PR #2704, opened 2017-02-20
dlgScatterPlot

lilyclements:
In Progress

We need this ucrReceiver to inherit ucrCore


maxwellfundi:
Nice start here @lilyclements . I was heading to these dialogs and it would be great to have a chat sometime on implementation of these plot dialogs.

lilyclements:
Interesting! I'll be in the office in half an hour so I'll take a look at it then if that's okay.
Make sure you check the spreadsheet to see if someone else is working on a dialog, or to put your name by a dialog however to avoid two people on one dialog

maxwellfundi:
@lilyclements  I was a little stuck with the setting operators here Danny  said I had to be careful with how i position them. Looks you have nailed this which is great!

The dialog produces a graph, which is okay but from the output window  the code looks funny with the geom function inappropriately positioned. 

`# dlgScatterPlot code generated by the dialog Scatter Plot
survey_temp <- InstatDataObject$get_data_frame(data_name="survey")
Scatter1 <- ggplot(data=survey_temp, mapping=aes(y=Field, x=Variety., colour="Yield"))
InstatDataObject$add_graph(data_name="survey", graph=Scatter1, graph_name="Scatter1")
InstatDataObject$get_graphs(graph_name="Scatter1", data_name="survey") + geom_point()`

The geom function should be after the ggplot function. Could you look at how to adjust the positions and see if this gets there?

lilyclements:
Yes I noticed that and I'm not sure how I did it! Do you know how? I've said to have the clsOverallfunction on the left hand side of the operator, however, clearly it thinks the clsOverallFunction is this line

maxwellfundi:
Yeah, I wonder why that is. It might be something we discuss when you guys come. I could also try  playing around and see if I get to somewhere. 

Enjoy your flight here and see you tomorrow :-)

instat/dlgScatterPlot.vb
```diff
@@ -133,6 +132,8 @@ Public Class dlgScatterPlot
         ucrBase.clsRsyntax.SetOperatorParameter(0, clsRFunc:=clsOverallFunction)
         ucrBase.clsRsyntax.SetOperatorParameter(1, clsRFunc:=clsRGeomScatterplotFunction)
 
+        ucrBase.clsRsyntax.SetAssignTo("last_graph", strTempDataframe:=ucrSelectorForScatter.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempGraph:="last_graph")
```
lilyclements: @maxwellfundi so this is good to know - the method we discussed at dinner does sort it out. I set it to ucrBase.clsRSyntax which my operator is set to.
I imagine if I were to give my operator a class and set it to that class then it would work like this too?

instat/dlgScatterPlot.vb
```diff
@@ -148,7 +149,7 @@ Public Class dlgScatterPlot
     End Sub
 
     Private Sub SetRCodeForControls(bReset As Boolean)
-        ucrSaveGraph.SetRCode(clsOverallFunction, bReset)
+        ucrSaveGraph.SetRCode(ucrBase.clsRsyntax.clsBaseOperator, bReset)
```
maxwellfundi: Could you tell me why you doing this line? Why did you change this to baseoperator other than overall function?

maxwellfundi:
@lilyclements this now works the way it is expected, but i have a slight problem when i click reset. I doesnt seem to reset the ucrSave. 

Could you have a look at that??

instat/dlgScatterPlot.vb
lilyclements: I changed this because the ucrSaveGraph is connected to the base operator (since the setassignto is on the ucrBase.clsRsyntax line)
You are right this does not reset the check box, but this does reset the text in the textbox at least. I need to fix it so that it does reset if the check box is checked. However, I've noticed this problem in other dialogs. I'm not sure if people have managed to fix that or if it is a problem overall. 

In the code - having this as to base operator instead of to the clsOverallFunction means that I have last_graph where I have put myName in the code (the rest runs the myName name fine)

InstatDataObject$add_graph(data_name="Blocking", graph=myName, graph_name="myName")
InstatDataObject$get_graphs(graph_name="myName", data_name="Blocking")

We want it to have myName not last_graph if I change the name to myName

PR #2710, opened 2017-02-21
Start of Rains

lilyclements:
For next release on Friday in preparation for the conference next week I have fixed a couple of small bugs on this.
I have also edited the main form to make this dialog enabled.

I would say to not review this (yet?)

### DIFF
--- a/instat/dlgStartofRains.vb
+++ b/instat/dlgStartofRains.vb
@@ -12,7 +12,9 @@
 '
 ' You should have received a copy of the GNU General Public License k
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
 Imports instat.Translations
+
 Public Class dlgStartofRains
     Public clsAddKey, clsDayFromAndTo, clsRainyDays, clsRollingSum, clsMinimumRainfall, clsRollingRainDays, clsFirstDOYPerYear, clsDryPeriodTen, clsWithinThirtyDays, clsDrySpell, clsYearGroupDaily, clsManipulationFirstDOYPerYear As New RFunction
     Public clsDryPeriodTenList, clsSubRainDays1, clsSubRainDays2, clsCombinedFilter, clsCombinedList, clsWetSpell, clsWetSpellList, clsDPOverallInterval, clsDPOverallIntervalList, clsDPRain, clsDPRainList, clsDPRainInDays As New RFunction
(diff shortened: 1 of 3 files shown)


stevekogo:
@lilyclements Thanks for getting the dialog to this stage. Despite the periodic bugs when you try re-running the dialog, I think we should eanable autofill for the receivers. Also, I don't like the property of overwriting the previous calculated column. Using the new column name to generate new ones seems sensible.

lilyclements:
Great, would you mind explaining to me how to do the new autofill?

lilyclements:
Also, I don't know how to generate different datasets each time since this dialog doesn't run a save button?

Having a few issues with autofill. Either I've used it wrong or I've run into a problem.

My doy column is called doy_366 and so it doesn't automatically fall into the doy field when doing define climatic data, and so when I press OK it doesn't actually run the doy in the R Code
Then if I open up my Start of Rains dialog, I can't manually put it in as it will only display datasets which are defined as "doy" columns

lilyclements:
@stevekogo what are your thoughts on this now? Is it ready for Danny to review?

instat/dlgStartofRains.vb
```diff
@@ -102,11 +102,11 @@ Public Class dlgStartofRains
         clsDPRainList.SetRCommand("list")
 
         ucrReceiverDate.AddIncludedMetadataProperty("Climatic_Type", {Chr(34) & "date" & Chr(34)})
-        '        ucrReceiverDOY.AddIncludedMetadataProperty("Climatic_Type", {Chr(34) & "doy" & Chr(34)})
+        ucrReceiverDOY.AddIncludedMetadataProperty("Climatic_Type", {Chr(34) & "doy" & Chr(34)})
```
stevekogo: this should also take in doy_366

dannyparsons: Do we have btoh doy and doy_366 climatic types?

lilyclements: By putting doy this does take in doy_366, or it did for me when I defined it in the Climatic Definitions (the version which has just been merged, that is!)

stevekogo:
@lilyclements There seems to be a conflict with the designer that need to be fixed. Please address this then @dannyparsons can merge this as it is.

PR #2711, opened 2017-02-21
updating dlgAddkey

maxwellfundi:
@lilyclements  So how did you resolve the resetting of the ucrinputs in your previous problem you earlier told me. The reset here does work so well for the ucrinput

### DIFF
--- a/instat/dlgAddKey.Designer.vb
+++ b/instat/dlgAddKey.Designer.vb
@@ -22,24 +22,15 @@ Partial Class dlgAddKey
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()>
     Private Sub InitializeComponent()
-        Me.lblKeyName = New System.Windows.Forms.Label()
         Me.cmdCheckUnique = New System.Windows.Forms.Button()
         Me.lblKeyColumns = New System.Windows.Forms.Label()
+        Me.lblKeyName = New System.Windows.Forms.Label()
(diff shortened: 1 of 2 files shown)


maxwellfundi:
Yeah it is ready for review. I am not sure i understand. What do you feel needs to  addressed on the layout?

lilyclements:
The receiver to move so the top of the receiver aligns with the top of the selector's list box, and then the label for that to move accordingly. 
The Key Name: needs to line up with it's input
The Check Unique's input length could end where the comment box on the base ends

maxwellfundi:
@lilyclements  i have now addressed the issues on layout. Thanks.

lilyclements:
Great, the labels still need to lined up slightly. "Key Columns" at say 254, 47
"Key Name" at about 10, 218

maxwellfundi:
I have now moved the labels to your preferred locations and they really have tiny visible change which doesnt make a difference really. 
These are too precise measurements and I guess we might need a document to guide us on this or we agree as a team else these are quite off head measurements will be inconsistent everywhere.

lilyclements:
We generally always have the ucrDataframe at 10,10
The first ucrReceiver at x, 60
The label for that directly above it at x, 47
On any I have been reviewing, I've been making sure we line everything up so it's all in the same positions so we are slowly but surely getting that consistency as we go through every dialog

lilyclements:

PR #2718, opened 2017-02-23
Ucrgrid basic Structure

Lunalo:
This is my thoughts on how we should implement .

Both The factor and General grid should inherit from ucrGrid hence being able to override methods of ucrGrid. 

This is in progress


dannyparsons:
Just wondering if we need both ucrGrid and ucrGeneral? 
Maybe you could explain what you would see as the different uses of these two?

Lunalo:
@dannyparsons 

I was basing on the libraries used but after further thinking I think it should be only one control ?

dannyparsons:
I think so yes. One standard grid control and one specific factor grid makes sense.

Lunalo:
@dannyparsons 

So I think the specific factor grid should inherit from standard grid control ?

dannyparsons:
Yes. Any specific grid control should inherit from a base grid control.

PR #2741, opened 2017-03-02
Corruption Dialogs

lilyclements:
Here I have made
dlgFitCorruptionModel
sdgFitCorruptionModelDisplay

I will make the other dialog shortly


instat/dlgFitCorruptionModel.vb
```diff
@@ -0,0 +1,160 @@
...
+        strSelectedColumn = strColumn
+        strSelectedDataFrame = strDataFrame
+        bUseSelectedColumn = True
+    End Sub
+
+    Private Sub SetDefaultColumn()
+        ucrSelectorFitModel.ucrAvailableDataFrames.cboAvailableDataFrames.SelectedItem = strSelectedDataFrame
+        ucrReceiverControlVariables.Add(strSelectedColumn, strSelectedDataFrame)
+        ucrReceiverIndicators.Add(strSelectedColumn, strSelectedDataFrame)
+        ucrReceiverOutput.Add(strSelectedColumn, strSelectedDataFrame)
```
lilyclements: I'm not sure if this is correct, and I think that's because I don't fully understand! Would I need three different strSelectedColumn's

instat/dlgFitCorruptionModel.vb
```diff
@@ -0,0 +1,160 @@
...
+
+    Private Sub SetRCodeForControls(bReset As Boolean)
+        SetRCode(Me, ucrBase.clsRsyntax.clsBaseFunction, bReset)
+    End Sub
+
+    Private Sub TestOKEnabled()
+        If Not ucrReceiverOutput.IsEmpty AndAlso Not ucrReceiverIndicators.IsEmpty AndAlso Not ucrReceiverControlVariables.IsEmpty AndAlso ucrSaveCorruptionModel.IsComplete Then
+            ucrBase.OKEnabled(True)
+            ucrBase.clsRsyntax.RemoveParameter("formula")
+            ucrBase.clsRsyntax.clsBaseFunction.AddParameter("formula", clsROperatorParameter:=clsModel)
```
lilyclements: I'm not sure how to treat operators in the new system still, I've had this problem in other dialogs so it would be helpful to go over it sometime

instat/dlgFitCorruptionModel.vb
```diff
@@ -0,0 +1,160 @@
...
+    End Sub
+
+    Private Sub Controls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverOutput.ControlContentsChanged, ucrReceiverIndicators.ControlContentsChanged, ucrReceiverControlVariables.ControlContentsChanged, ucrSaveCorruptionModel.ControlContentsChanged
+        TestOKEnabled()
+    End Sub
+
+    Private Sub cmdDisplayOptions_Click(sender As Object, e As EventArgs) Handles cmdDisplayOptions.Click
+        '        sdgFitCorruptionModelDisplay.SetRFunction(ucrBase.clsRsyntax.clsBaseFunction, bResetSubdialog)
+        bResetSubdialog = False
+        sdgFitCorruptionModelDisplay.Show()
```
lilyclements: This dialog is currently empty, just thought I should put it in while I'm adding windows forms

instat/dlgFitCorruptionModel.vb
```diff
@@ -0,0 +1,160 @@
...
+        'ucrReceiverOutput.SetParameter(New RParameter("", 1))
+        'ucrReceiverOutput.SetParameterIsString()
+        ucrReceiverOutput.Selector = ucrSelectorFitModel
+        ucrReceiverOutput.SetIncludedDataTypes({"integer", "numeric", "logical"})
+
+        'ucrReceiverControlVariables.SetParameter(New RParameter("", 2))
+        'ucrReceiverControlVariables.SetParameterIsString()
+        ucrReceiverControlVariables.Selector = ucrSelectorFitModel
+
+        'ucrReceiverIndicators.SetParameter(New RParameter("", 3))
```
lilyclements: I'm not sure what the parameters are for the two explanatory variables (presumably multiple receiver ones are the explanatory variables and the ucrReceiverOutput is the response/y)

dannyparsons: yes that's right. So we need an Operator for the formula - or probably two operators because it's going to be:
y ~ x1 + x2 + x3 where the x's come from both the multiple receivers.

instat/dlgFitCorruptionModel.vb
dannyparsons: I think I just need default columns for the control variables. So it is just one, but since it's a multiple, the default could be a list. Possibly the Add method could take a String() argument instead?

instat/dlgFitCorruptionModel.vb
dannyparsons: Hopefully it works almost the same as RFunctions. This looks right.

instat/dlgFitCorruptionModel.vb
```diff
@@ -0,0 +1,160 @@
...
+            ucrBase.clsRsyntax.SetFunction("lm")
+        Else
+            ucrBase.clsRsyntax.SetFunction("glm")
+        End If
+    End Sub
+
+    Private Sub SetFormula()
+        If Not ucrReceiverOutput.IsEmpty AndAlso Not ucrReceiverIndicators.IsEmpty AndAlso Not ucrReceiverControlVariables.IsEmpty Then
+            ucrBase.clsRsyntax.RemoveParameter("formula")
+            clsModel.AddParameter(iPosition:=0, clsRFunctionParameter:=ucrReceiverOutput.GetVariables())
```
dannyparsons: We should be able to do this part through the new system. Parameters for models can have names, they just don't display, so you can use any name. And obviously the parameter positions must be right for model parameters.

instat/dlgFitCorruptionModel.vb
dannyparsons: Could this easily be linked to the same sub dialog we have in other model dialogs? That would be nice since we want the same things I think.

lilyclements: Okay, I guess it would be sdgSimpleRegOptions (the one that other glm/lm dialogs link to). Sorry for this, is it easy for me to get rid of this new sub?

dannyparsons: Yeah I think you can just delete the new one.

instat/dlgDefineCRI.vb
```diff
@@ -0,0 +1,172 @@
...
+        '  ucrBase.iHelpTopicID =
+
+        'Selector
+        '        ucrSelectorCRI.SetParameter(New RParameter("", 0))
+        ucrSelectorCRI.SetParameterIsString()
+
+        'Receivers
+        '        ucrReceiverRedFlag.SetParameter(New RParameter("", 1))
+        ucrReceiverRedFlag.SetParameterIsString()
+        ucrReceiverRedFlag.Selector = ucrSelectorCRI
```
lilyclements: So, here I want to say if the item in the receiver is of type logical/numeric then display nud, otherwise show grid.
I've currently done this in the dialog later on because I wasn't sure how to say objValues:={"numeric", "logical"} as that's not a value of the object?

instat/dlgFitCorruptionModel.vb
lilyclements: Okay sure, why would we only want this for teh Control Variables and not the Indicators?
What are the indicators?

instat/dlgDefineCRI.vb
dannyparsons: That's fine for now I think.

instat/dlgFitCorruptionModel.vb
dannyparsons: There are some defaults for control variables that we can suggest, but the indicators are chosen by the user.

Lunalo:
@dannyparsons I guess these are for fitting models.  Could you share the specific models with us to make the reviewing process  easier.

dannyparsons:
Yes, thanks for asking. For the logical response variable case it should be something like:
`glm(y ~ x1 + x2 + x3 + z1 + z2 + z3, data = data, family = "binomial")`

and for numeric response variables:

`lm(y ~ x1 + x2 + x3 + z1 + z2 + z3, data = data)`

where `xi` are control variables and `zi` are indicators and `y` is corruption output.

lilyclements:
1. Should I put a ucrDistributions in?
2. How can I say to have `xi + zi` for i at any length?

instat/dlgFitCorruptionModel.vb
lilyclements: I don't think I understand. Do you mean something like
clsModel.SetParameter(New RParameter("y", 0))
Where "y" is the parameter I have associated to ucrReceiverOutput?

Lunalo:
@dannyparsons 

Thanks very much.
Logical response case:
 Why haven't we used family=binomial(link="logit")  since we are doing logistic regression?

dannyparsons:
@Lunalo I believe they are equivalent because the default link is "logit" for binomial. However, to be consistent with other dialog and to make it clear I think you are right. Let's do it as you suggest.

dannyparsons:
@lilyclements We don't need a distribution control because we only want to allow logistic or linear regression and it's determined by the explanatory variable anyway.

lilyclements:
The default for glm is gaussian, should I put in to have `family = binomial` or is that sorted by variables automatically?

dannyparsons:
I was talking about the default for link. I think we put `family=binomial(link="logit")` as John suggests.

dannyparsons:
Since there's vbproj changes I'll accept this now and we can work on it again later.

PR #2744, opened 2017-03-02
Tidying up

Lunalo:


Lunalo:

This is done already . So you can review these small edits

maxwellfundi:
@dannyparsons  this is ready to be merged

instat/dlgHideShowColumns.Designer.vb
```diff
@@ -57,7 +58,7 @@ Partial Class dlgHideShowColumns
         'lblHiddenColumns
         '
         Me.lblHiddenColumns.AutoSize = True
-        Me.lblHiddenColumns.Location = New System.Drawing.Point(248, 26)
+        Me.lblHiddenColumns.Location = New System.Drawing.Point(249, 38)
```
lilyclements: This needs to move down so it is like this on the receiver
image

instat/dlgHideShowColumns.vb
```diff
@@ -18,6 +18,7 @@ Imports instat.Translations
 Public Class dlgHideShowColumns
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
+    Private clsDefaultFunction As New RFunction
```
lilyclements: You can also update in SetDefaults so you are not "As New RFunction" in there, and additionally do not need to clone

instat/dlgView.Designer.vb
```diff
@@ -41,7 +41,7 @@ Partial Class dlgView
         'lblSelected
         '
         Me.lblSelected.AutoSize = True
-        Me.lblSelected.Location = New System.Drawing.Point(254, 46)
+        Me.lblSelected.Location = New System.Drawing.Point(248, 36)
```
lilyclements: As in HideShow, this needs to move to the appropriate place

instat/dlgHideShowColumns.Designer.vb
lilyclements: This needs to move so that it is more like

image

instat/dlgHideShowColumns.vb
lilyclements: While on this dialog, you can remove in the Set Defaults the "As New RFunction" and change that to the newest system and remove .clone/etc

dannyparsons:
I think this can be fixed by having a default state set for the panel, since it doesn't always have a value. Happy for you to find an alternative solution, at least temporarily.

Lunalo:
@dannyparsons 

The current version is working fine. 
This is not ready since we are still working on it. So we can leave it for now.

Lunalo:

 I think debugging tools have helped me identify the developer error and also the cause of breaking of this dialog. Although I managed to fix this dialog, I think there is better way of doing it. Someone can look at this code and test after which you can suggest the improvement. Remember this is tricky dialog, We have one main Rfunction with three R commands.

Thanks.

instat/dlgView.Designer.vb
lilyclements: This is too high, try moving this to a height of 45. Additionally, the name of this label should change to say "Variables to View:" as discussed in pull request #2862

instat/dlgView.vb
```diff
@@ -57,13 +55,13 @@ Public Class dlgView
         ucrPnlDisplayWindow.AddRadioButton(rdoDispSepOutputWindow)
         ucrPnlDisplayWindow.AddToLinkedControls(ucrChkSpecifyRows, {rdoDispOutputWindow}, bNewLinkedHideIfParameterMissing:=True, bNewLinkedAddRemoveParameter:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=True)
         ucrPnlDisplayWindow.AddFunctionNamesCondition(rdoDispSepOutputWindow, "View")
-
+        ucrPnlDisplayWindow.bAllowNonConditionValues = True
```
lilyclements: Okay, so I recognise that this line fixes the bug but I'm not sure if this is the correct method to do so. Perhaps I should explain what I think this control is doing and you can explain too and we can see?

To my knowledge, this control is saying to "Allow parameters/functions/values for this panel which are not specified".
If the first rdo is checked then it gives the function "View" 
If the second is checked then it gives the function of either
1. Nothing (e.g. if Specify is not checked)
2. Head()
3. Tail()

So, we shouldn't have bAllowNonConditionValues here because we don't need to allow for non conditional values since both radio buttons have condition values associated to them. The developer error without this line is not coming from the first rdo because it has a parameter, whereas the second rdo hasn't been set up properly as it doesn't have a parameter associated with it?

This is my understanding of the situation, I can see where I might be wrong but do not fully understand why
What is your understanding of this?

Lunalo:
@lilyclements 

You can now test the view dialog

instat/dlgView.vb
```diff
@@ -141,7 +145,7 @@ Public Class dlgView
         TestOKEnabled()
     End Sub
 
-    Private Sub FunctionControls_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlDisplayFrom.ControlValueChanged, ucrPnlDisplayWindow.ControlValueChanged, ucrChkSpecifyRows.ControlValueChanged
+    Private Sub FunctionControls_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlDisplayFrom.ControlValueChanged, ucrPnlDisplayWindow.ControlValueChanged, ucrChkSpecifyRows.ControlValueChanged, ucrReceiverView.ControlValueChanged
         SetIcallType()
```
lilyclements: I think the SetIcallType sub should be called SetiCallType

instat/dlgView.vb
```diff
@@ -37,16 +36,15 @@ Public Class dlgView
 
     Public Sub SetRCodeForControls(bReset As Boolean)
         ucrNudNumberRows.Maximum = Decimal.MaxValue
```
lilyclements: Why is this being done here out of interest?

instat/dlgView.vb
```diff
@@ -37,16 +36,15 @@ Public Class dlgView
 
     Public Sub SetRCodeForControls(bReset As Boolean)
```
lilyclements: I don't see why this needs to be a public sub?

instat/dlgView.vb
```diff
@@ -37,16 +36,15 @@ Public Class dlgView
...
         ucrNudNumberRows.Maximum = Decimal.MaxValue
-        SetRCode(Me, clsMainFunction, bReset)
+        SetRCode(Me, ucrBase.clsRsyntax.clsBaseFunction, bReset)
         DataFrameLength()
     End Sub
 
     Private Sub SetDefaults()
-        Dim clsDefaultFunction As New RFunction
+        Dim clsMainFunction = New RFunction
```
lilyclements: I don't think this needs to be a dim?

instat/dlgView.vb
```diff
@@ -73,11 +73,14 @@ Public Class dlgView
         ucrChkSpecifyRows.SetText("Specify Rows")
         ucrChkSpecifyRows.AddToLinkedControls(ucrPnlDisplayFrom, {True}, bNewLinkedHideIfParameterMissing:=True, bNewLinkedAddRemoveParameter:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=rdoTop)
         ucrChkSpecifyRows.AddToLinkedControls(ucrNudNumberRows, {True}, bNewLinkedHideIfParameterMissing:=True, bNewLinkedAddRemoveParameter:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=6)
+        ucrChkSpecifyRows.AddFunctionNamesCondition(True, {"head", "tail"})
+        ucrChkSpecifyRows.AddFunctionNamesCondition(False, {frmMain.clsRLink.strInstatDataObject & "$get_columns_from_data"})
+        ucrChkSpecifyRows.bAllowNonConditionValues = True
```
lilyclements: Why are you running this here?

instat/dlgView.vb
Lunalo: We want to do it every time R code for controls is being set

instat/dlgView.vb
Lunalo: true

instat/dlgView.vb
Lunalo: @lilyclements 

You can refer to Danny's comments on the #2883

instat/dlgView.vb
Lunalo: @lilyclements  Why do you think so?

instat/dlgView.vb
dannyparsons: Yes we discussed this is acceptable for now since this is how the system works currently.

instat/dlgView.vb
dannyparsons: Why do you need to do this every time the R code is set?

Lunalo: We set the maximum of the nud to maximum. Then when we set Rcode we reset it to correct maximum after which the dataframeLength is called.

This is done sequentially because if we set the value before setting the maximum we may get an error

dannyparsons: Ok yes I remember this now. That's good.

instat/dlgView.vb
lilyclements: because when we run the code it's `.iCallType = ...`
I can see why the "I" could be capitalised but the "C" needs to be capitalised.

instat/dlgView.vb
Lunalo: This is fixed

Lunalo:
@lilyclements 

Those changes you suggested are fixed

Lunalo:
@lilyclements  Thanks We have default for col_names=c() in our instat method

Lunalo:
@lilyclements  this last commit fixes the problem. Thanks to @dannyparsons .

lilyclements:
@Lunalo Is this ready for a re-review?

stevekogo:
@Lunalo respond to @lilyclements query on whether this is ready for a re-review. I think she is interested in reviewing it.

Lunalo:
@lilyclements 

This ready

dannyparsons:
@Lunalo This is why we don't work on more than one dialog on a branch :)

Lunalo:
@lilyclements 

It is now fixed

lilyclements:
Great, @dannyparsons this is ready for you to review

dannyparsons:
The View dialog could be nicely redone with the new way of doing multiple functions in dialogs. However, this is an improvement on the original so I can accept it now. If any bugs appear we should switch this to the new way of managing multiple functions.

PR #2746, opened 2017-03-03
Permute Column New Implementation.

AlexSananka:
I have made changes to this dialogue,I think it's now updated. @africanmathsinitiative/developers Can some one review it.

### DIFF
--- a/instat/dlgPermuteColumn.vb
+++ b/instat/dlgPermuteColumn.vb
@@ -16,7 +16,7 @@
 
 Imports instat.Translations
 Public Class dlgPermuteColumn
-    Private clsSetSampleFunc, clsSetSeedFunc, clsOverallFunction As New RFunction
+    Private clsSetSampleFunction, clsSetSeedFunction, clsOverallFunction As New RFunction
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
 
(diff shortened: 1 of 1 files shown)


instat/dlgPermuteColumn.vb
```diff
@@ -68,12 +67,12 @@ Public Class dlgPermuteColumn
         ucrBase.iHelpTopicID = 66
         ucrReceiverPermuteRows.Selector = ucrPermuteRowsSelector
         ucrReceiverPermuteRows.SetMeAsReceiver()
+        ucrChkSetSeed.AddFunctionNamesCondition(False, "set.seed")
```
lilyclements: @dannyparsons what is the difference between running

ucrChkSetSeed.AddFunctionNamesCondition(False, "set.seed")

And

ucrChkSetSeed.SetParameter(ucrNudSetSeed.GetParameter(), bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)

lilyclements: @dannyparsons what is the difference between running

ucrChkSetSeed.AddFunctionNamesCondition(False, "set.seed")

And

ucrChkSetSeed.SetParameter(ucrNudSetSeed.GetParameter(), bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)

dannyparsons: The add function name is done automatically for you when you set the parameter in that way. I recommend doing it this way and only using `AddFunctionNamesCondition` when you need to do something different to the standard.

PR #2747, opened 2017-03-03
Define Corruption

lilyclements:
In progress


instat/dlgDefineCorruption.vb
```diff
@@ -15,152 +15,182 @@
...
+        ucrReceiverWinnerName.Tag = "winner_name"
+        ucrReceiverWinnerCountry.Tag = "winner_country"
+        ucrReceiverContractValue.Tag = "original_contract_value"
+        ucrReceiverFiscalYear.Tag = "fiscal_year"
+
+        'ucrChk
+        ucrChkAutoGenerate.SetText("Auto-Generate")
+        ucrChkAutoGenerate.SetParameter(New RParameter("auto_generate", 1))
+        ucrChkAutoGenerate.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
+        ucrChkAutoGenerate.SetRDefault("TRUE")
```
lilyclements: @dannyparsons Is this our R Default or just out Default?

dannyparsons: At the moment it's both. But I think I want to change this so that there is no R default. So that the argument is always shown in the command. So keep our default as TRUE and remove the R default.

instat/sdgCorruptionCountryLevelColumns.vb
```diff
@@ -1,25 +1,67 @@
...
+        bControlsInitialised = True
+    End Sub
+
+    Public Sub SetRFunction(clsNewRFunction As RFunction, Optional bReset As Boolean = False)
+        If Not bControlsInitialised Then
+            InitialiseControls()
+        End If
+        clsCountryLevel = clsNewRFunction
+        SetRCode(Me, clsCountryLevel, bReset)
+    End Sub
```
lilyclements: Would I add in the autofillreceivers sub here? Or do I amend the one on the main dialog to contain this sub as well?

lilyclements: Would I add in the autofillreceivers sub here? Or do I amend the one on the main dialog to contain this sub as well?

instat/sdgCorruptionCountryLevelColumns.vb
```diff
@@ -1,25 +1,67 @@
...
+    Public bControlsInitialised As Boolean = False
+    Public clsCountryLevel As New RFunction
+    Dim lstReceivers As New List(Of ucrReceiverSingle)
+    Dim lstRecognisedTypes As New List(Of KeyValuePair(Of String, List(Of String)))
+
+    Private Sub sdgCorruptionCountryLevelColumns_Load(sender As Object, e As EventArgs) Handles Me.Load
+        autoTranslate(Me)
+    End Sub
+
+    Public Sub InitialiseControls()
```
lilyclements: Where would I reset the selector and do .SetMeAsReceiver for this sub?

dannyparsons: You don't need to reset the selector since that comes from the code passed in to the sub dialog. Setting the current receiver could be done when SetRFunction is called with Reset is True.

instat/sdgCorruptionCountryLevelColumns.vb
dannyparsons: Yes there should be an auto fill method here.

instat/sdgCorruptionCalculatedColumns.vb
```diff
@@ -1,3 +1,168 @@
...
+        ucrCurrentReceiver = ucrCalculatedColumnsSelector.CurrentReceiver
+
+        For Each ucrTempReceiver As ucrReceiver In lstReceivers
+            ucrTempReceiver.SetMeAsReceiver()
+            lstRecognisedValues = GetRecognisedValues(ucrTempReceiver.Tag)
+
+            If lstRecognisedValues.Count > 0 Then
+                For Each lviTempVariable As ListViewItem In ucrCalculatedColumnsSelector.lstAvailableVariable.Items
+                    For Each strValue As String In lstRecognisedValues
+                        If Regex.Replace(lviTempVariable.Text.ToLower(), "[^\w]|_", String.Empty).Contains(strValue) Then
```
lilyclements: Regex is under the caret if I add in the Autofill Receivers and I can't work out why. Do you know how I can solve this?

instat/dlgDefineCorruption.vb
```diff
@@ -15,152 +15,196 @@
...
+
+    Private Sub SetRCodesforReceivers(bReset As Boolean)
+        Dim ucrTempReceiver As ucrReceiver
+        For Each ucrTempReceiver In lstReceivers
+            ucrTempReceiver.SetRCode(clsTypesFunction, bReset)
+        Next
+    End Sub
+
+    Private Sub SetRSelector()
+        Dim ucrTempReceiver As ucrReceiver
```
lilyclements: Presumably I need to run something similar to this. Would I run this in the sub dialog or in the main dialog?
Perhaps should I run a sub similar to this in the sub dialogs, and then call that sub in the SetRCode in the main dialog?

lilyclements:

instat/dlgDefineCorruption.vb
```diff
@@ -15,152 +15,198 @@
...
+
+    Public Sub SetRCodeForControls(bReset As Boolean)
+        ucrDefineCorruptionSelector.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
+        sdgCorruptionCountryLevelColumns.ucrCountryLevelSelector.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
+        sdgCorruptionCalculatedColumns.ucrCalculatedColumnsSelector.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
+        ucrChkAutoGenerate.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
+        SetRCodesforReceivers(bReset)
+    End Sub
+
+    Private Sub SetRCodesforReceivers(bReset As Boolean)
```
dannyparsons: Is this needed to be a separate sub?

instat/dlgDefineCorruption.vb
dannyparsons: I don't think this needs to be a separate sub either

PR #2750, opened 2017-03-06
work on import from climsoft dialog

dannyparsons:


dannyparsons:
@stevekogo The basis of this is now done. The main dialog needs to be set up properly. I believe the link to the sub dialog is set up. The SQL codes run in the main dialog are there, but you should check these with @mhabimana. We do not have a CLIMSOFT database to test them.

I have sent you a separate file with the R script that @mhabimana worked on with @volloholic which shows how to run SQL commands on a CLIMSOFT database from R. This needs to be adapted to go in the `import_from_climsoft` method, which is currently blank, and then this will be run from the main dialog.

PR #2758, opened 2017-03-08
Regular sequence

AlexSananka:
I think this has made a great improvement though am still working on it.


AlexSananka:
@africanmathsinitiative/developers just before this becomes ready for review it has a problem on appending the new column.On first load it doesn't append but when you run the second time it appends.Could anyone have a suggestion on how to solve this?

Lunalo:
@AlexSananka 

I am carefully looking at the code for any error.

AlexSananka:
@Lunalo That will be great.

instat/dlgRegularSequence.designer.vb
```diff
@@ -232,98 +201,132 @@ Partial Class dlgRegularSequence
...
         Me.txtMessage.TabIndex = 5
         '
+        'grpSequenceType
+        '
+        Me.grpSequenceType.Controls.Add(Me.rdoNumeric)
+        Me.grpSequenceType.Controls.Add(Me.rdoDates)
+        Me.grpSequenceType.Controls.Add(Me.ucrPnlSequenceType)
+        Me.grpSequenceType.Location = New System.Drawing.Point(16, 53)
+        Me.grpSequenceType.Name = "grpSequenceType"
+        Me.grpSequenceType.Size = New System.Drawing.Size(199, 52)
```
lilyclements: This should have the same length (and starting point therefore) as the other grpbox

instat/dlgRegularSequence.designer.vb
```diff
@@ -120,64 +75,88 @@ Partial Class dlgRegularSequence
...
         Me.grpSequenceDefinition.TabStop = False
         Me.grpSequenceDefinition.Tag = "Sequence_definition"
         Me.grpSequenceDefinition.Text = "Sequence Definition"
         Me.grpSequenceDefinition.UseCompatibleTextRendering = True
         '
+        'ucrNudNumberOfDecimalPlaces
+        '
+        Me.ucrNudNumberOfDecimalPlaces.DecimalPlaces = New Decimal(New Integer() {0, 0, 0, 0})
+        Me.ucrNudNumberOfDecimalPlaces.Increment = New Decimal(New Integer() {1, 0, 0, 0})
+        Me.ucrNudNumberOfDecimalPlaces.Location = New System.Drawing.Point(142, 180)
```
lilyclements: This should move a little to around 143, 178

instat/dlgRegularSequence.designer.vb
```diff
@@ -120,64 +75,88 @@ Partial Class dlgRegularSequence
...
+        Me.ucrNudNumberOfDecimalPlaces.Name = "ucrNudNumberOfDecimalPlaces"
+        Me.ucrNudNumberOfDecimalPlaces.Size = New System.Drawing.Size(50, 20)
+        Me.ucrNudNumberOfDecimalPlaces.TabIndex = 13
+        Me.ucrNudNumberOfDecimalPlaces.Value = New Decimal(New Integer() {0, 0, 0, 0})
+        '
+        'ucrNudRepeatValues
+        '
+        Me.ucrNudRepeatValues.DecimalPlaces = New Decimal(New Integer() {0, 0, 0, 0})
+        Me.ucrNudRepeatValues.Increment = New Decimal(New Integer() {1, 0, 0, 0})
+        Me.ucrNudRepeatValues.Location = New System.Drawing.Point(99, 105)
```
lilyclements: These four nuds and the input need to have equal spacing

instat/dlgRegularSequence.designer.vb
```diff
@@ -232,98 +201,132 @@ Partial Class dlgRegularSequence
...
-        Me.lblTimes1.TabIndex = 10
+        Me.lblTimes1.TabIndex = 8
         Me.lblTimes1.Tag = "Times"
         Me.lblTimes1.Text = "Time(s)"
         '
         'lblRepeatValues
         '
         Me.lblRepeatValues.AutoSize = True
-        Me.lblRepeatValues.Location = New System.Drawing.Point(12, 108)
+        Me.lblRepeatValues.Location = New System.Drawing.Point(6, 105)
```
lilyclements: This needs to be aligned with the other labels

instat/dlgRegularSequence.designer.vb
```diff
@@ -186,16 +165,6 @@ Partial Class dlgRegularSequence
...
-        Me.nudInStepsOf.Maximum = New Decimal(New Integer() {2147483647, 0, 0, 0})
-        Me.nudInStepsOf.Minimum = New Decimal(New Integer() {2147483647, 0, 0, -2147483648})
-        Me.nudInStepsOf.Name = "nudInStepsOf"
-        Me.nudInStepsOf.Size = New System.Drawing.Size(51, 20)
-        Me.nudInStepsOf.TabIndex = 7
-        Me.nudInStepsOf.Value = New Decimal(New Integer() {1, 0, 0, 0})
-        '
         'dtpSelectorA
         '
         Me.dtpSelectorA.Location = New System.Drawing.Point(98, 24)
```
lilyclements: These two dtpSelectors should be the same size

instat/dlgRegularSequence.vb
```diff
@@ -13,78 +13,117 @@
 '
 ' You should have received a copy of the GNU General Public License k
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
-Imports instat.Translations
-Imports RDotNet
-
+Imports instat
```
lilyclements: This should be removed

instat/dlgRegularSequence.vb
```diff
@@ -280,13 +214,37 @@ Public Class dlgRegularSequence
...
+        CheckSequenceLength()
     End Sub
 
+
+    Private Sub ucrNudRepeatValues_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrNudRepeatValues.ControlValueChanged
+        SetBaseFunction()
+        SetInStepsOfParameter()
+    End Sub
+
+    Private Sub ucrNewColumnName_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrNewColumnName.ControlContentsChanged, ucrNudTo.ControlContentsChanged, ucrNudFrom.ControlContentsChanged, ucrNudInStepsOf.ControlContentsChanged, ucrNudRepeatValues.ControlContentsChanged, ucrPnlSequenceType.ControlContentsChanged
```
lilyclements: This does not include everything in your TestOK

instat/dlgRegularSequence.designer.vb
```diff
@@ -22,18 +22,15 @@ Partial Class dlgRegularSequence
...
-        Me.nudNumberofDecimalPlaces = New System.Windows.Forms.NumericUpDown()
-        Me.nudTo = New System.Windows.Forms.NumericUpDown()
-        Me.nudFrom = New System.Windows.Forms.NumericUpDown()
-        Me.nudRepeatValues = New System.Windows.Forms.NumericUpDown()
         Me.ucrDataFrameLengthForRegularSequence = New instat.ucrDataFrameLength()
         Me.dtpSelectorB = New System.Windows.Forms.DateTimePicker()
-        Me.nudInStepsOf = New System.Windows.Forms.NumericUpDown()
         Me.dtpSelectorA = New System.Windows.Forms.DateTimePicker()
         Me.lblLength = New System.Windows.Forms.Label()
         Me.chkDefineAsFactor = New System.Windows.Forms.CheckBox()
```
lilyclements: This control hasn't been updated

instat/dlgRegularSequence.vb
```diff
@@ -99,142 +138,38 @@ Public Class dlgRegularSequence
...
-                If (nudInStepsOf.Value = 1 AndAlso frmMain.clsInstatOptions.bIncludeRDefaultParameters) OrElse nudInStepsOf.Value <> 1 Then
-                    If nudTo.Value >= nudFrom.Value Then
-                        clsSeqFunction.AddParameter("by", nudInStepsOf.Value)
+            If ucrNudInStepsOf.GetText <> "" Then
+                If (ucrNudInStepsOf.Value = 1 AndAlso frmMain.clsInstatOptions.bIncludeRDefaultParameters) OrElse ucrNudInStepsOf.Value <> 1 Then
+                    If ucrNudTo.Value >= ucrNudFrom.Value Then
+                        clsSeqFunction.AddParameter("by", ucrNudInStepsOf.Value)
                     Else
-                        clsSeqFunction.AddParameter("by", -nudInStepsOf.Value)
+                        clsSeqFunction.AddParameter("by", -ucrNudInStepsOf.Value)
```
lilyclements: currently, if I try to go in reverse I get an error

instat/dlgRegularSequence.vb
```diff
@@ -280,13 +214,37 @@ Public Class dlgRegularSequence
...
-    Private Sub nudNumberofDecimalPlaces_TextChanged(sender As Object, e As EventArgs) Handles nudNumberofDecimalPlaces.TextChanged
-        nudFrom.DecimalPlaces = nudNumberofDecimalPlaces.Value
-        nudTo.DecimalPlaces = nudNumberofDecimalPlaces.Value
-        nudInStepsOf.DecimalPlaces = nudNumberofDecimalPlaces.Value
-        nudFrom.Increment = 10 ^ -(nudNumberofDecimalPlaces.Value)
-        nudTo.Increment = 10 ^ -(nudNumberofDecimalPlaces.Value)
-        nudInStepsOf.Increment = 10 ^ -(nudNumberofDecimalPlaces.Value)
+    Private Sub ucrSelectDataFrameRegularSequence_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSelectDataFrameRegularSequence.ControlValueChanged
+        If ucrSelectDataFrameRegularSequence.cboAvailableDataFrames.Text <> "" Then
+            ucrNudTo.Value = frmMain.clsRLink.GetDataFrameLength(ucrSelectDataFrameRegularSequence.cboAvailableDataFrames.Text)
```
lilyclements: Here we want
    Try
                ucrNudTo.Value = frmMain.clsRLink.GetDataFrameLength(ucrSelectDataFrameRegularSequence.cboAvailableDataFrames.Text)
            Catch ex As Exception
    End Try

Otherwise it crashes when the value is >100

instat/dlgRegularSequence.vb
```diff
@@ -13,78 +13,117 @@
...
+        ucrNudFrom.SetParameter(New RParameter("from", 0))
+        ucrNudTo.SetParameter(New RParameter("to", 1))
+        ucrNudRepeatValues.SetParameter(New RParameter("each", 1))
+        ucrNudInStepsOf.SetParameter(New RParameter("by", 2))
+        ucrNudInStepsOf.Minimum = 1
+        ucrNudInStepsOf.SetRDefault(1)
+        ucrNudInStepsOf.Maximum = Integer.MaxValue
+        ucrNudFrom.Minimum = 1
+        ucrNudFrom.Maximum = Integer.MaxValue
+        ucrNudTo.Minimum = 1
```
lilyclements: We want the maximum for ucrNudTo to be Integer.MaxValue here

lilyclements:
@AlexSananka how is this going? Is this ready for re-review?

AlexSananka:
@lilyclements This dialog is working fine except one issue which I really do not understand and I wish to get any suggestions from @africanmathsinitiative/developers on how to solve it.

On changing the repeat values Nud the dialog pops up an error on first load and does not assign To but when I run it the second time everything will work fine.

lilyclements:
Commented in the other Regular Sequence ( #2782 ) accidentally, but the same comments from there apply here

Put a stop in on the CheckSequenceLength() when the nuds change to work out the bug with the preview box

lilyclements:
So, I spoke to @rdstern and there are a few changes that can be made to this dialog. Mainly to get rid of the numeric up downs and instead replace them with ucrInputs (and the inputs have a `.SetValidationTypeAsNumeric` so that only numerics can be written in them). This means that the nud for decimal places can be removed

AlexSananka:
@lilyclements what could be the reason for changing this controls?If I may ask

dannyparsons:
Yes, you should be asking these questions - keep asking them. This is part of improving your understanding of decisions being made. Well done.

AlexSananka:
@rdstern just explained to me why we're changing this controls and it makes sense.Thanks

dannyparsons:
I agree but with one exception which is the nud for number of times to repeat values. This is always a positive integer so I think it is natural to keep the nud for this.

AlexSananka:
I see the point here,we need to keep this Nud for number of times to repeat values.We didn't notice this earlier thanks @dannyparsons

AlexSananka:
@africanmathsinitiative/developers  This is ready for review

instat/dlgRegularSequence.vb
```diff
@@ -13,78 +13,119 @@
 '
 ' You should have received a copy of the GNU General Public License k
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
-Imports instat.Translations
-Imports RDotNet
-
+
+Imports instat
```
lilyclements: This line can be removed

instat/dlgRegularSequence.vb
```diff
@@ -13,78 +13,119 @@
...
+        ucrPnlSequenceType.AddRadioButton(rdoDates)
+
+        ucrInputFrom.SetParameter(New RParameter("from", 0))
+        ucrInputFrom.AddQuotesIfUnrecognised = False
+        ucrInputFrom.SetValidationTypeAsNumeric()
+        ucrInputTo.SetParameter(New RParameter("to", 1))
+        ucrInputTo.AddQuotesIfUnrecognised = False
+        ucrInputTo.SetValidationTypeAsNumeric()
+        ucrNudRepeatValues.SetParameter(New RParameter("each", 1))
+        ucrNudRepeatValues.Minimum = 1
```
lilyclements: So, the other day I had a data frame where I wanted to repeat everything 2000 times. Currently you have not set a maximum which means the default is only 100

instat/dlgRegularSequence.vb
```diff
@@ -95,146 +136,44 @@ Public Class dlgRegularSequence
...
-            dtpSelectorA.Visible = False
-            dtpSelectorB.Visible = False
-            chkDefineAsFactor.Visible = False
-        ElseIf rdoDates.Checked Then
-            dtpSelectorA.Visible = True
-            dtpSelectorB.Visible = True
-            nudFrom.Visible = False
-            nudTo.Visible = False
-            chkDefineAsFactor.Visible = True
+            If ucrInputTo.GetText < ucrInputFrom.GetText Then
```
lilyclements: What if I wanted to decrease in values

instat/dlgRegularSequence.vb
```diff
@@ -280,13 +219,36 @@ Public Class dlgRegularSequence
...
+    Private Sub ucrInputFrom_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrInputFrom.ControlContentsChanged, ucrInputTo.ControlContentsChanged, ucrInputInStepsof.ControlContentsChanged
+        CheckSequenceLength()
+        'SetInStepsOfParameter()
+    End Sub
+
+    Private Sub ucrNewColumnName_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrNewColumnName.ControlContentsChanged, ucrInputTo.ControlContentsChanged, ucrInputFrom.ControlContentsChanged, ucrInputInStepsof.ControlContentsChanged, ucrNudRepeatValues.ControlContentsChanged, ucrPnlSequenceType.ControlContentsChanged, ucrSelectDataFrameRegularSequence.ControlContentsChanged
+        TestOKEnabled()
+    End Sub
+
+    Private Sub ucrInputFrom_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputInStepsof.ControlContentsChanged, ucrPnlSequenceType.ControlContentsChanged
```
lilyclements: Naming convention

instat/dlgRegularSequence.vb
rdstern: I don't have strong views on this, because it is filling the data frame anyway.

So (with my data frame of 8599 values, even with the existing system I was able to to do 1 to 1 in steps of 0, i.e. 1 8599 times.  It did it, because it was filling the data frame.
I then tried 1,2,1,2 4000 odd times and that works fine too.  So I am not too sure of the need for this feature and hence am quite happy whatever the limits.  I just left it at 1 for these examples.

instat/dlgRegularSequence.vb
AlexSananka: I hope you mean at `ucrInputInStepsof`

instat/dlgRegularSequence.vb
AlexSananka: Am fixing this

instat/dlgRegularSequence.vb
AlexSananka: @lilyclements I remember the data set you're saying(diamonds),But when I already fixed that and it does well for any data frames larger than 100 I guess.I don't see the need of setting the maximum here.

instat/dlgRegularSequence.vb
AlexSananka: OK,I keep forgetting this.

AlexSananka:
@lilyclements I have made the changes and it's now ready for a re-review

instat/dlgRegularSequence.vb
AlexSananka: This is now fixed

AlexSananka:
@africanmathsinitiative/developers this can be tested and reviewed,I think it's working fine.

instat/dlgRegularSequence.vb
lilyclements: @AlexSananka I see the confusion - a previous issue was that we could not import datasets over a length of 100 on this dialog. Apologies, I'm not referring to that situation. What I mean is, what if I want to repeat my values over 100 times. Currently there is a repeat limit of 100 on it and I think this could be increased to the VB maximum (something along the lines of `ucrNudRepeatValues.Maximum = Integer.MaxValue`)

So on Friday, I was working on a Climatic dataset where I was creating longitude and latitude columns. Say I want my longitude to go from 0 to 10 in one column and similarly my latitude to go from -50 to 50 in another column (both in steps of 0.05). However, I also wanted every combination possible, that is,
-50.00 for 0 longitude
-49.95 for 0 longitude
-49.90 for -0 longitude
...
49.95 for 0 longitude
50.00 for 0 longitude
-50.00 for 0.05 longitude
-49.95 for 0.05 longitude
...
50.00 for 0.05 longitude
etc etc, so in this situation I was repeating my longitude values 2000 times so I could go through all the latitude values at 0 degrees, then all latitude values at 0.05 degrees.

What I mean is, this is a situation where I would want to have a limit of over a repeat of 100. I could do this in R very easily, so I don't see why this shouldn't be possible in R-Instat as well.

instat/dlgRegularSequence.vb
```diff
@@ -94,147 +134,29 @@ Public Class dlgRegularSequence
...
-        Else
-            clsSeqFunction.RemoveParameterByName("from")
         End If
     End Sub
 
-    Private Sub nudTo_TextChanged(sender As Object, e As EventArgs) Handles nudTo.TextChanged
-        SetToParameter()
-        CheckSequenceLength()
-        TestOKEnabled()
-    End Sub
```
lilyclements: There is double spacing here, one of these lines can be removed

AlexSananka: I can't see this line,are you sure this is my current branch?

instat/dlgRegularSequence.vb
AlexSananka: I get your point now,Thanks

AlexSananka: This is fixed

instat/dlgRegularSequence.vb
lilyclements: It was line 141 but it says it's been removed in the commit below anyway!

AlexSananka:
@lilyclements  could you kindly specify the exact location you want the dtpSelectorB to be?

lilyclements:
Okay, so if you select ucrInputTo and then select the combo box underneath it. Then you should click on the align tops button. This means that they will have the same height.
You should select the ucrInputTo first. This is because you want the combo box's height to move to the ucrInputTo's height, since the ucrInputTo is in equal alignment with the controls above and below it.

AlexSananka:
Okay,I think that's sorted now.Thanks

lilyclements:
Great, I've just realised that you still need to address the problem that "To" should be able to value less than the "From"

AlexSananka:
I didn't get you on your last comment,pardon!!

lilyclements:
Apologies. So, you currently run on your TestOK that OK should not be enabled if the Preview is empty.

The preview is empty when I put the value in the "To" input box being less than the value in the "From" input box and so OK is not enabled. However, this situation should work for if I want to decrease in values. (e.g. I want to have my sequence going 10, 9, 8, 7, ...)

So, we need this to be fixed. As a side note, if the "To" input value is less than the "From" input value, then the value in the "ucrInputInStepsOf" should be made negative
Additionally, it should be made visible on the preview screen.

AlexSananka:
@lilyclements I don't think you tested well on this,you can decrease in values if you so wish,It's working fine.Just test it again and you will notice that it's Okay.We fixed this with @Lunalo

lilyclements:
@AlexSananka I don't think you understand what I am referring to.
Open the dialog.
Put a value in the TO input box (e.g. 10)
Now put a value in the FROM input box which is less than the value in the TO input box (e.g. 1)
Now, OK is not enabled but it should be.
I've explained already why this is.

Once you fix this problem in the code, reopen R-Instat and reopen the dialog with the same values as before.
You will get an error, I have explained above how to fix this.

lilyclements:
@AlexSananka look at an old version of this branch and look at the sub SetInStepsOfParameter()

AlexSananka:
Once "To" is less than "From" then "InstepsOf "should be negative and everything will work.

I intentionally removed that sub for SetInStepsOf since it's not necessary now that we changed from Nud's to ucrInputs.

lilyclements:
> I intentionally removed that sub for SetInStepsOf since it's not necessary now that we changed from Nud's to ucrInputs.

Even when the ucrInputInStepsOf was a nud, the SetInStepsOfParameter sub was not necessary. It was still possible for the user to make the value negative - I thought it was decided that the user can input a positive value and for the negation to be done internally. The user might not know to make it negative, so this way the user can put in any value to the input and it will automatically detect whether they want it to be negative or not. At least, this was my understanding.
I can see why the user inputting a negative value makes sense, however, to me, "In Steps Of" doesn't necessarily imply "In Positive Steps Of", it's just asking how much to 'jump' in between the "From" and "To"
@dannyparsons What are your thoughts?

instat/dlgRegularSequence.vb
lilyclements: Just saw this, sorry - I meant that it's called ucrInputFrom_ControlValueChanged and it's handling different ucrInputs to this and not handling their values changing

dannyparsons:
I think whether the in steps of should be typed as negative or whether the sign should be automatically detected from the From and To values is a good discussion to have.

However, @AlexSananka, the dialog was originally implemented with in steps of as always being positive with the sign automatically added in the code, as Lily pointed out. Happy for you to suggest alternatives to this, but you can't just change the implementation of a dialog, without telling anyone, and then expect a reviewer to somehow know this when testing.

Related to this, when someone says they've found a bug in code you've written, telling them it's not a bug is not an acceptable response. If someone reports a bug it's your responsibility to find out what has happened, and not just saying they shouldn't use the dialog in that way. We won't be able to say this to users once they start using our dialogs!

AlexSananka:
Question:Is there logic in moving backwards positively?
My thought is when the user is decreasing in values then they should now be aware that they are moving backwards and hence negatively.Does this make sense?
Look at how the sequence operates in R,I don't know if we are allowed to borrow the same idea.
But before this discussion,I can return the dialog to it's initial status.

dannyparsons:
I think both methods are logical. The advantage of having the Steps always positive is there's less chance of users getting stuck with OK disabled, as Lily did, because they didn't know they had to change the sign of Steps. Since the sign of steps is always determined by From and To I don't see why the user needs to type this. Others may think not allowing negative is also confusing though?

lilyclements:
I feel there is logic.

1. IN STEPS OF does not imply positive steps, it is just saying how to get from A to B. Steps are positive or negative

2. The user will most likely change the TO and the FROM before the IN STEPS OF. The IN STEPS OF is positive by default. So, the user can this way change the TO, change the FROM and see in the preview window that their IN STEPS OF is already working by decreasing values.


Perhaps there is cause to change the label name. When To > From then something like "Increasing in Steps of" and when To < From then "Decreasing in Steps of" (not this, but you get the gist)

Or perhaps we should allow both.
If To < From then
If InStepsOf < 0 then
instepsof.value remains the same
Else
instepsof.value is made negative
Etc etc


I'm on my phone so apologies for bad formatting in this message.

rdstern:
This is taking too much time from too many people on one point.  It is worth it if we recognise there are 3 separate issues here:
1) How do we review - from now on.
2) How do we move forward quickly on this particular point - of negative values (possibly) in this dialogue.
3) Can we all please be as attentive to details on all aspects of the dialogues and not just on this one point.

1. On the reviewing process we need more reviewers - like Lily.  She is doing too much of this work.  It is similar to reviewing papers - which some of us now also have to do.  It is not easy to do well.  It is also not always easy to react to a reviewer's comments, because the author often feels the reviewer has not understood or done their job well enough.  It is always good practice to give the reviewer the benefit of the doubt.  This is a good example, where the authors changed something in the code that was originally there for a good reason.  They then didn't comment on that point in the code, or in the message - namely that they had made this change in the way the dialogue was being used.  We have wasted a lot of time by poor communications.

2. OK, on the actual change that was made.  This software is designed for beginners to statistics packages.  This is an aspect that is done very differently in R-Instat to a spreadsheet.  And you may be surprised, but some of our users will not be so comfortable with negative numbers.  So I am not keen that you are forced to put in a negative number when the sequence goes downwards, before OK is enabled.  It is called "Step" in the dialogue I believe.  I am happy that you could either put a positive or a negative number and it will do the same thing in the sequence.  (David suggested that if the To is less than the From, then there is a label that automatically appears with the minus sign there!  I think we are getting too complicated there, because we would still want the user to enter either -1 or 1 and it works fine.

I don't even mind if you insist on having a minus there.  But if you do, then I strongly suggest that once the **From** and **To**, then it includes a -1 (say) as a suggestion to help the user.  (Though you will have to be careful, because what would you put there if **From is 10 and To is 9.5?**  It is quite easy for seemingly simple suggestions to occupy a lot of time - often needlessly!    

By the way I looked at putting the **From** and **To** the same and then 0 in the step and I am pleased that worked - at least in the old version!  

3) Most packages have aspects of the dialogues that show the programmers has thought more about their programming convenience than the user.  We have a tough set of users, namely relative beginners.  That's one reason we are trying to be really careful in the details of the look and feel of the dialogues - so they are really simple and easy to use.  And I am afraid that getting all these details right will take time and occasionally will strain working relationships in the team.   Let's enjoy working together as much as we can.

dannyparsons:
From just speaking to @Lunalo I believe the dialog has now been returned to how it was originally and I think that is fine for now so we can move on.

instat/dlgRegularSequence.designer.vb
lilyclements: @AlexSananka while this control is not used yet, it would be good to update it to a ucrChk. Then in the main code all that needs to be done is to set the name to "Define as Factor" and disable the ucrcheckbox.

instat/dlgRegularSequence.vb
```diff
@@ -280,13 +215,34 @@ Public Class dlgRegularSequence
...
+        CheckSequenceLength()
+    End Sub
+
+    Private Sub ucrNewColumnName_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrNewColumnName.ControlContentsChanged, ucrInputTo.ControlContentsChanged, ucrInputFrom.ControlContentsChanged, ucrInputInStepsOf.ControlContentsChanged, ucrNudRepeatValues.ControlContentsChanged, ucrPnlSequenceType.ControlContentsChanged, ucrSelectDataFrameRegularSequence.ControlContentsChanged
+        TestOKEnabled()
+    End Sub
+
+    Private Sub ucrInputFrom_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputInStepsOf.ControlContentsChanged, ucrPnlSequenceType.ControlContentsChanged
+        SetInStepsOfParameter()
+    End Sub
```
lilyclements: So, this comment refers to the subs in line 233-235, 237-239 and 241-243. The issue with the preview box is because you want to call "SetInStepsOfParameter" before "CheckSequenceLength".
You need to set the parameter as negative in order for the preview box to work, rather than the other way around which you accidentally had. To fix this, I would suggest merging some subs, so that all of the input.ControlContentsChanged are in one sub like so:

```
    Private Sub ucrNewColumnName_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrNewColumnName.ControlContentsChanged, ucrNudRepeatValues.ControlContentsChanged, ucrPnlSequenceType.ControlContentsChanged, ucrSelectDataFrameRegularSequence.ControlContentsChanged
        TestOKEnabled()
    End Sub

    Private Sub ucrInputFrom_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputFrom.ControlContentsChanged, ucrInputTo.ControlContentsChanged, ucrInputInStepsOf.ControlContentsChanged
        SetInStepsOfParameter()
        CheckSequenceLength()
        TestOKEnabled()
    End Sub
```

This does pose one problem still, however, that if the value is "1" in the ucrInputInStepsOf then the preview does not update

dannyparsons:
What's the state of this? It seemed like it was getting close to complete last week but I haven't seen any activity this week?

AlexSananka:
Yeah sure I haven't worked on it this week,I gave more attention dlgRatingScales.
This dialog needs small adjustments more on the preview window.Just small changes and this might be ready for Thusday's release.

dannyparsons:
That's fine to focus more on another dialog. But I also expect you should be able to work on multiple dialogs at the same time, especially if this only needs small changes and could be merged soon.

stevekogo:
@AlexSananka Can you make the adjustments so that this can be merged?

AlexSananka:
@stevekogo  I  have made the changes and I think this is ready for @dannyparsons

maxwellfundi:
@africanmathsinitiative/developers  this is urgent for today and needs urgent review. 

Thanks.

AlexSananka:
@maxwellfundi  maybe you can have a look at it just before @dannyparsons

lilyclements:
@africanmathsinitiative/developers I will review this in approximately 20 minutes. Hope that's okay!

lilyclements:
@dannyparsons By the way, before this gets merged (and after that problem I outlined above is fixed) I would like to tidy up the dialog and code a bit!

AlexSananka:
@lilyclements I am not getting this error in my branch,maybe we should skype and you can help me trace the issue

dannyparsons:
This dialog seems to be stable in the current version so I don't see this as urgent to be merged today if there are still issues here.

AlexSananka:
@lilyclements just showed me the issues and am trying to fix them,Thanks.

lilyclements:
To summarise the skype:
1. The repeat function should always run to be able to fit the data frame length
2. The assign to should always be on the repeat function
3. We should not run the `each` parameter both functions, only the `rep `function

AlexSananka:
@lilyclements I have made the changes,you can test this again,It should be working fine.

instat/dlgRegularSequence.vb
```diff
@@ -13,78 +13,128 @@
...
+
+        ucrInputInStepsOf.SetParameter(New RParameter("by", 2))
+        ucrInputInStepsOf.AddQuotesIfUnrecognised = False
+        ucrInputInStepsOf.SetValidationTypeAsNumeric(dcmMin:=0)
+
         ucrDataFrameLengthForRegularSequence.SetDataFrameSelector(ucrSelectDataFrameRegularSequence)
+
+        ucrPnlSequenceType.AddRadioButton(rdoNumeric)
+        ucrPnlSequenceType.AddRadioButton(rdoDates)
+        ucrPnlSequenceType.bAllowNonConditionValues = True
```
lilyclements: Do you understand what this line does and do you understand why you are getting the developer error?

instat/dlgRegularSequence.vb
```diff
@@ -13,78 +13,128 @@
...
-        nudInStepsOf.Value = 1
-        nudRepeatValues.Value = 1
-        SetNumericOrDatesParameters()
-        CheckSequenceLength()
-        nudNumberofDecimalPlaces.Value = 2
-        ucrNewColumnName.SetPrefix("Regular")
-    End Sub
+        ucrNewColumnName.Reset()
+
+        rdoNumeric.Checked = True
```
lilyclements: You said on the other version of this branch that you've fixed this. Can you send your updating code showing the fix?

instat/dlgRegularSequence.vb
AlexSananka: Yes I do,but I can't fix.Could you be in a position to explain it to me better?

AlexSananka:
@lilyclements and @dannyparsons  I think this dialog is ready.What are your thoughts?

instat/dlgRegularSequence.vb
lilyclements: Okay, so the developer error is occurring because you have that nothing on the ucrPnl is checked (until SetDefaults, where you manually checked it). You want it to be checked in the new system.

AlexSananka: This is now fixed

instat/dlgRegularSequence.vb
```diff
@@ -13,78 +13,125 @@
...
+        SetRCodeForControls(True)
+        TestOKEnabled()
     End Sub
 
     Private Sub TestOKEnabled()
-        If Not ucrNewColumnName.IsEmpty Then
+        If ucrNewColumnName.IsComplete Then
             If rdoNumeric.Checked Then
-                If nudFrom.Text <> "" AndAlso nudTo.Text <> "" AndAlso nudInStepsOf.Text <> "" AndAlso nudRepeatValues.Text <> "" AndAlso ucrDataFrameLengthForRegularSequence.ucrDataFrameSelector.cboAvailableDataFrames.Text <> "" Then
+                If ucrInputFrom.GetText <> "" AndAlso ucrInputTo.GetText <> "" AndAlso ucrInputInStepsOf.GetText <> "" AndAlso ucrNudRepeatValues.GetText <> "" AndAlso ucrSelectDataFrameRegularSequence.cboAvailableDataFrames.Text <> "" AndAlso txtGetPreview.Text <> "" Then
```
dannyparsons: I don't like the Test OK being based on the preview window - the way we do the preview could change but shouldn't effect the OK enabled. It should be based on the controls and the R code. 
What is done in the preview that we can't check on the controls?

AlexSananka: This was in an old version while we were still testing something.I have removed this.

PR #2760, opened 2017-03-09
Describe one var summarise problem  fixes #2753

### LINKED ISSUE #2753: Describe > One Variable > Summarise problem
Open the diamonds dataset from the ggplot library.

The command works OK with the default settings.  If you customise and just use the factors, there is a bug.  I think it isn't coping well with ordered factors.

I also tried with the survey dataset.  When I made the Fert column into an ordered factor then I got the same bug.

stevekogo:


stevekogo:
@rdstern This fixes the bug in #2753

stevekogo:

stevekogo:
@lilyclements Roger found a bug when summarizing ordered factors for describe one variable. Check the additions on the summaries method and try to understand, If not let me know.

maxwellfundi:

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -109,6 +110,9 @@ instat_object$set("public", "summary", function(data_name, columns_to_summarise,
     else if(col_data_type == "factor"){
       column_summaries = intersect(summaries, factor_collection)
     }
+    else if(col_data_type == paste0(c("ordered","factor"), collapse = ",")){
```
dannyparsons: We should be doing these checks better using the `is` functions instead of checking class directly, as we get to issues like these. This should be part of a bigger update to these summary methods which need to go through the new summary system. It's probably not a job for before the 20th but maybe for the version after that.

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -277,6 +282,10 @@ summary_sd <- function(x, na.rm = FALSE,...) {
   return(sd(x,na.rm = na.rm))
 }
 
+summary_var <- function(x, na.rm = FALSE,...) {
+  return((sd(x,na.rm = na.rm))**2)
```
dannyparsons: is there a reason not to use the `var` function?

stevekogo: I will use var function

PR #2763, opened 2017-03-09
DlgOneWayFrequencies

stevenndungu:
In Progress...
Issue: In R studio, the output of `sjt.frq`  by default is given in the output window  but this can be redirected to HTML  by changing  `use.viewer`  parameter from default as `TRUE`  to `FALSE`.
`sjt.frq(efc$e42dep,use.viewer=FALSE)` -HTML
`sjt.frq(efc$e42dep,use.viewer=TRUE)`- Rstudio output window
@dannyparsons @stevekogo 
Now coming to R-instat I am having a challenge redirecting the output to the R-Instat output window for the case of tables.  The tables (output) are redirected to  HTML which I guess is not what is required. I have tried using iCallType 2 and adding the parameter  `use.viewer` to control this, but apparently this does not help. 

For the graphs - its displaying well on the output window. 
@rdstern , Is the Weights check box supposed to pass `weight.by` parameter ? Just to be sure. otherwise not fully setup but almost

### DIFF
--- a/instat/dlgOneWayFrequencies.Designer.vb
+++ b/instat/dlgOneWayFrequencies.Designer.vb
@@ -22,13 +22,6 @@ Partial Class dlgOneWayFrequencies
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()> _
     Private Sub InitializeComponent()
-        Me.ucrSelectorOneWayFreq = New instat.ucrSelectorByDataFrameAddRemove()
-        Me.ucrReceiverOneWayFreq = New instat.ucrReceiverSingle()
-        Me.ucrSaveGraph = New instat.ucrSave()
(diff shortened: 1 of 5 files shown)


dannyparsons:
There is a parameter in RunScript for HTML output but this is not accessible from ucrBase yet.  We just need to add that as something which can be set there and passed through to RunScript. If you can do this great, otherwise I can work on it.

dannyparsons:
In future just create the blank dialog first so that can be merged in.

stevenndungu:
Yes, I created it yesterday and I needed to add some changes. So i opted to add more changes since I new you were busy in the conference presentations.

stevenndungu:
Let me try setting it up, in case i miss out on something I will consult you. Thanks

PR #2767, opened 2017-03-10
Describe two variable new implementation. Fixes #2762

### LINKED ISSUE #2762: Describe > Two Variables > Summary
I wonder what has happened to this dialogue?  Could someone please check and correct.  I think Steve was involved earlier.

Once the bugs are fixed I suggest we swap the receivers.  Make the first one a single receiver  "First Variable" and the second should be the multiple receiver.  But perhaps check what is happening first, before changing this.

I tried with the diamonds dataset.  With first receiver a set of factors and the second a variate.  I got nothing   I wonder if it is in the middle of being updated to the new system.  Or is it the problem again that these are ordered factors?

I also got nothing with them the other way round.

Then I tried with 3 variates and got (I think) correlations.  So that's OK.

Then I tried with just factors.  But it insists that nothing has changed and gives me the correlations again.

So I tried by opening the survey data set.  But it will not clear the idea it now has to just give those damn correlations from the diamonds dataset.  So I gave up!

This needs a careful lookup and then some testing.

Thanks

stevekogo:


stevekogo:
This is not ready for review.

stevekogo:
@africanmathsinitiative/developers This is almost ready but there is a bug on reset and changing between data frames. Someone should help me try figure out what causes this.

instat/dlgDescribeTwoVariable.vb
```diff
@@ -49,63 +64,107 @@ Public Class dlgDescribeTwoVariable
...
+        ucrReceiverSecondVar.SetParameterIsString()
+
+        ucrChkOmitMissing.SetText("Omit Missing Values")
+        ucrChkOmitMissing.SetParameter(New RParameter("na.rm"))
+        ucrChkOmitMissing.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
+        ucrChkOmitMissing.bUpdateRCodeFromControl = True
+
+        ucrChkSaveResult.SetText("Save Result")
+        ucrChkSaveResult.SetParameter(New RParameter("store_results"))
+        ucrChkSaveResult.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
```
lilyclements: I believe if you set the dataframe selector for the ucrChkSaveResult then the problem with changing dataframes should sort itself out

stevekogo: Thanks for the advice @lilyclements but I think setting dataframe selector is not a property of a ucrcheck.

lilyclements: Okay, I don't understand - why is this a ucrChk for a save and not a ucrSave?

stevekogo: ucrcheck was doing the job but I think ucrsave is the better option as it has set data frame property

lilyclements: You've raised a good point, however, that I don't know if we can do a ucrSave where only the checkbox is present?

instat/dlgDescribeTwoVariable.vb
```diff
@@ -49,63 +64,107 @@ Public Class dlgDescribeTwoVariable
...
+        ucrBaseDescribeTwoVar.iHelpTopicID = 414
         ucrBaseDescribeTwoVar.clsRsyntax.iCallType = 2
+
+        ucrSelectorDescribeTwoVar.SetParameter(New RParameter("data_name", 0))
+        ucrSelectorDescribeTwoVar.SetParameterIsString()
+
         ucrReceiverFirstVar.Selector = ucrSelectorDescribeTwoVar
-        ucrReceiverFirstVar.SetSingleTypeStatus(True)
-        ucrReceiverFirstVar.SetMeAsReceiver()
+        'ucrReceiverFirstVar.SetSingleTypeStatus(True)
```
lilyclements: As you pointed out, to remind you, this should be uncommented

stevekogo:
This will be closed and addressed in #2866.

PR #2777, opened 2017-03-15
Main form

lilyclements:

I've added Dry Spells and enabled it on the main form


rdstern:
If I understand correctly there is already a **Spells** item on the menu - under **Climatic > Prepare > Specific Events**.
I notice there is also both **End of the Rains** and **Water Balance** - so that's good.  No need to add anything more there either!

lilyclements:
Yes sorry I didn't clarify - I meant, there is a Spells item in that location already. All I did was enable it and connect it to the dlgDrySpells dialog. At the moment it is there and disabled so this pull request is just enabling it

lilyclements:
I wanted to check a few of these new features. So, on "Condition:" with the between/>/<, what does this do? Is this just the threshold so I want a threshold >0.85 (say) or <0.85 etc?
Currently, presumably, we have Rain >= 0.85 for the threshold?

Also, can I do >= and =< instead of > and <?

rdstern:
Great.  Thanks.

PR #2781, opened 2017-03-16
Water balance

lilyclements:
In Progress


lilyclements:
@africanmathsinitiative/developers This is ready for review. If you use this and change the date parameters, you have to change the water balance and capacity nuds afterwards. This is a temporary problem that will be resolved but cannot be in time for Monday when Roger wants to use this

lilyclements:
@Lunalo you reviewed this earlier. What were your thoughts, do I need to do anything else or is it ready for Danny?

PR #2829, opened 2017-03-21
Import odk

stevekogo:


dannyparsons:
If this is ready then it should be urgently tested as it's needed for papers and videos @africanmathsinitiative/developers

stevekogo:

Lunalo:
The dialog works well and I think it is in a good shape but I have a number of observations:
We have a developer error Message that appears on reset.
If I pass in a wrong password There is an error from r which seems complicated. We should capture it.
We should also capture errors that results from using wrong user names. 
What if when the form is Empty? Still there is an error since that is an empty data frame

dannyparsons:
These are all good points to work on. I suggest only the first one needs to be addressed now as the other three are improvements that need to be made but don't cause a serious problem at the moment.

stevekogo:
I have fixed the developer error message on reset.

instat/dlgImportFromODK.vb
```diff
@@ -16,26 +16,28 @@ Imports RDotNet
...
     End Sub
 
     Private Sub SetRCodeForControls(bReset As Boolean)
-        SetRCode(Me, ucrBase.clsRsyntax.clsBaseFunction, bReset)
-        TestOKEnabled()
+        ucrPnlPlatform.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
+        ucrInputUsername.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
+        ucrInputChooseForm.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
+        ucrPnlPlatform.SetRCode(clsGetFormsFunction, bReset)
+        ucrInputUsername.SetRCode(clsGetFormsFunction, bReset)
```
dannyparsons: Setting the R code for the same control twice shouldn't be done since they just overwrite each other. I suggest you just set the controls to the main function and you add the parameters to the internal function manually in the code.

instat/dlgImportFromODK.vb
```diff
@@ -44,56 +46,49 @@ Public Class dlgImportFromODK
...
     End Sub
 
-    Private Sub pnlPlatform_ControlContentsChanged() Handles ucrPnlPlatform.ControlValueChanged
-        If rdoOna.Checked Then
-            clsGetFormsFunction.AddParameter("platform", Chr(34) & "ona" & Chr(34))
-        ElseIf rdoKobo.Checked Then
-            clsGetFormsFunction.AddParameter("platform", Chr(34) & "kobo" & Chr(34))
-        End If
-    End Sub
+    'Private Sub pnlPlatform_ControlContentsChanged() Handles ucrPnlPlatform.ControlValueChanged
```
stevekogo: Then I will have to keep this.

dannyparsons: Yes. The way you have it now, it will work for the get forms function but won't work for the main function.

stevekogo:
@dannyparsons This is now fixed.

PR #2838, opened 2017-03-23
Dlg water balance

lilyclements:
Added column name option

PR #2840, opened 2017-03-23
Dlg spells

lilyclements:
@rdstern I have added in column names to this dialog. The following shows the
Year, Default-Name and then some names I added in to get multiple columns on the sheet.

Is this what you had in mind?
Similarly this has been implemented on #2838 

image

### DIFF
--- a/instat/dlgSpells.Designer.vb
+++ b/instat/dlgSpells.Designer.vb
@@ -253,7 +253,7 @@ Partial Class dlgSpells
         '
         Me.ucrSaveSpells.Location = New System.Drawing.Point(9, 295)
         Me.ucrSaveSpells.Name = "ucrSaveSpells"
-        Me.ucrSaveSpells.Size = New System.Drawing.Size(231, 24)
+        Me.ucrSaveSpells.Size = New System.Drawing.Size(290, 24)
         Me.ucrSaveSpells.TabIndex = 14
         '
         'dlgSpells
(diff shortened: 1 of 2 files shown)


lilyclements:
I'm having issues with the prefix of names and might want to discuss this with @dannyparsons - this dialog brings up a new dataframe, and so if I set a prefix of "Spells" it will always go to "Spells1" since the column Spells1 goes in a different dataframe to the dataframe in the selector.

rdstern:
That's great.  These dialogues will usually be used multiple times to try different options.  You don't have to be very clever in the names, as long as the user can choose a name for themselves if they wish.  That also gives them the option to overwrite if they want to.  Sometimes you do, because you realise the settings were wrong before.  For example yesterday we forget, and had the earliest start date still as day 1.  Then we corrected it.  That's when we would (ideally) overwrite the column.

rdstern:
Ok.  This could perhaps also be a David question?

lilyclements:
I've spoken to Danny about it (will catch David later when he comes back up!) who says he doesn't think we've come across this problem before and as a result we probably won't be able to fix it today. We at least have the column names for now so you can have multiple columns from the same dialog

lilyclements:
@Lunalo Can you re-review this for me please? Then we can get it merged asap for the new version this evening

maxwellfundi:

PR #2844, opened 2017-03-23
Setting up receiver on standardise country names

maxwellfundi:
@dannyparsons  was going through these corruption dialogs and this was not set up properly


dannyparsons:
Thanks for noticing, I have also fixed this in my branch that I will merge soon. Sorry I didn't tell you beforehand.

PR #2849, opened 2017-03-23
DlgConvertColumns- Bug Fixed

stevenndungu:
The dialog on checking the specify number of decimals checkbox, still would allow variables of character and factor which should not be the case. This would create a bug, but I have fixed it. @africanmathsinitiative/developers someone can  review


dannyparsons:
@lilyclements Is this stable and bug free now?

lilyclements:

PR #2866, opened 2017-03-28
Describe two var new implementation fixes #2762

### LINKED ISSUE #2762: Describe > Two Variables > Summary

stevekogo:


stevekogo:
This solves the issues in #2762 and #2767. I will tidy up then open it up for review.

stevekogo:
@africanmathsinitiative/developers This is now in shape. I welcome your reviews.

instat/sdgDescribeDisplay.Designer.vb
```diff
@@ -20,138 +20,117 @@ Partial Class sdgDescribeDisplay
...
+        Me.ucrChkProportions.Size = New System.Drawing.Size(100, 20)
+        Me.ucrChkProportions.TabIndex = 4
+        '
+        'ucrChkPercentages
+        '
+        Me.ucrChkPercentages.Checked = False
+        Me.ucrChkPercentages.Location = New System.Drawing.Point(127, 40)
+        Me.ucrChkPercentages.Name = "ucrChkPercentages"
+        Me.ucrChkPercentages.Size = New System.Drawing.Size(97, 20)
+        Me.ucrChkPercentages.TabIndex = 5
```
lilyclements: The tab order needs to be addressed

instat/sdgDescribeDisplay.Designer.vb
```diff
@@ -20,138 +20,117 @@ Partial Class sdgDescribeDisplay
...
-        Me.chkMeans.Name = "chkMeans"
-        Me.chkMeans.Size = New System.Drawing.Size(58, 17)
-        Me.chkMeans.TabIndex = 2
-        Me.chkMeans.Tag = "Means"
-        Me.chkMeans.Text = "Means"
-        Me.chkMeans.UseVisualStyleBackColor = True
+        Me.ucrChkSignifStars.Checked = False
+        Me.ucrChkSignifStars.Location = New System.Drawing.Point(10, 19)
+        Me.ucrChkSignifStars.Name = "ucrChkSignifStars"
+        Me.ucrChkSignifStars.Size = New System.Drawing.Size(100, 20)
```
lilyclements: The size of these should be increased. On both significant checkboxes, only the word "Visible" is noticeable. Check if this applies to other checkboxes too.

instat/dlgDescribeTwoVariable.vb
```diff
@@ -49,174 +61,179 @@ Public Class dlgDescribeTwoVariable
...
         End If
-        If (Not ucrReceiverSecondVar.IsEmpty()) Then
-            strSecondVarType = ucrReceiverSecondVar.strCurrDataType
+        If (ucrReceiverSecondVar.GetCurrentItemTypes.Count > 0) Then
+            strSecondVarType = ucrReceiverSecondVar.GetCurrentItemTypes.Item(0)
         Else
             strSecondVarType = ""
         End If
 
         If ((strVarType = "numeric" OrElse strVarType = "integer") And (strSecondVarType = "numeric" OrElse strSecondVarType = "integer")) Then
```
lilyclements: How come you have used And and not AndAlso?

instat/sdgDescribeDisplay.vb
```diff
@@ -15,112 +15,64 @@
...
 
-    Private Sub chkSignifStars_CheckedChanged(sender As Object, e As EventArgs) Handles chkSignifStars.CheckedChanged
-        If chkSignifStars.Checked Then
-            clsRAnovaDispOptions.AddParameter("signif.stars", "TRUE")
-        Else
-            clsRAnovaDispOptions.RemoveParameterByName("sign.stars")
-        End If
+        ucrChkMeans.SetText("Means")
+        ucrChkMeans.SetParameter(New RParameter("means"), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:="TRUE", strNewValueIfUnchecked:="FALSE")
+        ucrChkMeans.SetRDefault("FALSE")
```
lilyclements: After this you need, otherwise if you check any of the checkboxes, press OK on the main dialog, and then reopen the dialog, uncheck the checkboxes - it still runs the code for them as TRUE.

       ```
 lstCheckboxes = New List(Of ucrCheck)
        lstCheckboxes.AddRange({ucrChkAddMargins, ucrChkMeans, ucrChkPercentages, ucrChkProportions, ucrChkSignifLevel, ucrChkSignifStars, ucrChkTranspose})
        For Each ctrTemp As ucrCheck In lstCheckboxes
            ctrTemp.clsParameter.SetArgumentValue(Chr(34) & ctrTemp.clsParameter.strArgumentName & Chr(34))
            ctrTemp.clsParameter.bIncludeArgumentName = False
        Next
        bControlsInitialised = True
```

lilyclements:
@stevekogo what is the state of this? I think we would want this in tomorrow's release

stevekogo:
@lilyclements The main problem was receiver changing from factors to numeric columns. With Danny's update on running multiple functions, I will have a look at it then get back to you with more details.

stevekogo:
@lilyclements Danny's update fixed the problem or multiple functions. You can go ahead and test it once more as I deal with the issue on describe display subdialog.

instat/sdgDescribeDisplay.vb
stevekogo: I dont understand what you are suggesting. Note that that the checkboxes behave in a different way compared to those of sdgsummaries

lilyclements:
Tried this on the most recent branch.
This bug is still present, e.g. if I put a factor in the both receivers, and then remove it from the multiple and put a numeric in there is an issue.
I think @dannyparsons update has fixed the issue but that this fix means that the code for the receivers needs to be implemented in this dialog differently (that statement may not be correct, @dannyparsons can you confirm this for me?)

stevekogo:
@lilyclements I now get it. Let me try to figure out what the problem is. Your suggestions on potential fix are welcome.

instat/sdgDescribeDisplay.vb
lilyclements: Okay, I realised this after I commented it and forgot to remove the comment.
I thought that the dialog wouldn't run if the user didn't check everything in the sub dialog. However, that is not the case, apologies.

dannyparsons:
@stevekogo my example of one way frequencies from here shows how you can now implement multiple functions/parameters for controls in a dialog https://github.com/africanmathsinitiative/Instat/pull/2909 and particularly the review I did explaining some parts.

Doing this should simplify the dialog as well as preventing those kinds of bugs that I think you are getting.

stevekogo:
@dannyparsons Thanks for the advice. I am on it.

stevekogo:
@dannyparsons I tried fixing this using the new system of managing multiple functions without success. The biggest challenge is that the choice of the function depends on the datatypes of the variables in the receivers and notany specific control. In addition, the problem seems to occur when you shift from a function that requires the variables in the receivers to be strings to a function that requires the variables as R-functions. 

You can have a look at this dialog then call me any time for a discussion.

stevekogo:
@dannyparsons I am still waiting for your feedback on this. Have a look and advice accordingly.

dannyparsons:
Yeah we can have a look at this

stevekogo:
@dannyparsons I this this will have to wait to be implemented in the next version.

dannyparsons:
Shall I have a quick look at this now?

stevekogo:
ok

dannyparsons:
Changing this now is quite a big task. Unfortunately, it will have to wait for next time.

Is this an improvement on the current version and should be merged now?

lilyclements:
I think this should be merged now. Despite the error with the receivers, it runs a lot better than the current version

stevekogo:
It is an improvement. The only problem is a small bug which doesn't affect the functonality.

PR #2869, opened 2017-03-29
dlgRugPlot

maxwellfundi:
In progress!


maxwellfundi:
@dannyparsons  I tried implementing what we discussed yesterday and looks like we are heading in the right path, but, even after updating the parameter with the current value of the current receiver, it passes the value twice. I have showed this @stevekogo  and we think you could have a look now at this.

dannyparsons:
Ok I'll have a look through this and see.

maxwellfundi:
Great. Lots discussions needed for this though after you have a look.

dannyparsons:
looking at this now

lilyclements:
@maxwellfundi I think this branch can be deleted and closed? My understanding is that this is sorted in Danny's branch he merged yesterday

PR #2871, opened 2017-03-29
fixed bug in levels labels dialog

### LINKED ISSUE #2861: dlgLabels - R-Error
I have moved the issues of the labels bug to here. @africanmathsinitiative/developers let me know your thoughts on the following -

If the user opens Levels/Labels to explore their data and presses OK without adding a new level, an R error comes up saying that new_levels is not specified.
If I then add a level and press OK then it runs fine, as expected.
If I then re-open the dialog and do not add another level then it runs without an error and just runs  `new_levels = c(*all levels already in the receiver*)` 

In the first opening of the dialog, I have not specified any new levels and it does not run
In the third opening of the dialog, I have not specified any new levels and it **does** run

I think we need some consistency, either to not be able to press OK if I do not add any levels
Or, to run `new_levels = c("....")` every time even if I do not add a new level

dannyparsons:
fixes #2861


dannyparsons:
@lilyclements can you check this fixes the bug with the dialog?

PR #2876, opened 2017-03-30
added user controls for all mdi child forms and implemented new window system in frmMain using split containers

dannyparsons:
@africanmathsinitiative/instat Here is a suggested change to the frmMain layout - partly prompted by the difficultly we saw students had with the mdi child form system. This uses split containers to keep sections of the form separate while still allowing flexibility in resizing.

Please make a copy of this branch and play around and test. Let me know what you think, whether it is an improvement and any suggestions.

### DIFF
--- a/instat/clsGridLink.vb
+++ b/instat/clsGridLink.vb
@@ -17,6 +17,7 @@ Imports RDotNet
 Imports unvell.ReoGrid
 
 Public Class clsGridLink
+    Public ucrDataViewer As ucrDataView
     Public grdData As ReoGridControl
     Public grdMetadata As ReoGridControl
     Public grdVariablesMetadata As ReoGridControl
@@ -104,7 +105,7 @@ Public Class clsGridLink
                     Else
                         FillSheet(dfTemp, strDataName, grdData, bInstatObjectDataFrame:=True, bIncludeDataTypes:=True, iNewPosition:=i, bFilterApplied:=False, bCheckFreezeColumns:=True)
                     End If
-                    frmEditor.SetColumnNames(strDataName, dfTemp.ColumnNames())
+                    ucrDataViewer.SetColumnNames(strDataName, dfTemp.ColumnNames())
                     clsSetDataFramesChanged.AddParameter("data_name", Chr(34) & strDataName & Chr(34))
                     clsSetDataFramesChanged.AddParameter("new_val", "FALSE")
                     frmMain.clsRLink.RunInternalScript(clsSetDataFramesChanged.ToScript())
@@ -181,10 +182,7 @@ Public Class clsGridLink
             grdVariablesMetadata.Visible = True
             grdMetadata.Visible = True
         End If
-        'TODO TEMPORARY THIS MUST BE REMOVED
-        'Cannot refer to frmEditor directly
-        'Could fix by having user control for grid and here raising event handled in frmEditor
-        frmEditor.UpdateCurrentWorksheet()
+        ucrDataViewer.UpdateCurrentWorksheet()
     End Sub
 
     Public Sub SetMetadata(tmpStrMetadata As String)
@@ -209,8 +207,9 @@ Public Class clsGridLink
         End If
     End Sub
 
-    Public Sub SetData(grdTemp As ReoGridControl)
-        grdData = grdTemp
+    Public Sub SetDataViewer(ucrNewDataViewer As ucrDataView)
+        ucrDataViewer = ucrNewDataViewer
+        grdData = ucrNewDataViewer.grdData
         bGrdDataExists = True
         bGrdDataChanged = True
         UpdateGrids()
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -15,6 +15,7 @@
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 Imports RDotNet
+Imports unvell.ReoGrid
 
 Public Class RLink
     ' R interface class. Each instance of the class has its own REngine instance
@@ -44,6 +45,8 @@ Public Class RLink
 
     Public strGraphDisplayOption As String = "view_output_window"
 
+    Private grdDataView As ReoGridControl
+
     Public Sub New(Optional bWithInstatObj As Boolean = False, Optional bWithClimsoft As Boolean = False)
 
     End Sub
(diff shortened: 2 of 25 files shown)


lilyclements:
I really like this, I would very much advise everyone else to have a play around and look at this.
I have a few comments on why I really like this, mainly it is because it removes so many problems we have been having:

- It's easier to rearrange windows. In the current system, often windows have been getting layered upon one another and lost.

- Students were pressing "X" thinking it would delete their sheet when actually it closes the window. Since there is no "X" then this eliminates that issue

- The output window and data window are harder to remove now which I think is good. Someone shouldn't want to be able to easily remove those items from their main screen!

- They're impossible to minimise/maximise now. On Saturday, one student couldn't find their data view window. They had minimised their data view window and then it was hidden in the bottom left corner. I only knew it would be there as I've seen this problem before. The student then maximised their Data View Window thinking it would come up and cover the half of the screen it was covering before. It instead covered the entire screen.

- Students have tried to remove the output window since they want to see more of their data and less of the code, and then they've wondered where their graphs are going. This helps ensure that the output window is always there, and also means they can easily increase the side of their data

- We have had the problem where students have looked at their metadata and got it confused with their overall data since their main data is hidden somewhere. This helps remove that since we can have them easily next to each other

I really like it and feel it solves quite a lot of confusion I've seen when others have been trying out the software.

lilyclements:
Downside: I'm not keen on that blue colour between windows, sorry.

lilyclements:
Downside: I'm not keen on that blue colour between windows. Also is it possible for the spacing between all the windows to be the same?

lilyclements:
Issue: If I insert two data frames and then select for Column Metadata, I only see the metadata for one of the data frames

lilyclements:
Issue: If I insert two data frames and then select for Column Metadata, I only see the metadata for one of the frames because the scrolly bar along the bottom of that window covers the tabs for the data frames (or at least it is on my laptop!)

dannyparsons:
Fixed the spacing and tab appearance for column metadata control. Any colour suggestion? Or no colour?
I don't like the blue flashing when you show/hide windows but haven't found a way around that yet.

Also, if we go with this layout we need to rethink the View menu since that was set up for quite a different layout.

stevekogo:
@dannyparsons The new look looks awesome. Thanks for the implementation.
Here are a few suggestions:

- There should be provision to rearrange the windows as you wish and probably specify the separation color.(Separation color may go to options while the rearrange windows could be somewhere in the View Menu or enable rearranging within the main window)

- I would prefer to have titles on the windows. The minimize and maximize icons may not be necessary in this case.

dannyparsons:
Thanks Steve. Keeping the titles is a good idea, I will investigate that. 
The separator colour can easily be an option.
Rearranging I think is possible but I haven't figured it out yet. One question on that is how much flexibility we allow. At the moment we have two "main" windows and four "minor" windows and you can't change which ones are in the top and bottom half. For example, would we allow the log window in place of the output window or the data frame metadata where the data grid currently is? Should we allow all 6 windows side by side? etc. Don't think we have to decide now but something to think about.

maxwellfundi:
@dannyparsons  i really like the new layout. It looks really good. 
i would like it more if the metadata, script and log window to be at the bottom. it looks much better that way. 
i would also like to have titles on the windows as we currently have. Further a close button on the windows would be useful so that we can click it and the window disappears rather than going back to view, and clicking the window menu. It would save the user a click.

lilyclements:
I am not keen on a close button on all the windows. I think a close button on the windows that go along the top of the screen sounds reasonable, however, a close button on the Data View or Output View I'm not so sure. This is because we have had an issue up to now of students closing the Data View because they think it deletes their sheet (or closing the Output Window for one reason or another). Then they are not sure how to get it back. I like the idea that to close the Data View/Output View, you have to go through the same routine as to open it. Otherwise they can close it and not know how to get it back. However, I don't mind on closing on the other four windows.

Titles on the windows sounds like a good idea

I don't see the need for a minimise or maximise since the user can reshape them to the size they want. I don't know how a minimise or maximise would work on this format.

dannyparsons:
Interesting that you prefer the extra windows at the bottom - I like them at the top. I can check if it's easy to do. I think this is one disadvantage to this method, we may not always be able to position the windows exactly as we want them. The longer term solution is to have a proper docking system like in VS so I'm not so concerned if we can't do everything we want yet.

I agree with Lily about the issues of a close button. I like the idea of hiding and showing being done in the same way so you can't "lose" your windows. But I agree the View menu is not user friendly. We were thinking of having icons on the menu bar for each window so that you can quickly hide/show windows. As well as an option to reset to the default view. There could also be an item to hide the whole top bar so that you don't have to hide/show the windows one by one.

dannyparsons:
I've now added titles for each window. Apologies for the colours. Someone with design ideas should help us with that.

I think this can be merged now since no one has reported any issues this branch?

rdstern:
But of course that gave me the chance to use it!

I found I could open the survey data set and then also Dodoma and it all looks great to me.  This is an excellent improvement.

But, before that I had tried opening the diamonds data set from the library and it crashed.  I wondered if that was because it was large, so next ime I looked at one of the default data sets, and pressed on the special help that puts the information in a browser.  That was supposed to be a preliminary, but that makes R-Instat very unhappy too.  It doesn't give an error message - and (of course) is very easy for you to try!

Now I'll try some analyses!

dannyparsons:
Thanks I'm looking into this

rdstern:
Are there options for the behaviour of the "horizontal scrolling bar" - if that is the right name.  It is in the data window, where it is currently quite large and it is impossibly large in the column metadata window.  It also doesn't change proportionally when you resize the window.  If its behaviour on resizing is fixed, then I would still start with it shorter on the data window and much shorter on the column metadata window.  Currently, on the metadata there is only room for a single data frame label to be visible.

In thinking of options we would like, in the new (I assume) Tools > Options, then one I would like is the initial height of the extra windows section.  My choice would be to have it a bit bigger than the currwent value, but I am sure others would want it to be quite short.

Before seeing the "system" I would have been quite happy with just 3 windows - and the "extra" 4 taking it in turn.  But I now really like the system you have, where you can see all 4 if you wish.  It is going to be good for teaching too!

lilyclements:
On the visual side of things, as a minor comment - I'm not keen on this green. I think since the icon for the programme is blue, then it should be a similar blue to that?

dannyparsons:
Good point - but also what will our icon be? And what colours will it have? We still don't have a new one.

lilyclements:
Okay, that's a fair point. I imagine since the R icon is blue with some grey, and the Instat icon is blue with some white, then the R-Instat icon will be mainly blue. However, I understand this is neither the time or place to discuss this! We can keep this current military green for the time being!

PR #2877, opened 2017-03-31
Dlg D

lilyclements:


lilyclements:
@africanmathsinitiative/developers this pull request is fixing a bug so needs to go into tomorrow's version. This is still waiting for review!

stevenndungu:
@lilyclements I am Still having the error popping up whenever I have median and N None missing values check boxes unchecked with a categorical variable inside the the receiver.
image

stevenndungu:
Two more things noted:
- Only few parameters are positioned, other parameters are not positioned especially those in the sub dialogue
- Should we set the base function after setting its respective parameters or vice versa? check line 32 of the main dialogue, not a big concern though..

lilyclements:
@stevenndungu I wrote an issue on the bug you noted ( #2878 ) and Steve said he would look into it. Its an issue in the summaries method not the dialog so can be sorted separate to this. Thanks!

By positioned do you mean in terms of the layout? They are positioned to me. Have you tried this on your most recent master branch? 

I don't understand your other bullet point, sorry. Do you mean line 32 of dlgDescribeOneVariable? To me, this is in the load part?

lilyclements:
Ah - I see what you mean about the positioning! Apologies. I'll fix that now. Thank you!

stevenndungu:
Sorry, before reviewing it I checked on the bug reported. So I thought you hhad fixed that bug, that's why on the comment I commented that it was still giving the same error. Thanks for clarifying that.

stevenndungu:
@dannyparsons  this is ready for you to review

stevekogo:
@lilyclements I have gone through this and I have two issues to point out.
First, there seems to be a bug when selecting columns. After loading the survey data frame, I selected field and size column to the receiver while the customize checkbox in unchecked. I then checked the customize checkbox and added Yield column to the receiver. Unchecking customize check box again hides the yield column and unhides upon checking......try looking into this.
Secondly, I would like to comment on the bug found by @stevenndungu which is something I was looking into but yet to fix. I noticed that the bug can be fixed in two places, either at the subdialog level or at the r-code level. At the subdialog, we will need additional checks to disable OK until all all the columns selected have atleast a summary that can be performed on them. The later case needs adjusting the summary function but this might be additional work for nothing since we are using the old summary system and we need to migrate this to the new summary system. @dannyparsons can comment further on this. However, I know the exact place in the old summary system which causes this bug.

lilyclements:
Thanks for this. I'll look into the first bug you described when I get into the office.

As for the second bug, I'm starting to change the summaries calculation methods to the new system so you're right. We don't need to worry about that bug for now. That's helpful you know where it is, however. I'm a little intrigued by it, is it possible to have a Skype later where you can explain it to me? I'm trying to understand this R method more since this is the first week I've worked on it and so that explanation might help

dannyparsons:
Thanks for the useful reviews. The bug with the receiver I believe is caused by having multiple functions in the dialog. I am working on adding features in the Core to make multiple functions easier to manage. I hope this will solve that issue.

On the R code issue, Lily is changing those summary methods to now use the new system so it's good to know the bug but probably not worth fixing now it's going to change. @lilyclements if you need to understand the old summary system you could ask @stevekogo since he implemented much of those methods.

stevekogo:
@lilyclements I will be happy to take you through the old summary system. We can have a skype to discuss this.

PR #2883, opened 2017-03-31
Lunalo stack new

dannyparsons:


instat/dlgStack.vb
```diff
@@ -16,115 +16,138 @@
...
+
+        ucrColumnsToCarryReceiver.Selector = ucrSelectorStack
         ucrColumnsToCarryReceiver.SetParameter(New RParameter("id.vars", 2))
         ucrColumnsToCarryReceiver.SetParameterIsString()
+        ucrColumnsToCarryReceiver.bAddRemoveParameter = False
+        ucrColumnsToCarryReceiver.SetValuesToIgnore({"NULL"})
+
+        ucrChkCarryColumns.SetText("Carry Columns")
+        ucrChkCarryColumns.AddParameterValuesCondition(True, "id.vars", "NULL", False)
+        ucrChkCarryColumns.AddParameterValuesCondition(False, "id.vars", "NULL")
```
dannyparsons: These conditions were needed to set the check box correctly.

instat/dlgStack.vb
```diff
@@ -16,115 +16,138 @@
...
+        ' I would like to do this since the checkbox also controls the value of id.vars
+        ' but there is confusion with both controls setting the value of the same parameter at the same time. For now the writing is done manually in this dialog.
+        'ucrChkCarryColumns.SetParameter(ucrColumnsToCarryReceiver.GetParameter(), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:="", strNewValueIfUnchecked:="NULL")
```
dannyparsons: I wanted the checkbox to have a parameter as well, but it would be too confusing in this case to have the same parameter as the receiver.

instat/dlgStack.vb
```diff
@@ -16,115 +16,138 @@
...
-        'ucrStackDataInto.SetValidationTypeAsRVariable()
-        'ucrFactorInto.SetValidationTypeAsRVariable()
         ucrStackDataInto.SetValidationTypeAsRVariable()
-        clsMelt.AddParameter("variable.name", Chr(34) & ucrFactorInto.GetText() & Chr(34))
-        clsMelt.SetRCommand("melt")
+
+        ucrSaveNewDataName.SetIsTextBox()
+        ucrSaveNewDataName.SetLabelText("New Data Frame Name:")
+        ucrSaveNewDataName.SetSaveTypeAsDataFrame()
+        ucrSaveNewDataName.SetDataFrameSelector(ucrSelectorStack.ucrAvailableDataFrames)
```
dannyparsons: This control was did not have a complete setup.

instat/dlgStack.vb
```diff
@@ -16,115 +16,138 @@
...
-    Private Sub AddSuffix()
-        If (Not ucrNewDataName.bUserTyped) AndAlso ucrSelectorStack.ucrAvailableDataFrames.cboAvailableDataFrames.Text <> "" Then
-            ucrNewDataName.SetName(ucrSelectorStack.ucrAvailableDataFrames.cboAvailableDataFrames.Text & "_stacked")
-        End If
-    End Sub
+        clsMelt.SetRCommand("melt")
+        clsMelt.AddParameter("variable.name", Chr(34) & "variable" & Chr(34))
+        clsMelt.AddParameter("value.name", Chr(34) & "value" & Chr(34))
+        clsMelt.AddParameter("id.vars", "NULL")
+        clsMelt.SetAssignTo(ucrSelectorStack.ucrAvailableDataFrames.cboAvailableDataFrames.Text & "_stacked", strTempDataframe:=ucrSelectorStack.ucrAvailableDataFrames.cboAvailableDataFrames.Text & "_stacked")
```
dannyparsons: In here you should only be setting up the function and parameters. The controls should be able to read their value from the function if the Initialise has been done properly.
If you are setting values of controls in here you are probably doing something wrong.

instat/dlgStack.vb
```diff
@@ -16,115 +16,138 @@
...
+    Private Sub SetCarryColumnsOptions()
+        If ucrChkCarryColumns.Checked Then
+            ucrColumnsToCarryReceiver.SetMeAsReceiver()
+            If Not ucrColumnsToCarryReceiver.IsEmpty() Then
+                clsMelt.AddParameter("id.vars", ucrColumnsToCarryReceiver.GetVariableNames())
+            End If
         Else
-            ucrBase.clsRsyntax.RemoveAssignTo()
+            ucrReceiverColumnsToBeStack.SetMeAsReceiver()
+            clsMelt.AddParameter("id.vars", "NULL")
```
dannyparsons: Because the checkbox does not have the id.vars parameter this needs to be done manually. If this is not done then the receiver may remove the parameter when it is empty.

instat/dlgStack.vb
```diff
@@ -16,115 +16,138 @@
...
         End If
     End Sub
+
+    Private Sub ucrSelectorStack_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSelectorStack.ControlValueChanged
+        SetDataFrameName()
+        SetCarryColumnsOptions()
```
dannyparsons: This also needs to be called here before the receiver will remove id.vars when it is cleared.

instat/ucrCore.vb
```diff
@@ -85,6 +85,11 @@ Public Class ucrCore
 
     Public bIsVisible As Boolean = True
 
+    ' Values which the parameter associated to the control may have but which shouldn't be used to set the control's value
+    ' Individual controls can determine what value to set when a parameter value is contained in strValuesToIgnore
+    ' (Currently only implemented for receivers)
+    Protected strValuesToIgnore As String()
```
dannyparsons: I have added this array which we use in the Stack dialog so that the receiver can ignore a value of NULL and clear instead of filling with NULL..

instat/dlgStack.vb
```diff
@@ -16,115 +16,138 @@
...
+
+        ucrReceiverColumnsToBeStack.Selector = ucrSelectorStack
         ucrReceiverColumnsToBeStack.SetParameter(New RParameter("measure.vars", 1))
         ucrReceiverColumnsToBeStack.SetParameterIsString()
+
+        ucrColumnsToCarryReceiver.Selector = ucrSelectorStack
         ucrColumnsToCarryReceiver.SetParameter(New RParameter("id.vars", 2))
         ucrColumnsToCarryReceiver.SetParameterIsString()
+        ucrColumnsToCarryReceiver.bAddRemoveParameter = False
+        ucrColumnsToCarryReceiver.SetValuesToIgnore({"NULL"})
```
dannyparsons: In the core I have implemented a list of values to ignore. So if the control's parameter is one of these values, it will not set it to that value and can do something else. This is only implemented for the receiver currently.
So if id.vars = NULL the receiver will ignore the value and instead clear.

dannyparsons:
@africanmathsinitiative/developers This needs careful testing before merging since there are many places for potential bugs.
A known issue is if the second receiver is the current receiver when you run the code, when you reopen the first receiver is now the current receiver. This is caused by the base buttons disabling the controls while running code and is not easily fixable until we implement a better method of running R code which will not freeze the software while running.

Lunalo:
@dannyparsons 

I cannot load the data from library . I think you should look at how you are setting the range styles in `  Public Sub FormatDataView(shtCurrent As Worksheet)`

Lunalo:
I will be set to test this when this problem is fixed

rdstern:
I think the latest version fixes the problem.  Danny found the bug and I can now load from the R library again.

dannyparsons:
Yes this is fixed in the latest version

Lunalo:
Thanks.

I have tested this and is functioning well.

@lilyclements  @stevekogo 
can do a final test.

Lunalo:
@lilyclements  

I have now fixed the tab order in the other branch

dannyparsons:
@Lunalo should this be closed then?

Lunalo:
This is fine

PR #2874, opened 2017-03-30
Dlg transpose text

lilyclements:
@africanmathsinitiative/developers this is ready for review

Changes:
- Updated the SetDefaults()
- Fixed developer error for a disabled control
- Sorted layout

PR #2860, opened 2017-03-27
dlgDuplicateColumns

lilyclements:
@africanmathsinitiative/developers This is ready for review

Changes made:
- Bug fix. It was reading in the name of the column not the contents
- Updated some of the code in the SetDefaults sub
- Bug fix. ucrInputColumnName only has a name once something has been inputted in the receiver, so there was a developer error upon opening the dialog when nothing was in the receiver.

PR #2632, opened 2017-02-01
Dlg unstack

lilyclements:
Not finished. Reached a problem with operators so will continue later.

PR #2884, opened 2017-03-31
dlgAppend

lilyclements:
In Progress
EDIT: @africanmathsinitiative/developers This is ready for review
Changes made - 

* Tab order
* Layout
* Updated SetDefaults() sub
* Fixed minor issues

### DIFF
--- a/instat/dlgAppend.vb
+++ b/instat/dlgAppend.vb
@@ -16,7 +16,7 @@
 
 Imports instat.Translations
 Public Class dlgAppend
-    Public bFirstLoad As Boolean = True
+    Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
     Private clsBindRows As New RFunction
 
@@ -99,11 +99,5 @@ Public Class dlgAppend
 
     Private Sub ucrReceiverAppendDataframe_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverAppendDataframe.ControlContentsChanged, ucrSaveGraph.ControlContentsChanged, ucrChkIncludeIDColumn.ControlContentsChanged, ucrInputIDColName.ControlContentsChanged
         TestOKEnabled()
-    End Sub
-
-    Private Sub ucrInputIDColName_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputIDColName.ControlValueChanged
-        'If ucrInputIDColName.GetText = ucrSelectorDataframes.lstAvailableVariable.Then Then
-        '    MsgBox("")
-        'End If
     End Sub
 End Class
--- a/instat/dlgAppend.vb
+++ b/instat/dlgAppend.vb
@@ -77,7 +77,7 @@ Public Class dlgAppend
     End Sub
 
     Private Sub TestOKEnabled()
-        If Not ucrReceiverAppendDataframe.IsEmpty AndAlso ucrSaveGraph.IsComplete() Then
+        If ucrReceiverAppendDataframe.lstSelectedVariables.Items.Count > 1 AndAlso ucrSaveGraph.IsComplete() Then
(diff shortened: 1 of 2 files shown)


instat/dlgAppend.vb
```diff
@@ -13,7 +13,7 @@
 '
 ' You should have received a copy of the GNU General Public License k
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
-Imports instat
+
 Imports instat.Translations
 Public Class dlgAppend
     Public bFirstLoad As Boolean = True
```
Lunalo: I think this should be private  just for consistency

lilyclements: Thanks this isn't ready for review yet. However, #2874 #2877 and #2860 (and potentially #2632) should be in for Friday's release.

Lunalo: okay great. Just remind me when it is ready .

Thank you

lilyclements: Thanks for this! I just changed it. This is ready for review now!

lilyclements:

maxwellfundi:
@lilyclements  I have been reviewing this dialog and 

1. The selector looks strangely huge, i think it needs to be same size as others because:-
     a. We need to keep to consistence
     b. With its size , it makes the dialog not appealing

2. With the idea of **appending dataframes** as the dialog and the menu says, I  expected okay to be enabled when at least two dataframes are added to the receiver. Apparently, this is not the case. was there a reason for this?

lilyclements:
1. I agree, I'm not keen on the size of it either. It isn't as the usual default size because it doesn't have the data frame selector, but perhaps I should reduce it so it's a similar size to that! I'll look into that now.

2. is an interesting point, currently OK is enabled when one data frame is in the receiver. I'm not sure what works better. Either -
* OK should be enabled when there is only one data frame in the selector because it works in R so why not work in R-Instat?
* OK should not be enabled when there is only one data frame in the selector because as you say, the dialog is about appending data frames not copying one. Perhaps it would make it clearer to the user what "appending" means and that that requires more than one data frame.

@africanmathsinitiative/developers what are your thoughts on 2.?

dannyparsons:
Yes I think minimum two data frames for OK to be enabled is sensible.

lilyclements:
Great thanks @africanmathsinitiative/developers this is ready for re-review.

lilyclements:
@maxwellfundi I made the necessary changes. Can you re-review this?

stevekogo:
@Lunalo @maxwellfundi You reviewed this sometime back. Can you check if your suggestions were addressed?

maxwellfundi:
@dannyparsons  this looks good. Please have a look.

instat/dlgAppend.vb
```diff
@@ -57,32 +57,27 @@ Public Class dlgAppend
...
 
         ucrSelectorDataframes.Reset()
         ucrSaveGraph.Reset()
 
-        clsDefaultBindRows.SetRCommand("bind_rows")
-        clsDefaultBindRows.AddParameter(".id", Chr(34) & "id" & Chr(34))
-        clsDefaultBindRows.SetAssignTo(strTemp:="Append", strTempDataframe:="Append")
+        clsBindRows.SetRCommand("bind_rows")
+        clsBindRows.AddParameter(".id", Chr(34) & "id" & Chr(34))
+        clsBindRows.SetAssignTo(strTemp:="Append", strTempDataframe:="Append")
```
dannyparsons: This isn't the default name is it? Because above you have set the prefix for the save control. I think the reset is enough here since there is no fixed default assign to name.

lilyclements: This is a good point, I have fixed and committed it now, thanks. This is ready for you to continue your review!

PR #2888, opened 2017-04-03
Import gridded data from IRI

stevekogo:


stevekogo:
This is not yet open for reviews.

stevekogo:
@rdstern , @lilyclements This is almost taking shape in terms of functionality. You can now try your hands on it as I implement more options and tidy up the code and dialog.

I recommend you try chirps and Tamsat data sources with their data files.

stevekogo:
The data frame is still being imported into a single data frame but that will be sorted out soon.

lilyclements:
This is looking good.
One problem I've come across is trying to import where the latitude is negative/southern

If I input a negative value then it crashes (e.g. Longitude: 10, 10.5 and Latitude -10.5, -10)

I believe on the IRI site they always have positive values and use S instead of N when it's below the equator.
From what I've tried, it always takes in the values as Northern. I think we need to have the option for the Southern hemisphere? Or, have I misunderstood how to use this for southern data?

Similar issue for East/West.

stevekogo:
In my understanding, importing data using expert mode from IRI site can take negative signs for the coordinates. The problem is that data may not be available on the selected coordinates. I am sending an example which works for CHIRPS

stevekogo:
Change the latitude ranges to -2 and -1.5 and see what happens. It seems to be downloading some data on my side but my internet is weak

lilyclements:
Okay I see. I tried with negative co-ordinates for the latitude (looking at Ghana so I could compare to other datasets I have) and it was crashing on this version. Yesterday morning it was giving "ERROR 404" on the data frame in R-instat. Did it download it in the end on your one? I'll try out the values you've suggested and explore.

From what I could see on the chirps site, the latitude ranges from -50 to 50. However on the IRI website, it was 50S - 0 - 50N. Do we input negative values for it to Work? Since the output for latitude gives an "N" after the number I assumed this meant we should input positive values and specify which cardinal direction

lilyclements:
Sorry just read the expert mode part of your message. I'll look into this further tomorrow! Thanks!

stevekogo:
@africanmathsinitiative/developers  This is in a fairly stable state. I welcome your reviews.

stevekogo:
@rdstern You are welcome to test and comment on this.

dannyparsons:
@stevekogo What does fairly stable mean? If there are known bugs or issues it would be helpful to list these if you want people to review it now or specify what aspects you would like feedback on, so reviewers don't spend time finding issues you already know about. Alternatively, delay asking for reviews until you feel it's complete.

stevekogo:
@dannyparsons There are additional functionalities which I am yet to implement but I the current version is stable enough. I also welcome suggestions on the implementation since this is purely my point of view.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -231,104 +231,243 @@ yday_366 <- function(date) {
...
+open_NetCDF <- function(nc_data){
+  variables = names(nc_data$var)
+  lat_lon_names = names(nc_data$dim)
+  lat_names = c("lat", "latitude","LAT","Lat", "LATITUDE")
+  lon_names = c("lon", "longitude","LON","Lon", "LONGITUDE")
+  time_names = c("time", "TIME","Time","period", "Period", "PERIOD")
+  lat_found = FALSE
+  lon_found = FALSE
+  time_found = FALSE
+  if (!lat_found){
```
dannyparsons: You don't need this check because you have set `lat_found` to FALSE just above.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -231,104 +231,243 @@ yday_366 <- function(date) {
...
+    for (i in lat_lon_names){
+      if(!is.na(match(i, lat_names))){
+        lat <- as.numeric(ncvar_get(nc_data, i))
+        lat_found = TRUE
```
dannyparsons: You can replace this whole loop with something like:
```
lat_in <- which(lat_lon_names %in% lat_names)
lat_found <- (length(lat_in) == 1)
if(lat_found) {
   lat <- as.numeric(ncvar_get(nc_data, lat_in))
}
```
And similarly for other variables. 

However, what if the latitude name is not one of these? We just give an error. Shouldn't there be a way for the user to optional specify the lat, long, time names in their file? Then if they don't specify it, we can try to work it out and give an error if we can't find it.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -231,104 +231,243 @@ yday_366 <- function(date) {
...
+        time <- as.numeric(ncvar_get(nc_data, i))
+        time_found = TRUE
       }
     }
+  }
+  
+  if(!lon_found || (!lat_found))stop("Latitude and longitude names could not be recognised.")
+  if(!time_found){
+    warning("Time variable could not be found/recognised. Time will be set to 1.")
+    time = 1
```
dannyparsons: Is this sensible? Why are we giving an error if we don't have lat or long but a warning for the time missing? If we don't have to have the time variable, is it sensible to just set it to 1? Can we just not include it?
I don't know about these files so maybe this does make sense.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -231,104 +231,243 @@ yday_366 <- function(date) {
...
+  } 
+  period <- rep(time, each = (length(lat)*length(lon)))
+  lat_rep <- rep(lat, each = length(lon))
+  lon_rep <- rep(lon, length(lat))
+  lat_lon <- as.data.frame(cbind(lat_rep, lon_rep))
+  names(lat_lon) = c("lat","lon")
```
dannyparsons: Shouldn't we be using the names of lat and long from the original file?

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -231,104 +231,243 @@ yday_366 <- function(date) {
...
+    if(lat_lon[j,1] < 0 && lat_lon[j,2] >= 0){
+      station = append(station, paste(paste("S", abs(lat_lon[j,1]), sep = ""), paste("E", lat_lon[j,2], sep = ""), sep = "_"))
+    }
+    if(lat_lon[j,1] >= 0 && lat_lon[j,2] < 0){
+      station = append(station, paste(paste("N", lat_lon[j,1], sep = ""), paste("W", abs(lat_lon[j,2]), sep = ""), sep = "_"))
+    }
+    if(lat_lon[j,1] < 0 && lat_lon[j,2] < 0){
+      station = append(station, paste(paste("S", abs(lat_lon[j,1]), sep = ""), paste("W", abs(lat_lon[j,2]), sep = ""), sep = "_"))
+    }
+  }
```
dannyparsons: This could become very slow. Avoiding loops when working with vectors is usually possible because most functions work on vectors as well as single values. Something like:

```
col1 <- lat_lon[ ,1]
col2 <- lat_lon[ ,2]
ifelse(col1 >= 0 & col2 >= 0, paste(paste("N", col1, sep = ""), paste("E", col2, sep = ""), sep = "_"), 
       ifelse(col1 < 0 & col2 >= 0, "second case", 
              ifelse(col1 >= 0 & col2 < 0, "third case..." , 
                     "final case"))))
```

I don't think you need to check the last case since it's the only option at that point. However, you might need to consider missing values in col1 and col2? That check could be done first if needed.

dannyparsons: Also you could use `-` instead of `abs`, since you know the sign of the value. Minus is probably faster than `abs`.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -231,104 +231,243 @@ yday_366 <- function(date) {
...
+  lat_lon_df <- cbind(lat_lon, station)
+  my_data <- cbind(period, lat_lon_df)
```
dannyparsons: `bind_cols()` from `dplyr` could do this in one line.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -231,104 +231,243 @@ yday_366 <- function(date) {
...
+    else if (length(dim(dataset))==2){
+      year <- dataset[1:length(lon), 1:length(lat)]
+      year = as.data.frame(t(year))
+      year = stack(year)
+      g <- as.numeric(year$values)
+      nc_value = append(nc_value, g)
+    }
+    else if (length(dim(dataset))==3){
+      for (k in 1:length(time)){
+        year <- dataset[1:length(lon), 1:length(lat), k]
```
dannyparsons: I think I need to understand better this part of doing but it looks inefficient having a loop inside a loop and then doing a subset which is very slow. We could probably do this better.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -231,104 +231,243 @@ yday_366 <- function(date) {
...
-      my_data = cbind(my_data, nc_value)
-      names(my_data)[length(names(my_data))]<-current_var
     }
-    return(list(my_data, lat_lon_df))
-  }
\ No newline at end of file
+    else{
+      stop("The format of the data cannot be recognised")
+    }
+    my_data = cbind(my_data, nc_value)
```
dannyparsons: Again `bind_cols` could be used outside the loop on a list to avoid doing a bind for every item in the loop.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -231,104 +231,243 @@ yday_366 <- function(date) {
...
+import_from_iri <- function(download_from, data_file, X1, X2,Y1,Y2){
+  gaugelocdir = getwd()
+  
+  if(download_from == "CHIRPS_V2P0"){
+    prexyaddress <- paste("https://iridl.ldeo.columbia.edu/SOURCES/.UCSB/.CHIRPS/.v2p0")
+    if(data_file == "daily_0p05"){
+      prexyaddress = paste(prexyaddress, ".daily/.global/.0p05/.prcp", sep="/")
+    }
+    else if(data_file == "daily_0p25"){
+      prexyaddress = paste(prexyaddress, ".daily/.global/.0p25/.prcp", sep="/")
```
dannyparsons: You could avoid doing paste many times by setting a string in here, like `extension <- ".daily/.global/.0p25/.prcp"` and then at the very end doing `prexyaddress <- paste(prexyaddress, extension, sep="/")`

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -231,104 +231,243 @@ yday_366 <- function(date) {
...
+      stop("The format of the data cannot be recognised")
+    }
+    my_data = cbind(my_data, nc_value)
+    names(my_data)[length(names(my_data))]<-current_var
+  }
+  return(list(my_data, lat_lon_df))
+}
+  
+import_from_iri <- function(download_from, data_file, X1, X2,Y1,Y2){
+  gaugelocdir = getwd()
```
dannyparsons: I don't think the working directory is the best place to create a temporary file. We can get into issues when users don't have write access to this location since the working directory will be in the user's Program Files. Often only admin users have write access there. We might have to pass in a temp folder location from the dialog?

stevekogo:
@danmyparsons Thanks for the detailed comments. I will have a look at them and improve the codes accordingly.

instat/static/InstatObject/R/stand_alone_functions.R
stevekogo: @dannyparsons Which location do you suggest?

dannyparsons: I think it should be a parameter to the function. From VB we can get a correct temporary directly so this could be passed in from a dialog. Happy that the working directory could be the default for that parameter.

instat/static/InstatObject/R/stand_alone_functions.R
stevekogo: @dannyparsons I agree with your suggestion of using original names. A challenge occurs when you want to create the link between the data frames. How are we going to access the latitude and longitude columns?

instat/static/InstatObject/R/stand_alone_functions.R
stevekogo: In this case, the data are stored in a 3 dimensional array with the third dimension being time hence we need to extract data for a specific time, unstack and move to the next time interval and do the same while appending the resulting stacked data

instat/static/InstatObject/R/stand_alone_functions.R
dannyparsons: This looks ok for how you've done it now using lat_lon_names and the index calculated.

However, also if these names can be passed in my the user it makes it even easier.

instat/static/InstatObject/R/stand_alone_functions.R
dannyparsons: I see, thanks. If you could send me one of these 3 dimensional arrays I can see how we might do this reshaping more efficiently.

dannyparsons: This might be useful for us too http://stackoverflow.com/questions/34968209/read-multidimensional-netcdf-as-data-frame-in-r

instat/static/InstatObject/R/stand_alone_functions.R
stevekogo: bind_cols works for data frames but not data_frame and a vector. In this case, nc_value is always a vector unless yu recommend converting it to a data frame and adding it to a list then use bind_cols thereafter

instat/static/InstatObject/R/stand_alone_functions.R
stevekogo: This links helps alot. The importation is  now faster.

stevekogo:
Hi Danny,

Attached is 3 dimensional netcdf fie.

Steve

instat/static/InstatObject/R/stand_alone_functions.R
dannyparsons: Ok I didn't realise that. That's not ideal but possibly is still faster, the binding will be slow so only doing that once should still improve the speed. You could try that and see which is faster. 
The `system.time` function can time how long code takes to run.

stevekogo:
@dannyparsons I am also fixing open netCDF dialog in this branch since I am editing the code here. I think this is almost taking shape

stevekogo:
@africanmathsinitiative/developers I have worked on two dialogs in this branch and it now ready for testing. The dialogs are open_netCDF and Import gridded data. 

Import gridded data has a bug on ordering the parameters and I am yet to fix it. Also, If someone is going to test this, try not to use wide longitude and latitude ranges since you will have to download big data files.

dannyparsons:
Is there a guide for testing the netCDF import and some example files?

dannyparsons:
Can you give some examples for gridded data for the Maseno team to try which are small downloads? I just a daily one and it's taking over 10 minutes to download so would be good to know good examples of small downloads.

stevekogo:
Attached are two netcdf files which can be used for testing. As with the
IRI data, I recommend less than 0.5 degree latitude and longitude
difference starting with the default.

dannyparsons:
You can't send netCDF file attachments in a GitHub comment, you need to email them to us separately.

stevekogo:
Github does not support .nc  extension files. I will send it as an email and copy everyone.

dannyparsons:
Just a few comments from trying the IRI dialog:

1. Latitude and Longitude columns are in a different order in the two data frames

2. There's inconsistencies about when N/S and E/W are specified. Sometimes they are added, sometimes not. I don't know what we want to do (or have an option) but we should be consistent. We could either include N/S and E/W or use positive and negative numbers.
The factor levels are sometimes strange. Look at the Logitude value in row 8. This is because the exact numeric value is used in making the factor (and it's confusingly in scientific form)
image

3. Sometimes I get download error - maybe because there is no data in that range? Can we give better error messages for this?

4. I can't get download from point working. Should it be working?

5. I wasn't thinking that the temporary file location should be an option on the dialog. I was thinking we should find a temporary folder through VB and pass this through in the function automatically. 
However, I quite like this now on the dialog since if the user has an error with access, they can specify their own location. But what we should do is fill in a location by default - you can see how a temporary location is found in temp saving graphs for the output window.
Also, you have the wrong dialog for this. You have save file dialog when it should be choose folder dialog. If I use this dialog I get an error.

6. How should the data be sorted on import? Currently it is by Time then Latitude then Longitude I think. It looks a bit strange to me to have all the "Jan 1981" together. I would have sorted by Time last - but I'm not sure what's a sensible default for us is? This isn't a big deal as the user can always sort using the sort dialog.

7. However! When I did sort, my time factor has the levels in the wrong order - "Apr 1981" is the first level. This should be corrected. Also, since this is climatic data we really should have a column of type Date as well/instead of? a factor.

8. It's not defined as climatic data. Can we do this step in the import? It should be possible since we know what the data looks like. If this will take long then maybe it's not a priority as the user can do it manually for now.

9. What should happen when I run the dialog twice? Should I get 4 data frames or 3? i.e. should there only be one lat/long data frame linked to both data frames? I can see cases where either would be wanted. Probably a longer discussion needed on this.

stevekogo:
@dannyparsons Thanks for the detailed review. I will try to respond/fix each of them.

stevekogo:
@dannyparsons download from point should now be working.

stevekogo:
@dannyparsons Can you check the order of parameter. I don't know why they do not appear according to the specified order.

dannyparsons:
Possibly you need to the add the parameter positions to the default function parameters as well. That would be my guess.

stevekogo:
@dannyparsons Try 10.025, 10.025 for the point chirps data

stevekogo:
39, 4 works fine for chirps 0p25

dannyparsons:
@stevekogo I'm assuming this is urgent for today? I think most of my points from yesterday should be addressed before it's merged apart from a few which might need more thought or discussion.

stevekogo:
@africanmathsinitiative/developers This is ready for a re-review

stevekogo:
@dannyparsons I took out the options for specifying where to download the data and provided a temporary directory through the dialog

dannyparsons:
As I said, I was happy with that on the dialog but fine to take it out for now.

stevekogo:
The deadline is so tight and I could not get that functioning well. I will bring that feature back after we get a version.

dannyparsons:
No problem, good idea to remove if not working.

I get this error on IRI import but after the error the data still imports. Do you know the cause?

image

lilyclements:
I'm on the designer, shall I remove it?

stevekogo:
@dannyparsons I had deleted some line accidentally. I was not affecting the import but was affecting the linking.

instat/dlgImportGriddedData.vb
```diff
@@ -33,24 +33,218 @@ Public Class dlgImportGriddedData
...
+        TestOkEnabled()
+    End Sub
+
+    Private Sub pnlControls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrPnlGetArea.ControlContentsChanged
+        If rdoArea.Checked Then
+            lblMinLat.Text = "Minimum Latitude"
+            lblMinLon.Text = "Minimum Longitude"
+        ElseIf rdoPoint.Checked Then
+            lblMinLat.Text = "Latitude"
+            lblMinLon.Text = "Longitude"
```
lilyclements: These all need colons, we want:
```
            lblMinLat.Text = "Minimum Latitude:"
            lblMinLon.Text = "Minimum Longitude:"
        ElseIf rdoPoint.Checked Then
            lblMinLat.Text = "Latitude:"
            lblMinLon.Text = "Longitude:"
```

stevekogo:
@lilyclements Notify me once you are done editing the designer.

dannyparsons:
Great that's fixed now.

What about my other points? 2, 7 (and 8 but not urgent) still seem unresolved.

Also, when I run the dialog twice with the same names, why does IRI_Library_Data overwrite the previous data but the Lat_Lon_Data doesnt?

stevekogo:
"Also, when I run the dialog twice with the same names, why does IRI_Library_Data overwrite the previous data but the Lat_Lon_Data doesnt?"
I am using a prefx for lat_lon. I should use a prefix for the main data frame too

stevekogo:
That should now be fixed

stevekogo:
As with issue 2, 7 and 8 we can implement those later. Issue 2 is somehow technical since we are getting the data in raw for and the files differ in the way longitude and latitudes are stored.

lilyclements:
What changes, if any, need to be made now before Danny can merge?

stevekogo:
This can now be merged

dannyparsons:
If that's the format the data is coming in from IRI by default then I am less concerned.

PR #2895, opened 2017-04-04
added feature to edit column and data frame metadata

dannyparsons:


dannyparsons:
@africanmathsinitiative/developers this now makes parts of the metadata grids editable. Please test with interesting data sets and check for bugs

rdstern:
This is great.  It is sort of what SPSS does except:
a) In SPSS you choose between the data view and the variable (metadata) view.  We have both!
b) there is no sheet concept, because you only have one sheet at a time.
Enough praise - here is what I have found so far.
Small data set - from File > New with 100 observations.  Then generated 2 random samples each of length 100 from exponential.
a) Names seems to work.  I changed 2 names and they changed in the main grid.
b) Label field **could be made wider**, but works.  I put a (variable) label in and then tried the Describe > One Variable > Frequencies and it was picked up
c) The label field for the generated columns had NA in them.  The initial ones were blank.  I would like blank.
d) in the Dataframe metadata the name field spills out into the next field.
e) The label in the dataframe metadata is in a small font.  It also spills into the next field if full. 
f) When I changed the name of the sheet it also changed the order.  (I had 2 more sheets and it moved from first to last in the grid and also in the tab in the column metadata.
g) Now a nasty one!  I opened the SPSS file I was using earlier - with 260 columns.  It took a few seconds before, but this time it took a very long time (perhaps 3 minutes - I wasn't timing.   
h) The (unnecessary) bottom scroll bar in the column metadata is still too wide

PR #2897, opened 2017-04-05
improved efficiency of some R methods (particularly for column metadata)

dannyparsons:


dannyparsons:
@rdstern you could now try your wide data set on the latest version.

PR #2907, opened 2017-04-07
Summary Tables - layout

lilyclements:
@dannyparsons This is ready for you. @shadrackkibet sorted the TestOK and I've just edited the layout on Summary Tables.
I'm not sure if I understand this dialog fully however. It is very similar to "New Summary Tables", are they meant to be the same dialog where this one was made around a year before that dialog?
I updated the controls and layout regardless as that's been waiting for me to attack it for a few weeks now...

### DIFF
--- a/instat/dlgSummaryTables.vb
+++ b/instat/dlgSummaryTables.vb
@@ -13,11 +13,20 @@
 '
 ' You should have received a copy of the GNU General Public License k
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
-Imports instat
 Imports instat.Translations
 
 Public Class dlgSummaryTables
+    Private bFirstload As Boolean = True
     Private Sub dlgSummaryTables_Load(sender As Object, e As EventArgs) Handles MyBase.Load
+        autoTranslate(Me)
+        If bFirstload Then
+            InitialiseDialog()
+            bFirstload = False
+        End If
+        TestOkEnabled()
+    End Sub
+
+    Private Sub InitialiseDialog()
         cboWeights.Visible = False
         ucrBase.clsRsyntax.SetFunction("summary")
         ucrBase.clsRsyntax.iCallType = 2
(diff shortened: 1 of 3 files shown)


instat/dlgSummaryTables.vb
```diff
@@ -1,46 +1,73 @@
...
     End Sub
 
-    Private Sub ucrReceiverFactor_Leave(sender As Object, e As EventArgs) Handles ucrReceiverFactor.Leave
-        ucrBase.clsRsyntax.AddParameter("x", clsRFunctionParameter:=ucrReceiverFactor.GetVariables())
+    Private Sub TestOkEnabled()
+        If Not ucrReceiverFactor.IsEmpty AndAlso Not ucrReceiverVariate.IsEmpty Then
+            ucrBase.OKEnabled(True)
+        Else
+            ucrBase.OKEnabled(False)
+        End If
```
dannyparsons: @shadrackkibet This might be the correct code when the dialog is working, but for now we want OK to always be disabled to avoid an error, don't we?

shadrackkibet: @dannyparsons .Yes, that is the correct code required but for now  i have disabled OK to avoid error on this dialogue.

dannyparsons: Where have you disabled OK?

shadrackkibet: Initially this dialog was not run on first load since we didn't have bFirstLoad so that's why on first load testOk is enabled.So I decided enable to always be false which solves the issue partially.

dannyparsons: This doesn't work. If I reopen the dialog and both receivers have variables in, then OK will be enabled. What you can do is comment out the code you have now (because that can be used later) and instead make TestOK just call `ucrBase.OKEnabled(False)` for now.

shadrackkibet: @lilyclements so can I go on and edit this branch or can you make the suggested changes by @dannyparsons ??

dannyparsons: You can either pull this branch into your branch, or ask Lily to make the change. Normally, I would say Lily could just make the change because it's quicker, but in this case you may want the practice of working with branches and doing the code and comments correctly.

lilyclements: You can merge my amended branch into your branch and make the suggested changes in your branch

PR #2912, opened 2017-04-10
implemented package name for RFunctions with example on scatter and one variable model dialogs

dannyparsons:


dannyparsons:
@africanmathsinitiative/developers this implements having the package name associated with RFunctions. I've done examples on two dialogs but it's just one extra line for each function. This can be implemented now in all our dialogs. 

Once this is done we can then change the setup so that we do not need to load a big list of packages every time the software starts.

instat/clsRSyntax.vb
```diff
@@ -46,6 +46,17 @@ Public Class RSyntax
         bUseCommandString = False
     End Sub
 
+    Public Sub SetPackageName(strName As String)
+        If clsBaseFunction Is Nothing Then
+            MsgBox("Developer error: base function must be set before package name is set.")
```
dannyparsons: @africanmathsinitiative/developers note that if you try to set the package name through the R syntax before the base function has been set you will get this message and it won't apply.

instat/dlgOneVarFitModel.vb
```diff
@@ -241,7 +243,8 @@ Public Class dlgOneVarFitModel
     End Sub
 
     Private Sub SetNonSignTest()
-        clsRNonSignTest.SetRCommand("signmedian.test:: signmedian.test")
+        clsRNonSignTest.SetPackageName("signmedian.test")
+        clsRNonSignTest.SetRCommand("signmedian.test")
```
dannyparsons: @africanmathsinitiative/developers Note I have only done this for a few functions in this dialog. The others still need their package names added.

dannyparsons:
Also remember that base R functions and Instat object methods do not need a package name to be set.

instat/dlgOneVarFitModel.vb
lilyclements: Why are we doing it this way and not as:
`clsRNonSignTest.SetRCommand("signmedian.test:: signmedian.test")` ?

dannyparsons: So that we have the package name and function name stored separately. This is more flexible. For example, with this way we could easily have an option of displaying/not displaying package names. If the way of calling a package somehow changed in R we would only need to change one line of code and not every dialog. We could also do things like get a list of all packages needed for all the functions used on a dialog. 
This is all easy when the package name and function name are stored separately but would be much harder if they weren't. Good question.

PR #2916, opened 2017-04-11
Maxwellfundi ggplots lineplot

dannyparsons:
@maxwellfundi This implements ucrVariablesAsFactor as the core control and implemented on scatter plot dialog (sorry @lilyclements I didn't realise you had already done this)

I believe the links to the sub dialogs are still working but we need to quickly get those working in the new system too.

Also, the graph dialogs use ucrSaveGraph and I suggest we replace this with ucrSave since it''s already set up well in the new system. For example, the resetting of ucrSaveGraph currently doesn't work on scatter.

I suggest this could be merged if it looks in good shape and then @maxwellfundi can continue with this.

### DIFF
--- a/instat/dlgScatterPlot.vb
+++ b/instat/dlgScatterPlot.vb
@@ -110,38 +110,6 @@ Public Class dlgScatterPlot
         ucrFactorOptionalReceiver.SetRCode(clsRaesFunction, bReset)
     End Sub
 
-    'Private Sub ucrSelectorForScatter_DataFrameChanged() Handles ucrSelectorForScatter.DataFrameChanged
-    '    clsRggplotFunction.AddParameter("data", clsRFunctionParameter:=ucrSelectorForScatter.ucrAvailableDataFrames.clsCurrDataFrame)
-    'End Sub
-
-    'Private Sub SetXParameter()
-    '    If Not ucrReceiverX.IsEmpty Then
-    '        clsRaesFunction.AddParameter("x", ucrReceiverX.GetVariableNames(False))
-    '    Else
-    '        clsRaesFunction.AddParameter("x", Chr(34) & Chr(34))
-    '    End If
-    'End Sub
-
-    'Private Sub SetYParameter()
-    '    If Not ucrVariablesAsFactorForScatter.IsEmpty Then
-    '        clsRaesFunction.AddParameter("y", ucrVariablesAsFactorForScatter.GetVariableNames(False))
-    '    Else
-    '        clsRaesFunction.AddParameter("y", Chr(34) & Chr(34))
-    '    End If
-    'End Sub
-    'Private Sub ucrReceiverX_SelectionChanged(sender As Object, e As EventArgs) Handles ucrReceiverX.SelectionChanged
-    '    SetXParameter()
-    '    TestOkEnabled()
-    'End Sub
-
-    'Private Sub ucrFactorOptionalReceiver_SelectionChanged(sender As Object, e As EventArgs) Handles ucrFactorOptionalReceiver.SelectionChanged
-    '    If Not ucrFactorOptionalReceiver.IsEmpty Then
-    '        clsRaesFunction.AddParameter("colour", ucrFactorOptionalReceiver.GetVariableNames(False))
-    '    Else
-    '        clsRaesFunction.RemoveParameterByName("colour")
-    '    End If
-    'End Sub
-
     Private Sub TestOkEnabled()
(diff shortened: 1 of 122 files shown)


instat/dlgScatterPlot.vb
```diff
@@ -49,12 +49,16 @@ Public Class dlgScatterPlot
...
         ucrVariablesAsFactorForScatter.SetIncludedDataTypes({"factor", "numeric"})
+        ucrVariablesAsFactorForScatter.SetValuesToIgnore({Chr(34) & Chr(34)})
+        ucrVariablesAsFactorForScatter.bAddParameterIfEmpty = True
 
         ucrReceiverX.SetParameter(New RParameter("x", 0))
         ucrReceiverX.SetParameterIsString()
         ucrReceiverX.bWithQuotes = False
         ucrReceiverX.Selector = ucrSelectorForScatter
         ucrReceiverX.SetIncludedDataTypes({"factor", "numeric"})
+        ucrReceiverX.SetValuesToIgnore({Chr(34) & Chr(34)})
```
dannyparsons: So this is saying - if the parameter has this value "ignore it" i.e. don't put this value in the receiver. Since we want the parameter to be able to have this value, but don't want this value displayed in the receiver. If you see where these list of values are used, if it is any of these, it will instead clear the receiver which is what we want. Good question.

maxwellfundi:
@dannyparsons  i think this is in a sort of stable state and the bug for empty string is now fixed. You could merge this and get the rest  done

instat/dlgScatterPlot.vb
lilyclements: Okay, so that I can check I understand - in the original version of this (and in my version) in the dialog is an "If" statement that if the receiver is empty then to pass through `Parameter.Name = ""`, is this just a method in one line to say to pass in an empty string if nothing else is in the receiver?

instat/ucrReceiver.vb
```diff
@@ -359,19 +359,25 @@ Public Class ucrReceiver
         End If
         'Could need bParameterIsString and bParameterIsRFunction to be properties of RParameter if two functions need string/function
         If bParameterIsString Then
-            clsTempParam.SetArgumentValue(GetVariableNames(bWithQuotes))
+            'TODO this currently only works with one value to ignore. Also may need option not to set parameter value to strValuesToIgnore
+            '     although this currently can be done with bAddParameterIfEmpty = True
+            If IsEmpty() AndAlso strValuesToIgnore IsNot Nothing AndAlso strValuesToIgnore.Count = 1 Then
+                clsTempParam.SetArgumentValue(strValuesToIgnore(0))
```
dannyparsons: @lilyclements Here is the line in the receiver where the `= ""` is being added if the receiver is empty.

lilyclements:
I agree that this can be merged, I can't find any bugs on the Scatterplot dialog when I tested it on this version. Lots of little finicky changes to make, but I will make them gradually once the dialog codes are updated/etc

instat/ucrReceiver.vb
lilyclements: Okay, I understand. However, what is `strValuesToIgnore`?

dannyparsons: I think I explained this here https://github.com/africanmathsinitiative/Instat/pull/2916#discussion_r110919034 - not sure if it's clear?

dannyparsons: The `bAddParameterIfEmpty = Ture` is slightly different and is saying, if the receiver is empty do NOT remove the parameter. Of course this usually wouldn't make sense unless you also set something for `strValuesToIgnore`.

lilyclements: Yes that makes sense, thanks.

instat/dlgScatterPlot.vb
```diff
@@ -89,6 +93,8 @@ Public Class dlgScatterPlot
 
         clsRaesFunction.SetPackageName("ggplot2")
         clsRaesFunction.SetRCommand("aes")
+        clsRaesFunction.AddParameter("x", Chr(34) & Chr(34))
+        clsRaesFunction.AddParameter("y", Chr(34) & Chr(34))
```
maxwellfundi: @dannyparsons , why are you adding parameter y and x here? i thought we just needed the x only to have ""

dannyparsons: why do we only need `x = ""`?

maxwellfundi: Because we can have a scatter plot with just the y, so we need to pass in an empty x

dannyparsons: Yeah that's why we need it for x. Maybe I should rephrase: why don't we need this for y?

PR #2922, opened 2017-04-12
fixed bug with sub dialog combo boxes not resetting

dannyparsons:


instat/sdgOneWayFrequencies.vb
```diff
@@ -92,6 +92,7 @@ Public Class sdgOneWayFrequencies
         dctHorizontalPositionLabel.Add("Outward", Chr(34) & "outward" & Chr(34))
         ucrInputHorizontalLabels.SetItems(dctHorizontalPositionLabel)
         ucrInputHorizontalLabels.SetRDefault(Chr(34) & "center" & Chr(34))
+        ucrInputHorizontalLabels.bUpdateRCodeFromControl = False
```
dannyparsons: @stevenndungu This fixes the resetting issue. The problem was that this is set to True by default for input controls. This means that the control can update the R code if it has a value - even when resetting. For some controls this is useful - like the text box in the save control where we want the default value to update the R code. Here we always want to reset to the R default so we set this to True.

PR #2924, opened 2017-04-13
dlgLineplot

maxwellfundi:
The ucrChkPoints does not reset properly yet. 
This will be fixed soon.


stevekogo:
@maxwellfundi What is the current state of this? Can be opened for reviews any time soon?

maxwellfundi:
@stevekogo this is still on the list of things to be done. 

Soon it will . 

Thanks.

maxwellfundi:
@africanmathsinitiative/developers this is ready for review. I have fixed the issue with the checkbox

maxwellfundi:
@lilyclements I have added in the parameter positions as well as renaming the dialog from dlgPlot to dlgLinePlot. 

@dannyparsons this is now ready for you to review.

dannyparsons:
There is certainly a general issue with the multiple variables and the options sub dialogs which we need to address, but it doesn't need to be sorted in this updating. Disabling the sub dialog when multiple variables selected is the best we can do for now I think.

lilyclements:
I still managed to crash it despite that disabling, by making the sdg changes when on "Single" and then switching to "Multiple". If this is okay as it is, however, then @dannyparsons this is ready for you to review

dannyparsons:
Yes that is still a way to break it but hopefully fairly uncommon. Once the sub dialogs have been updated to the new system I think this will be easier to resolve.

instat/dlgLinePlot.vb
```diff
@@ -1,203 +1,200 @@
...
+        End If
+    End Sub
+
+    Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
+        SetDefaults()
+        SetRCodeForControls(True)
+    End Sub
+
+    Private Sub TempOptionsDisabledInMultipleVariablesCase()
+        ' Why is this done??
```
dannyparsons: Is this clear now? You can remove this if so.

maxwellfundi: Yeah, i know why we needed this. I think it will be a good temporary solution.

lilyclements: We have something similar to this in some other dialogs, such as line plot. However, it needs to be implemented in some dialogs - such as scatter

dannyparsons: Good so this comment should be removed from the code.

maxwellfundi:
@dannyparsons @lilyclements This works well, we are disabling the options when multiple receiver is the active receiver for ucrvariables as factor

PR #2938, opened 2017-04-20
One way frequencies

stevenndungu:


stevenndungu:
Note: Happy to get critics, but one thing to note is that on the rdoBoth and rdoGraph, Only one graph is produced (actually the last variable graph) to be plotted on the  output window when n graphs should be produced depending on the number of variables selected (n) . Still thinking on how to solve this (suggestions on how to solve it are welcome).

rdstern:
My understanding is that you will only produce one graph, but it will be a facetted/combined graph.  Here is an example from page 101 in the sjplot guide.  This was a main reason you changed (I hope) to using the sjplot command.

# frequencies, as plot grid
efc %>%
select(e42dep, c172code, e16sex, c161sex) %>%
sjplot() %>%
plot_grid()

instat/dlgOneWayFrequencies.vb
```diff
@@ -75,7 +80,6 @@ Public Class dlgOneWayFrequencies
         ucrChkGroupData.AddToLinkedControls(ucrNudGroups, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=10)
         ucrChkGroupData.AddParameterPresentCondition(True, "auto.group")
         ucrChkGroupData.AddParameterPresentCondition(False, "auto.group", False)
```
lilyclements: Should this be
ucrChkGroupData.AddParameterPresentCondition(False, "auto.group", "False")

instat/dlgOneWayFrequencies.vb
```diff
@@ -84,40 +88,43 @@ Public Class dlgOneWayFrequencies
...
+        clsSjPlot = New RFunction
 
         ucrSelectorOneWayFreq.Reset()
         ucrReceiverOneWayFreq.SetMeAsReceiver()
 
-        clsSjtFreq.SetRCommand("sjt.frq")
-        clsSjtFreq.AddParameter("show.summary", "FALSE")
-        clsSjtFreq.AddParameter("skip.zero", "FALSE")
-        clsSjtFreq.AddParameter("digits", 0)
+        clsSjTab.SetRCommand("sjtab")
```
lilyclements: You need to do .AddPackageName for both classes

stevenndungu:
Yes that was what we agreed (the two forms of using sjplot and sjtab got me confused a little bit). As well, it  applies the gglots format of implementation which apparently the system of passing the operators had not been fully in place to see the graphs changed to the new system. But Maxwell tells me that everything is set so I can change this quite easily. Thanks

@lilyclements  could you hesitate reviewing this ...I make some changes first...

lilyclements:
@stevenndungu I'm mid-review, I'll send my review to you now of what I've put so far.
I've also made minor changes to the code. If you haven't already made changes, I can send this to you to push into your branch. Otherwise, i'll explain the changes and you can add them

instat/dlgOneWayFrequencies.vb
stevenndungu: Yap, so that when the checkbox is unchecked the parameter `auto.group` is not passed. first `false` is for the state of the control, then the last `false` is for ensuring the parameter is absent on uncheck of the checkbox.

lilyclements: Oh I see! Thank you for explaining.

dannyparsons:
I didn't know about the plot_grid command - this makes it much easier. @stevenndungu you are right that operators can be done easily in the new system, however note that the code Roger sent is essentially the same as:
`plot_grid(sjplot(survey))`

So we can do it in the standard way we do functions.

Note: when I tried this on the survey data the output doesn't look very nice depending on the size of the image, see below. Does anyone else see this with their data sets?


image

rdstern:
I agree the default presentation is not always great.  It is sometimes better flipped.  And that is also why we have the options to not display the counts or the percentages.  We could even make that the default, so the user needs to opt to add them.

stevenndungu:
Thanks @rdstern @dannyparsons this is a good suggestion since it works in the same way. I have actually changed to `plot_grid` for the place of `sjplot` function. But retained the pipe way for the `sjtab ` function.

 Few things worth Noting as far as `sjplot`  and `sjtab` functions are in concern for this dialogues.
 -We have  a nice additional function  `group_by` option . Should we add this as an optional feature on the table option (?)
 for example     `efc %>%
   group_by(e16sex, c172code) %>%
  select(e42dep, n4pstu, e16sex, c172code) %>%
  sjtab()`
 - `weight.by` parameter behaves strangely to me for the case of `sjtab` function. It applies the weights to the last variable of the variables applied. for instance in the example below it provides a table for c172code variable only excluding others.
                    `efc %>%
         select(e42dep, n4pstu, e16sex, c172code) %>%
     sjtab(weight.by=efc$c172code)`
Note: for the graphs, the weights applies to all variables (my observation)
 - Both `plot_grid()` and using pipe function to plot several variables, produces individual plots for all the variables the finally generates a single plot grid of all the variables. This explains the why it takes long to plot the graph in the output window. I would suggest that we have both cases generated for the variables selected, ie, individual plots then finally the grid plot. So the user is at liberty to use whichever type of plot generated.

stevenndungu:
@dannyparsons This is in good shape now with the issue of adding additional pair of the second function is still issue. Two way frequencies has the same issue remaining to be fixed othws everything working.

stevekogo:
@stevenndungu follow up with @dannyparsons to get the issue sorted.

dannyparsons:
I will look at this today and try resolve the remaining issues.

stevenndungu:
After solving the problem I will be happy to know what was wrong, so that I can be able to solve the same issue on two way and as well help Shadrack on the same issue with String Handling (which has same issue of additional pair)

dannyparsons:
Great, hopefully it's a simple issue which I can discuss with you both and then you can resolve on other dialogs.

stevenndungu:
I suppose, since with slight alteration of the setrcode it works perfectly.

stevenndungu:

stevenndungu:
@mmumbo thanks for offering to test this.

instat/dlgOneWayFrequencies.vb
```diff
@@ -41,8 +41,9 @@ Public Class dlgOneWayFrequencies
         'HelpID
         ' ucrBase.iHelpTopicID = 
 
-        ucrReceiverOneWayFreq.SetParameter(New RParameter("data", 0))
+        ucrReceiverOneWayFreq.SetParameter(New RParameter("data"))
```
maxwellfundi: I think you could also give this parameter a position

stevenndungu:
Checking on it. Thanks

stevenndungu:
Not able to notice this. Perhaps you we can discuss on skype

stevenndungu:
Seen that, sorry

stevenndungu:
@lilyclements this can be solved by reseting the ucrSave control by adding `sdgOneWayFrequencies.ucrSaveGraph.Reset()`  on SetDefault() of the main dialogue. Which I need to consult with @dannyparsons first.

dannyparsons:
I'll merge this so you can move on to other dialogs but we will come back to the save graph reset issue.

PR #2940, opened 2017-04-20
dlgBoxplot

maxwellfundi:


stevekogo:
@maxwellfundi What is the current state of this?

maxwellfundi:
@africanmathsinitiative/developers this is ready for reviews. 

Thanks.

lilyclements:
Ive made some changes to merge, but it wasn't thorough enough. This needs a much more thorough review.
Also, I think the main receiver has a bug, you're telling it to set numeric but it isn't setting numeric only. We saw this problem on Histogram too where it causes a crash.

maxwellfundi:
@lilyclements yes, i have realized that this has that the main receiver has that bug. Let me investigate it actually.

instat/dlgBoxPlot.vb
```diff
@@ -13,137 +13,178 @@
...
+        ucrSecondFactorReceiver.SetParameterIsString()
+        ucrSecondFactorReceiver.bWithQuotes = False
+
+        ucrChkVarWidth.SetParameter(New RParameter("varwidth", 1))
+        ucrChkVarWidth.SetText("Variable Width")
+        ucrChkVarWidth.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
+        ucrChkVarWidth.SetRDefault("FALSE")
+
+        clsCoordFlipFunc.SetPackageName("ggplot2")
+        clsCoordFlipFunc.SetRCommand("coord_flip")
```
lilyclements: Just wondering, what is the benefit to setting this in `Initialise `as opposed to `SetDefaults`?

maxwellfundi: We are defining a paremeter for the ucrChkHorizontalBoxplot  which is a  function. Which we are setting it below on line 95. We can set a parameter that does not exist and so we have to do it here .

lilyclements: Ah okay, I see! Thank you

lilyclements:
@dannyparsons This is ready for review

PR #2944, opened 2017-04-21
DlgView enhancement

lilyclements:
Adding in `sjt.df` option to dlgView.
In progress


instat/dlgView.vb
```diff
@@ -73,13 +78,26 @@ Public Class dlgView
...
+        ' Linking for when rdoHTML is checked
+        ucrPnlDisplayWindow.AddToLinkedControls(ucrChkSortColumn, {rdoHTMLOutputWindow}, bNewLinkedHideIfParameterMissing:=True, bNewLinkedAddRemoveParameter:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=True)
+        ucrChkSortColumn.SetText("Sort by Column")
+        ucrChkSortColumn.AddToLinkedControls(ucrReceiverSortCol, {True}, bNewLinkedHideIfParameterMissing:=True, bNewLinkedAddRemoveParameter:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=rdoTop)
+        ucrChkSortColumn.SetRDefault("NULL")
+
+        ucrReceiverSortCol.SetParameter(New RParameter("sort.col", 3))
+        ucrReceiverSortCol.SetParameterIsString()
+        ucrReceiverSortCol.Selector = ucrSelectorForView
+        ' TODO: Options in this selector should be limited to options which are also in the other selector.
```
lilyclements: @dannyparsons I don't think what I want to do here has been implemented yet. if it has, could you point me in the right direction to a similar dialog? 
What I need to do is restrict the items that can go into this single receiver with items that are also in the multiple receiver.

Since this isn't a priority, if it can't be fixed easily now - my alternative solution is to write up an issue on it and disable TestOk if the `.Text` in the two receivers are not the same, with a warning stating "The item in the "Sort Column" Receiver is not also an item in the Main Receiver. OK will not be enabled" or something along those lines.
However, I don't like this idea of fixing it. I want to avoid pop-up boxes as much as possible since I feel they can be quite irritating for the user.

Is there currently a way to implement this?

lilyclements:
@dannyparsons I have been working on this with @stevenndungu and it's ready for you to review

AlexSananka:
Just before @dannyparsons reviews this,I want to test this dialog first.

AlexSananka:
image

AlexSananka:
@lilyclements are you aware of that bug?It appears when I only put one variable in the receiver.should we restrict our users to put more than one variable in the receiver or is anyone thinking of a better solution to this?Maybe @stevenndungu  can clarify.

stevenndungu:
This could be to fixed  by the line `ucrReceiver.bForceAsDataFrame = True`. However for this dialogue it is slightly special, more than that but am figuring out on how to fixed this easily.

dannyparsons:
Before thinking of ways to avoid the bug in the dialog, understand what the error means. 

The function is requesting a data frame for the `mydf` argument. By default, when a single column is selected, the instat object returns it as a vector because that is what is normally required. For this dialog we always need a data frame, even when there is only one column, hence Steven's suggestion will solve this.

So there is no need to restrict the dialog to move than one variable when there's no reason this shouldn't be possible.
Making the dialog avoid a bug by restricting its functionality is a last resort only after understanding the error and then if we are sure there is not another way to fix it.

AlexSananka:
@dannyparsons and that's why I asked if their is a better solution to solve that.Of course it was so obvious that a data frame is always needed in the receiver(the error is very informative itself).Sounds great that @stevenndungu suggestion solves the issue.

rdstern:
And the discussion itself has been interesting, because it has produced some principles!

instat/dlgView.vb
```diff
@@ -41,6 +41,7 @@ Public Class dlgView
         ucrReceiverView.SetParameter(New RParameter("x", 0))
         ucrReceiverView.SetParameterIsRFunction()
         ucrReceiverView.Selector = ucrSelectorForView
+        ucrReceiverView.bForceAsDataFrame = True
```
lilyclements: Sorry for my delay in responding! Thank you @stevenndungu for suggesting how to fix it.
I've added it here and checked and it has solved the bug.

instat/dlgView.vb
```diff
@@ -198,7 +199,7 @@ Public Class dlgView
         DataFrameLength()
     End Sub
 
-    Private Sub Controls_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlDisplayFrom.ControlValueChanged, ucrPnlDisplayWindow.ControlValueChanged, ucrChkSpecifyRows.ControlValueChanged
+    Private Sub Controls_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlDisplayFrom.ControlValueChanged, ucrPnlDisplayWindow.ControlValueChanged, ucrChkSpecifyRows.ControlValueChanged, ucrReceiverView.ControlValueChanged
```
lilyclements: While checking that the `.bForceAsDataFrame = True` works I came across another bug which I've fixed here

instat/dlgView.vb
lilyclements: Update: I was wrong, it hasn't solved the bug! Sorry

instat/dlgView.vb
```diff
@@ -88,8 +91,28 @@ Public Class dlgView
...
+        ucrReceiverView.SetRCode(clsMainFunction, bReset)
+        ucrPnlDisplayWindow.SetRCode(clsMainFunction, bReset)
+        ucrPnlDisplayFrom.SetRCode(clsMainFunction, bReset)
+        ucrNudNumberRows.SetRCode(clsMainFunction, bReset)
+        ucrChkSpecifyRows.SetRCode(clsMainFunction, bReset)
+        ucrReceiverView.AddAdditionalCodeParameterPair(clsViewDataFrame, New RParameter("mydf"), iAdditionalPairNo:=1)
+        ucrSelectorForView.SetRCode(clsMainFunction, bReset)
         DataFrameLength()
+        bControlsUpdated = True
     End Sub
```
lilyclements: I believe it stops the base function from changing when pressing reset? Although perhaps @dannyparsons can clarify this further for us!

instat/dlgView.vb
```diff
@@ -169,7 +169,7 @@ Public Class dlgView
             clsViewDataFrame.SetRCommand("sjt.df")
             clsViewDataFrame.RemoveParameterByName("x")
             clsViewDataFrame.RemoveParameterByName("title")
-            clsViewDataFrame.AddParameter("mydf", clsRFunctionParameter:=ucrReceiverView.GetVariables)
+            clsViewDataFrame.AddParameter("mydf", clsRFunctionParameter:=ucrReceiverView.GetVariables(True))
```
lilyclements: This fixes the bug. By default, in `.GetVariables `it runs `bForceAsDataFrame = False` by default, meaning that we were no longer forcing it to be a dataframe. Adding `True` means that it works with one variable

lilyclements: @stevenndungu This fixes the bug. By default, in `.GetVariables `it runs `bForceAsDataFrame = False` by default, meaning that we were no longer forcing it to be a dataframe. Adding `True` means that it works with one variable

instat/dlgView.vb
```diff
@@ -139,19 +160,50 @@ Public Class dlgView
             Else
                 ucrBase.clsRsyntax.SetBaseRFunction(ucrReceiverView.GetVariables())
             End If
-        Else
-            clsMainFunction.SetRCommand("View")
+        ElseIf rdoDispSepOutputWindow.Checked Then
             ucrBase.clsRsyntax.SetBaseRFunction(clsMainFunction)
+            clsMainFunction.SetRCommand("View")
```
lilyclements: I have called this sub in the `.ClickOK` event now. Is that okay?
What are we calling this change of base functions when the user clicks ok?

lilyclements:
@AlexSananka this is ready for a re-review

AlexSananka:
Sounds great,I am going to test it.

AlexSananka:
This dialog is working fine just one tiny thing and it will be ready,when I click on Display in output window and then reset it doesn't reset but it works when when I click on display HTML in ouput window.

Once this is sorted out it will be ready for @dannyparsons

AlexSananka:
Until I uncheck the specify rows checkbox and it resets,otherwise when it is checked it doesn't reset.

lilyclements:
Thanks @AlexSananka I've fixed this now in the most recent commit. This is ready for re-review

AlexSananka:
This now works fine and it's ready for @dannyparsons

instat/dlgView.vb
stevenndungu: Wauw! Great. I had not checked that.

lilyclements:
@dannyparsons I have made the necessary changes. This is ready for review

instat/dlgView.vb
```diff
@@ -120,38 +159,58 @@ Public Class dlgView
...
+    Private Sub ucrSelectorForView_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSelectorForView.ControlValueChanged
+        DataFrameLength()
+    End Sub
+
+    Private Sub Controls_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlDisplayFrom.ControlValueChanged, ucrPnlDisplayWindow.ControlValueChanged, ucrChkSpecifyRows.ControlValueChanged, ucrReceiverView.ControlValueChanged
+        ChangeFunctionParameters()
+    End Sub
+
+    Private Sub ucrBase_ClickOk(sender As Object, e As EventArgs) Handles ucrBase.ClickOk
+        ChangeFunctionParameters()
```
dannyparsons: I don't see why this needs to be done on ClickOK? It should already be correct by this point I think.

lilyclements:
@dannyparsons this is ready for re-review

PR #2947, opened 2017-04-23
dlgImportFromODK - Tidy up

lilyclements:

After showing this dialog today I noticed a few things to change. Here I have:
* Changed radio buttons to button buttons
* Tidied dialog
* Moved code around a bit
* Potential issue fix? Outlined below

### DIFF
--- a/instat/dlgImportFromODK.vb
+++ b/instat/dlgImportFromODK.vb
@@ -95,7 +95,16 @@ Public Class dlgImportFromODK
         TestOKEnabled()
     End Sub
 
-    Private Sub AllControls_ContentsChnaged(ucrChangedControl As ucrCore) Handles ucrInputChooseForm.ControlContentsChanged, ucrInputUsername.ControlContentsChanged ', ucrInputPassword.ControlContentsChanged 
+    Private Sub ucrInputUsername_ContentsChanged(ucrChangedControl As ucrCore) Handles ucrInputUsername.ControlContentsChanged
+        If ucrInputUsername.IsEmpty() Then
+            cmdFindForms.Enabled = False
+        Else
+            cmdFindForms.Enabled = True
+        End If
+        TestOKEnabled()
+    End Sub
+
+    Private Sub ucrInputChooseForm_ContentsChanged(ucrChangedControl As ucrCore) Handles ucrInputChooseForm.ControlContentsChanged, ucrInputUsername.ControlContentsChanged ', ucrInputPassword.ControlContentsChanged 
         TestOKEnabled()
     End Sub
 
--- a/instat/dlgImportFromODK.vb
+++ b/instat/dlgImportFromODK.vb
@@ -81,8 +81,10 @@ Public Class dlgImportFromODK
     Private Sub pnlPlatform_ControlContentsChanged() Handles ucrPnlPlatform.ControlValueChanged
         If rdoOna.Checked Then
             clsGetFormsFunction.AddParameter("platform", Chr(34) & "ona" & Chr(34))
+            ucrInputChooseForm.SetText("")
         ElseIf rdoKobo.Checked Then
             clsGetFormsFunction.AddParameter("platform", Chr(34) & "kobo" & Chr(34))
+            ucrInputChooseForm.SetText("")
         End If
     End Sub
 
--- a/instat/dlgImportFromODK.vb
+++ b/instat/dlgImportFromODK.vb
@@ -32,27 +32,17 @@ Public Class dlgImportFromODK
         TestOKEnabled()
     End Sub
 
-    Private Sub SetRCodeForControls(bReset As Boolean)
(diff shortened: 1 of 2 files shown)


instat/dlgImportFromODK.vb
```diff
@@ -93,7 +95,16 @@ Public Class dlgImportFromODK
...
 
-    Private Sub AllControls_ContentsChnaged(ucrChangedControl As ucrCore) Handles ucrInputChooseForm.ControlContentsChanged, ucrInputUsername.ControlContentsChanged ', ucrInputPassword.ControlContentsChanged 
+    Private Sub ucrInputUsername_ContentsChanged(ucrChangedControl As ucrCore) Handles ucrInputUsername.ControlContentsChanged
+        If ucrInputUsername.IsEmpty() Then
+            cmdFindForms.Enabled = False
+        Else
+            cmdFindForms.Enabled = True
+        End If
+        TestOKEnabled()
+    End Sub
```
lilyclements: This is a sub I added in. I think "find forms" should be disabled when ucrInputUsername is empty.

1. Doesn't make sense to find a form for something which doesn't have a username...
2. Clicking "Find Forms" automatically brings up a pop-up dialog where the user inputs their password. They can only input their password if their username has been inputted and so if they click this before entering a name, it brings up a bug.

I can't see any reasons why Find Forms should be enabled when Input Username is empty. Hopefully I haven't missed anything obvious! Let me know if I have.

stevekogo:
@lilyclements I am going to review this in a short while.

instat/dlgImportFromODK.vb
stevekogo: I agree with you on how this should behave. in addition, we may need to impose further restrictions such that having an empty space as user name does not enable the button.

stevekogo:
What should happen if you change from change from one platform to another? Should you keep the contents of the username and choose form combobox? I think you should not.

lilyclements:
Change between KOBO and ONA? I think the user might have KOBO automatically selected, input their username and then realise they meant to select ONA. It could be frustrating if they then had to reinput everything?

stevekogo:
We may decide to keep the username but what about the contents of the combo box?

lilyclements:
Yes, that should probably clear to avoid an error. I'll sort that now.

lilyclements:
@stevekogo I've done this. What other changes should be made?

stevekogo:
@lilyclements I will review this in a short while.

stevekogo:
@lilyclements So far, what you have implemented seems sensible but I have one more issue with the way this dialog works. 
Upon inputing the user name an successfully loging in and finding forms, the forms are listed in the combobox. If you then change the username to something different, the forms listed in the combobox stays there and OK is enabled. If you press OK, then you get an error due to inconsistencies.
What is your view on how to address this?

lilyclements:
I thought I implemented so that the ucrCombo clears when you change radio buttons? Are you using the most recent version?
I'll check when I'm in the office.

stevekogo:
This is a different concern which I thought of when testing the dialog. Clearing the combo box when move from one radio button to another works fine.

lilyclements:
I understand the problem now.

So how about when the username input text changes, the find forms clearers?

I don't want the combo box to clear when the username is empty as the user might not make it empty to change the username. However, when the user changes the username we can clear the combo box.

I didn't like this idea at first but I think it makes sense. 
The order you run this dialog has to go input username then find forms. 
The user wouldn't ever find forms and then change the username and then feel annoyed that their forms are lost. The form they want isn't visible until they're on the right username.

I want to see @dannyparsons view on this before implementation

dannyparsons:
Once the username has been changed in any way then the forms should clear because they are no longer valid and you need to click Find Forms again. So happy for this to be added.

instat/dlgImportFromODK.vb
```diff
@@ -103,6 +105,8 @@ Public Class dlgImportFromODK
         Else
             cmdFindForms.Enabled = True
         End If
+        ucrInputChooseForm.cboInput.Items.Clear()
+        ucrInputChooseForm.SetText("")
```
lilyclements: If the username changes, the ucrInput clears and the item selected with it does too

instat/dlgImportFromODK.vb
```diff
@@ -81,9 +81,11 @@ Public Class dlgImportFromODK
     Private Sub pnlPlatform_ControlContentsChanged() Handles ucrPnlPlatform.ControlValueChanged
         If rdoOna.Checked Then
             clsGetFormsFunction.AddParameter("platform", Chr(34) & "ona" & Chr(34))
+            ucrInputChooseForm.cboInput.Items.Clear()
```
lilyclements: I realised, the item chosen was clearing but not the actual cbo list so I could still access them. I have sorted this here

lilyclements:
@stevekogo This is ready for a re-review

stevekogo:
This is working well. @dannyparsons can have a final look then merge it in.

PR #2956, opened 2017-04-24
View factor labels/Levels

stevenndungu:


stevenndungu:
I have added most of the options I felt are very necessary to be part of the dialogue as we had discussed with @lilyclements . Suggestions of more parameters that seem important to this dialogue are welcome. This is the current design look...
image

stevenndungu:
I have added most of the options I felt are very necessary to be part of the dialogue as we had discussed with @lilyclements . Suggestions of more parameters that seem important to this dialogue are welcome. This is the current design look...
image

stevenndungu:
The checked check boxes are checked on default. others are unchecked on default

instat/dlgFactorLabels.vb
```diff
@@ -33,19 +35,83 @@ Public Class dlgViewFactorLabels
...
+        'Defining the function
+        clsViewFunc.SetPackageName("sjPlot")
+        clsViewFunc.SetRCommand("view_df")
+        clsViewFunc.AddParameter("show.id", "FALSE")
+        ucrBase.clsRsyntax.SetBaseRFunction(clsViewFunc)
     End Sub
 
     Private Sub TestOkEnabled()
-
+        If (Not (ucrReceiverFactorColumns.IsEmpty) AndAlso (ucrChkAlternateColour.Checked OrElse ucrChkShowFrequencies.Checked OrElse ucrChkShowId.Checked OrElse ucrChkShowLabels.Checked OrElse ucrChkShowMissingValues.Checked OrElse ucrChkShowPercentage.Checked OrElse ucrChkShowType.Checked OrElse ucrChkShowValues.Checked OrElse ucrChkSortByName.Checked)) Then
```
lilyclements: Do we want this? Or do we want just for ucrReceiverColumns to be full? It runs if I don't check anything, it's just not a very helpful run!

stevenndungu: The line says that Ok should be disabled when no checkbox is checked and I saw that happening...let me confirm then get back to you

stevenndungu:
Nop, let me check that as well

stevenndungu:
@lilyclements Issue of output given twice is an interesting one. I have  identified where the problem is, though sounds quite tricky to solve...

lilyclements:
Hmm okay. I'm intrigued, why is it occurring?

stevenndungu:
@africanmathsinitiative/developers  I thought of fixing the issue of double output by using select function in dplyr instead. So that instead of 
` sjPlot::view_df(x=InstatDataObject$get_columns_from_data(col_names=c("Field","Variety.","Fertgrp."), data_name="survey"),show.type=TRUE, show.na=TRUE, show.prc=TRUE, show.frq=TRUE, sort.by.name=TRUE)`
I have;
` sjPlot::view_df(x=dplyr::select(.data=survey_temp, c(Village.,Variety.,Fertgrp.)),show.type=TRUE, show.na=TRUE, show.prc=TRUE, show.frq=TRUE, sort.by.name=TRUE)`
Because essentially we want the selected variables to of class `data.frame` when extracted which can be easily achieved by `dplyr select` function or `[].` function in R. 
@lilyclements @dannyparsons I dont know what your views are on this.

stevenndungu:
Othws it is ready for review

dannyparsons:
I prefer we don't use select to be consistent with other functions but since this is an issue with the view_df function then I don't mind that we do it this way for now. We should try to determine the cause of the bug as well so we can report it.

lilyclements:
I've made some minor changes waiting to be merged, explained in that pull request why these changes were made.
After you've merged that in I think this should be ready for Danny.

@dannyparsons out of interest, why don't you like using the pipe? Is it because of how it looks in the code output or is there more to it?

dannyparsons:
I actually prefer how the code looks with pipes (and we could consider using them more) but I don't like the inconsistency that we only select columns with pipes in this dialog. It will be confusing to users and they might think this is doing something different to the other way of doing the code. So I just want us to be consistent. But here we don't have an alternative to the pipes.

stevenndungu:
Yeah, I understand. I arrived at the alternative way after the normal way failed after much trial.

lilyclements:
@dannyparsons Okay thank you for explaining. This is ready for you to review

instat/dlgFactorLabels.vb
```diff
@@ -33,19 +35,103 @@ Public Class dlgViewFactorLabels
     End Sub
 
     Private Sub InitialiseDialog()
+        'TODO: HElP ID
+
+        ucrReceiverFactorColumns.SetParameter(New RParameter("x", 1))
+        ucrReceiverFactorColumns.SetParameterIsString()
+        ucrReceiverFactorColumns.SetParameterIncludeArgumentName(False)
+        ucrReceiverFactorColumns.bWithQuotes = False
+        ucrReceiverFactorColumns.SetIncludedDataTypes({"factor", "ordered"})
```
dannyparsons: When you do "factor" it will include ordered factors as well so I don't think you need "ordered" here.

instat/dlgFactorLabels.vb
dannyparsons: If you want to avoid a blank table then some of these should be removed. For example, Alternative colours I don't see why OK enabled should depend on this. I would understand more if it was just the checkboxes that relate to what is displayed in the table.

instat/dlgFactorLabels.vb
stevenndungu: Yes, I agree with you.

instat/dlgFactorLabels.vb
stevenndungu: Well, having at least  the main three checkboxes checked for the OK to be enabled is quite Okay. Making that adjustment.

stevenndungu:
@africanmathsinitiative/developers  this is ready for re-review after several changes.
 Note: The design currently looks like this;
image

While testing use the `efc` data set in R-Instat. It is a dataset in `sjmsc` package in R, to clearly see how it works.

lilyclements:
@stevenndungu I will review this now. I really like the group boxes.
While I will try it with the efc data set as I've been looking at that in R with this sort of code, I will also try with other data sets as it's good to try different data on dialogs

lilyclements:

PR #2958, opened 2017-04-24
converted columns retain variable metadata by default

dannyparsons:


instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -775,8 +775,8 @@ instat_object$set("public", "rename_dataframe", function(data_name, new_value =
 } 
 )
 
-instat_object$set("public", "convert_column_to_type", function(data_name, col_names = c(), to_type, factor_numeric = "by_levels", set_digits, set_decimals = FALSE) {
-  self$get_data_objects(data_name)$convert_column_to_type(col_names = col_names, to_type = to_type, factor_numeric = factor_numeric, set_digits = set_digits,set_decimals = set_decimals)
+instat_object$set("public", "convert_column_to_type", function(data_name, col_names = c(), to_type, factor_numeric = "by_levels", set_digits, set_decimals = FALSE, keep_attr = TRUE) {
```
dannyparsons: @africanmathsinitiative/developers Here you can see I've added a parameter to the convert method which allows us to choose whether to keep attributes (metadata) of a column when we convert it. The default is TRUE. Can someone add this as a check box on the convert dialog? The dialog default can be TRUE as well.

lilyclements: @dannyparsons I'll do this

rdstern:
I have seen that this seems to work by default.  So I can now take the SPSS imported data and make  a column into a factor.  It then is OK (sort of) for the commands that include the value labels.  But as a factor it is still just keeping the numeric values.  How could our own commands choose the real labels instead?  Happy to discuss.

PR #2963, opened 2017-04-25
Keys and links

Lunalo:
I think we will work with @dannyparsons  when he comes on Thursday . We need to split the ouput into two columns


dannyparsons:
I've just remembered we do something similar in the RLink. If you look on line 555 it gets a list from R - and then uses the names and values separately.
You could look at that, otherwise we will do it when I arrive.

Lunalo:
@dannyparsons 

Great, I will look at it.
Thanks. I can see from the code is listing the elements vertically .

Lunalo:

This is now ready for review. 

How it works:

The dialog has two dataframe selectors for to and from dataframes. On load if the to dataframe has keys then they will be populated in the listbox in two-column; first column having the name of the key and the second one having respective key columns.
If from dataframe has columns with the same name with any of the columns in the listbox then the back color of the listbox turns to green indicating the link can be created  else the back color turns  red. 

If you select a key in the listbox then it is populated in the selected key ucrInput and this is overriden in case  another key is selected.

Link name ucrInput is used to specify the name of the link.

dannyparsons:
Thanks for this useful description John. Just to add that there will also be a subdialog for specifying a link when the matching columns in the two data frames have different names. Currently the subdialog should be disabled as it is not yet implemented.

For testing @africanmathsinitiative/developers could start with daily climatic data from the library. Then use the column summaries dialog to summarise by say, Year and Month. You will now have two data frames which are linked by the Year and Month columns (link automatically created by the dialog). You could then export these two data frame separately e.g. as csv. Then in a new session import them back into R-Instat. Of course they will now not be linked and you can test the dialog by trying to create the link (after first defining Year and Month as a key in the second data frame).

To check that a link has really been created you can look at the data frame and column metadata after running the dialog to see if appropriate metadata has been added or use the view link dialog if it is functioning.

stevenndungu:
@Lunalo Great, reviewing this shortly.

lilyclements:
I can't do a thorough review right now, so I am making minor changes and will send that to you shortly

stevekogo:
This is working fine despite a few concerns. First, if you run the code from this create key dialog in R with intention to create a new key same columns already used to create a key, you get the following warning *A key with these columns already exists. No action will be taken.
A link with these columns already exists. A new link will not be added.* Could this be captured in the dialog and probably disable Ok?

dannyparsons:
Good point. Maybe the easiest thing to do now is to change the warning into an error so that the message appears to the user. Not ideal because we should try to avoid error boxes but at least it tells the user what is happening.

Lunalo:
yes @dannyparsons 

 We were thinking of using try catch statement but i can just change the warning into error in r for now.

Lunalo:
This is now fixed

dannyparsons:
Sorry I've just realised this will give us problems in other dialogs when we try to add keys automatically, we don't want to get an error. Can you change it back to a warning. And then write this as an issue we need to work on.
@stevekogo any other issues with these?

lilyclements:
@dannyparsons I would like to make edits to this before you merge/review it, so once @stevenndungu gives the all-clear on this branch I'll make my changes!

Lunalo:
@lilyclements 
have you finished making edits?

PR #2966, opened 2017-04-25
Three way frequencies

AlexSananka:
@africanmathsinitiative/developers  This is in progress


AlexSananka:
@dannyparsons you can merge in this for now

instat/dlgThreeVariableFrequencies.vb
```diff
@@ -1,3 +1,260 @@
...
+        ucrPnlFrequencyDisplay.AddRadioButton(rdoGraph)
+
+        ucrChkWeights.Enabled = False
+        cmdOptions.Enabled = False
+
+        ucrPnlFrequencyDisplay.AddFunctionNamesCondition(rdoTable, "sjtab")
+        ucrPnlFrequencyDisplay.AddFunctionNamesCondition(rdoGraph, "sjplot")
+
+        ucrPnlFreqType.AddParameterValuesCondition(rdoCount, "fun", "grpfrq")
+        ucrPnlFreqType.AddParameterValuesCondition(rdoCount, "margin", False)
```
dannyparsons: This doesn't quite look right. What is the condition you wanted here?

AlexSananka: Yes sure you're right,I should have edited this.At first what I wanted to achieve here is like telling the code that margin is not associated with count,But I think it is already by default.I think we're doing this  when we want extra bars for either column,rows and cells.
I will delete this.

dannyparsons: I think what you want is a parameter present condition. This is when the condition is about whether the parameter is in the code and not about the value. This is what @stevenndungu did for two variable dialog.

AlexSananka:
@dannyparsons This is ready

PR #2977, opened 2017-04-27
Updatingdialogs

Lunalo:
This is still work in progress


Lunalo:
@dannyparsons 

This is ready. I was updating the r code from the previous version

PR #2980, opened 2017-04-27
DlgView graph

stevenndungu:

### DIFF
--- a/instat/dlgViewGraph.Designer.vb
+++ b/instat/dlgViewGraph.Designer.vb
@@ -1,9 +1,9 @@
-﻿<Global.Microsoft.VisualBasic.CompilerServices.DesignerGenerated()> _
+﻿<Global.Microsoft.VisualBasic.CompilerServices.DesignerGenerated()>
 Partial Class dlgViewGraph
     Inherits System.Windows.Forms.Form
 
     'Form overrides dispose to clean up the component list.
-    <System.Diagnostics.DebuggerNonUserCode()> _
+    <System.Diagnostics.DebuggerNonUserCode()>
     Protected Overrides Sub Dispose(ByVal disposing As Boolean)
         Try
             If disposing AndAlso components IsNot Nothing Then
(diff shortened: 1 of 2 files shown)


maxwellfundi:
@stevekogo I this is working well but just a few comments on the options for viewing the graph. Neither is the R Viewer nor the Output window are interactive. Do we need these options here if we can do the same on use graph dialog?

lilyclements:
I think these are appropriate here as this is a View Graph dialog. However, I could be the only person in the world who feels this way! Perhaps for clarity we need to define the difference between View Graph dialog and the Use Graph dialog

View Graph:
This is simply viewing your graph in different ways. (Similar to the "View" dialog for data sets where you can view your data in different ways: Output Window, R-Window, HTML, etc)

Use Graph:
In "Use Graph", if the user does not click on any options - the graph views in the output window.
However, there is a command on this dialog "Plot Options". This command takes the user to the sub dialog on plot options where they can edit their graph. (Albeit this cmd is currently disabled which means this dialogs functionality is to just view the graph in the output window)
The user can also save their new edited graph
This graph always(?) views in the output window

So the Use Graph dialog is more for editing a graph whereas the View Graph dialog is for viewing the current graph. The view options aren't editing the graph.

However, perhaps it's the word "Use" that is a little ambiguous here?

rdstern:
I agree that Use Graph and View Graph will be different.  I am waiting (patiently???) for the options to be added to the Use graph.  Can't these be as soon as the "dialogue system" is updated.  For example we could use a graph (box plot) and add a layer, or change the theme.  And talking of themes, there is a big area here still.  

I keep being told I can't have my box plot labels vertical, because this is part of themes.  That would be sooooo nice to have soon!

lilyclements:
This is nearly ready for @dannyparsons - I just want to address the points I mentioned earlier

1. What should the default be?

2. How do we tackle the problem of interactive plots not working for ggplots or do we not worry about that too much? At least, not for now.

dannyparsons:
I'm happy with interactive as the default since I think this will be the most commonly used option.

We should have a way of getting either the saved graphs or only the saved ggplot graphs. I can look into this in the R code but we can still have the dialog in the meantime.

lilyclements:
Great sounds good. In which case, @dannyparsons this is ready to review

dannyparsons:
Three remaining issues here:
1. A blank R viewer displays when viewing interactive graph
2. The conditions for the panel are not yet correct but this may not be possible since some options only differ by the method of display and the code is identical
3. We need to be able to select just ggplot graphs (or disable interactive for non ggplot graphs)

PR #2986, opened 2017-04-28
Two sample t-test

stevekogo:


stevekogo:
This fixes a bug upon running two-sample t-test but I am yet to migrate the code to the new system.

maxwellfundi:
@africanmathsinitiative/developers could you lease find sometime and review this? This needs to go into todays version.

stevenndungu:
Let me have a look at it.

lilyclements:
I'll review it too, I'm going to make some minor changes while I do

instat/dlgRegressionSimple.vb
```diff
@@ -112,17 +112,19 @@ Public Class dlgRegressionSimple
 
     Private Sub SetTTest()
         clsRTTest.SetRCommand("t.test")
+        clsRTTest.SetPackageName("stats")
```
lilyclements: Okay, so there are two t.test functions in two packages - mosaic and stats.
This was written with the mosaic package in mind, which is why the parameter names were "mu", "x", etc
How come this was changed to the stats package?

stevenndungu:
Lily are getting errors when you run the dialog? Two sample radio button.. Errors like
image

stevenndungu:
followed by;
image

lilyclements:
Yes, but also an error here would occur anyway as the second factor needs two levels only. Is there an easy way to implement this @dannyparsons ?

I'm not sure if these errors are occurring because this is now using the `stats` package and it was made for the `mosaic` package. Something to explore after I hear from @stevekogo why the package was changed

stevenndungu:
I see the dialogue is good but needs one enhancement such that the factor variable is restricted to only accept factor variables that are STRICTLY of two levels

dannyparsons:
The issue of restricting to 2-level factors is something we can implement but haven't got the functionality for yet. If the function gives a sensible error for this then I suggest we don't need to implement this today.

stevekogo:
Alot still needs to be implemented here but can't be fixed for today's deadline. Someone can tidy this up on my behalf not necessarily keeping my changes as long as something works better. I on the way to Maseno and will be there in the evening.

dannyparsons:
now been fixed by #2996

PR #2988, opened 2017-04-28
Dlg Two Way freq

stevenndungu:
This are design changes made by Lily and change change of defined parameter names only @dannyparsons  this is ready for you


dannyparsons:
There are no changes here. Was it already merged from the other branch?

PR #2991, opened 2017-04-30
Three way frequencies

AlexSananka:
@africanmathsinitiative/developers  This is ready for review


lilyclements:
Also, how come you are doing

`ucrReceiverGroups.SetParameter`

in the SetDefaults out of interest?

rdstern:
In the dialogue I suggest 3 changes:
1) Put the Group By field below the row column fields. (The row and column fields are still the more important.)
2) Change the label for the **Row Factor** to **Row Variable (Factor)**.  Also **allow any type of data there**, not just factors, i.e. especially number columns, but also character, etc. (This change also to be made for the 2-way frequencies, to be consistent with the one-way frequencies.)
3) There could I think be a second Group by factor below the first.  Perhaps, instead the receiver can be changed to be a multiple receiver, thought there can't be more than 2 columns here.  The label could become "Group by Factor(s)".  Please discuss which way this should be done?
4) Also discuss whether the Group by column could be optional?  I know this is then not 3-way, but it could be that someone is here and then decided they don't want the Grouping factor.
5) I assume the options are the same as for the 2-way frequencies.
6) The menu item could be changed to become "Up to 4 Variables"

AlexSananka:
@rdstern We have discussed this with @dannyparsons 
1 and 2 makes sense and I am going to fix this.
3.We think introducing the fourth receiver is better because if we go for the multiple receiver we have to introduce some pop up message to the user if they try adding more than two variable, which we would prefer to avoid.
4 and 5. The options in the two way frequencies are similar, and this will use the same subdialog, but there are small differences which makes it a bit complicated to allow the dilaog to do both 2 and 3/4 variables.  Unless this is really needed we suggest keeping it as minimum 3 variables.
6. This point makes sense but if we change the menu item will it still make sense for the other items in the same menu?

AlexSananka:
@lilyclements This is ready for a re-review

lilyclements:
@AlexSananka This looks ready, however, I am waiting to hear back from @rdstern before I give the go-ahead to @dannyparsons in case his response means more changes to be made.

lilyclements:
1. My changes are waiting to be merged, so make sure you merge them before any future commits

2. There is an issue with the ucrSave on the sub dialog.
If you go to "Both" and then open the sub dialog to the graph tab and close it (don't have to click anything necessarily!) then press OK the following error occurs.

> Error in self$get_objects(data_name = data_name, object_name = graph_name,  : 
>   three_way_freq1not found.

AlexSananka:
@lilyclements  I just saw this bug when I said it's ready,am going to fix this
Thanks

AlexSananka:
@lilyclements this is fixed

lilyclements:
@AlexSananka I have made changes that are ready to be merged.

Also, the code does work if the two "Group By"'s are the same variable
However, if a "Group by" variable is the same as the Row Variable or Column Variable then we get an error:

> Error in .subset2(x, i, exact = exact) : subscript out of bounds

Should we disallow the user to input in the "Group by" a variable already in a previous receiver?

AlexSananka:
@lilyclements I have made the necessary changes,this is again ready for you to review.

lilyclements:
@dannyparsons This is ready for you to review

instat/dlgThreeVariableFrequencies.vb
```diff
@@ -243,10 +238,86 @@ Public Class dlgThreeVariableFrequencies
...
+            If ucrChkFlip.Checked Then
+                clsColumnParam.Position = 1
+                clsRowParam.Position = 2
+            Else
+                clsRowParam.Position = 1
+                clsColumnParam.Position = 2
+            End If
+            ucrReceiverRowFactor.SetParameter(clsRowParam)
+            ucrReceiverColumnFactor.SetParameter(clsColumnParam)
+        End If
```
dannyparsons: This is a repeat of above. Why do you have this twice?

instat/dlgThreeVariableFrequencies.vb
```diff
@@ -163,62 +156,62 @@ Public Class dlgThreeVariableFrequencies
...
-    'Still in progress
-    Private Sub cmdOptions_Click(sender As Object, e As EventArgs) Handles cmdOptions.Click
-        sdgTwoWayFrequencies.SetRFunction(clsSjTab, clsSjPlot, bResetSubdialog)
-        bResetSubdialog = False
-        sdgTwoWayFrequencies.ShowDialog()
-        TestOkEnabled()
-    End Sub
+    Public Sub SetRCodeForControls(bReset As Boolean)
+        Dim clsTempParamOne As RParameter
+        Dim clsTempParamTwo As RParameter
```
dannyparsons: These names don't have any meaning related to what they do. Could you rename them? At least TempParamX would be more informative.

lilyclements:
Can't believe I missed that. Clearly I need to be more thorough in my reviews from now on!

stevenndungu:
image
You are setting rcode for sjtab function when it does not pass `cord.flip` parameter

stevenndungu:
Great, this looks good. Working well. @dannyparsons  this is ready for you

AlexSananka:
@africanmathsinitiative/developers someone should test this.It is working fine.

instat/dlgTwoWayFrequencies.vb
```diff
@@ -128,7 +128,7 @@ Public Class dlgTwoWayFrequencies
         ucrPnlFreqDisplay.AddToLinkedControls(ucrSaveGraph, {rdoGraph, rdoBoth}, bNewLinkedHideIfParameterMissing:=True)
         ucrPnlFreqType.SetLinkedDisplayControl(grpFreqTypeGraph)
 
-        ucrSaveGraph.SetPrefix("Graph")
+        ucrSaveGraph.SetPrefix("Two_way_freq")
```
lilyclements: On three_way_freq, the first letter was not capitalised but on Two_way_freg, it is. Either they should both be capitialised or not.

lilyclements: Noticed that one_way_freq is not capitalised too. Perhaps it's easiest for you to make two_way_freq without a capital T

AlexSananka:
@lilyclements sorry I don't get the error in my branch,do you mind if we maybe have a skype and we compare if we're doing the same thing or could you try and update your branch and try again.

lilyclements:

PR #2999, opened 2017-05-02
dlgHistogram

maxwellfundi:


maxwellfundi:

lilyclements:
Review/editing this now, so make sure you've sent all your commits

lilyclements:
@dannyparsons @maxwellfundi I have found a way to make this dialog along with other graphic dialogs crash. I'm not sure if I should write this in an issue as I know it is a problem with at least one other dialog - Scatterplot. (I haven't tested all plot dialogs because this causes a crash and so I have to keep reopening R-Instat to re-test)


If you open the dialog and select "Multiple Variables" and open the "Plot-Name Options" sub (e.g. Scatterplot options - not "Options") and change anything on that (e.g. colour), then it will crash when you press OK.

On this histogram dialog, I noticed that these options were not enabled when "Multiple Variables" was selected. However, these sub options still run if I select "Single Variables", make the sub dialog changes, and then switch to "Multiple Variables" and press OK. Since it's running the sub dialog options still, it crashes.

maxwellfundi:
@lilyclements  I think you need to be careful when pull requesting on peoples branches after you make changes to avoid conflict. I have had a huge conflict here when I merged in your changes. 

Thanks.

dannyparsons:
@maxwellfundi you should also be looking at the changes from a pull request before you merge it in your branch.

maxwellfundi:
@dannyparsons yeah I should also be quite careful when merging things.

lilyclements:
I don't understand how that happened. I always check the files I change before committing them. I only changed
Scatterplot designer
Histogram designer
Histogram dialog

I understand that my branch was more updated than your one, but usually it doesn't come up with all these as new changes once they go into your branch.

How has this occurred?

The only way I can think to avoid this is if we make sure that everyones branches are up to date before they say it's ready to review?

maxwellfundi:
@lilyclements yeah, we could be updating branches. That will avoid these kind of things.

lilyclements:
When somethings ready for review, I tend to get to it asap. I know it's a pain to update your branch before asking for review, but I think this would help us to avoid this problem. Should I make an issue to let all developers know?

dannyparsons:
I don't think that was the issue here. Only Lily's branch needs to be up to date when a pull request is done into Maxwell's so not sure what caused this today.

dannyparsons:
I'm not sure exactly either. This is where doing your changes at the end might be safer because then you just do a pull request to the main repo without Maxwell needing to merge in changes from you.

lilyclements:
The downside to that is, when do you review it? Do you review then don't merge but instead tell me it's been reviewed then I only make layout edits (to the designer and code)? This would mean I no longer fix little bugs (which is fine by me!) and just point them out during a review like I used to.
This would be a lot quicker for me because I wouldnt edit a branch multiple times.
What do you think?

dannyparsons:
This shouldn't happen often and I expect you shouldn't have to edit a branch more than once. Let's just see if we get in this situation again and check the pull requests carefully.

maxwellfundi:
Oh my God, this has so many conflicts. I think i will have to close this and open a new pull request. Or @dannyparsons are they easy to fix from your side?

dannyparsons:
No sorry seems there are too many conflicts.

PR #3005, opened 2017-05-02
Bug Fixes on Dlg Copy Sheet and Dlg Rename

### LINKED ISSUE #2528: Copy Data Frame
When you right-click at the bottom of a data frame, then "Copy" is disabled.

Could it become enabled and go to the dialogue Prepare > Data Object > Copy Data Frame

It should automatically fill in the data frame that sent it there.

Could you also check that this is done by other right-clicks that call dialogues.  For example try with 2 library data sets, from ggplot2, namely diamonds and then economics.  When I right click on Rename it correctly went to the **Prepare > Data Object > Rename Data Sheet** and "Economics" is automatically filled in.  Good.

But  then I go back to the **diamonds sheet** and right-click again.  Lo and behold I go to the Prepare > Data Object > Rename Data Sheet with Economics still automatically filled in!

A related task, with both these dialogues is when you access them directly, i.e. through Prepare > Data Object > Rename Data Sheet or Copy Data Sheet
I suggest that the correct thing is for them to "know" the visible sheet and always start from there?  You may want to check this point.  It is what happens with the Filter dialogue, but most others return with what was completed before.  This might need some more general discussion - which relates perhaps to the new changes in dialogue structure.  Currently it seems inconsistent between dialogues.

stevenndungu:
fixes #2528 
This fixes the following issues;
Dlg Copy Sheet: 
1) Fixing issue 2528
2) The Newname InputTexBox was empty instead of having the current data frame loded with "_copy" extension
image

Dlg Rename Sheet;
Fixing this bug
image


instat/dlgCopySheet.vb
```diff
@@ -13,23 +13,31 @@
...
         End If
         SetRCodeForControls(bReset)
         bReset = False
         autoTranslate(Me)
+        TestOKEnabled()
+    End Sub
+
+    Private Sub ReopenDialogue()
+        'Reseting ucrDataFrame to ensure that it displays the current data frame on the grid 
+        ucrDataFrameCopySheets.Reset()
```
dannyparsons: This is quite a nice solution, although we don't want it to reset every time we open the dialog. For example, if I open this from the main menu, then go back to the gird and change data frame and open it again, it should reopen as I closed it, but it will now refresh to the current data frame because it's always resetting.

If you look at `dlgName`, which is on the column right click menu, you can see how what we want is achieved by storing `strSelectedDataFrame` in the dialog and also `bUseSelectedColumn` which says when it should auto fill on reopen. 

Then when this is called from the right click menu in `frmEditor`, first we do         `dlgName.SetCurrentColumn(SelectedColumnsAsArray()(0), grdCurrSheet.Name)`. So it only auto fills when it's opened from the right click menu.

dannyparsons: And I've just saw that what you've done here is what is currently done on rename data frame dialog. This is going to have similar issues. Could you update that one as well to how I've described?

dannyparsons:
While you are here, can you also change the titles of the dialog so they say "data frame" instead of "sheet"?

lilyclements:
@dannyparsons before this is merged but after @stevenndungu has addressed your points, I would like to make changes to this.

lilyclements:
My changes are waiting to be merged so do merge that in before making any commits!

Can you explain what you have fixed in the r code?

I noticed I can copy a data frame and name it the name of a current data frame opened
(e.g. I have Survey and Blocking opened, I can copy Blocking and name it Survey. Just for fun. And it copies over the old Survey).

However, I noticed in rename that I can't rename it the name of a current data frame opened which is good.

lilyclements:
Also, Danny mentioned earlier about renaming the dialogs:
Rename dlgCopySheet to dlgCopyDataFrame
Rename dlgRenameSheet to dlgRenameDataFrame

stevenndungu:
In the Rcode I have added in the label option  for the data frame to be copied. In addition, after discussing with Danny, we want the rename  data frame dialog to behave in the same manner as copy data frame dialogue is behaving. That is, the dialogue to be able to rename a data frame if provided a similar name to an existing data frame.
I remember that, Alex was editing the `vdproj` so I could not make any changes meanwhile

dannyparsons:
Sorry Steven, thinking about it again, we don't want to be able to copy a data frame with the same name as a current data frame. I think we need to re-edit the R code. I like the addition of the label though, let's keep that.

lilyclements:
@stevenndungu how is this going out of interest? Let me know when it's ready to re-review

stevenndungu:
This is ready for re-review now.

stevenndungu:
Note: I have edited `frmMain`, `frmEditor` and `vbproj` in this pull request

dannyparsons:
I'll have a look at this if no one else is since there's some issues I want to fix.

Issue #2976, opened 2017-04-27
Adding an extremes dialogue


rdstern:
There are still 2 new dialogues to add to **Climatic > Prepare > Specific Events**, which will become **Climatic > Prepare > Extremes** and **Climatic > Prepare > End of the Rains**.  This issue is about the **Climatic > Prepare > Extremes**.  This dialogue will take (usually) daily data and find the annual extremes.   Then there is a second dialogue, Climatic > Model > Extremes which will model the extremes we have calculated from the Prepare dialogue.

So the Climatic > Prepare > Extremes does exactly that - it prepares the extremes for fitting a model!
a) At the top of the dialogue there are 2 buttons, namely **Max/Min** and **Peaks**. 
b) The middle of the dialogue can be the same as the Climatic > Prepare > Transform dialogue.  There is no threshold button needed, but the others, namely 1)Station, 2) Date, 3) Year, 4) Day of Year, 5) Data are the same.
c) There are also From and To fields as in the Start of the rains dialogues.
d) Then there are two alternative blocks one with the label **Max/Min** and the other with the label **Peaks**.
e) In the Max/Min block there is a checkbox  - default checked - with the label Maxima.

f) The second block is labelled Peaks
g) The first field is labelled Threshold - any number can be entered into this field.  
h) The second is a checkbox Default checked - "Values above threshold"
i) The year is saved automatically.  Two more columns can be saved. The usual one is **Value**.  The second one has a checkbox, default checked, **Day Number**.

shadrackkibet:
I am working on this

stevekogo:
@shadrackkibet and I are suggesting we have a skype call to discuss this dialog. The main focus is on peaks above threshold where I think we can use filters or subset in order to address the issue.
@dannyparsons, can you call us anytime so that we can discuss this.

dannyparsons:
We could Skype after the meeting with the team, but yes that part should be done with filters

shadrackkibet:
@dannyparsons this can now be closed.

PR #3007, opened 2017-05-03
Export r objects and R WorkSpace

AlexSananka:
@dannyparsons  this is ready to be merged


AlexSananka:
@dannyparsons am done editing this,you can now merge?Or should @africanmathsinitiative/developers  have a quick look?

lilyclements:
I THink we already have an extremes dialog which is currently empty? Perhaps it isn't empty. You should check before this is merged so we can avoid having multiple dialogs?

AlexSananka:
@stevekogo and @shadrackkibet should confirm this,I think this a new dialog according to #2976

rdstern:
Just to clarify there are actually 2 different extremes dialogues - or there will be.  This is not the same dialogue in 2 places.  They have different functions.

In **Climatic > Prepare > Extremes** we prepare the data by taking the (usually) daily records and finding the (usually) annual extremes.

In **Climatic > Modelling > Extremes**  we fit extreme value models to the extremes we prepared earlier.  This uses the extRemes package in R.

We have an AIMS MSc student from AIMS Tanzania doing his essay in this area.

lilyclements:
Thank you for clarifying. I think eventually we should rename them as we have dlgExtremes and dlgClimaticExtremes which have names which aren't too informative about what the difference is. However, I don't think we need to do this now.
@dannyparsons this is ready to merge.

lilyclements:
@africanmathsinitiative/developers Am I the only one having a very weird situation where I have updated my branch but cannot find these dialogs?

I'm on the updated branch which has read in other changes. It's also in the Source > Repos > Instat > instat folder as we would expect.

I took a copy of Alex's branch and I couldn't find the dlgExportRWorkspace in the solution explorer. When I looked for where it's being run it's underlined in red and saying there's an error. However, I am still able to run this branch despite all this.

image

image

it's even letting me run the dialog.

Does anyone know whats happened? What is going on?!

dannyparsons:
Not too sure. I will try it too. Maybe delete the branch and start again?

AlexSananka:
@lilyclements Which branch are you using,I have two separate branches am working on this two dialogs on,for Export R workspace I have a branch by the same name as the dialog and that is it for Export R object.

So you could consider pulling my two branches and test the dialog separately.

lilyclements:
My branch I'm using is called "AlexExportRWorkspace"
I checked many times it pulled the right one and it has. It must have because it's running the dialog well and showing all you changes.

I'll delete it and pull it again. This is strange!

lilyclements:
Fixed it. So, I opened "dlgExportRWorkspace" on my master branch, and then changed back to AlexExportRWorkspace and I can now see the changes. I have no idea why this issue happened in the first place, but I've got it now at least!

AlexSananka:
Great

PR #3008, opened 2017-05-03
Climatic transform dialog

stevekogo:


stevekogo:
This is work in progress

stevekogo:
This is partially working with part of the dialog disabled. Reviews are welcome.

stevekogo:

stevekogo:
This dialog is still fragile. If need be, I recommend that it should not be merged into today's version unless it is necessary.

dannyparsons:
How fragile? Can it do something useful fairly easily yet and without completely crashing out of the software? If not then probably you're right.

stevekogo:
It does not crash the software but reset does not work properly and you need to manually change the new column name since it is not a prefix.

dannyparsons:
I think that is ok to at least put in for now as long as other non functioning parts are disabled.
If someone agrees with this after testing then I'll merge.

dannyparsons:
@stevenndungu agrees. There are big issues still but it is usable for someone to look at at least.

lilyclements:
@stevekogo Are you continuing work on this today? I would like to make design changes but would like to wait until you've got a more stable version.

PR #3012, opened 2017-05-03
Export r objects

AlexSananka:
This is in progress


AlexSananka:

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1347,6 +1347,11 @@ instat_object$set("public", "import_from_iri", function(download_from, data_file
 } 
 )
 
+instat_object$set("public", "export_objects", function(data_name,file,object_names) {
+  save(list=object_names,file=file)
```
dannyparsons: We need to do this differently as an RDS file. And actually we won't need a method for this because we can just extract the object and save it directly.

Instead of this you can use the saveRDS function and get the object instead of just the object name.

AlexSananka:

instat/dlgExportRObjects.vb
```diff
@@ -32,21 +34,72 @@ Public Class dlgExportRObjects
...
+        SetRCodeForControls(True)
+        TestOkEnabled()
     End Sub
 
+    Private Sub cmdBrowse_Click(sender As Object, e As EventArgs) Handles cmdBrowse.Click
+        Dim dlgSave As New SaveFileDialog
+
+        dlgSave.Title = "Export file dialog"
+        dlgSave.InitialDirectory = frmMain.clsInstatOptions.strWorkingDirectory
+        dlgSave.Filter = "Serialized r objects (*.rds)|*.rds"
```
lilyclements: Two points:

1. Does serialized have to be US English here? It worked for me as UK or US English, however, since this is a name I imagine it should remain as US English.

2. Do we need to put "Serialized r objects" at all? On other dialogs, such as dlgExportGraphAsImage, it would be more like `RDS (*.rds)|*.rds`?

AlexSananka: It doesn't matter what you put in there,I think that's more of just a comment.
And now that it doesn't matter what you put in I do not think it matters which English you would use.

So for me that's just a comment and it's fine the way it is at the moment.

lilyclements: I think it does matter, because it is visible to the user and we need consistency along the project.
On dlgExportGraphsAsImage, we see it as this:

image

Whereas on this dialog it is like so:
image

So I think we should have this as RDS not `Serialized r objects`

However, if we do that this as `Serialized r objects` then I think we need a capital R and O for "R Objects". Additionally, I'm not sure if Serialized should be American English or not. I suppose US English if it's a name.

instat/dlgExportRObjects.vb
```diff
@@ -32,21 +34,73 @@ Public Class dlgExportRObjects
...
+    Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
+        SetDefaults()
+        SetRCodeForControls(True)
+        TestOkEnabled()
     End Sub
 
+    Private Sub cmdBrowse_Click(sender As Object, e As EventArgs) Handles cmdBrowse.Click
+        Dim dlgSave As New SaveFileDialog
+
+        dlgSave.Title = "Export file dialog"
```
lilyclements: I think this should read "Export R Objects"

instat/dlgExportRObjects.vb
AlexSananka: I see,Thanks.
but in dlgExportToCPT we have `Text separated file`
so should this be changed as well? I am not sure which one I should use now, @dannyparsons give your thoughts here.

instat/dlgExportRObjects.vb
AlexSananka: sure Thanks

instat/dlgExportRObjects.vb
lilyclements: Good point. I don't really mind which is used on any of these dialogs, as long as it's consistent! We'll see what @dannyparsons thoughts are.

dannyparsons:
I looked at this with @maxwellfundi before and on something like MS Paint you see:
image
So a mixture. If it's well known like JPEG or PDF we use just that otherwise we can use a longer description. I'd be happy with a longer description like `Serialized R Objects` but doesn't it stand for `R Data Structure`? I thought it did but I can't find it anywhere now.

lilyclements:
R Data Structure makes much more sense and it rings a bell.
All I could find was this - https://biostar.usegalaxy.org/p/7748/ where someone writes "R data structure (RDS) file"

I did a quick look on google and couldn't get anywhere. Then tried rseek.org and had the same issue! Maybe no one knows what RDS stands for.

lilyclements:
I think it is Serialized R Objects - https://cran.r-project.org/web/packages/rio/vignettes/rio.html

AlexSananka:
This is useful but I have to change from` .RDS `to` .rds`
oops!!

dannyparsons:
Good find. Looks good now. Any other issues @lilyclements?

lilyclements:
This looks ready for you to review!

dannyparsons:
Looking good. @AlexSananka Can you also change the rds name in the other dialog that use this now that we've agreed on the format here?

AlexSananka:
yeah sure I am going to change this

PR #3021, opened 2017-05-05
New Histogram 

maxwellfundi:
@lilyclements @dannyparsons  this is the redone histogram with the disabling feature. This works well from last time.


instat/ucrVariablesAsFactor.vb
```diff
@@ -331,4 +331,20 @@ Public Class ucrVariablesAsFactor
...
+            ucrMultipleVariables.SetIncludedDataTypes(strInclude)
+        End If
+    End Sub
+
+    Public Overrides Sub SetExcludedDataTypes(strExclude As String())
+        If bSingleVariable Then
+            ucrSingleVariable.SetExcludedDataTypes(strExclude)
+        Else
+            ucrMultipleVariables.SetIncludedDataTypes(strExclude)
+        End If
```
dannyparsons: The error Lily gets is because you should set the types for both receivers. For this it doesn't matter which is the current one. They both should have the same types.

maxwellfundi:
@dannyparsons this last commit solves the problem of crashing when multiple receiver is the current receiver. @lilyclements this is now fixed.

dannyparsons:
Looks good @lilyclements any other issues here?

PR #3023, opened 2017-05-05
Ucr dist update control

Lunalo:
this still in progress I will update you on how it goes


dannyparsons:
We could now do this as a simpler way of getting multiple columns:
`replicate(3, rnorm(sd=1, mean=0, n=36))`
It should simplify the control code as well.

Lunalo:

I have managed to fix the resseting and and  unusual indexing of the parameters list.

This will be ready by tomorrow to be reviewd

Lunalo:

This is ready for the review, except the automatic resetting of distribution parameters on Reopening the dialog and the developer error that is occurring on load.

Lunalo:

This is now ready for review. I would like to start update model dialogs so this is argent

lilyclements:
Going to review and make designer amendments to this. Make sure you've sent all your commits!

instat/dlgRandomSample.vb
```diff
@@ -13,180 +13,140 @@
...
+        ucrNudNumberOfSamples.SetRCode(clsMultipleSamplesFunction, bReset)
     End Sub
 
     Private Sub TestOKEnabled()
-        If ucrDistWithParameters.bParametersFilled AndAlso nudNumberOfSamples.Text <> "" _
-            AndAlso nudNumberOfSamples.Text <> "" AndAlso (Not chkSetSeed.Checked OrElse (chkSetSeed.Checked AndAlso nudSeed.Text <> "")) _
-            AndAlso ((nudNumberOfSamples.Value = 1 AndAlso Not ucrNewColumnName.IsEmpty) OrElse (nudNumberOfSamples.Value <> 1 AndAlso Not ucrPrefixNewColumns.IsEmpty)) Then
+        If ucrDistWithParameters.bParametersFilled AndAlso ucrNudNumberOfSamples.GetText <> "" _
+            AndAlso (Not ucrChkSetSeed.Checked OrElse (ucrChkSetSeed.Checked AndAlso ucrNudSeed.GetText <> "")) _
+            AndAlso ((ucrNudNumberOfSamples.GetText = 1 AndAlso ucrSaveRandomSamples.IsComplete) OrElse (ucrNudNumberOfSamples.GetText <> 1 AndAlso ucrSaveRandomSamples.IsComplete)) Then
```
lilyclements: I don't understand this last part Can you explain it to me?
You're saying 

> if the nud = 1 and the save is complete or if the nud is not 1 and the save is complete.

Why can't you just say 

> if the save is complete

instat/dlgRandomSample.vb
```diff
@@ -13,180 +13,140 @@
...
-                ucrPrefixNewColumns.Visible = True
-                lblPrefixColumnName.Visible = True
-            End If
-            SetAssignTo()
-        End If
+    Private Sub SetRCodeforControls(bReset As Boolean)
+        ucrNudSeed.SetRCode(clsSetSeed, bReset)
+        ucrChkSetSeed.SetRCode(clsSetSeed, bReset)
+        ucrSaveRandomSamples.SetRCode(clsMultipleSamplesFunction, bReset)
+        ucrNudNumberOfSamples.SetRCode(clsMultipleSamplesFunction, bReset)
```
lilyclements: Why is the ucrDists not being called here?

instat/dlgRandomSample.vb
Lunalo: Thanks for spotting this. This was happening before I started using `replicate`

I forgot to adjust it. I will do it to fit in the current state

instat/dlgRandomSample.vb
Lunalo: From how I understood this: Setting of distributions and there parameters are done in the ucr and so there is no need of setting r code here in the dialog

Lunalo: and so all I need is to  do `ucrDistWithParameters.SetRDistributions()
        ucrDistWithParameters.SetParameters()`

lilyclements: Okay, so what I would expect if I had n = 36 and did the normal distribution is to get:

`rnorm(sd=1, mean=0, n=36)
`(_or at least that's what the old code would give me_)

But here I get
`replicate(expr=(n=36), n=1)`

and here it gives all my columns as "36" in that case.

Lunalo: We set rcode so that the controls can know which functions and parameters are associated with it but in this case for ucr distributions everything is done there and the only thing we need to do is to set rcode for external control not in the ucrDistributions

lilyclements: Okay I understand. However, the output that I am now getting is giving me an entire column all containing the same numbers (that is, 36 in this case, my data frame length)

Perhaps it's not an issue in the SetRCode, however, I think it is still an issue

Lunalo: What I did, I associated replicate with ucrNudSampleSize and by defualt I passed the daframelength value to n for the distributions

Lunalo: I get you. I think if the user sets the number of samples to be 2 he s/he expects two samples.
`replicate(expr=(n=36), n=1)` this is not achievable currently since  `expr` is getting the output of simulated values.

lilyclements: I understand what you're saying, however, I'm getting the rcode as 
`replicate(expr=(n=36), n=1)`
regardless of which distribution I choose so there is a bug here somewhere that needs to be resolved

Lunalo: I think you might have deleted something . I have asked @stevenndungu to get the branch and it is working fine for him

stevenndungu:
I have two comments to make in addition to Lily's review:

1. @Lunalo , I just noted there is an issue change of data frames. When you change data frames some controls  do not update their parameter values.
This should should be easy to fix....here is the error
image

2. @africanmathsinitiative/developers stacking of appended columns together. I do not understand why this happens at times, I have experienced this today for the second time after some time now while testing this dialogue. 
 While appending column variables at  times it creates empty columns and stacks the column values on top of an already existing column. Which is strange and does not look neat at all.  Dragging the columns side by side separates  the stacked columns but leaves the columns created empty.  
image

instat/dlgRandomSample.vb
dannyparsons: @Lunalo remember our group discussions on testing previously. If a tester finds an issue on your dialog, it is your responsibility to find out why they get that bug, not the other way around.
If two users seem to get different results from the same use of a dialog you need to follow up to understand why, not expect the tester to do that for you. This is so we make testing a quick process of identifying bugs and passing back to the author to fix. 
Your assumption shouldn't be that the tester has made a mistake, although it could be the cause, you should follow up with the tester to understand why they are getting a bug that you are not.

lilyclements: Yes, I think I may have deleted something. I'll look for that now.
I tend to outline all my changes made before you merge them, however, since I'm doing a 10-minute-review on something I haven't worked long-term on, it would be really helpful for you to check my changes before merging them back in!

Lunalo:
@lilyclements 

You can have a look at it now .
Thanks.

instat/dlgRandomSample.vb
```diff
@@ -150,4 +150,8 @@ Public Class dlgRandomSample
     Private Sub ucrSaveRandomSamples_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrSaveRandomSamples.ControlContentsChanged, ucrSelectorRandomSamples.ControlContentsChanged, ucrChkSetSeed.ControlContentsChanged, ucrNudSeed.ControlContentsChanged, ucrSampleSize.ControlContentsChanged, ucrDistWithParameters.ControlContentsChanged
         TestOKEnabled()
     End Sub
+
+    Private Sub ucrSelectorRandomSamples_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSelectorRandomSamples.ControlValueChanged
+        SetDataFrameAndDistributionParameters()
+    End Sub
```
lilyclements: You have two subs running `.ControlValueChanged` and both running the sub `SetDataFrameAndDistributionParameters()`.

Can you make this into one sub together? So, line 146 can become
```
    Private Sub ucrNudNumberOfSamples_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrNudNumberOfSamples.ControlValueChanged. ucrSelectorRandomSamples.ControlValueChanged
        SetNewColumName()
    End Sub
```

And then this sub (line 154-156) can be removed

lilyclements:
A couple of distributions can cause a bug, e.g.:

> Error in rhyper(k = 10, nn = 36, m = 144, n = 10, 0) : 
>   unused argument (0)

This "unused argument(0)" has occurred for me for the following:
Hypergeometric
Lognormal
Cauchy


Additionally, F doesn't have it's parameters labels come up. I don't think this was an existing problem but perhaps it was? Can you give it's parameter labels in the code.
image
_under Distribution Parameters, the parameter names do not come up_


Third of all, 



Also, on hypergeometric the text cuts off over the inputs. What should we do about this? Can we move the inputs for when hypergeometric is checked?
image

What should we do about these at this stage? Other than this it's okay.

lilyclements:
You should go through and test each distribution at it's default level. I have counted at least 3 different errors with at least 5 different distributions by just selecting the distribution and pressing OK.
I didn't go through all the distributions as there were too many errors so you need to check all the distributions.

Additionally, F doesn't have it's parameters labels come up. I don't think this was an existing problem but perhaps it was? Can you give it's parameter labels in the code.
image
_under Distribution Parameters, the parameter names do not come up_

Also, on hypergeometric the text cuts off over the inputs. What should we do about this? Can we move the inputs for when hypergeometric is checked?
image

dannyparsons:
I see a lot of these bugs were existing bugs with the control and distributions before you started updating this, but I think now's a good time to sort these out and get everything working well.

Lunalo:
@lilyclements 

This is now fixed

lilyclements:
@Lunalo you still need to add the distributions package names. This can fix the problem outlined in #3109 

Additionally, the gamma distributions are having issues finding the shape parameter.

Lunalo:
I have fixed the testOk 

I realized that even from the old version the user is allowed to type in the combobox containing distributions . I think we should prevent  this. 

otherwise @africanmathsinitiative/developers  this is ready for review

lilyclements:
This is looking good, nice job. I am struggling to break this, which is somewhat upsetting but probably ultimately good. I can get errors but that's through doing "silly" things going negative on a positive distribution or writing my name in the parameters for distributions. I can't seem to crash it so it's looking pretty good.

Lunalo:
@dannyparsons 

This is ready for review

PR #3029, opened 2017-05-08
OneWayFrequencies bug fix

AlexSananka:


shadrackkibet:
I am going to review this now.

AlexSananka:
Great

dannyparsons:
@lilyclements you were so quick, didn't even let @shadrackkibet do his review first!

shadrackkibet:
just noticed a bug when i was testing this dialog though it has been merged.
When you type into the ucrsave and then you try resetting and then check the checkbox for the ucrsave you find the prefix changes to "last_graph".why does this changes!

image

dannyparsons:
Thanks Shadrack, sorry I merged it too soon! I think this is a bug we found on another dialog with the save control with someone? I should look into this and see if it's something I should fix on the control itself. Thanks for the reminder and good spot to find this in testing.

AlexSananka:
No I can fix this,I only need to reset the ucrsave in the SetDefaults.

shadrackkibet:
I am also suggesting that instead of disabling the ucrsave you can just hide it totally as it is not needed for table and and "Both" cases.Is this possible?

dannyparsons:
That sounds good.

lilyclements:
If we hide it totally, I would also like the ucrChkCoordinates to hide as well. It would be weird for one thing to hide and the other to be disabled.

lilyclements:
@dannyparsons @shadrackkibet The place this has happened before is on a sub dialog for either the one way frequencies or the two way frequencies.

AlexSananka:
ucrSave reset is fixed now

dannyparsons:
great

stevenndungu:
@AlexSananka I have question: I thought the ucrSave control in the sub-dialogue performs the role of saving the graphs? I thought the bug would be solved at the sub-dialogue level of the graph option. 

I just noted that we now have two save controls on the same dialogue one at dialogue level and the other at the sub-dialogue level. Which I find confusing as they both work depending on what you check. We need to do away with one (preferring that we still retain the one on the sub-dialogue as it was before - the bug can be solved at sub-dialogue level i think)

AlexSananka:
@stevenndungu good point,I thought I had disabled the savegraph on the subdialog but I didn't.

I think we can disable it for now or even do away with it since the one on the main dialog solves our issues.I have no problem if the issue can be fixed on the subdialog anyway.

dannyparsons:
Thanks Steven, yes there should only be one. I actually quite like it on the main dialog to be consistent with other dialogs. So the one on subdialog could now be removed.

AlexSananka:
I just realized I did the disabling on the two way subdialog instead of the oneway.

stevenndungu:
Yap, that sounds good because of consistency since our idea of having it in the subdialogue was only because it was specific to the the graph option. Alex, I did not check on two way freq, deleting it is better to disabling..hehe

PR #3042, opened 2017-05-09
dlgSummaryBarOrPieChart

maxwellfundi:


stevenndungu:
started testing it and when I get to pie chat option with Y variable as `depth` and xvariable as `cut` it runs without stopping,,,what could be the problem? Am using diamond data set

lilyclements:
Diamonds is a long dataset, could it be that it is taking a while to compute? I'm trying it now and see what you mean. That is strange!

lilyclements:
It's still going! I'm now running
`ggplot(data = diamonds, mapping = aes(x =cut, y = depth))+coord_polar(theta = "y") + geom_bar(width = 1, stat = "identity")`
in R, and this is taking a while too. I think it just must be because it's such a long dataset?

Strangely, if I run: `system.time(ggplot(data = diamonds, mapping = aes(x =cut, y = depth))+coord_polar(theta = "y") + geom_bar(width = 1, stat = "identity"))`
Then I get
```
   user  system elapsed 
   0.01    0.00    0.01 
```
So I don't understand how this is taking so long in R

stevenndungu:
Yes it takes pretty long time to run that.. though am told its because that is not how exactly the dialog should work. We need to have a summarised table like of data before preceding to the dialog.

If the user understands what they want to achieve with the dialog, then it works pretty fine but am concerned if this is not the case! We need to think about this as well.

stevenndungu:
Just one thing to change then its ready, have prefixes of "Bar", "Save Graph" and "Pie" in lower case to be consistent with other dialogues as well as "last_graph" in the  same graph

dannyparsons:
I think it's important for everyone to understand the expected outcome from running the dialog in this way. For example, when should you use the summary vs. normal dialog for Pie/Bar? If you're using the summary dialog is it sensible to use the (unsummarised) diamonds data set? i.e. what is it trying to do with the columns given that takes so long? How are you expecting it to group the numeric variable by the factor? Is there an option for that? Why not? Possibly because the dialog isn't for this case.
Of course the long running is an issue anyway, but is this expected (because we've done something not very sensible) or is it an issue with the code/dialog. 
If you think about the one way frequency tables with a big continuous variable - the issue isn't that it takes long (because we should expect that with so many unique values) but it highlights that we should be able to break out of commands when we want to - however there's nothing wrong with the code of the dialog itself, it's doing exactly what the user asks which will always take a long time.

So the summary bar/pie is meant for summaried data i.e. where you have a column of unique values and a matching numeric value to plot as the section of the pie. For diamonds you could summarise depth by cut, possibly doing a sum or mean and then plot the summary columns in this dialog.
To plot the raw data you would use the original data but then you only give one factor column and it does counts. We then wouldn't expect this to take a long time.

The example Steven gave is interesting though. Should we be able to do that and specify the summary we want? Is it possible in ggplot? Is that what the stat parameter is for?

I think my general point is the need for understanding the dialog when testing i.e. is what I expect it to do what it will do? What are sensible and non sensible uses of the dialog? etc. are really important. It can take more time to understand this but ultimately will help you test better.

stevenndungu:
@dannyparsons thanks, this is ready for you

lilyclements:
@dannyparsons there are changes I have made waiting to be merged. This is not ready for review yet!

maxwellfundi:
@lilyclements Changes merged and now ready for @dannyparsons

instat/ucrCore.vb
```diff
@@ -180,7 +180,7 @@ Public Class ucrCore
                 End If
             End If
             If ucrControl.bLinkedAddRemoveParameter Then
-                ucrControl.AddOrRemoveParameter(bTemp)
+                ucrControl.AddOrRemoveParameter(bTemp AndAlso (ucrControl.CanAddParameter() OrElse Not ucrControl.bAddRemoveParameter))
```
dannyparsons: @africanmathsinitiative/developers we have made a change here to the core control. This potentially effects any linked control which adds/removes another control's parameter. We needed this fix for this dialog and expect it won't harm other dialogs but be on the lookout for linked controls where maybe the parameter is now not being added correctly. This could be the cause in case it occurs.

lilyclements:
@maxwellfundi I liked @stevenndungu's suggestion of the prefixes being small case. Was that implemented in the end? If not, I'll quickly run a pull request and sort that now.

maxwellfundi:
@lilyclements  I had not changed that yet. But its okay for today's release and will fix those for future releases.

PR #3044, opened 2017-05-09
regular sequence fix

AlexSananka:
@lilyclements  This is ready,you can maybe test it and it can be merged.


AlexSananka:
@dannyparsons  I think this is ready for you since I just added the else statement.

PR #3050, opened 2017-05-10
Changing location of frmSetupLoading

maxwellfundi:


maxwellfundi:
@dannyparsons this is now moved appropriately to a higher position.

dannyparsons:
This isn't a great solution because it's a fixed position that will show differently on different machines but for now it's an improvement.

maxwellfundi:
Okay we will have a better lasting solution for this issue.

PR #3053, opened 2017-05-10
Dlgstringdist Enhancement

shadrackkibet:
In progress


shadrackkibet:
@africanmathsinitiative/developers This is making good progress.

lilyclements:
@shadrackkibet does that mean this is ready to review?

shadrackkibet:
@lilyclements this is not yet ready.

shadrackkibet:

instat/dlgStringDistance.vb
```diff
@@ -34,30 +34,46 @@ Public Class dlgStringDistance
...
         ucrInputPatternStringDistance.SetParameter(New RParameter("b", 1))
+        ucrInputPatternStringDistance.SetLinkedDisplayControl(lblPattern)
+
+        ucrPnlStringDist.AddParameterValuesCondition(rdoColumn, "b", ucrReceiverColumn.GetVariableNames.Trim(Chr(34)))
+        ucrPnlStringDist.AddParameterValuesCondition(rdoString, "b", Chr(34) & ucrInputPatternStringDistance.GetText & Chr(34))
+        ucrPnlStringDist.AddToLinkedControls(ucrReceiverColumn, {rdoColumn}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrPnlStringDist.AddToLinkedControls(ucrInputPatternStringDistance, {rdoString}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
 
         ucrInputComboBoxMethod.SetParameter(New RParameter("method", 2))
+        ucrInputComboBoxMethod.cboInput.DropDownStyle = ComboBoxStyle.DropDownList
```
lilyclements: This is interesting,  but I don't think this is how we eventually want a drop down list feature to work? @dannyparsons can probably explain this a lot better than me?

shadrackkibet: In this case i wanted the user not to write into the comboBox  because i noticed that when the user writes into the comboBox TeskOk is enabled and this will results into an error when the user clicks Ok. @dannyparsons  what do you think?

dannyparsons: This is definitely a sensible idea for this combo box. Good idea. It's something we've discussed before.

My issue is just on doing this generally across the software. This will currently be the only dialog with this kind of combo box which is not consistent. However, if we want to make this change, we have to start somewhere. There is some general work needed on combo boxes (to avoid the error you mention) and also to have combo boxes which can be typed in but also only allow specific values. There will also be some combo boxes which allow typing and can be any value (even one not in the list).
This could be the first example, or a third example of a combo box we cannot type in.
If we are happy with these 3 different types then I think your choice is right for this combo box.

My other issue is that this should be implemented as a method in the combo box control and not by accessing the combo box directly - so that any changes could be done in the method and not needed on every dialog. I wouldn't expect you to have known this, so I can work on your branch to show you what I mean by this.

A quick check with @rdstern and @volloholic would be useful before I proceed.

shadrackkibet:
@lilyclements  this is now ready for re-review after danny worked on the comboBox issue.

lilyclements:
@shadrackkibet I've made very minor edits. Let me know once these are merged then it should be ready for Danny.

lilyclements:

shadrackkibet:
I would like this to be merged quickly so that i can access the added combobox  functionality on use date dialog , there is a comboBox that behave similarly to this on this dialog.

instat/dlgStringDistance.vb
```diff
@@ -33,31 +33,47 @@ Public Class dlgStringDistance
...
+        ucrReceiverColumn.SetLinkedDisplayControl(lblColumnString)
+
         ucrInputPatternStringDistance.SetParameter(New RParameter("b", 1))
+        ucrInputPatternStringDistance.SetLinkedDisplayControl(lblPattern)
+
+        ucrPnlStringDist.AddRadioButton(rdoString)
+        ucrPnlStringDist.AddRadioButton(rdoColumn)
+
+        ucrPnlStringDist.AddParameterValuesCondition(rdoColumn, "b", ucrReceiverColumn.GetVariableNames.Trim(Chr(34)))
+        ucrPnlStringDist.AddParameterValuesCondition(rdoString, "b", Chr(34) & ucrInputPatternStringDistance.GetText & Chr(34))
```
dannyparsons: These can't be the correct conditions here because you are using values from the controls, and I expect these controls are empty in Initialise. I believe what you want is `AddParameterIsStringCondition` and `AddParameterIsRFunctionCondition` conditions to check what type of value a parameter has.

instat/dlgStringDistance.vb
```diff
@@ -67,14 +83,17 @@ Public Class dlgStringDistance
...
         ucrSelectorStringDistance.Reset()
         ucrSaveStringDistance.Reset()
         ucrInputComboBoxMethod.Reset()
-        ucrInputPatternStringDistance.Reset()
+
         clsStringDistFunction.SetPackageName("stringdist")
         clsStringDistFunction.SetRCommand("stringdist")
+        ucrInputPatternStringDistance.Reset()
         clsStringDistFunction.AddParameter("method", Chr(34) & "osa" & Chr(34))
+        clsStringDistFunction.AddParameter("b", ucrInputPatternStringDistance.GetText)
```
dannyparsons: Above you set the text of this control to empty, so this is just adding `b = ""`, if that's what you want as the default then fine, but then just replace this with `Chr(34) & Chr(34)`.

dannyparsons:
It would be great to update the Use Date dialog, as I mentioned before, if someone's going to edit this I'd like them to consider a redesign of the dialog so it's properly functional.

instat/dlgStringDistance.vb
shadrackkibet: will sort out this now.

instat/dlgStringDistance.vb
shadrackkibet: will also sort out this  now.

shadrackkibet:
@dannyparsons . I have sorted out the issues above this is ready for you again.

instat/dlgStringDistance.vb
```diff
@@ -53,8 +53,8 @@ Public Class dlgStringDistance
         ucrPnlStringDist.AddRadioButton(rdoString)
         ucrPnlStringDist.AddRadioButton(rdoColumn)
 
-        ucrPnlStringDist.AddParameterValuesCondition(rdoColumn, "b", ucrReceiverColumn.GetVariableNames.Trim(Chr(34)))
-        ucrPnlStringDist.AddParameterValuesCondition(rdoString, "b", Chr(34) & ucrInputPatternStringDistance.GetText & Chr(34))
+        ucrPnlStringDist.AddParameterIsStringCondition(rdoColumn, "b")
+        ucrPnlStringDist.AddParameterIsRFunctionCondition(rdoString, "b")
```
dannyparsons: Are you sure these are the correct way round?

shadrackkibet: Sorry, I have interchanged the conditions.

instat/dlgStringDistance.vb
```diff
@@ -53,8 +53,8 @@ Public Class dlgStringDistance
         ucrPnlStringDist.AddRadioButton(rdoString)
         ucrPnlStringDist.AddRadioButton(rdoColumn)
 
-        ucrPnlStringDist.AddParameterIsStringCondition(rdoColumn, "b")
-        ucrPnlStringDist.AddParameterIsRFunctionCondition(rdoString, "b")
+        ucrPnlStringDist.AddParameterIsRFunctionCondition(rdoColumn, "b")
+        ucrPnlStringDist.AddParameterIsStringCondition(rdoString, "b")
```
shadrackkibet: @dannyparsons I think this are now the right conditions for the respective controls.

PR #3057, opened 2017-05-10
Bar from Summary - bug fix

lilyclements:


instat/dlgSummaryBarOrPieChart.vb
```diff
@@ -175,19 +175,21 @@ Public Class dlgSummaryBarOrPieChart
...
+            If ucrSaveSummaryBar.bUserTyped = False Then
+                ucrSaveSummaryBar.SetPrefix("bar")
+            End If
+        ElseIf rdoPieChart.Checked Then
+            cmdPieChartOptions.Visible = True
+            cmdBarChartOptions.Visible = False
+            clsRgeomBarFunction.AddParameter("width", "1")
+            clsBaseOperator.AddParameter(clsRCoordPolarParam)
+            If ucrSaveSummaryBar.bUserTyped = False Then
+                ucrSaveSummaryBar.SetPrefix("pie")
```
lilyclements: Moved the `If ucrSaveSummaryBar.bUserTyped = False` to just be for the prefix. this was for everything before which meant that when you changed ucrSave name, the other things would not update!

dannyparsons: great but better to be `If Not ucrSaveSummaryBar.bUserTyped` in both places.

lilyclements: Yes I agree. I've fixed that now

PR #3059, opened 2017-05-10
Flat frequency tables

AlexSananka:
@africanmathsinitiative/developers This is making good progress.


AlexSananka:
@africanmathsinitiative/developers This is  ready for review

instat/dlgFlatFrequencyTable.vb
```diff
@@ -18,11 +18,14 @@ Imports instat.Translations
...
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
+    Private clsFtable, clsTable, clsAddMargin As New RFunction
     Private Sub dlgFlatFrequencyTable_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         autoTranslate(Me)
         If bFirstLoad Then
             InitialiseDialog()
             bFirstLoad = False
+        Else
+
```
lilyclements: Just noticed this - you haven't completed this Else. What did you want to put here?

instat/dlgFlatFrequencyTable.vb
```diff
@@ -34,18 +37,85 @@ Public Class dlgFlatFrequencyTable
...
-    Private Sub SetRCodeForControls(bReset As Boolean)
+    Private Sub ucrchkaddmargins_controlvaluechanged(ucrchangedcontrol As ucrCore) Handles ucrChkAddMargins.ControlValueChanged
+        If ucrChkAddMargins.Checked Then
+            clsFtable.RemoveParameterByName("table")
+            clsFtable.AddParameter("x", clsRFunctionParameter:=clsAddMargin)
+        Else
+            clsFtable.RemoveParameterByName("x")
+            clsFtable.AddParameter("table", clsRFunctionParameter:=clsTable)
+        End If
+    End Sub
```
lilyclements: How come this is written in the sub and not running in an initialise?

instat/dlgFlatFrequencyTable.vb
AlexSananka: This is unnecessary and can be deleted.

instat/dlgFlatFrequencyTable.vb
AlexSananka: This is because we want this to be happening when the receivers and the selector are not empty.

AlexSananka: Again we want this to happen at those specific events.

lilyclements:

instat/dlgFlatFrequencyTable.vb
```diff
@@ -33,19 +34,86 @@ Public Class dlgFlatFrequencyTable
...
     End Sub
 
+    Private Sub ucrChkAddMargins_ControlValueChanged(ucrchangedcontrol As ucrCore) Handles ucrChkAddMargins.ControlValueChanged
+        If ucrChkAddMargins.Checked Then
+            clsFTable.RemoveParameterByName("table")
+            clsFTable.AddParameter("x", clsRFunctionParameter:=clsAddMargin)
+        Else
+            clsFTable.RemoveParameterByName("x")
+            clsFTable.AddParameter("table", clsRFunctionParameter:=clsTable)
+        End If
```
dannyparsons: The names of these do not need to be different. In both cases you are passing in a table, one just has margins and other doesn't. The parameter could be either `x` or `table`, I suggest `x` since that's the named parameter in the function definition for `ftable`.

instat/dlgFlatFrequencyTable.vb
```diff
@@ -33,19 +34,86 @@ Public Class dlgFlatFrequencyTable
...
+    End Sub
+
+    Private Sub ucrRowReceiver_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrRowVariable.ControlValueChanged, ucrColumnVariable.ControlValueChanged, ucrSelectorDataFrame.ControlValueChanged
+        Dim clsGetVar As New RFunction
+        If Not ucrRowVariable.IsEmpty AndAlso Not ucrColumnVariable.IsEmpty Then
+            clsGetVar.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_columns_from_data")
+            clsGetVar.AddParameter("data_name", Chr(34) & ucrSelectorDataFrame.ucrAvailableDataFrames.cboAvailableDataFrames.SelectedItem & Chr(34))
+            clsGetVar.AddParameter("col_names", "c(" & ucrColumnVariable.GetVariableNames & "," & ucrRowVariable.GetVariableNames & ")")
+            clsTable.AddParameter("y", clsRFunctionParameter:=clsGetVar)
```
dannyparsons: I'm interested in this choice of parameter name. Is there a reason you chose `y`? Again, table accepts `...` as arguments so any name is allowed, I'm being quite picky, but we should try to pick something that makes sense. If I see an R function with parameter `y` I would expect there to be an `x` too. If there isn't a good reason for `y`, `x` seems like a sensible choice to me since it's a commonly used parameter name in R.

AlexSananka:
@dannyparsons all the above have been fixed.

instat/dlgFlatFrequencyTable.vb
```diff
@@ -95,11 +95,11 @@ Public Class dlgFlatFrequencyTable
 
     Private Sub ucrChkAddMargins_ControlValueChanged(ucrchangedcontrol As ucrCore) Handles ucrChkAddMargins.ControlValueChanged
         If ucrChkAddMargins.Checked Then
-            clsFTable.RemoveParameterByName("table")
+            clsFTable.RemoveParameterByName("x")
             clsFTable.AddParameter("x", clsRFunctionParameter:=clsAddMargin)
         Else
             clsFTable.RemoveParameterByName("x")
-            clsFTable.AddParameter("table", clsRFunctionParameter:=clsTable)
+            clsFTable.AddParameter("x", clsRFunctionParameter:=clsTable)
```
dannyparsons: Given that these parameters now have the same name, there is no need to remove the parameter before replacing it.

AlexSananka: Sure,I'll update this.

instat/dlgFlatFrequencyTable.vb
```diff
@@ -33,19 +34,84 @@ Public Class dlgFlatFrequencyTable
...
     Private Sub SetRCodeForControls(bReset As Boolean)
-
+        ucrChkAddMargins.SetRCode(clsAddMargin, bReset)
+        ucrRowVariable.SetRCode(clsFTable, bReset)
+        ucrColumnVariable.SetRCode(clsFTable, bReset)
+        ucrSelectorDataFrame.SetRCode(clsTable, bReset)
     End Sub
 
     Private Sub TestOkEnabled()
+        If Not ucrRowVariable.IsEmpty AndAlso Not ucrColumnVariable.IsEmpty AndAlso ucrSelectorDataFrame.ucrAvailableDataFrames.cboAvailableDataFrames.Text <> "" Then
```
dannyparsons: what happens if I only have row variables? doesn't it still work?

AlexSananka: This is now fixed,one issue we had is that you can't have one variable in the table function.We have fixed TestOk such that we must have more than one variable.either two in one receiver or one in each receiver.
We tried as much as we can to avoid introducing an error message box to tell the user to add more than one variable.

dannyparsons: Just to be clear, when there's only one variable the table function drops the column name, meaning we would have to change the way we use ftable to get it work. Since this would be more work just to be able to do a not very nice looking one way table, I'm happy we require at least two variables for this dialog.

It should be in the help so that it is clear when OK is enabled.

instat/dlgFlatFrequencyTable.vb
```diff
@@ -80,10 +80,14 @@ Public Class dlgFlatFrequencyTable
     End Sub
 
     Private Sub TestOkEnabled()
-        If Not ucrRowVariable.IsEmpty AndAlso Not ucrColumnVariable.IsEmpty AndAlso ucrSelectorDataFrame.ucrAvailableDataFrames.cboAvailableDataFrames.Text <> "" Then
-            ucrBase.OKEnabled(True)
+        If ucrSelectorDataFrame.ucrAvailableDataFrames.cboAvailableDataFrames.Text <> "" Then
+            If ((ucrRowVariable.lstSelectedVariables.Items.Count > 1 AndAlso ucrColumnVariable.IsEmpty) OrElse (ucrColumnVariable.lstSelectedVariables.Items.Count > 1 AndAlso ucrRowVariable.IsEmpty) OrElse (Not ucrRowVariable.IsEmpty AndAlso Not ucrColumnVariable.IsEmpty)) Then
```
dannyparsons: This should simply be:
```
ucrRowVariable.lstSelectedVariables.Items.Count + ucrColumnVariable.lstSelectedVariables.Items.Count > 1
```

instat/dlgFlatFrequencyTable.vb
```diff
@@ -81,13 +81,13 @@ Public Class dlgFlatFrequencyTable
 
     Private Sub TestOkEnabled()
         If ucrSelectorDataFrame.ucrAvailableDataFrames.cboAvailableDataFrames.Text <> "" Then
-            If ((ucrRowVariable.lstSelectedVariables.Items.Count > 1 AndAlso ucrColumnVariable.IsEmpty) OrElse (ucrColumnVariable.lstSelectedVariables.Items.Count > 1 AndAlso ucrRowVariable.IsEmpty) OrElse (Not ucrRowVariable.IsEmpty AndAlso Not ucrColumnVariable.IsEmpty)) Then
+            If ((ucrRowVariable.lstSelectedVariables.Items.Count + ucrColumnVariable.lstSelectedVariables.Items.Count > 1) OrElse (Not ucrRowVariable.IsEmpty AndAlso Not ucrColumnVariable.IsEmpty)) Then
```
dannyparsons: Do you think both conditions are still needed here?

instat/dlgFlatFrequencyTable.vb
```diff
@@ -112,6 +112,16 @@ Public Class dlgFlatFrequencyTable
...
+        ElseIf Not ucrRowVariable.IsEmpty AndAlso ucrColumnVariable.IsEmpty Then
+            clsGetVar.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_columns_from_data")
+            clsGetVar.AddParameter("data_name", Chr(34) & ucrSelectorDataFrame.ucrAvailableDataFrames.cboAvailableDataFrames.SelectedItem & Chr(34))
+            clsGetVar.AddParameter("col_names", (ucrRowVariable.GetVariableNames))
+            clsTable.AddParameter("x", clsRFunctionParameter:=clsGetVar)
+        ElseIf ucrRowVariable.IsEmpty AndAlso Not ucrColumnVariable.IsEmpty Then
+            clsGetVar.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_columns_from_data")
+            clsGetVar.AddParameter("data_name", Chr(34) & ucrSelectorDataFrame.ucrAvailableDataFrames.cboAvailableDataFrames.SelectedItem & Chr(34))
+            clsGetVar.AddParameter("col_names", (ucrColumnVariable.GetVariableNames))
+            clsTable.AddParameter("x", clsRFunctionParameter:=clsGetVar)
```
dannyparsons: There is a lot of repetition here. There is only one line that I see that varies depending on the conditions in the If and If Else. Do any of the other lines need to be repeated under each condition? Could they be done just once outside the conditions?
A general programming rule is not to repeat code unnecessarily and if you find yourself writing the same code in many places that's a sign it's probably not the best way to do it.

instat/dlgFlatFrequencyTable.vb
AlexSananka: No I should have removed the second condition since it's already included in the first one.

instat/dlgFlatFrequencyTable.Designer.vb
```diff
@@ -45,8 +117,15 @@ Partial Class dlgFlatFrequencyTable
...
         Me.ResumeLayout(False)
+        Me.PerformLayout()
 
     End Sub
 
     Friend WithEvents ucrBase As ucrButtons
+    Friend WithEvents ucrSelectorDataFrame As ucrSelectorByDataFrameAddRemove
+    Friend WithEvents ucrRowVariable As ucrReceiverMultiple
+    Friend WithEvents ucrColumnVariable As ucrReceiverMultiple
+    Friend WithEvents lblRowvariable As Label
```
lilyclements: This should be called `lblRowVariable` (capital V). Sorry I missed this before!

AlexSananka:
@lilyclements  and @dannyparsons  this is ready for re-review

lilyclements:

PR #3063, opened 2017-05-12
dlgCummulativeDistribution

maxwellfundi:


instat/dlgCumulativeDistribution.Designer.vb
```diff
@@ -75,17 +64,18 @@ Partial Class dlgCumulativeDistribution
...
         Me.chkIncludePoints.Size = New System.Drawing.Size(93, 17)
-        Me.chkIncludePoints.TabIndex = 7
+        Me.chkIncludePoints.TabIndex = 6
         Me.chkIncludePoints.Tag = "Include Points"
         Me.chkIncludePoints.Text = "Include Points"
         Me.chkIncludePoints.UseVisualStyleBackColor = True
         '
         'cmdLineOptions
         '
+        Me.cmdLineOptions.Enabled = False
```
maxwellfundi: Diasbled this temporarily before i add the stat_ecdf() on the list of geoms in the layer options dialog.

instat/dlgCumulativeDistribution.vb
```diff
@@ -17,110 +17,126 @@
...
-        clsRaesFunction.SetRCommand("aes")
-        clsRggplotFunction.AddParameter("mapping", clsRFunctionParameter:=clsRaesFunction)
-        ucrBase.clsRsyntax.SetOperatorParameter(True, clsRFunc:=clsRggplotFunction)
-        clsRgeom_CumDistFunction.SetRCommand("stat_ecdf")
-        ucrBase.clsRsyntax.SetOperatorParameter(False, clsRFunc:=clsRgeom_CumDistFunction)
 
+    Private Sub InitaliseDialog()
+        Dim clsScaleYReverseFunc As New RFunction
+        Dim clsScaleYReverseParam As New RParameter
+        ucrChkIncludePoints.Enabled = False ' temporary
```
dannyparsons: Is this difficult to do?

maxwellfundi: From my understanding we are not sure about we want this to do but, from my research, if we wanted just points we would simply add a parameter inside the `stat_ecdf` called `geom="point"` and we could get points instead of lines. 

I am not sure if that is what we want. maybe @rdstern could shed a light on this.

dannyparsons: So that's an alternative to a line not as well. Do we want the option of points or lines in the dialog? Or just fixed as a line?

rdstern:
Sometimes it could be useful to have the points as well as the line(s).  That helps to emphasise how many data points there are.  I assume this could always be another layer?  here is an example from the old Instat.  Default is just line.
cumulative plot in old instat

instat/dlgCumulativeDistribution.vb
maxwellfundi: Yeah, that would be an alternative. I see @rdstern would want the points to be on the line, not as an alternative. 
I will have to look up how to get those points on the lines.

dannyparsons: If doing both is complex then an temporary intermediate step could be a choice of points or lines. Or if points on their own are not sensible leave out points until we can do both.

dannyparsons:
What did you find out about adding points? I just quickly searched and found adding `geom_point(stat = "ecdf")` adds the points on the line. Can't we add that?

maxwellfundi:
@dannyparsons Adding just points is easy by either having `stat_ecdf(geom="point")` or `geom_point(stat = "ecdf")` which work the same, but we want to have both at the same time.

maxwellfundi:
@dannyparsons  I have seen how to add the point on the line now, i think i was not getting it correctly. 

So, I dont know whether there will ever be a time we would only want point?

dannyparsons:
I guess that `stat_ecdf(geom="point")` and `geom_point(stat = "ecdf")` are identical so it doesn't matter which we use - might be good to just check the difference quickly though.

maxwellfundi:
Couldn't find substantial difference. They work the same way. @dannyparsons that is now ready.

PR #3071, opened 2017-05-14
added shift day/year R code

### LINKED ISSUE #2968: Adding shift option to the Prepare > Column: Dates > Use Date dialogue
I suggest this work could be for Shadrack with support from Steve and Danny.

This is under (or next to to the Day in Year (366).  It is called something like Shifted day.  You might change the layout of the boxes perhaps, so there is one that goes right across with the label "Day" and is above , 3 of which are there already"Other Functions".  There are 4 options inside this box 3 of which are there already (and don't change).  The first could now be labelled "In Month" and is the one called Day now.  The second and third are as now and labelled "In Year" and "In Year (366)".  
The new one is called "Shifted".  If checked, then 2 up-down controls become also visible,  The first goes from 1 to 12 with 8 as the default.  The second goes from 1 to 31 with 1 as the default.
This will simply generate a variate which is the same sa the "In Year (366)" but starting from a different month/day.  
1. This will generate a column called shiftx e.g. shift8 (where x = 1 to 12), if the day of the month is 1.  Otherwise the name of the column could be shift8.11 if the start day is 11th August.
2. Spend just a little time on the code (15-30 minutes) and then discuss, with Steve and/or Danny, on how to do this calculation neatly.
3. Discuss also how (and particularly where) you will check that the date is valid, You could (perhaps) organise that the second up-down control (day of month) only goes to 30 if the month is 4, 6, 9, 11 and 29 if it is 2.  Alternatively this could be a check that is a preliminary when OK is pressed.  (There must be other dialogues where combinations of controls are not OK, though individually they are OK.)
4. Finally there will also be a small change in the **Climatic > Define Climatic Data** dialogue.  The "Day of Year" field is currently filled by one of the two options if they have been generated.  Now, if a Shiftx column exists this has priority.  (No extra field is needed.)  And while on that dialogue you may wish to change the label above from **"Day" to "Day in Month"**.

dannyparsons:
fixes #2968

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -1801,7 +1801,7 @@ data_object$set("public","set_contrasts_of_factor", function(col_name, new_contr
   }
 )
 #This method gets a date column and extracts part of the information such as year, month, week, weekday etc(depending on which parameters are set) and creates their respective new column(s)
-data_object$set("public","split_date", function(col_name = "", week = FALSE, month_val = FALSE, month_abbr = FALSE, month_name = FALSE, weekday_val = FALSE, weekday_abbr = FALSE, weekday_name = FALSE, year = FALSE, day = FALSE, day_in_month = FALSE, day_in_year = FALSE, leap_year = FALSE, day_in_year_366 = FALSE, dekade = FALSE, pentad = FALSE) {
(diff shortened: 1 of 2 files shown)


dannyparsons:
@stevekogo this branches from my previous branch. It could either be merged into yours or merged later.

PR #3074, opened 2017-05-15
Factor menu updates

Lunalo:
I have updated and made some edits on Combine Text, Factor Dataframe and Reorder levels dialogs


Lunalo:

This is ready for review

lilyclements:
@Lunalo I have made changes waiting to be merged

dannyparsons:
What's happening with this? Nothing seems to have been done since the last review.

Lunalo:
@lilyclements 

from the  code  if replace is set to TRUE which is the  rdefault  then the existing factor dataframe which is linked to the main dataframe is replaced when you press okay and nothing happens if the value of the parameter is set to FALSE

lilyclements:
I can see in the code what TRUE and FALSE do. However, when running it I get a little confused. Let me be a little clearer:
Does "Replace if already exists" mean to replace that data frame if the contents of that data frame already exists? Or to replace that data frame if the name already exists?

My issue comes from here -
* Using "Survey", I make a dataframe with the factor "Village" with the name "Vill"
* Then I change factors, keep the name "Vill" and press OK. It makes "Vill1". That's fair enough. My "Replace if already exists" is checked, so I assume it doesnt mean to replace the data frame if that name already exists, but to replace the contents if those contents are in a data frame. I haven't yet done a data frame with this factor so that's fair enough.
* Then I keep this factor, uncheck "Replace if already exists" and press OK. It replaces my data frame instead of making a new one. Surely this should make a new data frame? If I uncheck "Contrasts" and ask to not replace if it already exists, it still replaces it.
* If I keep this unchecked, use the same factor and now change the name of the data frame to say "Variety" then it keeps the name as Vill1. However, from the code, it is running `factor_data_frame_name="Variety", replace=FALSE`
But, it is replacing my data frame and it is not renaming it "Variety"

Can you explain to me what this check box does?

lilyclements:
I have made other bug fixes and layout edits so please merge them in before making any new changes here

dannyparsons:
Some of these dialogs have now been updated, can this be closed? Or redone as a new pull request?

PR #3081, opened 2017-05-15
Transform text

lilyclements:


instat/dlgTransformText.vb
```diff
@@ -18,15 +18,14 @@ Imports instat.Translations
 Public Class dlgTransformText
     Public bFirstLoad As Boolean = True
     Private bReset As Boolean = True
-    Private clsConvertFunction, clsLengthFunction, clsPadFunction, clsTrimFunction, clsWordsFunction, clsSubStringFunction As New RFunction
+    Private clsConvertFunction, clsLengthFunction, clsPadFunction, clsTrimFunction, clsWordsFunction, clsSubstringFunction As New RFunction
```
lilyclements: Changed SubString to substring

instat/dlgTransformText.vb
```diff
@@ -172,70 +171,44 @@ Public Class dlgTransformText
...
         clsTrimFunction = New RFunction
         clsWordsFunction = New RFunction
-        clsSubStringFunction = New RFunction
+        clsSubstringFunction = New RFunction
 
         ucrNewColName.Reset()
         ucrSelectorForTransformText.Reset()
 
         NewDefaultName()
+        clsConvertFunction.SetPackageName("stringr")
```
lilyclements: Added package names

instat/dlgTransformText.vb
```diff
@@ -172,70 +171,44 @@ Public Class dlgTransformText
...
-                Case "Title"
-                    ucrBase.clsRsyntax.SetFunction("str_to_title")
-            End Select
-        End If
-        TestOkEnabled()
-        SetRCodeForControls(False)
-    End Sub
-
-    Public Sub SetRCodeForControls(bReset As Boolean)
+    Private Sub SetRCodeForControls(bReset As Boolean)
```
lilyclements: Changed Public to Private sub

lilyclements:
@africanmathsinitiative/developers should I change the layout of this dialog to be consistent with other dialogs in the same sub menu (Prepare > Text)

Currently it looks like this:

image

Other examples in this menu are like this:

image

Should I move the rdo's up to the top. The rdo's control the base function so I think it would be appropriate to do so here. What do you think?

dannyparsons:
I think they could be moved yes, they are the same situation.

AlexSananka:
@lilyclements is this ready for review?

lilyclements:
There is an error which comes up with this dialog. I know where the error is, however, I'm not sure why this error is coming up. Danny was going to look into it if/when he got spare time.

The error is a line in the designer. This line has been automatically added

lilyclements:
@dannyparsons this is ready, except, there is a bug on this branch. We discussed it a while ago and you said you'd look into it. The issue was, if I change the designer then the following line was automatically added to my panels and inputs: `.AddQuotesIfUnrecognised = True`

When added to my inputs, it caused a developer error. I can just remove the line, but then it may come back for someone else if/when they edit this dialog again!

dannyparsons:
Try removing it for now and let's see if it comes back.

lilyclements:
@dannyparsons I have commented out the line causing issue. I have also sorted one of the combo boxes to not be editable. This is ready to review

PR #3068, opened 2017-05-13
Enhancing netCDF dialog

stevekogo:

PR #3085, opened 2017-05-16
Stevekogo net cdf

dannyparsons:


dannyparsons:
@lilyclements I will test further when I get to the airport but in case you want to look at it before then here it is

lilyclements:
I managed to crash this for one of my datasets if I put the same variable in the long and lat.

> Error in `row.names<-.data.frame`(`*tmp*`, value = c("period", "X", "X",  : 
>   duplicate 'row.names' are not allowed

(However, it didn't crash for another dataset)
We should think about putting in an exclude however to prevent this error.

stevekogo:
@dannyparsons I am working on a different netcdf branch #3068. Why do we have this other branch?

dannyparsons:
I branched from yours to edit parts of the dialog. Do you want to merge this back into yours?

lilyclements:
@dannyparsons this can be closed since these changes have also been merged into #3395

PR #3086, opened 2017-05-16
Dlg Jitter Bug Fix on change of data frames

stevenndungu:
The parameter `n` could be updated wgen you changed from one data frame to the other.
image

### DIFF
--- a/instat/dlgJitter.vb
+++ b/instat/dlgJitter.vb
@@ -146,7 +146,7 @@ Public Class dlgJitter
         clsRunif.AddParameter("n", ucrSelectorForJitter.ucrAvailableDataFrames.iDataFrameLength)
     End Sub
 
-    Private Sub ucrSelectorForJitter_DataFrameChanged() Handles ucrSelectorForJitter.DataFrameChanged
+    Private Sub ucrSelectorForJitter_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSelectorForJitter.ControlValueChanged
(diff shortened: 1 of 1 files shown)


shadrackkibet:
i am going to test this now.

stevenndungu:
Please do.

shadrackkibet:
@africanmathsinitiative/developers I have been testing this dialog but i didn't understand how the dialog works.could someone  explain how jitter works?

image


I noticed that when the input textboxes on the group box are empty TeskOk is  still enabled and i was wondering whether the output on teskOk click  is right.

dannyparsons:
The dialog takes a numeric column and adds a random component to it. The input controls allows you to specify the maximum amount that will be added/subtracted from the column.

This is useful when you have confidential data. For example, if you have a latitude column for someone's house and you want to make the data public, you might not want to give the exact location of a house because it could be used to identify the people who live there. One option is deleting the column, but you could also just add a random component to it. So that now the latitude column is slightly different so it doesn't identify the house, but still gives an idea of where it is roughly located (which could be useful for analysis) if you specify the maximum to be small.

The reason it is called jitter is because you jitter the column i.e. make small changes to the values of the column.

shadrackkibet:
This is a good example to understand the use and importance of jitter dialog. So it is okey for TestOk to be enabled when input textboxes on the group box are empty because it appends a column which is different from the original column.

shadrackkibet:

PR #3087, opened 2017-05-16
re-writing sdgplots 

maxwellfundi:
This work has just began, its not close to be reviewed yet.


instat/sdgPlots.vb
```diff
@@ -41,19 +46,36 @@ Public Class sdgPlots
...
+        clsGraphTitleFunction.SetRCommand("ggtitle")
+        clsBaseOperator.AddParameter("ggtile", clsRFunctionParameter:=clsGraphTitleFunction)
+
+        clsLegendFunction.SetPackageName("ggplot2")
+        clsLegendFunction.SetRCommand("labs")
+        clsBaseOperator.AddParameter("ledendtitle", clsRFunctionParameter:=clsLegendFunction)
+
+        clsRFacetFunction.SetPackageName("ggplot2")
+        clsRFacetFunction.SetRCommand("facet_wrap")
+        clsBaseOperator.AddParameter("facets", clsRFunctionParameter:=clsRFacetFunction)
```
maxwellfundi: @dannyparsons  could you please have a look at this. I am not sure having these here is the right thing for our linking.

instat/sdgPlots.vb
```diff
@@ -167,42 +235,40 @@ Public Class sdgPlots
...
+        'ucrPlotsAdditionalLayers.SetAesFunction(clsAesFunction)
+        ''The following two setup the ucrAdditionalLayers on the sdgPlots. Shares the global ggplot function, as well as the whole PLots RSyntax.
+        'ucrPlotsAdditionalLayers.SetGGplotFunction(clsRggplotFunction)
+        'ucrPlotsAdditionalLayers.SetRSyntax(clsRsyntax)
+        ''This is necessary to make sure the minimum number of fixed rows or columns is 1.
+        'ucrNudNumberofRows.Minimum = 1
+
+        ''Set's the X Axis tab to X mode and the YAxis tab to Y mode (each tab contains a generic ucrAxis with internal X or Y boolean setting).
+        ''Also carry the RSyntax through to these ucr's .
+        'ucrXAxis.SetXorY(True)
```
maxwellfundi: @dannyparsons  i also want to use this method on the ucrAxes to link up theucr with the subdialog. I am not sure we will do it the normal way we do for subdialogs with the setrfunction method

dannyparsons: I'm not quite sure what you mean here. But I assume this method will still be needed for the axis control. I would suggest getting the other parts of the dialog working before working on ucrAxis though, as that's going to be more difficult anyway.

instat/sdgPlots.vb
```diff
@@ -22,30 +22,210 @@ Public Class sdgPlots
...
+        clsGraphTitleFunction.SetRCommand("ggtitle")
+        clsBaseOperator.AddParameter("ggtile", clsRFunctionParameter:=clsGraphTitleFunction)
+
+        clsLegendFunction.SetPackageName("ggplot2")
+        clsLegendFunction.SetRCommand("labs")
+        clsBaseOperator.AddParameter("ledendtitle", clsRFunctionParameter:=clsLegendFunction)
+
+        clsRFacetFunction.SetPackageName("ggplot2")
+        clsRFacetFunction.SetRCommand("facet_wrap")
+        clsBaseOperator.AddParameter("facets", clsRFunctionParameter:=clsRFacetFunction)
```
dannyparsons: You can't be adding parameters here, because that depends on what is currently in the base operator. I think you can remove all the adding parameters since if they need to be added, that should have been done from the main dialog.

maxwellfundi: Okay, i understand, but then there would be a problem when we add say facets parameter which carries the facet function to the base operator , then opening the subdialog, then going back to the main dialog. That would cause a crash because nothing would have been added to the function.

dannyparsons: The sub dialog can add and remove the parameter - but only when the checkbox is changed, not when you open the sub dialog.

maxwellfundi: Great, yeah I see.

instat/sdgPlots.vb
```diff
@@ -121,6 +121,8 @@ Public Class sdgPlots
         'titles tab
 
         ucrInputGraphTitle.SetParameter(New RParameter("title"))
+        ucrInputGraphSubTitle.SetParameter(New RParameter("subtitle"))
+        ucrInputGraphCaption.SetParameter(New RParameter("caption"))
```
maxwellfundi: @dannyparsons  after adding these two parameters for the labs function, it doesn't run the the labs function the first time for some reason.  

It runs the code properly when i debugged and I do not seem to find where precisely the issue is.

maxwellfundi:
@dannyparsons I have now sent out all the changes including disabling tabs for things that dont work. its ready for merging

dannyparsons:
merging this because of vbproj changes, not functional but will be by end of today.

PR #3090, opened 2017-05-16
Allowing all data types in the selector of Random Subset dialog

### LINKED ISSUE #3089: Random subset of a factor column
I have this dataset in eSMS which contains a variety column with 10000 cases. The column is a factor. I want to select 4 random samples of 250 each for some exercise with my students. I thought the random subset was build for such task but I could not use it because in the dialogue, the variety column does not appear in the Variables box for selection. I think it is because it is of the factor type. Trying the same task with a numeric column seems to work well. Is there a bug there or am I using the wrong dialogue?

stevenndungu:
@dannyparsons I have allowed all data types now. fixes #3089


dannyparsons:
Great. Did you test it on factors?

stevenndungu:
Yap, I tested.

Issue #3067, opened 2017-05-12
Display of daily climatic data


rdstern:
This is one of the last set of features I am keen to see soon and certainly in the first release.  I think the only other is the crops dialogue that should be easy once the start, end and summary (already specified) have been done.

In the old Instat there is a dialogue to display daily data in 2 different ways.  I assume this might again be a single dialogue with our new buttons at the top to permit the results as either a Table or a Graph.
Also the main initial fields will again be the same as usual.  Perhaps we need a column for the day of the month here?  At least for the table.  We need the day of the year for the graph.

I give the requirements below, but not (yet) the layout.  I am happy to discuss the layout once we know who is doing what, and also when it might be done.  I suspect the table part might depend on the progress with the main tabulation.  The graphs would be great - and useful as soon as someone has the time. 

**1) As a "table".**  
a) This has the months across (columns) and the day of the month as the rows.
b) it is a trivial example of a table (or an unstack by the month factor). 
c) It is "trivial" as a table because it is usually used to display single values, rather than a summary.
d) It may later become a summary when we deal better with within-day data, though it would usually have a "shifted day" rather than starting at midnight.
e) Ideally it could be able to include a symbol for a range of values (or a single special value (e.g. "." for zero, so rainfalls stand out.  
f) This would display the data in the output window, as in the old Instat.
g) We would be able to choose the number of decimals to display.
h) It could be for up to 12 months and could start from any month, but the default would be 12 months starting from January.
i) It could have margins after each year (i.e. for each month as well as annual - including sum (for rainfall),mean (for temperature) count (from a certain threshold, min, max, and number of missing.  
j) A non-trivial set of margins is the annual values.  That's odd because we don't need the extra margin on the right! (Once we have margins in the table Danny says this is trivial.)

**2 As a graph or set of graphs**   
a) The graphs were hardly used in the old Instat.  They are "needle plots - sort of bar charts - for rainfall and possibly line charts for temperatures and other elements.
b) here they could be great, because of facets, i.e. many years on a single graph.
c) Default is for a graph per year.
d) Another alternative could be for a graph per station - over all the years (filtered usually).  This is very common in other software, but usually not so easy, because it confuses the seasonality with the trend.
e) ideally labels could be months or day of the year.
f) Ideally it would also be able to indicate missing values clearly.  
g) David has the commands for much of this from his work for Guyana.

rdstern:
[...]

PR #3103, opened 2017-05-19
Display daily climatic data

AlexSananka:
In progress


AlexSananka:
@africanmathsinitiative/developers I am still trying to figure out on how the design of this dialog should like.
image

What are your thoughts on this.
You can have a look at #3067 and get the picture.

rdstern:
It looks neat.  Usually we put the receiver on the right-hand side of the dialogue.  And that is the same for the table and the graph.  So it could go under the **year** field.
Then the dialogue is currently indicating table, but the remaining field are all associated with a graph.  So they could go into a group box with the label **Graph**.
Inside there is the limits for the y-axis, Lower and Upper.  They are numeric and could include an up-down control.  The x-axis is also there and the default is the doy column.
There will be another group for the Table.

Issue #2635, opened 2017-02-02
Some dialogue not to be updated


rdstern:
In the **model** menu the last 4 main items are all called **Other**, e.g. **Other (One Variable)**.  They correspond to at least 10 dialogues.

Ignore all these dialogues - at least for now.  It is likely that all their functionality will be handled by the earlier (more general) modelling dialogues in the same menu.

In the Excel list of dialogue names Lily is adding "Ignore for Now" as a comment.
 
Thanks

lilyclements:
I believe I have done this now for all dialogs which are in the "other" part and not also above the line (such as dlgRegressionSimple)

These are
dlgGeneralANOVA
dlgGeneralRegression
dlgGoodnessofFit
dlgNon-ParametricOneSampleTests
dlgNon-ParametricOneWayANOVA
dlgOneSample
dlgOneWayAnova

I've put "ignore for now" next to these dialogs

maxwellfundi:
@dannyparsons This can now be closed

PR #3107, opened 2017-05-22
Rewriting Non-Parametric one way ANOVA

stevenndungu:
Few changes made

1.  The first receiver was not passing its variable ie, 
       `kruskal.test(formula=variable2, x=data)`
instead of 
       `kruskal.test(formula=variable1~variable2, x=data)` ....fixed this semantic error

2. Restricted variable1(Y-variate variable) to numeric variables only instead of all variables. This is how kruskal should work, ie, numeric vector vs a factor variable, in order it can compare medians across the different levels given by variable 2


rdstern:
Can we have a discussion on the work on the dialogues that are "below the line" in the modelling menu.  My understanding is that they will all be incorporated within the dialogues above the line.  So this one is an example of a dialogue using 2 variables and is (or will be) put there.

stevenndungu:
Yes we can have a discussion on this.

dannyparsons:
This looks like it's still waiting for review @africanmathsinitiative/developers

lilyclements:
On the google docs I put a "Ignore for now" next to this since it is listed here #2635 as a dialog to not update.
Happy to review later today if everyone agrees it should be.

dannyparsons:
Since it's been updated, might as well review quickly. If it's going to be more work to get it ready then we leave it for now.

stevenndungu:
@dannyparsons I have tested this...now that I had already re-written it. It can just be merged in.

PR #3108, opened 2017-05-22
Re-writing dlgInsert columns

shadrackkibet:


shadrackkibet:
@africanmathsinitiative/developers this is in progress.

shadrackkibet:
@africanmathsinitiative/developers This is ready for review now.

shadrackkibet:
@lilyclements  this is ready for re-review after addressing your points above.

lilyclements:
Two small issues:

1. If I open the dlg, and say to insert **after** column x, it inserts before column x. If I then select before and then select after, then press OK, it will run after.
So, the R-Code you're using is correct, but upon first time using it, it runs `before = TRUE` when we don't want to be running this

2. If I click reset, I get the following developer error:

> Developer error: The value given cannot be converted to a decimal or is outside the range of the control. Value will be unchanged.

lilyclements:
EDIT: The developer error occurs if you insert a new row, then press reset. I assume the nud it is having issue with is the "Insert at Position" nud

shadrackkibet:
This is ready for re-review again.

lilyclements:
@shadrackkibet when I try to run this dialog it crashes on ucrNud. Are you having this problem too?

instat/dlgInsertColumn.vb
```diff
@@ -201,9 +201,10 @@ Public Class dlgInsertColumn
 
     Private Sub ucrReceiverColumnsToInsert_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverColumnsToInsert.ControlContentsChanged, ucrPnlColumnsOrRows.ControlContentsChanged, ucrPnlBeforeAfter.ControlContentsChanged, ucrPnlInsertColumns.ControlContentsChanged, ucrInputPrefixForNewColumn.ControlContentsChanged, ucrInputDefaultValue.ControlContentsChanged, ucrInputBeforeAfter.ControlContentsChanged, ucrNudNumberOfRows.ControlContentsChanged, ucrNudStartRow.ControlContentsChanged, ucrNudNumberOfColumns.ControlContentsChanged
         TestOKEnabled()
+        ucrNudStartRow.SetMinMax(1, ucrDataFramesList.iDataFrameLength)
```
lilyclements: Running this line here is causing the crash

shadrackkibet: ooooh, let me look into this now. I have noticed this too.

shadrackkibet:
@lilyclements this is ready for re-review again.

lilyclements:
It crashed on ucrNud again. This time, if I have two datasets open (of different lengths)

If I open on the shorter data frame, it runs fine.
If I open on the longer data frame, it crashes.

Similarly, if I open on the shorter data frame and then change data sets to the larger data frame it will crash.
It crashes on 
`            nudUpDown.Value = dcNewMax`

Presumably it's struggling with this change in maximum values, since the nudUpDown is saying the maximum value it can take is the length of the smaller data frame
But then it's trying to set a value greater than this (since it's setting the value as the new maximum).

instat/dlgInsertColumn.vb
```diff
@@ -1,360 +1,218 @@
...
-            If Not nudInsertColumns.Text = "" Then
-                ucrBase.clsRsyntax.AddParameter("num_cols", nudInsertColumns.Value)
-            Else
-                ucrBase.clsRsyntax.RemoveParameter("num_cols")
-            End If
-        Else
-            ucrBase.clsRsyntax.RemoveParameter("num_cols")
-        End If
+    Private Sub ucrSelectorInsertColumns_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrDataFramesList.ControlValueChanged
+        ucrNudStartRow.Value = ucrDataFramesList.iDataFrameLength
```
lilyclements: To solve our problem, we want to set the MinMax value before setting the new value. So, how about trying to run
`        ucrNudStartRow.SetMinMax(1, ucrDataFramesList.iDataFrameLength)` 
here before running
`.Value = ...`

shadrackkibet: This solves our problem, however i am having another problem which i am yet to fix now.

lilyclements: Whats the new problem?

shadrackkibet: when you load two datasets the selector updates  to both dataframe names.

shadrackkibet:
@lilyclements i have addressed the issue above and this is now fine.Have a look at it please.

instat/dlgInsertColumn.vb
```diff
@@ -1,360 +1,228 @@
...
+    Private Sub SetDefaults()
+        clsInsertColumnFunction = New RFunction
+        clsInsertRowFunction = New RFunction
+
+        ucrSelectorInsertColumns.Reset()
+        ucrInputBeforeAfter.Reset()
+        ucrInputDefaultValue.Reset()
+        ucrInputPrefixForNewColumn.Reset()
+        ucrInputPrefixForNewColumn.SetName("X")
+        ucrInputDefaultValue.SetName("NA")
```
dannyparsons: You can't set the values for controls, these must be added to the default function. Or set as the default state.

instat/dlgInsertColumn.vb
```diff
@@ -1,360 +1,228 @@
...
 
-    Private Sub ucrInputPrefixForInsertedColumns_NameChanged() Handles ucrInputPrefixForInsertedColumns.NameChanged
-        ColName()
-        TestOKEnabled()
+    Private Sub ucrPnlStartEnd_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlInsertColumns.ControlValueChanged
+        If rdoAtEnd.Checked Then
+            clsInsertColumnFunction.AddParameter("before", "FALSE")
+        ElseIf rdoAtStart.Checked
+            clsInsertColumnFunction.AddParameter("before", "TRUE")
+        End If
```
dannyparsons: Does this really need to be done? Can't the control do this automatically?

instat/dlgInsertColumn.vb
```diff
@@ -1,360 +1,228 @@
...
+        ucrPnlColumnsOrRows.AddToLinkedControls(ucrPnlBeforeAfter, {rdoInsertRows}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrPnlColumnsOrRows.AddToLinkedControls(ucrNudNumberOfRows, {rdoInsertRows}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrPnlColumnsOrRows.AddToLinkedControls(ucrNudStartRow, {rdoInsertRows}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+
+        ucrNudNumberOfColumns.SetLinkedDisplayControl(lblNumberOfColumnsToInsert)
+        ucrInputDefaultValue.SetLinkedDisplayControl(lblDefaultValue)
+
+        ucrInputPrefixForNewColumn.SetParameter(New RParameter("col_name", 6))
+        ucrInputPrefixForNewColumn.SetLinkedDisplayControl(lblInputColumnPrefix)
+        ucrInputPrefixForNewColumn.SetPrefix("X")
```
dannyparsons: If this control is to enter a prefix then you shouldn't set a prefix for the control itself, otherwise it will automatically add a number to the prefix, which isn't what you want if the user is supposed to type a prefix

shadrackkibet: I am then going to remove this

instat/dlgInsertColumn.vb
shadrackkibet: will set this for the  defaultfunction.

instat/dlgInsertColumn.vb
shadrackkibet: This can't work automatically on a the control. we tried this with @stevenndungu but it could't work.

shadrackkibet:
@dannyparsons i have addressed concerns above except the issue of passing the  parameter on the two radio buttons automatically which i couldn't achieve. I tried this with @stevenndungu but it could't work and that is why we opted for a manual way.

instat/dlgInsertColumn.vb
```diff
@@ -1,360 +1,227 @@
...
-        BeforeParameter()
+    Private Sub SetDefaults()
+        clsInsertColumnFunction = New RFunction
+        clsInsertRowFunction = New RFunction
+
+        ucrSelectorInsertColumns.Reset()
+        ucrInputBeforeAfter.Reset()
+        ucrInputDefaultValue.Reset()
+        ucrInputPrefixForNewColumn.Reset()
+        ucrInputPrefixForNewColumn.SetName("X")
```
dannyparsons: The same needed for this.

shadrackkibet: I have fixed this.

PR #3110, opened 2017-05-23
Rewriting dlgOne var use model

AlexSananka:
@africanmathsinitiative/developers this is in progress


AlexSananka:
Note:You should only use only ` fitdist` objects.that is make sure you fit your objects first.

lilyclements:
Two more minor problems I didn't get to:
- Upon reopening the dialog, the button for "bootstrap" is visible, regardless of if ucrChkProduceBootstrap is checked or not. (this is at least the case if you check ucrChkProduceBootstrap, uncheck it, close, reopen the dlg). I think this can be fixed if you run the sub regarding this button on reopen()? Then it can check if ucrChkProduceBootstrap is checked or not.

- When running it, the graphs should go into the output window, this is just an iCallType fix

AlexSananka:
@africanmathsinitiative/developers I have two subdialogs in the same dialog and I was wondering if we can merge them and them in one subdialog with different tabs having the different options.Currently when I click the fitmodel option I have 
image
When I click the bootstrap option I have 
image.
What are your views on this?

dannyparsons:
Can you explain your current thinking on this? I don't know the dialog well so it would be useful for you to explain why you think it should be two sub dialogs or why you think they should be merged or why you're not sure?
Currently, do both sub dialogs apply whatever model I select? Or what do they depend on? I think we need more information on these things to be able to comment on the layout. 
Also, can you post the main dialog as well?

lilyclements:
Another point to discuss with @rdstern is about the layout of the subdialog.

It used to be like this
image


I changed it to this:

image

I changed it because we had a conversation back in January about reducing tabs on sub dialogs if they had a small amount on each tab. I'm not sure about the reduction in the instance of this dialog. I think I prefer the former. It would be good to hear thoughts.

dannyparsons:
It is a bit long now. If the Plot option was a combo box then it would fit nicely all together.

lilyclements:
I think we should make it one sub dialog.

The plot dialog adds parameters to the `quantile` function and the bootstrap dialog adds parameters to the `bootdist` function. 
On the Plot dialog, as just mentioned, there is one option under "plots" which is only enabled when bootstrapping is checked
On the bootstrap dialog, there is one option (CI's) which does not affect the `bootdist` function, but instead affects the `quantile` function on the other sub. I always felt this was a little out of place.

Having it as one dialog could be nice since it will reduce this overlap between the two subs. 

So to conclude, we can have one sub dialog and it can have three tabs:
Plot, which can remain as radio buttons for options for reasons I'll explain below.
Quantile
Bootstrap

And the Bootstrap tab can be hidden or disabled when Bootstrapping is not checked.


As for the plot options - I like the idea of the plot option being a combo box, however, one of those radio buttons are only visible when bootstrapping is checked. So, we'd have to make it that that can't be an option unless it's checked. I think for this reason it works better as radio buttons, especially if we're going to combine everything to one sub.

AlexSananka:
@lilyclements this is ready for re-review just something small that i need to figure out when I want a CDcdf plot.

AlexSananka:
@lilyclements this is ready for re-review.

AlexSananka:
@lilyclements  this is again ready for review.

lilyclements:
I've made changes waiting to be merged.

Things left to do:
- Finish setting the defaults for the Quantiles tab. I've done it for the Plot tab and started on the Quantiles tab
- Sort the code for the "No Plot" option

AlexSananka:
@lilyclements  the above issues have been sorted,this is ready for you to review just before @dannyparsons 
have a look.
I have realized that @lilyclements  you checked the call type and your'e setting all to be 3 which I think is only for graphs and hence the tables cannot be visible since it's trying to plot graphs.Should we maintain the call type 2 for now? I have not got a way of making both the htmltables and the plots come to the output window,this is because some tables like the bootstrap aree done before click ok and the plots are done on click Ok.

AlexSananka:
I haven't got an if statement to do this very well.

AlexSananka:
the above commit fixes the above  call type issue,thanks @lilyclements  for your suggestion.

lilyclements:
* This dialog used to automatically give me my graph if I didn't open the sub dialog. Whereas, at the moment it won't do it automatically unless I open the sdg

* Insert values doesn't have a default value currently, so it doesn't reset properly

* If I pick my distribution (and don't pick multi-plot) then fill the receiver I get an error as it tries to run:
`fitdistrplus::qqcomp(x=dist1, ft=())`
as opposed to
`fitdistrplus::qqcomp(ft=dist1)`

AlexSananka:
@lilyclements on your first point that happened when I was implementing the rdoNoPlot,I had to link the main dialog with the sdg.Can we make No Plot the default?

lilyclements:
We need to find a better reason to make No Plot the default beyond that we can't find a way to do this.
I believe I managed to get it to work in the last version you merged in, this may well have been the wrong method which is why it's now removed. How come this was removed out of interest? We should have a look into that to find the correct way to implement it.
If we can't find a correct method, then I think we should discuss with Danny before deciding to set the default as No Plot.

AlexSananka:
What did you say has been removed?

AlexSananka:
@lilyclements This is ready for you to review

AlexSananka:
@dannyparsons You could have a quick  look at this pull request,we are having  2 issues here,which are outlined in #3120

lilyclements:
@AlexSananka this looks good! What's happening about the two known issues?
(ucrInput problems and running a plot without opening the dialog)

AlexSananka:
@lilyclements Is this ready for @dannyparsons  to review?

lilyclements:
I don't think this is ready for review until the problems in #3120 have been addressed

PR #3114, opened 2017-05-24
Summary calculations

lilyclements:

Here we have added methods to use the summaries system. These have started to be implemented in dlgNewSummaryTables so this dialog is ready for testing.

We have tested the R-code changes with existing dialogs and fixed bugs found. However, we may have missed some. During review, please test further in existing dialogs which use the summary functions.

Also, two new packages need to be downloaded: `htmlTable` and `rje`


dannyparsons:
This is needed to be merged quite quickly, but it should be tested first, at least by one person. Roger can also test further once it is merged.

dannyparsons:
This needs to be merged in this morning, can someone review it @africanmathsinitiative/developers?

stevenndungu:
Testing this now

stevenndungu:
This dialogues working quite well with one bug experienced only. 
I like the html tables given in the output window. They really look nice.
New Summary Tables
-Should the check box ` display NA as`  have a default NA value? instead of putting "" (?)
-If for instance I wanted to have summary tables. I do my first summary table the do the second summary time, after making some changes the second time. The results are saved in the same name. Could we have a save control in place of what we have now? with prefixes i think it would be better.
-otherwise the sub-dialogue and general functionality of the dialogue is just perfect!

stevenndungu:
Column Summaries
This works quite well but with `survey` dataset but when I change to `diamonds` dataset i get a bug. I at first thought its because the dataset was big compared to survey but that sounds not to be the case. This is what happens.
i) Load diamond dataset.
ii) Have the variables you want to get their summaries on first receiver.
iii) Add check several summary statistics that you want added to your summary table (from the sub-dialogue)
iii) Then finally have the factor variables you want to factor by; but for this case begin with one variable, then two variables and finally three variables 
Note the results! plus some  bug pops up though gives some output

this is the error I get
image
But not with three variables added

dannyparsons:
Thanks Steven.
For NA display the default in R is actually `""` so that NA values are blank in the table. In the VB code this should be `Chr(34) & Chr(34)` so that it doesn't just pass nothing. Thanks for noticing this, Lily can correct it.
You are right, currently the saving is not fully implemented, if you notice it just saves it as a variable but not within the instat object. So we do need to have a way of saving tables which is still to do.
Glad you have reproduced the bug from #3115. We haven't resolved this yet but good to know another way to get it.
These will be worked on soon. If there's no other major bug then I'd like to merge it for Roger to see.

stevenndungu:
Yes, this could be merged in. That is the only thing I noted, no other bug!

stevenndungu:
On issue of NA, my thoughts were;
 instead of "" empty string we could have a something like NA set to the inputcombo box... But the user can change to something of his preference. I got a table of missing values which was just empty with only row and column header. After replacing it with "NA" and "Missing", it looked much better compared to  empty space table

dannyparsons:
I agree that when all the values are missing, it can look quite confusing, but also, a table of all missing values is pretty useless however it looks. 
The reason I like `""` as the default is when you have a table with a few missing or empty levels it looks much nicer with blank cells than `NA`, although I can see the reason for us to use `NA` to be consistent with our grids.

It's a good discussion, let's merge it and see what other people think when they try it out.

Issue #3031, opened 2017-05-08
Factors in R-Instat updated


rdstern:
I am excited by the new facilities in R-Instat that will allow:
1) What I call "Levels", i.e. numbers as well as labels for a factor.
2) Intelligently transform between text, factor and numeric without losing information on either levels or labels.
I will be bug-hunting and will report in comments below.  Meanwhile I have a few items - not for the forthcoming release - but that could usefully be discussed while you are all together.
1) The Factor > Levels/Labels dialogue needs to be "enhanced" by adding a further column.  There is a column to edit the combined levels/labels.  Now there needs to be one column for the levels, i.e. the numbers (e.g. 0,1, or 1,2,3,8,9) and another column for the corresponding labels. 
2) The Factor > Reorder Levels could usefully be enhanced for factors with many levels.  The simplest (to me would be to be able to make it the same order as the rows in a corresponding factor data frame.  There might be a number of steps before this can be implemented.  As an example my factor could be (Climatic) Station and I want the order to be latitude.  The latitude is stored in the factor data frame.  It could (of course) become the levels of the factor.
3) In the column metadata the label display is tricky if the factors have many levels.  I wonder what should be done about that?  There is nothing that can be done with that field, so I suggest there is a "..." with a maximum width for that field. (David doesn't like the "..." but agrees that the width of that field should be limited.)

rdstern:
I took the Dodoma data:
1. The obvious factor column was the month.  I made that into a factor from character. etc.  I fiddled with this column and it behaved well.  Character or numeric it did what I wanted.  The metadata showed the labels were there, but of course in alphabetical order.  So far so good.
2. I used the Factor > Reorder Levels (so Jan =1, Feb = 2, etc)  and I think that has worked.  But the labels are no longer in the labels metadata.  Shouldn't they be?
3. I made the year column - numeric - into a factor.  It now said it had labels with 1935 = 1935 as the first label - which I know!  If I make it back to numeric it keeps those labels!  All looks very odd.  (While it is a factor I would have been happy with 1 = 1935... as being informative.  With numeric to factor do we need labels?  I did the same with the rainfall column.
![image](https://cloud.githubusercontent.com/assets/11226469/25821923/cce96e30-342e-11e7-99d6-a4c686353361.png)

rdstern:
1. In the survey data I made Fert into an ordered factor.  It seemed to have no labels.
[...]

PR #3123, opened 2017-05-26
Implemented labels in R method

dannyparsons:
This makes substantial changes to the way converting columns is done to incorporate labels, when they exist. It should not affect normal column conversion but this should be checked.


dannyparsons:
fixes #3031

Issue #1779, opened 2016-09-21
Adding infilling of missing values in Organise > Data Frame > Replace


rdstern:
This is still to be added.
There is a routine, called fill in the tidyr package.  I think Danny is reluctant to use this, because it takes (or returns) a data frame, rather than a column.  There is an alternative called na.locf in the zoo package.  I can not judge if this is better.

If neither of these is simple to implement, then I suggest we mark this as a "still-to-do" task and consider writing our own routine.  It should be quite easy to do.  This would then wait until after the beta is released.  (I would like this feature eventually, but it could wait.)

Roger

dannyparsons:
na.locf looks very powerful with lots of extra options and happy using something from zoo package. I think it's good enough.

stevekogo:
@rstern If this task is not taken by someone else, I  am happy to handle it.

rdstern:
Can do, but there is a lot starting to come in on climatic.  I think that might be better?

stevenndungu:
I think this has been implemented on dlg Replace. @dannyparsons this could closed.

dannyparsons:
We haven't actually done this part yet and Roger will be very pleased you're raised this again, it's still to do!

stevenndungu:
Ok great. I was looking at both functions; `na.locf` and `fill` which sounds easy and quick to put this in place. I will be OK to have  `na.locf`  done now that you seemed to have agreed on with Roge unless otherwise.

dannyparsons:
Thanks. What do they both return? I think my issue before was the one function was done on a whole data frame, and returned a data frame, whereas it's more sensible for us to use a function that works on a column.

stevenndungu:
`fills`  replaces missing values in, using the previous entry. This can either be  backwards or forward approach.  This is to mean  that for the forward approach it  replaces subsequent missing values using the  first value encountered till the next value in the column that is found to be non-missing, again the next non- missing value is used to replace missing values to the next non-missing value all the way to the bottom of the column. For the backward approach the vice vice-versa happens.

For  `fill`;
`fill(data, ..., .direction = c("down", "up"))`
-this can take multiple columns at the same time doing the same thing through the columns intended.
-updates the existing data frame

`na.locf`
`na.locf(object, na.rm = TRUE, ...)` Does a similar thing to  `fill` with change of parameter  `fromLast` from `TRUE` to `FALSE`.
-This can only work with one column at a go. ie works with a vectors
[...]

PR #3139, opened 2017-05-30
Dlg replace

stevenndungu:


stevenndungu:
Note: @dannyparsons, Just correction  from what I had written on  issue #1779 on na.locf function ` "if the first or the last value of the vector is missing ,it reverses using the opposite approach to fill up the missing value which is a plus for this over fill"` ...  While implementing I have seen that that is not the case, what happens is that na.rm is TRUE by default it thus truncates missing values not replaced by the applied aproach. It thus works in a similar case as fill.

This is what i mean.

We can use this data set for testing:
[DATA_MISSING.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/1040636/DATA_MISSING.xlsx)
`> bz <- c(NA,NA,NA,4,NA,NA,5,NA,5,NA,NA,NA)
> bz
 [1] NA NA NA  4 NA NA  5 NA  5 NA NA NA
> na.locf(bz)
[1] 4 4 4 5 5 5 5 5 5
> na.locf(bz, fromLast = TRUE,na.rm=F)
 [1]  4  4  4  4  5  5  5  5  5 NA NA NA
> na.locf(bz, fromLast = FALSE,na.rm=F)
 [1] NA NA NA  4  4  4  5  5  5  5  5  5`

Everything is working well as it should be. and the dialogue is ready for testing

dannyparsons:
I think that's fine and we are not as interested in that functionality anyway. We don't want it to do the truncating so as long as we always use `na.rm = FALSE` it should be fine.

stevenndungu:
Great, I actually set `na.rm = FALSE` on the function.

lilyclements:
Reviewing/making amendments to this now

stevenndungu:
Point 1: Not sure when that is happening exactly. Will skype you so that I can be able to see that.
Point 2: I made a mistake on the rdo to be checked when oldmissing rdo is checked. Thanks for noticing that.
Point 3: You are the one who implemented that, so I did not tamper with that since I thought you had a good reason of doing that. Check line 213 of the dialogue

stevenndungu:
@lilyclements  I have fixed point 2

lilyclements:
Point 1: If you open the dialog, just put a variable in the receiver and press OK (for the first time opening the dialog) then there is an error:
>   argument "new_value" is missing, with no default


Point 2: Great, good job for fixing


Point 3: I knew you didn't cause that issue, it was just one I noticed when using it.
Great, thanks. I remember why we've implemented that now, you can do an interval with a factor. Makes perfect sense.

Just point 1 then it should be ready!

lilyclements:

stevenndungu:
@dannyparsons  this is now conflict free.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -601,103 +602,119 @@ data_object$set("public", "replace_value_in_data", function(col_names, rows, old
...
     }
     if(!done) {
-      replace_rows[is.na(replace_rows)] <- FALSE
-      if(sum(replace_rows) > 0) {
-        if(filter_applied) {
-          replace_rows <- replace_rows & curr_filter
+      if(locf){
+        
+        print(nrow(private$data))
+        print(length(my_data))
```
dannyparsons: can you remove these prints

dannyparsons:
Is this done now?

stevenndungu:
Nop, still has one developer error. Figuring out how it comes about and why?This  happens when you run  either of last radio buttons, check misssing radio button, run and reopen the dialogue. The developer error occurrs.Else other things are okay.

dannyparsons:
@stevenndungu here I fixed the issue https://github.com/africanmathsinitiative/R-Instat/pull/3281/commits/6188aa3cc82d2c8f9b749d3e9452d235669f5238
The `locf` and `from_last` parameters we not being removed when you unchecked from above/below, so the conditions were never satisfied for other radio buttons.
Also I noticed From Above/Below were the wrong buttons and I changed the text of the buttons.

stevenndungu:
Aaah! I see, I was wondering why there was an issue. Thanks for helping in sorting that out. I

Issue #3136, opened 2017-05-29
Calculations dialog


dannyparsons:
Try should be disabled when the expression is blank

dannyparsons:
Same needed for the Enter dialog.

shadrackkibet:
i am working on this.

shadrackkibet:
@dannyparsons this can now be closed.

PR #3144, opened 2017-05-31
dlgCalculations

shadrackkibet:
@dannyparsons  this solves one of #3136.


dannyparsons:
Great, are you doing the calculator one on this branch as well?

shadrackkibet:
@dannyparsons  i have done both now on this branch . This now fixes both #3144

instat/dlgCalculator.vb
```diff
@@ -81,6 +79,8 @@ Public Class dlgCalculator
         ucrCalc.ucrSaveResultInto.SetDataFrameSelector(ucrCalc.ucrSelectorForCalculations.ucrAvailableDataFrames)
         ucrCalc.ucrSelectorForCalculations.Reset()
         ucrCalc.ucrSaveResultInto.SetValidationTypeAsRVariable()
+        ucrCalc.cmdDoy.Enabled = False
+        ucrCalc.cmdDek.Enabled = False
```
dannyparsons: These aren't the Try buttons, these are just disabled temporarily.
Also, I'm expecting the change you make for Try to be on the calculator control itself, not this dialog, so that it works the same for any dialog which uses this control.

shadrackkibet: I will have to work on this with @Lunalo

shadrackkibet:
@dannyparsons  i have fixed this on the ucrcalculator control and also sorted out reset issue of the controls.

instat/dlgEnter.vb
```diff
@@ -50,6 +50,8 @@ Public Class dlgEnter
         ucrSaveEnterResultInto.SetDefaultTypeAsColumn()
         ucrSaveEnterResultInto.SetDataFrameSelector(ucrDataFrameEnter)
         ucrSaveEnterResultInto.SetValidationTypeAsRVariable()
+        cmdTry.Enabled = False
+        ucrInputTryMessage.Enabled = False
```
dannyparsons: This doesn't need to be disabled, only the button.

instat/ucrCalculator.vb
```diff
@@ -33,8 +33,17 @@ Public Class ucrCalculator
     Public Sub InitialiseControl()
         ucrInputCalOptions.SetItems({"Basic", "Maths", "Logical and Symbols", "Statistics", "Strings (Character Columns)", "Probability", "Dates", "Rows"}) ' "Rows" is a temp. name
         ucrReceiverForCalculation.Selector = ucrSelectorForCalculations
+        ucrInputTryMessage.Enabled = False
+        cmdTry.Enabled = False
```
dannyparsons: Same for this, and also I'm not seeing where this gets enabled again?

instat/dlgEnter.vb
shadrackkibet: will remove this line now.

instat/ucrCalculator.vb
shadrackkibet: @dannyparsons i am not sure where this should be enabled on this control. could you explain to me where this should be enabled?

dannyparsons: As soon as the receiver becomes non empty it should be enabled. Therefore it can be enabled on the receiver changing when it is non empty. And if it's empty it should be disabled.

dannyparsons:
so we still need to enable/disable when the receiver changes?

shadrackkibet:
Exactly, that is what we need to do now which i am not yet sure how this can be achieved in the control. could you help me on this as i learn at the same time?

dannyparsons:
In the event when the receiver changes, just check if it is empty or not.

shadrackkibet:
@dannyparsons i think this is what we need here, have a look at it.

dannyparsons:
Great this looks good.

PR #3150, opened 2017-06-02
Sdg Themes

stevenndungu:


stevenndungu:
For a start, this is how I have designed the X axis tab with slight  changes to be made on arrangement of controls since I will be adding more controls on the same tab. 
image
Now, from the defined parameters values shown above, here is how a barchart would appear. Plotted using Dlg Barchart.
image
 
And this is how the code looks like for all the controls added so far...
`# dlgBarAndPieChart code generated by the dialog Bar and Pie Chart
diamonds_temp <- InstatDataObject$get_data_frame(data_name="diamonds")
last_graph <- ggplot2::ggplot(data=diamonds_temp, mapping=ggplot2::aes(x=cut, fill=clarity)) + theme(axis.text.x=element_text(angle=45, face="italic", colour="blue", lineheight=0.6, size=12, hjust=0.5, vjust=0.5), axis.text.x.top=element_text(size=12, hjust=0.5, vjust=0.5, lineheight=0.6, angle=45, face="italic", colour="blue"), axis.ticks=element_line(colour="yellowgreen", linetype="dotted", size=0), axis.ticks.x=element_line(size=0, linetype="dotted", colour="yellowgreen")) + ggplot2::geom_bar()
InstatDataObject$add_graph(data_name="diamonds", graph=last_graph, graph_name="last_graph")
InstatDataObject$get_graphs(graph_name="last_graph", data_name="diamonds")`

the theme from this subdialogue overwrites the already defined themes in the sdgPlots dialogue...

stevenndungu:
Questions worth a discussion:
1. Should I leave out some of the parameters that for some reasons I feel they may not be impacting a lot on the a plots ? Or exhaustively use almost all parameters a theme needs. For instance, `axis.text.x.top` and `axis.text.x` have similar parameters with some of the parameters overriding one from the other parameter. I  think this is because of inheritance between the two parameters.
2. Controls of the sub-dialogue are not directly linked to the base function or base operator of the main dialogue. This makes resetting of the sub-dialogue controls not functional for now. Should we have setdefault sub for this sub-dialogue? or create a sub with defaults then link it to the default of the main dialogue that will be calling it? Not sure how best to do this
3.

stevenndungu:
Here is an example with a boxplot with labels at 90 degrees and blue colour
image

rdstern:
Brilliant.  I am very pleased and impressed!

PR #3161, opened 2017-06-02
dlgextremes

shadrackkibet:
This is in progress.


shadrackkibet:
@stevekogo i have done some polishing up of the code here and also disabled  some controls  which are not yet functioning now.Have  a quick review before @dannyparsons checks.

lilyclements:
Reviewing/making changes to this now. Make sure you've sent all commits!

lilyclements:
I've made a lot of bug fixes here which are outlined and ready to be merged here:
https://github.com/shadrackkibet/Instat/pull/87#pullrequestreview-42855749
So address that before the following

I have two questions:
 * Should we autofill the optional receivers? (i.e., the Year and Station ones). I don't think we should. I didn't know that they didn't have to be filled when running it, until I looked in the code. @dannyparsons what are your thoughts on this too?
 *  Why do we have a date and DOY receiver options?

rdstern:
1. I am happy for Station to be filled, if it is there as a column.  Sometimes the data frame will be for a single station and there may therefore not be a station to be filled.
2. I am surprised that Year is optional.  Usually we will either give the extreme each year (could be a shifted year, but that is automatic).  Or it will be the extremes that exceed a threshold, and we would then usually want to output the year as a column in the resulting data frame.
3. On the date column, that is not strictly necessary, but the R functions may choose to use it.  The day of year will sometimes be saved, i.e. we sometimes what to know when in the year we had the extreme as well as how large it is.

shadrackkibet:
@lilyclements  i have fixed the reseting of the ucrinput to address point 3 above which is now fine. For Point 1 & 2 have been well answered by roger above.You can now check the reset of ucriput and some few minor edits which i found necessary.

lilyclements:
Excellent work! And thanks @rdstern for answering the questions. I just wanted to double check on those things.

@rdstern to your second point: This dialog does currently allow the user to only input the station and there is code written for this instance.  Do you think this is needed? I think it's nice to have the maximum and minimums for the different stations without year taken into account, however, my experience with crops work is very very very minimal. Should this be removed?

I'll re-review now.

lilyclements:
Working well. Out of interest, what should "Peaks" do? From what I can see it acts as a filter to subset the data to where rainfall is greater than 15mm (an option which I assume can change in the future when the other options are enabled?). I was wondering what the benefit of this is?

Also the checkbox should be disabled since it doesn't yet run any code

rdstern:
If the code is there, then (with multiple stations) it could be useful to have the overall extremes.  These would be an example where you would want the date to be available as well, so you know the value for the maximum and also when it occurred.

We could take that further.  The peaks over threshold already would let us choose all the values more than a certain value (e.g. 100mm) with their associated dates.  We could be asking not just for one maximum, but perhaps for the x largest values, e.g. the 5 greatest rainfalls in each record!

rdstern:
There are two ways of studying extremes.
a) Taking the maximum (minimum) each year.
b) Defining a threshold and taking all the values more than this threshold.  Usually with tropical rainfall this would be a lot bigger than 15mm - say 40, or 50 or 60mms.

The extreme value theory has the generalised extreme value distribution for the maxima and the generalised Pareto distribution to model the peaks.

That will be the second half of the work - for later.  The R extRemes package is designed to model extremes once our Prepare > Extremes dialogue has calculated them.

Note also that there will be further options for the Prepare > Extremes dialogue, e.g. 

a) What is the maximum  3 day total each year.  Answer, for the Prepare > Transform first to get the running 3-day totals.  Then use the Prepare > Extremes
b) What is the extreme in a single rainy season, between March and May each year.  Answer use the filter to just have March to May, then use the Prepare > Extremes dialogue.

shadrackkibet:
@lilyclements .Peaks is not yet functioning fully because of a feature we lack from the calculation system.This has been reported as an issue by @stevekogo  at #3201.We obtained a peaks_data_frame for now which can then then be improved later once the feature is added on the calculation system so allow us select specific columns after applying filter.

shadrackkibet:
@rdstern is giving  good explanation here of obtaining extremes i really like this. I think this is what we are working towards on this dialog. The second part of using peaks is what is not yet fully functional which is due to the lack of a feature in calculation system. I am happy that we are making good progress on this.

dannyparsons:
My understanding is that the only extra feature needed in the calculation system is to select a subset of columns, which you want to do after filtering the peaks to only show a few columns.
Is peaks working on the dialog apart from choosing a sub set of columns?

lilyclements:
Yes, it's not choosing a sub set of columns yet, it's more filtering (all instances?) where Rain > 15mm for now.
Thank you @shadrackkibet and @rdstern for the descriptions!

shadrackkibet:
just before @dannyparsons  reviews this i will make an amendment on this to enable the two  controls work as required so that the value >15mm(threshhold) can be changed from the control with a given operator.

rdstern:
Perhaps at the same time you could make the default into 40 rather than 15?  If you are giving a label of mm, then this could be (mm/degrees), because it could be occasions more than 40 degrees (say) if temperature data.

dannyparsons:
Couldn't this be for any variable? In which case we shouldn't give a label for units.

shadrackkibet:
@dannyparsons  i have not made some changes to enable the two controls (operator and threshold value controls).  Have a look at how i have implemented this to work i hope this is now fine.

dannyparsons:
Do you mean you have now made some changes?

shadrackkibet:
Yes I mean I have  made some changes on this dialog.

dannyparsons:
OK thanks I'll have a look at this.

shadrackkibet:
@dannyparsons this is still waiting you to review.

dannyparsons:
This is looking pretty good. I've made a pull request with a correction to the conditions for the main panel. Also a few comments to fix:

1. When I'm on Min/Max the Operator label is always visible.
2. On peaks I think the default new data frame name can be "Peaks" there's no need to cal it "data_frame"
3. The threshold selection is confusing because I'm not sure how the operation is going to be applied. It would be much clearer if it was all on one line i.e. Values >= 70
where `Value` is a label, `>=` is the combo box and `70` is the text box. `>=` should also be the default.
5. An option for ignoring missing values would be good for the Max/Min option otherwise I could get missing for each year.

@rdstern Should this only be yearly extremes? I can only do by Year because it's a year receiver. Should we be able to do other periods here, or call this dialog yearly extremes?

dannyparsons:
If you implement 3. then 1. isn't an issue any more. If 4. is easy to add then add it, otherwise it isn't urgent.

shadrackkibet:
@dannyparsons  i have sorted out 1,2 and 3 above on my latest commit. This is now ready for you.

Issue #1924, opened 2016-10-19
File > Import from OpenRefine


lilyclements:
This dialog is to import the dataset from OpenRefine.

This dialog will consist of
1. An input box to input the name of the Project
2. A combo box for 'Format', where the user can specify the format which their project is in in OpenRefine
3. Another Input combo box where the user can put what they wish to call their new data frame on R-Instat. (perhaps with Sheet1 as the default?)
4. The base buttons

Again, this is using 'rrefine' which is yet to be added into the programme.

The code is
refine_export(project.name = "", format = "")

Where
project.name = the name of the project on OpenRefine, specified in 1.
format = formatting of the file. Default is CSV.

e.g. refine_export(project.name = "MyDataFrame_cleanup", format="csv")

Issue #1914, opened 2016-10-17
Organise - linking to OpenRefine and developing the Organise menu further


rdstern:
This is not a high priority.  But it could be very useful, particularly if someone becomes interested in this area.  

It could also be very easy to implement, because the **rrefine** package is designed simply to export data to open refine, and then to be able to transfer results back from OpenRefine to R data frames. This could be implemented either in the File menu or in the Organise > Check data menu - or both.

OpenRefine describes itself as "A free, open source, power tool for working with messy data."  Now we also want our own Organise menu to be able to handle messy data, but OpenRefine goes a lot further than we could hope to do ever.  So in a follow-up message I suggest what we might still do in the Organise menu.  That will not negate the value of the links to OpenRefine.

It is possible that Danny/David might suggest that OpenRefine be given a higher priority because of the corruption work.  It would do some of the work of cleaning messy data sets from the web-crawling software using the internet.  In the current corruption project this step is being done by the Cambridge team, but this could do that sort of cleaning for data from other countries, and that is one of the longer-term aims of the corruption project.

In addition I like the 3 videos that describe the use of the software.  They should be of general interest, even though they were produced for when it was called GoogleRefine.  These videos are interesting in their own right.  They are also examples of videos of a sensible length, and very clear, and with subtitles that help the understanding - and would make translation easy.  They don't look difficult to produce and might be examples that we could follow, for R-Instat videos. Here is a link to the OpenRefine web page and the three videos: http://openrefine.org/ 

Roger

rdstern:
I am keen that R-Instat is as good as R can be in handling messy data.  And we are getting close with the Organise menu.
One area to be completed is the facilities to handle text (character) columns.  This will be partly through making the **Organise > Column: Calculate >  Calculations** dialogue as powerful as possible, and then by completing the **Organise > Column: Text** dialogues.

I particularly would like R-Instat to be able to cope with questionnaires that have "open" answers.  They could be one or more sentences and we want to extract the key points from them in successive columns.  (There will still be some messy things we can't do, shown in the OpenRefine second video and that's fine.  We just need to know the limits of what we can do, and then be able to use more appropriate software if the data go beyond those limits.  And we need to document what we can't do, as well as what we can do!

 I hypothesise that (with one exception) we don't need to do anything other than is in the **stringr** package.   And the missing features there are largely the missing dialogues in the **Organise > Column: Text** menu.  However, before I get to the exception, notice that **stringr** does have commands:

[...]

PR #3183, opened 2017-06-05
Import/Export Open refine

### LINKED ISSUE #1923: File > Export to OpenRefine
This dialog and 'Import from OpenRefine' are very similar. 

The idea is that you can export your dataframe to OpenRefine where you can make advanced changes that R-Instat cannot offer. We do not need to worry about coding anything in OpenRefine, only this dialog to export to OpenRefine, (and then another dialog which I will write up to Import From OpenRefine)

This dialog will consist of
1. A Dataframe selector, where the user can select which dataframe they wish to export
2. An input box where the user can note what they want the new name to be in OpenRefine.
The default for this could be "TheCurrentDataSetName _ cleanup" or something.
3. A check box to open the browser for OpenRefine. This is default unticked.
4. The base buttons

We could perhaps have an option to convert to CSV file too since that file type seems to work best on OpenRefine?

The R-Code for this is using the package 'rrefine'. You will need to work with @dannyparsons to get this added in

The code is
refine_upload ( file = "", project.name = "", open.browser = )

Where
File = the dataframe in 1.
project.name = the new name given in 2.
open.browser = TRUE or FALSE, depending whether 3. is checked or not.

_e.g. refine_upload(file = "MyDataFrame.csv", project.name = "MyDataFrame_cleanup", open.browser = FALSE)_

lilyclements:

For this to work, Open Refine has to be downloaded: 
https://github.com/OpenRefine/OpenRefine/releases/tag/2.7-rc.2

Otherwise you get an error which says:
```
Error in curl::curl_fetch_memory(url, handle = handle) : 
  Couldn't connect to server
```

Once downloaded, open the `.exe` file and then you can run it fine, and it opens in an internet browser.

Should a "warning" be on that open-refine has to be downloaded? I don't know how we can implement this "warning" if so. Ideas are welcome!

This will fix #1923 #1924 #1914


stevenndungu:
Will be testing this dialogue

lilyclements:
@stevenndungu Did you review this? It would be nice for this to get into 0.3 since the SSD will be looking at 0.3 and Open Refine was a suggestion from them

stevenndungu:
Yes, I reviewed.  It is working fine...Only one problem when I use survey data set. Let me change branch the get to you..

lilyclements:
This dialog converts the file to a csv in order to then export to Open Refine. @stevenndungu was having issue with an RDS data set and when we explored further we saw that this is a common issue that RDS files cannot have the write.csv function used on them. 
@dannyparsons @stevenndungu how do you think we should proceed with this dialog for an RDS?

dannyparsons:
I'm not sure I understand this correctly. Doesn't it just take a data frame from the instat object and export it as a csv? Where does RDS come into it?

lilyclements:
It takes a data frame and turns it into a csv and then exports it to Open Refine.
However, the RDS files are bringing up an error in the `write.csv` code part.

e.g. with my Damango data frame which is an RDS:

> Error in is.data.frame(x) : object 'Damango' not found

dannyparsons:
ok, but you shouldn't be exporting the file that you imported into R-Instat. You should be getting the data frame from within the instat object and exporting that as a csv. If you're exporting your Damango RDS file that you imported then it's not the same as the Damango data frame in the instat object. Or am I missing something about this?

lilyclements:
I'm trying to convert my data frame to a csv and then it can be exported. The stage which is causing issues is the converting-to-csv-part. How do I get that data frame from the instat object to convert to a csv?

dannyparsons:
If the data frame is in a receiver then it's just `GetVariables()`. 
Manually it's by running `InstatDataObject$get_data_frame(...`

lilyclements:
Thanks @dannyparsons I fixed the problem

instat/dlgExportToOpenRefine.vb
```diff
@@ -92,11 +102,9 @@ Public Class dlgExportToOpenRefine
...
     Private Sub ucrOpenRefineDataFrame_ControlContentsChanged() Handles ucrDataFrameOpenRefine.ControlContentsChanged
         NewDefaultName()
-        clsDefaultWrite.RemoveParameterByName("x")
         clsDefaultWrite.RemoveParameterByName("file")
         clsDefaultRefine.RemoveParameterByName("file")
         clsDefaultRefine.RemoveParameterByName("project.name")
-        clsDefaultWrite.AddParameter("x", ucrDataFrameOpenRefine.cboAvailableDataFrames.SelectedItem)
         clsDefaultWrite.AddParameter("file", Chr(34) & ucrInputDatasetName.GetText() & ".csv" & Chr(34))
         clsDefaultRefine.AddParameter("file", Chr(34) & ucrInputDatasetName.GetText() & ".csv" & Chr(34))
         clsDefaultRefine.AddParameter("project.name", Chr(34) & ucrInputDatasetName.GetText() & Chr(34))
```
stevenndungu: Why are you removing parameter "file" and adding it again in the same sub

lilyclements: It's being added manually. I am removing the parameter name before adding it again incase it otherwise adds in multiple "file = " each time the df is updated

dannyparsons: There's no need to remove a parameter if it's definitely going to be replaced in the same part of the code. It will always override.

lilyclements: Well clearly I'm forgetting how the "old system" works. I'll fix that now!

lilyclements:
@stevenndungu I've made the necessary changes

lilyclements:
@stevenndungu is this ready for Danny to review?

lilyclements:

Issue #3138, opened 2017-05-29
Consistent display of separator characters etc. in Text dialogs


dannyparsons:
@africanmathsinitiative/developers 
I noticed that Text dialogs don't have consistent ways of displaying characters in combo boxes. For examples, I've seen "Space" vs "Space ( )", and "Underscore" vs "Underscore _" in different dialogs.

I'm happy with the second display in each case so we show the name of the symbol and the symbol itself. Unless there's disagreement, could this be changed to be implemented consistently throughout the Text dialogs?

shadrackkibet:
I am working on this now.

PR #3189, opened 2017-06-06
consistency of symbols in textdialogs

shadrackkibet:
@dannyparsons this is ready for you. I found only one dialog not consistent here i.e transform text.  can this comboboxes be changed to non-editable because they contain specific symbols required here?


dannyparsons:
I believe the pad value can be anything so should it be non editable?

shadrackkibet:
sure , pad value can be anything therefore this should then be editable. This now this fixes #3138

PR #3199, opened 2017-06-06
fixed bug with ucrColours passing correct hex value for ggplot

dannyparsons:


dannyparsons:
@maxwellfundi fyi a bug fix on the colour control for ggplot

PR #3206, opened 2017-06-07
Fixing reset for summaryBar dialog

maxwellfundi:


maxwellfundi:
@lilyclements  this now fixes the reset problem on the dlsSummaryBarorPiechart

lilyclements:

instat/dlgSummaryBarOrPieChart.vb
```diff
@@ -125,6 +125,7 @@ Public Class dlgSummaryBarOrPieChart
         clsLabsFunction = GgplotDefaults.clsDefaultLabs.Clone()
         clsXScalecontinuousFunction = GgplotDefaults.clsXScalecontinuousFunction.Clone()
         clsRFacetFunction = GgplotDefaults.clsFacetFunction.Clone()
+        clsBaseOperator.RemoveParameter(clsRCoordPolarParam)
```
dannyparsons: This issue is caused by SetDialogOptions above which is setting things based on the radio button values. But the radio buttons do not have correct values until the R code has been set to them.
Therefore, the solution to this is to only call SetDialogOptions once the R code has been set. Then this line won't be needed since the wrong parameter will not be added.

maxwellfundi:
@dannyparsons , i fixed this too. Why was it removed from the menu?

rdstern:
It shouldn't be removed, but changed in location.  The idea is that both for tables and graphs you can choose to summarise the data first and then produce the neat picture or table.  Or you can proceed in one go.
We also heard that the bar chart and bar chart from Summaries caused some confusion in Maseno recently?  

So the Bar chart from Summary is now under General > Use Summaries.  Currently it is just "your" dialogue, but there will soon (I hope) be another button for Tables.

If you wish we could duplicate the dialogue also back where it was.  I would then like a line before that one to indicate that it is slightly different.

maxwellfundi:
Okay I see. 
I dont think duplicating it would be a bad idea. When is the extra button for table needed?

rdstern:
The extra button could be any time.  Initially it should be disabled.
I am not sure when Danny/David will add the extra feature to the tabulation.  I assume it will be for the first release

rdstern:
Later I would also expect that there would be more chart types also from the summary, including lines and possibly points.  Whether these are more buttons, or there would just be 2 buttons, namely Table and Graph is still to be discussed.

PR #3212, opened 2017-06-08
Implementing UcrAxes for sdgplots 

maxwellfundi:


dannyparsons:
I've resolved the conflict but I think you need to check the SetRCode on scatter plot as this got modified.

dannyparsons:
I've merged my changes from this branch now into the master. The axes should be working for continuous scales now. The main thing needed is to add the extra parameters for the scales function. There is also some tidying on the layout and positioning of labels needed.

Then make sure all main dialog pass through the correct functions. e.g. I noticed boxplot was passing an x continuous scales function but not one for y.

If this is done then this tab will be ready for the release.

PR #3213, opened 2017-06-08
dlgStringHandling Enhancement.

shadrackkibet:
@africanmathsinitiative/developers this is in progress.


shadrackkibet:
@africanmathsinitiative/developers this dialog is now working fine and it is ready for review. 

NOTE:
There is an issue with resetting of radiobuttons and a checkbox on reopen which i have fixed it temporarily.kindly, someone help me fix this.

lilyclements:
Reviewing this now

shadrackkibet:
@lilyclements  on your point 1 above i realised that when you use a regex expression which is not sensible you get a such a bug. I don't know how this can be avoided for now. I think the regexkeyboard still need to be imroved to avoid such combinations.

lilyclements:
Yes, I wouldn't mind if it came up with the error, but when it crashes it's not ideal!
We can't have it crashing on 0.3.0 but will think of an alternative after this release for the future. For the time being, can we disable the keyboard/regex options just so people can see what can potentially be done here in the future for when that's fixed?

shadrackkibet:
@lilyclements  can you test this?

PR #3235, opened 2017-06-09
Package names: Unstack, Population Pyramids

lilyclements:


instat/dlgUnstack.vb
```diff
@@ -37,6 +37,7 @@ Public Class dlgUnstack
     Private Sub InitialiseDialog()
         ucrFactorToUnstackReceiver.Selector = ucrSelectorForunstack
         ucrColumnToUnstackReceiver.Selector = ucrSelectorForunstack
+        ucrBase.clsRsyntax.SetPackageName("reshape2")
         ucrBase.clsRsyntax.SetFunction("dcast")
```
lilyclements: Just want to confirm: There is reshape2 and data.table.
Which package are we using for dcast? They both seem to have the same parameters

dannyparsons: yep I think that's right

PR #3238, opened 2017-06-09
Lunalo ucr dist update control

dannyparsons:


dannyparsons:
@Lunalo I have made a few small changes to your work but otherwise it looked good. 
I noticed the random samples dialog is very wide, do we need this?

PR #3244, opened 2017-06-10
One var freq rdos

AlexSananka:
@dannyparsons This  are tiny edits on oneway  and it's ready for you.


AlexSananka:
@dannyparsons  the above two commits fixes a bug on dlgonevarFitModel.
Currently when you pick a factor there is a checkbox that converts it to numeric,so when you don't check  it and click Ok,it bugs because it is still as.vector.
So I have made it be like whenever the selected variable is a factor the chechbox is checked and so it converts it to numeric.

instat/sdgOneVarFitModDisplay.vb
```diff
@@ -58,27 +58,27 @@ Public Class sdgOneVarFitModDisplay
...
         If rdoPlotAll.Checked Then
             clsRplotFunction.ClearParameters()
-            clsRplotFunction.SetRCommand("plot")
+            clsRplotFunction.SetRCommand("fitdistrplus::plot")
             clsRplotFunction.AddParameter("x", clsRFunctionParameter:=clsModel)
             frmMain.clsRLink.RunScript(clsRplotFunction.ToScript(), 3)
         ElseIf rdoPPPlot.Checked Then
             clsRplotFunction.ClearParameters()
-            clsRplotFunction.SetRCommand("ppcomp")
+            clsRplotFunction.SetRCommand("fitdistrplus::ppcomp")
```
dannyparsons: This is not right, why have you done this instead of setting the package name?

dannyparsons:
Thanks for spotting these issues. I'm fine with temporary solutions but only if they make sense.

If we are to prevent unchecking then the check box should be disabled. Currently when I uncheck it rechecks again, this will be very frustrating behaviour for users!

You also set the R command and included the package in it, why did you not set the package name? Also, plot is not a fitdistrplus function, it was crashing with this package.

I've made these few corrections on another branch and will merge this.

AlexSananka:
I have been away for a while,sorry for the late response and thanks for making this corrections

PR #3251, opened 2017-06-10
Themes layout

lilyclements:
@dannyparsons this is ready to review. I've edited the themes layout here for the ucr controls for hte themes and for sdgThemes


dannyparsons:
Sorry had to merge something else first and then resolve a conflict here. Might need to check it's still ok.

PR #3261, opened 2017-06-10
Disabling ThreeVariable Options in the frmMain

AlexSananka:


AlexSananka:
Here I have disabled threevar options that are above frequencies and Set the rcode for panels to be `ucBase.ClsRsyntax.ClsBaseFunction`

AlexSananka:
@dannyparsons this is ready

PR #3277, opened 2017-06-11
Transform text - function name parameter fix

lilyclements:


instat/dlgTransformText.vb
```diff
@@ -335,11 +335,11 @@ Public Class dlgTransformText
                 ucrBase.clsRsyntax.SetBaseRFunction(clsConvertFunction)
                 Select Case ucrInputTo.GetText
                     Case "Lower"
-                        ucrBase.clsRsyntax.SetFunction("str_to_lower")
+                        ucrBase.clsRsyntax.SetFunction("stringr::str_to_lower")
```
lilyclements: Is this an okay way to fix this? I don't want to set it to ucrBase.clsRsyntax

dannyparsons: As a temp fix yes, but can also be solved by having a separate function which is set as the base function.

lilyclements: Okay, great. I'll properly fix it after 0.3 release.

lilyclements: Realised this isn't needed since it's being called elsewhere. THere's also a lot of developer errors in this to be fixed, I'll look into these tomorrow.

PR #3283, opened 2017-06-12
Fixing condition for setdefaults

Lunalo:


Lunalo:
Still in progress

Lunalo:
@dannyparsons 

I have fixed this but the dialog is doing automatic resseting on reopening . If it takes long to fix it, we can do a temprary fix in the set defaults. thanks.

dannyparsons:
ok do that for now, if I can I will try to fix it as well.

Lunalo:
I think I have now implemented that

PR #3289, opened 2017-06-12
Duplicate columns - bug fix

lilyclements:


instat/dlgDuplicateColumns.vb
```diff
@@ -94,6 +96,10 @@ Public Class dlgDuplicateColumns
         End If
     End Sub
 
+    Private Sub ReopenDialog()
+        ucrInputColumnName.bAllowNonConditionValues = True
+    End Sub
```
lilyclements: Adding on ReopenDialog() to prevent a developer error occurring when reopening this dialog

dannyparsons: Should this not be done on initialise then if it's needed?

lilyclements: If I set on initialise I still get the error when reopening?

dannyparsons: Ok with this for now then but should come back to investigate.

dannyparsons: Can you add a comment that this a temp fix?

lilyclements: Sure, added the comment :)

PR #3291, opened 2017-06-12
Convert Columns - bug fix

lilyclements:


instat/dlgConvertColumns.vb
```diff
@@ -130,9 +130,13 @@ Public Class dlgConvertColumns
...
+    Private Sub ucrChkSpecifyDecimalsToDisplay_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkSpecifyDecimalsToDisplay.ControlValueChanged, ucrPnlConvertTo.ControlValueChanged
+        If rdoFactor.Checked OrElse rdoOrderedFactor.Checked Then
+            If ucrChkSpecifyDecimalsToDisplay.Checked Then
+                ucrReceiverColumnsToConvert.SetDataType("numeric")
+            Else
+                ucrReceiverColumnsToConvert.SetIncludedDataTypes({"integer", "numeric", "character", "ordered", "factor"})
+            End If
         Else
             ucrReceiverColumnsToConvert.SetIncludedDataTypes({"integer", "numeric", "character", "ordered", "factor"})
         End If
```
lilyclements: We had to only accept numeric if ucrChkSpecifyDecimals was checked. Added that also rdoFactor, rdoOrdered factor need to be checked too

lilyclements: Okay so before I edited this branch, it was:
```
 If ucrChkSpecifyDecimalsToDisplay.Checked Then
                ucrReceiverColumnsToConvert.SetDataType("numeric")
            Else
          ucrReceiverColumnsToConvert.SetIncludedDataTypes({"integer", "numeric", "character", "ordered", "factor"})
End If
```


I added the `If rdoFactor.Checked OrElse rdoOrderedFactor.Checked Then` to this because at the moment if you check the check box and then go onto another rdo, such as numeric, the checkbox is no longer visible but the user can't put factors in the receiver!

Are you saying I should remove all of this completely? :)

dannyparsons: yep I think so. The check box can be visible when factor/ordered is checked, and I don't think it should change the data type of the receiver

lilyclements: Great, the checkbox being visible is set in Initialise. I've removed this sub now and it works well. This is ready to merge

PR #3293, opened 2017-06-12
various bug fixes

dannyparsons:
@maxwellfundi here I have linked the selector on the main dialog to the facet selector on the sub dialog so that the facet variables can be included if we do a temporary stack.
I noticed that none of the main dialog are linked to the layer options dialogs, the setup lines are commented out. Why is this not done?


maxwellfundi:
@dannyparsons I am looking at this now.

PR #3318, opened 2017-06-13
Split textand climdex indices

AlexSananka:
@dannyparsons This is ready to go in.

### DIFF
--- a/instat/sdgClimdexIndices.vb
+++ b/instat/sdgClimdexIndices.vb
@@ -184,6 +184,9 @@ Public Class sdgClimdexIndices
         dctInputFreqPairs.Add("monthly", Chr(34) & "monthly" & Chr(34))
         ucrInputFreq.SetItems(dctInputFreqPairs)
         ucrInputFreq.cboInput.SelectedItem = "annual"
+        ucrInputFreq.SetRDefault(Chr(34) & "annual" & Chr(34))
+        ucrInputFreq.SetDropDownStyleAsNonEditable()
+        ucrInputFreq.bAllowNonConditionValues = True
 
         ucrInputTempQtiles.SetParameter(New RParameter("x"))
         ucrInputTempQtiles.SetValidationTypeAsNumericList()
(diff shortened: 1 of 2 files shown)


instat/sdgClimdexIndices.vb
```diff
@@ -184,6 +184,9 @@ Public Class sdgClimdexIndices
         dctInputFreqPairs.Add("monthly", Chr(34) & "monthly" & Chr(34))
         ucrInputFreq.SetItems(dctInputFreqPairs)
         ucrInputFreq.cboInput.SelectedItem = "annual"
+        ucrInputFreq.SetRDefault(Chr(34) & "annual" & Chr(34))
+        ucrInputFreq.SetDropDownStyleAsNonEditable()
+        ucrInputFreq.bAllowNonConditionValues = True
```
dannyparsons: If it's non editable we don't need to allow non condition values. Does it bug without this?

instat/sdgClimdexIndices.vb
```diff
@@ -223,6 +226,8 @@ Public Class sdgClimdexIndices
         dctNumericPairs.Add("GSL_sum", Chr(34) & "GSL_sum" & Chr(34))
         ucrInputGSLMode.SetItems(dctNumericPairs)
         ucrInputGSLMode.cboInput.SelectedItem = "GSL"
```
dannyparsons: Just noticed this line shouldn't be here. We shouldn't be setting the value. Can you remove this?

instat/sdgClimdexIndices.vb
AlexSananka: yes it bugs without setting the defaults

instat/sdgClimdexIndices.vb
AlexSananka: yes sure

AlexSananka:
@dannyparsons  I have added package names and sorted out the above highlighted issues.

dannyparsons:
Is it still giving the developer error?

AlexSananka:
No I have now fixed them.

dannyparsons:
I mean, is it only giving a developer error when you type something in different from the list?

AlexSananka:
I have now do away with making it non editable and fix it by setting the "R default"

PR #3319, opened 2017-06-13
Dlg correlation

stevenndungu:
@dannyparsons 
Added in Package names
Disabled rdo scatter matrix. It has a bug. Thought it wise to disable it. Your Views over the same?


dannyparsons:
Yes happy with that if it's crashing.

PR #3325, opened 2017-06-14
New data to library

### LINKED ISSUE #3095: Case study from Kenya - climatic
[WKenya_StationInfo.zip](https://github.com/africanmathsinitiative/R-Instat/files/1008263/WKenya_StationInfo.zip)

Here are the data from James Musyoka for the rainfall stations.  

There also should be some stations with multiple elements.  It would be very good to get those as well.  We have some stations (particularly Dodoma) with rain, maxtemp, mintemp and sun hours.  It would be good to get at least one example with both multiple stations and multiple elements, and with more elements than these four.

maxwellfundi:
Fixes #3095 

@dannyparsons  this is now added too.


dannyparsons:
Can you also check Procurement section of the library? Anything that says "Corruption" should be replaced with "Procurement"

maxwellfundi:
Okay, Right away..

PR #3337, opened 2017-06-14
developer error fix

shadrackkibet:
@dannyparsons  this is a minor fix.This is ready to be merged.


dannyparsons:
What about the other input control?

shadrackkibet:
The other input control doesn't bug.

dannyparsons:
I think it should be non editable as well anyway, because it doesn't allow custom values?

shadrackkibet:
i have changed both now.

PR #3346, opened 2017-06-14
Models Dialogs 

stevenndungu:
Added package names for sdgSimpleRegOptions
Disabled chkMultiplePlots. Checked with @lilyclements function autoplot does not support glm/lm functions.
Linked three variable modelling dialog with ucrDistributions-was bugging

### DIFF
--- a/instat/dlgOneVarCompareModels.vb
+++ b/instat/dlgOneVarCompareModels.vb
@@ -37,7 +37,10 @@ Public Class dlgOneVarCompareModels
         UcrReceiver.Selector = ucrSelectorOneVarCompModels
         UcrReceiver.SetMeAsReceiver()
         UcrReceiver.SetItemType("model")
-        ucrBase.clsRsyntax.SetFunction("gofstat")
+        ucrBase.clsRsyntax.SetFunction("fitdistrplus::gofstat")
+        'Temporry fix- If you run the second time it breaks, if the check box remains checked. Needs to be fixed
+        sdgOneVarCompareModels.chkCDF.Enabled = False
+
         sdgOneVarCompareModels.SetModelFunction(ucrBase.clsRsyntax.clsBaseFunction)
         sdgOneVarCompareModels.SetReceiver(UcrReceiver)
         sdgOneVarCompareModels.DisplayChiSquare()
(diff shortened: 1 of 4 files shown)


AlexSananka:
Testing this now.

AlexSananka:
one quick thing to fix,can you change the call type so that we can have our plots in the output window for consistency?

stevenndungu:
But i think there was a good reason why the plots are previewed from R otherwise that would have been done before...i guess

AlexSananka:
This is a very tiny fix and I think we've always changed them to the output window,I have done this before in a current pull request,Unless there is a good reason why we shouldn't change on this one.

dannyparsons:
should be consistent and all graphs go in the output window

instat/dlgOneVarCompareModels.vb
```diff
@@ -37,7 +37,10 @@ Public Class dlgOneVarCompareModels
         UcrReceiver.Selector = ucrSelectorOneVarCompModels
         UcrReceiver.SetMeAsReceiver()
         UcrReceiver.SetItemType("model")
-        ucrBase.clsRsyntax.SetFunction("gofstat")
+        ucrBase.clsRsyntax.SetFunction("fitdistrplus::gofstat")
+        'Temporry fix- If you run the second time it breaks, if the check box remains checked. Needs to be fixed
+        sdgOneVarCompareModels.chkCDF.Enabled = False
+
```
AlexSananka: I understand that this was disabled because this will cause the dialog to crash when you run the dialog the second time when this checkbox is checked.But will we never have CDF plot?
@dannyparsons  do you have a suggestion on how this can be fixed?

AlexSananka: It's a good temporary fix for now.I was just wondering if in `onevarusemodel` we can do same(for now) since we have a checkbox with the same issue.It crashes when when you run the dialog the second time.

dannyparsons: ok yes if it's the same on another dialog then that's fine to disable there too

stevenndungu:
@AlexSananka @dannyparsons is what we have bugging???  am having trouble with this branch so I can't make any more changes. Can this  be merged then the  Icall type be changed in another branch.

dannyparsons:
Ok I'm merging this. Please check in the master now as I'm not up to date on the latest bugs with these dialogs.

PR #3369, opened 2017-06-14
Climaticsummary disable

shadrackkibet:
@dannyparsons this is an empty dialog that need to be disabled.lily must have forgotten this on a closed issue.


dannyparsons:
great thanks

PR #3375, opened 2017-06-15
A temporary fix for resetting on regular sequence

Lunalo:


dannyparsons:
Thanks for spotting this bug. However, now that we're passed the 0.3 release we need to get back to "proper" fixes and remove all our temporary fixes.
Can you find the cause of this bug? If not, write it up so others can help.

Issue #3345, opened 2017-06-14
Graphics oddities


rdstern:
I use the diamonds data set.
a) I have a boxplot of carat by cut (as the factor)
b) I go to the **options** and **themes** and **All Options**
c) I set the **x-axis angle to 90**
d) I return to the themes and set the **theme to minimal**
e) I return and do the plot.  The theme is correctly minimal, but the x-axis angle is zero.
f) I repeat and get the same thing - still minimal with still not my angle
g) I return to the sub-dialogue and find it has reset to the default - grey.
h) I change to Economist as the theme and it gives an error - tells me there isn't that theme.
i) I return to the subdialogue and see it is grey.  I try that plot.  No bug, but also no graph - it is blank:  The command is now:
# dlgBoxplot code generated by the dialog Boxplot
diamonds_temp <- InstatDataObject$get_data_frame(data_name="diamonds")
last_graph <- ggplot2::ggplot(data=diamonds_temp, mapping=ggplot2::aes(y=carat, x=cut)) + ggplot2::geom_boxplot(varwidth=TRUE)
InstatDataObject$add_graph(data_name="diamonds", graph=last_graph, graph_name="last_graph")
InstatDataObject$get_graphs(graph_name="last_graph", data_name="diamonds") 

j) I check the options again, and then All Options and see the 90 setting is still in place.  I untick and tick again then return to plot - with theme back to the default.  (So nothing is changed on the dialogue from just before.)
I now get my graph with the angle at 90!  The command now is:
# dlgBoxplot code generated by the dialog Boxplot
diamonds_temp <- InstatDataObject$get_data_frame(data_name="diamonds")
last_graph <- ggplot2::ggplot(data=diamonds_temp, mapping=ggplot2::aes(y=carat, x=cut)) + ggplot2::geom_boxplot(varwidth=TRUE) + ggplot2::theme(axis.text.x=ggplot2::element_text(angle=90))
InstatDataObject$add_graph(data_name="diamonds", graph=last_graph, graph_name="last_graph")
InstatDataObject$get_graphs(graph_name="last_graph", data_name="diamonds")

dannyparsons:
@stevenndungu @maxwellfundi please look into this urgently

dannyparsons:
Found out the theme name was overriding the theme options - fixing this now.
Will now look into the theme name resetting.

stevenndungu:
[...]

PR #3407, opened 2017-06-16
Linking Sdgplot and SdgTheme

stevenndungu:
This fixes issue #3345 in a better way than what we now. Position of the parameters is critical on how theme parameter overwrite those of the applied theme.


dannyparsons:
Position shouldn't affect overwriting, only the parameter names can overwrite. However, there are issues with the current version. We should actually be using functions for themes instead of strings, as I did this as a quick fix. Will look into what is done here.

maxwellfundi:
@stevenndungu This seems like it has not been addressed. Is it something you could fix quickly for this version?

stevenndungu:
This was addressed by Danny. I will close  pull request but see if there is any enhancement I can make in  what we have now.

PR #3417, opened 2017-06-16
Rewriting One Variable Model fit

stevenndungu:


dannyparsons:
Great that this is getting updated. See the changes I made to one variable compare model in the latest to see how to handle multiple additional functions and link to sub dialogs - particularly how to add additional codes running after the base code.

stevenndungu:
@africanmathsinitiative/developers this is ready for review...

AlexSananka:
Testing this now

AlexSananka:
Sort out the issues we have discussed together,also the designer needs some edits so that it can look nit(currently it is somehow wide).
Both ucrSaves in the subdialogs are not working,can we make them working?If not then can we disable them until we can make them working?
I also understand that some four distributions have an `error 100`.I don't know how you're planning to sort out this.
Once this issues are sorted out then this will be ready for @dannyparsons .

stevenndungu:
@africanmathsinitiative/developers  this is ready for re-review..
Note: 
there is one bug which was there  in the previous installer though had not been identified yet.
If i choose for example negative binomial distribution, this error is given
image

the `error code 100 ` is given which i guess is an issue with the `fitdist` function.
-One var fit Model in the latest version 0.3.1 is not working (has a no of bugs)

stevenndungu:
the error above occurs with  Hypergometric Distribution,Negative Binomial,Binomial,extreme value and von mises only. Others are working quite well.

AlexSananka:
@dannyparsons I have tested this and it's working fine and more better than what we have currently.
I think it can go in.

PR #3436, opened 2017-06-18
Alex sananka one var compare

dannyparsons:


dannyparsons:
@AlexSananka here I have updated your one var compare to include new method for multiple functions. It should be clear what the new methods do but have a check.

I was unclear what the ucrSave controls on the sub dialog should be doing. For example, they had combo boxes with no items. I have disabled them for now. Everything else should be working. You probably should discuss with @rdstern and @lilyclements about these disabled options and what they should do.

PR #3447, opened 2017-06-19
to script option for RCode to be string

dannyparsons:


instat/clsRCodeStructure.vb
```diff
@@ -54,6 +54,11 @@ Public Class RCodeStructure
     'Will be cleared after running unless bClearFromGlobal = False
     Public bClearFromGlobal As Boolean = False
 
+    ' If True when running ToScript the function will be returned as a string that could be passed to R
+    ' e.g. "seq(from = 1, to = 10)" instead of seq(from = 1, to = 10)
+    ' When True, assignment cannot be used for the function or its parameters
+    Public bToScriptAsRString As Boolean = False
```
dannyparsons: @lilyclements @shadrackkibet use this option when you have an RFunction/ROperator which is a string argument to another function.

lilyclements: @dannyparsons I'm struggling with this slightly. When I run it, it just runs
`parameter_name = False`
rather than anything else. How do I use this?

dannyparsons: How are you using this? Can I see it in some of your code?

PR #3459, opened 2017-06-20
Start of rains

lilyclements:
@dannyparsons This contains a few changes now to various classes/data object/core control.

Start of Rains is not yet completed (in fact, it's just come up with a new bug that it's not reading in sub1/sub2/etc anymore), however, you may find you want to merge this for the other chnages. I'll continue with start of rains later.


dannyparsons:
Will merge this because of core changes to give longer time for testing of some of these.

PR #3467, opened 2017-06-21
Start of rains

lilyclements:
@dannyparsons there are several core changes made here so if you think suitable then perhaps merge this?
I don't mind either way.

Changes still to make:
* Work out where to call sub_calculation parameters
* Sort calculated_from parameters
* Finish chkTotalRainfall combined filter set up


instat/ucrCore.vb
```diff
@@ -221,14 +221,16 @@ Public Class ucrCore
         UpdateLinkedControls(bReset)
     End Sub
 
-    Public Overridable Sub SetRCode(clsNewCodeStructure As RCodeStructure, Optional bReset As Boolean = False)
+    Public Overridable Sub SetRCode(clsNewCodeStructure As RCodeStructure, Optional bReset As Boolean = False, Optional bUpdate As Boolean = True)
```
lilyclements: Fun fact, David added this in here. I then had a bug with two other subs - Geoms and another one so I added in this optional on their sub title too but didn't change anything within the sub. Is this okay?

dannyparsons: yep that's correct, so they match now.

Can you also pass bUpdate through to SetRCode inside this i.e. line 49 in Geom

instat/ucrCore.vb
```diff
@@ -221,14 +221,16 @@ Public Class ucrCore
         UpdateLinkedControls(bReset)
     End Sub
 
-    Public Overridable Sub SetRCode(clsNewCodeStructure As RCodeStructure, Optional bReset As Boolean = False)
+    Public Overridable Sub SetRCode(clsNewCodeStructure As RCodeStructure, Optional bReset As Boolean = False, Optional bUpdate As Boolean = True)
         If clsRCode Is Nothing OrElse Not clsRCode.Equals(clsNewCodeStructure) Then
             clsRCode = clsNewCodeStructure
-            If bUpdateRCodeFromControl AndAlso CanUpdate() Then
+            If bUpdateRCodeFromControl AndAlso CanUpdate() Then 'andalso bUpdate?
```
dannyparsons: Yes this should probably be included

PR #3468, opened 2017-06-22
Updating dlgFromLibrary

maxwellfundi:
@dannyparsons This dialog works well but I need help in two things here. 

1. When it loads it trows an error that I wasn't able to fix quickly 
image

2. The AddToBeforeCodes method doesnt hve an option to add the comments. So the BeforeClickOk for this dialog still runs. If that could be added then it would be great. 

Apart from these two, everything else works well. 

Thanks.


dannyparsons:
Thanks for specifying the bugs. I've fixed these and merged.

maxwellfundi:
You welcome. Its been good to get it done.

dannyparsons:
there's still couple things to do which I've commented on the issue

PR #3470, opened 2017-06-22
Rewriting Three variable modelling

AlexSananka:
@africanmathsinitiative/developers this is making good progress,once it's subdialogs are merged i will go on and do the linking.


dannyparsons:
Are the sub dialogs the same ones for 2 variables? Is that why you're waiting?

Don't do anything temporary if it's going to change soon. You can just leave it as it is.

AlexSananka:
Yes they are the same,until those are merged so that any work for this can continue.

Am on a different task now.

dannyparsons:
working on this now

PR #3477, opened 2017-06-23
End of rains

lilyclements:
@dannyparsons would you mind taking a look at this? It crashes any time I run it to the first line in Set Defaults which isn't to do with adding a key, days or groups. (i.e., is new code, i.i.e, it crashes on "clsWBReplaceNAMax")

I can't figure this out. If you get a spare moment could you have a look? Thanks


lilyclements:
@dannyparsons Not to worry, I worked it out. I was calling As RFunction not As New RFunction

lilyclements:
Sorry for poor formatting. On phone on plane soon to take off!

This dialog contains two checkboxes (at least one has to be checked for OK to be enabled. That has not yet been set in TestOKEnabled)

One checkbox: end of season.
This has been implemented and may need a few tweaks, most of which will have a todo by them but in general is good.

Second checkbox: end of Rains
I've started on this and written in the r-code (so SetRCommand, AddParameters etc are sorted)
Things need to be set such as - calculated_from needs to be moved to a sub, TESTOK needs to be set up for this instance, this needs to run if OK is checked (So we need to run apply_instat_calculation if this is checked). I also haven't had a chance to check what code I did write does work.

I have to go! Can someone please finish this? I can't tag everyone in this comment as on phone so Danny can you edit my comment here and tag developers in?

lilyclements:
Just landed! Realised on the flight that the definition is slightly wrong because it's meant to be the last instance where the rainfall over X days is greater than the threshold.
This means before the filter calculation there needs to be another calculation where we look at the rolling sum of rain days and this creates a new column. Then the filter calculation looks at this new column not the rain column. This means the "calculated_deom" should be removed from this calculation and instead a "sub_calculation" should be added.

lilyclements:
Realised on the flight the definition is slightly wrong. Just corrected it and managed to sync it in. Laptop is off now and phone is going away!
There's a few TODOs on there for if someone gets the time if we want this in 0.3.1, and this needs to be tested more, these fixes should be okay to make!
I believe I've outlined some of these above. All changes needed should be fairly straight forward I hope.

shadrackkibet:
@lilyclements is this ready for review?

lilyclements:
This is not yet complete, however, I am traveling so finding it difficult to get time on this.

maxwellfundi:
@lilyclements  could you revisit this now? It would be good to have it fixed.

lilyclements:
I fixed the remaining issues on this dialog at the start of July. This is now at a standstill until I can fix the radio buttons. Similar problems are seen in Transform and in Steve's View.

lilyclements:

lilyclements:
@africanmathsinitiative/developers we would like this in today's version. Someone needs to review this.

AlexSananka:
reviewing this

AlexSananka:
I have tested this and seems to work fine for me. @dannyparsons have a look.

PR #3480, opened 2017-06-26
Various bug fixes

dannyparsons:
- fixed issues reading old options files
- added separate class for default options
- auto recovery files now deleted correctly after use
- fixed bug in reading metadata with one column data frames


dannyparsons:
@africanmathsinitiative/developers please can you test the following in the latest version:

1. auto recovery files now deleted correctly after use - if you crash, then reopen, then close properly, then reopen again - you should no longer get auto recovery detected i.e. auto recovery should only appear the first time after reopening from a crash

2. fixed issues reading old options files - when you run the first time after updating, please capture any errors relating to the options file that appear and let me know. There shouldn't be any but if there is it's caused by these changes

Thanks

PR #3485, opened 2017-06-27
SoR bug fixes

lilyclements:
@africanmathsinitiative/developers this is ready to review. Here i have fixed a few minor problems on Start of Rains and Spells


instat/dlgStartofRains.vb
```diff
@@ -248,7 +243,7 @@ Public Class dlgStartofRains
         clsDPCombineOperator.Clear()
 
         ucrReceiverDate.SetMeAsReceiver()
-        ucrInputNewColumnName.Reset()
+        ucrInputNewColumnName.SetName("Start_of_Rains")
```
dannyparsons: This should be done by setting the parameter in the default function instead of here, if possible.

lilyclements: I don't understand. I can see how to do it for End of Rains because the "Save columns" are linked up to a checkbox. What do you mean here?
Because it's not a ucrSave but is instead a ucrInput, it has that problem we often get that "reset" doesn't reset it.

dannyparsons: I mean you should add result_name = "Start of Rains" somewhere in a function and then this control will be linked to that function and set automatically from this value so you don't need to set it manually. I can do it on your branch when I get time if it's easier.

lilyclements: Of course. I've fixed this now so this is ready for re-review

PR #3503, opened 2017-06-30
Bug fixes

lilyclements:


lilyclements:
What is the problem with fac.nested out of interest? Is there an issue with the code which is why it's disabled? When I tried to run it, the only error I could see was that we haven't set the package name for fac.nested. I'm not sure what the package is (Is it "dae" from a quick google? Slow internet so I can't really try to download that package!). However, I didn't want to spend time exploring this further in case this has OK disabled for a non-quick-fix-reason. Why is this dialog disabled?

lilyclements:
Update: Fixed the Count in Factor dialog. The issue seemed to just be that the package hadn't been added.

@africanmathsinitiative/developers This may mean that `dae` has to be installed into R-Studio and onto laptops now too for this to work

dannyparsons:
I think the dae package hadn't been added in to our list yet, but I can add it now for the next release since this is working, thanks.

PR #3527, opened 2017-06-30
Three variable update

AlexSananka:


AlexSananka:

AlexSananka:
This dialog is working fine only one developer error which I am struggling to replicate it but it popped up when I filled the response receiver then go to the display options `subdialog `then I return to the main dialog and then reset and I have this 
Developer error: Cannot set value of `ucrDistributionChoice `because cannot find a distribution matching the function and parameter given.

Once I retrieve this and fix it then this is ready for today's release.

PR #3537, opened 2017-07-01
Opening Logs in R-Studio

### LINKED ISSUE #3460: Opening log file in R-Studio directly
This should be a really easy but powerful addition. We could have this as a right click option from the log window for example. 
I believe if we save the log text to a file in an appropriate directory and then it's just a question of running:
 
Process.Start("logname.R") 

If R-Studio is installed and the default program to open .R files with this should simply work!

maxwellfundi:
Fixes #3460 
@africanmathsinitiative/developers  there is this new nice feature now in R-Instat.  Once you have run things on R-Instat, you can now go to log window, right click and be able to run the same in R-Studio. 

Thanks.

### DIFF
--- a/instat/frmLog.vb
+++ b/instat/frmLog.vb
@@ -20,20 +20,11 @@ Public Class frmLog
         frmMain.mnuViewLog.Checked = Me.Visible
     End Sub
 
-    Private Sub mnuOpenLogFile_Click(sender As Object, e As EventArgs) Handles mnuOpenLogFile.Click
-        Dim clsProcessStart As New RFunction
-        Dim strLogFilename As String = ""
-        Dim dlgOpenLogFile As New OpenFileDialog
-        clsProcessStart.SetRCommand("Process.Start")
-        dlgOpenLogFile.Filter = "All R files|*.R;|All files|*.*;"
-        dlgOpenLogFile.Title = "load Log File"
-        dlgOpenLogFile.Title = "load Log from file"
-        If dlgOpenLogFile.ShowDialog() = DialogResult.OK Then
-            If dlgOpenLogFile.FileName <> "" Then
-                strLogFilename = dlgOpenLogFile.FileName
-            End If
-        End If
-        clsProcessStart.AddParameter("filename", strLogFilename, bIncludeArgumentName:=False)
-        frmMain.clsRLink.RunScript(clsProcessStart.ToScript, strComment:="Loading Script")
+    Private Sub mnuOpenLogFile_Click(sender As Object, e As EventArgs)
+
+    End Sub
+
+    Private Sub frmLog_Load(sender As Object, e As EventArgs) Handles MyBase.Load
+
     End Sub
 End Class
--- a/instat/frmLog.vb
+++ b/instat/frmLog.vb
@@ -1,4 +1,5 @@
-﻿Public Class frmLog
+﻿Imports RDotNet
+Public Class frmLog
     Protected Overrides Sub OnFormClosing(ByVal e As FormClosingEventArgs)
         MyBase.OnFormClosing(e)
         If Not e.Cancel AndAlso e.CloseReason = CloseReason.UserClosing Then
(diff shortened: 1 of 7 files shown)


dannyparsons:
Can you add it to the Tools menu as well?

instat/ucrLog.vb
```diff
@@ -13,13 +13,26 @@
...
+    Private Sub mnuOpenLogFile_Click(sender As Object, e As EventArgs) Handles mnuOpenLogFile.Click
+        clsInstatOptions = New InstatOptions
+        Dim clsProcessStart As New RFunction
+        Dim strLogFilename As String = "RInstatLog.R"
+        Try
+            File.WriteAllText(Path.Combine(clsInstatOptions.strWorkingDirectory, strLogFilename), txtLog.Text)
+        Catch
+            MsgBox("Could not save the log file." & Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.", MsgBoxStyle.Critical)
+        End Try
+        Process.Start(clsInstatOptions.strWorkingDirectory & "\" & strLogFilename)
```
dannyparsons: This should also be in the Try.
You should be getting the path in exactly the same way as you do when you save the file, otherwise there's a bigger chance for errors appears. Better yet, create a variable for the path so you are sure you are opening the path that was saved.

instat/frmLog.vb
```diff
@@ -1,4 +1,5 @@
-﻿Public Class frmLog
+﻿Imports RDotNet
```
dannyparsons: This shouldn't be needed now?

instat/ucrLog.Designer.vb
```diff
@@ -53,6 +58,18 @@ Partial Class ucrLog
...
+        '
+        Me.mnuContextLogFile.Items.AddRange(New System.Windows.Forms.ToolStripItem() {Me.mnuOpenLogFile})
+        Me.mnuContextLogFile.Name = "mnuContextLogFile"
+        Me.mnuContextLogFile.Size = New System.Drawing.Size(148, 26)
+        '
+        'mnuOpenLogFile
+        '
+        Me.mnuOpenLogFile.Name = "mnuOpenLogFile"
+        Me.mnuOpenLogFile.Size = New System.Drawing.Size(152, 22)
+        Me.mnuOpenLogFile.Text = "Open Log File"
```
dannyparsons: How do I copy on the right click menu in the log now?

maxwellfundi: What do you mean copy? The script?

maxwellfundi: Just a quick thought, Should we have this as "open log in R-Studio"?

stevenndungu:
One more comment. If you open a log file on RStudio for the first time then create another log window with different code from the previous one, all the log windows bear the same name and therefore they overwrite the previous ones.
 Could it be possible to have have a warning that the log file already exists? To avoid overwriting of previous important files used before.

maxwellfundi:
@dannyparsons I have fixed your issues. 

@stevenndungu thank you for your concern. I think we can have this for this version then we can improve it later. Its really such a good point. 
however, I think once you quit an R-Instat session, then next time you will have a new log and if you wanted to keep the old one you can easily save from R-Studio.

instat/ucrLog.Designer.vb
dannyparsons: It's not open in RStudio because it might open with another program. 
Yes can I still copy the log from the right click?

instat/ucrLog.vb
```diff
@@ -13,13 +13,29 @@
...
+
+    Private Sub mnuOpenLogFile_Click(sender As Object, e As EventArgs) Handles mnuOpenLogFile.Click
+        clsInstatOptions = New InstatOptions
+        Dim clsProcessStart As New RFunction
+        Dim strLogFilename As String = "RInstatLog.R"
+        Dim strWD As String = clsInstatOptions.strWorkingDirectory
+
+        Try
+            File.WriteAllText(Path.Combine(strWD, strLogFilename), txtLog.Text)
+            Process.Start(strWD & "\" & strLogFilename)
```
dannyparsons: You are still getting the full file path in two completely different way. Why are you using Path.Combine in the first line but not the second?

dannyparsons:
Steven's point is quite important. Saving it in the temporary folder is quite dangerous actually, because the files could be cleared, and there's also the overwriting issue. I think we should either save it somewhere like Documents, which won't get deleted, or ask the user to pick a location first. I think the first option would be fine. But we should also then create an "R-Instat Logs" folder in documents for this, and make sure we don't overwrite.

maxwellfundi:
Okay, then i think we could have an option of saving which the user will choose a location and give the file a name then another item on the list to load so that that select the file from where they saved it because they could have multiple logs

dannyparsons:
Yes we could have those as options, but for now a folder in Documents is fine. From the software we only need one menu item to open the current log file - it doesn't matter if they've saved others before.

instat/frmLog.vb
maxwellfundi: Yes

maxwellfundi:
@stevenndungu  this last commit solves your issue. thanks for raising it. Its a great addition.

maxwellfundi:
@dannyparsons  i have now put in the copying feature. Its always enabled. If no text is selected then everything is copied and if a portion is selected then only that portion is selected.

PR #3545, opened 2017-07-01
adjusting chckbox size to fit text

maxwellfundi:


maxwellfundi:
@dannyparsons I had in mind that i did this already. now its done

PR #3599, opened 2017-07-04
One var fit Model

### LINKED ISSUE #3560: One Variable Fit Model
1. I'm not keen on that when you use a factor, a checkbox comes up that says "Convert to Variate" and is checked and won't let you uncheck it.
If this bug is planned to be fixed in the future, then can we make it so this option is disabled as otherwise it may confuse the user?
If this isn't a bug but is instead that you have to convert it to a variate in order to use a factor, then I don't agree with using a checkbox to do this since a checkbox implies the user has a choice.
It would be helpful if someone would explain to me why it's currently permanently checked

2. If I click the "Plot Likelihood" checkbox and then select "Plot Log-Likelihood" radio button. When I reopen that sub dialog, the "Plot Likelihood" checkbox is unchecked, which is not as I left it when I pressed return.

Other than that, this is looking really good!
@AlexSananka @stevenndungu @dannyparsons (I think it was Alex and Steven who were working on these? If I'm wrong then very sorry!)

stevenndungu:
@dannyparsons @lilyclements there is a slight difference in the behaviour of  "Convert to Variate" for one variable compared to 2 and 3. Setting it that way introduces an unnecessary bug.
I have therefore set it to behave in exactly the same way as it was before.
Fixes #3560


lilyclements:
Reviewing this now

instat/dlgOneVarFitModel.vb
```diff
@@ -340,6 +338,7 @@ Public Class dlgOneVarFitModel
                 ucrChkConvertVariate.Visible = False
             Else
                 ucrChkConvertVariate.Visible = True
+                ucrChkConvertVariate.Enabled = False
```
lilyclements: Okay thank you for disabling it.
I much prefer that than it looking like the user has a choice when they don't. Is the idea for "Convert to Variate" to always be disabled? Or is this a temporary fix to be fixed in a later version? 
If temporary can you put a comment next to it here stating as such?

I'm not sure if we should have it disabled here or at Initialise. I know it doesn't make a difference how the dialog runs, however, it is generally in Initialise. Perhaps I'm being overly picky, that can happen..

lilyclements:
This looks good, thanks for fixing the issue. I've noticed a new problem, however. I know we've had this issue on other dialogs but not sure how it was solved, but hopefully you know (otherwise, I think we had this problem on one of Maxwells dialogs so he might be able to help). 

If I open the sub, click "Save model" for likelihood then I see it will be called likelihood1.
I rename it something
Then press reset on the main dialog.
Reopen the sub, and click "Save model" for likelihood. The name is now automatically last_likelihood.

Similarly this occurs on the plots tab of this sub dialog.

AlexSananka:
Just a quick thing to note:
OK is always enabled in this dialog.You can fix this quickly.

AlexSananka:
No sorry this is fixed now.

instat/dlgOneVarFitModel.vb
stevenndungu: Thanks... for this case it is not a temporary fix. The ucrinputDistributions is never empty for 1 var model based on the conditions set to it unlike 2 and 3. Therefore, you cannot have  "ucrDistributionChoice.ucrInputDistributions.IsEmpty" check on testOK.

stevenndungu: Yeah we can have it on initialise, thats ok

stevenndungu:
@lilyclements  the issue on reopen on ucrSave it is an issue with ucrsave control (known bug).  Which I remember Danny saying he will edit the ucrsave code some time back...

lilyclements:
Great, thanks for clarifying! I know we've had the problem elsewhere, I just didn't know if there was a solution yet.
Have you fixed the TestOK issue? Let me know once you have because then it should be ready for Danny

stevenndungu:
Yap testOK was fixed. Alex had checked someday before not today

lilyclements:

dannyparsons:
Why can't the Convert to Variate option work the same way it does on 2 and 3 variable?

stevenndungu:
the ucrdistribution is never empty  when a factor variable is selected. I think makiing it work in the same way implies I have to make some possibly significant changes on how the ucrdistribution is set in this dialogue..We can discuss,,will get to you shortly

dannyparsons:
The important thing is that the 1/2/3/4 dialogs behave consistently. When you select a multi level factor, there are no distributions available for this type. If the user would like to treat the factor as numeric, they can, and then choose from the numeric distributions. But this is a choice by the user and doing it automatically without them seeing what happens is confusing - they may not even see the check box and notice it's been converted.

If this is going to be difficult then I don't mind doing something quick at the moment, but I don't think this is ideal. I would prefer to implement as in 2/3 variable, unless it's going to take a long time.

dannyparsons:
Is this complete?

stevenndungu:
Yes, now it's done

PR #3603, opened 2017-07-04
Fixing One variable graph bug

### LINKED ISSUE #2879: One Variable Graph - Single Graphs bug
On One Variable Graph, facets and combine graphs save well. However, there is an error with "Single Graphs"

If I open the dialog for the first time, and select "Single Graphs" and "Save Graph", the graph does not save: 1. it always calls the graph _onevar1_ if I reopen the dialog and add different variables/etc, 2. It doesn't show up on the dlgUseGraph dialog.

If I then save another graph (e.g. Facets or Combine) then the Single Graphs, the "Single Graphs" starts to try to save. They try to save as last_graph if the checkbox is unchecked, or as _onevarX_ if checked (as we would want!). Irregardless of the checkbox being checked or not, the following error occurs:

> Error in self$get_data_objects(data_name)$get_objects(object_name = object_name,  : 
>   one_var2 not found in objects
> 
> 
> The error occurred in attempting to run the following R command(s):
> 
> one_var2 <- InstatDataObject$graph_one_variable(data_name="Blocking", columns=c("x2","x3"), output="single", numeric="geom_boxplot", categorical="geom_bar")
> 
> InstatDataObject$add_graph(data_name="Blocking", graph=one_var2, graph_name="one_var2")
> 
> InstatDataObject$get_graphs(graph_name="one_var2", data_name="Blocking")

I'm not sure why this isn't working for "Single Graphs".
However, I also do not understand how we can save multiple single graphs under one name so could the issue be to do with this?

Any clarity in the error would be appreciated!

maxwellfundi:
@dannyparsons  I have now removed the strange parameter and disabled the the single graphs. 

Fixes #2879


dannyparsons:
Great, can you write another issue for later that we need our output window to be able to display multiple graphs produced from a single function?

PR #3628, opened 2017-07-05
Generalfit model updating 

Lunalo:


maxwellfundi:
@dannyparsons have a loo at this.

Lunalo:
am finishing up on something

dannyparsons:
still working on this?

Lunalo:

This is ready for review

AlexSananka:
Am reviewing this

instat/dlgFitModel.vb
```diff
@@ -14,85 +14,175 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports instat
 Imports instat.Translations
```
AlexSananka: Do we need this?

instat/dlgFitModel.vb
```diff
@@ -14,85 +14,175 @@
...
 
+        ucrBase.clsRsyntax.SetBaseRFunction(clsLM)
+        ucrBase.clsRsyntax.AddToAfterCodes(clsAnovaFunction, 1)
+        ucrBase.clsRsyntax.AddToAfterCodes(clsSummaryFunction, 2)
     End Sub
 
     Private Sub TestOKEnabled()
         If (Not ucrReceiverResponseVar.IsEmpty()) AndAlso (Not ucrReceiverExpressionFitModel.IsEmpty()) Then
-            ucrBase.clsRsyntax.AddParameter("formula", clsROperatorParameter:=clsModel)
             ucrBase.OKEnabled(True)
```
AlexSananka: TestOk will always be enabled even when the `ucrDistributionChoice `is empty ,which will cause a bug.
You can add this line` Not ucrDistributionChoice.ucrInputDistributions.isEmpty`

instat/dlgFitModel.vb
```diff
@@ -216,68 +299,47 @@ Public Class dlgFitModel
...
-            ucrModelName.Visible = False
-        End If
-        AssignModelName()
-        TestOKEnabled()
-    End Sub
 
     Private Sub AssignModelName()
-        If chkSaveModel.Checked AndAlso Not ucrModelName.IsEmpty Then
-            ucrBase.clsRsyntax.SetAssignTo(ucrModelName.GetText, strTempModel:=ucrModelName.GetText, strTempDataframe:=ucrSelectorByDataFrameAddRemoveForFitModel.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
+        If ucrModelName.ucrChkSave.Checked AndAlso Not ucrModelName.IsComplete Then
```
AlexSananka: Fix this as we have discussed,we don't need this sub,you can attain this in the Setdefaults

AlexSananka:
The distributions we have in the main dialog are OK and working fine ,this are the same distributions that should be in the sub dialog for model options,you can check on threevariable modelling and see how this was achieved.

In general this dialog is working fine,once those issues have been addressed this will be ready.

instat/dlgFitModel.vb
Lunalo: I think we didnt say we dont need it.  I said we should improve on it since we have a checkbox on the ucrSave

instat/dlgFitModel.vb
Lunalo: This should be ucrFamily I think

AlexSananka:
@dannyparsons This is ready for you

lilyclements:
Reviewing/making changes to this now

lilyclements:
1. If I check the checkbox and reopen, it is unchecked. This is presumably because it is not being called in the Set R Codes? This causes a developer error upon reopening.

2. So, I was getting the following bug. Then I reopened and it wasn't occuring anymore. I can't replicate it so this is something to look out for I guesss in case it occurs again! If you change the ucrSave name and click reset, then it is called "last_model" not "Genx". I know this is a known bug with sub dialog ucrSaves, so perhaps this is part of that bug. Apologies if that is the case and this is a known bug!

3. If I do Factor ~ Variate and use Binomial, I get an error:

> Error in if (ssr < 1e-10 * mss) warning("ANOVA F-tests on an essentially perfect fit are unreliable") : 
>   missing value where TRUE/FALSE needed
>

lilyclements:
Fixing these problems outlined above

lilyclements:
Fixed 1.

Haven't been able to replicate 2. so we won't worry about that for now. If I can't replicate it then I can't know how to fix it

for 3. -  if I check convert to variate, it works. It's when I don't check the box that it doesn't work. However, it should allow me to do Factor ~ Variate.
I've noticed that it is not calling in `family = DISTRIBUTION`, and also it is doing a `lm` not a `glm` as we would want in this case.
Is there a reason why this dialog does not have a `glm` function?

PR #3637, opened 2017-07-06
Jitter

stevenndungu:


dannyparsons:
If this is easy, great. Otherwise, it's not the highest priority at the moment.

maxwellfundi:
@stevenndungu  this has not been a priority for sometime. Could you have a look at it again now so that it gets to the master? Is it too much work still left here?

Lunalo:
@dannyparsons 

I remember discussing with you on this dialog. We had agreed that the dialog needed more thoughts and possibly restructuring.  I would suggest the dialog's design to be restructured .

dannyparsons:
Let's then merge this and disable the dialog on the main form so that it doesn't need to be complete. The form main edit can be done on a separate branch. Can someone do this?

stevenndungu:
It is already disabled.

PR #3659, opened 2017-07-11
Adding ghana dataset

Lunalo:


dannyparsons:
You also need to set the build and copy action. Check with @maxwellfundi

PR #3671, opened 2017-07-15
Adding cast dataset

### LINKED ISSUE #3667: Sam's Splitting
Email from Sam which can be easily implemented as additional functionality on the split text dialog can we do this immediately? @lilyclements  @maxwellfundi @Lunalo We will discuss on the train I believe this might be worth doing immediatly or for 4.1

"I have had a thought about something that could be really useful to include in R-Instat, especially since you are trying to link it up with ODK.

When you have a multiple select variable in an ODK form you might get output that looks like the column in the file attached. You can’t really do much, if any, analysis on that column, and transforming it into something useful is a pain for most people to deal with. The excel route of splitting it into binary variables is tedious and conceptually hard to explain since it relies on using a combination of really non-obvious functions, or alternatively some even more tedious manual labour. Genstat/SPSS/Stata can do it in a slightly more intuitive way but they all still require a 3-step process which often leaves people confused. There is a lovely one line solution in R, questionr::multi.split, but (for some reason) not everybody is converted over to R yet. And it’s not a really a simple one-line solution when you also have to explain how to install R, read the data in, install packages from CRAN (dealing with any weird internet issues arising), and then export the data back to excel. 

I have just come across someone with this issue for about the 3rd time this year from different McKnight grantees who aren’t comfortable using R (yet); hence why I am thinking about it now.  I’ve never actually had this problem myself since I have always got the binary variables anyway when outputting multiple select data from my ODK aggregator of choice. 

So it might be nice if it were possible to build the one line solution into a one-click solution in R-Instat? You’ve got the functions there already to do it with the same 3 moves I would use in Genstat/SPSS/Stata (split, stack, reshape), so you can just have a menu which wraps around those existing functions if that’s easier to do than loading in a new library? Since this is a specific problem dealt with by a subset of your existing target audience which is not currently directly addressed by any of your ‘competition’, then this gives you another bit of a USP if you can solve that problem in a quick and intuitive way.

Although as we convince more people to use Kobo rather than any of the other free ODK aggregators then this problem will mostly go away; so you probably need to strike now to hit this current trend of confusion!
[legumes 2008.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/1150197/legumes.2008.xlsx)

Sam"

Lunalo:
fixes #3667


dannyparsons:
That's the wrong issue. I can't find the CAST one. If you do then comment so I can close it.

PR #3678, opened 2017-07-16
Split text

lilyclements:
@africanmathsinitiative/developers this is ready to review. We want this in today's version so this is a high priority.

Added a new radio button at the top to give the option for the `multi.split` function. This is under the radio button "Into Binary Columns"

See #3667 for a further explanation of this


lilyclements:
I have made the changes we discussed @dannyparsons

PR #3681, opened 2017-07-16
Bug fixing 

### LINKED ISSUE #3284: General Graphics
**General**
1. When I press **Edit**, and go to the **Layer Parameters** tab, it seems to be back to its default state. That is, all my checkboxes are no longer checked. This occurred more often than not, except in geom_violin a couple (such as draw_quantiles) did remain checked.

2. Had an issue of updating. So the **Layer Dimensions** and the **Layer Parameters** tabs both update the data frame if you change it on one tab to the other. However, it doesn't update how this affects the options.
a) I open the sub, I go to **Parameters**, then to **Dimensions** and change to **geom_bar**. I then go back to **Parameters** and my options are as I would get for **geom_boxplot**

image

b) I open the sub, I go to **Parameters** and I change to **geom_point**. Then I go to **Dimensions** to input my variables. My options are given to me as I would get them for **geom_boxplot** (the default) not **geom_bar**:
image

3. There are several ucrInputs here which we presumably want to make non-editable

4. When pressing **reset**, could we have a warning that this will erase the current layers?

5. In the future, is it possible to add the option to rearrange layers?

**geom_point**
This may be an issue with other geoms.
I did a layer of a factor against a variate. (Constituency against Conservative Party votes)
Then I did another layer of a factor against a factor (Constituency against Labour Party votes)
since I "forgot" to convert my second factor into a variate.
This crashed it.

**geom_violin**
When the 7th option (trim) is checked, and you then click **edit** to reopen that option, there is a developer error

> Developer error: no state of control ucrInputCboParamValue satisfies it's condition. Cannot determine how to set the control from the RCode. Modify control setup so that one state can satisfy its conditions.

**geom_bar**
This crashes on `stat = "count"` 

**geom_jitter**
This allows you to press OK if none of the receivers are filled. This results in a crash
This needs y to be filled to not crash.

**geom_freqpoly**
This seems to crash when other ones are used but not by itself. This is not always the case - it's only if certain things are checked, `centre = 0`

Causes crash with geom_hline
geom_dotplot
geom_jitter

On R, when running centre = 0, you get this warning:

> Warning: Ignoring unknown parameters: center

This is so far, I'll add to this. I just want to press submit in case I lose this list!

maxwellfundi:

Just made a change on ucr color to make it non editable ... 
And put in mandatory aes for thw geom jitter to avoid a crash 

Fixes #3284


maxwellfundi:
@dannyparsons this is now done.

Issue #3697, opened 2017-07-17
Modelling bugs


lilyclements:
@africanmathsinitiative/developers these are fixes needed by the end of the day. I'm currently working on the One Variable case where I found other bugs.

**Display Options**
* If I check "Residual Plots", close the sub and reopen, then the checkbox is checked but the groupbox isn't visible. I think this should be a manual fix for now.
* Similar problem to above, but with "Fitted Model"
* On the Fitted Model tab, I think the significance level should be 0.95
* On Model > Fitted Model we are running the visreg function. Is this running the bExcludeAssignedFunctionOutput = False line? It needs to be if it's not because we're not getting the plot

**Model Options**
* On Three Variables: If I use the Normal distribution, I get an error on "Three Variable Modelling" if I run on Model Options "log" or "inverse". Should we disable the "Model Options" box for now if the **Normal, Gamma, Inverse Gamma or Quasi** distributions are used?
* On Three Variables: If I use a log-link for Binomial or Quasibinomial, I also get an error. However, the other options seem to run okay. What should we do about this?
* I don't think this code is running on Four Variables. Should it be disabled?

**Three Variables**
* Sometimes I get a bug if I use certain operators with certain variables
e.g. Poison dataset I'm using two-level factors. I get a bug if I do Sick ~ Vomiting / Nausea
What do we do about these sort of bugs?

**Four Variables**
* I put in my four variables. I did Village. ~ Variety + Variety + Fertgrp. Converted Village. to numeric, clicked on the Display tab and it crashed. I could not replicate this error when I tried!

stevenndungu:
Addressing the bugs mentioned above: starting with display option

stevenndungu:
@lilyclements  elaborate on this
"On Model > Fitted Model we are running the visreg function. Is this running the bExcludeAssignedFunctionOutput = False line? It needs to be if it's not because we're not getting the plot" I did not get you right here

AlexSananka:
'I put in my four variables. I did Village. ~ Variety + Variety + Fertgrp. Converted Village. to numeric, clicked on the Display tab and it crashed. I could not replicate this error when I tried!''

I cannot replicate this as well.

AlexSananka:
[...]

PR #3700, opened 2017-07-17
Model bug fixes

stevenndungu:
@africanmathsinitiative/developers  this ready for review
 This fixes all the Bugs mentioned in issue #3697.  We have tested together @AlexSananka   @shadrackkibet and could not replicate any of the bugs. 
Note:
Visreg function has parameter `gg` which converts plots to be of ggplot type but this does not work with  three  and four variables models. Therefore I have set it to plot based on lattice package. Which works fine

### DIFF
--- a/instat/dlgThreeVariablesModelling.vb
+++ b/instat/dlgThreeVariablesModelling.vb
@@ -176,7 +176,8 @@ Public Class dlgThreeVariableModelling
         clsVisReg.SetPackageName("visreg")
         clsVisReg.SetRCommand("visreg")
         clsVisReg.AddParameter("type", Chr(34) & "conditional" & Chr(34))
-        clsVisReg.AddParameter("gg", "TRUE")
+        'changed gg parameter to FALSE since when TRUE no plot is given (not compatible with ggplot2 package)
+        clsVisReg.AddParameter("gg", "FALSE")
         clsVisReg.iCallType = 3
 
         clsLM.SetAssignTo(ucrSaveModel.GetText, strTempDataframe:=ucrSelectorThreeVariableModelling.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempModel:=ucrSaveModel.GetText, bAssignToIsPrefix:=True)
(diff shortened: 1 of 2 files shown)


instat/dlgThreeVariablesModelling.vb
```diff
@@ -176,7 +176,8 @@ Public Class dlgThreeVariableModelling
         clsVisReg.SetPackageName("visreg")
         clsVisReg.SetRCommand("visreg")
         clsVisReg.AddParameter("type", Chr(34) & "conditional" & Chr(34))
-        clsVisReg.AddParameter("gg", "TRUE")
+        'changed gg parameter to FALSE since when TRUE no plot is given (not compatible with ggplot2 package)
+        clsVisReg.AddParameter("gg", "FALSE")
         clsVisReg.iCallType = 3
```
lilyclements: I think here we need to add the line
clsVisReg.bExcludeAssignedFunctionOutput = False line

Shall I add this in?

stevenndungu: Let me add it and see how it helps..

stevenndungu: This does not help. Line 57 does the same thing. I have explained why it does not work for 3 and 4 variables. Perhaps you could test in R to see how it behaves...I think the way I have fixed it, is the best way

lilyclements: No this is not to fix the crash problem. this is to fix that visreg isn't giving a plot, but is instead bringing up an empty box.

stevenndungu: This is not to fix the crush problem, but to fix the empty box given by visreg function as I explained on line 179 of three variable model fit dlg.

The crash bug is fixed by ucrsave controls that are disabled by now. The last ucrsave had its setrcode set inappropriately hence causing the crash since it linked to other functions in the dialogues. Line 315 of the sdgSimpleOptions sorts its out unless it is still occuring?

lilyclements: Okay thanks for explaining. So, are we still not getting the visreg box filled with the plot?

Would the crash problem not still occur occasionally as its still calling it last_graph or something

lilyclements:
This is running nicely for me, really good job! :)

PR #3698, opened 2017-07-17
One var fit fixes

lilyclements:
Still to do:

* Sort last_likelihood bug
* Checkbox is checked upon reopening
* OK is disabled upon reopening
* Error if I run a factor with the wrong distribution (e.g. Normal)

PR #3707, opened 2017-07-18
Model bug fixes

AlexSananka:


instat/dlgOneVarUseModel.vb
```diff
@@ -210,10 +210,10 @@ Public Class dlgOneVarUseModel
     Private Sub ucrChkProduceBootstrap_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkProduceBootstrap.ControlValueChanged
         If ucrChkProduceBootstrap.Checked Then
             clsQuantileFunction.AddParameter("x", clsRFunctionParameter:=clsRBootFunction)
-            ucrBase.clsRsyntax.AddToAfterCodes(clsRBootFunction, iPosition:=0)
+            ucrBase.clsRsyntax.AddToBeforeCodes(clsRBootFunction, iPosition:=0)
         Else
             clsQuantileFunction.AddParameter("x", clsRFunctionParameter:=ucrReceiverObject.GetVariables())
-            ucrBase.clsRsyntax.RemoveFromAfterCodes(clsRBootFunction)
+            ucrBase.clsRsyntax.RemoveFromBeforeCodes(clsRBootFunction)
```
dannyparsons: We think this should be after not before. First it should display the model and then additional things like the bootstrap and graphs.
I think the bootstrap should come before the graph though.

AlexSananka: Okay I was also not sure of this,I just changed for consistency but am going to change it again.
Thanks.

AlexSananka: This is sorted

AlexSananka:
@dannyparsons This is ready for you to re-review.

instat/dlgOneVarFitModel.vb
```diff
@@ -221,32 +222,26 @@ Public Class dlgOneVarFitModel
...
 
         clsRplotDenscomp.SetPackageName("fitdistrplus")
         clsRplotDenscomp.SetRCommand("denscomp")
         clsRplotDenscomp.iCallType = 3
-        clsRplotDenscomp.bExcludeAssignedFunctionOutput = False
 
         clsRLogLikFunction.SetPackageName("fitdistrplus")
         clsRLogLikFunction.SetRCommand("llplot")
         clsRLogLikFunction.iCallType = 3
-        clsRLogLikFunction.bExcludeAssignedFunctionOutput = False
```
dannyparsons: Why have you moved these to the sub dialog? They should be done here, it belongs with the other definitions of the function. It only need to be done once so doesn't make sense to do it on checked changed in the sub dialog.

instat/dlgOneVarUseModel.vb
```diff
@@ -142,7 +142,7 @@ Public Class dlgOneVarUseModel
         clsQuantileFunction.SetAssignTo(ucrNewDataFrameName.GetText, strTempModel:="last_model", strTempDataframe:=ucrSelectorUseModel.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
         'clsRBootFunction.SetAssignTo(ucrSaveObjects.GetText, strTempModel:="last_bootstrap", strTempDataframe:=ucrSelectorUseModel.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
         ucrBase.clsRsyntax.ClearCodes()
-        ucrBase.clsRsyntax.AddToAfterCodes(clsRPlotAllFunction, iPosition:=1)
+        ucrBase.clsRsyntax.AddToAfterCodes(clsRPlotAllFunction, iPosition:=2)
```
dannyparsons: Why is the plot not type 3?

instat/dlgOneVarFitModel.vb
AlexSananka: This was causing the bugs I raised on #3698.This edits happen to fix it.

instat/dlgOneVarUseModel.vb
AlexSananka: Sorry what do you mean?I just changed the position of the functions here.I wanted the graph to come after the bootstrap.

dannyparsons: Yep sorry I was confusing this with call type.

instat/dlgOneVarFitModel.vb
dannyparsons: It changing this solves a bug then there is a bigger issue with the code because this shouldn't effect other parts of the code. This code should be here and then we should solve remaining issues.

AlexSananka: When we have this line in the SeDefaults every time you reset it runs this line and hence we have the bug.I am not sure what we should do to solve this.

dannyparsons: There is no problem this line running more than once. I suspect the issue is related to something else in the code. I will have a look at this from here so you can move on to other tasks.

instat/dlgExportRWorkspace.vb
```diff
@@ -40,6 +40,7 @@ Public Class dlgExportRWorkspace
         ucrReceiverMultiple.SetParameter(New RParameter("data_names", 0))
         ucrReceiverMultiple.SetParameterIsString()
         ucrReceiverMultiple.Selector = ucrSelectorForDataFrames
+        ucrReceiverMultiple.strSelectorHeading = "data frame(s)"
```
lilyclements: Can this just be "Data Frames"?
We have Variables not Variable(s) in places
And also need a capital D and F

AlexSananka: Okay making this minor edit.

AlexSananka: But wait,we put variables because we can't or it's rare to have a data set with one variable but in this case sometimes you might just have one data  frame in the data file.Will it make sense to have data frames when it's just one?

lilyclements: We have "Variables" display even when there is only one variable in the data set.
Or in the climatic dialogs, for example, we'll have "Date Variables" even though the user would only have one defined Date Variable. So for consistency that's why I was thinking "Data Frames"

Issue #3389, opened 2017-06-15
Delete Data Frame enhancement


dannyparsons:
This should change to have a selector and a multiple receiver so that multiple data frames can be deleted in one go.

This will affect the right click Delete Data Frame option so take this into consideration when updating the dialog.

PR #3713, opened 2017-07-19
Delete Data frames Dialogue changes

stevenndungu:


stevenndungu:
Fixes issue  #3389

instat/ucrDataView.vb
```diff
@@ -295,15 +295,16 @@ Public Class ucrDataView
...
+    Private Sub deleteSheet_Click(sender As Object, e As EventArgs) Handles deleteDataFrame.Click
+        'Dim strScript As String
+        'Dim Delete = MsgBox("Are you sure you want to delete this dataframe?" & Environment.NewLine & "This action cannot be undone.", MessageBoxButtons.YesNo, "Delete Sheet")
+        'If grdData.Worksheets.Count > 0 Then
+        '    If Delete = DialogResult.Yes Then
+        '        strScript = frmMain.clsRLink.strInstatDataObject & "$delete_dataframe(data_name =" & Chr(34) & grdData.CurrentWorksheet.Name & Chr(34) & ")"
+        '        RunScriptFromDataView(strScript)
+        '    End If
+        'End If
+        dlgDeleteDataFrames.ShowDialog()
```
dannyparsons: I would still like the auto filling of the selected data frame here even though it's now a multiple receiver. It should work in almost exactly the same way.

stevenndungu: Yeah, I have thought of adding at as well. But requires slightly different method from the one that is already existing. Thinking on how to have it added in as well. Thanks

dannyparsons: Just do as the autofilling is done for other right clicks. It will be exactly the same idea for this dialog. You don't need to think of a new way to do this.

stevenndungu: since we are autofilling data frames and not columns, it then implies this is to be done in a different way compared to the other dlgs. 

I have actually got stuck on this.

maxwellfundi: @stevenndungu  could you tag someone who can help with this, or should @dannyparsons  help with this?

stevenndungu:
"I would still like the auto filling of the selected data frame here even though it's now a multiple receiver. It should work in almost exactly the same way"

This is still remaining in this pull request. I have still not managed to set it as required. 
I just tested the dlg after updating the branch - I realised that while deleting multiple data frames it gives an error "object of type 'environment' is not subsettable". A problem with `delete_dataframe` method. This was not there before. 

This are the only two things remaining in this pull request.

dannyparsons:
merging this because of many file changes and vbproj edits, but not yet complete.

Issue #3456, opened 2017-06-20
The View Data using the R viewer


rdstern:
I hope this is a simple one too.  I am calling it s bug, but perhaps it is a feature!  It reflects the increased importance of the R viewer now, for large data sets.

For information: On my (quite slow machine) the R view took about 3 seconds to show the data for a set with 15000 rows and 145 columns.  It took 12 seconds for 1.3 million rows and 10 columns.  I think this is good.

1. When I use the R viewer from the tab at the bottom of the data view it takes about 3 seconds (good) and then it is displayed - **and it stays on top**!
When I try the same for the **Prepare > Data Frame > View data** I have to select the variables.  No problem, I Add them all with the right click.  It is just as quick, but the focus then returns to the main screen and the data are behind.  It is easy to find, but **is it also easy to keep it on top**?
2. What about one option on this dialogue to view the data frame completely?  This might now include the famous buttons at the top with **View All** and **Selected Columns/Rows**.  If View All, then there is just the data frame selector.  Otherwise there is the dialogue we have now.
(1 could usefully be examined now, i.e. for July.  2 is less urgent, but might as well be done only if really quick for someone)

rdstern:
If using 2 screens then this R data view is an example that I might want on the second screen.

dannyparsons:
1. My guess is that it doesn't stay on top when you use it from a dialog, possibly because when the dialog closes, the main form gets focus and this causes other windows to hide. We could investigate this but not sure what we could do about it if this is the issue.

dannyparsons:
2. Is now implemented.

PR #3714, opened 2017-07-19
Dlg View Data  Enhancement

stevenndungu:


stevenndungu:
@africanmathsinitiative/developers  this is ready for re-view
Check issue #3456

stevenndungu:
Noted a developer error on second reset.

lilyclements:
Reviewing this now

lilyclements:
I'm getting an error when I run this without pressing anything (except the View All option)

> utils::View(x=Survey, title="Survey")
> Error in as.data.frame(x) : object 'Survey' not found

If I run the code in R by itself then it runs fine.
If I run that line in R after the instat setup then I get the error

dannyparsons:
@stevenndungu can you respond to this?

instat/dlgView.vb
```diff
@@ -115,6 +131,7 @@ Public Class dlgView
         clsSeparateWindowFunction.SetPackageName("utils")
         clsSeparateWindowFunction.SetRCommand("View")
 
+        clsSeparateWindowFunction.AddParameter("x", ucrSelectorForView.ucrAvailableDataFrames.cboAvailableDataFrames.SelectedItem, iPosition:=0)
```
dannyparsons: This isn't right. We want the data frame from the selector not the name of it. Anyway, this shouldn't be done in SetDefaults, it's not a default.

lilyclements: I believe I removed this last week because it's being called in SetRCodes anyway. It's waiting to be merged here:
https://github.com/stevenndungu/R-Instat/pull/305

Sorry I thought I brought your attention to it already

lilyclements: I didn't remove this so Danny's comment still applies.

AlexSananka: I believe this is what is causing the bug @lilyclements was experiencing both here and `dlgduplicates`.
If this can be done automatically then it will solve the bug we were experiencing in both dialogs.

stevenndungu: Let, have a look at it once again. I can't remember the reason why I set it that way, because i remember doing it that way intentionally after experiencing some trouble..

dannyparsons: The only reason this "works" when you load a data frame from the library or from a file is that the variable still exists in the global environment. However, this is completely wrong because it's not taking the data from within the instat object, so it's displaying the wrong data. 
All you need to do is change this to get the R function with selector which will get the data.

stevenndungu: Ok. Making the necessary changes

maxwellfundi: @stevenndungu  has the necessary changes here been made?

Lunalo:
@stevenndungu 

Is this now ready for review?

lilyclements:
We still have an issue with the radio buttons not being set up correctly.

stevenndungu:
I don't see a better way of solving this, now that the commit cannot be reverted. Unless somebody has suggestions on how to fix this, I will close this branch and transfer the changes to another branch.

stevenndungu:
@maxwellfundi Note that, its is not part of the files changed at the top.Only two files are changed!
Replicated this branch on pull request  #3759

PR #3719, opened 2017-07-20
Duplicates dialog

AlexSananka:


AlexSananka:
@africanmathsinitiative/developers This is making good progress

AlexSananka:

lilyclements:
Reviewing this now

lilyclements:
I'm having the same issue here that I had with View for when I'm using just a data frame and not the variables.
I restarted everything and shut down RStudio and still getting the problem. Can you update your branch and see if this is occurring for you? Perhaps it's just my laptop.

AlexSananka:
Yes I know why this is happening,you must be using a dataset that is already pre-loaded(I mean a data set that is not loaded from library.).Can you try and find a dataset from the library and see if you get this error?

lilyclements:
Is this a known bug? I think an issue should be written up on this if there isn't already?

AlexSananka:
This is a known bug in at least two dialogs,unless someone has a different way which we can make the data frame to pass this parameter because for now we're doing it manually which is why we're having this bug.

instat/dlgDuplicatesConstructed.vb
```diff
@@ -33,18 +34,121 @@ Public Class dlgDuplicatesConstructed
...
+    End Sub
+
+    Private Sub ucrPnlOptions_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrPnlOptions.ControlValueChanged
+        DataFrameParameter()
+    End Sub
+
+    Private Sub DataFrameParameter()
+        If rdoDataFrame.Checked Then
+            clsDuplicate2.AddParameter("x", ucrSelectorDuplicateswithVariables.ucrAvailableDataFrames.cboAvailableDataFrames.SelectedItem)
+            clsDuplicated.AddParameter("x", ucrSelectorDuplicateswithVariables.ucrAvailableDataFrames.cboAvailableDataFrames.SelectedItem)
```
dannyparsons: This is why you're getting the error with the data. It's not a general bug, you are passing the data frame name, not the data.

AlexSananka: Yes sure I was having a problem passing this,maybe we can skype when I have sometime so that I can explain what was happening?

dannyparsons: You just need to pass the data frame from the selector. Here you are only passing the name of the data frame.

maxwellfundi:
@AlexSananka has this been addressed?

AlexSananka:
I am not sure if I can get time to fix this,but it's something easy for anyone to take up and fix quickly.

Issue #3725, opened 2017-07-21
Problem with the data selector with multiple columns


rdstern:
In our data selector I am now usually using data files with more columns than can be shown in the selector.
When I select a column to transfer it into the receiver the selector always returns to the start of the set of columns.  Could it please stay where it is!

If easy it might be that an option could also be to have the variables in the selector in either Data Frame order or Alphabetical order.

lilyclements:
This is solved!

PR #3726, opened 2017-07-24
Fixes the problem with the data selector with multiple columns

stevenndungu:
This perfectly fixes issue #3725


stevenndungu:
Note: I have chose to have the variables in the selector retain the Data Frame order

dannyparsons:
This does solve the issue, though we lose some functionality that is nice. Having the focus on the receiver after adding is natural for tabbing etc. We might need a more complex solution. I will leave this open so we can discuss this after thinking about it more.

Lunalo:
@stevenndungu  @dannyparsons 

How far?

dannyparsons:
This needs some more thought. I think it would be still be good to fix for 0.4.1.

shadrackkibet:
@dannyparsons  ?

dannyparsons:
I don't think we have a good solution for this yet. I think we can close this as we have the issue still open, which I have set to 0.x milestone.

PR #3727, opened 2017-07-24
Climatic summary

shadrackkibet:
@africanmathsinitiative/developers in  progress.

PR #3730, opened 2017-07-25
Dlg Summary of Daily Data 

stevenndungu:


stevenndungu:
Discussed with Roger and Lily . This is similar to pull request #3727. There was a problem with replication of an issue, by mistake

Issue #3508, opened 2017-06-30
Extremes for peaks.  


volloholic:
Extremes for peaks.  It is currently just a filter.  That’s fine, but could it also only include just the columns that are in the dialogue – or at least have that as an option – and it would be the default.

shadrackkibet:
This relates to issue #3201 of a feature request. @dannyparsons can this now be implemented?

dannyparsons:
There's some work to implement this still. It's not urgent until the 0.4.1 release i.e. August

lilyclements:
@dannyparsons is there still some work needed before this can be implemented?

PR #3734, opened 2017-07-26
Updating Extremes Climatic

lilyclements:


instat/dlgExtremesClimatic.vb
```diff
@@ -266,7 +266,7 @@ Public Class dlgExtremesClimatic
                 clsMinMaxFuncExp.SetRCommand("max")
             End If
             clsMinMaxSummariseFunction.AddParameter("function_exp", clsRFunctionParameter:=clsMinMaxFuncExp, iPosition:=1) ' TODO1: fix this so that we don't lose the function_exp when we change the r function. Should we do two bases here clsMaxFuncExp and clsMinFuncExp and then the problem is similar to TODO1. We want to change the class being called for function_exp when the radio buttons are changed
-        Else
+        ElseIf rdoPeaks.Checked Then
```
lilyclements: @dannyparsons on this dialog there are two radio buttons: Max/Min and Peaks.
Max/Min is to be checked by default and this occurs upon pressing reset. However, upon first opening I was having the issue that Peaks was checked.
Through tracing back, I found this problem occurred if I had "Else" in this sub as opposed to "ElseIf rdoPeaks.Checked".

I do not fully understand why that caused this bug. Why do I need to explicitly say which radio button rather than just "Else"?

lilyclements:

You may notice two minor TODO's in the same sub where I've done a temp. fix. These shouldn't affect the review.
However, if anyone gets some spare time then any help would be appreciated, particularly on TODO1 (I know TODO2 is an issue on other dialogs). Alternatively I can talk to Daniel about it when he's back next week. These temp. fixes aren't introducing any new bugs (that I know of).

lilyclements:
@africanmathsinitiative/developers this is still ready to review

dannyparsons:
If it's doing something different on Else as opposed to ElseIf it means that no radio button is checked. This is why the ElseIf is safer than the Else. The Else doesn't mean that the other radio button is checked.

lilyclements:
Well, I have learnt my lesson and am converted to using `ElseIf`'s now. Thanks.

maxwellfundi:
@lilyclements  whats the state of this?

lilyclements:
@africanmathsinitiative/developers this is still waiting for review.

maxwellfundi:
Great, @shadrackkibet  could you please help with reviewing this?

shadrackkibet:
Reviewing this now.

instat/dlgExtremesClimatic.vb
```diff
@@ -66,111 +72,128 @@ Public Class dlgExtremesClimatic
         ucrReceiverDOY.AddIncludedMetadataProperty("Climatic_Type", {Chr(34) & "doy" & Chr(34)})
         ucrReceiverDOY.strSelectorHeading = "Day variables"
 
-        ucrReceiverElement.SetParameter(New RParameter("x", 0))
+        ucrReceiverElement.SetParameter(New RParameter("x", 0, bNewIncludeArgumentName:=False))
```
shadrackkibet: I think this should be restricted to numerics.

instat/dlgExtremesClimatic.vb
```diff
@@ -40,137 +46,197 @@ Public Class dlgExtremesClimatic
...
+        clsDayFromAndTo.AddParameter("type", Chr(34) & "filter" & Chr(34), iPosition:=0)
+        clsDayFromAndTo.AddParameter("function_exp", clsROperatorParameter:=clsDayFromAndToOperator, iPosition:=1)
+        clsDayFromAndToOperator.SetOperation("&")
+        clsDayFromAndToOperator.AddParameter("from_operator", clsROperatorParameter:=clsDayFromOperator, iPosition:=0)
+        clsDayFromOperator.SetOperation(">=")
+        clsDayFromOperator.AddParameter("from", 1, iPosition:=1)
+        clsDayFromAndToOperator.AddParameter("to_operator", clsROperatorParameter:=clsDayToOperator, iPosition:=1)
+        clsDayToOperator.SetOperation("<=")
+        clsDayToOperator.AddParameter("to", 366, iPosition:=1)
+        clsDayFromAndTo.SetAssignTo("Day_From_and_To")
```
shadrackkibet: Naming convention?

shadrackkibet:
Issues to address:

- The From and To Nud's should work in a way that From number < To number.This is affects some climatic dialogs which uses this  filter.I think this is possible to control on the code.

- Peaks is still just a filter.I think it is now time to address issue #3508

lilyclements:
Thanks @shadrackkibet I've addressed your issues and this is ready for a re-review.

As for issue #3508 - these are changes still to be made in the R-Code. Daniel is about to work on that, so I'll talk to him about the priorities, i.e., is it worth me spending time on implementing that or on fixing up other dialogs.

lilyclements:
I don't mind that the user can do the to/from values in the wrong order because they can see it's incorrect and it doesn't give an error.
I do not like the nuds since they are difficult to type values into. Adding a restriction that `From < To`, it makes it even more difficult to type into it.
Since having this restriction makes the nuds even more frustrating, and since it doesn't give an error to have it the "wrong way", I don't mind that the user can do From > To.
Perhaps this is something to discuss in an issue to hear others thoughts on it?
Or perhaps they shouldn't be nuds but inputs since they have restrictions.

What do you think? I think perhaps this is something to discuss after Friday since if we change this on one of the climatics we have to change it on all.

dannyparsons:
It's a good discussion and I think leave it as it is currently implemented now and discuss changes for the next version.

shadrackkibet:
Re-reviewing this.

shadrackkibet:
Noted the following;

- When doing min/max , i think it is useful that the resulting dataframe should always start with stations when using data with stations. Is this possible to sort out?

- The include Missing Values checkbox isn't resetting.

lilyclements:
For the station/year point:
This is something that would be easy to do, because all we need to do is switch the Year/Station options in the grouping class. However, after a discussion with Danny we've agreed that this is something which would need to be changed on all climatic dialogs if we change it here.
It's a good point, however, due to time we'll keep it like this for now. Perhaps it's something to write an issue up on to discuss for the next release?

I've fixed the resetting issue

dannyparsons:
> When doing min/max , i think it is useful that the resulting dataframe should always start with stations when using data with stations. Is this possible to sort out?

That's definitely possible and a useful point. I can see a user may sometimes want that order and but sometimes want the other order. So I'm not sure I agree with you that it should **always** start with stations, unless you have a good reason for this?

I suggest that what is ideal is to allow the user to specify the order and not have it fixed either way. Possibly a 'global climatic option' somewhere because it's done from many dialogs. This could be a good enhancement to think about. I'm not too worried about it right now though because the user can always use the Sort dialog to change the order after creating it. I do think it's worth writing an issue up on it though so we do it in future.

PR #3735, opened 2017-07-27
Transform climatic

lilyclements:
In progress - only (foreseeable) issue is with radio buttons. 
EDIT: @africanmathsinitiative/developers This is now ready to review.


instat/dlgTransformClimatic.vb
```diff
@@ -38,9 +52,20 @@ Public Class dlgTransformClimatic
...
+        ucrPnlTransform.SetParameter(New RParameter("function_exp", 1))
+        ucrPnlTransform.AddRadioButton(rdoMoving, "rollapply")
+        ucrPnlTransform.AddRadioButton(rdoCount, "rollapply")
+        ucrPnlTransform.AddRadioButton(rdoSpell, Chr(34) & "cumsum(" & strRainDay & ")-cummax((" & strRainDay & "==0)*cumsum(" & strRainDay & "))" & Chr(34))
+        ucrPnlTransform.AddRadioButton(rdoWaterBalance, "Reduce")
+
+        'ucrPnlTransform.AddParameterValueFunctionNamesCondition(rdoMoving, "sub1", "instat_calculation$new", False) ' clsRRainday
+        'ucrPnlTransform.AddParameterValueFunctionNamesCondition(rdoCount, "sub1", "instat_calculation$new")
+        'ucrPnlTransform.AddParameterValueFunctionNamesCondition(rdoSpell, "sub1", "instat_calculation$new")
+        'ucrPnlTransform.AddParameterValueFunctionNamesCondition(rdoWaterBalance, "sub1", "instat_calculation$new", False)
```
lilyclements: @dannyparsons can we discuss these radio buttons? (Line 57 to 67)

Two things change when I change the radio buttons:
1. the function_exp of the base function changes to one of four classes (one class per radio button)
2. the sub_calculations of the main function is added or removed for the same class

You can see I've *attempted* my implementation of 1 and 2 respectively.

My issues are as such:
* For 1: The Moving and Count rdos both have that the `function_exp` equals the class which runs `rollapply` as the function. There are two different classes I'm referring to here: `clsCountFunction` and `clsMovingFunction`. How else can I differentiate between the two classes if their function has the same name?
* For 2: Many an item are called `instat_calculation$new` and so I haven't been able to specify which one I'm referring to here. While I can connect to the correct class (`clsRRainday`) via SetRCodes, I also want to connect this panel another class due to 1. If I connect to both classes, how will it know I don't want it to add/remove the instat_calculation$new function for the other class?


PS, yes I am running that lovely long string as a string. I have my justifications ready.

Lunalo:
@lilyclements 

How far are yyou with this?

lilyclements:
I believe the main remaining issue here is radio buttons. Once that's resolved, this shouldn't be far off from being finished.

Lunalo:
Great

lilyclements:

shadrackkibet:
Reviewing this now.

shadrackkibet:
Few things to note here:

- On **Moving** the Nud & the combobox isn't resetting.

- On the different selected items on the Combobox, should we change naming conventions for the resulting column name? I find "moving_Sum" not really nice. should we have something like "Moving_Sum" or "moving_sum"? then this is applied to others here.

-**On count & Spell ** .I think we should change the content of the combobox i.e >="Amount of Rain" to be just ">=" etc, Roger had suggested this on Spells and we changed. This will result into the size of the combobox becoming smaller.

lilyclements:
* Good eye. I'll fix the resetting
* Okay, so it's moving_Sum not moving_sum because it's copying the name in the combo box so code-wise it's "shorter". However, that has been bugging me, and so I agree. I'll change that.
* Good point on three, I was meaning to do that. Thanks!

instat/dlgTransformClimatic.vb
```diff
@@ -38,203 +50,253 @@ Public Class dlgTransformClimatic
...
-        ucrInputColName.SetParameter(New RParameter("result_name", 4))
-        ucrInputColName.SetName("moving_" & ucrInputSum.cboInput.SelectedItem)
-        ucrInputColName.bUserTyped = False
+        ' Save Options
+        ucrInputColName.SetParameter(New RParameter("result_name", 2))
+        ucrInputColName.SetName("moving_sum")
 
-        clsSubCalcList.SetRCommand("list")
+        ucrPnlTransform.AddToLinkedControls(ucrInputSum, {rdoMoving}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:="Sum")
+        ucrPnlTransform.AddToLinkedControls(ucrNudSumOver, {rdoMoving}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=2)
```
lilyclements: @africanmathsinitiative/developers I need to run `bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=...` for these two controls here because otherwise if I run another rdo, then switch back to this rdo (the default one!), it doesn't read in the parameter value.

However, this also prevents resetting from working properly. How do I overcome this problem?

lilyclements:

shadrackkibet:

PR #3739, opened 2017-07-31
Climatic boxplots

maxwellfundi:
Ready for review


maxwellfundi:
@rdstern  this is now the design for the climatic boxplot. Please let me know what you think. 

It is currently producing  boxplots of X as years and Y as rain which is autofilled on the data receiver. I am working to get the other functionalities. 

image

rdstern:
That's starting to look nice.
1. Could you change the "Month" label to "Within Year:":  The default is to fill it with the **Month** column.
2. I assume the "More Data checkbox would open a single receiver if checked?  That is what is needed.
3. In the layout section:
a) Instead of the first 2 check-boxes have 2 radio buttons.  They are only enabled if there are 2 data columns.  One says "Year then Data" and the other says "Data then Year".  
(Example - try with the ordinary boxplot.  If you have max and min temperatures as your data, then one layout would give max then min for each year.  The other would give max for all the years and then min for all the years.  Max and min for each year is the default, which is the option "Data then Year.
b) The label Month should become **Facet.**  This opens a single receiver with the default being Month.
c) The label Station should become **2nd Facet**.  If checked this opens a receiver with the default being the Station column.

maxwellfundi:
Great, thanks for this clarification. I am gretting them added now.

maxwellfundi:
@dannyparsons On this two things are still left. 
1. Faceting with single variable, for some reason , it does not pass in the operator properly for some reason. 
2.  The functionality Roger wanted with the radio buttons on the layout have not been implemented

lilyclements:
@maxwellfundi how is this getting on?

maxwellfundi:
@lilyclements  I had a bit of things pending to be done. Am now back to Maseno i hope to finish them off by the end of the week.

dannyparsons:
We want at least an initial implementation of this for the next release so that sounds good.

dannyparsons:
Tomorrow we are both travelling so I hope this will be finished or almost finished today so that it will be in the release?

instat/dlgClimaticBoxPlot.vb
```diff
@@ -274,12 +273,14 @@ Public Class dlgClimaticBoxPlot
...
-            clsFacetFunction.SetRCommand("facet_wrap")
-        ElseIf (ucrChkFacet.Checked AndAlso Not ucrReceiverFacet.IsEmpty) AndAlso (ucrChk2ndFacet.Checked AndAlso Not ucrReceiver2ndFacet.IsEmpty) Then
-            clsFacetFunction.SetRCommand("facet_grip")
+        If ucrChkFacet.Checked Then
+            If (Not ucrReceiverFacet.IsEmpty) AndAlso Not ucrChk2ndFacet.Checked Then
+                clsFacetFunction.SetRCommand("facet_wrap")
+            ElseIf (Not ucrReceiverFacet.IsEmpty) AndAlso (ucrChk2ndFacet.Checked AndAlso Not ucrReceiver2ndFacet.IsEmpty) Then
+                clsFacetFunction.SetRCommand("facet_grid")
+            End If
+            clsFacetFunction.AddParameter(clsROperatorParameter:=clsFacetOp)
```
maxwellfundi: @dannyparsons  this dialog is almost finished but i would like your help in setting up these facets properly. the case of the both receivers filled works great, but not for the first one.

maxwellfundi:
@rdstern  we have been thinking about how we could restructure this dialog with @dannyparsons to be much more efficient. 
Now for the two ways of producing graphs of Data than Year or Year then Data, we produced these two graphs and i was wondering which is more sensible to have, if we have this 

image

or this which is a graph with both Tmax and Tmin like this

image

rdstern:
Both are possible but I find the second to be more useful.  It will be clearer with the option to not display the outliers as such, and possibly adding a line connecting the medians.

maxwellfundi:
Okay great.  We are going to restructure this dialog now do that kind of graph.

Thanks.

rdstern:
There is the same issue on the main boxplot dialogue.  It is well designed to do the first of the 2 options, and needs a checkbox to be able to swap the order of the x-variables

maxwellfundi:
We willhave to put that there actually to be able to swap the axis. Here is the new design for this

image

lilyclements:
There's a few alignment edits there that I would like to fix up. Let me know when I can do so :)

maxwellfundi:

Thanks.

maxwellfundi:
@rdstern  I am not sure if the within year is being used in plotting the graphs. Whats the purpose of the within year?

rdstern:
You might have the year as the x and then the within year (e.g. month) as the facet.  

Or you could have the month as the x and that's all.  In that case you might alternatively want the dekad or the week as the x.

If you do have the month as the x, you might have the year, or the station, as the facet.  Or you could have the station and month as 2 x's.  Probably then too much to have the year as the facet, but not impossible.

maxwellfundi:
@rdstern  This is now  stable for testing.

instat/dlgClimaticBoxPlot.vb
```diff
@@ -26,24 +45,338 @@ Public Class dlgClimaticBoxPlot
             SetDefaults()
         End If
         SetRCodeForControls(bReset)
+        If bReset Then
+            AutoFill()
+        End If
```
lilyclements: How come you are running `If bReset Then` twice?

dannyparsons: The autofill needs to be run after SetRCode but only on the first load. (We don't want to auto fill on reopen).

dannyparsons:
Do you have the command for the graph where the tmin and tmax were the same colour? That would be useful to see unless @maxwellfundi has already solved it?

maxwellfundi:
@lilyclements  the jitter needs colour parameter not fill and thats something i had not noticed. Thanks for bringing out the facets thing because its an error on the sdgplots too. I wonder why we have not noticed this for all this long!

maxwellfundi:
@lilyclements this is ready for a re-review. All issues have been addressed now.

instat/dlgClimaticBoxPlot.vb
lilyclements: I see, thanks. I didn't properly look into this autofill before. It's nice that it autofills, however, I am a little confused that it autofills the element as rain.
I was talking to @rdstern the other day about adjusted boxplots and such because he said that a regular boxplot isn't appropriate for rainfall. How come we are then autofilling Rain to be in my Boxplot receiver when I also have other variables such as Tmin, Tmax, Sunh?

lilyclements: I think that I am asking more of a statistical question than a programming question!

instat/dlgClimaticBoxPlot.vb
```diff
@@ -17,33 +17,396 @@ Imports instat.Translations
...
+        AutoFill()
+    End Sub
+
+    Private Sub CoreControls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrSavePlot.ControlContentsChanged, ucrVariablesAsFactorForClimaticBoxplot.ControlContentsChanged
+        TestOKEnabled()
+    End Sub
+    Private Sub FacetsCheck()
+        If Not ucrReceiverFacetBy.IsEmpty AndAlso Not ucrReceiver2ndFacet.IsEmpty Then
+            If ucrReceiverFacetBy.txtReceiverSingle.Text = ucrReceiver2ndFacet.txtReceiverSingle.Text Then
+                MsgBox("You can not do facets with two same variables", vbOKOnly)
```
lilyclements: Can this change to:
"You cannot do facets with two of the same variables"

lilyclements:
Final couple of issues:
1. If I check margins and press OK - it doesn't seem to be giving me any margin
2. If I check margins, press OK and reopen - it is unchecked
3. If I check "Vertical X Tick Markers", close and reopen - it is unchecked

Finally! More a technical question because I'm intrigued now.
If I run the dialog (jitter in this case) and select "Single Variables" on the ucrVariatesToFactor and input one variable, it runs a different code to if I were to select "Multiple Variables" on this ucr and input only one variable.
Why is this? I can understand why `colour = ` needs to be added, but I'm interested why the rest of it needs to be different.

`# Code generated with "Single Variables"
Survey <- InstatDataObject$get_data_frame(data_name="Survey")
jitter1 <- ggplot2::ggplot(data=Survey, mapping=ggplot2::aes(y=Field, x=as.factor(x=Variety.))) + ggplot2::geom_jitter() + theme_grey() + ggplot2::facet_wrap( ~ Village.)

# Code generated with "Multiple Variables"
Survey <- InstatDataObject$get_data_frame(data_name="Survey", stack_data=TRUE, id.vars=c("Variety.","Village.","Field"), measure.vars="Field")
jitter2 <- ggplot2::ggplot(data=Survey, mapping=ggplot2::aes(y=value, x=as.factor(x=Variety.), colour=variable)) + ggplot2::geom_jitter() + theme_grey() + ggplot2::facet_wrap( ~ Village.)`

maxwellfundi:
@lilyclements  the above issue on id vars is yet to be fixed i believe and @dannyparsons  is well aware of this. We will be fixing that eventually.

maxwellfundi:
@lilyclements  The issues raised here have been fixed now. have another test.

instat/dlgClimaticBoxPlot.vb
```diff
@@ -17,33 +17,395 @@ Imports instat.Translations
...
+        ucrReceiverFacetBy.Selector = ucrSelectorClimaticBoxPlot
+        ucrReceiverFacetBy.SetParameterIsString()
+        ucrReceiverFacetBy.bWithQuotes = False
+
+        ucrReceiver2ndFacet.SetParameter(New RParameter("var2", 1))
+        ucrReceiver2ndFacet.Selector = ucrSelectorClimaticBoxPlot
+        ucrReceiver2ndFacet.SetParameterIsString()
+        ucrReceiver2ndFacet.bWithQuotes = False
+
+        ucrChkVarWidth.SetParameter(New RParameter("varwidth"))
```
lilyclements: Very minor - but this needs a parameter position

instat/dlgClimaticBoxPlot.vb
```diff
@@ -17,33 +17,395 @@ Imports instat.Translations
...
+        ucrChkVarWidth.SetRDefault("FALSE")
+
+        clsCoordFlipFunc.SetPackageName("ggplot2")
+        clsCoordFlipFunc.SetRCommand("coord_flip")
+        clsCoordFlipParam.SetArgumentName("coord_flip")
+        clsCoordFlipParam.SetArgument(clsCoordFlipFunc)
+        ucrChkHorizontalBoxplot.SetParameter(clsCoordFlipParam, bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)
+        ucrChkHorizontalBoxplot.SetText("Horizontal Plot")
+
+        ucrChkMargins.SetParameter(New RParameter("margins"))
```
lilyclements: This also needs a parameter position

lilyclements: Parameter position also needed line 131, and also in some of the `AddParameter`'s in `SetDefaults`, and in other subs later in the dialog (at least another four subs?)

instat/dlgClimaticBoxPlot.vb
```diff
@@ -17,33 +17,395 @@ Imports instat.Translations
...
+            clsRaesFunction.AddParameter("x", Chr(34) & Chr(34))
+        Else
+            clsRaesFunction.AddParameter("x", clsRFunctionParameter:=clsAsFactor)
+        End If
+    End Sub
+
+    Private Sub ucrVariablesAsFactorForClimaticBoxplot_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrVariablesAsFactorForClimaticBoxplot.ControlValueChanged
+        SetColourFillAes()
+    End Sub
+    Private Sub SetColourFillAes()
```
lilyclements: Can you add a line break between the end of the last sub and start of the next?

instat/dlgClimaticBoxPlot.vb
```diff
@@ -17,33 +17,395 @@ Imports instat.Translations
...
+    End Sub
+
+    Private Sub ucrSelectorClimaticBoxPlot_DataFrameChanged() Handles ucrSelectorClimaticBoxPlot.DataFrameChanged
+        AutoFill()
+    End Sub
+
+    Private Sub CoreControls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrSavePlot.ControlContentsChanged, ucrVariablesAsFactorForClimaticBoxplot.ControlContentsChanged
+        TestOKEnabled()
+    End Sub
+    Private Sub FacetsCheck()
```
lilyclements: And can you add another line break here?

maxwellfundi:
@dannyparsons  this is now ready for you!

Issue #3740, opened 2017-07-31
New Column Names - Capitalised first letters?


lilyclements:
There is an inconsistency among the software of new column names. I think we need to clarify what formatting our new column names use.
e.g. I have a new column name called "New Name" we have used different formatting layouts, such as:

new_name
newName
NewName
new.name

I think the one I like the most is the one which is implemented the least which isn't ideal and suggests others disagree to it. So, I don't have any preference - I just want this to be consistent.
What should we do in the case of "new name"?

dannyparsons:
Can you give an example so I'm clear what you mean? We normally have just one word for default column names I thought?

lilyclements:
I see your one point, and raise three:

1. For one worded columns would we want a capital first letter?

2. In dlgTransformClimatic, there are column names called "moving_sum", "moving_mean", etc. are you suggesting these should just be one word?

3. In addition, with any ucrSave can we have consistency. For example, sometimes the plot names have capitals and sometimes they do not. Also we have "last_plot" but then we have "frequencypolygon1" so consistency with "_" and capitals and what not would be nice.

dannyparsons:
I think because these are R names we should follow something that's sensible in R, which any of your original examples are. Lower case and underscores for spaces is used quite a lot e.g. `ggplot2` so I think going for that in R-Instat is fine.

lilyclements:
Great. I know some dialogs have a capital first letter. I won't go through everything but will check when editing dialogs in the future for this.

Another question - so when we assign something to our instat_calculation$new function, should that have a capital first letter? I know I've used day_from_and_to in one place and Day_From_And_To in another.

dannyparsons:
Those are R variables as well so I think should be the same formatting without capitals

lilyclements:
Great. Those are just things I need to fix in my climatic dialogs I'm on (Start of Rains, End of Rains, Extremes and Transform)
@shadrackkibet @stevenndungu in any climatic that you're working on now can you make sure that your R-variables are non-capitalised?

dannyparsons:
@africanmathsinitiative/developers let's check this on dialog before the next release

PR #3742, opened 2017-07-31
dlgSpells

### LINKED ISSUE #2772: Simple dry spells dialogue
I suggest a simple dialogue for now - as sort of "partner" to your start dialogue.  I suggest that both (and water balance!) will change, but these will be useful.

1. One missing aspect (to ignore for now) is that there is no mention of "Station" in the existing start dialogue.  For now I suggest similarly for the spells one.  Need to think about this later.  

2. Also, in the start, I don't quite understand the need for the Year and day of year fields?

3. For the start, I assume the resulting worksheet is linked, so another definition would go into the same sheet?  Hence a similar calculation of spell lengths would also (automatically?) go into the same sheet?

4. So, in the new dialogue, instead of threshold you have "Condition".  This has 2 parts to it, with the first having just 3 options, namely "<", "Between" and ">".  Default is "Between" and this then has 2 text boxes open with default values 0 and 0.85.  If < the second box is hidden, and if > the first box is hidden.

5. The "From" and "To" fields remain as you have them.  perhaps they could instead say "From Day:" and "To Day:" - on both dialogues.

6.  So, for now, the spells dialogue will produce just a single column - as does the start of the rains.

7.  There is a checkbox underneath - default unchecked, with label: "Conditional on Rain at start of spell".  If checked this starts the count of the spell length only as 1 on the first day.

That's all for now!

shadrackkibet:
@africanmathsinitiative/developers  this is ready for review. Fixes #2772.


lilyclements:
Reviewing this now

maxwellfundi:
@lilyclements  i was just about volunteering you to test this.

instat/dlgSpells.vb
```diff
@@ -48,13 +48,13 @@ Public Class dlgSpells
...
-        ucrReceiverRainfall.AddIncludedMetadataProperty("Climatic_Type", {Chr(34) & "rain" & Chr(34)})
-        ucrReceiverRainfall.bAutoFill = True
-        ucrReceiverRainfall.strSelectorHeading = "Rain Variables"
+        ucrReceiverElement.SetParameter(New RParameter("rain", 0))
+        ucrReceiverElement.SetParameterIsRFunction()
+        ucrReceiverElement.bWithQuotes = False
+        ucrReceiverElement.Selector = ucrSelectorForSpells
+        'ucrReceiverElement.AddIncludedMetadataProperty("Climatic_Type", {Chr(34) & "rain" & Chr(34)})
+        ucrReceiverElement.bAutoFill = True
+        ucrReceiverElement.strSelectorHeading = "Variables"
```
lilyclements: This needs to accept only numeric variables.

shadrackkibet:
@rdstern can you clarify this.

lilyclements:
It will need to be a numeric as we are looking at values in a range and can't do that for a factor.
Presumably it can be an integer as well.

shadrackkibet:
what about the issue of Nuds being changed to receivers? I didn't see this at #2772 latest issues to address.

shadrackkibet:
@lilyclements  any other issues on this? apart from the issue of changing From and To Nuds to Receivers which is not yet clear?

lilyclements:
Nope that's it I believe. It looks good other than that

lilyclements:
In light of #3740 can you change the column name here from "Spells" to "spells"?

maxwellfundi:
@shadrackkibet  How is it going with this?

shadrackkibet:
@lilyclements  what can you say about this? is this ready for @dannyparsons ?

instat/dlgSpells.vb
```diff
@@ -275,12 +276,12 @@ Public Class dlgSpells
     End Sub
 
     Private Sub ucrBase_BeforeClickOk(sender As Object, e As EventArgs) Handles ucrBase.BeforeClickOk
-        frmMain.clsRLink.RunScript(clsAddKey.ToScript, strComment:="Spells: Defining column(s) as key")
+        frmMain.clsRLink.RunScript(clsAddKey.ToScript, strComment:="spells: Defining column(s) as key")
```
lilyclements: Can this spells still be a capital? This is the start of a sentence rather than the column name

shadrackkibet:
Sorted.

lilyclements:
excellent, @dannyparsons this is ready to review

dannyparsons:
Why are there changes to sdgCanonicalCorrelation designer? Is this going to cause conflicts with another pull request?

shadrackkibet:
I did  not intend to do this designer edits, I have therefore reverted that commit.

dannyparsons:
Checking all the file changes on Visual Studio before committing will make sure you don't make changes to files accidentally.

Issue #3498, opened 2017-06-29
Climatic > Define Climatic data and also the Use Date dialogue


rdstern:
This is a minor change in the dialog.

1. When there is a shifted year (and a shifted day of year) then it would be good if the Define Climatic data could pick that up - rather than the ordinary year.  The names could be "syear" and "sdoy", see below.
2. It currently picks up the rainfall and temperature columns automatically.  If the sunshine hours is labelled "sunh" could that also be picked up automatically?
3. In the Use Date dialogue the shifted year column is currently called **shift_year**.  (The user has no choice - which is OK) This could be changed to be **syear**.  The day of year is currently called **shift_214** and this could become **sdoy**. 
4. This may be later - in which case leave the issue open) or it may be simple to do now.  For the column syear add the **variable label** "Shifted year starting August".  (Of course depends on the month.  If it is simpler, then give the day number for now, in the label i.e. "Shifted year by 214 days".) Similarly for the day of year, give the label "Shifted day of year by 214 days".

Here is an example: I use dodoma from the climatic datasets
1. I use Dates > Make Date to provide a date column
2. I use Dates > Use Date with the default name to make a shifted year column and day of year column.  I keep the default names, which are (currently named) shift_year and shift_214.  
3. Now I use the Define Climatic Data.

dannyparsons:
@stevenndungu @maxwellfundi can you look at 2, 3 and 4? Should be fairly simple to do.

3. I think `s_year` and `s_doy` to be consistent with our naming convention if that will still be OK.
4. Should be simple too. We have access to both the day number and the day/month combination so could do "Shifted year starting 1 August" just as easily, which I guess is preferable?

dannyparsons:
2, 3 and 4 now done. 1 we should discuss @rdstern

dannyparsons:
Point 1. moved to #3815 so this can be closed.

PR #3752, opened 2017-08-02
Define Climatic data and  Use Date dialogues

stevenndungu:


stevenndungu:
Fixes issue #3498
Note: This is working correctly as indicated on the issue but still not sure why I have "=" before the labels as shown here 
image

PR #3759, opened 2017-08-02
Dlg View - Enhancement 

stevenndungu:


stevenndungu:
I explained the problem to this to @dannyparsons on skype. 
State of the dialogue: - It is working properly only that how some parameter (`x` parameter of  ` View` function) is passed is debatable.

dannyparsons:
@lilyclements could you look at this? I think it should just be passing the data frame correctly.

PR #3761, opened 2017-08-04
Display daily dialog

Lunalo:


dannyparsons:
This is the R code for the tables:

`InstatDataObject$summary_table(data_name="Dodoma", summaries=sum_label, columns_to_summarise = c("Rain"),  factors=c("Year", "Month", "Day"), n_column_factors=1, include_margins=TRUE, rnames=FALSE, as_html = FALSE, page_by = "Year", store_results = FALSE)`

Some of these parameter will NOT be options on the dialog. `summaries` is always `sum_label`, `page_by` is always the year column and `n_column_factors = 1` always. `stored_results = FALSE` always as we are not doing any "new" summaries.

The order the factors appear in the list is important so that the rows are days and columns are months in the table.
`include_margins`, `rnames` and `as_html` are all options. Check with @rdstern about the defaults for these.

Lunalo:
Okay , Great.

Thanks

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1202,6 +1202,21 @@ instat_object$set("public", "get_climatic_column_name", function(data_name, col_
...
+        } else {
+          return(ggplot(data_name, aes(x=doy, y=climatic_element)) + geom_bar(stat="identity", fill="blue") + geom_rug(data=subset(data_name, is.na(climatic_element)==1), mapping = aes(x=doy), sides="b", color="red") + theme_minimal() + coord_cartesian(ylim=c(0,100)) + scale_x_continuous(breaks=c(1,32, 61, 92, 122, 153, 183, 214, 245, 275, 306, 336, 367), labels = c(month.abb,""), limits =c(0,367)) + facet_wrap(~year,ncol=2) + ggtitle(paste0(station_name, " Daily ", climatic_elementfall)) + theme(panel.grid.minor = element_blank(), plot.title = element_text(hjust = 0.5, size=20), axis.title = element_text(size=16)) + xlab("Day of the year") + ylab(climatic_element))
```
dannyparsons: This will need to be adapted a bit to work properly from the code David made. For example, data and columns need to be extracted from the instat object from the parameters passed in.
I can work on this method and that might be more efficient now, but if you can get the dialog linked up correctly that is needed.

Lunalo:
OKay thanks. I will continue with the linking

dannyparsons:
Great, let me know when you get to a point where I should take over on the R code.

dannyparsons:
@Lunalo where are things on this? I have some time today to look at the R method but then I'm travelling tomorrow. So if we don't get most of this done today then it's not going to be ready for the release.

Lunalo:
I think if you show me by one example on how we need to extract parameters from the instat object then I can do the rest and ask for help in case  of anything . 

You can just get the r code above and I can still pull your changes on that . thanks

dannyparsons:
The example to look at which I mentioned before is the inventory plot function and discuss with @maxwellfundi about doing this. Currently you have an instat object method only. You need both a data object and instat object method, with the actually code going in the data object method. All this is similar to the inventory plot code which should be your starting point.

Have a go at this but let me know how you are progressing so that it's not too late for me to help out.

Lunalo:
Great ... I now get it . thanks

dannyparsons:
The work from this has now been merged through other branches.

Issue #3525, opened 2017-06-30
climatic indices


volloholic:
Check on climatic indices dialog and consider subsequent analyses.

dannyparsons:
will be in a working state when #3778 is merged. More improvements are needed and nothing on subsequent analysis yet. I suggest that should be a separate issue from getting this dialog in a working state.

lilyclements:
Closing as requested

PR #3778, opened 2017-08-09
Climdex fix

### LINKED ISSUE #3226: Climdex bug
So I believe the issue is occurring because of a core problem. If I run the contents of this dialog on a branch from a few days ago, it runs fine - although it does not give any output. I'm not sure if this is what we expect by just pressing OK on the dialog.
If I run the contents of this dialog on an updated branch, I get the following error:

> "no applicable method for 'as.PCICt' applied to an object of class "character""
> The error occurred in attempting to run the following R command(s):
> climdex_input <- climdex.pcic::climdexInput.raw(prec.dates=PCICt::as.PCICt(x=as.character(x=InstatDataObject$get_columns_from_data(data_name="Damango", col_name="Date1")), cal="gregorian"), prec=InstatDataObject$get_columns_from_data(data_name="Damango", col_name="Rain"))
> 

There are also clear issues in the setup for `SetRCodeForControls` which need to be addressed!

lilyclements:
fixes #3226


dannyparsons:
fixes #3525

lilyclements:
The dialog code (so far) is going well. However, I'm having issues:
1. The `as.character` error is occurring if I don't run `library(PCICt)`
2. I'm getting a missing values error which occurs if a `climdex.something` function is run

> Error in if (!all(names(new_data) == valid_names)) { : 
>   missing value where TRUE/FALSE needed

Seems like a fairly straight forward fix (I think...) but I'm unsure where `if (!all(names(new_data) == valid_names)) {` lays.

dannyparsons:
1. I've added `library(PCICt)` to the setup in the latest version so that should be fine now.

2. Is this in the climdex functions itself? If you can't find it in our R code then it probably is. If so, then we could report it, but don't have to fix anything ourselves.

shadrackkibet:
Is this ready for review?

lilyclements:

shadrackkibet:
Reviewing this.

shadrackkibet:
Two points to note:

- The tooltip waiting time is too small, can we do something about this? when you are reading the tooltip you find that it disappears quickly even if the cursor is still on a specific checkbox. Is it possible to restrict the tooltip visibility so long as the cursor is on a specific checkbox?

- When on precipitation and check the checkbox for precipitation exceeding a specified amount per day here  i a get a bug that ;**"please specify a single numeric threshhold _value"**...I don't know much about climatic Extremes indices and probably i am doing something not sensible here!. Here is the bug i am getting:

lilyclements:
The precipitation checkbox is meant to bring up a threshold checkbox which adds a value for this. Thanks, I'll look into that now!

As for the tooltip, that's a good point. I'm not sure how to do that or how easy that is, hopefully it's quick! @dannyparsons do you know about this?

lilyclements:
@shadrackkibet I have fixed the threshold problem. This basically used to be a nud, which I changed to a ucrInput. When I made that change, I forgot to say that this value shouldn't have quotation marks around it.
I'll wait to hear Danny's thoughts on the tool tip. Otherwise, this should be good to go!

dannyparsons:
Yes tooltips are important to have working well. It seems you can here https://msdn.microsoft.com/en-us/library/system.windows.forms.tooltip.autopopdelay.aspx
But this says the default is 5000 (5s) so it shouldn't be too short. It only stays if the mouse is not moving of course. You could try setting this value but unless it's been set somewhere else it should be 5000 already.

lilyclements:
I think the tooltips are for a good length of time. It is 5 seconds, which gives more than enough time to read it. 

PR #3780, opened 2017-08-10
updated netcdf import

dannyparsons:


instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1112,21 +1112,69 @@ instat_object$set("public","make_inventory_plot", function(data_name, date_col,
 }
 )
 
-instat_object$set("public", "import_NetCDF", function(nc_data, main_data_name, loc_data_name, latitude_col_name = "", longitude_col_name = "") {
-  nc_result <- open_NetCDF(nc_data = nc_data, latitude_col_name = latitude_col_name, longitude_col_name = longitude_col_name)
-  if(length(nc_result) != 3)stop("Output from open_NetCDF should be a list of length 3")
-  
-  data_list = nc_result[c(1,2)]
+instat_object$set("public", "import_NetCDF", function(nc, name, only_data_vars = TRUE, keep_raw_time = TRUE, include_metadata = TRUE) {
```
dannyparsons: @lilyclements this is the new netcdf import function

- `nc` - a `ncdf4` object i.e. output from `ncdf4::nc_open(filename = "file.nc")`
- `name` - prefix for the data frames that will be created. Default should be the file name.
- Others - all simple checkboxes. These options will become more complicated in future so we don't need much thought on the layout yet.

The ncdf4 object also needs to be closed after use. So full example would be:
```
nc <- ncdf4::nc_open(filename = "dat.nc")
InstatDataObject$import_NetCDF(nc = nc)
ncdf4::nc_close(nc = nc)
```

PR #3720, opened 2017-07-20
Canonical Bug fix

shadrackkibet:
@lilyclements  this is ready for review.Here i have sorted out the following;

- Display of graphs on the output window.This was producing an empty grid on the output window.

- Sorted out iposition problem. The display on the output window was not in order.

Issues:

- The two checkboxes on the sdg behaves funny. When one is checked the other one checks automatically on getting out of the sub.

- The data_name parameter BUG. This is an issue with the settingRcode in this dialog. It seems adding additional parameter pairs don't work properly. I don't understand why this happens! However i have added them manually for now. Have a look at this please.

PR #3786, opened 2017-08-10
CCA fixes

lilyclements:
I've fixed the remaining data_name issue outlined in #3720
This means that #3720 can now be closed and @africanmathsinitiative/developers this is ready to review.

### DIFF
--- a/instat/dlgCanonicalCorrelationAnalysis.vb
+++ b/instat/dlgCanonicalCorrelationAnalysis.vb
@@ -19,7 +19,7 @@ Public Class dlgCanonicalCorrelationAnalysis
     Public bFirstLoad As Boolean = True
     Private bResetSubdialog As Boolean = False
     Private bReset As Boolean = True
-    Private clsDefaultFunction, clsRCanCorFunction, clsRXCoefFunction, clsRYCoefFunction, clsRGraphicsFunction As New RFunction
+    Private clsDefaultFunction, clsRCanCorFunction, clsRCoefFunction, clsRGraphicsFunction As New RFunction
 
     Private Sub dlgCanonicalCorrelationAnalysis_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
@@ -69,8 +69,7 @@ Public Class dlgCanonicalCorrelationAnalysis
         clsDefaultFunction = New RFunction
         clsRGraphicsFunction = New RFunction
         clsRCanCorFunction = New RFunction
-        clsRXCoefFunction = New RFunction
-        clsRYCoefFunction = New RFunction
+        clsRCoefFunction = New RFunction
 
         ucrSelectorCCA.Reset()
         ucrSaveResult.Reset()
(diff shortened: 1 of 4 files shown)


instat/sdgCanonicalCorrelation.vb
```diff
@@ -28,13 +28,17 @@ Public Class sdgCanonicalCorrelation
...
         ucrChkCanonicalCorrelations.SetText("Canonical Correlations")
+        ucrChkCanonicalCorrelations.SetValueIfChecked(Chr(34) & "cor" & Chr(34))
+        ucrChkCanonicalCorrelations.AddParameterPresentCondition(True, "value1")
+        ucrChkCanonicalCorrelations.AddParameterPresentCondition(False, "value1", False)
 
-        ucrChkCoefficients.AddRSyntaxContainsFunctionNamesCondition(True, {frmMain.clsRLink.strInstatDataObject & "$get_from_model"})
-        ucrChkCoefficients.AddRSyntaxContainsFunctionNamesCondition(False, {frmMain.clsRLink.strInstatDataObject & "$get_from_model"}, False)
+        ucrChkCoefficients.SetParameter(New RParameter("value1", 2))
         ucrChkCoefficients.SetText("Coefficients")
+        ucrChkCoefficients.SetValueIfChecked(Chr(34) & "xcoef" & Chr(34))
```
lilyclements: I'm having an issue here. When this one checkbox is checked it runs two functions:

`IDO$get_from_model(..., value1 = "xcoef")`
`IDO$get_from_model(..., value1 = "ycoef")`

So the `.SetValueIfChecked` therefore has two values.

At the moment, this is just running `value1="xcoef"` on both times.

@africanmathsinitiative/developers does anyone know how I can fix this?

Why would the user not specify between running their x-coefficients and their y-coefficients? (i.e., why do we not have two checkboxes for this...) - is that not common in CCA?

dannyparsons: I think since we can't easily do these together automatically, then just do one of them manually in the code and keep it only linked to one.
The checkbox is just to display the coefficients so that's why it's doing both.
Alternatively, add this small function in the stand alone functions R file and then use this:
```
concor_coef <- function(object) {
  object[c("xcoef", "ycoef")]
}
```
so that the checkbox does runs this single function now which should be much easier. And probably nicer for the code.

lilyclements: Great thanks! Fixed it.

instat/dlgCanonicalCorrelationAnalysis.vb
```diff
@@ -130,20 +121,18 @@ Public Class dlgCanonicalCorrelationAnalysis
...
         If ucrSaveResult.ucrChkSave.Checked Then
             clsRCanCorFunction.AddParameter("model_name", Chr(34) & ucrSaveResult.GetText & Chr(34))
-            clsRXCoefFunction.AddParameter("model_name", Chr(34) & ucrSaveResult.GetText & Chr(34))
-            clsRYCoefFunction.AddParameter("model_name", Chr(34) & ucrSaveResult.GetText & Chr(34))
+            clsRCoefFunction.AddParameter("object", Chr(34) & ucrSaveResult.GetText & Chr(34))
         Else
             clsRCanCorFunction.AddParameter("model_name", Chr(34) & "last_CCA" & Chr(34))
-            clsRXCoefFunction.AddParameter("model_name", Chr(34) & "last_CCA" & Chr(34))
-            clsRYCoefFunction.AddParameter("model_name", Chr(34) & "last_CCA" & Chr(34))
+            clsRCoefFunction.AddParameter("object", clsRFunctionParameter:=ucrBase.clsRsyntax.clsBaseFunction)
```
dannyparsons: This should be the base function either way, so you can do this line outside the End If and remove the line in the If.

lilyclements: Good point, fixed it now

PR #3797, opened 2017-08-14
SOR fix

lilyclements:


lilyclements:
@dannyparsons I'm not sure why this fixes my problem... But, sometimes if I run this dialog, then I

lilyclements:
@dannyparsons I'm not sure why this fixes my problem... But, sometimes when I run this dialog one of my nud values aren't read in. 
This seems to occur if I check the 4th checkbox, press OK, uncheck, press OK, re-check, press OK. When I do this, it runs `width = 45 - + 1` where it should run `width = 45 - 30 + 1` in the `dry_period_overall_interval` function. 
However, if I instead run `.GetParameter` (which means making a few other amendments) then it runs fine. Can this work as a temp. fix?

I've also sorted naming conventions while on this dlg.

dannyparsons:
It should probably work either way, but it also makes sense that they are the same parameter so I think it's fine.

PR #3805, opened 2017-08-15
Right Clicks Bug Fix

### LINKED ISSUE #3777: Sort from right click uses the wrong data frame
When I have multiple data frames I right-click and choose Sort.
It takes me to the dialogue - which is fine.  But it is not always the visible data frame, which I would assume for the right-click option.
The example I had was climatic data where I had used sort before on the date column.
I then got a subset (using the filter) and wanted to sort on a column in the new data frame.  It went back to the previous data frame - as would (I assume) be correct if I had gone directly to the Sort dialogue.

But this way should (I assume) clear the fields and assume my new data frame?
Perhaps there are other right-click options that have the same problem?

stevenndungu:
Fixes #3777


instat/dlgDuplicateColumns.vb
```diff
@@ -131,6 +134,11 @@ Public Class dlgDuplicateColumns
         bUseSelectedColumn = False
     End Sub
 
+    Public Sub SetCurrentDataframe(strDataFrame As String)
+        strSelectedDataFrame = strDataFrame
+        ucrSelectorForDuplicateColumn.SetDataframe(strSelectedDataFrame)
+    End Sub
```
dannyparsons: I don't think we need these SetCurrentDataframe methods when the data frame is specified when you set the column. So I don't think this is needed.

instat/dlgDuplicateColumns.vb
```diff
@@ -46,6 +46,9 @@ Public Class dlgDuplicateColumns
         If bReset Then
             SetDefaults()
         End If
+        If bUseSelectedColumn Then
+            SetDefaultColumn()
+        End If
```
dannyparsons: Does this work here? I think even if it does it's better that this part goes after setting the R code.

dannyparsons: And the same for the other diaogs.

stevenndungu:
I  have made the edits.

PR #3806, opened 2017-08-16
Fixing selector bug in graphics

### LINKED ISSUE #3789: Changing data frame - graphics dialogs
When you change data frame in a graphics dialog the general options don't change unless you open the sub dialog again. This is serious for facets because you will get an error when it looks for a variable that doesn't exist in your current data frame. 

I have added a field to the selector (#3788) to be able to store the ggplot operator, and set this on the boxplot dialog as an exaple, so that clearing the facet parameter can be done from there. This means one line needs to be added to all specific graphics dialogs. @maxwellfundi can you finish this task? This is urgent for next release as it's a bug.

In general, are there other options we should/want to clear when the data frame is changed e.g. title, axes options etc? This could be less urgent as I think facets is the only error causing option, but there might to others that cause problems as well, like scales so I think we should discuss this. @rdstern @volloholic

maxwellfundi:
@dannyparsons  this has now been done for all graphics dialogs.

Fixes #3789


dannyparsons:
Great - you'll need to do this on the climatic boxplots dialog as well, maybe after it's merged. And have a check if there are other graphics dialogs needing this in other places, maybe not at the moment.

PR #3818, opened 2017-08-17
Climatic Summary

shadrackkibet:

### DIFF
--- a/instat/dlgClimaticSummary.vb
+++ b/instat/dlgClimaticSummary.vb
@@ -197,6 +197,8 @@ Public Class dlgClimaticSummary
 
         ucrSelectorVariable.SetRCode(clsDefaultFunction, bReset)
         ucrReceiverElement.SetRCode(clsDefaultFunction, bReset)
+        ucrChkStoreResults.SetRCode(clsDefaultFunction, bReset)
+        ucrChkPrintOutput.SetRCode(clsDefaultFunction, bReset)
 
         ucrPnlAnnual.SetRCode(clsDefaultFactors, bReset)
         ucrReceiverStation.SetRCode(clsDefaultFactors, bReset)
--- a/instat/dlgClimaticSummary.vb
+++ b/instat/dlgClimaticSummary.vb
@@ -50,12 +50,12 @@ Public Class dlgClimaticSummary
         ucrSelectorVariable.SetParameterIsString()
 
         'panel setting
-        ucrPnlAnnual.AddRadioButton(rdoAnnual)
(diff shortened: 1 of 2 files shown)


shadrackkibet:
Should i also change the receiver focus when on "Within-Year rdo" to be the "Within-Year" Receiver?

dannyparsons:
I can't quite remember the dialog layout but generally we only need to change the current receiver if we are hiding the current receiver. If we're just changing other things about the dialog, no need to change this.
I'm not sure which case applies to this dialog so you'll have to decide that.

lilyclements:
So, when we are on the Within-Year rdo, there is a new receiver which pops up. Hence this disappears when you click on the Annual rdo. Therefore, when you click onto the Annual rdo, we need to change the current receiver, otherwise the focus is on a non-visible receiver.

shadrackkibet:
@dannyparsons @lilyclements  how can restrict the withinYear receiver to only pick the within-Year variables i.e Month,dekad, etc .Is this possible by setting the receiver's metadata property?. I noticed there this has a bug when you allow the user to select any Element for example "Rain" into this receiver.

lilyclements:
I think it's okay as it is. If the user wishes to do it by "Rain", then they can - not that this is in any way advised! However, this way we can allow things like "Day", "doy_366", "Dekad" "Week", etc - variables which are generally automatically numeric.

dannyparsons:
That's a great point. We should restrict to those "within year types", but since we don't have record this at the moment we can't quite do it. So I think better to allow anything and explain in future we will restrict.
One question - is within year restrict to numeric now? What if my month is a factor/character? Should this be allowed? If so then I suggest no restriction on this receiver at the moment.

lilyclements:
There is no restriction on it at all at the moment, which is what we want for now because then the user can pick factors or numerics
I'll write an issue up on creating a within-year type set. In which case, this should be good for @dannyparsons

instat/dlgClimaticSummary.vb
```diff
@@ -231,6 +234,12 @@ Public Class dlgClimaticSummary
         End If
     End Sub
 
+    Private Sub ReceiverFocus()
+        If rdoAnnual.Checked Then
+            ucrReceiverStation.SetMeAsReceiver()
+        End If
```
dannyparsons: This could be modified so that the current receiver only changes if the receiver that's about to be hidden is the current one. i.e. if another receiver is the current then no need to do this when changing radio buttons.
Can you look into this? but I will still merge this version now because it's a small improvement.

Issue #3834, opened 2017-08-19
Describe > View Graph


rdstern:
The default is to view in the interactive viewer.  I wanted to **Display in R Viewer**, which it did.  But it doesn't seem to keep this option.  When I returned to the dialogue it was back to **Display in Interactive Viewer**.

Second problem.  I used the Describe > One Variable Frequencies on the 6 factors in the Sudan data.  I then wanted to Describe > View Graph > Display in R Viewer.  It doesn't give an error, but it doesn't display either.

lilyclements:
To the first problem:
I have fixed part of the problem. That is, it now it does not go to Interactive Viewer when you reopen the dialog (unless you had that checked beforehand!)

However, there is another problem which was hidden by this problem.

We have three (enabled) radio buttons, one runs
* Interactive runs:    ggplotly(p = IDO$get_graphs(...) )
* R-Viewer runs:      IDO$get_graphs(...)
* Output runs:         IDO$get_graphs(...)

The difference between R-Viewer and Output window isn't in the r-code, it's an option added in VB.

If I check R-Viewer, press OK and reopen - it, unsurprisingly, doesn't know what to check because two radio buttons fulfil the conditions.
Similarly so for Output Window.

(If I disabled one of R-Viewer or Output window and ran it, then the radio buttons worked great, so I'm confident with my solution to fix Roger's problem)



To the second problem:
The object created in One Way Frequencies (graph) isn't a ggplot object so this shouldn't be visible in this dialog. It's a "gtable" "gTree"  "grob"   "gDesc" object.
A similar problem is seen with One Variable Graph if you use the "Combined Graph" option.
I'm not sure how to fix this problem in the dialog - I know you can't get a base-R graph in this dialog so there is a specification somewhere.

dannyparsons:
Yes there's no way to do the automatic detection of the radio buttons because it relies on VB options, not R code. We could just do all the code for these radio buttons manually to avoid this.

We could either detect the graph type and disable the Interactive option or have an informative error message.

lilyclements:
disable the Interactive option? Let's not get too extreme now.
We are having problems here with all the radio buttons for a "gtable" "gTree" "grob" "gDesc" object not just with the Interactive rdo.

Great, I'll fix this to be manual (I think I can still remember the old system).

dannyparsons:
I meant disable interactive when the selected graph is not a gg type.

lilyclements:
[...]

PR #3838, opened 2017-08-21
View graph fix

lilyclements:
This is at a standstill until the issues I have mentioned in #3834 can be addressed


maxwellfundi:
@dannyparsons  the issue #3834 needs to be addressed so that lily can finish up this. @lilyclements  does this work properly except for the issues in the #3834 ?

lilyclements:
Yep this works fine asides from the final issue in #3834

lilyclements:
@africanmathsinitiative/developers this is fine (in my opinion) as it is so perhaps could be reviewed now?

There is still a bug which could be done on a new pull request. Otherwise, someone can take over this branch if you prefer.

The remain bug is that this dialog should only accept `ggplot` graphs, however, it is currently accepting
`gtable` `gTree` `grob` `gDesc` graphs (from the dialogs One Way Frequencies or One Variable Graphs (if you use combine)). These graphs do not work for any of the radio buttons so we need to fix this.

maxwellfundi:
@dannyparsons  this is ready for you!

dannyparsons:
I think the remaining issue from Lily is written up somewhere but can you check @maxwellfundi and write an issue if not?

PR #3841, opened 2017-08-22
SOR bug fix

lilyclements:
@dannyparsons this is ready to merge.

Here I have removed "save = 2" for two functions which were not meant to be printing their results in the main data frame.


dannyparsons:
I think I noticed this with the dialog adding something to the daily data. I think not adding is a good default but maybe this should be an option too, since those columns could be useful.

PR #3849, opened 2017-08-23
Duplicates additional option

AlexSananka:
@africanmathsinitiative/developers  this dialog is almost where @rdstern  wants it to be.I mean it's making good progress.


AlexSananka:

AlexSananka:
I just realised this still has issues and am still working on them.

lilyclements:
@AlexSananka have you fixed the additional issues? Is this ready to review?

AlexSananka:
Yes please go ahead and review/Test this.

instat/dlgTransformClimatic.Designer.vb
```diff
@@ -130,7 +130,7 @@ Partial Class dlgTransformClimatic
         'rdoMoving
         '
         Me.rdoMoving.Appearance = System.Windows.Forms.Appearance.Button
-        Me.rdoMoving.Cursor = System.Windows.Forms.Cursors.IBeam
+        Me.rdoMoving.Cursor = System.Windows.Forms.Cursors.Default
```
lilyclements: @AlexSananka make sure you only do relevant changes per pull request.
My understanding is that this is fixing a different issue and should warrant a new pull request. This is to help in the case of issues such as conflicts.
In this instance, I have edited Transform Climatic in a pull request too, and now I probably shouldn't edit the layout on it in case of a conflict. I know you only actively changed one line, but a lot of lines have automatically moved around here.

AlexSananka: I thought that change wasn't enough to create a new pull request,I just changed one line.
I can revert this commit and you make it in the open pull request.

lilyclements: There's a lot of other lines which have changed which I know weren't intentional but that does happen when you edit something in the designer. If you look at the changes before committing (or now, if you look at the changes made to the designer) you can see that a lot of things have been moved around - even though you didn't actively do that! Those are the changes which means I shouldn't edit the designer now.
It is something you should do in a separate pull request even if it's a one line edit.
It's also good practice to do it in another pull request in case you notice a bigger problem that needs to be fixed because then it isn't one line. For example, I just opened a pull request to fix the alignment for something - a very trivial task. While fixing that alignment, I noticed other bugs that need to be fixed. By being in a new branch, I could fix them there and then.

AlexSananka: I just reverted that commit,you can go on and make the desired designer changes.

lilyclements: Okay sure, my point wasn't to get you to revert it but more to explain to not do that in the future! But thank you! :) shall I make these changes to 3848 on my pull request on transform climatic?

AlexSananka:
I will try solve point one and two.And am not sure what you mean in point three.As for the ignore case yes you're right we do not need to pass a string there so those quotes shouldn't be there,Thanks.

lilyclements:
You said:
>As for the ignore case yes you're right we do not need to pass a string there so those quotes shouldn't be there

I meant that fourth point as two points: Do we want quotation marks? Do we want to call it as a string? There's a difference between quotations and string. For example, in start of rains I call the day of year variable as a string, however, I do not have quotations around it.

In point three, look at the output code. There is often a double space between the end of one code and the start (i.e., the comment line) of the next.

AlexSananka:
for point one and two we don't need doing either and I have corrected that.
For the output code that's not happening on my side.

AlexSananka:
@lilyclements this is ready for you to re-review.

dannyparsons:
@AlexSananka if you're not getting the same issue someone else is with your code, then you need to find out why!

instat/dlgDuplicatesConstructed.vb
```diff
@@ -72,29 +111,43 @@ Public Class dlgDuplicatesConstructed
...
         clsDuplicated.AddParameter("x", clsRFunctionParameter:=ucrSelectorDuplicateswithVariables.ucrAvailableDataFrames.clsCurrDataFrame, iPosition:=0)
 
+        clsStreakFunction.AddParameter("ignore", "NULL")
         clsDuplicated2.SetPackageName("questionr")
         clsDuplicated2.SetRCommand("duplicated2")
         clsDuplicated2.AddParameter("x", clsRFunctionParameter:=ucrSelectorDuplicateswithVariables.ucrAvailableDataFrames.clsCurrDataFrame, iPosition:=0)
-
+        'clsDuplicated2.AddParameter("x", frmMain.clsRLink.strInstatDataObject & "$get_data_frame", iPosition:=0)
+        clsStreakFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$duplicated_cases")
+        clsSubsetCol.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_columns_from_data")
```
lilyclements: How come you are running this here?

lilyclements:
This is what I mean by the spacing:
image

AlexSananka:
I see what you mean't, I previously thought it's the spacing  between each line.I will look into this.

instat/dlgDuplicatesConstructed.vb
AlexSananka: we need the column from the data and not the column names so we need this so that we can pass it's script in the ignore parameter.

lilyclements: Ok, so are you doing for the other radio buttons?

AlexSananka: The other radio buttons are not passing the parameter ignore.

lilyclements: One of the other ones is passing the variable in the receiver, however, just under a different parameter name

lilyclements: How about you try running the middle radio button and look at the code output?

instat/dlgDuplicatesConstructed.vb
```diff
@@ -130,43 +121,39 @@ Public Class dlgDuplicatesConstructed
         SetDataFrameOrColumns()
         ucrNewColumnName.Reset()
         ucrSelectorDuplicateswithVariables.Reset()
+        ucrReceiverForSelectedVariables.SetMeAsReceiver()
+        ucrReceiverForSuccessiveValues.SetMeAsReceiver()
```
lilyclements: These lines can be removed because we don't need to set them as a receiver here. We're setting them as a receiver when their rdo's are checked, however, in the default rdo we don't have a receiver

instat/dlgDuplicatesConstructed.vb
```diff
@@ -130,43 +121,39 @@ Public Class dlgDuplicatesConstructed
         SetDataFrameOrColumns()
         ucrNewColumnName.Reset()
         ucrSelectorDuplicateswithVariables.Reset()
+        ucrReceiverForSelectedVariables.SetMeAsReceiver()
+        ucrReceiverForSuccessiveValues.SetMeAsReceiver()
 
-        ' For the first two radio buttons, we run either clsDuplicated or clsDuplicated2
+        ucrBase.clsRsyntax.iCallType = 2
```
lilyclements: How come you have moved this from Initialise to SetDefaults?

instat/dlgDuplicatesConstructed.vb
```diff
@@ -130,43 +121,39 @@ Public Class dlgDuplicatesConstructed
...
         ucrSelectorDuplicateswithVariables.Reset()
+        ucrReceiverForSelectedVariables.SetMeAsReceiver()
+        ucrReceiverForSuccessiveValues.SetMeAsReceiver()
 
-        ' For the first two radio buttons, we run either clsDuplicated or clsDuplicated2
+        ucrBase.clsRsyntax.iCallType = 2
         clsDuplicated.SetRCommand("duplicated")
         clsDuplicated.AddParameter("x", clsRFunctionParameter:=ucrSelectorDuplicateswithVariables.ucrAvailableDataFrames.clsCurrDataFrame, iPosition:=0)
 
+        clsStreakFunction.AddParameter("ignore", "NULL")
```
lilyclements: This needs a parameter position here and elsewhere

instat/dlgDuplicatesConstructed.vb
```diff
@@ -189,11 +176,17 @@ Public Class dlgDuplicatesConstructed
 
     Private Sub SetIgnoreVals()
         If ucrChkOmitValues.Checked Then
-            clsSubsetCol.SetAssignTo(ucrReceiverForSuccessiveValues.GetVariableNames(False))
```
lilyclements: How come you removed this set assign to? I think it is needed for both consistency and efficiency, but also because it makes the R-Code less clunky and easier to understand

instat/dlgDuplicatesConstructed.vb
```diff
@@ -189,11 +176,17 @@ Public Class dlgDuplicatesConstructed
 
     Private Sub SetIgnoreVals()
         If ucrChkOmitValues.Checked Then
-            clsSubsetCol.SetAssignTo(ucrReceiverForSuccessiveValues.GetVariableNames(False))
-            ucrBase.clsRsyntax.AddToBeforeCodes(clsSubsetCol, iPosition:=1)
-            clsStreakFunction.AddParameter("ignore", ucrReceiverForSuccessiveValues.GetVariableNames(False) & "[" & ucrReceiverForSuccessiveValues.GetVariableNames(False) & ucrInputConditions.GetText & ucrNudOmit.GetText & "]", iPosition:=2)
+            If Not ucrReceiverForSuccessiveValues.IsEmpty Then
```
lilyclements: I don't understand why you are running this here.
My understanding is that the receiver being empty or not doesn't decide whether ignore = "NULL" or ignore = Var[Var==0]. How come you've added this in?

instat/dlgDuplicatesConstructed.vb
```diff
@@ -204,63 +197,53 @@ Public Class dlgDuplicatesConstructed
...
 
+    Private Sub ucrPnlOptions_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrPnlOptions.ControlValueChanged
+        SetDataFrameOrColumns()
+    End Sub
+
     Private Sub SetDataFrameOrColumns()
         If rdoDataFrame.Checked Then
             ucrSelectorDuplicateswithVariables.SetVariablesVisible(False)
-            ' note that we have to run this here because the parameter x is used for both functions and all four radio buttons
+            ucrSelectorDuplicateswithVariables.SetParameterIsrfunction()
```
lilyclements: I am really sceptical about this affecting the set up. But if this is a'okay then great

instat/dlgDuplicatesConstructed.vb
```diff
@@ -204,63 +197,53 @@ Public Class dlgDuplicatesConstructed
...
     End Sub
 
+    Private Sub ucrPnlOptions_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrPnlOptions.ControlValueChanged
+        SetDataFrameOrColumns()
+    End Sub
+
     Private Sub SetDataFrameOrColumns()
         If rdoDataFrame.Checked Then
             ucrSelectorDuplicateswithVariables.SetVariablesVisible(False)
-            ' note that we have to run this here because the parameter x is used for both functions and all four radio buttons
```
lilyclements: How come you removed this comment? I think we need an explanation here as to why it's being done manually here.

When I worked on this dialog a few weeks ago, I understood why we were running this manually and I didn't add a comment in. A quick comment would have saved myself a lot of time:

Then when I came back to this dialog a couple of days ago, there was a bug with the `x` parameters. I assumed this bug must be coming from the addition of the new radio button.
I did not realise that someone had removed this bit of code as they didn't realise it was necessary. Had there been a comment, it probably would not have been removed.

However, I also forgot that it is meant to be in the code here because it was weeks ago that I last looked at this dialog. So I was searching elsewhere for the bug! It wasn't until I looked at a branch from before this new rdo was added in that I realised what was wrong. Had there been a comment there, I would have remembered why we needed this in and why this would fix that bug!

When I added the code back in, I added a comment stating why we need this code to prevent this issue in a few more weeks time.

Comments can be really helpful to understand the code for both your future self (There's 200-300 dialogs. After working a little bit on lots of them, it's easy to forget why you did something one way a few weeks or months ago! Your future self can be really thankful if you a quick comment now!) or for someone else looking at it (open source software does mean this is likely! Or it can help another developer looking at it to understand why its needed - either in a review or if they work on it at a later date)

instat/dlgDuplicatesConstructed.vb
```diff
@@ -72,29 +112,44 @@ Public Class dlgDuplicatesConstructed
...
         clsDuplicated2 = New RFunction
+        clsStreakFunction = New RFunction
+        clsSubsetCol = New RFunction
 
+        SetDataFrameOrColumns()
         ucrNewColumnName.Reset()
         ucrSelectorDuplicateswithVariables.Reset()
-        ucrReceiverMultipleForDuplicates.SetMeAsReceiver()
+        ucrReceiverForSelectedVariables.SetMeAsReceiver()
+        ucrReceiverForSuccessiveValues.SetMeAsReceiver()
```
lilyclements: These lines can be removed because we don't need to set them as a receiver here. We're setting them as a receiver when their rdo's are checked, however, in the default rdo we don't have a receiver

instat/dlgDuplicatesConstructed.vb
```diff
@@ -72,29 +112,44 @@ Public Class dlgDuplicatesConstructed
...
 
+        ucrBase.clsRsyntax.iCallType = 2
```
lilyclements: How come you have moved this from Initialise to SetDefaults?

instat/dlgDuplicatesConstructed.vb
```diff
@@ -114,6 +169,23 @@ Public Class dlgDuplicatesConstructed
         End If
     End Sub
 
+    Private Sub SetIgnoreVals()
+        If ucrChkOmitValues.Checked Then
+            If Not ucrReceiverForSuccessiveValues.IsEmpty Then
```
lilyclements: I don't understand why you are running this here.
My understanding is that the receiver being empty or not doesn't decide whether ignore = "NULL" or ignore = Var[Var==0]. How come you've added this in?

instat/dlgDuplicatesConstructed.vb
lilyclements: This line here is what I'm referring to! But there are other comments

instat/dlgDuplicatesConstructed.vb
```diff
@@ -14,6 +14,7 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports RDotNet
```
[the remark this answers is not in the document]
AlexSananka: Do I'll delete this.I don't think it's necessary.

instat/dlgDuplicatesConstructed.vb
AlexSananka: good spot

instat/dlgDuplicatesConstructed.vb
AlexSananka: These are being repeated I need to remove them here

instat/dlgDuplicatesConstructed.vb
AlexSananka: This is also wrongly placed.

lilyclements: Is this the correct call type by the way?

instat/dlgDuplicatesConstructed.vb
AlexSananka: The ignore parameter takes whatever that is in the receiver so when you check the checkbox when the receiver is empty it complains because there are no col_names to add.So basically we always need something in the receiver so that you can add the parameter ignore.

AlexSananka:
Generally the code might look messier but I was still concentrating on making it work rather than arranging and commenting.I am going to add in the comments when am done with everything anyway.

instat/dlgDuplicatesConstructed.vb
AlexSananka: Not yet sure In fact am still not sure why we're having this triple spacing between this lines of output.

instat/dlgDuplicatesConstructed.vb
lilyclements: Ah okay I see! Thanks

instat/dlgDuplicatesConstructed.vb
lilyclements: The issue is coming because of an incorrect call type

lilyclements: I just tried iCallType = 0 and that solves the problem.

AlexSananka: Aha! great,Thanks.I didn't try 0 before.

AlexSananka:
@lilyclements I have sorted out the above issues,you could have a final review on this please.

lilyclements:
How come you removed the assignment for the new radio button for the IDO$get_columns_from_data?

AlexSananka:
I don't think it's necessary for now,as you've seen before it messes up our code because we need to be adding and removing during several and different events and has to change lot's of things.If you can make it work perfectly it would be really good.

AlexSananka:
@lilyclements Is this ready for @dannyparsons to have a look?

lilyclements:
I want to discuss this dialog with Danny before it goes any further. I'm not sure on the justification of removing the assign to because if it, as you said, "messes up our code" then there's either a problem in the core code which we'll have to address, or there's a problem in how we're doing it. Either way, a discussion is needed.

AlexSananka:
@dannyparsons did you have a discussion with @lilyclements on this? I think this needs to get done.

dannyparsons:
Yes we should get it done. Is this complete from your point of view? I thought we discussed that the new duplicates method should not be in the data object and could be a stand alone function?

AlexSananka:
I am going to move this method to the stand alone function.

AlexSananka:

AlexSananka:
Finally done with this. @dannyparsons you can have a look at this.

dannyparsons:
This is looking good, I have made a pull request to format the R code correctly. Check to see how it now follows our convention.
A few small changes you could make:
1. The omit values could be a textbox instead of a nud. I can only do integer values at the moment.
2. The label for the receivers should be "Variables to duplicate" because it's not duplicating them, it's checking them for duplicates, so it could be something like "Variables to check" instead.

AlexSananka:
@dannyparsons suggested changes have been made.

instat/dlgName.vb
```diff
@@ -24,27 +24,21 @@ Public Class dlgName
...
         bReset = False
-        autoTranslate(Me)
         If bUseSelectedColumn Then
             SetDefaultColumn()
         End If
     End Sub
 
-    Private Sub ReopenDialog()
-        ucrSelectVariables.Reset()
-    End Sub
```
dannyparsons: I should have noticed this before I merged but why did you remove these parts?

instat/dlgDuplicatesConstructed.vb
```diff
@@ -114,6 +170,24 @@ Public Class dlgDuplicatesConstructed
         End If
     End Sub
 
+    Private Sub SetIgnoreVals()
+        If ucrChkOmitValues.Checked Then
+
+            If Not ucrReceiverForSuccessiveValues.IsEmpty Then
+                clsSubsetCol.AddParameter("col_name", ucrReceiverForSuccessiveValues.GetVariableNames())
+                clsSubsetCol.AddParameter("data_name", Chr(34) & ucrSelectorDuplicateswithVariables.ucrAvailableDataFrames.cboAvailableDataFrames.SelectedItem & Chr(34))
+                frmMain.clsRLink.RunInternalScriptGetValue(clsSubsetCol.ToScript)
```
dannyparsons: Why are you running the script here?

instat/dlgName.vb
AlexSananka: I was testing something and forgot to put this back.I think I can add this on a new pull request.

instat/dlgDuplicatesConstructed.vb
AlexSananka: This shouldn't be here as well,making this edits on now.

dannyparsons: yes please

PR #3854, opened 2017-08-24
Summaries Extension.

shadrackkibet:
In progress


instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -440,6 +442,16 @@ summary_median <- function(x, na.rm = FALSE,...) {
...
 
+# upper quartile function
+upper_quartile<-function(x,na.rm=TRUE, probs=0.75,...){
+  return(quantile(x,na.rm = na.rm, probs = probs ))
+}
+
+# lower quartile function
+lower_quartile<-function(x,na.rm=TRUE, probs=0.25,...){
+  return(quantile(x,na.rm = na.rm, probs = probs ))
+}
```
dannyparsons: This is looking good. Remember to use correct spacing in the R code. Also I suggest we have one `quartile` function which takes `probs` as an argument. And then we can have an `upper` and `lower` function as well. But for `upper` and `lower`, `probs` shouldn't be an argument because if you change `probs` then it might not be the upper quartile anymore! We can look at this together if we need to.

shadrackkibet: sure , i am happy to look at this with you tomorrow.

shadrackkibet:

lilyclements:
Reviewing this now

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -440,6 +442,21 @@ summary_median <- function(x, na.rm = FALSE,...) {
...
+
+# lower quartile function
+lower_quartile <- function(x, na.rm = FALSE, ...) {
+  return(quantile(x, na.rm = na.rm, probs = 0.25 ))
+}
+
+# upper quartile function
+upper_quartile <- function(x, na.rm = FALSE, ...) {
+  return(quantile(x, na.rm = na.rm, probs = 0.75 ))
+}
```
lilyclements: If I run Lower or Upper Quartiles and do not select to omit missing values then there is an error. (At least, this occurred for Dodoma: Rain by Year)

shadrackkibet: sure, This is an issue with missing values when calculating quartiles. @dannyparsons  what are your suggestions about this?

dannyparsons: Strange that this is an issue, but this is why we have our own wrappers! So what we can do is modify our `summary_quantile` function to behave the same as other summary functions when there's missing values that we're not removing.
Can you have a think about how to modify the function to achieve this? Then we'll discuss together.

lilyclements:
I know this is in progress, but I had a little look and have a request. I noticed you have on the proportions sub dialog a checkbox at the top for "Calculate Proportions or Percentages".
Can I request that this instead moves to the main dialog, and if checked, the button for Proportions/Percentages is enabled - otherwise, the button is disabled? This would mean it works similar to "One Var > Describe" with our customise checkbox.

dannyparsons:
Interesting suggestion. I'm not sure I agree because I don't really like how this works on one variable describe at the moment. I was seeing this as more like how facets works on the sub dialog. On the other hand the check box for proportions could do a default proportions of by all the factors which is probably most common, without the user needing to see all the options on the sub dialog.

lilyclements:
I'm not keen on one variable describe either and I know we've had discussion of it in the past when Mike wrote an issue on it. Since that dialog is hopefully going to change perhaps we shouldn't use its layout to base this layout on!
However, I do like your suggestion of the default proportions and then this sub dialog is for more option or to customise the options a little more.

Out of interest, why does proportions have it's own sub dialog but other options (Scale, Skewness, Kurtosis, etc) go on a different tab?

Finally! I just noticed a Missing Options tab, what plans to go here?
May I plug the `naniar` package because that has some nice missing options! In terms of what's relevant for that package and what seems to be relevant for this dialog is it deals with: Are there missing values? Number of missing values and Proportion of missing values. (It has a lot more than that! However, while enthusiastic about the package I do know that not all of it needs to be in a tab on a sub dialog!)

dannyparsons:
The idea of separating proportions from summaries is to show users that proportions can be applied to any summary,, they are not a summary. Also a practical reason that the options became quite complex and need space.

The missing values tab is about allowing flexibility in choosing when the summary value should be missing dependening on the number or proportion of missing values. Unlike base R where the only option is to return missing if any value is missing. So this will require modifying our summary functions. I'm not sure if naniar has functions to help with this, but if so then great to know about it.

rdstern:
Danny says: the idea is to "show users that proportions can be applied to any summary,, they are not a summary".
I'd like more explanation on this topic reported, particularly the idea that proportions "are not a summary".  That would help me, and perhaps the others in the team to understand the general concept more.
By the way I have no problem with the solution, which is to consider the proportions and percentiles separately.  Either another tab, or another sub-dialogue.  I just want to understand the thinking more.

I think we are agreed that the median IS a summary?  That's also an example of a percentile - namely the 50th percentile.  So that is finding the value, so that 50% of the variable has smaller values and 50% has larger.  So, in a common use of proportions they are the inverse of percentiles.  Instead of fixing a percentage and asking for the value (50% gives the median), we fix a value and ask for the percent (e.g. what percent of temperatures are less than 30 degrees).  So, I find it curious that when a percentile (e.g. median) is a summary, then Danny states that the inverse is not a summary.

Let me express this a different way.  We have a dialogue called **Describe > Specific > Cumulative Distribution**.   If I use this on the yields in our survey, then I get the graph below:

image

Now if I start on the y-axis at 0.5 I find the corresponding point on the x-axis is just over 40.  That's the median.  If I start on the x-axis, say 50, then I find the corresponding point on the x-axis is just over 0.75 - so about 77%.  

So please explain a bit more **why the 77% is not a summary**.

Thanks

shadrackkibet:
@africanmathsinitiative/developers this is ready for re-review.

lilyclements:
Reviewing this now

lilyclements:

dannyparsons:
I think I'm not saying that the quantiles calculation including median etc. are not summaries, but I was more talking about the separation of choosing a summary and then also for example, displaying that summary within a group as a percentage of that summary for the whole data.
We initially thought about having proportions/percentage of counts as a summary on the summary sub dialog but then decided that these are two separate things 1. choosing counts as the summary and 2. choosing to display the counts as a proportion. And also our summary system works in this way.
Then by separating these two things out we can more easily explain and do percentages of other summaries, like mean, which I think often people don't consider. So median and quantiles, and it's inverse, are summaries and you can separately calculate the proportion of this value over some other value through the proportions/percentages sub dialog.
So I think we are agreeing but I was only thinking about a specific kind of proportion calculation for this sub dialog. Maybe this needs a different name so there's not confusion between summaries that actually calculate a percentage or proportion and calculating a proportion of that summary?

shadrackkibet:

dannyparsons:
thanks for the reminder

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -440,6 +449,66 @@ summary_median <- function(x, na.rm = FALSE,...) {
...
+    }
+}
+
+# Sn function
+summary_Sn <- function(x, constant = 1.1926, finite.corr = missing(constant), na.rm = FALSE,...){
+  if(anyNA(x) && !na.rm){
+    return(NA)}else{
+      x <- x[!is.na(x)]
+      return(robustbase::Qn(x, constant = constant, finite.corr = finite.corr))
+  }
```
dannyparsons: Well done @africanmathsinitiative/developers for solving this issue!

PR #3869, opened 2017-08-29
dlgBarPieChart

lilyclements:
@africanmathsinitiative/developers this is ready to review. Edits here:
* Alignment of rdos
* Receiver focus if you change rdos
* Removed duplicate subs
* Fixed the size of the ucrSave
* Sub arrangement on the dialog code
* Naming conventions


shadrackkibet:
Reviewing this now.

instat/dlgBarAndPieChart.vb
```diff
@@ -242,11 +242,22 @@ Public Class dlgBarAndPieChart
...
+        ReceiverFocus()
     End Sub
 
-    Private Sub CoreControls_ContentsChanged() Handles ucrFactorReceiver.ControlContentsChanged, ucrSaveBar.ControlContentsChanged
+    Private Sub CoreControls_ContentsChanged() Handles ucrReceiverFactor.ControlContentsChanged, ucrSaveBar.ControlContentsChanged
         TestOkEnabled()
     End Sub
+
+    Private Sub CoreControls_ContentsChanged(ucrChangedControl As ucrCore) Handles ucrSaveBar.ControlContentsChanged, ucrReceiverFactor.ControlContentsChanged
+
```
shadrackkibet: This is an empty sub it should be removed.

shadrackkibet:
- In the options subdialog is it easy to fix the defaults of fill and color?

- when on ucrsave and check the checkbox then select "last_graph" on the dropdown then close and reopen the dialog it unchecks.

instat/dlgBarAndPieChart.vb
lilyclements: Thanks. This sub was on there twice initially running the same thing both times. Removed it now.

lilyclements:
That's an interesting one about ucrSave. I think that will be an issue in every dialog but hadn't noticed that before! Nice. Perhaps this needs an issue written up on it or do we not think this is a big deal? If the user wishes to call it last_graph then it will be called as such regardless of it's checked. I can see why we would still want it checked. This will be an issue in the ucrSave not in the dialog.

Options sub dialog - presumably you're referring to the Layer Options one on the second tab? I'll look into this quickly. As a fun fact, I did just crash it by choosing the colours to be NULL. I'll write an issue up on this as that's a problem in the sdgLayerOptions?

lilyclements:
So, into the problem on the colour/fill aesthetics: They don't have a default colour - it's either that the parameter runs `colour = A COLOUR` or it doesn't run the parameter. However, `NULL` will crash either `colour` or `fill` on `geom_bar` so we definitely shouldn't have this option of `NULL`!
However, we do want the option of `colour = NULL` on other geoms, for example `geom_rug`, so I don't want to remove the option altogether. I'm not sure how to remove the option for just one geom.

The problem of an empty combo box is a problem on other geoms, not just `geom_bar`. So for example, `geom_rug` gives an empty colour combo box.

I've had a quick look on sdgPlotLayers and ucrGeom and elsewhere to see fixes for these but unsuccessfully.
I haven't worked with geoms before so I'm going to tag @dannyparsons into this to see his opinion.

rdstern:
I reply on the last_graph point.  I like the idea that our previous names are there for us to choose and perhaps overwrite.  But I suggest **we don't include last_graph in that list**.

This wouldn't stop you typing **last_graph** yourself, but then you are asking for a problem.  I am happy that when the last graph is called last_graph it is assumed that this was the R-Instat default, and then the save graph is unchecked.

dannyparsons:
Yes we could exclude `last_graph` from the list. Since it can't tell the difference between `last_graph` typed and `last_graph` from not saving then it will uncheck. It would be good if we could detect this but I don't think it's a big deal since it's might take a while to fix.

instat/dlgBarAndPieChart.vb
```diff
@@ -250,7 +224,29 @@ Public Class dlgBarAndPieChart
         TestOkEnabled()
     End Sub
 
-    Private Sub AllControlsChanged() Handles ucrSaveBar.ControlContentsChanged, ucrFactorReceiver.ControlContentsChanged
+    Private Sub SetDialogOptions()
```
shadrackkibet: should we do this in the event  to reduce the code? not unless you are calling this subroutine elsewhere.

lilyclements: I think this is okay as a sub?, although I will move the contents of ReceiverFocus() into here.

lilyclements: I think sometimes having things as subs makes it an easier read. But additionally, it has the sub header of "SetDialogOptions" which makes it easy to know what this sub is for.

lilyclements:
@africanmathsinitiative/developers is there anything else I need to fix here or is it ready for @dannyparsons

shadrackkibet:
This is ready for @dannyparsons to review.

Issue #3877, opened 2017-08-30
Refresh preview button in Open from File


dannyparsons:
I suggest we add a "Refresh preview" button in the Open From File dialog. We now don't refresh the previews when the dialog is reopened so this button can be used to update it. It's also possible the file is edited in another program while also in the dialog so refreshing could be useful then, or if the preview failed for some reason.

This just needs a simple button somewhere sensible on the dialog to run the following lines:
```
            RefreshFilePreview()
            RefreshFrameView()
```

PR #3878, opened 2017-08-30
Import option

lilyclements:
This is ready for review @africanmathsinitiative/developers. See #3877 for details


dannyparsons:
great, I wasn't thinking it would be that easy to enable/disable it at the right times so I didn't mention that but good that you've got this implemented too.

PR #3890, opened 2017-09-05
DlgTransformText

shadrackkibet:


instat/dlgTransformText.vb
```diff
@@ -206,7 +206,11 @@ Public Class dlgTransformText
         clsWordsFunction.SetRCommand("word")
         clsWordsFunction.AddParameter("start", 1)
         clsWordsFunction.AddParameter("end", 2)
-        clsWordsFunction.AddParameter("sep", "fixed(" & Chr(34) & " " & Chr(34) & ")")
+
+        clsFixedFunction.SetPackageName("stringr")
+        clsFixedFunction.SetRCommand("fixed")
+        clsFixedFunction.AddParameter("pattern", strParameterValue:=Chr(34) & " " & Chr(34))
+        clsWordsFunction.AddParameter("sep", clsRFunctionParameter:=clsFixedFunction)
```
lilyclements: I know this isn't up for review, but I took a quick glance at it anyway. There's a little issue here because we don't always want to call `fixed` before our item in the receiver. Perhaps I as a user would want to separate by the first character (i.e., `sep = "."`) rather than by the first full stop (i.e., `sep = fixed(".")`).
Perhaps I'm misunderstanding, but it's looking as if you're setting that it will always read the `fixed` in?

shadrackkibet: Thanks for bringing up this difference here , i didn't notice option for separating with the first character.

shadrackkibet:

lilyclements:

instat/dlgTransformText.vb
```diff
@@ -145,24 +147,26 @@ Public Class dlgTransformText
...
-        dctInputSeparator.Add("Space ( )", "fixed(" & Chr(34) & " " & Chr(34) & ")")
+        dctInputSeparator.Add("Space ( )", "stringr::fixed(" & Chr(34) & " " & Chr(34) & ")")
         dctInputSeparator.Add("Colon :", Chr(34) & ":" & Chr(34))
-        dctInputSeparator.Add("Period .", "fixed(" & Chr(34) & "." & Chr(34) & ")")
+        dctInputSeparator.Add("Period .", "stringr::fixed(" & Chr(34) & "." & Chr(34) & ")")
         dctInputSeparator.Add("Underscore _", Chr(34) & "_" & Chr(34))
         ucrInputSeparator.SetItems(dctInputSeparator)
         ucrInputSeparator.SetLinkedDisplayControl(lblSeparator)
-        ucrInputSeparator.SetRDefault("fixed(" & Chr(34) & " " & Chr(34) & ")")
+        ucrInputSeparator.SetRDefault("stringr::fixed(" & Chr(34) & " " & Chr(34) & ")")
```
dannyparsons: @shadrackkibet you were right to think that this wasn't a good way to implement this and a better way would be to use RFunctions, however there's also complications with that because of some options not having the fixed function which I think you noticed. So I think this is a reasonable compromise, but likely we'll have to return to this at some point.

PR #3892, opened 2017-09-06
Exact solutions

lilyclements:
In progress

### DIFF
--- a/instat/dlgOneVarFitModel.vb
+++ b/instat/dlgOneVarFitModel.vb
@@ -108,7 +108,6 @@ Public Class dlgOneVarFitModel
 
         ' Different default states for each of these radio buttons. enorm also doesn't connect to this option.
         ucrPnlNormal.AddToLinkedControls(ucrNudHyp, {rdoMeanNormal, rdoVarNormal}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
-        'ucrPnlStats.AddToLinkedControls(ucrNudHyp, {rdoVarNormal}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=1)
 
         ' Non parametric tests
         ucrPnlWilcoxVarTest.AddRadioButton(rdoWilcoxSignTest)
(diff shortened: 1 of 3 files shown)


lilyclements:
@africanmathsinitiative/developers This needs to be taken over. I have started to implement the "Exact Solutions" options on the One Variable Fit Model dialog here.

**For Exact Case: Normal** 
 I think when we run `t.test`, we actually want to use the `mosaic` package, not the `stats` one. The `mosaic` one is a wrapper. This should be checked up on.

**For Exact Case: Bernouli** 
We have a checkbox where the user can "Modify Options for Success". If the user does not press this, we expect the code to be
`binom.test(x = Variable, p = 0.5, conf.level = 0.95)` by default
If the user does check this and we have a numeric, then some options come up: The user can pick an operator ("==" by default) and they can type in an amount (e.g. 1) this means that the `x` parameter which now runs is changed to 
`x = (Variable == 1)`
If the user does check this and we have a factor, then a combo box comes up. Here the user can decide which factor level means it's a success. Here, we have a TODO still because we want the options in the combo box to be the levels of the factor in the receiver. The `x` parameter which now runs is changed to, for example, 
`x = (Village. == "SABEY")`
This means, to my knowledge, this will have to be run manually since it's the property of the variable which determines which code we run if the checkbox is checked.


**Notes:**
* I have set up the hypothesis tests for Poisson, No Distribution and Normal distributions. I have done part of Bernouli distribution. I have not set up the "Exact" rdo.
* I have rearranged the subs to try to make a little more sense. We have the subs for the "General" rdo option, then subs for both rdo options and then finally subs for the "Exact" rdo options. Hopefully this will make reading the code a little easier!
* There are bugs in this since I only just started on updating the "Exact" rdo. Hopefully nothing too difficult to manage. Bugs include: Haven't set up the rdo for "Exact"
* It needs to be discussed if the plotting options should be available on both rdo's or just on "General". When you run this on the "Exact" option at the moment, you may get errors but this is most definitely to do with the sub dialogs running. If you scroll up, you'll see your hypothesis test has (hopefully!) worked.

shadrackkibet:
@rdstern this has stayed here for long. What are your thoughts on this? we can take over and try to make this option work.

rdstern:
Yes, this is one that I would like us to do.  First this and then the button on the Two Variables.  We need a new label for the button.  Then this looks as though we may be able to start with Lily's code.  There is a lot of good sense above.  
It is (like the 2 variable) an example to start (possibly) with the existing code, but I would then like to build on the information above, before the work starts.  Then I assume this would be a good task for @wyclife with your support?

shadrackkibet:
@rdstern, since @Wycklife is now working on improving this dialog here #5861. Above ideas pointed out by @lilyclements are also put into consideration in the issue #5810, with enhancements on the dialog design. I, therefore, suggest that this can be closed.

lloyddewit:
This PR is superceded by PR #5861

PR #3901, opened 2017-09-08
View and remove links fix

AlexSananka:
Here the panel wasn't resetting and noticed we didn't have a click reset a sub.


AlexSananka:
But I think we already have that feature in this dialog.

lilyclements:
So we do! I only used it with one link so I'm not sure why it didn't work for me. Perhaps because I opened the dialog, closed it, made a link, reopened the dialog and it doesn't re-run `initialise` (ergo `autofill = TRUE`) on reopen. Apologies!

PR #3902, opened 2017-09-10
added subsetting options for net cdf files

dannyparsons:


dannyparsons:
merging this so that @rdstern can test but can still be tested by @africanmathsinitiative/developers 

Also, install `cmsaf` R package to use this dialog. It will be added in the next release.

PR #3938, opened 2017-09-18
Display daily data dialog

shadrackkibet:
In progress.


instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1463,3 +1463,23 @@ instat_object$set("public", "set_links", function(new_links) {
...
+                  } else {
+                      return(ggplot(data_name, aes(x=doy, y=climatic_element)) + geom_bar(stat="identity", fill="blue") + geom_rug(data=subset(data_name, is.na(climatic_element)==1), mapping = aes(x=doy), sides="b", color="red") + theme_minimal() + coord_cartesian(ylim=c(0,100)) + scale_x_continuous(breaks=c(1,32, 61, 92, 122, 153, 183, 214, 245, 275, 306, 336, 367), labels = c(month.abb,""), limits =c(0,367)) + facet_wrap(~year,ncol=2) + ggtitle(paste0(station_name, " Daily ", climatic_elementfall)) + theme(panel.grid.minor = element_blank(), plot.title = element_text(hjust = 0.5, size=20), axis.title = element_text(size=16)) + xlab("Day of the year") + ylab(climatic_element))
+                  }
+            }
+        }
+    }
+}
+)
```
dannyparsons: I think this function can be removed. The `display_daily_graph` is the correct one now.

shadrackkibet:
@africanmathsinitiative/developers this is ready for initial testing.

dannyparsons:
What are the remaining issues with the tables part?

shadrackkibet:
just only the page_by parameter issue... That is what i have noticed so far.If this is sorted out i think the table part will be now ok.

shadrackkibet:
I haven't been able to sort out the station issue on graphs as well ,, could you help me look into this?

shadrackkibet:
When you use data with station you get this bug;;
image

I think this is because we don't have an option to deal with missing values in the method.

dannyparsons:
ok yes, that's not actually to do with the station. You're right it's about missing values in the element column which isn't dealt with.

dannyparsons:
I'm going to fix a few of these issues now.

rdstern:
I have tried running this dialogue.  I am just using the graph .  It is looking very nice.
I list a few issues, then a bug.
1. There is no option to save the graph.  Could that control be added?  I was keen to look at the figure with our View Graph dialogue.  I assume that is possible?
2. The same dialogue should be both in Climatic > Check Data as well as in Climatic > Prepare.
3. I tried with a single year and would then like to edit the graph options.  Could we simply add the Options button?
4. The limits are set for rainfall.  I like to start with the 0 to 100 limits.  Could there be a lower limit as well, with default 0.  Upper limit could be 100 though it sometimes seems to say 0.  (Note for some places we will have temperatures below zero.)
5. Could it change to being a line plot for temperatures?
6. I tried plotting 2 variables and it gives an error.  It should work for 2 (or more) with the line plots for temperature and bars for rainfall.
7. I just note for the future that earleir I used it with rainfall and then temperature on the full data and then did a filter to the first 10 years and it crashed me out.  But the filtering is working well now.

Issue #3864, opened 2017-08-28
Docking on main windows


maxwellfundi:
We need to have all the controls with main windows docked because some machines machines with different displays, display the controls differently when they are not docked.

dannyparsons:
I think I can see which ones are not docked correctly from my machine because they display differently in design view. So I can look into this first.

rdstern:
How about the message that the command is taking a long time?  When I use 2 screens with R-Instat on the second, then this is displayed on the other screen.

dannyparsons:
Separate issue, but @maxwellfundi could look into this, I assume it's a start position option.

dannyparsons:
@maxwellfundi have you looked into this?

maxwellfundi:
This is just about the start position and is quite easy to fix. Will have this and the icon done

maxwellfundi:
A solution to @rdstern issue is to make sure the setuploading dialog start position would be centered but then there is a potential problem when this loads for the first time and we need to have the the dialog boxes from R visible. @dannyparsons  is there a way we could bring those dialog boxes from R to the top of everything else?

dannyparsons:
I don't think we have any control over how the RLink displays pop up boxes. We could try changing the Waiting dialog so that it isn't set to Top Most? What are other options?

maxwellfundi:
About this, I think we can have it start in a manually in a position but then that would mean @rdstern  problem is not solved. The other options are center parent which would make it appear on the centre of the data view window as i have noticed. the other is to have it at defaultwindows position which is not great. 

@dannyparsons  my suggestion is get to start in the centerscreen then make a way of making it not the topmost window.

maxwellfundi:
@dannyparsons  something interesting that i think is great is that, we can set the frmsetuploading to center screen but have it topmost to false. All our dialogs have topmost property to false, so the frmsetuploading appears on top of the dialog which would solve rogers issue. Now i am not sure if the dialogs from R when we install R-Instat for the first time will appear topmost since our frmsetuploading  will be false.

dannyparsons:
That sounds like a good idea. But we need to see it with the R windows first as you say, because that is the most important issue.

dannyparsons:
You can test my moving your packages folder to another location and then running R-Instat so that it tries to create it again.

dannyparsons:
[...]

PR #3952, opened 2017-09-21
added table containers to main windows (to have better positioning on screens that zoom) and corrected docking on receiver controls

dannyparsons:
@africanmathsinitiative/developers if I have done this well then you shouldn't see any change (almost). I have changed how controls in the main form are positioned. This links to #3864 where controls do not resize correctly on all screens. I can replicate this on my machine so I have been tested.
So hopefully this shouldn't affect the way controls display on normal screens like yours. If you do get any strange designer changes let me know. A small visible change is that there is now a small margin between controls within the main windows. It's hardly noticeable, I think it looks ok, but if it's preferred the other way without margins I can change that.

### DIFF
--- a/instat/ucrColumnMetadata.Designer.vb
+++ b/instat/ucrColumnMetadata.Designer.vb
@@ -65,27 +65,27 @@ Partial Class ucrColumnMetadata
         Me.columnFilterToolStripMenuItem = New System.Windows.Forms.ToolStripMenuItem()
         Me.clearColumnFilterToolStripMenuItem = New System.Windows.Forms.ToolStripMenuItem()
         Me.lblHeader = New System.Windows.Forms.Label()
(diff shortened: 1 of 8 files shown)


shadrackkibet:
I notice a blue margin on the output window.Is this what you mean here?

dannyparsons:
Yes, that's one difference. There's now a thin margin between controls within each window. So there is now a small gap between the blue label at the top of each window and the controls below it, and a gap around the edge of the window, which is why you can now see the blue outline of the output text box. This is because I've put everything in a table container so it has default margins. If we want to go back to how it looked before I think we could set the margins to 0.

shadrackkibet:
I notice a strange behavior, when you open any dialog the colour of the margin in the output window changes. Do you notice this?...

dannyparsons:
I'm not getting this, is anyone else? I get the blue margin when I hover over the output window with the mouse. But nothing changes when I just open a dialog.

PR #3958, opened 2017-09-22
Sizing, docking and dialog format corrections

dannyparsons:
@africanmathsinitiative/developers I discovered a few dialogs with FormBorderStyle set to FixedDialog instead of FixedToolWindow. I could tell because they have a slightly different way of displaying the X in the top right (red box for FixedToolWindow and no background for FixedDialog).
I think these are all of them because I went through all the dialogs, but if you notice any looking different then this could be the fix.


dannyparsons:
Also some of the top radio buttons had text not centred in the middle. It's not easy to see on most screens. Remember to check that when creating dialogs.

Issue #3845, opened 2017-08-22
Column Statistics extended


rdstern:
This dialogue is working well.
The first extension is to add proportions and percentiles.  This might be on an additional tab.  It could perhaps be similar to what is offered in the old Instat.
The second is to extend the options for missing values.  The 2 options currently are what is offered by R as standard.  This is first to give a summary as missing if any values in the calculation are missing.  The second is to only make the result missing if all observations are missing (or so many that the calculation cannot be done.
It would be good to add an intermediate case.  This is to give the result as missing if more than a given fraction (or number) are missing.  For example if more than 3 observations are missing, or more than 10% are missing.  This might add yet a further Missing tab to the Summaries sub-dialogue (or it could be a separate button.)  That might be like Facets in the graphics where there is an initial checkbox labelled:
Missing value options.  If checked the dialogue opens.  There are 2 radio buttons for **Number** or **Percent**.  If Number there is an up-down.  Text says "Summary set to missing if more than this value are missing."  The number for the up-down starts at 0 and is up in 1's with no upper limit.  Default is 3.
The percent starts at 0 and is up in 10s to 90.  The limit if you type in which you can is 99%.
When these are set the number of missing (and the total number) are always calculated (as well as the other statistics.  Then the result is set to missing if needed.

shadrackkibet:
@dannyparsons is percentiles and proportions ready to implement in this dialog?

rdstern:
There are also some other summaries we could add.
1. Make the sub-dialogue a bit wider.
2. Move Range to after Maximum.  It is (after all) just the (Max-Min)
3. Add the quartiles, but separate them, i.e. Lower Quartile, Upper Quartile  and implement them.
4. Put Variance before Standard Deviation.
5. Add Coefficient of Variation - it is just 100*s/xbar  (of course missing if xbar =0)
6. If you are feeling a bit ambitious you could also add skewness and kurtosis (3rd and 4th moments).  This needs a bit of thought on which ones to add.  I suggest you could be consistent with SPSS and SAS perhaps?)  You could look at [https://www.researchgate.net/post/Different_result_of_skewness_and_kurtosis-any_thoughts](url) 
There are various packages and I suggest R package e1071 looks good - use type = 2 in their skewness and kurtosis commands.  (It isn't important enough a topic to allow options - which they do.)

rdstern:
[...]

PR #3969, opened 2017-09-26
Summary

shadrackkibet:
In progress!


shadrackkibet:
@dannyparsons did you discuss with @rdstern the issue of mean function here? Do we need to have a separate function for Trimmed mean?

dannyparsons:
I think what we need is to modify the column summaries function so that we can pass the trim argument to the correct mean function.

shadrackkibet:
@dannyparsons  i think this can go in then we can come into the issue of trimmed mean later since we have an issue open at  #3845

dannyparsons:
Yes that sounds good. I don't see any changes in the R code for the new `summary_skewness_mc` function. Or is it already a function there?

shadrackkibet:
yep,`summary_skewness_mc` is already a function.

PR #3982, opened 2017-09-29
Adding icons to project. 

maxwellfundi:
@dannyparsons  this is a strange thing that i would not have expected. I added the proper icon everywhere its needed but only the resource file for the main form changed... I would not understand whats on the resource file as its strange code really but it works well.


dannyparsons:
Can someone test this on their machine by cloning the branch. I thought it would need the icon file to be uploaded but maybe it doesn't work that way.

shadrackkibet:
Testing this now.

shadrackkibet:
I think the icon is already added to the project. This can be closed.

maxwellfundi:
@shadrackkibet  great the issue seem to have been closed already so i close this too.

dannyparsons:
Yes I did add the icon to frmMain which still had the old one set.

PR #3983, opened 2017-09-29
Website updated 

maxwellfundi:
@dannyparsons  The link is now updated


dannyparsons:
There are two downloads links on the website and they now go to different places. I don't actually see why we should have both. But if we do they should be the same. Are there any links on the ADI website as well to download?

maxwellfundi:
yeas we have the latest link on the ADI website already...

Issue #3880, opened 2017-08-30
Transform Text BUG!


shadrackkibet:
I was testing this dialog, unfortunately i found  a bug .This is what i did:
I clicked on words at the top radiobuttons and then selected a variable to the receiver then on options groupbox i checked on the checkbox  ("or Column" ) to display a receiver then  select a variable to the receiver. On changing to another radiobutton at the top then it breaks at ucrNud. I don't understand why this happens, perhaps someone can look into this?

dannyparsons:
> it breaks at ucrNud

Can you be more specific? What was the error? Is this an error that was caught by the system or did it crash completely? If so, which line caused the crash and did you look at the values of any of the objects? Did you notice anything unexpected from that?

These are some of the details I expect a developer to mention when reporting a bug as compared to a user.

AlexSananka:
can this be closed now?

dannyparsons:
@shadrackkibet you will need to specify the details of the error here for others to be able to help.

shadrackkibet:
These are the two lines causing a crush in this dialog.
`ucrNudFirstWord.SetRCode(clsWordsFunction, bReset)`
`ucrNudLastWord.SetRCode(clsWordsFunction, bReset)`

dannyparsons:
What are the error messages? Can you look at variables causing the crash using watches with @maxwellfundi?

shadrackkibet:
sure, will like to know how to use watches as well. will look into this with @maxwellfundi .

Patowhiz:
@shadrackkibet @dannyparsons @maxwellfundi   just looked into the form. After going through the same procedure as @shadrackkibet  did, I got an error at line 157,inside the subroutine of `SetToValue(objTemp As Object)` of class ucrNud.vb. The error is an `InvalidCastException` .  It seems like during runtime there is an invalid conversion of **RFunction** to String, probably because the code implicitly tries to cast **RFunction** to the **String**.  Am still trying to understand why the code is the way it is. @shadrackkibet are you still working on this?

Patowhiz:
Well upon investigating the issue further. The same error gets thrown every time I check the checkbox of the First word option or the Last word option (the or Column) .  @africanmathsinitiative/developers could someone explain to me the output am to expect from the words first word option or the last word option from the end users perspective.

maxwellfundi:
Great that we have discussed. I think the bug in the ucrNud would be good to be fixed because it might be appearing somewhere else.

[...]

PR #3985, opened 2017-09-29
Transform Text Update and Bug Fixes.

shadrackkibet:
In Progress!


shadrackkibet:
@africanmathsinitiative/developers this is ready for review. However, I have not fixed the bug at #3880

Muthenya:
I have tested for this. @dannyparsons it is ready for review.

instat/dlgTransformText.vb
```diff
@@ -297,69 +302,65 @@ Public Class dlgTransformText
...
             grpParameters.Size = New Size(grpParameters.Width, igrpParameterFullHeight)
-            ucrNewColName.Location = New Point(ucrBase.Location.X, iNewColMaxY)
+            ucrSaveNewColumn.Location = New Point(ucrBase.Location.X, iNewColMaxY)
             ucrBase.Location = New Point(ucrBase.Location.X, iBaseMaxY)
             Me.Size = New Size(Me.Width, iFullHeight)
         End If
     End Sub
 
     Private Sub ucrPnl_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlOperation.ControlValueChanged, ucrInputTo.ControlValueChanged
-        If bRCodeSet Then
```
dannyparsons: You've taken out this line, was that deliberate?

instat/dlgTransformText.vb
```diff
@@ -223,16 +221,25 @@ Public Class dlgTransformText
...
-        bRCodeSet = False
         ucrPnlPad.AddAdditionalCodeParameterPair(clsPadFunction, clsNewRParameter:=New RParameter("side", 2), iAdditionalPairNo:=1)
-
-        ucrReceiverTransformText.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
-        ucrNewColName.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
+        ucrReceiverTransformText.AddAdditionalCodeParameterPair(clsConvertFunction, clsNewRParameter:=New RParameter("string", 0), iAdditionalPairNo:=1)
+        ucrReceiverTransformText.AddAdditionalCodeParameterPair(clsPadFunction, clsNewRParameter:=New RParameter("string", 0), iAdditionalPairNo:=2)
+        ucrReceiverTransformText.AddAdditionalCodeParameterPair(clsTrimFunction, clsNewRParameter:=New RParameter("string", 0), iAdditionalPairNo:=3)
+        ucrReceiverTransformText.AddAdditionalCodeParameterPair(clsWordsFunction, clsNewRParameter:=New RParameter("string", 0), iAdditionalPairNo:=4)
+        ucrReceiverTransformText.AddAdditionalCodeParameterPair(clsSubstringFunction, clsNewRParameter:=New RParameter("string", 0), iAdditionalPairNo:=5)
```
dannyparsons: I think this is causing the error isn't it? You can't link the receiver to the "string" parameter in `clsSubstringFunction` because sometimes it's from the receiver and sometimes the nud. I think that was why it was done manually before, which is the lines you removed below? So I think the only way to prevent that error is to keep the manual parts.
I haven't looked much into the code, but does that sound right?

shadrackkibet: This bug was there even with manual parts.

shadrackkibet: clsWordsFunction is the one with the problem actually.

dannyparsons: Ok I see. I still think you might get a problem with clsSubstringFunction now because you have two controls set to the "string" parameter. But anyway, which parameter was giving the problem in Words?

instat/dlgTransformText.vb
shadrackkibet: This was deliberate.

dannyparsons: And why?

instat/dlgTransformText.vb
shadrackkibet: Start and end which are sometimes passed by Nud's and Receivers.

dannyparsons: Ok so I think all those will need to be done manually then otherwise it gets into a mess trying to set both together.

dannyparsons: I think all these functions do work with a vector of values, but you could check if it does for substring.

shadrackkibet: Yep, i see this can take a vector of values.I will enable this option.

instat/dlgTransformText.vb
shadrackkibet: I don't see the reason why we need this. Do we?

dannyparsons: You should definitely understand what a piece of code does before you decide to remove it, because you can't know whether to remove it without understanding why it was there in the first place.
This was set to `False` at the start of SetRCode and set to `True` at the end. It meant that the manual stuff wouldn't run until SetRCode was finished, and this was important because the manual stuff checks the control values, and they are not up to date until SetRCode has finished, so I think this is definitely still needed because of the manual code we still have. And we don't want manual code to run in the middle of SetRCode.

shadrackkibet:
@africanmathsinitiative/developers this is ready for re-review..

instat/dlgTransformText.vb
```diff
@@ -297,103 +304,104 @@ Public Class dlgTransformText
...
         ElseIf rdoSubstring.Checked Then
             grpParameters.Visible = True
-            grpParameters.Size = New Size(grpParameters.Width, igrpParameterFullHeight / 2.14)
-            ucrNewColName.Location = New Point(ucrNewColName.Location.X, iNewColMaxY / 1.28)
-            ucrBase.Location = New Point(ucrBase.Location.X, iBaseMaxY / 1.26)
-            Me.Size = New Size(Me.Width, iFullHeight / 1.2)
+            grpParameters.Size = New Size(grpParameters.Width, igrpParameterFullHeight / 1.22)
+            ucrSaveNewColumn.Location = New Point(ucrSaveNewColumn.Location.X, iNewColMaxY / 1.09)
+            ucrBase.Location = New Point(ucrBase.Location.X, iBaseMaxY / 1.09)
+            Me.Size = New Size(Me.Width, iFullHeight / 1.07)
```
dannyparsons: Have these controls changed size/location? How did you come up with the new values?

instat/dlgTransformText.vb
dannyparsons: I still think this shouldn't have been removed.

instat/dlgTransformText.vb
shadrackkibet: The location of this controls have changed because i have enabled receivers for the substring function.

instat/dlgTransformText.vb
shadrackkibet: I will add this back.

instat/dlgTransformText.vb
dannyparsons: Ok great

instat/dlgTransformText.vb
dannyparsons: You still have the parameters with two different controls being done automatically instead of manually, so are the bugs still occurring?

shadrackkibet:
Closing this.More discussions and fixes at #4121

Issue #3973, opened 2017-09-27
Country Coloured Map Dialog


dannyparsons:
Following from #3937. I now have an idea for a first dialog for the Procurement menu for mapping. This will be for the simple country coloured map. I'm not sure where this will go in the Procurement menu at the moment, for now it can just go at the bottom. It will probably also go under Describe > Maps as it will be general.

The code for this map is quite simple but there are several steps.

1: `world <- ggplot2::map_data(map = "world")`

From the help, `map_data` "Easily turn data from the maps package in to a data frame suitable for plotting with ggplot2." So this is just getting the outlines of each country in a form that `ggplot2` can plot.

2: `merged_country <- right_join(world, country_counts, by = c(region = "country"))`

This joins our own data (`country_counts`) with the `world` data because the `country_counts` has the values for the colouring - see #3937.

3: `ggplot() + geom_polygon(data = merged_country, aes(x = long, y = lat, group = group, fill = count)) + coord_quickmap()`

The `x` and `y` are obvious, `group` is a variable from `world` and there is one group for each country, in `geom_polygon` this makes sure the points are connected in the right order, `fill` is a variable from `country_counts` for colouring the inside of the polygons.
`coord_quickmap` does sensible default scaling for a map.

I would like more flexibility on choosing subsets of maps etc. but that can come later on a subdialog. For example, there is a `region` argument to `map_data` which allows choosing a subset of countries, and then we might do a `left_join` instead of a `right_join` so that we get the outline of countries not in our data. We could also subset the world data for a latitude, longitude box.

For now, this should be a simple main dialog. A selector with a receiver for the Country column ("country" in my example) and a receiver for the Colour By column ("count" in my example). So the complete code is:

```
world <- ggplot2::map_data(map = "world")
merged_country <- right_join(world, country_counts, by = c(region = "country"))
ggplot() + geom_polygon(data = merged_country, aes(x = long, y = lat, group = group, fill = count)) + coord_quickmap()
```

Other than the country_counts data frame and the "country" and "count" column, everything else is fixed because it comes from world which is sort of an internal data frame - the user won't see this.

In terms of dialog code, I think map_data will be a before function of the base, and then the plot will be the base function, with the right_join as the data argument assigned as a fixed name "merged_country" for now.

[...]

PR #4010, opened 2017-10-06
Enhanced dlgCountryColouredMap

maxwellfundi:
@dannyparsons  have a look at this. 

Thanks.


maxwellfundi:
@africanmathsinitiative/developers  this is ready for review. You could use the world map to test the geom polygon on the general graphics dialog. 

Please make sure to test for both aesthetics and layer parameters.

instat/ucrGeom.vb
```diff
@@ -878,8 +878,21 @@ Public Class ucrGeom
...
+        clsgeom_polygon.AddLayerParameter("position", "list", Chr(34) & "identity" & Chr(34), lstParameterStrings:={Chr(34) & "identity" & Chr(34)})
+        clsgeom_polygon.AddLayerParameter("lineed", "list", Chr(34) & "round" & Chr(34), lstParameterStrings:={Chr(34) & "round" & Chr(34), Chr(34) & "butt" & Chr(34), Chr(34) & "squire" & Chr(34)})
+        clsgeom_polygon.AddLayerParameter("linejoin", "list", Chr(34) & "round" & Chr(34), lstParameterStrings:={Chr(34) & "round" & Chr(34), Chr(34) & "mitre" & Chr(34), Chr(34) & "bevel" & Chr(34)})
+        clsgeom_polygon.AddLayerParameter("linemitre", "numeric", "0", lstParameterStrings:={1})
+
+        'Arrow parameter is a function and should be passed like this
+        'arrow(angle = 30, length = unit(0.25, "inches"), ends = "last", Type = "open")
+        'currently not sure how to do this 
+        'clsgeom_polygon.AddLayerParameter("arrow")
+
```
maxwellfundi: @dannyparsons  i have added in the layer parameters for the geom polygon now and this looks good when i tested with the wold map. That was a good suggestion actually. 

However, there is a layer parameter called arrow as indicated here that am not sure how well we can implement it because the arrow parameter is a function that takes in more parameters  within the geom _polygon

dannyparsons: Seems that arrow can be used in many geoms like line too. I'm fine with not having it for now. The comment is useful so know it's something not there yet.

instat/ucrReceiverExpression.vb
```diff
@@ -38,7 +38,7 @@ Public Class ucrReceiverExpression
         End If
     End Sub
 
-    Public Overrides Sub Add(strItem As String, Optional strDataFrame As String = "")
+    Public Overrides Sub Add(strItem As String, Optional strDataFrame As String = "", Optional bFixReceiver As Boolean = False)
```
dannyparsons: You're not doing anything with the new parameter here? You should pass it through to the base method and do something to the receiver expression so that the parameter works as expected.
Don't spend long on this because it's not a priority at the moment, if it takes too long then maybe just remove the parameter.

maxwellfundi: I think that was an ommision but am adding that now. If it doesnt seem sensible then we can get rid of it

dannyparsons: Let's keep it and make it disable the receiver so it works like the others.

maxwellfundi:
Fixes #3973

instat/ucrGeom.vb
maxwellfundi: great, i thought it was useful to make sure that the next people who would take over this in the event that its not me would have an easier time. We need to comment code better actually for newbies by the way.

instat/ucrReceiverExpression.vb
maxwellfundi: great!

Muthenya:
@dannyparsons this is ready for review.

PR #4017, opened 2017-10-06
VariableSets

Patowhiz:
Added menu item Variable Sets under memu Prepare > Data Frame.  …
I welcome any suggestions to improve on the UI of the Variable Sets dialogue. 
Its functionality is not yet enabled.


instat/dlgVariableSets.Designer.vb
```diff
@@ -135,17 +103,113 @@ Partial Class dlgVariableSets
...
+        '
+        Me.UcrReceiverMultipleVariables.frmParent = Nothing
+        Me.UcrReceiverMultipleVariables.Location = New System.Drawing.Point(21, 109)
+        Me.UcrReceiverMultipleVariables.Margin = New System.Windows.Forms.Padding(0)
+        Me.UcrReceiverMultipleVariables.Name = "UcrReceiverMultipleVariables"
+        Me.UcrReceiverMultipleVariables.Selector = Nothing
+        Me.UcrReceiverMultipleVariables.Size = New System.Drawing.Size(180, 154)
+        Me.UcrReceiverMultipleVariables.strNcFilePath = ""
+        Me.UcrReceiverMultipleVariables.TabIndex = 21
+        Me.UcrReceiverMultipleVariables.ucrSelector = Nothing
```
dannyparsons: This should have a lower case "u"

Patowhiz:
Oh yeah! My machine has a high resolution. Then there seem to be an issue
with the end of file character. Perhaps the change of the designer file

Patowhiz:
Changed the groupbox to panel. Will look at naming

dannyparsons:
End of file characters can be an issue, but that's different, it would show changes to a whole block of lines. In the designer here it's just the sizes which have changed must be caused by this zooming thing new Windows laptops can do now. It's a known issue with VIsual Studio which is quite annoying. 
A work around I have to do, is to set the zooming back to 100%, but then you will probably need to change resolution so that everything isn't too small and use a lower resolution. You will need to sign out and in again for it to take effect. You only need to do this when editing a designer, not Visual Studio in general.
So you will also have to undo the changes on the designer possibly undoing the commit if it's separate, otherwise it's a bit more difficult but others can help you.

Patowhiz:
Okay. I'll ask maxwell for assistance in that part.

PR #4018, opened 2017-10-06
added methods to use variable sets in dialogs

dannyparsons:
@Patowhiz @maxwellfundi these are also useful for the variable sets dialog


Patowhiz:
@dannyparsons will be using them.

Issue #3979, opened 2017-09-28
Improving the right-click menu


rdstern:
In #3978 I have suggested deleting 2 items from the right-click menu and adding the Levels/Labels dialogue.
I also use the **Prepare > Data Frame > Reorder Columns** dialogue quite a lot.  Perhaps that could go into the top section, either before or after **Duplicate Column**.
If we want to make the right-click menu shorter, then I have never used the section on hiding columns.  There are 3 dialogues there.  Could they be deleted from this right-click menu- at least for now?

shadrackkibet:
@dannyparsons?

dannyparsons:
I agree making it shorter is needed and some of the converts could go, all the hide and freeze could go as well. 
Then adding levels/labels seems sensible too. But this should only be enabled when only one column is selected and that column is a factor.

shadrackkibet:
@dannyparsons This can now be closed.

rdstern:
The Levels/Labels on <Right-Click> works really well for Factor Columns.  But it is not enabled for Ordered Factors.  It should be.  Thanks

dannyparsons:
Thanks, hadn't noticed that.

rdstern:
Please could Levels/Labels be spelt like this.  It is currently Levels/Lables.
At the same time I suggest we could get rid of (or hide at this stage) the 2 entries **Insert Column Befor**e and **Insert Column After**.  Does anyone use them?

PR #4064, opened 2017-10-17
Right Click edits.

shadrackkibet:
@dannyparsons this are minor edits suggested at #3979


instat/ucrDataView.vb
```diff
@@ -47,6 +47,8 @@ Public Class ucrDataView
         mnuFreezeToHere.Enabled = False
         mnuUnfreeze.Enabled = False
         grdData.Visible = False
+        mnuInsertColsBefore.Enabled = False
+        mnuInsertColsAfter.Enabled = False
```
dannyparsons: Let's hide them instead of disabling. And remember any change here you must also make on the column metadata right click.

shadrackkibet: We also have freeze disabled.Shall i hide this as well?

dannyparsons: I think freeze we probably want to get working eventually so good to leave it there so people see it is coming.

shadrackkibet: great!

shadrackkibet: @dannyparsons i have now changed this.

dannyparsons:
Great, is levels/labels spelt correctly on column metadata already?

shadrackkibet:
yep, that is spelt correctly.

PR #4073, opened 2017-10-18
Adding options to show or hide special menus

maxwellfundi:
@dannyparsons  thats now done.  I need to add the same on the view menu now

### DIFF
--- a/instat/dlgOptions.vb
+++ b/instat/dlgOptions.vb
@@ -69,6 +69,8 @@ Public Class dlgOptions
         rdoSpanish.Enabled = False
         ucrNudDigits.SetMinMax(0, 22)
         ucrChkIncludeCommentsbyDefault.SetText("Include Comments by Default")
+        ucrChkViewClimaticMenu.SetText("Show Climatic Menu")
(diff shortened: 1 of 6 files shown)


instat/dlgOptions.vb
```diff
@@ -360,6 +366,22 @@ Public Class dlgOptions
...
+        End If
+    End Sub
+
+    Private Sub ucrChkViewProcurementMenu_ControlValueChanged() Handles ucrChkViewProcurementMenu.ControlValueChanged
+        If ucrChkViewProcurementMenu.Checked Then
+            frmMain.mnuProcurement.Visible = True
+        Else
+            frmMain.mnuProcurement.Visible = False
+        End If
+    End Sub
```
dannyparsons: Why are you changing the visibility while the dialog is open? And why is the visibility being changed here and not in frmMain or the options class?

dannyparsons: Nothing should happen when the check box is changed. That's not how this dialog works. Have a look at other controls on the Options, do we do anything when their value changes?

PR #4077, opened 2017-10-19
Adding Geom_smooth

maxwellfundi:


maxwellfundi:
@africanmathsinitiative/developers  I have put up some parameters for the geom_smooth here which can be tested. Just one issue to be addressed here though, layer parameters reset on reopening

rdstern:
What is the state of this now?  It would be great to be able to include this geom soon.

maxwellfundi:
The current pull request on this is working with a couple of parameters not
all. It has an issue with resetting too but I will have it fixed by end of
week hopefully.
Thanks

rdstern:
Great that geom smooth is there.  

Is there re strong reason why the **formula** parameter is missing?  That is what makes it really powerful.  I realise it has to be given in terms of y and x, rather than our own column names.

maxwellfundi:
The formula, as indicated on the comments on the code, is missing because we currently do not show textboxes that the user would type into for the layer parameters. This will be possible when we will be able to do that.

PR #4091, opened 2017-10-23
Additions to Prepare menu

### LINKED ISSUE #4062: Add copy of Describe > Specific > Boxplot to Prepare menu
Could we replace the current **Prepare > Check Data > Jitter** by a copy of the Boxplot dialogue and enable it.  It would be given as **Prepare > Check Data > Boxplot**.

maxwellfundi:
Fixes #4062 

@rdstern  this is the new look for the Prepare>Check Data menu

image


dannyparsons:
ellipses needed on the new items

maxwellfundi:
@dannyparsons  Added Ellipses already

rdstern:
Looks good.  Minor suggestions:
a) Have the 4 new options (Boxplots etc) all just below Duplicates
b) then a line before the 2 Open Refine options
c) If the One way Frequencies is the same as Describe > One Variable > Frequencies, then call it One Variable Frequencies rather than One-Way.

Issue #3992, opened 2017-09-30
Prepare > Data Frame > Replace Values issue 


rdstern:
[METADATA_RAINFALL_GUYANA_PICSA.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/1346287/METADATA_RAINFALL_GUYANA_PICSA.xlsx)

This is the first Excel file where I had to infill missing values (from above).  You may not be able to repeat this error message.  Try it as I did.  **It was the second time I opened the dialogue that I got the error statement**.
1. **I opened the Excel file** given here.  But I got it wrong and did not leave one row when opening.  I also did not notice that immediately.  (You will see the odd looking column names.)
2. I opened the dialogue **Prepare > Data Frame > Replace Values**.  
3. I changed the options so I was **replacing missing by values from above**.
4. Then I realised the data frame wasn't appropriate, so I **cancelled and then deleted the data frame**.
5. I then **Opened again from the second line**.  So the column names were now fine.
6. I then re-opened this dialogue and got the following message:

**instat

Developer error: no state of control ucrPnlOld satisfies it's condition. Cannot determine how to set the control from the RCode. Modify control setup so that one state can satisfy its conditions.

OK**   

I pressed ok and the dialogue opened fine and I was able to proceed.  I did notice that the radio buttons were both on Missing, which was not where they were before.

dannyparsons:
We've prevent the error message but the code still needs to be fixed so that this doesn't occur.

lilyclements:
Closing as presumably resolved - I cannot replicate but it was reported eight years ago

PR #4099, opened 2017-10-25
Replace Values

shadrackkibet:
@dannyparsons  this is a temporary fix for the developer error #3992.


dannyparsons:
Don't mind this for now to prevent the error as long as the issue stays open. Is there another open pull request editing this file? I won't merge this yet if there is.

shadrackkibet:
There is no open pull request  on this elsewhere.

dannyparsons:
I think I was confusing this with transform text so that's ok.

PR #4113, opened 2017-10-31
Simple changes

Muthenya:
@maxwellfundi please review this. Thank you.


maxwellfundi:
@Muthenya  it seems you need to use the ucrSave here. Check how its used on other dialogs and implement it here too. I am also sending you a climatic data set that you can use to test them too.

Muthenya:
Thank you for the insight @maxwellfundi.

dannyparsons:
I've mentioned in the issue related to this the problem using the ucrSave in these dialogs. The difference is they produce columns in a **different** data frame to the one selected. So the ucrSave can't detect the correct column name in the same way.
I think we might have to find a different temporary solution for this.

PR #4119, opened 2017-11-01
ShowModel Panel Reset on Reopening.

shadrackkibet:
@africanmathsinitiative/developers this fixes reset on reopen.


Patowhiz:
Just checked it. it works fine. However why `rdoQuantiles `the default checked rdo yet `rdoProbabilities `is the one that starts in arrangement ? I propose you change its arrangement if its the quantiles is the default. Just for UX.

Patowhiz:
Does this set the `rdoProbabilities `as the default checked radio button ? I noticed in design the arrangements of the 2 radio buttons the `rdoProbabilities `comes first then followed by `rdoQuantiles`, yet by on initialisation and on reset the checked radio button is `rdoQuantiles`.
dlgshowmodel

dannyparsons:
Sounds fine. I don't think there's a reason they need to be in this order so they can be swapped for consistency of the first being the default one checked.

shadrackkibet:
Will change this so that probabilities becomes the default.

shadrackkibet:
> Will change this so that probabilities becomes the default.

I switched the radio buttons on the designer instead .This is to avoid more changes i the dialog code setup.

@dannyparsons this is ready for re-review.

PR #4121, opened 2017-11-03
Transform text

Patowhiz:
#3880 Transform Text BUG
#3985 Transform Text Update and Bug Fixes

Fixed ucrNud and dlgTransformText
Eliminated the InvalidCastException

However I still have several questions about dlgTransformText.vb . Seems like it has a lot of redundant code but I could change because am still trying to understand why the author of the code structured it the way it is. Looking forward to a discussion over this.


dannyparsons:
Firstly I just notice that your variable sets dialog is in this pull request. We want to be very strict about having one pull request per task. It's difficult to review, manage and prevent conflicts when tasks are mixed in the same branch.
It's easy to miss this but once you get in the routine of doing this it should be easy.
We can leave this as it is and try to get it merged to not waste time, but in future these should have been separate pull requests and done on separate branches.

dannyparsons:
@shadrackkibet Can you test the transform text dialog and review the changes, since you were working on this it should be easier for you. I'm happy with doing parameters manually if they use a mix of value types to prevent errors and it's sensible. 
I'm not convinced that doing things like `ucrSelectorForTransformText.CurrentReceiver Is ucrReceiverLastWord` is needed since we haven't needed this check in any other dialog yet. I haven't looked carefully how this is being used so might be ok, but would like someone else to review first and then convince me.

instat/ucrNud.vb
```diff
@@ -147,14 +147,31 @@ Public Class ucrNud
...
-        Else
-            If Decimal.TryParse(objTemp, dNewValue) AndAlso dNewValue >= nudUpDown.Minimum AndAlso dNewValue <= nudUpDown.Maximum Then
+        Else
+
+            'If its NECESSARY parameter objTemp to be of type OBJECT then I  think it should 
+            'explicitly be converted to STRING to prevent future errors of InvalidCastException.
+            'the Decimal.TryParse(<String>, result as decimal) takes a string as one of its parameters
+            'In my testing a situation can occur where the passed objTemp value can't be implicitly cast to string
+            'this has happened with inappropriate setting of ucrNud controls.
+            If Decimal.TryParse(objTemp.ToString, dNewValue) AndAlso dNewValue >= nudUpDown.Minimum AndAlso dNewValue <= nudUpDown.Maximum Then
```
dannyparsons: You're absolutely right that we shouldn't pass `objTemp` directly into `Decimal.TryParse` because it must take a string argument. We never want to pass variables into functions without knowing they are the right type. So very well done spotting this.

Notice that `Protected Overrides Sub SetToValue(objTemp As Object)` is overriding a method in ucrCore which is why the argument is an `Object` since `ucrCore` is completely general to all controls. And we can't change the type of arguments when overriding a method. So yes having the argument as decimal would be preferable but not possible in this case. (I think we could avoid this by "shadowing" the method instead of overriding but that's a bigger change to look into later).

As long as `objTemp.ToString()` doesn't throw an error this is now safe. I think the only thing that could cause that is if objTemp is nothing but that's checked above so this looks good. I'm happy with this, thanks for spotting and fixing the problem.

You probably just want to update your inline comments now that we've agreed on this.

Patowhiz:
@dannyparsons sorry I forgot to create a new local branch for Text Transform Bug from my master. I realised later that the new created branch was from my variable branch. Mistake noted.

@shadrackkibet please review the variable sets and Text Transform dialogues.

Patowhiz:
@africanmathsinitiative/developers I hope this has been reviewed. We agreed with @dannyparsons  this review to be for Transform Text Bug and Variable Sets. So @dannyparsons I'm going remove the inline comments for ucrNud then I'll open a pull request that will be for variable sets and transform text as we had agreed. So that this can be merged and closed. Is that okay?

shadrackkibet:
@Patowhiz  i see you have changes from Transform Text and Variable sets which one should be reviewed?

shadrackkibet:
It will be good if you have this changes on separate branches.

Patowhiz:
Yeah. I made a mistake there. So forgive me in that.

shadrackkibet:
Two issues to sort out:

- The two Nud's  when on word radiobutton do not reset. The defaults were 1 and 2 previously. you can set this defaults on linking since we are now passing the parameters manually.Have `bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=1` and on other nud as well.

- When on words, i check the two checkboxes for "or column" and then reopen the dialog this checkboxes unchecks. You could have a look at this? I understand this issue was there previously.

instat/dlgTransformText.vb
```diff
@@ -404,5 +396,30 @@ Public Class dlgTransformText
...
+            clsWordsFunction.AddParameter("start", strParameterValue:=ucrNudFirstWord.Value, iPosition:=1)
+        End If
+
+        If ucrChkLastOr.Checked Then
+            clsWordsFunction.AddParameter("end", clsRFunctionParameter:=ucrReceiverLastWord.GetVariables(), iPosition:=2)
+        Else
+            clsWordsFunction.AddParameter("end", strParameterValue:=ucrNudLastWord.Value, iPosition:=2)
+        End If
+    End Sub
+
```
shadrackkibet: please remove this spaces.

shadrackkibet:
The other issue was enabling receivers so that we can pass a column of values for the` start `and `end` parameters on substring option but this could come later.Could you enable this since it is the similar to words options?

Patowhiz:
@shadrackkibet to which conrol are you proposing me to link ucrNuds to ?

shadrackkibet:
I am suggesting you to add `bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=1` on this line which is already there;`ucrChkFirstOr.AddToLinkedControls(ucrNudFirstWord, {False}, bNewLinkedDisabledIfParameterMissing:=True)` and on the other linking as well.

Patowhiz:
That will reset the ucrNuds whenever a user checks or unchecks the checkboxes for or column

shadrackkibet:
oooh i see. I think we can now reset this on defaults.

instat/dlgTransformText.vb
Patowhiz: Done

dannyparsons:
We can't have values of controls being set in SetDefaults because this conflicts with the controls reading from the R code to set their value.
I'm aware this has taken a while to solve this, so I'm happy if we get to something stable now and we can come back to fix this properly later as there are more urgent tasks for everyone at the moment.

instat/ucrNud.vb
```diff
@@ -147,14 +147,13 @@ Public Class ucrNud
...
+    Protected Overrides Sub SetToValue(objTemp As Object)
         Dim dNewValue As Decimal
-
         If objTemp Is Nothing Then
             'If no value reset to a default value
             nudUpDown.Value = nudUpDown.Minimum
-        Else
-            If Decimal.TryParse(objTemp, dNewValue) AndAlso dNewValue >= nudUpDown.Minimum AndAlso dNewValue <= nudUpDown.Maximum Then
+        Else
+            If Decimal.TryParse(objTemp.ToString, dNewValue) AndAlso dNewValue >= nudUpDown.Minimum AndAlso dNewValue <= nudUpDown.Maximum Then
```
Patowhiz: Removed the previous comments I had put in here for discussions. We agreed `objTemp `parameter type cannot be change to Decimal. So `objTemp.ToString` is necessary to avoid `InvalidCastException` for `ucrNud `controls

Patowhiz: @shadrackkibet this is now ready for review

Patowhiz:
*@dannyparsons *I welcome any suggestion on how we can initialise such
controls. In this case I have given them default values in setDefaults. The
2 controls do not have any R code attached to it including parameters.

instat/ucrNud.vb
shadrackkibet: This is ok now.

shadrackkibet:

dannyparsons:
We do need all controls to be linked to functions in the same way otherwise we will not be able to populate controls from R code. But let's come back to this later when we have time to work on it.

@Patowhiz there's also lots of manual setting of R code in your variable sets dialog which I want to question but given your issues with branches we can merge this now and come back to it.

I suggest you now delete your repo if there are no other active branches and then refork to have a clean master. If you need help understanding how manage branches then you can check with the team for help on that.

Issue #4058, opened 2017-10-16
Changes to One Variable Summarise dialogue


rdstern:
First, just a change of name!  
- [x] 1. Could the dialogue please be called **One Variable Summarise**.  It is the numerical "partner" for **One Variable Graph**.  But it is currently a bit confusing (in writing the tutorial) that it is called Describe One Variable.

- [x] 2. Also Change Max to Maximum in the **Max Factor Levels Shown** control.

- [ ] 3. Now a bit more on the changes I am afraid.  I wonder also if some of the small points below should be covered in the initial testing of the dialogue.  I show the problem first:

a) Open the diamonds data set.
b) For the two columns x and carat get the default summaries and see how they are laid out.
c) Now include customise.
d) Get the number of observations and the number not missing, and then the same as the default, i.e. the min, max, median and the upper and lower quartiles.
- [x] e) Compare the results.  I am concerned by the names of these summaries.  Could we change them to get rid of "summary_". 
- [x] f) Could we also look into not having scientific notation unless it is needed.  If you omit the number of observations and number not missing than the format is OK.
- [ ] g) Could we check on the number of decimals (significant figures).  Possibly have an option on that in the dialogue.  But make the default the same as R does in its summaries.

shadrackkibet:
@dannyparsons   Here **e,f,g** needs changes in the R code of our summaries function. what are you views on this?

shadrackkibet:
@rdstern  should we also rename multiple selector label to Variable(s) to Summarise?

rdstern:
Well spotted.  Yes indeed.  That is then consistent with the **Prepare > Column: Reshape > Column Summaries** dialogue.

dannyparsons:
This whole R function actually needs a rewrite for other reasons too so would be good to sort these out.

dannyparsons:
@shadrackkibet are 1 and 2 done here?

shadrackkibet:
Yep

PR #4136, opened 2017-11-06
OneVariableSummarise,dlgHideDataframes & FrmMain changes.

shadrackkibet:
@africanmathsinitiative/developers .

- Fixed changes suggested at #4058 on onevariablesummarise.
- Sorting out naming convention on the climatic menu.
- Fixed reset of Tailored menus.
- Resized selector on dlgHidedataframes.


instat/frmMain.vb
```diff
@@ -220,6 +220,8 @@ Public Class frmMain
         mnuViewDataFrameMetadata.Checked = False
         mnuViewColumnMetadata.Checked = False
         mnuViewScriptWindow.Checked = False
+        mnuViewClimaticMenu.Checked = False
+        mnuViewProcurementMenu.Checked = False
```
dannyparsons: This isn't the default at the moment. Currently the default is that both are visible. Interesting question whether the "defaull layout" includes changing which menus are hidden. I think maybe not but it's a good question. Could you remove this for now and we have it as an issue of what should be done here.

shadrackkibet: Noted.

PR #4143, opened 2017-11-06
add link dialog localisable

dannyparsons:


Patowhiz:
@dannyparsons @africanmathsinitiative/developers now that we have decided to use the multilingual-app-toolkit does this mean code `autoTranslate(Me)` of class `Translations `will not be used? And if so should we omit it in all future dialogues?

Patowhiz:
@maxwellfundi @shadrackkibet @Muthenya . On checking the videos that are on the multilingual-app-toolkit page I think the translation process is easy. I have tested how this works and am comfortable with how its implemented.  How do we split this task ? I suggest that we list the dialogues that each one is working on and are unmerged so that we exclude from working on them. Then we can assign to each other which dialogues to work on.

maxwellfundi:
@africanmathsinitiative/developers have have agreed we do this according to menus.
@Patowhiz - File Menu
@Muthenya - Prepare menu
@shadrackkibet - Climatic
And I will be on the Describe Menu. 

@Patowhiz  We might not need to have the autotranslate(me) because this was used when we were doing it manually by adding tags of respective strings on the resource files. I think @dannyparsons  understands how the translations class came about.

Patowhiz:
@maxwellfundi great. In the meantime, I think dlgVariableSets and dlgTransformText should not be touched. I have made changes to them and they are yet to be merged.

dannyparsons:
Sounds like a good strategy.

The Translations class will still be needed for text that is not picked up by the localisation of the form. e.g. anywhere we use SetText within a dialog will not be picked up and will require use of the Translations class, although it will have to be modified now.

maxwellfundi:
I am now done with the describe menu. I am going to the user controls and dialogues. 

@dannyparsons  you are right if it has to be used then it needs to be modified. 

Thanks.

Patowhiz:
All dialogues under file menu have been set localized to true

shadrackkibet:
I am done with  climatic menu. I am going to work on procurement menu.

PR #4145, opened 2017-11-07
 localizable property for Describe>One Variable dialogs

maxwellfundi:


dannyparsons:
These changes look correct, do you want it merged now?

maxwellfundi:
Yeah they could be merged and used as "guinea pigs" to see if it works properly

dannyparsons:
There won't be any visible changes but I've checked carefully and there's two file changes for each dialog so that looks good.
The only thing to check is that the dialog design still looks the same. It should be because it's linked to the resource file.

PR #4153, opened 2017-11-07
Translations prepare

Muthenya:
@dannyparsons please review these as I work on the rest.


dannyparsons:
You seem to be missing designer changes for Replace Values and for Row Names there are changes to the French resource file and not the main resource file. So something not quite right there.

Muthenya:
@dannyparsons I have included the missing changes. Though, there seems to be a problem with the dlgAddKey dialog. It only gives the change in the designer and not in the resource file, I haven't included it here.

dannyparsons:
ok let's do that one later then. Row names is still an issue with the French file appearing here.

Muthenya:
Okay. Please direct me on how to fix that.

dannyparsons:
I'm not sure, there shouldn't be any changes on the French file. You can try undoing those changes.

PR #4157, opened 2017-11-07
Import dataset designer.

shadrackkibet:
@dannyparsons this is a tiny fix on the designer.Some part of the ucrbase seems to be hidden.


dannyparsons:
There seems to be changes to the size of every control here? Was this branch made after the File menu dialogs were localised or before?

shadrackkibet:
Yep i made this changes after dialogs were localized. I Just resized the dialog size.

dannyparsons:
@Patowhiz did you not resolve the issue of your machine having high resolution causing changes to the sizes in designer within VIsual Studio? I'm now seeing here that the sizes on this dialog have changed.
From this I suspect the File menu dialogs will have to be redone, and you should not make any designer changes while your screen is in high resolution with zooming.

PR #4162, opened 2017-11-08
File menu designer fixes

shadrackkibet:
@dannyparsons .This are designer fixes.This is ready for merging.


dannyparsons:
yes there should be a pair of files for each dialog

dannyparsons:
I think it's actually only the resource file changes we need so this is ok.

Issue #4128, opened 2017-11-04
Calculations - new keyboard


rdstern:
I (very imaginatively) called this keyboard **More**, which Danny doesn't like.  So ideas welcome on a more informative and appropriate name.  Perhaps **Transform**?

Two important functions are to lag a column and also to take differences.

Here is the suggestion.  4 rows as always, and 4 columns (so far).

1. First column from the top is **lag, diff, ecdf, ntile**.  The lag and diff are from the zoo package (which is already installed).  Then ecdf is also simpler as cume_dist in dplyr and ntile is also in dplyr.
2. Second column is all from zoo and is moving (running) statistics, namely movmax, movsum, movmean, movmed (or movmedian if there is space).  (Note there isn't a moving min, but you could do -movmax(-x)!
3. Third column is cumulative functions, namely cummax, cumsum, cummean, cummin.  (Here all except cummean are in base, while cummean is in dplyr.)
4. Fourth column is ranks in different ways and is all from dplyr. It is rank, m_rank, d_rank, %rank.  (rank is row_number, m_rank is min_rank, d_rank is dense_rank and %rank is percent_rank.

And advance notice that we may also need a symbols keyboard, with :, %%, %/%, [ ], etc.  Not sure what else is needed or where it should go?

shadrackkibet:
Starting some work on this. Where should this be in climatic menu for now.

rdstern:
That's great.  For now it should just stay where it is.  So it is just an extra keyboard in the Prepare Columns: Calculations > Calculations dialogue.

The functions are very useful for climatic, but when they go there it could be a simpler dialogue for lagging, differences, etc.  We start with the general facility

shadrackkibet:
I see we already have most of this options in our current calculator.Does this mean i add this options since they are from this packages?

rdstern:
I don't understand? Where do we have these options for this keyboard, i.e. diff, lag, etc?

shadrackkibet:
In the dropdown we have;

![image](https://user-images.githubusercontent.com/26301973/32546071-e6ef306c-c48e-11e7-934a-27f0b4b00173.png)

shadrackkibet:
We have some of this options on **Rows** and **logical and symbols**.

rdstern:
You are right.  I hadn't looked there, because (on the rows) keyboard it is only really the pmax and pmin that are row operations.  But you could expand that keyboard, rather than starting a new one.  Let's call it transform for now.

[...]

Issue #4126, opened 2017-11-03
Calculations - Logical keyboard


rdstern:
We currently have a keyboard called Logical and symbols.

I suggest it becomes a Logical keyboard only.  It still has  4 rows and 4 columns; the keys could become wider if needed in some of the columns.

1. First column could be **< , <=, > >=**.  They are all there now.
2. Second column could be **==, !=, |, &.**  (i.e. equals, not equals, or, and). They are all there now, but & is currently given as And.
3. Third could be **ifelse, match, when and between**.  These are new.  **ifelse** is really **useful**. match is the same as %in% and the others help to turn missing rows into logical.  For **ifelse** consider using the if_else function from dplyr. (Still call it ifelse on the keyboard.)  when is case_when from dplyr and between is also from dplyr.
4. The fourth column is **is.na, !is.na, duplic (for duplicated) and near**.  These are new again.  near is from dplyr.

In addition we could have a 5th column with any, all, isTRUE and isFalse.  However, they are different as they give a single value.  Needs discussion to see if they are needed.

dannyparsons:
There's no `isFALSE` command in base, unless it's in another package I don't think it exists. Otherwise this looks good.

rdstern:
I am not sure we need the 5th column of functions at all in the keyboard.  One item missing is xor(), so that could be instead of isFALSE.  But neither isTRUE or isFALSE may be that useful.

[https://stat.ethz.ch/R-manual/R-devel/library/base/html/Logic.html](url) does give the impression that isFALSE is available?

dannyparsons:
Very strange it's there in documentation but not in R. I updated to the latest version of R but still not there. Maybe it's been removed? Does anyone else have this function in their R?

shadrackkibet:
Working on this in combination with #4128

shadrackkibet:
@dannyparsons  this can now be closed.

PR #4174, opened 2017-11-09
dlgCalculator & ucrCalculator.

### LINKED ISSUE #4125: Calculations - mainly Statistics keyboard
There are going to be a series of issues concerned with the calculator.  Perhaps Maxwell could propose one person who might concentrate on it.  Perhaps the work can start on this dialogue in 0.4.7 - which I have given as the current milestone.  But it is unlikely to finish then.  It also depends on the other tasks for AMI staff.

The first set - in this issue concerns the current Statistics keyboard.

1. Please could it be called **Summary** instead, both in the pull-down and also above the keyboard.
2. The keyboard is currently 3 columns of 4 rows.  Others are larger.  I now suggest **5 columns by 4 rows**.
3. **All the existing 12 items remain**, though I suggest in a different order.
4. Most of them use the default argument na.rm=FALSE.  So when there are any missing values the result is missing.  I suggest this should be changed to **na.rm=TRUE**.  
5. I think they all work, except quantile.  This has an additional argument names=TRUE.  It should change to **names=FALSE**.
6. The width of the buttons could change, so the full names can be on a single line.  This is currently not the case for **quantile**.
7. The button called count should change to say **length**.  The word count is perhaps clearer, but length is the command in R and we might as well use that when we can. 
8.  The common feature of this keyboard is that each button here produces a single summary value.  If it displays in the output window, or with the helpful Try feature in the dialogue we can see this.  When writing to a column it produces a column with this single value repeated.
9. It is a good time to add tool-tips.  I'll give them below.
10. The first column is then length (old count), non-miss (that's new and in our summary), miss (already there) and distinct (also new, it is the n_distinct function from dplyr.)
11. The second column is sum, mean, var, sd (all there now).
12. 11. The third column is max, median, min, range  (all there now).
13. The 4th column is cv, iqr, skew, anyDup, (iqr is there now, cv, and skew are all now part of the Summarise dialogue.  anyDup is short for anyDuplicated in the Data Table package and can be typed now.)
14. The fifth column is mad, mc, propn, quantile  (mad and mc are part of the summarise, quantile is there now and propn is new.)

So propn (proportion) is new.  It is the inverse of quantile.  For quantile you supply a value, e.g. 0.5 and it gives the corresponding value.  For example with 0.5 it gives the median.  For proportion you give the value, e.g. 10 and it gives the proportion.

It is described in the R discussions as coming from the ecdf function.  Calculated directly it is (I think) just mean(x= **column** <= **value**, na.rm = TRUE).  Put the default as 1 as the value, just as 0.5 is the default for the quantile.

The only other new one is the number of unique values in a column.  This is given by length(unique(x = **column**)).

shadrackkibet:
@africanmathsinitiative/developers I have make improvements to the calculator here. This fixes #4125,#4128 and #4126.

This is ready for review and testing.

### DIFF
--- a/instat/dlgCalculator.designer.vb
+++ b/instat/dlgCalculator.designer.vb
@@ -39,19 +39,19 @@ Partial Class dlgCalculator
     <System.Diagnostics.DebuggerStepThrough()>
     Private Sub InitializeComponent()
         Dim resources As System.ComponentModel.ComponentResourceManager = New System.ComponentModel.ComponentResourceManager(GetType(dlgCalculator))
-        Me.ucrCalc = New instat.ucrCalculator()
         Me.ucrBase = New instat.ucrButtons()
+        Me.ucrCalc = New instat.ucrCalculator()
         Me.SuspendLayout()
-        '
-        'ucrCalc
-        '
-        resources.ApplyResources(Me.ucrCalc, "ucrCalc")
-        Me.ucrCalc.Name = "ucrCalc"
         '
         'ucrBase
         '
         resources.ApplyResources(Me.ucrBase, "ucrBase")
         Me.ucrBase.Name = "ucrBase"
+        '
+        'ucrCalc
+        '
+        resources.ApplyResources(Me.ucrCalc, "ucrCalc")
+        Me.ucrCalc.Name = "ucrCalc"
         '
         'dlgCalculator
         '
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -134,19 +134,19 @@ Public Class dlgCalculator
     Private Sub CalculationsOptions()
         Select Case ucrCalc.ucrInputCalOptions.GetText
             Case "Maths"
-                Me.Size = New System.Drawing.Size(iBasicWidth * 1.33, Me.Height)
+                Me.Size = New System.Drawing.Size(iBasicWidth * 1.38, Me.Height)
             Case "Logical and Symbols"
-                Me.Size = New System.Drawing.Size(iBasicWidth * 1.23, Me.Height)
-            Case "Statistics"
-                Me.Size = New System.Drawing.Size(iBasicWidth * 1.15, Me.Height)
+                Me.Size = New System.Drawing.Size(iBasicWidth * 1.46, Me.Height)
+            Case "Summary"
+                Me.Size = New System.Drawing.Size(iBasicWidth * 1.49, Me.Height)
             Case "Strings (Character Columns)"
-                Me.Size = New System.Drawing.Size(iBasicWidth * 1.24, Me.Height)
+                Me.Size = New System.Drawing.Size(iBasicWidth * 1.41, Me.Height)
             Case "Probability"
                 Me.Size = New System.Drawing.Size(iBasicWidth * 1.59, Me.Height)
             Case "Dates"
                 Me.Size = New System.Drawing.Size(iBasicWidth * 1.3, Me.Height)
-            Case "Rows"
-                Me.Size = New System.Drawing.Size(iBasicWidth * 1.21, Me.Height)
+            Case "Transform"
+                Me.Size = New System.Drawing.Size(iBasicWidth * 1.35, Me.Height)
             Case Else
                 Me.Size = New System.Drawing.Size(iBasicWidth, Me.Height)
         End Select
--- a/instat/ucrCalculator.Designer.vb
+++ b/instat/ucrCalculator.Designer.vb
@@ -19,7 +19,7 @@ Partial Class ucrCalculator
     Inherits System.Windows.Forms.UserControl
 
     'UserControl overrides dispose to clean up the component list.
-    <System.Diagnostics.DebuggerNonUserCode()> _
+    <System.Diagnostics.DebuggerNonUserCode()>
     Protected Overrides Sub Dispose(ByVal disposing As Boolean)
         Try
             If disposing AndAlso components IsNot Nothing Then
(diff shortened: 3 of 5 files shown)


dannyparsons:
Great, can you post screenshots? Then Roger can check them too

shadrackkibet:
Here are the screenshots;

image

image
image
image
image
image
image

rdstern:
They look great.  This is a big improvement.  Well done.
1. Maths: sign is higher than the rest in the key.  It should be sign.  Is it currently that and not sigin?
2. Logical and symbols.  I had not thought of having more keys below, but quite like the arrangement?  Interesting what others think.
3. In Logical have **between**, not Between.  Also (if you can keep the size, perhaps you can have **duplicated** in full, rather than duplic? 
4. In transform swap movsum and movmax.

shadrackkibet:
@africanmathsinitiative/developers  this is still ready for review.

shadrackkibet:

dannyparsons:
Can someone review and test all the new buttons add the correct commands?

Patowhiz:
I'll test this

Patowhiz:
@shadrackkibet how do I run something like sin(90 degrees) ? I have realised the sin,cos and tan function operate on radians

shadrackkibet:
> @shadrackkibet how do I run something like sin(90 degrees) ? I have realised the sin,cos and tan function operate on radians

By default R give this in Radians .Mathematically, `1 degree =0.0174532925 radians`
so, just do this in the calculator; `sin(90*0.0174532925)` e.tc.

Patowhiz:
@shadrackkibet  well my suggestion is, if possible a functionality that allows you to operate on degree directly. Such that internally you convert it to radians before passing it to R. This could be by checking for something like sin(90 degrees). The presence of degrees prompting the conversion even for columns.
Though this can be left as it is since the user can decide to have a column  that has degrees converted to radians then get a sin of that column.

Patowhiz:
@shadrackkibet when I click on button IQR of the summary calculator the focus doesn't head to "x=<here>" in the expression IQR(x=, na.rm = TRUE, type =7). So in a case where I have column Size instead of the generated expression to be IQR(x=Size, na.rm = TRUE, type =7) , its IQR(xSize=, na.rm = TRUE, type =7). Seems like there is a problem with the next character to put cursor on

dannyparsons:
There does seem to be a few R packages which have these conversion functions. But the conversion is so simple, `(deg * pi) / (180)`, I don't think we want to add a new package just for that. I hope we have `pi` in the calculator already? Then it's very easy.
If users are seriously using angle columns then we should encourage them to use radians since a lot of mathematical operations on angles only work when using radians. I don't think this will be very common for a stats package though.

Patowhiz:
Yeah ,True. A tooltip or a help  explanation that explains your **(deg * pi) / (180)** conversion in performing the trigonometric functions will help the users

Patowhiz:
@shadrackkibet when I click on button Try to run this command `stringr::str_locate(string= Village., pattern= 'SAB')` For caculator Strings. It gives me start end... as the output in the try textbox. Is this the expected behaviour or should it give me a concrete value like 1    3 . Not this works as expected when I click on Ok . 2 columns are generated, one with the start value and another with the end value.

Patowhiz:
@shadrackkibet there is some inconsistency in the command generated for the following commands on the part of pattern=

- Button **Replace** generates - ` stringr::str_replace(string= ,pattern= ,replacement= )`

- Button **Locate** generates   -  `stringr::str_locate(string= , pattern= )`

- Button

 **Count** generates    -  `stringr::str_count(string= , pattern ='')`

- Button

 **Combine** generates -  `stringr::str_c(string= , pattern='')`

- Button **Detect** generates -  `stringr::str_detect(string= , pattern= )`

- Button **Split** generates - `stringr::str_split_fixed(string= , pattern='', n= )`

- Button **Extract** generates - `stringr::str_extract(string= , pattern= )`

Am not sure which is the best option between generating the commands with pattern='' or just pattern=
But I think it will be good UX to maintain consistency. 
However when applied with the right input the commands execute as expected.

rdstern:
I am enjoying the discussions on this checking.  And this reminds me that we could spend some time next week on adding tooltips.

Patowhiz:
@shadrackkibet Is the label **Strings (Character column)** meant to tell the user that the calculator is for columns **with** strings or is for columns of **type** character? Am curious to know why the extension "Character column"

Nevertheless I think all the commands generated in this calculator can operate even on factors and even numeric columns. I have tested this.

Patowhiz:
@shadrackkibet I have only found the above issues. The rest appears to be okay to me.

shadrackkibet:
ok. Will address some of this issues.Thanks .

shadrackkibet:
> @shadrackkibet when I click on button IQR of the calculator Summary the focus doesn't head to "x=" in the expression IQR(x=, na.rm = TRUE, type =7). So in a case where I have column Size instead of the generated expression to be IQR(x=Size, na.rm = TRUE, type =7) , its IQR(xSize=, na.rm = TRUE, type =7). Seems like there is a problem with the next character to put cursor on.

@shadrackkibet there is also a similar problem when I click on button Propn

@shadrackkibet the button pad of calculator Strings seems to have a similar problem

@shadrackkibet the button pnom of calculator Probability seems to have a similar problem. Check also if the it gives the expected result when executed, am not sure about it.

@shadrackkibet the button movmax ntile movesum movmed movmean of calculator Transform seems to have a similar problem.

All this are working for me.... I noticed that the comma is close to the equal sign which makes it hard for the user to  insert the cursor after the equal sign. However i have moved the comma away so that at least it is clear to the user.

Patowhiz:
Yeah that should sort out the focus issue

shadrackkibet:
@Patowhiz i have sorted out some of the of the issues you raised above. This is ready for re-review.

Patowhiz:
Will check

shadrackkibet:
@Patowhiz  is this ready for @dannyparsons ?

Patowhiz:
@dannyparsons The above issues have been addressed. @shadrackkibet I think this is ready

PR #4177, opened 2017-11-09
layer order bug fix and general graphics dialog display fixed

### LINKED ISSUE #4098: Adding layer to boxplot dialog
I added a layer to the boxplot dialog. It also worked except the second layer appeared under the boxplot, which wasn't very useful. I assume this is just an issue with parameter positions. Can this be fixed? And checked on all other graphics dialogs?

dannyparsons:
fixes #4098

### DIFF
--- a/instat/dlgScatterPlot.vb
+++ b/instat/dlgScatterPlot.vb
@@ -130,7 +130,7 @@ Public Class dlgScatterPlot
 
         clsBaseOperator.SetOperation("+")
         clsBaseOperator.AddParameter("ggplot", clsRFunctionParameter:=clsRggplotFunction, iPosition:=0)
-        clsBaseOperator.AddParameter("scatter", clsRFunctionParameter:=clsRScatterGeomFunction)
+        clsBaseOperator.AddParameter("scatter", clsRFunctionParameter:=clsRScatterGeomFunction, iPosition:=2)
 
         clsRggplotFunction.SetPackageName("ggplot2")
         clsRggplotFunction.SetRCommand("ggplot")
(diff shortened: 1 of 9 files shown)


instat/dlgScatterPlot.vb
```diff
@@ -130,7 +130,7 @@ Public Class dlgScatterPlot
 
         clsBaseOperator.SetOperation("+")
         clsBaseOperator.AddParameter("ggplot", clsRFunctionParameter:=clsRggplotFunction, iPosition:=0)
-        clsBaseOperator.AddParameter("scatter", clsRFunctionParameter:=clsRScatterGeomFunction)
+        clsBaseOperator.AddParameter("scatter", clsRFunctionParameter:=clsRScatterGeomFunction, iPosition:=2)
```
dannyparsons: @maxwellfundi here I have added a position for the geom on all graphics dialogs, this was part of what was needed to solve layer order. I think position 2 is sensible for the first geom layer. Let's remember to add a position to every parameter, it's especially important for ggplot.

instat/ucrAdditionalLayers.vb
```diff
@@ -46,6 +48,7 @@ Public Class ucrAdditionalLayers
             strTempGeomName = GetGeomName(clsTempParam)
             If strTempGeomName <> "" Then
                 AddLayers(strTempGeomName, bFromLayerSubDialog:=False)
+                iParameterPosition = Math.Max(iParameterPosition, clsTempParam.Position)
```
dannyparsons: @maxwellfundi here I'm calculating the maximum current position of all geoms to ensure that additional geoms always have a larger position. This needed to be different from iLayerIndex because the layer index always starts at 1, whereas the position depends on the positions of the current layers, which are set on the main dialog.

instat/dlgScatterPlot.vb
maxwellfundi: Great, thanks.

PR #4187, opened 2017-11-10
added missing resource file for add key

dannyparsons:


dannyparsons:
@africanmathsinitiative/developers I just found the resource file for add keys was missing meaning opening the dialog caused R-Instat to crash. I added it here, luckily it's not a very important dialog. Can you all check opening all dialogs before we create the next version to check for missing resource files?

Patowhiz:
Checking

Issue #4189, opened 2017-11-11
Issues in Version 0.4.7


rdstern:
There are some issues with the new radio buttons.  I don't know how serious.

1. Prepare > data Frame > Insert Columns/Rows ony has one button for columns (I think).
2. Delete columns rows looked fine - the first time I went in.  But then not the second time.
3. Check data > Duplicates buttons moved but all usable.
4. Same for boxplots.  I guess they are always a bit messed up.  I'll only report when they are unusable.
5. One Variable > Frequencies ok first time then buttons move.
6. Factor > Contrasts crashes, but can press Continue.  Guess resource file again?
7. The text columns have the new buttons to be fixed.  **Transform** gave errors a few times, but then showed the dialogue.  No error message second time.
8. String > Distance is more extreme than the others in moving the buttons.
9. Add comment is not yet used, but the button is seriously moved.
10. Rename Data frame doesn't work and gives error.
11. Copy Data Frame has same problem.
12.  Now into Describe menu - 2 variable graph has a problem.  Then shows the dialogue, but it is a bit messed up.
13. Bar chart had radio button so pie chart is not visible.
13a Model all OK - ish
14. Climatic Open SSTs doesn't and gives error.
15. Check > Display daily has unusable button after first time in dialogue.
16.  Rest of climatic sort of Ok as is Procurement.
17. ODK has unusable buttons on second time round.  Rest of File is OK.

maxwellfundi:
Am checking on Describe now.

maxwellfundi:
@africanmathsinitiative/developers I have not figured exactly what is causing the mess in the new style radio buttons. 

I suggest for this version, we simply have a temporary solution by getting back the localizable= False for the dialogs that have issues as it was before, before we figure out completely whats happening.  

What are your thoughts?

Patowhiz:
And also I think it's important to **LOCK** the controls once you are **DONE** with designing the form. There is a **feature** in visual studio  for locking a form to ensure the controls don't move even when they are opened in other computers. I have gone through the entire forms and I have seen several forms having the radio buttons offset.

dannyparsons:
@patowhiz I'm not sure Lock will solve this problem because the issue is occurring when we apply the resources during translating - but it might do so worth checking. The reason for the offset is the way in which we use ucrPanel and buttons in them, it's an issue about the relative location being different after the first time it's opened.

[...]

PR #4190, opened 2017-11-11
Boxplot designer fixes

maxwellfundi:
Messed up version 
image

Current version

image


rdstern:
I understand the messing up is because of the resource files.  It is a feature of most of the dialogues with our new style buttons.  Sometimes the mess is worse than with boxplots as I found and reported in #4189, for example in Prepare:

**Prepare > Data Frame > Insert Columns/Rows** only has one button for columns (I think).
**Prepare > Data Frame >Delete columns/rows** looked fine - the first time I went in. But then not the second time. 

I don't know whether we correct individually or there could be a more general fix?

maxwellfundi:
I am trying to look at that issue of dialogs ending up with one radio button on the panel as it is on the barchart and it is different from the boxplot problem so i think these would be looked at separately. 

I realize when i change localizable property of the bar chart, the radio buttons are okay on the panel. But when i edit the designer file for the same dialog when localizable=true, i still get one radio button. 

Am trying to figure out what is changing on the resource file that affects the designer of these panels.

dannyparsons:
The problem can be solved - at least temporarily - by not translating the panels so that the locations don't get shifted. I'll send a pull request on this soon, then hopefully no dialogs will need editing.

PR #4202, opened 2017-11-16
Scriptwindow

### LINKED ISSUE #4115: Making the script window editable
This is not just from me, but now from David!  We think it is a good possible task for Maxwell?

It is to make the Script window editable.
Then to be able to submit (run) either the window or the selected section.

According to David there are then 3 or 4 ways this could be run.  So the alternatives would be on a right-click menu.
There is then a health warning that running after editing the code carries a health risk.  Do you want to continue.  Like the warning when you delete columns.

But if it crashes we do now have our recovery options, so that's a good thing!

maxwellfundi:

Fixes #4115

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -14,11 +14,7 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-
 Public Class ucrScript
-
-    Public clsRLInk As New RLink
-
     Public Sub CopyText()
         txtScript.Copy()
     End Sub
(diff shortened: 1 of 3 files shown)


instat/ucrScript.vb
```diff
@@ -51,9 +51,9 @@ Public Class ucrScript
         clsRLInk.bOutput = True
 
         For Each strLine As String In strScript.Split(Environment.NewLine)
-            strLine = strLine.Replace(vbLf, String.Empty)
-            strLine = strLine.Replace(vbCrLf, String.Empty)
-            clsRLInk.RunScript(strScript:=strLine, iCallType:=0, strComment:="Code run from Script window", bSeparateThread:=False, bSilent:=True)
+            If strLine.Trim(vbCrLf).Count > 0 Then
```
maxwellfundi: @dannyparsons  so this code works and the only issue i got now  is that it doesnt sent the strline to the output window. 

Am not sure why and from look of things, the runscript is used properly. Its the issue i have have for both log and script window. 

What are your thought on this so far?

instat/ucrScript.vb
```diff
@@ -14,7 +14,11 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+
 Public Class ucrScript
+
+    Public clsRLInk As New RLink
```
dannyparsons: Why are you using your own instance of the RLink?

maxwellfundi:
have merged stuff fom here to #4209

PR #4225, opened 2017-11-22
Opening Script from Script Window 

### LINKED ISSUE #4115: Making the script window editable

maxwellfundi:

To test it, run the software and send something to script, then from the script window right click > Open Script

Fixes #4115

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -41,7 +41,7 @@ Public Class ucrScript
     Private Sub mnuRunWholeScript_Click(sender As Object, e As EventArgs) Handles mnuClearContents.Click
         Dim dlgResponse As DialogResult
         If txtScript.Text <> "" Then
-            dlgResponse = MessageBox.Show("Are you sure you want to clear the " & Me.Text, "Clear " & Me.Text, MessageBoxButtons.YesNo)
+            dlgResponse = MessageBox.Show("Are you sure you want to clear the contents of the script window " & Me.Text, "Clear " & Me.Text, MessageBoxButtons.YesNo)
             If dlgResponse = DialogResult.Yes Then
                 txtScript.Clear()
             End If
@@ -51,6 +51,9 @@ Public Class ucrScript
     Private Sub mnuRunSelectedText_Click(sender As Object, e As EventArgs) Handles mnuRunSelectedText.Click
         If txtScript.SelectionLength > 0 AndAlso txtScript.SelectedText <> "" Then
             RunText(txtScript.SelectedText)
+        Else
+            MessageBox.Show("You need to select some text before running" & Me.Text, "No text selected" & Me.Text, MessageBoxButtons.OK, MessageBoxIcon.Warning)
+
         End If
     End Sub
 
@@ -84,4 +87,5 @@ Public Class ucrScript
             End
         End Try
     End Sub
+
 End Class
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -14,8 +14,10 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports System.IO
 Public Class ucrScript
     Private strComment As String = "Code run from Script Window"
+    Public strRInstatLogFilesFolderPath As String = Path.Combine(Path.GetFullPath(FileIO.SpecialDirectories.MyDocuments), "R-Instat_Log_files")
 
     Public Sub CopyText()
         txtScript.Copy()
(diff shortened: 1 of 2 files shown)


maxwellfundi:
@dannyparsons  I have gotten the script to open including adding the top 3 lines to set up. My query is that, how will i have now to extract the data that is being used in the script. 

For example in this case, i loaded the Survey data, and produced a boxplot, then I went back to the boxplot and did ToScript, which i have run and opened in RStudio like shown here. 

On the import dialog, i did not sent anything to script and so there Survey  could not be foung when i run this script. 

Assuming that someone may do this more than once. (Sending things to script) and from different dialogs which are using different dataframes, how can we extract the data from the instatobject so that the script doesnt throw an error. 

Whats your opinion?
 
`setwd(dir="C:/Users/Fundi/Source/Repos/R-Instat/instat/bin/Debug/static/InstatObject/R")
source(file="Rsetup.R")
InstatDataObject <- instat_object$new()
# Code generated by the dialog, Boxplot

survey <- InstatDataObject$get_data_frame(data_name="survey")
last_graph <- ggplot2::ggplot(data=survey, mapping=ggplot2::aes(y=Yield, fill=Fertgrp., x=Village.)) + ggplot2::geom_boxplot() + theme_grey()
InstatDataObject$add_graph(graph_name="last_graph", graph=last_graph, data_name="survey")
InstatDataObject$get_graphs(data_name="survey", graph_name="last_graph")
rm(list=c("last_graph", "survey"))`

dannyparsons:
By opinion is quite simple, that's for the user to worry about. If they take a partial script into R then they need to sort out getting it to work. If that's not something they yet understand how to do then they probably shouldn't be using the script in this way.
I don't think this should be a button because I don't think it will be used that much. It's only slightly quicker than doing copy and paste so I'd be happy if it was on the right click. Then it's slightly hidden and won't confuse users as much.

shadrackkibet:
Testing this.

maxwellfundi:
Great thanks.

shadrackkibet:
**Just two  quick things to note:**

This perfectly opens in R as expected. 
1. when you right click and click on **clear contents** the message box doesn't display full message , some contents are hidden.
image
2.I see **Run selected text** is enabled when no text is selected.I think it should be disabled at this instance.

maxwellfundi:
@shadrackkibet  I have now addressed the issues on the message box. I have also added the message to show when no text is selected. 

About not seeing the selected text when the message box appears i think its generally how it works but these could be features we can look at in the future. 

@dannyparsons  this is now ready for you.

dannyparsons:
This sounds good. Can someone test the changes are ok?

Patowhiz:
Yeah sure. @shadrackkibet should I tackle this or you are on it.

shadrackkibet:
@Patowhiz  you can go ahead and test this.

Patowhiz:
Okay

instat/ucrScript.vb
```diff
@@ -41,7 +41,7 @@ Public Class ucrScript
     Private Sub mnuRunWholeScript_Click(sender As Object, e As EventArgs) Handles mnuClearContents.Click
         Dim dlgResponse As DialogResult
         If txtScript.Text <> "" Then
-            dlgResponse = MessageBox.Show("Are you sure you want to clear the " & Me.Text, "Clear " & Me.Text, MessageBoxButtons.YesNo)
+            dlgResponse = MessageBox.Show("Are you sure you want to clear the contents of the script window " & Me.Text, "Clear " & Me.Text, MessageBoxButtons.YesNo)
```
shadrackkibet: I think this is a question we should have a question mark at the end.

Patowhiz: @maxwellfundi @dannyparsons  I think the best way will be to have the icon of the MessageBox be a question mark, rather than the "?". Or any other icon that expresses the message. The MessageBox has default icons that can be used.

maxwellfundi: I think we need to be consistent with our message boxes. We have not thought about that really but i agree we might want to have a question mark other than the warning sign.

maxwellfundi: Thos is what we got now. 

image

shadrackkibet: @dannyparsons  this is ready for you to review.

Patowhiz:
@dannyparsons I think the changes are are okay. The dialogue works correctly and the issues have been addressed

dannyparsons:
Looks good, and I like the question mark icon. We should use this whenever we have questions in message boxes.

Patowhiz:
Yeah. VB.NET has a couple of the icons. And you can have custom made for
rinstant. People react to and understand visuals much better than plain
text.

Issue #4184, opened 2017-11-10
Restore data from backup


dannyparsons:
Original idea from #3959

This is a new dialog on the File menu, not sure where yet.

When R-Instat crashes, you have the option of opening the AutoSaveRecovery dialog, and one of the options there is to save/load the backup data file.

It would be useful to access the backup data at any time, not just after a crash.

So this new dialog will allow you to see the current backup data files (currently at most one but could be more in future) and then allow you to select one to save it somewhere as a file or load it directly into R-Instat.
The loading would replace the currently open data so a big warning is needed about saving your current file before doing this.

This will not be a standard dialog because it is not running R code in the same way. So it won't need our usual base buttons.

I suggest it has:

- A listbox showing the current backup data files (could be 0 so you can't do anything). This will have two columns for the filename and the Date/Time it was saved. In future there could be more than one which is why a listbox.
- Two radio buttons to choose whether you are saving the backup or loading the backup into R-Instat.
- If you are saving: A button, label and textbox appears to choose where to save a selected backup. This can be very similar to what is currently on the AutoSaveRecovery dialog.
- If you are loading: A button appears to load the selected data into R-Instat (replacing the current data). A big warning needed before clicking this button.
- A Help and Cancel button at the bottom of the dialog as well.

This is just a suggestion to start, so it's probably good to do the design first and then share with others for feedback.

Patowhiz:
![dlg2](https://user-images.githubusercontent.com/21179007/33014442-5b44b230-cdf8-11e7-8527-70a7a856b0ae.JPG)
@africanmathsinitiative/developers more suggestions here.
@dannyparsons I have followed the suggestions of @rdstern and the designs below show how am planning to achieve this functionality. Am not using the usual base buttons and I have chose to use the listview instead of the listbox. 

The default selected tab will depend on whether there is an existing opened data frame. If there is one then the Backup Data tab will be selected otherwise the Previous backups tab will be the one selected.

The backup operation will maintain a registry of file locations used in the case of a user initiated backup. This will help in displaying the system default backups and the user backups. The find button will help the user find a backup from an unregistered location e.g a flashdisk ,network etc .

Am still thinking about the schedule tab. Though its supposed to give the user the freedom to choose backup time frequency. I know it has not been mentioned so probably I could ignore this.

Looking forward to suggestions.

[...]

PR #4235, opened 2017-11-23
Backup manager

Patowhiz:
@africanmathsinitiative/developers This is ready for review, this dialogue is supposed to solve the issue raised in #4184 .
This form can be loaded just like it is i.e ` dlgBackupManager.ShowDialog() ` without setting any additional properties .


Patowhiz:
@shadrackkibet I have raised all the other previous issues except this 

>6. when i close R-Instat and then run it again.On coming back to the dialog i find my buckup file.When i click at it it doesn't load it in the data view window.

I'll get back to it once @dannyparsons reverts back to me

PR #4241, opened 2017-11-24
Shadrackkibet rainfall

dannyparsons:
@shadrackkibet I've fixed the filter OR combination issue for the checkboxes. There is still a few things to get sort out on the dialog so please test it further but it is looking good and usable for Roger next week so that's great and we will merge this for today.


shadrackkibet:
Great.It is good to see you have helped to sort out this . We are going  to apply the same with @Muthenya on the temperature dialog because they are similar.

dannyparsons:
It would be very good if the temp dialog could be working today as well, as that would be really useful next week for Roger.

shadrackkibet:
ok, we are going to fix this for the release.

PR #4245, opened 2017-11-24
Adding group for geom_line

### LINKED ISSUE #4238: Group aes for geom_line
The group aes is very useful for geom_line http://ggplot2.tidyverse.org/reference/aes_group_order.html

In the layer subdialog for Line Plot there is no group aes so it needs to be added.

maxwellfundi:
Fixes #4238


maxwellfundi:
@dannyparsons  I have added this to take in Factors/numerics because it can take either

PR #4260, opened 2017-11-27
Adding group parameter for Lineplot

### LINKED ISSUE #4237: Line Plots with factor on x axis
It is possibly in `geom_line` to have a factor on the x axis and get the points to connect. By default you get nothing because it groups by the factor and there's only one point per group. We can override this by adding `group = 1`:

`ggplot(data=Dodoma_by_Year, mapping=aes(x=Year, y=sum_Rain, group = 1)) + ggplot2::geom_line()`

image

So I think we just need to modify the Line Plot dialog to add `group = 1` when x is a factor.

It also works when x is character, although I'm not sure we want to do this for character as well since we should probably encourage having the column as factor or numeric if you are plotting a line?

maxwellfundi:

Fixes #4237


shadrackkibet:
Testing and reviewing this.

shadrackkibet:
group parameter works fine. However, i noticed that  **x** and **y** parameters in the **aes** function don't have parameter positions. Is there a reason why this don't have parameter positions?

maxwellfundi:
yeah thanks, that was not included in the setdefaults but i have not set those. @dannyparsons  this is now ready for you :-)

shadrackkibet:

dannyparsons:
Can you test what happens when other aes are added? I think we may get similar issue as we have found with boxplot i.e. if group parameter is present what happens when I also add colour aes?

maxwellfundi:
No it does not have same issue as boxplot. This works well.

PR #4264, opened 2017-11-27
Adding Coordflip to the sdgplots

### LINKED ISSUE #4251: How do I do coord_flip?
This may be a silly question, but I couldn't find it.

I can do coord_flip on some of the graphs, because it is on the main dialogue.  

I don't mind that it is missing on some, (e.g. Line Plots, Scatterplots)  because they have enough on the main dialogue.  And I haven't yet missed it for those graphs - but where is it?

I assumed it might be on the Coordinates Tab, but there is nothing there.  If I am right, then could it go there? The only complication would be that it should automatically be the same as on the main dialogue, for those plots that now have it in 2 places.

I can now do it, by editing the command in the script window.  Great we have this at last.  But there should be an easier way for this feature.

maxwellfundi:

I have moved the horizontal plot check boxes from main dialog to the sdgplots under coordinates tab. So to test this, run any graphics dialog and check how it works. 

Fixes #4251

### DIFF
--- a/instat/dlgBarAndPieChart.Designer.vb
+++ b/instat/dlgBarAndPieChart.Designer.vb
@@ -47,7 +47,6 @@ Partial Class dlgBarAndPieChart
         Me.rdoBarChart = New System.Windows.Forms.RadioButton()
         Me.cmdBarChartOptions = New System.Windows.Forms.Button()
         Me.ucrSaveBar = New instat.ucrSave()
-        Me.ucrChkFlipCoordinates = New instat.ucrCheck()
         Me.ucrReceiverByFactor = New instat.ucrReceiverSingle()
         Me.ucrReceiverFirst = New instat.ucrReceiverSingle()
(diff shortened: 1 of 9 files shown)


dannyparsons:
I don't think we want to remove it from the main dialogs that we already have it on, but just add it to the subdialog so any graph can do this.

maxwellfundi:
Okay.

shadrackkibet:
Testing and reviewing this now.

maxwellfundi:
This actually needs to be redone @shadrackkibet  because there is a commit that i removed the coordflips for main dialogs that we need to keep.

shadrackkibet:
This is working fine. One point to note ` theme_grey() ` has no package name.

maxwellfundi:
@shadrackkibet  thanks for your review. I think that must have been an ommision. will add that too.

maxwellfundi:
Will open a new one with coord flips only on the sdgplots.

maxwellfundi:
Opening a new pull request on this

Issue #2392, opened 2016-12-20
Including the QC from Rclimdex and Rclimdex extra in R-Instat


rdstern:
[Aguilar-Nanjing-2013-Presentation2 QC R-CLIMDEX.pdf](https://github.com/africanmathsinitiative/Instat/files/663948/Aguilar-Nanjing-2013-Presentation2.QC.R-CLIMDEX.pdf)
[Manual_rclimdex_extraQC.r.pdf](https://github.com/africanmathsinitiative/Instat/files/663949/Manual_rclimdex_extraQC.r.pdf)

The quality control in R-Climdex is fairly clearly shown in the presentation, which is the first of the files above.  It is by Enric Aguilar, who works in this area.  It produces a whole set of files for each station, for users to examine.  It also calculates the 27 indices from WMO.

These indices are a separate dialogue for us.  They use the climdex-pcic package that is in CRAN.  I strongly suggest we make use of this package, even if we can produce those indices ourselves.  This is also the recommendation from one of the original authors, from Canada.

In contrast, the qc procedures are written in R, but are not a CRAN package and were written a long time ago.  I wrote to Enric (who visited Reading earlier and I also met in a subsequent workshop.) to ask his advice.  His reply was also sent to Patrick Parrish, from WMO, who we will meet in Reading in January.  It is below:

"First (suggestion from me) ) is to keep the results within the same software environment, rather than writing output files. : this is the way it should go. 100% agree. RClimdex was written 15 years ago (time flies); it started as Climdex, with excel files, and then the "R" was added, so there was a story of using external excel files. Besides, we were ensuring that computers with limitted resources could work. The extra-qc was a plugin done very quickly and I had no time to think of a better interface.

A couple years ago, I wrote a piece of code called quickqc.R, to extend the capabilities of extra-qc to whole networks and add a few more capabilities. I have a descriptive power point, only in Spanish, but I know this is not a problem in your capable team." 

[QUICKQC.pptx](https://github.com/africanmathsinitiative/Instat/files/663982/QUICKQC.pptx)

"Second is to broaden the checks to examine all climatic elements.

We did some attempts to do so for the UERRA project. I coded some QC for pressure data. There is no way you can use it, as it needs very specific data, but I am attaching it, as it can give you some ideas about what to look at. I know some other colleagues also coded QC for wind. I know that the guys at the SMN Argentina (Dr. Maru Skansi, mms@smn.gov.ar) did a tremendous work on this."

[...]

Issue #2339, opened 2016-12-02
Climatic Analyses - quality control


rdstern:
This issue is to set out some of the general facilities we need for the quality control of the (daily) climatic data.

This is perhaps the parallel to the climate indices, in that the same basic software that calculates the 27 indices (RCLIMDEX) has RCLIMDEX-PCIC for the indices, and RCLIMDEX-QC for the quality control.  My general view on our proposed strategy have been re-enforced by my visit to the ClimDex people in Canada, and also correspondence with a key person on quality control of climatic data - Enric Angular.

For the indices we (at least initially) fit within the existing R package as far as possible. But for the QC we take the ideas from RCLIMDEX-QC and totally redo it ourselves.  I suggest there could be 3 key components to what we produce and they are all relatively general concepts already in R-Instat.
1) The rows of data where observations look "odd" needs a filter.  That is already working.  What is not yet operational, but has already been asked for, is to permit "or" in the filtering conditions.  To help users we may have (say) 10 different QC checks and we would like to know if the data - in any month - or from any element is "odd" for at least one of the checks.
2) The set of checks will produce a monthly data frame - from daily data.  (It might later produce a daily data frame from hourly data).  This is because the data are usually entered from monthly sheets.  Possibly we will produce a monthly data frame for all months/stations and then have special comments for the oddities.  Or we may produce one with just the months where there are oddities.  It won't matter, because if we have all the months, we can always filter later.
3) Initially we may want to do the checks one at s time.  Each check will add a column to the monthly data frame.  We don't even need the "or" on the filter for this option.  Ideally (as in the current R-Climdex-QC) we would be able to produce a set of checks all together, producing (probably) multiple columns in the monthly data frame.
4) We should also be able to indicate any "suspect" month within the daily data frame.  Now our boxplots come in.  We do the boxplots as we can now, but with a different colour - say - for any month that is suspect.  Possibly we can do this already.
5) The only real inconvenience in the current boxplot "system" is the axis labels for the years.  Ideally we should be able to just give a subset of the labels if it is an ordered factor.  Or we use the theme aspect to be able to display the x-axis at 90 degrees.  The latter will be easy.  I am not sure about the former.

dannyparsons:
@rdstern are we doing anything on this before 0.4.1?

PR #4266, opened 2017-11-27
QC Temperature.

shadrackkibet:
In Progress!

### DIFF
--- a/instat/dlgClimaticCheckDataTemperature.Designer.vb
+++ b/instat/dlgClimaticCheckDataTemperature.Designer.vb
@@ -30,8 +30,6 @@ Partial Class dlgClimaticCheckDataTemperature
         Me.lblMonth = New System.Windows.Forms.Label()
         Me.lblDay = New System.Windows.Forms.Label()
         Me.lblElement1 = New System.Windows.Forms.Label()
-        Me.lblRangeElement1 = New System.Windows.Forms.Label()
-        Me.lblRangeElement2 = New System.Windows.Forms.Label()
         Me.lblRangeElement1to = New System.Windows.Forms.Label()
         Me.lblRangeElement2to = New System.Windows.Forms.Label()
         Me.lblElement2 = New System.Windows.Forms.Label()
@@ -102,16 +100,6 @@ Partial Class dlgClimaticCheckDataTemperature
         '
         resources.ApplyResources(Me.lblElement1, "lblElement1")
         Me.lblElement1.Name = "lblElement1"
-        '
-        'lblRangeElement1
-        '
-        resources.ApplyResources(Me.lblRangeElement1, "lblRangeElement1")
-        Me.lblRangeElement1.Name = "lblRangeElement1"
-        '
-        'lblRangeElement2
-        '
-        resources.ApplyResources(Me.lblRangeElement2, "lblRangeElement2")
-        Me.lblRangeElement2.Name = "lblRangeElement2"
         '
         'lblRangeElement1to
         '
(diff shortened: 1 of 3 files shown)


shadrackkibet:
@rdstern  i am thinking of adding another checkbox for Range in this dialog. Here is a screenshot i suggest. What are your suggestions?
image

shadrackkibet:
@rdstern ?

rdstern:
I like the new layout:
a) Space in labels before (Tmax) and (Tmin)
b) Could make the defaults for Tmax 10 to 50 and 0 to 30 for Tmin.
c) Jump I would make the default 10mm
d) Difference the default is 0.
e) Outlier is default of 2.5 and it isn't mm.  Make it **multiplier**.  There could be a tooltip to say **"Values that are further than this number of IQRs from the corresponding quartile."**

shadrackkibet:

dannyparsons:
There should be a space after Range on the checkboxes. Also not sure why it is **Range (Element 1) Tmax**. I don't think the **Tmax** part is needed.

The Range options also confused me when I first used the dialog as I wasn't sure if it was a range for acceptable or unacceptable values. Can the label be changed to **Acceptable Range** instead?

Also, just to confirm the condition for range should be `!(Element1 > 0 & Element1 < 50)` i.e. show me all the values outside (0, 50) range. I may have specified this wrong in the issue so just checking this is what's implemented.

shadrackkibet:

shadrackkibet:

Patowhiz:
@shadrackkibet I'm reviewing this too

shadrackkibet:
@Patowhiz  are you still reviewing this?

Patowhiz:
@shadrackkibet 
1 )
This is more of a question and a suggestion .
In the **Define Climatic Data** dialog we have **Minimum Temperature** and **Maximum Temperature** , is it possible to have them pre filled by default to **element2(Tmin)** and **element1(Tmax)** respectively ? I can see the **Station** and **Date** are being pre filled by default or has that been omitted on purpose? especially after the user has defined the climatic data.

2) 
In the **Acceptable Range (Element1)** and **Acceptable Range (Element2)**  , I presume the "mm" after their respective `ucrNud` controls are unit of measurements. Is "mm" a  correct unit of measurement for this dialog? Or does it refer to something else other than millimeter. I presume the same applies to **Jump** and **Difference**. Temperature measurement are recorded in Kelvin or Celsius.

3)
This is more of design. @rdstern  now that we according to the current dialog design Element1 will always be **Tmax** and **Element2** will always be **Tmin**, why can't we just name Acceptable Ranges checkboxes to **Acceptable Range(Tmax)**, **Acceptable Range(Tmin)**. And also remove the naming **Element1(Tmax)**, **Element(Tmin)**  to remain with just **Tmax**, **Tmin** as labels of their respective receivers? I think that will make the design much simpler and straightforward.

4)
 @rdstern  Still on labelling from my understanding, we are using **Acceptable Range** to mean that we don't want values of **ElementX** _from_ .. _to_ ... . I find that a bit confusing cause in that case we are simply defining **Unacceptable/Undesired Range** (any of its antonym) , which clearly expresses the intention.

5)
@shadrackkibet I can see the default lowest and highest values for **Tmax** range  is **-50** to **65** 
 respectively . But for **Tmin** its **-50** to **50**. Is that intentional? I thought they are to be the same.

6)
From my understanding of Difference, its meant to get the difference between Tx and Tmin, if that's is its true purpose then I think it's odd to have its highest value set as 5. I could want to have a difference of 10.

7)
Currently a new data frame is created with data after the "filtering" has been done. The naming is done automatically, @rdstern I think it will be much better if we had a ucrSave control for a user to get control over the naming of the new data.

8)
The` GMetData2.RDS` (Ghana_Data) has its **Tmax** and **Tmin** data in 1 decimal place **(float values)** yet the filter ranges(ucrNuds  controls used here) only accept whole values (integer values). Is this meant to to clearly tell the user that you can't get ranges and difference in decimal numbers. For instance if I didn't want values that are in a range of 21 to 22.4 probably because I want to assume 22.4 as 22 and 22.5 as 23 . This won't be possible.

Patowhiz:
https://github.com/africanmathsinitiative/R-Instat/issues/2392

shadrackkibet:
@Patowhiz  thanks for testing this.
Since there is no major bug in this dialog, I suggest we have this changes in the issue since it is still open then we can have more discussion and input from others and we can then agree on what to do and what not to.I think this can go in for today's release since we have better functionality compared to what we have now in the master.

Patowhiz:
I agree.

shadrackkibet:
Could you please move the above comments to issue #2339.

Patowhiz:
Done

shadrackkibet:
great!

PR #4277, opened 2017-11-30
Calculator

### LINKED ISSUE #4205: Calculator with filters
Currently the calculator ignores a filter.
The calculator in Genstat gives a popup menu when a filter is applied.  It has 3 options:
1) Remove filter before calculation
2) Remove filter before calculation and then re-apply it
3) Keep the filter in operation

I don't like the popup menu in Genstat - it is annoying. But I would like to be able to use the calculator on only a filtered portion of the data.

We could have a checkbox, or have an option in the Data Options.  Or we just do the 3rd option above - as follows:
a) If the results go into a new column, then the unfiltered rows are made into NA.
b) If the results go into an existing column, it leaves the unfiltered rows as they were, and just changes those that are active.

Note there is an issue on the row statistics, which mentions the same point for those calculations.

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes #4205.

### DIFF
--- a/instat/ucrCalculator.Designer.vb
+++ b/instat/ucrCalculator.Designer.vb
@@ -204,6 +204,9 @@ Partial Class ucrCalculator
         Me.cmdAcos = New System.Windows.Forms.Button()
         Me.cmdCos = New System.Windows.Forms.Button()
         Me.cmdSqrt = New System.Windows.Forms.Button()
+        Me.cmdnth = New System.Windows.Forms.Button()
+        Me.cmdLast = New System.Windows.Forms.Button()
+        Me.cmdFirst = New System.Windows.Forms.Button()
         Me.ucrInputCalOptions = New instat.ucrInputComboBox()
         Me.ucrSaveResultInto = New instat.ucrInputComboBox()
         Me.ucrInputTryMessage = New instat.ucrInputTextBox()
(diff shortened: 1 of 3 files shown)


rdstern:
Great.  Could you also add cor and cov to the summary keyboard.

shadrackkibet:
@rdstern i noticed that **rad** and **grad** in the calculator doesn't work. Which functions should be included here? i had a quick look at what might be useful here https://stackoverflow.com/questions/32370485/r-convert-radians-to-degree-degree-to-radians

shadrackkibet:
@rdstern ?

rdstern:
Well spotted and sorry, I was confused by your grad above.  I assume you meant deg?
Anyway my suggestion is that the package circular (which is already loaded in R-Instat) has **deg** and **rad** as functions.  So let's use those.

They can already be typed in to the calculator, e.g. circular::rad(90) or circular::deg(2*pi)

shadrackkibet:
yep i meant **deg** sorry for the confusion.

shadrackkibet:

shadrackkibet:
@africanmathsinitiative/developers  i have added `nth`,`first `and `last ` functions to the calculator. I also corrected cursor position on other functions. 

This is still waiting you reviews!

Patowhiz:
Reviewing this.

shadrackkibet:
@Patowhiz  did you finish reviewing this?

Patowhiz:
The nth,first and last functions of the calculator , cursor positions are all good.  @shadrackkibet this works fine on my side

shadrackkibet:

PR #4295, opened 2017-12-05
dlgRestorelevelHierachy

maxwellfundi:
@rdstern  

Here is now the design for the dlgRestoreLevelHierachy. 

image

I am not sure the function to be used on this dialog. An instatobject method or an R function.


dannyparsons:
I think it's going to be using the same function as Column Summaries and just doing the `first` summary.
I also suggested some changes to the dialog design in the issue which could update the design.

rdstern:
That looks good.  Two changes are as follows:
1) Make the first receiver into a multiple receiver.  My initial uses are for a single receiver, but it should work equally well for multiple columns.  I guess it may not be essential, but I suggest we restrict these to be factor columns.
2) Call the second receiver **Carry Variables**.  They will often also be factor columns, but latitude and Longitude, etc may be numeric.  We should check it also will work for Character columns.  I think so.

From input by Danny the simplest route seems to be to adapt the same sort of code used in the Summary dialogue, using functions **First** and **Last** that are also part of dplyr, and useful generally.  That's the reason for #4309, which I hope is quick and can be added soon, to the calculator and summary system.  Then we return to the code for this dialogue.

maxwellfundi:
Great am on it. Will have to discuss this with then to see how to implement that function here.

maxwellfundi:
This is no longer needed at all as it will be done on the current unstack dialog

Issue #3396, opened 2017-06-15
Rename Columns improvements


dannyparsons:
Data Frame > Rename columns. This currently is only for one name at a time, and that is probably a good default. Ideally it would also be possible a) to make a sequence of names and b) to have a grid where the existing names are in one column and the new names can be typed into the second. An alternative (is it possible?) is to type directly into the column header field?

previously here #1781

dannyparsons:
b) not possible because of reogrid.
We can now type column names into the column metadata grids though

rdstern:
I like the way the column names can be typed into the column metadata grid.  I suggest this is sufficient for now.  Later, perhaps keeping the issue live, there is a whole issue of selecting names of columns, i.e. Find, etc which links also to the need for filters to be able to choose a subset of columns as well as rows.

rdstern:
A small improvement that should be very easy to do is as follows:
Please compare  Rename column (from right click) at the top
With Rename from the bottom right click, which is for the data frame.

They look slightly different - which is one problem.

But the particular small point is that the **rename data frame** suggests a new name which is the same as the old one.  That's nice, because I usually want a small edit of the existing name.

In contrast the rename column suggests the last name - which is dangerous and confusing and has never been what I wanted.  This should be easy to fix, so I have made it 0.4.10.

Patowhiz:
I'm addressing this

Patowhiz:
@rdstern what about the column label. Do you think it will be a plus if it could show the label name of the selected column i.e when a user selects the column to rename, the receiver is filled with column name from the selector and the label is filled with label of the selected column.
Though this functionality is already addressed in the column metadata grid.

rdstern:
I agree and it would be good for the 2 dialogues (rename column and rename data frame to look similar.  They can't be the same, because the data frame one doesn't have the same selector.

Patowhiz:
@dannyparsons , @shadrackkibet was proposing that for dialog Column Rename, when a user selects a column in the DataView and goes to the dialog through the **Prepare > DataFrame > Rename Column**, then the dialog should know the column selected and should automatically put it as the selected variable. Should this functionality be added?

That calls for minor modification of `ucrDataView `access of its subroutine, in particular the `SelectedColumnsAsArray()` from private to public

Patowhiz:
[...]

PR #4299, opened 2017-12-06
Column rename

Patowhiz:
@africanmathsinitiative/developers this fixes issue #3396 Rename Columns improvements


Patowhiz:

shadrackkibet:
Reviewing and testing this.

instat/dlgName.vb
```diff
@@ -119,8 +121,34 @@ Public Class dlgName
...
+            End If
+
+            'if the user has not typed anything then change the ucrInputVariableLabel contents
+            If Not ucrInputVariableLabel.bUserTyped Then
+                'get the label of the column selected(from ucrReceiverName) and set it as ucrInputVariableLabel value
+                Dim colmnLabelsRFunction = New RFunction
+                Dim expItems As SymbolicExpression
+
+                colmnLabelsRFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_column_labels")
+                colmnLabelsRFunction.AddParameter("data_name", Chr(34) & ucrSelectVariables.strCurrentDataFrame & Chr(34))
```
shadrackkibet: parameter position?

instat/dlgName.vb
```diff
@@ -119,8 +121,34 @@ Public Class dlgName
...
+                colmnLabelsRFunction.AddParameter("columns", ucrReceiverName.GetVariableNames(bWithQuotes:=True))
```
shadrackkibet: Also here?

instat/dlgName.vb
```diff
@@ -56,39 +62,33 @@ Public Class dlgName
...
-        End If
     End Sub
 
     Private Sub TestOKEnabled()
-        If Not ucrReceiverName.IsEmpty() AndAlso Not ucrInputNewName.IsEmpty() Then
-            ucrBase.OKEnabled(True)
-        Else
+        If ucrReceiverName.IsEmpty() AndAlso ucrInputNewName.IsEmpty() Then
             ucrBase.OKEnabled(False)
+        Else
```
shadrackkibet: I think you interchanged you testOk here.

instat/dlgName.vb
```diff
@@ -24,21 +24,27 @@ Public Class dlgName
...
+
+        'always reset the selector to reflect any column name change. 
+        'Reset the ucrInputs too because their inputs are also determined by ucrSelectVariables selection
+        ucrSelectVariables.Reset()
+        ucrInputNewName.Reset()
+        ucrInputVariableLabel.Reset()
+
+        ucrInputNewName.SetName("")
+        ucrInputVariableLabel.SetName("")
+
```
shadrackkibet: I think you should not reset this controls here. This is bringing problem when you reopen the dialog and the column is  still selected.

instat/dlgName.vb
```diff
@@ -56,39 +62,33 @@ Public Class dlgName
...
     Public Sub SetDefaults()
         clsDefaultRFunction = New RFunction
-        ucrInputVariableLabel.Reset()
+
         ucrSelectVariables.Reset()
         ucrInputNewName.Reset()
+        ucrInputVariableLabel.Reset()
+
+        ucrInputNewName.SetName("")
+        ucrInputVariableLabel.SetName("")
```
shadrackkibet: I think you should not reset this controls here.This is bringing an issue when you click reset and in the dataframe the column is still still selected.

instat/dlgName.vb
```diff
@@ -24,21 +24,27 @@ Public Class dlgName
     Private clsDefaultRFunction As New RFunction
 
     Private Sub dlgName_Load(sender As Object, e As EventArgs) Handles Me.Load
-        autoTranslate(Me)
```
shadrackkibet: This should be here, to be consistent with other dialogs.

instat/dlgName.vb
```diff
@@ -24,21 +24,27 @@ Public Class dlgName
...
 
     Private Sub dlgName_Load(sender As Object, e As EventArgs) Handles Me.Load
-        autoTranslate(Me)
         If bFirstLoad Then
             InitialiseDialog()
-            bFirstLoad = False
-        Else
-            ReopenDialog()
-        End If
-        If bReset Then
```
shadrackkibet: Not sure why you removed this.

instat/dlgName.vb
Patowhiz: In this case, I didn't see the need. Check on how parameters are sorted in the RunScript(). It will be neatly arranged to the way I have added them.

instat/dlgName.vb
Patowhiz: Because its purpose was to execute `ucrSelectVariables.Reset()` every time the dialog is loaded. There is no need for this.

instat/dlgName.vb
Patowhiz: @shadrackkibet  Except for `ucrSelectVariables.Reset()` , the others are a necessary. The blank problem that you were referring to is necessary for this form. Remember single receivers don't have the functionality to know that the column selected changed its name.

Patowhiz:
@africanmathsinitiative/developers this is ready for reveiw

instat/dlgName.vb
shadrackkibet: Do you mean we don't need parameter positions here?

Patowhiz: Well technically I think it's not necessary to have the parameters. Because of how the function is written in R and how the parameters are sorted before the R syntax is constructed(as long as you add them in the correct order in the way the function is written). Nevertheless to make the code consistent and clear to others I have added the positions in my commit.

instat/dlgName.vb
```diff
@@ -131,8 +131,8 @@ Public Class dlgName
                 Dim expItems As SymbolicExpression
 
                 colmnLabelsRFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_column_labels")
```
shadrackkibet: I suggest you define this function at the top and then SetRcommand in setdefaults, then you can add parameters here.

Patowhiz: I wanted to compartmentalise the functionality so that it can be clear when looking at the code in future. This was an afterthought functionality and its specific function is to retrieve the label of the column label only when a column has been selected and the user has not typed into the textbox.  To add on that putting the RCommand on `SetDefault()` supersedes its purpose, the RFunction does not affect any control, ONLY its output is used. I have put this functionality in another function `GetColLabel()` to make it much more elaborate.

shadrackkibet:

instat/dlgName.vb
dannyparsons: We do want to keep bReset and bFirstLoad separate. We're not making use of this now but in future we may want to be able to open a dialog as if it had been reset. So we could set bReset = True before opening, and then SetDefaults would be run so it resets. But we don't need the initialise stuff to run so bFirstLoad could still be False.
So the reopen we might not need but the SetDefaults in Reset we should keep.

instat/dlgName.vb
dannyparsons: When you link the controls to the R code in SetRCode then the correct text should be set for these controls. If that's not happening then there's something wrong in the way it's linking up. But yes, there should be no need to clear the text here. It may be hiding a problem elsewhere in the code.

instat/dlgName.vb
```diff
@@ -118,7 +116,47 @@ Public Class dlgName
         TestOKEnabled()
     End Sub
 
-    Private Sub ucrReceiverName_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverName.ControlValueChanged
-        DefaultNewName()
+    Private Sub ucrReceiverName_ControlValueChanged() Handles ucrReceiverName.ControlValueChanged
+        'if the receiver is not empty
+        If Not ucrReceiverName.IsEmpty Then
+            'if the user has not typed anything then change the ucrInputNewName contents
+            If Not ucrInputNewName.bUserTyped Then
```
dannyparsons: I know this is how the dialog is currently working but I think this actually causes confusion. I think even if the user has typed, once they select a new column they are likely to not want the old name they typed in. So I think we should set the name even if bUsertyped. And the same for the label.
Generally we prefer not to change things automatically but in this case I think not changing causes more confusion because users often forget to update this.

instat/dlgName.vb
```diff
@@ -118,7 +116,47 @@ Public Class dlgName
...
+            ucrInputVariableLabel.SetName("")
+        End If
+
     End Sub
-End Class
+
+    'Gets the label of the column selected(from ucrReceiverName) 
+    Private Function GetColLabel()
+        Dim strColLabel As String = ""
+        Dim colmnLabelsRFunction = New RFunction
```
dannyparsons: `cls` should be the prefix for this variable

instat/dlgName.vb
```diff
@@ -118,7 +116,47 @@ Public Class dlgName
...
+        Dim expItems As SymbolicExpression
+
+        colmnLabelsRFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_column_labels")
+        colmnLabelsRFunction.AddParameter("data_name", Chr(34) & ucrSelectVariables.strCurrentDataFrame & Chr(34), iPosition:=0)
+        colmnLabelsRFunction.AddParameter("columns", ucrReceiverName.GetVariableNames(bWithQuotes:=True), iPosition:=1)
+
+        expItems = frmMain.clsRLink.RunInternalScriptGetValue(colmnLabelsRFunction.ToScript(), bSilent:=True)
+
+        If expItems IsNot Nothing AndAlso Not (expItems.Type = Internals.SymbolicExpressionType.Null) Then
+            For Each strLabel As String In expItems.AsCharacter.ToArray
```
dannyparsons: If there's more than 1 item here then something's gone wrong. If you want, you can add a developer error if this array has more than length 1, but anyway you can just take the first item in the list.
Either way a for loop doesn't make sense because strColLabel is always taking the last value of the array.
We should check that there is at least 1 item in the list though.

Patowhiz: @dannyparsons its the checks that I wanted to avoid in the first place. The expected size/length of the array returned is 1 but the loop acts as an empty/length check and at the sametime it achieves my intention. I know rarely will there be a scenario where a column has more than one label , but if such a "rare incidence" occurs then just pick the **last label** for that column since we can only display one column in this dialog. I thought that will be the desired user expectation if things "go south" . Getting the first element means introducing checks which means introducing  extra statements. But if the explicit intention is to notify the user of an error in that "rare incidence" then I can Introduce the length check.

instat/dlgName.vb
Patowhiz: Done

instat/dlgName.vb
Patowhiz: Done. Removing this functionality makes the code even much simpler

instat/dlgName.vb
Patowhiz: Great. I had never thought of such functionality in such a manner. Its good to know why this has been done this way in all the other dialogs.

Patowhiz: Done. Though out of curiosity do we need to set  `bFirstLoad = False` after the `SetDefaults()` ?
I still don't see why not setting it to false will tamper with not  "initialising stuff"

instat/dlgName.vb
dannyparsons: We should be doing a check on the length, because if it's anything other than length 1 then something's gone wrong so we should know about, probably as a developer error.
I don't particularly mind if we decide to take the first or last but using a For loop when we only expect one item is confusing code so that was my main issue. There's no problem with adding an extra check.

instat/dlgName.vb
dannyparsons: No I don't think it matters whether `bFirstLoad = False` comes before or after `SetDefaults()` at the moment as long as its within the `If`. I feel it makes sense to come at the end of the If just in case we did have anything before that needs to know whether it was first load or not, but I don't think we actually have anything like this at the moment.

instat/dlgName.vb
Patowhiz: Okay

Patowhiz: @dannyparsons  this is done.

instat/dlgName.vb
```diff
@@ -118,7 +114,48 @@ Public Class dlgName
...
+        If Not ucrReceiverName.IsEmpty Then
+            'if the user has not typed anything then change the ucrInputNewName contents
+            ucrInputNewName.SetName(ucrReceiverName.GetVariableNames(bWithQuotes:=False))
+            'get the label of the column selected(from ucrReceiverName) and set it as ucrInputVariableLabel value
+            ucrInputVariableLabel.SetName(GetColLabel())
+
+        Else
+            'if the receiver is empty. These 2 controls should be empty
+            ucrInputNewName.SetName("")
+            ucrInputVariableLabel.SetName("")
```
dannyparsons: Not sure if these should be cleared or left as they are when the receiver is empty, but let's go with this and see what people think when they use it.

instat/dlgName.vb
dannyparsons: These shouldn't be needed, I hope.

instat/dlgName.vb
```diff
@@ -27,18 +29,18 @@ Public Class dlgName
         autoTranslate(Me)
         If bFirstLoad Then
             InitialiseDialog()
+            SetDefaults()
             bFirstLoad = False
-        Else
-            ReopenDialog()
         End If
-        If bReset Then
+        If bFirstLoad Then
```
dannyparsons: This part still doesn't look right. You've got SetDefaults twice on first load and you're using bFirstLoad instead of bReset here

Patowhiz:
That's correct. Changed that.

instat/dlgName.vb
```diff
@@ -32,7 +32,7 @@ Public Class dlgName
             SetDefaults()
```
dannyparsons: And SetDefaults() shouldn't be in `bFirstLoad`.

Patowhiz: Done.

dannyparsons: It's still here?

Patowhiz: I have uploaded the change.

instat/dlgName.vb
dannyparsons: What about this?

Patowhiz: I have removed the two unnecessary statements
```
 ucrInputNewName.SetName("")
 ucrInputVariableLabel.SetName("")
```

Patowhiz:
Changes submitted

PR #4300, opened 2017-12-06
Display daily data

shadrackkibet:
In progress!


shadrackkibet:
@rdstern  this is how the table part looks now. This function requires your data to  have stations for it to work. So station is not optional.
image

dannyparsons:
The R code will need to be adapted a bit before it will work for us.
Stations can't be required because we have data for 1 station which doesn't have a station column.
The function also assumes certain columns have specific names, which we don't want to require.
I'm less sure it's as time saving to use this now as we initially thought, @rdstern @volloholic?

shadrackkibet:
Here is the sample output from Hellen's function in R-Instat: However, it takes forever when you have many stations so you have to filter you data to few stations to get the output.

image

image

image

rdstern:
1. I notice it isn't doing an output of the rain?  
2. I agree that when you have many stations and many years it will take a long time - and produce a lot of output that no one will read.  

I am however keen that we do get this facility to work initially.  Then we can also see whether there could be an option to:
a) Write to a file, or
b) Write to a data frame

When writing to the output window I would be happy if (for now) we restricted to a single station per use of the dialogue.  You could still leave it as a multiple receiver for now, but perhaps add a field for write to file.  This could be disabled for now, but will remind us that this would be a useful option for the future.

That is what Helen does in her routines.  She writes lots of output files. 

This is a more general problem with the quality control procedures.  I am puzzling also on what to do in the QC when we produce graphs, because the equivalent graphical output would entail many graphs.

dannyparsons:
Great that you got this working in R-Instat, I assume with a carefully constructed dataset as my comments below still stand about needing to generalise this code for our use. I think there should be a discussion on this before any further work.

> The R code will need to be adapted a bit before it will work for us.
> Stations can't be required because we have data for 1 station which doesn't have a station column.
> The function also assumes certain columns have specific names, which we don't want to require.
> I'm less sure it's as time saving to use this now as we initially thought, @rdstern @volloholic?

rdstern:
On further discussion (including David and Danny) this adaptation of Helen's function might be a fair bit of work.  I suggest, however, that it might still be useful because her functions are pretty well written and hence this will sort of typify the type of adaption of R-code that is needed, if it is not already part of an existing package.
This could therefore be useful to document as the changes are made.  The documentation could initially be within the code, but could later become an example of "proper" documentation, once the key concepts are identified.  Typically some things will be hard-wired, which we don't want, or data are used where we just want to reference the variable names, or columns have specially assumed names, which we could adapt to our "climatic system", and so on.
This will need close collaboration between Danny (or David) and someone from the Maseno team.  I suggest it be just one person, and later they might give a short workshop to the others.

shadrackkibet:
I am going to close this now since this changes has been merged from another branch.

PR #4301, opened 2017-12-06
sdg summary

shadrackkibet:
In progress.


shadrackkibet:
@africanmathsinitiative/developers  this is now ready for review and testing.

instat/dlgOneVariableSummarise.vb
```diff
@@ -126,7 +125,9 @@ Public Class dlgOneVariableSummarise
     Private Sub cmdSummaries_Click(sender As Object, e As EventArgs) Handles cmdSummaries.Click
         sdgSummaries.SetRFunction(clsSummariesList, clsInstatSummaryFunction, ucrSelectorOneVarSummarise, bResetSubdialog)
         bResetSubdialog = False
+        sdgSummaries.bEnableTab = False
         sdgSummaries.ShowDialog()
+        sdgSummaries.bEnableTab = True
         TestOKEnabled()
```
maxwellfundi: Was this the only place that needed to have the tab disabled? I thought there was other dialogs

shadrackkibet: This has been added elsewhere.

instat/dlgSummaryTables.vb
```diff
@@ -175,7 +174,9 @@ Public Class dlgSummaryTables
     Private Sub cmdSummaries_Click(sender As Object, e As EventArgs) Handles cmdSummaries.Click
         sdgSummaries.SetRFunction(clsSummariesList, clsDefaultFunction, ucrSelectorSummaryTables, bResetSubdialog)
         bResetSubdialog = False
+        sdgSummaries.bEnableTab = False
```
shadrackkibet: @maxwellfundi were you asking for this?

dannyparsons: can you put the name of the tab in this variable name. There's multiple tabs so it's not clear what this does.

shadrackkibet: okay.

instat/sdgSummaries.vb
```diff
@@ -16,14 +16,22 @@
...
+    Public strDataFrame As String
+    Public bEnable2VariableTab As Boolean = True
 
     Private Sub sdgDescribe_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         autoTranslate(Me)
+        If Not bEnable2VariableTab Then
+            tbTwoVariables.Enabled = False
+        Else
+            tbTwoVariables.Enabled = True
+        End If
```
dannyparsons: This can be much simpler: `tbTwoVariables.Enabled = bEnable2VariableTab`

dannyparsons:
Anything else to review on this?

instat/sdgSummaries.vb
shadrackkibet: Thanks, i will be changing this shortly.

rdstern:
Here are my initial comments on this new set of sub-dialogues.
1) Great to see these additions
1) The trimmed mean has one of those (damn) up-down controls.  I tried and failed to trim 0.05.  If we have to have this control then it should go in steps of 0.01.  But really nice anyway to be able to type into it - with then a check that it remains between 0 and 0.5.
2) It seems  to give the trimmed mean, but how do I get both.  I would usually like to know what difference the trimmed mean is from the ordinary mean.  If only 1 is possible, then put the trimming on the main sub-dialogue (as an option with the mean).
3) Trimmed mean again.  When it is trimmed, can the name indicate the degree of trimming?
4) Why are the first, last, etc and proportions, etc on the 2 variable tab?  I thought they would go on the "More" tab.  And with the proportions, etc at the top.  
5) The quantile(s).  I tried 0.2.  It seemed to work, but the name isn't clear. could it be q0.2_Yield say.  Then we could re-use the dialogue to get further quantiles.
6) More later

Issue #4275, opened 2017-11-29
Runtime Warnings


Patowhiz:
@africanmathsinitiative/developers  Everytime I run R-Instant in my visual studio I get all these warnings that I think are fixable. Are the unused variables left there intentionally? 

![warnings](https://user-images.githubusercontent.com/21179007/33403802-cea17ab6-d572-11e7-82d2-a7833b283fe6.JPG)

dannyparsons:
No. Unused variables can be removed. It's good practice to solve all the warnings so we should try to clear all of these.

Patowhiz:
Fixed some of those warnings above

Patowhiz:
This has been solved.

PR #4303, opened 2017-12-06
Runtime warnings

Patowhiz:
This solves issue  #4275 Runtime Warnings.
I have removed the unused variables and initialised the uninitialised ones.
I'm still not sure why there is a reference to 'System.Net.Http' . I haven't been able to locate where its being used, so I haven't removed it. However for this reference to work then the path to it needs to be set up to point to System.Net.Http.dll. 

@africanmathsinitiative/developers anyone who could quickly point to me where and why this reference was added to the project.


dannyparsons:
I remember solving that issue before and it was something to do with matching the reference with the correct .NET version. I'll try to find the solution again.

Patowhiz:
Yeah 4.5 and above have removed the System.Net.Http . To reference it you will have to manually add the  System.Net.Http.dll . But am still skeptical where this is being used in RInstant.

dannyparsons:
I believe it's required by other reference, but I'm not absolutely sure. It's easy to check - if you remove the reference and it's needed in the project, you will get an error.

Patowhiz:
@dannyparsons I removed it and there was no error in my project during runtime

PR #4307, opened 2017-12-07
dlgTheme

maxwellfundi:
We now have a designer for the themes dialog. 

image

@dannyparsons  @rdstern  I am wondering if these themes will be associated to a dataframe, now that we are using the usual selector. I would have guessed we could use the selector without the dataframe selector. Perhaps i might be wrong.


instat/dlgThemes.vb
```diff
@@ -17,21 +17,33 @@
...
+    End Sub
+
+    Private Sub SetRCodeForControls(bReset As Boolean)
+
+    End Sub
+
     Private Sub InitialiseDialog()
         ucrBase.iHelpTopicID = 440
+
+        ucrChkDeleteTheme.SetText("Delete Theme")
```
maxwellfundi: @dannyparsons  Now that we will be saving the themes on the instatobject as well as other places, do we need to use this ucrsave? if we need it then there is not method of savetype as theme which might be needed.

rdstern:
I agree that we don't want the selector with the data frames and (for now) I presume a theme is saved as part of an R-Instat Object.  Later we may (will) want to think of the themes (and perhaps some other objects, but certainly themes) separately, because we are likely to want to use them across Instat objects and perhaps send them to other people!

maxwellfundi:
Great, I will change that and start implementing the dialog.

rdstern:
After discussing with David, he agreed that the main themes sub-dialogue could usefully have the items I mentioned somewhere, namely display legend, horizontal and vertical grid-lines, formatting of axis labels and (of course) angle of x-axis labels.
So my understanding is that the main themes dialogue would then simply have the same themes tab as its sub-dialogue.
Then the harder part is how to store the user-defined themes, then make them available in the dialogue to edit further or delete.  Then also to add it to the list of themes to be used by all the plotting dialogues.  Exciting!

dannyparsons:
I would think deleting themes would be a separate dialog since it won't have any of the other options on this dialog. The list of deletable themes is also different from useable themes. And there may be other options about how and where you delete a theme. So I think we don't have the delete checkbox here.

The save control could say New Theme Name.

Instead of Theme to Edit I think it could be Base Theme, because you're not really editing the theme, but creating a new theme with some theme as the base. I think having a base theme is also optional and you could build a completely new theme from scratch so then the base theme selection is optional.

instat/dlgThemes.vb
maxwellfundi: I think now that we need to save these themes we could have some code done for saving the themes.

maxwellfundi:
@dannyparsons  this had just the design and has been lying here for so long, I think we can merge the current state as it is. The discussion about how this works would be done in the future. 

The menu is disabled and therefore this dialog would never show.

Issue #4297, opened 2017-12-05
Missing word


ZIRANA11:
On the Histogram dialog, On plot options, when **Vertical** is selected the word **Columns** is missing on the statement **Fixed number of**
![column](https://user-images.githubusercontent.com/28622260/33611222-a0f19694-d9de-11e7-89ba-8eebf9e909cb.PNG)

dannyparsons:
@africanmathsinitiative/developers a trivial change for someone to make please

shadrackkibet:
will make this  change.

shadrackkibet:
@maxwellfundi  i see you have a pullrequest working on this file. Could you make this change?

maxwellfundi:
yeah i have been working on this. I think i will have to change it before
it causes conflicts

Regards
Maxwell Fundi
|| Phone :+254728508241||

PR #4311, opened 2017-12-09
Adding coordflip to sdgplots

### LINKED ISSUE #4251: How do I do coord_flip?

maxwellfundi:

Fixes #4251 and #4297

### DIFF
--- a/instat/dlgBarAndPieChart.vb
+++ b/instat/dlgBarAndPieChart.vb
@@ -192,11 +192,7 @@ Public Class dlgBarAndPieChart
         'Warning, when coordinate flip is added to coordinates tab on sdgPLots, then link with ucrChkFlipCoordinates...
 
         'this syncs the coordflip in sdgplots and this main dlg
-        If clsBaseOperator.clsParameters.Contains(clsBaseOperator.GetParameter("coord_flip")) Then
-            ucrChkFlipCoordinates.Checked = True
-        Else
-            ucrChkFlipCoordinates.Checked = False
-        End If
+        ucrChkFlipCoordinates.SetRCode(clsBaseOperator, bReset)
     End Sub
 
     Private Sub cmdBarChartOptions_Click(sender As Object, e As EventArgs) Handles cmdBarChartOptions.Click
--- a/instat/dlgBarAndPieChart.vb
+++ b/instat/dlgBarAndPieChart.vb
@@ -190,6 +190,13 @@ Public Class dlgBarAndPieChart
         sdgPlots.ShowDialog()
         bResetSubdialog = False
         'Warning, when coordinate flip is added to coordinates tab on sdgPLots, then link with ucrChkFlipCoordinates...
+
+        'this syncs the coordflip in sdgplots and this main dlg
+        If clsBaseOperator.clsParameters.Contains(clsBaseOperator.GetParameter("coord_flip")) Then
+            ucrChkFlipCoordinates.Checked = True
+        Else
+            ucrChkFlipCoordinates.Checked = False
+        End If
     End Sub
 
     Private Sub cmdBarChartOptions_Click(sender As Object, e As EventArgs) Handles cmdBarChartOptions.Click
--- a/instat/dlgBoxPlot.vb
+++ b/instat/dlgBoxPlot.vb
@@ -203,6 +203,14 @@ Public Class dlgBoxplot
         sdgPlots.SetRCode(clsBaseOperator, clsNewThemeFunction:=clsThemeFunction, dctNewThemeFunctions:=dctThemeFunctions, clsNewGlobalAesFunction:=clsRaesFunction, clsNewXScalecontinuousFunction:=clsXScaleContinuousFunction, clsNewYScalecontinuousFunction:=clsYScaleContinuousFunction, clsNewXLabsTitleFunction:=clsXlabsFunction, clsNewYLabTitleFunction:=clsYlabFunction, clsNewLabsFunction:=clsLabsFunction, clsNewFacetFunction:=clsRFacetFunction, ucrNewBaseSelector:=ucrSelectorBoxPlot, bReset:=bResetSubdialog)
         sdgPlots.ShowDialog()
         bResetSubdialog = False
+
+        'this yncs the coordflip in sdgplots and the main dlg
+        If clsBaseOperator.clsParameters.Contains(clsBaseOperator.GetParameter("coord_flip")) Then
+            ucrChkHorizontalBoxplot.Checked = True
+        Else
+            ucrChkHorizontalBoxplot.Checked = False
+        End If
+
     End Sub
 
     Private Sub cmdBoxPlotOptions_Click(sender As Object, e As EventArgs) Handles cmdBoxPlotOptions.Click
@@ -236,13 +244,6 @@ Public Class dlgBoxplot
                 ucrSecondFactorReceiver.Add(clsParam.strArgumentValue)
             End If
         Next
-
-        'this yncs the coordflip in sdgplots and the main dlg
-        If clsBaseOperator.GetParameter("coord_flip").strArgumentName = "coord_flip" Then
-            ucrChkHorizontalBoxplot.Checked = True
-        Else
-            ucrChkHorizontalBoxplot.Checked = False
-        End If
 
         'Question to be discussed: After running through the sdgLayerOptions, the clsCurrDataFrame parameters seem to have been cleared, such that in the multiple variable case, clsCurrDataFrame needs to be repopulated with "stack", "measure.vars" and "id.vars" parameters. Actually, even when repopulated, they are still not appearing in the script. ??
         'This resets the factor receiver and causes it to be cleared of the correct variable. We don't want this.
--- a/instat/dlgBoxPlot.vb
+++ b/instat/dlgBoxPlot.vb
@@ -204,7 +204,7 @@ Public Class dlgBoxplot
         sdgPlots.ShowDialog()
         bResetSubdialog = False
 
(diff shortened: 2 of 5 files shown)


shadrackkibet:
Reviewing and testing this.

instat/dlgBoxPlot.vb
```diff
@@ -236,6 +236,14 @@ Public Class dlgBoxplot
                 ucrSecondFactorReceiver.Add(clsParam.strArgumentValue)
             End If
         Next
+
+        'this yncs the coordflip in sdgplots and the main dlg
```
shadrackkibet: tiny edit for the comment here.

shadrackkibet:
@dannyparsons  this is working fine. This is ready for you to review.

@rdstern  Why do we have `coorflip` twice in  Boxplot and Bar chart?. We have this in the main dialog and in the subdialog. I thought we should have this in one place.

instat/dlgBoxPlot.vb
shadrackkibet: i see you edited this.

rdstern:
My main concern was that it should at least be in one place for all plots.  Before the changes that maxwell made it was on the main dialogue for boxplots and bar charts and nowhere for the others.
Now it is on the sub-dialogue for all charts that can access this - which is fine.  

The main dialogue is (of course) easier to access than the sub-dialogues.  So I am happy for it to be there too, on those charts where it is most useful.  That is until that main dialogue gets too cluttered.  And I was also wanting to have minimal change, so we proceed to fix lots of things.

maxwellfundi:
I now have these in two places and are in sync and I actually like it when they are in both places.

dannyparsons:
@shadrackkibet have you tested changing the coord flip on the main dialog and then the sub dialog to make sure that they update each other on those dialogs which have it on both?

shadrackkibet:
yep i checked that. I noticed that when you check the one in the main dialog it automatically checks in the subdialog which updates fine in the code.

dannyparsons:
and the opposite? when you check on the subdialog does it update the main dialog?

shadrackkibet:
yep it updates in the main dialog as well.

instat/dlgBarAndPieChart.vb
```diff
@@ -190,6 +190,13 @@ Public Class dlgBarAndPieChart
         sdgPlots.ShowDialog()
         bResetSubdialog = False
         'Warning, when coordinate flip is added to coordinates tab on sdgPLots, then link with ucrChkFlipCoordinates...
+
+        'this syncs the coordflip in sdgplots and this main dlg
+        If clsBaseOperator.clsParameters.Contains(clsBaseOperator.GetParameter("coord_flip")) Then
+            ucrChkFlipCoordinates.Checked = True
+        Else
+            ucrChkFlipCoordinates.Checked = False
+        End If
```
dannyparsons: This is better done by setting the R code for ucrChkFlipCoordinates, (which basically does this in SetRCode) but in general setting R code is how we sync controls with code, instead of manually doing what is implemented in the core system already.

maxwellfundi: What you mean is i can do setrcode here instead of doing this?

instat/sdgPlots.vb
```diff
@@ -193,11 +196,19 @@ Public Class sdgPlots
...
         ucrInputThemes.SetDropDownStyleAsNonEditable()
 
+        'coordiantes tab
+        ucrChkHorizontalplot.SetText("Horizontal Plot (coord-flip)")
+        clsCoordFlipFunc.SetPackageName("ggplot2")
+        clsCoordFlipFunc.SetRCommand("coord_flip")
+        clsCoordFlipParam.SetArgumentName("coord_flip")
+        clsCoordFlipParam.SetArgument(clsCoordFlipFunc)
+        ucrChkHorizontalplot.SetParameter(clsCoordFlipParam, bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)
+        ucrChkHorizontalplot.AddParameterPresentCondition(True, "coord_flip", True)
```
dannyparsons: I think this condition is already set in SetParameter - have a check. Anyway, if you are adding this condition here then you should have added the False condition as well. So there should be 0 or 2 conditions set here.

maxwellfundi: Okay. Sounds good.

instat/sdgPlots.vb
```diff
@@ -58,6 +58,9 @@ Public Class sdgPlots
         Dim dctThemes As New Dictionary(Of String, String)
         Dim strThemes As String()
 
+        Dim clsCoordFlipFunc As New RFunction
+        Dim clsCoordFlipParam As New RParameter
```
dannyparsons: I'm not sure how this works. These two are defined in this sub, so after the sub is run they no longer exist? They are not attached to any other piece of code so I don't see how this is doing anything?

instat/dlgBarAndPieChart.vb
dannyparsons: Yes

dannyparsons:
To properly test the syncing you need to test this from multiple dialogs i.e. set coord flip to be different from different main dialogs and see if they keep correct value when you go back to each dialog.

instat/sdgPlots.vb
maxwellfundi: We have used this before in other dialogs which are run on the initialize dialog so I dont see how it would work differently here.

dannyparsons: The issue I was concerned about is that these could also be defined on a main dialog and passed through. So there could be some overwriting of this parameter happening here. A sub dialog is very different from a main dialog in this way, so you can't just assume things will work in the same way without thinking about them.
But I think this is actually ok here because the coord_flip will be the same function as coming from the main dialog, so it's ok if one overwrites the other. So I don't think there is an issue here anymore, I thought there might have been because this is on a sub dialog.

instat/dlgBarAndPieChart.vb
maxwellfundi: Okay

instat/sdgPlots.vb
maxwellfundi: No need to add that actually its added in the setParameter for the ucrChk

dannyparsons: great, I thought that happened in there

maxwellfundi:
@dannyparsons  this now works as expected. I have been testing it with different dialogs and resting stuff and it syncs properly.

instat/sdgPlots.vb
```diff
@@ -193,11 +196,19 @@ Public Class sdgPlots
...
+        clsCoordFlipParam.SetArgumentName("coord_flip")
+        clsCoordFlipParam.SetArgument(clsCoordFlipFunc)
+        ucrChkHorizontalplot.SetParameter(clsCoordFlipParam, bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)
+        ucrChkHorizontalplot.AddParameterPresentCondition(True, "coord_flip", True)
         InitialiseTabs()
 
         'temporary disabled until implemented
         ' tbpLayers.Enabled = False
-        tbpCoordinates.Enabled = False
+        tbpCoordinates.Enabled = True
```
dannyparsons: I assume you can remove this line, since it's enabled by default?

instat/sdgPlots.vb
```diff
@@ -664,4 +678,24 @@ Public Class sdgPlots
...
         End If
     End Sub
+
+    Private Sub ucrChkIncludeFacets_CheckedChanged(ucrChangedControl As ucrCore) Handles ucrChkIncludeFacets.ControlValueChanged, ucr2ndFactorReceiver.ControlValueChanged, ucr1stFactorReceiver.ControlValueChanged
+
+    End Sub
+
+    Private Sub ucrChkFreeSpace_CheckedChanged(ucrChangedControl As ucrCore) Handles ucrChkFreeSpace.ControlValueChanged
+
+    End Sub
```
dannyparsons: these shouldn't be here

instat/sdgPlots.vb
```diff
@@ -664,4 +678,24 @@ Public Class sdgPlots
...
+
+    Private Sub ucrPnlHorizonatalVertical_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlHorizonatalVertical.ControlValueChanged, ucrChkMargin.ControlValueChanged
+
+    End Sub
+
+    Private Sub LabsControls_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputGraphTitle.ControlValueChanged, ucrInputGraphSubTitle.ControlValueChanged, ucrInputGraphCaption.ControlValueChanged
+
+    End Sub
```
dannyparsons: or any of these

instat/sdgPlots.vb
maxwellfundi: No, I have checked and for some reason that tab is disabled unless i have this line then it wont be enables. So i will leave it there. I have removed the rest though.

dannyparsons: The tab may be disabled in design view then. Not a big deal but better to enable in design view and disable in the code if needed to avoid future confusion.

PR #4318, opened 2017-12-11
dlgColumnstats

### LINKED ISSUE #4302: Prepare > Column: Reshape > Column Summaries
We wanted to use the column summaries to get summary statistics (median) of 2 columns into a dataframe.  It would only have one row.

But OK is not enabled unless there are some factor columns.  I suppose we could have had a factor with a single level - but it could be enabled anyway!

shadrackkibet:
@africanmathsinitiative/developers  this is ready for review. Fixes #4302

### DIFF
--- a/instat/dlgColumnStats.vb
+++ b/instat/dlgColumnStats.vb
@@ -128,7 +128,7 @@ Public Class dlgColumnStats
     End Sub
 
     Public Sub TestOKEnabled()
-        If ((ucrChkStoreResults.Checked OrElse ucrChkPrintOutput.Checked) AndAlso Not ucrReceiverSelectedVariables.IsEmpty AndAlso Not clsSummariesList.clsParameters.Count = 0) Then
+        If ((ucrChkStoreResults.Checked OrElse ucrChkPrintOutput.Checked) AndAlso Not clsSummariesList.clsParameters.Count = 0) Then
             ucrBase.OKEnabled(True)
         Else
             ucrBase.OKEnabled(False)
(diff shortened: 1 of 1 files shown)


instat/dlgColumnStats.vb
```diff
@@ -128,7 +128,7 @@ Public Class dlgColumnStats
     End Sub
 
     Public Sub TestOKEnabled()
-        If Not ucrReceiverByFactor.IsEmpty AndAlso (ucrChkStoreResults.Checked OrElse ucrChkPrintOutput.Checked) AndAlso Not clsSummariesList.clsParameters.Count = 0 Then
+        If ((ucrChkStoreResults.Checked OrElse ucrChkPrintOutput.Checked) AndAlso Not ucrReceiverSelectedVariables.IsEmpty AndAlso Not clsSummariesList.clsParameters.Count = 0) Then
```
dannyparsons: You don't have to have variables selected either, because if there are no variables to summarise it can still do the counts within each group.

shadrackkibet: ok. I will remove this receiver from testok.

dannyparsons:
Did you check that it does work without any factors or variables?

shadrackkibet:
yep, it works fine.  when you have factors only it produces counts. when you have variables it produces other summaries.

dannyparsons:
great, and when you have neither?

shadrackkibet:
it  gives you the number of rows in your dataframe.

dannyparsons:
ok, yes that's the count by the factors, but since there's no factors you just get one count.

PR #4319, opened 2017-12-11
dlgParallelCoordinatePlot

### LINKED ISSUE #4253: Adding parallel coordinate plots
This is a function from the **ggally** package.  It should be relatively easy to add and could be extremely useful.  It would go into the specific graphs under the current mosaic plots menu item.  I have put Maxwell down, because it is graphics, but on the assumption he may suggest who could do this.

I will explain more when I visit and show a variety of examples.  It is a different sort of graph and is designed to look at lots of variables (columns) at the same time.  

It returns a ggplot object so will fit nicely into our system.  I think it should be easy to do.  I consider the main dialogue first.

0. The title is Parallel Coordinate Plot.
1. Left hand side is as for all the other plotting dialogues.  Our usual selector and the Two buttons for the sub-dialogues (Parcoord Options, Plot Options) below that.
2. On the right-hand side is a multiple receiver with the label X Variables (note the s on Variables).  Let's start with it just for our set of numeric columns. 
3. You may be surprised there is no Y variable - but read on!
4. Below that is a single receiver with our usual label Factor (Optional).  (Just factor columns for this receiver.
5. On the lhs at the bottom (as for all the other graph dialogues) is our usual **Save Graph** and then the comment plus 5 button control.
6.  Below the 2 buttons is a Scale label.  This is a drop down with 6 possibilities - default is std.  The others are robust, uniminmax, globalminmax, center, centerObs.  Later the last 2 will have to open a further control.
7. Now I am not sure.  I would like to add the following, but I don't want the main dialogue to get cluttered.  If you do add all the items below, then some could perhaps go on the right hand side of the dialogue.  Or they could go on the sub-dialogue.  Let's see.
a. Checkbox with the label **Boxplots**.  Default is unchecked.
b) Checkbox with the label **Points**.  Default is unchecked.
c) A label **Transparency**. Then our **up-down** from 0 to 1 in steps of 0.01.  Default is 1.  Or a text field that checks 0 to 1 that is typed.
d) The coord-flip, as for boxplots, with same label **Horizontal Plot**.
e) Possibly the Legend option if that is going on all main dialogues.

Now the sub-dialogues.  The Plot options connects to our usual sub-dialogue. 
The Parcoord Options buttons opens a special sub-dialogue, because this is a function and not a geom.

This might have some of the items above, if the main dialogue is looking cluttered.
The 2 main items are Missing and Order

Checkbox for **Missing**.  If activated there is a label for a drop down with a set of options on how to cope with missing values.  Default is Exclude and is the first.  The others are mean, median, min10, random.
Checkbox for **Order** is the order of the columns in the display.  This is a pull-down with 5 options: Namely As Given, AnyClass, AllClass, Skewness and Outlier.

maxwellfundi:
Fixes  #4253


maxwellfundi:
@africanmathsinitiative/developers  This is ready for review. Check that it works well with both the mainplots subdialog as well as its own subdialog. I need a third eye to make sure the linking is working properly. 

Thanks.

maxwellfundi:
@Muthenya  I have changed this to now work properly with operators to make it possible work with plots subdialog. You might want to have a look at the changes i made.

shadrackkibet:

shadrackkibet:
**Some few designer issues to note:**

1. You forgot your colons.
2. The parcoord options sdg designer is not in good shape. Labels and far from their respective controls, and the base is not centered properly.
3. In the main dialog the scale label is not well aligned with it's control.


- I think we should always have more than one element in the x-variable receiver , otherwise this will always bug since we cannot get parallel coordinate plot for one element.
- When you have your elements in x-variable receiver and you don't have a Factor then your order is anyClass in the your sdg then you get a bug. I think you should change the default to be skewness?
- I click on plot options subdialog and then on include facets i try to check this checkbox then it crashes.So i am unable to do anything in this subdialog.

instat/sdgParallelPlots.vb
```diff
@@ -26,13 +26,15 @@ Public Class sdgParallelPlots
         Dim dctMissing As New Dictionary(Of String, String)
         Dim dctOrder As New Dictionary(Of String, String)
 
+        ucrInputMissing.SetParameter(New RParameter("missing"))
```
shadrackkibet: parameter position.

instat/sdgParallelPlots.vb
```diff
@@ -26,13 +26,15 @@ Public Class sdgParallelPlots
...
 
+        ucrInputMissing.SetParameter(New RParameter("missing"))
         dctMissing.Add("Exclude", Chr(34) & "exclude" & Chr(34))
         dctMissing.Add("Mean", Chr(34) & "mean" & Chr(34))
         dctMissing.Add("Median", Chr(34) & "median" & Chr(34))
         dctMissing.Add("min10", Chr(34) & "min10" & Chr(34))
         dctMissing.Add("Random", Chr(34) & "random" & Chr(34))
         ucrInputMissing.SetItems(dctMissing)
 
+        ucrInputOrder.SetParameter(New RParameter("order"))
```
shadrackkibet: also here.

maxwellfundi:
Thanks @shadrackkibet for the review. Addressing those issues soon

maxwellfundi:
Issues raised by reviewer addressed.

lilyclements:
Hola, I've got a bit of time off today so had a quick look at this - can I make some quick (probably mainly design) changes to this and merge it back into your branch?

maxwellfundi:
Holla @lilyclements  perfecto!  Por favor edítelo.

lilyclements:
Great thanks! It is only designer changes. I did notice that there isn't **always** parameter positions - is this allowed now or should there be parameter positions in?

Also `alphaLines = 1` is the default, however, this comes up in the R-Code in R-Instat. Is that intentional? Perhaps you've changed how we treat default values now or something so I didn't want to edit that in case!


Finally, are you able to get this to work for `AlphaLines < 1` for a larger data set - such as Diamonds? If I try on R-Instat it ran for about five minutes before I gave up and quit. (note it took quite a long time without altering `AlphaLines`, but once I did alter this nud it took even longer!)
On R, running Diamons with AlphaLines altered took 2 minutes 23 seconds (which is long for R) and then didn't give me the plot!

Another time when I ran it on R I got the message

`Warning message:
In strsplit(names, .grid.pathSep) : reached elapsed time limit`

I'm just wondering if other people are having this issue with changing this nud on their laptops with larger data sets or is my laptop being slow today? I'm running the following in R - how long does this take for you -

```
utils::data(package="ggplot2", X=diamonds)
diamonds <- diamonds
InstatDataObject$import_data(data_tables=list(diamonds=diamonds))
rm(diamonds)

# Code generated by the dialog, Parallel Coordinate Plot
diamonds <- InstatDataObject$get_data_frame(data_name="diamonds")
last_graph <- GGally::ggparcoord(data=diamonds, columns=c("carat","depth"), groupColumn="cut", boxplot=FALSE, showPoints=FALSE, alphaLines=1, missing="exclude", order="skewness") + theme_grey()
InstatDataObject$add_graph(graph_name="last_graph", graph=last_graph, data_name="diamonds")
InstatDataObject$get_graphs(data_name="diamonds", graph_name="last_graph")
```

Since it's a problem in R then it's not a problem in the programming so not fixable through the VB code (I assume!)

maxwellfundi:
@lilyclements  Very well done for having tested this brutally and its obvious that the bigger the dataset the longer it takes. 
@shadrackkibet says he does not get the issue.

lilyclements:
Okay interesting, so that was trying with `alphaLines` as 1. @shadrackkibet can you try

```
# Code generated by the dialog, Parallel Coordinate Plot
diamonds <- InstatDataObject$get_data_frame(data_name="diamonds")
last_graph <- GGally::ggparcoord(data=diamonds, columns=c("carat","depth"), groupColumn="cut", boxplot=FALSE, showPoints=FALSE, alphaLines=0.7, missing="exclude", order="skewness") + theme_grey()
InstatDataObject$add_graph(graph_name="last_graph", graph=last_graph, data_name="diamonds")
InstatDataObject$get_graphs(data_name="diamonds", graph_name="last_graph")
```

So here I've changed alphaLines to be less than 1. This took me over 2 minutes when I ran it, however, would be interesting to know if this is an issue with my laptop or with the R-Code.

maxwellfundi:
Maybe thats an issue with your laptop. Our laptops which i assume are not as fast as Thinkpad take less time. Am convinced its your laptop !

lilyclements:
Okay must be my laptop but just wanted to confirm - thanks! With `alphaLines` as 1 it ran fine on R (maybe a little slow) but less than 1 took over two minutes.

Will send up bfixes/param positions/etc shortly!

maxwellfundi:
Exelente!

Patowhiz:
LOL. I'm loving this discussion. benchmarking of applications execution time is always hectic since there is so many parameters involved.

maxwellfundi:
Yea, it can never be the same for two different machines

instat/dlgParallelCoordinatePlot.vb
```diff
@@ -32,23 +43,144 @@ Public Class dlgParallelCoordinatePlot
...
+        ucrReceiverFactor.SetParameter(New RParameter("groupColumn", 2))
+        ucrReceiverFactor.strSelectorHeading = "Factors"
+        ucrReceiverFactor.Selector = ucrSelectorParallelCoordinatePlot
+        ucrReceiverFactor.SetIncludedDataTypes({"factor"})
+        ucrReceiverFactor.SetParameterIsString()
+
+        ucrChkBoxplots.SetParameter(New RParameter("boxplot", 3))
+        ucrChkBoxplots.SetText("Boxplot")
+        ucrChkBoxplots.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
+        ucrChkBoxplots.SetRDefault("FALSE")
```
lilyclements: You've set the R default to be FALSE but when the R Code runs in R-Instat, it still runs `boxplot = FALSE`.
Do we know why this is out of interest?

maxwellfundi: @dannyparsons  this is interesting. I am not sure why this still shows false and its the R  Default. I would have expected it to remove it from the code. Else everything else works well now.

instat/dlgParallelCoordinatePlot.vb
```diff
@@ -32,23 +43,141 @@ Public Class dlgParallelCoordinatePlot
...
+        ucrReceiverXVariables.SetMeAsReceiver()
+        ucrSelectorParallelCoordinatePlot.Reset()
+        ucrSelectorParallelCoordinatePlot.SetGgplotFunction(clsBaseOperator)
+        ucrSaveGraph.Reset()
+        bResetSubdialog = True
+
+        clsGGParCoordFunc.SetPackageName("GGally")
+        clsGGParCoordFunc.SetRCommand("ggparcoord")
+
+        clsGGParCoordFunc.AddParameter("boxplot", "FALSE", iPosition:=11)
```
lilyclements: Since this is a default, this doesn't need to run here? At least, that was the case a few months ago! There are other defaults that are being run in this bit too like alphaLines = 1, etc

maxwellfundi: Sounds reasonable but how will I reset the already checked checkbox ? Also, how would i set R code for those controls?

lilyclements: Just checked and all of these `.AddParameter` bits are the R-Default except for the `order` parameter.

Additionally, since `missing = "exclude"` is in the sub dialog I can't remember how that default works and if that needs to be set here even if it's the R-Default.
You should be able to remove all these `.AddParameters` except definitely not the `order` parameter and perhaps not the `missing` one?

I can't remember the specifics for why this is.
They reset from your `.SetRCode` part so will reset without this `.AddParameter` part in here. @dannyparsons can you explain this? I can't remember.

This will hopefully fix the issue that you have with the R defaults being read into the r-code?

maxwellfundi: Yes, that would fix that issue actually. I actually agree with you, am going to remove the points and boxplot on the sedefaults

maxwellfundi:
@lilyclements  This should be good enough now to be merged in. I am going to merge it unless you have got another querry

lilyclements:
Does the AddParameter for alphalines =1 and something = "std" need to be in SetDefaults still? I thought these were the defaults for the function so wouldn't need to be?
On my phone on github so apologies for poor formatting

maxwellfundi:
Oo yeah, those could be removed too actually must have been an omission

PR #4240, opened 2017-11-23
Dialog backup manager

Patowhiz:
@africanmathsinitiative/developers this is ready for review.
Solution to issue #4184  Restore data from backup. This is a continuation of issue #4224 Backup Manager.

@shadrackkibet this has addressed all the issues that you have found except:

> 6.when i close R-Instat and then run it again.On coming back to the dialog i find my buckup file.When i click at it it doesn't load it in the data view window.

I'll address it once @dannyparsons  reverts back to me.

PR #4321, opened 2017-12-11
Dialog Backup Manager

Patowhiz:
Changes to the `clsRLink`, Subroutine `LoadInstatDataObjectFromFile` . 
@dannyparsons you can review the subroutine `LoadInstatDataObjectFromFile` .  I have commented out the original code statements so that you can compare the changes.

@africanmathsinitiative/developers we can have an interesting discussion on  `clsImportRDS.ToScript(strScript)` subroutine with @volloholic .

This is continuation of #4240

### DIFF
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -178,16 +178,34 @@ Public Class RLink
         Return RunInternalScript(clsLoadPackages.ToScript(), bSilent:=bSilent)
     End Function
 
-    Public Sub LoadInstatDataObjectFromFile(strFile As String, Optional strComment As String = "")
+    Public Sub LoadInstatDataObjectFromFile(strFile As String, Optional bKeepExisting As Boolean = False, Optional strComment As String = "")
+        ' Dim clsReadRDS As New RFunction
+        ' Dim strScript As String = ""
+        ' Dim strTemp As String = ""
+
+        ' clsReadRDS.SetRCommand("readRDS")
+        ' clsReadRDS.AddParameter("file", Chr(34) & strFile.Replace("\", "/") & Chr(34))
+        ' clsReadRDS.SetAssignTo(strInstatDataObject)
+        ' strTemp = clsReadRDS.ToScript(strScript)
+        ' RunScript(strScript, strComment:=strComment)
+        ' bInstatObjectExists = True
+
+        Dim clsImportRDS As New RFunction
         Dim clsReadRDS As New RFunction
         Dim strScript As String = ""
         Dim strTemp As String = ""
 
         clsReadRDS.SetRCommand("readRDS")
         clsReadRDS.AddParameter("file", Chr(34) & strFile.Replace("\", "/") & Chr(34))
-        clsReadRDS.SetAssignTo(strInstatDataObject)
-        strTemp = clsReadRDS.ToScript(strScript)
-        RunScript(strScript, strComment:=strComment)
+        clsReadRDS.SetAssignTo("new_RDS")
+
+        clsImportRDS.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$import_RDS")
+        clsImportRDS.AddParameter("data_RDS", clsRFunctionParameter:=clsReadRDS)
+        'This RFunction takes booleans in capitals hence ToUpper
+        clsImportRDS.AddParameter("keep_existing", bKeepExisting.ToString.ToUpper)
+
+        strTemp = clsImportRDS.ToScript(strScript)
+        RunScript(strScript & strTemp, strComment:=strComment)
         bInstatObjectExists = True
     End Sub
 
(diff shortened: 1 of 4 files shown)


instat/clsRLink.vb
```diff
@@ -178,16 +178,34 @@ Public Class RLink
...
 
         clsReadRDS.SetRCommand("readRDS")
         clsReadRDS.AddParameter("file", Chr(34) & strFile.Replace("\", "/") & Chr(34))
-        clsReadRDS.SetAssignTo(strInstatDataObject)
-        strTemp = clsReadRDS.ToScript(strScript)
-        RunScript(strScript, strComment:=strComment)
+        clsReadRDS.SetAssignTo("new_RDS")
+
+        clsImportRDS.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$import_RDS")
+        clsImportRDS.AddParameter("data_RDS", clsRFunctionParameter:=clsReadRDS)
```
dannyparsons: This looks good - just need to add parameter positions.

dannyparsons:
Looking good. 
I think my suggestion here still can be implemented https://github.com/africanmathsinitiative/R-Instat/pull/4240#discussion_r155761639

Patowhiz:
@dannyparsons  I used the suggestion you just mentioned above. Or have I'm I far from what you expected?

dannyparsons:
Ok I missed it in the code before, looks good.

instat/clsRLink.vb
Patowhiz: Will add the parameter positions

instat/dlgBackupManager.vb
```diff
@@ -1,5 +1,209 @@
...
+        If (ctrLstViewDataBackups.SelectedIndices.Count = 1) Then
+            'get the selected file path. To be used in opening the file name in form Main
+            strSelectedDataFilePath = strAutoSavedDataFilePaths(ctrLstViewDataBackups.SelectedIndices(0))
+            If strSelectedDataFilePath <> "" Then
+                'pass the selected file to the sub which will load the file to the instant object
+                frmMain.clsRLink.LoadInstatDataObjectFromFile(strSelectedDataFilePath, strComment:="Loading auto recovered data file")
+            End If
+        Else
+            'don't allow opening of more than 1 file at a time. NOTE in future this can be changed
+            MsgBox("You can only open one data file at a time")
```
dannyparsons: Does the list view have multiselect disabled? Don't mind this being here as precaution still if it is.

Patowhiz: Yes it has the multiselect enabled. The delete operation allows deleting of more than 1 file. So you can select several file paths

dannyparsons: In that case can we should either disable Open and Save buttons when multiple are selected, or disable multiselect completely. I don't have strong views either way, but it's good to avoid unnecessary pop ups if we can, because people find them annoying and confusing. Better to prevent something that isn't allowed instead of allowing users to do it and then telling them it isn't allowed.

Patowhiz: True. Thats a good way of looking at it

Patowhiz:
@dannyparsons I have made the recommended changes
On multiple selection, button open and save will be disabled. Opening and Saving of more than 1 file will NOT be allowed.

instat/dlgBackupManager.vb
```diff
@@ -1,5 +1,202 @@
...
+            Next
+        End If
+    End Sub
+
+    Private Sub ctrLstViewDataBackups_ItemSelectionChanged(sender As Object, e As ListViewItemSelectionChangedEventArgs) Handles ctrLstViewDataBackups.ItemSelectionChanged
+        TestOk()
+        If ctrLstViewDataBackups.SelectedIndices.Count > 1 Then
+            'this is meant to restrict a user from attempting to open or save more than 1 file
+            cmdOpen.Enabled = False
+            cmdSave.Enabled = False
```
dannyparsons: I can see this works but it is a bit messy, setting Enabled state of the buttons should all be done in one place. Or certainly setting the Enabled of Open and Save shouldn't be done in multiple ways.
I think you can edit SetButtonStates to maybe accept an integer of number of selected items and then set Enabled based on that.
And your TestOK is now maybe confusing since different buttons are enabled under different conditions. It's now really "Delete enabled" or "At least one item selected" but I don't mind as much if this stays the same.

Patowhiz: Done.

Patowhiz:
Please not this is not accessible via the frmMain yet. We never finalised
on which menu to put it.

Patowhiz:
@dannyparsons @rdstern  from which menu should this dialog be accessible ? Note its already merged but not accessible from the frmMain.vb

dannyparsons:
I think this is going to be under Tools, could possibly be under File as well.

Issue #4306, opened 2017-12-07
Bug on simulation of samples of n>1000


MainaJohnO:
![capture_bug_simulating_samples_greater_than1000](https://user-images.githubusercontent.com/32262789/33705490-953038d8-db41-11e7-8a7e-5db779639b11.PNG)
In regard to the eSMS assignments for R-instant during the review of topic 3 - Assignment1, under section(2.1) question 3 on simulation of samples when I try generating a sample of n=4000, the attached bug pops up. 
 An indication that the Rice_Yield_Population data sets we are using is not up to date. May be an update of the Rice_Yield_Population data sets is required
 The bug happens even if the sample size n=4000 is the first one to be generated. Trying to generate a sample of n=1001 the bug still pops up a clear indication that the bug pops up for any sample size of n>1000.

Patowhiz:
@MainaJohnO Did this happen when the number of columns =1 or even when the columns are multiple?

MainaJohnO:
It happens when the number of columns=1, seems when the columns >1 it can
work

dannyparsons:
Yes we determined the issue with @Patowhiz and @shadrackkibet

dannyparsons:
This should be fixed now by #4322. @MainaJohnO Can you check and close if it's ok?

MainaJohnO:
# Code generated by the dialog, New Data Frame

data <- data.frame(data=matrix(data=NA, ncol=1, nrow=1001))
InstatDataObject$import_data(data_tables=list(data=data))

rm(data)

I have tried out to generate a sample of size of n=1001 and it does not run
successfully. the code is as above and the bug that pops up is attached.

@Patowhiz <https://github.com/patowhiz> and @shadrackkibet
<https://github.com/shadrackkibet> the issue is not fixed

dannyparsons:
It seems to be working fine for me now. Are you using the latest version? @Patowhiz can you check this with @MainaJohnO?

shadrackkibet:
@MainaJohnO  is using our last version which was released when this was not yet fixed.
I checked this with him from the current master and we confirmed it is now working fine.So this can now be closed.

dannyparsons:
Great thanks, it would be good if @MainaJohnO can have access to the current development version through Visual Studio for testing.

Issue #4276, opened 2017-11-30
New Data Frame Error


jkmusyoka:
I could not create a new data frame of more than 1000 rows in R-Instat 0.4.9. The attached error popped up 

![newdataframeerror](https://user-images.githubusercontent.com/8112611/33421068-e939e422-d5c1-11e7-81ce-2dae7306260b.png)

rdstern:
This was to generate a new random sample.  It used to work, but does not now.

dannyparsons:
Is this the same issue as https://github.com/africanmathsinitiative/R-Instat-Help/issues/13?

Patowhiz:
@dannyparsons I'm able to create a random sample sample of n=4000 from Experimental Survey Dataset. I'm trying to figure out why the user received the error in  [https://github.com/africanmathsinitiative/R-Instat-Help/issues/13](url) . Where can I get the  Rice_Yield_Population data set ? so that I can also use it to test this.

rdstern:
This is an issue that has been raised elsewhere.  The bug is when you do File > New and create a new worksheet with **just a single column and more than 1000 rows**.
It works fine with 2 columns, etc and is just on this issue.  It is nothing to do with creating the random sample.

Patowhiz:
@rdstern  there is an issue that has been raised here  [ africanmathsinitiative/R-Instat-Help #13](https://github.com/africanmathsinitiative/R-Instat-Help/issues/13) . I was avoiding to comment it there because its an R-Instant issue. So I commented in this issue because Danny referenced it.

Patowhiz:
Nevertheless, I'm looking at both of them.

Patowhiz:
Looking deeper into the problem of generating new data frame, it appears that the real problem is in generating a data frame of 1 column with >1000 rows.  If the number of columns are > 1 then the dialog data frame is generated without any error.

dannyparsons:
Ok thanks for clarifying. That makes more sense, we've had issues with single column data frames before - usually related to it getting converted to a vector at some point. Following the error and seeing the error message from R should show what the cause is.

Patowhiz:
This is interesting. I have alienated the problem here.
The data.frame  Rcommand works fine and it creates the data frame successfully. The issue arises when the grid is being updated.  The problem is in the `$get_data_frame()` Rcommand that which throws  Error in `if (nrow(data) == 0) { : argument is of length zero `
[...]

PR #4322, opened 2017-12-12
New Data Frame Error

Patowhiz:
For issue  #4276 New Data Frame Error
Changed the **data_object_R6.R**
And minor changes on dlgNewDataFrame.vb
@jkmusyoka this solves the problem of creating a new data frame with  1 column .


instat/dlgNewDataFrame.vb
```diff
@@ -53,33 +50,26 @@ Public Class dlgNewDataFrame
         ucrNewDFName.SetLabelText("New Data Frame Name:")
```
Patowhiz: @maxwellfundi could you please check for me the text of this label. In my dialog it results to label1 . I'm suspecting the autotranslate(me) subroutine is interfering with the text of this label.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -288,7 +288,7 @@ data_object$set("public", "get_data_frame", function(convert_to_character = FALS
       }
     }
     if(!missing(max_cols) && max_cols < ncol(out)) out <- out[1:max_cols]
-    if(!missing(max_rows) && max_rows < nrow(out)) out <- out[1:max_rows, ]
+    if(!missing(max_rows) && max_rows < nrow(out)) out <- out %>% slice(1:max_rows, )
```
Patowhiz: @africanmathsinitiative/developers is there a better alternative to this? 
R seems to be changing the **object type** whenever it returns a single column data frame( i.e `out[1:max_rows, ]`  where number of columns in the dataframe out is 1), the returned object type is a **vector** and not a data frame. We had to use `slice()` here.

instat/dlgNewDataFrame.vb
```diff
@@ -14,28 +14,25 @@
...
-        If bReset Then
             SetDefaults()
+            bFirstLoad = False
         End If
         SetRCodeforControls(bReset)
         bReset = False
-        autoTranslate(Me)
-    End Sub
-
+        ucrNewDFName.SetLabelText("New Data Frame Name:")
```
Patowhiz: @maxwellfundi I had to put this code statements here cause `autoTranslate(Me)` seems to be changing the Label Text to **Label1** . We should discuss about this.

Patowhiz:

instat/dlgNewDataFrame.vb
shadrackkibet: There is a problem with this is some of our dialogs now. I am not sure whether this is the problem.

shadrackkibet: When you comment out `autoTranslate(Me)` the label text works fine. Seems this is the problem.

Patowhiz: @shadrackkibet Yeah thats what I realised too. I'm yet to check the Translation class

instat/dlgNewDataFrame.vb
```diff
@@ -53,33 +50,26 @@ Public Class dlgNewDataFrame
...
         clsOverallFunction.SetRCommand("data.frame")
 
         'matrix(nrow = 10, ncol = 2, Data = NA)
-        clsMatrixDefaultFunction.SetRCommand("matrix")
-        clsMatrixDefaultFunction.AddParameter("data", "NA")
-        clsMatrixDefaultFunction.AddParameter("ncol", 2)
-        clsMatrixDefaultFunction.AddParameter("nrow", 10)
+        clsMatrixFunction.SetRCommand("matrix")
+        clsMatrixFunction.AddParameter("data", "NA")
+        clsMatrixFunction.AddParameter("ncol", 2)
```
shadrackkibet: i think we need parameter positions here.

instat/dlgNewDataFrame.vb
```diff
@@ -44,7 +41,7 @@ Public Class dlgNewDataFrame
         ucrNudRows.SetMinMax(1, Integer.MaxValue)
 
         'nudCols
-        ucrNudCols.SetParameter(New RParameter("ncol"))
+        ucrNudCols.SetParameter(New RParameter("ncol"))
```
shadrackkibet: also here.

instat/static/InstatObject/R/data_object_R6.R
shadrackkibet: We also noticed that  `out[1:max_rows, ]` cannot be coerced as a dataframe using `as.data.frame.`.Could someone explain why this does not work?

Patowhiz:
Also fixes #4306 Bug on simulation of samples of n>1000 .

shadrackkibet:

instat/dlgNewDataFrame.vb
dannyparsons: Please add a comment in the code to say why this is here and that it is caused by a bug in the translation.

instat/dlgNewDataFrame.vb
dannyparsons: yes, position needed here

instat/dlgNewDataFrame.vb
```diff
@@ -53,33 +50,27 @@ Public Class dlgNewDataFrame
...
-        clsMatrixDefaultFunction.AddParameter("ncol", 2)
-        clsMatrixDefaultFunction.AddParameter("nrow", 10)
+        'matrix(data = NA,nrow = 10, ncol = 2)
+        clsMatrixFunction.SetRCommand("matrix")
+        clsMatrixFunction.AddParameter("data", "NA", iPosition:=0)
+        clsMatrixFunction.AddParameter("nrow", 10, iPosition:=1)
+        clsMatrixFunction.AddParameter("ncol", 2, iPosition:=2)
+
 
+        clsOverallFunction.AddParameter("data", clsRFunctionParameter:=clsMatrixFunction)
```
dannyparsons: noticed there's no parameter position here too!

instat/static/InstatObject/R/data_object_R6.R
dannyparsons: I think the only issue with `slice` is that it drops all the data frame level attributes and row numbers.
The attributes shouldn't be a problem because max_rows is only used to display the data so attributes not needed. However, row numbers might be a problem with a filter.
Can you try this:
Get a data frame with more than 1000 rows. Filter it and see if the rows numbers are displayed correctly. My suspicion is that the row numbers will reset to 1:1000. Can you check?

Patowhiz:
@africanmathsinitiative/developers I don't understand how this conflict happened. 
@dannyparsons this was waiting for your review.

maxwellfundi:
@Patowhiz  this is conflicting witht he master so likely you might not have had the latest version when working on this. You could try see whats conflicting from the pull request and see if its easy to fix it there

dannyparsons:
I've managed to resolve the conflicts but please check that this has kept the code as it should have been.

instat/dlgNewDataFrame.vb
```diff
@@ -26,25 +26,25 @@ Public Class dlgNewDataFrame
     Private Sub dlgNewDataFrame_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
             InitialiseDialog()
-            bFirstLoad = False
-        End If
-        If bReset Then
             SetDefaults()
+            bFirstLoad = False
```
dannyparsons: This section is missing the bReset If like in the other dialog

Patowhiz:
It's telling me that my commits will fast forward the changes because of no new commits. I'll consult @maxwellfundi  on this

instat/dlgNewDataFrame.vb
Patowhiz: Done

maxwellfundi:
Seems like this works for you now. great :-)

instat/static/InstatObject/R/data_object_R6.R
dannyparsons: Was this checked?

Patowhiz: Yeah @dannyparsons . I tested this and found that the filter works well and the rows are displayed as expected from my end.

dannyparsons: That's not what I was expecting, because slice clears the row names. If the filtered data has more than 1000 then slice should be called and I expect row names to clear. I will have a check of this too.
Can you add `dplyr::` to `slice` and remove the `,` in the function call?

Patowhiz: Let me investigate this further and see what happens when I do that.

Patowhiz: @dannyparsons I now get where your suspicion is. Its true slice usually clears the row names and replaces it with just character numbers.  I'm using the GMet rainfall and the numbering is indeed being affected when you do a filter that returns more than the set max rows. `The slice()` call indeed returns different row names . This happens even after I add `dplyr::` to `slice` and removing  ` , `

Patowhiz: @shadrackkibet and I had thought of using `as.data.frame` but R couldn't  coerce `out[1:max_rows, ]`   as a dataframe. Let me check if library `tibble` can help in this.

Patowhiz: @dannyparsons I think I know a way out of this.

Patowhiz: @dannyparsons I'm proposing a solution similar at this. 

```
out <- mtcars
max_cols <- 3
 
#all data frames have row names. 
#To preserve them, convert row names attr to column. 
out <- rownames_to_column(out,var = "tempcolrowname")
#slice the number of rows in the data table
#After this the data frame also gets a tibble data type
out <- out %>% dplyr::slice(1:max_cols) 
#convert column that has row names to row names attribute.
#After this the tibble data type attr is lost and only data frame is retained
out <- column_to_rownames(as.data.frame(out),var = "tempcolrowname") 

out
```

The above code is a longer alternative but it preserves the row names.
P.S I'm still catching up on R , I hope I'm not badly off.

dannyparsons: Ok thanks for confirming. So your idea is right we are basically going to have to reapply the rownames. Since this is only a problem in the case of having one column, we should separate the two cases because we won't need to do this extra stuff in most cases. I suggest something like:

```
if(ncol(out) > 1) out <- out[1:max_rows, ]
else {
    rnames <- row.names(out)[1:max_rows]
    out <- as.data.frame(dplyr::slice(1:max_rows))
    row.names(out) <- rnames
}
```

So what's happening in the else is first getting the row names before slicing. Then slice but convert the output from slice (a tibble) to a data frame, so that finally we can reapply the correct row names (since tibbles don't use row names). I prefer this over using the tibble functions because they are generally for when you want to use tibbles and we are not in this case.

Try that out and see how it works.

Patowhiz: LOL. You have really thought it out.
I have modified you code a bit to include the data frame in the out.

```
rnames <- row.names(out)[1:max_rows]
out <- as.data.frame(dplyr::slice(out,1:max_rows))
row.names(out) <- rnames
```

dannyparsons: Oh yes thanks that's right.

Patowhiz: This reminds me to ask this. Is there any chance that in future the data frame could come from a relational databases. I'm asking this because I know `slice()` is faster when it comes to in memory data manipulation but it doesn't work for relational database. An alternative in this case will be `filter()`

dannyparsons: Well with our keys and links we have already implemented a relational structure. But yes, database connections are something R is getting better at handling, so in future we could have a connection to a database in the instat object instead of actual data frame(s). It would just need some modifying of internal methods to handle the extra case.

Patowhiz:
So should we leave it as    slice(out,1:max_rows)  or we
use  filter(between(row_number(), 1, max_rows))   ?

instat/static/InstatObject/R/data_object_R6.R
Patowhiz: @dannyparsons  so should we leave it as  ` slice(out,1:max_rows) `  or we use  ` filter(out,between(row_number(), 1, max_rows))  ` ?

dannyparsons: slice is fine for now. There will be other bigger changes anyway to implement databases.

Patowhiz: Okay. Its done.

Issue #4324, opened 2017-12-13
Generating cookbook document from data frame (survey data)


dannyparsons:
Carlos and Sebastian have suggested a feature to be able to generate a codebook document from a data frame. This would consist of a description of each labelled/factor variable displaying the variable label along with the value labels for each value. Particularly useful for survey data sets which are very wide and have lots of factor variables normally.

Since we use these things (labels) already I think it would be quite straightforward to create this. I think we would want to generate an R markdown document that could then be saved and converted to other formats.

rdstern:
Sounds good.  It is an extended version of our existing Prepare > Column: Factor > View Labels, though that doesn't seem to work any more.  It doesn't produce an error, but also doesn't produce any output.

Earlier the output went (I think) into a browser.

Patowhiz:
@shadrackkibet probably we can brainstorm about the _codebook document_

dannyparsons:
The specification of what a codebook should look like can largely come from our colleagues who suggested it. But it's not an urgent feature to add.

PR #4328, opened 2017-12-15
View labels And Levels

shadrackkibet:
@africanmathsinitiative/developers this fixes display of output into the browser. Fixes part of #4324
This is ready for review.

### DIFF
--- a/instat/dlgViewLabelsAndLevels.Designer.vb
+++ b/instat/dlgViewLabelsAndLevels.Designer.vb
@@ -178,7 +178,7 @@ Partial Class dlgViewFactorLabels
         Me.Controls.Add(Me.lblFactorColumns)
         Me.Controls.Add(Me.ucrReceiverVariables)
         Me.Controls.Add(Me.ucrSelectorViewLabelsAndLevels)
-        Me.FormBorderStyle = System.Windows.Forms.FormBorderStyle.FixedDialog
+        Me.FormBorderStyle = System.Windows.Forms.FormBorderStyle.FixedToolWindow
         Me.MaximizeBox = False
         Me.MinimizeBox = False
         Me.Name = "dlgViewFactorLabels"
--- a/instat/dlgViewLabelsAndLevels.resx
+++ b/instat/dlgViewLabelsAndLevels.resx
@@ -121,6 +121,10 @@
   <data name="lblFactorColumns.AutoSize" type="System.Boolean, mscorlib">
     <value>True</value>
   </data>
+  <assembly alias="System.Windows.Forms" name="System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" />
+  <data name="lblFactorColumns.ImeMode" type="System.Windows.Forms.ImeMode, System.Windows.Forms">
+    <value>NoControl</value>
+  </data>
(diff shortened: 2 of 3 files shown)


Patowhiz:
Checking

shadrackkibet:
@Patowhiz ?

Patowhiz:
@africanmathsinitiative/developers this is ready to be merged. The change fixes the output.

Patowhiz:
@rdstern @dannyparsons @africanmathsinitiative/developers 
We had an extensive discussion with @shadrackkibet  and we had a few concerns / enhancements about this dialog. 
For instance checkbox **Show Variables Labels**, I think it will be less confusing if it were to be **Show Value Labels**. This will also make it consistent with dialog output.
@shadrackkibet has fixed the output window not showing up but the column generated for checkbox **Show Numeric Values** is simply stated as **Values**, this can be less confusing if it was stated as **Numeric Values** .

Looking further to the issue of labels, we also found in the column metadata grid, the column supposed to show the _labels of the values_ of a column was stated as **Labels** , I think it will be less confusing if its stated as **Value Labels**.

The images below illustrate what we thought could be a good addon: 

dlgviewlabels

dlgviewlabels output

column metadata

rdstern:
On the main point, then if that is what it means, I am happy to change it to **Value labels**.  But if you look at the output, then there is also a column for the **Variable labels**, only you don't have any in this data set, and that is what I thought it referred to.  So before you change the name, can you check exactly what it does refer to?

Patowhiz:
@rdstern  What is meant by **Variable labels** . Does it mean the labels of the variables(the columns of a data frame)?

In the output there is also a column which is named **Label** that refers to what I think is meant by **Variable Labels**

rdstern:
That's right.  Each variable has a name, and also (possibly) a label.  In a survey the name is usually short and the label often then gives the survey question for this variable, e.g. Livstk as the name and "Does your household have any livestock?" as the label.  

Then (in the special case of factor columns they may have labels corresponding to each level.  For example None, Few, Many.  These are called Value labels.

So a question on a course might have a variable label "How did you find the length of this workshop?, to which the value labels could be Much Too Short, Too Short, OK, Too Long, Much Too Long.

They are both useful, but different.  In our data selector, if you hover over a column it shows the Variable label for that column.  SPSS allows you to choose whether you use the name or the Variable label.

Patowhiz:
Exactly!, that's why we thought, changing 

**Show Variables Labels** to **Show Value Labels** for the view labels dialog
**Values** to **Numeric Values** for the view labels dialog output
**Labels** to **Value Labels** for the column metadata

Will make more descriptive and less confusing.
Then we can stick to such a naming convention in the entire Rinstant App where

Label will refer to this : 

> That's right. Each variable has a name, and also (possibly) a label. In a survey the name is usually short and the label often then gives the survey question for this variable, e.g. Livstk as the name and "Does your household have any livestock?" as the label.


And 
Value Labels will refer to this:

> Then (in the special case of factor columns they may have labels corresponding to each level. For example None, Few, Many. These are called Value labels.

Currently we have dialogs that are using the Label to refer to Value Label or Variable(Column) Label and this might be unclear to some until they are clear.

Patowhiz:
Okay. That can be left the way it is.

shadrackkibet:
@dannyparsons this is ready for merging.

maxwellfundi:
@dannyparsons  this looks good too and its just simple changes.

PR #4340, opened 2017-12-20
Importing Files

### LINKED ISSUE #4339: Importing Files - bug?
Two small bugs. Firstly, If I pick a csv file such as Survey, I get this

image

All great, except for something is called "Label1"

Then I reopened the dialog and it's like this -

image

1. How come reopening the dialog looks different?
2. There's a label called "Label1" that shouldn't be. I also noticed another "Label1" on the Open from Library dialog which shouldn't be called label1.

Patowhiz:
@africanmathsinitiative/developers  this is ready for reveiw
This fixes #4339 Importing Files - bug?
Label1 and group boxes visibility fix.
Please note there is redundant code in this dialog but my sole goal was to solve the Label1 and group boxes visibility.


instat/dlgImportDataset.vb
```diff
@@ -93,6 +93,16 @@ Public Class dlgImportDataset
                 RefreshFrameView()
             End If
         End If
+        'if none of the above then try setting the displayed values from the contents of ucrInputFilePath.
+        'This happens using EditLastDialogueToolStrip to open this dialog 
+        If Not bDialogLoaded AndAlso Not String.IsNullOrEmpty(ucrInputFilePath.GetText()) AndAlso File.Exists(ucrInputFilePath.GetText()) Then
```
dannyparsons: Instead of being a separate If statement, shouldn't this be an Else of something in the above If? Possibly the Else of:
`             If Not File.Exists(strFilePathSystem) Then`
Checking `Not bDialogLoaded` isn't the same as opening the dialog from EditLastDialogueToolStrip so this doesn't seem quite right for the fix.

Patowhiz: @dannyparsons . What you just observed is true. I had a similar opinion at first but then I didn't want to interfere with original flow of the code and I chose an alternative path instead. The if statement when put in the else section will go further and make the code that its already there redundant. Anyway let me change the code to my original perception.

maxwellfundi:
I believe there many more dialogs especially on the prepare menu that have same issue. Can you fix those too?

Patowhiz:
@maxwellfundi I have committed the temporary fix for the other dialogs too

maxwellfundi:
Great, are these all the dialogs that use that?

Patowhiz:
Yeah

PR #4357, opened 2018-01-15
Adding legend position to sdgPlots themes tab

### LINKED ISSUE #4252: The theme tab on sdgplot
The legend is part of the them tab.
I wonder if we could also have one aspect of the legend on the main plot dialogues?
This is initially for discussion.  It would be an easy way to add the argument theme(legend.position) to the ggplot command.
I assume it might be a check box, but I am not sure. 
The options could then be Default, right, left, bottom, top, none.  Default would the default! - which is why I wonder if we need the checkbox.  This would not give the argument.  I can see some uses for Bottom, but am particularly interested in none.  That gets rid of the legend and makes more space for the plot.  That can be particularly useful with facets when it is good to have all the area we need.

maxwellfundi:
Fixes #4252

### DIFF
--- a/instat/sdgPlots.vb
+++ b/instat/sdgPlots.vb
@@ -47,6 +47,7 @@ Public Class sdgPlots
     Private bRCodeSet As Boolean = False
     Private bResetThemes As Boolean = True
     Private ucrBaseSelector As ucrSelector
+    Dim clsElementText As New RFunction
 
     'See bLayersDefaultIsGolobal below.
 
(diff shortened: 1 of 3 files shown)


maxwellfundi:
@dannyparsons  this works properly with the linking and i wanted to make sure you approve this code so that i get to finosh of the rest. Is there is any potential issue you see about this?

Muthenya:
I've tested it and it looks good but I've noted that the second checkbox, that is, ucrChkXtickMarkerAngle, unchecks itself after it is applied on the graph. Otherwise, they work well. Nice work.

maxwellfundi:
@Muthenya  I edited the condition i had put for the checkbox and now works fine :-)

Muthenya:
Nice.

instat/sdgPlots.vb
```diff
@@ -47,6 +47,7 @@ Public Class sdgPlots
     Private bRCodeSet As Boolean = False
     Private bResetThemes As Boolean = True
     Private ucrBaseSelector As ucrSelector
+    Dim clsElementText As New RFunction
```
dannyparsons: I think this may not be quite right. Because there should be an `element_text` function also on the themes sub dialog so if you are creating one here it's not going to be in sync with the one of the themes sub dialog.

maxwellfundi: I see you are right, i had tested this one way, will see how to get them to be in sync

dannyparsons: Instead of creating a new function here you need to link it to the corresponding function which is either in the main theme function, or I think we have a list of theme functions passed through to the sub dialogs?

maxwellfundi: We have a theme function thats passed through the sub dialogs, i think i could use that. Will look at it.

dannyparsons: The element_text is the function of a parameter of the main theme function. But I think we also have those functions that make up parameters in a list/dictionary in the sub dialog too.

dannyparsons:
The `dctThemeFunctions` is the set of functions that go inside `theme()` so this function should go in there if its a part of `theme()` so that it syncs up correctly.

shadrackkibet:
@maxwellfundi  is this ready for review?

maxwellfundi:
@bethanclarke  If you are taking up the task on the themes tab, then this conversation would be of help.

PR #4362, opened 2018-01-16
Saving Script to file

### LINKED ISSUE #4348: Saving and opening a script file?
I have two problems, both saving and loading:

1. Saving a script file: I had some data in a script file, as follows:

**code generated by the dialog Calculations
age <- wakefield::area(20,x=c("a","b","c","d"))
InstatDataObject$add_columns_to_data(col_data=age, col_name="roger", data_name="data1")
rm(age)**

I used **File > Save As** to save these contents.  It seemed to save as a text file.  But the file was then empty.  I tried, both with the text selected and without.

2. So I copied the file to the clipboard and saved it as a txt file myself instead.  Then I went to the script file, cleared it and then used the right-click option **Open Script**.  This was a surprise as it took me to RStudio.  Perhaps that is also useful, but how do I open a script file I have saved?

3. I do rather like the **right-click** option to open a script file, so I wonder if we have both - assuming opening from RStudio is what is also wanted.  And perhaps we could also have **Save As** on the right-click menu as well?  Can **Save As** also be to either save the whole script window or the selected text?

maxwellfundi:
@rdstern  @dannyparsons  here is the current menu for the right click on the script window. 

image

The Save Script to File  saves the script to a text file and the Open Script From File clears the current text on the script window before loading the script window with the contents of the selected, previously saved text file. 

Fixes #4348

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -117,7 +117,7 @@ Public Class ucrScript
             dlgOpen.InitialDirectory = frmMain.clsInstatOptions.strWorkingDirectory
 
             If dlgOpen.ShowDialog() = DialogResult.OK Then
-                msgWarning = MessageBox.Show("Opening a script from file will clear your current script", "Open Script From File", MessageBoxButtons.YesNo)
+                msgWarning = MessageBox.Show("Opening a script from file will clear your current script" & Environment.NewLine & "Do you want still want to open?", "Open Script From File", MessageBoxButtons.YesNo)
                 If msgWarning = DialogResult.Yes Then
                     Try
                         txtScript.Clear()
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -88,4 +88,15 @@ Public Class ucrScript
         End Try
     End Sub
 
+    Private Sub SaveScriptToolStripMenuItem_Click(sender As Object, e As EventArgs) Handles mnuSaveScript.Click
+        Using dlgSave As New SaveFileDialog
+            dlgSave.Title = "Save Script as Text File"
+            dlgSave.Filter = "Text file (*.txt)|*.txt"
+            dlgSave.InitialDirectory = frmMain.clsInstatOptions.strWorkingDirectory
+
+            If dlgSave.ShowDialog() = DialogResult.OK Then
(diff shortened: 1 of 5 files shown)


maxwellfundi:
@rdstern  the saving of log and script from the file menu is now fixed too.
 Just a clarification, I would like to know if we want to be able to save two types of files in both cases- an R file or a txt file. In the File>Save As for both log and script window, you can be able to save either a txt file or an R file. 
I find that R can still open a txt file comfortably and so am not sure we need to be able to save both. @dannyparsons  @volloholic  what are your thoughts on this?

maxwellfundi:

dannyparsons:
1. Yes .R and .txt should both be options but with .R as the default. .R is the "correct" format for an R script, but in case someone doesn't have RStudio then .txt might be easier for their machine to automatically open.

2. **Open Script from File** - Can there be a warning that this will clear the current script before they open the file?

3. **Open Script in RStudio** - This is only true if they have RStudio installed. Change to **Open in RStudio/text editor**?

4. Change **Clear Contents** to **Clear Script**.

5. Change **Save Script to File** to **Save Script**.

Also add `...` to items leading to dialogs.

maxwellfundi:
@dannyparsons this is ready.

instat/ucrScript.vb
```diff
@@ -88,4 +89,45 @@ Public Class ucrScript
...
+        Using dlgOpen As New OpenFileDialog
+            dlgOpen.Title = "Open Script From Text File"
+            dlgOpen.Filter = "Text File (*.txt)|*.txt|R Script File (*.R)|*.R"
+            dlgOpen.InitialDirectory = frmMain.clsInstatOptions.strWorkingDirectory
+
+            If dlgOpen.ShowDialog() = DialogResult.OK Then
+                msgWarning = MessageBox.Show("Opening a script from file will clear your current script" & Environment.NewLine & "Do you want still want to open?", "Open Script From File", MessageBoxButtons.YesNo)
+                If msgWarning = DialogResult.Yes Then
+                    Try
+                        txtScript.Clear()
```
dannyparsons: I think this line should be removed. If the text can't be read from the file, then we should leave the text as it was. By setting the Text in the next line we automatically overwrite the previous text.

maxwellfundi: okay

PR #4368, opened 2018-01-19
Column Structure

### LINKED ISSUE #4367: Prepare > Data Frame > Column Structure
The check box  colour columns by structure does not maintain its previous state of being checked.
column structure

Patowhiz:
Fixed ucrChkColourColumnsByStructure default state ONLY.
Fixes #4367
I've commented some of the code statements

### DIFF
--- a/instat/dlgColumnStructures.vb
+++ b/instat/dlgColumnStructures.vb
@@ -18,7 +18,8 @@ Imports instat.Translations
 Public Class dlgColumnStructure
     Private bReset As Boolean = True
     Private clsColourByMetadata, clsColumnStructure, clsUncolourByMetadata As New RFunction
-    'clsCourByStructure is here to construct the R-command that will colour columns according to their type in case it is required (see relevant tick box).
+    'this RFunction is just used for setting the appropriate RCODE of the ucrChkColourColumnsByStructure control
+    Private clsColourStructure As New RFunction
     Private bFirstLoad As Boolean = True
     Private Sub ucrSelectorColumnStructures_Load(sender As Object, e As EventArgs) Handles Me.Load
         autoTranslate(Me)
@@ -57,17 +58,16 @@ Public Class dlgColumnStructure
         ucrReceiverMeasurement.bExcludeFromSelector = True
 
         ucrChkColourColumnsByStructure.SetText("Colour Columns by Structure")
-        'ucrChkColourColumnsByStructure.AddFunctionNamesCondition(True, frmMain.clsRLink.strInstatDataObject & "$set_column_colours_by_metadata")
-        'ucrChkColourColumnsByStructure.AddFunctionNamesCondition(False, frmMain.clsRLink.strInstatDataObject & "$remove_column_colours")
-        'ucrChkColourColumnsByStructure.AddFunctionNamesCondition(True, frmMain.clsRLink.strInstatDataObject & "$remove_column_colours")
+        ucrChkColourColumnsByStructure.AddFunctionNamesCondition(True, frmMain.clsRLink.strInstatDataObject & "$set_column_colours_by_metadata")
+        ucrChkColourColumnsByStructure.AddFunctionNamesCondition(False, frmMain.clsRLink.strInstatDataObject & "$remove_column_colours")
     End Sub
 
     Private Sub SetDefaults()
         clsColourByMetadata = New RFunction
         clsColumnStructure = New RFunction
         clsUncolourByMetadata = New RFunction
+        clsColourStructure = clsUncolourByMetadata
 
-        ucrChkColourColumnsByStructure.Checked = False
         ucrSelectorColumnStructure.Reset()
         SetColumnStructureInReceiver()
         ucrReceiverLayout.SetMeAsReceiver()
@@ -90,7 +90,7 @@ Public Class dlgColumnStructure
         ucrReceiverLayout.SetRCode(clsColumnStructure, bReset)
         ucrReceiverTreatment.SetRCode(clsColumnStructure, bReset)
         ucrReceiverMeasurement.SetRCode(clsColumnStructure, bReset)
-        'ucrChkColourColumnsByStructure.SetRCode(clsUncolourByMetadata, bReset)
+        ucrChkColourColumnsByStructure.SetRCode(clsColourStructure, bReset)
     End Sub
 
     Private Sub TestOKEnabled()
@@ -110,8 +110,10 @@ Public Class dlgColumnStructure
     Private Sub ucrBase_ClickOk(sender As Object, e As EventArgs) Handles ucrBase.ClickOk
         If ucrChkColourColumnsByStructure.Checked Then
             frmMain.clsRLink.RunScript(clsColourByMetadata.ToScript)
+            clsColourStructure = clsColourByMetadata
         Else
             frmMain.clsRLink.RunScript(clsUncolourByMetadata.ToScript)
+            clsColourStructure = clsUncolourByMetadata
         End If
     End Sub
 
--- a/instat/dlgColumnStructures.vb
+++ b/instat/dlgColumnStructures.vb
@@ -14,6 +14,7 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports instat
 Imports instat.Translations
 Public Class dlgColumnStructure
     Private bReset As Boolean = True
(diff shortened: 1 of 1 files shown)


instat/dlgColumnStructures.vb
```diff
@@ -57,16 +57,17 @@ Public Class dlgColumnStructure
         ucrReceiverMeasurement.bExcludeFromSelector = True
```
Patowhiz: @dannyparsons  what's the essence of   `ucrReceiverMeasurement.bExcludeFromSelector = True` of doing this in this context? I didn't understand it.

dannyparsons: When a column is added to a receiver, it is removed from the selector (so that you can't add a column in two receivers at the same time) since each column can only be one type.

Patowhiz: Hahaha. I didn't see it that way. Thanks

instat/dlgColumnStructures.vb
```diff
@@ -57,16 +57,17 @@ Public Class dlgColumnStructure
...
+        'ucrChkColourColumnsByStructure.AddFunctionNamesCondition(False, frmMain.clsRLink.strInstatDataObject & "$remove_column_colours")
+        'ucrChkColourColumnsByStructure.AddFunctionNamesCondition(True, frmMain.clsRLink.strInstatDataObject & "$remove_column_colours")
     End Sub
 
     Private Sub SetDefaults()
         clsColourByMetadata = New RFunction
         clsColumnStructure = New RFunction
         clsUncolourByMetadata = New RFunction
 
+        ucrChkColourColumnsByStructure.Checked = False
```
dannyparsons: We don't want to set the value of any controls in SetDefaults() because we want the controls to be able to read their value from the code. That's why we have conditions and link the control to an R code. From looking at the conditions, there are two conditions for True, which can't be satisfied at the same time. They should go into one condition with the two function names a list so it will check if it's either of them, instead of both.
Also the SetRCode is linking it to clsUncolourByMetadata where I think it should like to the base R code in ucrBase because this is dependant on the function.

Patowhiz: The base R code is clsColumnStructure .  The clsUncolourByMetadata and clsColourByMetadata are being independently of clsColumnStructure . Linking thecolour metadata RFunctions and setting the control to read from the base function will still make the check box not to retain its state.

Patowhiz: The colour and uncolour metadata RFunctions are being run after the base function has been executed

dannyparsons: OK yes, now I remember how this dialog works. In that case we probably need to add another "dummy" function which is set to `clsColourByMetadata` when checked and `clsUncolourByMetadata` when unchecked. Then the simple True and False conditions on the dummy function will work.
Not sure if I've explained that well, if not I can make the changes.

Patowhiz: You mean a "dummy"  function that sets the **RCode** for the check box to `lsColourByMetadata `when checked and `clsUncolourByMetadata `when unchecked?

Patowhiz: Something like this

 ```
Private Sub Dummy(bReset As Boolean)
        If ucrChkColourColumnsByStructure.Checked Then
            ucrChkColourColumnsByStructure.SetRCode(clsColourByMetadata, bReset)
        Else
            ucrChkColourColumnsByStructure.SetRCode(clsUncolourByMetadata, bReset)
        End If
    End Sub
```

Comment (dannyparsons) [reply instat/dlgColumnStructures.vb]: No I mean a dummy RFunction which is set as the R code for the checkbox, and then when the checkbox is changed the dummy RFunction is updated to the appropriate colour RFunction.

Comment (Patowhiz) [reply instat/dlgColumnStructures.vb]: Okay. I get you now. but don't you find it such additions as so unnecessary for this case. Adding an addition `RFunction `just to achieve a default unchecked state of a control which doesn't have a parameter attached to it. Only the check box status (Checked / Unchecked) is involved and it's only in determining which `RFunction `will be run. (I'm not suggesting we do this but just to show how much the checkbox contributes  ) We could literally use a normal check box and the dialog won't be affected.

Comment (Patowhiz) [reply instat/dlgColumnStructures.vb]: I find it much direct to remove `AddFunctionNamesCondition ` for the check box and have a comment as to why we chose to set the check box status in the set default for this special scenario. I think the author of the dialog had originally thought that this check box will have an R Code otherwise the implementation would have been easier if a normal checkbox control was used

**Comment by Patowhiz** [comment]
Okay

On Jan 19, 2018 7:46 PM, "Danny Parsons" <notifications@github.com> wrote:

*@dannyparsons* commented on this pull request.
------------------------------

In instat/dlgColumnStructures.vb
<https://github.com/africanmathsinitiative/R-Instat/pull/4368#discussion_r162673561>
:

>      End Sub

     Private Sub SetDefaults()
         clsColourByMetadata = New RFunction
         clsColumnStructure = New RFunction
         clsUncolourByMetadata = New RFunction

+        ucrChkColourColumnsByStructure.Checked = False

Our structures of implementing our own user controls with structures for
linking controls to RCode and using conditions definitely adds a layer of
complexity to creating dialogs and makes the code less intuitive. However,
without this we don't have the functionality of the controls being able to
read their value from a piece of RCode - we need this for subdialog and it
is a really powerful feature in general.
This will allow us, for example, to click on a piece of R code in the
output window and automatically open the correct dialog and set the
controls exactly as they were when that code was run. We're not doing this
yet but with a bit of work on the Output Window we could because the
structures are there in the controls. If we go back to setting values of
controls directly in the dialog, the controls are no longer reading from
the code and we lose the potential of these features.
So yes in cases like this the code can look unnecessarily complex but I
think the pays off in the long run will be worth it!

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/africanmathsinitiative/R-Instat/pull/4368#discussion_r162673561>,
or mute the thread

File: instat/dlgColumnStructures.vb [hunk 3]
```diff
@@ -109,8 +110,10 @@ Public Class dlgColumnStructure
     Private Sub ucrBase_ClickOk(sender As Object, e As EventArgs) Handles ucrBase.ClickOk
         If ucrChkColourColumnsByStructure.Checked Then
             frmMain.clsRLink.RunScript(clsColourByMetadata.ToScript)
+            clsColourStructure = clsColourByMetadata
         Else
             frmMain.clsRLink.RunScript(clsUncolourByMetadata.ToScript)
+            clsColourStructure = clsUncolourByMetadata
         End If
     End Sub
```
Patowhiz: @dannyparsons I've made the changes. Can I move this

 ```
If ucrChkColourColumnsByStructure.Checked Then
            clsColourStructure = clsColourByMetadata
   Else
             clsColourStructure = clsUncolourByMetadata
   End If
```
to an **event of the check box**  and just call

```
        frmMain.clsRLink.RunScript(clsColourStructure .ToScript)
  End Sub
```

this will also take care of the check box when the user just clicks on **close** instead of **ok** button.

If that's not necessary to you then this is done.

File: instat/dlgColumnStructures.vb [hunk 4]
```diff
@@ -18,7 +18,8 @@ Imports instat.Translations
 Public Class dlgColumnStructure
     Private bReset As Boolean = True
     Private clsColourByMetadata, clsColumnStructure, clsUncolourByMetadata As New RFunction
-    'clsCourByStructure is here to construct the R-command that will colour columns according to their type in case it is required (see relevant tick box).
+    'this RFunction is just used for setting the appropriate RCODE of the ucrChkColourColumnsByStructure control
+    Private clsColourStructure As New RFunction
```
Patowhiz: I'm not sure about the name to give this RFunction

instat/dlgColumnStructures.vb
dannyparsons: Yes exactly, you can use the ControlValueChanged of the checkbox and then simplify as you have above.

instat/dlgColumnStructures.vb
dannyparsons: That seems fine.

instat/dlgColumnStructures.vb
Patowhiz: Done.

Patowhiz:

Patowhiz:
@maxwellfundi can you review this

Patowhiz:

shadrackkibet:

instat/dlgColumnStructures.vb
```diff
@@ -14,11 +14,13 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports instat
```
shadrackkibet: Do you need this import?

Patowhiz: Nope. That was originally in the dialog but it's unused. I have removed it.

shadrackkibet:

PR #4370, opened 2018-01-22
Climdex indices

### LINKED ISSUE #4350: Improvements to climdex dialog
I have modified the R method for importing the climdex indices to get the linking working better (#4349) which mean the dialog needs updating to these changes.

We decided that year (for annual summaries) and year & month (for monthly summaries) should be required columns on the dialog. This isn't required to calculate the climdex indices because it uses the date column, but we need those columns to do the linking correctly. They do not have to be a specific data type.

So the dialog needs a year and month receiver added. The choice of annual or monthly summaries is currently on the sub dialog. I think this has to come to the main dialog now since the visibility of the month receiver depends on it. I'm not sure whether it should still be a combo box, and where it should go on the main dialog, or whether it should change to radio buttons at the top of the dialog. Maybe @shadrackkibet @rdstern @volloholic can discuss.

The sub dialog also needs a rethink probably. The R method we've written only allows either annual or monthly summaries to be calculated in one go. If we're happy with that restriction, then the choice of indices visible should change depending on this choice which I don't think it does currently. I know @lilyclements had a go at tidying up the sub dialog before so she might have thoughts on this too?

shadrackkibet:
@africanmathsinitiative/developers  this is ready for review.Fixes  #4350


shadrackkibet:
@rdstern this is how the designer looks for the main dilaog and the subdialog.

**When Annual radiobutton is checked.** We have the following enabled in the subdialog.
image
image
image
image

**When Monthly radiobutton is checked.** We have the following enabled/disabled in the subdialog.
image

image
image
image

dannyparsons:
This looks great. 
I am wondering what should happen when you switch between Annual/Monthly because the indices you selected may not apply when you change this. 

a) The easiest would be to clear them all. But this could be annoying if you spent time setting them up and then they are gone.
b) Remove all the Annual only summaries when moving from Annual to Monthly. Since all summaries can be Annual, no need to clear when going from Monthly to Annual.

b) is probably better but I don't mind too much either way as long as we have something that is stable to use.

shadrackkibet:
@dannyparsons   **b** above has been implemented. Please have a look at it.

dannyparsons:
@africanmathsinitiative/developers can someone test this?

Muthenya:
This looks good. @dannyparsons this is ready for you to review.

PR #4371, opened 2018-01-22
Calculator

### LINKED ISSUE #4345: The calculations dialogue continued - particularly cor and cov
I can't find where we first discussed cor and cov for the calculations.  It is in the summary dialogue and fine as far as it goes.  But they need 2 more arguments, namely **use** and **method**, see:

cor(x, y = NULL, use = "everything",
    method = c("pearson", "kendall", "spearman"))

The default of use is "everything", but that means that any missing values in the data give a missing result.  The other obvious one is use="c" short for "complete.obs", which will give the result for the non-missing combinations.

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes #4345


Patowhiz:
Reviewing this

Patowhiz:
@shadrackkibet probably this is the expected behaviour and how its supposed to be but I'm asking this just for clarification.  Right now, when I can click on any of the the functionality buttons like length,cor , cov  twice, the expression is generated twice. I mean for every length button click results in an expression being added in where the cursor position is at.  I thought it should clear what's already there and have a newly generated expression.

rdstern:
I am OK that it appears twice.  But what I also understand is difficult to change and am not so happy with is when you select the existing contents of the receiver, then press anything and it adds and doesn't replace what was selected.  You need to press the delete key first.

instat/ucrCalculator.vb
```diff
@@ -1147,15 +1147,15 @@ Public Class ucrCalculator
 
     Private Sub cmdCor_Click(sender As Object, e As EventArgs) Handles cmdCor.Click
         If chkShowArguments.Checked Then
-            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("cor(x= ,y= )", 5)
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("cor(x= ,y= ,use = ""everything"", method = c(""pearson"", ""kendall"", ""spearman""))", 71)
```
Patowhiz: Works fine

instat/ucrCalculator.vb
```diff
@@ -1147,15 +1147,15 @@ Public Class ucrCalculator
...
         Else
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("cor()", 1)
         End If
     End Sub
 
     Private Sub cmdCov_Click(sender As Object, e As EventArgs) Handles cmdCov.Click
         If chkShowArguments.Checked Then
-            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("cov(x= ,y= )", 5)
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("cov(x= ,y= ,use = ""everything"", method = c(""pearson"", ""kendall"", ""spearman""))", 71)
```
Patowhiz: works fine

Patowhiz:
@shadrackkibet just noted this is to enable the the functions be able to be concatenated. Replacing the expression will make this impossible to do.

shadrackkibet:
I think sometimes you might want to nest this functions for example having ` sum(range(Size))`.So it is useful to be able to click on one button and add a function automatically inside another function.

shadrackkibet:
I recently added ` mode` could you please that as well?

instat/ucrCalculator.vb
```diff
@@ -1192,4 +1192,8 @@ Public Class ucrCalculator
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("dplyr::nth()", 1)
         End If
     End Sub
+
+    Private Sub cmdMode_Click(sender As Object, e As EventArgs) Handles cmdMode.Click
+        ucrReceiverForCalculation.AddToReceiverAtCursorPosition("summary_mode()", 1)
```
Patowhiz: Works fine too

Patowhiz:
This works fine to me.

shadrackkibet:

PR #4380, opened 2018-01-30
Find/Replace.

### LINKED ISSUE #4379: String Handling dialog should allow to Replace a character with an empty string
We have had a case where we had a year column with a y character at the start (y1923, y1924......) except for the 365 rows of 1928 which a had a proper year(1928). 

We wanted to replace the y in all the years after 1922 with an empty string but the dialog couldnot allow that. 

A solution is to make the dialog pass in an empty string as the replace value when the replace value is empty.

shadrackkibet:
@africanmathsinitiative/developers this is ready for review.Here i have allowed empty string on replacement parameter. Fixes #4379


maxwellfundi:
I have been reviewing this but there is a bug when you leave the Replace By empty or i enter an empty string ""

image

I think you could pass in the empty string when the replace by input has nothing in it.

shadrackkibet:
@maxwellfundi thanks for testing this.I noticed` SetRCodeForControls `has been commented in the code.So commenting out this will fix the issue.

maxwellfundi:
great, this looks good now @dannyparsons

PR #4382, opened 2018-01-31
fixed bug with filter being applied incorrectly in calculations

dannyparsons:


dannyparsons:
@africanmathsinitiative/developers I'm not sure if there is an issue on this bug, but this fixes the problem that sometimes the Column Summaries was ignoring the filter applied to the data when summarising. It was only an issue when there was more than one condition which is why the problem appeared inconsistent.

PR #4383, opened 2018-01-31
Three way frequency.

shadrackkibet:
@africanmathsinitiative/developers this is ready for review.This solves the developer error when you switch between radio buttons on graph frequencies and then reopen the dialog.


dannyparsons:
These changes in the conditions class fix a bug which caused this problem, so looks good.

PR #4395, opened 2018-02-06
Adding Cut/Copy/Paste for the script window 

### LINKED ISSUE #4389: Cut/Copy/Paste right click items needed on script window

maxwellfundi:

As of now, this copies/cuts the entire text in the script window. I wonder if we could include an option of copying/cutting  selected text. 

Fixes #4389

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -142,4 +142,24 @@ Public Class ucrScript
     Private Sub mnuPaste_Click(sender As Object, e As EventArgs) Handles mnuPaste.Click
         txtScript.Text = Clipboard.GetText()
     End Sub
+
+    Private Sub EnableCopyCut()
+        If txtScript.SelectionLength > 0 Then
+            mnuCopy.Enabled = True
+            mnuCut.Enabled = True
+            mnuRunSelectedText.Enabled = True
+        Else
+            mnuCopy.Enabled = False
+            mnuCut.Enabled = False
+            mnuRunSelectedText.Enabled = False
+        End If
+    End Sub
+
+    Private Sub ucrScript_Load(sender As Object, e As EventArgs) Handles Me.Load
+        EnableCopyCut()
+    End Sub
+
+    Private Sub txtScript_MouseUp(sender As Object, e As EventArgs) Handles txtScript.MouseUp, txtScript.KeyUp
+        EnableCopyCut()
+    End Sub
 End Class
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -129,7 +129,17 @@ Public Class ucrScript
         End Using
     End Sub
 
-    Private Sub PasteToolStripMenuItem_Click(sender As Object, e As EventArgs) Handles PasteToolStripMenuItem.Click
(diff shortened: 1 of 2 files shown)


dannyparsons:
Yes it should only copy/cut the selected text, otherwise it works differently from copy and paste in almost every other software.

maxwellfundi:
@africanmathsinitiative/developers This is now ready for review. 

The copy, Cut and run selected should only be enabled when some text is selected
the clear script, open script in Rstudio, save script and the run all button should be enabled when there is some text on the script

shadrackkibet:
reviewing and testing this.

instat/ucrScript.vb
```diff
@@ -129,7 +129,17 @@ Public Class ucrScript
...
 
-    Private Sub PasteToolStripMenuItem_Click(sender As Object, e As EventArgs) Handles PasteToolStripMenuItem.Click
+    Private Sub mnuCopy_Click(sender As Object, e As EventArgs) Handles mnuCopy.Click
+        Clipboard.SetText(txtScript.SelectedText)
+    End Sub
+
+    Private Sub mnuCut_Click(sender As Object, e As EventArgs) Handles mnuCut.Click
+        Clipboard.Clear()
+        Clipboard.SetText(txtScript.SelectedText)
+        txtScript.Text = ""
```
shadrackkibet: I think this will set the script text to empty even if i select  just select part of the text. This should be .selectedText?

instat/ucrScript.vb
```diff
@@ -129,7 +129,17 @@ Public Class ucrScript
...
+    End Sub
 
+    Private Sub mnuPaste_Click(sender As Object, e As EventArgs) Handles mnuPaste.Click
+        txtScript.Text = Clipboard.GetText()
```
shadrackkibet: I think there should be a warning here to let the user know that pasting into the script window will replace the content of the script window. 

Sometimes the user might want to paste to the  selected text on the script window. Should you also consider that case as well?

instat/ucrScript.vb
maxwellfundi: ooo yeah you are right there, we need to just remove the cut text only..

instat/ucrScript.vb
maxwellfundi: Will put this in place too :-) 

I am not sure we want to always paste on some selected text but that could be a case. then we need to check if any text is selected and give a warning there.

rdstern:
This looks like a good and constructive testing of this issue.  
Usually pasting automatically replaces the selected text - with no warning.  But there is also usually an undo in case that was a mistake.  I assume it isn't easy to have an undo here - and so it would be good to have a warning?

maxwellfundi:
With the textbox its easy to do an undo.

rdstern:
Oooh, that would be nice then - to include an undo and redo in a little section on the right-click!  And in that case I would not bother with a warning.

maxwellfundi:
@shadrackkibet  you can now re-review this

shadrackkibet:
This is now working fine for me.

maxwellfundi:
Great, thanks for review :-)

dannyparsons:
I think you have made this harder than needed. The textbox control has Paste, Cut and Copy methods, we should be using those instead of writing our own functionality.

PR #4399, opened 2018-02-08
ucrCalculator

### LINKED ISSUE #4128: Calculations - new keyboard

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes #4128

### DIFF
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -1002,7 +1002,11 @@ Public Class ucrCalculator
     End Sub
 
     Private Sub cmdDiff_Click(sender As Object, e As EventArgs) Handles cmdDiff.Click
-        ucrReceiverForCalculation.AddToReceiverAtCursorPosition("lag()", 1)
+        If chkShowArguments.Checked Then
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("c(NA,diff(x= , lag = 1, differences = 1))", 29)
+        Else
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("c(NA,diff())", 2)
+        End If
     End Sub
 
     Private Sub cmdEcdf_Click(sender As Object, e As EventArgs) Handles cmdEcdf.Click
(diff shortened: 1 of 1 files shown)


instat/ucrCalculator.vb
```diff
@@ -1002,7 +1002,11 @@ Public Class ucrCalculator
     End Sub
 
     Private Sub cmdDiff_Click(sender As Object, e As EventArgs) Handles cmdDiff.Click
-        ucrReceiverForCalculation.AddToReceiverAtCursorPosition("lag()", 1)
+        If chkShowArguments.Checked Then
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("c(NA,diff(x= , lag = 1, differences = 1))", 29)
+        Else
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("c(NA,diff())", 2)
```
dannyparsons: Can you include a space after the comma before the `diff` parameter? Good to check this formatting generally in the calculator.

shadrackkibet: okay.

Issue #4386, opened 2018-02-01
Giving the by command?


rdstern:
Is there any way we can currently give the **by** command in R-Instat.  I am thinking initially with a command we type into the calculator or in the script window.  That sort of gives the same power for functions (like the statistical tests) that the summary dialogues gives for data manipulation.

The information says it is to  "Apply a Function to a Data Frame Split by Factors"

dannyparsons:
If your calculation returns a single value you can use `dplyr`, e.g.

`survey %>% dplyr::group_by(Village.) %>% dplyr::summarise(t.test(Field, Size)$p.value)`

```
# # A tibble: 4 x 2
#   Village. `t.test(Field, Size)$p.value`
#     <fctr>                         <dbl>
# 1    KESEN                     0.5246753
# 2    NANDA                     0.0008996
# 3     NIKO                     0.6953557
# 4    SABEY                     0.0022165

```
works in the calculator. 

In the script window this would be:
```
survey <- InstatDataObject$get_data_frame(data_name="survey")
survey %>% dplyr::group_by(Village.) %>% dplyr::summarise(t.test(Field, Size)$p.value)
rm(list=c("survey"))
```

To get the full output for a statistical test you can use `by`:
`by(survey, survey$Village., function(x) t.test(x$Field, x$Size))`

```
# survey$Village.: KESEN
# 
# Welch Two Sample t-test
# 
# data:  x$Field and x$Size
# t = 0.66, df = 12, p-value = 0.5
# alternative hypothesis: true difference in means is not equal to 0
# 95 percent confidence interval:
#   -2.162  4.019
# sample estimates:
#   mean of x mean of y 
# 5.143     4.214 
# 
# ------------------------------------------------------------ 
#   survey$Village.: NANDA
# 
# Welch Two Sample t-test
# 
# data:  x$Field and x$Size
# t = 3.9, df = 19, p-value = 9e-04
# alternative hypothesis: true difference in means is not equal to 0
# 95 percent confidence interval:
#   4.86 16.00
# sample estimates:
#   mean of x mean of y 
# 15.071     4.643 
# 
# ------------------------------------------------------------ 
#   survey$Village.: NIKO
# 
# Welch Two Sample t-test
# 
# data:  x$Field and x$Size
# t = -0.41, df = 7.6, p-value = 0.7
# alternative hypothesis: true difference in means is not equal to 0
# 95 percent confidence interval:
#   -4.033  2.833
# sample estimates:
[...]

PR #4409, opened 2018-02-13
Hypothesis Calculator.

### LINKED ISSUE #4385: Adding hypothesis tests to the model menu
I have made this into a 4.11 milestone because I would like it to be discussed before David and Danny leave - at least discussed.

I suggest that **adding a tests keyboard** to the calculator will be:
a) Very useful in its own right.
b) Useful for the next workshop in Riga in early May (with e-learning component in April)
c) Quick and easy to do
d) Sets us free in the modelling to concentrate on the more important components.

I see 3 potential problems:
a) Not as easy as using ordinary dialogues.  (I don't care.  If we need, then they can come later in addition to this feature.)
b) In the wrong menu, i.e. prepare, rather than model.  I don't care - at least for now.
c) The calculator doesn't take account of the filters.  That's annoying, but perhaps that can be sorted out at a later stage - maybe even just for this keyboard?

I am particularly interested in the tests that have been included in the stats package, because that is by the core R team.  There are 27 tests and I think most of them can be included on the keyboard.  I think they are all similar to the t.test command in just giving a few lines of results in the output window.  There are then a few more in other packages, particularly in **lawstat** and **outliers**.

So if the keyboard is called tests, the keys can mainly be labelled simply with the name of the test.  The command, in each case is stats::xxx.test(...).

Here is the list from the stats documentation.  If someone is going to take this on, then I'll look at a sensible order for the keyboard.

bartlett, binom, Box, chisq, cor, fisher
fligner, friedman, ks, kruskal, mauchly
mcnemar, mood, oneway
pairwise: prop, t, wilcox
power: anova, prop, t
poisson, prop, prop.trend, shapiro, t, var, wilcox

shadrackkibet:
In progress. Fixes #4385 

**Designer layout:**

image

### DIFF
--- a/instat/dlgHypothesisTestsCalculator.Designer.vb
+++ b/instat/dlgHypothesisTestsCalculator.Designer.vb
@@ -82,7 +82,7 @@ Partial Class dlgHypothesisTestsCalculator
         '
         Me.lblData.AutoSize = True
         Me.lblData.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.lblData.Location = New System.Drawing.Point(27, 29)
+        Me.lblData.Location = New System.Drawing.Point(19, 29)
         Me.lblData.Margin = New System.Windows.Forms.Padding(2, 0, 2, 0)
         Me.lblData.Name = "lblData"
         Me.lblData.Size = New System.Drawing.Size(30, 13)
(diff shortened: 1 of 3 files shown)


dannyparsons:
- checkbox for include arguments?
- the order of buttons probably needs discussion as I would assume things like `t-test` are most common.
- Not sure what the R Package: label is for, are they not functions?

shadrackkibet:
> checkbox for include arguments?

I have it at the top.
> the order of buttons probably needs discussion as I would assume things like t-test are most common.

This is for a start , i am to happy change to something sensible after discussion.
> Not sure what the R Package: label is for, are they not functions?

@rdstern  suggested we have this label somewhere. Maybe i didn't get it correctly!. Should this be the label of a combobox then we have  packages listed here?, which will then lead to different keyboards? Roger can clarify this.

shadrackkibet:
@africanmathsinitiative/developers  this is ready for testing.

dannyparsons:
I think @rdstern will have comments on the layout from the image above.

rdstern:
Exciting.  

1. On layout I think we may later want to change the order on the main keyboard.  But I'd like to get it working and try using it for now.
2. Could the second (little) keyboard go below the first, so the dialogue doesn't have to be so wide.
3. The Package could be a drop-down, or a label for now saying stats, because all those significance tests are from the stats package.  Later there will be other packages with their own significance tests that we may wish to add.  For now, I'll try just writing the command for other packages as we sometimes do in calc.

shadrackkibet:
@africanmathsinitiative/developers this is still ready for review.

dannyparsons:
I think Roger can best test this once it is merged. Is it accessible from frmMain?

shadrackkibet:
This is not yet accessible on the frmmain. I am going to add that now.Probably down on the models menu?

dannyparsons:
anywhere around there is fine to start with

rdstern:
This is looking very promising.  It isn't easy to use, but we never thought it would be.  It gives access to many R commands in one go, and is part of providing more of the power of R, for those willing to take this extra step!

1. On its position in the menu I was thinking of the first, rather than the last item, below the line in the model menu.   And it could be called Hypothesis Test on the menu.
2. I am sure we will want to edit the keyboard at some stage.  But (for now) I am happy with that as it is.
3. I was going to suggest getting rid of the **Try**.  But I have now found that when I do the t-test with no column, then click on Try, it tells me I have made an error.  I like that.  When I add a column it just gives we the 3 dots.  And even that is useful.
4. The By isn't working yet.  This is considered in #4386.  Is is possible to add this powerful feature?
5. I have yet to try to tests in turn.  So more to come!

PR #4417, opened 2018-02-21
changed defaults for codes in display daily

dannyparsons:
@shadrackkibet This is now ready to be implemented in the Display Daily dialog. The parameters are slightly different but similar. Note that the margin summaries (`monstats`) can now be a list of summaries.

There are no defaults for the codes because they are optional. But I suggest the dialog defaults are:
```
Misscode = "-"
Tracecode = "tr"
Zerocode = "--"
```


shadrackkibet:
Thanks, I will implement this in the dialog.

PR #4428, opened 2018-02-22
Display daily climatic data

shadrackkibet:
@africanmathsinitiative/developers this is now ready for review. I have implemented Display daily climatic data function. To test this effectively  filter your data first.


dannyparsons:
I can review this.

instat/dlgDisplayDailyData.vb
```diff
@@ -101,6 +101,7 @@ Public Class dlgDisplayDailyData
         dctSummary.Add("max", Chr(34) & "max" & Chr(34))
         dctSummary.Add("IQR", Chr(34) & "IQR" & Chr(34))
         dctSummary.Add("sum", Chr(34) & "sum" & Chr(34))
+        ucrInputComboMonStat.SetRDefault(Chr(34) & "min" & Chr(34))
```
dannyparsons: This isn't the R default. You can set a default for the dialog, in the default function, and sum would be the most sensible.

dannyparsons:
This is looking good.

The parameters for missing, trace and zero values need to be added to the dialog. 
I think they should each be a checkbox, with a combo box, but editable so values can be typed in.

instat/dlgDisplayDailyData.vb
shadrackkibet: ok, i will remove this.

shadrackkibet:
>The parameters for missing, trace and zero values need to be added to the dialog.
I think they should each be a checkbox, with a combo box, but editable so values can be typed in.

Noted i will implement this.

shadrackkibet:
Must we have this checkboxes?. What about having a label for each combobox.

dannyparsons:
They need to be optional, you may not want to replace zeros with a code, for example, so that's why I thought checkboxes.

shadrackkibet:
Done.

dannyparsons:
Looks good. I will merge so Roger can continue testing.
Roger can suggest more common options for the drop downs for the codes.
The function also allows for multiple summaries so that could check from a drop down to checkboxes.

rdstern:
Great

rdstern:
Here are my comments on a first test of the new display daily data dialogue:
1) It works.  Great that we have it.  Thanks.  
2) It gives an error before the dialogue loads - saying: 

instat

Developer error: no state of control ucrInputComboMonStat satisfies it's condition. Cannot determine how to set the control from the RCode. Modify control setup so that one state can satisfy its conditions.

OK   
3.  Then the dialogue loads fine.  Only do items below that are quick.
4. I first chose sum and was then pleased that I got multiple summaries - plus the sum.  Given multiple summaries are possible it would be good to be able to choose more  than 1.  Later, when I chose median, I only got the median, etc.
5. Is there the possibility of getting a count of the missing values in each month?
6. Is there the possibility of getting annual summaries also?
7. I did multiple summaries (Tmax and Tmin).  It did that, but with all the Tmax followed by all the Tmin.  If it only does one order, then I would prefer Tmax and then Tmin for each year.  (I can always get all the Tmax, followed by all the Tmin, by running the dialogue twice.)
8. The auto-filling does not yet include the station or the year columns.
9. I like the options to specify the code for missing. I also like the fact I can type into there.  We could have more options in the drop down.  I suggest the default could be m, and we should also suggest NA..
10. The default, when I reset is for the Missing, Trace and Zero all to be ticked.  I suggest missing be ticked and the other two be unticked.  (They only apply really for rainfall anyway.)
11.  I haven't checked - what happens with missing values if that option is not ticked?

rdstern:
It is slow to provide the results.  Is this:
a) Just the way it has to be?
b) the way it is programmed?
c) Because the summary statistics (i.e. the mean, median) take a lot of time?
d) Because it is writing to the output window?

I wonder of the possibility/ease of giving an option to write to a file?  Then in the output window giving a link to that file?

And this may become a more general option for commands that produce quite a lot of output? That is a sort of table equivalent to the options for where the graphical output goes?

rdstern:
I have now looked again at Helen's original function.  To me this looks well programmed, i.e. it is written by someone talented, but who would not claim to be a professional programmer.  I would guess it is written at least as well as most code to do climatic analyses.  So it is a great example for us to be able to discuss what might be different in this sort of function to one in R-Instat.

I can see that a few of the things I asked for above could be easy, and others are likely to be more difficult - and probably therefore shouldn't be done.

1. Adding the number of missing values each month as another summary statistic should be very easy?
2. Getting annual summaries (as well as monthly) is not in her code.  Let's not even start that.
3. Changing the order (Station, then Element then Year) seems pretty fixed in her code.  Let's leave that for now.
4. The code for trace and zero is anyway only for rainfall columns in her code.  If that is the same for us, then that's fine
5. Now a big one we could perhaps add - which isn't in her code.  Could we have an option of an additional logical column in the dialogue.  Then (like giving blank for zero) we give blank or a different code if that column is FALSE.  So you only get the number if that extra column is true.  Otherwise you get the code that was selected, e.g. blank.  
This option would be used together with our procedures for quality control.  There we can filter to isolate the suspect days.  We now use that filter (back on the original data frame) to be able to highlight those days in a nice - easy to examine format!

It may be that there is an easier way to do this step.  So perhaps there should be some discussion first.

shadrackkibet:
I will be resolving some of this issues.

shadrackkibet:
It will be good if we have this in the original issue #3067

shadrackkibet:
> It gives an error before the dialogue loads - saying:
instat
Developer error: no state of control ucrInputComboMonStat satisfies it's condition. Cannot determine how to set the control from the RCode. Modify control setup so that one state can satisfy its conditions.

I will sort out this.

> I first chose sum and was then pleased that I got multiple summaries - plus the sum. Given multiple summaries are possible it would be good to be able to choose more than 1. Later, when I chose median, I only got the median, etc.

I couldn't replicate this.Perhaps you could give more clear description of what you did exactly.

>  I did multiple summaries (Tmax and Tmin). It did that, but with all the Tmax followed by all the Tmin. If it only does one order, then I would prefer Tmax and then Tmin for each year. (I can always get all the Tmax, followed by all the Tmin, by running the dialogue twice.)

Actually, we should have a single receiver for the element instead of multiple receiver.

> The auto-filling does not yet include the station or the year columns.

The autofilling works perfectly for me.It will be good if you give a detailed description of what you did.

>  I like the options to specify the code for missing. I also like the fact I can type into there. We could have more options in the drop down. I suggest the default could be m, and we should also suggest NA..

I can add this options what about the existing option i.e `-` , should we remove it and have this two options(NA ,m)?

>  I haven't checked - what happens with missing values if that option is not ticked?

currently it gives empty spaces .

rdstern:
"I first chose sum and was then pleased that I got multiple summaries - plus the sum. Given multiple summaries are possible it would be good to be able to choose more than 1. Later, when I chose median, I only got the median, etc.

I couldn't replicate this.Perhaps you could give more clear description of what you did exactly."

What I would prefer is an option to choose the summaries from the list - perhaps a set of check boxes.  With mis added to the list, so we could get a count of missing values.

"I haven't checked - what happens with missing values if that option is not ticked?

currently it gives empty spaces ."

I don't think we should have <blank> as the default for missing.  The obvious one is NA, as that is missing in R.  I am happy that you can type into the field to choose your own code for missing.  Perhaps there should be a check that it isn't too wide?

"The auto-filling does not yet include the station or the year columns.

The autofilling works perfectly for me.It will be good if you give a detailed description of what you did."

I did have a file where the climatic define was used before.  But there were two year-type columns, one called year and the other called year1.  Likewise there were 2 acceptable identifiers for the station, though one of the had been designated in the define dialogue.

shadrackkibet:
Should we automatically fill in the element receiver? If so with which element?

rdstern:
I would like that, but I don't think it is so easy, because it could apply to any element.  If rainfall were the default initially, but it could be replaced by any other manually, then I would go for rainfall.

shadrackkibet:
Adding one element is easy, then others could be added manually.

rdstern:
I note the original issue has just been closed.
I can accept the display as it is now.
But the one addition that would enhance it, (and is in Helen's code) is to permit multiple summaries on a monthly basis.  So, instead of a pull-down where you choose one summary you have check-boxes.
And the one addition to those summaries, which I think may be trivial is to add the count of missing values.  I think (as that is just a statistic , then adding that code - possibly mis (plus the checkbox) - would enable that feature.

PR #4430, opened 2018-02-26
Adding Dates option to Regular sequence.

shadrackkibet:


shadrackkibet:
@dannyparsons  this is now working properly.

rdstern:
Great that this is now working.  Very useful addition.  At some stage it could be useful to add it to the Prepare > Column: Dates menu and also possibly to the Climatic > Dates menu similarly.  There the Dates would be the initial default, while in Prepare > Column: Generate it would remain as now.

rdstern:
Absolutely.  And ideally there the dates are the initial default for the radio button.

shadrackkibet:
@dannyparsons please merge this, then we can see how we can have different defaults on the menu.

instat/dlgRegularSequence.vb
```diff
@@ -56,71 +61,127 @@ Public Class dlgRegularSequence
...
-        ucrNudRepeatValues.SetLinkedDisplayControl(lblRepeatValues)
-        'ucrPnlSequenceType.AddToLinkedControls(dtpSelectorA, {rdoDates}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
-        'ucrPnlSequenceType.AddToLinkedControls(dtpSelectorB, {rdoDates}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrPnlSequenceType.AddToLinkedControls(ucrDateTimePickerFrom, {rdoDates}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=Chr(34) & "2018/1/1" & Chr(34))
+        ucrPnlSequenceType.AddToLinkedControls(ucrDateTimePickerTo, {rdoDates}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=Chr(34) & "2018/12/31" & Chr(34))
+        ucrPnlSequenceType.AddToLinkedControls(ucrInputComboDatesBy, {rdoDates}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:="days")
```
dannyparsons: I think we don't want default states here. It shouldn't be needed, because there are always values to read from in the function since we now have separate `seq` functions. We don't want the options to reset when you change between numbers and dates.
Is there is issue not having this?

shadrackkibet: There are developer errors when on Numeric radiobutton then you reopen the dialog. Having this default states fixes the issue, However i am not sure if this is the right way to tackle this problem.
image
image

dannyparsons: I believe its because you have `bNewLinkedAddRemoveParameter:=True` so all the parameters are getting removed when you change to numeric. If you only have it doing the hiding and nothing else that should work.

shadrackkibet: Thanks, this work now.

PR #4443, opened 2018-03-08
Added geoms to ucrGeom 

bethanclarke:
Progress so far: 
geom count is working 
geom encircle is not because the code is trying to get the command from ggplot2 not ggalt
Still need to test geom segment and curve but not sure what to do with the x, y, xend and yend parameters
geom area and geom tile have been updated 

Notes on some issues:
Geom_count we need to be able have options for the size parameter within the aes function e.g. geom_count(aes(size = ..prop..))

Geom_encircle needs to be looked at more closely but should work okay. Documentation is not very helpful, but here are a couple of useful links;
http://r4ds.had.co.nz/visualize.html
https://github.com/hrbrmstr/ggalt/blob/master/R/geom_encircle.r

Geom segment and curve have been added. We want to assign variables that only have one observation to the following parameters within aes function: x, y, xend, yend. So we need ensure you cannot select a variable that contains more than one observation, otherwise we get a error in R.


maxwellfundi:
You could find more info on geom_encircle here- https://cran.r-project.org/web/packages/ggalt/vignettes/ggalt_examples.html 
Does this geom has mandatory aesthetics? they should be bold. Further, i find that only linetype accepts only factors, could you change the selector heading to factors. 

Another comment would be, geom count looks good.

bethanclarke:
@maxwellfundi thanks for your comments. Geom encircle does have mandatory aesthetics, so I've changed that now. Linetype cannot be mapped to a continuous variable, but the other optional aes can be mapped to numeric and factors

dannyparsons:
Let us know when you're done with this changes and ready to start a new branch

bethanclarke:
I am ready to start a new branch. I just had to change some aesthetics parameters to not mandatory

PR #4451, opened 2018-03-13
Summaries Subdialog

shadrackkibet:
@africanmathsinitiative/developers Proportions ,Counts and Percentiles are now working.
This is ready for review.


shadrackkibet:

maxwellfundi:
Will test this soon. :-)

maxwellfundi:
@shadrackkibet  this looks good and works well from Prepare>Column Reshape > Column Summaries. However, since the summaries sub dialog is used even in Describe > One Variable> Summarise, i find that it does not work there because these dialogs use different functions.  The Prepare>Column Reshape > Column Summaries uses the `calculate_summary` while the Describe > One Variable> Summarise uses the `summary` function. 
@rdstern  @dannyparsons  I wonder if we can use the `calculate_summary` in both dialogs so that this works consistently.

maxwellfundi:
@shadrackkibet what is the status of this?

shadrackkibet:
There is an issue with passing` proportions , percentiles` and `quantile` parameters into ` summary` function  which should be sorted out. Otherwise `calculate_summary` is working fine.

dannyparsons:
@shadrackkibet just to check that `proportions`, `percentiles` and `quantile` are not under "Two Variables" tab. Roger suggested they go at the top of the "More" tab I think.

dannyparsons:
For the `summary` function I can correct this for those three in a separate pull request.
@maxwellfundi yes the correct thing to do is use `calculate_summary` on all the dialogs then we won't have these issue. `summary` is an older function we wrote which needs to be replaced.

shadrackkibet:
> @shadrackkibet just to check that proportions, percentiles and quantile are not under "Two Variables" tab. Roger suggested they go at the top of the "More" tab I think.

This are at the bottom of the "More Tab". I am going to move them up now.

shadrackkibet:
This is how **More** & **2-Variables** Tabs looks now.
**More Tab**
image

image

dannyparsons:
Sorry, can you also move First, nth, Last, n_distinct to the bottom of the More tab in it's own box.

shadrackkibet:
> Sorry, can you also move First, nth, Last, n_distinct to the bottom of the More tab in it's own box.

I thought `First, nth `and` last` will sometimes require a vector to `order_by `which will be from a receiver in the 2-variable Tab.

rdstern:
I agree with the move for now.
I also like the order_by option though.  So, once we have more time, then I suggest (on the 2-variable tab) we opt for either Correlations or First/Last.  I don't think it is much of a hardship to have to use the dialogue twice if you happened to want both!
A third option on this tab - later - may be to save a related column for some summaries - like the day of the maximum.

dannyparsons:
Maybe **Position**?

`n_distinct` is actually the only one that isn't related to a position - and it doesn't have the `order_by` argument. So **Position** is good title and `n_distinct` can go in the first box on the first tab with the other counting summaries.

shadrackkibet:
Done now.

dannyparsons:
@shadrackkibet The alignment of controls with the checkboxes is not right. I can't tell which controls are for which summary. Can you tidy these up?
image
and
image

Also from the One Variable dialog instead of disabling the controls on the 2 Variables tab, it would be better to hide the tab completely.

shadrackkibet:
ok, i am using the same controls for both proportion and count. I will then have to have different controls for this two checkboxes to fix alignment.

Issue #4435, opened 2018-03-03
The View R Objects dialogue


rdstern:
The recent Hypothesis tests dialogue has produced more R Objects.

It would be good if the Prepare > R Objects > View dialogue had more options available.  Currently all options are disabled except graphs.

rdstern:
Would this be a dialogue Patrick could look at?  I hope that, just checking that the str() command can be given would allow it to be useful for viewing.  And we should have the option to delete objects as well.  They used to work I think?

There are lots of ways the dialogue could be made more useful, but only if they can (at this stage) be done quickly.  Here is my initial list:
a) Can some information about the selected object be easily available within the dialogue, e.g. number of elements in the list?  The object type perhaps?
b) The data frame is itself an object.  Should this be included?
c) Perhaps the listviewer could be an option, i.e. listviewer::jsonedit(object).  This shows the content in an browser in a way that the levels of items in the list can be "unfolded".

Patowhiz:
Let me have a look at it.

dannyparsons:
The simple thing to get working first is just printing the object 'as is' in the output window. For certain objects things like structure and list items are very long and generally not useful whereas printing normally does something sensible. If the printing was working it would already be useful and other options can be added later.

Patowhiz:
@dannyparsons I'm making that to fall under the "All contents" option.

Patowhiz:
@dannyparsons what's your opinion about the data frame itself also being viewed as an object. Should this be included?

dannyparsons:
- Deleting objects should be in a separate dialog
- I don't think data frames should be viewable here - we are looking at objects associated with a data frame and we already have a view dialog for data frames.

There's lots of interesting functions we could think of using here which could be discussed. But I think the priority now is to get something that's useful and works. So I suggest two options which are easy to get working:
"Print" which just gets the object so it displays in the output window in the default way (no need for "View Graph" radio button as a separate option as this is what "Print" should do for graphs)
"Structure" which calls `str` on the object

Print should be the default and first option.

The idea of Component is to view an item (component) from the object instead of the whole object. For example, an `lm` object has a `coefficients` component. This radio button could be disabled for now as it will need some R code to be written to get working well.

Patowhiz:
@dannyparsons are you suggesting that I make the All contents radio button to be Print?

dannyparsons:
I'm suggesting a radio button called "Print" which just gets the object, calling the get_objects method.
[...]

PR #4469, opened 2018-03-27
View Objects

Patowhiz:


instat/dlgViewObjects.vb
```diff
@@ -61,7 +61,7 @@ Public Class dlgViewObjects
 
         'we are disabling this for now until they're working correctly.
         rdoStructure.Enabled = False
-        rdoAllContents.Enabled = False
+        'rdoAllContents.Enabled = False
```
maxwellfundi: Is this all what was needed for this dialog @Patowhiz ?

Patowhiz:
@africanmathsinitiative/developers this is ready for review. Fixes issue #4435  The View R Objects dialogue.

@africanmathsinitiative/developers  @dannyparsons  I need help in how to determine if the object selected is really a graph so as to present the appropriate feedback to the user, currently if you try viewing an object that is not a graph as a graph, an error message pops up.

Patowhiz:
@africanmathsinitiative/developers  @dannyparsons I'm thinking of running a script to get the graphs and checking if there is a graph with the selected object name. If the returned graphs have a graph name similar to the selected object then it gets shown otherwise we display a feedback that the selected object is not a graph. What's your opinion on this?

dannyparsons:
My suggestion on the issue means there isn't a need for the "View Graph" radio button. But you will still probably need to check if the selected item is a graph because this will affect the call type.

The easiest was is probably to run `get_graph_names` internally when the data frame is changed and then when the selected item is changed, check if it is in the list of graph names.

shadrackkibet:
reviewing and testing this now.

instat/dlgViewObjects.vb
```diff
@@ -21,6 +21,7 @@ Public Class dlgViewObjects
     Private clsGetObjectsFunction As RFunction
     Private clsShowObjectStructureFunction As RFunction
     Private clsShowObjectContentsFunction As RFunction
+    Private clsShowGraphFunction As RFunction
```
shadrackkibet: I think this should be ` As New RFunctions` if not why?

instat/dlgViewObjects.vb
```diff
@@ -86,14 +88,18 @@ Public Class dlgViewObjects
         clsShowObjectContentsFunction.SetRCommand("show")
         clsShowObjectContentsFunction.AddParameter(New RParameter("object", clsGetObjectsFunction, iNewPosition:=0))
 
+        'set R function for showing graph
+        clsShowGraphFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_graphs")
+        clsShowGraphFunction.AddParameter("print_graph", "TRUE")
```
shadrackkibet: I think you need parameter position here.

instat/dlgViewObjects.vb
```diff
@@ -60,32 +61,40 @@ Public Class dlgViewObjects
...
     Private Sub SetDefaults()
-        clsViewObjectFunction = New RFunction
+        clsGetObjectsFunction = New RFunction
+        clsObjectStructureFunction = New RFunction
 
         ucrSelectorForViewObject.Reset()
-        rdoViewGraph.Checked = True
-        SetiCallType()
+        rdoStructure.Checked = True
+        SetICallType()
```
shadrackkibet: why do you have this on setDefaults(). I thought this is only required in the panel control contents changed event

instat/dlgViewObjects.vb
```diff
@@ -60,32 +61,40 @@ Public Class dlgViewObjects
...
 
     Private Sub SetDefaults()
-        clsViewObjectFunction = New RFunction
+        clsGetObjectsFunction = New RFunction
+        clsObjectStructureFunction = New RFunction
 
         ucrSelectorForViewObject.Reset()
-        rdoViewGraph.Checked = True
-        SetiCallType()
+        rdoStructure.Checked = True
```
shadrackkibet: I think you should not be doing this here. This can fixed by setting Rcode for the panel

instat/dlgViewObjects.vb
```diff
@@ -60,32 +61,40 @@ Public Class dlgViewObjects
...
+        ucrReceiverSelectedObject.SetRCode(clsGetObjectsFunction, bReset)
     End Sub
 
     Private Sub TestOKEnabled()
-        If Not ucrReceiverSelectedObject.IsEmpty Then
-            ucrBase.OKEnabled(True)
-        Else
-            ucrBase.OKEnabled(False)
-        End If
+        'If Not ucrReceiverSelectedObject.IsEmpty Then
```
shadrackkibet: why is this commented out?

instat/dlgViewObjects.vb
```diff
@@ -120,19 +126,32 @@ Public Class dlgViewObjects
...
+
+    Private Sub ucrBase_BeforeClickOk(sender As Object, e As EventArgs) Handles ucrBase.BeforeClickOk
+        SetGraphParameters()
+    End Sub
+
+    'This is being called in several place because we don't have a BeforeToScripts Click event
+    Private Sub SetGraphParameters()
+        If rdoViewGraph.Checked Then
+            clsShowGraphFunction.AddParameter(strParameterName:="data_name", strParameterValue:=Chr(34) & ucrSelectorForViewObject.strCurrentDataFrame & Chr(34), iPosition:=0)
+            clsShowGraphFunction.AddParameter(strParameterName:="graph_name", strParameterValue:=ucrReceiverSelectedObject.GetVariableNames(bWithQuotes:=True), iPosition:=1)
```
shadrackkibet: I think you should have and else to remove this manually added parameters.

instat/dlgViewObjects.vb
```diff
@@ -52,39 +53,47 @@ Public Class dlgViewObjects
...
-        clsObjectStructureFunction.SetRCommand("str")
-        clsObjectStructureFunction.AddParameter(New RParameter("object", clsGetObjectsFunction, iNewPosition:=0))
+        'set R function for showing selected object structure
+        clsShowObjectStructureFunction.SetRCommand("str")
+        clsShowObjectStructureFunction.AddParameter(New RParameter("object", clsGetObjectsFunction, iNewPosition:=0))
 
-        ucrBase.clsRsyntax.SetBaseRFunction(clsObjectStructureFunction)
+        'set R function for showing selected object contents
+        clsShowObjectContentsFunction.SetRCommand("show")
+        clsShowObjectContentsFunction.AddParameter(New RParameter("object", clsGetObjectsFunction, iNewPosition:=0))
```
shadrackkibet: I think this should be  ` clsShowObjectContentsFunction.AddParameter("object", clsRFunctionParameter:= clsGetObjectsFunction, iPosition:=0))`

instat/dlgViewObjects.vb
```diff
@@ -52,39 +53,47 @@ Public Class dlgViewObjects
...
         SetICallType()
 
+        'set R function for getting objects
         clsGetObjectsFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_objects")
 
-        clsObjectStructureFunction.SetRCommand("str")
-        clsObjectStructureFunction.AddParameter(New RParameter("object", clsGetObjectsFunction, iNewPosition:=0))
+        'set R function for showing selected object structure
+        clsShowObjectStructureFunction.SetRCommand("str")
+        clsShowObjectStructureFunction.AddParameter(New RParameter("object", clsGetObjectsFunction, iNewPosition:=0))
```
shadrackkibet: Also here.

instat/dlgViewObjects.vb
Patowhiz: Cause I'm already initialising that in the SetDefaults subroutine. Removed unnecessary repetition

instat/dlgViewObjects.vb
Patowhiz: Why dp you think should I remove them? Right now, I have 4 RFunctions doing separate things. You can have a look at them in my declaration. The clsShowGraphFunction is specifically for showing graphs and is called iff the selected object is a graph

instat/dlgViewObjects.vb
Patowhiz: I didn't see a reason for this because it's the last parameter which means it's position will be set as -1 automatically making it to be the last one, since the other 2 have their position set. I can set this as 3 if that makes it clear. I prefered the -1 because technically in the rlink its position is not 3 but it will still work .

instat/dlgViewObjects.vb
Patowhiz: Yeah, you are right. There is no need to call SetICallType() here

instat/dlgViewObjects.vb
Patowhiz: The new line added  ucrBase.OKEnabled(Not ucrReceiverSelectedObject.IsEmpty) does exactly what the commented lines do. 5 lines of code reduced into 1 line to achieve the same functionality

instat/dlgViewObjects.vb
Patowhiz: Well I didn't see the necessity of that. If you check on the RParameter constructor used here then you will realise that the first 2 parameters are NOT optional. Which means that they have to be there and in the order shown.  And you can see I'm using the iNewPosition:=0 because that is an optional parameter. VB.NET compiler would have complained if the two were optional and set this way. Generally, we do things like iNewPosition:=0 for parameters that are optional to tell the compiler which parameter to set the passed variable. I hope this makes it understandable.

instat/dlgViewObjects.vb
Patowhiz: Currently the panel has no Rcode. I could be wrong with my implementation.

Patowhiz:

maxwellfundi:
@dannyparsons @Patowhiz  I have been testing this and I have been wondering if we need the to have a structure for graphs. it produces so much output that am wondering if its useful.
Further i have noticed that I can only print graphs that are ggplot object.

Patowhiz:
@maxwellfundi @dannyparsons I wasn't quite sure about the output expected when it comes to the structure of the graph. But that is what str outputs for objects currently.

dannyparsons:
It's not particularly useful for graphs but the dialog is more likely to be used for other objects anyway as we already have a view graph dialog, so I don't think there's much harm in having it for all objects. We wouldn't encourage people to look at graphs in this dialog in most cases but its there if anyone would like.

For non ggplot graphs, is this the same from the View Graph dialog? i.e. is it a problem with this dialog or more general?

Patowhiz:
Its general not only this dialog

instat/dlgViewObjects.vb
dannyparsons: We should set the position for all parameter, as good practice. It doesn't really make sense to want a parameter to have no position (-1)

instat/dlgViewObjects.vb
dannyparsons: The panel's value is the radio button in the panel which is checked, so it is dependent on the R code. This can be set through using the conditions to state how to detect which radio button to check from the R code.

instat/dlgViewObjects.vb
dannyparsons: If code has been replaced it can be removed, no need to keep it commented.

instat/dlgViewObjects.vb
```diff
@@ -48,44 +51,47 @@ Public Class dlgViewObjects
...
 
-        ucrPnlContentsToView.bAllowNonConditionValues = True ' temporary
-        ' rdo's
-        'ucrPnlContentsToView.SetParameter(New RParameter("", 2))
-        'ucrPnlContentsToView.AddRadioButton(rdoStructure, Chr(34) & "" & Chr(34))
-        'ucrPnlContentsToView.AddRadioButton(rdoAllContents, Chr(34) & " " & Chr(34))
-        'ucrPnlContentsToView.AddRadioButton(rdoComponent, Chr(34) & " " & Chr(34))
-        'ucrPnlContentsToView.AddRadioButton(rdoViewGraph, Chr(34) & " " & Chr(34))
-        'ucrPnlContentsToView.SetRDefault(Chr(34) & "" & Chr(34))
+        ucrPnlContentsToView.bAllowNonConditionValues = True
```
dannyparsons: This should not be True. Through setting conditions you can get the panel to determine the correct radio button to check.
@shadrackkibet has done this for many dialogs if you need help

instat/dlgViewObjects.vb
```diff
@@ -48,44 +51,47 @@ Public Class dlgViewObjects
...
-        clsViewObjectFunction = New RFunction
+        'initialise the Rfunctions
+        clsGetObjectsFunction = New RFunction
+        clsShowObjectStructureFunction = New RFunction
 
+        'reset controls to default states
         ucrSelectorForViewObject.Reset()
-        rdoViewGraph.Checked = True
-        SetiCallType()
+        rdoPrint.Checked = True
```
dannyparsons: When you have the conditions set correctly you won't need to have this.

instat/dlgViewObjects.vb
```diff
@@ -94,19 +100,54 @@ Public Class dlgViewObjects
...
+        clsGetGraphNames = New RFunction
+        clsGetGraphNames.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_graph_names")
+        clsGetGraphNames.AddParameter(strParameterName:="data_name", strParameterValue:=Chr(34) & ucrSelectorForViewObject.strCurrentDataFrame & Chr(34), iPosition:=0)
+
+        expItems = frmMain.clsRLink.RunInternalScriptGetValue(clsGetGraphNames.ToScript(), bSilent:=True)
+
+        If expItems IsNot Nothing AndAlso Not (expItems.Type = Internals.SymbolicExpressionType.Null) Then
+            strArr = expItems.AsCharacter.ToArray
+            If strArr IsNot Nothing AndAlso strArr.Length > 0 Then
+                If strArr(0) = ucrReceiverSelectedObject.GetVariableNames(bWithQuotes:=False) Then
```
dannyparsons: What you should be checking here is whether `strArr` contains `ucrReceiverSelectedObject.GetVariableNames(bWithQuotes:=False)`
Here you are only checking if the first item is the selected object.

instat/dlgViewObjects.vb
```diff
@@ -94,19 +100,54 @@ Public Class dlgViewObjects
...
+        Dim strArr As String()
+
+        clsGetGraphNames = New RFunction
+        clsGetGraphNames.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_graph_names")
+        clsGetGraphNames.AddParameter(strParameterName:="data_name", strParameterValue:=Chr(34) & ucrSelectorForViewObject.strCurrentDataFrame & Chr(34), iPosition:=0)
+
+        expItems = frmMain.clsRLink.RunInternalScriptGetValue(clsGetGraphNames.ToScript(), bSilent:=True)
+
+        If expItems IsNot Nothing AndAlso Not (expItems.Type = Internals.SymbolicExpressionType.Null) Then
+            strArr = expItems.AsCharacter.ToArray
```
dannyparsons: This is all fine, but I mentioned that there is already a `GetGraphNames` function in the `RLink` which would have saved you having to write the function and check the check the expression yourself.  Since you've done it now, no need to change it unless you want to.

instat/dlgViewObjects.vb
Patowhiz: I don't know how I missed this

instat/dlgViewObjects.vb
Patowhiz: I thought I'm already using the function by setting the RCommand to $get_graph_names, that was my understanding . Perhaps you could help me grasp an alternative better way of using the GetGraphNames.

instat/dlgViewObjects.vb
Patowhiz: @shadrackkibet help me with this. I have redone it in my latest code upload check to see if I have done it correctly. This is what I wanted; the panel to check/select the `rdoPrint` when the base RFunction is `clsGetObjectsFunction` and to check `rdoStructure` when the base RFunction is `clsShowObjectStructureFunction`. The radio buttons mentioned are not our custom controls. 
To achieve the above. This is what I did 

Inside subroutine `InitialiseDialog()`
 ```
ucrPnlContentsToView.AddFunctionNamesCondition(rdoPrint, frmMain.clsRLink.strInstatDataObject & "$get_objects")
        ucrPnlContentsToView.AddFunctionNamesCondition(rdoStructure, "str")
```

Inside `SetRCodeforControls(bReset As Boolean)`
```
 ucrPnlContentsToView.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
```

File: instat/dlgViewObjects.vb [hunk 14]
```diff
@@ -48,44 +51,47 @@ Public Class dlgViewObjects
         ucrReceiverSelectedObject.SetMeAsReceiver()
         ucrReceiverSelectedObject.strSelectorHeading = "Objects"
         ucrReceiverSelectedObject.SetItemType("object")
-        ucrReceiverSelectedObject.bAutoFill = True
+        'ucrReceiverSelectedObject.bAutoFill = True
```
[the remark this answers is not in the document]
Patowhiz: Done

instat/dlgViewObjects.vb
Patowhiz: Done

instat/dlgViewObjects.vb
Patowhiz: @shadrackkibet could you help me understand how I could utilise the function that @dannyparsons  is mentioning above from the RLink

Patowhiz:

Patowhiz:
Okay

shadrackkibet:
I have tested this dialog and it is working fine. Just work on Tab order and this should be ready for @dannyparsons

instat/dlgViewObjects.vb
dannyparsons: Did you want to swap this for the RLink method or leave as is? Either works.

Patowhiz:
Just Leave it as is.

PR #4476, opened 2018-04-03
Day of year range subdialog

dannyparsons:
Implemented on climatic summaries dialog

### DIFF
--- a/instat/UcrPanel.vb
+++ b/instat/UcrPanel.vb
@@ -81,7 +81,7 @@ Public Class UcrPanel
         End If
     End Sub
 
-    Protected Overrides Sub SetToValue(objTemp As Object)
+    Public Overrides Sub SetToValue(objTemp As Object)
         Dim rdoTemp As RadioButton
 
         If objTemp IsNot Nothing Then
--- a/instat/dlgClimaticSummary.Designer.vb
+++ b/instat/dlgClimaticSummary.Designer.vb
@@ -47,18 +47,10 @@ Partial Class dlgClimaticSummary
         Me.lblYear = New System.Windows.Forms.Label()
         Me.lblDay = New System.Windows.Forms.Label()
         Me.lblElement = New System.Windows.Forms.Label()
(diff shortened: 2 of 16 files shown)


dannyparsons:
@shadrackkibet this can be checked and tested - it will need some improvement before its ready

dannyparsons:
@shadrackkibet have you tested this? This is the most urgent work to be in the next version.

shadrackkibet:
I am on it now.

shadrackkibet:
This is working fine;
 Just one question and designer suggestions;
1. Why don't we have fixed difference in From box?
2. Some designer fixes i suggest is to have the receivers and other controls aligned to the respective radio buttons on Both From and To. Also colons for the label?, In the main dialog some part of the group box is hidden also.
3.The size of the dialog could be adjusted depending on which radio button is checked to reduce the amount of empty space when the selector is hidden, But i think this could come later?

Otherwise this is working well and could be improved in future , i think.

dannyparsons:
> Why don't we have fixed difference in From box?

How would this work in the filter? See how fixed difference works for the To and it should be clear why it isn't in From.

> Some designer fixes i suggest is to have the receivers and other controls aligned to the respective radio buttons on Both From and To. Also colons for the label?, In the main dialog some part of the group box is hidden also.

Great, could you make these changes, either on this branch or after merging?

> The size of the dialog could be adjusted depending on which radio button is checked to reduce the amount of empty space when the selector is hidden, But i think this could come later?

Yes at the moment the layout isn't great and empty space could be reduced. I agree this can come later.

shadrackkibet:
I just noticed something. When you open the  subdialog and click on Variable Day rdo button then go back to the main dialog it resets to Fixed Day rdo. This happens for both From and To. However, when you add a variable to the receiver this doesn't happen.

dannyparsons:
I think that's ok because if you haven't selected a column it's ok for it to go back to a fixed number.

I think we just need to add a line to "Reset" the selector on first load to get it to display.

PR #4477, opened 2018-04-04
Adding undo and rearranging menu

maxwellfundi:
@rdstern  @dannyparsons  here is how the menu looks.  Redo is currently disabled

image


dannyparsons:
Why is redo disabled? 
I would have a line under Redo to separate it from the others.
Change "Open Script from File" to "Load Script from File"
Change "Open Script in RStudio/Text Editor" to "Open Script as File"
Add "Run All Text" under "Run Selected Text"

maxwellfundi:
Because am figuring out how to do it properly like what we have in other software. For redo, it  mainly  does
1. Last undone action
2. redoes the last typed in characters, like repeating typing. However, this functionality of redo I feel we might not need on the script window. Why would anyone want to produce many characters which are the same?

So I actually feel we can have the first functionality of this redo.

Am on the changes you suggested now.

maxwellfundi:
@dannyparsons  should we then remove the button at the top now that we are adding "Run All Text" under "Run Selected Text"? or should we keep both?

maxwellfundi:
The ReDo is not available for text boxes and even in the MSDN documentation. Looking around stackoverflow, the redo and undo are both implemented via Undo method. I suggest then we have the undo for now.

dannyparsons:
Ok fine, let's get rid of redo then.

maxwellfundi:
@africanmathsinitiative/developers  this is ready for review. Check the shortcuts work and the undo.

Muthenya:
Testing this now.

Muthenya:
I've noticed that the undo can only undo the last "undoable" action, which is okay. Just a thought, if necessary, that it be possible to undo more than one action, probably a series of actions as is the case in other programs.

Otherwise, this looks good. Good work Max.

instat/ucrScript.vb
```diff
@@ -188,14 +188,28 @@ Public Class ucrScript
...
-            mnuOpenScript.Enabled = False
+            mnuOpenScriptasFile.Enabled = False
             mnuClearContents.Enabled = False
             mnuSaveScript.Enabled = False
         End If
     End Sub
+
+    Private Sub Menu_Undo(sender As Object, e As EventArgs) Handles mnuUndo.Click
+        'Determine if last operation can be undone in text box.   
+        If txtScript.CanUndo = True Then
```
dannyparsons: Don't need `= True` here.

dannyparsons:
I don't see a need for this message box. This is how Paste works in other software so I found it annoying to have to confirm this. 
image

The other message box when pasting non text has a question mark at the end of the text which isn't needed.

Also, I tried Ctrl C and Ctrl Z in the script window and they didn't do anything. Ctlr V did work.

PR #4478, opened 2018-04-04
added check that forces R 3.4.4 or later to be used

dannyparsons:
@africanmathsinitiative/developers this ensures users have R 3.4.4 running which we need because R 3.4.3 has a bug. Please check this doesn't cause any unexpected errors.

### DIFF
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -71,6 +71,7 @@ Public Class RLink
         Dim expTemp As SymbolicExpression
         Dim strMajor As String = ""
         Dim strMinor As String = ""
+        Dim iMinor2 As Integer
         Dim iCurrentCallType As Integer
         Dim bClose As Boolean = False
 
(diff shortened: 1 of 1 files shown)


shadrackkibet:
Are you aware of this issue? I am using window 10 and my machine is Hp pavilion x360.I also have R-Studio version ‘1.1.423’ installed.

I installed two versions of R deliberately to check  i.e **R-3.4.0** and **R-3.4.4** when i load R-Instat it detected R-3.4.0 instead, i also noticed my Rstudio uses R-3.4.0. 

Here are my screenshots:
image

image

image

I always thought R-Studio should always use the latest version and so does our R-Instat?.

shadrackkibet:
image

Uninstalling **R-3.4.0** fixes the issue.

Should we include something in this message box or have it "elsewhere" to notify the users that they might have old versions of R installed and they might need to uninstall to remain with the latest version to fix the problem?

dannyparsons:
That's very interesting. It might be because you installed the older version later?

I think what happens is when you install R it updates the registry with that version. And RStudio and R-Instat are checking the registry. So it detects the last version installed.

Can you try uninstalling and installing both again but install R 3.4.4 second and see what happens.

maxwellfundi:
@shadrackkibet  yes, this happened to me sometime back as well. So the last one is the last installed version is what is used. I solved it by uninstalling and installing afresh.

dannyparsons:
Ok so this could be in help/documentation that this is a potential issue if you have multiple versions.

maxwellfundi:
yes, I agree completely it could be among the known issues

Regards
Maxwell Fundi
|| Phone :+254728508241||

Muthenya:
I have tried it as well, it works with latest installed version. I'm now running on R 3.4.4.

PR #4479, opened 2018-04-04
CurrentCulture fixed to English to ensure numbers are passed to R in correctly

dannyparsons:


instat/frmMain.vb
```diff
@@ -71,6 +71,11 @@ Public Class frmMain
         'temporary
         mnuHelpAboutRInstat.Visible = False
 
+        ' This must be fixed because CurrentCulture affects functions such as Decimal.TryParse
+        ' e.g. "1.0" fails Decimal.TryParse if CurrentCulture = "fr-FR" because it expects "1,0"
+        ' Decimal point must be `.` and not `,` because R only accepts `.`
+        Thread.CurrentThread.CurrentCulture = New CultureInfo("en-GB")
```
maxwellfundi: Am not sure we wanted en-GB, I think what we currently have is en-US

dannyparsons: Where do we have en-US? In the Options class we have en-GB.

maxwellfundi:
I see, that must have changed :-). Thought we had US

Regards
Maxwell Fundi
|| Phone :+254728508241||

dannyparsons:
We should probably get rid of the country specific language anyway but don't want to make big changes at this point.

PR #4481, opened 2018-04-05
Doy Range sdg

shadrackkibet:


dannyparsons:
I would prefer if the controls appear in the same place on the left side of the group box - they are alternatives so they don't need to appear in different places.

shadrackkibet:
Done now.

PR #4486, opened 2018-04-06
Ucrscript

maxwellfundi:


maxwellfundi:
@dannyparsons  this now fixes your issue of shortcuts but i have one more improvement to suggest
When pasting, with no selected text, I suggest we paste where we have the current cursor position not at the end of the text in the text box. I cant remember why we did it that way. 
What do you think?

dannyparsons:
Yes always paste at the cursor position. This should all work as Word does for example.

maxwellfundi:
@dannyparsons  thats now fixed. You can test this now.

instat/ucrScript.vb
```diff
@@ -130,34 +130,20 @@ Public Class ucrScript
     End Sub
 
     Private Sub mnuCopy_Click(sender As Object, e As EventArgs) Handles mnuCopy.Click
-        Clipboard.SetText(txtScript.Text)
+        Clipboard.SetText(txtScript.SelectedText)
```
dannyparsons: Why arent' you using textbox.Copy and textbox.Cut?

maxwellfundi: I could use that too, they work in same way.

dannyparsons: We should be consistent. In another place you are using textbox.Paste. I suggest we follow the documentation https://msdn.microsoft.com/en-us/library/system.windows.forms.textboxbase.copy(v=vs.110).aspx

maxwellfundi: Okay.

PR #4527, opened 2018-04-13
fixes bug causing extra boxplot layer added to graphs

dannyparsons:
@rdstern finally found the cause of this, couldn't find an issue on it to close
@maxwellfundi @bethanclarke I can explain the cause of this issue if you would like to know

### DIFF
--- a/instat/sdgLayerOptions.vb
+++ b/instat/sdgLayerOptions.vb
@@ -64,7 +64,11 @@ Public Class sdgLayerOptions
         clsGeomFunction = clsNewGeomFunc
         clsAesFunction = clsNewGlobalAesFunc
         clsGgplotFunction = clsNewGgPlot
+        ' This prevents ucrGeomWithAes affecting ucrLayerParameter before ucrLayerParameter.Setup has run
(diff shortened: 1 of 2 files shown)


rdstern:
Brilliant!

maxwellfundi:
Whats the issue here @dannyparsons I see you introduced a bPauseChanges boolean now

dannyparsons:
So in sdgLayerOptions.vb it first sets up the geom + aes control and then the geom + layer parameters control (the two tabs). When it sets up the aes control it sets a new geom function (boxplot for a new layer) and also in the aes control when the geom changes its linked to the layer parameters control so it also changes the geom on that layer. However during the setup aes has the new layer (boxplot) and layer parameters still has the previous layer (point if on scatter plot) hence the aes control changes the name of the geom in the layer parameters control from point to boxplot.
So the bPauseChanges tell the two controls not to affect each other until they are both set up, because during the set up one layer has the new geom and another has the old geom, and the old geom is one of the original layers of the graph so that's why that changes an original layer.
Not too difficult in the end but difficult to find.

PR #4542, opened 2018-04-17
Fixes on describe menu

### LINKED ISSUE #4537: Improving the Describe > Specific > Boxplot dialogue - jitter plot defaults
There are a number of issues with this dialogue. This issue is concerned with the defaults for the jitter plot component.  
1. In the instructions on jitter plots it says that if you don't give a value for the horizontal jitter, then it defaults to 40% (i.e. 0.4).  I would prefer (our) default to be smaller than this and suggest 0.2.
2. The vertical jitter also defaults to 0.4 and I would prefer our default be 0, i.e. no vertical jitter.

These will have to be given in the command when jitter is used. 

In the section where the changes can be made for this layer, i.e. the sub-dialogue choosing jitter, then our defaults should be given there, i.e. 0.2 and 0.

In addition the limit for the jitter is 0.5.  So changing in steps of 0.1 is quite large.  I suggest we change in steps of 0.05 with a bottom limit of 0 and a top limit of 0.5.
Could there also be either a tool-tip or a label  e.g. (0 gives no jitter, maximum is 0.5).

Once this is done could you look also at the **Describe > One Variable > Graphics** so the jitter plot here and on the boxplot dialogue are consistent.

maxwellfundi:
Fixes #4537

### DIFF
--- a/instat/dlgBoxPlot.vb
+++ b/instat/dlgBoxPlot.vb
@@ -49,6 +49,7 @@ Public Class dlgBoxplot
         bReset = False
         autoTranslate(Me)
         TestOkEnabled()
+        SetOptionsButtonText()
     End Sub
 
     Private Sub InitialiseDialog()
@@ -260,7 +261,7 @@ Public Class dlgBoxplot
             clsRgeomPlotFunction.RemoveParameterByName("width")
             ucrSaveBoxplot.SetPrefix("boxplot")
             ucrSecondFactorReceiver.ChangeParameterName("fill")
-            cmdBoxPlotOptions.Text = "Boxplot Options"
+            SetOptionsButtonText()
         ElseIf rdoJitter.Checked Then
             clsRgeomPlotFunction.SetRCommand("geom_jitter")
             clsRgeomPlotFunction.RemoveParameterByName("outlier.colour")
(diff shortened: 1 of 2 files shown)


instat/dlgBoxPlot.vb
```diff
@@ -256,22 +256,41 @@ Public Class dlgBoxplot
         'Sets geom function, fill and colour aesthetics and ucrsave prefix
         If rdoBoxplot.Checked Then
             clsRgeomPlotFunction.SetRCommand("geom_boxplot")
+            clsRgeomPlotFunction.AddParameter("outlier.colour", Chr(34) & "red" & Chr(34), iPosition:=1)
```
dannyparsons: There's an issue with adding parameters here:

- I choose boxplot, go to the sub dialog to change all the options (change outlier colour to green)
- Then on the main dialog I choose jitter
- Then when I go back to boxplot I've lost my green outliers because its set back to red

Similarly for jitter with height and width.

I understand why you need to do it here because you have the same geom function for all the geoms. Really these should be 3 separate functions so you don't have to add and remove parameters a lot and it will remove the options for each geom. Not saying this should be done now because there is lots to do but this will need to be done at some point - however if the dialog is changing a lot it will save time correcting this now. If not now, it should be an issue to be done.

maxwellfundi: you are right I figured out that we have same geom function and sensible place to add the parameters is where we are changing the geoms. I think by setting the outlier colour in our layer options should then update the outlier.colour parameter while still on boxplot and remembers it even when we change to other options if we had three functions. But as of now, I see why you got that issue.

Issue #4281, opened 2017-12-01
Improving the Prepare > Column: Date > Use Date dialogue


rdstern:
This should be used by more people, but there are a number of obvious edits that would improve this dialogue.  They could perhaps be done in stages:
1) Check the order that the columns are produced.  I suggest a logical order is Year, then month, then day, etc.
2) Have (ordinary) radio buttons with text Label.  Numeric, Abbreviated Name, Full Name.  They then apply to Month, week, dekad, pentad etc.  Then those are only included once.
3) Include Dekad and Pentad in the "scheme".  Currently they only produce numeric results. Correct the spelling of Dekad in the output.
4) Have a checkbox for Shifted Dates, default unchecked.  If checked it gives the Month field as now.  Default is January.  The day field is omitted.
5) The shifted applies to all the within-year functions except weeks, i.e. Year, Month, Dekad, Pentad, Day of Year (366).

Patowhiz:
@rdstern Please expound more on number 2 & 3.

rdstern:
This might be good to discuss over skype from tomorrow.  

Items 2/3: Take the example of generating months from the date.  Currently can generate 3 columns at the same time, by ticking all three entries.  Then you get numeric month, and abbreviated name and full name.  I don't see this as very useful.  I would be happy with just one of these and the user chooses.  Then whatever is chosen has the simple name **month**.  
To do this, the user first chooses which of these alternatives.  Then the resulting month is of the chosen type.  
This choice also applies to generating weeks, and dekades and pentades.

Patowhiz:
@rdstern can we discuss about this issue today ? I'm optimistic that this can be solved before the week ends.

Patowhiz:
@rdstern  in  item (b) do you mean, change the name of the **resulting column** from **day_in_year** to **doy_365** or the actual **checkbox text name** from **Day In Year** to **doy_356**?

![usedate](https://user-images.githubusercontent.com/21179007/39192182-5095d1c6-47e1-11e8-9cf1-56e802b12fdb.JPG)

Patowhiz:
Then on the same. I noticed the resulting column of Dekad has prefix **dekad** which starts with **small d** and not **capital D**. In your request you suggest **Dekad** and not **dekad**. Should I use the capital D or just small d for consistency? I don't know if this matters anyway.

Patowhiz:
[...]

PR #4558, opened 2018-04-24
Use date dialog

Patowhiz:
partly fixes issue #4281


Patowhiz:

instat/DlgUseDate.vb
```diff
@@ -203,7 +195,24 @@ Public Class dlgUseDate
...
+        dctMonth.Add("April", 4)
+        dctMonth.Add("May", 5)
+        dctMonth.Add("June", 6)
+        dctMonth.Add("July", 7)
+        dctMonth.Add("August", 8)
+        dctMonth.Add("September", 9)
+        dctMonth.Add("October", 10)
+        dctMonth.Add("November", 11)
+        dctMonth.Add("December", 12)
+        Return dctMonth
```
shadrackkibet: This works fine , just for consistency should this be done on `InitialiseDialog()`

Patowhiz: That sub gets called and executed in the initialiseDialog() for consistency. I separated it to make the code readable and have the initialiseDialog() short, The initialiseDialog() method appeared unnecessarily long partly because of this , good programming practice. But if it's making the entire class complex then I'm happy to remove it.

shadrackkibet: It will be good to know what other people think. Personally i prefer doing all this on  initialiseDialog() rather than a function.

Muthenya: I see your point @Patowhiz but I would argue the same as @shadrackkibet . Still, it will be good to hear what everyone else has to say.

Patowhiz:
Okay. For consistency and preference, let me move this to InitialiseDialog()

instat/DlgUseDate.vb
dannyparsons: Not a big difference, but I would say to only separate out code into subs/functions when they need to be called in more than one place. If I see a function, I would assume its being used multiple times, so I then need to know everywhere it's being called to understand the code, whereas if it's not separate I know immediately this is only done there. So that's a bit clearer to me.
There are other ways to make a long function more readable e.g. use comment to split into sections.

dannyparsons:
@shadrackkibet has this been tested now?

shadrackkibet:
Testing this now.

instat/DlgUseDate.vb
```diff
@@ -139,17 +111,30 @@ Public Class dlgUseDate
...
+
+        ucrNudShiftStartDay.SetParameter(New RParameter("s_start_day_in_month", 22))
+        ucrNudShiftStartDay.SetRDefault(1)
+
+        ucrInputComboBoxMonth.SetParameter(New RParameter("s_start_month", 23))
+        ucrInputComboBoxMonth.SetItems(getMonths())
+        ucrInputComboBoxMonth.SetRDefault(8)
+        ucrInputComboBoxMonth.SetDropDownStyleAsNonEditable()
+        'TODO This should probably be the default for this control, hence this is temporary
+        ucrInputComboBoxMonth.bUpdateRCodeFromControl = False
```
shadrackkibet: Is this really needed? what is it meant to do?

shadrackkibet:
@dannyparsons  i have tested this and it works fine. This is ready for you.

PR #4560, opened 2018-04-26
Edits Lineplot

### LINKED ISSUE #4556: Line plot bug with factors on x axis plus other factors
In the line plot dialog we have added `group = 1` in the `aes` when `x` is a factor. This forces all the points to be connected (otherwise there is no line). However, if the other factor receiver is also filled in then it still connects all the points instead of just the ones in the same factor group. e.g. in the graph below only the points of the same colour should be connected but because of `group = 1` they are all connected.
image

So instead what should happen is if `x` is a factor and the other factor receiver is empty then still add `group = 1` otherwise if the other factor receiver has variable `f` then add `group = f` so that each set of points in the groups of f are connected.

h/t @mmumbo

maxwellfundi:
@mmumbo I have edited this dialog now to produce this kind of graph as you would like 

image

Fixes  #4556


shadrackkibet:

rdstern:
Excellent!

mmumbo:
Thanks guys! this is exactly what i was looking for.

PR #4576, opened 2018-04-30
Adding , and NA Buttons - Calculator

### LINKED ISSUE #4574: Minor changes to the calculator
We don't have a comma yet, and that is needed for many functions.
I suggest it go in the bottom left corner of the Basic keyboard, instead of Del.

Then we also don't have NA.  I suggest this go next to the " " and above when on the Logical and symbols keyboard.

I have made it for 4.16, simply because I assume it should be very quick.  If not, then change to Pending.

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes  #4574


shadrackkibet:
@maxwellfundi  please have a look at this.

Issue #4111, opened 2017-10-27
Simple change in Start of rains etc


rdstern:
In the 4 dialogues, **start, spell, extreme** and **end**, the columns overwrite unless you consciously change the names.
These are dialogues that one uses repeatedly.  

So could they use the code we have in other dialogues, so they are **start1**, etc, i.e. unless you change them yourself, then they automatically give a new name.

I am suggesting Maxwell could allocate someone to do this, possibly under his supervision.

There are other changes needed for these dialogue, but this is the one that is the same for all 4!

Muthenya:
@rdstern what's the best dataset for testing these dialogs?

rdstern:
In the library the simplest is the Dodoma data.
There is also the Ghana data, which has 2 stations - also useful.
Third (and useful you you, given where you are, is the Kenya rainfall data from James that has more than 50 stations.

I think all are simple to access from the library.  If not, then please tell me.

dannyparsons:
This isn't simple to do in the same way we do on other dialogs. The difference is we don't know which data frame the result is going to before running so we can't detect the correct suffix number to use.

We could do something less intelligent like increase the number each time the dialog runs. This doesn't guarantee to not override but it's unlikely to. It would be when you change data frames you might start at a higher number than needed.

With a bit more time we could get the code to detect which data frame the result is going to and then have the same intelligent suffix numbers. But I suggest for now we either leave as it, or implement the less good fix.

rdstern:
I understand more now.  I suggest this be left open and moved from 0.4.7.
And we leave it as it is for now.  It is a "feature" of moving from one data frame to another.  
And when we teach about these dialogues (and in the Help) we mention this point.

If it were easy, then one possibility would be to put a label into the name field.  e.g. Enter name of resulting column, e.g. spell.  But it would have to be a label such that OK then isn't enabled until you have entered a valid name.  And this would be better if we could then start to type and this text would automatically be removed.  This sometimes happens in on-line forms to complete.

lilyclements:
Closing as we have moved past this in SoR

PR #4579, opened 2018-05-02
Qc Temperature

### LINKED ISSUE #2392: Including the QC from Rclimdex and Rclimdex extra in R-Instat

shadrackkibet:
Fixes #2392


shadrackkibet:
@africanmathsinitiative/developers this is now ready for review and testing. Fixes #2392

shadrackkibet:
@rdstern  Here is the screenshot of this dialog.
image

rdstern:
Looks nice.  
I wonder about the New Column Name field.  I think I would have "New Column Prefix", because won't you usually have a number of columns you are generating, e.g. qc_diff, qc_same, etc?

shadrackkibet:
If i have qc_diff, qc_same, etc? as defaults what would be the default name when more than one checkbox is checked? 
I am currently using an input textbox for saving like other climatic dialogs so this is not the prefix. This is also discussed here #4111

rdstern:
That's the whole point.  When I use the dialogue I usually have more than one checkbox selected and then will be generating multiple logical columns.  That's why (at least for now) I would be giving them these fixed names.  So I would not mind if you didn't have that name field for now.  If the names are to be changed, (and hence not overwritten the next time the dialogue is used), then we need to change the column name in the data frame, once the dialogue is executed.

PR #4584, opened 2018-05-08
Adding Dodge param to Barchart

### LINKED ISSUE #4581: Position for Bar Chart dialog
From using the bar chart dialog a lot it would be useful to have the position parameter on the main dialog. It only affect the bars when you have a second factor but I find that people often want to change the position to "dodge" instead of the default "stack". There's not much on the main dialog currently so I think this can be added there.

I also wonder if we should have "dodge" as our default instead of the ggplot default "stack". It seems much easier when you have a lot of levels but maybe there's a reason "stack" is a better default?

maxwellfundi:
fixes #4581


maxwellfundi:
What do you mean change parameter position? Dodge is the first parameter on the geom_bar

We need to add this parameter when we have two factors, so that why I added them there. Maybe my understanding was wrong. What do we want?

dannyparsons:
I mean the parameter called "position". As a user, what do I do if I don't want dodge as the position? My suggestion was to have the "position" parameter on the dialog. I don't see anything added on the dialog, just that you've hard coded "dodge" as the position.

maxwellfundi:
ooo okay, I will then add a combo box with the positions so that the user can be able to change.

dannyparsons:
Great.

maxwellfundi:

dannyparsons:
Why have a checkbox on here as well?

maxwellfundi:
So that we can control how we pass in the parameter. Further, we can have the parameter removed if the user does not want to have the dodge, by simply unchecking it.

maxwellfundi:
@africanmathsinitiative/developers  Can someone test this?

shadrackkibet:

shadrackkibet:
This is working fine. Just a few two things to sort.

1) When i am in the  subdialog and then check the checkbox , select  `stack` ,go back to the  dialog and then back to the subdialog i find the checkbox is unchecked. That is not the case for the other options.
2) You need to sort out Tab order.
3) For consistency the options in the combobox are in title case while those in the subdialog are all small letters with quotes around. I think this should be consistent.

maxwellfundi:
@shadrackkibet  Your issue no 1 is happening because `stack` is the default and so, when that is selected, you obviously dont expect the checkbox to be selected. 

I will sort tab order. 

On issue number 3, I used a dictionary for the combo box, I can change that so that from title case to lower case for consistency, but I would not add the quotes on to the items of the combo box. Its uncommon to have such.

dannyparsons:
I think it's ok that they are not consistent in this case. The subdialog has the actual parameter values but on main dialogs we are able to have nicer looking names, so although consistency is good, if we can have the main dialog looking nicer then we should.

PR #4594, opened 2018-05-10
removed French resource files and disabled changing language

dannyparsons:
@africanmathsinitiative/developers please could you test this for build errors and crashes.


maxwellfundi:
@dannyparsons  I have been testing this and dialogs open okay, and no build errors.  Why was the resource files removed and translations disabled?

Muthenya:
This seems to work fine from my end.

dannyparsons:
Ok thanks for testing. The translation into French was very messy, not just the words but controls shifted. It would be the same in Swahili, but I removed the French files because users with French machines were getting the French version of R-Instat by default and in some case it was no easy to go back to English. So removing the resource files was the easiest solution it defaults to English for everyone now - the resource files are easy to add back in later since the translations were all automatic.

PR #4602, opened 2018-05-16
Column hide

### LINKED ISSUE #4372: Hide and Unhide Columns and Sheets?
Well, probably this was discussed before and could be clarified to me. Currently to hide a column, I have to go to Prepare > Data Frame > Hide Columns . Could we have a short cut for this in when I right click on column, probably just after the duplicate column. I find hide column as one of the most frequent functionalities used in grid like user controls. Under this shortcut, the dialog doesn't necessarily have to appear, it could just call the function involved in hiding column directly.

Then on the Prepare > Data Frame > Hide Columns which is better between that and Prepare > Data Frame > Hide/Show Columns . I believe the dialog does both functions.

Patowhiz:
fixes partly #4372


Patowhiz:

PR #4604, opened 2018-05-19
Quality control-Temperature.

shadrackkibet:


shadrackkibet:
@africanmathsinitiative/developers this is ready for review and testing.

shadrackkibet:
@dannyparsons  i suggest this to be merged so that @rdstern  can test it as soon as possible.

rdstern:
This is a good advance.  Lots seems now to work.

a) Please check the settings when the check-boxes are implemented for the logical columns and the calculated columns.  When I checked for both I only got the logical.  
When I checked for just the calculated, then I got both.

My only other concern is the names of the calculated columns.  I would like them to be shorter.  For example, could range_element1 become either range_1 or even range_tmax if you could pickup the name of the column for element1.  Perhaps check with Danny?

Then instead of largest_same_element1 have Same_1, or Same1 (i.e. the logical column is same1 and the calculated column is Same1, or Same_tmax, etc?

I am really pleased with this.

a) Will you also do this for the rainfall?
b) Beth says she has also completed the changes to the outliers.  Perhaps you could test this and then incorporate the 2 sets of changes.  Then that's done and very impressive!

shadrackkibet:
@rdstern  Thanks for testing this , i am going to look into this issues , changing the names is actually easy. Once this is done i will go ahead incorporate that on the Rainfall dialog.

rdstern:
Looks very good now.  I saw you were able to capture the names of the columns (e.g. same_Tmax) in the logical columns.  
1. In the calculated columns I expected Same_Tmax and Same_Tmin, but I only got one of them, called Same.  Also Jump should be Jump_Tmax and Jump_Tmin.
2. Still a small issue with what is displayed.  When I click to have **calculated**, but **not logical**, I still get logical. Oh, it is more than that.  I have now opted for neither and I still get both.  I suspect perhaps it isn't clearing?  I have now reset and asked for neither and I still get both!
3. I now tried Outlier only.  Unticked all the others, but asked for Calculated and Logical.  I still get them for all the checks.
4. I have not checked the Outlier fully, but I tried 2 different limits and still get the same number of outliers.  So I suspect this is not connected yet?

PR #4624, opened 2018-05-25
UseDate

### LINKED ISSUE #4603: Continue improvements in the use date dialogue
1. The addition of the quarter function has been a good improvement.
2. The main suggestions I make now are to improve the layout of the check-boxes.
3. In addition, I think the Pentad does not work correctly.  It should produce a factor with 72 levels, and with just 5 (or 6) days in each month.  Perhaps Beth can look quickly at this?
4. And more of the variables should have an option of labels.  For example Quarter is just 1 2, 3, 4 which is correct if numeric.  It could be Jan-Mar, Apr-Jun, Jul-Sep, Oct-Dec as short labels.  And these should (like the months), ideally be done so that they are automatically in the computer's language where it is not English.  
5. Similarly Dekad would have the label option of Jan1, Jan2, Jan3, Feb1... With Pentad it would be Jan1 to Jan6. 
6. And more variables should be shifted as an option, namely Month, Pentad and Dekad.  When these are labelled they will look the same, but Aug would be level 1, if shifted to Aug, with July as level 12.  For Dekads Aug1 would become Level 1, with Jul1 being level 34.

So my suggested layout is roughly as shown by Danny who may take a  photo.  The 6 variables that can be shifted are first, each with a column, namely 
**Year, Month, Day in Year(366), Pentad, Dekad, Quarter.**
In the left hand side are the 3 options **Numeric, Abb Name, Name**, with check-boxes to indicate what is possible.
So:
a) Year is just numeric,
b) Month is Numeric, or Abb Name or Name.
c) Day in Year(366) is just numeric
d) Pentad, Dekad, Quarter are Numeric or Abb Name.

Below that block is:
**Shifted Month and pull-down** as we have.

Below that is another block of variables to which shifted does not apply, i.e. 
a) Day in Month - just numeric
b) Day in Year - just numeric
c) Week is numeric or abbrev name Wk1, Wk52 or full name Week1, to Week52.
d) Weekday also has the 3 options
e) Leap Year is just numeric (it is actually logical, but let's not be too purist!

I suggest that Patrick could work on the layout and organisation of the dialogue, while Beth concentrates (with him) on the R code.

Patowhiz:
Partly Fixes issue #4603 Continue improvements in the use date dialogue .
Still in progress


Patowhiz:

shadrackkibet:
@Patowhiz i see another pull request with this changes here #4755 should this be closed?

Patowhiz:
Yes, this can be closed.

PR #4627, opened 2018-05-28
Climatic boxplot

### LINKED ISSUE #4024: Climatic > Check > Boxplots
Here are some specific points about this dialogue.  But it would be useful if John and/or Beryl could look systematically at some of the graphics and try them.  I start the list here.

They could also use the climatic data from the library, both the Dodoma data (1 station, 4 elements) and the Ghana data (2 stations, more elements).  I started with the Dodoma data and applied a filter for just 1988.  So just 366 values.

I would like the default (for the climatic) to have the x as the month and the y (if it exists) as Tmax.  If not, then leave the y variable empty.
No facet by default.

1. I tried a jitter plot.  Then did the plot. The default jitter is too wide (0.4).  So I returned, and used jitter options and made it 0.2.  This worked well, but the x variable of month_abbr was no longer the x variable.  Could you make sure the x-variable is remembers correctly - whatever it is.  It is remembered when you return to the dialogue, but not when you return from the sub-dialogue.
2. Having set the jitter option to 0.2 I decided to do a violin plot.  I found the violin plot also used the 0.2 setting of jitter.  And when I changed the jitter width to 0.4 it changed the width of the violin plot.  I wonder why?
3. I was then asked to add the jittered points to the violin plot.  I therefore started with the violin plot and added a layer for the geom jitter.  I returned and checked that these were the 2 geoms, (i.e. violin + jitter).  I then did the plot - and it worked except I got boxplots with the jittered points.  That was great if I had wanted that!  But I had asked for violin!
4. There is also another problem with the boxplots (and the forthcoming violin plot.)  I seem to be getting jitter + boxplot, rather than boxplot + jitter.  By this I mean that the boxplot (and later the violin plot means that I don't see all the jittered points.  I suggest the R command for the layers is given the wrong way round. (This may be a general problem?)
5. So I do the plot the other way round.  Namely jitter plot with the added layer on top of box plot.  I now get the plot I wanted, but had to "remember" the x-variable was month_abbr, because that wasn't included.
6. Now I decide I would like to edit the boxplot layer I just added.  (The dialogue is still on jitter).  I go to the sub-commands and notice that R-Instat has swapped the order of the layers.  (So boxplot is now the first layer - even though it wasn't!)  It now says I can't do an edit, because that is the main layer.  Hey - who is in charge here!!! 
7. If that is not challenge enough, then (and I would like John and Beryl to do this as well!) I would like to be able to pose problems and see how to do the corresponding plot.  The first one is the be able to connect the medians in the boxplots.  This is shown here [https://stackoverflow.com/questions/18599411/create-a-boxplot-for-multiple-stations-and-line-joining-median-of-the-stations](url), but I don't see how it fits into our current system.  (I can do it in my old Instat, and I can also add the means!) A second challenge is in the Hadley Wickham book and is where there is an additional factor (here say the day of the month) and I would like to connect those values across my successive boxplots.  

I think the points above may be easier to see if John or Beryl try to repeat them so you can see them locally.  (I would be happy to show them over skype)  Then I hope they might do similar analyses to try more aspects of the plotting, and not just this dialogue.  In general the graphics is looking very good indeed!

Muthenya:
Fixes #4024


maxwellfundi:
Will review this shortly.

Muthenya:
@dannyparsons I've made the edit on the ucrCore. This should be ready now.

PR #4631, opened 2018-05-29
Adding start_dates as an option - Start of Rains

shadrackkibet:
@dannyparsons  this is ready for you to review. I have implemented start_doy and start_dates as sub_calcuations and added them as combined calculation. 

@rdstern  here is the latest design of start rains dialog.

image


dannyparsons:
I think you could move the checkboxes to the left of the textboxes with labels "Display doy column" "Display Date column". Then you probably don't need the labels.

shadrackkibet:
@dannyparsons  that is now done.

PR #4639, opened 2018-05-30
Clean up for Column Summaries sdgs

lilyclements:


shadrackkibet:
I don't think that is possible for our ucrInputtextboxes as at now . We might want to change this to an Nud for now?

lilyclements:
Found a way to fix it by `SetValidationTypeAsNumericList(dcmMin:=0, dcmMax:=1)`

So the remaining bug is having two checkboxes controlling one input means that there is an issue in the inputs visibility when you check/uncheck the boxes. Can you think of an instance where we have one input being controlled by two checkboxes elsewhere or is this the first that this issue has come up?

dannyparsons:
The percentile textbox is ok, the min/max values were needed
@shadrackkibet anything else?

shadrackkibet:
checking this again.

PR #4645, opened 2018-06-01
improved error messages when using incorrect R version

dannyparsons:
@maxwellfundi @lilyclements are these messages now clearer?

image


maxwellfundi:
@dannyparsons  This looks much better now actually and more informative

Issue #4606, opened 2018-05-21
Column summaries - small issues


rdstern:
The test set of data I used had 2 columns - from File > New
data_1 is 1,2,3,4,5,6,7,8,9,10
data_2 is 1,2,3,4,NA,5,NA,6,7,8

So very simple - no factors!
Use Prepare > Column: Reshape > Column Summaries

1. Get the proportion > 4
For data_1 it is 0.6 - correct.
For data_2 it is still 0.6 - **incorrect**.  It should be 0.5.  Presumably it is setting the missing values to TRUE.
2. Get the correlation (and covariance) of these 2 columns with data_2.  Both results are given as missing.  So the setting is wrong for those two functions.
3. Delete data_2 from the 2-variables tab.  Now run again.  It gives an error, because there is no element to get the correlations.  But (with our usual rule) if the correlation or covariance is ticked, then OK should not be enabled unless that field is filled.

dannyparsons:
@africanmathsinitiative/developers This would be good for someone to look into carefully, particularly the problem with proportions and understand the R code and identify the problem. I'd like to point out that this was both implemented and tested without anyone noticing that it sometimes gives the wrong answer.

bethanclarke:
I will look at these issues @rdstern

bethanclarke:
@dannyparsons I can't seem to find the calculate_summary command anywhere. This the code for proportion > 4:
`InstatDataObject$calculate_summary(data_name="test1", columns_to_summarise=c("a","b"), test="'>'", value=4, summaries=c("proportion_calc"), silent=TRUE)`

bethanclarke:
There is a problem when it comes to calculating the correlations. 

`summary_cor` function requires arguments x and y, however x is not defined in VB. So I am trying to implement the argument x, where x is equal to the column(s) from the multiple receiver. But when we have multiple columns it either needs to a vector, matrix or data frame containing the variables from the multiple receiver. 

First, I tried to obtain the correlation between data1 and data2 (only selected data1 in multiple receiver) and this did not work in R-Instat because x wasn't included in the code and it didn't work when I ran it in R including x parameter: 

```
InstatDataObject$calculate_summary(data_name="test1", columns_to_summarise=c("a","b"), x=a, summaries=c("summary_cor"), silent=TRUE, y=b, use = "complete.obs")
Error in summarise_impl(.data, dots) : 
  Evaluation error: formal argument "x" matched by multiple actual arguments.
[...]

PR #4647, opened 2018-06-01
Column Summaries

shadrackkibet:

### DIFF
--- a/instat/static/InstatObject/R/Backend_Components/summary_functions.R
+++ b/instat/static/InstatObject/R/Backend_Components/summary_functions.R
@@ -618,6 +618,19 @@ count_calc <- function(x, count_test = "==", count_value, na.rm = FALSE, ...){
   }
 }
 
+#standard error of mean function
+
+standard_error_mean <- function(x, na.rm = FALSE, ...){
+  if (!na.rm){
+    if(sum(is.na(x) > 0)) return(NA)
+     return(sd(x)/sqrt(length(x)))
+  }
+  else{
+    y <- na.omit(x)
+    return(sd(y)/sqrt(length(y)))
+  }
+}
+
 
 instat_object$set("public", "summary_table", function(data_name, columns_to_summarise = NULL, summaries, factors = c(), n_column_factors = 1, store_results = TRUE, drop = TRUE, na.rm = FALSE, summary_name = NA, include_margins = FALSE, return_output = TRUE, treat_columns_as_factor = FALSE, page_by = "default", as_html = TRUE, signif_fig = 2, na_display = "", na_level_display = "NA", weights = NULL, caption = NULL, result_names = NULL, percentage_type = "none", perc_total_columns = NULL, perc_total_factors = c(), perc_total_filter = NULL, perc_decimal = FALSE, margin_name = "(All)", additional_filter, ...) {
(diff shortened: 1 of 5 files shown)


rdstern:
Aha.  I am getting some odd results testing proportions and counts.  So perhaps I am doing this too soon?

shadrackkibet:
Perhaps that is because you had same controls for both counts and proportions so obtaining them at the same time for different parameters wasn't possible. I have now added separate controls.

At what instance are you getting odd results?
image

rdstern:
Now I have added the percentage (in a proper calculation) and it comes with 13 decimal places!  It also has the % sign straight after the number, which I am not sure we need.
When I look at the information it says it is a **noquote** class and the number of significant figures is NA.  When I change this to 3 it doesn't seem to have any effect.

shadrackkibet:
Missing values is not yet fixed for counts, I will look into that.

For proportions as  percentage, what we are doing is multiplying it by 100 and that is why we have % sign. If you would like this sign to be removed I can do that easily bit I think it is necessary to have.

I can also look into rounding decimals values to a certain number of decimal places if that is needed.

rdstern:
That would be great.  And, if it is easy, could you look at the naming of the resulting columns?  They are very long just now and don't indicate the proportion or %.  Ideally for a column, called SDU it could be SDUgt2, etc, with lt, le, eq, gt, ge for the 5 options.  I assume that . is allowed in the name, so that SDUge2.5 is also allowed.  The name could be the same for proportion and percentage.  Perhaps for count it could be with a c in front.

dannyparsons:
I think the percentage option is returning a string with "%" added at the end, so probably why it is coming out with many decimal places. I think we could modify the R function to return a number instead.

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -595,20 +595,27 @@ proportion_calc <- function(x, prop_test = "==", prop_value, As_percentage = FAL
   if(!na.rm){
     if(sum(is.na(x)) > 0) return(NA)
     y <- x[eval(parse(text = paste("x", prop_value, sep = prop_test)))]
-    if(!As_percentage){return(length(y)/length(x))}
-    else {return(noquote(paste0((length(y)/length(x))*100 ,"%")))}  
+    if(!As_percentage){return(round(length(y)/length(x),digits = 2))}
+    else {return(noquote(paste0(round((length(y)/length(x)*100),digits = 2 ),"%")))}
```
shadrackkibet: @dannyparsons i am rounding off this results to 2dp here. what are your thoughts on this?

dannyparsons: That's ok and I think also remove the noquote and paste with "%" so that what is returned is a number.
@rdstern I assume doing percentages should still return a number and not a string e.g. "34.5%"

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes #4651 and #4606

bethanclarke:
@shadrackkibet I will test this and implement missing values options

shadrackkibet:
Please go ahead.

bethanclarke:
Count with missing values, proportion as percentage and standard error of mean are working well. @dannyparsons this is ready for you to review. 
@rdstern are there any changes to the variable names you would like done?

PR #4666, opened 2018-06-08
Logical Column for the Event - Start of Rains

shadrackkibet:
In progress!


shadrackkibet:
@dannyparsons i have now implemented logical column for the event on SOR as discussed earlier. Please test this.

dannyparsons:
I ran the dialog with all the defaults and got this error. I think it's caused by `grouping_by_station_year` which doesn't have calculated from columns. I think this should just be removed from everywhere when there is no station column.
image

dannyparsons:
I used the Ghana data with one condition: Total rainfall over 2 days >= 60 mm

Seems to be a problem in the start or status column for Saltpond 1984. If the status is NA then the start should be NA, as this indicates there are missing values in the rain column so neither start nor status can be determined.
image

For the years at the end it seems to be correct, the start is missing but the status is FALSE - indicating that the start didn't happen and this is not caused by missing values (the order of the years will be fixed by new version of dplyr).
image

dannyparsons:
Also see here that there are missing values in some other years. So it is good to check the start values in these years and where the missing values are. If the missing values are before the start day then start should be missing, if the missing values are after start, then it shouldn't be missing.
image

dannyparsons:
I think this is one cause of the problem: `"doy <= ifelse(366, start_doy, is.na(start_doy[1]))", ` parameter are in the wrong order to `ifelse` and `start_doy` should have `[1]` in both places.

shadrackkibet:
We are also grouping by station only elsewhere, should we instead group by Year when there is no station?

dannyparsons:
For the rollings sums which are done by station only, when there's no station we don't need to group.

shadrackkibet:
Thanks for testing, i will look into this issues.

shadrackkibet:
@dannyparsons  i have fixed this problems. This is ready for you.

dannyparsons:
This looks good. I think we might want "start_status" for the status column name, as there could be different status columns coming from different calculations. Other than that it looks good, will need careful testing after merging still

PR #4672, opened 2018-06-11
PICSA Rainfall Graphs

maxwellfundi:
@africanmathsinitiative/developers  this is the new dialog for the PICSA rainfall graphs. It works similarly to line plot dialog. 
I have one issue here of setting  the ucrvariablesAsfactor the current receiver on first load because i have the X receiver being automatically filled in with the year column. We have tried the known fix for this with @shadrackkibet by arranging the receivers in order with the ucrvariablesAsfactor  as the last but did not seem to work either. 
This is needed for today for installer tomorrow and hence important that it gets tested soon. @africanmathsinitiative/developers its ready for review and a work around of the above issue is welcome


maxwellfundi:
I have just released @rdstern has raised a similar issue on other dialog on #4686. We need to investigate this further i think.

Muthenya:
This looks good from my end.

maxwellfundi:
@dannyparsons  you can have a look at this now its needed for todays version

dannyparsons:
I will merge for @rdstern to test as this is more urgent

maxwellfundi:
okay great, am putting it on the frmMain

PR #4692, opened 2018-06-15
EOR

bethanclarke:
copied files across from End of Rains


dannyparsons:
Thanks Beth. It's now merged. Can you check in the latest version that this is working correctly? just to double check the copying over worked correctly.

PR #4703, opened 2018-06-19
Hypothesis Tests

### LINKED ISSUE #4440: Hypothesis Calculator
I had a look at the hypothesis calculator. For testing the keys/dialog as a whole:

* I noticed there were a few hypothesis tests that are missing - is this intentional? (I can't say I've heard of these, but they are `ansari`, `mantelhaen`, `quade`, `PP`)
* There's also the normal bugs which I assume are known (bugs are similar problems in the other calculator - e.g., you can insert a factor when inappropriate and hence get an error)
* Fun bug - If I click "Show Arguments" then select the test I want then it works fine, however, if I select the test I want and then click "Show Arguments", the arguments are not shown. (If I do get arguments shown and then click it to undo the arguments then it doesn't do anything)
* If I had a test selected, and then clicked a new test, it comes up as an argument in the last test (e.g. `bartlett(binom())`).
This doesn't make sense to me but just to check in case I'm missing something - would there be a situation where one test is read in as an argument to another test? (e.g. `bartlett(binom())`)
If not, then is it possible eventually perhaps to have these buttons act more like the radio buttons as boxes so that only one can be selected at a time? It could then overwrite if you select a new function (if you haven't inputted any arguments?) 
This would also help the issue I had where I selected, say, `poisson`, then wanted to remove `poisson`, and selecting `poisson` again won't remove it, it instead does `poisson(poisson())`.

image

_e.g. I've just done it for_ `bartlett`, `binom`, `Box`, `fligner`, _and_ `friedman`_here - and_ `bartlett` _is selected_

* Could there also be the tool-tips to briefly explain the tests using the descriptions in R - e.g., for the `bartlett` test _"Tests the null that the variances in each of the groups (samples) are the same."_

* For the question of how can the keyboard be improved/divided up, I just was wondering; is the idea here that the user has their data and they're looking for a test in mind? I think that the most sensible way is to do the common tests as one keyboard and less common as another.

I'll look into the `by` tomorrow, hopefully!

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes #4440


Muthenya:
This looks good from my end. I have also learnt that **Help** works when the package being looked into has been installed.

shadrackkibet:

PR #4711, opened 2018-06-19
Changes adding Surv object

lilyclements:


lilyclements:
Adding "surv" as a new save type. It is still not working properly - I am conscious that I have changed a lot of files, however, so perhaps should be merged before I try to fix the problem later this week when I go back to Reading (Probably Thursday?).
It has not reduced functionality in any of the current code.

The issue is when I run the `$get_surv` part I get an error

``` Error in self$get_object_names(type = type) : object 'out' not found  ```

A traceback in R takes this initially to line 1619 in data_object_R6.R
`self$get_object_names(type = type)`

And then to line 473 in instat_object_R6.R 
`self$get_data_objects(data_name)$get_objects(object_name = object_name, 
    type = type, force_as_list = force_as_list)`
It's on this part that it says that if it doesn't work to give an error: "object_name not found".

dannyparsons:
Nothing obviously wrong I can see but looks like something missing somewhere so should be possible to debug with some help.

lilyclements:
@dannyparsons looked over this with David and fixed the bug. This is ready for review.

lilyclements:

lilyclements:
@africanmathsinitiative/developers can this get looked at soon please? I'm concerned as it's affecting quite a few files and we don't want a conflict

shadrackkibet:
I have tested this Rcode and it is working well. @lilyclements is there an example(dialog) where you have used this objec?

lilyclements:
Thanks! I made temporary dialogs to test it where it worked well, however, removed them for this because I didn’t want to merge these temporary dialogs

shadrackkibet:

PR #4746, opened 2018-06-26
Qc Temperature

### LINKED ISSUE #4649: The rainfall and temperature quality control dialogues
They are now looking great.  But still to complete:

a) I cannot get the rainfall to work.  Even with a simple example for Dodoma (from library) checking just rain > 50mm it says:

**Error running R command(s)
Error in self$apply_instat_calculation(manipulation, curr_data_list, previous_manipulations) : 
  No data specified for calculation.
.temp_val <- capture.output(InstatDataObject$run_instat_calculation(calc=rainfall_filter, display=FALSE))
OK**   

b) The rainfall enables OK even when there is no data in the rain control.  I first tried (accidentally) with the wrong data frame and it was happy to enable OK.

c) For the outlier option I think in rainfall and temperature the calculated column, logical column (and please check the calculation also) have to be completed?  I get the limits it uses (which I don't need, but would like the value of the offending observation (compared to the 1.5 say).  And I don't think the logical column for this case is yet connected?

Sorry for starting a new issue, but I couldn't easily find the original.

shadrackkibet:
In progress Fixes #4649

### DIFF
--- a/instat/clsQualityControl.vb
+++ b/instat/clsQualityControl.vb
@@ -272,8 +272,8 @@ Public Class clsQcOutliers
         clsOutlierUpperLimitCalc.AddParameter("result_name", Chr(34) & strUpperCalcName & Chr(34), iPosition:=4)
         clsOutlierUpperLimitCalc.SetAssignTo("upper_outlier_limit" & strElementName)
 
-        clsOutlierUpperLimitFunc.AddParameter("bupperlimit", "TRUE")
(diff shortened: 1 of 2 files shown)


shadrackkibet:

shadrackkibet:
Issues fixed.

- Logical and calculated columns for outliers.
- Bug on outliers limits not changing when coefficient value is changed.

maxwellfundi:
@rdstern  We have tested this fr the Rain and temperature and now works properly. Please test them further from your side too.

PR #4753, opened 2018-07-05
Enhanced filter

Muthenya:
In progress.


Muthenya:
@africanmathsinitiative/developers this contains the addition of date-picker and setting up of a drop down with options for logical, numeric and character types of data. It's still a work in progress especially on part of the date-picker and more work needs to be done here. 

The remaining aspect is the issue of the updating of the display. A temporary fix was proposed of clearing the receiver when the column data type is changed but I need your take and suggestions on how to implement this properly since this change clears the receiver every time the sub dialog loads. Otherwise, I'm looking forward to getting a better solution from the team.

Muthenya:
The proposal is still in order. I'll reopen the issue soon.

PR #4757, opened 2018-07-05
Duplicates

### LINKED ISSUE #4677: Duplicates dialogue small bug on 3rd option
If I am right, then this should be an easy one to fix.  

The 3rd option on the Prepare > Check Data > Duplicates calls a Sam-written function to give the index number of the duplicates.  I tried it on the attached data set and it gave some duplicates, but not all.  I think it is only giving the duplicate index when the whole record is duplicated - which is fine if I was using that option.  But I wanted the duplicate rows for just 2 of the variables, i.e. Date and Station. 

image

[Petion_Perrin_All.zip](https://github.com/africanmathsinitiative/R-Instat/files/2090699/Petion_Perrin_All.zip)

The sheet used is Petion_Perrin_stacked_Subset1

Muthenya:
@africanmathsinitiative/developers please review this.
Fixes #4677.

### DIFF
--- a/instat/dlgDuplicates.vb
+++ b/instat/dlgDuplicates.vb
@@ -67,6 +67,7 @@ Public Class dlgDuplicates
         ' for rdoSuccessiveVariables this is ran as a string, it isn't run for rdoSelectedVariables.For rdodataframe this is ran as an r-function so we are doing this manually but passing the parameter in the additional code pair
         ucrSelectorDuplicateswithVariables.SetParameter(New RParameter("x", 0))
         ucrSelectorDuplicateswithVariables.SetParameterIsString()
+        ucrSelectorDuplicateswithVariables.bUseCurrentFilter = False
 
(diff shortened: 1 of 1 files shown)


Muthenya:
@africanmathsinitiative/developers?

Patowhiz:
Reviewing this

Patowhiz:
@Muthenya I'm still getting a similar error even after merging the latest commit.

error3

Patowhiz:
I'm using the data attached to the issue, After doing a filter of say, Date == '1950-06-01', then trying to get a duplicate on the 2 variables shown above, Rain and Day, I get the above error

shadrackkibet:
@Muthenya you need to set the current filter as false for the receivers as well.

Patowhiz:
@Muthenya this looks good now. No errors even after the filter is applied. The output is what I expect from the various options. @dannyparsons this is now ready for you.

PR #4765, opened 2018-07-11
Initial receiver focus

dannyparsons:


instat/ucrReceiverSingle.vb
```diff
@@ -309,15 +309,19 @@ Public Class ucrReceiverSingle
...
         '     don't always want to autofill when dataframe is changed.
         '     Something like AndAlso Selector.CurrentReceiver.bAttachedToPrimaryDataFrame
         '     except always want to autofill when resetting regardless of current receiver
         If bAutoFill AndAlso Selector IsNot Nothing AndAlso (Selector.CurrentReceiver Is Nothing OrElse Selector.CurrentReceiver.bAttachedToPrimaryDataFrame) Then
+            ucrCurrentReceiver = Selector.CurrentReceiver
             SetMeAsReceiver()
             If Selector.lstAvailableVariable.Items.Count = 1 Then
                 Add(Selector.lstAvailableVariable.Items(0).Text, Selector.strCurrentDataFrame)
             End If
+            ucrCurrentReceiver.SetMeAsReceiver()
```
shadrackkibet: I think we need to check if  `ucrCurrentReceiver` is not nothing then do this , otherwise when the current receiver is not set on load this breaks here;
image

dannyparsons: You're right. I read the If statement on 318 incorrectly, I thought that was being checked. So yes adding that check is needed, can you do that?

shadrackkibet: yes.

PR #4769, opened 2018-07-13
Qc Temperature

### LINKED ISSUE #4768: Minor changes on the QC for temperature dialogue
This is looking really good now.

I am using it with the Guinee data - 2 stations and both Tmax and Tmin.

a) When I use the jump option it gives the Jump_Tmax (value) and jump.Tmax (logical).  But for Tmin (the second column) it doesn't give the value column, i.e. it gives jump_Tmin but not Jump_Tmin.
b) There could also be a column that gives the value of the difference, (column called Diff) as well as the logical column (diff) that is there already.
c) In the outlier case (I changed the limit from 1.5 to 3, so it gave 139 rows in the Guinee data) it gave the column called upper_outlier_limit_tmax, but not the others.  It used to give all 4.  It now gives all 4 logical columns.  It should give all 4 limits as well.
d) The names of the resulting columns are far too long and also have the tmax and tmin at the end.  They are more like column labels than names!  Let's call them Ulim_tmax, Llim_tmax, then ulim_tmax and llim_tmax for the logical, ditto for the minimum.
e) In the names given to the logical and calculated columns you use a generic name (tmax and tmin).  Ideally you would (instead) pick up the actual name of the column (here Tmax and Tmin).  David says this isn't trivial, so it might wait, i.e. have lower priority than the other changes - which I hope are easy!

shadrackkibet:
Fixes #4768


shadrackkibet:

shadrackkibet:
@bethanclarke please test this.

bethanclarke:
All of the columns for tmax and tmin are produced for jump now. Logical and calculated columns work well for all options.

The outliers is working well, however is it possible to change the order of the columns that are produced in the filtered dataset? Could the order be Ulim_element1 logical (L), Ulim_element1 calculated (C), Llim_element1 (L), Llim_element1 (C), Ulim_element2 logical (L), Ulim_element2 calculated (C), Llim_element2 (L), Llim_element2 (C).

And the label for upper limit element 1 logical column is missing (L)

shadrackkibet:
> The outliers is working well, however is it possible to change the order of the columns that are produced in the filtered dataset? Could the order be Ulim_element1 logical (L), Ulim_element1 calculated (C), Llim_element1 (L), Llim_element1 (C), Ulim_element2 logical (L), Ulim_element2 calculated (C), Llim_element2 (L), Llim_element2 (C).

Yes this is possible. I will wok on it.

shadrackkibet:
> And the label for upper limit element 1 logical column is missing (L)

Will also look into this. Thanks for testing this.

shadrackkibet:
> And the label for upper limit element 1 logical column is missing (L)

Probably you need to expand the column in order to see it. In this screenshot from Ghana_Data the label is there.
image

maxwellfundi:
@bethanclarke  This seems to be a better version of the dialog that what we currently have. Is it acceptable to have the columns in their current state then create an issue for them for the next release maybe?

bethanclarke:
@maxwellfundi yes that’s fine, the columns created are correct, just the ordering would be nice to have differently but that doesn’t matter for the release

PR #4777, opened 2018-07-17
Show non numeric columnc

### LINKED ISSUE #4766: Dialogue to show non-numeric values
I suggest this starts as a very simple dialogue.  It is in the **Prepare > Check Data > Non-numeric Cases**. It is the third in that menu, i.e. below Compare Columns and a horizontal line is just below that.

It starts at least as a very simple dialogue.  There is the usual data selector, which only shows character columns.  I suggest a multiple receiver, but would accept a single receiver if that made things easier.  

It would then simply do a filter to show the rows where the values are non-numeric.

Later there might be options, for example it could provide a count of the number of non-numeric rows in the output window.

We need the R code of course.  From @dannyparsons 
is.na(x) != is.na(as.numeric(x)) comes close, but I thought it also includes missing values as non-numeric.  Perhaps it is all we need?  Otherwise it needs to be adapted a bit.  

Once we have a more general dialogue for filters, then it should also be possible to use the filter dialogue directly.  But, with our audience it would still be useful to have a special dialogue.

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes #4766

### DIFF
--- a/instat/dlgShowNonnumericValues.vb
+++ b/instat/dlgShowNonnumericValues.vb
@@ -44,13 +44,13 @@ Public Class dlgShowNonnumericValues
 
         ucrReceiverColumn.SetParameter(New RParameter("x", 1))
         ucrReceiverColumn.SetParameterIsRFunction()
-        ucrReceiverColumn.SetIncludedDataTypes({"character"})
+        ucrReceiverColumn.SetExcludedDataTypes({"numeric"})
 
-        ucrSaveLogicalColumn.SetPrefix("logical")
+        ucrSaveLogicalColumn.SetPrefix("nonnum")
         ucrSaveLogicalColumn.SetSaveTypeAsColumn()
         ucrSaveLogicalColumn.SetDataFrameSelector(ucrSelectorShowNonNumericValues.ucrAvailableDataFrames)
         ucrSaveLogicalColumn.SetIsTextBox()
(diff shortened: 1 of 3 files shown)


instat/dlgShowNonnumericValues.vb
```diff
@@ -14,8 +14,90 @@
...
+        TestOKEnabled()
+    End Sub
+
+    Private Sub InitialiseDialog()
+        ucrReceiverColumn.Selector = ucrSelectorShowNonNumericValues
+        ucrSelectorShowNonNumericValues.bUseCurrentFilter = False
+
+        ucrReceiverColumn.SetParameter(New RParameter("x", 1))
+        ucrReceiverColumn.SetParameterIsRFunction()
+        ucrReceiverColumn.SetIncludedDataTypes({"character"})
```
dannyparsons: Instead exclude numeric because factors could also go here.

shadrackkibet: okay.

maxwellfundi:
@dannyparsons this is now ready for you

maxwellfundi:
@shadrackkibet  have you added this to the frm main yet?

shadrackkibet:
Yes i have now added that.

maxwellfundi:
great
Regards
Maxwell Fundi
|| Phone :+254728508241||

PR #4778, opened 2018-07-17
 Importing shape files 

jackykemboi:
This is the initial implementation of the dialog. You can now import shape file, convert to data frame and load to R instat. @shadrackkibet and I have managed to do this as shown in the figure below;
image
@dannyparsons @maxwellfundi @rdstern @volloholic

### DIFF
--- a/instat/dlgImportShapeFiles.vb
+++ b/instat/dlgImportShapeFiles.vb
@@ -67,7 +67,7 @@ Public Class dlgImportShapeFiles
         Dim strTemp As String = ""
 
         Using dlgOpen As New OpenFileDialog
-            dlgOpen.Filter = "All Shape files|*.shp;*.dbf"
+            dlgOpen.Filter = "Shapefiles|*.shp|All Shapefiles|*.shp;*.dbf;*shx"
             dlgOpen.Title = "Import Shape File"
 
             If bFromLibrary Then
@@ -85,7 +85,7 @@ Public Class dlgImportShapeFiles
                     strFilePath = dlgOpen.FileName
                     strFileExt = Path.GetExtension(strFilePath)
                     ucrInputFilePath.SetName(Replace(strFilePath, "\", "/"))
-                    If strFileExt = ".shp" OrElse strFileExt = ".dbf" Then
+                    If strFileExt = ".shp" OrElse strFileExt = ".dbf" OrElse strFileExt = ".shx" Then
                         ucrSaveDataframeName.SetName(frmMain.clsRLink.MakeValidText(strFileName))
                         clsReadShapeFiles.AddParameter("dsn", Chr(34) & Replace(strFilePath, "\", "/") & Chr(34))
                     End If
--- a/instat/dlgImportShapeFiles.vb
+++ b/instat/dlgImportShapeFiles.vb
@@ -67,7 +67,7 @@ Public Class dlgImportShapeFiles
         Dim strTemp As String = ""
 
         Using dlgOpen As New OpenFileDialog
-            dlgOpen.Filter = "All Shape files|*.shp|Shape files|*.shp"
+            dlgOpen.Filter = "All Shape files|*.shp;*.dbf"
             dlgOpen.Title = "Import Shape File"
 
             If bFromLibrary Then
@@ -85,7 +85,7 @@ Public Class dlgImportShapeFiles
                     strFilePath = dlgOpen.FileName
                     strFileExt = Path.GetExtension(strFilePath)
                     ucrInputFilePath.SetName(Replace(strFilePath, "\", "/"))
-                    If strFileExt = ".shp" Then
+                    If strFileExt = ".shp" OrElse strFileExt = ".dbf" Then
                         ucrSaveDataframeName.SetName(frmMain.clsRLink.MakeValidText(strFileName))
                         clsReadShapeFiles.AddParameter("dsn", Chr(34) & Replace(strFilePath, "\", "/") & Chr(34))
                     End If
--- a/instat/dlgImportShapeFiles.vb
+++ b/instat/dlgImportShapeFiles.vb
@@ -23,7 +23,7 @@ Public Class dlgImportShapeFiles
     Private bFromLibrary As Boolean = False
     Private strFilePath As String = ""
     Private strLibraryPath As String = Path.Combine(frmMain.strStaticPath, "Library", "Climatic", "Shapefiles/")
-    Private clsReadOGRFunction, clsMergeFunction, clsTidyFunction, clsAsDataFrameFunction As New RFunction
+    Private clsReadShapeFiles As New RFunction
 
     Private Sub dlgImportShapeFiles_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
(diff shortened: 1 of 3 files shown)


maxwellfundi:
well done @jackykemboi for creating your first dialog. I find it okay as of the earlier design we all discussed and conceptualized. @dannyparsons @rdstern  have a look at this and give feedback.

rdstern:
This looks great.  Is it ready to be merged.
For information, 

1. I assume there are a variety of sources of shape files.  Does this dialogue read lots of different types or only from one source?
2. When importing the netcdf files (which is the closet equivalent - and which is working very well) the import also includes quite a lot of the meta-data, e.g. units.  Is there much metadata associated with a shape file?  If so, then is this imported and available?

instat/dlgImportShapeFiles.vb
```diff
@@ -20,8 +20,9 @@ Imports System.IO
 Public Class dlgImportShapeFiles
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
+    Private bFromLibrary As Boolean = False
     Private strFilePath As String = ""
-
+    Private strLibraryPath As String = Path.Combine(frmMain.strStaticPath, "Library", "Climatic", "Shapefiles/")
```
maxwellfundi: @JacklineKemboi @shadrackkibet I would suggest we lead people just onto the library not to the shape files like we do for everything else. 
You could have `Private strLibraryPath As String =Path.Combine(frmMain.strStaticPath & "\Library")`

instat/dlgImportShapeFiles.vb
```diff
@@ -91,6 +88,14 @@ Public Class dlgImportShapeFiles
             dlgOpen.Filter = "All Shape files|*.shp|Shape files|*.shp"
             dlgOpen.Title = "Import Shape File"
 
+            If bFromLibrary Then
+                dlgOpen.InitialDirectory = Path.GetDirectoryName(Replace(strLibraryPath, "/", "\"))
+            ElseIf Not ucrInputFilePath.IsEmpty() Then
```
maxwellfundi: Do you want to have `bFromLibrary = False` after setting the InitialDirectory

jackykemboi:
- @rdstern the dialog reads shape files from different source. 

- The import includes the meta data e.g for the case of Rwandan shape file the meta data from the source is:
image


This metadata is included except the description as shown:

image

- The dialog is   ready to be merged

instat/dlgImportShapeFiles.vb
shadrackkibet: yes , that is done on click event.

instat/dlgImportShapeFiles.vb
shadrackkibet: Yes we want the user to go to the shape files directory. That is where we are suggesting we have the shapefiles.

instat/dlgImportShapeFiles.vb
```diff
@@ -33,16 +39,109 @@ Public Class dlgImportShapeFiles
...
+        Else
+            ucrBase.OKEnabled(False)
+        End If
+
+    End Sub
 
+    Private Sub cmdLibrary_Click(sender As Object, e As EventArgs) Handles cmdLibrary.Click
+        bFromLibrary = True
+        GetFileFromOpenDialog()
+        bFromLibrary = False
```
shadrackkibet: @maxwellfundi this is where we are setting this to false `bFromLibrary = False`

instat/dlgImportShapeFiles.vb
maxwellfundi: Okay sounds good then, if we want the library button to lead to the shapefiles then this is fine.

instat/dlgImportShapeFiles.vb
maxwellfundi: Okay i see where that is used... Looks good.

instat/dlgImportShapeFiles.vb
maxwellfundi: This is okay actually, i had not seen where it was used.

shadrackkibet:
I suggest we merge this so that @rdstern  can test it as well.

maxwellfundi:
@rdstern your queries have been addressed. @dannyparsons this is dialog is now ready for your review.

instat/dlgImportShapeFiles.vb
```diff
@@ -33,16 +39,109 @@ Public Class dlgImportShapeFiles
...
+        clsAsDataFrameFunction = New RFunction
+
+        ucrSaveDataframeName.Reset()
+
+        clsReadOGRFunction.SetPackageName("rgdal")
+        clsReadOGRFunction.SetRCommand("readOGR")
+
+        clsMergeFunction.SetRCommand("merge")
+        clsMergeFunction.AddParameter("tidy", clsRFunctionParameter:=clsTidyFunction, iPosition:=0, bIncludeArgumentName:=False)
+        clsMergeFunction.AddParameter("as.data.frame", clsRFunctionParameter:=clsAsDataFrameFunction, iPosition:=1, bIncludeArgumentName:=False)
```
dannyparsons: For merging I suggest you use dplyr's join functions instead of the base merge as it is much faster, and also then consistent with the merge dialog.

dannyparsons:
It would be very nice to have the Description from the metadata imported into R-Instat. Have you looked into this? It could be that this is available in the shape object that is returned by `readOGR`. It would be good to see what else is stored in this object as that extra information is being lost when converting to a data frame.

dannyparsons:
I have been looking at an example that Jackline sent me. She had two shape files. One for the regions of Rwanda and one for the cities.

So in R the difference is that the regions import as a `SpatialPolygonsDataFrame` class and the cities is a `SpatialPointsDataFrame`. 
The dialog then uses `broom::tidy` but this only works with a `SpatialPolygonsDataFrame` as it extracts the polygons into a data frame. For a `SpatialPointsDataFrame` it only needs to be directly converted to a data frame to get each point as a row.

So from my understanding, this means that the dialog will currently only works for polygon shape files and not point or line shape files because `broom::tidy` will fail if it is not polygons? @JacklineKemboi is that correct? And if so, do you have an idea of what we might do for non-polygon shape files?

jackykemboi:
>So from my understanding, this means that the dialog will currently only works for polygon shape files and not point or line shape files because broom::tidy will fail if it is not polygons? @JacklineKemboi is that correct? And if so, do you have an idea of what we might do for non-polygon shape files?

@dannyparsons The dialogue works for spatial polygons. We want to add a condition if the shape file isn't a polygon.

dannyparsons:
I think the main use will be for polygons as it's easier to get point data in a standard format already, but since it's easy to import these as well I think the dialog should cope with the different types.
So to do this I think the class of the object will need to checked when importing and this will determine if it's polygon or not then you can do the appropriate thing depending on the type.
Since this is turning into more of a script then a few lines of code I suggest you create a simple `import_shape` function in R and call that from the dialog instead now.

maxwellfundi:
Testing this, and just a note, package called `sf` is needed for this to work

maxwellfundi:
@jackykemboi  @shadrackkibet  This dialog seems not to do anything when I click reset. Possibly no code is being run.

jackykemboi:
@maxwellfundi Reset is now fixed.

instat/dlgImportShapeFiles.vb
```diff
@@ -52,7 +52,7 @@ Public Class dlgImportShapeFiles
     Private Sub SetDefaults()
         clsReadShapeFiles = New RFunction
 
-        ucrSaveDataframeName.Reset()
+        ucrSaveDataframeName.SetName("")
```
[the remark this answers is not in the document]
shadrackkibet: I think we need this because we do have a default for this control anywhere in the dialog and we want to clear this when we reset.

shadrackkibet:
@dannyparsons  this working fine.  This is ready for you.

instat/dlgImportShapeFiles.vb
```diff
@@ -33,16 +39,91 @@ Public Class dlgImportShapeFiles
...
+        ucrBase.clsRsyntax.SetBaseRFunction(clsReadShapeFiles)
+    End Sub
+
+    Public Sub GetFileFromOpenDialog()
+        Dim strFileName As String = ""
+        Dim strFileExt As String = ""
+        Dim strTemp As String = ""
+
+        Using dlgOpen As New OpenFileDialog
+            dlgOpen.Filter = "All Shape files|*.shp|Shape files|*.shp"
```
dannyparsons: I don't think we need two options here which are the same extensions.

shadrackkibet: Yes, this should be `.shp | .dbf ` extensions

shadrackkibet: @dannyparsons This is now done.

dannyparsons:
Why are `dbf` files included here? From what I have read about them they are database files https://whatis.techtarget.com/fileformat/DBF-dBASE-file

shadrackkibet:
From what i read [here](https://blogs.library.duke.edu/data/2015/09/14/shapefiles-vs-geodatabases/) and [here](https://gis.stackexchange.com/questions/28105/how-do-i-load-a-dbf-file-into-qgis/28109) is that a valid Shapefile must contain at least `.shp, .dbf` and `.shx` files. 

The reason why we included `.dbf` file is because  we didn't find any difference in using any of the three file extensions so long as you have all three in the same place. 
So , i think we should also include `.shx` extension.

dannyparsons:
Ok so there could be a case where you want to choose the dbf or shx or just see if these files are there. So I suggest we have two filters: 1. shp and 2. shp, dbf and shx.

If there's no reason at the moment not to choose the shp file then let's have 1. as the default and 2. there as an option.

dannyparsons:
And from your second link and [here](https://en.wikipedia.org/wiki/Shapefile) there are other optional files but I don't see that we need to show them for the moment.

maxwellfundi:
@dannyparsons  The last issue here was solved so this is now ready to go in.

PR #4786, opened 2018-07-18
Access R documentation dialog

### LINKED ISSUE #4762: Access to R documentation in Help menu
In the File > Open from Library we can get access to any of the library data sets, and also the help information that describes the data.
Could we have a menu item in the help menu that gives similar access to the information on each function in the package?  I expect this to work in a similar way.
1. To provide this information (in a browser).
2. To have a full list of the packages accessible from R-Instat  (Whether they have data sets or not.)
3. The information includes examples.  These will usually run from our script window.  That would be a great feature!
.

shadrackkibet:
Fixes #4762 . @africanmathsinitiative/developers .This is ready for review and testing.


Muthenya:
Reviewing.

Muthenya:
The dialog gives the link but the html page does not load, i.e, nothing is shown on web page.

maxwellfundi:
@Muthenya i get a similar thing here too with this dialog. Not sure why. In fact, if you try using this dialog and then go click on the help from the import from library dialog for the packages, that doesn't work either and the vice versa works

dannyparsons:
It sounds like there's something causing this if it's affecting multiple machines. Can you systematically try to replicate when it does and doesn't work? Check things like other R programs being open and anti virus...

shadrackkibet:
@dannyparsons seems there is a problem which i am not yet able to figure out here. 

When i first run the software and  use import from library dialog and then use this dialog to get help this works properly. Vice versa doesn't work for this. Could you please have a look into this?

dannyparsons:
Have you tried the options for running the script like not running in a separate thread?

shadrackkibet:
> Have you tried the options for running the script like not running in a separate thread?

Okay, i will try to see if it works.

shadrackkibet:
@Muthenya it will be good if you give a detailed explanation of what exactly you did to get this errors.At what instance did this work/not work? when did this crash?.

Muthenya:
That's the thing @shadrackkibet, the bug doesn't seem consistent, but just to describe how I produced the bug;

The first instance I loaded the dialog, I just clicked okay and it gave me the html page. Then I clicked on the **Vignettes** option and `clicked okay but it didn't give the html page, on the code in the Output Window. I then went back to the **Help** option and changed the package name to **ggplot2** and inserted the function name **ggplot**. It did give the html page but now I crashed out.

The second instance, I loaded the dialog and immediately pressed okay. The output was give but I crashed out.

The third instance, I clicked on **Vignettes** and changed the package to **ggplot2**, clicked okay and I crashed out. It did the same when I tried with **dplyr**, i.e, for the fourth time.

The issue might be around changing of the package name and adding the function name. I would like that others test it as well to see if they will encounter the same bug.

dannyparsons:
@Muthenya When you do this can you check the R code that is being produced and whether it is correct. First we need to know if the R command is correct or if that is causing the error.
And, what do you mean by "crash out"? Is there any error message?
Try to do more systematic testing, if you think it could be caused by changing the package name, try running the dialog many times without changing the package name, what happens?
If that is ok, then try changing the package name without adding a function name, does that always work? Do some packages work but not others? Or does it sometimes work and sometimes not with the same code?
To understand this you need to be testing where you only change one thing at a time because there are too many options to know what could be the cause.

shadrackkibet:
@dannyparsons could you please have a look at this. Seems this problem is deeper in the code!

dannyparsons:
Yep I will look at this

shadrackkibet:
@dannyparsons did you find time to look at this?

dannyparsons:
I'll look at this before the next version

maxwellfundi:
@dannyparsons please also have a look at this at some point too for the next version.

dannyparsons:
I'm not sure of the issue yet so I don't think it will make it in this version.

shadrackkibet:
There a problem here and i am going to explain how to reproduce it:

- First go to **vignettes** option , then change package to **dplyr** and click Ok.
- Go to **Help** option and then select **dplyr** package  and click Ok.
- Go back to **Help** option and then select **dplyr** package and then input `select `  on the function name textbox.
- Click Ok.

I tried doing this when rstudio is open and closed and this problem was reproducible.

you get a bug as shown below:
image
 
@dannyparsons  I am handing this over to you as this problem seems to be deeper in the code.

rdstern:
I tried to test where we are with this improvement.  I get:
image

So I can not get further.  Could @Patowhiz or @maxwellfundi see if there is something obvious here?  Otherwise perhaps the other gurus?

shadrackkibet:
@rdstern please test this.

PR #4790, opened 2018-07-19
Length of the  season dialog

### LINKED ISSUE #4744: New climatic dialogue for length of season?
I propose an additional (simple) dialogue to find the length of the season.

If agreed, then it will go just under the dialogue to calculate the End of the Rains.

It is a very simple dialogue which (unlike the others in this menu) uses the summary data that has been produced on the start and end of the season.

It is (essentially) just the calculation of the End minus the Start.  It is currently done with the Prepare > Column: Calculate > Calculations.  The reason I am suggesting a dialogue is also to cope with the information in the logical columns (for the start and end) at the same time.

It will produce a column of lengths, plus a factor column.  The factor column has 4 levels (plus NA).

The dialogue has our usual selector and 4 receivers.  They Are labelled Start, Start(L), End, End(L).  The Start and End columns are needed for the command.  They may be numeric or dates. The others may be optional - they are restricted to logical columns.

Underneath are the 2 columns to be saved, called Length and Type.

If the Start and End are numeric, then the Length is simply (End - Start).  If Length is <=0 it is set to missing and the Type column is set to be "Not Positive".

The Type column had 4 levels, namely Normal, Not Positive, No Start and No End.  It is also NA when there are missing values.

This may need Beth's help in the R code, for the logical columns.

I also welcome confirmation from Maxwell, David and Danny on the value of this new dialogue.

shadrackkibet:
Fixes #4744 In progress!

### DIFF
--- a/instat/dlgClimaticLengthOfSeason.Designer.vb
+++ b/instat/dlgClimaticLengthOfSeason.Designer.vb
@@ -41,7 +41,7 @@ Partial Class dlgClimaticLengthOfSeason
         'lblStartofRains
         '
         Me.lblStartofRains.AutoSize = True
-        Me.lblStartofRains.Location = New System.Drawing.Point(304, 31)
+        Me.lblStartofRains.Location = New System.Drawing.Point(300, 31)
         Me.lblStartofRains.Name = "lblStartofRains"
         Me.lblStartofRains.Size = New System.Drawing.Size(32, 13)
         Me.lblStartofRains.TabIndex = 1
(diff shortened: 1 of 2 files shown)


shadrackkibet:
@africanmathsinitiative/developers this is now done. This is ready for review and testing.

shadrackkibet:
@rdstern  here is the design for this dialog.
image

rdstern:
Looks great.  

I would just change the name type to length_status (for now at least)!

shadrackkibet:
@africanmathsinitiative/developers this is still ready for review!

maxwellfundi:
@dannyparsons  this is ready for your review. It works well

Issue #4529, opened 2018-04-14
Improving the filter sub-dialogue


rdstern:
Currently it is easy to choose levels of a factor in the filter sub-dialogue.  But the system with other types of column would be helped if a keyboard could be added.  
With the current layout there could at least be some help for users, e.g. a pull-down for what is typed that includes NA and some other examples.  Better could be a simple keyboard with the numbers, a minus and also NA.  (Anything else?  Does a character column need quotes, if so, these should be added.

Could there possibly also be an option - perhaps one of our new radio buttons that says General as opposed to Basic or Simple.  If General then there is our whole logical keyboard and we just give the condition exactly as in the calculator, possibly including the Try facility, which is useful.

One use of this is to add the OR possibility to the filter, rather than having to go through the calculator and produce the column first.  The example I had on the workshop was 
(Tmin == NA) & (Tmax == NA).  This was actually not what was wanted - we needed OR (|) rather than &.  But that doesn't work in the calculator.  - you use is.na(Tmin) instead.  (I am still curious as to how it works in the filter!)

This is for discussion.  I realise there are many improvements needed to the filter.

maxwellfundi:
In addition, it got people very confused when, the year was a factor in the data and they wanted to do a filter of the year > 1970. 
So when the year is selected while still a factor, it displays the grid showing the levels of factors, which of course you can tick the ones you want. However, to do a proper filter on this, it was needed to get the year as a numeric column, but when you come back on to the filter dialog, while the year is selected it still shows the grid with the levels of factor. It needed to have the year removed from the receiver and then added afresh to be able have the input to type in 1970 and the drop down with the symbols. Can we keep updating the column type when this dialog is reopened?

dannyparsons:
Updating the display when reopened should be done, yes.

I wouldn't have thought a keyboard for numbers would be needed since its just a single number being entered but if users would like that there no problem. NA is useful but should be a drop down item in the textbox which should become a combo box, because NA should be there on its own if you're using it. I'm not sure why you would need TRUE and FALSE for a numeric column? The evaluation should always be against a number or NA. For logical columns TRUE, FALSE and NA could be in the drop down and you would not be able to type anything else in because there are no other options.

I don't see anything difficult about any of these changes so someone could take this up.

Having the option of AND or OR is also just a small change in the R code if someone wants to investigate that.

rdstern:
In discussion on 26 June the suggestions were clarified.
[...]

PR #4794, opened 2018-07-20
New filter

Muthenya:
@africanmathsinitiative/developers this contains the addition of date-picker and setting up of a drop down with options for logical, numeric and character types of data. It's still a work in progress especially on part of the date-picker and more work needs to be done here.

The remaining aspect is the issue of the updating of the display. A temporary fix was proposed of clearing the receiver when the column data type is changed but I need your take and suggestions on how to implement this properly since this change clears the receiver every time the sub dialog loads. Otherwise, I'm looking forward to getting a better solution from the team.


Muthenya:
@africanmathsinitiative/developers this is open for review and suggestions so that work on it may continue.

Muthenya:
@africanmathsinitiative/developers this still requires review and suggestions.

maxwellfundi:
What issue is this addressing @Muthenya

Muthenya:
There are some minor changes there were to be made as mentioned in issue #4529. The last comment by Roger.

Muthenya:
@maxwellfundi I have made the added the drop down for the NA values for date. Please have a look at it again and let me get your insight on how to go about the updating of the display since I'm still stuck on it. Otherwise, if this can be merged it will be good so that the discussion on how to improve the filter can continue. Still without it being merged the discussions may still continue.

dannyparsons:
Can you share how this looks with the new NA options?

Muthenya:
This is how it looks.

image

dannyparsons:
The date picker looks good. 
But I don't understand how the NA option works - both are visible so how do I use the NA combobox?

Muthenya:
When the NA combo box is filled, its value, i.e "NA", is what will be taken. When the combo box is empty, the date picker value is taken.

dannyparsons:
But both are visible, so that's confusing. What if I change the date in the date picker but with NA still filled, that will take NA and not the new date I selected, also confusing. 
Is a combo box really the right control here? Combo boxes are for list of values, is this just NA and blank?

I think discuss as a group to try and get something that's more intuitive here. The main condition is that only the value being used should be visible. I don't think a combo box is right for this. We can't incorporate NA into the data picker, so there are two options: choose a date or NA. There are different ways that could be done.

dannyparsons:
I have a suggestion of having two additional items in the operation combo box **"== NA"** and **"!= NA"**. This would be quicker because you wouldn't need to do anything different for other data types, date, numeric, character etc. all use that combo box.
And NA is different and not like other values, I shouldn't be able to do `> NA` for example, which is possible if we allow NA to be a value.

Muthenya:
I agree. Having it on the operations is much easier.

Muthenya:
@dannyparsons this is the comment from Roger containing the improvements I was to make on the Filter from issue #4529.
> In discussion on 26 June the suggestions were clarified.

1.  Updating the display as mentioned above should be done.
2. Adding the simple improvements, should also be done, namely
a) When date is chosen there should be a date picker. Also dates could be NA.
b) With numeric and character columns NA should be on a pull-down.
c) With logical columns NA, TRUE and FALSE should be on a pull-down.

These were what I was working on. So for:
1. I have not been able to set this up well (the display isn't updating properly).
2. a) The date picker is set up. But not the NA.
    b) This has been done.
    c) This has also been done.

dannyparsons:
Is it possible this is something you could now do with @shadrackkibet?

Patowhiz:
@Muthenya I'm sorry this took me a while. I've thought about this issue and I think I have a good suggestion. The original problem, I think was to make it easier for people to create "date filters" by not having to 1. guess the format, 2. not having to add the single quotes.  3. And probably the NA issue.

I think having this on the operator beats the purpose of the operator control. Because ==NA is NOT and operator.  My suggestion will be to extend the textbox such that, on click or on type , based on the column type selected an auto suggestion in terms of the structure of the input expected (including the NA) is shown to the user. For instance if I select Date column and click on the textbox, then a date format like 'dd/mm/yyyy' pops up above it even as I type. Below is a simple sketch to give you a rough idea. Then based on the column datatype we can add the additionals like single quotes to the user input if we don't want them to type it. We can discuss this further if the @africanmathsinitiative/developers  team supports the idea.

filter sketch

Patowhiz:
The above option will remove all the complexities of hiding and removing controls, makes it simple to the user(user won't have to memorise anything), and our code becomes simpler

dannyparsons:
The problem is that we are using the date picker control to choose a date, so it's not possible to type in NA. We also don't want to be typed in because Na or na will not be accepted. I see your point that `is.na` or `!is.na` is not an operator, however doing `== NA` is incorrect R syntax so that should be avoided. We could rename operator to operation and then `is.na` and `!is.na` is valid. Doing it this way prevents users doing strange comparisons like `> NA` that you could do with other values.

Patowhiz:
@dannyparsons and that is what the autosuggestion will be telling the user. It gives tells the user the options ALLOWED. For NA, it can easily be omitted when the operator selected is NOT != or ==

Patowhiz:
Well if that's the case then my option won't help.

rdstern:
I have had a look at what we have here so far.  
1. I like the little "pull-down" with the TRUE and FALSE for logical columns.
2. When I tried this with a logical column it gave be the date picker, so there is a small problem here.
3. I also like the idea that we could add is.na and !is.na and call it Operations.
4. Should there also be a simple keyboard for numeric (and integer) columns, just with the numbers and the "." perhaps.
5. If we have this little keyboard, then would a simple alternative to the pull-down be to have the keyboard also for logical columns and add TRUE and FALSE to the keyboard?  Aren't they treated as 0 and 1 otherwise.  And is there a problem typing x < 6 for a logical - it might always be true, but might not cause a problem?
6. Could the rhs easily "remember" what has been used before - at least in the session - as it does in the calculator?
7.  It is a sort of halfway dialogue.  So could we consider having a Try control, as we do on other dialogues?
8. When there is a mistake in the condition could the error message be changed.  This "you should restart" message is a bit drastic.

dannyparsons:
I like all those suggestions - except adding a keyboard in the case of logical columns. Although things like `x < 6` is possible because it will be treated as 0 and 1 (but then everything is less than 6 so filter does nothing), I can't see any advantage to it and I can't think of a filter on logical which wouldn't be `== TRUE` or `== FALSE` therefore the drop down is all that's needed and easy to use.

rdstern:
This has been quiet for a long time.  The filtering continues to be so useful, so is there someone who could re-start work on this issue.  
If so, then I note the proposed addition above of a simple keyboard, was (sort of) agreed by @dannyparsons except for logical variables.  
So now I am getting ambitious.  Sometimes the condition has multiple elements, e.g. Year > 1957 & Day >29.  I am happy that & is assumed, but could we add or i.e. | somewhere?  A dangerous way would be to allow the preview to be editable!  Could we instead add an "Operator" at the front of the control where the 2 columns are currently Variable and then Condition.  It is blank for the first condition.  When there are 2 or more conditions it has a &, that could be editable to |?

rdstern:
This is another pull request that needs an update to use the new version of R.  It says:

image

Is this something I can do - if not, then can it be done?

rdstern:
@dannyparsons and @volloholic I am up to filters in the help documentation.  While you are together is this a possible time to resurrect this work on filters or, at least to plan how they might improve in the not-too-distant future.  What is here is improvements to the keyboard.  That would be most useful.

It would be great to consider also whether it is too much work to add filters to the list of objects we can manage - like models and graphs.  Then they could also be renamed and deleted.  (They are already remembered as part of the metadata.)  I am less concerned about editing filters, which (I presume) is a bigger task.

Patowhiz:
@maxwellfundi  I think this can be closed.

rdstern:
This looks like things that we could now close?  @dannyparsons is it.  I am writin g more issues on filters and this has gone on a long time.

shadrackkibet:
This can be closed.

Issue #4659, opened 2018-06-07
New dialogue to reshape climatic data


rdstern:
In the Haiti workshop I decided to include the steps to reorganise daily data into the "shape" needed.  This was tough and - in hindsight - unnecessary.

When data are a real mess the initial steps are usually taken outside R-Instat.  There are then 3 common shapes:
a) One column (variable) per year.  This is the old Instat.
b) One variable per month - so the 12 months have to be stacked.  There is a day-of-month column.
c) One variable per day of the month - so the 31 days have to be stacked.  One row is a month.

If a) then it is a single element from a single station in a file (or Excel sheet).  It is usually rainfall.  
If b) Then there will also be a year variable, possibly a station variable and possibly an element variable.
If c) then there will be a year variable.  There can be a station variable, and there can be an element variable.

dannyparsons:
Yes this would be good to have as a dialog.

We also considered whether it should be able to do this for multiple files/data frames in one go e.g. if you have the format in a) for the same element for 20 stations it could do them altogether using the sheet name as the station ID.

Should the dialog take file(s) as the input or data frame(s)? i.e. do they have to import the data into R-Instat first and then use the dialog to reshape, or do it directly as a special import?

rdstern:
I give a possible design for the dialogue below.  I would also really like it to be able to use multiple data frames, but would be happy if we started with just one for now.
In the use of multiple data frames, as Danny says, this could then be through multiple files, or the import could be done first and then the multiple data frames are used.  I would be happy with (just) multiple data frames.  This is for 2 reasons:
1) It keeps life simple.  We also work on making it simple to import multiple (identical) files, or multiple identical sheets from Excel.  But we don't get clever by mixing the two.  Both these steps are pretty ambitious, perhaps for us, but also for users.
2) I am a believer in encouraging users to look at data before doing ambitious steps.  So, once imported, then perhaps the names of the sheets need to be changed.  And perhaps some are not quite as identical as we thought.
So I am quite happy that this remains a 2-step process.

rdstern:
There are 3 or our modern buttons at the top.  Perhaps a label first, on the left **One column per:**.  Then the first button says **Year**, the second says **Month (1 to 12)**, and the third says **Day (1 to 31)**.

Then there is our usual data selector on the left.

On the right, start with the second and third options.  

The second option is one column per month.  So there will be 12 monthly columns and they are always (in examples I have seen) Jan to Dec in that order.
[...]

PR #4813, opened 2018-07-25
Tidy climatic

Muthenya:
In progress.


instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -1120,6 +1120,94 @@ compare_columns <- function(x, y, use_unique = TRUE, sort_values = TRUE, firstno
...
+  
+  y$Date<-as.Date(paste(y$year,y$month,y$day,sep="/"))
+  
+  y<-subset(y,is.na(Date)==FALSE)
+  if(length(sortcols)==0){ y<-y[order(y$Date),] }
+  if(length(sortcols)>0){ y<-y[order(y[,sortcols],y$Date),] }
+  rownames(y)<-NULL
+  return(y)
+  
+
```
dannyparsons: Missing a final `}` here. This causes errors when starting R-Instat.

instat/dlgReshapeClimatic.vb
```diff
@@ -1,3 +1,265 @@
...
+        'ucrReceiverDayofYear.SetParameter(New RParameter(""))
+        'ucrReceiverDayofYear.SetParameterIsString()
+        'ucrReceiverDayofYear.bAutoFill = True
+
+        'ucrReceiverMonth.SetParameter(New RParameter(""))
+        'ucrReceiverMonth.SetParameterIsString()
+        'ucrReceiverMonth.bAutoFill = True
+
+        ucrReceiverStation.SetParameter(New RParameter("sortcols", 6))
+        ucrReceiverStation.SetClimaticType("station")
```
dannyparsons: The station receiver is always empty because of this. It shouldn't be set as any climatic type because this dialog is used before the data is defined.

instat/dlgReshapeClimatic.vb
```diff
@@ -1,3 +1,265 @@
...
+        ucrReceiverStation.SetParameterIsString()
+        ucrReceiverStation.bAutoFill = True
```
dannyparsons: This too shouldn't be

instat/dlgReshapeClimatic.vb
```diff
@@ -1,3 +1,265 @@
...
+        ucrReceiverLastDay.Selector = ucrSelectorReshapeClimaticData
+
+        'rdoYear
+        ucrReceiverFirstYear.SetParameterIsString()
+        ucrReceiverFirstYear.SetParameter(New RParameter("first", 4))
+        ucrReceiverFirstYear.bAutoFill = True
+
+        ucrReceiverLastYear.SetParameterIsString()
+        ucrReceiverLastYear.SetParameter(New RParameter("last", 5))
+        ucrReceiverLastYear.bAutoFill = True
```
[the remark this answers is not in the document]
Muthenya: Yes, this should not be there. Let me remove them.

instat/static/InstatObject/R/stand_alone_functions.R
Muthenya: I think this was removed when the conflict was being resolved. I have put it back.

dannyparsons:
Great, once you've made those changes I'll work on this branch to add a few extensions of the R code. Then I'll send it back to you as there will likely be some changes to the dialog after that.

Muthenya:
@dannyparsons you may have a look at it again.

dannyparsons:
related issue is here #4659

Muthenya:
Okay.

dannyparsons:
This is looking good. Some comments:

- `ignore_invalid` should be a checkbox, and the default is `FALSE` not `TRUE`
- `silent` parameter should be a checkbox
- The first time I ran it, I left "element name" blank and the code added `element_name = ""`, and it didn't produce the element column! This is an issue of the R code as well (I'll look into), but the default for `element_name` is "value", this should be in the textbox by default.
- Element name text boxes doesn't have correct validation. It's a column name, so should be a valid R name
- The default name could have `_tidy` at the end instead of the start.
- Element receiver could be renamed "Multiple Elements" to avoid confusing with this meaning on other dialogs
- Only one of "Multiple Elements" and "Element Name" should be entered, "Element Name" is to name the single element column that will be created. Possibly put these two in a groupbox and have an "or" label in between them. Also move "Element Name" above "Multiple Elements" as this is more common.
- Maybe reorder the receivers(?) to have: stack columns, year, month, day (if appropriate), station, element name, multiple elements. Not sure about this, or if Roger will agree, but that puts them in order of importance. It's inconsistent with other climatic dialogs, but because Station is less common here. Roger can have the final say.
- Current receiver doesn't change when changing layout. If I'm on "Element" receiver and then switch to "Year Columns" I don't have a current receiver.
- We're not using Station, Day and Month columns in the "Year Columns" format anymore (they might be added later), they can be removed.
- But "Element Name" should be in the "Year Columns" format.
- The top of the dialog doesn't look good, I don't see why the Layout label is there. It could be on top or to the side of the buttons so it at least lines up. We could get rid of it altogether.
- "Columns to Stack" is not a great label, I don't think users will know what this means, it's not obvious that it's doing a stack. The label could change based on the format "Year Columns", "Month Columns (12)", "Day Columns (31)". The numbers show how many columns are expected. And actually I think these could be the labels at the top too. I think "12" is better than "1 - 12" because they don't need to be labelled 1 - 12, there just needs to be 12 of them.

Lots of comments, but it did work with data in all 3 formats so that's very good already!

Muthenya:
So I have done almost all the changes. Just a few areas that need to be sorted out.

1. I have put the Element name in the "Year Columns" format but it is still within the group  box. I'll share it so that I can get your input in this.
2. I haven't yet reordered the controls. Awaiting Roger's input.

3. I wanted to know how the **stack_years** parameter is going to be incorporated into the dialog.

Muthenya:
Currently this is what I have for the "Year Columns" format
image

This is similar for "Month Columns" and "Day Columns"
image

dannyparsons:
Looking at the layout, I am more convinced that the alternative order would be better, then it wouldn't matter about the group box size for the year option. I think go ahead and reorder them, we can always switch them back later.

`stack_years` is to give the list of years if the year column names are not in the correct format. I suggest we don't implement this yet.

The checkbox labels are not very informative, the user should be able to know what they will do from the labels. They should be more like the description of the parameters not the parameter names. So ignore invalid is doing something like "Ignore values on invalid dates". For silent, it's clearer to have a positive statement so something like "display details of warnings & errors" so if this is checked `silent = FALSE`.

dannyparsons:
Yes Element Value is the name to give the new element column (unless you have multiple elements). I thought of deliberately having a vague default instead of rain so that it would be clear if you forgot to change it because its just "value". I thought this should go above the Elements receiver because it's more common, and you could miss it if it's too far down.

Muthenya:
Something not very clear Danny. Should I include the entire name "display details of warnings & errors" for the silent checkbox or do you have another alternative? I feel it's a bit too long.

instat/dlgTidyClimatic.vb
```diff
@@ -203,7 +185,7 @@ Public Class dlgTidyClimaticData
     Private Sub ucrPnlReshapeClimaticData_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlReshapeClimaticData.ControlValueChanged, ucrReceiverMonthTwo.ControlValueChanged, ucrReceiverYear.ControlValueChanged, ucrReceiverYearTwo.ControlValueChanged
         UpdateParameters()
         ShowReceivers()
-        MultipleStackElementLocation()
+        ElementGroupLocation()
```
dannyparsons: These don't all need to be run on all these events. e.g. `ElementGroupLocation()` only needs to run when the panel changes, I'm not sure about the others. It's better to separate these out so they only run when needed, otherwise it slows the dialog.

instat/dlgTidyClimatic.vb
```diff
@@ -245,23 +227,34 @@ Public Class dlgTidyClimaticData
...
+            grpElements.Location = New Point(grpElements.Location.X, iTextBoxMaxY / 1.5)
         Else
-            lblColumnstoStack.Location = New Point(lblColumnstoStack.Location.X, iReceiverLabelMaxY)
-            ucrReceiverMultipleStack.Location = New Point(ucrReceiverMultipleStack.Location.X, iReceiverMaxY)
             grpElements.Location = New Point(grpElements.Location.X, iTextBoxMaxY)
         End If
     End Sub
 
+    Private Sub ElementAddRemoveParam()
+        clsTidyClimaticFunction.RemoveParameterByName("element_name")
```
dannyparsons: This should go inside the `If Not`

instat/dlgTidyClimatic.vb
Muthenya: Actually, there other events are not necessary so I have removed them.

Muthenya:
@dannyparsons please have a look at it again.

dannyparsons:
This looks better now and works well. A few comments:

- element name textbox doesn't have the right validation still
- I ran the dialog with month format. When I reopened, the label for the multiple receiver had changed back to "Year Columns"
- I think it would be easy to have just one Year receiver that is linked to Month and Day formats. If so, then that would be better, because now it looks like its clearing when switching between Month and Day, because they are not the same.
- I think the receiver labels should have colons at the end
- Few small alignment things: everything on left side should be in line with x location 10. I think it would look better if the element textbox and receiver aligned with the other receivers, instead of the edge of the groupbox, so shift to the left a bit. 
- The checkboxes could go in a groupbox called "Options"

instat/dlgTidyClimatic.vb
```diff
@@ -0,0 +1,260 @@
...
+    End Sub
+
+    Private Sub InitialiseDialog()
+        'ucrBase.clsRsyntax.iCallType = 0
+
+        ucrPnlReshapeClimaticData.AddRadioButton(rdoYear)
+        ucrPnlReshapeClimaticData.AddRadioButton(rdoMonth)
+        ucrPnlReshapeClimaticData.AddRadioButton(rdoDay)
+
+        ucrSelectorTidyClimaticData.SetParameter(New RParameter("x", 0, bNewIncludeArgumentName:=False))
```
dannyparsons: Any reason to not have the argument name?

instat/dlgTidyClimatic.vb
```diff
@@ -0,0 +1,260 @@
...
+        ucrReceiverMonth.SetLinkedDisplayControl(lblMonth)
+
+        ucrReceiverElement.SetLinkedDisplayControl(lblMultipleElement)
+        ucrTextBoxElementName.SetLinkedDisplayControl(lblElementName)
+
+        ucrNewDFName.SetIsComboBox()
+        ucrNewDFName.SetSaveTypeAsDataFrame()
+        ucrNewDFName.SetCheckBoxText("New Data Frame Name")
+        ucrNewDFName.SetDataFrameSelector(ucrSelectorTidyClimaticData.ucrAvailableDataFrames)
+        ucrNewDFName.SetAssignToIfUncheckedValue("last_data_frame")
```
dannyparsons: Saving is not optional, and we're not using the checkbox on the save control here so this isn't needed.

instat/dlgTidyClimatic.vb
```diff
@@ -0,0 +1,260 @@
...
+    End Sub
+
+    Private Sub SetDefaults()
+        clsTidyClimaticFunction = New RFunction
+
+        ucrSelectorTidyClimaticData.Reset()
+        ucrReceiverMultipleStack.SetMeAsReceiver()
+        ucrNewDFName.Reset()
+
+        clsTidyClimaticFunction.SetAssignTo(ucrNewDFName.GetText, strTempDataframe:=ucrSelectorTidyClimaticData.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
```
dannyparsons: These two strings should be the same. Look at e.g. Stack dialog to see how this is done there.

instat/dlgTidyClimatic.vb
```diff
@@ -0,0 +1,260 @@
...
+
+    Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
+        SetDefaults()
+        SetRCodeforControls(True)
+        TestOkEnabled()
+    End Sub
+
+    Private Sub NewDefaultName()
+        If (Not ucrNewDFName.bUserTyped) AndAlso ucrSelectorTidyClimaticData.ucrAvailableDataFrames.cboAvailableDataFrames.Text <> "" Then
+            ucrNewDFName.SetPrefix(ucrSelectorTidyClimaticData.ucrAvailableDataFrames.cboAvailableDataFrames.Text & "_tidy")
```
dannyparsons: Don't set as prefix here because we don't want an extra number at the end. Just setting the name is fine, as in e.g. Stack dialog

instat/dlgTidyClimatic.vb
```diff
@@ -0,0 +1,260 @@
...
+    Private Sub ucrPnlReshapeClimaticData_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlReshapeClimaticData.ControlValueChanged
+        UpdateParameters()
+        ShowReceivers()
+        ElementGroupLocation()
+        ColumnstoStackText()
+    End Sub
+
+    Private Sub UpdateParameters()
+
+        clsTidyClimaticFunction.RemoveParameterByName("format")
```
dannyparsons: No need to remove here because it will be replaced by the code below.

instat/dlgTidyClimatic.vb
```diff
@@ -0,0 +1,260 @@
...
+
+        If rdoYear.Checked Then
+            clsTidyClimaticFunction.AddParameter("format", Chr(34) & "years" & Chr(34), iPosition:=1)
+            clsTidyClimaticFunction.RemoveParameterByName("element")
```
dannyparsons: If the element receiver is linked to the Month and Day options of the panel then this will be done automatically. Like this you might have a bug where it doesn't get added back correctly.

instat/dlgTidyClimatic.vb
Muthenya: Not really. I can change that.

instat/dlgTidyClimatic.vb
Muthenya: Okay.

instat/dlgTidyClimatic.vb
Muthenya: I had set it this way since from an earlier comment from Roger he wanted the extra number for the output (new data frame). But I'll still change it.

dannyparsons: Ok if we want the numbers then that's fine.

Muthenya:
I have made the edits. I hope it is ready now.

instat/dlgTidyClimatic.vb
```diff
@@ -0,0 +1,257 @@
...
+        ucrChkIgnoreInvalid.SetText("Ignore values on invalid dates")
+
+        ucrChkSilent.SetParameter(New RParameter("silent", 10))
+        ucrChkSilent.SetValuesCheckedAndUnchecked("FALSE", "TRUE")
+        ucrChkSilent.SetRDefault("FALSE")
+        ucrChkSilent.SetText("Display details of warnings and errors")
+
+        ucrPnlReshapeClimaticData.AddToLinkedControls(ucrReceiverDayofMonth, {rdoMonth}, bNewLinkedHideIfParameterMissing:=True)
+
+        ucrPnlReshapeClimaticData.AddToLinkedControls(ucrReceiverMonth, {rdoDay}, bNewLinkedHideIfParameterMissing:=True)
```
dannyparsons: The year, station and element receivers should also be linked to the panel in a similar way. They are visible when `rdoMonth` or `rdoDay` are checked.

instat/dlgTidyClimatic.vb
```diff
@@ -0,0 +1,257 @@
...
+    End Sub
+
+    Private Sub SetDefaults()
+        clsTidyClimaticFunction = New RFunction
+
+        ucrSelectorTidyClimaticData.Reset()
+        ucrReceiverMultipleStack.SetMeAsReceiver()
+        ucrNewDFName.Reset()
+
+        clsTidyClimaticFunction.SetAssignTo(ucrSelectorTidyClimaticData.ucrAvailableDataFrames.cboAvailableDataFrames.Text & "_tidy", strTempDataframe:=ucrSelectorTidyClimaticData.ucrAvailableDataFrames.cboAvailableDataFrames.Text & "_tidy")
```
dannyparsons: If we're using the prefix then here you should set the prefix, otherwise this is doing something different.

instat/dlgTidyClimatic.vb
```diff
@@ -0,0 +1,257 @@
...
+        If rdoYear.Checked Then
+            ucrReceiverYear.Visible = False
+            ucrReceiverElement.Visible = False
+            ucrReceiverStation.Visible = False
+            lblOr.Visible = False
+            ucrReceiverMultipleStack.SetMeAsReceiver()
+        Else
+            ucrReceiverYear.Visible = True
+            ucrReceiverElement.Visible = True
+            ucrReceiverStation.Visible = True
```
dannyparsons: If these receivers are linked to the panel then all the visibility will be done automatically (expect the Or label probably).

instat/dlgTidyClimatic.vb
dannyparsons: This is still to do.

dannyparsons:
> everything on left side should be in line, with x location 10

This is still not looking good e.g. base buttons, save control, groupbox. None of them are in line, they just need x location 10.

instat/dlgTidyClimatic.vb
Muthenya: I had done this but they were only  visible for the rdoMonth since they are linked to it first. That's why I did them in the **ShowReceiver()**

instat/dlgTidyClimatic.vb
Muthenya: I have explained this above.

instat/dlgTidyClimatic.vb
dannyparsons: Because the link needs to be with both `rdoMonth` and `rdoDay` not just `rdoMonth`.

instat/dlgTidyClimatic.vb
Muthenya: Okay. I guess I'll do this for **Or** alone, or is there a better way to do it.

dannyparsons: We can't link the Or label as easily to a core control so this can still be done in this way.

instat/dlgTidyClimatic.vb
Muthenya: We already have a set prefix. Is there another reason why I should set it here? Or maybe I haven't understood your point.

dannyparsons: Here you're setting the assign to, so the new data frame will be called `my.data_tidy`, for example. But then when the prefix is set, the assign to will now become `my.data_tidy1`. So my point is they are doing different things. You either have a prefix or a fixed name. Here we want a prefix. It might be better to replace this line with a call to the method that sets the prefix.

Muthenya: Okay. I understand now.

instat/dlgTidyClimatic.vb
```diff
@@ -0,0 +1,253 @@
...
+            End If
+        End If
+        If rdoMonth.Checked Then
+            If Not ucrReceiverYear.IsEmpty AndAlso Not ucrReceiverMultipleStack.IsEmpty AndAlso Not ucrReceiverDayofMonth.IsEmpty AndAlso ucrNewDFName.IsComplete Then
+                ucrBase.OKEnabled(True)
+            Else
+                ucrBase.OKEnabled(False)
+            End If
+        End If
+        If rdoDay.Checked Then
```
dannyparsons: Just noticed that these should be `Else If`s

dannyparsons:
Last comment (I think), when I'm on Month receiver in Day format, or Day receiver in Month format, when I switch formats I don't have a current receiver, because they're hidden and the current receiver isn't changed. In this case I think go to the stack column receiver.

Muthenya:
I guess this means that whenever the panel changes, the focus then goes to the stack column receiver.

dannyparsons:
Only when those receivers which aren't visible any more are the current receiver..

instat/dlgTidyClimatic.vb
```diff
@@ -238,6 +236,14 @@ Public Class dlgTidyClimaticData
...
     End Sub
 
+    Private Sub ChangeCurrentReceiver()
+        Dim CurrReceiver = ucrSelectorTidyClimaticData.CurrentReceiver
+
+        If (CurrReceiver Is ucrReceiverDayofMonth AndAlso Not rdoMonth.Checked) OrElse (CurrReceiver Is ucrReceiverMonth AndAlso Not rdoDay.Checked) OrElse (rdoYear.Checked) Then
+            ucrReceiverMultipleStack.SetMeAsReceiver()
+        End If
+    End Sub
+
```
Muthenya: Danny I'm not certain if this was the best way to do this but it works.  Let me know what you think about.

dannyparsons: Yes this covers all the cases

dannyparsons:
Can you add it to the menu

PR #4815, opened 2018-07-26
Correlations

### LINKED ISSUE #3942: Using GGally::ggpairs more in Describe menu?
I'm looking into the correlations dialog as it's going to be useful in the procurement menu to compare red flag columns. Some of the red flag columns are binary i.e. just 0 or 1. In R this could be represented as a numeric column of 1s and 0s or as logical type column of `TRUE`s and `FALSE`s.

I first noticed that the dialog only allows numeric columns, so logical type columns couldn't be used. But testing all the functions used in the dialog, logical columns can be treated as numeric (sort of). This links to #3925 on how our selectors could exclude types instead of including them, so that logical is considered "numeric".

The `GGally::ggpairs` function does give slightly different outputs for logical columns compared to numeric columns.
This is with `a` and `b` as numeric 0-1 columns:
image
And this is with `a` and `b` as logical columns:
image

So slightly strange graphs on the diagonal for logical columns but otherwise the same.

I also tried `GGally::ggpairs` with factor columns just to see and it gives an interesting output:
image

So two questions:

1. Should we allow logical columns in the correlations dialog? I feel we probably should because I think we want to encourage binary columns to be stored as logical types instead of 0/1 numeric columns but if we don't allow logical here then users would have to convert. And there are probably other numeric-like columns which could work but are currently excluded.

2. Should we be making more use of `GGally::ggpairs` since it seems more general than correlations? The graphs part are sort of what we have under the two variable graph dialog. Could we include this in that dialog, or have a separate item under Describe > Two Variables since it's a table/graph mix?

Muthenya:
Partly fixes #3942.

### DIFF
--- a/instat/dlgCorrelation.vb
+++ b/instat/dlgCorrelation.vb
@@ -140,7 +140,6 @@ Public Class dlgCorrelation
         clsWrapFunction.SetRCommand("wrap")
         clsWrapFunction.AddParameter("points", Chr(39) & "points" & Chr(39), bIncludeArgumentName:=False, iPosition:=0)
         clsListFunction.AddParameter("continuous", clsRFunctionParameter:=clsWrapFunction)
-        clsRGraphicsFuction.AddParameter("upper", clsRFunctionParameter:=clsListFunction, iPosition:=2)
 
         clsRGGscatMatrixFunction.SetPackageName("GGally")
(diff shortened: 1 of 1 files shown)


dannyparsons:
The way you have written your comment means that the issue is automatically closed by merging this. If it only address part of an issue then do not include "fixes #xxx" in the comment.

PR #4816, opened 2018-07-26
Adding labeller parameter on facet function

### LINKED ISSUE #4809: Changing facet labels
The labeller functions in ggplot2 are useful for changing the display of facet labels
https://ggplot2.tidyverse.org/reference/labellers.html
https://ggplot2.tidyverse.org/reference/labeller.html

There's quite a lot here, including providing a list of values to manually override the labels and using mathematical expressions.

What we could do quickly and is useful is to have the option of adding the following to the facet function:

`labeller = label_value` only display level name (the default)
`labeller = label_both` display variable name + level name
`labeller = label_context` display variable name + level name when there is more than one facet factor on the same axis.

The third is not really useful for us at the moment as we don't have two factors on the same axis in our options. But we could have a checkbox on the facets tab which says "include variable name in labels" and that would be very quick.

maxwellfundi:
Fixes #4809 

### DIFF
--- a/instat/sdgPlots.vb
+++ b/instat/sdgPlots.vb
@@ -89,7 +89,7 @@ Public Class sdgPlots
         ucrChkIncludeFacets.AddParameterPresentCondition(False, "facets", False)
 
         ucrChkMargin.SetText("Margins")
-        ucrChkMargin.SetParameter(New RParameter("margins"), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=False)
+        ucrChkMargin.SetParameter(New RParameter("margins", 2), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=False)
         ucrChkMargin.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
         ucrChkMargin.SetRDefault("FALSE")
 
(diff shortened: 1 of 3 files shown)


instat/sdgPlots.vb
```diff
@@ -123,6 +123,14 @@ Public Class sdgPlots
         ucrChkFreeSpace.SetRDefault(Chr(34) & "fixed" & Chr(34))
         ucrChkFreeSpace.AddToLinkedControls(ucrChkNoOfRowsOrColumns, {False}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
 
+        ucrChkLabeler.SetText("Include Variable Name in Labels")
+        ucrChkLabeler.SetParameter(New RParameter("labeller"))
```
shadrackkibet: Do we need position here?

maxwellfundi: @shadrackkibet  We currently dont have positions for the parameters going on to the facets function. I will now put positions on all of these.

Muthenya:
This looks good from my end.

maxwellfundi:
Thank you @Muthenya

Issue #4400, opened 2018-02-09
Calculator problem with a particular variable name


rdstern:
I have had a problem with the use of the calculator and a particular variable name I have been using.

I had a variable called Tmin.  I changed some values in that Variable and put the result in a new variable called Tmin1.

I then calculated another variable when Tmin1<0 and it worked fine.  But Tmin1<-5 gave the wrong answer.  It gave FALSE every time, even when Tmin1 was NA.  I tried brackets etc, but nothing seemed to help.

When I renamed Tmin1 to Tmina, then Tmina<-5 works fine.  

I wonder what I, or R-Instat, or R are doing wrong?

dannyparsons:
It should work with `Tmin1<(-5)` or `Tmin1< -5`. WIthout the brackets you have `<-` which is assignment so that messes up the inequality you want. I think that's why it's always suggested to put spaces between symbols.

shadrackkibet:
I think to fix this problem we need to add a space on both sides automatically on this symbols  (<,>, <=, >=,==).

rdstern:
I agree

shadrackkibet:
This can now be closed.

PR #4833, opened 2018-07-30
Calculator 

### LINKED ISSUE #4828: More small items on the calculator
1. We now have NA on the logical keyboard, but not TRUE or FALSE.  I don't think they are anywhere else.  Can they be added?
2. This may not be so easy?  In the maths keyboard, when you tick to show arguments on round or signif keys the function also has a digits argument, but this isn't shown.  If easy, then it would be good to show.  Otherwise (perhaps anyway) it is good for a tooltip, perhaps: 
"round(x) to round to whole numbers; round(x,2) to round to 2 decimal places, round(x,-2) to round to the nearest 100".
"signif(x,3) to round to 3 significant figures".

### LINKED ISSUE #4831: Small edit on the call to the View command.

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes  #4828  and  #4400


shadrackkibet:
yes, i have corrected that. Thanks.

Muthenya:
You missed one symbol, i.e "==". It  doesn't have space. Otherwise, this looks good from my end.

shadrackkibet:
That is  now done. @dannyparsons this is ready for you.

instat/ucrCalculator.vb
```diff
@@ -64,6 +64,8 @@ Public Class ucrCalculator
         cmdIsTrue.Enabled = False
         cmdIsFalse.Enabled = False
         bControlsInitialised = True
+        ttCalculator.SetToolTip(cmdRound, "round(x) to round to whole numbers, round(x,2) to round to 2 decimal places, round(x,-2) to round to the nearest 100")
```
dannyparsons: If we're going to have this then we should probably have tooltips on almost every button. It's going to look strange if this is the only button with a tooltip now.

instat/ucrCalculator.vb
```diff
@@ -1215,10 +1217,18 @@ Public Class ucrCalculator
...
-        ucrReceiverForCalculation.AddToReceiverAtCursorPosition("%in%")
+        ucrReceiverForCalculation.AddToReceiverAtCursorPosition(" %in% ")
     End Sub
 
     Private Sub cmdDoubleSqrBrackets_Click(sender As Object, e As EventArgs) Handles cmdDoubleSqrBrackets.Click
         ucrReceiverForCalculation.AddToReceiverAtCursorPosition("[[]]", 2)
     End Sub
+
+    Private Sub cmdTrue_Click(sender As Object, e As EventArgs) Handles cmdTrue.Click
+        ucrReceiverForCalculation.AddToReceiverAtCursorPosition("TRUE")
```
dannyparsons: Do you know why this function has the string "TRUE" instead of the boolean True?

shadrackkibet: My understanding is that we need this button to produce TRUE when clicked and not True.

dannyparsons: Sorry I thought this was a different function in the receiver. This is correct.

instat/ucrCalculator.vb
dannyparsons: If we're going to have more tootips then I suggest that the ttoltip has the same name as the button.

shadrackkibet: I think we just use one tooltip then  set the button and the text separately instead of having separate tooltips. Like for this case i have one tooltip control and i set it for two different buttons and it works.

dannyparsons: Ok that's good. I suggest one tooltip per keyboard then to keep this well organised.

PR #4837, opened 2018-07-30
View Dataframe on Right click

### LINKED ISSUE #4831: Small edit on the call to the View command.

shadrackkibet:
@africanmathsinitiative/developers  this is ready for review. Fixes #4831


instat/ucrDataView.vb
```diff
@@ -38,6 +38,7 @@ Public Class ucrDataView
     Private clsFreezeColumns As New RFunction
     Private clsUnfreezeColumns As New RFunction
     Private clsViewDataFrame As New RFunction
+    Private clsAssignOperator As New ROperator
```
dannyparsons: You don't need an operator for this. You can set the assignment of the function.

instat/ucrDataView.vb
```diff
@@ -681,10 +683,12 @@ Public Class ucrDataView
...
     End Sub
 
-    Private Sub ViewSheet_Click(sender As Object, e As EventArgs) Handles ViewSheet.Click
-        clsViewDataFrame.AddParameter("x", clsRFunctionParameter:=clsGetDataFrame)
-        clsViewDataFrame.AddParameter("title", Chr(34) & grdCurrSheet.Name & Chr(34))
-        RunScriptFromDataView(clsViewDataFrame.ToScript, strComment:="Right Click Menu: View R Data Frame", bSeparateThread:=False)
+    Private Sub ViewSheet_Click(sender As Object, e As EventArgs) Handles ViewSheet.Click
+        clsAssignOperator.AddParameter("left", grdCurrSheet.Name, iPosition:=0)
+        clsAssignOperator.AddParameter("right", clsRFunctionParameter:=clsGetDataFrame, iPosition:=1)
+        clsViewDataFrame.AddParameter("x", grdCurrSheet.Name)
```
dannyparsons: Doing this is dangerous because if your assign name changes then this code no longer works. This is one of the advantages of using the assignment, the R syntax does this for you if the function has been assigned so you don't have to make sure these are the same.

Patowhiz:
@shadrackkibet have you resolved the issues above? I would like to help in testing this

shadrackkibet:
@africanmathsinitiative/developers can someone volunteer to take over this?.

Patowhiz:
I'm taking over

stevenndungu:
@Patowhiz what is the state of this?

Patowhiz:
Still working on this

dannyparsons:
Actually this is working (almost) correctly from the dialog already. In the dialog, the `title` parameter just needs to be removed, the assigning is being done correctly automatically.
From the right click in the grid this needs to be corrected. It is in `ucrDataView` under the handling of the `ViewSheet.Click` event. The `title` parameter should be removed and the function should be assigned to the name of the data frame to get the correct code.

dannyparsons:
@Ivanluv You could complete this task.

PR #4840, opened 2018-07-31
Changing R-Instat Default Font

lilyclements:
@africanmathsinitiative/developers I think we should change the default font in R-Instat to Calibri (size 13) because it looks nicer and more updated than the current font. It is also the default font on word so is easy-to-read.

I think to see the changes on your laptop you would have to do a factory reset on the Options after adding this packages but a screen shot of differences are below.

It does not affect the font in the grid, nor does it affect font from output from dialogs. It also doesn't affect the script or log window.

image

Let me know what you think and/or if you think it should be merged.


maxwellfundi:
@africanmathsinitiative/developers  I think i like this new font more than the existing one. Thanks @lilyclements

dannyparsons:
I think it's an improvement too. How does it look with the output? This is still size 11, should it be increased too?

maxwellfundi:
image
@dannyparsons  I think it still looks good with the output. The size here is 13 also.

lilyclements:
Did you change the output font manually Maxwell? That shouldn’t be a change in this pull request so I’m just checking. If you didn’t change it yourself, I should fix that!

dannyparsons:
@maxwellfundi That doesn't look good at all! Compare that to the current font, the columns don't line up at all.

The font of the output shouldn't be changed but the size could, which is what I was asking.

maxwellfundi:
I agree the columns really are misplaced but we get a lot of that sometimes. We shouldnt have that then. 
@lilyclements  I did not change anything on your pull request so i was just playing with the fonts and sizes

shadrackkibet:
This change looks okay in my machine, However, I do not see much difference. 

**Current**
image

**New**
image

lloyddewit:
@rdstern I approved from the technical side. I would like to close this PR (this week if possible). We can either merge it before closing, or just close it. This depends on whether you prefer Calibri 13, or the current font. It also depends on whether the new font causes issues with column alignment. Please let me know your preference.

@lilyclements Do you have anything you want to add? 

Thanks!

lilyclements:
> This change looks okay in my machine, However, I do not see much difference. This can be merged.

@shadrackkibet I also am now no longer seeing a difference if I put these changes into a new branch (I cannot access the old branch - it just crashes when I open R-Instat on it). I am using a different computer now, however, to two years ago so perhaps it is to do with that.
@africanmathsinitiative/developers can anyone else see changes if they try this now?

shadrackkibet:
@lilyclements I would suggest you update this branch with the master.

lloyddewit:
I will test this change to see if the new font makes any difference on my machine

lloyddewit:
I tested on my laptop.
Before change:
image
After change:
image

I cannot see a difference.

Therefore I'll close this PR, rather than merging it.

PR #4842, opened 2018-08-01
Make date problem with day 60 for Non-leap years.

shadrackkibet:

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -2027,6 +2027,8 @@ data_object$set("public","make_date_yeardoy", function(year, doy, year_format = 
     if(is.factor(year_col)) {
       year_col <- as.numeric(levels(year_col))[year_col]
     }
+    #Replacing day 60 with 0 for non-leap years.This will result into NA dates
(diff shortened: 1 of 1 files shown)


instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -2027,6 +2027,8 @@ data_object$set("public","make_date_yeardoy", function(year, doy, year_format =
     if(is.factor(year_col)) {
       year_col <- as.numeric(levels(year_col))[year_col]
     }
+    #Replacing day 60 with 0 for non-leap years.This will result into NA dates
+    doy_col[(!lubridate::leap_year(year_col)) & doy_col == 60] <- 0
```
dannyparsons: Great idea. Could this just be made NA instead of 0. Both works, but in case anything further is done to `doy_col` in future, it's clear where NA comes from.

shadrackkibet: I tried NA but this caused problems with creating dates which i didn't understand.

image

dannyparsons: That's because the next line as condition `doy_col > 60` which returns `NA`. Let's not complicate and leave it like this then.

PR #4852, opened 2018-08-07
Weighted summaries

bethanclarke:
Included R-code for weights option for the following summary functions: mean, sd, var, median, quantile, skewness, kurtosis, covariance and coefficient of variation


shadrackkibet:
@dannyparsons  @rdstern  do you know what remains on this?

rdstern:
No I don't except that Beth was working on this with @volloholic .  So really good if you could look into this and discuss with David.  Both kept saying it is easy.  It would be an excellent addition to have to the summary system.

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -551,10 +568,21 @@ upper_quartile <- function(x, na.rm = FALSE, na_type = "", ...) {
...
-    return(e1071::skewness(x, na.rm = na.rm, type = type))
+    if (missing(weights)) {
+      return(e1071::skewness(x, na.rm = na.rm, type = type))
+    }
+    if (length(weights) != length(x)) 
+      stop("'x' and 'weights' must have the same length")
+    w <- as.double(weights)
+    if (na.rm) {
+      i <- !is.na(x)
+      weights <- weights[i]
```
shadrackkibet: what if the weights have missing values?

dannyparsons:
I would guess missing values in weights are handled the same as in the data. If `na.rm = TRUE` then missing values in data and values where the corresponding weight is missing are excluded. If `na.rm = FALSE` then if the data, or the weights, have any missing values then the result is missing.

Good to discuss with David while he is there though. Some of these checks on the weights that are needed in every summary should be done in a single place.

rdstern:
I am closing this, because it is also within @shadrackkibet current work

PR #4853, opened 2018-08-07
Added fixed value options in Make Date dialog

### LINKED ISSUE #4233: Date columns with non daily data
It's useful to have a Date type column even with monthly or yearly data e.g. to plot time series. At the moment our Make Date dialog requires year, month and day columns. 
But monthly data often only have Year and Month columns and I can't do time series line/point plots without a continuous variable like a Date. So I want to create a Date column and just take a fixed day for each month e.g. 1. I can do this in two ways:
1. Create a Day column in the calculator of all 1s and then use that in the Make Date dialog
2. Use the calculator to create: `as.Date(paste(Year, Month, 1), format = "%Y %m %d))`

But neither of these I like explaining to users, especially beginners, as it's not obvious or very logical.

Our Make Date dialog should be able to do this. I suggest we allow Year, Month and Day to also be fixed numbers as well as columns. Then I could flexibly make a Date with only some of these columns.

dannyparsons:
fixes #4233

- Implemented option (in R code and dialog) to have year, month or day as a fixed value in Year Month Day format, as described in #4233 
- Tidied and simplified dialog code and small bug fixes

Because of the general tidying of the code, please test that the general functionality of the dialog is working as before, as well as the new features.

### DIFF
--- a/instat/dlgMakeDate.Designer.vb
+++ b/instat/dlgMakeDate.Designer.vb
@@ -41,24 +41,43 @@ Partial Class dlgMakeDate
         Me.components = New System.ComponentModel.Container()
         Dim resources As System.ComponentModel.ComponentResourceManager = New System.ComponentModel.ComponentResourceManager(GetType(dlgMakeDate))
         Me.grpTwoColumns = New System.Windows.Forms.GroupBox()
+        Me.ucrChkTwoDigitYear = New instat.ucrCheck()
+        Me.ucrNudCutoff = New instat.ucrNud()
         Me.lblCutOffTwo = New System.Windows.Forms.Label()
+        Me.ucrInputComboBoxMonthTwo = New instat.ucrInputComboBox()
         Me.lblYearTwo = New System.Windows.Forms.Label()
+        Me.ucrReceiverYearTwo = New instat.ucrReceiverSingle()
         Me.lblDayofYear = New System.Windows.Forms.Label()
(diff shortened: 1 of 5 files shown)


shadrackkibet:
Testing this now.

shadrackkibet:
Two small issues i have observed.

- When i am on **Year-Month-Day** radio button the  **format** comboboxes reset only when i click reset twice.
- Still on the third option **Year-Month-Day**, when  i click on the value radio button(month) the combobox changes to numeric and is disabled which is okay because the the input only allows you to input numeric values, i wonder if we might allow typing of characters and allow other options in future?. Another issue here is that when i reset the combobox is still disabled and my option has changed to column radio button as shown below.
image

Otherwise the dialog works properly for both previous and added functionalities.

dannyparsons:
Thanks for comments.
1. Fixed the resetting issue of the format controls
2. No I wasn't planning on allowing other formats in future, in fact it was more work to restrict it like this. The output, the date column, will be no different with a numeric month or a character input and numeric has obvious advantages over typing the name of a month. The format is deliberately fixed so that it's clear to the user what to enter. I can't see any advantage to allowing a month to be specified in any other way?
3. Fixed the disabled reset issue.
4. I've added min and max validation for day value (1 to 31) and month value (1 to 12). We don't have integer validation (we probably should) but this is clear enough already I think.

Please test again.

shadrackkibet:
My recent testing shows that above issues have been addressed. This is ready for merging.

PR #4857, opened 2018-08-08
Qc Rainfall

### LINKED ISSUE #4854: Rainfall quality control with the Same option
It doesn't work!  This is the option in Climatic > Prepare > Rainfall qc that checked whether x or more (default 2) non-zero rainfall values are the same.  

This seems to work fine for temperatures qc, but not for rainfall.  It doesn't give any values and David said the code is still wrong.  (As I guess it was initially.)  It may be easy to fix, given the temperatures qc works, but otherwise he can help.

I tried with Dodoma as usual.  It doesn't give an error, just no true values.  Then we mad 2 values the same in the file at the top - and still nothing!

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes #4854


shadrackkibet:
Can someone test this please?

shadrackkibet:
@bethanclarke  could you please test this?

shadrackkibet:

PR #4868, opened 2018-08-13
Clearing Output Window

### LINKED ISSUE #4792: Additional menu item for clearing the output window
@africanmathsinitiative/developers  There is very well designed icon for clearing the output window in the toolbar, still, most of the time I find myself right clicking on the output window to access such functionality. Most apps have a similar functionalities(clearing contents of input controls) built right into the control on the right click menu. We already have a similar one on the script window.  Wouldn't be nicer to also have a menu item for this on right clicking the output window?

output

Patowhiz:
Fixes #4792


Muthenya:
@Patowhiz when I was testing the branch, I realized that when the output window is cleared only the **Delete selection** option is disabled and the **clear** option is enabled, yet the output window is "empty". Is there a reason for this, since I thought that both the options should be disabled?

dannyparsons:
I think that's not a problem that Clear is still enabled when it's already blank.

Patowhiz:
Just like @dannyparsons  suggests, I didn't see the need to disable the clear when richtextbox is empty. The user will just be clearing nothing. I'm disabling the Delete Selected, to show that it only applies to selected text.

@Muthenya did you find any other issue apart from the above. If not this can be merged since it has changes on `frmMain`. I

Muthenya:
Okay. No other issue from me.

Patowhiz:

PR #4873, opened 2018-08-15
Themes sub dialog

bethanclarke:
This pull request is for a new simpler sub themes dialog. It is very similar to the Themes sub dialog, but it only has options for the x and y axis. 

@maxwellfundi please could take over this? The code structure is the same as it is in sdgThemes, so it should be very simple to finish.

One issue with themes in general is with ucrColour, which is used on ucrElementText, ucrElementRect, ucrElementLine. Because of this issue, we are unable to view the dialogs since they contain these ucr's.


dannyparsons:
@maxwellfundi this is the work that needs to be completed on themes. Can you take this over to complete?

maxwellfundi:
Yes, pulling this now?

PR #4875, opened 2018-08-15
Compare Dialog

Muthenya:


Muthenya:
Here is the current designer for the dialog
image

shadrackkibet:
@dannyparsons we have managed to have difference option working now. Please test it. @rdstern any comments on the design so far?

dannyparsons:
Looks good, I will test this evening to have feedback for you tomorrow.

The second label could be "Absolute Difference"

dannyparsons:
Looks good and works well, great! Few small comments:

- No current receiver on first opening
- Restrict both receivers to numeric columns
- Title should be more than "Compare" maybe "Compare Satellite Data"

shadrackkibet:
@dannyparsons Anomalies option is also done. Please test it as well.

dannyparsons:
This is looking great, the code that runs is completely correct so that's great. Lots of small dialog issues below but nothing major.

**Differences**
- align the bias and abs dev checkboxes
- align the bias and abs dev textboxes
- code would look tidier if the combination calculation was assigned to something
- bias and absdev column names don't auto change numbers (set prefix as "bias" and "absdev" instead of setting name, and link to selector and set item type as columns)
- can you make both bias and abs dev checked by default
- tab order
- Selector label should indicate its numeric variables

**Anomalies**
- Rename the checkbox "Moving Average" and have it unchecked (so it's less distracting)
- Align checkboxes & textboxes
- both column names don't auto change numbers after running. If easy, set prefix as "col.a_anom" and "col.b_anom" where "col.a" and "col.b" are the names of the satellite and station columns chosen. Otherwise "satellite_anom" and "station_anom" would also be fine prefixes. And link to selector and set item type as columns.
- Station isn't required (if it's only one station). It can be set to autofill.
- code would look tidier if the combination calculation was assigned to something
- if current receiver is station or within year, when I change to "Differences" there no visible current receiver (can go back to satellite element)
- if bias is unchecked when I reopen the dialog it switched to "Anomalies" (because of how you've set up conditions). I suggest a better condition is to check the presence of the manipulations parameter in the combination calculation
- When I reopened, if the differences tab was selected sometimes the "Moving" controls were visible (possibly same conditions issue as previous?)

dannyparsons:
If you again send up whatever you get to by this evening even if you don't finish the list, I'll test again tonight.

shadrackkibet:
@dannyparsons  this is what we have now. Please test it again.

instat/dlgCompare.vb
```diff
@@ -1,3 +1,349 @@
...
+        ucrInputStationAnomalies.SetDataFrameSelector(ucrSelectorCompare.ucrAvailableDataFrames)
+        ucrInputStationAnomalies.SetDefaultTypeAsColumn()
+        ucrInputStationAnomalies.SetPrefix("station_anom")
+    End Sub
+
+    Private Sub SetDefaults()
+        'clsBiasCalculation.Clear()
+        clsDiffOperator.Clear()
+        clsListFunction.Clear()
+        'clsAbsDevCalculation.Clear()
```
dannyparsons: Why are these two commented out?

dannyparsons:
Thanks for doing this quickly, I've merged with a few minor changes.

shadrackkibet:
great.

PR #4910, opened 2018-08-29
Tidy data

Muthenya:
@dannyparsons I have made the changes. This is just waiting for the edits to the RCode and possibly other changes to the dialog after that.


instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1871,7 +1871,16 @@ instat_object$set("public","tidy_climatic_data", function(x, format, stack_cols,
...
   else if(format == "years") {
     
-    if(nrow(x) != 366) stop("data must have exactly 366 rows when format = 'years'")
+    by_cols <- c()
+    if(!missing(station)) by_cols <- c(by_cols, station)
+    if(!missing(element)) by_cols <- c(by_cols, element)
+    
+    if(length(by_cols) > 0) {
+      group_lengths <- x %>% dplyr::group_by(!!! rlang::syms(by_cols)) %>% dplyr::summarise(n = n())
+      print(group_lengths)
```
dannyparsons: @Muthenya please remove this print,I left it in by mistake

Muthenya: Done.

Muthenya:
@dannyparsons this is ready now. I don't think we currently have a data frame to test the new changes.

dannyparsons:
If you read what Roger wrote originally when he reopened the issue you will be able to create your own example for this, at least for multiple stations. Get in touch if it's not clear.

rdstern:
Here are 2 datasets to assist in the testing
[BVUMBWE1.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/2368475/BVUMBWE1.xlsx)
and
[CHITEDZE1.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/2368477/CHITEDZE1.xlsx)

You could start with Bvumbwe and **append** the Tmax and Tmin (they have the same years).  That would start the testing of multiple elements, for one station.

Then adding the appended Tmax and Tmin also for Chitedze would have 2 stations and 2 elements.  

Then try for Bvumbwe with all the elements - rainfall has more years and I think sunshine has fewer.  Does this work also?

Then anything more that you like.

Good luck.  These examples are "for real", because we may well have to do this in Malawi for lots of stations before the end of 2018.

Muthenya:
Great. Thanks.

Muthenya:
I have managed to test the dialog and it works well for the following instances:
1. When we have one station and two elements.
2. When we have two stations and two elements.
3. When we have two stations and one element.
4. When we have one station with different elements with different years. (It just gives NAs where there is no observation for a particular year for an element, e.g, for Bvumbwe using MinTemp, Rainfall and Sunshine, it gives NAs for MinTemp and Sunshine from 1962 to 1970 since there are no records for the two only for Rainfall).

5. And it still works for one station with one element.

Muthenya:
I was thinking about the need of having a check or checkbox for when there is multiple stations or multiple elements for Ok to be enabled. Currently, Ok is enabled when the Year multiple receiver is filled but following the changes, the elements' and/or stations' receivers require to be filled otherwise an error message will be thrown. I don't know if it is that obvious that when dealing with multiple elements and/or stations that their respective receivers need to be filled.

rdstern:
Well done on the testing.  There is one more thing you could do usefully when testing.  You were always sensible.  It is also useful to make deliberate mistakes and then see what happens.  
[PluviometrieRegionKenscoffDrSternCorrDGc_0917.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/2397963/PluviometrieRegionKenscoffDrSternCorrDGc_0917.xlsx)

Here is an example data set.  It imports easily and is the 3rd option, with a column for each day of the month.  

I am not using your most recent version.  On mine I tried with **the first option** - it is the year one  which is clearly wrong. But people may do mistakes like this:  They may forget to use the option buttons at the top and year is the default.

Put the 31 columns x1-x31 into the receiver.  Then press OK.  Then R-Instat first says: 

**Error running R command(s)
Error in data_book$tidy_climatic_data(x = PluviometrieRegionKenscoffDrSternCorrDGc_0917,  : 
  data must have exactly 366 rows when format = 'years'

.temp_val <- capture.output(data_book$tidy_climatic_data(x=PluviometrieRegionKenscoffDrSternCorrDGc_0917, format="years", stack_cols=c("X1","X2","X3","X4","X5","X6","X7","X8","X9","X10","X11","X12","X13","X14","X15","X16","X17","X18","X19","X20","X21","X22","X23","X24","X25","X26","X27","X28","X29","X30","X31"), element_name="pluie", new_name="PluviometrieRegionKenscoffDrSternCorrDGc_0917_tidy1"))
OK**   

That is sort of OK as an error message.  But when you click OK on that error message, then it says:


**Error retrieving R variable
Error: object '.temp_val' not found

The error occurred in attempting to retrieve:.temp_val
OK**   

That's not a helpful message.  What happens in your version.

dannyparsons:
> That is sort of OK as an error message. But when you click OK on that error message, then it says:
> 
> **Error retrieving R variable
> Error: object '.temp_val' not found
> 
> The error occurred in attempting to retrieve:.temp_val
> OK**

This message is coming from how we manage the link to R so is more general. We should be able to have something better instead of this error message but it's nothing to do with this dialog.

dannyparsons:
> I was thinking about the need of having a check or checkbox for when there is multiple stations or multiple elements for Ok to be enabled. Currently, Ok is enabled when the Year multiple receiver is filled but following the changes, the elements' and/or stations' receivers require to be filled otherwise an error message will be thrown. I don't know if it is that obvious that when dealing with multiple elements and/or stations that their respective receivers need to be filled.

This is a good point. I can see it would be easy to miss out filling the element or station receiver. I'm not sure what extra use a checkbox would be though? If you forget to fill the receiver, what if you forget to check the checkbox? And what if you do fill the receiver but didn't check the checkbox, is that correct or not? I can see what you're waiting to do with this but feel it probably just adds complications.

One way to do this would be to try to make the software "clever". We could try to guess when there are multiple elements/stations and fill in the receivers if they have recognised standard names. I'm generally cautious about doing this. It might be useful in some cases but when it goes wrong it can be even more confusing. So far our approach has been not to automate too many things, unless its very clear why. It's easy to explain to a user why they made a mistake and forgot something, but if the software did something wrong its much harder for a user to understand and put it right.
Another option is to understand common errors and have specific helpful error messages in these cases when we suspect this is happened. For example, the error message

> data must have exactly 366 rows when format = 'years'

could happen when you have multiple stations. The message could be adapted to suggest they might have multiple elements/stations that wasn't specified. But being careful not to say that it **is** the problem, because it might be something else (as Roger showed above). The [tidyverse style guide on error messages](http://style.tidyverse.org/error-messages.html) is quite useful to explain what they think is good practice for error messages.

There's other things we do to try prevent this mistake. Just the dialog design can play a big part in this. A clear and logical design helps the user to fill it in correctly. 
I don't have immediate thoughts on what we could do about this, other than look at doing better error trapping, but open to thoughts and suggestions.

Muthenya:
@dannyparsons any suggestions for the way forward on this?

dannyparsons:
I put my initial thoughts above to see if others had anything to comment on. We could do more to improve this but it would be good to get this improved version in for this release. I would suggest that someone tests this now so it can be merged.

Muthenya:

dannyparsons:
This looks good, I'll merge for further testing by @rdstern

PR #4912, opened 2018-08-29
Evapotranspiration

### LINKED ISSUE #4500: Adding an evapotranspiration dialogue
This is a new dialogue.  At least for now it could go at the bottom of Climatic > Prepare with a line under End of the rains and then the dialogue saying Evapotranspiration.

It could have two tabs with the first saying **Penman-Monteith** and the second saying **Hargreaves**.  This is consistent with the FAO recommendations to use P/M all the time, but with Hargreaves possible if there is only temperature data.
The definitive guide is the 1998 FAO irrigation and drainage paper number 56 (I think) [here](http://academic.uprm.edu/abe/backup2/tomas/fao%2056.pdf) .  It recommends P/M with the suggestion for Hargreaves on page 64.
For now I suggest these 2 be implemented as they are in the package called Evapotranspiration.  The simpler could be the function called ET.Hargreaves.Samani, which just needs Tmin and Tmax.  We then save the (daily) date in a further column, possibly called ETHar.
The P/M method needs more data.  I suggest we first implement the function from the package assuming the relevant elements exist, i.e. using example data from the manual.

Later there can be formulae to estimate those variables that do not exist in the data.  I do not think these are in the Evapotranspiration package.  They are each simple calculations, given in the FAO guide.  

We would need some discussion as to whether those are produced through the same dialogue, or as part of dialogues for the respective elements.  That can come later.

This work is for Beth and has been discussed with her.  I still don't know how to include her from Github?

shadrackkibet:
In progress Fixes #4500

### DIFF
--- a/instat/dlgEvapotranspiration.Designer.vb
+++ b/instat/dlgEvapotranspiration.Designer.vb
@@ -35,9 +35,6 @@ Partial Class dlgEvapotranspiration
         Me.lblMaxMissingDays = New System.Windows.Forms.Label()
         Me.lblMaxDurationMissingData = New System.Windows.Forms.Label()
         Me.lblMaxPerctMissindData = New System.Windows.Forms.Label()
-        Me.grpSolar = New System.Windows.Forms.GroupBox()
-        Me.lblCloudCover = New System.Windows.Forms.Label()
-        Me.lblSunshineHrs = New System.Windows.Forms.Label()
         Me.lblSolar = New System.Windows.Forms.Label()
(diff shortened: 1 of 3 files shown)


shadrackkibet:
@africanmathsinitiative/developers this is ready for testing.

shadrackkibet:
@rdstern  here is the current design for this dialog.

image
image

shadrackkibet:
@africanmathsinitiative/developers Can someone test this please?

maxwellfundi:
@rdstern Please test this too. @shadrackkibet asked me to merge it for you.

PR #4922, opened 2018-09-03
Conversions

### LINKED ISSUE #4845: New Conversions dialogue in Climatic menu
This is not really a new topic, but a new formulation of (parts) of other issues.

In discussion with Danny it seems that the sunshine hours calculation and the conversion of temperatures, etc, should have its own climatic dialogue.

It still needs a bit more discussion on the details of the dialogue.  Here is my go.
a) It is called Conversions and goes in **Climatic > Prepare and just under Transform**.
b) It has 2 of our new-style buttons at the top.  The first is **Units** and the second is **Day Length**.
c) If Units, then:

1. Our ordinary selector.
2. A single receiver label perhaps "Element"
3. At the bottom the control for saving the result, with label Result into:  
4. In the middle there are 3 combo boxes in a row.  The labels could be on top.  The first is labelled **Element** and has the 3 possibilities of Rain, Temperature, Wind.  The second is labelled **From** and the third is labelled **To**.

If Rain, then From and To contain: Inches, mm, cm
If Temp then: Centigrade, Fahrenheit, Kelvin
If Wind then: • knots: Knots, mph: Miles per hour, mps: Meters per second, ftps: Feet per second, kmph: Kilometers per hour

5. There is also an up-down **Decimals**: 0 to 3 default is 1.
See the package for more details

They will run commands such as
convert_precip(x, old="m",new="i",round=2)
convert_temperature(x,old="c",new="k",round=1)
convert_wind_speed(x,old="kmph",new="kn",round=1)

 If Day Length, then:
1. Usual Selector
2. 2 single receivers on the right for **Station** and **Date**.
3. A receiver for Latitude.  This is a bit like the receivers in the Climatic > Prepare > End of Rains in the end of Season bit.  
a) It has 2 buttons, labelled Value and Variable.  
b) When Value is selected, then you type the Value, and after the Value field have a label Decimal.  
c) If Variable, then you enter a column.
d) In the equivalent Climatic > End of Rains there is also a "replace NA with.  We don't need that here.

In the future the latitude (when it is a Variable) can be either from the same data frame as the Station and Date columns, or it could be from a linked data frame.  That's like the End of Rains dialogue.  Unless there is time, for now, have the data (including latitude) in the same data frame.

Then Result into for the result.
This just runs the command geosphere::daylength(lat, doy), and doy can be a date column.

Muthenya:
Fixes #4845


Muthenya:

shadrackkibet:
Yes, this is ready for @dannyparsons to review.

instat/dlgConversions.vb
```diff
@@ -257,9 +257,11 @@ Public Class dlgConversions
         If rdoDayLength.Checked Then
             ucrBase.clsRsyntax.SetBaseRFunction(clsDayLengthFunction)
             ucrReceiverDate.SetMeAsReceiver()
+            grpLatitude.Show()
```
shadrackkibet: I suggest this to be done by `.setlinkeddisplaycontrol() `  to one of the control linked this radiobutton.

Muthenya: This is now done.

maxwellfundi:
@dannyparsons  You could have a look at this now. issues seem to have been resolved.

Muthenya:
@maxwellfundi there is still some few edits from the issue that are yet to be made and are awaiting Danny's input.

maxwellfundi:
Okay, just saw @shadrackkibet marked this as resolved. Thanks.

dannyparsons:
What input do you need from me here?

Muthenya:
This is what I'm referring to Danny. You mentioned this in the issue.

> Also for the Units option, the From and To combo boxes depend on the Element combo box. But this isn't shown in the current design, they just look like 3 independent options. I'll send an image of a website where the design makes it clear Element should be chosen first, which is what we want

dannyparsons:
Oh yes thanks, I had forgot to write back on this. This is what I was mentioning [from here](https://www.calculator.net/conversion-calculator.html).

Not saying we should copy the design, although I do quite like the tab buttons, which is "Element" for us. But you see how the choice of Element is separate and clearly comes before the From and To.

image

Muthenya:
I agree. It looks much better and has better order. How do you propose we implement this in the dialog?
I was thinking of replacing the Element combo box with a group box and three radio buttons for Rain, Temperature and Wind Speed that link to their respective From and To drop downs.

dannyparsons:
Yes I think that would look good.

Muthenya:
This is now ready. @dannyparsons please have a look at this.

dannyparsons:
Not quite sure why but there are conflicts here with sdgSummaries. Could you have a look at this with @maxwellfundi ?

maxwellfundi:
@dannyparsons This is happening because there were some commits that  were from @shadrackkibet which had this change but then they were reverted. I am going to fix this on his branch and it will be okay when he updates.

dannyparsons:
Thanks. So the conflicts are now fixed but there are changes to the summary sdg which I assume are not wanted?

maxwellfundi:
Yes, it gets very tricky when a revert was done. @Muthenya could you just close this and reopen a new one with just the changes for the conversions by copying and pasting your changes

PR #4930, opened 2018-09-06
Summaries subdialog

### LINKED ISSUE #4497: Summaries - first and last and nth
The summaries dialogue (and hence climatic summaries has changed a lot. It is looking very impressive now. 

We now have First and Last and nth. What remains is the add the **OrderBy** facility to these summaries. I suggest we keep First/Last/nth where they are now. 

In addition we repeat them in the 2-variable tab - with the OrderBy label for that block of summaries. So it is effectively another statistic - I hope that is OK?

Muthenya:
I have taken up the issue from @shadrackkibet. This fixes  #4497.


instat/sdgSummaries.vb
```diff
@@ -344,6 +347,15 @@ Public Class sdgSummaries
         End If
     End Sub
 
+    Private Sub OrderByCheckEnabled()
+        If ucrChkFirst.Checked OrElse ucrChkLast.Checked OrElse ucrChknth.Checked Then
+            ucrChkOrderBy.Enabled = True
+        Else
+            ucrChkOrderBy.Enabled = False
+        End If
```
shadrackkibet: let's do this automatically through linking of controls.

dannyparsons:
Seems to work well. Two things:

- When I check order by and then uncheck all the summaries, order by is disabled but still checked and the selector receiver still visible. I think it should be unchecked with controls hidden
- The `n` label and text box doesn't seem to line up with the `nth` label

instat/sdgSummaries.vb
Muthenya: @shadrackkibet how should I do this and still ensure that the Check Order by checkbox is disabled and unchecked when all the summaries are unchecked?

dannyparsons: I think that needs to be done manually.

shadrackkibet: okay

instat/sdgSummaries.resx
```diff
@@ -1663,7 +1663,7 @@
     <value>3</value>
   </data>
   <data name="ucrInputN.Location" type="System.Drawing.Point, System.Drawing">
-    <value>279, 17</value>
+    <value>280, 19</value>
```
maxwellfundi: @Muthenya  this is a minor one but you need to move this a little more 1px to be aligned properly.

maxwellfundi:
@dannyparsons  you can have a look at this now.

instat/sdgSummaries.resx
dannyparsons: I agree

dannyparsons:
Working well now, great.

These three controls still seem completley out of line though:

image

Muthenya:
I'm on it.

Muthenya:
I guess this is now okay.

image

Issue #4381, opened 2018-01-31
Executing commands from the script file


rdstern:
Having got this far with the script file I would like to go a bit further - of course - or at least understand better why we can't.

Suppose I would like to execute a new R command that is not yet available in R-Instat.  The t.test command is one example and the prop.test command is another and wilcok.test is a third.  They are all in the main stats package.  And we can't yet do this simple inference.

I know I can go into RStudio, but I don't want to do this for just one command - and there are many users who (I strongly suggest) would like to be able to do this sort of thing in R-Instat itself.  

1. The obvious was (to me) is to execute a similar command and then edit the script file and run.  (I accept the dangers!).  So I used the samrain data from the Instat library and did a Model > One Variable > Fit Model dialogue.  I opted to not have any graphs then got the numerical results fine, then put the code into the script file.  It is quite complicated code but easy to see what to change.  Nice that it gives the name of the package.  So I substituted:

last_model <- stats::wilcox.test(Strt1) 

for the call to the fitdistrplus package.  
This runs fine, but (as Danny has told me) it produces no results in the output window.  It will do graphs.  So (undaunted) I went to method 2!

2) I put stats::wilcox.test(Strt1) into the calculator and opted not to store the results.  This now gives just what I wanted.  Yipee.  The code in the script window is quite simple too.

So I can do it.  Is Version 2 to be our route?  Or at least one of our routes?  If so then could we have another keyboard with these functions?  They are designed simply to put results into the output window.  If so, the one useful additional feature could be to disable the saving when that keyboard is selected.  These commands simply put results into the output window.  So they don't add anything to the Instat object - which the last_model line in method 1 is trying to do.

3) If we are to enhance the calculator as our route to the execution of these simple commands, then we might want to use a command from a package that is not currently in R-Instat.  Currently we have to load it through RStudio.  Could we have a facility to load a package through R-Instat

dannyparsons:
From looking into this it is now possible to display output through the script window, and I have updated it to allow this. So now by default it expects output from every line and tries to display it. This doesn't seem to cause any issues when there is no output to display, but this should now be tested to see if there are issues. If there are, we could add a "suppress output" checkbox on the script window to prevent this.

Separately, we could also have a keyboard for these functions as long as we don't try to save this back to the data.

It is possible to install and load packages through R-Instat using the R commands, although I would still be recommending users don't do this in case of unintended consequences.

rdstern:
[...]

PR #4931, opened 2018-09-06
Script Window Executing commands

Patowhiz:
Partly Fixes issue #4381 Executing commands from the script file.


maxwellfundi:
Will be testing this @Patowhiz  soon!

maxwellfundi:
@dannyparsons  I am happy with how this is working and it works well. I think its good now. I have seen you  had the concerns on some shortcuts thats why they were removed. 

ready to be merged

instat/ucrScript.vb
```diff
@@ -1,203 +1,226 @@
...
+        Else
+            mnuCopy.Enabled = False
+            mnuCut.Enabled = False
+            mnuRunSelectedText.Enabled = False
+        End If
+    End Sub
+
+    Private Sub ucrScript_Load(sender As Object, e As EventArgs) Handles Me.Load
+        mnuHelp.Enabled = False 'TODO. Remove this after help is developed
+        mnuRunCurrentLine.ShortcutKeys = Keys.Enter Or Keys.Control
```
dannyparsons: Can this not be done on the designer as for the other items?

dannyparsons:
This looks good. From testing this a few things I noticed, probably unrelated to these changes but since we're improving this now:

1. When I select all and delete I can undo to get it back, which is good. But when I use do a **clear** from the menu I can't undo. Not a serious issue as there's a warning before clearing, but if easy to also make clearing undo-able that would be good.

2. When I load script from file the file filter is .txt. I think this should be both .txt and .R (the two separate filters can be kept as well)

3. When I Save Script the default is .txt but I think this should be .R

dannyparsons:
4. I think it's correct not to have the warning now when you run every line. But I think there should be a separate warning before running the whole script. Something like are you should you want to run the whole script, changes made cannot be undone, process cannot be stopped once started etc.

(I can add a setting for this in options to turn off, and later it could be a checkbox on the warning message to turn the option off)

rdstern:
I am delighted this sounds to be moving forward so nicely.  If the warnings can be turned on and off in the Tools > Options then perhaps they could also become another option on the right-click?

instat/ucrScript.vb
Patowhiz: I didn't find a way of doing this at the designer ( mnuRunCurrentLine.ShortcutKeys = Keys.Enter Or Keys.Control) . The designer doesn't have the Enter Key. If someone has a  way of doing this I'll appreciate the help

dannyparsons: That's fine if it's not obviously possible from designer, just checking.

dannyparsons:
Yes I like that on the right click. But let's do that as a separate pull request.

Patowhiz:
@dannyparsons would you like me do this under this  pull request?

> This looks good. From testing this a few things I noticed, probably unrelated to these changes but since we're improving this now:
> 
> When I select all and delete I can undo to get it back, which is good. But when I use do a clear from the menu I can't undo. Not a serious issue as there's a warning before clearing, but if easy to also make clearing undo-able that would be good.
> 
> When I load script from file the file filter is .txt. I think this should be both .txt and .R (the two separate filters can be kept as well)
> 
> When I Save Script the default is .txt but I think this should be .R

dannyparsons:
Yes please, if quick.

Patowhiz:
@dannyparsons in regard to your [comment](https://github.com/africanmathsinitiative/R-Instat/pull/4931#issuecomment-419405556) above. Is it still necessary to have the warning message for the clear?

Patowhiz:
@dannyparsons  I'm not sure what to label the combination of the 2 file types. Is the current label okay with you?

script window

dannyparsons:
The label looks fine, though you need a space before the `(`.

instat/ucrScript.vb
```diff
@@ -1,203 +1,229 @@
...
+    End Sub
+
+    Private Sub Menu_Undo(sender As Object, e As EventArgs) Handles mnuUndo.Click
+        'Determine if last operation can be undone in text box.   
+        If txtScript.CanUndo Then
+            'Undo the last operation.
+            txtScript.Undo()
+            'Clear the undo buffer to prevent last action from being redone.
+            txtScript.ClearUndo()
+        ElseIf strClearedText <> "" Then
```
dannyparsons: I'm not sure about this. What happens if `CanUndo` is `False` for some other reason other than clearing and the cleared text is still there? Then we may get the cleared text back when it wasn't wanted.

I had a look to see if this is possible to get the clearing working correctly but there are only hacks which I'm not keen on https://stackoverflow.com/questions/34091319/how-to-clear-a-textbox-so-that-the-default-undo-method-still-functions. It's not so important so I'd be fine with leaving this out than spending more time on.

Patowhiz: I had ruled out all of such possibility as at the way the current code is structured. The clearedtext gets cleared everytime the text is appended through code. CanUndo will always be true as long as the change came from the user. I've also tested the possible scenarios to verify this. But I'm happy to remove this feature if it's more confusing the way it's currently implemented. I had to tweak it this since the textbox doesn't support undoing changes made internally through the code

Patowhiz: I had thought about mimicking the backspace as outlined in the link above but I wasn't quite sure about it. Would you like me use this alternative over the other or I just remove it all together?

textBox1.Focus();
textBox1.SelectAll();
SendKeys.Send("{BACKSPACE}");

dannyparsons: It would be a shame to remove it, since it's working. I'd be happy with a comprise: if you add comments anywhere the cleared text appears that will be useful for any future modifications needed or for debugging if something strange occurs.

Patowhiz: I'd go with this 
```
txtScript.Focus()
txtScript.SelectAll()
SendKeys.Send("{BACKSPACE}")
```
I've tested it and seems to be much direct to me. But I'll put a comment as to why this was preferred.

Patowhiz:
The above suggested changes have been made.

Patowhiz:

dannyparsons:
@africanmathsinitiative/developers this has been waiting review for 8 days

Muthenya:
Testing this now.

Muthenya:
This is just a thought, can the undo be improved since currently it just undoes the last typed or deleted action? 
Otherwise, this looks good from my end.

Patowhiz:
Yeah it could be improved. I didn't focus on it so much because my strong view is this control should be a rich text box or something better. I have opened an issue to improve and change this control.

dannyparsons:
Great, I just made a small change of added back the confirmation message when running all text (https://github.com/africanmathsinitiative/R-Instat/commit/945b87daa47a3f7472d8985955c687a1ee83934f). I think the warning is still needed in that case

Issue #4835, opened 2018-07-30
Issues concerned with commands in R-Instat


rdstern:
This issue is the opposite of what Danny and David keep asking for!  It relates to individual issues that have been raised and perhaps some new ones that will be added for specific tasks once there is general agreement.

Some users of R-Instat may later migrate to RStudio - either completely or partially.  They include the AIMS MSc students for whom we will be using R-Instat in October and also the health students envisioned by Prof Jim Todd in his interest in R-Instat, in relation a b it to Lily's MSc project.

The same could apply to climatic users, many of whom have been exposed to R, but found the learning curve to be rather steep.

[Reading, tweaking and using R commands.docx](https://github.com/africanmathsinitiative/R-Instat/files/2240725/Reading.tweaking.and.using.R.commands.docx)

I attach a guide I have drafted for these users.  It is incomplete, in that the figures are only included for the first main section.  But it would be good to agree on some of the points below, because they would affect both the text and the figures.  And I would welcome feedback of course.  (I have nominated some of the team in this issue.  Happy for any others to also become involved.)
It includes the script window, the log file (including moving to RStudio), and the use of what I have called halfway commands.  This is particularly the Calculator, but also the hypothesis tests, the Prepare > Column: Generate > Enter and hopefully the enhanced File > New dialogue. 

1. We rename **Instat Data Object** as **Data Sheet**, and **Instat Object** as **Data Book**.  David says that is very easy to do.  It affects the code in this guide for every command! 
2. We decide on changes to the toolbar (if any).  
a) Add an item to switch on and off the commands being in the output window.  We don't have a good idea for the icon yet.  Thoughts welcome.
b) Anything else?
3. There are small corrections to be made to the right-click on the log window - see #4822 
4. For the script window
a) Add the short-cut key <ctrl> <enter> to run selection (as in RStudio)
b) As in RStudio, the same key runs the current line if nothing is selected.  (Danny claims this should be fairly easy?)
5. The File > New has an option to include commands that populate the data frame, #4439.

rdstern:
I reply again on these items:
[...]

PR #4937, opened 2018-09-10
Output Window

Patowhiz:
Addresses issue #4835
Hiding and showing of commands in the output window.


Muthenya:
I've tested this and it's working well. I just have a thought.

Is there a need to have the comment when the commands are hidden? I don't find it necessary. I think it is more logical when hiding and showing both.

image

rdstern:
Good question.  That was discussed, because Patrick brought it up earlier.  We would still like to keep the comments - or at least keep the option of having comments and having R-commands separate.  I am hoping that soon users will not accept the default comments, but put in something more useful that reflects the output, rather than the commands.  I also like the bit of green being there to separate different parts of the output.

At this stage we are not even having the similar action for comments on the right-click.  So turning off comments is through Tools > Options.

But if you (and Patrick) still feel strongly, then perhaps we could have a similar right-click option for the Comments as we now have for the Commands?.

Patowhiz:
@rdstern to be honest. I strongly feel that the comments we are sending to the output window look and feel more of log events than output comments. They fit well in the log window,knowing where the code came from etc. But I'd expect comments on the output window to simply describe the output and not necessarily if its coming from the dialog or from which event(The user already knows this and if he wants to confirm he can easily go to the log window). If we could improve the output window comments then having the 2 options;switching of commands and comments will seem logical. 

Probably I miss the goal of the output window, I'm open to correction. 

For now, if we have to retain this the way it is then adding an option for switching off the comments too will be a safe bet. I'll remember to discuss this design with students at AIMS.

dannyparsons:
The option for comments is very different from the option to show/hide commands. Show/hide commands is purely an output window setting - the command always gets added to the log, or script, but might not be shown in the output window if the user wishes.

The option for displaying comments **by default** (since this is still an option on every dialog) determines whether a comment is associated with the code that is being run, this effects whether a comment is added to the output window, the log window and script window. So it's not an option of the output window it's more general, so I feel it shouldn't go on the Output Window right click.

Patowhiz:
Yep. That's a good point. Now I understand this. @dannyparsons what about having a separate setting for the output comments, just like the commands. Then we can switch the comments on the output window separately.  Then on our right click menu we can combine the switching of commands and comments to one. Something like hide (future) commands & comments.

Muthenya:
I agree Danny. My thought was more on the fact that the comment mentions the dialog or action that has generated the code yet in the event one has chosen **Hide (future) commands**, the comment is still there.

My thought now is if one wishes to hide the comments as well, they can just clear the comment from the dialog.

dannyparsons:
It can be useful to have the comments displayed without the command. If your command produces no output you can tell that anything actually happened if both are not visible - users rarely have the log window open. If users want to turn off commands I would be encouraging them to keep comments on so that they do see what's happening in the Output.

We probably should have a way of turning comments off from output window only. But I don't easily see a way to do this that isn't confusing with 1. the global show comments by default option and 2. the comments option on each dialog.

I can see users getting confused if they have output comments turned off because they won't see any effect of the comments checkbox on each dialog. This is sort of a failure of how we've set up these options so far but I see it as a bigger job to fix this at the moment and not having the additional option would be clearer at the moment.

Patowhiz:
I agree. @rdstern do you concur. If yes, then I think @dannyparsons this will be ready for your review.

Muthenya:
I thought this was ready @dannyparsons.

PR #4938, opened 2018-09-11
Log Window Functionality Addition

### LINKED ISSUE #4822: Right click on log file
I'm writing a guide about using commands when running R-Instat.

So I have opened the Log file.  Then I used <Right Click>.  
It has 3 items 
a) Open Log file
b) Copy
c) Run Selected Text

1. I tried Open Log File and it failed for me though it tried to go to pdf.  I assume it means Open Log File in RStudio?  Perhaps the name of the entry could change.
2.  I tried Copy and that seems OK
3. Run Selected Text is disabled.  Can it have a short cut key - just as I would like the same key in the Script window.  <Ctrl> <Enter> as in RStudio would be good.
4. In the script window there is Save Script.  There is no equivalent Save Log.

Patowhiz:
Fixes #4822 Right click on log file


Muthenya:
This looks good from my end.

PR #4940, opened 2018-09-11
ucrButtons Comment

### LINKED ISSUE #4825: Simple possible improvement to Comments field in dialogues
In preparing the guide to help those who may graduate to R it would be good if the Comments - that is on all dialogues could sometimes be more extensive.
In discussion with Danny we now suggest a small ... button which then provides a textbox into which the comment can be typed.  The default comment might be copied into it.  It would permit multiline comments.  It is sent to R just as the existing comment, but with a # in front of each line.

No special urgency, except it will affect all dialogues and hence would be good to do soon if we are going to do it.  I have therefore made it as 5.1.
If it is quite a bit of work, then I suggest we add the ... button very soon, (which is on all dialogues) but make it disabled for now.

Patowhiz:
Fixes #4825


Patowhiz:
Not yet

Patowhiz:

Muthenya:
I've tested and it works well. This is ready.

Patowhiz:

instat/ucrButtons.vb
```diff
@@ -275,4 +279,79 @@ Public Class ucrButtons
             cmdLanguage.FlatStyle = FlatStyle.Popup
         End If
     End Sub
+
+    'construct and format the comment
+    Private Function getFormattedComment(strComment As String, Optional bAppendFirstLineWithHash As Boolean = True) As String
```
dannyparsons: This should be done at the level of the `RunScript()` in the `RLink`. Then you will not need the workaround `bAppendFirstLineWithHash`.

instat/sdgComment.vb
```diff
@@ -14,12 +14,18 @@
...
+        txtComment.Focus()
+    End Sub
+
+    Private Sub txtComment_KeyDown(sender As Object, e As KeyEventArgs) Handles txtComment.KeyDown
+        If e.Control AndAlso e.KeyCode = Keys.Enter Then
+            Me.Close()
+        End If
     End Sub
 
     Public Sub setComment(strNewComment As String)
```
dannyparsons: Please use upper camel case for function names (first letter of each word capitalised), here and a few other places.

instat/ucrButtons.vb
```diff
@@ -275,4 +279,79 @@ Public Class ucrButtons
...
         End If
     End Sub
+
+    'construct and format the comment
+    Private Function getFormattedComment(strComment As String, Optional bAppendFirstLineWithHash As Boolean = True) As String
+        Dim strReconstructedComment As String = ""
+        Dim strCommentParts As String()
+        If strComment.Length > 0 Then
+            strCommentParts = strComment.Split(ControlChars.CrLf.ToCharArray(), StringSplitOptions.RemoveEmptyEntries)
+            For Each part As String In strCommentParts
```
dannyparsons: Use the naming convention for strings here.

instat/ucrButtons.vb
```diff
@@ -275,4 +279,79 @@ Public Class ucrButtons
...
+        'then set the  button properties
+        btnMoreComment.Text = ":::" 'temp. This will be shown as centered ... An image is preferred
+        'btn.Image = Image.FromFile("C:\patowhiz\3dots.png")
+        btnMoreComment.Size = New Size(25, txtComment.ClientSize.Height + 2)
+        btnMoreComment.TextAlign = ContentAlignment.TopCenter
+        btnMoreComment.FlatStyle = FlatStyle.Standard
+        btnMoreComment.FlatAppearance.BorderSize = 0
+        btnMoreComment.Cursor = Cursors.Default
+        'btn.Margin = New Padding(btn.Margin.Left, 0, 0, 0)
+        'btn.Location = New Point(txtComment.ClientSize.Width - btn.Width - 1, -1)
```
dannyparsons: Commented code should have a comment explaining why it is there (as you have below for location). If this isn't needed then just remove.

instat/ucrButtons.vb
```diff
@@ -275,4 +279,79 @@ Public Class ucrButtons
             cmdLanguage.FlatStyle = FlatStyle.Popup
         End If
     End Sub
+
+    'construct and format the comment
+    Private Function getFormattedComment(strComment As String, Optional bAppendFirstLineWithHash As Boolean = True) As String
+        Dim strReconstructedComment As String = ""
+        Dim strCommentParts As String()
+        If strComment.Length > 0 Then
+            strCommentParts = strComment.Split(ControlChars.CrLf.ToCharArray(), StringSplitOptions.RemoveEmptyEntries)
```
dannyparsons: The separator should be `Environment.NewLine` (`CrLf `is only there for compatibility with older VB code) and we should not remove empty lines, they may be deliberate by the user.

Patowhiz: What you are proposing here won't work the way it's needed. Not unless you propose to me another way of constructing the comments.

dannyparsons: `strComment.Split(Environment.NewLine.ToCharArray())`

Patowhiz:
@dannyparsons I have modified the code in a way that we don't need the sdgComment subdialog. I was influenced into creating it because I wasn't quite sure if this kind of a popup will be accepted as an alternative to our usual sub-dialogs. I'm using a form because I didn't find a good way of having a control show up in a similar manner without it being inside a form container. If it exists, I'm happy to change this implementation.

instat/sdgComment.vb
Patowhiz: This subdialog can now be deleted

instat/ucrButtons.vb
Patowhiz: `strComment.Split(Environment.NewLine.ToCharArray())` will create duplicate empty lines in the resulting array that's why I had to use` StringSplitOptions.RemoveEmptyEntries` parameter option to remove them . Try this in the code and you will see the problem I faced, I had to settle with removing all the empty lines. I didn't know the CrLf was obsolete so I have changed this to use what you have recommended
`strComment.Split(Environment.NewLine.ToCharArray(), StringSplitOptions.RemoveEmptyEntries)`

instat/ucrButtons.vb
```diff
@@ -275,4 +279,112 @@ Public Class ucrButtons
...
+                                             'if user wanted below the control. Left here for future reference
+                                             'sdgComment.Location = New Point(ctlpos.X - 2, ctlpos.Y + txtComment.Height - 2) 'then locate its position
+                                             frmPopup.Location = New Point(ctlpos.X - 2, ctlpos.Y - frmPopup.Height - 2) 'set location to show the form just above the examples button
+                                             frmPopup.Show()
+
+                                         End Sub
+
+    End Sub
+
+    'TODO. This can be removed once the above method is agreed to be a better implementation.
```
dannyparsons: I agree this can be removed now.

dannyparsons:
Looks good. Two small points:

- When the textbox is open, when I click the `...` button again I expected it to close, but it just reopens again. If it could close that would be more what would be expected I think.
- When the textbox opens all the text is highlighted each time, could it not be?

These are very minor so if there is any complication to do these then leave it.

instat/ucrButtons.vb
Patowhiz: Removed

PR #4943, opened 2018-09-12
data_book rename

### LINKED ISSUE #4835: Issues concerned with commands in R-Instat

dannyparsons:
fixes #4835

- Renamed `instat_object` class to `DataBook`
- Renamed R-Instat's instance of this to `data_book`
- Renamed `data_object` class to `DataSheet`

@africanmathsinitiative/developers please test this carefully as it effect much of our R code. Please test a variety of simple and complex dialog which use the DataBook (instat object)


Patowhiz:
But this is already merged?

dannyparsons:
Sorry, I automatically merged without thinking. But please test in the master and report any problems.

maxwellfundi:
This was merged before it was tested properly and it has caused a lot of trouble with @rdstern because an important dialog for the workshop- QC Temperatures is unusable at the moment because Instatobject was  not renamed properly on this dialog. 
 
Below is the error we get that clearly shows it cant find instatdataobject 

image

Issue #4607, opened 2018-05-22
Rainfall quality control at month level


rdstern:
In the **Climatic > Check Data > Rainfall QC** dialogue the one item that hasn't been implemented is the check that the total rainfall for the month is totally zero.  (This is important, because sometimes it may really be zero for the whole month, or it could be that the data were missing and mistakenly replaced by zero.  Excel makes that sort of thing easy!)

I suggest a simple way this could be implemented on the dialogue.  The item is already there as Dry Month. 
 It has < with a numeric field having a default of 0.05mm.  Then a button with Omit Months.  Or it is a control on the main dialogue.  But it opens a control like the filter, when there is a factor and you can tick, which months to omit.

These are the months when having totally zero isn't a surprise.  For example, in Niger stations that would be November to April. 

Now the calculation.  My breakthrough in thinking is that this would be easy if it gives the results in a monthly data frame, rather than the daily data of the other QC checks.  So that first.

There is the filter if the Omit months is there.  Then the monthly rainfall summary, giving the total and then a filter on that monthly data frame to just show the months with zero total.

Of course this is a different data frame to the other QC checks.  I don't see that as a problem.  Indeed I now see this as an opportunity, because the paper records are usually also at a monthly level.

So, as a further step I wonder (perhaps later) whether the other QC checks might have an option to also output summary results to the monthly data frame.  They might be the sum of the logical columns (at the daily level), which would therefore give the number of times that check has failed in the month.  That could be really useful in the checking process back to the paper records - and even more useful if we could sometimes have images of the paper records linked to our data.

PR #4950, opened 2018-09-13
Qc Rainfall.

### LINKED ISSUE #4947: Problem in Climatic > Check > Rainfall QC outliers
To see the problem load the Dodoma.rds data and choose the option to get outliers.  I get 972. The upper limit of the outliers is zero, which is why there are so many.

I tried with and without omit zero in the dialogue and can not see anything different in the R-code.  It gives 972 either way.  So I think it may be just they this bit isn't working.

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes #4947


shadrackkibet:

shadrackkibet:
Can someone test this soon? will it be efficient if this is merged and @rdstern  could check? I would like this to be merged before i work on another issue (#4607) which has also been discussed with @volloholic

PR #4953, opened 2018-09-14
Conversion

Muthenya:
This is the new Conversions pull request. Fixes issue #4845.
@dannyparsons you can now have a look at this.


dannyparsons:
Could @shadrackkibet test this now he's back?

shadrackkibet:
Great, Testing this now.

shadrackkibet:
I think this is now good as an initial implementation of this dialog. So i recommend this to be merged.

dannyparsons:
This runs well so great, but I have some questions and comments:

**Units**
- The **To** combo box seems to be in a different position and/or a different size for Rain than for the other two.
- I found this format a bit strange for Wind Speed. The other two elements don't have this format so it was offputting to me at first. I think they could be: Knots, Miles per hour (mph), Metres per second (mps), Feet per second (ftps), Kilometres per hour (kmph). I think we are generally using English spellings in the software instead of American so its metre.
image
- The units for temperature should be capitalised.
- The default units for the From and To should be different. From inches To mm seems a good default for rainfall @rdstern could suggest for the others.
- Parameters weren't in the correct order when I ran it
- When `old_metric` was the same as `new_metric` I got an error that they had to be different. Shouldn't we then prevent this on the dialog?
- Decimals is maximum 3, why?
- Why not default of 2 for decimal, which is the function default?

**Day Length**
- I would suggest a better prefix for the new column when on Day Length
- What is the station receiver for? It doesn't appear in the code.
- When I use a column for latitude and reopen the dialog I get an **Entry is not numeric** message box
- The parameter name is missing for `lat`

Also, what's the logic behind the current order of units in the combo boxes? It's not alphabetical and I can't work out if it's anything else?

Muthenya:
I have already made some corrections on the dialog but there are some which require your input @rdstern.

rdstern:
You leave me to guess what it is that needs my input.  Perhaps the items can be discussed with @volloholic when he arrives later today?

Muthenya:
Sorry for not being clear Roger. What I meant was I would like to know if there is anything from Danny's suggestions that you needed to answer to or had a different opinion on. I will as well discuss with David.

shadrackkibet:
@rdstern  I would like you to answer the following questions , We will address  others accordingly!

- > The default units for the From and To should be different. From inches To mm seems a good default for rainfall @rdstern could suggest for the others.

- > Decimals is maximum 3, why?

- > What is the station receiver for?

rdstern:
temperature can be Kelvin to Celsius
wind speed can be Knots to mps (Metres per second

The sensible default for these measurements is 1.  So I can't see why you even need 3, but let's make it 5 if you like.

You are right on the station.  I was including it, because it is there for all the other climatic dialogues.  But let's omit it here.

Muthenya:
> When I use a column for latitude and reopen the dialog I get an **Entry is not numeric** message box
> The parameter name is missing for `lat`

I given different parameter names for the textbox and receiver, `lat` and `latitude` respectively, so as to avoid the error message on load.

For the case of the parameter missing, I have changed it to be included when the parameter value is from the textbox and to be excluded when the parameter value is from the receiver.

rdstern:
I would be happy simply for OK not to be enabled when they are the same.

Muthenya:
This is ready now.

instat/dlgConversions.vb
```diff
@@ -84,11 +84,13 @@ Public Class dlgConversions
         dctPrecipitationUnits.Add("mm", Chr(34) & "mm" & Chr(34))
         ucrInputFromPrecipitation.SetItems(dctPrecipitationUnits)
         ucrInputFromPrecipitation.SetDropDownStyleAsNonEditable()
+        ucrInputFromPrecipitation.SetRDefault("inches")
         ucrInputFromPrecipitation.bAllowNonConditionValues = True
 
         ucrInputToPrecipitation.SetParameter(New RParameter("new_metric", 2))
         ucrInputToPrecipitation.SetItems(dctPrecipitationUnits)
         ucrInputToPrecipitation.SetDropDownStyleAsNonEditable()
+        ucrInputToPrecipitation.SetRDefault("mm")
```
dannyparsons: Are these really the R defaults? It looks like you are confusing R default and dialog default. In general, they are not the same. And when they are different, like in this case, setting something as the R default when its not will cause bugs.

dannyparsons:
The station receiver was there because we might want to get the latitude column from a linked data frame where we have latitude for each station. This would be great, but we'll need to think how best to do this so I agree to removing it for now.

dannyparsons:
I agree to disabling OK when they are the same. I don't think it will be that confusing, they should notice they are trying to do something that isn't sensible.

dannyparsons:
> I have given different parameter names for the textbox and receiver, lat and latitude respectively, so as to avoid the error message on load.
> 
> For the case of the parameter missing, I have changed it to be included when the parameter value is from the textbox and to be excluded when the parameter value is from the receiver.

I don't like doing this because its not how we want to discuss R code, and it's not consistent. But I can see why this was needed to be done to get it done within our system. Can you add comments in the code where you have done this to say why this has been and that in future we need a better way to do this?

instat/dlgConversions.vb
```diff
@@ -259,6 +259,7 @@ Public Class dlgConversions
             ucrReceiverElement.SetMeAsReceiver()
             ucrSaveConversions.SetPrefix("Conversion")
         End If
+        ChangeLatParameter()
```
Muthenya: This is here because the ucrPnlConversions controlValueChanged event results in the removal of the `lat` parameter. It fixes the bug. Is there a better way to deal with this @dannyparsons?

rdstern:
Sorry I may have misled you above, when I said there was no need for a Station field in this dialogue.  There is no need when doing the changes in units, but the Station is useful when getting the day lengths (or it could be).  We certainly need latitudes for the day lengths.  If they are there now, then we can get them through merging the station-level data with the daily data to give a latitude column.

If that is what is done now, then fine.  Later (as with some other dialogues, e.g. End of the rains) it would be good to be able to have the option of the Station and then to link to the data frame with the latitudes.  Let's leave that for later.

instat/dlgConversions.vb
```diff
@@ -268,4 +268,17 @@ Public Class dlgConversions
     Private Sub ucrPnlElements_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlElements.ControlValueChanged
         SetBaseFunction()
     End Sub
+
+    Private Sub ChangeLatParameter()
```
Muthenya: This was done since giving different parameter name for the latitude text box and receiver produced a bug (both the controls' parameters were in the function)

instat/dlgConversions.vb
dannyparsons: Yes looks good.

shadrackkibet:
Testing this again

shadrackkibet:
Only one issue remaining here and this should now be ready: On **Rain** From and To comboboxes do not reset to the default values.

Muthenya:
This is ready now @dannyparsons.

instat/dlgConversions.vb
```diff
@@ -158,6 +156,8 @@ Public Class dlgConversions
         ucrSelectorConversions.Reset()
         ucrSaveConversions.Reset()
         ucrInputLatitude.SetName("")
+        ucrInputFromPrecipitation.SetName("inches")
+        ucrInputToPrecipitation.SetName("mm")
```
shadrackkibet: we shouldn't be setting name here.

Muthenya: This has now been resolved.

shadrackkibet:
This is now working better. @dannyparsons  this is ready for you.

dannyparsons:
Looks good, please add to frmMain

PR #4958, opened 2018-09-17
Adding options to work with non-day units on Make date 

### LINKED ISSUE #4864: Extending the Make Date dialog: origin option with non-day units
I think this may be a relatively simple addition in the Make Date dialogue.
It is where there is a single column and the origin is specified, e.g. midnight on 1 Jan 1970, but the existing numbers are in seconds from that origin.

We already have the structure in place in **Prepare > Column: Date > Make Date > Single Column > Origin > Specify**.

This gives a date picker - looks nice.
a) Change the label Specify to **Specify Date** and add another option to the list **Specify DateTime**.
b) The sort of control is shown [here and looks nice](https://stackoverflow.com/questions/93472/datetimepicker-pick-both-date-and-time). 
c) This gives the time in hours/minutes/seconds and the examples I have seen so far have seconds as the units.  Do we stop there, or also offer further options for situation where the time is given in minutes or hours?  
d) In that case we might have either 3 options for **Specify Time (Seconds), Specify Time (Minutes), Specify Time (Hours)**.  Each would have a different up-down control for the time component.  Or we stick with the single option and add radio buttons with the label **Units** and the 3 options of Seconds, Minutes, Hours.

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes #4864


maxwellfundi:
testing this

maxwellfundi:
Looks good @shadrackkibet

dannyparsons:
@shadrackkibet @maxwellfundi how well did you test this before merging? I just tried, for example:
1. Import Dodoma data
2. Convert Date column to numeric
3. Multiply by `24 * 60` to get into minutes
4. Try convert back into a date using Make Date. This is what I got (`Date3`) which is obviously wrong.

image

shadrackkibet:
Thanks for spotting this. I tested for Hours that is by `*24 `and it worked for me, so i thought it is going to work for others as well.Now i see there is an issue having this without brackets. So i think to fix the issue we need to add brackets so that we have `x/(24*60)` and `x/(24*60*60)` or we have `x/1440 `and `x/86400 `as  denominators.

dannyparsons:
Yes that looks like the problem. Roger preferred that brackets so if that's easy then do that, otherwise there isn't a big problem with the other way.
This shows the importance of testing being done properly, which is why we have the two levels of testing, everything added or changed needs to be checked.

PR #4961, opened 2018-09-18
Saving fitted values, residuals and hatvalues for models

### LINKED ISSUE #4803: Saving fitted values and residuals
The modelling dialogues use the same Display button which goes to a sub-dialogue.  This includes a Save Tab, with the 4 options to save, but they are currently disabled.

I am not clear whether the code is already there?  If not, then I think the commands are just fitted, resid, rstandard, hatvalues.  They each produce a column of the same length as the data, so can go back into the same data frame.

I hope this will be easy for someone from the AMI team?

shadrackkibet:
In progress Fixes  #4803


shadrackkibet:

shadrackkibet:
@africanmathsinitiative/developers can someone look at this please? . I would like to use changes on the subdialog from here in another pullrequest!

maxwellfundi:
@dannyparsons  This looks good and the save tab now works well and produces the needed columns. However as shown by below code, it seems to repeat code- say you so just residuals then it will have the residuals code, when you come back to the sdg and do fitted values then it will run both the residuals and fitted values code and so on as shown below 

image

dannyparsons:
@shadrackkibet can you see the cause of this?

shadrackkibet:
All my changes are now pushed. @dannyparsons  Please  have a look at this.

dannyparsons:
Not yet found the cause of this problem but there's a more serious issue that all these calculations remove missing values so if the data has missing values the fitted values etc. will be shorter than the data frame length and won't match up when trying to add as a column. We'll probably need a special method to do this properly.

dannyparsons:
Actually I have seen from here https://stackoverflow.com/a/19065945/6424231 we can just add `na.action = na.exclude` to pad the result will missing values and get something or the right length!

But also see from here https://stackoverflow.com/questions/19065279/in-r-how-to-add-the-fitted-value-column-to-the-original-dataframe#comment72923299_19065945 that this only works if you use the `residuals` function not accessing them directly.

We'll need to look to see if similar functions exist for the other components or write them ourselves.

rdstern:
And the fitted values are sometimes OK.  If the x-s are available, but not the y, then you still get the fitted values.  If some x's are missing, then not.  The result is then missing.
I would be happy if we could sort out fitted values and one of the residuals.  But I think if we can sort out 1 then possibly all.

This feature is important for AIMS November, but not necessarily for Version 4.23

dannyparsons:
Even if we have to write our own functions for other components it won't be too difficult, just a bigger task than we thought.

shadrackkibet:
@dannyparsons I have found similar functions which we can use instead of writing special methods to do this.
```
residuals()/resid()
fitted.values()
rstandard()
hatvalues()
```
All this will work with `na.action = na.exclude` and give a column of the same length as the original dataframe. However, i have noticed that `hatvalues()` will give zeros when there are NA values which is not consistent with other option.

dannyparsons:
Ok that's great. And interesting about `hatvalues()`. I would suggest we use these for now, even for hat values, as it's the best we have got and its going to work with missing values. Then we can see if we want to do anything different after we have something stable in place.

shadrackkibet:
I have noticed that allowing missing values in a model will lead into another problem. Currently we are using autoplot() from ggplot2 to get plots from a model. Running `autoplot` function with NA values gives an error. 
`Error in ``$`<-`.data.frame``(``*tmp*``, .hat, value = c(0, 0, 0, 0, 0, 0, 0,  : 
  replacement has 28855 rows, data has 20194`

My suggestion is to have `na.action = na.exclude` as a option in the dialog. Preferably as a checkbox.I don't have a better way to avoid this bug. @dannyparsons  and @rdstern  what are your views on this?

dannyparsons:
Well noticed. Can you look up to see what others say about this problem? I'm sure it's common enough to have been looked into by others.

shadrackkibet:
> Well noticed. Can you look up to see what others say about this problem? I'm sure it's common enough to have been looked into by others.

@dannyparsons haven't found something sensible to do here. From [here](https://stackoverflow.com/questions/47232728/error-in-dataframe-tmp-replacement-has-x-data-has-y) they are suggesting use of `na.omit`. I am not sure if this is something sensible.

dannyparsons:
I don't think we want to use `na.omit` on the whole data frame because it will remove rows with missing values in any column - even if the column isn't in the model.

Can you confirm that it's `ggplot2::autoplot` and not `fortify::autoplot`? We use both packages and when I ran it it didn't have the package name. We should use `ggplot2`, if it still gives an error we might find something different in `broom` or raise it as an issue.

shadrackkibet:
I have added package name and there is still a problem. I am going to raise this as an issue.

dannyparsons:
There might be alternatives to autoplot in another package, or a solution for this case. I would suggest posting it as a question here https://community.rstudio.com/c/tidyverse is the best place. Make sure to give a reproducable example when you're including code.

shadrackkibet:
I have posted this [here](https://community.rstudio.com/t/error-in-ggplot2-autoplot-function-with-linear-model-lm-when-na-action-na-exclude/16188) . let's wait and see what the R Studio community say about it.

shadrackkibet:
@africanmathsinitiative/developers the saving of this columns is done and working well. So, What is remaining here is a bug with the `autoplot` function when you want to get the  residual plots. I have reported this bug as discussed above. I have also discussed this with @Muthenya

dannyparsons:
There didn't seem to be any response on RStudio so I [reposted on stackoverflow](https://stackoverflow.com/questions/53061605/error-in-ggplot2autoplot-on-linear-model-with-na-action-na-exclude).

dannyparsons:
@shadrackkibet I fixed the issue of the R code repeating in the output. It was an issue in clsRSyntax constructing the code. Other than the `autoplot` issue is there anything else that needs to be done?

shadrackkibet:
What should be done now is linking models objects into respective functions in other model dialogs that use this subdialog.

dannyparsons:
I'm about to upload that soon.

PR #4964, opened 2018-09-19
Delete rows columns

### LINKED ISSUE #4955: Opening Prepare > Data Frame > Delete Rows/Columns
**Describe the bug**
I go into R-Instat and try Prepare > Data Frame > Delete Rows/Columns

I get the error: 

**instat
Developer error: The value given cannot be converted to a decimal or is outside the range of the control. Value will be unchanged.
OK**   

Now I am trying to do something pretty silly, because I have just started R-Instat and don't have a data frame.  (I just wanted to see what the dialogue looked like).
But this is a meaningless error message.  It gives it twice and then goes into the dialogue.
Once I try again it is fine.
Other dialogues that I tried, e.g. Insert Rows/Columns don't seem to have this problem.  I have tried with all the dialogues in **Prepare > Data Frame** and **Prepare > Check Data**.  The only problem I found was in Export to Open Refine.  I report that separately.

Muthenya:
Fixes #4955.
@maxwellfundi, @dannyparsons this is ready.


shadrackkibet:
This seems to fix the problem but now, this two Nud's do not reset.

instat/dlgDeleteRowsOrColums.vb
```diff
@@ -75,13 +75,13 @@ Public Class dlgDeleteRowsOrColums
         clsOperatorRowNames = New ROperator
 
         ucrSelectorForDeleteColumns.Reset()
+        ucrNudFrom.SetText(1)
+        ucrNudTo.SetText(1)
```
dannyparsons: there shouldn't be any settext done in defaults, but the parameters can be added

Muthenya: I initially had added the parameters but the issue is the value in the Rcode would reset well but the value in the nud would remain as it was before reset.

dannyparsons: Did you work out the cause of the original issue? I suspect it is from this line https://github.com/africanmathsinitiative/R-Instat/pull/4964/files/d0e6225b347a743272ae31ba60ce07dbd6809cfb#diff-4d8f9c71a81eaf9492e10debb7c68633R61 which does not belong in initialise anyway and will set the max to 0 when there is no data.

maxwellfundi:
@Muthenya could you fix this so that we have it in the next release. @dannyparsons this should be easy?

dannyparsons:
I would put it very low priority for the release as its a small issue. Everything else should be done first.

Muthenya:
@dannyparsons I haven't been able to make progress on this. I know it's not urgent but I would like to know the way forward on this.

Muthenya:

Ogik99:
This looks good

Muthenya:

Patowhiz:
Well the problem stated in the issue is solved. The only other issue I found which is a different error based on an action that should not be allowable.

Muthenya:
Well spotted. This happens on reopen.

dannyparsons:
The nud should be limited to the maximum number of rows. This should be set every time the dialog opens as the number of rows can change between opening.

Muthenya:
I thought we might need a check for when the user has deleted all the rows but it doesn't make sense for all the rows or columns to be deleted.

instat/dlgDeleteRowsOrColums.vb
```diff
@@ -133,12 +134,19 @@ Public Class dlgDeleteRowsOrColums
...
+        SetMaxMin()
     End Sub
 
     Private Sub CoreControls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverForColumnsToDelete.ControlContentsChanged, ucrNudFrom.ControlContentsChanged, ucrPnlColumnsOrRows.ControlContentsChanged, ucrNudFrom.ControlContentsChanged
         TestOKEnabled()
     End Sub
+
+    Private Sub SetMaxMin()
+        If ucrSelectorForDeleteColumns.ucrAvailableDataFrames.cboAvailableDataFrames.Text <> "" Then
+            ucrSelectorForDeleteColumns.Reset()
```
dannyparsons: The selector shouldn't be reset on reopen.

dannyparsons:
Try it out. It is actually possible in R-Instat to have a data frame with 0 rows - this can actually happen quite a lot e.g. doing a filter or subset where the condition is never met. We do not allow a data frame with 0 columns however. So you do need to check that the row count is not 0.

Muthenya:
I understand Danny, and agree with the case of the filter, but my reason for saying this is I couldn't think of a reason why one could delete all rows using the Delete Rows/Columns dialog.

dannyparsons:
Ah ok I see. Well we don't need to find a reason why it might be done. It possible in other way, by subset or by right click on all rows, so there's no reason it shouldn't also be possible through the dialog.

However, deleting all columns will delete the data frame so this should not be allowed from the dialog. (Consistent with it not being possible to delete all columns from the right click too)

Patowhiz:
@Muthenya did you make the above suggested changes. I would like to retest this dialog if it's ready

Muthenya:
I'm not yet through with all the changes. I'm having issues with the nuds changing their limits after rows are deleted.

Muthenya:
@Wycklife please test this.

Muthenya:
@africanmathsinitiative/developers this is still ready for review.

Muthenya:
@Wycklife please test this.

Wycklife:
@dannyparsons @maxwellfundi  I have tested this and it works well without bugging

Issue #4915, opened 2018-08-30
Enhance the Climatic > File > CLIMSOFT dialogue


rdstern:
There is a bug that I hope can be fixed by Verison 4.23.  Then there are various enhancements that would be useful in their own right, and would also set a good example for CLIMSOFT dialogues.

Only the bug need be fixed for Rwanda - other stuff would be nice!
1. The bug is that the connection to the database can only be made once in a session.  It works fine.  But if you try again after disconnection, then it gives an error.  It has a line in the error message I don't quite understand which is something like widget followed by a number.  This number gets one larger every time you try the connection.
2. **The sub-dialogue** (making the connection could now be made easier to use.  There are 4 fields.
The first is the name of the database.  There are 2 names now.  I suggest the first is given as the default, and there is a pull-down, tha currently just contains the 2 names.  But there could be more later I assume. The first (default) is mariadb_climsoft_db_v4 and the second is mariadb_climsoft_test_db_v4.
The third field could perhaps also be a pull-down.  It contains 3308 as the default and I'll check with Marcellin what other values might sensibly be there.  For now also have 3306.
3. The **main dialogue** could be made a bit easier.  It could also add flexibility in what is copied over.  Some aspects need more thought, but: 
a) A simple improvement would be to be able to display the station names as well, or instead of the IDs in the selector.  And possibly, with a tooltip, to be able to display other station information, particularly the start (and possibly the end) of the record at that station.
b) A second could be to use the Element information in the dialogue.  Elements each have a **type**.  They are Daily, Monthly, Dekadal, Hourly, Synop and AWS and All.  If just one type is selected, then only the elements of that type are shown.  Eventaully perhaps only those stations are shown in the selector that has data of that type - but that might be tricky with the current software.
What would be useful from that, is the date-time column that is in the output.  It is currently not a date column, but could become one if the data were daily or monthly.
c)  There could also usefully be a checkbox for Transfer Element table.  Currently the Element information is given in the data table.  It is given in 3 ways, and that is tricky, because when made into factors, those columns will be in different orders of their levels. It would be better to just give the element number abd give the other information in the element table.
d) When multiple elements are transferred there could be a Tidy checkbox.  This does an unstack.
d) Later there might also be an Options button perhaps, as there is in the NetCDF dialogue.  
3. a) The main output data frame would become simpler in just giving the Element numbers.  And this column is made into a factor.
b) The Date column is made into a date when the data are daily or monthly.
[...]

PR #4967, opened 2018-09-20
Fixing Climsoft dialog widget bug

maxwellfundi:
@dannyparsons  I have added the button onto the sdgClimsoft that leads to dlgOptions where we have a new Climsoft tab with the fields- Database name, Port, Host and username.


dannyparsons:
This wasn't the suggestion as I understood it. I thought the button on the climsoft subdialog was to save the current values as the new dialog "defaults" (which can also be set in the options dialog)? The button means you can do this without going to the options dialog.

maxwellfundi:
Okay, this is was my understanding. if that is the case then its just changing its the text there.

maxwellfundi:

maxwellfundi:
@Patowhiz   the bug described on issue #4915 other improvements will be done later.

instat/sdgImportFromClimSoft.vb
```diff
@@ -121,4 +126,23 @@ Public Class sdgImportFromClimSoft
             cmdEnterPassword.Text = "Enter Password"
         End If
     End Sub
+
+    Private Sub cmdOptions_Click(sender As Object, e As EventArgs) Handles cmdOptions.Click
+        Dim msgSetDefaults = MsgBox("Are you sure you want to update Climsoft defaults?", MessageBoxButtons.YesNo, "Factory Reset")
+
+        If msgSetDefaults = DialogResult.Yes Then
+            If Not ucrInputDatabaseName.IsEmpty() Then
```
dannyparsons: I think actually this should change, even when its empty it should be set as the default. You may want to clear the default and this prevents it.

PR #4970, opened 2018-09-20
PICSA Rainfall

maxwellfundi:


maxwellfundi:
@dannyparsons  Have now added a little more functionality on to this from previous pull request. Please have a look at these issues that are remaining 
1.  On the main dialog have a check for levels of factors on the factor optional receiver 
On the sub dialog, 
1. Seeting up the R code properly for UcrInputBorderLinetype and

dannyparsons:
I should finish this by tomorrow.

maxwellfundi:
@dannyparsons  nice that this has made some progress. I have been testing it and lots of these options are working. 

I would like you to add na.rm= true for the y intercepts on the mean, median and quantile because if we have the missing values in the data then that throws a bug. 

Second, I was wondering if we have more that one level of factor on the factor receiver, since we have the number of lines as many as the number of factors, are we able to create equivalent number of lines for each of the factor? @rdstern is it more likely that people will have this situation?

PR #4971, opened 2018-09-21
New Data Frame

### LINKED ISSUE #4439: Extending the File > New dialogue even more
I now propose 3 of our radio buttons for the File > New dialogue.

One is just **New** and is the current dialogue. The second could be **Wakefield** and is discussed in another issue, see #4347.  The third is called **Data** and is a facility to add lines of data automatically see #4438. 
 The third option should be an easy addition?  It needs to have the number of rows (as does the first option), but not the number of columns.  It also needs the name of the resulting data frame.  There also is a wide text field for multiple lines, into which you can paste R commands.
There could also be a second, smaller field - perhaps on all options - where you could include a data frame label.

Then the R code needed is as follows:

**data2 <- data.frame(data=matrix(data=NA, nrow=20, ncol=0))
data2<-within( data2,{**
block<-c(rep("I",7),rep("II",6),rep("III",7))
trt<-c("A","B","C","D","g","k","l","A","B","C","D","e","i","A","B","C","D","f","h","j")
yield<-c(83,77,78,78,70,75,74,79,81,81,91,79,78,92,79,87,81,89,96,82)**}**
**)**

Here the data2 is the data frame name.  And 20 is the number of rows.  The three lines pasted into the text field are those starting block, trt, yield, see #4438.  

Piece of cake?
What isn't so clear is what messages will be given if those lines contain a mistake.  This can be checked partly by leaving out say one value from the yield column, so it is only of length 19 - in this example.

Patowhiz:
Fixes #4439 Extending the File > New dialogue even more.


Patowhiz:

maxwellfundi:
@Muthenya @JacklineKemboi @shadrackkibet can someone test this please?

Muthenya:
This seems to work well. Just some few issues
I noticed that there is an issue with the **ControlContentsChanged** since as shown it is not yet enabled, not until I click away from the cell.

image

I particularly like the Command option. Probably we could have more examples, but I still think it's okay.

Patowhiz:
@Muthenya Ideally I was thinking that the validation should happen after a user **commits** the change(sort of as I type I could be making mistake but correcting it before commiting shouldn't be a "crime"). So testOk is happening under `dataGridView.CellValueChanged` which in VB.NET is called when the user goes leaves the cell. The DataGridView is not among our custom controls. I could still implement this under keydown if the team feels it's a more viable option as you have stated above.

Muthenya:
I understand.

dannyparsons:
It's not just key down event e.g pasting should be covered too. There is no text changed for cells? If not I think `CellValueChanged` is ok for now. But this is something to change in future, possibly when it becomes a custom control, so that it's consistent with how other dialogs work.

dannyparsons:
I'll test this now as well.

instat/dlgNewDataFrame.vb
```diff
@@ -42,64 +43,248 @@ Public Class dlgNewDataFrame
...
+        'reset the controls
+        ucrNewDFName.Reset()
+        txtCommand.Text = "data.frame(data=matrix(data=NA, nrow=10, ncol=2))"
+        'empty and create 6 default rows
+        dataGridView.Rows.Clear()
+        For i As Integer = 1 To 6
+            dataGridView.Rows.Add()
+        Next
+
+        rdoConstruct.Checked = True
```
dannyparsons: This should not be done manually but through conditions on the panel, might be non trivial for the different cases but we should attempt it.

instat/dlgNewDataFrame.vb
```diff
@@ -42,64 +43,248 @@ Public Class dlgNewDataFrame
...
 
-        clsOverallFunction.AddParameter("data", clsRFunctionParameter:=clsMatrixFunction, iPosition:=0)
-        ucrBase.clsRsyntax.SetBaseRFunction(clsOverallFunction)
+        'reset the controls
+        ucrNewDFName.Reset()
+        txtCommand.Text = "data.frame(data=matrix(data=NA, nrow=10, ncol=2))"
+        'empty and create 6 default rows
+        dataGridView.Rows.Clear()
+        For i As Integer = 1 To 6
+            dataGridView.Rows.Add()
```
dannyparsons: This seems to add 6 more rows to the default 1 row so you get 7 in total. Also you can just do `dataGridView.Rows.Add(5)` instead of a loop.

instat/dlgNewDataFrame.vb
```diff
@@ -42,64 +43,248 @@ Public Class dlgNewDataFrame
...
+        lstView.Scrollable = True
+        lstView.View = View.Details
+        lstView.FullRowSelect = True
+
+        'add columns
+        lstView.Columns.Add("Command", 250)
+        lstView.Columns.Add("Comment", 200)
+
+        lstView.Items.Add(New ListViewItem({"data.frame()", "Empty data frame"}))
+        lstView.Items.Add(New ListViewItem({"data.frame(data = matrix(data = NA, nrow = 10, ncol = 2)", "10 rows and 2 columns filled with missing values"}))
```
dannyparsons: Missing a close bracket here, it gives an error running.

instat/dlgNewDataFrame.vb
```diff
@@ -42,64 +43,248 @@ Public Class dlgNewDataFrame
...
 
-    Private Sub ucrInputDataFrameName_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrNudRows.ControlContentsChanged, ucrNudCols.ControlContentsChanged, ucrNewDFName.ControlContentsChanged
+    Private Sub ucrPnlDataFrame_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlDataFrame.ControlValueChanged
+        If rdoConstruct.Checked Then
+            ucrNudCols.Visible = False
+            ucrNudRows.Visible = False
+            txtCommand.Visible = False
+            dataGridView.Visible = True
+            lblCommand.Visible = True
+            btnExample.Visible = False
```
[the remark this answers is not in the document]
Patowhiz: @shadrackkibet could you help me with the linking that @dannyparsons suggests here? I'm not quite sure on how this could be done by through linking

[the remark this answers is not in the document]
shadrackkibet: yes, i can do this and explain what i have done.

instat/dlgNewDataFrame.vb
Patowhiz: @shadrackkibet could you help with setting this through conditions on the panel, keeping in mind we have 2 options run the different functions but with same the same command, and the 3rd option doesn't pass a Function but rather a raw string command

instat/dlgNewDataFrame.vb
Patowhiz: I wanted the 7 columns intentionally, to remove the grey area of the datagridview. I have replaced the loop with `dataGridView.Rows.Add(6)`  and changed the comment to be more understandable

instat/dlgNewDataFrame.vb
Patowhiz: Thank you.

instat/dlgNewDataFrame.vb
```diff
@@ -42,64 +43,248 @@ Public Class dlgNewDataFrame
...
+
+        'Construct option function
+        clsConstructFunction = New RFunction
+        clsConstructFunction.SetRCommand("data.frame")
 
-        clsOverallFunction.AddParameter("data", clsRFunctionParameter:=clsMatrixFunction, iPosition:=0)
-        ucrBase.clsRsyntax.SetBaseRFunction(clsOverallFunction)
+        'reset the controls
+        ucrNewDFName.Reset()
+        txtCommand.Text = "data.frame(data=matrix(data=NA, nrow=10, ncol=2))"
```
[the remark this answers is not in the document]
Patowhiz: Yep. The ucrInputText control will do.

instat/dlgNewDataFrame.vb
shadrackkibet: How i am i going to add a row if i need more than 7?. We probably need a button somewhere to add a row.

Patowhiz:
@dannyparsons I have merged @shadrackkibet. I think the code is now consistent with way it's done in other dialogs and I'm okay with it. Works just as good too. You can have a look at this.

instat/dlgNewDataFrame.vb
Patowhiz: @shadrackkibet datagridview are internally made to add new rows by default whenever a user works on the last row. That's why I didn't implement the addition of row.

dannyparsons: An odd number of rows seems strange as a default, I'd still rather have 6, I don't mind about the grey area being visible. A smaller number is better as well - since users often feel they have to fill in everything that's possible.

instat/dlgNewDataFrame.vb
```diff
@@ -111,6 +123,23 @@ Public Class dlgNewDataFrame
                             ucrBase.OKEnabled(False)
                             Exit For ' if not valid text then exit for
                         End If
+
+                        'check for duplicates
+                        bDuplicate = False
+                        For i As Integer = dataGridView.Rows.IndexOf(row) - 1 To 0 Step -1
```
dannyparsons: A double For loop is very inefficient. Better to create an array of the column names in the first loop and then check the array for duplicates.

dannyparsons:
Looking better now. Has this been addressed? I can't see a change related to it in the code.

> Column numbers are editable in Construct. Didn't seem to cause any bug but they shouldn't be editable.

dannyparsons:
For the **Construct** option I was able to give column names and have the expression blank, or put in an expression but the column name was blank. In both cases the row is ignored which isn't silly because the row is incomplete, but it wasn't clear to me that was going to happen. We could require that rows are complete before to click ok, or highlight incomplete rows somehow/display message. Not sure if any of those are preferable to what's there already though. Any thoughts?

instat/dlgNewDataFrame.vb
Patowhiz: Removed duplicate inefficient for loop

instat/dlgNewDataFrame.vb
Patowhiz: changed to dataGridView.Rows.Add(5)

Patowhiz:
@dannyparsons this was addressed in the designer 

> Column numbers are editable in Construct. Didn't seem to cause any bug but they shouldn't be editable.

Patowhiz:
@dannyparsons in regard to this 

> For the Construct option I was able to give column names and have the expression blank, or put in an expression but the column name was blank. In both cases the row is ignored which isn't silly because the row is incomplete, but it wasn't clear to me that was going to happen. We could require that rows are complete before to click ok, or highlight incomplete rows somehow/display message. Not sure if any of those are preferable to what's there already though. Any thoughts?

Would you prefer disabling click ok?

shadrackkibet:
> For the Construct option I was able to give column names and have the expression blank, or put in an expression but the column name was blank. In both cases the row is ignored which isn't silly because the row is incomplete, but it wasn't clear to me that was going to happen. We could require that rows are complete before to click ok, or highlight incomplete rows somehow/display message. Not sure if any of those are preferable to what's there already though. Any thoughts?

If possible i would prefer highlighting the row and then disable Ok.

dannyparsons:
I'm not sure what would be best, I was opening it up for discussion.

Patowhiz:
@shadrackkibet keep in mind that highlighting the row will only happen when the user is not working on the datagrid probably (on leave event).

dannyparsons:
Why not on cell edit?

Patowhiz:
What do you mean by?
> Why not on cell edit?

On cell edit the selected row is the one that has the cell being edited. On cell end edit the next selected cell(which makes the next selected row) by default goes to the one below the cell which was being edited.

dannyparsons:
This could be done on `CellValueChanged` event so the user wouldn't have to leave the datagrid before this was done.

Patowhiz:
Should we have yellow for incomplete columns and red for columns with an error?

dannyparsons:
It depends what you mean by error. An error in the R command (which we want to implement later) could have just the cell highlighted in red.
It looks like you are using red for duplicate column names. I think that's different and I thought we were going to require unique names by not being able to leave the cell if there's already a column with that name. I prefer this because its not obvious that this is the problem if its just highlighted.

Patowhiz:
We have a message box outlining this. It tells the user that there's a duplicate. I don't like preventing the user from leaving a control. It goes against the rules of a good UX. For instance the user could want to correct the other cell not the current selected one

dannyparsons:
Ok that sounds fine. Then I think we have it highlighted in yellow too, and reserve red for when we check errors in R code.

Patowhiz:
Oka. We will have red for invalid column name and invalid expression

Patowhiz:
invalid column name to mean it's a reserved R keyword

shadrackkibet:
Testing this again.

shadrackkibet:
I have tested this and it now works better. @dannyparsons this is ready.

dannyparsons:
I'm getting a few inconsistencies here. Firstly, it only shows one incorrect cell at a time. When I had multiple gaps only the first one was highlighted in yellow. It would be good to check all cell and not stop when an error is found. Because of this, I could for example, type an invalid column name, but it didn't show as an error because an earlier cell was empty. 
I tried TRUE as a column name and it correctly told me that wasn't allowed. When I corrected it to a different name it stayed red when there was a previous cell highlighted in yellow. I think this is the same cause as above where it stops checking after the first incorrect value.

It's good that OK was always disabled when I had something wrong. It never let me click ok when rows were incomplete or had invalid/duplicate names.

Patowhiz:
@dannyparsons should we then remove the message boxes? they will be annoying if that's the case

dannyparsons:
Yes we probably don't want multiple message boxes appears each time a cell in changed.

Patowhiz:
And how will the user know what S/he is doing wrong?

Patowhiz:
Lets just leave the messageboxes then.

dannyparsons:
Sounds fine.

Patowhiz:
@dannyparsons this is ready for your reveiw

Patowhiz:
@dannyparsons this is ready for your review

instat/dlgNewDataFrame.vb
```diff
@@ -99,7 +200,100 @@ Public Class dlgNewDataFrame
...
+        lstView.View = View.Details
+        lstView.FullRowSelect = True
+
+        'add columns
+        lstView.Columns.Add("Command", 250)
+        lstView.Columns.Add("Comment", 200)
+
+        lstView.Items.Add(New ListViewItem({"data.frame()", "Empty data frame"}))
+        lstView.Items.Add(New ListViewItem({"data.frame(data = matrix(data = NA, nrow = 10, ncol = 2))", "10 rows and 2 columns filled with missing values"}))
+        lstView.Items.Add(New ListViewItem({"data.frame(x=1:30, y=rnorm(30,mean = 100, sd = 15), z = runif(30, min = 10, max = 30))", " "}))
```
dannyparsons: Just a spacing issue here.
```suggestion
        lstView.Items.Add(New ListViewItem({"data.frame(x = 1:30, y = rnorm(30, mean = 100, sd = 15), z = runif(30, min = 10, max = 30))", " "}))
```

instat/dlgNewDataFrame.vb
```diff
@@ -99,7 +200,100 @@ Public Class dlgNewDataFrame
...
+        lstView.Items.Add(New ListViewItem({"data.frame( block = gl(4,3), treat = c(""C"",""A"",""B"",""B"",""C"",""A"",""A"",""B"",""C"",""A"",""C"",""B""), yield = c(74, 68, 50,62,68,57,70,56,83, 67, 67,59))", " "}))
```
dannyparsons: and here
```suggestion
        lstView.Items.Add(New ListViewItem({"data.frame(block = gl(4, 3), treat = c(""C"", ""A"", ""B"", ""B"", ""C"", ""A"", ""A"", ""B"", ""C"", ""A"", ""C"", ""B""), yield = c(74, 68,  50, 62, 68, 57, 70, 56, 83, 67, 67, 59))", " "}))
```

dannyparsons:
This is ready other than the 2 spacing issues.

Patowhiz:
@dannyparsons changes applied

PR #4972, opened 2018-09-21
Rainfall Quality Control at Month level- Checking of dry month

### LINKED ISSUE #4607: Rainfall quality control at month level

shadrackkibet:
In progress Fixes #4607


shadrackkibet:
@rdstern Here is a screenshot of this option. What are your views in this?

image

dannyparsons:
It's a bit confusing that the button is **Omit Months** and the sub dialog is **Months to select** (the opposite). Maybe the button could be **Choose Months**?

shadrackkibet:
The point is that we need to select months to omit from the filter. So i suggest the label to be **Select Month(s) to Omit** for the subdialog.

rdstern:
Instead of Select Level could it be Omit Level?  The default for the Missing Value (not Values) Threshold should be 0.05 and the after the field it could say mm.  (Is that the value of the monthly total, or for any day?)  Can we make it so any positive number can be typed there?

shadrackkibet:
I think i misunderstood what this is meant for, i thought this is the sum of missing values in a month so Threshold will be something like 5.

 So, this is to check that the amount( Preferably Monthly total) is less than 0.05.
 If this is what you want then i can change this and also change this to an input so that you can easily type in any +ve value.

rdstern:
That's right.  It is the idea that if the total is really zero, then perhaps the month should be recorded as missing.

shadrackkibet:
@africanmathsinitiative/developers  this is now ready for review.

shadrackkibet:
@dannyparsons  i suggest you merge this for @rdstern to test.

PR #4973, opened 2018-09-21
Weighted Summaries Rcode.

shadrackkibet:


shadrackkibet:
Weighted summaries have been implemented for the following summaries:
**Mean,Standard deviation, Variance, Median, Quantiles, Skewness, Kurtosis, Coefficient of Variation, Median Absolute deviation,Correlation and Covariance.**

From my quick check on google haven't found functions to implement for the following functions. 
**Minimum, Maximum, n_distinct, Percentiles, Proportions, counts, Trimmed mean, Qn , Sn, mc, first, last, nth and standard error of mean.**

In discussion with @volloholic  he claims that he can implement this in many ways but he needs to discuss this with @rdstern first before we do it.

maxwellfundi:
@shadrackkibet  how is this getting on?

shadrackkibet:
There is still more to be done here. This will not be ready for .23

Muthenya:
There is another conflict here @maxwellfundi.

dannyparsons:
This may need to be checked carefully and copied into a new branch. @maxwellfundi can you have a look at this?

shadrackkibet:
@dannyparsons @rdstern we need to discuss this again. Is this still a priority?

dannyparsons:
Not sure how much of a priority but would be very nice to have.
It would be good to get the R code in place even if we don't change dialogs yet.
I see you looked into packages before as well. If they're common packages we're using already then good to use it but we should avoid adding a package for a single function if we can.

shadrackkibet:
Moved this here #5683

PR #4986, opened 2018-09-25
Transform

### LINKED ISSUE #4802: Raster package and more in Climatic Transform?
I needed to calculate a circular smoothing function for climatic work recently.  This will become useful generally when we have more plots with the x being the seasonality (day in the year) as well as time series plots (x is the dates or year etc.)

I needed a moving average which is circular.  There is a nice function in the raster package:
**raster::movingFun(prop10,20,circular=TRUE,na=TRUE)**

I couldn't find any other circular moving average?  With our current work on mapping I wonder whether we will need the raster package anyway?  That's a question partly to @JacklineKemboi ?

Then I suggest this could usefully go into the Transform menu of the calculator.  Perhaps it is called CircMov on the key?

And just to add the new functions that I hope we might have in the calculator in Version 5.0, i.e. for the Pretoria workshop.  They include:

1) This circular moving average on the transform keyboard.
2) Daylength added to the dates keyboard (#4749 )
3) Adding %in% #4795 to the logical and symbols keyboard
4) Adding a new Conversions keyboard.  Initially with the weathermetrics package and three keys.  This is from #4341:

This to have (initially) buttons called rain, temp and wind. They would implement 3 different commands, namely

convert_precip(x, old="m",new="i",round=2)
convert_temperature(x,old="c",new="k",round=1)
convert_wind_speed(x,old="kmph",new="kn",round=1)
We found, in Pretoria last year that temperatures were in degrees Kelvin and this would have been useful.
I am sure there will be more conversions later.

### LINKED ISSUE #4969: Cumulative distribution on Climatic > Prepare > Transform
This is probably a separate **Cumulative** button at the top of this **Transform** dialogue, perhaps at the start.  It would be good to discuss with @volloholic and @dannyparsons may add to this entry.

It is "simply" the cumulative sum, just as we can already do in the calculator.  It is on the Transform keyboard where there is cumsum, cummean, cummax, cummin.  For now I can only think we really need the cumsum, but it will be almost as simple to give them all at the same time.
The part that is not possible with the calculator is that we want to start the cumulative feature again each year and in each station.  
We may want to consider what happens when observations are missing.  The only simple action  is that the cumulative values are then missing from the first missing value to the end of that year..  
Once we have these cumulative sums (in another column) then we can see later how to move towards making the graphs shown in #4285 easier to produce.
If we want to "get clever" then the options I could think of are:
a) to be able to use the shifted year, rather than the January-December year, i.e. whatever is used in the Define Climatic Data.
b) Be able to start cumulating (and possibly end) within the year, rather than at 1 and 366.  In that case, all days up to the start day are (I assume) zero, and all days at the end are the value on the last defined day. (This is OK for the sum, but may not be sufficiently general.  Perhaps they have to be NA?)

Muthenya:
Fixes #4802 and fixes #4969.


Muthenya:
@dannyparsons this goes against our rule of having one pull request answering one issue, but I have added the cumulative option under this pull request.

Muthenya:
This is the implementation of the **circular** option and addition of the **cumulative** option.

instat/dlgTransformClimatic.vb
```diff
@@ -267,11 +332,20 @@ Public Class dlgTransformClimatic
         clsRMovingFunction.bToScriptAsRString = True
         clsRMovingFunction.SetPackageName("zoo")
         clsRMovingFunction.SetRCommand("rollapply")
-        clsRMovingFunction.AddParameter("width", 2, iPosition:=1)
+        clsRMovingFunction.AddParameter("width", 3, iPosition:=1)
```
dannyparsons: Just checking this is a deliberate change?

instat/dlgTransformClimatic.vb
```diff
@@ -267,11 +332,20 @@ Public Class dlgTransformClimatic
...
-        clsRMovingFunction.AddParameter("width", 2, iPosition:=1)
+        clsRMovingFunction.AddParameter("width", 3, iPosition:=1)
         clsRMovingFunction.AddParameter("FUN", "sum", iPosition:=2)
         clsRMovingFunction.AddParameter("fill", "NA", iPosition:=3)
         clsRMovingFunction.AddParameter("align", Chr(39) & "right" & Chr(39), iPosition:=4)
 
+        clsRasterFuction.bToScriptAsRString = True
+        clsRasterFuction.SetPackageName("raster")
+        clsRasterFuction.SetRCommand("movingFun")
+        clsRasterFuction.AddParameter("n", 3, bIncludeArgumentName:=False, iPosition:=1)
```
dannyparsons: We should be including argument names unless we need to exclude them for some reason

dannyparsons:
@shadrackkibet could you have a look at testing this?

instat/dlgTransformClimatic.vb
Muthenya: Yes. From the issue, Roger wanted this to be 3.

shadrackkibet:
Yes, i am going to test this.

maxwellfundi:
@shadrackkibet  how is this going?

shadrackkibet:
There are some few components i would like us to add into this before this is merged.

maxwellfundi:
@shadrackkibet are you able to do that this evening?

shadrackkibet:
I am working on another urgent task now. So i suggest we work on this by tomorrow Midday.

maxwellfundi:
great sounds good.

instat/dlgTransformClimatic.vb
```diff
@@ -204,33 +209,27 @@ Public Class dlgTransformClimatic
...
         ucrPnlTransform.AddToLinkedControls(ucrInputCircularPosition, {rdoMoving}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:="Right")
         ucrPnlTransform.AddToLinkedControls(ucrChkCircular, {rdoMoving}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrPnlTransform.AddToLinkedControls(ucrNudMultSpells, {rdoMultSpells}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=21)
+        ucrPnlTransform.AddToLinkedControls(ucrPnlEvap, {rdoWaterBalance}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrPnlTransform.AddToLinkedControls(ucrNudWBCapacity, {rdoWaterBalance}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=60)
+
+        ucrChkCircular.AddToLinkedControls(ucrInputPosition, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
```
Muthenya: @shadrackkibet I'm thinking of removing this line and setting up the controls manually as before since I haven't been able to resolve the bugs that came with this change.

instat/dlgTransformClimatic.vb
```diff
@@ -656,24 +639,7 @@ Public Class dlgTransformClimatic
         End If
     End Sub
 
-    Private Sub RasterPosition()
```
Muthenya: @shadrackkibet from my previous comment above, this is what I intend to add back. But I would be happy if you have a better way or possibly just have a look at the code again and let me know your thoughts.

Muthenya:
Changes pushed.

instat/dlgTransformClimatic.vb
shadrackkibet: What was the bug with this?

instat/dlgTransformClimatic.vb
shadrackkibet: I have looked at your code and I think I have fixed this issue. please test it and let me know if there is any problem.

Muthenya:
This has resolved the bug and is working well.

instat/dlgTransformClimatic.vb
```diff
@@ -137,7 +137,7 @@ Public Class dlgTransformClimatic
         dctInputSumPairs.Add("Sum", "sum")
         dctInputSumPairs.Add("Maximum", "max")
         dctInputSumPairs.Add("Minimum", "min")
-        dctInputSumPairs.Add("Mean", "(mean)")
+        dctInputSumPairs.Add("Mean", "mean")
```
Muthenya: I just noticed that the value for the mean had parentheses. I couldn't tell the reason for this so I removed them.

Muthenya:

instat/dlgTransformClimatic.vb
shadrackkibet: Beth added this after discussion with David for some sort of reason which I don't remember.

shadrackkibet:
I suggest we get what is currently here done and merged before we do anything else.

instat/dlgTransformClimatic.vb
Muthenya: Okay, I've added them back.

dannyparsons:
Yes, if its working with the brackets I suggest we leave them in and try to find out what they are for before changing.

Muthenya:
@dannyparsons please have a look at this. I have made some changes to the **Calculate by Year** checkbox. I'm not sure if this is what we wanted since if the **Year** receiver is filled it will always calculate by year. That is you cannot uncheck. This makes me think that the **Year** receiver can be set up like the **Station** receiver which always **Calculates by Station** when filled. Then we wont need the checkbox.

dannyparsons:
I discussed with Roger and we think that it's ok that cumulative is always by year. Unless you've already implemented it to be optional. However, fine if it's just always by year, and then as you say, the checkboxes are not needed.

Muthenya:
Okay. I'll set it up like so but just to be clear on this, this means that whenever the **Year** receiver is filled for all options it will always calculate by Year. Then the **Calculate by Year** checkbox will now be removed for all options.

dannyparsons:
No, if we don't need to do the calculation by year then we shouldn't. Cumulative will be done by year. I'm not sure which others are currently/should be done by year as well. We don't need to do the same for each calculation. Check with @rdstern.

rdstern:
Yes, I missed that point in the discussion earlier.  Some of the other calculations definitely aren't usually by year.  For example spell lengths can continue over years, so can water balance.

Muthenya:
This is how I've now implemented the Calculate by Year:
1. For the **Cumulative** option, it is calculated by year so long as the year receiver is filled. The **Calculate by Year** checkbox is not available for this option.
2. For the other options (radio buttons), the **Calculate by Year** checkbox is still their and is unchecked by default.

@dannyparsons if my implementation is okay, then I can ask for it to be reviewed.

Muthenya:
@dannyparsons have you been able to have a look at this?

dannyparsons:
Yes I think this sounds fine then.

Muthenya:

Muthenya:
@dannyparsons this has been slightly tested though I don't think it has been thorough. Could you please review and test it and if okay, Roger can test it further?

Muthenya:
@dannyparsons ?

Muthenya:
@Patowhiz have you been able to look at this?

instat/dlgTransformClimatic.vb
```diff
@@ -279,10 +279,10 @@ Public Class dlgTransformClimatic
         clsRRaindayMatch.bToScriptAsRString = True
 
         clsRRaindayAndOperator.SetOperation("&")
-        clsRRaindayAndOperator.AddParameter("lower", clsROperatorParameter:=clsRRaindayLowerOperator, iPosition:=0)
-        clsRRaindayLowerOperator.SetOperation("<=")
         clsRRaindayAndOperator.AddParameter("upper", clsROperatorParameter:=clsRRaindayUpperOperator, iPosition:=0)
-        clsRRaindayUpperOperator.SetOperation(">=")
+        clsRRaindayUpperOperator.SetOperation("<=")
```
Muthenya: I made changes on the operators to make them more in line with the limit they represent. Hope this is fine.

Muthenya:

instat/dlgTransformClimatic.vb
```diff
@@ -454,25 +460,29 @@ Public Class dlgTransformClimatic
                 clsRRaindayAndOperator.AddParameter("upper", clsROperatorParameter:=clsRRaindayUpperOperator, iPosition:=0)
                 clsRRaindayUpperOperator.AddParameter("max", ucrInputSpellUpper.GetText, iPosition:=1)
                 clsRCountOperator.AddParameter("x", clsROperatorParameter:=clsRRaindayUpperOperator, iPosition:=0)
```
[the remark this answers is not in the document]
Muthenya: @dannyparsons ?

Muthenya:
This is still awaiting review.

Patowhiz:
Reviewing this

instat/dlgTransformClimatic.vb
dannyparsons: They are often left out when writing R code but a general rule for dialogs is to include them.

instat/dlgTransformClimatic.vb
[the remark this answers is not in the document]
dannyparsons: I have not checked but if you are certain that it is not used anywhere in the dialog then yes remove. Make sure to do a search for all references of the class to check before you remove it.

dannyparsons:
@Patowhiz did you review this?

Patowhiz:
@Muthenya  I didn't manage to review  cause of a few problems in my machine. I've had to format it. Will review this today after re-downloading R-Instat

instat/dlgTransformClimatic.vb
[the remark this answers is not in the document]
Muthenya: I had gone through it before and I have gone through it again and haven't seen where it's being used. I have now removed it.

Muthenya:
@Patowhiz were you able to review this?

Patowhiz:
@Muthenya I can see from the code that the receivers are set to autofill, though the selector filters the variables depending on the receiver, I still have to select the variable from the selector. Can this be picked automatically?

Muthenya:
> @Muthenya I can see from the code that the receivers are set to autofill, though the selector filters the variables depending on the receiver, I still have to select the variable from the selector. Can this be picked automatically?

This should be for the Element variable?
I'm don't think having it set to autofill as it is right now is correct, but the reason why it is not autofilling has to do with its `SetClimaticType `property. The **element** can be any climatic element and thus it will be restrictive to other climatic elements if only set to one type of element.

dannyparsons:
That's correct for the element receiver. Autofilling is only when there's just one of those columns e.g. often only one date or year column, but any numeric could be an element.

instat/dlgTransformClimatic.vb
```diff
@@ -1,555 +1,704 @@
...
+
+        ucrPnlEvap.AddRadioButton(rdoEvapValue)
+        ucrPnlEvap.AddRadioButton(rdoEvapVariable)
+        ucrPnlEvap.AddParameterPresentCondition(rdoEvapValue, "evaporation.value")
+        ucrPnlEvap.AddParameterPresentCondition(rdoEvapVariable, "evaporation.value", False)
+
+        'ucrPnlTransform.AddParameterValueFunctionNamesCondition(rdoMoving, "sub1", "instat_calculation$new", False) ' clsRRainday
+        'ucrPnlTransform.AddParameterValueFunctionNamesCondition(rdoCount, "sub1", "instat_calculation$new")
+        'ucrPnlTransform.AddParameterValueFunctionNamesCondition(rdoSpell, "sub1", "instat_calculation$new")
+        'ucrPnlTransform.AddParameterValueFunctionNamesCondition(rdoWaterBalance, "sub1", "instat_calculation$new", False)
```
Patowhiz: What's the intentions of these comments?

instat/dlgTransformClimatic.vb
```diff
@@ -1,555 +1,704 @@
...
+
+        ' Setting receivers
+        ucrReceiverStation.Selector = ucrSelectorTransform
+        ucrReceiverStation.SetClimaticType("station")
```
Patowhiz: @dannyparsons don't you think we it's safer to have these climatic types as typed enumerations rather than plain strings

instat/dlgTransformClimatic.vb
```diff
@@ -1,555 +1,704 @@
...
+    Private Sub SetAssignName()
+        If rdoCumulative.Checked Then
+            CumulativeColNames()
+        ElseIf rdoCount.Checked Then
+            ucrInputColName.SetName("count")
+        ElseIf rdoMoving.Checked Then
+            MovingColNames()
+        ElseIf rdoSpell.Checked Then
+            ucrInputColName.SetName("spell")
+        Els
```
Patowhiz: There is another way of accessing the current data frame from the selector. It has a variable that store the current selected dataframe. It's safer to use that than accessing the controls inside the selector. (OOP)

Patowhiz:
This dialogue works well from my end. I like how the code is so neat and well arranged.

instat/dlgTransformClimatic.vb
dannyparsons: Yes that would be good to do, but not in this pull request.

PR #4994, opened 2018-09-28
Formmain

Wycklife:
@dannyparsons , please merge this. Adding the conversions dialog to the Climatic menu.


instat/frmMain.resx
```diff
@@ -192,34 +204,34 @@
     <value>False</value>
   </data>
   <data name="mnuDescribeThreeVariableGraph.Size" type="System.Drawing.Size, System.Drawing">
-    <value>196, 30</value>
+    <value>142, 22</value>
```
dannyparsons: There's a lot of size changes here like this one. Could you check on this? Is this a screen resolution issue? @maxwellfund

maxwellfundi: @dannyparsons  am slightly nervous about this, it has so many size changes even outside climatic menu. I would be happy with just climatic menu size changes because when you add a tool strip menu, depending with its size then it can make the other toolstrips to enlarge automatically to be proportional. Am seeing even describe and other here. @shadrackkibet  could you help @Wycklife  to get this right

shadrackkibet: Yes, We are going look into this.

PR #4998, opened 2018-09-28
Adding Conversions and General fit model dialogs

Wycklife:
@dannyparsons @maxwellfundi this is ready for you.


shadrackkibet:
I think this is okay, I have pulled this and tested on my end and it working properly.

dannyparsons:
I think possibly a previous change had resized controls and this is correcting automatically, so it looks ok.

Issue #4196, opened 2017-11-14
Adding mosaic plots


rdstern:
In extending bar charts, and generally plotting multiple factors, the way seems to be through mosaic plots.
 Another variation seems to be a double-decker plot.

These now seem to be implemented in ggplot2 through the ggmosaic package.  It basically adds geom_mosaic to the types of geom that can be used.  

There is a useful vignette 

My guess is that it will become a **Describe > Specific > Mosaic** menu item, but it needs a bit more exploration first.

It may be easy, which is why I am suggesting we at least look at it for version 0.4.9.

maxwellfundi:
The link here seems to be broken. It does not lead to a page

rdstern:
Odd.  If you look for "CRAN ggmosaic" on google, that was where you go.  But you are right.  It does not appear.

The team in Colombo, Sri Lanka have asked for a dialogue they can use to learn to start programming.  I have suggested this is a good one.  They are statisticians with computing - mainly R (and I think python) experience.  

We should also be adding the geom_mosaic to the list of geoms.  That seems potentially tricky as a starter, but perfect for Maxwell.  I am beginning to wonder if it is the right starting dialogue?

What do you think?

dannyparsons:
> There is a useful vignette https://cran.r-project.org/web/packages/ggmosaic/vignettes/ggmosaic.html

maxwellfundi:
I have looked at mosaic and I have a feeling it might be confusing and a little complicated for a beginer who might not understand how ggplot works.  

I will add the geom Mosaic on the list of geoms and maybe create the dialog too for the mosaic.

rdstern:
Yes, I wasn't sure about the dialogue, which we could discuss next week.  It might become another button on the bar chart dialogues instead.  meanwhile ig the geom can be added that would be great.

maxwellfundi:
Sounds good.

dannyparsons:
@maxwellfundi Could you or could you assign someone to add this geom now? I would like this for the workshop in the next release. I am going to think about the design of the dialog and specify that but the geom can already be added.

dannyparsons:
@rdstern I think this is a separate **Describe > Specific > Mosaic** dialog as opposed to extending the existing bar chart dialog. Although bar char is technically a subset of what `geom_mosaic` can do, there's lots of different concepts and I think keeping them separate makes sense.

maxwellfundi:
I will take on this task today.

[...]

PR #5012, opened 2018-10-07
Mosaic plots dialog

dannyparsons:

Part of #4196

Note:
- You need to install `ggmosaic` package first. This will be added to package list in next release
- `Mosaic Plot Options` currently doesn't work as the geom is not yet added to our list of geoms. It will be added as part of #4196.

### DIFF
--- a/instat/dlgMosaicPlot.Designer.vb
+++ b/instat/dlgMosaicPlot.Designer.vb
@@ -1,0 +1,205 @@
+﻿<Global.Microsoft.VisualBasic.CompilerServices.DesignerGenerated()> _
+Partial Class dlgMosaicPlot
+    Inherits System.Windows.Forms.Form
+
+    'Form overrides dispose to clean up the component list.
+    <System.Diagnostics.DebuggerNonUserCode()> _
+    Protected Overrides Sub Dispose(ByVal disposing As Boolean)
+        Try
+            If disposing AndAlso components IsNot Nothing Then
+                components.Dispose()
+            End If
+        Finally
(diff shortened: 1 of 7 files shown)


shadrackkibet:
Testing this now.

shadrackkibet:
Haven't been able to test this properly.  Using Dodoma dataset i filled up receivers as shown below. I click on mosaic plot options i get two this developer errors. When i click Ok and proceed the subdialog doesn't show any geom. If you proceed to click Ok i get a bug. Resetting the dialog after this, the dialog seems to work properly.

image

image

rdstern:
It is good you are trying that on something like the Dodoma set.  It should work, so it is good you found that problem.  The results will be boring with the data as they stand.  

In this set of data try the logical calculation for rain days and then you have another factor.  You could use the date variable to get quarters of the year and try those instead of the months.  You could also get factors (or logical) by doing the same sort of calculation on the temperature data, i.e. splitting into 2 groups and on sunshine say > 5 hrs) in the day.

You could take the years and use the recoding of a factor into Early, Middle, Late.  

This will give you lots of factors to play with.

maxwellfundi:
@shadrackkibet  thats because the layer options subdialog is not connected properly at the moment. I am working to adding the geom_mosaic on the list of geoms then that should be okay.

shadrackkibet:
Aha! i see that. Thanks

dannyparsons:
A survey type dataset with many factors would be better to test this with e.g. `efc` from R library or any of the procurement datasets, which was the motivation for getting this working now.

Issue #4951, opened 2018-09-14
Improve the Two Variables and Three Variables > Frequencies dialogues 


rdstern:
**Is your feature request related to a problem? Please describe.**
This currently puts the table only in a browser.
The graph has a useful feature, that could be added to include facets.

**Describe the solution you'd like**
For the table it now has an option to put an HTML table in the output.  It would be good to include that.  I am not sure how to do that, but @dannyparsons says it should be easy, or at least possible.  It is odd that these commands are the only ones that put the results into a browser.  Perhaps there should be further discussion on whether that option is kept, perhaps in the sub-dialogue?
For the graph, adding a tick box (at this stage) for would be useful and should be easy.  It just adds facet.grid = TRUE.  Perhaps that might be in the sub-dialogue, on the graphics tab?

**Additional context**
I have always felt that the facility to easily add counts and/or percents to the graphs is useful.  (And frustrating that there seems no control on the size of the font for these, nor can we get rid of the % sign or alter the number of decimals.)  I now find that having the coord flip, plus the option of facets can give presentable graphs - see below for the diamonds data:

![diamonds frequencies](https://user-images.githubusercontent.com/11226469/45535315-f9c35a00-b7f5-11e8-9a66-c1796ddf718c.jpeg)

dannyparsons:
These look much better than the current. I would suggest we make this the default behaviour and that is already an improvement.

Muthenya:
This is the current implementation of the facet.grid option and I've set it as the default. @dannyparsons I need your help on having the HTML table on the output window.

![image](https://user-images.githubusercontent.com/32125033/46608132-591e3c80-cb0c-11e8-97f8-4b7a280d565a.png)

Muthenya:
@rdstern just for clarification, the facet option is only for the **Count** option in the **Graph Frequencies** ?

dannyparsons:
We are still yet to get the tables to display in the output window with these functions so I don't think we can do anything other than a browser still for now.

dannyparsons:
Yes that sounds fine.

Gioche6:
@rdstern , On the Options for the two Variable frequencies, I am thinking of including more graph options like dot, boxplot and violin. They have since been added to the function. I have tested them using the existing code and they are produced well. 

I also noted the current option for line has been disabled for an error it produced. Well, a line option is only used for a line-styled histogram, for which the histogram option does not apply to grouped frequencies in this function. 


Here's where I plan on placing them
[...]

PR #5027, opened 2018-10-10
Including the facet.grid option in the Two Variable Frequencies dialog

Muthenya:
This fixes partly issue #4951.


dannyparsons:
This looks good. The facet option works well. The label for facets is not very clear, "Include Facets". I don't know if this means I specify another variable to facet by or it uses one of the existing variables. Our labels should be clear and not make the user guess what might happen. Since what it does is facet by the column factor specified I would have "Facet by Column Factor".

Muthenya:
Okay. I wasn't sure about the name. Let make that change.

Muthenya:
@dannyparsons this is ready now.

Issue #5013, opened 2018-10-08
QC Temperature dialog 


maxwellfundi:
**Is your feature request related to a problem? Please describe.**
Difference  option and Ok are enabled when we have only one element.

**Describe the solution you'd like**
Difference options should  be disabled when we got one element selected

Muthenya:
From the dialog I noticed that when the Difference checkbox is checked, okay can only be enabled when both the element receivers are filled. I believe this is the current "fix" for this. I just thought why is this (enable/disable) only proposed for the Difference checkbox and not the rest as well. It's not going to be uniform since for like the case of acceptable range checkboxes they are specific to their respective receivers.

Muthenya:
I would propose it stays as it is for now. @maxwellfundi ?

maxwellfundi:
I think if Ok is not enable is fine at the moment however it would be nice to have enhancements for these options. I dont see this as urgent @rdstern what do you think?

rdstern:
Thanks for the reminder.  In the temperature QC all the other options are possible with just a single element.  Hence, as suggested at the top of this thread, the logical (and clear, clean) solution is for the difference option to be disabled if only one temperature element is included.  

The solution above, of keeping the difference checkbox enabled, but not allowing OK remains alright for now though.  It may confuse some users, but not for long I think.  There is so much else to do, that I am happy with this temporary solution.

shadrackkibet:
@Wycklife please proceed with this task. I am available to provide support if you need.

Wycklife:
@shadrackkibet It's fine, thanks

N-thony:
@shadrackkibet  I would like to work on this issue. Could you please clarify where the issue is? Thanks.

shadrackkibet:
You only need to ensure that the difference checkbox and its corresponding NUD are disabled if one element is missing otherwise they are enabled. I also think you need to ensure that when it is disabled it is also unchecked.

PR #5042, opened 2018-10-15
  Setting the enabled property of the difference checkbox in QC temperature

Ogik99:
Fixes issue #5013 
@shadrackkibet kindly review


Muthenya:
Okay. I will have a look at it with them. Though I would also like Shadrack to have a look at it since I'm the one who took @Ogik99 through the changes.

maxwellfundi:
@Muthenya thats fine but we need to have this tested

Muthenya:
Okay. We have already tested it and we found a bug which I'll be resolving with Yvonne today.

maxwellfundi:
@Ogik99  @Muthenya  was the bug found here resolved yet?

dannyparsons:
Any update on this?

Muthenya:
Sorry for taking long to respond to this. There is still a pending discussion on it's issue, that is #5013. This will depend on the outcome of the discussion.

dannyparsons:
@Muthenya are you able to proceed now from Roger's response?

Muthenya:
>The solution above, of keeping the difference checkbox enabled, but not allowing OK remains alright for now though. It may confuse some users, but not for long I think. There is so much else to do, that I am happy with this temporary solution.

@dannyparsons this is how the dialog is at the moment. So this means that no change is needed for now.

dannyparsons:
Ok thanks, so can this be closed?

Muthenya:
Yes.

PR #5050, opened 2018-10-16
Updated RDotNet and ReoGrid packages and changed target framework to 4.5.1

dannyparsons:
@africanmathsinitiative/developers this is ready for review. Please test that everything is working as excepted.

RDotNet and ReoGrid packages updated to the latest versions. This required increasing the target .NET framework to 4.5.1 which is still compatible with Windows Vista and above. This isn't supported by XP but we were not supporting that anyway so this change doesn't lose anything.


Muthenya:
I have tested this and it's working well.

Issue #5048, opened 2018-10-16
Sorting needs to remove the previous variables from receiver.


maxwellfundi:
**Describe the bug**
When you had sorted by a certain column(s), the previous columns still remain on the receiver when you want to filter with another column. This happens when you use the right click on the data. 

**Expected behaviour**
On right clicking on data view on a column and click on sort, then the receiver on the sort dialog should remove previous selected columns and add the current ones. If multiple columns are selected then they should be added on to the receiver too.

PR #5055, opened 2018-10-17
Clearing the previously selected columns on reopen in the sort dialog.

Wycklife:
This fixes issue #5048. @africanmathsinitiative/developers this is ready for review.

### DIFF
--- a/instat/dlgSort.Designer.vb
+++ b/instat/dlgSort.Designer.vb
@@ -125,6 +125,7 @@ Partial Class dlgSort
         '
         'ucrSelectForSort
         '
+        Me.ucrSelectForSort.bDropUnusedFilterLevels = False
         Me.ucrSelectForSort.bShowHiddenColumns = False
         Me.ucrSelectForSort.bUseCurrentFilter = True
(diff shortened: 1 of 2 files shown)


dannyparsons:
@Wycklife Think also about the use of this dialog in other cases, not just from the column right click. It is also in the Prepare menu. With what you have implemented now, if I use the Sort dialog from the Prepare menu, open it and run it with some columns, then when I reopen the dialog again, it will be blank as the columns will have cleared. I hope you can see that this is inconsistent with how dialogs work in R-Instat. They should look the same when you reopen them again.

However, you are close with your solution. It just needs to be cleared only when we are filling in a default column from accessing the dialog in the right click menu.

Wycklife:
@dannyparsons Thanks for the observstion, let me go through it and make the required changes.

Wycklife:
This is ready for review @dannyparsons

Wycklife:
This is ready for review @dannyparsons

Wycklife:
This is ready for review @dannyparsons

dannyparsons:
Great, looks good. Could you check other dialogs which do similar things from the right click menu to see if any have the same problem?

PR #5059, opened 2018-10-19
Added non missing option to the summary subdialog

### LINKED ISSUE #5006: Extending the missing option in the summary statistics dialogues
The missing option is working well in the summaries now.  However it does not cope with the (common) climatic system when data does not start in January, or end in December. 
However the summaries for climatic have the useful feature that the years and months) are (almost) each the same length.
I propose a simple extension to the setting of summaries to missing.

This goes on the missing tab.  Currently the summary is set to missing if there are too many missing values on a group. 

I'd like the option added to set the summary to missing if the number of non-missing values is too low.

So, for annual summaries you can now have the result is set to missing if num_miss > 25.  The new option is identical except it is if num_nonmiss < 340.

It is a third radio button on the missing options.
The first becomes **Number Missing**.
The second becomes **Percentage Missing**
The new third one becomes Number Not Missing

If you choose it then the line reads: "Summary set to missing when there are less than this value"

Possibly when there are less observations than this value.

I am hoping this might be easy?  If not, then change the milestone!

Ogik99:
Fixes #5006


Muthenya:
@dannyparsons this required additions to the summary_functions.R file. Please have a look at it.

Muthenya:
@dannyparsons ?

dannyparsons:
What additional changes are required? I see that some have already been made.

dannyparsons:
From a quick look it looks sensible. The first check is for someone to test that it gives the results as expected.

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -427,12 +427,15 @@ summary_mode <- function(x,...) {
   else return(out)
 }
 
-na_check <- function(x, na_type = "n", na_max_n = NULL, na_max_prop = NULL, na_FUN = NULL, ...) {
+na_check <- function(x, na_type = "n", na_max_n = NULL, na_max_prop = NULL, min_non_miss = NULL, na_FUN = NULL, ...) {
```
dannyparsons: I suggest changing to keep the same format as the other parameters.
```suggestion
na_check <- function(x, na_type = "n", na_max_n = NULL, na_max_prop = NULL, na_min_n = NULL, na_FUN = NULL, ...) {
```

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -427,12 +427,15 @@ summary_mode <- function(x,...) {
...
 
-na_check <- function(x, na_type = "n", na_max_n = NULL, na_max_prop = NULL, na_FUN = NULL, ...) {
+na_check <- function(x, na_type = "n", na_max_n = NULL, na_max_prop = NULL, min_non_miss = NULL, na_FUN = NULL, ...) {
   if(na_type == "n") {
     return(summary_count_missing(x) <= na_max_n)
   }
   else if(na_type == "prop") {
     return(summary_count_missing(x)/summary_count(x) <= na_max_prop)
+  }
+   if(na_type == "n_non_miss") {
```
dannyparsons: Else If since `na_type` will only be one of these.
```suggestion
   else if(na_type == "n_non_miss") {
```

Muthenya: Okay. Thanks. We'll get right to it.

Ogik99: This is now ready for review.

Ogik99:

Patowhiz:
Reviewing this now.

instat/sdgSummaries.vb
```diff
@@ -212,6 +215,8 @@ Public Class sdgSummaries
 
         ucrNudNumber.SetParameter(New RParameter("na_max_n", 11))
 
+        ucrNudNumberNotMissing.SetParameter(New RParameter("min_non_miss", 12))
```
dannyparsons: The issue is here where the parameter name was changed in the R function but not updated here.
```suggestion
        ucrNudNumberNotMissing.SetParameter(New RParameter("na_min_n", 12))
```

This will prevent the error. But when I ran this the result was always missing even when there were more than the specified number of non missing values, so there's something that needs to be looked at here.

Muthenya:
Okay. Thank you. Then this is ready.

rdstern:
I have looked at this and it seems to work fine now, though I haven't tried all possible combinations.  I have tried with the number of missing and it all seems ok.

One small change I suggest.  It says now:

Summary set to missing when there is less than this value.

Could you change the word **value** to **number**.
And then this should also be changed with the first option

Summary set to missing if more than this value are missing.  Value here becomes number too.

Then happy to look again.  Perhaps it can now be merged?
Roger

Ogik99:
This is now ready

dannyparsons:
Was the upper limit removed?

Muthenya:
> Was the upper limit removed?

@dannyparsons the range for the "Non Missing" nud is from 1 to 340. Does this answer your question?

Muthenya:
@maxwellfundi this is another conflict caused by the summary_functions.R file. Once this is resolved then this should be ready to be merged.

dannyparsons:
@Muthenya with an upper limit of 340 it means I cannot perform a check for more than 340 missing values. If we are placing such a restriction on the user then you should be thinking very carefully if we have a good reason to have this restriction.
So the question is, "is there a good reason we are preventing users from checking for more than 340 missing values?".  If not, then we should not have the restriction.

dannyparsons:
The conflict may need to be handled manually, @maxwellfundi or me could look into this.

Muthenya:
My reasoning was that if the number of missing is > 25 then this corresponds to the number of non missing being < 340. Thus we wont need to go higher than this but now I don't think this is a good enough reason for the restriction.

dannyparsons:
That's only true for one particular type of dataset. You need to think much more generally than one example when designing a dialog.
> My reasoning was that if the number of missing is > 25 then this corresponds to the number of non missing being < 340.

Muthenya:
Yes and that's why I don't think it is a good enough reason to have the restriction.

maxwellfundi:
Moving this too... @Ogik99 will finish later

maxwellfundi:
@Muthenya  Has this been fixed yet?

Muthenya:
I need to check with @Ogik99. Yvonne have you been able to make this change?

Ogik99:
@Muthenya I have made changes from 344 to 360

Muthenya:
> @Muthenya I have made changes from 344 to 360

360 or 366?

Ogik99:
> > @Muthenya I have made changes from 344 to 360
> 
> 360 or 366?

366

Muthenya:
@maxwell once you resolve this, I can test this.

instat/sdgSummaries.vb
```diff
@@ -142,7 +142,7 @@ Public Class sdgSummaries
         ucrChkIncludeMissingOpt.AddToLinkedControls(ucrPnlMissingOptions, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=rdoNumber)
         ucrPnlMissingOptions.AddToLinkedControls({ucrNudNumber}, {rdoNumber}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=1)
         ucrPnlMissingOptions.AddToLinkedControls({ucrNudPercentage}, {rdoPercentage}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=1)
-        ucrPnlMissingOptions.AddToLinkedControls({ucrNudNumberNotMissing}, {rdoNumberNotMissing}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=340)
+        ucrPnlMissingOptions.AddToLinkedControls({ucrNudNumberNotMissing}, {rdoNumberNotMissing}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=366)
```
dannyparsons: The change was not to the default. I am happy with the default of 340 because that's the common use for this.

We were discussing that 340, or 366 should be the maximum, and there should be no maximum, because there is no reason to restrict this to a maximum number.

Muthenya: I agree with the default of 340.
 
> We were discussing that 340, or 366 should be the maximum, and there should be no maximum, because there is no reason to restrict this to a maximum number.

Then @dannyparsons I'm not sure I agree with this. Is there an instance where we would wish to check for a non-missing value greater than 366? If so, how can we allow for this without setting a maximum value for the nud yet its default maximum is 100?

dannyparsons: An important point here is that I shouldn't need to give a reason why there should not be a maximum, because its not putting a restriction on what the user can do. If you suggest that the maximum should be 366, then you do need to give a reason for this, because it restricts what we allow the user to do. So there needs to be a good reason (i.e. it causes errors or it always leads to doing something fundamentally wrong) why we want to prevent this value being more than 366. If there's no good reason then there is no harm in not having a maximum, even if there is no case we can think of where it will be used (and often users will have cases that you never think of until you see it).

However, if you want an example, take any data set with more than 366 values per group. Then this could need the value to be more than 366.
A specific example could be an agriculture experiment where I have done trials in 4 regions: 505, 503, 506 and 501 trials in each village. I want to get the average yield in each region, but I only want to get an average when there are at least 400 non missing values. Then I need to set the value to be 400, so why is the maximum 366? Because you are only thinking of climatic data, which has 365/366 values per year, but this is a general feature, not a climatic one, and we need to think of any possible data set. And there is nothing special about my 400, it could have been 4000. This would be extremely frustrating for this user to have this strange limit of 366 when the idea applies the same to any number.

Technically you can't actually have "no maximum" for an nud, because it can never be higher/lower than then largest/smallest possible integer on the machine. But practically, these are so large we consider it to have no maximum/minimum. If you look at the method for setting the min/max for the nud you can see the defaults for min/max are the smallest/largest integer values. You can just set the min and max will be the default i.e. largest integer which we consider as no maximum.

Muthenya: I agree. It's now very clear. Thanks.

Ogik99:
@africanmathsinitiative/developers this should now be ready

maxwellfundi:
@Patowhiz @Muthenya Could you please re-review this?

Muthenya:
@maxwellfundi the unresolved conflicts make me unable to merge the changes to my branch.

maxwellfundi:
@Muthenya  I suggest we (@dannyparsons ) exclude the R file then you can merge this work onto a new updated branch, then add the R edits back in and open a new PR. we will need to merge  climsoft dialog too because it has this R file too so that it doesnt cause more conflicts.

@Muthenya  below are the changes you will have to make on to the R file 

image

Muthenya:
Okay. Let me get to it.

maxwellfundi:
great @Muthenya

Muthenya:
This can now be closed.

maxwellfundi:
@Ogik99  I have closed this because @muthenya has opened a new PR on this because of conflict

PR #5063, opened 2018-10-19
Adding the missing values option to Export to CPT and other edits

### LINKED ISSUE #5018: Export to CPT Enabled bug
**Describe the bug**
- [ ] Okay is enabled even when the file path is empty
- [ ] After saving there is no message or indication that saving has been successful
- [ ] Currently the output file is a txt file
- [ ] Currently with missing values are left blank on the output file

**Expected behaviour**
- Okay should not be enabled when the save file path is empty. 
- We need a message box after saving which would give the directory of where it file has been saved
- We need the output from this dialog to be a csv file
- the missing values on the output file should be -99999

Muthenya:
Fixes #5018.
In progress.
@dannyparsons I need your assistance on how to go about the RCode.


Muthenya:
I've also noticed a bug with the radio buttons. This is caused by how the conditions have been set.

dannyparsons:
I'm going to add a parameter called `na_code` with default value "-999". You could update the dialog with this while I am implementing in the R code.

PR #5069, opened 2018-10-23
Added geom_col and added color parameter to geom_smooth

### LINKED ISSUE #4895: Add geom_col to the set of geoms
Hope this is easy?

### LINKED ISSUE #5067: Missing parameter(s?) from geom_smooth
Colour is missing as a layer parameter for `geom_smooth`. Can it be added to the geom definition? This would be a good time to check if any other parameters are missing from `geom_smooth`.

Ogik99:
fixes #5067 and fixes #4895 
@africanmathsinitiative/developers kindly review


instat/ucrGeom.vb
```diff
@@ -1476,4 +1504,8 @@ Public Class ucrGeom
     Public Function GetGeomName() As String
         Return ucrInputGeoms.GetText()
     End Function
+
+    Private Sub lblTypeofLayer_Click(sender As Object, e As EventArgs) Handles lblTypeofLayer.Click
```
dannyparsons: This seems to have been added and can be removed.

instat/ucrGeom.vb
```diff
@@ -1240,6 +1240,7 @@ Public Class ucrGeom
         'formula has to be an input and we dont have that currently. its passed in like this formula= y ~ x or  formula= y ~ poly(x, 2) or formula= y ~ log(x) so the user has to type in stuff
         'clsgeom_smooth.AddLayerParameter("formula",)
         clsgeom_smooth.AddLayerParameter("se", "boolean", "TRUE")
+        clsgeom_smooth.AddLayerParameter("colour", "colour", Chr(34) & "black" & Chr(34))
         clsgeom_smooth.AddLayerParameter("na.rm", "boolean", "FALSE")
```
maxwellfundi: @Ogik99 i think we need to add the formula parameter for the geom_smooth here.

maxwellfundi:
Will be reviewing this afternoon.

instat/ucrGeom.vb
```diff
@@ -1438,6 +1439,33 @@ Public Class ucrGeom
...
+        'Aesthetics as layer parameters... Used to fix colour, transparence, ... of the geom on that Layer.
+        clsgeom_col.AddLayerParameter("fill", "colour", Chr(34) & "white" & Chr(34))
+        clsgeom_col.AddLayerParameter("colour", "colour", Chr(34) & "black" & Chr(34))
+        clsgeom_col.AddLayerParameter("linetype", "numeric", "1", lstParameterStrings:={0, 0, 6})
+        clsgeom_col.AddLayerParameter("alpha", "numeric", "1", lstParameterStrings:={2, 0, 1}) 'Note: alpha only acts on the fill for bars. The outline is not getting transparent.
+        clsgeom_col.AddLayerParameter("size", "numeric", "0.5", lstParameterStrings:={1, 0}) ''Varies the size of outline. Note: negative size gives size 0 in general, but 'Warning: sometimesgive errors...
+
+        lstAllGeoms.Add(clsgeom_col)
+
     End Sub
```
maxwellfundi: @Ogik99 This needs to be moved to the correct alphabetical position, else it will always be the last on the list and we do not want that.

maxwellfundi:
This looks good and working well. @Ogik99 just move the geom_col code to the correct position and will be ready to be merged to master.

instat/ucrGeom.vb
Ogik99: geom_col is now correctly placed

maxwellfundi: Great, this looks good.

maxwellfundi:
@dannyparsons This looks good now!

instat/ucrGeom.vb
```diff
@@ -376,6 +376,32 @@ Public Class ucrGeom
 
         lstAllGeoms.Add(clsgeom_boxplot)
 
+        clsgeom_col.SetGeomName("geom_col")
+        'Mandatory Aesthetics
+        clsgeom_col.AddAesParameter("x", strIncludedDataTypes:=({"factor"}), bIsMandatory:=True)
+        clsgeom_col.AddAesParameter("y", strIncludedDataTypes:=({"numeric"}), bIsMandatory:=True)
```
dannyparsons: I just tried a modified example from the `geom_col` help with x as numeric and y as factor and it worked. So this doesn't look correct yet.

```
df <- data.frame(trt = c("a", "b", "c"), outcome = c(2.3, 1.9, 3.2))
ggplot(df, aes(x = outcome, y = trt)) +
    geom_col()
```

Ogik99: I have tried testing using dodoma data and for the x variable for example, both character and numeric data types are acceptable. The y variable only accepts numeric data types. This conforms to the above code.

dannyparsons: My example shows that the y variable can be a factor.

Ogik99: This is true, however it gives meaningless output.

dannyparsons: I've taken your comment as a challenge! Here's an example which I claim is meaningful. A data frame with the ages of 10 objects and their quality from low, medium, high. To compare age and quality I have produced a bar chart with age (numeric) on the x axis and quality (factor) on the y axis. This looks perfectly meaningful and sensible to me.

```
df <- data.frame(age = c(1:10), quality = factor(sample(c("low", "medium", "high"), replace = TRUE, size = 10), levels = c("low", "medium", "high")))
ggplot(df, aes(x = age, y = quality)) + geom_col() + scale_x_continuous(breaks = 1:10)
```
image

Even if we couldn't think of an example which we thought was "meaningful" or even sensible, what we are trying to do when we implement geoms from ggplot2 is to allow them to be used as flexibly as possible (like in R) but without causing problems. So, if having x as numeric meant that ggplot2 gave an error then we would prevent x being numeric. However, if we only thought that it wasn't sensible to have as numeric then we would still allow it for two reasons:
1. We want the general graphics system to be as flexible and functional as possible so we only prevent users doing something if it causes a problem (like an error) even if we don't think doing that is sensible.
2. We might think that something isn't sensible and no one should do, but we can't possibly know what the many, many different uses of this that there might be and there may be an application which is rare, but would be perfectly sensible. It would then be very annoying if this wasn't allowed by us.

So for the general system of using geoms we are making it as flexible as possible, we don't need to consider what might or might not be sensible. The basic rule is that if ggplot2 allow it then we try to allow it.

For main dialogs it might be different as our main aim is to make them easy to use, and therefore we might restrict functionality to make it easier to get something "sensible", but we would often try to keep the flexibility through the (more advanced) sub dialogs.

Ogik99: My understanding is that a factor variable can be a character string or a numeric value. In this case age is a numeric value but its actually a factor (a categorical variable). So in this case having x (age) as a numeric value is okay but in essence it is actually a factor.

Ogik99:
@dannyparsons  kindly review

instat/ucrGeom.vb
dannyparsons: These are good points to bring up and there are lots of different ideas in here that are mixed up together. Let's be clear on variable types. When you define the aesthetics for a geom you are setting the types of variables that are valid for each parameter. These are the **R** variable types. Each variable has a type. It can be factor, numeric or character (or others) but it can't be more than one of those 3 types at the same time, so you're not quite correct to say that a factor can be a character string or numeric value. A factor is a factor only.

A factor could be **created** from a character or numeric column but then it is no longer numeric or character, it's factor. R will only recognise it as factor even if it originally came from a numeric. For example, in R these two columns are exactly the same type:
```
x <- factor(c(1, 2, 3))
y <- factor(c("a", "b", "c"))
```
To R, x and y are the same type of variable: `factor`. You can check this by doing `class(x)` or `class(y)`. Even though x was made from a numeric vector and y from a character vector that is unknown to R now.

You are right that in my example I was treating x as a categorical variable because there is one bar for each distinct value of x. And there would be lots of numeric columns where treating them as categorical wouldn't make sense. However, categorical is not a variable type in R. What we've discovered is that `geom_col` will treat the x aesthetic as categories even if it isn't a factor. So if I want to use age on the x axis I don't need to convert it to factor. So in essence age it is being _treated_ as categorical for this graph but it is definitely **not** a factor.

Categorical and factor are very similar terms but in R factor is a technical term for a type of variable it stores. Any variable is either a factor or not which can be checked with `class()`. Anything can be made into a factor, even if its not sensible to do so. Whereas categorical is a statistical term. You could say that age is categorical even though its stored in R as numeric. There is no way in R to say that a variable is categorical without converting it to a factor. So when we restrict something to factors only we need to remember that this will exclude any non factor columns, even if they are thought of as "categorical", like the age.

For implementing geoms its much easier because we use the simple rule that if ggplot2 allow it, then we try to allow it. For geoms, we want maximum functionality and so we don't need to consider what is sensible and what is not (which is much harder to decide).

I'm not sure if that will clear things up or make things more confusing, but this was a good point so please keep raising these when you have thoughts.

PR #5073, opened 2018-10-24
Fixing bug when running code from Hypothesis dialog on Script Window

### LINKED ISSUE #5036: Small correction needed to Model > Hypothesis
**Describe the bug**
The code for this dialogue is not written to the script file correctly

**To Reproduce**
Open a dataset, say Samrain from the climatic datasets in the climatic guide.
Goto the Model > Hypothesis dialogue
Run a 2 sample t-test, say ttest(Strt2,Strt2,paired=TRUE)
Check it works OK.  No error in Try and produces some results in the output window.
Now go back to the dialogue
Press the Script button
Now try to run from the script window.  It gives errors.

**Expected behaviour**
You should get the same results from running the script window as from pressing OK.

**What's the problem?**
Look at the output of the R commands in the output window.  You will see that the first 2 lines, namely, in my case:
Samrain <- data_book$get_data_frame(data_name="Samrain")
attach(what=Samrain)

These lines are given before the Comment and are not part of the code copied to the script window.
It should be easy to swap the order of the code, so this is done.

Can this be given to a new programmer?  I have nominated Wycliffe, but it could be any of the new team.

Wycklife:
Fixes #5036.


Muthenya:
This looks good from my end.

instat/dlgHypothesisTestsCalculator.vb
```diff
@@ -666,4 +651,12 @@ Public Class dlgHypothesisTestsCalculator
     Private Sub ucrSaveResult_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrSaveResult.ControlContentsChanged, ucrReceiverForTestColumn.ControlContentsChanged
         TestOKEnabled()
     End Sub
+
+    Private Sub ucrSaveResult_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSaveResult.ControlValueChanged
```
dannyparsons: Is this not done automatically by the save control?

Wycklife: We thought the same but this was not the case. That's why we did it manually.

dannyparsons: If it doesn't work as expected then something is not right and you need to find out why. Doing a manual fix is not as good as we want to get to the root of the problem, without getting to the main cause it may lead to other unexpected bugs.

Patowhiz: `SetCommandString `i s being used when there are changes in the receiver, this effectively makes the base RSyntcat to be a command string and not a base function, making toScript function to NOT set assign a value. @dannyparsons I think we had discussed about this before, in one of my PRs. I don't know if we solved it but I made some changes to the underlying toScript method to include setAssign. I think He won't be able to achieve this in the normal until that change is merged

```
 Private Sub ucrReceiverForTestColumn_SelectionChanged(sender As Object, e As EventArgs) Handles ucrReceiverForTestColumn.SelectionChanged
        ucrBase.clsRsyntax.SetCommandString(ucrReceiverForTestColumn.GetVariableNames(False))
        ucrInputTryMessage.SetName("")
        cmdTry.Enabled = Not ucrReceiverForTestColumn.IsEmpty()
        TestOKEnabled()
    End Sub
```

Patowhiz: Exactly. Yes.

Muthenya:
This is working well now.

Muthenya:
This is working well now.

Wycklife:

maxwellfundi:
Looks good and save control is linked to the right thing as @dannyparsons  suggested. Merging this.

PR #5074, opened 2018-10-25
Changed geom linetype layer parameter from numbers to respective 'string names'

### LINKED ISSUE #4810: Line types in ggplot2
In our graphics system we currently have linetype (for many geoms) being a number from 0 to 6. Alternatively, you can use the names for these lines, see here: http://www.cookbook-r.com/Graphs/Shapes_and_line_types/
e.g. `ggplot(mpg, aes(class, hwy)) + geom_line(linetype = "dashed")`

I propose this is changed in our geom definitions to give the six names in an non editable combo box instead of a numeric up down with the numbers.

Ogik99:
fixes #4810 
@africanmathsinitiative/developers kindly review


dannyparsons:
@Ogik99 Thanks for looking at this. It is a good improvement to have the list of line types instead of the numbers. My concern with this is that this list is now repeated in several places which relates to the "duplicate code" and "don't repeat yourself" programming principles.

"duplicated code" can result in errors being introduced when you are trying to keep multiple similar copies up-to-date in different places whenever something changes or needs fixing e.g. to add another linetype in future would need to find all the places linetype is currently .
The solution is to restructure the code so that the repeated code only exists is one place (and then only needs to be updated in a single place in the future). In this case you could define the list of line types at the top of the file and then use this variable in each geom, instead of repeating the list in each. Then you've removed the duplicate code and avoided the repetition.

Ogik99:
Could we make similar changes to other aesthetics as well? There are also other aesthetics such as color and alpha with the same repeated code in each geom. What is your view on this @dannyparsons

dannyparsons:
Absolutely, that would be great. We should be trying to avoid repeated code like this wherever possible to make the software more robust.

maxwellfundi:
@Ogik99  This seems to have some work pending. would you like some help from anyone from @africanmathsinitiative/developers

Ogik99:
This is now ready. However, changes were not made to other aesthetics such as color and alpha because they do not have similar characteristics across all geoms.

maxwellfundi:
@Ogik99  Have checked this and looks good. @dannyparsons  am not sure how we ensure the aesthetics here also done in a similar way because some are quite different from geom to geom

dannyparsons:
If they are not the same across geoms then we have to keep them separate and try to ensure they are consistent with the geom they are for.

Muthenya:
Is this ready for review?

maxwellfundi:
@Muthenya  it would be nice if someone could check that any line that had something like `.AddLayerParameter("linetype", "numeric", "1", lstParameterStrings:={0, 0, 6})` is changed to `.AddLayerParameter("linetype", "list", Chr(34) & "blank" & Chr(34), lstParameterStrings:=strLineType)` for both commented out code and non commented out code.

Muthenya:
Okay.

Ivanluv:
line 400 has not been changed

dannyparsons:
@Ivanluv thanks for checking. When your comment refers to a particular line of code its even more useful to go to "file changes" and comment on the specific line. It just makes it a bit easier for the author to identify the line then.

Muthenya:
@dannyparsons I have also seen the lines in the code in Visual Studio but I am finding it difficult to comment on the line in GitHub.

The lines are 271 and 400.

dannyparsons:
For lines that the author hasn't changed in this pull request you can expand the sections of the code and still comment on them.

Ogik99:
The line 400 that @Ivanluv is referring to is in 'clsgeom_col' However, I don't have 'clgeom_col' on this particular branch. 
That line 400 actually exists on another branch - Geom_smooth.
@maxwellfundi could you kindly help out with this issue

Ogik99:
I have made changes to lines 271. And I no longer have this line of code on my branch .AddLayerParameter("linetype", "numeric", "1", lstParameterStrings:={0, 0, 6})

linetype

PR #5076, opened 2018-10-25
Enahncing dlgGeneralForfraphics:- Fixed reset for ucrSaveGraph and the  ucrSave graph bug when not checked

maxwellfundi:
@dannyparsons we seem to have the old save graph control here, Am not sure why we decided to keep it when we were rewriting dialogs.


dannyparsons:
@africanmathsinitiative/developers this looks ready for review.

Muthenya:
This looks good from my end. 
I've just noticed that after running the dialog for the first time, when you go back to the **"Add"** subdialog, the checkbox for "Apply on all layers" is unchecked yet it is checked by default. Is this how this is supposed to behave?
Otherwise, this is okay.

dannyparsons:
It's supposed to be checked by default for the first layer and then unchecked by default for all other layers.

Muthenya:
Okay. Thanks.

Muthenya:
Which error?

Patowhiz:
The bug that this PR is addressing. It's not attached to any issue

Muthenya:
I see:

- Go to Describe>General>Graphics
- Click on **Add** and add any layer, e.g, geom_boxplot.
- Depending on the data you are using add the aesthetics, in my case, I used Dodoma data and set the y=Rain and x=Month.
- Click return and then Ok. The error message will pop up.


Note that this doesn't happen when you check on Save Graph.

PR #5102, opened 2018-11-12
File > New Data Frame

### LINKED ISSUE #5097: Adding Try to the File > New dialogue
The new File > New Data frame is working very well.  I am delighted!

a) Could the **Try** control be added?  It is in the **Prepare > Calculate > Calculations**, and **Prepare > Generate > Enter**.  It would be good to have it, in each of our "half-way" dialogues.
b) Now wakefield is added automatically when R-Instat is loaded I can just have the command r_data(500,the_works) and I get an interesting data set with 49 columns.  This could be added as one the built-in examples perhaps.  (While we await the random option in the dialogue.
We might want to give it as wakefield::r_data(500,"the_works") just to emphasise the package, though it isn't necessary.
c) Would it make sense for some right-click operations to be included - a subset of what is in the script window?  An example I have now would benefit (in the command option) if the first line could just be run separately.  Then the remainder would be run with the last line (as usual) making the data frame.  I can do this now by running the first line from the script window. , but a bit more convenient to simply run from here.

Patowhiz:
Fixes #5097


rdstern:
Great that there is now the try.

I tried it an couldn't get anything to happen though.  I tried both with Command and Construct and with errors and ok commands.  Nothing seems to happen when I click Try whatever I do.

dannyparsons:
@Patowhiz what is the progress on this?

Patowhiz:
I'm closing this PR due to the conflicts it has with the main branch. I've opened another one that fixes the issue.

Patowhiz:
This is now addressed in PR #5207

PR #5105, opened 2018-11-13
Adding the facility to remember the previous functions and designer edits on the General Fit Model dialog

### LINKED ISSUE #5104: Making the pull-down work in the Model > General > Fit dialogue
This dialogue has the facility to remember the previous functions used.  Just like the **Prepare > Column: Calculations > Calculate** dialogue.  But it is inactive.

It would be good if it could work.  
Could it start with a default formula remembered - doesn't need to be visible, but ok if it is.  It contains simply (1).  This fits a null model.  It is sometimes a useful starting point.

At the same time it would be good if some of the keys in the keyboard (especially + and * were a bit larger.  They are larger in the keyboard in the Model > Model dialogue.

Muthenya:
Fixes #5104.


instat/dlgFitModel.vb
```diff
@@ -107,7 +107,7 @@ Public Class dlgFitModel
 
         clsFormulaOperator = clsRegressionDefaults.clsDefaultFormulaOperator.Clone
 
-        ucrReceiverExpressionFitModel.AddToReceiverAtCursorPosition("1")
+        ucrReceiverExpressionFitModel.AddToReceiverAtCursorPosition("(1)")
```
dannyparsons: I think this should just be `"1"` without brackets but let @rdstern confirm.

instat/dlgFitModel.vb
```diff
@@ -398,4 +398,8 @@ Public Class dlgFitModel
         clsVisReg.SetAssignTo("last_visreg", strTempDataframe:=ucrSelectorByDataFrameAddRemoveForFitModel.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempGraph:="last_visreg")
         clsAutoPlot.SetAssignTo("last_autoplot", strTempDataframe:=ucrSelectorByDataFrameAddRemoveForFitModel.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempGraph:="last_autoplot")
     End Sub
+
+    Private Sub ucrBase_ClickOk(sender As Object, e As EventArgs) Handles ucrBase.ClickOk
```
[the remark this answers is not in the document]
dannyparsons: Yes this is fine. This is a good use of the `ClickOK` event. We now want to avoid running any R code on `ClickOK` event because this can be handled correctly with the additional codes of RSyntax.
There are still things like this which should genuinely happen after OK is clicked in this way.

instat/dlgFitModel.vb
Muthenya: He wanted it to have the brackets.

Muthenya: Then instead this "1" can be added to the ComboBox in Initialise since it is not required in the Defaults.

rdstern:
I have tried this facility and it seems to work fine.  I suggest it can be merged.

maxwellfundi:
@rdstern great, I am merging it.

PR #5106, opened 2018-11-14
Adding Try to the Modelling Dialog

### LINKED ISSUE #5098: Adding Try to the new Model > Model dialogue
Could it also be added to this dialogue?

Muthenya:
Fixes #5098.


rdstern:
This seems to work fine.  
But I notice that the save model is not enabled on this dialogue.  Could that be added too - perhaps before it is merged?

rdstern:
I also think that the facility to look at the previous models is not available.  This was the feature recently added to the Model > General > Model dialogue.  Could it be added here too please?  Happy to check once it is done.

maxwellfundi:
@Muthenya will add the desired features here.

Muthenya:
> I also think that the facility to look at the previous models is not available. This was the feature recently added to the Model > General > Model dialogue. Could it be added here too please? Happy to check once it is done.

@rdstern do we need a default for this as well?

rdstern:
Good question.  Also it made me see what is there now.  Danny has made the default to be lm( ) and that is fine.
However, when I change the keyboard to Extremes and then press reset it currently gives me an error, because it tries to put lm( ) but that isn't a possibility with that keyboard.

So either a) Have no default
or b) When press reset go back to the stats keyboard and use lm( ) 
or c) When press reset have a sensible default for each keyboard.

You could also check with @dannyparsons .  However I notice that the Prepare > Calculations > Calculate does go back to the default keyboard.  So perhaps you could look at option b) and then keep lm( ) as the default.  That will mean you need to edit the action on reset.

Muthenya:
@rdstern I have not been able to produce an error. I have made the changes. Please have a look at it again.

@africanmathsinitiative/developers this can be reviewed again.

Muthenya:
@rdstern I have gone through the dialog again and tried doing the same model as you but all my results are consistent with your second finding, that is,

> **# Code generated by the dialog, Modelling survey <- data_book$get_data_frame(data_name="survey") attach(what=survey) model2 <- lm(Yield~Variety.+Fert) data_book$add_model(model_name="model2", model=model2, data_name="survey") data_book$get_models(data_name="survey", model_name="model2")**
> (Here it didn't give an error, but now told me the model it was going to fit - it didn't print the data frame)
> **summary(object=model2)**
> (Now it gave the results as before)
> **detach(name=survey, unload=TRUE) rm(list=c("model2", "survey"))**

And this happens whether or not I have saved the model.

I'll continue to test this and also check this with @dannyparsons.

rdstern:
Interesting.  In that case I suggest you check with @dannyparsons , but then get it merged and we can then continue to check.

Muthenya:
@rdstern this is now ready. Please have look at it again.

dannyparsons:
> Good question. Also it made me see what is there now. Danny has made the default to be lm( ) and that is fine.
> However, when I change the keyboard to Extremes and then press reset it currently gives me an error, because it tries to put lm( ) but that isn't a possibility with that keyboard.
> 
> So either a) Have no default
> or b) When press reset go back to the stats keyboard and use lm( )
> or c) When press reset have a sensible default for each keyboard.
> 
> You could also check with @dannyparsons . However I notice that the Prepare > Calculations > Calculate does go back to the default keyboard. So perhaps you could look at option b) and then keep lm( ) as the default. That will mean you need to edit the action on reset.

I agree with implementing b) here.

dannyparsons:
I think Roger's error that has not been reproduced is related to options on the display sub dialog.

I haven't checked this but it could be that when you change the model without reopening the subdialog that some options there don't update correctly.

@rdstern does this sound possible? If so @Muthenya try investigating this first.

rdstern:
Yes, that does sound possible.  Perhaps either that could be checked quickly in the code, or I'll go through the tests again once the whole thing is merged.  It is clearly useful now anyway.

And I should have thought of that possible reason too.  The display sub-dialogue is used  by many dialogues.

rdstern:
Yes, I'll do that today.

rdstern:
It seems ok to me.  The history is remembered, and try works also.  I have tried saving and changing the display options and can't generate an error.

The one very minor change I suggest is in Try.  Currently when the model is ok it says "Model is working properly".  That's quite a strong sentence.  I suggest instead "Model seems OK".

In passing I notice that OK seems recommended compared with ok and Ok (which we have as the OK in the control in all dialogues!)  Should that be changed?

dannyparsons:
I would suggest the even weaker: "Model runs without error" since this is what it is actually checking.

rdstern:
Fine.  And then I suggest it can be merged.

Muthenya:
@dannyparsons this is now ready.

PR #5108, opened 2018-11-15
New Visualize Data dialog using Visdat package

### LINKED ISSUE #5064: Dialogue for the visdat package
I propose this as a new dialogue.  It seems very simple as there are few options - but extremely useful.

The precise layout of the dialogue needs to be discussed.  I suggest there could be a single dialogue for the 3 functions visdat::vis_dat, vis_miss and vis_guess

We can choose between a whole data frame and selected columns in each case (I hope).  And it would be possible to filter also.  In each case it produces a ggplot.

It would go into Prepare > Check Data > Visualise Data.  It could be at the top.  It could also be at the top of the Climatic > Tidy and Examine menu.> Visualise Data

There could then be the three radio buttons Data, Missing and Guess.
Then a checkbox, default checked Whole Data Frame.

If checked, then it just gives the data frame control.  Otherwise it gives our usual control with the columns.  There is also a multiple receiver.  No receiver is needed if the whole data frame.

Then there is a different Options box for each of the three cases.  There aren't many options, so they can all be there.

Then the save graph and that's it!

The code should be simple - I was able to try it all using the Script window - for whole data frames that is.
I suggest this could be a nice small project for one of the new programmers.  Wait though, till Danny has commented on the idea and the structure of the dialogue.

Ogik99:
Fixes #5064 
@dannyparsons and @rdstern  this is the initial dialog with three functions : vid_dat, vis_miss and vis_guess

initial implementation


dannyparsons:
The design looks good to me. I suggest someone in the team tests the functionality.

maxwellfundi:
@Ogik99  just a few issues on this dialog. When the dialog loads for the first time, none of the two option2- **Whole Dataframe, Selected Columns-** is checked and the layout is for the Selected Columns option. Still on the same issue, even after selecting the columns, then **OK** is not enabled. see below 

image


On clicking the **Missing** option while on the whole dataframe option then one argument - `x`- is not passed in to the function and leads to the following error. 

image


Do these different options need different datatypes? If so can they be restricted?

Ogik99:
@maxwellfundi  kindly review this. Some changes have been made.

Ogik99:
@Wycklife  kindly test this.

Muthenya:
This is working well in general. Just two things I have noted:
1. The Missing option always gives an error message, shown below;

visdat

2. The Guess option graph only displays in the R Viewer.

dannyparsons:
@Muthenya 1. is a memory allocation issue in R. Can you try this with a smaller data set? Can you also try this directly in R with the same data set?

dannyparsons:
2. Should be fixed by the call type.

Muthenya:
1. @dannyparsons I have tested both with a smaller data set and in R and in both I haven't encountered any problem. 

2. The iCallType was already set to 3.

Muthenya:
I have also noticed that the Guess option produces the graph in the output window when you use another option, either Data or Missing, before it.

dannyparsons:
Are you still getting the same problem consistently in R-Instat?

Muthenya:
> Are you still getting the same problem consistently in R-Instat?
I haven't had the same issues in R.

And I'm now getting this error message

image

dannyparsons:
Can you check with @Ogik99? Is this just for this dataset or any?

Ogik99:
I have used the Survey data set but did not getting any errors. This is the output for the Guess option
guess option

Ogik99:
@Muthenya the Guess option worked well for me with the Dodoma data set
dodoma guess

dannyparsons:
This might be a general memory issue in R for machines with lower memory, if this function uses a lot of memory for some reason.

It would be good if others could also check this on different machines.

Muthenya:
@Wycklife and @Ivanluv are testing this as well.

Ivanluv:
@dannyparsons am gettinng the same memory error

dannyparsons:
@Ivanluv only with Dodoma data? Can you try different sizes so we work out if its the size of the data that causes it.

Ivanluv:
@dannyparsons yes,its only the Dodoma Data thats throwing the error .The other Data frames are working fine.I  think large data frames  cause the error.

dannyparsons:
Ok that's what I expected. That's a shame because then its an issue with the R function itself.

Could you confirm with another large dataset? I guess the problem is with very long dataset i.e. many rows rather than very wide datasets i.e. many columns.

shadrackkibet:
@Ogik99 what is the status of this?

Ogik99:
> Ok that's what I expected. That's a shame because then its an issue with the R function itself.
> 
> Could you confirm with another large dataset? I'm cautious if its only been tested properly with Dodoma. I guess the problem is with very long dataset i.e. many rows rather than very wide datasets i.e. many columns.

This package generally has issues with large data sets. Its advised that users downsample the data.

rdstern:
I was hoping to try this feature on my machine - which has quite a lot of memory. But it says:

**_Cannot initialise R connection.
Object reference not set to an instance of an object.
Could not establish connection to R.
R-Instat requires version 3.4.4 of R.
Note that R-Instat does not currently work with R 3.5.0 or above. Try reruning the installation to install R 3.4.4 or download R 3.4.4 from https://cran.r-project.org/bin/windows/base/old/3.4.4/ and restart R-Instat.
Object reference not set to an instance of an object.
OK_**   

Can the old pull requests be updated as we have now switched, so we use R Version 3.6 and can't use Version 3.44 any more.

Ogik99:
> > Ok that's what I expected. That's a shame because then its an issue with the R function itself.
> > Could you confirm with another large dataset? I'm cautious if its only been tested properly with Dodoma. I guess the problem is with very long dataset i.e. many rows rather than very wide datasets i.e. many columns.
> 
> `Some datasets might be too large to plot, sometimes creating a blank plot - if this happens, I would recommend downsampling the data, either looking at the first 1,000 rows or by taking a random sample. This means that you won't get the same "look" at the data, but it is better than a blank plot! See example code for suggestions on doing this.` - This is from the R- documentation for this package.

We may need to give users an option to downsample the data (incase of a large dataset).

maxwellfundi:
This looks like the issue is the huge dataset as described in the documentation. i suggest @Ogik99  you try updating this PR with the latest Master so that @rdstern  can be able to test it.

Ogik99:
> This looks like the issue is the huge dataset as described in the documentation. i suggest @Ogik99 you try updating this PR with the latest Master so that @rdstern can be able to test it.

This PR is now updated.

rdstern:
@Ogik99 I am probably doing something wrong.  I used your updated code - I think - and am nicely in R-Instat.  But I expected to see a new dialogue somewhere and couldn't find it?  I assume it should be under Describe > Specific, but also looked elsewhere.

Is there something else I should do first?

Ogik99:
@rdstern sorry for that error. The dialog had not been added to any of the menus. I have added this to the describe menu as "Visualize data".
However we are currently working on an issue---of the three functions only the visdat function gives the desired output. The other two give no output. We are trying understand this issue and check whether these two functions work with the update R version

Ogik99:
` library(visdat)
Warning message:
package ‘visdat’ was built under R version 3.6.1`  
- warning message when you load the visdat library in R

rdstern:
The dialogue looks ok.  For consistency with the Prepare > Check Data > Duplicates dialogue call the 2 options:
**_Data Frame
Selected Variables_** 

I tried the dialogue with the new Dodoma data (30,000 rows by about 10 variables) together with the annual summaries (85 rows by 15 variables)

On the summaries it all seems to work and I get graphs for each option.  The odd result was the missing option, where I have missing values at the top of the data, but they come at the bottom in the missing graph.
image

In the missing documentation for visdat it says:
"Note
Some datasets might be too large to plot, sometimes creating a blank plot - if this happens, I would
recommend downsampling the data, either looking at the first 1,000 rows or by taking a random
sample. This means that you won’t get the same "look" at the data, but it is better than a blank plot!
See example code for suggestions on doing this."
 
So the blank plots might just be the size of the data.

I tried with the daily data and the first option works.  With the missing option it stated that it needed 3.5gbytes and that was too much.  I made a subset (I don't know whether the filter works with this?) and it still complained that 716 mbytes was too much.  I deleted variables and then tried with just 2 variables and it still wanted 716 Mbytes of data.  So the memory needed seems to relate purely to variable length?  Is this true.  Should we discuss with them?

I tried a bit with the library Rockart data (DAAG).  That has 103 rows and 641 variables!  That works fine (at least for missing) except you can't read any of the names of course!

dannyparsons:
Thanks for finding this @Ogik99. It is a shame there is no netter solution to this. It's going to be more of an issue for some of our users with low computer memory. Taking less rows or a random sample is not what we want to do for climatic data, for example.

I do think its worth looking at the other options out there in different packages to see if there's somewhere that has addressed this problem, if so it could replace visdat or we have options for doing the plot. If there's no other package that looks useful, we could add an option of choosing the first and last row as a subset. Random sample could also be done, although not sure that's as useful for us?

> `Some datasets might be too large to plot, sometimes creating a blank plot - if this happens, I would recommend downsampling the data, either looking at the first 1,000 rows or by taking a random sample. This means that you won't get the same "look" at the data, but it is better than a blank plot! See example code for suggestions on doing this.` - This is from the R- documentation for this package.

We may need to give users an option to downsample the data (incase of a large dataset).

dannyparsons:
Also, @Ogik99 I know you added the dialog to frmMain so that Roger could view it, but we want to keep `frmMain` changes done on separate pull requests. Now we have some conflicts here.
We can find alternative ways to helping Roger test new dialogs.

shadrackkibet:
I see conflicts on the frmmain here, can this be sorted out? Otherwise, these changes can be transferred to a new branch for testing.

shadrackkibet:
@rdstern I think this links with the topic on visualizing missing values by @HawardKetoyoMsatsi and @penina-AI. Might be sensible to look into this and continue with it.

lloyddewit:
@lilyclements Please could you try and complete this PR? @rdstern suggested this.

lilyclements:
Just to summarise, to check I'm understanding correctly:

1. The R-code won't run for large data sets. The largest data set I have tried is the dodoma data, where I am only having a problem with this for `vis_miss`.
I will have a look to see if there are alternative packages, or another way around handling large data sets.
2. The plots sometimes need some clicking between "data frame" and "selected variables" to display the graphic, rather than an empty white box (i.e., to run the `$add_graph` and `$get_graphs` lines)

Also - sidenote of issues I came across:

3. The TestOK can currently be clicked when the ucrSave is incomplete.
4. Problems in the `ucrSave`:
Under `vis_miss` and `vis_dat` if I check the save, click OK, and reopen, the save is no longer checked. If I then check the save, the usual default name has gone and is now "last_graph".
This does not occur for `vis_guess`, however, `vis_guess` can get the wrong "default name" upon reopening.
This seems to be fixed by changing the prefix from `graph` to `vis_dat` on line 69:
`ucrSaveGraph.SetPrefix("vis_dat")`

I moved the code relating to this dialog into a new branch to be able to check/run the code without conflicts.
Here I have fixed the minor issues I put on bullet points 3 and 4.
There may be minor design changes too.

@dannyparsons shall I make a new pull request with these fixes on? Otherwise, these are only minor fixes so I can try to merge into this branch (only concerned with the current conflicts).

lloyddewit:
@lilyclements Thank you for digging up and investigating this historical PR! Regarding the conflicts, I suggest that:
- You make a new PR (branched from current master) that only contains the `frmMain` changes, you may also need to add a stub for `dlgVisualizeData.ShowDialog()`
- I approve and merge this PR promptly, this will reduce the risk of future merge conflicts
- You make a second PR that contains the `dlgVisualizeData` changes. This should not create any merge conflicts so we'll have time for a proper review and test cycle.

Hope this helps

lilyclements:
@lloyddewit I have done bullet points one and three:
For 1 (frmMain changes) - see #5897 
For 3 (dlgVisualiseData changes) - see #5898

Should this be closed?

lloyddewit:
@lilyclements Thank you for moving this to #5897 and #5898. This PR is now OK to close.

PR #5114, opened 2018-11-19
start of rains missing values fixes

dannyparsons:


dannyparsons:
corrected missing value problems in start of rains and spells dialog

PR #5135, opened 2019-01-25
Added keys to the MASS and stat key boards

### LINKED ISSUE #5125: Adding to the Model dialogue - 1 - existing keyboards
From looking at the Prediction documentation I suggest our new Model > Model dialogue be extended by a) Adding missing keys to the existing keyboards
b) Adding new keyboards
Here is propose the first.  We may then have further issues for each new keyboard or a larger issue for a set of new keyboards.

There are various keys missing from the stats and the MASS keyboards.  They are as follows:
1. From stats
Add nls
Add ppr
Add princomp
2. From MASS
Add lda
Add lqs
Add mca
Add qda

I assume this may be quite easy?  So I wonder who could do this quickly?

Ogik99:
fixes #5125 


Wycklife:
@dannyparsons I have reviewed this, all the required keys have been added on the expected two keyboards.

PR #5137, opened 2019-01-25
Added cumulative and exceedance radio buttons

### LINKED ISSUE #5057: Improving the exceedance graphs
I needed to do these for PICSA in Tanzania.  They are important and can look good in R (and hence R-Instat).  The dialogue is in Describe > Specific > Cumulative Plots and also in Climatic > Seasonal Forecast Support > Cumulative/Exceedance Graphs.  Once it is working a bit better, then  in Climatic I suggest we move it (or possibly add it also to the Climatic > PICSA menu..  

I have suggested Maxwell and Danny, not because they will fix the problems.  But it is a good time to consider who could be looking at the plotting with this possibly as a first example.  This will need supervision by Maxwell, and some edits, mentioned below, could be easy to do.

Here are 2 examples:
exceedence plot - seasonal totals

exceedence plot - lengths

They are the opposite of what statisticians call cumulative frequency graphs.  The Exceedance graphs go down from 1 to zero and the probability (left hand side) shows the chance of more than the given x value.  

It is possible that the improvements needed are general, rather than just this dialogue.

a) When I change the y-axis options to Continuous I find it is initially set to **discrete**.  (I assume this setting is somewhere in the code for this dialogue, rather than general?  It is an interesting issue, because the y variable is sometimes an integer column (like length in complete days), but the y-scale in the graph is from 0 to 1, so is always continuous. 
b)The default on the y-variable is to give me labels at 0, 0.25, etc and the request (in PICSA is for more labels to help read off the corresponding values more easily.  I change it to numeric and make the necessary edits.  It ignores me and when I return it is reset to discrete again. (I had to edit the script file to get the 0.1 spacing in the figures above.  The problem is clear from the code below.
c) It is because the R code seems contorted and repetitive.  I give an example below.  It shows the settings of 0.1 were accepted, and then overwritten by later code.
d) The Line options in the dialogue are disabled.  That stops me being able to change the colours, thickness etc.
e) Curiously the main options - please call it **Plot Options** (rather than just **Options**) were enabled.  I say curiously, because I was using multiple Y variables (to get the 2 lines in the examples above) and on other plots, the sub-dialogues are disabled.  I like it enabled!
f) We still have the general problem that when we add the points - which we can do, from the main dialogue - and I like that - we can't change the options on the points layer, because they are from the main dialogue.  This is a general problem that is reported elsewhere and has been an issue for a long time.
g) What I am not sure if it is easy is to change the y scale.  There is the option in the dialogue - which is good, and that is to have counts, rather than proportions.  Is it easy to also 1) Have percentages, and 2) Be able to multiple by 10 rather than 100 - so it gives the risks in 10 years, rather than in 100!

Here is an example of the current code from the command:

**Code generated by the dialog, Cumulative/Exceedance Graphs**

merge2 <- data_book$get_data_frame(data_name="merge2", stack_data=TRUE, id.vars=NULL, measure.vars=c("sum_Dodoma","sum_Kondoa"))
last_graph <- ggplot2::ggplot(data=merge2, mapping=ggplot2::aes(x=value, colour=variable)) + ggplot2::stat_ecdf() + theme_grey() + ggplot2::geom_point(stat="ecdf") + **ggplot2::scale_y_continuous(breaks=seq(from=0, to=1, by=0.1))** + **ggplot2::scale_y_continuous(breaks=seq(by=0.1, to=1, from=0))** + ggplot2::ylab(label="") + ggplot2::**scale_y_reverse(breaks=seq(1,0,-0.25), labels = seq(0,1,0.25))** + ggplot2::xlab(label="Seasonal Total (mm)") + ggplot2::scale_x_continuous(breaks=seq(by=50, to=1200, from=100), limits=c(), expand=c(0.0, 0))
data_book$add_graph(data_name="merge2", graph=last_graph, graph_name="last_graph")
data_book$get_graphs(graph_name="last_graph", data_name="merge2")
rm(list=c("merge2", "last_graph"))

default exceedance graphs

Ogik99:
fixes #5057 
@africanmathsinitiative/developers kindly review.


instat/dlgCumulativeDistribution.vb
```diff
@@ -179,4 +193,19 @@ Public Class dlgCumulativeDistribution
     Private Sub CoreControls_ControlContentsChanged() Handles ucrVariablesAsFactorforCumDist.ControlContentsChanged, ucrSaveCumDist.ControlContentsChanged
         TestOkEnabled()
     End Sub
+
+    Private Sub ucrPnlOption_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlOption.ControlValueChanged
+
+        If rdoExceedance.Checked Then
+            clsBaseOperator.RemoveParameterByName("Yscales")
+            clsScaleYReverseFunc.SetPackageName("ggplot2")
+            clsScaleYReverseFunc.SetRCommand("scale_y_reverse")
```
Muthenya: @Ogik99 the `SetPackageName` and `SetRCommand` are done in the setdefaults so they are not needed here. Is there a reason why you put them here?

Muthenya: @Ogik99 were you able to look into this?

maxwellfundi:
@Ogik99 there has been a little more conversation on how to improve this dialog. Do you have sometime to have a look at these issues and fix these issues that are left on here?

rdstern:
@Ogik99 you were able to update the visdat pull request to the latest version of R.  Is it possible to do the same with this?

maxwellfundi:
@Ogik99  I see you have conflicts here but all of them are on frmMain. i am going to solve this, and keep all the changes that are coming from master. 

i think you do the small bit of frm main later.

maxwellfundi:
@Ogik99  please try update this now

Ogik99:
This is now updated.

rdstern:
@Ogik99 thank you for updating this pull request.  I have looked at this together with #5057.  There you changed the design.  I hope you kept the new design, because I don't understand now why I didn't accept it!
That is the one with the new radio buttons.
The one I checked now seems to be the same as the original.  It says it can't run but then lets me continue.  
Can you return to the dialogue in #5057?  
image

If so, then the **_From_** value can stay, but it is fixed as 0.  It would still be good to show this.
The **_To_** value is currently always 1, but a drop down would be good with 1, 10 100
The **_In Steps of_** would then change to be steps like 0.1 or 0.25 (0.25 is the default.  The only sensible values are probably 0.05, 0.1, 0.2, 0.25.  
If To is changed to 10, or 100 then these values multiply appropriately.

Also the probabilities plotted need to multiply appropriately.

I sense a bit of nifty coding here, where Maxwell or Patrick may be able to help.

The sub-dialogue for Plot options needs to be implemented.  There is no need anymore to implement the Line Options button, because that can be done from the main plot options.

maxwellfundi:
@rdstern  great, we will remove the line options and get the **Plot Options** working properly.

dannyparsons:
There are some difficult issues here which need some thought.
As a first step @Ogik99 can you resolve the frmMain conflicts here, possibly through a new pull request? Then I can have a look at some of the latest points from Roger.
frmMain changes need to be kept in separate pull requests to avoid conflicts like this.

maxwellfundi:
@Ogik99  I believe we can actually close this now.

PR #5140, opened 2019-01-30
Adding Wakefield keyboard to the calculator

Wycklife:
This is still in progress.


Wycklife:
@rdstern, I am adding the arguments for each variable, in the case where we have the arguments defined with values, should I ignore the values, for example; children(n, x = 0:10, prob = c(0.25, 0.25, 0.15, 0.15, 0.1, 0.02, 0.02,
  0.02, 0.02, 0.01, 0.01), name = "Children")

dannyparsons:
If these are the defaults then keep them in. This is what we are doing for other functions so we need to keep consistency.

Wycklife:
@maxwellfundi help me resolve this, then it will be ready for review.

Muthenya:
@maxwellfundi kindly resolve this then I review.

dannyparsons:
I've resolved the conflicts here.

Muthenya:
@Wycklife I have come across this problem with the size of the dialog (the base buttons appear to be "cut").

image

Some resizing will do.

Muthenya:
I have also observed a number of buttons whose names are not fully seen. Namely:
1. `internet_browser`
2. `wakefield_month`
3. `wakefield_upper`
4. `wakefield_year`
5. `sex_inclusive`

The buttons can be widened a bit to accommodate the longer names.

dannyparsons:
Wider buttons or smaller font could be done. We should keep buttons the same size if possible.

rdstern:
If it gets too difficult to fit, then we should consider Wakefield1 and Wakefield2, i.e. 2 keyboards.

Muthenya:
@Wycklife what's the current state of this?

Wycklife:
@dannyparsons We managed to make the changes together with Lazarus on his machine. When I pulled the branch containing the changes, I could not see the changes we made, what could be the problem and how do I resolve this?

dannyparsons:
Do you know the name of the commits you made with Lazarus? They should appear in the list of commits on the pull request. You could first check that they appear there.

maxwellfundi:
@rdstern @dannyparsons This is now finished. It works well. We would like you to test it to see if the outputs are what is expected.

Wycklife:
@africanmathsinitiative/developers I have managed to make the changes on Wakefield Calculator, and this is how it looks like
calculations

dannyparsons:
The design looks good, but there are 414 file changes. Can you check why?

maxwellfundi:
@dannyparsons  I have checked this and it appears that the files changes are many because he kept updating his branch and so it has file changes from other commits from other people.

dannyparsons:
Yes something like this has happened because it is removing all of the recent changes that have been made.
In general there shouldn't be a need to update a branch you are working on, because the work is usually independent of recent changes, and it also can lead to issues like this.

I will close this so it doesn't get merged accidentally. I suggest moving the changes to a new branch.

PR #5142, opened 2019-02-05
French controls (made designer changes to the describe menu)

### LINKED ISSUE #5096: Improving the controls for the French (and other) translations
The French translation has moved on a bit in Version 0.4.26.
There are some basic issues still to be resolved as follows:
a) When French (or any other language than English) the custom control at the bottom of the dialogue add a Lan button.  This is very nice and switches the language within that dialogue.  But it is sometimes in an odd (overlapping) place in some dialogues.  @dannyparsons will investigate.
b) In the main custom control to select columns the Data Options button becomes Options des (Donnees) and the Donnees is not there.  That button will be made bigger.
c) In dialogues with check boxes the translations do not work yet.
d) We have many Francophone people here and will be working on translating some of the text.

The main additional improvement that is needed is to go through the dialogues to see where the controls are too short in French so the text is incomplete.  Then to adjust and report on those dialogues.  For example Prepare > Data Frame > View Data the final option in English is Display HTML in Output Window and there is not enough space for the French equivalent.  

When making corrections to the size of the controls look also at the English phrase and consider whether that could be simplified.

It needs someone to go through the dialogues in turn.

Ogik99:
fixes #5096 


dannyparsons:
@Ogik99 there are no changes to the boxplot dialog here, so I assume I can go ahead and make changes to the dialog before you do layout changes?

PR #5144, opened 2019-02-05
Adding common options to sdgPlots

maxwellfundi:
This is in progress, it has many files changes which some will need to be removed.


maxwellfundi:
@dannyparsons  This works well!

PR #5177, opened 2019-02-13
Climatic summaries

### LINKED ISSUE #5035: Add date option to the Climatic > Prepare > Climatic Summaries dialogue
There are 3 options (radio buttons at the top) for this dialogue.  This option is only needed for the second button, where the summary is by both year and within year, e.g. month, i.e. Annual + Within.

A checkbox in this situation labelled **Add Date Column**.
It already adds a Year column, and a Month (say) column.  It combines them to add a date column, which could be called Date.  This is the same as using the Year and Month columns and then going to the Use Date dialogue.  It will make it easier to immediately do time series plots of the resulting data.

For now - at least - the control can take the place of the disabled Omit Unused Levels for this second option.

Wycklife:
This relates to #5035

### DIFF
--- a/instat/dlgClimaticSummary.vb
+++ b/instat/dlgClimaticSummary.vb
@@ -128,11 +128,6 @@ Public Class dlgClimaticSummary
 
         ucrChkAddDateColumn.SetText("Add Date Column")
 
-        'ucrChkOmitMissingValues.SetParameter(New RParameter("na.rm", 6))
-        'ucrChkOmitMissingValues.SetText("Omit Missing Values")
-        'ucrChkOmitMissingValues.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
-        'ucrChkOmitMissingValues.SetRDefault("FALSE")
-
         'linking controls
         ucrPnlAnnualWithin.AddToLinkedControls({ucrReceiverYear}, {rdoAnnual, rdoAnnualWithinYear}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
         ucrPnlAnnualWithin.AddToLinkedControls({ucrReceiverWithinYear}, {rdoAnnualWithinYear, rdoWithinYear}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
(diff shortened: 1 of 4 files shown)


instat/dlgClimaticSummary.vb
```diff
@@ -1,307 +1,309 @@
...
+        ucrChkDropUnusedLevels.SetText("Drop Unused Levels")
+        ucrChkDropUnusedLevels.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
+        ucrChkDropUnusedLevels.SetRDefault("FALSE")
+
+        ucrChkAddDateColumn.SetText("Add Date Column")
+
+        'ucrChkOmitMissingValues.SetParameter(New RParameter("na.rm", 6))
+        'ucrChkOmitMissingValues.SetText("Omit Missing Values")
+        'ucrChkOmitMissingValues.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
+        'ucrChkOmitMissingValues.SetRDefault("FALSE")
```
dannyparsons: why are these now commented out?

dannyparsons:
@Wycklife I also edited your original comment:
1. We only want to use `fixes` when it completely resolves the issue, because fixes will close the issue. Since this pull request adds the checkbox but not the R code, the issue won't be resolved so we don't want it to close when this is merged. It's good to still link to the issue so I use something like `relates to ...`
2. If you are requesting a review of your pull request then you need to write "this is ready for review" otherwise it is assumed you are still working on it. (I assume you would like this reviewed now, but if not please edit the comment.) Also, make sure to tag `africanmathsinitiative/developers` not `africanmathsinitiative` because only developers should get these notifications.

instat/dlgClimaticSummary.vb
Muthenya: These were for the "Omit Missing Values" check box which was removed. These should be removed as well @Wycklife

Muthenya: @Wycklife these were supposed to be removed.

Muthenya:
The checkbox has now been disabled. Once the above mentioned commented lines are removed this will be ready for merging.

Wycklife:
@Muthenya I have resolve this;  removed omit missing value

Wycklife:
@dannyparsons This is ready for you

PR #5178, opened 2019-02-15
Adding Tidy to the Open CDF and Import Shapefiles dialog and ensuring the names are consistent.

### LINKED ISSUE #4896: Emphasise tidy data in more dialogues
In the File > Open NetCDF menu item an important feature is that the facility also puts the data into a (tidy) data frame.  Could we please emphasise this in the dialogue by changing the title to mention the word tidy.

So Open NetCDF and tidy would be possible or maybe better "Open and Tidy NetCDF file".

The importing of shape files is the same.  In Climatic > > File > Import Shape Files, once in the dialogue I propose we say "Import and Tidy Shape Files"

A further small point.  Is it **Shape Files** or **Shapefiles**?  Currently the menu uses one and the dialogue uses the other.  Please can we be consistent.

Wycklife:
This fixes #4896 .
@africanmathsinitiative/developers. This is ready for review.


Wycklife:
@rdstern @dannyparsons  Is there any dialog i can do the same to?

dannyparsons:
I think that's it for now thanks, unless you've found others with similar inconsistencies.

Wycklife:
Not really, only that there are inconsistency of menu naming for some dialogs. What do we do about this

dannyparsons:
Can you put this down in an issue and then we can discuss how to resolve it.

Patowhiz:
Looks good from my end

PR #5183, opened 2019-02-16
Import multiple Excel sheets

dannyparsons:
@africanmathsinitiative/developers this is ready for review. As well as the new feature I'm particularly concerned that it has not had effected the current import features.

### DIFF
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -437,6 +437,7 @@ Public Class dlgImportDataset
 
             If dlgOpen.ShowDialog() = DialogResult.OK Then
                 ucrSaveFile.Reset()
+                'TODO This is in place for when we allow multiple files selected.
                 bMultiFiles = (dlgOpen.FileNames.Count > 1)
                 If NumberOfFileTypes(dlgOpen.FileNames) > 1 Then
                     MsgBox("All files must be of the same type", MsgBoxStyle.Information, "Multiple file types")
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -22,6 +22,11 @@ Public Class dlgImportDataset
 
     Private clsImportFixedWidthText, clsImportCSV, clsImportDAT, clsImportRDS, clsReadRDS, clsImportExcel, clsImport As RFunction
     Private clsGetExcelSheetNames As RFunction
+    ' Functions for multi file import
+    Private clsLapply As RFunction
+    Private clsFileList As RFunction
+    ' Functions for multi Excel sheet impoty
+    Private clsImportExcelMulti As RFunction
     Private bFirstLoad As Boolean
     Public bFromLibrary As Boolean
     Private strLibraryPath As String
@@ -36,6 +41,7 @@ Public Class dlgImportDataset
     Public strFileToOpenOn As String
     Private bDialogLoaded As Boolean
     Private iDataFrameCount As Integer
+    Private bMultiFiles As Boolean
 
     Public Sub New()
         ' This call is required by the designer.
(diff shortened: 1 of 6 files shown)


instat/dlgImportDataset.vb
```diff
@@ -379,7 +408,15 @@ Public Class dlgImportDataset
 
     Private Sub TestOkEnabled()
         If (ucrSaveFile.IsComplete OrElse strFileType = "RDS") AndAlso bCanImport Then
-            ucrBase.OKEnabled(True)
+            If strFileType = "XLSX" OrElse strFileType = "XLS" Then
+                If lstSelectedExcelSheetNumbers.Count > 0 Then
+                    ucrBase.OKEnabled(True)
```
Patowhiz:
```suggestion
                    ucrBase.OKEnabled(lstSelectedExcelSheetNumbers.Count > 0)
```

instat/dlgImportDataset.vb
```diff
@@ -404,8 +442,14 @@ Public Class dlgImportDataset
 
             If dlgOpen.ShowDialog() = DialogResult.OK Then
                 ucrSaveFile.Reset()
-                If dlgOpen.FileName <> "" Then
-                    SetControlsFromFile(dlgOpen.FileName)
+                bMultiFiles = (dlgOpen.FileNames.Count > 1)
+                If NumberOfFileTypes(dlgOpen.FileNames) > 1 Then
```
Patowhiz: Why are we checking for number of file file names and types yet we have `dlgOpen.Multiselect = False` already set above.

instat/dlgImportDataset.vb
```diff
@@ -36,6 +41,15 @@ Public Class dlgImportDataset
...
     Private iDataFrameCount As Integer
+    Private bMultiFiles As Boolean
+
+    Private strFileName As String
+
+    Private bSupressCheckAllSheets As Boolean = False
+    Private bSupressSheetChange As Boolean = False
+
+    Dim lstSelectedExcelSheetNumbers As New List(Of Integer)
+    Dim lstSelectedExcelSheetNames As New List(Of String)
```
Patowhiz: A dictionary of integer and string would simplify and reduce the lines of code since the 2 lists correlate to each other

Patowhiz:
I have tested this with text,csv and excel files, works well from my end.
Is it necessary to give an indication that the multiple sheets will be imported as individual data frames? Something like importing the following sheets as data frames

instat/dlgImportDataset.vb
dannyparsons: I should have mentioned this, but I started implementing importing multiple files in this pull request as well. Then decided it needs more thought to do fully, but I left these bits of code in because they will be useful for when we eventually allow multifiles.

instat/dlgImportDataset.vb
dannyparsons: Thanks, but I don't think you've done the suggested change correctly. This will just replace line 413 with your new line, whereas you want to replace the whole If statement. Let me see if I can do the suggested change correctly.

dannyparsons: Ok so you can only do a suggestion to one line of code not a whole section. I'll have to just make this change manually. 
So we should only be using the "Suggested change" to make an edit to a single line of code - good to know.

instat/dlgImportDataset.vb
dannyparsons: Thanks, good idea.

dannyparsons:
Thanks for the review. 
I'm not sure what else users might think they will get imported as if not individual data frames? I don't think anyone would expect them to be merged into one sheet automatically.

dannyparsons:
@Patowhiz I've made those suggested changes. I also resized the Excel options box to remove blank space now that selected sheet is gone. Can you test again?

instat/dlgImportDataset.vb
Patowhiz: Okay. Probably a TODO of the intent will help the future developer of this dialog

dannyparsons: Good point, done.

Patowhiz:
Yes I'm testing again

Patowhiz:
@dannyparsons this looks good from my end

dannyparsons:
Thanks.

PR #5186, opened 2019-02-16
Create Survival Object dialog

lilyclements:
Not ready for review.

Current issues
1. I set the ValidationTypeAsNumeric for an input, however, it still takes non-numeric values. I saw that this is an issue in other dialogs (such as climatic ones with threshold as an input). However, it is not an issue for the ValidationTypeAsNumericList option. @dannyparsons is this a known issue?
2. Changing the survival object name doesn't update in the code, it just keeps the default name. It is unclear if this is me coding it wrong in the dialog, or if this is an issue in the setup of the Surv object outside of the dialog. Something to explore.
3. Issue with logic columns for "interval" and "interval2" types in R. It claims to accept numeric or logical columns, however, if I put a logical column in, it throws an error. Need to look into this a little more.

```
library(survival)
data(cancer)
cancer$status.logic <- with(cancer,
                              ifelse(status == 1, TRUE, FALSE))
class(cancer$status.logic)
[1] logical
with(data=cancer, exp=survival::Surv(time = time,
                                        time2 = time,
                                        event = status.logic,
                                        type = "interval"))
Error in survival::Surv(time = time, time2 = time, event = status.logic,  : 
  Invalid status value, must be logical or numeric
```

4. (small one) work out how to say "if this receiver is currently selected then ..." for when we switch tabs and one receiver is no longer visible.
5. Work on changing the status indicator

### DIFF
--- a/instat/dlgSurvivalObject.Designer.vb
+++ b/instat/dlgSurvivalObject.Designer.vb
@@ -32,6 +32,7 @@ Partial Class dlgSurvivalObject
         Me.rdoCounting = New System.Windows.Forms.RadioButton()
         Me.rdoMstate = New System.Windows.Forms.RadioButton()
         Me.rdoInterval2 = New System.Windows.Forms.RadioButton()
+        Me.ucrInputOrigin = New instat.ucrInputTextBox()
         Me.ucrPnlType = New instat.UcrPanel()
         Me.ucrSaveObject = New instat.ucrSave()
         Me.ucrReceiverEvent = New instat.ucrReceiverSingle()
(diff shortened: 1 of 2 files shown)


dannyparsons:
No I wasn't aware of that, it might be a general issue.

> 1. I set the ValidationTypeAsNumeric for an input, however, it still takes non-numeric values. I saw that this is an issue in other dialogs (such as climatic ones with threshold as an input). However, it is not an issue for the ValidationTypeAsNumericList option.

It might be because the columns have extra attributes? We've seen that before. Try taking the script into RStudio to get more details on the errors maybe.

> 3. Issue with logic columns for "interval" and "interval2" types in R. It claims to accept numeric or logical columns, however, if I put a logical column in, it throws an error. Need to look into this a little more.

lilyclements:
Current issues to work on:

1 - 4 outlined above

5. factor grid runs `c(c())`

6. The ucrReceiverEvent is set to only receive numeric columns when some rdos are selected, and to receive numeric or factors otherwise. However, it is never accepting factors. Tried debugging and it is running this line.

7. Running bits manually (for now?). Need to look into the possibility of running these bits in the automatic system.
e.g.
```
If the check box is checked then
    If the variable in the event receiver is a numeric variable
         Numeric input is visible
         the r-code is event = variable %in% c(x = ...)        [i.e. run ucrInput value]
    Else
         Factor grid is visible
         the r-code is event = variable %in% c(y = ...)        [i.e. run ucrFactorGrid value]
    End If
Else
   the r-code is event = variable
```

Struggling to run this in the automatic system because we can't specify (to my knowledge) for a certain ucr to be visible/not visible based off the current data type in the receiver. Therefore, also, the `bAddRemoveParameter` line can't be run so x and y have to manually added and removed from c()

Also unsure how to do this in the automatic system for the checkbox, but starting to get an idea with the 
`.AddParameterIsROperatorCondition` command.
When it is checked, the parameter "event" has value "%in%"
Otherwise, the parameter "event" has value of the item in the ucrReceiverEvent

8. Need to check how the factor grid works for logic columns. Hopefully treats them as a factor.

9. The factor grid resets upon opening

PR #5192, opened 2019-02-20
I have added Receiver Group to the Line plot main dialog

### LINKED ISSUE #5184: Add group to Line Plot main dialog
On the line plot dialog we set the `group` aesthetic automatically sometimes - when the x axis is a factor otherwise you don't get a line. It also then relates to what is set for the colour aesthetic.

This works fine for many graphs, but if you then want to override the `group` aesthetic to get different lines by another variable by going to the subdialog, when you return it can still get reset because of the automatic code and you don't know until run the dialog.

A suggestion is to include a receiver for `group` on the main dialog. It can still be set automatically in the same way, but now it will be clear to users what is happening and they can easily override it. When we require the group to be fixed because x is a factor the receiver can be disabled but still contain the variable.

@maxwellfundi if this makes sense can you assign to someone and work on it with them?

Wycklife:
This fixes #5184 

@africanmathsinitiative/developers. This is ready for review


instat/dlgLinePlot.vb
```diff
@@ -249,11 +257,22 @@ Public Class dlgLinePlot
...
         End If
     End Sub
+
+    Private Sub AllControl_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrVariablesAsFactorForLinePlot.ControlContentsChanged, ucrSave.ControlContentsChanged, ucrReceiverX.ControlContentsChanged
+
+    End Sub
+
+    Private Sub UcrVariablesAsFactor_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrVariablesAsFactorForLinePlot.ControlValueChanged
+
+    End Sub
```
maxwellfundi: @Wycklife  you seem to have empty sub here.. could you remove them

Wycklife:
@maxwellfundi I'm going through it now.

Wycklife:
@maxwellfundi I have resolve this.

Muthenya:
@maxwellfundi were you able to review this again?

maxwellfundi:
This looks good!

dannyparsons:
@maxwellfundi Did you test this with a graph that does multiple lines through group? This was the original intention of adding this. Also, does it sync with the receiver in the layer options?

PR #5197, opened 2019-02-21
Enhancing Windrose Dialog

maxwellfundi:
@rdstern  @dannyparsons  We now have the main dialog looking like this with the titles... they are actually working too. Now its getting the subdialog to work today

image


rdstern:
This looks good.  As you say it is what Samuel asked for.  My own view is that it sits better on a sub-dialogue, but I like it there for now, because (as you said) it was stated by Samuel to be particularly important.  I know the March meeting is for CLIMSOFT, but hope there is time to show this, at least to Samuel, and probably also to Marcellin and the other Samuel as evidence we are "listening".  And to get his views if he shows interest.

dannyparsons:
Looks good. Probably "Wind Rose Options" for the button to be consistent. And I would move it to the left side similar to how it is on other graphics dialogs.

maxwellfundi:
@Muthenya Can you please review this.. its needed for today release. report any bug you might find. Check that the colors work properly and also the linking that its working for those colors radio buttons.

Muthenya:
I'm on it.

Muthenya:
The dialog seems to work well. Just a few issues with the subdialog;
1. The conditions for the radio buttons are not well set
image

2. The subdialog does not reset when the dialog is reset.

3. The combo boxes for the radio buttons in the subdialog do not have default values (that is, are empty when they load). There is also no default col_pal for the radio buttons.

image

4. A more minor issue. The Type of Palette group box can be moved to be more centrally positioned in the tab and possibly the same be done for the controls within the group box.

These are what I have come across so far. I'm still reviewing.

PR #5207, opened 2019-02-27
File > New Data Frame

### LINKED ISSUE #5097: Adding Try to the File > New dialogue

Patowhiz:
Fixes #5097


instat/dlgNewDataFrame.vb
```diff
@@ -296,4 +297,31 @@ Public Class dlgNewDataFrame
...
+            If ucrInputCommand.IsEmpty Then
+                ucrInputTryMessage.SetText("")
+                ucrInputTryMessage.txtInput.BackColor = Color.White
+            Else
+                strVecOutput = frmMain.clsRLink.RunInternalScriptGetOutput(ucrInputCommand.GetText, bSilent:=True)
+                If strVecOutput IsNot Nothing Then
+                    'If strVecOutput.Length > 0 Then
+                    'End If
+                    ucrInputTryMessage.SetText("Command Ok.")
+                    ucrInputTryMessage.txtInput.BackColor = Color.White
```
Patowhiz: @dannyparsons though not urgent, we will have add some extra methods "features" into our custom controls. Looking at the code for most dialogues, I'm concerned about how we have to access the inner controls of the custom controls, this violates OOP principle (Encapsulation).

dannyparsons: Yes agreed, and happy for you to add an issue on that with the methods to be done, or just to do it yourself in a separate PR.

maxwellfundi:
@Patowhiz  any progress on this?

Patowhiz:
@maxwellfundi this is indicated as ready for review above

maxwellfundi:
@dannyparsons have a look at this too.

dannyparsons:
I can have a look but someone needs to review as well.

stevenndungu:
Reviewed. It works as expected.
@Patowhiz Should Ok be enabled iff the try solves to "command ok" or?

stevenndungu:
Noticed similar characteristics in dlg Enter and Calculations, @dannyparsons this is ready for you to review.

Patowhiz:
@stevenndungu not really, the user can actually just go directly to click **ok** even without trying the command first.

stevenndungu:
Yes for sure. I understand why it is left like that. Thanks.

PR #5213, opened 2019-02-28
Split text

### LINKED ISSUE #5194: Split Text into Maximum components
In the Split Text dialog we can split a column into components split by a character e.g. `:` but you have to give the number of components (columns) to return. Often you just want the maximum possible number of components and to have blank values when there were less than the maximum in that row.

The dialog uses `stringr::str_split_fixed` where you must specify `n`. There is also `stringr::str_split` which can have `n = Inf` but it returned a list of vectors not a data frame, so we didn't use it.

Is there now a function which does what `stringr::str_split` but returns a data frame/matrix with padded empty values? Or will we have to write our own function for this?

Ivanluv:
fixes #5194 


instat/dlgSplitText.vb
```diff
@@ -14,11 +14,12 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports instat
```
maxwellfundi: @Ivanluv this line can be removed... its not needed anymore

instat/dlgSplitText.vb
```diff
@@ -134,4 +154,26 @@ Public Class dlgSplitText
...
         TestOKEnabled()
     End Sub
+
+    Private Sub rdoFixedNumberOfComponents_CheckedChanged(sender As Object, e As EventArgs) Handles rdoFixedNumberOfComponents.CheckedChanged
+        ucrNudPieces.Show()
+    End Sub
+
+    Private Sub rdoMaximumNumberOfComponents_CheckedChanged(sender As Object, e As EventArgs) Handles rdoMaximumNumberOfComponents.CheckedChanged
+        ucrNudPieces.Hide()
+    End Sub
```
Patowhiz: Though probably trivial, to reduce the code lines, the above 2 radio buttons events can be simplified and made into one, since by default radio buttons behave like group buttons(_Toggle Group_). Checking one automatically unchecks the other one. You could have something like 
`ucrNudPieces.visible = rdoFixedNumberOfComponents.Checked` in `rdoFixedNumberOfComponents_CheckedChanged`. You can investigate this using break points in regard to how the 2 events are raised.

I think CheckChanged event gets raised even when the radio button is unchecked.

instat/dlgSplitText.vb
```diff
@@ -14,11 +14,12 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports instat
 Imports instat.Translations
 Public Class dlgSplitText
     Public bFirstLoad As Boolean = True
     Private bReset As Boolean = True
-    Private clsTextComponents As New RFunction
+    Private clsTextComponents, clsTextComponents2 As New RFunction
```
Patowhiz: I think its safer to name the 2 RFunctions using much more suggestive names e.g `clsTextComponentsFixed, clsTextComponentsMaximum`. Just to ease code readability.

instat/dlgSplitText.vb
```diff
@@ -134,4 +154,26 @@ Public Class dlgSplitText
...
+    Private Sub rdoFixedNumberOfComponents_CheckedChanged(sender As Object, e As EventArgs) Handles rdoFixedNumberOfComponents.CheckedChanged
+        ucrNudPieces.Show()
+    End Sub
+
+    Private Sub rdoMaximumNumberOfComponents_CheckedChanged(sender As Object, e As EventArgs) Handles rdoMaximumNumberOfComponents.CheckedChanged
+        ucrNudPieces.Hide()
+    End Sub
+
+    Private Sub ucrPnlTextComponents_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlTextComponents.ControlValueChanged
+        SplitTextOptions()
```
Patowhiz: You have called `SplitTextOptions()` in 2 places, `ucrPnlTextComponents_ControlValueChanged` and `ucrPnlSplitText_ControlContentsChanged`. I think calling it in only 1 of those events in regards to the panel changes will achieve your intentions. Since both will be raised on Checked status of the radio buttons in them.

instat/dlgSplitText.vb
Ivanluv: @maxwellfundi  got it.

instat/dlgSplitText.vb
Ivanluv: good idea

instat/dlgSplitText.vb
Ivanluv: I agree

Ivanluv:
@patowhiz  made the necessary changes .Please review this

instat/dlgSplitText.vb
dannyparsons: I agree with the change of code but I would prefer to keep the events of both radio buttons. While one
`rdoFixedNumberOfComponents_CheckedChange` is sufficient, its better to keep both to ensure bugs are not introduced if the dialog functionality changes.

Muthenya:
@Patowhiz have reviewed this again.

Muthenya:
Apart from the minor edits to be made, the dialog looks good from my end.

maxwellfundi:

instat/dlgSplitText.vb
dannyparsons: Actually, why are not using the `ControlValueChanged` of the panel here? This will be raised when the `CheckedChanged` of either radio button is raised so you can do it in one sub with one event. I agree with Patrick's suggested change on setting the visibility.

Muthenya:
@dannyparsons ?

Muthenya:

instat/dlgSplitText.vb
```diff
@@ -10,15 +10,14 @@
 ' but WITHOUT ANY WARRANTY; without even the implied warranty of
 ' MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 ' GNU General Public License for more details.
-'
-' You should have received a copy of the GNU General Public License 
-' along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
dannyparsons: This shouldn't be removed.

instat/dlgSplitText.vb
dannyparsons: Yes and it should only be on `ControlValueChanged`. `ControlContentsChanged` is used for doing non code checks, like setting OK enabled.

maxwellfundi:
@Ivanluv  this seems to need a very minor change, could you have a look at it?

maxwellfundi:
@Ivanluv  I have tested this and gone through the code and looks good. One thing to change here, the NUD for the number of pieces to return, does not reset to default value when reset is clicked

dannyparsons:
@Ivanluv?

Ivanluv:
Am working on it today

Ivanluv:
@dannyparsons  would you look at this

dannyparsons:
If someone can review and is happy with it I can do a final check.

instat/dlgSplitText.vb
```diff
@@ -42,13 +43,22 @@ Public Class dlgSplitText
...
 
-        ucrPnlSplitText.AddToLinkedControls(ucrNudPieces, {rdoTextComponents}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=2)
+        ucrPnlTextComponents.AddFunctionNamesCondition(rdoMaximumNumberOfComponents, "str_split")
+        ucrPnlTextComponents.AddFunctionNamesCondition(rdoFixedNumberOfComponents, "str_split_fixed")
+
+        ucrPnlTextComponents.AddToLinkedControls(ucrNudPieces, {rdoFixedNumberOfComponents}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, objNewDefaultState:=2, bNewLinkedChangeParameterValue:=True)
         ucrPnlSplitText.AddToLinkedControls(ucrSaveColumn, {rdoTextComponents}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrPnlSplitText.AddToLinkedControls(ucrPnlTextComponents, {rdoTextComponents}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        'ucrPnlSplitText.AddToLinkedControls(ucrSaveColumn, {rdoTextComponents}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
```
shadrackkibet: Commented code should be removed entirely not unless there is a reason why it should remain. For this case I see a duplicate in line 56.

instat/dlgSplitText.vb
```diff
@@ -134,4 +153,18 @@ Public Class dlgSplitText
     Private Sub Controls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrInputPattern.ControlContentsChanged, ucrReceiverSplitTextColumn.ControlContentsChanged, ucrNudPieces.ControlContentsChanged, ucrSaveColumn.ControlContentsChanged, ucrPnlSplitText.ControlContentsChanged
```
shadrackkibet: If I remember correctly, our format is that this should always be the last sub. It doesn't make much difference but consistency is important.

maxwellfundi:
@shadrackkibet  Could you review this?

shadrackkibet:
Yes, I have tested this and fixed small issues we had. @dannyparsons this is ready for you.

instat/dlgSplitText.vb
dannyparsons: You now have repeated code in `ControlContentsChanged` that is in `SplitTextOptions()`. I think you can remove `ControlContentsChanged`?

instat/dlgSplitText.vb
```diff
@@ -121,17 +141,28 @@ Public Class dlgSplitText
...
     End Sub
+    Private Sub ucrPnlTextComponents_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlTextComponents.ControlValueChanged
+        SplitTextOptions()
+    End Sub
 
-    Private Sub Controls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrInputPattern.ControlContentsChanged, ucrReceiverSplitTextColumn.ControlContentsChanged, ucrNudPieces.ControlContentsChanged, ucrSaveColumn.ControlContentsChanged, ucrPnlSplitText.ControlContentsChanged
+    Private Sub SplitTextOptions()
+        If rdoFixedNumberOfComponents.Checked Then
+            ucrBase.clsRsyntax.SetBaseRFunction(clsTextComponentsFixed)
+            clsTextComponentsFixed.SetAssignTo("Split", strTempDataframe:=ucrSelectorSplitTextColumn.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempColumn:="Split", bAssignToIsPrefix:=True)
```
dannyparsons: The assign to should be handled automatically by the save control. This shouldn't be needed. Also, the column name definitely shouldn't be fixed like this - it means the text the user types will be ignored and always "Split".

instat/dlgSplitText.vb
```diff
@@ -78,31 +88,41 @@ Public Class dlgSplitText
...
+        clsTextComponentsFixed.AddParameter("n", strParameterValue:=2, iPosition:=2)
+
+        clsTextComponentsMaximum.SetPackageName("stringr")
+        clsTextComponentsMaximum.SetRCommand("str_split")
+        clsTextComponentsMaximum.AddParameter("pattern", Chr(34) & "," & Chr(34), iPosition:=1)
+        clsTextComponentsMaximum.AddParameter("n", "Inf", iPosition:=2)
+        clsTextComponentsMaximum.AddParameter("simplify", "TRUE", iPosition:=3)
 
-        ucrBase.clsRsyntax.SetBaseRFunction(clsTextComponents)
+        clsTextComponentsFixed.SetAssignTo(strTemp:="Split", strTempDataframe:=ucrSelectorSplitTextColumn.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempColumn:="Split", bAssignToIsPrefix:=True)
```
dannyparsons: I think you can set the assign to for both functions here.
Also, can you change to lower case "split" which is the convention for column names.

instat/dlgSplitText.vb
Ivanluv: the codes for `controlsContentChanged ` and `SplitTxtOption()` are for different panels.However,I think changing  `controlsContentsChanged`  to `controlValueChanged` might be necessary

dannyparsons: oh yes that's right. But then you have the same issue there with the set assign to. And yes use `controlValueChanged` instead.
Probably SplitTextOptions() needs to be called under ucrPnlSplitText as well since this also updates the base function.

instat/dlgSplitText.vb
```diff
@@ -139,14 +141,13 @@ Public Class dlgSplitText
...
+    Private Sub ucrPnlSplitText_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlSplitText.ControlValueChanged
 
+        If rdoBinaryColumns.Checked Then
             ucrBase.clsRsyntax.SetBaseRFunction(clsBinaryColumns)
-            clsBinaryColumns.SetAssignTo("Split", strTempDataframe:=ucrSelectorSplitTextColumn.ucrAvailableDataFrames.cboAvailableDataFrames.Text, bAssignToColumnWithoutNames:=True)
         ElseIf rdoTextComponents.Checked Then
-            ucrSaveColumn.SetName("Split")
-        Else
+            SplitTextOptions()
+            ucrSaveColumn.SetName("split")
```
dannyparsons: Do we need to set the name each time you click Text Components? 
I think its probably not needed so this line could go.

dannyparsons:
@Ivanluv what is the state of this?

Ivanluv:
I removed the line you suggested so it can be tested

dannyparsons:
Looks good.

PR #5220, opened 2019-03-01
Improving the menu names of import dialogs on the Form Main

### LINKED ISSUE #4983: Import or Open?
We have both Open and Import in various places on File and Climatic menus. On the climatic menu particularly it is looks very inconsistent. There's also just 'Climsoft' and 'Clidata' with no open or import.

image

This relates to #4981 and might be good to resolve while doing the help.

I would suggest these are all **Import** because there is some kind of converting going on - they are not _R-Instat files_ - but then I would like to keep **Open from File** on the File menu and I don't have a good reason for this difference. 

Any thoughts?

Muthenya:
Fixes #4983.
@maxwellfundi, @dannyparsons this is ready.


maxwellfundi:
@dannyparsons  this is ready for you with suggested form main changes.

PR #5222, opened 2019-03-04
Adding the as.character function to Make Date Dialog

### LINKED ISSUE #5203: Make Date from a numeric column
If data comes in with a date column in the format `20190101` (known an YEARMODA) or any other variation of this without `-` or `/` separators then it will be a numeric column by default. The Make Date dialog only accepts non numeric columns when using a format. This is because:
```
as.Date(19990101, format = "%Y%m%d")
# Error in as.Date.numeric(19990101, format = "%Y%m%d") : 
# 'origin' must be supplied 
```
R assumes a numeric value is a number from an origin date, not a format.

So you can easily convert this to a character column before using the Make Date dialog, but this is hard to explain to new users why this needs to be done - and its a common initial step with climatic data.

To get around this we could do:
```
as.Date(as.character(19990101), format = "%Y%m%d")
[1] "1999-01-01"
```
Which of course works if the value was already a character (or factor).

So we could modify the Make Date dialog to always wrap `as.character()` around the column when specifying a format. That makes the code slightly more complex than needed in most cases when its already character, but we may not care about this because it solves a confusing issue.

Muthenya:
Fixes #5203.


maxwellfundi:
I have been testing this and I have a challenge selecting the column I wanted. To explain my test

1. I got a dataset with normal dates `1935-01-01`
2. Made the date column a factor
3. I split the column into three splits(year, month, day)
4. I then combine the spilts into one to make a column with values like `19350101` which is initially a character column called `Combine1`
5. I convert this to numeric so that I test the issue raised by @dannyparsons in #5203 

When i now open the make date dialog, i could not  see my `Combine1`  as shoen below. I think we need to allow the receiver to receive numeric columns.  @dannyparsons  @rdstern  What do you think?

image

dannyparsons:
Yes absolutely, this is a good example. Now that we are using `as.character()` the receiver can allow any column, and needs to allow numeric to solve the issue. This was commented in the issue https://github.com/africanmathsinitiative/R-Instat/issues/5203#issuecomment-467378643

instat/dlgMakeDate.vb
```diff
@@ -441,12 +448,14 @@ Public Class dlgMakeDate
                 clsDateFunction.AddParameter("origin", clsRFunctionParameter:=clsDefaultDate)
             ElseIf ucrInputOrigin.GetText = "Gregorian" Then
                 clsDateFunction.AddParameter("origin", clsRFunctionParameter:=clsGregorianDefault)
-            ElseIf ucrInputOrigin.GetText = "Julian Day Number"
+            ElseIf ucrInputOrigin.GetText = "Julian Day Number" Then
                 clsDateFunction.AddParameter("origin", clsRFunctionParameter:=clsJulianDateDefault)
             End If
         ElseIf rdoSpecifyFormat.Checked Then
             ucrReceiverForDate.SetExcludedDataTypes({"numeric"})
```
maxwellfundi: @Muthenya  I think this what needs to change so that numeric columns are accepted.

maxwellfundi: @Muthenya  Please make this simple change then will be ready to be merged.

Muthenya: Done.

Muthenya:
This is ready now.

instat/dlgMakeDate.vb
```diff
@@ -411,10 +411,12 @@ Public Class dlgMakeDate
...
 
     Private Sub SelectorHeader()
-        If rdoOrigin.Checked Then
+        If rdoDefaultFormat.Checked Then
+            ucrReceiverForDate.strSelectorHeading = "Non Numerics"
+        ElseIf rdoOrigin.Checked Then
             ucrReceiverForDate.strSelectorHeading = "Numerics"
         Else
-            ucrReceiverForDate.strSelectorHeading = "Non Numeric"
+            ucrReceiverForDate.strSelectorHeading = "Date Variables"
```
Muthenya: I wasn't sure on the appropriate Heading. Hope this will do.

dannyparsons: I think just `Variables` since its everything. Check what it is on other dialogs which allow all columns.

Muthenya: Okay. Thanks.

Muthenya: This is done.

PR #5233, opened 2019-03-07
Rename data frame

Ivanluv:
fixes#5210


Ivanluv:
@Wycklife  please review this

Wycklife:
This works well from my review

dannyparsons:
This change will only effect the dialog on reopening. It doesn't look like it will do what I was suggesting:

> when I change the data frame, the new name should default to the current name.

Instead of running `SetDefaults()` on load, you could modify `CheckAutoName()` to always update the data frame name, regardless of whether the user has typed or not.

Ivanluv:
@dannyparsons  is this fine?

dannyparsons:
I think that looks ok. Can someone re-test to confirm it does update all the times we want?
You could then delete the commented lines.

PR #5234, opened 2019-03-07
dlgOpenNetCDF changes

### LINKED ISSUE #5209: Go to import NETcdf file dialog if chosen from Open From File Dialog
A confusion for users I've seen is that they choose a NETcdf (.nc) file from Open From File and then it says it does not recognise the file. They didn't realise there is a separate dialog for NETcdf files and then they think R-Instat can't import them.

To help with this if a user chooses a NETcdf file from the Open From File dialog we should be able to close that dialog and automatically open the NETcdf dialog with the file they selected already chosen.

Ivanluv:
fixes #5209 


instat/dlgOpenNetCDF.vb
```diff
@@ -63,11 +64,28 @@ Public Class dlgOpenNetCDF
         If bStartOpenDialog Then
             GetFileFromOpenDialog()
             bStartOpenDialog = False
+        ElseIf bRefered Then
+            dlgImportDataset.Close()
+
```
maxwellfundi: This could be removed. I realize you are closing this in the import dialog too. So this could be removed.

instat/dlgOpenNetCDF.vb
```diff
@@ -302,4 +320,13 @@ Public Class dlgOpenNetCDF
...
         End If
     End Sub
+    Public Sub checkOkChange()
+        If bRefered And ucrBase_ClickOk() Then
+            SetDefaults()
+        End If
+    End Sub
+
+    Private Function ucrBase_ClickOk() As Boolean
+
```
maxwellfundi: An empty sub here could be removed.

maxwellfundi:
@Ivanluv  This is good and you could also take up the other files too- Shapefiles

instat/dlgImportDataset.vb
```diff
@@ -600,10 +600,10 @@ Public Class dlgImportDataset
             strFileType = "DAT"
             clsImportDAT.AddParameter("file", Chr(34) & strFilePathR & Chr(34))
             ucrBase.clsRsyntax.SetBaseRFunction(clsImportDAT)
-        ElseIf strFileExt = ".nc" Then
-            dlgOpenNetCDF.ShowDialog()
+        ElseIf strFileExt = ".nc" Then3
```
Muthenya: @Ivanluv the "3" after "Then" should be removed.

instat/dlgOpenNetCDF.vb
Muthenya: @Ivanluv this still needs to be removed.

instat/dlgImportDataset.vb
maxwellfundi: @Ivanluv this only needs to remove the 3 on line 603. I remember it was reviewed and ready?

instat/dlgOpenNetCDF.vb
```diff
@@ -302,4 +318,12 @@ Public Class dlgOpenNetCDF
             clsNcOpenFunction.SetAssignTo(strFileAssignName)
         End If
     End Sub
+    Public Sub checkOkChange()
+        If bRefered And ucrBase_ClickOk() Then
```
dannyparsons: What's the purpose of this if `ucrBase_ClickOk()` is empty?

instat/dlgOpenNetCDF.vb
```diff
@@ -63,11 +64,26 @@ Public Class dlgOpenNetCDF
         If bStartOpenDialog Then
             GetFileFromOpenDialog()
             bStartOpenDialog = False
+        ElseIf bRefered Then
```
dannyparsons: If `bStartOpenDialog` is `True` then this code won't run even if coming from the import dialog.

instat/dlgOpenNetCDF.vb
Ivanluv: confirms if the ok button is clicked

dannyparsons: If it's an empty function, then that's not what it's doing.

Ivanluv:
I've removed the function and replaced it with a boolean variable in the commit above

dannyparsons:
Why is that needed? SetDefaults() shouldn't be run twice on load. You also haven't addressed by comment on line 67 yet.

Ivanluv:
@dannyparsons  would you please look at this .I have made some few changes which include not running setDefaults() twice on load.

instat/dlgOpenNetCDF.vb
```diff
@@ -36,9 +36,11 @@ Public Class dlgOpenNetCDF
     Private bCloseFile As Boolean = False
     Private strFileAssignName As String = "nc"
     Private iExpandedWidth As Integer
-    Private strLibraryPath As String = Path.Combine(frmMain.strStaticPath, "Library", "Climatic", "Satellite/")
+    Private strLibraryPath As String = Path.Combine(frmMain.strStaticPath, "Library", "Climatic", "Satellite/")
     Private bFromLibrary As Boolean = False
     Private bSubDialogOKEnabled As Boolean = True
+    'bFromImportDatasets checks whether the dialogue is opened from dlgImportDatasets
+    Private bFromIMportDatasets As Boolean = False
```
dannyparsons: typo here: should be lower case `m`

instat/dlgOpenNetCDF.vb
```diff
@@ -66,10 +68,34 @@ Public Class dlgOpenNetCDF
...
+        If strFilePath <> "" And bFromIMportDatasets Then
+            Dim strFileName As String
+            Dim strTemp As String = ""
+
+            ucrInputFilePath.SetName(strFilePath)
+            strFileName = Path.GetFileNameWithoutExtension(strFilePath)
+            ucrInputDataName.SetName(frmMain.clsRLink.MakeValidText(strFileName))
+            clsNcOpenFunction.AddParameter("filename", Chr(34) & strFilePath & Chr(34))
+            clsNcOpenFunction.ToScript(strTemp)
+            frmMain.clsRLink.RunScript(strTemp, strComment:="Opening netCDF file", bUpdateGrids:=False)
```
dannyparsons: With this section of code we should be considering the DRY (Don't repeat yourself) principle https://en.wikipedia.org/wiki/Don%27t_repeat_yourself.
If you find yourself copying and pasting the same piece of code then we should consider if we can separate out the repeated code to avoid having it in two places.

Everything above here is already repeated in the dialog so you should try and put these parts into their own function and then call it from each place its needed. Then there will be no repeated code.

shadrackkibet:
@Ivanluv what is the status of this?

Ivanluv:
@shadrackkibet  this can be reviewed

shadrackkibet:
Conflicts again?

Issue #5216, opened 2019-03-01
Custom Utility Classes, Functions, Enumerations


Patowhiz:
**Is your feature request related to a problem? Please describe.**
Currently we have lots of code that could be standardized through utility classes, functions and enumerations. 
**Describe the solution you'd like**
For instance ;

1. The column types are currently defined hard coded strings "constants", as a developer I have to peek into the receiver to correctly get the spellings of the column type string that I want if I had to manually pass it as a parameter. We can have these as Enumerations which is much safer and standardizes the hard coded strings.

2. If I want to force the vb.net string to have quotes that R "recognizes" I have to include Chr(34) for double quotes. If these feature was to change for whatever reason then we will have to go and change these in all dialogs. I propose to have custom utility classes that can return e.g such strings properly formatted and have this centrally done. This is just one scenario.

Patowhiz:
One of the classes that need improvements in regards to this are `RSyntax ` in file **clsRSyntax**, the `iCallType `could be implemented as an enumerator.

lilyclements:
@Patowhiz what is the current status of this?

Patowhiz:
@lilyclements I looked into this a bit when rewriting the output system.

It’s something a junior developer could handle fairly easily since VB.Net is a static language. However, I think it should remain a low priority given the plans to eventually rewrite the dialogs in a different language.

PR #5247, opened 2019-03-14
View Data Frame As HTML

### LINKED ISSUE #5239: View dialogue gives error
When you use Prepare > Data Frame > View and opt to select the columns the 3rd option is for html in the output window.
This gives the sjplot stuff and gives the error below.  I suspect the command has changed?  

If it isn't easy to fix then perhaps we can disable for now?

**Error running R command(s)

Error: 'sjt.df' is not an exported object from 'namespace:sjPlot'

sjPlot::sjt.df(describe=FALSE, altr.row.col=TRUE, hide.progress=TRUE, mydf=data_book$get_columns_from_data(data_name="Moorings_monthly", col_names=c("s_year","month","rain","raindays"), force_as_data_frame=TRUE))
OK**

Patowhiz:
Fixes #5239

replaced old deprecated `sjPlot::sjt.df ` and replaced it with `sjPlot::tab_d`


dannyparsons:
@Patowhiz is this ready for review?

Muthenya:
@Patowhiz is this ready for review?

maxwellfundi:
@Patowhiz  Whats the situation with this?

dannyparsons:
@Patowhiz?

Patowhiz:
Not yet. There is a bug that I'm fixing on this one. Let me do the commit

dannyparsons:
@Patowhiz has this been done now?

shadrackkibet:
@Patowhiz  what is the status of this?

instat/dlgView.vb
```diff
@@ -213,7 +223,7 @@ Public Class dlgView
                 ucrBase.clsRsyntax.iCallType = 0
                 ucrBase.clsRsyntax.SetBaseRFunction(clsViewColumnsFunction)
             Else
-                ucrBase.clsRsyntax.iCallType = 0
+                ucrBase.clsRsyntax.iCallType = 2
```
Patowhiz: This was the bug. @dannyparsons could you help me understand what iCallType means. Its a variable of clsRsyntax class use in its RunScript method. It took me a while to understand why iCallType  0 couldn't pop up the browser and iCallType  2 could. There is no comment related to what this variable does, it's a public one

Patowhiz:

instat/dlgView.vb
Patowhiz: This was the bug. @dannyparsons could you help me understand what iCallType means. Its a variable of clsRsyntax class use in its RunScript method. It took me a while to understand why iCallType  0 couldn't pop up the browser and iCallType  2 could. There is no comment related to what this variable does, it's a public one

dannyparsons: It's used because we need to run the R code slightly differently through R.Net depending on whether output is expected and also whether the output is text, a graph or HTML.

Patowhiz: And when do we use 0, 1, 2, 3 or 4 (what do they mean). Couldn't we use enumerations for this, to make it clear enough for our future contributors

Patowhiz:
Please note when reviewing this. The display as html option doesn't have the feature to set the number of rows to display yet. Therefore when viewing a data frame with many rows it may take a lot of time before displaying the results

instat/dlgView.vb
dannyparsons: Yes enumerations would definitely be nicer to avoid needing to look in the RLink to see what each does. Would need to plan when to make that change as it will effect many files.

Patowhiz: Okay, though not a priority, I'll try to see how we can solve this in a way that maintains backward compatibility with the existing code, to reduce the number of files affected at a go. I just remembered that I opened an issued sometimes back related to this. Issue #5216

dannyparsons: Yes it might be possible to add enumeration without needing to change existing code.

Patowhiz:
@africanmathsinitiative/developers did anyone review this?

rdstern:
I have just had a quick look with a tiny data frame 3 rows  (3 rows and 4 variables).  It still came in a browser Window  - though it seems to imply it should be in the output window.  And with unreasonable formatting.
image

A larger set - 310 rows also has formatting issues:

image

It does seem to use the command very sensibly: 
# Code generated by the dialog, View Data
sjPlot::tab_df(show.type=FALSE, alternate.rows=TRUE, use.viewer=TRUE, x=data_book$get_columns_from_data(data_name="wind", col_names=c("wind_dir","wind_deg","Date","Time"), force_as_data_frame=TRUE))

There is so much to do on R-Instat that I don't think this feature is worth more time now.  Especially as @Patowhiz says - it is a bit dangerous without the options to choose a limited number of rows.

My inclination is to omit this option, at least for now.  The other 2 options work well and are sensible.  Let's either disable this option or hide it for now.

Patowhiz:
@rdstern by sensible formatting you mean the output of numbers?
Are you suggesting we disable it in this PR?

rdstern:
That's right.
It currently displays in the browser, rather than the output window.  It isn't sensible - as you have said - for large data sets without the options to display the first 6, etc.  In addition it displays some numeric variables with many decimal places.

So, I like the other 2 options a lot and they are both simple and they complement each other.  So, let's disable, or hide that third option for now.

Patowhiz:
Okay, I'm going to disable this for now if there is no any other review.
Anyone who still wants to do further review you can enable the option if you wish to do so.

Patowhiz:
@rdstern after having another look at this dialog, if we were to add the functionality of setting the number of rows to display, we could have a script that creates a temporary dataframe with the given rows then pass it for printing to the `sjPlot::tab_df()` function. But that would be a long short

Patowhiz:
I can also set the number of digits to the numerical values. The R function has a parameter for this.
Below is the full list of parameters of the R function used. I checked the new released [version](https://cran.r-project.org/web/packages/sjPlot/sjPlot.pdf)of the package. The author added the digits parameter. @rdstern Do you still want me to disable the feature

```
tab_df(x, title = NULL, footnote = NULL, col.header = NULL,
show.type = FALSE, show.rownames = TRUE, show.footnote = FALSE,
alternate.rows = FALSE, sort.column = NULL, digits = 2,
encoding = "UTF-8", CSS = NULL, file = NULL, use.viewer = TRUE,
...)
```

dannyparsons:
I agree with Roger, the issues with the function and its display make it a low priority for me for this feature. The other two options are useful already.

Patowhiz:
Okay. This is now ready then. I have disabled the option

PR #5277, opened 2019-03-27
changes to enable renaming a data frame while only changing the cases

### LINKED ISSUE #5136: Renaming a data frame and only changing the case
**Describe the bug**
If you try rename and data frame and only change the case e.g. "Dodoma" to "dodoma" you get an error:

image

This looks like an error from reogrid where it does not distinguish between sheet names that only differ by case.

**Details (optional):**
 - R-Instat Version 0.4.27

Ivanluv:
fixes 
#5136 


Muthenya:
Testing this now.

Muthenya:
This looks good.

PR #5278, opened 2019-03-27
use Dialogue form

Ivanluv:
@dannyparsons  this is ready


instat/instat.vbproj
```diff
@@ -1055,6 +1055,12 @@
     <Compile Include="sdgWindrose.vb">
       <SubType>Form</SubType>
     </Compile>
+    <Compile Include="static\InstatObject\R\dlgUseModel.Designer.vb">
```
dannyparsons: Looks like this is in the wrong place under `static`.

PR #5281, opened 2019-03-28
Corrected calculation for end of season

dannyparsons:


dannyparsons:
@rdstern can you review this from here or should it be merged first?

rdstern:
I can try, but why not merge?

dannyparsons:
I can merge if that's easier.

dannyparsons:
There are some issues with the two main checkboxes resetting here but I have not worried about this as they will soon be replaced with radio buttons at the top.
That will also give much more space on the dialog so the layout can be improved then too.

PR #5283, opened 2019-03-29
 Changing the default "width" value for the Count option in Climatic Transform

### LINKED ISSUE #5280: Climatic > Prepare > Transform small issue
I hope this is an easy one to check and fix.  That is if I have understood the command properly.

I wanted to use the Climatic > Prepare > Transform for a very simple calculation, namely to provide a column to give the raindays.
I usually use the calculator and prepare a logical column using (for Dodoma) (rainday <- rain > 0.85).
I tried with the Transform Count, just counting over 1 day.  It seemed to want to count over 2 days, but was inconsistent.  When I returned to the dialogue the code seemed different.  Here is what I get the first time I go into the dialogue with the count being over 1 day:
image

Code generated by the dialog, Transform

rain_day <- instat_calculation$new(type="calculation", function_exp="(Rain >= 0.85)", result_name="rain_day", calculated_from= list("Dodoma"="Rain"))
transform_calculation <- instat_calculation$new(type="calculation", function_exp="zoo::rollapply(data=rain_day, width=2, FUN=sum, align='right', fill=NA)", result_name="rainday", sub_calculations=list(rain_day), manipulations=list(), save=2)
data_book$run_instat_calculation(calc=transform_calculation, display=FALSE)
rm(list=c("transform_calculation", "rain_day"))

If I go back to the dialogue it has changed my name back to count and it now generates something different - as follows:

# Code generated by the dialog, Transform

rain_day <- instat_calculation$new(type="calculation", function_exp="(Rain >= 0.85)", result_name="rain_day", calculated_from= list("Dodoma"="Rain"))
transform_calculation <- instat_calculation$new(type="calculation", function_exp="zoo::rollapply(data=rain_day, width=2, FUN=sum, align='right', fill=NA)", result_name="count", calculated_from=list("Dodoma"="Rain"), sub_calculations=list(rain_day), manipulations=list(), save=2)
data_book$run_instat_calculation(calc=transform_calculation, display=FALSE)
rm(list=c("transform_calculation", "rain_day"))

Neither give me the answer I want.

Muthenya:
Fixes #5280 


dannyparsons:
I don't think it was just the default. Roger said when he changed it from 2 to 1 the command didn't change. So it sounded like there was a problem with that. Can you test that?

Ogik99:
I have reviewed and the code reflects the actual width value in the dialog. I even tried intervals of three days. Please see below the code:
width3

dannyparsons:
Thanks @Ogik99. I think we need some more specific information from @rdstern on this issue. Because here only the default has changed which wouldn't effect any problems the dialog might have linking to the code. So we need to first replicate the issue of @rdstern, which might only occur in specific cases if no one has seen the same issue yet.

stevenndungu:
I have tested and looked closely with @Muthenya on issue #5280 raised by @rdstern. The two raised issued have been addressed:

1. Width parameter value problem and 

2. Correct result name - "count". 

@dannyparsons, this is ready for you to review.

PR #5285, opened 2019-03-29
Enabling the Options Button for the Two Columns Option and other edits on the Correlation Dialog

### LINKED ISSUE #5255: Add options button to correlations when just 2 variables etc?
I thought we already had an issue on this topic, but couldn't find it.  I have also found some other problems with the correlations dialogue.  I think they should all be easy?

1. Could we add the Options button to the 2-variable case?  I hope it might be trivial to add?
We now have to treat 2 as an example of many.  This adds an unnecessary step.  
If, for some reason this is difficult, then could we make Many the default, when the dialogue is first opened.

2. When I return to the dialogue (having previously completed it - including a scatterplot (by a factor) from the options I get the following error when I return to the dialogue:

**instat
Developer error: no state of control ucrPnlCompletePairwise satisfies it's condition. Cannot determine how to set the control from the RCode. Modify control setup so that one state can satisfy its conditions.
OK**   
It then goes into the dialogue fine and I can continue.

3. In the Options I have the possibility to save the graph.  In the scatterplot at least it ignores this. and continues with the default of last_graph.

Muthenya:
Fixes #5255
In progress.


Muthenya:

Ogik99:
@Muthenya this seems to work well from my end. There are no error messages and the save graph option works well.

instat/dlgCorrelation.vb
```diff
@@ -260,26 +260,16 @@ Public Class dlgCorrelation
     Private Sub ucrPnlColumns_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlColumns.ControlValueChanged
         If rdoTwoColumns.Checked Then
             ucrReceiverFirstColumn.SetMeAsReceiver()
-            cmdOptions.Visible = False
         ElseIf rdoMultipleColumns.Checked Then
             ucrReceiverMultipleColumns.SetMeAsReceiver()
-            cmdOptions.Visible = True
         End If
+        ReceiverColumns
```
dannyparsons: This should have `()` at the end.

dannyparsons:
@Muthenya I don't see any change here that would fix the graph saving issue? I would expect that would have been in the sub dialog.

instat/dlgCorrelation.vb
Muthenya: This is done

Muthenya:
I understand @dannyparsons, but I haven't had the same problem with the **Save Graph**. In my case, it seems to work well. I have tagged Roger in the issue concerning this.

dannyparsons:
Thanks, let's try and better understand that issue from Roger.

dannyparsons:
@Muthenya is it just one of the graphs the currently doesn't work with the two variables option?

Muthenya:
Yes @dannyparsons. But now the Correlations plot in the subdialog has been disabled for the Two Columns option.

Muthenya:
@stevenndungu kindly review this.

Muthenya:
@stevenndungu were you able to review this?

stevenndungu:
I reviewed this, thanks for the reminder. However, I got errors that I could not reproduce twice (I could not understand why). So let me try again and see. Giving my comments shortly.

stevenndungu:
@Muthenya it works perfectly well. 
What do you think of;
 -  Having the options group box hidden when `None` and ` Pair plot` radio buttons are checked (on subdialogue)?
 - Having the  radio button `None`  checked if two columns are checked - when we check multiple columns >Options> correlation plot. Then decide to perform a two column > options >...
image
Otherwise is good.

dannyparsons:
@stevenndungu are you able to take this over and make the minor changes? 
I suggest we keep it as no plot by default for either two or multiple columns, mainly because the plots can take a long time to produce.

dannyparsons:
Thanks for taking it up. 
We actually don't need to close this PR because once your PR is merged, if it contains all the commits from here, GitHub is clever enough to close it automatically. If it doesn't close, then it might mean there are some changes here not merged, so it's a good check to keep this open.

PR #5294, opened 2019-04-08
Adding geom_label_repel in  Climatic>Mapping dialog for better labeling of station maps

Wycklife:
Fixes#5290

@africanmathsinitiative/developers In progress


Wycklife:
@maxwellfundi Could you please check whether this is okay.

maxwellfundi:
@Wycklife I am revewing this now

maxwellfundi:
@Wycklife I have tested this and it is making good progress. Just a quick fix here now is - You seem not to be adding parameters on your `geom_label_repel()` properly as shown in bold below. 

# Code generated by the dialog, Climatic Maps
gadm36_MWI_1 <- data_book$get_data_frame(data_name="gadm36_MWI_1")
malawi_AWS_Stations <- data_book$get_data_frame(data_name="malawi_AWS_Stations")
last_map <- ggplot2::ggplot() + ggplot2::geom_sf(data=gadm36_MWI_1, ggplot2::aes(fill=NAME_1)) + **ggrepel::geom_label_repel(, station_name)** + ggplot2::geom_point(data=malawi_AWS_Stations, ggplot2::aes(x=longitude, y=latitude)) + theme_grey()
data_book$add_graph(graph_name="last_map", graph=last_map, data_name="gadm36_MWI_1")
data_book$get_graphs(data_name="gadm36_MWI_1", graph_name="last_map")

You just need to add this layer like shown below when the **Station** receiver is not empty. - I think that name too need to change to **Station Names**. 

`+ geom_label_repel(data = malawi_AWS_Stations,aes(x=longitude, y=latitude, label= station_name)) `

maxwellfundi:
@Wycklife  this looks good now. However one more check I would like you to do it- 

`Ok should be enabled if- The Longitude and Latitude are filled(The current option) and disabled if Station names is not empty and Longitude and Latitude are not filled`

maxwellfundi:
@dannyparsons @rdstern this is now working as required. My concern is the stations can be be so many that the labels still look very cluttered. An example below is just the AWS for Malawi, imagine how worse it could get if we included all stations. 

image

I am suggesting @Wycklife we have a smaller `size `for the labels and then also add to `geom_lable_repel ` `force` parameter with a big value, and also `direction = "both"` so that the labels are more spread and alot more away from each other. 

@rdstern  @dannyparsons what are your thoughts on this?

Wycklife:
@maxwellfundi I have added the three parameters you suggested.

maxwellfundi:
@Wycklife did you check that the size is okay with the labels? @rdstern what are your thoughts on labels being too many

rdstern:
I agree that we should have the labels smaller, etc.  But one advantage of this mapping system is also the facility for facets, either by region or by type of station.  And that should help too.  We should be checking what happens to the size once you facet.

maxwellfundi:
Yeah I agree

maxwellfundi:
@dannyparsons We added these parameters because in normal context we will have lots of station names and so we needed a smaller size for the labels, the lables to be repelled more so that they so not overlap.
Am sorry the default for direction is both so that did not need to be added.

dannyparsons:
I'm happy if we have our own default as a smaller size, because we feel this will be appropriate for many cases. However, if we do this then it must be changeable in the dialog (or a subdialog) so users can change this if needed. The same is true for other parameters. I don't mind if these are on the main dialog for now, as there is space and the controls will be simple.

In general, I don't think you can say,
> in normal context we will have lots of station names

You've only shown one example here, so you can't claim that's normal in most uses. Many places would produce maps with less stations (small countries or a subset of stations), or do Roger's suggestion of using facets to reduce the number. In these cases, the smaller size might not be appropriate and so we must have a way to change it. I am not even sure if the many stations or few stations will be the "normal context" of people using this dialog, but as long as this is changeable on the dialog, it doesn't matter too much what our own default is. What could be frustrating, is having a fixed value chosen by us which a user can't change if they don't like it.

maxwellfundi:
@dannyparsons I understand your concerns here and am happy for now, because we have this working properly, remove them then decide on how we can implement them in the future in a way that includes easy ways of changing them. 

@Wycklife  will remove them now, then it will be ready.

Wycklife:
@maxwellfundi I've removed them,

maxwellfundi:
@Wycklife  Great, @dannyparsons  this is now good.

instat/dlgClimaticStationMaps.vb
```diff
@@ -154,6 +162,14 @@ Public Class dlgClimaticStationMaps
         clsFacetFunction.SetRCommand("facet_wrap")
         clsFacetFunction.AddParameter("facet", clsROperatorParameter:=clsFacetOp, bIncludeArgumentName:=False)
 
+        clsLabelRepelFunction.SetPackageName("ggrepel")
+        clsLabelRepelFunction.SetRCommand("geom_label_repel")
+
+        clsLabelRepelAesFunction.SetPackageName("ggplot2")
+        clsLabelRepelAesFunction.SetRCommand("aes")
+        clsLabelRepelFunction.AddParameter("aes", clsRFunctionParameter:=clsLabelRepelAesFunction, bIncludeArgumentName:=False)
```
dannyparsons: Why not include the argument name?

instat/dlgClimaticStationMaps.vb
```diff
@@ -258,4 +282,12 @@ Public Class dlgClimaticStationMaps
             clsGGplotOperator.RemoveParameterByName("facets")
         End If
     End Sub
+
+    Private Sub ucrReceiverStation_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverStation.ControlValueChanged
+        If Not ucrReceiverStation.IsEmpty Then
+            clsGGplotOperator.AddParameter("labels", clsRFunctionParameter:=clsLabelRepelFunction, bIncludeArgumentName:=False, iPosition:=2)
```
dannyparsons: Isn't there a standard format we use for parameter names for geoms in the operator?

maxwellfundi: I think this is still okay. We dont have specific  format. This is simply a parameter name.

instat/dlgClimaticStationMaps.vb
maxwellfundi: Because we dont need it at all. it would obviously bug otherwise.

maxwellfundi:
@dannyparsons  The current version works perfectly for what we need, other additions/enhancements can be done later.

instat/dlgClimaticStationMaps.vb
maxwellfundi: We just want to pas in `aes()` without its augment name

dannyparsons: Why? The argument name is `mapping`. Why not use it?

instat/dlgClimaticStationMaps.vb
dannyparsons: I know it doesn't change the functionality, but we should aim for consistency, even if we don't have it everywhere. But I actually think we do have fairly good consistency for this in our dialog. For the `geom_sf` layer in this dialog the parameter name is `geom_sf`. This is also what's done in other dialogs. So I suggest the same naming convention here.

instat/dlgClimaticStationMaps.vb
maxwellfundi: Doesn't make a difference on how it works but I see your point

maxwellfundi:
@dannyparsons I think this has addressed your concern

instat/dlgClimaticStationMaps.vb
```diff
@@ -94,6 +95,11 @@ Public Class dlgClimaticStationMaps
         ucrReceiverFacet.SetParameterIsString()
         ucrReceiverFacet.bWithQuotes = False
 
+        ucrReceiverStation.SetParameter(New RParameter("geom_label_repel", 2))
```
dannyparsons: It's label here because this is in `aes()` of `geom_label_repel`.
```suggestion
        ucrReceiverStation.SetParameter(New RParameter("label", 2))
```

dannyparsons:
Looks good by the way!
image

instat/dlgClimaticStationMaps.vb
maxwellfundi: Sorry yea thats right.

PR #5296, opened 2019-04-10
changes to the Rename Column Dialog 

### LINKED ISSUE #5019: Renaming shape files columns bug
**Describe the bug**
When a shape file is loaded and you would like to rename a column, below developer error appears. When you say okay a few times then the renaming dialog appears. Renaming happens normally. 

**To Reproduce**
1. Import a shape file
2. Right click on the column you want to rename

**Expected behaviour**
Clicking on rename columns right click menu should lead straight to the rename column dialog

**Screenshots**
If applicable, add screenshots to help explain your problem, particularly of error messages.
image

Ivanluv:
fixes #5019 


instat/dlgName.vb
```diff
@@ -144,8 +144,8 @@ Public Class dlgName
                 If strArr.Length = 1 Then
                     strColLabel = strArr(0)
                 ElseIf strArr.Length > 1
-                    MessageBox.Show(Me, "Developer error: retrieved column label should be one.", "Developer Error", MessageBoxButtons.OK, MessageBoxIcon.Error)
-                    strColLabel = strArr(strArr.Length - 1)
+                    'MessageBox.Show(Me, "Developer error: retrieved column label should be one.", "Developer Error", MessageBoxButtons.OK, MessageBoxIcon.Error)
+                    'strColLabel = strArr(strArr.Length - 1)
                 End If
```
stevenndungu: Commenting on ` 'strColLabel = strArr(strArr.Length - 1)` implies removal of the functionality. Which is definately not correct.
I think for this particular case we need to focus on what exactly can be done if the column labels are greater than one.

stevenndungu: We can have a look at it together.

Muthenya: @stevenndungu and @Ivanluv have you been able to have a look at this?

stevenndungu: Yes, we were reflecting on what instances would lead to column labels to be more than one. Luckily, shapefiles data set provides a good example where for example geometry column will have several labels describing the variable.

My thinking, `Developer error: retrieved column label should be one`, is not always true. This may be true in most data sets but for some cases (a good example shapefiles) this does not hold. @maxwellfundi @dannyparsons. What are your views.

dannyparsons: @stevenndungu its very good to check how this works for shapefiles as they are very different to a regular data frame. 
I just tried one shape file and the geometry column didn't have anything for the `label`. It does have lists for the other metadata properties, but not `label`. That might just be my file, do you have a file where there is something already for the label? Below is the metadata where I checked this.
If shapefiles do sometimes have lists for the label property then I agree we should not have a developer error.
image

stevenndungu: You are right, shapefiles have fairly different structure though it is the geometry column that has more specific details.

I have a different file and the behavior is the same as to what you have just shown above.  I have been exploring the behavior of the shapefiles with respect to labels in particular and good news is that the behavior is equivalent to regular data frames (in R). 
Rsn: As seen above the geom type, projections,... attributes of geometry column are not considered as labels rather classified as bbox, class, crs unlike I thought in my previous comment. If there are no labels coming along or introduced we have "".

However coming to the VB side, `get_column_labels` extracts double `""` which is not the case in R - this results to the length of 2 instead of 1 as expected. Finding out why.

Ivanluv:
we are currently stuck

dannyparsons:
@Ivanluv please give more details so others can try to help. Just saying you're stuck isn't very useful.

Ivanluv:
@dannyparsons  am not sure of how we get dataframe column labels from shape files

PR #5297, opened 2019-04-10
Hiding the Regex group box

Ivanluv:


Muthenya:
@Ivanluv what the current state of this? Is it ready for review or still in progress?

Ivanluv:

Muthenya:
@Ivanluv kindly reference the pull request to the issue it is addressing.

instat/dlgStringHandling.vb
```diff
@@ -90,6 +90,8 @@ Public Class dlgStringHandling
 
         'temporary disabling
         grpRegex.Enabled = False
```
Muthenya: Just a thought. Since we are hiding the group box, could we remove the disabling of the group box?

Muthenya:
This looks good from my end.

Muthenya:
Could we 
1. Bring the receiver and textboxes up a bit?
2. Move the unseen radio buttons (regex and fixed) to come after the receiver and textboxes?

image

dannyparsons:
Could the column receiver and pattern textbox also be the same width? They both look a bit longer than normal actually, but anyway it would look better if they have the same width.

maxwellfundi:
@Ivanluv These looks good but look like it have very simple changes to make. Could you have a look at them.

Ivanluv:
@maxwellfundi am on it

maxwellfundi:
Tested this and its all good.

instat/dlgStringHandling.vb
```diff
@@ -153,7 +155,7 @@ Public Class dlgStringHandling
         clsCountFunction.SetAssignTo(ucrSaveStringHandling.GetText, strTempDataframe:=ucrSelectorStringHandling.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempColumn:=ucrSaveStringHandling.GetText, bAssignToIsPrefix:=True)
         ucrBase.clsRsyntax.SetBaseRFunction(clsCountFunction)
         NewColumnName()
-        ChangeSize()
+        'ChangeSize()
```
dannyparsons: Why is this commented out? And also below.

dannyparsons: Ok that makes sense. So if the method is no longer needed it should be deleted along with all calls to it.

PR #5305, opened 2019-04-17
Doy range

### LINKED ISSUE #5118: Date control problem in climatic
I occasionally get a developer error in the climatic commands when I press the button to change the start/end dates button that is on most of the dialogues, e.g. Climatic > Prepare > Start of the Rains.  

When it occurs it gives 2 successive messages about invalid dates.  Then I get to the sub-dialogue and it works fine.  So it isn't stopping any work.

The error message comes twice and is with the user control ucrDoyTo, i.e.

**"Developer error: Invalid month and/or day value. For control: ucrDoyTo."**

I have yet to diagnose a sequence that always leads to this.  I think it relates to the use on the same control on multiple dialogues.  This last time I used Climatic > Prepare > Summarise with the dates set to Feb 15 till June 30.  No problems there, but got a problem the first time I did the Start of the Rains afterwards.  I think the problem is the number 30 in June.

Aha, I think the problem is simple to repeat.  
a) Go into the (start of the rains) dialogue for the first time.  Can use Dodoma.
b) Set the day range to be (say) 15 Feb to End April.  (Any value (I think) so it is less than 31 - with April it is no more than 30)
c) Press reset.  Now the day range is set to 1 to 366.
d) Press the day Range button again
This gives the error above.  

Notice there is no issue with the other user control, i.e. I assume called ucrDayFrom so that must be resetting correctly.  The To says 366 (which should correspond to 31), but there is a little left-over problem here.

Ivanluv:
fixes #5118 
@dannyparsons  I've tried looking at this and made some of the changes.Would you look at this please


dannyparsons:
It would be useful if you could explain the changes you've made, and why, so we can understand the reasoning and consider the consequences.

dannyparsons:
What happens if the month is Feb and it tries to change the day to 31?

Also, you've added two lines on 137-138 which repeat 133-134, what's the reason for that?

Ivanluv:
It does not since I've made it update its from_date to January 1st and to_date to December 31st when it from a new dialog(i.e climatic summary or end date or start date dialog)
The changes I made to the first two pull request have been undone in the final pull request.

Ivanluv:
I mean the last commit.The first two have errors

dannyparsons:
Ok that makes more sense. 

> Where is it set up to reset the dates from a new dialog?
> Also, two lines on 137-138 don't seem sensible to me, they are outside the If so you don't know what dtTemp will be.

maxwellfundi:
@Ivanluv Any progress on this?

Ivanluv:
@maxwellfundi  this was fine from my end

dannyparsons:
@Ivanluv you haven't answered the question I've raised three times now:

> Also, two lines on 137-138 don't seem sensible to me, they are outside the If so you don't know what dtTemp will be.

Ivanluv:
@dannyparsons  I realised the mistake I made in lines 137-138.Which I have eliminated .
So in the `ucrDayOfYear` dialog I've swapped ` ucrInputMonth.cboInput.SelectedIndex = dtTemp.Month - 1
        ` and `  ucrInputDay.SetName(dtTemp.Day)` so that the month values update before the days,this is in line 279-280.
Also in the `sdgDoyRange` dialogue I added the if statement to make sure the `clsDayToOperator` updates correctly when the value of `ucrDoyFrom`  changes and `rdoLength` is checked which is what the `updateToValues ` is doing

PR #5308, opened 2019-04-24
PICSA rainfall graphs

### LINKED ISSUE #5120: Remaining minor issues with PICSA rainfall graphs
These are broadly working well.
 
Remaining aspects to be addressed include:

a) The Median line doesn't seem to work.  When I look at the code it seems that some hasn't been changed from the Mean.  So it sort of works if you do Mean and Median together.  But it doesn't do the Median alone, because of the code for the mean.
b) On the Y-axis tab the Specify Lower and Upper limits always reset to be enabled.  They don't seem to do any harm if no numbers are present in the text boxes.
c) On the Y-axis dates there is currently no option to set the lower and upper limits.
d) Still on the Y-axis dates, (when I changed data frame) I set the dates to 5 or 10 days, but it didn't give them.  When I looked at the code they were set back to months - I may have done a reset.  But the dialogue said days and the code said months.

stevenndungu:
Fixes #5120 
- median line works.
- specify limits resets to unchecked.
- faceting works correcting with mean, median and percentiles.
- facetting in the main dialogue overriding the plot options for 2 variable facetting.
- rest  - in progress


rdstern:
Great!

stevenndungu:
@maxwellfundi could you review the effected changes, in the meanwhile?

maxwellfundi:
@stevenndungu will be reviewing this.

stevenndungu:
fixes #5309

stevenndungu:
@maxwellfundi, review together with issue #5309 - I have made some changes that solves that as well. Note: Use both the multiple receiver and single receiver (stack the data first).

maxwellfundi:
@stevenndungu I am on this now.

maxwellfundi:
@stevenndungu  at the moment, there is an issue with the `group_by` when using the single receiver.  We need to have the group by for this case not the `variable` but the value of the colour by and/or facet to work properly. 

image

For the multiple receiver, it works perfectly even with the mean median and tercile lines. Its (y). 

Doing the stacking then doing the plot also works great, and you will see that we need to have `dplyr::group_by(station, variable)`. 

Another odd thing i have found is that, after using the dialog a couple of times, then do reset, it behaved weirdly by going back to the multiple receiver and no current receiver selected. 

The facets also work well!

stevenndungu:
Yes, you are right on the error! I was do it conditionally not when we have a facet. Great, thanks for capturing that - I have to figure out how to handle that particular case.

stevenndungu:
Why is that happening? I have not touched on that. 
`Another odd thing i have found is that, after using the dialog a couple of times, then do reset, it behaved weirdly by going back to the multiple receiver and no current receiver selected.`

rdstern:
There are oddities, partly because the code can be so complicated.  So it might have been there all the time!

stevenndungu:
Okay, thanks @rdstern. Looking at it

instat/sdgPICSARainfallGraph.vb
```diff
@@ -1232,7 +1253,7 @@ Public Class sdgPICSARainfallGraph
     Private Sub ucrYLimitControls_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkYSpecifyLowerLimit.ControlValueChanged, ucrChkYSpecifyUpperLimit.ControlValueChanged, ucrInputYSpecifyLowerLimitNumeric.ControlValueChanged, ucrInputYSpecifyUpperLimitNumeric.ControlValueChanged
         If bRCodeSet Then
             If Not ucrChkYSpecifyLowerLimit.Checked Then
-                clsCLimitsYContinuous.AddParameter("min", "NA", bIncludeArgumentName:=False, iPosition:=0)
+                clsCLimitsYContinuous.AddParameter("min", "0", bIncludeArgumentName:=False, iPosition:=0)
```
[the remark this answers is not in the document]
stevenndungu: Yes I agree. I changed to 0 to allow reset since I noticed that from the latest version we had 0 and not NA. I therefore thought there were good reasons why it was set to 0 and not NA like the max.

I have effected the changes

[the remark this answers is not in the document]
dannyparsons: Ok I hadn't noticed that actually. Let's check with Roger, @rdstern when the Y axis is numeric, should the default be that it's fixed to 0 as the lower limit or its automatically set?

rdstern:
Interesting question - fixed to zero or set automatically.
For most things I like it fixed to zero and consider that (often) to be a good practice.

But I can't claim it as the default.  So the default must be automatic.  If (later) we add one or two check-boxes to the main dialogue, then that would be a contender.

dannyparsons:
Makes sense, so we keep it as `NA` for the limits.

stevenndungu:
@africanmathsinitiative/developers this is good/finished for review from my side. @reviewer, I am happy to provide guidance where need be on review.

maxwellfundi:
Great, am going to test it first thing today.

instat/sdgPICSARainfallGraph.vb
```diff
@@ -1303,18 +1302,20 @@ Public Class sdgPICSARainfallGraph
...
+            If ucrChkYSpecifyLowerLimit.Checked Then
+                ucrInputYSpecifyLowerLimitDateMonth.Visible = True
+            Else
+                ucrInputYSpecifyLowerLimitDateMonth.Visible = False
+            End If
 
+            If ucrChkYSpecifyUpperLimit.Checked Then
+                ucrInputYSpecifyUpperLimitDateMonth.Visible = True
+            Else
+                ucrInputYSpecifyUpperLimitDateMonth.Visible = False
```
shadrackkibet: @stevenndungu I guess you tried linking controls here. I would like to know why that doesn't work.

stevenndungu: yes, I tried linking the controls - check some commented lines...reason why it failed to work - looking at how the parameters and function (line 1288) of the two controls added (those with the months- combo boxes). The behaviour was not as I wanted exactly. So I opted out that way.

maxwellfundi:
@stevenndungu  The issue i had raised before when i tested this, still persists whilst using single receiver. 

My dialog if filled in like this 

image

And the error i get is 

image


This is because of the you are having grouping with **Variable and station**,  when we are adding lines hence this causes this bug... we only need the grouping with the station only for this case. 

guinea_2_by_station_year <- guinea_2_by_station_year %>% **dplyr::group_by(variable, station)** %>% dplyr::mutate(.mean_y=mean(x=sum_rain, na.rm=TRUE), .median_y=median(x=sum_rain, na.rm=TRUE), .lower_ter_y=quantile(x=sum_rain, probs=1/3, na.rm=TRUE), .upper_ter_y=quantile(x=sum_rain, probs=2/3, na.rm=TRUE))

This only happens when i add the lines, so please check this out. The case of the multiple receiver works perfectly well.

stevenndungu:
@maxwellfundi I am afraid what you just did is not the dialogue is meant to do. I stand to be corrected however.
What you did should be performed by some other dialogues in describe.

maxwellfundi:
what do you mean? I might want to have the graph of the sum of the rain for a place with the stations being the facets and i want them have different colour.. whats wrong with my dialog?

stevenndungu:
When using the single receiver your data will always be stacked. Therefore, `variable` will always be present, unlike how the error says. It is actually done internally (melting of the data frame provided) `stacked= TRUE` . The user is not meant to necessarily know

maxwellfundi:
@stevenndungu  I dont think that is quite right, stacking only happens when we use the multiple one(internally) unless  i misunderstood what you are saying.

dannyparsons:
I think we were not clear in describing what should be able to work here.

There are different cases:
1. Plotting a single variable (possibly for multiple stations). This is Maxwell's dialog and it will be the most common case of using the dialog.
2. Plotting multiple variables (possibly for multiple stations) using the "Multiple Variables" button of the receiver. In this case the data will get stacked for the user temporarily so that the multiple variables can be easily plotted together. In this case, the code should know to group by **variable**, because this column will be generated from the stacking.
3. Plotting multiple variables (possibly for multiple stations) using already stacked data. In this case, the user has already stacked their multiple variables into one column. This should be a "simple" case because the code doesn't need to do anything special. The user may have a **variable** column (or with a different name!) and they should put it in the appropriate receiver to get the graph they want.

Of these, number 1, Maxwell's dialog, is going to be by far the most common option for users. Most cases will be plotting a single summary on a graph, possibly faceted by station.
Number 2, is then the next most common, to get a more complicated graph with two or more summaries, maybe average min and max temperature. In this case the dialog needs to recognise the data will get stacked and new **variable**/**value** columns will be used.
Number 3 is very unlikely to be used. Most users will not stack their elements (which is why we have the multiple variables button). However, if they do it would be nice if the dialog could cope with the data in this format. But its the least essential. However, if the `group_by` is done by the colour variable and the facet variable, then I claim the dialog will also work in this case.

I think in our discussion there was confusion between options 1 and 3 and we were possibly talking about different ones to each other. 
We don't assume that the data will be stacked, and we definitely don't assume that certain columns exist in the data, because users choose and rename columns, even if stacking. We should be less worried about option 3 working if 1 and 2 work.

stevenndungu:
Great, thanks Danny and Maxwell for pointing it out.  I agree. 

Actually I had much thoughts on stacking as the second option that is why I added the `group_by(variable)`. I have cleared off the line and now the three options 1,2,3 should be working well unless if we still could have some errors. @maxwellfundi you might want to have a re-review.
Thanks.

maxwellfundi:
Great, looking at it soon.

maxwellfundi:
@stevenndungu  @dannyparsons  This is now working how i would expect it to work with no bugs. @rdstern is there a single time we would like all the lines at the same time in the same graph, I am finding the lines labels overlapping and i was wondering if this would be useful to use geom_label_repel

image

@dannyparsons  this is now ready for you

dannyparsons:
Great, thanks. Out of interest, how does it look with `geom_label_repel` instead for these graphs? i.e. just change this in the script.

maxwellfundi:
Ooops, i didnt try that option. I suggest we merge this then see if other options could be viable. we could write an issue on that @dannyparsons

instat/dlgPICSARainfall.vb
```diff
@@ -240,60 +243,62 @@ Public Class dlgPICSARainfall
...
-        clsCLimitsYContinuous.AddParameter("min", "0", bIncludeArgumentName:=False, iPosition:=0)
+        clsCLimitsYContinuous.AddParameter("min", "NA", bIncludeArgumentName:=False, iPosition:=0)
         clsCLimitsYContinuous.AddParameter("max", "NA", bIncludeArgumentName:=False, iPosition:=1)
 
         clsYScaleDateFunction = GgplotDefaults.clsYScaleDateFunction.Clone()
         clsYScaleDateFunction.AddParameter("date_labels", Chr(34) & "%d %b" & Chr(34), iPosition:=0)
+        clsYLimitsYDate.SetRCommand("c")
+
+        ' clsCLimitsYDate.AddParameter("min", "January", bIncludeArgumentName:=False, iPosition:=0)
+        'clsCLimitsYDate.AddParameter("max", "December", bIncludeArgumentName:=False, iPosition:=1)
```
dannyparsons: These lines can be removed.

instat/dlgPICSARainfall.vb
```diff
@@ -720,16 +735,27 @@ Public Class dlgPICSARainfall
...
         If clsBaseOperator.ContainsParameter("hlinemean") OrElse clsBaseOperator.ContainsParameter("hlinemedian") OrElse clsBaseOperator.ContainsParameter("hlinelowertercile") OrElse clsBaseOperator.ContainsParameter("hlineuppertercile") Then
-            If clsGroupByFunction.iParameterCount > 0 Then
-                clsPipeOperator.AddParameter("group_by", clsRFunctionParameter:=clsGroupByFunction, iPosition:=1)
+            clsPipeOperator.AddParameter("group_by", clsRFunctionParameter:=clsGroupByFunction, iPosition:=1)
+            If Not ucrVariablesAsFactorForPicsa.bSingleVariable Then
+                clsGroupByFunction.AddParameter("x", "variable", bIncludeArgumentName:=False, iPosition:=0)
             Else
-                clsPipeOperator.RemoveParameterByName("group_by")
+                If ucrReceiverColourBy.IsEmpty AndAlso Not ucrReceiverFacetBy.IsEmpty Then
+                    'clsGroupByFunction.AddParameter("x", ucrReceiverFacetBy.GetVariableNames(bWithQuotes:=False), bIncludeArgumentName:=False, iPosition:=0)
```
dannyparsons: Should there be something here? If not, this should be removed.

instat/dlgPICSARainfall.vb
```diff
@@ -720,16 +735,27 @@ Public Class dlgPICSARainfall
...
+                ElseIf (Not ucrReceiverColourBy.IsEmpty AndAlso Not ucrReceiverFacetBy.IsEmpty) Then
+                    clsGroupByFunction.AddParameter("x", ucrReceiverColourBy.GetVariableNames(bWithQuotes:=False), bIncludeArgumentName:=False, iPosition:=0)
+                ElseIf (Not ucrReceiverColourBy.IsEmpty AndAlso ucrReceiverFacetBy.IsEmpty) Then
+                    clsGroupByFunction.AddParameter("x", ucrReceiverColourBy.GetVariableNames(bWithQuotes:=False), bIncludeArgumentName:=False, iPosition:=0)
```
dannyparsons: A small point, a better parameter name convention for operators like this is to have the parameter name as `"0"`. Then it matches the position and it is easier to have the order needed.

instat/dlgPICSARainfall.vb
```diff
@@ -720,16 +735,27 @@ Public Class dlgPICSARainfall
...
         AddRemoveHLineCalculations()
     End Sub
 
     Private Sub AddRemoveHLineCalculations()
         If clsBaseOperator.ContainsParameter("hlinemean") OrElse clsBaseOperator.ContainsParameter("hlinemedian") OrElse clsBaseOperator.ContainsParameter("hlinelowertercile") OrElse clsBaseOperator.ContainsParameter("hlineuppertercile") Then
-            If clsGroupByFunction.iParameterCount > 0 Then
-                clsPipeOperator.AddParameter("group_by", clsRFunctionParameter:=clsGroupByFunction, iPosition:=1)
+            clsPipeOperator.AddParameter("group_by", clsRFunctionParameter:=clsGroupByFunction, iPosition:=1)
+            If Not ucrVariablesAsFactorForPicsa.bSingleVariable Then
+                clsGroupByFunction.AddParameter("x", "variable", bIncludeArgumentName:=False, iPosition:=0)
```
dannyparsons: Although the order doesn't really matter at this point, I think make this position `1` and parameter name `"1"`.

instat/sdgPICSARainfallGraph.vb
```diff
@@ -203,19 +207,22 @@ Public Class sdgPICSARainfallGraph
...
+        'ucrInputYSpecifyUpperLimitDate.Visible = False
 
         ucrPnlYAxisType.AddToLinkedControls(ucrChkSpecifyYAxisTickMarks, {rdoYNumeric}, bNewLinkedHideIfParameterMissing:=True)
         'TODO these two will not be only for numeric once limits for dates implemented
-        ucrPnlYAxisType.AddToLinkedControls(ucrChkYSpecifyLowerLimit, {rdoYNumeric}, bNewLinkedHideIfParameterMissing:=True)
-        ucrPnlYAxisType.AddToLinkedControls(ucrChkYSpecifyUpperLimit, {rdoYNumeric}, bNewLinkedHideIfParameterMissing:=True)
+        'ucrPnlYAxisType.AddToLinkedControls(ucrChkYSpecifyLowerLimit, {rdoYNumeric}, bNewLinkedHideIfParameterMissing:=True)
+        'ucrPnlYAxisType.AddToLinkedControls(ucrChkYSpecifyUpperLimit, {rdoYNumeric}, bNewLinkedHideIfParameterMissing:=True)
+        'ucrPnlYAxisType.AddToLinkedControls(ucrChkYSpecifyLowerLimit, {rdoYDate}, bNewLinkedHideIfParameterMissing:=True)
+        'ucrPnlYAxisType.AddToLinkedControls(ucrChkYSpecifyUpperLimit, {rdoYDate}, bNewLinkedHideIfParameterMissing:=True)
```
dannyparsons: These should be removed if not needed.

instat/dlgPICSARainfall.vb
```diff
@@ -302,22 +307,28 @@ Public Class dlgPICSARainfall
...
         clsGeomHlineMean.AddParameter("mapping", clsRFunctionParameter:=clsGeomHlineAesMean, iPosition:=0)
         clsGeomHlineMean.AddParameter("size", "1.5", iPosition:=2)
 
-        clsGeomHlineAesMean.SetPackageName("ggplot2")
+        clsGeomHlineAesMean.SetPackageName("ggplot2")
         clsGeomHlineAesMean.SetRCommand("aes")
         clsGeomHlineAesMean.AddParameter("yintercept", strMeanName, iPosition:=3)
 
         clsMeanFunction.SetRCommand("mean")
+        clsMeanFunction.AddParameter("x", "value")
```
dannyparsons: This doesn't seem right for the default? I would expect there not to be a default for this as it is linked a receiver or set manually somewhere else.

instat/dlgPICSARainfall.vb
stevenndungu: yes

stevenndungu:
Fixing the mentioned comments shortly.

instat/dlgPICSARainfall.vb
stevenndungu: This runs when the multiple receiver is used if I am not wrong. Let me confirm this.

dannyparsons: I think its in the `SetDefaults()`

stevenndungu: Yes it is on setdefaults of the main dialogue. I should transfer it to the sub dialogue? 
That is the two parameters, `x` and `na.rm`

dannyparsons: I don't see why there should be any default for `x`? Shouldn't this just be updated when changing between single and multiple?

rdstern:
I reply here to the point raised by @maxwellfundi above that when all the lines are labelled, the labels overlap.  And we could repel them.  If that is really simple, then fine.  But otherwise I would strongly discourage that option anyway.  To me:
a) Having a single labelled line - mean or median - will be frequent.  And is the main thing
b) Having (instead) the two terciles is also relevant in relation to the seasonal forecast.
It will be rare to have them together.
In the longer term I would expect possible requests for  more options, particularly:
c) Quartiles, instead of terciles.  More sensible and links with boxplots.  But does not tie with the current seasonal forecast.
d) A fixed line, e.g. 600mm to be allowed.

As Danny says the main use of these graphs will be for one (or more stations, with station as facet) and a single variable on each line.  Second will be 2 variables, particularly start and end of the season to be on the same graph.

So, I am very happy if these main aspects work well and for them to be merged without clever ddions on the labels.

There will (later) be one important addition and that is an extra (logical) layer to the graph to be able to indicate years when there was no planting opportunity - as in the logical column for the start and end of the season.  This may need to be done as part of the PICSA graph while it remains tricky to add layers when the special method for labelling with dates is used.

instat/dlgPICSARainfall.vb
stevenndungu: You are right. However, I was checking how the function is depended to the other functions in the default - and I think transferring implies lots of changes. A number of functions require it. It is not actually needed in the   changing between single and multiple. You might want to have a look.

stevenndungu: Note that there is similar cases for median, percentiles as well.

dannyparsons:
Thanks, I think this is a good time to merge and allow further testing in use.

PR #5320, opened 2019-05-03
Model dialogues (Allow appending of results and Disabling of Plots option)

### LINKED ISSUE #5307: Fitted values etc not working on some modelling dialogues
I think this is a general problem, so could use our "standard" survey data and fit any model.  I was using temperature data from Ghana.

I used the **Model > General > Fit Model**.  It works OK, but when I try to **save the fitted values** it gives an error.  It says it needs data of the right length.

The **Model > Two Variable > Fit Model** works fine (as far as I can see.  So copying what is there may be all that needs to be done to fix this.  If so, then please also include **Three Variable** (which I would like to use) and Four Variable at the same time.

(As usual I am finding these problems in writing the climatic guide.  I am now temporarily OK, because I can now use the 2-variable case, then put the correct model into the script.)

I am suggesting that either Steven or Lazarus could fix this?

stevenndungu:
fixes #5307

Note:
The problem was that when performing model fit, missing values were excluded implying that the prediction had reduced length relative to the size of the original data frame (appending becomes a problem). This gives the reason why this error happened in the Ghana temperature data set (it has many missing values). 

Similarly case to Three var Model fit but I will be solving it on a separate pull request since this results to another bug. I need to get a way around it.


stevenndungu:
Adding `na.action = na.exclude` correctly padded NAs so it was the right length which serves a a solution for appending the fitted, residual, etc  columns to the data frame across all the fit model dialogues. 

However, this brings up another problem the `autoplot` function which can only plot models without the padded `NA's` . Solving one problem results to another problem. [this happens to all the fit model dialogues as well -  two var model included]

A quick solution could be running `lm` or `glm` twice with and without it. Or consider Roger's suggestion of using alternative packages.

NB: This problem happens when handling data with missing values ONLY.

stevenndungu:
this is the error that autoplot function gives
image

dannyparsons:
Thanks, this sounds consistent with what we found before with @shadrackkibet. I can't find the previous discussion on this @shadrackkibet or @rdstern may be able to find it?

But I think the suggestion should be to keep `na.action = na.exclude` as this is "correct" for fitted values etc. The issue with `autoplot` is a bug in that function but it did not look like it will be resolved, and it would be complex to solve ourselves. I suggest we find an alternative way of getting these plots, using other functions or packages, or combining them using our own function.

dannyparsons:
So, should we disable the `autoplot` option for now until we find an alternative? I think it is more important to be able to deal with missing values than to be able to do those plots, if we can only do one currently?

shadrackkibet:
Our previous discussion can be found [here](https://github.com/africanmathsinitiative/R-Instat/pull/4961). Unfortunately, our posts on Rstudio and StackOverflow have no response yet.  I agree with @dannyparsons suggestion on using other functions to get these plots. Seems this is simple using `plot()` function by passing in model object and set `which` parameter to `1,2,3,4,5,6`.

dannyparsons:
Thanks @shadrackkibet. Using `plot` seems a good option for now.

stevenndungu:
@shadrackkibet  `plot()` is a a great option. We have to press **`ENTER`** in order to see subsequent plots which i think is okay. The challenge is capturing the plot to save  - tricky.  I have not found a nice way of saving it in the convention of R-Instat - unless we remove the option of saving where that will be OK.

@dannyparsons @rdstern I disabled the option temporarily. Is that okay or what do you propose? 
We can either disable the autoplot function controls or disable appending the fitted values, residuals etc or use plot but not allow saving the plots in the meanwhile.

dannyparsons:
Ok yes now I remember why `plot` is not easy to use because the way it works doesn't allow us to retrieve those easily in R-Instat.
Saving of fitted values, residuals is more important than the plot I think so the plot could be disabled for now.

stevenndungu:
Okay great. I did disable already and set the saving of fitted values, residuals,etc possible without the error we had before - pad in NA's. It should be ready for merging then

PR #5326, opened 2019-05-08
Enabling the Options Button for the Two Columns Option and other edits

### LINKED ISSUE #5255: Add options button to correlations when just 2 variables etc?

stevenndungu:
Fixes #5255
Taken up changes made by Lazarus.

Following this pull request I will be closing PR #5285 by Lazarus for efficiency.

@africanmathsinitiative/developers this is ready for review find more comments on the PR closed for proper and quick review.


stevenndungu:
@dannyparsons  this had several reviews. It was ready for you before I made the small changes from what Lazarus had done already.

instat/dlgCorrelation.vb
```diff
@@ -246,10 +245,11 @@ Public Class dlgCorrelation
         ElseIf rdoMultipleColumns.Checked Then
             ucrBase.clsRsyntax.SetBaseRFunction(clsCorrelationFunction)
         End If
+        sdgCorrPlot.BeforeAndAfterCodes()
```
dannyparsons: This might need a bit of restructuring. It's not good to call something on the sub dialog which relies on the controls being checked/enabled etc. because the sub dialog will not be in the correct state if it hasn't been opened. See other comments for an alternative.

instat/dlgCorrelation.vb
```diff
@@ -288,4 +279,28 @@ Public Class dlgCorrelation
...
+    End Sub
+
+    Private Sub CorrelationsEnable()
+        If rdoTwoColumns.Checked Then
+            sdgCorrPlot.rdoCorrelationPlot.Enabled = False
+            If sdgCorrPlot.rdoCorrelationPlot.Checked Then
+                sdgCorrPlot.rdoNone.Checked = True
+            End If
+        Else
+                sdgCorrPlot.rdoCorrelationPlot.Enabled = True
```
dannyparsons: Setting checked/enabled of sub dialog properties should only happen when the dialog opens. This code can run when the sub dialog opens. There can be a parameter in SetRCode of the sub dialog to say if it is two or multiple variables.
On the main dialog, I think all you need to do is remove the problem function in the two variable case `ucrBase.clsRsyntax.RemoveFromAfterCodes(clsRGGcorrGraphicsFunction)`

instat/sdgCorrPlot.vb
```diff
@@ -143,10 +144,34 @@ Public Class sdgCorrPlot
...
         If rdoCorrelationPlot.Checked Then
-            clsRsyntax.AddToAfterCodes(clsRGGcorrGraphicsFunction, iPosition:=1)
-            clsRsyntax.RemoveFromAfterCodes(clsRGraphicsFuction)
-            clsRsyntax.RemoveFromAfterCodes(clsRGGscatmatrixFunction)
+            If Not rdoCorrelationPlot.Enabled Then
+                clsRsyntax.RemoveFromAfterCodes(clsRGGcorrGraphicsFunction)
+            Else
+                clsRsyntax.AddToAfterCodes(clsRGGcorrGraphicsFunction, iPosition:=1)
+                clsRsyntax.RemoveFromAfterCodes(clsRGraphicsFuction)
+                clsRsyntax.RemoveFromAfterCodes(clsRGGscatmatrixFunction)
```
dannyparsons: You could simplify this code by removing all the codes outside the first `If` and then only adding them inside each `If` statement.

stevenndungu:
@africanmathsinitiative/developers  this is ready for review. Basically reviewing @dannyparsons 's comments above and general testing if there is any other bug probably available. Thanks

Ateamate:
@dannyparsons, this is ready for you to review. I haven't found any error on my side.

dannyparsons:
@Ateamate thank you for doing a review of this work, it's a great way to start getting involved in the project.
However, you did miss one bug which is in the dialog! 

Here's what I did to test it:

1. Choose multiple columns, put some columns in
2. Go to sub dialog
3. Choose "Correlation Plot"
4. Go back to main dialog and change to two variables, put two columns in
5. Run the dialog

This gives an error because it still tries to run the correlation plot that I chose for multiple variables, and the problem that was noticed was that it doesn't work for the two variable case.
If I went back to the subdialog between 4. and 5. it would have worked, because Steven had implemented a check when you open the subdialog for it to go back to "None". But here I wanted to check if it would work if I didn't open the subdialog again.

This is the level of detail that's needed in testing. It is not a trivial task because there are many possibilities of what a user can do on a dialog, using different combinations and in different orders, and we need to consider all of them. Actually, when you're testing, instead of trying to see if it works, you should be trying to break it, trying things that might not have been anticipated to see if that causes a bug. Looking at the comments and the original issue will help you to know what to target in testing by understanding the original problem well.
Testing a dialog really well can take a long time because there's lots of possible places for errors that you can check. But if you can gain the skills to do this well then its really important because it makes the whole software more stable and its so much better if we can detects problems here and fix them before they get to any user. Thanks for getting involved and please keep up the testing!

PR #5337, opened 2019-05-15
Setting TestOk and Adding ucrNudThreshold to the outlier check box

### LINKED ISSUE #5276: Change in Climatic > Check > Rainfall QC
I have been impressed at how the rainfall and temperature dialogues have worked!
In rainfall, one option is called Consecutive.  It notes whenever there are more than x (default 10) consecutive rain days.
One edit that would be useful, and just may be a small change - I don't know.  It is that the logical is set to true, only from the 10th onwards.  It would be good for it to be true for all days in that sequence, so we see the whole thing?
A second change is likely to take a bit longer.  This would add a threshold to the dialogue for that option.  

But the first is the only urgent one.

Wycklife:
Fixes #5276

This fixes the second and last part of the issue.


shadrackkibet:
@Wycklife what about the other parts of the issue?

PR #5338, opened 2019-05-15
Polar

### LINKED ISSUE #5331: Add polar to coord tab in plotting
In our plots we currently have a special facility for wind roses.  It would be good to be able to do other (simpler) circular plots.

I have finally realised there is no point in looking for special facilities in another package.  We can easily get sensible plots with our current system.  The one item we need to add is a 3rd checkbox on the Coordinates tab.  The label could be:
 Use polar coordinates.  The default is to give coord_polar(start = 0)
There could be an option called Starting Angle.  This has an up-down from 0 to 2 in steps of 0.1. After the up-down give * pi.  If 0 (default, then it is given as 0.  Otherwise it changes start = value * pi) (That's quite nice and clear!
There is also a checkbox, default unchecked, with the label Direction Anticlockwise.  If unchecked (default) it isn't given.  If checked, then add direction = -1 to the coord_polar command.

Here is an example of what we can do now - currently with the script from a bar chart or histogram etc.
image

The other "trick" that I assume we can do now already is to be able to set the ylim.  The "trick" is ylim(-500,500) in the graph above.  We might have to enable this sort of feature in the y-axis for discrete.  Let's see that later.

Ivanluv:
fixes #5331 
@dannyparsons  does this look fine?


dannyparsons:
Can you send an image of the new tab?

Ivanluv:
image

dannyparsons:
That looks good, apart from the checkboxes are not inside the groupbox, they can move to the right slightly. I would also make the textbox a bit smaller since its going to be a small number and then its closer to the `pi`.

rdstern:
I agree with the above.  Could there also be 0 (zero) as the default there?

Ivanluv:
making the changes now

instat/sdgPlots.vb
```diff
@@ -727,4 +733,24 @@ Public Class sdgPlots
...
+
+    Private Sub ucrChkIncludeFacets_CheckedChanged(ucrChangedControl As ucrCore) Handles ucrChkIncludeFacets.ControlValueChanged, ucr2ndFactorReceiver.ControlValueChanged, ucr1stFactorReceiver.ControlValueChanged
+
+    End Sub
+
+    Private Sub ucrChkFreeSpace_CheckedChanged(ucrChangedControl As ucrCore) Handles ucrChkFreeSpace.ControlValueChanged
+
+    End Sub
+
+    Private Sub chkScales_CheckedChanged(ucrChangedControl As ucrCore) Handles ucrChkFreeScalesY.ControlValueChanged, ucrChkFreeScalesX.ControlValueChanged
```
shadrackkibet: Why all these empty subs?

maxwellfundi:
@dannyparsons @Ivanluv  Could we use an NUD other than a text box, i presume whatever we need there is a numeric value.

maxwellfundi:

Ivanluv:
@maxwellfundi  yes,it is.

maxwellfundi:
Great testing it now.

maxwellfundi:
@ivan I have a few issues here

1. After running the dialog(all options are working when I check each of them, the direction and the angle).  On doing a reset on the main dialog, all the controls relating to your polar coordinates do not reset. 

2. The text box seems a little big, I think we will not have anything more than three figure- max is 360? So it could be reduced significantly. 

3. When I go to the subdialog and check the start angle and not have anything on the text box, this works because `start` parameter is not added. However, when I go back to the subdialog, with the start angle checked and the text box previously filled in and remove the values in the text box, then `start` is passed in but without a value - `start= * pi` which bugs.

when these are fixed this will be ready for you @dannyparsons

dannyparsons:
Since the text box might have different decimal places e.g. 0.5*pi or 0.25*pi I don't think this will be easy to do with an NUD. NUDs are good for a fixed range and fixed increment, but we don't have an obvious fixed increment here.

maxwellfundi:
@dannyparsons okay that now well understood

instat/sdgPlots.vb
```diff
@@ -80,19 +78,12 @@ Public Class sdgPlots
...
         ucrChkUsePolarCoordinates.SetText("Use Polar Coordinates")
-
-        ucrChkDirectionAnticlockwise.Hide()
-        ucrChkStartingAngle.Hide()
-        ucrtxtStartingAngle.Hide()
-        lblPi.Hide()
-
+        ucrChkUsePolarCoordinates.AddToLinkedControls({ucrChkDirectionAnticlockwise, ucrChkStartingAngle}, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrChkUsePolarCoordinates.AddParameterPresentCondition(True, "coord_polar", True)
+        ucrChkDirectionAnticlockwise.AddParameterPresentCondition(True, "coord_polar", True)
```
maxwellfundi: @Ivanluv I think this `ucrChkDirectionAnticlockwise` should be checking if the `direction` parameter and be checked when the `direction= -1`

Ivanluv: @maxwellfundi  I've made the change

instat/sdgPlots.vb
```diff
@@ -72,6 +75,22 @@ Public Class sdgPlots
...
 
         ucrBaseSubdialog.iHelpTopicID = 136
+
+        'Use Polar Coordinates
+        ucrChkUsePolarCoordinates.SetText("Use Polar Coordinates")
+        ucrChkUsePolarCoordinates.AddToLinkedControls({ucrChkDirectionAnticlockwise, ucrChkStartingAngle}, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrChkUsePolarCoordinates.AddParameterPresentCondition(True, "coord_polar", True)
+        ucrChkDirectionAnticlockwise.AddParameterPresentCondition(True, "direction", True)
+        ucrChkStartingAngle.AddParameterPresentCondition(True, "start", True)
+        ucrtxtStartingAngle.AddParameterPresentCondition(True, "start", True)
```
maxwellfundi: @Ivanluv  we dont need this condition for the input text box.

instat/sdgPlots.vb
```diff
@@ -727,4 +759,40 @@ Public Class sdgPlots
...
     End Sub
+
+    Private Sub ucrChkUsePolarCoordinates_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkUsePolarCoordinates.ControlValueChanged
+
+
+        clsCoordPolarFunc.SetPackageName("ggplot2")
+        clsCoordPolarFunc.SetRCommand("coord_polar")
+
+        If ucrChkUsePolarCoordinates.Checked Then
+            clsBaseOperator.AddParameter("coord_polar", clsRFunctionParameter:=clsCoordPolarFunc, bIncludeArgumentName:=False)
```
maxwellfundi: @Ivanluv I was testing this with different graphics dialog, with the barchart dialog, I was having trouble with the coordpolar- When i use your coord polar after creating the bar charts, then the pie chart is automatically checked because of we use `coord_polar `to create the pie chart. I suggest you change this parameter name- to say = `coord_polars` to avoid this confusion.

maxwellfundi:
@Ivanluv  this looks good now just one thing- The start when you type in a certain angle, resets on reopen. 

So if you have **20.78 * pi**   suggest to solve this issue, you create an operator with the * as operation with the two parameters on the right and left, with no argument names passed, then have the start input setrcode to the operator.

Ivanluv:
@maxwellfundi  this last commit solves the error of reset on reopen.However when I reset from the main dialogue,the start angle check box fails to reset.Any ideas on this?

maxwellfundi:
Okay great, will have a look here, how you have solved the issue of reset is not quite right, I think if I have time on monday we can have a look at using the operator instead, that way we will be having it work better and the issues will be resolved.

Ivanluv:
@maxwellfundi  could you review this please

shadrackkibet:
@maxwellfundi ?

rdstern:
I have had a look at this branch.  I also very much appreciate the care that has been taken in the testing of this feature, including features I didn't think to look at!  I tried boxplots by month and this all worked well. 

Looking at the above I tried bar charts, by month.  Here is an example:
![image](https://user-images.githubusercontent.com/11226469/62702453-ad20d880-b9de-11e9-9041-224f9febb147.png

I now find I can't do facets?  I hope that is something else and not linked to the changes made here?

Apart from that, I would be very happy to have this merged.

maxwellfundi:
@rdstern  I dont see how these changes could have affected the facets. @Ivanluv  with this branch, are you able to do facets?

rdstern:
I didn't understand that aspect either and it may be something else I did.  The non-availability of facets has appeared for me occasionally and I haven't been able to pin down why!

I would really like to have polar in the next version.

Ivanluv:
@maxwellfundi  it does work from my branch but when i try adding free scale to the facets ,I get an  error  that  coord polar does not support free scale

maxwellfundi:
Thats interesting. i wonder how we can restrict users from getting into this trap.

maxwellfundi:
@rdstern  I have tested this and the polar works well.  Actually and it supports facets, just some options on the facets that are not supported. I think the message is informative enough that people can understand that the options are not supported as shown 

image


@dannyparsons this looks good! i think this can be merged too.

rdstern:
Let's leave the last word with @dannyparsons but I would be very happy to see this merged.

instat/sdgPlots.vb
```diff
@@ -307,6 +333,14 @@ Public Class sdgPlots
             strDataFrame = ucrBaseSelector.strCurrentDataFrame
             ucrFacetSelector.SetDataframe(strDataFrame, False)
         End If
+
+
+        If bReset Then
+            clsCoordPolarFunc = New RFunction
```
dannyparsons: As a general rule, if you find yourself doing something in a different way just for one piece of code, when there's no real difference with how it works compared to other parts of code, then that should be a warning message to you that you may be missing something in understanding why the code works in the way it does.

In this case `clsCoordPolarFunc` should be no different to the other `ggplot` functions such as `clsLabsFunction` or `clsFacetFunction` in that they all add to the ggplot operator. So you should be questioning why you're treating it differently in the code.

Here, all the `ggplot` functions must be passed through from the main dialog to `SetRCode` otherwise it will not work correctly when using it from different main dialogs. `clsCoordPolarFunc` is no different so should follow this implementation.

@maxwellfundi being more familiar with how this sub dialog works I expected you would have picked this up from reviewing. Reviews need to look at the coding structures not just how it works from a user perspective. Can you support @Ivanluv to do this through `SetRCode`? It will need changes on main dialogs, so check other PRs to avoid conflicts.

instat/sdgPlots.vb
```diff
@@ -727,4 +767,24 @@ Public Class sdgPlots
         sdgThemesSub.ShowDialog()
         SetRcodeForCommonThemesControls(False)
     End Sub
+
+    Private Sub ucrChkUsePolarCoordinates_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkUsePolarCoordinates.ControlValueChanged
+
+
+        clsCoordPolarFunc.SetPackageName("ggplot2")
+        clsCoordPolarFunc.SetRCommand("coord_polar")
```
dannyparsons: These lines don't need to be run here each time. Once implemented through `SetRCode` this won't be needed.

Issue #1908, opened 2016-10-14
Wind roses


rdstern:
I would now like to catch up with the current state of the windrose dialogue and plot.  This also linkis to the possible packages we could be using.  Maxwell, perhaps you have already?  I am also copying Steve Kogo on this topic, because I raise some general issues.

The general issue is that Met Services seem to be obsessed by wind roses.  So it would be good to "nail" that area by the end of November.  And it should be quite easy.
However, wind roses are (usually) to plot wind direction and wind speed together.  There are also:

a) Other types of data where the same type of plot would apply.  For example extreme rainfall and date at which it occurs.  Time of day for rainfall starting together with the amount.

b) The wind rose is also special in plotting a circular variable (wind direction) together with another numerical variable.  It would be good to also have other graphs for circular data, e.g. to plot just the wind direction.

There are 6 packages I have found with procedures for wind roses.  Most also have other types of graph for circular data too.  I list them below, because they are also interesting for other parts of R-Instat.

**1) clifro.**  I list this one first, because it is the only package I have found that uses ggplot2 for the wind roses.  It is designed mainly to make it easy to extract and use New Zealand Met data.  It would be good to explore the package more.  It could have other routines that are of interest, and also could give ideas for our future work in relation to climsoft. I have not been able so easily to run the examples in clifro. 
More in ggplot2.  Before going to other packages I found this reference useful.  http://rstudio-pubs-static.s3.amazonaws.com/3369_998f8b2d788e4a0384ae565c4280aa47.html 
It shows a sensible histogram using polar coordinates in ggplot2.  It then uses the circular package for more plots.  But it raises the question for me on whether we could (at least initially) stay with just ggplot2 graphs?

**openair.** This is a package for air pollution data.  It does not use ggplot, but does have a facility for facets. 
**plotrix.**  This has lots of plotting, but not using ggplot2.  It was an early contender, because it has an example of the wind rose in the way done by the Australian Met Service.
**oce.**  This is a huge package and for oceanographic data.  It has a vignette and claims to be also for teaching.
**circular** This has not been updated since early 2015 and is not much approved of by CRAN.  But it has a range of circular plots.  We will (I think) have to make use of this package for other aspects of analysing circular data - though perhaps not before the end of November.
However this package does have other types of graph for circular data that goes round a circle.  I wonder how easily they can be done in ggplot2?
**CircStats** This may be one we don't need to use.  It is the other package on circular statistics.

[...]

PR #5344, opened 2019-05-20
Enhancing Windrose

maxwellfundi:
@rdstern I have fixed your issues on windrose except the color pallete one that seems to be taking bits of time. 
I am looking more onto it though. 

@africanmathsinitiative/developers this is ready for review for the #1908 issues that have been checked on the last comment on that issue by @rdstern


stevenndungu:
reviewing this one

maxwellfundi:
Great! Thanks @stevenndungu

dannyparsons:
This looks good except you should be defining the speed cuts function on the main dialog, not the subdialog. Otherwise it doesn't exist outside the subdialog which is why you had to do the resetting in the subdialog. If you define it on the main dialog, pass it through to the sub dialog in SetRFunction, then everything will work in the same way as the wind rose function. The resetting with happen through setting the defaults on the main dialog.

maxwellfundi:
@dannyparsons the issue is now fixed here

dannyparsons:
Great, that looks much cleaner right!

maxwellfundi:
Cool!

Issue #5302, opened 2019-04-13
Factor columns sometimes not appearing in selector


dannyparsons:
**Describe the bug**
I haven't been able to reproduce this but noticed several times that the selector sometimes doesn't display the factor columns. I've only noticed this after running several dialogs first. Strangely if you convert a factor column to factor it then reappears in the selector. I'm not sure if it effects all dialogs or just some or its related to certain kinds of data files.

**To Reproduce**
@rdstern has also noticed this and may be able to be more specific.

If someone can first get a reproduceable example that would help greatly in resolving this.

Ideally someone can solve it too but getting a reproduceable example first will really help in getting to the cause of this.
@africanmathsinitiative/developers this is quite a serious problem so it should be a priority to resolve.

rdstern:
I think the problem is with the **Climatic > Dates > Infill Missing Dates** dialogue.
I opened the Instat data climatic, ghana.
a) If I go directly to the Define climatic data the station is fine.  It always appears.
b) If I then go to the Infill missing - there are some to be infilled - and then return it seems ok.  But if I press reset it then can't find the station.

I wonder if the infill has changed the type of the station to character and this has not changed in the grid.  If so, it would explain why the station is suddently not recognised.

Wycklife:
@rdstern I don't seems to be getting the same problem. If I go to the infill dialog, only Station appears in the selector as a factor in the By factor receiver. When I reset the dialog, the station disappears from the factor column receiver but present in the selector. When I go back to the original dialog(Define Climatic data). the station is still present. Please could we discuss this?

Wycklife:
Another observation is that, after resetting the dialog, the station disappears also from the metadata.

rdstern:
Oh good.  This should make the discovery of the bug easier, noting that the problem is actually within the Infill dialogue.  Can you see where the R code in that dialogue is changing the station?  I assume it is then changed into a character variable and is therefore no longer a factor.  So possibly simply adding an as.factor to that expression may be needed?

shadrackkibet:
This Bug is a serious one and also disturbing. It is worth someone spending time to look into it.This might be related to issue #4127 . I will also try to replicate it next week if i get some free time.

Wycklife:
I have added as.factor to the code on the script window but it produces an error.
data_book$infill_missing_dates(data_name="ghana", date_name="date", factors=as.factor ("station"))
[...]

PR #5346, opened 2019-05-27
Delete Rows - factors

### LINKED ISSUE #5325: Delete rows has a problem with factors
I hope this will be an easy one to fix.  It is the problem we had with another dialogue that once used, then a factor isn't recognised any more.

[rr_amj_1981-2014_2.txt](https://github.com/africanmathsinitiative/R-Instat/files/3156869/rr_amj_1981-2014_2.txt)

Here is a data file.  Without the "-2" it can also go into the library in the Climatic > Ghana directory.

Once imported, the first column - called STN comes in as a factor.  That's because of the first 2 rows of data.

Now right-click and delete the 2 rows.  Now the column isn't recognised as a factor anymore.  That caused me some problems in the later analysis. 

If you right-click and make it a factor again all seems ok.

stevenndungu:
fixes  #5325
Whenever the rows were deleted. The factor columns would lose the variable names in the metadata - to NA. They would therefore disappear (factor columns).
@africanmathsinitiative/developers ready for review

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -948,6 +948,7 @@ DataSheet$set("public", "remove_rows_in_data", function(row_names) {
   #since it removes column attributes
   self$set_data(dplyr::slice(curr_data, -rows_to_remove))
   self$append_to_changes(list(Removed_row, row_names))
+  self$add_defaults_variables_metadata(self$get_column_names())
   self$data_changed <- TRUE
 }
 )


dannyparsons:
@stevenndungu good that you spotted this. This also happens when rows are added. So we need to do something similar when inserting rows and in the infill missing dates function.
I'm not sure why this is now happening, I don't think it was before, unless we have only just noticed it. Is it because of a change in `dplyr` functions like `slice`?

dannyparsons:
I've noticed that `dplyr::slice` has a new parameter `.preserve`, which preserves the grouping structure. Not sure if this is related, but try `.preserve = TRUE` and see if that has any difference on the metadata.

Also, do you understand why only factor columns are having this issue?

stevenndungu:
Great, I am not sure if it was there before. It is actually a strange behaviour... I have tried to find out why but not found the precise reason why factors only. If I use `dplyr::slice` on R the behaviour can not be traced.
I have added `.preserve = TRUE`, good addition but does not solve the problem the factors still disappear. 

"Do you understand why only factor columns are having this issue? " Not sure yet - under investigation

looking for the addition of rows and infilling cases as well. Thanks

dannyparsons:
Ok thanks. I think then add the same fix for the other two cases because we know that solves it, but add a comment that this needs to be understood as it wasn't caused by any code changes from us.

stevenndungu:
Ok fine. let me add.

dannyparsons:
Thanks @africanmathsinitiative/developers someone could test this now to confirm it fixes the issue.

stevenndungu:
@dannyparsons , I have noticed that this fixes #5302. It is actually the same issue.

rdstern:
In the testing I have noticed in the version before fixing that if I a) make a factor column, b) Add or delete a row - so it isn't a factor anymore, and then use right-click and the Levels/Labels dialogue that's still ok - even though the column isn't a factor anymore.  I assume that's because the (f) labels are still there and that is probably what is tested by this dialogue - at least when you get there through right-click.  So that's not a good test for the fix!

If, however, you try Prepare > Column: Reshape > Column Summaries, then in my version (with no fix) it isn't recognised as a factor anymore.  So that dialogue is a better test.

dannyparsons:
@africanmathsinitiative/developers this is important to review quickly as its a major bug currently.

stevenndungu:
@Wycklife is reviewing

Wycklife:
I have reviewed this and it solve the two issues #5325, i.e when I delete the two rows the Station still remain as a factor which is what is required. Also on #5302, The Station seems to appear as a factor even after resetting the dialog. All these seems to fix the issues from my end.

PR #5353, opened 2019-05-28
Adding library(circular) function to before code

Wycklife:
Fixes#5336


dannyparsons:
I'm happy with this solution of loading the package, but the `circular` package only needs to be loaded for the circular distributions, otherwise the code will look confusing if I use a non circular distribution is used.

So instead of adding the library() function to the before codes in the defaults, you can just check when the distribution changes, if its circular, add it, if not, remove it.

Wycklife:
@dannyparsons The condition has been set.

Wycklife:
@dannyparsons Changes has been made.

dannyparsons:
Great. Have you checked all the other distributions too. I wonder if any others depend on a package?

Wycklife:
Yes, other distribution are working well without the library(), 
@stevenndungu Could you please review this.

rdstern:
Good.  Please check the Extreme Value distribution.  That was another one that doesn't work currently.  That's in the extRemes package.  It may also have been fixed?

stevenndungu:
Extreme value distribution has an error and apparently adding package extRemes does not solve the problem. The error is;
image


Hypergeometric distribution has this error.
image

The rest i find them to work work well

rdstern:
The error in the extreme value distribution is odd.  An example of the command line is:

last_graph <- mosaic::qdist(dist="evd", p=c(0.5), shape=0, scale=0.5, loc=0)
So where has the code c(0,0.001, 0.999,1) come from?  Is it within the qdist code, or did we produce it?
Because I don't understand why we need that.  And it is the 0 and 1 in that line that has given the error.

maxwellfundi:
@dannyparsons @Wycklife I am suggesting that, we have then temporarily hide the distributions that have issues if we can not find a quick fix for now- it will just be the Extreme value, which needed better defaults and hypergometric. The von Mises is okay. 

@rdstern would you be happy with this?

rdstern:
ok by me.  I am disappointed with the extreme value distribution not working, but looked briefly and can see what's wrong, but not where it comes from!

maxwellfundi:
great @Wycklife  If you could look at leving out these two on the distributions list would great!

Wycklife:
@maxwellfundi I'm looking at it now

dannyparsons:
Ok I'm happy with that

PR #5354, opened 2019-05-28
Picsa Rainfall dlg - Minor fixes 

### LINKED ISSUE #5016: PICSA Rainfall dates labels on Y axis have issues with facets
**Describe the bug**
When you create a PICSA Rainafall graph with an one of the options of format as date selected then later add facets throws an error shown below. (sometimes with a swap of order then it works)

**To Reproduce**
1. Go to Climatic >PICSA Rainfall Graph
2. Fill in the all the receivers 
3. Click Ok to produce the graph
4. Go back to PICSA rainfall main dialog, go to PICSA Options and select the format as Date options and select Day Month(1 jan) and Click return
5. Click Okay
6. Go back to PICSA rainfall main dialog then click on Plot options and fill in a facet and click return 
7. Then click OK

**Expected behaviour**
The y axis for the facetted graphs should be in date format selected in PICSA Options subdialog format as date option. 

**Screenshots**
If applicable, add screenshots to help explain your problem, particularly of error messages.
image

**R script or commands**

`# Code generated by the dialog, PICSA Rainfall Graphs
Rwanda_daily_RR_Temp1981_2018_by_Station_Name_year <- data_book$get_data_frame(data_name="Rwanda_daily_RR_Temp1981_2018_by_Station_Name_year")
last_graph <- ggplot2::ggplot(data=Rwanda_daily_RR_Temp1981_2018_by_Station_Name_year, mapping=ggplot2::aes(x=as.numeric(x=year), y=as.Date(x=sum_PRECIP, origin="2015-12-31"), colour=Station_Name)) + ggplot2::geom_line(size=0.8) + theme_grey() + ggplot2::theme(panel.background=ggplot2::element_rect(colour="white", fill="white", size=0.5, linetype="solid"), panel.grid.major=ggplot2::element_line(colour="lightblue", linetype="longdash", size=0), panel.grid.minor=ggplot2::element_line(colour="lightblue", linetype="longdash", size=0), panel.border=ggplot2::element_rect(colour="black", fill="NA", size=0), axis.text.x=ggplot2::element_text(angle=90), axis.title.x=ggplot2::element_text(size=14), axis.title.y=ggplot2::element_text(size=14), title=ggplot2::element_text(size=20), plot.subtitle=ggplot2::element_text(size=15), plot.caption=ggplot2::element_text(size=8)) + ggplot2::geom_point(size=3, colour="red") + ggplot2::xlab(label="") + ggplot2::ylab(label="") + ggplot2::scale_y_date(date_label="%d %b") + ggplot2::scale_y_continuous(breaks=seq(from=, to=, by=)) + facet_wrap(facets= ~ Station_Name, dir="h")
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="Rwanda_daily_RR_Temp1981_2018_by_Station_Name_year")
data_book$get_graphs(data_name="Rwanda_daily_RR_Temp1981_2018_by_Station_Name_year", graph_name="last_graph")`

### LINKED ISSUE #5309: The Climatic > PICSA > Rainfall graphs again
image

This is a multiple graph, i.e. there are 2 variables plotted, showing the start and the end of the rains together. (It is with the Samaru - Nigeria data from the library.  The steps to get here are in the draft climatic guide.)

It isn't quite what I would like, because the y-axis doesn't show dates - it is numeric only.  I can get the dates and the line(s).  Here they are:
image

You notice the single line, rather than the separate line for the 2 means.  If that can be fixed, then I have the graph I need.  So that is one request.  David says that might be reasonably easy.

When, with the above graph, I return to the dialogue and click on Single and then on **Plot Options** I get this message:

**instat
Developer error: More than one state of control ucrChkExpand satisfies it's condition. Cannot determine how to set the control from the RCode. Modify conditions so that only one state can satisfy its conditions.
OK**   

Then it did let me into the sub-dialogue.  I chose Themes  and not to display the Legend and that works.

I then tried adding Hline as additional layers, having first established the mean of the start and end myself.  I can get it to display if I don't have dates as the Y-axis, but not if I do.  
I also tried (with dates) adding the end as a second layer and it says I need to give a date origin.  Adding layers is not easy, when dates are displayed.

I realise @stevenndungu is working on #5120, (also on this dialogue) but have made these separate, because they seem a bit different.  And possibly harder to fix?

### LINKED ISSUE #5351: PICSA Rainfall Graph Issues
This is working much better. A couple of issues from testing the newest version. These are in order of importance roughly.

1. When I do facets, and add the mean line, I still get the overall mean, not the mean for each group. The `group_by` is not added in the code. It only gets added when colour is specified.
2. When I add a mean line the display on the label is not rounded. Here's the part of the code. So the `round()` part is not quite right. It should just be `round(.mean_y)`. Looks like its just being added twice in different ways.
`ggplot2::geom_label(mapping=ggplot2::aes(x=-Inf, y=.mean_y, label=paste("Mean:", round(.mean_y, x=mean(x=start_doy, na.rm=TRUE))))`
3. If I only specify the upper limit for dates and not the lower limit, then nothing gets added to the code. I should be able to specify one without the other.
4. When I open the sub dialog for the first time and check the date limits it looks like this. I'm not sure how someone is supposed to know what to type in the boxes that have `NA` currently, either there needs to be a label that says `Day`, but the `NA` is still confusing so better would be a default of `1` then I think its clear without a label. If I leave it as `NA` then it puts the date as "2016/January/NA" so something is wrong there.
image.
I've just seen its using the same textbox for the numeric and date upper/lower limits, which now makes sense why `NA` appears there. I suggest the code would be much easier if there were separate textboxes for numeric and dates, because they do different things.
5. Minor issue, it accepts any number for the day, so I can put "56 January" and it doesn't complain. It doesn't give an error but the date will comes out as `NA`, so ignored.
6. Very minor issue, the parameters are in the wrong order in `as.Date` here: `ggplot2::scale_y_date(date_labels="%d %b", limits=as.Date(format="%Y/%B/%d", x=c("2016/February/1", "2016/September/30")))`

stevenndungu:
fixes #5351 and fixes #5309 and fixes #5016


instat/dlgPICSARainfall.vb
```diff
@@ -746,16 +749,18 @@ Public Class dlgPICSARainfall
             If Not ucrVariablesAsFactorForPicsa.bSingleVariable Then
                 clsGroupByFunction.AddParameter("0", "variable", bIncludeArgumentName:=False, iPosition:=0)
             Else
-                If (Not ucrReceiverColourBy.IsEmpty AndAlso Not ucrReceiverFacetBy.IsEmpty) Then
-                    clsGroupByFunction.AddParameter("0", ucrReceiverColourBy.GetVariableNames(bWithQuotes:=False), bIncludeArgumentName:=False, iPosition:=0)
+                If (ucrReceiverColourBy.IsEmpty AndAlso Not ucrReceiverFacetBy.IsEmpty) Then
+                    clsGroupByFunction.AddParameter("0", ucrReceiverFacetBy.GetVariableNames(bWithQuotes:=False), bIncludeArgumentName:=False, iPosition:=0)
+                ElseIf (Not ucrReceiverColourBy.IsEmpty AndAlso Not ucrReceiverFacetBy.IsEmpty) Then
+                    clsGroupByFunction.AddParameter("0", ucrReceiverColourBy.GetVariableNames(bWithQuotes:=False), bIncludeArgumentName:=False, iPosition:=1)
```
dannyparsons: This could be simplified and then easier to read. 
Have one If to check the colour receiver. 
A separate If to check the facet receiver.  
Then check the number of parameters in the clsGroupByFunction and add/remove from the pipe if 0 or not.
And fix the order of the parameters, I suggest facet is "0" and colour is "1". The order doesn't really matter but the code is simpler if its consistent.

stevenndungu: Sure. That will be more easier to read.

dannyparsons: Or you can check if at least one of the receivers is non empty then the grouping gets added.

instat/dlgPICSARainfall.vb
```diff
@@ -749,16 +749,35 @@ Public Class dlgPICSARainfall
...
+                    Else
+                        clsGroupByFunction.RemoveParameterByName("1")
+                    End If
+
+                    If Not ucrReceiverFacetBy.IsEmpty Then
+                        clsGroupByFunction.AddParameter("0", ucrReceiverFacetBy.GetVariableNames(bWithQuotes:=False), bIncludeArgumentName:=False, iPosition:=0)
+                    Else
+                        clsGroupByFunction.RemoveParameterByName("0")
+                    End If
+                ElseIf Not (ucrReceiverColourBy.IsEmpty OrElse ucrReceiverFacetBy.IsEmpty) Then
```
dannyparsons: This is still very confused. Your first If: `If (ucrReceiverColourBy.IsEmpty OrElse ucrReceiverFacetBy.IsEmpty)` is not right, if this is only false when both receivers are filled - not what you want.
Then on this line, you shouldn't need to check both receivers together, you have the individual checks above and that's enough.

stevenndungu: @dannyparsons  this is quite confusing trying to split it up as you suggested since there are alot of interdependence of the parameters and content to be passed when the three receivers are empty and non-empty. I was using permutation sort of and the combined if statement I had at first was quite fair for me.

Check what I have adapted to - fairly readable now.

stevenndungu: @Wycklife kindly help in reviewing this.

stevenndungu:
@africanmathsinitiative/developers this is ready for review. kindly check if I missed anything or if there are still other issues to be addressed

Wycklife:
This works well from my end. It can be merged.

instat/dlgPICSARainfall.vb
Patowhiz: I agree with @dannyparsons I think individual checks will make the code be easier to read and will achieve the same intention or probably I'm missing something.

stevenndungu: Sorry, for this case we cannot do individual  checks per receiver. This  will not achieve the objective otherwise I would  have done the changes. 

From my side those that is what I have thought of adjusting. You might want to make the changes and I will be okay to merge the changes in my branch. Thanks

dannyparsons: @stevenndungu I can see the complications as different parts of the code have got mixed with each other before you started working on this. You have done well to get some of this mess to work, but sometimes when we find that it can only work in a messy way, it means we need to re-look the code overall and consider if we can restructure it so that we can use less messy options. In this case, the adding of `mutate` and `group_by` functions have got mixed up with each other and they don't need to be. Separating these will help reduce complexity for us. From looking at the code I have some suggestions which will get us to much better code that we can work with going forward:

1. The inclusion of `clsMutateFunction` in the pipe is dependant only upon the checkboxes for lines in the sub dialog. Therefore, I don't know why the adding/removing of this from the pipe is not done on the subdialog. I must have not thought of this originally. I suggest that is moved to the subdialog and this will simplify this code since mutate is independent of the `group_by`.
2. Then, here on the main dialog, combine the code from `AddRemoveGroupBy()` and `AddRemoveHLineCalculations()` since they are both working with `group_by` and so they should be together to simplify. Now because of 1. your first check is whether `mutate` is in the pipe. If its not, ignore the `group_by`.
3. If `mutate` is there, when do we need `group_by`? We need `group_by` if any of the following are true:
a. we have multiple variables
b. we have a colour variable
c. we have a facet variable(s)
The confusion comes because we can have combinations of these. But I claim we can check these three cases independently of each other.
a. If we have multiple variables then add "variable" to `group_by`
b. If we have a colour variable then add it to `group_by`, **if it's not already in the group_by**, (because "variable" might be the colour).
c. If we have a facet variable(s) then add them to the `group_by`, **if they are not already in the group_by**, (because you can use the same variable for facet and colour). You need to use the method in `AddRemoveGroupBy()` to check what the facet variables are, so you actually don't need to check the facet receiver.
So these can be three separate If statements and then its all simple.
4. Finally, check if `group_by` has any parameters. If yes add it to the pipe, if not remove it.

I think this then simplifies most of these issues and the code will be much shorter and cleaner.
If this isn't clear let me know, as its not always easy to express suggestions like this in comments, and I'll make a suggested change to your branch if you're not sure about this.

stevenndungu: @dannyparsons  I have addressed the suggestions above with one thing remaining

"If we have a facet variable(s) then add them to the group_by, if they are not already in the group_by, (because you can use the same variable for facet and colour). You need to use the method in AddRemoveGroupBy() to check what the facet variables are, so you actually don't need to check the facet receiver."
Avoiding `group_by(variable,variable)` when the facet and colour receivers has same variable name.

instat/dlgPICSARainfall.vb
```diff
@@ -715,68 +714,54 @@ Public Class dlgPICSARainfall
...
+        If clsPipeOperator.ContainsParameter("mutate") Then
+            clsGroupByFunction.ClearParameters()
+            If clsBaseOperator.ContainsParameter("facets") Then
+                For Each clsTempParam As RParameter In clsFacetOperator.clsParameters
+                    If clsTempParam.strArgumentValue <> "" AndAlso clsTempParam.strArgumentValue <> "." Then
+                        clsGroupByFunction.AddParameter(i, clsTempParam.strArgumentValue, bIncludeArgumentName:=False, iPosition:=i)
+                    End If
+                    i = i + 1
+                Next
             Else
```
dannyparsons: You can remove this `Else`

instat/dlgPICSARainfall.vb
```diff
@@ -715,68 +714,54 @@ Public Class dlgPICSARainfall
...
-                    clsGroupByFunction.AddParameter("1", ucrReceiverColourBy.GetVariableNames(bWithQuotes:=False), bIncludeArgumentName:=False, iPosition:=1)
-                Else
-                    clsGroupByFunction.RemoveParameterByName("1")
-                End If
-
-                If (ucrReceiverColourBy.IsEmpty AndAlso Not ucrReceiverFacetBy.IsEmpty) Then
-                    clsGroupByFunction.AddParameter("0", ucrReceiverFacetBy.GetVariableNames(bWithQuotes:=False), bIncludeArgumentName:=False, iPosition:=0)
+            End If
+            If clsBaseOperator.ContainsParameter("hlinemean") OrElse clsBaseOperator.ContainsParameter("hlinemedian") OrElse clsBaseOperator.ContainsParameter("hlinelowertercile") OrElse clsBaseOperator.ContainsParameter("hlineuppertercile") Then
+                clsPipeOperator.AddParameter("group_by", clsRFunctionParameter:=clsGroupByFunction, iPosition:=1)
```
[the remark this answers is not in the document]
dannyparsons: Just because we have a `hline` doesn't necessarily mean we need `group_by`. For example, plotting a single variable for a single station only required mutate because there is only one `hline`.

instat/dlgPICSARainfall.vb
dannyparsons: I've realised we don't actually have a function written to be able to check if there are two parameters with the same values in our RFunctions yet. 
It's easy to add a function to check this but I think that can be done separately from this pull request since the code still works if a variable is repeated.

dannyparsons:
Shouldn't it be `clsPipeOperator.ContainsParameter("mutate")`? This function is used in many places so it should be working correctly.

stevenndungu:
> Shouldn't it be `clsPipeOperator.ContainsParameter("mutate")`? This function is used in many places so it should be working correctly.

Yes it should be that. I was doing a check on the wrong operator. similarly to colour

dannyparsons:
so is that working now?

instat/dlgPICSARainfall.vb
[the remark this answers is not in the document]
dannyparsons: these things still need to be changed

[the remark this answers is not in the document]
stevenndungu: > Just because we have a `hline` doesn't necessarily mean we need `group_by`. For example, plotting a single variable for a single station only required mutate because there is only one `hline`.

You are right. I have found it better to do away with it, together with `mutate` since there checks are not really important. Checks on `colour`, `facets `and `receivers`. What is your take on this...

stevenndungu:
> so is that working now?

Yes this works for now. have a look at it.

stevenndungu:
Great, thanks. I have merged the changes.

dannyparsons:
Great, if that now works ok then its ready.

stevenndungu:
Yes,it works fine. I have tested.

PR #5355, opened 2019-05-28
update the end date in sdgDoyRange subdialog

### LINKED ISSUE #5315: Small day range problem
I suggest this is a simple problem that anyone can look at.  (It maybe that @maxwellfundi could fix it very quickly.

The Day range sub-dialogue is used by many of the dialogues in Climatic > Prepare.  There are 2 controls, one for the start day and the other for the end day.  

Sometimes the End Day setting depends on the Start Day.  So, **if the Start day setting is changed, then the End day control should refresh**.  Currently it doesn't.

If data are needed, then the Moorings data is a good example, see #5313 for the data.  But Samaru, Nigeria  in the library will do.  

image

In the figure above first with Fixed start day of Jan 1st and End day is the third option, Length - make it 90.  Press Return.  This is fine.  You get 154 as the start and 154 + 90 as the end.
Now change the Start Day to be Feb 1st.  Press Return.

Now it is wrong.  You get 185 as the start and still 154 + 90 as the end. 

Other situations are more complicated, but I think all may be fixed if the End date refreshes.

Ivanluv:
fixes #5315 


maxwellfundi:
@dannyparsons  The values here updates correctly, this is ready for you

PR #5357, opened 2019-05-29
reusing the code from model dialogue in Hypothesis testing dialogue

### LINKED ISSUE #5341: Small improvements and explanation of problem in Model > Hypothesis
I am trying new tests in the Model > Hypothesis Tests dialogue.
a) I am finding the **Try button** to be very useful.  
But when the test works it gives "...", which is hard to see.
The Model >  Model dialogue is much better and says (in green) that the command runs without error.
Please could we copy that code into the -  oooh - re-use that code  - in both dialogues.  I can see no reason why they should be different!
b) Similar task is that the hypothesis tests dialogue doesn't remember the last commands we used.  Can we re-use that function/code too from the modelling dialogue, so it does? 

c) When the command I am using runs it still gives an error in the Hypothesis Tests dialogue.  It runs fine in the Model dialogue.

In each case the command itself runs ok.  In the hypothesis tests dialogue it then has the commands:
**data_book$import_data(data_tables=list(mydata=Last_Test))
data_book$get_data_frame(data_name="mydata")**
This gives an error.
In the model dialogue it has:
**data_book$add_model(model_name="Last_Test", model=Last_Test, data_name="mydata")
data_book$get_models(data_name="mydata", model_name="Last_Test")**

and that runs fine.

An example, using the mydata from openair package is as follows

circular::watson.test(wd)

(or properly circular::watson.test(circular::circular(wd, units="degrees")) - but error is the same)
It says:
**Error running R command(s)
Error in self$set_data(data, messages) : 
  Data set must be of type: data.frame
Last_Test <- circular::watson.test(circular::circular(wd,units="degrees"))
data_book$import_data(data_tables=list(mydata=Last_Test))
OK**   
(When I run a line at a time, it is the data_book$import_data line that is giving the error.)

I have suggested @dannyparsons first - not to do it, but to say what needs to be done, and then presumably to allocate it.

Ivanluv:
fixes #5341 
@africanmathsinitiative/developers  this is  can be reviewed


maxwellfundi:
@dannyparsons  from the dialog side, this works well, ready for you!

dannyparsons:
@maxwellfundi did you check that all suggestions have been done, in particular:

> The Model > Model dialogue is much better and says (in green) that the command runs without error.
> Please could we copy that code into the - oooh - re-use that code - in both dialogues. I can see no reason why they should be different!
> b) Similar task is that the hypothesis tests dialogue doesn't remember the last commands we used. Can we re-use that function/code too from the modelling dialogue, so it does?
> c) When the command I am using runs it still gives an error in the Hypothesis Tests dialogue. It runs fine in the Model dialogue.

rdstern:
I looked also at this pull request and confirm that the Try works better now.  But - as Danny says above - it still doesn't remember the previous test, which it does in the Modelling dialogue.

dannyparsons:
@Ivanluv Is this ready to be re-tested?

Ivanluv:
@dannyparsons  yes

dannyparsons:
@rdstern do you want to check this again?

rdstern:
Now it remembers the previous tests as well.  Great.  I think it can be merged.

Issue #4540, opened 2018-04-16
jitter plot plus boxplot or jitter plot plus violin plot


rdstern:
There are 3 ways to do this.  They are equally easy, but logically different.

Usually you want **jitter and then boxplot**, because then the central box will overwrite the middle part of the jitter.  Similarly with violin plots, i.e. jitter plus violin rather than violin plus jitter.

So there are 3 obvious possibilities:
1) Put Add boxplot checkbox on the jitter plot.  We could just have the one, or also **Add violin**. They could be 2 check-boxes, but I can't see why you would want both.
2) Put Add jitter checkbox on the boxplot (and on the violin plot) dialogues.  I still think you would then (by default) do jitter plus boxplot, rather than boxplot + jitter.
3) Do both 1 and 2 above.  In that case jitter + boxplot is in that order and boxplot + jitter is in the other order.

I am slightly inclined to option 3.

rdstern:
We now have boxplot plus jitter, which is good.  But I claim we usually need jitter plus boxplot!

Here is an example:
![image](https://user-images.githubusercontent.com/11226469/58792650-e6128a80-85f4-11e9-89ba-20067cff6856.png)

I used the existing "system" and then swapped them round in the script file.  This looks much better, because you don't see the jitter in the middle half, i.e. in the box.

In the existing dialogue this could either be in the existing boxplot, where you have the option to add points.  Or we could add the boxplot to the jitter plot.

What Danny says is really easy is to have the jitter behind the boxplot for now, as in the above.  Later we can add the option to choose the order of the layers.

I add to here some more trivial improvements in this dialogue.

a) In the Boxplot options one of the arguments is outlier.shape.  Please add NA to the possible shapes.  That's a good way to get rid of the outliers.  (We might later want to have that on the main dialogue, but this will do for now.)
b) In the violin plot radio button options there are 3 arguments called trim.  The second one should be relabelled as scale.  It doesn't seem to do anything currently.  It should add scale="count" etc.  This is the same as having variable width in a boxplot, so is important.  Later that might go on the main violin main dialogue.  I can only assume the third current trim is actually inherit.aes?

maxwellfundi:
@rdstern  With the boxplot we already have the points we need just to swap the layers so that they appear in the way you would like.

maxwellfundi:
Am now working on this!

volloholic:
[...]

PR #5358, opened 2019-05-30
Added label aesthetic to geom_point

### LINKED ISSUE #5254: Add label as a geom aesthetic at least for geom_point?
Adding the label doesn't change the visible plot.  But the information is picked up by the plotly display and that's great.  It may do no harm to add it to lots of geoms?  But I have (so far) looked at geom_point and geom_jitter and it works fine.  (It doesn't seem to do anything in geom_boxplot.) So I'd be happy if it is added to point and jitter for now.

For information, my label is the Date column.  The reason that is so nice is that for outliers it tells you exactly the date - just what is needed.

Ogik99:
fixes #5254 


instat/ucrGeom.vb
```diff
@@ -851,6 +851,7 @@ Public Class ucrGeom
         clsgeom_jitter.AddAesParameter("size", strIncludedDataTypes:={"factor", "numeric"})
         clsgeom_jitter.AddAesParameter("stroke", strIncludedDataTypes:={"factor", "numeric"})
         clsgeom_jitter.AddAesParameter("fill", strIncludedDataTypes:={"factor", "numeric"})
+        clsgeom_jitter.AddAesParameter("label", strIncludedDataTypes:={"factor", "numeric", "character", "Date"})
```
maxwellfundi: @Ogik99 This seems to have been added for the geom_jitter only. Could you still add this for the point, like you had done in the past PR?

instat/dlgScatterPlot.vb
```diff
@@ -83,6 +83,15 @@ Public Class dlgScatterPlot
         ucrReceiverX.SetValuesToIgnore({Chr(34) & Chr(34)})
         ucrReceiverX.bAddParameterIfEmpty = True
 
+        ucrReceiverLabel.SetParameter(New RParameter("label"))
+        ucrReceiverLabel.SetParameterIsString()
+        ucrReceiverLabel.bWithQuotes = False
+        ucrReceiverLabel.Selector = ucrSelectorForScatter
+        ucrReceiverLabel.strSelectorHeading = "Variables"
```
maxwellfundi: Can we have any kind of variable as a label?

instat/dlgScatterPlot.vb
```diff
@@ -138,10 +147,13 @@ Public Class dlgScatterPlot
         clsRaesFunction.SetRCommand("aes")
         clsRaesFunction.AddParameter("x", Chr(34) & Chr(34))
         clsRaesFunction.AddParameter("y", Chr(34) & Chr(34))
+        clsRaesFunction.AddParameter("label", Chr(34) & Chr(34))
```
maxwellfundi: I thought we should be adding this to the geom function?

instat/ucrGeom.vb
Ogik99: This has now been added to geom point as well

maxwellfundi: Great! See my other comments above

instat/dlgScatterPlot.vb
dannyparsons: Yes I think so since its text. Do we need to then specify `strIncludedDataTypes`? If we don't specify this I assume it includes everything?

maxwellfundi:
@Ogik99 A quick request here to edit the ucrGeom- Please update this branch from master, then add these two on ucrgeom

I add to here some more trivial improvements in this dialogue.

**a) In the Boxplot options one of the arguments is outlier.shape. Please add NA to the possible shapes. That's a good way to get rid of the outliers. (We might later want to have that on the main dialogue, but this will do for now.)
b) In the violin plot radio button options there are 3 arguments called trim. The second one should be relabelled as scale. It doesn't seem to do anything currently. It should add scale="count" etc. This is the same as having variable width in a boxplot, so is important. Later that might go on the main violin main dialogue. I can only assume the third current trim is actually inherit.aes?** relatin to an issue am working on #4540

instat/ucrGeom.vb
```diff
@@ -1414,14 +1417,13 @@ Public Class ucrGeom
...
-        clsgeom_violin.AddLayerParameter("trim", "list", Chr(34) & "area" & Chr(34), lstParameterStrings:={Chr(34) & "area" & Chr(34), Chr(34) & "count" & Chr(34), Chr(34) & "width" & Chr(34)}) 'if "area" (default), all violins have the same area (before trimming the tails). If "count", areas are scaled proportionally to the number of observations. If "width", all violins have the same maximum width.
+        clsgeom_violin.AddLayerParameter("trim", "boolean", "TRUE", lstParameterStrings:={"TRUE", "FALSE"}) 'If TRUE (default), trim the tails of the violins to the range of the data. If FALSE, don't trim the tails.
+        clsgeom_violin.AddLayerParameter("trim", "list", Chr(34) & "area" & Chr(34), lstParameterStrings:={Chr(34) & "area" & Chr(34), Chr(34) & "scale" & Chr(34), Chr(34) & "width" & Chr(34)}) 'if "area" (default), all violins have the same area (before trimming the tails). If "count", areas are scaled proportionally to the number of observations. If "width", all violins have the same maximum width.
```
maxwellfundi: @Ogik99   Here you needed to change the parameter name, not the `count`. We needed the second trim to be scale.

instat/dlgScatterPlot.vb
maxwellfundi: yes we can simply not specify and it will accept everything.

instat/dlgScatterPlot.vb
maxwellfundi: I think we already solved this and its okay to have this here.

instat/ucrGeom.vb
maxwellfundi: @Ogik99  one more thing here, I have checked and the third trim should be `inherit.aes` on the geom violin. please add that too.

Ogik99: @maxwellfundi  should we put FALSE as one of the options to inherit.aes. If you put FALSE you get the following error:

geom_violin

Ogik99: I don't think there's much the user can do about this.

Ogik99: this should be ready now

maxwellfundi:
@Ogik99 this looks good except just one thing you need to fix. 

The layer subdialog and the main dialog are not in syc with the label- When you fill in the label on the main dialog, then it automatically populates the label receiver on the layer subdialog, which is good and expected- 
However, if i remove the label from the layer subdialog, the code runs okay but then the label receiver in main dialog still remains filled in. Also when you have the label receiver filled in the layer sdg then it does nop populate the label receiver in the main dlg. 

I suggest you run rcode for that receiver as soon as you go out of the layer subdialog so that they are always in sync

Ogik99:
@maxwellfundi I have been able to solve atleast one of the issues you raised. 
This is now resolved:
> Also when you have the label receiver filled in the layer sdg then it does nop populate the label receiver in the main dlg

This other issue persists though
> However, if i remove the label from the layer subdialog, the code runs okay but then the label receiver in main dialog still remains filled in.

I have tried running the following code,which I thought could resolve this issue (however it does not):
`ElseIf clsParam.strArgumentName = "label" Then
                If clsParam.strArgumentValue = Chr(34) & Chr(34) Then
                    ucrReceiverLabel.Clear()
                Else ucrReceiverLabel.Add(clsParam.strArgumentValue)
`


Its interesting that the following code works well for the x aesthetic (and y aesthetic as well).
`If clsParam.strArgumentName = "x" Then
                If clsParam.strArgumentValue = Chr(34) & Chr(34) Then
                    ucrReceiverX.Clear()
                Else
                    ucrReceiverX.Add(clsParam.strArgumentValue)
                End If`

instat/ucrGeom.vb
```diff
@@ -1414,14 +1417,14 @@ Public Class ucrGeom
...
         clsgeom_violin.AddLayerParameter("adjust", "numeric", "1", lstParameterStrings:={1, 0}) 'The bandwidth used is actually adjust*bw. This makes it easy to specify values like ‘half the default’ bandwidth.
         clsgeom_violin.AddLayerParameter("kernel", "list", Chr(34) & "gaussian" & Chr(34), lstParameterStrings:={Chr(34) & "gaussian" & Chr(34), Chr(34) & "rectangular" & Chr(34), Chr(34) & "triangular" & Chr(34), Chr(34) & "epanechnikov" & Chr(34), Chr(34) & "biweight" & Chr(34), Chr(34) & "cosine" & Chr(34), Chr(34) & "optcosin" & Chr(34)}) 'A character string giving the smoothing kernel to be used.
         'This is only a parameter in developer version of ggplot. May soon be in release version.
         'clsgeom_density.AddLayerParameter("n", "numeric", "512", lstParameterStrings:={0, 0}) 'The number of equally spaced points at which the density is to be estimated. When n > 512, it is rounded up to a power of 2 during the calculations. If negative, sends an error.
-        clsgeom_violin.AddLayerParameter("trim", "boolean", "FALSE")
+        clsgeom_violin.AddLayerParameter("inherit.aes", "boolean", "TRUE", lstParameterStrings:={"TRUE", "FALSE"})
```
dannyparsons: Please don't add this to the geom definition. This is for all geoms for whether to use the global aes or not. There is already an option for this on the Layer Options sub dialog.

shadrackkibet:
@Ogik99  what is the status of this?

instat/ucrGeom.vb
maxwellfundi: @dannyparsons this is actually okay because the `trim` was repeated and in the current version 
`inherit.aes` is missing for the geom_violin.

dannyparsons: We don't add `inherit.aes` for any geom because its done with a separate checkbox on the aes tab. It's missing for all geoms deliberately.

maxwellfundi: Ooo yeah i  get it. thanks :-)

maxwellfundi:
@shadrackkibet I have actually checked this now and looks good.

PR #5364, opened 2019-06-06
Reset the input filter name

### LINKED ISSUE #5363: Filter name should reset when defining a new filter
When defining a new filter and typing a name for the filter, if I later define another filter it displays the same name as used previously. This is not really sensible because the rest of the dialog resets since it is a new filter. But currently its too easy for users to overwrite a previous filter.

So I suggest the filter name resets when we define a new filter.

Ogik99:
fixes #5363 


Ivanluv:
From my review it is working as expected

instat/ucrFilter.vb
```diff
@@ -54,6 +54,11 @@ Public Class ucrFilter
             ucrFilterByReceiver.Add(strDefaultColumn)
             SetDefaultColumn("")
         End If
+
+        If ucrInputFilterName.bUserTyped Then
+            ucrInputFilterName.SetPrefix("filter")
```
dannyparsons: I don't think this should only be if the user has typed but every time the dialog opens. For example, if some filters have been deleted we still want the name to reset, but the user might not have typed.

Also, does it work if you use `.Reset()` instead? Setting the prefix again is not the ideal way to reset the control (this might be that we need to improve the reset function if not)

Ogik99: @dannyparsons  the .Reset() method doesn't seem to work at all.

dannyparsons: Ok that's fine, keep the SetPrefix then. But still remove the user typed check I think.

maxwellfundi: I remember there was an issue with the `.Reset( )`. @Ogik99 please remove the usertyped here and should be fine.

Ogik99: This should be okay now.

PR #5369, opened 2019-06-10
Enabling the sf options button on the dialog. 

maxwellfundi:
@africanmathsinitiative/developers this is ready for review. @Wycklife  you have worked on this dialog before, please test this


maxwellfundi:
@rdstern @dannyparsons Other than the `Station Names` - Should we call this `Labels`  I also suggest we have optional on all the labels with optional variables


image

rdstern:
Let's also check with Danny but I wonder about not having the (Optional) anywhere?  We have other dialogues where some of the fields can be left empty?

maxwellfundi:
yes but mostly we have the optional after the receiver label. I think it makes it easier for the user to know that it is not mandatory.

dannyparsons:
Please add a space before `(Optional)`. We do need to think how to be consistent with optional receivers but I think its useful here because there are so many.

Fill should also optional.

I agree stations names should be label.

maxwellfundi:
Okay , then for this version we will have the above layout.. hopefully we could have the size and colour for the points before they tomorrow

maxwellfundi:
Thanks @Wycklife  for the designer changes

PR #5389, opened 2019-06-21
Removing the upper Limit for Time Scale

### LINKED ISSUE #5375: Two small edits to the SPI dialogue
This is working ok - at least for the SPI.  
Two small points:
1) Remove the upper limit from the  up-down control for the Time scale.  The limit is currently 24, i.e. it goes from 1 to 24.  There are situations when we want more than this.  The limit of 24 is 2 years if our data are monthly.  But the dialogue also works for 10-day or even daily data, where the limit would be much higher than 24.  So I suggest no top limit at all.  Still fine to have an up-down here, and going from 1 in steps of 1.
2) Currently the detailed results are printed in the output window.  This is partly because earlier it didn't save the resulting column.  Now it does.  I don't mind the model parameters being printed out.  But the detailed results are at least 12 times the length of the data frame.  Ideally printing them would be an option, but if that takes too long, then just don't print.  

I have put this as 5.1 in the expectation that it can be very quick.  Change the milestone if it isn't.

Ivanluv:
fixes #5375


maxwellfundi:
@Ivanluv  is this ready for review? if so could @Wycklife review?

Wycklife:
@maxwellfundi doing it now

Wycklife:
The upper limit doesn't exist anymore for the time scale

instat/dlgSPI.vb
```diff
@@ -148,7 +148,7 @@ Public Class dlgSPI
         clsDolarOperator.AddParameter("model", clsRFunctionParameter:=clsSpiFunction, iPosition:=0)
         clsDolarOperator.AddParameter("fitted", "fitted", iPosition:=1)
 
-        ucrBase.clsRsyntax.SetBaseRFunction(clsSummaryFunction)
```
dannyparsons: @Ivanluv This is good to stop the output. Now in the dialog we have no base function and an "After" function, which is a bit odd, because its not after anything.
Can you edit so that `clsAsVectorFunction` is now the base function?

Ivanluv:
@dannyparsons  does this look fine?

dannyparsons:
Not quite, because you have `clsAsVectorFunction` as the base function and as an "After" function.

Ivanluv:
so should I remove the "After" function?

dannyparsons:
Well, isn't the function going to run twice otherwise? How does the script look? We only need it as the base function so we don't need to run it again after.

PR #5392, opened 2019-06-25
Added key_glyph layer parameter to ucrGeoms

### LINKED ISSUE #5387: Key glyphs for legends - ggplot2
[Key glyphs](https://ggplot2.tidyverse.org/reference/draw_key.html) are a new feature of ggplot2 3.2.0 for display of the legend.

Every geom has a new `key_glyph` parameter. It can be one of the `draw_key` functions but easier is just to give the name of the function. There are 16 in the documentation e.g. this is quite nice

```
p <- ggplot(economics, aes(date, psavert, color = "savings rate"))
# key glyphs can be specified by their name
p + geom_line(key_glyph = "timeseries")
```
image

We can easily add the 16 options as a new geom parameter for every geom. Since this explicitly says it is for every geom then we should add this to all the geoms in one go with a for loop.

Ogik99:
fixes #5387 


maxwellfundi:
This looks good, can someone review it? @Ogik99 you can store the list of the parameters then pass the list . 

But this would still work okay

instat/ucrGeom.vb
```diff
@@ -1506,6 +1506,10 @@ Public Class ucrGeom
 
         lstAllGeoms.Add(clsgeom_vline)
 
+        For Each Geom In lstAllGeoms
+            Geom.AddLayerParameter("key_glyph", "list", "draw_key_point", lstParameterStrings:={"draw_key_point", "draw_key_abline", "draw_key_rect", "draw_key_polygon", "draw_key_blank", "draw_key_boxplot", "draw_key_crossbar", "draw_key_path", "draw_key_vpath", "draw_key_dotplot", "draw_key_pointrange", "draw_key_smooth", "draw_key_text", "draw_key_label", "draw_key_vline", "draw_key_timeseries"})
+        Next
+
```
maxwellfundi: from @dannyparsons issue - `geom_line(key_glyph = "timeseries") `seems you might need `chr(34)` around the parameter values? 
I also find that others would not need that, like -  `geom_line(key_glyph = draw_key_rect)`. COuld you check these more to see if there are more like the timeseires ones?

Ogik99: We could implement this in two ways: geom_line(key_glyph = "timeseries") or geom_line(key_glyph = draw_key_timeseries). Both methods have the same output.
But I think having it in the first form is much better

maxwellfundi:
@Ivanluv  while you on geoms, could you test this too.

maxwellfundi:
@Ivanluv Please review this this this morning

Ivanluv:
@maxwellfundi  this is  fine from my review

instat/ucrGeom.vb
dannyparsons: I like the first too, its shorter and looks like other parameters.

PR #5393, opened 2019-06-26
Added rug plot to the PICSA rainfall graph

### LINKED ISSUE #5024: Adding a rugplot to the PICSA rainfall graphs
The rainfall dialogues on the start and and length of the rains can each produce a logical column.  It would be good if this could also be displayed on the PICSA rainfall graphs

A rug-plot [see here](https://ggplot2.tidyverse.org/reference/geom_rug.html) would seem ideal.  It would need an additional layer, either in the PICSA options or on the main dialogue.  I suggest an additional tab in the PICSA options, because there will be details of the size and colour of the rug.

### LINKED ISSUE #5094: Adding a rug plot to the PICSA rainfall graphs
start with rug plot

I have added (manually) what might be an acceptable rugplot, see above.  It isn't quite right, because it displays missing (last 2 years) as no date, but that would be easy to fix.  I did this by making the status numeric (I couldn't get the logical to show.) and then plotting the rug with the alpha parameter being 1-status.  I set the rug to have no legend and be plotted only on the bottom axis.

I assume this may be quite easy to add, though we may want to explore more options first, e.g. to have 3 possibilities, i.e. faint (as now), bold line as now, and a third for missing?  For other graphs (e.g. rainfall totals, we may want a rugplot just to indicate missing?

Ogik99:
fixes #5094 


instat/dlgPICSARainfall.vb
```diff
@@ -302,6 +314,7 @@ Public Class dlgPICSARainfall
         clsGeomLine.AddParameter("colour", Chr(34) & "blue" & Chr(34))
         clsGeomLine.AddParameter("size", "0.8")
         clsBaseOperator.AddParameter(clsPointsParam)
+        clsBaseOperator.AddParameter(clsRugParam)
```
maxwellfundi: @Ogik99 I dont think we want to add this parameter here by default. I will discuss more with @rdstern  and revert on this

maxwellfundi:
@dannyparsons we have been trying to think how we want to add the rugplot here but not sure how to do it properly. 
@rdstern says you did this for the right thing which we cant figure out hoe to do now. Any ideas?

Carolnikie:
- I have tested this using the dodoma dataset, selected the first month of all the years greater than 1996 and came up with the graph below using tmax as the single variable
image
- Is it possible to have the rug plot printed outside the graph like I did with the graph below in R studio using the following code:  p + geom_rug(outside =T,length = unit(0.05, "npc")) + scale_y_continuous(expand=c(0.1,0.1))+coord_cartesian(clip = "off")
Rplot

maxwellfundi:
@dannyparsons @rdstern ???

rdstern:
I do like the idea of rug plots generally.  But for the PICSA graphs the plan is for the graphs to be able to show some extra information.  Wait a bit with this, because I am no longer so sure that a rug plot is the way to go.  Let me explain the problem first.

When plotting the start (or the end) of the rains in PICSA rainfall there are some years when there was no start.  Maxwell, you may remember this was mentioned by Keenness in the workshop last week.  

In the dialogue for the start and the end we capture this information (well we will do) with a logical column.  This logical column can be missing, or TRUE or FALSE.  If it is missing that means there are missing values in the data.  If TRUE, then there was a start.  If FALSE then there wasn't a start.

So I would like to add the information from this extra variable as another layer in the PICSA graph.  It could be an "Add Logical" button somewhere.  I think it may be an extra Tab in the sub-dialogue, because it only applies to particular graphs.

I thought that perhaps the rug-plot might work.  It would be on the x-axis, with the years, or it might (better) be on the top X axis, but it would be a rug plot **of the logical column**.  So it isn't the default of a rug plot of the data in an exiting layer.  One way might be to make the logical column into missing when it is either missing or TRUE.  So it only gives a point when it is FALSE.  Those are the years when no planting was possible.

There are quite a lot of options for the rug-plot, so I think it might work and it might have its own tab in the sub-dialogue.  You can have it to plot an extra variable, so I think this might work.  You can choose whether to have it on the bottom, or the top (or both).  You can also choose whether it should be inside or outside the graph area.  I'll do some experimenting myself, with adding it as an extra layer.

(It may be better to just plot it as a point plot?  The x is then the years, and the y is either at the bottom or the top?  But I am just not sure how to specity top and bottom of the plot, and the rug-plot does that automatically.

rdstern:
image

So here is an example of what I would like.    It isn't quite right for PICSA, but close enough.

Here are the data I used: 
[dodoma_by_year.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/3361963/dodoma_by_year.xlsx)

Here is another example:
image

dannyparsons:
We're not quite decided on how to add this yet so I think wait a bit as Roger suggested. You can add the code for the geom but don't add it to the graph yet.

What could be done now is adding the `outside` parameter to the `geom_rug` definition.

dannyparsons:
@Ogik99 I suggest not going further than you have already with `geom_rug` at this stage until we have a clearer idea of how we want it to work for PICSA graphs, which is still to be discussed. The code you added in the dialog is already useful but we'll need to have more ideas on implementation before you can make more progress.

PR #5403, opened 2019-07-02
Right justify

### LINKED ISSUE #5261: Minor edit on File > Open
image

Here is what I would like on this dialogue.
Currently the file name at the top is left justified, so you have no idea of the name of the file.  The field is also very small. 
If it could be right-justified, as above, then you would know the name of the file you are opening.

Ivanluv:
fixes #5261 


Ogik99:
Is it possible to have the box wide enough to accommodate all contents of the 'path directory'
rightjust

rdstern:
I agree to make it as wide as possible.  The limit is the size of the group box though, because there is another control next to it - for multiple Excel sheets, etc.

maxwellfundi:
Yeah sure, i would also prefer a longer box to accommodate the path

Wycklife:
This now seems okay. Just that the length of Input File Path is a bit large but I realized it depend on the length name of the data set.

dannyparsons:
I like the longer box, but you've introduced some issues:

- Left and right arrows are swapped when using them in the textbox
- When the text is shorter than the length of the textbox, the text is aligned to the right of the textbox

@Ogik99 @maxwellfundi @Wycklife these are things you should have noticed through testing and raised.

I think your solution of making the textbox right-to-left is not really what we want to solve this issue. We just want it to scroll to the end of the textbox when a file is selected.

Some suggestions to do this seem to be:
`Textbox.SelectionStart = Textbox.Text.Length`
or
`Textbox.ScrollToCaret()`

so this is a starting idea.

PR #5404, opened 2019-07-02
Net cdf directory

### LINKED ISSUE #5401: wrong directory for the library for File > Open and Tidy Netcdf.
The wrong directory for the library for File > Open and Tidy Netcdf.  This was a problem that Maxwell encountered.  I don’t know if it would also have been the same in the Open from Library > Instat > directory.  Maxwell understood the issue better than I did.

Ivanluv:
fixes #5401 
@africanmathsinitiative/developers  this can be reviewed


dannyparsons:
Great, thanks for this! It was an issue in the workshop here so good its now done. 
I hadn't thought about needing to update this when we changed the folder structure.

PR #5410, opened 2019-07-03
Bar chart bug

### LINKED ISSUE #5409: Small bug in bar chart
This should be a simple one to fix!
In the bar chart one argument is called position.
The default is **stack** and that is fine and works.  The next option is what could be labelled **stack (reverse).**  Currently it is labelled position = position_stack(reverse =TRUE), which looks very odd and out of place there!  That's essentially the code, not the label.

But the bug is more what is what it produces in code, which is **position = position = position_stack(reverse=TRUE)**, i.e. the **position =** is repeated twice.

Ivanluv:
fixes #5409 
@africanmathsinitiative/developers this  is ready for review

### DIFF
--- a/instat/ucrGeom.vb
+++ b/instat/ucrGeom.vb
@@ -301,7 +301,7 @@ Public Class ucrGeom
         'Global Layer parameters
         clsgeom_bar.AddLayerParameter("stat", "list", Chr(34) & "count" & Chr(34), lstParameterStrings:={Chr(34) & "count" & Chr(34), Chr(34) & "identity" & Chr(34)})
         clsgeom_bar.AddLayerParameter("show.legend", "list", "TRUE", lstParameterStrings:={"NA", "TRUE", "FALSE"})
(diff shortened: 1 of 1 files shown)


maxwellfundi:
Am not sure how this was missed-  @Ogik99 could you check this to see its looking okay?

maxwellfundi:
@Ogik99 could you have a review of this this morning?

maxwellfundi:
Additionally here, these two files have been edited by two people, i suggest, when its reviewed then @Ivanluv merges his changes to @Ogik99 s branch with the changes on `ucrGeom` called `label_geom`

Ogik99:
> Am not sure how this was missed- @Ogik99 could you check this to see its looking okay?

This now works well (there's no longer code repetition)

However, there is another unrelated issue (which is similar to an existing issue under #5358. Changes made to the bar chart options sub dialog for the position parameter do not get updated on the main dialog specifically for position_stack(reverse), Dodge2.
This is something I will be looking into.

dannyparsons:
@Ogik99 if you run SetRCode when returning to the dialog controls will update based on the code and this should be fixed, good to notice this issue.

PR #5411, opened 2019-07-03
Added label to the Use summaries dialog

### LINKED ISSUE #5398: Adding labels to bar chart, etc.
We have the Describe > General > Use Summaries dialogue which seems to work ok for bar charts - I haven't checked pie charts.  Here are data with an example:

[Dodoma_normals.zip](https://github.com/africanmathsinitiative/R-Instat/files/3344731/Dodoma_normals.zip)

image

Getting this is reasonably easy by adding the label column as a second layer.  But there are some issues to raise, in making it a bit easier.

1. The label layer has the same x and y as the data.  Then, the label is also (just about) the data.  But the default gives many decimal places.  So I used round(column).  I could easily produce a new variable (using the calculator), with the round function.  An alternative would be to allow me to type the round function into that field?  
2. I also used th4 nudge y and made it 3 for the above plot.  That's fine from the layer options.
3. Could we consider having the labels as an option on the main dialogue?  There is plenty of space.  It would be an Add Label checkbox.  If checked then there may be more options, but the obvious would be 
a) The Label - as we assume the same x and y.  Perhaps here is where we could allow typing (maybe as a further option, or by default?  (So could type round.  Or here we could have the round as a further option.  It would only apply to numeric columns.  Then it could add the round function - default is zero, but could include an up-down for decimals - say up to 5. .  
b) Then the nudge x and y I guess
c) Possibly the option to have geom_label or geom_text.  

If this was in place for one example dialogue it would not be so bad that others had to use the (usual) sub-dialogue.

I suggest this is useful and perhaps quite an easy addition to make for AMI?.

Ogik99:
fixes #5398 
Work in progress


instat/dlgSummaryBarOrPieChart.vb
```diff
@@ -259,4 +261,12 @@ Public Class dlgSummaryBarOrPieChart
     Private Sub CoreControls_ContenctsChanged() Handles ucrReceiverYVariable.ControlContentsChanged, ucrReceiverFactor.ControlContentsChanged, ucrSaveSummaryBar.ControlContentsChanged
         TestOkEnabled()
     End Sub
+
+    Private Sub ucrChkLabel_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkLabel.ControlValueChanged
+        If ucrChkLabel.Checked Then
+            clsBaseOperator.AddParameter("labelfunc", clsRFunctionParameter:=clsLabelgeomFunction)
```
dannyparsons: parameter name should be the name of the geom

dannyparsons:
@Ogik99 can you send an image of the new design when you have it? I think it will need some discussion before being agreed upon.

Ogik99:
Use summaries

Ogik99:
I added the 'second check box' and the single receiver - 'Label variable'

dannyparsons:
Thanks, I've had an idea and made a comment here https://github.com/africanmathsinitiative/R-Instat/issues/5398#issuecomment-510467597 which may simplify what's needed on the dialog. So might be good to wait for discussion before changing more.

maxwellfundi:
@Ogik99 whats going on with this?

Ogik99:
> @Ogik99 whats going on with this?

This is still subject to discussion

shadrackkibet:
@Ogik99 what is the state of this.

lloyddewit:
@shadrackkibet Please could you try and complete this PR? (@rdstern suggested that you would be the best person for this task)

shadrackkibet:
Happy to take over. Thanks.

shadrackkibet:
Since PR #5896  contains all the changes in this PR, I expect this to merged automatically once PR #5896 is merged!

lloyddewit:
@shadrackkibet is correct, PR #5896 includes all these commits so this PR will merge automatically when PR #5896 is merged. However, in order to reduce the number of open PRs, I decided to close this PR now rather than waiting.

PR #5412, opened 2019-07-03
Correcting names for corruption dialogs

### LINKED ISSUE #2841: Naming corruption dialogs
I have been going through some corruption dialogs this morning and we have not had a standard way of naming them. 
Some are named with this format- dlgCorruptionDescribe, while others have just the standard way of naming dialogs. 
My suggestion is, we use the first format as it is easier to distinguish the corruption dialogs easily unlike when they mix up with all other dialogs. 
We do not do this for all other menus, I wonder if this would be an appropriate approach.

Wycklife:
Fixes #2841

@africanmathsinitiative/developers In progress


maxwellfundi:
@Wycklife  I am merging this, having all lots of files changed is definitely going to cause problems at some point. I suggest you keep track of what has been renamed so that you can finish the task.

PR #5417, opened 2019-07-09
Added outside parameter to geom rug

### LINKED ISSUE #5413: Add extra argument to the rug plot geom
In the geom_rug there is a parameter called outside which is not included yet.  Please could it be added.  It is just TRUE or FALSE and default is FALSE.

In the guide, note in the R code the extra bit that needs to be added:

**outside** "logical that controls whether to move the rug tassels outside of the plot area. Default is off (FALSE). You will also need to use coord_cartesian(clip = "off"). When set to TRUE,

Ogik99:
fixes #5413


dannyparsons:
@Ogik99 I took over this branch as I needed to do bug fixes on the dialog and didn't want to conflict with your branch. So the code changes here have been merged but I hid the Rug plot tab so there are no visual changes related to this yet, which will need some further discussion.

Issue #5310, opened 2019-04-25
Add further missing option to summaries (and climatic summaries


rdstern:
We currently have a Missing Tab in the Prepare > Column: Reshape > Column Summaries dialogue.

I would like a further checkbox to be added, underneath the existing materials.

Default is unchecked.  Label is **Limit on consecutive missing values**
If checked, then another label is (as the label above, but with consecutively added, i.e. 
**Summary set to missing if more than this number are missing consecutively**.  

Then an up-down (perhaps underneath the label if not enough space.  Default could be 4.

This should be easy to add and leave disabled for now.  It maybe that the R-code is easy - it is a "spell" of missing values.  Danny or David to advise.

If it can be added, this is all that is needed for R-Instat to be able to provide climate normals - and that would greatly impress Climsoft installers!

Wycklife:
@rdstern What could be the appropriate name for the checkbox?

rdstern:
As above:
Default is unchecked. Label is **Limit on consecutive missing values**
If checked, then another label is: 
**Summary set to missing if more than this number are missing consecutively.**

Then an up-down (perhaps underneath the label if not enough space. Default could be 4.

shadrackkibet:
@Wycklife any update on this?

shadrackkibet:
@dannyparsons  @volloholic I have implemented this check as a function in a similar way we did for the count of missing values. I did this initially for mean function. This is all that is needed to calculate climate normals.  Below is my implementation. Any comments?

```
na_consecutive_check <- function(x, threshold = NULL, ...){ 
  is_na_rle <- rle(is.na(x)) 
  is_na_rle$values <- is_na_rle$values & is_na_rle$lengths >= threshold 
  return(any(is_na_rle$values)) 
  }

summary_mean <- function (x, add_cols, weights="", na.rm = FALSE, trim = 0, na_type = "", threshold = NULL, ...) { 
  if( length(x)==0 || (na.rm && length(x[!is.na(x)])==0) ) return(NA) 
  else { 
    if(na.rm && na_type != "" && !na_check(x, na_type = na_type, ...) || na_consecutive_check(x, threshold = threshold, ...)) return(NA) 
    else return(mean(x, na.rm = na.rm, trim = trim)) }
}
```

**Example**
```
x <- c(NA,6,3,4,NA,NA,NA)
summary_mean(x)
[1] NA
summary_mean(x, threshold = 2, na.rm = TRUE)
[1] NA
[...]

PR #5423, opened 2019-07-13
Amending missing values options - Climate normals implementation.

### LINKED ISSUE #5300: Calculating climate normals in R-Instat
[climate normals.pdf](https://github.com/africanmathsinitiative/R-Instat/files/3075718/climate.normals.pdf)

We are proposing that R-Instat become the products and applications partner to Climsoft.  A recent request (by Mozambique Met service) is for Climsoft to be able to calculate climate normals.  This would be quite a challenge for Climsoft and we should consider (reasonably urgently) providing this facility in R-Instat.
The WMO guide (attached) is very helpful and what we should use.

There are perhaps three steps towards being able to do these calculations:
1. Be able to do them manually for some stations.
2. Be able to check and improve station records to facilitate the calculations
3. Be able to automate the process, so it is really easy for NMS staff.

When there are good data we can calculate climate normals now, using our Climatic > Prepare > Climatic Summaries (or the main Prepare > Column: Reshape > Column Summaries) dialogue.  I assume the climatic one will soon anyway be able to handle missing values, see #5291.

I propose step 1 is easy and we do it reasonably quickly.  It is an extension of our criteria for calculating summaries when there are missing values.  It extends the missing values calculation in the following way:

The definition in WMO is to calculate the monthly value unless:
a) Observations are missing for 11 or more days in the month (which we can already do)
b) Observations are missing for 5 or more consecutive days in the month - which we can't do.

Some countries use 5 or more instead of 11 above and 3 or more instead of 5 above, which we would automatically accommodate.
I don't think we need to get clever about missing periods that overlap months.  We just look for consecutive missing values in the month.

So we need to be able to add this to our missing criteria and set defaults and we are away.

The WMO guide discusses three main items that I interpret here as Step 2.
a) Being able to **infill missing values** in ways that can be recommended.  This is vague, but we are starting towards this with our work on satellite data.  The WMO guide does not like methods that don't add information (e.g. interpolating) and discusses radar, neighbouring stations, satellite, etc.  
b) Checking on **homogeneity** and being able to adjust.  This is a big and important area that I discuss separately.  We would need to include this by a launch - if we have one! 
c) In homogeneity it particularly mentions changing stations, including automatic stations.  That reminds us that we need to start with facilities to be able to handle within-day records easily - and by the launch!

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes #5300 and #5310. This is the extension of missing values options to enable calculation of climate normals. I am available to provide guidance on testing if needed.

### DIFF
--- a/instat/dlgClimaticSummary.vb
+++ b/instat/dlgClimaticSummary.vb
@@ -21,7 +21,7 @@ Public Class dlgClimaticSummary
     Private iReceiverMaxY As Integer
     Private iReceiverLabelMaxY As Integer
     Private bResetSubdialog As Boolean = False
-    Private clsDefaultFunction, clsSummariesList, clsDefaultFactors, clsDayFilterCalc, clsDayFilterCalcFromConvert, clsDayFilterCalcFromList As New RFunction
+    Private clsDefaultFunction, clsConcFunction, clsSummariesList, clsDefaultFactors, clsDayFilterCalc, clsDayFilterCalcFromConvert, clsDayFilterCalcFromList As New RFunction
     Private clsFromAndToConditionOperator, clsFromConditionOperator, clsToConditionOperator As New ROperator
 
     Private Sub dlgClimaticSummary_Load(sender As Object, e As EventArgs) Handles MyBase.Load
@@ -147,6 +147,7 @@ Public Class dlgClimaticSummary
         clsDefaultFunction = New RFunction
         clsSummariesList = New RFunction
         clsDefaultFactors = New RFunction
+        clsConcFunction = New RFunction
 
         clsFromAndToConditionOperator = New ROperator
         clsDayFilterCalc = New RFunction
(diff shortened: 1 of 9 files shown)


shadrackkibet:
@rdstern @dannyparsons Here is the design showing missing values tab. Any comments?
image

shadrackkibet:
Since **Add date column** option is straight forward to implement, I have enabled it in this **PR**
image

shadrackkibet:
@africanmathsinitiative/developers it will be good for someone to test this.

dannyparsons:
@shadrackkibet I've tested this and it's working well.
Just one thing is can you make some of the words in the missing options in bold?
I think we said:
- **number of missing**
- **number of non missing**
- **percentage of missing**
- **number of consecutive missing**

shadrackkibet:
@dannyparsons I didn't find an easier way to **bold** some part of a checkbox label through code. However, I added a label control on top of a checkbox and **bolded** label text as shown below. What are your views on this?
image

dannyparsons:
OK I thought that would be easier. I think it looks a little blurry with the extra textbox on top so I think we could just go back to how it was before.

shadrackkibet:
What about making them upper case?

shadrackkibet:
Done. I have reverted to the previous design.

dannyparsons:
I think this is good now, unless we come up with another idea at a later point.

PR #5405, opened 2019-07-02
Added geom_contour

Ogik99:
fixes #5402 

@africanmathsinitiative/developers this is ready for review.

PR #5424, opened 2019-07-15
Minor edit on geom contour bin width parameter

### LINKED ISSUE #5402: Adding geom_contour to list of geoms
could we add a geom for contour plots potentially for the new mapping.  I think this is just geom_contour?

The geom exists and so please check it out, comment the code and test it.

Ogik99:
fixes #5402

### DIFF
--- a/instat/ucrGeom.vb
+++ b/instat/ucrGeom.vb
@@ -420,8 +420,8 @@ Public Class ucrGeom
         clsgeom_contour.AddAesParameter("weight")
 
         ''add layer parameters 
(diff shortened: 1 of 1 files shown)


Ogik99:
#5405

dannyparsons:
@Ogik99 what about the issue with `position` and `stat` parameters from here? https://github.com/africanmathsinitiative/R-Instat/pull/5405#discussion_r302473773
At least if we don't have other options for them then they need to be editable boxes.

Ogik99:
I have made changed to geom_tile, tufteboxplot, polygon, density2d and contour. This should be okay now.

maxwellfundi:
I have looked into this and the changes made here okay as expected.

PR #5425, opened 2019-07-15
Infill missing date- Adding data limits and fixed limits options

shadrackkibet:
Design preview.

image
image

### DIFF
--- a/instat/dlgInfill.Designer.vb
+++ b/instat/dlgInfill.Designer.vb
@@ -41,20 +41,21 @@ Partial Class dlgInfill
         Dim resources As System.ComponentModel.ComponentResourceManager = New System.ComponentModel.ComponentResourceManager(GetType(dlgInfill))
         Me.lblDate = New System.Windows.Forms.Label()
         Me.lblByFactors = New System.Windows.Forms.Label()
-        Me.rdoDataLimits = New System.Windows.Forms.RadioButton()
-        Me.rdoFixedLimits = New System.Windows.Forms.RadioButton()
-        Me.ucrDtpEndDate = New instat.ucrDateTimePicker()
-        Me.ucrChkEndDate = New instat.ucrCheck()
-        Me.ucrChkStartDate = New instat.ucrCheck()
-        Me.ucrChkResort = New instat.ucrCheck()
+        Me.lblLimits = New System.Windows.Forms.Label()
+        Me.grpOptions = New System.Windows.Forms.GroupBox()
+        Me.lblStartDate = New System.Windows.Forms.Label()
+        Me.lblEndDate = New System.Windows.Forms.Label()
+        Me.lblCompleteYears = New System.Windows.Forms.Label()
         Me.ucrReceiverFactors = New instat.ucrReceiverMultiple()
         Me.ucrReceiverDate = New instat.ucrReceiverSingle()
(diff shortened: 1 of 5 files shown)


Wycklife:
@shadrackkibet I'm getting the bug below when opening the dialog.


Capture

shadrackkibet:
Thanks, this is not yet ready for review.

Wycklife:
@shadrackkibet Okay, thanks

dannyparsons:
Maybe the radio buttons could be replaced with a label and combo box. So you have
"Limits:" and then a combo box with "From Data" and "Fixed" as the two options.

I think that would make it look better on the dialog. 

You could also put a group box around the whole part.

shadrackkibet:
updated design.

image
image

rdstern:
This looks very good.  Does this still allow what to me would be the default - namely not to complete the years, but just to complete the missing dates within the existing data?
So, when the option is Data Limits, then could there be a checkbox before the "Complete Years".  Default is unchecked.  If you check it, then it adds the Starting and is as is shown above.

shadrackkibet:
Yes, we programmed to allow that as the default option. I will add a checkbox for this purpose. Actually, **complete years starting from** should be the label of the checkbox.

shadrackkibet:
@africanmathsinitiative/developers this is now ready for review and testing.

rdstern:
I will look at this

rdstern:
1. The bug I found earlier when there is no station seems to have gone.  It used to insist on saying Infilling xxx missing dates for <Station Name>.  Now, without the station name it correctly omits that part of the sentence.
2. I tried with single and with multiple stations.  It is working well.  This is a much stronger feature now.
3. The only issue I have is with the selection of dates when they are specified.  It currently defaults to today's date I think.  I tried with an earliest date that is later than the existing data and it gave me an error.  The error was clear - so that's fine.  But I would prefer that once the date field is filled, it checks for the earliest and latest dates in the data frame.  Then it puts those as the default.  If you then choose something later than the minimum or earlier than the maximum, that is your problem.
4. The name of the dialogue is now correctly changed to Fill Date Gaps.  The menu item is still "Infill Missing Dates" and should also be changed.

Overall this is great!

shadrackkibet:
@dannyparsons do you have any ideas on how I could approach **3** above? It is something unique I  don't think we have done such before.

rdstern:
I assumed that finding the min and max of the date variable would be an easy thing to do and hence to add.  If this isn't the case, then the dialogue is ok as it stands.

maxwellfundi:
@shadrackkibet I think there could be a way of doing this internally maybe when the dialog loads and after populating the date receiver then we  take the date column and find for the **Max** and **Min** dates.

shadrackkibet:
Great. I now have an idea on how to approach this, I will try out something.

dannyparsons:
Yes we can internally find the min and max - not when the dialog loads, but when the date receiver is filled. We just need to be careful about when we set these values. i.e. if I change the dates but then change the date column, what should happen.

dannyparsons:
@shadrackkibet I suggest this could be merged to have the current improvements for the next version, and the additional min/max feature can be done separately?

shadrackkibet:
Agreed. I am going to open a separate issue for the min/max feature.

Issue #5321, opened 2019-05-04
Order of layers still swapping on ggplot


rdstern:
The plotting is holding up much better now, than previously.  And I have been producing many plots for the climatic guide.

But I have (at last) found a simple reproducible example where the order of the layers still swaps (and where it then again says the offending layer can't be edited, because it is on the base layer.

The graph below has 3 layers.  Main one is a boxplot and then the additional layers are a line plot and then a point plot.

![image](https://user-images.githubusercontent.com/11226469/57176071-41f9c000-6e4b-11e9-8180-14d86b50648d.png)

The data are here:
[mooringsplottest.zip](https://github.com/africanmathsinitiative/R-Instat/files/3144139/mooringsplottest.zip)

Main plot is from Moorings and is a boxplot of cumsum against dekad_abbr.

Then add  2 more layers: first is line plot and is for the newyear data frame.  y  is cumsum and x is dekad
second is point plot with the same x and y.  Change the size of the points to 3.5 and the colour to blue.
You should now get the plot above.
(If not, then perhaps the filter needs to be set again for Moorings.)

Now go in and edit both the 2 additional layers and change the x variable to dekadoct.  Do it for both together (one at a time and it is ok).  Now do the plot again.

The blue points have now gone, the order of these layers is now changed and neither can be edited further.

I hope this helps!

Ivanluv:
@maxwellfundi can I take this up?

dannyparsons:
This relates to editing layers issue so good if you can look at this as well.

volloholic:
There are probably now multiple issues mixed up here. I looked quickly at #5397  and #4899 with @rdstern  and the problem was simply that the main geom has not been assigned a position when first set. This is the iposition parameter when adding the geom to the base operator. 
We will follow up this issue over the next day or two.

PR #5426, opened 2019-07-16
removing the message check

Ivanluv:
@dannyparsons  is this  whats required for issue #5321?


volloholic:
@Ivanluv here you are simply removing the message that will not change functionality.

Ivanluv:
removing this enables the function to be found even after the order of the layers have been swapped.The order before swapping might be say `2.geom_line` and `3.geom_point`.After swapping it will be `2.geom_point` and `3.geom_line` but the listViewItem will still have them as before swapping .So on editing the condition
will search for `2.geom_point` while it is `3.geom_point ` in the listViewItems hence the message on only editing layers from the main dialog.So removing the numbering before the functions might solve this

maxwellfundi:
@Ivanluv  The concerns raised by @rdstern   #5441 from @dannyparsons  fixes them so am 
closing this.

Issue #5431, opened 2019-07-18
Line plot has multiple plot bug in Version 5.1


rdstern:
**Describe the bug**
In the graphs you have the option of Single or Multiple y.  In the line plot:
a) Complete it with single - I think any x and y.  Then close. Then return to the dialogue it has lost the y and the multiple is now there.

Also the multiple still says Single - which is a bug on other graphics dialogues too.  And on the line plot there is a funny area added half into the control.

**Screenshots**
If applicable, add screenshots to help explain your problem, particularly of error messages.

![image](https://user-images.githubusercontent.com/11226469/61436397-984aab00-a932-11e9-82a0-d4bc6fab0427.png)

Ivanluv:
@dannyparsons  I think commenting of the `SetReceiverStatus()` on the load event of the 

`ucrVariablesAsFactors` is the cause of this bug

maxwellfundi:
This seems to have been fixed with #5432  from @Ivanluv

PR #5432, opened 2019-07-18
Line plot changes

Ivanluv:
Removing the unused lblGroup on issue #5431

### DIFF
--- a/instat/dlgLinePlot.designer.vb
+++ b/instat/dlgLinePlot.designer.vb
@@ -44,7 +44,6 @@ Partial Class dlgLinePlot
         Me.cmdOptions = New System.Windows.Forms.Button()
         Me.cmdLineOptions = New System.Windows.Forms.Button()
         Me.lblFactorOptional = New System.Windows.Forms.Label()
-        Me.lblGroup = New System.Windows.Forms.Label()
         Me.lblGroupLine = New System.Windows.Forms.Label()
         Me.ucrChkValley = New instat.ucrCheck()
         Me.ucrChkPeak = New instat.ucrCheck()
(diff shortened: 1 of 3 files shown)


Wycklife:
This seems okay from my end

dannyparsons:
@Ivanluv this did force the text to be set correctly on reopen. But we should understand why it was changing when it reopened in the first place? Because this is the root of the problem and it may have other consequences. Its important generally that we understand why its working as it is even if we can fix it without understanding.
It's also better coding practice to not repeat code unnecessarily so it would be better if we can find a solution which doesn't require setting the text in a second place on load.

I had a look into this. In the sub `SetReceiverStatus()` this is only place currently where the text for the button gets set. I put a break point there but surprisingly it didn't run when I reopened the dialog, yet the text was still changing back to single. That told me the problem must be in some internal code, and probably related to automatic "translation" which sets it back to "Single Variable". On the designer for the control the button has text "Single Variable" hard coded as the button text, I think this was being put back in every time the dialog reopens. So I removed that text from the designer and now it seems to be working correctly, with no need to set the text again on load.

It was quite a difficult bug to find because it relates to the auto translation that's set up, but I wanted to explain the process of trying to get the root cause of a bug before adding more code to "cover up" the problem. I think by finding the main cause of this issues we've now got to a cleaner solution in the code.

I've sent the changes to your branch. Please have a look and check its working correctly as expected.

dannyparsons:
My changes can be seen here https://github.com/Ivanluv/R-Instat/pull/41/commits/cb25eb6a9e98d32177e7f745fd77cbbd24433576

Ivanluv:
@dannyparsons  thank you for explaining the process,it will help definitely be of  importance to me next time am faced with a similar issue.
The changes you have just made are working perfectly

PR #5436, opened 2019-07-22
Added the scale_x_date function

### LINKED ISSUE #5266: Add scale_x/y_dates option to Plot Options sub dialog
Here is a useful graph that is really simple to do, with a line plot
image

I would now like to scale the x-axis differently, giving a label every (say) 5 years, rather than every 20 years.
In the sub-dialogue, we therefore need to be able to specify the limits and the increment for dates.  Currently, the x-axis has 3 options, i.e. Continuous, Discrete, and Date.  But the Date options have yet to be completed

a) Please could they be, even if initially only partially.  And initially perhaps for the x-axis, though it should be the same for x and y.

In the meantime, I assume I could alter the script file.  I assume that is an easy thing to add?

And, by the way, in the Rwanda workshop we were asked the same question for the inventory plot.  It will greatly enhance the inventory plots to have access to the sub-dialogues.

Ogik99:
fixes #5266 
 
Work in progress

### DIFF
--- a/instat/clsGgplotDefaults.vb
+++ b/instat/clsGgplotDefaults.vb
@@ -94,6 +94,17 @@ Public Class GgplotDefaults
         End Get
     End Property
 
+    Public Shared ReadOnly Property clsXScaleDateFunction As RFunction
+        Get
+            Dim clsXScaleDateFunctionTemp As New RFunction
+
+            clsXScaleDateFunctionTemp.SetPackageName("ggplot2")
+            clsXScaleDateFunctionTemp.SetRCommand("scale_x_date")
+
+            Return clsXScaleDateFunctionTemp
+        End Get
+    End Property
+
     Public Shared ReadOnly Property clsYScaleDateFunction As RFunction
         Get
             Dim clsYlabScalesDateTempFunc As New RFunction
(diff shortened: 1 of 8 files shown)


Ogik99:
Initially implemented the breaks and limits designer. Any reviews @africanmathsinitiative/developers on the designer changes.
date breaks

dannyparsons:
The limits will need to be dates, so wouldn't date pickers be better?
And breaks will need to be a list of dates so not quite clear how to make that easy to input.

rdstern:
Date pickers would be nice.
Can we have our main option then being being date_breaks, e.g. date_breaks(width = "1 month")

 Then we have one of the dreaded up-down controls to give the number and a combo box to say what the number is of.  The options are "sec", "min", "hour", "day", "week", "month", "year".  Default might be "year" for us.

maxwellfundi:
Yea this is entirely possible. We thought we could have an input to type in
the breaks but later one thought this would be ideal to minimise errors
whilst typing.

maxwellfundi:
@Ogik99  this is making good progress, it would be great to have it for next version.

shadrackkibet:
Lots of conflicting files here, @Ogik99 can you do something?

lloyddewit:
@Ivanluv Did you already solve with with your recent work? If so, can we close this PR? thanks

Ivanluv:
@lloyddewit  I fixed this with PR #5803

lloyddewit:
@Ivanluv fixed this with PR #5803

PR #5437, opened 2019-07-24
removed id.vars from stacking data for multiple variables

dannyparsons:
This removes one potential cause of an error when using multiple variables in a plot. 
With this change, all columns are now carried when stacking so we do not need to keep track of all columns used in the dialog. Errors such as "column not found in data" should now be eliminated.


dannyparsons:
This means we also should have no reason to have the "Plot Options" button disabled when using multiple variables now.

maxwellfundi:
Great! This is a nice enhancement

PR #5441, opened 2019-07-25
graphics layers bug fixes

### LINKED ISSUE #5321: Order of layers still swapping on ggplot

dannyparsons:
fixes #5321 

@africanmathsinitiative/developers this is ready for review and will need much testing

@Ivanluv I have built on your attempt in #5426 to remove the message. Looking at the code of the control, it needed some reworking to work properly so I've made more substantial changes to the layers control and simplified it hopefully.

Also, see what I did on the boxplot dialog. One thing was to ensure all geom parameters have a position, so the layers order correctly. The second was to make variable for all the geom parameter names and then pass this list to the plot subdialog. This is to ensure that you can't delete main dialog layers from the sub dialog.
This will need to be done on every plot dialog to work correctly. Maybe you could look at this with @shadrackkibet?

### DIFF
--- a/instat/dlgBoxPlot.vb
+++ b/instat/dlgBoxPlot.vb
@@ -46,6 +46,12 @@ Public Class dlgBoxplot
     ' Jitter function that can be added to the boxplot/violin base layer
     Private clsAddedJitterFunc As New RFunction
 
+    'Parameter names for geoms
+    Private strFirstParameterName As String = "geomfunc"
+    Private strStatSummaryParameterName As String = "stat_summary"
+    Private strAdditionalPointsParameterName As String = "add_jitter"
+    Private strGeomParameterNames() As String = {strFirstParameterName, strStatSummaryParameterName, strAdditionalPointsParameterName}
+
     Private Sub dlgBoxPlot_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
             InitialiseDialog()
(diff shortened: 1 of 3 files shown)


shadrackkibet:
Testing this now.

shadrackkibet:
I also remember there is this annoying bug we experienced at TMA with Plot options when It is used in Climatic Boxplots then you try to open it in other dialogs, facets are disabled. Resetting doesn't help,

dannyparsons:
Yes thanks, I've fixed that one in another pull request that I will push shortly.

shadrackkibet:
I did quick testing of this,  seems to be working well for me.

rdstern:
It seems to be working well for me too.  This is a great improvement.

shadrackkibet:
There is this Multiple receiver issue resetting back to a single receiver sometimes when you move to the sub dialog and back, I don't remember exactly how this happens. Right now when you activate multiple receiver and leave it empty then close the dialog and then reopen, it resets back to the single receiver unlike when you have a variable in.

dannyparsons:
ok thanks, that will be good to look at in another pull request.

Ivanluv:
> There is this Multiple receiver issue resetting back to a single receiver sometimes when you move to the sub dialog and back, I don't remember exactly how this happens. Right now when you activate multiple receiver and leave it empty then close the dialog and then reopen, it resets back to the single receiver unlike when you have a variable in.

I think the bug from activating  the multiple receiver and leaving it empty then on reopen it sets back to the single receiver is from the `setControlValue()` method in the ucrVariableAsFactor. This is because on reopen, the method is checking if the is any current  variables present and if true it is setting the receiver to Multiple otherwise its setting it to single

shadrackkibet:
Are you able to fix it?

Ivanluv:
I Think it requires  discussion to make any code changes

PR #5446, opened 2019-07-26
Climate normals

### LINKED ISSUE #5310: Add further missing option to summaries (and climatic summaries

shadrackkibet:
Fixes #5310

### DIFF
--- a/instat/ucrInput.vb
+++ b/instat/ucrInput.vb
@@ -335,7 +335,7 @@ Public Class ucrInput
         Dim dcmText As Decimal
         Dim iType As Integer = 0
 
-        If strText <> "" AndAlso strValuesToIgnore IsNot Nothing AndAlso Not strValuesToIgnore.Contains(strText) Then
+        If strText <> "" Then
             If Not IsNumeric(strText) Then
                 iType = 1
             Else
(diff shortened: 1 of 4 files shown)


instat/ucrInput.vb
```diff
@@ -335,7 +335,7 @@ Public Class ucrInput
         Dim dcmText As Decimal
         Dim iType As Integer = 0
 
-        If strText <> "" AndAlso strValuesToIgnore IsNot Nothing AndAlso Not strValuesToIgnore.Contains(strText) Then
+        If strText <> "" Then
```
shadrackkibet: @dannyparsons this is where the problem is with the input textboxes not detecting non-numerics. It is because `strValuesToIgnore` is always nothing. This seems to sort out the issue but not sure if this is the right way to fix it.

shadrackkibet:
Fixes #5310

instat/ucrInput.vb
dannyparsons:
```suggestion
        If strText <> "" AndAlso (strValuesToIgnore Is Nothing OrElse (strValuesToIgnore IsNot Nothing AndAlso Not strValuesToIgnore.Contains(strText))) Then
```
Thanks, can you try this? We still want to check for values to ignore, but only its not nothing.

PR #5453, opened 2019-07-29
corrected spells calculations

dannyparsons:


rdstern:
I have been checking the spells, now merged into the master.

As far as I can tell on the first check, the unconditional spells is fine (I have yet to add missing, but so far so good even on that).  But not the conditional spells.  
They should start with a spell of 0 on the day before the period starts.  So they can't be more than the unconditional spells.  Now they are.  
A good test is the longest spell in 1 November to 31 December at Dodoma.  This seems ok for conditional when the longest spell is the one at the start of the period.  But 1951 is odd where the conditional spell is 23 days and unconditional is 21 days.  The conditional can't be longer than the unconditional.
If the longest spell is Dodoma Jan 1 to March 31, the difference is greater.  Some years on conditional are correct, e.g. 1935.  But I can't see where the 29 days comes from in 1936, when unconditional is (correctly) 12 days.  Then 1937 is ok and in the middle of the period.  In 1938 the unconditional is correct (as usual - 24 days), and the conditional should be 21, but is 22 days instead?  Then 1939 to 1941 are ok.
Looking at the longest dry spells in April perhaps adds another clue.  In 1941 the result is (correctly again) 34 days for the unconditional.  But it is 41 days for the conditional.  The conditional can never be longer than the period (i.e. 30 days).

PR #5456, opened 2019-07-30
Graphs edits

maxwellfundi:
@dannyparsons I have now gone through your boxplot to see what was done about the layers stuff and started implementations on other dialogs on  graphs. Please have a look at this just to counter check that this is whats needed.


dannyparsons:
Remember to check that all geoms have a position in the ggplot. I think its bar chart where the layer doesn't have a position, which will mess up the ordering.

maxwellfundi:
Okay great, will do.

maxwellfundi:
As well as the ones we have used the `SetParameter` which does not have a position parameter especially the ones passed in  check boxes like the `geom_smooth `on scatter plot.

dannyparsons:
If passing in a parameter object then there's no position, but it can be added to the parameter before adding to the function.

PR #5458, opened 2019-07-30
Additional feature conversions dialog

### LINKED ISSUE #5395: Additional feature - Conversions dialog or elsewhere
I have stations catalogue and I would like to map all stations at KMD. Below is a section of the data.

image

I would like to have a facility in R-Instat to convert Degrees and minutes + seconds(when available) to latitude and longitudes.

This is possible in R using `dms2dd()` function from **biogeo** package.  Using `str_extract()` from **stringr** package I extract `letters (N, S, E, W)` and minutes of latitude or longitude from Minutes columns. This is already possible using Find Replace dialog . You need to know some regex to be able to give the correct pattern. How can this be made easy for users who are not familiar with Regex? Here is a section of result columns.

image

Another alternative is using  `measurements::conv_unit()` but this doesn't deal with "E/W" "N/S" entries, it requires positive/negative degrees.

I suggest this as another option in the conversions dialog or we can have this as a special dialog under mapping section? @rdstern @dannyparsons what are your views?

shadrackkibet:
@africanmathsinitiative/developers this is ready for review. Fixes #5395. Here is a dataset to test this dialog [synoptic_stations.zip](https://github.com/africanmathsinitiative/R-Instat/files/3447019/synoptic_stations.zip). Below is the design. @dannyparsons  @rdstern any comments?
image

### DIFF
--- a/instat/dlgConversions.resx
+++ b/instat/dlgConversions.resx
@@ -135,10 +135,10 @@
     <value>NoControl</value>
   </data>
   <data name="rdoDayLength.Location" type="System.Drawing.Point, System.Drawing">
-    <value>207, 21</value>
+    <value>170, 21</value>
   </data>
   <data name="rdoDayLength.Size" type="System.Drawing.Size, System.Drawing">
-    <value>91, 28</value>
+    <value>77, 28</value>
   </data>
   <assembly alias="mscorlib" name="mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" />
   <data name="rdoDayLength.TabIndex" type="System.Int32, mscorlib">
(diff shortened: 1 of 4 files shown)


dannyparsons:
Maybe change the direction label to `Direction (N/S E/W)`

dannyparsons:
Otherwise looks good.

shadrackkibet:
done

instat/dlgConversions.resx
```diff
@@ -345,59 +345,641 @@
...
+    <value>262, 228</value>
+  </data>
+  <data name="lblLetters.Size" type="System.Drawing.Size, System.Drawing">
+    <value>104, 13</value>
+  </data>
+  <data name="lblLetters.TabIndex" type="System.Int32, mscorlib">
+    <value>12</value>
+  </data>
+  <data name="lblLetters.Text" xml:space="preserve">
+    <value>Direction(N/S E/W):</value>
```
dannyparsons: sorry just add a space here 
```suggestion
    <value>Direction (N/S E/W):</value>
```

PR #5462, opened 2019-07-31
Added a circular keyboard to the calculator

### LINKED ISSUE #5332: Adding the circular mean, etc into R-Instat
I write this issue in detail.  I have added it (temporarily) to the first milestone to ensure a quick discussion on the strategy proposed below - rather than the programming.  And the results may be useful for a current MSc essay, which is one reason for the urgency.

I start with the circular mean - which is useful in its own right - later there should be a Circular keyboard for the calculator and a Circular tab for the **Prepare > Column: Reshape > Column Summaries**.  This could even start with the single calculation of the mean.

One question is whether we use a package, or write our own functions.  There are 2 packages for circular statistics, namely **circular** and **CircStats**.  They have lots of things, but neither seem ideal for the simple calculations.  Circular makes you define a new data type - which might be useful - that's a point for discussion.  CircStats is much simpler, but their function for the mean doesn't seem to include the adjustment na.rm=TRUE/FALSE to allow for missing values.

[Here](https://stackoverflow.com/questions/32404222/circular-mean-in-r) is an article on StackOverflow that explains what is happening and usefully discussed (and solves) what happens when the answer is negative!

And Wikipedia has a really simple example to try [here](https://en.wikipedia.org/wiki/Mean_of_circular_quantities). 

If we were to write our own functions, then I am sure we could get Sam to write them for us.

The example I have is in degrees, while the R functions work in radians.  The examples in Wikipedia are 10, 20, 30 degrees, with a mean of 20 degrees.  Then subtract 15 to give 355, 5, 15 and a mean of 5 degrees.

Here is the formula:

a) sinpart <- mean(sin(wd*2*pi/360),na.rm=TRUE)
b) cospart <- mean(cos(wd*2*pi/360),na.rm=TRUE)
Then: c) circmean=atan2(sinpart,cospart)*360/(2*pi)

We could presumably assume the data are in radians only?  We shouldn't fix on degrees, because the circle could also be 360 degrees, or 366 days, or 12 months, or 24 hours.  Perhaps our function could include a Units argument?

So I can now use **Prepare > Column: Calculations > Calculate** to get the circular mean.  Could I also use **Prepare > Column: Reshape > General Summaries** to get the means per day from hourly data, which is what I need now?

What should our strategy be.  I was assuming before that we will probably make good use of the circular (and possibly circstats) packages.  We may do, for the distributions and fitting parts.  But I am now wondering what is for the best for the summary statistics calculations?

Wycklife:
Fixes #5332

In progress

### DIFF
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -1768,7 +1768,7 @@ Public Class ucrCalculator
 
     Private Sub cmdCircMean_Click(sender As Object, e As EventArgs) Handles cmdCircMean.Click
         If chkShowArguments.Checked Then
-            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::mean.circular(x = , na.rm=FALSE, control.circular=list(), ...)", 44)
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::mean.circular(x = , na.rm=FALSE, control.circular=list(), ...)", 45)
         Else
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::mean.circular()", 1)
         End If
@@ -1776,7 +1776,7 @@ Public Class ucrCalculator
 
     Private Sub cmdCircMedian_Click(sender As Object, e As EventArgs) Handles cmdCircMedian.Click
         If chkShowArguments.Checked Then
-            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::median.circular(x = , na.rm = FALSE, ...)", 21)
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::median.circular(x = , na.rm = FALSE, ...)", 22)
         Else
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::median.circular()", 1)
         End If
@@ -1784,7 +1784,7 @@ Public Class ucrCalculator
 
     Private Sub cmdMedianHL_Click(sender As Object, e As EventArgs) Handles cmdMedianHL.Click
         If chkShowArguments.Checked Then
-            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::medianHL.circular(x = , na.rm=FALSE, method=c(HL1,HL2,HL3), prop=NULL)", 47)
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::medianHL.circular(x = , na.rm=FALSE, method=c(HL1,HL2,HL3), prop=NULL)", 49)
         Else
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::medianHL.circular()", 1)
         End If
@@ -1792,7 +1792,7 @@ Public Class ucrCalculator
 
     Private Sub cmdCircRange_Click(sender As Object, e As EventArgs) Handles cmdCircRange.Click
         If chkShowArguments.Checked Then
-            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::range.circular(x = , test=FALSE, na.rm = FALSE, finite = FALSE, control.circular=list(), ...)", 73)
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::range.circular(x = , test=FALSE, na.rm = FALSE, finite = FALSE, control.circular=list(), ...)", 75)
         Else
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::range.circular()", 1)
         End If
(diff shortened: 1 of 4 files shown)


Wycklife:
@africanmathsinitiative/developers This is ready for review. Someone can have a look at it.

instat/ucrCalculator.vb
```diff
@@ -1744,4 +1765,108 @@ Public Class ucrCalculator
...
+        If chkShowArguments.Checked Then
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::var.circular(x = ,  na.rm = FALSE, ...)", 23)
+        Else
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::var.circular()", 1)
+        End If
+    End Sub
+
+    Private Sub cmdA1_Click(sender As Object, e As EventArgs) Handles cmdA1.Click
+        If chkShowArguments.Checked Then
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::besselI(x = , nu, expon.scaled = FALSE)", 28)
```
shadrackkibet: This is not the correct function.

shadrackkibet: Tooltip is also needed for this.

instat/ucrCalculator.vb
```diff
@@ -1744,4 +1765,108 @@ Public Class ucrCalculator
...
+        If chkShowArguments.Checked Then
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::angular.deviation(x = , na.rm = FALSE)", 17)
+        Else
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::angular.deviation()", 1)
+        End If
+    End Sub
+
+    Private Sub cmdCircQ1_Click(sender As Object, e As EventArgs) Handles cmdCircQ1.Click
+        If chkShowArguments.Checked Then
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::quantile.circular(x = , probs = seq(0, 1, 0.25), na.rm=FALSE, names = TRUE, type = 7, ...)", 69)
```
shadrackkibet: `probs=0.25`

instat/ucrCalculator.vb
```diff
@@ -1744,4 +1765,108 @@ Public Class ucrCalculator
...
+        Else
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::quantile.circular()", 1)
+        End If
+    End Sub
+
+    Private Sub cmdCircMin_Click(sender As Object, e As EventArgs) Handles cmdCircMin.Click
+        If chkShowArguments.Checked Then
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::quantile.circular(x = , probs = seq(0, 1, 0.25), na.rm=FALSE, names = TRUE, type = 7, ...)", 69)
```
shadrackkibet: `probs=0`

instat/ucrCalculator.vb
```diff
@@ -1744,4 +1765,108 @@ Public Class ucrCalculator
...
+        Else
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::quantile.circular()", 1)
+        End If
+    End Sub
+
+    Private Sub cmdCircMax_Click(sender As Object, e As EventArgs) Handles cmdCircMax.Click
+        If chkShowArguments.Checked Then
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::quantile.circular(x = , probs = seq(0, 1, 0.25), na.rm=FALSE, names = TRUE, type = 7, ...)", 69)
```
shadrackkibet: `probs=0.75`

instat/ucrCalculator.vb
```diff
@@ -1744,4 +1765,108 @@ Public Class ucrCalculator
...
+        Else
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::quantile.circular()", 1)
+        End If
+    End Sub
+
+    Private Sub cmdCircQ3_Click(sender As Object, e As EventArgs) Handles cmdCircQ3.Click
+        If chkShowArguments.Checked Then
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("circular::quantile.circular(x = , probs = seq(0, 1, 0.25), na.rm=FALSE, names = TRUE, type = 7, ...)", 69)
```
shadrackkibet: `probs=0.75`

shadrackkibet:
@dannyparsons this is now ready for you,

Ateamate:
All the requested buttons have been added. I've tested them and they are all functioning well.

PR #5471, opened 2019-08-13
Enabling plot options

### LINKED ISSUE #5438: Enable Plot Options sub dialog for multiple variables
I think we can now enable the Plot Options sub dialog on graphics dialog when we have multiple variables. Facets and other options should now work correctly when stacking.

I am not sure if we can also enable the Layer Options, this could be tested but I suggest leaving it disabled for now.

Ivanluv:
fixes #5438
@africanmathsinitiative/developers  this can be reviewed


dannyparsons:
@africanmathsinitiative/developers can someone review this?

rdstern:
I have had a quick look.
I did a simple line graph graph with 2 y-variables and all ok.
I then added points and fitted lines and all ok.
Then I went back to the dialogue and it warned me (3 times) I assume once for each layer, that I had multiple layers and may want to reset.  I didn't want to.  The graph was fine.
I then did my old trick of going back to single and added a horizontal line.  Then returned to multiple.  It then complained again, but then did it fine!

I will now test the other aspects.  I really like the general facility.  But I was hoping to add layers - though I realise the interpretation may not always be clear.

maxwellfundi:
@dannyparsons  this seems okay now.

dannyparsons:
This is only for the line plot dialog. Is that the only one with plot options not enabled now? Or is the same needed for the other dialogs?

Ivanluv:
Its needed also for the histogram and boxplot dialogs.I have enabled the plot options there too

Ivanluv:
this can be closed

PR #5473, opened 2019-08-16
Added column chart to dlg Bar and Pie Chart

### LINKED ISSUE #5464: Adding a column chart button to the bar chart dialogue

image

Ogik99:
fixes #5464 

### DIFF
--- a/instat/dlgBarAndPieChart.vb
+++ b/instat/dlgBarAndPieChart.vb
@@ -366,20 +366,27 @@ Public Class dlgBarAndPieChart
     End Sub
     Private Sub ChangeReceiverLocation()
         If rdoBarChart.Checked Then
+            lblVariable.Text = "Variable"
             lblVariable.Location = New Point(255, 76)
             ucrReceiverFirst.Location = New Point(255, 91)
+            lblByFactor.Text = "By Factor (Optional):"
             lblByFactor.Location = New Point(255, 126)
             ucrReceiverByFactor.Location = New Point(255, 141)
+            lblPosition.Text = "Position"
             lblPosition.Location = New Point(255, 171)
             ucrInputBarChartPosition.Location = New Point(255, 184)
         ElseIf rdoColumnChart.Checked Then
+            lblYvariable.Text = "Y Variable:"
             lblYvariable.Location = New Point(255, 76)
             ucrReceiverY.Location = New Point(255, 91)
+            lblVariable.Text = "X Variable:"
             lblVariable.Location = New Point(255, 126)
             ucrReceiverFirst.Location = New Point(255, 141)
+            lblByFactor.Text = "By Factor (Optional):"
             lblByFactor.Location = New Point(255, 171)
             ucrReceiverByFactor.Location = New Point(255, 184)
         ElseIf rdoPieChart.Checked Then
+            lblVariable.Text = "Variable"
             lblVariable.Location = New Point(255, 76)
             ucrReceiverFirst.Location = New Point(255, 91)
         End If
--- a/instat/dlgBarAndPieChart.vb
+++ b/instat/dlgBarAndPieChart.vb
@@ -67,7 +67,7 @@ Public Class dlgBarAndPieChart
         ucrPnlOptions.AddRadioButton(rdoPieChart)
         ucrPnlOptions.AddRadioButton(rdoColumnChart)
         ucrPnlOptions.AddParameterPresentCondition(rdoPieChart, "coord_polar")
-        ucrPnlOptions.AddParameterPresentCondition(rdoBarChart, "coord_polar", False)
+        ucrPnlOptions.AddParameterPresentCondition(rdoBarChart, "geom_bar")
         ucrPnlOptions.AddParameterPresentCondition(rdoColumnChart, "geom_col")
 
         ucrPnlOptions.AddToLinkedControls({ucrChkFlipCoordinates}, {rdoColumnChart, rdoBarChart}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
--- a/instat/dlgBarAndPieChart.vb
+++ b/instat/dlgBarAndPieChart.vb
@@ -76,6 +76,9 @@ Public Class dlgBarAndPieChart
         ucrPnlOptions.AddToLinkedControls(ucrReceiverByFactor, {rdoBarChart}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
         ucrReceiverByFactor.SetLinkedDisplayControl(lblByFactor)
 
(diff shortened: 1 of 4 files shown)


instat/dlgBarAndPieChart.vb
```diff
@@ -273,6 +333,21 @@ Public Class dlgBarAndPieChart
             If Not {"factor", "logical"}.Contains(ucrReceiverFirst.strCurrDataType) Then
                 ucrReceiverFirst.Clear()
             End If
+        ElseIf rdoColumnChart.Checked Then
+            clsRggplotFunction.AddParameter("mapping", clsRFunctionParameter:=clsColAesFunction, iPosition:=1)
+            cmdPieChartOptions.Visible = False
+            cmdBarChartOptions.Visible = False
+            cmdColumnChartOptions.Visible = True
+            ucrReceiverFirst.Visible = True
```
shadrackkibet: Why are is this done manually? I think it should be done through the linking system.

maxwellfundi: @Ogik99 ???

Ogik99: This was amended.

Ogik99:
can this be reviewed further @africanmathsinitiative/developers

maxwellfundi:
@Ogik99  I was looking at this and pretty much looks okay. I get one error though- 

image

You need to check the conditions for the panel for the radio buttons, especially for the new button! 

Secondly- some minor designer issue 

image

One more designer issue


image

Else everything else looks okay :-)

instat/dlgBarAndPieChart.vb
```diff
@@ -63,15 +65,20 @@ Public Class dlgBarAndPieChart
 
         ucrPnlOptions.AddRadioButton(rdoBarChart)
         ucrPnlOptions.AddRadioButton(rdoPieChart)
+        ucrPnlOptions.AddRadioButton(rdoColumnChart)
         ucrPnlOptions.AddParameterPresentCondition(rdoPieChart, "coord_polar")
         ucrPnlOptions.AddParameterPresentCondition(rdoBarChart, "coord_polar", False)
+        ucrPnlOptions.AddParameterPresentCondition(rdoColumnChart, "coord_polar", False)
```
maxwellfundi: This is actually the issue @Ogik99  you need to give the condition also for when the rdoColumnChart is checked... 

I see you are adding a parameter - `geom_col` on the basefunction when this rdoColumnChart  is checked so you need to have a condition for that- 
` ucrPnlOptions.AddParameterPresentCondition(rdoColumnChart, "geom_col")`

Ogik99: I managed to amend this.

rdstern:
[Climatic-guide_datasets.zip](https://github.com/africanmathsinitiative/R-Instat/files/3873773/Climatic-guide_datasets.zip)

@Ogik99 this is looking good.  But there is still some tidying to be done and then some re-arrangements so it is easy to get the charts you want.  In the attached file that has somehow disappeared from the R-Instat library - should go in the Climatic directory - perhaps ask @maxwellfundi about this.

Then I tried with the file called samrain and here is an example of a column chart for start1 against year.  
image

This is the sort of thing users have requested.  I did it (initially) through the bar chart options.  You might like to try it that way too so you see what is to be made easier through the column chart options.

So with bar chart set the variable is **_year_**.  This works with ok, but you just get lines each year of height 1.  Not very interesting!
So go to the bar chart options and set the geom parameter "stat" to identity.  Leave the width for now, but later you could change it from 0.9 to 0.1 for the figure above.
Now, with the stat set to identity you need a y-variable.  So go (still in the bar-chart options) to the Aeshetics and make the variable Start1 as the y.  
Now press ok.  
So this is what you need to make easy with the column chart option.  Note that the column chart is simply the bar chart with a y-variable and the stat set to "identity".

a) One detail is that when you return to the dialogue it does not remember that it is set to Column.  It resets to bar chart.
b) Next is the variables that are needed, and those that are optional.  This works directly in your dialogue:
image

So, all that is changed is the width and you get the graph above.  Great.

So I suggest you change the order of the variables for the column chart, so it is the same, and possibly using the same code as the scatterplot or line plot dialogues, i.e.
image

I am not sure about the multiple option, but it may do no harm.  Work with @maxwellfundi on the code that is needed.
So you are swapping round the x and the y variables in your dialogue and the factor (Optional comes below those.
Both the y and the x are needed.  So OK is only enabled when they are both completed.

Good luck!

Roger

rdstern:
That's much better and easier to use now.
One aspect that still remains is that the dialogue still does not "remember" that it is on a column chart.  So, when I return to the dialogue after a column chart, it has reverted to being a bar chart.  I assume this is easy to fix, because it does remember to remain on a pie chart.  So it should just be a question of adapting that code?

One other small change.  There is an **_Options_** Button just below the Column Chart Options.  In the other dialogues this is called Plot options.  Could you please add the Plot here too for consistency.  This is for all 3 settings, i.e. Bar, Pie and Column. 

The column chart below takes some steps to produce, but is the sort of thing people seem to like.  It shows the start, end and length.  I could only do it by calculating the lengths, then stacking, then adding an extra variable for the level of alpha (transparency).
image

Ogik99:
@rdstern  the dialog now "remembers" the column chart.
I have also amended the options button name--- it now reads "plot option"

rdstern:
@Ogik99 this seems to work fine now.
One small change still, I am afraid.  The first field reads By Variable.  It is **_Y Variable:_**.
And the one underneath, where you have Variable, should be **_X Variable:_**.

If it is easy, then could you try a bigger challenge - which I think @dannyparsons may ask for.  Otherwise perhaps he can do that.  If you look at the Scatterplot or Line Plot you will now see they are very similar in this aspect, namely Y Variable, X Variable and Factor (Optional).

The other dialogues have the Y Variable as Single/Multiple.  And when you change to Multiple it also changes the Factor (Optional) so that is effectively disabled.  It would be good to swap that code into this dialogue if you can.  Maybe @Patowhiz can also help if needed.  Then we are ready for this to be merged.

Ogik99:
> @Ogik99 this seems to work fine now.
> One small change still, I am afraid. The first field reads By Variable. It is **_Y Variable:_**.
> And the one underneath, where you have Variable, should be **_X Variable:_**.
> 
> If it is easy, then could you try a bigger challenge - which I think @dannyparsons may ask for. Otherwise perhaps he can do that. If you look at the Scatterplot or Line Plot you will now see they are very similar in this aspect, namely Y Variable, X Variable and Factor (Optional).
> 
> The other dialogues have the Y Variable as Single/Multiple. And when you change to Multiple it also changes the Factor (Optional) so that is effectively disabled. It would be good to swap that code into this dialogue if you can. Maybe @Patowhiz can also help if needed. Then we are ready for this to be merged.

It's interesting that it still reads "By Variable" and "Variable" (for the column chart). I had changed these. But let me check.

Ogik99:
bar_chart
This is what I have @rdstern

rdstern:
So do I now!

Thanks

Roger

maxwellfundi:
Great looks good here :-)

PR #5475, opened 2019-08-20
Added a few keys to the circular keyboard and made the proposed changes 

### LINKED ISSUE #5332: Adding the circular mean, etc into R-Instat

Wycklife:
Fixes #5332

@africanmathsinitiative/developers Ready to be reviewed. Someone to review the changes.

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -148,7 +148,7 @@ Public Class dlgCalculator
             Case "Transform"
                 Me.Size = New System.Drawing.Size(iBasicWidth * 1.35, Me.Height)
             Case "Circular"
-                Me.Size = New System.Drawing.Size(iBasicWidth * 1.3, Me.Height)
+                Me.Size = New System.Drawing.Size(iBasicWidth * 1.39, Me.Height)
             Case "Wakefield"
                 Me.Size = New System.Drawing.Size(iBasicWidth * 1.7, Me.Height)
             Case Else
--- a/instat/ucrCalculator.Designer.vb
+++ b/instat/ucrCalculator.Designer.vb
@@ -270,6 +270,16 @@ Partial Class ucrCalculator
         Me.cmdChildren = New System.Windows.Forms.Button()
         Me.cmdHeight = New System.Windows.Forms.Button()
         Me.grpWakefield = New System.Windows.Forms.GroupBox()
+        Me.cmdLikert = New System.Windows.Forms.Button()
+        Me.cmdCar = New System.Windows.Forms.Button()
+        Me.cmdAnswer = New System.Windows.Forms.Button()
+        Me.cmdPet = New System.Windows.Forms.Button()
+        Me.cmdAnimal = New System.Windows.Forms.Button()
+        Me.cmdAge = New System.Windows.Forms.Button()
+        Me.cmdIncome = New System.Windows.Forms.Button()
+        Me.cmdHair = New System.Windows.Forms.Button()
+        Me.cmdGroup = New System.Windows.Forms.Button()
+        Me.cmdGrade = New System.Windows.Forms.Button()
         Me.grpCircular = New System.Windows.Forms.GroupBox()
         Me.cmdCircQ3 = New System.Windows.Forms.Button()
         Me.cmdCircQ1 = New System.Windows.Forms.Button()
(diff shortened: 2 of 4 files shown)


rdstern:
@Wycklife In the calculator, circular keyboard there is a further key to add. 

I wrote before - but well hidden!  
"The first key, circular has default code:
circular::circular( ,modulo= "2pi",units="radians")"

I think it could now become the last key, that just fills the set of 4 by 4.

Wycklife:
@rdstern Should I add more parameters or just these two?

Wycklife:
@rdstern I've made the proposed changes. I think it's now okay.

Wycklife:
@africanmathsinitiative/developers Someone can review this

rdstern:
ok.  I'll check now

rdstern:
1. The first command circular: I think just the 2 arguments I have mentioned would be better.  I am checking it now.
In the first key - circular the default should be 
circular::circular(xxx,modulo="2pi")
Then, when you give the arguments, I think you have missed the quotes, e.g. "degrees".
Perhaps the default command (when arguments isn't ticked) could be given as:
circular::circular(xxx, modulo="2pi",units="radians")
Then it will be clear what a user should change if the data are in degrees or hours.

2. When the arguments are given, then for the mean you have (correctly) made na.rm = TRUE.  This needs to be done for all the other summaries.
3. The bottom row of keys doesn't work when the arguments checkbox is not ticked.  It needs the probs argument anyway so probs = 0 for the min, 0.25 for q1, 0.75 for q3 and 1 for the max.
4. It also needs to be added for the quantile key.
5. Lastly - the medianHL key needs to move down so it is in line with the others

Wycklife:
@rdstern Thanks for the observation, I'm making the changes.

Wycklife:
@rdstern I've implemented the changes, could you please have a look at them.

rdstern:
They all look fine.  Can you just align the medianHL key label, so it is the same position as the other keys.  It is the top-right key.
Then it is ready for @dannyparsons to merge.
It looks very good.  Very well done for your work on this and also on the summaries tab.  I will now complete the chapter in the climatic guide.

Wycklife:
@dannyparsons @rdstern Should I change na.rm =TRUE to default state?

dannyparsons:
I've just commented here https://github.com/africanmathsinitiative/R-Instat/issues/5332#issuecomment-525840241. Sorry we're changing back again! 
Yes, please, we want the R default so we want `na.rm = FALSE`. Both for circular and summary keyboards.

Wycklife:
@dannyparsons This has been implemented. Kindly have a look

dannyparsons:
Great, can you also check the `na.rm` default in the normal summary keyboard? For example, Roger noticed mean had `na.rm = TRUE` which we now want as `FALSE`.

Wycklife:
Okay

Wycklife:
@dannyparsons This is now consistent, have a look

maxwellfundi:
@dannyparsons  this seems ready for you.

PR #5476, opened 2019-08-21
Added a circular tab to the Summaries subdialog

Wycklife:
This fixes the second part of #5332 

This is the designer for the circular tab


rdstern:
@Wycklife I can only look superficially and the tab looks ok to me.  None of the actual labels are yet added?
I don't know if you could work with @shadrackkibet to add one of the functions - then I presume you could proceed with the others.  The most important is the circular mean, which is the command circular::mean.circular( ).  It should work just like the ordinary mean and ideally (of course) the missing options would also apply to these summaries just as they do to the others.

shadrackkibet:
@Wycklife is it okay if I pick your branch now and work on the R code?

Wycklife:
@shadrackkibet It's okay but I want to learn how you are doing it

Wycklife:
@shadrackkibet Could we do it together, then I carry on with the rest as suggested by @rdstern

shadrackkibet:
Okay.

Wycklife:
@shadrackkibet Let me know when you're ready

shadrackkibet:
@Wycklife the Rcode is now ready [here](https://github.com/Wycklife/R-Instat/pull/102) now ready you can merge the changes. I have implemented `circular.mean` in the summaries subdialog. The other should be straight forward. Let me know if you have any questions.

Wycklife:
@africanmathsinitiative/developers This is ready for review. Someone might have a look at it

rdstern:
@Wycklife   I tried to have a quick look.  Perhaps I have done something wrong, because the tab still looks like this when it is running:
 image

I added a data set and tried ticking a few boxes - assuming perhaps the first was the mean.  I asked for the results in the output window and got ordinary summaries, but nothing circular.

Wycklife:
@rdstern  The design seems okay from my end this is how it looks like.

image

rdstern:
@wycliffe that looks nice.  I have it running now.
a) In general it seems to work well
b) I have (quickly) checked that the missing options also seem to work.  That is really great.
c) Two minor points from your figure above are the mis-alignment of the Range checkbox and the last one in the Scale group-box that still says Check Box.  I wonder what that is?
d) In the scale group-box the ang_var and the ang-dev don't work.  I suspect that is simple to fix.
d) The quantile can't work like that.  It needs the actual quantile to be given as well.  For the ordinary quantiles this is in the More tab, see below.  So perhaps move it to a 3rd group-box that says More.  For now there is only Quantile.  Then either disable it because what is working is great.  Or adapt the similar control and code in Percentile, see below.  Hope that idea is clear:
image

shadrackkibet:
> d) The quantile can't work like that. It needs the actual quantile to be given as well. For the ordinary quantiles this is in the More tab, see below. So perhaps move it to a 3rd group-box that says More. For now there is only Quantile. Then either disable it because what is working is great. Or adapt the similar control and code in Percentile, see below. Hope that idea is clear:

@Wycklife I didn't include the quantile function. Could you include that? It is just similar to what we have for `min` and `max` but now we allow changing of  `probs` parameter. Then in the dialog you add a control(input or Nud) to pass `probs` parameter.

Wycklife:
@africanmathsinitiative/developers This is now okay. Someone to review.

rdstern:
This is working well now, as far as I can see.
I suggest it is ready to go to @dannyparsons for a final look and then to merge.
I have checked with all the functions, getting circular summaries by date and all seem to work.  
I have also checked that they work with the missing value options - and that is brilliant.

There are 2 points for @dannyparsons to investigate, which he thinks are small.
a) We should have a symbol to indicate when a variable is circular.  Perhaps wind_dir (ci) or (cr)?
b) I also defined the columns as circular in the calculator.  Then the circular package captures functions it recognises.  So the mean becomes the circular mean.  But the summary has not carried this information through properly.

What this does mean is the quantile works already - though the special key isn't there!  So no need to add it.

dannyparsons:
This looks good. The missing options tab should be at the end though.

rdstern:
Agreed.  I should have thought of that!

Wycklife:
@dannyparsons Okay

Wycklife:
@rdstern It's done. This is now ready to be merged.

Wycklife:
@dannyparsons Kindly have a look at the change.

PR #5477, opened 2019-08-21
Mapping polygons

### LINKED ISSUE #5422: Finding points in a polygon?
From where we are now, this would be very useful, but I don't know how to do this. 

Here is my problem.  In each country with climatic data we are being given the station data to draw on a map.  The station file needs checking - in each case.  Often, for example in Tanzania, the station file does not include the region, or the district variables.  It would be helpful if it did, both to support checking and also for the plotting.  I have the polygons from the gadm file and also the locations (lat and long from the station file.  

So I would like to do a sort of merge.  I think the answer is [here](https://gis.stackexchange.com/questions/282750/identify-polygon-containing-point-with-r-sf-package/318629#318629)?

If so, then it may be quite easy?  It would certainly be useful and appreciated!  Not just by me!

If it is really easy then I wonder where it should go in the menu system?  It is essentially a merge, but a pretty special one.  Perhaps it should go under the mapping, which would now have a map sub-dialogue and an Add Variables dialogue.

Ivanluv:
fixes #5422

### DIFF
--- a/instat/dlgCategorizingMapsFromPoints.Designer.vb
+++ b/instat/dlgCategorizingMapsFromPoints.Designer.vb
@@ -1,0 +1,218 @@
+﻿<Global.Microsoft.VisualBasic.CompilerServices.DesignerGenerated()> _
+Partial Class dlgCategorizingMapsFromPoints
+    Inherits System.Windows.Forms.Form
+
+    'Form overrides dispose to clean up the component list.
+    <System.Diagnostics.DebuggerNonUserCode()> _
+    Protected Overrides Sub Dispose(ByVal disposing As Boolean)
+        Try
+            If disposing AndAlso components IsNot Nothing Then
+                components.Dispose()
+            End If
+        Finally
+            MyBase.Dispose(disposing)
+        End Try
+    End Sub
+
+    'Required by the Windows Form Designer
+    Private components As System.ComponentModel.IContainer
+
+    'NOTE: The following procedure is required by the Windows Form Designer
+    'It can be modified using the Windows Form Designer.  
+    'Do not modify it using the code editor.
+    <System.Diagnostics.DebuggerStepThrough()> _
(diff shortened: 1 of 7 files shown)


Ivanluv:
@dannyparsons  am stuck on the implementation of the Shape file receiver and the Station receivers  using the same selector with different data frames.Could you me in this

dannyparsons:
I thought there is a separate data frame selector for the shape file, and this is separate from the selector for the points? That's how it looks in your image here https://github.com/africanmathsinitiative/R-Instat/issues/5422#issuecomment-522481210.
If so then aren't they just separate?

rdstern:
That was ok, until we realised that we need to choose a variable from the shape file - for the names.  Then it was like the main mapping dialogue, where there are the 2 data frames.
I thought - with the single selector - we could do what you already manage in the PICSA Crops dialogue - where there are 2 data frames.  Otherwise we move to having 2 selectors - as with the main mapping dialogue.

dannyparsons:
Ah ok I see. @Ivanluv you can have a look at the PICSA crops dialog to see how to have receivers with variables from different data frames. The idea is the same.

maxwellfundi:
@dannyparsons  I have a question here, from what I have found on the PICSA crops, when we click on the second set of the receivers then we are unable to use a second dataframe. 

 On the dataframe change, even on the PICSA crops it appears to clear the receivers. I dont think this achieves our purpose as we want to use two data frames on this.

maxwellfundi:
@dannyparsons ?

dannyparsons:
You need linked data frames on this dialog. This is climatic data so the first set of receivers are daily data and the other two are yearly data for start/end of rains. Start with e.g. Dodoma and calculate start and end of rains to use the dialog.
The requirement for the second data frame to be linked to the first can be turned off, which is what we want want on this dialog.

Issue #5268, opened 2019-03-23
Clearing problems with Climatic Boxplots


rdstern:
There are some odd things happening with the clearing and updating with this dialogue.  It initially works fine, and I think it is OK after reset is pressed - I will check.

But sometimes the graph is not what I expect from the dialogue.  Here is an example.
![image](https://user-images.githubusercontent.com/11226469/54870494-1d67ee00-4d9f-11e9-8b0f-0d12dc1c9201.png)

This graph is consistent with the command generated - not surprising:
Code generated by the dialog, Climatic Boxplot
Dodoma <- data_book$get_data_frame(data_name="Dodoma")
last_graph <- ggplot2::ggplot(data=Dodoma, mapping=ggplot2::aes(y=Rain, x=as.factor(month_abbr))) + ggplot2::geom_boxplot(varwidth=TRUE, fill="green") + theme_grey() + ggplot2::theme(axis.text.x=ggplot2::element_text(angle=90)) + ggplot2::xlab("")
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="Dodoma")
data_book$get_graphs(data_name="Dodoma", graph_name="last_graph")

But that is not the same as the dialogue in 2 ways:
![image](https://user-images.githubusercontent.com/11226469/54870524-85b6cf80-4d9f-11e9-8c34-5b48fbb040d0.png)

The dialogue asked for Omit below 0.85 and also Variable width.
This was after using the dialogue before on another element and also Filtering and clearing the filter through the dialogue.  Until I changed to rainfall it worked fine.

It maybe that there is an issue in the code nit updating that is easy to spot - at least for @Muthenya working with @dannyparsons ?  I'll also see whether I can do a sequence that makes it go wrong.

rdstern:
OK, I have a simple sequence that goes wrong.  
Use the dodoma.rds data.  Complete the dialogue as above, i.e. rain, omitting > 0.85 and variable width.  It gives the graph that is needed.
Now use boxplot options, change the geom so the Fill colour = Blue.

Now return and (at least the omitting below 0.85 is now not done.
You can untick the Omit below and tick it again - no difference.

This may not be the only problem, but it is one that should not be difficult to find.

rdstern:
And another problem.
boxplots for sunshine hours, variable **sunh**
filter to **year > 1972**
x is year and month is facet wrap
boxplot colour green seems OK now.
[...]

PR #5481, opened 2019-08-28
Climatic boxplot

Ivanluv:
@dannyparsons  is this right?for issue #5268

### DIFF
--- a/instat/dlgClimaticBoxPlot.vb
+++ b/instat/dlgClimaticBoxPlot.vb
@@ -372,6 +372,7 @@ Public Class dlgClimaticBoxPlot
             End If
         Next
         'TODO need to check for the variable width from the geom function in a way that when we check the var width checkbox, it triggers controlvaluechanged
+        clsRggplotFunction.AddParameter("data", clsROperatorParameter:=clsFilteredDataOperator, iPosition:=0)
     End Sub
 
     Private Sub AutoFill()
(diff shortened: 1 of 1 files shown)


instat/dlgClimaticBoxPlot.vb
```diff
@@ -524,6 +525,7 @@ Public Class dlgClimaticBoxPlot
     End Sub
 
     Private Sub OmitFilter()
+        clsRggplotFunction.AddParameter("data", clsROperatorParameter:=clsFilteredDataOperator, iPosition:=0)
```
dannyparsons: Why does it need to be here as well?

dannyparsons:
The real solution is the change the `ucrGeomWithAes` control so that it doesn't modify the data parameter in this way. That's a bit more work to figure out, so this is ok as a temporary solution.

rdstern:
I haven't checked what is improved here.  But at least one bug remains, which is easy to show as follows. (and which came up in the ICPAC workshop.)
Use the Dodoma.rds data from the library.  Get the default boxplot for rain with the checkbox ticked, and possibly the checkbox for the  variable width as well (though that may not be relevant.)
The boxplots are nice and include the code:
`ggplot2::ggplot(data=dodoma %>% dplyr::filter(rain > 0.85) `

Now go to boxplot options > Layer Parameters > Geoms and Fill is blue
Return to the main dialogue and ok.  The code has lost the filter part above.

Even simpler press Reset and start with the Fill = Blue.   Then set the checkbox for rain > 0.85 and it is still ignored.

PR #5484, opened 2019-08-28
System Reference Error

Patowhiz:
@dannyparsons  I don't know why I'm getting this error. After merging the recent updates from the master, I'm unable to build the application due to lack of reference of RDotNet SymbolicExpression class as a requirement.

The screenshots below explain this better. I had to add the reference to the assembly to be able to build the application, @africanmathsinitiative/developers  is there anyone else getting this error? Any help will or suggestion? I had to make a separate PR for this change.

error2

error3


dannyparsons:
It's not something I've seen before. Check the Nuget package dependencies, maybe remove and re-install Rdotnet. Some restarts may also help. Next is to Google the error messages for more clues.

Patowhiz:
This was resolved when I used Visual Studio 2019

dannyparsons:
That's odd as you shouldn't need to use VS 2019, although good to have anyway. Will keep an eye out for any other issues that might relate to this.

PR #5492, opened 2019-09-02
Wake field minor changes

### LINKED ISSUE #4347: Random samples and Wakefield package
This seems a very nice package - don't know why it is called Wakefield - for the use of R-Instat in teaching.  I suggest:
a) We add it to the packages that are installed with R-Instat.  Then it can already be used in our calculator to generate interesting random samples.
b) We consider adding it in 3 places, namely:
1) In the **File > New** dialogue.
2) In the **Prepare > Generate > Random Samples** dialogue - where I suggest it should be a sub-dialogue - perhaps called Options or even Wakefield. 
3) Possibly also a keyboard in the calculator - that could be fun!  There it would generate a single sample (column) whereas the dialogues could allow the generation of multiple columns.

What I am not clear is on the structure of the sub-dialogue.  Are there any principles we could use here?

I would like this perhaps to be considered by our team in AMI (the documentation team) in discussion with David.

I suggest that, in the short term, it could become a great new keyboard in our calculator and this would be relatively simple and hence a possible job for a newcomer, perhaps under supervision from Shadrack?  One complication here is that the package suggests a sensible name for the resulting column.  We could ignore this feature or try to incorporate it.

This is also an example of a package where we could investigate how it can be used in teaching?  I like some of the functions for different types of data and also the way it suggests an automatic type fore the resulting column.  So it generates factors and ordered factors and logical columns, etc.

In the File > New and Random Samples dialogue I suggest the Climatic > Climdex sub-dialogues give a possible structure.  But I am not sure how to allow for the probabilities.  That isn't yet so easy with our dialogue structure.

Ivanluv:
fixes #4347

### DIFF
--- a/instat/ucrCalculator.Designer.vb
+++ b/instat/ucrCalculator.Designer.vb
@@ -270,6 +270,16 @@ Partial Class ucrCalculator
         Me.cmdChildren = New System.Windows.Forms.Button()
         Me.cmdHeight = New System.Windows.Forms.Button()
         Me.grpWakefield = New System.Windows.Forms.GroupBox()
+        Me.cmdLikert = New System.Windows.Forms.Button()
+        Me.cmdCar = New System.Windows.Forms.Button()
+        Me.cmdAnswer = New System.Windows.Forms.Button()
+        Me.cmdPet = New System.Windows.Forms.Button()
+        Me.cmdAnimal = New System.Windows.Forms.Button()
+        Me.cmdAge = New System.Windows.Forms.Button()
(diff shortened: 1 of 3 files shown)


dannyparsons:
@Ivanluv there are conflicts here, presumably because of other changes to this dialog by @Wycklife for circular? Please check open pull requests before working on a new dialog to avoid potential conflicts.

maxwellfundi:
@Ivanluv  the conflicts here are too many to be solved manually. I think you can just close this, then copy paste your changes on updated branch.

Ivanluv:
@maxwellfundi  I will do that

PR #5493, opened 2019-09-03
Open Library Listview Selection bug

### LINKED ISSUE #5485: Small but on Open from Library
This is annoying rather than serious.  It is on Open From Library with the R package sets.  I think it only applies with packages where there are sufficient data sets that there is scrolling.  For example the datasets package, or DAAG.  
In that case scroll and choose a package near the end.  It jumps back to the start.  If you persist and scroll again then it is now fine.

As I said it is slightly annoying, but no big deal!  To me it looks like a @Patowhiz type puzzle to fix?

Patowhiz:
Fixes #5485


rdstern:
Great.  That seems to work.  Excellent service from @Patowhiz .  Can this be merged?

PR #5500, opened 2019-09-04
Added a new menu item and a dialog

### LINKED ISSUE #5474: New menu item and new dialogue to define circular
I propose a new menu item in the Prepare menu.  It is called **_Column: Define_**.   and is below Column: Date and above Column: Reshape.
It can start with just two items.  The first is a copy (at least for now) of Prepare > Data Frame > Convert Columns and is also called **_Convert Columns_**.  Then there is a line.  Under the line is Circular.
This is a new dialogue.  There is the usual data selector restricted to numeric columns.  There is a multiple receiver with the label **To Circular**.  (I can't think of many situations when you will want more than one variable.  So if it is more work, then make it a single receiver.)

There are two sets of the ordinary radio buttons.
One is called Units.  It has 3 options, namely Radians, Degrees, Hours with Radians the default.
The second is Modulo.  It has 3 options, namely Asis, pi, 2pi with 2pi the default.

There are other arguments, but they are mainly for the plotting routines in the package and w are not using those.  So stop there, at least for now.

This just issues the command circular::circular(x, modulo="2pi", units="radians")

Our default is 2pi, so we have to give that, because the package default is asis.  Our default of radians is the same as the package.

If it is a multiple receiver then the other controls remain simple, i.e. if one variable is in degrees and the other is in radians, then you use the dialogue twice.

Wycklife:
Fixes #5474 

In progress


maxwellfundi:
merging this because it has chages on frm main and vb proj

PR #5502, opened 2019-09-04
Climsoft dialogue

### LINKED ISSUE #4915: Enhance the Climatic > File > CLIMSOFT dialogue

Ivanluv:
fixes #4915


maxwellfundi:
@dannyparsons @rdstern We discussed with @Ivanluv about having the  Unstack Data on climsoft dialog and here is what we have come up with. 

@dannyparsons please have a look at the R code to see if there is anything else we might need to add. 

@rdstern you can also have a look at how this works, it might impress you :-)

rdstern:
I will have a look.  @maxwellfundi I am generally getting impressed at all the progress that is being made.  I want to write to the Climsoft team as well, once we have this next version, so this is all good.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1574,6 +1574,11 @@ DataBook$set("public", "import_from_climsoft", function(stations = c(), elements
       my_stations = paste0("(", paste(as.character(unique(station_data$recordedFrom)), collapse=", "), ")")
       station_info <-  DBI::dbGetQuery(con, paste0("SELECT * FROM station WHERE stationID in ", my_stations, ";"))
     }
+    
+   
+    if (unstack){
+      station_data <- reshape2::dcast(data=station_data, formula=recordedFrom + obsDatetime ~ abbreviation, value.var="obsValue")
```
dannyparsons: spaces needed around `=`

This looks good to get the data in a nice format, but we are then losing some of the other columns: `describedBy` and `elementName` which we shouldn't just forget.

With this format we should also get another small data frame of element metadata with one row per element. And these two data frames are linked.
We can get this by running another database query on the `obselement` table and get the columns: `describedBy`, `abbreviation`, `elementName`.
Then these two data frames have the same information as in the stacked data format.

Does this make sense? @Patowhiz could help with constructing the query if you need help.

Patowhiz: @Ivanluv I'm happy to help with the SQL query. To get a resultset form mysql database you just need to use a select statement. 
The structure goes like 

`SELECT * FROM table_name` 

where the table_name is replaced with the real table name you wish to get the data from.

In our case the query could be something like(Assuming you don't want all the columns from the database table);

`SELECT describedBy, abbreviation, elementName FROM obselement`

Ivanluv:
@Patowhiz  where should this be run exactly within `import_from_climsoft` ?

maxwellfundi:
From Danny's explanation, my understanding was where you are stacking the
data, then you need to run the query meaning you will then have three
dataframes in the end namely station info, data , elements info.
@dannyparsons is this correct?

dannyparsons:
yes exactly, that's my understanding of what we want. @rdstern can confirm. 
There should also be a link created between the data and element info (by the element ID column)

instat/static/InstatObject/R/instat_object_R6.R
Patowhiz: @dannyparsons we are checking on this issue with @Ivanluv , checking on your comment above, about getting the element information from obselement, we would like to point out that obselement doesn't have `describeBy `column. Should we assume you only mean't `abbreviation,elementName` ?

dannyparsons: Yes, that's right thanks for clarifying. I only meant to indicate the field above which are coming from `obselement`.
Essentially, we want to make sure that whether you stack or not, you get the same information from the database (just in a different form).

Ivanluv:
@dannyparsons  are we to have every column from `obselement`?Is this implementation right?

dannyparsons:
That's a good question. We should have flexibility in choosing which columns to include in general in the dialog.
For now just include the same columns from `obselement` as in the original code. This is something we should look at in further improvements to the dialog.

shadrackkibet:
@Ivanluv what is the state of this? is testing needed?

Ivanluv:
Ive not made any changes on this

lloyddewit:
@Ivanluv Please could you clarify the status? thanks

Ivanluv:
This requires more disscussion

Patowhiz:
@Ivanluv I'm happy to be involved in those discussions. This is one of the dialogs that the ICPAC staff were interested in. And I strongly felt that it could be made much better.

Patowhiz:
@rdstern @Ivanluv what's the state of the discussion on this?

Patowhiz:
@rdstern @Ivanluv I'm also happy to look into this and help solve the remaining core items raised in issue if need be.

Patowhiz:
@Ivanluv based on the discussions I had with @rdstern and @lloyddewit, I'm now looking into the issue. I'll be creating a new PR in regards to the issue and I'm looking into taking some of the changes you have made in this PR. I can see you have spent a considerable amount of time in solving this and thus I'll be consulting you to help me on some of the changes in the new PR. Thank you.

lloyddewit:
@Patowhiz @Ivanluv I was wondering how much of this PR's functionality is covered by  PR #5960 and @Patowhiz 's other Climsoft work. The number of code changes in this PR is actaully quite small. Could @Patowhiz integrate this into his work so that we can close this PR? 
thanks

lloyddewit:
@Ivanluv I see that you closed this PR. We need to ensure that these changes are not overlooked. Please could you add a comment to this PR explaining

- if these changes are still needed
- if so, in which PR you expect the changes to be made
- if you'll make the changes yourself, or if you've coordinated with someone else to make the changes

Thanks

Ivanluv:
@lloyddewit the changes I made are no longer needed .@Patowhiz took over this task

Patowhiz:
@Ivanluv thank you for clarifying.

lloyddewit:
Understood, thanks for the clarifications :)

PR #5505, opened 2019-09-09
Minor issue#5499

### LINKED ISSUE #5499: Minor issue with day-of-year range sub-dialogue
This sub-dialogue is called by the Climatic > Prepare > Climatic Summary and by other dialogues in Climatic > Prepare.
The minor problem - unless it affects other things is the To field which is here set to the variable end_rains
image

When I return to the main dialogue it is fine:

image

Now I return to the sub-dialogue (just to check) and it has reset the **_To_** to start_rains:
image

It seems to be just on Variable Day that it remembers incorrectly.  Fixed  Day and Length seem to be ok.

Ivanluv:
fixes #5499

### DIFF
--- a/instat/sdgDoyRange.vb
+++ b/instat/sdgDoyRange.vb
@@ -128,7 +128,7 @@ Public Class sdgDoyRange
                         End If
                     End If
                     rdoToVariable.Checked = True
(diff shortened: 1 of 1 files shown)


dannyparsons:
This looks correct. @africanmathsinitiative/developers can someone test?

maxwellfundi:
Looks good!

PR #5524, opened 2019-09-19
Shadrackkibet infillmissingdates

dannyparsons:
@shadrackkibet @rdstern some minor changes to the options display and a correction on the limits with a starting month. Please check this is working ok.


instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -2562,12 +2564,12 @@ DataSheet$set("public","infill_missing_dates", function(date_name, factors, star
     else if(!missing(start_month)) {
       if(start_month <= lubridate::month(min_date)) min_date <- as.Date(paste(lubridate::year(min_date), start_month, 1, sep = "-"), format = "%Y-%m-%d")
       else min_date <- as.Date(paste(lubridate::year(min_date) - 1, start_month, 1, sep = "-"), format = "%Y-%m-%d")
-      if(start_month <= lubridate::month(max_date)) max_date <- as.Date(paste(lubridate::year(max_date), 12, 1, sep = "-"), format = "%Y-%m-%d")
-      else max_date <- as.Date(paste(lubridate::year(max_date) - 1, 12, 1, sep = "-"), format = "%Y-%m-%d")
+      if(end_month >= lubridate::month(max_date)) as.Date(paste(lubridate::year(max_date), end_month, lubridate::days_in_month(as.Date(paste(lubridate::year(max_date), end_month, 1, sep = "-", format = "%Y-%m-%d"))), sep = "-"), format = "%Y-%m-%d")
```
dannyparsons: @shadrackkibet see here that `max_date` is not always in month 12. It should be 1 month before `start_month`. See addition of `end_month` calculation further above.
Then here, the year is either the current year or the next year (depends where the data ends).
Also, the day needs to be at the end of the month, not the start so `lubridate::days_in_month` is used to calculate this.
This should now give correct limits when using `start_month`.

shadrackkibet: Thanks. That is noted.

PR #5525, opened 2019-09-20
PICSA Graphs sub dialog bug fixes

### LINKED ISSUE #5515: Updated PICSA graphs for Version 5.3
It does seem to work much better.  I tried the mean and that seemed fine.  As did the y-axis with the lower limit set to 0.
Then I wanted to change the colour of the points.  So I pressed the Plot Options and got this message:
image

It worked ok though and I could change the layer.  That was all I changed but it reset the y-axis lower limit again, so I had to put the zero back in again.  So I tried the Plot Options again, without making any change.  It gave me the above message again - and still reset the y-axis limits back to the default.

I tried a few more things staying within the PICSA options and all seems ok.  Then I went to the Line Options.  It went in ok, but now cleared the variable I was plotting.  When I put it back in again and tried to run I got the following:
image

I would be fine if we don't have that Line Options button on the PICSA dialogue.  We needed it when we couldn't edit the base layers.  Now we can.  Let's just have PICSA options and Plot Options

dannyparsons:
#5515

### DIFF
--- a/instat/dlgPICSARainfall.vb
+++ b/instat/dlgPICSARainfall.vb
@@ -602,11 +602,9 @@ Public Class dlgPICSARainfall
         If ucrVariablesAsFactorForPicsa.bSingleVariable Then
             'Added this because this sub is called on sed defaults and  it over writes the enabled property 
             cmdPICSAOptions.Enabled = True
-            cmdLineOptions.Enabled = True
             cmdOptions.Enabled = True
         Else
             'cmdPICSAOptions.Enabled = False
-            cmdLineOptions.Enabled = False
             cmdOptions.Enabled = False
         End If
     End Sub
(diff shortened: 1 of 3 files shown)


instat/dlgPICSARainfall.vb
```diff
@@ -630,7 +628,7 @@ Public Class dlgPICSARainfall
         'AddRemoveGroupByAndHlines()
     End Sub
 
-    Private Sub cmdLineOptions_Click(sender As Object, e As EventArgs) Handles cmdLineOptions.Click
+    Private Sub cmdLineOptions_Click(sender As Object, e As EventArgs)
```
shadrackkibet: Handle is missing here

dannyparsons: Thanks, the button has been removed so this sub needs to be delete.

PR #5528, opened 2019-09-23
New polar

### LINKED ISSUE #5331: Add polar to coord tab in plotting

Ivanluv:
fixes #5331 
@dannyparsons  do this changes on adding the theta option look okay?


instat/sdgPlots.vb
```diff
@@ -255,6 +260,32 @@ Public Class sdgPlots
...
+        ucrInputStartingAngle.SetLinkedDisplayControl(New List(Of Control)({lblPi, lblStartingAngle}))
+
+
+        dctTheta.Add("x", Chr(34) & "x" & Chr(34))
+        dctTheta.Add("y", Chr(34) & "y" & Chr(34))
+        ucrInputPolarCoordinates.SetParameter(New RParameter("theta", iNewPosition:=1))
+        ucrInputPolarCoordinates.SetLinkedDisplayControl(lblPolarCoordinate)
+        ucrInputPolarCoordinates.SetItems(dctTheta)
+        ucrInputPolarCoordinates.SetDropDownStyleAsNonEditable()
+        'ucrInputPolarCoordinates.SetRDefault(Chr(34) & "x" & Chr(34))
```
dannyparsons: why is this commented? `x` is the correct default.

dannyparsons:
I put the default back in and that worked well. It's now merged.
@Ivanluv please now implement for all graphics dialogs.

PR #5537, opened 2019-10-03
Predict dialogue#5127

### LINKED ISSUE #5127: Adding to the model dialogue 3 - a predict dialogue
I suggest this is important.  I indicate this through showing the "system" in Genstat.  I think we can do even better!
Here is the dialogue for fitting a model.  This is like our Model > General > Fit or Model > Model

image

Notice that Predict is not enabled.  I now run the model.  This is like our Fit and this enables the Predict button.  So I press Predict to give the following sub-dialogue:

image

In the above dialogue I changed the levels of the fertiliser to 0,1,2,3 and then it gave the following results in the output window:
image

I had asked, so it also saved the (same) results in 2 data frames.  It has a table structure for data frames, so it saved them separately.  Here is one of them:

image

I could also do a graph, and when I click on the options in Predict I get as follows:
image
 This gives the following graph:
image

Note that in Genstat you have access to all the columns in the data frame.  So what happens if you a) Don't use all the variables in the model.  Here is the result ignoring the variate:

image

And here ignoring that factor:
image

These are like the margins in the 2-way table given above.  Those were "Marginal weights", i.e. using the observed frequencies of the 3 varieties.
image

These are equal weights, i.e. weighting each variety equally.

b) There is nothing in Genstat to stop you using predict with variables that are not in the model.  However, then it just gives a warning that you have used terms that are not in the model, and doesn't give you anything.   We can do better, because we could just include the terms that are in the model - from the corresponding object.

I claim this is all very important for us in R-Instat. It sort of "completes" the model component.

In "simple (descriptive) statistics" we use the describe menu to prepare informative tables and graphs.  They are direct summaries of the data.

Then we move to modelling (statistical inference).  We fit models and the "recent" advances are on the range of models that we can fit with a common framework.

Once we have a suitable model we then (should) want to prepare the corresponding tables and graphs using the chosen model.  That's statistics!  And (see above) the predict feature is the way we do this.  So in an improved (more advanced?) version of our statistical problem-solving course we could include this idea!

We have some decisions to make for R-Instat.  I suggest we use the **prediction package** where possible.  Here are some possibilities - not mutually exclusive:
1. We have a separate dialogue for the predictions.  I think this will be useful.  That's easy for us (and not for Genstat), because we always save the model object  So the prediction dialogue can use the saved models.
2. We may sometimes wish to (also) be able to display the predictions at the same time as we fit the model.  That's what Genstat does.  In our case it could be another tab on the Display sub-dialogue.  This could call the same prediction dialogue.
3. We will often want to produce the predictions into a new data frame - see the saving from Genstat above.  Unlike Genstat this will be an ordinary data frame, so the example above would have 12 rows (4 for the levels of Fert, by the 3 varieties.  That means that the predictions are in a single column and the standard errors, etc can be in further columns.
4.  We will usually also want to produce some output in the output window.  There it is probably appropriate to produce tables - as is done by Genstat.
5. I assume we will usually get the Prediction dialogue to produce the new data frame.  An alternative is to use an existing data frame.  We might even choose to have a predict tab on our **File > New Data Frame** dialogue.  So we could then access an existing data frame to get predictions.  This will also be appropriate if (as above) we want to try alternative options in the predictions.  They would then simply become extra columns in an existing data frame.  As with other summaries the prediction data frames would (by default) be linked to the data frame with the data.

The structure of the prediction data frame will need careful discussion and planning.

Ivanluv:
fixes #5127 
adding the show argument check box. 
@dannyparsons  how  are the `iSetCursorBackCharacters`  determined in  the function `AddToReceiverAtCursorPosition`  which is in `ucrReceiver expression`


Ivanluv:
@africanmathsinitiative/developers  can someone review this

rdstern:
I like the look of the dialogue now.
I assume we could use any data from the extremes package to test.  I checked by using the moorings data from the climatic library where I just then used climatic > prepare > extremes and got the maximum for the year.  This produced a variable with one complication that it has missing values.

Then I used the Fit Model > Extremes and fevd, adding the argument na.action=na.omit to get the fitted model.  Then I used the new Use Model dialogue.

a) It looks great now.
b) The dialogue name is given as dlgUseModel in the top left.  Let's make it **_Use Model_**.
c) The show arguments seems to work fine.
d) Remembering the past commands used seems to work fine.
e) I suggest the Clear button be moved to the right-hand side of the dialogue  and the same level as the keys on the general keyboard.  There is space there.
f) The Try button needs to move up and a Save Result checkbox be added below the Try.  It is the equivalent of the Save Model checkbox in the Model dialogue.
g) I first tried the ci (extremes keyboard) which works fine.  I then tried another key and it didn't work.  When it doesn't work the previous result is displayed in the output window.  It should be cleared and not displayed if possible.
h) The Try option always seems to give the information that there is an error.  This seems independent of whether there is one or not.

rdstern:
This now all seems to sort of work and looks really good.  
When @dannyparsons checks I hope he can look at the following "features".

I took an example datatset from the extRemes package (temperatures).  Then used the Model > Model dialogue to fit an extreme value distribution.

Then the Model > Use Model dialogue.

I am back to having the same problem as before - which went away temporarily!  When I try the plot, then the first time the plot appears very quickly and then is gone again.  After this I don't get the plot?

The print and summary seem to work fine.  But when I opt to save, then I no longer get any results in the output window.  And I can't find the object I saved.  I assume it isn't yet being added to the workbook?

Ivanluv:
@rdstern  I've made some changes and I think saving this now works

instat/dlgModelling.vb
```diff
@@ -255,7 +255,7 @@ Public Class dlgModelling
     Private Sub cmdlm_Click(sender As Object, e As EventArgs) Handles cmdlm.Click
         Clear()
         If ucrChkIncludeArguments.Checked Then
-            ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("lm(formula = , data, subset, weights, na.action,method = ""qr"", model = TRUE, x = FALSE, y = FALSE, qr = TRUE, singular.ok = TRUE, contrasts = NULL, offset)", 143)
+            ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("lm(formula = , data, subset, weights, na.action,method = ""qr"", model = TRUE, x = FALSE, y = FALSE, qr = TRUE, singular.ok = TRUE, contrasts = NULL, offset)", 133)
         Else
```
volloholic: Why are there any changes here? This file is not relevant.

Furthermore when we checked this it was correct before at 143!

maxwellfundi: @Ivanluv  could you change this minor issue here?

rdstern:
Now that the segmented keyboard is included in the master it would be good if the issue raised by @volloholic could be resolved as well as the conflicts, so this could be merged.

instat/dlgModelling.vb
maxwellfundi: @volloholic  @rdstern This is already resolved now.

PR #5538, opened 2019-10-03
Added tooltips for Logical and Symbols keyboard

### LINKED ISSUE #5520: Add tooltips to the calculator, logical keyboard
This is quite a big job, but very useful, particularly as the help is taking too long.

I hope it is easy, and that @dannyparsons can confirm it would be useful to start on this soon.  I am not sure of the priority, and it depends on other work.

I am also not sure how it will be translated, because we are close to a new system for translations.  Good if tooltips can be considered for translation, if there are alternative ways they can be given?

**_Simple keyboard:_**
^ power or exponent and can also be given as **.  For example 2^3 = 8
Logical and Symbols
< less than.  For example (2 < 3) is TRUE.  (2 < 2) is FALSE, (2 < 1) is FALSE
<= less than or equals.  For example (2 <= 3) is TRUE,  (2 <=2)  is TRUE, (2 <= 1) is FALSE
> greater than.  For example (2 > 3) is FALSE, (2 > 2) is FALSE, (2 > 1) is TRUE
>= greater than or equals.  For example (2 >= 3) is FALSE, (2 >=2) is TRUE, (2 >= 1) is TRUE
== logical equals.  For example (2 == 3) is FALSE, (2 == 2) is TRUE, (2 == 1) is FALSE
!= not equals (opposite of ==). For example (2 ! 3) is TRUE, (2 != 2) is FALSE and (2 != 1) is TRUE
|  logical OR.  For example  (2 < 3) | (2 < 1) is the same as (TRUE | FALSE) which is TRUE
& logical AND.  For example  (2 < 3) & (2 < 1) is the same as (TRUE & FALSE) which is FALSE
! logical NOT.  For example !(2 < 3) is FALSE, !(2  < 1) is TRUE
: colon is from:to.  For example 3:6 is 3, 4, 5, 6. 6:3 is 6, 5, 4, 3 (It is the seq function with a step-length of 1 or -1.)
%%   modulus operator gives the remainder after integer division.  For example 7 %% 3 is 1, 12 %% 3 is 0
%in% is like the match function and returns a logical vector.  For example (11:15 %in% c(11,13)) gives TRUE, FALSE, TRUE, FALSE, FALSE
match (see also %in%) gives the positions of the matching elements.  For example match(11:15, c(11,13)) gives (1,NA, 2, NA, NA). match(11:15,c(11,13),nomatch=0) gives 1, 0, 2, 0, 0)
 
%/% div operator is for integer division.  For example (7 %/% 3) is 2, (13 %/% 3) is 4

Wycklife:
Fixes #5520
For logical and Symbol keyboard, the tooltips have been added for most of the keys.
@africanmathsinitiative/developers Someone can review this

### DIFF
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -84,7 +84,7 @@ Public Class ucrCalculator
         ttCalculator.SetToolTip(cmdwhen, "when is multiple ifelse, for example case_when(1:5>320,1:5>110) gives NA, 10, 10, 20, 20")
         ttCalculator.SetToolTip(cmdIfelse, "is what it says, for example ifelse((1:5 > 3,20,10) gives 10, 10, 10, 20, 20")
         ttCalculator.SetToolTip(cmdBetween, "between two values, for example between(1:5, 3,4) is FALSE, FALSE, TRUE, TRUE, FALSE")
-        ttCalculator.SetToolTip(cmdMiss, "is.na detects missing values, for example is.na(c(1,3,NA, 5)) gives FALSE, FALSE, TRUE, FALSE")
+        ttCalculator.SetToolTip(cmdIsNa, "is.na detects missing values, for example is.na(c(1,3,NA, 5)) gives FALSE, FALSE, TRUE, FALSE")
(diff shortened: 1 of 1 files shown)


Wycklife:
@rdstern I have added the tooltips for most of the keys except for isTRUE and isFALSE plus a few of them. Just have a look at them if that's what is required. Happy to make any possible change. Again just have check at is.na, I don't understand why it doesn't display the information even after adding them.

PR #5540, opened 2019-10-06
New treemap dialog

dannyparsons:
Implements the `treemap` graph from the `treemapify` package, see details here https://cran.r-project.org/web/packages/treemapify/vignettes/introduction-to-treemapify.html

An interesting test is to use the WorldBank97 data from the Procurement library and use `country` as the identifier. You could also weight by the contract value. This will test the "Primary data" option. Use the "country" sheet to test the "Summary data" option.

- Install the `treemapify` package to test as it's not yet included in R-Instat
- I have excluded the dialog from the project so that there are no vbproj changes. To test, include in the project and add the dialog somewhere in frmMain. Eventually this will be a graph under the describe menu, and on the procurement menu
- The geom only works with summarised data so the dialog has "Primary data" and "Summary data" options at the top. If "Primary data" then the data is summarised in the code to use the geom. In this case, facets are disabled.
- For now I haven't added a "Treemap Options" button to get to layer options, like on most dialog, as this could add bugs due to the summarising in the code. I think its not essential.
- Layers tab is disabled on Plot Options dialog since the geom have not yet been added. This can be enabled once geoms are defined.

### DIFF
--- a/instat/dlgTreemap.vb
+++ b/instat/dlgTreemap.vb
@@ -280,7 +280,7 @@ Public Class dlgTreemap
         TestOkEnabled()
     End Sub
 
-    Private Sub CoreControls_ContentsChanged() Handles ucrReceiverIdentifier.ControlContentsChanged, ucrSaveTreemap.ControlContentsChanged, ucrPnlDataType.ControlContentsChanged
+    Private Sub CoreControls_ContentsChanged() Handles ucrReceiverIdentifier.ControlContentsChanged, ucrSaveTreemap.ControlContentsChanged, ucrPnlDataType.ControlContentsChanged, ucrReceiverWeightBy.ControlContentsChanged
         TestOkEnabled()
     End Sub
 
--- a/instat/dlgTreemap.vb
+++ b/instat/dlgTreemap.vb
@@ -43,11 +43,20 @@ Public Class dlgTreemap
     Private clsXScalecontinuousFunction As New RFunction
     Private clsYScalecontinuousFunction As New RFunction
     Private clsRFacetFunction As New RFunction
+    Private clsCoordPolarFunction As New RFunction
+    Private clsCoordPolarStartOperator As New ROperator
     Private clsThemeFuction As New RFunction
     Private dctThemeFunctions As New Dictionary(Of String, RFunction)
 
+    Private strIdentifier As String = "Identifier"
+    Private strIdentifierAndWeighty As String = "Identifier & Weight"
+    Private strWeight As String = "Weight"
+    Private strNone As String = "None"
+
     Private bReset As Boolean = True
     Private bFirstLoad As Boolean = True
+    Private bRCodeSet As Boolean = False
+    Private bResetSubdialog As Boolean = True
 
     Private Sub dlgTreemap_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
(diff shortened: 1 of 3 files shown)


instat/dlgTreemap.vb
```diff
@@ -0,0 +1,441 @@
...
+'
+' This program is distributed in the hope that it will be useful,
+' but WITHOUT ANY WARRANTY; without even the implied warranty of
+' MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+' GNU General Public License for more details.
+'
+' You should have received a copy of the GNU General Public License 
+' along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+Imports instat
```
maxwellfundi: This is not needed.

instat/dlgTreemap.vb
```diff
@@ -0,0 +1,441 @@
...
+    Private Sub cmdOptions_Click(sender As Object, e As EventArgs) Handles cmdOptions.Click
+        sdgPlots.SetRCode(clsNewOperator:=ucrBase.clsRsyntax.clsBaseOperator, clsNewYScalecontinuousFunction:=clsYScalecontinuousFunction, clsNewXScalecontinuousFunction:=clsXScalecontinuousFunction, clsNewXLabsTitleFunction:=clsXlabFunction, clsNewYLabTitleFunction:=clsYlabFunction, clsNewLabsFunction:=clsLabsFunction, clsNewFacetFunction:=clsRFacetFunction, clsNewCoordPolarFunction:=clsCoordPolarFunction, clsNewCoordPolarStartOperator:=clsCoordPolarStartOperator, clsNewThemeFunction:=clsThemeFuction, dctNewThemeFunctions:=dctThemeFunctions, clsNewGlobalAesFunction:=clsGeomTreemapAesFunction, ucrNewBaseSelector:=ucrSelectorTreemap, bReset:=bResetSubdialog)
+        'this is a temporary fix
+        If rdoPrimary.Checked Then
+            sdgPlots.tbpFacet.Enabled = False
+            sdgPlots.tbpLayers.Enabled = False
+        End If
+        sdgPlots.ShowDialog()
+        sdgPlots.tbpFacet.Enabled = True
+        sdgPlots.tbpLayers.Enabled = True
```
maxwellfundi: i think these two can go into the else of the **rdoPrimary.checked** then the sdgplots.showdialog remains  outside the if statement.

maxwellfundi:
@rdstern  A quick one here- if you could install - `treemapify ` package and have a look at this dialog. @dannyparsons  where on the menu is this supposed to be in?

rdstern:
@maxwell ok will do

rdstern:
@maxwellfundi could you quickly add the dialogue to the project. If we remain cautious then let's start by calling it Treemap and putting it at the bottom of Procurement > Describe > Categorical.

I am happy to test then or for you to merge first - it can't do much harm sitting there!

maxwellfundi:
@rdstern  Okay am adding it there. Thanks.

maxwellfundi:
@dannyparsons  there seems not to have had vbproj  file on this dialog?

rdstern:
@dannyparsons I have followed your suggestions and - so far so good.
image

On the individual data it seems to work ok, with and without a filter.  The example above is for just the smallest contracts - using the forcats::lump faility to set up the data for the filter!

Aha, my first problem:
image
Can the weights be done better than this.  These are the summary data so I could round myself - but that isn't possible with the raw data.

Then I tried the different colours - seems ok

I wasn't sure what to expect from Fill, so I tried with a variable and got an error.  (I was on summary)
image

PR #5563, opened 2019-10-28
Improvements to mapping#5368

### LINKED ISSUE #5368: Improvements to the mapping dialogue
This is now working well and is a potnetial "hit" for us.

a) On the lhs, the sf options is disabled.  @maxwellfundi is investigating.
b) Is the right-hand side correctly called "Points".  It also includes labels.  Could we call it **Data** instead?
c) Then the 2 key fields are the long and lat.  I suggest they come first.
d) Then perhaps there could be 2 group boxes, with one for **Label** and the other for **Points** .
If Label only has one field, then that's ok, but are there options, e.g. geom Label or geom text, and also the size.  In that case there is a gropu box.
For the Points group can we also add size?
e) Then I like the simple Facets. 
f) I am not so keen on the (Optional) label.  Do we need them.

Ivanluv:
fixes #5368
@dannyparsons would you kindly look at this code.


instat/dlgClimaticStationMaps.vb
```diff
@@ -344,11 +346,10 @@ Public Class dlgClimaticStationMaps
...
         clsSetGeometry.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_geometry")
         clsSetGeometry.AddParameter("data", ucrSelectorOutline.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
-        GeometryOutput = frmMain.clsRLink.RunInternalScriptGetOutput(clsSetGeometry.ToScript(), bSilent:=True)
-
+        GeometryOutput = frmMain.clsRLink.RunInternalScriptGetValue(clsSetGeometry.ToScript(), bSilent:=True)
+        Dim str As String = GeometryOutput.AsCharacter(0)
         If GeometryOutput IsNot Nothing AndAlso Not GeometryOutput.Type = Internals.SymbolicExpressionType.Null Then
-            ucrReceiverGeometry.txtReceiverSingle.Enabled = True
-            ucrReceiverGeometry.SetText("sd")
+            ucrReceiverGeometry.Add(str, ucrSelectorOutline.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
```
maxwellfundi: @dannyparsons  we have this sort of working with @Ivanluv. One issue i am unable to immediately see why its happening is- the `str` above which is geometry, is not being added to the geometry receiver even after using the correct method- .`Add`. I have gone through the add method and everything appears Okay. @Patowhiz could you help look at why thats happening?

rdstern:
I like the improvements.  The only problem I had was with facets.
I had a simple level 3 map for one county in Kenya - I think called Taita Taveta.
[Taita Taveta shape.zip](https://github.com/africanmathsinitiative/R-Instat/files/3873532/Taita.Taveta.shape.zip)

I read in this file and used it for a level 3 map.  Then I used the points option with the same file and used Name3 as the points and Name2 as the facets.  That works fine.  Then I went to Plot Options > Themes and got rid of the legend.  That worked, but the facets also disappeared, theough they were still there on the main dialogue.  Then I looked at the Plot Options > Facets and found they were not set there.  
When I added Name2 into the facets field again, then the facets reappeared.  Good to check what is happening, both when other plot options change, and also whether setting facets on the main diualogue should also set it automatically on the sub-dialogue - so they amount to the same thing, with the main dialogue just being more convenient.  Perhaps this is something where @maxwellfundi   can help?  

Then I would really like to get the new dialogue merged as soon as possible.  These aee nice improvements.

Patowhiz:
@Ivanluv this branch has conflicts. I want to test and possibly help. Can you resolve the changes?

maxwellfundi:
let me see if i can be able to resolve them @Patowhiz

maxwellfundi:
@Ivanluv @Patowhiz  I have decided to keep what was in the master on the sdgplots dialog. it was a single line that had conflicts.

maxwellfundi:
@rdstern  I have already fixed the build errors, the geometry is not yet being added automatically yet but we should be able get this fixed with Danny.  You can recheck this

rdstern:
Great - will do.

instat/dlgClimaticStationMaps.vb
```diff
@@ -232,12 +254,17 @@ Public Class dlgClimaticStationMaps
...
     Private Sub cmdMapOptions_Click(sender As Object, e As EventArgs) Handles cmdMapOptions.Click
         sdgMapOption.SetRCode(clsBaseOperator:=clsGGplotOperator, clsXlim:=clsXlimFunction, clsylim:=clsYlimFunction, bReset:=bResetSubdialog)
         sdgMapOption.ShowDialog()
         bResetSubdialog = False
     End Sub
 
+    Private Sub ucrChkAddPoints_Load(sender As Object, e As EventArgs)
+
+    End Sub
+
```
volloholic: This looks like it's been added by mistake?

instat/dlgClimaticStationMaps.vb
```diff
@@ -288,11 +315,56 @@ Public Class dlgClimaticStationMaps
...
+        clsGetDataFrame.AddParameter(clsParam)
+        clsGetDataFrame.SetAssignTo(ucrSelectorOutline.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
+        clsGetDataFrame.ToScript(strScript)
+        frmMain.clsRLink.RunInternalScript(strScript)
+
+        clsSetGeometry.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_geometry")
+        clsSetGeometry.AddParameter("data", ucrSelectorOutline.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
+        GeometryOutput = frmMain.clsRLink.RunInternalScriptGetValue(clsSetGeometry.ToScript(), bSilent:=True)
+
+        frmMain.clsRLink.RunInternalScript(clsRemoveFunc.ToScript())
```
volloholic: This looks ok but it would be good to order the rcommands more logically!

PR #5565, opened 2019-10-29
Improvements and tooltips for the calculator transform keyboard

Wycklife:
Fixes ##5554

### DIFF
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -69,6 +69,7 @@ Public Class ucrCalculator
         ttCalculator.SetToolTip(cmdLag, "Shift a variable down. For example lag(1:5) = (NA,1,2,3,4); lag(1:5,3) = (NA,NA,NA, 1,2)")
         ttCalculator.SetToolTip(cmdLead, "Shift a variable up. For example lead(1:5) = (2,3,4,5,NA); lead(1:5;3) = (4,5, NA,NA,NA)")
         ttCalculator.SetToolTip(cmdDiff, "difference between successive elements. For example diff(c(1,4,3,7)) = (NA 3,-1,4)")
+        ttCalculator.SetToolTip(cmdpmax, " maximum of a set of variables. For examples pmax(c(1,3,5),c(6,4,2)) = (6,4,5)")
 
     End Sub
 
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -67,6 +67,7 @@ Public Class ucrCalculator
         ttCalculator.SetToolTip(cmdRound, "round(x) to round to whole numbers, round(x,2) to round to 2 decimal places, round(x,-2) to round to the nearest 100")
         ttCalculator.SetToolTip(cmdSiginf, "signif(x,3) to round to 3 significant figures")
         ttCalculator.SetToolTip(cmdLag, "Shift a variable down. For example lag(1:5) = (NA,1,2,3,4); lag(1:5,3) = (NA,NA,NA, 1,2)")
+        ttCalculator.SetToolTip(cmdLead, "Shift a variable up. For example lead(1:5) = (2,3,4,5,NA); lead(1:5;3) = (4,5, NA,NA,NA)")
 
     End Sub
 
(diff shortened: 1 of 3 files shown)


rdstern:
@Wycklife this is looking excellent.  Just some minor suggestions on the keyboard.
1.  I have made another (more general) issue on the length of time a tooltip remains visible.  That needs a bit of discussion, before being implemented.
2. The naapprox tooltip starts inear.  Add the l for linear.
3. In the naapprox code include the argument na.rm=FALSE.  Otherwise it gives an error in examples where there are initial missing values (e.g. the dodoma data when used on the temperatures. 
4. I wonder now if you tried the code you included.  It is worth doing with the dodoma temperature variables.  Anyway when you tick on the show arguments the 5 functions for the keys movmax to movmin each have an extra bracket at the end.  There should be just 2 and there are 3.
5. The same functions have a complicated code for the fill argument - which I don't understand and it doesn't run like that.  Change them to fill = NA as in the tooltip.
6. On these same keys, when the arguments is not ticked you still need a value of the moving average to run.  So include ,3,fill=NA) then.
6. The function for minimums doesn't exist - I don't know why.  So that is why we have the movapply function there.  But include FUN="min" in the call so it produces minima.
7.  The key nafill gives a function na.fill0(). I think you need to delete the zero.

Wycklife:
@rdstern I have implemented the changes above

rdstern:
See above.  Some have been done, but some still remain.  Overall it is looking good and is very powerful now.

maxwellfundi:
@Wycklife  can you please address all the issues remaining here.

Wycklife:
@maxwellfundi That's what am doing now.

maxwellfundi:
great, sounds good.
Regards
Maxwell Fundi
|| Phone :+254728508241||

Wycklife:
@rdstern Could you please check whether the updated changes are okay?

rdstern:
Close now:
6. On these same keys, (movsum, max, mean, median) when the arguments is not ticked you still need a value of the moving average to run. So include ,3,fill=NA) then. Check the code. You need the comma to be included, and perhaps also include it as ,k=3. Then the active point should be just in front, so when you click ona variable it adds into the right place.  **_Currently to commands looks ok, but when you add the variable it should automatically go just after the bracket and before the first comma.  It doesn't yet.  This is for the 5 keys_** .
7. The function for minimums doesn't exist - I don't know why. So that is why we have the movapply function there. But include FUN="min" in the call so it produces minima. _**Still not yet done**_

instat/ucrCalculator.vb
```diff
@@ -66,6 +66,31 @@ Public Class ucrCalculator
         bControlsInitialised = True
         ttCalculator.SetToolTip(cmdRound, "round(x) to round to whole numbers, round(x,2) to round to 2 decimal places, round(x,-2) to round to the nearest 100")
         ttCalculator.SetToolTip(cmdSiginf, "signif(x,3) to round to 3 significant figures")
+        ttCalculator.SetToolTip(cmdLag, "Shift a variable down. For example lag(1:5) = (NA,1,2,3,4); lag(1:5,3) = (NA,NA,NA, 1,2)")
+        ttCalculator.SetToolTip(cmdLead, "Shift a variable up. For example lead(1:5) = (2,3,4,5,NA); lead(1:5;3) = (4,5, NA,NA,NA)")
+        ttCalculator.SetToolTip(cmdDiff, "difference between successive elements. For example diff(c(1,4,3,7)) = (NA 3,-1,4)")
```
Patowhiz: @dannyparsonswe are discussing about this dialog with @Wycklife, why did you prefer the setting of the tooltip to be made through code instead of using the designer? This could easily be done through visual studio designer reducing the number lines in this subroutine and thus improving clarity. Or I'm I missing something?

tooltip

Wycklife:
@rdstern This is now okay

Wycklife:
@rdstern It's done

maxwellfundi:
@Wycklife  There are a couple of conflicts here some of which are in the designer, and .resx which are hard to fix from my end. If you can update your master, create a new branch and transfer these changes.  that will be great.

PR #5566, opened 2019-10-30
Climdex#5367

### LINKED ISSUE #5367: Two small issues with Climatic > Prepare > climdex
There is one big issue that the climdex dialogue does not include the Station field.  It is currently designed for just a single station.  I am curious to know what size of job it is to extend the code for multiple stations?  I assume that is a @dannyparsons question.

There are 2 smaller issues so far.
I am testing with Dodoma that is just a single station and Ghana, where there are 2 stations.  I am using the library data for both.
Initially I am using **annual indices** and just the rainfall ones.  In the rainfall tab I ticked everything, so I produce 12 columns
The two problems I found are are as follows:
1) When I use the indices and it makes a new data frame (a opposed to adding to an exiting annual one) it makes the year column into a **factor**.  I need the year as numeric for plotting.  It is the linking column, so not trivial to change it.
So, could it please leave the Year column as the same type as in the annual data - usually numeric. 
2) For the Ghana data it insists on making a new data frame for the annual summaries.  This is even though I have either filtered to a single station, (I tried with Saltpond) or have made a subset - which I did with Tamale.  This may be a more general issue, that the new data frame is called Ghana by station_year (or Tamale_by_Station_Year), even though there is a single station.  The climdex dialogue doesn't have a Station field, so the result is a data frame which is just Tamale_by_Station.
  
  This result means that Dodoma is nice and does exactly what I would like.  I can get some other summaries and then the summaries from climdex add to the existing data frame.  (And as climdex isn't writing the year column it isn't a factor there.  For Ghana this doesn't happen.

a) My suggested (temporary) solution is to add a **Station field** to the dialogue which is visible, but disabled.  This would be blank for Dodoma, but completed for the Ghana examples - even when the Station has just a single level. 

b) Also leave the year column, i.e. don't have it automatically as a factor, when making a new data frame.

c) And if you agree with these steps, then (in preparation) could a Station field also be added - in the same way - to the SPI dialogue.  I will be coming to that one next!

Ivanluv:
fixes #5367 
@rdstern  does this look okay


maxwellfundi:
@rdstern  the abbreviations here seem to have been set properly. Can you please counter check?

dannyparsons:
@Ivanluv avoid using "fixes" when it doesn't fix the whole issue, otherwise the issue will be closed without everything being solved. The main issue was including by station in climdex indices, which isn't going to be solved here yet.

Also, is the count of number of indices included here?

dannyparsons:

Ivanluv:
Yes

PR #5570, opened 2019-11-07
Fit model#5561

### LINKED ISSUE #5561: Add a Try control to the Model > General > Fit dialogue
In this dialogue the proposed model can be typed.  Mistakes are therefore possible.  So could we have the same Try control that we have on (for example) the Model > Modelling dialogue.

At the same time a small change to the keyboard would be useful.  Instead of 0 there could there be I( ), with the focus then inside the brackets.  That enables an ordinary formula to be included in the model formula.

Ivanluv:
fixes #5561

### DIFF
--- a/instat/dlgFitModel.vb
+++ b/instat/dlgFitModel.vb
@@ -251,7 +251,7 @@ Public Class dlgFitModel
     End Sub
 
     Private Sub cmdZero_Click(sender As Object, e As EventArgs) Handles cmdZero.Click
-        ucrReceiverExpressionFitModel.AddToReceiverAtCursorPosition("l()", 1)
+        ucrReceiverExpressionFitModel.AddToReceiverAtCursorPosition("I()", 1)
     End Sub
 
     Private Sub cmdClear_Click(sender As Object, e As EventArgs) Handles cmdClear.Click
--- a/instat/dlgFitModel.vb
+++ b/instat/dlgFitModel.vb
@@ -15,7 +15,10 @@
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 Imports instat.Translations
+Imports RDotNet
 Public Class dlgFitModel
+    Private clsAttach As New RFunction
+    Private clsDetach As New RFunction
     Public bFirstLoad As Boolean = True
     Private bReset As Boolean = True
     Public bRCodeSet As Boolean = False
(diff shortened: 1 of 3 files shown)


instat/dlgFitModel.vb
```diff
@@ -403,4 +414,61 @@ Public Class dlgFitModel
     Private Sub ucrBase_ClickOk(sender As Object, e As EventArgs) Handles ucrBase.ClickOk
         ucrReceiverExpressionFitModel.AddtoCombobox(ucrReceiverExpressionFitModel.GetText)
     End Sub
+
+    Private Sub TryScript()
```
Patowhiz: @dannyparsons this is the `TryScript` of the Fit Model dialog. @Ivanluv and I have implemented it in a different way compared to other dialogs like hypothesis test calculator dialog.

Instead of using the dialog's declared before,base and after scripts code (in this case `clsAttach, clsLMOrGLM, clsAnnovaFunction ,clsSummaryFunction, clsDetach`), we are just accessing them from base buttons controls and running all of them in a loop line by line, each time checking for the length of the returned vector. 

We preferred this over the other one because, we are relying on the `RSyntax `object to create the scripts for us (Which it's what happens when you click **To Scrip**t button or **Ok** button). We have also used this concept because it's how the Script function of `ucrButtons `control is implemented, so we thought doing a similar implementation is probably more ideal.

This works well, I'm trying to think of a scenario when it may not work, an ideas? And do you think this is better than trying to construct the scripts manually like how it has been done in the other dialogs? Is there a reason why we should construct the scripts _manually_.

Any feedback and the reason behind why this may not be a good implementation will help, and we recognise that we can manually create the before,base and after scripts from the declared functions.

Ivanluv:
@rdstern  could you have alook at this again

rdstern:
Great.  It all seems to work now.  Perhaps @dannyparsons could do a final check and merge?

instat/dlgFitModel.vb
maxwellfundi: @dannyparsons ??

volloholic: @Patowhiz This is a really good question. The main reason to construct it manually is that things like anova and summary are not part of the model fitting they use the model and then are part of extra's that can be presented. This could also include extra graphs ect, some of which may give errors. So if we're really just wanting to try the model then these shouldn't be done, they can also be much slower. 
However, what you have implemented is sensible and so it is probably a useful discussion to decide what the try should be applying to, just the model or the whole process. I suggest we merge the current implementation but keep an issue open for this discussion.

maxwellfundi:
@volloholic  great, i am merging this into the master now.

PR #5574, opened 2019-11-13
One & two variable summarise improvements

dannyparsons:
- dialog design modified, checkbox for customise replaced with radio buttons
- improved display of names of summaries in output
- removed scientific notation of summaries
- corrected summaries for Dates
- corrected error for invalid summaries on some data type
- ordered summaries in output by size and groupings

**Edit**
Added improvements to two variable summarise dialog here since the improved summary methods were needed
- added a box on the dialog showing the summary that will be done based on the variables selected
- allowed for broad categorical/numeric types (implemented in multiple receiver) to be chosen in the first receiver (e.g. factor and character are allowed together, or numeric and date)
- created a correlation function that ensures columns are converted to numeric first (e.g. date and logical)

### DIFF
--- a/instat/dlgDescribeTwoVariable.Designer.vb
+++ b/instat/dlgDescribeTwoVariable.Designer.vb
@@ -40,15 +40,22 @@ Partial Class dlgDescribeTwoVariable
     Private Sub InitializeComponent()
         Dim resources As System.ComponentModel.ComponentResourceManager = New System.ComponentModel.ComponentResourceManager(GetType(dlgDescribeTwoVariable))
         Me.cmdSummaries = New System.Windows.Forms.Button()
-        Me.cmdDisplayOptions = New System.Windows.Forms.Button()
         Me.lblFirstVariable = New System.Windows.Forms.Label()
         Me.lbSecondVariable = New System.Windows.Forms.Label()
+        Me.grpOptions = New System.Windows.Forms.GroupBox()
+        Me.lblSummary = New System.Windows.Forms.Label()
+        Me.lblFirstType = New System.Windows.Forms.Label()
+        Me.lblBy = New System.Windows.Forms.Label()
+        Me.lblSecondType = New System.Windows.Forms.Label()
+        Me.lblSummaryName = New System.Windows.Forms.Label()
+        Me.grpSummaries = New System.Windows.Forms.GroupBox()
+        Me.ucrChkOmitMissing = New instat.ucrCheck()
         Me.ucrReceiverSecondVar = New instat.ucrReceiverSingle()
         Me.ucrReceiverFirstVars = New instat.ucrReceiverMultiple()
         Me.ucrSelectorDescribeTwoVar = New instat.ucrSelectorByDataFrameAddRemove()
         Me.ucrBase = New instat.ucrButtons()
-        Me.ucrChkSaveResult = New instat.ucrCheck()
-        Me.ucrChkOmitMissing = New instat.ucrCheck()
+        Me.grpOptions.SuspendLayout()
+        Me.grpSummaries.SuspendLayout()
         Me.SuspendLayout()
         '
         'cmdSummaries
@@ -57,13 +64,6 @@ Partial Class dlgDescribeTwoVariable
         Me.cmdSummaries.Name = "cmdSummaries"
         Me.cmdSummaries.Tag = "Summaries"
         Me.cmdSummaries.UseVisualStyleBackColor = True
-        '
-        'cmdDisplayOptions
-        '
-        resources.ApplyResources(Me.cmdDisplayOptions, "cmdDisplayOptions")
-        Me.cmdDisplayOptions.Name = "cmdDisplayOptions"
-        Me.cmdDisplayOptions.Tag = "Display_Options"
-        Me.cmdDisplayOptions.UseVisualStyleBackColor = True
         '
         'lblFirstVariable
         '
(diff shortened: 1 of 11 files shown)


rdstern:
On the one variable I really like the fact that the minimum to maximum are now displayed in a sensible order.
With a date variable it now works except it gives NA for the quartiles.  If you use methods 1 to 3 (default is 7 for numeric variables, and that should stay) - guess method 1 is ok, then they should be possible.
I also tried a summary on the More tab and chose again the percentile and took the default of 0.25.  Now it gives missing for both dates and numeric?
Then I tried more summaries.  I tried the sd and var and range.  They seem to be given as dates, when they are not, see below:
image
Also, in the above, if you can fix the quantile could the label add the number - either 0.25 or 25%?

rdstern:
I like the two variable improvements a lot.  I do like the way it explains what it is doing.  I know this is still work in progress.
a) In numeric by factor could there be a choice of order.  Currently it is by summary.  The alternative would be by level of the factor.
b) In the frequency tables could there be options of margins or not?  And frequencies or any of the 3 percentages?

rdstern:
**_For numeric by numeric_** you give correlations, which is fine.  But after your work on ANOVA would you also consider other analyses, and particularly providing an ANOVA table?  Could also be a regression line as another obvious summary.  Perhaps there could be these options, or a default (correlations) and an options button?
**_For numeric by factor_** you permit all the options which is great.  But could there also be a default as there is, in the one variable?  Effectively this is a summary by a factor.  There are many packages that allow this, but perhaps better could be to use the standard summary function wrapped into a by?  Might this also show a neat bit of code instead of an ordinary loop?

rdstern:
I looked again at the one variable and really like the improvements now in the dates particularly:  Here is an example and I suggest there is more here (except for factor variable) than the default.
image

a) I wonder about the number of significant figures?
b) It would be nice to do a bit more for factor variables.  The only easy improvement I can see is that we could interpret the range as being the number of levels - so we could give that.
c) The quantile function is still to be done - or disabled?  What I asked for was 0.25.

rdstern:
Sorry you can 't stop now.  This is potentially better for teaching than any of the R functions!
a) Please check 2 way with the second being logical.  I tried with a Wakefield derived data set and it complained.  I think there might have to be an as.factor there.
b) If I use the one-variable default for a logical it treats it like a factor and gives the label and the frequency.
So could we do more for factors in the following 2 ways:
1) Interpret Min and Max as the levels with the smallest and largest frequency.  And give their labels.  And (as suggested above) interpret range as the number of levels.
2) For logical do this as well, but continue with the other summaries as you have now.  That shows clearly that it is halfway between a factor and numeric.
3) On the 2-variable logical is  interpreted only as categorical.  Could the categories for logical  become a drop down, so you can choose numeric or categorical!
4) This could also apply to any 2-level factor.
5) When the second variable is a factor (or logical) could the categorical again be a drop down with the options **_categorical_** and **_by variable_**! That would also apply to a logical.
5)  When the second level is a factor you could also have a third option, labelled **_as.numeric (take care!)_**
This treats the factor simply as numeric, which is sometimes sensible - particularly for factors with 2 levels or for ordered factors, like Likert scales.
6) With categorical by categorical the by variable for the second one wiould then give a series of one-way tables, rather than a 2-way table.  With numeric by categorical it simply changes the order of the summaries, so you have all the summaries for the first level before moving on to the second level, etc.

Currently Numeric by categorical doesn't seem to work when the Categorical is Logical.

So I changed Death from Logical to Factor and returned to the dialogue.  But I then got an error, because I had the logical (with all the other numeric variables) as a Y as well as the X.  I got:
image

It then correctly loaded the dialogue having taken Death out of the Y.  I suggest in such an instance it does the same action without the message.

Currently, when the X is logical it gives an error as follows:
image

The dialogue is as follows:
image
 The problem is that the variable Dummy is both an X and a Y.  It then needs to be treated as numeric as Y and categorical as X!

rdstern:
I used the File > New Data Frame, Command and generated "The works" from wakefield.
Then tried your One Variable summary on it.
It failed because it said it couldn't make zip into a numeric.  So I got rid of zip and it worked, but gave very curious results.  I am not convinced by you wanting to change character into numeric - into factor I can understand.

rdstern:
@maxwellfundi could you have a quick look at this and then hopefully merge.  I had a small problem above and suggest @dannyparsons may still want to make more changes.  But it is generally working well and is already much better than what is there now.  So merging will certainly be an improvement.

maxwellfundi:
@rdstern Will do. If having it working the way it is now, as an improvement then I will get it into the master.

dannyparsons:
> I used the File > New Data Frame, Command and generated "The works" from wakefield.
> Then tried your One Variable summary on it.
> It failed because it said it couldn't make zip into a numeric. So I got rid of zip and it worked, but gave very curious results. I am not convinced by you wanting to change character into numeric - into factor I can understand.

I tried this but couldn't get the same error. I got the works dataset from File > New Data Frame and then did a one variable summarise of all variables and there was no error.

What summaries did you try and what were the strange results?

There shouldn't be any coercing of characters to numeric - the error is probably in trying to convert the summary column to numeric, but I wasn't able to replicate.
Note that min and max are defined on character vectors already in R so that should return a value. Values are sorted sorted lexicographically

rdstern:
I still get this error message:
image

I started from scratch, used File > New and loaded the works.
Then Describe > One Variable Summary with all variables.  Customised and quite a number of summaries.  Here is the command:
image

dannyparsons:
The culprit here is `sd`. For some reason it coerces character columns to numeric and therefore can return a numeric value (if you have numbers as characters like Zip). 
Try: `sd(c("1", "2", "3", "4"))` - very strange functionality and seems dangerous.
Since we have our wrapper summary functions we can adjust this ourselves to fix it.

PR #5581, opened 2019-11-18
New Data Frame

### LINKED ISSUE #5579: Adding the Random button in File > New Data Frame
One cosmetic improvement is the try feature.  It works fine.  
a) After using it, it sort of remains until used again.  Could it clear when there is a change in the entry.
b) When ok could it add the green colour for consistency with the other dialogues with Try.

Please could we activate the Random option in the F**_ile > New Data Frame_** dialogue.  It can be just the same as the current Command button, but move over the 2 Wakefield examples and then I would like to add some more Wakefield, (plus some examples not using Wakefield.  But the Wakefield ones are potentially exciting.
First move over the third example: data.frame(x = 1:30, y = rnorm(30, mean = 100, sd = 15), z = runif(30, min = 10, max = 30)) into the new place.
Add: data.frame(data = replicate(20,rbinom(n=25, p=0.4,size=1)))
Add
data.frame(data = matrix(data=c(rbinom(7300,p=0.4,size=1)*rexp(7300,0.1)), nrow = 365, ncol = 20))
Add 
data.frame(replicate(10,(sample(c(TRUE,FALSE),size=50,replace=TRUE,prob=c(0.3,0.7)))))

In addition to the 2 for Wakefield that are there, add:
a) wakefield::r_data_theme(n = 200, data_theme = "survey") 
b) wakefield::r_data_theme(n=500, data_theme = "survey2") %>% r_na(prob=0.1)
c) wakefield::r_data_frame(n = 50,id, r_dummy(color),r_series(likert,3),grade,grade,grade)

Finally, I would sometimes like to put extra commands in front of the main command.  If they are on different lines it seems to run ok, though the Try button says otherwise.  An important example is set.seed(995) say, so you can have reproducible examples. I would also like to have:
n=25;p=0.1 say, which would be better programming.

If I put them on the same line it says it can't run, because it expects a data frame.

Patowhiz:
Fixes #5579


rdstern:
Great.  Next "problem".
I am testing adding a new package (segmented) to R-Instat - you don't need the package, but one example there gives the following code:
set.seed(12)
xx<-1:100
zz<-runif(100)
yy<-2+1.5*pmax(xx-35,0)-1.5*pmax(xx-70,0)+15*pmax(zz-.5,0)+rnorm(100,0,2)
data.frame(x=xx,y=yy,z=zz)

When I run this (File > New Data Frame  - Command tab) it works fine.  But when I press the Try it says there is an error.  I wonder why?

The code in R-Instat is:
# Code generated by the dialog, New Data Frame
set.seed(12)
xx<-1:100
zz<-runif(100)
yy<-2+1.5*pmax(xx-35,0)-1.5*pmax(xx-70,0)+15*pmax(zz-.5,0)+rnorm(100,0,2)
data1 <- data.frame(x=xx,y=yy,z=zz)
data_book$import_data(data_tables=list(data1=data1))

rm(data1)

Patowhiz:
@rdstern I wonder whether we are using the same version. When I run the above code I get an error. Even in R studio.

dataframe

rstudio

rdstern:
Sorry, yes there was a typo there - should be * in front of the pmax:
set.seed(12)
xx<-1:100
zz<-runif(100)
yy<-2+1.5*pmax(xx-35,0)-1.5*pmax(xx-70,0)+15*pmax(zz-.5,0)+rnorm(100,0,2)
dati<-data.frame(x=xx,y=yy,z=zz)

My main point is that this runs ok, but the Try button says otherwise.  It may be - like with our dialogue for merging that you could check first if if it ok as as set of R-commands.  Then perhaps if so, then will it produce a data frame.

Patowhiz:
That code still throws an error. Though I get your point. I need to change the implementation of the try, I will have to break the lines and run them each at a time.

rstudio

rdstern:
I see what has happened.  My code was ok, but the use of the star for times has turned into italic in that line in github. 
I am sending my draft help here and the Try is not so easy to use. 
[Examples for File _ New Help.docx](https://github.com/africanmathsinitiative/R-Instat/files/3862471/Examples.for.File._.New.Help.docx)

a) In the example for the **_Construct_** tab I find it a bit awkward to use.  Please try an example or two.

Then I also found it said that the commands in the attached file wouldn't work.  Eventually it relented.
I would also like to have (usually initial) lines there where I complete the expression part, but don't give a variable.  I have assumed that isn't possible
For example I could have a line with:
# This is a comment
or a<-10
or a <- 10 ; n <- 50
etc

Ideally at the end - for good practice I might have a last command
rm(a,n)
But I would be happy with commands just at the beginning lines if this is easy to organise.  If not, then those examples could always use the Command tab.

Then (in the Command and Random tabs) you would, as now, be able to have these commands.  I think there it seems all to work ok - it is just the checking in Try that needs to be tidied.

And Random is the same as Command, from your point of view - it is just the examples that are different.

Thanks

Roger

rdstern:
Comments above updated a bit to clarify!

Patowhiz:
The construct example are clear to me. You have given the example in terms of how it's written in R and how it can be replicated using the grid.

It's clear to me that I need to enhance the try code. I'm doing that.

I'm struggling to understand this:

> I would also like to have (usually initial) lines there where I complete the expression part, but don't give a variable. I have assumed that isn't possible
> For example I could have a line with:

And about your comment, is it in regards to the command option or the construct option?

Patowhiz:
I have fixed the try issue. Commands should now be able to span multiple lines

rdstern:
@Patowhiz I am checking again - and looks even better - exciting!
a) One command in random isn't random and should go back into the command examples:  This is:
`data.frame(block = gl(4, 3), treat = c("C", "A", "B", "B", "C", "A", "A", "B", "C", "A", "C", "B"), yield = c(74, 68,  50, 62, 68, 57, 70, 56, 83, 67, 67, 59))`
It should swap with on that has not moved over and is random, i.e. 
`data.frame(replicate(10,(sample(c(TRUE,FALSE),size=50,replace=TRUE,prob=c(0.3,0.7)))))`

b) Ideally, in the random set of examples, could you swap the order, so the Wakefield ones come after the others.  That's because each of the others illustrates a different idea - so nice to see them all without scrolling.

c)  The first example with the command button is:
data.frame()
I tried it, just as it is.  It works, but I can't then do anything with an empty data frame.  Unless there is something that can be done, then could you replace it by:
`set.seed(101) `
d) Of course the command set.seed(101) doesn't run on its own - and gives an error.  But the try doesn't pick that up.  I am not sure what it does pick up now, see below:

image

Patowhiz:
@rdstern you want the first command example to be just set.seed(101) ? Won't that be misleading? It doesn't produce any data frame.

I have implemented a better try. It will give an okay command as long as one of the complete commands in textbox produce a data frame. This is the best I can check for, since I can't find any other way of explicitly evaluating how the overall script produce results. To improve on the try, I have had to try running each command line by line, it's what happens when you click Ok in the dialog. I'll consult with Danny once he is back.  

@dannyparsons is the away of explicitly checking for object of  type `data.frame`. The sample codes in R-Instant are only checking for nothing and length of `CharacterVector`

rdstern:
Sounds good.  
On the checking, I don't have a problem if the checking is incomplete.  We have that with the checking for the merging dialogue.  If so, then perhaps the text could read "Commands could run Ok" - notice "Commands" in plural.  And (as in Merge) it become gray rather than green.  So, green is when we are pretty sure the command(s) will run, and gray is when we have not found a problem.  This dialogue seems a good case for gray.  The checks should pick out obvious errors like my xxx (not a command at all) or mismatched brackets.  Those will be the most common problems.

On your point on the set.seed(101) I realise we can't use it by itself.  It is very useful for people to consider before generating random samples, because it makes them repeatable.  I wonder if we add a comment?  so for example 
`set.seed(101) #Generate a repeatable sequence.  Add the data.frame commands `

Patowhiz:
> On your point on the set.seed(101) I realise we can't use it by itself. It is very useful for people to consider before generating random samples, because it makes them repeatable. I wonder if we add a comment? so for example
> `set.seed(101)` #Generate a repeatable sequence. Add the data.frame command

Yes I can see the point why set.seed command is important. I'm just trying to be critical on the general purpose of the textbox and the dialog. We want the user to understand that, the commands given should result to a dataframe. Not a column or anything else. By having the default statement as something which if I click on okay gives me an error, won't that contradict me as a user? All the examples that we have by default produce a dataframe even without tweaking or adding anything. My opinion is we maintain this consistency, then in the tutorial we could mention the extra command that enhances randomisation and how it could be used in conjunction with the example.

Or if it's too important, then we can have the `set.seed(101)` followed by a command that produces a dataframe.

Initially we had `data.frame()` as the default because we wanted the user to have a starting point. The he/she could type in the expressions that will produce the rows of the dataframe. This was mentioned in the previous issue.

I think I have now nailed the try functionality. Could you try and break it. how click Ok will execute the statements is how the try will. If it still has an issue I will just adopt the colour codes you have suggested.

rdstern:
Ideally there could be lines with no lhs, i.e. it just runs those commands without making a variable in the data.frame. For example comment lines or the set.seed(999) command  - of course!

rdstern:
It looks very good indeed now.
I tried with obvious errors, etc
The only things not allowed seem to be:
a) Multiple commands per lines.  So I can do set.seed(111) but not `set.seed(111) # this is fun`
b) `a<- 30; data.frame(x = 1:a, y = rnorm(a, mean = 100, sd = 15), z = runif(a, min = 10, max = 30)) says it is an error.  It runs fine, but just gives me the 30 in a data frame.

Just being picky here.  Very happy with what you have.
Can the lines in the examples have tooltips associated with them?

`

Patowhiz:
> The only things not allowed seem to be:
> a) Multiple commands per lines. So I can do set.seed(111) but not set.seed(111) # this is fun
> b) `a<- 30; data.frame(x = 1:a, y = rnorm(a, mean = 100, sd = 15), z = runif(a, min = 10, max = 30)) says it is an error. It runs fine, but just gives me the 30 in a data frame.

Yes, this is because of how R-Instant is structured to run commands. I think the script window can't do multiple commands on the same line too. So that can be subject to discussion in future when we want fully enhance running of commands directly in R-Instant.

> Can the lines in the examples have tooltips associated with them?

Yes the command examples can have tooltips

rdstern:
I now get build errors.  I assume you are in the midst of the final changes?

rdstern:
@Patowhiz the one item remaining - to me anyway - is the possibility of some examples with the first (i.e. the Construct) option selected.  I cannot judge how much work that might be.  If that is on your list for the next few days, then great.  If not, then I am very happy with the improvements so far and perhaps they could be merged first.  Then the pressure is off.

There I would be happy with a single example, e.g. from the Command examples
The:
data.frame(block = gl(4, 3), treat = c("C", "A", "B", "B", "C", "A", "A", "B", "C", "A", "C", "B"), yield = c(74, 68,  50, 62, 68, 57, 70, 56, 83, 67, 67, 59)) becomes
block <- gl(4, 3)
treat <- c("C", "A", "B", "B", "C", "A", "A", "B", "C", "A", "C", "B")
yield <- c(74, 68,  50, 62, 68, 57, 70, 56, 83, 67, 67, 59))

If it is possible to have lines of commands that don't correspond to variables, then add a first line of
#Randomized block example with 12 plots.

This assumes you have a single field in the examples.  If 2 fields , then 
Variable  Command
                #Randomized block example with 12 plots.
block       gl(4, 3)
treat        c("C", "A", "B", "B", "C", "A", "A", "B", "C", "A", "C", "B")
yield        c(74, 68,  50, 62, 68, 57, 70, 56, 83, 67, 67, 59))

Patowhiz:
@rdstern for the Construct examples, I have already added the example you [mentioned ](https://github.com/africanmathsinitiative/R-Instat/issues/5579#issuecomment-555903181) on the issue #5579 which is being addressed by this PR. If you are not okay with the previous example that you gave I can change to the new example that you are proposing.

The try was the one that was giving me a bit of challenge. I needed to tidy it up a bit.
The only other thing remaining is the tooltips that you wanted for the command examples. I have already setup the code structure for that, the only thing remaining are the examples themselves.

rdstern:
@Patowhiz I had forgotten that - and the examples you have now are better than the ones I made quickly today.
With the new Try I put all the example lines in and it was happy - said ok.  (And it was fine.)
I then clicked in the last line - first field - and got rid of the name.  The green Command Ok remained intact 
Without the name it is not ok, but the Ok button is still enabled.  When I click on ok nothing happens and the Ok is now disabled.
I tried obvious illegal names and it traps those nicely. 
Here is an example in this vein:
image


All pretty minor though.  I also get the following:
image

I did have the closing bracket - to get the Command Ok on Try.  Then got rid of the bracket and the message still stays.

Patowhiz:
@rdstern in regards to empty column name, the try gives an okay because in R, it is okay to create a dataframe without giving column names. So your above missing column names will translate to: 

`data.frame(no=c(0,0.5,1,2:28), seq(30,1,-1)) `

Where the last column will just be **seq(30,1,-1)**. So what the try gives that as an okay command cause it's an okay command. That is what I had in mind and you can try this in the script window. However, the R-Instat Ok button is disabled because it validates not just correct names but missing names too even though you can create a dataframe by just giving expressions .  But for consistency I have changed this implementation to not confuse users. So as long as okay is disabled, the try will be show error (though I wonder if that is a correct thing to do?).

I have fixed the other bug too

rdstern:
@maxwell - do you want to have a quick look at the code - or just merge?

Or would you prefer to do the tooltips?
Construct:
c( )  produces a vector.  Here it produces a variable (column) of 30 values between 0 and 28 
seq produces a regular sequence.  Here from 30 down to 1 in steps of -1
rep repeats one or more values as shown here.  Try also rep(1:5, each =6) and rep((1:5,length = 30)
c( ) and rep or seq can be combined as shown here.  Try also c(rep(1:5,each=2),seq(6,100,length=20))
regular sequence of dates. Alternatively use Prepare > Column: Date > Generate Dates
Alternatively use Prepare > Column: Date > Generate Dates
More to come!

rdstern:
Tooltips for the Command Option
I am still concerned with your first option, namely the empty data frame.  The problem is that it does work and gives the following:
image
 Now you can't do anything with it and it crashes later.
What about **_data.frame(row=1:100)_** instead as the simplest that R-Instat can begin with. 

This is actually quite useful - more useful than the default from the empty data frame.

Tooltips:
Data frame with a single variable to build on.
The same data frame as the default from the Empty option of this dialogue. 
4 variables, showing use of seq and rep functions 
2 variables with dates.  Or use Prepare > Column: Date > Generate Dates to add a date variable.
2 variables including generating a sequence of times
illustrates the gl function and generates data for a simple experiment on 12 plots

rdstern:
Random keyboard
3 variables including both random normal and uniform data
20 variables, each a random sample from a binomial distribution 
**_Next one change the command slightly to read initially data.frame(year = matrix...
Then the variables are given as year.1 etc rather than data.1_**
20 years of rainfall-type data using binomial (p=0.4) and exponential (mean = 10) variables
10 variables and 50 observations, sampling with replacement and probability 0.3 of TRUE 
49 variables, illustrating most of the types of data in the Wakefield package
10 variables from the Wakefield package
ID plus 10 categorical variables each with 5 levels
ID plus 10 categorical variables (as factors) and 10% of missing values
ID, plus 12 variables.  6 are indicator (dummy) variables, 3 Likert and 3 grades

Patowhiz:
@rdstern  Remaining is just the tooltip for the last example for construct option. "occasions"

rdstern:
Yes, I didn't have anything more to say.  Could just be "Regular sequence of monthly data".

rdstern:
Shadrack makes an excellent point, about the tidyverse style guide, because I view this dialogue as a good illustration of how readers can use R-Instat to start learning about writing R commands.  So they should know about good grammar.

@shadrack in the example you give the only issue I can see is the guideline on spacing.  So

`data.frame(x=1:30, s=rep("Reading",30),r=seq(1, 6.8, length=30),t=seq(1,60,2))`, becomes 

`data.frame(x = 1:30, s = rep(""Reading"", 30), r = seq(1, 6.8, length = 30), t = seq(1, 60, 2))`

So here the rules are:
a) Space after a comma - and not before a comma, i.e. as in ordinary English.
b) Space round "ordinary operators, like = and +, but not round operators with high precedence, e.g. 1:30 remains without spaces, so does x^3.
@shadrackkibet   Are there any other of these rules that we have broken in the examples here?

If not, then I suggest we obey those rules - Patrick, ok for you to add the spaces?  The style guide is interesting and is [here](https://style.tidyverse.org/_main.pdf).

(I was going to suggest we break the rule if the line gets very long.  But I see our best example almost keeps the rule - so that's not needed.  So there are already quite a lot of spaces here:

`data.frame(block = gl(4, 3), treat = c("C", "A", "B", "B", "C", "A", "A", "B", "C", "A", "C", "B"), yield = c(74, 68,  50, 62, 68, 57, 70, 56, 83, 67, 67, 59)) `

This seems to obey the rules already - probably because I copied it from R documentation!

rdstern:
@Patowhiz I now have a new example for the Commands button.
`data.frame(Iindex = c(1:10,NA,12:15), site =c( "Kisumu",rep(NA,7),"Naivasha",rep(NA,6)),x=c(1:7,99,8,5,7,3,6,88,77))`

The tool-tip can say.  "Data to illustrate replacing of missing values (Prepare > Data Frame > Replace Values) " 

Then @maxwellfundi I hope this could be merged soon.

Thanks

rdstern:
As still not yet merged I have yet another suggestion - but only if easy.  I am now looking at operations in R that only work for character variables, i.e. not for factors.

On our importing could we have a checkbox **_Convert Strings to Factor Columns_**.  
The default in R is TRUE and I am happy with that.  So the default is that it is ticked.  Then the command proceeds as now.  If unticked, then it just adds  **_,stringsAsFactors = FALSE_** to the data frame command.

Patowhiz:
@rdstern and I have agreed that it's not really necessary to add the option of  convert strings to factor columns.

Patowhiz:
More work is needed in the ucrTry control to adapt it to this dialog. Therefore this dialog will use it's own implementation for now.

@africanmathsinitiative/developers.  this is ready for review.

instat/dlgNewDataFrame.vb
```diff
@@ -272,37 +287,123 @@ Public Class dlgNewDataFrame
         'shows a popup that displays the example commands
         Dim frm As New Form
         Dim lstView As New ListView
+
         frm.ShowInTaskbar = False
         frm.FormBorderStyle = FormBorderStyle.None
-        frm.Size = New Size(ucrInputCommand.Width, ucrInputCommand.Height)
+        frm.Size = New Size(ucrInputCommand.Width, ucrInputCommand.Height + 50)
```
dannyparsons: We can't use any fixed sizes like `50`. This needs to be relative to other sizes otherwise it won't look correct on all machines.

Patowhiz: @dannyparsons this is technically relative to other sizes(dynamic). I'm just adding 50 pixels more to make the borderless form show all the commands. This was previously requested by @rdstern. I could remove the 50 but that means users will have to scroll the listview to see all the random example commands.

Patowhiz: I've opted to remove the 50, I didn't see a better easy way to do this to just add the extra pixels that were needed for the random examples

dannyparsons:
Ok that's fine. @volloholic can you review?

instat/dlgNewDataFrame.vb
```diff
@@ -312,36 +413,52 @@ Public Class dlgNewDataFrame
...
+
             If rdoConstruct.Checked Then
-                strScript = clsConstructFunction.ToScript
-            ElseIf rdoCommand.Checked Then
-                If ucrInputCommand.IsEmpty Then
-                    ucrInputTryMessage.SetText("")
-                    ucrInputTryMessage.txtInput.BackColor = Color.White
-                    Return
-                End If
+                Dim clsTempConstructFunction As RFunction
```
rdstern: Move this to the top

Patowhiz: Done

PR #5583, opened 2019-11-18
Added data types for the wakefield keys

### LINKED ISSUE #4347: Random samples and Wakefield package

Wycklife:
Fixes #4347
 This fixes the last part of the issue #4347
@africanmathsinitiative/developers This can be reviewed.

### DIFF
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -1514,6 +1514,7 @@ Public Class ucrCalculator
     End Sub
 
     Private Sub cmdGroup_Click(sender As Object, e As EventArgs) Handles cmdGroup.Click
+        ucrReceiverForCalculation.SetDataType("factor")
         If chkShowArguments.Checked Then
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("wakefield::group(n = , x = c(""Control"" , ""Treatment"") , prob = NULL , name = ""Group"")", 65)
         Else
@@ -1522,6 +1523,7 @@ Public Class ucrCalculator
     End Sub
 
     Private Sub cmdHair_Click(sender As Object, e As EventArgs) Handles cmdHair.Click
+        ucrReceiverForCalculation.SetDataType("factor")
         If chkShowArguments.Checked Then
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("wakefield::hair(n = , x = c(""Brown"" , ""Black"" , ""Blonde"" , ""Red""), prob = c(0.35 , 0.28, 0.26 , 0.11), name = ""hair"")", 98)
         Else
@@ -1546,6 +1548,7 @@ Public Class ucrCalculator
     End Sub
 
     Private Sub cmdInternet_Browser_Click(sender As Object, e As EventArgs) Handles cmdInternet_Browser.Click
+        ucrReceiverForCalculation.SetDataType("factor")
         If chkShowArguments.Checked Then
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("wakefield::internet_browser(n = , x = c(""Chrome"" , ""IE"", ""Firefox"" , ""Safari"" , ""Opera"" , ""Android"") , prob = c(0.5027 , 0.175 , 0.1689 , 0.0994, 0.017, 0.0132) , name = ""Browser"")", 149)
         Else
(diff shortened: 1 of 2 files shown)


Wycklife:
@rdstern I have added the data types for the keys proposed, Area and Zip keys are missing, should I add them? Otherwise, kindly have a look at the data types.

instat/ucrCalculator.vb
```diff
@@ -1695,6 +1720,8 @@ Public Class ucrCalculator
     End Sub
 
     Private Sub cmdSex_Inclusive_Click(sender As Object, e As EventArgs) Handles cmdSex_Inclusive.Click
+        ucrReceiverForCalculation.SetDataType("factor")
+        ucrReceiverForCalculation.SetDataType("factors")
         If chkShowArguments.Checked Then
```
maxwellfundi: @Wycklife  why do you have these two types here - `factor `and `factors`?

Wycklife: @maxwellfundi Thanks for that observation. Am correcting this now.

rdstern:
@Wycklife and @maxwellfundi I must have done something wrong here.  I thought I took your branch - it is with the improved keyboard.  I expected to see for example coin replaced by coin (f).  And also Hair replaced by hair (f), i.e. also the lower case h.  But it seems to be the same as before.

instat/ucrCalculator.vb
```diff
@@ -1599,6 +1615,8 @@ Public Class ucrCalculator
     End Sub
 
     Private Sub cmdLorem_Ipsum_Click(sender As Object, e As EventArgs) Handles cmdLorem_ipsum.Click
+        ucrReceiverForCalculation.SetDataType("character")
+        ucrReceiverForCalculation.SetDataType("character")
```
maxwellfundi: @Wycklife  there seeems to have some repeated code here. Please get rid of one of these lines.

instat/ucrCalculator.vb
```diff
@@ -1464,6 +1472,7 @@ Public Class ucrCalculator
     End Sub
 
     Private Sub cmdEducation_Click(sender As Object, e As EventArgs) Handles cmdEducation.Click
+        ucrReceiverForCalculation.SetDataType("factors")
```
maxwellfundi: @Wycklife here you also have `factors` other than `factor`

instat/ucrCalculator.vb
```diff
@@ -1360,6 +1361,7 @@ Public Class ucrCalculator
     End Sub
 
     Private Sub cmdAnswer_Click(sender As Object, e As EventArgs) Handles cmdAnswer.Click
+        ucrReceiverForCalculation.SetDataType("factors")
```
maxwellfundi: @Wycklife `factors `here again- change it to `factor`

rdstern:
@maxwellfundi and @Wycklife I still wonder if there is a misunderstanding on what is needed here.  I think the functions automatically specify the type of the resulting variables.  What I was asking for was that the visible keys on the keyboard carry this information.  So when you see the key, then you know what type of variable to expect so the key itself says hair (f).  I got the variable types by simply generating "the works" and seeing what types of variable results.

Wycklife:
@rdstern There was a misunderstanding on this, let me redo the issue. Should I close this pull request?

rdstern:
Ok,  sorry I didn't detect earlier.  Yes, I think it could be a new pull request.

Wycklife:
I'm closing this due to the unwanted features

PR #5586, opened 2019-11-19
Segmented keyboard#5584

### LINKED ISSUE #5584: Adding a segmented keyboard to the Model > Use Model dialogue
I hope this will be an interesting challenge for @Ivanluv ?  
It uses the new features in the Use Model dialogue that he has added recently.  So I hope this can be merged first.  I will discuss with @dannyparsons 

It also extends analyses that uses the **_Model > Two Variables_** dialogue that I hope can soon be extended.

The new keyboard has the following keys:
segmented, (double size) **_print, summary
confint, vcov, predict, plot
lines, points_**, aapc, broken, 
davies, intercept, pscore, slope

In the commands that are generated the ones in bold don't have segmented:: in front.  The others do.  In these 
broken is broken.line
davies is davies.test
pscore is pscore.test.

One complication is that the main command - segmented seems to need both the model and also the data.  So in addition to providing access to the model I also had to add the lines for the data as in:
data1 <- data_book$get_data_frame(data_name="data1")
attach(what=data1)
I think this is just for the segmented key.

Ivanluv:
fixes #5584

### DIFF
--- a/instat/dlgUseModel.vb
+++ b/instat/dlgUseModel.vb
@@ -286,7 +286,7 @@ Public Class dlgUseModel
     End Sub
 
     Private Sub cmdPlotLines_Click(sender As Object, e As EventArgs) Handles cmdPlotLines.Click
-        ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("plot lines()", 1)
+        ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("plot()", 1)
     End Sub
 
     Private Sub cmdPoints_Click(sender As Object, e As EventArgs) Handles cmdPoints.Click
--- a/instat/dlgUseModel.vb
+++ b/instat/dlgUseModel.vb
@@ -23,6 +23,7 @@ Public Class dlgUseModel
     Public bUpdating As Boolean = False
 
     Private clsAttach As New RFunction
+
     Private Sub dlgUseModelLoad(sender As Object, e As EventArgs) Handles Me.Load
         If bFirstLoad Then
             InitialiseDialog()
@@ -53,6 +54,7 @@ Public Class dlgUseModel
         ucrInputModels.IsReadOnly = True
 
         bUpdating = False
+
     End Sub
 
     Private Sub SetDefaults()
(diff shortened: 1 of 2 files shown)


instat/dlgUseModel.vb
```diff
@@ -229,4 +232,60 @@ Public Class dlgUseModel
...
+        ucrReceiverForTestColumn.AddToReceiverAtCursorPosition(" predict()", 1)
+    End Sub
+
+    Private Sub cmdPlotLines_Click(sender As Object, e As EventArgs) Handles cmdPlotLines.Click
+        ucrReceiverForTestColumn.AddToReceiverAtCursorPosition(" plot lines()", 1)
+    End Sub
+
+    Private Sub cmdPoints_Click(sender As Object, e As EventArgs) Handles cmdPoints.Click
+        ucrReceiverForTestColumn.AddToReceiverAtCursorPosition(" points()", 1)
+    End Sub
```
maxwellfundi: @Ivanluv  this looks good, Just a point of curiosity, why do we have consistent spaces when you add things to the receiver? is this to make sure there is proper spacing?

if so @dannyparsons can we use chr(32) just to be consisistent with what we do for quotes - chr(34)?

Ivanluv: we don't need the spacings @maxwellfundi  .Thank you for noticing this,am correcting this now

Ivanluv:
@rdstern  could you look at this

rdstern:
@Ivanluv this is working well now.
One thing was that plot lines should just be plot - both on the key and the command.

Then I suggest this be "parked" until the new main Use Model dialogue has been merged.  Then this branch should be updated next.

For reporting the area that I still can't get to work is the plotting, and that is because these are not ggplots.  That is a whole new ball game.  And the aapc key also doesn't work.  But this keyboard would still be useful even without those.  So no work is needed on those aspects now.

Wycklife:
@maxwellfundi I'm looking at it

rdstern:
Pleased this new keyboard has been merged.  Note that this is being merged into the "old" Use Model dialogue.  There is a new one - pull request #5537 that will update the same dialogue in a different way.  I hope there won't be conflicts

Issue #5587, opened 2019-11-23
Remember the directory when using the R-Instat library


rdstern:
Open from Library seems to work well.  I am thinking here of the Instat files.  When I use Open from Library or the library in Open and tidy NetCDF files it always goes to the right place.
I would like the ordinary File > Open from File to be separate from the Open from Library.  

Currently if I use Open from File I then go to the directory I need.  Then, the next time it returns to that directory.  However if I have an Open from Library then the Open from File has forgotten the previous directory and now takes me to the library.  I don't like that, because if I wanted to return to the library then I would have used that dialogue (or Library button in the Open NetCDF dialogue).
@volloholic says this is easy to do and could be a good task for @Patowhiz?  It should be done only in the Open from Library code.  It would proceed by saving the location of the last file there and then restoring that location when the library call is done.  There shouldn't need to be a change anywhere else.

I hope he is right!

Patowhiz:
Well, yes I agree this is doable at the **Open from Library** dialog. The reason as to why the dialog overwrites the previous values of open from file is because they are using a shared instance. When you just call an **anyformdialog.ShowDialog()**, basically what VB.Net does is it just instantiates the dialog for you and puts it in memory. When you call the same statement even somewhere else, it retrieves it from the memory instead of creating another. So you end up manipulating just 1 object. This is somehow memory efficient but it leads to such problems.

I can use another easier way and just have the **Open from Library dialog** instantiate it's **own** **Open from file dialog**. This will ensure that the dialog doesn't overwrite the shared object. That way I don't need to bother with saving and restoring whatever values were set when opening the dialog from a different location. Which will help in code separation etc.

rdstern:
Two more suggestions for the Open from Library.  At least the first (I assume) will be easy.
When I use Open from Library I su=ometimes want to look at multiple examples from the same package.  Currently it doesn't seem to remember the last file I used.  

As an example try the agrecolae package where there are many datasets.  When I return to the dialogue having opened a data frame I return to the top of the list of datasets, and not to where I was before.  I would like to carry on down  - looking at the datasets.  I assume it could remember where I was last?

More difficult (I assume) is to investigate reading more of the datasets, i.e. some that are more complicated than a single data frame.  I think quite a lot are a list of multiple data frames.  Here is an example I could read 

```
# Code generated by the dialog, Open Dataset from Library
utils::data(package="agricolae", X=DC)
DC <- DC
[...]

PR #5593, opened 2019-11-26
Remembering Path in Open File location

Patowhiz:
Fixes issue #5587

### DIFF
--- a/instat/dlgFromLibrary.vb
+++ b/instat/dlgFromLibrary.vb
@@ -116,7 +116,7 @@ Public Class dlgFromLibrary
     End Sub
 
     Private Sub cmdLibraryCollection_Click(sender As Object, e As EventArgs) Handles cmdLibraryCollection.Click
-        dlgImportDataset.bFromLibrary = True
+        dlgImportDataset.bFromLibrary = True 'set flag that this should open from library
         dlgImportDataset.bStartOpenDialog = True
         dlgImportDataset.ShowDialog()
         Me.Hide()
--- a/instat/dlgFromLibrary.vb
+++ b/instat/dlgFromLibrary.vb
@@ -26,6 +26,7 @@ Public Class dlgFromLibrary
     Private clsDataFunction As New RFunction
     Private dctPackages As New Dictionary(Of String, String)
     Private strAvailablePackages() As String
+    Private newDlgImportDataset As dlgImportDataset
 
     Private Sub dlgFromLibrary_Load(sender As Object, e As EventArgs) Handles Me.Load
         autoTranslate(Me)
@@ -106,9 +107,14 @@ Public Class dlgFromLibrary
     End Sub
 
     Private Sub cmdLibraryCollection_Click(sender As Object, e As EventArgs) Handles cmdLibraryCollection.Click
-        dlgImportDataset.bFromLibrary = True
-        dlgImportDataset.bStartOpenDialog = True
-        dlgImportDataset.ShowDialog()
+        'we need a new instance of the import dialog to not overwrite the open from file import dialog settings
+        If newDlgImportDataset Is Nothing Then
+            newDlgImportDataset = New dlgImportDataset
+            newDlgImportDataset.Text &= " from Library" 'to distinguish it in the list of recent used dialogs
+        End If
+        newDlgImportDataset.bFromLibrary = True
+        newDlgImportDataset.bStartOpenDialog = True
+        newDlgImportDataset.ShowDialog()
         Me.Hide()
     End Sub
 
--- a/instat/dlgFromLibrary.vb
+++ b/instat/dlgFromLibrary.vb
@@ -190,14 +190,12 @@ Public Class dlgFromLibrary
     End Sub
 
     Private Sub TestOkEnabled()
-        If rdoDefaultDatasets.Checked AndAlso lstCollection.SelectedItems.Count > 0 Then
-            ucrBase.OKEnabled(ucrNewDataFrameName.IsComplete)
-            cmdHelp.Enabled = True
-        Else
-            ucrBase.OKEnabled(False)
-            cmdHelp.Enabled = False
-        End If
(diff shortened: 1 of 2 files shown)


maxwellfundi:
@Patowhiz  Are you able to look at this issues/change requested by @rdstern

Patowhiz:
I suggest the changes to the Open and Tidy NetCDF dialog (which @dannyparsons may be working on) be moved to another issue.

@africanmathsinitiative/r-instat this is ready for review

rdstern:
I agree this should be merged now.  Danny is working on the netCDF dialogue for now. I will add the requirement as another issue later.  It is very useful to have this implemented.

instat/dlgFromLibrary.vb
```diff
@@ -106,9 +116,14 @@ Public Class dlgFromLibrary
     End Sub
 
     Private Sub cmdLibraryCollection_Click(sender As Object, e As EventArgs) Handles cmdLibraryCollection.Click
-        dlgImportDataset.bFromLibrary = True
-        dlgImportDataset.bStartOpenDialog = True
-        dlgImportDataset.ShowDialog()
+        'we need a new instance of the import dialog to not overwrite the open from file import dialog settings
```
dannyparsons: I'd like to discuss this method with @volloholic before merging.

Patowhiz:
@dannyparsons @volloholic  For this PR to now fully fix the issue modification will have to be made on frmMain. So that it can locally save the current selected file name to automatically reopen it. I can only do that after the other frmMain changes are merged.
 I've already built in the methods required for this in the import dataset dialog.

Patowhiz:
@rdstern @dannyparsons this is ready for review

instat/dlgFromLibrary.vb
```diff
@@ -178,10 +192,11 @@ Public Class dlgFromLibrary
     Private Sub TestOkEnabled()
         If rdoDefaultDatasets.Checked AndAlso lstCollection.SelectedItems.Count > 0 AndAlso ucrNewDataFrameName.IsComplete Then
             ucrBase.OKEnabled(True)
+            cmdHelp.Enabled = True
         Else
             ucrBase.OKEnabled(False)
+            cmdHelp.Enabled = False
```
dannyparsons: The help should only be disabled when no item is selected. If `ucrNewDataFrameName` is not complete I should still be able to click the help.

instat/dlgFromLibrary.vb
```diff
@@ -203,14 +218,6 @@ Public Class dlgFromLibrary
         frmMain.clsRLink.RunScript(clsHelp.ToScript, strComment:="Opening help page for" & " " & lstCollection.SelectedItems(0).Text & " " & "dataset. Generated from dialog Open Dataset from Library", iCallType:=2, bSeparateThread:=False, bUpdateGrids:=False)
     End Sub
 
-    Private Sub EnableHelp()
-        If rdoDefaultDatasets.Checked AndAlso lstCollection.SelectedItems.Count > 0 Then
```
dannyparsons: This is the correct check for enabling the help.

instat/dlgFromLibrary.vb
Patowhiz: This is now fixed

Patowhiz: This is fixed

instat/dlgFromLibrary.vb
```diff
@@ -190,13 +190,14 @@ Public Class dlgFromLibrary
     End Sub
 
     Private Sub TestOkEnabled()
-        If rdoDefaultDatasets.Checked AndAlso lstCollection.SelectedItems.Count > 0 AndAlso ucrNewDataFrameName.IsComplete Then
-            ucrBase.OKEnabled(True)
+        If rdoDefaultDatasets.Checked AndAlso lstCollection.SelectedItems.Count > 0 Then
+            ucrBase.OKEnabled(ucrNewDataFrameName.IsComplete)
```
dannyparsons: It would be clearer to separate out TestOKEnabled from setting the Help button enabled. This works, but is not good practice to do two different things in one place.

Help button changes should only be triggered by `rdoDefaultDatasets` and `lstCollection` changing.

Patowhiz: They are both relatively doing the same things, checking when to enable or disable controls. Separating them means I have to call 2 functions in several places instead of one that's intended to do the same thing, this adds more unnecessary lines of code making it not to be concise and thus reducing clarity. Nevertheless, I have returned it to the way it was (split into 2 and calling the other inside the other)

dannyparsons: They are doing "relatively" similar things, but not the same things. Setting the OK button and Help button are different concepts that are set under similar, but not the same conditions. 
If their conditions were to diverge further in future, then this code becomes more difficult to understand and maintain.

So separating them into two function calls **adds clarity** and is not "unnecessary" because they are different concepts. Having them mixed together is very unclear. 
Separating them does add lines of code, but we are much more concerned by clarity than minimising number of lines.

instat/dlgFromLibrary.vb
```diff
@@ -176,14 +190,14 @@ Public Class dlgFromLibrary
     End Sub
 
     Private Sub TestOkEnabled()
-        If rdoDefaultDatasets.Checked AndAlso lstCollection.SelectedItems.Count > 0 AndAlso ucrNewDataFrameName.IsComplete Then
-            ucrBase.OKEnabled(True)
-        Else
-            ucrBase.OKEnabled(False)
-        End If
+        ucrBase.OKEnabled(rdoDefaultDatasets.Checked AndAlso lstCollection.SelectedItems.Count > 0 AndAlso ucrNewDataFrameName.IsComplete)
         EnableHelp()
```
dannyparsons: I know this was already here before, but this shouldn't be run within `TestOkEnabled()`. It's the same issue of clarity and separating out different concepts. This should be run under a method that handles changes to the two controls it depends on.

Patowhiz: Fixed it. Now calling this in areas that affect help button

dannyparsons:
@rdstern can you do a final test of this?

PR #5610, opened 2019-12-05
Adding column chart option to dlgBar and Pie chart

### LINKED ISSUE #5464: Adding a column chart button to the bar chart dialogue

Ogik99:
fixes #5464 


Ogik99:
@Patowhiz you can check this now.

rdstern:
It looks good.  I have checked that it seems to do all as before, and returns nicely to the dialogue on repeated occasions.  I have not checked the code itself.
@Patowhiz From what I have seen I would be happy for this to be merged.

Later a useful challenge for the graphics generally will be to be able to reproduce examples where the articles use the code.  Here is an example, called dot plots and bar charts from [Iowa University](https://homepage.divms.uiowa.edu/~luke/classes/STAT4580/dotbar.html).

Ogik99:
@Patowhiz any final comments

maxwellfundi:
@Ogik99 does this have improvements ready?

PR #5621, opened 2019-12-16
Improvements to two var graph dialog

### LINKED ISSUE #5582: Two variable graph dialogue
Basing this on the design of the 2-variable summarise might be quick - I hope, and could be great.
a) There would be just the same main part with numeric by numeric, etc as in summarise.
b) Then the current sub-dialogue is very limited.  It doesn't give options of display, just the type of graph.  I strongly propose that this be moved to the main dialogue, where only 1 of the 4 options is available - depending on the combination of numeric by numeric, etc.
c) If there are multiple elements in the receiver, then there will usually be facets - I assume with one for each of the receivers.  This will often be a situation where Free y scale is needed, so that could also perhaps be on the main dialogue.  It is currently nowhere.
d) The by option would have a facet for each level of the second variable.  (Rather than considering it as the x-variable in the plot.  So with multiple there would usually be 2 facet variables.  So, thyere might be by (row) and by (col) as options where you choose how to use the second variable.
e) We keep a sub-dialogue for other options of the plotting.  This could usefully become a mini-version of the main plotting sub-dialogue.  For example, for facets the only option may be margins.  (Or we may be able to include our main graphics sub-dialogue.  In that case I would like margins to be on the main dialogue, i.e. we add a few more common potions to make it rare that the sub-dialogue is needed.  There may also be a useful half-way house, where we have the full set of options and also a special small set of common additions.)

dannyparsons:
- dialog design consistent with new two variable summary layout
- additional types of graphs included in lists
- options not yet available and to be added
fixes #5582


dannyparsons:
- options subdialog and common options added to the main dialog

rdstern:
This is looking great now.  I am trying mainly the many factors by a variate - as that has changed.
The default is fine - here is jitter:
image


But when I flip it isn't fine:

Aha - the code is always `ggplot2::facet_wrap(facets=~variable, scales="free_x")`  If you flip, change this to `"free_y"`.  Then you get this - which is what you want.  Currently it is a mess.

image

More generally on facets you don't give me access to them.  I can see why not, but I might want to change the number of columns, etc.  I know, by giving access, you could mess up the plot but does that matter?  Or is it the stacking of the data?  If I can't have access to them generally, then a simple facets options sub-dialogue might be what's needed?

Now a bigger feature - just might be really easy?  The 2-variable summary and graph are now partners.  On the Two-variable summary could I have a simple **_Graph_** checkbox.  Default is unchecked.  If I check, then I get whatever graph is the default for that combination.  But it picks up the current settings of the graph dialogue.  So, if I wanted violin plots, then I just opt for those on the graphs dialogue and that is then what I get with the summarise!

I agree that it is good enough to merge and we continue testing.

rdstern:
I checked bar charts.  Very promising.  But I am supposed to find things that are wrong.
a) In the survey data I used the same variable in the Y and the X.  (I know that's not sensible, but that's not the point!) It gave an error that it couldn't find the variable that I had duplicated.
b) So I took it out of the Y.  Then OK, but your free axis still needs to change here too.  I was flipping and here is the graph:
image

PR #5626, opened 2020-01-07
Improvements to the calculator.

### LINKED ISSUE #5623: Improvements to the calculator - mainly character keyboard
1. In the **_factor keyboard_** the concat key is disabled.  Please replace it by the **_labelled_** key from the **_labelled package_**.
2. The remainder is the strings  keyboard.  
a) The split key currently gives stringr::str_split_fixed().  This gives an error.  Delete the **__fixed_**, so it is str_split(). 

3. There are some useful additions to the stringr package to go onto the Strings (Character Columns) keyboard.
a) Key with label **_starts_** for stringr::str_starts.
b) Key with label **_ends_** for stringr::str_ends.
c) Key with label **_squish_** for stringr::str_squish.
d) Key with label **_encode_** for stringr::str_conv.
e) Key with label **_remove_** for stringr::str_remove
f) Key with label **_remove2_** for stringr::str_remove_all
g) Key with label **_extract2_**, for stringr::str_extract_all
h) Key with label **_locate2_**, for stringr::str_locate_all
i) Key with label **_replace2_**, for stringr::str_replace_all

4 The following keys all need a pattern= argument, namely 
a) count, 
b) detect, 
c) extract, 
d) locate, 
e) replace, 
f) split, 
g) remove, 
h) remove2, 
i) starts, 
j) ends.
I suggest that the code generates, for example, stringr::starts( ,pattern=) with the focus, as now just after the open bracket.  (As the pattern argument is compulsory it would help to give it.) @dannyparsons may wish to comment on this.  An alternative is not to give ",pattern=", but just the comma.  
And for **_replace_** there should be another comma, because the replacement string is also needed.  Or it could be **_,replace=_**.

5. Order of the keys.  The keyboard is currently:
image

a) I suggest leaving the top 2 rows except put combine where replace is now.
b) Make it as bit wider and add squish and encode to the top 2 rows.
c) Then have the 3rd and 4th rows in roughly alphabetical order (they are now all the commands that include a pattern, so possibly
Row 3: count, detect, end, extract, extract2
Row 4: locate, locate2, remove, remove2, split 
Row 5: replace, replace2, starts
6. A new section (row of the keyboard with label **_modifier_**. 
This has 4 keys, namely:
boundary, collate, fixed, regex. 
These keys are used within one of main commands on the keyboard.  When the key is pressed it needs to include the name of the package, e.g. stringr::boundary("words").  So, for example the command stringr::count(texts,stringr::boundary("words")) would count the number of words in each string, rather than characters.

7. Beneath this is a further set of keys  - see next message.

Wycklife:
Fixes #5623
This fixes the first 5 parts of the issue.
@africanmathsinitiative/developers This can be reviewed


Wycklife:
@rdstern I have fixed number 1-5. I need more clarification on number 6 and 7 kindly.

rdstern:
Hi @Wycklife 
I reply first on the changes you have done to items 1 to 5.  Then later to clarify 6 and 7.
1) Make the dialogue a bit wider for this keyboard.
2) Add the "stringr::" to the new keys as shown in item 3 in the list.  I assume it is needed?
3) Did you do item 4?  I couldn't see the pattern = argument in those keys?

Wycklife:
@rdstern Thanks for the reply, I did item 4. but I can as well have a check on them.

rdstern:
Here is an indication of the complete keyboard(s) with the items 6 and 7:
(Not quite like this of course - I hope the other keyboards won't go so far down in the dialogue.  But I hope the picture explains what could be useful.  And this relates also to your proposed presentation in the Zambia meeting.

image

Wycklife:
@rdstern I have implemented the two keyboards. Could you kindly check whether the parameters passed by the keys are correct.  I have changed the location of the base so as to allow full display of the string keyboard but I have realized it affects other keyboards i'e creating more space. I will ask @dannyparsons to help me solve the problem once he's done with the session.

Wycklife:
@rdstern I'm happy to make changes you'll suggest.

Wycklife:
@africanmathsinitiative/developers This is now ready for review.

rdstern:
@Wycklife this is looking good.  So just small suggestions.
a) The group box for Modifiers has an extra line that can be deleted (Compare with the others, Strings or Symbols where there is no extra line.)
b) The symbols keys would ideally be of similar size (as far as possible) so they line up vertically.
c) In the symbol keys you have sometimes added **_brackets_** - first 2 rows - and I am not sure they are needed.  Can you try without.  Then I assume the \ s symbol will have \s space where there wasn't room for \ s (space)?
d) In \ d and \ s there should be no space between \ and s - or \ and d.
e) In the [ ] and [^ ] leave the cursor within the brackets - after the ^. It is now after the brackets.
f) In the {n} and {m,n} keys the key should deliver { } and { , }, i.e. without the n or m and with the cursor just after the first bracket.
g) On the bottom row the + 0 or 1 key should just deliver the + and the same for the other 2 keys.

Wycklife:
@rdstern I have made the changes. Kindly have a look.

maxwellfundi:
@rdstern  this seems to have been fixed from your comments above. 

Cheers.

PR #5634, opened 2020-01-22
View data frame

### LINKED ISSUE #4831: Small edit on the call to the View command.

Ivanluv:
@africanmathsinitiative/developers  this is ready for review.Fixes #4831

### DIFF
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -682,9 +682,12 @@ Public Class ucrDataView
     End Sub
 
     Private Sub ViewSheet_Click(sender As Object, e As EventArgs) Handles ViewSheet.Click
+        Dim strScript As String = ""
         clsViewDataFrame.AddParameter("x", clsRFunctionParameter:=clsGetDataFrame)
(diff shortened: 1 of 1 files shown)


instat/ucrDataView.vb
```diff
@@ -683,10 +683,12 @@ Public Class ucrDataView
 
     Private Sub ViewSheet_Click(sender As Object, e As EventArgs) Handles ViewSheet.Click
         Dim strScript As String = ""
+        Dim strTemp As String
         clsViewDataFrame.AddParameter("x", clsRFunctionParameter:=clsGetDataFrame)
-        clsViewDataFrame.SetAssignTo(grdCurrSheet.Name)
-        clsViewDataFrame.ToScript(strScript)
-        RunScriptFromDataView(strScript, strComment:="Right Click Menu: View R Data Frame", bSeparateThread:=False)
+        clsGetDataFrame.SetAssignTo(grdCurrSheet.Name)
+        clsGetDataFrame.ToScript(strScript)
```
dannyparsons: This line shouldn't be run. You only want to call `ToScript` on `clsViewDataFrame`.

Issue #5613, opened 2019-12-09
Problem deleting the only data frame


rdstern:
The bottom **_right-click Delete_** calls the **_Prepare > Data Object > Delete Data Frames_** dialogue.

This is generally working well, but gives a nasty message if you try to delete the only (or possibly also all) the data frames.  It also appears not to work, but I think it sort of does, because the data frame disappears when a new one is opened.  It possibly doesn't know what to do when there is no data frame left. 
It is easy to delete all (or the last) data frame - use **_File > Close Data File_** instead.

Could this dialogue trap this situation.  It would ideally either call the File > Close Data File code, or at least refuse and give the message for the user to use File > Close Data File option instead.

lloyddewit:
I'd be happy to look at this. Please could you assign it to me? thanks

dannyparsons:
@lloyddewit This message box appears, which is generated in the RLink. The first check would be to see what is triggering this.
I believe it should be able to cope with deleting all the data frames, whether you do it through Close File or deleting all the data frames.
It may be that the RLink is not recognising that all the data frames are gone and is trying to access something that isn't there. Or possibly the empty `Data Book` is not returning the correct information from the R side.

![image](https://user-images.githubusercontent.com/11214914/72810499-c3f67700-3c6e-11ea-8863-9f9db8b00e51.png)

lloyddewit:
@dannyparsons Thank you for the extra information.
The R code generated by dlgDeleteDataFrames looks correct:
```R
data_book$delete_dataframes(data_names="diamonds")
```
As described above, this seems to work correctly unless the last data frame is deleted. In this case, the grid still displays the last data frame's data and displays the error message box.

The message box is triggered in `clsGridLink.UpdateGrids` line 135. The conditions that trigger the message box are:
```vb
        If frmMain.clsRLink.bInstatObjectExists Then
            'various operations and checks that don't trigger errors ...
            ':
            If (bGrdDataExists AndAlso (bGrdDataChanged OrElse bRDataChanged)) OrElse (bGrdVariablesMetadataExists AndAlso (bGrdVariablesMetadataChanged OrElse bRVariablesMetadataChanged)) Then
                expTemp = frmMain.clsRLink.RunInternalScriptGetValue(clsGetDataNames.ToScript(), bSilent:=True)
                If expTemp IsNot Nothing Then
                    lstDataNames = expTemp.AsList
                Else
                    If Not bMessageShown Then
[...]

PR #5637, opened 2020-01-24
Solves #5613 problem deleting only data frame.

lloyddewit:
Created new function `clsRLink.GetDataFrameCount()` which returns the number of data frames in `data_book`.
Added an extra check in `clsGridLink.UpdateGrids()` so that the grid is only updated if there is at least one data frame.


dannyparsons:
Thanks @lloyddewit I've also tested and the bug is resolved.

dannyparsons:
If you write "fixes #5613" in the commit message or pull request comment then it will automatically close the issue when the pull request is merged.

PR #5646, opened 2020-01-29
Tab order for the Describe>Specific dialogues

### LINKED ISSUE #5510: Tab order/index on all dialogs
@africanmathsinitiative/developers in preparation for the change coming here https://github.com/africanmathsinitiative/R-Instat/issues/5508, this will only work correctly if the tab order is set correctly for each dialog. 
Can I ask that as you do other work, have a quick check that the dialog you're working on has tab order set up correctly. It's very quick to do, if you're not sure about it ask @maxwellfundi.
If anyone feels particularly inspired to do more than this, then checking sets of dialogs for the tab order would be really useful.

Ivanluv:
fixes #5510 


rdstern:
@Ivanluv that's great and I am happy to test.  I assume having a defined tab order will also solve a problem I have had, which is to be able to press OK when the bottom row of controls is not visible.

What are the "rules" you have applied in setting the tab order?  Just so I know what I am checking for?

Ivanluv:
I do not think this will solve  the problem of solving the press Ok when the bottom row is not visible  since it will move through all the buttons  at the bottom row.You therefore won't know on which button is the tab.

The "rule" is that ,the tab order starts with the dataframe selectors on the left of the dialogues ,then moves to the receivers on the  right of the dataframe selectors.This starts from the first receiver to the last before moving to the other  controls.We then end with the base control

rdstern:
Many thanks.  I'll check now.

rdstern:
@Ivanluv I am checking the Specific dialogues in turn.  At the same time I am also checking the Help button.  This correct from some, but not others.  It seems a good time to sort that out too.
**_Frequency Tables_** - tab looks ok.  Help not working.  Context is 425
**_Summary Tables_** - tab looks ok.  Help not working. Context is 426
**_Scatterplot_** - please check with @dannyparsons .  It goes from the first receiver to the Single Variable button and then to the second receiver.  Is that going to be ok, or should the Single/Multiple Variable button be before the first receiver?  Help works ok.
**_Line plot_** Single/multiple query as for scatterplot.  After Add peaks, logically tab to add valleys. Help works ok.
**_Histogram_** Where do/should the buttons at the top feature in the tab order?  It doesn't seem to tab back to them?  Same Single/Multiple query as scatterplot. Help works ok.
Can you check on the points above - I'll continue testing later.

maxwellfundi:
@Ivanluv  this looks good but a few files are missing.  Have these already have working help?

**Parallel Coordinate** plot. ok. Needs help - ID = 593
**Mosaic plot** - seems ok Needs help - ID = 594

dannyparsons:
@maxwellfundi this will conflict with https://github.com/africanmathsinitiative/R-Instat/pull/5610. Please look into what to do.

PR #5656, opened 2020-02-04
File Save As Changes

### LINKED ISSUE #5594: Minor change in File menu
In writing the Help for the R-Instat dialogues I suggest a small change in the position of one of them.  This is easy to do, but there are probably reasons for the current position.

We have File > Save As and then File > Export
The distinction seems reasonable for most of the items.
1. I suggest that **_File > Export > Export Graph_** as Image be moved to **_File > Save As_**.  It could then be at the bottom and **perhaps** with a line before.  It is (sort of) saving a graph window. It changes it's name to Save Graph as Image.

2. In the dialogue itself the heading should also change.  So should the comment - though that is perhaps automatically generated?

The distinction between the File > Save As and File > Export menus is not that great.  But this would leave the File > Export to be exporting data - usually somewhere else for more work on the analysis.

The File > Save As currently includes File > Save As > Save Output Window As... which already saves the graphs that are in that window.  It seems odd to distinguish that from the Save Graph as Image.

3. At the same time, 2 of the entries in the File > Save As menu mention the Window, and 2 don't.
I suggest (for consistency) that we add Window to Save Log As.  So it reads Save Log Window As...

(I then quite like it that 3 of the Windows are saved.  This isn't the case for the data, where the Window you see is only showing part of the data - they are in R and often much more than is in the window. )  

This is all cosmetic.  If no response from @volloholic or @dannyparsons I assume they agree with these 2 small changes.

Patowhiz:
Fixes #5594
@maxwellfundi this may need to be merged sooner because it's a minor design change on the frmMain


Patowhiz:

PR #5665, opened 2020-02-05
Set the The time for the tooltip and also added the new try to the calculator

### LINKED ISSUE #5567: Change the time a tooltip is visible
Tool-tips have now been added to a variety of commands.  This is currently mainly in the calculator, but I suggest we should extend this feature to other dialogues.

One problem I am having is that the tooltip disappears automatically after a fixed time.  The default, which we are using now is 5000 milliseconds (5 seconds).  I am finding this is too short for some of the recent tooltips - where I have given a short example.

If you move the focus away from a control then it goes off anyway.  So I don't see a big problem if we lengthen this time.

I had a look [here](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.tooltip.autopopdelay?view=netframework-4.8) where there is a description on how to make the change.

I wonder if this could/should be a global change?  I suggest doubling the period it is visible would be useful.  Please could this be discussed and hopefully implemented.

Wycklife:
Fixes #5567 

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -59,7 +59,7 @@ Public Class dlgCalculator
         ucrCalc.chkSaveResultInto.Checked = True
         SaveResults()
         ucrCalc.ucrSelectorForCalculations.bUseCurrentFilter = False
-        'ucrCalc.ucrTryModelling.SetRSyntax(ucrBase.clsRsyntax)
+        ucrCalc.ucrTryCalculator.SetRSyntax(ucrBase.clsRsyntax)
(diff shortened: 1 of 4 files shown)


maxwellfundi:
@Wycklife these seems to conflicts here! Did you update the branch with the master before starting work?

maxwellfundi:
I get it, you did two pieces of work on same dialogue on different branches and this was due to the other PR being merged so it causes conflicts. We will try solve this tomorrow

maxwellfundi:
Closing this! @Wycklife  is opening a new PR with these changes

Issue #5539, opened 2019-10-04
Adding new keys to the extRemes keyboard in Use Model dialogue


rdstern:
There is an existing keyboard for the extRemes package in Model > Use Model dialogue.

I propose an extra first row with print, summary and plot as the 3 keys.  They are each described in the extRemes guide within the fevd entry.

The extRemes print and summary currently work well.  The plot seems to work, i.e. it doesn't give an error, but there also isn't a visible plot!  I hope @dannyparsons can diagnose this problem.

I don't know whether this might be undertaken by @Wycklife who seems to be an expert on keyboards or by @Ivanluv who is looking as other aspects of this dialogue?  Perhaps they could discuss.

Wycklife:
@rdstern I'm happy to pick it up.

rdstern:
The three keys at the top of this menu each give
extRemes::plot.fevd(fevd object) etc 
This doesn't work.  Instead just give `plot(object)` and `summary(object)` and `print(object)`.  

It doesn't need the package name in front and it doesn't need plot.fevd, etc.  Just plot will do!

The other complication is that plot and print don't produce another object.  They just do the plot.  If you then save the resulting object and display it, it just says NULL.

What should that mean for the code?  I suggest that is a @dannyparsons question.

It would be really good if this keyboard were working.  I will use the ideas in the AIMS course.

I have been using the POTRw dataset - from the extRemes package to do the testing.  It is one of the examples in the documentation for the fevd command.

lilyclements:
This is a really simple one! @Ag-Derek can you assign?

This is for changes in the dialog found under "Model > Use Model > Use Model Keyboards..."
Then under the "extRemes" package, the top row has three buttons. These currently run

`print.fevd`, `summary.fevd`, `plot.fevd`

Can this be changed to just be

`print`, `summary`, `plot`

![Image](https://github.com/user-attachments/assets/4119ea3e-cde5-4821-9bad-4efed07ad08f)

Ag-Derek:
@Emmanuel-Afrifa can you take this up

PR #5666, opened 2020-02-05
Resized the model dialog

Wycklife:
Related to #5539

@africanmathsinitiative/developers Someone to check whether the size of the dialog is now okay and that the controls can be clearly seen


Wycklife:
@rdstern I have removed the new keys as you suggested but resized the dialog

PR #5667, opened 2020-02-05
Frm Main Changes

### LINKED ISSUE #5616: Improving the toolbar and windows easily!
Here is our current toolbar:

image

Two icons also include the arrow for a menu, and the File > Open is used here.

I have a small series of suggestions to make the toolbar more useful:

1.  The icon above shows 5 previous files and that is the same as at the bottom of the file menu.  In the file menu 5 is enough, but here we could easily have more and that could be useful.  I suggest 10 here. 
2. In #5228 I suggest having a similar pull-down menu on the graph icon above, particularly to make it easier to use plotly.
3. In the icon for the log window could we have another with **_Log Window_** and **_Script Window_** being the 2 options.
4. Similarly for Column Metadata could we have one with Column Metadata and Data Frame Metadata being the 2 options.
5. Something a bit different and I am not sure where it fits.  Our current windowing system is limited, but useful.  This is not the time for an overhaul.  However, there is one layout that would be very useful on occasions.  I would call it "Swap data and metadata view".   I would use the column metadata much more if that were there.  It would show all variables and the right-click menu ius already there.  
This would be another option on the View menu and then on the new little pull-down menu.

shadrackkibet:
Fixes  #5616 and #5670


dannyparsons:
@rdstern have you tested this?

rdstern:
Yes - there are some further minor changes to be done.  It is looking good.

dannyparsons:
@shadrackkibet @ZIRANA11 The name of the file should follow the [tidyverse style guide for names](https://style.tidyverse.org/files.html#names) i.e. `climatic_guide_datasets.xlsx`

shadrackkibet:
@rdstern could please test this again.

maxwellfundi:
@dannyparsons  @rdstern  Could you have a look at this soon. has many files so worth to merged soon.

instat/clsRLink.vb
```diff
@@ -1471,6 +1471,19 @@ Public Class RLink
...
+        Dim clsLastGraph As New RFunction
+
+        clsInteractivePlot.SetPackageName("plotly")
+        clsInteractivePlot.SetRCommand("ggplotly")
+        clsLastGraph.SetRCommand(strInstatDataObject & "$get_last_graph")
+        clsLastGraph.AddParameter("print_graph", "FALSE", iPosition:=0)
+        clsInteractivePlot.AddParameter("p", clsRFunctionParameter:=clsLastGraph, iPosition:=0)
+        'Need to set iCallType = 2 to obtain a graph in an interactive viewer.
+        RunScript(clsInteractivePlot.ToScript(), iCallType:=2, strComment:="View last graph as Plotly", bSeparateThread:=False)
+    End Sub
```
rdstern: This should be an option on the view last graph function

shadrackkibet: This is now done

instat/clsRLink.vb
```diff
@@ -1464,24 +1464,22 @@ Public Class RLink
...
-        clsLastGraph.SetRCommand(strInstatDataObject & "$get_last_graph")
-        clsLastGraph.AddParameter("print_graph", "FALSE", iPosition:=0)
-        clsInteractivePlot.AddParameter("p", clsRFunctionParameter:=clsLastGraph, iPosition:=0)
-        'Need to set iCallType = 2 to obtain a graph in an interactive viewer.
-        RunScript(clsInteractivePlot.ToScript(), iCallType:=2, strComment:="View last graph as Plotly", bSeparateThread:=False)
+        If bAsPlotly Then
+            clsInteractivePlot.SetPackageName("plotly")
+            clsInteractivePlot.SetRCommand("ggplotly")
+            clsLastGraph.SetRCommand(strInstatDataObject & "$get_last_graph")
+            clsLastGraph.AddParameter("print_graph", "FALSE", iPosition:=0)
```
dannyparsons: These two lines should be the same for both versions, and therefore should be done once outside the `If` statement.

shadrackkibet: This is fixed here #5686

PR #5671, opened 2020-02-05
Fixes #5262 Data receiver in View dialogue does not clear

### LINKED ISSUE #5262: Data receiver in View dialogue does not clear
I used the View Data dialogue to put some data (12 rows - monthly data) into the output Window.
These columns were rainfall totals and also number of rain days.  I first displayed the set of rainfall totals.
image

I then returned to the dialogue.  I **cleared the receiver** (right-click) then Clear), and put in a different set of variables (number of rain days) to display in the output window.

It worked fine, except it displayed what was previously in the receiver, i.e. it ignored the fact the receiver now had a different set of columns to display.

I assume (hope) this issue is just for this dialogue?
I attach a picture of the dialogue with the receiver inconsistent with the output.

image

The previous set of columns are still being displayed.

After reset (of course) I get what I wanted, see below:

image

lloyddewit:
Fixes #5262 by ensuring that that the R script is updated whenever the contents of the `Variable(s) To View` window is updated.


lloyddewit:
I made a mistake with the commits so I'll close this pull request, fix the commits and make a new pull request.

PR #5672, opened 2020-02-05
Changing right-click in reogrid a lttle

### LINKED ISSUE #5242: Changing right-click in reogrid a lttle?
Currently in R-Instat the main right-click menu is only activated from the name field.  
In Genstat there is a similar right-click menu, but it works from within the grid as well.

It (like R-Instat) has some options that recognise the variable - like convert to factor, and others that simply open a dialogue - like Reorder columns.

I find that I have hardly, if ever, used the special current right-click from within the grid itself.  So I suggest we extend the main right-click, so it can be activated from within the grid as well as from the name field.

I have suggested this for Version 0.4.31.  But it may need discussion, and may not be a trivial change - even if agreed.  If so then change the milestone

Wycklife:
Fixes #5242

### DIFF
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -635,7 +635,7 @@ Public Class ucrDataView
         dlgSort.ShowDialog()
     End Sub
 
-    Private Sub mnuFreezeToHere_Click(sender As Object, e As EventArgs) Handles mnuFreezeToHere.Click
+    Private Sub mnuFreezeToHere_Click(sender As Object, e As EventArgs) 
         Dim strLastSelectedColumn As String
         Dim strSelectedColumns As String()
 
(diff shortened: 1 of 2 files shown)


Wycklife:
@rdstern The change is made.

instat/ucrDataView.vb
```diff
@@ -786,7 +784,11 @@ Public Class ucrDataView
...
     End Sub
 
-    Private Sub mnuHelp_Click(sender As Object, e As EventArgs) Handles mnuHelp.Click
+    Private Sub mnuHelp_Click(sender As Object, e As EventArgs) Handles mnuCellHelp.Click
         Help.ShowHelp(Me, frmMain.strStaticPath & "\" & frmMain.strHelpFilePath, HelpNavigator.TopicId, "134")
     End Sub
+
+    Private Sub lblHeader_Click(sender As Object, e As EventArgs) Handles lblHeader.Click
+
+    End Sub
```
dannyparsons: This empty sub should be removed.

dannyparsons:
Was this code reviewed? There is some redundant code that should be removed.

Patowhiz:
@dannyparsons I'm now working on this control in the pasting to data frame PR. I also noticed so much redundant code. I can remove the empty sub. The rest I think can be part of the bigger discussions we have on improving the code quality.

PR #5675, opened 2020-02-06
Fixed the button examples problems on the new dataframe dialog

Patowhiz:
I was still looking at what was causing the change to the text for button examples


dannyparsons:
It's useful to document in the pull request the cause and solution. 
The cause was the translation class changing the button text and removing the translation code in the dialog fixed this.

Patowhiz:
Sorry forgot that.

PR #5585, opened 2019-11-19
Added the label with (click Ok to export data) to dlgExportDataset

Wycklife:
Fixes #5530
Fixes the last part of #4347
@africanmathsinitiative/developers Someone to review this.

PR #5677, opened 2020-02-06
Added a label text to export dialog

Wycklife:
Fixes  #5585

@rdstern This now works well


rdstern:
Here is the dialogue now:
image

I like the comment - which was the main reason for the change.  But:
a) Why the extra space below the OK buttons
b) Why the extra space to the right - after the browse button
c) Why is the browse button larger than before
d) Why the incomplete "box" for the filename

This was supposed to be such a simple change.  It seems to go on forever?

Wycklife:
@rdstern I have resolve the problem. The designer is now working fine.

Wycklife:
@rdstern Done.

shadrackkibet:
Can we change the colour of this label to alert the user on clicking OK? perhaps make it red?

rdstern:
I'd be happy with that change.  It would make the label stand out.

shadrackkibet:
Good to also improve the design of this dialog. The **Export File** label is far from the corresponding control.  Align the label with the textbox.

Wycklife:
@rdstern I have changed the color of the text to red.

shadrackkibet:
You should do the same on the save dialog.

PR #5680, opened 2020-02-06
Import multiple NetCDF files 

### LINKED ISSUE #4849: Import multiple similar netcdf files
Suggestion from @rdstern 
> It would be good to be able to read data from multiple (similar?) NETCDF files at once.

This could be needed quite often e.g. CM SAF data comes as a zipped tar file containing one netcdf file per time point.

A common operation is to choose a time period (subset of files) and location(s) (subset within each file). This could be done in one step at import.

If we implement this, I would suggest we don't do this directly from a zipped file, but a folder or set of files could be chosen.

dannyparsons:
fixes #4849


rdstern:
Great.  For my part this is ready to be merged

PR #5681, opened 2020-02-06
Implementing right click

### LINKED ISSUE #5242: Changing right-click in reogrid a lttle?

Wycklife:
Fixes #5242

@rdstern This is now okay


dannyparsons:
@Wycklife try to have informative commit messages. They are useful if we ever need to look for specific changes in the history.

Wycklife:
@dannyparsons noted with thanks

maxwellfundi:
@dannyparsons  i think this code is looking good

Issue #3152, opened 2017-06-02
Weighted analyses in R-Instat?


rdstern:
When we share R-Instat we will get comments relating to details, and some that are more major and general.  Here is an example of a major query, from Carlos - SSD.  This is just from describing what we have now done with the thought he would be impressed.
Danny/Roger:  We now have quite a lot on tables, particularly summary tables and will also have frequency tables, which is one of the things you particularly asked for.
Carlos: Great.  Can they do weighted tables?  In SPSS there is a very simple dialogue to specify a column of weights.  Then the statistics are subsequently based on those weighted samples.
let me provide an example from our simple survey dataset.  This has 36 farmers.  One of the variables is the field size.  There are 161.5ha in total  Suppose you wanted the mean yield.  Then it could be the simple mean per farmer.  This is unweighted sum(yield)/36.  Or it could be per ha, in which case we want to weight each yield by the size of the field i.e. sum(size*yield)/Sum(size).

What to do?  We could 
a) Ignore it on the grounds that this is not our primary audience.
b) Accommodate it, but we don't have much time.
c) See what we can do first.

I have taken option c) hence this post.
a) It would be very easy to have a dialogue to specify a column as being of weights - like SPSS.  We could note this in the "type" of column if we wish.  We could certainly store it in our meta-data.  Note usually this will be a numeric column, but (with our new "system" perhaps is could sometimes be defined as a factor column.
b) Some dialogues already allow for weights, in particular the one-way frequencies and two way frequencies.  We could (by default) change those dialogues - as we do with climatic - so, when a column of weights exists, then the box is ticked by default and the weight column is inserted.  (I prefer this to SPSS - where it does it, without anything on the dialogue to indicate the weights.)
c) Why didn't the more general 3-way frequencies dialogue have the weights?  I remember we discussed this feature.  It is now important, if it is easy to add.
d) We could add the survey package to our set of packages.  We will need this package in the future.  It seems to have some nice examples of surveys as well as methods for using weights.  We will need this package in the future - so let's add it now.
e) The survey package includes a command ftable.svystat which modifies (I think) the ftable we use already for frequency tables into one that would cope with weights.
f) We can ask David and Danny how feasible it is to add weights to their summary system.  It is there in the "old" Instat package summary/tables system which David wrote.  Why isn't it in our new system!!!

rdstern:
[...]

PR #5683, opened 2020-02-06
Weighted summaries  and hydroGOF package  verification functions 

### LINKED ISSUE #5630: Verification of satellite data in R-Instat
This is an important area to work on and also to add facilities to R-Instat.  I collect here the initial e-mail correspondence by Shadrack and Danny.  I then consider what we might do in R-Instat, starting with activities in the forthcoming code sprint.

"Dear Roger,

This is just to report some of my findings on the methods used to compare Satellite and in-situ data. Following up on my discussion with Rija (CDT developer) he shared some papers. One looks interesting for me and can be found [here](https://www.scirp.org/pdf/ACS_2016102811260514.pdf). However, I am still going to check others. There are various methods used in this paper which include;
1.	Bias - which we have in R-Instat but the way it is calculated here is a bit different, so we might want to offer this option as well.
2.	Mean Absolute Error (MAE)
3.	Root Mean Square Error (RMSE)
4.	Pearson's correlation (r)
5.	Probability of Detection (POD)
6.	False Alarm Ratio (FAR)
7.	Critical Success Index (CSI) 
8.	Volumetric Hit Index (VHI) 
9.	Volumetric False Alarm Ratio (VFAR) 
10.	Volumetric Miss Index (VMI) 
11.	Volumetric Critical Success Index (VCSI)    
12.	More in the "hdroGOF" package
We already have some of these somewhere in R-Instat and there is also a workaround in R-Instat to obtain some of these. The "hydroGOF" R package has a number of these measures implemented and Rija uses this package for some of the methods he implemented for CDT. I wonder if we can take advantage of this package and implement some in R-Instat. However, algorithms for these methods look simple so implementing isn't a huge task at least. 
My thoughts.
Regards,
Shadrack"

Danny replies as follows:

"Hi Shadrack,

That list looks good. It's similar to what I've seen others using. I've separately sent some papers from Ghana doing comparisons, but I think they mainly use methods from your list.

This site is very comprehensive [with methods](https://www.cawcr.gov.au/projects/verification/) . It's called methods for forecast verification but it mentions these are general for comparing two columns of data.
What I like about the site is it describes what each method will tell you and what it is appropriate for and it's advantages and disadvantages.

Danny"

The site from Danny is excellent.  I now use these two messages as background to propose some activities to make the verification easy - and useful - in R-Instat of course.

shadrackkibet:
Fixes #5630 and #3152


shadrackkibet:
Here are the screenshots
image
image

dannyparsons:
I would move the checkbox to the right so its not on the edge of the group box.

shadrackkibet:
Updated design.

image
image

shadrackkibet:
This can now be tested fully. All weights and hydroGOF methods have been implemented.

rdstern:
This is great and I have started testing.
My main questions (minor) are on the layout.  
1. On the main dialogue I would still prefer to have the weights between the 2 multiple controls.  They apply to the variables in the top control, so it is good if they are just below that.
2. The sub-dialogue has become very wide.  Would it be possible for the check-boxes to be re-ordered, and then make it closer to the previous width.  I am thinking of the second 6 being 
Index of agreement, Mean error, Mean absolute error, Mean squared error, Percent bias, Volumetric efficiency, in that order.  They are all quite short, so quite a lot of space can be saved.
3. Could there be a help button in that block that calls the hyderoGof package - as there is for some of the modelling dialogues?

shadrackkibet:
1. Great, I will move the receiver.
2. Another alternative is to use initials e.g **RMSE** for Root Mean Square Error as the names for the checkboxes and then have the full names as part of the tooltip. 
3. Yes, it is possible to add the help button.

shadrackkibet:
Updated design

image
image

shadrackkibet:
Yes, you are right I am using `spatstat` package for the weighted median. Since we are already using  `Hmisc`  package elsewhere I think it is ideal to make use of it and I am already using it for other weighted summaries. They do not have a weighted median as a function, but since it has weighted quantile function I can use this for the weighted median by just fixing `prob=0.5` and that should work.

rdstern:
Sounds good.  I am less concerned about omitting the weights package.  It looks quite useful.  The problem with spatstat was its great size, and even thew size of the manual with over 1700 pages I think.  The weights package looks pretty small.

rdstern:
@shadrackkibet might you get this ready to be merged?  I would continue with the weights package - which should then be added.  We should omit the spatstat package.  It could then be merged even if the (many) statistics that don't permit weights are not disabled.  Just make a new issue for that to be done later.

shadrackkibet:
Yes, I will do that now.

rdstern:
I may be doing something wrong, but I am unable to test your branch.  It says "Object reference not set to an instance of an object".  I wonder if you know what might be going on.  I am able to move to other branches.

shadrackkibet:
I have updated my branch with the master. Could you try it again?

dannyparsons:
I'll also try and test this.

shadrackkibet:
@rdstern are you able to test this now?

rdstern:
This is a major new set of features.  There are weighted summaries and the goodness of fit statistics from the hydroGOF package.
I would like to use these on the forthcoming AIMS climatic statistics course.  I am not sure if the weights package is yet included - I have added it, so can't test.
From my point of view this could now be merged.  I think @dannyparsons or @lloyddewit may have something to say about the code first?

shadrackkibet:
Anything remaining for me to address here? Otherwise this can be merged.

lloyddewit:
@rdstern asked me to review. Sorry I didn't do it yet. Please could you give me until Thu 30 April?

rdstern:
From me there are some new points - sorry.
Not new, but on the weighting you were going to do something to indicate which statistics can be weighted?  Could be to disable the others - though there are a lot, or add a (w) on those where weights would apply, or perhaps put them in in italic?
Then I used your branch to prepare a lot of the session for the students.  You were there in week 2. 
 So I am really excited that the facilities can soon be merged.  But it did show that some improvements - almost all cosmetic - would be useful.

[Session 2 - Gridded Data part 2 (pptx) (Roger Stern's conflicted copy 2020-04-22).pptx](https://github.com/africanmathsinitiative/R-Instat/files/4521480/Session.2.-.Gridded.Data.part.2.pptx.Roger.Stern.s.conflicted.copy.2020-04-22.pptx)

You will see there 5 or 6 slides that used your functions.  Of course when they are used I found various improvements I would like - particularly to make them simpler.  Sorry - but I think the results will be great!

a) I found it quite difficult to relate some of the names you give to the variables compared to those in the hydroGOF manual.  And the names are too long.  So could you change the names to the short versions that are used in the guide. So the names would be cp, d, etc, both in the dialogue and in the resulting data frame.
b) That will mean the dialogue can be narrower again!  Also, ideally the names you have given, e.g. Coefficient of persistence could be included as the variable labels - but only if that is easy to do   
c) In the slides I prepared I spent quite a time thinking of the order. that would be most useful for the resulting variables.  Could you consider this.  To me the order under ggof on page 16 of the guide looks sensible - though it may omit some statistics.  This again would be both in the dialogue and the resulting data frame.
d) Some functions have extra arguments.  One to include is j=1 (with the default of 1)  This is for the modified Index of agreeement and modified Nash Sutcliffe.  I would just include it once after all the statistics - and as j = .  (There could be a tooltip for an explanation if you wish.)
e) If easy do you want to include a checkbox in the block for Select All.  (This is just like in the filter sub-dialogue to select all levels of a factor.) 
f) Finally you didn't give the correlation, which is fine.  But you did give R2 and it didn't seem to me to be the same as the square of the correlation.  It would be good to check that.

shadrackkibet:
I was wondering whether having the long name as the tooltip is necessary, might be informative than the abbreviations.  

**Question**: Which order is preferable to you in the dialog, should I go vertically or horizontally?

rdstern:
I would be happy to have the long names as tooltips.  I still wonder whether (in addition) they could also be labels in the variables derived.
I would go vertically

shadrackkibet:
I have addressed most of these points (above). Two remaining (see below), @rdstern please test to confirm.

> f) Finally you didn't give the correlation, which is fine. But you did give R2 and it didn't seem to me to be the same as the square of the correlation. It would be good to check that

I will be checking today.

> Not new, but on the weighting, you were going to do something to indicate which statistics can be weighted? Could be to disable the others - though there are a lot, or add a (w) on those where weights would apply, or perhaps put them in italic?

I would like to hear what @dannyparsons would say about this. I can implement either.

rdstern:
I'll check and good idea to discuss with @dannyparsons

shadrackkibet:
I was wondering whether you might have different values of  `j`  for the two options. So having this for each option will simplify things rather than having to use the dialog twice. 

Even if we remove the two inputs there won't be much change in the width of the dialog. See below comparison with the previous size (not significantly different in size). This is because of the tabs at the top (in yellow) if we reduce the width some of the tabs will have to be hidden. 

image

rdstern:
ok, then don't worry about the width.  I still think that having the individual values of j is not particularly important.  And just having j = 1 visible in the bottom right-hand corner - near the Help button will be fine.  Usually j will be kept at 1.  In the rare situation that you want to have it different from 1 and also different for the 2 situations (very rare!) then you would just have to use the dialogue twice.

shadrackkibet:
@rdstern I checked what you reported about coefficient of determination. 

From `hydroGOF` manual the coefficient of determination (`br2()` the function name) is more robust compared to the usual coefficient of determination (`r2`). There is a nice example which illustrates one problem with the "usual" `r^2`  given in the manual and I copy it below;
```# Looking at the difference between r2 and br2 for a case with systematic
# over-prediction of observed values
obs <- 1:10
sim1 <- 2*obs + 5
sim2 <- 2*obs + 25
# The coefficient of determination is equal to 1 even if there is no one single
# simulated value equal to its corresponding observed counterpart
r2 <- (cor(sim1, obs, method="pearson"))^2 # r2=1
# 'br2' effectively penalises the systematic over-estimation
br2(sim1, obs) # br2 = 0.3684211
br2(sim2, obs) # br2 = 0.1794872
```
Therefore, what is computed by `br2()` which is referred to as the coefficient of determination in `hydroGOF` package is not necessarily the square of correlation `r`. I didn't go deep into how this is calculated in the package, but from my understanding, the interpretation remains the same.

lloyddewit:
On 23 April I said that I'd try and peer review before 30 April (see above). However there still seems to be ongoing work. It's best if I peer review when the code changes are done. 
@shadrackkibet - please let me know when you think the code changes are complete, thanks

shadrackkibet:
@rdstern what is your take on this. I have worked on issues above maybe you could have a look at the latest changes before @lloyddewit reviews the code?

rdstern:
will do - later today

shadrackkibet:
@lloyddewit this is to let you know that you can now proceed with the review. Thanks.

instat/sdgSummaries.vb
```diff
@@ -43,7 +45,6 @@ Public Class sdgSummaries
         ucrChkNTotal.SetText("N Total")
 
         ucrChkMean.SetParameter(New RParameter("summary_mean", 4), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "summary_mean" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
-        ucrChkMean.SetText("Mean")
```
lloyddewit, line 46: I understand that the standard place to call `SetText` would be here in the `InitialiseControls` function. However, I see that for these 14 controls,  `SetText` is now called in `SetRFunction`. What is the rationale for this? is it so that `strWeightLable` can be appended?

instat/sdgSummaries.vb
```diff
@@ -274,25 +261,131 @@ Public Class sdgSummaries
...
         Next
         bControlsInitialised = True
-        MissingOptionsVisibilty()
+        TwoVariablesControls()
         PositionOptions()
         OrderByCheckEnabled()
     End Sub
 
-    Public Sub SetRFunction(clsNewRFunction As RFunction, clsNewDefaultFunction As RFunction, clsNewConcFunction As RFunction, Optional ucrNewBaseSelector As ucrSelector = Nothing, Optional bReset As Boolean = False)
+    Public Sub SetRFunction(clsNewRFunction As RFunction, clsNewDefaultFunction As RFunction, clsNewConcFunction As RFunction, Optional ucrNewBaseSelector As ucrSelector = Nothing, Optional bReset As Boolean = False, Optional strNewWeightLabel As String = Nothing)
```
lloyddewit, line 363: `strWeightLabel` is not related to the RFunction, it seems strange to set it in `SetRFunction`. Is there a more logical place where it could be set?

instat/sdgSummaries.vb
shadrackkibet: The reason why I have this here is that `InitialiseControls` subroutine is only run once which won't work for in this case because the label needs to be updated whenever the user comes back to the sub dialog.  This depends on whether the weights receiver is filled or not. I agree with you that this might not be an ideal place to append the label but clearly this won't work if I set the label on `InitialiseControls`  otherwise this label needs to updated elsewhere to add (**W**) text. Do you have a suggestion where I could do this?
Yes, 
>so that strWeightLable can be appended

 The rationale for this is to indicate which summaries are weighted because not all summaries are weighted at the moment. Another alternative would have been to disable those summaries that are not weighted but these summaries are many so I opted to add (**W**) at the moment.

instat/sdgSummaries.vb
shadrackkibet: Yes, `strWeightLabel` is not related to RFunction. My understanding is that this subroutine I nickname it "a bridge" could be used to pass anything from the main dialog to the sub dialog. So, it is not always things related to RFunction that is passed through, which means the naming of this subroutine isn't precise!  Another alternative is to have `public` variable in the main dialog class which I can then access it in sub dialog which will need to update accordingly. At the same time, a subdialog can be used in more than one dialog so passing variables etc through one place will make it easy to track. I, therefore, do not see any other logical place to set this.

instat/dlgColumnStats.vb
```diff
@@ -114,7 +124,10 @@ Public Class dlgColumnStats
     End Sub
 
     Public Sub SetRCodeForControls(bReset As Boolean)
-        SetRCode(Me, ucrBase.clsRsyntax.clsBaseFunction, bReset)
+        'temporary fix
```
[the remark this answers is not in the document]
shadrackkibet: Yes, this is still temporary at the moment. I will try set conditions for the weight checkbox which is the reason I added this so that the state of this control is retained whenever the dialog is reopened. So a long term solution is to set conditions for the weights checkbox.

lloyddewit:
@shadrackkibet Thank you for the clear explanations. My peer review comments are all addressed and I'm happy for the PR to be merged (even though it has already been done!).

shadrackkibet:
Thanks, @lloyddewit for your review.

PR #5684, opened 2020-02-06
Minor edit on dlg Row names/numbers

Ogik99:


dannyparsons:
Even for a trivial change, it's important to say what the change is that you're making. Otherwise I can't tell if what you've changed in the code is doing what you intended to do.

PR #5686, opened 2020-02-06
Minor Rlink change

shadrackkibet:


instat/clsRLink.vb
```diff
@@ -1468,16 +1468,16 @@ Public Class RLink
         Dim clsLastGraph As New RFunction
         Dim clsInteractivePlot As New RFunction
 
+        clsLastGraph.SetRCommand(strInstatDataObject & "$get_last_graph")
+
         If bAsPlotly Then
+            clsLastGraph.AddParameter("print_graph", "FALSE", iPosition:=0)
```
dannyparsons: Shouldn't this be done for both or it shouldn't be there for the normal viewing?

shadrackkibet: We need this to be "TRUE" for normal viewing.

dannyparsons: Ok good, just checking

PR #5689, opened 2020-02-07
Added stat_identity options for the pie chart

### LINKED ISSUE #5464: Adding a column chart button to the bar chart dialogue

Ogik99:

fixes #5464
This implements the "stat_identity" option for the pie chart.


instat/dlgBarAndPieChart.vb
```diff
@@ -78,11 +78,11 @@ Public Class dlgBarAndPieChart
...
         ucrReceiverByFactor.SetLinkedDisplayControl(lblByFactor)
-        ucrPnlOptions.AddToLinkedControls(ucrReceiverY, {rdoBarChart}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
-        ucrReceiverY.SetLinkedDisplayControl(lblYvariable)
+        'ucrPnlOptions.AddToLinkedControls(ucrReceiverY, {rdoBarChart}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        'ucrReceiverY.SetLinkedDisplayControl(lblYvariable)
 
-        ucrPnlOptions.AddToLinkedControls({ucrInputYValue}, {rdoBarChart}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
-        ucrInputYValue.SetLinkedDisplayControl(lblYValue)
+        'ucrPnlOptions.AddToLinkedControls({ucrInputYValue}, {rdoBarChart}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        'ucrInputYValue.SetLinkedDisplayControl(lblYValue)
```
dannyparsons: Should these be removed if they're not needed any more?

rdstern:
I assume by "stat identity" you mean the option for the pie chart to use a variable - as is done for the bar chart.
If so, then this works.  
The only issue  - that is different from the bar chart is that when you choose the option "Variable", then it correctly opens the control for the variable.  But OK is enabled - which it isn't for the bar chart in the same way.  So you can press OK without giving a variable and that generates an error.  So just to sort out that small point and it can be merged.

shadrackkibet:
@Ogik99 did you sort out the second point from @rdstern above? any help needed?

lloyddewit:
@shadrackkibet I have a favour to ask. @Ogik99 is not currently active on R-Instat and it would be nice to close this PR. Please could you resolve the last issue identified by @rdstern so that we can merge? thanks if this is possible

lloyddewit:
@shadrackkibet Please ignore my request above. I noticed that @Ogik99 has a number of open PRs. I will look at all these together to see how best to resolve. 
But be warned, I may ask for your help again later ;)

lloyddewit:
@Ivanluv Please could you try and complete this PR? (@rdstern suggested that you would be the best person for this task)

lloyddewit:
@rdstern I had to make some small changes to resolve merge conflicts. Therefore your approval is needed again. The merge conflicts only affected the position of the controls on the dialog. If the dialog's appearance looks OK to you then it is probably safe to approve, thanks

rdstern:
@Ivanluv the changes made have resulted in a slight problem with the position label and corresponding control.  Once that is resolved (again!) this can be merged.
image

Ivanluv:
Am looking at this

PR #5690, opened 2020-02-07
UcrSave TextBox Resizing Bug

Patowhiz:
Fixes a  bug that made the textbox keep on shrinking everytime you set the label text


dannyparsons:
@Patowhiz which dialog was this causing a bug on so that I can test?

Issue #3577, opened 2017-07-03
Complete in-line documentation for all code. 


volloholic:
All code should have appropriate in line documentation describing all functions methods and structures.

dannyparsons:
Inline documentation can be done as shown here https://docs.microsoft.com/en-us/dotnet/visual-basic/reference/language-specification/documentation-comments 
It has lots of uses such as:
- can be automatically utilised within the code as tooltip to developers using that code e.g. when using a function the documentation for that function and its parameters displays as tooltips to the developer
- can be used by other tools to produce automated printed and online documents

It's quite easy to make, I have started using it for Climsoft. Visual Studio has autocomplete features which makes it easy to write out.

@rdstern @africanmathsinitiative/developers something to discuss at the code sprint. We can make specific issues if we decide on any tasks related to this.

dannyparsons:
@michaelclapham link to inline documentation above

lloyddewit:
Inline documentation was also discussed in PR #5985 by @dannyparsons , @Patowhiz , @shadrackkibet and @lloyddewit.

It was agreed that we should select one or more dialogs to be 'model dialogs' for commenting. The model dialogs should be fully commented and referenced in the developer's guide. The other dialogs would only need comments for their non-standard methods and behaviour. 

**Selected comments from PR #5985:**

> We have been using inline documentation for functions. Explain what the function does and it's parameters.
> Also, we have comments on global variables or properties of the dialog, briefly explaining what they are used for.
> 
> In addition to this, I'm usually keen on commenting my "choice" of Implementation especially when it deviates from the standard norm of doing things in R-Instat e.g manually checking for R-Parameters. And possibly pointing on the issue that explains the problem further.
> 
> Sometimes due to limitations of the "base code", I would also normally comment on my choice of implementation and how it could be made better once the "base code" is improved.

> That sounds reasonable. In general I would suggest documenting core parts of the code and code which is non standard or non obvious. I think that is a reasonably general view in software.
> 
> So in dialogs I would only comment things which are both not common to dialogs in general and might not be clear to someone looking at it for the first time. But the standard stuff and non standard stuff that is obvious what it's doing and why I don't think should be cluttered with comments.

[...]

PR #5694, opened 2020-02-07
Contributes to #3577 partial inline documentation for `clsRLink.vb`

lloyddewit:
Contributes to #3577 but does not solve it. `clsRLink.vb` still needs inline documentation completed. I just checked it in because I will not be able to work on it again for at least a week and I want to avoid merge problems later.
@dannyparsons You do not need to review the contents of the comment blocks now. It is probably easier to review all the comments together when the inline documentation of this file is complete.


lloyddewit:
Got message 'This branch has conflicts that must be resolved'. But the conflicts did not seem to be real. I will close this pull request and make a new one to see if this solves the problem.

PR #5696, opened 2020-02-09
Access to R documentation in Help menu

### LINKED ISSUE #4762: Access to R documentation in Help menu

shadrackkibet:
@africanmathsinitiative/developers This is ready for review. Fixes #4762


rdstern:
This is a good improvement and I suggest could be merged as it is.
If it is trivial, then 2 small suggestions are as follows:
a) When I look for vignettes are none exist, then it just gives the command in the output window.  Can you detect this instance and add "No vignettes found for this package" or "No vignettes found for Evapotranspiration" - i.e. detect the package name?

b) If I open the function name and then leave it blank it says:
# Code generated by the dialog, Packages Documenation
utils::help(topic="xx", package="EnvStats")

No documentation for '' in specified packages and libraries:
you could try '??'

If I try xx it says:
# Code generated by the dialog, Packages Documenation
utils::help(topic="xx", package="EnvStats")

No documentation for 'xx' in specified packages and libraries:
you could try '??xx'

If I try ?? then it says
# Code generated by the dialog, Packages Documenation
utils::help(topic="xx", package="EnvStats")

No documentation for '??' in specified packages and libraries:
you could try '????'

Can we do better? I would be happy if just the first line were given, i.e. omit the "you could try ..." line.

The code remains to be checked.

shadrackkibet:
These messages are from R directly. However, we can change them but since it is not an error message and doesn't inhibit the user from doing anything then it might not be worth doing it at the moment.

maxwellfundi:
Sounds good then @shadrackkibet  and @rdstern  I will merge this now.

PR #5710, opened 2020-02-17
Modelling fixes

### LINKED ISSUE #5421: Problems with the fit 3-variable model
In the **Model > 3 variable > Fit model** it does not cope with missing values in the data.

The **Model > 2 Variable > Fit Model** and the **Model > General > Fit Model** do cope nicely.

@stevenndungu  worked on them and they seem to be working well.  The 3 variable must have been omitted.  This is both for the fitting itself and then for saving the fitted values and residuals etc.

The example I used is the Dodoma data.  From the data, use Climatic > Prepare > Climatic Summaries with the Within Year option, usinf the month_abbr.  Then summarise Tmin, getting the monthly mean.

This is now to be fitted using the **s_year** and the **month_abbr** variables. There are missing values at the top of the monthly summary columns.

There is a second problem that is more general.  In the 2-variable case the residual plots option is disabled.  And it is not ticked.
In the 3 variable and the general case it is also disabled, but ticked.  However there is no plot - at least initially.  When you use the dialogue a few times (3 variable and possibly general, then it appears.  As it is disabled it now appears all the time.

I would like the options to be possible, i.e. individual plots, etc.  I believe @stevenndungu had a problem there, which is why they were disabled.  I copy him in, in case he can advise - even though he has now left.

shadrackkibet:
Fixes #5421. @africanmathsinitiative/developers This is ready for review.

### DIFF
--- a/instat/dlgModelling.vb
+++ b/instat/dlgModelling.vb
@@ -109,7 +109,7 @@ Public Class dlgModelling
         ucrReceiverForTestColumn.SetMeAsReceiver()
 
         ucrBase.clsRsyntax.SetCommandString("")
-        ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("lm()", 1)
+        ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("lm(, na.action = na.exclude)", 25)
 
 
         ucrSaveResult.Reset()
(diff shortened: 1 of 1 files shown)


maxwellfundi:
From the code i do not see any obvious problems. Looks good.

PR #5720, opened 2020-02-18
Changes on the PICSA Rainfall sub dialog (Caption control)

### LINKED ISSUE #5491: PICSA Rainfall graphs with multi-line caption
One of the sub-dialogues for PICSA has Title, Sub-Title and Caption.

I suggest some edits to the title tab.
a) Delete the word Graph from Graph Title etc.  It is in a Graph box, so that's clear.
b) Move the controls over a bit and add a checkbox after the font size Default unchecked "As Caption".  If checked then it changes the **_vjust_** argument appropriately, so it is at the bottom left.  I assume this will work.  
My reason for this is that I would like to recommend that PICSA graphs usually have 2 discrete captions.  On the bottom left is the definition used in the graph.  And the bottom right - usual caption is the Source: Tanzania Met Authority (TMA).
c) On my machine the up-down controls don't seem to work very well.  I can type into them, but the up-down arrows don't work much.  I haven't checked them elsewhere.

Wycklife:
Fixes #5491


instat/sdgPICSARainfallGraph.vb
```diff
@@ -1500,6 +1500,14 @@ Public Class sdgPICSARainfallGraph
...
     End Sub
 
+    Private Sub ucrInputGraphTitle_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputGraphTitle.ControlValueChanged, ucrInputGraphSubTitle.ControlValueChanged, ucrInputGraphcCaption.ControlValueChanged
+
+    End Sub
+
+    Private Sub ucrInputGraphcCaption_Load(sender As Object, e As EventArgs)
+
+    End Sub
+
```
maxwellfundi: @Wycklife  these sub seems empty. You can get rid of them

Wycklife:
@maxwellfundi I'm removing them now

rdstern:
That's better.  But see the figure below:
image

The label for the x-axis and y-axis titles are now only half visible.
The Return and Help buttons are now visible but what is that half 3rd button (Car?) that is also half-visible as part of that control?

Wycklife:
@rdstern Kindly get the changes a gain. My machine seems to be having a problem with the solution. I have tried it on Ivan's machine and it seems to work well.

Wycklife:
@rdstern I have resolve the designer issue, my visual basic scaling was not correctly set. Now it's working fine. Now you can review this then it will be ready for merge

PR #5721, opened 2020-02-19
Changes in the Prepare menu

### LINKED ISSUE #5615: Changes in the Prepare menu
It is currently like this:
image

I suggest 
a) we move Convert Columns above the line - still just below Replace Values.
b) We get rid of the Column Metadata option
(There could be a reason for keeping it - as I get to writing the help - it does give an easy route to renaming columns, adding labels, and also changing the significant figures displayed.  We don't seem to have another way to do that.  
However, if we do that, then I suggest we move it up, so it is just under Rename Column.)
c) We move **_Column Structure_** down to just above **_Colour by Property_**.  And we get rid of that line below the current Column Structure position.
d) We check on the Protect Column and Freeze Columns.  They are disabled and have been there a long time.  Are we going to do anything soon, or should we either hide them for now, or get rid of them?

Wycklife:
Fixes #5615


Wycklife:
@maxwellfundi This is ready for you

Issue #5730, opened 2020-02-23
Including a single text as a variable


rdstern:
I quite often have wanted to add a single character variable to a data frame.  It is usually climatic data for a single station and I would like to add a station name.  
 
One easy way would be to add a third button to the **_Prepare > Calculate > Regular Sequence_** dialogue.  
I can already use this dialogue to add a single numeric value or a single date, by just making the To field the same as From:  So it is not an unreasonable dialogue to add a single character (text) variable.

1. This third button would say "Text" perhaps.  (Currently there is Numeric and Date.)
2. The 1st field could then say Text:  (instead of From:)
3. The other 3 fields are not needed.
4. The Length would remain as now.

When numeric the current R-code is say:
`Regular3 <- rep(x=seq(from=1, to=1, by=1), each=, length.out=46)  `
With the single text you just need to replace the seq part of the command, so:
`Regular1 <- rep(x="Roger", each=1, length.out=46)  `
or just:
`Regular1 <- rep(x="Roger", length.out=46) `

While adding to that dialogue there are 2 small edits to make.

a) The To: field doesn't remember the previous value.  If set as 1 in the example above, then it resets to the length of the data frame when the dialogue is next opened.
b) If you do 1 to 46 (say) in steps of 1 then the command generated is:
`Regular5 <- rep(x=seq(from=1, to=46, by=1), each=) ` i.e. the value 1 for each= is omitted.  It works, but is messy!

dannyparsons:
Not needed as this functionality exists in the Insert Rows/Columns dialog.

PR #5733, opened 2020-02-24
Minor change

Ivanluv:
@rdstern  I've started looking into  issue #5730 by making the edits you've suggested . I've specifically done the edit for part(B).Would you kindly look at this


rdstern:
Damn - you were too quick.  We have since looked and the facility in b is available in Prepare > Data Frame > Insert Rows/Columns!  I forgot that!

dannyparsons:
Yes, sorry we won't need the additions to the Regular Sequence dialog now as this is available in the Insert Rows/Columns dialog. So I suggest closing this.

Issue #5728, opened 2020-02-22
Add range argument when reading Excel files


rdstern:
I often find it awkward to just have the skip and lines to read with an Excel sheet.  An alternative in readxl is the **_range_** argument.  When it is given the **_rows to skip_** and **_maximum rows to import_**  are ignored, so they could be disabled, but that isn't essential.

We could have different formats in the dialogue.  I suggest a checkbox called Range.  If checked, then There is a **_From:_** with a default of **_A1_** and a **_To:_** with a default of the extent of the sheet if nothing were given.  (There should be space for 2 letters and quite a lot of numbers, e.g. AA1000000 I guess is the maximum?)

dannyparsons:
We could have a small section for "Subset" or something, with two options: range and rows, which give these different choices.

rdstern:
Sounds good.

Ivanluv:
@rdstern is you suggestion adding range as  an argument to the `excel_sheets` command?

shadrackkibet:
We discussed this with @Ivanluv.  The function we are using to read excel files is `rio::import` which internally uses `read_excel` according to the documentation. Using the power of  `...` in the `import` function we can pass the `range` argument into `read_excel` by adding it into `import` function.

lloyddewit:
In PR #5738 @Patowhix reported:

> I can easily break the dialog by just not including the A in the range textbox(which I think any user may omit the A character). Other than the error shown by the dialog, it breaks on reopening. The extent of the sheet is not automatically detected and the default is always AA100.

For more details, please see @Patowhiz 's screendumps, added 15 Aug 2020, in PR #5738.

PR #5738, opened 2020-02-25
Import Files With Multiple Missing Values And Retaining Selected Sheets

### LINKED ISSUE #5700: Minor bugs/features to fix in Open from file dialogue
The import dialog has several bugs/features that need to be fixed and added. 

- [x] 1. The data frame name textbox needs to be pushed down to be consistent with the other dialogs. The textbox also needs to be resized, currently it's very small.

- [ ] 2. Can we have the option to enter more than 1 missing value representation.

- [x] 3. When importing from file. The maximum rows to skip is currently set as 100. This limits the user, we have found files that require us to skip more than 100 rows. The 100 limit needs to be removed.

- [ ]  4. When importing from excel, the selected sheet changes when I reopen the dialog, can the dialog be made to retain the selected sheet?

Patowhiz:
Fixes #5700


rdstern:
I tried the existing version anyway.
[Climatic Data Sadore 1983 to 2014.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/4251118/Climatic.Data.Sadore.1983.to.2014.xlsx)
I used importing of sheet2 on the attached file.  You need to omit the first 4 rows to get the naming of the variables correctly.
a) Still seems to import sheet2 correctly.  But when I return to the dialogue it has not remembered the selected sheet, i.e. it isn't ticked in the list of sheets.  It is still in the preview.  So that control doesn't seem to remember.
b) I tried with **_`0,Janvier`_** (2 missing values) and the preview seems to correspond correctly.  But it doesn't set them as missing on the import.  When either one is the only value, then it seems to work fine.

Patowhiz:
@rdstern you can now test this. I have implemented multiple missing values for excel files only. Once @dannyparsons finds the implementation to be okay I can try to do the same for other file types.

Patowhiz:
@rdstern the first issue of not remembering the sheet name is just a bug due to redundant code that is not so well organised. Best way to fix this is to rewrite most of the dialogs code but for this PR that will make it pending for long. So I have fixed it by doing a small check.

On the second issue, what you are asking for is a prefix I think,and @dannyparsons can confirm this, this is implemented deeper into the `ucrSave` custom control. The problem is the sheet names are different and thus setting the prefix like the way it's done in other dialogs will be suggesting the wrong prefix for the different sheet names. The alternative to this is implementing a custom way of determining the prefix in this dialog, I can do this if @dannyparsons thinks that is a good way forward.

Patowhiz:
On the third issue, the values spring back cause currently we are only hiding things on reset but not clearing the controls. I have fixed this by clearing the excel controls

shadrackkibet:
@Patowhiz what is the state of this? l see conflicts and seems there was some pending work.

maxwellfundi:
@Patowhiz  How is this getting along? Are yous till working on this?

Patowhiz:
Sorry for the late reply. This was ready. but now I have to go through it again, there has been several changes on this dialog by different people. There is also another pending PR by @Ivanluv on the dialog touching functions edited in this PR. Because of this I'll have to wait until those PRs are merged then quickly solve the conflicting files for this to be reviewed

Patowhiz:
@africanmathsinitiative/developers this is now ready for review. Please note this PR fixes multiples issues outlined in issue #5700

lloyddewit:
@Ivanluv Please could you review this? thanks

lilyclements:
@Patowhiz I have a small request of two lines to add into this code:

There is a group box called `grpCsv`. This runs for csv files when a csv file is selected, and runs for txt files when a txt file is selected, and a certain radio button is clicked.
For txt files, the radio buttons are visible and above this group box.
For csv files, the radio buttons are not visible, but the location has not changed. This results in a bit of blank space when a csv file is selected. As a result, it would be good for the location of the group box to change when a CSV file is chosen.

Hence just after line 577, where the code states `ElseIf strFileExt = ".txt" Then`, can you add
`grpCSV.Location = New System.Drawing.Point(9, 99)`
Hence when a text file is selected, it will be below the radio buttons

And after line 584, under `ElseIf strFileExt = ".csv" Then`, can you add
`grpCSV.Location = New System.Drawing.Point(9, 50)`
Hence when a csv file is selected, there will no longer be a blank space at the top of the dialog.

I'm not sure if you have updated this with the new merged design, but these changes should align with that design.

Ivanluv:
The code changes look fine from my review

Ivanluv:
I have however noticed the changes on adding the range functionality,which I added ,seems to have errors that were solved from my end
image

lloyddewit:
@Patowhiz Did you get a chance to look at this? thanks

Patowhiz:
@lloyddewit I added the code you suggested in your above [comment](https://github.com/africanmathsinitiative/R-Instat/pull/5738#issuecomment-656733232) . The groupbox location is now adjusted accordingly.

@Ivanluv this is strange to me. Those controls were added but there is no reference of them in the code. They have not been used anywhere. My understanding is they are supposed to allow entry of range of rows to import but this functionality has not been added. @lloyddewit @rdstern were these controls added through an issue raised?  I can be able to build in this functionality or rather understand more on why they are in the designer but not used. I looked at my past repo branches and I can confirm they were not there. I can also confirm that my current branch has the latest merged code. The name of the controls; `ucrChkRange` , `ucrInputTextFrom` , `ucrInputTextTo` .

Patowhiz:
> I have however noticed the changes on adding the range functionality,which I added ,seems to have errors that were solved from my end
> image

@Ivanluv from your comment above. Do you mean you had added the code for the range? If so then the code is not there. Could you please point me to the PR that added the code. Would help in saving time. I didn't touch these controls and the functionality, I could just paste in the code.  To confirm this, can you try running your latest master branch, does the functionality exist?

lloyddewit:
> 
> 
> @lloyddewit I added the code you suggested in your above [comment](https://github.com/africanmathsinitiative/R-Instat/pull/5738#issuecomment-656733232) . The groupbox location is now adjusted accordingly.

@Patowhiz Did you mean @lilyclements not @lloyddewit ?

instat/dlgImportDataset.vb
```diff
@@ -807,20 +790,24 @@ Public Class dlgImportDataset
...
+            'if there were previously checked items then restore them
+            If lstCheckedItems.Count > 0 Then
+                For i As Integer = 0 To clbSheets.Items.Count - 1
+                    For Each strSelected As String In lstCheckedItems
+                        If strSelected = clbSheets.Items(i).ToString Then
+                            clbSheets.SetItemChecked(i, True)
+                        End If
+                    Next
+                Next
+            End If
```
lloyddewit, line 810: The `clbSheets` loop could be nested inside the `lstCheckedItems` loop. Then we wouldn't need the `If` statement. Would this be more readable?

instat/dlgImportDataset.vb
```diff
@@ -1022,4 +1018,32 @@ Public Class dlgImportDataset
...
+            For Each strTemp As String In arrStr
+                If strRmissingValueString = "" Then
+                    strRmissingValueString = Chr(34) & strTemp.Trim & Chr(34)
+                Else
+                    strRmissingValueString = strRmissingValueString & "," & Chr(34) & strTemp.Trim & Chr(34)
+                End If
+            Next
+            strRmissingValueString = "c(" & strRmissingValueString & ")"
+        End If
+        Return strRmissingValueString
```
lloyddewit, line 1046: We could avoid having to indent the main function code (the code after line 1037), is this more readable (your decision)?
```suggestion
        Dim arrStr() As String = strText.Split(",")

        'if length is < 2 return an R string else return a vector of strings for R
        If arrStr.Length = 0 Then
            Return ""
        End If
        If arrStr.Length = 1 Then
            Return Chr(34) & arrStr(0) & Chr(34)
        End If

        Dim strRmissingValueString As String = ""
        For Each strTemp As String In arrStr
            If strRmissingValueString = "" Then
                strRmissingValueString = Chr(34) & strTemp.Trim & Chr(34)
            Else
                strRmissingValueString = strRmissingValueString & "," & Chr(34) & strTemp.Trim & Chr(34)
            End If
        Next
        strRmissingValueString = "c(" & strRmissingValueString & ")"
        Return strRmissingValueString
```

lloyddewit:
> @Ivanluv from your comment above. Do you mean you had added the code for the range? If so then the code is not there. Could you please point me to the PR that added the code. Would help in saving time. I didn't touch these controls and the functionality, I could just paste in the code. To confirm this, can you try running your latest master branch, does the functionality exist?

@Ivanluv I'll wait for your response before I merge

Ivanluv:
> @Ivanluv from your comment above. Do you mean you had added the code for the range? If so then the code is not there. Could you please point me to the PR that added the code. Would help in saving time. I didn't touch these controls and the functionality, I could just paste in the code. To confirm this, can you try running your latest master branch, does the functionality exist?

@Patowhiz  this was solved by PR #5772.

Patowhiz:
@Ivanluv thank you for the feedback. For some reason these changes are no longer there even in the master(even after synchronising my master). @lloyddewit @Ivanluv  can you confirm that you have the changes in your master branches? If not should I include the code that was added by @Ivanluv in this PR or he does it on another PR? I prefer @Ivanluv taking over since this is something he already did but I'm also happy to quickly add it. The screenshot below is from my master which has the latest changes.

import file

Ivanluv:
@Patowhiz  I also dont have the changes in my master.I suggest you add the code

instat/dlgImportDataset.vb
Patowhiz: @lloyddewit this is a good alternative suggestion. However, in this case of multiple if statements, when it comes to functions (methods that return a value) I think it's more readable and easier to debug if the value returned is from a particular variable. The developer can easily track logically how the single variable is updated inside the body of the function by just highlighting it. I find this easier than having multiple return statements that are executed in different if statements in the body of the function.

lloyddewit:
@Patowhiz I agree with @Ivanluv . If the code is no longer in the master then the safest thing to do is just add the code again to this PR.
If I misunderstood (e.g. you could use help looking for something) then let me know.

instat/dlgImportDataset.vb
lloyddewit: @Patowhiz It's true that if you highlight the return variable then you can see how it's updated in the function. However you can also double-click on one of the returns and all the returns in the function are highlighted.
Immediately before R-Instat I programmed in Golang. the Golang idiom is 'happy path stays left' (see https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88 for the rationale).
I like this idiom and feel it makes the code more readable, but I respect your decision if you leave the code as it is.

Patowhiz: @lloyddewit  thank you for the reference article, it has good points, I'm happy with either way as a standard of coding. I've committed your suggestion with a slight modification.

The modification is, in this case the array length will never be 0. So checking for length = 0 is redundant and the code inside the block will never be executed.

lloyddewit:
@Patowhiz Thank you for the changes. Please could you accept/reject the last remaining review comment? I can then merge, thanks

Patowhiz:
@lloyddewit will add the code and have it ready for review by end of today.

Patowhiz:
> I often find it awkward to just have the skip and lines to read with an Excel sheet. An alternative in readxl is the **_range_** argument. When it is given the **_rows to skip_** and **_maximum rows to import_** are ignored, so they could be disabled, but that isn't essential.
> 
> We could have different formats in the dialogue. I suggest a checkbox called Range. If checked, then There is a **_From:_** with a default of **_A1_** and a **_To:_** with a default of the extent of the sheet if nothing were given. (There should be space for 2 letters and quite a lot of numbers, e.g. AA1000000 I guess is the maximum?)

@rdstern above is your statement from the issue #5728 that was fixed by @Ivanluv in PR #5772.
I have added the code that @Ivanluv had implemented in the PR. But then I noticed 2 things that probably both of you are aware.

1. I can easily break the dialog by just not including the **A** in the range textbox(which I think any user may omit the **A** character). Other than the error shown by the dialog, it breaks on reopening. 
2. The extent of the sheet is not automatically detected and the default is always **AA100**.

@lloyddewit in relation to what was already implemented this PR is now ready. If @rdstern @Ivanluv would like my above concerns to be solved, then I suggest reopening issue #5728 and doing a PR that will fix them. 

Below is an image that demonstrates how I was able to find the bug.

import bug

import bug 2

lloyddewit:
@Patowhiz Thanks for the updates. There is still an unresolved comment from 08 Aug (see above). If you can resolve this then I can approve, thanks

@rdstern @Ivanluv I reopened issue #5728 (see @Patowhiz 's comments from 15 Aug above).

lloyddewit:
@Patowhiz There is just one small comment open (from 08 Aug, see below). Please could you accept/reject the suggestion? I can then approve, thanks

image

instat/dlgImportDataset.vb
Patowhiz: @lloyddewit sorry I missed this. I agree, there are fundamental merits in eliminating the `if`.  
I've now modified the code, included your changes and made it better.

Patowhiz:
@lloyddewit I accept your suggestion, you can have a look at my modified code

lloyddewit (approved):

lloyddewit:
@rdstern Please could you approve? thanks

rdstern (approved):
Great

PR #5740, opened 2020-02-25
Added first dialog (Glance) of the broom package to Use model dialog

### LINKED ISSUE #5647: Add broom to the Model > Use Model menu
broom::augment(last_model)
broom::glance(last_model)
broom::tidy(last_model)

Wycklife:
Fixes #5647


Wycklife:
@rdstern @lloyddewit To test this, you create a new branch, then get changes from these pull requests, #5740 and #5741 . These changes are for one dialog(Glance).

rdstern:
@Wycklife I don't quite understand.  I assume the programmer - namely you - would always do the initial testing and hence you would set up this branch - then tell us when you are ready.  It seems odd that I and (separately @lloyddewit) would set this up?  What am I missing here?

Wycklife:
@rdstern Danny advised me not to do the formmain changes on the same branch, these changes are related. Create Branch you would like to test this, get these changes to the branch one by one. I can as well combine these changes into one branch but I have to ask @dannyparsons and @maxwellfundi .

rdstern:
I agree with that process, but the current version (0.6.1) has already merged the changes in frmmain - as far as I can see.  So perhaps this caution is no longer needed and the later branches can simply update from the master?  (Though I am not sure I have fully understood what is going on!)

Wycklife:
@rdstern These branches are updated. So I don't think if there can be conflicts. I keep on updating them.

Wycklife:
@rdstern What I have done with @Ivanluv is that I created a branch called test from the updated master in Ivan's machine. Then I have two pull requests related to the Glance dialog, I get the two changes one by one into my test branch. Merged and pulled the changes. Then I can see the dialog. 
On the remaining two dialogs I created  another branch called Test2, since they have different formmain changes as for the Glance. I then have 3 pull request for these two dialogs. I merged them one by one, then I pulled the changes. Kindly try this. I have as well inserted screen shots for the dialogs on the first pull request (#5754) You can as well have a look at them

shadrackkibet:
@Wycklife Frmmain changes (new dialogs) should be done on a separate branch and merged as soon as possible.  As soon as that is done then you can update your branch with the master and then proceed with any further changes. This will make it easy for anybody testing your work rather than having two pull requests.

Wycklife:
@shadrackkibet That's making a lot of sense. I have talked to @maxwellfundi and he is going to merge them

rdstern:
This works, as far as it goes.  There are 2 next steps to undertake.  It needs to write to a new data frame.  The controls are there, but ignored by the command.

And it needs to be able to do this for multiple models.  Then I also suggest the dialogue could usefully move to the Compare Models section

Wycklife:
@rdstern This is now working for multiple models. @dannyparsons The todo, in the code is left for you to discuss with @volloholic

instat/dlgGlance.vb
```diff
@@ -1,3 +1,116 @@
...
+        ucrModelReceiver.ResetText()
+        ucrModelSelector.Reset()
+
+        clsGlance.SetRCommand("glance")
+        clsGlance.SetPackageName("broom")
+
+        'todo implement as a function properly
+        clsMap_df.SetRCommand("map_df")
+        clsMap_df.SetPackageName("purrr")
+        clsMap_df.AddParameter(strParameterName:=".f", strParameterValue:="broom::glance", iPosition:=1)
```
shadrackkibet: @dannyparsons I suggest we have an option not to include `()` when setting an R command rather than having to add this manually `broom::glance`

Wycklife:
@rdstern These are fixed. Could you please check.

Wycklife:
@rdstern Done!

PR #5743, opened 2020-02-26
Minor change to ucrVariableAsFactor

### LINKED ISSUE #5707: PICSA Rainfall Graph Dialog
When you select from the receiver from single Variable to multiple variable, the selector shows date columns only instead of numeric variables.

Ivanluv:
@africanmathsinitiative/developers  this is ready for review .Fixes #5707

### DIFF
--- a/instat/ucrVariablesAsFactor.vb
+++ b/instat/ucrVariablesAsFactor.vb
@@ -315,7 +315,7 @@ Public Class ucrVariablesAsFactor
 
     Public Overrides Sub SetExcludedDataTypes(strExclude As String())
         ucrSingleVariable.SetExcludedDataTypes(strExclude)
-        ucrMultipleVariables.SetIncludedDataTypes(strExclude)
+        ucrMultipleVariables.SetExcludedDataTypes(strExclude)
     End Sub
 
     Public Overrides Sub SetSelectorHeading(strNewHeading As String)


lloyddewit:
@Ivanluv Thank you for the change. I reviewed and tested it. It seems to fix #5707.
There is still something I don't understand. 
`dlgPICSARainfall.InitialiseDialog()` contains the following lines:
```
        ucrVariablesAsFactorForPicsa.SetIncludedDataTypes({"numeric"})
        ucrVariablesAsFactorForPicsa.SetExcludedDataTypes({"Date"})
```
I assume this means that the selectors should show numeric data excluding dates. However the selectors also show non-numeric data (e.g. station names). Is this correct?
Also, do you know if there's any reason for having 'Date' capitalised and 'numeric' not?

Patowhiz:
"Also, do you know if there's any reason for having 'Date' capitalised and 'numeric' not?"

This ties to an issue which I created of using enumerations instead of hard coded string literals in R-Instat code.

Ivanluv:
@lloyddewit   the two functions  `SetIncludedDataTypes({"numeric"})` and `SetExcludedDataTypes({"Date"})` seem to be editing the same list of excluded Metadata properties ie `lstExcludedMetadataProperties` thats why its only showing non numeric data types.This is because after  this `  ucrVariablesAsFactorForPicsa.SetExcludedDataTypes({"Date"})` is run the list only contains date datatype to exclude

lloyddewit:
@Patowhiz I agree. Using hard-coded string literals increase the risk of bugs. It would be safer to use something like:
```
Enum IncludeType
    numeric
    date
End Enum
```
We can address this if/when we refactor these classes.

@Ivanluv Thank you for the comment. I investigated a bit more and I think there is still a bug that needs to be fixed before this issue can be closed. Details below. Please could you verify my analysis below? If you'd like to discuss then I'd be very happy to have a WhatsApp or Skype chat.

----

The ultimate purpose of `ucrVariablesAsFactor.SetIncludedDataTypes()` and `ucrVariablesAsFactor.SetExcludedDataTypes()` is to populate the following 'ucrReceiver' data fields:
```
Public lstIncludedMetadataProperties As List(Of KeyValuePair(Of String, String()))
Public lstExcludedMetadataProperties As List(Of KeyValuePair(Of String, String()))
```
These lists are used by `clsRLink.FillListView()` to populate the controls.
`dlgPICSARainfall` calls:
```
ucrVariablesAsFactorForPicsa.SetIncludedDataTypes({"numeric"})
```
This sets `lstExcludedMetadataProperties ` to:
```
"class", {"factor", "character", "ordered,factor"}
```
I think this is correct and should mean that only numeric (e.g. not station names) should be listed.
However, the problem comes when `dlgPICSARainfall` calls:
```
ucrVariablesAsFactorForPicsa.SetExcludedDataTypes({"Date"})
```
This ***overwrites*** `lstExcludedMetadataProperties` to contain only:
```
"class", {"Date"}
```
I think the bug is in `ucrReceiver` line 339:
```
        iExcludeIndex = lstExcludedMetadataProperties.FindIndex(Function(x) x.Key = strProperty)
        If iExcludeIndex <> -1 Then
            lstExcludedMetadataProperties(iExcludeIndex) = kvpExcludeProperty
```
This code checks if the key (in our case "class") already exists in the exclude list. If so, it overwrites the whole entry. 
The solution for `dlgPICSARainfall` would be to append "Date" to the existing list of excluded fields (rather than overwriting the list). However we first need to clarify if other dialogs rely on the existing behaviour of `ucrReceiver`.

lloyddewit:
@Ivanluv @maxwellfundi @rdstern 
I think there could still be a bug in this PR (see my comment from 02/03 above). There are no commits after 02/03. Was the bug fixed in another PR or does it still exist?

Ivanluv:
@lloyddewit  Ive found out that while using the `setIncludeDataType()` with the `bStrict` argument being true ie  `  ucrVariablesAsFactorForPicsa.SetIncludedDataTypes({"numeric"}, bStrict:=True)` ,the non numeric variables like station are not included in the selector

lloyddewit:
@Ivanluv Thank you, I tested your proposal and it seems to work well. I changed `dlgPICSARainfall.InitialiseDialog` line 127 to:
```
ucrVariablesAsFactorForPicsa.SetIncludedDataTypes({"numeric"}, True)
```
and I **deleted** line 128:
```
ucrVariablesAsFactorForPicsa.SetExcludedDataTypes({"Date"})
```
Line 128 is no longer needed because date is not a numeric so is anyway excluded by line 127.
Will you make these changes? 

P.S. Here's a screenshot from my test. You can see that the numeric columns are listed but the date and text columns are included. This is correct.

image

rdstern:
I am confused by the result of this change.  The initial problem in #5707 was when the Multiple Plot, not the Single Plot control as displayed displayed above.  
The y-variable in the single display should only be displaying numeric variables, not date, etc. 
The original problem was that with the Multiple option, then only date variables were provided whereas what's needed is only numeric.

lloyddewit:
@rdstern what you describe is another symptom of the same bug. This PR fixed this specific symptom but not the complete bug. This PR ensured that the selector listed all the numeric fields correctly, and excluded date columns, but it did not exclude non-numeric columns like month name. The change proposed above fixes this.

lloyddewit:
@Patowhiz Thank you for the clear analysis, I think it's really good to document quality issues we spot while we're working on the code. I agree that a 'set' function should only do what the developer expects from the function's name. 
Please could you raise an issue for this (you could reuse the text above)?

PR #5749, opened 2020-02-26
Adding the stat_cor and the stat_regline_equation

### LINKED ISSUE #5744: Add 2 more stats to the list of stats and geoms for ggplot2
I assume this may be relatively easy for either Ivan or Wyclife to add?
Here is an example of a graph with summaries added.  This is something we have needed for a long time.

image

This uses the ggpubr package that is already installed.  It uses `stat_cor` for the correlation and `stat_regline_equation` for the regression equation. 

They can now be added simply to the script for a scatteplot (or line plot) as I did below, to get the graph above.

`last_graph <- ggplot2::ggplot(data=Sadore, mapping=ggplot2::aes(x=Tmx, y=Tmn)) + ggplot2::geom_point() + theme_grey()+ggpubr::stat_cor()+ggpubr::stat_regline_equation(label.y=30) `
We already have 4 stats in the list when layers are added to a ggplot, so this is adding 2 more.

a) For future reference I note that the documentation for the regression line equations says it was inspired by a function in the ggpmisc package (that is also installed). Later there will be other stats and geoms to add from there, particularly in relation to the broom package.  But for now these 2 stats should be simple to add and would be very useful.
b) In the figure above it should be possible to adapt the displays, once the stats have been implemented. In particular I would like to be able to give the correlation without the significance level.  And, with the regression line it should be possible to add further summaries - but I failed, when adding them in the script file.  This can be investigated once the stats are installed.

Ivanluv:
@africanmathsinitiative/developers  this is ready for review.Fixes #5744

### DIFF
--- a/instat/ucrGeom.vb
+++ b/instat/ucrGeom.vb
@@ -166,6 +166,31 @@ Public Class ucrGeom
         Dim clsgeom_violin As New Geoms
         Dim clsgeom_vline As New Geoms
 
+        Dim clsgeom_statcor As New Geoms
+
+        clsgeom_statcor.SetGeomPackage("ggpubr")
+        clsgeom_statcor.SetGeomName("stat_cor")
+
+        'add aes parameter
+        clsgeom_statcor.AddAesParameter("x", strIncludedDataTypes:={"factor", "numeric"})
+        clsgeom_statcor.AddAesParameter("y", strIncludedDataTypes:={"factor", "numeric"})
+
+        'add layer parameter
+        clsgeom_statcor.AddLayerParameter("geom", "list", Chr(34) & "point" & Chr(34), lstParameterStrings:={Chr(34) & "point" & Chr(34), Chr(34) & "text" & Chr(34), Chr(34) & "label" & Chr(34), Chr(34) & "line" & Chr(34), Chr(34) & "rug" & Chr(34), Chr(34) & "hline" & Chr(34), Chr(34) & "vline" & Chr(34)})
+        clsgeom_statcor.AddLayerParameter("span", "list", Chr(34) & "5" & Chr(34), lstParameterStrings:={Chr(34) & "5" & Chr(34), Chr(34) & "NULL" & Chr(34)})
+        clsgeom_statcor.AddLayerParameter("ignore_threshold", "numeric", "0", lstParameterStrings:={1, 0, 1})
+        clsgeom_statcor.AddLayerParameter("strict", "boolean", "FALSE")
+        clsgeom_statcor.AddLayerParameter("label.fmt", "editablelist", Chr(34) & "%.4g" & Chr(34), lstParameterStrings:={Chr(34) & "%.4g" & Chr(34)})
+        clsgeom_statcor.AddLayerParameter("x.label.fmt", "editablelist", Chr(34) & "NULL" & Chr(34), lstParameterStrings:={Chr(34) & "NULL" & Chr(34)})
+        clsgeom_statcor.AddLayerParameter("y.label.fmt", "editablelist", Chr(34) & "label.fmt" & Chr(34), lstParameterStrings:={Chr(34) & "label.fmt" & Chr(34)})
+        clsgeom_statcor.AddLayerParameter("position", "editablelist", Chr(34) & "identity" & Chr(34), lstParameterStrings:={Chr(34) & "identity" & Chr(34)})
+        clsgeom_statcor.AddLayerParameter("na.rm", "boolean", "FALSE")
+        clsgeom_statcor.AddLayerParameter("show.legend", "list", "FALSE", lstParameterStrings:={"NA", "TRUE", "FALSE"})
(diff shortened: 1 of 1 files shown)


Ivanluv:
@rdstern  am not sure on how to implement your suggestion on label.x,part (b).

rdstern:
For b) I was hoping to find another example, but can't
The idea is that in a graph of y against x this is the x-value where the text for the correlation will start.  Now, the x-scale could be small, e.g. 0 to 0.5 or large, e.g. 1900 to 2020.  So it is tricky to have an up-down control for that value.  Just have a field where you can type a number.  

So when completed it could look like this

( tick) x-value   (1930.  Then this would add `value.x = 1930` to the command.

By the way - start with the correlation stat.  @dannyparsons is looking at the regression one and it may change?

instat/ucrGeom.vb
```diff
@@ -182,6 +205,28 @@ Public Class ucrGeom
...
+        clsgeom_statreglineequation.AddLayerParameter("position", "list", Chr(34) & "identity" & Chr(34), lstParameterStrings:={Chr(34) & "stack" & Chr(34), Chr(34) & "dodge" & Chr(34), Chr(34) & "dodge2" & Chr(34), Chr(34) & "identity" & Chr(34), Chr(34) & "jitter" & Chr(34), Chr(34) & "fill" & Chr(34)})
+        clsgeom_statreglineequation.AddLayerParameter("label.y.npc", "editablelist", Chr(34) & "left" & Chr(34), lstParameterStrings:={Chr(34) & "left" & Chr(34), Chr(34) & "center" & Chr(34), Chr(34) & "right" & Chr(34)})
+        clsgeom_statreglineequation.AddLayerParameter("colour", "colour", Chr(34) & "black" & Chr(34))
+        clsgeom_statreglineequation.AddLayerParameter("size", "numeric", "5", lstParameterStrings:={0, 0})
+        clsgeom_statreglineequation.AddLayerParameter("alpha", "numeric", "1", lstParameterStrings:={2, 0, 1})
+        clsgeom_statreglineequation.AddLayerParameter("label.x", "editablelist", Chr(34) & "NULL" & Chr(34), lstParameterStrings:={Chr(34) & "NULL" & Chr(34)})
+        clsgeom_statreglineequation.AddLayerParameter("label.y", "editablelist", Chr(34) & "NULL" & Chr(34), lstParameterStrings:={Chr(34) & "NULL" & Chr(34)})
```
lloyddewit: Are these parameter's identical to the `clsgeom_statcor` parameters?
If so, could they be put in a private function to avoid duplication?

Ivanluv: Am not sure if they are more parameters to add for these geoms so I dont know if they are identical

shadrackkibet:
What is the status of this? looks like there is some good progress according to @rdstern report above.

Ivanluv:
@more discussion is required on this

lloyddewit:
please let me take another look at this before this is merged, thanks

lloyddewit:
@Ivanluv Is there something blocking this, or have you just been too busy? thanks

lloyddewit:
@Ivanluv Please could you clarify the status? thanks

Ivanluv:
@rdstern suggested that this issue requires more discussion since its complex than he initially thought .

lloyddewit:
@rdstern It would be good to close this old PR. Do you have any ideas how we could get this PR into a state where we could merge it (and then raise an issue for anything outstanding)? thanks

rdstern (commented):
This is a feature I would love to see added.  I hope @dannyparsons could take over the advice needed to add these facilities.  Otherwise we should leave this pull request for now.

dannyparsons:
Was the discussion on whether we want to be using things from `ggpubr` or not?

maxwellfundi:
@Ivanluv @rdstern  i have opened a PR on the list of geoms #6324 that has the geoms liste in the issue #6293 . For these two, `stat_cor` and the `stat_regline_equation` i will add them maybe in same PR.
So quick question @rdstern  was the issue raised by @dannyparsons agreed on yet? 

Again, is whatever is on here reasonable to be merged @rdstern  then open issues for specific problems as @lloyddewit is suggesting?

dannyparsons:
I think the priority is to get the core geoms from ggplot2 added first. I think those from other packages are lower priority.

lloyddewit:
@dannyparsons @rdstern @maxwellfundi I read through the above comments again but I don't have the knowledge to understand what the 'core geoms' are and whether they have been (partially)implemented in this PR. I'm still not clear whether we need to:
- merge this PR as is
- update and merge this PR
- close this PR
- do any of the above and raise extra issues

Sorry for my confusion! Any guidance?

dannyparsons:
I hadn't looked at the changes before commenting, sorry. Happy for this to be merged if it's working. I remember Roger having doubts about `ggpubr` after looking into it further, but I don't know if that was more generally or for these geoms specifically.

rdstern:
I suggest with all the other things we do, that we close this one for now

lloyddewit:
Closed as recommended by @rdstern

Issue #5618, opened 2019-12-11
Improve the Prepare > Data Frame > Delete Columns/Rows dialogue


rdstern:
Here is a picture of the dialogue:

![image](https://user-images.githubusercontent.com/11226469/70644118-446e9480-1c3a-11ea-8aaf-c459e6f96965.png)

1. Small change could the controls with the row numbers become wider, so they can show row numbers up to 10 million at least.  The example here has 1.3 million rows.
2. More important is what the dialogue should do when a filter is in operation.  This needs discussion, because there are options.  Currently it ignores the filter and just deletes the row numbers you specify, whether visible or not.
One neat thing would be if a filter is on, then it changes and says: 
o  Delete all filtered rows (the rows you don't see, even with the R viewer))
o  Delete the unfiltered rows (those that remain visible)

Patowhiz:
@rdstern was this fixed? I can see @Wycklife mentioned this in his PR #5750. If it was fixed then I suggest closing the issue

rdstern:
@Patowhiz happy for you to do this.

From a quick check it looks as though the simple first issue is fixed - the controls are wider.  But it still ignores the filter.  I tried with the (usual) survey data from the library and filtered so the first visible row was case 11.  It still deleted the first 2 rows in the data frame, even though they were not visible.

I am not sure what should be done.  It is logical that it deletes row numbers 1-2 if that is what I ask for.  I can delete the first 2 **_visible_** rows by marking them and then using right-click on the left-hand side.  

So I think I would be happy if the dialogue didn't work when a filter is in operation.  What would be the best way to indicate this?

Patowhiz:
@rdstern I like your suggestion in item (2). Currently, there is no clear way of letting the user know what he/she is about to delete. We could have the 2 options when a filter is detected. Should I go ahead and do it that way? Of course the easier and quicker alternative is to disable the dialog when filter is detected as you have suggested above. I'm happy to do either.

lilyclements:
@rdstern can we revisit this? Perhaps just disable the "Delete Rows/Columns" dialog when a filter is applied?

PR #5750, opened 2020-02-26
Improve the Prepare > Data Frame > Delete Columns/Rows dialogue 

Wycklife:
Fixes  part one of #5618

### DIFF
--- a/instat/ucrNud.vb
+++ b/instat/ucrNud.vb
@@ -146,12 +146,16 @@ Public Class ucrNud
     End Function
 
     Public Function GetText() As String
-        Return CStr(Value)
+        If nudUpDown.Text = "" Then
+            Return ""
+        Else
+            Return CStr(Value)
+        End If
     End Function
 
     Public Overrides Sub SetToValue(objTemp As Object)
         Dim dNewValue As Decimal
-        If objTemp Is Nothing Then
+        If objTemp Is Nothing OrElse objTemp.ToString() = "" Then
             'If no value reset to a default value
             Value = nudUpDown.Minimum
         Else
--- a/instat/dlgDeleteRowsOrColums.Designer.vb
+++ b/instat/dlgDeleteRowsOrColums.Designer.vb
@@ -131,6 +131,7 @@ Partial Class dlgDeleteRowsOrColums
         '
         'ucrSelectorForDeleteColumns
         '
+        Me.ucrSelectorForDeleteColumns.bDropUnusedFilterLevels = False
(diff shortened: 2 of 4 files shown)


lloyddewit:
I understand part one of #5618 requests that 'the controls with the row numbers become wider, so they can show row numbers up to 10 million at least'.
I see 3 row number controls ('number of rows', 'delete from' and 'to'). I understand that all 3 of these controls should be widened to show at least 8 digits.
@rdstern did I understand correct?
@Wycklife In the 'files changed' section I see that the 'number of rows' control has been widened but not the other 2 controls. Do the other 2 controls also need widening?

Wycklife:
@lloyddewit  Thank you, I misunderstood this. I will work on the remaining two controls.

rdstern:
@Wycklife this is an "interesting dialogue!
It is much better now.  But I took the (usual) survey data, then deleted rows 34 to 36.
Then - when I returned to the dialogue it opened fine - it used to give an error first.  Then the From still said 34 and the To (correctly) said 33.  There were now 33 rows - so pressing ok with the 34 - -which I did (of course) gave an error.
Also when I delete the From or To in the controls the OK is still enabled.  

So I tried to see what happens with deleting variables.  (And found another error, but it is close there.)
I deleted the last variable.  Then returned and it gave an error.  Then the receiver was empty - which is fine.  So - we need to trap that error.

I deleted another variable and it was fine - when I returned the control was empty.  I am happy with that.

So, back to the rows.  Can the up-down start empty?  If so, then I think it should.  And ok needs both to have a number in to be enabled.
So, if it ever becomes empty, then you disable ok.
Also =, it doesn't remember anything.  When you return to the dialogue the From and To are empty again.  Is that easy to do?

Wycklife:
@rdstern This is done and now working well. Kindly confirm.

instat/ucrNud.vb
```diff
@@ -145,13 +145,17 @@ Public Class ucrNud
         Return bTemp
     End Function
 
-    Public Function GetText() As String
-        Return CStr(Value)
+    Public Function GetText() As String
+        If nudUpDown.Text = "" Then
+            Return ""
+        Else
+            Return CStr(Value)
```
Patowhiz, line 152: This is trivial, but for future reference I'm mentioning it here. This is unnecessary cast. The `Value` which is a `Decimal` has a concrete implementation of a `ToString()` function which returns the `Decimal` as a `String`. This function is called implicitly by the compiler.

PR #5752, opened 2020-03-02
Added the augment broom dialog use models

### LINKED ISSUE #5647: Add broom to the Model > Use Model menu

Wycklife:
Fixes #5647
This is working for a single model.

### DIFF
--- a/instat/dlgAugment.vb
+++ b/instat/dlgAugment.vb
@@ -48,12 +48,11 @@ Public Class dlgAugment
         ucrModelSelector.SetParameter(New RParameter("data", 0))
         ucrModelSelector.SetParameterIsrfunction()
 
-
         ucrSaveNewDataFrame.SetIsComboBox()
         ucrSaveNewDataFrame.SetSaveTypeAsDataFrame()
+        ucrSaveNewDataFrame.SetCheckBoxText("Save New data frame")
+        ucrSaveNewDataFrame.SetPrefix("Augment_data")
         ucrSaveNewDataFrame.SetDataFrameSelector(ucrModelSelector.ucrAvailableDataFrames)
-        ucrSaveNewDataFrame.SetPrefix("Newdataframe")
-
 
         ucrModelReceiver.SetParameter(New RParameter("x", 0))
         ucrModelReceiver.Selector = ucrModelSelector
@@ -79,12 +78,13 @@ Public Class dlgAugment
 
     Private Sub TestOKEnabled()
         'Tests when ok can be enabled
-        If ucrModelReceiver.IsEmpty AndAlso ucrSaveNewDataFrame.IsComplete Then
-            ucrBase.OKEnabled(False)
-        Else
+        If (ucrSaveNewDataFrame.IsComplete AndAlso Not ucrModelReceiver.IsEmpty() AndAlso ucrChkDisplayInOutput.Checked) Then
             ucrBase.OKEnabled(True)
+        Else
+            ucrBase.OKEnabled(False)
         End If
     End Sub
+
 
     Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
         SetDefaults()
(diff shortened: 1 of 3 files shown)


rdstern:
This dialogue starts to implement the broom augment function.  
@volloholic may be able to confirm, but I think this receiver should change to be a single receiver.  It is producing mode variables at the original data frame level.  

Currently the output is to the results window only.  It should now also be able to write back to the original data frame.

This is unlike the other 2 dialogues that will write to a new data frame.  This dialogue will add variables to the existing data frame.  There is also a question of whether the dialogue is in the right place - perhaps it should be in the Fit model section.  Perhaps it should even be "behind the scenes" and become part of the saving results from the fit-model dialogues.

So, leave this for now, until we resolve these ideas.

Wycklife:
@rdstern Could you please check if this is okay.

Wycklife:
@rdstern .Hi Roger, @dannyparsons suggested that I should leave the two check boxes in the Augment dialog for now, since the the dialog is only adding extra columns to the existing data frame.

instat/dlgAugment.vb
```diff
@@ -1,3 +1,91 @@
...
+        SetRCodeForControls(True)
+        TestOKEnabled()
+    End Sub
+
+    Private Sub CoreControls_ControlContentsChanged() Handles ucrModelReceiver.ControlContentsChanged ', ucrSaveNewDataFrame.ControlContentsChanged
+        TestOKEnabled()
+    End Sub
+
+    Private Sub UpdateAssignTo()
+        ucrBase.clsRsyntax.SetAssignTo(strAssignToName:="augment_data", strTempDataframe:=ucrModelSelector.ucrAvailableDataFrames.cboAvailableDataFrames.Text, bAssignToColumnWithoutNames:=True)
```
shadrackkibet: I think this can be done on SetDefaults()

instat/dlgAugment.vb
```diff
@@ -1,3 +1,91 @@
...
+        clsAugment.SetRCommand("augment")
+        clsAugment.SetPackageName("broom")
+        ucrBase.clsRsyntax.SetBaseRFunction(clsAugment)
+        UpdateAssignTo()
+    End Sub
+
+    Public Sub SetRCodeForControls(bReset As Boolean)
+        'ucrSaveNewDataFrame.SetRCode(ucrBase.clsRsyntax.clsBaseCommandString, bReset)
+        ucrModelReceiver.SetRCode(clsAugment, bReset)
+        UpdateAssignTo()
```
shadrackkibet: Since you have one RFunction then this could be simplified.

instat/dlgAugment.vb
```diff
@@ -1,3 +1,91 @@
-﻿Public Class dlgAugment
+﻿
+Imports instat.Translations
+Public Class dlgAugment
+    Public bfirstload As Boolean = True
+    Public bReset As Boolean = True
+    Private bResetSubdialog As Boolean = False
+    Private clsRSyntax As RSyntax
```
shadrackkibet: I don't think this is needed.

instat/dlgAugment.vb
```diff
@@ -1,3 +1,91 @@
...
+
+        ucrModelReceiver.SetParameter(New RParameter("x", 0))
+        ucrModelReceiver.Selector = ucrModelSelector
+        ucrModelReceiver.SetMeAsReceiver()
+
+
+    End Sub
+
+    Private Sub SetDefaults()
+
```
shadrackkibet: You should declare New instance of your RFunction here.

Wycklife:
@rdstern This is fixed too

rdstern:
Just the trivial changes for consistency with the other dialogues, namely Add to Data Frame and New Data Frame Name.  Then please merge

Wycklife:
@rdstern Done!

PR #5753, opened 2020-03-02
Added the tidy broom dialog to use model

### LINKED ISSUE #5647: Add broom to the Model > Use Model menu

Wycklife:
Fixes #5647

This is working for a single model for now!


rdstern:
This works, as far as it goes.  There are 2 next steps to undertake.  It needs to write to a new data frame.  The controls are there, but ignored by the command.

And it needs to be able to do this for multiple models.  Then I also suggest the dialogue could usefully move to the Compare Models section

Wycklife:
@rdstern The changes are noted! But for the implementation for the data frame and multiple models, I will need @volloholic help.

Wycklife:
@rdstern This is ready for you.

Wycklife:
@rdstern  Done!

PR #5755, opened 2020-03-03
Script window

### LINKED ISSUE #5742: Issue multiline commands in script window?
Currently R-Instat commands in the script window are sent to R each line at a time.  I often have to collect lines from simple files for this.  Here is an example:

```
# generate artificial data
set.seed(4321)
x <- 1:100
y <- (x + x^2 + x^3) + rnorm(length(x), mean = 0, sd = mean(x^3) / 4)
my.data <- data.frame(x = x, y = y,
group = c("A", "B"),
y2 = y * c(0.5,2),
w = sqrt(x))

# give a name to a formula
formula <- y ~ poly(x, 3, raw = TRUE)
# no weights
ggplot2::ggplot(my.data, aes(x, y)) +
geom_point() +
geom_smooth(method = "lm", formula = formula) +
ggpmisc::stat_poly_eq(formula = formula, parse = TRUE)
```
This works fine as long as I combine the 4 lines from my.data into a single line and then also the 4 lines for the ggplot2 graph.
I often find a very long line for a ggplot2 command and it would be highly convenient and more readable if they could be split over multiple lines.

Would it be possible for certain ends of line to be recognised as only part of a command.  The 2 above are the most obvious, i.e. + and comma.  It would also be good to be able to define a function and so detecting when { is open and waiting for the closure would also be good.

Might that be all?  Perhaps @dannyparsons can comment?  equals is another possibility.  I am not keen on checking on whether the brackets match, because there can often be mistakes there and so I would prefer that be checked on a single line.

Patowhiz:
Fixes #5742.
This is not yet ready for review. Still working on the reconstructing the text passed on the  window

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -39,6 +39,8 @@ Public Class ucrScript
     Private Sub RunAllText()
         If txtScript.TextLength > 0 Then
             If MsgBox("Are you sure you want to run the entire contents of the script window?", MessageBoxButtons.YesNo, "Run All") = MsgBoxResult.Yes Then
+                'todo. check for symbols + , and %>% and reconstruct the text
+
                 RunText(txtScript.Text)
             End If
         End If


lloyddewit:
I think that this PR has been superceded by PR #5769 and can be closed.

Patowhiz:
Yes, thank you.

PR #5760, opened 2020-03-09
Added a simple keyboard to the filter dialog

### LINKED ISSUE #5731: Add a simple keyboard to the filter sub-dialogue
The need for this should be checked by @dannyparsons.  
We have made some small improvements to the filter sub-dialogue.
It now is different for factors, dates and logical variables.
For numeric variables I suggest we add the same simple keyboard as for the calculator now.  
We need most of the numeric keys, and it does work with simple calculations, e.g. val > (3*2) so it does no harm to have exactly the same calculator.  I hope that might be very simple to add?

And, while on this dialogue, when a factor is selected in the filter, the dialogue gives a choice of levels to select.  Initially the first one is selected - and that is a pain.  So could no levels be selected initially.

Wycklife:
Fixes #5731

### DIFF
--- a/instat/sdgCreateFilter.Designer.vb
+++ b/instat/sdgCreateFilter.Designer.vb
@@ -19,7 +19,7 @@ Partial Class sdgCreateFilter
     Inherits System.Windows.Forms.Form
 
     'Form overrides dispose to clean up the component list.
-    <System.Diagnostics.DebuggerNonUserCode()> _
+    <System.Diagnostics.DebuggerNonUserCode()>
     Protected Overrides Sub Dispose(ByVal disposing As Boolean)
         Try
             If disposing AndAlso components IsNot Nothing Then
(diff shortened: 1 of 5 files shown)


rdstern:
@wyclife I have started the review.  I describe what I did in detail, because I would like you to be doing some of these analyses.
a) Opened the survey data from the Intro guide in the library.  It is a good (small) one to have in testing.
b) Decided to filter to yield > 40.  So went to New Filter, used yield variable and got your keyboard.
Changed the = to >.
c) Used your keyboard to input 40 - all ok so far.
d) Tried the Clear button - just for the heck of it.  I am testing after all.  Clear doesn't work.  **_There is the first thing I found to correct._**
e) The Add Condition is not enabled.  _It should be._  (I just went into the control and did > again and it was now enabled!) That's the second thing to correct.
f) Tried the filter - and it worked fine.  Great.
g) Came back to the dialogue and new condition.  In the sub-dialogue it has remembered the 40 from last time, but not the >.  This is now reset to =.  For consistency it should remember both components.  That's the third thing to correct.
h) So finally - so far - I went back to the main Filter dialogue and used Reset.  Then I returned to the sub-dialogue and it had still remembered yield > 40.  Perhaps check with @danny but I thought Reset would reset the sub-dialogue as well?
i) And finally, finally the new keyboard has the label "Basic".  I wonder if a label is needed in this dialogue.

By the way the main change is working

Wycklife:
@rdstern I also experienced these happening , I shared these with you via the skype but then, I asked Patrick why the dialog is like that. But he said that's how it should work except that the Add condition button should be disabled when the receiver is empty which I have fixed already.

Wycklife:
@rdstern Could you please try this a gain.

Wycklife:
@rdstern We have discussed part 4 above with @dannyparsons and @volloholic . We need to add a new validation type for the control. I'm going to write this as a new issue and will be fixed by me as well.

rdstern:
Sounds good.  That means that the existing progress - without that checking - could be merged.  Then the resolution of that issue can proceed later.

PR #5761, opened 2020-03-09
More changes in the Model menu

### LINKED ISSUE #5758: More changes in the Model menu
The current Model menu looks like this:
image

I suggest:
a) This gives too much importance to the Hypothesis tests.  It is reasonable for them to be (instead) within the Fit Model section, because accepting a null hypothesis usually implies that a simpler model is ok.  So I propose that the current Fit Model sub-menu - currently like this:
image
Should instead move Hypothesis Tests... to below the line, i.e. just above "Model Keyboard".  I further suggest it be called **_"Hypothesis Tests Keyboard..."_**
b) The Model Keyboard... becomes **_"Fit Model Keyboard..."_**
c) See below for the reasoning to also add a new "section above the keyboards with **_Augment Data..._**.  (This moves the broom augment dialogue from Use Model to Fit Model..)

d) I further suggest that the Glance and Tidy dialogues (from broom) are more appropriate in the Compare Models section.  So, there is a line, in the Compare Models section and then Glance and Tidy move there.  I remind you that glance gives a single row for each model.  So it is at the model level.  Tidy gives a row for each parameter of each model - so it is at the parameter level.  

So, they are not Use Model.  Broom works for a single model only - so that would be more appropriate at the Fit Model level.  But we are extending it to provide information (multiple receivers).  So it becomes useful at the compare models menu?

e) The broom **_augment_** dialogue is still at the data level.  This is for a single model and should therefore 1) Have only a single receiver, 2) Be at the Fit Model level and 3) Needs to allow the user to choose what to add to the data.
So, for the menu, I suggest (for now) that it should move to its own little section just above the Fit Model Keyboard...  And it should probably be **_Augment Data..._**? 
f) While changing this menu I suggest we hide (rather than disable) the 4 entries that start with Other. and are currently at the bottom of the Model menu.

Wycklife:
Fixes #5758

### DIFF
--- a/instat/frmMain.Designer.vb
+++ b/instat/frmMain.Designer.vb
@@ -92,8 +92,6 @@ Partial Class frmMain
         Me.mnuModelProbabilityDistributionsShowModel = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuModelProbabilityDistributionsCompareModels = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuModelProbabilityDistributionsRandomSamplesUseModel = New System.Windows.Forms.ToolStripMenuItem()
-        Me.ToolStripSeparator60 = New System.Windows.Forms.ToolStripSeparator()
-        Me.mnuModelHypothesisTests = New System.Windows.Forms.ToolStripMenuItem()
         Me.ToolStripSeparator3 = New System.Windows.Forms.ToolStripSeparator()
         Me.mnuModelFitModel = New System.Windows.Forms.ToolStripMenuItem()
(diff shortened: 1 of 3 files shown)


Wycklife:
@rdstern I have made few changes on this after discussions with David.

Wycklife:
@dannyparsons Could you please merge this.

maxwellfundi:
@Wycklife I see this has been approved but frmMain has merge conflicts. Could you please close this, update your repo and submit these changes. ,

Wycklife:
I am closing this pull request since the branch has conflicts.

PR #5767, opened 2020-03-24
Fixes #5725 Correctly handles special characters in Excel file names

### LINKED ISSUE #5725: Illegal file name in File Open dialogue
At last I am (again) back on the track of problems we had reading Excel - and also perhaps csv files into R-Instat.  I thought it was sheet names with spaces.  But I think that may be ok.  It seems to be the file names that are not fully trapped?  Here is an example where there are some French accents.
[DonnéesClimatiquesSadoré 1983_2014.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/4235867/DonneesClimatiquesSadore.1983_2014.xlsx)

I hope this might be easy to fix?  Of course the work-around is just to rename the file!

lloyddewit:
Fixes #5725 Excel files with special characters in the file name cannot be opened.

**Summary of issue:**

If the user selects `File / Open From File ...` then the file open dialog opens. If the user selects a file with special characters in the name (e.g. `DonnéesClimatiquesSadoré 1983_2014.xlsx`) and tries to open it, then R-Instat displays the error message below, and the Excel file sheet(s) are not loaded.

image

**Analysis:**

R-Instat calls the following R command:

`readxl::excel_sheets(path=""C:/Users/steph/OneDrive/Desktop/DonnéesClimatiquesSadoré 1983_2014.xlsx"")`

This command incorrectly encodes the special characters in the file name. This makes the file name incorrect, and means that the file cannot be found or opened.

A second issue is that, if the user selects all the sheets in a spreadsheet for import, then R-Instat also uses the file name as a variable name for subsequent R commands. If the file name contains special characters then these R commands also fail.

**Solution:**

I solved the first issue by forcing R to correctly encode the file name. I did this by using the R `enc2native` function:

`readxl::excel_sheets(path=enc2native(""C:/Users/steph/OneDrive/Desktop/DonnéesClimatiquesSadoré 1983_2014.xlsx""))`

I solved the second issue by ensuring that, if the file name was used as a variable name, then all special characters were removed from the variable name. For example file name `DonnéesClimatiquesSadoré 1983_2014' is converted to variable name 'DonnesClimatiquesSador1983_2014'. If the resulting variable name is empty (unlikely but possible) then it's set to a default name.

**Testing**
I tested with a folder name that also contained a special character. I tested with the following file names. Note that the last file contains excel sheet names that also contain special characters. All tests passed.

image

Note: This issue was specific to Excel files, and the error was in R commands only used when importing Excel files. I did not test special characters in file names of other file types. If there are bugs in other file types then I'd prefer to raise a new issue for these.

@rdstern - if you think it's a priority then could you test other file types? I think your testing would be much more efficient and effective than mine.


instat/dlgImportDataset.vb
```diff
@@ -1001,24 +989,37 @@ Public Class dlgImportDataset
...
+    ''' <param name="strFilePathTmp">      The full path of the file. </param>
+    '''
+    ''' <returns>   Returns the file name (without extension) with any special characters removed.
+    '''             If the cleaned name is an empty string then returns 'defaultCleanFileName'.
+    ''' </returns>
+    '''--------------------------------------------------------------------------------------------
+    Private Function GetCleanFileName(strFilePathTmp As String) As String
+        Dim strCleanFileName As String
+        strCleanFileName = System.Text.RegularExpressions.Regex.Replace(Path.GetFileNameWithoutExtension(strFilePathTmp), "[^A-Za-z0-9_\-]", "")
+        If strCleanFileName Is "" Or strCleanFileName Is Nothing Then
```
Patowhiz, line 1019: This looks good. And the alternative way is using the VB.NET utility function   `String.IsNullOrEmpty()`

lloyddewit: @Patowhiz Thank you, good suggestion. This PR is already merged so I implemented your suggested change in PR #5806.

PR #5769, opened 2020-03-28
Fixes #5742 Issue multiline commands in script window

### LINKED ISSUE #5742: Issue multiline commands in script window?

lloyddewit:
Fixes #5742 Issue multiline commands in script window.

Note that this issue also has an optional enhancement to allow multi-line functions (enclosed with '{' and '}') to be executed in the script window. This PR does **not** implement this optional enhancement.

**Issue**

In the R-Instat script window, if the user splits commands over multiple lines then R-Instat displays an error. 

@rdstern and @dannyparsons requested that if a script line ends with '+', ',', or '%>%', then the command may continue onto the next line.

**Analysis and solution**

The R  script window commands are processed by 'ucrScript'. It would be possible to add the multi-line functionality to this class.

However, R scripts can also be manually executed from the log window. Therefore I decided to add the multi-line functionality to the 'clsRLink.RunScriptFromWindow' function. This executes R scripts from both the script and log windows. Implementing the functionality here guarantees that a script will be executed in the same way regardless of whether it comes from the script window or the log window. 

Counter argument: The log window only contains computer-generated R commands and therefore is unlikely to have multi-line commands (at least for now). 

**Testing**

I tested with the following scripts:

```
# generate artificial data
set.seed(4321)
x <- 1:100
y <- (x + x^2 + x^3) + rnorm(length(x), mean = 0, sd = mean(x^3) / 4)
my.data <- data.frame(x = x, y = y,
group = c("A", "B"),
y2 = y * c(0.5,2),
w = sqrt(x))

# give a name to a formula
formula <- y ~ poly(x, 3, raw = TRUE)
# no weights
ggplot2::ggplot(my.data, aes(x, y)) +
geom_point() +
geom_smooth(method = "lm", formula = formula) +
ggpmisc::stat_poly_eq(formula = formula, parse = TRUE)

# Initialize `x` 
x <- rnorm(100)

# Update value of `x` and assign it to `x`
x %<>% abs %>% 
sort
```
 All tests passed correctly.

### DIFF
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -187,7 +187,8 @@ Public Class RLink
     ''' <summary> This method executes the <paramref name="strNewScript"/> R script and displays 
     '''           the output as text or graph (determined by <paramref name="strNewScript"/>).
     '''           R commands may be split over multiple lines. This is only allowed if the  
-    '''           non-final line ends with '+', ',', or '%>%'.
+    '''           non-final line ends with '+', ',', or '%>%'; or there are one or more '{'
+    '''           brackets that have not been closed with an equivalent '}' bracket.
     '''           This function is named '...FromWindow' because it's designed to execute scripts 
     '''           entered by a human from a dialog window (e.g. a script window). These scripts 
     '''           may contain R commands split over multiple lines to make the commands more 
@@ -196,16 +197,36 @@ Public Class RLink
     ''' <param name="strNewScript">  is the R script to execute.</param>
     ''' <param name="strNewComment"> is shown as a comment. If this parameter is "" then shows 
     '''                              <paramref name="strNewScript"/> as the comment.</param>
-    Public Sub RunScriptFromWindow(strNewScript As String, strNewComment As String)
-        Dim strScriptCmd As String = ""
+    ''' <param name="bClearScriptCmd"> (Optional) If true then the command buffer is flushed 
+    '''                                before <paramref name="strNewScript"/> is processed. 
+    '''                                Otherwise <paramref name="strNewScript"/> is treated as a 
+    '''                                continuation of the previous string that was sent to this 
+    '''                                function.</param>
+    Public Sub RunScriptFromWindow(strNewScript As String, strNewComment As String, Optional bClearScriptCmd As Boolean = True)
+        Static strScriptCmd As String = ""
+        Static iOpenCurlyBraces As Integer = 0
+
+        'by default, flush out any unexecuted full or partial commands
+        If bClearScriptCmd Then
+            strScriptCmd = ""
+        End If
 
         'for each line in script
         For Each strScriptLine As String In strNewScript.Split(Environment.NewLine)
-            strScriptLine = strScriptLine.Trim(vbLf) 'remove linefeed at end of line
-            strScriptLine = Trim(strScriptLine)      'trim any spaces from front or end
+            'remove any comments (character '#' and anything after)
+            Dim iCommentPos As Integer = strScriptLine.IndexOf("#")
+            Select Case iCommentPos
+                Case 0      'a normal comment line (starts with '#')
+                    Continue For
+                Case Is > 0 ' a line with an appended comment (e.g. 'x <- 1 # generate data' converted to 'x <- 1 ')
+                    strScriptLine = strScriptLine.Substring(0, iCommentPos - 1)
+            End Select
 
-            'if line is empty or a comment, then ignore line
-            If strScriptLine.Count <= 0 Or strScriptLine.StartsWith("#") Then
+            'remove any linefeeds and spaces from front or end
+            strScriptLine = strScriptLine.Trim(vbLf).Trim()
+
+            'if line is empty then ignore line
+            If strScriptLine.Length <= 0 Then
                 Continue For
             End If
 
@@ -218,7 +239,15 @@ Public Class RLink
             If strScriptLine.Length >= 3 Then
                 strLast3Chars = strScriptLine.Substring(strScriptLine.Length - 3)
             End If
-            If cLastChar = "+" Or cLastChar = "," Or strLast3Chars = "%>%" Then
+            If cLastChar = "+" OrElse cLastChar = "," OrElse strLast3Chars = "%>%" Then
+                Continue For
+            End If
+
+            'if there are open curly braces then assume command is not complete
+            Dim iNumOpenCurlies = strScriptCmd.Where(Function(c) c = "{"c).Count
+            Dim iNumClosedCurlies = strScriptCmd.Where(Function(c) c = "}"c).Count
+            If iNumOpenCurlies <> iNumClosedCurlies Then
+                strScriptCmd &= "; " 'allow next command to be on same line
                 Continue For
             End If
 
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -197,24 +197,22 @@ Public Class RLink
     ''' <param name="strNewComment"> is shown as a comment. If this parameter is "" then shows 
     '''                              <paramref name="strNewScript"/> as the comment.</param>
     Public Sub RunScriptFromWindow(strNewScript As String, strNewComment As String)
-        Dim strScriptLines() As String = strNewScript.Split(Environment.NewLine)
-        Dim strScriptCmds(strScriptLines.Length - 1) As String 'num of cmds is always <= num of lines
-        Dim iScriptCmdIndex = 0
+        Dim strScriptCmd As String = ""
 
-        'create an array of clean, single-line R commands
-        For i = 0 To strScriptLines.Length - 1
-            Dim strScriptLine As String = strScriptLines(i).Trim(vbLf) 'remove linefeed at end of line
+        'for each line in script
+        For Each strScriptLine As String In strNewScript.Split(Environment.NewLine)
+            strScriptLine = strScriptLine.Trim(vbLf) 'remove linefeed at end of line
+            strScriptLine = Trim(strScriptLine)      'trim any spaces from front or end
 
             'if line is empty or a comment, then ignore line
             If strScriptLine.Count <= 0 Or strScriptLine.StartsWith("#") Then
                 Continue For
             End If
 
-            'else add line of script to array of cmds
-            strScriptCmds(iScriptCmdIndex) &= strScriptLine
+            'else append line of script to command
+            strScriptCmd &= strScriptLine
 
-            'if line ends in a '+', ',', or '%>%' then assume cmd is not complete
-            strScriptLine = Trim(strScriptLine) 'trim any spaces from front or end
+            'if line ends in a '+', ',', or '%>%' then assume command is not complete
             Dim cLastChar As Char = strScriptLine.Last
             Dim strLast3Chars As String = ""
             If strScriptLine.Length >= 3 Then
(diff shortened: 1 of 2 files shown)


instat/clsRLink.vb
```diff
@@ -184,29 +184,52 @@ Public Class RLink
...
-                End If
-                RunScript(strScript:=strLine.Trim(vbLf), iCallType:=iCallType, strComment:=strComment, bSeparateThread:=False, bSilent:=False)
-                strComment = ""
+        'for each line in script
+        For Each strScriptLine As String In strNewScript.Split(Environment.NewLine)
+            strScriptLine = strScriptLine.Trim(vbLf) 'remove linefeed at end of line
+            strScriptLine = Trim(strScriptLine)      'trim any spaces from front or end
+
+            'if line is empty or a comment, then ignore line
+            If strScriptLine.Count <= 0 Or strScriptLine.StartsWith("#") Then
```
Patowhiz:
```suggestion
            If strScriptLine.Length<= 0 OrElse strScriptLine.StartsWith("#") Then

```

Patowhiz: A minor change though I find the short circuit orElse more optimal than the or based on VB.NET documentation.
Same also for string.Length which is a property vs string.count which is function inherited by all IEnumerable objects.

instat/clsRLink.vb
```diff
@@ -184,29 +184,52 @@ Public Class RLink
...
-                If strLine.Contains(strInstatDataObject & "$get_graphs") Then
-                    iCallType = 3
-                Else
-                    iCallType = 2
-                End If
-                RunScript(strScript:=strLine.Trim(vbLf), iCallType:=iCallType, strComment:=strComment, bSeparateThread:=False, bSilent:=False)
-                strComment = ""
+        'for each line in script
+        For Each strScriptLine As String In strNewScript.Split(Environment.NewLine)
+            strScriptLine = strScriptLine.Trim(vbLf) 'remove linefeed at end of line
```
Patowhiz:
```suggestion
            strScriptLine = strScriptLine.Trim(vbLf).Trim() 'remove line feed at end of line and any spaces from front or end

```

instat/clsRLink.vb
```diff
@@ -184,29 +184,52 @@ Public Class RLink
...
+            'else append line of script to command
+            strScriptCmd &= strScriptLine
+
+            'if line ends in a '+', ',', or '%>%' then assume command is not complete
+            Dim cLastChar As Char = strScriptLine.Last
+            Dim strLast3Chars As String = ""
+            If strScriptLine.Length >= 3 Then
+                strLast3Chars = strScriptLine.Substring(strScriptLine.Length - 3)
+            End If
+            If cLastChar = "+" Or cLastChar = "," Or strLast3Chars = "%>%" Then
```
Patowhiz:
```suggestion
            If cLastChar = "+" OrElse cLastChar = "," OrElse strLast3Chars = "%>%" Then

```

instat/clsRLink.vb
lloyddewit: @Patowhiz thank you for the review! I agree and accept both your suggested changes above.

You triggered me to research the difference between Or and OrElse in VB. It's different to other languages. I now understand that `Or` forces evaluation of both conditions, whereas `OrElse` stops if the first condition fails. I think in 99% of cases, `OrElse` is best.

I declared variables inside a loop because the last coding guidelines I followed (for Golang) said to declare variables just before they are used rather than at the top of the function. This is so that maintainers don't have to scroll up to see the declaration (aids readability). It also sometimes saves a line of code resetting the variable every ieration. There is no measurable performance impact for the user.

instat/clsRLink.vb
lloyddewit: Agreed and implemented, thanks

instat/clsRLink.vb
lloyddewit: Agreed and implemented, thanks

Note: You'll see al the above changes included in my next commit.

lloyddewit:
@Patowhiz I committed your suggested changes and implemented the enhancements requested by Roger. If you have time then another code review would be great.

@rdstern As requested, I implemented functionality to allow multi-line commands to be entered one at a time (by ctrl-enter). One side-effect is that the buffer is not cleared until the software considers the command is complete. If the user abandons a command half-way through and then tries to run a new valid command then R-Instat will try and run the abandoned partial command together with the new command and generate an error. I worked around this problem by clearing the command buffer whenever the user executes the entire script window, or just the selected text.

I implemented functionality to allow comments to be appended to intermediate lines in a command.

I also implemented functionality to allow multi-line functions to entered. This works by checking if there are any open curly braces. So the function needs to be entered in a certain way. For example:

```
# this is processed correctly
odds1 <- function(x) {
  seq_along(x) %% 2 > 0 
}
# this doesn't work because open curly brace is on next line
odds2 <- function(x) 
{
  seq_along(x) %% 2 > 0 
}
```

Please note that this functionality is only in beta. It'll work for some simple functions. But larger and more complex functions may not be processed correctly. It would be alot of work to correctly parse all R functions. One option is for you to test the types of functions you're likely to enter and let me know what fails. I can then fix the highest priority cases. Alternatively I can remove this beta functionality.

**Testing**

I tested with the following scripts:

```
# one-line comment
  # indented one-line comment
# one-line comment with trailing spaces
odds <- function(x) {
  seq_along(x) %% 2 > 0 
}
odds(1:5) # appended comment

pow <- function(x, y) {
  # one-line comment in function
result <- x^y
  print(paste(x,"raised to the power", y, "is", result)) #appended comment in function
}
pow(8, 2)
```

All tests passed correctly.

lloyddewit:
@rdstern The example you gave worked fine for me and I wasn't able to recreate the error. 

The only thing I did differently was that I needed to add the line `library(ggpmisc)` to get `geom_grob` to run, but I don't think that is relevent.

One possible reason for your error was that maybe there was some old text still in the buffer that corrupted your line-by-line commands. As I stated above:

> The buffer is not cleared until the software considers the command is complete. If the user abandons a command half-way through and then tries to run a new valid command then R-Instat will try and run the abandoned partial command together with the new command and generate an error. I worked around this problem by clearing the command buffer whenever the user executes the entire script window, or just the selected text.

Please could you try and recreate the error? Thanks

Also, if you have any other examples of functions (or multi-line commands) that fail then I'd be very grateful.

lloyddewit:
@rdstern I improved the code to correctly handle carriage returns within functions. The user can now enter multi-line functions (even large ones). I tested with the R commands and functions in the attached file. Please could you also test? Many thanks
Note: If you are testing hard and ctrl-enter doesn't seem to work as you expect, then it may be that there is still old text remaining in the buffer. Executing selected text (or the whole script window) will clear the buffer.

[PR5769multiLineTest.txt](https://github.com/africanmathsinitiative/R-Instat/files/4598032/PR5769multiLineTest.txt)

@Wycklife, @Ivanluv, @Patowhiz - Please could one of you peer review? Many thanks

instat/clsRLink.vb
```diff
@@ -184,29 +184,76 @@ Public Class RLink
...
+    ''' <param name="strNewScript">    The R script to execute.</param>
+    ''' <param name="strNewComment">   Shown as a comment. If this parameter is "" then shows 
+    '''                                <paramref name="strNewScript"/> as the comment.</param>
+    ''' <param name="bClearScriptCmd"> (Optional) If true then the command buffer is flushed 
+    '''                                before <paramref name="strNewScript"/> is processed. 
+    '''                                Otherwise <paramref name="strNewScript"/> is treated as a 
+    '''                                continuation of the previous string that was sent to this 
+    '''                                function.</param>
+    Public Sub RunScriptFromWindow(strNewScript As String, strNewComment As String, Optional bClearScriptCmd As Boolean = True)
+        Static strScriptCmd As String = "" 'static so that script can be added to with successive calls of this function
```
Patowhiz, line 206: From your comment, `Static strScriptCmd As String = ""` is meant for successive recursive calls (my understanding). May I know why you preferred doing it this way rather than using parameters to achieve the same.

I like the changes done to achieve the features requested, and the more I look into this subroutine the more I realise it's implementation as a pure function is more preferable. I don't what your views are about this.

Patowhiz:
@lloyddewit sorry this took me long to have a look at. I've checked the code changes, looks good to me. I've given you my thoughts and questions on some of the implementations.

What I have mainly struggled with is `Static strScriptCmd As String = ""` and why you prefered implementing the "caching" of successive calls that way. I haven't understood that bit well.

instat/clsRLink.vb
lloyddewit: @Patowhiz Great that you reviewed - Thank you!
I made `strScriptCmd` static so that the user can execute R script line by line, even when the complete R command is split over 2 or more lines. The `static` allows the function to remember the previoius lines of script, until the function is called with the final line of script. The command is then executed and `Static strScriptCmd` is then cleared.
I don't think `RunScriptFromWindow` is called recursively, and that wasn't my intention for 'Static strScriptCmd`.
If I misunderstood your question, or you've ideas for improvement then let me know (we can chat if it's easier). Thanks!

lloyddewit: please see my comments in the conversation tab.

lloyddewit:
@Patowhiz Great that you reviewed. Your comments really made me question how I'd implemented things - thank you!
I made `strScriptCmd` static so that the user can execute R script line by line, even when the complete R command is split over 2 or more lines. The `static` allows the function to remember the previous lines of script, until the function is called with the final line of script. The command is then executed and `strScriptCmd` is then cleared.
Your earlier comment that it is not a 'pure' function is very interesting (sorry your comments are no longer visible here because I mistakenly clicked 'resolve conversation'). You are correct and I agree that implementing this as a pure function would be better. We would still need to cache incomplete commands so I think we would still need a static somewhere making a different function 'impure'. We could potentially move the caching into `ucrScript.RunText`. If you think that would be better (or have other ideas) then let me know, and I'll look into the implications.
In an earlier comment, you mentioned recursion. I don't think `RunScriptFromWindow` is called recursively, and that wasn't my intention for 'Static strScriptCmd`.
If I misunderstood your comments, or you've ideas for improvement then let me know (we can chat if it's easier). Thanks!

Patowhiz:
@lloyddewit thanks for the response. I've now understood your use of static. Looks good on my side. We could explore implementing it as a pure function called in another function just to have the code execution much clearer for future developers(I think). But so far seems to be working fine and nicely written!

lloyddewit:
@Patowhiz The more I thought about your comment, the more convinced I was that you were correct. This PR has already been merged so I implemented your suggested change in PR #5796. I think the new version is more understandable and maintainable. So thanks again for the good comment.

@maxwellfundi This PR has been superceded by PR #5796 so I think this PR can be closed.

PR #5790, opened 2020-05-15
Enabling display of numeric data in scientific format in the grid

### LINKED ISSUE #5782: Display of numeric data in R-Instat grid - possibly scientific?
Here is an example of the display of data currently:
image

The last variable (p-values) is my concern.  I have set the number of significant figures to 1, but that doesn't help.  (Setting to zero gives either 0 or 1.)  This is going to be a typical display for p-values, where some may be very small, while others are close to 1. 

The metadata allows for scientific, but that option is not active and I don't know if it is allowed in reogrid.
Another option - for us, would be to be able to choose decimals?
Is anything possible easily?

shadrackkibet:
@rdstern this is ready for testing. Fixes #5782


shadrackkibet:
Great, I will get in touch with him.

dannyparsons:
The edits shouldn't be here in converting a column to a different type. Scientific is not a column type but just a format to display. There is no change to the data when using scientific notation.
The change needed is in the same place where the decimal places option is used. This is used in the `convert_to_character_matrix()` function which takes the data frame and formats it to display in the grid. Notice that there is a `decimal_places` parameter which is a vector passed in from the metadata. There should be an equivalent for `scientific` which is currently hard coded as `FALSE`.

I hope that gives you a starting point. Let me know if you want to discuss in more detail.

shadrackkibet:
Thanks, @dannyparsons your guidance was informative enough and I was able to fix the issue. @rdstern this seems to work for me now. Please check.

shadrackkibet:
yes, that is easy to implement. I will have it implemented in another pull request since this is merged.

rdstern:
@shadrackkibet great, and please can you have a quick chat to Maxwell, as I have still to send him the description of the way we are working now - so he waits for a second approval, from Stephen, before merging.  We will circulate the "system" today.

PR #5796, opened 2020-05-22
Improves PR #5769 Issue multiline commands in script window

lloyddewit:
PR #5769 implements multiline commands in the the script window. After the merge @Patowhiz provided peer review comments. This PR addresses those peer review comments.

In particular, 'clsRLink.RunScriptFromWindow' is now a pure function (removed static variable). 

The static variable was moved to 'ucrScript.mnuRunCurrentLine_Click' (this is the function that allows the user to execute script line by line using ctrl-enter). It is better to have the static here because this is the only function that requires unexecuted script to be kept for successive calls.

@Patowhiz please could you peer review again?

@rdstern when @Patowhiz has finished his peer review, please could you test again. The functionality should be exactly the same. 
The only side-effect of the change is that the buffer used by executing line by line (ctrl-enter) is only cleared when a complete command is executed. I'm concerned that theoretically, the user could enter incorrect code that causes subsequent (valid) commands not to run correctly. In this case, the only way to clear the buffer would be to restart R-Instat. I tried to trigger this error but could not. Maybe in practice it's not a big issue.

### DIFF
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -18,55 +18,132 @@ Imports RDotNet
 Imports unvell.ReoGrid
 Imports System.IO
 
+'''--------------------------------------------------------------------------------------------
+''' <summary>   An object of this class represents an R interface. 
+'''             Each instance of this class has its own R environment.
+'''             <para>
+'''             This class provides methods and fields to:</para>
+''' <list type="bullet">
+'''     <item><description>
+'''             Define the R environment including paths and R versions.
+'''     </description></item><item><description>
+'''             Open a direct link to the R environment and load the required R packages. 
+'''     </description></item><item><description>
+'''             Execute a specified R script and display the output as text, graph or in a 
+'''             web browser.
+'''     </description></item><item><description>
+'''             Write to log and debug files.
+'''     </description></item><item><description>
+'''             Specify the fonts and colours to be used in the output window.
+'''     </description></item><item><description>
+'''             Populate and configure the data frame control combo box, and the list view control.
+'''     </description></item><item><description>
+'''             Get names of data frames, columns, models, tables, filters, graphs and survs.
+'''     </description></item><item><description>
+'''             Get properties and other metadata.
+'''     </description></item><item><description>
+'''             Execute specific climatic and corruption commands.
+'''     </description></item>
+''' </list></summary>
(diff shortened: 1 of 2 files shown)


lloyddewit:
@Patowhiz There are many changes shown in `clsRLink.vb`. Almost all these changes are comment blocks which I added to resolve the merge conflicts. The only changes you need to review are those in the `RunScriptFromWindow` function starting from line 340.

@rdstern I thought more about the potential situation where the line by line input is affected by previous incorrect R script which is still in its buffer. One option would be to add a `Clear current line buffer' option to the right-click menu. If you think this is a priority then please let me know. Thanks

rdstern:
I seem to have found an example that doesn't work well.  I am not sure if we just accept this as an exception, but I wonder if a quote is included, then this might also work?  Here it is:
```
Input = ("
 Individual  Hand     Length
 A           Left     17.5
 B           Left     18.4
 C           Left     16.2
 D           Left     14.5
 E           Left     13.5
 F           Left     18.9
 G           Left     19.5
 H           Left     21.1
 I           Left     17.8
 J           Left     16.8
 K           Left     18.4
 L           Left     17.3
 M           Left     18.9
 N           Left     16.4
 O           Left     17.5
 P           Left     15.0
 A           Right    17.6
 B           Right    18.5
 C           Right    15.9
 D           Right    14.9
 E           Right    13.7
 F           Right    18.9
 G           Right    19.5
 H           Right    21.5
 I           Right    18.5
 J           Right    17.1
 K           Right    18.9
 L           Right    17.5
 M           Right    19.5
 N           Right    16.5
 O           Right    17.4
 P           Right    15.6
")

Data = read.table(textConnection(Input),header=TRUE)
```

lloyddewit:
@rdstern If I add a check that only executes the line when there are an even number of `"` then this test case should also pass. I will implement this.

rdstern:
Great!  Thanks.

Patowhiz:
@lloyddewit I can't test this due to the conflicting files.

Patowhiz:
The code looks good.

lloyddewit:
I will close this PR because I could not fix the merge conflicts in a good way so I made a new PR from the master branch and reimplemented the changes.

This PR is replaced by PR #5808 'Replaces PR #5796 Issue multiline commands in script window'.

PR #5797, opened 2020-05-23
Renaming filters and calculations using Rename objects dialog

### LINKED ISSUE #5792: Rename, delete and reorder filters, like graphs and models
image

shadrackkibet:
Fixes #5792. @rdstern  this is the implementation to be able to rename filters and calculations from rename objects dialog. I request you to do initial testing.


rdstern:
This seems to be working fine.  At least the dialogue itself.  That's great.  I have tried with filters, but not yet with a calculation.  Very exciting to see this progress!
I am using the usual survey data and filtering to get 2 or 3 of the four villages.
Three very minor points:
a) When you return to the dialogue it has cleared the old name, but kept the new name.  I am not sure what's right, but the **_Prepare > Data Frame > Rename Column_** is slightly different.  It doesn't keep either.  So perhaps clear both the old and the new name?
b) There is also one further minor difference.  You propose Filter11 as the new name instead of Filter1.  If I change that to Filter1, i.e. no change, then you disable OK.  The Rename Column proposes the new name as the same as the old name and allows OK, thugh (of course) then it doesn't do anything.  I am not sure what I prefer here - back to you on that.  It is a very minor point.
c) I really like the drop down with the Default of Objects.  Discussing with David he confirms that splitting and having an All would be a bit complicated.  But he thought that adding (for example) a line and then allowing the Objects under that line to be given by their type might be easy.  So the drop down might change into:
Objects
Filters
Calculations
___________
Tables
Graphs
Models

But don't do that unless it is really easy.  I don't want to lose focus on the changes to the filter system.

Now a bigger problem.  Although the dialogue works ok t causes other problems - you may know about them.
a) I had filtered to 26 observations.  At the bottom it says Showing 26 of 26 rows.  When I use your new rename dialogue it now says Showing 26 of 0 rows.  
b) And also some dialogues no longer work.  For example the Rename Column no longer works.  Also I can't re-use the new filter name in the Filter dialogue.

shadrackkibet:
@rdstern The main problem you reported above is now fixed, you could check.

shadrackkibet:
closing this because of many files changes. Moved changes into a new pull request.

shadrackkibet:
Yes, I know about this. This should be fixed on the filter dialog. I will fix it when I start work in the filter dialog

PR #5802, opened 2020-05-26
Added anew keyboard from coin package to the Hypothesis Test dialog

### LINKED ISSUE #4460: |Adding the coin package for more hypothesis tests
The coin package looks interesting for us.  I have included Lily in case it could be useful for her MSc project?  It is mentioned in the clinical trials and survival task views and so just may be of some interest to consider.
More generally, for us it includes a pretty general system for hypothesis testing.  It looks to be as close as hypothesis testing can get to our ideas of looking for general methods.

We now have an obvious place to start looking at it, namely as a keyboard in our hypothesis testing dialogue - and I have been on the lookout for useful extra packages/keyboards.  It also has an interesting set of example data sets.

The keyboard could be sensibly structured with categories of tests, namely 
1) Contingency Tests, (4)
2) Correlation Tests (4)
3) Location Tests (6)
4) Scale Tests (6)
5) Survival Tests (1)
6) Symmetry Tests (4)
There are also a few additional individual tests, etc, so this may become 2 keyboards.

This package seems obvious enough for the future that I suggest we include it in Version 4.12, but leave the coding etc to later.  Then if anyone does want to use those tests they would be available, through the hypothesis testing dialogue, though not on a keyboard.

Wycklife:
Fixes #4460


rdstern:
Well done.  That looks neat and I like the layout.
In the location section the correct commands are not there - it repeats oneway 6 times.
And there is an instance where it is wilcoksign and not wilcoxsign

That apart it seems to work.

Wycklife:
@rdstern I have fixed the changes above.

shadrackkibet:
Great to see progress on this, I will be code checking.

instat/dlgHypothesisTestsCalculator.vb
```diff
@@ -659,4 +672,188 @@ Public Class dlgHypothesisTestsCalculator
             ucrBase.clsRsyntax.bExcludeAssignedFunctionOutput = True
         End If
     End Sub
+
+    Private Sub cmdOneway1_Click(sender As Object, e As EventArgs) Handles cmdOneway1.Click
+        If ucrChkIncludeArguments.Checked Then
+            ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("coin::oneway_test(formula = , data = , subset = NULL, weights = NULL)", 42)
```
shadrackkibet, line 678: These are not the only parameters to these functions. I am not sure which rule we follow generally for keyboards. I think the idea is to include all parameters where necessary including `...` to indicate that the function allows other parameters which could be passed internally.

Wycklife: For the keyboards, we have been omitting the three dots as this was suggested by @rdstern

shadrackkibet: Okay, what about the other parameters?

rdstern:
@shadrack, great that you are checking the code.  On the addition of parameters we now add them when they are always needed.  
It was almost ok, and @Wycklife says he has made those has made changes.  Perhaps you could check that too.  Then tell @lloyddewit that it just remains for him to do a final check and it can be merged.

instat/dlgHypothesisTestsCalculator.vb
```diff
@@ -659,4 +672,188 @@ Public Class dlgHypothesisTestsCalculator
...
+        If ucrChkIncludeArguments.Checked Then
+            ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("coin::wilcoxsign_test(formula = , data = , subset = NULL)", 26)
+        Else
+            ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("coin::wilcoxsign_test()", 1)
+        End If
+    End Sub
+
+    Private Sub cmdFriedman1_Click(sender As Object, e As EventArgs) Handles cmdFriedman1.Click
+        If ucrChkIncludeArguments.Checked Then
+            ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("coin::friedman_test(object = )", 2)
```
shadrackkibet, line 742: Following up on a comment by @rdstern I claim for consistency purposes you need to add the following parameters. `formula = , data = , subset = NULL` here.

instat/dlgHypothesisTestsCalculator.vb
```diff
@@ -659,4 +672,188 @@ Public Class dlgHypothesisTestsCalculator
...
+        Else
+            ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("coin::friedman_test()", 1)
+        End If
+    End Sub
+
+    Private Sub cmdQuade1_Click(sender As Object, e As EventArgs) Handles cmdQuade1.Click
+        If ucrChkIncludeArguments.Checked Then
+            ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("coin::quade_test(object = )", 2)
```
shadrackkibet, line 750: here?

rdstern:
For now you can often give the function as coin::function(y ~ x).  We don't usually provide the data frame, because that is shown in the dialogue.  And the subset = null isn't always necessary.  So, at least initially I am happy to leave them as they are now.  I did wonder about coin::function( ~  ), but it is simpler to use the ~ key that is available.  So, initially I am happy with just the () as is given now.  We may later want to add some arguments, once we practice in its use.

shadrackkibet:
Thanks, @rdstern. Then we should remove these parameters from other functions where it is added for consistency? otherwise, this is ready for @lloyddewit to review.

Wycklife:
@rdstern I have made this change. You can now approve and then merge since I need to work on the same dialog.

lloyddewit:
This was merged before I reviewed (which I understand will be prevented in the new system). 
I've now reviewed and it looks good from my side.

PR #5804, opened 2020-05-26
Renaming filters and calculations using Rename objects dialog

### LINKED ISSUE #5792: Rename, delete and reorder filters, like graphs and models

shadrackkibet:
Fixes #5792. I moved changes from #5797 because of many files changes after trying to fix conflicts.

### DIFF
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -322,6 +322,7 @@ Public Class RLink
     End Function
 
 
+    '''--------------------------------------------------------------------------------------------
     ''' <summary> This method executes the <paramref name="strNewScript"/> R script and displays 
     '''           the output as text or graph (determined by <paramref name="strNewScript"/>).
     '''           <para>R commands may be split over multiple lines. This is only allowed if the  
@@ -331,6 +332,7 @@ Public Class RLink
     '''           entered by a human from a dialog window (e.g. a script window). These scripts 
     '''           may contain R commands split over multiple lines to make the commands more 
     '''           readable.</para>
+    '''</summary>
     ''' <param name="strNewScript">    The R script to execute.</param>
     ''' <param name="strNewComment">   Shown as a comment. If this parameter is "" then shows 
     '''                                <paramref name="strNewScript"/> as the comment.</param>
@@ -338,8 +340,8 @@ Public Class RLink
     '''                                before <paramref name="strNewScript"/> is processed. 
     '''                                Otherwise <paramref name="strNewScript"/> is treated as a 
     '''                                continuation of the previous string that was sent to this 
-    '''                                function.</param></summary>
-
+    '''                                function.</param>
+    '''--------------------------------------------------------------------------------------------
     Public Sub RunScriptFromWindow(strNewScript As String, strNewComment As String, Optional bClearScriptCmd As Boolean = True)
         Static strScriptCmd As String = "" 'static so that script can be added to with successive calls of this function
 
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -331,7 +331,6 @@ Public Class RLink
     '''           entered by a human from a dialog window (e.g. a script window). These scripts 
     '''           may contain R commands split over multiple lines to make the commands more 
     '''           readable.</para>
-
     ''' <param name="strNewScript">    The R script to execute.</param>
     ''' <param name="strNewComment">   Shown as a comment. If this parameter is "" then shows 
     '''                                <paramref name="strNewScript"/> as the comment.</param>
@@ -339,7 +338,7 @@ Public Class RLink
     '''                                before <paramref name="strNewScript"/> is processed. 
     '''                                Otherwise <paramref name="strNewScript"/> is treated as a 
     '''                                continuation of the previous string that was sent to this 
-    '''                                function.</param>
+    '''                                function.</param></summary>
 
     Public Sub RunScriptFromWindow(strNewScript As String, strNewComment As String, Optional bClearScriptCmd As Boolean = True)
         Static strScriptCmd As String = "" 'static so that script can be added to with successive calls of this function
@@ -1491,6 +1490,10 @@ Public Class RLink
                     clsGetItems.AddParameter("file", Chr(34) & strNcFilePath & Chr(34))
                 Case "variable_sets"
                     clsGetItems.SetRCommand(strInstatDataObject & "$get_variable_sets_names")
+                Case "table"
+                    clsGetItems.SetRCommand(strInstatDataObject & "$get_table_names")
+                Case "calculation"
+                    clsGetItems.SetRCommand(strInstatDataObject & "$get_calculation_names")
             End Select
             clsGetItems.AddParameter("as_list", "TRUE")
             If lstView.TopItem IsNot Nothing Then
--- a/instat/dlgRenameObjects.vb
+++ b/instat/dlgRenameObjects.vb
@@ -45,10 +45,11 @@ Public Class dlgRenameObjects
     Private Sub ReopenDialog()
         ' temp. fix, the receivers should clear only if the name of the object in it has changed
         ucrSelectorForRenameObject.Reset()
-        NewDefaultName()
+        ucrInputNewName.SetName("")
     End Sub
 
     Private Sub InitialiseDialog()
+        Dim dctType As New Dictionary(Of String, String)
         ucrBase.iHelpTopicID = 350
 
         'ucrSelector
@@ -60,14 +61,19 @@ Public Class dlgRenameObjects
         ucrReceiverCurrentName.Selector = ucrSelectorForRenameObject
         ucrReceiverCurrentName.SetMeAsReceiver()
         ucrReceiverCurrentName.SetParameterIsString()
-        ucrReceiverCurrentName.SetItemType("object")
-        ucrReceiverCurrentName.strSelectorHeading = "Objects"
-
         'ucrNewName
         ucrInputNewName.SetParameter(New RParameter("new_name", 2))
         ucrInputNewName.SetValidationTypeAsRVariable()
 
-        NewDefaultName()
+        ucrInputType.SetParameter(New RParameter("object_type", 3))
+        dctType.Add("Objects", Chr(34) & "object" & Chr(34))
+        dctType.Add("Filters", Chr(34) & "filter" & Chr(34))
+        dctType.Add("Calculations", Chr(34) & "calculation" & Chr(34))
+        dctType.Add("Tables", Chr(34) & "table" & Chr(34))
+        dctType.Add("Graphs", Chr(34) & "graph" & Chr(34))
+        dctType.Add("Models", Chr(34) & "model" & Chr(34))
+        ucrInputType.SetItems(dctType)
+        ucrInputType.SetDropDownStyleAsNonEditable()
     End Sub
 
     Private Sub SetDefaults()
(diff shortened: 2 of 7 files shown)


shadrackkibet:
> This seems to work now. From my view it could be merged. I tried also using the rename variable dialogue, with the filter in operation. Also, the new filter name seems to work fine.
The only minor point is that the filter dialogue when I returned there) still includes the old name. I think that is a problem in a number of dialogues. Some are ok, when I change the name of the column, the Fit Model > Two variables has cleared the corresponding receiver.

rdstern:
This is now ready for a code check, by @lloyddewit before merging.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -6,51 +6,51 @@ DataSheet <- R6::R6Class("DataSheet",
...
-      data_name <- make.names(iconv(data_name, to = "ASCII//TRANSLIT", sub = "."))
-    }
-    self$append_to_metadata(data_name_label, data_name)
-  }
-  else if (!self$is_metadata(data_name_label)) {
-    if (( is.null(data_name) || data_name == "" || missing(data_name))) {
-      self$append_to_metadata(data_name_label,paste0("data_set_",sprintf("%03d", start_point)))
-      if (messages) {
-        message(paste0("No name specified in data_tables list for data frame ", start_point, ". 
+                           {
```
lloyddewit, line 9: why is all this code indented?
General comment: with all the indentation changes it's hard to see which lines have functionaly changed. Is it possible to fix the indentation to make a cleaner diff (or alternatively state which lines have changed and need peer reviewing)?

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -476,8 +476,19 @@ DataSheet$set("public", "get_calculations", function() {
...
+    ex_ind = which(out %in% excluded_items)
+    if(length(ex_ind) != length(excluded_items)) warning("Some of the excluded_items were not found in the list of calculations")
+    if(length(ex_ind) > 0) out = out[-ex_ind]
+  }
+  if(as_list) {
+    lst = list()
+    lst[[self$get_metadata(data_name_label)]] <- out
+    return(lst)
+  }
+  else return(out)
```
lloyddewit, line 491: remove the `else` here (not needed)?

Or is the following more readable (less indentation)?
```
if(!as_list) {
  return(out)
}
lst = list()
lst[[self$get_metadata(data_name_label)]] <- out
return(lst)
```

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -476,8 +476,19 @@ DataSheet$set("public", "get_calculations", function() {
 }
 )
 
-DataSheet$set("public", "get_calculation_names", function() {
-  return(names(private$calculations))
+DataSheet$set("public", "get_calculation_names", function(as_list = FALSE, include = list(), exclude = list(), excluded_items = c()) {
```
lloyddewit, line 479: Some parameters are not used, only `excluded_items` seems to be supported, does that matter?

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1680,10 +1691,25 @@ DataSheet$set("public", "get_last_graph", function() {
 }
 )
 
-DataSheet$set("public", "rename_object", function(object_name, new_name) {
-  if(!object_name %in% names(private$objects)) stop(object_name, " not found in objects list")
-  if(new_name %in% names(private$objects)) stop(new_name, " is already an object name. Cannot rename ", object_name, " to ", new_name)
-  names(private$objects)[names(private$objects) == object_name] <- new_name
+DataSheet$set("public", "rename_object", function(object_name, new_name, object_type = "object") {
+  if(!object_type %in% c("object","filter","calculation", "graph", "table","model")) stop(object_type, " must be either object, filter or a calculation")
```
lloyddewit, line 1695: Is the error message complete? Should it be:
```
stop(object_type, " must be either object, filter, calculation, graph, table or model")
```
Also (small point) the spacing is inconsistent in the if condition.

instat/dlgRenameObjects.vb
```diff
@@ -110,10 +117,36 @@ Public Class dlgRenameObjects
         bUseSelectedColumn = False
     End Sub
 
-    Private Sub ucrInputNewName_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverCurrentName.ControlValueChanged
+    Private Sub ucrReceiverCurrentName_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverCurrentName.ControlValueChanged
         NewDefaultName()
```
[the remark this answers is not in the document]
shadrackkibet: Yes, I removed this in places where it was redundant so we can do this.

instat/dlgRenameObjects.vb
```diff
@@ -77,6 +83,7 @@ Public Class dlgRenameObjects
         ucrInputNewName.Reset()
```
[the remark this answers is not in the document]
shadrackkibet: Actually we should remove this. Controls are reset through `setRcodeforcontrols`.

instat/dlgRenameObjects.vb
```diff
@@ -60,14 +61,19 @@ Public Class dlgRenameObjects
...
         ucrReceiverCurrentName.SetMeAsReceiver()
         ucrReceiverCurrentName.SetParameterIsString()
-        ucrReceiverCurrentName.SetItemType("object")
-        ucrReceiverCurrentName.strSelectorHeading = "Objects"
-
         'ucrNewName
         ucrInputNewName.SetParameter(New RParameter("new_name", 2))
         ucrInputNewName.SetValidationTypeAsRVariable()
 
-        NewDefaultName()
```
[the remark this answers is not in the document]
shadrackkibet: We shouldn't do this for the same reason as above.

instat/dlgRenameObjects.vb
```diff
@@ -110,10 +117,36 @@ Public Class dlgRenameObjects
...
-    Private Sub ucrInputNewName_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverCurrentName.ControlValueChanged
+    Private Sub ucrReceiverCurrentName_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverCurrentName.ControlValueChanged
         NewDefaultName()
     End Sub
 
+    Private Sub ucrInputType_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputType.ControlValueChanged
+        Dim strType As String
+        strType = ucrInputType.GetValue()
+
+        Select Case strType
```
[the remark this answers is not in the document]
shadrackkibet: yes, we can.

instat/dlgRenameObjects.vb
```diff
@@ -77,6 +83,7 @@ Public Class dlgRenameObjects
         ucrInputNewName.Reset()
 
         clsDefaultFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$rename_object")
+        clsDefaultFunction.AddParameter("object_type", Chr(34) & "object" & Chr(34), iPosition:=3)
```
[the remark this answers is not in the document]
shadrackkibet: We need this here to set the default for the control.

instat/static/InstatObject/R/data_object_R6.R
shadrackkibet: I didn't intend to do indentation changes. I think that happened when I was trying to move changes from another branch that had conflicts with `Rlink`. I would like to revert these indentation changes but I don't know how to. Any ideas? Although there is no functionality change due to indentation changes.

instat/static/InstatObject/R/data_object_R6.R
shadrackkibet: It doesn't matter at the moment. Actually I should have added a comment here to say the parameters are currently not used. I added it because these parameters are added in the `Rlink` similar to other `get_?_names` methods. I think it was added for future use because we have it in other methods as well e.g `get_filter_names`

shadrackkibet: Actually, I think we should have one method `get_object_filter_calculation_names` and make these objects, filters and calculations options instead of having different methods.

instat/static/InstatObject/R/data_object_R6.R
shadrackkibet: That works. I like the simplification. Thanks

instat/static/InstatObject/R/data_object_R6.R
shadrackkibet: Objects refer to `graphs, tables` and `models` in this case.  I added that because we want to be able to distinguish them in the dialog. Please see my comment in line 1696.

instat/clsRLink.vb
```diff
@@ -331,15 +331,14 @@ Public Class RLink
     '''           entered by a human from a dialog window (e.g. a script window). These scripts 
     '''           may contain R commands split over multiple lines to make the commands more 
     '''           readable.</para>
-
```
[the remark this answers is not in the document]
lloyddewit: Resolved, thanks

instat/clsRLink.vb
```diff
@@ -331,15 +331,14 @@ Public Class RLink
...
     ''' <param name="strNewScript">    The R script to execute.</param>
     ''' <param name="strNewComment">   Shown as a comment. If this parameter is "" then shows 
     '''                                <paramref name="strNewScript"/> as the comment.</param>
     ''' <param name="bClearScriptCmd"> (Optional) If true then the command buffer is flushed 
     '''                                before <paramref name="strNewScript"/> is processed. 
     '''                                Otherwise <paramref name="strNewScript"/> is treated as a 
     '''                                continuation of the previous string that was sent to this 
-    '''                                function.</param>
+    '''                                function.</param></summary>
```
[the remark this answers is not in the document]
lloyddewit: Resolved, thanks

instat/dlgRenameObjects.vb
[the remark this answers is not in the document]
lloyddewit: Resolved, thanks

instat/dlgRenameObjects.vb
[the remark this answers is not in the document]
lloyddewit: Resolved, thanks

instat/dlgRenameObjects.vb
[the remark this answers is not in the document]
lloyddewit: Resolved, thanks

instat/dlgRenameObjects.vb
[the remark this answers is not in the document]
lloyddewit: Resolved, thanks

instat/static/InstatObject/R/data_object_R6.R
lloyddewit: Resolved, thanks

instat/static/InstatObject/R/data_object_R6.R
lloyddewit: The new indetation looks incorrect so it'll have to be fixed. Does VS have any automatic feature to fix R code indents (like it does vb)? There are probably 3rd party tools that do it but it's probably quicker just to fix it manually.

instat/static/InstatObject/R/data_object_R6.R
lloyddewit: I would personally remove the uncalled parameters to avoid confusion. If you prefer to keep the unused parameters then please could you add a comment explaining why they're unused?
 
Will you implement the `get_object_filter_calculation_names` method? If not maybe you could add a comment suggesting that this method is written when this file is next refactored?

instat/static/InstatObject/R/data_object_R6.R
lloyddewit: resolved, thanks

instat/dlgRenameObjects.vb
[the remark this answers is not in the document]
lloyddewit: OK, I think I understand (I'm still learning about how R-Instat dialogs work).
Resolved, thanks

lloyddewit:
Thank you for all the feedback and changes. From my side, there are only a couple of things still open (both in the R code). If you comment the missing parameters and repair the indentation then I can approve, thanks.

PR #5805, opened 2020-05-26
Adding and disabled buttons into the filter user control and the subdialog

shadrackkibet:
@maxwellfundi this are just designer changes I would like you to have a quick look and merge. I have added required buttons to improve the filter and disabled them as they are not working at the moment.


lloyddewit:
For the record: I gave a quick review and I have no comments.

PR #5808, opened 2020-05-27
Replaces PR #5796 Issue multiline commands in script window

lloyddewit:
PR #5796 had merge conflicts. I could not fix them in a good way so I made this new PR from the master branch and reimplemented the changes.

@Patowhiz You have already peer reviewed these changes. The only new part is the check for open double quotations (") on lines 372 and 376. Please could you quickly review these lines? Thanks

@rdstern The test case below should now pass. Please could you test? Thanks
P.S. It's strange that the '=' works on the last line, but not the first!
```
Input <- ("
 Individual  Hand     Length
 A           Left     17.5
 B           Left     18.4
 C           Left     16.2
 D           Left     14.5
 E           Left     13.5
 F           Left     18.9
 G           Left     19.5
 H           Left     21.1
 I           Left     17.8
 J           Left     16.8
 K           Left     18.4
 L           Left     17.3
 M           Left     18.9
 N           Left     16.4
 O           Left     17.5
 P           Left     15.0
 A           Right    17.6
 B           Right    18.5
 C           Right    15.9
 D           Right    14.9
 E           Right    13.7
 F           Right    18.9
 G           Right    19.5
 H           Right    21.5
 I           Right    18.5
 J           Right    17.1
 K           Right    18.9
 L           Right    17.5
 M           Right    19.5
 N           Right    16.5
 O           Right    17.4
 P           Right    15.6
")

Data = read.table(textConnection(Input),header=TRUE)
```

### DIFF
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -321,7 +321,7 @@ Public Class RLink
         Return bClose
     End Function
 
-
+    '''--------------------------------------------------------------------------------------------
     ''' <summary> This method executes the <paramref name="strNewScript"/> R script and displays 
     '''           the output as text or graph (determined by <paramref name="strNewScript"/>).
     '''           <para>R commands may be split over multiple lines. This is only allowed if the  
@@ -331,23 +331,17 @@ Public Class RLink
     '''           entered by a human from a dialog window (e.g. a script window). These scripts 
     '''           may contain R commands split over multiple lines to make the commands more 
     '''           readable.</para>
-
+    ''' </summary>
     ''' <param name="strNewScript">    The R script to execute.</param>
     ''' <param name="strNewComment">   Shown as a comment. If this parameter is "" then shows 
     '''                                <paramref name="strNewScript"/> as the comment.</param>
-    ''' <param name="bClearScriptCmd"> (Optional) If true then the command buffer is flushed 
-    '''                                before <paramref name="strNewScript"/> is processed. 
-    '''                                Otherwise <paramref name="strNewScript"/> is treated as a 
-    '''                                continuation of the previous string that was sent to this 
-    '''                                function.</param>
-
-    Public Sub RunScriptFromWindow(strNewScript As String, strNewComment As String, Optional bClearScriptCmd As Boolean = True)
-        Static strScriptCmd As String = "" 'static so that script can be added to with successive calls of this function
-
-        'by default, flush out any unexecuted full or partial commands
-        If bClearScriptCmd Then
-            strScriptCmd = ""
-        End If
+    ''' 
+    ''' <returns> Any text at the end of <paramref name="strNewScript"/> that was not executed.
+    '''           If all the text in <paramref name="strNewScript"/> was executed then returns "".
+    '''           </returns>
+    '''--------------------------------------------------------------------------------------------
+    Public Function RunScriptFromWindow(strNewScript As String, strNewComment As String) As String
+        Dim strScriptCmd As String = ""
 
         'for each line in script
         For Each strScriptLine As String In strNewScript.Split(Environment.NewLine)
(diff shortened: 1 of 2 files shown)


Patowhiz:
The code looks good to me. Tested this, also well on my side.

lloyddewit:
@rdstern if you can test/approve then Maxwell can merge. If possible, I'd like to merge quickly because clsRLink is prone to merge conflicts.

@maxwellfundi the merge conflict is easy to resolve, I can explain how. Just let me know when Roger has approved and you're ready to merge. Thanks

maxwellfundi:
@lloyddewit  which changes should i keep on here?

lloyddewit:
@maxwellfundi Please keep the `<<<<<<<` block (i.e. the Issue5742MultilineCmdsMergeFix branch, this is the version of the function with only 2 parameters), thanks

maxwellfundi:
Okay, noted.

instat/clsRLink.vb
```diff
@@ -369,19 +364,17 @@ Public Class RLink
...
             strScriptCmd &= strScriptLine
 
-            'if line ends in a '+', ',', or '%>%'; or there are open curly braces, 
+            'if line ends in a '+', ',', or '%>%'; or there are open curly braces; or open quotations, 
             '    then assume command is not complete
             Dim cLastChar As Char = strTrimmedLine.Last
             Dim strLast3Chars As String = ""
             Dim iNumOpenCurlies = strScriptCmd.Where(Function(c) c = "{"c).Count
             Dim iNumClosedCurlies = strScriptCmd.Where(Function(c) c = "}"c).Count
+            Dim iNumDoubleQuotes = strScriptCmd.Where(Function(c) c = """"c).Count
```
Patowhiz, line 373: @lloyddewit sorry I had not seen this line change. May I ask why you preferred using dynamic variable over explicit variables to achieve this functionality? In my experience in .NET, I found that it's safer to use explicit types not unless you are doing something that truly requires a dynamic type (like when you don't know the object type in design time). Smae applies for variables `iNumOpenCurlies` and `iNumClosedCurlies` .

Patowhiz: I have to add this works perfectly fine, so this question is mainly to understand any merit to this and whether the team can take advantage of this in other parts of the software.

lloyddewit: @Patowhiz Good observation. I think you are correct. For these 3 lines, it would be better to be explicit:
```
            Dim iNumOpenCurlies As Integer = strScriptCmd.Where(Function(c) c = "{"c).Count
```
I think I used dynamic out of habit (I was coding in Golang before R-Instat and this would be normal in Golang).
I looked at Microsoft's coding guidelines (
https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/program-structure/coding-conventions). They don't mention it explicitly and use both dynamic and explicit in their examples.
However, I still agree with you that explicit is better in this case. I will make a new PR to change it.

Patowhiz: @lloyddewit I've also noticed that you have 3 `where` LINQ clause functions, my understanding is that these will do 3 separate loops on the same string `strScriptCmd` doing checks for the characters and counting them. This looks neat and very easy to follow, but my question is will this not impact performance when `strScriptCmd` is long (Considering the fact that these are 3 separate loops on the same thing). If there is no such possibility or if the impact will always be negligible I'm happy with the way it's implemented (The `where` LINQ extension is also among my favourite neat and short way of achieving something). 

One of the alternatives to this will be of course a single loop with a `Select Case` statement that checks for the characters and adds the count to the respective variables(which is long in code but I think more optimal in this case).

         ```
            Dim iNumOpenCurlies As Integer
            Dim iNumClosedCurlies As Integer
            Dim iNumDoubleQuotes As Integer
            For Each c As Char In strScriptCmd
                Select Case c
                    Case "{"c
                        iNumOpenCurlies += 1
                    Case "}"c
                        iNumClosedCurlies += 1
                    Case """"c
                        iNumDoubleQuotes += 1
                End Select
            Next
```

lloyddewit: @Patowhiz Thanks. Your comments make me think and help me learn about VB (this one triggered me to read about LINQ).
I agree that calling these 3 LINQ command in sequence will likely require 3 loops through the script rather than one. There's a trade-off between readability and performance. This loop will only traverse a single R function at most. So any performance improvements will be negligible and will not affect the user. Therefore in this case I think readability is the priority and the 3 LINQ commands are the better option. What do you think?

PR #5809, opened 2020-05-28
Adding a check to convert non upper cases and string of length one and updating the cell in column metadata

shadrackkibet:
@rdstern this is the implementation of your request in #5790. Please test it.

### DIFF
--- a/instat/ucrColumnMetadata.vb
+++ b/instat/ucrColumnMetadata.vb
@@ -112,7 +112,7 @@ Public Class ucrColumnMetadata
         Dim iTemp As Integer
         Dim iNameColumn As Integer = -1
         Dim strNewValue As String
-        Dim strBooleanValsAllowed As String() = {"TRUE", "T", "FALSE", "F"}
+        Dim strBooleanValsAllowed As String() = {"T", "TR", "TRU", "TRUE", "F", "FA", "FAL", "FALS", "FALSE"}
 
         If e.NewData.ToString() = strPreviousCellText Then
             e.EndReason = unvell.ReoGrid.EndEditReason.Cancel
--- a/instat/ucrColumnMetadata.vb
+++ b/instat/ucrColumnMetadata.vb
@@ -138,7 +138,16 @@ Public Class ucrColumnMetadata
             ElseIf strProperty = strLabelsScientific Then
                 strNewValue = e.NewData.ToString.ToUpper
 
-                If Not strValAllowed.Contains(strNewValue) Then
+                e.NewData = strNewValue
+                If strValAllowed.Contains(strNewValue) Then
+                    If strNewValue.Length = 1 Then
+                        If strNewValue.Contains("F") Then
+                            e.NewData = "FALSE"
+                        Else
+                            e.NewData = "TRUE"
+                        End If
+                    End If
+                Else
                     MsgBox("Type TRUE/T to change to scientific display and FALSE/F back to numeric display", MsgBoxStyle.Information)
                     e.EndReason = unvell.ReoGrid.EndEditReason.Cancel
                     Exit Sub
--- a/instat/ucrColumnMetadata.vb
+++ b/instat/ucrColumnMetadata.vb
@@ -112,7 +112,7 @@ Public Class ucrColumnMetadata
         Dim iTemp As Integer
         Dim iNameColumn As Integer = -1
         Dim strNewValue As String
-        Dim strValAllowed As String() = {"TRUE", "T", "FALSE", "F"}
+        Dim strBooleanValsAllowed As String() = {"TRUE", "T", "FALSE", "F"}
 
         If e.NewData.ToString() = strPreviousCellText Then
             e.EndReason = unvell.ReoGrid.EndEditReason.Cancel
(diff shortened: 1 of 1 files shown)


instat/ucrColumnMetadata.vb
```diff
@@ -138,11 +138,20 @@ Public Class ucrColumnMetadata
             ElseIf strProperty = strLabelsScientific Then
                 strNewValue = e.NewData.ToString.ToUpper
 
-                If Not strValAllowed.Contains(strNewValue) Then
+                e.NewData = strNewValue
+                If strValAllowed.Contains(strNewValue) Then
```
lloyddewit, line 142: The name 'strValAllowed' is quite generic and sounds as if it contains the values allowed for any string. Would it be more readable to change the name to somethink like 'strBooleanValAllowed' or 'strScientificValAllowed'?

instat/ucrColumnMetadata.vb
```diff
@@ -138,11 +138,20 @@ Public Class ucrColumnMetadata
             ElseIf strProperty = strLabelsScientific Then
                 strNewValue = e.NewData.ToString.ToUpper
```
lloyddewit, line 139: `strNewValue` is set but it doesn't seem to be used in this subroutine. Could we remove this variable completely from this subroutine?

instat/ucrColumnMetadata.vb
```diff
@@ -138,11 +138,20 @@ Public Class ucrColumnMetadata
             ElseIf strProperty = strLabelsScientific Then
```
lloyddewit, line 138: Could this `ElseIf` block is be shortened? E.g.:
```
                e.NewData = e.NewData.ToString.ToUpper
                If strValAllowed.Contains(e.NewData) Then
                    If e.NewData(0) = "F" Then
                        e.NewData = "FALSE"
                    Else
                        e.NewData = "TRUE"
                    End If
```

instat/ucrColumnMetadata.vb
shadrackkibet: We can't remove this because it is being passed into the Rfunction in line 169 and 178. We sometimes want it with quotes and sometimes without quotes.

instat/ucrColumnMetadata.vb
shadrackkibet: I agree with your suggestion @lloyddewit that is more sensible. Thanks

instat/ucrColumnMetadata.vb
shadrackkibet: I get the idea to simplify this @lloyddewit. However, we still need to make sure we update `strNewValue` to avoid it being empty. So I just add that below your suggested code. i.e

```
                e.NewData = e.NewData.ToString.ToUpper
                If strBooleanValAllowed.Contains(e.NewData) Then
                    If e.NewData(0) = "F" Then
                        e.NewData = "FALSE"
                    Else
                        e.NewData = "TRUE"
                    End If
                    strNewValue = e.NewData
                 else

```

instat/ucrColumnMetadata.vb
lloyddewit: Yes, you are quite right. `strNewValue` on lines 169 and 178 was not visible on my screen and I forgot to scroll right. I should have checked more carefully! Sorry for that.

instat/ucrColumnMetadata.vb
lloyddewit: Yes, that's a good solution.

shadrackkibet:
@rdstern?

shadrackkibet:
@maxwellfundi this can be merged now.

lloyddewit:
@rdstern To pass your 'awkward' test case, for any combination of upper and lower case (e.g. `FaLs', 'tR' etc.), only one line would need to be changed:
```
115: Dim strBooleanValsAllowed As String() = {"T", "TR", "TRU", "TRUE", "F", "FA", "FAL", "FALS", "FALSE"}
```
Should this be done before @maxwellfundi merges?

P.S. if there are other places in R-Instat where the user enters true or false, and more sophistication is required, then we could use a fuzzy logic lib to handle other user typos (e.g. 'fslse').

shadrackkibet:
@lloyddewit  I think we will be complicating this at this stage. We will also need to add checks to convert this "awkward" cases to either `TRUE` or `FALSE` so that it is displayed in the cell.

lloyddewit:
> 
> 
> @lloyddewit I think we will be complicating this at this stage. We will also need to add checks to convert this "awkward" cases to either `TRUE` or `FALSE` so that it is displayed in the cell.

Only line 115 needs to change (I tested). The existing conversion on line 140 already works for the new 'awkward' cases:
```
                e.NewData = e.NewData.ToString.ToUpper
                If strBooleanValsAllowed.Contains(e.NewData) Then
                    If e.NewData(0) = "F" Then
                        e.NewData = "FALSE"
                    Else
                        e.NewData = "TRUE"
                    End If
```

shadrackkibet:
Yes, you are right, I see the reason why it works because we are checking using the first letter here `e.NewData(0)`. So for all those cases, it is converted. Thanks

rdstern:
@shadrackkibet and @lloyddewit this is great and the new "system" is working.  Shadrack if that is convincing and only a very small time is needed, then I suggest you make the change, before merging.  
I also want to use this to add another component.  Shadrack, you may have the feeling that you have spent enough time on this small task - and without yet another change.  And Stephen, you may soon find (as Danny often did) that discussing with the team was more time consuming than just making the change yourself.  

So, using this as an example, if you both felt it was more efficient, from this point, for Stephen to just make these changes - then you agee that Stephen should do that.  In that case I should have a final quick check - possibly anyway - before it is merged.

rdstern:
Oh, I am too late - it seems to be resolved.  I should add that I have the strong impression that Shadrack is learning well and also working pretty damn efficiently now.  I really like what I am seeing here!

lloyddewit:
@rdstern thank you for the suggestion. Generally I think the PR author should maintain control over the code changes. But if the PR author agrees and the change is small then it should be OK for the reviewer to save the PR author some time and make a small update.
@shadrackkibet As far as I can see, the change isn't made yet.  If you're OK with @rdstern 's suggestion then I can try and make this one-line change. It would be an interesting experiment for me to see if I have the necessary GitHub permissions.

lloyddewit:
@shadrackkibet Please ignore my comment above, Roger was right and you were too fast for me!

PR #5811, opened 2020-05-28
Reading of numeric values  in the PICSA Rainfall dialogue

Ivanluv:
@lloyddewit  this fixes the issue in pull request #5743


Wycklife:
@rdstern and @lloyddewit  I have tested this using the dodoma dataset. The selector now accepts only numeric variables when you're on the multiple receiver. This therefore seems to be fixed from my end.

shadrackkibet:
The code looks okay. Not sure if this fixes the original issue @rdstern please confirm and approve so that it can be merged.

shadrackkibet:
The code looks okay. Not sure if this fixes the original issue @rdstern please confirm and approve so that it can be merged.

PR #5825, opened 2020-06-09
Auto Switching of Receivers

### LINKED ISSUE #5508: Auto Switching of Single Receivers Focus on Selector Variables Selections
At the moment you have to manually change the focus of the single receiver before selecting the variable you want to use.

Changing focus of single receivers automatically on selection of variables in the selector

**Additional context**
This issue is in relation to the design changes suggested in issue #5480

Patowhiz:
Fixes #5508
@lloyddewit I hope my comments are suggestive enough to help other coders understand what the code additions are meant to do. I tried to follow the commenting style you use. Looking forward to any corrections.


Patowhiz:
I tested these changes using Line Plot and Column Summaries dialogs. @rdstern you may test this with I hope @rdstern and the team will test this with the other dialogs to check if the autoswitching will break. 

I noticed there are multiple controls inheriting from the receivers and these may need some slight minimal modifications to correctly autoswitch. I only modified those that I think needed modifications.

I added the addition of receivers in the Selector property based on @dannyparsons suggestion, this was the most viable way to add this change with very minimal required changes to the dialogs for this to work.

lloyddewit:
@shadrackkibet @Ivanluv I had a quick look at this and it seems good. But you should do your review first. Please could one of you peer review? thanks

shadrackkibet:
@Ivanluv can you review this? otherwise, I am happy to do it.

instat/ucrReceiver.vb
```diff
@@ -76,6 +76,9 @@ Public Class ucrReceiver
     ' Currently only implemented for multiple receiver.
     Public bForceVariablesAsList As Boolean = False
 
+    'used by the selector to autoswitch from the current receiver
+    Public bAutoSwitchFromReceiver As Boolean = True
```
shadrackkibet, line 80: I understand that we need to auto-switch only for single receivers. So I will prefer to have this as `False` by default and then we will not have to do anything on multiple receiver. We can then enable the feature only for the single receiver. What do you think?

instat/ucrSelector.vb
```diff
@@ -440,4 +449,62 @@ Public Class ucrSelector
...
+    ''' The receiver is only added if it does not exist in the list
+    ''' </summary>
+    ''' <param name="ucrNewReceiver"></param>
+    Public Overridable Sub AddReceiver(ucrNewReceiver As ucrReceiver)
+        AddReceiver({ucrNewReceiver})
+    End Sub
+
+    ''' <summary>
+    ''' Adds passed enumerable of receivers to the end of the receivers list
+    ''' The reciver is only added if it does not exist in the list
```
shadrackkibet, line 465: Minor typoo
```suggestion
    ''' The receiver is only added if it does not exist in the list

```

instat/ucrSelector.vb
```diff
@@ -440,4 +449,62 @@ Public Class ucrSelector
...
+    ''' changes the focus of the next receiver from passed receiver and sets it as the 'current receiver' of the selector
+    ''' </summary>
+    ''' <param name="FromSelectedReceiver"></param>
+    Private Sub AutoSwitchCurrentReceiver(FromSelectedReceiver As ucrReceiver)
+        Dim ucrNewCurrentReceiver As ucrReceiver
+        For i As Integer = 0 To lstOrderedReceivers.Count - 2
+            If lstOrderedReceivers.Item(i) Is FromSelectedReceiver Then
+                ucrNewCurrentReceiver = lstOrderedReceivers.Item(i + 1)
+                If ucrNewCurrentReceiver.Visible AndAlso
+                   ucrNewCurrentReceiver.Enabled Then
```
shadrackkibet, line 499: Should this be in one line to make it more readable?
```suggestion
                If ucrNewCurrentReceiver.Visible AndAlso ucrNewCurrentReceiver.Enabled Then

```

shadrackkibet: I wonder if at some point ` ucrNewCurrentReceiver` might be Nothing, then you might want to check for Nothing for safety?

shadrackkibet:
@rdstern @Patowhiz This is a general question on this feature. I think it will just need a small tweak on this implementation.

The current implementation goes through all single receivers even if it is already infilled. We infill receivers automatically in most climatic dialogs, or a user might have infilled some receivers before switching options then more receivers are now visible. So I wonder if we want to only auto switch to the next empty receiver? 

This is a nice feature, I see that this will enhance user experience.

Patowhiz:
> @rdstern @Patowhiz This is a general question on this feature. I think it will just need a small tweak on this implementation.
> 
> The current implementation goes through all single receivers even if it is already infilled. We infill receivers automatically in most climatic dialogs, or a user might have infilled some receivers before switching options then more receivers are now visible. So I wonder if we want to only auto switch to the next empty receiver?
> 
> This is a nice feature, I see that this will enhance user experience.

@rdstern, @shadrackkibet raises a very good question. My first assumption is the auto switch will strictly follow the order and in this case assume the use intends to "overwrite"  the contents of the receiver and thus switches to the next receiver regardless of whether it has been filled or not . I think this is a good general assumption, but what about the climatic dialogs? should we retain this functionality or adopt what @shadrackkibet proposes? 

I can see how we could have both functionalities by adding  the functionality of checking when it's been "system filled" or "user filled". A long shot but very doable now if its a priority.

instat/ucrReceiver.vb
Patowhiz: I agree, thanks

instat/ucrSelector.vb
Patowhiz: Thanks, I agree this should be in 1 line. 
`ucrNewCurrentReceiver` will never be nothing. The list holds "concrete" receivers only, and if  "nothing" was ever passed  
 that error ("a developer error") will be thrown when adding the "nothing" to the list (have a look at subroutine `AddReceiver(ucrNewReceiver As ucrReceiver)`  in line459)

lloyddewit:
@rdstern Please could you test and approve? thanks

instat/ucrVariablesAsFactor.vb
```diff
@@ -295,9 +295,19 @@ Public Class ucrVariablesAsFactor
...
         End Get
         Set(ucrNewSelector As ucrSelector)
+            'this by default adds the parent receiver to the list of receivers in the selector
             MyBase.Selector = ucrNewSelector
+            'we remove it here because the selector will manipulate the children receivers only
+            'todo. Currently the selector only uses the list of receiver to manipulate the focus switching, 
+            'in future the usage of the receivers could change which may require this implementation to be changed 
+            If MyBase.Selector IsNot Nothing Then
+                MyBase.Selector.RemoveReceiver(Me)
+            End If
```
lloyddewit, line 305: I don't really understand what's happening here. please could you explain a bit more (or we can briefly discuss)? thanks

instat/ucrSelector.vb
```diff
@@ -440,4 +449,63 @@ Public Class ucrSelector
...
+
+    ''' <summary>
+    ''' changes the focus of the next receiver from passed receiver and sets it as the 'current receiver' of the selector
+    ''' </summary>
+    ''' <param name="FromSelectedReceiver"></param>
+    Private Sub AutoSwitchCurrentReceiver(FromSelectedReceiver As ucrReceiver)
+        Dim ucrNewCurrentReceiver As ucrReceiver
+        For i As Integer = 0 To lstOrderedReceivers.Count - 2
+            If lstOrderedReceivers.Item(i) Is FromSelectedReceiver Then
+                ucrNewCurrentReceiver = lstOrderedReceivers.Item(i + 1)
```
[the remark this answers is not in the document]
Patowhiz: @lloyddewit from the user's perspective, I'm not sure if that's desirable. If I'm in the last receiver as a user, I think that means, I'm okay and done with items that I have filled from the first receiver to the last one (in most scenarios I think). If you strongly think that's not the case I'm happy to commit the suggested change.

@rdstern what's your opinion on this. Do you think most of the time the user would want the focus to automatically shift from the last receiver to the first receiver? At the moment, as @lloyddewit has correctly pointed out, the implementation is, the focus gets stuck on the last receiver until the user manually changes the focus to another receiver other than the last.

instat/ucrSelector.vb
```diff
@@ -440,4 +449,63 @@ Public Class ucrSelector
...
+                If ucrNewCurrentReceiver.Visible AndAlso ucrNewCurrentReceiver.Enabled Then
+
+                    ucrNewCurrentReceiver.Focus() 'sets current focus enabling correct tab navigation
+                    SetCurrentReceiver(ucrNewCurrentReceiver)
+                Else
+                    AutoSwitchCurrentReceiver(ucrNewCurrentReceiver)
```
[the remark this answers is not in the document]
Patowhiz: @lloyddewit this a good point that I had not thought about. I cannot see a scenario where all receivers are invisible and disabled. However, your point on 100% guaranteed certainly is a important. I agree that we should add a check for that. Will add the check.

instat/ucrVariablesAsFactor.vb
Patowhiz: @lloyddewit thank you for asking.

From the comments just above the line code
```
  'this by default adds the parent receiver to the list of receivers in the selector
   MyBase.Selector = ucrNewSelector
   'we remove it here because the selector will manipulate the children receivers only
   'todo. Currently the selector only uses the list of receiver to manipulate the focus switching, 
   'in future the usage of the receivers could change which may require this implementation to be changed 
    If MyBase.Selector IsNot Nothing Then
                MyBase.Selector.RemoveReceiver(Me)
     End If
```
What I mean here is, in this PR I had to override the code that sets the receivers selector. So calling 
` MyBase.Selector = ucrNewSelector`      
automatically adds the `ucrVariablesAsFactor`  which is `Me` as the receiver of `ucrNewSelector` . This is undesirable because ucrVariablesAsFactor is a receiver that has 2 receivers(it's children controls) which it uses to manipulate data . So I'm removing `ucrVariablesAsFactor` which is `Me` by calling `MyBase.Selector.RemoveReceiver(Me)` and leaving only it's children controls in the list of receivers used for autoswitching. I have a `todo` comment there because the list of receivers could be used for other things that may prompt this implementation to be changed 

I hope that makes it a bit clear. I'm happy to have a Skype call and discuss further.

instat/ucrSelector.vb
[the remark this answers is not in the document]
lloyddewit: @Patowhiz There is only a risk of an infinite loop if my previous comment is implemented (lines 503-505, circling through to set focus to first receiver in list). I suggest we wait until @rdstern responds to your question above. If he finds the current implementation OK then we don't need to check for an infinite loop here.

instat/ucrVariablesAsFactor.vb
```diff
@@ -295,9 +295,19 @@ Public Class ucrVariablesAsFactor
             Return MyBase.Selector
         End Get
         Set(ucrNewSelector As ucrSelector)
+            'this by default adds the parent receiver to the list of receivers in the selector
             MyBase.Selector = ucrNewSelector
+            'we remove it here because the selector will manipulate the children receivers only
+            'todo. Currently the selector only uses the list of receiver to manipulate the focus switching, 
+            'in future the usage of the receivers could change which may require this implementation to be changed
```
lloyddewit, line 302: Thank you for the clear explanation below, could we include this in the comments?
```suggestion
            'This line automatically adds ucrVariablesAsFactor (Me) as the receiver of ucrNewSelector. 
            'This is undesirable because ucrVariablesAsFactor is a receiver that has two receivers 
            '(its child controls) which it uses to manipulate data. 
            MyBase.Selector = ucrNewSelector
            'Therefore remove ucrVariablesAsFactor (Me) and leave only it's child controls in the 
            'list of receivers used for autoswitching.
            ' 
            'TODO Currently the selector only uses the list of receivers to manipulate the focus 
            'switching. In future the usage of the receivers could change which may require this 
            'implementation to be changed 
```

Patowhiz: Thanks for the elaborate suggestion

Patowhiz:
Added and committed all suggestions except the one awaiting @rdstern feedback.

rdstern:
@Patowhiz if I understand your query properly, then I checked also with Danny and we propose that there should be no move back to the first receiver.  Once it reaches the last receiver, then it stays there.

Patowhiz:
@lloyddewit from @rdstern comment above, then do you agree we should just retain my current implementation?

lloyddewit:
@Patowhiz yes, I agree, if no circular move back to the first receiver is needed, then your code is good. I approved.

@rdstern please could you re-approve? thanks

instat/ucrSelector.vb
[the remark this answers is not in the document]
Patowhiz: Based on @rdstern feedback this is now resolved.

lloyddewit:
@rdstern Please could you re-approve? thanks

rdstern (approved):

PR #5787, opened 2020-05-07
Added an additional validation property to ucrInput control

Wycklife:
Fixes #5762

In progress

PR #5826, opened 2020-06-09
Numeric Expression Validation In UcrInput

### LINKED ISSUE #5762: Validation of numeric expressions in input controls
At the moment a custom control like ucrTextbox can validate if a value is numeric or not, but if I type in , for instance, 80/2 which results to 40, it is unable to recognize this as an expression which could result to a numeric expression.  

For this kind of validation the current control needs to be enhanced.

Patowhiz:
Fixes #5762
This PR builds on additions made in PR #5787

I have tested the code changes using regular sequencer and random samples dialog. Please also test with other dialogs.

dlgRegularSequence

dlgRandomSample


lloyddewit:
@shadrackkibet Please could you review this first (it's related to one of your PRs so you're a good candidate)? thanks

instat/ucrInput.vb
```diff
@@ -337,7 +342,17 @@ Public Class ucrInput
...
         If strText <> "" AndAlso (strValuesToIgnore Is Nothing OrElse (strValuesToIgnore IsNot Nothing AndAlso Not strValuesToIgnore.Contains(strText))) Then
             If Not IsNumeric(strText) Then
-                iType = 1
+                iType = 1 'reset as invalid entry 
+                'if numeric expressions are allowed check the expression results to a valid numeric
+                If bNumericExpressionAllowed Then
+                    Dim vecOutput As CharacterVector
+                    'is.numeric(x) returns true if the x expression is a valid one. 
+                    'So we use it here to check validity of the entry
+                    vecOutput = frmMain.clsRLink.RunInternalScriptGetOutput("is.numeric(" & strText & ")", bSilent:=True)
```
shadrackkibet, line 351: I think the correct function to use here is `is.integer()`  rather than `is.numeric()` which checks for numeric values including decimals.

instat/ucrInput.vb
```diff
@@ -337,7 +342,17 @@ Public Class ucrInput
...
+                    If vecOutput IsNot Nothing AndAlso vecOutput.Length > 0 AndAlso Mid(vecOutput(0), 5).ToUpper = "TRUE" Then
```
shadrackkibet, line 352: Please ignore my previous comment deleted: 

I was wondering why it is necessary to check for length `vecOutput.Length > 0` here, isn't it always greater than 0. Could you please explain why it is needed?

Patowhiz: This is meant to ensure that `Mid(vecOutput(0), 5).ToUpper = "TRUE"` check will not through a runtime error incase the returned `CharacterVector` collection is empty. I asked myself the same question "isn't always greater than 0" when implementing this, and I wasn't sure of the answer, so to be safe I did that check (of course the alternative would have been a try catch but I found that to be an overkill). 

I'm happy to remove that check if what's returned will never be an empty collection.

shadrackkibet: Isn't that case checked by `vecOutput IsNot Nothing`? which you already have?

Patowhiz: `vecOutput IsNot Nothing` checks for `Nothing` or simply `NULL`. `vecOutput.Length > 0` checks for empty Collection. A collection can be not `NULL` but empty. By empty here I means it has no elements in it. For instance; take a "type of" collection like an `array`, it can have 0 length meaning empty but not Null, same applies to `List`. 

For clarity, notice I'm checking for `Nothing`(which is simply NULL), then empty(by checking the length), then if those 2 checks pass, I'm taking the first element (index 0) in the collection in this case `vecOutput(0)`, then I'm stripping away unwanted characters using the VB.NET `Mid` function, then I'm converting to upper case, after which I'm checking if the result is `TRUE`. If this check passes then the rest follows...

If I call `vecOutput(0)` when the collection is empty or rather zero length, then a runtime error will be thrown, so in this case `vecOutput IsNot Nothing` will pass but `vecOutput(0)` will be an error (accessing an element in an index position that does not exist).

I'm happy to remove that check if what's returned will never be an empty collection or rather a collection of zero length.

lloyddewit: @shadrackkibet @patowhiz Thank you for the question and nicely explained response. I am very happy to see the null and empty collection checks here. I would like to keep these checks even if `RunInternalScriptGetOutput` was expected to always return a non-empty collection. The number of checks is always a question of balance, but in my view, these checks make the code safer and more readable.

lloyddewit:
@rdstern You already approved above, please can you 'approve' again so this can be merged? thanks

PR #5828, opened 2020-06-11
Improvements to the Display Daily climatic dialogue

### LINKED ISSUE #5342: New possible plot for the climatic data
When we use Climatic > Check Data > Display daily

we currently have a table of the daily data.  But we don't have a good graph.  This is the openair::summaryPlot function with the data from one of the Ghana stations.

image

It (I assume) uses the lattice system, rather than ggplot and there are some things I am not so sure about.  But it is the sort of thing we should have.

I wonder if we might initially test this function more and then perhaps (initially?) include it.  We might later want to prepare our own version in ggplot.

Ivanluv:
fixes #5342


Ivanluv:
@rdstern @shadrackkibet  could you review this

shadrackkibet:
@rdstern I suggest you initially test this.

rdstern:
@Ivanluv this already looks very promising.
a) Danny was saying that the second line of code may not be needed?  In my example it was:
`date <- data_book$get_columns_from_data(data_name="ghana", col_names="date")`  maybe check this with @shadrackkibet ?
b) When using a single station (Dodoma) there were 4 elements and it should have included the "scales = free_y".  That should be the default in this dialogue (unless you only include a single element.)  Perhaps there could be a checkbox with the free y scale as the default.
c) The rugplot on the x-axis is designed to show where there are missing values.  That isn't trivial.  For the Ghana data @dannyparsons did this by changing the default rug to 
`geom_rug(data = ghana %>% filter(!is.na(value))`
d) The rugplot on the y-axis isn't very useful (except possibly for the rainfall.  It does no harm now, but later it might be interesting instead to investigate ggmarginal (in the ggextra package) to put a boxplot (or violin plot there instead)  Or nothing perhaps.
e) I suggest the default layout is to have the plots as wide as possible.  So they are in a single column.  I hope that is possible even when there are multiple stations?  Ideally again there might be  a layout control, but with the daily data we will usually want each plot to be as wide as possible.

Ivanluv:
> e) I suggest the default layout is to have the plots as wide as possible. So they are in a single column. I hope that is possible even when there are multiple stations? Ideally again there might be a layout control, but with the daily data we will usually want each plot to be as wide as possible.

Any ideas on how to implement this?

rdstern:
It is really coming on!  Well done.

Something like this?
image

To do that I changed your facet_wrap to facet_grid i.e.  `ggplot2::facet_grid(rows=vars(variable),scales="free_y")`

While I am there I would also not have a legend, because it uses space and repeats information in the graph itself.
And I suggest we don't need the x-axis label, because it is obviously the Date.

shadrackkibet:
@rdstern I think we do not the legend as well. That information displayed in the legend is already in the facets. So we could have more space by removing it.

Ivanluv:
@rdstern  can you look at this again

rdstern:
One simple small improvement could be to have the default colour of the rug plot different to the actual plot.  I think, for example, just putting colour="black" in the rug geom as the default would do.

The legend is still visible.  Is that difficult to remove?

Soon I would like to see this dialogue merged.  Once we use it, then I am sure that more controls could be added.  But quite useful to use it first.

Ivanluv:
@rdstern  could you look at this again

rdstern:
will do

instat/dlgDisplayDailyData.vb
```diff
@@ -216,11 +216,16 @@ Public Class dlgDisplayDailyData
...
+        clsYLabFunction = New RFunction
 
         clsXLabFunction.SetPackageName("ggplot2")
         clsXLabFunction.SetRCommand("xlab")
         clsXLabFunction.AddParameter("label", Chr(34) & "" & Chr(34),, iPosition:=0, bIncludeArgumentName:=False)
 
+        clsYLabFunction.SetPackageName("ggplot2")
+        clsYLabFunction.SetRCommand("ylab")
+        clsYLabFunction.AddParameter("label", Chr(34) & "" & Chr(34),, iPosition:=0, bIncludeArgumentName:=False)
+
```
shadrackkibet, line 228: @Ivanluv I think you might have misunderstood this. You need to remove the legend and not the y-axis label.

rdstern:
@Ivanluv good that the legend is gone.  However I couldn't get the nice display you had in skype.  Here was my dialogue:
image

And I still got a single set of plots:
image

I also wonder whether the usual Options button would be possible with this plot?

rdstern:
This now works very nicely!
Could you check quickly whether the "usual" Options button works?  Then I could change the x-axis which is a date variable for example - within the existing dialogue?
Otherwise I am ready to approve now.  There will be further enhancements, but they can be discussed later.  And they may not be needed if the Options button works!

Ivanluv:
@rdstern  the "Data options" is working from my end.

rdstern:
@Ivanluv I would hope that Data Options works, i.e. you can filter.  I was thinking of the Options button, by which I mean the Plotting options sub-dialogue.

Ivanluv:
@rdstern when using the plot options  subdialog some functionality may not be possible  eg changing the x-axis which is a date variable  within the existing dialogue,which is the idea you had.This is because we are initially stacking  all the columns  except the date and station hence the columns wont be available

rdstern:
@lloyddewit I hope that this might soon be ok to merge?

instat/dlgDisplayDailyData.vb
```diff
@@ -278,21 +374,70 @@ Public Class dlgDisplayDailyData
...
+            clsFacetOperator.AddParameter("station", ucrReceiverStations.GetVariableNames(False), bIncludeArgumentName:=False, iPosition:=1)
+            clsFacetFunction.RemoveParameterByName("rows")
+            clsFacetFunction.AddParameter("facets", clsROperatorParameter:=clsFacetOperator, iPosition:=1)
+            clsFacetFunction.AddParameter("space", Chr(34) & "fixed" & Chr(34), iPosition:=2)
+        Else
+            clsFacetFunction.RemoveParameterByName("facets")
+            clsFacetFunction.RemoveParameterByName("space")
+            clsFacetFunction.AddParameter("rows", "vars(variable)", iPosition:=1)
+            clsFacetFunction.RemoveParameterByName("facets")
+            clsFacetFunction.RemoveParameterByName("space")
```
lloyddewit, line 428: Why are these parameters removed twice?

instat/dlgDisplayDailyData.vb
```diff
@@ -278,21 +374,70 @@ Public Class dlgDisplayDailyData
...
+                ucrSelectorDisplayDailyClimaticData.ucrAvailableDataFrames.clsCurrDataFrame.RemoveParameterByName("measure.vars")
+                ucrSelectorDisplayDailyClimaticData.ucrAvailableDataFrames.clsCurrDataFrame.RemoveParameterByName("stack_data")
+                ucrSelectorDisplayDailyClimaticData.ucrAvailableDataFrames.clsCurrDataFrame.RemoveParameterByName("id.vars")
+            End If
+        End If
+    End Sub
+
+    Private Sub FacetControl()
+        If rdoGraph.Checked AndAlso Not ucrReceiverStations.IsEmpty Then
+            clsFacetOperator.AddParameter("station", ucrReceiverStations.GetVariableNames(False), bIncludeArgumentName:=False, iPosition:=1)
```
lloyddewit, line 419: The `station` parameter is not removed in the `else` block below. Is that correct?

instat/dlgDisplayDailyData.vb
```diff
@@ -278,21 +374,70 @@ Public Class dlgDisplayDailyData
...
+        FacetControl()
     End Sub
 
     Private Sub ucrSelectorDisplayDailyClimaticData_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSelectorDisplayDailyClimaticData.ControlValueChanged
         clsDisplayDailyGraphFunction.AddParameter("data_name", Chr(34) & ucrSelectorDisplayDailyClimaticData.ucrAvailableDataFrames.strCurrDataFrame & Chr(34), iPosition:=0)
+        StackingFunction()
+        FacetControl()
+    End Sub
+
+    Private Sub StackingFunction()
```
lloyddewit, line 403: Please could you add an XML comment block. Please follow the style of the examples in RCodeStructure, thanks
Note: from now on, I will ask developers to do this for all new functions. This will gradually improve the readability of the code. It makes sense for the developer who wrote the code to write the comment.

instat/dlgDisplayDailyData.vb
```diff
@@ -278,21 +374,70 @@ Public Class dlgDisplayDailyData
...
+                ucrSelectorDisplayDailyClimaticData.ucrAvailableDataFrames.clsCurrDataFrame.AddParameter("id.vars", clsRFunctionParameter:=clsIdVars)
+            Else
+                ucrSelectorDisplayDailyClimaticData.ucrAvailableDataFrames.clsCurrDataFrame.RemoveParameterByName("measure.vars")
+                ucrSelectorDisplayDailyClimaticData.ucrAvailableDataFrames.clsCurrDataFrame.RemoveParameterByName("stack_data")
+                ucrSelectorDisplayDailyClimaticData.ucrAvailableDataFrames.clsCurrDataFrame.RemoveParameterByName("id.vars")
+            End If
+        End If
+    End Sub
+
+    Private Sub FacetControl()
```
lloyddewit, line 417: Please could you add an XML comment block. Please follow the style of the examples in RCodeStructure, thanks

instat/dlgDisplayDailyData.vb
Ivanluv: yeah.This is wrong.Am removing this right away

instat/dlgDisplayDailyData.vb
Ivanluv: yes.Thats correct

instat/dlgDisplayDailyData.vb
lloyddewit: thanks. I suggest to modify to:
```
    ''' <summary> 
    ''' If the Graph radio button is checked, adds the rows, facets and space parameters to the 
    ''' facet function.
    ''' If the radio button is unchecked, it removes these parameters.
    ''' </summary>
```

instat/dlgDisplayDailyData.vb
lloyddewit: That's great, thanks. I suggest to modify to:
```
    ''' <summary> 
    ''' If the Graph radio button is checked, takes all the columns in the multiple receiver, 
    ''' excluding date and station columns, and stacks them.
    ''' If the radio button is unchecked, unstacks the columns in the multiple receiver.
    ''' If the multiple receiver is empty then this subroutine does nothing.
    ''' </summary>
```

lloyddewit:
@rdstern Please could you approve again? We can then merge, thanks

PR #5829, opened 2020-06-11
Added Filter from Factor subdialog 

### LINKED ISSUE #4800: New filter from factor sub-dialogue
David claims this would be quite easy for our AMI team to undertake, with a bit of help from him.

It would produce a set of filters for a data frame, that are each a particular level of a factor.  We can do this one-at-a-time from the current Filter sub-dialogue.  This new dialogue would produce them all together and give them each the name which is the label of that factor level.
The dialogue could either be next to the current Filter dialogue in the **Prepare > Data Frame** menu, (called Filter by Factor) or it could be in the **Prepare > Column: Factor Menu**, and simply be called Filter.

The dialogue would have our usual data selector.  Then it would look a bit like the Levels/Labels dialogue.  But when you select a factor the main control could perhaps be a copy of the control that is currently in the Filter sub-dialogue so you could select whichever levels you want. And that control has a special button to **select all levels**.  In this case that might be the default.

Wycklife:
Fixes #4800

@africanmathsinitiative/developers This is work in progress. Designer part is done, currently implementing the code.

### DIFF
--- a/instat/dlgRestrict.resx
+++ b/instat/dlgRestrict.resx
@@ -119,14 +119,10 @@
   </resheader>
   <assembly alias="System.Drawing" name="System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
   <data name="cmdDefineNewFilter.Location" type="System.Drawing.Point, System.Drawing">
-    <value>339, 117</value>
-  </data>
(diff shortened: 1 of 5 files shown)


lloyddewit:
@Wycklife  Please could you clarify the status? thanks

lloyddewit:
@Wycklife Thanks for the new commits, is this now ready for test/review?

Wycklife:
@rdstern  I'm working on this now.

Wycklife:
I'm closing this pr since the branch have conflicts

PR #5840, opened 2020-06-14
Initial Implementation of homogenization 

### LINKED ISSUE #5461: Homogenization of climatic data
I have now started thinking about the implementation of `homogen` function from `climatol` package in R-Instat.

I suggest this as a special dialog in the climatic menu. I am yet to come up with the design of the dialog. This is still at an early stage as the Rcode is not yet ready.

**Describe alternatives you've considered**
There exist several methods to homogenize climatic data. `Climatol` R package implements the famous **SNHT** which is a widely accepted. Good to check other R packages as well.

**Additional context**
I have tried using this function in `R` initially, applied this on KMD dataset. A suggestion is to ask the author of `climatol` package to add an option to pick a dataframe, this will make it easy for us to implement this in R-Instat. Currently, this function has lot's of outputs which as printed in **pdf** format. Not sure if we want all this. I think we just need the homogenized column. This is currently saved as a separate file `.dah`, it will be good if this is saved as a column in the original dataframe. This function also prints some summaries, we might want this printed in the output window. There are other separate files saved in the directory.

Below is my code, at the moment this is just for viewing, I will edit it soon and supply sample files so that it is reproducible.

```R
#loading required libraries
library(climatol)
library(tidyverse)
library(reshape2)
library(magrittr)

setwd("F:\\AIMS Cameroon\\AIMS Essay\\data")
rainfall_monthly <- read_csv("rainfall_monthly_data_kmd.csv")
est.c <-  read_csv("synoptic_stations.csv")

#reshape data into the required shape.
dat <- dcast(data = rainfall_monthly, formula = year + month_val ~ station_id, value.var = "sum_Rain")
dat %<>% select(V1="8635000",V2="8639000",V3="8641000",V4="8737000",V5="8834098",V6="8840000",V7="8934096",V8="8935115",V9="8935181",V10="8937022",V11="8937065",V12="9034025",V13="9034088",V14="9034103",V15="9035279",V16="9035318",  
                V17="9036135",V18="9036261",V19="9036288",V20="9037202",V21="9037262",V22="9039000",V23="9135001",V24="9136087",V25="9136130",  
                V26="9136164",V27="9136168",V28="9136208",V29="9136269",V30="9137048",V31="9137089",V32="9237000",V33="9240001",V34="9338001",  
                V35="9339036",V36="9340007",V37="9340009",V38="9439021")

dat <- as.matrix(dat)

#writing required files
write(dat,'pcp_1973-2018.dat')
write.table(est.c, 'pcp_1973-2018.est', row.names = FALSE, col.names = FALSE)
homogen('pcp', 1973, 2018, std = 2, nm = 12) #homogenization

#loading homogenized data
View(import("F:\\AIMS Cameroon\\AIMS Essay\\data\\pcp_1973-2018.rda",which = 'dah'),title = 'homogenized_data')

```

shadrackkibet:
Fixes #5461. @rdstern this contains an initial implementation of homogenization. This implements the three core functions from `changepoint` package i.e `cpt.mean(), cpt.var` and `cpt.meanvar()`. I did initial testing using the datasets from the `changepoint` package although some of these datasets are not well labelled. I also used Dodoma data at some point to check how these methods work in our area of application (Climate). But first I aggregate the data into Annual using climatic summaries for the reasons I outline below. So for this case, I obtained annual rainfall totals for testing.

My understanding is that homogenization of daily and sub-daily data is still an active area of research therefore applying homogenization methods is still not recommended. This is because daily data is generally "noisy" and also daily data is mostly used to study extremes which make detecting changepoints difficult because it is not easy to distinguish extremes and inhomogeneities. I am still looking into the to be released WMO guide you shared to check if they now have recommendations based on recent advancements. In the previous guide, they only made recommendations on aggregated data.

I, therefore, envisage that the main use of this dialog will be on aggregated data, monthly, annual, seasonal, dekadal etc. However, the user is not restricted to apply these methods into daily and sub-daily data.

I made `BinSeg` the default method of detecting inhomogeneities because this method allows for multiple changepoints detection defined by `Q` which looks sensible in climatology. A previous check with Rija (CDT) is that he was planning to change his algorithms in CDT to use this method.

Below are some output results;
```

Class 'cpt' : Changepoint Object
       ~~   : S4 class containing 14 slots with names
              cpts.full pen.value.full data.set cpttype method test.stat pen.type pen.value minseglen cpts ncpts.max param.est date version 

Created on  : Sat Sep 21 05:12:37 2019 

summary(.)  :
----------
Created Using changepoint version 2.2.2 
Changepoint type      : Change in mean 
Method of analysis    : BinSeg 
Test Statistic  : Normal 
Type of penalty       : MBIC with value, 13.03 
Minimum Segment Length : 1 
Maximum no. of cpts   : 5 
Changepoint Locations : 11 12 13 20 49 
Range of segmentations:
     [,1] [,2] [,3] [,4] [,5]
[1,]   13   NA   NA   NA   NA
[2,]   13   12   NA   NA   NA
[3,]   13   12   11   NA   NA
[4,]   13   12   11   20   NA
[5,]   13   12   11   20   49

 For penalty values: 48032 48032 48032 48032 38714 
```
image

### DIFF
--- a/instat/dlgHomogenization.vb
+++ b/instat/dlgHomogenization.vb
@@ -90,7 +90,6 @@ Public Class dlgHomogenization
         dctMethodOptions.Add("BinSeg", Chr(34) & "BinSeg" & Chr(34))
         ucrInputComboMethod.SetItems(dctMethodOptions)
         ucrInputComboMethod.SetDropDownStyleAsNonEditable()
-        ucrInputComboMethod.SetRDefault(Chr(34) & "BinSeg" & Chr(34))
 
         ucrInputComboDistribution.SetParameter(New RParameter("test.stat", 3))
         dctDistributionOptions.Add("Normal", Chr(34) & "Normal" & Chr(34))
@@ -120,7 +119,7 @@ Public Class dlgHomogenization
         ucrSaveResult.SetPrefix("Result")
         ucrSaveResult.SetAssignToIfUncheckedValue("last_result")
 
-        ucrInputComboPenalty.AddToLinkedControls(ucrInputPenValue, {"Asymptotic"}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrInputComboPenalty.AddToLinkedControls(ucrInputPenValue, {"Asymptotic", "CROPS"}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
         ucrInputComboMethod.AddToLinkedControls(ucrInputQ, {"BinSeg"}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
         ucrPnlOptions.AddToLinkedControls(ucrReceiverNeighbour, {rdoNeighbouring}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
         ucrReceiverNeighbour.SetLinkedDisplayControl(lblNeighbouring)
@@ -149,6 +148,7 @@ Public Class dlgHomogenization
         clsCptMeanFunction.AddParameter("data", clsRFunctionParameter:=clsExcludeNAFunction, iPosition:=0)
         clsCptMeanFunction.AddParameter("Q", 5, iPosition:=5)
         clsCptMeanFunction.AddParameter("pen.value", 0, iPosition:=6)
+        clsCptMeanFunction.AddParameter("method", Chr(34) & "BinSeg" & Chr(34), iPosition:=2)
 
         clsCptVarianceFunction.SetPackageName("changepoint")
         clsCptVarianceFunction.SetRCommand("cpt.var")
@@ -165,6 +165,7 @@ Public Class dlgHomogenization
         clsSummaryFunction.SetRCommand("summary")
 
         ucrBase.clsRsyntax.ClearCodes()
+        AddPlotSummaryParameters()
         ucrBase.clsRsyntax.SetBaseRFunction(clsCptMeanFunction)
     End Sub
 
@@ -217,14 +218,22 @@ Public Class dlgHomogenization
     Private Sub ucrPnlMethods_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlMethods.ControlValueChanged
         If rdoCptMean.Checked Then
             ucrBase.clsRsyntax.SetBaseRFunction(clsCptMeanFunction)
+        ElseIf rdoCptVariance.Checked Then
+            ucrBase.clsRsyntax.SetBaseRFunction(clsCptVarianceFunction)
+        ElseIf rdoMeanVariance.Checked Then
+            ucrBase.clsRsyntax.SetBaseRFunction(clsCptMeanVarianceFunction)
+        End If
+        AddPlotSummaryParameters()
+    End Sub
+
+    Private Sub AddPlotSummaryParameters()
+        If rdoCptMean.Checked Then
             clsPlotFunction.AddParameter("x", clsRFunctionParameter:=clsCptMeanFunction, iPosition:=0)
             clsSummaryFunction.AddParameter("object", clsRFunctionParameter:=clsCptMeanFunction, iPosition:=0)
         ElseIf rdoCptVariance.Checked Then
-            ucrBase.clsRsyntax.SetBaseRFunction(clsCptVarianceFunction)
             clsPlotFunction.AddParameter("x", clsRFunctionParameter:=clsCptVarianceFunction, iPosition:=0)
             clsSummaryFunction.AddParameter("object", clsRFunctionParameter:=clsCptVarianceFunction, iPosition:=0)
         ElseIf rdoMeanVariance.Checked Then
-            ucrBase.clsRsyntax.SetBaseRFunction(clsCptMeanVarianceFunction)
             clsPlotFunction.AddParameter("x", clsRFunctionParameter:=clsCptMeanVarianceFunction, iPosition:=0)
             clsSummaryFunction.AddParameter("object", clsRFunctionParameter:=clsCptMeanVarianceFunction, iPosition:=0)
         End If
--- a/instat/dlgHomogenization.vb
+++ b/instat/dlgHomogenization.vb
@@ -19,7 +19,7 @@ Imports instat.Translations
 Public Class dlgHomogenization
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
-    Private clsCptMeanFunction, clsCptVarianceFunction, clsCptMeanVarianceFunction, clsExcludeNAFunction, clsPlotFunction, clsSummaryFunction, clsSnhtFunction, clsPettittFunction As New RFunction
+    Private clsCptMeanFunction, clsCptVarianceFunction, clsCptMeanVarianceFunction, clsExcludeNAFunction, clsPlotFunction, clsSummaryFunction, clsSnhtFunction, clsPettittFunction, clsBuishandFunction As New RFunction
     Private Sub dlgHomogenization_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         autoTranslate(Me)
         If bFirstLoad Then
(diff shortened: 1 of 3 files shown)


shadrackkibet:
@rdstern this is now enabled under climatic>check data>homogenization. I would like you to do some initial testing for this.

shadrackkibet:

lloyddewit:
@Patowhiz Please could you peer review? thanks

shadrackkibet:
> I am approving, because there will later need to be changes, but it is already a dialogue where testing a merged version would be very useful.
> 
> In the important snht option the default is to save a new data frame, while all the other options save the result. This raises a number of questions. If @shadrack can fix them today, or @lloyddewit has serious queries on the code then fine, but otherwise I would prefer to merge as it is, subject to these improvements being considered as a new issue.
> a) In shnt it always saves a new data frame, should this be optional? I don't understand the content.
> b) The name of the new data frame is snht_dataframe1. I would prefer it to relate to the initial data frame, e.g. dodoma_snht1.
> c) Should the option to save result be there as well?
> d) I suggest (unless I have misunderstood) that the default should be to produce a summary and not to save the data frame?

All these depend on the package I am using. I think it is easier to understand snht from `trend` package which returns an htest rather than one from `snht` package which returns a data frame. However, I am concerned that in the WMO guide on homogenization they recommend this option.

I will switch to the `trend` option for now.

instat/dlgHomogenization.vb
```diff
@@ -14,8 +14,325 @@
...
+            clsPlotFunction.AddParameter("x", clsRFunctionParameter:=clsSnhtFunction, iPosition:=0)
+            clsSummaryFunction.AddParameter("object", clsRFunctionParameter:=clsSnhtFunction, iPosition:=0)
+        ElseIf rdoPettitt.Checked Then
+            clsPlotFunction.AddParameter("x", clsRFunctionParameter:=clsPettittFunction, iPosition:=0)
+            clsSummaryFunction.AddParameter("object", clsRFunctionParameter:=clsPettittFunction, iPosition:=0)
+        ElseIf rdoBuishand.Checked Then
+            clsPlotFunction.AddParameter("x", clsRFunctionParameter:=clsBuishandFunction, iPosition:=0)
+            clsSummaryFunction.AddParameter("object", clsRFunctionParameter:=clsBuishandFunction, iPosition:=0)
+        End If
+    End Sub
```
Patowhiz, line 309: @shadrackkibet After looking at the entire dialog code. My understanding is that `AddPlotSummaryParameters()` is meant to simply set the base function as the parameter of the  _after codes_ (`clsPlotFunction`, `clsSummaryFunction`). 

Since this subroutine is also called in `ucrPnlMethods.ControlValueChanged` event which involves checking which checkbox is checked, this leads to double checking of checked checkbox which I think isn't necessary

My suggestion then would be implementing this subroutine to something like

```
Private Sub AddPlotSummaryParameters(clsSelectedBaseFunction As New RFunction )
    clsPlotFunction.AddParameter("x", clsRFunctionParameter:=clsSelectedBaseFunction, iPosition:=0)
    clsSummaryFunction.AddParameter("object", clsRFunctionParameter:=clsSelectedBaseFunction, iPosition:=0)
End Sub
```

Then when calling it, you just need to pass in the the current set base function. 
This removes the need of double checking of the checkbox checked since it's already done when setting the base function and I think will not change the functionality.

Of course a more radical implementation would be not having the subroutine parameter (`clsSelectedBaseFunction As New RFunction` ) and just accessing the base function from `ucrBase.clsRsyntax` but then we don't have a `GetBaseFunction()` function and also it may not be immediately clear what the base function is at a glance.

instat/dlgHomogenization.vb
```diff
@@ -14,8 +14,325 @@
...
+        ucrPnlOptions.AddFunctionNamesCondition(rdoNeighbouring, {"cpt.mean", "cpt.var", "cpt.meanvar", "snh.test", "pettitt.test", "br.test"}, False)
+        ucrPnlOptions.AddFunctionNamesCondition(rdoMultiple, {"cpt.mean", "cpt.var", "cpt.meanvar", "snh.test", "pettitt.test", "br.test"}, False)
+
+        ucrChkPlot.SetText("Plot")
+        ucrChkPlot.AddRSyntaxContainsFunctionNamesCondition(True, {"plot"})
+        ucrChkPlot.AddRSyntaxContainsFunctionNamesCondition(False, {"plot"}, False)
+
+        ucrChkSummary.SetText("Summary")
+        ucrChkSummary.AddRSyntaxContainsFunctionNamesCondition(True, {"summary"})
+        ucrChkSummary.AddRSyntaxContainsFunctionNamesCondition(False, {"summary"}, False)
```
Patowhiz, line 82: @shadrackkibet this is more of a question than a suggestion. cause I'm not sure of how its been internally implemented, I thought by just calling 
`ucrChkSummary.AddRSyntaxContainsFunctionNamesCondition(True, {"summary"})`, that effectively  sets even the false condition, since a checkbox has just 2 states; checked(True) or unchecked(False). Which means 
 `ucrChkSummary.AddRSyntaxContainsFunctionNamesCondition(False, {"summary"}, False)` , may be redundant?

If that's not the case, then @lloyddewit I think this will be good to be "specially" implemented for the checkbox control.

instat/dlgHomogenization.vb
```diff
@@ -14,8 +14,325 @@
...
+        End If
+    End Sub
+
+    Private Sub ucrChkSummary_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkSummary.ControlValueChanged
+        If ucrChkSummary.Checked Then
+            ucrBase.clsRsyntax.AddToAfterCodes(clsSummaryFunction, iPosition:=0)
+            clsSummaryFunction.iCallType = 2
+        Else
+            ucrBase.clsRsyntax.RemoveFromAfterCodes(clsSummaryFunction)
+        End If
```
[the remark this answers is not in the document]
shadrackkibet: Agreed!

instat/dlgHomogenization.vb
shadrackkibet: @Patowhiz that for your suggestion here and above. 

I like your radical implementation I choose to do that, it is straight forward and simplifies code. For me, it is clear what the base function is since that is set when we do `ucrBase.clsRsyntax.SetBaseRFunction()` and I can directly access the set base function using `ucrBase.clsRsyntax.clsBaseFunction` or did you mean we should have a function `GetBaseFunction()` which returns the set base function in the `Rsyntax` class? If that is the case I see that we will need to refactor code elsewhere because I think it will be appropriate to change `clsBaseFunction` to private in the `Rsyntax` class.

Patowhiz: @shadrackkibet yes that's what I meant. And I wasn't quite sure if directly accessing that variable might have unintended consequences (in future), but if that's what you prefer, I have no strong views against it. It's good that we are aware of such possibility.

instat/dlgHomogenization.vb
shadrackkibet: Apparently, that works but haven't checked how it is implemented internally. The second line is added to make the set conditions explicit when reading the code. Without the second line, it is not clear to me that `false` condition is set! 

Happy to remove it if the first line is clear to someone else. And I would suggest we always not include it and gradually remove it in other places to be consistent!

lilyclements:
> Also important is that you used an ordinary plot. I assume this could change to be a ggplot. And could it also include the date column as the x. This may not be simple R. I wonder if @lilyclements can help if you need help here?

I have had a look into doing this by `ggplot`:

image

image

Only issue I'm having is with the `date` variable. I assume from the code this only runs where the element is complete, however this gives a variable with an extra value than the `TK` variable. Should it start from the second date in the data set, or finish at the penultimate date, or is it a bit more complicated than that?

R Code:
```
test.dat <- snh.test(x=na.exclude(object=dodoma$rain), m=10)

data <- data.frame(Tk = test.dat$data,
                               Date = 1:length(test.dat$data))

ggplot(data = data, aes(x = Date, y = Tk)) + geom_line() +
  labs(title = "Standard Normal Homogeneity Test (SNHT)")
```

shadrackkibet:
@rdstern @lloyddewit this PR is slightly over a month old now. I am of the opinion that this should be merged because what is here is already useful. I know there is still a lot **TODO** which I will work on preferably as new PR.

PR #5847, opened 2020-06-21
Verification of forecasts and satellite data - Summaries dialog

### LINKED ISSUE #5843: New dialogue called Verification Summary
This builds on the current verification facilities.  It uses the ideas of the summary dialogue to be able to produce the verification summaries for any combination of factor levels, including station and season.  The other difference is that (like the existing summaries) the results will usually be saved in a new data frame.

There will now be 2 dialogues.  We suggest Compare be moved to its own section below the one for start, end, extremes. Compare: Calculation.  Then the new one is Compare: Summary.  

This has the same structure as the top of the existing dialogue (no radio buttons though).  Under the second receiver is another multiple receiver - like the ordinary summary dialogue, i.e. for Factors and labellled the same.

As with the ordinary summary dialogue there is then a Summary button.  

Some of the other options of the ordinary summary dialogue might also be present, though it is always saved to the summary level.

What is not so clear is the structure of the sub-dialogue.  It may have 3 (or 4) tabs, though perhaps at least initially, there could just be a single tab.  There would then be 3 (or 4) blocks with labels:

Continuous, Categorical, Binary and possibly Probabilistic

The Continuous would combine the summaries from the HydroGOF and Verification, while the others would all come from the Verification.  

There will later be a tricky bit to detect the type of data in the single receivers, so the summaries are appropriate.  Perhaps that code can be added later?

shadrackkibet:
Fixes #5843. This is still work in progress. Continuous options working now. Coming soon!

### DIFF
--- a/instat/dlgCompareSummary.vb
+++ b/instat/dlgCompareSummary.vb
@@ -79,6 +79,7 @@ Public Class dlgCompareSummary
         clsSummaryFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$calculate_summary")
         clsSummaryFunction.AddParameter("obs.type", Chr(34) & "'cont'" & Chr(34), iPosition:=1)
         clsSummaryFunction.AddParameter("summaries", clsRFunctionParameter:=clsListFunction, iPosition:=6)
+        clsListFunction.AddParameter("summary_cor", Chr(34) & "summary_cor" & Chr(34), bIncludeArgumentName:=False, iPosition:=20)
 
         clsListFunction.SetRCommand("c")
 
@@ -94,10 +95,11 @@ Public Class dlgCompareSummary
         sdgVerificationSummaries.SetRFunction(clsNewSummaryFunction:=clsSummaryFunction, clsNewListFunction:=clsListFunction, bReset:=bResetSubdialog)
         sdgVerificationSummaries.ShowDialog()
         bResetSubdialog = False
+        TestOkEnabled()
     End Sub
 
     Private Sub TestOkEnabled()
-        If ucrReceiverSatellite.IsEmpty OrElse ucrReceiverStation.IsEmpty Then
+        If ucrReceiverSatellite.IsEmpty OrElse ucrReceiverStation.IsEmpty OrElse clsListFunction.clsParameters.Count = 0 Then
             ucrBase.OKEnabled(False)
         Else
             ucrBase.OKEnabled(True)
--- a/instat/dlgCompareSummary.vb
+++ b/instat/dlgCompareSummary.vb
@@ -120,8 +120,9 @@ Public Class dlgCompareSummary
 
     Private Sub ucrPnlObservationType_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlObservationType.ControlValueChanged
         SetSelectorDataTypes()
-        ucrReceiverStation.Clear()
         ucrReceiverSatellite.Clear()
+        ucrReceiverStation.Clear()
+        ucrReceiverStation.SetMeAsReceiver()
         If rdoContinuous.Checked Then
             iTabIndexSelected = 0
             clsSummaryFunction.AddParameter("frcst.type", Chr(34) & "'cont'" & Chr(34), iPosition:=5)
@@ -154,10 +155,10 @@ Public Class dlgCompareSummary
             ucrReceiverStation.SetIncludedDataTypes({"numeric"}, bStrict:=True)
             ucrReceiverStation.strSelectorHeading = "Numerics"
         ElseIf rdoBinary.Checked Then
-            ucrReceiverSatellite.SetIncludedDataTypes({"logical", "factor"}, bStrict:=True)
-            ucrReceiverSatellite.strSelectorHeading = "Logical and Factors"
-            ucrReceiverStation.SetIncludedDataTypes({"logical", "factor"}, bStrict:=True)
-            ucrReceiverStation.strSelectorHeading = "Logical and Factors"
+            ucrReceiverSatellite.SetIncludedDataTypes({"logical", "numeric"}, bStrict:=True)
+            ucrReceiverSatellite.strSelectorHeading = "Logical and Numerics"
+            ucrReceiverStation.SetIncludedDataTypes({"logical", "numeric"}, bStrict:=True)
+            ucrReceiverStation.strSelectorHeading = "Logical and Numeric"
         ElseIf rdoCategorical.Checked Then
             ucrReceiverSatellite.SetIncludedDataTypes({"logical", "factor"}, bStrict:=True)
             ucrReceiverSatellite.strSelectorHeading = "Factors"
--- a/instat/dlgCompareSummary.vb
+++ b/instat/dlgCompareSummary.vb
@@ -115,6 +115,8 @@ Public Class dlgCompareSummary
 
     Private Sub ucrPnlObservationType_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlObservationType.ControlValueChanged
         SetSelectorDataTypes()
+        ucrReceiverStation.Clear()
+        ucrReceiverSatellite.Clear()
         If rdoContinuous.Checked Then
             iTabIndexSelected = 0
             clsSummaryFunction.AddParameter("frcst.type", Chr(34) & "'cont'" & Chr(34), iPosition:=5)
--- a/instat/dlgCompareSummary.vb
+++ b/instat/dlgCompareSummary.vb
@@ -101,7 +101,7 @@ Public Class dlgCompareSummary
         sdgVerificationSummaries.SetRFunction(clsNewSummaryFunction:=clsSummaryFunction, clsNewListFunction:=clsListFunction, iNewTabIndexSelected:=iTabIndexSelected, bReset:=bResetSubdialog)
         sdgVerificationSummaries.ShowDialog()
         bResetSubdialog = False
-        AddUseParameter()
+        AddRemoveUseParameter()
         TestOkEnabled()
     End Sub
 
@@ -166,7 +166,7 @@ Public Class dlgCompareSummary
         ucrBase.clsRsyntax.iCallType = If(ucrChkPrintOutput.Checked, 2, 0)
     End Sub
 
-    Private Sub AddUseParameter()
+    Private Sub AddRemoveUseParameter()
         If ucrChkIgnoreMissing.Checked AndAlso clsListFunction.ContainsParameter("summary_cor") Then
             clsSummaryFunction.AddParameter("use", Chr(34) & "'complete.obs'" & Chr(34), iPosition:=8)
         Else
@@ -175,7 +175,7 @@ Public Class dlgCompareSummary
     End Sub
 
     Private Sub ucrChkIgnoreMissing_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkIgnoreMissing.ControlValueChanged
-        AddUseParameter()
+        AddRemoveUseParameter()
     End Sub
 
     Private Sub ucrReceiverStation_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverStation.ControlContentsChanged, ucrReceiverSatellite.ControlContentsChanged, ucrChkPrintOutput.ControlContentsChanged, ucrChkStoreResults.ControlContentsChanged
--- a/instat/dlgCompareSummary.vb
+++ b/instat/dlgCompareSummary.vb
@@ -101,6 +101,7 @@ Public Class dlgCompareSummary
         sdgVerificationSummaries.SetRFunction(clsNewSummaryFunction:=clsSummaryFunction, clsNewListFunction:=clsListFunction, iNewTabIndexSelected:=iTabIndexSelected, bReset:=bResetSubdialog)
         sdgVerificationSummaries.ShowDialog()
         bResetSubdialog = False
+        AddUseParameter()
         TestOkEnabled()
     End Sub
 
@@ -171,14 +172,16 @@ Public Class dlgCompareSummary
         ucrBase.clsRsyntax.iCallType = If(ucrChkPrintOutput.Checked, 2, 0)
     End Sub
 
-    Private Sub ucrChkIgnoreMissing_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkIgnoreMissing.ControlValueChanged
-        If ucrChkIgnoreMissing.Checked Then
-            If sdgVerificationSummaries.ucrContinuousVerification.ucrChkCorrelations.Checked Then
-                clsSummaryFunction.AddParameter("use", Chr(34) & "'complete.obs'" & Chr(34), iPosition:=8)
-            End If
+    Private Sub AddUseParameter()
+        If ucrChkIgnoreMissing.Checked AndAlso clsListFunction.ContainsParameter("summary_cor") Then
+            clsSummaryFunction.AddParameter("use", Chr(34) & "'complete.obs'" & Chr(34), iPosition:=8)
         Else
             clsSummaryFunction.RemoveParameterByName("use")
         End If
+    End Sub
+
+    Private Sub ucrChkIgnoreMissing_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkIgnoreMissing.ControlValueChanged
+        AddUseParameter()
     End Sub
 
     Private Sub ucrReceiverStation_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverStation.ControlContentsChanged, ucrReceiverSatellite.ControlContentsChanged, ucrChkPrintOutput.ControlContentsChanged, ucrChkStoreResults.ControlContentsChanged
--- a/instat/dlgCompareSummary.vb
+++ b/instat/dlgCompareSummary.vb
@@ -1,5 +1,116 @@
-﻿Public Class dlgCompareSummary
-    Private Sub dlgCompareSummary_Load(sender As Object, e As EventArgs) Handles MyBase.Load
+﻿' R- Instat
+' Copyright (C) 2015-2017
+'
+' This program is free software: you can redistribute it and/or modify
+' it under the terms of the GNU General Public License as published by
+' the Free Software Foundation, either version 3 of the License, or
+' (at your option) any later version.
+'
+' This program is distributed in the hope that it will be useful,
+' but WITHOUT ANY WARRANTY; without even the implied warranty of
+' MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+' GNU General Public License for more details.
+'
+' You should have received a copy of the GNU General Public License 
+' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports instat.Translations
+Public Class dlgCompareSummary
+    Private bFirstLoad As Boolean = True
+    Private bReset As Boolean = True
+    Private bResetSubdialog = False
+    Private clsSummaryFunction As New RFunction
+    Private clsListFunction As New RFunction
+    Private Sub dlgCompareSummary_Load(sender As Object, e As EventArgs) Handles MyBase.Load
+        autoTranslate(Me)
+        If bFirstLoad Then
+            InitialiseDialog()
+            bFirstLoad = False
+        End If
+        If bReset Then
+            SetDefaults()
+        End If
+        SetRcodeForControls(bReset)
+        bReset = False
+        TestOkEnabled()
+    End Sub
+
+    Private Sub InitialiseDialog()
+        ucrSelectorVerificationSummary.SetParameter(New RParameter("data_name", 0))
+        ucrSelectorVerificationSummary.SetParameterIsString()
+
+        ucrPnlObservationType.SetParameter(New RParameter("obs.type", 1))
+        ucrPnlObservationType.AddRadioButton(rdoContinuous, Chr(34) & "'cont'" & Chr(34))
+        ucrPnlObservationType.AddRadioButton(rdoCategorical, Chr(34) & "'cat'" & Chr(34))
+        ucrPnlObservationType.AddRadioButton(rdoBinary, Chr(34) & "'binary'" & Chr(34))
+
+        ucrReceiverStation.SetParameter(New RParameter("columns_to_summarise", 2))
+        ucrReceiverStation.Selector = ucrSelectorVerificationSummary
+        ucrReceiverStation.SetParameterIsString()
+
+        ucrReceiverSatellite.SetParameter(New RParameter("pred", 3))
+        ucrReceiverSatellite.Selector = ucrSelectorVerificationSummary
+        ucrReceiverSatellite.SetParameterIsString()
+
+        ucrReceiverMultipleFactors.SetParameter(New RParameter("factors", 4))
+        ucrReceiverMultipleFactors.Selector = ucrSelectorVerificationSummary
+        ucrReceiverMultipleFactors.strSelectorHeading = "Factors"
+        ucrReceiverMultipleFactors.SetParameterIsString()
+        ucrReceiverMultipleFactors.SetIncludedDataTypes({"factor"}, bStrict:=True)
+    End Sub
+
+    Private Sub SetDefaults()
+        clsSummaryFunction = New RFunction
+        clsListFunction = New RFunction
+        'data_book$calculate_summary(columns_to_summarise="tmax", data_name="dodoma", factors=c("year"), y="tmin", additional_filter=day_filter, j=1, summaries=c("summary_cor", "summary_count_non_missing", "summary_count", "summary_sum"), silent=TRUE)
+        ucrSelectorVerificationSummary.Reset()
+        ucrReceiverStation.SetMeAsReceiver()
+
+        clsSummaryFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$calculate_summary")
+        clsSummaryFunction.AddParameter("obs.type", Chr(34) & "'cont'" & Chr(34), iPosition:=1)
+        clsSummaryFunction.AddParameter("summaries", clsRFunctionParameter:=clsListFunction, iPosition:=6)
+
+        clsListFunction.SetRCommand("c")
+
+        bResetSubdialog = True
+        ucrBase.clsRsyntax.SetBaseRFunction(clsSummaryFunction)
+    End Sub
+
+    Private Sub SetRcodeForControls(bReset As Boolean)
+        SetRCode(Me, ucrBase.clsRsyntax.clsBaseFunction, bReset)
+    End Sub
+
+    Private Sub cmdSummaries_Click(sender As Object, e As EventArgs) Handles cmdSummaries.Click
+        sdgVerificationSummaries.SetRFunction(clsNewListFunction:=clsListFunction, bReset:=bResetSubdialog)
+        sdgVerificationSummaries.ShowDialog()
+        bResetSubdialog = False
+    End Sub
+
+    Private Sub TestOkEnabled()
+        If ucrReceiverSatellite.IsEmpty OrElse ucrReceiverStation.IsEmpty Then
+            ucrBase.OKEnabled(False)
+        Else
+            ucrBase.OKEnabled(True)
+        End If
+    End Sub
+
+    Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
+        SetDefaults()
+        SetRcodeForControls(True)
+        TestOkEnabled()
+    End Sub
+
+    Private Sub ucrPnlObservationType_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlObservationType.ControlValueChanged
+        If rdoContinuous.Checked Then
+            clsSummaryFunction.AddParameter("frcst.type", Chr(34) & "'cont'" & Chr(34), iPosition:=5)
+        ElseIf rdoBinary.Checked Then
+            clsSummaryFunction.AddParameter("frcst.type", Chr(34) & "'binary'" & Chr(34), iPosition:=5)
+        ElseIf rdoCategorical.Checked Then
+            clsSummaryFunction.AddParameter("frcst.type", Chr(34) & "'cat'" & Chr(34), iPosition:=5)
+        End If
+    End Sub
+
+    Private Sub ucrReceiverStation_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverStation.ControlContentsChanged, ucrReceiverSatellite.ControlContentsChanged
+        TestOkEnabled()
     End Sub
 End Class
--- a/instat/dlgCompareSummary.vb
+++ b/instat/dlgCompareSummary.vb
@@ -128,6 +128,16 @@ Public Class dlgCompareSummary
         End If
     End Sub
 
+    Private Sub ucrChkIgnoreMissing_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkIgnoreMissing.ControlValueChanged
+        If ucrChkIgnoreMissing.Checked Then
+            If sdgVerificationSummaries.ucrContinuousVerification.ucrChkCorrelations.Checked Then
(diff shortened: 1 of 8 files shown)


shadrackkibet:
@rdstern this is now working. Please do some initial testing?

rdstern:
Great - will do!

shadrackkibet:
@dannyparsons any comments on the structure of the dialog? Here are the screenshots.

image

image

image
image

shadrackkibet:
@dannyparsons any comments on the structure of the dialog? Here are the screenshots.

image

image

image
image

shadrackkibet:
Thanks for the clarification. I will do this change in another open pull request which contain frmmain changes.

rdstern:
This looks great.  On the main dialogue could you change the 2 receivers to items that are more general.  I showed it to Danny and he suggested Observed: and Estimated:.

shadrackkibet:
@rdstern that is now sorted out.

shadrackkibet:
I am going to update this branch with the master then you will be able to see it in the new position.

shadrackkibet:
It should be available in the new position now. Thanks

shadrackkibet:
@rdstern the feature you just requested is now implemented, please check.

shadrackkibet:
@rdstern above issues you reported are now fixed. Please check.

**Known Bug:**
When you are on a particular option say **Binary** make sure nothing is check on either **continuous** or **categorical** tab. Otherwise, you will get a bug because those summaries are added yet they do not apply to that option. 
One way I thought to fix the problem is by unchecking any checkbox that is disabled but that might inconvenience users.

rdstern:
That known bug makes it quite difficult to check all options!

a) It always writes a data frame.  I think with the results in the output window that should be instead of the new data frame.
b) In categorical it only (I think) allows factors, while binary allows either logical or factors.  Could categorical also allow logical, then I can check item c) there also.
c) In Dodoma I made logical variables from tmin>17 and tmax > 29.  Then I found that binary seems to give sensible results for the logical, but not for same data as a factor, see below:

```
# Code generated by the dialog, Verification Summaries
data_book$calculate_summary(data_name="dodoma", obs.type="'binary'", columns_to_summarise="tmax29a", y="tmin17a", return_output=TRUE, frcst.type="'binary'", summaries=c("PODy", "TS", "ETS", "FAR", "HSS", "PC", "BIAS", "EDS", "SEDS", "EDI", "SEDI"), na.rm=TRUE, j=1)

  PODy_tmax29a TS_tmax29a ETS_tmax29a FAR_tmax29a HSS_tmax29a PC_tmax29a
1       0.7195     0.5187      0.2255      0.3498       0.368     0.6832
  BIAS_tmax29a EDS_tmax29a SEDS_tmax29a EDI_tmax29a SEDI_tmax29a
1        1.106      0.3871        0.293      0.5229       0.5071

# Code generated by the dialog, Verification Summaries
data_book$calculate_summary(data_name="dodoma", obs.type="'binary'", columns_to_summarise="tmax29", y="tmin17", return_output=TRUE, frcst.type="'binary'", summaries=c("PODy", "TS", "ETS", "FAR", "HSS", "PC", "BIAS", "EDS", "SEDS", "EDI", "SEDI"), na.rm=TRUE, j=1)

  PODy_tmax29 TS_tmax29 ETS_tmax29 FAR_tmax29 HSS_tmax29 PC_tmax29 BIAS_tmax29
1           1         1          0          0          0         1           1
  EDS_tmax29 SEDS_tmax29 EDI_tmax29 SEDI_tmax29
1         -1          -1        NaN         NaN
```
d) I accidentally still had correlations for binary (from continuous).  It gave them!  Maybe it should be an additional summary for 2*2 tables?

shadrackkibet:
@rdstern the known bug is now fixed! also (a) and (b) reported above. After allowing logical variables on categorical, I checked the bug (c) you reported and both gave same results. 

For the binary case it doesn't give same results like you have reported, they do not allow factor variables for binary case in verifications package. Surprisingly numeric variables specifically of two "levels" i.e `0,1` is allowed. That is perhaps treated as a specific case internally. I therefore suggest that we do not allow factor variables for binary case but instead allow numeric. If needed elsewhere, I think we should add an option to hide numeric variables that are non 0-1 then that will be appropriate here.

shadrackkibet:
@rdstern 
(a) Agreed and fixed. 
(b) It is a general issue with the summary system, the same issue is with Column summaries and climatic summaries dialogs. So I suggest a separate issue to fix that.
(c) When both checkboxes **print results to output** and **store results** are unchecked, OK should not be enabled. I notice it is enabled in climatic summaries, it should be disabled.

lloyddewit:
@Ivanluv @Patowhiz Please could one of you peer review this. @rdstern has already tested the functionality so your review may focus on the code. I can then complete the final steps. Thanks!

instat/dlgCompareSummary.vb
```diff
@@ -1,5 +1,191 @@
...
+        SetSelectorDataTypes()
+        ucrReceiverSatellite.Clear()
+        ucrReceiverStation.Clear()
+        ucrReceiverStation.SetMeAsReceiver()
+        If rdoContinuous.Checked Then
+            iTabIndexSelected = 0
+            clsSummaryFunction.AddParameter("frcst.type", Chr(34) & "'cont'" & Chr(34), iPosition:=5)
+            sdgVerificationSummaries.tbBinary.Enabled = False
+            sdgVerificationSummaries.tbCategorical.Enabled = False
+            sdgVerificationSummaries.tbContinuous.Enabled = True
```
Patowhiz, line 131: @shadrackkibet I'm assuming `sdgVerificationSummaries` is a new subdialog created by you in this PR. If so, I suggest creating subroutines for setting the Enable properties of the 3 tabs (`tbBinary`, `tbCategorical`, `tbContinuous`). Accessing controls of a window(form) is not a good practice. If those controls are changed in future then also this dialog's functionality will have to be rewritten because refactoring may not be possible. I know we have several places in R-Instat where controls of another form are accessed directly, still that code may have to be refactored in future to conform with good principle of **OOP**. Again this is purely a suggestion and doesn't need to be implement.

instat/ucrContinuousVerification.vb
```diff
@@ -1,5 +1,157 @@
...
+                If ucrChkSelectAll.Checked Then
+                    ucrChkSelectAll.Checked = False
+                End If
+            End If
+        Next
+
+        ucrInputJmia.SetRCode(clsNewSummaryFunction, bReset)
+
+        If bReset Then
+            ucrChkSelectAll.Checked = False
```
Patowhiz, line 124:
```suggestion
            ucrChkSelectAll.Checked = Not bReset 
```

Patowhiz: Removes the need for the if block

instat/ucrContinuousVerification.vb
```diff
@@ -1,5 +1,157 @@
...
+        clsHelp.AddParameter("package", Chr(34) & strPackageName & Chr(34))
+        clsHelp.AddParameter("help_type", Chr(34) & "html" & Chr(34))
+        frmMain.clsRLink.RunScript(clsHelp.ToScript, strComment:="Opening help page for" & " " & strPackageName & " " & "Package. Generated from subdialog" & " " & ParentForm.Text, iCallType:=2, bSeparateThread:=False, bUpdateGrids:=False)
+    End Sub
+
+    Private Sub ucrChkSelectAll_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkSelectAll.ControlValueChanged
+        If ucrChkSelectAll.Checked Then
+            For Each ctrTemp In lstContinuousCheckboxes
+                If Not ctrTemp.Checked Then
+                    ctrTemp.Checked = True
```
Patowhiz, line 144:
```suggestion
                    ctrTemp.Checked = Not ctrTemp.Checked
```

instat/ucrContinuousVerification.vb
```diff
@@ -1,5 +1,157 @@
...
+                End If
```
Patowhiz, line 145: If Not ctrTemp.Checked Then
                    ctrTemp.Checked = True
             End If

@shadrackkibet correct me if I've misunderstood this, looking at this code block, I think it's meant to ensure that all the checkboxes in the list are checked? if so, why the the need to check if they are not checked? I find it more direct to just call `ctrTemp.Checked = True` regardless of whether the checkbox is checked or not. I think it's better in terms of clarity and I think the .NET JIT compiler may optimise this by doing just that.

If you agree with my thoughts, there are other similar instances in this file that you may want to do a similar change.

instat/dlgCompareSummary.vb
```diff
@@ -1,5 +1,191 @@
...
+            ucrReceiverSatellite.SetIncludedDataTypes({"logical", "factor"}, bStrict:=True)
+            ucrReceiverSatellite.strSelectorHeading = "Factors"
+            ucrReceiverStation.SetIncludedDataTypes({"logical", "factor"}, bStrict:=True)
+            ucrReceiverStation.strSelectorHeading = "Factors"
+        End If
+    End Sub
+
+    Private Sub ucrChkPrintOutput_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkPrintOutput.ControlValueChanged
+        If ucrChkPrintOutput.Checked Then
+            ucrBase.clsRsyntax.iCallType = 2
```
Patowhiz, line 172:
```suggestion
            ucrBase.clsRsyntax.iCallType = If(ucrChkPrintOutput.Checked, 2, 0)
```

Patowhiz: Removes the need for the whole block

instat/sdgVerificationSummaries.vb
```diff
@@ -13,9 +13,94 @@
...
+        If bControlsInitialised Then
+            InitialiseControls()
+            bControlsInitialised = False
+        End If
+        clsListFunction = clsNewListFunction
+
+        For Each ctr In lstSummaryCheckboxes
+            ctr.SetRCode(clsNewListFunction, bReset)
+            If ctr.Enabled = False AndAlso ctr.Checked Then
+                ctr.Checked = False
```
Patowhiz, line 91:
```suggestion
                ctr.Checked = (ctr.Enabled = False AndAlso ctr.Checked)
```

Patowhiz, line 91:
```suggestion
                ctr.Checked = (ctr.Enabled = False AndAlso ctr.Checked)
```

instat/dlgCompareSummary.vb
Patowhiz: Removes the need for the whole block

instat/sdgVerificationSummaries.vb
Patowhiz: Removes the need for the whole if block

instat/dlgCompareSummary.vb
shadrackkibet: I agree with you. I will do it in a flexible way. Currently, this subdialogs is used in one dialog. I know this will also cause problems if this subdialog is used in different dialogs (Classes).

instat/ucrContinuousVerification.vb
shadrackkibet: Perhaps I didn't understand what you mean by "removes the need for the if block". `Not bReset` will be essentially `False` in this case. If `bReset` is `False` this won't get to `ucrChkSelectAll.Checked = Not bReset`.

instat/ucrContinuousVerification.vb
shadrackkibet: Yes, it is meant to check all checkboxes in the list. But there is no need to check those checkboxes that are already checked by default or checked by the user using **Select all** checkbox.

instat/dlgCompareSummary.vb
shadrackkibet: Is it correct to assume that you mean we just have `ucrBase.clsRsyntax.iCallType = If(ucrChkPrintOutput.Checked, 2, 0)` inside the `Sub` ? and not 
```
If ucrChkPrintOutput.Checked Then
    ucrBase.clsRsyntax.iCallType = If(ucrChkPrintOutput.Checked, 2, 0)
End if
```
from your suggestion above? actually, if you change your suggestion I can commit it directly from Github without having to change it from VS.

shadrackkibet: Thanks, @Patowhiz although this is not what is done in most places (the later is), I am happy to change it because it reduces the lines of code.

Patowhiz: @shadrackkibet yes I meant so and yes it's purely to reduce lines of code. In the whole sub. I don't know how to do a github code suggestion to a whole block of code(multiple code lines), I only know how to do for a single line.

instat/ucrContinuousVerification.vb
Patowhiz: @shadrackkibet sorry for not being clear on this, I meant

```
If bReset Then
   ucrChkSelectAll.Checked =  False
End if
```

can be 

`ucrChkSelectAll.Checked = Not bReset`

shadrackkibet: This change will cause trouble with retaining the correct state of `ucrChkSelectAll` checkbox. So I still prefer to remain with the if statement. I only want to uncheck this checkbox when `bReset` is `True`.

instat/ucrContinuousVerification.vb
Patowhiz: @shadrackkibet what I meant is I think it's not really necessary to add the if check. I agree there is no need to check those checkboxes that are already checked, but checking an already checked checkbox or rather doing  `ctrTemp.Checked = True` in this case, won't cause any change to the CheckSate of the checkBox thus the if statement in this case is not really necessary.

I'm happy to let this  be the way it is if it's a bit confusing, I think the .NET JIT compiler may just optimise this.

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -1174,10 +1174,109 @@ VE <- function(x, y, na.rm = FALSE, na_type = "", ...){
...
+  return(A$EDI)  
+}
+
+##Symmetric Exremal dependence index
+SEDI <- function(x, y, frcst.type, obs.type, ...){
+  A <- verification::verify(obs = x, pred = y,  frcst.type = frcst.type, obs.type = obs.type)
+  return(A$SEDI)  
+}
+
+##Probabilistic/binary?
```
Patowhiz, line 1277: @shadrackkibet is this a future todo?

shadrackkibet: Yes, these options will be enabled in future when the need arises. I will  add a comment to make it clear.

instat/dlgCompareSummary.vb
```diff
@@ -1,5 +1,191 @@
...
+    End Sub
+
+    Private Sub ucrChkIgnoreMissing_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkIgnoreMissing.ControlValueChanged
+        If ucrChkIgnoreMissing.Checked Then
+            If sdgVerificationSummaries.ucrContinuousVerification.ucrChkCorrelations.Checked Then
+                clsSummaryFunction.AddParameter("use", Chr(34) & "'complete.obs'" & Chr(34), iPosition:=8)
+            End If
+        Else
+            clsSummaryFunction.RemoveParameterByName("use")
+        End If
```
Patowhiz, line 185: @shadrackkibet for clarity, do we have a scenario where `ucrChkIgnoreMissing` will be checked but `sdgVerificationSummaries.ucrContinuousVerification.ucrChkCorrelations` will be unchecked (I haven't seen this in the code)? And if so won't there be a need to also remove the "use" parameter if it exists in the `clsSummaryFunction`. 

Looking at this code block 

```
If 98*- .Checked Then
     If sdgVerificationSummaries.ucrContinuousVerification.ucrChkCorrelations.Checked Then
             clsSummaryFunction.AddParameter("use", Chr(34) & "'complete.obs'" & Chr(34), iPosition:=8)
     End If
Else
       clsSummaryFunction.RemoveParameterByName("use")
End If
```

It appears to me that the parameter "use" is only removed when `ucrChkIgnoreMissing` is checked but not when `sdgVerificationSummaries` is unchecked. 

--My suggestion if my below assumption is correct--
If parameter "use" should only be present when `sdgVerificationSummaries` and `clsSummaryFunction` are both checked then my suggestion is the code block can be changed to

```
If ucrChkIgnoreMissing.Checked AndAlso 
                             sdgVerificationSummaries.ucrContinuousVerification.ucrChkCorrelations.Checked  Then
        clsSummaryFunction.AddParameter("use", Chr(34) & "'complete.obs'" & Chr(34), iPosition:=8)
Else
       clsSummaryFunction.RemoveParameterByName("use")
End If
```
which ensures that  "use" is always removed when both conditions are not met.

instat/sdgVerificationSummaries.vb
```diff
@@ -13,9 +13,94 @@
...
+
+        lstSummaryCheckboxes.AddRange({ucrChkHitRate, ucrChkThreatScore, ucrChkEquitableThreatScore, ucrChkFalseAlarmRatio, ucrChkHeidkeSkillScore, ucrChkPercentCorrect, ucrChkBias, ucrChkExtremeDependencyScore, ucrChkSymmetricExtremeDependencyScore, ucrChkExremalDependencyIndex, ucrChkSymmetricExremalDependenceIndex, ucrChkCatPercentCorrect, ucrChkCatHeidkeSkillScore, ucrChkCatPierceSkillScore, ucrChkGerrityScore})
+
+        For Each ctr In lstSummaryCheckboxes
+            ctr.SetParameterIncludeArgumentName(False)
+            ctr.SetRDefault(Chr(34) & Chr(34))
+        Next
+    End Sub
+
+    Public Sub SetRFunction(clsNewSummaryFunction As RFunction, clsNewListFunction As RFunction, Optional bNewEnableDisableTab As Boolean = False, Optional iNewTabIndexSelected As Integer = 0, Optional bReset As Boolean = False)
```
Patowhiz, line 81: @shadrackkibet I'm not seeing where the subroutine parameter `bNewEnableDisableTab ` is used inside  the subroutine. Is it added as "feature" that could be used in future?

instat/ucrContinuousVerification.vb
Patowhiz: Sorry. Yes I agree. You can mark this as resolved.

instat/dlgCompareSummary.vb
shadrackkibet: Yes, there is a scenario. Notice that `ucrChkIgnoreMissing` is passing `na.rm` parameter which is required even if `sdgVerificationSummaries.ucrContinuousVerification.ucrChkCorrelations` is unchecked. However, I agree with the point you raise above `use` parameter is only needed when `ucrChkCorrelations` is checked and `ucrChkIgnoreMissing` is checked. I will implement this with a slight modification of what you suggest above.

instat/sdgVerificationSummaries.vb
shadrackkibet: You are right. I will remove it.

instat/dlgCompareSummary.vb
shadrackkibet: I implemented this with a slight change in your suggestion above. Instead of three subroutines, I have only one subroutine that set Enable property for the three tabs.

lloyddewit:
@Patowhiz thank you for the peer review. Are all your comments resolved? May @rdstern  and I proceed with final test and review? thanks

Patowhiz:
@shadrackkibet Thank you for the responses. @lloyddewit I'm satisfied with the changes you can take over.

instat/ucrContinuousVerification.vb
```diff
@@ -1,5 +1,157 @@
...
+    Private Sub cmdHelp_Click(sender As Object, e As EventArgs) Handles cmdHelp.Click
+        Dim clsHelp As New RFunction
+        Dim strPackageName As String
+
+        strPackageName = "hydroGOF"
+        clsHelp.SetPackageName("utils")
+        clsHelp.SetRCommand("help")
+        clsHelp.AddParameter("package", Chr(34) & strPackageName & Chr(34))
+        clsHelp.AddParameter("help_type", Chr(34) & "html" & Chr(34))
+        frmMain.clsRLink.RunScript(clsHelp.ToScript, strComment:="Opening help page for" & " " & strPackageName & " " & "Package. Generated from subdialog" & " " & ParentForm.Text, iCallType:=2, bSeparateThread:=False, bUpdateGrids:=False)
```
lloyddewit, line 137: Replace with the following?
```
        frmMain.clsRLink.RunScript(clsHelp.ToScript, strComment:="Opening help page for " & strPackageName & " Package. Generated from subdialog " & ParentForm.Text, iCallType:=2, bSeparateThread:=False, bUpdateGrids:=False)
```

instat/ucrContinuousVerification.vb
```diff
@@ -1,5 +1,157 @@
...
+                If Not ctrTemp.Checked Then
+                    ctrTemp.Checked = True
+                End If
+            Next
+            ucrChkSelectAll.SetText("Unselect all")
+        Else
+            For Each ctrTemp In lstContinuousCheckboxes
+                If ctrTemp.Checked Then
+                    ctrTemp.Checked = False
+                End If
```
lloyddewit, line 152: Same comment as above

instat/dlgCompareSummary.vb
```diff
@@ -1,5 +1,184 @@
...
+        clsSummaryFunction = New RFunction
+        clsListFunction = New RFunction
+
+        ucrSelectorVerificationSummary.Reset()
+        ucrReceiverStation.SetMeAsReceiver()
+
+        clsSummaryFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$calculate_summary")
+        clsSummaryFunction.AddParameter("obs.type", Chr(34) & "'cont'" & Chr(34), iPosition:=1)
+        clsSummaryFunction.AddParameter("summaries", clsRFunctionParameter:=clsListFunction, iPosition:=6)
+        clsListFunction.AddParameter("summary_cor", Chr(34) & "summary_cor" & Chr(34), bIncludeArgumentName:=False, iPosition:=20)
```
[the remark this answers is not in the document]
shadrackkibet: This is the position of the parameter in the `list` function which contain all summaries. Actually, they are over 30!

instat/dlgCompareSummary.vb
```diff
@@ -1,5 +1,184 @@
...
+            ucrReceiverStation.SetIncludedDataTypes({"numeric"}, bStrict:=True)
+            ucrReceiverStation.strSelectorHeading = "Numerics"
+        ElseIf rdoBinary.Checked Then
+            ucrReceiverSatellite.SetIncludedDataTypes({"logical", "numeric"}, bStrict:=True)
+            ucrReceiverSatellite.strSelectorHeading = "Logical and Numerics"
+            ucrReceiverStation.SetIncludedDataTypes({"logical", "numeric"}, bStrict:=True)
+            ucrReceiverStation.strSelectorHeading = "Logical and Numeric"
+        ElseIf rdoCategorical.Checked Then
+            ucrReceiverSatellite.SetIncludedDataTypes({"logical", "factor"}, bStrict:=True)
+            ucrReceiverSatellite.strSelectorHeading = "Factors"
```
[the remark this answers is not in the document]
shadrackkibet: Should be that!

instat/sdgVerificationSummaries.vb
```diff
@@ -13,9 +13,100 @@
...
+        ucrChkExtremeDependencyScore.SetText("Extreme Dependency Score")
+
+        ucrChkSymmetricExtremeDependencyScore.SetParameter(New RParameter("SEDS", 29), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "SEDS" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
+        ucrChkSymmetricExtremeDependencyScore.SetText("Symmetric Extreme Dependency Score")
+
+        ucrChkExremalDependencyIndex.SetParameter(New RParameter("EDI", 30), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "EDI" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
+        ucrChkExremalDependencyIndex.SetText("Exremal Dependency Index")
+
+        ucrChkSymmetricExremalDependenceIndex.SetParameter(New RParameter("SEDI", 31), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "SEDI" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
+        ucrChkSymmetricExremalDependenceIndex.SetText("Symmetric Exremal Dependence Index")
```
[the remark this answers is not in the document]
shadrackkibet: Yes.

instat/sdgVerificationSummaries.vb
```diff
@@ -13,9 +13,100 @@
...
+
+        ucrChkCatPercentCorrect.SetParameter(New RParameter("pc", 32), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "pc" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
+        ucrChkCatPercentCorrect.SetText("Percent Correct")
+
+        ucrChkCatHeidkeSkillScore.SetParameter(New RParameter("hss", 33), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "hss" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
+        ucrChkCatHeidkeSkillScore.SetText("Heidke Skill Score")
+
+        ucrChkCatPierceSkillScore.SetParameter(New RParameter("pss", 34), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "pss" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
+        ucrChkCatPierceSkillScore.SetText(" Pierce Skill Score")
```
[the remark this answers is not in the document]
shadrackkibet: yes

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -1174,10 +1174,109 @@ VE <- function(x, y, na.rm = FALSE, na_type = "", ...){
...
+  return(A$eds)  
+}
+
+##Symmetric extreme dependency score
+SEDS <- function(x, y, frcst.type, obs.type, ...){
+  A <- verification::verify(obs = x, pred = y,  frcst.type = frcst.type, obs.type = obs.type)
+  return(A$seds)  
+}
+
+##Exremal dependency index
```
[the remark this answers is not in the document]
shadrackkibet: yes

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -1174,10 +1174,109 @@ VE <- function(x, y, na.rm = FALSE, na_type = "", ...){
...
+EDI <- function(x, y, frcst.type, obs.type, ...){
+  A <- verification::verify(obs = x, pred = y,  frcst.type = frcst.type, obs.type = obs.type)
+  return(A$EDI)  
+}
+
+##Symmetric Exremal dependence index
```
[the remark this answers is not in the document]
shadrackkibet: yes

instat/ucrContinuousVerification.vb
```diff
@@ -1,5 +1,157 @@
...
+        ucrChkRatioOfRootMeanSquaredError.SetText("Ratio of root mean square error")
+
+        ucrChkRatioOfStandardDeviation.SetParameter(New RParameter("rSD", 7), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "rSD" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
+        ucrChkRatioOfStandardDeviation.SetText("Ratio of standard deviations")
+
+        ucrChkNashSutcliffeEfficiency.SetParameter(New RParameter("NSE", 8), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "NSE" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
+        ucrChkNashSutcliffeEfficiency.SetText("Nash-Sutcliffe efficiency")
+
+        ucrChkModNashSutcliffeEff.SetParameter(New RParameter("mNSE", 9), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "mNSE" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
+        ucrChkModNashSutcliffeEff.SetText("Modified Nash-Sutcliffe effieciency")
```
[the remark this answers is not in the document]
shadrackkibet: yes

instat/dlgCompareSummary.vb
shadrackkibet: If I understood you correctly, you are asking for where the parameter passed by `ucrChkIgnoreMissing` added and removed? If that is correct then check line 62 where the parameter is set and then this is passed to the base function by setting Rcode in line 97.

instat/ucrContinuousVerification.vb
shadrackkibet: Since this doesn't change the functionality I will remove it here and elsewhere. But I am not convinced why an already checked checkbox should be checked again or an unchecked is unchecked again!

instat/dlgCompareSummary.vb
```diff
@@ -1,5 +1,184 @@
...
+    Private Sub cmdSummaries_Click(sender As Object, e As EventArgs) Handles cmdSummaries.Click
+        sdgVerificationSummaries.SetRFunction(clsNewSummaryFunction:=clsSummaryFunction, clsNewListFunction:=clsListFunction, iNewTabIndexSelected:=iTabIndexSelected, bReset:=bResetSubdialog)
+        sdgVerificationSummaries.ShowDialog()
+        bResetSubdialog = False
+        AddRemoveUseParameter()
+        TestOkEnabled()
+    End Sub
+
+    Private Sub TestOkEnabled()
+        If ucrReceiverSatellite.IsEmpty OrElse ucrReceiverStation.IsEmpty OrElse clsListFunction.clsParameters.Count = 0 OrElse Not (ucrChkStoreResults.Checked OrElse ucrChkPrintOutput.Checked) Then
```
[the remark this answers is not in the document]
shadrackkibet: The two are not the same. `Not ucrChkStoreResults.Checked OrElse Not ucrChkPrintOutput.Checked` means both checkboxes have to be checked for OK to be enabled. By putting `Not` outside i.e `Not (ucrChkStoreResults.Checked OrElse ucrChkPrintOutput.Checked)` it means if one is checked then Ok is enabled, which is what we want here. 

This is a poor explanation, I hope it helps

instat/ucrContinuousVerification.vb
shadrackkibet: ok

instat/sdgVerificationSummaries.vb
```diff
@@ -13,9 +13,100 @@
...
+            InitialiseControls()
+            bControlsInitialised = False
+        End If
+        clsListFunction = clsNewListFunction
+
+        For Each ctr In lstSummaryCheckboxes
+            ctr.SetRCode(clsNewListFunction, bReset)
+            If ctr.Enabled = False AndAlso ctr.Checked Then
+                ctr.Checked = False
+            End If
```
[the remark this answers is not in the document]
shadrackkibet: I remove this since it doesn't affect functionality, but I am not convinced as pointed above.

instat/dlgCompareSummary.vb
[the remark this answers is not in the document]
lloyddewit: yes you are correct

instat/dlgCompareSummary.vb
lloyddewit: @shadrackkibet sorry for my confusing comment (I copied and pasted the wrong variable name). Please ignore it. I intended to ask about `ucrChkCorrelations`. But I now understand what you've done, and it looks fine.

lloyddewit:
@rdstern You already approved this before. The changes triggered by the peer review should not have affected the software logic so the risk of regression is small. Please could you approve again so we can merge? thanks

PR #5848, opened 2020-06-22
dlgStringHandling

### LINKED ISSUE #5830: String handling design
The design of this dialog doesn't look good. The labels are not well aligned and the size, spacing of the receiver and inputs, are unnecessarily wide. There is also an empty space on the right which can be substantially minimized by reducing the size of the top radio buttons.

image

lilyclements:

Fixes #5830 

Changes
- radio buttons along the top slightly less wide so the dialog width can decrease
- running a sub to change the dialog size
- small minor changes to straighten up and align the design
- tab order
- changed name of a private sub (was called "visibile()" which I assumed is a typo for "visible()". "visible()" caused a small warning so changed it to "visiblerdo()")


instat/dlgStringHandling.vb
```diff
@@ -289,7 +289,7 @@ Public Class dlgStringHandling
             Me.Size = New Size(iFullWidth, Me.Height)
         Else
             grpRegex.Visible = False
-            Me.Size = New Size(iFullWidth / 1.57, Me.Height)
+            Me.Size = New Size(536, Me.Height)
```
shadrackkibet, line 292: I am concerned about this kind of change. Calculation of width and height by a factor e.g `iFullWidth / 1.57` is meant to enable resizing of dialogs to adapt well in computers with different resolutions. I think we should stick with that rather than fixing height  e.g `536`.

lilyclements: Good to know, that makes a lot of sense. I will fix that now - thanks!

lilyclements: Sorted it

shadrackkibet:
Please add the previous and the current designer look.

lilyclements:
Before:

image


After:
image

Note that it does extend if the regex box is visible, however, since a checkbox is disabled this currently will not be visible.

shadrackkibet:
The designer looks good to me now. However, I noted that the code could be refactored better on the linking bits which are done manually. Since the regex keyboard is not working at the moment I think we can do it once it is enabled. 

@rdstern  @lloyddewit can you review now?

lloyddewit:
@rdstern Please could you approve? then we can merge, thanks

Issue #5835, opened 2020-06-12
Decreasing some dialogue depth and width


rdstern:
There are many ways that I suggest the dialogues could be made more consistent and hence easier to use.
One simple change is that some dialogues are very deep.  This is not just annoying, but they may become unusable on some screen resolutions.  

There are two dialogues in particular.
a) **_Prepare > Date > Use Date_** is very deep and I think quite easy to change somewhat.
b) **File > Open from File** is similarly very deep.  There seems to be a gap at the top on the left of the dialogue.  On the right it is quite "full", but I suggest the preview at the top could be made smaller.  I also note that the dialogue could be made less wide.  There seems to be an unnecessary gap in the middle of the dialogue.

I think these are the two main "culprits" on depth.
On width (not so vital):
c) The Prepare > Dates > Make Date dialogue seems unnecessarily wide.
d) So does the Model > Fit Model> Fit Model Keyboard
I am sure there are more!

shadrackkibet:
I think it will be good to discuss and agree on some of these things and document them so that it is one of the resources for developers. Good that @lilyclements will soon be involved again. She wrote the document [Making Dialogs.docx](https://github.com/africanmathsinitiative/R-Instat/files/4774199/Making.Dialogs.docx) attached which is still relevant however it might need updating.

lloyddewit:
@shadrackkibet Thanks for sharing the document. 

@lilyclements This appears to be a sort of style guide that could be applied to any UI. If you have time at some point, please could you tell me a little about the document's background? 

@africanmathsinitiative/developers I'm interested in any R-Instat development documentation that may exist. I'd like to collect it together and share it on the Taiga Wiki, or in some other repository. I currently have:
- R-Instat Developer Documentation from @maxwellfundi 
- Adding Data Sets in R-Instat from @berylwaswa 

Please let me know if you're aware of any other developer documentation, thank you

lilyclements:
@lloyddewit For the "Making dialogs" document, I wouldn't say I wrote it at all - I just took the appropriate bits out of a Windows guide and popped it into a document.
The document it is based off guidelines here - https://docs.microsoft.com/en-us/windows/win32/uxguide/guidelines
Perhaps that should be noted in the document.

@shadrackkibet you mentioned that some bits may need updating. This is most definitely a discussion for elsewhere, however, I am interested in what aspects need updating?
[...]

PR #5853, opened 2020-06-22
dlgImportData design

lilyclements:

Changes:
- design amendments to decrease the depth of the dialog from 746x645 to 714x541

One of the dialogs discussed in #5835 

Before:
image

After:
image


shadrackkibet:
Please, can we also see the designer when you have `.csv` and any of the excel file `.xlsx`or`.xls` selected? Import options will be different.

lilyclements:
**BEFORE -**
Excel:
image

RDS:
image

csv:
image


**AFTER -** 
Excel:
image

RDS:
image

CSV:
image

shadrackkibet:
All looks better now. My suggestions below.

`.csv`
- Can we also push csv options group box up so that it is aligned at the top with Text File Preview? This also applies to `.txt` option we could align the radio buttons (**Separator, Fixed width** etc) with the preview at the top.

**Excel**

- Aligning options group box with sheets list box?
- Rows to skip label is far from the control. Perhaps align it with missing value string input?
- The checkbox label seems not to show on run time. Perhaps that was not added. The checkbox should be linked to the controls below it **From** and **To**.

shadrackkibet:
I think the checkbox label should be **Range**  @rdstern, please confirm.

lilyclements:
**To the csv comment -** 
The group box for csv is the same as for txt but with the labels changed. 
This means I am unsure how to include the rdo buttons to be visible only for the txt option, and not the csv option, other than the manual method.
This will mean when csv is selected, there is a big space at the top of the group box where the rdo options are not visible as well.
Additionally, there's a separate group box for other txt files. I'm not sure of the distinction between these two txt groupboxes, and if the other one needs the rdo options visible.

I can align the other boxes (rds, excel) to the top, however.

**To the excel comment -** 
I agree and will implement 1 and 2
I did note the third point - this ucrChk is not referenced in the code at all. Is this checkbox only meant to link to the to/from ucrs? If that is the case, I'm sure I can fix that! (If it's a little more complex, it has been a few years since I did R-Instat code so I may need a bit of a refresh)

**To the data frame comment -**
This ucr has changed somewhat from when I last used it. When I used to shorten the ucrSave, it would push the text box closer to the text. Now it seems to shorten the text box and not push it closer to the text.
Is there a way I can shorten the ucrSave such that it pushes the text box closer to the text, rather than shortens the text box?

shadrackkibet:
Happy for you to fix the designer parts for now for this to be merged. If you are okay we can take-up those bits that need writing of code as a separate issue.

Yes, that should be linked to the to/from ucrs and it is only meant for excel options `.xls` and `.xlsx`.

The ucrSave issue is more general we might want to fix it by setting position and size through code so that it is updated on runtime, Okay to have it the way it is for now.

Patowhiz:
@lilyclements I'm hoping you won't touch the "VB code file" attached to this dialog. I'm also working on the dialog but just on it's code file so I think there shouldn't be any conflicts.

lilyclements:
@shadrackkibet I have changed the designer bits we discussed, see updated screenshots below of excel and RDS (I didn't change anything on csv/txt aside from the dialog width, which can be seen from the xlsx/RDS screenshots). This is ready for re-review.

@rdstern 
1. I have moved the block up for xlsx and RDS files. I am unsure how to do this for csv files (see previous comment) - this requires editing the code which @Patowhiz has requested I do not do
2. Done (see screenshots below)
3. I have asked about the ucrSave problem which I think has to be edited through the code.

Excel:
image

RDS:
image

shadrackkibet:
Happy with these changes. Looks good. I would suggest we have open issue(s) for issues not yet addressed. Especially on setting checkbox text(Range) and linking the checkbox to From/To inputs and making sure that works.

lilyclements:
@shadrackkibet agreed. Made an issue - #5864 
Anything else that should be added?

shadrackkibet:
Thanks @lilyclements, all is captured in the issue. This is ready for @rdstern and @lloyddewit to review and approve.

lloyddewit:
@rdstern please could you let me know if your requested changes are now implemented? I can then do the final review and merge, thanks

PR #5842, opened 2020-06-17
Added a new Structured menu to R-instat

Wycklife:
Fixes #5839

@africanmathsinitiative/developers I have added a new menu called Structured with a list of items. Currently the items are doing nothing except for circular>Define which opens circular dialog. At this, there are still discussions on how to implement the items within the menu. @rdstern @dannyparsons @volloholic @lloyddewit

PR #5856, opened 2020-06-23
Added a new Structured Menu to R-Instat

### LINKED ISSUE #5839: Add a new Structured menu to R-Instat
This is from @volloholic and has grown from looking where to add time series analysis into the current menu structure in R-Instat.
The idea is to have a **_new Menu, called Structured,_** just after the Model menu, so before Climatic.
This would have the same general structure as the whole system, so Prepare, Describe, Model.  (I don't think we need File, at least initially.)
Currently I suggest the Menu as follows:
```
Structured
Define (dialogue moved from Prepare menu.  Only item currently in the Prepare section.)
-------
Describe
-------
Model
```
In the **_Describe_** section we either  have a submenu, perhaps initially as follows: 
```
Circular
Low-Flow
Survival
Time Series
```

Or we have **_Describe: Circular_** on the main Structured menu, etc

Similarly for the Model section.

We have (at least) four types of structured data.
1) **_Circula_**r is currently the only type of data in the **_Prepare > Define_** menu.  We are ready to define time series data (ts, xts, zoo).
2) We are also adding the lfstat package (for hydrology) and this has the **_lfobj_** object, which is usually created from a ts object.
3) The **_surviva_**l package defines a survival object
4) **_Time series_** may be ts, or xts, or zoo object.

I am including quite a number of people, simply so we discuss this issue and can proceed reasonably quickly.  It is quite a big change, but is growing on me.  It fits with the R principle of defining structures and these are all examples where there is a key variable, plus additional meta-data.  The extra data may be because of an additional variable, e.g. the censoring variable in survival or in information derived from the structure or additional variables, such as time series.

Wycklife:
Fixes #5839

@lloyddewit I have managed to redo this since the other pr had conflicts. Could you review this so that it get merged to a void conflicts?

### DIFF
--- a/instat/frmMain.Designer.vb
+++ b/instat/frmMain.Designer.vb
@@ -602,6 +602,7 @@ Partial Class frmMain
         Me.mnuStructuredLow_FlowDefine = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuStructuredSurvival = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuStructuredSurvivalDefine = New System.Windows.Forms.ToolStripMenuItem()
+        Me.mnuStructuredTimeSeries = New System.Windows.Forms.ToolStripMenuItem()
         Me.stsStrip.SuspendLayout()
         Me.Tool_strip.SuspendLayout()
         Me.mnuBar.SuspendLayout()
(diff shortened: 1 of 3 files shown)


lloyddewit:
@Wycklife Thank you for closing PR #5842 and replacing it with this one. If you close a PR and reopen a new one to solve merge issues, this then I suggest that you add a comment to each PR to reference the connection between them. I just did it for these PRs.

PR #5860, opened 2020-06-25
Adding options to delete filters and calculations - delete objects dialog

### LINKED ISSUE #5792: Rename, delete and reorder filters, like graphs and models

shadrackkibet:
Fixes #5792. @rdstern this is now working please test. I am concerned that we are enabling the user to delete a current filter and also "no_filter" this will cause bugs. One option is to hide them from the selector so that the user does not delete it or stop the user from deleting and give notification that it is not allowed.

### DIFF
--- a/instat/dlgDeleteObjects.vb
+++ b/instat/dlgDeleteObjects.vb
@@ -96,18 +96,12 @@ Public Class dlgDeleteObjects
     End Sub
 
     Private Sub ucrInputComboType_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputComboType.ControlValueChanged
-        Dim value As String
-        Dim key As String
+        Dim key As String = dctTypes.Keys(ucrInputComboType.cboInput.SelectedIndex)
+        Dim value As String = ""
 
-        If dctTypes.Count <> 0 Then
-            value = dctTypes.Values(ucrInputComboType.cboInput.SelectedIndex)
-            key = dctTypes.Keys(ucrInputComboType.cboInput.SelectedIndex)
-            If (dctTypes.ContainsKey(key) AndAlso dctTypes.ContainsValue(value)) Then
-                ucrReceiverObjectsToDelete.SetItemType(value.Replace(Chr(34), ""))
-                ucrReceiverObjectsToDelete.strSelectorHeading = key
-            End If
-        Else
-            Exit Sub
+        If key IsNot Nothing AndAlso dctTypes.TryGetValue(key, value) Then
+            ucrReceiverObjectsToDelete.strSelectorHeading = key
+            ucrReceiverObjectsToDelete.SetItemType(value.Replace(Chr(34), ""))
         End If
     End Sub
 
--- a/instat/dlgDeleteObjects.vb
+++ b/instat/dlgDeleteObjects.vb
@@ -16,7 +16,8 @@
 
 Imports instat.Translations
 Public Class dlgDeleteObjects
-    Public bFirstLoad As Boolean = True
+    Private bFirstLoad As Boolean = True
+    Private dctTypes As New Dictionary(Of String, String)
     Private bReset As Boolean = True
     Private clsDefaultFunction As New RFunction
 
@@ -37,7 +38,6 @@ Public Class dlgDeleteObjects
     End Sub
 
     Private Sub InitialiseDialog()
-        Dim dctTypes As New Dictionary(Of String, String)
         ucrBase.iHelpTopicID = 352
 
         ' Selector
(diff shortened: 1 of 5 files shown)


shadrackkibet:
Another issue is when you have a filter in operation (current_filter). I think we should not allow the user to delete the filter.

lloyddewit:
@shadrackkibet Please could you clarify the status? thanks

shadrackkibet:
A check in R to stop the user from deleting the `current_filter` will suffice for this. I will endeavour to add it soon then this will be ready for re-review. Thanks for the reminder.

shadrackkibet:
> It does show that I have no way of knowing which filter is currently being used. I suggest that could become a new issue. The rows are shown in red, so perhaps there could be a Current Filter item.

@rdstern I didn't quite understand what you mean by this statement. From your example above the error tells you the current filter is `Filter2`. Do you mean there should be a way the user knows the current filter before they select it to be deleted?

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1714,6 +1715,26 @@ DataSheet$set("public", "rename_object", function(object_name, new_name, object_
...
 
+DataSheet$set("public", "delete_objects", function(data_name, object_names, object_type = "object") {
+  if(!object_type %in% c("object", "graph", "table","model","filter", "calculation")) stop(object_type, " must be either object (graph, table or model), filter or a calculation.")
+  if(any(object_type == c("object", "graph", "table","model"))){
+      if(!all(object_names %in% names(private$objects))) stop("Not all object_names found in overall objects list.")
+      private$objects[names(private$objects) %in% object_names] <- NULL
+    }else if(object_type == "filter"){
+      if(!all(object_names %in% names(private$filters))) stop(object_names, " not found in filters list.")
+      if("no_filter" %in% object_names) stop("no_filter cannot be deleted.")
+      if(any(private$.current_filter$name %in% object_names))stop(private$.current_filter$name, " is a current filter it cannot deleted.")
```
shadrackkibet, line 1726:
```suggestion
      if(any(private$.current_filter$name %in% object_names))stop(private$.current_filter$name, " is currently in use and cannot be deleted.")

```

rdstern:
Yes, this is more general than deleting a filter.  When a filter is in operation the row variable is in red, so you know a filter is being applied.  But which?  I thought - as a separate issue, perhaps youi could right-click in the rows (i.e. on the red bit) and there would be simply a label that Says Filter: Filter2 if a filter is in operation and perhaps Filter: None if no filter is currently applied?

lloyddewit:
@shadrackkibet Thanks for all your recent work on this. I understand that this PR is now ready for review and that the outstanding points will be raised as a new issue. Is this correct?

shadrackkibet:
Yes.

shadrackkibet:
Yes.

lloyddewit:
@shadrackkibet Thanks for the quick reply, please could you create the new issue and reference this PR in it?
@Ivanluv Please could you peer review this PR?

shadrackkibet:
Anything that requires my attention here? otherwise, this should be merged.

lloyddewit:
@Ivanluv This is one of the PRs you are reviewing, did you get a chance to do this yet? thanks

shadrackkibet:
If there is nothing for me to look into here then I think it is appropriate to get this merged!

rdstern:
Could you get @Ivanluv to review the code - as suggested above.  If that is ok, then perhaps @maxwellfundi could do the final code review as @lloyddewit is not around now, and then Maxwell can also merge once he approves.

Ivanluv:
This looks ok from my review @maxwellfundi  could you also check

lloyddewit:
@rdstern @maxwellfundi I'll approve and merge after @Ivanluv  has peer reviewed.

shadrackkibet:
> This looks ok from my review @maxwellfundi could you also check

Looks like @Ivanluv already peer-reviewed?

instat/dlgDeleteObjects.vb
```diff
@@ -37,6 +37,7 @@ Public Class dlgDeleteObjects
     End Sub
 
     Private Sub InitialiseDialog()
+        Dim dctTypes As New Dictionary(Of String, String)
```
lloyddewit, line 40: Make this a private class variable (see comment below)?

instat/dlgDeleteObjects.vb
```diff
@@ -84,6 +95,29 @@ Public Class dlgDeleteObjects
         TestOKEnabled()
     End Sub
 
+    Private Sub ucrInputComboType_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputComboType.ControlValueChanged
```
lloyddewit, line 98: Would this function be smaller and safer if we used the dictionary above? For example:
```
Dim value As String
If (dict.TryGetValue(ucrInputComboType.GetValue(), value)) Then
    ucrReceiverObjectsToDelete.SetItemType(value)
    ucrReceiverObjectsToDelete.strSelectorHeading = ucrInputComboType.GetValue()
End If
```

instat/dlgDeleteObjects.vb
shadrackkibet: Perhaps I did not understand your suggestion. Please clarify.

By declaring `Dim` it means that the scope of `dctTypes` is within `InitialiseDialog()` sub-routine which is what we want in this case.

instat/dlgDeleteObjects.vb
shadrackkibet: @lloyddewit thanks for your suggestion. I have made a slight modification to achieve this. What do you think? See below.

```
     Dim value As String
     Dim key As String

        If dctTypes.Count <> 0 Then
            value = dctTypes.Values(ucrInputComboType.cboInput.SelectedIndex)
            key = dctTypes.Keys(ucrInputComboType.cboInput.SelectedIndex)
            If (dctTypes.ContainsKey(key) AndAlso dctTypes.ContainsValue(value)) Then
                ucrReceiverObjectsToDelete.SetItemType(value.Replace(Chr(34), ""))
                ucrReceiverObjectsToDelete.strSelectorHeading = key
            End If
        Else
            Exit Sub
        End If
```

shadrackkibet: I also think we could make things clear by having `GetSelectedValue()` and `GetSelectedKey()` methods in the combobox class. I find `GetValue()` and `GetText()` methods not clear enough.

lloyddewit: @shadrackkibet Thank you for the reply. I agree that we need to check for an empty dictionary. I also like that you created the `key` and `value` variables, I think this makes things more readable. I think we can simplify the code a bit more while still maintaining the readability:

```
    Private Sub ucrInputComboType_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputComboType.ControlValueChanged
        Dim key As String = dctTypes.Keys(ucrInputComboType.cboInput.SelectedIndex)
        Dim value As String = ""

        If key IsNot Nothing AndAlso dctTypes.TryGetValue(key, value) Then
            ucrReceiverObjectsToDelete.strSelectorHeading = key
            ucrReceiverObjectsToDelete.SetItemType(value.Replace(Chr(34), ""))
        End If
    End Sub
```
What do you think?

shadrackkibet: That is okay.

shadrackkibet:
@rdstern this is a reminder for you to approve.

shadrackkibet:
@lloyddewit ?

PR #5867, opened 2020-06-27
Taylor diagrams implementation

### LINKED ISSUE #4944: Adding the Taylor diagram to R-Instat
I am not sure if the plotrix package is already installed.  It is in my version of R-Instat.  If not, then I don't really mind if it is installed for the following work, or not.
It has a lot of plotting features that users of R-Instat may value.  One example, from our recent Pretoria workshop is **Taylor diagrams**.  It also has Gantt charts and Australian Met style wind roses.

It is a big package with lots of features and lots of authors including Hadley Wickham.  But the graphs are not ggplot2 graphs.  So I don't expect us to want to do lots of dialogues - at least in the near future.

Instead I propose that they form examples for the "tweaking R-Instat" guide, where these could be done either through our script window, or through moving to RStudio.  And the guide I am writing for AIMS (and now also for Met Services) could give examples.
(This started as a request for the plotrix package.  I think this is now added, but the t=Taylor diagram still remains to be included.  Hence the change of name of this issue.)
It is also useful, to explain some of the complications in R that R-Instat has hidden from users.  For example the Wind roses needs the data in a matrix, rather than in 2 columns, which is the situation with our example and other wind roses.  So it is a good opportunity to explain this sort of stuff.

If plotrix is already installed, then that's fine.  Otherwise this section in the guide can cover adding a new package.

shadrackkibet:
Fixes #4944. @rdstern this is the initial implementation of Taylor diagrams. Please do some initial testing.

### DIFF
--- a/instat/dlgTaylorDiagram.Designer.vb
+++ b/instat/dlgTaylorDiagram.Designer.vb
@@ -1,9 +1,9 @@
-﻿<Global.Microsoft.VisualBasic.CompilerServices.DesignerGenerated()> _
+﻿<Global.Microsoft.VisualBasic.CompilerServices.DesignerGenerated()>
 Partial Class dlgTaylorDiagram
     Inherits System.Windows.Forms.Form
 
     'Form overrides dispose to clean up the component list.
-    <System.Diagnostics.DebuggerNonUserCode()> _
+    <System.Diagnostics.DebuggerNonUserCode()>
     Protected Overrides Sub Dispose(ByVal disposing As Boolean)
         Try
             If disposing AndAlso components IsNot Nothing Then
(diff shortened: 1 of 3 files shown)


rdstern:
Not directly for this dialogue, but I can't easily transfer this graph into the R viewer.  It works there if I use the Tools > Options menu.  But the option on the toolbar and the Dialogue > Display Graph > Display in R viewer is only for ggplot graphs.  We may want an issue soon to be able to either use ggplotify routinely or to have simpler facilities to display in R viewer - perhaps this could be another option on the toolbar graph item?

shadrackkibet:

lloyddewit:

lloyddewit:
@rdstern @shadrackkibet I raised issue #5885 for the remianing small issues.

shadrackkibet:
Thanks, @lloyddewit for putting up an issue on the remaining small issues.

rdstern:
Thanks from me too.  When I see all the options for the Taylor diagram this may become quite a rich dialogue.  That would be consistent with it having its own dialogue even in the future.
For now these addition features can be tested by starting in the dialogue and then continuing in the script window.

Issue #5838, opened 2020-06-13
Controls Design Properties


Patowhiz:
**Is your feature request related to a problem? Please describe.**
Currently some important design properties(e.g labels) of the custom controls can't be set in design time, developers have to rely on doing this in code because the "setter" methods aren't VB.NET properties(which Visual Studio is fond of), see issue #5831. This may prompt trial and error from the developers side to set the correct sizes. 

**Describe the solution you'd like**
I suggest redoing some of these "setters" in VB.NET properties, this will help in reducing the code needed to be typed and help developers take advantage of Visual Studio properties feature

lloyddewit:
Note: There is an interesting discussion on the pros and cons of this approach in PR #5875.

PR #5875, opened 2020-06-30
Check box text property

Patowhiz:
Partly Fixes , issue #5838. (More controls to await such enhancements)

Currently to set the text of the checkbox, developers have to call `ucrCheck.SetText()`. This sometimes increases substantially the number of lines needed to write and during peer reviewing the team has to go through these lines which are basically more or less "labels" of the checkboxes. To take advantage of visual studio and have it write such trivial code for us, this subroutine has to be rewritten or complemented with a VB.NET property. The VB.NET property Text will now appear in the design properties panel of visual studio. 

Another critical advantage of this enhancement is, when designing the dialog the developers can now see the size that the text will occupy on the dialog, without necessarily having to run the application, all they need to do is find the property in the design view and set the text.

@lloyddewit If this is merged before PR #5847, @shadrackkibet could take advantage of this in labelling the many checkboxes on those dialogs.

### DIFF
--- a/instat/ucrCheck.vb
+++ b/instat/ucrCheck.vb
@@ -23,6 +23,15 @@ Public Class ucrCheck
     Private strValueIfChecked As String = "TRUE"
     Private strValueIfUnchecked As String = "FALSE"
 
+    Public Overrides Property Text() As String
+        Get
+            Return chkCheck.Text
+        End Get
+        Set(value As String)
+            chkCheck.Text = value
+        End Set
+    End Property
+
     Public Sub SetValueIfChecked(strNewValueIfChecked As String)
         Dim clsTempCond As New Condition
 


instat/ucrCheck.vb
```diff
@@ -23,6 +23,15 @@ Public Class ucrCheck
...
     Private strValueIfUnchecked As String = "FALSE"
 
+    Public Overrides Property Text() As String
+        Get
+            Return chkCheck.Text
+        End Get
+        Set(value As String)
+            chkCheck.Text = value
+        End Set
+    End Property
```
shadrackkibet, line 33: Correct me If I might have misunderstood how this will work. This change will encourage the setting of Labels using the properties tab. I don't think this is what we want to encourage generally. My understanding is that we should always try as much as we can to have properties set in code rather than setting them via properties Tab which can make it easier to debug code. I know we sometimes set properties via the Tab, especially for non-customized controls. Furthermore, I am also concerned that this change will cause inconsistencies across the software because we will have labels set in two different places, not unless we want to change it in all other places.

Does this change reduce code from `.vb` by just moving text label to another file (`.Designer.vb`) as an extra line? I will have to move labels for over 30 checkboxes to the properties Tab for each checkbox. That appears to be inefficient for me.

Patowhiz: @shadrackkibet thank you for your good queries. I'm happy you are concerned about this as much as I am. Yes, this code is precisely meant to encourage setting of the checkbox text through the properties palette and also act as one of the steps of refactoring R-Instat code.
 
Below are my thoughts in regards to your good queries;

> I don't think this is what we want to encourage generally.

That is want I think we should encourage generally, It's why Visual Studio is designed to take advantage of this, these helps you see the labels in design time and reduces time taken to resize, reposition controls like this one. The WYCIWYG Visual Studio feature is fundamentally designed around the properties feature 

> I know we sometimes set properties via the Tab, especially for non-customized controls.

We've been doing this because the custom controls never took advantage of the VB.NET property feature, the property feature is basically one of the fundamental features of VB.NET, it's a no brainer to me why we shouldn't be taking advantage of such a well thought out language feature.

> Furthermore, I am also concerned that this change will cause inconsistencies across the software because we will have labels set in two different places, not unless we want to change it in all other places.

The basic idea here is actually to create consistency with how VB.NET controls are designed and implemented, remember the custom controls inherit the `Control`, which is the parent of VB.NET controls, thus all `"ucrs"` are basically controls and they should have characteristics that are designed like the parent as much as possible. In regards to confusion, remember we have properties that are changed by the application logic and others that are never changed, this is always very direct in the code. This is meant to target those that are never changed but affect the design of the form. 

> Does this change reduce code from .vb by just moving text label to another file (.Designer.vb) as an extra line?

Yes, at the moment as you design the window, adding controls, positioning controls etc, all that code is being written in the designer file, why?, cause that is designer code, and it's good practice to separate designer code from the application logic. This is one of the principles of an MVC architecture and it is one of the reasons Visual Studio and VB.NET was very popular among beginners. This architecture was borrowed from internet applications technologies and found its way to desktop technologies, ever since we've had variants of the MVC architecture ported to even mobile technologies like Android. Which is a strong prove that it works, even for bigger projects.

> I will have to move labels for over 30 checkboxes to the properties Tab for each checkbox. That appears to be inefficient for me.

For work already done, yes probably. But look at it this way, 30 lines of designer code pushed to the designer(where it belongs) makes life easier for everyone. When going through the code, generally the designer code is usually not so important, cause you can visually see it when you open the form or run the code (and when changing it, mostly you never write the code anyway). So just diving in and starting to look at the application code makes my life much easier. You can imagine if all the hundreds of lines of designer code was mixed with application logic code, life would have been a bit harder for VB.NET developers. 

To sum it up. When going through R-Instat it was a no brainer to me that we should:

1. Use the VB.NET language features where possible (they have been battle tested for years).

2. Try as much as possible to embrace good recommended design and implementation practices even if their merits are not so obvious (these practices are well thought out, after years of software development and any future developer with a computer science background will easily be able to relate with and be productive ).

I'm happy to be convinced otherwise and to learn more about why such a feature shouldn't be included.

lloyddewit: @Patowhiz @shadrackkibet Thank you for this interesting discussion, I think this is an important design decision.
My strong technical preference is to manage this property in the way that @Patowhiz suggests. I also understand @shadrackkibet 's concerns that this is different from the current way of doing things, and may have repurcussions for existing checkbox labels.

@Patowhiz - Does this change affect existing dialog boxes?

Patowhiz: @lloyddewit no this doesn't affect the implementation of the existing dialogs and also it won't require them to be reimplemented. The developers may choose to use this to simplify their work if need be.

Patowhiz: Just to add on that, this new change is already compatible with the current `ucrCore.SetText()` functionality. So `ucrCheck.SetText()` will just internally call `ucrCheck.Text`.

shadrackkibet: If we are going to encourage this practice generally then I would like to have a general discussion involving other developers with @dannyparsons and/or @volloholic. If agreed then definitely we should refactor dialogs for consistency. Not a must to be done now, it can be done in bits. 

This will also mean we change the way we set other properties like `visible, enabled, read-only, backcolor` etc which we always encourage to be set via code. Also tooltips?

Issue #5876, opened 2020-07-01
Popup Component


Patowhiz:
Currently we have code that is responsible for creating a custom "popup" in different dialogs and controls. As discussed with @dannyparsons there's need to have this code in a central place which will help in future maintenance, enhancement and avoid repetition of code as more dialogs continue taking advantage of this feature. 

Check screenshots of the popups below
![popup1](https://user-images.githubusercontent.com/21179007/86271073-f40a5480-bbd4-11ea-8757-3fded7450ed9.png)

![popup2](https://user-images.githubusercontent.com/21179007/86271091-f9679f00-bbd4-11ea-8eb1-764232b5ca82.png)

Issue #5596, opened 2019-11-28
Add examples to the Prepare > Generate > Enter dialogue


rdstern:
@Patowhiz here is the current Enter dialogue.

I am interested in adding some examples.  If so, then I wonder how this should be done?  
We can have tooltips, of course.
But in addition I wonder about helping users in the writing of examples of the commands as we do in the File > New Data Frame dialogue.

Here there are 2 obvious ways.  We could have a button - as we do in the File > New Data Frame.  I assume that would be quick and easy to do, following the File >New Data Frame work.  It would be just the right-hand side of the formula.  It could go just after the show arguments checkbox.

Or the dialogue automatically remembers past commands.  When you start it is empty.  It could start with a few examples - perhaps below a horizontal line.  

@Patowhiz What do you think?  Should we add the examples?  Would it be very easy with a button?  If so, then I am slightly inclined that way.

I got to this because I would like to build towards a guide to help beginners on their first commands that use these "halfway" dialogues.

Patowhiz:
@rdstern yes we can have a similar "examples" functions like in the data frame dialog. I suggest we have the examples button just above the combobox to be consistent with the data frame dialog and also future dialogs that may need this. I'm have no reservations with having it in another place though.

Please share some of the examples that you have. 

![dlgEnter](https://user-images.githubusercontent.com/21179007/86086373-43a03180-baaa-11ea-9a44-05c7e9432621.JPG)

rdstern:
@Patowhiz happy to have you add this.  Could you move the Show Arguments over a little and put it at the same level to the right of that?

Patowhiz:
Yes I can do that. Will have this ready for your review as soon as you share the examples.

Patowhiz:
@rdstern that is the new design. Awaiting the sample commands.

![dlgEnter](https://user-images.githubusercontent.com/21179007/86311513-3f4e5280-bc29-11ea-96b6-cc8d6ca7702f.png)

rdstern:
@Patowhiz that is great.  I am looking at it now.  I am finding other small items I would like to change.  Happy for you to complete all this, or for Wyclife to do some of the tasks if that fits.  Then I would like @lilyclements to review, because it is a good example where the layout can be really helpful and she is great on this type of detail.

a) I would prefer the Examples button after the Show arguments checkbox, rather than before.  That's a trivial one.
[...]

PR #5877, opened 2020-07-01
Empty clsPopup

Patowhiz:
@lloyddewit @maxwellfundi this is an empty `class` meant to hold the code that will be creating the popup used in several dialogs. 

Since I last created such a popup for the comment section, github issues have been created demanding more use of it (currently there is another dialog that needs this feature outlined by @rdstern in issue #5596 and possibly more will arise). Because of this we had discussed with @dannyparsons to have this as a reusable component that can be used in any dialog without much of re-implementation( and copy paste). Check out more about this `component` in issue #5876 . 

So I have added an empty class so that this can be merged quickly and avoid conflicts incase another item is added to the R-Instat `solution`. After this is merged I'll push the code needed to create the actual popup and demonstrate how this `class` can be used to dynamically create the `component`.

### DIFF
--- a/instat/clsPopup.vb
+++ b/instat/clsPopup.vb
@@ -1,0 +1,3 @@
+﻿Public Class clsPopup
+
+End Class
--- a/instat/instat.vbproj
+++ b/instat/instat.vbproj
@@ -179,6 +179,7 @@
     <Compile Include="clsCustomRenderer.vb" />
     <Compile Include="clsGgplotDefaults.vb" />
     <Compile Include="clsInstatOptionsDefaults.vb" />
+    <Compile Include="clsPopup.vb" />
     <Compile Include="clsQualityControl.vb" />
     <Compile Include="clsRCodeStructure.vb" />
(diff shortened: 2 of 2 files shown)


Patowhiz:
I thought about this and realized that this will not have designer and resx. It's just a class that will dynamically create the GUI during runtime.  The size, controls of this "component" are only know by the callers of this class. That's why I used cls and not sdg,dlg,ctrl. I've actually already implemented it and pushed the commits on a separate PR.

Patowhiz:
You can check out it's Implementation on PR #5878. I'm still happy to rename it though, if you still strongly think that the prefix should change.

lloyddewit:
@rdstern or @maxwellfundi This is a placeholder and there's nothing to test here. Please could one of you approve so that this PR can be merged? Many thanks

Patowhiz:
@lloydewit I don't know of any classes like this one In R-Instat no. If classes had been arranged into  packages(namespaces) I would have put it under the UI package.

PR #5891, opened 2020-07-09
Infilling missing values using zoo package methods

shadrackkibet:
This is still a work in progress. @rdstern we do not yet have an issue for this. Can you also do some preliminary testing? It is in a state that can be tested. Here is a screenshot.

image

### DIFF
--- a/instat/dlgInfillMissingValues.vb
+++ b/instat/dlgInfillMissingValues.vb
@@ -35,6 +35,8 @@ Public Class dlgInfillMissingValues
     End Sub
 
     Private Sub InitialiseDialog()
+        Dim dctFunctionNames As New Dictionary(Of String, String)
+
         rdoNeighbouring.Enabled = False
         rdoMultiple.Enabled = False
         ucrPnlOptions.AddRadioButton(rdoSingle)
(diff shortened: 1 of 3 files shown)


rdstern:
@shadrackkibet this is a great start.
a) We need to add a checkbox with the label Max Gap:  (up-down from 1 with no upper limit) Rows.  If unticked then there is no limit to the maximum gap.
b) The first 4 options work for temperatures from Dodoma.  
c) locf and structTS don't work for tmax and tmin.  They give different errors, but the problem is the missing values at the start of the series.  At least they work for rainfall which doesn't have this problem.
d) A big one to consider is that they don't work when a filter is applied.  @shadrackkibet good you are also working on filters!  Can we do anything about this?  I am happy that the interpolated data would have missing values for the filtered portions.  
It is quite common to have an initial missing block.
e) The structTS assumes the data are a time series and then fits a time series model, so taking the structure of the data into account when infilling the missing values.  For us now the data will usually be daily.  I wonder if we can have a combo box for hourly, daily, 5-day, 7-day, 10-day, monthly, yearly, but you could alternatively type the periodicity.  Then the time series object could be defined on the fly and could be a very good way to interpolate.  (Leave hourly for now, because that has 2 periodicities.  I think a zoo object can cope with that?  For daily data the periodicity may be 365.25?)

rdstern:
More suggestions: this is exciting

a) We do need to take account of the possibility of having to **_ignore missing at the start and end of the series_**.  So I propose (if possible) we always use the zoo::na.fill function.  This includes a list of 3 values for the start, middle and end of the series.  So it is ideal for us.  I think, as an example:  `zoo::na.fill(object=tmax,fill=list(NA,"structTS",NA))` is an example.
b) So I think we don't need to permit a filter - or we just ignore it  i.e. take it off and then possibly reapply it - like the calculator now.
c) So I suggest an additional control.  Could be a box with label Start and End.  Inside are 2 radio buttons for now.  First is Leave as Missing.  Second is "Extend infill".  (That's the option Extend )
d) We should change the names, so they are more informative.  The first could be Interpolate, and the second could be average.  I am not sure we need the third in its current form

rdstern:
None of the obvious options above are very useful for infilling daily rainfall data.  Usually we should use data from neighbouring stations, but it would be good to have at least one option that is sort of acceptable.  Often there will be relatively few missing values.  The method I am suggesting is not intended to infill (say) complete missing months.  Then it may be better to omit the year from the analysis.  Infilling from neighbouring stations may sometimes be possible for longer periods.

One simple way of infilling a few missing values could be to sample (at random) from the remaining non-missing values in the same month (and year?).  With the zoo::na.aggregate you can get a mean or median and it implies any function.  So could we choose the sample function and get a sample value from the non-missing data instead of the mean or median.  It needs to be possibly different for each missing observation.  @dannyparsons is of the view that this should be an easy option with the current zoo function and this should obviously be checked first.  If that doesn't work then a new function may have to be written.

shadrackkibet:
@rdstern I have now implemented most of your suggestions above. Please check the latest changes. 3 points still remain as stated below.
> e) The structTS assumes the data are a time series and then fits a time series model, so taking the structure of the data into account when infilling the missing values. For us now the data will usually be daily. I wonder if we can have a combo box for hourly, daily, 5-day, 7-day, 10-day, monthly, yearly, but you could alternatively type the periodicity. Then the time series object could be defined on the fly and could be a very good way to interpolate. (Leave hourly for now, because that has 2 periodicities. I think a zoo object can cope with that? For daily data the periodicity may be 365.25?)

**Comment:** It is not yet clear to me how to achieve this with the `zoo(), zooreg()` or `as.ts()` functions. 

>The spline looks fine as the name. Again (as with all the options we would like to be able to use the 3-element list as well.

**Comment:** You mention that we want the three-element list eg `fill = list(NA,"extend",NA)` work with all options. I only found this to work for `na.fill()`-constant. I don't think it works for others directly.
>Now a possible big development. I suggest two controls at the top for Station and Date as we usually have for the climatic dialogues. Then the idea above of Top. Middle and Bottom. needs to apply separately for each station. This could use purrr, etc to handle the manipulation. It could use some R-skills and @dannyparsons can help.

**Comment:** Not yet.

>None of the obvious options above are very useful for infilling daily rainfall data. Usually we should use data from neighbouring stations, but it would be good to have at least one option that is sort of acceptable. Often there will be relatively few missing values. The method I am suggesting is not intended to infill (say) complete missing months. Then it may be better to omit the year from the analysis. Infilling from neighbouring stations may sometimes be possible for longer periods.

>One simple way of infilling a few missing values could be to sample (at random) from the remaining non-missing values in the same month (and year?). With the zoo::na.aggregate you can get a mean or median and it implies any function. So could we choose the sample function and get a sample value from the non-missing data instead of the mean or median. It needs to be possibly different for each missing observation. @dannyparsons is of the view that this should be an easy option with the current zoo function and this should obviously be checked first. If that doesn't work then a new function may have to be written.

**Comment:** Not yet.

rdstern:
This is excellent as a start.  I quickly give some very small suggestions - mainly for tidying.  

a) Add the colon to Max Gap and move the checkbox a bit to the right.  It looks now as though it is connected to the Function control.  Perhaps move it to the right and under that single receiver as it is there for all the 6 settings.  (You are better than I am at layout.)  Also, add **_row(s)_** after the updown control.
b) Under the Constant option the label is `Insert Constant:` Change it to **_Value:_** (Just Value, not insert as it is replacing missing by the chosen value.)
c) I really like the fact you have already included the Position control.  In some dialogues it picks up the variable (if completed) from the main dialogue.  Could we do this here?
d) Perhaps a bit more to follow!

shadrackkibet:
I am of the opinion that this should be made available to the Interns particularly @HawardKetoyoMsatsi and @penina-AI so that they can test it over the week mainly because this is what they have been working on. This is also a good challenge for them to propose solutions to the remaining issues because it is mainly in R and also propose further improvements.

rdstern:
Agreed!

lloyddewit:
@Ivanluv Please could you peer review this PR? thanks

Ivanluv:
@lloyddewit  this looks fine from my review

instat/dlgInfillMissingValues.vb
```diff
@@ -14,8 +14,238 @@
...
+        ucrPnlMethods.AddFunctionNamesCondition(rdoNaFill, "na.fill")
+        ucrPnlMethods.AddFunctionNamesCondition(rdoNaLocf, "na.locf")
+        ucrPnlMethods.AddFunctionNamesCondition(rdoNaStructTS, "na.StructTS")
+        ucrPnlMethods.AddFunctionNamesCondition(rdoNaSpline, "na.spline")
+
+        ucrPnlStartEnd.SetParameter(New RParameter("fill", 1))
+        ucrPnlStartEnd.AddRadioButton(rdoLeaveAsMissing, "list(NA, ""extend"", NA)")
+        ucrPnlStartEnd.AddRadioButton(rdoExtendFill, Chr(34) & "extend" & Chr(34))
+        ucrPnlStartEnd.AddRadioButton(rdoInsertConstant)
+        ucrPnlStartEnd.AddParameterValuesCondition(rdoLeaveAsMissing, "fill", "list(NA, ""extend"", NA)")
```
lloyddewit, line 63: @shadrackkibet Just a question (you don't need to change anything).
In R-Instat quotes in strings are normally implemented with `& chr(34)`. Here you use the alternative notation `""`. Is there a special reason for this?

shadrackkibet: That is correct @lloyddewit. I can't say it is a special reason I just found it convenient otherwise ` "list(NA, " & Chr(34) & "extend" & Chr(34) & ", NA)"` should work.

shadrackkibet:
@rdstern Please check my latest changes, this bug should now be fixed and also mean should not return NaN values anymore.

PR #5803, opened 2020-05-26
Adding the scale_x_date function to the ucrAxes 

Ivanluv:
fixes #5266

PR #5896, opened 2020-07-10
Adding bar labels-use summaries dialog

### LINKED ISSUE #5398: Adding labels to bar chart, etc.

shadrackkibet:
This is a continuation of work done in PR #5411. Fixes #5398.


shadrackkibet:
@rdstern please check if this is working as required. It is currently implemented on use summaries dialog.

lloyddewit:
This PR replaces PR #5411

shadrackkibet:
@rdstern the bug you have reported above(b) was caused by changes done by @Ivanluv and it is more general. Plot Options doesn't work in most dialog now. I reported this bug in https://github.com/africanmathsinitiative/R-Instat/pull/5803#issuecomment-657019813 @lloyddewit was able to trigger this exception in his machine and he reported. I wonder if @Ivanluv has been able to investigate it further. I thought it should be easy to fix the problem you just need a new instance of the function somewhere `New`.

Point (c) should be an easy to fix. I will have a look at it at some point.

Ivanluv:
@shadrack am looking at the bug now

lloyddewit:
@shadrackkibet PR #5902 is now merged. Does this mean that this PR is now ready for review? thanks

shadrackkibet:
Not yet ready. We still need to make sure labels work properly.

lloyddewit:
@shadrackkibet Is it possible to start work on this PR again or is there something external blocking? thanks

shadrackkibet:
Yes I will look into this again. Thanks for the reminder.

lloyddewit:
@shadrackkibet are you still planning to work on this PR or is something blocking this? thanks

shadrackkibet:
@rdstern I didn't want to withdraw this PR without giving this a try. Could you please check if the label positioning is now fixed?

rdstern:
@shadrackkibet great.  I'll try

rdstern:
@shadrackkibet well done.  This seems now to work.  Well sort of. But I think you may have cracked it! 
I tried with the survey data as usual and with summaries by village and variety.
[survey_with_summary.zip](https://github.com/africanmathsinitiative/R-Instat/files/5432991/survey_with_summary.zip)

a) Initially it works fine, and the one change I suggest is within the geom_label to add show.legend=FALSE so you don't get the a added in to the colour of the legend.
b) I then did plot options > Layers and tried to edit the layer and got the message:
image

However it then allowed be to edit the layer.  I made the legend change myself and it worked fine.  I capture the screen below to compare with later:
image

It is still working fine.  So I got ambitious and used y-nudge to edit a bit more.  Now it was odd.  So I returned and undid that change.
However, now the command and the graph is still different:
image

So I reset the main dialogue, but don't now know how to get back to the labels as I had them before: 
image

shadrackkibet:
b) This is caused by how the `position` parameter is added into the layers sub dialog. This needs discussion to fix correctly which will also affect other graphics dialogs. I note that the current implementation is a bit messy! Not all position options are included eg `nudge, jitterdodge` and these options are inconsistent between layers. There are comments about it in code.

In summary, how do want the parameter position to be added into a layer? e.g `"position = "dodge"`or `position = position_dodge()` etc? This might also affect how that is displayed in the control. The advantage to using `position = position_dodge()` is that it allows fixing of the parameter `width=1` which ensures that labels are displayed in the correct position. Other options `position_identity(), position_jitterdodge(), position_jitter(), position_nudge(), position_stack()` allow some other parameters. Do we want to allow changing of these values from the dialog/subdialog?

rdstern:
This all sounds sensible and useful to discuss more generally.  Where you are also makes me feel we should proceed.  But my R isn't up to this.  Can @lilyclements and/or @dannyparsons help here?

rdstern:
@shadrackkibet well done, this seems to be working pretty well now.  Until I edit the layer it remains pretty sensible.  Once I start doing that, then it goes a bit haywire and I wonder if the problem is more general and just uncovered here.

I simply tried nudging the y to put the labels a bit lower.  I then tried changing from the mean yield to the counts.  It wasn't going well so I did a reset.  That doesn't seem to affect the settings in the sub-dialogue.  When I changed back to the means I got:

image

And the last graph that had worked was:
image

This had the counts, though I was now plotting the means.  And it had put the legend for that layer back, i.e. the a's though I hadn't touched that.

If, as I suspect, I am finding more general problems, then perhaps this change could be merged and then the further problems are attacked.

shadrackkibet:
@rdstern I am for the opinion of disabling the labels checkbox for now. The positioning of bars is already and improvement!

lilyclements:
> This all sounds sensible and useful to discuss more generally. Where you are also makes me feel we should proceed. But my R isn't up to this. Can @lilyclements and/or @dannyparsons help here?

Is this something to discuss here or elsewhere on an issue? I am happy to look at it later today or tomorrow in more depth, but ask because I am not sure how urgently you wanted this PR to be merged, and this affects other dialogs.

lilyclements:
Just had a brief look at this and had a couple of comments:

**Dialog Design:**
1. The "X-Variable", "Y-value", and "Y-Variable" (when visible) labels need colons
2. If I select "Variable" under "Y-Value", you get an overlap on the labels for the Position label. 
image

3. Can the "Position" input box be increased in size to the "Y count" input box? (Or both of them to be the same size as the receivers?
This is just for aesthetic, but also because then all of the "Stack in reverse" option can be read when that's selected
image

**Position Parameter:**
As for the `position` parameter question: My understanding is that currently the layer options in the sub-dialog offer `position = "dodge"` whereas the main dialog offers `position = position_dodge()` (and similarly so for `"stack"`, etc). This causes two problems:
1. Inconsistency if you select one parameter value (e.g. `"dodge"`) on the sub-dialog but another one (e.g. `position_stack()`) on the main dialog. Presumably it should update on the main dialog if you edit it on the sub-dialog?
2. Inconsistency between whether `position_dodge` or `"dodge"` should be selected. (Similarly so for `position_stack`, `position_jitter`, ... etc). I guess whether `position_dodge()` or `"dodge"` is used depends on whether we want to alter the parameter options in the function? (Similarly so for `position_stack`, etc). But presumably it should be the same option in both places. Perhaps then it is worth changing them to be `position_...()` and then there is the option to adding the extra parameter options for the `position` functions later?

Finally, is there a reason why `position_dodge2()` is not offered on this dialog? (I only ask since `"dodge2"` is an option on the sub-dialog)

shadrackkibet:
@lilyclements good that you have pointed out inconsistencies under "Bar and Pie chart" dialog. However, this PR is about  "Use summaries dialog" under **Describe>General>Use summaries**.

lilyclements:
@shadrackkibet my apologies - thank you for pointing this out! I can make those aesthetic changes to "Bar and Pie chart" later this week then while it's on my mind?
In this case, there is still one inconsistency - if I keep it on the default of `dodge` on the main dialog, and then select `jitter` on the layers sub-dialog, the main dialog still is on the `dodge` option (although `jitter` runs).
If I reopen, and change the position option on the main dialog, the `jitter` option from the sub still runs, and the main dialog option does not update.

shadrackkibet:
@rdstern I wish to disable labels checkbox for now then this can be merged? It can be enabled once sync between labels checkbox with the subdialog options is fixed properly!

dannyparsons:
@shadrackkibet I suggest you hide any features which are not currently implemented so that the dialog will look complete to a user. I haven't looked at the design for this dialog but if it's appropriate I would prefer controls that are not implemented to be hidden rather than disabled (which typically suggests a user can do something to enable them).

lloyddewit:
@shadrackkibet thanks for the changes, is it ready for test again or do you plan more commits?

shadrackkibet:
Yes, it is ready.

instat/ucrCheck.vb
```diff
@@ -179,4 +179,13 @@ Public Class ucrCheck
...
     End Sub
+
+    Public Overloads Property Visible As Boolean
+        Get
+            Return chkCheck.Visible
+        End Get
+        Set(bVisible As Boolean)
+            chkCheck.Visible = bVisible
+        End Set
+    End Property
```
lloyddewit, line 190: I assume that this change has no functional effect and that it was implemented just to be consistent with other `ucrCheck` overloaded properties.
Is that correct?
I have no objection to this change, I just wanted to make sure I understood it, thanks

lloyddewit (approved):

lloyddewit:
@rdstern On 26 Oct (above) you suggested that if the labels checkbox was hidden then this PR could be merged. @shadrackkibet has now hidden the labels checkbox. I reviewed the code and it looks safe.
Please could you re-approve? thanks

instat/ucrCheck.vb
shadrackkibet: Correct.

rdstern (approved):
This is fine

PR #5902, opened 2020-07-15
Adding scale date to all graphics dialogue

Ivanluv:
this fixes  the  errors in the graphics dialogue  raise in PR #5803 and #5896

### DIFF
--- a/instat/dlgBarAndPieChart.vb
+++ b/instat/dlgBarAndPieChart.vb
@@ -39,6 +39,8 @@ Public Class dlgBarAndPieChart
     Private bResetBarLayerSubdialog As Boolean = True
     Private clsCoordPolarFunction As New RFunction
     Private clsCoordPolarStartOperator As New ROperator
+    Private clsXScaleDateFunction As New RFunction
+    Private clsYScaleDateFunction As New RFunction
 
     Private Sub cmdOptions_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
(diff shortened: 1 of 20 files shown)


lloyddewit:
@shadrackkibet Please could you peer review? thanks

instat/dlgBarAndPieChart.vb
```diff
@@ -192,6 +194,8 @@ Public Class dlgBarAndPieChart
         clsBaseOperator.AddParameter(GgplotDefaults.clsDefaultThemeParameter.Clone())
         clsCoordPolarStartOperator = GgplotDefaults.clsCoordPolarStartOperator.Clone()
         clsCoordPolarFunction = GgplotDefaults.clsCoordPolarFunction.Clone()
+        clsXScaleDateFunction = GgplotDefaults.clsXScaleDateFunction.Clone()
+        clsYScaleDateFunction = GgplotDefaults.clsYScaleDateFunction.Clone()
```
shadrackkibet, line 198: @Ivanluv this is a question. Why is this needed in the main dialog? Is it appropriate to clone these function from the sub dialog because that is where these functions are updated? I do not see where it is used in the main dialog rather than just being passed to the sub dialog. I think if you do it in the sub dialog you will have to do in one class rather than all these 20 classes.

Ivanluv: main dialogue may need to add parameters to the function

shadrackkibet: Okay, technically that is possible. But in all these dialogs I can't see where you added parameters to these functions. Perhaps you could point me to the line that is done in the main dialogs.

Ivanluv: for the dlgPICSARainfall dialogue  addition of parameters  to `clsYScaleDateFunction` was  done on the main dialogue

shadrackkibet: @Ivanluv apart from PICSARainfall graph dialog is there any other dialog? PICSA Rainfall graph dialog is a very specific kind of dialog.  y_axis, x_axis and also other options can be modified from two sub dialogs, **Plot options** and **PICSA options** sub dialogs, which means these options should be in sync. However, this doesn't prevent moving of these code into `ucrAxes` where it belongs!

@rdstern I question why we need **Plot Options** subdialog here, this is a very specific dialog which means we should make those options that are required for PICSA graphs in the **PICSA options**  subdialog, otherwise, we don't need **PICSA options** sub dialog. I don't know the long terms plans for this but I can clearly see that in future will not need both.

rdstern:
@shadrackkibet and @Ivanluv I have provisionally approved, so this could just depend on approval by @lloyddewit .  But are there other dialogues where I should check the sub-dialogue first.  I would really like this change to be merged, unless it may cause new problems?

shadrackkibet:
@lloyddewit if you and @rdstern are happy that this is working fine, I am aware you are about to leave so I suggest this be merged as it is for the new version. I further suggest @Ivanluv to open an issue and link it with this PR then we can address the code issues later.

PR #5914, opened 2020-07-30
Removed automatic selection of the first level of a factor when filtering and elsewhere

shadrackkibet:
@rdstern I didn't find an issue for this but I remember this is something you mentioned at some point during our discussions. This change removes the selection of the first level of a factor when filtering for convenience purpose. Please test.

**Note:** This affects other places where this control is used, my assumption is that we do not need to always select the first level.

### DIFF
--- a/instat/ucrFactor.vb
+++ b/instat/ucrFactor.vb
@@ -279,7 +279,7 @@ Public Class ucrFactor
                 If strSelectedLevels IsNot Nothing AndAlso strSelectedLevels.Count > 0 Then
                     bSelect = strSelectedLevels.Contains(shtCurrSheet(i, 0))
                 Else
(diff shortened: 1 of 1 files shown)


rdstern:
Sounds good.  Will check.  One other (smaller) item in the filter sub-dialogue is that the control with the function ==, > etc doesn't get remembered between successive uses of the dialogue.

And really pleased you are back to this dialogue.

shadrackkibet:
> Sounds good. Will check. One other (smaller) item in the filter sub-dialogue is that the control with the function ==, > etc doesn't get remembered between successive uses of the dialogue.
> 
> And really pleased you are back to this dialogue.

I will fix this on a separate PR.

PR #5915, opened 2020-08-03
Retaining an operator(==,> etc) on filter input

shadrackkibet:
@rdstern this PR fixes a small issue you reported in PR #5914. Please check, I might have misunderstood your statement below.

> Sounds good. Will check. One other (smaller) item in the filter sub-dialogue is that the control with the function ==, > etc doesn't get remembered between successive uses of the dialogue.

### DIFF
--- a/instat/ucrFilter.vb
+++ b/instat/ucrFilter.vb
@@ -160,8 +160,7 @@ Public Class ucrFilter
 
     Private Sub ucrFilterReceiver_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrFilterByReceiver.ControlContentsChanged
         'for logical columns add {"==", "is.na", "!is.na"} only
-        Dim selectedIndex As Integer
-        selectedIndex = ucrFilterOperation.cboInput.SelectedIndex
+        Dim selectedIndex As Integer = ucrFilterOperation.cboInput.SelectedIndex
         ucrFilterOperation.SetItems(If(ucrFilterByReceiver.strCurrDataType.ToLower = "logical", {"==", "is.na", "! is.na"}, {"==", "<", "<=", ">", ">=", "!=", "is.na", "! is.na"}))
         If ucrFilterByReceiver.strCurrDataType.ToLower = "logical" AndAlso selectedIndex > 2 Then
             ucrFilterOperation.cboInput.SelectedIndex = 0
--- a/instat/ucrFilter.vb
+++ b/instat/ucrFilter.vb
@@ -160,9 +160,14 @@ Public Class ucrFilter
 
     Private Sub ucrFilterReceiver_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrFilterByReceiver.ControlContentsChanged
         'for logical columns add {"==", "is.na", "!is.na"} only
+        Dim selectedIndex As Integer
+        selectedIndex = ucrFilterOperation.cboInput.SelectedIndex
         ucrFilterOperation.SetItems(If(ucrFilterByReceiver.strCurrDataType.ToLower = "logical", {"==", "is.na", "! is.na"}, {"==", "<", "<=", ">", ">=", "!=", "is.na", "! is.na"}))
-        ucrFilterOperation.cboInput.SelectedIndex = 0
-
+        If ucrFilterByReceiver.strCurrDataType.ToLower = "logical" AndAlso selectedIndex > 2 Then
+            ucrFilterOperation.cboInput.SelectedIndex = 0
+        Else
+            ucrFilterOperation.cboInput.SelectedIndex = selectedIndex
+        End If
         VariableTypeProperties()
         CheckAddEnabled()
     End Sub
--- a/instat/ucrFilter.vb
+++ b/instat/ucrFilter.vb
@@ -162,11 +162,10 @@ Public Class ucrFilter
         'for logical columns add {"==", "is.na", "!is.na"} only
         Dim selectedIndex As Integer = ucrFilterOperation.GetSetSelectedIndex
         ucrFilterOperation.SetItems(If(ucrFilterByReceiver.strCurrDataType.ToLower = "logical", {"==", "is.na", "! is.na"}, {"==", "<", "<=", ">", ">=", "!=", "is.na", "! is.na"}))
-        Dim ItemsCount As Integer = ucrFilterOperation.GetItemsCount()
-        If ucrFilterByReceiver.strCurrDataType.ToLower = "logical" AndAlso selectedIndex > ItemsCount Then
-            ucrFilterOperation.GetSetSelectedIndex = 0
-        Else
+        If ucrFilterByReceiver.strCurrDataType.ToLower IsNot "logical" OrElse selectedIndex < ucrFilterOperation.GetItemsCount Then
             ucrFilterOperation.GetSetSelectedIndex = selectedIndex
+        Else
+            ucrFilterOperation.GetSetSelectedIndex = 0
         End If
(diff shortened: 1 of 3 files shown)


shadrackkibet:
@rdstern I am hoping the functionality requested above is now working. Please check.

shadrackkibet:
@rdstern another issue I found and now fixed here is that items of history are not unique. Items are added even if it already existed in history. I presume an item should not be added if it is already in history. Let me know if the previous behaviour is needed then I can revert to it.

shadrackkibet:
@lloyddewit I would like to get this merged soon so that I can continue with other tasks involving the filter.

shadrackkibet:
@lloyddewit I would like to get this merged soon so that I can continue with other tasks involving the filter.

instat/ucrFilter.vb
```diff
@@ -160,9 +160,14 @@ Public Class ucrFilter
 
     Private Sub ucrFilterReceiver_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrFilterByReceiver.ControlContentsChanged
         'for logical columns add {"==", "is.na", "!is.na"} only
+        Dim selectedIndex As Integer
+        selectedIndex = ucrFilterOperation.cboInput.SelectedIndex
```
lloyddewit, line 164: combine into a single line?

instat/ucrFilter.vb
```diff
@@ -160,9 +160,14 @@ Public Class ucrFilter
 
     Private Sub ucrFilterReceiver_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrFilterByReceiver.ControlContentsChanged
         'for logical columns add {"==", "is.na", "!is.na"} only
+        Dim selectedIndex As Integer
+        selectedIndex = ucrFilterOperation.cboInput.SelectedIndex
         ucrFilterOperation.SetItems(If(ucrFilterByReceiver.strCurrDataType.ToLower = "logical", {"==", "is.na", "! is.na"}, {"==", "<", "<=", ">", ">=", "!=", "is.na", "! is.na"}))
-        ucrFilterOperation.cboInput.SelectedIndex = 0
-
+        If ucrFilterByReceiver.strCurrDataType.ToLower = "logical" AndAlso selectedIndex > 2 Then
```
lloyddewit, line 166: It's dangerous to hard-code `2` here. It would be better to define a constant for `{"==", "is.na", "! is.na"}`and then check that the index is not greater than the number of items in the constant.
If you define a constant for the logical comparisons then, to be consistent, you should also define a constant for `{"==", "<", "<=", ">", ">=", "!=", "is.na", "! is.na"}`.

instat/ucrFilter.vb
shadrackkibet: Sometimes we have such declarations in one/two lines. Which one should we encourage?

lloyddewit: In this case I find 1 line more readable. What's your view?

shadrackkibet: No preference for me. Happy to change, I am just concerned with what we want to always encourage.

lloyddewit: I vote for 1 line :)
I agree that we should also consider consistency with what's been done elsewhere in the code, even if we don't personally agree with it. We don't have any R-Instat specific coding standards so we'll just have to apply common sense and be flexible.

shadrackkibet: ok

shadrackkibet: ok

instat/ucrFilter.vb
shadrackkibet: What do you think about this?

**Class ucrFilter**
```
        Dim selectedIndex As Integer = ucrFilterOperation.cboInput.SelectedIndex
        ucrFilterOperation.SetItems(If(ucrFilterByReceiver.strCurrDataType.ToLower = "logical", {"==", "is.na", "! is.na"}, {"==", "<", "<=", ">", ">=", "!=", "is.na", "! is.na"}))

        Dim ItemsCount As Integer = ucrFilterOperation.ItemsCount()

        If ucrFilterByReceiver.strCurrDataType.ToLower = "logical" AndAlso selectedIndex > ItemsCount - 1 Then
            ucrFilterOperation.cboInput.SelectedIndex = 0
        Else
            ucrFilterOperation.cboInput.SelectedIndex = selectedIndex
        End If
```
**Class ucrInputComboBox**

```
  Public Function ItemsCount() As Integer
        Return cboInput.Items.Count
    End Function
```

instat/ucrFilter.vb
```diff
@@ -160,9 +160,14 @@ Public Class ucrFilter
...
+        Dim selectedIndex As Integer = ucrFilterOperation.GetSetSelectedIndex
         ucrFilterOperation.SetItems(If(ucrFilterByReceiver.strCurrDataType.ToLower = "logical", {"==", "is.na", "! is.na"}, {"==", "<", "<=", ">", ">=", "!=", "is.na", "! is.na"}))
-        ucrFilterOperation.cboInput.SelectedIndex = 0
-
+        Dim ItemsCount As Integer = ucrFilterOperation.GetItemsCount()
+        If ucrFilterByReceiver.strCurrDataType.ToLower = "logical" AndAlso selectedIndex > ItemsCount Then
+            ucrFilterOperation.GetSetSelectedIndex = 0
+        Else
+            ucrFilterOperation.GetSetSelectedIndex = selectedIndex
+        End If
```
lloyddewit, line 170: I like your approach. 

Should the condition be `selectedIndex > (ItemsCount -1)`?
To make it more readable we could invert the condition.
We could also inline `ItemsCount`.
```suggestion
        If ucrFilterByReceiver.strCurrDataType.ToLower IsNot "logical" OrElse selectedIndex < ucrFilterOperation.GetItemsCount Then
            ucrFilterOperation.GetSetSelectedIndex = selectedIndex
        Else
            ucrFilterOperation.GetSetSelectedIndex = 0
        End If
```

instat/ucrFilter.vb
```diff
@@ -160,9 +160,14 @@ Public Class ucrFilter
 
     Private Sub ucrFilterReceiver_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrFilterByReceiver.ControlContentsChanged
         'for logical columns add {"==", "is.na", "!is.na"} only
+        Dim selectedIndex As Integer = ucrFilterOperation.GetSetSelectedIndex
```
lloyddewit, line 163: Do we need this variable? Could we use `ucrFilterOperation.GetSetSelectedIndex` directly in the code below? Or does the selected index value change when `ucrFilterOperation.SetItems` is called?

instat/ucrInputComboBox.vb
```diff
@@ -161,6 +161,19 @@ Public Class ucrInputComboBox
...
+            Return cboInput.SelectedIndex
+        End Get
+        Set(value As Integer)
+            cboInput.SelectedIndex = value
+        End Set
+    End Property
+
+    Public Function GetItemsCount() As Integer
+        Return cboInput.Items.Count
+    End Function
```
lloyddewit, line 175: I really like your use of Property above. Can we do something similar here?
```suggestion
    Public ReadOnly Property GetItemsCount As Integer
        Get
            Return cboInput.Items.Count
        End Get
    End Property
```

instat/ucrFilter.vb
shadrackkibet: Yes, it changes.

lloyddewit:
@rdstern the code has changed but the functionality should be the same. Please could you retest and approve? thanks

PR #5921, opened 2020-08-08
Enabling use of & and | operators in the filter 

### LINKED ISSUE #5791: Add the or condition to the filter sub-dialogue
This is a small task, which I propose be done by @Wycklife, with support (if needed - it may not be) from @shadrackkibet .

It is simply to enable the option of the or condition to be an alternative to the current and condition, when there are multiple conditions in a filter.  Here is my suggestion on the Filter sub-dialogue.

image

a) Add a button above the 3 given in the dialogue.  Default is `Combine with and`. If you press it then it says `Combine with or`.  The **_and_** & **_or_** should be in italic or bold, or both.  Perhaps we should say `Combine with &` and `Combine with | `instead?  Let's start with that.  
b) We could also have a tooltip that says "With more than one condition, e.g. (year > 1990) & (year < 2021) they have all to be TRUE"  Could we program it so the tooltip depends on the setting of this button?  So if `or` is selected it says instead "With more than one condition, e.g. (sunhrs >14) | (sunhrs <0) then just one need be TRUE". 
c) Make the 4 buttons all the same size and with equal spacing.  Make them slightly wider so each is on a single line.  Then they don't need to be quite so deep.
d) The new & and | button is only enabled when there is more than one condition.  (If you can't enable a button, then fine - it just doesn't do anything when there is a single condition.

shadrackkibet:
Fixes #5791. @rdstern this is ready for you to test. This is the implementation of the use of **&** and **|** operators in the filter.

### DIFF
--- a/instat/ucrFilter.vb
+++ b/instat/ucrFilter.vb
@@ -369,7 +369,13 @@ Public Class ucrFilter
         ucrReceiverExpression.Clear()
     End Sub
 
-    ''' <summary> Handles event triggered when cmdCombineWithAndOr Button is clicked.</summary>
+    ''' <summary> Handles event triggered when cmdCombineWithAndOr Button is clicked.
+    ''' The purpose of this button is to change the filter operator which to either AND(&amp;) or OR(|).
+    ''' The user clicks the button to trigger a change in the operator. 
+    ''' This subroutine changes the oparator, passes "and_or" parameter into clsFilterFunction, updates the button label and set the tooltip text for the button.
+    ''' Filter preview is also updated.
+    ''' Button text and the tooltip displayed on hover is changed whenever the button is clicked.
+    ''' </summary>
     ''' <param name="sender"> Not used. </param>
     ''' <param name="e"> Not used. </param>
     Private Sub cmdCombineWithAndOr_Click(sender As Object, e As EventArgs) Handles cmdCombineWithAndOr.Click
--- a/instat/ucrFilter.vb
+++ b/instat/ucrFilter.vb
@@ -372,7 +372,8 @@ Public Class ucrFilter
     ''' <summary> Handles event triggered when cmdCombineWithAndOr Button is clicked.
     ''' The purpose of this button is to change the filter operator to either AND(&amp;) or OR(|).
     ''' The user clicks the button to trigger a change in the operator. 
-    ''' This subroutine changes the oparator, passes "and_or" parameter into clsFilterFunction, updates the button label and set the tooltip text for the button.
+    ''' This subroutine changes the oparator, passes "and_or" parameter into clsFilterFunction, 
+    ''' updates the button label and set the tooltip text for the button.
     ''' Filter preview is also updated.
     ''' Button text and the tooltip displayed on hover is changed whenever the button is clicked.
     ''' </summary>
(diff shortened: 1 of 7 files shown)


shadrackkibet:
@rdstern this is ready for you to test.

rdstern:
ok @shadrackkibet .  Will do.

shadrackkibet:
I find it useful to be enabled when there is no or one condition. The user will be able to change a condition so that any further conditions added will be combined using that specified condition.

shadrackkibet:
@rdstern I would like to get this merged soon so that I can open another PR with implementation on filters from factor levels.

lloyddewit:

Ivanluv:
@lloyddewit  am on this

shadrackkibet:
Is there anything for me to look into here? otherwise, this can be merged.

Ivanluv:
this is looking fine from my review

instat/ucrFilter.vb
```diff
@@ -368,4 +368,19 @@ Public Class ucrFilter
     Private Sub cmdClear_Click(sender As Object, e As EventArgs) Handles cmdClear.Click
         ucrReceiverExpression.Clear()
     End Sub
+
+    Private Sub cmdCombineWithAndOr_Click(sender As Object, e As EventArgs) Handles cmdCombineWithAndOr.Click
```
lloyddewit, line 372: I think we should add XML comment blocks for new functions.

Please add one for this function of the form:
```
    ''' <summary>   Handles event triggered when ... </summary>
    '''
    ''' <param name="sender">    Not used. </param>
    ''' <param name="e">    Not used. </param>
```

lloyddewit: Thanks for the quick change!
The summary currently gives the same information as the subroutine declaration, so it doesn't add value. 
The next developer maintaining this code will be interested to know things like:
- What is the purpose of this button?
- How/why will the user use it?
- What specific things does this subroutine do when this button is clicked? 
- Is the action of the button different under different circumstances?

Is it possible to provide this type of information in the comment summary?

shadrackkibet: This is the first time I add such XML comments into R-Instat code which I have seen now added in other places. Personally I am not a fan of commenting everything in code I prefer to make code readable as much as I can which is not easy sometimes or might need more time, that is when I find it useful to leave comments. If the maintainer wants to get answers to such kind of questions then reading the code is better. This is just my opinion,  but happy to add this information to the summary as you suggest.

lloyddewit: @shadrackkibet Thank you for the great comments. I agree with you that the ideal is to have code so readable that additional in-line comments aren't needed. The XML comment blocks still add value, even to the most readable code, because Intellisense displays the XML comment when a developer hovers the mouse over a function name (plus of course there are other potential benefits like auto doc generation).
I think we should normally add XML comments to new functions (unless they're standard functions used in all dialogs). I'm trying to consistently ask for these in peer reviews.

instat/ucrFilter.vb
```diff
@@ -368,4 +368,28 @@ Public Class ucrFilter
     Private Sub cmdClear_Click(sender As Object, e As EventArgs) Handles cmdClear.Click
         ucrReceiverExpression.Clear()
     End Sub
+
+    ''' <summary> Handles event triggered when cmdCombineWithAndOr Button is clicked.
+    ''' The purpose of this button is to change the filter operator to either AND(&amp;) or OR(|).
+    ''' The user clicks the button to trigger a change in the operator. 
+    ''' This subroutine changes the oparator, passes "and_or" parameter into clsFilterFunction, updates the button label and set the tooltip text for the button.
```
lloyddewit, line 375:
```suggestion
    ''' This subroutine changes the oparator, passes "and_or" parameter into clsFilterFunction, 
    ''' updates the button label and set the tooltip text for the button.
```

lloyddewit:
@rdstern Please could you approve? thanks

PR #5932, opened 2020-08-17
Generating filters from factor levels

### LINKED ISSUE #4800: New filter from factor sub-dialogue

shadrackkibet:
Fixes #4800. @rdstern this is ready for you to test. This implements a new subdialog that generates filters from factor levels.

### DIFF
--- a/instat/dlgRestrict.resx
+++ b/instat/dlgRestrict.resx
@@ -225,9 +225,6 @@
   <data name="$this.ClientSize" type="System.Drawing.Size, System.Drawing">
     <value>418, 388</value>
   </data>
-  <data name="cmdFilterFromFactors.Enabled" type="System.Boolean, mscorlib">
-    <value>False</value>
-  </data>
   <data name="cmdFilterFromFactors.ImeMode" type="System.Windows.Forms.ImeMode, System.Windows.Forms">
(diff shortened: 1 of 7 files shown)


lloyddewit:
@Ivanluv please can you peer review? thanks

Ivanluv:
@lloyddewit  am reviewing this now

Ivanluv:
@lloyddewit  the code looks okay from my end

lloyddewit (approved):

PR #5900, opened 2020-07-14
Adding the climatic extremes value dialogue 

Ivanluv:
fixes #5887

PR #5933, opened 2020-08-17
Changes for the extremes Dialogue

### LINKED ISSUE #5887: Adding an extreme value dialogue
This is already provided under Climatic > Model > Extremes.

But that dialogue is simply copied from the old Instat and almost certainly the code and dialogue need to be very different.  

This will almost certainly use the fevd function from the extRemes package.  It allows trends etc and so may be "Interesting" to construct.  
We may also call other functions to use the resulting object, in the same dialogue, or have another new dialogue to Use Extreme Model.  Then this one would become Fit Extreme Model.

We need to ensure that the structure of this (or these) dialogues is consistent with other dialogues in R-Instat.  We may also find that the structure of the equivalent dialogues in the front-end to the extRemes package is helpful?

Ivanluv:
@africanmathsinitiative/developers  this fixes #5887 and it from PR #5900


Ivanluv:
@shadrackkibet  you can now review this

instat/sdgExtremesDisplayOptions.vb
```diff
@@ -1,3 +1,67 @@
...
+        clsRsyntax = clsNewRSyntax
+        clsFevdPlotFunction = clsNewFevdPlotFunction
+
+        ucrPnlExtreme.SetRCode(clsFevdPlotFunction, bReset, bCloneIfNeeded:=True)
+        ucrSavePlots.SetRCode(clsFevdPlotFunction, bReset, bCloneIfNeeded:=True)
+    End Sub
+
+    Private Sub ucrPnlExtreme_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlExtreme.ControlValueChanged
+        If rdoNoPlot.Checked Then
+            clsFevdPlotFunction.RemoveParameterByName("type")
```
shadrackkibet, line 61: This function should not run when this radio button is selected. Therefore no need to remove this parameter.

instat/sdgExtremesDisplayOptions.vb
```diff
@@ -1,3 +1,67 @@
-﻿Public Class sdgExtremesDisplayOptions
+﻿Imports instat
+Imports instat.Translations
+Public Class sdgExtremesDisplayOptions
```
shadrackkibet, line 3: The licence text is needed here.

instat/sdgExtremesMethod.vb
```diff
@@ -1,3 +1,20 @@
 ﻿Public Class sdgExtremesMethod
```
shadrackkibet, line 1: Licence text needed here

instat/sdgExtremesMethod.vb
```diff
@@ -1,3 +1,20 @@
...
+    Private clsFevdFunction As New RFunction
+    Public bControlsInitialised As Boolean = False
+    Public Sub InitialiseControls()
+        ucrPanelExtreme.SetParameter(New RParameter("method"))
```
shadrackkibet, line 5: Parameter position?

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports instat
```
shadrackkibet, line 17: This is not needed.

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
+    Private clsDetach As New RFunction
+    Private clsFevdFunction, clsNaExclude, clsConvertVector, clsFevdPlotsFunction As New RFunction
+    'clsLocationScaleResetOperator is not run but affects reset of the check box.Any better method of implementation?
+    Private clsLocationScaleResetOperator As New ROperator
+    Private clsLocationParamOperator As New ROperator
+    Private bFirstLoad As Boolean = True
+    Private bReset As Boolean = True
+    Private bResetFittingOptions As Boolean = False
+    Private bResetDisplayOptions As Boolean = False
+    Private bResettingDialogue As Boolean = False
```
shadrackkibet, line 31: Why is this needed?

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
+        ucrReceiverExpressionExplanatoryModelForLocParam.SetParameterIsString()
+        ucrReceiverExpressionExplanatoryModelForLocParam.SetRDefault(1)
+        ucrReceiverExpressionExplanatoryModelForLocParam.bWithQuotes = False
+
+        ucrTryModelling.SetReceiver(ucrReceiverExpressionExplanatoryModelForLocParam)
+        ucrTryModelling.SetIsModel()
+
+        ucrSaveExtremes.SetPrefix("extreme")
+        ucrSaveExtremes.SetIsComboBox()
+        ucrSaveExtremes.SetCheckBoxText("Save Model")
```
shadrackkibet, line 95:
```suggestion
        ucrSaveExtremes.SetCheckBoxText("Save Model:")

```

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
+        clsNaExclude = New RFunction
+        clsConvertVector = New RFunction
+        clsLocationParamOperator = New ROperator
+        clsLocationScaleResetOperator = New ROperator
+
+        ucrBase.clsRsyntax.ClearCodes()
+
+        ucrReceiverVariable.SetMeAsReceiver()
+        ucrSelectorExtremes.Reset()
+        ucrInputThresholdforLocation.SetText("0")
```
shadrackkibet, line 113: This is not needed.

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
+        ucrReceiverExpressionExplanatoryModelForLocParam.Selector = ucrSelectorExtremes
```
shadrackkibet, line 114: This should be done on `InitialiseControls()`

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
+
+        clsLocationScaleResetOperator.SetOperation("")
```
shadrackkibet, line 116: Why do you need an empty operation?

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
+        clsLocationScaleResetOperator.bBrackets = False
+
+        clsLocationParamOperator.SetOperation("~")
+        clsLocationParamOperator.AddParameter(strParameterValue:="", iPosition:=0, bIncludeArgumentName:=False)
```
shadrackkibet, line 120: why is this needed?

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
+
+        clsFevdFunction.SetAssignTo(ucrSaveExtremes.GetText(), strTempDataframe:=ucrSelectorExtremes.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempModel:="last_model", bAssignToIsPrefix:=True)
+        clsFevdPlotsFunction.SetAssignTo("last_graph", strTempDataframe:=ucrSelectorExtremes.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempGraph:="last_graph")
+        clsFevdPlotsFunction.AddParameter("x", clsRFunctionParameter:=clsFevdFunction, iPosition:=0)
+
+        clsAttach.SetRCommand("attach")
+        clsDetach.SetRCommand("detach")
+        clsAttach.AddParameter("what", clsRFunctionParameter:=ucrSelectorExtremes.ucrAvailableDataFrames.clsCurrDataFrame)
+        clsDetach.AddParameter("name", clsRFunctionParameter:=ucrSelectorExtremes.ucrAvailableDataFrames.clsCurrDataFrame)
+        clsDetach.AddParameter("unload", "TRUE")
```
shadrackkibet, line 150: positions?

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
 
-    Private Sub chkRestrictValues_CheckedChanged(sender As Object, e As EventArgs) Handles chkRestrictValues.CheckedChanged
-        If chkRestrictValues.Checked = True Then
-            lblBetween.Visible = True
-            txtBetween.Visible = True
-            lblAnd.Visible = True
-            txtAnd.Visible = True
+    Private Sub cmdDisplayOptions_Click(sender As Object, e As EventArgs) Handles cmdDisplayOptions.Click
+        sdgExtremesDisplayOptions.SetRCode(clsNewFevdPlotFunction:=clsFevdPlotsFunction, clsNewRSyntax:=ucrBase.clsRsyntax)
+        bResetDisplayOptions = False
```
shadrackkibet, line 178: This should be done after `dgExtremesDisplayOptions.ShowDialog()`

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
         End If
     End Sub
 
-    Private Sub chkMaximumLikelihood_CheckedChanged(sender As Object, e As EventArgs) Handles chkMaximumLikelihood.CheckedChanged
-        If chkMaximumLikelihood.Checked = True Then
-            cboMaximumLikelihood.Visible = True
-        End If
+    Private Sub cmdFittingOptions_Click(sender As Object, e As EventArgs) Handles cmdFittingOptions.Click
+        sdgExtremesMethod.SetRCode(clsNewFevdFunction:=clsFevdFunction, bReset:=bResetFittingOptions)
+        bResetFittingOptions = False
```
shadrackkibet, line 192: also here

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
+    Private Sub cmdExp_Click(sender As Object, e As EventArgs) Handles cmdExp.Click
+        ucrReceiverExpressionExplanatoryModelForLocParam.AddToReceiverAtCursorPosition("exp( )", 2)
+    End Sub
+
+    Private Sub cmdTan_Click(sender As Object, e As EventArgs) Handles cmdTan.Click
+        ucrReceiverExpressionExplanatoryModelForLocParam.AddToReceiverAtCursorPosition("tan( )", 2)
+    End Sub
+
+
+
```
shadrackkibet, line 265: single space

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
+    Private Sub ucrReceiverVariable_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverVariable.ControlValueChanged
+        TestOkEnabled()
+    End Sub
+
+    Private Sub control_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputExtremes.ControlValueChanged, ucrReceiverVariable.ControlValueChanged, ucrChkExplanatoryModelForLocationParameter.ControlValueChanged, ucrInputThresholdforLocation.ControlValueChanged
+        ParameterControl()
+        ucrTryModelling.ucrInputTryMessage.SetName("")
+        ucrTryModelling.ucrInputTryMessage.txtInput.BackColor = Color.White
+        ucrTryModelling.ucrInputTryMessage.txtInput.Controls.Clear()
```
shadrackkibet, line 274: We should have methods in `ucrTry` class to do this.

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
+            If Not bResettingDialogue Then
+                clsLocationScaleResetOperator.AddParameter("scaleLocation", clsROperatorParameter:=clsLocationParamOperator, iPosition:=0)
+            End If
+        Else
+            ucrReceiverVariable.SetMeAsReceiver()
+            clsFevdFunction.RemoveParameterByName("scale.fun")
+            clsFevdFunction.RemoveParameterByName("location.fun")
+            clsLocationScaleResetOperator.RemoveParameterByName("scaleLocation")
+        End If
+        HideAndShowKeyboard()
```
shadrackkibet, line 302: This is only used here, can you inline code here.

instat/sdgExtremesDisplayOptions.vb
```diff
@@ -1,3 +1,67 @@
...
+        ucrPnlExtreme.AddParameterValuesCondition(rdoQQ2, "type", Chr(34) & "qq2" & Chr(34))
+        ucrPnlExtreme.AddParameterValuesCondition(rdoQQPlot, "type", Chr(34) & "qq" & Chr(34))
+        ucrPnlExtreme.AddParameterValuesCondition(rdoRlplot, "type", Chr(34) & "rl" & Chr(34))
+        ucrPnlExtreme.AddParameterValuesCondition(rdoZPlot, "type", Chr(34) & "Zplot" & Chr(34))
+        ucrPnlExtreme.AddParameterValuesCondition(rdoTrace, "type", Chr(34) & "trace" & Chr(34))
+
+        ucrSavePlots.SetPrefix("plot")
+        ucrSavePlots.SetSaveTypeAsGraph()
+        ucrSavePlots.SetDataFrameSelector(dlgExtremes.ucrSelectorExtremes.ucrAvailableDataFrames)
+        ucrSavePlots.SetCheckBoxText("Save Graph")
```
shadrackkibet, line 40:
```suggestion
        ucrSavePlots.SetCheckBoxText("Save Graph:")
```

instat/dlgExtremes.vb
```diff
@@ -14,64 +14,301 @@
...
+        bResetFittingOptions = True
+    End Sub
+
+    Private Sub SetRCodeForControls(bReset As Boolean)
+        ucrInputExtremes.SetRCode(clsFevdFunction, bReset)
+        ucrReceiverVariable.SetRCode(clsNaExclude, bReset)
+        ucrSaveExtremes.SetRCode(clsFevdFunction, bReset)
+        ucrChkExplanatoryModelForLocationParameter.SetRCode(clsLocationScaleResetOperator, bReset)
+        ucrReceiverExpressionExplanatoryModelForLocParam.SetRCode(clsLocationParamOperator, bReset)
+    End Sub
```
shadrackkibet, line 167: space here

instat/sdgExtremesDisplayOptions.vb
Ivanluv: This helps in setting the conditions

lloyddewit:
@Ivanluv There are still some open peer review comments above. Please could you clarify whether they're already resolved, or if you still plan to accept/reject them? thanks

instat/dlgExtremes.vb
Ivanluv: Used in Setting  the RCode for the explanatory model check box

instat/dlgExtremes.vb
Ivanluv: the operator required is, ~"value" ,the operation has no parameter on the left

instat/dlgExtremes.vb
Ivanluv: The parameter is added by the explanatory check box and the checkbox is unchecked by default

Ivanluv:
@shadrackkibet could you review this again

shadrackkibet:
- Looks better. Can you try to improve the design, alignments, naming etc? Otherwise, @lilyclements might find some time to help. See below.
image
image
image

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,287 @@
...
+        clsFevdPlotsFunction.SetPackageName("extRemes")
+        clsFevdPlotsFunction.SetRCommand("plot.fevd")
+        clsFevdPlotsFunction.iCallType = 3
+        clsFevdPlotsFunction.bExcludeAssignedFunctionOutput = False
+
+        clsFevdFunction.SetPackageName("extRemes")
+        clsFevdFunction.SetRCommand("fevd")
+
+
+        clsConvertVector.SetRCommand("as.vector")
```
shadrackkibet, line 133: Why do you need this function?

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,287 @@
...
+
+        clsLocationParamOperator.SetOperation("~")
+        clsLocationParamOperator.AddParameter(strParameterValue:="", iPosition:=0, bIncludeArgumentName:=False)
+        clsLocationParamOperator.bSpaceAroundOperation = False
+
+        clsNaExclude.SetPackageName("stats")
+        clsNaExclude.SetRCommand("na.exclude")
+
+        clsFevdPlotsFunction.SetPackageName("extRemes")
+        clsFevdPlotsFunction.SetRCommand("plot.fevd")
```
shadrackkibet, line 125: Just `plot` is enough

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,281 @@
...
+            clsFevdFunction.RemoveParameterByName("scale.fun")
+            clsFevdFunction.RemoveParameterByName("location.fun")
+            clsLocationScaleResetOperator.RemoveParameterByName("scaleLocation")
+        End If
+        HideAndShowKeyboard()
     End Sub
 
-    Private Sub ucrReceiverDataToFit_Leave(sender As Object, e As EventArgs) Handles ucrReceiverDataToFit.Leave
-        ucrBase.clsRsyntax.AddParameter("xdat", clsRFunctionParameter:=ucrReceiverDataToFit.GetVariables())
+    Private Sub HideAndShowKeyboard()
```
Patowhiz, line 288: @Ivanluv though trivial and probably not necessary subroutine  `HideAndShowKeyboard()` could be simplified to;

```
Private Sub HideAndShowKeyboard()	
	With ucrChkExplanatoryModelForLocationParameter
	     grpFirstCalc.Visible = .Checked
             grpSecondCalc.Visible = .Checked
	End With
 End Sub
```

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,281 @@
...
-        If chkRestrictValues.Checked = True Then
-            lblBetween.Visible = True
-            txtBetween.Visible = True
-            lblAnd.Visible = True
-            txtAnd.Visible = True
+    Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
+        bResettingDialogue = True
+        SetDefaults()
+        SetRCodeForControls(True)
+        TestOkEnabled()
```
Patowhiz, line 162: @Ivanluv `TestOkEnabled()` will by default be called when `SetRCodeForControls(True)` is called(because `ControlContentsChanged` will be raised). This makes calling `TestOkEnabled()` here to be called twice. I noticed this when I put breakpoints in during runtime

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,281 @@
...
         autoTranslate(Me)
-        ucrReceiverDataToFit.Selector = ucrAddRemove
-        ucrReceiverDataToFit.SetMeAsReceiver()
-        ucrBase.iHelpTopicID = 488
+        TestOkEnabled()
+    End Sub
+
+    Private Sub InitialiseDialog()
+        grpFirstCalc.Visible = False
+        grpSecondCalc.Visible = False
```
Patowhiz, line 46: @Ivanluv I think the 2 lines
```
 grpFirstCalc.Visible = False
 grpSecondCalc.Visible = False
```

could be moved to the designer(set via visual studio properties palette) since they are called once during form load, meaning they are part of the first default state of the form. Also its clear that they are only set in subroutine `HideAndShowKeyboard()`, which is certainly called when contents of the controls concerned are changed.

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,281 @@
 Imports instat.Translations
 
 Public Class dlgExtremes
+    Private clsAttach As New RFunction
+    Private clsDetach As New RFunction
+    Private clsFevdFunction, clsNaExclude, clsPlotsFunction As New RFunction
+    'clsLocationScaleResetOperator is not run but affects reset of the check box.Any better method of implementation?
+    Private clsLocationScaleResetOperator As New ROperator
+    Private clsLocationParamOperator As New ROperator
```
Patowhiz, line 25: @Ivanluv though trivial. I think `clsFevdFunction`, `clsNaExclude`, `clsPlotsFunction`, `clsLocationScaleResetOperator`, and `clsLocationParamOperator `don't need to be initialised here. This is done in `SetDefaults()` and the subroutine is always called before they are used.

I also noticed `clsAttach `and `clsDetach` are not initialised in `SetDefaults()`

Patowhiz (commented):
@Ivanluv would also be helpful if comments you could add comments explaining certain parts of code that are not clear in first glance.

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,281 @@
...
+        If ucrInputExtremes.GetText() = "GP" OrElse ucrInputExtremes.GetText() = "PP" OrElse ucrInputExtremes.GetText() = "Exponential" Then
+            ucrInputThresholdforLocation.Visible = True
+            clsFevdFunction.AddParameter("threshold", ucrInputThresholdforLocation.GetText(), iPosition:=4)
         Else
-            cboMethodOfMoments.Visible = False
+            ucrInputThresholdforLocation.Visible = False
+            clsFevdFunction.RemoveParameterByName("threshold")
         End If
+        If ucrChkExplanatoryModelForLocationParameter.Checked Then
+            If (ucrInputExtremes.GetText() = "GP" OrElse ucrInputExtremes.GetText() = "PP" OrElse ucrInputExtremes.GetText() = "Exponential") Then
```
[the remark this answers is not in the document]
Ivanluv: Thanks for this suggestion

instat/dlgExtremes.vb
Ivanluv: True.Am making changes right away

lloyddewit:
@Ivanluv does the commit above resolve all the peer review comments above, or are you still working on this? thanks for clarifying

Ivanluv:
@lloyddewit  this resolves all the peer reviews

instat/dlgExtremes.vb
lloyddewit: @Ivanluv Sorry, I didn't understand your comment above. Could this sub be inlined?
```suggestion
        With ucrChkExplanatoryModelForLocationParameter
            grpFirstCalc.Visible = .Checked
            grpSecondCalc.Visible = .Checked
        End With
```

lloyddewit:
@Ivanluv I went through the previous peer review comments. Most of them seemed to be resolved, so I marked them as resolved. However, some still seem to be open. Please could you review all the comments above that are not marked as resolved?
If you reply to each comment (either accepting suggestion or giving reason for rejecting) then that would be very helpful, thanks

instat/dlgExtremes.vb
Ivanluv: True.I have made this change

instat/dlgExtremes.vb
Ivanluv: Done

instat/dlgExtremes.vb
Ivanluv: Done

instat/dlgExtremes.vb
Ivanluv: Done

instat/sdgExtremesDisplayOptions.vb
Ivanluv: I've made the change

instat/dlgExtremes.vb
Ivanluv: ` ucrInputThresholdforLocation` is not setting Rcode  so `ucrInputThresholdforLocation.SetText("0")` resets the text to zero on reseting .`ucrReceiverVariable.SetMeAsReceiver()` is also required on resetting since we have two receivers

instat/dlgExtremes.vb
Ivanluv: I have removed this since its not required

instat/dlgExtremes.vb
Ivanluv: Done

instat/dlgExtremes.vb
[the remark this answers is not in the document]
Ivanluv: Done

Ivanluv:
@lloyddewit  could you look at this again

instat/dlgExtremes.vb
lloyddewit: @Ivanluv Please could you resolve the comment above? thanks

instat/dlgExtremes.vb
lloyddewit: @Ivanluv I can't see the change in the code, please could you clarify? thanks

lloyddewit:
@Ivanluv I think there is one comment in `instat/dlgExtremes.vb` that is still open (see above). Please can you clarify? thanks

instat/dlgExtremes.vb
Ivanluv: Initialisation of `clsAttach` and `clsDetach` is done by ` clsAttach = New RFunction` and  `clsDetach = New RFunction` on lines 102 and 103.As for the whether the initialisation of the `clsFevdFunction`, `clsNaExclude`, `clsPlotsFunction`, `clsLocationScaleResetOperator` functions needs not to be done there am not sure

lloyddewit: @Ivanluv Thank you for the clarification about `clsAttach` and `clsDetach`.  Regarding lines 22 to 25 I agree with @Patowhiz that `New` is not needed here. But I don't want to delay because of this so I'm happy to approve.

lloyddewit (approved):

lloyddewit:
@rdstern Please could you test/approve? thanks

rdstern:
There are some important aspects that may need an @dannyparsons discussion.

This is the file I tried with:

[Niger4dly_extremes.zip](https://github.com/africanmathsinitiative/R-Instat/files/5367302/Niger4dly_extremes.zip)

a) In the summary data frame in these data, the y data (in variable max) has missing values.  This works if there is no model, but gives an error if a model is fitted.  The command tries to exclude the missing values, but doesn't work as given, perhaps because the y and x are then inconsistent.  If I filter to the non-missing values in max, then I can fit a model, say station or year, etc.

If that is what is needed, then I am happy for this to continue in that way.  But then perhaps don't do anything in the R code to the dependent variable, but include the default na.action=na.fail, so the error message is clearer.

b) I can have a graph and the default is a nice combined graph.  By default this appears in the output window.  The graph can (sort of) be saved in the dialogue, but it is a graph in the base graphics, so I don't think it does save anything - at least not that is recognised by R-Instat.

I still wonder if ggplotify can be used, so it becomes a ggplot instead.  Then all should work.

As it isn't saved, then I can't display the graph in the R viewer.  I have discussed this aspect in general with @Patowhiz .  However, in the current code I changed the plotting to the following:
```
plot(x=data_book$get_models(data_name="Data_by_year_station", model_name="extreme9"), type="primary")
last_graph <-recordPlot()
```
Can we use this in general and could this also enable the base graphs in general to be saved in objects and then plotted again in the future?

Ivanluv:
Am ready to complete to once  @dannyparsons  has responded

lloyddewit:
@dannyparsons is there anything you can do from your side to unblock this PR? thanks

shadrackkibet:
> a) In the summary data frame in these data, the y data (in variable max) has missing values. This works if there is no model, but gives an error if a model is fitted. The command tries to exclude the missing values, but doesn't work as given, perhaps because the y and x are then inconsistent. If I filter to the non-missing values in max, then I can fit a model, say station or year, etc.

@rdstern please give more details about this error. I am unable to replicate it. Perhaps post the error message as well.

rdstern (changes requested):
@shadrackkibet sorry for the delay in trying to answer this.  Now it needs updating I think.  It currently says that the version of R is incorrect.

rdstern:
@Ivanluv this needs updating I think.  It says it needs the new  version of R?

Ivanluv:
@rdstern I have updated the branch

rdstern (changes requested):
Here is the code for one example now:
extRemes::fevd(x = stats::na.exclude(object = max), location.fun = ~station + 
    year, type = "GEV", method = "MLE")
I'm not sure this is a good way to try to cope with the missing values, so perhaps that should be undone.  It still gives the same odd message when there are missing values in the data.  I notice now that it also can't cope with missing values in the x'variables.  We might just have to accept that for now.
A minor problem is that when the code, as in the example above, is on 2 lines, then it over-types in the title of the plot - at least for a combined plot.
Perhaps a quick look by @shadrackkibet could help.  Then we merge and improve later.

Ivanluv:
@Shadrack how can I solve  the over typing problem

shadrackkibet:
- The function `fevd` has the parameter `na.action` which is meant to deal with missing values. I think to avoid the error we need to exclude missing. Therefore you need to set `na.action = na.omit` or `na.action = na.exclude` . For example
`extRemes::fevd(x = max, na.action = na.exclude, location.fun = ~station +
year, type = "GEV", method = "MLE")` 
- I can't find an easy way to deal with the over-typing issue.

shadrackkibet:
@Ivanluv can we try to get this merged? anything remaining?

Ivanluv:
@shadrackkibet  everything is done except for the over-typing issue

rdstern:
The overtyping in the figure isn't important.  It could be merged now, or the problems with missing values could be investigated further.  Perhaps merge and have this as an issue?

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,275 @@
...
+        clsLocationParamOperator = New ROperator
+        clsLocationScaleResetOperator = New ROperator
+        clsAttach = New RFunction
+        clsDetach = New RFunction
+
+        ucrBase.clsRsyntax.ClearCodes()
+
+        ucrReceiverVariable.SetMeAsReceiver()
+        ucrSelectorExtremes.Reset()
+        ucrInputThresholdforLocation.SetText("0")
```
shadrackkibet, line 103: Also, reset the save control here.

shadrackkibet:
Only a few comments and then this should be ready. Since this is an initial implementation it will be good to get it in. Further improvements can be done after further testing.

- When I check the checkbox (Explanatory model for location or scale) and then reopen the dialog the controls below disappear.
- The save control doesn't reset.
- Rename the receiver "**Response Variable:**" which is consistent with **"Model>Fit Model>General"**.

image

Ivanluv:
@shadrackkibet  I have resolved the issues you raised. Could you look at this again

shadrackkibet:
Looks better now. Over to @lloyddewit for the next steps.

lloyddewit (changes requested):
@Ivanluv this looks good, just some small comments, thanks

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,277 @@
...
+    End Sub
+
+    Private Sub cmdTan_Click(sender As Object, e As EventArgs) Handles cmdTan.Click
+        ucrReceiverExpressionExplanatoryModelForLocParam.AddToReceiverAtCursorPosition("tan( )", 2)
     End Sub
 
-    Private Sub chkMethodofMoments_CheckedChanged(sender As Object, e As EventArgs) Handles chkMethodOfMoments.CheckedChanged
-        If chkMethodOfMoments.Checked = True Then
-            cboMethodOfMoments.Visible = True
+    Private Sub ucrReceiverVariable_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverVariable.ControlValueChanged
```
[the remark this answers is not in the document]
Ivanluv: Done

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,277 @@
...
+    Private Sub cmdFittingOptions_Click(sender As Object, e As EventArgs) Handles cmdFittingOptions.Click
+        sdgExtremesMethod.SetRCode(clsNewFevdFunction:=clsFevdFunction)
+        sdgExtremesMethod.ShowDialog()
+    End Sub
 
-    Private Sub chkMaximumLikelihood_CheckedChanged(sender As Object, e As EventArgs) Handles chkMaximumLikelihood.CheckedChanged
-        If chkMaximumLikelihood.Checked = True Then
-            cboMaximumLikelihood.Visible = True
-        End If
+    Private Sub cmdPlus_Click(sender As Object, e As EventArgs) Handles cmdPlus.Click
```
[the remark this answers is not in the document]
Ivanluv: Any suggestions on how to this?

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,277 @@
...
+    Private Sub TestOkEnabled()
+        If ucrReceiverVariable.IsEmpty Then
+            ucrBase.OKEnabled(False)
         Else
-            lblBetween.Visible = False
-            txtBetween.Visible = False
-            lblAnd.Visible = False
-            txtAnd.Visible = False
+            ucrBase.OKEnabled(True)
         End If
```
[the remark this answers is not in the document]
Ivanluv: Done

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,277 @@
...
 Public Class dlgExtremes
+    Private clsAttach As New RFunction
+    Private clsDetach As New RFunction
+    Private clsFevdFunction, clsPlotsFunction As New RFunction
+    'clsLocationScaleResetOperator is not run but affects reset of the check box.Any better method of implementation?
+    Private clsLocationScaleResetOperator As New ROperator
+    Private clsLocationParamOperator As New ROperator
+    Private bFirstLoad As Boolean = True
+    Private bReset As Boolean = True
+    Private bResettingDialogue As Boolean = False
```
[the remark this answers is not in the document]
Ivanluv: `bResettingDialogue  `prevents the addition of the `scaleLocation `parameter to the `clsLocationScaleResetOperator `operator on resetting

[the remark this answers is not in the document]
lloyddewit: Thank you, I took another look and I understand now (`ParameterControl` is called indirectly when `SetRCodeForControls` changes the value of the controls).

instat/dlgExtremes.vb
```diff
@@ -17,61 +17,277 @@
...
+    Private Sub SetRCodeForControls(bReset As Boolean)
+        ucrInputExtremes.SetRCode(clsFevdFunction, bReset)
+        ucrReceiverVariable.SetRCode(clsFevdFunction, bReset)
+        ucrSaveExtremes.SetRCode(clsFevdFunction, bReset)
+        ucrChkExplanatoryModelForLocationParameter.SetRCode(clsLocationScaleResetOperator, bReset)
+        ucrReceiverExpressionExplanatoryModelForLocParam.SetRCode(clsLocationParamOperator, bReset)
+    End Sub
+
+    Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
+        bResettingDialogue = True
```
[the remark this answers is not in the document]
lloyddewit: I understand now, please see my other comment below related to `bResettingDialogue `

instat/dlgExtremes.vb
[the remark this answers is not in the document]
lloyddewit: Would something like this work?
```
    Private Sub cmdButton_Click(sender As Object, e As EventArgs) Handles cmdColon.Click, cmdMultiply.Click
        Select Case DirectCast(sender, Button).Name
            Case "cmdColon"
                ucrReceiverExpressionExplanatoryModelForLocParam.AddToReceiverAtCursorPosition(":")
            Case "cmdMultiply"
                ucrReceiverExpressionExplanatoryModelForLocParam.AddToReceiverAtCursorPosition("*")
        End Select
    End Sub
```

lloyddewit:
@Ivanluv Thanks for all the changes, there is just one comment still open

instat/dlgExtremes.vb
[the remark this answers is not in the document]
shadrackkibet: @lloyddewit I think it is dangerous to use the name of the control among the cases here. If a developer changes the name of the control then this will definitely not work and it might not be easy to know exactly where names are used.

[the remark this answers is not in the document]
lloyddewit: > 
> 
> @lloyddewit I think it is dangerous to use the name of the control among the cases here. If a developer changes the name of the control then this will definitely not work and it might not be easy to know exactly where names are used.

@shadrackkibet thank you for the comment. The button event handlers all do roughly the same thing so I tried to think of a way to reduce the duplication. The `sender` parameter contains the type and current state of the calling object so I tried to use this information to identify which button was clicked. I agree that if, in future, a developer changes the control name then the suggested code would also need to be updated. I agree this reduces the code adaptability. I can't think of any better `sender` property we could use. I think vbdotnet expects you to derive a new `EventArgs` class and pass extra information about the event in the `e` parameter. However I think this is too complex for this situation.
In summary, I agree with you, and I can't think of a better solution for now. So I will resolve this comment and approve the PR.
Thanks again for the insightful comment.

lloyddewit (approved):

lloyddewit:

rdstern (approved):
Long-term my only confusion is with missing values.  But I see the comment for the fevd function is as follows:  "function to be called to handle missing values. Generally, this should remain at he default (na.fail), and the user should take care to impute missing values in
an appropriate manner as it may have serious consequences on the results."

So I suggest it be merged and we continue to test.

PR #5938, opened 2020-08-21
Infilling climatic element using multiple elements from different sources and Missing value plots from imputeTS package

shadrackkibet:
@rdstern this is ready for you to test. It is the implementation of the function from **chillR** package for infilling missing values using auxiliary data sources.


rdstern:
@shadrack ok.  Will do.

rdstern (changes requested):
1. I am happy that the control about whether to include output has been omitted.  Not important, but (at some stage) it would be good to reposition the other 2 controls.
2. Very good to have the saving variable in the dialogue.  But odd that it isn't the same control as there is, when a single variable is involved.  Then there is the control that includes the position button.
3.  I am not sure why it needs the date variable to be unique?  I first tried with a data frame having multiple stations and it gives an unclear error message.  When I made a new data frame with a single station, then it worked fine.  It is only comparing the 2 (or more) variables, so I am unclear why the date needs to be unique?
4. I need to check, but assume it won't work when a filter (say to a single station) is in operation.  This is a more general issue, and may need to be attacked at some stage - but not yet just for this dialogue.

shadrackkibet:
> Very good to have the saving variable in the dialogue. But odd that it isn't the same control as there is, when a single variable is involved. Then there is the control that includes the position button.

This is like what we have in start of rains, end of rains etc

rdstern:
@shadrackkibet but start and end of the rains writes to a new summary data frame, while here we are adding to the existing one.  And the method should be consistent between how we add when we use just the one variable, and whn there are others.

shadrackkibet:
Fixes  #5944. @rdstern this now includes missing value plots from `imputeTS` package. This is ready for you to test.

rdstern (changes requested):
@shadrackkibet this is great.  
a) It needs a check of the code when the Station is not filled, or when there is a single station, i.e. the wrapper isn't then needed.  Currently it doesn't then give a graph.  One example I used was the dodoma data which doesn't have a Station variable.
b) It doesn't transfer the graph into the R window.  I can understand that the combined graph doesn't work with Plotly, but it should transfer as an R graph, either from the toolbar or from the dialogue.  @Patowhiz is also looking at this feature in general.
c) In the 4th button there needs to be a 3rd single receiver, perhaps with the label True (Optional): or True Values (Optional):.  
d) The graphs are looking good enough that it will be worth trying to add further options (largely through the sub-dialogue) and also to have more flexibility to display (or not) the titles, so more of the screen can be for the graphs.

shadrackkibet:
@rdstern I had identified some of these issues. a) and b) should now work and at least you should be able to use plotly when you have one station eg Dodoma.

shadrackkibet:
Another improvement that is now working is for the **Fit multiple** i.e chillR should now work when you have multiple stations at a time. The station receiver is now enabled and should be working as expected.

rdstern:
With one station it is great that plotly works.  I hope, eventually, that the R graph display will become available for any graph.  Currently the toolbar icon and the dialogue don't seem to work.  It is available as can be seen by opting in the Tools > Options dialogue to plot there.  Here is an example:
image

I hope @Patowhiz is examining this in general.
Currently, if Station is selected, then it is a multiple graph, even if the data is filtered to a single station, which is something that could be a useful feature.  I wonder if this can be detected?

In the dialogue there is a lot of space at the top.  The top radio buttons and the controls could move up.  That's partly because there is another control that would be useful - and quite nice not to change the size for this.

So, this could be for the intervals option which can show Percents or Counts.  Maybe just a checkbox (default checked) Show Percents?

shadrackkibet:
> So, this could be for the intervals option which can show Percents or Counts. Maybe just a checkbox (default checked) Show Percents?

@rdstern I suggest a Combobox with the two options **Percent** and **Counts**. What do you think about that?

shadrackkibet:
> Currently, if Station is selected, then it is a multiple graph, even if the data is filtered to a single station, which is something that could be a useful feature. I wonder if this can be detected?

Do you mean it always gives you a combined plot with all stations?

rdstern:
RDS: Currently, if Station is selected, then it is a multiple graph, even if the data is filtered to a single station, which is something that could be a useful feature. I wonder if this can be detected?

SK: Do you mean it always gives you a combined plot with all stations?

RDS: No. It correctly applies the filter.  But it seems to go through your code for a combined graph, so the current system for plotly or R plot don't work.

shadrackkibet:
Aha! understood now. I can add a check to fix that.

shadrackkibet:
@rdstern ?

shadrackkibet:
@rdstern I have now found a way to fix the issue with R graph display for multiple plots. **View last graph** button should now work with the combined plot. `patchwork` package, an extension package to `ggplot2` provides an alternative to `gridExtra` function which work well with `gg` objects. Unfortunately, it doesn't work well with `plotly`, the author of the package is aware of it, according to him this is beyond the scope of the package as reported [here](https://github.com/thomasp85/patchwork/issues/70). I think we can find a workaround using `plotly::subplot()` which works with combining plots.

rdstern:
That's great.  I am less concerned with plotly with multiple plots than fixing the R display.  We usually want plotly to look into the details of a plot, so will often want to start with a single station.

shadrackkibet:
@rdstern are you happy with this?

lloyddewit:

Ivanluv:
@lloyddewit  the vb code looks fine from my end. I will need help looking at the R script

instat/dlgInfillMissingValues.vb
```diff
@@ -275,25 +435,25 @@ Public Class dlgInfillMissingValues
...
-                        ucrBase.clsRsyntax.AddToBeforeCodes(clsSetSeedFunction, 0)
-                    Case Else
-                        ucrBase.clsRsyntax.RemoveFromBeforeCodes(clsSetSeedFunction)
-                End Select
-            Else
-                ucrBase.clsRsyntax.RemoveFromBeforeCodes(clsSetSeedFunction)
-            End If
+    Private Sub ucrInputComboFunction_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputComboFunction.ControlValueChanged, ucrPnlMethods.ControlValueChanged, ucrChkSetSeed.ControlValueChanged, ucrPnlOptions.ControlValueChanged
+        If rdoFitSingle.Checked AndAlso rdoNaAggregate.Checked AndAlso ucrChkSetSeed.Checked Then
+            Select Case ucrInputComboFunction.GetValue()
```
lloyddewit, line 440: If there is only 2 cases then `If-Then-Else` is normally better

instat/dlgInfillMissingValues.vb
```diff
@@ -306,7 +466,75 @@ Public Class dlgInfillMissingValues
...
+                    Case "Distribution"
+                        ucrChkAddLegend.Visible = False
+                        clsVisualizeElementNa.AddParameter("type", Chr(34) & "distribution" & Chr(34), iPosition:=8)
+                    Case "Gap size"
+                        ucrChkAddLegend.Visible = True
+                        clsVisualizeElementNa.AddParameter("type", Chr(34) & "gapsize" & Chr(34), iPosition:=8)
+                    Case "Intervals"
+                        ucrChkAddLegend.Visible = True
+                        clsVisualizeElementNa.AddParameter("type", Chr(34) & "interval" & Chr(34), iPosition:=8)
+                End Select
```
lloyddewit, line 487: should we have a `Case Else` to raise a developer error if `GetText` returns something unexpected?

instat/dlgInfillMissingValues.vb
```diff
@@ -306,7 +466,75 @@ Public Class dlgInfillMissingValues
...
+    End Sub
+
+    Private Sub ucrReceiverDispMultShowStation_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverDispMultShowStation.ControlContentsChanged, ucrPnlOptions.ControlContentsChanged
+        If (rdoDisplay.Checked OrElse rdoShow.Checked) AndAlso Not ucrReceiverDispMultShowStation.IsEmpty Then
+            ucrNudFacetColumns.Visible = True
+            lblFacetColumns.Visible = True
+        Else
+            ucrNudFacetColumns.Visible = False
+            lblFacetColumns.Visible = False
+        End If
```
lloyddewit, line 534:
```suggestion
        ucrNudFacetColumns.Visible = (rdoDisplay.Checked OrElse rdoShow.Checked) AndAlso Not ucrReceiverDispMultShowStation.IsEmpty
        lblFacetColumns.Visible = ucrNudFacetColumns.Visible
```

lloyddewit (changes requested):
@shadrackkibet This looks good, I only had minor comments. 
@dannyparsons I reviewed the R code with my limited knowledge of R. You are also very welcome to review if you wish

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -3628,3 +3627,117 @@ DataSheet$set("public", "get_variable_sets", function(set_names, force_as_list)
...
+      for (i in seq_along(station_names)) {
+        temp_data <- curr_data[station_col == station_names[i], ]
+        plt_list[[i]] <- imputeTS::ggplot_na_imputations(x_with_na = temp_data[, element_col_name], x_with_imputations = temp_data[, element_col_name_imputed], x_axis_labels = temp_data[, x_axis_labels_col_name], title = station_names[i], xlab = xlab, ylab = ylab, legend = legend, x_with_truth = x_with_truth)
+      }
+    } else {
+      plt <- imputeTS::ggplot_na_imputations(x_with_na = element_col, x_with_imputations = element_imputed_col, x_axis_labels = curr_data[, x_axis_labels_col_name], xlab = xlab, ylab = ylab, legend = legend, x_with_truth = x_with_truth)
+    }
+  }
+  if (!missing(station_col_name) && dplyr::n_distinct(station_names) > 1) {
+    patchwork::wrap_plots(plt_list, ncol = ncol)
```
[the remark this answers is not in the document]
dannyparsons: It's not necessary and fractional less efficient apparently but I prefer to include an explicit `return` statement for clarity.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -3628,3 +3627,117 @@ DataSheet$set("public", "get_variable_sets", function(set_names, force_as_list)
...
 }
 )
+
+DataSheet$set("public", "patch_climate_element", function(date_col_name = "", var = "", vars = c(), max_mean_bias = NA, max_stdev_bias = NA, column_name, station_col_name) {
+  if (missing(date_col_name)) stop("date is missing with no default")
+  if (missing(var)) stop("var is missing with no default")
+  if (missing(vars)) stop("vars is missing with no default")
+  date_col <- self$get_columns_from_data(date_col_name, use_current_filter = FALSE)
+  if (!lubridate::is.Date(date_col)) stop("This column must be a date or time!")
+  curr_data <- self$get_data_frame(use_current_filter = FALSE)
```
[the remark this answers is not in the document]
shadrackkibet: `date_col` is also used in line 3636.

shadrackkibet:
@lloyddewit I have addressed points from your review above. This is ready for you to review again. Thanks

lloyddewit (dismissed):

lloyddewit:
@rdstern please could you approve? thanks

shadrackkibet:
@rdstern this is a reminder for you to approve?

rdstern:
@shadrackkibet this is looking very good.  I am testing with a new data set namely the 4 stations from Niger.
[niger4day.zip](https://github.com/africanmathsinitiative/R-Instat/files/5412147/niger4day.zip)

a) When going into the dialogue for the first time it gave the error below.  I press OK and it continues fine.
image

b) I then used the first option for a graph (for minimum temperatures) and it was fine.  When I returned to the dialogue it lost the label for the facets control.

c) I then went to Multiple and used tmax with tmin as the other variable:  This gave the error:
image

d) I unstacked the data by station to give minimum temperatures for each of the four stations.  I carried all the date variables.  Then I tried with Birni N'Konni with infilling from Niamey.  This gave:

image

shadrackkibet:
@rdstern I investigated the source of the bug reported in c) above. I think I understand why that happens. This is because some dates are missing in the data you provided above. I used infill missing dates which showed that 30 and 31 days are missing for **Niamey Aero** and **Birni N'Konni** respectively. After infilling it works well. I thought of adding this as a check and infill where necessary. 

d) For this I assume is a non-issue because this dialog is designed for stacked data i.e station in a single column, so unstacking is not needed.

rdstern:
@shadrackkibet you are quite right with the gaps in the data.  When I corrected that, then it worked.
The next thing I tried was to unstack the data from the 4 stations.  I unstacked tmin by station and carried all the date variables.  Then I sorted by date, so the data are in date order.

When I tried infilling for Agades, using just one station, Niamey, I get:

image

Perhaps this is because of the missing values at the start for Agades?

I have a feeling you are close here!

shadrackkibet:
@rdstern after infilling dates you don't have to unstack the data.

shadrackkibet:
So something like this should work.
image

rdstern:
@shadrackkibet I realise that I could have done as you suggested.  But I **_also_** wanted to see if an element could be filled-in using the same element from the other stations.  That should also be possible - no?  So infilling Agades tmin, using the values for the same element from Niamey, etc should also work I reckon?

shadrackkibet:
> @shadrackkibet you are quite right with the gaps in the data. When I corrected that, then it worked.
> The next thing I tried was to unstack the data from the 4 stations. I unstacked tmin by station and carried all the date variables. Then I sorted by date, so the data are in date order.
> 
> When I tried infilling for Agades, using just one station, Niamey, I get:
> 
> image
> 
> Perhaps this is because of the missing values at the start for Agades?
> 
> I have a feeling you are close here!

Aha! I now understand what you were trying to do. That should actually work! I tried unstacking **tmin** and carried date variables as you did. The only other thing I did is to check if there are missing dates in between. So I used our usual dialog "Fill date gaps" I found out that 30 rows for date gaps are missing. Once that is done it works.

shadrackkibet:
@rdstern I have three options to guide the user accordingly when there are missing dates. What do you think? 

1) Add a check for missing dates which will notify the user through the output window if there are missing dates. In that case, the user will then have to infill missing dates.
2) Have this as an option on the dialog as a checkbox - Fill missing dates.  When checked it infills missing dates. If unchecked a message is printed into the output window.
3) Infill missing dates internally and notify the user that missing dates has been infilled.

rdstern:
@shadrackkibet this now seems to work well.  I am a bit disappointed in myself that I didn't realise that the missing dates were causing the problem, particularly after you had told me exactly the same reason for the original data.

Anyway I would be happy with your first option, namely to detect and report, and then leave the user to infill the missing gaps.  That's usually done before the data are specified as climatic.  So, if - as I did - this is missed, then the user is reminded.

shadrackkibet:
@rdstern I have now added a check for the missing dates. Please have a look.

rdstern (approved):
@shadrack.  This is sensible now and I would be happy to see it merged and then used.  So I am approving.

It is an example that traps an error and now gives a good message.  We have a few other climatic dialogues that try to do similar checks and I wonder about putting the message in the output window instead of in the error message.  I like the fact if gives some details about the infilling when it works, so this would be a shorter message when it doesn't?  What do others including @dannyparsons think?

However I would like to also get the 2nd approval, by @lloyddewit soon so it can be merged

Patowhiz:
@rdstern I second your idea. 

If it's agreed that having such messages in the output window is the way to go.  Then I suggest such messages to be in a different colour.  To show the difference between normal output and user feedback.

shadrackkibet:
@rdstern I have now modified this to allow printing of the message into the output window. I also changed the output message when it works.

rdstern (approved):
@shadrackkibet I like the change and am approving the work at this stage.  It is now a dialogue that will benefit from being used and discussed in the next version.
The Niger data is odd in a variety of ways.  This time I infilled the data in the long form.  The command worked nicely.  I then unstacked tmax for the 4  stations and it first gave an error.  I found that was because the data were not in date order.  When put into date order it now said that there was a gap of -31 days.  Interesting that this was because there were duplicate days.  It could be good to check for duplicates explicitly at some stage, but I also wonder how that happened to the data.

instat/static/InstatObject/R/data_object_R6.R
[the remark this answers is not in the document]
lloyddewit: agreed, sorry I missed that

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -3628,3 +3627,134 @@ DataSheet$set("public", "get_variable_sets", function(set_names, force_as_list)
...
+  }
+  else {
+    out <- chillR::patch_daily_temperatures(weather = weather, patch_weather = patch_weather, vars = var, max_mean_bias = max_mean_bias, max_stdev_bias = max_stdev_bias)
+    col <- out[[1]][, var]
+  }
+  if(length(col)==dim(curr_data)[[1]]){
+  gaps_remaining <- summary_count_missing(col)
+  gaps_filled <- (summary_count_missing(curr_data[, var])-gaps_remaining)
+  cat(gaps_filled, " gaps filled", gaps_remaining, " remaining.", "\n")
+  self$add_columns_to_data(col_name = column_name, col_data = col)
```
lloyddewit, line 3695: indent?

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -3628,3 +3627,134 @@ DataSheet$set("public", "get_variable_sets", function(set_names, force_as_list)
...
+  }else{
+    if (gaps != 0) {
```
lloyddewit, line 3697: use `else if (gaps != 0) {`?

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1,2333 +1,2341 @@
-DataBook <- R6::R6Class("DataBook",
```
lloyddewit, line 1: I can't review because something is wrong with the diff, please could you redo your changes so that they are visible in the diff? thanks

lloyddewit (changes requested):
@shadrackkibet this looks good, a couple of very minor comments. Also, there's a problem with one of the diffs, thanks

lloyddewit (approved):
@shadrackkibet thanks for the quick changes - all good!

lloyddewit:
@rdstern there are no functional changes since your last approval, please could you re-approve? thanks

shadrackkibet:
@rdstern ?

rdstern (approved):

PR #5795, opened 2020-05-18
Handling missing values in One variable fit Model

Ivanluv:
Ive made some changes to the code from the pull request #5370.Fixes  #5323. @shadrackkibet  could you look at this

PR #5941, opened 2020-08-21
Adding stringdist package

### LINKED ISSUE #5940: Add "stringdist" package into R-Instat
When running `Prepare > Column: Text > Distance` the `stringdist` package isn't in R-Instat.

This should be added into the Rsetup.R file (most recent version is 0.9.6)

lilyclements:
Fixes #5940 

Added in the `stringdist` package used in dlgStringDistance.


instat/static/InstatObject/R/Rsetup.R
```diff
@@ -68,14 +68,15 @@ packs <- c("abind", "acepack", "agricolae",
...
            "SparseM", "spatial", "SPEI", "splus2R", 
-           "stringi", "stringr", "survival", "svglite", 
-           "sys", "systemfonts", "testthat", "tibble", 
-           "tidyr", "tidyselect", "translations", "treemapify", 
-           "units", "usethis", "utf8", "vctrs", 
-           "verification", "viridis", "viridisLite", "visreg", 
-           "wakefield", "weathermetrics", "whisker", "withr", 
-           "xfun", "XML", "xml2", "xopen", "xtable", 
-           "xts", "yaml", "zip", "zoo")
+           "stringi", "stringr", "stringdist", "survival",
```
shadrackkibet, line 71: I think packages are arranged alphabetically here so `stringdist` should be before `stringi`?

shadrackkibet (changes requested):

instat/static/InstatObject/R/Rsetup.R
lilyclements: Of course. Apparently I do not know the alphabet! Sorted now.

lloyddewit (approved):

lloyddewit:
@rdstern please could you approve? thanks

dannyparsons:
Did you edit this file directly? It should only be updated automatically by running through `install_packages.R` and generating the local repo again. When this is done many of the packages and versions change, hence we shouldn't edit this file directly.

dannyparsons:
We would usually do this just before a release for all the new packages so as not to require developers to download many packages again. However, now that we don't store packages in the repo we could do this more regularly.
@lloyddewit @lilyclements I could show you the process for doing this if you like.

lloyddewit:
@dannyparsons Thank you for spotting this! Yes, I would like to know the correct process. Can you explain this via text or is verbal better?

dannyparsons:
Basically the current process is that we need to create a mini package repository so that all the packages and their dependencies are all available without any further downloading. The `miniCRAN` R package does this for us which we use in the `install_packages.R` script. It can list and download all dependant packages based on a list of required packages and make a mini repository for this.

So to update our list:
1. (Optional) Update `r_version` if needed
2. In `install_packages.R` add the new package to the `pkgs` list
3. Rerun the following lines to make the new repo of packages
4. Run the last two lines starting with `enquote` which gives the lists of packages and versions.
5. Copy the output from these to replace the lists of packages and versions in `Rsetup.R`

So that's how to update the `Rsetup.R`.

The reason for not doing this regularly is that it downloads all R packages and potentially changes a lot of package versions as it pulls the latest of each package. Developers would need to update some of their R packages when changes to `Rsetup.R` are made.

However, note that this was the process when the actual local repository was included in the R-Instat installer and R packages were installed on the first load. Now, we don't actually need the repository of zipped package files, because we include a working version of R with all packages in the software. We probably just need a way to list all the packages required and their versions which we then use to install download the packages when creating the installer.

I'm not sure if that all makes sense?

I think we can think about how to improve this process for the next installer it will take a bit of time to investigate.

Have any other packages been added in this way? If so, we need to record them as they will be lost when we overwrite this lists. I suggest we have an issue which lists the new packages required so that they are added for the next release.

lloyddewit:
@dannyparsons please see my responses below.

> So that's how to update the `Rsetup.R`.

The instructions seem to make sense. I'd have to try it to be sure.

> 
> The reason for not doing this regularly is that it downloads all R packages and potentially changes a lot of package versions as it pulls the latest of each package. Developers would need to update some of their R packages when changes to `Rsetup.R` are made.
> 

If changes to `Rsetup.R` require developers to update their development environments, then I agree that we should not change `Rsetup.R` too regularly. The next releasable version may be a good opportunity. If we had a test event for the releasable version then we could also test that the updated packages have not caused any regression.

> However, note that this was the process when the actual local repository was included in the R-Instat installer and R packages were installed on the first load. Now, we don't actually need the repository of zipped package files, because we include a working version of R with all packages in the software. We probably just need a way to list all the packages required and their versions which we then use to install download the packages when creating the installer.

From the above, I understand that we currently deliver R and all the R packages required by R-Instat. Are you suggesting that we still deliver R, but only a list of required packages? These packages would then be downloaded by the R-Instat installer (or downloaded by R-Instat the first time it runs). Did I understand correctly?

> I think we can think about how to improve this process for the next installer it will take a bit of time to investigate.

yes it would be good to have an issue raised for this.

> Have any other packages been added in this way? If so, we need to record them as they will be lost when we overwrite this lists. I suggest we have an issue which lists the new packages required so that they are added for the next release.

On 29/05/20 PR #5795 updated `Rsetup.R` (looks like a manual update):

image

The 2 previous changes to `Rsetup.R` were on 06/02/20 and 03/12/19 (PR #5692 and PR #5607 ). These look like they were done with your involvement so I assume they're OK.

**Next steps:**
Maybe @lilyclements could regenerate `Rsetup.R` using the correct steps above and update this PR. We could then merge this PR when we are ready to create the next releasable version. 

In parallel, we could agree on the new approach for delivering and installing packages.

dannyparsons:
> From the above, I understand that we currently deliver R and all the R packages required by R-Instat. Are you suggesting that we still deliver R, but only a list of required packages? These packages would then be downloaded by the R-Instat installer (or downloaded by R-Instat the first time it runs). Did I understand correctly?

No sorry, I mean we download the packages when creating the R-Instat installer and put in R library folder of the installer (but we don't need to create a local repo for them as they are now not stored in the R-Instat repo). Then the `RSetup.R` script just checks against the packages and versions to make sure they are installed when running.

> On 29/05/20 PR #5795 updated Rsetup.R (looks like a manual update):

That manual change is actually ok, because this is just the list of packages being loaded when R-Instat starts. However, that does concern me for another reason because loading that package masks other functions and I think may give unexpected results for circular columns (will look into separately).

The next steps sounds fine, though the package versions change day by day so any pull request done now may need to be updated again nearer the time. But good for you or @lilyclements to try this anyway to see how it works.

dannyparsons:
The downside of only updating this at the last minute is that it doesn't go through proper testing. They may be bugs in newer versions of R packages for example. Maybe we could introduce a cut off period for introducing new R packages in a version? So we update the R packages scripts some time before a release and continue testing and making changes but don't allow any changes to R packages again before the release (except for bug fixes).

lilyclements:
@dannyparsons this is good to know, and seems to make a lot more sense than manually adding in packages.

Should I regenerate the `Rsetup.R` using the steps outlined above and update this PR, or should I close this PR?

dannyparsons:
I think you could close this for now but then follow the steps to update the file nearer the time we make the next release and include any new packages.

PR #5945, opened 2020-08-24
Changed Options text property to Plot Options in Cumulative/Exceedance graph and Two Variable graphs dialogs

### LINKED ISSUE #5899: Inconsistencies graphics dialogs - Plot Options or Options
@rdstern is there a reason why the button(Options) in Two variable Graph and Cumulative/Exceedance dialogs is different from that in other graphics dialogs under Describe>Specific, there it is called **Plot Options**. I claim these should be consistent because both buttons call the same sub dialog. I, therefore, suggest a change of the two-variable graph and Cumulative/Exceedance button to **Plot Options**. See below.

image
image

image

Wycklife:
Fixes #5899
@africanmathsinitiative/developers Ready for review

### DIFF
--- a/instat/dlgDescribeTwoVarGraph.resx
+++ b/instat/dlgDescribeTwoVarGraph.resx
@@ -126,14 +126,14 @@
     <value>10, 197</value>
   </data>
   <data name="cmdOptions.Size" type="System.Drawing.Size, System.Drawing">
-    <value>120, 23</value>
+    <value>119, 23</value>
   </data>
   <assembly alias="mscorlib" name="mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" />
   <data name="cmdOptions.TabIndex" type="System.Int32, mscorlib">
     <value>5</value>
   </data>
   <data name="cmdOptions.Text" xml:space="preserve">
-    <value>Options</value>
+    <value>Plot Options </value>
   </data>
   <data name="&gt;&gt;cmdOptions.Name" xml:space="preserve">
     <value>cmdOptions</value>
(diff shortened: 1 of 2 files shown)


lloyddewit (approved):

instat/dlgDescribeTwoVarGraph.resx
```diff
@@ -126,14 +126,14 @@
...
-    <value>120, 23</value>
+    <value>119, 23</value>
   </data>
   <assembly alias="mscorlib" name="mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" />
   <data name="cmdOptions.TabIndex" type="System.Int32, mscorlib">
     <value>5</value>
   </data>
   <data name="cmdOptions.Text" xml:space="preserve">
-    <value>Options</value>
+    <value>Plot Options </value>
```
lloyddewit, line 136: The extra space at the end is inconsistent with the other label. I'm not sure if it makes any difference. Was it intentional?
```suggestion
    <value>Plot Options</value>
```

lloyddewit (changes requested):

instat/dlgDescribeTwoVarGraph.resx
Wycklife: I don't think  it makes a difference but I can remove it.

lloyddewit (approved):

lloyddewit:
@Wycklife Thanks (I'm not sure if it makes a difference either), I committed the change.

@rdstern This PR just changes `Options` to `Plot Options` in two places. Please could you check it still looks OK and then approve? thanks

rdstern (approved):
Fine

PR #5957, opened 2020-09-02
Importing from Climsoft

### LINKED ISSUE #4915: Enhance the Climatic > File > CLIMSOFT dialogue

### LINKED ISSUE #6033: Unstacking data when importing from climsoft
Currently there is no option to automatically unstack data when importing from climsoft.

From discussions we had with @rdstern, a new option needs to be added that unstacks the data when the user wants to when importing from climsoft.

When this option is checked, imported data will automatically be unstacked and a new data frame produced.

Below is what has been proposed by @rdstern 

> We have a checkbox which is labelled Unstack (Tidy the imported data) Default is ticked. 
> If there is only one element (e.g. rainfall) then it is disabled.  - there is no unstacking to do. If multiple elements, then it is ok to use. 
> Now, what about the transfer of the flags.  I don't want this to take too much time, because the flag information in the current Climsoft isn't that great so far.  So, if you tick the flag information and there is just one element that's fine.  It transfers both - as now.
> If multiple elements, then I suggest the default is for unstacking to be unchecked.  Then it does as now.  If you check then it does as now and also includes a copy of the unstacked data (not flags). If everyting is used, then this could therefore be up to 4 data frames, i.e. element, station, stacked data (with flag(s)) and unstacked data.

Patowhiz:
Fixes #4915 and Fixes #6033

@africanmathsinitiative/developers this PR includes changes to the `instat_object_R6.R` file(still working on it). Hopefully no one  is also working on the file.


Patowhiz:
This PR introduces a new concept in the R-Instat. The use of wizards. By wizard, I mean taking a big complex task and splitting it up into a series of small simple steps aimed to completing the task. In this case, we have a climsoft main import dialog that has all the functionality of importing data from climsoft database, a user is expected to grasp all the controls shown in the main dialog and use them to achieve the task. Alternatively, we have a "wizard" dialog, that gives the same functionality but as a series of steps.

I chose the long path in implementing the wizard dialog, in the hope that if its widely accepted by the team, we can build a base code out of it and have it set as a standard of building other wizards especially in climatic tasks where I feel that we can take big tasks that are common in most Met services and lay them down nicely as a series of steps that a user can follow to get his/her desired outputs.

Please note, the current base buttons control(the OK, Reset, Help,To Script buttons) used was not conceptualised with "wizards" in mind, so I had to hide the comment text box and have back button just besides (A little bit of tweaking). But once we have a base code for "wizards", the base buttons control can be easily extended to support them.

For screenshots and more discussions explaining this concept check on the issue #4915.

This is now ready for review.

lloyddewit:

shadrackkibet:
First I comment on inconsistencies.

- Do we use CLIMSOFT, climsoft, climSoft or ClimSoft? I have seen all this in different places.
- Do we use `..`, `...` or no dots as all for buttons that open another form? A wizard opens a completely new form and closes the main form then perhaps that should be different? Not sure what though.
- The date pickers do not have the default dates. I am not sure what is sensible to here, perhaps that should be determined by the selected data from the dialog and what is in the database. Without changing dates the script showed empty `start_date=, end_date=)`. Same case when I use the wizard.
- The wizard looked like forms in climsoft in design. Controls are centered, not much strict alignment of controls. Is this deliberate because users of this dialog are likely to have used climsoft? therefore the wizard should just be familiar to what they have seen in climsoft? If that is the case then I would suggest alignment of controls to conform to what we have in R-Instat. See below marks on the screenshot just to point a few places that might need adjustments. I understand the base controls are different but I think that should not make much of a diffence with our ordinary forms if we need to conform!
image
- In the main form we have date pickers aligned vertically in the wizard they are aligned horizontally. I think we should be consistent.
image
image
- In the main dialog you have "Elements" while in the wizard you have " Selected Elements" also for "Stations" and "Selected stations"
image
image
image

- When I use the wizard the main form is closed. I wonder if it is sensible to return to the main dialog from the wizard without having to close it and then reopen the dialog i.e the reverse of what is done when wizard is opened.

shadrackkibet:
- I imported the data from a station and then used **close data book**. When I return back to the dailog I got this. That is because when this run, the database connection is lost although it still indicate connected from the dialog?
```
# Creating New Instat Object
data_book <- DataBook$new()
```
Going back to the dialog I got receivers still filled in but the selector is now empty without variables. I notice Ok is still enabled so I click it and got the error below.
image
Disconnecting and then connecting again fixes the problem!

- I think it is sensible to arrange stations names and element names in alphabetical order in the selector, this will make it easy to find a station/element from the selector. I imagine a situation where there are many stations and I know the name of the station that will simplify my work? Looked like station id's and element ids are arrange are arranged.

lloyddewit:
@Patowhiz I saw you'd made a significant commit after @shadrackkibet 's review. Does this address all his comments or do you expect to make additional commits? thanks

lloyddewit:
@Patowhiz Please could you clarify the status of theis PR? thanks

Patowhiz:
@shadrackkibet thank you for the suggestions. I have added some of the changes you suggested. A few points to note;

> The wizard looked like forms in climsoft in design. Controls are centred, not much strict alignment of controls. Is this deliberate because users of this dialog are likely to have used climsoft? therefore the wizard should just be familiar to what they have seen in climsoft? If that is the case then I would suggest an alignment of controls to conform to what we have in R-Instat. See below marks on the screenshot just to point a few places that might need adjustments. I understand the base controls are different but I think that should not make much of a difference with our ordinary forms if we need to conform!

The wizard has multiple panels stacked in the same dialog. The panels have different sizes because they have different controls that  have different sizes and in different locations. That's why I chose to centre the panels, the maximum boundaries of the dialog is however determined by the base buttons. 

> In the main form we have date pickers aligned vertically in the wizard they are aligned horizontally. I think we should be consistent.

> When I use the wizard the main form is closed. I wonder if it is sensible to return to the main dialog from the wizard without having to close it and then reopen the dialog i.e the reverse of what is done when the wizard is opened.

I chose to arrange the date pickers horizontally to reduce the overall dialog height. I can still arrange them vertically if you strongly think so, however I didn't see this as ideal visually, because the other steps will have a big gap between them and the base buttons. 
Treat the main and wizard dialogs as 2 separate dialogs that are used differently to achieve the same thing(that's the intention). I have added 2 separate menu items to reinforce this(as mentioned in the issue, the button was meant to temporarily showcase this functionality). The 2 dialogs are designed (both in code and UX) differently but to achieve the same thing.

> Is it possible to centre the password wizard into the form? Whenever I opened it always opens at the top of the software

The password sub-dialog is an R input prompt dialog. I'm not sure how if it can be centred. The command parameter available are; `getPass::getPass(msg = "PASSWORD:", noblank = FALSE, forcemask = FALSE)` 

> I think we always align lefts of the labels see below.

The original sub-dialog had these labels in the arrangement shown. I'm not sure if we have a similar design elsewhere in R-Instat. The consistent design that I'm sure of is, putting the labels on top of the textbox. I tried this but then the overall panel looked thinner visually(due to the overall size of the dialog, in the wizard). So I reverted to it's original design and used the same in the wizard . I'm happy to change the arrangements in any way that would improve the visual appeal and UX.

> Should I be allowed to do things like this? Then this sub dialog doesn't reset when I reset from the main dialog.

The port should always be a number, I've added the validation. The host is always a string. So any string input should be allowed e.g localhost.

> Why change Ok from the base control to Finish? I find that Ok still fits in. I managed to trick the wizard and it showed Ok disabled and enabled at some point.

This wizard dialog takes a user through a process, step by step. The "Finish" explicitly informs the user that he/she has finished all the steps. In addition, I used the word "finish" after looking into other software packages that have similar wizard concepts. Seems as a standard button text when it comes to wizards. I'm happy to change this to "Ok" if its agreed to be a big change in regards to R-Instat. 
I suspect the translation is overwriting the "Finish" to "Ok". I have corrected this. The enabling and disabling is dictated by validations done by the dialogs in each step.

> I think it is sensible to arrange stations names and element names in alphabetical order in the selector, this will make it easy to find a station/element from the selector. I imagine a situation where there are many stations and I know the name of the station that will simplify my work? Looked like station id's and element ids are are arranged.

This is a good suggestion. I initially thought the same. However, I think changing the order might technically confuse the user. A position of a particular station will keep on changing based on the order. I'm happy to add that functionality back. @rdstern what's your opinion on this.







.

mhabimana:
@Patowhiz @shadrackkibet: I am following your progress on this. Re-the right name to use,we have discussed these inconsistencies and agreed as a team to use  **Climsoft**in all our communication to be consistent. I hope this is clear now.

rdstern:
@shadrackkibet and @Patowhiz I like the discussions you have here.  I reply now on the final point of whether the variables should be in alphabetical order of the name when the names are given.  They are in numerical order when given that way.

I agree (so far) with the yes/no answer and am happy (for now) for them to remain as now, so they are in the same order numerically as they are with the names.  This is vaguely consistent with the current system for variable names that are in the order in the data frame.  We don't yet have a good system in general for dealing with lots of variables in the data selector.  So:

a) For now, what you can do is to select all the stations and then put them into an R-Instat data frame where both are available.  Then use this to check on the stations you want.
b) Make an issue to be able to sort names in the right-click on the data selector.  Here numerical order for variables is the order in the data frame, and alphabetical is again in alphabetical order of the names.  This will be useful for this dialogue and more generally.

Patowhiz:
@mhabimana you can now test this.

As @dannyparsons mentions in this [comment](https://github.com/africanmathsinitiative/R-Instat/issues/4915#issuecomment-705449106), the wizard dialog still needs broader discussions. If current design is not desirable it can be redone in another PR. 

Both the main and wizard dialog use the same R script code

.

mhabimana:
@Patowhiz -Thanks and noted.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1559,51 +1563,124 @@ DataBook$set("public", "database_disconnect", function() {
...
+	  }else{
+	    data_list <- list(db_station_info, db_observation_data)
+        names(data_list) = c("stations_info", "observation_data")
+	  }  
+  }else{
+       data_list <- list(db_station_info)
+       names(data_list) = "stations_info"
+  }
+  #import the data as separate data frames
   self$import_data(data_tables = data_list)
```
Patowhiz, line 1655: @dannyparsons running this `self$import_data(data_tables = data_list)` line will always make sure data frames with an existing name are renamed. I'm struggling with how I could get the renamed data frame to be used below. Basically if the user imports data from climsoft twice, the data frame names are automatically renamed in this function. Yet I need the same names for converting columns, linking the data frames etc.

The original code assumed that the user will only import once.

Patowhiz:
@lloyddewit @africanmathsinitiative/developers this is now ready for review.

lloyddewit:
@rdstern I think it's probably most efficient if you do your testing first. I will then do the final review, thanks

instat/static/InstatObject/R/instat_object_R6.R
dannyparsons: The method we have used in other places is to first ensure that the name will be unique before importing the data. Then you can be sure that the name will not be renamed. There is a helper function we created for this in `stand_alone_function.R`
e.g. `next_default_item("stations_info", self$get_data_names())`

lloyddewit:
@Patowhiz Do you need to address the comment from @dannyparsons before @rdstern reviews?

Patowhiz:
@lloyddewit yes. Let me address @dannyparsons comments first.

Patowhiz:
The only changes made in `instat_object_R6.R`  are in functions `database_connect` and `import_from_climsoft`.
For some reason Github is showing the other changes I think due to how I resolved the conflicts. I confirmed the code is the same in the master in those code lines indicated as changed.

@lloyddewit this is ready for review.

rdstern (commented):
The early problem I found has gone.  I will check at my end tomorrow with @Patowhiz and hope @dannyparsons can also check, so this can be merged for a video by the Manor House team.

instat/sdgImportFromClimSoft.vb
```diff
@@ -20,123 +20,160 @@ Imports RDotNet
...
+        clsRDatabaseConnect.AddParameter("user", frmMain.clsInstatOptions.strClimsoftUsername, iPosition:=3)
+
+        'set database disconnect R command
+        clsRDatabaseDisconnect.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$database_disconnect")
+
+        'disconnect if was already connected 
+        'If CheckIfConnectionIsActive() Then
+        '   Disconnect()
+        'End If
+        'bConnected = False
```
lloyddewit, line 85: This looks like it can be removed now

instat/sdgImportFromClimSoft.vb
```diff
@@ -20,123 +20,160 @@ Imports RDotNet
...
             lblConnection.ForeColor = Color.Green
-            cmdEnterPassword.Text = "Disconnnect"
+            btnConnect.Text = "Disconnect"
+
+            'disable all other controls. Entry allowed only when there is no existing connection
+            ucrComboBoxDatabaseName.Enabled = False
+            ucrComboBoxPort.Enabled = False
+            ucrTxtHost.Enabled = False
+            ucrTxtUserName.Enabled = False
+            chkRememberCredentials.Enabled = False
```
lloyddewit, line 108: These 4 lines are duplicated in the else part.
Replace with a small private function e.g. `Private Sub SetControlsEnabled(newEnabled As Boolean)`?

instat/sdgImportFromClimSoft.vb
```diff
@@ -20,123 +20,160 @@ Imports RDotNet
...
+    End Sub
 
-        If ucrInputPort.GetText() <> "" AndAlso ucrInputUserName.GetText() <> "" AndAlso ucrInputHost.GetText() <> "" AndAlso ucrInputDatabaseName.GetText() <> "" Then
-            cmdEnterPassword.Enabled = True
+    Private Sub btnConnect_Click(sender As Object, e As EventArgs) Handles btnConnect.Click
+        btnConnect.Enabled = False 'temporary disable 
+        'if was already connected, then user action is meant to disconnect else, try connecting to database
+        If bConnected Then
+            Disconnect()
+            bConnected = False
```
lloyddewit, line 153: Safer to set `bConnected = False` inside `Disconnect()`?

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1531,11 +1531,15 @@ DataBook$set("public", "has_database_connection", function() {
...
 DataBook$set("public", "database_connect", function(dbname, user, host, port, drv = RMySQL::MySQL()) {
+  #launches an input box prompt for entering password. 
+  #done this way so that password characters are not displayed in the output window
   password <- getPass::getPass(paste0(user, " password:"))
-  out <- NULL
-  out <- DBI::dbConnect(drv = drv, dbname = dbname, user = user, password = password, host = host, port = port)
-  if(!is.null(out)) {
-    self$set_database_connection(out)
+  if(length(password) > 0){
+    out <- NULL
```
lloyddewit, line 1538: Sorry for my lack of R knowledge, what's the purpose of this line?

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1558,53 +1562,128 @@ DataBook$set("public", "database_disconnect", function() {
...
+      station_ids_values <- paste0("(", paste0("'", as.character(unique(db_observation_data$station) ), "'", collapse = ", "), ")")
+      db_station_info <- DBI::dbGetQuery(con, paste0("SELECT * FROM station WHERE stationId IN ", station_ids_values, ";" ))
+    }
+    
+    station_data_name <- next_default_item("stations_info", self$get_data_names(), include_index = FALSE)
+    elements_data_name <- next_default_item("elements_info", self$get_data_names(), include_index = FALSE)
+    observation_data_name <- next_default_item("observation_data", self$get_data_names(), include_index = FALSE)
+    
+    #elements info could be optional
+    if (include_elements_info) {
```
lloyddewit, line 1644: can we guarantee that `db_station_info` has always been set when we reach this line?

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1558,53 +1562,128 @@ DataBook$set("public", "database_disconnect", function() {
...
     
-    data_list <- list(station_info, station_data)
-    names(data_list) = c("station_info","station_data")
-  }
-  else{
-    data_list <- list(station_info)
-    names(data_list) = "station_info"
+    flags_column_col_sql <- " "
+    if (include_observation_flags) {
+      db_flags_info <- DBI::dbGetQuery(con, "SELECT characterSymbol, description FROM flags")
```
lloyddewit, line 1607: Is this ever used?

lloyddewit (changes requested):
@Patowhiz your code looked very good, I only have small suggestions

instat/static/InstatObject/R/instat_object_R6.R
Patowhiz: @lloyddewit yes. this is meant to create the observation flags data frame if the option is included in the parameter.

instat/static/InstatObject/R/instat_object_R6.R
Patowhiz: @lloyddewit yes. If observation data flag is indicated, the it's guaranteed that the db_station_info will be set.

instat/static/InstatObject/R/instat_object_R6.R
Patowhiz: @lloyddewit I think it was previously used to initialise the variable before calling the database connection function. 
I have removed the line as this is no longer needed

instat/sdgImportFromClimSoft.vb
Patowhiz: @lloyddewit I agree. I have introduced a variable that removes repetition. And renamed the subroutine name.

instat/sdgImportFromClimSoft.vb
Patowhiz: @lloyddewit I agree it's safer.

instat/dlgClimSoft.vb
```diff
@@ -36,118 +44,204 @@ Public Class dlgClimSoft
...
-        If ucrReceiverMultipleStations.IsEmpty Then
-            ucrReceiverMultipleElements.strDatabaseQuery = "SELECT obselement.elementName FROM obselement,observationfinal WHERE obselement.elementId=observationfinal.describedBy GROUP BY observationfinal.describedBy;"
-        Else
-            ucrReceiverMultipleElements.Clear()
-            ucrReceiverMultipleElements.strDatabaseQuery = "SELECT obselement.elementName FROM obselement,observationfinal WHERE obselement.elementId=observationfinal.describedBy AND observationfinal.recordedFrom IN (" & String.Join(",", ucrReceiverMultipleStations.GetVariableNamesList(strQuotes:=Chr(39))) & ") GROUP BY observationfinal.describedBy;"
+        If bIgnoreReceiverChanges Then
+            Exit Sub
         End If
-        TestOKEnabled()
+        SetElementsRecieverQuery()
```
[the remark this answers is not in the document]
Patowhiz: Done.

instat/dlgClimSoft.vb
```diff
@@ -158,4 +252,40 @@ Public Class dlgClimSoft
         End If
     End Sub
 
+    Private Sub ucrUnstackDataControlsValueChanged_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkUnstackData.ControlValueChanged, ucrReceiverMultipleElements.ControlValueChanged, ucrChkObservationData.ControlValueChanged
```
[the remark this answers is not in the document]
Patowhiz: @lloyddewit I agree.

Patowhiz:
@lloyddewit I have made the changes that you have suggested above. Please have a look so that I can have this dialog ready for the videos that @rdstern is keen on. Thanks.

rdstern (approved):
I have been checking through interacting with @Patowhiz and it is unseful to merge as it is.  There is an addition that could be added as an additional issue.  It may be done in a different way to other dialogues, but is a potential general problem.
It is here the fact that there may be many stations to choose from.  In general with data frames having many variables this is the general issue that the data selector may have many entries.  We would like to restrict attention to a subset.  In SPSS you can define "sets" and then restrict attention to just those who are in a given set.  Here we have a qualifier and I would like to be able to choose from a given values of a qualifier.  But it could alternatively be from all the values in a district, etc.  So here we could choose from a value in another field in the database.  This could be by perhaps having additional options on the right-click in the data selector.  They could perhaps (in general) a value from a factor, or all names starting with xxx, etc.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1558,53 +1561,130 @@ DataBook$set("public", "database_disconnect", function() {
...
```
lloyddewit, line 1606: Sorry, I still don't fully understand, please could you show me where `db_flags_info` is used (i.e. on which line(s))?

lloyddewit (changes requested):
@Patowhiz Thanks for all the changes. I just have one open question remaining

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1369,7 +1369,7 @@ DataBook$set("public", "add_single_climdex_index", function(data_name, indices,
     linked_data_name <- self$get_linked_to_data_name(data_name, year)
     if(length(linked_data_name) == 0) {
       if(c("numeric","integer") %in% year_class) ind_data[[year]] <- as.numeric(levels(ind_data[[year]]))[ind_data[[year]]]
-      if("factor" %in% year_class) ind_data[[year]] <- make_factor(levels(ind_data[[year]]))[ind_data[[year]]]
+      if("factor" %in% year_class) ind_data[[year]] <- as.factor(levels(ind_data[[year]]))[ind_data[[year]]]
```
shadrackkibet, line 1372: Did you really want to make this change?

instat/static/InstatObject/R/instat_object_R6.R
Patowhiz: @lloyddewit I have remove the code line. Was meant to load in the flags data frame. But based on my latest changes, It's not needed

instat/sdgImportFromClimSoft.vb
```diff
@@ -20,123 +20,160 @@ Imports RDotNet
...
-        ucrInputUserName.SetParameter(New RParameter("user", 3))
-        cmdEnterPassword.Enabled = False
-        bControlsInitialised = True
+    Private Sub InitialiseDialog()
+        Dim dctDatabaseNames As New Dictionary(Of String, String)
+        Dim dctPorts As New Dictionary(Of String, String)
+
+        'database names
+        dctDatabaseNames.Add("mariadb_climsoft_db_v4", Chr(34) & "mariadb_climsoft_db_v4" & Chr(34))
+        dctDatabaseNames.Add("mariadb_climsoft_test_db_v4", Chr(34) & "mariadb_climsoft_test_db_v4" & Chr(34))
```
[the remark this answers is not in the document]
Patowhiz: @lloyddewit yes. These are the climsoft default database names. However, the user can still type in another name in the  combobox

instat/sdgImportFromClimSoft.vb
```diff
@@ -20,123 +20,160 @@ Imports RDotNet
...
+        ucrComboBoxDatabaseName.SetParameter(New RParameter("dbname", 0))
+        ucrComboBoxDatabaseName.SetItems(dctDatabaseNames)
+        ucrComboBoxDatabaseName.bAllowNonConditionValues = True
+
+        'host e.g 127.0.0.1
+        ucrTxtHost.SetParameter(New RParameter("host", 1))
+
+        'ports
+        dctPorts.Add("3308", "3308")
+        dctPorts.Add("3306", "3306")
```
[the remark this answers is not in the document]
Patowhiz: @lloyddewit yes. These are the climsoft default mariadb database port names. However, the user can still type in another port in the combobox.

In future the default database names and port numbers could come from a config file. I just didn't see that as part of this PR.

instat/sdgImportFromClimSoft.vb
```diff
@@ -20,123 +20,160 @@ Imports RDotNet
...
-    Private Sub ucrControlsContents_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrInputDatabaseName.ControlContentsChanged, ucrInputHost.ControlContentsChanged, ucrInputPort.ControlContentsChanged, ucrInputUserName.ControlContentsChanged
-        Dim strHost As String = ""
-        Dim strUser As String = ""
-        Dim strPort As String = ""
-        Dim strDatabaseName As String = ""
+    ''' <summary>
+    ''' checks if there is an active database connection
+    ''' </summary>
+    ''' <returns>true if connected to database, false if not</returns>
+    Public Function CheckIfConnectionIsActive() As Boolean
```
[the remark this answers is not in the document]
Patowhiz: I agree

instat/static/InstatObject/R/instat_object_R6.R
Patowhiz: @shadrackkibet thanks. The change was not intentional. Reverted it.

lloyddewit (approved):
@shadrackkibet Thanks for the extra review
@Patowhiz I reviewed the new changes and it all looks good - thank you!

lloyddewit:
@rdstern Thanks for approving this PR on 18 Dec, since then only a few minor changes have been made. Please could you approve again?
@Patowhiz the only change since 18 Dec that could potentially affect the functionality was that `as.factor` was changed to `make_factor`. If you think its worth Roger doing extra tests on this change then let us know? thanks

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -2109,7 +2188,7 @@ DataBook$set("public","tidy_climatic_data", function(x, format, stack_cols, day,
   if(!continue) return()
   
   # Standard format of slowest varying structure variables first (station then element then date) followed by measurements
-  if(!missing(station)) z <- data.frame(station = make_factor(y$station), date = y$date)
+  if(!missing(station)) z <- data.frame(station = forcats::as_factor(y$station), date = y$date)
```
shadrackkibet, line 2191: Still, I will use `make_factor()` here not unless there is a reason to use `forcats::as_factor()` just to make sure we are consistent in how we deal with factors. We had issues with the conversion of factors before which was solved using `make_factor()` which might not be the case here.

lloyddewit:
@Patowhiz If you decide to change the code based on @shadrackkibet's suggestion then please could you also take the opportunity to rename `IsConnectionIsActive` to `IsConnectionActive`? (I noticed this after I'd approved, thanks)

@rdstern Danny's currently on leave. Is there an urgency to merge this PR (for the videos), or should we wait for Danny to return and review this PR?

rdstern:
I was about to comment on our recent use of forcats, but @shadrackkibet has already covered that point.
As you say @dannyparsons is on leave now.  I have mixed views on whether we should wait or take advantage of his absence!  I guess we should wait.

lloyddewit:
@dannyparsons ready for your review!

rdstern (approved):
I had a comment earleir, that should become an issue.  But this version can be merged.

lloyddewit:
@dannyparsons Please could you review (you asked to review this before it was merged)?, thanks

dannyparsons (approved):

dannyparsons:
David and I have reviewed this. We suggest this can be merged, particularly for the changes to the main dialog. We like the idea of wizards as a potential option for some dialogs. This is a good starting point on a single dialog. However, the implementation of this in R-Instat needs to be reconsidered in order for it to be usable across dialogs e.g. new child users controls for wizard dialogs. Since this does not affect any core control or classes then we agree to merge this now and re-implementation can be done separately.

PR #5958, opened 2020-09-02
Download gridded data from IRI as Netcdf format

### LINKED ISSUE #3995: Gridded data and Netcdf
This dialogue **Climatic > File > Import Gridded Data (IRI)** seems to work - but at closer inspection it has serious problems.
(See the next comment before trying to correct them.)
 
1. I assume it needs an internet connection and we should perhaps mention that?
2. The up-down controls should be updated to our new version, so it is easier to type into them.  And the up-down steps should be 0.005 and not 0.0001.  (I assume this should be quite easy to do.)
3. It has its own waiting dialogue, though that isn't very useful.  Perhaps it should be replaced by information on how many rows of data will be imported, so we can be cautious if there are too many.  But this would be for later.
4. The date column is imported as a factor.  I presume could be a date column.  It can be made into a date column, though the format (%d<space>%b<space>%Y) could usefully be added to our standard ones in the Make Date dialogue.
5. Now the big one.  It only imports part of the data.  When I imported one point, it was from 1981 to 1987.  With an area (500 points) it was less than 2 years.  I don't know if this was to do with my internet connection and would like to try with a better one.  I don't know whether the IRI site allows a choice of first and last date?  

It does seem to import sensibly into 2 data frames.  I would like to use it in the forthcoming training in Guyana.

shadrackkibet:
Fixes #3995. @dannyparsons @rdstern I am putting up this for you to have a look. Danny, I am getting below bugs especially with **CHIRPS** when downloading data. Sometimes it works, when I change the time range. Perhaps it is an internet issue? I tried copying the link from VS output window into the browser, sometimes it downloads the file but then when I try to open the file using Netcdf dialog I get an error as shown below. I am not sure why. 

**Download from IRI**
image

**Import Netcdf**
image

### DIFF
--- a/instat/dlgImportGriddedData.vb
+++ b/instat/dlgImportGriddedData.vb
@@ -226,14 +226,11 @@ Public Class dlgImportGriddedData
     End Sub
 
     Private Sub cmdBrowse_Click(sender As Object, e As EventArgs) Handles cmdBrowse.Click
-        Dim strPath As String
-
-        Using dlgFolder As New FolderBrowserDialog
-            dlgFolder.Description = "Choose Folder"
-            If dlgFolder.ShowDialog() = DialogResult.OK Then
-                ucrInputFilePath.SetName("")
-                strPath = dlgFolder.SelectedPath
-                ucrInputFilePath.SetName(Replace(strPath, "\", "/"))
+        Using dlgSave As New SaveFileDialog
+            dlgSave.Title = "Import from IRI Data Library Dialog"
+            dlgSave.Filter = "NetCDF files|*.nc"
+            If dlgSave.ShowDialog() = DialogResult.OK Then
+                ucrInputFilePath.SetName(dlgSave.FileName.Replace("\", "/"))
             End If
         End Using
     End Sub
--- a/instat/dlgImportGriddedData.vb
+++ b/instat/dlgImportGriddedData.vb
@@ -155,6 +155,8 @@ Public Class dlgImportGriddedData
                 dctFiles = New Dictionary(Of String, String)
                 dctFiles.Add("RFE2 Daily Precipitation", Chr(34) & "daily_rfev2_est_prcp" & Chr(34))
                 dctFiles.Add("RFE2 10-day Precipitation", Chr(34) & "10day_rfev2_est_prcp" & Chr(34))
+                dctFiles.Add("Daily estimated Precipitation", Chr(34) & "daily_est_prcp" & Chr(34))
+                dctFiles.Add("Monthly estimated Precipitation", Chr(34) & "monthly_est_prcp" & Chr(34))
                 ucrInputData.SetItems(dctFiles)
                 ucrInputData.cboInput.SelectedItem = "RFE2 Daily Precipitation"
                 ucrDtpMinDate.MinDate = New Date(2000, 10, 31)
--- a/instat/dlgImportGriddedData.vb
+++ b/instat/dlgImportGriddedData.vb
@@ -40,20 +40,19 @@ Public Class dlgImportGriddedData
         ucrBase.iHelpTopicID = 526
 
         ucrInputSource.SetParameter(New RParameter("source", 0))
-        dctDownloadPairs.Add("NOAA_RFE2", Chr(34) & "NOAA_RFE2" & Chr(34))
-        dctDownloadPairs.Add("CHIRPS_V2P0", Chr(34) & "CHIRPS_V2P0" & Chr(34))
+        dctDownloadPairs.Add("NASA", Chr(34) & "NASA_TRMM_3B42" & Chr(34))
+        dctDownloadPairs.Add("NOAA", Chr(34) & "NOAA_RFE2" & Chr(34))
         dctDownloadPairs.Add("TAMSAT", Chr(34) & "TAMSAT" & Chr(34))
-        dctDownloadPairs.Add("NOAA_ARC2", Chr(34) & "NOAA_ARC2" & Chr(34))
-        dctDownloadPairs.Add("NOAA_CMORPH_DAILY", Chr(34) & "NOAA_CMORPH_DAILY" & Chr(34))
-        dctDownloadPairs.Add("NOAA_CMORPH_3HOURLY", Chr(34) & "NOAA_CMORPH_3HOURLY" & Chr(34))
-        dctDownloadPairs.Add("NOAA_CMORPH_DAILY_CALCULATED", Chr(34) & "NOAA_CMORPH_DAILY_CALCULATED" & Chr(34))
-        dctDownloadPairs.Add("NASA_TRMM_3B42", Chr(34) & "NASA_TRMM_3B42" & Chr(34))
+        dctDownloadPairs.Add("UCSB CHIRPS", Chr(34) & "CHIRPS_V2P0" & Chr(34))
         ucrInputSource.SetItems(dctDownloadPairs)
         ucrInputSource.SetDropDownStyleAsNonEditable()
 
         ucrInputData.SetParameter(New RParameter("data", 1))
-        dctFiles.Add("Daily_Est._Prcp.", Chr(34) & "daily_estimated_prcp" & Chr(34))
+        dctFiles = New Dictionary(Of String, String)
+        dctFiles.Add("TRMM 3B42 3-Hourly Precipitation", Chr(34) & "3_hourly_estimated_prcp" & Chr(34))
+        dctFiles.Add("TRMM 3B42 Daily Precipitation", Chr(34) & "daily_estimated_prcp" & Chr(34))
         ucrInputData.SetItems(dctFiles)
+        ucrInputData.cboInput.SelectedItem = "TRMM 3B42 3-Hourly Precipitation"
         ucrInputData.SetDropDownStyleAsNonEditable()
 
         ucrInputFilePath.SetParameter(New RParameter("path", 2))
@@ -118,8 +117,8 @@ Public Class dlgImportGriddedData
         rdoEntireRange.Checked = True
 
         clsDownloadFromIRIFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$download_from_IRI")
-        clsDownloadFromIRIFunction.AddParameter("source", Chr(34) & "NOAA_RFE2" & Chr(34), iPosition:=0)
-        clsDownloadFromIRIFunction.AddParameter("data", Chr(34) & "daily_estimated_prcp" & Chr(34), iPosition:=1)
+        clsDownloadFromIRIFunction.AddParameter("source", Chr(34) & "NASA_TRMM_3B42" & Chr(34), iPosition:=0)
+        clsDownloadFromIRIFunction.AddParameter("data", Chr(34) & "3_hourly_estimated_prcp" & Chr(34), iPosition:=1)
         clsDownloadFromIRIFunction.AddParameter("path", (Environment.GetFolderPath(Environment.SpecialFolder.UserProfile) & "\Downloads").Replace("\", "/"), iPosition:=2)
         'opted to set this as default location since it has data for all sources
         clsDownloadFromIRIFunction.AddParameter("min_lon", 10, iPosition:=3)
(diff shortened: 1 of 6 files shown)


rdstern:
One option did work at your defaults.  But I then got similar error messages to the above, i.e.
image

Then I re-tried the same options as worked before and got something like the above error again.

I am keen on being able to import the different sources of rainfall data.  But I assume other elements are also available.  At this stage I would not object to a long list as it seems to be an excellent source of data.

dannyparsons:
I would try testing in RStudio so that you figure out which parts of the R code are giving the error. And also check that the URL is being constructed correctly each time by printing it and trying to use the URL directly.
That may be linked to the second issue of opening the file.

dannyparsons:
I'm struggling to get any NetCDF files to download from the IRI site directly. Can you try this as well? I tried CHIRPS and TAMSAT it just gives files of size 0.

shadrackkibet:
@rdstern @dannyparsons this is now working better. I found out that fixing the `method = "libcurl" or "curl"` in `download.file()` works better rather than `auto`.

rdstern:
I tried with roughly the defaults and TAMSAT data.  After quite a long time - 5 minutes it said:
image

In addition:

a) There is the option to save the file.  Is this instead (or as well) as putting into the data frame?  Would that be an rds file or an nc file? It could be that we wish to download a large file (say for an area) and then only open the subset that corresponds to the stations?  Then we might want to save it, but not put it in R-Instat.  If not, then maybe that could be added and called called Transfer and be additional to save?
b) I am given no prior information on the size of the file I am downloading.  Is this possible.  I think it is worked out incorrectly by IRI?  It would be like a Try feature?
c) We have the problem in R-Instat that this operation might take a long time and we can't interrupt a command.  Could a transfer be "in the background" so we could continue with other work?

dannyparsons:
> @rdstern @dannyparsons this is now working better. I found out that fixing the method = "libcurl" or "curl" in download.file() works better rather than auto.

I preferred the "auto" method because it gives a pop up download box. If it doesn't work then no problem, but could you check it definitely doesn't work this way? There was an issue with the IRI site so it might work now.

dannyparsons:
> a) There is the option to save the file. Is this instead (or as well) as putting into the data frame? Would that be an rds file or an nc file? It could be that we wish to download a large file (say for an area) and then only open the subset that corresponds to the stations? Then we might want to save it, but not put it in R-Instat. If not, then maybe that could be added and called called Transfer and be additional to save?

The saved file is the downloaded NetCDF file. This could be clarified on the dialog if it's not clear. The dialog should give the option to download without importing exactly for the reason you mention. Importing would import everything.

> b) I am given no prior information on the size of the file I am downloading. Is this possible. I think it is worked out incorrectly by IRI? It would be like a Try feature?

If we list the spatial and temporal resolution of each data source hard coded in the dialog code then we can estimate quite accurately the number of rows to be imported. We could then try to estimate how this relates to file size by comparing sizes with number of observations, but this is really not simple. It may be different for different data sources and will also depend on each dimension e.g. a long time series for one point may be bigger/smaller than the same number of observations for a single time point. We would really be guessing as this can only be known from IRI and it doesn't seem to provide it for subsets well.

> c) We have the problem in R-Instat that this operation might take a long time and we can't interrupt a command. Could a transfer be "in the background" so we could continue with other work?

It could be but I don't think I would recommend this as it needs a lot of processing power to run two versions of R in separate threads. It would need a lot of testing and checking to see how stable this is and I guess it's not important enough for that. I hope we can have the download progress bar as that is very useful.

dannyparsons:
- It allows me to have the min date after the max date. I think we can prevent this on the dialog.

shadrackkibet:
@dannyparsons can you have a look into this again? I have now made the changes you requested above.

instat/ucrDateTimePicker.vb
```diff
@@ -185,4 +185,12 @@ Public Class ucrDateTimePicker
     Public Function DateValue() As Date
         Return dtpDateTime.Value
     End Function
+
+    Public Function CurrentDate(Optional bWithQuotes As Boolean = True) As String
```
dannyparsons, line 189: Could you rename this to `ValueAsString` to be consistent with `ValueAsRDate` and also add a parameter for the separator which can be `"/"` by default?

dannyparsons: Actually, you shouldn't need this function because the control can return a Date function automatically.

instat/dlgImportGriddedData.vb
```diff
@@ -31,114 +33,122 @@ Public Class dlgImportGriddedData
...
+        ucrChkWholeRange.AddParameterPresentCondition(True, {"min_date", "max_date"})
+        ucrChkWholeRange.AddParameterPresentCondition(False, {"min_date", "max_date"}, False)
+
+        clsDefaultStartDate = New RFunction
+        clsDefaultStartDate.SetRCommand("as.Date")
+        clsDefaultStartDate.AddParameter("x", Chr(34) & "2000/10/31" & Chr(34))
+
+        clsDefaultEndDate = New RFunction
+        clsDefaultEndDate.SetRCommand("as.Date")
+        clsDefaultEndDate.AddParameter("x", ucrDtpMaxDate.CurrentDate())
```
dannyparsons, line 119: The date picker control is set up to return an `as.Date()` function you can pass in directly so you don't need create these R functions in the dialog.

dannyparsons:
The min date is correctly limited to the minimum date of the data, but the max date doesn't have the same lower limit. Can you add that too?

Also, the pop up about the max date being higher than the min date comes up a lot of times as you're changing the dates. Could you set this so that it only appears once and then doesn't appear again until the dialog is reset?

dannyparsons:
Thanks, that seems to address those issues. I'm making a comment on the issue for design changes.

lloyddewit:
@shadrackkibet thank you for the changes. Is this PR now ready for review or do you still expect more commits? thanks

shadrackkibet:
I still have more TODO before this is ready for re-review. Thanks

shadrackkibet:
@dannyparsons I have now incorporated the changes as outlined in the issue, please have a look. I left out CMORPH from the dialog but the code is there, we will enable once we agree on which options to include. 

I am unable to download TAMSAT v3 data, even directly from the browser using URLs. e.g [http://iridl.ldeo.columbia.edu/SOURCES/.Reading/.Meteorology/.TAMSAT/.TARCAT/.v3p1/.daily/.rfe/X/%2810E%29/%2810.5E%29/RANGEEDGES/Y/%2810N%29/%2810.5N%29/RANGEEDGES/data.nc](http://iridl.ldeo.columbia.edu/SOURCES/.Reading/.Meteorology/.TAMSAT/.TARCAT/.v3p1/.daily/.rfe/X/%2810E%29/%2810.5E%29/RANGEEDGES/Y/%2810N%29/%2810.5N%29/RANGEEDGES/data.nc). It is taking long to load with no success. Could you also try? Not sure if this is an issue from IRI servers. There seem to be issues accessing data from time to time from IRI servers. Yesterday I was unable to download CMORPH data, on enquiry via email they said it was an issue from their end which seems to have been fixed now.

dannyparsons:
Great thanks I'll check.

The TAMSAT data also didn't download for me. I also tried restricting it to a few days and it still doesn't work. Text seems to work ok but not NetCDF. Can you send them another query about this? It's generally been stable for me in the past so I hope this is just a temporary instability.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -2088,62 +2088,56 @@ DataBook$set("public","package_check", function(package) {
...
+    else stop("Data file does not exist for TAMSAT data")
   #monthly,climatology and TAMSAT RFE 0p1 are yet to be implemented.
   }else if(source == "NOAA_ARC2") {prexyaddress <- paste0(init_URL, ".NOAA/.NCEP/.CPC/.FEWS/.Africa/.DAILY/.ARC2")
-  if(data == "daily_estimated_prcp") {extension <- ".daily/.est_prcp"}#(days since 1960-01-01 12:00:00) ordered (1 Jan 1983) to (12 Sep 2020)
-  else if(data == "monthly_average_estimated_prcp") {extension <- ".monthly/.est_prcp"}#(months since 1960-01-01) ordered (Jan 1983) to (Aug 2020)
-  else stop("Data file does not exist for NOAA ARC2 data")
+   if(data == "daily_estimated_prcp") {extension <- ".daily/.est_prcp"}#(days since 1960-01-01 12:00:00) ordered (1 Jan 1983) to (12 Sep 2020)
+    else if(data == "monthly_average_estimated_prcp") {extension <- ".monthly/.est_prcp"}#(months since 1960-01-01) ordered (Jan 1983) to (Aug 2020)
+    else stop("Data file does not exist for NOAA ARC2 data")
   }else if(source=="NOAA_RFE2") {prexyaddress <- paste0(init_URL, ".NOAA/.NCEP/.CPC/.FEWS/.Africa")
```
dannyparsons, line 2113: The sources should now match the dialog so this should just be "NOAA" and it should just add the NOAA part of the URL, then the rest is specific to the choice so it should go in the data part.
The same applies to the other sources.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -2088,62 +2088,56 @@ DataBook$set("public","package_check", function(package) {
...
-  if(data == "daily_estimated_prcp"){extension <- ".DAILY/.RFEv2/.est_prcp"}#(days since 2000-10-31 12:00:00) ordered (31 Oct 2000) to (12 Sep 2020)
-  else stop("Data file does not exist for NOAA RFE2 data")
+      if(data == "daily_estimated_prcp"){extension <- ".DAILY/.RFEv2/.est_prcp"}#(days since 2000-10-31 12:00:00) ordered (31 Oct 2000) to (12 Sep 2020)
```
dannyparsons, line 2114: The data names should also be updated to include the data set e.g. `data == "rfe2_daily_estimated_prcp"`

instat/dlgImportGriddedData.vb
```diff
@@ -31,114 +33,116 @@ Public Class dlgImportGriddedData
...
 
     Private Sub SetDefaults()
-        clsRDefaultFunction = New RFunction
+        clsDownloadFromIRIFunction = New RFunction
+        bShowMessageBox = True
 
-        'ucrInputExportFile.IsReadOnly = True
-        'ucrInputExportFile.SetName("")
+        'temp fix
+        rdoEntireRange.Checked = True
```
dannyparsons, line 117: This should be easy to have conditions based on whether the data parameters are present? The default is the whole range so the date parameters should not be included by default.

instat/dlgImportGriddedData.vb
```diff
@@ -31,114 +33,116 @@ Public Class dlgImportGriddedData
...
-        clsRDefaultFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$import_from_iri")
-        clsRDefaultFunction.AddParameter("download_from", Chr(34) & "CHIRPS_V2P0" & Chr(34))
-        clsRDefaultFunction.AddParameter("data_file", Chr(34) & "daily_0p05" & Chr(34))
-        clsRDefaultFunction.AddParameter("data_frame_name", "IRI_Library_Data")
-        clsRDefaultFunction.AddParameter("location_data_name", "Lat_Lon_Data")
-        clsRDefaultFunction.AddParameter("path", Chr(34) & System.IO.Path.Combine(System.IO.Path.GetTempPath() & "R_Instat_Temp_IRI_Download").Replace("\", "/") & Chr(34))
+        clsDownloadFromIRIFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$download_from_IRI")
+        clsDownloadFromIRIFunction.AddParameter("source", Chr(34) & "NASA_TRMM_3B42" & Chr(34), iPosition:=0)
+        clsDownloadFromIRIFunction.AddParameter("data", Chr(34) & "3_hourly_estimated_prcp" & Chr(34), iPosition:=1)
+        clsDownloadFromIRIFunction.AddParameter("path", (Environment.GetFolderPath(Environment.SpecialFolder.UserProfile) & "\Downloads").Replace("\", "/"), iPosition:=2)
```
dannyparsons, line 122:
```suggestion
        clsDownloadFromIRIFunction.AddParameter("path", Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments).Replace("\", "/"), iPosition:=2)
```

I see that Downloads isn't part of Special Folders so in case of any issues on some machines with this workaround let's use Documents instead which shouldn't ever cause problems.

instat/dlgImportGriddedData.vb
```diff
@@ -31,114 +33,116 @@ Public Class dlgImportGriddedData
...
         'opted to set this as default location since it has data for all sources
-        clsRDefaultFunction.AddParameter("X1", 10)
-        clsRDefaultFunction.AddParameter("X2", 10.5)
-        clsRDefaultFunction.AddParameter("Y1", 10)
-        clsRDefaultFunction.AddParameter("Y2", 10.5)
-        ucrBase.clsRsyntax.SetBaseRFunction(clsRDefaultFunction)
+        clsDownloadFromIRIFunction.AddParameter("min_lon", 10, iPosition:=3)
+        clsDownloadFromIRIFunction.AddParameter("max_lon", 10.5, iPosition:=4)
+        clsDownloadFromIRIFunction.AddParameter("min_lat", 10, iPosition:=5)
+        clsDownloadFromIRIFunction.AddParameter("max_lat", 10.5, iPosition:=6)
```
dannyparsons, line 127: The default for these doesn't really matter as long as it's somewhere in Africa that is in most of the sources, but we though a nice default would be the Maseno station: lon = 34.58, lat = 0. Then add 0.5 to each for max.

dannyparsons: And if it's easy, can we have a single point as the default?

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -2086,4 +2086,62 @@ DataBook$set("public","package_check", function(package) {
...
+    else if(data == "daily_recovered_rainfall_estimate ") {extension <- ".daily/.rfe_filled/"}#grid: /T (julian_day) ordered (1 Jan 1983) to (10 Sep 2020) by 1.0 N= 13768 pts :grid 
+    else if(data == "dekadal_rainfall_estimates") {extension <- ".dekadal/.rfe/"}#grid: /T (days since 1960-01-01) ordered [ (1-10 Jan 1983) (11-20 Jan 1983) (21-31 Jan 1983) ... (1-10 Sep 2020)] N= 1357 pts :grid
+    else if(data == "dekadal_recovered_rainfall_estimate") {extension <- ".dekadal/.rfe_filled/"}#grid: /T (days since 1960-01-01) ordered [ (1-10 Jan 1983) (11-20 Jan 1983) (21-31 Jan 1983) ... (1-10 Sep 2020)] N= 1357 pts :grid
+    else if(data == "monthly_rainfall_estimates") {extension <- ".monthly/.rfe/"}#grid: /T (months since 1960-01-01) ordered (Jan 1983) to (Aug 2020) by 1.0 N= 452 pts :grid
+    else if(data == "monthly_recovered_rainfall_estimate") {extension <- ".monthly/.rfe_filled/"}#grid: /T (months since 1960-01-01) ordered (Jan 1983) to (Aug 2020) by 1.0 N= 452 pts :grid
+    else stop("Data file does not exist for TAMSAT data")
```
dannyparsons, line 2107: I got an error here with this code: 
`data_book$download_from_IRI(source="TAMSAT", data="daily_recovered_rainfall_estimate", path="C:/Users/Danny/Documents", min_lon=34.58, min_lat=0, min_date=as.Date(x="2000/10/31"), max_date=as.Date(x="2000/11/1"), name="TAMSAT_v3.1 Daily Rainfall Complete (Filled)_31_10_2000_01_11_2000", download_type="point")`

I think it's missing this data option from the list.

dannyparsons (commented):
The design looks much better and it's working well. A few comments below, and two other points:

- the new data frame name is quite long. Can this instead be just the source name (in lower case)? And if easy in R we add a label attribute to the data frame which gives full details e.g. "TAMSAT v3.1 Daily Rainfall 31/10/2000 - 1/11/2000 Point: (34,58, 0)"
- If I check "Don't import after downloading" then the data frame name box should disappear

instat/dlgImportGriddedData.vb
shadrackkibet: @dannyparsons I tried removing `min_date` and `max_date` from the default with conditions set as shown below

```
ucrPnlDateRange.AddParameterPresentCondition(rdoCustomRange, {"min_date", "max_date"})
ucrPnlDateRange.AddParameterPresentCondition(rdoArea, {"min_date", "max_date"}, False)
```
On testing, no radio button is checked. What could be the problem?

I also tried adding `min_date` by default with the following conditions

```
ucrPnlDateRange.AddParameterPresentCondition(rdoCustomRange, "min_date")
ucrPnlDateRange.AddParameterPresentCondition(rdoArea, "min_date", False)
```
this worked as expected i.e `rdoCustomRange` is checked by default.

dannyparsons: Not sure why that should be if they are both added by default. Still, the condition on `min_date` is still valid.

shadrackkibet: Fixed now. Thanks

shadrackkibet:
@dannyparsons could you have a look into this again?

shadrackkibet:
@dannyparsons I am trying to add TAMSAT v3.0 but seems restricting of values and ranges doesn't wok for v3.0. How did you get this to work? For example, what could be the problem with this link? [https://iridl.ldeo.columbia.edu/SOURCES/.Reading/.Meteorology/.TAMSAT/.TARCAT/.v3p0/.daily/.rfe/X/(34.58E)/VALUES/Y/(0N)/VALUES/](https://iridl.ldeo.columbia.edu/SOURCES/.Reading/.Meteorology/.TAMSAT/.TARCAT/.v3p0/.daily/.rfe/X/(34.58E)/VALUES/Y/(0N)/VALUES/)

dannyparsons:
If you go to the data source https://iridl.ldeo.columbia.edu/SOURCES/.Reading/.Meteorology/.TAMSAT/.TARCAT/.v3p0/.daily/.rfe and then use the website to restrict the range you see that it uses lon and lat instead of X and Y e.g. https://iridl.ldeo.columbia.edu/SOURCES/.Reading/.Meteorology/.TAMSAT/.TARCAT/.v3p0/.daily/.rfe/lon/%2819.0125W%29%2840E%29RANGEEDGES/lat/%284S%29%2835.9625S%29RANGEEDGES/
So we need to be able to have different dimension names when restricting the range. I wasn't actually aware it wasn't always X and Y.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -2086,4 +2086,138 @@ DataBook$set("public","package_check", function(package) {
...
+      extension <- ".monthly/.rfe/"
+    } # grid: /T (months since 1960-01-01) ordered (Jan 1983) to (Aug 2020) by 1.0 N= 452 pts :grid
+    else if (data == "monthly_rfe_filled") {
+      extension <- ".monthly/.rfe_filled/"
+    } # grid: /T (months since 1960-01-01) ordered (Jan 1983) to (Aug 2020) by 1.0 N= 452 pts :grid
+    else {
+      stop("Data file does not exist for TAMSAT data")
+    }
+  } else if (source == "NOAA_ARC2") {
+    prexyaddress <- paste0(init_URL, ".NOAA/.NCEP/.CPC/.FEWS/.Africa/.DAILY/.ARC2")
```
dannyparsons, line 2134: Can you change this so that the source is just "NOAA". And then include "ARC2" in the data part? We may have other NOAA products to add in future so the base URL will be different.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -2086,4 +2086,138 @@ DataBook$set("public","package_check", function(package) {
...
+    if (data == "daily_est_prcp") {
+      extension <- ".daily/.est_prcp"
+    } # (days since 1960-01-01 12:00:00) ordered (1 Jan 1983) to (12 Sep 2020)
+    else if (data == "monthly_est_prcp") {
+      extension <- ".monthly/.est_prcp"
+    } # (months since 1960-01-01) ordered (Jan 1983) to (Aug 2020)
+    else {
+      stop("Data file does not exist for NOAA ARC2 data")
+    }
+  } else if (source == "NOAA") {
```
dannyparsons, line 2144: So this combines with "NOAA_ARC2" above with "NOAA" as the source and ARC2 Daily, RFE2 daily etc. being data options for that source.

instat/dlgImportGriddedData.vb
```diff
@@ -148,110 +141,91 @@ Public Class dlgImportGriddedData
...
                 dctFiles = New Dictionary(Of String, String)
-                dctFiles.Add("Mean Microwave Only Est. Prcp.", Chr(34) & "mean_microwave_only_est_prcp" & Chr(34))
-                dctFiles.Add("Mean Morphed Est. Prcp.", Chr(34) & "mean_morphed_est_prcp" & Chr(34))
-                dctFiles.Add("Orignames Mean Microwave Only Est. Prcp.", Chr(34) & "orignames_mean_microwave_only_est_prcp" & Chr(34))
-                dctFiles.Add("Orignames Mean Morphed Est. Prcp.", Chr(34) & "orignames_mean_morphed_est_prcp" & Chr(34))
-                dctFiles.Add("Renamed102015 Mean Microwave Only Est. Prcp.", Chr(34) & "renamed102015_mean_microwave_only_est_prcp" & Chr(34))
-                dctFiles.Add("Renamed102015 Mean Morphed Est. Prcp.", Chr(34) & "renamed102015_mean_morphed_est_prcp" & Chr(34))
-                ucrInputDataFile.SetItems(dctFiles)
-                ucrInputDataFile.cboInput.SelectedItem = "Mean Morphed Est. Prcp."
+                dctFiles.Add("Daily Precipitation 0.25 degree", Chr(34) & "daily_improved_global_0p25_prcp" & Chr(34))
```
dannyparsons, line 176: Can you move this one to the bottom of the list? I think it will be least common to want the low resolution version.

instat/dlgImportGriddedData.vb
```diff
@@ -148,110 +141,93 @@ Public Class dlgImportGriddedData
...
                 dctFiles = New Dictionary(Of String, String)
-                dctFiles.Add("Daily Est. Prcp.", Chr(34) & "daily_estimated_prcp" & Chr(34))
-                dctFiles.Add("Monthly Average Est. Prcp.", Chr(34) & "monthly_average_estimated_prcp" & Chr(34))
-                'monthly,climatology and TAMSAT RFE 0p1 are yet to be implemented.
-                ucrInputDataFile.SetItems(dctFiles)
-                ucrInputDataFile.cboInput.SelectedItem = "Daily Est. Prcp."
-            ElseIf ucrInputDownloadFrom.cboInput.SelectedItem = "NOAA_RFE2" Then
+                dctFiles.Add("RFE2 Daily Precipitation", Chr(34) & "daily_rfev2_est_prcp" & Chr(34))
+                dctFiles.Add("RFE2 10-day Precipitation", Chr(34) & "10day_rfev2_est_prcp" & Chr(34))
+                dctFiles.Add("Daily estimated Precipitation", Chr(34) & "daily_est_prcp" & Chr(34))
```
dannyparsons, line 158:
```suggestion
                dctFiles.Add("ARC2 Daily Precipitation", Chr(34) & "daily_est_prcp" & Chr(34))

```

instat/dlgImportGriddedData.vb
```diff
@@ -148,110 +141,93 @@ Public Class dlgImportGriddedData
...
+                dctFiles.Add("Monthly estimated Precipitation", Chr(34) & "monthly_est_prcp" & Chr(34))
```
dannyparsons, line 159:
```suggestion
                dctFiles.Add("ARC2 Monthly Precipitation", Chr(34) & "monthly_est_prcp" & Chr(34))

```

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -1283,6 +1283,40 @@ summary_sample <- function(x, size, replace = FALSE){
   else{sample(x = x, size = size, replace = replace)}
 }
 
+add_xy_area_range <- function(path, min_lon, max_lon, min_lat, max_lat) {
```
dannyparsons, line 1286: Add two parameters here, `dim_x = "X"` and `dim_y = "Y"` so that we can change the dimension names to "lon" and "lat" for TAMSAT case.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -1283,6 +1283,40 @@ summary_sample <- function(x, size, replace = FALSE){
...
+         "(", ifelse(min_lon < 0, paste0(abs(min_lon), "W"), paste0(min_lon, "E")), ")", "/",
+         "(", ifelse(max_lon < 0, paste0(abs(max_lon), "W"), paste0(max_lon, "E")), ")", "/",
+         "RANGEEDGES", "/",
+         "Y", "/",
+         "(", ifelse(min_lat < 0, paste0(abs(min_lat), "S"), paste0(min_lat, "N")), ")", "/",
+         "(", ifelse(max_lat < 0, paste0(abs(max_lat), "S"), paste0(max_lat, "N")), ")", "/",
+         "RANGEEDGES", "/")
+}
+
+add_xy_point_range <- function(path, min_lon, min_lat) {
```
dannyparsons, line 1297: Same here as above.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -2086,4 +2086,133 @@ DataBook$set("public","package_check", function(package) {
     }
   }
 }
-)
\ No newline at end of file
+)
+
+DataBook$set("public", "download_from_IRI", function(source, data, path = tempdir(), min_lon, max_lon, min_lat, max_lat, min_date, max_date, name, download_type = "Point", import = TRUE) {
+  init_URL <- "https://iridl.ldeo.columbia.edu/SOURCES/"
```
dannyparsons, line 2092: Define `dim_x = "X"` and `dim_y = "Y"` as variables here. Then in the TAMSAT section set `dim_x = "lon"`, `dim_y = "lat"` so that correct names are used.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -2086,4 +2086,133 @@ DataBook$set("public","package_check", function(package) {
...
+    }
+  } else {
+    stop("Source not specified correctly.")
+  }
+  prexyaddress <- paste(prexyaddress, extension, sep = "/")
+  if (download_type == "Area") {
+    URL <- add_xy_area_range(path = prexyaddress, min_lon = min_lon, min_lat = min_lat, max_lon = max_lon, max_lat = max_lat)
+  }
+  else if (download_type == "Point") {
+    URL <- add_xy_point_range(path = prexyaddress, min_lon = min_lon, min_lat = min_lat)
```
dannyparsons, line 2200: add the new `dim_x` and `dim_y` defined parameters here to ensure correct url.

instat/dlgImportGriddedData.vb
```diff
@@ -148,110 +141,93 @@ Public Class dlgImportGriddedData
...
+                ucrInputData.cboInput.SelectedItem = "RFE2 Daily Precipitation"
+                ucrDtpMinDate.MinDate = New Date(2000, 10, 31)
+                ucrDtpMaxDate.MinDate = New Date(2000, 10, 31)
+            Case "TAMSAT"
                 dctFiles = New Dictionary(Of String, String)
-                dctFiles.Add("Daily Est. Prcp.", Chr(34) & "daily_estimated_prcp" & Chr(34))
-                ucrInputDataFile.SetItems(dctFiles)
-                ucrInputDataFile.cboInput.SelectedItem = "Daily Est. Prcp."
-            ElseIf (ucrInputDownloadFrom.cboInput.SelectedItem = "NOAA_CMORPH_DAILY" OrElse ucrInputDownloadFrom.cboInput.SelectedItem = "NOAA_CMORPH_3HOURLY" OrElse ucrInputDownloadFrom.cboInput.SelectedItem = "NOAA_CMORPH_DAILY_CALCULATED") Then
+                dctFiles.Add("v3.1 Daily Rainfall", Chr(34) & "daily_rfe" & Chr(34))
```
dannyparsons, line 166: Can you add v3.0 options here as well? Put them at the top for now since there is an issue with v3.1.

dannyparsons (commented):
This looks good now, just the TAMSAT v3.0 options need to be added and the dimension names issue to be resolved. I have commented below on how to approach these.

shadrackkibet:
@dannyparsons TAMSAT v3.0 is now added. Please check.

shadrackkibet:
@dannyparsons @rdstern this is a reminder for you to check.

rdstern (commented):
@shadrack this is impressive.
a) The only minor issue I could find was with the ARC2 data, where the minimum date should be 1/1/1983, and it is currently the same are the (correct) RFE data at October 2000.
b) At the top there is correctly) the name of the dialogue.  Should (internet connection needed) be added?

These are just comments:
c) I don't have a good way to do this, but I has to wait about 25 minutes for a single point to be copied - for the whole time period.  
d) The place it is saved reminds me that I wonder if we should soon have an issue to be able to set a working directory?  This might even be an item that would be on the file menu and also perhaps on our new front menu.

shadrackkibet:
@rdstern a) and b) done!

instat/dlgImportGriddedData.vb
```diff
@@ -31,114 +32,107 @@ Public Class dlgImportGriddedData
...
-        ucrNudMaxLat.SetMinMax(-50, 50)
-        ucrNudMaxLat.DecimalPlaces = 4
-        ucrNudMaxLat.Increment = 0.0001
-        ucrNudMaxLat.SetLinkedDisplayControl(lblMaxLat)
+        ucrInputSource.SetParameter(New RParameter("source", 0))
+        dctDownloadPairs.Add("NASA", Chr(34) & "NASA" & Chr(34))
+        dctDownloadPairs.Add("NOAA", Chr(34) & "NOAA" & Chr(34))
+        dctDownloadPairs.Add("TAMSAT_v3.0", Chr(34) & "TAMSAT_v3.0" & Chr(34))
+        dctDownloadPairs.Add("TAMSAT_v3.1", Chr(34) & "TAMSAT_v3.1" & Chr(34))
+        dctDownloadPairs.Add("UCSB_CHIRPS", Chr(34) & "UCSB_CHIRPS" & Chr(34))
```
dannyparsons, line 46: Can you remove the `_`  from the display name? That will make them more readable.

dannyparsons (changes requested):
Just one minor change.

instat/dlgImportGriddedData.vb
shadrackkibet: Done!

dannyparsons (dismissed):
Great.

lloyddewit:
@rdstern I understand from the comments above that your concerns have all been addressed. Please could you re-approve?

@dannyparsons If I see that you have already approved a PR then I will assume that my final review/approval is no longer needed. If I misunderstood then please let me know, thanks

rdstern (approved):
@shadrackkibet this is great.  I assume @lloyddewit to give the final review, unless already done

dannyparsons (approved):
I take it from Stephen's comment that I can approve and merge.

PR #5970, opened 2020-09-08
Added menu items under Structured>Circular

### LINKED ISSUE #5964: The circular menu
We need to check that no one else is editing frm main, but this should be quick - I hope.
There is already Structured > Circular, with the Define dialogue.
a) Add `Calculator `below Define (This should have the ordinary Calculator from the Prepare Menu.  Ideally when opened it has the Circular keyboard automatically.
b) Add `Summaries` below that.  Ideally when you open the sub-dialogue it starts initially with the Circular tab.
c) Then a horizontal line
d) Then **_Scatter Plot_**.  This can have a blank dialogue for now.  It will implement the ScatterPlot function from the Circular package.
e) Then Density Plot.  Same as d) from Circular package
f) Then (possibly) Rose Plot.  I need to check so leave this for now.
g) Then a horizontal line.
h) Then Wind Rose with the current Wind Rose dialogue
i) Then Wind/Pollution Rose with a similar dialogue from the Openair package.
j) Then (these next are disabled for now) Percentile Rose, Frequency Rose, Polar Plot, Polar Cluster, Polar Annulus, Polar Frequencies
k) Then a horizontal line
l) Then (also disabled) Circlize.

HawardKetoyoMsatsi:
Fixes #5964

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -25,6 +25,7 @@ Public Class dlgCalculator
     Public bFirstLoad As Boolean = True
     Public iHelpCalcID As Integer
     Private iBasicWidth As Integer
+    Public strDefaultItem As String
 
     Private Sub dlgCalculator_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         autoTranslate(Me)
@@ -53,7 +54,7 @@ Public Class dlgCalculator
 
     Private Sub SetDefaults()
         ucrCalc.ucrSaveResultInto.SetPrefix("Calc")
-        ucrCalc.ucrInputCalOptions.SetName("Basic")
+        ucrCalc.ucrInputCalOptions.SetName(strDefaultItem)
         ucrCalc.Reset()
         ucrCalc.chkShowParameters.Checked = False
         ucrCalc.chkSaveResultInto.Checked = True
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -2202,4 +2202,12 @@ Public Class frmMain
     Private Sub mnuClimaticCompareTimeSeriesPlot_Click(sender As Object, e As EventArgs) Handles mnuClimaticCompareTimeSeriesPlot.Click
         dlgTimeSeriesPlot.ShowDialog()
     End Sub
+
+    Private Sub mnuClimaticCompareSeasonalPlot_Click(sender As Object, e As EventArgs) Handles mnuClimaticCompareSeasonalPlot.Click
+        dlgSeasonalPlot.ShowDialog()
+    End Sub
+
+    Private Sub mnuClimaticCompareConditionalQuantiles_Click(sender As Object, e As EventArgs) Handles mnuClimaticCompareConditionalQuantiles.Click
+        dlgConditionalQuantilePlot.ShowDialog()
+    End Sub
 End Class
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -559,6 +559,7 @@ Public Class frmMain
     End Sub
 
     Private Sub mnuPrepareRechapeColumnSummaries_Click(sender As Object, e As EventArgs) Handles mnuPrepareColumnReshapeColumnSummaries.Click
+        sdgSummaries.iIndex = 0
         dlgColumnStats.ShowDialog()
     End Sub
 
@@ -1101,6 +1102,7 @@ Public Class frmMain
     End Sub
 
     Private Sub mnuDescribeGeneralColumnSummaries_Click(sender As Object, e As EventArgs) Handles mnuDescribeGeneralColumnSummaries.Click
+        sdgSummaries.iIndex = 0
         dlgColumnStats.ShowDialog()
     End Sub
 
@@ -1492,6 +1494,7 @@ Public Class frmMain
             dlgColumnStats.strDefaultVariables = Nothing
             dlgColumnStats.strDefaultFactors = Nothing
         End If
+        sdgSummaries.iIndex = 0
         dlgColumnStats.ShowDialog()
     End Sub
 
@@ -1644,6 +1647,7 @@ Public Class frmMain
             dlgColumnStats.strDefaultVariables = Nothing
             dlgColumnStats.strDefaultFactors = Nothing
         End If
+        sdgSummaries.iIndex = 0
         dlgColumnStats.ShowDialog()
     End Sub
 
@@ -1670,6 +1674,7 @@ Public Class frmMain
             dlgColumnStats.strDefaultVariables = Nothing
             dlgColumnStats.strDefaultFactors = Nothing
         End If
+        sdgSummaries.iIndex = 0
         dlgColumnStats.ShowDialog()
     End Sub
 
@@ -2002,6 +2007,7 @@ Public Class frmMain
     End Sub
 
     Private Sub mnuPrepareCalculateCalculations_Click(sender As Object, e As EventArgs) Handles mnuPrepareCalculator.Click
+        dlgCalculator.strDefaultItem = "Basic"
         dlgCalculator.ShowDialog()
     End Sub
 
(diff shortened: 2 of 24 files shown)


shadrackkibet:
I think we can continue further with this work by adding new empty forms  for the proposed dialogs into this PR.  With help from others, we can support @HawardKetoyoMsatsi to implement these dialogs.

HawardKetoyoMsatsi:
@Wycklife and I have added empty dialogs for the Rose plots and others mentioned on #5974 . @africanmathsinitiative/developers  Someone to review these changes and the merging should be quick so that we can begin work on these dialogs. Thank you.

HawardKetoyoMsatsi:
@rdstern  we have decided to have all the six additional Rose plots in one dialog named Other Rose Plots, is this okay?

lloyddewit:
@africanmathsinitiative/developers For speed, I think we can bypass your review for this PR. If @rdstern approves, then I can review/approve/merge.

HawardKetoyoMsatsi:
@dannyparsons @Wycklife have added a time series dialog and a sub dialog with controls. Kindly get the changes from this branch. @africanmathsinitiative/developers This also fixes #5973 .

dannyparsons:
There are probably too many changes here for one pull request, but I am ok with the time series dialog added here for speed of getting that done.

@HawardKetoyoMsatsi there are two small changes in `dlgCalculator.vb` and `sdgSummaries.vb` which seem unrelated to the issue. If these are intentional could you explain them?

HawardKetoyoMsatsi:
> There are probably too many changes here for one pull request, but I am ok with the time series dialog added here for speed of getting that done.
> 
> @HawardKetoyoMsatsi there are two small changes in `dlgCalculator.vb` and `sdgSummaries.vb` which seem unrelated to the issue. If these are intentional could you explain them?

@dannyparsons These changes are related to this issue. 
The Structured>Circular>Calculator menu opens dlgCalculator with which the default calculator should be Circular Calculator. The Structured>Circular>Summaries>Summaries opens sdgSummaries  which needs open with the default being Circular in order to give column statistics for circular data...

lloyddewit:
@HawardKetoyoMsatsi Thenak you for the recent changes. Do you plan any more commits to this PR or can we go ahead with review and merge? thanks

HawardKetoyoMsatsi:
@lloyddewit I don't have any changes. You can go ahead with review and merge. 
Thanks

instat/sdgSummaries.vb
```diff
@@ -26,7 +26,7 @@ Public Class sdgSummaries
     Private strWeightLabel As String
     Public bEnable2VariableTab As Boolean = True
     Public bOkEnabled As Boolean = True
-
+    Public iIndex As Integer
```
lloyddewit, line 29: Give this a default value of zero?

instat/frmMain.vb
```diff
@@ -559,6 +559,7 @@ Public Class frmMain
     End Sub
 
     Private Sub mnuPrepareRechapeColumnSummaries_Click(sender As Object, e As EventArgs) Handles mnuPrepareColumnReshapeColumnSummaries.Click
+        sdgSummaries.iIndex = 0
```
lloyddewit, line 562: Not needed if iIndex has default value of zero?
This comment applies to all the similar cases below

lloyddewit (changes requested):
@HawardKetoyoMsatsi Thank you for the changes, I just had a couple of comments, thanks

instat/frmMain.vb
dannyparsons: I think if the dialog was opened previously then it may now not be 0 so would need to be reset. However, I don't think this is the best way to implement this. I will explain in the general comment.

instat/dlgCalculator.vb
```diff
@@ -25,6 +25,7 @@ Public Class dlgCalculator
     Public bFirstLoad As Boolean = True
     Public iHelpCalcID As Integer
     Private iBasicWidth As Integer
+    Public strDefaultItem As String
```
dannyparsons, line 28: I like this approach to add a variable to control which tab to show by default. 
However, good practice would be to make this `Private` and then add a method to set this such as `SetDefaultItem()` which can be called to set this variable.
The reason for this is that a method will allow us to make checks on what this variable is set as. For example, what happens if someone sets this to something that doesn't exist? The method allows us to have a consistent way of setting the variables and ensure errors and trapped. The method can be simple for now with just a line to set the variable and the checks can be added later.
Check with @shadrackkibet if any of this is not clear or get back to me.

instat/sdgSummaries.vb
dannyparsons: Can you change this to be a character instead of an integer? The reason is that it will be easy to make a mistake and set the wrong number whereas a character is clearer, and also when we add new tabs we potentially change the order of the tabs which would mean the tab numbers will be wrong whereas we can associate the names to a tab number and change them in one place if needed.

Can you do a similar thing to the suggestion for the calculator dialog? i.e. make this `Private` and have a method to set? You need to be careful about when this is then applied. It is not only when resetting, it could be that the tab needs to be changed after the first time it's opened because it's opened from another place. I suggest we use `""` to indicate not to change the tab (unless it's being reset then it goes back to first time). There are a few different cases here so think them out carefully.

instat/frmMain.vb
```diff
@@ -2002,6 +2007,7 @@ Public Class frmMain
     End Sub
 
     Private Sub mnuPrepareCalculateCalculations_Click(sender As Object, e As EventArgs) Handles mnuPrepareCalculator.Click
+        dlgCalculator.strDefaultItem = "Basic"
```
dannyparsons, line 2010: So my comment above means this line will be replaced by a call to a method e.g. `dlgCalculator.SetDefaultItem("Basic")`

instat/frmMain.vb
```diff
@@ -1101,6 +1102,7 @@ Public Class frmMain
     End Sub
 
     Private Sub mnuDescribeGeneralColumnSummaries_Click(sender As Object, e As EventArgs) Handles mnuDescribeGeneralColumnSummaries.Click
+        sdgSummaries.iIndex = 0
```
dannyparsons, line 1105: We can avoid having to set this back to the default in every other place by setting the variable back to the default in the sub dialog after it's been applied.

instat/sdgSummaries.vb
dannyparsons: Also, I think there should be an equivalent variable in `dlgColumnSummaries` to be able to set this. We should not be setting something on the sub dialog directly without going through the main dialog, because the main dialog may also have functionality which conflicts. For example, if we had a method in the main dialog to also set the default tab, then it would overwrite an external change.

The right place to be setting a default for the sub dialog is when the main dialog sets up the sub dialog. So we should add a way of setting this through the main dialog.

dannyparsons (commented):
Sorry I have made what seems to be trivial changes more complicated but there are some important software development principles that we want to stick to even for small changes. I have tried to explain the ideas in the comments below but get back to me if anything's not clear.

dannyparsons:
@HawardKetoyoMsatsi now that you have the new Seasonality plot dialog to do would you like me to finish off the changes I suggested to this above if you haven't already done this? Then you can concentrate on the new dialog.

HawardKetoyoMsatsi:
Sure, you finish off the changes let me continue with the Seasonality plot dialog.

dannyparsons:
Ok I'll work on this now. We need to be careful about conflicts, because the seasonality dialog is added in this pull request. So you need to wait for this to be merged before making a new pull request for the seasonality dialog, but you could start work on the dialog and copy over changes once this is merged. Check with @shadrackkibet about that.

HawardKetoyoMsatsi:
I am working on it on a new branch, once the merging is done I will copy my work to update the seasonality dialog accordingly.

instat/dlgCalculator.vb
dannyparsons: @HawardKetoyoMsatsi if you're interested in how I implemented this see the changes here https://github.com/africanmathsinitiative/R-Instat/pull/5983/commits/caef5d6ef573440da9b6a1a23dda393c81cd0688

lloyddewit:
@HawardKetoyoMsatsi I understand that PR #5983 includes all your updates (with some adjustments from Danny). So I think it's OK to close this PR. Please let me know if I misunderstood, thanks

dannyparsons:
That's fine to close but because my new PR is branched from here it would automatically close when the new PR is merged. The advantage of leaving this to be done automatically is that any linked issues will get closed at the correct time when the new PR is merged.

lloyddewit:
@dannyparsons yes, I forgot that GitHub would detect that the commits are merged and close this PR automatically, that's a better way of doing it.

PR #5971, opened 2020-09-09
Fixed small issues in the General Graphics dialogue

### LINKED ISSUE #5552: Problems with the Describe > General > Graphics dialogue
I give an example to illustrate the problems.  
I use the library Dodoma data.  I have 4 layers (though the problems may be the same with just 2 levels.
I do first a jitter plot with tmax as the Y and month_abbr as the X.
The second layer is a boxplot.  Same x and y.

I then Edit and make the **_width of the jitter 0.2_** and the **_alpha into 0.1_**.
Then I Edit the boxplot and make the outlier colour into blue
Then edit the jitter again and you see that the width of the jitter is still 0.2, but the alpha argument has become unset.
Interesting!

I don't know how general this problem is.

Three, more problems.
1. In the boxplot colour the first option is Null.  I don't know what this does? I did try it. I would like the option to not display the outliers and that is NA, rather than Null.  I suspect that is what Null is trying to do, so should we change it to NA?
2. When you start the general graphics dialogue the only option is Add and that is correct.  But if you have the 2 layers above and press Reset, then it still leaves the Edit and Delete buttons enabled.  This gives a nasty error if you try them, which (of course) I did! 
3) The dialogue itself is called **General**.  It should be called **Graphics** instead.

Ivanluv:
@africanmathsinitiative/developers this fixes #5552.Its still in progress


instat/ucrGeom.vb
```diff
@@ -1651,9 +1651,11 @@ Public Class ucrGeom
             If iNewGeomIndex <> -1 Then
                 clsCurrGeom = lstAllGeoms(iNewGeomIndex)
                 If iNewGeomIndex <> iCurrentGeomIndex Then
-                    For Each clsParam In lstAllGeoms(iCurrentGeomIndex).clsLayerParameters
-                        clsGeomFunction.RemoveParameterByName(clsParam.strLayerParameterName)
-                    Next
+                    'Why do we have to remove the parameters here?
+
+                    'For Each clsParam In lstAllGeoms(iCurrentGeomIndex).clsLayerParameters
```
Ivanluv, line 1656: @Patowhiz  I am looking at issue #5552 and I think the code I have commented causes the issue raised ie 

> I then Edit and make the width of the jitter 0.2 and the alpha into 0.1.
Then I Edit the boxplot and make the outlier colour into blue
Then edit the jitter again and you see that the width of the jitter is still 0.2, but the alpha argument has become unset.
Interesting!

I  however don't understanding why and how  the code  is of use.  Could you kindly take a look at this and help me understand

lloyddewit:
@Ivanluv please could you clarify the status of this PR? do you still need help from @Patowhiz ? thanks

Ivanluv:
@lloyddewit I got help but am still working on this

Ivanluv:

lloyddewit:
@shadrackkibet Please could you review? thanks

instat/ucrAdditionalLayers.vb
```diff
@@ -117,6 +119,11 @@ Public Class ucrAdditionalLayers
     End Sub
 
     Private Sub SetEditDeleteEnabled()
+        If lstLayers.Items.Count = 0 Then
+            'Disabling of the Edit and Delete buttons when the list of Layers is cleared
+            cmdDelete.Enabled = False
+            cmdEdit.Enabled = False
+        End If
```
lloyddewit, line 126: The selected items will always be a subset of the total items. Therefore, do we need this extra check?
If we do need the extra check then would it be more readable to make this function:
```
    Private Sub SetEditDeleteEnabled()
        If lstLayers.Items.Count > 0 AndAlso lstLayers.SelectedItems.Count = 1 Then
            cmdDelete.Enabled = True
            cmdEdit.Enabled = True
        Else
            cmdDelete.Enabled = False
            cmdEdit.Enabled = False
        End If
    End Sub
```

instat/ucrGeom.vb
```diff
@@ -56,7 +56,9 @@ Public Class ucrGeom
                 iGeomIndex = lstAllGeoms.FindIndex(Function(x) x.strGeomName = clsTempFunc.strRCommand)
                 If iGeomIndex <> -1 Then
                     clsCurrGeom = lstAllGeoms(iGeomIndex)
+                    bResetGlobal = bReset
```
lloyddewit, line 59: `ucrInputGeoms.SetName` does not seem to use `bResetGlobal `, and it is set to false 2 lines later. Do we need this line?

instat/ucrGeom.vb
```diff
@@ -1651,9 +1653,12 @@ Public Class ucrGeom
             If iNewGeomIndex <> -1 Then
                 clsCurrGeom = lstAllGeoms(iNewGeomIndex)
                 If iNewGeomIndex <> iCurrentGeomIndex Then
-                    For Each clsParam In lstAllGeoms(iCurrentGeomIndex).clsLayerParameters
-                        clsGeomFunction.RemoveParameterByName(clsParam.strLayerParameterName)
-                    Next
+                    'Why do we have to remove the parameters here?
```
lloyddewit, line 1656:
```suggestion
                    'TODO Why do we have to remove the parameters here?
```

lloyddewit (changes requested):

instat/ucrGeom.vb
Ivanluv: its used by the event `ucrInputGeoms.ControlValueChanged` thats called during `ucrInputGeoms.SetName`

lloyddewit: OK, thanks for the explanation

lloyddewit (approved):

lloyddewit:

lloyddewit:
@Ivanluv As suggested by @dannyparsons, this PR affects all graphics dialogs, so @dannyparsons or @volloholic will also review before we merge.

dannyparsons:
Thanks I'll discuss with @volloholic about reviewing.

rdstern (approved):

PR #5982, opened 2020-09-15
Pasting into data frame columns

### LINKED ISSUE #5619: Adding a simple paste into a data frame

Patowhiz:
Fixes #5619


dannyparsons:
The approach here is not quite what we want. You shouldn't do anything directly in the Data Grid as is done in `PasteFromClipboard` for example. The Data Grid is only ever updated from R, otherwise we are not in sync with the data in R. It's always a one-way process and that's fundamental to how R-Instat works.

The values to paste should be sent directly to R. This can actually be much simpler because there are R functions (`readClipboard()`) and an R package `clipr` (https://cran.r-project.org/web/packages/clipr/index.html and used by `rio`) which can directly read from the clipboard themselves, so we don't even need to process the clipboard values in VB before they go into R.

I think all we actually need is a method in the DataBook/DataSheet which will read from the clipboard and attempt to paste at a cell value given from the front end e.g. `data_book$paste_from_clipboard(data_name = "sheet1", row = 1, column = "date")`. All of the validation, checking and inserting into the data will then be done in this method.

This approach could also replace our current method for typing into a single cell since it's the same principle.

rdstern (changes requested):
@Patowhiz I started testing this, but, having seen @dannyparsons comment, he is right on the correct way to do this.  It will (I am afraid) change the code quite a lot. 
It will probably be an automatic fix for the one thing I tried, namely to copy a few numbers from the column of an R data frame and then paste them back.  This didn't work, because the copy added a ' at the start and end!

Patowhiz:
@dannyparsons thanks for looking at this. I don't understand what you mean by doing anything directly in data grid. The method never reads or writes to the datagrid. As you've suggested, It's simply sends values with the R command to R, afterwhich R will update the datagrid. Values are read from the clipboard.

I absolutely agree that we could push the checking into R, I did the "checking" of values in VB.NET to make it consistent with the editing of cells code. As you have correctly questioned this, I had also asked myself why it was done that way instead of using an R package but then I could see benefits of that.

I will now push the whole functionality to R. I think the algorithm used would relatively be the same.

@rdstern do you mean copying from R-Instat data frame? If so I think the problem could be in the copy but I'll check this. Fundamentally,  any application will read from the same windows OS clipboard. That means they will be exposed to the same text format; whether it's VB or R.

dannyparsons:
Great, I think we are on the same page.

Patowhiz:

dannyparsons:
This looks good. I'll go through in more detail soon. One first point, I don't think we need to require the user to highlight the same number of columns as is being pasted. I was assuming the pasting starts from a cell as the top left corner. This is how pasting works in Excel. 
I can see the use of getting the user to highlight the columns they think they are pasting into to confirm it is the same. But I think this will be a bit too unnatural and difficult to easily do for a lot of columns.

dannyparsons:
Maybe there should be a pop up box before pasting which says, "You are about to paste into x columns and y rows. Are you sure you want to do this?" Since we don't have an undo like Excel I don't mind having a pop up to confirm.

rdstern (commented):
@Patowhiz this is exciting in making R-Instat easier to use.  Also dangerous (as you say) because we don't (yet) have an undo.  And, unlike the old Instat, we can't "lock" variables.  I don't know if this is easy/possible in R?
I have only tried from within a data frame and that is already interesting.
a) We do have Prepare > Data Frame > Add Columns/Rows as a dialogue, which I used.  Then I copied easily into that new variable.  Good.
b) I wasn't able to copy the name over.  I didn't try just the name.
c) I copied a numeric variable and it made it character.  I could live with that, but could it copy as numeric instead, either because it was numeric before, or detecting the variable is numeric?
d) I am protecting the right-click menu now.  So I suggest that we don't even have the Cell > Copy or Cell > Paste.  We have them on the edit menu and we also have an icon for each on the toolbar.  This is the time to bring those into operation.
e) It overwrites with no warning.  Should we ask if it is copying where there is existing data?  Just like we do, when deleting a variable?
f) In Excel I am able to make a new sheet (there is a little (+) at the bottom.  Then I often copy into that.  Could we copy into a new sheet?
g) If so, then could we start from no data, into a sheet.  That could link with an issue to make our initial no data window look better!

dannyparsons:
I've just realised that one big drawback to this approach is that we lose all reproducibility when reading the clipboard is handled by R. Scripts won't re-run the same when the clipboard is different and scripts can't be checked for accuracy when the data is unknown.
I think we can fix this by keeping the same structure but by reading the clipboard in VB and passing the contents to the R method. The code might be a bit ugly for large copies, and we need to pass it through in a compatible way, but this will give the reproducibility as the pasted data will be in the script.

Patowhiz:
@dannyparsons in regards to users highlighting columns, I thought it wise to introduce this restriction because I may have unintentionally copied 2 columns yet I wanted to paste just one. Going by the fact that we don't have an undo then making this as strict as possible, doing exactly what the users intends at the sametime doing the obvious standard "practices" alleviates unintentional changes to the data. Excel notifies you on the differences in the length of the data but again if you happened to have ignored the warning, you can just undo. However I agree that if I start pasting from the first cell in excel, it naturally pastes horizontally and vertically.  Looking into this feature, I'm limited by the fact that to make it natural to excel as possible, new columns will have to be dynamically created if the pasted columns are more than what the data frame has. The alternative would be ignoring the extra pasted columns which as you correctly stated previously is not a good user experience .

@dannyparsons in regards to reading the clipboard in VB.NET, that is just one of the benefits I had foreseen, prompting me to use the earlier approach(which clearly was going to involve more work in reshaping the data but a huge plus in looking at the pasted data in the output window). But then this operation has no dialog involved and all the checks to be done can be exhausted, which means there is very little if any, that can go wrong. I think there's also not much to look at in the output window  and showing all the data copied would be an unnecessary information to the user. So I'm struggling in understanding why we may want reproducibility in this case especially when we can be strict enough to ensure nothing goes against the intention of the user. 

I'm happy to reimplement the pasting like initially done, I can see benefits in both ways though I'm not fully convinced why this is far from a good alternative. 

I agree we could present a warning that pasting cannot be undone and advice users to paste on a new column(which is a feature that I strongly advocate for). But again it's important to be cognizant of the fact that users hate unnecessary warnings. So for instance,  as a user if I clicked on paste, on existing data. I already know it's existing because I highlighted it , that's why the right click appeared.  So warning me or prompting me will be double input. Excel tries to warn or prompt only in special cases. 

Of course to fix the annoying prompts/warns, we can simply enable an option to switch off the warns and remembering the user input.

@rdstern on item (e),  looking at the different spreadsheet packages I thought it's now standard that pasting on existing data will overwrite. Excel gives no warning in regards to this. So should we really give thise kind of warning?

@rdstern I will look at your item (c), it seems odd to me. Should copy numeric as a numeric. It's been designed to do so.

@rdstern on item (g), yes. Actually,  to me, this will be the holy grail of paste feature in R-Instat. Simply copying a whole sheet from a spreadsheet to R-Instat in 1 step. In addition,  I expect users to paste to new columns than existing columns, or to new sheets. Probably it's cause that's how I do it. My observations show that it's natural to see a user pasting data to a new column just to be sure instead of to an existing column. So I expect that feature of copying to new columns to be used more than copying to existing columns.

@rdstern on item (d), I think looking at the common spreadsheets it's seems standard to have copy paste on the right click menu toolstrip. Users are already used to this. I think omitting this would feel unnatural from the user's perspective.  Of course not to mention they compliment the Ctrl + V shortcut keys.

@rdstern on pasting to an existing column, I wasn't sure if it would be the correct thing to do in regards to renaming the column. I can have the column renamed, I  think it's sensible to have it renamed. Because uf a user doesn't want this, then he/she can paste starting from the first cell.

Thank you @dannyparsons and @rdstern for the concrete feedback.

dannyparsons:
Let's separate out some of these things because they don't all necessarily relate to each other:

Highlighted columns: Let's see how Roger finds it, there are issues with either approach. We might keep the highlighting approach and see how users respond to it. If it's too awkward to use then we can consider pasting multiple columns without highlighting.

Pasting new columns: This is independent of the first point. Even with pasting from a single cell I would still want an error if the pasting had too many columns. I don't think it's very important to have this feature (and I am not sure it's sensible as its easy to introduce mistakes) so let's do the main implementation first. It would be easy to allow this after if we decided.

Warning box: I don't think pasting is going to be very common, and it should be done carefully, so I don't mind having a warning box which tells you how many rows and columns are being pasted, and include an option to turn it off if they wish.

Clipboard in VB: Reproducibility is science and research in general is so important. The reason its important in R specifically is that it allows others to scrutinise and reproduce your analyses from the same scripts. If I send you my scripts and data files you can rerun everything I did and check the results are accurate. Lack of reproducibility is a big downside of a menu driven software, unless it has a language behind it producing scripts, which is why we have insisted on all data processing in R-Instat being done in R. If someone produces research using R-Instat, they can provide the R scripts and there is transparency in what was done. It's a huge topic in general, this page is quite useful in relation to R https://nceas.github.io/sasap-training/materials/reproducible_research_in_r_fairbanks/index.html

So this is the main reason to have the clipboard done in VB. Otherwise, if I send you my R script that includes reading from the clipboard it won't work because you don't have the same thing copied. Or if I want to run my own script again tomorrow it won't work because I won't have the same thing copied.

So unfortunately reproducibility has to be our highest priority. I agree it's not necessary for the user to see the copied data in the output window, and it could look ugly and it's much neater if this is hidden, but we will have to do it as its the only way to ensure reproducibility.

However, this doesn't mean we have to go back to doing all the checks and manipulations in VB. We want to keep the checking in R, just with the data coming from VB. What I suggest if we pass the clipboard data from VB as a simple string into the R method as an extra parameter to the existing paste method. Then all the manipulating and checking on this data is done within the R method. The R method can be adapted to be able to cope with copied data in whatever form it is provided from VB. In fact `read_clip_tbl` allows you to provide an alternative string as a parameter. So I don't think you will have to change much about this implementation.

These are all useful discussions. I hadn't thought through all these issues before so it's good we are figuring out the details now together.

Patowhiz:
@dannyparsons reproducibility in that context sounds sensible to have. Will include it. Yep the clipr package has the feature neatly built in.

Will do that as I wait for @rdstern feedback on the other items.

rdstern:
I reply on some items.  I should say I am also fully in support of the idea of reproducibility.  @Patowhiz thank you for your patience on the extra work.  It is in a good cause!

P: "on item (e), looking at the different spreadsheet packages I thought it's now standard that pasting on existing data will overwrite. Excel gives no warning in regards to this. So should we really give this kind of warning?"

R: I think that once the feature is completed, then the main use will be pasting into new variables (perhaps with NA), and here there would be no warning.  If we had undo, then we could do without a warning - which almost all the other software would have.  We also have the warning for other operations, like deleting variables, which I use regularly and don't find it a problem.  Excel also has a facility for locking sheets or variables, which we don't have yet.  In the future we might also have an option (on the toolbar) to temporarily turn warnings off.  So, for now I would like to be cautious and warn.

R: You agree on my items c) copying numeric as numeric.  That was into a new variable.  It also isn't such a big deal, because itg is easy to make the variable numeric.

R: You also are in agreement on item g) copying whole variables or whole sheets.  Very exciting to be able to do this!

P: "on item (d), I think looking at the common spreadsheets it's seems standard to have copy paste on the right click menu toolstrip. Users are already used to this. I think omitting this would feel unnatural from the user's perspective. Of course not to mention they compliment the Ctrl + V shortcut keys."

R:  Yes, many packages seem to have mainly items like Copy and Paste on their right-click menu.  However, our right-click menu is already pretty long with items from the Prepare menu.  And we have a pleasantly small toolbar with Copy and Paste there.  So I would be going "in the other direction", i.e. the odd Cell sub-right-click could now be deleted!  This is not something I feel very strongly about though I would like the toolbar to be active and (of course) I really like <ctrlC> and <ctrlV>.

P "on pasting to an existing column, I wasn't sure if it would be the correct thing to do in regards to renaming the column. I can have the column renamed, I think it's sensible to have it renamed. Because uf a user doesn't want this, then he/she can paste starting from the first cell."

R: I am also not sure what is best here.  And I don't feel strongly about this issue.  The one area where it may be more important is when pasting to a new sheet.  That might become an option, or a Paste Special that gives an option?

dannyparsons:
I think pasting as a new data frame should be a separate option and that's much safer as there's no overwriting. On the Edit menu we could have Paste and Paste as new Data Frame. On the toolbar Paste could be a single item with a dropdown for the two options. 
Paste as new Data Frame could be its own dialog as there are options such as whether headers are included and the name for the new data frame. Or it could also be a separate tab on the New Data Frame dialog. I think it probably needs to be in a dialog in some way because of the options to specify.

Patowhiz:
@rdstern @dannyparsons  you can now test the new additions.

rdstern (commented):
@Patowhiz I think this is working as far as it goes.  The one problem I found was when it changed the name of the Yield variable to the first observation in the data.  Here is the command:
# Paste values in Data
data_book$paste_from_clipboard(data_name="survey", col_names=c("yield"), start_row_pos=1, first_clip_row_is_header=TRUE, clip_board_text="2.5

I look forward to the toolbar being active

I look forward to the ambitious extensions, namely to paste into new column(s) and to paste a new sheet.  When pasting into new columns, then it could roughly be the same as Enter.
1) If the paste is too long, then it is truncated.
2) If the paste is too short, then the rest is NA.  (It could be - like a regular sequence, that there is an option to recycle the data.

Patowhiz:
@rdstern  yes. I explained the implementations. When you paste from the column header. The first row of the copied data is used to rename the column. So in this case you copied a single line of text and pasted from the column header. However if you paste from a specific cell, then the column header will not be renamed. This functionality is meant to be similar to excel.

I have created individual issues for the other "types" of pasting you've mentioned above.

lloyddewit:

dannyparsons:
I've discussed with @rdstern and we think the pasting into an existing sheet should be kept simple and therefore it does not allow for also renaming the columns. So we suggest always to only paste values and not rename columns. I think it's getting too clever and too many potential issues to try and do both at the same time.

I do think it could be useful to rename columns through pasting, but I think we can think of another way to allow users to paste a new set of column names, which could be from a special paste option/dialog. Also, when pasting to create a new sheet, as mentioned before, I think this should go through a dialog and there would be an option to include headers in the paste for that case.

Patowhiz:
@dannyparsons @rdstern  so you suggest that the entire data copied to always be pasted? Should the first row be skipped when pasting into column through the header right click?

Originally, pasting from a cell meant all the copied data including the first row was going to be pasted. Pasting from the column header right click meant that the first row was going to be skipped and assumed to be the header(this is the default header option in clipr).

What would you like me to do in the 2 different scenarios?
One point to note is, for instance if I copy numeric values from a spreadsheet as column, the first row would most of the time be a character.

Patowhiz:
@rdstern I've made the changes  suggested above. I have simplified the pasting to be just the data. Row headers not recognised for both pasting from a cell or from the column header.
This is now ready for review.
This PR blocks issues #5991, #5992, #5993 and #6020 to be done. They affect the files.

rdstern (approved):
@Patowhiz this is now working well - as far as it goes.  I hope this can be merged and then more features can be added.
a) I was able to paste columns, or parts of columns if they fit.
b) If we want to add columns then first we need to add them with NA and then paste into the now existing columns.  
c) Similarly, for now, I can make a new (empty) data frame and paste multiple columns into it.  Great.  So this is already useful.
I therefore would be very happy to see it merged.  Over to others, including @lloyddewit to check the code, so it can be merged quickly.

lloyddewit:
@shadrackkibet please could you complete your review? thanks

dannyparsons:
My view is that column names are only included when pasting to create a new data frame. When you're pasting into an existing data frame - whether it's empty or not - you only paste into the values. That makes sense to me as it keeps the pasting simple and clear and reduces risk of error, which is important especially without an undo option.

That could mean we don't have the pasting option from the header right click if that makes it confusing for users.

I will discuss with Roger offline so that we agree on the details and can confirm with you the functionality to settle on so that you can finalise this.

(I do think renaming columns though copy-and-paste is also a useful feature, I just think it should be a separate process to pasting values into a data frame, possibly as a separate dialog or menu option.)

Patowhiz:
@dannyparsons yes. @rdstern also explained the same. I have simplified this functionality to just pasting the data. Row headers are always assumed. I have also left the  code for pasting the row headers if in future that functionality is desired by the developer elsewhere(probably in the paste special dialog for new columns that I'll be implementing)

In terms of the data, the column types that can be pasted on are; factors, ordered factors, numerics, integers, characters and logicals

Patowhiz:
@dannyparsons @rdstern this is also ready for review.

rdstern (changes requested):
I looked at the last changes made.  The types of data don't include Date (D) variables, which would be useful.  And there will soon be date/time variables.  There are also more specialised variables, for example circular, which perhaps one shouldn't be able to paste into?

The paste operation is currently mirrored into the output window, which I suggest isn't needed.

However, I would like work to proceed on this topic, so I am happy to approve soon.  Pasting is useful even if there are some exceptional circumstances where it doesn't work.

Patowhiz:
> I looked at the last changes made. The types of data don't include Date (D) variables, which would be useful. And there will soon be date/time variables. There are also more specialised variables, for example circular, which perhaps one shouldn't be able to paste into?

@rdstern yes, clipr also doesn't automatically recognise date columns, however, we could technically detect and have them included if need be. We can try and do this in pasting new columns and pasting new data frame.

> The paste operation is currently mirrored into the output window, which I suggest isn't needed.

@rdstern that is what I had implemented earlier,  but @dannyparsons made a good and important point on reproducibility. Thus the new change. If we had a way of detecting amd collapsing the long parameter values, we could achieve both suggestions. But at the moment all parameter values are shown in the output window, therefore the clipboard data which is now a parameter value also has to be shown in the output window.

rdstern:
I looked at this again and it seems to work.  I guess we have to accept the pasting to the output window.  I pasted 50,000 values into the data frame and it was odd to have them all in the output window.  R has some functions that just put a limited number of rows and then say "Continued with xxxxx rows" which would be much nicere, but may not be possible.

I think the one problem I found may have nothing to do with this dialogue.  I produced new variables using the Prepare > Data Frame > Insert Columns/Rows dialogue, and then pasted into those variables.  Very nice!  But then I got this error when I returned to that dialogue:
image

I think this may be unrelated to the pasting and to be fixed elsewhere by @Patowhiz .  Good to check.

dannyparsons:
There are ways for the copied data to be retrieved from R directly without having to be pasted through in the code that me and Patrick looked into at first, but then the code is not reproducible because when you run it again with something else copied to the clipboard it will do something different. So this is the only way to have a reproducible script file. This would be a big negative of R-Instat for some people if we had copy-and-paste that wasn't captured in the R code. Although it makes the output look a bit ugly I think it's necessary. In any case, I don't think it would be good practice to copy-and-paste large amounts of data because then you should be importing a file.

lloyddewit:
@rdstern Based on @dannyparsons  response, are you now willing to approve?
@shadrackkibet did you get a chance to progress your peer review?
Thanks

rdstern (dismissed):

lloyddewit:
@shadrackkibet waiting for your review. PR #6061 is currently blocked by this PR
@dannyparsons I'll give you a chance to review this PR before we merge

shadrackkibet:
Can someone sort out these conflicts so that I can proceed with my review? Thanks

lloyddewit:
@Patowhiz as PR author, please could you resolve the conflicts? thanks

dannyparsons:
@Patowhiz Can you have a look at the conflicts here? Let me know if you need any help. Then I am ready to review.

dannyparsons:
@Patowhiz can you look at the conflicts here and see what can be resolved? Roger is asking for this to be reviewed and merged.

maxwellfundi:
@Patowhiz I can help with the conflicts on the form if am advised which changes to keep. The ones on this branch or on the master branch. I assume it would be the changes on this branch?
@dannyparsons I wonder if you are able to look at the data file conflict to determine what we keep so that t his is reviewed?

maxwellfundi:
Alternatively @Patowhiz  we can accept the master data object and you quickly make the changes again on it? Does that work?

Patowhiz:
The conflicts here are so many. I tried to resolve them but wasn't successful. So I'm creating a new PR that will replace this.

Patowhiz:
This PR has been replaced with PR #6334. So I'm now closing it.

PR #5983, opened 2020-09-16
Haward ketoyo msatsi circularmenu

dannyparsons:
This replaces #5970 

The Calculator and Summaries sub dialog opens on Circular options the first time the dialog is opened (if opened from the Circular menu). In subsequent opens the dialog opens as it was closed, regardless of the opening location.


rdstern (dismissed):
This looks fine from my side.  I note the currently non-existent dialogues ae disabled.  Work will start on them immediately and I assume the dialogues can be added without changing frm main?  Over to @lloyddewit for a second check and then merge

instat/sdgSummaries.vb
```diff
@@ -481,9 +484,23 @@ Public Class sdgSummaries
...
+                        Exit For
+                    End If
+                Next
+                ' Could give a developer error here as it means strDefaultTab was not recognised.
+                If Not bTabFound Then
+                    tbSummaries.SelectedIndex = 0
+                End If
+            Else
+                tbSummaries.SelectedIndex = 0
+            End If
```
lloyddewit, line 503: Simplify?
```suggestion
            tbSummaries.SelectedIndex = 0
            If strDefaultTab <> "" Then
                For i As Integer = 0 To tbSummaries.TabPages.Count - 1
                    If tbSummaries.TabPages(i).Text = strDefaultTab Then
                        tbSummaries.SelectedIndex = i
                        Exit For
                    End If
                Next
            End If
```

instat/dlgCalculator.vb
```diff
@@ -25,6 +25,9 @@ Public Class dlgCalculator
     Public bFirstLoad As Boolean = True
     Public iHelpCalcID As Integer
     Private iBasicWidth As Integer
+    Private strDefaultKeyboard As String
+    ' Note: This list needs to be updated when a new keyboard is added.
+    Private strKeyboards() As String = {"Maths", "Logical and Symbols", "Summary", "Strings (Character Columns)", "Factor", "Probability", "Dates", "Transform", "Circular", "Wakefield", "Modifier", "Symbols", "hydroGOF"}
```
lloyddewit, line 30:
```suggestion
    Private strKeyboards() As String = {"Maths", "Logical and Symbols", "Summary", "Strings (Character Columns)", "Factor", "Probability", "Dates", "Transform", "Circular", "Wakefield", "Modifier", "Symbols", "HydroGOF"}

```

instat/dlgCalculator.vb
```diff
@@ -88,6 +96,14 @@ Public Class dlgCalculator
         ucrCalc.ucrTryCalculator.StrvecOutputRequired()
     End Sub
 
+    Public Sub SetDefaultKeyboard(strNewDefaultKeyboard As String)
+        If Not strKeyboards.Contains(strNewDefaultKeyboard) Then
+            MsgBox("Developer error: there is no Calculator keyboard called" & Chr(34) & strNewDefaultKeyboard & Chr(34) & vbNewLine & "Default keyboard will be selected.")
+            strDefaultKeyboard = ""
+        End If
+        strDefaultKeyboard = strNewDefaultKeyboard
```
lloyddewit, line 104: Move this line to before the `If` statement (else it cancels the affect of line 102)?

instat/dlgColumnStats.vb
```diff
@@ -25,6 +25,7 @@ Public Class dlgColumnStats
     Public strDefaultDataFrame As String = ""
     Public strDefaultVariables() As String
     Public strDefaultFactors() As String
+    Private strDefaultTab As String = ""
```
lloyddewit, line 28: Cleaner to have this as a `Property`?

lloyddewit (changes requested):
@dannyparsons this looks good, I just had a couple of small suggestions

instat/dlgColumnStats.vb
dannyparsons: Probably yes, but for consistency with the other variables already in the dialog I thought having it like this.

instat/sdgSummaries.vb
dannyparsons: Not sure why I didn't add the developer error in straight away, I've added it now. With the developer error added I like this approach so we explicitly check when an incorrect value is given. Does this look ok or could it be more simple?

instat/dlgCalculator.vb
dannyparsons: Thanks, done.

instat/sdgSummaries.vb
```diff
@@ -481,9 +484,23 @@ Public Class sdgSummaries
...
+                For i As Integer = 0 To tbSummaries.TabPages.Count - 1
+                    If tbSummaries.TabPages(i).Text = strDefaultTab Then
+                        tbSummaries.SelectedIndex = i
+                        bTabFound = True
+                        Exit For
+                    End If
+                Next
+                If Not bTabFound Then
+                    MsgBox("Developer error: there is no Summary tab called" & Chr(34) & strDefaultTab & Chr(34) & vbNewLine & "Default tab will be selected.")
+                    tbSummaries.SelectedIndex = 0
```
lloyddewit, line 499: If line 484 above is not deleted, then this line is not needed.

instat/sdgSummaries.vb
```diff
@@ -481,9 +484,23 @@ Public Class sdgSummaries
...
+                End If
+            Else
+                tbSummaries.SelectedIndex = 0
```
lloyddewit, line 502: If line 484 above is not deleted, then these lines are not needed.

lloyddewit (changes requested):
@dannyparsons Just a couple of very small comments. Please accept/reject as you wish.

instat/sdgSummaries.vb
dannyparsons: Yes that was deleted here https://github.com/africanmathsinitiative/R-Instat/pull/5983/files#diff-e8e5e936008fdc29b42fa8745c491d97L484

lloyddewit (approved):
@dannyparsons from your comment, I assume that you'd like to leave the extra lines in, no problem

@rdstern Please could you re-approve? thanks

rdstern (approved):
It seems as though the "system" of checking is working!

PR #5984, opened 2020-09-17
New seasonality plots dialog

### LINKED ISSUE #5980: Seasonality Plots dialog
This is the design for the seasonality plots dialog. This is to compare the seasonality of two columns, in general "reference" data and "estimates". The graphs are of summary data based on the seasonality column chosen, but the dialog uses the original (usually daily) data because that is the level where the calculations are done. Because of this the R code is slightly none trivial. Here is the type of output that we want to produce:

The colours are the reference and estimates data, the values on the y-axis as raindays (proportion of days above 0.85) and day of the year on the x-axis. The smooth line is some kind of model fitted to the data and the wiggly line is the 5 day moving average. The moving average line shows us that the curves fit quite well and the curves show that both get the same seasonality shape but the estimates consistently overestimate raindays.

image

There 4 receivers: reference & estimates are the columns to compare, seasonality (X axis) is the seasonal column e.g. day of year or dekad or month etc. The last is station, I forgot it on the design but add it to the bottom. This is optional.

Under reference and estimates there is a drop down box to choose the summary to display (for each value of the seasonality column). The options there are: "Prop Above", Prop Below" and "Mean". For example, for rainfall we would usually use "Prop Above" to give proportion of values above 0.85mm which is rainy days. For temperature, we would usually want "Mean". If "Prop Above" or "Prop Below" is chosen then a text box appears next to it to give the value. Before the text box add a label for ">" or "<".
This is for each column because you may want different summaries for the reference and estimates data e.g. Prop Above 0.85 for reference and Prop Above 5 for estimates.

The check box: "Remove rows with any missing values" is checked by default. This filters the data to only use values where both columns are not missing. This is usually sensible so that the comparisons are using the data time periods. If unchecked, the filter is removed.

The two group boxes set what is displayed on the graph.

The first is "Values" which has a check box to display the points and another to display the moving average. The points is `geom_point()` and moving average adds `geom_line(aes(y = moving))` and adds the calculation: `mutate(moving = raster::movingFun(prop, n = 5, circular = TRUE, na.rm = TRUE))`. A text box appears to set the value of `n` which is `5` by default. Notice that `circular = TRUE` which means the values are "wrapped round" so that, for example, for the day of year the values at 365, 366, 1, 2 and 3 are used in calculating the 5 point moving average for day 1.
I think we could have moving average checked by default with 5 and points unchecked.

The "Smoothing" group box sets what smooth curve is fitted to the data. A smooth curve is useful to see the general shape in a smoother way than the points or moving average which makes it easier to compare. There are different options that do different things but will give similar curves. The options in the drop down box are: "Fourier Series", "Periodic Splines", "Natural Splines" and "LOESS". 
For Fourier Series there are 2 parameters to specify, Harmonics (default 3) and Period (default 366). Harmonics is number of sin/cos pairs to include in the fitting and the period is the length of the seasonal cycle e.g. 366 for day of year.
For Periodic and Natural Splines the d.f. (degrees of freedom) parameter is given, default 6. This is how flexible the line can be basically.
For LOESS the span parameter controls how much of the data is used to fit at each point, so a lower span is more flexible. I suggest our default for this could be 0.5 (the R default is 0.75). It could be a nud with 2 decimal places. The minimum is 0 but it has an unlimited maximum.

Here is the R code to do the dialog. In this example, `zm` is the data frame chosen, `x1` is the reference column and `x2` is the estimates column. `sdoy` is the seasonality column and `station` is the station column.
I've included two versions of the `zm_seasonal` calculation to show how it should work for a proportion and how it should work for a mean.
We always calculate `n` in `zm_seasonal` because this is the needed as the `weight` aesthetics which is important for the fitting to know how many observations make up each point.
I've included all the possible curves in the graph to show how they work but the dialog will never allow such a complex graph. There will only be one `geom_smooth` in the graph at a time.

``` r
# add this function to stand alone functions file
fourier_series <- function(x, n, period) {
  p2 <- "2 * pi"
  h <-  seq_len(n)
  paste0("sin(", x, " * ", h, " * ", p2, " / ", period, ")", " + ", 
         "cos(", x, " * ", h, " * ", p2, " / ", period, ")", 
         collapse = " + ")
}

# if the summary is prop
zm_seasonal <- zm %>%
  # this is controlled by the missing checkbox
  filter(!is.na(x1) & !is.na(x2)) %>%
  mutate(reference = x1 > 0.85,
         estimates = x2 > 0.85) %>%
  # this stacks the data which is needed for the graph
  pivot_longer(cols = c(reference, estimates), names_to = "data", values_to = "thresh") %>%
  group_by(station, data, sdoy) %>%
  summarise(n = sum(!is.na(thresh)),
            prop = sum(thresh, na.rm = TRUE)/n) %>%
  # this is only added if moving average option is selected
  mutate(moving = raster::movingFun(prop, n = 5, circular = TRUE, na.rm = TRUE))

# if the summary is mean
zm_seasonal <- zm %>%
  # this is controlled by the missing checkbox
  filter(!is.na(x1) & !is.na(x2)) %>%
  mutate(reference = x1,
         estimates = x2) %>%
  # this stacks the data which is needed for the graph
  pivot_longer(cols = c(reference, estimates), names_to = "data", values_to = "value") %>%
  group_by(station, data, sdoy) %>%
  summarise(n = sum(!is.na(value)),
            m = mean(value, na.rm = TRUE) %>%
  # this is only added if moving average option is selected
  mutate(moving = raster::movingFun(m, n = 5, circular = TRUE, na.rm = TRUE))

ggplot(zm_seasonal, aes(x = sdoy, y = prop, colour = data, weight = n)) +
  # points
  geom_point() +
  # Moving average
  geom_line(aes(y = moving)) +
  # Fourier series
  geom_smooth(method = "glm", formula = as.formula(paste("y ~", fourier_series("x", 3, 366))),
              method.args = list(binomial(link = "logit")), se = FALSE) +
  # periodic spline
  geom_smooth(method = "glm", formula = y ~ pbs::pbs(x, df = 6),
              method.args = list(binomial(link = "logit")), se = FALSE) +
  # natural spline
  geom_smooth(method = "glm", formula = y ~ splines::ns(x, df = 6), 
   method.args = list(binomial(link = "logit")), se = FALSE) +
  # loess
  geom_smooth(method = "loess", formula = y ~ x, se = FALSE, span = 0.5) +
  # facet by station if given
  facet_wrap(~station)
```

Then there is the usual save control because the output is a graph.

If you're interested in the background here are some links but I suggest looking at the code and getting started on it and not spending lots of time reading:
- A short explanation of why you might use Fourier Series for seasonal data instead of polynomials https://stats.stackexchange.com/questions/224990/whats-wrong-to-fit-periodic-data-with-polynomials
- An explanation of splines with something at the end of natural splines http://www.clayford.net/statistics/using-natural-splines-in-linear-modeling/. Circular splines are almost the same except with additional requirements that the beginning and end of the fitted values line up in the same way it does naturally for the Fourier Series.
- A good explanation of LOESS https://www.itl.nist.gov/div898/handbook/pmd/section1/pmd144.htm 

seasonality_dialog

@HawardKetoyoMsatsi would like to take this up with help from @shadrackkibet? It's not the most simple dialog but I think it's a good one if you are keen. It does need to be done quite soon so we should discuss priorities with @rdstern if you do take it up. Let us know what you think.

HawardKetoyoMsatsi:
Fixes #5980. @dannyparsons @rdstern the seasonality comparisons dialog is ready please have a look.  @shadrackkibet and I worked on the issue.

### DIFF
--- a/instat/dlgSeasonalPlot.Designer.vb
+++ b/instat/dlgSeasonalPlot.Designer.vb
@@ -22,13 +22,6 @@ Partial Class dlgSeasonalPlot
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()> _
     Private Sub InitializeComponent()
-        Me.ucrBase = New instat.ucrButtons()
-        Me.ucrSelectorSeasonalityComparisons = New instat.ucrSelectorByDataFrameAddRemove()
-        Me.ucrReceiverReference = New instat.ucrReceiverSingle()
-        Me.ucrReceiverStation = New instat.ucrReceiverSingle()
-        Me.ucrReceiverSeasonality = New instat.ucrReceiverSingle()
-        Me.ucrReceiverEstimate = New instat.ucrReceiverSingle()
-        Me.ucrChkRemoveMissing = New instat.ucrCheck()
         Me.grpValues = New System.Windows.Forms.GroupBox()
         Me.ucrInputN = New instat.ucrInputTextBox()
         Me.lblN = New System.Windows.Forms.Label()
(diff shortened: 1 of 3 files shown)


shadrackkibet:
I note that the labelling on the y-axis will be clearer with either "prop" or "mean". Currently, it always labels it "calc_summary". We will update that accordingly.

rdstern (commented):
That's exciting.  It is a complicated dialogue and works!  
a) I suggest that doy will usually be the seasonality.  So it could be included in the dialogue as the default.
b) There are the 3 options of Prop >, Prop < and Mean.  I suggest we make Mean the default and it will be hardly any extra work to add Median.
c) The Station will define the facets.  Could we copy the control for the Station from the Climatic Boxplots dialogue.  The default could be Columns, see below, but including None is also useful.  Then it averages over the stations, which works now if you omit the Station field and could sometimes be useful.  (I used the Ghana data.)
d) I tried changing the seasonality to month and it just gives the 12 points - as points.  We need to consider what is wanted here.  We could plot each as a line I assume?



I found a couple of oddities.  I tried lots of things and changed the seasonality from doy to month.  Then got this:
image

But I couldn't repeat it, so couldn't find where it is.  Then I got this odd graph:
image

That was from this code:
image
 and this dialogue:
image

shadrackkibet:
It is clear where problems are from the code. We will fix!

dannyparsons:
It should work in the same way if month is used instead of doy. I guess the issue was month was a factor. I checked a few things in my R script with a factor for the seasonality and it seemed to work ok but there may be something I missed.

instat/dlgSeasonalPlot.Designer.vb
```diff
@@ -22,31 +22,450 @@ Partial Class dlgSeasonalPlot
...
+        Me.lblDf.Text = "Degrees of freedom:"
+        '
+        'lblPeriod
+        '
+        Me.lblPeriod.AutoSize = True
+        Me.lblPeriod.Location = New System.Drawing.Point(128, 49)
+        Me.lblPeriod.Name = "lblPeriod"
+        Me.lblPeriod.Size = New System.Drawing.Size(45, 13)
+        Me.lblPeriod.TabIndex = 18
+        Me.lblPeriod.Text = "Periods:"
```
dannyparsons, line 239:
```suggestion
        Me.lblPeriod.Text = "Period:"
```

shadrackkibet:
Above issues should be fixed now. This is working better.

rdstern:
1. On a change in design, I wonder if there is ever a situation when we would want a different summary for the Reference and Estimated Variables?  So perhaps delete the first instance, so there are both variables and then you give the summary.
2. Now I am not getting the smoothing results plotted.  And the code looks odd.  Some lines are repeated, and it is doing a binomial link when I am fitting the temperatures.  So the fitted curves are not plotted at all. 

```
# Code generated by the dialog, Seasonal Comparisons
ghana <- data_book$get_data_frame(data_name="ghana")
max_temperature <- data_book$get_columns_from_data(data_name="ghana", col_names="max_temperature")
min_temperature <- data_book$get_columns_from_data(data_name="ghana", col_names="min_temperature")
max_temperature <- data_book$get_columns_from_data(data_name="ghana", col_names="max_temperature")
min_temperature <- data_book$get_columns_from_data(data_name="ghana", col_names="min_temperature")
station <- data_book$get_columns_from_data(data_name="ghana", col_names="station")
doy <- data_book$get_columns_from_data(data_name="ghana", col_names="doy")
doy <- data_book$get_columns_from_data(data_name="ghana", col_names="doy")
station <- data_book$get_columns_from_data(data_name="ghana", col_names="station")
last_graph <- ghana %>% dplyr::filter(!is.na(max_temperature) & !is.na(min_temperature)) %>% dplyr::mutate(reference=max_temperature, estimate=min_temperature) %>% tidyr::pivot_longer(cols=c(reference,estimate), names_to="data", values_to="value") %>% dplyr::group_by(station, data, doy) %>% dplyr::summarise(n=sum(!is.na(value)), calc_summary=mean(value, na.rm=TRUE)) %>% dplyr::mutate(moving=raster::movingFun(calc_summary, n=5, circular=TRUE, na.rm=TRUE)) %>% ggplot2::ggplot(mapping=ggplot2::aes(x=doy, y=calc_summary, colour=data, weight=n)) + ggplot2::geom_line(mapping=ggplot2::aes(y=moving)) + ggplot2::geom_smooth(method="glm", formula=as.formula(paste("y~", fourier_series("x", n=4, period=366))), method.args=list(binomial(link="logit")), se=FALSE) + ggplot2::facet_wrap( ~ station)
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="ghana")
data_book$get_graphs(data_name="ghana", graph_name="last_graph")
```

dannyparsons:
> 1. On a change in design, I wonder if there is ever a situation when we would want a different summary for the Reference and Estimated Variables? So perhaps delete the first instance, so there are both variables and then you give the summary.

I had this option specifically to do rainday occurrence with different thresholds, which could be a really nice way to understand gridded data thresholds. This is maybe only sensible for the proportion options but this was the simplest way I could think of to allow it. I guess it may be sensible for summaries to be different if the columns are different variables, for example comparing cloud fraction cover with mean rainfall?

shadrackkibet:
> Now I am not getting the smoothing results plotted. And the code looks odd. Some lines are repeated, and it is doing a binomial link when I am fitting the temperatures. So the fitted curves are not plotted at all.

Do you mean we do not need the binomial link for temperatures?

rdstern:
This is an exciting new dialogue.  My suggestions on the points above, by @shadrackkibet and @dannyparsons are as follows:
1. The binomial link is only for binomial data, i.e. the prop above and prop below.  When modelling the mean we would usually use lm, rather than glm (though it is a special case of glm).
2. On the Summary, I take @Danny's point.  about the summary being able to be different for the two datasets.  There is also more to do on these summaries.  So how about the following:

a) Have a checkbox before the Second Summary, which becomes `Summary (Optional): `  The drop-down options only appear when checked.  If necessary we could add a tooltip "Only needed if the summary is different for the 2 series"
b) Add to the options for the summaries as follows:  `Mean Above`, `Mean Below`, and also `St. Dev`.  (With Mean above and Mean below, the control to choose the value becomes visible - as with Prop Above.
c) There can be many more properties we may wish to compare.  But we can already get these summaries easily with the **_Climatic > Prepare > Climatic Summaries_**.  The third button is **_Within Year_** and is designed to get any summary we need.  So one additional option in the new dialogue is `None`.
d) Ideally, when `None` is Chosen there is another single receiver - so usually disabled - Weight (Optional):.  This would allow the Counts, when making the summary, to become the weights.

dannyparsons:
1 Yes for mean use `lm` and no link function.

a) The second summary label could be replaced with a check box saying "Different summary for estimates" which displaying the dropdown. This might make the dialog code quite a bit more complex to work correctly which is kind of one reason I had the original implementation but I can see how this would be nicer.

b) Mean above/below would require adding a `filter()` after stacking the data. Then the summary can be mean.

c) and d) is not quick to implement because all the code needs to change to be able to use directly with summary data, and the dialog would need some changes. This would be appropriate as a separate button at the top of the dialog for Primary/Summary data. Then it's almost like coding another dialog so I would suggest getting it to work well with primary data first before doing this.

rdstern:
On c) and d) I like the idea of having Data: Primary and Summary as two radio buttons at the top.  Happy that this becomes a separate issue.  I am not sure @dannyparsons approves, but I would quite like the buttons there now, with the Summary disabled, so it is clear that the intention is there.  That's like our quality control, where we still don't have spatial control, but it is important that potnetial users know that it is a future agenda item.

shadrackkibet:
> b) Mean above/below would require adding a filter() after stacking the data. Then the summary can be mean.

It is not clear to me why a `filter` is needed after stacking. Please explain further.

rdstern:
Take rainfall data as the example.  I want the mean rainfall on rainy days only, say days with > 0.85mm.  So I filter the data to get rid of all the dry days, then just take the ordinary mean.

Or maybe I have misunderstood the question.

dannyparsons:
> It is not clear to me why a filter is needed after stacking. Please explain further.

It would be like this. If I wanted mean above 0.85 I filter value > 0.85 before doing the summaries.

``` r
# if the summary is mean
zm_seasonal <- zm %>%
  # this is controlled by the missing checkbox
  filter(!is.na(x1) & !is.na(x2)) %>%
  mutate(reference = x1,
         estimates = x2) %>%
  # this stacks the data which is needed for the graph
  pivot_longer(cols = c(reference, estimates), names_to = "data", values_to = "value") %>%
  filter(value > 0.85) %>%
  group_by(station, data, sdoy) %>%
  summarise(n = sum(!is.na(value)),
            m = mean(value, na.rm = TRUE)
```

shadrackkibet:
Understood now thanks!

shadrackkibet:
> 1 Yes for mean use lm and no link function.

A `glm` with no link function is just `lm`. This will simplify code. Is that okay?

dannyparsons:
`lm` would be clearer when looking at the code as people don't think about `lm` being a special case of `glm`, so I'd prefer putting in `lm` to make that clear

HawardKetoyoMsatsi:
@dannyparsons @rdstern kindly have a look at the changes you proposed, they have been addressed.

shadrackkibet:
While you test this. We are not sure what could be sensible default thresholds for **mean above/mean below**, clearly `0.85` is not appropriate here.

shadrackkibet:
@dannyparsons @rdstern this is ready for you to review.

dannyparsons:
Thanks, I'll review. Happy with 0.85 as the threshold. That option is mostly for rainfall anyway so it doesn't matter if the default is not very appropriate for other elements.

PR #5985, opened 2020-09-17
Time series plots dialog

dannyparsons:
The basic main dialog is working.

To do:

- [x] Autofill year and station climatic variables
- [x] Allow add/remove of points and mean lines
- [x] Add summaries subdialog for annotated summaries

@rdstern the basic functionality could be tested now or could wait until the whole dialog is complete

### DIFF
--- a/instat/dlgTimeSeriesPlot.vb
+++ b/instat/dlgTimeSeriesPlot.vb
@@ -53,19 +53,31 @@ Public Class dlgTimeSeriesPlot
     Private clsPivotLonger As RFunction
     Private clsPivotCFunction As RFunction
 
-    ' Calculate means
-    ' These functions construct the R code below if mean lines are added.
-    ' df_mean <- df_stack %>%
+    ' Calculate Individual Summaries.
+    ' These functions construct the R code below if mean lines and/or if annotated individual summaries are added.
+    ' df_individual <- df_stack %>%
     '   group_by(station, data) %>%
-    '   summarise(mean = mean(value, na.rm = TRUE))
-    Private clsMeansOperator As ROperator
-    Private clsMeansGroupBy As RFunction
-    Private clsMeansSummarise As RFunction
-    Private clsMeanFunction As RFunction
+    '   summarise(mean = mean(value, na.rm = TRUE)
+    '             sd = sd(value, na.rm = TRUE))
+    Private clsIndividualSummariesOperator As ROperator
+    Private clsIndividualSummariesGroupBy As RFunction
+    Private clsIndividualSummariesSummarise As RFunction
+    Private clsNIndividual As RFunction
+    Private clsMean As RFunction
+    Private clsSd As RFunction
+    Private clsSlope As RFunction
+    Private clsPasteNIndividual As RFunction
+    Private clsPasteMean As RFunction
+    Private clsPasteSd As RFunction
+    Private clsPasteSlope As RFunction
+    Private clsSignifMean As RFunction
+    Private clsRoundSd As RFunction
+    Private clsSignifSlope As RFunction
 
-    ' Calculate Text Summaries
-    ' These functions construct the R code below if annotated summaries are added.
-    '  df_summary <- df %>%
+
+    ' Calculate Comparison Summaries
+    ' These functions construct the R code below if annotated comparison summaries are added.
+    '  df_comparison <- df %>%
     '    group_by(station) %>%
     '    summarise(n = n(),
     '              cor = cor(estimate, gauge, use = "na.or.complete"),
@@ -74,7 +86,7 @@ Public Class dlgTimeSeriesPlot
     Private clsSummaryOperator As ROperator
     Private clsSummaryGroupBy As RFunction
     Private clsSummarise As RFunction
-    Private clsN As RFunction
+    Private clsNComparison As RFunction
     Private clsCor As RFunction
     Private clsMe As RFunction
     Private clsMae As RFunction
@@ -101,7 +113,7 @@ Public Class dlgTimeSeriesPlot
     ' A dictionary of the summary functions above. The key is the summary name in lower case e.g. "bias"
     ' The value is a tuple of two RFunctions. Item1 is the summary function. Item2 is the paste function containing the summary.
     ' This dictionary is created to avoid passing all the RFunctions to the sub dialog individually.
-    Private dctSummaries As Dictionary(Of String, Tuple(Of RFunction, RFunction))
+    Private dctComparisonSummaries As Dictionary(Of String, Tuple(Of RFunction, RFunction))
 
     ' ggplot functions
     Private clsGgplotOperator As ROperator
@@ -113,9 +125,18 @@ Public Class dlgTimeSeriesPlot
     Private clsGeomHLineParameter As RParameter
     Private clsGeomHLine As RFunction
     Private clsGeomHLineAes As RFunction
-    Private clsGeomText As RFunction
-    Private clsGeomTextAes As RFunction
-    Private clsPasteLabel As RFunction
+    Private clsComparisonGeomText As RFunction
+    Private clsComparisonGeomTextAes As RFunction
+    Private clsComparisonPasteLabel As RFunction
+
+    Private clsReferenceGeomText As RFunction
+    Private clsReferenceGeomTextAes As RFunction
+    Private clsReferencePasteLabel As RFunction
+    Private clsReferenceFilter As RFunction
+    Private clsEstimatesGeomText As RFunction
+    Private clsEstimatesGeomTextAes As RFunction
+    Private clsEstimatesPasteLabel As RFunction
+    Private clsEstimatesFilter As RFunction
 
     Private clsLabsFunction As New RFunction
     Private clsXlabsFunction As New RFunction
@@ -219,15 +240,22 @@ Public Class dlgTimeSeriesPlot
         clsPivotLonger = New RFunction
         clsPivotCFunction = New RFunction
 
-        clsMeansOperator = New ROperator
-        clsMeansGroupBy = New RFunction
-        clsMeansSummarise = New RFunction
-        clsMeanFunction = New RFunction
+        clsIndividualSummariesOperator = New ROperator
+        clsIndividualSummariesGroupBy = New RFunction
+        clsIndividualSummariesSummarise = New RFunction
+        clsNIndividual = New RFunction
+        clsMean = New RFunction
+        clsSd = New RFunction
+        clsSlope = New RFunction
+        clsPasteNIndividual = New RFunction
+        clsPasteMean = New RFunction
+        clsPasteSd = New RFunction
+        clsPasteSlope = New RFunction
 
         clsSummaryOperator = New ROperator
         clsSummaryGroupBy = New RFunction
         clsSummarise = New RFunction
-        clsN = New RFunction
+        clsNComparison = New RFunction
         clsCor = New RFunction
         clsPbias = New RFunction
         clsRSd = New RFunction
@@ -251,15 +279,15 @@ Public Class dlgTimeSeriesPlot
         clsSignifRmse = New RFunction
         clsRoundKge = New RFunction
 
-        dctSummaries = New Dictionary(Of String, Tuple(Of RFunction, RFunction))
-        dctSummaries.Add("n", New Tuple(Of RFunction, RFunction)(clsN, clsPasteN))
-        dctSummaries.Add("cor", New Tuple(Of RFunction, RFunction)(clsCor, clsPasteCor))
-        dctSummaries.Add("pbias", New Tuple(Of RFunction, RFunction)(clsPbias, clsPastePbias))
-        dctSummaries.Add("rsd", New Tuple(Of RFunction, RFunction)(clsRSd, clsPasteRSd))
-        dctSummaries.Add("me", New Tuple(Of RFunction, RFunction)(clsMe, clsPasteMe))
-        dctSummaries.Add("mae", New Tuple(Of RFunction, RFunction)(clsMae, clsPasteMae))
-        dctSummaries.Add("rmse", New Tuple(Of RFunction, RFunction)(clsRmse, clsPasteRmse))
-        dctSummaries.Add("kge", New Tuple(Of RFunction, RFunction)(clsKge, clsPasteKge))
+        dctComparisonSummaries = New Dictionary(Of String, Tuple(Of RFunction, RFunction))
+        dctComparisonSummaries.Add("n", New Tuple(Of RFunction, RFunction)(clsNComparison, clsPasteN))
+        dctComparisonSummaries.Add("cor", New Tuple(Of RFunction, RFunction)(clsCor, clsPasteCor))
+        dctComparisonSummaries.Add("pbias", New Tuple(Of RFunction, RFunction)(clsPbias, clsPastePbias))
+        dctComparisonSummaries.Add("rsd", New Tuple(Of RFunction, RFunction)(clsRSd, clsPasteRSd))
+        dctComparisonSummaries.Add("me", New Tuple(Of RFunction, RFunction)(clsMe, clsPasteMe))
+        dctComparisonSummaries.Add("mae", New Tuple(Of RFunction, RFunction)(clsMae, clsPasteMae))
+        dctComparisonSummaries.Add("rmse", New Tuple(Of RFunction, RFunction)(clsRmse, clsPasteRmse))
+        dctComparisonSummaries.Add("kge", New Tuple(Of RFunction, RFunction)(clsKge, clsPasteKge))
 
         clsGgplotOperator = New ROperator
         clsGgplotFunction = New RFunction
@@ -268,9 +296,18 @@ Public Class dlgTimeSeriesPlot
         clsGeomPoint.Clear()
         clsGeomHLine.Clear()
         clsGeomHLineAes.Clear()
-        clsGeomText = New RFunction
-        clsGeomTextAes = New RFunction
-        clsPasteLabel = New RFunction
+        clsComparisonGeomText = New RFunction
+        clsComparisonGeomTextAes = New RFunction
+        clsComparisonPasteLabel = New RFunction
+
+        clsReferenceGeomText = New RFunction
+        clsReferenceGeomTextAes = New RFunction
+        clsReferencePasteLabel = New RFunction
+        clsReferenceFilter = New RFunction
+        clsEstimatesGeomText = New RFunction
+        clsEstimatesGeomTextAes = New RFunction
+        clsEstimatesPasteLabel = New RFunction
+        clsEstimatesFilter = New RFunction
 
         ucrSelectorTimeSeriesPlots.Reset()
         ucrSelectorTimeSeriesPlots.SetGgplotFunction(clsGgplotOperator)
@@ -314,23 +351,33 @@ Public Class dlgTimeSeriesPlot
 
         ' Calculate means
 
-        clsMeansOperator.SetOperation("%>%")
-        clsMeansOperator.AddParameter("0", clsROperatorParameter:=clsStackOperator, iPosition:=0)
-        clsMeansOperator.AddParameter("1", clsRFunctionParameter:=clsMeansGroupBy, iPosition:=1)
-        clsMeansOperator.AddParameter("2", clsRFunctionParameter:=clsMeansSummarise, iPosition:=2)
-        clsMeansOperator.bBrackets = False
+        clsIndividualSummariesOperator.SetOperation("%>%")
+        clsIndividualSummariesOperator.AddParameter("0", clsROperatorParameter:=clsStackOperator, iPosition:=0)
+        clsIndividualSummariesOperator.AddParameter("1", clsRFunctionParameter:=clsIndividualSummariesGroupBy, iPosition:=1)
+        clsIndividualSummariesOperator.AddParameter("2", clsRFunctionParameter:=clsIndividualSummariesSummarise, iPosition:=2)
+        clsIndividualSummariesOperator.bBrackets = False
 
-        clsMeansGroupBy.SetPackageName("dplyr")
-        clsMeansGroupBy.SetRCommand("group_by")
-        clsMeansGroupBy.AddParameter("1", strData, iPosition:=1, bIncludeArgumentName:=False)
+        clsIndividualSummariesGroupBy.SetPackageName("dplyr")
+        clsIndividualSummariesGroupBy.SetRCommand("group_by")
+        clsIndividualSummariesGroupBy.AddParameter("1", strData, iPosition:=1, bIncludeArgumentName:=False)
 
-        clsMeansSummarise.SetPackageName("dplyr")
-        clsMeansSummarise.SetRCommand("summarise")
-        clsMeansSummarise.AddParameter("mean", clsRFunctionParameter:=clsMeanFunction, iPosition:=0)
+        clsIndividualSummariesSummarise.SetPackageName("dplyr")
+        clsIndividualSummariesSummarise.SetRCommand("summarise")
+        clsIndividualSummariesSummarise.AddParameter("mean", clsRFunctionParameter:=clsMean, iPosition:=0)
 
-        clsMeanFunction.SetRCommand("mean")
-        clsMeanFunction.AddParameter("x", strValue, iPosition:=0)
-        clsMeanFunction.AddParameter("na.rm", "TRUE", iPosition:=1)
+        clsNIndividual.SetPackageName("dplyr")
+        clsNIndividual.SetRCommand("n")
+
+        clsMean.SetRCommand("mean")
+        clsMean.AddParameter("x", strValue, iPosition:=0)
+        clsMean.AddParameter("na.rm", "TRUE", iPosition:=1)
+
+        clsSd.SetRCommand("sd")
+        clsSd.AddParameter("x", strValue, iPosition:=0)
+        clsSd.AddParameter("na.rm", "TRUE", iPosition:=1)
+
+        clsSlope.SetRCommand("mean")
+        clsSlope.AddParameter("y", strValue, iPosition:=0)
 
         ' Calculate summaries
 
@@ -345,8 +392,8 @@ Public Class dlgTimeSeriesPlot
         clsSummarise.SetPackageName("dplyr")
         clsSummarise.SetRCommand("summarise")
 
-        clsN.SetPackageName("dplyr")
-        clsN.SetRCommand("n")
+        clsNComparison.SetPackageName("dplyr")
+        clsNComparison.SetRCommand("n")
 
         clsCor.SetRCommand("cor")
         clsCor.AddParameter("use", Chr(34) & "na.or.complete" & Chr(34))
@@ -395,7 +442,7 @@ Public Class dlgTimeSeriesPlot
 
         clsGeomHLine.SetPackageName("ggplot2")
         clsGeomHLine.SetRCommand("geom_hline")
-        clsGeomHLine.AddParameter("data", clsROperatorParameter:=clsMeansOperator, iPosition:=0)
+        clsGeomHLine.AddParameter("data", clsROperatorParameter:=clsIndividualSummariesOperator, iPosition:=0)
         clsGeomHLine.AddParameter("mapping", clsRFunctionParameter:=clsGeomHLineAes, iPosition:=1)
 
         clsGeomHLineAes.SetPackageName("ggplot2")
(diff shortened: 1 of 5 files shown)


dannyparsons:

rdstern:
This is working well so far

1) I suggest moving the Summary Annotations button to the right-hand side.  That's where it is for the Summaries dialogue. (No longer - I suggest instead a Time Plot Options sub-dialogue and they can then go there.
2) Add a button, as for the ordinary line plots dialogue, for the trend line.  The default for that is un-ticked.
3) Change the default for the mean line to be un-ticked.

4) The summaries works well, except the default is large.  I changed to 3.  When I used all 8 they are not all visible, even with size 3.  I wonder about another option of making them the plot footnote.
5) As you have the means - and I would also like the slopes - could there be another set of summaries, namely Means, Slopes.  They each give 2 values if chosen.  Perhaps they are only enabled if the relevant option is chosen, but they could be of interest anyway.
6) Do you intend to add a Time Plot Options sub-dialogue?  If so, then perhaps this could include the annotations and that button would then go in the usual place for the plotting sub-dialogues.  One option I would like is the presence/absence and position of the legend.  And I would like to be able to specify, or omit, the x-axis labels as for the PICSA rainfall.  In particular, when it is Date or year, then it is obvious, so I don't need an x-axis label. 
7) I tried with, and then without, your "Remove rows with any missing values" and couldn't see any difference in the results?  When it is without, then I expect 2 values of n and some statistics, like the bias, and the means should change?I think maybe it is different on the graphs, but not on the summaries?  There is an important statistical (and ENACTS) point here. 

The only bug I have found so far, was when I analysed rainfall data - still the daily data - which worked fine.  Then I changed to tmax and tmin and it gave an error.  I didn't copy the error - and can't get it again.  I give the R-commands that gave the error, blank plot - and the later one, after reset - that worked.

This one had the error:
```
# Code generated by the dialog, Time Series Comparisons
data_RDS <- data_book$get_data_frame(data_name="data_RDS")
data_RDS <- data_RDS %>% dplyr::mutate(tmax=ifelse(test=is.na(x=tmin), yes=NA, no=tmax), tmin=ifelse(test=is.na(x=tmax), yes=NA, no=tmin), rain=ifelse(test=is.na(x=tmin), yes=NA, no=tmax), enacts_rain=ifelse(test=is.na(x=tmax), yes=NA, no=tmin), chirps_rain=ifelse(test=is.na(x=tmax), yes=NA, no=tmin))
data_RDS_stack <- data_RDS %>% tidyr::pivot_longer(cols=c(tmax, tmin), names_to="data", values_to="value")
data_RDS_summary <- data_RDS %>% dplyr::group_by(station) %>% dplyr::summarise(n=dplyr::n(), cor=cor(x=tmin, y=tmax, use="na.or.complete"), pbias=hydroGOF::pbias(sim=tmin, obs=tmax), rsd=hydroGOF::rSD(sim=tmin, obs=tmax))
last_graph <- ggplot2::ggplot(data=data_RDS_stack, mapping=ggplot2::aes(x=date, y=value, colour=data)) + ggplot2::geom_line() + ggplot2::geom_point() + ggplot2::geom_text(data=data_RDS_summary, mapping=ggplot2::aes(label=paste(n=paste("n", n), cor=paste("cor", round(x=cor, digits=2)), pbias=paste("% Bias", round(x=pbias, digits=2)), rsd=paste("rSD", round(x=rsd, digits=2)), sep=" ; ")), x=-Inf, y=Inf, hjust=0, vjust=1, size=3, inherit.aes=FALSE) + theme_grey() + ggplot2::facet_wrap(facets= ~ station)
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="data_RDS")
data_book$get_graphs(data_name="data_RDS", graph_name="last_graph")

```
And this one worked ok:
```
# Code generated by the dialog, Time Series Comparisons
data_RDS <- data_book$get_data_frame(data_name="data_RDS")
data_RDS <- data_RDS %>% dplyr::mutate(tmax=ifelse(test=is.na(x=tmin), yes=NA, no=tmax), tmin=ifelse(test=is.na(x=tmax), yes=NA, no=tmin))
data_RDS_stack <- data_RDS %>% tidyr::pivot_longer(cols=c(tmax, tmin), names_to="data", values_to="value")
data_RDS_mean <- data_RDS_stack %>% dplyr::group_by(station, data) %>% dplyr::summarise(mean=mean(x=value, na.rm=TRUE))
last_graph <- ggplot2::ggplot(data=data_RDS_stack, mapping=ggplot2::aes(x=date, y=value, colour=data)) + ggplot2::geom_line() + ggplot2::geom_point() + ggplot2::geom_hline(data=data_RDS_mean, mapping=ggplot2::aes(yintercept=mean, colour=data)) + theme_grey() + ggplot2::facet_wrap(facets= ~ station)
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="data_RDS")
data_book$get_graphs(data_name="data_RDS", graph_name="last_graph")
```

instat/sdgTimeSeries.vb
```diff
@@ -203,4 +215,34 @@ Public Class sdgTimeSeries
...
+                clsGeomText.AddParameter("x", "-Inf", iPosition:=3)
+                clsGeomText.AddParameter("y", "-Inf", iPosition:=4)
+                clsGeomText.AddParameter("hjust", "0", iPosition:=5)
+                clsGeomText.AddParameter("vjust", "0", iPosition:=6)
+            ElseIf ucrInputPosition.GetText() = strBottomRight Then
+                clsGeomText.AddParameter("x", "Inf", iPosition:=3)
+                clsGeomText.AddParameter("y", "-Inf", iPosition:=4)
+                clsGeomText.AddParameter("hjust", "1", iPosition:=5)
+                clsGeomText.AddParameter("vjust", "0", iPosition:=6)
+            End If
```
shadrackkibet, line 245: Will this be much clearer with `select`?

dannyparsons:
> Add a button, as for the ordinary line plots dialogue, for the trend line. The default for that is un-ticked.

I assume you mean checkbox?

> Change the default for the mean line to be un-ticked.

I quite like it there by default as it very quickly gives you an idea of bias and it doesn't seem too distracting (unless you add trend and then probably remove it). Why didn't you like it checked by default?

> The summaries works well, except the default is large. I changed to 3. When I used all 8 they are not all visible, even with size 3. I wonder about another option of making them the plot footnote.

Don't mind changing the default size to be a bit smaller. But this was based on having 3 or 4 summaries. I can't see it being sensible to have all 8, one because its too much text but also some of them give the same basic information and are alternatives to each other.

Footnotes (caption in ggplot) is just for the whole graph I think and not for each facet so that wouldn't work for this.

> As you have the means - and I would also like the slopes - could there be another set of summaries, namely Means, Slopes. They each give 2 values if chosen. Perhaps they are only enabled if the relevant option is chosen, but they could be of interest anyway.

I'll need to think about how to incorporate this. I don't think it would be sensible to display the comparison summaries and the individual summaries at the same time without doing careful manual positioning which we can't accommodate on the dialog. So my initial idea is these are two different options on the summary sub dialog - comparison or individual summaries. When there is a slope the best place for the two sets are on diagonally opposite each other so that would probably be the options for this.

> Do you intend to add a Time Plot Options sub-dialogue? If so, then perhaps this could include the annotations and that button would then go in the usual place for the plotting sub-dialogues. One option I would like is the presence/absence and position of the legend. And I would like to be able to specify, or omit, the x-axis labels as for the PICSA rainfall. In particular, when it is Date or year, then it is obvious, so I don't need an x-axis label.

I think I would suggest trying to get the Plot Options subdialog to work with the dialog instead of a new subdialog. A specific sub dialog could become a never ending task and never quite have all the options everyone wants. Unlike for the PICSA graphs, I think it's fair to expect users to learn how to use the general Plot Options subdialog to add extra options.

> I tried with, and then without, your "Remove rows with any missing values" and couldn't see any difference in the results? When it is without, then I expect 2 values of n and some statistics, like the bias, and the means should change?I think maybe it is different on the graphs, but not on the summaries? There is an important statistical (and ENACTS) point here.

The statistics don't change because the comparison statistics can only use complete rows anyway. If there's an `NA` in the reference data and not for the estimates then the correlation is the same with or without that row because it's always ignored. What changes is what's plotted when those rows are removed - and the mean lines would also change, but comparison summaries are unaffected by removing rows with a missing value - the `n` summary refers to the number of points used in the comparisons, so this doesn't change either. Maybe different text for the checkbox would make this clearer but I couldn't think of anything better that is as short as this.  Of course if we do include individual summaries as annotations such as mean then these would change.

instat/sdgTimeSeries.vb
dannyparsons: thanks, done

dannyparsons:
> The only bug I have found so far, was when I analysed rainfall data - still the daily data - which worked fine. Then I changed to tmax and tmin and it gave an error. I didn't copy the error - and can't get it again. I give the R-commands that gave the error, blank plot - and the later one, after reset - that worked.

I reran your code and there is an error in calculating the summaries with `hydroGOF`. I think this is because Moorings has no temperature data and `hydroGOF` can't cope with calculations on no data. When I filter out Moorings the code runs ok so pretty sure that's it. 
This could cause us other problems when doing summaries with `hydroGOF` so we might want to try do something about this.

Patowhiz:
@dannyparsons please can you add comments to the code. 

It may seem trivial to have comments on some of the functions, subroutines and choice of current implementation but it helps a lot in future especially if another developer adds implementation. Sometimes as a developer it helps a lot to understand with surety the current choice of implementation. Changing it or adding to it requires absolute understanding of what is currently implemented, otherwise this may lead to redundant additions and possibly bugs to the code in future.

Commenting saves a lot of time to other developers and even helps current reviewers to understand more about the current implementation, improving reviews and reducing bugs that may pass.

I would also generally encourage all team members to be commenting on their implementation especially where it may not seem so obvious on the choice of implementation. It may seem time consuming and unnecessary but it is one of the good tenets of good programming and software development.

dannyparsons:
Happy to do that. Do we have an agreement on the form of comments? Are we using the inline documentation for functions? Are we just using this or also adding comments as well? Do we have an agreement on what parts of dialog code we put comments in general? I have some thoughts but interested what is already happening on this @Patowhiz @lloyddewit

Patowhiz:
@lloyddweit has been helping me in improving on the sections to comment. He can explain better.

We have been using inline documentation for functions. Explain what the function does and it's parameters. 
Also, we have comments on global variables or properties  of the dialog, briefly explaining what they are used for.

In addition to this, I'm usually keen on commenting my "choice" of Implementation especially when it deviates from the standard norm of doing things in R-Instat e.g manually checking for R-Parameters. And possibly pointing on the issue that explains the problem further. 

Sometimes due to limitations of the "base code", I would also normally comment on my choice of implementation and how it could be made better once the "base code" is improved.

dannyparsons:
That sounds reasonable. In general I would suggest documenting core parts of the code and code which is non standard or non obvious. I think that is a reasonably general view in software.

So in dialogs I would only comment things which are both not common to dialogs in general and might not be clear to someone looking at it for the first time. But the standard stuff and non standard stuff that is obvious what it's doing and why I don't think should be cluttered with comments.

shadrackkibet:
I like @dannyparsons suggestions above on comments. I like to reserve comments to things that are not standard in the way we code. Although that is subjective, probably we need examples or define clearly what is not obvious to us. Commenting on everything is not a good idea to me. I would like to encourage programmers to make code more readable rather than using comments at the expense of making code readable. We also need to be careful not to put in misleading comments. There is also a potential problem if we are to adopt this in general which will mean we need to maintain comments i.e update comments when code changes which tend to be forgotten and therefore with time some comments will be obsolete.

dannyparsons:
@Patowhiz @lloyddewit @shadrackkibet have a look at the comments I added to the main dialog only. I have commented global variables and non-standard parts of the code. Does this align with what you expect to be commented on a dialog?

instat/dlgTimeSeriesPlot.vb
```diff
@@ -1,3 +1,633 @@
...
+        If ucrReceiverEstimates.IsEmpty Then
+            If strEstimates <> "" Then
+                clsAdjustNAMutate.RemoveParameterByName(strEstimates)
+                strEstimates = ""
+            End If
+        Else
+            strEstimates = ucrReceiverEstimates.GetVariableNames(False)
+            clsAdjustNAMutate.AddParameter(strEstimates, clsRFunctionParameter:=clsIfElseEstimates, iPosition:=1)
+        End If
+    End Sub
```
Patowhiz, line 715: @dannyparsons, my understanding is `strEstimates` is used in these subroutine only, to store the parameter name from `ucrReceiverEstimates`. Which is added or removed from  `clsAdjustNAMutate`.  And that the parameter name could be different but the instance reference of `clsIfElseEstimates` doesn't change (unless its on reset).

Being in this subroutine only, means that this happens every time there's value changes in the `ucrReceiverEstimates`, not anywhere else. 
If so, I wonder if the following alternatives could work.

```
 If ucrReceiverEstimates.IsEmpty Then
      clsAdjustNAMutate.RemoveParameterByName(strEstimates)
      strEstimates = ""
Else
     strEstimates = ucrReceiverEstimates.GetVariableNames(False)
     clsAdjustNAMutate.AddParameter(strEstimates, clsRFunctionParameter:=clsIfElseEstimates, iPosition:=1)
End If
```

or (I think this could be preferable especially because it removes the need of having the extra variable for holding the parameter name)

```
 If ucrReceiverEstimates.IsEmpty Then
      clsAdjustNAMutate.RemoveParameter(clsIfElseEstimates)
Else
     clsAdjustNAMutate.AddParameter(ucrReceiverEstimates.GetVariableNames(False), 
                                                             clsRFunctionParameter:=clsIfElseEstimates, iPosition:=1)
End If
```

If the above is true, I think the similar implementation could be done for `ucrReceiverReference.ControlValueChanged`

instat/sdgTimeSeries.vb
```diff
@@ -1,3 +1,282 @@
...
+' You should have received a copy of the GNU General Public License 
+' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports instat.Translations
+
+Public Class sdgTimeSeries
+    Private bRCodeSet As Boolean = False
+    Private bControlsInitialised As Boolean = False
+
+    Private strGeomTextParameterName As String = "geom_text"
```
Patowhiz, line 23: I've noticed you used `strGeomTextParameterName` variable to hold the parameter name for `clsGgplotOperator` , but you haven't done the same for `clsSummarise` and `clsPasteLabel`. My understanding is you probably found no substantial merit to have an explicit variable to hold parameter names for `clsSummarise` and `clsPasteLabel` because its being set in one subroutine. I wonder if the same could be done for `clsGgplotOperator`, removing the need to have this extra variable in this subdialog?

instat/dlgTimeSeriesPlot.vb
```diff
@@ -1,3 +1,763 @@
...
+    Private bReset As Boolean = True
+    Private bResetSubdialog As Boolean = True
+
+    ' Constants
+    ' These store the names of the selected columns. This is used to ensure old parameters are correctly removed from dplyr::mutate when selected columns are changed.
+    ' See ucrReceiverEstimates.ControlValueChanged and ucrReceiverReference.ControlValueChanged event handles for their use.
+    Private strEstimates As String = ""
+    Private strReference As String = ""
+    ' The name of the "names" column when the data is stacked
+    Private strData As String = "data"
```
Patowhiz, line 31: From the comment
 
`' The name of the "names" column when the data is stacked`

wouldn't `strName` be preferable? to relatively match the comment.

Patowhiz:
@dannyparsons the comments in `dlgTimeSeriesPlot` were helpful, they align with what I expected to be commented on the dialog.

Patowhiz:
@dannyparsons I'm unable to see the issue linked to this PR. So it wasn't completely clear to me in terms of what to test other than checking on if the dialog works without bugs.

instat/sdgTimeSeries.vb
dannyparsons: These variables are needed for geoms when the dialog is linked to the Plot Options subdialog as the list of dialog geoms is passed to the set up of the Plot Options sub dialog. Although that's not on this dialog yet the intention is to have it added at some point so when this added these variables will be needed. Currently there is no other reason to have them, which is why they are not there for non-geom parameters.

instat/dlgTimeSeriesPlot.vb
dannyparsons: Yes, makes sense.

dannyparsons:
Sorry I don't think we have an issue which explains the idea behind the dialog. Briefly, the main idea is to compare the time series of two columns e.g. station and satellite data to see how similar they are. So it is really just a line plot, but the options allow you to add a line for the mean (or trend that I'm about to add). The sub dialog allows to display comparison statistics which gives some numerical measures of how similar the two columns are.

This might mainly be used with summary data, such as yearly data of total rainfall, number of rain days, average temperature, but could also be done on more frequent data, such as daily temperature data.

instat/dlgTimeSeriesPlot.vb
dannyparsons: Your first suggestion is just removing `If strEstimates <> ""` I think? Seems pretty much the same either way to me since if `strEstimates = ""` then nothing will happen anyway. In case a parameter did get added with name `""` somehow then your suggestion is probably slightly safer so I'll remove the `If strEstimates <> ""`.

Your second suggestion wouldn't work since `clsIfElseEstimates` is not an `RParameter` so it's not accepted by `.RemoveParameter()`. To use that function would require defining an `RParameter` for this. Then this part of the code would just be changing the name of the parameter. But then it's not really any simpler than this because we need another variable to define the `RParameter` so it's much the same as have a string variable.

dannyparsons:
> @dannyparsons I'm unable to see the issue linked to this PR. So it wasn't completely clear to me in terms of what to test other than checking on if the dialog works without bugs.

Great. I suggest we could have something like a wiki page which explains our expectations for code comments and in line documentation.

dannyparsons:
Also, I am making some more substantial changes to this dialog today from Roger's comments so I suggest anyone wanting to review from now waits until the new changes are done.

dannyparsons:
@africanmathsinitiative/developers this is now ready for review again

shadrackkibet:
I make minor design comments;
- Did you miss colons? I think that is because the same label is used in two places.

image
image

- Save control label is far from the input box.
image

- Should OK be enabled when you don't have Time (X- axis)? I got an error when that is empty because of missing x aesthetic.
image

rdstern (dismissed):
@dannyparsons this seems to work fine and looks excellent.  I am happy fpr it to be merged as is, if the code is ok according to @lloyddewit 

I did (of course) try odd things to check.  It still works fine if the two variables are the same.  It isn't a bad graph for a single line for multiple stations.  Then, obviously the comparisons don't work, but the summaries do.  

Later I understand that the Plot sub-dialogue may be added.  That would be great, but could become a later issue.

shadrackkibet:
I notice that when I check `N` on individual summaries then I return back to the dialog, when I go back I find all other individual summaries `mean, sd, slope`are checked. Sometimes when I check `mean,sd,slope` on return they are unchecked.

dannyparsons:
Thanks. @shadrackkibet I've fixed the designer issues and sub dialog checkbox issues, please check and confirm.

shadrackkibet:
Other issues have been addressed except for TestOk when "Time(X -axs)" is empty if you agree.

dannyparsons:
Thanks, forgot that, that's done now.

shadrackkibet:
`geom_text` bugs when font size is left empty, I think Ok should be disabled? Not much of a concern though, it might confuse users when they find that Ok is not enabled.  Otherwise, it looks good for me.

dannyparsons:
That's a good spot. Disabling Ok could also be confusing for a user to know they need to go back to the sub dialog and set the font size. Instead of this I have set the sub dialog to set the font size to the default value if it is empty when the sub dialog is closing. This then avoids messages, errors or confusion in case it is left empty.

rdstern (approved):
Here is an example:
image
a) It is working well now.
b) There remain quite a lot of improvements I would like still.  They include being able to get rid of the x-axis label, changing or deleting the legend, etc.  They are all consistent with something I know that @dannyparsons is planning to try soon, i.e. to be able to access the usual Plot Options sub-dialogue.
c) So depends on him.  I have approved in case we merge now, and add this later.  Or the button could be added and disabled for now - so it shows the intention.  Or it could be added and tested now.

shadrackkibet:
Can something be done on the scientific notation for the slope?

image

dannyparsons:
That's more of a global option in R which controls when it chooses to display something in scientific notation. We could force it by instead using `format()` and `scientific = FALSE`. But then it's forced as never being scientific and could be very long and unreadable, which might not be what some people want. So we would have to add an option to control `scientific` on the subdialog which I think gets too detailed. I think for numeric output this is more important but for display on a graph scientific is ok to know it won't be too long, and if it's that close to 0 anyway it shouldn't be much of an issue.

dannyparsons:
@rdstern I think I would prefer this to be merged now and the Plot Options added later as I'm not sure exactly when I'll get to that so the dialog could be merged and used in the meantime.

rdstern (dismissed):
Still approved.  Over to @lloyddewit

lloyddewit:
@dannyparsons @Patowhiz Thank you for all the comments on comments. I found them sensible and realistic. Sorry for not responding sooner.

One additional observation. @dannyparsons wrote:

> So in dialogs I would only comment things which are both not common to dialogs in general and might not be clear to someone looking at it for the first time. But the standard stuff and non standard stuff that is obvious what it's doing and why I don't think should be cluttered with comments.

The 'standard stuff' is self-explantory for the experienced developers. New developers will need to invest time understanding this. The developers guide helps. I suggest that we nominate a dialog to be a model dialog. This should be a dialog that is clean, well-written, and contains examples of typical dialog behaviour. The developer's guide could reference this dialog.

The dialog and sub-dialog in this PR could be candidates because they have been written by the lead developer and are new (so still have a clean structure). We don't have to add the comments in this PR but if you agree then I could create an issue for this.

I also have some separate specific suggestions about comments in this PR but I'll include them as part of the normal peer review.

instat/dlgTimeSeriesPlot.vb
```diff
@@ -1,3 +1,799 @@
...
+
+    ' Constants
+    ' These store the names of the selected columns. This is used to ensure old parameters are correctly removed from dplyr::mutate when selected columns are changed.
+    ' See ucrReceiverEstimates.ControlValueChanged and ucrReceiverReference.ControlValueChanged event handles for their use.
+    Private strEstimates As String = ""
+    Private strReference As String = ""
+    ' The name of the "names" column when the data is stacked
+    Private strName As String = "data"
+    ' The name of the "values" column when the data is stacked
+    Private strValue As String = "value"
```
lloyddewit, line 33: Declare as `Const`?

instat/sdgTimeSeries.vb
```diff
@@ -1,3 +1,440 @@
...
+    Private bControlsInitialised As Boolean = False
+
+    Private strGeomTextReferenceParameterName As String = "reference_geom_text"
+    Private strGeomTextEstimatesParameterName As String = "estimates_geom_text"
+    Private strGeomTextComparisonParameterName As String = "comparison_geom_text"
+
+    Private strTopLeft As String = "Top Left"
+    Private strTopRight As String = "Top Right"
+    Private strBottomLeft As String = "Bottom Left"
+    Private strBottomRight As String = "Bottom Right"
```
lloyddewit, line 31: are these constants? declare as `Const`?

lloyddewit (changes requested):
@dannyparsons  thanks, I found this very clean and readable.

dannyparsons:
I like the idea of a model dialog fully commented for new developers. This dialog is quite a bit more complicated compared to standard dialogs, so I'm not sure if that makes it a good candidate or not. Or we might want a couple of fully commented dialogs. One completely standard dialog and one or two with some more complexity. I agree we should move this conversation over to an issue.

instat/dlgTimeSeriesPlot.vb
```diff
@@ -1,3 +1,799 @@
...
+
+    Public bFirstLoad As Boolean = True
+    Private bReset As Boolean = True
+    Private bResetSubdialog As Boolean = True
+
+    ' Constants
+    ' These store the names of the selected columns. This is used to ensure old parameters are correctly removed from dplyr::mutate when selected columns are changed.
+    ' See ucrReceiverEstimates.ControlValueChanged and ucrReceiverReference.ControlValueChanged event handles for their use.
+    Private strEstimates As String = ""
+    Private strReference As String = ""
```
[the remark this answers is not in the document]
dannyparsons: Agreed.

instat/dlgTimeSeriesPlot.vb
```diff
@@ -1,3 +1,799 @@
...
+    ' The name of the "names" column when the data is stacked
```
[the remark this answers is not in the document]
dannyparsons: Good idea.

instat/dlgTimeSeriesPlot.vb
```diff
@@ -1,3 +1,799 @@
...
+    Private strName As String = "data"
+    ' The name of the "values" column when the data is stacked
+    Private strValue As String = "value"
+
+    ' Adjust NA values
+    ' These functions construct the R code below if ucrChkNAValues is checked. This ensures the two columns have the same missing values.
+    ' df <- df %>% 
+    '   mutate(gauge = ifelse(is.na(estimate), NA, gauge),
+    '          estimate = ifelse(is.na(gauge), NA, estimate))
```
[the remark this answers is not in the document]
dannyparsons: Maybe if there was a way for some of these comments to be part of the documentation for the whole dialog itself, if that's possible.

instat/dlgTimeSeriesPlot.vb
```diff
@@ -1,3 +1,799 @@
...
+    Private clsRoundPbias As RFunction
+    Private clsRoundRSd As RFunction
+    Private clsRoundKge As RFunction
+
+    ' A dictionary of the comparison/individual summary functions above. 
+    ' The key is the summary name in lower case e.g. "bias"
+    ' The value is a tuple of two RFunctions. 
+    ' Item1 is the summary function. 
+    ' Item2 is the paste function containing the summary.
+    ' These dictionary are created to avoid passing all the RFunctions to the sub dialog individually.
```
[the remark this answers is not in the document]
dannyparsons: Done. I'll leave them as members of the dialog in case they may be needed in other parts of the dialog later.

instat/dlgTimeSeriesPlot.vb
```diff
@@ -1,3 +1,799 @@
...
+        clsReferenceGeomText = New RFunction
+        clsReferenceGeomTextAes = New RFunction
+        clsReferencePasteLabel = New RFunction
+        clsReferenceFilter = New RFunction
+        clsEstimatesGeomText = New RFunction
+        clsEstimatesGeomTextAes = New RFunction
+        clsEstimatesPasteLabel = New RFunction
+        clsEstimatesFilter = New RFunction
+
+        dctComparisonSummaries = New Dictionary(Of String, Tuple(Of RFunction, RFunction))
```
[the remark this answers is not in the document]
dannyparsons: Because the functions within the tuple are reinitialised in `SetDefaults()` e.g. `clsNComparison = New RFunction` that breaks the reference link between the `RFunction` and the `RFunction` in the `Dictionary` hence why I reinitialise the Dictionary too.
Not sure I have a great reason for keeping the Dictionaries here instead of within `cmdSummaries_Click` but feels sensible to have them initialised in a similar place to the `RFunctions` and only done when resetting the dialog not every time the sub dialog is opened.

instat/dlgTimeSeriesPlot.vb
```diff
@@ -1,3 +1,799 @@
...
+    Private clsEstimatesGeomText As RFunction
+    Private clsEstimatesGeomTextAes As RFunction
+    Private clsEstimatesPasteLabel As RFunction
+    Private clsEstimatesFilter As RFunction
+
+    Private clsLabsFunction As New RFunction
+    Private clsXlabsFunction As New RFunction
+    Private clsYlabFunction As New RFunction
+    Private clsXScaleContinuousFunction As New RFunction
+    Private clsYScaleContinuousFunction As New RFunction
```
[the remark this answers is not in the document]
dannyparsons: These will be needed when we have the Plot Options subdialog implemented on the dialog, but currently they are not used. I have commented them out with a comment why.

instat/dlgTimeSeriesPlot.vb
dannyparsons: Yes.

instat/sdgTimeSeries.vb
dannyparsons: Yes.

dannyparsons:
Thanks @lloyddewit I have tried to address your comments and pushed changes with the suggestions.

instat/dlgTimeSeriesPlot.vb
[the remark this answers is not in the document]
lloyddewit: yes, really each class should also have an XML comment block. But I suggest we don't block the merge of this PR for this.
Here's an example from from `RCodeStructure`:
```
''' <summary>   An object of this class represents an R command. 
'''             The R command may include parameters and an assignment part.
'''             <para>
'''             This class provides methods and fields to:</para>
''' <list type="bullet">
'''     <item><description>
'''             Insert, update and delete the R command's parameters.
'''     </description></item><item><description>
'''             Change the relative position of the R command's parameters.
'''     </description></item><item><description>
'''             Define what to assign to (e.g. data frame, column, model etc.).
'''     </description></item><item><description>
'''             Whether the assignment part has been created (or needs to be created).
'''     </description></item><item><description>
'''             Define what to do with the result returned from the R command (e.g. ignore, show 
'''             the result as text, show the result as a graph or store in a temporary variable).
'''     </description></item><item><description>
'''             Return the R command as an executable R string.
'''     </description></item>
''' </list>
'''             </summary>
Public Class RCodeStructure
```

instat/dlgTimeSeriesPlot.vb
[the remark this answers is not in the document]
lloyddewit: Understood, thanks for the explanation

lloyddewit (approved):

lloyddewit:
@rdstern Please could you re-approve (changes since your previous approval mainly affected comments so they are low risk)? thanks

lloyddewit:
> 
> 
> I like the idea of a model dialog fully commented for new developers. This dialog is quite a bit more complicated compared to standard dialogs, so I'm not sure if that makes it a good candidate or not. Or we might want a couple of fully commented dialogs. One completely standard dialog and one or two with some more complexity. I agree we should move this conversation over to an issue.

I moved this conversation to issue #3577

rdstern (approved):

PR #5987, opened 2020-09-18
Updating dplyr functions

lilyclements:
@dannyparsons here is the update of the `group_by` and `arrange` functions so far


lloyddewit:
@lilyclements is this ready for review or do you expect to make more commits? thanks

lilyclements:
@lloyddewit I am still working on this.

Current state:
* dplyr parameters are mostly updated except for a case of `group_by` in a for-loop that needs looking at
* need to add `.drop` and `.preserve` parameters for the `group_by` and `filter` functions respectively.

lilyclements:
@dannyparsons there are three things here I would like to discuss when you have the time:

* I've used `group_by` in the for-loop but under the `parse` system. I'm unsure how we can do this with the embrace system.
* Unsure how to add the `.drop` and `.preserve` parameters for the `group_by` and `filter` functions respectively.
* Getting an error between running a manipulation and `group_by` in the new system with embraces


I'll elaborate on the third point. I tried out the calculations system in R-Instat for dlgstartofrains for the Ghana data set.
We group by a station and a year.
When we run a manipulation for mutating a new column for the station variable, we get an error when we then `run_instat_calculation`:

```
grouping_by_station <- instat_calculation$new(type="by", calculated_from=list("ghana"="station"))
roll_sum_rain <- instat_calculation$new(type="calculation", function_exp="RcppRoll::roll_sumr(x=rainfall, n=2, fill=NA, na.rm=FALSE)", result_name="roll_sum_rain", calculated_from=list("ghana"="rainfall"), manipulations=list(grouping_by_station))
```

```
Error: Problem with `mutate()` input `..1`.
x Can't subset columns that don't exist.
x Column `station` doesn't exist.
i Input `..1` is `dplyr::across("station")`.
i The error occurred in group 1: station = "Saltpond".
```

This only occurs when `manipulations` is in the `roll_sum_rain` argument. Otherwise, there is no error.

I've attached the R code for this example because it may be easier to read there. 

[Error in manipulations.zip](https://github.com/africanmathsinitiative/R-Instat/files/5305566/Error.in.manipulations.zip)

lloyddewit:
@dannyparsons are you able to help @lilyclements unblock this PR? thanks

dannyparsons:
Yes I need to get back to looking at this.

dannyparsons:
@lilyclements for `group_by` I think we should use `dplyr::group_by(!!! rlang::syms(by))`. Although this is the "old new" method, it seems to work well. 

I think `syms` is more appropriate than `parse_expr` as `syms` is specifically for referring to the names of things, as explained here https://stackoverflow.com/questions/50158884/what-is-the-difference-between-sym-and-parse-expr-in-the-rlang-package

lilyclements:
@dannyparsons great - thanks for this, I agree it works well. This solves point 3 (and I assume gives the answer to point 1), which just leaves adding the `.drop` and `.preserve` parameters. I have tried this but am a little unsure - are you okay to sort that, otherwise would you mind having a skype so we can talk it through?

dannyparsons:
Yep, I'm fine to quickly do this. I can also show you how I do it, so you see what was needed for it.

lilyclements:
@dannyparsons sure, whichever is easier for you - I can always read the code and see where the changes were to work out what I was missing.

lilyclements:
@dannyparsons are you okay to add the `.drop` and `.preserve` parameters for this PR?

rdstern:
@dannyparsons this one seems to be "in your court"!  Does it wait, or can we quickly get rid of this pull request?

PR #5989, opened 2020-09-18
Reordering of Multiple Items in ucrReorder Control

### LINKED ISSUE #3589: Reorder control enhancement
I think that the reorder dialogs can be improved in a few ways.

It would be good if we could select multiple items to reorder. For example, with months the order sometimes goes: 1,10,11,12,2,3,4,5,6,7,8,9. I would just want to put 10,11,12 at the end. It would be good if I could just select 10,11,12 and click to move them all to the bottom at once

Patowhiz:
Fixes #3589

### DIFF
--- a/instat/ucrReorder.vb
+++ b/instat/ucrReorder.vb
@@ -14,29 +14,15 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-Imports instat
 Imports RDotNet
 Public Class ucrReorder
     Public Event OrderChanged()
     Public WithEvents ucrDataFrameList As ucrDataFrame
     Public WithEvents ucrReceiver As ucrReceiverSingle
-    Private strDataType As String
-    Dim selectedListViewItem As ListViewItem
-    Dim selectedIndex As Integer
-    Dim itemsCount As Integer
+    Private strDataType As String = ""
     Public bWithQuotes As Boolean = True
     Public bIsDataType As Boolean = True
     Public Event SelectedIndexChanged(sender As Object, e As EventArgs)
-
-    Public Sub New()
-
-        ' This call is required by the designer.
-        InitializeComponent()
-
-        ' Add any initialization after the InitializeComponent() call.
-        strDataType = ""
-        selectedListViewItem = New ListViewItem
-    End Sub
 
     Public Sub setDataType(strType As String)
         strDataType = strType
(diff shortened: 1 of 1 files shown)


rdstern (approved):
Great, this now seems to work.  Over to @lloyddewit to be able to merge

lloyddewit:

Ivanluv:
@lloyddewit  this looks okay

instat/ucrReorder.vb
```diff
@@ -57,78 +43,81 @@ Public Class ucrReorder
...
+
+        'remove all selected items
+        For Each kvpItem As KeyValuePair(Of Integer, ListViewItem) In dctSelectedItems
+            lstAvailableData.Items.Remove(kvpItem.Value)
+        Next
+
+        'reorder the items in the list view based on the option given
+        Select Case strDirection
+            Case "top"
+                For Each kvpItem As KeyValuePair(Of Integer, ListViewItem) In dctSelectedItems
```
lloyddewit, line 88: Move the `For` and `Next` lines outside the `Select` statement (makes code 6 lines shorter)?

instat/ucrReorder.vb
```diff
@@ -57,78 +43,81 @@ Public Class ucrReorder
...
+                    If kvpItem.Key >= lstAvailableData.Items.Count - 1 Then
+                        lstAvailableData.Items.Add(kvpItem.Value)
+                    Else
+                        lstAvailableData.Items.Insert(kvpItem.Key + 1, kvpItem.Value)
+                    End If
+                Next
+            Case "bottom"
+                For Each kvpItem As KeyValuePair(Of Integer, ListViewItem) In dctSelectedItems
+                    lstAvailableData.Items.Add(kvpItem.Value)
+                    iStartindex = iStartindex + 1
```
lloyddewit, line 111: is this line needed?

rdstern (dismissed):
I have checked again, because I needed to use this facility.  It works fine still from my point of view.  Back to @lloyddewit

lloyddewit (changes requested):
@Patowhiz this looks clear and elegant, I especially appreciated the excellent commenting. 
I just had a couple of observations

instat/ucrReorder.vb
Patowhiz: @lloyddewit aha. That was the catch 22. Make the lines shorter or make the code faster. Putting the Select statement outside means the same checking will be done several times in the same loop. Putting the for loop inside the select statement means the check happens once.

They will both produce the same output and they both look simpler and direct to me. 

So the questions I had were; Shorter lines? or Faster code?
What do you recommend, I'm happy with either, since in this case, optimised code may never have any noticeable effect, and the lines shortened are just 6.

instat/ucrReorder.vb
Patowhiz: Thanks. This is not needed

instat/ucrReorder.vb
lloyddewit: You probably know my view ;)
In this case (and in the vast majority of cases) readability is more important than performance. If we are reordering millions of lines then having the extra 6 duplicate lines may have a measurable performance improvement but I'm not convinced.
Accept or reject as you wish :)

instat/ucrReorder.vb
```diff
@@ -57,78 +43,81 @@ Public Class ucrReorder
...
-            lstAvailableData.Select()
-            selectedListViewItem.EnsureVisible()
-            RaiseEvent OrderChanged()
-        End If
+        ReorderListViewItems("top")
+    End Sub
+
+    ''' <summary>
+    ''' reorders the items in the listview based on the direction option given and raises reorder event
+    ''' opted to use string parameter because the sub is used inside this class only and by 4 events
```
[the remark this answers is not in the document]
Patowhiz: Agreed.

instat/ucrReorder.vb
Patowhiz: Shortened the lines

Patowhiz:
Made the requested changes

lloyddewit (approved):

lloyddewit:
@rdstern please could you re-approve? thanks

rdstern (approved):

PR #5990, opened 2020-09-21
Viewing Last Graph In R Viewer

### LINKED ISSUE #5884: Add to the graph option on the toolbar
We now have a good example of a non-ggplot graph in R-Instat.  (This is a Taylor plot in the Climatic > Prepare menuIt would be good for this to display "In R window".  This can be done via the Tools > Options > Graph Options menu.  

But this is tedious.  It would be good if it could be an option on the toolbar and the graph icon does just what is needed for a ggplot graph.  Could it also do it for an ordinary graph?  If not could the same code as is used in the Tools > Options, perhaps as an extra option.  
(The difference is that the Tools > Options then sets it this way for ever.  So ideally it would be a switch so it could be set back to the Output window again.  Perhaps both could be done?)

If it is easy to do this, then please consider a similar change in the Describe > Display graph dialogue.  There again, Display in R viewer only works for ggplot graphs.

Patowhiz:
Fixes #5884
@africanmathsinitiative/developers this ready for review.

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -1648,7 +1648,7 @@ DataSheet$set("public", "add_object", function(object, object_name) {
   if(object_name %in% names(private$objects)) message("An object called ", object_name, " already exists. It will be replaced.")
   private$objects[[object_name]] <- object
   self$append_to_changes(list(Added_object, object_name))
-  if(any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot") %in% class(object))) {
+  if(any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "openair") %in% class(object))) {
     private$.last_graph <- object_name
   }
 }
@@ -1672,7 +1672,7 @@ DataSheet$set("public", "get_object_names", function(type = "", as_list = FALSE,
   if(type == "") out = names(private$objects)
   else {
     if(type == model_label) out = names(private$objects)[!sapply(private$objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "htmlTable", "Surv") %in% class(x)))]
-    else if(type == graph_label) out = names(private$objects)[sapply(private$objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot") %in% class(x)))]
+    else if(type == graph_label) out = names(private$objects)[sapply(private$objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "openair") %in% class(x)))]
     else if(type == surv_label) out = names(private$objects)[sapply(private$objects, function(x) any(c("Surv") %in% class(x)))]
     else if(type == table_label) out = names(private$objects)[sapply(private$objects, function(x) any(c("htmlTable", "data.frame", "list") %in% class(x)))]
     else stop("type: ", type, " not recognised")
--- a/instat/static/InstatObject/R/instat_object_R6.R
+++ b/instat/static/InstatObject/R/instat_object_R6.R
@@ -554,8 +554,7 @@ DataBook$set("public", "get_objects", function(data_name, object_name, include_o
 DataBook$set("public", "get_object_names", function(data_name, include_overall = TRUE, include, exclude, type = "", include_empty = FALSE, as_list = FALSE, excluded_items = c()) {
   if(type == "") overall_object_names = names(private$.objects)
   else {
-    if(type == model_label) overall_object_names = names(private$.objects)[!sapply(private$.objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "
-	", "ggsurv", "ggsurvplot", "htmlTable", "Surv") %in% class(x)))]
+    if(type == model_label) overall_object_names = names(private$.objects)[!sapply(private$.objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "ggsurv", "ggsurvplot", "htmlTable", "Surv") %in% class(x)))]
     else if(type == graph_label) overall_object_names = names(private$.objects)[sapply(private$.objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "openair") %in% class(x)))]
     else if(type == surv_label) overall_object_names = names(private$.objects)[sapply(private$.objects, function(x) any(c("Surv") %in% class(x)))]
     else if(type == table_label) overall_object_names = names(private$.objects)[sapply(private$.objects, function(x) any(c("htmlTable") %in% class(x)))]
--- a/instat/static/InstatObject/R/instat_object_R6.R
+++ b/instat/static/InstatObject/R/instat_object_R6.R
@@ -523,7 +523,7 @@ DataBook$set("public", "get_objects", function(data_name, object_name, include_o
       if(!include_empty) out = out[sapply(out, function(x) length(x) > 0)]
     }
     if(!missing(object_name) && length(object_name) == 1) {
-      if(print_graph && (ggplot2::is.ggplot(out) || any(c("gg", "ggmultiplot") %in% class(out)))) return(print(out))
+      if(print_graph && (ggplot2::is.ggplot(out) || any(c("gg", "ggmultiplot", "openair") %in% class(out)))) return(print(out))
       else return(out)
     }
     else return(out)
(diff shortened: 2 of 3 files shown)


rdstern (commented):
Great.  This works.  However I have the idea from @Patowhiz that it is quite specific and I wonder how specific?  It certainly works for the Taylor plot from the openair package.  This package uses the lattice graphics system, rather than ggplot2.  We are about to add more graphs from the openair package. 
I think it depends on the class of the object, so is this a Taylor Plot class, or an Openair class or (ideally) a lattice plot class?

And we are just about to add a series of dialogues that give graphs from the circular package.  An example is the code here (we don't have the dialogues yet, but will soon.  The circular package (and also the extRemes package, that has graphs as well) use the ordinary R graphics system.  For now here is an example that runs in the script window.

```
# Generate 100 observations from a von Mises distribution.
# with mean direction 0 and concentration 3.
data.vm <- circular::rvonmises(n=100, mu=circular(0), kappa=1)
# Plot data set. All points do not fit on plot.
plot(data.vm, stack=TRUE, bins=150) 
```
When I run this example I seem to get the graph automatically in the R viewer, rather than the output window.  That's maybe a different problem, but interesting also?  That may be machine dependent to some extent?

Could @dannyparsons check the code, in parallel to further discussion and change, because there could also be a need for more detailed discussion on the way these graph objects are added to the data sheet.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -554,8 +554,9 @@ DataBook$set("public", "get_objects", function(data_name, object_name, include_o
 DataBook$set("public", "get_object_names", function(data_name, include_overall = TRUE, include, exclude, type = "", include_empty = FALSE, as_list = FALSE, excluded_items = c()) {
   if(type == "") overall_object_names = names(private$.objects)
   else {
-    if(type == model_label) overall_object_names = names(private$.objects)[!sapply(private$.objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "htmlTable", "Surv") %in% class(x)))]
-    else if(type == graph_label) overall_object_names = names(private$.objects)[sapply(private$.objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot") %in% class(x)))]
+    if(type == model_label) overall_object_names = names(private$.objects)[!sapply(private$.objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "
+	", "ggsurv", "ggsurvplot", "htmlTable", "Surv") %in% class(x)))]
```
lloyddewit, line 558: Is this newline intentional? Is `ggmultiplot` missing?

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -523,7 +523,7 @@ DataBook$set("public", "get_objects", function(data_name, object_name, include_o
       if(!include_empty) out = out[sapply(out, function(x) length(x) > 0)]
     }
     if(!missing(object_name) && length(object_name) == 1) {
-      if(print_graph && (ggplot2::is.ggplot(out) || any(c("gg", "ggmultiplot") %in% class(out)))) return(print(out))
+      if(print_graph && (ggplot2::is.ggplot(out) || any(c("gg", "ggmultiplot", "openair") %in% class(out)))) return(print(out))
```
lloyddewit, line 526: This line does not check for `gtable`, `grob` etc. Is this correct?
Same comment for line 547 below

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1648,7 +1648,7 @@ DataSheet$set("public", "add_object", function(object, object_name) {
   if(object_name %in% names(private$objects)) message("An object called ", object_name, " already exists. It will be replaced.")
   private$objects[[object_name]] <- object
   self$append_to_changes(list(Added_object, object_name))
-  if(any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot") %in% class(object))) {
+  if(any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "openair") %in% class(object))) {
```
lloyddewit, line 1651: The same (or similar) list of strings is checked multiple times in this file and also in `instat_object_R6.R`. Remove this duplication by creating a constant used by both R files?

lloyddewit (commented):
@Patowhiz Thanks, this looks good, I only had some minor comments.
@rdstern asked @dannyparsons to comment so he may also have feedback.

instat/static/InstatObject/R/data_object_R6.R
Patowhiz: @lloyddewit yes that would be possibly an ideal solution(even in other functions in the R script file). As I've mentioned in issue #5884, we may have to review this approach in a more general way. I looked at the current code and tried to offer an easier quicker solution to this issue. I'm happy look into a much better solution and refactoring of the code once I get @dannyparsons input.

instat/static/InstatObject/R/instat_object_R6.R
Patowhiz: @lloyddewit thank you this was not intentional change. I've undone the change.

instat/static/InstatObject/R/instat_object_R6.R
Patowhiz: @lloyddewit yes, I also didn't understand why the original code wasn't checking gtable, grob. I didn't to add them since I didn't understand why they were not originally included. I added the openair check cause I think this check could relate to the issue if the object_name parameter ever happens to be missing.

Patowhiz:
> Great. This works. However I have the idea from @Patowhiz that it is quite specific and I wonder how specific? It certainly works for the Taylor plot from the openair package. This package uses the lattice graphics system, rather than ggplot2. We are about to add more graphs from the openair package.
> I think it depends on the class of the object, so is this a Taylor Plot class, or an Openair class or (ideally) a lattice plot class?
> 
> So, the next challenge is (I think) and ordinary graph. This uses the fitdistrplus package. I continued with the dodoma data and used Model > One Variable > Fit Model. Then use, say tmax and check that the display will give a multiplot (say). This is in the output window, but (as expected) the plot icon doesn't yet work for this graph.
> 
> And we are just about to add a series of dialogues that give graphs from the circular package. An example is the code here (we don't have the dialogues yet, but will soon. The circular package (and also the extRemes package, that has graphs as well) use the ordinary R graphics system. For now here is an example that runs in the script window.
> 
> ```
> # Generate 100 observations from a von Mises distribution.
> # with mean direction 0 and concentration 3.
> data.vm <- circular::rvonmises(n=100, mu=circular(0), kappa=1)
> # Plot data set. All points do not fit on plot.
> plot(data.vm, stack=TRUE, bins=150) 
> ```
> 
> When I run this example I seem to get the graph automatically in the R viewer, rather than the output window. That's maybe a different problem, but interesting also? That may be machine dependent to some extent?
> 
> Could @dannyparsons check the code, in parallel to further discussion and change, because there could also be a need for more detailed discussion on the way these graph objects are added to the data sheet.

@rdstern when I analysed the class types of plot object produced in R-Instat, the object class type was of openair. I may need to look into all the plots produced by the package to determine if the plot objects produced differ in their class types. So based on this approach, if the class types are different, then it means keeping on adding the class types of the different object types produced. But as I mentioned in issue #5884, subject to  @dannyparsons opinion,  we may need to rethink on the entire approach in tracking the plot objects produced, and even possibly other types of objects.

@rdstern in regards to running the R scripts script window , the plots are always produced in the R viewer because of how the RunScript in the clsRLink is currently implemented. We could discuss this in a separate issue.

Patowhiz:
@rdstern  **Model > One Variable > Fit Model** dialog uses the graphics package to produce the plot. The `graphics::plot()` function returns `NULL` . So the function does not even produce an object. All graphs are handled as objects in R-Instat and since the plot function does not return any object, I think the current R-Instat way of handling graphs cannot work. @dannyparsons can verify this .

Script produced by the dialog
```
# Code generated by the dialog, One Variable Fit Model

tmax <- data_book$get_columns_from_data(data_name="dodoma", col_names="tmax")
last_model <- fitdistrplus::fitdist(data=as.vector(x=stats::na.exclude(object=tmax)), method="mle", distr="norm")
data_book$add_model(model_name="last_model", model=last_model, data_name="dodoma")
data_book$get_models(data_name="dodoma", model_name="last_model")
last_graph <- graphics::plot(x=data_book$get_models(data_name="dodoma", model_name="last_model"))
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="dodoma")
data_book$get_graphs(data_name="dodoma", graph_name="last_graph")
rm(list=c("last_model", "tmax", "last_graph"))
```

Of course the above script code shows that the produced script code is not optimised (and I think the dialog needs to be re-looked at, to ensure the redundant script code is not produced) but that is besides the point. The point is, I think running `data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="dodoma")` has no effect, because no object will added to the list of objects and thus last graph cannot be reproduced. the object ending up being produced by this dialog is the model object.

So I'm unable to solve this dialog and I think any dialog that uses the `graphics::plot()` function to produce graphs in the way I have done in this PR. In fact I'm strongly leaning towards proposing against handling graphs as objects. Should we ever use other packages that have plot functions that don't produce objects, using current R-Instat implementation, I think we won't be able to view and get last plot.

Patowhiz:
We have discussed with @rdstern that supporting the plots produced by the graphics package can be done in another PR

lloyddewit:
@rdstern Based on your discussion with @Patowhiz  yesterday, are you willing to approve this PR in its current state? thanks

rdstern (approved):
This is a useful addition.  It is incomplete, but already extends the graphs.  So good to merge now.

lloyddewit (approved):

PR #6006, opened 2020-10-06
Seasonality Plots dialog (branched from HawardKetoyoMsatsi:seasonalityplots)

dannyparsons:
Branched from #5984 

- Changed estimates summary options after discussion with @rdstern. Estimates' summary now fixed to the same as Reference Values' but thresholds can be different.
- Set mean as the default summary
- Fixed R code setting for controls where this was not set up
- Corrected variables from receivers to be as strings not RFunctions

### DIFF
--- a/instat/dlgSeasonalPlot.vb
+++ b/instat/dlgSeasonalPlot.vb
@@ -583,8 +583,8 @@ Public Class dlgSeasonalPlot
     End Sub
 
     Private Sub UpdateSummaryParameters()
-        Dim strReferenceName As String = ""
-        Dim strEstimatesName As String = ""
+        Dim strReferenceName As String
+        Dim strEstimatesName As String
 
         strReferenceName = ucrReceiverReference.GetVariableNames(False)
         strEstimatesName = ucrReceiverEstimates.GetVariableNames(False)
--- a/instat/dlgSeasonalPlot.vb
+++ b/instat/dlgSeasonalPlot.vb
@@ -32,12 +32,18 @@ Public Class dlgSeasonalPlot
     Private clsGreaterOperator As ROperator
     Private clsEst2GreaterOperator As ROperator
     Private clsEst2LessOperator As ROperator
+    ''' <summary> Dummy ROperator set to either clsRefGreaterOperator or clsRefLessOperator depending on summary option. This is used in SetRCode for ucrInputReferenceSummary. </summary>
+    Private clsRefGreaterOrLessOperator As ROperator
 
     Private clsMissingFilterFunction As RFunction
     Private clsRefIsNotNaFunction As RFunction
     Private clsEstIsNotNaFunction As RFunction
     Private clsMutateFunction As RFunction
     Private clsPivotLongerFunction As RFunction
+    Private clsPivotCFunction As RFunction
+    Private clsPivotListFunction As RFunction
+    Private clsPivotFactorFunction As RFunction
+    Private clsPivotFactorLevelsCFunction As New RFunction
     Private clsGroupByFunction As RFunction
     Private clsSummariseFunction As RFunction
     Private clsMovingMutateFunction As RFunction
@@ -225,6 +231,10 @@ Public Class dlgSeasonalPlot
         clsEstIsNotNaFunction = New RFunction
         clsMutateFunction = New RFunction
         clsPivotLongerFunction = New RFunction
+        clsPivotCFunction = New RFunction
+        clsPivotListFunction = New RFunction
+        clsPivotFactorFunction = New RFunction
+        clsPivotFactorLevelsCFunction = New RFunction
         clsGroupByFunction = New RFunction
         clsSummariseFunction = New RFunction
         clsMovingMutateFunction = New RFunction
@@ -261,13 +271,11 @@ Public Class dlgSeasonalPlot
 
         clsPipeOperator.SetOperation("%>%")
         clsPipeOperator.AddParameter("filter", clsRFunctionParameter:=clsMissingFilterFunction, iPosition:=1)
-        clsPipeOperator.AddParameter("mutate", clsRFunctionParameter:=clsMutateFunction, iPosition:=2)
         clsPipeOperator.AddParameter("pivot_longer", clsRFunctionParameter:=clsPivotLongerFunction, iPosition:=3)
         clsPipeOperator.AddParameter("group_by", clsRFunctionParameter:=clsGroupByFunction, iPosition:=5)
         clsPipeOperator.AddParameter("summarise", clsRFunctionParameter:=clsSummariseFunction, iPosition:=6)
         clsPipeOperator.AddParameter("moving_mutate", clsRFunctionParameter:=clsMovingMutateFunction, iPosition:=7)
         clsPipeOperator.AddParameter("ggplot", clsROperatorParameter:=clsPlusOperator, iPosition:=8)
-
 
         clsAndOperator.SetOperation("&")
         clsAndOperator.AddParameter("ref", clsRFunctionParameter:=clsRefIsNotNaFunction, iPosition:=0)
@@ -303,9 +311,20 @@ Public Class dlgSeasonalPlot
 
         clsPivotLongerFunction.SetPackageName("tidyr")
         clsPivotLongerFunction.SetRCommand("pivot_longer")
-        clsPivotLongerFunction.AddParameter("cols", "c(reference, estimates)", iPosition:=0)
+        clsPivotLongerFunction.AddParameter("cols", clsRFunctionParameter:=clsPivotCFunction, iPosition:=0)
         clsPivotLongerFunction.AddParameter("names_to", Chr(34) & "data" & Chr(34), iPosition:=1)
         clsPivotLongerFunction.AddParameter("values_to", Chr(34) & "value" & Chr(34), iPosition:=2)
+        clsPivotLongerFunction.AddParameter("names_ptypes", clsRFunctionParameter:=clsPivotListFunction, iPosition:=6)
+
+        clsPivotCFunction.SetRCommand("c")
+
+        clsPivotListFunction.SetRCommand("list")
+        clsPivotListFunction.AddParameter("data", clsRFunctionParameter:=clsPivotFactorFunction, iPosition:=0)
+
+        clsPivotFactorFunction.SetRCommand("factor")
+        clsPivotFactorFunction.AddParameter("levels", clsRFunctionParameter:=clsPivotFactorLevelsCFunction, iPosition:=1)
+
+        clsPivotFactorLevelsCFunction.SetRCommand("c")
 
         clsGroupByFunction.SetPackageName("dplyr")
         clsGroupByFunction.SetRCommand("group_by")
@@ -443,11 +462,13 @@ Public Class dlgSeasonalPlot
     Private Sub SetRCodeForControls(bReset As Boolean)
         ucrReceiverReference.AddAdditionalCodeParameterPair(clsRefGreaterOperator, ucrReceiverReference.GetParameter, iAdditionalPairNo:=1)
         ucrReceiverReference.AddAdditionalCodeParameterPair(clsRefLessOperator, ucrReceiverReference.GetParameter, iAdditionalPairNo:=2)
+        ucrReceiverReference.AddAdditionalCodeParameterPair(clsPivotCFunction, New RParameter("0", iNewPosition:=0, bNewIncludeArgumentName:=False), iAdditionalPairNo:=3)
 
         ucrReceiverEstimates.AddAdditionalCodeParameterPair(clsEst1GreaterOperator, ucrReceiverEstimates.GetParameter, iAdditionalPairNo:=1)
         ucrReceiverEstimates.AddAdditionalCodeParameterPair(clsEst1LessOperator, ucrReceiverEstimates.GetParameter, iAdditionalPairNo:=2)
         ucrReceiverEstimates.AddAdditionalCodeParameterPair(clsEst2GreaterOperator, ucrReceiverEstimates.GetParameter, iAdditionalPairNo:=3)
         ucrReceiverEstimates.AddAdditionalCodeParameterPair(clsEst2LessOperator, ucrReceiverEstimates.GetParameter, iAdditionalPairNo:=4)
+        ucrReceiverEstimates.AddAdditionalCodeParameterPair(clsPivotCFunction, New RParameter("1", iNewPosition:=1, bNewIncludeArgumentName:=False), iAdditionalPairNo:=5)
 
         ucrInputReferenceThreshold.AddAdditionalCodeParameterPair(clsRefLessOperator, ucrInputReferenceThreshold.GetParameter, iAdditionalPairNo:=1)
         ucrInputReferenceThreshold.AddAdditionalCodeParameterPair(clsLessOperator, ucrInputReferenceThreshold.GetParameter, iAdditionalPairNo:=2)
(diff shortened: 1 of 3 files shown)


instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,700 @@
...
+        ucrBase.clsRsyntax.iCallType = 3
+        ucrBase.clsRsyntax.bExcludeAssignedFunctionOutput = False
+
+        ucrSelectorSeasonalityComparisons.SetParameter(New RParameter("data", 0))
+        ucrSelectorSeasonalityComparisons.SetParameterIsrfunction()
+
+        ucrReceiverReference.SetParameter(New RParameter("x", 0))
+        ucrReceiverReference.Selector = ucrSelectorSeasonalityComparisons
+        ucrReceiverReference.SetParameterIsString()
+        ucrReceiverReference.bWithQuotes = False
```
dannyparsons, line 101: This was set as `.SetParameterIsRFunction()` for all receivers whereas it should be strings without quotes. For `dplyr` and `ggplot2` functions we only need the variable names, not to extract the columns. Before, there was lots of extra lines getting the columns from the data frames before doing the plot which were unnecessary.

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,700 @@
...
+        ucrReceiverReference.SetDataType("numeric")
+        ucrReceiverReference.strSelectorHeading = "Numerics"
```
dannyparsons, line 103: Data types were not set for any receivers.

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,700 @@
...
+        ucrReceiverStation.bWithQuotes = False
+        ucrReceiverStation.SetClimaticType("station")
+        ucrReceiverStation.bAutoFill = True
+        ucrReceiverStation.SetDataType("factor")
+        ucrReceiverStation.strSelectorHeading = "Factors"
+
+        ucrInputReferenceSummary.SetItems({"Mean", "Mean Above", "Mean Below", "Prop Above", "Prop Below", "Std.dev"})
+        ucrInputReferenceSummary.SetDropDownStyleAsNonEditable()
+
+        ucrInputEstimatesSummary.IsReadOnly = True
```
dannyparsons, line 135: Changed the functionality of this after discussing with Roger. This now displays the same summary as for Reference Values and can't be changed.

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,700 @@
...
+    Private clsMeanFilterFunction As RFunction
+    Private clsStdFunction As RFunction
+
+    'ggplot functions + operators
+    Private clsPlusOperator As ROperator
+    Private clsFacetWrapTildeOperator As ROperator
+    Private clsTildeOperator As ROperator
+
+    ''' <summary> Dummy RCodeStructure set to either clsAsFormulaFunction or clsTildeOperator depending on smoothing option. This is used in SetRCode for ucrInputSmoothing. </summary>
+    Private clsFormulaRCode As RCodeStructure
```
dannyparsons, line 58: To be able to set the R code for `ucrInputSmoothing` I created this "dummy" `RCodeStructure` which is set when the smoothing option is changed and allows the R code to be set. See below for where it's used.

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,700 @@
...
+            clsPlusOperator.AddParameter("geom_point", clsRFunctionParameter:=clsGeomPointFunction, iPosition:=1)
+        Else
+            clsPlusOperator.RemoveParameterByName("geom_point")
+        End If
+    End Sub
+
+    Private Sub ucrReceiverStation_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverStation.ControlValueChanged
+        ' Facets are handles manually since they could also be managed on a sub dialog.
+        ' The check for clsFacetWrapTildeOperator.clsParameters.Count = 0 ensures that facets added on a subdialog are not removed when ucrReceiverStation is empty.
+        If ucrReceiverStation.IsEmpty AndAlso clsFacetWrapTildeOperator.clsParameters.Count = 0 Then
```
dannyparsons, line 690: This is needed in case we later add the Plot Options subdialog and allow facets to be added from there also.

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,700 @@
...
+        UpdateSummaryParameters()
+    End Sub
+
+    Private Sub Smoothing_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputReferenceSummary.ControlValueChanged, ucrInputSmoothing.ControlValueChanged
+        UpdateSmoothingParameters()
+    End Sub
+
+    Private Sub ucrInputReferenceThreshold_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputReferenceThreshold.ControlContentsChanged
+        UpdateEstimatesThresholdParameter()
+    End Sub
```
dannyparsons, line 650: I separated out these different parts of the code so that they are only run when the controls they depend on change.

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,700 @@
...
+                clsSummariseFunction.AddParameter("summary", clsROperatorParameter:=clsDivideOperator, iPosition:=3)
+            Case "Prop Below"
+                clsPipeOperator.RemoveParameterByName("filter2")
+                clsMutateFunction.AddParameter("reference", clsROperatorParameter:=clsRefLessOperator, iPosition:=0)
+                clsMutateFunction.AddParameter("estimates", clsROperatorParameter:=clsEst2LessOperator, iPosition:=1)
+                clsSummariseFunction.AddParameter("summary", clsROperatorParameter:=clsDivideOperator, iPosition:=3)
+            Case "Mean"
+                clsPipeOperator.RemoveParameterByName("filter2")
+                clsMutateFunction.AddParameter("reference", ucrReceiverReference.GetVariableNames(False), iPosition:=0)
+                clsMutateFunction.AddParameter("estimates", ucrReceiverEstimates.GetVariableNames(False), iPosition:=1)
```
dannyparsons, line 576: For all options, these need to be `.GetVariableNames(False)` so that we just get the names of the column. `False` means do not add quotes to the column name.

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,700 @@
...
+        ucrInputSmoothing.SetRCode(clsFormulaRCode, bReset)
+        ' Cannot easily set R code using conditions because checks on multiple functions is needed hence done "manually" in this sub.
+        ReferenceSummarySetRCode(bReset)
+    End Sub
+
+    ''' <summary>
+    ''' Sets the R Code for ucrInputReferenceSummary. This is done manually by this Sub instead of through conditions because of the complex checks on multiple functions required.
+    ''' </summary>
+    ''' <param name="bReset">True if the dialog is being reset.</param>
+    Private Sub ReferenceSummarySetRCode(bReset As Boolean)
```
dannyparsons, line 491: It's not possible to automatically set the R code for `ucrInputReferenceSummary` because it depends on multiple parts of the code, however, we still need the value to be read from the code so that's what this function does.

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,700 @@
...
+        clsGgplotFunction.AddParameter("mapping", clsRFunctionParameter:=clsGgPlotAesFunction, iPosition:=0)
+
+        clsPlusOperator.SetOperation("+")
+        clsPlusOperator.AddParameter("ggplot", clsRFunctionParameter:=clsGgplotFunction, iPosition:=0)
+        clsPlusOperator.AddParameter("geom_line", clsRFunctionParameter:=clsGeomLineFunction, iPosition:=2)
+        clsPlusOperator.AddParameter("geom_smooth", clsRFunctionParameter:=clsGeomSmoothFunction, iPosition:=3)
+        clsPlusOperator.AddParameter("ylab", clsRFunctionParameter:=clsYlabFunction, iPosition:=4)
+
+        clsFacetWrapTildeOperator.SetOperation("~")
+        clsFacetWrapTildeOperator.bForceIncludeOperation = True
```
dannyparsons, line 376: Instead of adding a blank parameter use `.bForceIncludeOperation = True` to force the operator to appear even when there is only one parameter in the operator.

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,700 @@
...
+
+        clsPropSumFunction.SetRCommand("sum")
+        clsPropSumFunction.AddParameter("x", "value", iPosition:=0)
+        clsPropSumFunction.AddParameter("na.rm", "TRUE", iPosition:=1)
+
+        clsNotIsNaThreshFunction.SetRCommand("!is.na")
+        clsNotIsNaThreshFunction.AddParameter("x", "value", iPosition:=0)
+
+        clsMeanFunction.SetRCommand("mean")
+        clsMeanFunction.AddParameter("x", "value", iPosition:=0)
```
dannyparsons, line 341: In `mean`, `sd` functions etc. you had a different parameter name and had `bIncludeArgumentName:=False`. These should be used just like any other function with correct parameter names.

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,700 @@
...
+        autoTranslate(Me)
+        If bFirstLoad Then
+            InitiliseDialog()
+            bFirstLoad = False
+        End If
+
+        If bReset Then
+            SetDefaults()
+        End If
+        SetRCodeForControls(bReset)
```
dannyparsons, line 86: This is run every time the dialog opens, not just on reset.

dannyparsons (commented):
@HawardKetoyoMsatsi @shadrackkibet I've added comments here to explain the changes I've made from #5984. You can see the full comparison here https://github.com/HawardKetoyoMsatsi/R-Instat/compare/seasonalityplots...dannyparsons:HawardKetoyoMsatsi-seasonalityplots

The dialog was producing the correct output with all different options, but some of the things mentioned below are ways I corrected the R code that was being produced, or corrected the VB code to follow our normal conventions, particularly ensuring all controls are reading their value from the R code.

lloyddewit:

@dannyparsons I understand that:

- this PR is ready for test
- this PR supercedes PR  #5984, no more work is expected on PR #5984 
- PR #5984 will merge automatically when this PR merges

If I misundersood anything then please let me know, thanks

rdstern (changes requested):
I was going to  approve, because it could usefully be merged now.  But I have a variety of suggestions, either for now, or for later, once it has been merged.  No I would like point a) to be considered first.
a) The data are labelled as reference and estimated.  And these are given in the legend, in the opposite order to the dialogue.  But the main change is to give the actual name of the variable, rather than the standard name.

From here, these could be suggestions for a new issue, after merging.
b) Could the Missing checkbox be moved down, so it is at the bottom, rather than the top of the controls on the left?  That is more consistent with the missing option on the summary dialogue.  It also leaves a space for...
c) There are a lot more minor changes that may become useful, if it is practical to add the Plot Options, which should go just below the data selector.  Perhaps it could go there now, even though disabled?  (Among these are position of the legend, having dates in the year for the x-axis.
d) Could a default graph caption give details of what is being done in this dialogue?

dannyparsons:
@lloyddewit yes that's right exactly

dannyparsons:
@rdstern a) and b) have now been addressed. In particular, I found the `names_ptypes` parameter of `pivot_longer` which can define the factor column to be produced from stacking to ensure the labels are always in the same order (instead of being in alphabetical order)

I agree with making c) and d) a separate issue for further improvement after this.

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,743 @@
...
+        ucrNudDf.SetRCode(clsNsFunction, bReset)
+        ucrNudHarmonics.SetRCode(clsFourierSeriesFunction, bReset)
+        ucrNudPeriod.SetRCode(clsFourierSeriesFunction, bReset)
+        ucrInputSmoothing.SetRCode(clsFormulaRCode, bReset)
+        ' Cannot easily set R code using conditions because checks on multiple functions is needed hence done "manually" in this sub.
+        ReferenceSummarySetRCode(bReset)
+    End Sub
+
+    ''' <summary>
+    ''' Sets the R Code for ucrInputReferenceSummary. This is done manually by this Sub instead of through conditions because of the complex checks on multiple functions required.
```
lloyddewit, line 509:
```suggestion
    ''' Sets the R Code for ucrInputReferenceSummary. This is done manually by this sub instead of 
    ''' through conditions because of the complex checks on multiple functions required.
```

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,743 @@
...
+    Private clsRefGreaterOperator As ROperator
+    Private clsEst1GreaterOperator As ROperator
+    Private clsRefLessOperator As ROperator
+    Private clsEst1LessOperator As ROperator
+    Private clsDivideOperator As ROperator
+    Private clsLessOperator As ROperator
+    Private clsGreaterOperator As ROperator
+    Private clsEst2GreaterOperator As ROperator
+    Private clsEst2LessOperator As ROperator
+    ''' <summary> Dummy ROperator set to either clsRefGreaterOperator or clsRefLessOperator depending on summary option. This is used in SetRCode for ucrInputReferenceSummary. </summary>
```
lloyddewit, line 35:
```suggestion
    ''' <summary> 
    ''' Dummy ROperator set to either clsRefGreaterOperator or clsRefLessOperator depending on 
    ''' summary option. This is used in SetRCode for ucrInputReferenceSummary. 
    ''' </summary>
```

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,743 @@
...
+        SetRCodeForControls(True)
+        TestOkEnabled()
+    End Sub
+
+    Private Sub UpdateSummaryParameters()
+        Dim strReferenceName As String = ""
+        Dim strEstimatesName As String = ""
+
+        strReferenceName = ucrReceiverReference.GetVariableNames(False)
+        strEstimatesName = ucrReceiverEstimates.GetVariableNames(False)
```
lloyddewit, line 590:
```suggestion
        Dim strReferenceName As String = ucrReceiverReference.GetVariableNames(False)
        Dim strEstimatesName As String = ucrReceiverEstimates.GetVariableNames(False)
```

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,743 @@
...
+
+    Private Sub ucrInputReferenceThreshold_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputReferenceThreshold.ControlContentsChanged
+        UpdateEstimatesThresholdParameter()
+    End Sub
+
+    Private Sub UpdateEstimatesThresholdParameter()
+        ' By default estimates has the same threshold as reference values, unless the user has already typed in ucrInputEstimatesThreshold to make them different.
+        If Not ucrInputEstimatesThreshold.UserTyped Then
+            ucrInputEstimatesThreshold.SetName(ucrInputReferenceThreshold.GetText())
+        End If
```
lloyddewit, line 699:
```suggestion
        ' By default estimates has the same threshold as reference values, unless the user has 
        ' already typed in ucrInputEstimatesThreshold to make them different.
        If Not ucrInputEstimatesThreshold.UserTyped Then
            ucrInputEstimatesThreshold.SetName(ucrInputReferenceThreshold.GetText())
        End If
```

instat/dlgSeasonalPlot.vb
```diff
@@ -1,3 +1,743 @@
...
+        If ucrChkPoints.Checked Then
+            clsPlusOperator.AddParameter("geom_point", clsRFunctionParameter:=clsGeomPointFunction, iPosition:=1)
+        Else
+            clsPlusOperator.RemoveParameterByName("geom_point")
+        End If
+    End Sub
+
+    Private Sub ucrReceiverStation_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverStation.ControlValueChanged
+        ' Facets are handles manually since they could also be managed on a sub dialog.
+        ' The check for clsFacetWrapTildeOperator.clsParameters.Count = 0 ensures that facets added on a subdialog are not removed when ucrReceiverStation is empty.
```
lloyddewit, line 732:
```suggestion
        ' Facets are handled manually since they could also be managed on a sub dialog.
        ' The check for clsFacetWrapTildeOperator.clsParameters.Count = 0 ensures that facets added 
        ' on a subdialog are not removed when ucrReceiverStation is empty.
```

lloyddewit (changes requested):
@dannyparsons just some small comments

instat/dlgSeasonalPlot.vb
dannyparsons: We have been using a convention of defining variables separately to assigning them a value. I am not sure if this is the correct convention to use but that's why I did it like this. I will remove `= ""` as these are unnecessary.
I can't see any mention of what the convention should be for this in the VB coding conventions document. I did find this document https://en.wikibooks.org/wiki/Visual_Basic/Coding_Standards which doesn't mention this convention but in the "Function Return Values" section the example code separates defining the variable from assigning a value.

dannyparsons:
@lloyddewit thanks for the review and feedback. I've accepted all those changes apart from one where I've commented. Let me know if you think that's ok.

instat/dlgSeasonalPlot.vb
lloyddewit: @dannyparsons If that is the normal convention in R-Instat then that's OK for me. I personally find 2 lines more readable here than 4 lines but that's subjective. Consistency is also important.
Thank you for the wikibooks link, I wasn't aware of this. There's also the Microsoft VB coding guidelines (https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/program-structure/coding-conventions). These aren't so extensive as the wikibooks link. They briefly mention declarations. They don't have the exact example above but show the following as good practice. The Visual Studio 'quick actions' (the lightbulb symbol preceeding a code line) also recommends combining declaration and initailisation. So it's subjective!

image

lloyddewit (approved):

instat/dlgSeasonalPlot.vb
dannyparsons: Thanks. Yes I think either would be acceptable as long as we're consistent. Whatever we decide we should be documenting our coding conventions or stating which ones we're following. I think developers have mostly picked them up from the existing code but we should be explicit about it too. We probably want a separate issue on documenting coding conventions.

dannyparsons:
I can replicate this problem is `x` is a factor or if d.f. is < 3. When `x` is a factor none of the curves work. Do you get it in any other case? If it was very important for this to work with `x` as a factor we could probably do this or we restrict `x` to be numeric.
I think we should also restrict d.f. to be >= 3 for splines. It doesn't produce anything otherwise and can't see why < 3 would be especially needed.

rdstern:
hbere is the problem with doy1 numeric and d.f. = 6
image

Here are the results:
image

dannyparsons:
It's not really designed for the case of data summarised to the doy level but it should work still. I tried it with Dodoma and it worked producing the same R code that you showed. Could you send me the data you're using privately so that I can test?

rdstern (approved):
Found the last problem was because a new package needed to be installed.

I note that at some stage a logit transform may be needed.

dannyparsons:
This is now ready. We agreed on not needing any further changes at this point.

PR #6011, opened 2020-10-09
Bug fix: PICSA Crops dialog rain totals

### LINKED ISSUE #5959: Serious problem with the PICSA crops dialogue
This is serious, but I hope simple to solve.
It occurs when there are multiple stations in the file.  Then, as far as I can see, the totals used in the comparison are for all the sites, and not for the single site.
[Moz4CropsTest.zip](https://github.com/africanmathsinitiative/R-Instat/files/5167289/Moz4CropsTest.zip)

If needed the attached file shows the problem.  Look at ctop_prop5 that gives the summary results for the 4 stations and then at crop_def5 that gives the year-by-year results.  The problem is then obvious.  The total for comparison in (say) 1979 is 1305mm, which is very large!  It is also the same value for all 4 stations!  It appears it isn't getting the total for each station, but the overall total for the 4 stations together!

There is another (smaller) error.  shown below:

image

To solve the problem above I made a subset for a single station (Tete) and then changed the data frames to use these data.  Earlier I had used the data frames fr=or all 4 stations.  Something has not been cleared, because it has tried to give some results for all 4 stations.

A third - also minor - issue is when the linking is not correct.  What has happened is that the year variable used when defining climatic was the year factor for the shifted year, (1982-1983) and not the numeric variable.  This seems not to be a problem with the summaries or the start of the rains dialogues.  But here it says that the 2 data frames must be linked.  I am happy that there is an error generated here.  But they are linked - however incorrectly.  It would be good to change the message, so the one given now only happens when they are not linked (which is possible).  And there is another message to say when they are incorrectly linked.  

Another small issue identified earlier was that the end of the rains is not auto-completed.  @dannyparsons says this is actually pretty tricky to fix.

dannyparsons:
fixes #5959

This now has the correct rain totals for multiple stations in the PICSA crops dialog, and has the same correct functionality for a single station, but please check both cases in review.

### DIFF
--- a/instat/static/InstatObject/R/instat_object_R6.R
+++ b/instat/static/InstatObject/R/instat_object_R6.R
@@ -1685,9 +1685,11 @@ DataBook$set("public", "crops_definitions", function(data_name, year, station, r
   plant_day_name <- "plant_day"
   plant_length_name <- "plant_length"
   rain_total_name <- "rain_total"
+  
+  is_station <- !missing(station)
 
   if(missing(year)) stop("Year column must be specified.")
-  if(missing(station)) by <- year
+  if(!is_station) by <- year
   else by <- c(year, station)
   if(missing(season_data_name)) season_data_name <- data_name
   if(season_data_name != data_name) {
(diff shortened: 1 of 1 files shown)


lloyddewit:

rdstern (approved):
Great.

lloyddewit (approved):

PR #6014, opened 2020-10-12
Separating missing value options on summary dialogs and changing single element receiver to a multiple receiver

### LINKED ISSUE #6012: Separate missing value options on summary dialogs
I have a suggested change for missing value options on dialogs that produce summaries. Since videos will be made soon showing these dialogs I want to raise this now and if agreed implement quickly before videos are started.

Currently on e.g. the Column Summaries dialog there is a checkbox to exclude missing values, and then detailed missing value options a tab on the Summaries subdialog. There's several confusing things about this:

- If I don't check the missing check box then any missing options chosen on the sub dialog are ignored - but there's no way of knowing this from the dialog or sub dialog
- From the main dialog there's no indication that there are more missing options and its not very logical that I need to look on the Summaries sub dialog for this
- In general, there is a disconnect between the missing option on the main dialog and sub dialog

To resolve these I suggest a separate "Missing Options" sub dialog which contains the Missing tab currently on the Summaries sub dialog. The missing checkbox on the main dialog remains. The Missing Options button is placed next to the missing checkbox and is only enabled if the check box is checked. This then ensures missing options are only available when checked.
The functionality of all the controls stays exactly the same.

The only down side I see to this is that it uses a bit more space on the main dialog and makes it look a little bit more complex but I think this is worth it to have more clarity on the missing value options.

If it's agreed, @shadrackkibet might be able to do this quite quickly as he worked on the missing options originally? There are a number of affected dialogs but the work is essentially the same on all of them.

### LINKED ISSUE #6025: Small change in climatic summaries dialogue?
Can the data receiver in this dialogue be changed to become a multiple receiver?
I don't know now why we made it a single receiver - I think that was a mistake on my part.

If not easy, then leave it.  But it is a much used dialogue and would be needed for the SPI video as well as in its own right.

Ideally this would be for the forthcoming release?  Is that possible?

shadrackkibet:
Fixes #6012 and Fixes #6025. This is still a work in progress.

### DIFF
--- a/instat/dlgClimaticSummary.vb
+++ b/instat/dlgClimaticSummary.vb
@@ -53,6 +53,7 @@ Public Class dlgClimaticSummary
         ucrBase.clsRsyntax.iCallType = 0
         ucrBase.iHelpTopicID = 510
         ucrChkDropUnusedLevels.Enabled = False ' removed this functionality so this is disabled
+        cmdMissingOptions.Enabled = False
 
         ucrSelectorVariable.SetParameter(New RParameter("data_name", 0))
         ucrSelectorVariable.SetParameterIsString()
@@ -210,7 +211,6 @@ Public Class dlgClimaticSummary
 
         ucrBase.clsRsyntax.ClearCodes()
         ucrBase.clsRsyntax.SetBaseRFunction(clsDefaultFunction)
-        bResetSubdialog = True
     End Sub
 
     Private Sub SetRCodeForControls(bReset As Boolean)
@@ -296,6 +296,13 @@ Public Class dlgClimaticSummary
         Else
             clsDefaultFunction.RemoveParameterByName("use")
         End If
+        cmdMissingOptions.Enabled = ucrChkOmitMissing.Checked
+    End Sub
+
+    Private Sub cmdMissingOptions_Click(sender As Object, e As EventArgs) Handles cmdMissingOptions.Click
+        sdgMissingOptions.SetRFunction(clsNewSummaryFunction:=clsDefaultFunction, clsNewConcFunction:=clsConcFunction, bReset:=bResetSubdialog)
+        bResetSubdialog = False
+        sdgMissingOptions.ShowDialog()
     End Sub
 
     Private Sub ucrReceiverStation_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverDOY.ControlValueChanged, ucrSelectorVariable.ControlValueChanged, ucrChkAddDateColumn.ControlValueChanged, ucrReceiverDate.ControlValueChanged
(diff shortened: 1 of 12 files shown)


shadrackkibet:
@africanmathsinitiative/developers this PR fixes two issues as referenced above. This is ready for review.

lloyddewit:

shadrackkibet:
@rdstern, please test. I would like to get this done as soon as possible.

rdstern (dismissed):
@shadrackkibet this is great and it all seems fine.  This was @dannyparsons suggestion and is one that was blindingly obvious - in hindsight of course.

I am happy to approve so these can be merged.  However there are (at least) 2 more dialogues.  They do need to be changed soon, but this could be a separate issue if you prefer.

They are Describe > One Variable > Summarise and also 2 Variable > Summarise.  Both these dialogues call the same sub-dialogue and the missing options are now absent.  I assume they are simple to add here too.  I have never used the missing options in these dialogues, so it can be left to later if needed.

shadrackkibet:
> They are Describe > One Variable > Summarise and also 2 Variable > Summarise. Both these dialogues call the same sub-dialogue and the missing options are now absent. I assume they are simple to add here too. I have never used the missing options in these dialogues, so it can be left to later if needed.

Let's have it as a separate issue. Thanks

instat/sdgMissingOptions.vb
```diff
@@ -16,7 +16,73 @@
...
+        ucrChkMaxPercMissing.SetText("Maximum percentage of missing allowed")
+
+        ucrChkConsecutiveMissing.SetParameter(New RParameter("con", 4), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "'con'" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
+        ucrChkConsecutiveMissing.SetText("Maximum number of consecutive missing allowed")
+
+        ucrInputPercentage.SetParameter(New RParameter("na_max_prop", 10))
+        ucrInputPercentage.SetValidationTypeAsNumeric(dcmMin:=0, dcmMax:=100)
+        ucrInputPercentage.AddQuotesIfUnrecognised = False
+
+        ucrNudNumber.SetParameter(New RParameter("na_max_n", 11))
```
dannyparsons, line 49: This currently has a max of 100. Need to set min/max to fix this.

instat/dlgColumnStats.vb
```diff
@@ -191,6 +192,12 @@ Public Class dlgColumnStats
         Else
             clsDefaultFunction.RemoveParameterByName("use")
         End If
+        If Not ucrChkOmitMissing.Checked AndAlso clsDefaultFunction.ContainsParameter("na_type") Then
```
dannyparsons, line 195:
```suggestion
        If Not ucrChkOmitMissing.Checked Then
```
You don't want that second condition because it would mean that the parameter could get added if the checkbox was unchecked but the parameter was not included at that point i.e. it could go into `Else` when you don't want it to. There's no issue with trying to remove a parameter that doesn't exist anyway too.

instat/dlgClimaticSummary.vb
```diff
@@ -296,6 +295,18 @@ Public Class dlgClimaticSummary
         Else
             clsDefaultFunction.RemoveParameterByName("use")
         End If
+        If Not ucrChkOmitMissing.Checked AndAlso clsDefaultFunction.ContainsParameter("na_type") Then
```
dannyparsons, line 298:
```suggestion
        If Not ucrChkOmitMissing.Checked Then
```
Same here as explained for the other dialog.

dannyparsons (changes requested):
Looks good, a couple of changes needed.

Ivanluv:
@lloyddewit  the vb code looks okay by me

instat/sdgMissingOptions.vb
shadrackkibet: Done!

dannyparsons (approved):

lloyddewit (approved):

lloyddewit:
@rdstern please could you retest? The only functional change since your last approval was changing the max possible value of 'na_max_n' from 100 to maxint.

dannyparsons:
I've tested that point and it's working so happy for it to be merged.

PR #6028, opened 2020-10-19
Viewing graphs that are produced by openair and graphics package in graph dialog

### LINKED ISSUE #5996: View Graph Dialog
Currently this dialog shows list of ggplot graphs only. The dialog needs to be enhanced to show all graphs that can be displayed through the R Viewer toolbar option.

Patowhiz:
Fixes #5996.

This PR adds support for viewing graphs produced by openair and graphics package.

### DIFF
--- a/instat/dlgOneVarFitModel.vb
+++ b/instat/dlgOneVarFitModel.vb
@@ -259,13 +259,20 @@ Public Class dlgOneVarFitModel
 
         clsROneVarFitModel.SetAssignTo("last_model", strTempDataframe:=ucrSelectorOneVarFitMod.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempModel:="last_model")
         clsRLogLikFunction.SetAssignTo("last_likelihood", strTempDataframe:=ucrSelectorOneVarFitMod.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempGraph:="last_likelihood")
-        clsRplotFunction.SetAssignTo("last_graph", strTempDataframe:=ucrSelectorOneVarFitMod.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempGraph:="last_graph")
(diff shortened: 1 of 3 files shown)


rdstern:
As @dannyparsons was going to help with this, could the option be made comprehensive.  
In the Tools > Options the user can opt for all graphs to be displayed in the R-viewer.  So I suggest that the facility in the toolbar and in the View Graph dialogue should also be possible for any graph?
Currently it is just for a ggplot graph.

a) I believe the aim is to enable the facility for openair - which uses the lattice system.
b) And in the ordinary R graphs - @dannyparsons gave a suggestion on this.  
c) The third option is a combined ggplot graph, which currently isn't possible.  An example of this is in tutorial 1 where, for the ggplot diamonds data, we use the Describe > One Variable > Graph - for all the 10 variables.

rdstern:
@Patowhiz and @dannyparsons this work seems to have stalled.
So I suggest an addition or alternative that could be reasonably comprehensive and simple to implement.  It isn't as good as making the toolbar option available for any type of graph - though it may be useful in addition anyway.
Currently the user can use Tools > Options > Graphics to put graphs by default into either the R viewer or the output window.  Could this become also an option on the pull-down in the toolbar.  It could just be a switch perhaps?  If set to the R-viewer then automatically the graphs would be there instead of in the output window.  It already works from the Tools menu, so should be easy to add in the toolbar?

lloyddewit:
@Patowhiz when you are ready to continue with this PR, then please let @dannyparsons and I know. @dannyparsons would like to discuss the best implementation approach before this restarts, thanks

lloyddewit:
@dannyparsons in your mail of 19 Jan, you mentioned the possibility of @Ivanluv taking over this PR. How would you like to proceed? thanks

Ivanluv:
@lloyddewit  this is currently in progress with help from @volloholic

rdstern (commented):
@Patowhiz and @dannyparsons my understanding is that this has now been done?  So it can be closed?

Patowhiz:
Yes this being done elsewhere.

PR #6037, opened 2020-10-22
Remembering Last Save In Closing Databook

### LINKED ISSUE #6003: Small change to the File > Close Data Book option
I assume this will be quick to do. It is an obvious one for @Patowhiz and check back if it is not a small task.

It correctly gives a warning that all data will be lost.  If I don't want to lose any data, then I save, and then use this option.  But it still asks.  Most software would detect that the data have not been changed and only ask if there were changes.  Could this check be made here, so it does not ask if the data have not changed since the last save.

Patowhiz:
Fixes #6003


lloyddewit:

Patowhiz:
@lloyddewit @shadrackkibet this PR had unintended changes. I have replaced it with PR #6039.

PR #6043, opened 2020-10-25
Fixing Position Bug and Adding ucrSave control to dlgFindNonnumericValues

### LINKED ISSUE #6002: Small addition to Prepare > Check Data > Non-numeric cases
@Patowhiz could you add the position control to this dialogue once @lilyclements changes have been merged.  Could you also add the line, so the default position for the saved variable is after the variable in the receiver, once that has been added.  Ideally add a comment to this line, so Lily can repeat it, in further dialogues that she will examine.

### LINKED ISSUE #6042: Serious bug with the Position button
The recent version (0.6.3) has introduced a bug on the dialogues that use the Position button.
I originally thought this was linked to the new Calculator dialogue that has been updated to include this facility.  But it seems more general than this.
I just tried also with the Right-click > Duplicate dialogue and have the same bug.  After using this facility it doesn't recognise the new variable.  It gives an error when going into any dialogue after this and then when making any change in a receiver within that dialogue.

I think the Duplicate dialogue may give rise to the problem even if you use the default settings.

Added later:  I have just checked and the Duplicate dialogue does give a problem.  Here is how to get it - I am trying on the new master version.
a) Load (say) Dodoma from the Excel file in Open from Library > R-Instat > Browse > Climatic > Climatic data sets, dodoma.  (I think any file will do!)
b) Make year into a factor.
c) Right-click and Duplicate column.  Change type and make year1 numeric.
d) Try (for example) Describe > One Variable > Summarise.

It gives an error message multiple times and then goes into the dialogue.  Then it repeats the error whenever using a control.  It isn't recognising the new variable.

Patowhiz:
Fixes #6002
Fixes #6042

Extended the adding of columns functionality.
Removed bug in the calculation script.
Made dialog `dlgFindNonnumericValues` use `ucrSave` control.


lloyddewit:
@Patowhiz ready for review/test? thanks

Patowhiz:
@lloyddewit @africanmathsinitiative/developers this is now ready for review.

Please note, I had to extend the adding of columns functionality to support this and future dialogs which save columns in separate scripts. This seemed to be the best alternative rather than quickly extending the `calculations.R` scripts.

The current implementation of adding and at the same time positioning columns can be adapted to the older script code(which technically limited this function to columns that didn't exist). So currently there is a duplication of checks, I will add a comment in regards to this once this new implementation is generally agreed.

rdstern:
@Patowhiz I am trying to check your new code.  It says I need to update my version of R - which I have done.  I am using 4.0.3.  So I wonder if the problem is that your version wasn't updated?

Patowhiz:
@rdstern Oh! yes. This PR didn't have the latest changes from master. I've now updated. But let me test if bug mentioned in issue #6042 will now affect this PR.

Patowhiz:
@rdstern now you can test this PR. This dialog will not have the bug described in issue #6042. The functionality extended is designed differently. Please note, it's only dialog `dlgFindNonnumericValues`, because of how the dialog script produces the new logical column.

lilyclements (commented):
It seems that this PR means that the new column is often at the end of the data frame.

1. I think the new generated column is supposed to be next to character column in the receiver for this dialog (I may be wrong, however, this is going off the issue).
2. If I use the right-click on the main data frame to "Convert" a column, e.g. "Convert to Character", then the column moves to the end of the data frame, rather than stay in the position where it initially was. (I assume this happens elsewhere, for example, "Prepare > Data Frame > Convert Columns").
3. This does not occur everywhere, however. E.g., Prepare > Calculate > Duplicate Column we want the new column to be placed next to a current column (i.e., not necessarily at the end of the data frame) and this does as such.

This seems to cause a bug if I create a new column (where the new column is not placed at the end of the data frame) and open a dialog. Using the "diamonds" data set in the `ggplot2` library:
a) Go to Prepare > Calculate > Duplicate Column and put in "cut" as the column to duplicate. Press OK
b) Open the "Non-Numeric Cases" dialog

The following error comes up:

image

For me, this bug also occurred for other data sets I tried out, and for any dialog I tried to open in (b), however, I'm being specific here in case you try a data set/dialog which it happens to not occur with.

Patowhiz:
@lilyclements thank you for testing this.

I have solved your item (1). I had not linked the new save control to the receiver.
I have also solved item (2). 

@rdstern @dannyparsons  to solve the item (2) raised by @lilyclements  above. I had to fix issue #6042 also. So @rdstern this PR doesn't have the bug. You can now also test this.

rdstern (approved):
Great.  Please can @lloyddewit and perhaps also @dannyparsons check the code.  This could have solved the serious bug in 0.6.3

instat/ucrSave.vb
```diff
@@ -882,7 +882,7 @@ Public Class ucrSave
     ''' <summary>   Updates the control's R command's column position variables.
     '''             These variables specify the position of the new column in the data frame 
     '''             (start, end or before/after a specified column).<para>
-    '''             This subroutine is called automatically when the receiver (specifying a column name) linked to 
+    '''             This subroutine is called automatically when the receiver (specifying a column name) linked to clsColPosFunction
     '''             this control changes. </para><para>
```
lloyddewit, line 886:
```suggestion
    '''             This subroutine is called automatically when the receiver (specifying a column name) linked to 
    '''             clsColPosFunction changes. </para><para>
```

instat/clsRCodeStructure.vb
```diff
@@ -410,19 +410,9 @@ Public Class RCodeStructure
...
                     clsAddColumns.AddParameter("adjacent_column", strAdjacentColumn)
-                Else
-                    If bInsertColumnBefore Then
-                        clsAddColumns.AddParameter("before", "TRUE")
-                    Else
-                        If frmMain.clsInstatOptions.bIncludeRDefaultParameters Then
-                            clsAddColumns.AddParameter("before", "FALSE")
-                        End If
-
-                    End If
```
lloyddewit, line 425: I understand that:
- old code: add `before` parameter unless (`strAdjacentColumn` specified AND `before` false AND don't include default params)
- new code: always add `before` parameter

Therfore the only difference is related to including default parameters (visually different but functionally the same). Please could you clarify the purpose of this change?

lloyddewit (changes requested):
@Patowhiz thanks, looks good, just a couple of small questions

instat/clsRCodeStructure.vb
Patowhiz: It's actually also functionally different at the R backend level. 

Basically, by not defining the before parameter, at the backend level, the function responsible for for adding or replacing columns will not be able to know where to position the new or replaced column. So it will just add it but not position it.

Before my new changes, replaced columns could not be repositioned. To extend this functionality but at the same time not affect other backend R functions that use the add column R function, I had to change how this parameter is being used in the add column R function block.

lloyddewit: In the R backend, the `before` and `adjacent_column` parameters are used in the `add_columns_to_data` function. In this function, the `before` parameter is `FALSE` by default (see definition below from `data_object_R6.R` line 496). The change above explicitly forces the `before` parameter to be set even when it is `FALSE` (but it's anyway `FALSE` by default). Therefore I still don't see the purpose of the change above. Am I missing something?
Could you test with and without the above change to clarify if it makes any difference? thanks
```
function(col_name = "", col_data, use_col_name_as_prefix = FALSE, hidden = FALSE, before = FALSE, adjacent_column, num_cols, require_correct_length = TRUE)
```

Patowhiz: @lloyddewit I have changed `add_columns_to_data` function in this PR. It is now as shown below;
`function(col_name = "", col_data, use_col_name_as_prefix = FALSE, hidden = FALSE, before, adjacent_column, num_cols, require_correct_length = TRUE) `

Check line 496 in file `data_object_R6.R` in this PR. Look at the green line. Below is also a screenshot.

code_difference

lloyddewit: Yes, you're correct, I didn't notice the change on line 496, sorry for that. I'll approve.
Out of interest, what was the reason for removing the default parameter? is it to force the developer to be specific?

lloyddewit (approved):

lloyddewit:
@rdstern  since your last approval, only a comment changed. Please could you re-approve? thanks

rdstern (approved):

instat/clsRCodeStructure.vb
Patowhiz: @lloydewit 
Yes, if the developer doesn't specify any value the function assumes no positioning is required. Which technically is a good assumption.  
Before the change, there was no way of knowing whether positioning of the column added or replaced is needed(which is understandable, replaced columnscould not be repositioned). The alternative to solving this would have been adding another parameter for this,  which I think would have been redundant.

instat/dlgFindNonnumericValues.vb
```diff
@@ -51,9 +52,15 @@ Public Class dlgFindNonnumericValues
         ucrReceiverColumn.bWithQuotes = False
         ucrReceiverColumn.SetIncludedDataTypes({"character"})
         ucrReceiverColumn.strSelectorHeading = "characters"
+        'makes the ucrSave control to position new column after selected column
+        ucrSaveColumn.setLinkedReceiver(ucrReceiverColumn)
```
Patowhiz, line 56: @rdstern  below is my response in regards to the comment you made about making dialogs automatically create new columns after selected variable
@lilyclements for dialogs that already have the ucrSave control, a code line like this is the only change required to make the new column created by the dialog be automatically position after the selected column/variable. See the comment above the code.
ucrReceiverColumn.strSelectorHeading = "characters"

'makes the ucrSave control to position new column after selected column
ucrSaveColumn.setLinkedReceiver(ucrReceiverColumn).

PR #6047, opened 2020-10-27
Define Surv Object dialog

### LINKED ISSUE #6007: Define dialog for survival data (in the structured menu)
The new "Structured" menu has an option for Survival data. This is an idea for the Define Survival Object dialog for this menu where a `surv` object is created.

The function is `Surv` in the `survival` package. It has five parameters which are given below and then listed in more detail:
* `time`
* `time2`
* `event`
* `type`
* `origin`

The dialog is centred around the `type` parameter which relates to the type of censoring the data has. It takes six values: `right`, `left`, `interval`, `interval2`, `counting`, or `mstate` (multistate). As such, these six options can be in buttons across the top of the dialog. The `right` option should be the default (Note: this is not always the default in R).

If `right` or `left` censoring are chosen, then the `time` variable is the time that an individual left the study
If another censoring is chosen, the `time` variable is the time that an individual joined the study; `time2` is the time that they left the study.
The dialog can have two receivers therefore - "Entry Time" and "Exit Time". The first "Entry Time" receiver is only visible when `interval`, `interval2`, `counting`, or `mstate` are selected and takes the `time` parameter.
The second "Exit Time" receiver is always visible. It takes the `time` parameter if `right` or `left` are selected, otherwise takes the `time2` parameter.
"Exit Time" must always be visible and filled*
"Entry Time" must always be filled* when it is visible (that is, for `counting`, `interval`, `interval2`, `mstate`).
_*`mstate` can take just the `time` parameter. I need to confirm further whether that corresponds to the Entry Time or Exit Time if only one time variable is given for `mstate`._

There is a third receiver which corresponds to the `event` parameter. This takes a variable corresponding to the status for an individual. This is always visible, except for when `interval2` is selected since this option does not take an event parameter.
This receiver can be empty if `mstate` is selected. Otherwise, it must be filled (except in the case of `interval2` where it is not visible).
This receiver should take numerical variables. If `counting` or `mstate` are chosen, it can additionally take a factor variable.

The `origin` parameter is only visible if `counting` is selected. This is a nud and takes a default of `0`.

Finally, there should be an option to specify which value in the `event` variable corresponds to dead/alive. This can be offered with a checkbox which links to a factor grid when the `event` variable is a factor. Otherwise, it can link to a ucrInput where the user can input the value that denotes an event has occurred.

lilyclements:
Fixes #6007 

This dialog will go in the `Structured` menu under `Survival > Define`. It creates a `Surv` object used in survival analysis. This object is the response in (most) functions in the `survival` package, and hence is used throughout this survival sub-menu.
As a result, this would be used with survival data. There are lots of survival data sets given in the `survival` package, which can be accessed through the "Import from Library" dialog.


lloyddewit:

shadrackkibet:
First I comment on the design. 
- The name of this dialog is **Create Survival Object**. I suggest renaming this to either **Define Survival Object** or **Define Survival Data**.
- The name of the save control should have "name"? So I suggest something like **Survival Object Name:** or **New Survival Object Name:** we often use **Column Name:** or **New Data frame Name:**
- Combine **Modify Event** checkbox label with input label.  So I suggest the label of the checkbox to be **Modify event when the value is:**, then align the checkbox with receivers, below it the input textbox, then the origin input box. This means you don't have to adjust the width of the dialog. However, the height will increase a bit.

lilyclements:
Thanks for having a look and for the suggestions. I have sorted the second and third points.
To the first, the name of this dialog was conceptualised a couple of years ago when there was going to be both a define dialog and a dialog creating the object. The plan is to only have this dialog now so I am open to changing the name to **Define Survival Object.** However, in other define dialogs, my understanding it that defines the current data and changes the metadata rather than adds a new object or column to the data set (at least this my understanding for Climatic and Procurement. With Time Series, if we use `as_tsibble`, I suppose it replaces the data with time-series defined data. I am not sure about the other Define dialogs). Here, we are not defining our current columns or replacing the data with defined-data, but creating a new object to the data. Do you think that using **Define** in the same is suitable given this? Happy to change it if so.

shadrackkibet:
I do not strongly feel that **Define** is really appropriate compared to **Create**. I just thought **Define** is consistent. Based on your explanation I don't think we should change it.  Let's see what others think. Thanks.

shadrackkibet:
I earlier suggested the input textbox below the checkbox. Then reduce the width of the dialog. There is too much empty space on the side caused by the input textbox. See below.
image

lilyclements:
Will sort this shortly, apologies I missed that comment before. The reason for this space is that a factor can be inputted as the event for "Mstate" and "Counting". In this instance, a factor grid comes up instead of an input box, where you select the levels. Should this factor grid still go on the right, where the dialog increases in size if this option is chosen?

shadrackkibet:
> Will sort this shortly, apologies I missed that comment before. The reason for this space is that a factor can be inputted as the event for "Mstate" and "Counting". In this instance, a factor grid comes up instead of an input box, where you select the levels. Should this factor grid still go on the right, where the dialog increases in size if this option is chosen?

Yes.

shadrackkibet:
Comment on the R code. Below is an example code produced using **Right** option at the top.
```
# Code generated by the dialog, Create Survival Object
cancer <- data_book$get_data_frame(data_name="cancer")
time <- data_book$get_columns_from_data(data_name="cancer", col_names="time")
status <- data_book$get_columns_from_data(data_name="cancer", col_names="status")
surv <- with(data=cancer, exp=survival::Surv(time=time, event=status))
data_book$add_surv(surv_name="surv", surv=surv, data_name="cancer")

rm(list=c("surv", "cancer", "time", "status"))
```
Because data is passed into the `with()` function line 2 and 3 are not needed. So the receivers should not pass parameters as Rfunctions. Below is an example of how code should look like.

```
# Code generated by the dialog, Create Survival Object
cancer <- data_book$get_data_frame(data_name="cancer")
surv <- with(data=cancer, exp=survival::Surv(time=time, event=status))
data_book$add_surv(surv_name="surv", surv=surv, data_name="cancer")

rm(list=c("surv", "cancer"))
```

lilyclements:
Good point! I have sorted this now, along with the design changes for the "Modify" checkbox.

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+        clsModifyOperation.AddParameter(clsRFunctionParameter:=clsCFunction, iPosition:=1)
+
+        clsDefaultFunction.SetRCommand("with")
+        clsDefaultFunction.AddParameter("exp", clsRFunctionParameter:=clsRightLeftFunction)
+        ' TODO fix up the set assign to as currently it is always running surv()
+        clsDefaultFunction.SetAssignTo(strTemp:="surv", strTempDataframe:=ucrSelectorFitObject.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempSurv:="surv", bAssignToIsPrefix:=True)
+
+        ucrBase.clsRsyntax.SetBaseRFunction(clsDefaultFunction)
+
+        SetBaseRFunction()
```
shadrackkibet, line 172: Why is this needed here? Should only run when the state of the panel and the receiver changes.

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+        ElseIf rdoMstate.Checked Then
+            If ucrReceiverEntry.IsEmpty() Then
+                ' if no entry time, then mstate reads the "Exit Time" receiver as time.
+                ' if there is entry time, mstate reads the "Exit Time" receiver as time2; "Entry Time" receiver as time.
+                clsDefaultFunction.AddParameter("exp", clsRFunctionParameter:=clsRightLeftFunction)
+            Else
+                clsDefaultFunction.AddParameter("exp", clsRFunctionParameter:=clsStartEndFunction)
+            End If
+        Else
+            clsDefaultFunction.AddParameter("exp", clsRFunctionParameter:=clsInterval2Function)
```
shadrackkibet, line 249: Add parameter positions.

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+        End If
+    End Sub
+
+    Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
+        SetDefaults()
+        SetRCodeforControls(True)
+        TestOkEnabled()
+    End Sub
+
+    Private Sub SetBaseRFunction()
```
shadrackkibet, line 233: The name of this sub is misleading. I see `exp` parameter is added, the base function is not changed! Consider changing the name of this sub.

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+                    clsModifyOperation.AddParameter("factor_value", ucrModifyEventFactor.GetSelectedLevels(), bIncludeArgumentName:=False, iPosition:=1)
+
+                    ucrModifyEventNumeric.Visible = False
+                    ucrModifyEventFactor.Visible = True
+                    ucrModifyEventLogical.Visible = False
+
+                Else
+                    Me.Size = New System.Drawing.Size(458, Me.Height)
+                    clsModifyOperation.RemoveParameterByName("factor_value")
+                    clsModifyOperation.AddParameter(clsRFunctionParameter:=clsCFunction, iPosition:=1)
```
shadrackkibet, line 296: Is this correct? parameter name?

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+
+                    If ucrReceiverEvent.strCurrDataType = "logical" Then
+                        clsCFunction.RemoveParameterByName("x")
+                        clsCFunction.AddParameter("z", ucrModifyEventLogical.GetText(), bIncludeArgumentName:=False)
```
shadrackkibet, line 300: Add position?

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+                        ucrModifyEventNumeric.Visible = False
+                        ucrModifyEventFactor.Visible = False
+                        ucrModifyEventLogical.Visible = True
+
+                    Else '(if numeric or integer)
+                        clsCFunction.RemoveParameterByName("z")
+                        clsCFunction.AddParameter("x", ucrModifyEventNumeric.GetText(), bIncludeArgumentName:=False)
```
shadrackkibet, line 307: Position

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+            ucrChkModifyEvent.Visible = False
+            ucrModifyEventNumeric.Visible = False
+            ucrModifyEventFactor.Visible = False
+            ucrModifyEventLogical.Visible = False
+        End If
+    End Sub
+
+    Private Sub ucrPnl_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrPnlType.ControlContentsChanged
+        If rdoRight.Checked OrElse rdoLeft.Checked Then
+            ucrReceiverEvent.RemoveIncludedMetadataProperty("class")
```
shadrackkibet, line 335: Do this outside the `if` statement? I see it is needed in all conditions.

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+                ucrReceiverEvent.Clear()
+            End If
+        End If
+
+        ' if the variable in the event receiver is logical and you click rdoInterval, clear it.
+        If ucrReceiverEvent.strCurrDataType = "logical" Then
+            If rdoInterval.Checked Then
+                ucrReceiverEvent.Clear()
+            End If
+        End If
```
shadrackkibet, line 357: Combine this into one `if` statement?

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+        SetReceivers()
+        ModifyOptions()
+        SetBaseRFunction()
+        TestOkEnabled()
+    End Sub
+
+    Private Sub ucrSelectorFitObject_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrSelectorFitObject.ControlContentsChanged
```
shadrackkibet, line 364: Use `.ControlValueChanged` event.

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+                clsDefaultFunction.AddParameter("exp", clsRFunctionParameter:=clsRightLeftFunction)
+            Else
+                clsDefaultFunction.AddParameter("exp", clsRFunctionParameter:=clsStartEndFunction)
+            End If
+        Else
+            clsDefaultFunction.AddParameter("exp", clsRFunctionParameter:=clsInterval2Function)
+        End If
+    End Sub
+
+    Private Sub SetReceivers()
```
shadrackkibet, line 253:
```suggestion
    Private Sub SetCurrentReceiver()
```

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+        If ucrSelectorFitObject.CurrentReceiver.Name Is "ucrReceiverEvent" Then
```
shadrackkibet, line 254: Using the receiver name here is dangerous. If a developer changes the name of the receiver then this will not work. Just use the current receiver.

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+            If (rdoInterval2.Checked) Then
+                ucrReceiverExit.SetMeAsReceiver()
+            End If
+        ElseIf ucrSelectorFitObject.CurrentReceiver.Name Is "ucrReceiverEntry" Then
+            If (rdoRight.Checked OrElse rdoLeft.Checked) Then
+                ucrReceiverExit.SetMeAsReceiver()
+            End If
```
shadrackkibet, line 261: Combine this using `AndAlso`, also above?

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+
+    Private Sub ucrChkModifyEvent_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrChkModifyEvent.ControlContentsChanged, ucrReceiverEvent.ControlContentsChanged
+        TestOkEnabled()
+        ModifyOptions()
+    End Sub
+
+    Private Sub ucrReceiverEntryControl(ucrChangedControl As ucrCore) Handles ucrReceiverEntry.ControlContentsChanged
+        SetBaseRFunction()
+        TestOkEnabled()
+    End Sub
```
shadrackkibet, line 380: Use `.ControlValueChanged` event here. Also `TeskOkEnabled()` is not needed here move it into the last sub.

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,373 @@
...
+
+    Private Sub ucrModifyEventFactor_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrModifyEventFactor.ControlValueChanged, ucrModifyEventLogical.ControlValueChanged
+        ModifyOptions()
+    End Sub
+
+    Private Sub ucrChkModifyEvent_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrChkModifyEvent.ControlContentsChanged, ucrReceiverEvent.ControlContentsChanged
+        TestOkEnabled()
+        ModifyOptions()
+    End Sub
```
shadrackkibet, line 375: Use `.ControlValueChanged` event here. Also `TestOkEnabled()` is not needed.  Move `.ControlContentsChanged` event into the last sub. Then combine this with the sub above(368).

shadrackkibet (changes requested):

instat/dlgSurvivalObject.vb
lilyclements: Will do. Out of interest, when should I use `ControlContentsChanged`, and when should I use `ControlValueChanged`?

instat/dlgSurvivalObject.vb
lilyclements: I thought so - what is the problem (sorry)?

instat/dlgSurvivalObject.vb
shadrackkibet: Only use `.ControlContentsChanged` for TestOk

instat/dlgSurvivalObject.vb
shadrackkibet: I see this is an operation anyway. But I would still give it a parameter name and hide if required.

lilyclements: Thanks for clarifying, sorted it now.

lilyclements:
@shadrackkibet thank you for the review, I have implemented the changes you outlined.

instat/dlgSurvivalObject.vb
shadrackkibet: Any change on this?

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,356 @@
...
+            clsRightLeftFunction.AddParameter("event", clsRFunctionParameter:=ucrReceiverEvent.GetVariables, iPosition:=2)
+            clsStartEndFunction.AddParameter("event", clsRFunctionParameter:=ucrReceiverEvent.GetVariables, iPosition:=2)
+            ucrChkModifyEvent.Visible = False
+            ucrModifyEventNumeric.Visible = False
+            ucrModifyEventFactor.Visible = False
+            ucrModifyEventLogical.Visible = False
+        End If
+    End Sub
+
+    Private Sub ucrPnl_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrPnlType.ControlContentsChanged
```
shadrackkibet, line 326: Change this event and then move `TestOkEnabled` to the last sub. Sorry I didn't notice this in my previous review.

shadrackkibet (changes requested):

lilyclements:
@shadrackkibet I have made the changes

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,351 @@
...
+            If ucrChkModifyEvent.Checked Then
+                ' have to currently add these variables manually because the ucr that is added changed depending on the data type
+                ' and (to my knowledge) we can't yet do this in the automatic system
+                clsRightLeftFunction.AddParameter("event", clsROperatorParameter:=clsModifyOperation, iPosition:=2)
+                clsStartEndFunction.AddParameter("event", clsROperatorParameter:=clsModifyOperation, iPosition:=2)
+
+                If ucrReceiverEvent.strCurrDataType = "factor" Then
+                    Me.Size = New System.Drawing.Size(662, Me.Height)
+                    clsModifyOperation.ClearParameters()
+                    clsModifyOperation.AddParameter("event", clsRFunctionParameter:=ucrReceiverEvent.GetVariables, bIncludeArgumentName:=False, iPosition:=0)
```
shadrackkibet, line 279: Add this as an additional parameter pair

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,351 @@
...
+                        clsCFunction.AddParameter("x", ucrModifyEventNumeric.GetText(), bIncludeArgumentName:=False, iPosition:=0)
+                        ucrModifyEventNumeric.Visible = True
+                        ucrModifyEventFactor.Visible = False
+                        ucrModifyEventLogical.Visible = False
+                    End If
+                End If
+            Else
+                Me.Size = New System.Drawing.Size(458, Me.Height)
+                clsRightLeftFunction.AddParameter("event", clsRFunctionParameter:=ucrReceiverEvent.GetVariables, iPosition:=2)
+                clsStartEndFunction.AddParameter("event", clsRFunctionParameter:=ucrReceiverEvent.GetVariables, iPosition:=2)
```
shadrackkibet, line 309: Also here add additional parameter pair and remove this.

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,351 @@
...
+                ucrModifyEventNumeric.Visible = False
+                ucrModifyEventFactor.Visible = False
+                ucrModifyEventLogical.Visible = False
+            End If
+
+        Else 'if interval or interval2 checked
+            Me.Size = New System.Drawing.Size(458, Me.Height)
+            clsRightLeftFunction.AddParameter("event", clsRFunctionParameter:=ucrReceiverEvent.GetVariables, iPosition:=2)
+            clsStartEndFunction.AddParameter("event", clsRFunctionParameter:=ucrReceiverEvent.GetVariables, iPosition:=2)
```
shadrackkibet, line 318: Remove this

shadrackkibet (changes requested):
When testing this I notice that the Rcode is still not correct. Using the first option (Right). I got this.
```
# Code generated by the dialog, Create Survival Object
heart <- data_book$get_data_frame(data_name="heart")
event <- data_book$get_columns_from_data(data_name="heart", col_names="event")
surv <- with(data=heart, exp=survival::Surv(time=stop, event=event))
data_book$add_surv(surv_name="surv", surv=surv, data_name="heart")

rm(list=c("surv", "heart", "event"))
```

`event <- data_book$get_columns_from_data(data_name="heart", col_names="event")` is not needed. See my comments to fix this.

lilyclements:
@shadrackkibet thank you, I have made the changes outlined.

shadrackkibet:
I now get a bug when testing as explained below.

1) I first use the counting option as follows.
image
2) Then reopen the dialog and change top radio button to "Interval". On click Ok I get below error.
image

The code is as shown below, which shows that the operator (`%in% c(0)`) is not cleared. 
```
# Code generated by the dialog, Create Survival Object
kidney <- data_book$get_data_frame(data_name="kidney")
surv <- with(data=kidney, exp=survival::Surv(time=time, time2=time, event=status %in% c(0), type="interval"))
data_book$add_surv(surv_name="surv", surv=surv, data_name="kidney")

rm(list=c("surv", "kidney"))
```

Reopening the dialog and then click Ok, it works well.

lilyclements:
This links into a "fix" I did earlier:

* If the "Modify" checkbox is checked, then the `event` parameter in the `Surv` function equals the `%in%` operator.
* If the "Modify" checkbox is not checked, then the `event` parameter in the `Surv` function equals the value in the event receiver.

I am not sure how to do this in the `SetRCode` system, and cannot find an example of a dialog similar, so I have now reverted it back to the "manual" system. Is there a better way to do this?

shadrackkibet:
Happy with that for now. Thanks

lloyddewit:
@shadrackkibet Thank you for the review

shadrackkibet:
@lilyclements this is a general comment about the dialog. I think it is not clear to a user presumably who doesn't know about R when they use the dialog to know the dialog worked. The user always expects an output as a column, graph or some output into the output window which in this case it is not available. I, therefore, suggest a print message like what we do on define climatic indicating that a surv object has been created.

lilyclements:
@shadrackkibet that is a good point - I have added a comment in the output window. To follow on from your point, I am wondering if perhaps I should include a "View and Delete Survival Object" dialog after this dialog, like we have for keys and links. This can, however, already be achieved through the "Prepare > R Objects" menu so perhaps isn't necessary. What do you think?

shadrackkibet:
> @shadrackkibet that is a good point - I have added a comment in the output window. To follow on from your point, I am wondering if perhaps I should include a "View and Delete Survival Object" dialog after this dialog, like we have for keys and links. This can, however, already be achieved through the "Prepare > R Objects" menu so perhaps isn't necessary. What do you think?

 I think that depends on whether we want this to be "independent" sub-menu i.e do we want the user to be able to use the sub-menu without having to move elsewhere many times? If that is the case then it makes sense to add them into this sub-menu. Another factor to consider here is whether other sub-menus under structured menu will need the two dialogs, then we will need to think about repositioning them so that we don't have to add into each of the sub-menus. Perhaps it is too early to think about this now since other sub-menus are not yet working. Otherwise, I am happy for them to be added for now. Also, include Rename Objects? not sure about Reorder objects - it might not be useful here.

shadrackkibet:
- Can you amend the message to read: "col1,col2 column(s) are used to create the Survival Object: surv"? I think you just need to add a few words then have sep="," for `paste()`

- Then the prefix for save control doesn't work the save name is always "surv" is that deliberate? I think it should be surv1, surv2 etc when using it for the second time etc

- Then ensure column names are updated correctly on reopen. When I reopen the dialog and then click ok then not all column names are displayed in the output message.

lilyclements:
@shadrackkibet I have done 1. and 3, thanks!

The save name not being a prefix is deliberate (however, like a lot of this dialog, was initially planned a few years ago and so am very open to changing it). There is a small issue with naming the new surv object, but that goes deeper than in this dialog. I think it is to be sorted in the clsRCodeStructure.R file, and so is on my list to sort out (but on a separate branch to avoid conflicts).

lloyddewit:
@rdstern I think this PR is in a good state for your testing

shadrackkibet:
> @shadrackkibet I have done 1. and 3, thanks!
> 
> The save name not being a prefix is deliberate (however, like a lot of this dialog, was initially planned a few years ago and so am very open to changing it). There is a small issue with naming the new surv object, but that goes deeper than in this dialog. I think it is to be sorted in the clsRCodeStructure.R file, and so is on my list to sort out (but on a separate branch to avoid conflicts).

This is a serious bug. It should be fixed soon before this PR is merged, otherwise, the message is misleading to the user. I did have a quick look into how to fix this. The problem is actually with the incomplete implementation of how the **surv** object is handled.

- In `ucrInput.vb` file the option to set the default name for the **surv** object is missing. Check the method in line 150. **surv** should be among other options.
- There is a problem with the method used to set the next default item for the **surv** object. This links to the method for getting **surv** names in the `clsRlink.vb` file. See the function in line 1906, notice that a wrong function, I think caused by copy-pasting is used here in line 1912 i.e `clsGetSurvNames.SetRCommand(strInstatDataObject & "$get_graph_names")` it should be `       clsGetSurvNames.SetRCommand(strInstatDataObject & "$get_surv_names")` which is a method already implemented.. 

I think these are among places that need fixing for the prefix to work.

lilyclements:
> * In `ucrInput.vb` file the option to set the default name for the **surv** object is missing. Check the method in line 150. **surv** should be among other options.
> * There is a problem with the method used to set the next default item for the **surv** object. This links to the method for getting **surv** names in the `clsRlink.vb` file. See the function in line 1906, notice that a wrong function, I think caused by copy-pasting is used in line 1912 i.e `clsGetSurvNames.SetRCommand(strInstatDataObject & "$get_graph_names")` it should be `       clsGetSurvNames.SetRCommand(strInstatDataObject & "$get_surv_names")` which is a method already implemented..
> 
> I think these are among places that need fixing for the prefix to work.

Thanks for this @shadrackkibet - I have worked out how to fix it, which requires changing three files (the two you outlined, and `ucrSave.vb`). I initially made the changes on this branch so that I could test that it fixes it. Should I commit these changes to this PR, or should I open a new PR with only these three files edited to avoid conflicts?

shadrackkibet:
@lilyclements, I think it is easy to test it here. So I would prefer the changes done on the PR. Happy to test it separately if you would prefer it that way.

lilyclements:
@shadrackkibet I agree. Have committed and synced the changes.

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,383 @@
...
+                    Else
+                        ucrBase.OKEnabled(True)
+                    End If
+                End If
+            End If
+        End If
+    End Sub
+
+    Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
+        ucrBase.clsRsyntax.RemoveFromAfterCodes(clsCreateObjectScript)
```
shadrackkibet, line 241: Remove this and instead clear codes just before `SetBaseRFunction` in `SetDefaults()`

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,383 @@
...
+        clsCreateObjectScriptPaste.AddParameter("col1", strParameterValue:=ucrReceiverExit.GetVariableNames(), bIncludeArgumentName:=False, iPosition:=1)
+        clsCreateObjectScript.AddParameter("surv_name", Chr(34) & ucrSaveObject.ucrInputComboSave.GetText() & Chr(34), iPosition:=2, bIncludeArgumentName:=False)
+
+        If Not ucrReceiverEntry.IsEmpty AndAlso ucrReceiverEntry.Visible Then
+            clsCreateObjectScriptPaste.AddParameter("col0", strParameterValue:=ucrReceiverEntry.GetVariableNames(), bIncludeArgumentName:=False, iPosition:=0)
+        Else
+            clsCreateObjectScriptPaste.RemoveParameterByName("col0")
+        End If
+
+        If ucrReceiverEvent.Visible Then
```
shadrackkibet, line 373: Add `.IsEmpty` condition just like above?

instat/ucrInput.vb
```diff
@@ -184,6 +184,12 @@ Public Class ucrInput
                 Else
                     SetName("")
                 End If
+            ElseIf strDefaultType = "Surv" Then
+                If ucrDataFrameSelector IsNot Nothing AndAlso ucrDataFrameSelector.cboAvailableDataFrames.Text <> "" Then
+                    SetName(frmMain.clsRLink.GetNextDefault(strDefaultPrefix, frmMain.clsRLink.GetSurvNames(ucrDataFrameSelector.cboAvailableDataFrames.Text)))
+                    'Else
+                    '    SetName("")
```
shadrackkibet, line 191: Uncomment this, for now, I think it is sensible to have it, although I claim this should be a developer error.

instat/dlgSurvivalObject.vb
```diff
@@ -13,6 +13,383 @@
...
+        clsInterval2Function.SetRCommand("Surv")
+
+        clsCFunction.SetRCommand("c")
+
+        clsModifyOperation.SetOperation("%in%")
+        clsModifyOperation.AddParameter("c_function", clsRFunctionParameter:=clsCFunction, iPosition:=1)
+
+        clsDefaultFunction.SetRCommand("with")
+        clsDefaultFunction.AddParameter("exp", clsRFunctionParameter:=clsRightLeftFunction, iPosition:=1)
+        clsDefaultFunction.SetAssignTo(ucrSaveObject.GetText, strTempDataframe:=ucrSelectorFitObject.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempSurv:=ucrSaveObject.GetText)
```
shadrackkibet, line 170: Remove this it is not needed.

shadrackkibet (changes requested):

shadrackkibet:
@lilyclements I have a few comments above otherwise, this is now working fine. Thanks for working on my many reviews, I must admit now that I feel this is becoming too much at this stage. I suggest that @rdstern and @lloyddewit can now proceed on testing and reviewing this.

instat/dlgSurvivalObject.vb
lilyclements: Whenever this receiver is visible, it needs to be filled (unlike the receiver above).

instat/dlgSurvivalObject.vb
lilyclements: Theres an issue if I click reset that this bit of code runs twice. The `RemoveFromAfterCodes` helped with this, whereas this `Clear` doesn't fix this.

shadrackkibet: This should work because on click reset `SetDefaults()` is run which will clear codes.

instat/dlgSurvivalObject.vb
shadrackkibet: Sorry, perhaps I didn't understand what you mean. My understanding is that the parameter below should be added only if the receiver is not empty.

instat/dlgSurvivalObject.vb
lilyclements: Running the `.Clear` still meant the code would run twice if I clicked reset though

shadrackkibet: I mean `ucrBase.clsRsyntax.ClearCodes()` not `clsCreateObjectScript.Clear()`

instat/dlgSurvivalObject.vb
lilyclements: If I do just the `.isEmpty` argument then it adds `col2` provided it is not empty.
This is a problem if I fill it on, say, "Right" then click on "Interval2" (where the Event receiver is not visible). The receiver may not be visible on "Interval2", but it is still filled. Hence, the output code still says that `col2` is used to create the survival object when it is not.

So I could do `Not ....isEmpty AndAlso ....isVisible` but that is the same thing as running `....isVisible`. This is because for the other radio buttons (ie, not "Interval2"), where this Event receiver is visible, the Event receiver has to be filled for OK to be enabled.

instat/dlgSurvivalObject.vb
lilyclements: My apologies! Sorted that now. What is the benefit to this over what I was doing?

shadrackkibet: It just for simplicity and consistency. This will ensure that any other after/before codes added will be cleared. No need to remove one at a time on reset.

shadrackkibet: Clearing a function e.g `clsCreateObjectScript.Clear()` doesn't mean you clear it from Rsyntax i.e running on click ok. It basically means you are clearing some components of the function eg Setting Rcommand to "", R package to "" and clearing other components in the R code structure like parameters. It can be confused to declaring a new instance i.e `New RFunction`

lloyddewit:
@rdstern Please could you review? thanks

lloyddewit:
@rdstern Please could you review? thanks

rdstern (commented):
@lilyclements or @shadrackkibet I musr be doing something wrong as I see the Define, but no dialogue opens for me?

shadrackkibet:
That is because the dialog is not yet added into the main form. @lilyclements are you in a position to add it?

lilyclements:
I can add it. Shall I add it in this PR or create a new branch?

shadrackkibet:
A new branch with reference to this PR.

rdstern (changes requested):
@lilyclements I still can't see this dialogue.  I get to define and then nothing.  Is there something I should do?
Once I can, then what data file should I try to test it?

lilyclements:
@rdstern sorry for that - can you try it again now?

Data file wise I have been using some in the survival library. I think the CDC data set both start and end times to try other censoring options. Otherwise I have been using the lung dataset.

rdstern:
This is looking like a nice dialogue.  It would be good to see it merged so it can be used and built on.  
1) I was confused by the checkbox "Modify Event when Value is"  It finishes at the end of the dialogue, so I assumed something must be missing.  When I clicked, however it opens another field.  What's typed here?  It looks the size of an ordinary receiver, but I assume it just wants a number.  Can it be anything or would it be usually a small integer?  Or might it be TRUE or FALSE?  So I wonder if an up/down if integer, or a drop-down with say TRUE/FALSE/0,1,2,3.  It could still allow typing in if those values were not what was to be modified.  And could the label be something like "Value to Modify:" or "Event Value Modified:"
2) I don't want to delay merging, but could tooltips easily be given on some of the controls?  They might change however, depending on the top button.  .

lilyclements:
@rdstern thank you for your review.

To 1. - I have changed the checkbox to be a checkbox and label now to clear it up a bit.
The value to type in is the value that signifies an event has occurred for the event variable (so is default "1" in the Surv object for right censoring).
For some censoring types, the event variable can only be an numeric variable or a logical variable. If a numerical variable is in the "Event" receiver, an input box comes up where you write in the numeric values. Otherwise if it is logical, a combo box comes up with the options "TRUE"/"FALSE". I can change this if you think so?
For some censoring types (e.g. Multistate) a factor variable can be in the event receiver. If the user wishes to specify which factor levels signify an event has occurred, they can check the checkbox and a factor level grid comes up.

To 2 - I can add in tooltips. Where are you thinking they would be helpful?

rdstern (approved):
@lilyclements I am approving, because this looks fine now.  For the record I include details that could be added:
1. The buttons at the top could have a label Type: 
2.  When I click on Modify and can give the number to modify by.  I used the lung data and tried oddities that it didn't allow.  It also didn't allow TRUE or FALSE, which I assume may be needed sometimes.  Could there be a pull-down with 0, 1, 2, TRUE and FALSE as the suggestions.  It still allows typing in case of other values.
3. Good to have the summary in the output window.  Good to have even more, so next line is Type of Censoring:  Right, Then Censored when status = FALSE
All optional, before merging!

lilyclements:
Thank you for the review @rdstern. This PR makes changes in some more general files (`clsRlink`, `ucrSave`, and `ucrInput`) as well as the survival dialogs so I've put the changes you've recommended in an issue (#6165) and I will work on these after this PR is merged. This is to avoid conflicts in case someone else wants to edit those more general files while I'm making the changes on the survival dialogs.

rdstern (approved):
@lloyddewit and @lilyclements I am approving, because this now works nicely.  I have made some suggestions for cosmetic changes which are accepted, and will be added as a separate issue for later.

lloyddewit (approved):

Issue #6008, opened 2020-10-07
At least a minor change on the visdat dialogue


rdstern:
This is working well now.  Could we make a very minor change at least and perhaps add some further options.

a) The very minor change is that currently on Visualising data the default is sort = TRUE, which I find confusing, and would strongly prefer sort = FALSE.  (I use that in a proposed video and have to make the change in the script window)
b) Better would be to add some of the options, now that we have a working dialogue.   Could we have a checkbox called **_Sort Variables_** for this option?  Default is unticked (which is sort = FALSE).  I think this sort option may apply to the Missing option, but not to the Guess.
c) Perhaps, at the same time we could consider other options.  In particular add Palette for the Data and Guess.  It doesn't seem to be there for Missing.  Perhaps it could be 3 (ordinary) radio buttons.
d) Should we also have Maximum Size? If so, then perhaps that could be an updown with 9 as the minimum and going up one at a time.  It says Maximum Data Size 9 hundred-thousand bytes.  I assume it is bytes?

I hope @lilyclements could look at this and comment quickly, and perhaps specify in more detail.  Then (unless she wants to make the changes) I suggest @Wycklife could do this.  If this will take some time, then please could @Wycklife make the change in item a.

lilyclements:
To respond to the end of this issue - I will look at this and comment by the end of the week.

lilyclements:
b) There are two `sort_...` functions
`sort_type` applies to `vis_dat` and is default `TRUE`. Is this the one you want in, and for the default to be `FALSE`?
There is also `sort_miss` in the `vis_miss` function which sorts the columns by missing to non-missing. This is default `FALSE`. Do you want this one in as well?

d) The `large_data_size` argument corresponds to the number of cells in a data frame.
For example, the airquality data has 153 rows and 6 variables (918 cells):
```
# airquality has 918 cells. If we set the large_data_size to be equal to or greater than this, then the function runs
vis_miss(airquality, large_data_size = 918)` 

# however, if we set the large_data_size to be less than the number of cells, there is an error.
vis_miss(airquality, large_data_size = 917)
Error in vis_miss(airquality, large_data_size = 917) : 
  Data exceeds recommended size for visualisation, please consider
         downsampling your data, or set argument 'warn_large_data' to FALSE.
```

[...]

PR #6051, opened 2020-10-29
Added slice_sample function in the visdat dialog

Wycklife:
Fixes part (e) of #6008 

@africanmathsinitiative/developers This will be reviewed by @lilyclements


instat/dlgVisualizeData.vb
```diff
@@ -231,4 +235,34 @@ Public Class dlgVisualizeData
...
+                clsVisMissFunction.AddParameter("x", clsRFunctionParameter:=clsSamplingFraction, iPosition:=0)
+            End If
+        ElseIf rdoWholeDataFrame.Checked Then
+            clsVisDatFunction.RemoveParameterByName("x")
+            clsVisGuessFunction.RemoveParameterByName("x")
+            clsVisMissFunction.RemoveParameterByName("x")
+            If ucrNudSamplingFunction.Value = 1 Then
+                clsVisDatFunction.AddParameter("data", clsRFunctionParameter:=ucrSelectorVisualizeData.ucrAvailableDataFrames.clsCurrDataFrame, iPosition:=0)
+                clsVisGuessFunction.AddParameter("data", clsRFunctionParameter:=ucrSelectorVisualizeData.ucrAvailableDataFrames.clsCurrDataFrame, iPosition:=0)
+                clsVisMissFunction.AddParameter("data", clsRFunctionParameter:=ucrSelectorVisualizeData.ucrAvailableDataFrames.clsCurrDataFrame, iPosition:=0)
```
lilyclements, line 260: Add `bIncludeArgumentName:=False` for each of these `.AddParameter("data",...` arguments

lilyclements (commented):
Looks good, just a couple of comments:

1. Design wise can you level up the "Sampling Fraction" label and nud with the check box for "Sort Variables"?
2. As you highlighted to me elsewhere, the "Selected Variables" receiver clears upon reopening the dialog. I can have a better look at this tomorrow if you do not sooner?

instat/dlgVisualizeData.vb
Wycklife: @lilyclements I have fixed the two comments above.

instat/dlgVisualizeData.vb
```diff
@@ -231,4 +235,34 @@ Public Class dlgVisualizeData
     Private Sub ucrNudMaximumSize_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrNudMaximumSize.ControlValueChanged
         MaximumDataPoint()
     End Sub
+
+    Private Sub ucrNudSamplingFunction_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrNudSamplingFunction.ControlValueChanged, ucrReceiverVisualizeData.ControlValueChanged
```
lilyclements, line 239: You should add `ucrPnlSelectData.ControlValueChanged` here. 

At the moment, if I am on "Data" and select "Select Variable" and press OK, it works fine.
Then if I reopen, and click on "Data Frame" and press OK, the data argument does not read in

image

Adding `ucrPnl...` here solves this problem

lilyclements (commented):
Looks good. A couple of comments:

1. Click "Guess" (keep it on "Data Frame") on the first open and press OK; then on the reopen click "Data" (and keep it on "Data Frame"), the parameter value for the `sort_type` argument argument is not read in:
```
visdat::vis_dat(dodoma, sort_type=, palette="default", warn_large_data=TRUE, large_data_size=900000)
```

2. I'm unsure why the receiver clears upon reopen if the `nud < 1`. If I comment out line 249 (`clsVisDatFunction.AddParameter("x", clsRFunctionParameter:=clsSamplingFraction, iPosition:=0)`) then the issue does not occur for any of the three functions at the top. However, I'm unsure why this clears the receiver and so am unsure of the solution.

instat/dlgVisualizeData.vb
Wycklife: @lilyclements Adding ucrPnlSelectData.ControlValueChanged seems not to solve the problem. I think the bug above is related to the receiver clearing the variables when you reopen the dialog. So when you change the buttons, the parameter x is not available.

lloyddewit:
@lilyclements thank you for your review, please could you confirm that all your comments are resolved? thanks

instat/dlgVisualizeData.vb
lilyclements: Oh, my mistake. Nice fix!

lilyclements:
@lloyddewit @Wycklife Most of the comments are resolved with the exception of "2." - if the nud for selecting the fraction proportion is less than one and "Selected Variables" is used instead of "Data Frame", then the receiver is cleared upon reopening.

lloyddewit:
@lilyclements Thanks for the feedback
@Wycklife Please could you resolve the comment below?
> 
> 
> @lloyddewit @Wycklife Most of the comments are resolved with the exception of "2." - if the nud for selecting the fraction proportion is less than one and "Selected Variables" is used instead of "Data Frame", then the receiver is cleared upon reopening.

Wycklife:
@lloyddewit I have a discussion on this too

rdstern:
@Wycklife and @lilyclements I have been discussing this function with @dannyparsons .  
a) Just to confirm that the default is for the sampling fraction to be 1, i.e. no sample.
b) But more serious is the situation when you do sample.  It isn't a random sample, but a selection from the rows of the database.  So I hope that it keeps the sampled data in the same order.  That may not be so critical in general, but our data (at least in the climatic menu) are time series.  So if the sample is 0.5 (say) and the full rows are 1 to 10 then I would like either exactly 5 rows or about 5 rows (don't mind), but they could then be rows 1,4,5,7,9, which would be fine.  But if they were 4,1,9,7,5 then I would be unhappy!

So what is happening now?  If it is a random sample and also in random order, then I would prefer not to have it at all at least at this stage.  Perhaps @lilyclements could advise?

dannyparsons:
I've just had a check and it is a random sample, rows are in a random order. So I think this isn't what you want.

rdstern:
Why can't we generate a random TRUE/FALSE variable with probability `p < 1 `of TRUE.  Then just filter the rows to the TRUE values?  But not if it all takes a long time.

lilyclements:
Yes it is a random sample. I have had a little look and can't find a function which allows it to be randomly sampled in order. Your suggestion sounds good. So R code wise (if the nud is less than 1) can be something like

```
df <- data.frame(X = 1:1000)
prob <- 0.1

df_slice <- df %>% 
 mutate(random = rbinom(n = nrow(df), size = 1, prob = prob) %>%
 filter(random == 1)

vis_dat(df_slice, ...)
```

dannyparsons:
Simplified would be 

`vis_dat(df %>% filter(rbinom(n = nrow(df), size = 1, prob = prob) == 1), ...)` or
`vis_dat(df %>% filter(as.logical(rbinom(n = nrow(df), size = 1, prob = prob))), ...)`

Wycklife:
@dannyparsons , @lilyclements I have implemented the second part suggested by @dannyparsons. The code being produced is shown below.

set3 <- data_book$get_data_frame(data_name="set3")
last_graph <- visdat::vis_dat(set3 %>% dplyr::filter(.data=as.logical(x=rbinom(n=nrow(set3), size=1, prob=0.21))), sort_type=TRUE, palette="default", warn_large_data=TRUE, large_data_size=900000)
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="set3")
data_book$get_graphs(data_name="set3", graph_name="last_graph")
rm(list=c("last_graph", "set3"))

I don't understand the bug being produced  (no applicable method for 'filter_' applied to an object of class "logical"). Why is this error being produced?

Wycklife:
Another observation I have made is that when I removed the parameter .data and run the code in R, it seems to work.

Wycklife:
The first function vis_dat(df %>% filter(rbinom(n = nrow(df), size = 1, prob = prob) == 1), ...), seems to be fine, I'm now implementing it using this function.

Wycklife:
@dannyparsons Kindly help me fix the following;
1. When I check on the selected columns radio button and add the columns to the receiver, then close the dialog. On reopening, the receiver clears the selected columns and therefore empty. With the receiver being empty, change to the whole data frame option. There is a bug being produced.

2. The second problem is the one I have mentioned above. I don't understand why that error is being produced. Thank you.

lilyclements:
> I don't understand the bug being produced (no applicable method for 'filter_' applied to an object of class "logical"). Why is this error being produced?

The issue here is that the parameter value for `.data` should be a data frame. The `as.logical(...)` argument is not a data frame but instead what the data frame should be filtered to.
Since you call `set3 %>% filter(...)` you have already told the code that the data frame is `set3`, so do not need the `.data` argument.

E.g. the following two codes run the same code (note they do not give the same values since it randomly selects columns to filter to)

```
set3 <- airquality

# The data frame can be given at the start before %>% (the "pipe")
set3 %>% dplyr::filter(as.logical(x = rbinom(n = nrow(set3), size = 1, prob = 0.21)))

# Or you can call the data frame in through the .data parameter
dplyr::filter(.data = set3, as.logical(x = rbinom(n = nrow(set3), size = 1, prob = 0.21)))
```

This is why, as you found, removing the `.data` parameter fixed the issue.

Wycklife:
@lilyclements Thank you for replying back. I have removed that parameter and it's now working. The only problem now remaining is that of receiver clearing variables when you reopen the dialog.

lilyclements:
Great - I'll do a general review to see if there's anything else. Hopefully @dannyparsons can advise on the receiver clearing variables.

lilyclements:
@Wycklife I have reviewed this. If I put the fraction proportion less than one and check "Selected Variables", the `vis_...` code runs for the data set and not for the variables I've selected.

instat/dlgVisualizeData.vb
lilyclements: Coming back to this - surely ucrPnl should be run here since you want it to run when the rdo buttons change?

instat/dlgVisualizeData.vb
```diff
@@ -231,4 +258,34 @@ Public Class dlgVisualizeData
...
+            clsVisGuessFunction.RemoveParameterByName("data")
+            clsVisMissFunction.RemoveParameterByName("data")
+            If ucrNudSamplingFunction.Value = 1 Then
+                clsVisDatFunction.AddParameter("x", ucrReceiverVisualizeData.GetParameter.strArgumentValue, bIncludeArgumentName:=False, iPosition:=0)
+                clsVisGuessFunction.AddParameter("x", ucrReceiverVisualizeData.GetParameter.strArgumentValue, bIncludeArgumentName:=False, iPosition:=0)
+                clsVisMissFunction.AddParameter("x", ucrReceiverVisualizeData.GetParameter.strArgumentValue, bIncludeArgumentName:=False, iPosition:=0)
+            ElseIf ucrNudSamplingFunction.Value < 1 Then
+                clsVisDatFunction.AddParameter("x", clsROperatorParameter:=clsPipeOperator, iPosition:=0)
+                clsVisGuessFunction.AddParameter("x", clsROperatorParameter:=clsPipeOperator, iPosition:=0)
+                clsVisMissFunction.AddParameter("x", clsROperatorParameter:=clsPipeOperator, iPosition:=0)
```
lilyclements, line 274: You need to update the `clsPipeOperator` here such that the left hand side takes the columns from the data frame. Currently, `clsPipeOperator.AddParameter("x", clsRFunctionParameter:=ucrSelectorVisualizeData.ucrAvailableDataFrames.clsCurrDataFrame, bIncludeArgumentName:=False, iPosition:=0)` always runs, hence the selected columns are never read in, only the data frame.

instat/dlgVisualizeData.vb
```diff
@@ -231,4 +258,34 @@ Public Class dlgVisualizeData
     Private Sub ucrNudMaximumSize_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrNudMaximumSize.ControlValueChanged
         MaximumDataPoint()
     End Sub
+
+    Private Sub ucrNudSamplingFunction_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrNudSamplingFunction.ControlValueChanged, ucrReceiverVisualizeData.ControlValueChanged
```
lilyclements, line 262: Coming back to this - surely ucrPnl should be run here since you want it to run when the rdo buttons change?

lloyddewit:
@Wycklife please could you summarize what still needs to be done on this PR and if you still need input from anyone? thanks

PR #6053, opened 2020-10-30
Dlgfrequency

### LINKED ISSUE #5966: More circular plots from openair package
There are 6 more circular plots provided by the Openair package.
Percentile Rose
Frequency Rose
Polar Plot
Polar Cluster
Polar Annulus
Polar Frequency
These may become 6 dialogues or (say) 2 dialogues with 3 radio buttons in each or...?  From a quick look I fail to see the connections that would help in structuring the dialogues.  Once decided they should be easy to construct and code, because they have similar options to the Wind/Pollution Rose function.

This is work initially for @HawardKetoyoMsatsi with @Wycklife probably doing the coding then.

HawardKetoyoMsatsi:
Fixes #5966
Added controls to dlgPolarFrequency
Added a function polar_frequency to stand_alone_functions.R

### DIFF
--- a/instat/dlgPolarFrequency.Designer.vb
+++ b/instat/dlgPolarFrequency.Designer.vb
@@ -22,22 +22,187 @@ Partial Class dlgPolarFrequency
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()> _
     Private Sub InitializeComponent()
-        Me.UcrButtons1 = New instat.ucrButtons()
+        Me.ucrBase = New instat.ucrButtons()
+        Me.ucrSelectorPolarFrequency = New instat.ucrSelectorByDataFrameAddRemove()
+        Me.lblPollutant = New System.Windows.Forms.Label()
+        Me.lblColor = New System.Windows.Forms.Label()
+        Me.Label3 = New System.Windows.Forms.Label()
+        Me.Label4 = New System.Windows.Forms.Label()
+        Me.Label5 = New System.Windows.Forms.Label()
+        Me.Label6 = New System.Windows.Forms.Label()
+        Me.ucrReceiverPollutant = New instat.ucrReceiverSingle()
+        Me.ucrInputStatistic = New instat.ucrInputComboBox()
+        Me.ucrInputType = New instat.ucrInputComboBox()
+        Me.ucrInputPosition = New instat.ucrInputComboBox()
+        Me.ucrInputColor = New instat.ucrInputComboBox()
+        Me.ucrNudMinimumBins = New instat.ucrNud()
+        Me.ucrChkTransform = New instat.ucrCheck()
         Me.SuspendLayout()
         '
-        'UcrButtons1
+        'ucrBase
         '
-        Me.UcrButtons1.Location = New System.Drawing.Point(10, 329)
-        Me.UcrButtons1.Name = "UcrButtons1"
-        Me.UcrButtons1.Size = New System.Drawing.Size(410, 52)
-        Me.UcrButtons1.TabIndex = 0
+        Me.ucrBase.Location = New System.Drawing.Point(10, 329)
+        Me.ucrBase.Name = "ucrBase"
+        Me.ucrBase.Size = New System.Drawing.Size(410, 52)
+        Me.ucrBase.TabIndex = 0
+        '
+        'ucrSelectorPolarFrequency
+        '
+        Me.ucrSelectorPolarFrequency.bDropUnusedFilterLevels = False
+        Me.ucrSelectorPolarFrequency.bShowHiddenColumns = False
+        Me.ucrSelectorPolarFrequency.bUseCurrentFilter = True
+        Me.ucrSelectorPolarFrequency.Location = New System.Drawing.Point(14, 1)
+        Me.ucrSelectorPolarFrequency.Margin = New System.Windows.Forms.Padding(0)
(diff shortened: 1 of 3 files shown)


lloyddewit:
@shadrackkibet you seem to be reviewing other PRs from @HawardKetoyoMsatsi so please could you review this PR also?
(if you prefer another reviewer then let me know)

shadrackkibet:
Yes, I would prefer another reviewer on this.

lloyddewit:

Ivanluv:
@lloyddewit  the vb code looks fine from my end .

lloyddewit:
@HawardKetoyoMsatsi please could you resolve the file conflict before we proceed? thanks

HawardKetoyoMsatsi:
@lloyddewit the conflict is resolved, kindly take a look...

lloyddewit:

rdstern (commented):
There are some minor issues still, for example Ok is enabled when it shouldn't be.  But we also need to check on some more fundamental issues in these dialogues.  In our case we are delighted that these graphs extend to pollution variables, but I hope very much that they can also be used with just wind speed and wind direction as variations on the standard wind rose.  I may be wrong here, but we can't have pollutant as an essential element.  
Also we MUST test the function with datasets that don't have the standard names for the variables.  That is for ws, wd, no2, etc.  That's not the way we work in R-Instat.  I suggest we need to involve more people, but can start testing what is possible with the current dialogues and multiple datasets.  I will bring in @volloholic to check.

HawardKetoyoMsatsi:
Thanks, I will look into the issue. 
Something interesting I am getting from the openair package is that even thou code for these pollution roses functions can be written without specifying the pollutant, **nox** is supplied as the default pollutant, therefor pollutant (or any other variable that can be used in the place of pollutant) is technically a must add input. Same applies to variables like date, ws and wd where applicable.

rdstern:
@HawardKetoyoMsatsi when I look into Chapter 6 of the book on[ openair ](https://bookdown.org/david_carslaw/openair/sec-polarFreq.html) it seems clear that this dialogue can start with a frequency graph of wind speed and direction.   Then you can add the pollutant.  So it looks as though you could simply change your dialogue to have wind speed first and wind direction second and then Pollutant (Optional) as the third receiver.

Remember that the book is about pollution, whereas we partly want the dialogues to give flexible options for wind roses type graphs.  And our "system" is to define the data as climatic, where we keep account of the names of the wind speed and direction names, while openair fixes those variables as ws and wd.  
In the future we probably want to be able to also define pollutant variables as well in our define climatic data dialogue.

So, for each of these options we want (if possible) to have the first controls as wind speed and wind direction.  I am still of the opinion that most (if not all) these options can usefully become a single dialogue with option buttons at the top.  But start, for simplicity if you like with these separate dialogues.

And let's test the dialogues with a new version of the mydata, where we change each variable name from the defaults.

HawardKetoyoMsatsi:
@rdstern I have made the suggested changes, please take a look at the new state of the dialog. It is working fine with column headers changed.
I had initially put all the options in a single dialog but the code was too long and complicated that's why I resolved to having a dialog for each option.

rdstern:
@HawardKetoyoMsatsi I have only tried one and it is this frequency one.  It worked fine, but I noticed that when I just put in the wind speed and direction it still gave a frequency plot of NO2.

[mydata1.zip](https://github.com/africanmathsinitiative/R-Instat/files/5602629/mydata1.zip)

In the above file I have changed all the variable names and now it gives an error when I just use wind speed and direction.  Could you please test with this file, or make one yourself where all variable names are changed.

HawardKetoyoMsatsi:
I have gone through this and I realize polar frequency plots for all the pollutants seem alike when we use the default statistic "frequency" (even when I do it in R-Studio). However, when I use different statistic the differences in the plots is clearly visible. Something not interesting is the documentation state the default statistic to be "frequency" but when the  parameter "statistic" is not passed then statistic parameter used by default is "mean". @shadrackkibet  kindly try this out too.
@shadrackkibet please help me look at line 114 before we ask @rdstern to review the changes I made.

HawardKetoyoMsatsi:
@rdstern kindly have a look at this PR I have fixed the issues mentioned above and tested using the attached data.

rdstern:
@HawardKetoyoMsatsi this is looking good now.  WE are getting to the stage where we need to be able to check out the different things that are possible with this dialogue.
There are just two suggestions and one question I have first.  Then we should merge and assess.

a) If I put the same variable into with windspeed and the pollutant, then it gives an error.  Perhaps this could be trapped and not allowed?
b) The facets are limited to those that can be calculated.  I am sure it allows any factor and we should be able to include the Station as a factor.  I don't have a problem with the pull-down, but it should allow a factor from the data frame as an alternative.  If that is a bit complicated, then I would prefer just having a simple usual control as a single receiver.  Then, if we want to facet by the time variables we just have to get them ourselves.

c) I am still hoping that before long you will be happy to consider having the different plots on the same dialogue, with different radio-buttons at the top.  What do you think?  I suggest quite a lot of the controls will be the same and this might be more work now, but could be better in the long run?  However, I am keen to get them working in some way, so wouldn't want this change if it makes a very long delay till the dialogues are working.

lloyddewit:
@HawardKetoyoMsatsi if you close a PR, then please add a comment explaining why it was closed. If the closed PR is replaced by a new PR then please reference the closed PR in the new PR, thanks.

PR #6054, opened 2020-10-30
Removed exclusion of numeric data types - Column summaries

shadrackkibet:
@rdstern this PR removes the exclusion of numeric data types from the selector when Summarise By receiver is in focus. This is in response to a skype message. Please test.


rdstern (approved):
@shadrackkibet that is fine.
Please could @lloyddewit or @dannyparsons do the second approval, so it can be merged.

lloyddewit (approved):

PR #6061, opened 2020-11-03
Pasting and Copying Data

### LINKED ISSUE #5991: Pasting New Columns to Data Frame
As mentioned in issue #5619, a user may want to paste to a new column rather than an existing column.

It would be good to add this feature, currently there is no undo in pasting to existing columns, which overwrites existing data, this may be undesirable to the users .

### LINKED ISSUE #5992: Pasting New Data Frame
As mentioned in issue #5619, a user may want to paste to a new data frame as new sheet rather than to an data frame.

### LINKED ISSUE #5993: Copying data frame
Currently, the copy implementation creates a duplicate data frame rather than copying the whole data frame data.
A user may want to just copy the whole data frame to paste into a spreadsheet. This could be useful especially whenever a user wants to easily save several stages of the processed data frame to a separate file.

### LINKED ISSUE #6020: Typing/Replacing data in single cell in the grid
Currently the code implemented does validation at the VB.Net side.
As mentioned by @dannyparsons in this [comment](https://github.com/africanmathsinitiative/R-Instat/pull/5982#issuecomment-692847413) in PR #5982, this needs to be redone to use the same principles used in  PR #5982.

@dannyparsons comments
> The approach here is not quite what we want. You shouldn't do anything directly in the Data Grid as is done in `PasteFromClipboard` for example. The Data Grid is only ever updated from R, otherwise we are not in sync with the data in R. It's always a one-way process and that's fundamental to how R-Instat works.
> 
> The values to paste should be sent directly to R. This can actually be much simpler because there are R functions (`readClipboard()`) and an R package `clipr` (https://cran.r-project.org/web/packages/clipr/index.html and used by `rio`) which can directly read from the clipboard themselves, so we don't even need to process the clipboard values in VB before they go into R.
> 
> I think all we actually need is a method in the DataBook/DataSheet which will read from the clipboard and attempt to paste at a cell value given from the front end e.g. `data_book$paste_from_clipboard(data_name = "sheet1", row = 1, column = "date")`. All of the validation, checking and inserting into the data will then be done in this method.
> 
> This approach could also replace our current method for typing into a single cell since it's the same principle.

Patowhiz:
Fixes #5991 Pasting New Columns to Data Frame
Fixes #5992 Pasting New Data Frame
Fixes #6020 Typing/Replacing data in single cell in the grid
Fixes #5993 Copying data frame

The 3 PR's are related.

@rdstern @lloyddewit  this will be ready for review once PR #6052 and PR #5982 are merged. Some files in those PR need changes in this PR for it to be testable.


rdstern:
@Patowhiz should I be testing this now?  It was marked as draft?

Patowhiz:
@rdstern this will be ready for testing once PR #5982 is merged.

maxwellfundi:
@Patowhiz I can help with the conflicts on the form if am advised which changes to keep. The ones on this branch or on the master branch. I assume it would be the changes on this branch?
@dannyparsons I wonder if you are able to look at the data file conflict to determine what we keep so that t his is reviewed?

rdstern:
@Patowhiz I now have a good example to test yuor pasting when it is ready!
This also assumes I can paste into the label variable of the metadata.  If not yet, then that would be very useful - and fits pasting well!

Just to remind myself.  There is a new package called `childdevdata`.  It gives various case studies of research.  The documentation lists the names of the variables, but they are not included in the rds files - they are in the guide.
So I hope I can copy the information from the guide (pdf) and paste into a new 3 variable data frame.  Then take the description and paste it into the label variable of the metadata!

Patowhiz:
This PR has now been replaced by PR #6503.

PR #6067, opened 2020-11-05
Saving and position columns bug fixes and changes

dannyparsons:
1. Fixed functionality so that when replacing an existing column, it stays in same position by default and also added parameter to the R methods to allow it to be moved when replaced (a control for this can be added to the positioning sub dialog in a separate PR)
2. Fixed bug in calculator where the `before` parameter, and therefore positioning a new column, was changed when reopening the dialog
3. Added `strAdjacentColumn` as a parameter to assign to methods as this is now part of the saving options

@africanmathsinitiative/developers This is ready for the review. 1. needs to be reviewed across dialogs that use the save control, please test carefully as this edits core parts of saving columns, and 2. only needs to be checked for the calculator dialog

### DIFF
--- a/instat/clsRCodeStructure.vb
+++ b/instat/clsRCodeStructure.vb
@@ -250,8 +250,7 @@ Public Class RCodeStructure
     '''                                             names of new data frames if data frame list is
     '''                                             not named. </param>
     '''--------------------------------------------------------------------------------------------
-    Public Sub SetAssignTo(strTemp As String, Optional strTempDataframe As String = "", Optional strTempColumn As String = "", Optional strTempModel As String = "", Optional strTempGraph As String = "", Optional strTempSurv As String = "", Optional strTempTable As String = "", Optional bAssignToIsPrefix As Boolean = False, Optional bAssignToColumnWithoutNames As Boolean = False, Optional bInsertColumnBefore As Boolean = False, Optional bRequireCorrectLength As Boolean = True, Optional bDataFrameList As Boolean = False, Optional strDataFrameNames As String = "")
-
+    Public Sub SetAssignTo(strTemp As String, Optional strTempDataframe As String = "", Optional strTempColumn As String = "", Optional strTempModel As String = "", Optional strTempGraph As String = "", Optional strTempSurv As String = "", Optional strTempTable As String = "", Optional bAssignToIsPrefix As Boolean = False, Optional bAssignToColumnWithoutNames As Boolean = False, Optional bInsertColumnBefore As Boolean = False, Optional bRequireCorrectLength As Boolean = True, Optional bDataFrameList As Boolean = False, Optional strDataFrameNames As String = "", Optional strAdjacentColumn As String = "")
         strAssignTo = strTemp
         If Not strTempDataframe = "" Then
             strAssignToDataFrame = strTempDataframe
@@ -277,10 +276,10 @@ Public Class RCodeStructure
         Me.bAssignToIsPrefix = bAssignToIsPrefix
         Me.bAssignToColumnWithoutNames = bAssignToColumnWithoutNames
         Me.bInsertColumnBefore = bInsertColumnBefore
+        Me.strAdjacentColumn = strAdjacentColumn
         Me.bRequireCorrectLength = bRequireCorrectLength
         Me.bDataFrameList = bDataFrameList
         Me.strDataFrameNames = strDataFrameNames
-
     End Sub
 
     '''--------------------------------------------------------------------------------------------
--- a/instat/clsRCodeStructure.vb
+++ b/instat/clsRCodeStructure.vb
@@ -249,6 +249,7 @@ Public Class RCodeStructure
     ''' <param name="strDataFrameNames">            (Optional) Optional R character vector to give
     '''                                             names of new data frames if data frame list is
     '''                                             not named. </param>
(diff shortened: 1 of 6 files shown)


lloyddewit:
@Patowhiz You've done quite a bit of work in this area, please could you peer review? thanks

rdstern (approved):
Good

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -478,9 +478,9 @@ DataBook$set("public", "set_metadata_changed", function(data_name = "", new_val)
...
-  if(missing(use_col_name_as_prefix)) self$get_data_objects(data_name)$add_columns_to_data(col_name, col_data, hidden = hidden, before = before, adjacent_column = adjacent_column, num_cols = num_cols, require_correct_length = require_correct_length)
-  else self$get_data_objects(data_name)$add_columns_to_data(col_name, col_data, use_col_name_as_prefix = use_col_name_as_prefix, hidden = hidden, before = before, adjacent_column = adjacent_column, num_cols = num_cols, require_correct_length = require_correct_length)
+DataBook$set("public", "add_columns_to_data", function(data_name, col_name = "", col_data, use_col_name_as_prefix = FALSE, hidden = FALSE, before, adjacent_column, num_cols, require_correct_length = TRUE, keep_existing_position = TRUE) {
+  if(missing(use_col_name_as_prefix)) self$get_data_objects(data_name)$add_columns_to_data(col_name, col_data, hidden = hidden, before = before, adjacent_column = adjacent_column, num_cols = num_cols, require_correct_length = require_correct_length, keep_existing_position = keep_existing_position)
```
Patowhiz, line 482: @dannyparsons will parameter `use_col_name_as_prefix` ever be missing? I can see it has a default value `False`.

instat/clsRCodeStructure.vb
```diff
@@ -250,8 +250,7 @@ Public Class RCodeStructure
...
-    Public Sub SetAssignTo(strTemp As String, Optional strTempDataframe As String = "", Optional strTempColumn As String = "", Optional strTempModel As String = "", Optional strTempGraph As String = "", Optional strTempSurv As String = "", Optional strTempTable As String = "", Optional bAssignToIsPrefix As Boolean = False, Optional bAssignToColumnWithoutNames As Boolean = False, Optional bInsertColumnBefore As Boolean = False, Optional bRequireCorrectLength As Boolean = True, Optional bDataFrameList As Boolean = False, Optional strDataFrameNames As String = "")
-
+    Public Sub SetAssignTo(strTemp As String, Optional strTempDataframe As String = "", Optional strTempColumn As String = "", Optional strTempModel As String = "", Optional strTempGraph As String = "", Optional strTempSurv As String = "", Optional strTempTable As String = "", Optional bAssignToIsPrefix As Boolean = False, Optional bAssignToColumnWithoutNames As Boolean = False, Optional bInsertColumnBefore As Boolean = False, Optional bRequireCorrectLength As Boolean = True, Optional bDataFrameList As Boolean = False, Optional strDataFrameNames As String = "", Optional strAdjacentColumn As String = "")
```
Patowhiz, line 253: Would be good to add the parameter added `strAdjacentColumn` to the list of parameters in the comment.

Patowhiz: @dannyparsons probably not related to the added functionality.
In the same `SetAssignTo()` I have noticed that some optional parameters are always overwriting the initially set property values while others don't.

Shouldn't we have the same consistency? As a developer using the object, I would think that if I don't define the value of the optional parameter then the internal value that it's meant to assign to, won't be overwritten. 

Alternatively we could include a comment to the optional values that are always used in the subroutine regardless of whether they are used or not, just to be clear.

Patowhiz (commented):
This looks good. Tested it across multiple dialogs that use the ucrSave control. `ucrSave` control will need to be modified to enable positioning of existing columns and take advantage of the new parameter added. Issue #6002 will also need to be reopened to reimplement the functionality requested.

Just a few questions to @dannyparsons.

Patowhiz:
@dannyparsons @lloyddewit, on a separate note, now that @dannyparsons has worked on the `SetAssignTo()` subroutine, I think in line 218 `If bUseBaseFunction Then 'TODO SJL 07/04/20 Use ElseIf?` of `clsRSyntax.vb`.
The todo could now be done by adding the `ElseIf` as recommended

instat/static/InstatObject/R/instat_object_R6.R
dannyparsons: `missing(use_col_name_as_prefix)` is still `TRUE` if it is not specified by the user, even if it has a default. But I don't think that's relevant here and there isn't a need for the `if` `else` anymore so I've removed that.

instat/clsRCodeStructure.vb
dannyparsons: Added the comment.

Yes we probably don't want to overwrite unless the user has specified. We should probably have all the defaults as `Nothing` or `Null` and only change if not. I would want to first check if any uses of this rely on the existing implementation before changing. So I think this should be a separate issue/PR to do this.

dannyparsons:
Thanks, done and addressed the comments.

Patowhiz:
@lloyddewit this looks good to me.

PR #6076, opened 2020-11-09
Fixing convert to factor parameter order

### LINKED ISSUE #6005: Change as.factor to as_factor?

lilyclements:
As mentioned in #6005:

The parameters are in the wrong order when you use right-click to convert to a factor. The function specifies that `col_names` parameter should be before the `to_type` parameter:
If you right click to convert to factor:
`data_book$convert_column_to_type(data_name="cars", to_type="factor", col_names=c("speed"))`
If you go through dlgConvertColumns (accessed by the Prepare menu):
`data_book$convert_column_to_type(data_name="cars", col_names="speed", to_type="factor")`

This PR adds the parameter positions so that the order is consistent regardless of how converting the variable this is accessed.

@africanmathsinitiative/developers  this is ready to review


instat/ucrColumnMetadata.vb
```diff
@@ -289,9 +289,9 @@ Public Class ucrColumnMetadata
     'End Sub
 
     Private Sub mnuCovertToOrderedFactors_Click(sender As Object, e As EventArgs) Handles mnuCovertToOrderedFactors.Click
-        clsConvertOrderedFactor.AddParameter("col_names", GetSelectedVariableNames())
-        clsConvertOrderedFactor.AddParameter("to_type", Chr(34) & "ordered_factor" & Chr(34))
-        RunScriptFromColumnMetadata(clsConvertOrderedFactor.ToScript, strComment:="Right Click Menu: Convert to Ordered Factor")
+        clsConvertOrderedFactor.AddParameter("to_type", Chr(34) & "ordered_factor" & Chr(34), iPosition:=1)
+        clsConvertOrderedFactor.AddParameter("col_names", GetSelectedVariableNames(), iPosition:=2)
```
dannyparsons, line 293: I think you've switched them the wrong way round here?

instat/ucrDataView.vb
```diff
@@ -813,32 +813,32 @@ Public Class ucrDataView
...
+        clsConvertTo.AddParameter("col_names", SelectedColumns(), iPosition:=1)
+        clsConvertTo.AddParameter("to_type", Chr(34) & "logical" & Chr(34), iPosition:=2)
         RunScriptFromDataView(clsConvertTo.ToScript(), strComment:="Right click menu: Convert Column(s) To Logical")
     End Sub
 
     Private Sub mnuConvertToNumeric_Click(sender As Object, e As EventArgs) Handles mnuConvertToNumeric.Click
-        clsConvertTo.AddParameter("to_type", Chr(34) & "numeric" & Chr(34))
-        clsConvertTo.AddParameter("col_names", SelectedColumns())
+        clsConvertTo.AddParameter("to_type", Chr(34) & "numeric" & Chr(34), iPosition:=2)
+        clsConvertTo.AddParameter("col_names", SelectedColumns(), iPosition:=1)
```
dannyparsons, line 841: It looks like you've left these in the other order just to see if anyone noticed ;) It looks like the only one where they are added in a different order, doesn't make any difference but now I can't unsee the inconsistency :)

dannyparsons (changes requested):

instat/ucrDataView.vb
lilyclements: Yes, all a big test - congrats! I can't unsee it either, have sorted it now.

dannyparsons (approved):
@lloyddewit happy with this. There is no functionality change to R code being run from these changes.

lloyddewit:

rdstern (approved):
Happy to approve this change to the code.  And so I do.  But while changing the code I note there remains an issue of ordered factors.  I don't know if @lilyclements or @dannyparsons feels it is a good time to also change that.
If I have numeric I can make it either a factor, or an ordered factor.  But once it is a factor I am then not able to make it (instead) into an ordered factor.  I have tired clicking on ordered factor.  I have also tried making it numeric again and then an ordered factor.  It stubbornly remains labelled as (f), rather than (o.f).

lloyddewit (approved):

dannyparsons:
This should be mentioned on the issue about defining factors, but yes it is related.

PR #6077, opened 2020-11-09
Updating convert_column_to_type function

### LINKED ISSUE #6005: Change as.factor to as_factor?

lilyclements:
Fixes #6005 

When converting to a factor in the `convert_column_to_type` R function, this PR means that functions that work similarly to that of the `forcats` package are used.


instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1161,17 +1161,32 @@ DataSheet$set("public", "convert_column_to_type", function(col_names = c(), to_t
...
       if(set_decimals) curr_col <- round(curr_col, digits = set_digits)
       if(ignore_labels) {
+      if (is.numeric(curr_col)){
+        new_col <- forcats::fct_inseq(forcats::as_factor(curr_col), ordered = make_ordered)
+		} else if (is.factor(curr_col) | is.character(curr_col)){
+        new_col <- forcats::fct_inorder(curr_col, ordered = make_ordered)
+        } else if (is.logical(curr_col)){
+        new_col <- forcats::fct_inorder(forcats::as_factor(curr_col), ordered = make_ordered)
+		} else {
         new_col <- factor(curr_col, ordered = make_ordered)
```
lilyclements, line 1171: I'm still running `} else { new_col <- factor(curr_col, ordered = make_ordered)` for classes which `fct_inseq` or `fct_inorder` do not work for, such as dates.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1161,17 +1161,32 @@ DataSheet$set("public", "convert_column_to_type", function(col_names = c(), to_t
       make_ordered <- (to_type == "ordered_factor")
       if(set_decimals) curr_col <- round(curr_col, digits = set_digits)
       if(ignore_labels) {
+      if (is.numeric(curr_col)){
+        new_col <- forcats::fct_inseq(forcats::as_factor(curr_col), ordered = make_ordered)
```
lilyclements, line 1165: Is there a better way to do this? A numeric variable has to be a factor for `fct_inseq` to work, so I convert it to a factor anyway.
The addition that `fct_inseq` gives is this `ordered` parameter.
Should I instead do something to say to run `as_factor` if `make_ordered` is `FALSE`, and to otherwise run this line?

lilyclements: Also, I'm using `fct_inseq`, not `fct_inorder` here to keep the levels in sequential order

```
library(forcats)
y <- c(2:12, 1)
head(fct_inorder(as_factor(y)))
# [1] 2 3 4 5 6 7
# Levels: 2 3 4 5 6 7 8 9 10 11 12 1
head(fct_inseq(as_factor(y)))
# [1] 2 3 4 5 6 7
# Levels: 1 2 3 4 5 6 7 8 9 10 11 12
```

lilyclements (commented):
@dannyparsons I have updated the `convert_column_to_type` function. I have a few comments below just to check if there is a better way to do this.

Another two small questions:

1. There are two dialogs where `as.factor` is run as an additional function (not the main function). These dialogs are dlgClimaticBoxplot and dlgTwoVariableFitModel. Should I convert these to `forcats::as_factor` as well?
2. Additionally, `as.factor` is used in some R codes (DateManipulationMethods.R, climate_data_refclass.R, climate_refclass.R, display_rain_running_total_method.R, display_water_balance_method.R). Should these also be converted to `forcats::as_factor`?

instat/static/.old/data_object.R
```diff
@@ -586,7 +586,7 @@ data_obj$methods(convert_column_to_type = function(col_names = c(), to_type = "f
   
   for(col_name in col_names){
     if(to_type=="factor"){
-      data[,col_name] <<- as.factor(data[,col_name])
+      data[,col_name] <<- forcats::as_factor(data[,col_name])
```
lilyclements, line 589: Just had a thought about this one - `as_factor` calls an error for some classes of `f`, such as if it is a date. I think it only works for numerics, logicals, characters, and factors. Presumably this should remain as `as.factor` then?

dannyparsons: I think we need to do the same thing here and check the type of the column and then use `factor()` with the right levels/

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -1539,4 +1539,14 @@ slope <- function(y, x) {
 
 }
 
-
+make_factor <- function(x, make_ordered = FALSE) {
```
dannyparsons, line 1542: Could this be `ordered`? I think useful if it's the same name as in `factor`.

lilyclements: Good point. Sorted it

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -1539,14 +1539,14 @@ slope <- function(y, x) {
 
 }
 
-make_factor <- function(x, make_ordered = FALSE) {
+make_factor <- function(x, ordered = FALSE) {
```
dannyparsons, line 1542: let's have the same default, `is.ordered(x)` for `ordered`

dannyparsons:
This would have closed automatically, by the way, when #6088 is merged, because the commits are contained there.

lilyclements:
I thought this was the case. Github is a lot fancier than I give it credit for.

PR #6064, opened 2020-11-05
Added the first Circular Plot (Density Plot dialog)

Wycklife:
Fixes #6060
I have added data selector
Two single receivers
Two buttons which opens  a subdialog each
Code Implementation is in the process.

@africanmathsinitiative/developers Still in progress.

PR #6079, opened 2020-11-10
Added the second circular dialog (Scatter Plot)

### LINKED ISSUE #6060: Adding three plots dialogs from the circular package in Structur>Circular menu

Wycklife:
Fixes #6060 

  @africanmathsinitiative/developers  This is still being implemented. It's not yet ready for review.


Wycklife:
@lloyddewit I'm closing this PR since it doesn't implement what is required and therefore after discussion with Davis, I'm in the process of re-implementing it.

lloyddewit:
@Wycklife Same comment as PR #6064:  I'm not aware of any other open PRs for this issue so I assume that you'll open a new one. When you do this, please reference this PR, thanks

PR #6087, opened 2020-11-12
Making Improvements to the ucrSave control in various dialogs

### LINKED ISSUE #6080: Improve the position control in various dialogues
a) This can start with the calculator.  The Save Result control (which now has the position button added) I thought used to be a pull-down, so it was easy to choose the existing variables.  If so, it has gone now.  Please could it be either put back of put there if it wasn't ever there before.
 To see what is needed see, for example Prepare > Check > Duplicates where there is the pull-down.  That is what we should have consistently.
I am sure @Patowhiz can help if you have problems.

b) Also do the same with Prepare > Check Data > Duplicates.  On that dialogue in addition:
1) Make the save control smaller, so it doesn't overlap with the text.
2) Add the pull-down
3) When the Save option is unticked the Position button should not be visible
(Notice all the other dialogues in Prepare > Column: Calculate menu are ok.)
1) Save Control
2) Change the size of the position, which is larger here.

I hope this will be simple, but feel free to break this into a number of issues if you wish

Wycklife:
Fixes #6080
I have added SetIsComboBox property for all the listed dialogs in the issue above
I have also reduce the size of the save control as well as the size of the position button as mentioned in the issue.

@africanmathsinitiative/developers  This is issue is not yet complete. It's in progress.


Wycklife:
@rdstern This is now ready for review.

Wycklife:
@Patowhiz Could you please review this?

rdstern:
@Wycklife thanks.  I will do it now.

rdstern:
In the list in #6080 some are now ok
b) I gave the wrong name for the dialogue.  Please do this item, which is the Compare Columns dialogue.  There are 3 things to correct.

c) The position part of the control doesn't work here.
k) You didn't do this
l) you didn't do this

rdstern:
@Wycklife you have been handing over some tasks to @HawardKetoyoMsatsi, but this pull request goes to @Patowhiz instead, because it is his area.
In addition, I missed (at least) one dialogue.  This is the Climatic > Prepare > Transform dialogue.  Please could this also have the position control added, plus the pull-down.  And the default is to save the variable just after the varieble being transformed.
Meanwhile the Climatic > Prepare > Conversions dialogue has the Position button, but need to add the pull-down.

rdstern (changes requested):
@Patowhiz I hope this is one you have (or will now) take over from @Wycklife .

I checked the most urgent, which is the re-addition of the checkbox for the save option in the calculator.  You used to have the option to save.
@Wycklife has now re-instated that feature, but it has problems:
a) When you untick it still says Save Result Into - and the Position button remains visible.  Ideally just Save Result and without the new Position button.
b) In the output window you then do get the result, but also some extra control information.  That extra information should be deleted.

maxwellfundi:
@Patowhiz  this is now in your hands from @Wycklife ? while it has conflicts, it will be good to take it over and finish for this sprint.

Patowhiz:
@lloyddewit I think this can now be closed. It was done in another PR.

lloyddewit:
@Patowhiz Thanks, closed as suggested

PR #6088, opened 2020-11-12
Updating convert columns with new factor defaults

### LINKED ISSUE #3372: Convert to Ordered Factor not working for labelled data
Using efc data tried to make several different columns into Ordered Factor, but it just made it a Factor.

### LINKED ISSUE #3642: Factor columns are too persistent!
There still seem to be some issues concerned with factor columns.

Whether they can be dealt with by 4.0 is not clear.  
 Once a column is a factor I cannot make it an ordered factor it seems.  And I wonder about dates?
 here are my issues.
1) I had a date column.  From that column I got a year column and also a month column.  It made the month into an ordered factor which is fine.  I made the year into an ordinary factor column, which was fine.  Then I decided to change that into an ordered factor and it seems to refuse.  It persistently treats it as an ordinary factor - or at least it labels it as (f) rather than (o.f).
2. I made a date column into a factor column.  (I know that is strange, but I wanted to see if it caused any issues.  It seemed fine.)  Then I made a combined factor from the date and station.  Then I looked for the number of observations in each factor level.  That took over an hour on my machine, so I don't recommend this route - at least not with 200,000 rows of data and almost as many factor levels.  But it worked! Much to my surprise!
3. But then I wanted my date column to stop being a factor and become a date again.  Perhaps it is, but it still has the (f) rather than (D) after the column name.

### LINKED ISSUE #4041: Converting from ordered factor to Factor and Vice versa ?
Seems this is not working. I accidentally changed my column to ordered factor instead of factor(unordered).So i got my column as ordered factor and now  wanted to change this to factor but seems this isn't working . I checked the code and this is running well `InstatDataObject$convert_column_to_type(data_name="WKenyaRain", to_type="factor", col_names=c("Year","Month"))` but the column datatype hasn't changed it is still `o.f`. Is this something to worry about?

### LINKED ISSUE #4555: Having factor levels in data order
Here is an example from Mike.  There are 3 columns of data with the first being a character variable with 42 levels.  For the analysis it needs to be made into a factor.  But ideally the levels would remain the same as in the data and not in alphabetical order.  I think that is possible using the FORCATS package?  It would also usually solve our usual problem with making a Month factor in climatic data. 

[Germination rates by rep.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/1938863/Germination.rates.by.rep.xlsx)

### LINKED ISSUE #6005: Change as.factor to as_factor?

dannyparsons:
This PR builds on and replaces #6077 

Changes:
- converting to factor works similarly to `forcats` package fixes #6005 fixes #4555
- this has also been changed in some of our own R functions
- converting between ordered/unordered factor now works fixes #4041 fixes #3642 fixes #3372 


rdstern (approved):
Great change.  As @dannyparsons has said we have only been waiting 3 years for this!

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -1539,4 +1539,29 @@ slope <- function(y, x) {
...
+#    For any other data types, levels are given in order of appearance (the same as for "character").
+#    Note that this should be used cautiously for other data types and the default behaviour may not be the most sensible.
+# If anything other than this default behaviour is required, use factor().
+make_factor <- function(x, ordered = is.ordered(x)) {
+  if (is.factor(x)) {
+    if (ordered != is.ordered(x)) {
+      if (ordered) class(x) <- c("ordered", class(x))
+      else class(x) <- class(x)[class(x) != "ordered"]
+    }
+    x
```
lloyddewit, line 1557: @dannyparsons sorry for my limited knowledge of R, but what does this line do?

lloyddewit (dismissed):
@dannyparsons I reviewed with my limited knowledge of R. It looked good, I just had one small question on ine 1557. 
Should we also ask one of the developers to review? Or do you think it's safe to merge?

instat/static/InstatObject/R/stand_alone_functions.R
dannyparsons: This returns `x`, equivalent to `return(x)`. We often use `return()` explicitly but for simple functions I would do it this way too where it's fairly obvious. That's kind of consistent with section 3.3 https://style.tidyverse.org/functions.html. Though maybe this is technically an early return, but since there's just one main If Else block I would consider them all as standard returns.

dannyparsons:
If @lilyclements is able to review as well that would be good as she could also check the R code.

lilyclements (commented):
The smallest of all comments (I was desperate to find *something* clearly). Otherwise looks good to me. 
I'll write an issue later about the various R-codes/dialogs that should be updated with this function.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -1539,4 +1539,29 @@ slope <- function(y, x) {
 
 }
 
-
+# make_factor is intended to be somewhat equivalent to forcats::as_factor() or base::as.factor().
+# It provides default behaviour for converting to factor depending on the data type, similar to forcats::as_factor().
+# For "character" and "numeric" types make_factor is consistent with forcats::as_factor() in terms of the order of the factor levels.
+# It differs from forcats::as_factor() in two main ways:
+# 1. It includes an ordered parameter to allow for creating ordered factors, including converting a factor to an ordered factor (and vice versa).
+# 2. It works for any data types (e.g. Dates) whereas forcats::as_factor() is limitied to "factor", "character", "logical", "numeric".
```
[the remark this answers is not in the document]
dannyparsons: Thanks. Have you also seen "suggestions" in GitHub? Like this:
```suggestion
# 2. It works for any data types (e.g. Dates) whereas forcats::as_factor() is limited to "factor", "character", "logical", "numeric".
```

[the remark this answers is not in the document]
dannyparsons: I now have the option to commit the suggestion directly without having to go back to code, change, commit and sync. It's perfect for small corrections like these!

image

lloyddewit (approved):

lloyddewit:
@rdstern Only a comment changed since your last approval. If you approve again then I can merge, thanks

@dannyparsons Or you can merge directly if you prefer

instat/static/InstatObject/R/stand_alone_functions.R
lloyddewit: Understood, thanks

rdstern (approved):

PR #6091, opened 2020-11-16
Adding the station radio button to Climatic summaries dialogue

### LINKED ISSUE #6068: Minor addition in the Climatic > Prepare > Climatic Summaries dialogue
This is to add a 4th button - really the first button - at the start of the top of the dialogue.  It is called Station.  The other 3 buttons will have to become a bit smaller to keep the same size for the dialogue.
The default is still the Year button.
If the Station option is chosen then I am happy that the same other items remain visible, as are on the Year button.  But the Station must be completed for Ok to be enabled.  I don't think that is necessary in general.  

It then works just the same, but for overall summaries - so there will be one row per station.  In a file with just one station, then there is just a single row in the summary data.

I hope this will be easy.  It is needed for one of the videos.  It can be checked by comparing it to the main summary dialogue, where just the Station is used as the By variable.

I assume this will be quite quick for @Ivanluv to do?  Perhaps @shadrackkibet could confirm that?

Ivanluv:
fixes #6068


instat/dlgClimaticSummary.vb
```diff
@@ -62,12 +62,15 @@ Public Class dlgClimaticSummary
         ucrPnlAnnualWithin.AddRadioButton(rdoAnnual)
         ucrPnlAnnualWithin.AddRadioButton(rdoAnnualWithinYear)
         ucrPnlAnnualWithin.AddRadioButton(rdoWithinYear)
+        ucrPnlAnnualWithin.AddRadioButton(rdoStation)
 
+        ucrPnlAnnualWithin.AddParameterPresentCondition(rdoAnnual, "stations", False)
```
dannyparsons, line 67: That's not a correct condition for Annual. Annual will be by stations if there are stations.

instat/dlgClimaticSummary.vb
```diff
@@ -62,12 +62,15 @@ Public Class dlgClimaticSummary
...
         ucrPnlAnnualWithin.AddParameterPresentCondition(rdoAnnual, "within_variable", False)
         ucrPnlAnnualWithin.AddParameterPresentCondition(rdoAnnualWithinYear, "within_variable", True)
         ucrPnlAnnualWithin.AddParameterPresentCondition(rdoAnnualWithinYear, "year", True)
         ucrPnlAnnualWithin.AddParameterPresentCondition(rdoWithinYear, "within_variable", True)
         ucrPnlAnnualWithin.AddParameterPresentCondition(rdoWithinYear, "year", False)
+        ucrPnlAnnualWithin.AddParameterPresentCondition(rdoStation, "stations", True)
```
dannyparsons, line 73: The conditions for station should also have not year and not within year.

instat/dlgClimaticSummary.vb
```diff
@@ -354,21 +363,30 @@ Public Class dlgClimaticSummary
...
+                    clsDefaultFactors.RemoveParameterByName("within_variable")
+                    clsDefaultFactors.AddParameter(ucrReceiverYear.GetParameter())
+                ElseIf rdoAnnualWithinYear.Checked Then
+                    clsDefaultFactors.AddParameter(ucrReceiverWithinYear.GetParameter())
+                    clsDefaultFactors.AddParameter(ucrReceiverYear.GetParameter())
+                ElseIf rdoWithinYear.Checked Then
+                    clsDefaultFactors.RemoveParameterByName("year")
+                    clsDefaultFactors.AddParameter(ucrReceiverWithinYear.GetParameter())
+                End If
+                clsDefaultFactors.RemoveParameterByName("stations")
```
dannyparsons, line 389: Why are you removing this here? Stations should always be included for each option if there is a station column, as you have done on line 373.

dannyparsons (changes requested):

instat/dlgClimaticSummary.vb
Ivanluv: `stations` is a different parameter from `station`. I made it to simplify assigning of conditions

dannyparsons: Ok I think I've seen what you've done with the two parameters. But this is quite confusing when reading the code (and for someone editing this in future to easily make a mistake), and I don't think necessary. If the only reason is to shorten the number of conditions then I don't think it is an advantage over having clearer and more logical code, even if it's longer. It would be much clearer to have a single parameter for stations, and have four conditions for the station radio button. That also simplifies the code below when setting the parameters.

rdstern (changes requested):
@Ivanluv this is great.  It works well.  The one problem needs perhaps a comment from @dannyparsons .  If you don't have a Station factor, then Ok is still enabled.  But the summary doesn't work without a factor.  You can have a factor with a single level - that's fine.  But otherwise you just get a single number in the summary, rather than a single row.
I suggest, for this option, you don't have Ok enabled unless the Station factor is present.  Note that you don't need a station factor for any of the other 3 radio-button options, because then you have a year etc factor.  The problem is just when there is no factor.

dannyparsons:
I agree, if there are no station variables then ok is disabled.

Ivanluv:
@rdstern  could you look at this again. @dannyparsons  I have made some changes to the conditions but am not sure if its the correct way to do it

instat/dlgClimaticSummary.vb
```diff
@@ -62,12 +62,17 @@ Public Class dlgClimaticSummary
...
         ucrPnlAnnualWithin.AddRadioButton(rdoWithinYear)
+        ucrPnlAnnualWithin.AddRadioButton(rdoStation)
 
+        ucrPnlAnnualWithin.AddParameterPresentCondition(rdoAnnual, "year", True)
         ucrPnlAnnualWithin.AddParameterPresentCondition(rdoAnnual, "within_variable", False)
         ucrPnlAnnualWithin.AddParameterPresentCondition(rdoAnnualWithinYear, "within_variable", True)
         ucrPnlAnnualWithin.AddParameterPresentCondition(rdoAnnualWithinYear, "year", True)
         ucrPnlAnnualWithin.AddParameterPresentCondition(rdoWithinYear, "within_variable", True)
         ucrPnlAnnualWithin.AddParameterPresentCondition(rdoWithinYear, "year", False)
+        ucrPnlAnnualWithin.AddParameterPresentCondition(rdoStation, "station", True)
```
dannyparsons, line 73: This is technically correct, but I think you will have an issue when no factors are given, because then none of the conditions will be matched and you'll get a developer error. So if you remove this condition then when no factors are filled in, the station radio button will be selected, and there's no chance of errors.

instat/dlgClimaticSummary.vb
```diff
@@ -201,6 +206,8 @@ Public Class dlgClimaticSummary
         clsDefaultFunction.AddParameter("silent", "TRUE")
 
         clsDefaultFactors.SetRCommand("c")
+        'This a temporary fix to enable correct functioning of conditions
+        clsDefaultFactors.AddParameter("year", "")
```
dannyparsons, line 210: I think by removing the line mentioned above this won't be needed anymore and can be removed.
@rdstern are you happy that the station option is the default when opening the dialog?

rdstern:
Yes indeed

lloyddewit:
@Ivanluv did you resolve all the comments from @dannyparsons and is it ready for another review? thanks

Ivanluv:
@lloyddewit this is ready for review

lloyddewit:
@dannyparsons please could you confirm that your comments are all resolved? thanks

rdstern (approved):
It is time this was merged.

lloyddewit:
@Ivanluv Please could you clarify if/how @dannyparsons 's comments were resolved? thanks

Ivanluv:
@lloyddewit  have resolved the comment from @dannyparsons  by having the station radio button as the default and removing the unnecessary code

instat/dlgClimaticSummary.vb
lloyddewit: confirmed resolved

instat/dlgClimaticSummary.vb
lloyddewit: confirmed resolved

instat/dlgClimaticSummary.vb
lloyddewit: confirm resolved

lloyddewit (approved):

PR #6093, opened 2020-11-18
Minor Formmain Changes

Wycklife:
@lloyddewit  These are minor changes on the frmmain. Just changed the enabled property from False to True for the three dialogs. Also Set the ShowDialog() property for the same dialogs; Structured>Circular>Scatter Plot, Density Plot, Rose Plot.
I still have some pending already done work which require this PR to be merged for easy testing.

### DIFF
--- a/instat/frmMain.resx
+++ b/instat/frmMain.resx
@@ -3498,26 +3498,17 @@
   <data name="ToolStripSeparator46.Size" type="System.Drawing.Size, System.Drawing">
     <value>189, 6</value>
   </data>
-  <data name="mnuStructuredCircularScatterPlot.Enabled" type="System.Boolean, mscorlib">
-    <value>False</value>
-  </data>
   <data name="mnuStructuredCircularScatterPlot.Size" type="System.Drawing.Size, System.Drawing">
     <value>192, 22</value>
   </data>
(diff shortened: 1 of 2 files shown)


lloyddewit:
@Wycklife I understand that this PR will enable the 3 new circular plot dialogs. If this PR is merged then the main branch R-Instat version will allow the user to open any of the 3 dialogs, and will then display an empty dialog.
I understand that merging the `frmMain` changes early reduces the risk of merge conflicts later.
However, we also want to avoid the main branch accessing empty dialogs. We should only enable options when the functionality that is enabled has been implemented. 
What is the current state of the 3 new dialogs and when do you expect implementation to be complete?

@dannyparsons I understand why @Wycklife  would like these `frmMain` changes merged early. If you have extra advice about the optimum time to merge `frmMain` changes, then please let us know, thanks

dannyparsons:
@lloyddewit I agree. Dialogs should only be enabled once they are (fully or partially) functional. The master branch should always be in a "release ready" state. I know this has not been what's been happening recently but we should follow this rule. There are currently empty dialogs that can be accessed in the current version and I don't think they should be viewable either.

maxwellfundi:
I am closing this as it was done in another PR #6260

Issue #6095, opened 2020-11-20
Updating to make_factor


lilyclements:
With the addition of the `make_factor` function in the software, we need to update some functions in dialogs. The functions `factor`, `as.factor`, or `as_factor` should be updated where the levels are automatically set (i.e., where there is no `levels` parameter currently read in).

Places which need to be updated are:

* data_object_R6.R - in the `split_date` function (but not everywhere, such as for pentad_abbr, dekad_abbr_vector)
* data_object_R6.R - in the `make_inventory_plot` function, the line `factor(curr_data[[station_col]])` should be updated to `make_factor(curr_data[[station_col]])`
* dlgClimaticBoxplot - replace `as.factor` with `make_factor`

PR #6096, opened 2020-11-20
Updating split_date function with make_factor

lilyclements:

This PR updates the `split_date` function to use `make_factor` function instead of the `factor` function where the levels are automatically set (i.e., where there is no levels parameter currently read in), this is as outlined in #6095


lloyddewit:
@rdstern I think a review from you and I is adequate for this PR, please could you test? thanks

rdstern (approved):
Great.  This seems fine

lloyddewit (approved):

PR #6097, opened 2020-11-20
Added the first circular plot dialog (Rose diagram) in Structure>Circular 

### LINKED ISSUE #6060: Adding three plots dialogs from the circular package in Structur>Circular menu

Wycklife:
Fixes #6060
I have added and implemented the functionality of the dialog.
Added a few options as I'm still exploring more. This is also under discussions with @rdstern  and @volloholic 
I have also added a button for the circular options. Disabled for now.

@africanmathsinitiative/developers This is at the initial stage. Testing this would require you to first get the changes from PR #6093, where the dialog is enabled from the frmmain.


Wycklife:
I'm closing this PR since Haward is working on it

PR #6098, opened 2020-11-20
Added the second circular plot (Density plot) in Structured > Circular

Wycklife:
Fixe #6060

I have added and implemented the functionality of the dialog.
Added a few options as I'm still exploring more. This is also under discussions with @rdstern and @volloholic
I have also added a button for the circular options. Disabled for now.

These dialogs are not yet perfect including the designer as I'm still exploring the different options from the circular package.

@africanmathsinitiative/developers This is at the initial stage. Testing this would require you to first get the changes from PR #6093, where the dialog is enabled from the frmmain.

### DIFF
--- a/instat/dlgCircularDensityPlot.Designer.vb
+++ b/instat/dlgCircularDensityPlot.Designer.vb
@@ -70,7 +70,7 @@ Partial Class dlgCircularDensityPlot
         '
         'ucrSaveDensity
         '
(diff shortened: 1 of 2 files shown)


HawardKetoyoMsatsi:
Hello @Wycklife Could you please close this PR so that I can open an updated version of the dialogs.

PR #6107, opened 2020-11-26
Extremes calculation dialog improvements

### LINKED ISSUE #4760: Adding declustering to the calculation of extremes
The extRemes package includes a function called decluster.  I think this would be very useful for us for the **Climatic > Prepare > Extremes** dialogue when it is calculating the peaks over a threshold.

It needs someone better than I am at R to check exactly how it works.  The idea is that peaks over a threshold (say 40mm) should not be (say) neighboring days.  They need to be at least r days apart, where the user chooses r.

If r=1 then this function isn't needed.  If more than 1, then it could be useful prior to the existing filter.

The decluster function seems to produce a column of the same length as the original data, but (in any cluster where there are multiple values more than the threshold, it keeps the maximum and sets all the others to the threshold value.  So then if you take all the values more than the threshold (filter) after this, then it will just give the one value in each cluster.

It also has groups argument, so that neighboring values above a threshold would be OK if they belong to different groups.  We have 2 contenders for different groups, namely maybe different stations and also different years when we are getting the peaks for just a subset of the year.

The function also allows options for missing values.  I am not sure what happens now when there are missing values in the data?  I must check.

### LINKED ISSUE #5881: More options on extremes for peaks
This should involve Caleb once he can be assigned issues.

He has a Word document which is attached.
[extremes-dialogue (1).docx](https://github.com/africanmathsinitiative/R-Instat/files/4880460/extremes-dialogue.1.docx)

It includes a proposed addition to the **_Climatic > Prepare > Extremes_** dialogue at the end of the document.  That is a good start, but it needs some editing.  I propose the following - still a set of controls under the data selector.
There is currently an group box with the label Options.  

I suggest 2 radio buttons. Each has a box - no label for multiple controls.  
1) The first radio button has the label **_Threshold_** and the second has the label **_Investigate Thresholds_**.
2) Towards the bottom of the dialogue there is a control with the label **_New Data Frame Name_**.  This now has a checkbox and the label becomes **_Save New Data Frame_**.  Default is unchecked and it is disabled (though still visible) if the "Investigate Thresholds" button is chosen.
3) There is currently an group box with the label Options.  There will now be 2 group boxes, below each of the 2 radio buttons.  I don't think they need a label.
a) In the first group box (corresponds to threshold) there is the existing control to choose the **_Value_**.
b) Under that is a checkbox (default unchecked) with the label Decluster:  Gap: then an updown control with the minimum 1 and no maximum.  No - let's have a maximum of 99.  Default is 1.  After that is a label **_rows_**.
c) In the second group box I am less sure but possibly, first a set of 3 text fields into which you can type a number.  The first field has the label **_From:_**  the Second has the label **_To:_**  and the third has **_Steps:_** .  There could be defaults that are sensible for rainfall e.g. 30, 60, 5.  (The values could be negative except the third one and that must be positive and more than zero.  If the second value is less than the first, then it steps down rather than up.  - I am sure testers will try that.)
Then there are 3 options, each with a checkbox.  The first is Counts, the second is **_Mean Residual Life Plot_** and the third is **_Parameter Stability Plots_**.

It might be good to show a picture of the new controls, but perhaps they can be constructed from the instructions above and the dialogue then shown?

The R code is not trivial and still needs some investigation.  Here are some points:

a) Currently the analysis simply does a filter.  That is still what is done if the first box is used and there is no declustering and the new Save checkbox is ticked.  
b) In the new dialogue there is some output, which gives the Count at the chosen threshold.  This uses the calculation system and is the **_Counts_** for each station if there are multiple stations in the file.
c) If the Investigate Thresholds option is chosen, then the Counts checkbox would provide the Counts for each station at each threshold.
d) The code for this (in the dialogue) uses the R-Instat calculation "system".  I suggest @shadrackkibet could help here.  It would probably include a loop (I guess) for the multiple thresholds.  (@dannyparsons may be needed to help here?)
e) If the Decluster option is chosen, then the results from Caleb indicate there would be some output automatically.  I did not see how the filtered data would be provided (using the maximum (or minimum) from each cluster, but I presume that is an option from the function?
f) There is then a choice of package for the plotting options.  I look forward to the recommendations for the mean residual life plot and the parameter stability plots.  Caleb gave the impression that the parameter plots are not available from the extRemes package.  They are available and are called threshrange plots.
g) If both checkboxes are ticked for the plots, then it would result in 2 plots being produced.  I hope this isn't a problem?

shadrackkibet:
Fixes #4760 and Fixes #5881.  @rdstern I am putting up this for some quick testing just to check that this is what is needed although it is still work in progress. Mean Residual Life Plot is already working however the following still remain. 

- Add more options eg min and max thresholds, rug, line colour, fill etc... Some of these might go into a subdialog.
- Implement the following;
  - Tc plot
  - Declustering

FYI here current look of the dialog.
image


rdstern:
Yes - seems to work:
image

I note it also automatically adds the filter.  This plot would usually be prior to that.  The plots look good.  I assume that options for scaling the facets, etc will come.

shadrackkibet:
I actually thought of modifying the filter so that it gives the values within an interval say, PRECIP>= 10mm &PRECIP<=20mm. This will be concurrent with what is displayed on the graph.

shadrackkibet:
@rdstern this is ready for you to review. This PR implements Mean residual life plot. I would like to get this merged first before I proceed with the other components i.e Tc plot and declustering. 

- On your point above, the filter is still added when Mrl plot checkbox is checked. If that is not sensible then we can find a mechanism to avoid adding a filter. This can be by making Mrl plot explicit on top or hide the **value**  controls when Mrl plot checkbox is checked.

rdstern:
image

This is looking great.  I am happy to approve in its current form and do so.  Perhaps @dannyparsons can check and see if it is ready, or can soon be merged.

Being critical, the above default has the word excesses that overtypes the numbers below, particularly when the graph is smaller.  And the facets have the same scale above, so can we have them as proper facets, with the x part only at the bottom.  That will become more critical with more stations.


image

Here is the image "the other way".  Now I can't read the excesses, because of their size. Can this be changed?  And I wonder about having an option this way round, to have the same y axis for each graph - and hence save a lot of space again as well as compare the stations more easily.

Are there becoming too many options and we need a sub-dialogue button?

rdstern (approved):

instat/dlgExtremesClimatic.vb
```diff
@@ -188,8 +248,13 @@ Public Class dlgExtremesClimatic
         clsFilterExtremeSubCalcs.Clear()
         clsFilterExtremeExp.Clear()
 
+        clsPlotMrlFunction = New RFunction
```
lloyddewit, line 251: move this down to line 378?

instat/dlgExtremesClimatic.vb
```diff
@@ -442,6 +518,40 @@ Public Class dlgExtremesClimatic
...
+                minCharVector = exprMin.AsCharacter
+                ucrInputMin.SetText(minCharVector(0))
+            End If
+            If exprMax IsNot Nothing AndAlso Not exprMax.Type = Internals.SymbolicExpressionType.Null Then
+                maxCharVector = exprMax.AsCharacter
+                ucrInputMax.SetText(maxCharVector(0))
+            End If
+        End If
+
+        clsPlotMrlFunction.AddParameter("element_name", ucrReceiverElement.GetVariableNames, iPosition:=2)
```
lloyddewit, line 554: Once added, this parameter is never removed. Is that correct?

lloyddewit (changes requested):
looks good, just a couple of small comments

instat/dlgExtremesClimatic.vb
shadrackkibet: Keeping this here is consistent with what is done in other places. Clearing i.e `RFunction.Clear()` and setting of `New` instances of RFunction are normally done at the start of `setDefaults()` subroutine before adding parameters and setting other properties.

lloyddewit: understood, thanks

instat/dlgExtremesClimatic.vb
shadrackkibet: Done!

lloyddewit (approved):

lloyddewit:
@rdstern the only change since your previous approval is that the `element_name` parameter should now be removed when the receiver is empty. Please could you test/approve? thanks

rdstern (approved):
Happy to approve and this is an exciting addition.  An early next step is to separate the new graphs from the process of making a new data frame.
And I think there is another section to come?

lloyddewit (approved):

PR #6109, opened 2020-12-01
Adding sampling fraction option to dlgVisualiseData

### LINKED ISSUE #6008: At least a minor change on the visdat dialogue

lilyclements:
Replaces #6051 
Fixes #6008 - adds a sampling fraction option in the visualise data dialog.


lloyddewit:
@shadrackkibet Please could you review? thanks

lloyddewit (approved):

lloyddewit:
@shadrackkibet After @rdstern 's approval, I looked at the changes and decided that this PR could go straight to final review, without the need for a pre-review from you. I hope that's OK from your side

Issue #6110, opened 2020-12-02
Export to World Weather Record dialog


dannyparsons:
For WMO we have been asked to implement in R-Instat the ability to export data in the format required for submission to the World Weather Record, which national Met services are required to do for WMO. You don't need to know all the details about this but if you want to see here for more background https://www.wmo.int/pages/prog/wcp/wcdmp/GCDS_2.php and there is a WMO guide with specific details of the format https://www.wmo.int/pages/prog/wcp/wcdmp/documents/worldweatherrecords/1186_en.pdf

So I suggest we have a dialog in R-Instat to do this export. The data are monthly, there are up to 7 elements, plus station metadata.  Because of the number of variables to specify, I suggest a main dialog to specify the monthly data variables, and then a subdialog to specify the station metadata (from a separate data frame). 

Below is a rough design of the dialog and subdialog. The dialog has a "Station identifier" which can be either "Station Name" or "WMO Number" (selected from the combo box), next to or under this a button for "Station Metadata" to open the subdialog. Then there are receivers for year, month and the 7 elements. The subdialog just has the receivers shown. A different data frame will be selected for the station metadata.

The labels for the seven elements can be:
- mean station pressure
- mean sea-level pressure
- mean daily air temperature
- total precipitation
- mean daily max air temperature
- mean daily min air temperature
- mean daily relative humidity

Here is an R file with the R function created for the dialog [wwr_export.zip](https://github.com/africanmathsinitiative/R-Instat/files/5629352/wwr_export.zip). This should go in standalone functions.

Here is an example output file from WMO https://www.wmo.int/pages/prog/wcp/wcdmp/documents/WWR_txt_example_revised_2016_noMeanClino.txt
Note that the format specifies the exact spacing and alignment of values so this should be the same.

On the main dialog, the first receivers are required, then at least one element must be given. For the station metadata, everything apart from the last two are required.

I hope it's clear from the R file how this all links up. @HawardKetoyoMsatsi are you happy to take on this task? Check with @rdstern but I think this task would become highest priority now.

![WhatsApp Image 2020-12-02 at 12 18 48](https://user-images.githubusercontent.com/11214914/100871718-9edeb180-3498-11eb-8a6c-239b56b42f7b.jpeg)

dannyparsons:
@HawardKetoyoMsatsi Can you respond to this? This is an important task so if you can't take it up then we need to assign it to someone else.

HawardKetoyoMsatsi:
[...]

PR #6114, opened 2020-12-07
Adding Export to World Weather Record dialog 

HawardKetoyoMsatsi:
Added two blank forms dlgExportToWWR and sdgExportToWWR. This solves issue #6110
@africanmathsinitiative/developers and @rdstern this is ready for review. 
It contains frmmain changes therefore it should be merged as soon as possible.

### DIFF
--- a/instat/frmMain.resx
+++ b/instat/frmMain.resx
@@ -1332,7 +1332,7 @@
     <value>189, 22</value>
   </data>
   <data name="mnuClimaticPrepareSPI.Text" xml:space="preserve">
-    <value>SPI...</value>
+    <value>SPI/SPEI...</value>
   </data>
   <data name="ToolStripSeparator51.Size" type="System.Drawing.Size, System.Drawing">
     <value>186, 6</value>
--- a/instat/frmMain.resx
+++ b/instat/frmMain.resx
@@ -1058,6 +1058,12 @@
   <data name="mnuClimaticFileExportToCPT.Text" xml:space="preserve">
     <value>Export to CPT...</value>
   </data>
+  <data name="ExportToWWRToolStripMenuItem.Size" type="System.Drawing.Size, System.Drawing">
+    <value>247, 22</value>
+  </data>
+  <data name="ExportToWWRToolStripMenuItem.Text" xml:space="preserve">
+    <value>Export to World Weather Record</value>
+  </data>
   <data name="mnuClimaticFile.Size" type="System.Drawing.Size, System.Drawing">
     <value>212, 22</value>
   </data>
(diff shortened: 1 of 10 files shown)


dannyparsons:
Thanks for this. Are the menu items hidden? It's fine to add them to frmMain but they should be hidden until the dialogs are implemented. Alternatively, you could have created the dialog but not added the menu items yet. Either is ok.

Also, it's good to reference the issue, but you don't want this pull request to close the issue, because that will be when you actually do the dialog. In your next PR when you implement the dialog, you can use "fixes #xxx" to reference the issue and close it when merged.

HawardKetoyoMsatsi:
Well noted, I have disabled the menu.

instat/frmMain.resx
```diff
@@ -1058,6 +1058,15 @@
   <data name="mnuClimaticFileExportToCPT.Text" xml:space="preserve">
     <value>Export to CPT...</value>
   </data>
+  <data name="ExportToWWRToolStripMenuItem.Enabled" type="System.Boolean, mscorlib">
+    <value>False</value>
```
lloyddewit, line 1062: Should both the `Enabled` **and** `Visible` properties be set to false? (we don't want the user to see a greyed-out menu item, we prefer the user not to see the menu item at all)

instat/frmMain.resx
```diff
@@ -1058,6 +1058,15 @@
...
+  </data>
+  <data name="ExportToWWRToolStripMenuItem.Size" type="System.Drawing.Size, System.Drawing">
```
lloyddewit, line 1064: I'm not familiar with the `resx` naming convention but should this new menu item be called `mnuExportToWWRToolStrip`?
Please could you check with one of the other developers? thanks

instat/frmMain.resx
```diff
@@ -3544,31 +3553,31 @@
     <value>Wind/Pollution Rose...</value>
   </data>
   <data name="mnuStructuredCircularOtherRosePlotsPercentileRose.Size" type="System.Drawing.Size, System.Drawing">
-    <value>180, 22</value>
+    <value>168, 22</value>
```
lloyddewit, line 3556: Do you know why these Rose plot values changed (I assume they aren't connected to this PR)?

lloyddewit (changes requested):

instat/frmMain.resx
dannyparsons: Yes `mnuExportToWWRToolStrip` would be a sensible name.

instat/frmMain.resx
dannyparsons: It does happen automatically sometimes, I am not really sure why. If it doesn't look any different in the design then it should be ok.

dannyparsons:
@HawardKetoyoMsatsi while you are here, could you also change the label for Climatic > Prepare > SPI to "SPI/SPEI" if Roger agrees. I will update the dialog itself.

rdstern:
Agreed.  Good idea to include SPEI.  I assume we can include our own evapotranspiration.  I did check with the package people if we could use their routine, but it seems to be limited to monthly data.

instat/frmMain.resx
HawardKetoyoMsatsi: I have fixed naming of the menu item mnuExportToWWRToolStripMenuItem .

HawardKetoyoMsatsi:
Since Roger agreed I have left the dialog for @dannyparsons to update the last suggestion raised concerning this dialog.

instat/frmMain.resx
HawardKetoyoMsatsi: I have set both enabled and visible properties to false

lloyddewit (approved):

lloyddewit:
@rdstern I approved from my side but it still requires an update from @dannyparsons

dannyparsons (approved):

PR #6119, opened 2020-12-11
Adding survival dialog to frmmain

lilyclements:
As requested in #6047 this PR adds the define survival object dialog to the main form.

### DIFF
--- a/instat/frmMain.resx
+++ b/instat/frmMain.resx
@@ -3622,7 +3622,7 @@
     <value>Low_Flow</value>
   </data>
   <data name="mnuStructuredSurvivalDefine.Size" type="System.Drawing.Size, System.Drawing">
-    <value>117, 22</value>
+    <value>180, 22</value>
   </data>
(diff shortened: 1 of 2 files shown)


lloyddewit (approved):

lloyddewit:
@rdstern I don't think there's anything you can test here. This PR just enables the dialog in the `Structured` menu under `Survival > Define`. If we merge this PR then this will allow you to test PR #6047. Please could you approve?

@dannyparsons I understand that our policy is to only make menu options visible when the functionality behind that option is implemented. I considered asking @lilyclements to hide the `Survival` menu option. However, in the `Structured` menu there are several menu options without connected functionality. Therefore we should also hide these menu options for consistency.
I suggest we leave the `Structured` menu as it is for now, but still apply our new 'hiding' policy for future PRs. If you prefer something else then please let me know, thanks

rdstern (approved):
ok

PR #6125, opened 2020-12-12
Fixed summary tables bug and display of invalid dates in tidy climatic data

dannyparsons:
- `weights` parameter passed as `NULL` and correctly checked in summary functions so that Summary Tables dialog doesn't give error
- invalid dates data frame in tidy climatic data now a `data.frame` (not `tibble`) in all cases so that all rows are printed in output

### DIFF
--- a/instat/static/InstatObject/R/Backend_Components/summary_functions.R
+++ b/instat/static/InstatObject/R/Backend_Components/summary_functions.R
@@ -632,12 +632,12 @@ summary_quantile_circular <- function (x, probs = seq(0, 1, 0.25), na.rm = FALSE
   else return(circular::quantile.circular(x, probs = probs, na.rm = na.rm, names = names, type = type)[[1]])
 }
 
-summary_mean <- function (x, add_cols, weights = "", na.rm = FALSE, trim = 0, na_type = "", ...) {
+summary_mean <- function (x, add_cols, weights = NULL, na.rm = FALSE, trim = 0, na_type = "", ...) {
   if( length(x)==0 || (na.rm && length(x[!is.na(x)])==0) ) return(NA)
   else {
     if(na.rm && na_type != "" && !na_check(x, na_type = na_type, ...)) return(NA)
     else {
-      if (missing(weights)) 
+      if (missing(weights) || is.null(weights))
         return(mean(x, na.rm = na.rm, trim = trim))
       else 
         return(stats::weighted.mean(x, w = weights, na.rm = na.rm))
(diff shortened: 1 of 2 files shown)


instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -632,20 +632,20 @@ summary_quantile_circular <- function (x, probs = seq(0, 1, 0.25), na.rm = FALSE
...
 }
 
-summary_mean <- function (x, add_cols, weights = "", na.rm = FALSE, trim = 0, na_type = "", ...) {
+summary_mean <- function (x, add_cols, weights = NULL, na.rm = FALSE, trim = 0, na_type = "", ...) {
   if( length(x)==0 || (na.rm && length(x[!is.na(x)])==0) ) return(NA)
   else {
     if(na.rm && na_type != "" && !na_check(x, na_type = na_type, ...)) return(NA)
     else {
-      if (missing(weights)) 
+      if (missing(weights) || is.null(weights))
```
lloyddewit, line 640: if `weights` was not specified by the calling function then it will be `null` by default. Therefore the `missing(weights)` check is not needed.
This comment also applies to the other changed functions below.

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -632,20 +632,20 @@ summary_quantile_circular <- function (x, probs = seq(0, 1, 0.25), na.rm = FALSE
   else return(circular::quantile.circular(x, probs = probs, na.rm = na.rm, names = names, type = type)[[1]])
 }
 
-summary_mean <- function (x, add_cols, weights = "", na.rm = FALSE, trim = 0, na_type = "", ...) {
+summary_mean <- function (x, add_cols, weights = NULL, na.rm = FALSE, trim = 0, na_type = "", ...) {
   if( length(x)==0 || (na.rm && length(x[!is.na(x)])==0) ) return(NA)
   else {
```
lloyddewit, line 637: This function could be reduced to 4 lines
```
  if( length(x)==0 || (na.rm && length(x[!is.na(x)])==0) ) return(NA)
  else if (na.rm && na_type != "" && !na_check(x, na_type = na_type, ...)) return(NA)
  else if (is.null(weights)) return(mean(x, na.rm = na.rm, trim = trim))
  else return(stats::weighted.mean(x, w = weights, na.rm = na.rm))
```
This comment also applies to the other changed functions below.

instat/static/InstatObject/R/Backend_Components/summary_functions.R
dannyparsons: You're right that would be more concise.

instat/static/InstatObject/R/Backend_Components/summary_functions.R
dannyparsons: This looks better, though I would prefer to keep `else if` and `else` for clarity.

PR #6140, opened 2020-12-28
NCMP Dialogs

### LINKED ISSUE #6131: NCMP Dialogs
This is for a new menu of dialogs: Climatic > NCMP. NCMP stands for National Climate Monitoring Products. They are defined by WMO https://www.metoffice.gov.uk/hadobs/opace2_tt_ncmp/. In the folder here https://github.com/africanmathsinitiative/R-Instat/tree/master/instat/static/InstatObject/R/Backend_Components/NCMP each file defines a function which is based on R scripts produced by WMO for NCMP.

The calculations follow a series and the dialogs needs to be done in order. The output from previous steps are the input to the next steps.

So there will be 6 dialogs in the sub menu:
- Indices
- Variogram
- Region Average
- Trends Graphs
- Count Records
- Summary File

Here is a script showing how to use the functions [script.zip](https://github.com/africanmathsinitiative/R-Instat/files/5721139/script.zip). Here is the data used in the scripts [monthly (2).zip](https://github.com/africanmathsinitiative/R-Instat/files/5721142/monthly.2.zip)

In the R files I've tried to describe all the parameters so the dialog designs should be quite straightforward. The input are mainly data frames and column names. I suggest we don't include the optional parameters on the dialogs for now, so that we get main functionality working first.

All but the last function requires station metadata input. I suggest this could be done on a sub dialog, similarly to as described for another dialog https://github.com/africanmathsinitiative/R-Instat/issues/6110. Doing it this way means we only need to create one subdialog and use it in all the main dialogs that need it. The parameters are consistent across the functions.

Most of the functions have a `ncmp_folder` parameter which should be a folder selection. See the Import from IRI dialog for how this is done.

The design should be straightforward based on the parameters. There are a couple of dialogs with some differences:

**Region Average**
For `p4_region_average` there are two data frame inputs required (a2 and a3). But for a3 we don't need to choose any columns so I suggest to have the normal controls for a2 data and columns, and then use a `ucrDataFrame` to choose a3 data. That should mean there's enough room on the dialog.

**Trends Graphs**
Here we need a3 and a4 data and columns, so I suggest we use two `ucrSelectorByDataFrameAddRemove` controls, so the dialog will sort of be in two sections, like Export to CPT dialog.

lilyclements:
In progress.

Fixes #6131 

TODO:
* More rigid bug tests for p2-p4
* Bug test p5 onwards
* Issues with saving on dlgVariogram - it saves, but gives an error
* Issue with the nuds not giving reading the parameters unless you change the value.

### DIFF
--- a/instat/dlgClimaticNCMPCountRecords.Designer.vb
+++ b/instat/dlgClimaticNCMPCountRecords.Designer.vb
@@ -37,6 +37,7 @@ Partial Class dlgClimaticNCMPCountRecords
         Me.lblYear = New System.Windows.Forms.Label()
         Me.ucrReceiverStation = New instat.ucrReceiverSingle()
         Me.ucrSelectorForA2 = New instat.ucrSelectorByDataFrameAddRemove()
+        Me.ucrBase = New instat.ucrButtons()
         Me.grpOptions.SuspendLayout()
         Me.SuspendLayout()
         '
@@ -198,11 +199,19 @@ Partial Class dlgClimaticNCMPCountRecords
         Me.ucrSelectorForA2.Size = New System.Drawing.Size(210, 180)
         Me.ucrSelectorForA2.TabIndex = 0
         '
+        'ucrBase
+        '
+        Me.ucrBase.Location = New System.Drawing.Point(10, 363)
+        Me.ucrBase.Name = "ucrBase"
+        Me.ucrBase.Size = New System.Drawing.Size(405, 52)
+        Me.ucrBase.TabIndex = 17
+        '
         'dlgClimaticNCMPCountRecords
         '
         Me.AutoScaleDimensions = New System.Drawing.SizeF(6.0!, 13.0!)
         Me.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Font
-        Me.ClientSize = New System.Drawing.Size(397, 432)
+        Me.ClientSize = New System.Drawing.Size(415, 419)
+        Me.Controls.Add(Me.ucrBase)
         Me.Controls.Add(Me.ucrSaveCountRecords)
         Me.Controls.Add(Me.ucrSelectorForA4)
         Me.Controls.Add(Me.cmdStationMetadata)
@@ -241,4 +250,5 @@ Partial Class dlgClimaticNCMPCountRecords
     Friend WithEvents lblYear As Label
     Friend WithEvents ucrReceiverStation As ucrReceiverSingle
     Friend WithEvents ucrSelectorForA2 As ucrSelectorByDataFrameAddRemove
+    Friend WithEvents ucrBase As ucrButtons
 End Class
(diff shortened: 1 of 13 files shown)


dannyparsons:
- In the dialogs with saving files, it should be a folder selector not a file selector (see Browse Folder on NetCDF Import as an example)
- When you open the sub dialog for the first time it doesn't have any data selected. The selector should reset on first load which will ensure a default data frame is selected
- The default data frame names for saving should be in lower case
- `p5` outputs a list of two data frames as well as saving graphs, so this needs to be saved
- In the `p5` dialog, can you add the parameters `nyba` and `nyea`? These are the min and max years for trend calculation. I thought they weren't important to include but I think they are now. Default is 1950 - 2019.
- p7 also returns a data frame as well as saving a file so that needs to be saved

lilyclements:
@dannyparsons thank you for your review and comments. I have sorted out the bullet points you have listed, and additionally updated the R functions from the files you emailed earlier today. These dialogs seem to be working, aside from one (small, I hope) point -

Earlier today we discussed a different way of saving the data frames for region average (and trend graphs) since the output is a list. So I changed the ucrSave to be a label and ucrInput and set a new base function in the dialog to have the function `data_book$import_data`. However, changing the name of the new data frame for the ucrInput in the dialog is not going as I initially thought it would.

For example, with the Region Average dialog, I might type in the ucrInput for the new data frame names to be a prefix starting with `region_average_df`. This then gives the following R code:

```
[...]
region_average_df <- p4_region_average(a2=Indices1, station="station", year="Year", month="month", a3=variogram1, station_df=Metadata, name="station", lat="lat", lon="lon", nye=2019, uncode=288, res=1)
data_book$import_data(data_tables=region_average_df)
```

But the two new data frames are given their own, much more useful, names from lines 428-429 in the R code for `p4_region_average` (in this case, `ghana_avg_1_annual` and `ghana_region_avg`). Do we want to offer the ability to change these data frame names? If so, can we discuss a bit more how to go about this?

dannyparsons:
Yes, this is a good point. Ideally we could rename the data frames, but for now I don't mind them being fixed.

So I think you can remove the textbox completely and have a fixed name for `region_average_df` in the code (it could be `region_average_dfs`).

If we wanted, the easiest way to change the names is probably as new parameters to `p4_region_average()`, but that's not a priority now.

lilyclements:
@dannyparsons thanks I have made these changes now.
I think these dialogs are ready for review. The only other thing is with the UN code if I should perhaps add a tooltip with the link to the website to view the UN codes?
Should we wait until after I have compared the original NCMP R code to these dialogs code before I put these up for review, or are you happy for me to put these up for review (and I'll compare the R outputs in the meantime)

dannyparsons (approved):

dannyparsons:
Lily and I have been reviewing this and we've compared outputs from the original R-NCMP software to confirm it has the same functionality.

Issue #5974, opened 2020-09-10
Adding the conditional quantile plot


rdstern:
In #5928 we describe the set of dialogues needed for the graphical comparison of station and gridded data.  One of these is the conditional quantiles plot from the openair package.  [Chapter 19](https://davidcarslaw.com/files/openairmanual.pdf) of the sort of book on this openair package describes their use.  
I suggest @HelenaMwangila could design this dialogue and then try to implement it.  Once designed it will be similar to some existing dialogues and this should help with the coding.  The item is already in the menu under Climatic > Compare  > Conditional Quantiles.

As an initial step please could @Wycklife add a blank dialogue for Helena to start.  (At the same time could @Wycklife also add a blank dialogue for the time sereis and the seasonal plots.  Then @dannyparsons will be working on those dialogues, at least initially.

HelenaMwangila:
After looking at the controlQuantile() function, I tried to design a layout of the dialogue.
![image](https://user-images.githubusercontent.com/67306949/93300465-7d5a1c00-f7ff-11ea-97c9-aed6d997edaf.png)
I would like any useful input to improve it, before moving to the next step.

dannyparsons:
This looks good. For the Compare section in general we need to agree on a consistent name for the two column being compared. Lots of different combinations are used in different packages.

I quite like "Reference" and "Estimates" which is general enough for different uses and quite clear which column to put in each place. Reference is a bit more general than Observations and so it would allow, for example, comparing two satellite products.

rdstern:
@N-thony you were asking for a new task.  Could you take this on?  It completes our initial set of dialogues to compare station and satellite data.

N-thony:
@rdstern is sure! I would like to work on this

shadrackkibet:
The option  `openair::conditionalQuantile()` is flexible enough. @N-thony please proceed with this implementation. I am available to support in case you need help.

PR #6144, opened 2021-01-01
Implementing the Conditional Quantile and conditionalEval Options

N-thony:
@rdstern I have implemented the Conditional Quantile dialog of issue #5974. Could you please check this dialog if it works well? Thanks

### DIFF
--- a/instat/dlgConditionalQuantilePlot.vb
+++ b/instat/dlgConditionalQuantilePlot.vb
@@ -194,7 +194,7 @@ Public Class dlgConditionalQuantilePlot
     Private Sub TestOkEnabled()
         If rdoQuantiles.Checked AndAlso Not ucrReceiverObservations.IsEmpty() AndAlso Not ucrReceiverPredictions.IsEmpty() AndAlso ucrSavePlot.IsComplete AndAlso ucrNudBin.GetText <> "" AndAlso ucrNudKeyColumns.GetText <> "" Then
             ucrBase.OKEnabled(True)
-        ElseIf rdoEvaluation.Checked AndAlso Not ucrReceiverObservations.IsEmpty() AndAlso Not ucrReceiverPredictions.IsEmpty() AndAlso ucrSavePlot.IsComplete AndAlso Not ucrReceiverStatistics.IsEmpty Then
+        ElseIf rdoEvaluation.Checked AndAlso Not ucrReceiverObservations.IsEmpty() AndAlso Not ucrReceiverPredictions.IsEmpty() AndAlso ucrSavePlot.IsComplete Then
             ucrBase.OKEnabled(True)
         Else
             ucrBase.OKEnabled(False)
--- a/instat/dlgConditionalQuantilePlot.vb
+++ b/instat/dlgConditionalQuantilePlot.vb
@@ -142,4 +142,5 @@ Public Class dlgConditionalQuantilePlot
     Private Sub ucrReceiverObservations_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverObservations.ControlContentsChanged, ucrSavePlot.ControlContentsChanged, ucrReceiverPredictions.ControlContentsChanged
         TestOkEnabled()
     End Sub
+
 End Class
--- a/instat/dlgConditionalQuantilePlot.vb
+++ b/instat/dlgConditionalQuantilePlot.vb
@@ -95,16 +95,15 @@ Public Class dlgConditionalQuantilePlot
         ucrNudBin.Minimum = 1
         ucrNudBin.SetRDefault(31)
 
-        ucrInputMinBin.SetParameter(New RParameter("min.bin", 5))
-        ucrInputMinBin.SetValidationTypeAsNumeric()
-        ucrInputMinBin.SetRDefault("c(10,20)")
-        ucrInputMinBin.AddQuotesIfUnrecognised = False
+        ucrNudMinNumPoints.SetParameter(New RParameter("min.bin", 5))
+        ucrNudMinNumPoints.Minimum = 5
+        ucrNudMinNumPoints.SetRDefault(10)
 
-        ttConditionalQuatilePlot.SetToolTip(ucrInputMinBin.txtInput, "Min.bin can also be length of two e.g. min.bin = c(10,20)")
+        ' ttConditionalQuatilePlot.SetToolTip(ucrNudMinNumPoints.txtInput, "Min.bin can also be length of two e.g. min.bin = c(10,20)")
 
-        ucrNudKeyColumns.SetParameter(New RParameter("key.columns", 9))
-        ucrNudKeyColumns.Minimum = 1
-        ucrNudKeyColumns.SetRDefault(2)
+        ucrNudColumns.SetParameter(New RParameter("key.columns", 9))
+        ucrNudColumns.Minimum = 1
+        ucrNudColumns.SetRDefault(2)
 
         ucrInputLegendPosition.SetParameter(New RParameter("key.position", 10))
         dctLegendPosition.Add("Bottom", Chr(34) & "bottom" & Chr(34))
@@ -113,19 +112,17 @@ Public Class dlgConditionalQuantilePlot
         dctLegendPosition.Add("Left", Chr(34) & "left" & Chr(34))
         dctLegendPosition.Add("None", Chr(34) & "none" & Chr(34))
         ucrInputLegendPosition.SetItems(dctLegendPosition)
+        ucrInputLegendPosition.SetRDefault(Chr(34) & "none" & Chr(34))
         ucrInputLegendPosition.SetDropDownStyleAsNonEditable()
-        ucrInputLegendPosition.SetRDefault(Chr(34) & "bottom" & Chr(34))
-
 
         ucrPnlOptions.AddRadioButton(rdoQuantiles)
         ucrPnlOptions.AddRadioButton(rdoEvaluation)
         ucrPnlOptions.AddFunctionNamesCondition(rdoQuantiles, "conditionalQuantile")
         ucrPnlOptions.AddFunctionNamesCondition(rdoEvaluation, "conditionalEval")
-        ucrPnlOptions.AddToLinkedControls({ucrNudKeyColumns, ucrInputLegendPosition, ucrInputMinBin, ucrReceiverMultipleFacet}, {rdoQuantiles}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrPnlOptions.AddToLinkedControls({ucrInputLegendPosition, ucrNudMinNumPoints, ucrReceiverMultipleFacet}, {rdoQuantiles}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
         ucrPnlOptions.AddToLinkedControls({ucrReceiverMultipleObs, ucrReceiverMultiplePred, ucrReceiverStatistics, ucrReceiverType}, {rdoEvaluation}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
-        ucrNudKeyColumns.SetLinkedDisplayControl(lblKey)
-        ucrInputLegendPosition.SetLinkedDisplayControl(lblKeyPositions)
-        ucrInputMinBin.SetLinkedDisplayControl(lblColNumber)
+        ucrInputLegendPosition.SetLinkedDisplayControl(lblLegendPositions)
+        ucrNudMinNumPoints.SetLinkedDisplayControl(lblColNumber)
         ucrReceiverMultipleFacet.SetLinkedDisplayControl(lblTypes)
         ucrReceiverMultipleObs.SetLinkedDisplayControl(lblVarObserved)
         ucrReceiverMultiplePred.SetLinkedDisplayControl(lblVarPredicted)
@@ -147,6 +144,7 @@ Public Class dlgConditionalQuantilePlot
         ucrConditionalQuantilePlotSelector.Reset()
         ucrSavePlot.Reset()
         ucrReceiverObservations.SetMeAsReceiver()
+        SetVisibleColumnOption()
 
         clsConditionalQuantileFunction.SetPackageName("openair")
         clsConditionalQuantileFunction.SetRCommand("conditionalQuantile")
@@ -173,7 +171,8 @@ Public Class dlgConditionalQuantilePlot
         ucrReceiverPredictions.SetRCode(clsConditionalQuantileFunction, bReset)
         ucrReceiverMultipleFacet.SetRCode(clsConditionalQuantileFunction, bReset)
         ucrNudBin.SetRCode(clsConditionalQuantileFunction, bReset)
-        ucrNudKeyColumns.SetRCode(clsConditionalQuantileFunction, bReset)
+        ucrNudMinNumPoints.SetRCode(clsConditionalQuantileFunction, bReset)
+        ucrNudColumns.SetRCode(clsConditionalQuantileFunction, bReset)
         ucrInputLegendPosition.SetRCode(clsConditionalQuantileFunction, bReset)
         ucrSavePlot.SetRCode(clsConditionalQuantileFunction, bReset)
         ucrSavePlot.AddAdditionalRCode(clsConditionalEvalFunction)
@@ -190,9 +189,18 @@ Public Class dlgConditionalQuantilePlot
         SetRcodeForControls(True)
         TestOkEnabled()
     End Sub
+    Private Sub SetVisibleColumnOption()
+        If ucrInputLegendPosition.GetText = "None" Then
+            lblColumns.Visible = False
+            ucrNudColumns.Visible = False
+        Else
+            lblColumns.Visible = True
+            ucrNudColumns.Visible = True
+        End If
+    End Sub
 
     Private Sub TestOkEnabled()
-        If rdoQuantiles.Checked AndAlso Not ucrReceiverObservations.IsEmpty() AndAlso Not ucrReceiverPredictions.IsEmpty() AndAlso ucrSavePlot.IsComplete AndAlso ucrNudBin.GetText <> "" AndAlso ucrNudKeyColumns.GetText <> "" Then
+        If rdoQuantiles.Checked AndAlso Not ucrReceiverObservations.IsEmpty() AndAlso Not ucrReceiverPredictions.IsEmpty() AndAlso ucrSavePlot.IsComplete AndAlso ucrNudBin.GetText <> "" AndAlso ucrNudColumns.GetText <> "" Then
             ucrBase.OKEnabled(True)
         ElseIf rdoEvaluation.Checked AndAlso Not ucrReceiverObservations.IsEmpty() AndAlso Not ucrReceiverPredictions.IsEmpty() AndAlso ucrSavePlot.IsComplete Then
             ucrBase.OKEnabled(True)
@@ -201,7 +209,7 @@ Public Class dlgConditionalQuantilePlot
         End If
     End Sub
 
-    Private Sub ucrReceiverObservations_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverObservations.ControlContentsChanged, ucrSavePlot.ControlContentsChanged, ucrReceiverPredictions.ControlContentsChanged, ucrReceiverStatistics.ControlContentsChanged, ucrNudBin.ControlContentsChanged, ucrNudKeyColumns.ControlContentsChanged
+    Private Sub ucrReceiverObservations_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverObservations.ControlContentsChanged, ucrSavePlot.ControlContentsChanged, ucrReceiverPredictions.ControlContentsChanged, ucrReceiverStatistics.ControlContentsChanged, ucrNudBin.ControlContentsChanged, ucrNudColumns.ControlContentsChanged
         TestOkEnabled()
     End Sub
 
@@ -221,5 +229,9 @@ Public Class dlgConditionalQuantilePlot
             ucrSavePlot.Location = New Point(ucrSavePlot.Location.X, 401)
             cmdPlotOptions.Text = "Conditional Eval Options"
         End If
+    End Sub
+
+    Private Sub ucrInputLegendPosition_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputLegendPosition.ControlValueChanged
+        SetVisibleColumnOption()
     End Sub
 End Class
--- a/instat/dlgConditionalQuantilePlot.vb
+++ b/instat/dlgConditionalQuantilePlot.vb
@@ -80,7 +80,6 @@ Public Class dlgConditionalQuantilePlot
         ucrInputLegendPosition.SetItems(dctLegendPosition)
         ucrInputLegendPosition.SetDropDownStyleAsNonEditable()
         ucrInputLegendPosition.SetRDefault(Chr(34) & "bottom" & Chr(34))
-        ucrInputLegendPosition.bAllowNonConditionValues = True
 
         ucrSavePlot.SetPrefix("conditionalquantile")
         ucrSavePlot.SetCheckBoxText("Save Graph")
--- a/instat/dlgConditionalQuantilePlot.vb
+++ b/instat/dlgConditionalQuantilePlot.vb
@@ -35,6 +35,7 @@ Public Class dlgConditionalQuantilePlot
     End Sub
 
     Private Sub InitialiseDialog()
+        Dim dctLegendPosition As New Dictionary(Of String, String)
         ucrBase.clsRsyntax.iCallType = 3
 
         ucrConditionalQuantilePlotSelector.SetParameter(New RParameter("mydata", 0))
@@ -51,12 +52,17 @@ Public Class dlgConditionalQuantilePlot
         ucrReceiverPredictions.SetParameterIsString()
         ucrReceiverPredictions.SetIncludedDataTypes({"numeric"})
 
+        ucrReceiverMultipleFacet.SetParameter(New RParameter("type", 2))
+        ucrReceiverMultipleFacet.Selector = ucrConditionalQuantilePlotSelector
+        ucrReceiverMultipleFacet.SetParameterIsString()
+        ucrReceiverMultipleFacet.SetIncludedDataTypes({"factor", "character"})
+        ucrReceiverMultipleFacet.bExcludeFromSelector = True
 
-        ucrReceiverType.SetParameter(New RParameter("type", 2))
-        ucrReceiverType.Selector = ucrConditionalQuantilePlotSelector
-        ucrReceiverType.SetParameterIsString()
-        ucrReceiverType.bExcludeFromSelector = True
-        ucrReceiverType.SetClimaticType("station")
+        'ucrReceiverType.SetParameter(New RParameter("type", 2))
+        'ucrReceiverType.Selector = ucrConditionalQuantilePlotSelector
+        'ucrReceiverType.SetParameterIsString()
+        'ucrReceiverType.bExcludeFromSelector = True
+        'ucrReceiverType.SetClimaticType("station")
 
         ucrChkBin.SetText("Bins")
         ucrChkBin.SetParameter(New RParameter("bins", 3), bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)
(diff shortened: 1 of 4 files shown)


rdstern:
@N-thony that's great.  I'll check.

rdstern:
This is great and it is working already!  Well done.
a) Could you change the control currently called Data Split Methods, Type into Facet By (Optional):  And put it on the right-hand side below the other 2 controls.  It can take up to 2 factors, so perhaps make it into a multiple receiver.  
b) Move up the Plot Options button, so it is in the same position as in the other plotting dialogues - so just below the data selector.  Until it is active it should be disabled.
c) I tried bins.  Can you make the default into 31, which I think is the default in the function.  I am not sure what min.bin is doing, but make the default 10 for now.
d) It will be good to implement the alternative key positions and number of lines.  I leave to you whether rthis needs the sub-dialogue or can be added to the main dialogue.
e) Now a computing problem for you.  It (by default) saves a lattice graph.  It is saved in lasg_graph is you don't give it a separate name.  But it can be huge!  So for the dodoma data in the R-Instat library I used tmax and tmin and did the fraph by month.  Then the file (with this one graph added to the R-Instat data book) becomes 219 mbytes.  It is less than 2 mbytes otherwise.  I wonder what can be done about this.  It is slowing my machine down and I have a pretty fast machine!

N-thony:
@rdstern I solved the issues you raised yesterday.  Could you confirm that the issues are resolved?
For instance, I have added key positions to the main dialog. I may add the sub dialog if you request more flexibility in the graph.

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,146 @@
...
+        ucrBase.clsRsyntax.iCallType = 3
+
+        ucrConditionalQuantilePlotSelector.SetParameter(New RParameter("mydata", 0))
+        ucrConditionalQuantilePlotSelector.SetParameterIsrfunction()
+
+        ucrReceiverObservations.SetParameter(New RParameter("obs", 0))
+        ucrReceiverObservations.Selector = ucrConditionalQuantilePlotSelector
+        ucrReceiverObservations.SetParameterIsString()
+        ucrReceiverObservations.SetIncludedDataTypes({"numeric"})
+        ucrReceiverObservations.bAutoFill = True
```
shadrackkibet, line 48: I think you do not need to autofill here.

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,146 @@
...
+    End Sub
+
+    Private Sub InitialiseDialog()
+        Dim dctLegendPosition As New Dictionary(Of String, String)
+        ucrBase.clsRsyntax.iCallType = 3
+
+        ucrConditionalQuantilePlotSelector.SetParameter(New RParameter("mydata", 0))
+        ucrConditionalQuantilePlotSelector.SetParameterIsrfunction()
+
+        ucrReceiverObservations.SetParameter(New RParameter("obs", 0))
```
shadrackkibet, line 44: Check your parameter positions here, these two can't have the same position.

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,146 @@
...
+        ucrReceiverMultipleFacet.Selector = ucrConditionalQuantilePlotSelector
+        ucrReceiverMultipleFacet.SetParameterIsString()
+        ucrReceiverMultipleFacet.SetIncludedDataTypes({"factor", "character"})
+        ucrReceiverMultipleFacet.bExcludeFromSelector = True
+
+        'ucrReceiverType.SetParameter(New RParameter("type", 2))
+        'ucrReceiverType.Selector = ucrConditionalQuantilePlotSelector
+        'ucrReceiverType.SetParameterIsString()
+        'ucrReceiverType.bExcludeFromSelector = True
+        'ucrReceiverType.SetClimaticType("station")
```
shadrackkibet, line 65: Remove it? Otherwise, comment on why this is still needed.

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,146 @@
...
+
+        ucrChkBin.SetText("Bins")
+        ucrChkBin.SetParameter(New RParameter("bins", 3), bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)
+        ucrNudBin.SetParameter(New RParameter("bins", 3))
```
shadrackkibet, line 69: Two controls with the same parameter?

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,146 @@
...
+        ucrChkBin.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
+        ucrNudBin.Minimum = 1
+        ' ucrChkBin.SetRDefault("TRUE")
```
shadrackkibet, line 72: Remove?

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,146 @@
...
+
+        ucrChkMinBin.SetText("Min.bin")
+        ucrChkMinBin.SetParameter(New RParameter("min.bin", 4), bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)
+        ucrNudMinBin.SetParameter(New RParameter("min.bin", 4))
```
shadrackkibet, line 76: Also here.

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,146 @@
...
+        ucrNudMinBin.Minimum = 1
+        ' ucrChkMinBin.SetRDefault("TRUE")
```
shadrackkibet, line 78: Remove?

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,146 @@
...
+
+        ucrChkKeyColumns.SetText("Key.columns")
+        ucrChkKeyColumns.SetParameter(New RParameter("key.columns", 5), bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)
+        ucrNudKeyColumns.SetParameter(New RParameter("key.columns", 5))
```
shadrackkibet, line 82: Also here.

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,146 @@
...
+        ucrChkKeyColumns.AddToLinkedControls(ucrNudKeyColumns, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=2)
+
+        ucrChkBin.AddToLinkedControls(ucrNudBin, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=31)
+        ucrChkMinBin.AddToLinkedControls(ucrNudMinBin, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=10)
+
+        ucrChkLegendPosition.SetText("Key.position")
+        ucrChkLegendPosition.AddToLinkedControls(ucrInputLegendPosition, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:="Bottom")
+        ucrInputLegendPosition.SetDropDownStyleAsNonEditable()
+        ucrInputLegendPosition.SetParameter(New RParameter("key.position", 6))
+        ' dctLegendPosition.Add("None", Chr(34) & "none" & Chr(34))
```
shadrackkibet, line 94: Remove?

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,146 @@
...
+    Private Sub SetDefaults()
+        clsConditionalQuantileFunction = New RFunction
+
+        ucrConditionalQuantilePlotSelector.Reset()
+        ucrSavePlot.Reset()
+        ucrReceiverObservations.SetMeAsReceiver()
+
+        clsConditionalQuantileFunction.SetPackageName("openair")
+        clsConditionalQuantileFunction.SetRCommand("conditionalQuantile")
+
```
shadrackkibet, line 120: You need to set the assignment to `last_graph` here so that the save control resets properly i.e it is unchecked when the dialog is reset.

shadrackkibet (changes requested):

instat/static/InstatObject/R/Rsetup.R
```diff
@@ -177,7 +177,7 @@ load_R_Instat_packages <- function() {
   # PCICt needed to access PCICt class when importing NET cdf files
   # ggmosaic because geom_mosaic aes only work when ggmosaic is loaded
   # wakefield because many functions do not work without loading (https://github.com/trinker/wakefield/issues/11)
-  packs_to_load <- c("plyr", "dplyr", "ggplot2", "ggthemes", "svglite", "ggfortify", "PCICt", "sp", "ggmosaic", "wakefield", "circular")
+  packs_to_load <- c("plyr", "dplyr", "ggplot2", "ggthemes", "svglite", "ggfortify", "PCICt", "sp", "ggmosaic", "wakefield", "circular", "latticeExtra")
```
shadrackkibet, line 180: Add a comment above explaining why `latticeExtra` need to be loaded.

shadrackkibet:
> b) Move up the Plot Options button, so it is in the same position as in the other plotting dialogues - so just below the data selector. Until it is active it should be disabled.

Make sure this button is disabled now since it is not yet working!

> c) I tried bins. Can you make the default into 31, which I think is the default in the function. I am not sure what min.bin is doing, but make the default 10 for now.

This can be a vector e.g `c(10,20)` therefore this should be an input textbox so that it allows the user to type in a vector. Perhaps you need to add a tooltip to explain this and also give an example to guide the user.

shadrackkibet:
Comment on design and TestOk.

image

- Align labels properly with corresponding receivers.
- Rename checkboxes to **Key Position** and **Key Column(s)** .
- When `bins, min.bin` and `key columns` are empty as shown in the figure below, Ok is enabled. On click Ok it doesn't give an error but the output code is as shown below.
```
# Code generated by the dialog, Conditional Quantile Plot
Dagoretti_JKIA <- data_book$get_data_frame(data_name="Dagoretti_JKIA")
conditionalquantile2 <- openair::conditionalQuantile(obs="TMPMAX", mod="tmax_MERRA2", type="station", mydata=Dagoretti_JKIA, bins=, min.bin=, key.columns=, key.position="bottom")
data_book$add_graph(graph_name="conditionalquantile2", graph=conditionalquantile2, data_name="Dagoretti_JKIA")
```
image

To sort out these you need to set Rdefaults.

instat/dlgConditionalQuantilePlot.vb
N-thony: Yes, I need to set the same parameter to the ucrChk because it controls if the parameter is added or removed

N-thony:
@shadrackkibet I solved the issues you raised a couple of hours ago. Could you confirm that the issues are resolved? Thanks.

instat/dlgConditionalQuantilePlot.vb
shadrackkibet: The NUD should pass the parameter while the checkbox is only used to hide/unhide the NUD via linking controls.

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -41,61 +40,54 @@ Public Class dlgConditionalQuantilePlot
...
-
         ucrChkBin.SetText("Bins")
-        ucrChkBin.SetParameter(New RParameter("bins", 3), bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)
-        ucrNudBin.SetParameter(New RParameter("bins", 3))
-        ucrChkBin.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
-        ucrNudBin.Minimum = 1
-        ' ucrChkBin.SetRDefault("TRUE")
+        ucrChkBin.SetParameter(New RParameter("min.bin"), bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)
+        ucrNudBin.SetParameter(New RParameter("bins", 4))
+        ucrNudBin.SetMinMax(1, Integer.MaxValue)
```
shadrackkibet, line 62: Did you set the R defaults?

N-thony: I have added conditions in TestOk to avoid that.

instat/dlgConditionalQuantilePlot.vb
N-thony: Okay, but this will not allow that when the dialog is closed and ucrChk  is checked and open, ucrChk becomes unchecked.
I would like to see the code where this is done so I can understand how it works.

instat/dlgConditionalQuantilePlot.vb
shadrackkibet: You need to set R defaults as well.

instat/dlgConditionalQuantilePlot.vb
shadrackkibet: You need to set conditions to ensure that checkboxes retain state on reopening. In this case you should use `.AddParameterPresentCondition`. for example

```
    ucrChkBin.AddParameterPresentCondition(True, "bins")
    ucrChkBin.AddParameterPresentCondition(False, "bins", False)
```

N-thony: Okay

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,144 @@
...
+    End Sub
+
+    Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
+        SetDefaults()
+        SetRcodeForControls(True)
+        TestOkEnabled()
+    End Sub
+
+    Private Sub TestOkEnabled()
+        If Not ucrReceiverObservations.IsEmpty() AndAlso Not ucrReceiverPredictions.IsEmpty() AndAlso ucrSavePlot.IsComplete AndAlso ucrNudBin.GetText <> "" AndAlso ucrNudKeyColumns.GetText <> "" AndAlso Not ucrInputMinBin.IsEmpty Then
```
shadrackkibet, line 128: The corresponding checkboxes should be included in TestOk. Otherwise, if these Nuds and input are empty and the checkboxes are not checked OK will not be enabled which might be confusing to the user.

N-thony: Correct! Thanks

N-thony:
@shadrackkibet I solved the issues you raised a couple of hours ago. Could you confirm that the issues are resolved? Thanks.

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,150 @@
...
+        End If
+    End Sub
+
+    Private Sub ucrReceiverObservations_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverObservations.ControlContentsChanged, ucrSavePlot.ControlContentsChanged, ucrReceiverPredictions.ControlContentsChanged, ucrNudBin.ControlContentsChanged, ucrNudKeyColumns.ControlContentsChanged, ucrInputMinBin.ControlContentsChanged, ucrChkBin.ControlContentsChanged, ucrChkKeyColumns.ControlContentsChanged, ucrChkMinBin.ControlContentsChanged
+        TestOkEnabled()
+    End Sub
+
+    Private Sub ucrInputMinBin_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputMinBin.ControlValueChanged
+        If ucrInputMinBin.Text.Contains(",") Then
+            clsConditionalQuantileFunction.AddParameter("min.bin", "c(" & ucrInputMinBin.GetText() & ")")
```
shadrackkibet, line 147: - This parameter is passed automatically by setting Rcode and manually here which contradicts the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle. Could you ensure this done only once? If not I suggest you add a comment here.
- What does the else to this do? should you remove the parameter? should it be a developer error? 
- Always add parameter positions!

instat/dlgConditionalQuantilePlot.vb
shadrackkibet: I still can't see where this is done. If you decided otherwise then you should explain why!

instat/dlgConditionalQuantilePlot.vb
N-thony: min.bin is already added automatically, so I have set the condition if in case the user enters min.bin of the length of two e.g min.bin = c(10,20).

N-thony:
@shadrackkibet I solved the issues you raised a couple of minutes ago. Could you confirm that the issues are resolved? Thanks.

instat/dlgConditionalQuantilePlot.vb
shadrackkibet: `ucrNudBin.SetRDefault(31)`

N-thony: Setting the Rdefault to to Nud doesn't allow the ucrChk to keep the state if you close and open the dialog when they are checked. Remember that I have set the condition to the ucrChk as follow:
    ucrChkBin.AddParameterPresentCondition(True, "bins")
    ucrChkBin.AddParameterPresentCondition(False, "bins", False)

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -144,7 +144,9 @@ Public Class dlgConditionalQuantilePlot
 
     Private Sub ucrInputMinBin_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputMinBin.ControlValueChanged
         If ucrInputMinBin.Text.Contains(",") Then
-            clsConditionalQuantileFunction.AddParameter("min.bin", "c(" & ucrInputMinBin.GetText() & ")")
+            clsConditionalQuantileFunction.AddParameter("min.bin", "c(" & ucrInputMinBin.GetText() & ")", iPosition:=5) 'this is min.bin is length of two e.g min.bin = c(10,20) 
+        Else
+            clsConditionalQuantileFunction.AddParameter("min.bin", ucrInputMinBin.GetText(), iPosition:=5) 'this is min.bin is length of two e.g min.bin = c(10,20)
```
shadrackkibet, line 149: Why the same comment twice?

N-thony: Oh, sorry. It was cc.

shadrackkibet:
When you reopen the dialog `min.bin` checkbox is always checked. The dialog should always retain its state on reopening.

N-thony:
> When you reopen the dialog `min.bin` checkbox is always checked. The dialog should always retain its state on reopening.

I don't know but mine works well. closing.opening min.bin checkbox keeps the state.

instat/dlgConditionalQuantilePlot.vb
shadrackkibet: That is right. Setting R default ensures that the parameter is hidden when its value is the set R default. This is useful because it is not necessary to pass parameters when defaults are defined internally in the R function. For example to following lines of code give the same output. 

```
openair::conditionalQuantile(mydata=Dagoretti_JKIA, obs="prcp_CHIRPS", mod="est_prcp_CHIRPSv2", type="station")
```

```
openair::conditionalQuantile(mydata=Dagoretti_JKIA, obs="prcp_CHIRPS", mod="est_prcp_CHIRPSv2", type="station", bins=31, min.bin=c(10,20), key.columns=2, key.position="bottom")
```

This means it is not necessary to pass `bins, min.bin, key.columns, key.position` with default values. I hope this makes sense.

N-thony: Exact!

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,152 @@
...
+        ucrReceiverMultipleFacet.bExcludeFromSelector = True
+
+        ucrNudBin.SetParameter(New RParameter("bins", 4))
+        ucrNudBin.SetMinMax(1, Integer.MaxValue)
+
+        ucrChkBin.AddParameterPresentCondition(True, "bins")
+        ucrChkBin.AddParameterPresentCondition(False, "bins", False)
+        ucrChkBin.SetText("Bins:")
+
+        ucrInputMinBin.SetParameter(New RParameter("min.bin", 5))
```
shadrackkibet, line 66: This parameter is added into the R function manually elsewhere. Why is this needed?

instat/dlgConditionalQuantilePlot.vb
shadrackkibet: Do you now understand why you need to set R defaults?

instat/dlgConditionalQuantilePlot.vb
N-thony: Oh yes, this correct!

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,152 @@
...
+        ucrInputMinBin.SetValidationTypeAsNumeric()
+        ucrInputMinBin.AddQuotesIfUnrecognised = False
+        ucrInputMinBin.bAllowNonConditionValues = True
```
shadrackkibet, line 69: Why are these two lines of code needed?

shadrackkibet:
As a principle please make sure that you are always able to explain what the code does and why each line of code you write is needed. If anything doesn't make sense to you always ask!

instat/dlgConditionalQuantilePlot.vb
N-thony: Yap

N-thony:
@shadrackkibet I solved the issues you raised a couple of hours ago. Could you confirm that the issues are resolved? Thanks.

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -57,46 +57,29 @@ Public Class dlgConditionalQuantilePlot
...
         dctLegendPosition.Add("Left", Chr(34) & "left" & Chr(34))
         dctLegendPosition.Add("Right", Chr(34) & "right" & Chr(34))
         dctLegendPosition.Add("Top", Chr(34) & "top" & Chr(34))
         dctLegendPosition.Add("Bottom", Chr(34) & "bottom" & Chr(34))
         dctLegendPosition.Add("None", Chr(34) & "none" & Chr(34))
         ucrInputLegendPosition.SetItems(dctLegendPosition)
-        ucrChkLegendPosition.AddParameterPresentCondition(True, "key.position")
-        ucrChkLegendPosition.AddParameterPresentCondition(False, "key.position", False)
+        ucrInputLegendPosition.SetDropDownStyleAsNonEditable()
+        ucrInputLegendPosition.SetRDefault("bottom")
```
shadrackkibet, line 81: chr(34)?

N-thony:
@shadrackkibet I solved the issues you raised a couple of minutes ago. Could you confirm that the issues are resolved? Thanks.

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,128 @@
...
+        ucrInputLegendPosition.SetParameter(New RParameter("key.position", 10))
+        dctLegendPosition.Add("Bottom", Chr(34) & "bottom" & Chr(34))
+        dctLegendPosition.Add("Top", Chr(34) & "top" & Chr(34))
+        dctLegendPosition.Add("Right", Chr(34) & "right" & Chr(34))
+        dctLegendPosition.Add("Left", Chr(34) & "left" & Chr(34))
+        dctLegendPosition.Add("None", Chr(34) & "none" & Chr(34))
+        ucrInputLegendPosition.SetItems(dctLegendPosition)
+        ucrInputLegendPosition.SetDropDownStyleAsNonEditable()
+        ucrInputLegendPosition.SetRDefault(Chr(34) & "bottom" & Chr(34))
+        ucrInputLegendPosition.bAllowNonConditionValues = True
```
shadrackkibet, line 83: Can you explain why should non-conditioned values be allowed here?

shadrackkibet:
I also think the disabled **Plot options** should be called **Conditional Quantile Options** because this will be a special sub dialog with options only used in this dialog.

instat/dlgConditionalQuantilePlot.vb
N-thony: To allow the control value to be set to a value that is not one of the valid values predefined in the dictionary.

N-thony:
> I also think the disabled **Plot options** should be called **Conditional Quantile Options** because this will be a special sub dialog with options only used in this dialog.

Yes, you are right!

instat/dlgConditionalQuantilePlot.vb
shadrackkibet: But only options specified in the dictionary are allowed here. Isn't that the reason why this input Combobox is non-editable?

N-thony: > But only options specified in the dictionary are allowed here. Isn't that the reason why this input Combobox is non-editable?

Yes, I see your point.

shadrackkibet: Therefore this should be removed.

N-thony: Done, thanks!

shadrackkibet:
Looks better now. @lloyddewit, please proceed with the necessary steps.

lloyddewit:
@shadrackkibet thank you for the review
@rdstern Please could you test that your comments from 04 Jan above are now resolved? thanks

rdstern:
@N-thony and @shadrackkibet well done as this seems to be working well now.
I have tried most of the options in the dialogue and don't seem to be able to break it.
Importantly I also note that the graph icon on the toolbar is working, so we can (I assume) now display Lattice graphs as well as ggplot graphs in the R-viewer through the toolbar icon.
So the changes that I suggest are purely cosmetic.  
To me the dialogue breaks (slightly) with an important principle in our proposed use of R-Instat.  R-Instat is designed to promote good statistical practice - even for inexperienced users.  One aspect of the good practice is that you are tempted to concentrate on important points when using a dialogue.  This includes using facets etc.
Now currently, when I use this dialogue I am drawn to the options on the left-hand side.  So, I am likely to ask "How many bins should I use, and what does this minimum number mean and what on earth is c(10,20) here.  

They are not important and so should not distract the user.  

I need some help here, because I am not sure what the "Minimum number of points" does?  Assuming we keep that option on the main dialogue, then do we really need to be able to alter both options?  If so, then it could become 2 up-down controls.  

I therefore checked back to the guide and there seems to be an important new (I assume new - or perhaps I missed it before) function called conditionalEval that extends the use of the conditional quantiles function.  So, let's think short-term and longer-term.  In the longer term I note that the chapter in the book is still called conditional quantiles, so our menu remains fine.  Then I suggest we will soon have 2 radio buttons at the top with a label `Conditional`, and then the first button is `Quantiles` and the second is `Evaluation`.

This is for a new issue and very interesting, but not urgent.

For now it would be good to get the current version merged as soon as possible.  So I suggest minimal changes as follows:

a) Change `Bins` to `Histogram Bins`  
b) Make the Minimum number control invisible for now.
c) Possibly add the buttons at the top, as described above, to show our intentions to add the Evaluation function later - with the Evaluation button visible, but disabled.
d) Write a new issue to add the Conditional Evaluation function once @N-thony has time.

Then let's merge it.

N-thony:
@rdstern I solved the comments about the small changes in the design. Could you please check it? Thanks

N-thony:
@rdstern I solved the comments about the small changes in the design. Could you please check it? Thanks

N-thony:
@rdstern In addition of the conditionalQuantile function, I have implemented the conditionalEval function. Could you please review? Thanks.

rdstern (commented):
@N-thony The controls on the left were empty - I am not sure what they should have contained?  And I was unable to see what happens, because Ok wasn't enabled.
Could you also tell me what data you are using to test?

N-thony:
> @N-thony The controls on the left were empty - I am not sure what they should have contained? And I was unable to see what happens, because Ok wasn't enabled.
> Could you also tell me what data you are using to test?

Oh sorry, I set the condition also to the Statistic Receiver. I think you need to fill this Receiver too.

N-thony:
@rdstern thanks for your comment above and the quick chat. This is now ready. Please could you test?

maxwellfundi:
@shadrackkibet @rdstern  Kindly this is ready for your re-review.

rdstern (changes requested):
@N-thony this is potentially looking very good indeed.  But I couldn't test it properly, because of an issue that needs to be fixed before you proceed.
It isn't just your problem, but this dialogue has brought it to a head.
I found elsewhere that a lattice graph (at least one generated through openair) can be very large.  A simple graph changed the file size in R-Instat from 2mbytes to 200 mbytes.  The graphs here are complicated to the extent that the automatic backup doesn't work, and I can't save the file, because of the inclusion of the graph.  
I don't know what's happening here - is it trying to save a bitmap for example.
This is a bit peripheral, but needs a serious investigation before we can move on.
You may wish to investigate quickly, but don't spend more than (say) 15 minutes, before asking others - perhaps @volloholic ?
@shadrackkibet is also testing, but needs to be aware of this issue first.

rdstern:
I start with the quantiles
a) It is looking very good
b) In the minimum number of points could you make it an up-down control.  Perhaps minimum 5 and no maximum.  Default 10.
c) Please also check that the value is registered in the command if it isn't the default?  I don't know what the default is?
d) Please change the label Key Position to Legend Position.  
e) Move Key Columns control to the same row as the Legend Position.  Call it just Columns.  Also if the Legend Position is set to None, then it is made not visible.

I still need to check the Eval in more detail.  It have me the following error when I tried a simple default with just the 2 variables and nothing in either multiple receiver.
image

I need also to check with @maxwellfundi about @shadrackkibet taking over completion of this and the Taylor diagram.  @N-thony has done well getting this far, but now has too many other priorities.

N-thony:
@rdstern this is ready. Could you test? Thanks

rdstern (changes requested):
@N-thony thanks

The one item remaining is the Facet multiple receiver.  Can you change it so it also permits factors.

N-thony:
@rdstern thanks for your comment. I think it's resolved. Could you re-test now?

rdstern (dismissed):
@N-thony great, thanks.
@dannyparsons and @lloyddewit This dialogue has tow parts, with different functions, each from openair.
We have now disabled the Eval, so the plotting can be merged.  Then I hope we can proceed quickly to check and add the Eval part later.

instat/dlgConditionalQuantilePlot.vb
```diff
@@ -1,3 +1,237 @@
...
+'
+' You should have received a copy of the GNU General Public License 
+' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports instat.Translations
+Public Class dlgConditionalQuantilePlot
+    Private bFirstLoad As Boolean = True
+    Private bReset As Boolean = True
+    Private iDialogHeight As Integer
+    Private iBaseMaxY
```
lloyddewit, line 22:
```suggestion
    Private iBaseMaxY As Integer
```

lloyddewit (changes requested):

lloyddewit (approved):

lloyddewit:
@rdstern I made one tiny change that should not affect the functionality. Please could you approve again? thanks

rdstern (approved):
I am approving, because what I found is trivial and I don't want to slow the process now.
I used the default, which has legend as None.  But it then does the default, which is legend = Bottom.
That can be changes within the next issue - which is mainly for the Evaluation part

lloyddewit:
@shadrackkibet Thank you for all the reviewing you did for this PR. You requested changes so you also need to approve. Please could you approve and merge? thanks

shadrackkibet (approved):

PR #6147, opened 2021-01-05
Improvement to the Display Daily Dialogue

### LINKED ISSUE #6099: Additional option in Climatic > Check Data > Display Daily Graph
Here is the current graph for the Ghana data plotting 5 elements:
image

This is closer to what I would like

image

```
Here is the current code:
# Code generated by the dialog, Display Daily Data

ghana1 <- data_book$get_data_frame(data_name="ghana1", stack_data=TRUE, measure.vars=c("rainfall","min_temperature","max_temperature","sunshine_hours","windspeed"), id.vars=c("station", "date"))
last_graph <- ggplot2::ggplot(data=ghana1, mapping=ggplot2::aes(x=date, y=value, colour=variable)) + ggplot2::geom_line() + ggplot2::geom_rug(data=ghana1 %>% filter(is.na(value)), colour="red") + theme_grey() + ggplot2::theme(axis.text.x=ggplot2::element_text(), legend.position="none") + ggplot2::facet_wrap(scales="free_y", ncol=1, facet=station ~ variable, labeller=label_wrap_gen(multi_line=FALSE),strip.position="right") + ggplot2::xlab(NULL)
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="ghana1")
data_book$get_graphs(data_name="ghana1", graph_name="last_graph")
rm(list=c("last_graph", "ghana1"))
```
1. In the code above I have added `strip.position="right"` which has made the difference.  The default is "top", I wonder if we allow any of top, bottom, left, right?
2. In the existing graph a smaller improvement would be to not wrap the facet  - but you seem to have done that already with multi.line=FALSE, so I am not quite clear why the facet headings are so big.  Perhaps there is a limit to the names.

image

3. Is there also the possibility of the angle for the facet titles?  If so, then it may be that the angle of vertical or horizontal could be combined with the strip position option.  So there could be 6 options, when strip position is left or right, then the angle could be horizontal or vertical.

4. Is there also an easy way of changing the width of the facet label? They are quite big now.

Ivanluv:
fixes #6099 


Ivanluv:
@rdstern  could you look at this

rdstern (changes requested):
The dialogue now doesn't load for me.  It says:

"System.InvalidOperationException: 'Forms that are not enabled cannot be displayed as a modal dialog box. Set the form's enabled property to true before calling showDialog.'"

Ivanluv:
@rdstern  I have corrected the error

rdstern (approved):
Great

lloyddewit (approved):

PR #6152, opened 2021-01-10
Missing Values: Fixed small bug un design

N-thony:
image

Fixed small bug in design under Climatic>Check Data>Filling missing values. When clicking on Fit Single, the controls were not showing well.
@lloyddewit this is ready to review. Thanks


lloyddewit:
@N-thony Thank you for identifying and fixing this issue.
I was able to recreate this issue on my machine.
Superficially it looks like the problem is that user controls are not positioned correctly on the dialog, and they overwrite each other.
However, when I look at the proposed code changes, I see that the functionality is changed (`ucrChkAddLegend` and `ucrChkmeanBias` are made visible at different times). Does this mean that the user now has different options than before?

Please could you explain why you changed the visibility of the check boxes, how this solves the issue, and whether the user functionality is affected? thanks

N-thony:
> @N-thony Thank you for identifying and fixing this issue.
> I was able to recreate this issue on my machine.
> Superficially it looks like the problem is that user controls are not positioned correctly on the dialog, and they overwrite each other.
> However, when I look at the proposed code changes, I see that the functionality is changed (`ucrChkAddLegend` and `ucrChkmeanBias` are made visible at different times). Does this mean that the user now has different options than before?
> 
> Please could you explain why you changed the visibility of the checkboxes, how this solves the issue, and whether the user functionality is affected? thanks

The position of ucrChkAddLegend is the one that created this problem. After looking at the code I found that ucrChkAddLegend is supposed to be visible when (rdoShow and rdoDisplay) are checked. And in case of rdoDisplay is checked it will be visible depending on the option chosen in ucrInputComboType.
So, I have just linked ucrChkAddLegend with (rdoShow and rdoDisplay) to make it simple.

rdstern (dismissed):
@N-thony this now looks fine - good to have it fixed.

N-thony:
@rdstern Sorry, there was a small change I have made. Nothing much in the code. Could you check the dialog? Thanks

rdstern (approved):

rdstern:
That's what I did - and I thought I approved!  Now I have.

lloyddewit (approved):

lloyddewit:
@N-thony Thank you for fixing this under your own initiative, and for the clear explanation about your code changes

Issue #6112, opened 2020-12-07
Adding Sadore dataset to  the R-Instat library


Wycklife:
This is a new climatic dataset from Niger. It should be added in the Instat > Browse > Climatic > Niger > Sadore >Sadore.RDS.

rdstern:
Just to mention not to have a special directory called Sadore.  It can go into the Niger directory along with other datasets from the same country.

lloyddewit:
@Wycklife This PR seems to be covered by PR #6113. Can we close this PR? thanks

Wycklife:
@lloyddewit This is an issue being fixed by PR  #6113.

lloyddewit:
@Wycklife yes, you are correct. Sorry I misread this as a PR. I deleted my comment above.

PR #6113, opened 2020-12-07
Added a new Climatic dataset (Sadore.RDS) from Niger to the R-Instat library 

Wycklife:
Fixes #6112

@africanmathsinitiative/developers This is urgent as we're required to make a video using this dataset. Could it be reviewed and merged?

PR #6159, opened 2021-01-13
Adds 2 new climatic files for Niger, fixes issue #6112

lloyddewit:
PR #6113 was created to add two new Niger climatic files to R-Instat. However, PR #6113 did not follow the file name conventions, and `instat.vbproj` had a merge conflict. This PR resolves this and replaces PR #6113.

@rdstern Please could you check and approve? thanks


rdstern (approved):

instat/instat.vbproj
```diff
@@ -5363,6 +5363,12 @@
     <Content Include="static\InstatObject\R\Backend_Components\NCMP\P7_Summary.R">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="static\Library\Climatic\Niger\sadore.RDS">
```
shadrackkibet, line 5366: Should this also include the years for consistency i.e `sadore_1983_to_2014.RDS`?

lloyddewit: @shadrackkibet Thanks, good question. I named it `sadore.RDS` because this was consistent with the names of the other RDS files I looked at for other countries. These RDS files also have a specific date range but it is not included in the file name.
I opened `sadore.RDS` and it actually contains data from 1983 to 2015. So if we did rename it we should use these years.
@rdstern Do you have an opinion? Should we leave the file name as `sadore.RDS` to be consistent with the other countries, or should we name the file `sadore_1983_to_2014.RDS` to be consistent with the `xlsx` file?

dannyparsons: I think short names are better and the years aren't essential. I suggest renaming the Excel file to `sadore` for consistency. I can do this by branching this.

Issue #6082, opened 2020-11-11
Associating rds files to R-Instat


Patowhiz:
@rdstern requests

> If some users don't have R and RStudio installed, then their rds files will also relate to R-Instat.  Like xlsx files take you into Excel, could rds take you into R-Instat?

rdstern:
@N-thony I have just discussed this issue with David - i.e. @volloholic .  He says this is a task for you and should be easy?  That would be great.  But please be ready to ask for help quickly if it isn't obvious what you should do.  I understand it should be a small task.

I found that people have been trying to do this automatically, i.e. instead of going into R-Instat and then opening the data file, they just click on the data file itself.  Of course I am not sure what would - or should - happen for those users who also have RStudio?  I guess most of them will want to keep the situation where rds files are then associated with RSudio rather thasn R-Instat.

N-thony:
@rdstern sure, I'm working on it. Thanks

lloyddewit:
This issue is partially implemented by PR #6160 .

In PR #6160 , @dannyparsons commented 'I wouldn't want this to be done on start up, as mentioned this could be done as part of the installation setup. I think that is the usual place for most software. It can then also be an option in our Options dialog for users to turn this on and off.'

Therefore to fully resolve this issue, the installer should also be updated.

lilyclements:
@ChrisMarsh82 can you update on the status of this?

PR #6160, opened 2021-01-13
Automatically open R-Instat when user double-clicks on an associated data file

N-thony:
Associated the .RDS file with R-Instat so when double click on .RDS file, R-Instat will run. This partially implements issue #6082.
@africanmathsinitiative/developers and @rdstern this is ready to review. Thanks


lloyddewit:

N-thony:
> @Ivanluv please could you review? thanks

To associate R-Instat with .RDS file we need to set this association into to the registry, the code in this branch allows R-Instat to edit your registry. So, for this, you must change the right of your user/application to edit the registry.

lloyddewit:
@N-thony Thanks for the change
@Ivanluv This looks like a working software solution. It may also be possible to solve this issue via documentation (explaining to the user how to associate RDS files with R-Instat by changing their Windows settings). What's your view?
If we do set the association in the software, should we prompt the user for confirmation first (in case they already associate RDS files with another application)?

N-thony:
> @N-thony Thanks for the change
> @Ivanluv This looks like a working software solution. It may also be possible to solve this issue via documentation (explaining to the user how to associate RDS files with R-Instat by changing their Windows settings). What's your view?
> If we do set the association in the software, should we prompt the user for confirmation first (in case they already associate RDS files with another application)?

I think the setup project should add this key and if necessary set its permissions accordingly. As the installer will be running as admin this is a good approach. To avoid users setting by themselves the permissions. This can be done during the installation of R-Instat.

rdstern:
Just to add that regular R-users will probably (automatically) associate rds with RStudio.  I would not want R-Instat to change that setting - at least it should ask.  It should not be automatic.  Can one check if there is already an association?  If not, then it could be done.  If yes, then it either leaves it, or it asks.

Ivanluv:
> > @Ivanluv please could you review? thanks
> 
> To associate R-Instat with .RDS file we need to set this association into to the registry i.e Computer\HKEY_CLASSES_ROOT, the code in this branch allows R-Instat to edit your registry. So, for this, you must change the right of your user/application to edit the registry.

The code looks okay from my end.
Also I would prefer prompting the user for confirmation first incase we  set the association in the software

N-thony:
@Ivanluv and @rdstern thanks for the comments above, I think I resolved them all, please could you review them?

dannyparsons:
I wouldn't want this to be done on start up, as mentioned this could be done as part of the installation setup. I think that is the usual place for most software. It can then also be an option in our Options dialog for users to turn this on and off.

On Roger's point, although RStudio can launch .RDS files, I've never seen this being used and only just realised it's possible, so I don't have any concern setting this for users who already have RStudio as I highly doubt anyone uses this feature.

Although this is useful, the more important task is to tell R-Instat what to do when a user double clicks on an .RDS file. Somewhere in the code will need to say what R-Instat should do when an .RDS file is clicked. I suggest it launches the Open from File dialog with the clicked file already selected. @N-thony do you understand what's needed here?

The file association is of no use if R-Instat doesn't know what to do with a .RDS that is launched, so that needs to be done first.

dannyparsons:
This essentially requires being able to load R-Instat with parameters passed to it i.e. a file. I believe there are straightforward ways to configure this in VB.NET. @N-thony if you find this is not trivial then it may not be worth spending a lot of time on it.

N-thony:
> I wouldn't want this to be done on start up, as mentioned this could be done as part of the installation setup. I think that is the usual place for most software. It can then also be an option in our Options dialog for users to turn this on and off.
> 
> On Roger's point, although RStudio can launch .RDS files, I've never seen this being used and only just realised it's possible, so I don't have any concern setting this for users who already have RStudio as I highly doubt anyone uses this feature.
> 
> Although this is useful, the more important task is to tell R-Instat what to do when a user double clicks on an .RDS file. Somewhere in the code will need to say what R-Instat should do when an .RDS file is clicked. I suggest it launches the Open from File dialog with the clicked file already selected. @N-thony do you understand what's needed here?
> 
> The file association is of no use if R-Instat doesn't know what to do with a .RDS that is launched, so that needs to be done first.

Yes, the part of the code about setting the association in the registry is not important and I can remove it. Another point is also, actually if you run this, it's doing exactly what RStudio is doing. In this branch, R-Intat knows what to do even when an .RDS file is clicked.

N-thony:
@dannyparsons I have removed a part of the code that is not important for now.
So, after running this branch, you can just double click/right-click on .RDS file, open with, then you can find your instat.exe in Debug folder and then choose it and click okay. R-Instat will run normally and the dialog for importing data will show up directly.

instat/frmMain.vb
```diff
@@ -95,9 +95,14 @@ Public Class frmMain
...
-        strStaticPath = Path.GetFullPath("static")
-        strAppDataPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "RInstat\")
 
+        'Gets the path for the executable file that started the application, not including the executable name.
+        'because the application once double click in the associate file i.e. .RDS file.
+        'We need the full path of static folder to set the working folder containg files needed when the application is loading. 
+        Dim strPath As String = Application.StartupPath
+        strStaticPath = String.Concat(strPath, "\static")
+
+        strAppDataPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "RInstat\")
```
lloyddewit, line 105: As far as I can see, these changes have no affect, `strStaticPath` will be identical. Should these changes be undone?

instat/frmMain.vb
```diff
@@ -147,6 +154,21 @@ Public Class frmMain
...
 
+    Private Sub LoadDataOnFileClicked()
+        Try
+            If (Environment.GetCommandLineArgs.Length > 1) Then
+                Dim FilePath As String = Environment.GetCommandLineArgs(1)
+                dlgImportDataset.strFileToOpenOn = FilePath
+                dlgImportDataset.bStartOpenDialog = False
+                dlgImportDataset.ShowDialog()
+            Else
+                MsgBox("This file extension is not associated with R-Instat.")
```
lloyddewit, line 165: This error will be raised when there are no command line arguments. 
If there are no command line arguments then does it always mean that `This file extension is not associated with R-Instat`?

instat/frmMain.vb
```diff
@@ -147,6 +154,21 @@ Public Class frmMain
         translateMenu(mnuBar.Items, Me)
     End Sub
 
+    Private Sub LoadDataOnFileClicked()
+        Try
+            If (Environment.GetCommandLineArgs.Length > 1) Then
+                Dim FilePath As String = Environment.GetCommandLineArgs(1)
+                dlgImportDataset.strFileToOpenOn = FilePath
```
lloyddewit, line 161: Combine into 1 line?
```suggestion
                dlgImportDataset.strFileToOpenOn = Environment.GetCommandLineArgs(1)
```

lloyddewit (changes requested):

instat/frmMain.vb
lloyddewit, line 105: As far as I can see, these changes have no affect, strStaticPath will be identical. Should these changes be undone?

lloyddewit (changes requested):

instat/frmMain.vb
N-thony: You are right, it doesn't mean if there are no command-line arguments the file extension is not associated with R-Instat. 
I may change the message.

lloyddewit:
@N-thony I followed your explanation above. Now if I double click on an RDS file, then R-Instat starts automatically with the 'Open from File' dialog open. This seems to work well.

To completely resolve issue #6082, the installer also needs to be updated. I updated the opening comment in this PR so that the issue will not be closed automatically if this PR is merged. I also added a comment to issue #6082.

This PR has merge conflicts, please could you resolve (or replace with a new PR)? thanks

instat/frmMain.vb
N-thony: I found that for the first time you open a file for example .rds file either with RStudio or R-Instat it associates the file extension you have open with by adding that automatically into the registry(FileExts). So I think the association is done by the system. Therefore, I just added some code to tell R-Instat what to do when its opens from a file.

N-thony:
@lloyddewit I Improved the code to avoid that R-Instat to open another instance when one is already running when a user runs R-Instat from a different/same file (.RDS file) and open this file. Please could you test the change? Thanks.

instat/frmMain.vb
N-thony: When double click on for example .RDS file, Path.GetFullPath will consider the location of the file and find the full path of "static" to find the necessary files needed when R-Instat is executed and consider the .RDS for example as an executable of R-Instat. This create an error since the .RDS can be anywhere in computer. So, to avoid that, once double click on .RDS file for example, Application.Startup will get the path for the executable file that started the application, not including the executable name where we can find "static" folder to allow R-Instat to find the files needed when it is executed.

instat/frmMain.vb
lloyddewit: confirmed resolved

instat/frmMain.vb
lloyddewit: Thanks for the explanation. I understand (for example) that if the user double clicks on an RDS file on the desktop that `GetFullFilePath` will look for the `static` folder on the desktop, and this will give an error.
Before you added the double-click option, this was not a problem because `GetFullPath` would always return the path of the R-Instat executable which is where `static` is located.

N-thony: Right, this change is important to allow R-Instat to run correctly when it is opened from a file.

instat/ApplicationEvents.vb
```diff
@@ -0,0 +1,19 @@
+﻿Namespace My
+    ' The following events are available for MyApplication:
+    ' Startup: Raised when the application starts, before the startup form is created.
+    ' Shutdown: Raised after all application forms are closed.  This event is not raised if the application terminates abnormally.
+    ' UnhandledException: Raised if the application encounters an unhandled exception.
+    ' StartupNextInstance: Raised when launching a single-instance application and the application is already active. 
+    ' NetworkAvailabilityChanged: Raised when the network connection is connected or disconnected.
```
lloyddewit, line 7: Replace with XML comment?
```suggestion    ''' <summary>   
    ''' The 'My' namespace and the 'My.Application' objects are part of the dotnet framework.
    ''' This class extends the 'My.Application' object by creating a Partial class named MyApplication.
    ''' This class handles the event raised when R-Instat is launched when it's already running.
    ''' For more details on extending 'My.Application', please see: 
    ''' https://docs.microsoft.com/en-us/dotnet/visual-basic/developing-apps/customizing-extending-my/extending-the-my-namespace 
    ''' </summary>
```

instat/ApplicationEvents.vb
```diff
@@ -0,0 +1,19 @@
...
+    Partial Friend Class MyApplication
+        Private Sub MyApplication_StartupNextInstance(ByVal sender As Object, ByVal e As Microsoft.VisualBasic.ApplicationServices.StartupNextInstanceEventArgs) Handles Me.StartupNextInstance
```
lloyddewit, line 9: XML comment?
```suggestion        '''----------------------------------------------------------------------------------------
        ''' <summary>   Handles the event raised when when launching a single-instance application 
        '''             and the application is already active. 
        '''             If the event was triggered by double-clicking on a data file, then the sub 
        '''             displays a message box asking if the user would like to open the file in 
        '''             the currently running R-Instat instance. </summary>
        '''
        ''' <param name="sender">   Source of the event. </param>
        ''' <param name="e">        Startup next instance event information. </param>
        '''----------------------------------------------------------------------------------------
        Private Sub MyApplication_StartupNextInstance(ByVal sender As Object, ByVal e As Microsoft.VisualBasic.ApplicationServices.StartupNextInstanceEventArgs) Handles Me.StartupNextInstance
```

instat/ApplicationEvents.vb
```diff
@@ -0,0 +1,19 @@
...
+            If e.CommandLine.Count > 0 Then
+                If MsgBox("Would you like to open this file?", MsgBoxStyle.YesNo) = MsgBoxResult.Yes Then
+                    dlgImportDataset.strFileToOpenOn = e.CommandLine(0)
+                    dlgImportDataset.bStartOpenDialog = False
+                    dlgImportDataset.ShowDialog()
+                End If
+            End If
```
lloyddewit, line 16: If the user tries to launch R-Instat for a second time by just running the executable, then nothing happens. Would it be better to display a message telling the user that only one instance of R-Instat can be executed at a time?

instat/frmMain.vb
```diff
@@ -95,9 +96,14 @@ Public Class frmMain
         clsRLink.SetLog(ucrLogWindow.txtLog)
 
         SetToDefaultLayout()
-        strStaticPath = Path.GetFullPath("static")
-        strAppDataPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "RInstat\")
 
+        'Gets the path for the executable file that started the application, not including the executable name.
+        'because the application once double click in the associate file i.e. .RDS file.
+        'We need the full path of static folder to set the working folder containg files needed when the application is loading.
```
lloyddewit, line 102:
```suggestion
        'Gets the path for the executable file that started the application, not including the 
        'executable name.
        'We use `Application.StartupPath` because this returns the correct path even if the 
        'application was started by double-clicking a data file in another folder.
        'We need the full path of the static folder to set the working folder containing files 
        'needed when the application is loading.         
```

instat/frmMain.vb
```diff
@@ -95,9 +96,14 @@ Public Class frmMain
...
 
         SetToDefaultLayout()
-        strStaticPath = Path.GetFullPath("static")
-        strAppDataPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "RInstat\")
 
+        'Gets the path for the executable file that started the application, not including the executable name.
+        'because the application once double click in the associate file i.e. .RDS file.
+        'We need the full path of static folder to set the working folder containg files needed when the application is loading. 
+        Dim strPath As String = Application.StartupPath
+        strStaticPath = String.Concat(strPath, "\static")
```
lloyddewit, line 104: Convert to single line?
```suggestion
        strStaticPath = String.Concat(Application.StartupPath, "\static")
```

lloyddewit (changes requested):

lloyddewit:
@rdstern This PR was updated since you last reviewed. Now, if the user has associated 'RDS' files with R-Instat, and the user double-clicks on an RDS file when R-Instat is already running, then a msg box is displayed that gives the user the option to load the data into the currently running R-Instat instance. The side-effect is that R-Instat is now a single-instance application (I.e. you can only run one instance at a time. Is this acceptable from your side?

N-thony:
@lloyddewit Thanks for your comment above. I think I resolved it. Please could you review?

Another thing is that this solution also take into account any data file supported by R-Instat, not just a .RDS file.

rdstern:
I am not able to test at this stage (I think), but I was concerned by "this solution also take into account any data file supported by R-Instat, not just a .RDS file."  For example, does that include xlxs files for Excel.  Usually users wouldn't want all these file extensions to go into R-Instat.  I would still like Excel files to take me to Excel, etc.  Perhaps I have misunderstood?  Is it other R files that might go to the script window?

N-thony:
> I am not able to test at this stage (I think), but I was concerned by "this solution also take into account any data file supported by R-Instat, not just a .RDS file." For example, does that include xlxs files for Excel. Usually users wouldn't want all these file extensions to go into R-Instat. I would still like Excel files to take me to Excel, etc. Perhaps I have misunderstood? Is it other R files that might go to the script window?

I mean like if you have a .xlsx file and want to open with R-Instat, you can right-click on the file>open with>then you choose R-Instat, R-Instat will open and load the file without any problem.

instat/ApplicationEvents.vb
lloyddewit: @N-thony Thanks for the change. 
If R-Instat is already running and I double-click on an RDS file, then I first get the `R-Instat is already running` message box, and then when I close that, I get the `Would you like to open this file?` message box.
Would it be better to only display the `R-Instat is already running` message when there is no data file to open?

lloyddewit (changes requested):

lloyddewit:
@rdstern Thank you for the chat this morning. I understand that you are happy that:
- the user has to explicitly associate a data type with R-Instat (via the Windows settings)
- only 1 instance of R-Instat can be run at any time
If I misunderstood, then please let me know, thanks

N-thony:
@lloyddewit Thanks for your comment above. I think I resolved it. Please could you review?

instat/ApplicationEvents.vb
```diff
@@ -17,7 +17,9 @@
         ''' <param name="e">        Startup next instance event information. </param>
         '''----------------------------------------------------------------------------------------
         Private Sub MyApplication_StartupNextInstance(ByVal sender As Object, ByVal e As Microsoft.VisualBasic.ApplicationServices.StartupNextInstanceEventArgs) Handles Me.StartupNextInstance
-            MsgBox("R-Instat is already running", MsgBoxStyle.Information, "Information")
+            If e.CommandLine.Count = 0 Then
```
lloyddewit, line 20: This works correctly. An even better solution would be just to add an `Else` to the `If` statement below.

lloyddewit (changes requested):

N-thony:
@lloyddewit Thanks for your comment above. I think I resolved it. Please could you review?

instat/ApplicationEvents.vb
```diff
@@ -0,0 +1,31 @@
...
+        '''             the currently running R-Instat instance. </summary>
+        '''
+        ''' <param name="sender">   Source of the event. </param>
+        ''' <param name="e">        Startup next instance event information. </param>
+        '''----------------------------------------------------------------------------------------
+        Private Sub MyApplication_StartupNextInstance(ByVal sender As Object, ByVal e As Microsoft.VisualBasic.ApplicationServices.StartupNextInstanceEventArgs) Handles Me.StartupNextInstance
+            If e.CommandLine.Count = 0 Then
+                MsgBox("R-Instat is already running", MsgBoxStyle.Information, "Information")
+            Else
+                If MsgBox("Would you like to open this file?", MsgBoxStyle.YesNo) = MsgBoxResult.Yes Then
```
lloyddewit, line 23: You can combine these into an `Else If` starement then you can remove one of the `End If` statements below.
@N-thony You may be getting a bit tired of my requests, I'm just trying to get you into good habits, please be patient with me :)

lloyddewit (changes requested):

N-thony:
@lloyddewit Thanks for your comment above. I think I resolved it. Please could you review?

lloyddewit (approved):

lloyddewit:
@N-thony This looks good, thank you!

@dannyparsons @shadrackkibet -  What @N-thony did in this PR is quite interesting. I learned a bit more about about dotnet when I reviewed. It looks good to me. I'd be interested if you had any comments (if you have time to look),

rdstern (approved):
@lloyddewit sorry I am still not able to run it!  So I am being trustworthy and approving to give you the flexibility to merge for the new version if you feel confident!

N-thony:
> @lloyddewit sorry I am still not able to run it! So I am being trustworthy and approving to give you the flexibility to merge for the new version if you feel confident!
@rdstern I can help you to test this PR.

instat/ApplicationEvents.vb
```diff
@@ -0,0 +1,29 @@
...
+        '''             displays a message box asking if the user would like to open the file in 
+        '''             the currently running R-Instat instance. </summary>
+        '''
+        ''' <param name="sender">   Source of the event. </param>
+        ''' <param name="e">        Startup next instance event information. </param>
+        '''----------------------------------------------------------------------------------------
+        Private Sub MyApplication_StartupNextInstance(ByVal sender As Object, ByVal e As Microsoft.VisualBasic.ApplicationServices.StartupNextInstanceEventArgs) Handles Me.StartupNextInstance
+            If e.CommandLine.Count = 0 Then
+                MsgBox("R-Instat is already running", MsgBoxStyle.Information, "Information")
+            ElseIf MsgBox("Would you like to open this file?", MsgBoxStyle.YesNo) = MsgBoxResult.Yes Then
```
shadrackkibet, line 22: Why should you ask the user whether to open the file or not only when the software is already running? I do not see the difference between opening a file at first instance and when the application is already running. This would have made sense if you are checking if the file is already opened and would like to notify the user of an overwrite (Note: R-instat doesn't overwrite an existing file it indexes the file and opens it as a separate sheet). 

I think both instances should be the same i.e it should go directly into the import dialog. 

Does this make sense to you?

shadrackkibet (commented):
This works. Only a few comments below.

I like the fact that at first instance the application loads with the import dialog on top as shown below.
image

I do not like the inconsistency when I open the file while R-Instat is already running. The message box is hidden below in the taskbar, the user might keep double-clicking on the file leading to the following.
image

There should be a way to prompt the message box on top of the screen together with the application launched. If the user double clicks on the file while the message box is open then the user is prompted with some sound?

instat/ApplicationEvents.vb
N-thony: Yeah, you are right since R-Instat doesn't overwrite an existing file. It makes sense.

N-thony:
@shadrackkibet Thanks for your comments above. I think I resolved them. Please could you review?

lloyddewit:
@shadrackkibet Thank you for the review and helpful comments.

@N-thony I retested and everything seemed to work well apart from that I was able to generate the exception below:

image

I triggered this by:

- double-clicking on an RDS file
- double-clicking on the same RDS file again when the import dataset dialog was still open.

I think I also triggered this exception through some other combinations of double-clicking so you should test thoroughly.

Another question: If I double-clicked on an RDS file inside a file explorer window then the import dataset dialog and message box were hidden under the file explorer window. Is it possible to always have the import dataset dialog and message box on top of everything else?

instat/ApplicationEvents.vb
```diff
@@ -0,0 +1,30 @@
...
+    ''' This class handles the event raised when R-Instat is launched when it's already running.
+    ''' For more details on extending 'My.Application', please see: 
+    ''' https://docs.microsoft.com/en-us/dotnet/visual-basic/developing-apps/customizing-extending-my/extending-the-my-namespace 
+    ''' </summary>
+    Partial Friend Class MyApplication
+        ''' <summary>   Handles the event raised when launching a single-instance application 
+        '''             and the application is already active. 
+        '''             If the event was triggered by double-clicking on a data file, then the sub 
+        '''             displays a message box asking if the user would like to open the file in 
+        '''             the currently running R-Instat instance. </summary>
```
lloyddewit, line 14:
```suggestion
        ''' <summary>   Handles the event raised when launching a single-instance application. 
        '''             If the event was triggered by double-clicking on a data file, then the sub 
        '''             opens R-Instat (if not already running), displays the import dataset dialog 
        '''             and populates the dialog with the double-clicked file's path. 
        '''             If the event was triggered by just starting the R-Instat executable, then 
        '''             the sub displays a message box informing the user that R-Instat is already 
        '''             running.</summary>
```

lloyddewit (changes requested):

N-thony:
> @shadrackkibet Thank you for the review and helpful comments.
> 
> @N-thony I retested and everything seemed to work well apart from that I was able to generate the exception below:
> 
> image
> 
> I triggered this by:
> 
> * double-clicking on an RDS file
> * double-clicking on the same RDS file again when the import dataset dialog was still open.
> 
> I think I also triggered this exception through some other combinations of double-clicking so you should test thoroughly.
> 
> Another question: If I double-clicked on an RDS file inside a file explorer window then the import dataset dialog and message box were hidden under the file explorer window. Is it possible to always have the import dataset dialog and message box on top of everything else?

Okay, let me see this.

shadrackkibet:
I was able to "easily" replicate the exception reported above by @lloyddewit. You should be testing this thoroughly before you ask for retesting.

I do not like the fact that I can get the import dialog without the application below it as shown below.  

image


**To replicate**

- Minimize everything from your screen
- Double click on an RDS file
- Click Ok to load the file
- Minimize R-Instat
- Now double click on and RDS file

I would like to always have the application open below the dialog.

Rstudio does this properly I like it!

N-thony:
@lloyddewit  and @shadrackkibet thanks for your comments above. I think I resolved them. Please could you review?

lloyddewit (changes requested):
@N-thony Thank you for the changes. I can no longer recreate the exception and in my tests, the software always loaded the correct file.

@shadrackkibet asked for the main window and the file load window to be displayed when the user double clicks on a data file. This did not work completely for me.

Issue 1:
- double click on RDS file, R-Instat opens, import dataset opens on top (correct)
- minimise all windows using WindowsKey/m
- double click on an RDS file, import data set opens but main R-Instat window is not shown underneath (incorrect)

I can recreate the issue above with other types of minimizing. Occasionally it works correctly.

Issue 2:
- I start R-Instat 
- I maximise the window
- I minimise to access the file explorer
- I double click on an RDS file
- The import data set dialog opens correctly but the R-Instat main window is now smaller and is no longer maximised

As @shadrackkibet mentioned, it is more efficient if you can catch these types of issues before you ask for review. Thanks

N-thony:
> @N-thony Thank you for the changes. I can no longer recreate the exception and in my tests, the software always loaded the correct file.
> 
> @shadrackkibet asked for the main window and the file load window to be displayed when the user double clicks on a data file. This did not work completely for me.
> 
> Issue 1:
> 
> * double click on RDS file, R-Instat opens, import dataset opens on top (correct)
> * minimise all windows using WindowsKey/m
> * double click on an RDS file, import data set opens but main R-Instat window is not shown underneath (incorrect)
> 
> I can recreate the issue above with other types of minimizing. Occasionally it works correctly.
> 
> Issue 2:
> 
> * I start R-Instat
> * I maximise the window
> * I minimise to access the file explorer
> * I double click on an RDS file
> * The import data set dialog opens correctly but the R-Instat main window is now smaller and is no longer maximised
> 
> As @shadrackkibet mentioned, it is more efficient if you can catch these types of issues before you ask for review. Thanks

Many thanks @lloyddewit for this value in me when testing in this kind of case with WindowsKeys/m. I can sortout this the same time when resolving the others comments.

instat/ApplicationEvents.vb
```diff
@@ -19,13 +19,21 @@
         ''' <param name="e">        Startup next instance event information. </param>
         '''----------------------------------------------------------------------------------------
         Private Sub MyApplication_StartupNextInstance(ByVal sender As Object, ByVal e As Microsoft.VisualBasic.ApplicationServices.StartupNextInstanceEventArgs) Handles Me.StartupNextInstance
-            If e.CommandLine.Count = 0 Then
-                MessageBox.Show("R-Instat is already running", "Information", MessageBoxButtons.OK, MessageBoxIcon.Information)
+            Dim strPath As String
+            If e.CommandLine.Count > 0 Then
+                frmMain.TopMost = True
+                frmMain.TopMost = False
```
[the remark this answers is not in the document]
N-thony: Yes, it is correct.

N-thony:
@lloyddewit thanks for your helpful comments above. I think I resolve them. Please could you test the change?

lloyddewit:
@N-thony 
I did the following:
- opened R-Instat (opened maximised)
- minimised everything (using windows key + m)
- opened file explorer
- double-clicked on an RDS file

I expected the main R-Instat window to open maximized but instead it opened as shown below. Please could you correct this? thanks

image

N-thony:
@lloyddewit thanks for your comment above. I think to have corrected it.

lloyddewit:
@N-thony Thank you for the change. The issue I raised above is resolved. I noticed one new issue but I don't think it's important. If you can fix it easily then please do so, otherwise don't worry.
To recreate the issue:
- double click on `instat.exe` in the `Debug` folder
- resize the R-Instat window (so it is not maximized)
- minimize everything (Windows M)
- double click on RDS file
- minimize everything (Windows M)
- double click on `instat.exe` in the `Debug` folder (on my machine, R-Instat opens with an incorrect size)

N-thony:
> @N-thony Thank you for the change. The issue I raised above is resolved. I noticed one new issue but I don't think it's important. If you can fix it easily then please do so, otherwise don't worry.
> To recreate the issue:
> 
> * double click on `instat.exe` in the `Debug` folder
> * resize the R-Instat window (so it is not maximized)
> * minimize everything (Windows M)
> * double click on RDS file
> * minimize everything (Windows M)
> * double click on `instat.exe` in the `Debug` folder (on my machine, R-Instat opens with an incorrect size)

@lloyddewit following this, if R-Instat is already running, if you double click on .exe or RDS file, it will open with the same size before it to be minimized

instat/ApplicationEvents.vb
```diff
@@ -0,0 +1,53 @@
...
+    ''' https://docs.microsoft.com/en-us/dotnet/visual-basic/developing-apps/customizing-extending-my/extending-the-my-namespace 
+    ''' </summary>
+    Partial Friend Class MyApplication
+        ''' <summary>   Handles the event raised when launching a single-instance application. 
+        '''             If the event was triggered by double-clicking on a data file, then the sub 
+        '''             opens R-Instat (if not already running), displays the import dataset dialog 
+        '''             and populates the dialog with the double-clicked file's path. 
+        '''             If the event was triggered by just starting the R-Instat executable, then 
+        '''             the sub displays a message box informing the user that R-Instat is already 
+        '''             running.</summary>
```
lloyddewit, line 16: This statement is no longer correct. Remove?

instat/ApplicationEvents.vb
```diff
@@ -0,0 +1,53 @@
...
+                End If
+            Else
+                frmMain.TopMost = True 'Needed to force the window above the other windows
+                frmMain.TopMost = False 'After the window being above other windows this will allow going to other windows
+                If frmMain.isMaximised AndAlso frmMain.isMinimised Then
+                    frmMain.WindowState = FormWindowState.Maximized
+                ElseIf Not frmMain.isMaximised AndAlso frmMain.isMinimised Then
+                    frmMain.WindowState = FormWindowState.Normal
+                End If
+            End If
```
lloyddewit, line 50: This code contains duplication. Please carefully check my suggested replacement below. The functionality should be identical to the code above. Please only accept my changes when you are confident that my suggested code below is correct, thanks
```suggestion
            frmMain.TopMost = True 'Needed to force the window above the other windows
            frmMain.TopMost = False 'After the window is forced above other windows, this will allow other windows to be on top of this window
            If frmMain.isMinimised Then
                frmMain.WindowState = If(frmMain.isMaximised, FormWindowState.Maximized, FormWindowState.Normal)
            End If

            If e.CommandLine.Count > 0 Then
                If Application.OpenForms.OfType(Of dlgImportDataset).Any Then
                    dlgImportDataset.ucrInputFilePath.SetName(Replace(e.CommandLine(0).ToString, "\", "/"))
                Else
                    dlgImportDataset.strFileToOpenOn = e.CommandLine(0)
                    dlgImportDataset.bStartOpenDialog = False
                    dlgImportDataset.ShowDialog()
                End If
            End If
```

lloyddewit (changes requested):
@N-thony Thanks, this is looking good, just a couple of small suggestions

N-thony:
@lloyddewit thanks for your comment above. I think to have resolved it.

lloyddewit (approved):

lloyddewit:
@shadrackkibet Please could you test again, and approve if you're happy? @rdstern said that he was happy for the two of us to test and approve this PR.

shadrackkibet:
I tested and it is almost working correctly. However, I got a bug when I did the following.

- Double click on an RDS file and don't click Ok
- Go back to the directory and right-click on any other file format e.g `.csv`  and open with R-Instat
- Click Ok

image

This is because of the wrong function used to load the `.csv` file. I think it also applies to other file formats.

N-thony:
@shadrackkibet thanks for your comment above. I think have resolved it.

instat/frmMain.vb
```diff
@@ -2294,8 +2319,19 @@ Public Class frmMain
     Private Sub mnuClimaticNCMPSummary_Click(sender As Object, e As EventArgs) Handles mnuClimaticNCMPSummary.Click
         dlgClimaticNCMPSummaryFile.ShowDialog()
     End Sub
+    Private Sub frmMain_Resize(ByVal sender As Object, ByVal e As System.EventArgs) Handles Me.Resize
```
shadrackkibet, line 2322:
```suggestion

    Private Sub frmMain_Resize(ByVal sender As Object, ByVal e As System.EventArgs) Handles Me.Resize
```

N-thony:
@shadrackkibet thanks for your comment above. I think have resolved it.

shadrackkibet (approved):
Works now.

lloyddewit (approved):

instat/My Project/Application.myapp
```diff
@@ -2,7 +2,7 @@
 <MyApplicationData xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
   <MySubMain>true</MySubMain>
   <MainForm>frmMain</MainForm>
-  <SingleInstance>false</SingleInstance>
+  <SingleInstance>true</SingleInstance>
```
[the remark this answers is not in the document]
N-thony: @dannyparsons this was just to not confuse the user who is already working and wants to open another dataset. Instead of running another instance and load the data with this other instance.
I don't remember the advantages of having multiple instances in this case. We can make it multiple, and it will not disturb the implementation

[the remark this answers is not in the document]
dannyparsons: Ok thanks, that makes sense. Is there a way that when launching from a file there can be a choice by the user whether to load in the same instance or a new instance? If not, it doesn't matter. Will have a think about this option with others.

Issue #6148, opened 2021-01-05
Make the structured menu optional


rdstern:
@dannyparsons had fairly complex ideas about the items in the structured menu and the menu itself.  Currently it is unlike the other special menus in that it is not optional.  I suggest it should be optional, and that should be a simple task for @Patowhiz.  It would become another item in the View menu and be before Climatic - as it is in the list of menus?

![image](https://user-images.githubusercontent.com/11226469/103649004-a42a9280-4f55-11eb-9552-c58ffecc9418.png)

It would also be in the Tools > Options menu

![image](https://user-images.githubusercontent.com/11226469/103649108-d3410400-4f55-11eb-9caa-9d749cab0bc1.png)

(If this is done, then there is another change in the climatic menu that would also affect frm main that could perhaps be done at the same time.)

PR #6161, opened 2021-01-14
Making Structured Menu Option

N-thony:
Made the structured menu option. This solves issue #6148.
@africanmathsinitiative/developers and @rdstern this is ready to review. Thanks

### DIFF
--- a/instat/clsInstatOptions.vb
+++ b/instat/clsInstatOptions.vb
@@ -36,6 +36,7 @@ Imports RDotNet
     Public bCommandsinOutput As Nullable(Of Boolean)
     Public bIncludeCommentDefault As Nullable(Of Boolean) 'sets the default for comments on the dialog
     Public bShowProcurementMenu As Nullable(Of Boolean)
+    Public bShowStructuredMenu As Nullable(Of Boolean)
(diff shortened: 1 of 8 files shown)


rdstern (approved):
@lloyddewit I am accepting this and I think it is fine.  I notice that the options in the View menu are synchronized with the Tools > Options > Tailored Menu options.  So, they work fine and they both apply, not just to the current R-Instat session, but also for when you use R-Instat again.  I initially wondered whether View would be for the current session and Tools > Options > Tailored would be forever.  On reflection I think it is ok as is, but you may have different views.

dannyparsons (approved):

lloyddewit:
@rdstern thanks for testing, I agree that we should leave the synchronisation between the view options and tool options as it is. I think it is normal in applications that there are different routes to change the same options. I think it would be confusing for the user if a different route to the same option had hidden side effects.

shadrackkibet:
@N-thony whenever you open a PR that fixes an issue always start with referencing the issue with the prefix "fixes". For example fixes #6148. This will close the original issue automatically when your PR is merged.

PR #6163, opened 2021-01-18
Updated the file data_object_R6.R

### LINKED ISSUE #6156: Small problem to fix on Climatic > Dates > Infill dates
You will see this with any example of climatic data.  It has appeared recently.  The command works, but adds an extra odd liner in the output window.  Here is an example.  I think it is the new use of the dplyr commands.  It is the summaries line below which appears every time.

```
# Code generated by the dialog, Fill date gaps
data_book$infill_missing_dates(data_name="Quel", date_name="date", factors="station")

`summarise()` ungrouping output (override with `.groups` argument)
No missing dates to infill
```

HawardKetoyoMsatsi:
Fixes #6156 


dannyparsons:
@HawardKetoyoMsatsi we need to be careful about existing functionality changing as a result of this.

`.groups="keep"` is just one option, and from the docs, this is a new option, so I don't think we always want this.

I think the sensible options would be either:

> "drop_last": dropping the last level of grouping. This was the only supported option before version 1.0.0.

or leave it as it is and,

> In addition, a message informs you of that choice, unless the option "dplyr.summarise.inform" is set to FALSE, or when summarise() is called from a function in a package.

I would suggest we set `options(dplyr.summarise.inform = FALSE)` when R-Instat loads which will use the new behaviour but won't display any messages.

We should also check that this new behaviour doesn't affect any of our existing functionality.

dannyparsons:
@HawardKetoyoMsatsi I have implemented this here #6167 so will close this.

PR #6164, opened 2021-01-18
Declustering of extremes

### LINKED ISSUE #4760: Adding declustering to the calculation of extremes

shadrackkibet:
Fixes #4760  @rdstern this is ready for testing. 

NB: I am still investigating how we can obtain declustered data and save as desired.


rdstern (changes requested):
@shadrackkibet this is looking very nice.  You have 10 as the default.  I wonder about making it higher, say 40?
More seriously, with the Ghana data, that I can attach it crashes with 10mm and an overflow exception.  I assume this is because it is processing too many observations.  It produces the summary, but gives the overflow when I stipulate the graph.

shadrackkibet:
@rdstern please check if the overflow exception problem is now fixed. I also changed the default threshold to 40 which appears to be high for temperature data therefore sometimes the user will get a bug when this threshold is higher than the data maximum.

rdstern (approved):
@shadrackkibet I am approving now, so this can be merged.  If you would like to make a small change now first then add a tooltip to the threshold value "Choose the threshold value within the data range at each station".

lloyddewit:
@Ivanluv Please could you peer review, thanks

Ivanluv:
@lloyddewit  am currently looking at this

Ivanluv:
@lloyddewit  this is okay

instat/clsRLink.vb
```diff
@@ -1267,7 +1268,7 @@ Public Class RLink
                                                               strTempError = ex.Message
                                                               bReturn = False
                                                           End Try
-                                                      End Sub)
+                                                      End Sub, maxStackSize:=25000000)
```
lloyddewit, line 1271: @shadrackkibet I was curious where the 25MB figure comes from. Is this a recommendation from somewhere or a practical figure you derived by testing on your machine? I couldn't find much information about this online.

lloyddewit (approved):
@shadrackkibet Thanks, this looks good and I'm happy to approve. I had one question (doesn't affect approval)

instat/clsRLink.vb
shadrackkibet: This purely subjective. I had to test this using various datasets in R-Instat and this limit was never reached (at least on my machine). The change in `maxStackSize` was motivated by a solution [here](https://stackoverflow.com/questions/49623768/an-unhandled-exception-of-type-system-stackoverflowexception-occurred-in-rdotn). From this link they use 2.5MB but that didn't work with `ghana` dataset so I just added a `0` and that was never reached even using other datasets. The "ideal" value I thought it will work is `integer.MaxValue` but that crashed the software. 

I thought this change will impact the software negatively. I did test other dialogs after this change but I didn't find anything suspicious.

So this value is some sort of a practical figure derived by testing on my machine.

lloyddewit: @shadrackkibet very interesting, thanks for explaining

PR #6167, opened 2021-01-19
Suppress dplyr summarise message option

### LINKED ISSUE #6156: Small problem to fix on Climatic > Dates > Infill dates

dannyparsons:
Fixes #6156 as commented here https://github.com/africanmathsinitiative/R-Instat/pull/6163#issuecomment-762843577


shadrackkibet:
@dannyparsons this works as expected. I only have one question and a suggestion.

```
options(dplyr.summarise.inform=FALSE)

# Option: Number of digits to display
options(digits=4)

# Option: Show stars on summary tables of coefficients
options(show.signif.stars=FALSE)
```

We now have these three functions in the output window why don't we have these options in one function as follows?
`options(dplyr.summarise.inform=FALSE, digits=4, show.signif.stars=FALSE)`

If this doesn't work for some reason, then I think you need to add a comment above `options(dplyr.summarise.inform=FALSE)` like others.

shadrackkibet:
In future, we might want to add more options which will mean we add many `option()` lines?

dannyparsons:
Thanks. That would be better to do in one line. Currently they are done in different ways so it's not easy to combine them (the second two are through the options dialog whereas the first is fixed). But this is a good point to consider if we can get this to be more concise by combining.

shadrackkibet:
Then my second comment should be done? i.e

```
# Option: Suppress dplyr summarise messages
options(dplyr.summarise.inform=FALSE)
``

dannyparsons:
That makes sense, but also isn't possible because the first three lines of code are coming from one string, so they can only have one comment at the top of all of them. It would be good to change this, but that would need a lot more modification of the RLink.

PR #6171, opened 2021-01-19
Adding Export to WWR in frmMain

lilyclements:


rdstern (approved):
@lilyclements another complex dialogue and looks pretty neat.

The elements are numbered from 2 to 8 and I assume that ties with them somewhere and there is a good reason why number 1 is absent?

On the shape of the dialogue I would slightly prefer the Type to be below the first receiver, so the dialogue can be narrowere - or perhaps the box with the elements exported to be wider?

Approving anyway.

dannyparsons:
Yes, that comes from the WWR specifications, but I'm not sure why it starts at 2. actually.

PR #6183, opened 2021-01-26
Display of a filter in operation and retaining of state of apply option on load - Filter dialog

### LINKED ISSUE #5901: Displaying a filter in operation
>Yes, this is more general than deleting a filter.  When a filter is in operation the row variable is in red, so you know a filter is being applied.  But which?  I thought - as a separate issue, perhaps youi could right-click in the rows (i.e. on the red bit) and there would be simply a label that Says Filter: Filter2 if a filter is in operation and perhaps Filter: None if no filter is currently applied?

_Originally posted by @rdstern in https://github.com/africanmathsinitiative/R-Instat/pull/5860#issuecomment-658311939_

@rdstern I suggest adding this as part of the label below the data view window which is easy to do.

### LINKED ISSUE #6180: Problem retaining the state of an apply option on reopen - filter dialog
This dialog doesn't retain the state of an apply option (Apply as Filter or Apply as Subset) selected on reopen. Our principle is that dialogs should always retain states on reopen.

### LINKED ISSUE #6181: Filter doesn't overwrite?
I have considered this as a small bug - but you may disagree?  

I had filter1 for a data frame.  It was incorrect.  So I chose to name the new filter as filter1 to replace it.  It doesn't allow me to do that.  I think it should.

We can delete filters now, so I assume this should be an easy change?

shadrackkibet:
Fixes #5901, Fixes #6180 and Fixes #6181. @africanmathsinitiative/developers this is ready for review. 

Notes
- Removed resetting of state for apply options on load.
- Added a method for getting the filter name.
- Setting up the display of an active filter below the current sheet.
- Clearing the filter receiver.

### DIFF
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -42,6 +42,7 @@ Public Class ucrDataView
     Private clsConvertOrderedFactor As New RFunction
     Private clsFilterApplied As New RFunction
     Private clsHideDataFrame As New RFunction
+    Private clsGetCurrentFilterName As New RFunction
     Public lstColumnNames As New List(Of KeyValuePair(Of String, String()))
 
     Private Sub ucrDataView_Load(sender As Object, e As EventArgs) Handles MyBase.Load
@@ -84,6 +85,7 @@ Public Class ucrDataView
         clsFilterApplied.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$filter_applied")
         clsHideDataFrame.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$append_to_dataframe_metadata")
         clsViewDataFrame.SetRCommand("View")
+        clsGetCurrentFilterName.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_current_filter_name")
         UpdateRFunctionDataFrameParameters()
     End Sub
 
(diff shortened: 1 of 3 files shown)


lloyddewit:

Ivanluv:
@lloyddewit this is fine

shadrackkibet:
Please note that this PR now includes a fix that ensures the preview is updated correctly when the filter is overwritten.

rdstern:
@shadrackkibet great.  I'll check

rdstern (approved):
@lloyddewit all fine for me.

instat/ucrDataView.vb
```diff
@@ -343,7 +345,8 @@ Public Class ucrDataView
             iColumnCount = frmMain.clsRLink.GetDataFrameColumnCount(grdCurrSheet.Name)
             lblRowDisplay.Text = "Showing " & grdCurrSheet.RowCount & " of " & iRowCount & " rows"
             If frmMain.clsRLink.RunInternalScriptGetValue(clsFilterApplied.ToScript()).AsLogical(0) Then
-                lblRowDisplay.Text = lblRowDisplay.Text & " (" & frmMain.clsRLink.GetDataFrameLength(grdCurrSheet.Name, False) & ")"
+                Dim strFilterName As String = frmMain.clsRLink.RunInternalScriptGetValue(clsGetCurrentFilterName.ToScript(), bSilent:=True).AsCharacter(0)
+                lblRowDisplay.Text = lblRowDisplay.Text & " (" & frmMain.clsRLink.GetDataFrameLength(grdCurrSheet.Name, False) & ")" & "|Active filter:" & strFilterName
```
lloyddewit, line 349: Use '&=' operator?
Could also make the same change to line 351 below
```suggestion
                lblRowDisplay.Text &= " (" & frmMain.clsRLink.GetDataFrameLength(grdCurrSheet.Name, False) & ")" & "|Active filter:" & strFilterName
```

lloyddewit (approved):
@shadrack this looks good, just one small comment (which I don't think isn't worth delaying the merge for)

instat/ucrDataView.vb
shadrackkibet: This a convenient short-hand. I like it. Thanks for the insight.

PR #6189, opened 2021-01-28
Disabled remove current filter on right clicks

### LINKED ISSUE #4079: Disable the Remove Current Filter?
We currently disable the Levels/Labels dialogue when the current column isn't a factor.  I like that.
Could we also disable "Remove Current Filter" when no filter is applied in a particular data frame?

If so, then remember it is on the row right-click as well.

shadrackkibet:
Fixes #4079. @africanmathsinitiative/developers this is ready for review.

Notes
- Disabled remove the current filter on column, row and cell right clicks when there is no filter and enabled otherwise


rdstern (approved):
That's logical and useful.  I am also happy that @shadrackkibet is making the software more logical in this way

lloyddewit (approved):

PR #6191, opened 2021-01-29
Remove current filter and Add buttons - Data Options

### LINKED ISSUE #6085: Removing a filter from Data options and Add button
Currently, when you got a filter say `Filter2` you can just add it into the receiver from data options subdialog and once you click return the filter is activated (red rows on the data frame) and also it remains on the receiver. Can I be allowed to do the reverse? i.e when a filter is already added into the receiver and then cleared, once return is clicked then that specific filter is then removed if active? A tricky feature request!

I also claim that **Add** button should be added here for the filters selector. It is not trivial for the user to know that you have to double click on the filter for it to be added into the receiver especially for users who are used to **Add** button from the selector.

image

shadrackkibet:
Fixes #6085. @africanmathsinitiative/developers this is ready for review.

Notes.
- Changing selector
- Showing numeric keyboard appropriately
- Added remove current filter button


rdstern (approved):
Seems to work fine

lloyddewit (approved):

PR #6203, opened 2021-02-03
Added right-click functionality to the column metadata (bottom tab right-click) and the data frame metadata (row right-click)

### LINKED ISSUE #2967: Right-click in the column metadata window
I have been pondering more on how we "make it easy" to cope with "wide" data sets.  I suggest this is important as an advance we can make, because of the neat way we will be able to handle multiple data frames.
In this context it is clear we need simple ways to select subsets of the columns.  Effectively we need to be able to filter on column names.  This could be neat if it is available through a right-click in the rows of the column metadata! We already make good use of the right-click facility and also have a way of keeping the information when we show a subset of the data.  So, for the meta-data could we have filtering on the column names/labels, etc. and also enable (and show this) through the window?
This right-click is more general than the filtering, but I see this as a key element of it.

### LINKED ISSUE #5250: Adding a right-click to the Column metadata
In the Column metadata, see below, the right-click on the left-hand side gives the same as the right-click in the data, when you currently click in the name field.
Please could we add the right-click on the bottom tab.  This could be exactly the same as the right-click on the bottom tab in the actual data.  

That's actually pretty useful, particularly as the metadata is "complete", i.e. for all columns.  And, when you want to see the data, you have that option (last one in the bottom right-click).

I hope that will be easy.  If it would also be easy then (not so important, but consistent), would be the have the same right-click on the lhs in the data frame metadata!  (In the data frame metadata one row is a data frame, and hence equivalent to the tabs in the data and column metadata.

image

shadrackkibet:
Fixes #5250 and Fixes #2967 @africanmathsinitiative/developers this is ready for review.

Notes
- Deleted extra separator, disabled freeze and unfreeze menu items
- Added and implemented facilities for tab context menu strip
- Enabled hide sheet when sheets > 1
- Removed some commented out code

### DIFF
--- a/instat/ucrDataFrameMetadata.vb
+++ b/instat/ucrDataFrameMetadata.vb
@@ -21,9 +21,15 @@ Public Class ucrDataFrameMetadata
     Public WithEvents grdCurrSheet As unvell.ReoGrid.Worksheet
     Public strPreviousCellText As String
     Private lstNonEditableColumns As New List(Of String)
+    Private clsHideDataFrame As New RFunction
+    Private clsViewDataFrame As New RFunction
+    Private clsGetDataFrame As New RFunction
 
     Private Sub frmMetaData_Load(sender As Object, e As EventArgs) Handles Me.Load
         LoadForm()
+        clsViewDataFrame.SetRCommand("View")
+        clsGetDataFrame.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_data_frame")
+        clsHideDataFrame.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$append_to_dataframe_metadata")
     End Sub
 
     ' TODO this needs tidying up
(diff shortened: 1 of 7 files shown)


shadrackkibet:
I think in the long term we need to make these right-click as user controls. Maintaining consistency across windows is difficult.

lloyddewit:

shadrackkibet:
**Note:** Right-click is now also added into row click in data frame metadata.

shadrackkibet:
@rdstern what do you think about these additions?

rdstern (dismissed):
@shadrack this looks fine to me.

shadrackkibet:
@lloyddewit this is ready for review.

instat/ucrDataFrameMetadata.vb
```diff
@@ -155,4 +161,57 @@ Public Class ucrDataFrameMetadata
     Private Sub mnuHelp_Click(sender As Object, e As EventArgs) Handles mnuHelp.Click
         Help.ShowHelp(Me, frmMain.strStaticPath & "\" & frmMain.strHelpFilePath, HelpNavigator.TopicId, "544")
     End Sub
+
+    Private Function GetSelectedVariableNamesAsArray() As String()
+        Dim lstSelectedVars As New List(Of String)
```
lloyddewit, line 166: We know the size of the array (derived from `grdMetaData.CurrentWorksheet.SelectionRange.Row`). Therefore we can make a fixed size array and allocate to it directly in the loop. Is this simpler and more efficient than creating an empty variable length list and then expanding the list on each iteration?
Declare array something like:
```
Dim arrSelectedVars(grdMetaData.CurrentWorksheet.SelectionRange.Row) As String
```

instat/ucrDataFrameMetadata.vb
```diff
@@ -155,4 +161,57 @@ Public Class ucrDataFrameMetadata
...
+
+        For i As Integer = grdMetaData.CurrentWorksheet.SelectionRange.Row To grdMetaData.CurrentWorksheet.SelectionRange.Row + grdMetaData.CurrentWorksheet.SelectionRange.Rows - 1
+            lstSelectedVars.Add(grdMetaData.CurrentWorksheet(i, 0))
```
lloyddewit, line 169: replace with array? e.g.
```
arrSelectedVars(i - grdMetaData.CurrentWorksheet.SelectionRange.Row) = grdMetaData.CurrentWorksheet(i, 0)
```

instat/ucrDataFrameMetadata.vb
```diff
@@ -155,4 +161,57 @@ Public Class ucrDataFrameMetadata
     Private Sub mnuHelp_Click(sender As Object, e As EventArgs) Handles mnuHelp.Click
         Help.ShowHelp(Me, frmMain.strStaticPath & "\" & frmMain.strHelpFilePath, HelpNavigator.TopicId, "544")
     End Sub
+
+    Private Function GetSelectedVariableNamesAsArray() As String()
```
lloyddewit, line 165: This function returns an array of strings. However all the calling functions only use the first string in the array. Would it be better just to return the first string?

lloyddewit (changes requested):
@shadrackkibet this looks good, I just had a couple of comments.

lloyddewit:
@shadrackkibet as you know, GitHub will now generate the release notes automatically. I tried to change the name of this PR so that it provides a descriptive sentence for the release notes. Please could you check/update? thanks

instat/ucrDataFrameMetadata.vb
shadrackkibet: Yes, that would work now. However, I think one reason we want to leave it as an array is that some functionalities like deleting multiple columns at the same time will require this to return an array.

instat/ucrDataFrameMetadata.vb
shadrackkibet: Agreed.

instat/ucrDataFrameMetadata.vb
shadrackkibet: Agreed.

shadrackkibet:
> @shadrackkibet as you know, GitHub will now generate the release notes automatically. I tried to change the name of this PR so that it provides a descriptive sentence for the release notes. Please could you check/update? thanks

@lloyddewit there are no changes in the Data view here. Only Column metadata (bottom right-click) and Dataframe metadata (row right-click).

instat/ucrDataFrameMetadata.vb
lloyddewit: OK, thanks
I think it's important to have an extendable architecture but I prefer only to implement functionality as/when its needed. Your preference may be different.
In this case, the implementation is done so I agree it's not worth changing (and I agree it may help another developer in the future).

lloyddewit (approved):

lloyddewit:
@rdstern There are some small implementation changes. The functionality should be identical to before. Please could you retest? thanks

rdstern (approved):
Looks fine still

PR #6210, opened 2021-02-05
Added checkmarks to four toolbar dropdown menu options ('Log Window', 'Script Window', 'Column Metadata' and 'Data Frame Metadata')

### LINKED ISSUE #6207: Tool bar items - column and data frame metadata, log and script window
The following toolbar items should be checked when the respective windows are active. This is working when you use the view menu (See below). Ensure sync between toolbar icons and view menu.

image
image
image

A question to @rdstern. When you reset the layout (using the icon or via view menu) I think the tailored menus (climatic, procurement etc) should be enabled if they are hidden as shown below.
image

Then some inconsistency, the icon tooltip says (**Reset Layout**) while using the view menu we have **Reset to Default Layout**. Shouldn't this be the same?

shadrackkibet:
Fixes #6207 @africanmathsinitiative/developers this is ready for review.

image
image


shadrackkibet:
@rdstern what do you think about these changes?

rdstern (changes requested):
This seems to work fine.  
a) the ticks in the toolbar are right at the top of the box, could they be a bit lower, as they are in the View menu.
b) Less important, but if they were a bit narrower it might also be a minor improvement.  That would leave a small space before the text.

shadrackkibet:
> This seems to work fine.
> a) the ticks in the toolbar are right at the top of the box, could they be a bit lower, as they are in the View menu.
> b) Less important, but if they were a bit narrower it might also be a minor improvement. That would leave a small space before the text.

@rdstern I didn't find an easy way to sort out the display of ticks. If this is really crucial then I can spend some more time.

shadrackkibet:
@rdstern does this look better? Seems it is not easy to reposition the ticks. I checked the view menu and the ticks appear at the top.

rdstern (approved):
@shadrackkibet the width of the box is much better now.  I can live with the tick being so high so am approving.  I like it more in the view menu, where it is sort of high, but clearly in the box.

And I do like the feature.

lloyddewit (approved):

Issue #6192, opened 2021-01-31
Oddity when importing an Excel file


rdstern:
[MPAT Kenya.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/5899471/MPAT.Kenya.xlsx)

This is a potential file for the R-Instat library.
I import just the household data sheet.
If I do it simply - just press Ok, then it is fine.
But there are -99 values in the data.  If I allow for them, then it still imports fine, but now some columns are designated as character that were numeric before.  This is particularly Qu55 to Qu58.

(To see this either show more variables (Tools > Options) or look at the column metadata.)

Can we specify a missing value code - perhaps multiple, without this happening.  

(This links to the fact that changing from character to numeric in R-Instat is still problematic - long standing limitation!)

rdstern:
@N-thony in the File > Open from File dialogue there is a field called `Missing Value String`.  If that is empty, then importing works fine.  But if I put -99 into that field, then various variables are imported as character and not numeric.

N-thony:
@rdstern I think this is not a problem with R-Instat, it is how to function work. I have tried the code as you can see in the screenshot below and I understood that when you specify a missing value code, the function change automatically the type of columns containing the missing value to a character.
![image](https://user-images.githubusercontent.com/68591383/106583423-bfd18a80-654d-11eb-9461-ac8aa6abb9e4.png)

rdstern:
That's just my problem.  Note that the variable Q55.1 doesn't have any values of -99.  If you don't have the na there in the command, then Q55.1 comes in as numeric.  Why should rio's import change it to become character?  Is there an option to prevent that?

N-thony:
> That's just my problem. Note that the variable Q55.1 doesn't have any values of -99. If you don't have the na there in the command, then Q55.1 comes in as numeric. Why should rio's import change it to become character? Is there an option to prevent that?

Yeah, you are right. If the column doesn't have the missing value code in my case Q55.1 doesn't have a value of -99, rio will still keep then NA and change the type of the column to character. I'm going to look if there is an option to prevent that

lilyclements:
The issue is that there are missing values for some variables in the data set which are given by a blank cell (e.g. for `Q55.1`, but not for `Q7`). These cells, for some reason, become characters when imported into R if the `na` argument is specify.

To fix this you need to run `na = c("-99", "")`

e.g.
```
[...]

PR #6065, opened 2020-11-05
Import Dialog Refactoring

Patowhiz:
This PR fixes no issue.
It's meant to refactor current import dialog code. Removes redundant code, improves code clarity.

@africanmathsinitiative/developers this is ready for review.

PR #6222, opened 2021-02-15
Resolved issue when importing Excel files with missing data values

N-thony:
Fixing the issue when setting the missing value code in importing excel file. This solve issue #6192
@africanmathsinitiative/developers and @rdstern this is ready for review. Thanks.

### DIFF
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -1065,12 +1065,6 @@ Public Class dlgImportDataset
     Private Function GetMissingValueRString(strText As String) As String
         Dim arrStr() As String = strText.Split(",")
 
-        'if length is = 1 return an R string else return a vector of strings for R.
-        'please note the array length will never be 0
-        'If arrStr.Length = 1 Then
-        '    Return Chr(34) & arrStr(0) & Chr(34)
-        'End If
-
         Dim strRmissingValueString As String = ""
         For Each strTemp As String In arrStr
             If strRmissingValueString = "" Then
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -1079,7 +1079,7 @@ Public Class dlgImportDataset
                 strRmissingValueString = strRmissingValueString & "," & Chr(34) & strTemp.Trim & Chr(34)
             End If
         Next
-        strRmissingValueString = "c(" & strRmissingValueString & "," & Chr(34) & "" & Chr(34) & ")"
+        strRmissingValueString = "c(" & strRmissingValueString & "," & Chr(34) & Chr(34) & ")"
         Return strRmissingValueString
     End Function
 
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -882,8 +882,8 @@ Public Class dlgImportDataset
         'so instead we are using the strFileType to identify which RFunctions should be updated accordingly
         If strFileType = "XLSX" OrElse strFileType = "XLS" Then
             'set for single imports and multiple imports 
-            clsImportExcelMulti.AddParameter("na", "c(" & GetMissingValueRString(ucrInputMissingValueStringExcel.GetText()) & "," & Chr(34) & "" & Chr(34) & ")")
-            clsImportExcel.AddParameter("na", "c(" & GetMissingValueRString(ucrInputMissingValueStringExcel.GetText()) & "," & Chr(34) & "" & Chr(34) & ")")
+            clsImportExcelMulti.AddParameter("na", GetMissingValueRString(ucrInputMissingValueStringExcel.GetText()))
+            clsImportExcel.AddParameter("na", GetMissingValueRString(ucrInputMissingValueStringExcel.GetText()))
         ElseIf strFileType = "CSV" Then
             clsImportCSV.AddParameter("na.strings", GetMissingValueRString(ucrInputMissingValueStringCSV.GetText()), iPosition:=2)
         ElseIf strFileType = "TXT" Then
@@ -1067,9 +1067,9 @@ Public Class dlgImportDataset
 
         'if length is = 1 return an R string else return a vector of strings for R.
         'please note the array length will never be 0
-        If arrStr.Length = 1 Then
-            Return Chr(34) & arrStr(0) & Chr(34)
-        End If
+        'If arrStr.Length = 1 Then
+        '    Return Chr(34) & arrStr(0) & Chr(34)
+        'End If
 
         Dim strRmissingValueString As String = ""
         For Each strTemp As String In arrStr
@@ -1079,7 +1079,7 @@ Public Class dlgImportDataset
                 strRmissingValueString = strRmissingValueString & "," & Chr(34) & strTemp.Trim & Chr(34)
             End If
         Next
-        strRmissingValueString = "c(" & strRmissingValueString & ")"
+        strRmissingValueString = "c(" & strRmissingValueString & "," & Chr(34) & "" & Chr(34) & ")"
         Return strRmissingValueString
     End Function
 
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -882,8 +882,8 @@ Public Class dlgImportDataset
         'so instead we are using the strFileType to identify which RFunctions should be updated accordingly
         If strFileType = "XLSX" OrElse strFileType = "XLS" Then
             'set for single imports and multiple imports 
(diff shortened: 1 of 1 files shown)


instat/dlgImportDataset.vb
```diff
@@ -882,8 +882,8 @@ Public Class dlgImportDataset
         'so instead we are using the strFileType to identify which RFunctions should be updated accordingly
         If strFileType = "XLSX" OrElse strFileType = "XLS" Then
             'set for single imports and multiple imports 
-            clsImportExcelMulti.AddParameter("na", GetMissingValueRString(ucrInputMissingValueStringExcel.GetText()))
-            clsImportExcel.AddParameter("na", GetMissingValueRString(ucrInputMissingValueStringExcel.GetText()))
+            clsImportExcelMulti.AddParameter("na", "c(" & GetMissingValueRString(ucrInputMissingValueStringExcel.GetText()) & "," & Chr(34) & "" & Chr(34) & ")")
+            clsImportExcel.AddParameter("na", "c(" & GetMissingValueRString(ucrInputMissingValueStringExcel.GetText()) & "," & Chr(34) & "" & Chr(34) & ")")
```
Patowhiz, line 886: @N-thony the value returned by function `GetMissingValueRString()` could make this fail. It also return an R string vector if it detects the missing values input are more than one. Please have a look at its implementation.

I looked at Lily's comment. I think it's best to implement what you have correctly done here at `GetMissingValueRString()` level.

This could be done in this PR or just edited in PR #6065 that affects the same dialog to avoid future conflicts.

N-thony:
@Patowhiz thanksfor your comment. I think to have resolve it. Could you please test?

instat/dlgImportDataset.vb
```diff
@@ -1079,7 +1079,7 @@ Public Class dlgImportDataset
                 strRmissingValueString = strRmissingValueString & "," & Chr(34) & strTemp.Trim & Chr(34)
             End If
         Next
-        strRmissingValueString = "c(" & strRmissingValueString & ")"
+        strRmissingValueString = "c(" & strRmissingValueString & "," & Chr(34) & "" & Chr(34) & ")"
```
lloyddewit, line 1082: Can we simplify?
```suggestion
        strRmissingValueString = "c(" & strRmissingValueString & "," & Chr(34) & Chr(34) & ")"
```

instat/dlgImportDataset.vb
```diff
@@ -1067,9 +1067,9 @@ Public Class dlgImportDataset
 
         'if length is = 1 return an R string else return a vector of strings for R.
         'please note the array length will never be 0
-        If arrStr.Length = 1 Then
-            Return Chr(34) & arrStr(0) & Chr(34)
-        End If
+        'If arrStr.Length = 1 Then
+        '    Return Chr(34) & arrStr(0) & Chr(34)
+        'End If
```
lloyddewit, line 1072: If these lines are no longer needed, then they should be deleted (rather than commented out)

lloyddewit (changes requested):

instat/dlgImportDataset.vb
lloyddewit: @N-thony `GetMissingValueRString` is referenced in 5 places. If you move the changes to 'GetMissingValueRString' then you will affect all 5 references (rather than just the 2 above). Please could you test all 5 references? thanks

lloyddewit (changes requested):
> @N-thony the value returned by function `GetMissingValueRString()` could make this fail. It also return an R string vector if it detects the missing values input are more than one. Please have a look at its implementation.

@N-thony I don't think you addressed @Patowhiz 's comment above. What is returned by `GetMissingValueRString("a,b")`? is this correct?

N-thony:
> > @N-thony the value returned by function `GetMissingValueRString()` could make this fail. It also return an R string vector if it detects the missing values input are more than one. Please have a look at its implementation.
> 
> @N-thony I don't think you addressed @Patowhiz 's comment above. What is returned by `GetMissingValueRString("a,b")`? is this correct?

I think in @Patowhiz's comment, he suggested me to apply the change in GetMissingValueRString() level.

N-thony:
@lloyddewit  thanks for your comment above. I think to have resolved it. Could you please test?

Patowhiz:
> > > @N-thony the value returned by function `GetMissingValueRString()` could make this fail. It also return an R string vector if it detects the missing values input are more than one. Please have a look at its implementation.
> > 
> > 
> > @N-thony I don't think you addressed @Patowhiz 's comment above. What is returned by `GetMissingValueRString("a,b")`? is this correct?
> 
> I think in @Patowhiz's comment, he suggested me to apply the change in GetMissingValueRString() level.

@N-thony yes that was my suggestion. And also as @lloyddewit mentioned above, try testing it even for other files if the new implementation will work. I think it should work since the new implementation just adds an empty string as one of the missing values placeholders which is probably always the case.

@rdstern my only worry is if there is any situation where a user may not want this default implementation `e.g na = c("-99", "")`. Because the only way they will be able to change it is by going to the script window. The added `""` is never shown in the input control.

lloyddewit:
@N-thony thank you for the short chat. As discussed, we need to ensure that your solution works with a list of missing values (as opposed to a single specified missing value).
Please create a test Excel file that contains missing values represented by "-1", "-99" and "". Then try and import the Excel file and ensure that all the missing values are replaced by "NA" and the columns remain numeric.
Thanks

N-thony:
> @N-thony thank you for the short chat. As discussed, we need to ensure that your solution works with a list of missing values (as opposed to a single specified missing value).
> Please create a test Excel file that contains missing values represented by "-1", "-99" and "". Then try and import the Excel file and ensure that all the missing values are replaced by "NA" and the columns remain numeric.
> Thanks

N-thony:
> > @N-thony thank you for the short chat. As discussed, we need to ensure that your solution works with a list of missing values (as opposed to a single specified missing value).
> > Please create a test Excel file that contains missing values represented by "-1", "-99" and "". Then try and import the Excel file and ensure that all the missing values are replaced by "NA" and the columns remain numeric.
> > Thanks

@lloyddewit  using this file [MPAT.Kenya.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/6082698/MPAT.Kenya.xlsx) shared by @rdstern, I used Household_data. I have tested this solution with column Q56 that contains "-1", "-2", and "-3" and it works well.

lloyddewit (approved):

lloyddewit:

Patowhiz:
> > > > @N-thony the value returned by function `GetMissingValueRString()` could make this fail. It also return an R string vector if it detects the missing values input are more than one. Please have a look at its implementation.
> > > 
> > > 
> > > @N-thony I don't think you addressed @Patowhiz 's comment above. What is returned by `GetMissingValueRString("a,b")`? is this correct?
> > 
> > 
> > I think in @Patowhiz's comment, he suggested me to apply the change in GetMissingValueRString() level.
> 
> @N-thony yes that was my suggestion. And also as @lloyddewit mentioned above, try testing it even for other files if the new implementation will work. I think it should work since the new implementation just adds an empty string as one of the missing values placeholders which is probably always the case.
> 
> @rdstern my only worry is if there is any situation where a user may not want this default implementation `e.g na = c("-99", "")`. Because the only way they will be able to change it is by going to the script window. The added `""` is never shown in the input control.

@rdstern could you please answer my above question.

instat/dlgImportDataset.vb
```diff
@@ -1079,7 +1073,7 @@ Public Class dlgImportDataset
                 strRmissingValueString = strRmissingValueString & "," & Chr(34) & strTemp.Trim & Chr(34)
             End If
         Next
-        strRmissingValueString = "c(" & strRmissingValueString & ")"
+        strRmissingValueString = "c(" & strRmissingValueString & "," & Chr(34) & Chr(34) & ")"
```
Patowhiz, line 1076: @maxwellfundi @lloyddewit  This now looks good from my end

rdstern:
@Patowhiz I am not quite sure what you mean by not wanting the default?  On input I think we will never want to have the variables made into character, because of those missing values.  

But I do still want to be able to include multiple missing values.  For data from Zambia now I am using -99.0,-99, because R recognises them as different.  I assume I can still have this?  If not, then perhaps we might just allow a single string/value, but have a second field for another.  It is rare that we need more than 2 - so perhaps the script window can then be used.

Patowhiz:
@N-thony my question has been answered. @lloyddewit  this is ready for you.

lloyddewit:
@rdstern Please could you test/approve? thanks

maxwellfundi:
@rdstern this is ready for your re-review.

rdstern (approved):
Great

PR #6223, opened 2021-02-15
Fixed bug in Inventory plot y-axis scales

lilyclements:
There was a bug in the inventory dialog code where 1st January would repeat itself:

image

This occurred because `common_date` would run from 1st January 2000 - 31st December 2000. When the results were plotted, R would give an upper bound of "1st January 2001" (which came up as a second 1st January).
This PR fixes this problem by restricting the y-limits to the `common_date` bounds. As a result, a graph now looks like this:

image

The only change is in the axis labels, not in the plot itself.

To replicate:
1. Import a data set (for example, the two station Ghana data)
2. Open the Inventory dialog by `Climatic > Check Data > Inventory..`
3. Input the `date` variable and `elements` (and `station` if relevant)
4. Under "Plot Type", select "Year-DOY Plot"
5. Click "Inventory Plot Options"
6. Under the "Y Axis" tab check "Specify Breaks"
7. Click OK


lilyclements:

rdstern (approved):
Great. Thanks @lilyclements This was a relatively small but annoying feature in this dialogue!

lloyddewit (approved):

PR #6225, opened 2021-02-16
Improved usability of the Survival Object dialog

### LINKED ISSUE #6165: Changes to the Define Survival Dialog
From PR # the following changes have been recommended:

> 1. The buttons at the top could have a label Type:
> 2. When I click on Modify and can give the number to modify by. I used the lung data and tried oddities that it didn't allow. It also didn't allow TRUE or FALSE, which I assume may be needed sometimes. Could there be a pull-down with 0, 1, 2, TRUE and FALSE as the suggestions. It still allows typing in case of other values.
> 3. Good to have the summary in the output window. Good to have even more, so next line is Type of Censoring: Right, Then Censored when status = FALSE

lilyclements:
Fixes #6165

Changes made:
* Added "Type" label to the top of the dialog for the buttons
* Changed an input into a combo box when a numeric variable is selected
* Added "Type of Censoring" into the summary output
* Added "Censored when status is ..." into the summary output

To test:
* Use a data set from the "Survival" package in the R library (e.g. `cancer`, `cbd`)
* Access the dialog by `Structured > Survival > Define`
* It should create a survival object which is not visible (except through the `View Objects` dialog. The output summarises the object created


rdstern (changes requested):
@lily that looks fine.  Sorry though I keep finding new questions.  This one may need input from @dannyparsons .

The dialogue seems fine.  But I then checked the metadata and nothing has changed.  As we are starting with the structured menu I wondered if there should be a change.

I checked with the one example we did before and that was to define a variable as circular.  I used the wind data from the circular package and defined the variable as circular - saving it back into the same name.

In that case I found no change in the name.  (We do change if a variable is made into a factor (say), but the metadata does change.  I wonder if something should happen for survival data?  

And what happens when you save the survival data in R-Instat and open again.  Is it still a survival object?  How would we know?

shadrackkibet:
@lilyclements any updates on this PR?

lloyddewit:
@lilyclements This PR hasn't been worked on for nearly a year.
Can we complete it or close it?
Thanks

lilyclements:
@lloyddewit I suggest we close it for now. I have the branch on my laptop still so will reopen it when the time comes, but it isn’t a priority at the moment. Thanks

PR #6226, opened 2021-02-17
Remove Library data from repo

dannyparsons:
@lloyddewit As discussed the Library directory is now removed from the project and in a separate repository https://github.com/africanmathsinitiative/R-Instat-Data

Library folder is also added to the .gitignore so that developers can maintain the Library in this directory.

I will write to developers to explain the change once merged.


.gitignore
```diff
@@ -233,4 +233,5 @@ ModelManifest.xml
 /instat/static/InstatObject/R/examples_R6.R
 /instat/static/InstatObject/R/.Rhistory
 /instat/static/InstatObject/R/examples.R
+/instat/static/Library/
```
lloyddewit, line 236: In this PR, we are instructing Git to ignore the `Library` directory and, at the same time, to update the `Library` (by deleting files).
I think this is all done in the same commit so I don't think it's a problem. We should double-check after the merge to ensure that the data files were removed as expected.

lloyddewit (approved):
Should we also cover the required changes to the Installer in this PR?
If not, should we raise an issue to update the Installer, and reference the issue in this PR?

.gitignore
dannyparsons: Agreed, it seemed to work when I tested modifying files after the commit, but should be checked.
Yes, an issue on updating the installer docs would be good too.

PR #6227, opened 2021-02-18
Enabling the test radio button in the Two variable fit model 

### LINKED ISSUE #4029: Model > Fit Model > Two Variable Test to be enabled
The Two sample option is disabled for me.  Is this general?  I wonder why?  It will be needed for Stats Made Simple.

Ivanluv:
"feature"
@africanmathsinitiative/developers  this fixes #4029
Still in progress


rdstern (changes requested):
@Ivanluv this is a good start.
a) When done, add the number of levels for categorical.  I know you have started with the numeric by numeric.
b) In the null hypothesis for the t it is limited to 0 to 1.  This should allow any numerical value.  
c) The correlation for Spearman and Kendall should change to cor.test from cor.

rdstern:
@Ivanluv I notice you changed the limits I think for the hypothesis test.  I wonder if that was the limits, where I tried again and was still told it had to be between 0 and 1.  

The example below was changed in the script window to 35 and it runs fine.  It could equally have been -35.  Of course that is for the 2-sample t-test.  In other situations it may be different.

```
# Code generated by the dialog, Two Variable Fit Model

attach(what=survey)
two_var4 <- t.test(x=yield, y=size, conf.level=0.95, mu=35, paired=FALSE)
data_book$add_model(model_name="two_var4", model=two_var4, data_name="survey")
data_book$get_models(data_name="survey", model_name="two_var4")
detach(name=survey, unload=TRUE)
rm(list=c("two_var4", "survey"))
```

lloyddewit:
@Ivanluv Is this PR ready for test/review or is it still ongoing? thanks

Ivanluv:
@lloyddewit  its ready for review by @rdstern

rdstern (changes requested):
The main item missing is the set of 5 ones listed as under numeric by categorical 2, i.e. the t-test, etc.

For now perhaps they should be included for the numeric by categorical anyway, perhaps this time under the 4 you have. 

We still have to nail the issue of the number of non-zero levels of a factor.  I'll ask Lily!

rdstern (changes requested):
@Ivanluv it is getting there.
I tried the new t-test with the usual survey data, when y is yield and x is variety - so numeric by factor.  Nasty message!  (I understand, because variety has 3 levels)  Made 2-level factor - same nasty message!
Changed the main line to a formula, i.e. two_var3 <- t.test(yield ~ variety, conf.level=0.95, mu=0, paired=FALSE)
Now it works fine for 2 levels and gives a sensible message for 3 levels.  Please also check the same change with  the other tests you added. 
In general I like the formula code when it is possible.  You might even re-instat the Model control and show that model when it is used in the code.

maxwellfundi:

Ivanluv:
@rdstern suggested that I should wait for @lilyclements  work on the Distributions be completed before I can continue working on this

maxwellfundi:
@lilyclements  your work on distributions is important for this to continue. How is that going? Any progress with the developer error yet?

lilyclements:
I'm unsure how to fix the developer error. Would you mind taking a look at it?

Ivanluv:
@lilyclements that might require discussion before making any changes

lilyclements:
The developer error was present before, and on the PR (#6276) @dannyparsons commented "These are small issues so may not be that important." so I can always write it in an issue to resolve later?

dannyparsons:
Agreed.

lloyddewit:
@Ivanluv Are you able to work on this PR again or is something still blocking? thanks

Ivanluv:
@rdstern is there anything I can add on this?

rdstern:
I was waiting till we could implement the changes made by Lily in the General MOdelling dialogue.  But I would now prefer to merge the changes you have done and then make a new issue of that point afterwards!  So I have started testing!

image


a) This is a t-test with 2 numeric variables.  It complains about the number of factor levels.  It shouldn't be checking that.  

I'll continue and add to this message.

rdstern:
Now the first variable yield and the second a factor.  I used Prepare > Factor > Recode Factor to make a 2-level factor from Variety and used that first:
a) With just 2 levels it all works.  The only change is that the default is oneway, and should be the start of the list, i.e. t.  All the elements in the list should be there.
b) When there are more than 2 levels - e.g. yield against variety it should start with aov and the block starting t should not be present.

rdstern:
Finally, both variables categorical.
Then have a line in the list of tests before McNemar.  (It is slightly different.)

rdstern:
@Ivanluv that's good progress on a complicated dialogue:
I am using the survey dataset, so will use those variables:  I think (hope) the itrems below are all simple!

1) The Test control doesn't remember the setting.  For example it returns to t each time I open the dialogue.  
These are all for` numeric by categorical 2 levels:`
2) I made trad into a factor with 2 levels.  Then wilcox of yield against trad doesn't work.  Instead whenever you can it is better to have a formula.  So put yield ~ trad instead of the x and y.  This is what you did for the t-test in that situation.  Do the same here.
3.  **_ansari_** works for numeric by 2-level factor.  But good to use the formula when you can.  So change it to ansari.test(yield ~ trad, conf.level=0.95)
4. **_mood:_**  same - use two_var18 <- mood.test(yield ~ trad, conf.level=0.95)
5. **_kruskal_**   Use formula too:  two_var21 <- kruskal.test(yield ~ trad)
6. **bartlett**  Same change
7. **_fligner_** Same change
8.  categorical 2 levels by 2 levels you still have mosaic::prop.test to add.  Call it proportion in the menu.  It could go first
For example:  two_var28 <- mosaic::prop.test(fert0 ~ trad)


9. The Try control should clear when you change the Test control.

PR #6230, opened 2021-02-22
Save Control in dlgCalculator

### LINKED ISSUE #6080: Improve the position control in various dialogues

Patowhiz:
Fixes (Partly) #6080. Solves item **(a)** 


instat/ucrCalculator.vb
```diff
@@ -2738,4 +2739,21 @@ Public Class ucrCalculator
...
+    ''' <param name="ucrChangedControl"></param>
+    Private Sub ucrSaveResultInto_SaveNameChanged(ucrChangedControl As ucrCore) Handles ucrSaveResultInto.ControlContentsChanged
+        RaiseEvent SaveNameChanged()
+    End Sub
+
+    ''' <summary>
+    ''' this will be raised the input name is changed and even when the ucrSave checkbox checked status is changed
+    ''' </summary>
+    ''' <param name="ucrChangedControl"></param>
+    Private Sub ucrSaveResultInto_SaveResultChanged(ucrChangedControl As ucrCore) Handles ucrSaveResultInto.ControlValueChanged, ucrSaveResultInto.ControlContentsChanged
```
Patowhiz, line 2755: @lloyddewit this made me wonder if we should add raising of `ControlContentsChanged` in the `ucrCheck` custom control. At the moment, it only raises `ControlValueChanged` when checked or unchecked, and the `ucrSave` control only raises its `ControlValueChanged`. 

In this case, we want the calculator to detect  any changes in its `ucrSave` control; checked status and other controls contents changed.

So I implemented it this way cause I'm not sure what it will conceptually mean to also raise contents changed when the checkbox status is changed in the `ucrSave` control and whether that will change its conceptualisation without affecting other dialogs negatively.

However, I strongly think that `ucrCheck` control should also raise `ControlContentsChanged`. If you agree then the extra event defined here may not be necessary.

rdstern (changes requested):
@Patowhiz great to have this working again:
a) It says Save Result Into.  Either delete the Into or move the control over to the right, so it doesn't overwrite.
b) I suggest the default is to produce another variable.  So the initial state is with the Save control checked.
c) When you write to the output window, just give the data.  Now you get each part of the object.
d) Not to do - just a relief.  I like the way you can use the drop-down in the save control, to save back into an existing variable.  Delighted that is now returned.

Patowhiz:
@rdstern I have now made the above changes

rdstern (approved):
Looks ok now

instat/ucrCalculator.vb
lloyddewit: @Patowhiz Thank you for the interesting question.

My personal preference would be for controls to raise standard events whenever possible. If these events have a non-standard effect then I would prefer this non-standard behaviour to be implemented in the event handler. I think doing it this way makes it easier for other developers to understand the behaviour.

Another point: Currently the `ucrSaveResultInto.ControlContentsChanged` is handled in two subs. This could be confusing for other developers reading the code. The order in which the subs will be called is also indeterminate. 
Is it possible to handle the `ucrSaveResultInto.ControlContentsChanged` event in a single sub?

Patowhiz: > My personal preference would be for controls to raise standard events whenever possible.

@lloyddewit great. Let me implement the event in the `ucrCheck`. Then we won't need `ucrSaveResultInto.ControlContentsChanged` to be handle in 2 subs

Patowhiz:
@lloyddewit I have made the necessary changes needed in the `ucrCalculator`. I wasn't sure on the best event to raise  `OnControlContentsChanged()` in `ucrCheck` (I have used `CheckedChanged`, it's already being used for  `OnControlValueChanged()`).

Patowhiz:
This is ready for review.

lloyddewit:

Ivanluv:
@lloyddewit   the code looks okay from my end

rdstern (approved):

instat/ucrCalculator.vb
```diff
@@ -2738,4 +2739,21 @@ Public Class ucrCalculator
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("hydroGOF::VE(sim = , obs = )", 10)
         End If
     End Sub
+
+    ''' <summary>
+    ''' raised when the input name is changed
+    ''' </summary>
+    ''' <param name="ucrChangedControl"></param>
+    Private Sub ucrSaveResultInto_SaveNameChanged(ucrChangedControl As ucrCore) Handles ucrSaveResultInto.ControlContentsChanged
```
lloyddewit, line 2747: The `ucrSaveResultInto.ControlContentsChanged` event is already handled below.
Also, where is the `SaveNameChanged` event handled?
Could we delete this sub (or combine it with the sub below)?

instat/ucrCalculator.vb
```diff
@@ -2738,4 +2739,21 @@ Public Class ucrCalculator
...
+        RaiseEvent SaveNameChanged()
+    End Sub
+
+    ''' <summary>
+    ''' this will be raised the input name is changed and even when the ucrSave checkbox checked status is changed
```
lloyddewit, line 2752:
```suggestion
    ''' this will be raised when the input name is changed and even when the ucrSave checkbox checked status is changed
```

lloyddewit (changes requested):

instat/ucrCalculator.vb
Patowhiz: @lloyddewit it was being handled in the `dlgCalculator`. I added the extra `SaveResultChanged` event because of how `ucrCheck` was previously implemented. I have now removed it. `SaveNameChanged` as previously implemented will be enough.

Patowhiz: This is now ready for review.

Patowhiz:
@rdstern @lloyddewit this is ready for review.

rdstern (approved):

lloyddewit (approved):

instat/ucrSave.vb
```diff
@@ -495,6 +495,7 @@ Public Class ucrSave
         ucrInputComboSave.Enabled = ucrChkSave.Checked
         ucrInputTextSave.Visible = False
         ucrInputComboSave.Visible = False
+        btnColumnPosition.Visible = ucrChkSave.Checked
```
shadrackkibet, line 498: This change has caused a bug in all dialogs that use save control. The position button is not always visible when the checkbox is checked. See all graphic dialogs for example **Describe>Specific>**
image

Patowhiz: @shadrackkibet thank you for spotitng this. I have corrected it in PR #6255

PR #6232, opened 2021-02-22
Changed save control into a combobox in the following dialogs: Numeric Cases,  Enter, Duplicate, Count in Factor, Combine Factor, Find/Replace, Transform, Split, Combine, Distance

Patowhiz:
Fixes (Partly) #6080. Solves items;


rdstern (changes requested):
@Patowhiz this largely seems to be working.  Great:
a) Could you add the facility for the default position to be after the variable that is used in the dialogue, for:
Count in factor, Transform, split and distance
b) If easy then in Combine Factors and Text > Combine, where there is a multiple receiver, the resulting default is after the last variable in the receiver?
c) As mentioned in the issue, for Duplicates, there could be some output from the summary, as there is, when using non-numeric cases.
d) I haven't checked in the instances where the position button was already working, that  the default is also just after the column, where appropriate.  Can you do this - or I can investigate if you prefer?

Patowhiz:
Item (b) can be done in a separate PR. I have created issue #6240 for it.
Item (c)  can be done in a separate PR. I have created issue #6239 for it.

Patowhiz:
@rdstern I have solved item (a) and its ready for review.

lloyddewit:
@shadrackkibet Please could you peer review? Thanks

shadrackkibet:
Only one comment. There is some unnecessary space between the label and input in the following dialog.

image

rdstern (changes requested):
@Patowhiz also on the example from @shadrackkibet for non-numeric cases, I noticed that the default is also for the new variable to go just after the existing one.  But it didn't - it still went to the end.
I have not checked other examples.

shadrackkibet:
In fact, neither of the options in the position sub dialog works.

Patowhiz:
@rdstern  `dlgFindNonnumericValues` produces the script below.
```
# Code generated by the dialog, Find Non-Numeric Values

non_numerics_calculation <- instat_calculation$new(type="calculation", function_exp="is.na(x=Month) != is.na(as.numeric(x=Month))", calculated_from=list("Dodoma"="Month"), result_name="nonum1", save=2)
data_book$run_instat_calculation(calc=non_numerics_calculation, display=FALSE)
nonum1 <- data_book$get_columns_from_data(data_name="Dodoma", col_names="nonum1")
data_book$add_columns_to_data(data_name="Dodoma", col_name="nonum1", col_data=nonum1, before=FALSE, adjacent_column="Month")
summary(object=data_book$get_columns_from_data(data_name="Dodoma", col_names="nonum1"))
rm(list=c("non_numerics_calculation", "nonum1"))
```

I forgot that the functionality I wrote in which by default always repositioned columns(even existing ones) was changed by @dannyparsons after discussions. A new parameter was introduced that by default prevents repositioning existing columns.  Check below code in `data_object_R6.R` line 562.

```
  # If replacing existing columns and not repositioning them, or before and adjacent_column column positioning parameters are missing
  # then do not reposition.
  if((replaced && keep_existing_position) || (missing(before) && missing(adjacent_column))) return()
```

Currently we don't have any implementation that sets the `keep_existing_position` parameter to false from the VB.NET front end, which means even all the other dialogs won't reposition the produced columns if a user ends up using an existing column name(which makes this technically to be an existing bug for those dialogs).

@dannyparsons @lloyddewit to make this  dialog(and other similar dialogs) reposition the produced column (which is produced inside `run_instat_calculation`) we need to implement functionality inside `ucrSave` that is used to check whether the default implementation of not repositioning existing columns should be overridden or not. I think introducing a public boolean property could be a good possible solution. What are your thoughts?

We also did have wider discussions with @dannyparsons on how the `ucrSave` should be writing these parameters though the new implementation could be done after this since it won't be "visible" to the dialogs.

lloyddewit:
@Patowhiz Thank you for the analysis. I understand that the current functionality will not reposition existing columns, even when the calling dialog explicitly wishes to reposition an existing column. Did I understand correct?
In that case, your proposal sounds OK. The flag to overwrite existing columns should ideally passed as an optional parameter. If that's not practical then the boolean could be a public propeety of `ucrSave` (as you suggested).

Maybe @dannyparsons has a different view?

Patowhiz:
@lloyddewit can we do a follow up with @dannyparsons in regards to this PR.

dannyparsons (approved):

dannyparsons:
Further work on positioning will be done on a separate PR by Patrick.

lloyddewit:
@rdstern Please see comment from @dannyparsons above. Are you willing to approve this PR now so that we can merge? thanks

dannyparsons:
I've merged so Patrick can continue on a separate PR.

PR #6235, opened 2021-02-23
Added options for non-numeric values when converting to numeric

### LINKED ISSUE #5465: Add to the numeric option to the right-click menu
Could we rename the Convert to Numeric into **Convert to Numeric - Labelled** and add a further option to **Convert to Numeric**.  That second option could do a drop_label (or equivalent) from sjlabelled first. 

Or leave the first option as it is and the second one is **Convert to Numeric - no label**. 

I keep having the problem and have to do the **Prepare > Data Frame > Convert Columns** with a particular set of options.  This wasn't something I could have covered with the participants last week.  So I prepare the data myself first to avoid the topic.

dannyparsons:
fixes #5465

Convert to numeric from the right click menu now includes a pop up box if the character/factor column contains non-numeric values to give the user the choice of "Normal Convert" or "Labelled Convert". Note that columns are now converted to numeric individually if multiple are selected.

Please suggest any changes to the form design or the text if it can any clearer.

image


shadrackkibet:
I think this should also be done in column metadata.

rdstern (changes requested):
@dannyparsons in principle great.  Four  points all tried with my usual survey data with the 36 observations.
a) Used fert and changed the 2.5 into 2..5.  It works fine with this, but your message is not quite right.  It uses the levels (not the integers)  as the categories, so 2..5 is level 4.  The actual values are 0, 0.5, etc up to 3, so the offending value has the value 4.  The others are ok.
b) Then tried with the field variable, changing the 5 to the letter S.  This didn't work for me.  My data got into a real mess still.  (Not sure quite what combination of character/factor I used, but can try again if needed.
c) I do like the idea that if nothing is numeric it gives the option to go via a factor.  If complicated to do, then perhaps that's an issue for now.
d) Once it is working is there more we may want to do with the labelled data.  Not urgent for this pull request?  Like making it simple to get rid of labels?

rdstern:
@dannyparsons I think it is close.  I am now at least getting it to go right for most things and then go wrong consistently.
I am using the field (in the survey data) with the 17th value of 5 becoming S.  If I keep doing labelled it seems ok.  The factor levels/labels are sensible.  They are in ascending order with the numbers as strings.  So 1, then 10, etc.  The S is the last and is ordinal 22.
The problem comes when I do the convert with the S becoming missing.  (And after I have done it at least once with the labelled way.)  It is still fine and numeric.  But, if I continue and make that into a factor or character again, then it starts to go wrong.  The levels are different.  And now (with the NA there) all is numeric, so those wrong factor levels become the data.

dannyparsons:
I think that's a deeper issue with how labels are used and applied, which we do need to look at again and adjust or overhaul. Are you happy the current implementation being an improvement and acceptable to merge, despite the remaining bugs?

I like the ideas for additions too. If we can remove labels manually, it can prevent a lot of the issues. Can we have these are new issues?

dannyparsons:
> I think this should also be done in column metadata right click.

@shadrackkibet Thanks very much for bringing this up. It's important these are consistent. 

When I looked into this I realised that the process for converting to numeric is repeated in 3 places: on 2 right click menus in the data grid and on 1 right click menu in the metadata grid. So I've refactored the code to remove the repetition by:

- Creating a new class `GridROperations` for the functions to run the R commands, separating them from the grid controls, and avoiding them needing to repeated in both user controls
- Created a `ConvertToNumeric()` method in `GridROperations` that can be called in each of the three places
- In the data grid, the two right click events for convert to numeric are now handled by the same sub to avoid that repetition

@lloyddewit I found a way of keeping `GridROperations` independent of a `grid`. See `RunRCommand()` in `ucrDataView` which has an `action` parameter to allow passing an anonymous function. I found the solution here https://stackoverflow.com/a/51150606/6424231. I think this avoids the issue we discussed and I'm happy with this implementation as it separates R commands from the interface. My aim would be to work towards a refactor such that the `RDotNet` dependence can be removed from the two user controls. I'd be interested in your view on the implementation and any suggestions.

Having this kind of structure is really important to ensure we have consistency in how similar actions are done and we don't introduce differences or bugs by having the same things repeated in different places. I suggest this should that a similar refactoring should be done for all R commands from the grid controls. This would be separate to this PR.

rdstern:
The minute I add the labels I am in potential trouble.
a) Make the field non-numeric by adding the S in position 17.  That's to a character variable.
b) Now make it numeric by the labels.  Looks odd.  It is now labelled.
c) So make it character again.  Correct the S back to 5.
d) Make it numeric and it seems fine.
e) But it is a "time bomb".  Make it character again and it is odd - but all numeric
f) So now when you make it numeric again it is all wrong.

It is all pretty consistent though!

image

rdstern:
@dannyparsons I looked again when it goes wrong, and I wonder about a small (partial?) fix?
First I make it go wrong with the column in survey called field as above.
a) Duplicate field into character column.  Change 5 in row 17 to S.
b) To numeric and use labelled.  Notice it is all wrong.
c) Back to factor, or character - I don't think it matters which - but I haven't checked that aspect carefully.
d) Now to numeric not with labels.  Seems ok.
e) Now look by going to factor.  Oooh.  Now look at the Levels/Labels.  Aha the labels are visible and wrong.  It is now the levels you want now and not the labels.
f) So, when returning (all numeric?) labelled data back to numeric use the levels and not the labels.

Is that a hey presto - at least for some situations?

dannyparsons:
@rdstern I am able to replicate this problem. But I don't think the fix is as simple as suggested. In this case it would solve it, but I think there are other cases where the convert back to factor from numeric should use the labels.

We do need to solve these kind of issues, but I think is a bigger job. So I would suggest this is merged as is since it solves issues in most common cases. The next useful simple thing would be a way of removing/clearing the labels. This wouldn't solve the issue directly but would help get out of some of these messes. If agreed, that can be a separate issue.

maxwellfundi:
@rdstern I think as @dannyparsons  suggests we can have this approved and merged and the problem raised to be taken up as a different issue. 
@dannyparsons the conflict here doest seem to be big, we might fix it rather than re-opening a new PR

dannyparsons:
@maxwellfundi conflicts have now been resolved. I agree with merging.

maxwellfundi:
There seems to be a big conflict now @dannyparsons .

dannyparsons:
@maxwellfundi Is there a way to exclude the vbproj file changes from the PR? That could be added in a subsequent PR.

PR #6237, opened 2021-02-23
Changed the default option position at the top on view data dialog

### LINKED ISSUE #6234: View dataframe dialog - Change the default option position at the top
The default option in this dialog is All columns. I suggest we move this option to the left of this dialog so that we have All columns then choose columns so that it is consistent with other dialogs where we use custom radio buttons at the top.

image

shadrackkibet:
fixes #6234 

- Switching top radio buttons


dannyparsons (approved):

dannyparsons:
Although this is only a very small change, since it is a change to the interface I suggest we should include it in the release notes, probably as a "bug" fix.

PR #6241, opened 2021-02-25
Added a Circular Density Plot under Structured > Circular menu

HawardKetoyoMsatsi:
@africanmathsinitiative/developers @rdstern @Wycklife @lloyddewit this is ready for review but will require the review after PR #6220 is merged. It implements ggplot options similar to the ones under Describe > Specific > Histogram.

PR #6246, opened 2021-03-03
Added a Circular Density Plot under Structured > Circular menu

HawardKetoyoMsatsi:
@africanmathsinitiative/developers @rdstern @Wycklife @lloyddewit this is ready for review.
This PR replaces PR  #6241.
@shadrackkibet you can now have a look at this.
sdgLayerOptions and sdgPlots subdialogs are reused here to implement a circular density plot.


lloyddewit:
@HawardKetoyoMsatsi I changed the title of this PR to the title used by PR #6241 . This is because the title will be in the release notes, and the users will want to know about the feature being added, rather than details of how it was implemented.

lloyddewit:

rdstern (changes requested):
@HawardKetoyoMsatsi now I begin to understand why this has taken so long.  Unless I have misunderstood, you may be on a slippery slope here.

My idea, that I thought with @Wycklife was clear.  These 3 graphs are all in the circular package.  There they are ordinary R graphs, not ggplot2 graphs. 

To ge quick we implement them just as they are, i.e. just implementing the function from the circular package.  I thought this would be quick and easy, and that's all that is needed at this stage.

Later we can attempt to change them to our own ggplot graphs.  That's not so simple, partly because of the edge effects.

Once we have the ordinary graphs , directly from the circular package, then we can also continue with the separate (general) tasks of being able to display them effectively and perhaps (using ggplotify) of turning the completed graphs into ggplot ones.

HawardKetoyoMsatsi:
I am closing this as it was done in another PR #6262

PR #6046, opened 2020-10-27
Improvements to Date Keyboard

Wycklife:
Fixes part a to f of #5999
I have done the following;
Changed the name for the date keyboard toDates/Times
Added four keys namely; asdate, datetime, time and pm
Implemented the functionality of the keys.

@africanmathsinitiative/developers This can be reviewed

PR #6257, opened 2021-03-10
Improved the date/time features of the Calculator

### LINKED ISSUE #5999: Correct and extend the calculator for dates and times
This is initially for discussion and comment by @dannyparsons or others who are clearer on the date and time functions.  This is mainly on changes to the Prepare > Calculator dialogue.

We need to do more to handle (climatic) data that is measured within a day, and the Date keyboard in the calculator may provide most of what we need.

- [x] a) I suggest we rename the keyboard as Dates/Times 
- [x] b) We already have an hour, minute and second function.  The hours works fine, but the minutes and seconds should be corrected.  lubridate::min should become minute and sec becomes second.
- [x] c) There is a function `lubridate::date`.  I am not sure of the difference between that and `lubridate::as_date`, - do we need both? If not, I suggest there be three top buttons lubridate::date or as_date (called date on the key), lubridate::as_datetime( ) (and called datetime on the key) and hms::hms( ) and called time. 
- [x] d) This is because I suggest that we may sometimes wish to separate a datetime variable into the date and the time components.  For climatic data we ask for a date variable.  I am not sure we always need a time variable, because within day information will either be a) 3hr, b) hourly, c) more detailed than hourly.  So we might have a single within-day variable, which could either be hour, or time.
- [x] e) Lubridate has a quarter function that we may as well add.
- [x] f) On within day data there is already an am function in lubridate on the existing keyboard.
- [ ] g) A shifted date function would be useful, which I think @dannyparsons uses with the era5 hourly data.  I suggest we simply describe how this is done with a tooltip and add dplyr::lag( ) also on this keyboard?
- [ ] 

Separately from this dialogue, we (probably) should recognise datetime variables in the grid.  They do register in the metadata.  Do we call them dt?  And then t for time variables.

I suggest that, for now, this will do.  Later we probably want the Make Tme and Use Time dialogues, but that would take longer.

Patowhiz:
Fixes partly #5999.
Continuation of PR #6046.


Patowhiz:
> c) On checking the functions documentation. I think as_datetime and as_date is the same. Therefore one should be removed!
> 
> Another point I noted is that the key in the last row (d_in_m) doesn't work. I think it should be days in month and use the function lubridate::days_in_month() . Also move this button so that it next to day button then wday @rdstern can confirm?

@rdstern above is @shadrackkibet [comment](https://github.com/africanmathsinitiative/R-Instat/pull/6046#issuecomment-740602042). 

Which one should I remove, as_datetime or as_date ?

rdstern:
I'd like to check this also with @dannyparsons.  We do want to advance in our treatment of climatic data measured within a day.  I am thinking that we may also want to indicate the type of data more precisely in reogrid - as well - possibly in the column metadata.

So I would like to link what we do in the keyboard, with how we will then show the data type in the grid.  So possibly either:

1) We currently have Date(D) with the D to indicate date.  Perhaps it could also indicate datetime.?  In that case we just need one key - let's keep Date then as being simpler.  Should it be D or d?
2) Or we have D for date or maybe d, and perhaps d_t for date time.  If so, then I think it will be clearer to users if we keep both keys - even though date could be used for either.
3) In either case I assume we still want a t for time!

rdstern (approved):
Great.  I was told that the work on d in the grid will be automatic and isn't part of this issue.
Eventually we need tooltips, but that is for another day!

lloyddewit (approved):

PR #6262, opened 2021-03-12
Implemented functionality of circular plots dialog

### LINKED ISSUE #6060: Adding three plots dialogs from the circular package in Structur>Circular menu

HawardKetoyoMsatsi:
Fixes #6060
@africanmathsinitiative/developers @rdstern @lloyddewit this is ready for review.
Implemented code for dlgCircularDensity plot. The dialog implements the Circular Density plot, Rose plot, and Scatter plot.


Wycklife (commented):
I have done some basic tests on this and the following are my comments:
1. Clicked on the first option(Rose Plot) and select the units combobox. You have radinas instead of radians.

WhatsApp Image 2021-03-16 at 10 50 01

2.  When I select hours as the unit with the first option, the output still has pi. See below.
WhatsApp Image 2021-03-16 at 10 53 52

HawardKetoyoMsatsi:
```
# Code generated by the dialog, Circular Plots

myws <- data_book$get_columns_from_data(data_name="notmydata", col_names="myws")
plot(circular::density.circular(x=myws, bw=25, kernel="vonmises", na.rm=TRUE), main=" ", xlab=" ", ylab=" ", shrink=1.3, units="degrees")
last_graph <- grDevices::recordPlot()
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="notmydata")
data_book$get_graphs(data_name="notmydata", graph_name="last_graph")
rm(list=c("last_graph", "myws"))
```

@rdstern @dannyparsons we have modified the R-Instat script code to enable the above code produced by the dialog to display and save the graph. And our investigations show that the function recordPlot can also work for ggplots.

We now need to discuss with @dannyparsons on how this can be generally supported for other dialogs that use non ggplot package.

rdstern:
@HawardKetoyoMsatsi this is looking good.  There is a lot that can be added - with many options available, but I would like to merge this quickly and then make a separate issue for refinements.
a) There are the changes noted by @Wycklife above
b) You have Kernel visible for each type of plot, but I think it only applies to the density.
c) Keeping the overall size the same could you have `Circular: ` in front of the 3 buttons at the top.  Then the first one says Histogram (Rose Plot), the second says Density Plot and the third says Scatter Plot as now.

dannyparsons:
I think it's ok for enhancements to be left as a separate PR later, but bugs or errors should be addressed before merging.
e.g. I think these should be fixed before merging:
> You have radinas instead of radians.

> b) You have Kernel visible for each type of plot, but I think it only applies to the density.

HawardKetoyoMsatsi:
@Wycklife @rdstern The above issues have been fixed, kindly have a look.

@Wycklife When one selects hours as the unit with the first option, the output still has pi since conversion is done from Hour Angles to degrees which appear in terms of Pi, so it is just a conversion from one scale to another but they don't calibrate in the hour-scale.

HawardKetoyoMsatsi:
@dannyparsons @rdstern @Patowhiz suggestions that arose from the discussion among @Patowhiz and @dannyparsons and I have been implemented herein. kindly have a close look at it.

maxwellfundi:
@Wycklife  please re-review this PR

Wycklife:
> @Wycklife please re-review this PR

Okay

Wycklife:
I have tested this and it seems to work well from my end.

maxwellfundi:
@rdstern This now is ready for your review.

Wycklife (approved):

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1657,7 +1657,7 @@ DataSheet$set("public", "add_object", function(object, object_name) {
   if(object_name %in% names(private$objects)) message("An object called ", object_name, " already exists. It will be replaced.")
   private$objects[[object_name]] <- object
   self$append_to_changes(list(Added_object, object_name))
-  if(any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "openair") %in% class(object))) {
+  if(any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "openair","recordedplot") %in% class(object))) {
```
dannyparsons, line 1660:
```suggestion
  if(any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "openair", "recordedplot") %in% class(object))) {

```

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1681,7 +1681,7 @@ DataSheet$set("public", "get_object_names", function(type = "", as_list = FALSE,
   if(type == "") out = names(private$objects)
   else {
     if(type == model_label) out = names(private$objects)[!sapply(private$objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "htmlTable", "Surv") %in% class(x)))]
-    else if(type == graph_label) out = names(private$objects)[sapply(private$objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "openair") %in% class(x)))]
+    else if(type == graph_label) out = names(private$objects)[sapply(private$objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "openair","recordedplot") %in% class(x)))]
```
dannyparsons, line 1684:
```suggestion
    else if(type == graph_label) out = names(private$objects)[sapply(private$objects, function(x) any(c("ggplot", "gg", "gtable", "grob", "ggmultiplot", "ggsurv", "ggsurvplot", "openair", "recordedplot") %in% class(x)))]

```

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -2278,4 +2278,10 @@ plot_declustered <- function(data, station_col_name, element_col_name, threshold
       obj %>% ggplot2::ggplot(xlab = xlab, ylab = ylab)
     }
   }
+}
+
+#This function creates a wrapper around grDevices::recordPlot() to enable non-ggplot graphs to be saved as recorded_plot objects.
+record_graph <- function(x) {
+x
+return(grDevices::recordPlot())
```
dannyparsons, line 2286: Please use the correct indentation in a function. If you edit in RStudio this should be done automatically.

dannyparsons (changes requested):
- Some small formatting suggestions
- I tested with the three different graph display options: output window, R viewer and separate viewer and all worked
- However, when I click display last graph on the toolbox the  window pops up but it's blank. Can you check this?

rdstern (changes requested):
This is looking good now.
a) The kernel is no longer in the dialogue for non-density graphs, but it is still in the R code you generate.  It doesn't cause trouble, but should only be in the command for a density plot.
b) You have shrink=2 as your default.  The default in the package is 1, so I suggest you change it to 1.  
c) If you wish you could start the process of adding arguments to the dialogue by including Shrink as a label and then an up-down.  It could go from perhaps 0.5 to 3 in steps of 0.1.  It applies to all 3 buttons.

HawardKetoyoMsatsi:
@dannyparsons @rdstern the above issues have been fixed, kindly take a look.

instat/dlgCircularDensityPlot.vb
```diff
@@ -189,6 +197,10 @@ Public Class dlgCircularDensityPlot
         TestOkEnabled()
     End Sub
 
+    Private Sub CoreControls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrSaveDensity.ControlContentsChanged, ucrReceiverVariable.ControlContentsChanged
+
+    End Sub
```
dannyparsons, line 202: Empty sub?

instat/dlgCircularDensityPlot.vb
```diff
@@ -1,3 +1,220 @@
...
+        clsRosePlotFunction.AddParameter("bins", 36, bIncludeArgumentName:=True, iPosition:=5)
+        clsRosePlotFunction.AddParameter("col", Chr(34) & "blue" & Chr(34), bIncludeArgumentName:=True, iPosition:=6)
+        clsRosePlotFunction.AddParameter("units", Chr(34) & "degrees" & Chr(34), iPosition:=8)
+        clsRosePlotFunction.AddParameter("shrink", 1, bIncludeArgumentName:=True, iPosition:=9)
+
+        clsScatterPlotFunction.SetPackageName("circular")
+        clsScatterPlotFunction.SetRCommand("plot.circular")
+        clsScatterPlotFunction.AddParameter("na.rm", "TRUE", iPosition:=3)
+        clsScatterPlotFunction.AddParameter("stack", "TRUE", iPosition:=4)
+        clsScatterPlotFunction.AddParameter("sep", 0.0003, bIncludeArgumentName:=True, iPosition:=5)
```
dannyparsons, line 150: bIncludeArgumentName:=True is the default so you don't need to include that. You can remove this parameter and it will function the same.

dannyparsons (changes requested):
R code formatting looks good now. Couple of small comments below.

I checked the View Last Graph button again and it still shows a blank window. Did you check this?

HawardKetoyoMsatsi:
> * Some small formatting suggestions
> * I tested with the three different graph display options: output window, R viewer and separate viewer and all worked
> * However, when I click display last graph on the toolbox the  window pops up but it's blank. Can you check this?

Hi @dannyparsons,  @Patowhiz is investigating your third issue raised herein.

shadrackkibet (approved):
This is already working therefore we can have it as an initial implementation. I think we can have the issue with the last graph investigated as a separate issue. I therefore approve. Over to you @lloyddewit , @dannyparsons  and @rdstern

dannyparsons:
I'd like to understand the last graph issue before merging, as this may affect the dialog implementation. It may also have effects on other dialogs.

rdstern (dismissed):
@haward great this seems to work now.  I just notice that the plot.circular code still includes the kernel ="von.mises".  which is not needed there.  My understanding is that it is only needed for the density.  I am happy you have taken it out of rose plot.
I am approving as this isn't serious - it is ignored by the function.  But nice (assuming it is very quick) to make this small change.

lloyddewit (dismissed):

lloyddewit:
> 
> 
> @haward great this seems to work now. I just notice that the plot.circular code still includes the kernel ="von.mises". which is not needed there. My understanding is that it is only needed for the density. I am happy you have taken it out of rose plot.
> I am approving as this isn't serious - it is ignored by the function. But nice (assuming it is very quick) to make this small change.

@HawardKetoyoMsatsi Would you like to make the optional change suggested by @rdstern or can we go ahead and merge? 
@dannyparsons You asked to review before this PR was merged. You already reviewed above. Do you wish to do a final review?

Thanks

HawardKetoyoMsatsi:
> > @haward great this seems to work now. I just notice that the plot.circular code still includes the kernel ="von.mises". which is not needed there. My understanding is that it is only needed for the density. I am happy you have taken it out of rose plot.
> > I am approving as this isn't serious - it is ignored by the function. But nice (assuming it is very quick) to make this small change.
> 
> @HawardKetoyoMsatsi Would you like to make the optional change suggested by @rdstern or can we go ahead and merge?
> @dannyparsons You asked to review before this PR was merged. You already reviewed above. Do you wish to do a final review?
> 
> Thanks

@lloyddewit Yes please let let me make the requested changes before you merge.

HawardKetoyoMsatsi:
@lloyddewit I have removed the kernel ="von.mises" parameter as requested and it is okay now, back to you and @dannyparsons .

lloyddewit (approved):

lloyddewit:
@HawardKetoyoMsatsi Thank you

@rdstern @dannyparsons If you are both happy, then I'll merge, thanks

rdstern (approved):

dannyparsons (changes requested):
The issue with view last graph is potentially general and relates to the changes to the Data Book in this PR. So I would like to understand this better before merging.

@Patowhiz I think you're probably busy with other urgent tasks. @lilyclements Could you take a quick look at the displaying of last graph. I guess something might need to be updated in the view last graph R method to account for record_plot. If not I could have a look later.

dannyparsons:
@HawardKetoyoMsatsi I found the issue which was the recordPlot not saving objects correctly. To save time I opened a new PR with this small change #6346

Issue #6201, opened 2021-02-03
Correct Describe > One Variable > Frequencies 


rdstern:
@HawardKetoyoMsatsi I hope this particular correction is small.  The R-code seems over complicated, by doing wverything in one line and could be simplified.  You may be able to sort this all out on your own, but @shadrackkibet could help if needed.

Currently (for me) this doesn't work for the graph option or for both.  Whether there is a single variable or multiple, it gives an error.  This is one Variable which doesn't work.  The same for 2 variables does work, so that should help.  I haven't checked for 3 variables and perhaps you could?

As a topic I hope you are ok that this becomes the "tip of an iceberg" as follows:  I would like the Describe > One Variable, Two Variable and Three Variable to become useful at the start of most analyses.  

a) There are other improvements for one variable.  As we discuss they should become separate issues and be resolved.  6177 is an example.  I would also like the One Variable graph to offer bar charts for numeric variables as an option.  I think One Variable > Summarise is currently ok.
b) I hope the work on one variable will be quick, because there is more work needed on the 2 variables.  This includes tables stuff, which probably needs work by @lilyclements first, so move to 2-variable graphs.
c) Then 3 variables, both the Graph and Tables dialogues have to be constructed, or activated.  That's more.

I describe testing datasets in #6200 .  These dialogues should be useful for small datasets including one from the tomato game, and the survey that is in the library.  Also for larger ones like the climatic in #6200 and the survey in #6199

PR #6270, opened 2021-03-15
In the 'One Way Frequency' dialog box, allow multiple/single variables to be specified for 'Graph' or 'Both'

lilyclements:

Changes made:
* Changed `sjplot` to `plot_frq` since the plot wasn't working otherwise
* Added the `list` function depending if one variable is being plotted, or several variables. This fixes a bug (not outlined in any issue) where there was an error if only one variable was in the receiver for "Graph" or "Both".
* Minor design tidy up

Fixes some of #6201 
Issue #6201 has three other requests that I haven't looked at in this PR - I'm unsure if these three other fixes are in their own separate issues. If they are, then #6201 can be closed when this is merged.


rdstern (changes requested):
@lilyclements great.  This now works.  
There is just one item you may quickly be able to examine.  It currently calls plot_frq inside plot_grid.  This currently means it is harder to look at, in the R viewer.  We expect to improve that, but (only if very quick) do we need the plot_grid is only a single variable is plotted?  Could we then just call plot_frq directly?  That would make it a ggplot graph, rather than a grid.

Then I would like to see this merged - because it does now work.  There is a lot more work that is possible, and I suggest worthwhile, to add more options - and also to stop messy graphs.  here is an example:
image

But that should be investigated initially by others, not by Lily

If the change suggested above takes more than a few minutes, then leave it for later work and we'll merge this.

lilyclements:
@rdstern It's definitely possible, but I'm not sure how to do it because there are multiple functions. It would be quicker for someone else to do (either now or I can update the issue to add it in)

rdstern (approved):
Bug fixed so happy to approve.  There is lots more that can be done later, but useful to have this working nicely as the starting point.

I suggest this can close issue 6201.  I will open a new issue on possible further improvements.

lloyddewit (approved):

instat/dlgOneWayFrequencies.vb
```diff
@@ -119,15 +119,17 @@ Public Class dlgOneWayFrequencies
 
         clsPlotGrid.SetPackageName("sjPlot")
         clsPlotGrid.SetRCommand("plot_grid")
+        clsPlotGrid.AddParameter("x", clsRFunctionParameter:=clsSjPlotList)
```
shadrackkibet, line 122: Position? Also in other places?

shadrackkibet:
Also to note that **View last graph** doesn't work for this dialog. It is a good example where `ggplotify::as.ggplot()` will work. I do not think the graph will loose its features due to conversion.

instat/dlgOneWayFrequencies.vb
lilyclements: Thanks for this. I think someone may be working on this dialog shortly (see issue #6279) so I've added a comment requesting the parameter positions added then (just to avoid conflicts in case I add it in and it's not merged quickly!)

rdstern:
@HawardKetoyoMsatsi is going to look at #6279, but hasn't started yet.

There are 2 points to fix in this issue, and then I am hoping that @jkmusyoka can suggest further additions to the dialogue.  As @shadrackkibet is clearly back, then I would be delighted if he could take over Haward's role.  It is his sort of job, and I think quick for him, while tricky for Haward?

instat/dlgOneWayFrequencies.vb
shadrackkibet: That is fine @lilyclements

PR #6272, opened 2021-03-15
Default (column) names now more concise by not including unnecessary number suffix

### LINKED ISSUE #6271: Default names should start without a number
Default column/data frame/object names currently go in sequence e.g. `date1, date2, date3, ...` until an available name is reached.

I suggestion optionally allowing the sequence to start without a number e.g. `date, date1, date2, date3, ...`

This option can be the default. It will ensure more concise names and avoiding numbers unnecessarily.

dannyparsons:
fixes #6271 

For simplicity, I have just changed the default for `include_index` and not exposed the `include_index` parameter in `ucrSave` to allow it to be changed on individual dialogs. I'm not sure this flexibility is needed, so I suggest we only expose the parameter when we have a case of a dialog needing it.


lloyddewit (approved):

lloyddewit:
@shadrackkibet I don't think that @rdstern can usefully test this. Please could you do the 2nd review/approval? 
@lilyclements I think the above 2 reviews should be enough. So we'll go ahead and approve without waiting for your review, I hope that's OK.
Thanks

maxwellfundi (approved):
Works as needed.

PR #6276, opened 2021-03-15
Allowing more distributions for certain variable classes when modelling (the general case)

### LINKED ISSUE #6197: Correcting and improving the general and other modelling dialogues
There are real problems with this facility.  As a simple test I tried to do a simple logistic model.  I was using these data:

[class.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/5903667/class.xlsx)

The description of the analysis is [here](https://stats.idre.ucla.edu/r/dae/logit-regression/).

I quickly got this error and am stuck and having to restart.
image

But I suggest fixing the error is secondary to a bit of an overhaul of the way we are doing the fitting as I explain below.  This will be instructional - and challenging - for @shadrackkibet and I hope @lilyclements might also usefully get involved.

We were working towards a system that was very clever, but possibly too clever.  I therefore suggest we take a step back on cleverness and get a good working system first.  Once it is working well, then can add further aspects later.

The dependent variable starts it off.  We currently try to do different things if it is integer, etc and we don't consider integer-type data elsewhere in R-Instat.  What could be useful could be to note the type of data for the y-variable and report that.  @dannyparsons does that nicely in the Describe > Two Variables > Summarise and Graph.  

It is the same problem in Model > Fit > 2 Variable, 3 Variable and General.

The important distinctions are that the Y can be numeric or categorical or ordered categorical (Or perhaps even logical for binary data).  They are easy to report on.  If we take binary data from the example it is defined as integer, so I suggest it is reported as such, or maybe as numeric.  We could make it numeric!

So I suggest we have all the options for glm for any type of data.  That means we have to be sensible ourselves (at least initially), because R-Instat isn't helping us.  So, the pull-down has Normal first - and then uses lm automatically, then a line and then all the options for glm below.  (I am not sure where the glm links go currently because the Model Options is disabled.  Perhaps they should go there, because there is a default for each distribution.  

Then, we try later to build some checks.  For example Poisson models should have non-negative data, but do they have to be integer?  And, if so, then what about quasi-Poisson?

Then there is non-linear models, ordered categorical data models and multinomial models and hierarchical models, but let's add them later.  We can use them (except nnet) now via the model-fitting keyboards.

lilyclements:
Fixes #6197 

Changes
* Added a label in the general modelling dialog which states the variable type (or if the variable type is unsuitable for a model)
* Allowed numeric-type distributions to work for positive integers
* Added a new type for distributions - "two way numeric" - to account for when data is binary (note that for `binomial` or `quasibinomial`, a numeric is accepted only if it is binary, that is, it is 0/1)


instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1406,14 +1406,20 @@ DataSheet$set("public", "get_data_type", function(col_name = "") {
   else if(is.numeric(private$data[[col_name]])) {
     #TODO vectors with integer values but stored as numeric will return numeric.
     #     Is that desirable?
-    if(is.integer(private$data[[col_name]])) {
-      if(all(private$data[[col_name]]>0)) {
-        type = "positive integer"
+      if(is.binary(private$data[[col_name]])){
+        type = "two level numeric"
```
lilyclements, line 1410: I have kept "two level numeric" and "two level factor" separate for now. I can combine them together, however.
One reason I kept them separate is because I noticed in "Two Variable Fit Model" I think `ucrReceiverExplanatory.strCurrDataType = "two variable numeric"` would need to be added into l. 477 (`DataTypeAccepted` sub). However, do not think "two variable factor" is needed there.

instat/static/.old/data_object.R
```diff
@@ -683,7 +683,7 @@ data_obj$methods(get_data_type = function(col_name = "") {
   }
   
   if(is.factor(data[[col_name]]) & (length(levels(data[[col_name]]))==2)){
-    type = "2 level factor"
+    type = "two level factor"
```
lilyclements, line 686: I just noticed this is the only place we use "2 level factor" not "two level factor" so changed it in case of future bugs.
Should I add in the "two level numeric" type in this R file as well?

lilyclements (commented):
@dannyparsons I have made the changes we discussed earlier to the distributions. I've made a few comments below - would you mind taking a look?

instat/static/InstatObject/R/data_object_R6.R
dannyparsons: Good idea, that sounds very useful to keep them separate!

dannyparsons: Can you also change `private$data[[col_name]]` to `curr_col` and then on 1394 change `use_current_filter = TRUE` so that  it's checking on the filtered data.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1406,14 +1406,20 @@ DataSheet$set("public", "get_data_type", function(col_name = "") {
...
+        }
+        else type = "integer"
+      }
+      else type = "numeric"
   }
+
+
+
   else if(is.factor(curr_col)) {
     if(length(levels(curr_col))==2) type = "two level factor"
```
dannyparsons, line 1424: Can you change this such that it checks the number of non-empty levels? An example is a 3-level factor where we filter down to 2 levels in R-Instat. We would like this to return as `"two level factor"` because when you run the dialog there will be only 2 levels due to the filter.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1406,14 +1406,20 @@ DataSheet$set("public", "get_data_type", function(col_name = "") {
...
-    if(is.integer(private$data[[col_name]])) {
-      if(all(private$data[[col_name]]>0)) {
-        type = "positive integer"
+      if(is.binary(private$data[[col_name]])){
+        type = "two level numeric"
       }
-      else type = "integer"
-    }
-    else type = "numeric"
+      else if(is.integer(private$data[[col_name]])) {
```
dannyparsons, line 1412: Here we could do `all(a == as.integer(a))` so that it also captures numeric columns which only have integers (as opposed only to integer type columns).

instat/static/InstatObject/R/data_object_R6.R
lilyclements: Yes will do. The checks on numerics/integers/positive integers also use `col_name`. Should I change these to `curr_col` too?

dannyparsons: Yes, all the checks should be on `curr_col`

lilyclements:
@dannyparsons this is ready for re-review

dannyparsons (changes requested):
A few issues found:

- When I opened the dialog for the first time the checkbox for convert to numeric was visible but the receive is empty, so I think just needs to be checked on first opening
- When I changed response variable from numeric to factor it doesn't update the response variable type. Could the R function return "factor" so it can be displayed? The same happened for a logical column and also a 0-1 numeric column. Maybe the label is not updating correctly?
- Also it didn't let me choose a distribution when the response variable was logical
- When I had stuff filled in and clicked reset I got this message: "Developer error: Cannot set value of ucrFamily because cannot find a distribution matching the function and parameter given"

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1403,19 +1403,22 @@ DataSheet$set("public", "get_data_type", function(col_name = "") {
...
+        if(all(curr_col>0)) {
+          type = "positive integer"
+        }
+        else type = "integer"
+      }
+      else type = "numeric"
   }
   else if(is.factor(curr_col)) {
-    if(length(levels(curr_col))==2) type = "two level factor"
+    if(nlevels(curr_col)==2 | nlevels(factor(curr_col))==2) type = "two level factor"
```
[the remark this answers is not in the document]
lilyclements: Thanks! When should a `||` be used instead of a `|`?

lilyclements:
@dannyparsons thank you for the review - I've addressed the changes, can you have another look?

instat/static/InstatObject/R/data_object_R6.R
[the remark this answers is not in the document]
dannyparsons: Good question. `||` always returns a single logical value whereas `|` is vectorised. It doesn't matter in this case, because both give the same but `If` gives a warning if more than one logical value is provided, so this forces it as one.

Try this to see the difference:
```r
1:10 > 5 | (1:10 %% 2) == 0
1:10 > 5 || (1:10 %% 2) == 0
```
Here is quite a nice explanation https://stackoverflow.com/a/22251971/6424231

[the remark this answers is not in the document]
lilyclements: Ah interesting! Thank you!

instat/dlgFitModel.vb
```diff
@@ -42,6 +42,7 @@ Public Class dlgFitModel
         If bFirstLoad Then
             InitialiseDialog()
             bFirstLoad = False
+            ucrChkConvertToVariate.Visible = False
```
dannyparsons, line 45: Instead of this, can you run `ResponseConvert()` as the last line of `SetRCodeForControls()`? That way the settings for the checkbox is done consistently in a single place.

dannyparsons (commented):
Looks good. From my last list of issues, only the developer error remains. I identified when it happens. It seems to only be when the response variable is an "unsuitable type". Then click Reset then you get the dev error. I think this is coming from `ucrDistribution`.

Also, when the response variable is character, convert to numeric checkbox makes everything `NA` so the model gives an error. I think we only want convert to numeric for factor? Check with @rdstern 

These are small issues so may not be that important.

lilyclements:
Thanks for identifying this - I'm unsure how to fix it. Would you mind taking a quick look?

I've sorted the issue with the visibility of the ucrChk

Ivanluv:
@lilyclements  everything including the code looks good by me .For the issue on the developer error by the distribution control ,I think that requires more work within the distribution control

lilyclements:

dannyparsons (approved):
Since the developer error was an existing issue, and not a major one, I suggest this is approved due to other more urgent tasks.

dannyparsons:
@rdstern Could you also test?

@lilyclements Could you open a separate issue for the developer error?

lilyclements:
Have now reported the error in #6345 - thanks.

rdstern (approved):
@lilyclements great.

I would like a discussion now on what you consider to be the next steps in this "system".  For example I notice it correctly detected an integer variable in the survey data and suggested a Poisson model.  When I had a similar variable but with zeros, it said it wasn't a category.

However this is a big step forwards.  Next we avoid the developer bug by always having some possible distributions!

lloyddewit:
@lilyclements Thank you for this PR. Please could you check that the new title for this PR is correct (and change if not)? thanks

PR #6281, opened 2021-03-16
Separating the sub-dialog for Merge into two sub-dialogs

### LINKED ISSUE #6265: Simplify the use of the merge dialogue
This dialogue works well.

The one complication is that when you click on Merge Options, the resulting sub-dialogue looks more daunting than need be.  I suggest the Merge Options button be renamed as `Joining Columns`.  There is a second button on the right - or below the Joining Columns button.  It says `Columns to Include`.  

This stops the need for the sub-dialogue to be tabbed.  

Further, I suggest on the columns to merge by that it always goes to the second button at the top if there are no joining columns, i.e. if the main dailogue was red rather than grey.

lilyclements:
Replaces #6277
Fixes #6265

This PR separates the two tabs in the original sub-dialog into two separate sub dialogs (as described in #6265).

@dannyparsons this is ready for review. From what I can tell, the dialogs still automatically add in columns you merge by, even if you try to remove those columns in the "Columns to Include" tab. Is this okay?

### DIFF
--- a/instat/dlgMerge.vb
+++ b/instat/dlgMerge.vb
@@ -50,7 +50,6 @@ Public Class dlgMerge
 
     Private Sub InitialiseDialog()
         Dim dctJoinTypes As New Dictionary(Of String, String)
-
         ucrBase.iHelpTopicID = 60
 
         'sdgMerge.SetRSyntax(ucrBase.clsRsyntax)
@@ -103,14 +102,11 @@ Public Class dlgMerge
 
         ucrFirstDataFrame.Reset()
         ucrSecondDataFrame.Reset()
-
         ucrSaveMerge.Reset()
 
         clsMerge.SetPackageName("dplyr")
         clsMerge.SetRCommand("full_join")
-
         clsByList.SetRCommand("c")
-
         clsSuffixC.SetRCommand("c")
 
         ucrBase.clsRsyntax.SetBaseRFunction(clsMerge)
@@ -141,6 +137,13 @@ Public Class dlgMerge
     Private Sub cmdJoinOptions_Click(sender As Object, e As EventArgs) Handles cmdJoinOptions.Click
         sdgMerge.Setup(ucrFirstDataFrame.cboAvailableDataFrames.Text, ucrSecondDataFrame.cboAvailableDataFrames.Text, clsMerge, clsByList, bResetSubdialog)
         sdgMerge.ShowDialog()
+        bResetSubdialog = False
+        SetMergingBy()
+    End Sub
+
+    Private Sub cmdColumnOptions_Click(sender As Object, e As EventArgs) Handles cmdColumnOptions.Click
+        sdgMergeColumnstoInclude.Setup(ucrFirstDataFrame.cboAvailableDataFrames.Text, ucrSecondDataFrame.cboAvailableDataFrames.Text, clsMerge, clsByList, bResetSubdialog)
+        sdgMergeColumnstoInclude.ShowDialog()
         bResetSubdialog = False
         SetMergingBy()
     End Sub
@@ -227,7 +230,7 @@ Public Class dlgMerge
                 ucrInputMergingBy.SetName(String.Join(", ", lstJoinPairs))
                 ucrInputMergingBy.txtInput.BackColor = SystemColors.Control
             Else
-                ucrInputMergingBy.SetName("No columns to merge by!" & Environment.NewLine & "Click Merge Options to specify merging columns.")
+                ucrInputMergingBy.SetName("No columns to merge by!" & Environment.NewLine & "Click Join Options to specify merging columns.")
                 ucrInputMergingBy.txtInput.BackColor = Color.LightCoral
             End If
         Else
--- a/instat/dlgMerge.vb
+++ b/instat/dlgMerge.vb
@@ -229,9 +229,11 @@ Public Class dlgMerge
                 Next
                 ucrInputMergingBy.SetName(String.Join(", ", lstJoinPairs))
                 ucrInputMergingBy.txtInput.BackColor = SystemColors.Control
+                cmdJoinOptions.BackColor = SystemColors.ControlLight
             Else
                 ucrInputMergingBy.SetName("No columns to merge by!" & Environment.NewLine & "Click Join Options to specify merging columns.")
                 ucrInputMergingBy.txtInput.BackColor = Color.LightCoral
+                cmdJoinOptions.BackColor = Color.LemonChiffon
             End If
         Else
             ucrInputMergingBy.SetName("")
--- a/instat/dlgMerge.vb
+++ b/instat/dlgMerge.vb
@@ -229,7 +229,8 @@ Public Class dlgMerge
                 Next
                 ucrInputMergingBy.SetName(String.Join(", ", lstJoinPairs))
                 ucrInputMergingBy.txtInput.BackColor = SystemColors.Control
-                cmdJoinOptions.BackColor = SystemColors.ControlLight
+                cmdJoinOptions.BackColor = SystemColors.ButtonFace
+                cmdJoinOptions.UseVisualStyleBackColor = True
             Else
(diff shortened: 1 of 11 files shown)


instat/sdgMerge.vb
```diff
@@ -15,18 +15,19 @@
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 Imports System.ComponentModel
+Imports instat.Translations
 
 Public Class sdgMerge
     Private bControlsInitialised As Boolean = False
     Private clsByList As New RFunction
     Private clsMerge As RFunction
     Private bEnableColumnSelection As Boolean = True
```
dannyparsons, line 24: Can you remove this and also remove from the Setup method? This is only needed in the other subdialog now.

dannyparsons (changes requested):
Looks good. Some suggestions:
- Change "Merging by" label to "Joining by" for consistency
- Change "Column Options" button to "Column Selection"
- Can you make the multiline textbox to be longer like it was before? I think it was 3 or 4 lines deep before.
- Can you add something to highlight the "Joining Options" button when the no joining columns message is displayed? Something like changing the `BackColour` property to "LemonChiffon"? Then it goes back to normal colour when there some joining columns.

Also, I just noticed that `dlgMergeAdditionalData` will also need to be updated as well. It only need the "Joining Options" subdialog. Could you also update this dialog?

lilyclements:
@dannyparsons I have now implemented the changes you suggested. Can you re-review?

instat/dlgMerge.vb
```diff
@@ -201,28 +211,37 @@ Public Class dlgMerge
...
+                    clsMerge.AddParameter("by", clsRFunctionParameter:=clsByList, iPosition:=2)
+                End If
             End If
             If dctJoinColumns.Count > 0 Then
                 For Each kvpTemp As KeyValuePair(Of String, String) In dctJoinColumns
                     lstJoinPairs.Add(kvpTemp.Key & " = " & kvpTemp.Value)
                 Next
                 ucrInputMergingBy.SetName(String.Join(", ", lstJoinPairs))
                 ucrInputMergingBy.txtInput.BackColor = SystemColors.Control
+                cmdJoinOptions.BackColor = SystemColors.ControlLight
```
dannyparsons, line 232: I think that should just be `Control`?
```suggestion
                cmdJoinOptions.BackColor = SystemColors.Control
```

instat/sdgMergeColumnsToInclude.vb
```diff
@@ -1,3 +1,132 @@
...
+        If ucrChkMergeWithSubsetFirst.Checked Then
+            For Each clsTmpParam As RParameter In clsByList.clsParameters
+                If Not ucrReceiverFirstSelected.GetVariableNamesAsList.Contains(clsTmpParam.strArgumentName.Trim(Chr(34))) Then
+                    ucrReceiverFirstSelected.Add(clsTmpParam.strArgumentName.Trim(Chr(34)), ucrSelectorColumnsToIncludeFirst.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
+                End If
+            Next
+        End If
+        If ucrChkMergeWithSubsetSecond.Checked Then
+            For Each clsTmpParam As RParameter In clsByList.clsParameters
+                If Not ucrReceiverFirstSelected.GetVariableNamesAsList.Contains(clsTmpParam.strArgumentValue.Trim(Chr(34))) Then
```
dannyparsons, line 126: This should be the second selector. Possibly an existing bug!
```suggestion
                If Not ucrReceiverSecondSelected.GetVariableNamesAsList.Contains(clsTmpParam.strArgumentValue.Trim(Chr(34))) Then
```

dannyparsons (changes requested):
Looks good, couple of small comments.

Also, can you remove `Optional bNewEnableColumnSelection As Boolean = True` from the SetUp function in `sdgMerge` as it's not needed now?

instat/dlgMerge.vb
dannyparsons: Actually, I found an answer here to do it https://stackoverflow.com/a/25419219/6424231

VB is over complicated some times!

```vb
cmdJoinOptions.BackColor = SystemColors.ButtonFace
cmdJoinOptions.UseVisualStyleBackColor = True

```

lilyclements: Yeah control was giving a shade lighter for some reason! I'll try this - thanks!

lilyclements:
@dannyparsons thank you for the comments - I've made the changes now.

dannyparsons (approved):
Great, looking good. I've reviewed both code and functionality so either of @lloyddewit or @rdstern or another developer doing a second review will be sufficient.

rdstern (approved):
@lilyclements that's great and I am adding my approval.

But is there a good reason for the new button to be under the Options button.  I was a bit looking forward to it being on the right, rather than below?

rdstern:
I'm happy with the position of the Join Options button.  I wondered about Column Sewlection moving to be on the same "row" as the join options and its partner on the right of the dialogue, rather than below.

dannyparsons:
Yep good idea.

dannyparsons (approved):

lilyclements:
Sorry for the confusion - do you mean like this (which is the current PR)

image

Or like this?
image

Or like something else. Sorry!

dannyparsons:
The second!

rdstern:
I would like the third!  That's like the second, but with Column SElection moved to the right, so it is lined on the right with the end of the field above.  Possibly consider moving the Join Options button so it aligned on the left with the left of the field above and also the left of the Join button above.

dannyparsons:
Yes happy with that suggestion too.

lilyclements:
Great - I've moved it like this in the most recent commit:

image

dannyparsons:
Sorry for another change, could you make them smaller so there's more space between them in the middle?

lilyclements:
Sure - this is the most recent commit 
image

dannyparsons (approved):

instat/sdgMergeColumnsToInclude.vb
```diff
@@ -1,3 +1,132 @@
...
+            SetXParameter()
+            SetYParameter()
+        End If
+    End Sub
+
+    Private Sub SetXParameter()
+        If ucrChkMergeWithSubsetFirst.Checked AndAlso Not ucrReceiverFirstSelected.IsEmpty Then
+            clsMerge.AddParameter("x", clsRFunctionParameter:=ucrReceiverFirstSelected.GetVariables(bForceAsDataFrame:=True))
+        Else
+            clsMerge.AddParameter("x", clsRFunctionParameter:=ucrSelectorColumnsToIncludeFirst.ucrAvailableDataFrames.clsCurrDataFrame)
```
shadrackkibet, line 91: Parameter positions?

instat/sdgMergeColumnsToInclude.vb
```diff
@@ -1,3 +1,132 @@
...
+        End If
+        ucrSelectorColumnsToIncludeFirst.SetVariablesVisible(ucrChkMergeWithSubsetFirst.Checked)
+    End Sub
+
+    Private Sub SetYParameter()
+        If ucrChkMergeWithSubsetSecond.Checked AndAlso Not ucrReceiverSecondSelected.IsEmpty Then
+            clsMerge.AddParameter("y", clsRFunctionParameter:=ucrReceiverSecondSelected.GetVariables(bForceAsDataFrame:=True))
+        Else
+            clsMerge.AddParameter("y", clsRFunctionParameter:=ucrSelectorColumnsToIncludeSecond.ucrAvailableDataFrames.clsCurrDataFrame)
```
shadrackkibet, line 100: Also here

shadrackkibet:
@dannyparsons any reason why this PR is missing a label? I will add a feature label. Please remove if not needed.

instat/sdgMergeColumnsToInclude.vb
lilyclements: Good point - made a PR in #6304 which adds these parameter positions

dannyparsons:
@shadrackkibet Thanks for the label, I had forgot.

Issue #4806, opened 2018-07-23
Colour palettes for ggplot2


dannyparsons:
We are still yet to implement colour palettes in the graphics system. When we do, ggplot2 now supports viridis colours https://cran.r-project.org/web/packages/viridis/vignettes/intro-to-viridis.html which seem to have useful advantages over others (and has ggplot2 approval it seems)

> scale_colour_continuous()scale_col  and scale_colour_gradient() are now controlled by global options ggplot2.continuous.colour and ggplot2.continuous.fill. These can be set to "gradient" (the default) or "viridis".
> 
> New scale_colour_viridis_c()/scale_fill_viridis_c() (continuous) and scale_colour_viridis_d()/scale_fill_viridis_d() (discrete) make it easy to use Viridis colour scales.

Ogik99:
@maxwellfundi any thoughts on this.

Ogik99:
The viridis colour scale will be applicable to the colour and fill aesthetics. I found the link below to be useful
https://ggplot2.tidyverse.org/reference/scale_viridis.html

rdstern:
This seems to be as far as we have got in the past 2.5 years.  I suggest we need another tab on the plotting sub-dialogues.  It is probably called Colour or Colour Scales?

This is complicated, but is now well described here in [Chapter 11](https://ggplot2-book.org/scale-colour.html).  

I suggest that whatever we do, this is an instance where there will be more to add, and hence being able to edit the script file will be useful.  

Here is an example where I added a colour scale for boxplots:

`last_graph <- ggplot2::ggplot(data=Data, mapping=ggplot2::aes(y=tmin, fill=month_abbr, x=make_factor(month_abbr))) + ggplot2::geom_boxplot() + theme_grey() + ggplot2::theme(axis.text.x=ggplot2::element_text(angle=90, hjust=1, vjust=0.5)) + ggplot2::xlab(NULL)+ scale_fill_viridis_d(option="E",direction=1,end=1,begin=0.5)
`
![image](https://user-images.githubusercontent.com/11226469/108492055-9a0dea80-729c-11eb-944f-50dc164d14f1.png)


I notice we have viridis in R-Instat, but this command appears to be a part of ggplot2.

I strongly suggest that doing something simple to start would be very useful.  Initially this could be mainly discrete for the sort of boxplots as above, plus colours for line plots and points at different factor levels.  Later continuous scales will be needed for maps.

[...]

PR #6283, opened 2021-03-17
Adding the colour Tab Option to the plot sub dialogue

Ivanluv:
'Feature
@africanmathsinitiative/developers this fixes issue #4806 .Still in progress


instat/sdgPlots.vb
```diff
@@ -331,9 +336,84 @@ Public Class sdgPlots
...
+        ucrNudColourScaleMapEnds.SetMinMax(0, 1)
+        ucrNudColourScaleMapEnds.DecimalPlaces = 2
+        ucrNudColourScaleMapEnds.Increment = 0.01
+        ucrNudColourScaleMapEnds.SetLinkedDisplayControl(lblColourScaleMapEnds)
+
+        ucrChkColourScaleReverseOrder.SetParameter(New RParameter("direction", iNewPosition:=4))
+        ucrChkColourScaleReverseOrder.SetText("Reverse Order Of Colours")
+        ucrChkColourScaleReverseOrder.SetValuesCheckedAndUnchecked("1", "-1")
+
+        ucrChkApplyChanges.SetText("Apply changes to colour scale")
```
lilyclements, line 406: For consistency with other checkboxes, the first letter should be capitalised here. I.e., "Apply Changes to Colour Scale"

lilyclements: Additionally
1. This checkbox currently is too small so the user cannot read past "changes" on the sdg
2. This checkbox should presumably disable the colour scale groupbox since they can't apply changes to the colour scale and then alter it
3. Is the function to this checkbox set up? If not, can you set it up?

instat/sdgPlots.vb
```diff
@@ -331,9 +336,84 @@ Public Class sdgPlots
...
+        ucrNudFillScaleMapBegins.Increment = 0.01
+        ucrNudFillScaleMapBegins.SetLinkedDisplayControl(lblFillScaleBeginColour)
+
+        ucrNudFillScaleMapEnds.SetParameter(New RParameter("end", iNewPosition:=3))
+        ucrNudFillScaleMapEnds.SetMinMax(0, 1)
+        ucrNudFillScaleMapEnds.DecimalPlaces = 2
+        ucrNudFillScaleMapEnds.Increment = 0.01
+        ucrNudFillScaleMapEnds.SetLinkedDisplayControl(lblFillScaleMapEnds)
+
+        ucrChkFilLScaleReverseColourOrder.SetParameter(New RParameter("direction", iNewPosition:=4))
```
lilyclements, line 368: Naming convention - change the checkbox name "ucrChkFilLScaleReverseColourOrder" to "ucrChkFillScaleReverseColourOrder"

instat/sdgPlots.vb
```diff
@@ -331,9 +336,84 @@ Public Class sdgPlots
...
+        ucrChkFilLScaleReverseColourOrder.SetText("Reverse Order Of Colours")
+        ucrChkFilLScaleReverseColourOrder.SetValuesCheckedAndUnchecked("1", "-1")
```
lilyclements, line 370: This is checkbox currently default checked - the default for this should be unchecked (similarly so for the reverse colour order checkbox in the `scale_colour_viridis` function

lilyclements (commented):
Looks good in general - we should discuss how to do the "Discrete" option. Ideally it would be automatically detected, but I'm not sure how possible that is right now. I'll put it up for discussion on the issue

instat/sdgPlots.vb
Ivanluv: Done

instat/sdgPlots.vb
Ivanluv: Done

instat/sdgPlots.vb
Ivanluv: Done

Ivanluv:
@lilyclements could you review this

lilyclements (commented):
Additionally:
1. There is a double space for both of the labels on the sdg where it says "Map  Begins". Can this be changed to a single space (i.e., "Map Begins")
2. The label on the sdg is "Discrete" with options "Discrete" and "Continuous". Perhaps this parameter option is better as a checkbox since the options are TRUE/FALSE. (also note that "Continuous" is spelt wrong in the dialog options, however, I assume that won't be a problem if this is changed to a boolean).
3. If "Apply Changes to Colour Scale" is checked, the colour scale groupbox should always be disabled. At the moment it isn't disabled upon first check, or upon reopening
4. If "Apply Changes to Colour Scale" is checked, then `scale_colour_viridis` function should not run. Instead only `scale_fill_viridis` should run, with the parameter `aesthetics=c("fill", "colour")`. At the moment, both functions are running when that is checked with the `aesthetics` parameter in the `colour` function 
5. Upon checking, only factors can be used in the `fill` parameter for `geom_boxplot`, hence for this boxplot dialog `discrete = TRUE` should always run. Can you set the `discrete` parameter to always be `TRUE` (and disable or hide the ucrChk so the user can't change it) when it comes from the `geom_boxplot` dialog. 
**Note: For other `geoms`, the `fill` and `colour` options can be continuous still!**

instat/dlgOneVarFitModel.vb
```diff
@@ -342,7 +342,7 @@ Public Class dlgOneVarFitModel
         clsRplotFunction.SetRCommand("plot")
         clsRplotFunction.AddParameter("x", clsRFunctionParameter:=clsROneVarFitModelFunction, iPosition:=0)
         clsRplotFunction.iCallType = 3
-        clsRplotFunction.bExcludeAssignedFunctionOutput = False
+        ' clsRplotFunction.bExcludeAssignedFunctionOutput = False
```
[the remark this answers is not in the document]
Ivanluv: Its  a mistake .I have undone it

instat/sdgPlots.vb
```diff
@@ -787,4 +900,113 @@ Public Class sdgPlots
...
+    End Sub
+
+    Private Sub Discrete_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputColourDiscrete.ControlValueChanged, ucrInputFillDicrete.ControlValueChanged
+        If ucrInputColourDiscrete.GetText = "Discrete" Then
+            clsScaleColourViridisFunction.AddParameter("discrete", "TRUE", iPosition:=7)
+        Else
+            clsScaleColourViridisFunction.AddParameter("discrete", "FALSE", iPosition:=7)
+        End If
+
+        If ucrInputFillDicrete.GetText = "Discrete" Then
```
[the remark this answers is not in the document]
Ivanluv: I have changed this to a checkbox

instat/sdgPlots.vb
```diff
@@ -331,9 +342,92 @@ Public Class sdgPlots
...
+        'Colour
+        ucrInputFillScaleColour.SetParameter(New RParameter("option", iNewPosition:=0))
+        dctOptions.Add("viridis", Chr(34) & "viridis" & Chr(34))
+        dctOptions.Add("magma", Chr(34) & "magma" & Chr(34))
+        dctOptions.Add("inferno", Chr(34) & "inferno" & Chr(34))
+        dctOptions.Add("plasma", Chr(34) & "plasma" & Chr(34))
+        dctOptions.Add("cividis", Chr(34) & "cividis" & Chr(34))
+        ucrInputFillScaleColour.SetItems(dctOptions)
+        ucrInputFillScaleColour.SetLinkedDisplayControl(lblFillScaleColourPalettte)
```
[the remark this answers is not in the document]
Ivanluv: changed this to a check box

Ivanluv:
@lilyclements  I have added a panel to control the colour options ,it however does not contain the "Apply Changes to Colour Scale" checkbox .Setting of conditions for the controls with the "Apply Changes to Colour Scale" checkbox as part of the radio buttons in the panel was challenging that's why I had it out of the panel
Currently ,the colour option panel is visible if the "Apply Changes to Colour Scale" checkbox is unchecked and not visible if the checkbox is unchecked.

lilyclements:
Great. Is this ready for review?

Ivanluv:
Its ready

lilyclements (commented):
Looks good!

1. The discrete checkbox should be checked on when this dialog is accessed through boxplots, (and additionally should be disabled if possible since boxplot can only accept factor variables)
2. Can you set default values? Currently parameters like `alpha` come up where they take the default value
3. The value for the `aesthetics` parameter for the `scale_fill_viridis` function should just be `"colour"` (rather than `aesthetics=c("fill", "colour")` - apologies for the confusion!)
4. On the colour scale groupbox, the labels aren't aligned too well with the receivers
5. If I open another plot dialog (e.g. "Scatterplot") and go to the plot options, it crashes.

maxwellfundi:
@Ivanluv is this ready for re-review?

Ivanluv:
@maxwellfundi  am still adding this functionality to other graphics dialogue.

maxwellfundi:
Okay great, let us know as soon as its ready for review!

Ivanluv:
@lilyclements  could you look at this again. The default for  the controls is gotten from  the `viridis `Rfunction Parameters. In the latest changes I have added the colour option to all the graphics dialogue so this the scatter dialogue should not  crash

lilyclements (commented):
A few comments to discuss

1. I think it is best to remove the checkbox for aesthetics. It is introducing problems here (and in R) with the `options` parameter where it always runs the same parameter value. Additionally, the original design with two checkboxes "Add Fill Scale" and "Add Colour Scale" is ideal since the user should be able to add a fill scale without adding a colour scale. I've noticed that here the user can only add a colour scale if they add a fill scale. This is a problem for a few reasons, but one of them is that some `geoms` only allow colour and not fill. For example, `geom_point`, it doesn't make sense for `geom_point` to have a fill scale since there is nothing to fill, hence it yields an error. However, it does make sense for `geom_point` to have a colour scale.
2. This leads onto 2., the `fill` options should not be available for all `geoms`, and similarly for the `colour` options. I can create a list of which `geoms` allow `fill`, which allow `colour`, and which allow both? `geom_point`, for example, cannot have a fill colour since there is nothing to fill.
3. Is there a reason why the `alpha`, `options`, etc parameters are running even if their defaults are given?
4. There's developer errors and a few issues with the parameters for the `scale_colour_viridis` function. However, I think these will be fixed if there were two checkboxes (as there originally were) - one to add `fill` and one to add `colour`

Ivanluv:
@lilyclements  I have made the suggestions you had. Could you check that again?

lilyclements:
Great - it's working really well! Just a few things,

1. The `discrete` parameter isn't being read in on the boxplot dlg for me. Do you know why this is?

2. Both "Fill" and "Colour" shouldn't be options for all sub-dialogs. Some geoms, such as `geom_point` do not allow both parameters. I've started looking at this so have put a table below of things so far. Can you make it so when certain geom's are in, certain features are disabled.

For example
* For `geom_boxplot`, the `discrete` checkbox should be checked and disabled. The user should not be able to uncheck it since this geom only accepts discrete variables and not continuous variables (see table).
* With `geom_point`, the `Add fill colours` checkbox on the sdg should be disabled since the user cannot add a `fill` parameter to this geom, only a `colour` parameter.

I'm not sure if perhaps this second job is a bit too big for now? 

| `geom` | `fill`? | `colour`? | discrete? | continuous? |
| ------------- | ------------- |------------- |------------- |------------- |
| `geom_bar` | Y|Y|Y|N|
|`geom_boxplot`|Y|Y|Y|N|
|`geom_violin`|Y|Y|Y|N|
|`geom_jitter`|N|Y|Y|N|
|`geom_point`|N|Y|Y|Y|
|`geom_freqpoly`|N|Y|Y|N|
|`geom_mosaic`|Y|N|Y|N|

rdstern:
@Ivanluv I wonder if this can be merged now - or soon anyway.
a) Can you (did you) resolve @lilyclements point 1 above?
b) For point 2 this sounds complicated to fix.  The error message is quite clear, though, so perhaps we start with that for now?  And I wonder if we can easily trap specific messages and so report them in R-Instat, rather than in R?

Then we open another issue to fix what still remains?  For example we could have a message on the tab, plus a button with Lily's table

And a question from me maybe also to @lilyclements   Do there colours apply to a graph, or to layer?  So if I have boxplots + points, then does it apply to both, or should there be a control so you can choose the layer?

dannyparsons (approved):
This looks good and works well. (1) from Lily appears to be fixed.

There are definitely lots of ways to break it by choosing options that don't apply to certain geoms or by not choosing the correct discrete option but the error messages are fairly clear. So I suggest we can continue to improve the automatic checking to help avoid errors more often but this is very useful already.

dannyparsons:
@rdstern are you happy to approve as well?

rdstern (approved):

PR #6287, opened 2021-03-17
Added Summary checkbox to the Duplicates dialog

### LINKED ISSUE #6239: The Duplicates Dialog provides summary of the result in the Output window
> @Patowhiz you are correct to query why we need a checkbox for the Prepare > Check > Duplicates. Keep it as it is now.
> However, on the Duplicates could you add the summary command just the same as it is for Prepare > Check Data > Non-numeric
> This is the line that says something like:
> 
> `summary(object=data_book$get_columns_from_data(data_name="survey", col_names="dup2"))`
> 
> It is nice that there is a result, and this is also a useful one!

As mentioned in [comment](https://github.com/africanmathsinitiative/R-Instat/issues/6080#issuecomment-783515128) by @rdstern. A summary command should be added in the **_Prepare > Column: Calculate > Duplicate_** dialog

lilyclements:

Fixes #6239

Changes made:
* Added checkbox to run a summary of the new column (similar to as seen in Prepare > Check Data > Non-numeric Columns).
* Naming convention - changed "clsdupCountIndex" to "clsDupCountIndex"

### DIFF
--- a/instat/dlgDuplicates.vb
+++ b/instat/dlgDuplicates.vb
@@ -19,7 +19,7 @@ Imports instat.Translations
 Public Class dlgDuplicates
     Private bReset As Boolean = True
     Private bFirstLoad As Boolean = True
-    Private clsDuplicated2, clsDuplicated, clsStreakFunction, clsSubsetCol, clsdupCountIndex As New RFunction
+    Private clsDuplicated2, clsDuplicated, clsStreakFunction, clsSubsetCol, clsDupCountIndex, clsSummaryFunction, clsGetColumnsFunction As New RFunction
 
     Private Sub dlgDuplicatesConstructed_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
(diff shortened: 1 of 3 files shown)


rdstern (approved):
Great

maxwellfundi (approved):
Looks all good!

lloyddewit (approved):

instat/dlgDuplicates.vb
```diff
@@ -149,24 +155,31 @@ Public Class dlgDuplicates
         clsStreakFunction.SetRCommand("duplicated_cases")
         clsStreakFunction.AddParameter("ignore", "NULL")
 
-        clsdupCountIndex.SetRCommand("duplicated_count_index")
+        clsDupCountIndex.SetRCommand("duplicated_count_index")
+
+        clsGetColumnsFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_columns_from_data")
+
+        clsSummaryFunction.SetRCommand("summary")
+        clsSummaryFunction.AddParameter("object", clsRFunctionParameter:=clsGetColumnsFunction)
```
shadrackkibet, line 163: Position?

shadrackkibet:
I tested and got a bug when the **Duplicate only option/Index number of duplicates** is selected. This happens when the new checkbox **Display summary** is checked.

image

shadrackkibet:
I also suggest the summary be done internally for our functions i.e `duplicated_count_index` rather than as a separate function.  We just need to add a boolean for this. Perhaps this is difficult to do at the moment because some other functions used in this dialog are from other packages. I think we will need to write a wrapper for all these options in future so that we only have one function. Then we can print the summary internally.

lilyclements:
@shadrackkibet thanks for this - I've fixed the bug in PR #6303

Should we write an issue on adding the summary internally, if it is difficult to do at the moment?
Similarly, I looked at the "Non numeric columns" dialog code to implement the output in this "duplicates" dialog. Would we want to add the summary internally there as well?

shadrackkibet:
Actually, Let's have this as a discussion not an issue including Non-numeric columns. Once we discuss with others we can then open specific issues.

PR #6288, opened 2021-03-17
Improved detection & display of data types

### LINKED ISSUE #6266: Display of data types in the R-Instat grid
Currently we define some types of data.  This could usefully be extended - at least visually in the grid.
For example, from the structured menu, we can define columns to be `circular`.  Then this is noted in the column metadata, but nothing is indicated in the name in the grid.
We can now define columns as `survival`.  This is not noted (currently) in the metadata, nor the grid.
We will have low flow variables etc.  there are lots of different ones.
We define dates as D.  So far this is for daily data.  We have date-time and time coming.  How do we show this in the grid.

@volloholic suggestion is simple, and is as follows:
a) We use D - or should it be d for date, or date-time or time.  They are all a date-time sort of thing.
b) We use s for structured, so this could be survival, circular, low-flow.  It just indicates that it is defined.

Separately to this is that details of the structure, and the type of date is given in the metadata.  So that needs to be added for the survival.  

This needs @dannyparsons to check.  If agreed it needs to be implemented.

If quicker for him to do than explain, then he should.  Otherwise he could suggest who implements.

dannyparsons:
fixes #6266 
- Symbols now in upper case
- Dates group expanded to date and time related columns
- Added (S) for structured (circular only for now)
- Added (A) for application (not implemented yet)


rdstern (approved):
It seems to work fine, so I approve.  
I tried with data and should say I don't like the use of capital letters any more than I think they are frowned on by tidyverse?  It isn't so bad for D, but having lots of F and O.F and C is a bit shouty?

dannyparsons:
It is a bit, but I think it adds clarity. It's probably more noticeable for existing users who are used to the lower case. 
We already use it for D and also we had to have L for logical because "l" wasn't clear. So if we want consistency capitals is the only way to go. But I'm not completely against mixing upper and lower.

Could we try it out up to the next release and if you're still not convinced we go back to some lower case?

instat/clsGridLink.vb
```diff
@@ -438,20 +438,29 @@ Public Class clsGridLink
                 For k = 0 To dfTemp.ColumnCount - 1
                     strCurrHeader = lstColumnNames(k)
                     If vecColumnDataTypes(k).Contains("factor") AndAlso vecColumnDataTypes(k).Contains("ordered") Then
-                        fillWorkSheet.ColumnHeaders(k).Text = strCurrHeader & " (o.f)"
+                        fillWorkSheet.ColumnHeaders(k).Text = strCurrHeader & " (O.F)"
```
lloyddewit, line 441: The for loop could be made more readable by replacing with:
```
                For k = 0 To dfTemp.ColumnCount - 1
                    strCurrHeader = lstColumnNames(k)
                    Dim strType As String = vecColumnDataTypes(k)
                    Dim clsHeader As ColumnHeader = fillWorkSheet.ColumnHeaders(k)
                    clsHeader.Text = strCurrHeader
                    clsHeader.TextColor = Graphics.SolidColor.DarkBlue

                    Select Case True
                        Case strType.Contains("factor") AndAlso strType.Contains("ordered")
                            clsHeader.Text &= " (O.F)"
                            clsHeader.TextColor = Graphics.SolidColor.Blue
                        Case strType.Contains("factor")
                            clsHeader.Text &= " (F)"
                            clsHeader.TextColor = Graphics.SolidColor.Blue
                        Case strType.Contains("character")
                            clsHeader.Text &= " (C)"
                        Case strType.Contains("Date"), strType.Contains("POSIXct"), strType.Contains("POSIXt"),
                                 strType.Contains("hms"), strType.Contains("difftime"), strType.Contains("Duration"),
                                 strType.Contains("Period"), strType.Contains("Interval")
                            clsHeader.Text &= " (D)"
                        Case strType.Contains("logical")
                            clsHeader.Text &= " (L)"
                        Case strType.Contains("circular") ' Structured columns e.g. "circular" are coded with "(S)"
                            clsHeader.Text &= " (S)"
                            ' Types of data for specific Application areas e.g. survival are coded with "(A)"
                            ' No examples implemented yet.
                            'Case strType.Contains()
                            '    clsHeader.Text &= " (A)"
                    End Select
                Next
```

lloyddewit (changes requested):

instat/clsGridLink.vb
dannyparsons: Thanks for this suggestion I've implemented that. I kept the `If Else` as I think that's more readable and logical for other developers, particularly new ones who might look at this.

lloyddewit:

rdstern (approved):

instat/clsGridLink.vb
Patowhiz: @dannyparsons @lloyddewit sorry I came late to the discussion. Now that I can see we have Date instead of date, Character instead of character and factor instead of Factor(all of which are done that was for a good reason), wouldn't it be ideal to do first convert `strType` to lowercase before proceeding with checking `contains()`. I think assuming the case in this context could make the code be a bit "resilient" in terms of logical bugs incase future implementation of the R-Instat script that produces the data frame is changed.

lloyddewit: I thought that the checks were deliberately case-sensitive. I don't know enough about the context to judge if a case-insensitive comparison would be better. I'm OK with whatever @dannyparsons prefers.

dannyparsons: @Patowhiz I could argue it either way. Lower case removes case sensitivity, but R is case sensitive, so for someone familiar with R, it would look strange to see some of these in lower case. There is a very small chance of having different data types in R that only differ by case, so this avoids a potential issue arising from that.

PR #6059, opened 2020-11-03
Adding controls to dlgPolarAnnulus

HawardKetoyoMsatsi:
Fixes #5966
Added controls to dlgPolarAnnulus and implemented R code.
@africanmathsinitiative/developers, this is ready for review.

PR #6050, opened 2020-10-29
Adding controls to dlgPolarCluster

HawardKetoyoMsatsi:
Fixes #5966
Added controls to dlgPolarCluster and implemented R code.

@africanmathsinitiative/developers , @shadrackkibet , this is ready for review.

PR #6023, opened 2020-10-16
Added the second circular plot (Polar Frequency) from openair package #5966

HawardKetoyoMsatsi:
Fixes #5966 

@africanmathsinitiative/developers. This is ready for review.

PR #6290, opened 2021-03-17
Implemented dialog for Polar plot from Openair package

HawardKetoyoMsatsi:
5 different polar plots from the openair package are implemented here.
This PR replaces PRs. #6059 #6053 #6050 #6023 #6022


HawardKetoyoMsatsi:
@shadrackkibet @maxwellfundi @rdstern @lloyddewit Kindly review the current state of the dialog. I have removed the Facet options after a discussion with @shadrackkibet I am going to open an issue over the same.

rdstern (changes requested):
@HawardKetoyoMsatsi and @shadrackkibet I very much like the structure that you have now. I am afraid it is still a long way from being finished, simply because there are so many refinements that are possible with these plots. 
The examples I tried did work.  Two suggestions, then I hope we could merge and sort out the rest in a further issue.  The big plus point, from the usage point of view is that we can try adding features with the script window first.
a) Please look at the other circular dialogues.  I have a feeling they have wind speed before wind direction?  I quite like wind direction first.  But let's be consistent.  This might mean another issue to change the other circular dialogues.
b) The facets are absent.  I think Haward has greatly complicated his life by starting with the nice features in openair, and then trying to add the way we work in R-Instat.  For simplicity, let's add the facets the R-Instat way and then consider whether we can take advantage of the extra facilities in openair.  So add 2 single receivers, called Facet (Optional) and Facet2 (Optional) which call openair's Type argument, and assume the facets are from variables in the data frame - not from his special method from the date variable.  
Then let's merge and proceed.

We can't use the lattice package and not have lattices (facets!)

HawardKetoyoMsatsi:
@rdstern Kindly take a look, I have implemented the faceting options and placed with speed before wind direction for consistency with other dialogs. This however brought Plot against: label before wind direction (Kindly comment on whether we can leave it there or look for a space for it).

rdstern:
I comment as I go through these in turn.  It would be really good to get this merged in this version:
a) In the third option, i,.e. Polar Plot change Plot against to Windspeed  - as in most of the others.  It says that's the default, so let's say that too! 
b) Same for 5th option Polar cluster
c) In Polar cluster move the Cluster label to the left, so it is not under the next control.
d) Is there anything different in your code for the polar cluster plot, compared to the others?  The others all transfer to the R viewer.  This one doesn't. 

Overall, great - and clearly a lot of work.  The dialogues look neat.  Well done.


**_Polar Frequency_**

Seems ok to me:

image

Giving:
image

(I don't know what the Transform checkbox is doing, and later we should add a tooltip to explain.

**_Percentile Rose_**

image


Giving:

image


**_Polar Plot_**
image


Giving:
image



**_Polar Annulus_**

This gave an error of not enough data in certain circumstances.  For example it needed to Date-Time Variable, and I couldn't facet by time of day.  I could by another pollutant:

So this needs investigation, but I suggest we can merge first and then discover more.
But this works:
image


Giving:

image

**_Polar cluster_**

image

Giving:

image

HawardKetoyoMsatsi:
@rdstern I have made the required changes kindly take a look. I am not able to tell what is happening with the polar cluster, they all use the same function.

rdstern (dismissed):
Great @HawardKetoyoMsatsi quite a tour de force!

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -2280,6 +2280,41 @@ plot_declustered <- function(data, station_col_name, element_col_name, threshold
   }
 }
 
+#This function creates awrarper around functions from openair package
+other_rose_plots <- function(data, type1_col_name, type2_col_name, date_col_name, wd_col_name, ws_col_name, main_method, single_pollutant, multiple_pollutant, ...) {
+  type = "default"
```
shadrackkibet, line 2285:
```suggestion
  type <- "default"
```

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -2280,6 +2280,41 @@ plot_declustered <- function(data, station_col_name, element_col_name, threshold
...
+  if (missing(type1_col_name) == FALSE && missing(type2_col_name) == FALSE) {
+    type <- c(type1_col_name,type2_col_name)
+  }
+  if (missing(type1_col_name) == TRUE && missing(type2_col_name) == FALSE) {
+    type <- type2_col_name
+  }
+  if (missing(type1_col_name) == FALSE && missing(type2_col_name) ==TRUE) {
+    type <- type1_col_name
+  }
```
shadrackkibet, line 2294:
```suggestion
  if (!missing(type1_col_name) && !missing(type2_col_name)) {
    type <- c(type1_col_name, type2_col_name)
  }
  if (missing(type1_col_name) && !missing(type2_col_name)) {
    type <- type2_col_name
  }
  if (!missing(type1_col_name) && missing(type2_col_name)) {
    type <- type1_col_name
  }
```

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -2280,6 +2280,41 @@ plot_declustered <- function(data, station_col_name, element_col_name, threshold
   }
 }
 
+#This function creates awrarper around functions from openair package
+other_rose_plots <- function(data, type1_col_name, type2_col_name, date_col_name, wd_col_name, ws_col_name, main_method, single_pollutant, multiple_pollutant, ...) {
```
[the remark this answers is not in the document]
HawardKetoyoMsatsi: Two controls are passing the same parameter name that's why I chose to have different names for each control

instat/dlgOtherRosePlots.vb
```diff
@@ -1,3 +1,293 @@
...
+                If Not ucrReceiverX.IsEmpty AndAlso Not ucrReceiverMultiplePollutants.IsEmpty Then
+                    ucrBase.OKEnabled(True)
+                Else
+                    ucrBase.OKEnabled(False)
+                End If
+            End If
+            If rdoPolarAnnulus.Checked Then
+                If Not ucrReceiverWindSpeed.IsEmpty AndAlso Not ucrReceiverMultiplePollutants.IsEmpty Then
+                    ucrBase.OKEnabled(True)
+                Else
```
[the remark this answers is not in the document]
HawardKetoyoMsatsi: Done

shadrackkibet (dismissed):

instat/dlgOtherRosePlots.vb
```diff
@@ -1,3 +1,285 @@
...
+                ucrInputStatistic.Visible = True
+                lblStatistic.Visible = True
+            Else
+                clsOtherRosePlots.RemoveParameterByName("statistic")
+                ucrInputStatistic.Visible = False
+                lblStatistic.Visible = False
+            End If
+        ElseIf rdoPolarAnnulus.Checked OrElse rdoPolarPlot.Checked Then
+            clsOtherRosePlots.AddParameter("statistic", Chr(34) & ucrInputStatistic.GetText & Chr(34), iPosition:=11)
+        ElseIf rdoPolarCluster.Checked OrElse rdoPercentileRose.Checked Then
```
lloyddewit, line 261: I would expect this sub to either add or remove the `statistic` parameter but in its current implementation it may do neither.
Should this `If` statement have an `Else` at the end that removes the parameter?

rdstern (changes requested):
@HawardKetoyoMsatsi a very small point.  I put wind speed into the first dialogue.  That is cleared in the 3rd and 5th, where you used to have other variable.  It is remembered by the other cases.
I could accept that if you don't have time

instat/dlgOtherRosePlots.vb
HawardKetoyoMsatsi: Yea for the last ElseIf an Else would have been better

shadrackkibet (approved):

HawardKetoyoMsatsi:
@rdstern I think it will be okay to accept if the two plots with cleared inputs as they are then we can discuss on how to either restrict it to wind Speed. Some plots take wind speed as a parameter name "`wd`" while the two take wind speed as parameter name "`x`"

PR #6297, opened 2021-03-18
Display output message when creating a key or link in the Add Key or Add Link dialogs

### LINKED ISSUE #6072: Prepare > Keys and Links Add Key and add Link
They both seem to work.  But we now report something into the output window for most changes like this.

Could that we added here too.  It just needs one line.

A larger change (not urgent) would be to allow links between variables with different names.

lilyclements:
Fixes #6072 

This PR
* Adds an output message when a new key is created containing the columns used to create the key, and the name of the key
* Adds an output message when a new link is created containing the key used to create it, and the name of the link


dannyparsons:
Sorry Lily, could you do this internally in the `add_key` and `add_link` functions in R so that the output message is done in R? Instead of adding after functions to the dialog. The view key and view link dialogs already do this, so it can be consistent, but the output could be improved. In the `add_key` method you can call a common function, if that makes sense?

lilyclements:
@dannyparsons I have implemented for keys and can do for links. Do you mean something like this?

lilyclements:

shadrackkibet:
A question and a bug report.

- What should be the message when we are overwriting a **key**? Currently, no message is printed in the output window.
- When a key is deleted the receiver should be cleared for now. See an error below.
image
image

dannyparsons:
Yes this looks good. Ideally the printing code being repeated when adding/viewing could be a separate common function, but I suggest this is fine as it is.

I think the receiver clearer after deleting is a general issue in dialog, not sure there is anything specific to do on this dialog.

lilyclements:
For the bug in the remove keys dialog, is there a way to run in the dialog a "internal R-code"?
For example, I want to run something like

```
If [R-Code] then
ucrReceiver.ClearReceiver()
End if
```

Where the R-Code is
`!key_name %in% data_book$get_key_names(data_names = "DataFrame")`

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1819,6 +1819,9 @@ DataSheet$set("public", "add_key", function(col_names, key_name) {
     if(length(private$keys) == 1) self$append_to_variables_metadata(setdiff(self$get_column_names(), col_names), is_key_label, FALSE)
```
lilyclements, line 1819: Up on line 1816, the code is
`if(key_name %in% names(private$keys)) warning("A key called ", key_name, " already exists. It wil be replaced.")`

Can I change this to
`if(key_name %in% names(private$keys)) cat("A key called ", key_name, " already exists. It wil be replaced.")`

Given @shadrackkibet's suggestion to let the user know that the `key_name` is being replaced

shadrackkibet: I think this should be a warning. Not sure if that will be printed in the output window.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1813,7 +1813,7 @@ DataSheet$set("public", "add_key", function(col_names, key_name) {
   }
   else {
     if(missing(key_name)) key_name <- next_default_item("key", names(private$keys))
-    if(key_name %in% names(private$keys)) warning("A key called ", key_name, " already exists. It wil be replaced.")
+    if(key_name %in% names(private$keys)) warning("A key called", key_name, "already exists. It will be replaced.")
```
lilyclements, line 1816: Given @shadrackkibet's suggestion to let the user know that the key_name is being replaced, can I change this `warning` to `cat` such that is reads:

```if(key_name %in% names(private$keys)) cat("A key called ", key_name, " already exists. It wil be replaced.\n")```

lilyclements: @shadrackkibet - sorry I accidentally removed your comment since I moved this discussion to be on the correct line of R-code. You commented:

> I think this should be a warning. Not sure if that will be printed in the output window.

To which I reply: a warning does not print in the output window, however, `cat` does.

lilyclements: Had a little look [here](https://stackoverflow.com/questions/36699272/why-is-message-a-better-choice-than-print-in-r-for-writing-a-package) and the only real difference I can tell between `warning` and `cat` is that a `warning` is in red font, whereas `cat` is part of the output so is in black (or blue in R-Instat, assuming the settings in RStudio and R-Instat haven't been changed).

I see the advantage to it being in red font in RStudio, however warning messages aren't given in R-Instat. Is there a way to "un-suppress warnings" just for this function in R-Instat?

dannyparsons: Our main R-Instat problem is that we don't currently show warnings and message. We shouldn't really be using `cat()` at all, but until we can display those messages and warnings we have to use `cat()` for output. So I think leave it as warning, as that is "correct" and it's not essential to get the output.

lilyclements: Okay thanks. In that case, I've made all the changes and this is ready for review

shadrackkibet (approved):

maxwellfundi (approved):

rdstern (approved):
Great.

PR #6301, opened 2021-03-19
New dialog for climatic daily data entry

### LINKED ISSUE #6250: Adding data entry - grid (spreadsheet) style - to R-Instat for daily climatic data
I specify the problem, using Zambia Met service as a case study.  I suggest a simple solution could be useful more generally.

Zambia Met would like to update the daily records of rainfall and max and min temperature in 6 stations in Southern and Eastern Province. Their current data is in Clicom format.  It imports easily into R-Instat and is easy to tidy.  They currently have 2 data entry staff.  The data typically start in 1950 and is reasonably complete to about 2010.  It is then very patchy for the recent years.

We think we have a solution to this problem, which is for them to download and install Climsoft.  Then import the existing station and current data.  Then add the new months of data.  This will be combined with capacity-building, in that modest funds will be used to invite a Zambia Met person for 2 weeks from each province.  They will arrive with the data to be entered and then work with HQ staff to enter the data.

This is part of a more general problem.  Zambia has over 40 stations and 10 provinces.  Maybe this formula could be used for all their stations.  I assume (in addition) it also has daily rainfall data from many volunteer stations, but this was not discussed.

Staff have found it easy to download, install and use R-Instat.  

Now Climsoft has good data entry facilities for these data, but it still lacks a grid-style data entry system - which was a feature of its Clicom predecessor.  We have long wanted to add a grid-style data entry system to Climsoft.  

The Zambia situation is common, and is also reflected in Mozambique.  

It would be even simpler for Zambia Met if they could initially use a spreadsheet, or R-Instat, to enter these daily data.  Perhaps any proposed grid-style data entry system for Climsoft could also be included in R-Instat?  I explore what is needed here, particularly to check if a usable (limited) system could be easy to provide for R-Instat.  Then Zambia Met could start with R-Instat and add Climsoft later for more general data management of the climatic database.

I note that the 41 stations includes many other elements, but Zambia Met are happy to start with these three.  I should mention that a next stage in R-Instat will be the quality control.  This is already quite good in R-Instat, but temperaure QC could be enhanced if more temperature elements, e.g. dry bulb and wet bulb.

There are also other data, on a 3 or 4 hr basis.  But a simple solution for data entry of just daily data would be very useful to many services.

In Climsoft the Station details must be entered before the daily records can be added.  I suggest our proposed solution could include the same restriction, partly because it could simplify the process.
Similarly element information could be available.

In Climsoft there is also flag information that may be entered at the same time.

Most of the data entry sheets include monthly totals.  They are obvious for rainfall, but are often included for many elements and provide an excellent check.  (Of course if there is a difference it may be because of data entry or from original calculation errors.)

Often (as in Zambia) there will be some station data already, and the task will be to enter the gaps.  They will usually be at least monthly gaps.  In some circumstances there may be shorter gaps to infill.  Sometimes a new station may be added.

How could we start doing this?  Is there something we could specify, in R-Instat, It could be useful in R-Instat even if it starts with the entry of single elements - as in Climsoft - with additional facilities coming later.  

The Genstat grid system has included a general data entry facility with double-entry capabilities for many years.  This is general.  I wonder if something very simple could be setup quickly?

It may be that a simple solution is possible, but using a third package, maybe Excel or Open Office, or even something else.  I wonder what the options are?

N-thony:
Fixes #6250


lloyddewit:
@N-thony When you create a PR, please always add a description explaining the reason for the PR. Please also explicitly state whether you are still working on the PR, or whether it is ready for review.
Thanks

Patowhiz:
code generated by the dialog

```
data_book$save_data_entry_data(data_name="Sheet1", new_data=(Station="ACHEGO", Date1=as.Date(c("1999-06-30","1999-08-21","1999-10-12","1999-12-03")), Tmin=c(16,18,22,18), Tmax=c(26,24,28,20)), rows_changed=c(23557,23609,23661,23713))
rm()
```

N-thony:
@Patowhiz thanks. I think this works very well as needed by @dannyparsons .
I also checked the dodoma dataset. See the code below

# Code generated by the dialog, Climatic Data Entry

data_book$save_data_entry_data(data_name="dodoma", new_data=(date=as.Date(c("2001-01-03","2001-01-05","2001-01-04","2001-01-06")), rain=c(2.1,2.3,5.4,1.3), tmax=c(28.5,27.8,28,26.2), tmin=c(19.9,18.5,17,18.7)), rows_changed=c(24110,24112,24111,24113))
rm()

@dannyparsons if you are happy with this. Could you send now the R script for saving the data changed? Thanks.

N-thony:
@dannyparsons  I have tried to run the R function and I'm getting the error below

N-thony:
> @dannyparsons I have tried to run the R function '**save_data_entry_data**' and I'm getting the error below
> image

@dannyparsons I have fixed this bug

N-thony:
@africanmathsinitiative/developers @rdstern this is ready for the first review. Thanks.

rdstern (changes requested):
@N-thony I am excited to have my first look.  Please can you enable the menu item first.

N-thony:
@rdstern the Data Entry menu is enabled now. Could you proceed with your first test? Thanks

rdstern (changes requested):
@Patowhiz @N-thony @dannyparsons  wow it already works!

I only found 2 things at this stage.  Then I am so relieved that you guys have got this far.  
1. Initial date starts as today.  Could it default to the end of the data.
2. I tried with Dodoma, where there are actually 2 months to enter at the end of the file, so starting from November 2013.  It all worked!  Then I closed the data file and went to Zambia, where there are 6 stations - that's a real example where we DO want some data entry.  It crashed and still had Dodoma in its memory.  So I have not checked that yet.
3. First enhancement - it would be good if there were something, by default in the output window to show that it worked.  I am not sure what, so just a random text initially, e.g. "Whoopee" would be fine to start!

N-thony:
@rdstern I think your comments are resolved. Could you test the changes? Thanks

N-thony:
@rdstern I think your comments are resolved. Could you test the changes? Thanks

maxwellfundi:
@rdstern  @shadrackkibet @Patowhiz  this is ready for re-review.

dannyparsons:
@rdstern and I need to discuss an option for end dates. This will be an additional feature so I think it could still be tested now by one reviewer for any current issues, and then more once we have proposed the end dates option.

rdstern (changes requested):
I first tried with Dodoma - 1 station.  
Works fine just details compared to next item:
a) When there is nothing to update, then you can (correctly) enter data, but Ok is still disabled.
But having updated, with Ok, then it should be disabled again.
I would like an improvement to be that there are data to enter, and the other variables are locked.  I was able to change the dates and it was quite happy.
The other important feature there is the the Enter key is good to go down and tab goes across.  That's good, but it should only go across those fields I am entering.
I notice the date is yet to be fixed.  It is still today's date, rather than the last date in the file.  That's a luxury, because it does trap the error nicely when you leave it like that.
The big problem is with a file with multiple stations.  I tried the 6 Zambia stations and then to enter data from the start, i.e. 1 Jan 1945.  It freezes up.  I think it is because of multiple stations, but perhaps it is because I am asking from the start of the data?  I'll check.  It would be good for the developers to use the zambia data - ar at least data where there are multiple stations for their initial testing

rdstern (commented):
aha, I get the same problem with Dodoma, when I enter data from the start of that file.  It freezes up.  So it is probably a start of data problem and may have nothing to do with single or multiple stations

rdstern (commented):
No it freezes up for me when I try to enter data in the middle too!  I tried dodoma August 1979

rdstern:
@N-thony and @dannyparsons this is already starting to look very nice.
a) I assume only range is working so far?  I tried month, but it trapped this and said there was no data?
b) I included date as a variable to edit and it accepted this, but it was formatted as numeric not as a date.
c) It could also be useful to be able to view the date variable, but the second receiver didn't allow that.
d) You locked the initial date and station variables - good.  Could you also lock all the View variables in the same way?
e) Could you arrange for <TAB> to not take you into the locked variables.  (That will facilitate <TAB> to support entry of data across, rather than down.
f) You say (Refresh) rather than Reset which we use on the main dialogue.  Why not use Reset.  Then possibly the message could be "This deletes the data entered here.  (No other data are affected). Do you want to continue.  Perhaps also a tooltip "This is for situations where you have entered incorrect data and wish to start again.  It only affects the data that you see here."
g) The data show as NaN.  I assume it could be blank instead?  (This might even be an option in specifying the data entry.)

N-thony:
> @N-thony and @dannyparsons this is already starting to look very nice.
> a) I assume only range is working so far? I tried month, but it trapped this and said there was no data?
> b) I included date as a variable to edit and it accepted this, but it was formatted as numeric not as a date.
> c) It could also be useful to be able to view the date variable, but the second receiver didn't allow that.
> d) You locked the initial date and station variables - good. Could you also lock all the View variables in the same way?
> e) Could you arrange for to not take you into the locked variables. (That will facilitate to support entry of data across, rather than down.
> f) You say (Refresh) rather than Reset which we use on the main dialogue. Why not use Reset. Then possibly the message could be "This deletes the data entered here. (No other data are affected). Do you want to continue. Perhaps also a tooltip "This is for situations where you have entered incorrect data and wish to start again. It only affects the data that you see here."
> g) The data show as NaN. I assume it could be blank instead? (This might even be an option in specifying the data entry.) Danny says this may not be so easy to do.

@rdstern day and range, there is a small issue that we are trying to fix for month.

rdstern:
One more detail. When I use Zambia6 it correctly enters the Station field, but does not enter a particular station.  I suggest the default is the first station.  Once a station is entered, then there is no option to clear, so we might as well start with a station there.

N-thony:
> > @N-thony and @dannyparsons this is already starting to look very nice.
> > a) I assume only range is working so far? I tried month, but it trapped this and said there was no data?
> > b) I included date as a variable to edit and it accepted this, but it was formatted as numeric not as a date.
> > c) It could also be useful to be able to view the date variable, but the second receiver didn't allow that.
> > d) You locked the initial date and station variables - good. Could you also lock all the View variables in the same way?
> > e) Could you arrange for to not take you into the locked variables. (That will facilitate to support entry of data across, rather than down.
> > f) You say (Refresh) rather than Reset which we use on the main dialogue. Why not use Reset. Then possibly the message could be "This deletes the data entered here. (No other data are affected). Do you want to continue. Perhaps also a tooltip "This is for situations where you have entered incorrect data and wish to start again. It only affects the data that you see here."
> > g) The data show as NaN. I assume it could be blank instead? (This might even be an option in specifying the data entry.) Danny says this may not be so easy to do.
> 
> @rdstern day and range, there is a small issue that we are trying to fix for month.

N-thony:
@lilyclements can you try now? Thanks

rdstern:
@N-thony this continues to advance - which is great.
a) I know you are still working on the Month option.  Good that the day and range are working.
b) As we discussed, the Station and date field on the left in the grid should be read only.  You can click on them to go there, but you can't delete them   The same applies to the View variables on the right.
c) The <Tab> goes across from one field to the next.  It only goes to the variables you can edit.  So, it goes to the next row (and the first field to edit) after the last field to edit.  
d) So if there is just one variable to edit, then <Tab> goes down - just like the <Enter> key.
e) As discussed, move the Data antry button down towards the bottom - above the Comments control.  And the Data Checks button should move to the right.  

Now some new controls to add below the date controls.

f) Transforming: Add a transform checkbox. The label is Transform.  If checked it opens a pull-down control. You can choose a transformation, or type  a number. If you type a number, then it divides by that number, before copying back to the data frame. The three initial default options are `10 `(to divide by 10, so no decimals need be entered), `inch to mm` (to divide by 25.4) and `F to C` (to subtract 32 and divide by 9/5) 

(Note using it is kept simple, so you have only a single option at a time. If you wish to do more than one option, then they must be on separate data entry occasions.)

g) `View More Days:` This is another checkbox, default unchecked.  If checked then it opens an up-down, followed by pull-down choice.  The up-down is from 1 upwards, with perhaps an upper limit of 31.  The pull-down has 3 options with the first selected by default, namely `Before, After, Both`  

Implementing in the grid (simply!) adds this number of rows, as long as there are data available.  They are also read only rows - just like the extra variables are read only.  

h) A label `Allowed Entry:` Followed by a pull-down with 3 current options:   Numeric,  No decimal, Character.  This is again implemented for all variables.

i) A checkbox with Default Value:  Default unchecked.  If checked  then a numeric field opens with zero as the default.  Then another checkbox **_Rainfall Only_**, default unchecked.   (If checked, then fills all missing rainfalls with zero.  Non-missing rainfalls are unaffected.)

j) In the grid have an extra button on the right with `View Transformed`.    It is disabled unless the Transform option is chosen.  If clicked it applies the transformation - which could be useful for visual checking.  For example typing 312 is quick, but the useful to see it as 31.2 to check against the paper record.

rdstern:
This addition is just for the month option.  
[Ghana Tamale photos of daily data sheets.docx](https://github.com/africanmathsinitiative/R-Instat/files/6215079/Ghana.Tamale.photos.of.daily.data.sheets.docx)

Look at these data sheets for information on why this addition is useful.  At the bottom is the Sum and the Mean. Initially we'll just use the Sum.   (The mean isn't calculated for rainfall.  (Initially perhaps it could be, in the Code.  Perhaps the count of non-zero values would be more useful.)
So, in the monthly sheet there is at least one extra line.  Start with just Sum.  The word Sum is written if possible in the Date variable, below the last month, and perhaps in a  different colour.  This is to enter the sum, from the sheet.
Below that is `Data Sum`.  This is the sum calculated from the data.  This row is read only.  It is calculated from the data as entered (including those already there!)
Below that is `Difference`.  This could go green if it is zero.  It is Grey otherwise (I am not so keen on Red!)

PR #6307, opened 2021-03-20
split translation files into menu and non-menus

dannyparsons:


dannyparsons:
@lloyddewit merging this as only translation file changes

PR #6309, opened 2021-03-20
Improved the One Variable Frequencies Graph

### LINKED ISSUE #6279: Improve the One Variable Frequencies Graph
The one variable frequency graph could be very useful in teaching and in simple data processing.  
It has just been corrected, so it now works.  But there are many ways that options could be added to improve it.  The commands could also be improved in at least 2 ways.  This could be an early job for @HawardKetoyoMsatsi. 

- [x] Everything in the R code is in 1 line.  This makes it difficult to edit the graph options in the script window.  Here is a simple example.

```
# Code generated by the dialog, One Way Frequencies

last_graph <- sjPlot::plot_grid(x=list(x=sjPlot::plot_frq(data=data_book$get_columns_from_data(data_name="survey", col_names="yield", force_as_data_frame=TRUE, remove_labels=TRUE, drop_unused_filter_levels=TRUE),show.prc=FALSE,show.values=FALSE)))

rm(last_graph)

```
In most dialogues the data frame is got first, and then the data are processed.

- [x] The frequency graph is drawn in a grid.  This is needed if more than 1 variable is plotted - I assume facets can't be used?  But it isn't needed for a single variable.  So could just sgPlot::plot_frq be used then directly?

Then James, John and Tobias can start their work.  This function has many options that could usefully be included on the dialogue, and hence the plots.  I also don't know why they are labelled A, B, etc and not with the names of the variables.    

As a start I found:

- [x]  The line option doesn't work.  The box, violin histogram do work.  The dot is OK.

- [x] We can control the size, i.e. width of the bar.  If small, then it becomes a vertical line - looks quite nice.

- [x] I couldn't get the colors to change - needs a bit more work.

- [x]  I can't see how to get the name of the variable included.  But it does automatically include the label - unless you say otherwise.  (Nice to have encouragement to use variable labels.)

shadrackkibet:
fixes #6279

- Designer fixes and Tab order
- Added parameter positions
- Added tags
- Added ggplotify::as.ggplot - therefore View Last Graph should work.


shadrackkibet:
@rdstern what do you think about adding the following into the subdialog? Perhaps @jkmusyoka and the interns can continue to specify more from here?

image

rdstern:
@shadrackkibet this looks good.  Great!

shadrackkibet:
@rdstern you can now go ahead with the testing.

rdstern:
@shadrackkibet this is looking very nice now.  I really like the use of ggplotify, because that turns even the grid into a ggplot.  I wonder if that will also work on the 1 variable graph dialogue.  There we start with the diamonds data and plotting them all is a combined graph and so won't show in the R-viewer.  Now perhaps it might?

One point is that we need to add ggplotify to the list of packages.  It isn't yet included.
Then I returned to have a problem again with the line graph.  All the others worked for the survey data.  I particularly tried with the 6 variables from village to yield.  With the line graph it said
image

shadrackkibet:
@rdstern I would consider the use of `ggplotify` case-by-case because it doesn't always work, it depends on a number of factors.

In one variable the fix is easy but not using `ggplotify`, there we need to use `patchwork::wrap_plots()` to combine graphs instead of `gridExtra::grid.arrange()`. Patchwork is already a package in R-Instat and I like it because it always returns a `ggplot` object and developed by one of the ggplot2 authors. It's classified as ggplot extension [package](https://mode.com/blog/r-ggplot-extension-packages/). I can fix this on a separate PR or we assign it to someone else then I provide support.

On the bug. Which kind of a graph do you expect here? You have categorical variables on a line graph, they are best displayed on a bar graph? We can't do much to allow this to work because it's from sjplot, all we can do is to hide these variables when "line" is selected. Are you happy with this fix? @jkmusyoka can we use this as a teaching example where we allow students to get this kind of error? Can it be used to explain different types of variables and graphics?

maxwellfundi:
@rdstern  what are your thoughts on this? can we have the issue addressed here agreed on and another PR opened with the other changes?

rdstern (approved):
@shadrack - great.

lloyddewit (approved):

PR #6320, opened 2021-03-22
Added RDS file recognition functionality in Options dialog

N-thony:
@lloyddewit I added the functionality to recognize RDS file as suggested in issue #6082. This complete PR #6160.
@lloyddewit could you test, please? Thanks.


lloyddewit:
@dannyparsons @rdstern @N-thony  I think this PR was triggered by one of @dannyparsons 's comments that I added to issue #6082 (see below). Normally the associations between a file type and an application are managed by the user in the normal Windows settings. This is how it is done in other applications (e.g. Word, Excel etc.). For RDS files and R-Instat, the assocation is already done in this standard way and is working fine. This PR will need to directly change the Windows Registry which I am uncomfortable about. Also, it is a bit inconsistent to have RDS files in the options but not other file types. 
Therefore I am not sure we need to have this option in R-Instat. What's your view?

@N-thony - In retrospect, I should have made my comments earlier before you worked on this. Sorry for that.

image

N-thony:
> @dannyparsons @rdstern @N-thony I think this PR was triggered by one of @dannyparsons 's comments that I added to issue #6082 (see below). Normally the associations between a file type and an application are managed by the user in the normal Windows settings. This is how it is done in other applications (e.g. Word, Excel etc.). For RDS files and R-Instat, the assocation is already done in this standard way and is working fine. This PR will need to directly change the Windows Registry which I am uncomfortable about. Also, it is a bit inconsistent to have RDS files in the options but not other file types.
> Therefore I am not sure we need to have this option in R-Instat. What's your view?
> 
> @N-thony - In retrospect, I should have made my comments earlier before you worked on this. Sorry for that.
> 
> image

@lloyddewit you are right. Therefore, no need to complete PR #6160 since this association is done by Windows settings.

dannyparsons:
I agree. What is needed is the mechanism for R-Instat to be able to launch with an .RDS file, if the user chooses to open such a file in R-Instat. Has this been added yet?

N-thony:
> I agree. What is needed is the mechanism for R-Instat to be able to launch with an .RDS file, if the user chooses to open such a file in R-Instat. Has this been added yet?

Yes, this solution take account this point. R-Instat is able to launch with an .RDS file.

dannyparsons:
And where is that code which tells R-Instat what to do when launched with an .RDS file?

N-thony:
Line 336 but I think as @lloyddewit said, this is done already by Windows. It adds this association in Computer\HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\FileExts\.RDS\OpenWithList

maxwellfundi:
This seems to be in the review stage. Could we finalize this discussion? @Wycklife  did you gert to test this on your machine?

lloyddewit:
> 
> 
> This seems to be in the review stage. Could we finalize this discussion? @Wycklife did you gert to test this on your machine?

I think there is a strong case for closing this PR. Therefore I suggest that @Wycklife waits for Danny's response before testing.

dannyparsons:
Thanks @lloyddewit that was what I was asking about.

@lloyddewit Could you elaborate on why it might not be sensible for R-Instat to change the Registry directly? I agree we shouldn't automatically change the Registry but I do see other software has similar options e.g. IrfanView, an image viewer, has this:

image

Is it because you may need to be running as an Admin with the right permissions etc? And this may mean we need to trap errors and the process gets complicated? If so, that would be a good argument not to include it.

In terms of other file types, RDS is the only file type used to store the `data_book` so it makes sense to only have this file type association, whereas other file types e.g. csv we only import, but don't manage "natively" if that makes sense.

lloyddewit:
> 
> 
> @lloyddewit Could you elaborate on why it might not be sensible for R-Instat to change the Registry directly? I agree we shouldn't automatically change the Registry but I do see other software has similar options e.g. IrfanView, an image viewer, has this:
> 

Yes, after writing my comments I also thought of other examples where Desktop applications make themselves the default association for certain file types. 
My concerns were mainly about risk. If the Registry gets corrupted then it can be difficult for the user to fix. I see changing the Registry as relatively high risk. 
> Is it because you may need to be running as an Admin with the right permissions etc? And this may mean we need to trap errors and the process gets complicated? If so, that would be a good argument not to include it.
> 

No, I haven't tested it but I think a non-admin user can change their own registry settings.
> In terms of other file types, RDS is the only file type used to store the `data_book` so it makes sense to only have this file type association, whereas other file types e.g. csv we only import, but don't manage "natively" if that makes sense.

Yes, that's a good point.

In summary, if we merge this PR:
- benefits: user has an alternative way to associate RDS files with R-Instat
- costs: risks associated with changing Registry, extra complexity in code, extra testing

I think you and @rdstern are the best qualified to judge the relative priorities. What's your preference?

N-thony:
> > @lloyddewit Could you elaborate on why it might not be sensible for R-Instat to change the Registry directly? I agree we shouldn't automatically change the Registry but I do see other software has similar options e.g. IrfanView, an image viewer, has this:
> 
> Yes, after writing my comments I also thought of other examples where Desktop applications make themselves the default association for certain file types.
> My concerns were mainly about risk. If the Registry gets corrupted then it can be difficult for the user to fix. I see changing the Registry as relatively high risk.
> 
> > Is it because you may need to be running as an Admin with the right permissions etc? And this may mean we need to trap errors and the process gets complicated? If so, that would be a good argument not to include it.
> 
> No, I haven't tested it but I think a non-admin user can change their own registry settings.
> 
> > In terms of other file types, RDS is the only file type used to store the `data_book` so it makes sense to only have this file type association, whereas other file types e.g. csv we only import, but don't manage "natively" if that makes sense.
> 
> Yes, that's a good point.
> 
> In summary, if we merge this PR:
> 
> * benefits: user has an alternative way to associate RDS files with R-Instat
> * costs: risks associated with changing Registry, extra complexity in code, extra testing
> 
> I think you and @rdstern are the best qualified to judge the relative priorities. What's your preference?

@lloyddewit I agree with you about changing the registry. I think this PR needs to wait a bit before it to be merges or maybe not.

dannyparsons:
Thanks both for the discussion. I suggest we delay at least until after version 0.7 is released due to the possible risks.

lloyddewit:
@N-thony This PR was put on hold until after the R-Instat 0.7 release. that's now done so I think we should look at this PR again.
I just read through the comments above. I understand that @dannyparsons' and @rdstern's preference is:
- In the options dialog, allow the user to associate the RDS file type with R-Instat
- Set up the association in the Windows Registry directly from R-Instat
- Ensure that the functionality also works for users without admin privilege

Please could you implement the above?

@dannyparsons @rdstern Please let me know if I misunderstood anything.

Thanks

N-thony:
> @N-thony This PR was put on hold until after the R-Instat 0.7 release. that's now done so I think we should look at this PR again.
> I just read through the comments above. I understand that @dannyparsons' and @rdstern's preference is:
> 
> * In the options dialog, allow the user to associate the RDS file type with R-Instat
> * Set up the association in the Windows Registry directly from R-Instat
> * Ensure that the functionality also works for users without admin privilege
> 
> Please could you implement the above?
> 
> @dannyparsons @rdstern Please let me know if I misunderstood anything.
> 
> Thanks

@lloyddewit I don't know if my mind is still fresh on this but I think we agreed that since the association is done by the OS i.e in Computer\HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\FileExts\, We can then leave this PR. The only issue with R-Instat was not recognizing when tried to open it from a file i.e .RDS or any other. This issue was fixed and merged already. Let me know if I misunderstood something.

N-thony:
@dannyparsons the .RDS file association with R-Instat works well and this feature is in R-Instat version 0.7, so I don't see the need of adding again specific codes to do the association. D you still need this in R-Instat?

rdstern:
@lloyddewit I have finally got the association feature to work in my version (0.7.2) of R-Instat.  I right-clicked on an RDS file and set the association to R-Instat.
a) Then - when R-Instat wasn't open, it went into R-Instat and did essentially a File > Open on that RDS file, which was then read into R-Instat.  I was pretty happy.
b) The next step was to see what happened when R-Instat was already open and I clicked on another file.  This time it again went to the File > Open dialogue and added more data frames.

So I am with @N-thony in being ok with all this.  It is different to Excel, which opens multiple instances of the software.  But still ok.

I would therefore have no objection if we now simply closed this pull request.

lloyddewit:
@N-thony , @rdstern Thank you for the extra comments. If there is no requirement to set up the RDS association from **within** R-Instat, then I am also very happy to close this PR.

PR #6324, opened 2021-03-23
Updating list of geoms

### LINKED ISSUE #6293: Improve the graphics in R-Instat through the addition of more geoms
My claim is that we would substantially enhance the graphics capabilities in R-instat if we had more geoms.  I also had the impression that @maxwellfundi became quite a whizz at doing this efficiently.  If so, then perhaps this is also something that @Karakacha and @TobbyKe could take over, with initial instruction from @maxwellfundi?

Now the list is getting long and later perhaps we should worry about that!  For now I suggest the additions are far more important than adapting that control

On that supposition here is my initial list to add:

a) The remaining geoms in ggplot2.  I suggest we should have them all.  And there are some that sound really useful!  I have put the obviously important ones in bold below.
b) The geoms in ggridges as mentioned in #6278 
c) geoms and some stats from ggpmisc see  #5744 and #5037

From ggplot2
area
bin2d
blank
crossbar
errorbar
errorbarh
**function**
hex
linerange
path
**qq**
qq_line
quantile
**raster**
rect
**spoke**
steps

If you are able to proceed, then I suggest you do a number of pull requests, because there are quite some enhancements here.  Not one for each geom, but perhaps 4 or 5 overall with suitable names.  for example geom_spoke seems to be useful for circular data.

maxwellfundi:
Fixes #6293


maxwellfundi:
I have also now added initial `stat_density_ridges` 


image

rdstern:
@great!

rdstern (changes requested):
@maxwellfundi this is a tour-de-force!  I have started the checking, but it will take more time than is possible just now.
I found one problem in an important geom, namely rect needs for eastetics, namely xmin, xmax, ymin and ymax
The initial job (I suggest is simply to go through each geom quickly and check - as I did for about 5, that it gives something, not an error, it you complete all the aesthetics you put in bold and produces something - can be nonsense but no error.  I tried adding to a simple scatter plot with the usual survey data on 5 geoms and rect was the only error I got.  Do you think John can help as well?
I can also continue tonight and tomorrow, but not just now.

maxwellfundi:
@dannyparsons  we have a fill aesthetic we can add a column from the data to produce a plot like this one 

image

We can also have the fill as the different quantiles as shown below with` fill = factor(stat(quantile))` ( Am not sure how we can add this on the receiver, can we make it editable?

image

maxwellfundi:
Yes @rdstern  its exactly the job have given to @Karakacha  to check them one by one to identify any oddities or errors.

maxwellfundi:
@rdstern  I have now corrected the issues with rect and other geoms, thisi s ready for re-review

Karakacha (approved):
I tested all the geoms added, fixed all the bugs and now they are all working.

lloyddewit (approved):

rdstern (approved):
Great.  @maxwellfundi What an undertaking.  Now we need to try them out.

PR #6333, opened 2021-03-24
Fixing the issue of multiple output in the duplicates dialog 

lilyclements:

I noticed a very minor bug in the Duplicates dialog where if you clicked reset and clicked "OK", then the `summary` code was run multiple times. This PR fixes this so that this should not be an issue.

To test:
* Load data
* Open dialog (Prepare > Check Data > Duplicates)
* Click "Reset"
* Click "OK"
* The `summary` output should now only be printed once


dannyparsons (approved):
Code looks correct. A reviewer should also check functionality.

instat/dlgDuplicates.vb
```diff
@@ -137,8 +137,6 @@ Public Class dlgDuplicates
         clsStreakFunction = New RFunction
         clsSubsetCol = New RFunction
         clsDupCountIndex = New RFunction
-        clsGetColumnsFunction = New RFunction
-        clsSummaryFunction = New RFunction
```
shadrackkibet, line 141: Do you have `AddToAfterCodes/AddToBeforeCodes` somewhere? I think the correct way to fix this is by clearing codes i.e `   ucrBase.clsRsyntax.ClearCodes()` in `SetDefaults()`.

dannyparsons: Thanks Shadrack, I read this differently at first, I thought this was defining the RFunctions again. Yes, these lines should stay here and there should be a clearing of codes as Shadrack mentioned, which should solve the issue.

dannyparsons: An alternative would be to do `clsSummaryFunction.Clear()`. But I think I prefer explicitly clearing the codes as Shadrack mentioned.

lilyclements: Thanks for this. I have sorted it now and so this is ready for a re-review

rdstern (approved):

shadrackkibet (approved):
Looks good. I therefore approve.

PR #6334, opened 2021-03-24
Pasting into columns from clipboard

### LINKED ISSUE #5619: Adding a simple paste into a data frame

Patowhiz:
Fixes #5619

Please not this PR replaces PR #5982 but does all that is explained in that PR.

Pasting into existing columns is only supported if the data is of pasted is of the same length as the data frame, and the pasting the column header is not supported.


rdstern (dismissed):
@Patowhiz I am also approving in case my suggestion is not trivial.  
It warns me that my change cannot be undone.  I am usually happy with that.  But I added 2 new variables, filled with NA and pasted into them.  In that case I don't need a warning.  
As I said - if this is very quick to do, then please do.  If it takes a reasonable time, then let's make an issue and merge this now.  If it sounds really tricky, then ignore the suggestion.

Patowhiz:
@dannyparsons what's you suggestion on the above @rdstern proposal, we could run an R command every time there is a paste that checks on whether the column has NA's only and only show the message box if it has other values as well.

dannyparsons:
Ok I'll have a look.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -850,6 +850,62 @@ DataSheet$set("public", "replace_value_in_data", function(col_names, rows, old_v
...
+               "Current data frame rows:", nrow(current_tbl), ". Copied rows:", nrow(clip_tbl)) )
+  } 
+  
+  #check copied data integrity
+  for(index in seq_along(col_names)){
+    col_data <- current_tbl[, col_names[index]]
+    #get column type of column from the current table using column name
+    col_type <- class(col_data)
+    #check copied data integrity based on the data type expected
+    if (col_type == "factor" || col_type == "ordered" ){
```
dannyparsons, line 877: This is preferred way to check for a factor.
```suggestion
    if (is.factor(col_data)) {
```

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -850,6 +850,62 @@ DataSheet$set("public", "replace_value_in_data", function(col_names, rows, old_v
...
+      #get all the factor levels of the selected column in the current data frame
+      expected_factor_levels <- levels(col_data)
+      #check if all copied data values are contained in the factor levels
+      #if any invalid is found. exit function
+      for(val in clip_tbl[,index]){
+        if(!is.na(val) && !is.element(val,expected_factor_levels)){
+          stop("Invalid column values. Level not found in factor")
+        }
+      }#end inner for loop
+    }else if ( !(col_type == "numeric" || col_type == "integer" || col_type == "character" || col_type == "logical" ) ){
```
dannyparsons, line 887: Preferred way to check types.
```suggestion
    } else if ( !(is.numeric(col_data) || is.logical(col_data) || is.character(col_data)) {
```

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -850,6 +850,62 @@ DataSheet$set("public", "replace_value_in_data", function(col_names, rows, old_v
...
+      #clipr support above column types only. So pasting to a column not recognised by clipr may result to unpredictible results 
+      #if not in any of above column types then exit function
+      stop( paste("developer error. unknown column type:", col_type) )
```
dannyparsons, line 890:
```suggestion
      stop( paste("Cannot paste into columns of type:", col_type) )
```

dannyparsons (changes requested):
Looks good, works well, just a couple of small suggestions.

I don't think it's worth handling the very special case of pasting into a column that's entirely blank.

Patowhiz:
Due to new conflicts caused by additions to `ucrDataview`, this PR has a bug that requires me to take time and trace it. I may create another PR and redo the changes to it.

dannyparsons:
Sorry about that, I think other PRs on these files were merged recently. Hopefully it's not much work to resolve. It was generally looking good.

Patowhiz:
This PR has now been replaced by PR #6459, so I'm closing it.

PR #6338, opened 2021-03-25
Added new dialog for finding points in a polygon

### LINKED ISSUE #5422: Finding points in a polygon?

HawardKetoyoMsatsi:
Fixes #5422
@maxwellfundi , @rdstern @dannyparsons @shadrackkibet @lloyddewit  This is ready for review kindly take a look.

### DIFF
--- a/instat/dlgLocatingPointsInShapeFile.Designer.vb
+++ b/instat/dlgLocatingPointsInShapeFile.Designer.vb
@@ -22,31 +22,180 @@ Partial Class dlgLocatingPointsInShapeFile
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()>
     Private Sub InitializeComponent()
+        Me.grpShapeaFile = New System.Windows.Forms.GroupBox()
+        Me.lblGeometry = New System.Windows.Forms.Label()
+        Me.grpStationFile = New System.Windows.Forms.GroupBox()
+        Me.lblLongitude = New System.Windows.Forms.Label()
+        Me.lblLatitude = New System.Windows.Forms.Label()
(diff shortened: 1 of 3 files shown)


HawardKetoyoMsatsi:
@rdstern the changes e agreed upon have been made, kindly take a look.

rdstern (approved):
Great to see this working

shadrackkibet (approved):

lloyddewit (approved):

shadrackkibet:
@HawardKetoyoMsatsi This dialog needs to be now enabled from the frmmain.

shadrackkibet:
Note that this dialog currently works with only `level-0` shapefiles i.e boundary. It is easy to enable other levels to work, we just need to use `sf::st_combine()` which will combine several feature geometries into one.  @rdstern would you like this addition?

PR #6157, opened 2021-01-13
Improving Taylor Diagram Dialog

N-thony:
Improved the design of the Taylor Diagram dialog and small changes in the code. This fixes the comments in issues #5885 
@africanmathsinitiative/developers this is ready to review. Thanks.

PR #6339, opened 2021-03-25
Improved labeling of controls in the taylor diagram dialog 

### LINKED ISSUE #5885: Taylor diagram improvements
Related to Issue #4944. The initial implementation of  Taylor diagrams was done in PR #5867. This PR was merged but some small issues remain (see comments from @rdstern below). 

Minor points: Label should be Estimated, and not Estimate.
I wonder about changing both the position and label for the first receiver. Perhaps it could become Facet (Station): and also perhaps moved to be below the Estimated receiver? Then it is just above the Group control.
I suggest we may later only have a single receiver for the Groups. There can be more than one factor here, but the legend is only for the first. But this should not be changed now, so the use of the multiple receiver can be tested.
I have also been looking at the documentation for the Taylor diagram. There is a lot more we can add to this dialogue, including saving the results - possibly as an object or in a new data frame. And options for colours, etc. Also I notice there can be 2 facet variables. I wonder if it might be useful to have a receiver that takes up to 2 variables?

shadrackkibet:
fixes #5885
 
@africanmathsinitiative/developers this is ready for review. It is a continuation of PR #6157. 

- Minor designer changes
- Changed labels


maxwellfundi (approved):
This looks good! @lloyddewit @rdstern   this is ready for your review.

rdstern (changes requested):
Openair uses the term Type for what we call a facet.  So I would prefer that label to be Facet (Optional):  And Group could become Group (Optional):
I suggest @dannyparsons to have a quick look, before these minor changes, just to check if there is anything else.

dannyparsons (approved):
Looks good

shadrackkibet:
Danny did approve these changes. @rdstern are you happy to approve as well?

dannyparsons:
Yes, the code looks good, and I liked Roger's suggestions on the labels, though I didn't actually test out the dialog.

shadrackkibet:
Now I am confused. Should I change the labels to **Facet (Optional):** and **Group (Optional):**?

rdstern:
I would like that!

shadrackkibet:
Done!

image

rdstern (approved):

dannyparsons (approved):
Looks good. I think we might be able to improve on this when we know more about Taylor Diagrams, but this is now not a confusing design so definitely an improvement.

PR #6344, opened 2021-03-25
Graphs no longer saved in data_book by default to reduce file size when saving/auto-saving

### LINKED ISSUE #6158: Sorting out the graphics in R-Instat
This issue builds on #5996 and then in particular also on #6146.  They should be closed once this is solved.

We are suggesting that @Ivanluv take over from @Patowhiz on #5996, because some changes are needed, resulting from fixing #6146 that are deep in the R-Instat code.  Discussions will involve @dannyparsons , @volloholic and @lloyddewit both on strategy and implementation.

Currently graph objects are saved automatically as part of the appropriate data sheet.  They may be called last_graph if not given a particular name, or may be named.  (So, when there are multiple data sheets in  the R-Instat data_book there may be multiple last_graphs.)

We now want to make this saving of graphs optional, because they can result in a data file becoming large.  (The extreme example I found was a single graph from a modest dataset being over 200mbytes.)

This is paying quite a high price for the luxury of being able to recall graphs over different sessions of using R-Instat - something I have never needed to use.  We would still like this option, and it ties in with the idea of being able to save the current session of work (offered by other statistics packages, which we hope to add.)  But it must be more under our control.

I also find it a plus that R is efficient in its file sizes - and often about 10-15% of the size in Excel.  This is often useful when sharing data and a shame to sacrifice completely.

David and Danny have proposed there be an option - presumably in Tools > Options > Graphics on where to save the graphs.  The default is in the Environment and the alternative is perhaps as we have it now.  Once implemented, then I assume there would be no visible difference in the current R-Instat session?  We may also want a new "Save Graphs As" in the File > Save As menu? 

This will also affect the toolbar feature, which currently works for some graphs and not others.  So those changes need to wait for these alterations.  

This will also affect the Describe > View Graph dialogue.  This currently only works for ggplot graphs.  We are now producing graphs using all 3 systems, namely ggplot2, lattice and base graphs.  I think the toolbar now works for lattice graphs as well as ggplot graphs, but the dialogue does not.  The base graphs adds another "feature" discussed in #5996 that the current R-Instat code - adapted from ggplot2 code, doesn't work, because it doesn't produce an object.  It needs to change - in each instance - as described there, to produce an object.  Currently I think is is just in Model > One Variable, when using the fitdistrplus package, or in extremes, when using the extRemes package or when using the circular package.  These are not big changes.

dannyparsons:
fixes #6158

Implements the workaround outlined here: https://github.com/africanmathsinitiative/R-Instat/issues/6158#issuecomment-806906835


rdstern (changes requested):
@dannyparsons releif that that problem is solved.
But I don't have a graph saved at all. 
I did a Taylor plot, and then gave the graph a name.  Then a correlation with a graph.  So one ggplot2 and one lattice graph.
When I look at the objects I have last_model, but not last_graph, nor the graph I saved.  I thought I would have access to them for the session?

dannyparsons:
Thanks. They were there but weren't being displayed. Should be fixed now.

I tested saving a file. Closing R-Instat and opening. When reopening, graphs are gone but models stayed.

dannyparsons:
In case you're looking again already, there was a small bug, I've just fixed with latest commit.

rdstern (dismissed):
@dannyparsons this is great, and I hope it can be merged.
It has uncovered some (minor) problems in the Prepare > R Object dialogues.  There are 4 dialogues and graphs are not recognised as Objects in View object or Reorder Objects.  In rename objects and Delete objects it is different, because of Shadrack's work on the filters.  That has Objects and one category is Model.  A model appears twice.  Graphs appear there, but not under the general Object category.  I wonder if we need that, and perhaps should just have the categories.

Then Describe > View graph now includes the lattice graphs, even though it says it is just ggplot only.  Of course it can't display in the interactive viewer!
These are nothing to do with this issue, that I hope can be merged quickly.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -499,59 +499,85 @@ DataBook$set("public", "get_columns_from_data", function(data_name, col_names, f
...
+    if (!is.null(out)) return(out)
   }
   else {
-    if(data_name == overall_label) {
-      curr_objects = private$.objects[self$get_object_names(data_name = data_name, type = type)]
+    #TODO implement force_as_list in all cases
+    if(missing(data_name)) {
       if(!missing(object_name)) {
+        curr_objects = private$.objects[self$get_object_names(data_name = overall_label, type = type)]
         if(!(object_name %in% names(curr_objects))) stop(object_name, "not found.")
```
lloyddewit, line 541:
```suggestion
        if(!(object_name %in% names(curr_objects))) stop(object_name, " not found.")
```

dannyparsons:
@lloyddewit This may conflict with others PRs on the same R files. Since this is a small change I don't mind waiting for those to be merged first and then resolving any conflicts.

lloyddewit (approved):
I reviewed the R code as best I could.
I just have one small comment (feel free to ignore this comment)

dannyparsons:
Thanks, updated.

lloyddewit (approved):

dannyparsons:
@lloyddewit I'll wait till #6346 is merged before merging this in case of any conflicts.

PR #6346, opened 2021-03-26
New circular density plots dialog: Density plot, Rose plot, and Scatter plot & fixed saving of non-ggplot graphs

dannyparsons:
this replaces #6262 with a small addition. This R function is edited in order to save non-ggplot graphs correctly:

```r
# This function creates a wrapper around grDevices::recordPlot() to enable non-ggplot graphs to be saved as recorded_plot objects.
# It also handles graphics devices carefully.
record_graph <- function(x) {
  # store current device, which could be png
  d <- dev.cur()
  # create a new device, to ensure graph is displayed in device that recordPlot() can capture
  dev.new()
  # store the new device, so it can be turned off later
  d2 <- dev.cur()
  # Display graph to be captured by recordPlot()
  x
  y <- grDevices::recordPlot()
  # set device back to current, to ensure code after this run correctly
  dev.set(d)
  # turn off the new graphics device
  dev.off(which = d2)
  return(y)
}
```


dannyparsons:
@africanmathsinitiative/developers This is ready for review. The only additional change to #6262 is the saving of non-ggplot graph objects. So this is all that needs to be tested.

shadrackkibet (commented):
I tested it on my machine and now the last graph works. The only issue I got is that when I click Ok in the dialog I get the empty (black) graphic window and then it disappears.

dannyparsons:
Thanks, yes forgot to mention that now the graph will "flash" on screen, this is so that we can capture it with recordPlot. This will only be for non-ggplot graphs. Possibly by further investigation into graph devices we can avoid this but I think this is ok for now as a way of allowing these graphs to be saved.

shadrackkibet (approved):

rdstern (approved):
@shadrackkibet @dannyparsons I tried with a scatterplot from mydata (65000 rows) and it isn't a flash then.  It is building the graph in the R-Viewer and then (I presume) transferring it to the output window.  Quite nice!

dannyparsons:
That's not quite what it's doing but it sounds more logical so I'm happy for you to tell people that!

Patowhiz:
This is a nice addition to the basic implementation we had (I've learnt something). @dannyparsons I had though the "R screen devices" are just virtual devices. I'm surprised there will be flashes.

PR #6347, opened 2021-03-26
Graphs no longer saved in data_book by default to reduce file size when saving/auto-saving

dannyparsons:
Same changes as #6344 redone due to conflicts.


dannyparsons:
Merging as changes approved on #6344

PR #6351, opened 2021-03-27
New dialog for grid data entry of daily climatic data

dannyparsons:
replaces #6301 with some functional and code changes

I slightly changed how the day/month/range option is done. I think it will be clear from trying out the dialog. Basically, the end date is always visible but fixed and disabled for day/month. Month is forced to start at the first of the month, that could easily be changed.

I stopped it autofilling all elements to edit - I thought it's better to choose the ones you want, but can change this back if preferred.

I've done something on reopening the subdialog. It isn't perfect but I think sufficient for a first attempt. If you make any changes on the main dialog then it warns you that you'll lose your changes if you continue to reopen the subdialog. If you haven't changed anything on the main dialog and reopen the subdialog you can get back your changes to edit. Not perfect but anything more complex will need more work.

@rdstern This doesn't have all the features, but I suggest this is a sufficient minimum working dialog. 
Given the time taken to get to this point and the other urgent tasks needed for the next version, I suggest this is tested for bugs and merged if working well. Other changes can be started straight after, but won't hold up the next release.

@N-thony @Patowhiz I've made a number of significant code changes here, particularly to how the subdialog works. In general, the RFunctions should be passed through to subdialog and controls linked to them - we should avoid having to make a direct call to something on the main/subdialog from the other one - this can get very messy as subdialogs can be used by other main dialogs, and may not always be in sync. It's better that checks are done directly on the RFunctions and code where possible. This has changed the code quite a lot. I'm happy to discuss and go through it with you to explain.


rdstern:
This is looking good.
a) The date now is properly formatted to view on the right, as well as on the left.  But when included to edit, it is in numeric format.
b) I like the way it tells me that I will lose what I had before.  It queries this whether I have made a change in the data or not.  For example I pressed reset and it still asked.  That's not a problem, but nice if it doesn't bother me unless there is data to lose.
c) I caused a problem by entering non-numeric values.  Then when I tried to go back into the data entry it was unhappy.  I had 00 in one cell - easy to do.  And then kkk in the next.  I suggest, for now perhaps you restrict entry to **_numeric_** to avoid the problem (Not sure what happens with editing the dates then, but that's another thing that could be left perhaps for now?)  Or for now you leave the editing of dates as the numeric values they are now!
d) The one improvement I would like is the Tab stuff.  If I enter tmax and tmin and rain, then I hope the focus could be on the first value of tmax, and tab would go to tmin, rain and back to tmax.  That's how to do entry going across.  Currently Tab still goes to all cells.  I hope that's easy to add?  If not, then that's also after merging this version.  
I suggest it could then be merged.  I still need to check on saving though.

dannyparsons:
Thanks. Made a few small changes:

- month is the default option
- element columns can only be strict numeric (no dates)
- data entry only allows numeric values
- message display corrected after running dialog

Tab functionality was more complex sorry, so I suggest that's for a later improvement.

This is ready for re-review.

rdstern (approved):

Issue #3721, opened 2017-07-20
Initial multilingual translation


volloholic:
Establishment of the multi lingual functionality and translation into at least one other language. (Probably French)

Ogik99:
#4176 
#5096

dannyparsons:
We have now decided on the approach as suggested also for Climsoft here https://github.com/opencdms/Climsoft/issues/200#issuecomment-474824472.
This will be simpler to implement in Climsoft and so will be implemented there first before being used in R-Instat.

isedwards:
@dannyparsons - There are three commits in the [Climsoft/#592](https://github.com/climsoft/Climsoft/pull/592/commits) pull request for adding language translation.

- [bb1d840](https://github.com/climsoft/Climsoft/pull/592/commits/bb1d84071d95a9290629a4d499b083b3a2eddfd3) adds the command `SimpleTranslateTool.translateForm(Me)` to every form. You can ignore this commit. Hopefully in R-Instat all forms may inherit from a single parent base form, in which case you would only have to add this command in one place and all forms would then inherit the code.
- [7f3a803](https://github.com/climsoft/Climsoft/pull/592/commits/7f3a803790faba850b1175e0f6c75d01c22c389b) can also be ignored. It adds the binary SQLite file that contains translations for the Climsoft interface. R-Instat will need an equivalent file.
- [93f2ea7](https://github.com/climsoft/Climsoft/pull/592/commits/93f2ea7ad4f5ab67ac5b4bb94f608072a3490e7a) this commit contains [clsSimpleTranslateTool.vb ](https://github.com/climsoft/Climsoft/blob/93f2ea7ad4f5ab67ac5b4bb94f608072a3490e7a/ClimsoftVer4/ClimsoftVer4/clsSimpleTranslateTool.vb). You should be able to reuse some of this code. The majority of other changes in this commit are removing an older attempt at language translation from Climsoft (some of which is just hardcoded).


There are currently three known problems with [clsSimpleTranslateTool.vb ](https://github.com/climsoft/Climsoft/blob/93f2ea7ad4f5ab67ac5b4bb94f608072a3490e7a/ClimsoftVer4/ClimsoftVer4/clsSimpleTranslateTool.vb).

[...]

PR #6343, opened 2021-03-25
Added option to display menus in French

lloyddewit:
Partially fixes issue #3721 

Note: The translation mechanism uses the TranslateWinForms library. I uploaded this library to NuGet so that it can be integrated into R-Instat in the standard way. The source code for this library can be found at:
https://github.com/lloyddewit/TranslateWinForm
We can move the library to the `africanmathsinitiative` GitHub site when it is stable.

PR #6353, opened 2021-03-28
Added option to display menus in French.

lloyddewit:
Partially fixes issue #3721.
This PR replaces (and closes) PR #6343.

@dannyparsons Please could you review?

Note: The translation mechanism uses the TranslateWinForms library. I uploaded this library to NuGet so that it can be integrated into R-Instat in the standard way. The source code for this library can be found at:
https://github.com/lloyddewit/TranslateWinForm
We can move the library to the `africanmathsinitiative` GitHub site when it is stable.


dannyparsons (approved):
This works well. The first time I ran it complained about missing SQLite.interop.dll

After installing and uninstalling some packages it then worked. So there's a possible problem but not sure.

@rdstern or @shadrackkibet could you try out the branch and see if it works?

rdstern:
I get the error you mentioned too, namely:

image

Do I have to load that dll.  If so, then where?

dannyparsons:
Can you extract these two folders to `instat/bin/Debug` [dlls.zip](https://github.com/africanmathsinitiative/R-Instat/files/6223592/dlls.zip)

We're not yet sure why they don't appear automatically, but this should work for testing.

shadrackkibet:
- First I got the same warning message.
image
- I extracted the files into the directory as shown above and the menu translation seems to work!

rdstern (approved):
Great

lloyddewit:
I understand from @dannyparsons that if we make sure that `SQLite.Interop.dll` is copied in the bin folder when compiling the installer, then the R-Instat deliverable version should work fine for users.
Therefore I suggest that we merge this PR.

In parallel, we should also look for cleaner ways to manage packages within the R-Instat project. I raised issue #6356 for this.

PR #6357, opened 2021-03-30
Improving check station location dialog

shadrackkibet:
No open issue that this PR fixes. It is based on a come here https://github.com/africanmathsinitiative/R-Instat/pull/6338#issuecomment-808866906 and discussions with @rdstern elsewhere.

- Use of any level shapefile
- Omitting missing values


rdstern (commented):
@shadrack  this works and is am important improvement in handling the shape files at any level.  
I assume you also want to add the optional feature to save the sub-county, etc as well as the logical variable.  I assume both would become logical, as the station file may not have the information - so the logical becomes less useful.  

If this enhancement is to be left for the next release, then I will approve where this has got to.  Otherwise I'll wait

rdstern (approved):
There is still work to be done on this feature, but it is already useful.

dannyparsons (approved):

PR #6362, opened 2021-03-30
New modified import multiple files

### LINKED ISSUE #5789: input permitting multiple files
We currently use the rio::import to import a single file and rio::import_list to import multiple Excel sheets from an Excel book.
We quite often get multiple files to import, where there is a single file for each climatic station or for each element, or both.  Currently we have to import these files one-at a time.
This extension is to be able to import multiple files.
a) In the browse we simply choose more than one file.  I assume they would all have the same extension and we may wish to check this?
b) In the dialogue, this could result in a similar display we currently have for an Excel book.  But we don't need all the selection - we did that when we chose the files.  So we might have a listview control to see the list of the files.
c) Unless it is easy we would not have a preview.  If it is easy, then we might show a preview of the first file.
There are all sorts of ways we could spend a lot of time and make this complicated.  Let's keep it as simple as possible!

Patowhiz:
Fixes #5789. An alternative to PR #6355

Implements below @rdstern comment.
> @Patowhiz brilliant - it works! So now I assume it is cosmetic. And your current system didn't make it easy for me! I was reading the 6 files from Zambia. for an element. OK so here is what I had to do.
> a) There is still the developer error going in. Is that just the first time.
> b) Then it doesn't go to the dialogue to let me choose Single or Multiple - not the first time anyway. It takes me to the choice of a file. On "my proposal" that's fine, but on yours that step is a waste of time.
> c) Then I get to the dialogue and choose Multiple. Now I am thrown back to the very beginning (each time) and have to find the directory.
> d) When I find the directory it gives me the choice of 3 file types/extensions. Mine is DLY. So it doesn't find any files!
> e) I change it to txt for 4 of them and csv for 2 - just for the heck of it!
> f) Then it reads either way on. All 6 files.
> 
> So, how about "my system".
> 
> a) You start - always - with a sample file. It is a good start if it can read it.
> b) Most of the time we are still going to be reading single files - so that's a nice start.
> c) There is a checkbox somewhere towards the top saying Multiple Files. It is disabled initially.
> d) It has a tooltip that says "All files in a directory can be input if they are all of certain types, e.g. csv, txt, sav"
> e) Now you sort out the settings, e.g. missing values, and so on with this file.
> f) You now tick on multiple files. It now gives you a message in that box - I first got No files found. Then if 2 files found it gives the details just as you do now?
> 
> g) now I must try with sav for spss! That would be great!


rdstern (dismissed):
@Patowhiz great.  This new release is becoming quite something!

instat/dlgImportDataset.vb
```diff
@@ -128,6 +135,10 @@ Public Class dlgImportDataset
         ucrInputFilePath.SetParameter(New RParameter("file"))
         ucrInputFilePath.IsReadOnly = True
 
+        ucrChkMultipleFiles.SetText("Multiple Files")
```
dannyparsons, line 138: Checkbox text might need to be on two lines for this?
```suggestion
        ucrChkMultipleFiles.SetText("All files of this type")
```

dannyparsons (commented):
This looks good. I have two small suggestions for now and then some for later.

- Change checkbox text as below
- In the file preview if it's easy can you change it to only show the file names? We don't need the full path for each file, plus the path is shown above anyway.

Patowhiz:
@dannyparsons I have made the changes you requested above.

dannyparsons (approved):

dannyparsons:
Thanks, looks good. Further changes were small so merging.

PR #6363, opened 2021-03-30
Added facility to summarise sub-daily (hourly, synoptic, AWS) data to daily level

dannyparsons:

There's no specific issue. It adds a "Daily" button to the Climatic Summaries dialog in order to summarise sub-daily data to the daily level by summarising by date.


rdstern (approved):
@dannyparsons this is great, so approved.  It has set us on a track we need to follow though.  This dialogue assumes a doy variable.  The dialogue that easily produces a doy variable only works currently with a Date variable, not a Date-time variable.  

I don't think any steps are difficult here, but we should take a few more soon - in the next couple of months?

lloyddewit:
@dannyparsons I did not see @shadrackkibet 'c omment when I approved and merged (either my mistake or the PR was not refreshed). Do you wish to address his comment?

dannyparsons:
It's sort of an existing issue but it should be fixed. I suggest another developer could take it up.

lloyddewit:
> 
> 
> It's sort of an existing issue but it should be fixed. I suggest another developer could take it up.

Raise issue #6366

PR #6369, opened 2021-03-31
Improved declustering option in Extremes dialog

dannyparsons:
replaces #6368 with minor changes:

- removed second threshold
- moved position of declustering checkbox

@rdstern ready for review


rdstern (approved):
@dannyparsons great.  That's very nice and neat.

It also gives a good impression that it is filtering the declustered values.  We should be able to do that in a future release before anyone knows that it wasn't always there.

What's there now is a feature rather than a problem!

Issue #6352, opened 2021-03-28
Adding facilities to the current grid data entry system


rdstern:
This issue builds on the basic data entry system merged into Version 0.7.0.  It adds items that are needed before implementation, though testing can start now on the basic system.

It is possible that there will be time to merge some of the facilities, but others will wait for Release 0.7.1 - possibly mid-April.  And some will be for Version 0.7.2, hopefully by late May.  I have indicated my hope on the completion date with the issue.

In the list below (1) is for the most urgent, so aim for this release, but nothing is essential.  (2) would be nice in this release, but would probably wait for mid-April.  (3) is not considered now and may not be until release 7.2.  

a) (1) **Entry by day**:  Check that entry by day (rather than by element) is available.  The feature needed, is for at least one key to advance horizontally (Tab does now) but also only include the elements being entered.   So on the next row it omits the Station and Date fields and goes straight to the first element to be entered.

b) (1) **Menu item**. Add Daily and ... if not already there.  So it reads `Daily Data Entry...`  This may already have been done.

c)  (1) Add a `Transform: `checkbox. If checked it opens a pull-down into which you can type. Initial default is 10 and then Inch to mm and F to C. Perhaps just have the 10 there for now.  This is needed for the saving.  So the Transformation is applied when writing back to the data.

d) (1) Add a **_Transform button_** in the grid.  It is disabled for now and there is a tooltip.  "When implemented, this is an option to show the transformed data.)

e) (2) or (3) depending how easy it is.  Implement the Transform button.  It is disabled unless the Transform checkbox is used.  If enabled, it shows the data, e.g. 134, becomes 13.4. The grid is made read only and Transform is clicked again to be able to resume entry.      

f) (1) `Default Value` checkbox. Default unchecked. If checked you can type any number - default is 0. That changes all missing values to 0 in that variable.  Non-missing values are left unchanged.  There is another checkbox afterwards, with the label For Precip Only.  Default is unchecked.  It is disabled if the data are not defined as climatic.  Add a tooptip "The data must be defined as climatic to recognise which variable is precipitation."

g) (1) **_Restrict Entry:_** Add a group box with `Restrict Entry` as the label.  Inside there are currently just 2 checkboxes for now:  Default is unchecked for both. The first is `No Decimal`.  The second is `Allow t for Trace`.  The first is implemented now. You still allow minus, but not "." (So a temperature of 13.4 is entered as 134.  This links to the transform mentioned in c) 

[...]

PR #6365, opened 2021-03-31
Added facilities in the climatic data entry grid system

N-thony:
@Patowhiz this ready. Could you test and proceed with the next? Thanks.

PR #6370, opened 2021-03-31
Added functionality of entering Trace, default value, no decimal in climatic data entry

Patowhiz:
Fixes issue #6352 and its a continuation of PR #6365.

@africanmathsinitiative/developers @N-thony  this is ready for review


rdstern (approved):

rdstern (approved):
@Patowhiz great.  I also now saved values repeatedly and couldn't break it.
The only minor bug (in my view anyway) is that the Return button doesn't register the value in the last cell.  You have to exit the cell first.  At (say) the end of the month, users might not always do that.  Not serious enough to need to fix now.

Patowhiz:
@rdstern that's cause of the reogrid delay in raising the event before the dialog closes. Unfortunately most grida tend to behave like that but let me have a look into it.

lloyddewit:

dannyparsons (approved):

PR #6381, opened 2021-04-04
Bug fix: use of multiple factors on fill date gaps

### LINKED ISSUE #6380: Small improvement in the Climatic > Dates > Infill Missing dates dialogue when there are multiple station variables?
This dialogue has a multiple receiver for the Station information.  One use could be if there are stations within countries I suppose.  I am not sure how we would cope with that elsewhere.
 However, one common situation is when there are multiple Station variables, e.g. name and ID
[germany_4stations.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/6253999/germany_4stations.xlsx)

It would be nice to be able to put them all in to the receiver and they would be infilled.  One easy way would be to just use the first variable for the infilling and then use the facility to replace missing values by the value above as we do in the Replace Values dialogue.

shadrackkibet:
fixes #6380


rdstern (commented):
@shadrackkibet great.  It works.
However, it now works too well in fact.  
I tried also with a Month factor that I generated with the Use Date dialogue.  It still worked, but added many other rows than I wanted.  Can we restrict it to only work for additional station-type variables.  I assume as a factor they need to have the same number of levels and really be in 1-to 1 correspondence?

I have now checked on one possibility.  I took the 2 factors, each with 4 levels and used interaction as follows,i.e. with drop unused levels = TRUE:

`Interact <- interaction(x=data_book$get_columns_from_data(data_name="Append", col_names=c("station","station_id"), use_current_filter=FALSE), drop=TRUE)
`
The resulting interaction only had 4 levels, i.e. the same number as the first station variable.  Unless that is the case, we don't allow additional variables?

shadrackkibet:
@rdstern I added a check for the factor levels please test again.

rdstern (changes requested):
@shadrackkibet I think the check could be a bit cleverer than just the number of levels of each factor.  That's why I suggested we use the combine idea and get rid of unused levels.  Then the combined number of levels will be the same as the station factor.

lloyddewit:
@shadrackkibet Please could you clarify the status of this PR? thanks

shadrackkibet:
No advances on this yet. I am not sure how to do this kind of check.

shadrackkibet:
I also think the current design of the dialog is a limitation in doing this check. @rdstern if you agree we have the station factor separately then I think it is possible to implement your idea i.e add a station receiver.

rdstern:
I think it is easy to do a check with the current structure.  I use the example of the Kenya data from the library, for all elements where there are 6 stations.  There are 2 variables for the station as shown here:

image

Now these two variables need to be essentially the same.  I got an interaction variable for the 2 variables.  Here is the r-code from the interaction dialogue:

```
# Code generated by the dialog, Combine Factors
Interact <- interaction(x=data_book$get_columns_from_data(data_name="w_kenya_allelements", col_names=c("station","stationID"), use_current_filter=FALSE), drop=TRUE, sep="Période."):

```

This produces an interaction variable.  (If there are 3 variables, then all 3 could be included.  Then we simply check if the number of levels of this interaction is the same as the number of levels of the first variable in the control.  If ok we proceed.  If not we stop and say something like:  **_The multiple Station Variables are not in synch_** 

It would be clearer if we changed the label from `By Factors` to` Station Variable(s):`

shadrackkibet:
- I don't like the fact that we have to depend on the first variable in the multiple selector. What if the order of items in the receiver changes? 
-  This dialog is available under **Prepare>Column Dates**. Therefore it can be used generally although its main use is on climatic data. Changing **By Factors:** to **Station Variable(s):** will look like it is just climatic?

rdstern:
On your first comment, it only works if all variables are essentially identical - so any variable will do.
I take your second point.  I hadn't thought of that.  perhaps leave the label as it is.

shadrackkibet:
@rdstern I implemented the desired check. Please check.

rdstern:
@shadrackkibet many thanks - will do

rdstern (dismissed):
@shadrackkibet this seems to work well, so I am approving.  @lloyddewit to do the second check now?
@shadrackkibet when there is an error it is raised in the same way as errors in our code - that we haven't planned for.  This is a general point.  When we have something like this, I have no problem reporting it, in this way.  But I wonder if we could also give a neater version of the message in the output window?  We don't need to repeat the command there, because that's (usually) already displayed?

lloyddewit:

N-thony (commented):
@lloyddewit all is good from my end. Thanks

instat/dlgInfill.vb
```diff
@@ -133,7 +122,23 @@ Public Class dlgInfill
...
+        ElseIf ucrInputLimitOptions.GetText = "Fixed Start Limit" Then
+            clsDefaultFunction.RemoveParameterByName("end_date")
+            clsDefaultFunction.AddParameter("start_date", clsRFunctionParameter:=ucrDtpStartDate.ValueAsRDate, iPosition:=3)
+        ElseIf ucrInputLimitOptions.GetText = "Fixed End Limit" Then
+            clsDefaultFunction.RemoveParameterByName("start_date")
+            clsDefaultFunction.AddParameter("end_date", clsRFunctionParameter:=ucrDtpEndDate.ValueAsRDate, iPosition:=4)
+        Else
+            clsDefaultFunction.RemoveParameterByName("start_date")
+            clsDefaultFunction.RemoveParameterByName("end_date")
+        End If
```
lloyddewit, line 138: Can we simplify this sub?
```suggestion
        clsDefaultFunction.RemoveParameterByName("start_date")
        clsDefaultFunction.RemoveParameterByName("end_date")
        Select Case ucrInputLimitOptions.GetText
            Case "Fixed Limits", "Fixed Start Limit"
                clsDefaultFunction.AddParameter("start_date", clsRFunctionParameter:=ucrDtpStartDate.ValueAsRDate, iPosition:=3)
            Case "Fixed Limits", "Fixed End Limit"
                clsDefaultFunction.AddParameter("end_date", clsRFunctionParameter:=ucrDtpEndDate.ValueAsRDate, iPosition:=4)
        End Select
```

shadrackkibet: Committed the suggestion. Thanks for the simplification @lloyddewit

lloyddewit (approved):

lloyddewit:
@rdstern One small change since your last approval (functionality should not be affected), please could you approve again? thanks

rdstern (approved):

Issue #6384, opened 2021-04-06
Adding to the current Comments system


rdstern:
The addition of comments is important for R-Instat generally, and particularly for the quality control and the new data entry.  The idea - a bit like adding comments to a cell in Excel, is that you can note items of particular interest.  

In R-Instat they go onto their own sheet.  This is also particularly useful to keep a record of changes you may make of data in the data book.

You usually right-click to add a comment.

- [x]  The first change I suggest.  In the right-click for a cell, it is currently on a special cell sub-click.  I suggest (for now) we remove the cell category.  Put Add comment instead on the main right-click, maybe just above Filter.  

- [x]  Put Add comment just above Filter on column right-click.

- [x] Then its use depends on where you are.  If you are at the top, then it is a Column comment.  If you start from a cell, then it is a cell comment.  Either way it simply keeps that information and loads the dialogue, ready to add the comment.

- [x]  The second change is within the dialogue and is to add Column and Data Frame after the buttons at the top for Cell and Row.  I think they'll be used in that order, so I am happy with Cell, then Row, then Column, then Data Frame.

- [ ]   Currently only a single comment is added.  If it is easy, then for Cell and Row, I would like additionally to have a checkbox with the label Add to Each Visible Row.  Default is unchecked, and it does not remember.  Whenever you return to this, it is unchecked.  

The idea is that you might filter to all the rows with a particular feature.  Then you comment at the same time on all of these.  At the same time it would include a warning if the number of rows is too large, say more than 25.  "This will add xxx comments into the data frame.  Are you sure you wish to continue?"

- [x]  Unrelated, but while you are changing the right click I notice that the Levels/Labels behaves differently when in a cel, compared to the top row.

PR #6387, opened 2021-04-06
Enhancement of cell right click 

shadrackkibet:
This PR fixes "marked" points in issue #6384

- Moved add comment into the main right-click 
- Enable levels/labels only on the factor column

### DIFF
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -961,9 +961,12 @@ Public Class ucrDataView
     Private Sub cellContextMenuStrip_Opening(sender As Object, e As CancelEventArgs) Handles cellContextMenuStrip.Opening
         Dim iSelectedCols As Integer = grdData.CurrentWorksheet.SelectionRange.Cols
         Dim strColumns() As String = SelectedColumnsAsArray()
-        Dim strType As String = frmMain.clsRLink.GetColumnType(grdCurrSheet.Name, strColumns(0))
+        Dim strType As String
 
-        mnuLebelsLevel.Enabled = (iSelectedCols = 1 AndAlso strType.Contains("factor"))
+        If strColumns IsNot Nothing AndAlso strColumns.Count > 0 Then
+            strType = frmMain.clsRLink.GetColumnType(grdCurrSheet.Name, strColumns(0))
+            mnuLebelsLevel.Enabled = (iSelectedCols = 1 AndAlso strType.Contains("factor"))
+        End If
         mnuRemoveCurrentFilters.Enabled = Not String.Equals(strFilterName, strNoFilter)
     End Sub
 End Class
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -969,7 +969,8 @@ Public Class ucrDataView
             strType = frmMain.clsRLink.GetColumnType(grdCurrSheet.Name, strColumns(0))
             mnuLebelsLevel.Enabled = (iSelectedCols = 1 AndAlso strType.Contains("factor"))
         Else
-            'TODO: Add developer error?
+            MsgBox("Developer error: SelectedColumnsAsArray() expected to return an array with at least one element.")
+            mnuLebelsLevel.Enabled =  False
         End If
         mnuRemoveCurrentFilters.Enabled = Not String.Equals(strFilterName, strNoFilter)
     End Sub
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -959,6 +959,11 @@ Public Class ucrDataView
     End Sub
 
     Private Sub cellContextMenuStrip_Opening(sender As Object, e As CancelEventArgs) Handles cellContextMenuStrip.Opening
+        Dim iSelectedCols As Integer = grdData.CurrentWorksheet.SelectionRange.Cols
+        Dim strColumns() As String = SelectedColumnsAsArray()
+        Dim strType As String = frmMain.clsRLink.GetColumnType(grdCurrSheet.Name, strColumns(0))
+
+        mnuLebelsLevel.Enabled = (iSelectedCols = 1 AndAlso strType.Contains("factor"))
         mnuRemoveCurrentFilters.Enabled = Not String.Equals(strFilterName, strNoFilter)
     End Sub
 End Class
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -685,10 +685,6 @@ Public Class ucrDataView
         End If
     End Sub
 
-    Private Sub mnuCellCopyRange_Click(sender As Object, e As EventArgs) Handles mnuCellCopyRange.Click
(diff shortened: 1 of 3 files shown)


rdstern (approved):
@shadrackkibet great.  Changes the position of the Add Comment option and also sorts out a problem on the Levels/Labels
Over to @lloyddewit

instat/ucrDataView.vb
```diff
@@ -972,6 +959,11 @@ Public Class ucrDataView
     End Sub
 
     Private Sub cellContextMenuStrip_Opening(sender As Object, e As CancelEventArgs) Handles cellContextMenuStrip.Opening
+        Dim iSelectedCols As Integer = grdData.CurrentWorksheet.SelectionRange.Cols
+        Dim strColumns() As String = SelectedColumnsAsArray()
+        Dim strType As String = frmMain.clsRLink.GetColumnType(grdCurrSheet.Name, strColumns(0))
```
lloyddewit, line 964: `SelectedColumnsAsArray()` may potentially return an empty array, this will trigger an exception on this line.
Check for empty array?

lloyddewit (changes requested):

instat/ucrDataView.vb
shadrackkibet: Thanks, that is right. I have now added a check.

instat/ucrDataView.vb
```diff
@@ -961,9 +961,12 @@ Public Class ucrDataView
...
         Dim iSelectedCols As Integer = grdData.CurrentWorksheet.SelectionRange.Cols
         Dim strColumns() As String = SelectedColumnsAsArray()
-        Dim strType As String = frmMain.clsRLink.GetColumnType(grdCurrSheet.Name, strColumns(0))
+        Dim strType As String
 
-        mnuLebelsLevel.Enabled = (iSelectedCols = 1 AndAlso strType.Contains("factor"))
+        If strColumns IsNot Nothing AndAlso strColumns.Count > 0 Then
+            strType = frmMain.clsRLink.GetColumnType(grdCurrSheet.Name, strColumns(0))
+            mnuLebelsLevel.Enabled = (iSelectedCols = 1 AndAlso strType.Contains("factor"))
+        End If
```
lloyddewit: Do we need an `Else` statement?
Also, we should normally declare variables in the scope where they are needed
```
        Dim strColumns() As String = SelectedColumnsAsArray()
        If strColumns IsNot Nothing AndAlso strColumns.Count > 0 Then
            Dim iSelectedCols As Integer = grdData.CurrentWorksheet.SelectionRange.Cols
            Dim strType As String = frmMain.clsRLink.GetColumnType(grdCurrSheet.Name, strColumns(0))
            mnuLebelsLevel.Enabled = (iSelectedCols = 1 AndAlso strType.Contains("factor"))
        Else
            mnuLebelsLevel.Enabled = False
        End If
```

lloyddewit (changes requested):

instat/ucrDataView.vb
shadrackkibet: I don't think an else will be `mnuLebelsLevel.Enabled = False`. Perhaps a developer error will be more appropriate? 

Another point is that  `mnuLebelsLevel.Enabled = (iSelectedCols = 1 AndAlso strType.Contains("factor"))` is a short-hand which can be `false` so we already have an else in that case.

For the scope, I would define scope differently. I would consider the subroutine as the scope where I declare the variables and make assignments between checks. Therefore I would have the following in the subroutine.

```
        Dim iSelectedCols As Integer
        Dim strColumns() As String
        Dim strType As String

        strColumns = SelectedColumnsAsArray()
        If strColumns IsNot Nothing AndAlso strColumns.Count > 0 Then
            iSelectedCols = grdData.CurrentWorksheet.SelectionRange.Cols
            strType = frmMain.clsRLink.GetColumnType(grdCurrSheet.Name, strColumns(0))
            mnuLebelsLevel.Enabled = (iSelectedCols = 1 AndAlso strType.Contains("factor"))
        Else
            "developer error"
        End If
        mnuRemoveCurrentFilters.Enabled = Not String.Equals(strFilterName, strNoFilter)
``` 

What do you think?

rdstern (dismissed):
Still happy with this from my side

instat/ucrDataView.vb
lloyddewit: > I don't think an else will be `mnuLebelsLevel.Enabled = False`. Perhaps a developer error will be more appropriate?
> 
For me, a developer error is only when we have a situation that should never occur and indicates a coding error. If you think that an empty `strColumns` array meets this criteria then I agree that a developer error is the best solution.

> For the scope, I would define scope differently. I would consider the subroutine as the scope where I declare the variables and make assignments between checks. Therefore I would have the following in the subroutine.
> 
> ```
>         Dim iSelectedCols As Integer
>         Dim strColumns() As String
>         Dim strType As String
> 
>         strColumns = SelectedColumnsAsArray()
>         If strColumns IsNot Nothing AndAlso strColumns.Count > 0 Then
>             iSelectedCols = grdData.CurrentWorksheet.SelectionRange.Cols
>             strType = frmMain.clsRLink.GetColumnType(grdCurrSheet.Name, strColumns(0))
>             mnuLebelsLevel.Enabled = (iSelectedCols = 1 AndAlso strType.Contains("factor"))
>         Else
>             "developer error"
>         End If
>         mnuRemoveCurrentFilters.Enabled = Not String.Equals(strFilterName, strNoFilter)
> ```
> 
> What do you think?

This is 3 lines longer than my proposal but we have no formal coding standards so I think it comes down to personal taste.
Therefore, I've no problem if you prefer to implement like this, it looks good :)

rdstern (dismissed):

instat/ucrDataView.vb
```diff
@@ -972,6 +959,18 @@ Public Class ucrDataView
...
+        Dim strColumns() As String
+        Dim strType As String
+
+        strColumns = SelectedColumnsAsArray()
+        If strColumns IsNot Nothing AndAlso strColumns.Count > 0 Then
+            iSelectedCols = grdData.CurrentWorksheet.SelectionRange.Cols
+            strType = frmMain.clsRLink.GetColumnType(grdCurrSheet.Name, strColumns(0))
+            mnuLebelsLevel.Enabled = (iSelectedCols = 1 AndAlso strType.Contains("factor"))
+        Else
+            'TODO: Add developer error?
```
lloyddewit, line 972: Please could you add a developer error here?
Would it also be sensible to set `mnuLebelsLevel.Enabled` to `False` (otherwise it could potentially remain `True` even when the required conditions are not met)?
```suggestion
            MsgBox("Developer error: SelectedColumnsAsArray() expected to return an array with at least one element.")
            mnuLebelsLevel.Enabled =  False
```

lloyddewit (changes requested):

lloyddewit:
@rdstern Please could you approve again? thanks

PR #6402, opened 2021-04-09
Allow comments to be added to columns and dataframes

### LINKED ISSUE #6384: Adding to the current Comments system

shadrackkibet:
fixes #6384

@africanmathsinitiative/developers this is ready for review. This feature is also available via right-clicks.

### DIFF
--- a/instat/dlgAddComment.Designer.vb
+++ b/instat/dlgAddComment.Designer.vb
@@ -50,6 +50,8 @@ Partial Class dlgAddComment
         Me.ucrBase = New instat.ucrButtons()
         Me.ucrSelectorAddComment = New instat.ucrSelectorByDataFrameAddRemove()
         Me.ucrInputRow = New instat.ucrInputTextBox()
+        Me.rdoDataFrame = New System.Windows.Forms.RadioButton()
+        Me.rdoColumn = New System.Windows.Forms.RadioButton()
         Me.SuspendLayout()
         '
         'lblRow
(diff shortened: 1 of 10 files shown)


rdstern (approved):
@shadrackkibet this is great.  I am approving as it is.  But if you want me to be really picky:
a) It is nice to have it on the right-click and even at the bottom.  The column comment could also be added ro the right-click on the column metadata, and the data frame right-click could add a data frame comment.
b) I keep forgetting to have a key to start with.   It gives me an error and I define a key and it is happy.  I wonder if the dialogue could detect if there is no key and not enable Ok, but give me a message to make a key first?

shadrackkibet:
@rdstern 
a) Now implemented, please check. 
b) Let's consider it as a separate improvement to be done on a separate PR.  I will therefore open an issue for it.

rdstern (approved):

lloyddewit (approved):

PR #6247, opened 2021-03-03
Opening from script feature

Ivanluv:
This currently implements the opening of the split dialogue from script ,though not fully functional

To review this you need to :
1. Download this  file [RScript .zip](https://github.com/africanmathsinitiative/R-Instat/files/6106590/RScript.zip)
2. Extract the RScript .dll file and save  it to an appropriate folder on your system
3. Add the RScript .dll to your R-Instat project by 
  - opening the  `Solution Explorer`
 -  right click on `References` File
 -  select` Add Reference ...`
 -  select `Browse`
 -  navigate to the folder where you saved the .dll and select it

You can also view the source code for this RScript.dll file here https://github.com/lloyddewit/RScript

Among the things not functioning as required are:

- Incorrect updating of controls that use `AddAdditionalCodeParameterPair()`

PR #6407, opened 2021-04-12
Opening Dialogues From Script

Ivanluv:
This  is a continuation of PR #6247


Ivanluv:
@lloyddewit could you review this

lloyddewit:
The approach used in this PR to open dialogs from scripts works correctly.
However, we found that it requires each dialog to be adapted. Therefore it will be impractical to apply this approach to all the R-Instat dialogs.
We are currently trying to develop a semi automatic method to create/reopen dialogs from script. We will build on what we learned from this PR and it's important that we keep the changes made in this PR.
Further work will be in a new PR(s). So I propose to close this PR and keep it for reference.

PR #6408, opened 2021-04-12
Added menu for Setup data entry dialog

N-thony:
@lloyddewit this PR adds a menu for Setup Data Entry dialog.

### DIFF
--- a/instat/frmMain.Designer.vb
+++ b/instat/frmMain.Designer.vb
@@ -1607,8 +1607,8 @@ Partial Class frmMain
         '
         'mnuSetupForDataEntry
         '
-        resources.ApplyResources(Me.mnuSetupForDataEntry, "mnuSetupForDataEntry")
         Me.mnuSetupForDataEntry.Name = "mnuSetupForDataEntry"
+        resources.ApplyResources(Me.mnuSetupForDataEntry, "mnuSetupForDataEntry")
         '
         'mnuTidyandExamineClimaticDataEntry
         '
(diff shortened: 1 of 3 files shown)


Patowhiz:
@lloyddewit this PR adds a menu item to frmMain. @rdstern and I need it to test and add to implementations done by @N-thony

Patowhiz:
@lloyddewit it would be helpful if it could be merged before conflicts arise.

lloyddewit (changes requested):
This menu option will currently open an empty dialog. Normally, we should not have this user behaviour on the main branch.
Please could you set the new menu option's visibility to false? You could then set the menu option's visibility to true in the branch where you implement the Setup Data Entry dialog. The chances of a conflict on `frmMain` should be small.
Does this make sense? if not, then I'm happy to discuss
Thanks

N-thony:
> This menu option will currently open an empty dialog. Normally, we should not have this user behaviour on the main branch.
> Please could you set the new menu option's visibility to false? You could then set the menu option's visibility to true in the branch where you implement the Setup Data Entry dialog. The chances of a conflict on `frmMain` should be small.
> Does this make sense? if not, then I'm happy to discuss
> Thanks

@lloyddewit I would prefer to make it disable instead of visible false.

lloyddewit:
@N-thony disabled is also OK

N-thony:
@lloyddewit this is ready now.

shadrackkibet:
Disabled gives an impression to the user that they have to do something for it to be enabled. I would prefer setting visibility to false.

maxwellfundi (approved):
@lloyddewit over to you

lloyddewit (approved):

PR #6411, opened 2021-04-13
Editing Summary Tables- adding outer margins, creating long format to use for `mmtable2`

lilyclements:
This PR adds outer margins into the summary tables function. This then means that one summary table is produced in "long format" despite the number of columns to summarise or summary statistics chosen.

Note that this function runs the calculation side of the summary tables.
`mmtable2` is then can be used to run the aesthetic side of the summary tables.
Additionally, it is planned for `htmltable` to be another suitable alternative (this may require writing a function).


rdstern:
I realise this is "work-still-in-progress".  
I can see that there is an extra possible checkbox on treat summary columns as an additional factor.  I assume there will eventually also be one on the summaries too?
I tried with the survey data as before and then added the outer margins.  
image


It always says:

image

lilyclements:
The outer margins work for this function, however, the "summary margins" do not yet work in R-Instat. This is due to an issue in the linking outlined in #6477.

image
_Outer margins only_

image
_Summary and outer margins_

rdstern (commented):
@lilyclements is there anything I can test here?  I am happy to add the package to do this.  Does it need edits to the summary and frequency tables dialogues first?

rdstern:
@lilyclements and @Ivanluv I keep getting the following?
image

When I look at the commands generated I don't see why this argument should be a problem?
The code for one example was as below, but the error message is the same for all examples - with these data as with another example.

```
# Code generated by the dialog, Summary Tables

summary_table <- data_book$summary_table(data_name="survey", columns_to_summarise="yield", factors=c("village","variety"), prop_test="'>'", prop_value=30, summaries=c("summary_count_non_missing", "proportion_calc", "summary_mean"), As_percentage=TRUE, treat_columns_as_factor=FALSE, j=1)


last_table <- mmtable2::mmtable(data=summary_table, cells=value) + mmtable2::header_left_top('summary-variable') + mmtable2::header_top_left(village) + mmtable2::header_left_top(variety)
data_book$add_table(table_name="last_table", table=last_table, data_name="survey")
data_book$get_tables(data_name="survey", table_name="last_table")

rm(list=c("last_table", "survey", "summary_table")) 
```

Ivanluv:
@rdstern you have to  reinstall mmtable2 in R then the error will be fixed

rdstern:
@lilyclements and @HawardKetoyoMsatsi this now seems to have gone "backwards a bit?  The results are now in the output window, but not as they should be:

image

I get the same whether I have store results in Data or not.  I thought "store results in Data would put them into a data frame?
If I untick the HTML, then I get:

image

If I add to display outer margins then I get this:

image

Ivanluv:
@rdstern  the dialogue from the screenshots you have seem to be the "old" dialogue before the changes I added. 
Could you update from PR #6508 and try testing again

rdstern:
@Ivanluv and @lilyclements sorry about that.  Should this pull request be closed?

Ivanluv:
Yes, this be closed because this changes are included in my PR #6508.

PR #6412, opened 2021-04-14
Added new options to the Import dialog for importing CSV and text files

### LINKED ISSUE #6378: Adding options when importing multiple files
Currently this import is only for the default settings.  That is ok for importing from SPSS, Stata or SAS, so this is already useful.  But with csv or text files it would be good to include options, particularly to select missing values.

At the same time, the label to import multiple files has made the dialogue wider.  It would be good to reduce the width, so it is no wider than is needed by the preview window.  REduce the space after the File: label and the Browse button could be slightly smaller.  Finally, if needed the main field width could also be reduced slightly.

Patowhiz:
Fixes #6378


rdstern (approved):
@Patowhiz great.
@lloyddewit over to you

lloyddewit (approved):

Issue #6396, opened 2021-04-08
A null reference exception triggered in the translation file


shadrackkibet:
**Describe the bug**
There is an exception triggered in the `Translation.vb` file line 75 when you click "Yes" from auto-recovery options. 

**To Reproduce**
- Load any dataset.
- Do some analysis, say use any dialog
- Then close R-Instat
- Restart again. Hopefully, recovery options are shown, click "Yes".

![image](https://user-images.githubusercontent.com/26301973/114002007-26806680-9865-11eb-8b88-6d820ba9227f.png)
![image](https://user-images.githubusercontent.com/26301973/114002979-0f8e4400-9866-11eb-870f-e0e6cd669dda.png)

In the current version (0.7.0) I get the following.
![image](https://user-images.githubusercontent.com/26301973/114003833-c2f73880-9866-11eb-830e-01b98474664a.png)

I, therefore, tag @lloyddewit and @dannyparsons because you have recently been working on the translation.

shadrackkibet:
Interesting to also note that the directories are pointed to Danny's directories in my currently installed version.
![image](https://user-images.githubusercontent.com/26301973/114005353-2170e680-9868-11eb-9a11-6d09d9e1b16c.png)

lloyddewit:
@shadrackkibet Thank you for detecting and documenting this. I am able to recreate the error.
Clarification: the bug is only triggered when I close R-Instat by ending it's process (e.g. from within the debugger). The bug is not triggered when R-Instat exits normally.
I will fix this bug.

lloyddewit:
Fixed in PR #6413

PR #6413, opened 2021-04-14
Added option to display dialog boxes in French

lloyddewit:
Partially fixes issue #3721
Fixes issue #6396 

This PR adds translation functionality to the dialog boxes. The dialog button to toggle between English and French is also implemented. Currently the French translation is the English text prefixed with 'fr' e.g. 'File' is translated to 'frFile'. I did this deliberately so that during testing it is easier to see which text is translated and which is not. Before this PR is merged, I will swap in the database file that contains the real French translations.

@rdstern Not all text is translated yet. I will add a comment below that lists all the known untranslated text. I will resolve these in a future PR (or time allowing, resolve some in this PR). Please could you test and identify any issues that I have missed?

@shadrackkibet Please could you peer review? We need to merge by the end of the week (so that this can be included in the release for next week's event). However you may wish to wait for @rdstern to finish his testing first in case changes are needed.

Note: The translation mechanism uses the TranslateWinForms library. I uploaded this library to NuGet so that it can be integrated into R-Instat in the standard way. The source code for this library can be found at:
https://github.com/lloyddewit/TranslateWinForm
We can move the library to the africanmathsinitiative GitHub site when it is stable.


lloyddewit:
Text that is not yet translated:

- check box text
- input line text
- labels that change dynamically
- tool tips
- message boxes (e.g. startup and developer errors)
- other specific cases (see screen shots below)

image

image

image

image

image

image

image

image

image

image

image

image

image

rdstern (changes requested):
@lloyddewit it is looking promising - in a way?  But it now puts fr in front of each item instead of translating it.  I suppose that's sort of French, but it actually translated the menus before!

lloyddewit:
@rdstern If you look at my original comment above, it states:

> This PR adds translation functionality to the dialog boxes. The dialog button to toggle between English and French is also implemented. **Currently the French translation is the English text prefixed with 'fr' e.g. 'File' is translated to 'frFile'. I did this deliberately so that during testing it is easier to see which text is translated and which is not. Before this PR is merged, I will swap in the database file that contains the real French translations.**

I do not claim that 'frFile' is real French. My French is bad but not that bad :)

rdstern:
Ok, I started with the Prepare menu.  This is what I found so far:  I am sure some of the points may be covered in your note above, but I repeat anyway!

a) The reset key isn't translated - that's one of the 5 at the bottom of each dialogue.  The translation of `réinitialiser` is fine except it is long.  I am wondering about remettre or remonter?  I am checking.
b) The Row-numbers/Names dialogue has the first option, i.e. Column Name is not translated.  This is perhaps one of many situations that - when translated - may make us reflect on the English.  the French will be longer, but I suggest that just having the English label `Name`, rather than `Column Name` would be ok here.
c) I then jumped to the calculator.  I notice here that the drop-down with the names of the keyboards is not yet translated.  On the other hand the names of the keys are going to be translated.  I suggest this is an example where we should stick with the English - though as I say that I am not quite sure?  They are not always the function.  This needs some thought.  Perhaps stick with it for now, particularly as we (I hope) will have the option to look in English if we wish.  I think we may wish to translate sometimes and could perhaps allow for this by making the translation sometimes into the same thing, e.g. frNA could be NA!
d) A lot is translated!  It will be good.
e) The name at the top of the dialogue is not yet always translated.  This is often the same (or an extended version of the menu item.  Now I see it is sometimes.  Prepare > Data Frame > Sort is an example where it isn't.  Now I try to find more and can't!
f) the Prepare > Check Data > Visualise data is an example where I expect the label for sampling fraction to be too large, but there is plenty of space to move the start of the label to the left.
g) In the Prepare > Check Data > Duplicates the label New Column Name is not yet translated.  This time I slightly prefer to keep New Column Name in English - don't feel very strongly, but we might consider New Name to be translated into French?  There are quite a lot of instances where we will have either New Column Name or New Variable Name and it is pretty obvious what it is that will be new!
 I'll continue later.

rdstern:
@lloyddewit tell me when there is more I can look at.

lloyddewit:
@rdstern I committed changes that resolve some (but not all) of your observations above. I hope to make additional commits by end today that will resolve more. You are very welcome to test now, or if you prefer, after the next commit.

lloyddewit:
@rdstern I think I resolved your observations above (apart from the calculator pull-down list). 

Text that is not (yet) translated:

- Pull-down lists
- Tool tips
- Message boxes (e.g. startup and developer errors)

Other observations:

- Check boxes do not always respond correctly to the language toggle button
- 'Help' button not always translated

Please could you test again? It is probably most efficient if you report just 5 or 6 issues. I can then resolve and we can iterate. 
Many thanks!

rdstern:
It is looking really good now.
a) I checked various Prepare > Objects dialogues - seems ok
b) Describe > One Variable > All dialogues - seems ok
 Except I think that the Return and Help buttons are not translated, possibly that's a general issue.

(No, I now checked the Describe > Multivariate > Correlations and then the Return and Help on the sub-dialogue are ok.) 


c) Describe > Specific > Boxplot  (and seems the same on other plots)

Some problems with the sub-dialogues here.
1) In the boxplot options the aesthetics - one of them is translated, i.e. frlabel, but not the others.  These are arguments, so should they be translated or not?  Perhaps that's also a @dannyparsons or @volloholic question.  It could be a lot of work, but should there be tooltips (translated) if the arguments are not?
2) In the geom - that is the other tab, nothing is translated - not even the title.

3) In the main plot sub-dialogue most of the tabs are ok but not most of the Y-axis tab.  The x-axis looks ok in contrast.  Note that the default here is discrete, so look at continuous and date as well.

4) I think the Return and Help are not transalted here.

d) In the Model > Probability Distributions > Show Model when I come into the dialogue the first time the Mean is translated, but not the St Dev.  Once I click to Exponential, then Mean there isn't translated.  If I now go back to Normal, then the Mean isn't translated now.  I can't get it to translate the mean except by closing and re-opening the dialogue again.

lloyddewit:
@rdstern Thank you for the observations above. I think I resolved many of them. Please could you retest and report another batch of issues?

Please note that translation for the following text types is not yet implemented:
- drop-down lists
- tool tips
- R parameters
- cases where a control has >1 instance in a (sub)dialog

I added extra explanation for the above to issue #3721. 

I personally noticed some issues with checkboxes toggling between languages, and some untranslated tabs in subdialogs. I'll look into these.

Thanks for your help!

rdstern:
Having written the items below I have now added 3 pull requests - so I'll try again!

I thought the checkboxes were now ok:  Not always:
image

In the boxplot options the label is still as before.  Perhaps I haven't updated?

Now after the update - both of these are ok!  Great.

image

In the latest version the y-axis is still as before - but I wonder if that is really a problem.  If all instances are like this, then could we leave the y-axis mainly in English and add a note below, which is translated, and is to the effect that `Some terms here are not translated from English.  Use the x-axis tab to see the translation of each of these.`  So we turn this issue into a feature instead!

image

Checked Describe > Multivariate and found nothing to report.

I noticed from the Model > Probability Distributions > Show Model that even the pull-down with the distributions is now translated.  Great.  So I checked the Prepare > Calculations and there the pull down of the keyboards isn't translated?

rdstern (commented):
@lloyddewit there is lots that does translate.  So I am happy for a merge whenever needed.

lloyddewit:
@rdstern Thank you for the observations, they are very helpful and identified some things I missed.

> In the boxplot options the label is still as before. Perhaps I haven't updated?
> 

I investigated:
- If the language is French when R-Instat is started (i.e. it was set to French the last time it was closed), then the checkboxes display correctly in French.
- If R-Instat starts in English, is then changed to French, and a dialog is opened for the first time (i.e. has not yet been opened in the current session), then the checkboxes display correctly in French.
- If R-Instat starts in English, is then changed to French, and a dialog is reopened, then the checkboxes display incorrectly in English. The same happens if the user toggles languages in an open dialog

I think I know the cause and can fix it.


> In the latest version the y-axis is still as before - but I wonder if that is really a problem. If all instances are like this, then could we leave the y-axis mainly in English and add a note below, which is translated, and is to the effect that `Some terms here are not translated from English. Use the x-axis tab to see the translation of each of these.` So we turn this issue into a feature instead!

This is an interesting proposal! My personal preference is to try and fix the issue so it displays in French. But your suggestion could be a temporary workaround.

> I noticed from the Model > Probability Distributions > Show Model that even the pull-down with the distributions is now translated. Great. So I checked the Prepare > Calculations and there the pull down of the keyboards isn't translated?

I assume you mean 'Prepare > Calculator...'?
In R-Instat, the dropdown list items are normally added individually and there is no centralized place in the code to translate these items. Therefore, each drop down list needs one or two lines of code added to get it to translate. This shouldn't be hard to do but will take a little time. I translated the Show Model drop-down list to check that it could be done. I mentioned in a previous comment that pulldown lists are not currently translated.

**Conclusion:** The issues above can be resolved. We are getting closer to the release data and I would prefer to first merge this PR and fix the above issues in future PRs. I hope this is acceptable for you.

rdstern:
Very much do.  We have just had the request for Climsoft in French so this is all very timely.  That should be simpler

lloyddewit:
@shadrackkibet Please could you peer review this PR? 

Note: The translation mechanism uses the TranslateWinForms library. I uploaded this library to NuGet so that it can be integrated into R-Instat in the standard way. The source code for this library can be found at:
https://github.com/lloyddewit/TranslateWinForm

Please could you also peer review this library? It is a repo rather than a PR so I'm not sure about the best way to provide comments. Do you have any ideas? We can discuss if needed.

Many thanks

lloyddewit:
@rdstern Update to your comment below about the Box Plot dialog checkboxes not translating dynamically (i.e. if the dialog was first started in French then the check box labels are always displayed in French).

> In the boxplot options the label is still as before. Perhaps I haven't updated?
> 

I investigated and this is just another instance of the known problem 'cases where a composite control has >1 instance in a (sub)dialog' (i.e. identical to the issue with the `Y-Axis` tab in the Plot Options subdialog). This is because there are multiple checkboxes and each checkbox has a subcontrol with the same name (`chkCheck`) so only one of the checkbox labels gets translated (`Group to Connect`).

shadrackkibet:
@lloyddewit could you explain in one or two paragraphs how the translation mechanism works? This will be useful for me to do an effective review. I remember I had asked about it somewhere when work on translation started. You could add more information about translation in general. Perhaps a skype/zoom will be effective? Otherwise, a written explanation will be useful for other developers.

dannyparsons:
> In the latest version the y-axis is still as before - but I wonder if that is really a problem. If all instances are like this, then could we leave the y-axis mainly in English and add a note below, which is translated, and is to the effect that `Some terms here are not translated from English. Use the x-axis tab to see the translation of each of these.` So we turn this issue into a feature instead!

We need to fix this general issue as Stephen mentioned. I don't think this would be a great "feature" and also in other case of this issue the text to translate is different between two instances of the same control so we need to be able to translate all controls. However, I think this is rare enough to be an acceptable bug for the first version of this. So I don't think it needs to be solved before merging this.

@lloyddewit It might be more efficient for me to review the code changes here as I reviewed the initial version so that might save Shadrack having to learn the whole system.

instat/ucrCheck.vb
```diff
@@ -23,15 +23,6 @@ Public Class ucrCheck
...
     Private strValueIfUnchecked As String = "FALSE"
 
-    Public Overrides Property Text() As String
-        Get
-            Return chkCheck.Text
-        End Get
-        Set(value As String)
-            chkCheck.Text = value
-        End Set
-    End Property
```
dannyparsons, line 33: Why was this removed?

dannyparsons (commented):
It's hard to evaluate thoroughly as the changes are substantial but I have reviewed the main files changed and gone through all files changed briefly. I also looked at the translation library code.
Just one question for clarification below.

instat/ucrCheck.vb
lloyddewit: Good question. I discussed this with @Patowhiz on 18/04/21. He agreed with this change. Please see below for more details:
image

image

dannyparsons: Thanks that makes sense, I don't see any need for the extra property.

dannyparsons (dismissed):

lloyddewit:
@dannyparsons Thanks for approving
Next steps:
- you approve/merge PR #6444 (new English JSON file for non-menu)
- you make new 'fr' file with auto translation
- I make new database and add it to this PR
- @rdstern and I sanity check the French translations
- we approve and merge this PR

lloyddewit:
@dannyparsons I added the new database that contains the dynamic text translations. Please could you sanity check this PR.
Issue #3721contains a list of translation cases that still need to be implemented. If you find any extra missing cases then please let me know.
Thanks

dannyparsons:
This looks to be working well generally. Menus are translated, dialogs are translated, the speed is fast. The main issue I found which I don't see in the know issues is that checkboxes don't get translated after they've been opened once. 

e.g. If you start in English, open a dialog with a checkbox, close, set to French, reopen, the checkbox is still in English. The "Lan" button doesn't change the checkboxes either.

It does translate if you start in French e.g. restart R-Instat, set language to French, open the dialog, the checkbox will be in French, close and change to English, reopen the dialog and the checkbox is still in French. So it never changes more than once.

Could this be related to removing the `Text` property in `ucrCheckbox`? It's hard to debug as I can't debug in the code within the `TranslateWinForm` library, is that possible?

dannyparsons (approved):

Issue #6397, opened 2021-04-08
Completing the Position control for saving new variables


rdstern:
This system is just-about working.  I am going resaonably systematically through the dialogues where it is used.  I am particularly looking for 4 things:
a) Does it work - I am hoping for that.  That's called `Work`.
b) Is the Position button nicely aligned, it was various sizes and slightly off centre.  I call that `Align`.
c) Is the label about the saving ok? In some cases the control was pushed to the left and obscured the label.  I call that `Adjust`.
d) Is the default position ok.  Often the new variable relates to an existing one, and the obvious default is next to it.  That's called `Default` and I'll explain in detail when this occurs.

**_Prepare >Data Frame_**
1. Row names/Numbers:  ok

**_Prepare > Check Data_**
1. Duplicates:  Align.  Also Default:The new variables always come at the end.  For the third option it is simple, because there is a single receiver.  The default should be after the variable.  For the first option it takes a whole data frame, so the existing is fine.  The middle option takes multiple Variables.  Ideally the default is after the last variable in the multiple receiver.  I hope that will be easy.  (That will usually be the last variable in the data frame, but happy with the rule that what is last in the receiver is the basis for the position.
2. Compare Columns.  Align. also Adjust: Change to `Save result for second column: `  Default to after 2nd column.  It is currently at the end.
3. Non-Numeric.  Adjust:  Add colon and here there is too much space after the label.
4. Anonymise. Align, Change name from Hash to hash.  Default to after variable - currently at end.

**_Prepare > Calculator_**
Add colon ":" to Save Result.  Now also a challenge.  If a variable is clicked to go into the formula, is this detected?  If so, then default to after the last variable.  If tricky, but first variable is possible, then after that.  (But don't spend too long on this.  For example you may recall a formula and then not know.  Leave that.  Possibly have to leave with the End being the default.  Alternatively, is it easy to detect the current active variable in the data frame.  If so, then it might be easy for the default to be after that variable. 

**_Prepare > Column: Calculate_**
1. Regular Sequence.  Align.  This is another example where there is no variable.  Could it also go after the current variable in the grid, if that exists.  Otherwise at the end.
2. Enter.  There is a big gap after the label Enter Result into which should be smaller.  There is no position control yet.  As with the others, ideally after the current active variable, or at the end.
3.  Duplicate   Align only!
4. Row Summaries.  Align.  After last variable.
[...]

PR #6414, opened 2021-04-14
Added option to compare by row in the Compare Columns dialogue

### LINKED ISSUE #6393: Add an option to the Compare Columns Dialogue
There is a `Prepare > Check Data > Compare Columns` dialogue currently.
I propose we add 2 radio buttons at the top.  The first is `By Row` and the second is `By Value`
The option described below is to add is `By Row`.  The existing dialogue is comparing `By Value`.

The `By Row` should be simpler! Its main use is to compare 2 numerical variables, row, by row, though I consider below the situation with other types of column.
1. I suggest, the usual data selector and 2 single receivers, just as in the dialogue already.  The labels for those receivers can be the same.
2. And (initially) it writes out a logical variable - as does the other option.  And this is TRUE if the 2 values are the same and FALSE otherwise.  And there is also (like the other option), simply the same report in the output window namely the number of TRUE and FALSE.
3. Possibly, as now, allow any type of data.  But check that the data are the same type when press OK.  If not give a message.  Maybe once one control is completed, only allow the same type in the other field?
4. If numeric then there is a tolerance with an option, into which you can type, but there are some suggestions.  0 is an option, as is 0.05, and 0.001 and 1E-10.  Default can be 0.
5. Then the function can be `abs(first - second) <= tol`.
6. I hope this might work for logical, dates as well?
7. I assume for factor variables or character variables the tolerance isn't needed, and the function is simply `(first == second)` ?

(Note that 2 character variables can also be close, though not identical.  But then we would use the Prepare > Column: Text > Distance dialogue.)

 I am assuming that this compare will work with filtered data.  So, if the data are filtered, then (if a new variable is chosen) it is missing for all the data outside the filter.  If an existing (logical) variable is chosen then the results would "slot in" to part of that variable.  

In the next comment I mention extensions to this "system" that could be discussed/implemented.  But I am keen to get something simple that works for now.

HawardKetoyoMsatsi:
Fixes #6393

### DIFF
--- a/instat/dlgCompareColumns.vb
+++ b/instat/dlgCompareColumns.vb
@@ -20,8 +20,8 @@ Public Class dlgCompareColumns
 
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
-    Private clsCompareColumns As New RFunction
-    Private clsYinXOperator As New ROperator
+    Private clsCompareColumns, clsIfElseCompare, clsAbsoluteValue As New RFunction
+    Private clsYinXOperator, clsIsEqualToOperator, clsSubtractOperator, clsLessorEqualToOperator As New ROperator
 
     Private Sub dlgCompareColumns_Load(sender As Object, e As EventArgs) Handles Me.Load
         If bFirstLoad Then
@@ -40,6 +40,18 @@ Public Class dlgCompareColumns
 
     Private Sub InitialiseDialog()
         ucrBase.iHelpTopicID = 546
+
+        ucrPnlOptions.AddRadioButton(rdoByRow)
+        ucrPnlOptions.AddRadioButton(rdoByValue)
+
+        ucrPnlOptions.AddFunctionNamesCondition(rdoByRow, "ifelse")
+        ucrPnlOptions.AddFunctionNamesCondition(rdoByValue, {"%in%", "compare_columns"})
+
+        ucrPnlOptions.AddToLinkedControls({ucrChkSort, ucrChkUnique}, {rdoByValue}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrPnlOptions.AddToLinkedControls({ucrInputTolerance}, {rdoByRow}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+
+        ucrInputTolerance.SetLinkedDisplayControl(lblTolerance)
+
         ucrReceiverFirst.SetParameter(New RParameter("x", 0))
         ucrReceiverFirst.Selector = ucrSelectorCompareColumns
         ucrReceiverFirst.SetParameterIsRFunction()
@@ -53,6 +65,12 @@ Public Class dlgCompareColumns
         ucrReceiverSecond.bAttachedToPrimaryDataFrame = False
         ucrReceiverSecond.bOnlyLinkedToPrimaryDataFrames = False
         ucrReceiverSecond.bIncludeDataFrameInAssignment = True
+
+        ucrInputTolerance.SetParameter(New RParameter("tol", 1))
+        ucrInputTolerance.SetItems({"0", "0.005", "0.0000000001"})
+        ucrInputTolerance.SetValidationTypeAsNumeric()
+        ucrInputTolerance.AddQuotesIfUnrecognised = False
+        ucrInputTolerance.SetLinkedDisplayControl(lblTolerance)
 
         ucrChkUnique.SetParameter(New RParameter("use_unique", 2), bNewChangeParameterValue:=True)
         ucrChkUnique.SetText("Use unique values for comparison")
@@ -88,13 +106,21 @@ Public Class dlgCompareColumns
         ucrSaveLogical.SetIsTextBox()
         ' This ensures the assign text is correctly cleared when resetting
         ucrSaveLogical.bUpdateRCodeFromControl = False
+
+        ucrBase.clsRsyntax.iCallType = 2
     End Sub
 
     Private Sub SetDefaults()
         clsCompareColumns = New RFunction
+        clsIfElseCompare = New RFunction
+        clsAbsoluteValue = New RFunction
         clsYinXOperator = New ROperator
+        clsIsEqualToOperator = New ROperator
+        clsSubtractOperator = New ROperator
+        clsLessorEqualToOperator = New ROperator
 
         ucrBase.clsRsyntax.ClearCodes()
+        ucrInputTolerance.SetText("0")
 
         ucrSelectorCompareColumns.Reset()
         ucrReceiverFirst.SetMeAsReceiver()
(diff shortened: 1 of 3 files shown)


rdstern (changes requested):
@HawardKetoyoMsatsi great, this is basically working.
a) The tolerance doesn't seem to work.  I tried 0.2 and it always seemed to remain on zero.
b) I suggest, like the non-numeric dialogue you always save the resulting logical variable.  The label can be shorter, as in the other dialogue, so Logical Column:.
c) Could there be a suggested name, perhaps just compare.
d) Instead of the table, where there is the x as the output, perhaps also just use the summary command - again as in the non-numeric dialogue

HawardKetoyoMsatsi:
@rdstern I have made the requested changes, please take a look.

rdstern:
@HawardKetoyoMsatsi it is looking good now - and the tolerance works.  Just some small things.
a) The save logical column doesn't remember the setting from last time.
b) Could saving be the default, i.e. it is initially checked.
c) Suggest a default variable name, perhaps `compare`, then it would become compare1, etc.  There is a standard function for this.
d) When saved the output window prints Null see below.  Could this not be done?

image

HawardKetoyoMsatsi:
@rdstern Kindly have a look at the corrections I have made.

@dannyparsons @shadrackkibet I have removed two comments and a piece of code (below) below from the code and set the data frame selector because I find the save control automatically saves to the Dataframe that feeds the second receiver. Please take a look at it, it must have been implemented already.

`'Not setting data frame selector or prefix here because we need save control only linked to data frame of second selector which is not yet implemented`

    ```
Private Sub ucrReceiverSecond_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverSecond.ControlValueChanged
        ' Needs to be done manually because data frame name should only be the data frame name from the second receiver's varibable.
        ucrSaveLogical.SetGlobalDataName(ucrReceiverSecond.GetDataName())
    End Sub
```

**Review by rdstern — APPROVED** [review]
@HawardKetoyoMsatsi great.
Approving.  I hadn't realised that usually the other option may have variables from different data frames.  I am assuming the same data frame here.

**Comment by lloyddewit** [comment]
@N-thony Please could you peer review this PR? We can briefly discuss a good approach. I'll contact you on WhatsApp.

File: instat/dlgCompareColumns.vb [hunk 1]
```diff
@@ -141,26 +223,56 @@ Public Class dlgCompareColumns
...
+        If Not ucrReceiverSecond.IsEmpty Then
+            CheckDatatype()
+        End If
+    End Sub
+
+    Private Sub ucrReceiverSecond_ValueChanged(sender As Object, e As EventArgs) Handles ucrReceiverSecond.ValueChanged
+        If Not ucrReceiverFirst.IsEmpty Then
+            CheckDatatype()
+        End If
     End Sub
```
N-thony, line 277: Combine?

instat/dlgCompareColumns.vb
```diff
@@ -119,19 +184,36 @@ Public Class dlgCompareColumns
...
+            Else
+                ucrBase.OKEnabled(True)
+            End If
         Else
             ucrBase.OKEnabled(False)
         End If
     End Sub
 
-    Private Sub Controls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverFirst.ControlContentsChanged, ucrReceiverSecond.ControlContentsChanged, ucrSaveLogical.ControlContentsChanged
+    Private Sub Controls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrSaveLogical.ControlContentsChanged
```
N-thony, line 216: Only ucrSaveLogical?

N-thony (changes requested):

instat/dlgCompareColumns.vb
HawardKetoyoMsatsi: @N-thony Combining the two checks in one subroutine will not work in this case, the two checks need to work when separate events are called.

instat/dlgCompareColumns.vb
HawardKetoyoMsatsi: @N-thony this is now fixed please take a look.

N-thony: Great, this solves my comment.

N-thony:
@lloyddewit over to you now. Thanks.

lloyddewit (approved):

lloyddewit:
@rdstern since your last approval, there is one small change that should not affect the functionality. Please could you reapprove? thanks

rdstern (approved):
Very useful addition

instat/dlgCompareColumns.vb
shadrackkibet: @HawardKetoyoMsatsi I note that you ought to have used `.ControlValueChanged`  event and not `.ValueChanged` here.

shadrackkibet:
I did have a quick look and had a few designer points to report.

- Could you improve the design of this dialog? The top radio buttons are not well centred. See below
image

- Add colons(:)

instat/dlgCompareColumns.vb
shadrackkibet: I see also see a potential problem here; the check for the receivers should be done inside `CheckDatatype()` sub. Otherwise, if you run this sub elsewhere the message box is always shown. To check this just move from one radio button at the top when the receivers are empty you will always get the message which can be annoying to the user. Does this make sense to you?

shadrackkibet:
@HawardKetoyoMsatsi why do you need `ifelse()` here? Perhaps I am missing something?

`ifelse(test=abs(x=Dagoretti_JKIA.PRECIP - Dagoretti_JKIA.TMPMAX) <= 0, yes=TRUE, no=FALSE)`

Isn't this sufficient? because it always returns a logical value?

`abs(x=Dagoretti_JKIA.PRECIP - Dagoretti_JKIA.TMPMAX) <= 0`

instat/dlgCompareColumns.vb
lloyddewit: @shadrackkibet Thanks for the comment. I looked at the code, and indeed `ucrReceiver` objects normally handle the `ControlValueChanged` event rather than the `ValueChanged` event. I see that the `ValueChanged` event is a standard WinForms event, whereas the `ControlValueChanged` event is a `ucrCore` event (see screen shot below). The code above should work, but I understand that it is better to use handle the `ControlValueChanged` event because this is the standard way of handling this type of event from a `ucrCore` derived object. Is that correct?

image

shadrackkibet: @lloyddewit your statement is correct.

instat/dlgCompareColumns.vb
```diff
@@ -82,33 +104,76 @@ Public Class dlgCompareColumns
...
+        clsLessorEqualToOperator.AddParameter("first", clsRFunctionParameter:=clsAbsoluteFunction, iPosition:=0)
+        clsLessorEqualToOperator.AddParameter("tol", "0", iPosition:=1)
+
+        clsIfElseCompareFunction.SetRCommand("ifelse")
+        clsIfElseCompareFunction.AddParameter("test", clsROperatorParameter:=clsIsEqualToOperator, iPosition:=0)
+        clsIfElseCompareFunction.AddParameter("yes", "TRUE", iPosition:=1)
+        clsIfElseCompareFunction.AddParameter("no", "FALSE", iPosition:=2)
+
+        clsSummaryFunction.SetRCommand("summary")
+        clsSummaryFunction.AddParameter("x", clsRFunctionParameter:=clsIfElseCompareFunction, bIncludeArgumentName:=False, iPosition:=1)
```
shadrackkibet, line 160: Always use the correct parameter names. Here it should be `object` and then no need to hide the parameter.

instat/dlgCompareColumns.vb
```diff
@@ -82,33 +104,76 @@ Public Class dlgCompareColumns
...
+        clsAsCharacterFunctionOne = New RFunction
+        clsAsCharacterFunctionTwo = New RFunction
+        clsSummaryFunction = New RFunction
         clsYinXOperator = New ROperator
+        clsIsEqualToOperator = New ROperator
+        clsSubtractOperator = New ROperator
+        clsLessorEqualToOperator = New ROperator
 
         ucrBase.clsRsyntax.ClearCodes()
+        ucrInputTolerance.SetText("0")
```
shadrackkibet, line 129: This default should be set as a parameter value to the left of the operator and therefore no need to manually set it here.

instat/dlgCompareColumns.vb
```diff
@@ -119,13 +184,30 @@ Public Class dlgCompareColumns
...
+            If rdoByRow.Checked Then
+                If {"integer", "numeric"}.Contains(ucrReceiverFirst.strCurrDataType) AndAlso {"integer", "numeric"}.Contains(ucrReceiverSecond.strCurrDataType) Then
+                    ucrBase.OKEnabled(True)
+                ElseIf ucrReceiverFirst.strCurrDataType = "Date" AndAlso ucrReceiverSecond.strCurrDataType = "Date" Then
+                    ucrBase.OKEnabled(True)
+                ElseIf {"factor", "character"}.Contains(ucrReceiverFirst.strCurrDataType) AndAlso {"factor", "character"}.Contains(ucrReceiverSecond.strCurrDataType) Then
+                    ucrBase.OKEnabled(True)
+                ElseIf ucrReceiverFirst.strCurrDataType = "logical" AndAlso ucrReceiverSecond.strCurrDataType = "logical" Then
+                    ucrBase.OKEnabled(True)
+                Else
```
shadrackkibet, line 205: Simplify this by using an `OrElse` in the if statement?

instat/dlgCompareColumns.vb
```diff
@@ -119,13 +184,30 @@ Public Class dlgCompareColumns
...
+                    ucrBase.OKEnabled(False)
+                End If
+            Else
+                ucrBase.OKEnabled(True)
```
shadrackkibet, line 209: The `else` here means when the other radio button is checked Ok should be enabled as shown below. Which is not desired? By the way, look at the spacing of the checkboxes do you notice something?

image

instat/dlgCompareColumns.vb
```diff
@@ -141,26 +223,56 @@ Public Class dlgCompareColumns
...
-    Private Sub ucrReceiverSecond_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverSecond.ControlValueChanged
-        ' Needs to be done manually because data frame name should only be the data frame name from the second receiver's variable.
-        ucrSaveLogical.SetGlobalDataName(ucrReceiverSecond.GetDataName())
-    End Sub
-
-    Private Sub ucrSaveLogical_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSaveLogical.ControlValueChanged
-        If ucrSaveLogical.ucrChkSave.Checked Then
+    Private Sub ucrPnlOptions_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlOptions.ControlValueChanged
+        If rdoByValue.Checked Then
+            grpComparisions.Visible = True
```
shadrackkibet, line 228: This shouldn't be done manually here. Set it as a linked display control i.e `SetLinkedDisplayControl`

instat/dlgCompareColumns.vb
```diff
@@ -141,26 +223,56 @@ Public Class dlgCompareColumns
...
+                ElseIf ucrReceiverFirst.strCurrDataType = "logical" AndAlso ucrReceiverSecond.strCurrDataType = "logical" Then
+                    ucrInputTolerance.Visible = False
+                    clsIfElseCompareFunction.AddParameter("test", clsROperatorParameter:=clsIsEqualToOperator, iPosition:=0)
+                Else
+                    MsgBox("Receivers must have the same data type, OK will not be enabled")
+                    ucrInputTolerance.Visible = False
+                End If
             Else
-                ucrSaveLogical.SetName("in_" & ucrReceiverFirst.GetVariableNames(False))
+                ucrInputTolerance.Visible = False
```
shadrackkibet, line 261: Check if this is the correct `else`.

lloyddewit:
@shadrackkibet Thank you for your review.
@HawardKetoyoMsatsi This PR is now merged so I assume you'll address @shadrackkibet 's comments in a new PR. Is that correct?

rdstern:
I would be happy for @HawardKetoyoMsatsi to address these useful, points from @shadrackkibet .

Alternatively there is a task on quite a number of dialogues, concerned with the Save control and I can include it there.  I'll add the buttons at the top at the same time.  There is also quite a lot of space in the save control, between the label and the next field.  I am waiting for the French version, before commenting on that aspect.

HawardKetoyoMsatsi:
@lloyddewit Sure, I will address the issues in a new PR. I will however wait for @rdstern's comment concerning issues with the Save control in other dialogs I work on all of them at the same time.

rdstern:
I have included this in #6397 and have now added @HawardKetoyoMsatsi to that issue.  It is waiting for @Patowhiz and I would be happy for Haward to join me in reminding Patrick about this work.  I had hoped it would be ready for Ivan and Haward to take over by now.

shadrackkibet:
I would suggest that we consider the changes here as separate from issues in the save control. The change in the save control here is specific to this dialog which can just be fixed from the design. Therefore, I would expect the below change to be included with the changes suggested above.

>a) Check Data > Compare Columns: (closed pull request #6414 by Shadrack:) So centre the buttons at the top, add a colon in the position control label, sort out Alignment of the position control button and check that the default is to put the new (logical) variable after the second column. If easy (and only if easy) note the blank space, where Shadrack suggests the size of the dialogue might change with the setting of the top button.

rdstern:
@shadrackkibet I would be very happy with that.  Let's do it!  Can you check with @HawardKetoyoMsatsi if he or you or @Ivanluv  can do it?

shadrackkibet:
@rdstern from our quick discussion here we have agreed as follows;

- @HawardKetoyoMsatsi will work on reviews here and do the changes on the save control that are specific to compare dialog.
- @Ivanluv will look into issue #6397 excluding Compare dialog and fix issues reported on the save control from various dialogs.
-  I will support and review their work

instat/dlgCompareColumns.vb
HawardKetoyoMsatsi: The final commit that was merged had this piece of code updated.

shadrackkibet: That is right, the question here is to check if `ucrInputTolerance.Visible = False` is the correct as the else

PR #6428, opened 2021-04-26
Added missing value options into insert column(s)

### LINKED ISSUE #6424: Insert columns needs simple correction
In the Prepare > Data Frame > Insert Columns/Rows when inserting columns the code generated is as follows:

```
# Code generated by the dialog, Insert Columns/Rows

data_book$add_columns_to_data(data_name="fairview", col_data="NA", col_name="X", use_col_name_as_prefix=TRUE, before=FALSE)
rm()
```
The quotes round NA is a bad idea.  This enters the string NA not the missing value.  If you change to 0 say, then it enters "0".  So, please get rid of the quotes.

The one problem is that you may want to enter a string.  So also add a tooltip with the default value field, `Include the quotes if the default value is a string, e.g. "small"`

Ivanluv:
Fixes #6424


rdstern (approved):
Great

rdstern (dismissed):
Great

instat/dlgInsertColumn.vb
```diff
@@ -127,6 +130,8 @@ Public Class dlgInsertColumn
         clsInsertColumnFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$add_columns_to_data")
         clsInsertColumnFunction.AddParameter("use_col_name_as_prefix", "TRUE")
         clsInsertColumnFunction.AddParameter("before", "FALSE")
+        clsInsertColumnFunction.AddParameter("col_data", "NA")
+
```
shadrackkibet, line 134: Add parameter positions?

shadrackkibet, line 134: Add parameter positions?

shadrackkibet (changes requested):

instat/dlgInsertColumn.vb
```diff
@@ -124,18 +124,16 @@ Public Class dlgInsertColumn
         ucrInputBeforeAfter.Reset()
         ucrInputDefaultValue.Reset()
         ucrInputPrefixForNewColumn.Reset()
-        clsInsertColumnFunction.AddParameter("col_data", "NA")
-        clsInsertColumnFunction.AddParameter("col_name", "X")
 
         clsInsertColumnFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$add_columns_to_data")
         clsInsertColumnFunction.AddParameter("use_col_name_as_prefix", "TRUE")
```
shadrackkibet, line 129: Still missing position here?

shadrackkibet (changes requested):

shadrackkibet:
@rdstern do we want the inserted column to be of class Logical by default? I thought we want numeric class by default.

rdstern:
@shadrackkibet I had noticed that.  But if you type a number into the logical (all NA) variable, then it automatically becomes numeric.  So I thought that was OK?

shadrackkibet:
In R we have different `NA` types i.e `NA_integer_, NA_real_, NA_complex_, and NA_character`. If we want to have a numeric class by default then we need to use `NA_real_` as the default. Will this be confusing to users?

rdstern:
I'd be quite happy with that.  @dannyparsons I don't see any downside to having NA_real_ as the default?

rdstern:
Can I check this?  It would be good to get it merged today if possible.

Ivanluv:
@rdstern  yes you can check

rdstern (changes requested):
@Ivanluv this works fine and good to get into the next version.
It is a bit confusing that the default value is given in the dialogue as NA_real.  Could that still be shown as NA?

rdstern (commented):
@Ivanluv that's fine by me, because I wasn't so worried before.  But it now generates a logical column again!  I thought the idea was that you, would make it numeric.
If you would like to be more dramtic, then you could replace the control by a pull-down into which you could optionally type.  If so, then the default is still NA, perhaps but could be NA_num as you had before.  Then there could be NA, which generates logical and possibly also TRUE, FALSE "small" "Mali" as others.

shadrackkibet:
A workaround would be to use `as.numeric(NA)`. Make `as.numeric()` and Rfunction and then pass it into the insert column function. Then the default shown in control will remain as `NA`.

instat/dlgInsertColumn.vb
```diff
@@ -111,26 +112,32 @@ Public Class dlgInsertColumn
...
 
         ucrSelectorInsertColumns.Reset()
         ucrInputBeforeAfter.Reset()
         ucrInputDefaultValue.Reset()
         ucrInputPrefixForNewColumn.Reset()
-        clsInsertColumnFunction.AddParameter("col_data", "NA")
-        clsInsertColumnFunction.AddParameter("col_name", "X")
+
+        clsAsNumericFunction.SetRCommand("as.numeric")
+        clsAsNumericFunction.AddParameter("col_data", "NA", bIncludeArgumentName:=False, iPosition:=0)
```
shadrackkibet, line 130: Always use the correct parameter name. In this case, it should be x and also for the control above. Then there is no need to hide it.

shadrackkibet (changes requested):

shadrackkibet:
@Ivanluv I notice I might have misled you in this. Now everything that is imputed into the input control will be coarsed to numeric including characters etc. Therefore I suggest we go by Roger's suggestion above. Really sorry for that!

shadrackkibet:
Then you need to ensure that for characters you add "" and for numeric and logical no "".

instat/dlgInsertColumn.vb
```diff
@@ -74,9 +74,10 @@ Public Class dlgInsertColumn
         ucrNudNumberOfColumns.SetRDefault(1)
         ucrNudNumberOfColumns.SetMinMax(1, Integer.MaxValue)
 
-        ucrInputDefaultValue.SetParameter(New RParameter("x", 0))
+        ucrInputDefaultValue.SetParameter(New RParameter("col_data", 5))
+        ucrInputDefaultValue.SetItems({"NA", "NA_real_"}, bAddConditions:=True, bAddQuotes:=False)
```
shadrackkibet, line 78: I think there is no harm in adding the other types as well i.e `NA_integer_, NA_complex_` and `NA_character` @rdstern to confirm.
@Ivanluv I think `bAddConditions:=True, bAddQuotes:=False` are redundant here because of two lines(79,80) below?

rdstern:
Yes, I agree.  Once there is the pull-down it would be fun to create variables of different types.

shadrackkibet:
@rdstern this looks good. If you approve then we can merge.

rdstern (approved):
Looks good.  Perhaps @shadrackkibet can also approve and merge?

instat/dlgInsertColumn.vb
```diff
@@ -111,6 +113,8 @@ Public Class dlgInsertColumn
         ucrNudNumberOfRows.SetLinkedDisplayControl(lblNumberOfRowsToInsert)
         ucrNudStartRow.SetLinkedDisplayControl(lblStartPos)
         ucrReceiverColumnsToInsert.SetLinkedDisplayControl(lblColumn)
+
+        ttColumnDefaultValue.SetToolTip(ucrInputDefaultValue.cboInput, "Include the quotes if the default value is a string, e.g. ""small"".Do not include qoutes if its logical, e.g. TRUE")
```
shadrackkibet, line 117:
```suggestion
        ttColumnDefaultValue.SetToolTip(ucrInputDefaultValue.cboInput, "Include the quotes if the default value is a string e.g. "small". Do not include quotes if the default value is logical e.g. TRUE)

```

instat/dlgInsertColumn.vb
```diff
@@ -114,7 +114,8 @@ Public Class dlgInsertColumn
         ucrNudStartRow.SetLinkedDisplayControl(lblStartPos)
         ucrReceiverColumnsToInsert.SetLinkedDisplayControl(lblColumn)
 
-        ttColumnDefaultValue.SetToolTip(ucrInputDefaultValue.cboInput, "Include the quotes if the default value is a string, e.g. ""small"".Do not include qoutes if its logical, e.g. TRUE")
```
Ivanluv: @shadrackkibet  the double  quotes before and after small  place the string small within the whole string

shadrackkibet: @Ivanluv sorry my suggestion was not correct. Please ignore it.

shadrackkibet (approved):
@rdstern please re-approve then we can merge.

shadrackkibet:
@rdstern nothing changed since you last approved. I had a minor suggestion that was wrong.

rdstern (approved):

PR #6399, opened 2021-04-08
Enhancement of the Climatic Data Entry Dialog

N-thony:
@rdstern this solves the following items in issue #6352 and ready for your test:

1. Added the Options sub-dialogue.
2. Remain in the offending cell when an item is flagged and not possible, e.g. a decimal or typing text
3. Added Extra Rows checkbox
4. Summary of the number of observations shown in the main dialog
5. Added a button Comment
6. Added flag field data checkbox
Other items are still under implementation.

PR #6429, opened 2021-04-26
Bug fixes: Climatic data entry

shadrackkibet:


shadrackkibet:
@rdstern could you please test? There shouldn't be any functionality changes in the dialog. This is just tidying code and fixing bugs in the current version. I would like to merge this before I work on adding flags.

@lloyddewit I welcome you to peer review this. 

I noted some bugs and deviation in the way we set up sub dialogs in PR #6399 which I thought will be quick for me to fix. For learning purposes I would encourage @N-thony to look into changes here and ask questions.

Summary of some of the changes made:

- Setting Rfunctions in the main dialog and then passing them into subdialogs
- Proper resetting of subdialogs i.e have a boolean from the main dialog that is passed into `.setRcode` in the subdialog
- Visibility of tooltips i.e setting it into the parent control
- Restricted comments subdialog data frame selector
- Use of linking system to link controls i.e `AddToLinkedControls`
- Corrected `get_keys` function. We don't have to fix the key name

rdstern (dismissed):
@shadrackkibet it still looks fine.
@lloyddewit could you confirm and merge if ok.  It will then be opened again immediately for further work

lloyddewit:
@rdstern @shadrackkibet yes, I'm happy to review

instat/sdgClimaticDataEntryOptions.vb
```diff
@@ -14,29 +14,26 @@
...
     End Sub
+
     Private Sub InitialiseControls()
         ucrChkTransform.SetText("Transform:")
-        ttucrChkTransform.SetToolTip(ucrChkTransform, "The values written to the data frame are transformed, usually multiplied, by the value given here.")
+        ttucrChkTransform.SetToolTip(ucrChkTransform.chkCheck, "The values written to the data frame are transformed, usually multiplied, by the value given here.")
 
         ucrInputTransform.SetItems({"0.1", "25.4", "0.254"})
-        ucrInputTransform.SetValidationTypeAsNumeric(dcmMin:=0.1)
+        ucrInputTransform.SetValidationTypeAsNumeric()
```
lloyddewit, line 31: What was the reason for removing the minimum valid number?

instat/sdgClimaticDataEntryOptions.vb
```diff
@@ -106,8 +105,7 @@ Public Class sdgClimaticDataEntryOptions
 
     Public ReadOnly Property TransformValue As Double
         Get
-            'todo. do explicit conversion. Currently ucrInputTransform validation type is number
-            Return If(String.IsNullOrEmpty(ucrInputTransform.GetValue), 0, ucrInputTransform.GetValue)
+            Return ucrInputTransform.GetValue
```
lloyddewit, line 108: Would it improve fault tolerance if we did a `TryCast` here to ensure that the `GetValue` object can safely be cast to a double?

instat/sdgCommentForDataEntry.vb
```diff
@@ -17,36 +17,23 @@
...
-    Private bFirstLoad As Boolean = True
-    Private bReset As Boolean = True
+    Private ucrBaseSelector As ucrSelector
+    Private bControlsInitialised As Boolean = False
     Private bClearComments As Boolean
     Public clsList As RFunction
     Private clsCommentsList As RFunction
     Private clsSaveDataEntry As RFunction
-    Private clsGetKey As RFunction
+    Private clsGetKey As New RFunction
```
lloyddewit, line 29: Is the `New` consistent with the other subdialogs?

instat/sdgCommentForDataEntry.vb
```diff
@@ -124,34 +84,50 @@ Public Class sdgCommentForDataEntry
...
+        If ucrBaseSelector IsNot Nothing AndAlso ucrBaseSelector.strCurrentDataFrame <> "" Then
+            strSelectedDataFrame = ucrBaseSelector.strCurrentDataFrame
+            ucrSelectorAddComment.SetDataframe(strSelectedDataFrame, False)
+            clsGetKey.AddParameter("data_name", Chr(34) & strSelectedDataFrame & Chr(34), iPosition:=0)
+        Else
+            clsGetKey.RemoveParameterByName("data_name")
+        End If
+
+        clsSaveDataEntry = clsNewSaveDataEntry
+        clsGetKey = clsNewGetKey
```
lloyddewit, line 107: Will this cancel the effect of lines 101 and 103?

lloyddewit (changes requested):
@shadrackkibet Thanks, this looks good.
Please could you look at the comment in `sdgCommentForDataEntry.vb`?
My other comments are less important. I understood that this PR was also about showing good practice. So I used the opportunity to ask your opinion on a couple of things.

instat/dlgClimaticDataEntry.vb
```diff
@@ -23,13 +23,17 @@ Public Class dlgClimaticDataEntry
     Private clsSaveDataEntry As New RFunction
     Private clsEditDataFrame As New RFunction
     Private clsGetDataEntry As New RFunction
+    Private clsCommentsList As New RFunction
+    Private clsList As New RFunction
+    Private clsGetKey As New RFunction
```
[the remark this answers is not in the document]
shadrackkibet: I like adding `*Function` as well. Generally in the software, we have a mixture of the two. In my opinion, I think we should encourage adding `*Function` and `*Operator` at the end. Therefore I will change the naming here.

instat/dlgClimaticDataEntry.vb
```diff
@@ -217,20 +233,23 @@ Public Class dlgClimaticDataEntry
...
                                            dTranformValue:=sdgClimaticDataEntryOptions.TransformValue,
                                            MissingValueAsNA:=sdgClimaticDataEntryOptions.MissingValueAsNA,
-                                           strEntryType:=ucrInputType.GetText)
+                                           strEntryType:=ucrInputType.GetText, ucrNewBaseSelector:=ucrSelectorClimaticDataEntry, bReset:=bResetSubdialogs)
             End If
             sdgClimaticDataEntry.ShowDialog()
+            bResetSubdialogs = False
             SetNumberRowsChangedText(sdgClimaticDataEntry.NRowsChanged)
-            SetNumberCommentEnteredText(sdgCommentForDataEntry.NbCommentEntered)
+            SetNumberCommentEnteredText(sdgCommentForDataEntry.GetSetNumberOfCommentsEntered)
```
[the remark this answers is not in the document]
shadrackkibet: The visual basic naming convention encourages capitalization of the first letter when naming elements [here](https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/program-structure/naming-conventions) Generally in the software, we don't follow this. Checking elsewhere for example `ucrInputComboBox.vb` line 164 I find that we have used `GetSet...` therefore, in this case, I choose to remain with this.
image

instat/sdgCommentForDataEntry.vb
shadrackkibet: This is another place where we are not consistent generally. 

The reason why I added this is because I get an exception in line 129 because this is not declared as a `New` instance. I had chosen this fix but since it is not consistent I will remove `New` here and instead add a check in line129.

instat/sdgClimaticDataEntryOptions.vb
shadrackkibet: I think you meant `TryParse`? If this is correct then I agree with you.

instat/sdgClimaticDataEntryOptions.vb
shadrackkibet: I don't think the minimum transformation values has to be `0.1` a user might want to use any numeric number.

instat/sdgClimaticDataEntryOptions.vb
shadrackkibet: @lloyddewit I thought of changing this into a function as follows;
```
    Public Function TransformValue() As Object
        Dim dTemp As Double
        If Double.TryParse(ucrInputTransform.GetValue, dTemp) Then
            Return ucrInputTransform.GetValue
        Else
            Return CDbl(ucrInputTransform.GetValue)
        End If
    End Function
```
What do you think?

instat/sdgCommentForDataEntry.vb
shadrackkibet: Yes, It should be just above line 101 and 103

instat/dlgClimaticDataEntry.vb
```diff
@@ -112,6 +116,9 @@ Public Class dlgClimaticDataEntry
         clsSaveDataEntry = New RFunction
         clsEditDataFrame = New RFunction
         clsGetDataEntry = New RFunction
+        clsCommentsList = New RFunction
+        clsList.Clear()
```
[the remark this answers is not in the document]
shadrackkibet: No reason. Either work in this case. I will change for consistency.

rdstern:
@shadrackkibet you are right.  If in inches then 0.01 might be used, so smaller numbers shouldn't be excluded.  0.1 is the obviou value if data are in mms or degrees c, where measured to one decimal place.  Then you may want to enter (say) 12 for 1.2mm and not bother with the decimal point

instat/sdgClimaticDataEntryOptions.vb
lloyddewit: Yes, you're right `TryParse` is the correct function' I liked your function. For me, keeping it as a readonly property is also fine. I think if the `TryParse` fails then we should raise a developer error rather than trying `CDbl`. How about something like below?
```
    Public ReadOnly Property GetTransformValue As Double
        Get
            Dim dTemp As Double
            If Not Double.TryParse(ucrInputTransform.GetValue, dTemp) Then 'If TryParse fails then sets dTemp to zero
                MsgBox("Developer error: The ucrInputTransform combo box contains the value '" & ucrInputTransform.GetValue & "'. Cannot convert this to a double.")
            End If
            Return dTemp 
        End Get
    End Property
```

lloyddewit:
@shadrackkibet Thank you for the responses. All comments are resolved apart from one (please see above), thanks

instat/sdgClimaticDataEntryOptions.vb
shadrackkibet: Agreed

shadrackkibet:
> @shadrackkibet Thank you for the responses. All comments are resolved apart from one (please see above), thanks

@lloyddewit I resolved it now. Thanks

lloyddewit (approved):

lloyddewit:
@shadrackkibet Thanks for the updates, it all looks good.
@rdstern Please could you approve again? thanks

rdstern (approved):

PR #6430, opened 2021-04-27
Adding "link" as the default "link_name" in the Add link dialogue

### LINKED ISSUE #6350: Small changes in the Add link dialogue
The Add Key dialogue now provides a small summary in the output Window.  

Here there are 2 small items

a) Suggest a name as we do in other dialogues - `link` would be fine
b) As with the Add Key, it would be good to provide some output.  Possibly just 
`name`  variable in `name of data frame1` linked to `name `  variable in `name of data frame2 `
For example `station variable in data frame:germany_data linked to station variable in data frame germany_metadata` 

image

@dannyparsons to check quickly.  It would slightly simplify the import data from Germany video!

Ivanluv:
fixes #6350 


shadrackkibet:
@Ivanluv no changes here?

shadrackkibet:
@Ivanluv does this index automatically? i.e link1, link2, link3 etc?

Ivanluv:
@shadrackkibet  it does not index automatically

shadrackkibet:
We need to check with Danny/David first then I think we will need to use the save control here which I think we need to be adapt it to save keys and link.

shadrackkibet:
@Ivanluv what is the state of this PR? Any suggestions on how we can get this done?

Ivanluv:
@shadrackkibet  I have used the save control in saving the keys. Could you check PR #6590 first then if we come up with a better implementation I might do that for the links as well

shadrackkibet:
Great I will have a look at it. That means we close this PR right? Please reopen if you think otherwise.

Issue #3433, opened 2017-06-18
import .DAT files


dannyparsons:
Currently cannot import .DAT files as not supported by `rio`. If `rio` do not add this before 0.9 (from my suggestion) then we should do this ourselves using `data.table::fread`.

dannyparsons:
@stevenndungu I now have a fix for this using `format = "txt"`. I can show you were to add this when you get there.

stevenndungu:
@dannyparsons  looking at this now

dannyparsons:
fixed by #3463

Patowhiz:
@dannyparsons is there a reason why `format = "csv"` was not used for `dat` files. I checked and found that it `rio` reads them fine if you force it to read the files as `csv`. This issue now relates to issue #6431 and PR #6432.
@rdstern would like the `csv` options enabled for the `dat` files but I need to understand why `txt` was the preferred option.

rdstern:
@Patowhiz I may just be repeating your question.  In the particular example here, the files are actually comma separated.  I wouldn't have a problem with txt, but I just want to be able to have some flexibility e.g. in missing values.  I get nothing just now, as you see above.

Patowhiz:
@rdstern yes. And I would prefer them read as csv if there is no special reason for reading the as txt. Because the implementation will be simple. 
We have 3 options for reading text files, the 2 out of 3 use different package for reading the files. So reading dat as text files would force me to  also design the code in a way that it considers the 3 options.

@rdstern after looking at the dat files that you have sent me, they are separated by white spaces.

![data file](https://user-images.githubusercontent.com/21179007/116378615-2968da00-a7c7-11eb-9f88-f323ebaff2c2.PNG)

Patowhiz:
@rdstern @dannyparsons I also noticed, the .dat file is not included in the file open dialog. Is there any special reason for this? See screenshot below.
![file types](https://user-images.githubusercontent.com/21179007/116378765-456c7b80-a7c7-11eb-85dc-5fd62d1b02c1.png)

Patowhiz:
As agreed by @dannyparsons I have implemented this as text file.

PR #6432, opened 2021-04-28
Added import options for .dat file types

### LINKED ISSUE #6431: Importing multiple DAT file types
@rdstern says

> I think there is another issue on the File > Open.  We have data from 153 stations.  The files are actually csv files, but they have the extension dat.  Reading in as dat, you don't get the options to change the missing value code.  I thought things like this were always assumed to possibly read in as csv?  If not, then could you add dat files to be csv ones.

Patowhiz:
Fixes #6431.


Patowhiz:
I have added import options for DAT files. The DAT files will be read as csv when importing multiple files. The options are not available for single files because this was not previously implemented.

lloyddewit:
@Patowhiz Thank you for working on this issue. 
I understand that this PR:
- only affects the behaviour when all the files in a directory are imported
- assumes that any `.dat` files in the directory are CSV formatted

Some questions:
- can we always assume that `.dat` files are CSV?
- what is the reason not to update the functionality for importing single files?
- for single file import, the code processes `.dly` files the same as `.csv` files; but it processes `.dat` files in a separate way. Is this correct?

Thanks for clarifying

rdstern:
It is still not quite what is needed:  Here is the screen with a single file:
image

Now I need to change the default missing value code to -99999.  It is actually a csv file so this was all possible when I renamed the file to change the extension to csv.

This is being use for different countries on our AIMS Cameroon course.  Of course they can all change the extension, so I am seizing this opportunity to add to the flexibility of importing.

Patowhiz:
> @Patowhiz Thank you for working on this issue.
> I understand that this PR:
> 
> * only affects the behaviour when all the files in a directory are imported
> * assumes that any `.dat` files in the directory are CSV formatted
> 
> Some questions:
> 
> * can we always assume that `.dat` files are CSV?
> * what is the reason not to update the functionality for importing single files?
> * for single file import, the code processes `.dly` files the same as `.csv` files; but it processes `.dat` files in a separate way. Is this correct?
> 
> Thanks for clarifying

@lloyddewit thanks for the questions.

> can we always assume that .dat files are CSV?
 yes we can assume that .data files are csv.

> what is the reason not to update the functionality for importing single files?
I see no reason, I think this is a good feature that went unnoticed. @rdstern recommends to have it done.

> for single file import, the code processes .dly files the same as .csv files; but it processes .dat files in a separate way. Is this correct?
yes, that is what is happening right now. I think this was done due to issues raised in closed issue #3433 back in 2017. I think it was anticipated that in future a reading .dat files will be implemented differently, though I'm not sure.

@dannyparsons has convinced me why its preferable to read the `.dat `files as `.txt` files. So I'll go ahead and implement this for single and multiple .dat files and also include its file type in the open file dialog. I'm also enabling the import options.

Patowhiz:
@rdstern this is now ready for you.

rdstern (approved):
@Patowhiz great
@lloyddewit over to you.

lloyddewit (approved):

PR #6436, opened 2021-05-01
Adding flag fields on climatic data entry

### LINKED ISSUE #6394: Adding the writing of flag fields to the new data entry system
This is an important part of the data entry system.  It is to add "flag fields" to the data files as well as adding the data.  The main use will be to be able to kn ow which rows have been updated in the data, and hence to be able to selct those rows/dates for export  (to climsoft)

It is possible they can be for other purposes too, once the facility is done.  We thought this would have to be @dannyparsons but now suggest that @N-thony could set up the checkbox, and then @shadrackkibet would be able to do this, with guidance from @volloholic .  I hope that's ok?  By "do this" I hope that Shadrack would be able to write the R-code for this component.  Then @Patowhiz and @N-thony can include it within the data entry system.

This was item 14 in the improvements to the data entry system.  It is copied below here:

(Antoine/Danny) Add a checkbox in the sub-dialogue, default unchecked. Label is Add flag field data. Currently this is just to add the the saving code (Danny to do?) when Ok is pressed. It does the following:
a) If not present, then it creates those variables. If a variable is tmax, then its name could be tmax_flag. It is a factor variable. Ideally perhaps it has 3 levels, plus NA. It is NA when tmax is originally NA, data, where there are data that are unchanged, edit, when data have been changed, and add, when NAs have become data, and changed, when an add row is edited. Initially all we need is to distinguish between unchanged and changed, so we can filter on the changed values to write back to Climsoft.)
b) Once the flag variable is present it just writes back to it, similarly to writing to the data.
c) As I write this I wonder whether it could be simpler if the flag variables are present in the grid, possibly after all the other variables. They are visible, but are completed automatically?
d) Maybe it would be simpler if the flag fields were made, before editing starts if they are going to be used? Perhaps when the box it ticked it could check that the corresponding flag fields exist. There could be a button to Make flag variables which is disabled if they already exist and to that effect. This would add the flag variable(s)
e) So the R-code can assume the flag variables already exist.  I suggest there should be a separate (Setup) dialogue to produce these variables.  And they would be as follows:
1) They are called rain_fl, tmax_fl, etc.  
2) They are factor variables with 4 defined levels.  0) NA when the data variable is NA. 1) `data` when the variable has data. 2) `edit`, when an existing value is changed, 3) `entry` when a new value is entered 4) `change` when a new value is entered and then changed.  
3) So, to start with a flag variable is either all NA, or a mixture of `data and NA`.  If any data values are changed, then they become `edit` - they become edit if they are "changed", even if the actual value doesn't alter.  If an edit value is changed again, it stays as edit.  If an NA is changed, then it becomes `add`.    I am currently suggesting that if edit is changed to NA, or add is changed back to NA, they are still noted as edit or add.

### LINKED ISSUE #6437: Bug when adding a comment in data entry dialog
**Describe the bug**
image
The first one occurs when for the first time you select a row, and click on add comment. There is no row number in the input row, you should close and opened the comment sub dialog again to have the row number of the selected row in the grid.
image
This one is due by the fact that the add comment button is supposed to be disable if there is no key in the data frame when trying to add a comment, but here it is enable even if there is no key in a data frame that allow the add comment function to run.

shadrackkibet:
fixes  #6394 and fixes #6437

@rdstern this is ready for you to test. Thanks

### DIFF
--- a/instat/sdgClimaticDataEntryOptions.vb
+++ b/instat/sdgClimaticDataEntryOptions.vb
@@ -62,8 +62,6 @@ Public Class sdgClimaticDataEntryOptions
             ucrChkAllowTrace.Checked = False
             ucrChkTransform.Checked = False
             ucrChkNoDecimal.Checked = False
-            ucrInputDefaultValue.SetText(0)
-            ucrInputTransform.SetText(0.1)
         End If
     End Sub
 
@@ -85,7 +83,7 @@ Public Class sdgClimaticDataEntryOptions
         End Get
     End Property
 
-    Public ReadOnly Property GetDefaultValue As Double
+    Public Property GetSetDefaultValue As Double
         Get
             Dim dTemp As Double
             If Not Double.TryParse(ucrInputDefaultValue.GetValue, dTemp) Then
@@ -93,6 +91,9 @@ Public Class sdgClimaticDataEntryOptions
             End If
             Return dTemp
         End Get
+        Set(value As Double)
+            ucrInputDefaultValue.SetText(value)
+        End Set
     End Property
 
     Public ReadOnly Property AllowTrace As Boolean
@@ -107,7 +108,7 @@ Public Class sdgClimaticDataEntryOptions
         End Get
     End Property
 
-    Public ReadOnly Property GetTransformValue As Double
+    Public Property GetSetTransformValue As Double
         Get
             Dim dTemp As Double
             If Not Double.TryParse(ucrInputTransform.GetValue, dTemp) Then
@@ -115,6 +116,9 @@ Public Class sdgClimaticDataEntryOptions
             End If
             Return dTemp
         End Get
+        Set(value As Double)
+            ucrInputTransform.SetText(value)
+        End Set
     End Property
 
     Private Sub ucrChkMissingValues_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkMissingValues.ControlValueChanged
--- a/instat/sdgClimaticDataEntryOptions.vb
+++ b/instat/sdgClimaticDataEntryOptions.vb
@@ -71,10 +71,13 @@ Public Class sdgClimaticDataEntryOptions
         End Get
     End Property
 
-    Public ReadOnly Property UseDefault As Boolean
+    Public Property GetSetDefaultCheckboxState As Boolean
         Get
             Return ucrChkDefaultValue.Checked
         End Get
+        Set(value As Boolean)
+            ucrChkDefaultValue.Checked = value
+        End Set
     End Property
 
     Public ReadOnly Property MissingValueAsNA As Boolean
--- a/instat/sdgCommentForDataEntry.vb
+++ b/instat/sdgCommentForDataEntry.vb
@@ -18,8 +18,6 @@ Imports instat.Translations
 Imports RDotNet
 Public Class sdgCommentForDataEntry
     Private strSelectedDataFrame As String = ""
-    Private strSelectedRow As String = ""
-    Private strSelectedColumn As String = ""
     Private ucrBaseSelector As ucrSelector
     Private bControlsInitialised As Boolean = False
     Private bClearComments As Boolean
@@ -71,21 +69,6 @@ Public Class sdgCommentForDataEntry
         ttcmdAddComment.SetToolTip(ucrInputComment.txtInput, "Add Comment button is enabled when the key exists in the data frame and the input comment is not empty.")
     End Sub
 
-    Private Sub SetDefaultPosition()
-        ucrInputRow.SetName(strSelectedRow)
-        If strSelectedColumn <> "" AndAlso strSelectedRow <> "" Then
-            ucrReceiverColumn.Add(strSelectedColumn, strSelectedDataFrame)
-            rdoCell.Checked = True
-        ElseIf strSelectedColumn <> "" AndAlso strSelectedRow = "" Then
-            ucrReceiverColumn.Add(strSelectedColumn, strSelectedDataFrame)
-            rdoColumn.Checked = True
-        ElseIf strSelectedColumn = "" AndAlso strSelectedRow <> "" Then
-            rdoRow.Checked = True
-        Else
-            rdoDataFrame.Checked = True
-        End If
-    End Sub
-
     Public Sub SetUpCommentsSubdialog(clsNewSaveDataEntry As RFunction, clsNewGetKey As RFunction,
                       clsNewCommentsList As RFunction, clsNewList As RFunction, strDataFrame As String,
                                       Optional strRow As String = "", Optional strColumn As String = "", Optional ucrNewBaseSelector As ucrSelector = Nothing, Optional bReset As Boolean = False)
@@ -108,15 +91,23 @@ Public Class sdgCommentForDataEntry
             clsGetKeyFunction.RemoveParameterByName("data_name")
         End If
 
-        strSelectedRow = strRow
-        strSelectedColumn = strColumn
-
-        SetDefaultPosition()
-
         ucrSelectorAddComment.SetRCode(clsCommentsListFunction, bReset, bCloneIfNeeded:=True)
         ucrInputComment.SetRCode(clsCommentsListFunction, bReset, bCloneIfNeeded:=True)
         ucrInputRow.SetRCode(clsCommentsListFunction, bReset, bCloneIfNeeded:=True)
         ucrReceiverColumn.SetRCode(clsCommentsListFunction, bReset, bCloneIfNeeded:=True)
+
+        ucrInputRow.SetName(strRow)
+        If strColumn <> "" AndAlso strRow <> "" Then
+            ucrReceiverColumn.Add(strColumn, strSelectedDataFrame)
+            rdoCell.Checked = True
+        ElseIf strColumn <> "" AndAlso strRow = "" Then
+            ucrReceiverColumn.Add(strColumn, strSelectedDataFrame)
+            rdoColumn.Checked = True
+        ElseIf strColumn = "" AndAlso strRow <> "" Then
+            rdoRow.Checked = True
+        Else
+            rdoDataFrame.Checked = True
+        End If
 
         If bReset Then
             EnableDisableAddComment()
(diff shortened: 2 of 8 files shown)


rdstern:
I have now checked the problem I found more carefully.  I think it is quite small.
[fairview.zip](https://github.com/africanmathsinitiative/R-Instat/files/6410062/fairview.zip)
It is with the fairview data that just contains rainfall.
I add the flag field and then start the data entry.
However there is no rain from July 1 2014 to October 31 2014.  So I change to range and choose those 4 months  - I use Range rather than month.  I also have a default value of zero (0).  
It works fine and updates.  I chose the option to not save the flag fields.  That is all ok.
Then I return and go back into the data again.  It is still the default value - pale yellow.  
a) I now set to save the flag fields.  It does this and correctly calls them data, because they were there before.
b) But, they should not be pale yellow now, nor be entered.  The default value is only inserted as zero when there is NA.  Here there is now zero, so the default is not used.  Hence no new values are entered.  

That's the problem.  I think it should be easy - I hope - to fix.

shadrackkibet:
@rdstern that is because the default value checkbox is still checked. So you should uncheck the checkbox once the values are infilled. We can uncheck it once the user clicks Ok and therefore they will not appear as pale yellow despite having changed. Would you like that?

rdstern:
@shadrackkibet but ther default value is supposed to replace the NA by the default (0).  When it chenged NA into 0 it colours it a pale yellow.  Even if you leave it ticked, there are no NA values when you return to the same data.  You detect this, with the value in the flag field and the default should also.  So nothing should be yellow, or marked as changed the second time round.

rdstern (commented):
@shadrackkibet I think your part seems to be working fine.  Great.
I tried with the Dodoma data and including adding and changing.  It seems to work.
The only problem is that it is quite slow now and this is just for a single station?  

But the bad news is that other parts may have regressed!  In particular the default value may not work as well as before.  I will test further and report separately.

shadrackkibet:
@rdstern I have just fixed the default value. Could you please re-test now?

shadrackkibet:
@rdstern I expected it to be a bit slow when adding flags. That is because I have a number of loops inside the R function, I will try to optimize it before release and if not I think we will have to consider it as a separate improvement. However, I don't expect it to be slow with the previous functionality i.e without adding flags.

rdstern (approved):
Now seems to work well!.  Great

shadrackkibet:
@rdstern I noted a bug in the metadata. The variable names didn't update correctly and therefore the columns were not visible in the filter subdialog where we want to use the flag columns!

Note that I have also optimized adding of flags. It should be substantially fast now. There is a risk of regression on flag fields with this change. Please check that it works as expected. I did check and seems to work!

rdstern:
@shadrackkibet it is much faster.  So that's great.  
a) One small thing is that you have the add flag fields checkbox in the data entry.  I suggest the default - when there are the flag fields - is for this to be checked.
b) Then I got this bug.  It may have been there before, and not noticed.
image

I suspect something is not clearing, because at the bottom it mentions November and 10 days.  That's what I updated earlier for Dodoma.  Now I was using the small new dataset to test the default of zero and that's when the error occurred.

The commands generated were as follows:
image

I tried the same after pressing Reset.  Then it worked fine with the following code instead:
image

shadrackkibet:
@rdstern I have fixed the bug you reported above. 

> a) One small thing is that you have the add flag fields checkbox in the data entry. I suggest the default - when there are the flag fields - is for this to be checked.

We need an R function to be able to implement this or we use column names from the selector.  I also notice that we need to check if the flags fields exist when creating flags using the setup for data entry dialog. When this PR is merged the user can easily override the flag fields using the setup for data entry without noticing (very dangerous!).  Therefore let's open a separate issue with both of these.

b) Bug fixed!

rdstern (approved):
@shadrackkibet great - it all seems to be working.  Thanks for all that.
@lloyddewit to check now

instat/sdgClimaticDataEntryOptions.vb
```diff
@@ -85,14 +86,17 @@ Public Class sdgClimaticDataEntryOptions
         End Get
     End Property
 
-    Public ReadOnly Property GetDefaultValue As Double
+    Public Property GetSetDefaultValue As Double
         Get
             Dim dTemp As Double
```
lloyddewit, line 91: This get potentially returns `Nothing`
```suggestion
            Dim dTemp As Double = 0
```

instat/sdgClimaticDataEntryOptions.vb
```diff
@@ -107,14 +111,17 @@ Public Class sdgClimaticDataEntryOptions
         End Get
     End Property
 
-    Public ReadOnly Property GetTransformValue As Double
+    Public Property GetSetTransformValue As Double
         Get
             Dim dTemp As Double
```
lloyddewit, line 116: This get potentially returns `Nothing`
```suggestion
            Dim dTemp As Double = 0
```

instat/sdgCommentForDataEntry.vb
```diff
@@ -108,16 +91,24 @@ Public Class sdgCommentForDataEntry
...
+            ucrReceiverColumn.Add(strColumn, strSelectedDataFrame)
+            rdoCell.Checked = True
+        ElseIf strColumn <> "" AndAlso strRow = "" Then
+            ucrReceiverColumn.Add(strColumn, strSelectedDataFrame)
+            rdoColumn.Checked = True
+        ElseIf strColumn = "" AndAlso strRow <> "" Then
+            rdoRow.Checked = True
+        Else
+            rdoDataFrame.Checked = True
+        End If
```
lloyddewit, line 110: Should the checkboxes be set to `False` if the conditions are not met?
Can this be made more readable?
```suggestion
        Dim bStrColumnEmpty = String.IsNullOrEmpty(strColumn)
        Dim bStrRowEmpty = String.IsNullOrEmpty(strRow)
        If Not bStrColumnEmpty Then
            ucrReceiverColumn.Add(strColumn, strSelectedDataFrame)
        End If
        rdoCell.Checked = Not bStrColumnEmpty And Not bStrRowEmpty
        rdoColumn.Checked = Not bStrColumnEmpty And bStrRowEmpty
        rdoRow.Checked = bStrColumnEmpty And Not bStrRowEmpty
        rdoDataFrame.Checked = bStrColumnEmpty And bStrRowEmpty
```

instat/sdgCommentForDataEntry.vb
```diff
@@ -128,7 +119,11 @@ Public Class sdgCommentForDataEntry
         Dim expTemp As SymbolicExpression
         If clsGetKeyFunction IsNot Nothing Then
             expTemp = frmMain.clsRLink.RunInternalScriptGetValue(clsGetKeyFunction.ToScript(), bSilent:=True)
-            cmdAddComment.Enabled = expTemp IsNot Nothing AndAlso Not ucrInputComment.IsEmpty
+            If expTemp IsNot Nothing AndAlso expTemp.AsCharacter(0) = "TRUE" AndAlso Not ucrInputComment.IsEmpty Then
+                cmdAddComment.Enabled = True
+            Else
+                cmdAddComment.Enabled = False
+            End If
```
lloyddewit, line 126:
```suggestion
            cmdAddComment.Enabled = expTemp IsNot Nothing AndAlso expTemp.AsCharacter(0) = "TRUE" AndAlso Not ucrInputComment.IsEmpty
```

lloyddewit (changes requested):

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -3862,18 +3862,35 @@ DataSheet$set("public", "get_data_entry_data", function(station, date, elements,
...
   curr_data <- self$get_data_frame(use_current_filter = FALSE)
+  changed_data <- curr_data
   for (i in seq_along(rows_changed)) {
     for (k in seq_along(names(new_data))) {
-      curr_data[rows_changed[i], names(new_data)[k]] <- new_data[i, names(new_data)[k]]
+      changed_data[rows_changed[i], names(new_data)[k]] <- new_data[i, names(new_data)[k]]
+    }
+  }
+  if (add_flags) {
+    for (i in names(new_data)[-c(1:2)]) {
```
dannyparsons, line 3875: Why always not the first two columns? I don't think we should rely on this being correct in all cases.

dannyparsons:
Thanks for the comments Stephen. I've also added some comments/questions. I don't think any of these are significant functional issues hence I suggest merging as the additions are substantial and needed.

@shadrackkibet please still address the comments

instat/static/InstatObject/R/data_object_R6.R
shadrackkibet: Because we don't want to create flags for station and date columns which are always the first and the second columns in the constructed data frame. A better way will be to get elements separately.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -3862,18 +3862,35 @@ DataSheet$set("public", "get_data_entry_data", function(station, date, elements,
   curr_data
 })
 
-DataSheet$set("public", "save_data_entry_data", function(new_data, rows_changed, ...) {
+DataSheet$set("public", "save_data_entry_data", function(new_data, rows_changed, add_flags = FALSE, ...) {
   if (nrow(new_data) != length(rows_changed)) stop("new_data must have the same number of rows as length of rows_changed.")
   curr_data <- self$get_data_frame(use_current_filter = FALSE)
+  changed_data <- curr_data
```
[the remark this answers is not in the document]
shadrackkibet: The data frame is changed in line 3871. In line 3876 and 3877 we need to extract a column from the original(curr_data) data frame and the changed data frame(changed_data) to compare and produce the flag field.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -2527,7 +2527,7 @@ DataBook$set("public", "get_data_entry_data", function(data_name, station, date,
   self$get_data_objects(data_name)$get_data_entry_data(station = station, date = date, elements = elements, view_variables = view_variables, station_name = station_name, type = type, start_date = start_date, end_date = end_date)
 })
 
-DataBook$set("public", "save_data_entry_data", function(data_name, new_data, rows_changed, comments_list = list()) {
+DataBook$set("public", "save_data_entry_data", function(data_name, new_data, rows_changed, comments_list = list(), add_flags = FALSE, ...) {
```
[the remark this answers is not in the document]
shadrackkibet: It is not used anywhere. I will remove it. Thanks

Issue #6404, opened 2021-04-10
Serious bugs in Version 0.7.0 still arise


rdstern:
Serious bugs in Version 0.7.0 continue to arise.

a) I reported an odd one in #6383 
b) The video team had problems reported in #6398 

@Patowhiz should be able to find that one, because it is consistent and means the Instat side of the library is unusable.

It would be great if that is related to the intermittent problems I had - and Rachel now has.
She had the same problem I reported, but in the opposite order.  I tried to delete a data frame, and got a problem "stack overflow".  Then I tried a graph and got a different problem.

Rachel had this "the other way round".  She had a problem with the graph.  I then got her to produce and then delete a new data frame - and this produced the same error message I had started with and reported in #6383.

She temporarily "solved" her problem as I did.  This was to save the current data file.  Then use File > Close Data File and then re-open that file again.  Then all works ok at least for a time, i.e. you can make and delete data frames and draw graphs.  Is there a clue in this being a partial solution?  Are the problems anything to do with the new way we deal with graphs?

Or it is also related to the fact we started by copying a station as a subset.  Then it couldn't recognise that data frame.    Aha, if the data frame isn't recognised to be able to draw a graph, then perhaps that's why it is difficult to delete of course!  I will investigate further.  Is it to do with writing out subsets!

Here are her screen captures:

![image](https://user-images.githubusercontent.com/11226469/114259561-b38f0100-99c6-11eb-8114-ea8bdb6b58ae.png)

and then:

![image](https://user-images.githubusercontent.com/11226469/114258904-eda9d400-99c1-11eb-95d1-9f163ff3d44d.png)

rdstern:
Relief, I can make it go wrong consistently  - I think!
[germany_griefswaldtest.zip](https://github.com/africanmathsinitiative/R-Instat/files/6289317/germany_griefswaldtest.zip)

This is the data file:
1. Open it
2. Do a graph - I don't know if any graph, but I did Climatic > Compare > Correlations then sunh and sunh_sat, and options and scatterplot with month_abbr as the factor - 12 levels.
3. Now right-click and filter.  Use the filter called winter.  Write it to a subset - keep the default name.
4. Try to delete the subset - because you meant to just do a filter, not a subset.

Note if you do the subset without the graph it works fine.  What's happening now when you do a graph and then do a subset?  I suggest that's the problem!

lilyclements:
Fixed by #6443

PR #6443, opened 2021-05-03
Fixed bug deleting data frames with graphs included

### LINKED ISSUE #6383: Bug when deleting a data frame and then drawing a graph
**Describe the bug**
When deleting a data frame I got the following message:
image

**To Reproduce**
a) I used the germany merged data.  This is for 4 stations.
b) I did quite a lot of analyses before - and had R-Instat open a long time.
c) I then Filtered by factor with Station as the factor
d) I selected one station and then did a plot for this station.
e) Then I returned to the filter, and made it a subset instead.
f) I then chose the 4 months - December to March for further study.  This was intended to be a filter.  But the seting was still subset, so I got another data frame.
g) So I returned and did the filter properly.
h) Then deleted the data frame with the subset - and that gave the error. 

I have not had that error before.

I was using the new R-Instat version 0.7.0

This is the command:

```
# Code generated by the dialog, Delete Data Frames
data_book$delete_dataframes(data_names="griefswald_subset")
```

I now see that it did delete the data frame!
So I tried to continue - with a graph and got:
image

I found that I could save the rds file.  So I then closed the data file and opened it again.  (I did not close R-Instat.)  I have worked for another hour on various tasks without a problem.  I then did produce a yearly summary dataset.  When I dewleted that, I got the same error as started this issue.  Again, it did delete the data frame so I am continuing.

I take the opportunity to report a second aspect.  It may be related, because I have continued with the same data.  I have been using the correlations dialogue, with the gGally graph for a scatterplot.  It has been working fin, but I have now changed data frames and it gives an error.  Here is the code generated:

```
# Code generated by the dialog, Correlation

sunh <- data_book$get_columns_from_data(data_name="arkona", col_names="sunh")
sunh_sat <- data_book$get_columns_from_data(data_name="arkona", col_names="sunh_sat")
last_model <- cor.test(x=sunh, y=sunh_sat, alternative="two.sided", exact=NULL, method="pearson")
data_book$add_model(model_name="last_model", model=last_model, data_name="arkona")
data_book$get_models(data_name="arkona", model_name="last_model")
arkona <- data_book$get_data_frame(data_name="arkona", remove_attr=TRUE)
last_graph <- GGally::ggscatmat(data=arkona, columns=c("sunh", "sunh_sat"), color="month_abbr")
 
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="arkona")
data_book$get_graphs(data_name="arkona", graph_name="last_graph")
rm(list=c("last_model", "sunh", "sunh_sat", "last_graph", "arkona"))

```
The error is as follows:
image

It generates the graph fine.  The problem is at the line with the first data book.  Is this anything to do with the changes in the way we now save the graphs?

I again saved the rds file and opened it again.  Now the graph runs fine for arkona.  I went back to griefswald, and still ok.  And back to arkona.  Now it all works ok.  Puzzling?

I saved the file and generated another data frame.  Then deleted it.  At least that error is repeatable.  Here is the data file:

[germany_griefswald_plus.zip](https://github.com/africanmathsinitiative/R-Instat/files/6259735/germany_griefswald_plus.zip)

dannyparsons:
fixes #6383 

@rdstern Can you test this and also confirm if it solves #6404?

This definitely resolves one bug in deleting a data frame that has graphs in it but I'm not sure if there are other related bugs or if this was the cause of them. It will probably be easier for you to test to confirm.


rdstern (approved):
@dannyparsons This is better and the main problem was solved, at least in the example I tried.
So I approve this change.
However the second problem is still there.  It is also graph related, and I fear it may be due to your same changes in the system.  But it is not so severe.
I can't capture the error here.
The problem is shown after I load the germany data and do a graph.  For example a boxplot.
Then I load Dodoma and do a plot again.  It seems to give sensible code, but says it can't find Dodoma.  I will check further.

dannyparsons:
@rdstern that was quite a serious bug so good to find. I think it's now fixed

rdstern (approved):
@dannyparsons great - works fine now

PR #6459, opened 2021-05-06
Pasting into Existing columns latest

### LINKED ISSUE #5619: Adding a simple paste into a data frame

Patowhiz:
Fixes #5619

Please not this PR replaces PR #5982 and #6334 but does all that is explained in those PRs.

Pasting into existing columns is only supported if the data pasted is of the same length(which is dependent on the pasting position) as the data frame, and the pasting the column header is not supported.

A user can paste into the entire column(in this case the data pasted should be of equal length to the data frame selected) or paste from a particular cell(in this case the data pasted should be equal to the remainder of the number of rows from the selected cell row).


Patowhiz:
@dannyparsons in this PR, I have factored all the items we discussed on skype. After your review I can go ahead and now do pasting new columns and new data frame.

rdstern (approved):
@Patowhiz this seems to work well from my point of view.
@lloyddewit over to you.

Patowhiz:
@lloyddewit @dannyparsons I'll be glad if this could be reviewed. It's blocking me from making 2 new PRs.

maxwellfundi:
@lloyddewit Could you please review this?

dannyparsons (approved):

instat/ucrDataView.vb
```diff
@@ -984,4 +998,79 @@ Public Class ucrDataView
...
+    ''' <param name="startRowPos">starting row position. This starts at postion 1</param>
+    ''' <param name="firstClipRowHeader">flag indicating whether first row of clipboard data is a header</param>
+    Private Sub PasteValuesToDataFrame(lstColumnNames As IEnumerable(Of String), startRowPos As String, firstClipRowHeader As Boolean)
+        Dim clsPasteValues As New RFunction
+        Dim strColNamesVec As String = ""
+        Dim strClipBoardText As String = My.Computer.Clipboard.GetText
+
+        If String.IsNullOrEmpty(strClipBoardText) Then
+            MsgBox("No data available for pasting.", MsgBoxStyle.Information, "No Data")
+            Exit Sub
```
shadrackkibet, line 1040: If the clipboard is empty I think it is better to disable paste and therefore we don't need this message? That is what is done in other software such as excel and word? In Rstudio paste is enabled but it does nothing!

instat/ucrDataView.vb
```diff
@@ -984,4 +998,79 @@ Public Class ucrDataView
...
+        Next
+        strColNamesVec = "c(" & strColNamesVec & ")"
+
+        'add R parameters and run the command
+        clsPasteValues.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$paste_from_clipboard")
+        clsPasteValues.AddParameter("data_name", Chr(34) & grdCurrSheet.Name & Chr(34))
+        clsPasteValues.AddParameter("col_names", strColNamesVec)
+        clsPasteValues.AddParameter("start_row_pos", startRowPos)
+        clsPasteValues.AddParameter("first_clip_row_is_header", If(firstClipRowHeader, "TRUE", "FALSE"))
+        clsPasteValues.AddParameter("clip_board_text", Chr(34) & strClipBoardText & Chr(34))
```
shadrackkibet, line 1064: parameter postions?

shadrackkibet (commented):
- This should be added into column metadata right-click. The right clicks are not consistent anymore.
- We have paste under the Edit menu and the icon on the toolbar. I think it should be enabled now.
- Since we now have paste on right-click, we need copy as well?

I think we can explore ways to avoid right-click being long in some instances. We can decide to hide these features in some instances.

PR #6461, opened 2021-05-07
Ensure that the previously selected station is displayed when reopening the Daily Data Entry dialog

### LINKED ISSUE #6450: Proper setting of the station pull down to remember the station selected when you return to the dialog in Climatic>Tidy and Examine>Daily Data Entry dialog
**Describe the bug**
I have been using the Daily Data Entry dialog to enter data for Maseno and Kibos stations but I keep on selecting the station every time I want to enter a new record.
**To Reproduce**
1. Go to Climatic>Tidy and Examine>Daily Data Entry.
2. Select the the station to be entered from the station pull down (Kibos) and add the the element to the receiver.
3. Select the dates for the entry
4. Click on the Data Entry button and Click return. The station pull down returns to the default station (Achego).
**Expected behaviour**
I think when I click return the pull down should be showing the station selected before.

Patowhiz:
Fixes #6450.


rdstern (approved):
@Patowhiz this seems to be fixed

lloyddewit (approved):

lloyddewit:
@shadrackkibet This looks good to me. If you'd also like to review, then let me know. Otherwise I'm happy to merge

shadrackkibet:
Thanks, @lloyddewit. I have a question.

What happens when the control is reset? The first item in the control should be selected which I don't think it does at the moment.

lloyddewit:
> 
> 
> Thanks, @lloyddewit. I have a question.
> 
> What happens when the control is reset? The first item in the control should be selected which I don't think it does at the moment.

@shadrackkibet Interesting question. I assume that Combo Box will do the reset and will trigger the ucrInputFactorlevels event handled on line 86. Does that make sense? 
@Patowhiz please can you clarify? thanks

Patowhiz:
@shadrackkibet @lloyddewit  thanks for pointing the reset functionality. I have updated the implementation to cater for this scenario.
@lloyddewit I couldn't find a way to get the "reset" status of the linked receiver to the `ucrInputFactorLevels`. From what I gathered `ControlValueChanged` doesn't have "event properties" that could be used in situations like these. So in this case, the end developer will also always have to call the `ucrInputFactorLevels` `Reset()` method in the `SetDefaults()` of of the dialog.

rdstern (approved):
@lloyddewit this seems to be ok now from my end.

lloyddewit (approved):

lloyddewit:
@Patowhiz The PR title will be automatically added to the 0.7.2 release notes. These release notes are for both users and developers. I rephrased this PR title so that users would understand it. I hope that's OK

Patowhiz:
It's okay.

Issue #6473, opened 2021-05-17
Improving the consistency of our dialogues?


rdstern:
This suggestion relates to 2 existing issues.  Namely:
1) The comments below in issue #6469  and 2) a comment #6472 on implementing a dialogue in the standard way.

@lloyddewit I suggest after this being merged, to work on the Climatic menu in one PR.

@N-thony Thanks, good idea
Please could you try and resolve the remaining parts of issue #6454, issue #6464 and issue #6465; and then start checking/updating the dialogs in the Climatic menu?

@rdstern Is that OK from your side or did you have a different task you'd prefer @N-thony to work on?

I like to propose a slightly different system that examines dialogues for a number of aspects at the same time. 

My suggestion is to work dialogue-by-dialogue, rather than menu by menu.  In each "offending" dialogue, whenever there is anyone of the following points, then the list below is checked.  An issue is made for that dialogue, with the new Consistency label.  And a variety of aspects are checked - and if necessary - corrected at the same time.  In future then we will steadily go through quite a lot of the dialogues and those that are checked will all have the consistency label attached.

So here is my initial list of checks:
a)  Implementation is in the "standard" way.
b) Labels are appropriate length for the French.
c) The new Save control - when relevant - is completed and saves the new variable(s) in a consistent way.
d) Spacing of controls on the dialogue is consistent and well aligned.

Are there other topics to add?  What do you think?
If so, then this will be quite a bit of work.  It could wait (perhaps, or just be tested) until the next level of translation is available?  Then a dialogue for checking is "triggered" by any of the above aspects.  But once triggered each of the above is considered.  The pull request when triggered by labels can just report "Implementation is already standard".

lloyddewit:
@rdstern small comment: I assume you mean **PR** #6469 above, not issue

rdstern:
@lloyddewit yes I do.  

I also realise I was not sure what @N-thony was planning in regard to the climatic menu.  If it is the crowdin stuff, then that still stands as before.  My suggestions only affact the code/programming

lloyddewit:
@rdstern Thank you for the suggestions. I liked your checklist. I discussed with @N-thony 

> a) Implementation is in the "standard" way

There are many aspects to this, checking them all would be huge effort. I suggest to limit the scope just to ensure that the translation is done at the end of the load function. I will take this task. I hope that I can write a short script to identify which dialogs have this problem. I can then correct these dialogs. I will do this in a dedicated PR.

[...]

PR #6478, opened 2021-05-20
Ensure that dialog boxes are translated (e.g. into French) only after all text has been set

lloyddewit:
Partially fixes issue #6473.

When a dialog opens, then it's `_Load` sub is called (e.g. `Private Sub dlgAppend_Load(...`). This sub should contain the line `autoTranslate(Me)`. This line translates the dialog text into the required language (e.g. French). It should normally be the last line in the `_Load` sub. If the line appears too early in the sub, and dialog text is changed in a subsequent line, then the changed text is not translated.

Many of the dialogs already (correctly) have the `autoTranslate(Me)` line at the end of the sub. I wrote a script to identify all the dialogs that did not. For these dialogs, this PR moves the `autoTranslate(Me)` line to the end of the `_Load` function. 

There were a large number of dialogs that required this. Therefore to save time and reduce the risk of manual error, I wrote a script to update the code automatically. I also manualy updated subdialogs and other classes as required.

Please note that it's possible that some of the corrected dialogs did not have any user-detectable translation errors. However, best practice is to translate at the end of the `_Load` sub. I decided it was safer to update all dialogs that contained this potential bug, and have the `_Load` sub implemented in the standard way.

@rdstern There are a large number of changed files. I believe the change is low risk and it is sufficient just to test a sample of the updated dialogs. Please could you look at the list of changed files in this PR, test all the changed subdialogs (`sdg...`), plus a sample of the dialog boxes? 

@shadrackkibet Please could you review? Can you think of any side-effects from moving the `autoTranslate(Me)` to the end of the `_Load` sub?
Unfortunately GitHub diff flagged some lines as changed when they are not. The only real changes are relocations of `autoTranslate(Me)`.

Many thanks

### DIFF
--- a/instat/dlgAddComment.vb
+++ b/instat/dlgAddComment.vb
@@ -33,12 +33,12 @@ Public Class dlgAddComment
             SetDefaults()
         End If
         bReset = False
-        autoTranslate(Me)
         SetRCodeForControls(bReset)
         If bUseSelectedPosition Then
             SetDefaultPosition()
         End If
         TestOKEnabled()
+        autoTranslate(Me)
     End Sub
 
     Private Sub InitialiseDialog()
--- a/instat/dlgAddKey.vb
+++ b/instat/dlgAddKey.vb
@@ -25,7 +25,6 @@ Public Class dlgAddKey
     Private bUniqueChecked As Boolean = False
 
     Private Sub dlgAddKey_Load(sender As Object, e As EventArgs) Handles MyBase.Load
-        autoTranslate(Me)
         If bFirstLoad Then
             InitialiseDialog()
             bFirstLoad = False
@@ -38,6 +37,7 @@ Public Class dlgAddKey
         bReset = False
         bUniqueChecked = False
         TestOKEnabled()
+        autoTranslate(Me)
     End Sub
 
     Private Sub InitialiseDialog()
--- a/instat/dlgAddLink.vb
+++ b/instat/dlgAddLink.vb
@@ -23,7 +23,6 @@ Public Class dlgAddLink
     Private clsAddLink As RFunction
 
     Private Sub dlgAddLink_Load(sender As Object, e As EventArgs) Handles MyBase.Load
-        autoTranslate(Me)
         If bFirstLoad Then
             InitialiseDialog()
             bFirstLoad = False
@@ -36,6 +35,7 @@ Public Class dlgAddLink
         SetRCodeForControls(bReset)
         bReset = False
         TestOKEnabled()
+        autoTranslate(Me)
     End Sub
 
     Private Sub InitialiseDialog()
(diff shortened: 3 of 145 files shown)


shadrackkibet (approved):
@lloyddewit I checked the code and all looks good. I can't think of any side effects of moving the position of `autoTranslate(Me)`.  Just my observations below.

I only found out that some of the forms changed are not in use anymore. For example `dlgTwoSampleNonparametricTest`. I understand you did this automatically through code and perhaps it is not easy to avoid such dialogs. I thought of adding constraint somewhere in your code to restrict these changes to only dialogs that are added into the frmmain and active?

I think in future we should delete these forms from the project or re-use them instead of creating new forms.

I note that we should enforce this when peer reviewing and make sure that for any new dialogs `autoTranslate(Me)` comes last on the loading sub. Therefore all developers should be aware of this.

rdstern (approved):
@lloyddewit looks fine to me.

lloyddewit:
> 
> 
> @lloyddewit I checked the code and all looks good. I can't think of any side effects of moving the position of `autoTranslate(Me)`. Just my observations below.
> 
> I only found out that some of the forms changed are not in use anymore. For example `dlgTwoSampleNonparametricTest`. I understand you did this automatically through code and perhaps it is not easy to avoid such dialogs. I thought of adding constraint somewhere in your code to restrict these changes to only dialogs that are added into the frmmain and active?
> 
> I think in future we should delete these forms from the project or re-use them instead of creating new forms.
> 

I agree. I raised issue #6481 for this.

> I note that we should enforce this when peer reviewing and make sure that for any new dialogs `autoTranslate(Me)` comes last on the loading sub. Therefore all developers should be aware of this.

I agree. I thinkit would be helpful to have a checklist for peer reviewers. We could add your suggestion above  to the checklist. Do you think we should have a checklist on the R-Instat GitHub Wiki? Any other ideas?

shadrackkibet:
I agree on creating a checklist for peer reviewers. I think adding it on the R-Instat Github wiki is fine. In addition, to that, I would be happy if it will be available somewhere as suggestions so that when the reviewers click on the "Add your review" button on a PR the checklist is there and the reviewer can read it before starting a review. I think it will be helpful especially to the new programmers joining the project. The checklist gives a starting point to the reviewer.

There are some other resources I would recommend. 
- The first one is the [google engineering practices](https://google.github.io/eng-practices/) I think it is a good resource that is worth reading as a programmer. I found it useful and I am keen to see how we can adopt some of these practices in the R-Instat project.
- Second. There are very interesting and fun [videos](https://www.youtube.com/results?search_query=uncle+bob+clean+code) on clean code by Uncle Bob. There are about 6 videos (Lesson 1 to Lesson 6).

Issue #6196, opened 2021-02-01
Importing a file from a web site


rdstern:
I am wanting to improve logistic regression in R-Instat.  A useful resource is [here](https://stats.idre.ucla.edu/r/dae/logit-regression/).

It uses data from their web site and the commands there are as follows:

```
For our data analysis below, we are going to expand on Example 2 about getting into graduate school. 
We have generated hypothetical data, which can be obtained from our website from within R. 
Note that R requires forward slashes (/) not back slashes () when specifying a file location even if the file is on 
your hard drive.

mydata <- read.csv("https://stats.idre.ucla.edu/stat/data/binary.csv")
```
I would like to access this data file using our current Fil > Open from File dialogue.  I suppose in the current dialogue I could use the Browse and search for the location.  But it would be much simpler to copy the location from the chapter and then just paste it into the dialogue?

Is that true?  If so, then could we allow pasting into that field?  Would anything else need to change?  Otherwise could another solution be suggested?  

Aha - @dannyparsons doesn't want pasting, but thinks it should be pretty easy.  He will add to this message.  Contact him if you need to.

I checked this works by getting the script and pasting there.  I also then changed the names of the resulting data.

```
# Code generated by the dialog, Import Dataset

class <- rio::import(file="https://stats.idre.ucla.edu/stat/data/binary.csv")
data_book$import_data(data_tables=list(class=class))

rm(class)
```

For reference I attach a copy of the data.
[class.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/5903437/class.xlsx)

dannyparsons:
Actually, I wouldn't mind if the file location textbox was editable and you could paste a web link in there instead.

However, it's not a trivial change. 
- The dialog options need to adapt based on the file type of the selected file, this need to also work for a web link. This also has implications for the file control that Patrick has created.
- The preview currently displays automatically, and refreshes with every option change. I would suggest this shouldn't be done automatically for a weblink, only if you request a preview, to avoid continuously downloading files. So this needs a change in the dialog functionality.

This change probably shouldn't happen until after @Patowhiz has finished working on this dialog. Once that's done, this could be the next improvement to the dialog.

N-thony:
> Actually, I wouldn't mind if the file location textbox was editable and you could paste a web link in there instead.
> 
> However, it's not a trivial change.
> 
[...]

PR #6501, opened 2021-05-27
Added import of multiple text files; also ensured that Excel files are consistently imported correctly

### LINKED ISSUE #6476: Reading of multiple txt files
I am trying to read (climatic) data of course.  There are multiple files per element, with one file per year.  In a file that is shared the tmax has missing as -100 and reading an individual file it works fine reading 31 rows.  There are 13 columns with the day of the month plus one for each month:
image

These are txt files and in the import dialogue I use the fixed width option as shown below:
image

When I switch to read the multiple files is changes to csv as shown below and doesn't read the 31 rows.  It reads 28 usually and 29 in a leap year.

image

### LINKED ISSUE #6482: Small bug in importing Excel file
This is an awkward file - sent separately to @Patowhiz .  The bug may be more general and is (I think) easy to fix.
I am importing a selection from the file as shown:
image

(So no column names, skipping 281 rows and importing 75 rows.)

The commands generated were as follows:
```
# Code generated by the dialog, Import Dataset
vtary_muchinga <- rio::import(file="C:/Users/RogerStern/Dropbox (SSD)/Roger/Climatic/Zambia/Voluntary/Voluntary stations.xlsx", guess_max=Inf, which=1, skip=281, col_names=FALSE, n_max=74)
data_book$import_data(data_tables=list(vtary_muchinga=vtary_muchinga))

rm(vtary_muchinga) 
```
But I got it wrong first time and only imported 74 rows.  So I deleted the data file and tried again.  Then it went wrong and imported each variablew into a different data frame:

image

The commands were now:

```
# Code generated by the dialog, Import Dataset
vtary_muchinga <- rio::import(file="C:/Users/RogerStern/Dropbox (SSD)/Roger/Climatic/Zambia/Voluntary/Voluntary stations.xlsx", guess_max=Inf, which=1, skip=281, col_names=FALSE, n_max=75)
data_book$import_data(data_tables=vtary_muchinga)
```
You can see that the main rio:import command is the same, but the `data_book$import` line has changed.  I assume that is the problem?

Patowhiz:
Fixes #6482 and Fixes #6476


lloyddewit (dismissed):

lloyddewit:
@rdstern please can you test? thanks

Patowhiz:
@rdstern you can now also test if this PR fixes issue #6476. I had to push changes here because they affect the same dialog. Thanks.

Patowhiz:
@lloyddewit I think this should now also have a "feature" label.

rdstern:
@Patowhiz great.  This now works fine on the import, with just the extension given as the sheet name.
But I wonder what should happen next?  When I open again it seems ok as well.

However, I tried closing the Data Book to go back to your introductory screen.  I think then perhaps the Open dialogue should reset?    (I suggest this is a reasonable thing to do, if the import of multiple files didn't work as well as would be hoped - that's the situation with the Pakistan files, where we find they are inconsistent.)

Currently, when I then try to open an rds file I get:

image

If I try to open a similar single file to before I get:

image

In each case I have to press Reset.  Then I get the default directory:
image

Aha.  I need to press REset on the dialogue.  Then I close it and the try again from the start-up screen.  

I wonder what should happen?

lloyddewit:
> 
> 
> @lloyddewit I think this should now also have a "feature" label.

agreed

Patowhiz:
@rdstern thanks for the multiple testing. That's definitely a bug probably cause by my new additions. Let me fix that.

Patowhiz:
@rdstern I'm unable to reproduce the bug you described above. Closed the databook as explained above and reopened the dialog and it still retained its previous state. Perhaps a skype meeting to demonstrate how you got this bug or steps that I can take to reproduce it

Patowhiz:
@rdstern I have fixed the bug you described above. You can now review.

rdstern (approved):
From my point of view this seems ok

Patowhiz:
This PR is blocking issue #6196 and refactoring of the import dialog. It may potentially also block issue #6556.

instat/dlgImportDataset.vb
```diff
@@ -959,9 +1001,9 @@ Public Class dlgImportDataset
     End Sub
 
     Private Sub ucrBase_ClickOk(sender As Object, e As EventArgs) Handles ucrBase.ClickOk
-        ' add the item to the MRU (Most Recently Used) list...
-        'only add if its a file that was selected
-        If Not ucrChkMultipleFiles.Checked Then
+        'add the item to the MRU (Most Recently Used) list...
+        'only add if its a file that was selected, don't folder paths
```
lloyddewit, line 1005:
```suggestion
        'only add if its a file that was selected, don't add if folder path was selected
```

lloyddewit (changes requested):

lloyddewit (approved):

lloyddewit:
@rdstern I made a small change to one of the comments. If you can approve again, then I'll merge, thanks

rdstern (approved):

PR #6503, opened 2021-05-29
Added Paste New Data Frame dialog

Patowhiz:
Fixes #6061

@africanmathsinitiative/developers this is ready for review once PR #6505 is merged

### DIFF
--- a/instat/dlgPasteNewDataFrame.vb
+++ b/instat/dlgPasteNewDataFrame.vb
@@ -20,12 +20,9 @@ Public Class dlgPasteNewDataFrame
 
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
-    Private clsPasteFunction As RFunction
-    'used to prevent TestOkEnabled from being called multiple times when loading the dialog. 
-    Private bIsOnDialogLoad As Boolean
+    Private clsAddData, clsDataList, clsPasteFunction As New RFunction
 
     Private Sub dlgPasteNewDataFrame_Load(sender As Object, e As EventArgs) Handles MyBase.Load
-        bIsOnDialogLoad = True
         autoTranslate(Me)
         If bFirstLoad Then
             InitialiseDialog()
@@ -38,7 +35,6 @@ Public Class dlgPasteNewDataFrame
         SetClipBoardDataParameter()
         SetRCodeForControls(bReset)
         bReset = False
-        bIsOnDialogLoad = False
         TestOkEnabled()
     End Sub
 
@@ -51,7 +47,6 @@ Public Class dlgPasteNewDataFrame
         ucrSaveNewDFName.SetLabelText("New Data Frame Name:")
         ucrSaveNewDFName.SetPrefix("data")
 
-        'ucrChkRowHeader.Text = "First row is header"
         ucrChkRowHeader.SetText("First row is header")
         ucrChkRowHeader.SetParameter(New RParameter("header", 1))
 
@@ -60,7 +55,17 @@ Public Class dlgPasteNewDataFrame
     End Sub
 
     Private Sub SetDefaults()
+        clsAddData = New RFunction
+        clsDataList = New RFunction
         clsPasteFunction = New RFunction
+
+        'todo 29/05/2021. this is temporarily done this way because of how
+        'function ConstructAssignTo in clsRCodeStructure is currently implemented.
+        'It doesn't construct assignTo statements correctly
+        clsAddData.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$import_data")
+        clsAddData.AddParameter("data_tables", clsRFunctionParameter:=clsDataList, iPosition:=0)
+        clsDataList.SetRCommand("list")
+        clsDataList.AddParameter(strParameterName:=ucrSaveNewDFName.GetText, clsRFunctionParameter:=clsPasteFunction)
 
         clsPasteFunction.SetPackageName("clipr")
         clsPasteFunction.SetRCommand("read_clip_tbl")
@@ -70,7 +75,7 @@ Public Class dlgPasteNewDataFrame
         'please note. for some reason,this parameter is not used by clipr if x(data) parameter is specified.
         clsPasteFunction.AddParameter("nrows", 1000, iPosition:=2)
 
-        ucrBase.clsRsyntax.SetBaseRFunction(clsPasteFunction)
+        ucrBase.clsRsyntax.SetBaseRFunction(clsAddData)
 
         ucrNudPreviewLines.Value = 10
         ucrSaveNewDFName.Reset()
@@ -78,7 +83,8 @@ Public Class dlgPasteNewDataFrame
 
     Private Sub SetRCodeForControls(bReset As Boolean)
         ucrChkRowHeader.SetRCode(clsPasteFunction, bReset)
-        ucrSaveNewDFName.SetRCode(clsPasteFunction, bReset)
+        'todo 29/05/2021. this is temporarily commented out
+        'ucrSaveNewDFName.SetRCode(clsPasteFunction, bReset)
     End Sub
 
     Private Sub TestOkEnabled()
@@ -91,10 +97,14 @@ Public Class dlgPasteNewDataFrame
         TestOkEnabled()
     End Sub
 
-    Private Sub ucrControls_ControlContentsChanged(ucrchangedControl As ucrCore) Handles ucrSaveNewDFName.ControlContentsChanged, ucrChkRowHeader.ControlContentsChanged, ucrNudPreviewLines.ControlContentsChanged
-        If Not bIsOnDialogLoad Then
-            TestOkEnabled()
-        End If
+    Private Sub ucrControls_ControlContentsChanged(ucrchangedControl As ucrCore) Handles ucrChkRowHeader.ControlContentsChanged, ucrNudPreviewLines.ControlContentsChanged
+        TestOkEnabled()
+    End Sub
+
+    Private Sub ucrSaveNewDFName_ControlValueChangedChanged(ucrchangedControl As ucrCore) Handles ucrSaveNewDFName.ControlValueChanged
+        clsDataList.ClearParameters()
+        clsDataList.AddParameter(strParameterName:=ucrSaveNewDFName.GetText, clsRFunctionParameter:=clsPasteFunction)
+        TestOkEnabled()
     End Sub
 
     Private Sub btnRefreshPreview_Click(sender As Object, e As EventArgs) Handles btnRefreshPreview.Click
@@ -103,8 +113,7 @@ Public Class dlgPasteNewDataFrame
     End Sub
 
     Public Sub SetClipBoardDataParameter()
-        'todo 29/05/2021. this is temporarily commented because of the way function ConstructAssignTo in clsRCodeStructure is currently implemented
-        'clsPasteFunction.AddParameter("x", Chr(34) & My.Computer.Clipboard.GetText & Chr(34), iPosition:=0)
+        clsPasteFunction.AddParameter("x", Chr(34) & My.Computer.Clipboard.GetText & Chr(34), iPosition:=0)
     End Sub
 
     ''' <summary>
@@ -115,10 +124,10 @@ Public Class dlgPasteNewDataFrame
         Dim bValid As Boolean = False
         Dim dfTemp As DataFrame
         Dim expTemp As SymbolicExpression
-        Dim copiedDataRowCount As Integer
+        'Dim copiedDataRowCount As Integer
         Dim clsTempImport As New RFunction
         'split copied data to an array of indivindual lines. Used to validate length of data.
-        Dim arrStrTemp() As String = My.Computer.Clipboard.GetText().Split(New String() {Environment.NewLine}, StringSplitOptions.None)
+        'Dim arrStrTemp() As String = My.Computer.Clipboard.GetText().Split(New String() {Environment.NewLine}, StringSplitOptions.None)
 
         'set feedbcak controls default states
         panelNoDataPreview.Visible = True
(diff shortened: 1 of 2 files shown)


lloyddewit:

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,190 @@
...
+        clsTempImport.SetRCommand("read_clip_tbl")
+
+        'reconstruct the copied data from the array. No need will just be slower, 
+        'so let clipr use its clipboard functionality. commented code left here for future refernce
+        'clsTempImport.AddParameter("x", Chr(34) & Strings.Join(arrStrTemp, Environment.NewLine) & Chr(34))
+
+        clsTempImport.AddParameter("header", If(ucrChkRowHeader.Checked, "TRUE", "FALSE"))
+        'copied data could be long, so restrict to 10 rows only for preview. 
+        'please note. for some reason,this parameter is not used by clipr if x(data) parameter is specified.
+        clsTempImport.AddParameter("nrows", ucrNudPreviewLines.Value)
```
shadrackkibet, line 164: Why is this set manually? Can't this be done through `SetRcode`?

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,190 @@
...
+        clsPasteFunction.SetRCommand("read_clip_tbl")
+        SetClipBoardDataParameter()
+        clsPasteFunction.AddParameter("header", "TRUE", iPosition:=1)
+        'copied data could be long, so restrict to 1000 rows only for import. 
+        'please note. for some reason,this parameter is not used by clipr if x(data) parameter is specified.
+        clsPasteFunction.AddParameter("nrows", 1000, iPosition:=2)
+
+        ucrBase.clsRsyntax.SetBaseRFunction(clsAddData)
+
+        ucrNudPreviewLines.Value = 10
```
shadrackkibet, line 80: We shouldn't be setting this manually here. It should be added as a parameter value.

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,190 @@
...
+
+        ucrSaveNewDFName.SetIsTextBox()
+        ucrSaveNewDFName.SetSaveTypeAsDataFrame()
+        ucrSaveNewDFName.SetLabelText("New Data Frame Name:")
+        ucrSaveNewDFName.SetPrefix("data")
+
+        ucrChkRowHeader.SetText("First row is header")
+        ucrChkRowHeader.SetParameter(New RParameter("header", 1))
+
+        ucrNudPreviewLines.Maximum = 1000
```
shadrackkibet, line 53: I see the minimum is set in the control properties. I think we should set min/max here.

shadrackkibet (changes requested):

shadrackkibet:
- Sort out Tab order?
- Why the column name is SetRcode when nothing is on the clipboard? 
image
- I think it's better to have the preview controls at the bottom left. That is what we have in the import dialog.

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,190 @@
...
+        'Else
+        '    lblConfirmText.Text = "Click Ok to paste data to new data frames."
+        'End If
+
+        'use clipr package to check if structure of data can be pasted to a data frame 
+        clsTempImport.SetPackageName("clipr")
+        clsTempImport.SetRCommand("read_clip_tbl")
+
+        'reconstruct the copied data from the array. No need will just be slower, 
+        'so let clipr use its clipboard functionality. commented code left here for future refernce
```
shadrackkibet, line 158:
```suggestion
        'so let clipr use its clipboard functionality. commented code left here for future reference
```

shadrackkibet (changes requested):

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,190 @@
...
+        'clsTempImport.AddParameter("x", Chr(34) & Strings.Join(arrStrTemp, Environment.NewLine) & Chr(34))
+
+        clsTempImport.AddParameter("header", If(ucrChkRowHeader.Checked, "TRUE", "FALSE"))
```
shadrackkibet, line 161: Why is this not done through SetRcode?

instat/dlgPasteNewDataFrame.vb
Patowhiz: @shadrackkibet the R function variable used here is local to the VB.NET function.

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,190 @@
...
     End Sub
+
+    Private Sub ucrControls_ControlContentsChanged(ucrchangedControl As ucrCore) Handles ucrChkRowHeader.ControlContentsChanged, ucrNudPreviewLines.ControlContentsChanged
+        TestOkEnabled()
+    End Sub
+
+    Private Sub ucrSaveNewDFName_ControlValueChangedChanged(ucrchangedControl As ucrCore) Handles ucrSaveNewDFName.ControlValueChanged
+        clsDataList.ClearParameters()
+        clsDataList.AddParameter(strParameterName:=ucrSaveNewDFName.GetText, clsRFunctionParameter:=clsPasteFunction)
+        TestOkEnabled()
```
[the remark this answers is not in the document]
Patowhiz: Yes it can be placed there as is the case in other dialogs. But I have it here because of what happens in `TestOkEnabled`. I choose to have it here to avoid the temporary "blinks" on the dialog as the user types in the name. The validation involves reading the clipboard data and I find it optimal and still functional to have it done here when this control has received user changes.

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,190 @@
...
+    End Sub
+
+    Private Sub btnRefreshPreview_Click(sender As Object, e As EventArgs) Handles btnRefreshPreview.Click
+        SetClipBoardDataParameter()
+        TestOkEnabled()
```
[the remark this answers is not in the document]
Patowhiz: This is a click event for a VB.NET button. VB.Net click events have different signatures to the custom events in R-Instat so I had to have this event on its own, thus the repetition.

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,190 @@
...
+        'reset the clip board data parameter value
+        SetClipBoardDataParameter()
+        SetRCodeForControls(bReset)
+        bReset = False
+        TestOkEnabled()
+    End Sub
+
+    Private Sub InitialiseDialog()
+        'todo. attach the help id later
+        'ucrBase.iHelpTopicID = 332
```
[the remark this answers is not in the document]
Patowhiz: @shadrackkibet 332 is just a random commented placeholder number that will be replaced later as the todo. states. I've changed it to blank.
@rdstern what should be the help id for this dialog? I can still leave it as a todo comment if it needs time to be created.

instat/dlgPasteNewDataFrame.vb
Patowhiz: @shadrackkibet why should we be adding it as a parameter value? 

I'm doing it manually because the control never reads nor writes to the base R function or any R functional global to this dialog class. It's specifically used for preview locally to the function used to preview the clipboard data.

instat/dlgPasteNewDataFrame.vb
Patowhiz: Done.

instat/dlgPasteNewDataFrame.vb
Patowhiz: Same as above comment.

Patowhiz:
> * Sort out Tab order?
> * Why the column name is SetRcode when nothing is on the clipboard?
>   image
> * I think it's better to have the preview controls at the bottom left. That is what we have in the import dialog.

@rdstern what's you suggestion on the new design changes proposed by @shadrackkibet above. I'm happy with either. My main concern would be the UX flow but I can see the benefit of the dialog consistent with the other dialog.

rdstern:
@Patowhiz and @shadrackkibet I don't have strong views.  I like consistency as a default.  But this may be sufficiently different that the change is sensible.  Here is the import dialogue.
image

The preview isn't so wide, as here, so the Refresh preview is below the preview there.  And the important First Row is Column Names is top left, with other options.  However, it seems neat and clear where it is.  If there are other options, then I'd be inclined to put them together, and I suppose at the bottom.  If not, then I quite like it at the top, because that's where the first line - possibly the header line - will be too.

instat/dlgPasteNewDataFrame.vb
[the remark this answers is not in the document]
shadrackkibet: This means Ok will still be enabled even if the save control is empty although you can't click Ok?

instat/dlgPasteNewDataFrame.vb
shadrackkibet: Can't we make the function global then it will be possible to SetRCode?

instat/dlgPasteNewDataFrame.vb
shadrackkibet: If you agree to make `clsTempImport` function global then it will be possible to set defaults this way. 

Briefly, I understand that in future we want dialogs to open from script therefore we should try to pass parameters by setting Rcode.

instat/dlgPasteNewDataFrame.vb
```diff
@@ -50,8 +50,8 @@ Public Class dlgPasteNewDataFrame
         ucrChkRowHeader.SetText("First row is header")
         ucrChkRowHeader.SetParameter(New RParameter("header", 1))
 
+        ucrNudPreviewLines.Minimum = 10
         ucrNudPreviewLines.Maximum = 1000
```
shadrackkibet: Use `ucrNudPreviewLines.SetMinMax()` to make this one line?

Patowhiz:
@africanmathsinitiative/developers a few comments to add;
1. This dialog will load slowly if the data to be pasted is long e.g data with 10 columns and over 2000 rows.
2. Reproducibility has taken precedence over optmisation in the execution of code.
3. `clsRcodeStructure` object that constructs "assign to" statements doesn't yet correctly support script produced by this dialog, so modification have been added to make this dialog work(check inline comments).

instat/dlgPasteNewDataFrame.vb
Patowhiz: Done.

instat/dlgPasteNewDataFrame.vb
Patowhiz: > If you agree to make `clsTempImport `function global then it will be possible to set defaults this way.

@shadrackkibet yes that's one alternative. But I thought that will just complicate the dialog more and will still not achieve the required aim. The other alternative which is just to clone the R function object that is being "duplicated" by `clsTempImport`. I have chosen that alternative as it simplifies the code and could be used achieve the aim of reproducibility in future.

> Briefly, I understand that in future we want dialogs to open from script therefore we should try to pass parameters by setting Rcode.

@shadrackkibet yes. But `ucrNudPreviewLines` never reads nor writes to the dialog's base functions. Which means any changes made to it will never be logged in the R log file.

Patowhiz:
@africanmathsinitiative/developers after discussing @rdstern and @dannyparsons a few comments to add;
1. Reproducibility has taken precedence over optmisation in the execution of code (though this is only affected when the data is large).
2. This dialog may load slowly if the data to be pasted is large (e.g data with 10 columns and over 2000 rows) but this acceptable because we want to discourage users from "pasting" large data(for large data they should use import dialog).
3. `clsRcodeStructure` object that constructs "assign to" statements doesn't yet correctly support script produced by this dialog, so modification have been added to make this dialog work(check inline comments for further explanation).
4. Currently, this dialog will always show contents of the clipboard by reading from the system clipboard every time the dialog is loaded. Which means on reloading, the contents being previewed may be different. To make the dialog to open from script and preview the data from script, a design change will have to be made.

instat/dlgPasteNewDataFrame.vb
Patowhiz: This parameter is now read and written by the control. Done.

instat/dlgPasteNewDataFrame.vb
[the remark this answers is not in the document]
Patowhiz: @shadrackkibet thanks. I have modified the `TestOkEnabled` to work correctly when the save control is empty.

Patowhiz:
@rdstern based on @dannyparsons suggestion I have now restricted the allowable data paste to 1000 rows. Please retest this and check how it performs on your machine.

rdstern (changes requested):
@Patowhiz this is great.  For me it works fine.  I tried with longer than 1000 rows and it was quick in not letting me.
I imported with 999 rows and 10 variables and it was quick.  
The only item to correct is that it always says there are 10 rows to import.  That's under the preview.  It should give the number of rows in the clipboard, not in the proview.

Then I am ready to approve!

Patowhiz:
@rdstern as discussed, I have removed the display of columns and rows found. Ready for your review.

rdstern (dismissed):
@lloyddewit from my point of view this is working really well.

lloyddewit:
@shadrackkibet Thank you for your peer reviews on this PR. It looks like most of your comments are now resolved. However, there are still some open comments. Your comments about SetRCode are especially important.
Please could you check if all your comments are resolved? If so, please could you approve?
I will then review, approve and merge.
Thanks

shadrackkibet (dismissed):
Thanks @lloyddewit I don't think it is worth delaying anymore. Over to you.

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,183 @@
...
+Imports RDotNet
+Imports instat.Translations
+Public Class dlgPasteNewDataFrame
+
+    Private bFirstLoad As Boolean = True
+    Private bReset As Boolean = True
+    Private clsAddData, clsDataList, clsPasteFunction As New RFunction
+
     Private Sub dlgPasteNewDataFrame_Load(sender As Object, e As EventArgs) Handles MyBase.Load
+        autoTranslate(Me)
```
lloyddewit, line 26: This should be at the end of the sub.
Otherwise the text set in `InitialiseDialog` will not be translated.

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,183 @@
...
+' You should have received a copy of the GNU General Public License 
+' along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+Imports RDotNet
+Imports instat.Translations
+Public Class dlgPasteNewDataFrame
+
+    Private bFirstLoad As Boolean = True
+    Private bReset As Boolean = True
+    Private clsAddData, clsDataList, clsPasteFunction As New RFunction
```
lloyddewit, line 23: Could we rename these RFunctions?
```suggestion
    Private clsAddDataFunction, clsDataListFunction, clsPasteFunction As New RFunction
```

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,183 @@
...
+        'todo. this could be changed if the ucrCheck raising of ControlContentsChanged event is modified
+        TestOkEnabled()
+    End Sub
+
+    Private Sub btnRefreshPreview_Click(sender As Object, e As EventArgs) Handles btnRefreshPreview.Click
+        SetClipBoardDataParameter()
+        TestOkEnabled()
+    End Sub
+
+    Public Sub SetClipBoardDataParameter()
```
lloyddewit, line 123: Should this be private?
If it needs to be public, then we should add a header comment

instat/dlgPasteNewDataFrame.vb
```diff
@@ -1,5 +1,183 @@
...
+                'try to preview the data
+                frmMain.clsGrids.FillSheet(dfTemp, "temp", grdDataPreview, bIncludeDataTypes:=False, iColMax:=frmMain.clsGrids.iMaxCols)
+                bValid = True
+                lblConfirmText.Text = "Click Ok to paste data to new data frames."
+                lblConfirmText.ForeColor = Color.Green
+            Catch
+                lblConfirmText.Text = "Could not preview data. Cannot be pasted."
+            End Try
+        Else
+            lblConfirmText.Text = "Could not preview data. Cannot be pasted."
```
lloyddewit, line 176: Ensure text translated
```suggestion
                lblConfirmText.Text = GetTranslation("Click Ok to paste data to new data frames.")
                lblConfirmText.ForeColor = Color.Green
            Catch
                lblConfirmText.Text = GetTranslation("Could not preview data. Cannot be pasted.")
            End Try
        Else
            lblConfirmText.Text = GetTranslation("Could not preview data. Cannot be pasted.")
```

lloyddewit (changes requested):
This looks good, just a few suggestions

instat/dlgPasteNewDataFrame.vb
Patowhiz: Done.

instat/dlgPasteNewDataFrame.vb
Patowhiz: Yes, should be private. done

instat/dlgPasteNewDataFrame.vb
Patowhiz: Done.

lloyddewit (approved):

lloyddewit:
@rdstern The changes since your last approval should not have affected the functionality. Please could you retest and approve? thanks

rdstern (approved):

PR #6507, opened 2021-06-02
Added a 'More...' option to the file menu and icon toolstrip

### LINKED ISSUE #6466: Add the more facility to the File menu and the toolbar
When starting R-Instat the recent files are available, together with the paths.
And there is a More button. 

image

The toolbar also has the recent file names and so does the bottom of the File menu.  Please could the More be available here too.

Patowhiz:
Fixes #6466

### DIFF
--- a/instat/clsRecentFiles.vb
+++ b/instat/clsRecentFiles.vb
@@ -297,7 +297,8 @@ Public Class clsRecentFiles
                 End If
 
                 If Not bShowAll Then
-                    'if recent files are more than 5 then just the "more" link label and exit loop 
+                    'if recent files are more than 5 then display the "more" link label and exit loop 
+
                     If lstRecentOpenedFiles.Count - iCounter >= 5 Then
                         If mnuFile IsNot Nothing Then
                             Dim clsItem As ToolStripMenuItem = New ToolStripMenuItem(GetTranslation("More ..."))
--- a/instat/clsRecentFiles.vb
+++ b/instat/clsRecentFiles.vb
@@ -164,18 +164,20 @@ Public Class clsRecentFiles
         Dim strFilePath As String = ""
         'if tag is "more" then its more link that was clicked so exit sub
         If TypeOf sender Is ToolStripItem Then
+            Dim bIsMoreLink As Boolean
             Dim toolStripItem As ToolStripItem = DirectCast(sender, ToolStripItem)
-            strFilePath = toolStripItem.Tag.ToString().Substring(4) 'exclude MRU: 
+            strFilePath = toolStripItem.Tag.ToString()
+            bIsMoreLink = (strFilePath = "MRU:more")
             If toolStripItem.OwnerItem Is mnuFile Then
-                If strFilePath.Contains("more") Then
+                If bIsMoreLink Then
                     UpdateRecentFilesMenuItems(True, mnuFile:=mnuFile)
                     mnuFile.ShowDropDown()
                     AddHandler mnuFile.DropDownClosed, AddressOf MenuDropDownClosedAfterMoreClicked
                     Exit Sub
                 End If
             ElseIf toolStripItem.OwnerItem Is mnuFileIcon Then
-                If strFilePath.Contains("more") Then
-                    UpdateRecentFilesMenuItems(strFilePath.Contains("more"), mnuFileIcon:=mnuFileIcon)
+                If bIsMoreLink Then
+                    UpdateRecentFilesMenuItems(True, mnuFileIcon:=mnuFileIcon)
                     mnuFileIcon.ShowDropDown()
                     AddHandler mnuFileIcon.DropDownClosed, AddressOf MenuDropDownClosedAfterMoreClicked
                     Exit Sub
@@ -183,7 +185,7 @@ Public Class clsRecentFiles
             End If
         ElseIf TypeOf sender Is LinkLabel Then
             strFilePath = DirectCast(sender, LinkLabel).Tag.ToString()
-            If strFilePath.Contains("more") Then
+            If strFilePath = "more" Then
                 UpdateRecentFilesMenuItems(True, ucrDataViewWindow:=ucrDataViewWindow)
                 Exit Sub
             End If
@@ -226,14 +228,14 @@ Public Class clsRecentFiles
     ''' <param name="mnuFile"></param>
     ''' <param name="mnuFileIcon"></param>
     ''' <param name="ucrDataViewWindow"></param>
-    Public Sub UpdateRecentFilesMenuItems(bShowAll As Boolean, Optional mnuFile As ToolStripMenuItem = Nothing, Optional mnuFileIcon As ToolStripSplitButton = Nothing, Optional ucrDataViewWindow As ucrDataView = Nothing)
-        'exit sub if file toolstrip or file icon toolstrip or data view are not initialised
+    Public Sub UpdateRecentFilesMenuItems(bShowAll As Boolean, Optional mnuFile As ToolStripMenuItem = Nothing,
+                                          Optional mnuFileIcon As ToolStripSplitButton = Nothing,
+                                          Optional ucrDataViewWindow As ucrDataView = Nothing)
+        'exit sub if file toolstrip, file icon toolstrip and data view are not initialised
         If mnuFile Is Nothing AndAlso mnuFileIcon Is Nothing AndAlso ucrDataViewWindow Is Nothing Then
             Exit Sub
         End If
 
-        Dim strPath As String
-        Dim strFileName As String
 
         If mnuFile IsNot Nothing Then
             'remove all recent files from the menu items
@@ -258,17 +260,14 @@ Public Class clsRecentFiles
         End If
 
         'then add and displays items (_in reverse order) for recent files
-        Dim clsItem As ToolStripMenuItem
-        Dim clsItemIcon As ToolStripMenuItem
-        Dim linkMenuItem As LinkLabel
         For iCounter As Integer = lstRecentOpenedFiles.Count - 1 To 0 Step -1
             Try
-                strPath = lstRecentOpenedFiles(iCounter)
-                strFileName = Path.GetFileName(strPath)
+                Dim strPath As String = lstRecentOpenedFiles(iCounter)
+                Dim strFileName As String = Path.GetFileName(strPath)
 
                 'create new ToolStripItem, displaying the name of the file...
                 If mnuFile IsNot Nothing Then
-                    clsItem = New ToolStripMenuItem(strFileName)
+                    Dim clsItem As ToolStripMenuItem = New ToolStripMenuItem(strFileName)
                     clsItem.ToolTipText = strPath
                     'set the tag - will be used to identify the ToolStripItem as an most recent(MRU) item 
                     'and contains the full path so it can be opened later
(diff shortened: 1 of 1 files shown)


lloyddewit:
@Ivanluv - please could you peer review? thanks

rdstern:
It works - great.  But I currently have the maximum of 30 files and it is quite long.
a) One change I would like is to reset it once the menu is closed.  Currently, if you have ever pressed More, then it is there forever.
b) In the File menu the Exit option is the last one.  I suggest it be hidden when the More is pressed.
c) Sometimes the More is Plus?  I am happy with either, but wonder why the word changes?

Here is the file menu with the option in action.  

image

I am sort of happy with that, (as long as the Exit isn't still there at the bottom, but it is perhaps slightly odd that the menu doesn't keep some of the other options visible.

@dannyparsons was very keen that we keep the options the same for the toolbar and the file menu.  I would like it there somewhere.  So if others feel it makes the File menu look too odd, then perhaps we just add it to the toolbar?

Patowhiz:
@rdstern I have added the above requested changes. You can now review.

rdstern:
@Patowhiz you asked to see the Plus.  I am not sure why it sometimes happens and am not against it!  Here is is this time!
image

Happy with it now resetting each time.  I think that's better.

rdstern (approved):

Patowhiz:
@lloyddewit could the "Plus..." be related in anyway with the translation? In the code, I found some changes that make the "More..." text be retrieved through the translation function. That's my suspicion.

rdstern:
@Patowhi - sorry that's my mistake.  Of course - it IS the French!  So, nothing for you to do except (of course) include the time you have wasted looking for a non-existent problem that I have introduced!

Ivanluv:
@lloyddewit  the code looks okay from my end

lloyddewit (changes requested):
This looks good! just a few small suggestions

instat/clsRecentFiles.vb
```diff
@@ -286,28 +218,153 @@ Public Class clsRecentFiles
...
+        Dim clsItemIcon As ToolStripMenuItem
+        Dim linkMenuItem As LinkLabel
         For iCounter As Integer = lstRecentOpenedFiles.Count - 1 To 0 Step -1
-            strPath = lstRecentOpenedFiles(iCounter)
             Try
-                Dim linkMenuItem As New LinkLabel
-                linkMenuItem.Text = Path.GetFileName(strPath)
-                linkMenuItem.Tag = strPath 'path used when the link is clicked
+                strPath = lstRecentOpenedFiles(iCounter)
+                strFileName = Path.GetFileName(strPath)
```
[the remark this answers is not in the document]
Patowhiz: @lloyddewit I have declared the 2 variables just above the start of the for loop so that their addresses can be reused in the for loop and also be consistent with how I'm using the `ToolStripMenuItem `and `LinkLabel `objects that are also used in the for loop function.

instat/clsRecentFiles.vb
```diff
@@ -286,28 +218,153 @@ Public Class clsRecentFiles
...
+
+        If ucrDataViewWindow IsNot Nothing Then
+            'remove all the data view window recent file menu items
+            ucrDataViewWindow.ClearRecentFileMenuItems()
+        End If
+
+        'then add and displays items (_in reverse order) for recent files
+        Dim clsItem As ToolStripMenuItem
+        Dim clsItemIcon As ToolStripMenuItem
+        Dim linkMenuItem As LinkLabel
```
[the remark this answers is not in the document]
Patowhiz: Yes we can delete one of them.

instat/clsRecentFiles.vb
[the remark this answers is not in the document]
Patowhiz: @lloyddewit after relooking the code, I think any optimization in regards to this function is insignificant compared to the code clarity, the list will never be big enough to cause any noticeable slow performance. So I've moved the declarations to the individual blocks.

instat/clsRecentFiles.vb
```diff
@@ -154,105 +154,37 @@ Public Class clsRecentFiles
...
+    Private Sub OnMnuRecentOpenedFile_Click(sender As Object, e As EventArgs)
         Dim strFilePath As String = ""
-
+        'if tag is "more" then its more link that was clicked so exit sub
         If TypeOf sender Is ToolStripItem Then
-            strFilePath = DirectCast(sender, ToolStripItem).Tag.ToString().Substring(4)
+            Dim toolStripItem As ToolStripItem = DirectCast(sender, ToolStripItem)
+            strFilePath = toolStripItem.Tag.ToString().Substring(4) 'exclude MRU: 
+            If toolStripItem.OwnerItem Is mnuFile Then
+                If strFilePath.Contains("more") Then
```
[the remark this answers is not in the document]
Patowhiz: Agreed and redone.

Patowhiz:
@lloyddewit I have made the suggested code changes. You can now review.

instat/clsRecentFiles.vb
```diff
@@ -154,105 +154,39 @@ Public Class clsRecentFiles
...
     ''' raised when menu item of recent opened file is clicked
     ''' </summary>
     ''' <param name="sender">ToolStripItem or LinkLabel only</param>
     ''' <param name="e"></param>
-    Private Sub OnMnuRecentOpenedFile_Click(ByVal sender As Object, ByVal e As EventArgs)
+    Private Sub OnMnuRecentOpenedFile_Click(sender As Object, e As EventArgs)
         Dim strFilePath As String = ""
-
+        'if tag is "more" then its more link that was clicked so exit sub
         If TypeOf sender Is ToolStripItem Then
```
lloyddewit, line 166: If we add an extra condition to this `If` then could we remove lines 167, 169, 170, 172, 177, 179, 184?
```suggestion
        If TypeOf sender Is ToolStripItem AndAlso toolStripItem.Tag.ToString() = "MRU:more" Then
```

instat/clsRecentFiles.vb
```diff
@@ -286,28 +220,150 @@ Public Class clsRecentFiles
...
+                End If
+            Next
+        End If
+
+        If ucrDataViewWindow IsNot Nothing Then
+            'remove all the data view window recent file menu items
+            ucrDataViewWindow.ClearRecentFileMenuItems()
+        End If
+
+        'then add and displays items (_in reverse order) for recent files
```
lloyddewit, line 262:
```suggestion
        'then add and display items (in reverse order) for recent files
```

instat/clsRecentFiles.vb
```diff
@@ -286,28 +220,150 @@ Public Class clsRecentFiles
...
+                    linkMenuItem.Text = strFileName
+                    linkMenuItem.Tag = strPath 'path used when the link is clicked
+                    'attach link click event handler for opening the file
+                    AddHandler linkMenuItem.Click, AddressOf OnMnuRecentOpenedFile_Click
+                    ucrDataViewWindow.InsertRecentFileMenuItems(linkMenuItem)
+                End If
 
-                ucrDataViewWindow.InsertRecentFileMenuItems(linkMenuItem)
+                If Not bShowAll Then
+                    'if recent files are more than 5 then just the "more" link label and exit loop
```
lloyddewit, line 300:
```suggestion
                    'if recent files are more than 5 then display the "more" link label and exit loop 

```

instat/clsRecentFiles.vb
```diff
@@ -286,28 +220,150 @@ Public Class clsRecentFiles
...
+    ''' <param name="mnuFileIcon"></param>
+    ''' <param name="ucrDataViewWindow"></param>
+    Public Sub UpdateRecentFilesMenuItems(bShowAll As Boolean, Optional mnuFile As ToolStripMenuItem = Nothing,
+                                          Optional mnuFileIcon As ToolStripSplitButton = Nothing,
+                                          Optional ucrDataViewWindow As ucrDataView = Nothing)
+        'exit sub if file toolstrip, file icon toolstrip and data view are not initialised
+        If mnuFile Is Nothing AndAlso mnuFileIcon Is Nothing AndAlso ucrDataViewWindow Is Nothing Then
+            Exit Sub
+        End If
+
```
lloyddewit, line 238:
```suggestion
```

instat/clsRecentFiles.vb
```diff
@@ -286,28 +220,150 @@ Public Class clsRecentFiles
...
-                linkMenuItem.Text = Path.GetFileName(strPath)
-                linkMenuItem.Tag = strPath 'path used when the link is clicked
+                Dim strPath As String = lstRecentOpenedFiles(iCounter)
+                Dim strFileName As String = Path.GetFileName(strPath)
+
+                'create new ToolStripItem, displaying the name of the file...
+                If mnuFile IsNot Nothing Then
+                    Dim clsItem As ToolStripMenuItem = New ToolStripMenuItem(strFileName)
+                    clsItem.ToolTipText = strPath
+                    'set the tag - will be used to identify the ToolStripItem as an most recent(MRU) item
```
lloyddewit, line 272:
```suggestion
                    'set the tag - will be used to identify the ToolStripItem as a most recent(MRU) item 
```

instat/clsRecentFiles.vb
```diff
@@ -286,28 +220,150 @@ Public Class clsRecentFiles
...
     ''' </summary>
-    Private Sub ShowAllRecentFiles()
-        'clear all menu items previously added 
-        ucrDataViewWindow.ClearRecentFileMenuItems()
-        'displays items (_in reverse order) for recent files 
-        Dim strPath As String
+    ''' <param name="bShowAll"></param>
+    ''' <param name="mnuFile"></param>
+    ''' <param name="mnuFileIcon"></param>
+    ''' <param name="ucrDataViewWindow"></param>
```
lloyddewit, line 230: Thanks for including these XML comments.
There should normally be a short description of each parameter

lloyddewit (changes requested):
Thanks for the changes, I agree with them all.
I have a few extra comments

instat/clsRecentFiles.vb
Patowhiz: @lloyddewit I will have to cast the `sender` for me to access the `Tag` function, I wanted to avoid the the "try cast" here. So I'm choosing to check the type then do a "1 time" direct cast to return the actual "type" so that I can access all the properties of the `ToolStripItem`.

Patowhiz:
@lloyddewit I have made the suggested changes above.

instat/clsRecentFiles.vb
lloyddewit: @Patowhiz thanks, understood.
I don't think we need a `Trycast` here. We could safely write the statement as:
```
If TypeOf sender Is ToolStripItem AndAlso 
    DirectCast(sender, ToolStripItem).Tag.ToString() = "MRU:more" Then
```
It is safe because the `AndAlso` will guarantee that the `DirectCast` will only be called if `sender` is a `ToolStripItem`.
I'll go ahead and approve anyway, only implement the above suggestion if you wish to.

lloyddewit (approved):
Thanks, all looks good. there's just one optional suggestion. I'll go ahead and approve anyway.

lloyddewit:

instat/clsRecentFiles.vb
Patowhiz: @lloyddewit aha! I understand you now. Yes I could do that, but I still need the `Tag` regardless of whether its "MRU:More" option or not. To later be used as a file path if not "MRU:More" option. So I need to always do `If TypeOf sender Is ToolStripItem` check then after that store the `Tag` then exit the subroutine if the `Tag` is "MRU:More". Same applies for the LinkLabel option.

I could avoid do the above if I can somehow get the file path directly from the sender object, which in this case not possible without explicit cast. Not unless I assume it to always be an object that has a `Tag` property and let VB.NET handle implicit casting though that could open doors to logical errors in future.

rdstern (approved):
Still looks fine!

instat/clsRecentFiles.vb
lloyddewit: yes, fair point, thanks for explaining

Issue #6512, opened 2021-06-04
Improve French translations


lloyddewit:
I suggest that we use this PR to document French translation issues that we come across.

In PR #6510, @rdstern wrote:

> 
> 
> I did find some curious translations in passing, but they aren't for here! log in the 2-variable model is translated as journal (like the log of activities, rather than the log function.) So I checked in the calculator and it is fine there - log stays as log. However the next key is tan and that is translated as brun (colour tan!)

lloyddewit:
Comments from @rdstern in PR #6513:
> 
> 
> This looks good. I saw you solved your remaining issue of being able to add the translation to the calculation keyboard, which is where I started.
> This showed (possibly) a potential problem for those translations. The first is sqrt that is translated as racine. The tooltip is `frsquare root. For example sqrt(3) = 3 ^ 0.5 = 1.732`.
> 
> Another example is ceiling, translated as plafond. The tooltip is `frInteger above the given value. For example ceiling(3.5)=4, ceiling-3.5 = -3`
> 
> In each case we want to translate the first part, but we don't want to translate the formulae. So `ceiling` can be translated, but not `ceiling(` . Or perhap[s nothing when the sentence has an = sign.
> 
> I then looked at the column summaries. Here there isn't the fr and instead the translation is implemented - fine.
> 
> ![image](https://user-images.githubusercontent.com/11226469/120968283-907f9400-c760-11eb-89f8-4f4afaf424f4.png)

lloyddewit:
The Toolbar Help button tool tip text is incorrectly translated into French (see PR #6529).
To fix this, the CrowdIn translation of 'He&lp' needs to be correctly translated (e.g. to '&Aide').
@dannyparsons do you know why the text of some tool bar buttons conatins an '&'?

lloyddewit:
This spreadsheet was added to PR #6580 by @rachelkg . It describes additional open French translation issues.

[R-Instat.translation.changes.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/6830937/R-Instat.translation.changes.xlsx)

lloyddewit:
> 
> 
> The Toolbar Help button tool tip text is incorrectly translated into French (see PR #6529).
> To fix this, the CrowdIn translation of 'He&lp' needs to be correctly translated (e.g. to '&Aide').
> @dannyparsons do you know why the text of some tool bar buttons conatins an '&'?

[...]

PR #6513, opened 2021-06-06
Added translations for (sub)dialog tool tips

lloyddewit:
Partially fixes issue #3721
Translates tool tips associated with (sub)dialog controls. I still need to investigate tool bar tool tips.
Please note that this PR includes a test database that 'translates' tool tips into French by adding 'fr'. I will create a separate PR for the real French CrowdIn translations.

@shadrackkibet Almost all the changes are in the [TranslateWinform Nuget package](https://github.com/lloyddewit/TranslateWinForm) so there is not much to review in this PR. You'll get an opportunity to review the Nuget package code in the coming weeks.


rdstern:
This looks good.  I saw you solved your remaining issue of being able to add the translation to the calculation keyboard, which is where I started.
This showed (possibly) a potential problem for those translations.  The first is sqrt that is translated as racine.  The tooltip is `frsquare root.  For example sqrt(3) = 3 ^ 0.5 = 1.732`.

Another example is ceiling, translated as plafond.  The tooltip is `frInteger above the given value.  For example ceiling(3.5)=4, ceiling-3.5 = -3`  

In each case we want to translate the first part, but we don't want to translate the formulae.  So `ceiling` can be translated, but not `ceiling(` .  Or perhap[s nothing when the sentence has an = sign.

I then looked at the column summaries.  Here there isn't the fr and instead the translation is implemented - fine.

image

I like it.  Now the task is to use the tooltips more!

lloyddewit:
@rdstern Thank you for the test. If you are happy with the mechanism, then please could you approve?

I think that all your comments above can be solved by the CrowdIn translations. So  moved these comments to issue #6512

rdstern (approved):

lloyddewit:
@shadrackkibet please could you approve and merge? thanks

shadrackkibet (approved):

shadrackkibet:
@lloyddewit **Translation (mechanism)** is part of the labels to be used in PRs? I thought we only use it for issues. This is the first PR with such a label.

PR #6517, opened 2021-06-08
Bug fixes: ensure comboxes to remember states in reopen of bar chart dialogue

### LINKED ISSUE #6135: bar charts needs corrections and possibly single_multiple option
I have indicated @dannyparsons, not that I expect him to do this - and it isn't urgent - but to confirm the two points below a) that it is easy? b) Good to not disable the option buttons.

There are 2 bar chart dialogues.  One in the Describe > Specific and the other in Describe > General > Use Summaries.
In the other graphs the main receiver has a Single/Multiple option.  This is absent from the bar charts.  I needed it recently and so had to do stacking first.  If simple, then please could we add this feature?

At the same time the Multiple option disables the Plot Options and also the special options for that type of plot.  I still use those by switching between single and multiple.  Could this "feature" be reviewed?

[ghana_2_stations_for_rainnormals.zip](https://github.com/africanmathsinitiative/R-Instat/files/5746266/ghana_2_stations_for_rainnormals.zip)

Here is the data file I used.

Here is an example of the graph I wanted:
image

N-thony:
Partially fixes #6135
This PR fixes the bug in the two comboxes to remember the states in reopen, and the presentation of the labels.

### DIFF
--- a/instat/dlgBarAndPieChart.resx
+++ b/instat/dlgBarAndPieChart.resx
@@ -127,7 +127,7 @@
   </data>
   <assembly alias="System.Drawing" name="System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
   <data name="lblByFactor.Location" type="System.Drawing.Point, System.Drawing">
-    <value>252, 221</value>
+    <value>252, 255</value>
   </data>
   <data name="lblByFactor.Size" type="System.Drawing.Size, System.Drawing">
     <value>103, 13</value>
@@ -148,7 +148,7 @@
     <value>$this</value>
   </data>
   <data name="&gt;&gt;lblByFactor.ZOrder" xml:space="preserve">
-    <value>10</value>
+    <value>12</value>
   </data>
   <data name="cmdOptions.ImeMode" type="System.Windows.Forms.ImeMode, System.Windows.Forms">
     <value>NoControl</value>
@@ -175,7 +175,7 @@
     <value>$this</value>
   </data>
   <data name="&gt;&gt;cmdOptions.ZOrder" xml:space="preserve">
-    <value>9</value>
+    <value>11</value>
   </data>
   <data name="cmdPieChartOptions.ImeMode" type="System.Windows.Forms.ImeMode, System.Windows.Forms">
     <value>NoControl</value>
@@ -202,7 +202,7 @@
     <value>$this</value>
   </data>
   <data name="&gt;&gt;cmdPieChartOptions.ZOrder" xml:space="preserve">
-    <value>15</value>
+    <value>17</value>
   </data>
   <data name="rdoPieChart.Appearance" type="System.Windows.Forms.Appearance, System.Windows.Forms">
     <value>Button</value>
@@ -241,7 +241,7 @@
     <value>$this</value>
   </data>
   <data name="&gt;&gt;rdoPieChart.ZOrder" xml:space="preserve">
-    <value>8</value>
+    <value>10</value>
   </data>
   <data name="rdoValue.Appearance" type="System.Windows.Forms.Appearance, System.Windows.Forms">
     <value>Button</value>
@@ -277,7 +277,7 @@
     <value>$this</value>
   </data>
   <data name="&gt;&gt;rdoValue.ZOrder" xml:space="preserve">
-    <value>7</value>
+    <value>9</value>
   </data>
   <data name="cmdBarChartOptions.ImeMode" type="System.Windows.Forms.ImeMode, System.Windows.Forms">
     <value>NoControl</value>
@@ -304,7 +304,7 @@
     <value>$this</value>
   </data>
   <data name="&gt;&gt;cmdBarChartOptions.ZOrder" xml:space="preserve">
-    <value>14</value>
+    <value>16</value>
   </data>
   <data name="lblPosition.AutoSize" type="System.Boolean, mscorlib">
     <value>True</value>
@@ -313,7 +313,7 @@
     <value>NoControl</value>
   </data>
   <data name="lblPosition.Location" type="System.Drawing.Point, System.Drawing">
-    <value>252, 265</value>
+    <value>252, 297</value>
   </data>
   <data name="lblPosition.Size" type="System.Drawing.Size, System.Drawing">
     <value>47, 13</value>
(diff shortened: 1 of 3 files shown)


rdstern (changes requested):
@N-thony that's already much better.  The oddity now is that we have the Position control when the option is set to count, but not otherwise.
I suggest the right-hand-side is going to be a bit more complicated anyway, with the multiple options for x or y.  So could you move the Position control to the left hand side.  It can be above the Flip button and a single line, so the control is just after the label.
This will make the dialogue a bit longer, which is ok.  There will be more options to come.

N-thony:
> @N-thony that's already much better. The oddity now is that we have the Position control when the option is set to count, but not otherwise.
> I suggest the right-hand-side is going to be a bit more complicated anyway, with the multiple options for x or y. So could you move the Position control to the left hand side. It can be above the Flip button and a single line, so the control is just after the label.
> This will make the dialogue a bit longer, which is ok. There will be more options to come.

@rdstern I'm not sure about it but after looking at the code, it seems to be a reason that I ignore for instance the Position control is visible when the option is set to count otherwise it is Y variable Receiver control which is at the same location with Position control. I wonder if it is good to move those two controls above Flip button. Can we have a quick chat?

N-thony:
@rdstern  I made the improvement regarding your suggestions in #6135 from item a) to d). Could you test? Thanks.

rdstern (changes requested):
@N-thony that is a reasonable start, but there is a long way to go.
1) In the second option  there needs to be the x-variable label and control.  You have lost that.
2) Then there is a lot to do on the R code that is generated.  Currently the only difference in the R-code is whether the orientation is vertical or horizontal.  The R-code for the value option does not have the "identiry" option any more.  You need to examine the different options in the current Version 7.2 dialogue as a start.

N-thony:
@rdstern I have improved the code and design regarding your comments above. Could you test the change? Thanks.

N-thony:
@rdstern I have improved the code and design regarding your comments above. Could you test the change? Thanks.

rdstern (approved):
@N-thony excellent, this all seems to be working!

rdstern (dismissed):
@N-thony this all seems to be working.  Well done!  
@lloyddewit there is more to be done, but I suggest it would be good to merge this first.

lloyddewit:

rdstern:
@N-thony there seem to be some problems of clearing part of the R-code.  I show the sequence that led to an error:
I ran a bar graph with my usual survey data.  Here is the command generated - this runs fine:

```
# Code generated by the dialog, Bar and Pie Chart
Sheet.1 <- data_book$get_data_frame(data_name="Sheet.1")
last_graph <- ggplot2::ggplot(data=Sheet.1, mapping=ggplot2::aes(x=village, fill=variety)) + ggplot2::geom_bar(position="dodge", stat="count") + theme_grey()
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="Sheet.1")
data_book$get_graphs(data_name="Sheet.1", graph_name="last_graph")
```
This ran fine:
Then I loaded a dataset from the standard R library: datasets.  The data file is called BOD



Here is the completed dialogue - it is very simple:
image

```
utils::data(package="datasets", X=BOD)

data_book$import_data(data_tables=list(BOD=BOD))

# Code generated by the dialog, Bar and Pie Chart
BOD <- data_book$get_data_frame(data_name="BOD")
last_graph <- ggplot2::ggplot(data=BOD, mapping=ggplot2::aes(x=Time, y=demand)) + ggplot2::geom_bar(position="dodge", stat="identity") + ggplot2::geom_text(mapping=ggplot2::aes(fill=variety), stat="count") + theme_grey() + ggplot2::theme(axis.text.x=ggplot2::element_text())
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="BOD")
data_book$get_graphs(data_name="BOD", graph_name="last_graph")

```

Here is the error:
image

You see, in the error and also in the code above, that lots of the code remains from the previous graph.

If I first press Reset, then it delivers - instead - the following code:

```
# Code generated by the dialog, Bar and Pie Chart
BOD <- data_book$get_data_frame(data_name="BOD")
last_graph <- ggplot2::ggplot(data=BOD, mapping=ggplot2::aes(x=Time, y=demand)) + ggplot2::geom_bar(position="dodge", stat="identity") + theme_grey()
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="BOD")
data_book$get_graphs(data_name="BOD", graph_name="last_graph")

```
This runs fine and gives the graph I want.

instat/dlgBarAndPieChart.vb
```diff
@@ -259,39 +247,38 @@ Public Class dlgBarAndPieChart
...
+                    ucrVariablesAsFactorForBarChart.Clear()
+                Else ucrVariablesAsFactorForBarChart.Add(clsParam.strArgumentValue)
+                End If
+            ElseIf clsParam.strArgumentName = "x" AndAlso (clsParam.strArgumentValue <> "value" OrElse ucrVariablesAsFactorForBarChart.bSingleVariable) Then
+                If clsParam.strArgumentValue = (Chr(34) & Chr(34)) Then
+                    ucrVariablesAsFactorForBarChart.Clear()
+                Else ucrVariablesAsFactorForBarChart.Add(clsParam.strArgumentValue)
+                End If
+            End If
+        Next
```
Ivanluv, line 262: when  go to the `Bar Chart options ` button  and` Return `when I am within the `Value` radio button, I get the error below    
image

N-thony:
@Ivanluv thanks for your comment above, I fix it is fixed now. Please have a look. Thanks.

Ivanluv:
@lloyddewit  this is good from my end

rdstern (approved):
@lloyddewit I think @Ivanluv is also ok with this.

shadrackkibet:
@N-thony could you improve the PR header? **Small bug fixes in Bar And Pie chart Dialog** is too general. Remember this header will be used to automatically generate release notes. Could you include the details of the bug fixed?

lloyddewit (approved):

instat/dlgBarAndPieChart.resx
```diff
@@ -232,25 +202,28 @@
...
   <data name="&gt;&gt;cmdPieChartOptions.ZOrder" xml:space="preserve">
-    <value>19</value>
+    <value>17</value>
   </data>
   <data name="rdoPieChart.Appearance" type="System.Windows.Forms.Appearance, System.Windows.Forms">
     <value>Button</value>
   </data>
+  <data name="rdoPieChart.Enabled" type="System.Boolean, mscorlib">
+    <value>False</value>
+  </data>
```
shadrackkibet, line 212: @N-thony @rdstern could you explain why the "pie-chart" option was disabled in this dialog? I am afraid these changes might have caused regression!

PR #6529, opened 2021-06-13
Added Tool Bar button tool tip translation (e.g. into French)

lloyddewit:
This PR translates tool tips on the tool bar buttons. It partially fixes issue #3721.

If the tool tip text already has a French translation, then this is shown. Else the translation is shown as 'fr...'.
If you would like changes to the French translations, then please document these in issue #6512. I already documented the incorrect tranlation of the 'Help' tool tip.
Please note that the 'View Last Graph' and 'Column Metadata' button tool tips don't currently translate (I think because they're a different button type). I'll fix this in a separate PR.
I will also update the English JSON file in another PR (so that the remaining tool tip text is translated).
Thanks

@shadrackkibet The changes are all in the database and TranslateWinForms library, so there probably isn't much you can review here.


shadrackkibet (approved):

lloyddewit:
The tool tip text to translate was added to the English JSON file in PR #6533.

rdstern (approved):
Nice

PR #6539, opened 2021-06-17
Export file dialog bug fix

### LINKED ISSUE #6530: Possible problem with the Export dialogue
This uses files from the dslabs package.  But I think the main problem is for any dataset.  I have now tried with the usual survey dataset from the library.

If I export a single file, then it seems to always insist it is a csv.  If I choose anything else, it always goes back to csv.

Please could this be checked.

The dslabs adds another problem with an awkward format.  The package isn't currently installed - will be in the next version.  It is easy to add.

I should state that this is a very messy file I am deliberately exporting.  But the main problem may be more general?

I have opened 2 data frames, namely `research funding rates` and also `raw data research funding rates`

I now go to File > Export 

If I export a single sheet, then it only seems to allow csv.  If I choose anything else it goes back to csv.  Is that a general problem.

If I have multiple sheets, exported as a single file - then it seems to work ok and I can export to Excel.

Patowhiz:
Fixes #6530.


instat/dlgExportDataset.vb
```diff
@@ -126,7 +126,7 @@ Public Class dlgExportDataset
             lblConfirm.Text = "Files with the same names will be overwritten." & Environment.NewLine & "Click Ok to Confirm the Export."
         End If
         TestOkEnabled()
-        autoTranslate(Me)
+        'autoTranslate(Me)
```
Patowhiz, line 129: @lloyddewit I wasn't aware that this line was added ( `autoTranslate(Me)` ). I think because of the translations that you have been working on?
When the subroutine is executed `cboFileType.SelectedIndexChanged` is raised. I think the subroutine also tries to translate the contents of the `combobox` in a way that raises the event.

I have temporarily commented the line to wait for you feedback. It's the one causing the bug explained in the issue by @rdstern .

lloyddewit: @Patowhiz `autoTranslate(me)` is called because the sub changes the label text and this text needs to be translated. `autoTranslate(me)` will update control text, which may trigger events. This is expected behaviour. 

Normally we should fix the code so that it can handle the triggered events. However in this case, there may be a simpler solution. The sub only seems to change the `lblConfirm` text. So we could potentially just translate the updated text and then delete the `autoTranslate(Me)` line. See example below. This would need testing.

If/when you test, please note that line 126 will not currently be translated because it contains a new line. The translation library doesn't currently translate multi-line text. Line 116 should translate correctly to French.

Hope this helps

image

Patowhiz: @lloyddewit thanks.

in regards to the multi-line text( line 126), can I do something like;

Patowhiz: @lloyddewit thanks.

in regards to the multi-line text( line 126), can I do something like;

`GetTranslation("Files with the same names will be overwritten.") & Environment.NewLine & GetTranslation("Click Ok to Confirm the Export.")`

Would the above alternative translate?

Patowhiz: Looking at the translate api, it appears that it depends on whether the "Files with the same names will be overwritten." exists in the data source.

So briefly; the translation works by getting an entire sequence of characters(could be a word or an entire sentence) and tries to match it with an equivalent 'whole' sequence of characters in the data source from which it uses to get the translation 'equivalent'.

Patowhiz:
@rdstern this is ready for your testing. Please also test whether the translation works.

instat/dlgExportDataset.vb
lloyddewit: @Patowhiz 
Yes, you are exactly correct. Your proposed solution should translate correctly (assuming we also add the substrings and their French/Portguese translations to the translation database).
However, R-Instat has many multiline strings that need to be translated. Therefore I need to find a general solution for this and apply it consistently across R-Instat. Therefore, I suggest that, for now, you just set line 126 as it's done in the screenshot.

instat/dlgExportDataset.vb
```diff
@@ -123,10 +123,10 @@ Public Class dlgExportDataset
                 strCommand = strCommand & "rio::export( x = data_book$get_data_frame(data_name=""" & strItem & """), file = """ & ucrFilePath.FilePath & "/" & strItem & GetSelectedExtension(cboFileType.SelectedItem) & """)" & Environment.NewLine
             Next
             ucrBase.clsRsyntax.SetCommandString(strCommand)
-            lblConfirm.Text = "Files with the same names will be overwritten." & Environment.NewLine & "Click Ok to Confirm the Export."
+            'lblConfirm.Text = GetTranslation("Files with the same names will be overwritten." & Environment.NewLine & "Click Ok to Confirm the Export.")
+            lblConfirm.Text = GetTranslation("Files with the same names will be overwritten.") & Environment.NewLine & GetTranslation("Click Ok to Confirm the Export.")
```
lloyddewit, line 127:
```suggestion
            lblConfirm.Text = GetTranslation("Files with the same names will be overwritten." & Environment.NewLine & "Click Ok to Confirm the Export.")
```

lloyddewit (changes requested):

instat/dlgExportDataset.vb
Patowhiz: @lloyddewit done.

lloyddewit (approved):

lloyddewit:

Patowhiz:
@rdstern this waiting for your review.

rdstern (approved):
Looks fine now - relief!

PR #6540, opened 2021-06-17
Correctly translate the 'View Last Graph' and 'Column Metadata' button tool tips

lloyddewit:
@lloyddewit added the following comment to PR #6529:

>Please note that the 'View Last Graph' and 'Column Metadata' button tool tips don't currently translate (I think because they're a different button type). I'll fix this in a separate PR.

This PR resolves the above bug.

@shadrackkibet I fixed this bug by changing the database, I don't think there's much you can review.


shadrackkibet (approved):

lloyddewit:
Replaced by #6544

Issue #6519, opened 2021-06-09
Variable names sometimes translated


rdstern:
This took quite a time to find.  Here is an example:
![image](https://user-images.githubusercontent.com/11226469/121331636-f9a50a00-c90e-11eb-8090-7249dd1d581b.png)

No, it is easy to get directly - it seems to be dataset dependent.  I tried with our usual survey data and never got the problem.
But starting R-Instat, then open samrain from the climatic guide datasets.  Run the bar chart graph with the option set to value, and year as the variable and (say) strt1 as the y-variable.  Runs fine.  Then return to the dialogue and it has replaced year by annee!

lloyddewit:
@rdstern Thanks for reporting this.
This is an instance of missing functionality as described in issue #3721 (3rd item down) :
![image](https://user-images.githubusercontent.com/57253949/121511693-347b7080-c9e9-11eb-8e05-1c4f9ede858b.png)

I can fix the specific issue above by tagging the 'Variable X' receiver as 'do not translate'. This is a one-row change in the database (no code change needed). I tested this and it works.

I can raise a PR for this, but I would prefer to fix this in a more generalized way. I'm thinking that by default, receiver text should **not** be translated. This is because the receiver should take over the text from a selector without changing anything. Does this make sense?

rdstern:
I agree that receiver text, for single and multiple receivers, should not be translated.  They contain the names of variables and they need to stay in their "neat" form.

lloyddewit:
Fixed in PR #6544

PR #6544, opened 2021-06-20
Ensure that Single Receiver text not translated

lloyddewit:
Fixes issue #6519.

In issue #6519 @rdstern wrote:
> I agree that receiver text, for single and multiple receivers, should not be translated. They contain the names of variables and they need to stay in their "neat" form.

This PR ensures that Single Receiver text is not translated. 
In the current implementation, Multiple Receiver text is anyway not translated so no changes were required for this.

This PR also includes some other changes:
- Added back Portuguese translations (these were accidentally removed in PR #6513)
- Removed all tool tip **test** translations. These will be added back correctly (i.e. with real translations) when the JSON file in PR #6533 is translated in CrowdIn
- Correctly translated the 'View Last Graph' and 'Column Metadata' button tool tips (and closed PR #6540)

@shadrackkibet Please could you approve? The problem was solved in the database (my personal notes below), so there isn't much you can review.

----
Preventing translation of Single Receiver text:
- Each single receiver has a `txtReceiverSingle` property
- In the `forms_controls` table, for each Single receiver, I set this property to `DoNotTranslate`.

image

Correctly translating the 'View Last Graph' and 'Column Metadata' button tool tips:
- In the `forms_control` table, set `mnuLastGraph` and `mnuMetadata` to `ReplaceWithDynamicTranslation`

image


shadrackkibet (approved):

Patowhiz:
@lloyddewit going by @rdstern analogy of receivers containing 'variables' text that shouldn't be translated. I think other controls could also have texts that are related to contents of data frames that shouldn't be translated(subject to @rdstern  correction). For instance, the drop downs for selecting the factor levels or file types, input text boxes that display file paths etc. Basically some custom controls can be used to display text related to something in the data frame that shouldn't be translated. 

@lloyddewit I haven't delved much into the translation implementation and the above examples could have already been factored in (in which case my concerns should be ignored!).

lloyddewit:
> 
> 
> @lloyddewit going by @rdstern analogy of receivers containing 'variables' text that shouldn't be translated. I think other controls could also have texts that are related to contents of data frames that shouldn't be translated(subject to @rdstern correction). For instance, the drop downs for selecting the factor levels or file types, input text boxes that display file paths etc. Basically some custom controls can be used to display text related to something in the data frame that shouldn't be translated.
> 
> @lloyddewit I haven't delved much into the translation implementation and the above examples could have already been factored in (in which case my concerns should be ignored!).

@Patowhiz Thank you for the comments. You are correct that many controls have text that should never be translated. Winforms allows us to add tags to controls. We were planning to use this feature to tag controls with 'do not translate'. This will be one of the next features we implement (see issue #3721).

lloyddewit:
@rdstern Please could you test that that Single Receiver text is no longer translated? and approve if satisfied? thanks

Patowhiz:
@lloyddewit thanks for the feedback. 
As a side note, I'm a bit skeptical on using the **winforms** `Tag` property for controls, because we already use it in other places for other functionalities(you probably are aware of this). I would generally lean towards adding a new property explicitly for this purpose.

rdstern (approved):

PR #6545, opened 2021-06-21
Included trace when counting number of days entered in data entry

### LINKED ISSUE #6515: Minor improvement needed on the data entry dialogue
@Patowhiz the data entry is working quite well.

One very minor item is the summary of the number of days when trace values are included.  Most of the different summaries are all ok, namely total, and number of rain days.  But the number of days needs to include the trace values.  Currently if I have 30 days with rainfall entered, with 22 zero, and 6 non-zero and 2 trace, the number of days is given as 28.  It should be 30, i.e. trace is counted.  And the mean per day should also divide by 30, not 28.

Patowhiz:
Fixes #6515.


rdstern (approved):
Good change for trace values

instat/sdgClimaticDataEntry.vb
```diff
@@ -633,6 +634,11 @@ Public Class sdgClimaticDataEntry
                 strValue = grdCurrentWorkSheet.Item(row:=i, col:=iColIndex)
             End If
 
+            'assume trace values to be 0 (for data entry uses) here
+            If strValue.ToUpper = "T" Then
+                strValue = 0
```
lloyddewit, line 639: We are setting a string to an integer here. Is that correct?

lloyddewit (changes requested):

rdstern:
I just report that the situation is when - in data entry - a trace rainfall is typed as t (or T), because that's what people would find convenient.  There are then various points at which we need to do something about it.  Trace, is a very small quantity (smaller than 0.1mm), but recognised as rain.  In the data entry we consider it like a value of zero, but a day that was entered - so it contributes to the count of days.  Later (when we write back to R-Instat, we transform it to 0.03mm.

instat/sdgClimaticDataEntry.vb
Patowhiz: Aha yes. It should be "0", I missed this because .NET just implicitly casts it to "0".

Patowhiz:
@rdstern thanks for clarifying. I have made inline code comments for future developers to understand this.

instat/sdgClimaticDataEntry.vb
Patowhiz: Corrected.

lloyddewit (approved):

lloyddewit:
@rdstern if you're happy to approve then we can merge, thanks

rdstern (approved):

PR #6546, opened 2021-06-21
Added extra buttons to simplify the use of the Script window

### LINKED ISSUE #6311: Add more buttons to simplify the use of the script window
I have been using the script window quite a lot recently and have become fairly skilled at <Ctrl><Alt><T> to run a selection.  I can run All by clicking at the top.  Could we have a few more icons or buttons?
a) Keep `Run All`
b) Add `Run Current LIne` - as in RStudio
c) Add `Run Selection`
d) Possibly add `Clear 
`
They could be buttons or possibly icons with a tooltip each?

This needs a bit of discussion.  If we can agree, then I assume it will be easy to do?

Patowhiz:
Fixes #6311.


lloyddewit:
@N-thony please could you review?
Thanks!

rdstern (commented):
Could @shadrackkibet give a second opinion.
a) I did discuss the positioning of the buttons with @Patowhiz 
But now I see them there, I find that they are a long way from the script (if the window is full width and the lines are short.  
I had got ok before with the Run All being in the middle, but that is ill-defined.
I wonder about Run and Run All on the left, with Clear staying on the right?
And just a question - it says CTRL + Enter as the short-cut, which I use a lot.  Does this same short-cut now work for a selection, because I am used to using Ctrl Alt T.  I'd be delighted if it does?  Happy if both are ok

Patowhiz:
@rdstern I have left the short cut as they are because it's a separate issue. I can have the Ctrl + Enter still run selections if you want, then we can replace the 2 right click menus with just 1. Just like we have for the button. But this will now also be a feature change not just an addition.

I can also have the buttons aligned to the left.

Would you like me to do the 2 tasks above and have it ready for review.

instat/sdgClimaticDataEntry.vb
```diff
@@ -633,6 +634,11 @@ Public Class sdgClimaticDataEntry
                 strValue = grdCurrentWorkSheet.Item(row:=i, col:=iColIndex)
             End If
 
+            'assume trace values to be 0 (for data entry uses) here
+            If strValue.ToUpper = "T" Then
+                strValue = 0
+            End If
```
shadrackkibet, line 640: Are these changes needed here?

shadrackkibet (dismissed):
@N-thony I would like to give you a chance to review as requested by @lloyddewit otherwise I am happy to review.

rdstern:
@Patowhiz I would prefer the 2 buttons on the left, and separate from the Clear Button, that can remain on the right.
If I understand correctly then there is a run all button which does as before.  And a run line button, which becomes a sun-selection button if there is a selection.  And these are consistent with actions in RStudio?
Then I am more relaxed about the action of the short-cuts, because I suggest more people will then use the buttons.  So, at least for now, let's leave the short-cuts as they are.

Patowhiz:
@rdstern ready for your review.

instat/sdgClimaticDataEntry.vb
Patowhiz: Removed. Thanks.

rdstern (changes requested):
In the first tooltip it states (ctrl + Enter) as the short-cut.  Delete this line, because it doesn't work for the Selection.  Then I'll be happy to approve.

N-thony:
> @N-thony I would like to give you a chance to review as requested by @lloyddewit otherwise I am happy to review.
@shadrackkibet no problem. We can both review.

N-thony (commented):
The codes look okay from my end. @shadrackkibet could you also review? Thanks.

Patowhiz:
@lloyddewit I think this is ready for your review. @rdstern has already reviewed this.

rdstern (dismissed):

shadrackkibet:
I have a few comments on the functionality. Some comments are to query the previous functionality. I couldn't find where these points are mentioned. Apologies if I am repeating.

- Running code doesn't work like in Rstudio. Currently the cursor moves through the empty lines. In Rstudio if the line is empty the cursor is moved to the none-empty line. 
- I am not sure we need two button i.e Run and Run All, I think we should have one button that provides both functionality i.e the button runs a line where the cursor is if no selection and runs the selection if there is (basically if you want to Run all then you have to select all). In future we will need icons therefore we will just have one icon. _Hicks law!_
- I would prefer the buttons on the left side like in Rstudio, some of our users have exposure to Rstudio so they would expect similar positioning while those who don't have Rstudio exposure are likely to move to Rstudio in the future. Therefore good to maintain that! - _Jacobs law!_
- I am not sure we need this message when running all.
image

Feel free to ignore these points, I don't think we need to act on them now.

Patowhiz:
@shadrackkibet thanks for your feedback, I discussed your suggestions with @rdstern, below are some of the conclusions we made;

> Running code doesn't work like in Rstudio. Currently, the cursor moves through the empty lines. In Rstudio if the line is empty the cursor is moved to the non-empty line.

Yes, currently, R-Instat recognises the R commands in a rather _crude_ way. @lloyddewit and I have discussed this and we think with incoming work  by Chris that will address this issue, it's best we leave the current implementation it as it is.


> I am not sure we need two buttons i.e Run and Run All, I think we should have one button that provides both functionality i.e the button runs a line where the cursor is if no selection and runs the selection if there is (basically if you want to Run all then you have to select all). In future, we will need icons therefore we will just have one icon. Hicks law!

@rdstern says that he uses the **Run All** button quite a lot and therefore he prefers it to stay, he prefers having it compared to having to select all the text every time a user wants to run all the text. Personally I'm okay with either.

> I am not sure we need this message when running all.

I agree. @rdstern what are your thoughts about this? I do understand the warning was added because the user may have modified the script code and doesn't wish to lose the changes before pasting them elsewhere. But we do have a _crude_ undo anyway.

> One of the key CUA design principles is that users must be able to do any action or task using either the keyboard or the mouse. I, therefore, suggest that all shortcuts should work including situations when you have a selection.

I agree, I have made the changes that implements this consistent.

@rdstern you can now test the new changes. I have merged the Run current line with Run selected text menu.

Patowhiz:
@rdstern please have a look at this so that it can be merged in this sprint.

rdstern (dismissed):
Great

instat/ucrScript.vb
```diff
@@ -88,13 +64,62 @@ Public Class ucrScript
...
+
+    Private Function RunText(strText As String) As String
+        Return If(strText <> "", frmMain.clsRLink.RunScriptFromWindow(strNewScript:=strText, strNewComment:=strComment), "")
     End Function
 
+    Private Sub cmdRunAll_Click(sender As Object, e As EventArgs) Handles cmdRunAll.Click
+        'temporarily disable the buttons incase its a long operation
+        cmdRunLineSelection.Enabled = False
+        cmdRunAll.Enabled = False
+        RunAllText()
```
lloyddewit, line 95: `RunAllText` is called twice in this control (see line 279). 
Ij the lines above, buttons are disabled during script execution, but on line 279, this is not done.
Should the 2 calls be consistent?

lloyddewit (changes requested):
Just one small question

instat/ucrScript.vb
Patowhiz: @lloyddewit thanks, I have refactored the code.

lloyddewit (approved):

lloyddewit:
@rdstern The changes should not have affected the functionality, if you test/approve again then we can merge, thanks!

rdstern (approved):

PR #6555, opened 2021-06-24
Ensure that special characters display correctly in the Data View window

### LINKED ISSUE #6024: Grid to display accents and other symbols?
image
[Station and Elements Niger.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/5393964/Station.and.Elements.Niger.xlsx)

The elements imports easily into R-Instat.  And the information is correctly imported, including accents and symbols like the degree symbol.  But they don't display in the grid.

I wonder if this is an easy thing to correct.  It comes in one of the videos we will be preparing.

### LINKED ISSUE #6483: A simple - maybe - problem with data imported from Excel
Here is a simple data set in Excel.
[Voluntary stations_muchinga.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/6531626/Voluntary.stations_muchinga.xlsx)
I would like to import the data into R-Instat and the get the mean of the latitude data.

a) I can do it easily in Excel.  Just use the sum button and ask for the mean.  
b) But I want to import into R-Instat and then analyse
c) When I import the lat has odd characters (in the grid view) and hence the variable is a character variable.
image

d) When I look in R, the odd character appears as a space. 
e) If I try converting to numeric, it tells me 30 values are non-numeric.
f) So what should I do?

Looking up the answer maybe[ here](https://stackoverflow.com/questions/52722846/how-can-i-remove-non-numeric-characters-from-strings-using-gsub-in-r):

Can I "tidy" these data already in R-Instat - with an existing dialogue?  Does it need a new dialogue or a new option in an existing dialogue?

My main aim is to get rid of odd symbols/characters in numeric variables.  But note that an extension of the same option could be useful for character variables - see the station variable here?

Patowhiz:
Fixes #6024.
Fixes #6483.

I recommend the code changes to also be looked at by @dannyparsons, he is the original author and I may have missed something in regards to his original code intentions.


rdstern (approved):

lloyddewit:
@Patowhiz Thank you for the change. I also read the associated issues. The change is simple (just one line of R commented out). I accept that this solves this specific issue, what I don't understand is why. I'm concerned avbout side effects. Do you have any ideas about why the line was included and what it was trying to achieve?

rdstern:
@lloyddewit and @Patowhiz I think the concern was that it might have a side effect on the display of numeric variables, when they are in R-Instat as character.  Here is an example:

image

Here the 5th variable (EvBac) is numeric, but because of an error in a single row - where there is a non-numeric value - the whole variable is interpreted as character.  This leads to some numbers being displayed oddly, as is shown above.

a) I don't mind that, because it is a useful alert that something may be wrong.  You can see that from the (C) at the top, but I am quite happy with this formatting problem too.
b) More important just now, ais that this display is the same as before - the changed line has not affected that.

So I am happy to accept and stand by my approval.  I don't know if @volloholic had any other worries.  If he would occasionally reply here, then we would know more!

Patowhiz:
@lloyddewit on the why, I strongly think it has to do with R.NET link and it's data frame object, it doesn't convert results of the R `format` function correctly in scenarios that I have outlined in the issue(also look at the screenshots). I can dive deeper to confirm exactly at what point things go wrong in the R.NET but that would take me more time.

The side effects are as explained by @rdstern above. I tested this multiple times and couldn't find any other, even within the logic boundaries.

 What I'm doing here is simply changing the contents of a parameter, to allow the characters be converted by `as.character` function and not `format`. 

I think @dannyparsons intended the R `format` function be used because of formatting the digits in character columns. @rdstern has mentioned above, that he is now okay with that side effect and I strongly agree with his reasoning. 

@volloholic other concern is the level at which I'm solving this problem. 

I chose to solve the problem at this level based on my analysis and understanding of of how the different R functions are used inside the entire R-Instat package, but I may have missed something that the author was keen on (understandably, the entire script has no comments).

Patowhiz:
The other concern was the issue of significant figures, it's not working as expected for numeric columns, but I confirmed that the bug is already there in the master. @rdstern we may have to test together confirm and possibly open another issue on the bug. And probably add a new column metadata for decimal places(another separate issue).

lloyddewit:
@rdstern, @Patowhiz Thank you for the explanation. I'll wait a couple of days to see if @dannyparsons or @volloholic respond. If not, then I'll try and have a quick chat with @Patowhiz and then approve.

rdstern (approved):
@lloyddewit I approved 3 weeks ago and you then said you would wait a day or two before discussing and approving.  I assume that can be done now.  I have forwarded your message to @dannyparsons , just in case he wants to show a sign of life on this.

@Patowhiz I am assuming the change in the latest reogrid are not relevant to this?

dannyparsons:
I don't think there is a problem with this change. The issue seems to be with the conversion of the `format()` character columns by R.NET as Patrick identified - I am not sure why this is an issue either but I don't think we need to fully understand it if we have a workaround.

The main difference between `format()` and `as.character()` is that `format()` adds extra spaces to the text to make the aligned "pretty", but I don't think this makes much of a difference for our display currently and if it solves other issues then it's worth the change.

My only request before merging is that this is checked for all our other main column types: logical, Date, POSIXct/POSIXt, ordered factor. because I think this change affects any non-numeric columns - but I don't expect there to be an issue.

If it works ok for these then happy for it to be merged. I suggest removing the current comment lines and adding a comment where `as.character()` is being used to explain why `format()` isn't being used for non-numeric columns.

lloyddewit:
> 
> My only request before merging is that this is checked for all our other main column types: logical, Date, POSIXct/POSIXt, ordered factor, because I think this change affects any non-numeric columns - but I don't expect there to be an issue.
> 
> If it works ok for these then happy for it to be merged. I suggest removing the current comment lines and adding a comment where `as.character()` is being used to explain why `format()` isn't being used for non-numeric columns.

@Patowhiz Please could you:
- test the other main column types @dannyparsons mentions above
- update the comments as suggested above

@shadrackkibet I'm on leave next week, when @Patowhiz has done the above, please could you approve/merge?

thanks!

shadrackkibet:
Noted.

Patowhiz:
@dannyparsons I have added the comment as suggested.
@rdstern I have tested the changes for logical, Date, POSIXct/POSIXt and ordered factors as suggested by @dannyparsons. 

@lloyddewit @shadrackkibet this is ready for you.

lloyddewit (approved):

lloyddewit:
@rdstern If you can retest/approve then we can merge, thanks

Patowhiz:
@rdstern please retest this so that we can start working on significant figures and decimals.

Patowhiz:
@rdstern please retest this so that I can have the paste special PRs ready for review. Thanks.

rdstern (approved):

PR #6561, opened 2021-06-28
Added back to back checkbox to the bar chart dialog

### LINKED ISSUE #6525: Add back-to-back bar charts to dialogue
This is one of three major improvements to the bar chart dialogue.

image

Population pyramids are an example of "back-to-back" bar charts.  Other examples are also shown in the figure.  I think they would be easier to do in situations where the left-hand side could just be negative, and this would perhaps be an option.  But good to have the possibility of both sides being positive as in the examples shown.

Ideally this would just be a checkbox labelled "Back to Back", but it may have to be more complicated than that.  It is only enabled when the optional factor is completed.  This second factor presumably can only have 2 levels.

N-thony:
Fixes #6525


instat/static/InstatObject/R/Rsetup.R
```diff
@@ -42,7 +42,7 @@ packs <- c("abind", "agricolae", "agridat",
            "hydroGOF", "hydroTSM", "imputeTS", "ini", 
            "insight", "intervals", "isoband", "jpeg", 
            "jquerylib", "jsonlite", "Kendall", "KernSmooth", 
-           "keyring", "klaR", "knitr", "labeling", 
+           "keyring", "klaR", "knitr", "labeling", "lemon",
```
shadrackkibet, line 45: This should not be edited manually. See the comment at the beginning of this file.

shadrackkibet (changes requested):
Please resolve the conflicts in this PR so that others can review this PR.

instat/static/InstatObject/R/Rsetup.R
shadrackkibet: If you would like this to be added as a new package then add/open an issue for it. Also, make sure you link to this PR.

N-thony: @shadrackkibet I think this is resolved. Please, have a look. Thanks.

rdstern (changes requested):
@N-thony I hoped to check, but there were build errors

PR #6586, opened 2021-07-06
Add lollipop charts to R-instat in a ggplot2 way 

### LINKED ISSUE #6523: Add dumbbell and slope charts to R-instat
In our graphs it would be good to provide more of the graphs that we keep seeing.  We need to "make-it-easy" but we should also stick to some principles, so we are adding in a ggplot sort of way.  I suggest we can do this by adding a few more buttons to the existing specific dialogues.  One example is the lollipop charts.  

See [here](https://www.r-graph-gallery.com/301-custom-lollipop-chart.html#baseline) for many examples, with the corresponding R code when you click on a graph.

As explained here, a lollipop chart is basically a geom_segment and geom_point.  

Hence I suggest we have 2 buttons (for now) for the Describe > Specific > ScatterPlot.  The first is Points and is the current dialogue.  I also suggest we change the name from scatterplot to Point plot.  This means the same thing, but the name then is the same as the geom.  (That's what we do unless there is a good reason to be different.)

Then we add Segment as a second button.

We already do this with some plots that go together - like boxplot, which has jitter and violin.

In this case the second aspect is that we could do the points first and then add the segments, or vice versa.  I suggest (here) we only need to "Add Points" to the segment option.  We don't bother to "Add Segments" to the points option.  

The principle is that once we have an initial geom we can always use the Plot Options to add layers.  But there are many popular graphs that have multiple layers.  Here is a good example, where we want many options for segments under that button, and then we often want to add points also.  

@lilyclements has R code for lollipop plots from her MSc days, and so will look at this item first on the R side.  Then I hope @haward can include this in the (newly named) point Plot dialogue.

@dannyparsons has included lollipop in the Procurement menu.  I assume possibly as a special bar chart.  I prefer as here, partly because it is easy to include as double, and also cherry plots.  They seem to be an alternative to bubble plots, Perhaps we should also consider adding size of the points to the main point dialogue?

HawardKetoyoMsatsi:
This is not ready for review.


shadrackkibet:
@HawardKetoyoMsatsi what is the state of this PR?

rdstern:
My thinking is that this work started before we discovered geom_lollipop.  Now we have changed and agreed to use that the work shanged.
a) geom_lollipop is now added Ias is geom_dumbbell.  
b) The lollipop charts will now be included as another checkbox (I hope) within the bar charts, rather than in the way originally specified.

So, either this pull request needs to morph into adding that, or (probably better) we close this pull request and start a new branch once the existing back-to-back has been merged.  I hope that is now (just about ready!)

HawardKetoyoMsatsi:
I am closing this pull request because it is supposed to be solved using geom_lollipop and geom_dumbbell.

PR #6587, opened 2021-07-06
Bug fix: Displaying correct filter preview, corrected naming of filters and use of Not(!) 

### LINKED ISSUE #6101: Add Not condition to filter
Could this be really easy to add.  For now I suggest it could be below the & and or option and we could get rid of the Remove Condition button for now.  It would just reverse the condition, by adding` !`  in front of the condition.

I give an example that I needed recently.  Climatic data from Garoua has 9 days when tmin > 100 degrees.  That is an easy condition to examine.  Then I would like the opposite, namely !(tmin > 100).  That is not the same as (tmin <= 100), because the NAs are also omitted.  So (without the NOT) I have to do (tmin <= 100 ) & (tmin !is.na).

That's not too bad.  But I actually had (tmin>100) | (tmax >100).

I can't see how to do the reverse of that in the filter.  Easy if NOT is available.

While making changes I notice that the preview of a condition such as (Tmin !is.na) in the sub-dialogue appears as just (Tmin).  It is fine in the main dialogue.

shadrackkibet:
fixes #6101. @rdstern this fixes the following.
- Inconsistency in the display of filter previews when using `is.na` and `!is.na` in the filter sub dialogs.
- Sorted out the naming of filters (data frames) from `Filter1, Filter2` etc to `filter1, filter2` etc.

I am still not sure how to add Not(`!`) in the current way we do the filters. Looks like it's not trivial as such!


rdstern:
@shadrackkibet this looks ok

image

Adding the not could we 

a) Change the order in this pull-down, so `is.na` and `!is.na` are at the top, followed by `==` and `!=`
Then the four <, <-, >, >= as now.  Followed by four new ones, namely  `!<, !<=, !> and !>=`

Then the only tricky bit is that you can't have (tmax !> 100), but have to write that as !(tmax > 100)

rdstern (dismissed):
@shadrackkibet I am approving, but have also suggested how you could easily (I think) ass NOT.

shadrackkibet:
Thanks for the suggestion. Operators like `!<, !<=, !>` and `!>=` do not exist in R like `!=` which is not `!==`, therefore I do not think adding them into the drop-down is consistent with R. 

The challenge here is technical, for example, we currently run the filter as follows;
 
`# Create Filter subdialog: Created new filter
data_book$add_filter(filter=list(C0=list(column="TMPMAX", operation=">", value=25)), data_name="Dagoretti_JKIA", filter_name="Filter")`

`operation=">"` is where we give the operator but `!<` isn't recognized as an operator therefore it is not trivial as such. I think adding NOT will be a bit of work that is not yet clear to me. @dannyparsons any ideas on how I could approach this?

rdstern:
I'll check with @volloholic as @dannyparsons is travelling today.  I accept it isn't an operator.  How about `(not)>` etc.  That's to me more obviously not an operator.  It would also appear like that in the list of conditions and then be translated into `!(tmax > 100)` at the bottom.  I suggest that translation also makes the situation clear.  We can also have a tooltip to explain more clearly, either for the set of operators or for each.  Of course I'd be quite happy to add the tooltip to the suggestion above as well, so `!>` or `(not)>`n has a tooltip 
 `This negates the > condition  It becomes e.g. !(x > 100)`  
Or even
 `!> is not an R operator.  Instead it negates the condition and becomes e.g. !(x > 100)`.

shadrackkibet:
@rdstern could you check if (Not) work correctly in my current setup? I added it as a checkbox. What do you think?

rdstern:
This could be ok.  But: 
image
 It needs to cope with multiple conditions.

If you will do it this way, then when I add the 2nd condition the Not should then be unchecked again.  If checked, then it adds the Not to the second condition.  

Then - if you later want to uncheck the first condition you have to get rid of the 2nd.  That's ok.  The checkbox only works on the last condition.

shadrackkibet:
Actually, the filter preview here is not correct, I will try to update it correctly! The main filter dialog displays the correct preview. It should show for example  `!(yield<=20) & !(size<=3)` if you had **Not** checked for both conditions because that is what it does behind the scenes and that isn't the same as `!(yield<=20 & size<=3)` [of which in my understanding, we can't do with the current setup].

Ideally, the checkbox should add (!) so long as it is checked. If you don't want **Not** then you must uncheck it yourself not the program doing it!

shadrackkibet:
@rdstern  I have now implemented NOT(!) at two levels please test. 
- Applied to each condition and 
- Applied to all conditions
See below checkboxes.
image

rdstern (changes requested):
@shadrackkibet I am sorry this is taking you so long - there are so many other topics I would like to have you addressing.
A small point: I started with a factor variable and the not options didn't appear, until I had a numeric variable.  Once they appear they are there forever - which is fine.
I was hoping that the "for each" wouldn't necessarily do the NOT for all of them?  I may want it for the first and not the second?  But the examples I had would be OK (I think) with just the for all, so perhaps just have that option for now?

shadrackkibet:
My plan is to focus on sorting out the filter system again! My next task after this is merged would be adding to a filter. This will take a bit of time. A similar task is a selection for columns that will sort of apply the same principles! 

Happy to pause this and work on other tasks.

rdstern (dismissed):
@shadrackkibet great.  Very minor suggestion is that "Add Not for all" is now over complicated.  I would just have "Add Not" now.  Approved anyway

lloyddewit:
@Ivanluv Please could you review this PR? thanks

Ivanluv:
@lloyddewit  happy to review

Ivanluv:
@lloyddewit  the vb code looks okay from my end

instat/ucrFilter.vb
```diff
@@ -67,17 +68,19 @@ Public Class ucrFilter
         ucrFactorLevels.SetReceiver(ucrFilterByReceiver)
         ucrFactorLevels.SetIncludeLevels(False)
         ucrFactorLevels.bIncludeNA = True
-        clsFilterView.bForceIncludeOperation = False
+        clsFilterOperator.bForceIncludeOperation = False
```
lloyddewit, line 71: Would it be more readable to have this line with the other `clsFilterOperator` declarations (just after line 38)? Or is there a special reason to have it here?

lloyddewit (approved):
looks good, just one small optional comment

lloyddewit:
@rdstern Just one small change (to a label)  since your last approval, please could you test/approve again? thanks

rdstern (approved):

PR #6594, opened 2021-07-09
Added Tufte boxplots to the Boxplot, Jitter and Violin plot dialogue

### LINKED ISSUE #6251: Add boxplots without box when needed - from Tufte?
One example [here](https://stackoverflow.com/questions/2076370/most-underused-data-visualization) of a neglected graph is the adaptation of a boxplot, without the unnecessary box.

Here is an example - the code uses lattice, but you will get the idea:
```
library(lattice)
library(taRifx)
compareplot(~weight | Diet * Time * Chick, 
  data.frame=cw , 
  main = "Chick Weights",
  box.show.mean=FALSE,
  box.show.whiskers=FALSE,
  box.show.box=FALSE
  ).
```
The resulting boxplot is like this:
image

In ggplot2 I think this is a special geom, i.e. geom_tufteboxplot in the ggthemes package that is currently installed with R-Instat.  The latest version - later than in current R-Instat - has corrected some bits in this function.

This would be useful to test with the Ghana dataset.  I suggest - for the climatic boxplot - it might open the door to being able to perhaps have better boxplots for year, and also possibly to permit boxplots for both year and month together.  It is designed particularly for situations where there are many categorical levels, which is just what we have here.  

Useful to consider it also for the general boxplots dialogue - possibly just an option, or its own radio button even.  This needs investigation.  Really useful also for our teaching of "good graphics".

HawardKetoyoMsatsi:
Fixes #6251
@rdstern and @shadrackkibet take a look, I have added the tufte boxplots.


rdstern:
@HawardKetoyoMsatsi that's a good start.  Did you manage to get the Tufte boxplot to show outliers.  I checked on stack-overflow and found that you need to put stat="boxplot".  That should be the default, i.e. geom_tufteboxplot(stat="boxplot").  And that option should be added to the stat in the geom.
Could you also add coef (same default of 1.5 as for boxplot.  I couldn't see it documented anywhere, but it works fine.
Then I suggest your default could be median = "line".
In addition you need to check the code when any of the other options are used.
a) The proportional to size has no meaning and should be hidden.
b) The next 3 options should all work fine, except you then add an ordinary boxplot to the code.  Instead, if Tufte is selected, then put the tufte boxplot after the adjustment, where you currently have the additional ordinary boxplot.
c) In connecting the mean/median, change the width from 1.5 to 0.7.  That's at least for Tufte.

lloyddewit:
@N-thony please could you also review this one? thanks

N-thony:
> @N-thony please could you also review this one? thanks

@lloyddewit sure, happy to review.

HawardKetoyoMsatsi:
@rdstern  thanks so much for the observation, The three parameters are being added in a sub dialogue and it is unfortunate that when the geom_tufteboxplot() was added, the stat = "boxplot" option was left out we will have to add it in a new or. The bellow error will appear when you click on the button "Tufte Boxplot Options" but once we add the option (stat = "boxplot") in the subdialog the error will disappear.
image
Therefore we have to add  (stat = "boxplot") under geom_tufteboxplot.

rdstern (changes requested):
@HawardKetoyoMsatsi it seems to work fine now.

Will you be adding the boxplot option to the geom in this pull request, or in a new one?

When you do, then remember to also add the coef as well - in the same place in the list as for the ordinary boxplots.

rdstern:
@HawardKetoyoMsatsi I still get the developer error, but I assume that is because the pull request hasn't been updated?  I think the new geom changes have now been merged, but perhaps not.

I did notice that you can't (currently for Tufte boxplots with the 2 factors.  The fill factor is not visible for the Tufte.  Fill, of course doesn't work with Tufte, but colour does as here?
image

So, I wonder, if we could keep the second factor in the boxplot dialogue, but if Tufte, then use colour rather than Fill?

Then I am ready to see this one merged.

HawardKetoyoMsatsi:
@rdstern I have made the above corrections I added second factor receiver to the boxplot and made the parameter name to be colour rather than fill, kindly take a look.

rdstern (dismissed):
@HawardKetoyoMsatsi looks good and happy to have a Tufte' style graph as an option.  Perhaps @shadrackkibet to check and merge?

instat/dlgBoxPlot.vb
```diff
@@ -419,6 +444,14 @@ Public Class dlgBoxplot
         End If
     End Sub
 
+    Private Sub ucrPnlPlots_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlPlots.ControlValueChanged, ucrChkTufte.ControlContentsChanged
+
+    End Sub
+
+    Private Sub ucrVariablesAsFactorForBoxplot_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrVariablesAsFactorForBoxplot.ControlContentsChanged
+
+    End Sub
```
shadrackkibet, line 453: Remove?

instat/dlgBoxPlot.vb
```diff
@@ -380,8 +401,12 @@ Public Class dlgBoxplot
...
-            cmdBoxPlotOptions.Text = "Boxplot Options"
+        If rdoBoxplotTufte.Checked Then
+            If ucrChkTufte.Checked Then
+                cmdBoxPlotOptions.Text = "Tufte Boxplot Options"
+            Else
+                cmdBoxPlotOptions.Text = "Boxplot Options"
+            End If
         ElseIf rdoJitter.Checked Then
             cmdBoxPlotOptions.Text = "Jitter Options"
         Else
```
shadrackkibet, line 412: Combine with the sub above?

instat/dlgBoxPlot.vb
```diff
@@ -356,11 +369,19 @@ Public Class dlgBoxplot
     End Sub
 
     Private Sub SetGeomPrefixFillColourAes()
-        'Sets geom function, fill and colour aesthetics and ucrsave prefix
-        If rdoBoxplot.Checked Then
-            ucrSaveBoxplot.SetPrefix("boxplot")
-            ucrSecondFactorReceiver.ChangeParameterName("fill")
-            clsCurrGeomFunc = clsBoxplotFunc
+        'Sets geom function, fill and colour aesthetics, ucrsave prefix and stat summary parameters
+        clsStatSummary.AddParameter("size", 1.5)
```
shadrackkibet, line 373: Enforce parameter positioning here elsewhere.

shadrackkibet (changes requested):

HawardKetoyoMsatsi:
@shadrackkibet I have fixed the above please take a look.

shadrackkibet (approved):
@rdstern functionality shouldn't be affected. Could you re-test and re-approve?

rdstern (approved):

PR #6605, opened 2021-07-14
Added missing parameters for geom_tufteboxplot and geom_density_ridges

### LINKED ISSUE #6592: Adding more layers and geom parameters for density ridges
When geom_density_ridges, geom_density_ridges_gradient and stat_density_ridges were added to R-Instat, aes() and some geom() parameters added were the ones that feature in the package's documentation. We have other parameters that even though they are not included in the documentation/ manual for this package, are frequently used. I would like to suggest that these frequently used layers and parameters to be added even if they don't feature in the documentation of the ggridges package. These include :-
1. scale_fill_viridis() a function that is used alongside other functions like geom_density_ridges_gradient() .
2. geom, quantile_lines, quantiles, calc_ecdf ( parameters for stat_density_ridges())
3. jittered points, alpha, scale, point_shape,point_size, point_alpha etc (can be passed as parameters,for geom_density_ridges())
4.

HawardKetoyoMsatsi:
Fixes #6592
@rdstern I have added the geom parameters that had not been added to geom_tufteboxplot and geom_density_ridges yet they are frequently used.


rdstern (changes requested):
@HawardKetoyoMsatsi great, but can you also add coef as is there for boxplots.  Put it just before the outlier shape.  Perhaps (at the same time) can you change the order of outlier shape, colour, stroke, so their order is the same as for boxplots.  I notice there is size for Tufte, but missing for the ordinary boxplot.  Perhaps (at the same time) add it in for the ordinary boxplot.

HawardKetoyoMsatsi:
> @HawardKetoyoMsatsi great, but can you also add coef as is there for boxplots. Put it just before the outlier shape. Perhaps (at the same time) can you change the order of outlier shape, colour, stroke, so their order is the same as for boxplots. I notice there is size for Tufte, but missing for the ordinary boxplot. Perhaps (at the same time) add it in for the ordinary boxplot.

@rdstern this is now fixed

rdstern (changes requested):
@HawardKetoyoMsatsi that's great.  The one problem is that in dumbbell and lollipop you have restriced both the x, y and xend to be factors.  I don't think that's necessary for these geome.  
Then I would like to merge this one quickly so we can proceed with the front end.

HawardKetoyoMsatsi:
@rdstern I have made the requested changes.

lloyddewit (approved):

lloyddewit:
@rdstern If it's OK from your side, then we can merge, thanks

shadrackkibet (approved):

rdstern (approved):
Great.  I hope this means the tufte boxplots can now also be merged?

PR #6607, opened 2021-07-14
Bug fix: disable reogrid settings to not view text overflow

### LINKED ISSUE #6206: Small bug in the display of the column metadata
This is certainly in the Column metadata and possibly also in the data frame metadata.

image

The first 2 fields in the column  (and dataframe metadata are the variable (or data frame) name, followed by the label.  When the label is empty - as it often is, then the name field spreads out and sometimes into the following fields.  If there is something in the label field, then it is sensibly truncated.

N-thony:
Fixes #6206 
I have added the code @Patowhiz suggested and fixed the bug in issue #6206. For this code to run, I updated the reogrid control to the latest version, and fixed the small error in ucrFactor by changing AppendCols to AppendColumns as called in the latest version of reogrid.


Patowhiz:
@rdstern @lloyddewit I think we want this behaviour in all grids, don't we? If so, then I think it needs to be implemented in a general way that affects all grids. This just means adding the single line of code in one of the functions related to filling grids.

@N-thony  ignore this comment until we get feedback from @lloyddewit and @rdstern .

rdstern (dismissed):
Great @N-thony and @Patowhiz it seems fine now!  I wonder if this is new code, or new version or a mixture?

rdstern:
Aaaah, @Patowhiz and @N-thony you remind me that I approved the change now, but didn't check on the data frame metadata.  That had the same problem before.

N-thony:
> Aaaah, @Patowhiz and @N-thony you remind me that I approved the change now, but didn't check on the data frame metadata. That had the same problem before.

I think there is no problem, these changes will also be applied to data frame metadata.

Patowhiz:
@rdstern another test would be data view window last column, import a dataset with a few columns and one whose last column has long character contents. I think overflow should also be disabled.

instat/clsGridLink.vb
```diff
@@ -388,6 +388,7 @@ Public Class clsGridLink
             fillWorkSheet.SetSettings(unvell.ReoGrid.WorksheetSettings.Edit_AllowAdjustRowHeight, False)
         Else
             fillWorkSheet.SetSettings(unvell.ReoGrid.WorksheetSettings.Edit_AllowAdjustRowHeight, True)
+            fillWorkSheet.SetSettings(unvell.ReoGrid.WorksheetSettings.View_AllowCellTextOverflow, False)
```
Patowhiz, line 391: This is an open comment.
@N-thony could you explore whether adding this line inside subroutine `FormatDataView(fillWorkSheet)` would be more viable than here. (It's being called in line 419 ).
If in your opinion it's better to have this line here, then just ignore this comment.

N-thony: It is also fine. I just moved it inside the subroute FormatDataView. Have a look. Thanks.

Patowhiz:
@lloyddewit @rdstern this looks good to me. 
@rdstern @lloyddewit  the sooner this new version of `reogrid` is merged the better because we can start looking into other improvements required in the front end like pasting.

lloyddewit (approved):

lloyddewit:
@N-thony , @Patowhiz many thanks

@rdstern Please could you test/approve? we can then merge, thanks

rdstern (approved):
Great!

PR #6621, opened 2021-07-18
Added mechanism to ignore specified UI controls when translating to a different language (e.g. to French)

### LINKED ISSUE #3721: Initial multilingual translation

### LINKED ISSUE #6532: Minor problems with the split dialogue
[data to split.zip](https://github.com/africanmathsinitiative/R-Instat/files/6651387/data.to.split.zip)

These data need to be split:
Here is the dialogue in Prepare > Column: Text > Split

Note I have masse the prefix into x and am using a space to split the components.
a) The first time I ran this, it worked - well almost:

Here is the command it generated:

```
# Code generated by the dialog, Split Text Column
V1 <- data_book$get_columns_from_data(data_name="data", col_names="V1", use_current_filter=FALSE)
split <- stringr::str_split(string=V1, pattern=" ", n=Inf, simplify=TRUE)
data_book$add_columns_to_data(data_name="data", col_name="split", col_data=split, use_col_name_as_prefix=TRUE, before=FALSE)
```
It correctly identified the pattern = " " , but it ignored my request for the prefix geing x and insisted it still be split.  So it generated columns split1, to split8.

b) I deleted the variables and ran again.  

The dialogue still looks as shown, but now, although space is shown in the dialogue it has another bug.  The code now is as follows:

```
# Code generated by the dialog, Split Text Column
V1 <- data_book$get_columns_from_data(data_name="data", col_names="V1", use_current_filter=FALSE)
split <- stringr::str_split(string=V1, pattern="Comma ", n=Inf, simplify=TRUE)
data_book$add_columns_to_data(data_name="data", col_name="split", col_data=split, use_col_name_as_prefix=TRUE, before=FALSE)
```

It now says pattern = "Comma " in the code,  i.e. Comma + <space> and so doesn't now work well at all!

### LINKED ISSUE #6581: Fill > New Data Frame has oddity in French
I was checking the language and in French.
From the front menu (I had switched back to English, but don't think this is relevant?
I went into the File > New dialogue and the last of the top buttons and pressed Ok  It should generate a blank data frame with 2 variables and 10 rows:

Here was the error message
image

In the output window it generated the commands as follows:
```
# Code generated by the dialog, New Data Frame
données <- data.frame(data=matrix(data=NA, nrow=10, ncol=2))
data_book$import_data(data_tables=list(données=données))

rm(données)
```
I assume it is the translation of data to donnees that is the problem.  If I recall the last dialogue and press Ok again then it is fine, giving:
```
# Code generated by the dialog, New Data Frame
data1 <- data.frame(data=matrix(data=NA, nrow=10, ncol=2))
data_book$import_data(data_tables=list(data1=data1))

rm(data1)
```
If I return again to the dialogue and press on another radio button and then back again to the last one, then it gives the same error as above

lloyddewit:
Fixes #6581 
Fixes #6532 
Partially fixes #3721 

Added mechanism to ignore specified WinForm controls when the application or dialog is translated into a different language.
This will prevent the text from being translated. For example, if a control's text represents the name of an R parameter, then this text should not be translated.

@rdstern , @N-thony ,  Please could you review 'translateIgnore.txt' to make sure that the proposed approach is OK, and adequately explained. Updating the database from this file currently requires manual steps, but I'll automate this (maybe in a separate PR).
@rdstern Please could you also test that this fixes issue #6581 , thanks

@rachelkg You are also welcome to review if you're interested (optional)

@dannyparsons We discussed this before and you suggested that I could potentially use the Winform tag feature to label controls that should not be translated. However many dialogs use the tag for other purposes and I would have to rewrite these dialogs. The approach above also has the advantage that it allows groups of controls to be specified (using pattern matching), rather than having to manually specify each control individually. I hope you're Ok with this approach, we can discuss if needed.

@Patowhiz In one of your previous PR comments, you warned me that the 'control tag' approach may not work  - you were correct!

### DIFF
--- a/instat/Translations.vb
+++ b/instat/Translations.vb
@@ -289,28 +289,18 @@ Public Class Translations
             End If
 
             'execute the SQL command
-
-            'Dim con As New SqlConnection(Connection_String)
-            'Dim cmd As New SqlCommand(strSqlUpdate, con)
-            'Using con
-            '    con.Open()
-            '    Dim iRowsUpdated As Integer = cmd.ExecuteNonQuery()
-            'End Using
-
-            'Using conn As New SqlConnection(connetionString)
-            '    Using cmd As New SqlCommand(strSqlUpdate, conn)
-            '        cmd.CommandType = CommandType.Text
-            '        conn.Open()
-            '        Dim i As Integer = cmd.ExecuteNonQuery()
-            '        conn.Close()
-            '    End Using
-            'End Using
-
             Try
-                'connect to the SQLite database that contains the translations
+                'specify the path of the SQLite database that contains the translations (2 levels up from the execution folder)
+                Dim strDbPath As String = Directory.GetParent(Application.StartupPath).FullName
+                strDbPath = Directory.GetParent(strDbPath).FullName
+                'strDbPath = String.Concat(strDbPath, "\translations")
+                strDbPath = Path.Combine(strDbPath, "translations")
+                strDbPath = Path.Combine(strDbPath, "rInstatTranslations.db")
+
+                'connect to the database and execute the SQL command
                 Dim clsBuilder As New SQLiteConnectionStringBuilder With {
                     .FailIfMissing = True,
-                    .DataSource = GetDbPath()}
+                    .DataSource = strDbPath}
                 Using clsConnection As New SQLiteConnection(clsBuilder.ConnectionString)
                     Using clsSqliteCmd As New SQLiteCommand(strSqlUpdate, clsConnection)
                         clsConnection.Open()
@@ -330,7 +320,7 @@ Public Class Translations
         'accidentally in the release version. 
         MsgBox("The " & strPath &
                " ignore file was processed. The application will now exit.", MsgBoxStyle.Exclamation)
-        System.Windows.Forms.Application.Exit()
+        Application.Exit()
     End Sub
 
 End Class
--- a/instat/Translations.vb
+++ b/instat/Translations.vb
@@ -312,7 +312,7 @@ Public Class Translations
                     .FailIfMissing = True,
                     .DataSource = GetDbPath()}
                 Using clsConnection As New SQLiteConnection(clsBuilder.ConnectionString)
-                    Using clsSqliteCmd As New SQLiteCommand(strSqlUpdate)
+                    Using clsSqliteCmd As New SQLiteCommand(strSqlUpdate, clsConnection)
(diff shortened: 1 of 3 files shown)


lloyddewit:
On 10/08/21, I discussed with @dannyparsons . He thought that the proposals in this PR were good, and he was happy with them.
We discussed the alternative of using WinForm tags. He thought that:
- It may be possible to update dialogs that use WinForm tags in conflicting ways
- Developers may find it easier to add a tag using Visual Studio designer, compared to looking for the control name and adding an entry to the translate ignore file
- If a developer changes a control name, then this could have repurcussions in the translate ignore file
- However, he also saw the advantages of a translate ignore file and was happy to go in this direction
- We should consider adding a GitHub task to update the database automatically whenever changes to the translate ignore file are merged

rdstern (approved):
@lloyddewit that seems to be fixed.
Perhaps @shadrackkibet can also check and merge if ok?

lloyddewit:
@rdstern Thank you for approving.

@shadrackkibet By the end of the week, I'll commit some extra changes to this PR. I'll tell you when I'm done and ask you to review. Thanks in advance

shadrackkibet:
Thanks for the notification. Noted!

lloyddewit:
I tested with the following test cases:

- no specifications
- only negations
- ignore specifications without negations
- ignore specifications with negations

All tests passed. The patterns used for testing are below.

```
dlgAugment_ucrSaveNewDataFrame_%
dlgAugment_%DataFrame
dlgAugment_ucrBase_cmd%
!dlgAugment_ucrSaveNewDataFrame_btnColumnPosition

%lgWind%
!%lgWindrose%
```

lloyddewit:
@rdstern @shadrackkibet This is ready for review.
There is nothing you can test here because all the specifications in the the 'translateIgnore.txt' file are already in the Master database. Also, the only functionality added is the 'SetTranslateIgnore()' sub. This is currently commented out and not called. This is because it only needs be uncommented when the 'translateIgnore.txt' file is updated, and the changes need to be applied to the database. I already tested the code using the 'Db browser' tool and an uncommented version of the code (see above).

Therefore I suggest that @shadrackkibet reviews the code changes. If he approves, then I think it's safe for @rdstern to approve without taking any further action. We can then merge.
Thanks!

rdstern (approved):

shadrackkibet (approved):

PR #6630, opened 2021-07-20
Enabled flexible positioning of labels on bar charts 

### LINKED ISSUE #6549: Small improvement in geom_text and geom_label
I first put this improvement in a more general context.  We need to be able to add labels (and texts) to our plots.  Here is an example.  It is from the R Graphics cookbook which is available [here](https://r-graphics.org/index.html).

image

This shows the small improvement we need in geom_text and geom_label.  Here is a screenshot from the existing sub-dialogue for geom_text:
image

Looking at the code above for the graph, we already have position = dodge.  We need to add further options to include position = position_dodge(0.9).  So, in the pull-down, that should be added, i.e. position_dodge (I notice we already have position_jitterdodge(), which is similar.  There are others I assume.  

And when any of these are given, then we need to add a pull_down to be able to give the 0.9.  I assume 0.9 could usefully be the default (it is the default width of bars).

There is also position=position_stack where the obvious argument is hjust.  There is a second - reverse =  TRUE or FALSE
I am not sure we need position = position_jitterdodge() - that is mainly for jittered point, not for texts.  But position = position_nudge could be useful to add.  That has an x and a y to include.

So, I think we could consider getting rid of jitterdodge and include dodge, stack and nudge.  That is unless we have a single user control for this, which is included in more geoms?

shadrackkibet:
Fixes #6549 @rdstern this is perhaps the quickest way I could get the positioning of labels possible. Adding control to pass the `width` will require more time. So I have made the drop-down editable so that you can change the `width` parameter. From the user perspective, this is not the best! But now it is possible to position labels from the dialog (requires some bits of editing)?


rdstern:
@shadrackkibet I realise it may not be ideal.  At least for now I am quite happy with that.

image

Here is a graph I was able to produce.  I also had to add the vjust, which works fine as well.

I notice you have changed the order of the 2 tabs.  I am now having a doubt about that.  I liked that change when you are editing the bar charts, i.e. the individual geoms.  But the change is general.  So when I add a layer I also start with that view.  In that case I am much happier the way it was.  When we add a layer then the main things are the x and y etc.  So if it has to be the same in all instances, then I prefer what we had before.  When I choose a new geom, e.g. text, the first thing should be the x and y, and not the options.

You are reluctant (and I agree why) to allow typing into the control.  But in the graph above I was only able to get what you see, by editing the script.  If I could also type into the label field, that would be so nice.  I needed to change min_yield into round(min_yield,1).  I wonder if @dannyparsons et all would agree to that!

rdstern (approved):
I am approving, because that change is fine.  @shadrackkibet I don't know if the change in view of all layers is also part of this pull request.  If so, then perhaps that could be changed back first.

shadrackkibet:
@rdstern for the issue of rounding I would consider that a data preparation (wrangling) step i.e so you have to use the calculator first to create the column which you can then use for labels.

rdstern:
This may not be possible, but I would like to find a way of doing this (perhaps) through this text layer.

Here is the resulting graph:
image


I have added the Stat(count) which is an option.  That's not the label = stat(count), which is one of the items I can't see how to do easily.  But also I assume the x=village is automatic, but it says I need a y as well as a label?  I put anything I liked for the Y and the label and then edited those 2 into the `label=stat(count) in the script.  
`
```
last_graph <- ggplot2::ggplot(data=survey, mapping=ggplot2::aes(x=village, fill=variety)) + ggplot2::geom_bar(position="dodge", stat="count") + ggplot2::geom_text(mapping=ggplot2::aes(label=stat(count)), position=position_dodge(width = 0.9), vjust=-0.5, stat="count") + theme_grey() + ggplot2::theme(axis.text.x=ggplot2::element_text())
data_book$get_graphs(data_name="survey", graph_name="last_graph")
rm(list=c("last_graph", "survey
```"))

rdstern (approved):
I am ok with this to be merged.  @lloyddewit over to you

lloyddewit (approved):

PR #6655, opened 2021-08-02
Improved Frequencies dialog

### LINKED ISSUE #6640: Updating the "Frequencies" dialog
The summary tables function has been updated and is given on issue #6411 - as a result the frequencies table dialog needs to be updated.
Changes to be made to the summary tables dialog were outlined in issue #6502. Similar changes are given for this dialog - namely, there are two functions: `summary_table` and `mmtable2`.

Changes:
* Remove "Show Row Names", "HTML Table" checkboxes, "Page By" and it's input.
* If "Display Margins" is checked, then outer margins run. Hence if this checkbox is checked then run the parameters `include_margins = TRUE` (otherwise `FALSE`), and `margins = "outer"`.
(this is for the `summary_tables` function)
* The "Store Results" checkbox should only be visible if there is one column factor. If this checkbox is checked then `store_table = TRUE`, otherwise `store_table = FALSE`. Default is that it is unchecked.
* Assign this function to a string called "frequency_table"

The second function is `mmtable`. This can be set up in the same way as outlined in #6502. It has two parameters which are hard-coded in.

* The first parameter is `data`, this takes the value of the `frequency_table` so can just be hard coded in as `data = frequency_table`
* The second is `cells = value`

The rest of the set up for this is the same as for the summary tables dialog (see #6502).

Ivanluv:
fixes #6640
This is still in progress


Ivanluv:
@lilyclements  when using the summary and both margin options am getting the following error
image
Is there a way I can fix this?

lilyclements:
In the frequency dialog you can only do outer margins, at least for now.
I need to discuss and amend the code for summary margins.

Ivanluv:
@lilyclements  okay.Thanks

Ivanluv:
@lilyclements could you look at these and suggest any more additions or changes I can make

lilyclements (commented):
1. What is "Checkbox1" and the receiver by it? Should the receiver be disabled/hidden, and checkbox1? Or if not, can you name Checkbox1?
2. Can you align the group boxes a little more together?
3. Can you remove the margin group box. The only option is "Outer".
4. There is no option for the number of columns like in the Summary dialog. Can you add a "Column Factors:" control in and for it to work similar to the Summary Tables dialog?
5. Since there are no summary variables given, can you remove the "Treat Summaries as Further Factor" option, and the checkboxes that link to it when it is `TRUE` and `FALSE` except keep the "Display Summaries as Rows" checkbox.

rdstern (commented):
@Ivanluv and @lilyclements that looks good for a start.  I notice zero frequencies are omitted.  Can there be an option to include them?  And the percentages don't yet work.  But I know this is work in progress.

lilyclements:
@rdstern what of the percentages do not work? There shouldn’t be any lost functionality from the previous frequency tables
By zero frequencies are omitted do you mean combinations with no frequencies currently come up as a blank or NA, or do you mean something else?

rdstern:
@lilyclements and @Ivanluv it just doesn't work!

When I don't have percentages it correctly gives the counts!  Here is the code - without requesting the percentages:

```
# Code generated by the dialog, Frequency
frequency_table <- data_book$summary_table(data_name="survey_subset1", factors=c("village","variety"), store_results=FALSE, treat_columns_as_factor=FALSE, summaries=count_label)

last_table <- mmtable2::mmtable(data=frequency_table, cells=value) + mmtable2::header_top_left(variable=village) + mmtable2::header_left_top(variable=variety)
data_book$add_table(table_name="last_table", table=last_table, data_name="survey_subset1")
data_book$get_tables(data_name="survey_subset1", table_name="last_table")

rm(list=c("last_table", "frequency_table"))

`That's fine.  Then I tick to add percentages.  Here is an example - I tried different ones:

![image](https://user-images.githubusercontent.com/11226469/128629313-0a26db8c-9b5d-4135-b7a1-b697d14f3717.png)

The code generated remains identical to the above - so whatever I do I still just get the counts!

**Comment by Ivanluv** [comment]
@rdstern  I have fixed the added the frequency functionality

**Comment by rdstern** [comment]
@Ivanluv great.  I'll check

**Review by rdstern — CHANGES_REQUESTED** [review]
@Ivanluv and @lilyclements great.  The percentages now work.  

The one item to fix, before merging, at this stage, is the order of factor levels.  This is the same in the frequency and summary dialogues.  The order in the tables should be the same as the order of those factors.  The current tables insist on the default alphabetical order.

**Comment by Ivanluv** [comment]
@rdstern  I have fixed the ordering of factors

**Review by rdstern — APPROVED** [review]
@Ivanluv this now seems to work fine.  I am approving.
@shadrackkibet over to you.  We are now making real progress!

**Comment by shadrackkibet** [comment]
@N-thony could you review first.

**Review by lloyddewit — APPROVED** [review]

**Comment by lloyddewit** [comment]
@rdstern Please could you retest? thanks

**Review by rdstern — CHANGES_REQUESTED** [review]
@Ivanluv it now seems very solid.  I was really hoping to merge now, but one item to fix first is the display of margin totals.
The default is All, but you can type there.  However, even with the default I always get the margin title as NA - which is very confusing.  I was using the simple survey data set.   It should be obvious, whichever data you use.  I can add pictures if needed.  I hope it is simple to fix.

**Comment by Ivanluv** [comment]
@lilyclements  could you help me with this ,I think the error might be in the summary function in R

File: instat/static/InstatObject/R/Backend_Components/summary_functions.R [hunk 1]
```diff
@@ -1370,10 +1370,6 @@ DataBook$set("public", "summary_table", function(data_name, columns_to_summarise
   }
   shaped_cell_values <- cell_values %>% dplyr::relocate(value, .after = last_col())
 
-  # Converts factor columns to character so we can relabel values if needed
-  for (i in seq_along(factors)) {
-    shaped_cell_values[[i]] <- as.character(shaped_cell_values[[i]])
-  }
```
lilyclements, line 1376: To fix the problem highlighted by @rdstern can you add here:

```
for (i in seq_along(factors)) {
    levels(shaped_cell_values[[i]]) <- c(levels(shaped_cell_values[[i]]), margin_name)
  }
```

Ivanluv:
@rdstern  the margin total is now fixed

rdstern (approved):
Great!

lloyddewit (approved):

PR #6673, opened 2021-08-14
Added sample as a summary into column statistics dialog

### LINKED ISSUE #6646: Add the sample function to our summary system
I think this may be a very simple addition to the summary system.  I would like the summary to take a randome sample from each combination.  There is a potential complication, but we can do something (I think?)

I would like to include a random sample (at least of size 1.  I give an example to show what I want - I think it should be easy?
As in the textbook on [Introduction to Data Science](https://rafalab.github.io/dsbook/examples-of-algorithms.html) I use the dataset from the useful package, called HistData (that we should add!), called GaltonFamilies

There are 934 children from about 200 families.  

image

The exercise (in the book) is to examine the relationship in heights between son and father.  An initial bit of tidying gets the heights of a random son from each family.

The R code given in the book is as follows:

```
library(HistData)

set.seed(1983)
galton_heights <- GaltonFamilies %>%
  filter(gender == "male") %>%
  group_by(family) %>%
  sample_n(1) %>%
  ungroup() %>%
  select(father, childHeight) %>%
  rename(son = childHeight)
```

This produces a new data frame with 179 rows.
In R-Instat we can almost do this.  The interesting step (after filtering for all the male sons is to summarise for each family.

image

Then we can get the first son, or the last son here:  
image

Could we add a sample function (also from dplyr) with an updown - default is 1.  I don't mind if initially it only works for one!
If for more than 1, then it would have to produce more variables.

Then we have the additional argument Replace - that could be a checkbox, so unchecked means Replace = FALSE - the default.   

I wonder if there are other functions to add.  I was surprised to get so close!

shadrackkibet:
fixes #6646

@rdstern here is the addition of sampling into the summary system. Please test.

### DIFF
--- a/instat/sdgSummaries.vb
+++ b/instat/sdgSummaries.vb
@@ -170,7 +170,8 @@ Public Class sdgSummaries
         ucrChkCount.AddToLinkedControls(ucrInputComboCountTest, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:="==")
         ucrChkCount.AddToLinkedControls(ucrInputCountValue, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=0)
         ucrChkQuantile.AddToLinkedControls(ucrInputQuantile, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=0)
-        ucrChkSample.AddToLinkedControls(ucrChkSampleReplace, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True)
+        ucrChkSample.AddToLinkedControls(ucrChkSetseed, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrChkSetseed.AddToLinkedControls(ucrNudSeed, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=0)
 
         ucrInputN.SetLinkedDisplayControl(lblInputN)
         ucrNudFraction.SetLinkedDisplayControl(lblFractionTrimmed)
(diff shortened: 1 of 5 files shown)


rdstern (commented):
@shadrackkibet looks good.  Did you also add the Replace checkbox?  If that is to sample with replacement, then it has no relevance, because we are only taking a single observation.  It could (sometimes) be nice to be able to take say 3 random observations and then this would be relevant.  But that opens the same potewntial "can of worms" as when I asked Danny if we could have 2 percentiles at the same time!

shadrackkibet:
We could allow sampling many but then it will mean the other summaries are duplicated. For example, with size = 5 we will have something like this? 
image

rdstern:
aha.  Let's just stick with the 1 - otherwise it is different to all the other summaries and it isn't important enough for that.  At least not here!

shadrackkibet:
@rdstern done! You could test fully now. Thanks.

rdstern (changes requested):
@shadrackkibet this is working fine.  I think I would slightly prefer to not have the set seed by default.  Then I am ready to approve.

shadrackkibet:
@rdstern that is done now.

rdstern (approved):

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -1003,6 +1003,12 @@ summary_n_distinct<- function(x, na.rm = FALSE, ...) {
   return(dplyr::n_distinct(x = x, na.rm = na.rm))
 }
 
+# sample function
+summary_sample <- function(x, replace = FALSE, seed, ...){
```
lloyddewit, line 1007: Is '...' needed here?

instat/sdgSummaries.vb
```diff
@@ -170,6 +170,8 @@ Public Class sdgSummaries
         ucrChkCount.AddToLinkedControls(ucrInputComboCountTest, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:="==")
         ucrChkCount.AddToLinkedControls(ucrInputCountValue, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=0)
         ucrChkQuantile.AddToLinkedControls(ucrInputQuantile, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=0)
+        ucrChkSample.AddToLinkedControls(ucrChkSetseed, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
```
lloyddewit, line 173: rename to 'ucrChkSeed'?

lloyddewit (approved):
@shadrackkibet I just had a couple of questions but they're not important so I'll approve.
I'll also merge. If you want to update the code further, then we can always do this in a separate PR.

instat/static/InstatObject/R/Backend_Components/summary_functions.R
shadrackkibet: Yes, because there are some other unnamed parameters that could be passed to the sample function.

instat/sdgSummaries.vb
shadrackkibet: Sure.

Issue #6615, opened 2021-07-16
Add dot plots to the histogram dialogue


rdstern:
I am now following the order of graphs in the book on [Data Visualisation with R](https://rkabacoff.github.io/datavis/Univariate.html#dot-chart).  It has a section in chapter 3 on dotplots.  They are shown there as a simple alternative to a histogram - and particularly when the dataset isn't too large.

@HawardKetoyoMsatsi I am combining this addition for you to do, and also to use this example to introduce @Cal-Insta to being able to make simple improvements to a dialogue.  Here is what I suggest:
a) @HawardKetoyoMsatsi to add `Display as Dotplot `checkbox to the dialogue, just as you have done to add rifdgeplots to the density option on the same dialogue.
b) So, when the checkbox is checked the Histogram Options button will also change to become Dotplot Options.
c) @Cal-Insta to test the changes, using an example from his own data, and also the example from the book
d) The example from the book then considers options to make the plot better.  Check that these are all available from the sub-dialogue.
e) It could be useful to have some of these options available on the main dialogue.  Some will be equally useful whether the display is a histogram or a dotplot.  Others might be for just one or other?  If you think it would be useful, then propose which of these options could be on the main dialogue.  
f) Assuming there are some, then perhaps @HawardKetoyoMsatsi adds the first option and @Cal-Insta then adds the others.

HawardKetoyoMsatsi:
@Cal-Insta please how is the testing going on.

Cal-Insta:
@HawardKetoyoMsatsi the display as dot plot works ok. I think we can now look at the plot options.

PR #6675, opened 2021-08-16
Added Display as Dotplot checkbox to Histogram dialog

HawardKetoyoMsatsi:
@Cal-Insta @rdstern This will fix the first part of issue #6615, in case we find options to add, that will be done in another PR.

### DIFF
--- a/instat/dlgHistogram.designer.vb
+++ b/instat/dlgHistogram.designer.vb
@@ -55,6 +55,7 @@ Partial Class dlgHistogram
         Me.ucrBase = New instat.ucrButtons()
         Me.ucrPnlOptions = New instat.UcrPanel()
         Me.ucrChkRidges = New instat.ucrCheck()
+        Me.ucrChkDisplayAsDotPlot = New instat.ucrCheck()
         Me.SuspendLayout()
         '
         'cmdOptions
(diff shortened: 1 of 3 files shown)


Cal-Insta (commented):
This works well.

Cal-Insta (commented):
Some of the dotplot options I suggest we look at and add include: 
1. binwidth
2. method
3. stackdir
4. dotsize
5.stroke width
6. colour and fill

HawardKetoyoMsatsi:
> Some of the dotplot options I suggest we look at and add include:
> 
> 1. binwidth
> 2. method
> 3. stackdir
> 4. dotsize
>    5.stroke width
> 5. colour and fill

@Cal-Insta I guess this should better be put under issue #6615 not this PR. I find it to be a continuation of the discussion on the issue.

rdstern (changes requested):
@HawardKetoyoMsatsi @Cal-Insta that looks good so far.  But the y axis is different in the 2 cases.  It looks as though an option needs to be provided here, so counts can be given for the dot-plots.  At the same time, perhaps the current y-axis for the dot-plot could be given for the histogram option.

Also I assume the geom should change.  So when dotplots is checked, then the options should change from Histogram Options to dotplot Options?

@Cal-Insta has given a list of possible options.  This must go further and be much more precise as to hosk on the forthcoming dialogue.  Also, how many of those items will be just for the dot-plots and how many for both?

HawardKetoyoMsatsi:
@rdstern   I find the problem with the dot-plot giving inconsistent y-scale (kinda meaningless ) and one suggestion was to manually scale it to match the number of dots which depends on the data at hand.  
Could it be okay if we have it as it. Because if we manually scale it (as suggested by most people  [1](https://github.com/tidyverse/ggplot2/issues/1619) , [2](https://github.com/tidyverse/ggplot2/issues/2203) etc.), that cant be flexible to apply to different datasets?

HawardKetoyoMsatsi:
> @HawardKetoyoMsatsi Is this PR ready again for test/review? thanks

@lloyddewit it is not yet ready, part of @rdstern  changes are yet to be made.

rdstern (approved):
@HawardKetoyoMsatsi ok, let's consider the scale with the other improvements that @Cal-Insta is - I think - going to make.  You are right that we need to move on!.

HawardKetoyoMsatsi:
@lloyddewit Please you can review this. The scale and suggestions by @Cal-Insta will be done in a new PR as described in issue #6615.

lloyddewit (approved):

PR #6567, opened 2021-06-29
Copy from data view, columns and data frame grid to clipboard

Patowhiz:
Fixes #6071.
Fixes #5993.

@africanmathsinitiative/developers this is ready for review.

PR #6676, opened 2021-08-16
Bug fix: Passing the data name when copying data frames

shadrackkibet:
@rdstern this fixes the bug reported [here](https://github.com/africanmathsinitiative/R-Instat/pull/6567#issuecomment-891066484). Please test.


dannyparsons:
Looks good. Since it's a bug fix, it should be in the release notes.

shadrackkibet:
I don't think so since this bug was introduced by a PR that was recently merged and there has been no release since then.

PR #6679, opened 2021-08-17
Fixed Minor problems where Split By input keeps resetting back to "Comma ," in the split dialogue

### LINKED ISSUE #6532: Minor problems with the split dialogue

HawardKetoyoMsatsi:
Fixes #6532
@rdstern @africanmathsinitiative/developers this is ready for review.

### DIFF
--- a/instat/dlgSplitText.Designer.vb
+++ b/instat/dlgSplitText.Designer.vb
@@ -49,7 +49,7 @@ Partial Class dlgSplitText
         Me.ucrPnlSplitText = New instat.UcrPanel()
         Me.ucrSaveColumn = New instat.ucrSave()
         Me.ucrNudPieces = New instat.ucrNud()
-        Me.ucrInputPattern = New instat.ucrInputComboBox()
+        Me.ucrInputSplitByPattern = New instat.ucrInputComboBox()
         Me.ucrReceiverSplitTextColumn = New instat.ucrReceiverSingle()
         Me.ucrSelectorSplitTextColumn = New instat.ucrSelectorByDataFrameAddRemove()
         Me.ucrBase = New instat.ucrButtons()
@@ -128,13 +128,13 @@ Partial Class dlgSplitText
         Me.ucrNudPieces.Name = "ucrNudPieces"
         Me.ucrNudPieces.Value = New Decimal(New Integer() {0, 0, 0, 0})
         '
-        'ucrInputPattern
+        'ucrInputSplitByPattern
         '
-        Me.ucrInputPattern.AddQuotesIfUnrecognised = True
-        Me.ucrInputPattern.GetSetSelectedIndex = -1
-        Me.ucrInputPattern.IsReadOnly = False
-        resources.ApplyResources(Me.ucrInputPattern, "ucrInputPattern")
-        Me.ucrInputPattern.Name = "ucrInputPattern"
+        Me.ucrInputSplitByPattern.AddQuotesIfUnrecognised = True
+        Me.ucrInputSplitByPattern.GetSetSelectedIndex = -1
+        Me.ucrInputSplitByPattern.IsReadOnly = False
+        resources.ApplyResources(Me.ucrInputSplitByPattern, "ucrInputSplitByPattern")
+        Me.ucrInputSplitByPattern.Name = "ucrInputSplitByPattern"
         '
         'ucrReceiverSplitTextColumn
         '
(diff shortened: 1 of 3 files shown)


rdstern (approved):
@HawardKetoyoMsatsi greast - seems to be fixed.
@shadrackkibet if ok with you, then maybe it can be merged?

shadrackkibet (commented):
The only change I see is that the name of the control changed from `ucrInputPattern` to `ucrInputSplitByPattern`. Was this causing the bug? Otherwise, I couldn't see how the bug is fixed in the code.

HawardKetoyoMsatsi:
Yea I realized there was a problem  that with the control I had to change the name which fixed the issue. I it however appeared a little strange to me.

HawardKetoyoMsatsi:
@shadrackkibet I have looked at this issue together with @Ivanluv and it seems to be more complicated for us to fix. I would like someone to take over the problem and give a fix better than the temporary fix suggested in this PR.

shadrackkibet:
Thanks for the response. I will discuss this further with @lloyddewit when he comes back.

Please move on to another task.

lloyddewit:
> 
> 
> Thanks for the response. I will discuss this further with @lloyddewit when he comes back.
> 
> Please move on to another task.

I also looked at the code changes, and I can't see anything that would fix this bug. It's possible that:
- The bug is intermittent
- The bug has been resolved since the issue was created (as a side-effect of another change)

@rdstern Please could you try and recreate the bug using the latest master? If you can recreate the bug, then please could you perform the exact same steps on this branch to see if there is a difference?

Thanks

rdstern:
For me there remains a problem, but it seems now slightly different to before.  I use the same dataset given in #6532.
I change the character to a space, and the prefix to x.  It seems to run ok.  Here is the command generated:
```
# Code generated by the dialog, Split Text Column
V1 <- data_book$get_columns_from_data(data_name="data", col_names="V1", use_current_filter=FALSE)
x <- stringr::str_split_fixed(string=V1, pattern=" ", n=2)
data_book$add_columns_to_data(data_name="data", col_name="x", col_data=x, use_col_name_as_prefix=TRUE, before=FALSE, adjacent_column="V1")
```

I now realise the initial strin g is wider than I thought.  The new string x is Ok, but x1 (the other string in the split) still has some spaces remaining.

I return to the dialogue.  I now split x1.  Here is the dialogue.  It has now kept the space in the dialogue - which I don't think it did before.  So that's good.
image

I run this to give x2 and x3.  This doesn't work.

I examine the code.
```
# Code generated by the dialog, Split Text Column
x1 <- data_book$get_columns_from_data(data_name="data", col_names="x1", use_current_filter=FALSE)
x <- stringr::str_split_fixed(string=x1, pattern="Comma ", n=2)
data_book$add_columns_to_data(data_name="data", col_name="x", col_data=x, use_col_name_as_prefix=TRUE, before=FALSE, adjacent_column="x1")
```

So, although the dialogue said Space it now used Pattern = "Comma" in the command.

(If I press Reset, then it works OK - just once.  It then gives me x4 and x5 correctly with space as a split.  I then continue, because there are still more spaces in the remaining string.  But - although space( ) is now in the dialogue, it goes back to its default and puts a comma in the code.

HawardKetoyoMsatsi:
@lloyddewit @shadrackkibet , Please how do we go about this?.

rdstern:
@HawardKetoyoMsatsi I think this is all done and ok now

maxwellfundi:
@lloyddewit  this too needs your review :-)

lloyddewit:
This PR solves the issue by renaming the `ucrInputPattern` control. This has the side effect of bypassing the translation database. This is not the best way to solve this issue. The issue is best solved by updating the database and its associated 'ignore' file.
Therefore, I will close this PR and solve this issue in PR #6621 .

PR #6682, opened 2021-08-19
Simplified addition of labels from bar chart dialog

shadrackkibet:
@rdstern Please do some quick testing. I would like to merge then hand it over to @Cal-Insta to investigate the defaults we could've for other position options.

### DIFF
--- a/instat/dlgBarAndPieChart.Designer.vb
+++ b/instat/dlgBarAndPieChart.Designer.vb
@@ -50,6 +50,11 @@ Partial Class dlgBarAndPieChart
         Me.lblXvariable = New System.Windows.Forms.Label()
         Me.rdoPie = New System.Windows.Forms.RadioButton()
         Me.rdoDonut = New System.Windows.Forms.RadioButton()
+        Me.lblLabelPosition = New System.Windows.Forms.Label()
+        Me.lblLabelColour = New System.Windows.Forms.Label()
+        Me.ucrInputLabelPosition = New instat.ucrInputComboBox()
+        Me.ucrInputLabelColour = New instat.ucrInputComboBox()
+        Me.ucrChkAddLabels = New instat.ucrCheck()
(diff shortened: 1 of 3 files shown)


rdstern (approved):
@shadrackkibet I am happy to approve as it works in some circumstances, and does no harm.  I am warming to the idea of changing the special geom to Text Options when that is checked.  I wonder if you would like to include that before it is merged.
Then I assume you will be in touch with @Cal-Insta for him to take-over assessing on the next steps on the code for the labels, so it works in more instances.
I assume @dannyparsons needs to approve as @lloyddewit is still away.

rdstern (changes requested):
@shadrackkibet it is looking good now.
a) I think you mentioned that the stack reverse doesn't work well yet.  It has not reversed the labels.
b) I then tried to go to edit the text layer and it gave the dreaded developer error, before going in.
c) I then did a value plot as follows.  I added a variable from 1 to 36, using the row numbers.  Then had x as yield and y as observation number.  Then I can't get the labels at all.  It plots well without the labels.

shadrackkibet:
@rdstern a) and b) should be fixed now. However, I really doubt the usability of position options from plot options subdialog. Definitely, we've to improve on it if we find that users want to use further options to adjust labels.

rdstern:
@shadrackkibet this is now improved in that error.  It still gives a problem with the counts.  I used the sub-dialogue and found it now works.  

Then I tried to edit the text label.  I found the following:
image

I then returned, without changing anything.  It then didn't work, I assume because those 2 fields were now empty?

Unrelated to this is that I found a second complication.  I was using coord-flip.  IThe position of the label wasn't ideal when Out.  I changed to In, and found that it is changing vjust, when I think it should now be setting and changing hjust?

shadrackkibet:
@anastasia-mbithe could you have a quick look into this? Then we can try to fix the remaining issues together?

PR #6688, opened 2021-08-20
Added options for path and step plots into line plot dialog

### LINKED ISSUE #6651: Add step, path and smooth to the Describe > Specific > Line dialogue
- [x] The main suggestions here are to add geom_smooth.  But a small addition is to add geom_step and geom_path as options.  I suggest they be added like the density ridges option, i.e. a checkbox just below the receivers on the right.  This time there are 2, namely Step and Path.  I suggest they could go side by side as checkboxes.  Default is unchecked for either.  If either is checked, then the other can't be.  So, if you check the other, then it unchecks the other one.  
If either is checked, then the r-code simply substitutes geom_step or geom_path for geom_line.  These 2 geoms are already available.

It might be that the pull reuest first merges these changes.

- [x] The main (and big) change is to add Smooth as a button at the top.  This is an important geom.  I suggest as a button, because you may want it, without the data, i.e. the line.  And because there will be a different set of options for the smoothers.

I think the main receivers can be the same as for the line.  Then we also have the points and (now) line as checkboxes.  And I will add further options instead of the peaks/valleys.  I will add these later.

I have included @shadrackkibet on this issue, as I am sure he will be involved.  I am not sure who will do the main work.

- [x] Looking further, there are also some edits to be made to the smooth geom.  They need to be done anyway and are in a separate issue #6652 .  Then it should also be easy to add options for this geom to the main dialogue.  

a) Possibly add orientation to the dialogue.  The default is "x", but it can be "y".  The example [here](https://ggplot2.tidyverse.org/reference/geom_smooth.html) shows an example.

b) Include Method with a pull doan, and possibly Arguments on the same row also with a pull down.

c) Include Function with a pull-down 

d) Keep the Points checkbox - as in the Line button
e) Add a Line checkbox
f) Move the With Standard Error checkbox from the Line button

shadrackkibet:
fixes #6651

@rdstern this adds the step and path plots as options. Please check.

### DIFF
--- a/instat/dlgLinePlot.designer.vb
+++ b/instat/dlgLinePlot.designer.vb
@@ -45,6 +45,10 @@ Partial Class dlgLinePlot
         Me.cmdLineOptions = New System.Windows.Forms.Button()
         Me.lblFactorOptional = New System.Windows.Forms.Label()
         Me.lblGroupLine = New System.Windows.Forms.Label()
+        Me.rdoPath = New System.Windows.Forms.RadioButton()
+        Me.rdoStep = New System.Windows.Forms.RadioButton()
+        Me.ucrPnlStepOrPath = New instat.UcrPanel()
+        Me.ucrChkPathOrStep = New instat.ucrCheck()
         Me.ucrChkValley = New instat.ucrCheck()
         Me.ucrChkPeak = New instat.ucrCheck()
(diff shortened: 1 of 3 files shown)


rdstern (approved):
Great - 2 more geoms!

PR #6703, opened 2021-08-24
Adding the pivot table dialogue to the Describe>Two Variable menu

Ivanluv:
The PR adds the pivot table dialogue to the Describe> Two variable menu

### DIFF
--- a/instat/frmMain.Designer.vb
+++ b/instat/frmMain.Designer.vb
@@ -51,6 +51,7 @@ Partial Class frmMain
         Me.mnuDescribeTwoVariablesGraph = New System.Windows.Forms.ToolStripMenuItem()
         Me.ToolStripSeparator34 = New System.Windows.Forms.ToolStripSeparator()
         Me.mnuDescribeTwoVariablesFrequencies = New System.Windows.Forms.ToolStripMenuItem()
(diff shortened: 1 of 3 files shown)


rdstern:
@Ivanluv great - I am looking at that now.

Ivanluv:
@rdstern  This contains the form main changes only which shouldn't work. I have the main changes in another branch but I need this merged first for me to make a PR from the other branch. This is to reduce form main conflicts

rdstern:
@Ivanluv, then I am happy to approve, but would like @shadrackkibet to look quickly at what you have done first.  If it is just to get the blank dialogue, then I wonder if it should be limited to this one, or you might like to do the same for the 3 variable dialogue at the same time.  Up to you?  I am hoping that working on the 3-variable dialogue could perhaps be something for an intern, under your instruction, once the 2-variable has progressed.

shadrackkibet:
@Ivanluv since you already have changes in another branch you could just open a PR since those changes do not affect frmmain. I will suggest you liaise with @rdstern on the testing before we enable it in the frmmain. We really don't want to encourage enabling of dialogs from the main form before they are tested and completed.

Ivanluv:
okay am therefore closing this PR

PR #6704, opened 2021-08-25
Added Pivot Table dialog to the Describe> Two Variable menu

### LINKED ISSUE #6634: Adding pivot tables
I came across the `rpivotTable` package which introduces pivot tables as a little widget in R. The user can get the results of a single summary statistic and move variables around the table.
Function-wise it looks relatively straightforward, able to run with just a `data` variable given. 
A lot of the other parameters are all options that can be specified when running the R-code, or can be decided once the table is created
* `rows` - variables in the rows (however, can be changed after the table has run)
* `cols` - variables on the columns (however, can be changed after the table has run)
* `rendererName` -  whether a table, heattable, etc is shown (however, can be changed after the table has run)
* `inclusions` - levels of a variable to be included in the output
* `exclusions`, an alternative to `inclusions` - levels of a variable to not be included in the output (e.g. in `diamonds` data to not show the level `A` in the `color` variable, run the parameter: `exclusions=list(color="A")`)

For example, 

```
library(rpivotTable)
library(ggplot2)

rpivotTable(diamonds,
            cols = c("color"),
            rows = c("clarity", "cut")
            exclusions=list(color="E"))
```

image

But the columns can be shifted around easily

image

It looks like a flexible but simple function. I wasn't sure what thoughts were on perhaps including this as an option in R-Instat?

Ivanluv:
this fixes  #6634 and specifically for the Describe >Two Variable case.

### DIFF
--- a/instat/dlgPivotTable.vb
+++ b/instat/dlgPivotTable.vb
@@ -35,6 +35,7 @@ Public Class dlgPivotTable
     End Sub
 
     Private Sub InitialiseDialog()
+        Dim strOrder As String() = {"Given", "Reverse", "Ascending", "Descending"}
         ucrBase.clsRsyntax.iCallType = 2
 
         ucrSelectorPivot.SetParameter(New RParameter("data", iNewPosition:=0))
@@ -54,6 +55,10 @@ Public Class dlgPivotTable
         ucrChkIncludeSubTotals.SetParameter(New RParameter("subtotals", iNewPosition:=3))
         ucrChkIncludeSubTotals.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
         ucrChkIncludeSubTotals.SetRDefault("FALSE")
+
+        ucrInputAdditionalRowOrder.SetItems(strOrder)
+        ucrInputInitialColumnOrder.SetItems(strOrder)
+        ucrInputInitialRowOrder.SetItems(strOrder)
     End Sub
 
 
@@ -63,7 +68,7 @@ Public Class dlgPivotTable
         ucrReceiverInitialRowFactor.SetMeAsReceiver()
 
         clsRPivotTable.SetRCommand("rpivotTable")
-
+        ucrBase.clsRsyntax.SetBaseRFunction(clsRPivotTable)
     End Sub
 
     Private Sub SetRCodeForControls(bReset As Boolean)
@@ -73,7 +78,11 @@ Public Class dlgPivotTable
     End Sub
 
     Private Sub TestOkEnabled()
-
+        If ucrInputInitialRowOrder.IsEmpty OrElse ucrInputInitialColumnOrder.IsEmpty Then
+            ucrBase.OKEnabled(False)
+        Else
+            ucrBase.OKEnabled(True)
+        End If
     End Sub
 
     Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
--- a/instat/dlgPivotTable.vb
+++ b/instat/dlgPivotTable.vb
@@ -164,7 +164,6 @@ Public Class dlgPivotTable
 
     Private Sub ReceiversChanged(ucrChangedControls As ucrCore) Handles ucrReceiverInitialColumnFactor.ControlValueChanged,
             ucrReceiverInitialRowFactor.ControlValueChanged, ucrReceiverSelectedVariable.ControlValueChanged
-
         If Not bRcodeSet OrElse Not ucrChkSelectedVariable.Checked Then
             Exit Sub
         End If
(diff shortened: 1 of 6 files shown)


shadrackkibet:
@HawardKetoyoMsatsi could you review the following.

rdstern (approved):
@Ivanluv this is just great.  @shadrackkibet can you check too and then we can merge.

lloyddewit:
@HawardKetoyoMsatsi Reminder: please could you review? thanks

HawardKetoyoMsatsi (changes requested):
Do we have to do frmMain changes on the same PR? I don't see any problem with this, but I think according to what is required from R-Instat developers is the frmMain changes are done on a separate PR then merged quickly. 


I get this error when I try to either 
1.  when I try to enter a value in the receiver `Initial Row factor:` and the multiple receiver while the receiver` Initial Column factor:` is still empty.
3. delete an entry I have made in the receiver Initial Column factor while there is a value in the receivers `Initial Row factor:` and the multiple receiver 

image

Ivanluv:
@HawardKetoyoMsatsi could you check this again

HawardKetoyoMsatsi:
Yes, I see everything is fixed.
@shadrackkibet @lloyddewit  over to you.

instat/dlgPivotTable.vb
```diff
@@ -1,3 +1,205 @@
...
+                Else
+                    If Not ucrReceiverInitialColumnFactor.IsEmpty Then
+                        clsConcatenateFunction.AddParameter("column", ucrReceiverInitialColumnFactor.GetVariableNames(bWithQuotes:=False), bIncludeArgumentName:=False, iPosition:=0)
+                    End If
+                    If Not ucrReceiverInitialRowFactor.IsEmpty Then
+                        clsConcatenateFunction.AddParameter("row", ucrReceiverInitialRowFactor.GetVariableNames(bWithQuotes:=False), bIncludeArgumentName:=False, iPosition:=1)
+                    End If
+                End If
+                For Each strItem In ucrReceiverSelectedVariable.GetVariableNamesList(bWithQuotes:=False)
+                    If strItem <> ucrReceiverInitialRowFactor.GetVariableNames(bWithQuotes:=False) AndAlso strItem <> ucrReceiverInitialColumnFactor.GetVariableNames(bWithQuotes:=False) Then
```
lloyddewit, line 188: In the lines above, before we call `GetVariableNamesList`, we check that the receiver is not empty. Should we do similar checks before executing these lines?

maxwellfundi:
@lloyddewit  Please review this. I would like to reduce the ready PRs.

instat/dlgPivotTable.vb
```diff
@@ -1,3 +1,205 @@
...
+            clsRPivotTable.AddParameter("data", clsRFunctionParameter:=ucrSelectorPivot.ucrAvailableDataFrames.clsCurrDataFrame, iPosition:=0)
+        End If
+    End Sub
+
+    Private Sub ReceiversChanged(ucrChangedControls As ucrCore) Handles ucrReceiverInitialColumnFactor.ControlValueChanged,
+            ucrReceiverInitialRowFactor.ControlValueChanged, ucrReceiverSelectedVariable.ControlValueChanged
+        If bRcodeSet Then
+            If ucrChkSelectedVariable.Checked Then
+                clsConcatenateFunction.ClearParameters()
+                Dim iCount As Integer = 2
```
lloyddewit, line 170: This variable is only used in the `For` loop. Move this declaration to just before the `For` loop?

lloyddewit (changes requested):

instat/dlgPivotTable.vb
Ivanluv: I don't think it will make much difference because The `for `loop wont execute when the list of columns in the receiver ie`ucrReceiverSelectedVariable.GetVariableNamesList(bWithQuotes:=False)` will be empty

instat/dlgPivotTable.vb
Ivanluv: After adding the 'happy path keeps left' principle ' this will be okay

Ivanluv:
@lloyddewit  could you look at this again

instat/dlgPivotTable.vb
lloyddewit: `ucrReceiverInitialColumnFactor.GetVariableNames` is also called on line 188, and this is not part of the loop condition.
Could this potentially trigger an exception?

lloyddewit: @Ivanluv thanks for explaining on Skype, this comment is now resolved

lloyddewit (approved):

lloyddewit:
@rdstern There are some small changes since your last review, these should not have affected the functionality. Please could you test/approve again? thanks

rdstern (approved):
@lloyddewit and @lloyddewit great.

PR #6706, opened 2021-08-25
Improved usability of the General Graphics dialog

### LINKED ISSUE #6633: Small change and some questions on the general Graphics dialogue
The initial change is trivial.  At the start there is only one real control and that is the Plot Options.  Please could that control be moved, so it is in the same place as in the other dialogues.

@shadrackkibet the reason I am suggesting you for this is that I wonder what else we might add to this main dialogue that is perhaps different to the other plotting dialogues, and that might make it a more attractive dialogue to use.  Here are my initial suggestions.  Unless they are easy, and obviously sensible to you, this is for reactions and discussion initially.
a) A button with label `Main variable`s  Perhaps that's the equivalent to (say) Boxplot Options on the ordinary graphics dialogues.  If used, then it opens a dialogue with the usual(?) controls.  So Variable Selector on left.  Possibly even Single/Multiple, for Y, X Variable, Factor OPtional, and then Also Facet Variable - possibly even 2 facet variables.  This would also give access (automatically) to the Data Options:

image

Bit like this.  Not sure there what Group is there for?  So, roughly this plus Facet or Facet 1, Facet 2.  The factor (Optional)might be labelled as Variable (Fill) and perhaps the Groups could be Variable (Colour).

We might not allow the Single/Multiple as this could complicate the code - at least initially.
This adds the first part of the command, so something like:
`last_graph <- ggplot2::ggplot(data=survey, mapping=ggplot2::aes(y=yield, x=village, colour=fertgrp, fill=fertgrp)) + `

plus the facet part of the command!  No geom, because we don't have any yet.

Then  on the right-hand side we add some options that could apply to the dialogue as a whole.
a) Theme
b) Legend
c) Coord-flip
d) Title, Subtitle, Caption
e)? what else?

We still have to build up the layers, but it becomes quite nice (and simple) to use?

N-thony:
Fixes #6633
@rdstern please have a look. Thanks.

### DIFF
--- a/instat/dlgGeneralForGraphics.Designer.vb
+++ b/instat/dlgGeneralForGraphics.Designer.vb
@@ -40,9 +40,15 @@ Partial Class dlgGeneralForGraphics
     Private Sub InitializeComponent()
         Dim resources As System.ComponentModel.ComponentResourceManager = New System.ComponentModel.ComponentResourceManager(GetType(dlgGeneralForGraphics))
         Me.cmdOptions = New System.Windows.Forms.Button()
+        Me.lblXVariable = New System.Windows.Forms.Label()
+        Me.lblFillOrColor = New System.Windows.Forms.Label()
+        Me.ucrVariablesAsFactorForGraphics = New instat.ucrVariablesAsFactor()
+        Me.ucrFillOrColourReceiver = New instat.ucrReceiverSingle()
+        Me.ucrReceiverX = New instat.ucrReceiverSingle()
+        Me.ucrGraphicsSelector = New instat.ucrSelectorByDataFrameAddRemove()
         Me.ucrAdditionalLayers = New instat.ucrAdditionalLayers()
-        Me.ucrSaveGraph = New instat.ucrSaveGraph()
         Me.ucrBase = New instat.ucrButtons()
+        Me.ucrSave = New instat.ucrSave()
         Me.SuspendLayout()
         '
         'cmdOptions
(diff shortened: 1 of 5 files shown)


rdstern:
@N-thony that's a very good start:
a) The Layers starts with an initial line geom.  And then I couldn't get rid of it, because it was produced on the main dialogue.  It should start blank.
b) Could you move the Layers control down a little.  Then there could be room for the Plot Options in its usual place.  And the button for Facets and the one for Themes could then go in the middle of the dialogue.

N-thony:
> @N-thony this is looking really good. You could make use of this dialogue in your next report - I'll explain.
> Meanwhile, before merging there is a problem with the x-axis and y-axis in the sub-dialogues. I am not sure if something needs to be done on the main dialogue, or on the sub-dialogue?
> I think this may also need some help, because I suspect that the setting of the first geom might be used here. Compare a simple point plot with the survey data and the same point plot from the general dialogue. I assume it detects that this is continuous from that first geom. It needs to do this however that first geom was produced.

@rdstern I made a small change regarding the x-axis and y-axis problems. Please have a look.

maxwellfundi:
@rdstern @lloyddewit please re-review this.

rdstern:
@N-thony it still has the same problem.  I think this may be a general problem, rather than just this dialogue, and am copying @dannyparsons as we may need his help.
Here is the y-axis tab:  I am trying to edit a simple graph which has geom_point.
image
As you see, there is nothing we can do to edit the y-variable settings.  I think this is related to the fact that the default entry point is these variables for discrete data, but here even that menu is not visible, so we can't change to be able to edit the current settings.

N-thony:
> @N-thony it still has the same problem. I think this may be a general problem, rather than just this dialogue, and am copying @dannyparsons as we may need his help.
> Here is the y-axis tab: I am trying to edit a simple graph which has geom_point.
> image
> As you see, there is nothing we can do to edit the y-variable settings. I think this is related to the fact that the default entry point is these variables for discrete data, but here even that menu is not visible, so we can't change to be able to edit the current settings.
> I think this has been a casualty of the translation changes. @N-thony is looking into that.

@rdstern this is fixed now. Have a look. Thanks.

rdstern (dismissed):
@lloyddewit there will be more changes in the future, but this now seems to be working fine and it would be good to get it merged.
@N-thony seems also to have fixed a problem that is more general - with the x and y  sub-plot dialogues.  That is needed here and elsewhere.

N-thony:
> @lloyddewit there will be more changes in the future, but this now seems to be working fine and it would be good to get it merged.
> @N-thony seems also to have fixed a problem that is more general - with the x and y sub-plot dialogues. That is needed here and elsewhere.

@lloyddewit over to you.

lloyddewit (changes requested):

lloyddewit (approved):

lloyddewit:
@rdstern The changes since your last review should not have affected the functionality. If you test/approve then we can merge, thanks!

rdstern (approved):

PR #6708, opened 2021-08-30
Added smooth as a separate option in line plot dialog

### LINKED ISSUE #6651: Add step, path and smooth to the Describe > Specific > Line dialogue

shadrackkibet:
fixes #6651

@rdstern this PR add smoothing as a top radio button in the line plot dialog. Please test.


rdstern (changes requested):
@shadrackkibet this seems an excellent start.  I initially thought that you could do some quick improvements and then hand it over, but I think an initial acceptable version may be more acceptable than I considered previously.  So you may be able to "nail" a good initial version.
At the same time a small change in the other change, where you added step, etc, could be useful.
Then others will continue with the other buttons for the top of this dialogue.

- [x] a) Move the line option up to the same level as points,  So Add line and Add points checkbox are in the same "row", with the default perhaps Add line is checked (so the points are connected, and Add Points is unchecked.

- [x] b) When in this option the geom changes from Line Options to Smooth Options.

- [x] c) At the same time, (now it is merged) change the option with Line, so that when step is selected, then change to Step options.  When Path is selected it changes to Path options.

- [x] d) Omit the Orientation option.  I suggest that is distracting and not important enough for the main dialogue.

e) Add arguments for the different options of the smoothing.  Initially possibly as follows:

 

- - [ ] 1) function  argument is always there
- 
-  - [ ] 2) family is there for glm and gam.  (check with the family function for the options.
- 
- - [ ] 3) span is there for loess
- 
- - [ ] 4) se is there for all options and default is unchecked.
- 
- - [ ] 5) In the pull down, use lower case, e.g. gam not GAM.

This is all then incredibly powerful and fits well with the future improvements in modelling, and also our data science stull!

- [ ] Can you confirm that it will (later) be easy to add tooltips for each method.  So when loess is selected there could be a tooltip for that?

shadrackkibet:
@anastasia-mbithe could you pull this branch? I would like you to proceed with this task with my help.

shadrackkibet:
@rdstern could you specify the tooltips here for @anastasia-mbithe to add? Perhaps it can be a nice continuation after the sprint since it is something easy?

PR #6723, opened 2021-09-06
Adding the compare column dialogue to column  to the Climatic > Tidy and Examine menu

### LINKED ISSUE #6470: Two changes to the climatic menu on frm main
Could there be a new menu item in Climatic > File > Export to Climsoft.
It becomes the top of the 3 export dialogues and is ready for the dialogue to be written.  Perhaps the blank dialogue - for now - should also be included.
Then in Climatic > Tidy and Examine > Compare Columns.

This is simply another copy of the Prepare > Check Data > Compare Columns dialogue.  In the tidy and examine it is the 3rd menu item, so just below Daily Data Entry

mmulahvictor:
Fixes #6470

### DIFF
--- a/instat/dlgAddKey.resx
+++ b/instat/dlgAddKey.resx
@@ -123,10 +123,13 @@
   </data>
   <assembly alias="System.Drawing" name="System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
   <data name="cmdCheckUnique.Location" type="System.Drawing.Point, System.Drawing">
-    <value>10, 230</value>
+    <value>13, 283</value>
+  </data>
+  <data name="cmdCheckUnique.Margin" type="System.Windows.Forms.Padding, System.Windows.Forms">
+    <value>4, 4, 4, 4</value>
   </data>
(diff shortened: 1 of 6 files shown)


lloyddewit:
@mmulahvictor there isn't anything to review yet, I think you still need to push your changes.
I assume that you're part of the Kenya team and have local support, if not then let me know, thanks

mmulahvictor:
Thanks let me do the changes immediately

lloyddewit:
@mmulahvictor Thank you for the commits! I had a quick look and noticed some things:
- the code does not build
- the project is missing at least one file
- 'dlgImportt' does not follow the normal naming conventions. Should we rename?
- I assume that the purpose of this PR is to insert a placeholder (stub) into 'frmMain' for a new dialog(s). This approach is sensible because it reduces the risk of merge conflicts. However, if we do this, we should ensure that the stub is invisible to the user. Normally we do this by making the new menu option(s) invisible.

@Ivanluv Please could you help @mmulahvictor prepare this PR, and then do the first peer review? thanks

mmulahvictor:
Ok I am trying to create a new one with the help of Ivan. of which I will
make the changes.

PR #6724, opened 2021-09-06
Improved smooth options in Line plot dialog.

anastasia-mbithe:
@rdstern This is a continuation of the task started by @shadrackkibet in PR #6708. Could you kindly have a look at it.


rdstern (changes requested):
We discussed and there remain various options, particularly formula and family to add.  Then it is an exciting improvement to our system.

anastasia-mbithe:
@rdstern Kindly have a look at the changes made and test them.

rdstern:
Well done.  This is looking good already and I think all the options are all there.  Please now discuss the layout in detail with @shadrackkibet who is better than me on that aspect.  I describe below.  On substance, each of the controls Span, Formula, etc should have a checkbox, as they are all optional.  This would also apply to the R-code you generate.  Unless you click the checkbox you don't need that option in the code.  Then I think Shadrack said the default for span would be 0.75.

Then it is just the look of the dialogue.  I would prefer that the set of options for Smoothing is on a maximum of 2 "lines".  Include thinking about how close to the lhs you start the controls, and the Method label doesn't line yet with the control.  
Also put the se on the right i.e. after the formula and family and Span.  You might decide ona group box - perhpas called Smooth to help making it all neat!

anastasia-mbithe:
I have added a group box for the Smooth options and also check box for the Span and Formula Options. @rdstern you can check if its okay now.

rdstern (approved):
Great.  Very happy to approve.

shadrackkibet (approved):

Issue #6670, opened 2021-08-13
One Way frequency table: Functionalities allowing to decide if you want to display NA and summary in the table or not


IssoufKon:
**Is your feature request related to a problem? Please describe.**
The problem is that it is not possible to remove the "NA" and "summary" in the one way frequency table.  Why is it possible to do this with the two or three way frequency tables but not with the one way frequency table? In fact, these functionalities don't exist in the "more options" menu so we can't decide if we want them or not in the One Way frequency table.

**Describe the solution you'd like**
In fact, in "One Way frequencies", when we click on the "options" menu, in "more options" of the table, we think that we should add the following functionality:
•	Show summary
•	Show missing values
•	Decimal places for percentage 

**Describe alternatives you've considered**
As in the Two Way frequencies table, these different functionalities will allow us to decide if we want to display the missing value or not in the table, the summary or not, and also to decide the number of digits after the decimal point that we want.

**Additional context**
Capture of the area where we could have more functionalities,
![1](https://user-images.githubusercontent.com/88491307/129292254-dc4c474b-28b8-4670-8c8a-ef22ee7ad693.GIF)

Example with the case of frequency 2 table
![image](https://user-images.githubusercontent.com/88491307/129292348-2aa35b8f-4a4b-4c1c-993c-f1e50bcd86a0.png)

Capture of the table with the NA and Summary that we want to delete.
![image](https://user-images.githubusercontent.com/88491307/129292463-459b5a89-0c94-4437-b749-b83be5a75541.png)

Gioche6:
@Ivanluv , I am taking over this task.

Ivanluv:
@Gioche6 okay

Patowhiz:
@rdstern, @Gioche6 and I are looking into different ways we could add the suggested options. As you have mentioned in item (d) above, we will leave the other options if we find it's taking longer to get an alternative R package that does the same.

Patowhiz:
@rdstern we have decided to just solve item (a), (b) and (c) for this sprint. In future more research could be done on alternative packages that accommodates all the suggested options or we could write our own custom function that does the same.

Gioche6:
@rdstern , @Patowhiz .On solving item _(c)_, I have noted that the `frq` function does not return a data frame, but the result is of different class, namely `"sjmisc_frq", "list"` . However,  Once we convert the result into a data frame, some summary statistics of the result  are lost, namely: Total, valid , mean and sd. 
see the snip below. 
[...]

Issue #6729, opened 2021-09-07
Difference in Script Window codes and Output Window Codes for One Way Frequency Dialog


Gioche6:
**I have noted a difference in the codes produced by the One way frequency dialog, in Both option. the codes produced at the Window Script are different to those produced at the output Window**

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to 'One Way Frequency Dialog'
2. Click on 'Both'
3. Scroll down to 'Ok'
4. Open the Dialog again. 'Click the script button'
5. Compare the codes produced at the output window and the script Window. 
6. See the error

See the snips below :

Here's the Dialog
![one way frequency ](https://user-images.githubusercontent.com/73229979/132315988-b9f55dbe-9d97-464d-8638-39b043713c0f.png)

Below is a code produced by the Script Window. 
![new issue](https://user-images.githubusercontent.com/73229979/132315726-9d002040-daa8-4ace-a430-f40e76433ba0.png)

And below is the script produced at the output Window 
![new](https://user-images.githubusercontent.com/73229979/132316429-fa0b5718-e820-45f2-a985-2c4bd4785595.png)

PR #6730, opened 2021-09-07
Improved One Way Frequencies dialog, added table save and extra parameters for 'frq' function

Gioche6:
Partly fixes Issue #6670 

@rdstern , @Patowhiz This is ready for review

- I have added the `min.frq` parameter in the `frq` function for the One Way Frequency Dialog. Additionally, I have added the `show.na` parameter in the same function. 

- A function `as.data.frame` has been added converting the result into a data frame

- A save control has been added, to save the resulting data frame for the table and Both options in the dialog.

### DIFF
--- a/instat/sdgOneWayFrequencies.vb
+++ b/instat/sdgOneWayFrequencies.vb
@@ -36,10 +36,13 @@ Public Class sdgOneWayFrequencies
 
         'Table Only
         ucrChkShowStrings.SetParameter(New RParameter("show.strings", 7), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:="TRUE", strNewValueIfUnchecked:="FALSE")
-        ucrChkShowStrings.SetText("Show Strings")
+        ucrChkShowStrings.SetText("Omit Character Variabless")
 
-        ucrChkShowNa.SetText("Show NA")
-        ucrChkShowNa.SetParameter(New RParameter("show.na", 11), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:="TRUE", strNewValueIfUnchecked:="FALSE")
+        ucrPnlShowMissingFreq.SetParameter(New RParameter("show.na", 8))
+        ucrPnlShowMissingFreq.AddRadioButton(rdoShowMissingTrue, "TRUE")
+        ucrPnlShowMissingFreq.AddRadioButton(rdoShowMissingFalse, "FALSE")
+        ucrPnlShowMissingFreq.AddRadioButton(rdoShowMissingAuto, Chr(34) & "auto" & Chr(34))
+        ucrPnlShowMissingFreq.SetRDefault("TRUE")
 
         'Graph Only
         ucrPnlGraphType.SetParameter(New RParameter("type", 4))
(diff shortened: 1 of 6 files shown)


lloyddewit:
@Gioche6 I liked the way you wrote the PR description, thank you

lloyddewit (approved):
@Gioche6 This looks good to me

lloyddewit:

Patowhiz:
@lloyddewit @rdstern please note that we found another bug described in issue #6729. Our intention is to solve it in this PR too because we can't finish the items stated above without fixing the bug.

rdstern (changes requested):
This is an excellent start.  There is quite a bit to check here.
a) I get a developer error when I return to the dialogue.  That seems to happen on some dialogues when the settings are not quite right.  
Then I have yet to try saving a table.  That's because it now saves every time, and instead of giving me the results in the output window.  The default is not to save, and I don't think we really need a default last table object saved each time.  So, whn the Save Option isn't checked it should give either html, or simple results in a browser or the output window as before.
Try with the "usual" survey data.
I think the minimum frequencies is working ok.  I'll be able to check more once it displays - as it did before - in the output window.
When it does save, please chack it only saves the results.  There is an extra line on occasions, and whn you set the min.frq then there are 2 extra lines.

Gioche6:
@rdstern , I have made the changes Requested  on the One way frequency Dialog. It is now ready for review.

rdstern (changes requested):
@Gioche6 this is now working well.
I have realised, however, that another option would be useful
Could you add a label for a group box Show Missing Frequencies There are 3 radio buttons, namely True, False and Auto.  The default is True.  If the others are chosen, then it generates the option show.na = FALSE or show.na = "auto".  -

Gioche6:
@rdstern ,  `show.string` is  Logical parameter, if `TRUE`, frequency tables for character vectors will not be printed. This is useful when printing frequency tables of all variables from a data frame, and due to computational reasons character vectors should not be printed

see the example below for a one way frequency summary of a string Variable.

On this first image, ihave set show.string = TRUE,  and the results are as shown.
string1
 
On the second image, I have set show.string =FALSE, for the same. The results are as follows. 

string2

rdstern:
Aha, thank you for the explanation.  That could be reworded more clearly for us as "Omit Character Variables".

It isn't particularly important, for us, because would usually simply not select the character variables if we don't want them.  But it does no harm, so perhaps reword the label and then keep it as it is.

Then I am also happy with the NA stuff within the options, but we need to allow for all 3 options, possibly by the radio buttons.

Gioche6:
@rdstern , I have added the group box nd made necessary change on the design . This is now ready for review and testing.

rdstern (approved):

Gioche6:
@lloyddewit , could you please review this PR, I think it all better now and can be merged.

lloyddewit (approved):

instat/sdgOneWayFrequencies.vb
```diff
@@ -36,7 +36,13 @@ Public Class sdgOneWayFrequencies
 
         'Table Only
         ucrChkShowStrings.SetParameter(New RParameter("show.strings", 7), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:="TRUE", strNewValueIfUnchecked:="FALSE")
-        ucrChkShowStrings.SetText("Show Strings")
+        ucrChkShowStrings.SetText("Omit Character Variabless")
```
shadrackkibet, line 39: Typo? `Variabless`

Gioche6: @shadrackkibet , let me quickly create a new PR , correcting the typo. Thank you for pointing it out

PR #6731, opened 2021-09-07
Reordering f the summaries in the summary table

### LINKED ISSUE #6644: Reorder summary statistics in the Summary Table dialog
In #6508 @rdstern requested a way to rearrange the summary statistics in the summary table.

This is similar to reordering a factor level. Here, the "factor" is a string containing the summaries to run. The "factor levels" are a list of the summaries.

A way this can be done is by:

1. Open the sub-dialog
2. Select the summaries
3. On the main dialog, there is then a list of the summaries selected
4. The user can then rearrange the order of these summaries from the list (similar to the `Prepare > Factor > Reorder Levels` dialog)
5. The r-code just reads from this list, e.g., `summaries = c("first-item", "second-item", ..., "nth-item")`

So can a list of the summaries selected be added to the main dialog for the Summary Table dialog? This list has up/down buttons to rearrange the variables (I assume this is achieved with "ucrReorder" like in the `Prepare > Factor > Reorder Levels` dialog).
This list displays the summary statistics selected on the sub-dialog
For the `summary` parameter, the parameter value is this list.

Ivanluv:
Fixes #6644

### DIFF
--- a/instat/dlgSummaryTables.vb
+++ b/instat/dlgSummaryTables.vb
@@ -129,6 +129,8 @@ Public Class dlgSummaryTables
 
         ucrChkDisplayMargins.AddToLinkedControls({ucrInputMarginName}, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:="All")
         ucrChkDisplayMargins.AddToLinkedControls({ucrPnlMargin}, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=rdoOuter)
+
+        ucrReorderSummary.bDataIsSummaries = True
     End Sub
 
     Private Sub SetDefaults()
@@ -221,6 +223,7 @@ Public Class dlgSummaryTables
         ucrChkDisplayVariablesAsRows.SetRCode(clsMutableOperator, bReset)
         ucrChkStoreResults.SetRCode(clsDefaultFunction, bReset)
         ucrSaveTable.SetRCode(clsMutableOperator, bReset)
+        FillListView()
     End Sub
 
     Private Sub TestOKEnabled()
@@ -345,6 +348,7 @@ Public Class dlgSummaryTables
         Next
 
         clsSummariesList.ClearParameters()
+        'Changing the parameter positions
         For Each clsParameter In lstOrderedSummaries
             clsSummariesList.AddParameter(clsParameter.strArgumentName, clsParameter.strArgumentValue, iPosition:=iparameterIposition)
             iparameterIposition = iparameterIposition + 1
--- a/instat/dlgSummaryTables.vb
+++ b/instat/dlgSummaryTables.vb
@@ -25,6 +25,7 @@ Public Class dlgSummaryTables
             clsVariableHeaderLeftTopFunction, clsVariableHeaderTopLeftFunction,
             clsummaryVariableHeaderLeftTopFunction, clsSummaryVariableHeaderTopLeftFunction As New RFunction
     Private clsMutableOperator, clsColumnOperator As New ROperator
+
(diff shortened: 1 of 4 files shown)


rdstern (changes requested):
@Ivanluv this is looking very promising - except it doesn't yet seem to work!
I tried an example with the usual survey data and 5 summaries, namely count, max, mean, median, sd.  That's the order whetever I do in the new control.

Once it is working I also suggest we might have a Layout button that takes tyou to a sub-dialogue.  That's where we collect all the controls that disctate a different-from-default layout.  Once we get to it, then (with these summaries) we might also want to add a label option, so we can specify the label for each summary.  The default is as now, but we can change, e.g. Count non-missing to whatever we want?

Ivanluv:
@rdstern  this is currently working but the order of  the summaries appear in the inverse order as opposed to the order they occur in the code .I think @lilyclements  can help me with this

instat/dlgSummaryTables.vb
```diff
@@ -350,7 +350,8 @@ Public Class dlgSummaryTables
         clsSummariesList.ClearParameters()
         'Changing the parameter positions
         For Each clsParameter In lstOrderedSummaries
-            clsSummariesList.AddParameter(clsParameter.strArgumentName, clsParameter.strArgumentValue, iPosition:=iparameterIposition)
+            clsParameter.Position = iparameterIposition
```
lilyclements, line 353: Can you change this name from ‘iparameterIposition’ to ‘iParameterPosition’
Just for naming consistency.

lilyclements:
@Ivanluv when you run the dialog, does it put the summaries in the user specified order in the R code or is it reversed in the R code?

Ivanluv:
@lilyclements  it has the summaries in the user specified order in the Rcode. The output however, has the order reversed

lilyclements:
@Ivanluv I am little confused. The R function does seem to reverse it so I will fix that.
However, in addition to that, on the first time you run the dialog (or after clicking reset, inputting summary variables, and clicking OK), the R code in the output window reverses it

94D82770-4CCC-4E61-B004-380F48BCBF7C
126DFB4E-4A3A-417A-A380-E2E67B94D0C2

Ivanluv:
@lilyclements  could look at this again

instat/dlgSummaryTables.vb
```diff
@@ -328,12 +328,15 @@ Public Class dlgSummaryTables
     End Sub
 
     Private Sub FillListView()
+        Dim iposition As Integer = 0
```
lilyclements, line 331: Can you change this to ‘iPosition’ to be more consistent with elsewhere in the software

Ivanluv: This is done

lilyclements: Great. I will look at the bug tomorrow latest

lilyclements:
The order of the output changing occurs in the `instat_calculation$new` function when `manipulations` is a parameter.

This is in line 226 in the `calculate_summary` function in the "summary_functions.R" file

`combined_calc_sum <- instat_calculation$new(type="combination", sub_calculations = sub_calculations, manipulations = manipulations)`

@dannyparsons is this a known issue? I am not sure where to find the code from here that is causing the bug, and not sure if I will be able to fix it easily (or if I did fix it, if it would have implications elsewhere!). I can do a temporary fix in the of reordering the summary columns in the `calculate_summary` function (or the `summary_tables`) code for now if that is easier.

dannyparsons:
It might be something quite simple to do with the order the summary system is doing the calculations.

Would you be able to give me an example script I could run which has the issue, or maybe the two cases that are coming from the dialog? That might be easiest then I can try to debug in R.

lilyclements:
[calculate_summary_bug.zip](https://github.com/africanmathsinitiative/R-Instat/files/7140078/calculate_summary_bug.zip)

Here is an example running it with the `calculate_summary` function - I've run `return_output = TRUE` just to show that the order in the output is in the reverse to specified in the code.
As a side note - in R-Instat the output in the data frame window has the summary columns in the specified order. I'm intrigued if you happen to know why this is.

PR #6732, opened 2021-09-07
Added rug plot into point plot in Scatter Plot dialog

shadrackkibet:
@rdstern this is ready for review. It is the addition of rug plot into point plot dialog.


rdstern (changes requested):
@shadrack.  Nice so far.  The other options don't yet work - it is a rigplot on the left and bottom whichever you choose.  I also wonder about size.  Is colour perhaps more useful?

shadrackkibet:
@rdstern the bug is fixed now.

For now, if you want to change the colour you have to use the sub dialog. I claim that size is much more important than colouring for the reason that if you are printing your graph in black and white which is "often the case" then it is better to use size rather than colouring.

rdstern (approved):
@shadrackkibet this now looks fine.  I did try the change inside to outside.  The command still looks fine, but there no longer seems to be the rug-plot.
However, the options in the dialogue work well.

lloyddewit (approved):

PR #6733, opened 2021-09-07
Enabled New Data Frame dialog to 'try' running multiple lines

### LINKED ISSUE #6692: Improve the File > New Data Frame to be able to run multiple lines of code
Could we please enhance the File > New Data Frame dialogue, with the Command option to be able to process multiple lines, as is done in the script window.  

I give 3 examples of code that (sort of) works in the script.  But it is then separate to R-Instat and needs a command to make it into a data frame.  

This works fine already as is shown in the figure:

```
library(tidytext)
get_sentiments("bing")
```
image

The second example is as follows - taken from the Text Mining book, chapter 1

```
text <- c("Because I could not stop for Death -",
          "He kindly stopped for me -",
          "The Carriage held but just Ourselves -",
          "and Immortality")
library(dplyr)
text_df <- tibble(line = 1:4, text = text)

```
Here it is in the dialogue in R-Instat:
image

(I got rid of the library line, because I think we load dplyr anyway.  Then I don't assign it in the last line, because the dialogue does that.#
This doesn't work.  But if I get rid of the multiple lines, so it is just one long line, then it works fine.

The third example is what I am aiming for.  Interesting R code that produces a data frame.  And an example that is lengthy to do in R-Instat.  (And starting from there, R-Instat is fine!):

```
library(janeaustenr)
library(dplyr)
library(stringr)

original_books <- austen_books() %>%
  group_by(book) %>%
  mutate(linenumber = row_number(),
         chapter = cumsum(str_detect(text, 
                                     regex("^chapter [\\divxlc]",
                                           ignore_case = TRUE)))) %>%
  ungroup()

```
Again, if I take the text starting original_books and put all on one line, and move the original_books to the name of the data frame, then it works fine.

So could we arrange for multiple lines to run here - as they do in the script window.

That will be great!

Patowhiz:
Fixes #6692

@lloyddewit changes to the R link could be used to refactor code used in the script window. The functions added could also potentially be used in the new script dialog.
I also made the dialog to use the custom try control, this prompted me to enhance the custom try control.


lloyddewit:
@rdstern I suggest you first test, then I'll peer review, thanks

rdstern (changes requested):
@Patowhiz I tried different examples and the Jane Austen one in #6692 still doesn't work.  It does if the main part is put onto a single line.  That's as before.

rdstern (dismissed):
@Patowhiz great it now works.  I hope @lloyddewit finds it ok too.
I did find that Try said it wouldn't work!  That's for the complicated example that is mentioned in the issue.  I persisted and it worked fine

Patowhiz:
@rdstern yes, the _try_ is currently set up to be a bit more restrictive for commands that create a new data frame, partly because of the limitations of the R.Net engine. All you have to do for the 'complex' sample code is remove the extra assignment(we can relax this restriction and make it consistent with _OK_  behavior).

@lloyddewit this is ready for you. Please note that additional changes that may require your attention tend to reformat the script code sent to the R.Net engine. I did this to ensure consistency and reliability in runnable commands being executed by the engine. I have demonstrated the reformatting to @rdstern.

Below is an example of how a script is reformatted;
From
```
text <- c("Because I could not stop for Death -",
          "He kindly stopped for me -",
          "The Carriage held but just Ourselves -",
          "and Immortality")
library(dplyr)
tibble(line = 1:4, text = text)
```

To 

```
text <- c("Because I could not stop for Death -","He kindly stopped for me -","The Carriage held but just Ourselves -","and Immortality")
library(dplyr)
data <- tibble(line = 1:4, text = text)
```

instat/ucrTry.vb
```diff
@@ -112,35 +107,53 @@ Public Class ucrTry
...
+                            If vecOutput Is Nothing Then
+                                bValid = False
+                                Exit For
                             End If
+
+                            'todo. for data frames output, length is expected to be > 0(even though that is a bonus check)
+                            'todo SJL. not sure if other output types will have length > 0
+                            'todo. as at 07/08/2021. this block is only being used by dlgNewDataframe ONLY
+                            bValid = vecOutput.Length > 0
+                            'don't break the if, we probably want to loop through to the last command checking validity?
```
lloyddewit, line 142: In my view, if a command line fails then we don't need to continue. Therefore, I think it's OK to break the 'If' when a command line fails.

instat/clsRLink.vb
```diff
@@ -393,6 +393,80 @@ Public Class RLink
         Return strScriptCmd
     End Function
 
+    ''' <summary>
+    ''' extracts all the complete runnable R commands from the passed script R command string
+    ''' the command lines returned are re-formatted in a format that the R.Net engine can execute
```
lloyddewit:
```suggestion
    ''' Extracts all the complete runnable R commands from <paramref name="strScript"/>.
    ''' The command lines returned are re-formatted in a format that the R.Net engine can execute.
```

instat/clsRLink.vb
```diff
@@ -393,6 +393,80 @@ Public Class RLink
...
+                Continue For
+            End If
+            lstRunnableCommandLines.Add(strSplitScriptCmd)
+            strSplitScriptCmd = ""
+        Next
+        Return lstRunnableCommandLines.ToArray
+    End Function
+
+    ''' <summary>
+    ''' checks if the script passed is complete valid R script that can be run without an error
```
lloyddewit:
```suggestion
    ''' checks if <paramref name="strNewScript"/> is complete valid R script that can be run without an error
```

lloyddewit (dismissed):
@Patowhiz This all looks very good, I just pushed a couple of small changes to the XML headers, I hope that's OK

lloyddewit:
@rdstern There are no functional changes, please could you approve again? we can then merge, thanks

rdstern (approved):

lloyddewit (approved):

instat/ucrTry.vb
Patowhiz: @lloyddewit if so then we should only be checking for `Nothing` not length. Some commands return results of `Length = 0`

lloyddewit: @Patowhiz It's also a bit strange that we may set `bValid` to false, and then set it back to true with the next iteration of the loop.
Should we delete lines 138-142? Should we do this in a seprate PR?

PR #6734, opened 2021-09-07
Implemented the lollipop option in the Bar Chart dialog

### LINKED ISSUE #6726: Add lollipop option to the bar chart dialogue
We recently (I think) added geom_lollipop to the list.

This is a sort of bar chart.  It is (in geom terms) also a segment plus point.  But, as there is a geom lollipop I suggest the best place is in the bar chart dialogue.  It can be a single checkbox, on the left, perhaps after Polar.  If checked, then we use geom_lollipop rather than geom_bar, so the bar options button should change to Lollipop Options.
There should be 2 more controls on the same line as the Lollipop checkbox.  They are labelled Size, and Colour and use those controls from the Lollipop sub-dialogue.
This should be implemented for both the current options of the bar chart dialogue.

N-thony:
Fixes #6726 
@rdstern I have made the first implementation of the lollipop option, please have a look so that we can move further.

### DIFF
--- a/instat/dlgBarAndPieChart.Designer.vb
+++ b/instat/dlgBarAndPieChart.Designer.vb
@@ -52,6 +52,13 @@ Partial Class dlgBarAndPieChart
         Me.rdoDonut = New System.Windows.Forms.RadioButton()
         Me.lblLabelPosition = New System.Windows.Forms.Label()
         Me.lblLabelColour = New System.Windows.Forms.Label()
+        Me.lblLabelSize = New System.Windows.Forms.Label()
+        Me.lblLollipopSize = New System.Windows.Forms.Label()
+        Me.lblLollipopColour = New System.Windows.Forms.Label()
+        Me.ucrInputLollipopColour = New instat.ucrInputComboBox()
+        Me.ucrNudLollipopSize = New instat.ucrNud()
+        Me.ucrChkLollipop = New instat.ucrCheck()
+        Me.ucrInputLabelSize = New instat.ucrInputComboBox()
         Me.ucrInputLabelPosition = New instat.ucrInputComboBox()
         Me.ucrInputLabelColour = New instat.ucrInputComboBox()
         Me.ucrChkAddLabels = New instat.ucrCheck()
@@ -67,8 +74,6 @@ Partial Class dlgBarAndPieChart
         Me.ucrBarChartSelector = New instat.ucrSelectorByDataFrameAddRemove()
         Me.ucrBase = New instat.ucrButtons()
         Me.ucrPnlOptions = New instat.UcrPanel()
-        Me.ucrInputLabelSize = New instat.ucrInputComboBox()
-        Me.lblLabelSize = New System.Windows.Forms.Label()
         Me.SuspendLayout()
         '
         'lblByFactor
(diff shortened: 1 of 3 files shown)


rdstern (changes requested):
@N-thony this is almost working well.
Why just the 2 colours.
The only real change is that you have inclu=ded Horizontal=TRUE.  This should be deleted.  If we want horizontal then we'll use the coord-flip instead.

rdstern (changes requested):
@N-thony that's better.  WE are close to being able to merge.  
However, we will want to use these lollipops together with labels.  When I choose Add Labels, then it unticks the lollipop option.  The two should be separate.

lloyddewit (approved):

lloyddewit:

N-thony:
> @N-thony I keep checking with summary data from the usual survey. I wonder if you are checking too?
> a) Now it unsets the lollipop checkbox each time it executes.
> b) If you leave colour as blue, (default in your dialogue) then the point colour is not included - and it is black. So make that the default, so it is correct.
> I have not yet checked the interaction with labels - I'll do that on the next version

@rdstern sorry. There was a small change I made that has caused that small bugs. Now it seems resolved. I keep also checking. Thanks.

rdstern (changes requested):
@N-thony that's fixed.  But it still now doesn't allow me to add a text label to the lollipop chart 0 which I would often like to do.  When I click on Label it unchecks lollipop and vv.  I would like both together!

N-thony:
> @N-thony that's better. WE are close to being able to merge.
> However, we will want to use these lollipops together with labels. When I choose Add Labels, then it unticks the lollipop option. The two should be separate.

This was unclear about the functionalities of the two controls. That is why it is getting long. But hopefully, that is now fixed.

lloyddewit (approved):

rdstern (approved):
I am happy now!

N-thony:
> I am happy now!

@lloyddewit is it possible to merge this so that I can start working on the reorder feature? Thanks

rdstern:
@N-thony I am re-opening this, because I can't get a good lollipop chart.  This is work you did earlier.  And I suggest these are needed, because they are used to analyse the R-Instat evaluation data.  You may wish to work with one of the interns, because there is a general feature also to correct.  
Lollipop first.  I'll sort out some data and add it here.  There seem to be 2 problems - at least!
a) The code gives not just a lollipop, but also a bar chart.  The geom lollipop is instead of the bar chart.  
b) When you use lollipop the button changes to Lollipop Options.  But it still gives bar chart.  It should give Lollipop.

Also, cosmetic options on this dialogue.  It currently looks a bit overwhelming.  So:
c) Make the first Reorder label into a checkbox.  Default unchecked. The control is only visible when checked.
d) I think the 2nd Reorder, and also Position are only when the By Factor (Optional): is not empty.  So make them not visible until that is filled.

There is a related issue on the plotting geoms.  But that will be separate and should involve an intern.  A careful one!

lloyddewit:
> @N-thony I am re-opening this, because I can't get a good lollipop chart. This is work you did earlier. And I suggest these are needed, because they are used to analyse the R-Instat evaluation data. You may wish to work with one of the interns, because there is a general feature also to correct. Lollipop first. I'll sort out some data and add it here. There seem to be 2 problems - at least! a) The code gives not just a lollipop, but also a bar chart. The geom lollipop is instead of the bar chart. b) When you use lollipop the button changes to Lollipop Options. But it still gives bar chart. It should give Lollipop.
> 
> Also, cosmetic options on this dialogue. It currently looks a bit overwhelming. So: c) Make the first Reorder label into a checkbox. Default unchecked. The control is only visible when checked. d) I think the 2nd Reorder, and also Position are only when the By Factor (Optional): is not empty. So make them not visible until that is filled.
> 
> There is a related issue on the plotting geoms. But that will be separate and should involve an intern. A careful one!

@rdstern We can't reopen a merged PR, and there's a risk that your comments are lost.
So I reopened issue #6726 and moved your comments there.
I hope that's OK

Issue #6697, opened 2021-08-23
Error when opening OneVarCompareModels.vb


rachelkg:
Hello @Patowhiz and @lloyddewit, 

I receive this error when I try to open sdgOneVarCompareModels.vb

Thanks, 

Rachel
![sdgOnevarCompareModels](https://user-images.githubusercontent.com/85707390/130489349-ff06d7d1-f3df-490b-9e95-f9d84ebe7f6e.png)

Patowhiz:
@Gioche6 we can have a look at this together.

lloyddewit:
This is fixed in PR #6736

PR #6736, opened 2021-09-07
Ensure that One variable Compare Models subdialog opens correctly

Gioche6:
fixes issue #6697 

@Patowhiz , this is ready for review

The constructor was wrongly written. I have fixed it by changing it from `sdgOneVarCompareModels` to `sdgOneVarCompareModels_Load`

The sdgOneVarCompareModels Design now loads with no issue.

### DIFF
--- a/instat/sdgOneVarCompareModels.vb
+++ b/instat/sdgOneVarCompareModels.vb
@@ -21,7 +21,7 @@ Public Class sdgOneVarCompareModels
     Private clsChisqtableOperator, clsChisqbreaksOperator As New ROperator
     Private clsRSyntax As RSyntax
 
-    Private Sub sdgOneVarCompareModels(sender As Object, e As EventArgs) Handles MyBase.Load
+    Private Sub sdgOneVarCompareModels_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         autoTranslate(Me)
     End Sub
 


lloyddewit:
@Patowhiz or @shadrackkibet : In your opinion, does this fix work? I'm a bit confused, I thought that the name of the handler sub was not important, but maybe I'm wrong?

Patowhiz:
@lloyddewit yes this works. I looked at it with @Gioche6 before the change was made. 
This type of error is usually caused when the constructor is accidently overwritten the developer.
This [stack overflow](https://stackoverflow.com/questions/1956997/ctor-is-not-a-valid-identifier) answers explains more on the same.

lloyddewit (approved):

lloyddewit:
@Patowhiz  @Gioche6 Thank you for the explanation and the interesting link. I noticed that it had the same name as the constructor but the handler's signature was still unique so I assumed that wasn't the problem. It seems that the automatically generated code had a bug and renaming the handler solved this. That's great, I'm happy it's working now (and I learned something).

lloyddewit:

rdstern (approved):

Patowhiz:
@lloyddewit yes, that's one of the VB.Net shortcomings. I was also surprised that by the constructor generation.

PR #6737, opened 2021-09-07
Re-arranged specific graphs in the Describe menu

anastasia-mbithe:
@rdstern I have rearranged the Specific menu items as you suggested, with the help of @shadrackkibet. Kindly have a look at it.


rdstern (approved):
This is a good improvement and could be merged now.  Here are some -possibly future suggestions
The only very minor change I could suggest is to move the Mosaic up one further step, so it is the first under the line - just as bar charts are the first graph.
On more suggestions (perhaps for the future), I wonder about deleting pie charts from the bar chart menu.  Usually the names correspond to the radio buttons and we don't have one on pie charts.  I wonder about adding tooltips to the menu items.  Would that be useful or a distraction?
The line menu already has smooth, and I hear dumbbell and slope are coming now.  So perhaps that can already say Line/Smooth/Dumbbell/Slope
The rug item will change to be Map/Choropleth/Contour/Heatmap plot.  Perhaps add that now?
The Point plot is a bit lonely!  I don't have extra suggestions (yet!), but wonder whether we say Point (Scatter) Plot as the menu item?

lloyddewit (approved):

PR #6738, opened 2021-09-07
Added Heat Map Plot as a dialog in place of the Rug plot.

EstherNjeriLiberatta:
@africanmathsinitiative/developers @rdstern @shadrackkibet 

This is ready for review.


rdstern (changes requested):
@shadrackkibet I got build errors?

EstherNjeriLiberatta:
@shadrackkibet I have added the necessary implementations for the heat map. However, based on @Patowhiz advice, I think it's best I rename the dialog and the file names on another PR after this is merged. I noticed renaming the dialog requires me to make the changes on `frmMain` which could cause conflicts.

@lloyddewit I have changed the title of the dialog from **Rug Plot** to **Heat Map**, I noticed this change is not reflected during runtime, @Patowhiz has explained to me that it relates to the translation issue that you are working on. I think the title issue will be fixed automatically once you make the necessary entries in the translation database.

@shadrackkibet @HawardKetoyoMsatsi @rdstern this is ready for your review.

rdstern (changes requested):
This is an excellent start to this dialogue.  There are two elements to add, at least initially.
a) Add the lebels ad @shadrackkibet has implemented in the bar chart dialogue.  The inside/outside option isn't needed.
b) Add choices of colour scales.  Initially take the colour scales directly from the sub-dialogue.  In the longer term we should extend the range of scales in the sub-dialogue.
c) Optional at this stage is to investigate the Multiple option for this dialogue.  I think if there are multiple variables, then they are stacked and those are used for the tiles.  So, these are a set of variates - rather than factors.  Then the rows of the data become the y, the stacked variables make the x-s and the contents are used for the fill?  This will need a bit of tweaking of the R code, but should be ok.

EstherNjeriLiberatta:
@africanmathsinitiative/developers @rdstern @shadrackkibet 

@rdstern I have added the label buttons as you suggested, this is ready for review.

rdstern (approved):
Looks fine.

If you now just add the fill control from the plot sub-dialogue under the colour tab. this can be merged!

rdstern (changes requested):
Sorry, didn't mean to approve just yet.  It does need some flexibility on colours!

EstherNjeriLiberatta:
@rdstern @shadrackkibet 
I have made the requested changes on the dialogue.

rdstern:
Thanks - I'll look

rdstern (changes requested):
Great, well done.  The only detail is that the default should still be available.  So add a checkbox, default unchecked, for the colour palette option.

shadrackkibet:
@EstherNjeriLiberatta I would suggest that we work on getting this done, reviewed and merged today.

EstherNjeriLiberatta:
@rdstern @shadrackkibet @fran2or 
I have made the requested changes.

instat/dlgRugPlot.vb
```diff
@@ -17,7 +17,7 @@
 Imports instat.Translations
 Public Class dlgRugPlot
     Private clsRggplotFunction As New RFunction
-    Private clsRgeom_RugPlotFunction As New RFunction
+    Private clsRgeom_TileFunction As New RFunction
```
shadrackkibet, line 20: remove `_`

shadrackkibet (changes requested):
@rdstern please test and approve if you are happy to merge.

instat/dlgRugPlot.vb
EstherNjeriLiberatta: @shadrackkibet, it's done.

rdstern (approved):
WE will obviously have to change the menu name in the future.  But this is now working nicely!

shadrackkibet:
We should change the menu name now!

shadrackkibet (approved):

PR #6746, opened 2021-09-08
Unnest tokensfunction

### LINKED ISSUE #6687: Add dialogue or radio button for the unnest tokens function
This is all part of text mining.  The function is in the tidytext package, which is (I think) already in R-Instat.  The use is described in the [vignette](https://mran.microsoft.com/snapshot/2018-08-31/web/packages/tidytext/vignettes/tidytext.html).  In addition there is (yet another) interesting book that is available as a soft copy.  It covers the same materials and more in more detail.  It is called [Text Mining with R: A Tidy Approach.](https://www.tidytextmining.com/tfidf.html)

This function/dialogue takes texts and breaks it up into words.  Usually each row has a line of text, or perhaps a sentence or a paragraph.  Here is an example:
image

Then the function `unnest_tokens(jane,word, text)`  takes the variable called text, and produces a new data frame with a variable called word.

image

So this is like the current stack, except it is stacking the components of a single (quite complicated) variable.

So, my suggestion is to change the name of Stack to Stack (Gather).  Then to change the existing dialogue (at the same time) to use Gather.  There are 2 buttons.  One is Gather and that does what the reshape2::melt function does.  But we change it to use tidyR and gather.  Then we add another radio button called unnest and use that for the unnest_tokens function.

a) It has the usual data selector and a single receiver for the column with the texts.  This will (I assume be a character variable.
b) Then there is (at the bottom.the Stack Data Into is just Data into (and the default is word, but see below.
c) There is (like stack), new data frame name control.
d) Below the single receiver there is a label `Token`: with a pull-down - also into which you can type..  The default is words and the list is given in the command.  
e) Then label Format with pull-down and options given in the function.  Default is text.
f) Below that is a checkbox, default unchecked with label `to Lower Case`
g) Label Pattern/Option with a field into which you can type:  This gives pattern for regex or for use-defined function. There is a tooltip with Example "Chapter [\\d]" for regex token or " " for user defined token. 
h) Label Strip on left then checkbox default unchecked Punctuation another Non-Alphanumeric another Numeric,  another URL.  If checked then strip_punct = TRUE, strip_non_alphanum = TRUE, strip_numeric = TRUE, strip_url = TRUE

Vitalis95:
fixes #6687

### DIFF
--- a/instat/dlgCopyDataFrame.vb
+++ b/instat/dlgCopyDataFrame.vb
@@ -136,6 +136,7 @@ Public Class dlgCopyDataFrame
 
     Private Sub ucrPnlOptions_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlOptions.ControlValueChanged
         If Not ucrChkCopyToClipboard.Checked OrElse rdoDataFrame.Checked Then
+            clsCopyDataFrameFunction.AddParameter("data_name", strParameterValue:=Chr(34) & ucrDataFrameCopySheets.strCurrDataFrame & Chr(34), iPosition:=0)
             ucrBase.clsRsyntax.SetBaseRFunction(clsCopyDataFrameFunction)
         ElseIf rdoColumnsMetadata.Checked Then
             'todo. temporarily added here because this control's parameter pair condition is not updated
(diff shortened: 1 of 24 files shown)


lloyddewit:
Hi @Vitalis95 , thank you for the PR. Before you progress, you should resolve the conflicting files. I'm surprised that this happened, it may be that you created the PR in a non-standard way. Do you have someone in the Kenyan team you can ask for support?
Also, the PR title is important because it is pasted automatically into the release notes provided to the user. Please could you change the PR title so that it better describes to the user what has been done? Thanks!

Vitalis95:
@lloyddewit we are closing this.

PR #6750, opened 2021-09-09
Added reorder feature and minor bug fixes to the bar chart dialogue

### LINKED ISSUE #6582: Add a reorder feature to bar charts
Could you add a Reorder checkbox to the bar (and pie chart dialogue.  Default is unchecked.
If checked, then there could usefully be 4 options, namely Ascending, Descending, Reverse, None

In these options Ascending and Descending will be the common ones.  Ascending simply adds the reorder function to the command, DEscending makes it a minus!

image

To reverse the order of the bars look [here](https://stackoverflow.com/questions/42710056/reverse-stacked-bar-order) perhaps.

Note that this will also apply to the lollipop charts that we are doing by adding geom_segment to the point plots.

N-thony:
Fixes #6582
@rdstern please have a look. Thanks.

### DIFF
--- a/instat/dlgBarAndPieChart.vb
+++ b/instat/dlgBarAndPieChart.vb
@@ -124,7 +124,8 @@ Public Class dlgBarAndPieChart
 
         ucrReceiverX.Selector = ucrBarChartSelector
         ucrReceiverX.strSelectorHeading = "X Variable"
-        ucrReceiverX.SetParameterIsRFunction()
+        ucrReceiverX.bWithQuotes = False
+        ucrReceiverX.SetParameterIsString()
 
         ucrReceiverByFactor.Selector = ucrBarChartSelector
         ucrReceiverByFactor.SetIncludedDataTypes({"factor"})
@@ -339,8 +340,12 @@ Public Class dlgBarAndPieChart
         clsForecatsInfreqValue.SetPackageName("forcats")
         clsForecatsInfreqValue.SetRCommand("fct_infreq")
 
-        clsReorderFunction.SetRCommand("reorder")
         clsReorderFunctionValue.SetRCommand("reorder")
+        clsReorderFunctionValue.SetRCommand("reorder")
+
+
+        clsReorderFunction.SetRCommand("reorder")
+        clsReorderFunction.SetRCommand("reorder")
 
         clsAesFunction1.SetRCommand("aes")
 
@@ -371,7 +376,7 @@ Public Class dlgBarAndPieChart
         clsRgeomBarFunction1.SetPackageName("ggplot2")
         clsRgeomBarFunction1.SetRCommand("geom_bar")
         clsRgeomBarFunction1.AddParameter("data", clsRFunctionParameter:=clsSubsetFunction1, iPosition:=0)
-        clsRgeomBarFunction1.AddParameter("aes", clsRFunctionParameter:=clsAesFunction1, iPosition:=1, bIncludeArgumentName:=False)
+        clsRgeomBarFunction1.AddParameter("aes", clsRFunctionParameter:=clsAesFunction1, iPosition:=1)
 
         clsRgeomBarFunction2.SetPackageName("ggplot2")
         clsRgeomBarFunction2.SetRCommand("geom_bar")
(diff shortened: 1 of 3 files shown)


rdstern (changes requested):
@N-thony that is ok for the data option.  You can also do it for the frequency option.  It uses a slightly different function from forcats as explained in the new issue you are looking at.  (forcats is very powerful!)
I am confused about the label and also the position on the dialogue.  I was assuming it would go on the right, perhaps under the factor that it corresponds to.  Possibly just Reorder, but I am not sure now.  Once implemented for both options I suggest we merge and then discuss further.  I would then like to get @shadrackkibet view on this.

N-thony:
> @N-thony that is ok for the data option. You can also do it for the frequency option. It uses a slightly different function from forcats as explained in the new issue you are looking at. (forcats is very powerful!)
> I am confused about the label and also the position on the dialogue. I was assuming it would go on the right, perhaps under the factor that it corresponds to. Possibly just Reorder, but I am not sure now. Once implemented for both options I suggest we merge and then discuss further. I would then like to get @shadrackkibet view on this.

@rdstern Please have a look.

N-thony:
> > @N-thony that is ok for the data option. You can also do it for the frequency option. It uses a slightly different function from forcats as explained in the new issue you are looking at. (forcats is very powerful!)
> > I am confused about the label and also the position on the dialogue. I was assuming it would go on the right, perhaps under the factor that it corresponds to. Possibly just Reorder, but I am not sure now. Once implemented for both options I suggest we merge and then discuss further. I would then like to get @shadrackkibet view on this.
> 
> @rdstern Please have a look. There is a small issue for the frequency option that will good for me to fix after the quick chat with you.

I think reorder function requires x and y variables, that is why in my previous change I made it available only for the value option. Let me know if I'm wrong so that I can fix it quickly with your suggestion. Thanks.

N-thony:
@rdstern from our chat this moring, I made the first change by using forcats package. Please have a look and let me know on the next step.

rdstern (changes requested):
@N-thony it is getting there, but only slowly!  
You are now correctly using infreq for the frequency option.  But you are also now using it for the data option, where you were fine before.  There it should be by the variable.
On the position, please move the control up, so it is clearly associated with the first variable

Then add the same code for the second factor, when it is used.

N-thony:
> @N-thony it is getting there, but only slowly!
> You are now correctly using infreq for the frequency option. But you are also now using it for the data option, where you were fine before. There it should be by the variable.
> On the position, please move the control up, so it is clearly associated with the first variable
> 
> Then add the same code for the second factor, when it is used.

@rdstern please have a look.

rdstern (approved):
I am ok with this now.  It is an obvious dialogue for a check also by @shadrackkibet

instat/dlgBarAndPieChart.vb
```diff
@@ -113,7 +124,7 @@ Public Class dlgBarAndPieChart
 
         ucrReceiverX.Selector = ucrBarChartSelector
         ucrReceiverX.strSelectorHeading = "X Variable"
-        ucrReceiverX.SetParameterIsString()
+        ucrReceiverX.SetParameterIsRFunction()
```
shadrackkibet, line 127: But we do not need it as an Rfunction when it is an aesthetic?

instat/dlgBarAndPieChart.vb
```diff
@@ -464,7 +518,90 @@ Public Class dlgBarAndPieChart
...
+                    clsForecatsReverseValue.AddParameter("x", ucrReceiverByFactor.GetVariableNames(False), iPosition:=0, bIncludeArgumentName:=False)
+                    clsBarAesFunction.AddParameter("fill", clsRFunctionParameter:=clsForecatsReverseValue, iPosition:=1)
+                Case Else
+                    clsBarAesFunction.AddParameter("fill", ucrReceiverByFactor.GetVariableNames(False), iPosition:=1)
+            End Select
 
+        Else
+            Select Case strChangeTextReorder
+                Case strAscending
+                    clsReorderFunction.AddParameter("y", ucrVariablesAsFactorForBarChart.GetVariableNames(False), iPosition:=1, bIncludeArgumentName:=0)
```
shadrackkibet, line 580: - Use the correct parameters. In this case, it should be `x`? 
- What does ` bIncludeArgumentName:=0` mean? 
Check other places as well.

shadrackkibet (changes requested):

shadrackkibet:
@rdstern I don't like the default x-axis labels when you reorder. It is much presentable if we have the variable name. You could change it by going the subdialog but wouldn't be nice to retain it by default? 
image
I also feel like we could improve on the design of this dialog. Perhaps a group box will look nice?
IMG_20210917_150241_395

N-thony:
> @rdstern I don't like the default x-axis labels when you reorder. It is much presentable if we have the variable name. You could change it by going the subdialog but wouldn't be nice to retain it by default?
> image
> I also feel like we could improve on the design of this dialog. Perhaps a group box will look nice?
> IMG_20210917_150241_395

@shadrackkibet I agree with you about adding the possibility of renaming the x-axis labels. Aslo, the improvement of the design looks interesting but from your screenshot above I can't see all the options controls in the groupbox. If @rdstern agree with that I will make the change. Thanks.

rdstern:
I would be happy with a redesign of this dialogue.  We currently also have 2 possible sorts, but they are not x and y.  They are the instance of having 2 x variables.  If that is confusing I am happy that we just have one - the main x.  @N-thony is improving the main reorder dialogue, so we can always reorder before doing the graph.  

Remember there are also lollipop graphs.  So there is quite a lot to show if everything is in a group box on the left-hand side.

You also criticize - correctly 0 the "interesting" variable titles.  I have no problem going back to the names.  I would also like to geo 2 steps further.  First I would like the option to use variable labels as well as the name.  This seems to be described with a vignette from the [expss package here](https://cran.r-project.org/web/packages/expss/vignettes/labels-support.html)

I have a slightly different suggestion on these titles for your consideration.  This is as follows:
a) We have a `Titles` button on each of the plotting dialogues.  This takes you straight to the Titles tab in the sub-dialogue.
b) Currently that tab just has the 3 main titles.  (Title, sub-title and Caption.  The x-axis and y axis titles are on different (and rather crowded tab s).  So move the x-axis titles and y-axiss titles to this Titles tab.

N-thony:
@rdstern we discussed about this recently and agreed that we can keep it as it is for now. Or would you like me to go ahead with the changes you suggested above and the redesign?

rdstern (dismissed):
@N-thony and @shadrackkibet I would be happy to merge now and then consider further improvements later.

instat/dlgBarAndPieChart.vb
```diff
@@ -366,9 +426,11 @@ Public Class dlgBarAndPieChart
...
-        ucrReceiverByFactor.AddAdditionalCodeParameterPair(clsIsEqualToOperator1, New RParameter("left", 0), iAdditionalPairNo:=1)
-        ucrReceiverByFactor.AddAdditionalCodeParameterPair(clsIsEqualToOperator2, New RParameter("left", 0), iAdditionalPairNo:=2)
-        ucrReceiverByFactor.AddAdditionalCodeParameterPair(clsLevelsFunction, New RParameter("x", 0), iAdditionalPairNo:=3)
+        ucrReceiverByFactor.AddAdditionalCodeParameterPair(clsIsEqualToOperator1, New RParameter("left", 0), iAdditionalPairNo:=2)
+        ucrReceiverByFactor.AddAdditionalCodeParameterPair(clsIsEqualToOperator2, New RParameter("left", 0), iAdditionalPairNo:=3)
+        ucrReceiverByFactor.AddAdditionalCodeParameterPair(clsLevelsFunction, New RParameter("x", 0), iAdditionalPairNo:=4)
+        ucrReceiverByFactor.AddAdditionalCodeParameterPair(clsReorderFunctionValue, New RParameter("x", 0, bNewIncludeArgumentName:=False), iAdditionalPairNo:=5)
+        ucrReceiverX.AddAdditionalCodeParameterPair(clsReorderFunction, New RParameter("x", 0, bNewIncludeArgumentName:=False), iAdditionalPairNo:=1)
```
shadrackkibet, line 433: The parameters are still hidden for no reason? Same case in line 581, 592?

instat/dlgBarAndPieChart.vb
```diff
@@ -464,7 +523,82 @@ Public Class dlgBarAndPieChart
...
+        Else
+            ucrInputReorderValue.Visible = False
+            ucrInputReorderValue.SetText(strNone)
+        End If
+        If ucrVariablesAsFactorForBarChart.bSingleVariable Then
+            ucrInputAddReorder.Visible = True
+        Else
+            ucrInputAddReorder.Visible = False
+            ucrInputAddReorder.SetText(strNone)
+        End If
```
shadrackkibet, line 547: Can we make it a bit clean and easy to read? Perhaps nest the `if` statements? eg
```
  If rdoFrequency.Checked Then
            ucrChkLollipop.Checked = False
            If ucrVariablesAsFactorForBarChart.bSingleVariable Then
                ucrInputReorderX.Visible = True
            Else
                ucrInputReorderX.Visible = False
                ucrInputReorderX.SetText(strNone)
            End If
  Else
            If ucrVariablesAsFactorForBarChart.bSingleVariable Then
                ucrInputReorderValue.Visible = True
            Else
                ucrInputReorderValue.Visible = False
                ucrInputReorderValue.SetText(strNone)
            End If
 End If
```
What do you think?

shadrackkibet (changes requested):

N-thony:
@shadrackkibet please have a look. Thanks.

shadrackkibet:
@rdstern please retest and approve. Functionality shouldn't be affected.

rdstern (approved):

shadrackkibet (approved):

PR #6752, opened 2021-09-09
Script Dialog

### LINKED ISSUE #5986: Dialogue to facilitate running a new function in R-Instat
The script window allows me to run more or less any small sequence of R-commands.  I often use it for a new function from one of our existing packages, or maybe I add a new package and try a function.  
This works well on the examples that are almost always provided in the manual associated with the package.  
But I would then often like to:
a) run the example with data from R-Instat, rather than from within the example.
b) be able to save the results back into the Data Sheet or Data Book, so I can continue with my work using R-Instat in the "normal way".

(I know I can do this by migrating to RStudio, but that is essentially a one-way process.  I suggest sometimes I wish to continue with R-Instat, but there is one command missing in the process.)

I currently do this by using a dialogue such as Model > Fit Model > Fit Model Keyboard.  I choose any function and then write to the script window and edit what is there.  It would be much better to have a dialogue that is specially designed for this.

Here is an example.  I would like to do some declustering of extremes in climatic data.  This is in the extRemes package, which is already installed.  But the function isn't.

Here is a simple example adapted from the manual:

```
y <- rnorm(100, mean=40, sd=20)
y <- apply(cbind(y[1:99], y[2:100]), 1, max)
bl <- rep(1:3, each=33)
ydc <- extRemes::decluster(y, threshold= 60, r=1, groups=bl)
ydc
plot(ydc)
```
In the code above, the first 3 lines produce the data.  Then line 4 runs the function, which produces the object called ydc.  Then line 5 displays the object in the output window and the last line displays a plot.

Now I adapt this for the Ghana data using the Model > Fit > Fit Model dialogue as my guide:
 This gives the following code - when I also opt to save the model!
```
# Code generated by the dialog, Modelling

ghana <- data_book$get_data_frame(data_name="ghana")
attach(what=ghana)
model1 <- lm(rainfall~year, na.action = na.exclude)
data_book$add_model(model_name="model2", model=model2, data_name="ghana")
data_book$get_models(data_name="ghana", model_name="model2")
summary(object=model2)
detach(name=ghana, unload=TRUE)
rm(list=c("model2", "ghana"))
```

Now I adapt it for the new command:

```
# Code generated by the dialog, Modelling

ghana <- data_book$get_data_frame(data_name="ghana")
attach(what=ghana)
ydc <- extRemes::decluster(rainfall, threshold= 60, r=1, groups=year, na.action=na.exclude)
data_book$add_model(model_name="ydc", model=ydc, data_name="ghana")
data_book$get_models(data_name="ghana", model_name="ydc")
ydc
plot(ydc

detach(name=ghana, unload=TRUE)
rm(list=c("ydc", "ghana"))
```
I would like the new dialogue to facilitate this process for a variety of commands that I may wish to run.

I need help with the construction of this dialogue, so these are just my initial ideas.  
There are 2 (perhaps 3) parts to the dialogue, namely Get Data, Run Function, (optional), then Display and Save Results.
1) The last control is the 5 bottom buttons, plus the Comment.  Here the OK is always disabled.  The Script is enabled once the dialogue is complete.  Or perhaps OK writes to the script window, i.e. they are identical.  I quite like that.
2) I assume the command might need access to specific variables, or perhaps a whole data frame, or to an object from the data frame.  Perhaps these are buttons at the top, namely Data Frame, Variables, Objects.
3) So perhaps the Get Data part is quite simple?  If Data Frame, then the selector just shows the data frames and we select one in the usual way.  If Variables, then we get the "usual" selector and a multiple receiver.  
4. In some instances this is all we need.  In that case there is sufficient information to:
a) get the data - as shown above
b) Add a comment line for the function.  Perhaps mention `# object1 <- package::function(..., na.action=na.exclude)`
c) Add a further comment to suggest additional actions. # display and save results
d) add a version of the detach and rm commands, see above.

This would already be useful!  Maybe start there when constructing the dialogue and test this part.  Where does this dialogue go?  Perhaps in the Tools menu or in the Edit Menu.  Perhaps it is called Add Function?

5) There could be a Function part to the dialogue.  This may just ask for the package name and function name?  We have a dialogue for this already! It is **_Help > R Packages and Commands_**.  It could also include the second Help button which is only enabled once at least one of these is given.  Then it would change the line above to at least specify the package and command name.  It could have a control that asks for the object name, that the user supplies, default is object1.

6) Then the part I am less sure about in the dialogue and also the code.  The actions here would be useful, even if not exhaustive.  This might be a set of check-boxes. 
a) Display result in object window Uses Summary?
b) Display plot - Uses Plot
c) Save variable back to data frame - control with name - and even position if possible?
d) Save Object

I hope @dannyparsons and @shadrackkibet could critique and help conceptualize the dialogue.  I am not assuming they would then need to do the work, though they may add support.

Patowhiz:
Fixes #5986.


rdstern (changes requested):
Tried, but is the menu item not visible?

Patowhiz:
@rdstern you can now test this.

rdstern:
I wonder about having 3 or 4 group boxes on the left.  They may only contain (for now) a single control, but they show the sequence.
They might be called:
```
Preliminary
Use data from R-Instat
Save Results
Finally
```

Ok, the first group box just has the current Get Library control for now.  Possibly underneath there is a label (Not needed for base, stats, util, are the others loaded automatically?)
There is also an Add Comment button.  This adds the contents of the Comment field to the top of the script.  (I still suggest we leave the comment field where it is.)
In finally I - for now suggest 2 checkboxes  
First is unchecked and has label Detach Packages.  (I am suggesting this be unchecked, because it does no real harm now, if a few more packages are left in the memory.
The second is a checkbox, Remove Objects.  This is checked and is the rm command.  If checked it removes all the structures - data frames, etc, generated by the command as far as can be done.  It would have to be added to, if more objects are generated by code for the user.  The default is rm( ).

In Use `Data from R-Instat` I suggest 2 radio buttons labelled Data Frame and Variables, default is Data Frame.  If Data Frame, then as now.  If Variables then there is the usual data selector, with the data frame and the variables together with thr Add and also the Data options!  (Check this last bit with @dannyparsons.  If Data options then it would execute the option and also put the code for the filter into the script.

In the Save results there could again be 2 radio buttons.  First with the label  New Data Frame and second Same Data Frame.  If New, then just the option you have now.  If Same, then checkboxes, default unchecked.  Also, either the option Same, or these options are disabled, until you have chosen a data frame on which to work..  
Then one option is Save Object, and this is the Save control that is on many of the dialogues.  Do we need to be more specific.
Second option is Save Variable.  This gives your control, including the Position button!

 Is there anyway we can divide the main field into sections (5) so the information from each of the categories on the left goes into the appropriate section?  The fifth section - in the middle) is for the code that the user adds, often initially a copy/paste.

Patowhiz:
@lloyddewit the new requirements of comments suggested by @rdstern may require me change how comments are added to the output window. I'm happy to discuss this with you first and do it in a separate PR. This change will also be required for new data frame dialog comments addition(@rdstern  is yet to create the issue for this).

@rdstern for now, the comments from this dialog won't be shown in green in the output window.

Patowhiz:
@rdstern you can now test this.

lloyddewit:
@Patowhiz This PR currently has `frmMain` conflicts.
@shadrackkibet Do you think it would be practical to include these `frmMain` changes in PR #6850? Could we then pull them back into this PR and resolve the conflicts?

rdstern (approved):
When last seen I would like to see this merged.  It is a major new dialogue.  Needs @dannyparsons or @shadrackkibet to confirm this would do no harm - and also to make it visible

lloyddewit (approved):

Issue #6426, opened 2021-04-24
Adding the select facility for variables as a partner to the filter system


rdstern:
This has 2 parts.  It builds on, and supersedes topics mentioned in #4372, in #4957 and also in #6263 and in #4009.
The first part is to (probably) replace the current Prepare > Data Frame > Hide/Show Columns with a new Select Columns dialogue.
a) There will still be the Hide Columns (and Restore Hidden Columns) on the right-click menu as described in #4372, together with information on what is hidden at the bottom - just like for rows when there is a filter.
b) The main dialogue will include hide as an option.  It will probably be largely like the main Prepare > Data Frame > Filter dialogue, so the data selector has selects as objects.  You can choose to define a new select, or use an existing one.  One thing you can do with a select is to hide the columns.
c) As with filter, another route to select is from the Data Options button on each main dialogue.  This currently takes you to a version of the Filter dialogue, but just for that data frame.  Now that will change with an additional button, so you can opt for Rows or Columns.

![image](https://user-images.githubusercontent.com/11226469/115952616-dfd67180-a4de-11eb-9a05-b851fc274f92.png)

It is already here.  Do we keep these tabs - just what we want, or change them to radio buttons?

Currently when you choose Columns it is almost blank.  This would change to be like the dialogue for Rows.
I suggest it can for now be almost an exact copy of this.  
So the data selector is for selections.  Replace Filter by Selection in most of these controls.  Perhaps on the Rows option change Selected Filter Preview to just Filter Preview.  Then on the Columns have Select Preview.

d) Now the Select sub-dialogue.  This is a parallel of the Filter sub-dialogue and many parts are the same.  Here I start the suggestions on the structure of this sub-dialogue:

![image](https://user-images.githubusercontent.com/11226469/115952915-bae2fe00-a4e0-11eb-978d-b44320083e94.png)

1) It has the usual data selector at the left-hand side - though it is only needed for one of the options.
2) It has the same controls below, i.e. 
![image](https://user-images.githubusercontent.com/11226469/115953001-02698a00-a4e1-11eb-9b59-ed2a87a55f0c.png)
But the main box has a single entry - just Condition.
And Filter becomes Select, so Select Preview, and Select Name

3) Then the Filter By  becomes Select By: And the single receiver becomes a pull down with the options specified below.
![image](https://user-images.githubusercontent.com/11226469/115953218-12359e00-a4e2-11eb-9e57-6f50fe7f177d.png)

[...]

PR #6755, opened 2021-09-09
R Implementation of Column Selections

### LINKED ISSUE #6758: Add column selections (filters) to the DataBook
Add column selection objects in the DataBook and DataSheet which can be created like row filters as a list of conditions.

The full specification is here #6426

dannyparsons:
R implementation of variable filtering system described here https://github.com/africanmathsinitiative/R-Instat/issues/6426

fixes #6758

This adds change to the DataBook and DataSheet.

To test this use an example such as below.

Like row filtering, column selection is done with a list of conditions.

Conditions contain:
- "operation", which is an R function which will match column names from a condition. The supported functions are defined in `labels_and_default.R` as:
`column_selection_operations <- c("base::match", "tidyselect::starts_with", "tidyselect::ends_with", "tidyselect::contains", "tidyselect::matches", "tidyselect::num_range")` (`base::match` is used to give a list of columns). Internally, each of these functions returns the column numbers in the selection. We do this internally by passing the `vars` argument to the `tidyselect` function, so this is not done on the data directly. We can easily add to this list of functions.
- "parameters" which are a list of parameters from the "operation" function which should be passed e.g. "match" for the `tidyselect::` functions but also any other parameters.
- "negation" (optional) TRUE/FALSE. Default FALSE. If TRUE the negation of the condition is applied.

The main function is then `data_book$add_column_selection`. I've chosen "or" as the default for combining the conditions, unlike for rows where "and" is the default. I think for columns it will be more common for multiple conditions to be "or", but happy to change this.

Currently, when you do `data_book$set_current_column_selection` this only affects what is returned by `data_book$get_column_names` i.e. it only affects the list of variables in dialogs, not the displayed data in the grid. We can also get this to only show those columns in the Data View, this just needs a bit of extra code and checks e.g. what happens if no columns are selected.

@shadrackkibet Please test this out in R with some data and review. You could test how well it will work in dialogs by running the lines below in the script window. I would suggest merging this first before you work on the implementation in VB to keep the PRs small. Although there's no VB changes, please also test that dialogs work ok and still load columns correctly.

@lloyddewit I would appreciate your review on the changes to the definitions to the `DataBook` and `DataSheet`. In particular, as changes to these structures need to consider backwards compatibility of existing RDS files which don't have the column selection features. This should all be dealt by `data_sheet$clone_data_object` called from `data_book$import_RDS`. It seems to run ok on my existing files but a sanity check of the logic would be useful.

```r
conds <- list(list(operation = "base::match",
                   parameters = list(x = c("YEAR", "MONTH", "STATION"))),
              list(operation = "tidyselect::starts_with",
                   parameters = list(match = "ST", ignore.case = FALSE)))
data_book$add_column_selection("Append", 
                               name = "simple",
                               column_selection = conds, and_or = "|")
data_book$set_current_column_selection("Append", "simple")
data_book$get_column_names(data_name = "Append", use_current_column_selection = TRUE)
```


lloyddewit (approved):

lloyddewit:
@dannyparsons I reviewed and approved because I did not see that your review request was cancelled. I had no comments. This may be because my R knowledge is passive. I agree that it is better for an active R programmer like @shadrackkibet to review.

shadrackkibet:
Thanks, Danny this is noted. I will be testing it and then move it forward.

dannyparsons:
Thanks @lloyddewit. As long as it works with existing files and new files from Shadrack's tests then I think it's ok as it doesn't add any functionality to the front end.

shadrackkibet (approved):
I tested this and couldn't find anything to report. I think it is safe to merge.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -824,6 +835,65 @@ DataBook$set("public", "get_filter_as_instat_calculation", function(data_name, f
...
+
+DataBook$set("public", "get_column_selection", function(data_name, name) {
+  return(self$get_data_objects(data_name)$get_column_selection(name))
+}
+)
+
+DataBook$set("public", "get_column_selection_column_names", function(data_name, filter_name) {
+  return(self$get_data_objects(data_name)$get_filter_as_logical(filter_name))
+}
+)
```
shadrackkibet, line 864: @dannyparsons I found an issue here. This is using the filter method. We need a parallel to this for column selection.

PR #6756, opened 2021-09-10
Added Dumbbell and Slope graphs

HawardKetoyoMsatsi:
@rdstern @africanmathsinitiative/developers @lilyclements 
This is ready for review


rdstern (changes requested):
That's a really good start.
a) The label on dumbbell still says Line Options.  It correctly goes to the dumbbell geom, so it is just the label that needs to change.
b) On the slope graph the second x axis is now limited to be a factor variable.  And the label on the treat is optional - which it isn't!

Then some of the options will be included on the main dialogue.  I'll also have a look.
A big one will be to add labels, which should use the same set of controls (initially) from the bar chart dialogue.

instat/dlgLinePlot.vb
```diff
@@ -168,6 +188,10 @@ Public Class dlgLinePlot
         ucrChkAddSE.SetParameter(New RParameter("se", 1))
         ucrChkAddSE.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
 
+        ucrChkRemoveMissing.SetText("Remove Missing")
```
lilyclements, line 191: Perhaps "Remove Missing Values" (or whichever missing "thing" it removes)

lilyclements (commented):
1. "Smooth options" button should say "Smooth Options" to be consistent with other buttons on the dialog. Similarly with "Line options".
2. The "Use path or step" checkbox should say "Use Path or Step" to be consistent with other buttons on the dialog
3. Dumbbell colours for x and xend (or by default x and xend should be different colours and shapes, with options for this either on the dialog or on the sub). I'm not sure if further options should be on the sub or main dialog, however, a different shape and colour for the x/xend points should be given by default.
4. Just a strange one - We have "Smooth Options" as a button, and "Smooth Options" as the name of a group box. Perhaps the group box label should change to something more specific?
5. Another general one - the comment when this is run is "# Code generated by the dialog, Line". Can the dialog description change so that the comment is mentions the name of the whole dialog?  (Perhaps this needs to be done in another issue)
6. For dumbbell plots, if `x` is a factor, then `xend` can be a factor or numeric. At the moment you only let numeric variables as `xend`.

lilyclements:
To expand on point 3: The options in the `geom_dumbbell` function are `colour_x` and `colour_xend` to change the colours, so perhaps just pick two colours to be the start and end points. 
Having a look at the function, they don't seem to have shape options easily for the start/end points, so perhaps pick default colours which show up differently in black and white, and for colour-blind. (Orange and purple?)

I'm not sure if we would want the option to change start and end colour points on the main dialog or sub dialog

HawardKetoyoMsatsi:
@rdstern I have made the above corrections I however would like to hear from you if we can include the colour_x and colour_xend in the main dialog @lilyclements is not quite sure about that. 
@lilyclements  your points 1 and 2 the dialog seems to have captured that correctly.

lilyclements:
@HawardKetoyoMsatsi can you sync your changes

HawardKetoyoMsatsi:
> @HawardKetoyoMsatsi can you sync your changes

HI @lilyclements could you please check now.

rdstern:
I think now the x can only be a factor.  Numeric should be allowed.
I tried the example from [here](https://www.r-bloggers.com/2021/08/ggalt-make-a-dumbbell-plot-to-visualize-change-in-ggplot2/) - which I think is a useful exercise.  I was (quite) easily able to do the data wrangling, except my variables are of length 40, rather than 36, because I still have some missing values in the data.  I have left them there, because I am curious to know whether the dialogue copes with it.

Here are the data after the wrangling:  
[mpg for dumbbell.zip](https://github.com/africanmathsinitiative/R-Instat/files/7163582/mpg.for.dumbbell.zip)

Here is the plot:
image

By comparison here it is in the article:
image

It looks better there and the factor is sorted.  I can sort the factor here - perhaps, like the bar chart, this could be included in the dialogue?

I can use the reorder function and then also change the size and colour of the endpoints to give the following:
image

I suggest a checkbox labelled Colour.  If checked, then there are the 3 options X X End and Line
A second checkbox is labelled Size.  If checked, then again the controls for X X End and Line.

(I should say that the X and X End options seem to work fine, but I couldn't see the effect of the line options - even though the code generated looked fine.)

rdstern:
The slope chart.  here is the slope chart for the same data.  As it is a function, I note there are no options.  So anything we need, should be added to the dialogue.  
image

The first is the control over the legend.  I suggest the default is not to have a legend.  Then, what are the other options?  Nice to have the labels.  Could we opt to not have them?  If so, then, for consistency, I suggest we have a Labels checkbox, but the default is for it to be checked.  That's where any options for the labels - size, etc, will go.

Then, there is one oddity on the shape of the data for the dumbbell and slope.  For the dumbbell there is the Single Variable/Multiple Variable option.  But (I think) the data need to be unstacked for the dumbbell plots.  So I am not clear that the Single/Multiple applies.

However, for the slope plot I think the data need to be stacked.  That means we could have the Single/Multiple control there, and the variable is then the x-axis.  Perhaps @lilyclements could check this?

lilyclements:
> The first is the control over the legend. I suggest the default is not to have a legend. Then, what are the other options? Nice to have the labels. Could we opt to not have them? If so, then, for consistency, I suggest we have a Labels checkbox, but the default is for it to be checked. That's where any options for the labels - size, etc, will go.

A theme function should be running in R-Instat which should remove the labels (like it was in the `CGPfunctions` package). I called this `newggslopegraph_theme` or something. @HawardKetoyoMsatsi can you check why this is not running in the output given by @rdstern. The `newggslopegraph_theme` function contains lots of options where we can change parameters relating to the labels.

I will check over the other points shortly.

HawardKetoyoMsatsi:
@rdstern I made the above changes partly, a few are yet to be fixed. @lilyclements I will check that.

rdstern:
Great.  I'll check

HawardKetoyoMsatsi:
@lilyclements I have figured out the theme_grey() layer added automatically makes the themes function you created not work, swapping the order of the additional theme layers corrects this. @rdstern There is a lot of linking on this dialog messing resetting of the checkboxes and setting defaults once I fix it with support from @shadrackkibet I will push the changes.

HawardKetoyoMsatsi:
@rdstern I have made changes to the dialog please take a look.

HawardKetoyoMsatsi:
> @HawardKetoyoMsatsi well done. These both seem to be working pretty well. a) On the slope graphs the only problem I could fine was when I used the option NULL, which is in the Colour pul-down for the Label, Text and Line. It gives an error mostly as it looks for a colour called NULL. I think it might need to do something different here. b) Trivial point, but the vertical spacing between Label, Text and Line, would look neater if it were equal. Perhaps move Line down just a little. c) I assume this is now sort of our own function and hence within the R-Instat code. I hope the code includes a comment that gives the source. Can you give me the comment so I can also add it to the help file. d) The name newggslopegraph_amended is very long. Could it be ggslopegraph or maybe just slopegraph? e) In Dumbbell the main problem was the Dumbbell Options. I like the label, but it goes to geom_line, not geom_dumbbell.

@lilyclements could you please write a comment that Roger needs about the newggslopegraph_amended function which I changed the name to slopegraph in the stand_alone_functions.R file. I think you will be in a better position to write this comment since you wrote the function.

HawardKetoyoMsatsi:
@rdstern, the error is fixed please take a look.

rdstern (changes requested):
The one item now is that the dumbbell option has our usual Single/Multiple variable control.  The slope graph has (instead) a Y-variable.  Could you please use that control also for the dumbbell option, as it won't work if you change to Multiple.  Also the label Y Variable is clearer.

HawardKetoyoMsatsi:
@rdstern the above change has been made please take a look.

rdstern (approved):
Looks ok now!

lloyddewit:
@N-thony Please could you peer review?
How comfortable are you reviewing the R code?
Thanks

lloyddewit:
> @N-thony Please could you peer review? How comfortable are you reviewing the R code? Thanks

I discussed with @HawardKetoyoMsatsi .
@N-thony Please could you review the VB code?
@shadrackkibet Please could you review the R code?
Thanks!

N-thony (changes requested):
@lloyddewit sorry for the delay.
@HawardKetoyoMsatsi looks good, just few comments.

HawardKetoyoMsatsi:
@N-thony I have made part of the changes. You are right lines 664(now 663) and 700(now 699) should be `Else` and not  `ElseIf rdoLine.Checked Then` but let me confirm first with @shadrackkibet and @anastasia-mbithe they may have been the ones who made the changes first because they are changes that were already merged into master before I worked on the dialog.

HawardKetoyoMsatsi:
@N-thony the above corrections have been made, please take a look.

N-thony:
> @N-thony the above corrections have been made, please take a look.

@HawardKetoyoMsatsi nice.
@shadrackkibet over to you.

instat/dlgLinePlot.vb
```diff
@@ -47,6 +47,9 @@ Public Class dlgLinePlot
     Private clsAnnotateFunction As New RFunction
     Private clsListFunction As New RFunction
     Private clsFormulaFunction As New RFunction
+    Private clsggSlopeFunction As New RFunction
```
lloyddewit, line 50: Follow naming conventions?
```suggestion
    Private clsGgSlopeFunction As New RFunction
```

lloyddewit (changes requested):
Just one small suggestion for the VB code

HawardKetoyoMsatsi:
@lloyddewit am done back to you.

rdstern (approved):
These now seem to work.

lloyddewit (approved):

PR #6764, opened 2021-09-13
Adding Tree maps to the Bar chart dialog

anastasia-mbithe:
Hello @rdstern and @shadrackkibet, I have managed to add tree maps to the bar chart dialog and some basic functions. You can review this and give more suggestions on what should be added.
I am yet to work on adding labels to the tree maps.

### DIFF
--- a/instat/dlgBarAndPieChart.Designer.vb
+++ b/instat/dlgBarAndPieChart.Designer.vb
@@ -63,7 +63,6 @@ Partial Class dlgBarAndPieChart
         Me.lblPlace = New System.Windows.Forms.Label()
         Me.ttArea = New System.Windows.Forms.ToolTip(Me.components)
         Me.ttFill = New System.Windows.Forms.ToolTip(Me.components)
-        Me.ucrChkSameLabelAsFill = New instat.ucrCheck()
         Me.ucrInputPlace = New instat.ucrInputComboBox()
         Me.ucrReceiverLabel = New instat.ucrReceiverSingle()
         Me.ucrChkLayout = New instat.ucrCheck()
@@ -231,12 +230,6 @@ Partial Class dlgBarAndPieChart
         '
         resources.ApplyResources(Me.lblPlace, "lblPlace")
         Me.lblPlace.Name = "lblPlace"
-        '
-        'ucrChkSameLabelAsFill
-        '
-        Me.ucrChkSameLabelAsFill.Checked = False
-        resources.ApplyResources(Me.ucrChkSameLabelAsFill, "ucrChkSameLabelAsFill")
-        Me.ucrChkSameLabelAsFill.Name = "ucrChkSameLabelAsFill"
         '
         'ucrInputPlace
         '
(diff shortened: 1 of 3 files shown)


rdstern:
@anastasia-mbithe this is a very good start.
a) Can you add a tooltip to the area control.  "A numeric-type variable that gives the area of each tile, or rectangle"
b)  On the fill a tooltip "A scale to fill each tile or rectangle.  This may be the same as the variable used for the areas.
c) Allow for a factor in the area, variable.  This isn't allowed in the dialogue, so in the R code, if it is not numeric, then include as.numeric(x) in the area variable
d) Change the labelling option.  It is simpler than for the bar chart.  Instead it calls geom_treemap_text.  You can keep the position control (it is the place argument in this function) with alternatives topleft, top, topright, bottomleft, bottom, bottomright.  Then have a furst checkbox for the variable.  If that is unchekced, then the label is the same as the fill variable that you have already.  If you do, then it opens a single receiver and you can have any variable there.
e) The option button is still calling the bar chart geom.  It sometimes says Treemap, but even then it calls the bar chart.
f) Add a control to call a fill scale.  This can be the same as used in the new heatmap dialogue.  One complication (which may be the same as with the heatmap?) is that your fill variable could be numeric or a factor.  If it is a factor, then you have to add discret = TRUE to the R code.
g) Finally - not in this dialogue, but in the geom_treemap_text - in the plot sub-dialogue, the label aesthetic is currently limited to being a factor.  It should allow anything.

anastasia-mbithe:
Hello @rdstern and @shadrackkibet  I added the tooltips for area and fill receiver, and set area variable to pass as.numeric(x) and added the treemaptext geom in the option button.
I added the label checkbox but I feel its complicating the code when the receiver is linked twice; to the label checkbox and also to the add labels checkbox at the bottomleft.

rdstern:
@anastasia-mbithe just checking that the labels are only added in this graph using the special treemap_text geom.  This is instead of the old label, which can't be usef, because there isn't an x or y in the treemap graph.

anastasia-mbithe:
@rdstern  yeah, once the tree map radio button is selected the geom in add labels changes from geom_text to geom_treemap_text

rdstern:
@anastasia-mbithe it is looking better.  I now suggest not the checkbox for Another Variable.  When you open the label checkbox, just add the receiver there, (on the right) with (as you have) the variable from Fill completed as the default.  It can just say Variable:  
By the way, when I used another variable there I didn't get it displayed and it reverted back to the original when I reopened the dialogue.

anastasia-mbithe:
@rdstern yeah that's one issue am looking at, upon reopening, Fill does not pass the variable to Label as it should in the code. That's why it does not display the chart with labels.

anastasia-mbithe:
@rdstern @shadrackkibet  I am closing this PR because the branch had some conflicts. This work is available in PR #6805

anastasia-mbithe:
I am closing this PR because the branch had some conflicts. This work is available in PR #6805

PR #6766, opened 2021-09-14
Added shape receiver to the heat map dialogue

EstherNjeriLiberatta:
@africanmathsinitiative/developers @rdstern @shadrackkibet @fran2or 
This is ready for review.

### DIFF
--- a/instat/dlgRugPlot.vb
+++ b/instat/dlgRugPlot.vb
@@ -45,9 +45,9 @@ Public Class dlgRugPlot
     Private clsLabelAesFunction As New RFunction
     Private clsColourPaletteFunction As New RFunction
     Private clsGeomPointSizeFunction As New RFunction
-    Private clsSizeAes As New RFunction
     Private clsGeomPointShapeFunction As New RFunction
-    Private clsShapeAes As New RFunction
+    Private clsShapeAesFunction As New RFunction
+    Private clsSizeAesFunction As New RFunction
 
     'Parameter names for geoms
     Private strFirstParameterName As String = "geomrug"
@@ -72,7 +72,6 @@ Public Class dlgRugPlot
         Dim dctLabelPositions As New Dictionary(Of String, String)
         Dim dctLabelSizes As New Dictionary(Of String, String)
         Dim dctColourPallette As New Dictionary(Of String, String)
-
 
         ucrBase.iHelpTopicID = 476
         ucrBase.clsRsyntax.bExcludeAssignedFunctionOutput = False
@@ -102,6 +101,10 @@ Public Class dlgRugPlot
         ucrReceiverPoints.SetParameterIsString()
         ucrReceiverPoints.Selector = ucrHeatMapSelector
         ucrReceiverPoints.bWithQuotes = False
+
+        ucrNudShape.SetParameter(New RParameter("size", 2))
+        ucrNudShape.Visible = False
+        ucrNudShape.SetLinkedDisplayControl(lblPointsSize)
 
         ucrSaveGraph.SetPrefix("heatmap")
         ucrSaveGraph.SetSaveTypeAsGraph()
@@ -163,9 +166,9 @@ Public Class dlgRugPlot
         clsLabelAesFunction = New RFunction
         clsColourPaletteFunction = New RFunction
         clsGeomPointSizeFunction = New RFunction
-        clsSizeAes = New RFunction
+        clsSizeAesFunction = New RFunction
         clsGeomPointShapeFunction = New RFunction
-        clsShapeAes = New RFunction
+        clsShapeAesFunction = New RFunction
 
         ucrSaveGraph.Reset()
         ucrVariableAsFactorForHeatMap.SetMeAsReceiver()
@@ -197,27 +200,25 @@ Public Class dlgRugPlot
 
         clsLabelAesFunction.SetPackageName("ggplot2")
         clsLabelAesFunction.SetRCommand("aes")
-
+        clsLabelAesFunction.AddParameter("label", ucrVariableAsFactorForHeatMap.GetVariableNames(False), iPosition:=0)
 
         clsGeomPointSizeFunction.SetPackageName("ggplot2")
         clsGeomPointSizeFunction.SetRCommand("geom_point")
-        clsGeomPointSizeFunction.AddParameter("mapping", clsRFunctionParameter:=clsSizeAes, iPosition:=0)
+        clsGeomPointSizeFunction.AddParameter("mapping", clsRFunctionParameter:=clsSizeAesFunction, iPosition:=0)
 
-        clsSizeAes.SetPackageName("ggplot2")
-        clsSizeAes.SetRCommand("aes")
+        clsSizeAesFunction.SetPackageName("ggplot2")
+        clsSizeAesFunction.SetRCommand("aes")
 
         clsGeomPointShapeFunction.SetPackageName("ggplot2")
         clsGeomPointShapeFunction.SetRCommand("geom_point")
-        clsGeomPointShapeFunction.AddParameter("mapping", clsRFunctionParameter:=clsShapeAes, iPosition:=0)
+        clsGeomPointShapeFunction.AddParameter("mapping", clsRFunctionParameter:=clsShapeAesFunction, iPosition:=0)
+        clsGeomPointShapeFunction.AddParameter("size", 5, iPosition:=2)
 
-        clsShapeAes.SetPackageName("ggplot2")
-        clsShapeAes.SetRCommand("aes")
-        clsShapeAes.AddParameter("size", 5, iPosition:=1)
+        clsShapeAesFunction.SetPackageName("ggplot2")
+        clsShapeAesFunction.SetRCommand("aes")
 
         clsColourPaletteFunction.SetPackageName("viridis")
         clsColourPaletteFunction.SetRCommand("scale_fill_viridis")
-
-        clsLabelAesFunction.AddParameter("label", ucrVariableAsFactorForHeatMap.GetVariableNames(False), iPosition:=0)
 
         clsBaseOperator.AddParameter(GgplotDefaults.clsDefaultThemeParameter.Clone())
         clsXlabsFunction = GgplotDefaults.clsXlabTitleFunction.Clone()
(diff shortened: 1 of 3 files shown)


lloyddewit:
@N-thony Please could you peer review? thanks

instat/dlgRugPlot.vb
```diff
@@ -214,6 +234,7 @@ Public Class dlgRugPlot
 
     Public Sub SetRCodeForControls(bReset As Boolean)
         ucrReceiverFill.AddAdditionalCodeParameterPair(clsLabelAesFunction, New RParameter("label", 2), iAdditionalPairNo:=1)
+        ucrReceiverShape.AddAdditionalCodeParameterPair(clsShapeAes, New RParameter("size", 2), iAdditionalPairNo:=1)
```
N-thony, line 237: Set the position to New RParameter("size", 0)

instat/dlgRugPlot.vb
```diff
@@ -185,6 +196,15 @@ Public Class dlgRugPlot
         clsLabelAesFunction.SetPackageName("ggplot2")
         clsLabelAesFunction.SetRCommand("aes")
 
+
+        clsShapeFunction.SetPackageName("ggplot2")
+        clsShapeFunction.SetRCommand("geom_point")
+        clsShapeFunction.AddParameter("mapping", clsRFunctionParameter:=clsShapeAes, iPosition:=1)
```
N-thony, line 202: set iPosition = 0?

instat/dlgRugPlot.vb
```diff
@@ -133,6 +139,9 @@ Public Class dlgRugPlot
         ucrChkColourPalette.AddParameterPresentCondition(True, "option")
         ucrChkColourPalette.AddParameterPresentCondition(False, "option", False)
 
+        ucrReceiverShape.AddParameterPresentCondition(True, "geom_point")
+        ucrReceiverShape.AddParameterPresentCondition(False, "geom_point", False)
```
N-thony, line 143: Move these after line 100

instat/dlgRugPlot.vb
```diff
@@ -237,7 +259,7 @@ Public Class dlgRugPlot
         End If
     End Sub
 
-    Private Sub AllControlsContentsChanged() Handles ucrReceiverX.ControlContentsChanged, ucrSaveGraph.ControlContentsChanged, ucrVariableAsFactorForHeatMap.ControlContentsChanged, ucrChkAddLabels.ControlContentsChanged, ucrReceiverFill.ControlContentsChanged, ucrChkColourPalette.ControlContentsChanged
+    Private Sub AllControlsContentsChanged() Handles ucrReceiverX.ControlContentsChanged, ucrSaveGraph.ControlContentsChanged, ucrVariableAsFactorForHeatMap.ControlContentsChanged, ucrChkAddLabels.ControlContentsChanged, ucrReceiverFill.ControlContentsChanged, ucrChkColourPalette.ControlContentsChanged, ucrReceiverShape.ControlContentsChanged
```
N-thony, line 262: ucrChkColourPalette.ControlContentsChanged, ucrReceiverShape.ControlContentsChanged removed? since they are not in the sub TestOkEnabled

N-thony (changes requested):

instat/dlgRugPlot.vb
EstherNjeriLiberatta: Hello @N-thony, I've made the changes you have requested.

N-thony: @lloyddewit over to you.

rdstern:
image

I am unconvinced by the extra value of this shape control.  It is just adding a point of varying size in the middle of each cell.  It isn't an extra I am very familiar with, in a heatmap.  Is this common?  If so, then we probably need more options, or we call it a size option, because there is only a single shape - namely a circle!

fran2or:
> image
> 
> I am unconvinced by the extra value of this shape control. It is just adding a point of varying size in the middle of each cell. It isn't an extra I am very familiar with, in a heatmap. Is this common? If so, then we probably need more options, or we call it a size option, because there is only a single shape - namely a circle!

@rdstern  I think you are right that the shape type and size feature is not common with heatmaps, however I think having it as an option (dialog should still work without it being checked or filled ) may be useful when people want to look at two measurements together (eg. yield and fertilizer amount - this is the single shape with  varying size). Another example which may make use of the type of shape could be yield (heatmap) and grouped fertilizer amounts (different shapes representing varied categories of  fertilizer amount groups).

fran2or:
Having discussed with @rdstern, he now agrees that this is a very useful addition.  Could you kindly change the label for that control from Shape to Points (Optional):  It should also detect whether the extra variable is numeric or a factor.  If factor, it uses different shapes and if numeric it uses different sizes.  Also please tell @shadrackkibet  that an extra geom is being used, for when he alters to Options button.

EstherNjeriLiberatta:
Hello @rdstern @shadrackkibet @fran2or, this is ready for review.

EstherNjeriLiberatta:
Hello @africanmathsinitiative/developers, This is ready for review.
@rdstern @shadrackkibet @fran2or, I have made the requested suggestions on the shape receiver.

instat/dlgRugPlot.vb
```diff
@@ -185,6 +198,22 @@ Public Class dlgRugPlot
...
+        clsGeomPointSizeFunction.SetPackageName("ggplot2")
+        clsGeomPointSizeFunction.SetRCommand("geom_point")
+        clsGeomPointSizeFunction.AddParameter("mapping", clsRFunctionParameter:=clsSizeAes, iPosition:=0)
+
+        clsSizeAes.SetPackageName("ggplot2")
+        clsSizeAes.SetRCommand("aes")
+
+        clsGeomPointShapeFunction.SetPackageName("ggplot2")
+        clsGeomPointShapeFunction.SetRCommand("geom_point")
+        clsGeomPointShapeFunction.AddParameter("mapping", clsRFunctionParameter:=clsShapeAes, iPosition:=0)
```
shadrackkibet, line 211: Only one `geom_point` and `aes` function should be enough. You only need to change `aes` parameters depending on the variable in the receiver. Does this make sense to you? If this complicates things for you then it is not worth it.

shadrackkibet (changes requested):

shadrackkibet:
@EstherNjeriLiberatta your design looks clean and controls are well aligned. I like it! 

Could you also make sure that the Tab order is sorted? Ask if you are not sure what that means.

shadrackkibet:
> Having discussed with @rdstern, he now agrees that this is a very useful addition. Could you kindly change the label for that control from Shape to Points (Optional): It should also detect whether the extra variable is numeric or a factor. If factor, it uses different shapes and if numeric it uses different sizes. Also please tell @shadrackkibet that an extra geom is being used, for when he alters to Options button.

@fran2or thanks for the alert. Well noted!

instat/dlgRugPlot.vb
EstherNjeriLiberatta: @shadrackkibet Thank you for the corrections.

EstherNjeriLiberatta: It was quite confusing adding and removing parameters to the function so I decided to define more functions. If it is okay with you, please allow me to use the functions as I have them for now.

shadrackkibet: No problem. Let me know when you are done with other changes. I can help to sort it out.

instat/dlgRugPlot.vb
```diff
@@ -185,6 +198,22 @@ Public Class dlgRugPlot
...
+
+        clsShapeAes.SetPackageName("ggplot2")
+        clsShapeAes.SetRCommand("aes")
+        clsShapeAes.AddParameter("size", 5, iPosition:=1)
```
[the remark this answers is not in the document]
shadrackkibet: Just to add that if you are fixing size to numeric then you should add it into the `geom_point()` and not into `aes()`.

EstherNjeriLiberatta:
Hello, @rdstern @shadrackkibet This is ready for review.

rdstern (approved):
I am approving, because the only changes now are cosmetic, and I suggest there may be more later.  So, subject to @shadrackkibet or @lloyddewit it could be merged now.
a) Could change the label to simply Points (Optional):  
b) The label at the top could be changed to be Heat Map.  (It will soon change further!
c) In the special Options button, it goes to geom Tile.  So I would prefer to call it Tile Options here.

lloyddewit:
@shadrackkibet Thank you for the review you did earlier. If you're happy that all your comments are now resolved, then we can merge.

instat/dlgRugPlot.vb
```diff
@@ -289,4 +323,24 @@ Public Class dlgRugPlot
...
+    Private Sub ucrReceiverPoints_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverPoints.ControlValueChanged
+        If bRCodeSet Then
+            ucrNudShape.Visible = False
+            If ucrReceiverPoints.IsEmpty Then
+                clsBaseOperator.RemoveParameterByName("geom_point")
+            Else
+                If ucrReceiverPoints.strCurrDataType = "numeric" Then
+                    clsBaseOperator.AddParameter("geom_point", clsRFunctionParameter:=clsGeomPointSizeFunction, iPosition:=7)
+                ElseIf ucrReceiverPoints.strCurrDataType = "factor" Then
+                    ucrNudShape.Visible = True
```
shadrackkibet, line 336: @lloyddewit the linking system should be improved to allow the state of the receiver as a datatype. This feature will help in such a case where linking has to be done manually.

shadrackkibet (approved):

instat/dlgRugPlot.vb
shadrackkibet: Not done. But I will merge since this doesn't affect functionality.

shadrackkibet:
> I am approving, because the only changes now are cosmetic, and I suggest there may be more later. So, subject to @shadrackkibet or @lloyddewit it could be merged now.
> a) Could change the label to simply Points (Optional):
> b) The label at the top could be changed to be Heat Map. (It will soon change further!
> c) In the special Options button, it goes to geom Tile. So I would prefer to call it Tile Options here.

@EstherNjeriLiberatta could you work on these changes on a separate PR?

EstherNjeriLiberatta:
@shadrackkibet yes, I'll do that.

PR #6767, opened 2021-09-14
Added selection of columns functionality

### LINKED ISSUE #6426: Adding the select facility for variables as a partner to the filter system

shadrackkibet:
fixes #6426

This is not yet ready for review. I will send a notification when it is ready. 

Design screenshots below.

image
image
image
image

### DIFF
--- a/instat/dlgSelectColumns.vb
+++ b/instat/dlgSelectColumns.vb
@@ -19,7 +19,6 @@ Public Class dlgSelectColumns
     Private bReset As Boolean = True
     Private bFirstLoad As Boolean = True
     Private clsAddColumnSelection As RFunction
-    Private clsSetCurrentColumnSelection As RFunction
     Private clsConditionsList As RFunction
     Private clsFromToOperation As ROperator
     Private Sub dlgSelectColumns_Load(sender As Object, e As EventArgs) Handles MyBase.Load
@@ -82,7 +81,6 @@ Public Class dlgSelectColumns
 
     Private Sub SetDefaults()
         clsAddColumnSelection = New RFunction
-        clsSetCurrentColumnSelection = New RFunction
         clsConditionsList = New RFunction
         clsFromToOperation = New ROperator
 
@@ -90,7 +88,6 @@ Public Class dlgSelectColumns
         ucrInputSelectOperation.SetText("Columns")
 
         clsConditionsList.SetRCommand("list")
-        'clsConditionsList.SetAssignTo("conds")
 
         clsFromToOperation.SetOperation(":")
         clsFromToOperation.AddParameter("from", 1, iPosition:=0)
@@ -100,9 +97,6 @@ Public Class dlgSelectColumns
         clsAddColumnSelection.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$add_column_selection")
         clsAddColumnSelection.AddParameter("column_selection", clsRFunctionParameter:=clsConditionsList, iPosition:=2)
         clsAddColumnSelection.AddParameter("and_or", Chr(34) & "|" & Chr(34), iPosition:=3)
-
-        clsSetCurrentColumnSelection.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$set_current_column_selection")
-
     End Sub
 
     Private Sub SetRcodeForControls(bReset As Boolean)
@@ -116,20 +110,12 @@ Public Class dlgSelectColumns
     End Sub
 
     Private Sub cmdCombineWithAndOr_Click(sender As Object, e As EventArgs) Handles cmdCombineWithAndOr.Click
-        If cmdCombineWithAndOr.Text.Contains("All combined with |") Then
-            cmdCombineWithAndOr.Text = " All combined with &&"
-            clsAddColumnSelection.AddParameter("and_or", Chr(34) & "&" & Chr(34), iPosition:=3)
-        Else
-            cmdCombineWithAndOr.Text = "All combined with |"
+        If cmdCombineWithAndOr.Text.Contains("All combined with &") Then
+            cmdCombineWithAndOr.Text = " All combined with |"
             clsAddColumnSelection.AddParameter("and_or", Chr(34) & "|" & Chr(34), iPosition:=3)
-        End If
-    End Sub
-
-    Private Sub ucrSelectorForColumnSelection_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSelectorForColumnSelection.ControlValueChanged
-        If ucrSelectorForColumnSelection.ucrAvailableDataFrames.strCurrDataFrame <> "" Then
-            clsSetCurrentColumnSelection.AddParameter("data_name", Chr(34) & ucrSelectorForColumnSelection.ucrAvailableDataFrames.strCurrDataFrame & Chr(34), iPosition:=0)
         Else
-            clsSetCurrentColumnSelection.RemoveParameterByName("data_name")
+            cmdCombineWithAndOr.Text = "All combined with &&"
+            clsAddColumnSelection.AddParameter("and_or", Chr(34) & "&" & Chr(34), iPosition:=3)
         End If
     End Sub
 
@@ -137,7 +123,6 @@ Public Class dlgSelectColumns
         If clsAddColumnSelection IsNot Nothing Then
             If Not ucrInputSelectName.IsEmpty() Then
                 clsAddColumnSelection.AddParameter("name", Chr(34) & ucrInputSelectName.GetText() & Chr(34), iPosition:=1)
-                clsSetCurrentColumnSelection.AddParameter("name", Chr(34) & ucrInputSelectName.GetText() & Chr(34), iPosition:=1)
             Else
                 clsAddColumnSelection.RemoveParameterByName("name")
             End If
--- a/instat/dlgSelectColumns.vb
+++ b/instat/dlgSelectColumns.vb
@@ -49,10 +49,10 @@ Public Class dlgSelectColumns
         ucrInputSelectOperation.SetItems({"Columns", "Starts with", "Ends with", "Contains", "Matches", "Numeric range"})
         ucrInputSelectOperation.SetDropDownStyleAsNonEditable()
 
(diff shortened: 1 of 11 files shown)


rdstern (commented):
@shadrackkibet great this has started.  I couldn't see where to check?  I assume eventually on Data Options, and a replacement dialogue for Prepare > Data Frame > Hide/Show?

shadrackkibet:
Let us test it from my end at this stage. I will be in touch on skype.

shadrackkibet:
@rdstern hopefully you can test now.

rdstern:
This is great - and seems to work.  Happy to imporve if you think merging at this stage will help.
What's next for you (and for others if you are bringing them in.  Here are some points from this first look:
a) Could we make it easy to have all variables of a given type, e.g. all factors, all logical? 
b) Could it say how many variables are selected.  I tried some that didn't work and got none selected.  I wonder if there should be a warning in this case.
c) I wanted to select all that start with anything and then i in the second place.  I tried .i, but it didn't work.  Could there be a fixed, or regex option in the choice?
d) In the selector the top usually says Variables.  When there is a selection, could there be perhaps simply `(select1)` after this?
e) I look forward to being able to access these from the data options.  The tabs currently say Rows Columns .  I suggest `Rows (Filter) Columns (Select)`
Very exciting!

rdstern:
A job for @dannyparsons on the R side I assume.  I am keen to encourage more use of variable labels in our use of R.  Could @dannyparsons add the feature to be able to search in other attributes as well as the name.  Then, you add a group box in the sub-dialogue It has the label Select from:  Then there are 3 options inside.   Variable name,  Variable Label, Value Label.  Variable Name is ticked by default, and the others are not.

rdstern:
@shadrackkibet could you please also add the help context number 597.  Thanks

rdstern (approved):
Good to merge if possible now.  Then there are more dimensions of the general facility to add.  
@lloyddewit could you check?

instat/dlgSelectColumns.vb
```diff
@@ -1,3 +1,257 @@
...
+                strValue = clsFromToOperation.ToScript
+                'Case "Where"
+                'clsCurrentConditionList.AddParameter("operation", Chr(34) & "where" & Chr(34), iPosition:=0)
+                'If ucrInputColumnType.GetText = "Numeric" Then
+                'clsParametersList.AddParameter("fn", "is.numeric", iPosition:=0)
+                'ElseIf ucrInputColumnType.GetText = "Factor" Then
+                'clsParametersList.AddParameter("fn", "is.factor", iPosition:=0)
+                'ElseIf ucrInputColumnType.GetText = "Character" Then
+                'clsParametersList.AddParameter("fn", "is.character", iPosition:=0)
+                'End If
```
lloyddewit, line 196: Delete?

instat/dlgSelectColumns.vb
```diff
@@ -1,3 +1,257 @@
...
+        ucrInputSelectOperation.SetDropDownStyleAsNonEditable()
+
+        ucrInputColumnType.SetItems({"Numeric", "Factor", "Character"})
+        ucrInputColumnType.SetDropDownStyleAsNonEditable()
+
+        ucrInputSelectOperation.AddToLinkedControls(ucrChkIgnoreCase, {"Starts with", "Ends with", "Contains", "Matches"}, bNewLinkedHideIfParameterMissing:=True)
+        ucrInputSelectOperation.AddToLinkedControls(ucrReceiverMultipleVariables, {"Columns"}, bNewLinkedHideIfParameterMissing:=True)
+        ucrInputSelectOperation.AddToLinkedControls({ucrNudFrom, ucrNudTo}, {"Numeric range"}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
+        ucrInputSelectOperation.AddToLinkedControls(ucrInputText, {"Numeric range", "Starts with", "Ends with", "Contains", "Matches"}, bNewLinkedHideIfParameterMissing:=True)
+        'ucrInputSelectOperation.AddToLinkedControls(ucrInputColumnType, {"Where"}, bNewLinkedHideIfParameterMissing:=True)
```
lloyddewit, line 57: delete?

instat/dlgSelectColumns.vb
```diff
@@ -1,3 +1,257 @@
...
+                    bEnableOrDisable = False
+                End If
+            Case "Numeric range"
+                If ucrNudFrom.IsEmpty OrElse ucrNudTo.IsEmpty OrElse ucrInputText.IsEmpty Then
+                    bEnableOrDisable = False
+                End If
+            Case "Where"
+                '
+            Case "Last column"
+                '
```
lloyddewit, line 236: delete?

instat/dlgSelectColumns.vb
```diff
@@ -1,3 +1,257 @@
...
+        cmdAddCondition.Enabled = bEnableOrDisable
+    End Sub
+
+    Private Sub ucrReceiverMultipleVariables_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverMultipleVariables.ControlValueChanged, ucrInputSelectOperation.ControlValueChanged, ucrNudFrom.ControlValueChanged, ucrNudTo.ControlValueChanged
+        EnableDisableAddConditionButton()
+    End Sub
+
+    Private Sub ucrInputText_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrInputText.ControlContentsChanged
+        EnableDisableAddConditionButton()
+    End Sub
```
lloyddewit, line 247: combine these into a single handler?

lloyddewit (changes requested):
This looks good, I just had some small comments

instat/dlgSelectColumns.vb
shadrackkibet: I need it when we add functionality to select factors, characters etc columns. I will add a comment!

instat/dlgSelectColumns.vb
shadrackkibet: For the same reason as above

instat/dlgSelectColumns.vb
shadrackkibet: same case

instat/dlgSelectColumns.vb
shadrackkibet: agreed

lloyddewit (approved):

lloyddewit:
@rdstern Since your last approval, there are only small changes that should not affect functionality. Please could you approve again?
We can then merge, thanks

shadrackkibet:
@lloyddewit many thanks for the reviews.

shadrackkibet:
@rdstern?

rdstern (approved):
Great to merge, so I am approving.  There are many things to improve of course - and I just noticed another one - maybe two!
a) The default iin the sub-dialogue s to select the columns and that is fine.  There is still the Data Options button, and it is active.  I am not sure it should be there!
b) When I move to the Starts, etc, then the column selector remains visible and I like that.  I'd like even to be able to scroll there?  But we should disable the Add button to indicate that the list of variables is then just for viewing.

PR #6785, opened 2021-09-21
Added geom_text_wordcloud and geom_text_wordcloud_area geoms.

HawardKetoyoMsatsi:
@rdstern @africanmathsinitiative/developers 
This is ready for review.


rdstern (changes requested):
@HawardKetoyoMsatsi shouldn't the label be a character (possibly a factor variable.  Currently it is just a numeric variable.
I couldn't check fully, because the current master version has an odd problem that is also evident in your version.  This is nothing to do with your work.

HawardKetoyoMsatsi:
@shadrackkibet Ann wanted to use the geom_text_wordcloud and geom_text_wordcloud_area could you please have a quick look at this PR
 I found the aesthetic parameter called angle should be passed as a numeric value. However, all aesthetic parameters in R-Instar set up aesthetics to accept input from receivers only. Is there a way they can be set up the parameters to accept inputs from ucrNuds or text inputs? If not we can leave that to be passed b the dialog.

rdstern (changes requested):
@HawardKetoyoMsatsi now it accepts a factor, but not a text variable.  I am nit sure it makes sense with a numeric variable, but we can try.  So you might remove any restriction on the label?

HawardKetoyoMsatsi:
@rdstern I remoned thedatatype restriction on the label parameter. 
@shadrackkibet @anastasia-mbithe  In this PR angle parameter is restricted to a numeric variable however it can also accept a numeric value. All aes() parameters should be variables but these geom functions are a bit strange they take some aes() parameters either variables or values).

rdstern:
@HawardKetoyoMsatsi I can't see anything wrong, but it throws an exception, see below, with any example:  It also takes me to the same point when I type Library(ggwordcloud) into the script window.  Do you get that?  
image

HawardKetoyoMsatsi:
The functions geom_text_wordcloud() and geom_text_wordcloud_area() seem to give the above error even when used in the r script in a branch without the changes I made in this PR. Maybe there could be a bigger issue with these functions.

rdstern (approved):

instat/ucrGeom.vb
```diff
@@ -1875,6 +1877,81 @@ Public Class ucrGeom
...
+        clsgeom_text_wordcloud.AddLayerParameter("na.rm", "boolean", "FALSE")
+        clsgeom_text_wordcloud.AddLayerParameter("Show.legend", "boolean", "FALSE")
+        clsgeom_text_wordcloud.AddLayerParameter("inherit.aes", "boolean", "TRUE")
+        clsgeom_text_wordcloud.AddLayerParameter("show_boxes", "boolean", "FALSE")
+        clsgeom_text_wordcloud.AddLayerParameter("area_corr", "boolean", "FALSE")
+        'clsgeom_text_wordcloud.AddLayerParameter("rstep", "numeric", "0.01", lstParameterStrings:={2, 0, 1})
+        'clsgeom_text_wordcloud.AddLayerParameter("tstep", "numeric", "0.02", lstParameterStrings:={2, 0, 1})
+        'clsgeom_text_wordcloud.AddLayerParameter("perc_step", "numeric", "0.01", lstParameterStrings:={2, 0, 1})
+        'clsgeom_text_wordcloud.AddLayerParameter("max_steps", "numeric", "10", lstParameterStrings:={0, 0, 100})
+        'clsgeom_text_wordcloud.AddLayerParameter("area_corr", "boolean", "FALSE")
```
shadrackkibet, line 1915: I assume this is commented out because of the limitation on the number of controls we can show?

shadrackkibet (approved):

Issue #2761, opened 2017-03-09
New Menu Item: Describe > One Variable >Rating Data


rdstern:
This does not have a milestone, and might be done in April rather than for the current version.  Howver it may be very easy to do, because it follows the same format as the other graphs from the sjplot package.  

First you could look at what you are aiming for.  In the sjplot Reference Manual, page 61 there is a "See Also - sjPlot manual: sjp.likert and similarly on page 95 for sjp.stackfrq and page 146 for sjt.stackfrq.

1. The name can be "Rating Scales"
2. There is the usual selector and a multiple receiver.  The label for the receiver could be "Selected Ordered Factors".  If this is too long, then perhaps just "Ordered Factors".  This dialogue is only for Ordered Factors.
3. It is also for (Multiple) Ordered factors with at least all the same number of levels, and probably the same set of labels, e.g. a set of questions all with the same labels "Very Bad", "Bad", "Middling", "Good", "Very Good". (We need to think how to check this.  It relates to being able to easily distinguish different blocks of columns.)  We need to be able to check the number of levels at least and get access to this in the dialogue.
4. The next options are all similar to those from #2757 namely first :
It will generate either a frequency table, or a ggplot graph, or both. So there is the control to save a graph at the bottom of the dialogue.
5. There is also a checkbox (default unchecked) with the label Weights: If it is checked, then a single receiver is visible. The selector is now restricted to numeric columns.
6. There is a checkbox with label "Table". Default is checked.
7. There is a checkbox with label "Graph". Default is unchecked. (Don't worry now that both could be unchecked. If so, then for now we could still produce a ggplot graph object and not display it.
8. If checked then there is a choice (2 radio buttons) between "Stacked" and "Likert".
9. In item 8 if Likert is chosen and the number of levels is an odd number for the factor, then there is an Up-Down control visible with the label "Neutral Level".  It is between 1 and the number of levels with the default being (nlev+1)/2, i.e. the middle level.
9. There is a set of 5 radio buttons with the overall label "Sort". They have the 5 options "None, Low Ascending, Low Descending, High Ascending, High Descending". The default is "None".
10. There is a checkbox (as in the existing one variable graph, for "Flip Coordinates".
11. There is an Options button leading to a sub-dialogue. For now leave this blank.

stevenndungu:
@AlexSananka  how is this going?

AlexSananka:
@dannyparsons  was to work on it from somewhere.#2872

[...]

PR #6786, opened 2021-09-22
Updating R functions in the Rating Data Dialog 

Gioche6:
Fixes issue #2761 

@rdstern I have updated the functions as you advised, and changed parameter names according to the new updated package `sjPlot`. The dialog is now good and working. 

@Patowhiz , I have removed the unused panel as you have advised, Kindly check if its okay now. 

I'd however recommend several features for this dialog, like a combo of colors the user could choose from, and also probably a title option for both the graph and the table.

### DIFF
--- a/instat/dlgRatingScales.Designer.vb
+++ b/instat/dlgRatingScales.Designer.vb
@@ -19,7 +19,7 @@ Partial Class dlgRatingScales
     Inherits System.Windows.Forms.Form
 
     'Form overrides dispose to clean up the component list.
-    <System.Diagnostics.DebuggerNonUserCode()> _
+    <System.Diagnostics.DebuggerNonUserCode()>
     Protected Overrides Sub Dispose(ByVal disposing As Boolean)
         Try
             If disposing AndAlso components IsNot Nothing Then
@@ -36,7 +36,7 @@ Partial Class dlgRatingScales
     'NOTE: The following procedure is required by the Windows Form Designer
     'It can be modified using the Windows Form Designer.  
     'Do not modify it using the code editor.
-    <System.Diagnostics.DebuggerStepThrough()> _
+    <System.Diagnostics.DebuggerStepThrough()>
     Private Sub InitializeComponent()
         Dim resources As System.ComponentModel.ComponentResourceManager = New System.ComponentModel.ComponentResourceManager(GetType(dlgRatingScales))
         Me.lblOrderedFactor = New System.Windows.Forms.Label()
@@ -62,7 +62,6 @@ Partial Class dlgRatingScales
         Me.ucrReceiverOrderedFactors = New instat.ucrReceiverMultiple()
         Me.ucrBase = New instat.ucrButtons()
         Me.ucrSelectorRatingScale = New instat.ucrSelectorByDataFrameAddRemove()
-        Me.UcrDialogDisabled1 = New instat.ucrDialogDisabled()
         Me.grpSort.SuspendLayout()
         Me.grpLikertType.SuspendLayout()
         Me.SuspendLayout()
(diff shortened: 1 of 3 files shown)


rdstern:
@Gioche6 well done, it is working already!  I am now testing it with the efc data that is provided with the sjplot package.  here is their example and it works fine.
image

There are some aspects of these plots that I like as an example for our other examples of plotting.  For example it is automatically picking up the variable label.  And it is adding the sample size to the variable label.

I was almost ready to say let's merge and then add further options later.  But then I discovered some problems which you could perhaps examine.  I hope you can resolve them easily, i.e. without having to mess with the function in the sjplot package.

a) The 9 variables are numeric and labelled with the categoriesI tried making them into factors (and also ordered factors.  Then it gives an error.  We would often have them as factors when using R-Instat, so it should work.
b) I tried the 5 buttons you provided and couldn't see any difference in the resulting tables or plots.  What should happen with these options?

Gioche6:
@rdstern , I have been thinking and doing research on the thoughtful comments you gave and these are my findings. 

1. Item `(a)`. I tried developing the error you were having, and indeed i found it, however, As per the snips below, the error was resolved by making a few changes in the Convert Columns Dialog, option `Ignore labels` 
 

- **Case 1**. On this case, I left the Ignore labels unchecked , and working with the result in making the stacked and Likert graphs, an error of duplicated levels appeared, which is the same error you experienced . I tend to think that failing to ignore the labels makes them duplicated, and hence there's an error in the Rating Dialog when producing the graphs.  

snip1


- `Case 2 `On this case, I checked the ignore labels checkbox, and Tried using the results in  producing the graphs from the Rating Dialog which worked  fine. On the two cases, I worked with both convert to factor and convert to ordered factors. I would suggest you try converting the same, ignoring the  variable labels (clicking ignore labels) checkbox. I am excited to hear what you will find. 


snip 2

Gioche6:
@rdstern , On your item (c) comment,  the five buttons Indicates whether the items should be ordered by by highest count of first or last category of items.

"High Ascending"
to order ascending by lowest count of first category,

"High Descending"
to order descending by lowest count of first category,

"low ascending
to order ascending by lowest count of last category,

"low descending"
to order descending by lowest count of last category,

None
(default) for no sorting.

rdstern:
@Gioche6 I have no problem limiting the output (at this stage) where the function in the package has limitations. So forget the item c. except that we need to note (in the help) that variable labels need to be added for a good display.  This is awkward now, but it will soon be easy to paste into the variable label field - so copying the name into the label will become easy.

I tried with the "usual" survey data on Likert.  Not ideal, but I always like to try on "own data".  Here is a problem with the current dialogue:
image

In the Likert option the Number of categories checkbox keeps re-appearing whenever you reopen the dialogue.  Then it gives an error.  By the way, I always check dialogues by re-opening them to see if anything is different.  Good to fix this.

I would like to merge this dialogue as soon as possible, even if some options might be the subject of a new issue.  Then you can be moving on the the 2-way and 3-way frequency dialogues.  They are more important.

Gioche6:
@rdstern , this is still in progress. I would request you to test and review on the progress i have made in getting rid of the error produced when checking the Number of categories checkbox

I had a small challenge which I have managed to solve, that concerns the `nlevels` function. I noted that it is only applicable using only one Variable. and returns 0 whenever several multiple values are used. I think right now it works fine after adding several conditions in the code.

rdstern:
@Gioche6 thanks - I'll check

rdstern:
@Gioche6 I tried the default and then checked the sorting option.  These are the 5 radio buttons on the right of the dialogue.  It doesn't seem to work yet.  I replaced the  `sort.frq=NULL`, by (for example)  `sort.frq="neg.desc`" in the script window, which is all that is needed to get this component to work.  Could you add it to the code?

I used the efc data to test.  Are you testing with these same data?  I look forward to the next version.

Gioche6:
@rdstern , The five buttons for the `sort.frq` parameters works fine now. Could you test it?

I am using the `efc `data from the `sjPlot` package to do the testing. I believe right now the dialog is working fine.

Gioche6:
@rdstern , I have made the changes you requested. Kindly test to see if they are alright.

Additionally, the function does not allow a variable with two categories to have neutral levels, i have tested with the `e16sex` variable in the `efc` data which has two categories: male and female. I have however added a condition when working on a variable with two categories, you will not be able to check the Number of categories checkbox, therefore eliminating the function error that would appear.

lloyddewit:

Gioche6:
@lloyddewit  and @rdstern,  I have made minor changes on code implementation as requested by @N-thony . Could you test it and review. Thanks

lloyddewit:

PR #6803, opened 2021-09-27
Improved the Recode Factor Levels dialog in differents way with forcats functions

### LINKED ISSUE #6778: Improving the recode factor levels dialogue
This is a parallel exercise to the reorder levels dialogue.  It improves the current Prepare > Column: Factor > Recode Factor dialogue, in a similar way.
a) the currenct dialogue uses a plyr function.  Please check that it can be changed to use fct_recode from forcats instead.  I don't think this changes the dialogue.
b) At the top have 3 buttons.  Namely `Recode, Other, Lump`  The initial default is Recode and that gives the current dialogue.
c) The `Other` option will generate forcats::fct_other, so it generates the command:
`forcats::fct_other(f, keep, drop, other_level = "Other")`
e.g. forcats::fct_other(x, keep = c("A", "B"),other="other")
Note you give either keep or drop.
I suggest you use the control from the filter dialogue when a factor is chosen.  The code will also probably be useful too, to provide the list of levels selected.
There is a group box with label Selected Values.  Inside 2 radio buttons, first is Keep, second is Drop (i.e. Put into Other)
There is a control where you can type.  Default there is "Other".  If empty or Other, then default, otherwise it adds the other argument to the command.
d) the Lump option will generate one of the 4 forcat::fct_lump functions.  As with the others there is the selector and the songle receiver and the Save control.
Then possibly a group box with the label Keep.  Then 4 radio buttons inside
O Levels that appear at least this often.  Then up-down from 0 upwards - no limit - default perhaps 1 and in steps of 1.  You can type into it.
O Most (or least) common values.  This has an up-down into which you can type any integer.  Default could be 5 and there is no limit.  Negative values are allowed.
O Most (or least) frequent values Again an up-down into which you can type any values between -1 and +1.  It is an up-down with default 0.1 and going in steps of 0.01.  It can be negative. (command is forcats::fct_lumn_prop()
O Levels with more than Other  (command is forcats::fct_lump_lowfreq()

Add the Other control also as in the Other option!

N-thony:
Fixes #6778 

### DIFF
--- a/instat/dlgRecodeFactor.vb
+++ b/instat/dlgRecodeFactor.vb
@@ -22,6 +22,8 @@ Public Class dlgRecodeFactor
     Private clsFctRecodeFunction, clsFctOtherFunction As New RFunction
     Private clsFctLowFreqFunction, clsFctLumpPropFunction, clsFctLumpMinFunction, clsFctLumpnFunction As New RFunction
     Private clsCFunction As New RFunction
+    Private clsDummyFunction As New RFunction
+    Private clsOtherDummyFunction As New RFunction
     Private bReset As Boolean = True
     Private Sub dlgRecodeFactor_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
@@ -51,15 +53,18 @@ Public Class dlgRecodeFactor
         ucrPnlMethods.AddRadioButton(rdoKeep)
         ucrPnlMethods.AddRadioButton(rdoDrop)
 
+        ucrPnlMethods.AddParameterValuesCondition(rdoKeep, "checked", "keep")
+        ucrPnlMethods.AddParameterValuesCondition(rdoDrop, "checked", "drop")
+
         ucrPnlKeep.AddRadioButton(rdoLevels)
         ucrPnlKeep.AddRadioButton(rdoCommonValues)
         ucrPnlKeep.AddRadioButton(rdoFrequentValues)
         ucrPnlKeep.AddRadioButton(rdoMore)
 
-        ucrPnlKeep.AddFunctionNamesCondition(rdoLevels, "fct_lump_min")
-        ucrPnlKeep.AddFunctionNamesCondition(rdoCommonValues, "fct_lump_n")
-        ucrPnlKeep.AddFunctionNamesCondition(rdoFrequentValues, "fct_lump_prop")
-        ucrPnlKeep.AddFunctionNamesCondition(rdoMore, "fct_lump_lowfreq")
+        ucrPnlKeep.AddParameterValuesCondition(rdoLevels, "checked", "level")
+        ucrPnlKeep.AddParameterValuesCondition(rdoCommonValues, "checked", "lump_n")
+        ucrPnlKeep.AddParameterValuesCondition(rdoFrequentValues, "checked", "lump_prop")
+        ucrPnlKeep.AddParameterValuesCondition(rdoMore, "checked", "low_freq")
 
         ucrNudLevels.SetParameter(New RParameter("min", 1))
         ucrNudLevels.SetMinMax(0, Integer.MaxValue)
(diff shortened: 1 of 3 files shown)


rdstern:
@N-thony well done on the drop option.  This is already working well.
I checked the lump option with the (ggplot2) diamonds data where the clarity factor has 8 levels.
Could you label the first option 
`Levels with at least this frequency`

Also label the second option as 
`This number of common levels`

And the third as 
`Levels with at least this proportion`

I don't understand the current "Other" option.  Is this still to be done?  Did you look at the Filter sub-dialogue control for a factor, where you can choose a selection of levels?

N-thony:
@rdstern I have changed the control as we discussed under the Other option, please have a look. Thanks.

rdstern (dismissed):
@N-thony great - well done!  Over to @shadrackkibet or @lloyddewit for the next step

N-thony:
@lloyddewit over to you.

instat/dlgRecodeFactor.vb
```diff
@@ -19,8 +19,13 @@ Imports instat.Translations
 
 Public Class dlgRecodeFactor
     Private bFirstLoad As Boolean = True
-    Private clsReplaceFunction, clsRevalueFunction As RFunction
+    Private clsFctRecodeFunction, clsFctOtherFunction As New RFunction
+    Private clsFctLowFreqFunction, clsFctLumpPropFunction, clsFctLumpMinFunction, clsFctLumpnFunction As New RFunction
```
lloyddewit, line 23: Is this naming more consistent?
```suggestion
    Private clsFctLowFreqFunction, clsFctLumpPropFunction, clsFctLumpMinFunction, clsFctLumpNFunction As New RFunction

```

instat/dlgRecodeFactor.vb
```diff
@@ -51,41 +92,120 @@ Public Class dlgRecodeFactor
...
+            Else
+                ucrBase.OKEnabled(False)
+            End If
+        ElseIf rdoLump.Checked Then
+            If Not ucrReceiverFactor.IsEmpty AndAlso ucrSaveNewColumn.IsComplete Then
+                ucrBase.OKEnabled(True)
+            Else
+                ucrBase.OKEnabled(False)
+            End If
         End If
```
lloyddewit, line 209: Simplify `If` condition?
```suggestion
        If ucrReceiverFactor.IsEmpty OrElse Not ucrSaveNewColumn.IsComplete Then
            ucrBase.OKEnabled(False)
        ElseIf rdoRecode.Checked Then
            ucrBase.OKEnabled(ucrFactorGrid.IsColumnComplete("New Label"))
        ElseIf rdoOther.Checked Then
            ucrBase.OKEnabled(Not String.IsNullOrEmpty(ucrFactorLevels.GetSelectedLevels()))
        ElseIf rdoLump.Checked Then
            ucrBase.OKEnabled(False)
        End If
```

instat/dlgRecodeFactor.vb
```diff
@@ -104,29 +224,70 @@ Public Class dlgRecodeFactor
...
+                ElseIf rdoCommonValues.Checked Then
+                    clsDummyFunction.AddParameter("checked", "lumpn", iPosition:=0)
+                    ucrBase.clsRsyntax.SetBaseRFunction(clsFctLumpnFunction)
+                End If
+            End If
+        End If
+    End Sub
+
     Private Sub ucrReceiverFactor_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverFactor.ControlValueChanged
         DefaultNewName()
```
lloyddewit, line 287: I know that this is not your code, but ...
`DefaultNewName` is very small and only called here. Could we delete `DefaultNewName` and paste its code in here instead?
```suggestion
        If (Not ucrSaveNewColumn.bUserTyped) AndAlso (Not ucrReceiverFactor.IsEmpty) Then
            ucrSaveNewColumn.SetPrefix(ucrReceiverFactor.GetVariableNames(bWithQuotes:=False) & "_recoded")
        End If
```

lloyddewit (changes requested):
@N-thony Looks good, just a few small comments

N-thony:
@lloyddewit thanks for your review. This is ready for re-review.

rdstern (changes requested):
@N-thony the lump no longer allows ok.  Also it adds recorded to the existing name.  So I already had data1_recoded and it therefore made it data1_recoded_recoded.  I wonder what would have happened if I used the command 5 times?

N-thony:
> @N-thony the lump no longer allows ok. Also it adds recorded to the existing name. So I already had data1_recoded and it therefore made it data1_recoded_recoded. I wonder what would have happened if I used the command 5 times?

@rdstern I have corrected this and fixes some bugs. Please have a look.

rdstern (approved):
@N-thony that's better!

lloyddewit (approved):

Issue #6195, opened 2021-02-01
Add cluster analysis to R-Instat


rdstern:
This is one of the multivariate methods mentioned in WMO 100.  We already have the others, namely correlations, PCA, Canonical correlations, so this is the one tool missing.

It fits well, in that we are roughly limiting ourselves currently to descriptive methods.

There is a CRAN cluster analysis task view that lists 100 packages.  Two are distributed with base R, namely stats and cluster. 

There is a good case study [here](https://www.datanovia.com/en/blog/cluster-analysis-in-r-simplified-and-enhanced/).  It proposes the package factoextra be used.  Interestingly that isn't in the list in the task view packages.  But their arguments seem sound and we already have factoextra in R-Instat.  Also they use ggplot for the graphs - which of course we like!  (Aha the article is written by the author of factorextra, and he also has a book called "Practical Guide to Cluster Analysis in R"

As background, there are 2 distinct forms of clustering:
a) Hierarchical:  going up or down you could start with each observation as its own cluster, then put the 2 closest together.  Or start with everything in one big cluster and progressively move cases out.
b) Partitioning clustering where you specify the number of clusters, e.g. 4 and then see which cases to put in which cluster.

Here is an article:
![image](https://user-images.githubusercontent.com/11226469/106382886-59683300-63ba-11eb-97b1-4d2057d39b20.png)

If we proceed, then it would be an additional dialogue - or perhaps 2 dialogues - under Describe > Multivariate.

rdstern:
I think I now have the methods to use for clustering and it is simpler than I thought!  That's a relief.  I am now happy with the book called ["Practical Guide to Cluster Analysis in R: Unsupervised Machine Learning" ](https://xsliulab.github.io/Workshop/week10/r-cluster-book.pdf)

More than that we can follow the book - which like our others recently - can be read online freely.  And that's useful, because the code can be followed.

Cluster analysis needs the data to be "prepared".  That's fine we have a prepare menu.  So I suggest a preliminary - which follows the commands in the first chapters of the book is a new Transform dialogue.  This can go in the Prepare > Data Reshape menu.

I suggest 3 (no 2) buttons at the top, namely (Omit.na), Scale, Distance.  The selector, in each case, allows either a data frame or a set of variables from a data frame.  It returns a new matrix (from dist) and data frame, from scale.  I assume a matrix can be displayed as a new data frame and still be treated as a matrix if it qualifies?

[...]

PR #6812, opened 2021-09-30
Added a dialog to prepare data for cluster analysis

Vitalis95:
@rdstern @shadrackkibet, this is prepare dialog for cluster analysis, issue #6195. I have added two functions; scale() and na.omit(),
Please have a look at it.


shadrackkibet:
@HawardKetoyoMsatsi please test the functionality.

HawardKetoyoMsatsi:
Fine I will take a look.

rdstern (changes requested):
@Vitalis95 it says there are build errors.  This may be just me, but I can't test.  Perhaps you can fix, or I could also look at it working on your machine - shared perhaps with skype?

shadrackkibet:
@rdstern I suggest you wait for Haward to look into this first then you can test.

instat/dlgCluster.vb
```diff
@@ -1,3 +1,192 @@
...
+    Private Sub SetBaseFunction()
+        If rdoScaleData.Checked Then
+            If ucrChkOmitMissingRows.Checked Then
+                ucrBase.clsRsyntax.SetBaseRFunction(clsStatsNAOmitFunction)
+            Else
+                ucrBase.clsRsyntax.SetBaseRFunction(clsScaleFunction)
+            End If
+        Else
+
+        End If
```
HawardKetoyoMsatsi, line 143: Make this condition complete.

instat/dlgCluster.vb
```diff
@@ -1,3 +1,192 @@
...
+        If rdoScaleData.Checked Then
+            ucrSaveNewDataFrame.SetPrefix("scale")
+        Else
+            ucrSaveNewDataFrame.SetPrefix("distance")
+        End If
+    End Sub
+
+    Private Sub ucrCore_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverPrepareData.ControlContentsChanged, ucrSelectorPrepareData.ControlContentsChanged, ucrPnlSelectData.ControlContentsChanged
+        TestOkEnabled()
+    End Sub
```
HawardKetoyoMsatsi, line 178: This sub should also handle control contents change for rdoSelectedColumn.

HawardKetoyoMsatsi (changes requested):
1. I see frmMain changes in your PR kindly conform with @shadrackkibet whether you can commit frmMain main in this PR.
2. I see a problem with your frmMain resource file which is causing the build errors please take a look.

instat/dlgCluster.vb
Vitalis95: I will add the changes

Vitalis95: @HawardKetoyoMsatsi I think the panel controls this? Radio buttons doesn't have the control content changed property.

shadrackkibet:
@rdstern check on the functionality. 

We will sort out the changes on the main form later. Thanks

rdstern (changes requested):
@Vitalis95 it says there are build errors?  Sorry @shadrackkibet I don't know how to proceed further in such a case?

shadrackkibet:
Makes sense. Happy with your suggestions @lloyddewit

lloyddewit:
> 
> 
> @shadrackkibet @lloyddewit does it mean that I should close this PR?

As discussed on WhatsApp:

image

shadrackkibet:
@rdstern hold on a bit. We will sort this out soon. Thanks

Vitalis95:
@rdstern  have a look it now, I have added the dist function. The data I am using to test is from library (cluster: voters.repub). Thanks

shadrackkibet:
At the same time @Gioche6 could you test and review the code?

Gioche6:
@shadrackkibet , Sure. I'll be glad to to so

rdstern (dismissed):

lloyddewit (dismissed):
@Vitalis95 - Thank you for the changes, I'mm happy to approve.
@shadrackkibet - I saw that you also started to review. That's good, because you have good experience with the R-Instat dialog idioms and this will help Vitalis.
We can merge after you approve. Thanks

Vitalis95:
@shadrackkibet  have a look at it.

shadrackkibet (changes requested):
Not stable yet. From a user perspective, there are various issues which we need to sort out. Like when you use a data frame with non-numeric columns you get errors. Perhaps we should try to stop this? There are also reset issues for example the radio button selected doesn't reset to the default. Should we restrict selected variables to only numerics?

Vitalis95:
The receiver for the selected variables only allows numeric variables. I made the  default to be selection of variables as suggested by @rdstern but not the whole data frame  
The data frame should have numeric columns that is why  with non-numeric columns it gives an error. I will be later adding daisy function for non-numeric columns (factors) -  @shadrackkibet @rdstern

rdstern (approved):
Now seems to work.  Are there more checks or can this be merged?

lloyddewit:
> Now seems to work. Are there more checks or can this be merged?

@shadrackkibet Thank you for reviewing this PR earlier. You had some comments. Are these now resolved and may we merge?
Thanks

shadrackkibet:
> Not stable yet. From a user perspective, there are various issues which we need to sort out. Like when you use a data frame with non-numeric columns you get errors. Perhaps we should try to stop this? There are also reset issues for example the radio button selected doesn't reset to the default. Should we restrict selected variables to only numerics?

@Vitalis95 did anything change after this comment? Apologies if I missed out on something.

Vitalis95:
> > Not stable yet. From a user perspective, there are various issues which we need to sort out. Like when you use a data frame with non-numeric columns you get errors. Perhaps we should try to stop this? There are also reset issues for example the radio button selected doesn't reset to the default. Should we restrict selected variables to only numerics?
> 
> @Vitalis95 did anything change after this comment? Apologies if I missed out on something.

@shadrackkibet The receiver for the selected variables only allows numeric variables. I made the default to be selected variables as suggested by @rdstern but not the whole data frame
The whole data frame allows only numeric columns that is why with non-numeric columns it gives an error, @shadrackkibet you can guide me on this , to stop the error in case of a data frame with non-numeric columns.
Thanks.

shadrackkibet:
Hide non-numeric columns from the selector. To do this set the receiver to allow only numerics.  Pseudocode as follows `ucReceiverName.SetDataType("numeric")` also change the selector header to "Numerics" which should be something like `ucrReceivername.strSelectorHeader = "Numerics"`.

shadrackkibet (approved):
I approve and merge because what is here works already. Although there are still some issues, it qualifies as the initial implementation. I suggest we address them separately.

Issue #6102, opened 2020-11-23
Small changes in climatic boxplots


rdstern:
I have now been using this dialogue quite a lot in preparing a corresponding video.  It has held up quite well, but here are 3 small points:

a) When you first come into the dialogue it correctly completes the control, but not the within year one.  Usually this is the month control.  Could this be added when it exists.  The same lack occus when you press reset.

b) I am not sure what "Vertical x tick markers" is supposed to do?  Is it "Vertical x Labels".  If so, then could the label be changed.  And it doesn't seem to do anything yet.  (I can get vertical x labels from the sub-dialogue.)

c) With multiple stations I have been using the Station control set to `Facet Row`.  When I return to the dialogue it remains set, but if I go to the sub-dialogue and then return, it is still set on the control, but ignored when I press Ok.  If I press on the option, then it is Ok.

d) Could there be an extra up-down control on the right called `Outlier Coefficient`  This is the same as the Coeff value in the sub-dialogue.  It would be the same control as there, with an initial default of 1.5.

rdstern:
[ghana_2_stations_for_qcrain.zip](https://github.com/africanmathsinitiative/R-Instat/files/5648575/ghana_2_stations_for_qcrain.zip)

I found more oddities in the climatic boxplot code.  Maybe they can be fixed by @lilyclements, but I suspect it may need a look at the code by @dannyparsons to be efficient.

![image](https://user-images.githubusercontent.com/11226469/101277293-58949580-37ab-11eb-94f4-5a0eaaae317a.png)

This is the graph I wanted.  It was a boxplot for the rainfall, omitting the zero values.

But I started in the wrong data frame, i.e. the last data frame in the attached file that just had 2 rows of data in a single month.
I then moved to the correct data frame, but the dialogue found it diificult to forget the original data frame.

It did give me what I wanted on one occasion, but with the coefficient for the outlier set to 1.5.  I changed it to 3 and then found it had forgotten the condition of omitting values less than 0.85.  I could then no longer find a way (even pressing reset) to get it to include the condition.

I finally succeeded only by copying the closest code from the output window and then editing in the script window.


I suggest it might need a systematic examination of the code?

lilyclements:
[...]

PR #6813, opened 2021-09-30
Changes to the Climatic Boxplot Dialogue 

Ivanluv:
fixes issues Addressed in issue #6102 
This is still in progress


rdstern:
@Ivanluv this is looking much better now.  The really annoying feature where the rainfall limit was forgotten seems to be well-and-truly fixed.  Great.
The controls to remember the facetting don't seem (all?) to remember their previous settings.  
The figure below followed this graph - not a very nice one!

image

This had Station set to Facet wrap and the year set to be the x-axis.  When I return to the dialogue it is like this.  So with  both Station and syear1 set to "none".

image

PR #6814, opened 2021-10-01
Added functionality to rename multiple columns 

### LINKED ISSUE #6425: Add rename_with to be able to rename multiple columns
Have Single and Multiple as two buttons at the top of the Prepare > Data Frame > Rename Column dialogue.  In the multiple option implement the dplyr `rename_with` command.

 There are then 2 parts we need to specify:
a) The variables to select.  This could allow the usual data selector with the multiple receiver.  But it will also allow a Select object from the new Select dialogue.  It may only allow this.  We may want to complete that dialogue first.
b) The function to change the names.   This needs some thought and will probably be done after the new Prepare > Data Frame > Select dialogue

shadrackkibet:
fixes #6425

@rdstern I added the basic functionality to rename multiple columns. I will appreciate it if you could do some quick testing.


rdstern:
@shadrackkibet that's a great start.  
a) I did get the usual developer error going into the dialogue for the second time - only that once though? Perhaps it was linked to the system moving the current data frame?
b) Already a small cosmetic point, I think tidyverse promotes names in lower case?  If so, then change the order, so default is lower case, followed by the title.  This is already useful.
c) I didn't understand - and haven't checked - about "clean names".  Sounds good, and I need to look further.  I tried on some names that must already be "clean" so nothing changed.
d) My big surprise was that the results always go to a new data frame.  I hadn't even thought of that as an option!  To me the default is to tidy the data as it is, i.e. change the names in existing variables.  I am even nervous about making new variables with just the names changed.  So, if and when we start trying to implement an "undo" this will be a very strong candidate.  Now there's a challenge - could that be included as an option here?  
e) And I look forward to more options.  I see you are specifying a few examples of the function, so perhaps that will be easy.
f) Is there scope for perha[ps an Advanced button or a define option, where the user writes the function?

shadrackkibet:
@rdstern my response as follows;

a) Sorted out!
b) The tidyverse promotes `snake` case naming. If you would like this as the default then we should make "Clean names" the default. Actually, this option provides other options (upper,lower, sentence, title) in the drop down. I am for the opinion that we remove (upper,lower, sentence, title) radio buttons. 
c) An example that could be used to check this option is the `iris` dataset.
d) Yes we can make it an option. However, this is some bits of work. 
e) I do not intend to provide more options at this stage. I would like to try these options first before adding more with particular use cases.
f) We can provide a control to define the function. I wonder how useful it is. Is it not the reason why we have the script window?

shadrackkibet:
@rdstern I am also thinking about providing these options when importing data. The user can use the preview to check before importing. It saves the user a step of going to another dialog, reduces things to remember and hence improve the UI experience. What do you think?

rdstern:
@shadrackkibet useful replies.
b) and c) Yes I see tidyverse says:
"Variable and function names should use only lowercase letters, numbers, and _. Use underscores (_) (so called snake case) to separate words within a name."
So, I suggest as the drop down does everything, but we want to make life simple for the user that we currently have 2 radio buttons with Lower Case and Clean Names as the 2 current options.  And snake case is the default in the Clean Names.
d) There are multiple suggestions in d) so I am not sure what exactly is the "some bits of work".  To me the default is (like the simple rename) that the existing multiple variables are renamed.  I don't see why we introduce a new data frame!  
e) f) Ok  let's just this working now with the options you have.

shadrackkibet:
This is the sort of Danny's question. I will check with him.

By "some bits of work" I mean we need to make this an option within the rename function in the databook and therefore it "behaves" in a similar way as the simple rename.  

Another way we could achieve this is by providing an option in the save control to override an existing data frame.

shadrackkibet:
@rdstern this is now ready for your re-review. Thanks

shadrackkibet:
@lloyddewit sometimes the vb (name) i.e `dlg...` isn't consistent with the (text) of the dialog. In this case, I think we should use **Rename Column** rather than **Name** and therefore I altered the PR name. I suggest we always use dialog "text" in PR titles because that is what users see in the software. What do you think?

lloyddewit:
> 
> 
> @lloyddewit sometimes the vb (name) i.e `dlg...` isn't consistent with the (text) of the dialog. In this case, I think we should use **Rename Column** rather than **Name** and therefore I altered the PR name. I suggest we always use dialog "text" in PR titles because that is what users see in the software. What do you think?

Yes, that makes sense, thanks for improving the PR name

PR #6819, opened 2021-10-05
Changed the single receiver to multiple for the estimates on Taylor diagram

### LINKED ISSUE #6105: Multiple receiver for estimates on Taylor diagram
Taylor diagrams are quite popular for use in the inter-comparison analysis, I have seen it used in several papers. The function from openair package is flexible enough to allow multiple estimates at a time. I, therefore, suggest we change the single receiver "Estimate" into a multiple receiver because we often have several estimates to compare. This is similar to what I have suggested in #6100. @rdstern what do you think?
image

N-thony:
Fixes #6105 
@shadrackkibet this is ready for your review. Thanks.

### DIFF
--- a/instat/dlgTaylorDiagram.resx
+++ b/instat/dlgTaylorDiagram.resx
@@ -127,7 +127,7 @@
   </data>
   <assembly alias="System.Drawing" name="System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
   <data name="lblObserved.Location" type="System.Drawing.Point, System.Drawing">
-    <value>286, 42</value>
+    <value>286, 24</value>
   </data>
   <data name="lblObserved.Size" type="System.Drawing.Size, System.Drawing">
     <value>56, 13</value>
@@ -148,7 +148,7 @@
     <value>$this</value>
   </data>
   <data name="&gt;&gt;lblObserved.ZOrder" xml:space="preserve">
-    <value>8</value>
+    <value>9</value>
   </data>
   <data name="lblEstimated.AutoSize" type="System.Boolean, mscorlib">
     <value>True</value>
@@ -157,7 +157,7 @@
     <value>NoControl</value>
   </data>
   <data name="lblEstimated.Location" type="System.Drawing.Point, System.Drawing">
-    <value>286, 99</value>
+    <value>286, 72</value>
   </data>
   <data name="lblEstimated.Size" type="System.Drawing.Size, System.Drawing">
     <value>56, 13</value>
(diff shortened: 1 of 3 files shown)


instat/dlgTaylorDiagram.resx
```diff
@@ -478,6 +478,6 @@
     <value>$this</value>
   </data>
   <data name="&gt;&gt;ucrReceiverType.ZOrder" xml:space="preserve">
-    <value>3</value>
+    <value>4</value>
```
shadrackkibet, line 481: Sort out Tab order. I see a change here.

instat/dlgTaylorDiagram.vb
```diff
@@ -51,6 +51,7 @@ Public Class dlgTaylorDiagram
         ucrReceiverEstimated.SetParameterIsString()
         ucrReceiverEstimated.SetIncludedDataTypes({"numeric"})
         ucrReceiverEstimated.strSelectorHeading = "Numerics"
+        ucrReceiverEstimated.bExcludeFromSelector = True
```
shadrackkibet, line 54: This might incovenience users. Many users might not realise that you have to remove the variable from the receiver so that it is available to move to the other receiver. It might get users confused about what to do. I reserve this feature for instances when the dialog produces an error by using the same variable in multiple places. Otherwise, it might lead to confusion because users always see variables in the selector in many other places.

shadrackkibet (changes requested):

instat/dlgTaylorDiagram.resx
N-thony: I think the Tab order is okay/sorted already

shadrackkibet:
According to the documentation, it looks like we have to restrict the multiple receiver to only accept two variables. Any suggestions?

N-thony:
> Checking on the function documentation i.e `?openair::TaylorDiagram`

Yeah, I think you are right since the mod can be of length 2 only.

N-thony:
> According to the documentation, it looks like we have to restrict the multiple receiver to only accept two variables. Any suggestions?

@dannyparsons I was thinking in this kind of situation to have the possibility to set the maximum number of values that multiple receivers can receive. I mean to improve the multiple receivers with this restriction of the maximum number of value to receive if needed. What do you think?

lloyddewit:
@N-thony Could you look at the code and propose a way to set the maximum number of values that multiple receivers can receive? 
If you get stuck, then let me know
Thanks for trying this

dannyparsons:
@N-thony It should just be an extra few lines to remove the additional variables. So I suggest you do this now in this PR as it's much improved functionality for the user.

shadrackkibet:
Thanks for the improvement. I tried again but then I notice that although the content in the receiver is cleared when I click Ok I still see the variable in the output window. For example the following 
image
 Gives the following script 
```
# Code generated by the dialog, Taylor Diagram
Dagoretti_JKIA <- data_book$get_data_frame(data_name="Dagoretti_JKIA")
last_graph <- openair::TaylorDiagram(mydata=Dagoretti_JKIA, obs="PRECIP", mod=c("precip_CHIRP","rfe_TAMSAT","tp_ERA5"))
data_book$add_graph(graph_name="last_graph", graph=last_graph, data_name="Dagoretti_JKIA")
```

PR #6822, opened 2021-10-05
Changed the estimated single Receiver to Multiple on the Comparisons summaries Dialogue

### LINKED ISSUE #6100: Compare summaries dialog
I think it is much convenient to have "Estimated" receiver as a multiple receiver on this a dialog. We often have estimates from various sources which we want to compare with station observation. therefore changing the single receiver into multiple will make it convenient to obtain summaries rather than using the dialog multiple times. This is already possible via the summary function therefore it is easy to allow from the dialog. @rdstern @dannyparsons what do you think?
image

N-thony:
Fixes #6100
@shadrackkibet please have a look. Thanks.

### DIFF
--- a/instat/dlgCompareSummary.Designer.vb
+++ b/instat/dlgCompareSummary.Designer.vb
@@ -29,21 +29,21 @@ Partial Class dlgCompareSummary
         Me.rdoCategorical = New System.Windows.Forms.RadioButton()
         Me.rdoBinary = New System.Windows.Forms.RadioButton()
         Me.cmdSummaries = New System.Windows.Forms.Button()
+        Me.ucrReceiverSatellite = New instat.ucrReceiverMultiple()
+        Me.ucrChkStoreResults = New instat.ucrCheck()
         Me.ucrChkPrintOutput = New instat.ucrCheck()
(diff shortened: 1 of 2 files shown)


shadrackkibet:
@N-thony did you test this? I tried adding two or more into the new multiple receiver (Estimated) and got the following error immediately!

image

shadrackkibet:
Yes. 

**Solution:** `y` should be `columns_to_summarise` and vice versa! 

**Potential side effect:** Check that all verification summaries produced are still sensible and correct. Good to also check that there are no verification summaries that produce different results when you switch `x` and `y`.

N-thony:
> What exactly do you mean when you say the results produced are exactly the same? If you mean correlation = Mean error = MAE etc then that's is a big problem.
> 
> I suggest you find a minimum example (reprex) that shows that the results produced is still the same as what we had previously. Only that it has become efficient to do comparisons with multiple estimates.
Great, here is the comparisons of what we had previously and the new improvement with multiple estimates.
I used the experimental survey data frame

1. Here is the results previous with the single receiver
image
image
2. Here is the results with the multiple estimates
image

PR #6834, opened 2021-10-06
Implemented the option on Recod Factor dialogue for missing values

### LINKED ISSUE #6824: Recode factor levels for missing values
I wanted to recode the missing value, but couldn't.

I would like a checkbox that when checked would enable the recoding of the missing value with new level.

**Describe alternatives you've considered**
We considered using replace value which does not work easily for factors. We have created another issue for this.

**Additional context**

N-thony:
Fixes #6824 
@rdstern I have added the options for the record factor levels for missing values. Please have a look. Thanks.


rdstern (approved):
@N-thony that's great and seems to work fine.  Could you just add a default name, and I suggest the one proposed by forcats, namely (missing).
Then I hope we could merge

PR #6835, opened 2021-10-06
New branch

### LINKED ISSUE #6470: Two changes to the climatic menu on frm main

mmulahvictor:
Fixes #6470 
 -I added the Export to Climsoft to the Climatic> File> Export to Climsoft as shown below

Screenshot (19)
 -I also changed `ExportFromClimsoft` to `ExportToClimsoft` name and text.

@shadrackkibet @lloyddewit could you look at this.


rdstern:
I have yet to check, but the compare columns is just a repeat of a dialogue in the prepare menu, so should be a small task.  On the other hand, the export to climsoft needs a new dialogue to be constructed, so I am curious about progress.  I would not expect this to be done, without advice from @Patowhiz

PR #6838, opened 2021-10-07
Improved the Help feature in the Calculator dialog

### LINKED ISSUE #6796: Add Help context numbers plus other small changes to calculator
a) There are 6 keyboards in the calculator with their own help already.  Please add the context numbers for the other 5 keyboards, as follows:
Factor - 44
Transform 166
Wakefield 444
Circular 596
hydroGOF 598
b) Change Test to Text in the Text/String keyboard name
c) Check that currently there isn't a Sort function (just sorting a single variable - we have sort where you carry the whole data frame!  I can't find it!  So add Sort to the Transform keyboard.  Possibly make it the first key, so there are 5 keys (a bit narrower) on the first row.  It is the basic sort function from the base package.

@Wycklife this is similar to tasks in your previous "life".  Can you do this?  Or perhaps you could supervise one of the new interns?

Wycklife:
Fixes #6796

I have done the following;
1. Replaced the outdated help file with the updated version
2. Linked the updated help version to the main help button and added the context numbers for different keyboards.
3. Changed the text second help name from Help to R Help and linked it to R documentation for different keyboards.
4. Added an extra key (sort) to Transform keyboard
5.  Changes Test to Text in String keyboard

Still under discussion;
1. The position for the R Help button
2. Tooltips for the summary keyboard

@africanmathsinitiative/developers  This can be reviewed at this stage.

### DIFF
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -1319,7 +1319,7 @@ Public Class ucrCalculator
         clsHelp.AddParameter("help_type", Chr(34) & "html" & Chr(34))
         frmMain.clsRLink.RunScript(clsHelp.ToScript,
                                    strComment:="Opening help page for " &
-                                   strPackageName & " " & "Package. Generated from dialog Calculator",
+                                   strPackageName & " Package. Generated from dialog Calculator",
                                    iCallType:=2, bSeparateThread:=False, bUpdateGrids:=False)
     End Sub
 


Wycklife:
@rdstern Check if you are happy with the arrangement of the keys for the transform keyboard, also check the location for the Help button. Could you test this? Happy to make the any change you will need.

rdstern:
@Wycklife while you are there, 2 more small changes.
a) At the top add a colon to `Expression:`
b) Start with Save Result unticked.

Issue #6791, opened 2021-09-24
Implementing a Prepare > Column: Calculations > Transform dialogue


rdstern:
@rachelkg is now updating the help in R-Instat.  She typifies what we hope will be a common user of R-Instat.  She tells me that the Prepare > Calculator dialogue is pretty scary!

In the original Instat (long before R-Instat was born, we also had a calculator.  But we also had a Transform dialogue so users could do common data transformations without needing the general, but scary for some, calculator.

Here are some screen shots from that dialogue.  I suggest we produce an adapted version for R-Instat.
![image](https://user-images.githubusercontent.com/11226469/134626291-1b38592d-4ae7-4361-be36-0f9e4ebae5a5.png)

Here is a common transformation.  Notice the Inverse button so the inverse of log is exponential, etc.  Also a preview button to examine the formula!

In that dialogue, in the "old Instat", the "other Transformations button gives the following options:

![image](https://user-images.githubusercontent.com/11226469/134627644-b5ba76c7-72aa-4038-b49d-aca774d65966.png)

I suggest we need the equivalent in R-Instat.  I propose as follows:
![image](https://user-images.githubusercontent.com/11226469/134628083-2b762b62-5adc-4b16-ac59-2bb2080d1b20.png)

We use the existing Prepare > Column: Calculations > Ranks dialogue, which we rename to be Transform.  We might, at the same time, move it up one place, though I am not sure.  What is nice is that I propose a new name for the Menu item to be Column: Numeric!  That's neat, and important, because the 3 main types of data are Numeric, Factor, Character.  And that's what we now have in R-Instat!  

There will be quite a lot to do in this dialogue.  I suggest we merge in bits, as we go along.  I further suggest we have a new rule on the merging, i.e. we make everything visible as soon as it is ready, and have tooltips to mention when disabled items are not yet available.  That will make writing the help much easier.

The need for this dialogue came when writing the help for Sort.  We already have a general and powerful Sort (on right-click and in the Prepare > Data Frame menu).  But it sorts on multiple columns and carries the whole data frame.  How do we sort a single variable, leaving everything else undisturbed.  This should be in the calculator, but it is an omission there - that's another issue!  In the old Instat sort and Rank are both in this transform menu.

So, I propose the new Transform (from Rank) has the following buttons at the top.
Numeric, Non-negative, Sort, Rank.  
a) Rank is already there.
[...]

PR #6836, opened 2021-10-07
Updated 'Prepare>Column: Numeric' menu

mmulahvictor:
Partly fixes #6810
@africanmathsinitiative/developers  this is ready for review.

- I moved `Prepare> Column: Calculate> Duplicate Column` to `Prepare> Data Frame menu` below Rename Column.
- I changed the `Column: Calculate` menu name to `Column: Numeric`.
- I also changed the naming conventions of the items under Column: Numeric to follow the ColumnNumeric convention.
- I also changed the name of the `Prepare> Column: Numeric> Rank` To` Prepare> Column: Numeric> Transform`.

@shadrackkibet @lloyddewit @rdstern can you approve this?

PR #6840, opened 2021-10-08
Implemented the Prepare > Column: Calculate > Transform (Rank) dialog

anastasia-mbithe:
This fixes issue #6791 
@rdstern and @shadrackkibet I have added the Numeric, sort and Non- negative buttons in the Transform (Rank) dialog.
In Numeric, the scale radio button is not yet functional. Also, I have disabled the Non- negative for now. 
Am not sure of the prefix set in the save control for each function, you can suggest the Prefix to be used and any other change to be implemented.

I am yet to implement the naming to Transform dialogue.


mmulahvictor:
@anastasia-mbithe I have already made the changes on the Transform(Rank) dialog in PR #6836.

rdstern:
@anastasia-mbithe this is coming on well.
a) The sort now appears to work fine.
b) The numeric is now also working, though there are many cosmetic changes that will be useful
1) In Round you say `Round Of to:`  I suggest just `Round:`  I have chosen to keep the colon because you have to say to how many decimals.  Which you have as Digits: below.  I suggest changing Digits to Decimals for this option.
2) Also the up-down correctly has 0 as the default, but it can be a negative number.  (Then it gives to the nearest 10 or 100 etc.)
3) Significant Figures seems fine.  If you choose to keep the colon (:) in Round, then you need it here too.  I am happy with Digits as the labels> Make the default 4.  
4) Scale this is ok as a label.  No colon here, because this is complete as it is.  But you should change the name of the new variable
5) Diff doesn't work.  Have a look at what we have done in the calculator.  And notice that only works for first order differences.  So I would not bother with the second control.  Just have the Lag.  And notice that you will have to add in as many NA as you have for the value of the lag.  (Once it is working, then if you want to give 2nd order differences you just use the dialogue twice!

rdstern:
I think that the earlier code was probably closer to "proper" sorting, i.e. when the sorted variable carries the other variables in the data frame.  This was intended for the simple sort, i.e. where just that variable is sorted.  And it is sorted (usually) into a new variable 

As you raise the point, I wonder whether we also allow both options on this dialogue?  What we have should be the default, but we could have an option perhaps Carry the Whole Data Frame.  Then there is no new variable, so it is different!  And the general sort is for multiple variables.  I was wondering whether to give information in the dialogue to this effect?

anastasia-mbithe:
@rdstern and @shadrackkibet, the diff function works fine now. You can have a look at it.

shadrackkibet:
We need to discuss sort as well before inviting Roger to look into it.

PR #6841, opened 2021-10-08
Fixing the problem for a control hiding preview when loading multiple txt files

### LINKED ISSUE #6833: Problem with a control hiding preview when loading txt files
There is an issue with control visibility when importing multiple `txt` files. This should be easy to fix! See below.
image

Ivanluv:
fixes #6833

### DIFF
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -525,7 +525,6 @@ Public Class dlgImportDataset
     End Sub
 
     Public Sub SetDialogStateFromFile(strFileOrFolderPath As String, Optional strFolderFileExt As String = "")
-        strlastFileName = strFileName
         bImportFromFolder = False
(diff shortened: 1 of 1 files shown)


Ivanluv:
@shadrackkibet  from my understanding, when you have the `All files of this type ` checkbox checked you are referencing all the files within the current directory that's why the file name is not included as part of the directory. However, in doing this we loose the file name of the last selected file within the directory. That's why the checkbox gets unchecked when we check it again. This I have fixed by storing the last selected filename

shadrackkibet:
@EstherNjeriLiberatta could you please test and code review this?

EstherNjeriLiberatta:
@Ivanluv It's working well now. However, the issue with the checkbox not being able to check again seems to happen when I try importing a file that has a copy.

PR #6844, opened 2021-10-10
Fixed minor design bug on Describe>Specific>Histogram Plot Dialog

### LINKED ISSUE #6837: Small design bug on Describe>Specific>Histogram Plot Dialogue
There is a small design bug when changing the single to multiple as shown in the screenshot:
image

N-thony:
Fixes #6837 
@lloyddewit there is no coding in this change. Could it be merged? Thanks.


N-thony:
> @N-thony you say there was a small design bug in the histogram. But you don't say what the bug is? Is it that the Multiple Variables appear in Bold? If so, then it has yet to be fixed: image

It didn't display well like in the screenshot below
image

PR #6845, opened 2021-10-10
Added selection of columns into data options

### LINKED ISSUE #6263: Ways to handle wide data sets in R-Instat
A possible "solution" to this issue is here #4009 .  This adds the SPSS idea of sets of variables.

We have another interesting dataset to use as an example.  This is (or are) the MICS surveys. They are described here #6259. 

I also found an excellent discussion on [stackoverflow](https://stackoverflow.com/questions/9425194/what-is-the-most-efficient-way-to-select-a-set-of-variable-names-from-an-r-data) as usual!

I now feel perhaps we can make some progress in ways that are really easy.  My proposal (for discussion at this stage) is for a mixed strategy.  

a) We can already hide parts of a data frame.  There is work for 0.7.0 to slightly improve this facility.  
b) Some of the solutions suggested in the stack-overflow discussion involve swapping the data round, to be able to have one row (instead of a column) for a variable.  We already have this idea visually and can now work with the column metadata swapped with the data.
c) We do want to be able to select a whole set of variables easily, within a dialogue,  and should investigate how this can be improved.  That is discussed in the issue #5480 and perhaps we can make some progress for 0.7.0?  In 5480 @Patowhiz proposes to get rid of the data selector.  If we keep the data selector, then I would like to enhance the ways we can select multiple variables.  
d) In keeping the selector could we add a facility to autocomplete?
e) Where we may want to select lots of variables could we have a larger version of the selection control.  Then a combination of shift- click and ctrl-click that both work already, will be even simpler to use
f) In SPSS there are sets as discussed elsewhere.  Then selecting the sets of variables is like hiding variables very effectively.  That's like the filter for rows.  But SPSS only seems to handle one data frame at a time.  In R (and R-Instat we can easily make a new data frame with just a subset of the variables.  So I suggest we  below that we considerably extend the current Hide/Show dialogue and change its name to Select Variables.  (That will also justify, even more, its inclusion on the right-click menu!)

shadrackkibet:
fixes #6263

@rdstern this is ready for you to test. It is the implementation of the select columns feature as a tab under Data options in the selector. Please note this feature affect most dialogs therefore good to test it using different dialogs while making sure that there are no bugs introduced by these changes and there are no regressions.


rdstern:
@shadrackkibet this already seems to work.  I am ready to approve, except for a minor cosmetic suggestion.  So perhaps @lloyddewit could already examine the code, and then perhaps we can merge soon?
image

I would like to get rid of the word Column in certain places, above.  I suggest users will have to be shown that Filter is for Rows and Select is for columns.

So I would like the defailt name to be simply select, (and then select1, etc).  And it is clear on this tab that we are dealing with columns.  So the buttons can say Define New Selection and Remove Current Selection.

PR #6858, opened 2021-10-15
Added stem and leaf plot to Frequencies dialog

### LINKED ISSUE #6177: Adding a stem and leaf plot
This is a simple graph we don't yet have.  I did not think it was important, but it is mentioned in WMO 100.  And adding it is a simple deliverable for 2021.

There is a simple stem command in the base system and a more flexible one in aplpack package.  The latter includes back-to-back plots.  
They both produce results in the output window, which we may want?  But nicer to have as a graph.

[https://stackoverflow.com/questions/26532564/how-to-output-a-stem-and-leaf-plot-as-a-plot](url) shows how this can be done into a ggplot graph.  This could be in our Describe > One Variable > Graph perhaps as another graphics type.  Then perhaps the back-to-back as 2 variables.
As it is in ggplot, I am not sure how easily a factor could be added to have facets?  We should not spend too much time on it.

EstherNjeriLiberatta:
Fixes #6177 @rdstern @shadrackkibet Kindly have a look.

### DIFF
--- a/instat/dlgOneWayFrequencies.vb
+++ b/instat/dlgOneWayFrequencies.vb
@@ -261,7 +261,7 @@ Public Class dlgOneWayFrequencies
 
     Private Sub TestOkEnabled()
         If rdoStemAndLeaf.Checked Then
-            If Not ucrReceiverStemAndLeaf.IsEmpty OrElse ucrChkScale.Checked OrElse ucrChkWidth.Checked Then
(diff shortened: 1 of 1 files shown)


EstherNjeriLiberatta:
@rdstern @shadrackkibet @africanmathsinitiative/developers This is ready for review.

Ivanluv:
@shadrackkibet  this looks okay from my end

Issue #6534, opened 2021-06-15
Improve the find/replace dialogue in the prepare menu


rdstern:
I am using an example dataset from the dslabs package.  @HawardKetoyoMsatsi this will be added to R-Instat.  Meantime add it yourself, from Tools >Add package.

Then I am doing work described in chapter 24 of the [Introduction to Data Science ](https://rafalab.github.io/dsbook/string-processing.html)book. 
This uses the dataset called `reported_heights`

There are various problems with this dataset as explained in chapter 24.  Here is a screenshot of the variable first - the heights are a character variable.
![image](https://user-images.githubusercontent.com/11226469/122032711-cd363580-cdc7-11eb-9ed1-6315e5829396.png)

Selecting the non-numeric rows shows those problems:

![image](https://user-images.githubusercontent.com/11226469/122033057-18504880-cdc8-11eb-9039-fed4678fff54.png)

The data were supposed to be entered in inches, but some people have given feet and inches.
The way the inches are given is actually more complicated than is shown above, see below:

The double-quotes are sometimes given as 2 single quotes.
 I would like to replace the dougle quotes (and also the 2 single quotes) by ok.  Just to show I can do this in R-Instat.

Here is the dialogue - using \ as the escape character.

![image](https://user-images.githubusercontent.com/11226469/122034777-c3adcd00-cdc9-11eb-8c8a-349fe70c90c4.png)

It works!

![image](https://user-images.githubusercontent.com/11226469/122034551-8d704d80-cdc9-11eb-9d3a-519b480257ec.png)

But when I return to the dialogue the string has been changed.  It should not be.  That's the correction to make:

And please check the other options in this dialogue.  I think they will have the same problem.

I think the double quote is the same, but the single quote is not.
![image](https://user-images.githubusercontent.com/11226469/122035266-35861680-cdca-11eb-9a84-910399b4ea33.png)

HawardKetoyoMsatsi:
The dialog seems to remember what was input initially however there seems to be a problem with the way characters " . and ' are understood by the control. 
I tried to use . by replacing .. with xx 
![image](https://user-images.githubusercontent.com/16137042/123332623-67754680-d549-11eb-9047-02ad29a46e12.png)

look at what I get.
![image](https://user-images.githubusercontent.com/16137042/123332701-870c6f00-d549-11eb-9003-d69d264161b2.png)

Instead of looking for .. it takes any character it finds and replaces it with xx. 
[...]

PR #6859, opened 2021-10-16
Made literal expressions (rather than regex) the default in Find/Replace Dialog

N-thony:
Fixes partially #6534 
@rdstern I have disabled and implemented the regex on Find/Replace Dialogue in the way suggested in the issue that this PR refers to. Please take a look. Thanks.


lloyddewit:
@rdstern Since your last review, there are only small changes that don't affect the functionality. Please could you approve again?
thanks

PR #6866, opened 2021-10-19
Reorder data frames

### LINKED ISSUE #4557: Reordering data frames
This now works in reogrid - just the same as in Excel.  But I wonder if it works for us, i.e. in the Instat object? When you reorder using reogrid it is just in the front end.  Does it also reorder in the metadata?  

Is this feature going to be a blessing or a curse?

shadrackkibet:
fixes  #4557

@rdstern please test this once PR #6865 is merged.


shadrackkibet:
Please note, only column metadata do not update at the moment. I am investigating the problem.

shadrackkibet:
@dannyparsons @ChrisMarsh82 when I reorder data frames currently from the dialog it only works in the data frame metadata but not in data view and column metadata.  It used to work in data view and data frame metadata before the paging changes were merged.  Kindly look into this. Perhaps just some insights then I can try to fix it.

dannyparsons:
I hope this is just another symptom of the refreshing issues and this will also be taken care of, but good to check this once Chris resolves them.

dannyparsons:
@ChrisMarsh82 I think this part of the code below from GridLink may have got lost in the translation, which would detect if the order needed to be changed.
```vb
        If bFoundWorksheet Then
            fillWorkSheet = grdCurr.GetWorksheetByName(strName)
        Else
            fillWorkSheet = grdCurr.CreateWorksheet(strName)
            grdCurr.AddWorksheet(fillWorkSheet)
            iCurrPosition = grdCurr.Worksheets.Count - 1
        End If

        If iNewPosition <> -1 AndAlso iNewPosition <> iCurrPosition AndAlso iNewPosition < grdCurr.Worksheets.Count Then
            grdCurr.MoveWorksheet(fillWorkSheet, iNewPosition)
        End If
```

ChrisMarsh82:
@dannyparsons ok will add that in

dannyparsons:
I think this is easy to solve, but it would change a core part of the refreshing so suggest it's high risk just before the release.

@ChrisMarsh82 Can we discuss this early next week?

rdstern (changes requested):
It partially works:
a) In the column metadata window reorder is enabled and takes you correctly to the dialogue.  It is still disabled in the data window.
b) When the frames are reorded this seems to work in the data frame metadata, and the dialogue seems to remember the new order.
But the column metadata and data windows don't change the order.  I assume this might be a "refreshing" problem, but haven't found how to do that?

rdstern:
@shadrackkibet this has been quiet for almost 3 months.  It now has conflicts.  It would be useful, and I assume you will be getting back to it?  Is there a problem?

rdstern:
@shadrackkibet can this be closed now @N-thony is doing #7508?

shadrackkibet:
Yes, please.

PR #6878, opened 2021-10-22
Added Scale and Non-Negative buttons to the Transform dialog

anastasia-mbithe:
@rdstern  and @shadrackkibet  I have added Scale and Non negative buttons in the Transform dialog, This is still in progress, am yet to include the reverse function under the scale button.


rdstern:
@anastasia-mbithe and @shadrackkibet has anything changed recently on this Transform dialogue?

shadrackkibet:
Yes, I have some incomplete changes on my end which I will be pushing soon. Latest on Monday.

shadrackkibet:
>One attractive addition - if easy - could be a control with the R-command that is generated. This might be editable, or even (here) just for viewing. This would be instead of the Try field, that isn't needed in this dialogue - because (one the bugs are sorted) then it should generate an acceptable R command.

@rdstern is this a preview? like what we have in the models and filter?

rdstern:
exactly that

shadrackkibet:
@rdstern you can now test this.

rdstern:
This is looking really good now.  So just minor points, plus a question or two!
a) Allow negative integers in the round function.  For example -1 rounds to 10, -2 to nearest 100.  Could be useful occasionally - even for teaching.
b) Could the Preview control move up just a little.
c) The preview is great, but sometimes a bit daunting for the users of this dialogue.  So add a checkbox in front - default unchecked.
d) Now a question, on the name of the resulting variable.  It currently ignores the variable being transformed and is round, or round1.  I wonder if it should be yield1, yield2, etc.  Not sure, and can't have both, because I don't want names that become too long.
e) I really like the function for standardise.  Could we add an Omit NA

shadrackkibet:
@anastasia-mbithe I would like you to try and fix the following issues. I will be there to support you if needed. Just leave out (i) for now.

anastasia-mbithe:
@shadrackkibet okay.

anastasia-mbithe:
@rdstern At first I thought the Add operation was useful in getting the Inverse, but it seems like your suggestion is to add it in this already existing function. So, should the order of operations be like; (((yield-5)*1)/1)+3?

anastasia-mbithe:
@rdstern and @shadrackkibet I have managed to do all the changes apart from point(a). The last point, I have added the edit preview checkbox and it allows editing of the code in the preview, but it still runs the initial code not the edited(new) one.

anastasia-mbithe:
point a) is sorted out. Sorry I had misread the comment earlier.

rdstern:
It is looking reasonably good now:
a) The Preview setting checkbox doesn't remember when you open the dialogue again
b) As you say, the Edit doesn't yet work.  Let's just say Edit, rather than Edit Preview.
c) The Power we don't need to repeat the word Power.  And I am not a great fan of the up-down any more.  I suggest this be changed into just a control (like with the scale button) where you can type a number.  You could give suggestions in a pull-down e.g. 3, 2, 0.5, -1.  Possibly @dannyparsons won't like this?
d) Similarly for the constand.  And line the control with the others in the box.
e) With the Scale option when you type a number (instead of accepting one of the options provided) it makes it a string.  That produces an error of course.

dannyparsons:
c) I don't mind sensible common options as long as the user can enter their own too

anastasia-mbithe:
@rdstern and @shadrackkibet  Everything is sorted now apart from the edit preview function.

rdstern:
It is getting there, so the comments become ever more detailed.
a) On the scale option I like the way the preview updates as soon as you make a change and then click elsewhere.  But could you check whether you really need so many brackets?  You certainly need one set when you have subtract.  But do you need any more?  Suddenly having 3 lots of brackets is a bit overwhelming to our main audiance.
b) When I return to the dialogue - at least after the scale option, but perhaps generally, I get the developer error.
c) In the non-negative the preview doesn't seem to update when you type.  That's because it only updates when you click elsewhere, but here there isn't a good elsewhere.  Could you see if it is still ok if you update the preview after any character typed - might be a mess - .  If not, then perhaps you need an update button before the edit button?
d) And it will be great when the edit also works.

anastasia-mbithe:
@rdstern Kindly have a look at this. The edit preview is functional now.

rdstern (approved):
@anastasia-mbithe that's great.  I am happy to merge and this is an important improvement in R-Instat!  Over to @shadrackkibet or @lloyddewit for a second opinion.
@shadrackkibet I am building up to more suggestions on changes to frmmain and a relatively minor one is that I suggest this dialogue is more important than the row statistics - so should be promoted and swap places in the menu.

shadrackkibet:
> @shadrackkibet I am building up to more suggestions on changes to frmmain and a relatively minor one is that I suggest this dialogue is more important than the row statistics - so should be promoted and swap places in the menu.

I agree!

I was pondering about promoting some tidyverse verbs here although we don't use tidyverse packages in this dialog. So we call the dialog **Transform (Mutate)** and also in the frmmain. What do you think?

The tidyverse description of mutate is as follows;

> Description
mutate() adds new variables and preserves existing ones; transmute() adds new variables and drops existing ones. New variables overwrite existing variables of the same name. Variables can be removed by setting their value to NULL.

rdstern:
@shadrackkibet I rather like that idea of Mutate and then (also in the help we are writing) giving the idea from tidyverse.
By the way, on this dialogue I am keen to get it merged soon.  But there is then one (small?) improvement I would like to give in a new issue.  When you use Edit, it now works, but doesn't remember the edited calculation.  I am ok with that.  But the calculator and some others, like Enter, also keep information on past calculations as a pull-down.  It would be good to add that feature here.

lloyddewit:

N-thony:
> @N-thony Please could you review? thanks

@lloyddewit happy to review.

instat/dlgTransform.vb
```diff
@@ -161,9 +232,69 @@ Public Class dlgTransform
...
         ucrChkMissingLast.SetParameter(New RParameter("na.last"))
         ucrChkMissingLast.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
+        ucrChkMissingLast.SetRDefault("FALSE")
+
+        ucrInputSubtract.SetParameter(New RParameter("u", 1))
+        dctSubtractValues.Add("0", "0")
+        dctSubtractValues.Add("32", "32")
+        dctSubtractValues.Add("1970", "1970")
+        'dctSubtractValues.Add("Mean", "mean")
+        'dctSubtractValues.Add("Min", "min")
```
N-thony, line 242: Remove?

instat/dlgTransform.vb
```diff
@@ -161,9 +232,69 @@ Public Class dlgTransform
...
+
+        ucrInputDivide.SetParameter(New RParameter("z", 1))
+        dctDivideValues.Add("1", "1")
+        dctDivideValues.Add("5", "5")
+        dctDivideValues.Add("9", "9")
+        dctDivideValues.Add("25.4", "25.4")
+        dctDivideValues.Add("365.25", "365.25")
+        dctDivideValues.Add("1440", "1440")
+        'dctDivideValues.Add("sd", "sd")
+        'dctDivideValues.Add("range", "range")
```
N-thony, line 269: Even here?

N-thony (changes requested):
The update button seems to not work yet, maybe it should be hidden
image

N-thony (commented):
The preview input control seems to not remember the state on reopen after editing

anastasia-mbithe:
> 

The Update button is there for someone to click if you changed anything manually; especially in the textboxes, so that it can update in the Preview.

N-thony:
> > 
> 
> The Update button is there for someone to click if you changed anything manually; especially in the textboxes, so that it can update in the Preview.

Yes, but when I click nothing is happening. Can you maybe set it disabled?

N-thony:
@lloyddewit it is okay on my side, over to you.

instat/dlgTransform.vb
```diff
@@ -263,60 +461,182 @@ Public Class dlgTransform
...
+                clsConstantDummyFunction.AddParameter("preview", "FALSE", iPosition:=1)
+                If ucrChkEditPreview.Checked Then
+                    ucrInputPreview.IsReadOnly = False
+                Else
+                    ucrInputPreview.IsReadOnly = True
+                End If
+            Else
+                clsConstantDummyFunction.AddParameter("preview", "TRUE", iPosition:=1)
+            End If
+        End If
```
lloyddewit, line 623: can we make more readable?
```suggestion
        If Not bResetRCode Then
            Exit Sub
        End If

        If ucrChkPreview.Checked Then
            clsConstantDummyFunction.AddParameter("preview", "FALSE", iPosition:=1)
            ucrInputPreview.IsReadOnly = Not ucrChkEditPreview.Checked
        Else
            clsConstantDummyFunction.AddParameter("preview", "TRUE", iPosition:=1)
        End If
```

lloyddewit (changes requested):
@anastasia-mbithe This is clean, tidy and readable code, thanks.
I just had a couple of small comments

anastasia-mbithe:
@lloyddewit , Thank you for the suggestion, its simpler and more readable.

dannyparsons:
A general point, maybe not best on this PR but I see it was mentioned above.

I'm a bit concerned about over complicating the menu item names and dialog names. I've noticed that the menu items are getting longer and longer and they are starting to get overwhelming. I think we should try to use the minimal amount of text that clearly describes something, 1. because we are often working with beginners to (statistical) software and 2. some users do not have first language English so less is often better.

I think we could be in danger of getting too carried away with using all tidyverse terms. Sometimes they useful and have given us more clear terms, but sometimes more they are more technical than our own language which is simpler, and we don't need to be consistent with them.

So, for example, I don't think adding **Mutate** to **Transform** is necessary for clarity and in some cases might cause more confusion - e.g. "I think I know what Transform is but not Mutate too, so maybe I don't know what I'm doing".

Whereas on other dialogs like **Stack** the additional terms are useful because **Stack** isn't a widely used term and the other terms may help users who know it by another name.

I think it's great that we do have consistency with tidyverse ideas and the place for pointing this out and describing it is in the help file for users who want to know this in more detail.

Feel free to ignore for this PR and we can discuss in more details somewhere else.

lloyddewit (approved):

lloyddewit:
@rdstern There are some small changes since your last approval, please could you test again? thanks

rdstern (approved):
Looks good now, both in itself and as a teaching tool.

PR #6888, opened 2021-10-29
Improved the Spells Dialogue following ICPAC Dialogue

### LINKED ISSUE #5575: Improving the spells dialogue following comments from ICPAC
We currently have a spells dialogue, though some find it a bit difficult to use.  In the recent ICPAC training they currently use spells in a different way - a way that I don't like much.  But they asked still for it to be included, as that is the way they would like to process spells.  I can also see an adjustment to their proposed method that could be very useful.
Our current spells analysis takes a period in the year and finds the length of the longest (dry, wet, hot, etc) spell in that period.  For example it calculates the longest dry spell length each year between the date of the start and the end.
The additional option requested is the number of dry spells longer than a specified length.  For example **how many** dry spell in Jan-March period longer than 7 days.  I have never liked this because a very long dry spell in a particular year (say from 3rd Jan to 25 March is just one spell, but it doesn't leave any time for more spells.  So the answer in that year is 1.  While another year could have 5 dry spells of 8 days and sounds much worse.
However, I agree that if this is what a regional centre would like to produce, then we should include it.  Also if we want to critique it, then it is easier if we can produce it and then discuss the use of the results.
And I can also consider an adaptation of the idea that would be useful.  Instead of just giving the number of spells we could also include the number of days in the period belonging to a spell of longer than 7 days.  So, then this would be many days in the first case and just a total of 5 days in the second.  The 2 together could be a very useful summary.

N-thony:
Fixes #5575
I have made the improvement of the Spells Dialogue as requested in the corresponding issue. It doesn't yet with the Filter and it seems to be a general issue with calculations.

### DIFF
--- a/instat/dlgSpells.Designer.vb
+++ b/instat/dlgSpells.Designer.vb
@@ -44,19 +44,21 @@ Partial Class dlgSpells
         Me.lblDOY = New System.Windows.Forms.Label()
         Me.lblAnd = New System.Windows.Forms.Label()
         Me.grpRainParameters = New System.Windows.Forms.GroupBox()
-        Me.ucrInputFilterPreview = New instat.ucrInputTextBox()
         Me.cmdDoyRange = New System.Windows.Forms.Button()
-        Me.ucrInputSpellLower = New instat.ucrInputTextBox()
-        Me.ucrInputCondition = New instat.ucrInputComboBox()
-        Me.ucrInputSpellUpper = New instat.ucrInputTextBox()
         Me.lblStation = New System.Windows.Forms.Label()
         Me.lblNewColumnName = New System.Windows.Forms.Label()
+        Me.rdoAnnuel = New System.Windows.Forms.RadioButton()
+        Me.rdoSpells = New System.Windows.Forms.RadioButton()
+        Me.ucrPnlOptions = New instat.UcrPanel()
         Me.ucrInputNewColumnName = New instat.ucrInputTextBox()
-        Me.ucrChkConditional = New instat.ucrCheck()
         Me.ucrReceiverStation = New instat.ucrReceiverSingle()
         Me.ucrReceiverDOY = New instat.ucrReceiverSingle()
         Me.ucrReceiverYear = New instat.ucrReceiverSingle()
         Me.ucrReceiverDate = New instat.ucrReceiverSingle()
+        Me.ucrInputFilterPreview = New instat.ucrInputTextBox()
+        Me.ucrInputSpellLower = New instat.ucrInputTextBox()
+        Me.ucrInputCondition = New instat.ucrInputComboBox()
+        Me.ucrInputSpellUpper = New instat.ucrInputTextBox()
         Me.ucrReceiverElement = New instat.ucrReceiverSingle()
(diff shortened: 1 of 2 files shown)


lloyddewit:
@N-thony I assumed that this PR was ready for test. If I misunderstood, then let me know
Thanks!

rdstern (changes requested):
@N-thony I don't think this is yet ready.  But I noticed two things you may be able to fix, while you wait for @shadrackkibet to do the full code.
a) It produces a logical variable, which is TRUE each spell.  I suggest that can be deleted.
b) This may use the sub-dialogue for the dates, before you fixed it.  But when I look there it always says Jan1 to Jan 31 - rather than Dec 31 at the start.

N-thony:
> @N-thony I don't think this is yet ready. But I noticed two things you may be able to fix, while you wait for @shadrackkibet to do the full code. a) It produces a logical variable, which is TRUE each spell. I suggest that can be deleted. b) This may use the sub-dialogue for the dates, before you fixed it. But when I look there it always says Jan1 to Jan 31 - rather than Dec 31 at the start.

N-thony:
> @N-thony I don't think this is yet ready. But I noticed two things you may be able to fix, while you wait for @shadrackkibet to do the full code. a) It produces a logical variable, which is TRUE each spell. I suggest that can be deleted. b) This may use the sub-dialogue for the dates, before you fixed it. But when I look there it always says Jan1 to Jan 31 - rather than Dec 31 at the start.

@rdstern  for a) I'm going to look into it, but b) I think I have already fixed it, this branch doesn't have that change.

N-thony:
@rdstern I did a check for a) and also discussed with @shadrackkibet about the possibility to get rid of this extra variable. From @shadrackkibet's view that requires a change in the calculation which we need to check with @volloholic or @dannyparsons.

dannyparsons:
It could be done by doing the "rain day" calculation within the spells calculation instead of separately i.e.

`.spells(precip >= 0 & precip <= 0.85)`

Also, this shouldn't be called a "rain_day" because it could be any event. "spell_day" or "event" would be better. But this will be irrelevant if we don't save the column.

N-thony:
> It could be done by doing the "rain day" calculation within the spells calculation instead of separately i.e.
> 
> `.spells(precip >= 0 & precip <= 0.85)`
> 
> Also, this shouldn't be called a "rain_day" because it could be any event. "spell_day" or "event" would be better. But this will be irrelevant if we don't save the column.

@dannyparsons thanks, I have added the spell day calculation into the spells calculation.
@dannyparsons @rdstern please take a look.

rdstern (approved):
@dannyparsons this is a very useful new feature. I am approving, because I can live with it now.  But it ignores filters completely.  This includes ignoring the selection of part of the year, for the results, or of the selection of (say) some of the years.  This isn't a problem, because either filter can be applied to the resulting data frame.  But perhaps (in a new issue?) disable the control in the dialogue, or apply the filter (or filters) at the end - again later.

dannyparsons (approved):

dannyparsons:
@N-thony Could you create a new issue with the remaining issues Roger has outlined?

N-thony:
> @N-thony Could you create a new issue with the remaining issues Roger has outlined?

@dannyparsons sure.

PR #6892, opened 2021-11-01
Corrected and improved the Prepare > Column: Text > Split dialog

### LINKED ISSUE #6889: Correct and improve the Prepare > Column: Text > Split dialogue
There are a few minor problems with this dialogue - which I think cannot have been tested/used seriously so far.
a) It uses regex by default.  So the same control needs to be added as was used recently by @N-thony for the find/replace dialogues.  That is for us to use coll() as the default, so we opt in for regex, rather than the opposite.  Include the disabled option (for now) for the regex keyboard, which will need to be added more generally.
b) The pull-down gives various options to be added.  Please confirm that the implementation works.  When it says Underscore (_) then only the underscore, not the whole phrase is added?  I think it is Ok, but please check.
c) When I closed a data frame and then used a different one, it complained and clearly had not cleared the reference to the previous data frame.
d) When you change to have a specified number of splits, the command changes to use str_split_fixed.  I don't think that needs to be the case.  I suggest str_split can still be used.

N-thony:
Fixes #6889 

### DIFF
--- a/instat/dlgSplitText.vb
+++ b/instat/dlgSplitText.vb
@@ -50,7 +50,7 @@ Public Class dlgSplitText
         ucrPnlTextComponents.AddRadioButton(rdoMaximumNumberOfComponents)
 
         ucrPnlTextComponents.AddFunctionNamesCondition(rdoMaximumNumberOfComponents, "str_split")
-        ucrPnlTextComponents.AddFunctionNamesCondition(rdoFixedNumberOfComponents, "str_split")
+        ucrPnlTextComponents.AddFunctionNamesCondition(rdoFixedNumberOfComponents, "str_split_fixed")
 
         ucrPnlTextComponents.AddToLinkedControls(ucrNudPieces, {rdoFixedNumberOfComponents}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, objNewDefaultState:=2, bNewLinkedChangeParameterValue:=True)
         ucrPnlSplitText.AddToLinkedControls(ucrSaveColumn, {rdoTextComponents}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
@@ -110,7 +110,7 @@ Public Class dlgSplitText
         clsBinaryColumns.SetRCommand("multi.split")
 
         clsTextComponentsFixed.SetPackageName("stringr")
-        clsTextComponentsFixed.SetRCommand("str_split")
+        clsTextComponentsFixed.SetRCommand("str_split_fixed")
         clsTextComponentsFixed.AddParameter("pattern", Chr(34) & "," & Chr(34), iPosition:=1)
         clsTextComponentsFixed.AddParameter("n", strParameterValue:=2, iPosition:=2)
 
--- a/instat/dlgSplitText.vb
+++ b/instat/dlgSplitText.vb
@@ -64,17 +64,7 @@ Public Class dlgSplitText
         ucrReceiverSplitTextColumn.SetMeAsReceiver()
         ucrReceiverSplitTextColumn.bUseFilteredData = False
 
-        ucrInputPattern.SetParameter(New RParameter("pattern", 1))
-        dctPatternPairs.Add("Space ( )", Chr(34) & " " & Chr(34))
-        dctPatternPairs.Add("Period .", Chr(34) & "." & Chr(34))
-        dctPatternPairs.Add("Comma ,", Chr(34) & "," & Chr(34))
-        dctPatternPairs.Add("Colon :", Chr(34) & ":" & Chr(34))
-        dctPatternPairs.Add("Semicolon ;", Chr(34) & ";" & Chr(34))
-        dctPatternPairs.Add("Hyphen -", Chr(34) & "-" & Chr(34))
-        dctPatternPairs.Add("Underscore _", Chr(34) & "_" & Chr(34))
-        ucrInputPattern.SetItems(dctPatternPairs)
-        'ucrInputPattern.SetRDefault(Chr(34) & " " & Chr(34)) 'This is the default for clsTextComponents
-        ucrInputPattern.bAllowNonConditionValues = True
+        ucrInputPattern.SetItems({"Space ( )", "Period .", "Comma ,", "Colon :", "Semicolon ;", "Hyphen -", "Underscore _"})
 
         ucrNudPieces.SetParameter(New RParameter("n", 2))
         ucrNudPieces.SetMinMax(2, Integer.MaxValue)
@@ -105,6 +95,7 @@ Public Class dlgSplitText
         clsSplitDummyFunction = New RFunction
 
         ucrSelectorSplitTextColumn.Reset()
+        ucrInputPattern.cboInput.SelectedIndex = 2
 
         clsStringCollFunction.SetPackageName("stringr")
         clsStringCollFunction.SetRCommand("coll")
@@ -133,11 +124,8 @@ Public Class dlgSplitText
     Private Sub SetRCodeForControls(bReset As Boolean)
         ucrReceiverSplitTextColumn.AddAdditionalCodeParameterPair(clsTextComponentsMaximum, New RParameter("string", 0), iAdditionalPairNo:=1)
         ucrReceiverSplitTextColumn.AddAdditionalCodeParameterPair(clsBinaryColumns, New RParameter("var", 0), iAdditionalPairNo:=2)
-        ucrInputPattern.AddAdditionalCodeParameterPair(clsTextComponentsMaximum, New RParameter("pattern", 1), iAdditionalPairNo:=1)
-        ucrInputPattern.AddAdditionalCodeParameterPair(clsBinaryColumns, New RParameter("split.char", 1), iAdditionalPairNo:=2)
         ucrSaveColumn.AddAdditionalRCode(clsTextComponentsMaximum, bReset)
         ucrReceiverSplitTextColumn.SetRCode(clsTextComponentsFixed, bReset)
-        ucrInputPattern.SetRCode(clsTextComponentsFixed, bReset)
         ucrNudPieces.SetRCode(clsTextComponentsFixed, bReset)
         ucrChkIncludeRegularExpressions.SetRCode(clsSplitDummyFunction, bReset)
         ucrPnlSplitText.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
@@ -191,6 +179,7 @@ Public Class dlgSplitText
                 strPattern = Chr(34) & "_" & Chr(34)
         End Select
         clsStringCollFunction.AddParameter("pattern", strPattern, iPosition:=0)
+        clsBinaryColumns.AddParameter("split.char", strPattern, iPosition:=1)
         If ucrChkIncludeRegularExpressions.Checked Then
             clsTextComponentsFixed.AddParameter("pattern", strPattern, iPosition:=1)
             clsTextComponentsMaximum.AddParameter("pattern", strPattern, iPosition:=1)
--- a/instat/dlgSplitText.vb
+++ b/instat/dlgSplitText.vb
@@ -10,10 +10,9 @@
 ' but WITHOUT ANY WARRANTY; without even the implied warranty of
 ' MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 ' GNU General Public License for more details.
+'
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
-
 
 Imports instat.Translations
 Public Class dlgSplitText
--- a/instat/dlgSplitText.vb
+++ b/instat/dlgSplitText.vb
@@ -67,7 +67,7 @@ Public Class dlgSplitText
 
         ucrInputPattern.SetParameter(New RParameter("pattern", 1))
         dctPatternPairs.Add("Space ( )", Chr(34) & " " & Chr(34))
-        dctPatternPairs.Add("Period .", "stringr::fixed(" & Chr(34) & "." & Chr(34) & ")")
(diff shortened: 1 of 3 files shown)


rdstern (changes requested):
@N-thony sorry I was wrong, you do need the str_split_fixed, when you fix the number of splits.  Otherwise you get a single variable with a c(  ) when there is more than one split.
I notice when you use the option for any number of splits it always puts the result at the end - and ignores the position setting.  I don't know if there is anything you can do about this?  If not, then maybe the Position button should be disabled or hidden for that option?

N-thony:
> @N-thony sorry I was wrong, you do need the str_split_fixed, when you fix the number of splits. Otherwise you get a single variable with a c( ) when there is more than one split. I notice when you use the option for any number of splits it always puts the result at the end - and ignores the position setting. I don't know if there is anything you can do about this? If not, then maybe the Position button should be disabled or hidden for that option?

Done.

rdstern:
Great, the main one is done.  Just checking on the other point, namely when getting Maximum number of components they are all at the end of the data, and not just after the column used - though it says they will be when you press that button.  I assume the current code can't cope with not knowing how many variables to include.  Maybe you can fix that, or perhaps disable the Position button for this option?

N-thony:
> Great, the main one is done. Just checking on the other point, namely when getting Maximum number of components they are all at the end of the data, and not just after the column used - though it says they will be when you press that button. I assume the current code can't cope with not knowing how many variables to include. Maybe you can fix that, or perhaps disable the Position button for this option?

Resolved

rdstern:
@N-thony that's great, but I now detect another problem.  The world_bank_pop is a good dataset to use, because that has a variable called indicator with contents like `SP.URB.TOTL`.
So I tried first without the regex, and then I ticked regex.  With your code it never runs with regex.  Your command is still: 
`split <- stringr::str_split(string=indicator, pattern=stringr::fixed("."), n=Inf, simplify=TRUE)`

With regex ticked it should become simply:
`split <- stringr::str_split(string=indicator, pattern=".", n=Inf, simplify=TRUE)`

Try it, and you will see the difference.

I am a bit concerned that is might also still be incorrect in the Find/Replace options?

N-thony:
> world_bank_pop

@rdstern I tried on my side and it works fine
image

N-thony:
> @N-thony that's great, but I now detect another problem. The world_bank_pop is a good dataset to use, because that has a variable called indicator with contents like `SP.URB.TOTL`. So I tried first without the regex, and then I ticked regex. With your code it never runs with regex. Your command is still: `split <- stringr::str_split(string=indicator, pattern=stringr::fixed("."), n=Inf, simplify=TRUE)`
> 
> With regex ticked it should become simply: `split <- stringr::str_split(string=indicator, pattern=".", n=Inf, simplify=TRUE)`
> 
> Try it, and you will see the difference.
> 
> I am a bit concerned that is might also still be incorrect in the Find/Replace options?

@rdstern I have corrected the regex code. Please take a look.

rdstern (approved):
@N-thony if @shadrackkibet is ok, then this could be merged.

N-thony:
@HawardKetoyoMsatsi over to you.

HawardKetoyoMsatsi (changes requested):
The code looks good. Maybe if you could reduce blank lines 15-17 of dlgSplitText.vb to one line.

N-thony:
> The code looks good. Maybe if you could reduce blank lines 15-17 of dlgSplitText.vb to one line.

Done

N-thony:
> > The code looks good. Maybe if you could reduce blank lines 15-17 of dlgSplitText.vb to one line.
> 
> Done

@HawardKetoyoMsatsi I have resolved your comment above.

instat/dlgSplitText.vb
```diff
@@ -80,6 +77,12 @@ Public Class dlgSplitText
         ucrNudPieces.SetParameter(New RParameter("n", 2))
         ucrNudPieces.SetMinMax(2, Integer.MaxValue)
 
+        ucrChkIncludeRegularExpressions.SetText("Include Regular Expressions")
+        ucrChkIncludeRegularExpressions.SetDefaultState(False)
+
+        ucrChkAddKeyboard.SetText("Add keyboard")
+        ucrChkAddKeyboard.Enabled = False
```
shadrackkibet, line 84: Why manual checking?

instat/dlgSplitText.vb
```diff
@@ -80,6 +77,12 @@ Public Class dlgSplitText
         ucrNudPieces.SetParameter(New RParameter("n", 2))
         ucrNudPieces.SetMinMax(2, Integer.MaxValue)
 
+        ucrChkIncludeRegularExpressions.SetText("Include Regular Expressions")
+        ucrChkIncludeRegularExpressions.SetDefaultState(False)
```
shadrackkibet, line 81: Do you really need this?

instat/dlgSplitText.vb
```diff
@@ -92,9 +95,16 @@ Public Class dlgSplitText
         clsTextComponentsFixed = New RFunction
         clsTextComponentsMaximum = New RFunction
         clsBinaryColumns = New RFunction
+        clsStringCollFunction = New RFunction
+
+        ucrChkIncludeRegularExpressions.Checked = False
```
shadrackkibet, line 100: Same case here.

shadrackkibet (changes requested):

N-thony:
@shadrackkibet I resolved your comment above. Please take a look.

instat/dlgSplitText.vb
```diff
@@ -71,15 +72,16 @@ Public Class dlgSplitText
         dctPatternPairs.Add("Hyphen -", Chr(34) & "-" & Chr(34))
         dctPatternPairs.Add("Underscore _", Chr(34) & "_" & Chr(34))
         ucrInputPattern.SetItems(dctPatternPairs)
-        'ucrInputPattern.SetRDefault(Chr(34) & " " & Chr(34)) 'This is the default for clsTextComponents
```
shadrackkibet, line 77: Could you explain why you removed this? It is good practice to first understand why the code was added initially before removing it.

instat/dlgSplitText.vb
```diff
@@ -143,15 +156,44 @@ Public Class dlgSplitText
...
+                strPattern = Chr(34) & "," & Chr(34)
+            Case "Colon :"
+                strPattern = Chr(34) & ":" & Chr(34)
+            Case "Semicolon ;"
+                strPattern = Chr(34) & ";" & Chr(34)
+            Case "Hyphen -"
+                strPattern = Chr(34) & "-" & Chr(34)
+            Case "Underscore _"
+                strPattern = Chr(34) & "_" & Chr(34)
+        End Select
```
N-thony: This is because of the value coming from the input control e.g Comma , we just need the comma symbol. I don't see another way to do that if not like this. Otherwise, you can suggest to me how to do it simply.

instat/dlgSplitText.vb
```diff
@@ -202,7 +204,15 @@ Public Class dlgSplitText
         End If
     End Sub
 
-    Private Sub ucrAll_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrChkIncludeRegularExpressions.ControlValueChanged, ucrReceiverSplitTextColumn.ControlValueChanged, ucrInputPattern.ControlValueChanged, ucrPnlSplitText.ControlValueChanged
+    Private Sub ucrChkIncludeRegularExpressions_ControlConteChanged(ucrChangedControl As ucrCore) Handles ucrChkIncludeRegularExpressions.ControlContentsChanged
+        If ucrChkIncludeRegularExpressions.Checked Then
+            clsSplitDummyFunction.AddParameter("checked", "TRUE", iPosition:=0)
+        Else
+            clsSplitDummyFunction.RemoveParameterByName("checked")
+        End If
```
shadrackkibet, line 212: Simply `.SetParameter` for this control.

instat/dlgSplitText.vb
```diff
@@ -71,15 +72,16 @@ Public Class dlgSplitText
...
         ucrNudPieces.SetMinMax(2, Integer.MaxValue)
 
         ucrChkIncludeRegularExpressions.SetText("Include Regular Expressions")
-        ucrChkIncludeRegularExpressions.SetDefaultState(False)
+        ucrChkIncludeRegularExpressions.AddParameterPresentCondition(True, "checked")
+        ucrChkIncludeRegularExpressions.AddParameterPresentCondition(False, "checked", False)
 
+        'This is not yet implemented
         ucrChkAddKeyboard.SetText("Add keyboard")
         ucrChkAddKeyboard.Enabled = False
```
shadrackkibet: We encourage the hiding of controls that do not work.

rdstern (approved):

instat/dlgSplitText.vb
```diff
@@ -67,19 +65,26 @@ Public Class dlgSplitText
...
         dctPatternPairs.Add("Hyphen -", Chr(34) & "-" & Chr(34))
         dctPatternPairs.Add("Underscore _", Chr(34) & "_" & Chr(34))
         ucrInputPattern.SetItems(dctPatternPairs)
-        'ucrInputPattern.SetRDefault(Chr(34) & " " & Chr(34)) 'This is the default for clsTextComponents
         ucrInputPattern.bAllowNonConditionValues = True
 
         ucrNudPieces.SetParameter(New RParameter("n", 2))
         ucrNudPieces.SetMinMax(2, Integer.MaxValue)
 
+        ucrChkIncludeRegularExpressions.SetText("Include Regular Expressions")
```
shadrackkibet, line 80: Do we need this checkbox when splitting into binary columns?

instat/dlgSplitText.vb
shadrackkibet, line 188: This is clearly a duplication of what you've under `InitialiseDialog()`.

shadrackkibet (changes requested):
Under splitting into binary columns I got the following R code 
`split <- questionr::multi.split(var=village, split.char="instat.RFunction")`

@N-thony I would suggest you take time to first thoroughly test things before you even invite others to test them. Some obvious issues like the one above should be easy to detect if you test things thoroughly.

instat/dlgSplitText.vb
N-thony: @shadrackkibet any suggestion about this?

N-thony:
@shadrackkibet hopefully this is okay now. Take a look!

N-thony:
@shadrackkibet hopefully this is okay now? Take a look.

N-thony:
> @shadrackkibet hopefully this is okay now? Take a look.

@shadrackkibet how about this?

instat/dlgSplitText.vb
```diff
@@ -143,15 +148,47 @@ Public Class dlgSplitText
...
+                strPattern = Chr(34) & ":" & Chr(34)
+            Case "Semicolon ;"
+                strPattern = Chr(34) & ";" & Chr(34)
+            Case "Hyphen -"
+                strPattern = Chr(34) & "-" & Chr(34)
+            Case "Underscore _"
+                strPattern = Chr(34) & "_" & Chr(34)
+            Case Else
+                strPattern = Chr(34) & ucrInputPattern.GetText & Chr(34)
+        End Select
```
lloyddewit, line 182: If you have a dictionary that defines the correct character for each description (e.g. ',' for 'Comma ,'), then you can replace this `Select` with a single line.

lloyddewit (changes requested):

instat/dlgSplitText.vb
N-thony: I had a look at the code for the available option under splitting into binary and the R code, I think we don't need it. That is why is hidden when going to the splitting option

N-thony:
@lloyddewit I made the changes regarding your comment above. Please take a look.

instat/dlgSplitText.vb
```diff
@@ -143,15 +159,37 @@ Public Class dlgSplitText
...
         SplitTextOptions()
+        ChangeParametersValues()
+    End Sub
+
+    Private Sub ChangeParametersValues()
+        Dim strPattern As String
+
+        If dctPatternPairs.ContainsKey(ucrInputPattern.GetText) Then
+            strPattern = dctPatternPairs(ucrInputPattern.GetText)
+        Else
```
lloyddewit, line 180: Is the user allowed to enter patterns that are not in the pre-defined list?

lloyddewit (commented):
Looks good, just one open question

instat/dlgSplitText.vb
N-thony: @lloyddewit yes.

lloyddewit:
@shadrackkibet Thank you for your earlier review. There is one open comment. Please could you check that it's resolved?
I already set your other comments to resolved, I hope that's OK.
Thanks

lloyddewit (dismissed):

lloyddewit:
@rdstern There are changes since you last tested. Please could you test again? thanks

rdstern (changes requested):
@N-thony please try again with the tidyr world bank population.  There is a variable that splits nicely into 3 with dot as the separator.  But it also works with regular expression ticked, which it shouldn't!  It currently gives split <- stringr::str_split_fixed(string=indicator, pattern=stringr::fixed("."), n=3)
so I think it is still not permitting the regular expression?

N-thony:
> @N-thony please try again with the tidyr world bank population. There is a variable that splits nicely into 3 with dot as the separator. But it also works with regular expression ticked, which it shouldn't! It currently gives split <- stringr::str_split_fixed(string=indicator, pattern=stringr::fixed("."), n=3) so I think it is still not permitting the regular expression?

@rdstern this seems to be fixed now. Please take a look. Thanks.

rdstern (approved):
I am happy now and approving.  @lloyddewit there is just a tiny code change since you accepted before.

lloyddewit (approved):

lloyddewit:
@shadrackkibet there's still an open comment from you above, so we also need your approval before merging. Please could you take a look? thanks

shadrackkibet (approved):

PR #6906, opened 2021-11-09
Improved speed issues

### LINKED ISSUE #6901: Speed issues with large datasets
From https://github.com/africanmathsinitiative/R-Instat/issues/6882#issuecomment-960528018

> I am also concerned there could be serious speed issues with the new system.
> I opened a large set of files - both fairly long and fairly wide. Took 40 seconds on version 0.7.2 and 30 seconds longer on the merged version. I can live with that. But then:
> a) Converting one variable to be a factor 15 seconds. Almost instantaneous - certainly less than 2 seconds earlier
> b) Describe > One Variable > Summarise for that variable, 13.8 seconds on merged version and less than 1 second earlier.
> 
> It may all have nothing to do with the changes to the grid, but it is all now disturbingly slow when the data set is non-trivial.
> I have now tried with more modest sized sets, e.g. diamonds (from ggplot2) with 10 variables and 50,000 cases. All ok. Then I added rockart from daag package, with 102 cases and 600 variables and still all is ok.
> 
> The MICS data from Pakistan is where I am finding the problems. This has 8 sets imported from SPSS files. For example hh is about 50,000 cases by 350 variables. Others are up to 300,000 cases, etc. No problem with version 0.7.2, but (as mentioned above) has become annoyingly slow for simple things on merged version.

ChrisMarsh82:
Add Has Changed flag to dataframe page to stop unnecessary gird refresh.

### DIFF
--- a/instat/Model/DataFrame/clsDataFramePage.vb
+++ b/instat/Model/DataFrame/clsDataFramePage.vb
@@ -28,6 +28,7 @@ Public Class clsDataFramePage
     Private _clsRLink As RLink
     Private _clsDataFrame As DataFrame
     Private _lstColumns As List(Of clsColumnHeaderDisplay)
+    Private _hasChanged As Boolean
 
     Private ReadOnly Property iColumnIncrements As Integer
         Get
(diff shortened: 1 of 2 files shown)


rdstern:
@ChrisMarsh82 sounds good.  Just to mention a related issue on refreshing the grid, which might be in a similar place in the code?
When I do simple operations, like adding a variable, the grid refreshes fine, but the information on the bottom line doesn't refresh.  For example, if I start with 7 variables and add 3 more it still says 7.

dannyparsons:
Thanks, I've added a new issue for that #6907

rdstern:
@dannyparsons thanks.  At the same time a look at what to do with that status line when there is a filter would be good.  And now we sometimes may also have a selection of variables.

dannyparsons (approved):
This is definitely an improvement. 

For dialogs which display output there is now no unnecessary refreshing so this is almost instant again.

The loading and refreshing of data is about twice as fast as the current version. However, still slower than before. e.g. Dodoma from the library used to take ~1s and now ~4s. A larger file with 6 data frames used to take ~3s but now ~10s. So there's still some big efficiencies to find.

I'll merge this as it is an improvement on the current version.

Issue #6882, opened 2021-10-27
Resolving issues after merging the paging of wide data sets


rdstern:
I am making a new issue here.  @shadrackkibet comments about aspects that no longer work are important to fix before any new release.  This is in #6707.

After his comments I tried closing the data book.  I use it quite often and it used to return to our (now quite nice) front screen.  Now is appears not to work.  In fact it does work, in the sense that when you open a new data frame, then all the old ones disappear.  But when you delete it is as though nothing has happened.

Shadrack also mentions his new select features.  The action with filters is sort of ok, but not perfect.  I wonder what else there is to fix?

N-thony:
I have tried to load the quelimane.RDS dataset and I'm getting the following error
![image](https://user-images.githubusercontent.com/68591383/139633146-423223f4-da3a-43dc-92c8-d80f6bc60acd.png)
![image](https://user-images.githubusercontent.com/68591383/139633224-f3e85dba-de58-4d27-a7a8-e413d9c2d8a4.png)

Here is the dataset: [quelimane6.zip](https://github.com/africanmathsinitiative/R-Instat/files/7451233/quelimane6.zip)

rdstern:
There is a very minor point on deleting all **_visible_** columns.  It should direct you to `Prepare > Data Frame > Delete Columns/Rows`.  I had a data frame with 83 rows and wanted to delete 81-83.  (these were the only visible ones.   I certainly didn't want to delete the whole data frame!

N-thony:
I have tried to open the script window but unfortunately, I'm getting the error in the screenshot below
![image](https://user-images.githubusercontent.com/68591383/140041028-8a178dea-ec1b-4dfc-8889-94fc0ed69abb.png)
![image](https://user-images.githubusercontent.com/68591383/140024623-7dda97c0-67e1-49e7-a616-df78b2f9a153.png)

rdstern:
I am also concerned there could be serious speed issues with the new system.
I opened a large set of files - both fairly long and fairly wide.  Took 40 seconds on version 0.7.2 and 30 seconds longer on the merged version.  I can live with that.  But then:
a) Converting one variable to be a factor 15 seconds.  Almost instantaneous  - certainly less than 2 seconds earlier
b) Describe > One Variable > Summarise for that variable, 13.8 seconds on merged version and less than 1 second earlier.

It may all have nothing to do with the changes to the grid, but it is all now disturbingly slow when the data set is non-trivial.

[...]

PR #6908, opened 2021-11-09
Updated refresh of data from the databook

### LINKED ISSUE #6912: Data View Window not clearing contents after trying to delete a dataset that is loaded.
I loaded a dataset from the library and realised it was the wrong one. I decided to delete it but it didn't clear. I also tried to close the data book but it didn't clear as well.

To reproduce the bug
1. Try and import any dataset either from the library, creating a new one, or from a file.
2. Once data is loaded, right-click on the tab at the bottom and click on `delete`.
3. See error.
4. You could also go to File and scroll down to `Close Data Book` and click on that.
5.  You will notice nothing happens on the Data View window.

What is expected is the data view window to have cleared its contents after clicking on delete or clear form.

Note: Once you click on delete/ close data book and try importing a new dataset, it automatically replaces the one you had even if it was not cleared from the data view window.

shadrackkibet:
fixes #6912

@ChrisMarsh82 I just copied the code you provided [here](https://github.com/africanmathsinitiative/R-Instat/issues/6882#issuecomment-953742432). It seems to work. However, the status label below is still there and the flash screen we had is lost.

@dannyparsons this is also an improvement from what we have in the master branch. I suggest we merge.


dannyparsons:
@shadrackkibet Which bug should this fix so I can check it carefully?

shadrackkibet:
**Close Data Book** under File menu should now work partially.

ChrisMarsh82 (changes requested):
Separated out deleteAlldataframes to make the code more readable.

ChrisMarsh82:
@shadrackkibet  I may have done this wrong in GitHub but tried to commit a suggestion of how the code could be written. Split up the code into another procedure to make it more readable.

dannyparsons:
@ChrisMarsh82 The commit worked. Could you look into why the Start screen panels have disappear when the data book is closed? It looks like when `grid.AddWorksheet(fillWorkSheet)` runs in `AddAndUpdateWorksheets()` in `ucrDataView` when the data frame is first loaded then the panels get hidden. I can't see a line of code explicitly hiding them though, and it didn't do this before. Maybe related to a change in the designer?

The Start panels could always be visible because they are under the grid so there's no need to hide them when there's data in the grid.

shadrackkibet:
> @shadrackkibet I may have done this wrong in GitHub but tried to commit a suggestion of how the code could be written. Split up the code into another procedure to make it more readable.

This worked fine Chris. I like your idea of splitting the code.

ChrisMarsh82:
Added an update for the dataframe meta data that was not getting reset. 
@dannyparsons I've sorted the issues with the start screen however had to do it in #6919

dannyparsons (approved):
This fixes the deleting of data frames and closing the data book issues. Still some refreshing of status and info bar needed that should be in other PR

PR #6909, opened 2021-11-09
updated columnMetaData to display even if no databook open

### LINKED ISSUE #6895: Bug when trying to open the View>script and other windows
I have tried to open the script window but unfortunately, I'm getting the error in the screenshot below
image

ChrisMarsh82:
fixes #6895
ColumnMetaData was causing issues as if there isn't a worksheet then the reoGrid cannot be visible. Also updated order of remove to create to make sure current worksheet is updated.

### DIFF
--- a/instat/ucrColumnMetadata.Designer.vb
+++ b/instat/ucrColumnMetadata.Designer.vb
@@ -91,159 +91,162 @@ Partial Class ucrColumnMetadata
         Me.grdVariables.ContextMenuStrip = Me.cellContextMenuStrip
         Me.grdVariables.Dock = System.Windows.Forms.DockStyle.Fill
         Me.grdVariables.LeadHeaderContextMenuStrip = Nothing
-        Me.grdVariables.Location = New System.Drawing.Point(3, 23)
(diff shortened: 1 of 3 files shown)


dannyparsons:
@ChrisMarsh82 for GitHub to link the PR to the issue you need to put "fixes" in the description not the title of the PR and use a # e.g.

fixes #6895

dannyparsons:
@ChrisMarsh82 This works well but you can make the grids completely hidden when there's no data? Here is how it looks in 0.7.2

This is more consistent with the data view window. Also, with the grid visible, you can easily crash the software by typing into it. So all round hiding would be best. (Not sure why column and data frame window back colours are different, they should be the same as the Data View)

image

ChrisMarsh82:
@dannyparsons ok, that should be easy to change

ChrisMarsh82:
@dannyparsons updated so grid doesn't display. There is an issue with the dataframe meta when closing the databook but will sort that in #6908

dannyparsons (approved):

PR #6911, opened 2021-11-09
Fixed bug loading older files with filters applied

### LINKED ISSUE #6891: Bug when trying to load quelimane dataset
I have tried to load the quelimane.RDS dataset and I'm getting the following error
image
image

Here is the dataset: [quelimane6.zip](https://github.com/africanmathsinitiative/R-Instat/files/7451233/quelimane6.zip)

dannyparsons:
fixes #6891

This bug wasn't due to the recent changes to the grid but to the recent additions to the filter objects. It's a very good bug to find and a good reminder of the importance of backwards compatibility so we can support users with older `data_book` files.

@shadrackkibet This was caused by recent additions to the parameters for filters i.e. `"inner_not"` and `"outer_not"`. Because these parameters don't exist in filter objects in `data_book`s saved a long time ago, it gives an error when trying to access these parameters. I hadn't thought too much about this before. I should have put these structures in place at the beginner so it would be clear what to do when making changes.

It's easy to resolve. I have created `check_filter()` which checks an incoming filter object when the data is imported and adds defaults to parameters which may not exist. We just need to maintain this function for any further changes to the filter object.

For extra security, I've added `isTRUE()` to checks on logical objects in the filter function. It shouldn't be needed now, but I think it's good practice as it will catch any `NULL`s that slip through.

We should also do the same for the `column selection` objects when these change in future.

### DIFF
--- a/instat/static/InstatObject/R/Backend_Components/calculations.R
+++ b/instat/static/InstatObject/R/Backend_Components/calculations.R
@@ -53,6 +53,14 @@ calculation$set("public", "data_clone", function() {
 }
 )
 
+# This ensures cloned filter objects from older saved data_books have the expected parameters
+check_filter <- function(filter_obj) {
+  if (is.null(filter_obj$parameters[["and_or"]])) filter_obj$parameters[["and_or"]] <- "&"
+  if (is.null(filter_obj$parameters[["outer_not"]])) filter_obj$parameters[["outer_not"]] <- FALSE
+  if (is.null(filter_obj$parameters[["inner_not"]])) filter_obj$parameters[["inner_not"]] <- FALSE
+  return(filter_obj)
+}
+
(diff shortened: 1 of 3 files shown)


N-thony (approved):

shadrackkibet (approved):
@dannyparsons thanks for the information. This is something we should all keep in mind when making any future changes in the filter and column selection.

@lloyddewit I thought we should document the issue of backward compatibility. Perhaps we should have it somewhere under wiki? What do you think?

dannyparsons:
I agree, it would be good to document this somewhere.

lloyddewit:
> 
> 
> I agree, it would be good to document this somewhere.

@shadrackkibet Thanks for the suggestion, please could you document this in the R-Instat wiki? Do you think a couple of explanatory sentences and a link to this issue would be adequate?

PR #6916, opened 2021-11-10
Added Violin Plot, Box Plot, Line Graph and Dot Plot to Two Way Frequency Sub Dialog

### LINKED ISSUE #4951: Improve the Two Variables and Three Variables > Frequencies dialogues

Gioche6:
Fixes #4951

@rdstern , I have  added the new graphs as discussed. 

> I have added Violin Plot, Box Plot, Dot PLot and Line Graph.

> The Line graph and Bar graph are the ones used where parameter `fun="xtab"`. I have restricted that, however bar chart is used also when parameter `fun="grpfrq"`. I have also made that restriction on the sub dialog display.

> Additionally, the three other plots, ie, violin, plot and dot plots are only available on `fun="grpfrq"`. I have made that possible.

Kindly review and test if it works fine. Thanks


Gioche6:
@rdstern , The new graphs will be available for both the two variable frequency and three variable frequency dialogs as the use the same sub dialog and the same R function

rdstern:
That's fine and just what I would expect

rdstern:
I am using the `survey `data as a simple example.  This is really coming on nicely.  
1) With the new boxplot option here is an oddity:
image
It is ok if you don't use the facets.  I really like that checkbox - was that there before or did you add it?
2)  With the boxplots selected if you choose percents it gives an error and says don't do that.  I wonder if those options should be disabled if you choose that sort of plot.
3) I wonder if Count might need to be labelled Count (or Values).  With the boxplot, etc you are examining the values.

Gioche6:
@rdstern , This fixes item number 2. Item number 1 and 3, I am thinking of creating a separate PR for item 1 and 3 because I will be handling the main dialogs and not the sub dialog anymore. Or is it okay if I make the changes here for easier testing?

rdstern (commented):
To evaluate this I need to know what you decided to change and what was there before.  So, most of the time I see the function used is grpfrq and that gives facets.  Then I tried with row frequencies and got the following:
# Code generated by the dialog, Two Way Frequencies

last_graph <- sjPlot::sjplot(data=efc, c12hour, e15relat, type="violin", title="", fun="xtab", margin="row")
and that gave an error.  Have we always checged to xtab here?  I am wondering if this is getting into a more complicated mess through trying to add different flexibility?  I need to understand what was there before and what was added.

Gioche6:
@rdstern , The changes that i have made are the following. The changes are only for the two variable sub dialog

1. I added the three plots, namely Box pot, line graph and dot plot.  
2. I enabled the line graph option which was disabled. 
3. The plots (Box plot, dot plot, violin plot , bar graph ) are only used when the `fun=grpfrq` in the `sjplot` function . I ensured that line graph is disabled when the Count option is used. Count option in the radio buttons passes parameter `fun=grpfrq` while the rest of the radio buttons passes parameter `fun =xtab`. 
4. Since boxplot option bring an error when there is percentage, I have made the percentage checkbox disabled whenever box plot option is checked. 

Those are the only changes I have made. All of them revolve around the new plot option. I have not interfered with anything else in the function.

rdstern (approved):

instat/sdgTwoWayFrequencies.vb
```diff
@@ -129,6 +134,23 @@ Public Class sdgTwoWayFrequencies
...
+                rdoBoxPlot.Enabled = False
+                rdoLineGraph.Enabled = True
+                rdoBar.Enabled = True
+                clsTwoWayGraphFreq.RemoveParameterByName("facet.grid")
+            ElseIf clsTwoWayGraphFreq.GetParameter("fun").strArgumentValue = Chr(34) & "grpfrq" & Chr(34) Then
+                rdoLineGraph.Enabled = False
+                rdoDot.Enabled = True
+                rdoViolinPlot.Enabled = True
+                rdoBoxPlot.Enabled = True
+                rdoBar.Enabled = True
```
lloyddewit, line 150: If we order these statements the same as the `if` clause above (e.g. rdoViolinPlot, rdoDot, rdoBoxPlot etc.) then someone maintaining this code will find it easier to see the difference between the `If` and `ElseIf` clauses.

lloyddewit (commented):
Looks good, just one small comment.
Please can you also reserve the merge conflict? Thanks

Gioche6:
@lloyddewit , I have resolved the merge conflict.

instat/sdgTwoWayFrequencies.vb
Gioche6: Thank you for this. Indeed it will make it easier for the one maintaining the code.

lloyddewit: @Gioche6 You marked this comment as resolved but I don't see the code changes. Are the changes committed and pushed?
Thanks

Gioche6:
@lloyddewit . I have arranged them well now, Please have a look.

lloyddewit (approved):

lloyddewit:
@rdstern since your last approval, @Gioche6 resolved a merge conflict. Please could you test/approve again? thanks

rdstern (approved):

shadrackkibet:
@lloyddewit this merge might have introduced build errors in the master branch. I think that is caused by manually editing the resource file while sorting out conflicts.

image

Not sure how we can efficiently sort this out. Should we revert the merge? Looking forward to your guidance.

maxwellfundi:
@shadrackkibet I think this happened with the commit of merging master on to this branch in a conflict fixing. I'm future a normal revert commit can fix such. I strongly suggest not editing resource files @Gioche6 in future

PR #6919, opened 2021-11-10
fixes 6907 Updated refresh of info on refresh

ChrisMarsh82:
Add refresh and made sure Dataframe page knows current total from filter and resets paging if needed


dannyparsons (approved):
Refreshing of info and status bar now works well when updating, deleting and closing.

PR #6925, opened 2021-11-10
Updated grid so any new dataframes added become current tab

ChrisMarsh82:
This works as when the original dataframes are added the main form selects the first worksheet therefore this can be added without affecting the initial load


dannyparsons (approved):
This works well. I've found another case to resolve that I'll report as an issue, but will merge this first as it's an improvement in most cases.

ChrisMarsh82:
@dannyparsons I think I may have messed this up regarding pull requests. The code changes in this request should be reverted in #6938

dannyparsons:
Thanks for spotting that, I think I resolved it by adding back the refresh line in this PR #6940

PR #6926, opened 2021-11-10
Fixed bug when trying to show Levels/Labels Dialogue

### LINKED ISSUE #6900: Possible problem with the Levels/Labels dialogue
This bug may be to do with the new paging, but I think it could have arisen because of changes to the right-click Levels/Labels dialogue.  At least that is worth a check.
What I was doing is not sensible, but it still should not give an error.  It will be good to see if you can repeat it?
a) Load the example data called words from the stringr package.
b) Change the variable to a factor and then use Levels/Labels.  It confirms that each word is different!
c)  Now load sentences from the stringr package
d) Make this into a factor and again use the Levels/Labels 
e) Delete the sentences data frame
f) Now use Levels/Labels again on the words dataset.  
It gives me an error on the master copy and throws me out of the program.  The error message implies it is still looking for the sentences data.
Are you able to repeat it?  
I tried also with one > Variable Summary instead of Levels/Labels and that didn't give an error.  I also tried the Levels/Labels in Version 0.7.2 and that worked fine.

If you can get the error - hope so.  That would be a good one to fix

N-thony:
Partially fixes #6900 
This essentially concerns the first part of issue #6900 on the way the application was throwing due to the fact that the deleted dataset from the Data View after showing the Levels/Labels dialogue was still present when trying to show the Levels/Labels of another data set.
@rdstern @africanmathsinitiative/developers this is ready for review. Thanks.

### DIFF
--- a/instat/dlgLabelsLevels.vb
+++ b/instat/dlgLabelsLevels.vb
@@ -119,7 +119,7 @@ Public Class dlgLabelsLevels
         Dim clsGetColumnFunction As RFunction = ucrReceiverLabels.GetVariables()
         clsGetColumnFunction.RemoveAssignTo()
 
-        If Not ucrReceiverLabels.IsEmpty Then
+        If Not ucrReceiverLabels.IsEmpty AndAlso {"factor"}.Contains(ucrReceiverLabels.strCurrDataType) Then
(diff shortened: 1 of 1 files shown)


dannyparsons (approved):
This seems to work ok for me. @rdstern can you test?

rdstern (approved):
Great.  This first part works fine now.
I notice the sentences factor takes a little more time to load.  That's all a different issue!

PR #6930, opened 2021-11-11
fixes #6787 - simple fix that swaps grids and labels on buttons

### LINKED ISSUE #6787: Swap data and metadata has problems when a script window is opened.
With wide data it is often more useful to have the column metadata window open instead of the data.  Here is an example where I hav e chosen to swap the 2 windows:

image

Now, every dialogue has a script button.  This opens a 3rd window above the 2 main windows - at least it does when there is the data window.  But after swapping the 2 windows I get this!

image

That's not at all what I want.  I simply want to keep the 2 windows as before, i.e. column metadata on left and Output on the right.  And then the script window above.  That's what I get with the data on the left.  

I hope this is easy to fix?

ChrisMarsh82:
This will fix the issue simply however there may be an issue in the future as it swaps labels on buttons and if we were using tool tips this could get lost.
@dannyparsons do I need to set up translations on these buttons if I'm swapping labels?

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -517,20 +517,23 @@ Public Class frmMain
             Else
                 splOverall.Panel1Collapsed = True
             End If
-            If mnuViewSwapDataAndMetadata.Checked AndAlso mnuViewDataView.Checked Then
-                splDataOutput.Panel1.Controls.Add(ucrColumnMeta)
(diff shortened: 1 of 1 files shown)


dannyparsons (approved):
I agree there may be issues with this in future for tooltips or translations but this is an improvement on the current version.

PR #6931, opened 2021-11-11
Fixed bug recoding factor levels

### LINKED ISSUE #6898: Small problem with the new Prepare > Factor > Recode dialogue
The following shows the problem with the new merged code.
The example is from the library, Instat, Climatic Niger sadore.xlsx.  Read the second sheet, and omit 4 lines so it is named correctly.
Then make Mois into a factor and then use the dialogue.
The Mois variable has 14 levels - not 12 as it should.
Change Janvier to janvier and then also Fevrier to fevrier.  Then press Ok.
The command generated is as follows:
```
# Code generated by the dialog, Recode Factor
Mois <- data_book$get_columns_from_data(data_name="Sadore", col_names="Mois", use_current_filter=FALSE)
Mois_recoded <- forcats::fct_recode(.f=Mois, `mars`="mars", `avril`="avril", `mai`="mai", `juin`="juin", `juillet`="juillet", `aout`="aout", `septembre`="septembre", `octobre`="octobre", `novembre`="novembre", `decembre`="decembre", `fevrier`="Fevrier", `janvier`="janvier")
data_book$add_columns_to_data(data_name="Sadore", col_name="Mois_recoded", col_data=Mois_recoded, before=FALSE, adjacent_column="Mois")
```

You can see it correctly changes from Fevrier, but not from Janvier - see the code above.  So the resulting factor still has 13 levels - and not 12.

dannyparsons:
fixes #6898

I've resolved this by changing the Recode Factor dialog back to using `plyr::recode()`. This function works in a different way allowing the duplicate values.

@N-thony @rdstern Can you confirm this resolves the issue?

I also updated some text on the dialog. We should use `NA` in capitals for the missing value code. We should also not use `_` for text on dialogs, it should always be readable normal text e.g. `Add NA` instead of `add_na`.

### DIFF
--- a/instat/dlgRecodeFactor.resx
+++ b/instat/dlgRecodeFactor.resx
@@ -652,7 +652,7 @@
     <value>2</value>
   </data>
   <data name="rdoAddNa.Text" xml:space="preserve">
-    <value>add_na</value>
+    <value>Add NA</value>
   </data>
   <data name="rdoAddNa.TextAlign" type="System.Drawing.ContentAlignment, System.Drawing">
     <value>MiddleCenter</value>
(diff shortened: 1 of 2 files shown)


N-thony (approved):
@dannyparsons I have checked, and I confirm this solve the issue

Issue #6921, opened 2021-11-10
Column Metadata disappears/doesn't detect changes after deleting a data frame (refresh issue?)


dannyparsons:
- Open two data frames e.g. Rwanda data
- Open the Column Metadata Window
- Open the Dataframe Metadata Window
- Delete the first data frame from the Data View

The Dataframe Metadata has correctly updated to now have 1 row.

The Column Metadata correctly now has one tab when you click in it. But the data is still for the first data frame. The grid is also locked or frozen.

- Hide and show the Column Metadata

Now the correct metadata displays. However, I can type in any cell and it doesn't send the changes to R. Some column which should be restricted are not. It looks like these settings haven't been applied.

If I change the name of the data frame from the Dataframe Metadata Window then the Column Metadata refreshes again. And now when I type something it correctly sends the changes to R.

Looks like it isn't getting updated correctly and the settings and events triggering R code to run are not happening.

PR #6933, opened 2021-11-11
fixes refreshing issue on delete

### LINKED ISSUE #6927: Problem with deleting the dataset from the dataview seems to persist
I have loaded two datasets words and sentences from the library - stringr package and tried to delete the sentences dataset from Data View right-click on the tab of sentences but after deleting, the sentences were still present in the Data View. Then I have tried the load another dataset, you can try with any, in my case I have tried with rwanda dataset, and found that the sentences dataset was no longer in the Data View.
Here is the screenshot after deleting sentences dataset from Data View
image
Then here after loading rwanda dataset
image

But check first if you can reproduce that.

dannyparsons:
fixes #6927

### DIFF
--- a/instat/static/InstatObject/R/instat_object_R6.R
+++ b/instat/static/InstatObject/R/instat_object_R6.R
@@ -990,7 +990,7 @@ DataBook$set("public", "delete_dataframes", function(data_names, delete_graph_bo
   # TODO need a set or append
   for(name in data_names) {
     private$.data_sheets[[name]] <- NULL
-    data_objects_changed <- TRUE
+    self$data_objects_changed <- TRUE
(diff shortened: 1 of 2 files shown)


dannyparsons:
@africanmathsinitiative/developers this is ready for review. Please check this fixes the issue.

dannyparsons:
This also fixes #6921 which was actually an existing bug, so we can't blame everything on @ChrisMarsh82!

@ChrisMarsh82 The UI refresh functions didn't work so I went for a disable/enable as the least noticeable manual refresh.

rdstern (approved):
Great - all seems to work fine now

PR #6934, opened 2021-11-11
Fixed a bug when using column selection

shadrackkibet:
@dannyparsons thanks for bringing this into my attention. I did confirm that column selection isn't working properly. On testing, I identified two issues.

-  The name of the column selection wasn't updating correctly i.e it only takes the default name `sel`
-  Column selection wasn't running properly, it only worked once from using the dialog under prepare menu.

Both of these issues are due to my previous implementation.

@rdstern please test if column selection now works properly from Data options and directly from Prepare menu.

I will work on restoring columns selection via right-clicks separately.


dannyparsons:
Looks good to me. Couple of naming points:

- Can we call it "Column Selection" instead of "Selection", on the menu and within the dialog? "Selection" on it's own doesn't really mean much.
- When I do "Numeric Range" can the label for the textbox be "Prefix"? I was confused at first what "Matching String" meant in this context.

rdstern (dismissed):
@dannyparsons these features seem to work!  I wonder whether we include this, because it is just a start, but we have been waiting a long time to have something on this.  What's there seems to work and I am not sure it will worry users if it is included.  
There is tidying, of course, e.g. the Remove Current Selection is a useful button, but should be disabled when there is no selection.  We should be able to select on variable type, we should be able to view the selection in the grid, and so on.  But this is a start at least.

shadrackkibet:
@dannyparsons I did make the changes you suggested above. However, the dialog name will reflect on runtime once I make the translation db change which I will do on a separate PR together with the frmmain change.

dannyparsons (approved):
This works well and fixes some existing bugs which prevent this use of this feature, so I'm happy to merge it.

PR #6935, opened 2021-11-11
Changed the Infilling daily climatic dialog to use chillR::patch_daily_temps() plus minor improvement

### LINKED ISSUE #6915: Check on function for infilling daily climatic data
This should be a useful function and can be mentioned, and perhaps promoted.  This is the dialogue: 

image

It currently uses an out-of-date function from chillR and perhaps there could be a quick check that this works.  Also I wonder if it works - or could work - with a filter.  I wonder whether infilling might be better if there were an option to do it separatley in (say) Winter and Summer?
We should also get someone in the team to try this out.  Could we get some sample datasets - possibly using CHIRTS and/or ERA5?

N-thony:
Fixes #6915
@dannyparsons @rdstern I have changed chillR::patch_daily_temperatures() to the new function chillR::patch_daily_temps() as requested in the corresponding issue. Please take a look.


rdstern (dismissed):
@N-thony this seems to work fine.  I am not sure what the new feature brings, but think it could be important and wonder whether we gamble and include it this time?  Over to you @dannyparsons

dannyparsons:
I'd like to be cautious and not change this last minute, as this is an important feature and I don't yet understand why there was a completely new function. The additional parameter looks useful. It's great this change was done so quickly. I suggest merging it soon after the release and testing well in R the two functions and the new parameter.

lloyddewit:

instat/dlgInfillMissingValues.vb
```diff
@@ -416,6 +445,14 @@ Public Class dlgInfillMissingValues
         clsAveFunction.AddParameter("FUN", clsROperatorParameter:=clsBracketOperator, iPosition:=2)
     End Sub
 
+    Private Sub ucrchkTimeInterval_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkTimeInterval.ControlValueChanged
+        If ucrChkTimeInterval.Checked AndAlso ucrNudTimeInterval.GetText <> "" AndAlso ucrInputTimeInterval.GetText <> "" Then
```
Ivanluv, line 449: Could use ucrNudTimeInterval.isEmpty()  and ucrInputTimeInterval.isEmpty() instead

Ivanluv:
@shadrack could you help me in reviewing the data object code

instat/dlgInfillMissingValues.vb
N-thony: @Ivanluv I have resolved this.

Ivanluv (commented):
The vb code looks okay from my end

lloyddewit (dismissed):

lloyddewit:
@rdstern just one small change since your last approval, that shouldn't have affected the functionality. Please could you test/approve again?

@shadrackkibet I couldn't see anything controversial in the R code so I suggest that we go ahead and merge anyway.

Thanks!

dannyparsons:
I'd just like to check the difference between the two functions before we merge, particularly as this is an important feature for WMO workshops.

rdstern (changes requested):
@N-thony 2 small changes.
1) The setting of the new parameter isn't remembered when you return to the dialogue.
2) Could you change the saving control.  With save single (in the same dialogue) it includes the position button.  It would be good to use that as well here.

N-thony:
> @N-thony 2 small changes.
> 
> 1. The setting of the new parameter isn't remembered when you return to the dialogue.
> 2. Could you change the saving control.  With save single (in the same dialogue) it includes the position button.  It would be good to use that as well here.

@rdstern I had a quick look into the R code and in order to make the change concerning item 2. I needed to edit the R code and this result in the fact that we will not have the following output again in the Output Window.
@dannyparsons what do you think of that change?
image

dannyparsons:
Looks good, but what is the `NULL` in the output?

rdstern:
@N-thony are you saying you now get no output at all?  I did like the one-line summary and (in general) like to have something in the output window from the functions that change the data frame.

N-thony:
@rdstern as we discussed some minute ago, I put back the input new column control and got rid of the `NULL` in the output when printing the one-lin summary. Please take a look?

rdstern (approved):
Check with @dannyparsons but I would be happy to see this merged now.

dannyparsons (approved):

PR #6937, opened 2021-11-11
fixes #6902 deleting columns now look all visible columns and not jus…

### LINKED ISSUE #6902: Deleting all visible columns
From https://github.com/africanmathsinitiative/R-Instat/issues/6882#issuecomment-956098203

> There is a very minor point on deleting all visible columns. It should direct you to Prepare > Data Frame > Delete Columns/Rows. I had a data frame with 83 rows and wanted to delete 81-83. (these were the only visible ones. I certainly didn't want to delete the whole data frame!
> 
> Perhaps another small point is the refreshing of the settings after variables wk3 to wk76 have been deleted. In the example below there are (now) about 10 variables, but it still says 79, from before they were deleted.

image

ChrisMarsh82:
Set column check to look at total columns and sets blank row if row count is 0

### DIFF
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -128,13 +128,14 @@ Public Class ucrDataView
 
     Private Sub AddRowData(dataFrame As clsDataFrame, workSheet As Worksheet)
         Dim textColour As Color
-        Dim rngDataRange As RangePosition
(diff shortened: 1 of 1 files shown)


dannyparsons (approved):
Works well on both parts, great!

PR #6939, opened 2021-11-12
Name changes in the translation db and frmmain

shadrackkibet:
@dannyparsons these are just minor changes in the frmmain and translation db as suggested in #6934


dannyparsons (approved):
This looks good to me. 
@shadrackkibet Could you explain what you changed in the database file as there isn't a way of comparing the change to a previous version?

shadrackkibet:
I changed the highlighted text from **Select** to **Column Select**

image

PR #6953, opened 2021-11-16
Added R-Script package

ChrisMarsh82:
Added R-package for use later on when changing output screens


lloyddewit:
@ChrisMarsh82 As discussed, the current NuGet package includes a number of XUnit dependencies. I will try and make a new NuGet package without these dependencies. I can't predict how long this will take. I will aim for the end of the week. I hope that's OK

lloyddewit:
> 
> 
> @ChrisMarsh82 As discussed, the current NuGet package includes a number of XUnit dependencies. I will try and make a new NuGet package without these dependencies. I can't predict how long this will take. I will aim for the end of the week. I hope that's OK.

@ChrisMarsh82 I made a new RScript NuGet package that does not include the XUnit dependencies (v. 1.0.4). This should reduce the number of dependencies from 32 to 5:

image

The quickest thing may be to close this PR and make a new one with RScript 1.0.4.

ChrisMarsh82:
Created a new pull request that overrides the original see #6975

PR #6954, opened 2021-11-16
Replace form.Hide with from.Close in order to work on Linux

ChrisMarsh82:
Updated all form.Hide to form.Close

@shadrackkibet can you check this doesn't break any functionality. I have checked areas that I've changed and the forms close and reload correctly however I'm not confident on all functionality

### DIFF
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -1362,12 +1362,12 @@ Public Class RLink
                                                               End If
                                                               While thrRScript.IsAlive
                                                                   If bErrorMessageOpen Then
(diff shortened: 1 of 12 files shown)


dannyparsons:
As long as the dialogs can still reopen in their previous state this should be fine.

@shadrackkibet The only other thing I can think of to check is dialogs where we set a default before opening it e.g. Levels/Labels from the grid right click menu. I think we can still set properties, because it works before the dialog has ever been opened - but good to check this.

I can't think of any other potential functionality issue.

lloyddewit (approved):

dannyparsons (approved):

PR #6960, opened 2021-11-18
Added context numbers to the repeatedly used sub-dialogs and linked them to Help Information

### LINKED ISSUE #6967: Set Links for the sub-dialog Help so that  you are able to jump to a specified anchor point
First there are (at least for this task) 2 sorts of sub-dialogue.  If you look at Describe One Variable > Summarise and use the Customise button you will see there is an option Choose Summaries, for that sub-dialogue.  Now that is a sub-dialogue used by quite a number of dialogues, and Rachel has a special place in the Help for those.  The graph sub-dialogues is another example used by many dialogues.    

In contrast the Describe > One Variable > Graph has a Graph Options button and that is a sub-dialogue, just for that dialogue.  So then we jump to an anchor point in the Help for that dialogue.    

For that dialogue there isn't the proper help yet, but I have made the anchor point.  

Rachel has included 4 so far, that are used repeatedly, namely Data Options, Plot Options, Summaries and Save Column Position.  There may be more?  You could include those links in the code for now.

Wycklife:
Fixes #6967

I have done the following;

1. Replaced the old Help file with the updated one
2. I have set the context numbers for the four sub-dialogs so that when you click on the Help button, it opens help information/anchor points for that sub-dialog.
3. These sub-dialogs are used repeatedly in most of the dialogs. They include; Data Options, Summaries sub dialog, Plot Options sub-dialog, and Save Column Position sub-dialog
  
@rachelkg Could you test whether this is okay? @rdstern You could have a look too. Thank you.


rdstern (approved):
There were 2 tasks in the issue that has led to this pull request.  The first was to link the repeatedly used sub-dialogues to the help file.  That is working fine.
The second is to link those sub-dialogues only used once to an anchor point in the help for that dialogue.  That is not working yet.
I am approving, because what was done is already very useful and suggest we keep the issue open and have a separate pull request for the links to the anchor points.  
So happy if @lloyddewit or @shadrackkibet could also check and merge.

lloyddewit (approved):

PR #6973, opened 2021-11-22
Changed the increment in the acceptable range controls in the Climatic Check Data Temperature dialog (plus other small enhancements)

### LINKED ISSUE #6971: Small improvement to the Climatic > Check Data > QC Temp  dialogue
This is a trivial, but (for me) quite an important change.  In the temperature 
image

The acceptable range up-down controls go in integer steps.  The others shown here go in steps of 0.1.  Please use 0.1 also for the 2 acceptable range controls.

At the same time at the top get rid of the disabled Satelite label - which is also spelt wrongly.  And for the Multiple, keep it disabled but add the tooltip "Not yet implemented"

N-thony:
Fixes #6971
@africanmathsinitiative/developers  @rdstern I made a small improvement as suggested in the corresponding issue apart from the tooltip on the Multiple radio button that seems to not work yet since the control is disabled. Please take a look. Thanks.


rdstern (approved):
Great!

lloyddewit (approved):

PR #6974, opened 2021-11-22
Bug fixed when reading data sets from library

### LINKED ISSUE #6969: Reading more data sets from the R library
My latest example is the climatol package I have added recently to evaluate.  It has a number of datasets.  Here is a snapshot.
image

You will notice that there is a name for the file and then another name in the brackets.  Currently the code to read the first file is:

```
utils::data(package="climatol", X=dah)

data_book$import_data(data_tables=list(dah=dah))
```
This gives an error and fails to read.  I can change it to read the name in the brackets, as follows:
```
# Code run from Script Window
utils::data(package="climatol", X=pcp_results)

data_book$import_data(data_tables=list(pcp_results=dah))
```
This imports fine.
I suggest changing the reading to the name in the brackets, as above, when it finds a bracket.  (I am slightly confused that it uses both parts of the name - but then my R is really flaky!

N-thony:
Fixes #6969
@rdstern this fixes the bug that was appearing when trying to load the data library especially in the case described in the corresponding issue. Could you take a look? Thanks.

### DIFF
--- a/instat/dlgFromLibrary.vb
+++ b/instat/dlgFromLibrary.vb
@@ -265,7 +265,6 @@ Public Class dlgFromLibrary
         strSelectedDataNameFromBracket = GetStringInBracket(strSelectedText)
         strSelectedDataName = CheckString(strSelectedText)
 
-        ' strSelectedDataName = CheckString(lstCollection.SelectedItems(0).SubItems(0).Text)
         If strSelectedText.Contains("(") Then
             clsDataFunction.AddParameter("X", strSelectedDataNameFromBracket)
         Else
--- a/instat/dlgFromLibrary.vb
+++ b/instat/dlgFromLibrary.vb
@@ -210,21 +210,21 @@ Public Class dlgFromLibrary
     End Sub
 
     Private Function CheckString(ByVal strValue As String)
-        Dim strLength As Integer = strValue.IndexOf(" ")
-        If strLength = -1 Then
+        Dim iLength As Integer = strValue.IndexOf(" ")
+        If iLength = -1 Then
             Return strValue
         Else
-            Return strValue.Substring(0, strLength)
+            Return strValue.Substring(0, iLength)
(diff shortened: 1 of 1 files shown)


instat/dlgFromLibrary.vb
```diff
@@ -218,6 +218,16 @@ Public Class dlgFromLibrary
         End If
     End Function
 
+    Private Function GetStringInBracket(ByRef strValue As String)
+        Dim startIndex As Integer = strValue.IndexOf("(") + 1
+        Dim strLength As Integer = strValue.IndexOf(")") - startIndex
```
lloyddewit, line 223: R-Instat naming convention for integers is `iStartIndex` etc.

N-thony:
> Fixes #6969 @rdstern this fixes the bug that was appearing when trying to load the data library especially in the case described in the corresponding issue. Could you take a look? Thanks.

lloyddewit (approved):

instat/dlgFromLibrary.vb
N-thony: @lloyddewit I have made the change.

instat/dlgFromLibrary.vb
```diff
@@ -218,6 +218,16 @@ Public Class dlgFromLibrary
         End If
     End Function
 
+    Private Function GetStringInBracket(ByRef strValue As String)
+        Dim iStartIndex As Integer = strValue.IndexOf("(") + 1
+        Dim strLength As Integer = strValue.IndexOf(")") - iStartIndex
```
lloyddewit, line 223: naming convention?

lloyddewit (changes requested):

lloyddewit (approved):

rdstern (approved):
Nice.  I also tested on the survey package that has a similar situation and they now read ok too.

PR #6979, opened 2021-11-23
Function definition for specific R Functions 

### LINKED ISSUE #6853: For a specified list of R function names, return the corresponding function definitions
**Problem description**
If we want to populate dialogs from R script (see PR #6407), then we need to be able to correctly decode R function calls. These function calls may contain named and unnamed parameters, and the order is flexible. We need to be able to match parameters in the function call to parameters in the function definition.

The purpose of this GitHub issue is to write a function that supports the above. The function should, for a list of function names, return the corresponding function definitions.

**Proposed solution**
Please add the following function to ‘clsRLink’:
```
Private Function GetFunctionDefinitions(lstFunctionNames As List(Of String)) As Dictionary(Of String, String)
```
For example:
-	Input: (“readRDS”, “list”, “sign”, “ggplot2::ggplot”)
-	Returns: (“readRDS”:” file, refhook = NULL”, “list”:”...”, “sign”:”x”, “ggplot2::ggplot”:” data = NULL, mapping = aes(), ..., environment = parent.frame()”)

If, for a specified name, R cannot return a function definition, then the dictionary value is set to an empty string. For example, “myNonexistentFn” returns “myNonexistentFn”:””

**Additional context**
In R, if you enter the function name without parameters, then R displays the function definition.
For example:
````
readRDS
function (file, refhook = NULL) 
{
    if (is.character(file)) {
        :

list
function (...)  .Primitive("list")

sign
function (x)  .Primitive("sign")

ceiling
function (x)  .Primitive("ceiling")

ggplot2::ggplot
function (data = NULL, mapping = aes(), ..., environment = parent.frame()) 
{
  :
````

### PR DESCRIPTION
**Pull request by Ivanluv**
fixes #6853 and  replaces #6885
@lloyddewit  this is ready

### CODE REVIEW

File: instat/clsRLink.vb [hunk 1]
```diff
@@ -2377,4 +2378,101 @@ Public Class RLink
...
+    '''             clsParameter 4: Argument Name="simplify" , Parameter position = 3 , Default value = "FALSE")
+    '''     </description></item>
+    ''' </list></summary>
+    ''' 
+    ''' <param name="strFunctionName">  The function name provided. </param>
+    '''
+    ''' <returns>   The list of clsRParameters. </returns>
+    '''--------------------------------------------------------------------------------------------
+    Private Function GetRFunctionDefinitionParameters(strFunctionName As String) As List(Of clsRParameter)
+        'temporary object that retrieves the output from the environment
```
lloyddewit, line 2405:
```suggestion
        'Note: this function is not currently called but it will be used in future
        '      functionality to populate dialogs from script.
        '      Please do not delete this function. (@lloyddewit 24/11/21)
        
        'temporary object that retrieves the output from the environment
```

lloyddewit:
@rdstern This PR contains functionality required in the future to populate dialogs from scripts. There's nothing you can test. Please could you approve this PR anyway so that we can merge?
Thanks

rdstern (approved):

PR #6985, opened 2021-11-25
Improved the Climatic > Prepare > Conversions dialog

### LINKED ISSUE #6970: Improve the Climatic > Prepare > Conversions dialogue
An example is the Niger data, in the Instat library, with the station information.  It is given in degrees and minutes.
Here is the dialogue to convert into fractions of a degree.  
There are 2 improvements I would like:
a) Make the final entry of N, S, E, W much simpler.  Currently it needs a variable with (usually) a single letter throughout.  I would prefer a Direction groupbox with 4 radio buttons, namely o N o S o E o W and default N.  Possibly (if easy) there are 5 with the fifth being 
o Variable.  If you choose Variable, then it is as now, i.e. you need to give the variable.

b) Less important, but I envisage simple use of the dialogue could be for a single Value.  So, at the top there could be a checkbox with the name Variable.  Default is checked.  If unchecked, then it expects a single Value and then you type that value into the field, rather than the variable.  In that case the result goes into the output window, rather than a column.  (I have a feeling this is going to be more work than it is worth however.)

N-thony:
Fixes #6970
@africanmathsinitiative/developers @rdstern I have made the improvement regarding the corresponding issue. This is ready for review. Thanks.


lloyddewit:
@HawardKetoyoMsatsi Please could you review?
Thanks

rdstern (changes requested):
@N-thony that's great - it all works already, including just a single number.
So just some cosmetics on that option
a) The saving is still there, even though it isn't used.  So the bottom line should be hidden.
b) Similarly hide the option to use a variable in the set of N,S,E,W options.  You can't have a variable there when you are typing the lat or long coordinates.
c) I tried typing 6 degrees and 20 minutes and it gives many decimals in the output.  i.e. -6.3333333333.  I couldn't find the function to see whether there is an argument to limit the precision.  If not, then you could issue the command as round(   ,3) say.

N-thony:
@rdstern I resolved your comments above and made the change on the use of the input combo for direction as we discussed this morning. Please take a look. Thanks.

rdstern (dismissed):
Looks ok now.  Great!

lloyddewit:
@HawardKetoyoMsatsi reminder: please can you review? thanks

instat/dlgConversions.vb
```diff
@@ -328,7 +437,7 @@ Public Class dlgConversions
...
+    Private Sub ucrPnlConversions_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrPnlConversions.ControlContentsChanged, ucrReceiverElement.ControlContentsChanged, ucrReceiverDate.ControlContentsChanged, ucrNudDecimal.ControlContentsChanged, ucrPnlLatitude.ControlContentsChanged, ucrInputLatitude.ControlContentsChanged, ucrReceiverLatitude.ControlContentsChanged, ucrInputFromPrecipitation.ControlContentsChanged, ucrInputToPrecipitation.ControlContentsChanged, ucrInputFromTemperature.ControlContentsChanged, ucrInputToTemperature.ControlContentsChanged, ucrInputFromWindSpeed.ControlContentsChanged, ucrInputToWindSpeed.ControlContentsChanged, ucrPnlElements.ControlContentsChanged, ucrReceiverDegrees.ControlContentsChanged, ucrChkVariable.ControlContentsChanged, ucrInputDegree.ControlContentsChanged
```
HawardKetoyoMsatsi, line 440: Add ucrSaveConversions.ControlContentChanged here?

HawardKetoyoMsatsi (changes requested):

instat/dlgConversions.vb
N-thony: @HawardKetoyoMsatsi thanks, I resolved it. Please take a look.

instat/dlgConversions.vb
```diff
@@ -255,8 +283,12 @@ Public Class dlgConversions
             End If
         ElseIf rdoDayLength.Checked AndAlso Not ucrReceiverDate.IsEmpty AndAlso ((rdoSingleValue.Checked AndAlso Not ucrInputLatitude.IsEmpty) OrElse (rdoColumn.Checked AndAlso Not ucrReceiverLatitude.IsEmpty)) Then
             ucrBase.OKEnabled(True)
-        ElseIf rdoCoordinates.Checked AndAlso Not ucrReceiverDegrees.IsEmpty AndAlso ucrSaveConversions.IsComplete Then
-            ucrBase.OKEnabled(True)
+        ElseIf rdoCoordinates.Checked Then
+            If (Not ucrReceiverDegrees.IsEmpty AndAlso ucrSaveConversions.IsComplete AndAlso ucrChkVariable.Checked) OrElse (Not ucrChkVariable.Checked AndAlso Not ucrInputDegree.IsEmpty) Then
```
lloyddewit, line 287: Split over multiple lines so that we don't exceed 100 colums?
Please could you do the same for line 284 (even though it's not your code)? thanks

lloyddewit (changes requested):

instat/dlgConversions.vb
N-thony: Thanks @lloyddewit, I resolved this.

lloyddewit (approved):

lloyddewit:
@rdstern Since you last approved, there has just been a couple of small changes that should not have affected the functionality. Please could you test/approve again? thanks

rdstern (approved):
Looks good

PR #6992, opened 2021-11-29
fixed convert to numeric form closing issue and updated install scripts

dannyparsons:
@ChrisMarsh82 I noticed an issue with the new closing on `frmConvertToNumeric`. This is because the `DialogResult` property seems to be lost when the form is closed. I've changed it so that the form is closed by `ucrDataView` instead. This seems to work ok. Can you check you're happy with this?

(I also included updates to the installation scripts here, but that's unrelated to this change)


dannyparsons:

lloyddewit (approved):

shadrackkibet (approved):
I tested and I didn't find anything suspicious.

PR #6994, opened 2021-11-29
Implemented the regex sub dialog under Prepare > Column:Text > Split dialogue

### LINKED ISSUE #6893: Add regex sub-dialogue
@N-thony I am not sure whether this will work out, but it is worth a try.
a) Change the expressions to expression in the dialogues, and the checkbox to a regex button.  If you press the button then you get a sub-dialogue - just like the Position button.
b) The heading is Construct regex expression.  Then there is a field into which you can type.  Underneath is the keyboard (together with the tooltips) from the calculator.  Initially just the 3 by 5 block at the bottom of the character keyboard:
image

Then just Return and Help as usual.  The help context number is 89.  
When you press Return it copies the field, from the sub-dialogue, into the pattern. control on the main dialogue.

It is worth a try!

N-thony:
Fixes #6893 
@rdstern I have implemented the regex construction in its sub dialog and used that as an example in the corresponding issue of this PR as requested. Please could you take look? Thanks.


N-thony:
@rdstern I have added the use of regex sub-dialogue on the Find/Replace dialogue to facilitate the tests as we discussed, please take a look. Thanks.

lloyddewit:
@N-thony thanks for theis PR. I noticed that there are 66 files changed in this PR. Are all these changes intentional, or are some a side-effect of the recent PRs that were merged for the Mono work?

N-thony:
> @N-thony thanks for this PR. I noticed that there are 66 files changed in this PR. Are all these changes intentional, or are some a side-effect of the recent PRs that were merged for the Mono work?

@lloyddewit you are correct. I was trying to load the Twitter dataset and test the new changes in the Split dialogue but I couldn't then I noticed some improvement in the remote master branch and merged it into this local branch and that allowed me to load the dataset.

rdstern (changes requested):
@N-thony this is looking good for now.
a) When the keyboard is visible then the user types into the field there.  There is already the pull-down.  Now in the calculator (and some other dialogues) this pull-down "remembers" the previous entries.  It would be good to add this feature here.
b) There could usefully be an Ignore Case checkbox.  Default is checked.  This is always available.  It is an argument of either Fixed, or regex.
c) There are (at least) 3 other controls that become visible when regex is chosen. The first is Multiline and the second is Comments.  You may choose to put them on the same row in the dialogue, so it does not become too deep.  They are both checkboxes, and default is un checked.
d) Then (perhaps all together on one row is a checkbox called Boundary If checked this opens a pull-down with (I think) 4 options.  Character is first and is the default.  This adds the boundary argument.  I think this may be different in being available anyway - i.e. like the case checkbox it is available whether, or not, the regex option is chosen?
e) Finally did you find an example (or two) of data to illustrate and test these features?  Perhaps you could report?

N-thony:
e) Finally did you find an example (or two) of data to illustrate and test these features? Perhaps you could report?
@rdstern yes, I tried the regex `"@\\S+"` under the Extract option in the Find/Replace dialogue to extract the text in the tweets starting with the symbol `@` and it works fine. I just loaded the 13 first rows of the dataset for the test and better see.

Here is the dataset I used [archive.zip](https://github.com/africanmathsinitiative/R-Instat/files/7674195/archive.zip)

lloyddewit:
@N-thony There's still the issue with the 66 changed files. Are you able to resolve this using Git or is it easier to make a new PR?
It may be good to resolve this before @rdstern approves, else he'll need to retest after the changes.
Thanks

N-thony:
> @N-thony There's still the issue with the 66 changed files. Are you able to resolve this using Git or is it easier to make a new PR? It may be good to resolve this before @rdstern approves, else he'll need to retest after the changes. Thanks

@lloyddewit I will open a new one that will replace this one.

lloyddewit:
Replaced by PR #7062

PR #7033, opened 2021-12-02
Added column selection feature into the right clicks

### LINKED ISSUE #7034: Adding column selection into rights clicks

shadrackkibet:
fixes #7034

@ChrisMarsh82 here are the changes I have made while adding column selection into the right clicks for Data View and Column Metadata. Works well on windows from my end.  I am not sure what I should do on the Linux side of things.

I am looking forward to your guidance. We can do a quick call if that is much more efficient.

Thanks.


instat/Model/DataFrame/clsDataFrameFilter.vb
```diff
@@ -64,9 +71,17 @@ Public Class clsDataFrameFilter
         Return _RLink.RunInternalScriptGetValue(clsFilterApplied.ToScript()).AsLogical(0)
     End Function
 
+    Private Function GetColumnSelectionAppliedFromRCommand() As Boolean
+        Dim clsColumnSelectionApplied As New RFunction
+        clsColumnSelectionApplied.SetRCommand(_RLink.strInstatDataObject & "$column_selection_applied")
+        clsColumnSelectionApplied.AddParameter("data_name", Chr(34) & _strDataFrameName & Chr(34))
+        Return _RLink.RunInternalScriptGetValue(clsColumnSelectionApplied.ToScript()).AsLogical(0)
+    End Function
```
shadrackkibet, line 79: @ChrisMarsh82 I was a bit in a catch 22 situation here. Since we have Column Selections and Filters. Should we create a new class for column selections since we have `clsDataFrameFilter` or we rename this class to include column selections?

ChrisMarsh82:
@shadrackkibet looks good from the initial look of it and the way you have done it means it shouldn't be any different in Linux. 
I think putting it in the filter will work well. I need to check a few things that I should be able to do Monday. 

Question: I can't see where it would be displayed to users that they are looking at a subset of columns. Should this be added to the lblColDisplay in SetDisplayLabels?

shadrackkibet:
@ChrisMarsh82  thanks for looking into this.

> Question: I can't see where it would be displayed to users that they are looking at a subset of columns. Should this be added to the lblColDisplay in SetDisplayLabels?

This is not yet ready. I am tweaking the R method for this. I will alert you once it is ready.

instat/Model/DataFrame/clsDataFrameFilter.vb
ChrisMarsh82: Should this be changed to RowFilterApplied and ColumnFilterApplied? They can both sit in clsDataFrameFilter or you could create clsDataframeRowFilter and clsDataframeColumnFilter.  Once you have the R commands working it would be easier to see what the best route would be. I assume both filters would work in a very similar way?

rdstern:
@shadrackkibet is anything more needed, before I test.  I am keen to continue progress on select, including applying it to the grid.

rdstern (changes requested):
@shadrackkibet seems to work and looks good.  
The only minor change is that the term Remove Current Column Selection is so long that it has made the whole right-click menu wider.  I suggest Remove Column Selection is fine.  Or it could be Remove Current Selection, but I think I prefer the former.  Then it could be merged.

rdstern (approved):
Looks good.  I hope this can be merged soon?

ChrisMarsh82 (approved):

PR #7038, opened 2021-12-06
In Open From Library dialog, open R script file and write to the Script Window, also added Niger script file

### LINKED ISSUE #6984: Opening R scripts from R-Instat library into script window
[Niger_4stations_corrections.zip](https://github.com/africanmathsinitiative/R-Instat/files/7601052/Niger_4stations_corrections.zip)
This is a script file with the R extension.  
It makes corrections to the Niger_4station data.
I would like it to be included in the R-Instat library.
Then the other addition is to be able to open files, from the library, with an R extension, into the script window.  Currently we only have data files that open into a data frame.

The code for this is (almost) the same as when in the script window we can right-click and **_Load Script from File_**.  We just need this code.  So if the extension is R then it does that and if anything else it continues with the ordinary input.

I notice that the **_Load Script from File_** `always`  warns you that it will clear the current contents.  It does this even if there are no contents.  It would be good to give that warning only when the script was not empty.  And that applies also to the current command.

N-thony:
Fixes #6984 
This PR adds the Niger script file to the Climatic > Niger folder and also the facility to open an R script to the Script Window from Open From Library.

### DIFF
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -686,13 +686,13 @@ Public Class dlgImportDataset
                               "Load Script From File", MessageBoxButtons.YesNo) = DialogResult.Yes Then
                     Try
                         frmMain.ucrScriptWindow.txtScript.Text = File.ReadAllText(strFilePathSystem)
-                        SetDialogStateFromFile("")
                     Catch
                         MessageBox.Show("Could not load the script from file." &
                               Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.",
                               "Load Script", MessageBoxButtons.OK, MessageBoxIcon.Error)
                     End Try
                 End If
+                SetDialogStateFromFile("")
                 Me.Close()
             End If
         End If
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -569,7 +569,7 @@ Public Class dlgImportDataset
                 strlastFileName = Path.GetFileNameWithoutExtension(strFileOrFolderPath)
                 strCurrentDirectory = Path.GetDirectoryName(strFileOrFolderPath)
                 strFileExtension = Path.GetExtension(strFileOrFolderPath).ToLower 'extension check is done in lower case
-            ElseIf Directory.Exists(strFileOrFolderPath) AndAlso strFolderFileExt <> "" Then
+                            ElseIf Directory.Exists(strFileOrFolderPath) AndAlso strFolderFileExt <> "" Then
                 strCurrentDirectory = strFileOrFolderPath
                 strFileExtension = strFolderFileExt.ToLower 'extension check is done in lower case
                 bImportFromFolder = True
@@ -685,7 +685,8 @@ Public Class dlgImportDataset
                               Environment.NewLine & "Do you still want to load?",
                               "Load Script From File", MessageBoxButtons.YesNo) = DialogResult.Yes Then
                     Try
-                        frmMain.ucrScriptWindow.txtScript.Text = File.ReadAllText(strFileOrFolderPath)
+                        frmMain.ucrScriptWindow.txtScript.Text = File.ReadAllText(strFilePathSystem)
+                        SetDialogStateFromFile("")
                     Catch
                         MessageBox.Show("Could not load the script from file." &
                               Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.",
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -691,8 +691,8 @@ Public Class dlgImportDataset
                               Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.",
                               "Load Script", MessageBoxButtons.OK, MessageBoxIcon.Error)
                     End Try
-                    Me.Close()
                 End If
+                Me.Close()
             End If
         End If
 
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -660,25 +660,16 @@ Public Class dlgImportDataset
                     frmMain.mnuViewScriptWindow.Checked = True
                     frmMain.UpdateLayout()
                 End If
-                If frmMain.ucrScriptWindow.txtScript.TextLength > 0 Then
-                    If MessageBox.Show("Loading a script from file will clear your current script" & Environment.NewLine & "Do you still want to load?",
+                If frmMain.ucrScriptWindow.txtScript.TextLength = 0 OrElse MessageBox.Show("Loading a script from file will clear your current script" & Environment.NewLine & "Do you still want to load?",
                      "Load Script From File", MessageBoxButtons.YesNo) = DialogResult.Yes Then
-                        Try
-                            frmMain.ucrScriptWindow.txtScript.Text = File.ReadAllText(strFileOrFolderPath)
-                        Catch
-                            MessageBox.Show("Could not load the script from file." & Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.", "Load Script", MessageBoxButtons.OK, MessageBoxIcon.Error)
-                        End Try
-                    End If
-                Else
                     Try
                         frmMain.ucrScriptWindow.txtScript.Text = File.ReadAllText(strFileOrFolderPath)
                     Catch
                         MessageBox.Show("Could not load the script from file." & Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.", "Load Script", MessageBoxButtons.OK, MessageBoxIcon.Error)
                     End Try
+                    Me.Close()
                 End If
-                Me.Close()
             End If
-
         End If
 
         TryTextPreview()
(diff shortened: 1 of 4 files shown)


instat/dlgImportDataset.vb
```diff
@@ -655,6 +655,30 @@ Public Class dlgImportDataset
...
+                    frmMain.UpdateLayout()
+                End If
+                If frmMain.ucrScriptWindow.txtScript.TextLength > 0 Then
+                    If MessageBox.Show("Loading a script from file will clear your current script" & Environment.NewLine & "Do you still want to load?",
+                     "Load Script From File", MessageBoxButtons.YesNo) = DialogResult.Yes Then
+                        Try
+                            frmMain.ucrScriptWindow.txtScript.Text = File.ReadAllText(strFileOrFolderPath)
+                        Catch
+                            MessageBox.Show("Could not load the script from file." & Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.", "Load Script", MessageBoxButtons.OK, MessageBoxIcon.Error)
+                        End Try
```
lloyddewit, line 670: This try block is duplicated below. Remove the duplication?
(there are various ways you could do this)

instat/dlgImportDataset.vb
```diff
@@ -655,6 +655,30 @@ Public Class dlgImportDataset
                 ucrSaveFile.SetName(GetCleanFileName(strFileName), bSilent:=True)
             End If
 
+            If strFileExtension = ".r" Then
```
lloyddewit, line 658: general comment:
Are multiple imports allowed for `'r` files (see line 601)

instat/dlgImportDataset.vb
```diff
@@ -655,6 +655,30 @@ Public Class dlgImportDataset
...
+                        End Try
+                    End If
+                Else
+                    Try
+                        frmMain.ucrScriptWindow.txtScript.Text = File.ReadAllText(strFileOrFolderPath)
+                    Catch
+                        MessageBox.Show("Could not load the script from file." & Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.", "Load Script", MessageBoxButtons.OK, MessageBoxIcon.Error)
+                    End Try
+                End If
+                Me.Close()
```
lloyddewit, line 679: The dialog isn't closed for other file formats, whats the reason for doing it here?

instat/dlgImportDataset.vb
```diff
@@ -655,6 +655,30 @@ Public Class dlgImportDataset
...
+                    frmMain.mnuViewScriptWindow.Checked = True
+                    frmMain.UpdateLayout()
+                End If
+                If frmMain.ucrScriptWindow.txtScript.TextLength > 0 Then
+                    If MessageBox.Show("Loading a script from file will clear your current script" & Environment.NewLine & "Do you still want to load?",
+                     "Load Script From File", MessageBoxButtons.YesNo) = DialogResult.Yes Then
+                        Try
+                            frmMain.ucrScriptWindow.txtScript.Text = File.ReadAllText(strFileOrFolderPath)
+                        Catch
+                            MessageBox.Show("Could not load the script from file." & Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.", "Load Script", MessageBoxButtons.OK, MessageBoxIcon.Error)
```
lloyddewit, line 669: break up over several lines?

instat/ucrScript.vb
```diff
@@ -158,12 +158,20 @@ Public Class ucrScript
...
             If dlgLoad.ShowDialog() = DialogResult.OK Then
-                If MessageBox.Show("Loading a script from file will clear your current script" & Environment.NewLine & "Do you still want to load?",
+                If txtScript.TextLength > 0 Then
+                    If MessageBox.Show("Loading a script from file will clear your current script" & Environment.NewLine & "Do you still want to load?",
                                    "Load Script From File", MessageBoxButtons.YesNo) = DialogResult.Yes Then
+                        Try
+                            txtScript.Text = File.ReadAllText(dlgLoad.FileName)
+                        Catch
+                            MessageBox.Show("Could not load the script from file." & Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.", "Load Script", MessageBoxButtons.OK, MessageBoxIcon.Error)
+                        End Try
```
lloyddewit, line 168: remove the duplication?

instat/dlgImportDataset.vb
N-thony: @lloyddewit thanks for the quick call. We agreed to leave it as it is for now.

N-thony:
@lloyddewit I have resolved your comments above. Please take a look. Thanks

instat/dlgImportDataset.vb
```diff
@@ -676,6 +676,21 @@ Public Class dlgImportDataset
                 ucrSaveFile.SetName(GetCleanFileName(strFileName), bSilent:=True)
             End If
 
+            If strFileExtension = ".r" Then
+                If Not frmMain.mnuViewScriptWindow.Checked Then
+                    frmMain.mnuViewScriptWindow.Checked = True
+                    frmMain.UpdateLayout()
+                End If
+                If frmMain.ucrScriptWindow.txtScript.TextLength = 0 OrElse MessageBox.Show("Loading a script from file will clear your current script" & Environment.NewLine & "Do you still want to load?",
+                     "Load Script From File", MessageBoxButtons.YesNo) = DialogResult.Yes Then
```
lloyddewit, line 685: A line should not exceed 100 columns.
Add some newlines to the `If` statement?

instat/dlgImportDataset.vb
```diff
@@ -676,6 +676,21 @@ Public Class dlgImportDataset
...
+                    Try
+                        frmMain.ucrScriptWindow.txtScript.Text = File.ReadAllText(strFileOrFolderPath)
+                    Catch
+                        MessageBox.Show("Could not load the script from file." & Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.", "Load Script", MessageBoxButtons.OK, MessageBoxIcon.Error)
```
lloyddewit, line 689: A line should not exceed 100 columns.
Add some newlines?

instat/dlgImportDataset.vb
lloyddewit: Yes, this function contains functionality that is not relevent for R script imports.
Part of the problem is that the Import dialogwas designed for importing datasets. But we are using it to populate the script window. This is not a natural fit and I think it reduces the understandability of the code.
We could refactor dlgImportDataset but I think this is too much work.
On balance, I think your implementation is an OK option for now.

lloyddewit (changes requested):
Thanks, just a couple of small suggestions for readability

N-thony:
@lloyddewit I resolved your comment above. Please take a look. Thanks.

rdstern (changes requested):
I first tried with the script window empty and it worked fine.  It also didn't ask me - which it used to - and was unnecessary.
So - good so far.
Then I tried with a second file and with something already in the script window.  It warned me - that's good too.
So I said no  (I realise I should have tested yes as well, but didn't.)  Anyway after no it returned me to the ordinary Import file dialogue with that R file.  So I pressed Ok and it gave me an error.

Two points follow from this:
a) It would be good if the Import dialogue could also recognise an R-file and put it into the script window.  Currently the only way is via right-click once in the script window.
b) A separate point is whether it should go to the import dialogue when I say No to overwriting the existing file in the script window.  I'm not sure about this.

N-thony:
> I first tried with the script window empty and it worked fine. It also didn't ask me - which it used to - and was unnecessary. So - good so far. Then I tried with a second file and with something already in the script window. It warned me - that's good too. So I said no (I realise I should have tested yes as well, but didn't.) Anyway after no it returned me to the ordinary Import file dialogue with that R file. So I pressed Ok and it gave me an error.
> 
> Two points follow from this: a) It would be good if the Import dialogue could also recognise an R-file and put it into the script window. Currently the only way is via right-click once in the script window. b) A separate point is whether it should go to the import dialogue when I say No to overwriting the existing file in the script window. I'm not sure about this.

@rdstern I resolved your comments above, please take a look. Thanks,

rdstern (changes requested):
@N-thony this now works perfectly - though with a big "but"!  
As far as I have tested the File > Open From Library seems to work ok.  (As I write this I realis that I haven't tested fully when you load a sequence of R files and then a data file and then an R file again.  So that will be good to check.
But, now the new code interacts with the File > Input from file dialogue, and in a pretty fundamental way.
a) I had a file in the script window.  When I use File > Import File (instead of Input from library), it gives me the warning that I would expect from the Input from Library dialogue.  
b) So I cleared the script window and tried File > Import again.  Then it returned the previous script automatically!
Now I would also like the File > Import File dialogue to be able to import a file into the script window.  I assumed this would be a separate issue, but perhaps it may be simpler to add this feature to the File > Import dialogue now???

N-thony:
@rdstern the change I made temporarily resolved your comments above. With a possibility to import an r script From both File and Library.

rdstern (approved):
@N-thony great - this seems to work now.

lloyddewit (approved):

PR #7049, opened 2021-12-07
Fixes #7043 Added clear selection before adding columns as causes issue on linux

### LINKED ISSUE #7043: Using dialogs from the right click menus in Linux version crashes R-Instat
In the Linux version:

- Open airquality data from the library
- Right click a column header and rename
- Change the name to anything. Click Ok.
- R-Instat closes without error or message

ChrisMarsh82:
The issue is with updating grids when the column is selected.


dannyparsons (approved):
Looks good to me.

Is there going to be an easy way we can test PRs on Linux before merging?

PR #7059, opened 2021-12-08
Dynamically resize info text in Data View

### LINKED ISSUE #7008: Data Grid info text doesn't fit in display
When I had a filter, the info text was too long to fit in the Data View unless I make it bigger than half the screen. Can we reduce the text size or have it dynamically reduce?

image

N-thony:
Fixes #7008 
@africanmathsinitiative/developers @ChrisMarsh82 @dannyparsons this is ready for review. Thanks.

### DIFF
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -224,9 +224,25 @@ Public Class ucrDataView
             SetDisplayLabels()
             UpdateNavigationButtons()
             SetGridVisibility(True)
+            ResizeLabe()
         Else
             frmMain.tstatus.Text = GetTranslation("No data loaded")
             SetGridVisibility(False)
+        End If
+    End Sub
+
+    Private Sub ResizeLabe()
+        If lblRowDisplay.Width < TextRenderer.MeasureText(lblRowDisplay.Text, lblRowDisplay.Font).Width OrElse
+            lblRowDisplay.Width > TextRenderer.MeasureText(lblRowDisplay.Text, lblRowDisplay.Font).Width OrElse
+            lblRowDisplay.Width = TextRenderer.MeasureText(lblRowDisplay.Text, lblRowDisplay.Font).Width Then
+            lblRowDisplay.Font = New Font(lblRowDisplay.Font.FontFamily, lblRowDisplay.Font.Size - 0.5F, lblRowDisplay.Font.Style)
+            lblRowDisplay.Width = TextRenderer.MeasureText(lblRowDisplay.Text, lblRowDisplay.Font).Width
+        End If
+        If lblColDisplay.Width < TextRenderer.MeasureText(lblColDisplay.Text, lblColDisplay.Font).Width OrElse
+            lblColDisplay.Width > TextRenderer.MeasureText(lblColDisplay.Text, lblColDisplay.Font).Width OrElse
+            lblColDisplay.Width = TextRenderer.MeasureText(lblColDisplay.Text, lblColDisplay.Font).Width Then
+            lblColDisplay.Font = New Font(lblColDisplay.Font.FontFamily, lblColDisplay.Font.Size - 0.5F, lblColDisplay.Font.Style)
+            lblColDisplay.Width = TextRenderer.MeasureText(lblColDisplay.Text, lblColDisplay.Font).Width
         End If
     End Sub
 
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -232,22 +232,13 @@ Public Class ucrDataView
     End Sub
 
     Private Sub ResizeLabels()
-        Dim minSize As Single = 2
-        Dim maxSize As Single = 12
+        Dim minSize As Single = 7
+        TblPanPageDisplay.Font = New Font(TblPanPageDisplay.Font.FontFamily, 12, TblPanPageDisplay.Font.Style)
 
-        If lblRowDisplay.Width + TextRenderer.MeasureText(lblRowDisplay.Text, New Font(lblRowDisplay.Font.FontFamily, lblRowDisplay.Font.Size, lblRowDisplay.Font.Style)).Width > tlpTableContainer.Width Then
-            While lblRowDisplay.Width + TextRenderer.MeasureText(lblRowDisplay.Text, New Font(lblRowDisplay.Font.FontFamily, lblRowDisplay.Font.Size, lblRowDisplay.Font.Style)).Width > tlpTableContainer.Width AndAlso
-              lblRowDisplay.Font.Size > minSize
-                lblRowDisplay.Font = New Font("Microsoft Sans Serif", lblRowDisplay.Font.Size - 0.5F, lblRowDisplay.Font.Style)
-                lblColDisplay.Font = New Font("Microsoft Sans Serif", lblRowDisplay.Font.Size - 0.5F, lblRowDisplay.Font.Style)
-            End While
-        Else
-            While lblRowDisplay.Width + TextRenderer.MeasureText(lblRowDisplay.Text, New Font(lblRowDisplay.Font.FontFamily, lblRowDisplay.Font.Size, lblRowDisplay.Font.Style)).Width < tlpTableContainer.Width AndAlso
-          lblRowDisplay.Font.Size < maxSize
-                lblRowDisplay.Font = New Font("Microsoft Sans Serif", lblRowDisplay.Font.Size + 0.5F, lblRowDisplay.Font.Style)
-                lblColDisplay.Font = New Font("Microsoft Sans Serif", lblRowDisplay.Font.Size + 0.5F, lblRowDisplay.Font.Style)
-            End While
-        End If
+        While lblRowDisplay.Width + lblColDisplay.Width + lblColBack.Width + lblColFirst.Width + lblColLast.Width + lblColNext.Width + lblRowBack.Width +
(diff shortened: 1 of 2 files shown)


N-thony:
> Fixes #7008 @africanmathsinitiative/developers @ChrisMarsh82 @dannyparsons this is ready for review. Thanks.

Please wait a bit before the review. I will let you know, I found a small bug.

ChrisMarsh82:
@N-thony  If you delete the font from the label controls you then can control the font from the TblPanPageDisplay. You will also need to check the size in a loop so that the size can be decremented more than once. 
ResizeLabel should be called within SetDisplayLabels as this makes sure the labels are being resized every time they are changed.

ChrisMarsh82:
Here is a link to the type loop you will need. Its in c# but you get the idea
https://stackoverflow.com/questions/9527721/resize-text-size-of-a-label-when-the-text-gets-longer-than-the-label-size

dannyparsons:
Could you not calculate the size needed once instead of incrementally decreasing? - that seems inefficient, especially if it's done every time there's a refresh. 

The link from Chris links to another answer that does this using the ratios https://stackoverflow.com/questions/9527721/resize-text-size-of-a-label-when-the-text-gets-longer-than-the-label-size/20457976#20457976 Can we use the same idea?

N-thony:
@ChrisMarsh82 @dannyparsons I have tried different approaches you have provided in the above links. I found Danny's approach a bit longer and complicated than the one from Chris with some adaptation I done, please take a look.

N-thony:
> SetDisplayLabels

@ChrisMarsh82 please could you take a look now. This seems to work. Thanks

ChrisMarsh82:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:w="urn:schemas-microsoft-com:office:word"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Word.Document>
<meta name=Generator content="Microsoft Word 15">
<meta name=Originator content="Microsoft Word 15">
<link rel=File-List
href="file:///C:/Users/CHRIST~1/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

<link rel=themeData
href="file:///C:/Users/CHRIST~1/AppData/Local/Temp/msohtmlclip1/01/clip_themedata.thmx">
<link rel=colorSchemeMapping
href="file:///C:/Users/CHRIST~1/AppData/Local/Temp/msohtmlclip1/01/clip_colorschememapping.xml">

<style>

</style>

</head>

<body lang=EN-GB style='tab-interval:36.0pt;word-wrap:break-word'>


<p class=MsoNormal><span lang=EN-US style='mso-ansi-language:EN-US'><o:p>&nbsp;</o:p></span></p>



Test | Result
-- | --
Add filter that doesn’t fit | Text gets smaller
Add filter that does fit | Text stays same
Remove filter that didn’t fit | Text goes back to original size
Remove filter that did fit | Text stays same
Re-size dialog so text doesn’t fit | Text gets smaller
Re-size dialog so text fits (when   previously didn’t) | Text goes back to original size
  |  
Text so large would never fit | Text goes to minimum size




</body>

</html>

ChrisMarsh82:
@N-thony 

This is the pseudo-code you should be basing your code on:
```
Resize Procedure
  Set Text to standard size
  While Text does not fit and Text is greater than minimum size
     Decrement text by 0.5
  End While
```


If we look at the code you have so far:
```
        If lblRowDisplay.Width + TextRenderer.MeasureText(lblRowDisplay.Text, New Font(lblRowDisplay.Font.FontFamily, lblRowDisplay.Font.Size, lblRowDisplay.Font.Style)).Width > tlpTableContainer.Width Then
            While lblRowDisplay.Width + TextRenderer.MeasureText(lblRowDisplay.Text, New Font(lblRowDisplay.Font.FontFamily, lblRowDisplay.Font.Size, lblRowDisplay.Font.Style)).Width > tlpTableContainer.Width AndAlso
              lblRowDisplay.Font.Size > minSize
                lblRowDisplay.Font = New Font("Microsoft Sans Serif", lblRowDisplay.Font.Size - 0.5F, lblRowDisplay.Font.Style)
                lblColDisplay.Font = New Font("Microsoft Sans Serif", lblRowDisplay.Font.Size - 0.5F, lblRowDisplay.Font.Style)
            End While
        Else
            While lblRowDisplay.Width + TextRenderer.MeasureText(lblRowDisplay.Text, New Font(lblRowDisplay.Font.FontFamily, lblRowDisplay.Font.Size, lblRowDisplay.Font.Style)).Width < tlpTableContainer.Width AndAlso
          lblRowDisplay.Font.Size < maxSize
                lblRowDisplay.Font = New Font("Microsoft Sans Serif", lblRowDisplay.Font.Size + 0.5F, lblRowDisplay.Font.Style)
                lblColDisplay.Font = New Font("Microsoft Sans Serif", lblRowDisplay.Font.Size + 0.5F, lblRowDisplay.Font.Style)
            End While
        End If
```
The text is not set to standard size at the start of the procedure. - This is done in the second while which then overcomplicates the procedure.

The first IF statement doesn't do anything. The WHILE clause does the same comparison and there is no code that's in the if statement that is not in the while clause. 

While Clause: 
`            While lblRowDisplay.Width + TextRenderer.MeasureText(lblRowDisplay.Text, New Font(lblRowDisplay.Font.FontFamily, lblRowDisplay.Font.Size, lblRowDisplay.Font.Style)).Width > tlpTableContainer.Width`

This is not checking the right thing. If this works it is more of luck than anything else. 
`lblRowDisplay.Width` is added to the `TextRender.MeasureText(lblRowDisplay.Text ...` which is the same value. This needs to take the width of all the labels in the `tlpPanPageDisplay`

`> tlpTableContainer.Width` is not the right panel. Although this will work as it will generally be the same size as `tlpPanPageDisplay` it is not the panel where the labels are.

````
                lblRowDisplay.Font = New Font("Microsoft Sans Serif", lblRowDisplay.Font.Size - 0.5F, lblRowDisplay.Font.Style)
                lblColDisplay.Font = New Font("Microsoft Sans Serif", lblRowDisplay.Font.Size - 0.5F, lblRowDisplay.Font.Style)
````
This should be the font on the `tlpPanPageDisplay` as it will change all the fonts including the navigation symbols '<<<>>>'

Your min size is set to 2. Is this a viable size? have you tested what this looks like when the font is set to this size?


I think you are getting close. If you go through these points one at a time I'm sure you can get it to work.

**Comment by N-thony** [comment]
@ChrisMarsh82 thanks for your review and the pseudo-code above. This seems to work now even if we had a small issue with that. As discussed yesterday for the small issue that was coming from the code when resizing I made a small change by doubling the width of the column display label in the sum of labels width in the TblPanelLayout to allow to have access to the steps that will allow the resizing of the text to fit the TblPanelLayout. The minimum size can also go to less than five to allow texts to resize and fit when collapsing the splitContainer. Please take a look. Thanks.

**Comment by N-thony** [comment]
@ChrisMarsh82 thanks for your review and the pseudo-code above. This seems to work now even if we had a small issue with that. As discussed yesterday for the small issue that was coming from the code when resizing I made a small change by doubling the width of the column display label in the sum of labels width in the TblPanelLayout to allow to have access to the steps that will allow the resizing of the text to fit the `TblPanelLayout`. The minimum size `5` can also go to less than five to allow texts to resize and fit when collapsing the splitContainer. Please take a look. Thanks.

**Comment by ChrisMarsh82** [comment]
@N-thony This looks like it works really well now. Well done!

One minor change I would like you to make. The WHILE statement has `(lblColDisplay.Width * 2)` I don't understand why we would need to take the lblColDisplay into account twice. I think what may be needed is a buffer as there are other things taking up space other than the labels. I have done some testing and a 50-pixel buffer seems to work well. 
I would change the WHILE statement to be: 
```
While lblRowDisplay.Width + lblColDisplay.Width + 50 +
                    lblColBack.Width + lblColFirst.Width + lblColLast.Width + lblColNext.Width +
                    lblRowBack.Width + lblRowFirst.Width + lblRowNext.Width + lblRowLast.Width > TblPanPageDisplay.Width AndAlso
                    TblPanPageDisplay.Font.Size > minSize
```

Also putting the WHILE statement over multiple lines would help the readability.


Other than this I think the code is good to go and works very well.

**Review by ChrisMarsh82 — DISMISSED** [review]

File: instat/ucrDataView.vb [hunk 1]
```diff
@@ -230,6 +231,18 @@ Public Class ucrDataView
         End If
     End Sub
 
+    Private Sub ResizeLabels()
+        Dim minSize As Single = 5
```
lloyddewit, line 235: I suggest to use `Const` and follow the naming conventions.
I agree that the naming conventions add nothing to readability here but I want to stay consistent with the rest of R-Instat.
```suggestion
        Dim Const iMinSize As Single = 5
```

lloyddewit (changes requested):
@N-thony This looks good, just one small suggestion

instat/ucrDataView.vb
N-thony: Thanks @lloyddewit, I resolved this

lloyddewit (approved):

lloyddewit:
@ChrisMarsh82 please can you approve again? we can then merge, thanks

ChrisMarsh82 (approved):

PR #7084, opened 2022-01-10
Fixing order of labels in the summary tables R code

lilyclements:
@rdstern @Ivanluv @dannyparsons this should fix the ordering issue outlined in the email from @rdstern 

@rdstern @Ivanluv - can you check this sorts the problem?

### DIFF
--- a/instat/static/InstatObject/R/Backend_Components/summary_functions.R
+++ b/instat/static/InstatObject/R/Backend_Components/summary_functions.R
@@ -1456,7 +1456,7 @@ DataBook$set("public", "summary_table", function(data_name, columns_to_summarise
     }
   }
   shaped_cell_values <- shaped_cell_values %>% dplyr::mutate(value = round(value, signif_fig))
-if (!is.null(shaped_cell_values$summary)){
+if (is.null(shaped_cell_values$`summary-variable`)){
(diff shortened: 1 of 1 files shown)


dannyparsons (approved):

Ivanluv:
@lilyclements  I have tried this
image
  which produces the the following script 
image
 and the table does not seem to be ordered
image

lilyclements:
@Ivanluv thanks! This should be sorted now.

Ivanluv:
This is working okay from my end

Ivanluv:
@rdstern @lloyddewit  could you review this so that we can have it merged?

rdstern:
It is different, but I now can't get anything out!  For example for a simple table with the usual survey data:
image

And these were the commands:

image

lilyclements:
@rdstern good point - this should hopefully now be fixed. Thanks!
(I am working from the office without my R-Instat laptop today so was not able to test this properly. Hopefully this is now fixed but do check!).

rdstern:
@lilyclements sorry - still get the same error as above.

lilyclements:
@rdstern how about now? It is working for me now !

I came across a bug which is that if we run mmtable2 without a factor and just one summary and one variable (and perhaps set as factor as FALSE only) then mmtable2 doesn’t work. I’ll write it up in an issue tomorrow.

rdstern:
@lilyclements here is the table I have been aiming for!  Great.

image

dannyparsons (approved):

PR #7085, opened 2022-01-10
Updated AutoScaleMode

ChrisMarsh82:
Set localizable to false Simplified the resx file.
Updated Autoscale Mode to dpi to work better on Linux
This will be done in batches to help with the merge.


dannyparsons (approved):
Tested and runs ok

Issue #7006, opened 2021-11-30
The sheets tab is hidden by the scroll bar when a dataset is loaded


dannyparsons:

dannyparsons:
@Patowhiz Chris has been looking into this and can't seem to see why the sizing of the tab buttons is ignored/reset for the Data View. It works ok for the Column Metadata Window. Could you have a look at this? Possibly comparing to the previous code before Chris's changes might help.

Patowhiz:
@dannyparsons let me have a look at it.

Patowhiz:
@dannyparsons @ChrisMarsh82 my understanding is `ucrDataView`  is now using the `TableLayoutPanel` VB.Net control to align the 3 controls that are in it;  the _launch panel, linux grid, windows grid_. Then it dynamically changes the size of these docked controls by setting its column size.

@dannyparsons this issue is caused by the dynamic change of the `TableLayoutPanel` column sizes. The `reogrid` ends up changing its `SheetTabWidth` property from **450** to **60**, thus hiding sheet names from the user. When the column size of the `TableLayoutPanel` 'housing' is changed from **0** to **100%**, the `SheetTabWidth` of the `reogrid` remains at 60 pixels, thus sheet names remains hidden.

I have fixed this in PR #7111. inside the `SetGridVisibility()` subroutine, which I think is the best place to fix this.

@ChrisMarsh82 you may have a different idea in regards to where this `SheetTabWidth` resizing should be set (the data view control has significantly changed from how I last understood its base code, you now understand it better).

@dannyparsons below is a quick code preview of the current code and what I have added to save your time.

![ucrDataview](https://user-images.githubusercontent.com/21179007/149174421-53404f6c-cadb-4143-9621-e4c9352d2165.PNG)

dannyparsons:
Thanks very much for looking into this in detail! I've replied to the PR with a suggestion, but great that we have something that works.

PR #7098, opened 2022-01-11
Solving Build errors caused by frequency Sub Dialog

Gioche6:
@africanmathsinitiative/developers , This is ready for review
@shadrackkibet , I have fixed the build errors caused by PR #6916. 
I agree they occurred while solving the merge conflict. I wonder why most PRs are getting merge conflicts. I have however fixed this one and runs okay now without affecting the performance of the dialog.
@lloyddewit , Could you test and merge if its okay to get rid of the build errors ? 
Thank you.


rdstern (approved):
Seems to work for me.

maxwellfundi:
@Dennis Gioche ***@***.***> merge conflicts are common when
two people make changes to the same document. I suggest a practice of
frequently updating your branches from master to avoid conflicts.

Patowhiz:
@lloyddewit @shadrackkibet could this be prioritized for review. We now have a 'broken' master which I had merged its changes, and I'm needed to look into issue #7006 urgently.

Thanks.

shadrackkibet (approved):
@maxwellfundi please approve and merge.

lloyddewit:
@Gioche6 There have been lots of merge conflicts in the last days. I think because large numbers of files were changed for the new Linux version. I agree with @maxwellfundi that occasional merge conflicts are unavoidable.

lloyddewit:
@shadrackkibet @maxwellfundi @Gioche6 @rdstern Thank you for resolving this so quickly.
The master branch should never be broken. To prevent this happening again, I suggest:

- The tester (normally @rdstern or his delegate) should always do a sanity check before approving a PR that was already tested/approved, but has since changed. The sanity check should include building the PR. I think it makes sense for the testers to do this check because the peer review is focussed on analysing the source code and won't always involve building the software.
- I also agree with @maxwellfundi 's comment in PR #6916 that we should normally never manually edit `.resx` or `.designer.vb` files.

dannyparsons:
Thanks Stephen this is a useful process.

Related to this, Chris recently made a PR which changed all dialogs back to `Localizable = False`. This was previous used for the old translation system but now no longer needed. This now means that all designer details will be the designer file and the resx file will remain static. This should also avoid some conflicts and issues as there will only be 2 file changes per dialog. There are no software functional changes but from now on, or if any old ones are found, keep `Localizable = False` (the VB default).

PR #7100, opened 2022-01-11
updated paths to use / instead of \ in some places for Linux version

dannyparsons:
@africanmathsinitiative/developers This is ready for review. This should not have any changes to the Windows version. Please just confirm that open from file/library still works, also from Open NetCDF and paths are set correctly.

### DIFF
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -1,4 +1,5 @@
-﻿Imports System.IO
+﻿Imports System.Runtime.InteropServices
+Imports System.IO
 Imports RDotNet
 Imports instat.Translations
 
@@ -420,12 +421,16 @@ Public Class dlgImportDataset
             dlgOpen.Multiselect = False
             If bFromLibrary Then
                 dlgOpen.Title = "Import from Library"
-                dlgOpen.InitialDirectory = strLibraryPath
+                ' TODO There should be a way of using Path.GetDirectoryName to avoid needing the If but couldn't get this to work
+                If RuntimeInformation.IsOSPlatform(OSPlatform.Linux) Then
+                    dlgOpen.InitialDirectory = Replace(strLibraryPath, "\", "/")
+                Else
+                    dlgOpen.InitialDirectory = strLibraryPath
+                End If
             Else
                 dlgOpen.Title = "Open Data from file"
                 dlgOpen.InitialDirectory = If(String.IsNullOrEmpty(strCurrentDirectory), frmMain.clsInstatOptions.strWorkingDirectory, strCurrentDirectory)
             End If
-
             If DialogResult.OK = dlgOpen.ShowDialog() Then
                 'always reset the multiple files checkbox
                 ucrChkMultipleFiles.Checked = False
@@ -872,7 +877,7 @@ Public Class dlgImportDataset
         'add the item to the MRU (Most Recently Used) list...
         'only add if its a file that was selected, don't add if folder path was selected
         If Not bImportFromFolder Then
-            frmMain.clsRecentItems.addToMenu(strFilePathSystem)
+            frmMain.clsRecentItems.addToMenu(Replace(strFilePathSystem, "\", "/"))
         End If
 
         'Sets the current data frame as the first new data frame
--- a/instat/dlgOpenNetCDF.vb
+++ b/instat/dlgOpenNetCDF.vb
@@ -14,6 +14,7 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports System.Runtime.InteropServices
 Imports instat.Translations
 Imports System.IO
 Imports RDotNet
@@ -34,7 +35,7 @@ Public Class dlgOpenNetCDF
     Private bCloseFile As Boolean = False
     Private strFileAssignName As String = "nc"
     Private iExpandedWidth As Integer
-    Private strLibraryPath As String = Path.Combine(frmMain.strStaticPath, "Library", "Climatic", "_Satellite/")
+    Private strLibraryPath As String = frmMain.strStaticPath & "\" & "Library" & "\" & "Climatic" & "\" & "_Satellite" & "\"
     Private bFromLibrary As Boolean = False
     Private bSubDialogOKEnabled As Boolean = True
     Private bMultiImport As Boolean = False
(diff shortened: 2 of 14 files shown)


N-thony:
@dannyparsons this works fine on my side. I have tested open from file/library still works, also from Open NetCDF and paths.

shadrackkibet (approved):

instat/dlgOpenNetCDF.vb
```diff
@@ -34,7 +35,7 @@ Public Class dlgOpenNetCDF
     Private bCloseFile As Boolean = False
     Private strFileAssignName As String = "nc"
     Private iExpandedWidth As Integer
-    Private strLibraryPath As String = Path.Combine(frmMain.strStaticPath, "Library", "Climatic", "_Satellite/")
+    Private strLibraryPath As String = frmMain.strStaticPath & "\" & "Library" & "\" & "Climatic" & "\" & "_Satellite" & "\"
```
Patowhiz, line 38: @dannyparsons please you may ignore this, it's not part of the review for this PR. I just mentioned it here for convinience.
@lloyddewit this reminds me that we may need a global data object that stores 'configured' data in memory and accessible to different classes during runtime.

We have plenty of these 'global' variables that are scattered in different classes including in `frmMain`. I usually find it architecturally incorrect.

If you agree I can create an issue on the same or we can discuss more. 

Thanks.

Patowhiz:
@dannyparsons @N-thony my save data as dialog looks very wide on my machine. @N-thony could you check this? 

save data as

N-thony:
> @dannyparsons @N-thony my save data as dialog looks very wide on my machine. @N-thony could you check this?
> 
> save data as

@Patowhiz you are right. It looks very wide on my side too.

instat/dlgSaveAs.vb
```diff
@@ -75,7 +75,7 @@ Public Class dlgSaveAs
 
     Private Sub ucrBase_ClickOk(sender As Object, e As EventArgs) Handles ucrBase.ClickOk
         frmMain.strSaveFilePath = ucrFilePath.FilePath
-        frmMain.clsRecentItems.addToMenu(Replace(ucrFilePath.FilePath, "/", "\"))
+        frmMain.clsRecentItems.addToMenu(Replace(ucrFilePath.FilePath, "\", "/"))
         frmMain.bDataSaved = True
```
Patowhiz, line 79: @dannyparsons we actually now no longer need `Replace(ucrFilePath.FilePath, "\", "/")`, `ucrFilePath` component was already designed to always have the `ucrFilePath.FilePath` with forward slashes. So we just need to have pass it as it is to the `frmMain.clsRecentItems.addToMenu()`. Have a look at my debugging image below.

filepath


In addition, you have already made changes to `frmMain.clsRecentItems.addToMenu()` that ensure all back slashes(\) are converted to forward slashes(/).

instat/dlgImportDataset.vb
```diff
@@ -420,12 +421,16 @@ Public Class dlgImportDataset
             dlgOpen.Multiselect = False
             If bFromLibrary Then
                 dlgOpen.Title = "Import from Library"
-                dlgOpen.InitialDirectory = strLibraryPath
+                ' TODO There should be a way of using Path.GetDirectoryName to avoid needing the If but couldn't get this to work
+                If RuntimeInformation.IsOSPlatform(OSPlatform.Linux) Then
+                    dlgOpen.InitialDirectory = Replace(strLibraryPath, "\", "/")
+                Else
+                    dlgOpen.InitialDirectory = strLibraryPath
+                End If
```
Patowhiz, line 429: @dannyparsons I think we don't necessary need to do the Linux check here. `.Net` recognises forward slashes in windows as well, they actually have it already built in their `.Net `platform.

If we really have to hard code, we can simply call `dlgOpen.InitialDirectory = Replace(strLibraryPath, "\", "/")` and be sure it will work on both Linux and Windows platforms.

Below are `.Net` documentation links that explain the concept much further. To save your time, just have a look at the **Remarks** section.

[DirectorySeparatorChar or AltDirectorySeparatorChar](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.directoryseparatorchar?view=net-5.0) are both valid for separating directory levels in a path string.

[Path.GetDirectoryName Method](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getdirectoryname?view=net-5.0) works with both file separators.

If I have missed something in your intentions for explicitly doing the check, please ignore this suggestion.

Patowhiz:
@dannyparsons I had an interest in this PR because it touched most components that I added and refactored.

@dannyparsons @lloyddewit 
Currently all saving of files is done by R which we pass into it file paths with forward slashes, so the saving part is already handled for us, regardless of the platform. I re-wrote most of the save dialogs to use the file path custom control that already returns file paths with the expected file separator for R.

Looking at this PR, I summarized it into 2 critical issues; 
1. We need to have R-Instat display the paths to the user correctly based on the OS platform (its technically not important but I think paths should be familiar to users).
2. R-Instat needs to use the correct file path separators regardless of what is displayed to the user when opening initial directories.
3. Currently, default hard coded file open paths are using windows default file separator e.g. the library directory path.

My suggestion is, we could still merge this if its very urgent but then refactor the code in different components to;

1. Use a custom file path control (probably extend the existing file path custom control) for opening of directories .
2. Have standard utility functions that can return the correct file paths(with the appropriate file separators) for display and opening of directories based on the OS platform.

Then use the 2 above as the standard in various components that deal with file opening in R-Instat.

dannyparsons:
@Patowhiz Thanks for the review. I was hoping that you would help make this more consistent going forward. At first I thought it was simple and that we could use `/` everywhere since Windows and Linux accepts it and R uses it. However, I think I noticed that the `OpenFileDialog`in Windows only accepts `\` whereas it needs `/` in Linux - so I got a bit confused, hence the mixed up code.

I would appreciate merging this now because of the urgency for the Linux version, but welcome other issues to address this more globally. Some global functions for dealing with and coverting file paths would be very useful for consistency.

instat/dlgImportDataset.vb
dannyparsons: It seems that the OpenFileDialog is one place that is sensitive to \` and `/`. The Windows case of this does not work in Linux, and vice versa. I think use of `Path.GetDirectoryName` could avoid the need for an If, but I started with this to ensure I had something working in both cases. Hopefully this can be improved in future.

rdstern (dismissed):
@dannyparsons I have used Import with multiple files, and checked the locations for Import netCDF and also Import from Library.  All seems ok

dannyparsons:
@Patowhiz @N-thony I've also fixed the Save As dialog size. This was because we had set `AutoSize = True` on all dialogs and user controls which we thought was needed to make the dialogs display correctly in Linux, however, this wasn't the case, but we decided not to revert that change as it didn't affect most dialogs. If you see any with this sizing issue, just change `AutoSize = False` on the dialog and possibly users controls and reposition any controls that have moved.

PR #7104, opened 2022-01-12
Added the facets angle on the inventory sub dialogue

### LINKED ISSUE #6989: Two small edits for the climatic > check data > inventory dialogue
The first is trivial, I hope.  The control, on the main dialogue, with the label Facet By: has the initial setting of Default.  When you return to the dialogue it always rests to this setting, i.e. it doesn't remember the current setting.

The second may be trickier to add?  Here is an example:
image

One solution is to allow the y-facet label to be horizontal, instead of always vertical.  This was a question [here](https://stackoverflow.com/questions/40484090/rotate-switched-facet-labels-in-ggplot2-facet-grid). 
And here is a graph in the answer:
image

So it seems it is the ( theme(strip.text.y.left = element_text(angle = 0)) .  I hope we can just say "strip.text.y" or perhaps strip.text.y.right?  (I haven't checked other options we have to know whether it is always on the right for us.

This would add Angle to the sub-dialogue.  
image

There could be another checkbox on the same line perhaps as Angle Y-Axis Label.  And perhaps the Size Y-Axis should add the word Label.  

I see these are just a subset - perhaps - of the items given on #6029 earlier.  Perhaps also check what, if anything remains from there?

N-thony:
Fixes #6989 
@rdstern here is the change that corresponds to the issue. The first point mentioned in the corresponding issue will be fixed in #7100. Please take a look of the second point which was to add the facet angle. Thanks.


rdstern:
Maybe @dannyparsons can check both the code and the results.  He wrote the original inventory dialogue!

dannyparsons (approved):
This generally looks sensible to me, if it works well in testing.

rdstern:
@N-thony these changes are very nice.  I am approving so they can be merged.
But then could you check one point on this dialogue, and also make another very minor change.
a) On the main dialogue the control on layout always returns to Default.  It doesn't remember the setting.  This was raised earlier and may be fixed automatically in the merged version.
b) A detail, but we may as well get it right.  
image

You see the defaults are different for the x-axis and the y-axis.  I wonder why, and what they should be?  Perhaps 9 and 9?  Here is the graph - it looks a bit odd with such different sizes:
image

rdstern (approved):
Comments made, but approved.  Perhaps could be merged, and then the comments could be a new issue?

PR #7109, opened 2022-01-12
Fixed the message to not appear when no data is loaded

### LINKED ISSUE #7102: Annoying message on File > Close Data Book
This message appears even if no data was loaded, and when you click on Close Data Book. See screenshot below
image

And there is a typo error in the first line of the message, it should be Are you sure you want to close `your data` rather than `you data`

N-thony:
Fixes #7102 
@dannyparsons this PR fixed the small bug related to the corresponding issue. Please take a look. Thanks.

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -1578,7 +1578,7 @@ Public Class frmMain
 
     Private Sub mnuFileCloseData_Click(sender As Object, e As EventArgs) Handles mnuFileCloseData.Click
         If Not bDataSaved Then
-            If ucrDataViewer.GetWorkSheetCount() = 0 OrElse DialogResult.Yes = MsgBox("Are you sure you want to close your data?" &
+            If ucrDataViewer.GetWorkSheetCount() = 0 OrElse DialogResult.Yes = MsgBox("Are you sure you want to close you data?" &
                                          Environment.NewLine & "Any unsaved changes will be lost.",
                                          MessageBoxButtons.YesNo, "Close Data") Then
                 clsRLink.CloseData()
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -1578,7 +1578,7 @@ Public Class frmMain
 
     Private Sub mnuFileCloseData_Click(sender As Object, e As EventArgs) Handles mnuFileCloseData.Click
         If Not bDataSaved Then
-            If DialogResult.No = MsgBox("Are you sure you want to close you data?" &
+            If ucrDataViewer.GetWorkSheetCount() = 0 OrElse DialogResult.No = MsgBox("Are you sure you want to close your data?" &
                                          Environment.NewLine & "Any unsaved changes will be lost.",
                                          MessageBoxButtons.YesNo, "Close Data") Then
                 Exit Sub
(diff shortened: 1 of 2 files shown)


instat/frmMain.vb
```diff
@@ -1578,14 +1578,13 @@ Public Class frmMain
 
     Private Sub mnuFileCloseData_Click(sender As Object, e As EventArgs) Handles mnuFileCloseData.Click
         If Not bDataSaved Then
-            If DialogResult.No = MsgBox("Are you sure you want to close you data?" &
+            If ucrDataViewer.GetWorkSheetCount() = 0 OrElse DialogResult.Yes = MsgBox("Are you sure you want to close your data?" &
                                          Environment.NewLine & "Any unsaved changes will be lost.",
                                          MessageBoxButtons.YesNo, "Close Data") Then
```
Patowhiz, line 1583: @N-thony would be ideal to pass this `"Are you sure you want to close your data?"` and `"Any unsaved changes will be lost."` through the translations function. 
You don't necessarily have to do it in this PR and I'm also not sure if we the 2 corresponding translations in the database. (@lloyddewit  could easily add the translations later and not have to change this line)

N-thony: @Patowhiz we don't have the 2 corresponding translations in the database that is the reason I made directly the change here.

dannyparsons: Good point, we should work on having all our messages going through the translation system. Can be done separately from this PR.

dannyparsons (dismissed):
Looks good to me

instat/frmMain.vb
```diff
@@ -1578,7 +1578,7 @@ Public Class frmMain
 
     Private Sub mnuFileCloseData_Click(sender As Object, e As EventArgs) Handles mnuFileCloseData.Click
         If Not bDataSaved Then
-            If ucrDataViewer.GetWorkSheetCount() = 0 OrElse DialogResult.Yes = MsgBox("Are you sure you want to close your data?" &
+            If ucrDataViewer.GetWorkSheetCount() = 0 OrElse DialogResult.Yes = MsgBox("Are you sure you want to close you data?" &
```
shadrackkibet, line 1581: Looks like you reverted the typo correction?

shadrackkibet (changes requested):

instat/frmMain.vb
N-thony: Yes, as @Patowhiz, I will fix that in another PR. But first I will check with @lloyddewit for the possibility.

shadrackkibet: Okay

shadrackkibet (dismissed):

instat/frmMain.vb
lloyddewit: `MsgBox` text is not currently translated. Some `MsgBox` text is multiline and/or contains variable values. This text will be more difficult to translate.
I suggest that we implement the `MsgBox` translation mechanism, across all of R-Instat, in a separate PR.
In summary, I think @N-thony was correct to fix this typo here.

instat/frmMain.vb
lloyddewit: @N-thony your original typo fix was correct (see my comment above)
```suggestion
            If ucrDataViewer.GetWorkSheetCount() = 0 OrElse DialogResult.Yes = MsgBox("Are you sure you want to close your data?" &
```

lloyddewit (approved):

lloyddewit:
@shadrackkibet @rdstern Please could one of you test/merge? thanks

shadrackkibet (approved):

PR #7110, opened 2022-01-12
Fixed export dialog sizes

dannyparsons:
@africanmathsinitiative/developers Please check the Export dialogs are now the correct size. In the current version they are too large.


rdstern (approved):
Yes - seems to work

PR #7111, opened 2022-01-12
Fixed hiding of sheets when data frame is loaded to the data viewer

### LINKED ISSUE #7006: The sheets tab is hidden by the scroll bar when a dataset is loaded

Patowhiz:
Fixes #7006.


rdstern (dismissed):
@dannyparsons seems to work?  I noticed, when opening 6 sheets together, that the scroll bar seemed to click multiple times.  But, in the end it was fine.

dannyparsons (commented):
This is great, well done for finding the cause. It works well. But is there a way that this could be set just once per session? Because now if I have 6 sheets and make the tab wider, then when it refreshes again it makes it smaller than I had it.

A quick solution would be to keep it here but have a boolean to check that it's only done once. That would be fine for me.

Patowhiz:
@rdstern thanks for testing.
@dannyparsons yes, I didn't think the flickering will be noticeable to the human eye.

I agree with the suggestion, let me look into other possible solutions. If they take me more than 3 hours I'll just do a quick fix as suggested.

Patowhiz:
@rdstern the scrollbar flickering happens anytime you load a new sheet to the grid. 
This is caused by the `reogrid` adjusting its scrollable bar width because of the difference in column sizes of the sheets.
You will also notice this even as you change from one sheet to another. 
We could fix this but I think its beyond the scope of this PR and also it seems to be a better design for scrolling.
So I can confirm that the flickering is not cause by any change made by this PR.
Thanks.

Patowhiz:
@dannyparsons I've just added the `boolean` check with a comment that other possible solutions could be viable.

@dannyparsons @ChrisMarsh82 the other possible solution that I found ideal was to use the `TableLayoutPanel.RowStyles` property. 
So we arrange the controls of the data view as rows inside 1 column of the `TableLayoutPane`l, then we adjust the height of the row during runtime. 
This means only the height of the controls will be affected but the widths will always be intact (meaning even the internal widths of the child controls will be preserved during runtime).

This will make changes to the designer and only to the `SetGridVisibility()` subroutine. 
However, @dannyparsons I have not made that change because I'm still not sure if there was any other reason as to why @ChrisMarsh82 used `tlpTableContainer.ColumnStyles` instead of `TableLayoutPanel.RowStyles` other than just for visibility functionality.

Patowhiz:
@rdstern @dannyparsons this now ready for testing

dannyparsons (approved):
Thanks this works well. I would have slightly preferred the size to only be set once as a default, and then allow users to have it smaller as well as larger if they wish, but I can't really see why that's particularly needed so this is a good solution now.

shadrackkibet (approved):

PR #7107, opened 2022-01-12
Adding sub dialogue for formatting of summary tables

Ivanluv:
@africanmathsinitiative/developers this fixes  #6770 and replaces PR #6977

PR #7123, opened 2022-01-13
Improving the Describe > One Variable > Summarise dialogue

### LINKED ISSUE #7081: Improve the Describe > One Variable > Summarise dialogue and possibly 2-variable and 3-variable?
I use the data from the R-Instat review as an example:

I use the one variable summarise and wanted to get the sum (as in the review) and also the number of occasions with more than zero.
Here is the output:
```

# Code generated by the dialog, One Variable Summarise
data_book$summary(data_name="Scores", na.rm=TRUE, count_test="'>'", count_value=0, j=1, return_output=TRUE, summaries=c("count_calc", "summary_count_non_missing", "summary_count", "summary_sum"), columns_to_summarise=c("BlueSky","Deducer","JASP","jamovi","RAF","Rattle","Rcmdr","R_Instat","RKWard"))

                  BlueSky Deducer JASP jamovi RAF Rattle Rcmdr R_Instat RKWard
count                  43      43   43     43  43     43    43       43     43
count-non-missing      43      37   39     40  37     38    37       43     35
sum                   359     102  224    235  59    110   686      290    100
count-calc             NA      NA   NA     NA  NA     NA    NA       NA     NA

```
The main changes I would like are to the layout.  But here there is an oddity in the calculation.  What am I doing wrong?  The omit missing values option is ticked?  The count-calc is count values > 0, i.e. the number of values of 1 or more.  
This points to the urgency of doing something here.  But now I remember that we had the same issue with our Describe > Specific > Summary Tables dialogue.  And we solved the problem there!

Aha I have the solution in general.  One thing I would like is to be able to control the orientation, i.e. choose whether it is the variables or the summaries that go down.  I am sure there will be more.  But this is (almost) a special case of the Summary tables dialogue.  I say almost, because that dialogue is limited to numeric variables to summarise, while here we can summarise any type - though dates become numeric!  Anyway we can summarise factors and ordered factors include max and min.  (Do we want an option to do ordinary summarise on numeric levels of a factor - I guess not???)
We might give the results separately for the factors and the variates?

And I suggest these special cases will also be ok for the 2-variable and 3-variable summarise as well!  We will have a limited set of options - that's to keep it simple.  If you want full generality then you go to the main dialogues under the Specific menu.

I assume this is mainly work for @Ivanluv, with possible guidance from @lilyclements.  He will, I hope, be able to sub-contract some of the work to one of the interns?

Ivanluv:
fixes #7081 
@rdstern In this PR I have added the functionality to control the orientation, i.e. choosing whether the variables or the summaries go down in the summary table


rdstern:
@Ivanluv well done.  That looks great for a start.
a) Omit the Treat Columns as Factor checkbox, but always do that, so the 2 other controls are always visible.
b) Change the Label for the first from Summary to Summaries.  Default is for it to be checked.
c) Now a puzzle that might be more difficult to resolve, namely the receiver can take any type of variable.  So could be factor, or ordered factor or date, etc.  Currently it gives an error for dates or factors or character variables.  Daters should be ok (in the long run) to solve, and also need to be possible in the general tabulation.  I am happy for most summaries to be missing when they don't apply.  But I would like to be able to cope with those that do.  I hope you may have ideas on what can be done in mmtable2, e.g. dates, and what might be impossible.   
d) Could you also check on different types of output - this applies to the general tabulation as well.  Currently it is html.  @Patowhiz is looking into whether that may be possible in the output window.  I wonder whether there might be other options, particularly simple ones that could go into the current output window.

rdstern:
@Ivanluv it now seems to get the correct summaries, but in the wrong order.
image
This was the dialogue:
image

a) You can see that Blue Sky remains at the top in the results, but its summary is at the bottom!
b) The summary for R-Instat is in second place (towards the top - but it should be second from the bottom.
c) The labelling has become alphabetical.

And, may need @lilyclements help - it doesn't work for factor variables. etc.  I would like it to do something for variables of any type.  Not sure what to do here?

rdstern (changes requested):
Comments given for changes

Ivanluv:
@lilyclements  any suggestions here

lilyclements:
@Ivanluv have you tried updating your branch? I am not getting this error if I try it on the version in #7107 - which I assume is more updated.

Ivanluv:
> And, may need @lilyclements help - it doesn't work for factor variables. etc. I would like it to do something for variables of any type. Not sure what to do here?

@lilyclements  I have updated this branch and its now working okay

Ivanluv:
> And, may need @lilyclements help - it doesn't work for factor variables. etc. I would like it to do something for variables of any type. Not sure what to do here?

@lilyclements d you have any ideas how we could fix this?

rdstern (changes requested):
@Ivanluv this is now working fine.  Now I tried with other types of data (not-numeric) and it doesn't like that.
Now I thought you might need help from @lilyclements for this.  But then I tried with our summary system and that handles all types of data perfectly - even character variables.  And dates come out fine as dates in the summary.  So I wonder if that can provide the code to be used before going into mmtable?  
I treid with dodoma data, so there is a date variable  and also month as a factor.  Then I made month into character.  It just gives NA when it can't work out the summary!

On the actual dialogue could you have the Variables as rows checked by default.

Ivanluv:
@rdstern by the "summary system" do you mean summary tables?

rdstern:
No, I meant Prepare > Data: Reshape > Column Summaries.  The first receiver is Variable(s) to Summarise and it allows all types of variable - as does your Describe > One Variable > Summarise.  And it also has the same sub-dialogue as yours, and copes with any summary for any type!  Just what we need in your One Variable dialogue!

rdstern:
@Ivanluv you ask - " should we have the results stored as it is being done in the Describe > One Variable > Summarise?"  I assume you mean as is done in the Prepare > One Variable > Column Summaries"

I would not, at least for now.  Once we are able to do sensible summaries for variables of all types, then I would like to merge and then reflect and consider.  

What is done here is a great update, and very useful.  I am more concerned here with the tables of results.  And they are able to use any of the summaries we have in our system.  There will be more to do, but that can come later.  We need it to stay simple, but could usefully add the options for the missing values - now that we have them in a package.
Then I am waiting for Patrick to sort out html for the output window, which will be great.

Then (as you say) there will be 2 further directions:  
a) More flexibility on the output.  We could add the presentation options from the Summary tables work.  But let's finish that on the summary tables dialogue first.  Then this dialogue could allow users to practice with very simple structures.
b) We could permit the save - as you say.

We also apply these ideas to the 2-way and 3-way tables - if we have the 3-way - I think so!  

Then in simple graphs and tables (which IS descriptive statistics after all) we have a package to be envied!
This will be good for both teaching and also doing!

rdstern (commented):
@Ivanluv not sure what has happened here?  This now seems like the old dialogue again?

Ivanluv:
@rdstern  Am still making the changes  for the dialogue.

rdstern (changes requested):
Hi @Ivanluv I assume these changes are "work in progress?", but I had a quick look.
a) The options are now visible all the time.  Should not be, for the default where the summary is used.
b) The default of ticking the Variables as rows is still to be done.
c) The summary gives an error now.

Ivanluv:
@rdstern could you have a look

rdstern:
@Ivanluv I think you may be a few tweaks away from what's needed.
I used the review data as before.
a) It now does not give an error when you add the category variables.  Hooray!
b) The answers seem to be correct.
c) But the variables are in alphabetical order - not the order in the dialogue.  (We used to have that problem with the tables.
d) And the proportions don't work.  They give missing values.  (I had the count > 0 as the statistic)
e) And the Omit Missing Values checkbox doesn't seem to work.  - the summary sums are still given as missing.

Here is the table I got:
image

And here are the commands you generated:
image

Ivanluv:
> c) But the variables are in alphabetical order - not the order in the dialogue. (We used to have that problem with the tables.

@lilyclements could you look at the RCode and see what I might have missed in the code for this error to occur. Could you also help me find out why the proportions are not working (part (d) of Rogers review above)

rdstern:
@Ivanluv @lilyclements will soon be looking at the R code.  In the meantime could you make a cosmetic change to the dialogue as follows:
Change the radio buttons Default and Customised to our more usual ones (when they change other controls, i.e. they become 2 option buttons at the top.
With the customised we should also check - if it is easy - adding the Missing Value option button (sub-dialogue) from Summary - mI assume that will be easy?

lilyclements:
@Ivanluv if the user selects to modify/choose the summary statistics to display, you are currently using `$calculate_summary` not `$summary_table`. 

`$summary_table` is a wrapper for `$calculate_summary` therefore you do not need to use `$pivot_longer` with `$summary_table` as that is all incorporated into the code.
Additionally, the alphabetical order would be sorted with `$summary_table` since this is already written in.
Is there a reason to not use `summary_table`?

In PR #7230, the code is fixed to use date variables in the summary functions. Once this is merged in, then update your master and hopefully that will work! (Or you can merge it into this PR if you think that's more appropriate).

lilyclements:
@Ivanluv I'm not sure if this will fix the proportions problem. If it is still a problem after changing the function then I can look into it then

lilyclements:
@Ivanluv how is this work going?

Ivanluv:
@lilyclements  am waiting for PR #7230 to be merged

Ivanluv:
@rdstern @lilyclements  I have merged PR #7230 into this and made the changes as suggested by @lilyclements i.e using the `summary_table `function instead of the `$calculate_summary` function. There is however a challenge when trying to find summaries not supported by certain data types eg getting the sum of factor columns as below.

image


image

image

After the error the table won't be produced even if one of the selected columns supports the summary

rdstern:
It is looking nice
image


Note sensible summaries for date variables.  And I made month into an ordered factor, so it can get the min and max!  That's nice to explain to people!


But we do need to solve the problem that @Ivanluv mentioned, namely that it can't cope with some summary statistics - like sum, or mean, for a factor.  When values are missing, and that option isn't given, then that cell is blank.  We need the same for those statistics for a factor variable, etc.

One small change if @Ivanluv can't solve the current error, is to move the Missing option to below the other 2 checkboxes.  Then it will be in the same relative position as in the summary dialogue.  
But we can't merge until we can present tables of results for datasets (like the simple survey) that mix factors and variates.

lilyclements:
So the sum/mean for factor is a problem in `calculate_summary` function. The same error reported by @Ivanluv can occur elsewhere in R-Instat. For example, in the "Column Summaries" dialog - 

image

I'll work on getting this into the `calculate_summary` function.

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -1418,28 +1423,29 @@ DataBook$set("public", "summary_table", function(data_name, columns_to_summarise
       for (facts in power_sets_summary) {
         if (length(facts) == 0) facts <- c()
         summary_margins_df <- data_book$get_data_frame(data_name = data_name) %>%
-          dplyr::select(c(factors, columns_to_summarise)) %>%
-          tidyr::pivot_longer(cols = columns_to_summarise)
+          dplyr::select(c(tidyselect::all_of(factors), tidyselect::all_of(columns_to_summarise))) %>%
+          tidyr::pivot_longer(cols = columns_to_summarise, values_transform = list(value = as.character))
         data_book$import_data(data_tables = list(summary_margins_df = summary_margins_df))
         summary_margins[[length(summary_margins) + 1]] <- data_book$calculate_summary(data_name = "summary_margins_df", columns_to_summarise = "value", summaries = summaries, factors = facts, store_results = FALSE, drop = drop, na.rm = na.rm, return_output = TRUE, weights = weights, result_names = result_names, percentage_type = percentage_type, perc_total_columns = perc_total_columns, perc_total_factors = perc_total_factors, perc_total_filter = perc_total_filter, perc_decimal = perc_decimal, margin_name = margin_name, additional_filter = additional_filter, perc_return_all = FALSE, ...)
```
lilyclements, line 1429: here

instat/static/InstatObject/R/Backend_Components/summary_functions.R
```diff
@@ -1418,28 +1423,29 @@ DataBook$set("public", "summary_table", function(data_name, columns_to_summarise
...
         data_book$delete_dataframes(data_names = "summary_margins_df")
```
lilyclements, line 1430: finally here. Thanks!

lilyclements:
> @rdstern @lilyclements I have merged PR https://github.com/africanmathsinitiative/R-Instat/pull/7230 into this and made the changes as suggested by @lilyclements i.e using the summary_table function instead of the $calculate_summary function. There is however a challenge when trying to find summaries not supported by certain data types eg getting the sum of factor columns as below.

@Ivanluv @rdstern I have now fixed this bug in PR #7338

Ivanluv:
@lilyclements  I have pulled you changes from  PR #7338 and am getting the following error
image

image
image

instat/static/InstatObject/R/Backend_Components/calculations.R
```diff
@@ -471,7 +465,25 @@ DataBook$set("public", "apply_instat_calculation", function(calc, curr_data_list
...
+    calc_from_data_name <- curr_data_list[[c_link_label]][["from_data_frame"]]
+    formula_fn_exp <- as.formula(paste0("~", calc$function_exp))
+    # note: important that there is *no* space between | for grepl function
+    if (any(grepl("summary_sum|summary_mean", formula_fn_exp))){
+      col_data_type <- self$get_variables_metadata(data_name = calc_from_data_name,
+                                                   column = col_name,
+                                                   property = "class")
+      # note: important that there *is* a space between | for str_detect function
+      if (any(stringr::str_detect("factor | Date", col_data_type))){
+        curr_data_list[[c_data_label]] <- curr_data_list[[c_data_label]] %>%
```
lilyclements, line 477: @Ivanluv can you change “factor | Date” to
“factor | character | Date”

it is important to have a space between the | too

lilyclements (commented):
@Ivanluv is village or size a character variable? If so then the following should fix it. If not, then the following will fix another potential bug so can you implement the following anyway:

instat/static/InstatObject/R/Backend_Components/calculations.R
Ivanluv: On it

Ivanluv: @lilyclements  the bug is now fixed

rdstern:
@Ivanluv was this supposed to work now?  I tried with a minimum for a factor and it said that isn't allowed.

rdstern:
@Ivanluv and @lilyclements I am hoping this will be included in the Version 0.7.5 update?  That's in 2 weeks time.  OK?

lilyclements:
> @Ivanluv was this supposed to work now? I tried with a minimum for a factor and it said that isn't allowed.

@rdstern which combinations do we not want to work?

**Factors and characters:**
minimum/maximum
mean
median
sum
range
SD
variance

**Ordered factors:**
mean
median
sum
range
SD
variance

**Dates:**
sum?

I have not yet put in that min/max (amongst others in the list above) shouldn't work for factors or characters, but that is very quick to add. I'll add any other combinations when you've responded to this so I know if there's any I'm missing here. Thanks!

rdstern:
@lilyclements and @Ivanluv I'd prefer to look at this the other way round, namely what is sensible for different types.  I am just keen that we have an option to mix the data types in this dialogue, rather than having to think carefully what to put in the the receiver.
So numeric allows everything.  Logical and date are special cases of numeric
a) Ordinary factors (or character) only allow number of values, including number of unique values, number of missing, etc.
b) Ordered factors allow max and min (and they should therefore allow range - perhaps we don't yet use the range function?
c) Ordered factors currently don't allow anything else.  I actually don't wee why they can't allow median and other percentiles, because they are ordered.  But let's stick with what R does for now.  
d) So only the numeric types allow means, sd, skewness and everything else.  And for now percentiles, and proportions, etc.

Ivanluv:
@lilyclements what is you suggestion on how to proceed ?

PR #7131, opened 2022-01-17
Implemented the new change on Prepare > Column: Text > Find/Replace dialogue

### LINKED ISSUE #7120: Change Prepare > Column: Text > Find/Replace dialogue again!
Apologies @N-thony but, just after you agreed to make a new issue of a minor change in this dialogue, I am requesting larger changes.  Please bear with me, because I think this is a clear improvement.  I hope it can be quick, as it is almost all cosmetic!  These have become obvious as I am now trying to write the help.  I am getting to something simpler.

a) The menu says Find/Replace, so the heading in the dialogue, which is currently String Handling could also change to become Find/Replace.
b) Then the idea of the top radio buttons is that they correspond to different layouts in the dialogue.  This isn't the case here.  So please change them to just 2 (familiar) radio buttons, namely just `Find` and `Replace`.
c) So, the dialogue will change in shape a bit.  It can be a little less wide and a little deeper.  The 2 (or 3) existing controls could perhaps be raised a little.
d) Then Find will have a group box under the 2 existing controls.  It might have the label Options.  I would like a better word.  Inside are (at least) Count, Detect, Locate, Extract.  
e) The second has just Replace and Replace All.
f) The names for the resulting variables can remain as they are.

N-thony:
Fixes (partially) #7120 
@rdstern I have made the change requested in the corresponding issue. I left the Locate option under Find option. Please take a look of the change so far. Thanks.

### DIFF
--- a/instat/dlgStringHandling.Designer.vb
+++ b/instat/dlgStringHandling.Designer.vb
@@ -47,18 +47,28 @@ Partial Class dlgStringHandling
         Me.rdoReplace = New System.Windows.Forms.RadioButton()
         Me.rdoFind = New System.Windows.Forms.RadioButton()
         Me.grpModifiers = New System.Windows.Forms.GroupBox()
-        Me.ucrChkComments = New instat.ucrCheck()
-        Me.ucrChkMultiline = New instat.ucrCheck()
         Me.cmdAddkeyboard = New System.Windows.Forms.Button()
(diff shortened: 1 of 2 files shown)


lloyddewit:
@HawardKetoyoMsatsi Please could you peer review?
Thanks

rdstern (changes requested):
@N-thony it is looking good.  I suggest getting rid of the Match option.  Then I'll approve.

PR #6857, opened 2021-10-15
Added more recoding options into recode numeric dialog

mmulahvictor:
Fixes #6851
@africanmathsinitiative/developers this is ready for review
@rdstern @shadrackkibet have a look at it.

PR #7136, opened 2022-01-18
Added recode options into recode numeric dialog

### LINKED ISSUE #6851: Improve the Prepare > Factor > Recode Numeric dialogue
This is a relatively simple task.  Could @shadrackkibet suggest who could have the time to do this.

This dialogue currently uses the default cut function.  I propose instead we use the `Hmisc::cut2` function instead.
Here is the current dialogue:
image

1) We no longer need the `Closed on` group box, 
2) There are 3 radio buttons under the single receiver.  Possibly in a group box.  The 3rd is the Existing Break Points as it is now
3) The first is labelled Quantiles and is followed, if chosen, by an up-down with minimum 1 and no maximum, and default say 4.
4) the middle one is labelled Minimum and again has an up-down with miit includes nimum 1 and no maximum.
5) Change the Add Labels checkbox to `Label Groups with Means`.  Get rid of the field to type in.  If chosen the command adds levels.mean = TRUE and also digits=3

mmulahvictor:
Fixes #6851
This PR replaces #6857.
@africanmathsinitiative/developers @shadrackkibet 
This is ready for review


rdstern (changes requested):
@mmulahvictor sorry to be pedantic.  The default is now, correctly 4.
But if you use the third option forst, and then come back to the dialogue, and then move to the first option - it is set to 1 again as the default.

rdstern (approved):
Looks good now and an excellent improvement on the previous dialogue

shadrackkibet (approved):
Well done @mmulahvictor.

PR #7071, opened 2021-12-16
Correct Regular Sequence Dialogue

derekagorhom:
Fixes issue #6672
Moved the preview up in the dialogue to allow for room for the Position button to go after the main control
@maxwellfundi can you check if everything is in order

PR #7139, opened 2022-01-19
Resolved issues with layout of the Regular Sequence Dialog

### LINKED ISSUE #6672: Correct the Regular Sequence dialogue
@maxwellfundi if you are introducing the new interns then here is a simple bug to correct in the Prepare > Calculate > Regular Sequence dialogue.  Here is a screen shot:

image

It looks very untidy now and the Position control is obviously in the wrong place.
a) I suggest the preview be moved up in the dialogue, so there is room for the Position button to go after the main control - which it is, in other dialogues.
b) While there it seems a shame not to line up the 4 main controls for From, To, etc.  When doing that lining, make sure the spacing works for the dates button as well as for numeric columns.  And that the spacing is ok if the language changes.

derekagorhom:
fixes #6672 from the previous PR #7071 
moved the preview up in the dialogue to allow room for the Position button and also fixes issue with some controls
resetting back to the original default

### DIFF
--- a/instat/dlgRegularSequence.designer.vb
+++ b/instat/dlgRegularSequence.designer.vb
@@ -264,7 +264,7 @@ Partial Class dlgRegularSequence
         '
         Me.lblPreview.AutoSize = True
         Me.lblPreview.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.lblPreview.Location = New System.Drawing.Point(258, 80)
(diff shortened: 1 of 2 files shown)


derekagorhom:
@rdstern can you take a look at this to see if it is ok

lloyddewit (approved):

rdstern (approved):
@derekagorhom looks good to me now.

PR #7140, opened 2022-01-19
Small corrections to Prepare> Dataframe> Insert columns/rows dialog

### LINKED ISSUE #7075: Small corrections to Prepare > Data Frame > Insert Columns/Rows
a) The dialogue need not be so wide.
b) The pull-down currently gives alternative missing values only - and they seem to give an error.  Instead, keep NA as the default and have 0, "Kisumu", "New Zealand", TRUE, FALSE as other possibilities.

@shadrackkibet could you suggest someone?

mmulahvictor:
This PR fixes #7075

- I have slightly reduced the size of the dialog
- I have added new values in the pull-down and mantained NA as the default
- I have also enabled quotes to the some of the above value to be read as strings. This has removed the error when running the dialog.
@shadrackkibet can you look at this changes.


lloyddewit (approved):

rdstern (changes requested):
Sorry @mmulahvictor I think there is a problem that may have been there before?
When you give NA your command puts quotes round it.  So, what is delivered is a string containing the characters NA.  That's not the same as NA without the quotes, which would give missing.
While there, could you look at what give for the 0 options.  I wonder if you give "0"?  Certainly the result is a characet variable.  Would it be possible to give just 0 and then check it is a numeric variable?

shadrackkibet:
@mmulahvictor what is the state of this PR?

mmulahvictor:
> @mmulahvictor what is the state of this PR?

@shadrackkibet it seems to be working ok.

lloyddewit (approved):

lloyddewit:
@rdstern Please could you test again? thanks

rdstern (approved):
Great

PR #7145, opened 2022-01-21
Linking Direct to the sub-Help

Wycklife:
Fixing Parts of  #6967

@rdstern This is another set of sub-dialog that links direct to the sub-help. This was suggested by @rachelkg and below are the sub-dialogs;

1. Describe>Two Variables>Frequencies
2. Describe>Three Variables Frequencies
3. Climatic>Mapping>Map
4. Procurement>Mapping>Map Country Values

It's ready and can be reviewed.


rdstern (changes requested):
@Wycklife great if that is working.  But I can't tell for those dialogues, because they don't have much help.  So I can't see if it is going to the top or the anchor point.  Could you try for (for example) describe > one variable > graph, which is more complete.

Wycklife:
@rdstern I was given two list of sub-dialogs by @rachelkg. One of the list contains sub-dialogs that should link direct to the sub-help(This pull request), the second list contains sub-dialogs which links on to the anchor point(Still in progress).

rdstern:
Let me give my understanding of what we are aiming for.  Then perhaps @rachelkg can also comment, or maybe this is enough for @Wycklife to progress.
a) We now have a place in the help for those sub-dialogues that are repeated in multiple dialogues.  One example - used above - is the plot options, which is one of the sub-dialogues from Climatic > Mapping > Map.
In the pull request the help from this sub-dialogue correctly goes there.
b) When the sub-dialogue is used in only one (or perhaps also now two) dialogues, then the sub-dialogue help is together with the main dialogue help, but at an anchor point.  Then the help should jump straight to that anchor point.

@Wycklife tells me the lin king to the anchor points is still work in progress.  

I am not sure about the other 3 examples given by @Wycklife in this pull request.  They seem to go to the main place for the main-dialogue help.  That is except for the 3-variable which goes to the 2 variable position.

So, wycklife could you either explain what each example is supposed to be doing here.  Then perhaps also wait until you have finished on the anchor points.  Why is it taking so long?  If it isn't possible then please say so.

Also perhaps @rachel and I could review the list that she sent.  I am a bit confused about one or two of the titles.

rdstern:
@Wycklife I think we have agreed not to link to anchor points.  So I presume this can be closed?  There will soon be work on linking the sub-dialogues to new entries in the help file, but I assume that will be a different pull request.

Wycklife:
@rdstern These are some of the sub-dialogs which goes to the sub-help. May be you can review and see whether they are linking well to the sub-help not anchor points.

Wycklife:
This PR is being fixed on a new branch since this branch has got conlicts.

PR #7155, opened 2022-01-24
Small change to the facet label size on the Inventory subdialog

### LINKED ISSUE #7150: Improving Inventory plot dialog
@N-thony these changes are very nice.  I am approving so they can be merged.
But then could you check one point on this dialogue, and also make another very minor change.
a) On the main dialogue the control on layout always returns to Default.  It doesn't remember the setting.  This was raised earlier and may be fixed automatically in the merged version.
b) A detail, but we may as well get it right.  
image

You see the defaults are different for the x-axis and the y-axis.  I wonder why, and what they should be?  Perhaps 9 and 9?  Here is the graph - it looks a bit odd with such different sizes:
image

_Originally posted by @rdstern in https://github.com/africanmathsinitiative/R-Instat/issues/7104#issuecomment-1019073121_

N-thony:
Fixes #7150 
@rdstern I made the small on the size of the facet label for x-axis and y-axis to 9 as suggested. Please take look. Perhaps you could check how it looks when `x-axis angle = 0` and `y-axis angle = 90`. Thanks.


rdstern:
image

I have been using the Ghana tidy data from the library and the last option, namely Elements/Stations

I changed the orientation of x and y both to 0 and did nothing about the size.

This is where I would like the x labels to be a bit bigger.  And perhaps the same size as the y variables.
The default for the y is vertical and that seems fine.  I wonder (at least with this option) why the default for the x labels is vertical?

rdstern (approved):
@N-thony thanks

lloyddewit (approved):

PR #7167, opened 2022-01-26
Made Design Changes for Compare Columns Dialog

### LINKED ISSUE #7166: Prepare > Check Data > Compare Columns
Here is the dialogue when you first enter it.
image

This is not as odd as first appears.  The blank space at the bottom is because it is needed by the other radio button.  (By Value).  
If you go to that option, and then return to By Row, then the extra space isn't there.  

So it is just that on first entry it is using the size of the By Value, rather  than the By Row.
So it should be very easy to fix.  Perhaps @shadrackkibet can decide who can do this

Gioche6:
Fixes #7166

@africanmathsinitiative/developers , This is ready for review.
@shadrackkibet , @rdstern . This is ready for your review and testing.

- This issue was fixed by simply setting `AutoSize` to ` False` on the dilaog design properties


rdstern (approved):
This has been a small change, just in the size of the dialogue.  Hope @shadrackkibet or @lloyddewit could check quickly and merge

lloyddewit (approved):

PR #7168, opened 2022-01-26
In Script dialog, correctly execute command copied to Output window

### LINKED ISSUE #7151: Small fix needed in Edit > Script dialogue
Open a data frame.  Anyone will do.
Then I would like to know that size of the dataframe object.
That's simply the command `object.size(ghana) `- if ghana is the data frame name.
If you choose the data frame option and Add, then it adds the first line in the code shown below.
Then I just typed the command.
When I press Script it copies the code into the script window and it works perfectly.  
But I thought I could (more easily) just press Ok.  Then it correctly puts the commands in the output window, but doesn't seem to execute them.
That's a shame.  Is there more it could/should do?
 
image

HawardKetoyoMsatsi:
Fixes #7151
@rdstern, @africanmathsinitiative/developers this is ready for review.


rdstern:
@HawardKetoyoMsatsi it still seems the same as before:
image


When I press Ok it puts the commands in the output window, but not the result.
When I press To Script it puts the code into the script window.  Running from there gives the result.

HawardKetoyoMsatsi:
@rdstern I suppose you may not have pulled my changes correctly. 
The code produces correct results from my end.
Below is the output I get.

image

rdstern (approved):
Great - yes it does work!

instat/dlgScript.vb
```diff
@@ -378,5 +380,7 @@ Public Class dlgScript
         End If
     End Sub
 
-
+    Private Function RunText(strText As String) As String
+        Return If(strText <> "", frmMain.clsRLink.RunScriptFromWindow(strNewScript:=strText, strNewComment:=strComment), "")
```
lloyddewit, line 384: safer to use IsNullOrEmpty?
```suggestion
        Return If(String.IsNullOrEmpty(strText), "", frmMain.clsRLink.RunScriptFromWindow(strNewScript:=strText, strNewComment:=strComment))
```

instat/dlgScript.vb
```diff
@@ -191,6 +191,10 @@ Public Class dlgScript
         SetRCodeForControls(True)
     End Sub
 
+    Private Sub ucrBase_ClickOk(sender As Object, e As EventArgs) Handles ucrBase.ClickOk
+        RunText(txtScript.Text)
```
lloyddewit, line 195: What is the purpose of using a private function? could we delete the function and copy the code here?
We ignore the returned value, is that correct?

lloyddewit (changes requested):

HawardKetoyoMsatsi:
@lloyddewit  I have made the requested changes. Please take a look.

lloyddewit (approved):

instat/dlgScript.vb
lloyddewit: @HawardKetoyoMsatsi `clsRLink.RunScriptFromWindow()' returns any part of the script that was not executed. It should be rare that part of the script is not executed and I'm not sure it's worth showing a warning dialog for this. I saw that you decided to ignore this returned value. I think that's reasonable.

lloyddewit:
@rdstern Since your last approval, the code was simplified. The functionality should (in theory!) be the same. Please could you test/approve again?
Should we merge this PR for 0.7.4 or should we wait?

rdstern (approved):
Still seems fine

lloyddewit:
@rdstern This is a small, low risk change. So I went ahead and merged.

PR #7171, opened 2022-01-26
Improved Import from Library for Instat files

### LINKED ISSUE #7164: Improve Import from Library for Instat files
Please check this with Version 7.3.1.  It isn't so easy from the new master version.
Import from Library works pretty well for the R packages.
For the Instat files it is ok the first time.
But pressing the last dialogue is then not ideal.  This single operation has generated 2 entries in the recent files.  The most recent is the Import File dialogue.  But it has forgotten the path, so is of no use.  
image

So I always go back to square 1!

N-thony:
Fixes #7164 
@rdstern sorry for the delay. I have made a small change that fixes the problem raised in the corresponding issue. Please could you test it? Thanks. Then I will have a look at loading the multiple files with a given extension within subfolders.

### DIFF
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -1030,7 +1030,6 @@ Public Class dlgImportDataset
                 strFileExtension = ""
                 strFilePathSystem = ""
                 strFilePathR = ""
-                ucrInputFilePath.SetName("")
             Else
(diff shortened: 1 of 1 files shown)


rdstern (approved):
Great.  That is an improvement!

shadrackkibet (approved):

Issue #1504, opened 2016-06-24
Checking if receiver is having something every time the dialog is reopening 


Lunalo:
@africanmathsinitiative/instat 

I realized that a check for receiver stills holds something every time the dialog is being reopened is missing.

I hope we will look into this maybe as from next week.

Thanks.

dannyparsons:
If a receiver contains a column that no longer exists in the data, it should be cleared when the dialog is reopened.

maxwellfundi:
@danny this is resolved. it needs closing

maxwellfundi:
No, actually it is not resolved...

dannyparsons:
We also need to check the type of the column if the receiver is restricted. The column may exist but type is no longer valid for the receiver.
Some thought needed to resolve this.

dannyparsons:
This can be done with a method run in R by the receiver in the method reading the R code of the receiver.

AlexSananka:
What is the state of this?was it finally resolved?

dannyparsons:
We now have a partial fix - the single receiver clears on reopen if the column no longer exists.

Still to do:
- Checks columns exists on reopening in multiple receiver
- Check the column satisfies metadata conditions for single & multiple receivers on reopen
- Implement checks when receiver is for other types - not only columns

Ogik99:
What is the status of this issue?
@africanmathsinitiative/r-instat

dannyparsons:
There is still work to do on this for the remaining issues which will require some more discussion first.

Patowhiz:
@dannyparsons though not included in our recent milestone we could discuss more on this if it's included in the next milestone

dannyparsons:
I think what we really want to solve this and not slow down dialogs by having a lot of interaction with R is to store column metadata in the front end and use this as a central checking point. 
That requires quite a bit of restructuring of core code which is probably why its not been addressed sooner.

N-thony:
> 

@dannyparsons we can think about implementing this now. I will look into it.

lilyclements:
I beileve this is now complete

lilyclements:
Reopening after seeing the message

"I think what we really want to solve this and not slow down dialogs by having a lot of interaction with R is to store column metadata in the front end and use this as a central checking point.
That requires quite a bit of restructuring of core code which is probably why its not been addressed sooner."

PR #7172, opened 2022-01-26
Fixing a bug with the margins in the summary table function 

lilyclements:
Fixes issue found in #7107 

@rdstern found an issue with the margins in the `summary_calculations` function when the summary itself contained more than one `_` (for example, `summary_count_non_missing`, but not `summary_sum`).

> And then it gets more "interesting" when I add the margins! This might be a @lilyclements question? image

This problem was occurring because we have the columns saved as `summary-type _ variable-name` (e.g. `sum_field`, `count_non_missing_field`). This PR fixes the issue found.

@rdstern @africanmathsinitiative/developers this is ready to review

### DIFF
--- a/instat/static/InstatObject/R/Backend_Components/summary_functions.R
+++ b/instat/static/InstatObject/R/Backend_Components/summary_functions.R
@@ -95,7 +95,7 @@ DataBook$set("public", "append_summaries_to_data_object", function(out, data_nam
 } 
 )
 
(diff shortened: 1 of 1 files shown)


rdstern:
@lilyclements excellent - that works!  
Of course I managed to find something that doesn't, but it is much less important.  Hopefully easier to fix?  Otherwise I'll Approve this.  If so, then should Ivan incorporate it in his pull request, or could both be merged?

The variable is given in the comparison table as `R_Instat`.  In attempting to see what happened when it was different, I renamed it as ` R.Instat`.
That gave me the following error.

image

lilyclements:
That’s interesting - I checked the case for these things but clearly not thorough enough. I’ll have a look tomorrow and hopefully will fix it

lilyclements:
@rdstern this error is because you renamed it as "R.Instat", but did not remove and reenter the variable into the dialog. So the code is looking for "R_Instat" which no longer exists. It would occur on other dialogs and has been mentioned in #1504

PR #7175, opened 2022-01-27
Added distributions to show model, random samples and general dialogs

### LINKED ISSUE #7125: Add distributions from stats package for presentation, fitting and random sampling
This is not urgent, but useful when a programmer needs a new (small) task.
A number of dialogues make use of the set of distributions.  They are in particular the 
Model > Probability Distributions > Show Model and also 
Model > Probability Distributions > Random Samples and
Model > One Variable > General

a) There are 4 distributions missing from the stats package and these could usefully be added and checked: This could be done together or any one added would be useful:
a1) Beta distribution
a2) Cauchy Distribution
a3) Hypergeometric distribution
a4) Multinomial distribution

b) The birthday distribution is useful.  First check (script window by tweaking from another distribution) that it is useful for the show model dialogue.  Then it is different, because it is not relevant for the random samples or the fit model dialogues.

c) Similarly check if arima.sim can be added (just called arima) This is to simulate from an arima process.  It is for random samples, but not for the probability distributions or the fitting.  

d) We already have chi-square and t distributions.  But they each could have a non-centrality parameter.  This should be easy to add.  It applies generally to the show model, random sampling and fitting.

Vitalis95:
Fixes (partially)  #7125
@rdstern @shadrackkibet @lloyddewit  I have added Hypergeometric distribution and non-centrality parameter to chi-square and t distributions.
However, adding the remaining distributions has been a challenge. It is a bit complicated implementing the functions because it affects the 3 dialogs.  Please I need your help on this.
Thanks


lloyddewit:
@Vitalis95 Thank you for tackling this issue.
In your comment, I changed 'Partially fixes #7125' to 'Fixes (partially) #7125'. Otherwise, the issue willl be closed automatically when this PR merges.
You requested help. Do you need help:
- understanding the text in the issue?
- or what the dialogs should look like?
- or how to code the dialog changes?

thanks for clarifying

Vitalis95:
@lloyddewit thanks for that change.
The help I need is how to code the dialog changes. Thanks

lloyddewit:
@Vitalis95 Thank you for the clarification. @Ivanluv has good experience coding dialog boxes, maybe you could ask him initially for advice?

Vitalis95:
@rdstern @shadrackkibet @dannyparsons 
The multinomial distribution works well in random sample dialog however, in show model dialog I get the following errors,
For quantiles;
qmultinom

For probabilities;
pmult

The multinomial distribution from stats package has the following functions;
rmultinom(n, size, prob)
dmultinom(x, size = NULL, prob, log = FALSE)

qmultinom and pmultinom which brings the error are not there. 
Thanks.

Vitalis95:
@rdstern @shadrackkibet 
I have added multinomial distribution, it works well in random sample dialog. However, in show model dialog , I'm getting an error and I have shared above. Thanks

rdstern:
@Vitalis95 I have had a look and am impressed that you got this working for the random samples even!  I failed and so looked on [stackoverflow here](https://stackoverflow.com/questions/59160552/random-number-generation-from-multinomial-distribution-in-r-using-rmultinom-fu).  

The answer there indicated that what I think is useful is sort of the multiple equivalent of bernoulli, rather than real multinomial.  - which gives vectors.  There it indicates that we could use sample instead.

Then I checked with bernoulli and found it gives an error in the probability distributions dialogue, that we need to fix.  That's because it uses the binomial function, without saying it is for one trial.

I wondered whether - instead - there is a bernoulli function.  So, I now suggest as follows for both problems.

a) Add the mc2d package
b) Swap the bernoulli method (using binom) by the bern function from this package
c) Use the empiricalD function for the (sort of) multinomial sampling.  For example rempiricalD(43,1:5, prob=c(10, 10, 10, 0, 10))

I suggest that the user has to give the prob values as here, - single control and commas between.  At least 2 values I guess but perhaps 1 would work!  It would be trivial.  Then provide the values - 1 to however many probabilities.

I wonder what we should call it - perhaps multiple?

Then, if this works and you still feel like more, the triangular distribution is also in the same package and could be easy to add?

Vitalis95:
@rdstern @shadrackkibet , I have added discrete empirical distribution and triangular distribution from mc2d package.
For the bernoulli distribution, I had discussions with @shadrackkibet and he suggested I  retain the binom function but I add size=1 (the number of trials).

shadrackkibet:
@Vitalis95 I made the following additions;

- Loaded the package so that distribution work on Show Model.
- Added `mc2d` package as a new package to be installed in the next release.

Pull changes and test if it works from your end.

Vitalis95:
@shadrackkibet , I have pulled and tested. It works

rdstern (changes requested):
Looks good.  One trivial point that was there before.  could bernoulli please be spelled with the extra l.
Then I am concerned about the freedom in the typing in the empirical distribution.  I tried a negative value in the probabilities and it gave a sensible error message.  So that's ok.
But I wonder about the freedom in choosing the discrete values.  I tried 1,2,3,4,5 and it didn't like that.  I tried 1:10 when there were still only 5 probabilities and it was not happy.  
My suggestion is to be more restrictive.  Make the values obviously non-editable and always 1:x.  If I change the probabilities then it changes the second value?  So 1,1,1 would change the values into 1:3.  So you can easily do 3 point, 5, point, 7 point etc, and it always gives probabilities of the integers 1 upwards.

Vitalis95:
@rdstern , @shadrackkibet  I have changed the values in discrete empirical  to be non-editable and also when changing the probabilities then it changes the second value automatically.
For the birthday distribution, I was getting an error while generating quantiles and probabilities in R, I asked the mosaic people how to use birthday distribution from stats package in mosaic package, and they said the names of the functions are not really part of a standard suite of d-, p-, and q-functions for a family of distributions.
Here is the respond; have a look at it.
https://github.com/ProjectMOSAIC/mosaic/issues/792

Thanks.

Vitalis95:
@rdstern  I have added arima.sim to Random sample dialog to Simulate from an ARIMA model. It has AR parameter.
I will be adding a checkbox that Plots several time series . For time being, have a look at this.

rdstern (changes requested):
@Vitalis95 that was a good try on the arima.  But I have realised it is much more complicated than that.  It will probably become a button at the top, so there will be Independent and ARIMA.  But that can wait until we have more on time series.
So aplogies, but let's not have the ARIMA checkbox for now.
The empirical looks much better now.
Could you add a tooltip when there is that option, saying Input relative values separated by commas, For example 3,2,1.
In the field, don't make the user type the c( ).  Instead you add that in the code.
Then let's merge!

Vitalis95:
@rdstern  I have added the tooltip and also c() in the code.
Have a look at it.

rdstern (approved):
Looks good

instat/ucrDistributions.vb
```diff
@@ -29,6 +29,7 @@ Public Class ucrDistributions
     Private bParameterIsDistFunction As Boolean = False
     Private bFunctionIsDistFunction As Boolean = False
     Private bParameterIsDistName As Boolean = False
+    Private intNumber As Integer
```
lloyddewit, line 32: remove?

instat/ucrDistributionsWithParameters.vb
```diff
@@ -110,12 +114,20 @@ Public Class ucrDistributionsWithParameters
         OnControlValueChanged()
     End Sub
 
-    Private Sub ucrInputParameter2_ControlValueChanged() Handles ucrInputParameter2.ControlValueChanged
+    Private Sub ucrInputParameter2_ContentsChanged() Handles ucrInputParameter2.ContentsChanged
         If lstCurrArguments IsNot Nothing AndAlso lstCurrArguments.Count > 1 Then
-            AddParameter(lstCurrArguments(1), ucrInputParameter2.GetText)
+            If clsCurrDistribution.strNameTag = "Discrete_Empirical" Then
+                ttEmpirical.SetToolTip(ucrInputParameter2.txtInput, "Input relative values separated by commas, For example 3,2,1.")
```
lloyddewit, line 120: This tool tip is set every time `ucrInputParameter2` changes.
Should we just set the tool tip once somewhere else?

Vitalis95:
@lloyddewit have a look at it,
thanks

rdstern (approved):

shadrackkibet (approved):

lloyddewit (approved):

PR #7177, opened 2022-01-28
Added Up to button for importing multiple files on Import Dataset Dialogue

N-thony:
There is not yet an issue relative to this PR, this is just a feature @rdstern requested on the import of multiple datasets, to be able to import Files with a given extension in a folder and sub folders.
@rdstern could you test this if it works as you expected? Thanks.


rdstern (changes requested):
@N-thony that's great it worked.  Good if @lloyddewit could check the code.
a) I think I'd prefer a slightly more discrete black up-arrow?  @lloyddewit do you have thoughts on this?
b) After it worked I pressed the up-arrow again and it crashed.  I'd be quite happy if it was only for a single level up.  (I had done that before and it was ok - so don't know what the difference is here?)

N-thony:
> @N-thony that's great it worked. Good if @lloyddewit could check the code. a) I think I'd prefer a slightly more discrete black up-arrow? @lloyddewit do you have thoughts on this? b) After it worked I pressed the up-arrow again and it crashed. I'd be quite happy if it was only for a single level up. (I had done that before and it was ok - so don't know what the difference is here?)

@rdstern thanks for your comments. I changed the arrow button as requested and fixed some bugs found while testing. Could you also re-test?

rdstern (approved):
@N-thony great.  This is fine now.  I hope @shadrackkibet can test and then it could be merged.  (Not necessary for 0.7.4)
To test you need files of the right type.  I had directories each with one SPSS file (SAV) in their own directory.  Could try in the Instat climatic library with csv files.
I wonder if the up arrow could be a bit smaller - more discreet, but I am not a good person to advise on design and it is fine as it is.

N-thony:
> @N-thony great. This is fine now. I hope @shadrackkibet can test and then it could be merged. (Not necessary for 0.7.4) To test you need files of the right type. I had directories each with one SPSS file (SAV) in their own directory. Could try in the Instat climatic library with csv files. I wonder if the up arrow could be a bit smaller - more discreet, but I am not a good person to advise on design and it is fine as it is.

@rdstern you are right, I reduced the up arrow button size. Thanks.

instat/dlgImportDataset.vb
```diff
@@ -1059,12 +1067,23 @@ Public Class dlgImportDataset
         TestOkEnabled()
     End Sub
 
+    Private Sub SetUpButtonEnable(strCharacter As String)
+        cmdStepBack.Enabled = If(strCurrentDirectory.Count(Function(x) x = strCharacter) <= 1, False, True)
+    End Sub
+
+    Private Sub cmdStepBack_Click(sender As Object, e As EventArgs) Handles cmdStepBack.Click
+        SetDialogStateFromFile(Strings.Left(strCurrentDirectory, InStrRev(strCurrentDirectory, If(strCurrentDirectory.Contains("/"), "/", "\")) - 1), strFileExtension)
```
lloyddewit, line 1075: If we broke this into separate statements (and/or with line breaks), and added comments, then would it be more readable?

If `strCurrentDirectory` contains no file separators, then `Strings.Left()` will be called with `Length` parameter set to -1. The `Strings.Left()` documentation does not define what happens in this case. Does it trigger an exception? Would it be safer to set `Length` to 0 if there is no file seperator (or even better, set the path to "")?

instat/dlgImportDataset.vb
```diff
@@ -29,6 +29,7 @@ Public Class dlgImportDataset
     Private strFilePathSystemTemp As String = ""
     Private strFilePathR As String = ""
     Private strCurrentDirectory As String = ""
+    Private strCurrDirectory As String = ""
```
lloyddewit, line 32: We have `strCurrentDirectory` and `strCurrDirectory`, this is confusing.
Is `strCurrDirectory` only used when we're importing a single file? What's the reason for having a second variable for this?
If we need both, then could we give one a better name to make it clear what its purpose is?

instat/dlgImportDataset.vb
```diff
@@ -1092,6 +1111,10 @@ Public Class dlgImportDataset
         Return frmMain.clsRLink.MakeValidText(strCleanFileName)
     End Function
 
+    Private Sub ucrChkMultipleFiles_Load(sender As Object, e As EventArgs)
+
+    End Sub
+
```
lloyddewit, line 1117: do we need this?

instat/instat.vbproj
```diff
@@ -5166,6 +5166,10 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
     <None Include="bin\Debug\static\Library\Climatic\Niger\Niger_4stations_corrections.R" />
+    <None Include="Resources\up_to.png" />
+    <None Include="Resources\up_to_btn.png" />
+    <None Include="Resources\upward-arrow.png" />
+    <None Include="Resources\upward-arrow1.png" />
```
lloyddewit, line 5172: Are all these 'png' files used in the project or could we remove some of them?

lloyddewit (changes requested):
@N-thony Thanks for the good work. I have a few questions

PR #7178, opened 2022-01-28
Adding the sub dialogue for formatting of Table

Ivanluv:
@rdstern @lloyddewit I have the two branches #7107 and #7172 incorporated in this PR


rdstern (approved):
@Ivanluv and @lilyclements that's great now.  I would be happy to see this merged if @lloyddewit is ok with the code.
If @lilyclements can resolve the remaining "feature", then that's a bonus.  othersie it can remain to the new version.

lilyclements (commented):
This is looking good to me. There's some small things that can be changed/tidied up but that can wait for after the release.

For example, tidying up the sub dialog and main dialog aesthetically so things are aligned and what not. Adding a footnote currently adds it to every cell of the table which we probably do not want. On header tab you currently can only edit either the rows or columns - not both. Is this the case in R if you run the code? Etc etc

lilyclements:
I've commented on the bug in #7172. It's a bug first mentioned in issue #1504 and is to do with changing a variable name but not removing and re-entering it into the receiver. I think then this is ready to merge

dannyparsons (approved):

PR #7184, opened 2022-02-01
Added up arrow button to Import Dataset Dialog

### LINKED ISSUE #7154: Check sub-directories when using File > Import with multiple files
Only add this feature if it is a small job! (@dannyparsons says this is easy to do. )

A powerful feature of this dialogue is the ability to import multiple files of the same type into successive data frames.
The way this is used is fine and is as follows:
a) Go to a directory that contains these multiple files, e.g. 8 SPSS files (each xxx.sav). Click on any one of them.
image

b) Use the "All files of this type" checkbox, to go up a directory.  It then allows the importing of all these files.  (We could add checkboxes later to be able to select the files we want - default would be all?
image

c) However I now also have an example where this is not quite enough.  These are DHS data, also SPSS files,  that can be unzipped.  This gives 8 directories, each with 3 files, one of which is the SAV file.  So going up to the directory is not enough.  I would like to go up one more level:

image

What I would like is now another checkbox after the "All files of this type" .  It is only visible when All files is checked. It is disabled by default.  It is followed by the up-arrow symbol you see on the browse dialogue.  (So the whole control is small and unobtrusive.  If checked it moves up another directory - to DHS in this case.

The detection of the multiple files than has also to register those it finds from sub-directories too.

N-thony:
Fixes #7154
@rdstern could re-test this? Thanks.
@lloyddewit I made the change as requested in #7177. Could you check again? Thanks.


rdstern (dismissed):
Looks fine still

lloyddewit:
@N-thony is this ready for review now or do you still plan changes? thanks

N-thony:
> @N-thony is this ready for review now or do you still plan changes? thanks

yes @lloyddewit, we can also have a quick on that if it is okay with you.

rdstern:
@lloyddewit is this now ok to merge?

instat/dlgImportDataset.vb
```diff
@@ -1059,12 +1067,27 @@ Public Class dlgImportDataset
         TestOkEnabled()
     End Sub
 
+    Private Function GetCorrectSeparatorInPath() As String
+        Return If(strCurrentDirectory.Contains("/"), "/", "\")
+    End Function
+
+    Private Sub cmdStepBack_Click(sender As Object, e As EventArgs) Handles cmdStepBack.Click
+        If GetCorrectSeparatorInPath() <> "" Then
+            SetDialogStateFromFile(Strings.Left(strCurrentDirectory, InStrRev(strCurrentDirectory, "\") - 1), strFileExtension)
```
lloyddewit, line 1076: if the separator is '/', then will this fail?

lloyddewit (changes requested):
@N-thony thanks, just one question

instat/dlgImportDataset.vb
N-thony: In this case, yes, I have added the check of the correct separator in the path. Thanks.

instat/dlgImportDataset.vb
```diff
@@ -1059,12 +1067,27 @@ Public Class dlgImportDataset
         TestOkEnabled()
     End Sub
 
+    Private Function GetCorrectSeparatorInPath() As String
+        Return If(strCurrentDirectory.Contains("/"), "/", "\")
+    End Function
+
+    Private Sub cmdStepBack_Click(sender As Object, e As EventArgs) Handles cmdStepBack.Click
+        If GetCorrectSeparatorInPath() <> "" Then
```
lloyddewit, line 1075: `GetCorrectSeparatorInPath` always returns `/` or '\`, therefore this `If` condition will always be true and the `Else` will never be reached.

lloyddewit (changes requested):
@N-thony Thanks for resolving the last comments, I just spotted something else

instat/dlgImportDataset.vb
N-thony: You are right, I just resolved this. Thanks.

lloyddewit: I think we still need to check if `strCurrentDirectory` contains at least one separator.
If not, then the line below may give strange results.
```
        SetDialogStateFromFile(Strings.Left(strCurrentDirectory, InStrRev(strCurrentDirectory, GetCorrectSeparatorInPath()) - 1), strFileExtension)
```
I suggest that you put the `If/Else` statements back, and change the `If` condition so that it checks that there is at least one separator.
Does this make sense?

N-thony:
@lloyddewit I made the change as requested. Please take a look. Thanks.

instat/dlgImportDataset.vb
```diff
@@ -1059,12 +1067,33 @@ Public Class dlgImportDataset
...
     End Sub
 
+    Private Function GetCorrectSeparatorInPath() As String
+        Dim strSepearator As String = ""
+        If strCurrentDirectory.Contains("/") Then
+            strSepearator = "/"
+        ElseIf strCurrentDirectory.Contains("\") Then
+            strSepearator = "\"
+        End If
+        Return strSepearator
```
lloyddewit, line 1077:
```suggestion
        Dim strSeparator As String = ""
        If strCurrentDirectory.Contains("/") Then
            strSeparator = "/"
        ElseIf strCurrentDirectory.Contains("\") Then
            strSeparator = "\"
        End If
        Return strSeparator
```

lloyddewit (changes requested):

lloyddewit:
@rdstern Just a couple of small changes since you last approved, please could you test again? thanks

N-thony:
> @rdstern Just a couple of small changes since you last approved, please could you test again? thanks

@rdstern how about this?

rdstern (approved):
still fine to me!

PR #7185, opened 2022-02-01
Made Text file preview be shown on Linux Version

### LINKED ISSUE #7113: Text file preview doesn't display in Linux version
The text file preview in the Import from File dialog doesn't display in the Linux version.

I think this is because it is directly reading the file from VB and the path uses `\` instead of `/`, which won't work in Linux.

If this is changed to `/` in the code, it should work for Windows and Linux.

If someone can make this change and confirm it works in Windows still then I can test in Linux too.

image

Gioche6:
Fixes #7113

@africanmathsinitiative/developers , This is ready for review
@dannyparsons , Please test in the Linux version and see if the text file preview appears.


instat/dlgImportDataset.vb
```diff
@@ -729,7 +729,7 @@ Public Class dlgImportDataset
             Dim iLinesToPreview As Integer = ucrNudPreviewLines.Value + If(IsCSVFileFormat(), ucrNudRowsToSkipCSV.Value, ucrNudRowsToSkipText.Value)
             lblTextFilePreview.Text = "Text File Preview:"
             Try
-                Using sReader As New StreamReader(strFilePathSystem)
+                Using sReader As New StreamReader(strFilePathSystem.Replace("\", "/"))
```
lloyddewit, line 732: is this cleaner/safer? if you prefer the hardcoded characters then that's also ok
```suggestion
                Using sReader As New StreamReader(
                        strFilePathSystem.Replace(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar))
```

lloyddewit (approved):

Gioche6:
@lloyddewit , Thank you for that suggestion. I've merged it here.

lloyddewit (approved):

instat/dlgImportDataset.vb
```diff
@@ -729,7 +729,8 @@ Public Class dlgImportDataset
             Dim iLinesToPreview As Integer = ucrNudPreviewLines.Value + If(IsCSVFileFormat(), ucrNudRowsToSkipCSV.Value, ucrNudRowsToSkipText.Value)
             lblTextFilePreview.Text = "Text File Preview:"
             Try
-                Using sReader As New StreamReader(strFilePathSystem)
+                Using sReader As New StreamReader(
+                        strFilePathSystem.Replace(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar))
```
Patowhiz, line 733: @Gioche6 this looks good. 
@lloyddewit @dannyparsons I understand that in the same dialog, we have other instances that are doing this kind of separator replacements but using hardcoded "\" with "/". It would be useful (even to future developers) to have @Gioche6 also make those lines consistent with this one.
The bigger task would be to make this kind of replacements consistent in the whole software in a way that aligns with the Single Responsibility Principle.

lloyddewit: I see the value of a consistent approach within this class, and throughout R-Instat.
The dilema is do we refactor a whole class if we just change one line? 
My personal view:
- Refactoring takes time and introduces risk. It is best done as a task on its own.
- I think it's normally OK to refactor the specific code we're changing, or even the whole function.

For this PR, my vote is not to refactor, and limit the changes to line 733.

Of course its highly subjective and opinions may vary!

Patowhiz: @lloyddewit I agree, it can be done on a separate PR. And not just here, we have other classes that need the same too. I'll open an issue on the same.

lloyddewit:
@dannyparsons @ChrisMarsh82  Please could one of you test on Linux? thanks

ChrisMarsh82 (approved):

PR #7188, opened 2022-02-02
Fixing Reset issue in Prepare>Numeric>Transform dialog 

### LINKED ISSUE #7076: Small error on the Position control when using the same dialogue with different data-frames
The Position control in this dialogue does not update, when the data frame is changed.

The default position is after the variable being selected.

I used the dialogue with the rice survey data and multiplied the yield, by 67.25 to put them into kg per ha  It gave an error mentioning a variable Distance (from a previous data frame.  I forced it to work, and then moved to another data frame.  Again the error, this time mentioning the yield from the survey data.

anastasia-mbithe:
This partially fixes #7076 
Hello, @rdstern and @shadrackkibet. This fixes the reset issue in the Numeric Transform dialog.


rdstern (commented):
@anastasia-mbithe I am now confused by this.  I tested this and the error is still there.  But it now seems ok in the master version, so I think the error has been fixed in the position control.

In that case surely this can simply be closed?

anastasia-mbithe:
@rdstern if the issue was fixed in the position control then am good if its closed.

lloyddewit:
I closed because of comments above. We can always reopen if needed.

anastasia-mbithe:
@rdstern, we should consider reopening this PR, since PR #7076 did not fix the reset issue. I went ahead this morning to test the changes in the master and the reset issue is still there and this PR fixes that. 
You can consider re-testing the master.

rdstern (approved):
@anastasia-mbithe great.  This now seems to work well

instat/dlgTransform.vb
```diff
@@ -486,81 +486,82 @@ Public Class dlgTransform
...
-                ucrBase.clsRsyntax.SetBaseROperator(clsPowerOperator)
-            ElseIf rdoLogToBase10.Checked Then
-                clsPreviewTextFunction = clsLogBase10Function.Clone
-                clsNonNegativeDummyFunction.AddParameter("check", "log10", iPosition:=0)
-                ucrBase.clsRsyntax.SetBaseRFunction(clsLogBase10Function)
-            ElseIf rdoNaturalLog.Checked Then
-                clsPreviewTextFunction = clsNaturalLogFunction.Clone
-                clsNonNegativeDummyFunction.AddParameter("check", "log", iPosition:=0)
-                ucrBase.clsRsyntax.SetBaseRFunction(clsNaturalLogFunction)
+    Private Sub BaseFunctions()
```
shadrackkibet, line 489: This `sub` has only 1 reference. I suggest you remove the `sub` and move the code to where it is referenced.

shadrackkibet (changes requested):

shadrackkibet (approved):

instat/dlgTransform.vb
shadrackkibet: I will fix this while working on #7236. Please ignore it.

PR #7200, opened 2022-02-04
Enabled grid to reflect the current columns selected

shadrackkibet:
@africanmathsinitiative/developers this is ready for testing.

This PR allows columns selected to be applied onto the grid while updating metadata windows accordingly.

@rdstern  I would like to hand this over to another developer to update the colour of the columns on the grid so as to reflect the current column selection. The display below should also show the column selection? What about adding it at the top?

What are your thoughts?


rdstern (dismissed):
@shadrackkibet this seems to work well.  @lloyddewit over to you.  There is more to do in this general area, but good to get this merged first.

lloyddewit (changes requested):

lloyddewit (approved):
@shadrackkibet thanks, looks good to me

lloyddewit:
@rdstern One small bug fix since your last approval, please could you test again? thanks

rdstern (approved):
@lloyddewit and @shadrackkibet I am approving in case the one curiosity I found should not be part of this, but fixed later.
I took a data frame with 11 variables and selected 5 of them.  All worked fine except it now says at the bottom that we now have 1 to 5 of 11 variables.  So two arrow keys are now available (which they are not with all 11 variables.  If you select, then it thinks it goes to 31 upwards.  So that control is not distinguishing between not visible and not selected.

lloyddewit:
> @lloyddewit and @shadrackkibet I am approving in case the one curiosity I found should not be part of this, but fixed later. I took a data frame with 11 variables and selected 5 of them. All worked fine except it now says at the bottom that we now have 1 to 5 of 11 variables. So two arrow keys are now available (which they are not with all 11 variables. If you select, then it thinks it goes to 31 upwards. So that control is not distinguishing between not visible and not selected.

@shadrackkibet I didn't completely understand this comment, does it require a new issue/PR?

shadrackkibet:
Let us consider this as a new issue that will then translate to a new PR. I will check with Roger if we have captured it in one of the issues or we open a new one.

PR #7207, opened 2022-02-07
Enabled running multiple instances of R-Instat

### LINKED ISSUE #6950: Enable the possiblity to run simultaneaoulsy multiple instances of R-Instat?
This option was recently disabled after introducing the open from file feature i.e. the possibility to load a dataset into R-Instat directly by right-clicking on the file (.RDS file). The reason that multiple instances were disabled is to avoid users being confused when trying to load the dataset by right-clicking on the dataset file into R-Instat when there are multiple instances.

This is easy to fix but I did a quick check on that and then I found that when trying to load a dataset into R-Instat when right-clicking on its file when there are multiple instances, it load the dataset into the last instance of R-Instat that was launched and this is a bit tricky if the user to load it into the first instance for example.

This corresponds to item f) in issue #6947

N-thony:
Fixes #6950 

### DIFF
--- a/instat/Interface/IDataViewGrid.vb
+++ b/instat/Interface/IDataViewGrid.vb
@@ -19,6 +19,8 @@ Public Interface IDataViewGrid
 
     Event CellDataChanged()
 
+    Event DeleteValuesToDataframe()
+
     Event PasteValuesToDataframe()
 
     Event ReplaceValueInData(strNewValue As String, strColumnName As String, strRowText As String)
--- a/instat/Interface/IDataViewGrid.vb
+++ b/instat/Interface/IDataViewGrid.vb
@@ -19,8 +19,6 @@ Public Interface IDataViewGrid
 
     Event CellDataChanged()
 
-    Event DeleteValuesToDataframe()
-
     Event PasteValuesToDataframe()
 
     Event ReplaceValueInData(strNewValue As String, strColumnName As String, strRowText As String)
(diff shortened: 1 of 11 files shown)


rdstern (commented):
I am not sure what I can test here.  It now tries to run an old R-Instat - I have them installed still to be able to check back.  And that doesn't allow multiple instances.  I hope @shadrackkibet and @lloyddewit can perhaps check properly?

N-thony:
> I am not sure what I can test here. It now tries to run an old R-Instat - I have them installed still to be able to check back. And that doesn't allow multiple instances. I hope @shadrackkibet and @lloyddewit can perhaps check properly?

This option of running a `single instance application` was enabled in the recent versions, for example, if you have `0.7.2` and `0.7.3.1` you can't run them simultaneously. In this PR I have disabled that option as requested, therefore you can run multiple instances from this PR with any recent version of R-Instat but it will run only one of the recent where this option was activated.
image

shadrackkibet:
@rdstern I tried the following and it seems to work. Perhaps you can try as well?

- Open R-Instat 0.7.4
- Switch to this branch in Visual Studio (See below)
a) image

b) image

- Click Start button in Visual Studio

Result;

Both Instances of R-Instat Run i.e Development version and R-Instat 0.7.4 as shown below;
image

shadrackkibet (approved):

PR #7208, opened 2022-02-08
Added options to the Prepare > Data Frame > Rename Column dialogue

### LINKED ISSUE #6877: Add options to the Prepare > Data Frame > Rename Column dialogue
A trivial point to include in the next update on frmMain is to change Rename Column... to Rename Columns...  It is now also for multiple columns.

We can have Variable Labels as well as Variable Names.  R itself isn't very good on Variable Labels, but I would like to encourage their use more in R-Instat.  I like what SPSS and other packages encourage, namely relatively short variable names plus variable labels when needed.   In some of our examples the variable names are very long (up to 10 words) and no use is made of the variable labels.

There is always a space for them in the column metadata.  I have been asking about the possibility of pasting there.  But anyway I suggest we also add facilities also into the Rename dialogue, possibly as follows:

In the Single option I suggest we have a group box with label Column (Variable) Label:
It is a bit to the left of the current label.
Inside there are 3 radio buttons.
O Enter  - this is the initial default
O Copy From Current Name
O Copy From New Name
Then there is the field for the label, which is made wider than the name fields.

Now trickier is what to do on Multiple?  It is possible we need another top button, or we add to the options here.  Perhaps there are 2 ordinary radio buttons for Names Only and Include Labels.  Or there could be a checkbox, default unchecked, saying Include Variable Labels.  The control then needs 2 columns, one for the names and the other for the labels.  Then it will be easy to paste into the variable labels.  And there could be an option Copy names to Labels here.

I suggest the operations on the names (and there will be more) could be separate from those on Variable labels.  So, when the Variable Labels option is used, then the options relate just to the labels.  

I am suggesting Shadrack should check on a possible structure.  Then perhaps @N-thony could do the work, once agreed?

N-thony:
Fixes #6877 
This is still a work in progress. @shadrackkibet I have added the grid under the multiple options as suggested by @rdstern in the corresponding issue. At this stage, I would like your help on the backend side. I'm not sure about how the R code or proper package to use for renaming multiple variables through the grid.

### DIFF
--- a/instat/dlgName.vb
+++ b/instat/dlgName.vb
@@ -29,7 +29,6 @@ Public Class dlgName
     Private clsRenameColumns As New RFunction
     Private WithEvents grdCurrentWorkSheet As Worksheet
     Private bIncludeCopyOfLevels As Boolean
-    Private strCurrentName As List(Of String)
     Private dctRowsChanged As New Dictionary(Of Integer, String)
     Private dctNameRowsValues As New Dictionary(Of Integer, String)
 
@@ -74,10 +73,8 @@ Public Class dlgName
 
         ucrPnlOptions.SetParameter(New RParameter("type", 4))
         ucrPnlOptions.AddRadioButton(rdoSingle, Chr(34) & "single" & Chr(34))
-        ucrPnlOptions.AddRadioButton(rdoMultiple, Chr(34) & "multiple" & Chr(34))
+        ucrPnlOptions.AddRadioButton(rdoMultiple, Chr(34) & "" & Chr(34))
         ucrPnlOptions.AddRadioButton(rdoRenameWith, Chr(34) & "multiple" & Chr(34))
-        ucrPnlOptions.SetRDefault(Chr(34) & "single" & Chr(34))
-        'rdoMultiple.Enabled = False
 
         ucrPnlOptions.AddToLinkedControls({ucrReceiverName, ucrInputNewName, ucrInputVariableLabel}, {rdoSingle}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
         ucrPnlOptions.AddToLinkedControls(ucrReceiverColumns, {rdoRenameWith}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
@@ -91,7 +88,6 @@ Public Class dlgName
         ucrInputCase.SetLinkedDisplayControl(lblCase)
         ucrPnlCase.SetLinkedDisplayControl(grpOptions)
         ucrChkIncludeVariable.SetLinkedDisplayControl(grdRenameColumns)
-        ucrChkIncludeVariable.Enabled = False
 
         ucrPnlCase.SetParameter(New RParameter(".fn", 5))
         ucrPnlCase.AddRadioButton(rdoToLower, "tolower")
@@ -171,9 +167,9 @@ Public Class dlgName
         AddChangedRow(e.Cell.Row, e.NewData)
 
         If grdCurrentWorkSheet IsNot Nothing Then
-            For Each i As Integer In dctNameRowsValues.Keys
+            For i As Integer = 0 To dctNameRowsValues.Keys.Count - 1
                 If dctRowsChanged.ContainsKey(i) Then
-                    clsRenameColumns.AddParameter(dctRowsChanged(i), strCurrentName(i), iPosition:=i + 1)
+                    clsRenameColumns.AddParameter(dctRowsChanged(i), dctNameRowsValues(i), iPosition:=i + 1)
                 End If
             Next
         End If
@@ -215,6 +211,13 @@ Public Class dlgName
         If bIncludeCopyOfLevels Then
             grdCurrentWorkSheet.AppendColumns(1)
             grdCurrentWorkSheet.ColumnHeaders(grdCurrentWorkSheet.ColumnCount - 1).Text = "New Label"
+            For i As Integer = 0 To GetListColsLabel().Count - 1
+                For j As Integer = 0 To grdCurrentWorkSheet.Columns - 1
+                    If grdCurrentWorkSheet.ColumnHeaders(j).Text = "New Label" Then
+                        grdCurrentWorkSheet.Item(row:=i, col:=j) = GetListColsLabel().Item(i)
+                    End If
+                Next
+            Next
         End If
 
         grdCurrentWorkSheet.SetRangeDataFormat(New RangePosition(0, 0, grdCurrentWorkSheet.Rows, grdCurrentWorkSheet.Columns), DataFormat.CellDataFormatFlag.Text)
@@ -290,6 +293,32 @@ Public Class dlgName
             End If
         End If
         Return strColLabel
+    End Function
+
+    Private Function GetListColsLabel() As List(Of String)
+        Dim lstColLabel As New List(Of String)
+        Dim clsColmnLabelsRFunction = New RFunction
+        Dim expItems As SymbolicExpression
+
+        clsColmnLabelsRFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_column_labels")
+        clsColmnLabelsRFunction.AddParameter("data_name", Chr(34) & ucrSelectVariables.strCurrentDataFrame & Chr(34), iPosition:=0)
+        For i As Integer = 0 To dctNameRowsValues.Keys.Count - 1
+            clsColmnLabelsRFunction.AddParameter("columns", Chr(34) & dctNameRowsValues(i) & Chr(34), iPosition:=1)
+            expItems = frmMain.clsRLink.RunInternalScriptGetValue(clsColmnLabelsRFunction.ToScript(), bSilent:=True)
+            If expItems IsNot Nothing AndAlso Not (expItems.Type = Internals.SymbolicExpressionType.Null) Then
+                Dim strArr As String() = expItems.AsCharacter.ToArray
+                If strArr IsNot Nothing Then
+                    'the number of labels for a column expected is 1
+                    If strArr.Length = 1 Then
+                        lstColLabel.Add(strArr(0))
+                    ElseIf strArr.Length > 1 Then
+                        MessageBox.Show(Me, "Developer error: retrieved column label should be one.", "Developer Error", MessageBoxButtons.OK, MessageBoxIcon.Error)
+                        lstColLabel.Add(strArr(strArr.Length - 1))
+                    End If
+                End If
+            End If
+        Next
+        Return lstColLabel
     End Function
 
     Private Sub ucrPnlOptions_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlOptions.ControlValueChanged
--- a/instat/dlgName.vb
+++ b/instat/dlgName.vb
@@ -16,6 +16,8 @@
 
 Imports instat.Translations
 Imports RDotNet
+Imports unvell.ReoGrid
+Imports unvell.ReoGrid.Events
 
 Public Class dlgName
     Dim bFirstLoad As Boolean = True
@@ -24,6 +26,12 @@ Public Class dlgName
     Dim strSelectedColumn As String = ""
     Dim strSelectedDataFrame As String = ""
     Private clsDefaultRFunction As New RFunction
+    Private clsRenameColumns As New RFunction
+    Private WithEvents grdCurrentWorkSheet As Worksheet
+    Private bIncludeCopyOfLevels As Boolean
+    Private strCurrentName As List(Of String)
+    Private dctRowsChanged As New Dictionary(Of Integer, String)
+    Private dctNameRowsValues As New Dictionary(Of Integer, String)
 
     Private Sub dlgName_Load(sender As Object, e As EventArgs) Handles Me.Load
         If bFirstLoad Then
@@ -61,17 +69,15 @@ Public Class dlgName
 
         ucrChkIncludeVariable.SetText("Include Variable Labels")
 
-        ucrFactorLevels.SetVisible(False)
-
         'Label Input
         ucrInputVariableLabel.SetParameter(New RParameter("label", 3))
 
         ucrPnlOptions.SetParameter(New RParameter("type", 4))
         ucrPnlOptions.AddRadioButton(rdoSingle, Chr(34) & "single" & Chr(34))
-        ucrPnlOptions.AddRadioButton(rdoMultiple, Chr(34) & "" & Chr(34))
+        ucrPnlOptions.AddRadioButton(rdoMultiple, Chr(34) & "multiple" & Chr(34))
         ucrPnlOptions.AddRadioButton(rdoRenameWith, Chr(34) & "multiple" & Chr(34))
         ucrPnlOptions.SetRDefault(Chr(34) & "single" & Chr(34))
-        rdoMultiple.Enabled = False
+        'rdoMultiple.Enabled = False
 
         ucrPnlOptions.AddToLinkedControls({ucrReceiverName, ucrInputNewName, ucrInputVariableLabel}, {rdoSingle}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
         ucrPnlOptions.AddToLinkedControls(ucrReceiverColumns, {rdoRenameWith}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
@@ -84,6 +90,8 @@ Public Class dlgName
         ucrReceiverColumns.SetLinkedDisplayControl(lblColumns)
         ucrInputCase.SetLinkedDisplayControl(lblCase)
         ucrPnlCase.SetLinkedDisplayControl(grpOptions)
+        ucrChkIncludeVariable.SetLinkedDisplayControl(grdRenameColumns)
+        ucrChkIncludeVariable.Enabled = False
 
         ucrPnlCase.SetParameter(New RParameter(".fn", 5))
         ucrPnlCase.AddRadioButton(rdoToLower, "tolower")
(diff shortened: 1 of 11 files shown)


shadrackkibet:
@N-thony thanks for the discussion. Leave this to me for now. I will alert you when the R code is ready.

rdstern (changes requested):
@N-thony so far I have just looked at the multiple.  Not yet the rename_with.
a) Looks fine on manual changes.
b) When I paste from Excel they appear to change, but it doesn't register.
c) When I press Ok and have made a change, but not clicked elsewhere, then the last change doesn't register.
d) When I click to include labels it goes blank and freezes.  (And, once it is working please test copying the old names and pasting into the variable labels.

N-thony:
> @N-thony so far I have just looked at the multiple. Not yet the rename_with. a) Looks fine on manual changes. b) When I paste from Excel they appear to change, but it doesn't register. c) When I press Ok and have made a change, but not clicked elsewhere, then the last change doesn't register. d) When I click to include labels it goes blank and freezes. (And, once it is working please test copying the old names and pasting into the variable labels.

@rdstern thanks for your comment. During our discussion we agreed to get the multiple option working well before moving to the rename_with option. Thanks for reminding about the pasting feature on the grid. I made some change regarding your comments above. I didn't yet finish to make a proper test to check if it is stable, but you may wish to test it too at this stage.

rdstern (changes requested):
@N-thony that's progress!  
a) Paste is working, but only for the first cell thaqt is pasted.
b) Adding labels is now sensible.  I only checked pasting, and that seems the same as the names, so it just pastes the first label.

N-thony:
@rdstern I'm still looking into this. Meanwhile, @shadrackkibet I found that with the functions parameters passing the `columns `and `indexes` is a bit tricky. Using the survey dataset, I have the selected the three first column Village, field and size then through my dialogue I renamed both the name and label of village which the index is 1. Then click Ok. All runs fine, when I go back I expected to see only 2 columns here field and size since the new village name is not in the defined filter then, I do rename field column in the grid then I found it's still rename now the new village name column  since now in the grid the index of field is 1. I can demonstrate this if you have a minute for a quick call.

N-thony:
> @N-thony I am not sure I understand what you find tricky. I tried the following;
> 
> * Loaded Survey data
> * Prepare>Data Frame>Rename columns
> * Data Options>Columns>Define New Selection>Selected village,field,size>Add condition>Return>Return
> * Selected Multiple at the top
> * Checked Include Variable Labels checkbox
> * Renamed columns as follows & added new labels (see below);
>   image
> * Click Ok
> * I got the following code (Which is correct)
> 
> ```
> # Code generated by the dialog, Rename Columns
> data_book$rename_column_in_data(data_name="survey", type="multiple", .fn=tolower, new_column_names_df=data.frame(cols=c("village1","field2","size3"), index=c(1,2,3)), new_labels_df=data.frame(cols=c("Village","Field","Size"), index=c(1,2,3)))
> ```
> 
> Output in the Data View and Column Metadata is as follows; image image
> 
> From my analysis above I do not see any problem with the R code. There are definitely efficiency issues at the dialog level.
> 
> By the way, remove `.fn=tolower` from code when multiple and single options are selected.

@shadrackkibet thanks for reproducing that. I meant if you only rename the first row as below, it is fine.
image

Then back to the dialog, you will have the following column present in the grid since the renamed column is not in the selection condition. Then try to rename them, `field` and `size`
image

N-thony:
I meant by tricky since the indexes are supposed to come from the data frame?

rdstern (changes requested):
@N-thony did you want me to check again now - well here is what I found mon multiple.
a) It automatically used all the variables.  With the usual data selector I assumed you would start with it empty and then choose which variables to rename?  
b) If I manually edit, then it seems fine.  That is with the exception that I have to exit the last change, or it is not used.  I suggest either this change is closed (and used) when you press another key.  Alternatively (not quite so good) Ok and To Script) could be disabled when you are changing a name.
c) Paste still doesn't work, when you are pasting multiple cells.  I tried both pasting new names from Excel and also copying names into labels within the dialogue.  Just the first change is implemented, e.g. just x1, if I paste x1 to x7 in one go.

rdstern (changes requested):
@N-thony that's progress, but there is quite a lot to say.
a) Possibly change the title from New Label to Label.
b) I pasted into the labels, but OK was not enabled, until I changed a name.
c) It pasted correctly.  But when I returned, those labels did not appear in your grid.
d) I was not able to paste into the new names, largely because of that very annoying message.  I suggest it is turned off.  
e) So I entered names manually aqnd ignored the message which came repeatedly.  I deliberately now has a repeated name, and pressed Ok and it crashed.  I suggest you don't do any checks when the new names are being done, but check when Ok is pressed. If the set of new names are not valid, then decide on an action.  Probably don't change them, or perhaps give an option to tidy, i.e. make valid and unique.  (I am sure we must have the code for that when importing a data file?

rdstern (changes requested):
This is looking good.  You may need help mon the R code for the rename_with.  I wonder if @HawardKetoyoMsatsi can help.  It may have to be @shadrackkibet , but Haward is pretty good with R.
On 2 details, with the Abbreviation, the up-down can be less wide.  It will never need more than 2 numbers.  The default is 8 and the minimum is 1 for now.  (It may become 2).  
And after the up/down, put characters.  
Then, not only on this dialogue, and perhaps a separate pull request, could you look carefully at the details of the regex keyboard.  The cursor is not always in the right place, after each key.

N-thony:
@rdstern I and @shadrackkibet agreed that the changes on `Rename With` #6877 will be done in the separate PR. Here are just the changes concerning the `Multiple` option.
@shadrackkibet I would like to be able to also have the label names in the Label on filling the grid, I'm not sure there is an R code doing that. I have done with the one returning a single label name and made a list of that on the front end side but this makes the grid not fill faster. Do you have any suggestion on that?

shadrackkibet:
What about `data_book$get_column_labels()`? Did you try it?

instat/dlgName.vb
```diff
@@ -199,12 +393,62 @@ Public Class dlgName
...
+        Dim clsColmnLabelsRFunction = New RFunction
+        Dim expItems As SymbolicExpression
+
+        clsColmnLabelsRFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_column_labels")
+        clsColmnLabelsRFunction.AddParameter("data_name", Chr(34) & ucrSelectVariables.strCurrentDataFrame & Chr(34), iPosition:=0)
+
+        'This is to be done in R to avoid the app to slow down when getting the list of label
+        For i As Integer = 0 To dctNameRowsValues.Count - 1
+            clsColmnLabelsRFunction.AddParameter("columns", Chr(34) & dctNameRowsValues(i) & Chr(34), iPosition:=i)
+            expItems = frmMain.clsRLink.RunInternalScriptGetValue(clsColmnLabelsRFunction.ToScript(), bSilent:=True)
```
shadrackkibet, line 407: Instead of going to R in each loop why don't you just get labels once? Then populate the grid with the labels.

N-thony:
> What about `data_book$get_column_labels()`? Did you try it?

Sounds good, let me try it then.

N-thony:
> > What about `data_book$get_column_labels()`? Did you try it?
> 
> Sounds good, let me try it then.

@shadrackkibet could check this and probably add the check for uniqueness of column names?

N-thony:
@shadrackkibet how about this? Any update? Thanks.

shadrackkibet:
@N-thony the check for duplicate column names is now added. There are "things" you need to stop a user from typing at the grid level. 

- Numeric values
- Booleans - `TRUE, FALSE, T, F`
- Column names with spaces
- Ok should not be enabled if a cell is empty.

If any of these is added to the grid just display a message box when the user is about to click OK.

N-thony:
@rdstern with the update in R code function by @shadrackkibet and some small bug fixes by doing the following test using the experimental survey dataset

- Pasted both `New name` and `Label` columns at the same time it works **OK**
- Renamed all cells in both `New name` and `Label` columns at the same time it works **OK**
- Renamed single cell in `New name` column it works **OK**
- Renamed single cell in `Label` column it works **OK**
I also added the additional check as suggested by Shadrack.

Note that the renamed value is considered in the R code only after leaving the cells, I have tried to fix that but seems to be a particular problem with Reogrid. I will keep looking into it. @rdstern could you test too? Thanks.

rdstern:
@N-thony this does seem to be causing you a lot of problems.
a) I tried with the survey data, pasting x1,... x7 from Excel into the names.  It works!.  But when I return to the dialogue the survey is still there with all the old names.  If I move data frames and then come back, then it updates.  
b) I added a data frame from DAAG called rockart.  It has 641 variables.  It takes over 20 seconds to load into the grid.  And you can't stop it.  There will be wider ones still.  So we do need an Add button, even if that is another issue.  And you should liaise with Patrick, who is looking at efficiencies here.
c) I then tried pasting a set longer than the number of names.  Good that worked fine.  It ignored me.
d) I then tried making non-unique names.  It gives an error, that is reasonable.  (Ideally this is captured earleir so we give our own message, but this is fine.)
e) then I tried a name with an error, so x<space>4.  It crashed.  These are obvious things to check for.  We must have routines in our file import that check this.
f) I tried labels and it seems to add them ok.  But then I tried to delete them and change them.  Once I have labels they don't ever seem to go away, and they don't change as easily as labels as they do in the grid.
You seem to be making quite "heavy weather" of all this.  
I was also hoping you might have made more progress on the rename with as well?

N-thony:
@rdstern I have fixed a couple of bugs found during our last chat, also from your review,

- especially heavy datasets like `DAAG rockart`, 
- deleting all column or single or multiple no neighborhood Label cells with Delete Keyboard
- Also fixed some bugs on testing.

lloyddewit:
@HawardKetoyoMsatsi Please could you review? thanks

rdstern (changes requested):
@N-thony this is much better now.  I first tried a very wide data set.  This was with over 7000 variables.  It took 16 seconds to load into the ordinary data selector, and 11 seconds to load into this grid.  Ideally both may be improved further, but we could live with this.  I also tried:
a) Pasting names - ok
b) Updating names - ok
c) Adding, updating and deleting labels - ok
d) Names repeated - sort of ok.  Told me it wasn't allowed as an error.  But then I could continue.
e) The one thing - for me so far that needs to be fixed is that it crashes if you input an illegal name, e.g. x<space>1.  We must already have a routine for this.

N-thony:
@rdstern I resolved item e) and also added the check for the French accent in the name for example `variety` is `variété` and it is an invalid column name. Please test it. Thanks.

rdstern (changes requested):
@N-thony I still managed to break it.  I added a single variable with an illgal name and it traps it.
Then I pasted
y.1
y_2
y-3
y4
y5
and then pressed Ok.  It crashed.

rdstern:
@N-thony I still get the same error, namely with:

image

I get:

image

Then I am thrown out of R-Instat

rdstern (dismissed):
@N-thony great it now seems to work fine.  I was also keen that you removed the message about individual names being wrong.  I am happy that you did that.
I hope that @lloyddewit or @shadrackkibet can check quickly, so it can be merged.  It is an important improvement, and I hope the rename_with (to be done later) will be a relatively small task

lloyddewit:
> @N-thony great it now seems to work fine. I was also keen that you removed the message about individual names being wrong. I am happy that you did that. I hope that @lloyddewit or @shadrackkibet can check quickly, so it can be merged. It is an important improvement, and I hope the rename_with (to be done later) will be a relatively small task

@HawardKetoyoMsatsi I think it makes sense for you to complete your peer review first, please let us know when it's done, thanks

HawardKetoyoMsatsi (commented):
@N-thony I find the codes to be fine and running fine. I have no issue  data_object_R6.R but am having a very small issue (ignore if possible) with dlgName.vb that I have left as a comment rather than a request for change.

N-thony:
@rdstern @shadrackkibet I made the change as we discussed few minutes ago. Please could you check/test? Thanks.

rdstern (dismissed):
@N-thony looks fine to me.  I am approving, so it can be merged.  The next item is the abbreviate function, which I hope will be easy to add?

shadrackkibet:
@N-thony `abbreviate` can be another radio button. Implement as follows;

`data_book$rename_column_in_data(data_name="survey", type="rename_with", .fn=abbreviate)`

If a control (Nud/input) is needed for `minlength`(R default =4) or any other parameter can be added as follows;

`data_book$rename_column_in_data(data_name="survey", type="rename_with", .fn=abbreviate, minlength = 4)`

rdstern:
@shadrackkibet great to have this and a really good "quick win" for R-Instat.  Once that is added and merged, then I suggest that the last feature in renaming (which overlaps with all your select stuff, i.e. your select is the find, and then we add replace!) becomes far less urgent, and could wait till later in the year.  Certainly until you have everything you need in the Select stuff.  So, I can write that as a new issue.  (Maybe we don't even need the find, but it just works on a defined select!

shadrackkibet:
> @shadrackkibet great to have this and a really good "quick win" for R-Instat. Once that is added and merged, then I suggest that the last feature in renaming (which overlaps with all your select stuff, i.e. your select is the find, and then we add replace!) becomes far less urgent, and could wait till later in the year. Certainly until you have everything you need in the Select stuff. So, I can write that as a new issue. (Maybe we don't even need the find, but it just works on a defined select!

This should be straightforward to implement. In a nutshell, this will simply have a receiver to take in a select object then we just rename those columns in the select object.

rdstern:
Exactly!

N-thony:
> @N-thony `abbreviate` can be another radio button. Implement as follows;
> 
> `data_book$rename_column_in_data(data_name="survey", type="rename_with", .fn=abbreviate)`
> 
> If a control (Nud/input) is needed for `minlength`(R default =4) or any other parameter can be added as follows;
> 
> `data_book$rename_column_in_data(data_name="survey", type="rename_with", .fn=abbreviate, minlength = 4)`

@rdstern @shadrackkibet I have added this.

instat/dlgName.vb
```diff
@@ -109,33 +127,335 @@ Public Class dlgName
...
+        clsNewColNameDataframeFunction = New RFunction
+        clsNewLabelDataframeFunction = New RFunction
+        clsDummyFunction = New RFunction
+        clsDummyRenameWithFunction = New RFunction
 
         ucrSelectVariables.Reset()
+        dctRowsNewNameChanged.Clear()
+        dctRowsNewLabelChanged.Clear()
+
+        ucrNudAbbreviate.SetText(4)
```
shadrackkibet, line 141: If you have the control passing the parameter then the control will read from the R code.

instat/dlgName.vb
```diff
@@ -16,14 +16,27 @@
...
-    Dim strSelectedColumn As String = ""
-    Dim strSelectedDataFrame As String = ""
+    Private bUseSelectedColumn As Boolean = False
+    Private strSelectedColumn As String = ""
+    Private strSelectedDataFrame As String = ""
+    Private strEmpty As String = " "
     Private clsDefaultRFunction As New RFunction
+    Private clsNewColNameDataframeFunction As New RFunction
+    Private clsNewLabelDataframeFunction As New RFunction
+    Private clsDummyFunction As New RFunction
```
shadrackkibet, line 32: Try using conditions. Then you wouldn't need a dummy function. Happy to discuss.

instat/dlgName.vb
```diff
@@ -109,33 +127,335 @@ Public Class dlgName
...
+        Return strValue
+    End Function
+
+    Private Sub ValidateNamesFromDictionary(iColIndex As Integer)
+        If iColIndex = 1 Then
+            For Each value In dctRowsNewNameChanged.Values
+                If Not CheckNames(value, iColIndex) Then
+                    MsgBox("The column name must not be a numeric or contains space or french accent or be a boolean e.g TRUE, FALSE, T, F.")
+                    bCurrentCell = False
+                    Exit For
```
shadrackkibet, line 228: We need to be moving this kind of code into user control. I think we have this kind of code in climatic data entry. I suggest you add a TODO comment.

instat/dlgName.vb
```diff
@@ -109,33 +127,335 @@ Public Class dlgName
...
+                End If
+            Else
+                clsDefaultRFunction.RemoveParameterByName("new_labels_df")
+            End If
+        End If
+    End Sub
+
+    Private Sub RemoveLabelsParams()
+        clsDefaultRFunction.RemoveParameterByPosition(8)
+        clsDefaultRFunction.RemoveParameterByPosition(9)
```
shadrackkibet, line 361: In my opinion, `RemoveParameterByPosition` is difficult to maintain by developers. If a developer adds another parameter then the order isn't always "correct". I would prefer `RemoveParameterByName()`. Does this make sense to you?

shadrackkibet (changes requested):

rdstern (changes requested):
@N-thony I think the abbreviate works fine now.  But your function has now (behind the scenes) changed the janitor function for clean names.  I think it also applies the default width from the abbreviate function.  
At least the result of the clean names, with snake works fine in version 0.7.4 and is now different in your current function.  I was trying with the iris set.

rdstern (dismissed):
Great - this is now an excellent new feature.  Hope @shadrackkibet or @lloyddewit can approve, so it can be merged

instat/dlgName.vb
```diff
@@ -213,7 +551,7 @@ Public Class dlgName
         End If
     End Sub
 
-    Private Sub ucrCoreControls_ControlContentsChanged() Handles ucrInputNewName.ControlContentsChanged, ucrReceiverName.ControlContentsChanged, ucrSelectVariables.ControlContentsChanged, ucrPnlOptions.ControlContentsChanged
+    Private Sub ucrCoreControls_ControlContentsChanged() Handles ucrInputNewName.ControlContentsChanged, ucrReceiverName.ControlContentsChanged, ucrReceiverColumns.ControlContentsChanged, ucrSelectVariables.ControlContentsChanged, ucrPnlOptions.ControlContentsChanged
```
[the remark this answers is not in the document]
lloyddewit: @HawardKetoyoMsatsi thanks for the comment, I agree that this may not be strictly necessary but is safe enough

lloyddewit:
@N-thony I think there are 3 open comments from @shadrackkibet that are still open. Please could you fix these (without setting to resolved) and then ask @shadrackkibet to check. If he's happy, then he can resolve these remaining comments and approve.

@shadrackkibet thanks for reviewing this PR, I don't think I need to review also, but if you'd like me to review, then let me know, thanks

N-thony:
> @N-thony I think there are 3 open comments from @shadrackkibet that are still open. Please could you fix these (without setting to resolved) and then ask @shadrackkibet to check. If he's happy, then he can resolve these remaining comments and approve.
> 
> @shadrackkibet thanks for reviewing this PR, I don't think I need to review also, but if you'd like me to review, then let me know, thanks

@lloyddewit I think the three comments were already solved.

lloyddewit:
> 
> @lloyddewit I think the three comments were already solved.

@N-thony I looked at the code and couldn't see where the comments are solved. Maybe I'm missing something. For each comment, please could you explain where/how the comment is solved?
thanks

instat/dlgName.vb
N-thony: this is resolved

lloyddewit:
@N-thony thank you for the update
@shadrackkibet if you think that all your comments have been resolved, then please could you approve?
Thanks

rdstern (approved):

PR #7216, opened 2022-02-10
Added an option to detect and report on empty columns into delete rows/columns dialog

### LINKED ISSUE #6883: Detecting and possibly deleting empty variables
[dfid2_ug_210803.zip](https://github.com/africanmathsinitiative/R-Instat/files/7431804/dfid2_ug_210803.zip)
This is a stata file that reads easily into R-Instat.
It has 268 variables and about 47000 rows.
Some numeric variables are simply NA (all 47000 rows).  
Some character variables are just blank.  

I would like to tidy these data and one step could be to report on and then possibly to delete these variables.  How can I do this in R and hence what do we need in R-Instat for this?

That's the main question.  However, another issue is that it is a good example where the variable names are very long and some are more suited to being variable labels.

And the Data selector does need improvements!  One that I would have appreciated here, and haven't mentioned elsewhere is that it could remember where it is in the list of variables, when I return.  Currently it always starts at the top.

anastasia-mbithe:
This fixes #6883 
Hello @rdstern and @shadrackkibet, I added the empty button at the top of the dialog, as well as the two checkboxes as suggested. I also set the functions and the dialog is working properly; I tested with a procurement dataset that had a lot of empty variables. Am working on adding a report in the output window, i.e.. xxx empty variables deleted or rows or both as suggested in the issue.


shadrackkibet:
@rdstern should we remove empty columns in the original data frame or we should have it as a separate data frame?

anastasia-mbithe:
Hello @shadrackkibet, I have been researching on something here and still don't know how to implement it.
I want to add a message in the output window after the execution of the remove empty columns /rows dialog. Such that; if the function deletes two empty variables it gives a message in the output window i.e. "2 empty variables deleted" if its 55 then ; " 55 empty variables deleted", if its rows; "55 empty rows deleted" etc
Can you assist me in that?

shadrackkibet:
Ann, leave this to me for now. I will alert you when to proceed. I think I understand what you need to complete the task from your explanation. Keep up the good work!

Meanwhile, I suggest you move to another task.

shadrackkibet:
@anastasia-mbithe I have now made changes to this PR. You will notice I have added a `remove_empty` method into the `databook`. There are two main reasons for this change.
-  The method in the `databook` allows us to append the resulting data frame into the original data frame because we can work with the data frame "internally" - This means we do not need `ucrSave` which allows us to save the resulting object separately.
- It allows us to customize the message we need using `cat` which we can then capture and print into the output window - loosely speaking.

You can now pull the changes and test them from your end. Feel free to ask any questions or make changes at the dialog level, I didn't look into it much. Let me know if you find any bugs.

anastasia-mbithe:
@shadrackkibet, thank you so much for this, am following the code as well as testing the dialog. 
Am wondering what the default state of the checkboxes should be: should they be checked or unchecked or one of them to be checked. I realized after clicking reset, both appear checked, should I change that? and if so to which preferred state?

shadrackkibet:
The default for the R function is both checked so I will go with that. @rdstern might have a different opinion.

anastasia-mbithe:
@rdstern you can look at this and also see if you are okay with the checkboxes.

rdstern:
@anastasia-mbithe I can not check yet.  When I load your branch it gives an error right at the start, namely:
image

Then it doesn't let me load a data file.  Keeps giving that error.  Please tell me when I can check.  The dialogue looks fine.

anastasia-mbithe:
@rdstern The bug is fixed, check it now.

shadrackkibet:
@rdstern please check. I would like to close this one as soon as possible.

PR #7218, opened 2022-02-10
Added 2 new geoms from the moderndive package

### LINKED ISSUE #7206: Add 2 new geoms and the moderndive package
The moderndive package includes the geoms geom_parallel_slopes and geom_categorical_model.  @HawardKetoyoMsatsi could you please add them to the ggplot geoms and also request this package in the next version of R-Instat.  For now they don't need to be added to a dialogue.  I suggest later that categorical will be added to Model > Two Variable and possibly 3 variable (for facets?). The parallel geom will be added to 3-variable

HawardKetoyoMsatsi:
Fixes #7206
@rdstern and @africanmathsinitiative/developers 
This is ready for review.

In this Commit I added two geoms, (geom_categorical_model and geom_parallel_slopes) from moderndive package.

I also changed the location of geom_mosaic in the list of geoms so that geom_mosaic can follow the alphabetic order in the list of all geoms.

I however realised that some of the layer parameters do not display input boxes correctly (see "Alpha","Stroke" and "Fill" in the attached image). This problem is not caused by the changes I made. 
image

I could also not add layers using Describe > Use Graph > Options > Layers > Add.  The code breaks. This problem is also not caused by the changes I made.


rdstern (approved):
@HawardKetoyoMsatsi this seems to work fine.
a) You found some problems with the controls, and I noticed that too.  That's, as you say, a separate issue.  Can you please write it as an issue.
b) I report how I tested, in case others want to do something similar. I use3d Describe > General > Graphics dialogue, with the survey data.  For the parallel I used yield as the y and fert as the x, with variety as the factor.  (I had to install the package).  It worked fine.  I then added facets and this was ok.
c) Then I used the same y, and just variety as the x, for categorical.  That was also ok.  I then tried fert (numeric) as the x.  It didn't give an error, but the graph was blank.  Then I made fert into a factor and it worked fine.

So ready for a code check by @shadrackkibet or @lloyddewit and merging.

Oh @haward remember to include the package in the pull request.

instat/ucrGeom.vb
```diff
@@ -511,6 +513,27 @@ Public Class ucrGeom
...
+        clsgeom_categorical_model.AddAesParameter("x", bIsMandatory:=True)
+        clsgeom_categorical_model.AddAesParameter("y", bIsMandatory:=True)
+
+        'Optional Aesthetics 
+        clsgeom_categorical_model.AddAesParameter("colour")
+        clsgeom_categorical_model.AddAesParameter("size")
+
+        'Add  layer parameters
+        clsgeom_categorical_model.AddLayerParameter("position", "list", Chr(34) & "identity" & Chr(34), lstParameterStrings:={Chr(34) & "identity" & Chr(34), Chr(34) & "jitter" & Chr(34)})
+        clsgeom_categorical_model.AddLayerParameter("se ", "list", "TRUE", lstParameterStrings:={"TRUE", "FALSE"}) 'We add space after parameter name se ("se ") for correct labeling, if the space is omitted, vb labels "se" as "ifelse" but when the space is included "se " is correctly labelled as "se".
```
lloyddewit, line 528: Thanks for adding the comment about "se ", very helpful.

Normally, rows shouldn't go over 100 columns

lloyddewit (approved):
@HawardKetoyoMsatsi Thanks for the tidy implementation.
I only had one comment, but it's too small to prevent merging

PR #7219, opened 2022-02-10
Ensure output window always scrolls to bottom when new output is added

### LINKED ISSUE #6996: Output Window doesn't always scroll to bottom when new output is added
If you click somewhere at the top of the output window so that the cursor is within one of the output blocks.
Then go to a dialog that produces output. It doesn't scroll to the end of the window. I think it just scrolls to the end of the block that it is in.
So it works if the cursor wasn't in the output window before running the dialog. But not if you were somewhere within the output.

Patowhiz:
Fixes #6996. 

### DIFF
--- a/instat/UserControl/ucrOutputPage.vb
+++ b/instat/UserControl/ucrOutputPage.vb
@@ -166,7 +166,6 @@ Public Class ucrOutputPage
         }
         pnlMain.Controls.Add(panel)
         pnlMain.Controls.SetChildIndex(panel, 0)
-        panel.Focus()
         AddCheckBoxToElementPanel(panel, outputElement)
         AddHandler panel.Resize, AddressOf Panel_Resize
         Return panel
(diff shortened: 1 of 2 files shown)


Patowhiz:
@lloyddewit I noticed I can't change this to draft. I will close it once @dannyparsons has had a look at it.

Patowhiz:
@dannyparsons this now fixes #6996. Took me a lot of time but it should help with html output task.
@rdstern can also test it. Thanks.

rdstern (approved):
Seems fine to me, but I am only using Windows

lloyddewit (approved):

lloyddewit:
> Seems fine to me, but I am only using Windows

Issue #6996 didn't say that this issue was related to Linux so I assume it's OK to merge this PR.

Patowhiz:
@dannyparsons @lloyddewit this would be very interesting to me if it doesn't work in **Linux**, that would mean we need to be very careful with how `.Net` events are dispatched in a **Linux** platform. I'm also just using Windows.

PR #7220, opened 2022-02-10
Extended Prepare>Data Reshape>Transpose and ensured consistent with Tidyverse

### LINKED ISSUE #6905: Change "Transpose" to extend and be consistent with Tidyverse
I have been looking at the dialogs in the "Data Reshape" sub-menu to check for consistencies with Tidyverse.

Reading [here](https://newbedev.com/how-to-transpose-t-in-the-tidyverse-using-tidyr) shows that transposing with the Tidyverse is a little less simple than the current `t()` command. For example:

```
library(dplyr)

# Currently how we transpose data in R-Instat
as.data.frame(x=t(chickwts$weight))

# How we can transpose using tidyverse code
chickwts %>%
  dplyr::select(c(weight)) %>%
  tibble::rownames_to_column() %>%  
  pivot_longer(-rowname) %>%
  pivot_wider(names_from=rowname, values_from=value)
```

Additionally, transposing using the Tidyverse introduces new problems - such as an error if there is a factor and a numerical variable that are both being transposed.

@rdstern I suggest we keep this as it is for now using the `t` function. What do you think?

EstherNjeriLiberatta:
Fixes #6905 (Prepare>Data Reshape>Transpose)

In this PR, I have separated the get data line of code from the transpose function which now appears as;
`columns <- data_book$get_columns_from_data(data_name="survey", col_names="village", force_as_data_frame=TRUE)`.
I have replaced the t function with transpose function from the data.table package i.e. `data.table::transpose()`.
I have added the `keep.names`  argument in the function as an option when the checkbox is checked and the `make.names` argument in when added to the single receiver.

@lilyclements, @shadrackkibet I have a challenge with the `make.names `argument.  I have discussed this with @Ivanluv  but it still gives an error when I add variables to the receiver. Please could you help with that? Thank you.


EstherNjeriLiberatta:
@shadrackkibet the` make.names `argument in R runs as: 
`crisps_transposed<-transpose(crisps, make.names = "Sample")`  but the code in R-instat runs as 
`survey_transposed <- as.data.frame(x=data.table::transpose(l=columns, make.names=size))`.
The output transforms the first column from:
`crisps_transposed<-transpose(crisps)` to
image

 with the `keep.names` argument:
image

with the `keep.names` and `make.names `argument:
crisps_transposed<-transpose(crisps, keep.names = "Sample", make.names = "Sample")

image

I get this error when I run the code.
image


I hope this helps. Thank you

rdstern:
@EstherNjeriLiberatta your explanation above would be very helpful, except you use data called `crisps`, as your example.  Where is it?  I wanted to repeat what you did, but can't yet easily.

rdstern:
@EstherNjeriLiberatta as you say, it runs fine as long as you don't try to add the new names:
a) Consider changing Variable Name (Optional): to New Names (Optional):
b) Change Display Variable Names to `Include Old Names` and possibly give oldnames as the default name.

Then I also tried, and failed, with the offending feature.  I used the Excel file from the review and transposed the 9 variables.  I tried to make the variable, called Feature, into the new names.

It complained that this variable was not of length 1, and indeed it was of length 45, (number of rows) with the way you accessed it.  But I couldn't find the solution.  I suggest it should be very quick, for @shadrackkibet and useful to report once you know.  If not, then possibly try @lilyclements later.

EstherNjeriLiberatta:
Thank you @rdstern.

EstherNjeriLiberatta:
@rdstern I have fixed items 1 and 2. 
@shadrackkibet please assist me with fixing the error that is produced when a variable is added to `make. names` argument i.e. the `New names (Optional)`control. Thank you

lilyclements:
@EstherNjeriLiberatta at the moment you run `make.names = size`. We want it to be `make.names = "size"` (i.e., with quotes around the variable). Does that fix the problem?

rdstern:
Aha, Lily's suggest is different in another way.  The make.names variable is also included in the columns, and that also saves the extra line, where you got that variable separately.  I tried with the quotes as @lilyclements suggested, but didn't take the extra step of including that variable in the columns!

instat/dlgTransposeColumns.vb
```diff
@@ -28,56 +28,77 @@ Public Class dlgTransposeColumns
...
 
-        ' ucrReceiver
-        ucrReceiverColumsToTranspose.SetParameter(New RParameter("x", 0))
         ucrReceiverColumsToTranspose.SetParameterIsRFunction()
         ucrReceiverColumsToTranspose.Selector = ucrSelectorTransposeColumns
         ucrReceiverColumsToTranspose.SetMeAsReceiver()
-        ucrReceiverColumsToTranspose.bForceAsDataFrame = True
+
+        ucrReceiverVariableNames.SetParameter(New RParameter("make.names", 1))
+        ucrReceiverVariableNames.SetParameterIsRFunction()
```
lilyclements, line 44: By changing this line, the `size <- ...$get_columns_from_data(...)` line should be removed, and instead `size` will be read in with quotes.

instat/dlgTransposeColumns.vb
```diff
@@ -28,56 +28,77 @@ Public Class dlgTransposeColumns
...
         bReset = False
         autoTranslate(Me)
     End Sub
 
     Private Sub InitialiseDialog()
         ucrBase.iHelpTopicID = 277
 
-        ' ucrReceiver
-        ucrReceiverColumsToTranspose.SetParameter(New RParameter("x", 0))
         ucrReceiverColumsToTranspose.SetParameterIsRFunction()
```
lilyclements, line 39: While I'm here... can you fix this typo - should say "columns" not "colums" in the name

instat/dlgTransposeColumns.vb
EstherNjeriLiberatta: @lilyclements Replacing the `.SetParameterIsRFunction()` with `.SetParameterIsString` has helped adding the quotes which is great. However, I am getting a new error now.

image

Here is the code produced:
```
columns <- data_book$get_columns_from_data(data_name="survey", col_names="village", force_as_data_frame=TRUE)
survey_transposed <- as.data.frame(x=data.table::transpose(l=columns, make.names="field"))
data_book$import_data(data_tables=list(survey_transposed=survey_transposed))
```

lilyclements: So the value you put into the make.names argument has to be one of the columns in the multiple receiver. It wouldn’t make sense to set the names as field if the field column isn’t being transposed.

EstherNjeriLiberatta: Okay, thanks Lily. I have tried that and I get a another error.

rdstern:
@EstherNjeriLiberatta it is good that you are proceeding.  However, when I look at the first message you recently sent to @lilyclements 
image

Now ask yourself if, in future, you could have found this error yourself and tried a correction?  It says above "make.names field not found in names of input".  

I write because your latest message seems to say to me that "I have tried that and I get a another error."

Ask yourself if any help is possible to that message, when you give  - at least to me - no further details of the error.  So can you see if you can solve this yourself, from the message - if any that is given, or from something different you have typed.  If not, then you do need help, and should give all the details you can of what the did, what it resulted, in R-code, and perhaps details of the data and then - of course, the error message.

Are you now in Maseno?  Shadrack is now there and so is Antoine.  So they could perhaps also help.

EstherNjeriLiberatta:
@rdstern thank you. I'll do that.

lloyddewit:
@EstherNjeriLiberatta Thank you for the change. Do you still need to make changes, or have you now fully tested this PR yourself, and is it ready for peer review and @rdstern 's test? Please clarify the status, thanks

rdstern (changes requested):
@EstherNjeriLiberatta great, this is working now - well almost.
It now works (with the new names) when you include the new names variable in the multiple receiver - as well as the single receiver that is below.
That's good.  Now please get some help on the coding that is needed, when you include the names variable in the single receiver, but not in the multiple.  Your VB code should still add it to the variables in the R command that is generated.

rdstern:
@EstherNjeriLiberatta any progress on this?  Should it be handed to someone else?

shadrackkibet:
@Ivanluv happy to take over this task?

Ivanluv:
Yes, am taking over

shadrackkibet:
Thanks

lilyclements:
@shadrackkibet is there a way that we can restrict the variables that can be entered into the single receiver as only the variables in the multiple receiver?

lilyclements:
Or to automatically read the variable in the single receiver into the `l` parameter in the `transpose` function.

ie,

`transpose(l = c(columns, <var in single receiver>`)

or, to at least catch the error. So if they press OK and the variables don’t match to have a warning come up to say to add the `make.names` variable to the multiple receiver.

rdstern:
@lilyclements but does the user of the dialogue have to worry about that?  Couldn't it be just added through the (preumably VB?) code, before the R function is called?

shadrackkibet:
> @shadrackkibet is there a way that we can restrict the variables that can be entered into the single receiver as only the variables in the multiple receiver?

@lilyclements we do not have anything like this implemented yet. Three options to approach this;

- Implement this at dialog level if this is something that will not be useful elsewhere 
- Otherwise, implement at the receiver level if there are other use cases apart from this dialog. - Something similar to `   ucrReceiver.SetDataType()`
> Or to automatically read the variable in the single receiver into the l parameter in the transpose function.
ie,
transpose(l = c(columns, <var in single receiver>), …)
or, to at least catch the error. So if they press OK and the variables don’t match to have a warning come up to say to add the make.names variable to the multiple receiver.

- We can have an R wrapper to add variables at the `R` level. 
- This is also possible at dialog level.

As a way forward I suggest as follows;

Since @Ivanluv is now taking over this task - try out and report here.

EstherNjeriLiberatta:
> @lilyclements we do not have anything like this implemented yet. Three options to approach this;
> 
> * Implement this at dialog level if this is something that will not be useful elsewhere
> * Otherwise, implement at the receiver level if there are other use cases apart from this dialog. - Something similar to `   ucrReceiver.SetDataType()`
> 
> > Or to automatically read the variable in the single receiver into the l parameter in the transpose function.
> > ie,
> > transpose(l = c(columns, ), …)
> > or, to at least catch the error. So if they press OK and the variables don’t match to have a warning come up to say to add the make.names variable to the multiple receiver.
> 
> * We can have an R wrapper to add variables at the `R` level.
> * This is also possible at dialog level.
> 
> As a way forward I suggest as follows;
> 
> Since @Ivanluv is now taking over this task - try out and report here.

@lilyclements @rdstern , @Ivanluv and I have added this on the dialogue level this now works as you were suggesting. I have added a screenshot below of how its appearing on the dialogue and the script window. 
image
image
 
However, this error pops up when I try to run the code. We are still trying to find out what it means @shadrackkibet would you kindly help us out on this. Thank you.
image

rdstern:
@EstherNjeriLiberatta I hope you understand why there is an error here, and it is good that you try it.  
There are 2 separate points before this dialogue is completed.   I thought there was just one, but you have usefully found a second, and important one.  I think this may also indicate that you are not quite sure what the command is doing, but that is also fine, because users will often be in the same situation, and it shouldn't crash.
a) First my original request and that is that you shouldn't have to repeat the Features variable in the example below, also in the multiple receiver.  (The data I used are below.)
Here is sort of the command generated, which I put into the script file>
```
# Code generated by the dialogue. Transpose Columns

columns <- data_book$get_columns_from_data(data_name="Scores45", col_names=c("BlueSky","Deducer","JASP","jamovi","RAF","Rattle","Rcmdr","R_Instat","RKWard","Feature","Feature"), force_as_data_frame=TRUE)

Scores45_transposed <- as.data.frame(x=data.table::transpose(l=columns, make.names="Feature", keep.names="old name"))
data_book$import_data(data_tables=list(Scores45_transposed=Scores45_transposed))

rm(list=c("Scores45_transposed", "columns"))  
```
The first command you generate is the col_names line above.  I suggest if the make.names option is used, then you simply add that variable into the col_names command, see above.  Now, if it isn't there before, then it is added, which is what you want.  If it was there before then it appears twice, as above, and that seems to cause no problem.  Oh no it does cause a problem.  If it is there twice, then it makes it the names and also adds it as another row

So to solve my first problem you have to add the variable if it isn't there, and not again, if it is.  Try it perhaps with these data from our review:
[R GUI Comparisons.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/8409454/R.GUI.Comparisons.xlsx)

b) Now your problem.  Now I can't solve that, so it is good for @shadrackkibet or someone else to help on this.  But trying it above - when it works - shows why you have the problem.  
I don't get a problem with the code you generated.  But I assume you then made the village into the New Names field, and that gives the problem you show.  That's possibly because the Village doesn't have unique names for the successive columns.  There are just 4 village names and 36 columns you are producing!  (If you make field into a factor and then use Prepare > Factor > Combine, you can produce a unique set of names and it works fine.  So, the problem to be solved is what do you want to do, when the names variable isn't unique.  
So one way to solve the problem is 
a) Check first if the names variable is unique.  (That might be to see whether the length of the variable is the same as the number of unique values.  
b) If not, then there is a make.unique function in R base which appends a numeric value.  Then you submit that new variable to the transpose function.
I am sure @shadrackkibet could help, but check also first with @HawardKetoyoMsatsi if he is still in Maseno?  This should be quite easy.  Your problem needs to be solved.  If you can't solve a) above we may merge and the make it into another issue to be solved later.

EstherNjeriLiberatta:
@rdstern Thank you sir, I'll look at it.

shadrackkibet:
@EstherNjeriLiberatta did you get help on this? Happy to look into it.

EstherNjeriLiberatta:
@shadrackkibet Not yet, it's okay, thank you.

rdstern:
@EstherNjeriLiberatta could this be completed now you have @lilyclements to discuss it with?

EstherNjeriLiberatta:
@rdstern Yes.

lilyclements:
After talking with @ChrisMarsh82, the problem that occurs with the `make.names` argument occurs when the variable in the `make.names` does not have unique values. This then creates columns with the same name, and so VB has an error and it crashes.
A solution for this is:
* Have an if statement in VB for when the `make.names` receiver is not empty, and if `(length(unique(variable_in_receiver)) != nrow(data_frame))`
* If that is the case, highlight the receiver red and have a tooltip stating something like "Variable does not contain unique values. OK will be disabled."
* OK can then be re-enabled when the variable is removed (or when a new variable with unique values is read in).
* You can look at the make keys dialog to see about changing a colour.
* This solution requires VB to read R code to run the `(length(unique(variable_in_receiver)) != nrow(data_frame))` line.

lloyddewit:
> This solution requires VB to read R code to run the `(length(unique(variable_in_receiver)) != nrow(data_frame))` line.

Does this check need to be done in R? Would it be more readable to do the check in VB?

lilyclements:
@lloyddewit yes, but after speaking with @ChrisMarsh82 we felt this would be more efficient.

However, @Patowhiz fixed this issue in the `import` function last week.

@EstherNjeriLiberatta can you update your branch from the master? It is working for me well now if I update from the master - hopefully this bug is no longer occurring for you too!

EstherNjeriLiberatta:
@lilyclements  I've updated my branch it's working okay now, thank you. @Patowhiz  thanks for fixing this.

rdstern (changes requested):
@EstherNjeriLiberatta the main problem is indeed fixed.  Excellent - this was a name column that provided non-unique names.
I can also accept that you restrict the allowable variables so the named one must be also selected.  But I then need to be able to return to the multiple receiver and add that variable.
This is a cosmetic change, so I hope this dialogue can be included in the current upgrade.

EstherNjeriLiberatta:
@rdstern you can now be able to return to the multiple receiver and add a variable. This is ready for review.

rdstern (approved):
Another one it would be good to include in the release if the code is ok?

lloyddewit (changes requested):

lloyddewit (approved):
@EstherNjeriLiberatta Thank you, this looks good.
I had some small comments. I didn't know how urgently @rdstern wanted this PR merged so I resolved the comments myself, I hope you don't mind.

PR #7226, opened 2022-02-11
Added options to select columns by variable type, labelled variables, empty variables and NA variables

### LINKED ISSUE #7190: More options on choosing select object
We can now define selections of a set of variables, just as we can define filters of rows.  Each select is then given a name.  I am now terming "a select" as any selection.
They are going to have a variety of uses, including "taming" some of the wide data sets.

1. Please could we add their choice by type of data.  So we could easily select all numeric, all character, all factor, all date, all logical, etc.
2. Could we please include all with value labels as a select too.
3. Does our Not already work.  So could we have all factors, except those I have just given in the data receiver?
4. With factors could I please be able to select all factors with x levels.
5. While there, could there be other possible rules, e.g. all variables with missing values, etc.

As part of this, could we also consider the properties of a select.  Could selects also have a type, so if all are character, then this is also character.  And they can be mixed, or there could be a list if they are more than one type.  If they are all factors with 2 levels, could the number of levels also be included?

shadrackkibet:
fixes #7190

@rdstern  this PR add more options to select columns. I have enabled the following options under select.
image
The remaining bit is an option to select by the number of factor levels.

I used the following dataset to test it.
[dfid2_ug_210803.zip](https://github.com/africanmathsinitiative/R-Instat/files/8051556/dfid2_ug_210803.zip)

### DIFF
--- a/instat/dlgSelect.vb
+++ b/instat/dlgSelect.vb
@@ -35,6 +35,8 @@ Public Class dlgSelect
 
     Private Sub InitialiseDialog()
         ucrBase.iHelpTopicID = 597
+        ucrBase.clsRsyntax.iCallType = 2
+
         ucrInputSelectPreview.txtInput.ReadOnly = True
         ucrReceiverSelect.SetItemType("column_selection")
         ucrReceiverSelect.strSelectorHeading = "Column selections"
--- a/instat/dlgSelectColumns.Designer.vb
+++ b/instat/dlgSelectColumns.Designer.vb
@@ -34,6 +34,7 @@ Partial Class dlgSelectColumns
         Me.lblTo = New System.Windows.Forms.Label()
         Me.lblFrom = New System.Windows.Forms.Label()
         Me.lblColumnType = New System.Windows.Forms.Label()
+        Me.ucrChkNot = New instat.ucrCheck()
         Me.ucrInputColumnType = New instat.ucrInputComboBox()
         Me.ucrNudTo = New instat.ucrNud()
         Me.ucrNudFrom = New instat.ucrNud()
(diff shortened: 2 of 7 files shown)


rdstern (changes requested):
@shadrackkibet I think this may be unrelated to your current improvements, but it is something that needs to be fixed - if I am right.
I load 2 files into R-Instat.  I used efc from sjmisc and lateformeeting from rrefine, but I think that any two will do.
I then went into a dialogue - I used Describe > One Variable > Summarise - but I assume any will do.  Then I did a filter on efc.
I then went back to the dialogue and changed to the other data frame.  I now used Data Options again and your select.  It had the correct data frame. I made a new selection, and it still assumes that the data frame is the original, i.e. efc.  It can't be changed there, so I am stuck.

shadrackkibet:
@rdstern thanks for reporting this. I will fix it in this PR.

rdstern (changes requested):
@shadrackkibet this is looking great now.  I couldn't find any problem with multiple data frames.  And the Not works well.
I also tried from right-click, from the Prepare menu and from the data options.
All good!
The one problem was the labels.  I tried variable labels  and it closed all the data frames.  I would like both variable labels and value labels if possible, but if they are giving you problems, then they could be left for now.
If you do include them, then I think that all data frames have variable labels and asking for those (or Not) is whether they are blank.  With value labels, I am not sure that all data frames have them.  Once they do, I think that not means contains NA.

There will also be the options on Apply, but let's merge this first.  It is looking excellent.  (I really like the Not!)

shadrackkibet:
@rdstern Variable Label should now work.

Let us consider value labels on a separate PR

rdstern (changes requested):
@shadrack - damn!
That works nicely now - the variable labels.  I wonder why you call it where and not type?
Anyway I was about to approve, when I thought I should try and see what happens when I select something that isn't there.
So I selected all the factors in a data frame where there were none.
It gave me no rows or columns, but still the column names from the last select.  Then it crashed!  I wonder what we should do here?  Perhaps a message and also maybe something in the output window and then don't allow it?  Maybe just show the first variable, or make it the same as remove selection?

shadrackkibet:
@rdstern I choose not to allow it. You will get a message as an error -  This should clearly be a message but at the moment we are limited in how we capture error messages.

rdstern (dismissed):
Great.
Interesting that we can't capture and give a clearer message.  I wonder if that will be possible in the future?

shadrackkibet:
@rdstern I have a workaround in R. Instead of an error message we print something in the output window. Perhaps you might like it?

rdstern (approved):
@shadrackkibet I am still approving.
However, there is a new feature (or maybe it was there all the time?)  I chose last column, and then (in Columns) selected the first 3 columns.
They appeared fine, but in a new order.  The last column was first.  I kind-of thought they would remain in their original order.  I now realise I didn't check what now happens, when the selection is stopped.Do they return to their original order, or is the change forever?

lloyddewit (approved):

PR #7217, opened 2022-02-10
Fixes Bug in Regular Sequence

derekagorhom:
fixes issue #7199 

fixed the bug in `Regular sequence` by changing the code in `line 182` 
to 
`ucrNudRepeatValues.AddAdditionalCodeParameterPair(clsSeqDateFunction, New RParameter("each", 1), iAdditionalPairNo:=1)`

@africanmathsinitiative/developers  can anyone review this.

PR #7231, opened 2022-02-14
Fixes the Regular Sequence dialog bug, refactors and improves its code

### LINKED ISSUE #7199: Fix error or errors in Prepare > Numeric > Regular Sequence
I first found this nasty bug with an odd data set, but I can now generate it quite easily.  I am suggesting @Patowhiz to look reasonably generally at the code.
This is straight after starting R-Instat.  I open the standard survey data and use the Regular Sequence dialogue.  I put 40 as shown below, so the sequence needs to be truncated.  

image

Damn - it works.  On other occasions I have had an error that the resulting column must be divisible by the worksheet length. 

No matter.  I return to the dialogue and repeat - nothing changed, except R-Instat has made the name regular1.  Now I get this nasty error:

image

I press Continue.  Nothing has been generated into the output window.

Now when I return to the dialogue it says there is an invalid character in the name.
image

I have experimented a bit more.  If I use the dialogue in an ordinary way, but put my own invalid names, then it traps this - which is good.  But I then tried with 36 replaced by 15, so the sequence has to be extended - and got the nasty error again.

Tried again with multiple columns generated of right length.  All ok.  Then made 36 into 18 - so still a multiple, and got the nasty error immediately.

Patowhiz:
Fixes #7199
Replaces PR #7217

Fixes the original bug explained in the issue and also the other issues raised in comments.

@derekagorhom this PR includes you changes in PR #7217. I decided to create a new PR because it will be easier for me to answer any questions raised about the changes made to other files.

@rdstern @derekagorhom I looked into using the 'seq' function for both numeric and date options but I didn't do it in this PR, the changes made here surpassed my time estimates for such implementation. Perhaps @derekagorhom can have a look at it. I'm still not sure on how it can be correctly done and I could just create a new issue about it. Thanks.


rdstern (dismissed):
@Patowhiz now looks good to me

instat/ucrDataFrameLength.vb
```diff
@@ -15,24 +15,34 @@
...
-        SetLength()
-    End Sub
+    ''' <summary>
+    ''' Gets the length(no. of rows) of the data frame, if data frame is not set, returns 0
+    ''' </summary>
+    ''' <returns>length of the set data frame</returns>
+    Public Function GetDataFrameLength() As Integer
+        Dim iLength As Integer
+        Integer.TryParse(txtInput.Text, iLength)
+        Return iLength
```
lloyddewit, line 40: If `Integer.TryParse()` can parse the text into an Integer, then it returns True, else False.
Should we check the result of the parse?
If it fails, then should we raise a developer error or return a default value like zero?
```suggestion
        Return If(Integer.TryParse(txtInput.Text, iLength), iLength, 0)
```

instat/dlgRegularSequence.vb
```diff
@@ -20,12 +20,17 @@ Imports RDotNet
...
     Private bReset As Boolean = True
-    Private clsSeqFunction, clsRepFunction As New RFunction
-    Private clsSeqDateFunction As New RFunction
-    Private bUpdateBy As Boolean = False
-    Private clsByDateOperator As New ROperator
-    Private clsDefaultStartDate As RFunction
-    Private clsDefaultEndDate As RFunction
+    Private clsRepFunction, clsSeqFunction, clsSeqDateFunction As New RFunction
+    Private clsByOperator As New ROperator
+    Private strDefaultSequenceOption As String = ""
```
lloyddewit, line 25: Should this be a string? 

Or should it be an enumeration? Or a boolean?

instat/dlgRegularSequence.vb
```diff
@@ -63,170 +88,135 @@ Public Class dlgRegularSequence
...
-        SetDefaultRdo()
+        SetRCodeForControls(bReset)
+        SetSelectedDefaultSequenceOption()
         TestOKEnabled()
+        bReset = False
     End Sub
 
     Private Sub TestOKEnabled()
-        If ucrNewColumnName.IsComplete Then
+        If ucrSelectDataFrameRegularSequence.cboAvailableDataFrames.Text <> "" AndAlso
```
lloyddewit, line 212: safer?
```suggestion
        If Not String.IsNullOrEmpty(ucrSelectDataFrameRegularSequence.cboAvailableDataFrames.Text) AndAlso
```

instat/dlgRegularSequence.vb
```diff
@@ -235,157 +225,94 @@ Public Class dlgRegularSequence
...
-                Else
-                    clsSeqFunction.AddParameter("by", "-" & ucrInputInStepsOf.GetText())
-                End If
-            Else
-                clsSeqFunction.RemoveParameterByName("by")
+            'if the new peview script is the same as the previous one then no need to refresh preview
+            'this check is useful for large data frames.
+            strNewPreviewScript = clsAsCharacter.ToScript()
+            If strNewPreviewScript = strPreviewScript Then
+                Return
```
lloyddewit, line 253: better for a sub?
```suggestion
                Exit Sub
```

instat/dlgRegularSequence.vb
```diff
@@ -235,157 +225,94 @@ Public Class dlgRegularSequence
...
     End Sub
 
-    Private Sub ucrInputDateControls_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrDateTimePickerFrom.ControlValueChanged, ucrDateTimePickerTo.ControlValueChanged, ucrInputComboDatesBy.ControlValueChanged
-        If bUpdateBy Then
-            SetInStepsOfDateParameter()
+    Private Sub ucrInputStepsOfControls_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputFrom.ControlValueChanged, ucrInputTo.ControlValueChanged, ucrDateTimePickerFrom.ControlValueChanged, ucrDateTimePickerTo.ControlValueChanged, ucrInputComboDatesBy.ControlValueChanged, ucrPnlSequenceType.ControlValueChanged
+        'ucrInputInStepsOf will be empty when dialog loads for the first time
+        'no need to st value for ucrInputInStepsOf if its empty
+        If ucrInputInStepsOf.IsEmpty Then
+            Return
```
lloyddewit, line 290:
```suggestion
            Exit Sub
```

lloyddewit (changes requested):
Looks good, just a few small comments

instat/dlgRegularSequence.vb
Patowhiz: I used a string because it seemed simple and the option is never exposed externally to other classes. It's entirely used internally. I could also use a `boolean` if you strongly think it's the best this case, I preferred a string just incase in future we have an extra option. It will be easier to just add that option with a corresponding subroutine for external setting.

rdstern (approved):
Still seems ok

instat/dlgRegularSequence.vb
lloyddewit: @Patowhiz I agree that if we add another option in future, then `Boolean` is not extendable so may not be a good choice here.
I personally find `Enum` the most natural fit here, it is more readable, more constrained, and allows better static checking at compile time.

lloyddewit:
@Patowhiz just one outstanding comment above, what do you think?

rdstern:
@Patowhiz the last mention of this is conflicts.  I had earlier approved.

instat/dlgRegularSequence.vb
Patowhiz: @lloyddewit `Enums` it is. Done.

Patowhiz:
@rdstern you may want to test and review this urgently before conflicts. 
It has changes to `frmMain` and also other dialog changes due to enhancements made to control `ucrDataFrameLength`

lloyddewit:
> @Patowhiz I have approved. Has @lloyddewit remaining point been addressed?

yes, fully addressed thank you

lloyddewit (approved):

PR #7238, opened 2022-02-17
Added logical operations into Prepare>Column: Numeric>Transform dialog

### LINKED ISSUE #7236: A simple addition to the Prepare > Column: Numeric > Transform dialogue
@shadrackkibet I am assuming this is work for @anastasia-mbithe because it is "her" dialogue - and it is looking very good now.
But please check that this is a useful addition.  And possibly you should quickly do it, because it may take longer to explain!

We need to be able to do simple logical calculations - and this dialogue is the place to do it!
So please add another radio button on the first set, i.e. the set that starts with Round.
The label is Logical.
If chosen then it gives a choice of operator, from ==, downwards.  For most of these options there is a smallish field for a number to be typed.  
The order of the operators, and I assume the code is exactly the same as in the Filter sub-dialogue.  Which is why I wonder if @shadrackkibet might choose to do it quickly?  We might as well have even the options isna, so identical - deliberately - to the filter options!
Good for teaching too.  I find logical calculations are surprisingly new to many people!

anastasia-mbithe:
This fixes #7236
Hello @shadrackkibet, the controls are in place you can carry on.


shadrackkibet:
@anastasia-mbithe I have now added logical operations into this dialog. Could you first pull these changes locally and test them from your end?  I did some quick testing as follows;
- Loaded the famous `survey` data
- Selected the first Numeric option
- Switch to the logical radio button
- Added `field` into the receiver
- In the Input, I typed `3`
- Click OK

image

I also tried various logical options from the drop-down i.e `>,<,<=` etc
Also tried switching between different data frames and options at the top.

**Results**
- No bugs were identified.
- I got a logical column (`filed1`) added to the Survey data frame (snippet below)
image


You are welcome to test it your own way then we will invite @rdstern to test it.

anastasia-mbithe:
@shadrackkibet the only issue now is reset again. After testing all the operations, I clicked on Sort button then clicked reset and went back to logical instead of round.
Screenshot (68)

I can fix that.

shadrackkibet:
Please go ahead

anastasia-mbithe:
@shadrackkibet I fixed the reset issue and have also made changes to the test ok such that OK is not enabled when the ucrInputLogicalValues is empty for all operations except "is.na" and "!is.na"

shadrackkibet:
Great. @anastasia-mbithe I would like you to illustrate in writing here how you tested it. Then I think we are ready to invite others to test.

anastasia-mbithe:
@shadrackkibet Okay.
- So, I opened the survey data
- then I opened the dialog and filled the receiver,
- selected the Logical button and for all operations apart from "is.na" and "!is.na" test ok was disabled, when the input is empty
Screenshot (73)

But for "is.na" and "!is.na" test ok is enabled
Screenshot (74)

For the other operations when I filled the input with 6, test OK was enabled
Screenshot (75)

- I then did some different operations and then selected the Scale button,
- Then multiplied the column in the receiver by 5
- Then clicked reset and took me directly to the Round button.
Screenshot (76)

rdstern:
@anastasia-mbithe please check again.  Maybe I have done something wrong, but it doesn't seem to work at all.  I did this:
image

Notice the preview is not consistent with the logical function.  Neither are the results.  here they are:
```
field <- data_book$get_columns_from_data(data_name="survey", col_names="field", use_current_filter=FALSE)
field4 <- field
data_book$add_columns_to_data(data_name="survey", col_name="field4", col_data=field4, before=FALSE, adjacent_column="field")

```
It doesn't even give a logical function.

Please check each of the logical functions in turn, and also confirm that the preview updates correctly.

anastasia-mbithe:
@rdstern,The preview works fine now as well as the edit preview for any of the options. In the case of Logical, the operations update automatically in the preview but the values, it updates after clicking the Update button. But even without clicking Update button it still works to Script or OK. 

Screenshot (79)

You can test it from your side now.

rdstern (approved):
@anastasia-mbithe great.  From my point of view it all seems to be working now.  Over to @shadrackkibet or @lloyddewit to check the code

shadrackkibet (approved):

PR #7242, opened 2022-02-17
Updated the database to rename View Label dialogue

### LINKED ISSUE #7073: Add a new option to the Prepare > Data Frame > Convert dialogue
This needs an ok from @dannyparsons and then I hope it might be a relatively small task for @shadrackkibet or for him to allocate.  
This has become urgent, because we use labelled data - and have an outdated tutorial that needs improvement.  And I would like to explain about labelled data in the help system we are working on now.

The obvious dataset for testing is called efc and comes with the sjmisc package.
This is already nicely labelled.  

We also add labels in R-Instat.  For example, with our usual survey, see below, I made 2 numeric variables into factors and R-Instat has made value labels automatically!
  
image

That's all fine, but I would also like to be able to get rid of labels - and that is not currently so easy!  This is not important enough for a new dialogue.  I suggest either we have 2 buttons at the top of the current Convert dialogue (First is Convert - as now - though I still don't particularly like the options there).  Second is Remove Labels.

Perhaps better is to add a third button to the Delete Columns/Rows.  That already has Columns and Rows, so could we add a third button, which is Labels.  I think this might be a better place?

It has a multiple receiver as is there for the columns.  There is also a group box with the label Keep:  And 2 radio buttons The first is Labels and the Second is Numbers (or Levels?).  Default is labels.  

So, what does it do.  I suggest whatever they are currently (numeric, factor, or character), if you keep the label it becomes a character variable and if you keep the number/level it becomes numeric.

N-thony:
Partially fixes #7073 
@lloyddewit @shadrackkibet this PR just update the database by renaming the View Label column and Menu title.
image

### DIFF
--- a/instat/dlgTransposeColumns.vb
+++ b/instat/dlgTransposeColumns.vb
@@ -18,7 +18,7 @@ Imports instat.Translations
 Public Class dlgTransposeColumns
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
-    Private clsOverallFunction, clsTFunction As New RFunction
+    Private clsOverallFunction, clsTransposeFunction, clsGetColumnNames As New RFunction
 
     Private Sub dlgTransposeColumns_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
@@ -28,7 +28,7 @@ Public Class dlgTransposeColumns
         If bReset Then
             SetDefaults()
         End If
-        SetRCodeForControls(bReset)
+        SetRCodeforControls(bReset)
         bReset = False
         autoTranslate(Me)
     End Sub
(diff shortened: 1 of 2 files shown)


instat/dlgTransposeColumns.vb
```diff
@@ -28,56 +28,77 @@ Public Class dlgTransposeColumns
...
         clsOverallFunction.SetRCommand("as.data.frame")
         clsOverallFunction.SetAssignTo(ucrNewDataframe.GetText(), strTempDataframe:=ucrNewDataframe.GetText())
-        clsOverallFunction.AddParameter("x", clsRFunctionParameter:=clsTFunction)
-        clsTFunction.SetRCommand("t")
+        clsOverallFunction.AddParameter("x", clsRFunctionParameter:=clsTransposeFunction)
+
+        clsTransposeFunction.SetPackageName("data.table")
+        clsTransposeFunction.SetRCommand("transpose")
+        clsTransposeFunction.AddParameter("l", "columns", iPosition:=0)
+        clsTransposeFunction.AddParameter("make.names", "NULL", iPosition:=1)
```
shadrackkibet, line 83: But I see some other changes here.

shadrackkibet (changes requested):

instat/dlgTransposeColumns.vb
N-thony: Yeah, this contains the wrong changes, let me close and open another one. Thanks.

PR #7245, opened 2022-02-18
Added Pairs radio buttons to the Describe > Two Variables > Graphs dialog

Ivanluv:
This adds the `Pairs` radio buttons in the Describe > Two Variables > Graphs dialogue with its corresponding functionality that used the `ggpairs` function


rdstern:
@Ivanluv that's a good start.
a) Could you swap round the 2 buttons at the top.  Pairs is simpler than By, so it could be first.
b) The receiver in pairs is currently restricted to variables of a single type.  An exciting aspect of pairs is that the types can be mixed.  So can you not have this restriction.
c) When you change the button at the top, the selected variables in the multiple receiver, should clear.  Otherwise that will complicate things, given the restrictions are different.
d) There could well be (later) a Paits Option, but I am not sure the Plot options will work.  I need to check that more

rdstern:
@Ivanluv and @lilyclements I am adding Lily into the discussion, because I suggest this is a useful function, both in its own right and as a challenge to our dialogue system.  I suggest we could usefully merge in stages, but try - eventually - to offer much of the flexibility that you can get when using it in command mode.

I propose the first aspect of this flexibility could (perhaps) go on the main dialogue.  I am sure there will be more options for a Pairs Options button.  And the main Plot Options may, indeed be useful, though some tabs would be disabled - e.g. facets.

So I propose a big group box, called Type of Display.  (I is quite complicated, so it may  - later(?) - or after discussion, move to the Pairs Options sub-dialogue.  If not, then most of the other options would go on the sub-dialogue.

There are 3 checkboxes across (or down) labelled Upper, Diagonal, Lower (for the 3 arguments (upper, diag, lower) in the function.) Default is unchecked, and then that's all you see.  (If across then in Order Lower, Diagonal, Upper, because Lower nis the most important.)
If checked, then there is a set of pull-downs, perhaps combo boxes.  For Upper and Lower, they are Labelled Continuous, Combo, Discrete, NA.  If the check-boxes go across, then these go down.
For Continuous the options are:
('points', 'smooth', 'smooth_loess', 'density', 'cor', 'blank')
For Combo they are:
('box', 'box_no_facet', 'dot', 'dot_no_facet', 'facethist', 'facetdensity', 'denstrip', 'blank')
For Discrete - which could perhaps be better labelled as Factor:
('facetbar', 'ratio', 'blank')
For NA
('na', 'blank')
For Diagonal there are just 3 pull-downs.  (They are for a single distribution, so Combo makes no sense.
Continuous: ('densityDiag', 'barDiag', 'blankDiag')
Discrete (Factor): ('barDiag', 'blankDiag')
NA: ('naDiag', 'blankDiag')

Pretty exciting stuff.

Ivanluv:
This is still in progress

Ivanluv:
@rdstern could you look at this again. In the latest changes I have:

1. Swapped the pair and the by radio buttons where the pair radio button is now the first radio button and by is the second radio button
2. Removed restriction for the multiple receiver when within the pair radio button .Also the multiple receiver is now cleared when changing between the radio buttons
3. Added the colour single receiver for the pair option 
4. Added the Type of Display group box functionality for the pairs option

rdstern:
@Ivanluv great - will do

rdstern (dismissed):
Looks great.  Hope @shadrackkibet or @lloyddewit can approve the code.  This is a very atrractive addition!

instat/dlgDescribeTwoVarGraph.vb
```diff
@@ -54,23 +45,10 @@ Public Class dlgDescribeTwoVarGraph
...
-    Private clsGeomLine As New RFunction
-    Private clsStatSummaryHline As New RFunction
-    Private clsStatSummaryCrossbar As New RFunction
-    Private clsGeomFreqPoly As New RFunction
-    Private clsGeomHistogram As New RFunction
-    Private clsGeomDensity As New RFunction
-    Private clsAnnotateFunction As New RFunction
     Private strGeomParameterNames() As String = {"geom_jitter", "geom_violin", "geom_bar", "geom_mosaic", "geom_boxplot", "geom_point", "geom_line", "stat_summary_hline", "stat_summary_crossline", "geom_freqpoly", "geom_histogram", "geom_density"}
 
+    Public strFirstVariablesType, strSecondVariableType As String
```
lloyddewit, line 50: - could these be `private`?
- Should these be enumerations rather than strings?

instat/dlgDescribeTwoVarGraph.vb
```diff
@@ -100,17 +78,42 @@ Public Class dlgDescribeTwoVarGraph
...
+        ucrPnlByPairs.AddToLinkedControls({ucrReceiverSecondVar, ucrInputCategoricalByCategorical,
+                                          ucrChkFlipCoordinates}, {rdoBy}, bNewLinkedHideIfParameterMissing:=True)
+        ucrPnlByPairs.AddToLinkedControls({ucrChkLower, ucrReceiverColour}, {rdoPairs}, bNewLinkedHideIfParameterMissing:=True)
+
+        ucrChkLower.SetText("Lower")
+        ucrChkLower.SetLinkedDisplayControl(grpTypeOfDispaly)
+
+        ucrChkDiagonal.SetText("Diagnol")
+
+        ucrChkUpper.SetText("Upper")
```
lloyddewit, line 95: - should this be "Diagonal"?
- these `SetText` calls are duplicated further down, are they needed here?

instat/dlgDescribeTwoVarGraph.vb
```diff
@@ -755,11 +874,78 @@ Public Class dlgDescribeTwoVarGraph
...
+            End If
+        End If
+        AddRemoveColourParameter()
+    End Sub
+
+    Private Sub ucrSelectorTwoVarGraph_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSelectorTwoVarGraph.ControlValueChanged
+        AddDataFrame()
+    End Sub
+
+    Private Sub AddColumnParameter()
```
lloyddewit, line 910: This is only called once, dow e need a seperate `Sub` for this?

lloyddewit (changes requested):
Thanks looks good, just a few questions?

instat/dlgDescribeTwoVarGraph.vb
Ivanluv: > * could these be `private`?

yes

> * Should these be enumerations rather than strings?

They can't be an enumeration because there values are changing

instat/dlgDescribeTwoVarGraph.vb
Ivanluv: I have removed the duplications and correctly updated the spelling

instat/dlgDescribeTwoVarGraph.vb
Ivanluv: Done

Ivanluv:
@lloyddewit  could you have a look again

instat/dlgDescribeTwoVarGraph.vb
lloyddewit: ok understood

lloyddewit (approved):

rdstern (approved):
Still looks good!

PR #7251, opened 2022-02-22
Improvements on the grid in the Contrasts dialog. 

### LINKED ISSUE #4356: Improving the contrasts dialogue
This first topic is low priority, but other improvements, mentioned below, are straightforward.

With the default contrasts for a factor the output for coefficients when modelling is clear:

```
Coefficients:
             Estimate Std. Error t value Pr(>|t|)    
(Intercept)    59.600      3.288  18.127  < 2e-16 ***
Variety.OLD   -14.159      3.654  -3.875  0.00048 ***
Variety.TRAD  -29.600      3.700  -7.999 3.15e-09 ***
---
```

But after setting my own contrasts using the Contrasts dialog the default names for the new contrasts are 1 and 2, so I have to remember what they mean.

```
Coefficients:
            Estimate Std. Error t value Pr(>|t|)    
(Intercept)  45.0137     1.3431  33.514  < 2e-16 ***
Variety.1    14.5863     2.3254   6.273 4.33e-07 ***
Variety.2     0.4275     1.6285   0.262    0.795    
---
```
I've found a way of giving names to the contrasts in R. So we could add this to the Contrasts dialog? Currently we fill a grid to set the Contrasts, so we could allow editing the column names of the grid to name the contrasts, but I don't think this will be obvious to the user. Any other suggestions? Is it important to be able to name them at all?

anastasia-mbithe:
This Partially fixes #4356 
@shadrackkibet and @rdstern, This is still in progress.


rdstern:
@anastasia-mbithe This is very good progress on this dialogue.  As I see it, there are 2 small tasks and one larger one remaining:
a) When the dialogue runs it  would be good to have something displayed in the Output window.  I suggest something like 
`Factor xxx has yyy contrasts`.  xxx is the name of the factor and yyy is the same as the entry, e.g. Treatment, User Defined.
b) It is very good that you can paste into the grid, when `User Defined` is selected.  Could Copy (ctrl<C>) also be implemented?  (That should be very easy.  Perhaps check with @N-thony if need be.
c) Now the bigger one.  When in User Defined, for any factor it should (initially) show the user defined contrasts for that factor, if they have been defined.  If not, then it should show an empty grid, i.e. zeros.  Currently - I am using the usual survey dataset, it keeps the display within the dialogue, when you move from one factor to another.
This should be quite easy to do, because the next dialogue, i.e. Factor Data Frame, already does it.  So it must be using a function that displays the user defined values when they exist.  You should just use that function and display zeros otherwise.

rdstern (changes requested):
@anastasia-mbithe I had another look at the dialogue and believe you have now added the <ctrl> C, and that's great.  The other tasks should be quick, if you are (or will be) in Maseno and get some help from Antoine or Shadrack?  
And I am afraid I have thought of something else that would be good to add - and again I am sure @N-thony could help if you need it.  This is to display the labels in a factor as well.  I am using the usual survey data and the labels for the variety are NEW, OLD, TRAD.  Currently we just have 1, 2, 3 displayed.

anastasia-mbithe:
@rdstern We managed to have the grid show existing contrasts in the grid, when user defined option is selected. However, we did not find a way to change the labels that appear i.e village 1, village 2 etc. The contrasts function in R is set to give such output when another contrast other than Control Treatment is used.

rdstern (dismissed):
@anastasia-mbithe well done.  That is very nice now. I am approving and hope that @shadrack, or @lloyddewit can check the code, so this can be merged.
Once merged, then we can have another issue, but not urgent for a) Allowing fixed format for numbers when needed.  (That's for polynomial contrasts) and b) adding the factor labels.

shadrackkibet (commented):
Please first remove commented code. Otherwise, explain why we still need it.

anastasia-mbithe:
@shadrackkibet, The commented code in the Getcontrast() sub, allows editing of the column headers and row headers, hence could be a way to fix the label issue, but it works for the control Treatment option only and bugs with the other options. I thought of keeping it so that if needed, someone could improve it in future.

anastasia-mbithe:
@rdstern @shadrackkibet, The dialog now gives a message in the output window; Factor xxx has yyy contrasts.

shadrackkibet (changes requested):

instat/dlgContrasts.vb
```diff
@@ -194,9 +200,44 @@ Public Class dlgContrasts
...
+            TestOKEnabled()
+        End If
+    End Sub
+
+    Private Sub GetContrast()
+        Dim clsGetColumnFunction As RFunction = ucrReceiverForContrasts.GetVariables()
+        clsGetColumnFunction.RemoveAssignTo()
+        Dim clsGetContrastFunction As New RFunction
+        Dim expContrasts As SymbolicExpression
+        ' Dim strColumnHeaders() As String
```
shadrackkibet, line 215: Remove

instat/dlgContrasts.vb
```diff
@@ -33,10 +33,11 @@ Public Class dlgContrasts
 
         ' Add any initialization after the InitializeComponent() call.
         grdLayoutForContrasts.SetSettings(WorkbookSettings.View_ShowSheetTabControl, False)
-        grdLayoutForContrasts.SetSettings(WorkbookSettings.View_ShowHorScroll, False)
+        'grdLayoutForContrasts.SetSettings(WorkbookSettings.View_ShowHorScroll, False)
```
shadrackkibet, line 36: Remove

shadrackkibet:
@anastasia-mbithe could you also change the PR title? This is too general. What exactly improved?
image

anastasia-mbithe:
@shadrackkibet Is the new title okay?

shadrackkibet:
@rdstern please test and re-approve.

rdstern (approved):
image

This can be for the future, i.e. to be able to sort out the precision that is being used in these displays.  But I would be happy for this to be merged now, and that becomes a new, and perhaps more general issue.

shadrackkibet (approved):

PR #7243, opened 2022-02-17
Renamed Dialogue controls texts in the database

N-thony:
Fixes (partially) #7073 
@lloyddewit @shadrackkibet this just update the database by renaming the menu and dialogue text to `View/Delete Labels` for View Label dialogue.
image

PR #7253, opened 2022-02-22
Hide drop empty rows checkbox when no file is selected in import dataset dialog

### LINKED ISSUE #7173: Hide Drop empty Rows and All files of this type checkboxes
- When there is no file selected i.e the path is empty these checkboxes should not be visible?

- I also wonder why the preview message - "Preview not implemented for this type of file" is displayed here. Please hide the message or we change it to something else like "No file selected"

Note: Be aware of empty space in between controls.
 
- Why is Ok enabled? When I click Ok error!
image
 

@rdstern what do you think?

- To replicate select a file and click reset.

image

mmulahvictor:
Fixes #7173 

- I have hidden the drop checkbox when no file is selected.
- I have disable ok which appeared after reseting the dialog.

This is ready for review


rdstern (changes requested):
This is a nice example of a simple task where I would like the programmer, here @mmulahvictor to do more testing and also to read the task carefully, before sending it for review.
a) When I press Reset as suggested by @shadrackkibet it is perfect and all works fine.  But when I return to the dialogue again, the option has returned.  So it isn't obeying the rule that the dialogue re-opens as it was left. 
b) Shadrack suggested the message on not displaying files of this type could/should(?) also be changed.  It isn't yet.  Either there should be a comment that this remains to be done, or that the programmer doesn't agree, and so decided not to do it.

lloyddewit:
> 

@mmulahvictor when PR #7243 is merged, please could you pull the changes into this PR so that the new text can also be tested, thanks

mmulahvictor:
@shadrackkibet can you look at the changes I have made above?

shadrackkibet:
@Ivanluv could you review the following PR?

lloyddewit (approved):

lloyddewit:

rdstern (approved):

PR #7256, opened 2022-02-28
Added columns header colour when columns selection is made

N-thony:
@shadrackkibet I was not able to push directly into your branch, here is the change corresponding to changing the column header text colour in the datagridview when the columns selection is made. Have a look. Thanks.


shadrackkibet:
Thanks, did you push the changes you made?

PR #7263, opened 2022-03-03
Added options for renaming and deleting column selections

shadrackkibet:
fixes #can't trace the issue at the moment but it should be somewhere. Apologies, I will try to find it. @rdstern please help.

This PR adds options for renaming and deleting column selections into **Rename Objects** and **Delete Objects** dialogs. To test it;
- Create column selections using **Prepare> Data frame> Column selection**
- Rename column selection using **Prepare> R Objects> Rename Objects**
- Delete column selection using  **Prepare> R Objects>Delete Objects**

**Note**: You have to select **Type** - "Column selections" while using these dialogs.


rdstern (approved):
@shadrackkibet that seem to work for me.  I made 2 selections, and renamed them and then tried to delete the one I was using - which it doesn't allow.  Deleted the other one.
In the longer run I wonder about being able to view and also reorder these objects, but that can wait?
If @lloyddewit is happy we can merge.

PR #7265, opened 2022-03-03
Naming and adding textboxes into Qc temperature dialog

### LINKED ISSUE #7025: Naming on QC Temperature dialog
This dialog is designed for temperature but can be used for other elements. I suggest a few changes to make it a bit more general:

1. For "Acceptable Range":
- Change the display to `Acceptable Range Element 1:` `From:` [box] `To:` [box]
- Remove the "oC" labels as it could be any unit
- Change the nud to textboxes. They only allow fixed increments and I think a free textbox that restricts to numeric is better.

image

2. The produced column names have static names with "tmin" and "tmax" which is confusing for different columns. e.g. see the code below. I used a column `sunh` but the produced column was `range_tmax`.
The best solution would be that is uses the name of the column provided i.e. `range_sunh`. 
If that's difficult, then use a general standard name such as `range_e1` and `range_e2` (for element 1 and element 2)
```r
# Code generated by the dialog, Climatic Check Data Temperature

acceptable_range_test__tmax <- instat_calculation$new(type="calculation", function_exp="sunh < 0 | sunh >= 16", calculated_from=list("sunshine_6stns"="sunh"), result_name="range_tmax")
grouping <- instat_calculation$new(type="by", calculated_from=list("sunshine_6stns"="station_id"))
filter_calculation <- instat_calculation$new(type="calculation", function_exp="range_tmax", calculated_from=list("sunshine_6stns"="sunh"), sub_calculations=list(acceptable_range_test__tmax), manipulations=list(grouping), result_name="QC")
filtered_data <- instat_calculation$new(type="filter", function_exp="QC", calculated_from=list("sunshine_6stns"="sunh"), sub_calculations=list(filter_calculation), save=2, result_data_frame="qcTemp", manipulations=list(grouping))
data_book$run_instat_calculation(calc=filtered_data, display=FALSE)
rm(list=c("filtered_data", "filter_calculation", "acceptable_range_test__tmax", "grouping"))
```

3. I don't see the purpose of the `QC` column created at the end, which always seems to be `TRUE`?
4. Change "Same" label to "Days:" [box]. We should generally have the label in front of a textbox for consistency and for being clear in other languages where the order of words may be different.
5. Remove "oC" from the "Jump" label
6. Add : to "Coef:"
7. All the produced columns should follow naming conventions and be in lower case

Vitalis95:
Fixes #7025 
@rdstern @dannyparsons @shadrackkibet ,  this PR makes the following changes ;
- Changing the nud to textboxes in Acceptable Range
- produced column names to have standard name such as _e1 and _e2 (for element 1 and element 2)
- Removing QC column which was being created at the end
- Removing  "oC" label from Acceptable Range and Jump and Changing "Same" label to "Days:"

@N-thony  made the following changes in database #7243 
changing "to" to [box] To: [box] in Acceptable Range
Adding : to "Coef:" in Outlier.

### DIFF
--- a/instat/dlgClimaticCheckDataTemperature.Designer.vb
+++ b/instat/dlgClimaticCheckDataTemperature.Designer.vb
@@ -503,7 +503,7 @@ Partial Class dlgClimaticCheckDataTemperature
         Me.ucrInputRangeElement1Max.AutoSize = True
         Me.ucrInputRangeElement1Max.IsMultiline = False
         Me.ucrInputRangeElement1Max.IsReadOnly = False
-        Me.ucrInputRangeElement1Max.Location = New System.Drawing.Point(296, 265)
+        Me.ucrInputRangeElement1Max.Location = New System.Drawing.Point(297, 265)
         Me.ucrInputRangeElement1Max.Name = "ucrInputRangeElement1Max"
         Me.ucrInputRangeElement1Max.Size = New System.Drawing.Size(44, 21)
         Me.ucrInputRangeElement1Max.TabIndex = 49
--- a/instat/dlgClimaticCheckDataTemperature.Designer.vb
+++ b/instat/dlgClimaticCheckDataTemperature.Designer.vb
@@ -35,7 +35,6 @@ Partial Class dlgClimaticCheckDataTemperature
         Me.rdoIndividual = New System.Windows.Forms.RadioButton()
         Me.rdoMultiple = New System.Windows.Forms.RadioButton()
         Me.lblNudSame = New System.Windows.Forms.Label()
-        Me.lblNudDiff = New System.Windows.Forms.Label()
         Me.ttOutliers = New System.Windows.Forms.ToolTip(Me.components)
         Me.lblCoeff = New System.Windows.Forms.Label()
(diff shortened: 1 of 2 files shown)


Vitalis95:
@shadrackkibet I have made the changes you suggested.

rdstern:
@Vitalis95 a) when I open the dialogue it shows me the From:  for each of those controls.  I suggest (like the rest of them) that this is only shown when the options are chosen.
b) Could you also check the vertical location of the two sets of textboxes.  I am poor at this, but they don't seem quite lined up, both horizontally and vertically?
c) I assume it is supposed to always be e1 and e2 as you say above, rather than picking up the name of the element?  If so, then that's fine as you have it now.
d) I assume you are removing the oC in case the same dialogue is used for other elements than temperatures?  If so, then also remove it from Difference.
e) The Days check is to check how many consecutive days are the Same.  If that is clearer now with Days repeated, then that's fine.

Vitalis95:
@rdstern,  I have made the changes.

rdstern (approved):

shadrackkibet (approved):

Issue #7260, opened 2022-03-02
Further improve Two Variables Graph


rdstern:
@Ivanluv you have added the Pairs option to this dialogue.  Now to improve the (already good) By option.
a) I suggest the title could become `Two Plus Variables Graph`  can we improve on that?
b) Fix the bug, that the categories (e.g. numeric by numeric don't always display.  To see this, start (in the survey data) with 2 sets of factors, e.g. variety, village, by fertgrp.  There is now no group-box, below the Plot Options.  Now change the second (single receiver to yield.  Still no group box.  Press Ok.  Then return to the dialogue.  Lo and behold there is the group box!  
I think it is there for the other 3 options.  Is it there, and just not shown for the categoric by categoric?  If not, then make it there, even if there are currently no options.  (@Ivanluv I can show you this quickly on skype)
c) Now to possible improvements.  The single receiver is not quite in line with the multiple receiver.  I suggest that this could also move up slightly.  Then add a further single receiver with label `(More) Facets:`  This is just for a factor variable.  
(I need to check this option is the best way to add a third variable to the plots here?)
d) Whenever there are facets we could also add a checkbox, Include Margin in the group box below.  (This is either because there are multiple variables in the first receiver, or because the third receiver is used, or both.
e) When there are 2 facet variables, then add an option in the group box, `Swap Facets`.  Default unchecked.
f) If, as I suspect, there are currently no options for the factor by factor graph, then consider what can be added.  (They may be there now, but currently not visible.

I need to check some of these suggestions with @dannyparsons and @lilyclements before there is too much work.  But item b) at least, can be fixed immediately.

rdstern:
@Ivanluv this is already a good start.
But (of course) I have now found another bug!  I tried 2 categorical variables, namely village by fertgrp.  This works fine for either barchart or mosaic.  Great.  But when I add another variable into the multiple receiver, I get an error when I do a mosaic chart.  Perhaps it can't do facets in the same way?  Could you investigate.

I wonder also whether there could be other options?

@shadrackkibet do you think the labels could be added in this dialogue for the bar charts, and possibly also to the mosaic - if we keep it?

Ivanluv:
@shadrackkibet  I am getting the following error after adding the labels 
![image](https://user-images.githubusercontent.com/45068931/157208503-8aef6bd2-2646-4143-96cc-a63901357922.png)
[...]

PR #7266, opened 2022-03-03
 Improving the Two Variables Graph

Ivanluv:
This fixes part of the points addressed in issue  #7260 i.e

1. Changing the title of the dialogue to `Two Plus Variables Graph `
2. Hiding of the group box below the` Plot Options` button when having the categoric by categoric options (within the by radio button)
3. Aligning of the multiple and single receivers


lloyddewit:
@Ivanluv There's a conflict with the database. If you pull the database from master and make the change again, then it should resolve this, thanks

PR #7271, opened 2022-03-04
Fixed bug on the factor levels control being hidden in some dialogs

### LINKED ISSUE #7269: Disturbing bug in climate data entry in Version 0.7.4
@N-thony and @lloyddewit this is an innocent looking bug.  
image

The control for the station is not visible.  I hope I am wrong to be disturbed, but it is another oddity that has been introduced in Version 0.7.4.  Here is 0.7.3:
image

This seems different, but is the 4th oddity that we have found in Version 0.7.4 that has been introduced.  I am worried there may be more to be discovered, and hope there is perhaps a single reason for them?  

I also tried these data when there was a filter in place, and it told me there was no such period.  Using Version 0.7.3 the filter worked fine

N-thony:
Fixes #7269 
@rdstern  I had a quick discussion on this with @ChrisMarsh82 and then found that the problem was more general mainly under the following dialogues and ucrInputFacotrLevels itself:

- dlgCalculateTreatementDifferences
- dlgClimaticDataEntry
- dlgCalculateTreatementLines
- dlgSetupForDataEntry
- ucrInputFacotrLevels 
To fix this, we just set the `AutoSize property to False` and also adjusted the `Size` of ucrInputFacotrLevels and it corresponds to the above dialogues.

Please take a look. Thanks.

### DIFF
--- a/instat/dlgCalculateTreatmentDifferences.Designer.vb
+++ b/instat/dlgCalculateTreatmentDifferences.Designer.vb
@@ -82,25 +82,23 @@ Partial Class dlgCalculateTreatmentDifferences
         'ucrInputFactorOption2
         '
         Me.ucrInputFactorOption2.AddQuotesIfUnrecognised = True
-        Me.ucrInputFactorOption2.AutoSize = True
         Me.ucrInputFactorOption2.AutoSizeMode = System.Windows.Forms.AutoSizeMode.GrowAndShrink
         Me.ucrInputFactorOption2.GetSetSelectedIndex = -1
         Me.ucrInputFactorOption2.IsReadOnly = False
         Me.ucrInputFactorOption2.Location = New System.Drawing.Point(272, 164)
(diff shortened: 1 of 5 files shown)


rdstern (approved):
@N-thony great.  I have checked where I found the problem, and that is fixed.  I also looked at the setup for data entry and that also seems fine, but I didn't see it before.
I have not checked others, because you just give the function names, so I'm not quite sure which dialogues, nor what I should look for.  But I assume the problem is solved, and am very relieved.  Over to @lloyddewit or @shadrackkibet for second approval and merge.
That's a relief!

shadrackkibet (approved):
@africanmathsinitiative/developers I think we might have this kind of problem in other places. Can we get volunteers to quickly check and report on dialogs with this kind of problem? Let us get volunteers by menu.

Shadrack - Climatic Menu

PR #7272, opened 2022-03-04
New Random Split dialog

### LINKED ISSUE #7227: Add a new dialogue possibly called Random Split using rsample?
From what I understand of what's needed this is sufficiently different to anything we have, that it is a new dialogue.
a) We have subset, which uses our filter system, but has no element of randomness.  It leads to a single new data frame.
b) We have random subset, which takes a single variable and produces multiple subsets of that variable in a new data frame.
This dialogue will keep all the variables (like filter) and split the data frame into two components - called Train and Test.  So it produces 2 data frames - at least that is part of the rsample package.  
So, it could be called Split, or Random Split, or even Train & Test.  I was thinking we could refer to the package and call it Rsample, but maybe `Random Split` is a better name?  Looking further, this is only a small part of what rsample is doing.  So, Random Split looks like the name.
Looking at rsample, this is mainly the` initial_split `function, (which (I think) just adds a logical (or other variable to the set.  There is also the `initial_time_split` function.  It is then followed by the `training` and `testing` functions that are also described there.  

I think this is what is needed for the caret package, where we have started.  See [here](https://cran.r-project.org/web/packages/caret/vignettes/caret.html), for an example of the caret package that (I think) could use the results we produce.  (Good to use a package from tidymodels where we can and it looks simpler than the process in the caret package.)  So the results should help with the new caret dialogue.
The above should be investigated first.  
a) There is also an rsample2caret function, which I don't think we need, for our simple use of rsample.  (Maybe interesting later.)
b) It seems very easy to use rsample with strate.  It also discusses nesting (e.g. IPUMS data has households and individuals, and we would like to sample on the households - then taking all individuals within selected households.  Unless it is easy, I suggest we leave this for later.  (Not difficult to write, but let's start with the low-hanging fruit!) 

I suggest the dialogue will be very easy.

a) Possibly 2 buttons at the top maybe Sample and Time Series
b) With both there is just the data frame on the left.  
c) With Sample there is Stratifying Factor as a checkbox top0 right.  Default unchecked
d) With Sample: If checked, then there is the usual selector with the data limited to factors.    
(Note rsample allows numeric and I am not keen on that.  It complicates the dialogue, but - for this use - I would prefer users to think about the split separately.)
e) With Sample, and only if Strata an up- down called Pool appears.  It is an up-down limited to 0 to 0.15 in steps of 0.01.  Default is 0.1.  (It may not work with 0 - we should check!) 
e) For both:  Fraction.  Can be an up-down limited to 0.01 to 0.99 and default 0.75'
f) For Time series a checkbox labelled Lag.  I am suggesting a checvkbox, because I don't think it will be needed often.  If checked, then an updown with default 0 and going up (restricted to non-negative integers), though should check if negative makes sense and is allowed.
g) For both: Save Sample to: and a variable.  It produces an object, which is used by training and testing, see below.  I hope this includes a variable perhaps logical that we can add to t5he existing data frame.  
h) Then 2 checkboxes called `Save Training Data` and `Save Testing Data`.  Default not sure, perhaps unchecked.  If checked, then the data frame name adding _train and _test.

Note @lilyclements is looking at the package as a whole.  Perhaps @shadrackkibet can check that this subset and the design are sensible?  If so, then they should be relatively simple for an intern.

mmulahvictor:
Fixes #7227 
I have added controls to the Random split dialog.
I have also named the controls appropriately.
**This is not yet ready for review, it is still work in progress**


mmulahvictor:
@rdstern and @shadrackkibet can you test the first button, "Sample", if it is working correctly. In the mean time I am finishing up with some few changes on the second button, "Time series". Thank you.

rdstern:
@mmulahvictor where is it?  I was looking for the dialogue.  Where is it on the menu?

lloyddewit:
@mmulahvictor I assume that this dialog will be accessed via Prepare / Data Reshape / Random Split... .
This menu option is still greyed out.
Please could you let us know the status of this PR?
Thanks

lloyddewit:
Replaced by PR #7519

PR #7274, opened 2022-03-06
Bug fix: Removing labels when transforming a labelled numeric variable

### LINKED ISSUE #7249: Recoding a factor is problematic when it is labelled. So is Prepare > Data Frame > Replace Values
@N-thony this may also relate to your task of being able to delete labels #7073.
The example I have is the Sadore data, in Climatic > Niger - worksheet Sadore leaving 4 lines.
Make Mois into a factor.  Then perhaps in the Levels/Labels dialogue add Levels.  This adds value labels to the variable
The column named Mois has 14 levels.  Levels 1 and 14 are Janvier and janvier.  Levels 2 and 13 are fevrier and Fevrier.  (All the other months are lower case.)
I can use the Levels/Labels dialogue to change the labels.  Or I can use Prepare > Column: Factor > Recode Factor.
The result is different, depending on which I use!  I thought the solution was to delete the labelling, but I think there is a better one.  
a) In the `Recode Factor` I can recode the labels in the obvious way.  Then I (correctly) get 12 levels.  The only possible problem is that janvier has level 14!  That doesn't do any harm, but looks odd.
b) So one solution would be to allow the user (in this situation) to recode the levels too.  Maybe recode the levels would be an alternative to recoding the labels.  
image
So, in this dialogue, there is a checkbox, for labelled data, default unchecked, which says `Recode Levels`.  If checked then the Last column above is replaced by the levels and they have to be numeric, but can be changed.  
That would be neat!
c) I can also use the Right-Click Levels/Labels dialogue.  This allows me to change the labels.  But there are still 14 levels to the factor.  I am now stuck.  I think it may work ok, but it has become messy.  Worse, here I can edit the levels and then it gives me an error.
One drastic solution is to not permit editing in this dialogue.  Something needs to change though.  

I suggest @N-thony can make the changes, (he made the changes recently in the Factor Recode dialogue) but it needs a small initial input from @dannyparsons. 
Except I have now tried the Prepare > Data Frame > Replace Values dialogue, also on Labelled data.  I have tried this with our usual survey data.  Variety, make it numeric, and then back to a factor again.  Now it has labels attached.  Now make it numeric again and change 2 to 1.  This has now changed all the frequencies.  Also change 3 to 5 and then make it a factor again.  

So, I now wonder if we are getting ourselves into a mess, by being able to edit levels.  We don't use labelled data much - we like factors.  These are a luxury.  So I now suggest simply as follows:
a) We remain powerless to change levels - we just change labels, for Factors.
b) If a variable is labelled and is a factor, and we do a factor recode, we can warn and then, if we proceed, then the labelling is deleted.  
c) If we change a number in a labelled numeric variable (in the grid, or in recode, etc?) then we warn and delete the labelling.  

If you don't like this in your data, then make a copy of the variable (with labelling), before making the change.

Though maybe we can do better, by having options to also edit the levels?  I used the Village factor and lumped the values less than 8 together.  Then made it numeric - with labelled convert.  Then factor again.  Here is the current result - nice!  (That made the extra levels for me!
image

I assume we will also have to delete labelling if we change strings?  Oh I am confused!  

Maybe instead, when we edit a factor, we should delete labelling and then reinstate it.  Then, if we change numbers in a labelled numeric variable, could we then reinstate, with default labels.  So, we reinstate levels when we edit a labelled factor or character variable and we reinstate labels, when we edit a labelled numeric variable?  I like this solution?

This needs a @dannyparsons and/or @volloholic input and decision!

shadrackkibet:
fixes #7249

@rdstern wrote in the issue as follows;

> However, try this (silly) sequence and you see our labelling can cause nonsense.
a) Perhaps duplicate fert first. Then use the calculator to calculate size+fert, into calc. Then make calc a factor. This adds labels, and the factor has 17 levels.
b) Make fert into a factor, and then back to numeric. Now do a) above. the calculated column is now a factor with 22 levels and the levels/labels information is a mess.


lloyddewit (approved):

lloyddewit:

rdstern:
@lloyddewit I assume @shadrackkibet is making changes to this pull request first.  Currently it deletes the label information when a factor is made numeric, and it is often useful to keep the labelling then - so it can be made into a factor again, with both levels and labels.  What goes wrong is when a labelled numeric is transformed, e.g. multiplied by 2, or combined with another variable.  Then the labels should be deleted from the transformed variable.

rdstern (changes requested):
@shadrackkibet you may want to do this quickly, or pass it to someone else.
I think it is probably simpler than I thought before.  Check if I am right here, but it is just when you do a calculation or a transformation (Prepare > Calculate, or Prepare > Column: Numeric > Transform) that the labels could remain and become silly.
They don't seem to do much harm, and making a transformed variable into a factor makes new labels - that are ok.  So my solution is simply to delete any labels from those functions.  It doesn't matter if it isn't labelled, so it is just that those resulting variables should not be labelled?

shadrackkibet:
@rdstern this is easy todo.

@N-thony this is similar to what you did in recode numeric dialog, the only difference is you remove before you transform a variable. Can you do the same in  **Prepare > Calculate, or Prepare > Column: Numeric > Transform**.  We can run a check internally to check if labels exist then remove them. Either way, it doesn't matter if you remove the label even if it is not there.

rdstern:
@shadrackkibet and @N-thony I am just checking when the labels are deleted here.  You say "you remove before you transform a variable."  I was assuming that the resulting variable , i.e. after you transform, that you make sure that the result isn't labelled?

N-thony:
> @shadrackkibet and @N-thony I am just checking when the labels are deleted here. You say "you remove before you transform a variable." I was assuming that the resulting variable , i.e. after you transform, that you make sure that the result isn't labelled?

@shadrackkibet this was done in the Recode Factor dialogue  i.r. removing the labels after recoding the factor variable. I will do the same here as @rdstern was assuming.

rdstern (approved):
@N-thony great - from my point of view this now seems to work fine.
Sorry for the delay, but I have been struggling with being able to access the pull requests, following updates in Visual Studio.
I hope perhaps @shadrack can check the code so it could be merged?

rdstern:
@shadrack, isn't this what @N-thony has now done in #7321 ?  If so then perhaps this one can be closed?

PR #7278, opened 2022-03-07
Updated the data view info when selection is applied

### LINKED ISSUE #7277: Update the dataview info when column selection is applied
We recently adjusted the dataview info to be able to fit the layout panel when the filter is applied, but now we need the same thing when column selection is applied

image

N-thony:
Fixes #7277 
I have added the info concerning the column selection when it is applied to DataView, I also adjusted the size of the info when the application is minimized or maximized.

### DIFF
--- a/instat/Model/DataFrame/clsDataFrameFilter.vb
+++ b/instat/Model/DataFrame/clsDataFrameFilter.vb
@@ -26,7 +26,9 @@ Public Class clsDataFrameFilterOrColumnSelection
     Protected _bFilterApplied As Boolean
     Protected _bColumnSelectionApplied As Boolean
     Protected _iFilteredRowCount As Integer
+    Protected _iSelectedColumnCount As Integer
     Protected _strFilterName As String
+    Protected _strSelectionName As String
 
     Public ReadOnly Property iFilteredRowCount As Integer
         Get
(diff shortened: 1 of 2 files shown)


rdstern:
I am approving, because it now works.  But I wonder whether we should try to do better?  Here it is with a select and a filter
image

This is default and they are incredibly small.  I am not good on design, but suggest that 
a) When a filter if active we say Rows 1 to 108, i.e. we only have the word "Showing" when there is no filter.  Similarly for Selects?
b) We drop the word Active in each case as well.  So it is just Filter::, and Selection?

@N-thony and @shadrackkibet what is your view?

rdstern (approved):
Approving, but also see comment with a suggestion.

shadrackkibet:
@rdstern I agree.

@N-thony please change.

N-thony:
> I am approving, because it now works. But I wonder whether we should try to do better? Here it is with a select and a filter image
> 
> This is default and they are incredibly small. I am not good on design, but suggest that a) When a filter if active we say Rows 1 to 108, i.e. we only have the word "Showing" when there is no filter. Similarly for Selects? b) We drop the word Active in each case as well. So it is just Filter::, and Selection?
> 
> @N-thony and @shadrackkibet what is your view?

@rdstern I made the above change, please test. 
Just to check, is it `Filter:` or `Filter::` you suggested?

shadrackkibet (approved):
@rdstern if you approve then we can merge.

rdstern (approved):
@shadrackkibet I am approving, though I quite like @Patowhiz suggestion.  If merged as it stands, then perhaps this can be considered later?  As you wish.

shadrackkibet:
> Code looks good. @rdstern @N-thony I wonder why you preferred the column label selection going horizontally. I'd suggest it to be added vertically. Add this can be done only when there is a column selection. This way we won't need to reduce the font size.

@N-thony could you open an issue for this? then you can consider it on a separate PR.

@rdstern another option is to put the label at the top where variable's names are.

PR #7281, opened 2022-03-09
Improving the Two Variables Graph

### LINKED ISSUE #7260: Further improve Two Variables Graph

Ivanluv:
this replaces PR #7266 and fixes (partly) #7260 i.e

1. Changing the title of the dialogue to `Two Plus Variables Graph`
2. Hiding of the group box below the Plot Options button when having the categoric by categoric options (within the by radio button)
3. Aligning of the multiple and single receivers


Ivanluv:
@shadrack after adding the labels controls am getting the following error, any ideas on how to fix this?
image
image
image

rdstern (changes requested):
@Ivanluv I wonder if you have included the right bit of code from @shadrackkibet implementation in the full bar chart dialogue?  In the bar chart it works either for counts or for values (i.e. for summary data.).  It looks as though you are using the code for the values, when that for the counts is what you want.

Ivanluv:
@rdstern from the discussion we had with @shadrackkibet I will be using counts

Ivanluv:
@lloyddewit I have tried pulling the master and making the changes but still getting the conflict on the translation file. Is there a way I can fix this

lloyddewit:
> @lloyddewit I have tried pulling the master and making the changes but still getting the conflict on the translation file. Is there a way I can fix this

I suggest that you pull the database file from the master again. This should solve the conflict but you'll lose your database changes. We can then approve/merge this PR without the database change and make the database change in a new PR.
Does this make sense or can you think of a better way?
thanks and sorry for the inconvenience

Ivanluv:
@lloyddewit  I have created a new PR #7307 to replace this

PR #7282, opened 2022-03-09
Added an option to include frequencies into the Prepare>Column: Factor> Factor Data Frame Dialogue

### LINKED ISSUE #7229: Improve the Prepare > Column: Factor > Factor Data Frame dialogue
I have long felt we need this dialogue and now I know why!  There may be more reasons, but one is so we can tidy data efficiently.  I have been looking at the Open Refine package - that we link to.  It is designed to tidy messy data.  
One thing it does is to be able to tidy a messy factor variable - and at the factor, rather than at the data level.  We can almost do that and we will with the support of the improved factor data frame.
I used the usual survey data as an example.  I also made a missing value in one of the villages to check that it would work ok when there are missing values.
Here is the current dialogue:

image

a) By default it gives the contrasts.  It gives an error if you untick the contrasts.  So that is the first thing to fix.
b) It has an option to replace the existing data frame, when that exists.  That doesn't seem to happen.
c) Without the contrasts it would just give a single variable, namely just the levels.  Sometimes That's what we need.  But the Levels/Labels also gives the Frequencies.  So please could there also be an option to give the frequencies.  (We can do it already, so perhaps we just need to add the code from below.  That's what perhaps @shadrackkibet could check.

image
d) I think the above shows what I would like just from the dialogue (in addition to its use for contrasts!).  I got this by first defining my own contrasts - to check that aspect.  (That works, though we could use improvements on that dialogue later.) 
Then I used the Factor Data Frame dialogue.  That produced just 4 rows, i.e. not the NA row.  Then I use the Prepare > Data Reshape > Column Summaries dialogue.  I summarised Village, By Village, and the only summary was the Count.  
This added that variable and also the NA row.  

So could we have an extra checkbox, default checked, called `Frequencies`. 
And the Contrasts checkbox, by default is unchecked.  With the bug fixed, so it works - which it doesn't now.  It gives this error.

image

@shadrackkibet I have a feeling you could do this very quickly - and allocating it might actually take you longer, but up to you.  That's particularly because I think incorporating the summary dialogue stuff might just be the way to go, and you know that code very well.  However, @anastasia-mbithe is making excellent progress on the Contrasts dialogue, so perhaps she could undertake this one later?

We will then test the tidying process with the rrefine `lateformeeting` data, though that doesn't include missing values.  

The idea is that the tidying process for really messy factors (e.g. with 100s of levels) can then be done at the factor level.  So, the process will be: 
a) Make this factor data frame
b) Use Find-replace (and filter, etc) to tidy the levels.
c) Use the new simple merge dialogue to add it back to the original data (with the original factor variable as the key.)

This process with the lateformeeting data will probably be good for a video!

anastasia-mbithe:
This fixes #7229
Hello @rdstern and @shadrackkibet, The dialog now works. The replace checkbox is checked by default as well as the frequencies checkbox. 
When the Contrasts checkbox is unchecked, we have the levels and frequencies only, when checked we have the contrasts included. The same thing happens to the Frequencies checkbox, when its unchecked we don't have the frequencies in the output displayed.

### DIFF
--- a/instat/dlgFactorDataFrame.Designer.vb
+++ b/instat/dlgFactorDataFrame.Designer.vb
@@ -46,6 +46,7 @@ Partial Class dlgFactorDataFrame
         Me.ucrBase = New instat.ucrButtons()
         Me.ucrReceiverFactorDataFrame = New instat.ucrReceiverSingle()
         Me.ucrSelectorFactorDataFrame = New instat.ucrSelectorByDataFrameAddRemove()
+        Me.ucrChkFrequencies = New instat.ucrCheck()
         Me.SuspendLayout()
         '
         'lblSelectedFactor
@@ -129,12 +130,22 @@ Partial Class dlgFactorDataFrame
         Me.ucrSelectorFactorDataFrame.Size = New System.Drawing.Size(213, 183)
         Me.ucrSelectorFactorDataFrame.TabIndex = 0
         '
+        'ucrChkFrequencies
+        '
+        Me.ucrChkFrequencies.AutoSize = True
+        Me.ucrChkFrequencies.Checked = False
+        Me.ucrChkFrequencies.Location = New System.Drawing.Point(238, 176)
+        Me.ucrChkFrequencies.Name = "ucrChkFrequencies"
+        Me.ucrChkFrequencies.Size = New System.Drawing.Size(165, 23)
+        Me.ucrChkFrequencies.TabIndex = 8
+        '
         'dlgFactorDataFrame
         '
         Me.AutoScaleDimensions = New System.Drawing.SizeF(96.0!, 96.0!)
         Me.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Dpi
         Me.AutoSize = True
         Me.ClientSize = New System.Drawing.Size(428, 264)
+        Me.Controls.Add(Me.ucrChkFrequencies)
         Me.Controls.Add(Me.ucrChkAddCurrentContrasts)
         Me.Controls.Add(Me.ucrChkReplaceIfAlreadyExists)
         Me.Controls.Add(Me.ucrInputFactorNames)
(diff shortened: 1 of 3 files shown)


rdstern (changes requested):
@anastasia-mbithe that's excellent and almost working fine.  
The one issue is that when you have a factor data frame for a particular factor, then that's it.  It doesn't produce another one.  So - with the usual survey data, I get a factor data frame for the village - called village.  Then I change the name in your dialogue to give me a second one - perhaps with different options, called aaa.  It doesn't do this.  No error, just no data frame.  If I delete village, then it is fine again.  But I may want 2 different data frames (perhaps different contrasts) for the same factor.

shadrackkibet:
> @anastasia-mbithe that's excellent and almost working fine. The one issue is that when you have a factor data frame for a particular factor, then that's it. It doesn't produce another one. So - with the usual survey data, I get a factor data frame for the village - called village. Then I change the name in your dialogue to give me a second one - perhaps with different options, called aaa. It doesn't do this. No error, just no data frame. If I delete village, then it is fine again. But I may want 2 different data frames (perhaps different contrasts) for the same factor.

This will need some bits of change in the R code. We can look at it together if you would prefer.

anastasia-mbithe:
@rdstern Thank you for that. 
The dialog only replaces the previous data frame with the created one, instead of adding the new data frame in a new sheet. If you want the dialog to produce more than one data frame then, we can add that feature and do away with the replace option. From my end when you change the name of the data frame to something like aaaaa, with the Replace checkbox checked it does change.
Screenshot (82)
Screenshot (83)

However, the contrasts are similar because it uses the default function (control/treatment contrast). So, to get different contrasts for the same factor then maybe we add an option similar to Contrasts dialog to choose from.

anastasia-mbithe:
> @anastasia-mbithe that's excellent and almost working fine. The one issue is that when you have a factor data frame for a particular factor, then that's it. It doesn't produce another one. So - with the usual survey data, I get a factor data frame for the village - called village. Then I change the name in your dialogue to give me a second one - perhaps with different options, called aaa. It doesn't do this. No error, just no data frame. If I delete village, then it is fine again. But I may want 2 different data frames (perhaps different contrasts) for the same factor.

@rdstern This is fixed now.

rdstern (approved):
Great.  Very well done.

shadrackkibet (approved):

PR #7291, opened 2022-03-10
Added delete labels to the new recoded factor column in Recode Factor Dialogue

### LINKED ISSUE #7275: Sort out labelling in factor (and possibly in character?) variables
When I have a labelled factor variable I can edit the levels and labels with the right-click Levels/Labels (same as Prepare > Factor >Levels/Labels dialogue.  This runs the databook set_factor_levels function e.g. 

```
# The dialog, Labels/Levels
data_book$set_factor_levels(data_name="survey", col_name="fert2", new_labels=c("None4","0.5","1","1.5","2","2.5","Max4"), new_levels=c(0.2,0.5,1.0,1.5,2.0,2.5,3.5))
```
This works fine.  I can edit either the levels, or the labels, and this works fine.  I can also make the labels not unique.  But I can't make the levels not-unique.  If the factor isn't labelled, then non-unique labels will reduce the number of levels.  If it is labelled, then the number of levels stays the same, because they have different levels.  I am happy with all this.

But the `Prepare > Factor > Recode Factor` does (seemingly) roughly the same thing, except it gets in a (big) mess with labelled data.  It has 4 radio buttons at the top.  They don't know about labels, so the labelling doesn't change.  I suggest either our labelling function is done after its function, or we may delete labels first and then reinstate the new labels afterwards.  Then sometimes the levels will be NA and that's fine (I hope!).

I have checked the other factor functions briefly and I think they are ok!

And I think character operations lose the labelling so that's ok too.  

That's a relief.  I think a) the levels/labels is working ok, and there is just the one dialogue in the factors to correct. 

It is good that @shadrackkibet and @N-thony are together this week, and I hope this will turn out to be simple.

N-thony:
Fixes #7275 
@rdstern @shadrackkibet  as discussded on skype I have added the delete of labels to the new recoded column factor, please take a look. Thanks.

### DIFF
--- a/instat/dlgRecodeFactor.vb
+++ b/instat/dlgRecodeFactor.vb
@@ -156,8 +156,8 @@ Public Class dlgRecodeFactor
         clsRemoveLabelsFunction.AddParameter("property", Chr(34) & "labels" & Chr(34), iPosition:=2)
         clsRemoveLabelsFunction.AddParameter("new_val", Chr(34) & Chr(34), iPosition:=3)
 
-        clsAddLabelsFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$convert_column_to_type")
-        clsAddLabelsFunction.AddParameter("to_type", Chr(34) & "factor" & Chr(34), iPosition:=2)
+        'clsAddLabelsFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$convert_column_to_type")
+        'clsAddLabelsFunction.AddParameter("to_type", Chr(34) & "factor" & Chr(34), iPosition:=2)
 
         clsFctLumpMinFunction.SetPackageName("forcats")
         clsFctLumpMinFunction.SetRCommand("fct_lump_min")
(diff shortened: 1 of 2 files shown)


rdstern (approved):
@N-thony great - this works!  So I am approving.  
I did notice an inconvenience in the Recode (where you simply put the values in.
I give an example, which is to recode fert (7 levels) into None (0), Some (0.5 up to 2.5) Max (3).
a) When I do this, and return to the dialogue it has not remembered these new codes.
b) I would like to be able to copy down/paste into the new values.  I seem to have to enter them one-at a time.
This could be as a new issue if you prefer?  And it isn't that urgent.

N-thony:
> @N-thony great - this works! So I am approving. I did notice an inconvenience in the Recode (where you simply put the values in. I give an example, which is to recode fert (7 levels) into None (0), Some (0.5 up to 2.5) Max (3). a) When I do this, and return to the dialogue it has not remembered these new codes. b) I would like to be able to copy down/paste into the new values. I seem to have to enter them one-at a time. This could be as a new issue if you prefer? And it isn't that urgent.

@rdstern I have fixed item b), I think perhaps item a) to can be fixed in a separate PR. Please take a look. Thanks.

rdstern (approved):
Great

instat/dlgRecodeFactor.vb
```diff
@@ -21,7 +21,7 @@ Public Class dlgRecodeFactor
     Private bFirstLoad As Boolean = True
     Private clsPlyrRevalueFunction, clsReplaceFunction, clsFctOtherFunction, clsFctExplicitNaFunction As New RFunction
     Private clsFctLowFreqFunction, clsFctLumpPropFunction, clsFctLumpMinFunction, clsFctLumpNFunction As New RFunction
-
+    Private clsRemoveLabelsFunction, clsAddLabelsFunction As New RFunction
```
lloyddewit, line 24: remove `clsAddLabelsFunction `?

lloyddewit (changes requested):
@N-thony thanks, looks good, I just have one small comment

lloyddewit (approved):

rdstern (approved):

PR #7292, opened 2022-03-11
Updating the correlations dialog to use the newer corrr package

### LINKED ISSUE #7257: Update the correlations dialog to use the newer corrr package
A new `corrr` package is discussed on issue #6121. So here, I will outline potential ideas to edit in the "Multivariate > Correlations" dialog.

We currently have two options on the Correlation dialog - "Two Columns" and "Multiple Columns".

**I'll start with "Multiple Columns":**

"Multiple Columns" currently runs `cor`, which gives a correlation matrix between the specified variables.

I suggest we replace the function in the "Multiple Columns" tab with `correlate` from the `corrr` package. The parameters do not change, so should be a very simple change.

```
library(corrr)
data(mtcars)

# currently for a kendall method and missing as complete.obs:
cor(x = mtcars, method = "kendall", use = "complete.obs")

# with the new function in the corrr package
corrr::correlate(x = mtcars, method = "kendall", use = "complete.obs")
```

If we wanted, we can introduce other options.

1. Specify the value in the diagonal with a `diagonal` parameter in the function. E.g.
` corrr::correlate(x = mtcars, method = "kendall", use = "complete.obs", diagonal = "Hello")`
2. The `fashion` function gives a few extra options we could include. The user can then decide the number of decimal places, if they want the initial zero, or how to display NAs:
`cor_table <- corrr::correlate(x = mtcars, method = "kendall", use = "complete.obs", diagonal = "Hello")`
`fashion(cor_table, decimals = 2, leading_zeros = TRUE, na_print = "")`
3. Get just the triangular of the correlations with the `shave` function
`cor_table <- corrr::correlate(x = mtcars, method = "kendall", use = "complete.obs", diagonal = "Hello")`
`fashion(shave(cor_table))`

image

So we can add 
1. An option "Display on diagonal" with an input, default "NA" in. This is linked to the `diagonal` parameter.
2. An option "Decimal Places:" with a nud. This is linked to adding the `fashion` function and the `decimals` parameter
3. An option "Display NAs" with an input. This is linked to adding the `fashion` function and the `na_print` parameter
4. A checkbox "Leading zeros", default `TRUE`. If `FALSE`, this is linked to adding the `fashion` function and the `leading_zeros` parameter.
Option 2-4 might be less trivial because they add the same function. @rdstern do you think we want this?

**Onto "Two Columns"**

Currently we use `cor.test` which performs a correlation test between the two given variables. We can use the `correlate` function, but then lose the formalised test.

@rdstern presumably we want to keep the test in here? In which case, we can keep this option as `cor.test`.
If not, then all the functionality laid out above still work, but now in the `correlate` function, `x` is a variable, and a new `y` argument is used for the second variable.
`correlate(x = mtcars$mpg, y = mtcars$cyl)`.
I'll expand on this if you think we should use `correlate` here too.

**Sub dialog**
In the `corrr` package is `rplot`, but this does not seem to do anything that we do not already have in the "Correlation Plot" option.
In the output, we can just read in correlation object from the `corrr` package. E.g., for the "Correlation Plot" option:

```
last_model <-  corrr::correlate(x = mtcars, method = "kendall", use = "complete.obs")
GGally::ggcorr(last_model)
```

There is another plot but the output does not seem to work properly for me. I'm not sure why this is.

Vitalis95:
Fixes #7257 
@rdstern @shadrackkibet ,  this PR;

- Replaces the function in the "Multiple Columns" tab with correlate from the corrr package.
- Adds `diagonal` parameter 
- Adds `decimals ` parameter from `fashion` function
- Adds `na_print` parameter from `fashion` function
- Adds ` leading_zeros` parameter from `fashion` function 
- Adds `shave` function
- Saves the correlations in a new data frame

### DIFF
--- a/instat/dlgCorrelation.designer.vb
+++ b/instat/dlgCorrelation.designer.vb
@@ -38,35 +38,42 @@ Partial Class dlgCorrelation
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()>
     Private Sub InitializeComponent()
+        Me.components = New System.ComponentModel.Container()
         Me.lblFirstColumn = New System.Windows.Forms.Label()
         Me.lblSecondColumn = New System.Windows.Forms.Label()
         Me.grpMethod = New System.Windows.Forms.GroupBox()
         Me.rdoKendall = New System.Windows.Forms.RadioButton()
         Me.rdoPearson = New System.Windows.Forms.RadioButton()
         Me.rdoSpearman = New System.Windows.Forms.RadioButton()
+        Me.ucrPnlMethod = New instat.UcrPanel()
         Me.cmdOptions = New System.Windows.Forms.Button()
         Me.grpMissing = New System.Windows.Forms.GroupBox()
         Me.rdoCompleteRowsOnly = New System.Windows.Forms.RadioButton()
         Me.rdoPairwise = New System.Windows.Forms.RadioButton()
+        Me.ucrPnlCompletePairwise = New instat.UcrPanel()
         Me.lblConfInterval = New System.Windows.Forms.Label()
         Me.rdoTwoColumns = New System.Windows.Forms.RadioButton()
         Me.rdoMultipleColumns = New System.Windows.Forms.RadioButton()
         Me.lblSelectedVariables = New System.Windows.Forms.Label()
         Me.lblDisplayOnDiagonal = New System.Windows.Forms.Label()
+        Me.lblDecimalPlaces = New System.Windows.Forms.Label()
+        Me.lblDisplayNas = New System.Windows.Forms.Label()
+        Me.ttDisplayOnDiagonal = New System.Windows.Forms.ToolTip(Me.components)
+        Me.ucrInputDisplayNas = New instat.ucrInputComboBox()
+        Me.ucrChkLeadingZeros = New instat.ucrCheck()
+        Me.ucrSaveDataFrame = New instat.ucrSave()
+        Me.ucrNudDecimalPlaces = New instat.ucrNud()
         Me.ucrNudConfidenceInterval = New instat.ucrNud()
         Me.ucrChkCorrelationMatrix = New instat.ucrCheck()
         Me.ucrPnlColumns = New instat.UcrPanel()
-        Me.ucrPnlMethod = New instat.UcrPanel()
         Me.ucrReceiverMultipleColumns = New instat.ucrReceiverMultiple()
-        Me.ucrPnlCompletePairwise = New instat.UcrPanel()
         Me.ucrReceiverSecondColumn = New instat.ucrReceiverSingle()
         Me.ucrReceiverFirstColumn = New instat.ucrReceiverSingle()
         Me.ucrSelectorCorrelation = New instat.ucrSelectorByDataFrameAddRemove()
         Me.ucrBase = New instat.ucrButtons()
         Me.ucrSaveModel = New instat.ucrSave()
         Me.ucrInputDiagonal = New instat.ucrInputTextBox()
-        Me.ucrNudDecimalPlaces = New instat.ucrNud()
-        Me.lblDecimalPlaces = New System.Windows.Forms.Label()
+        Me.ucrChkShave = New instat.ucrCheck()
         Me.grpMethod.SuspendLayout()
         Me.grpMissing.SuspendLayout()
         Me.SuspendLayout()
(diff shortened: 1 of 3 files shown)


rdstern (approved):
@Vitalis95 well done.  This already looks good.  
a) The default option should probably be NA (proper NA) on the diagonal.  Blank should also be an option.
b) When it writes the data frame than the main columns should be numeric.  Currently they are all character, though not indicated.  That's probably because of the blank on the diagonal.  The first column (with the names can be character - and hopefully has the (C) if so.  If shaved and saved, then the other half of the matrix needs to be filled with NA.
c) If reasonably easy, then can you add the rearrange checkbox.  There will be a Method pull-down, with PCA as default and a checkbox for Absolute.  Default is unchecked, so gives absolute=FALSE.

rdstern (changes requested):
@Vitalis95 well done.  This already looks good.
a) The default option should probably be NA (proper NA) on the diagonal.  Blank should also be an option. b) When it writes the data frame than the main columns should be numeric.  Currently they are all character, though not indicated.  That's probably because of the blank on the diagonal.  The first column (with the names can be character - and hopefully has the (C) if so.  If shaved and saved, then the other half of the matrix needs to be filled with NA. c) If reasonably easy, then can you add the rearrange checkbox.  There will be a Method pull-down, with PCA as default and a checkbox for Absolute.  Default is unchecked, so gives absolute=FALSE.

rdstern:
@Vitalis95 this is starting to look really good.  I have been trying with the mtcars data from the datasets package - default in the library.
a) Is there a reason why clicking on the shave or rearrange checkbox seems to refresh the data selector?  The Leading zeros and result name don't seem to do that.
b) I'm not good on design - Antoine is better, so you may want to check with him.  But you currently have the Rearrange checkbox on the right.  I wonder about moving it to the left and swapping it with Shave.  Then the 3 controls can all go on a single line.  And you may want to move leading zeros to the right, and on the same row as decimal places - they sort of go together.
c) And could you also centre the top buttons.
d) Additionally - if easy - the same dialogue will now be in 2 places, namely 2/3 variables and Multivariate.  Currently the default is 2 variables.  Ideally I'd like to keep that for the 2/3 variables and have Multiple Columns as the default in Multivariate.  If that is tricky to arrange, then have Multiple Columns first in the top buttons, so it is always the default.
e) Now on the saving.  This is working well except for 2 features:
1) It is saving each variable as a character variable, but without putting the (C) there.
2) It is not saving NA as missing, but the 2 characters NA instead.
I don't know how flexible you can be.  But when the diagonal is NA (proper NA, i.e. missing) and you do not shave, then the matrix should be numeric.
Possibly otherwise, i.e. shaved, or diagonal as blank, then the whole matrix should be character - and with the names having (C) to confirm that.
You may find you are better able to save as a matrix, and then add the as.data.frame.  This might make the first column into the row labels.  So be it if that is what happens.
f) You may stop there for now.  But I am interested in the graphic option from corrr as well.  One reason could be that this could be the same code as producing the graph.  Currently the plotting commands are separate.  And currently this means that if the rearrange works, then it does not apply to the graphs.  We may want at least one option where it does?  (If that is a bit difficult, then we could merge first, and add this as a separate issue afterwards.)

Vitalis95:
@rdstern  I have sorted out the selector issue and also setting multiple columns as default
However, for the saving character issue, the `fashion` function with has the `leading_zero, na_print and decimals` parameter tend to convert the  correlation data frame into a noquote matrix (Character strings without quotes) with the correlations cleanly formatted  and the diagonal (or any NA) left blank.
If I run the correlation without `fashion` function the output generated is numeric matrix with the first column with names being a character as shown below:
The `fashion` function is for pretty printing, can we drop it because it converts the correlation data frame into a character string
correlate
Have a look at it, thanks.

rdstern (changes requested):
@Vitalis95 good progress!
I tried with the mtcars data from the standard datasetsd package.
a) I now get many decimal places, both in the output window and also in the grid, when I save.  It used to obey the decimals option.
b) Sorry, but I did warn you, I was no good on design.  Now, when I look at the dialogue it seems a bit overwhelming!  I think the disabled Matrix control can go.  Then could you try with a checkbox with label Display Options.  Default unticked, uses the default displays from the dialogue.  If ticked, then the 3 rows of countros are in a group box, which becomes visible.  I think that's everything, because rearrange is also a sort of display option!

Vitalis95:
@rdstern  I have reorganized the controls in the dialogue and created a group box for the rearrange option. Also I have sorted a reset issue I found in the Result Name checkbox. Have a look it if you are ok with that.
About the output generated, the `fashion` function convert the output into characters and alter the missing values, the missing values `NA` appear empty (`" "`) if Display NA checkbox is blank or any other value specified here.
Here is a `corrr::fashion` article; 
https://drsimonj.svbtle.com/fashion-your-correlations-with-corrr
Thanks

rdstern:
@Vitalis95 sorry I didn't make the previous message clear.  I prefer the rearrange on a single line, as it was.  And there are many more display options in the dialogue.  What I said above was:
"Then could you try with a checkbox with label Display Options. Default unticked, uses the default displays from the dialogue. If ticked, then the 3 rows of countrols are in a group box, which becomes visible. I think that's everything, because rearrange is also a sort of display option!"

So the checkbox is new, with label `Display Options`.  Default is unchecked.  Then Display on diagonal, down to Shave and also rearrange are all in the group box that opens when you check for the rich set of Display Options.

N-thony:
> @N-thony I checked and found that this order `...rearrange(x = shave(x=correlate...)` doesn't work, it produces an error, but I have changed it to `...shave(x = rearrange(x=correlate...)` and it works fine. @rdstern meanwhile, have a look at it as @N-thony check on the issue of data frame being displayed on the output window whenever you run.

@Vitalis95 remove lines 53,255,269,246 and then try.

Vitalis95:
@rdstern I am closing this branch because it has conflicts due to database translation

PR #7300, opened 2022-03-14
Corrected the convert in Numeric Popup dialog

### LINKED ISSUE #7293: Correct the convert to numeric popup dialogue
I hope @N-thony can do this, but it may need input from @shadrackkibet as well as a check that my suggestions below are correct.

When a factor (possibly also a labelled character) is converted to numeric, then there is usually a popup if the factor contains a mix of numeric and non-numeric values.  But if all is non-numeric, and it is unlabelled, then R-Instat currently simply uses the ordinal values, and makes them into labels!
To illustrate us the fert data from survey.  Make it into a factor.
a) Then recode into fert_recoded by just making 0 into None and 3 into Max.  The recoding doesn't have labels now - which is fine!
b) Also recode by the above, but make all the other levels into Some.  (So nothing is numeric in fert_recoded1)
Note that neither of them is now labelled.

Now right-click and make a) into Numeric.  The popup allows the option of the Labelled convert.  That is even though there are now no labels!  If you accept, then it adds labels again - which is fine - except it may not be what is wanted.  It does what I would call "Ordinal Convert".  I suggest we simply change Labelled Convert to `Ordinal Convert `in the popup.  Maybe the word Ordinal is in a Colour, so it stands out.
When nothing is numeric, and you convert to numeric it misses the check and uses the ordinals and makes it numeric.  This should be trapped and the same dialogue given.  It should again say Ordinal Convert.

N-thony:
Fixes #7293
I have added the checking when nothing is a numeric in the column to convert to numeric and also changed `Labelled Convert` to `Ordinal Convert`

### DIFF
--- a/instat/Model/RCommand/clsPrepareFunctionsForGrids.vb
+++ b/instat/Model/RCommand/clsPrepareFunctionsForGrids.vb
@@ -150,6 +150,46 @@ Public Class clsPrepareFunctionsForGrids
         End If
         _RLink.RunScript(clsConvertToNumeric.ToScript(), strComment:="Right click menu: Convert Column(s) To Numeric")
     End Sub
+
+    ''' <summary>
+    '''  
+    ''' </summary>
+
+    Public Function GetColumnLabels(strColumnName As String) As Boolean
+        Dim bLabels As Boolean
+        Dim clsColmnLabelsRFunction = New RFunction
+        Dim expItems As SymbolicExpression
+        Dim clsGetColumnsFromData As New RFunction
+
+
+        clsGetColumnsFromData.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_columns_from_data")
+        clsGetColumnsFromData.AddParameter("data_name", Chr(34) & _strDataFrame & Chr(34), iPosition:=0)
+        clsGetColumnsFromData.AddParameter("col_names", Chr(34) & strColumnName & Chr(34), iPosition:=1)
+        clsGetColumnsFromData.AddParameter("use_current_filter", "FALSE", iPosition:=2)
+
+        clsColmnLabelsRFunction.SetRCommand("get_column_attributes")
+        clsColmnLabelsRFunction.AddParameter("x", clsRFunctionParameter:=clsGetColumnsFromData, iPosition:=1)
+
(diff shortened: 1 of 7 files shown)


rdstern (changes requested):
@N-thony almost I think.  As discussed use Labelled convert - as before - when labelled, and ordinal convert when there are no labels.

rdstern (approved):
Looks good now

instat/frmConvertToNumeric.vb
```diff
@@ -30,6 +30,16 @@
         strDataFrameName = strDataFrame
     End Sub
 
+    Public Sub CheckLabels(bCheck As Boolean)
+        If bCheck Then
+            cmdLabelledConvert.Text = "Lebelled Convert"
```
lloyddewit, line 35:
```suggestion
            cmdLabelledConvert.Text = "Labelled Convert"
```

instat/frmConvertToNumeric.vb
```diff
@@ -30,6 +30,16 @@
...
+            lblOrdinal.Visible = False
```
lloyddewit, line 36: Move this out of the `if else` block and set it to `lblOrdinal.Visible = Not bCheck`?

instat/Model/RCommand/clsPrepareFunctionsForGrids.vb
```diff
@@ -150,6 +150,27 @@ Public Class clsPrepareFunctionsForGrids
...
     End Sub
+
+    ''' <summary>
+    '''  Check if the column factor contains labels.
+    ''' </summary>
+    Public Function CheckHasLabels(strColumnName As String) As Boolean
+        Dim clsColmnLabelsRFunction = New RFunction
+        Dim clsGetColumnsFromData As New RFunction
+
+
```
lloyddewit, line 161:
```suggestion
```

lloyddewit (changes requested):
@N-thony thanks, looks good, just a couple of small questions

N-thony:
@lloyddewit I resolved your comments. Please take a look. Thanks.

rdstern (approved):
Still looks good to me

lloyddewit (approved):

Issue #7297, opened 2022-03-13
Changes in menu items


rdstern:
I list changes - some new - in frm main.  I assume as agreed earlier, that @shadrackkibet will handle, or supervise this, so the results can be merged quickly.

a) In the Prepare > Check Data menu:
1) Change Duplicates... to Duplicate Rows... 
2) Make Import from Open Refine not visible.  (We may re-instate it later.
3) And just a pointer, that we will (soon?) add Statistical Data Disclosure, using the sdcMicro package.  I am pleased that this package is currently installed, so let's keep it.  It is important and a major undertaking.  It also has its own menu-driven front-end so we may just link to that.  That would be really nice!

b) In Describe > Two Variables
1) Change Describe > Two Variables to Describe > Two/Three Variables
2) Add Correlations just below the horizontal line there.  Add a copy of the Describe > Multivariate > Correlations dialogue.
3) Delete Describe > Three Variables

shadrackkibet:
@Vitalis95 include the following in your PR. Consistency is key!

Climatic>Tidy and Examine>Duplicate Rows...
Options by Context > Check data> Duplicate Rows...


Change Filter to Filter Rows... in all right clicks
The same applies to Column selection in all right clicks

rdstern:
@shadrackkibet great that your comments overlapped with mine.  a) Good to add your consistency items.
b) Pleased you liked my addition for Filter to `Filter Rows...`
c) Do you want to keep `Column Selection...`, rather than `Select Columns...`  I (only) slightly prefer `Select Columns...`, but happy to go with your preference.

shadrackkibet:
@rdstern I like **Select Columns...** let us change it. 

We need to change dialog names at the top but that can be done later.

rdstern:
@shadrackkibet great.  It shows it is a bit useful writing the R-Instat help, alongside the programming.  And I like calling "your" new objects `Selects  `, just as we have `Filters`!

N-thony:
> I list changes - some new - in frm main. I assume as agreed earlier, that @shadrackkibet will handle, or supervise this, so the results can be merged quickly.
> 
> a) In the Prepare > Check Data menu:
> 
> 1. Change Duplicates... to Duplicate Rows...
> 2. Make Import from Open Refine not visible.  (We may re-instate it later.
[...]

PR #7309, opened 2022-03-15
Updated Merge Additional Data dialog

### LINKED ISSUE #7029: Update Merge Additional Data dialog

N-thony:
Fixes #7029 
In this PR I have made the following changes:

- Changed the texts the dialogue and menu also updated the database from `Merge Additional Data dialog` to `Add (Merge) Data`
- Also added two possilities of saving the data either in the current dataframe or the new dataframe
I had a discussion with @Patowhiz where we decided on the use of `add_columns_to_data` for the option of saving the merged dataframe to the current dataframe.
I also made change in the database requested in #7297 items a.1) b.1) c.2)

### DIFF
--- a/instat/dlgMergeAdditionalData.Designer.vb
+++ b/instat/dlgMergeAdditionalData.Designer.vb
@@ -24,19 +24,19 @@ Partial Class dlgMergeAdditionalData
     Private Sub InitializeComponent()
         Me.lblVariablesToIncludeSecond = New System.Windows.Forms.Label()
         Me.cmdModify = New System.Windows.Forms.Button()
+        Me.lblMergeBy = New System.Windows.Forms.Label()
         Me.ucrInputMergingBy = New instat.ucrInputTextBox()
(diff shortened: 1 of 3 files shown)


lloyddewit:

rdstern (changes requested):
@N-thony excellent.  I assume it will later be in the Prepare menu.  Meantime I have checked it in the Options by Context menu.
I added the lat and long in the Niger data.  The one item to change is that it included the name variable by default.  I took it out, because it is not one I am adding (it is already there!).
I realise it needs to be included in the command.  But, I would like the code to include it, if it isn't added by the user.

Ivanluv (approved):

N-thony:
@rdstern I have a conflict here that I cannot resolve, it is coming from the database change. @Vitalis95 was happy to take over all the tasks concerning changing the text on the control and also in the database.
I will close this one and open another one.

PR #7312, opened 2022-03-16
Corrected the error on the check values type in Non Numeric Dialogue

### LINKED ISSUE #7310: Confirm and correct error in Non-numeric dialogue
We are now just about complete in our use of factors and labels in R-Instat.  This is making it easy to transform data between character, factor and numeric.
When you make a factor (or character?) into a numeric variable and it has some non-numeric values, then it invites you to check, and that loads the Prepare > Check Data > Non-Numeric Values dialogue.
a) Now that dialogue currently only accepts character variables, but it accepts a factor from this route.  So my first suggestion is that it should accept factors generally as well as character variables.
b) Here is an example where it seems to me to get the answer wrong for a factor.  And it is different for the factor and the character variable.  
1. With the usual survey data, make fert into a factor.  Then, with Levels/Labels make 0 into None and 3 into Max. 
2. Now try to make it numeric and it notes that 12 values are not-numeric.
3. Click to check, and it goes to the Nonum dialogue.
4. Accept and the result is FALSE for all rows!  That's wrong - it just said 12 values are not numeric!
5. Make fert into character
6. Try nonum again - and it gives the correct answer.  12 values are non-numeric.

@shadrackkibet please confirm that I am right here.  This anomaly needs to be corrected.  
Assuming so, please can @N-thony make the correction.  @shadrackkibet may give the R-code, but it should be evident, because it is detected in the popup when I tried to make the factor numeric (but that code isn't in the results window.)

N-thony:
Fixes #7310 
I made the correction of the R code as provided in the corresponding issue.

### DIFF
--- a/instat/dlgFindNonnumericValues.vb
+++ b/instat/dlgFindNonnumericValues.vb
@@ -22,7 +22,7 @@ Public Class dlgFindNonnumericValues
     Private bUseSelectedColumn As Boolean = False
     Private strSelectedColumn As String = ""
     Private strSelectedDataFrame As String = ""
-    Private clsIsNaFunction, clsIsNaNumericFunction, clsAsNumericFunction, clsSummaryFunction As New RFunction
+    Private clsIsNaFunction, clsIsNaNumericFunction, clsAsNumericFunction, clsAsCharacterFunction, clsSummaryFunction As New RFunction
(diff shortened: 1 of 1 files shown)


lloyddewit (approved):

lloyddewit:

rdstern (approved):

PR #7328, opened 2022-03-24
Export dialog enhancement

### LINKED ISSUE #7286: File > Export > Export Dataset UI issues
**Steps to reproduce**
- File > Export > Export Dataset
- Pick destination file ('Save As' field)
- Select data frame(s) to export
- After selecting data frame, the filename is removed from the destination ('Save As') field.

**Expected behaviour**
Filename should remain there.

image

**Details:**
 - OS: Ubuntu 18.04 (docker-version)
 - R-Instat Version: Commit [cdddbdf](https://github.com/IDEMSInternational/r-instat-docker/commit/cdddbdf665956dbd88505b8e093b26619b844413) from https://github.com/IDEMSInternational/r-instat-docker
Also occurs on Windows, version 0.7.4

Patowhiz:
Fixes #7286.


instat/dlgExportDataset.vb
```diff
@@ -175,43 +167,61 @@ Public Class dlgExportDataset
...
-        cboFileType.SelectedItem = strPrevSelectedFileType
-        If String.IsNullOrEmpty(cboFileType.SelectedItem) Then
+        If Not String.IsNullOrEmpty(strPrevSelectedFileType) AndAlso cboFileType.Items.Contains(strPrevSelectedFileType) Then
+            cboFileType.SelectedItem = strPrevSelectedFileType
+        Else
             cboFileType.SelectedIndex = 0
         End If
 
+        'if file save state has changed then make path changes to the file path control
+        If bSaveAsMultipleFiles <> ucrFilePath.FolderBrowse Then
```
lloyddewit, line 183: I understand that `bSaveAsMultipleFiles` is about saving to multiple files, and `FolderBrowse` is about whether to browse a dialog or save a file.
So I assume that `bSaveAsMultipleFiles` is only relevent when `FolderBrowse` is False.

I'm a bit confused. What is happening on lines 183 and 184?

instat/dlgExportDataset.vb
```diff
@@ -175,43 +167,61 @@ Public Class dlgExportDataset
...
+            ucrFilePath.FolderBrowse = bSaveAsMultipleFiles
+            If Not ucrFilePath.IsEmpty Then
```
lloyddewit, line 185: why do we only set `ucrFilePath.FilePath` if its not empty?
why shouldn't we set an empty `ucrFilePath.FilePath`?

instat/ucrFilePath.vb
```diff
@@ -82,17 +83,38 @@ Public Class ucrFilePath
...
             'its safer To, just incase number of filters set are different from previously set
             SelectedFileFilterIndex = 1
+
+            If String.IsNullOrEmpty(strFilePathDialogFilter) Then
+                Exit Property
+            End If
+
+            'for file browse, if file path name was already set,
+            'change the extension set to use the new one.
+            If Not FolderBrowse AndAlso Not String.IsNullOrEmpty(FilePath) Then
```
lloyddewit, line 107: add these conditions to the `If` statement on line 101?

instat/ucrFilePath.vb
```diff
@@ -117,28 +149,58 @@ Public Class ucrFilePath
...
             'first replace backward slashed with forward slashes cause of R path formats
             PathControl().SetName(value.Replace("\", "/"))
-            If value <> "" Then
-                'to be safe, always replace the forward slashes with back slashes. values will be used by vb.net sialog prompts
-                PreviousSelectedWindowsFilePath = value.Replace("/", "\")
-            End If
+            'If value <> "" Then
+            '    'to be safe, always replace the forward slashes with back slashes. values will be used by vb.net sialog prompts
+            '    PreviousSelectedWindowsFilePath = value.Replace("/", "\")
+            'End If
```
lloyddewit, line 158: delete?

lloyddewit (commented):
@Patowhiz thanks, just a couple of questions

instat/dlgExportDataset.vb
Patowhiz: @lloyddewit thanks for asking.

The file path is supposed to change depending on whether the user is saving files into a single file or as multiple files (and this is determined by the dialog options; number of data frames and a checkbox). 

That check is meant to check if the user has changed that option(as multiple files or as a single file, by comparing the dialogs option and what was already set in the file path control). 

This check and the code block under it is meant to retain the file path that the user set. 

So for instance if I had browsed to `C:\Users\LENOVO\Desktop\benchmark.xlsx` then I select multiple data frames for saving into a single file, the path shouldn't change. But what happens if in the dialog I set it to save the data frames as multiple files. The control should only display the path to the directory, in this case `C:\Users\LENOVO\Desktop`. Consequently if I change my mind and decide to save the multiple data frames into a single file, then it should revert back to `C:\Users\LENOVO\Desktop\benchmark.xlsx`.  

I hope that is a bit clear?

instat/dlgExportDataset.vb
Patowhiz: @lloyddewit thanks for asking.

Because of line 189. That line will display a wrong file path. And also conceptually speaking there is no point in trying to retain correct selected file paths if there was no file path set.

instat/ucrFilePath.vb
Patowhiz: @lloyddewit this meant to make the control align itself with the extensions(in this case file filters) set. Also helps with preventing developer logical errors.

For instance, if I had previous given this control `.txt` and `.xlsx` extensions. The selected a file save name patrick.txt. What should happen if later I change the list of extensions to `.tab` and `.csv` only? The file save name should change to patrick.tab right? Since retaining the .txt is a now a logical error if a user tries to change the save name and extension.

And also these file extensions checks only matter if the control is not in file (NOT folder) browse mode. hence that check.

instat/ucrFilePath.vb
Patowhiz: Done

rdstern:
@Patowhiz and seems to work.

The only thing I noticed was the overwriting message, shown below:
image

I think it was true initially, so I changed the file name, but the message always stays.  I don't think it is true in the example shown?

Patowhiz:
@rdstern because of the dialog design, which was 
 suggested by David. I thought it is paramount that we tell the user at any given time that files will be overwritten. The alternative of course is to consistently check if there's a similar file name and omit that warning if not. Do you prefer that?

rdstern:
@Patowhiz no I like the message to warn that a file is to be overwritten.  But when I then change the file name to a new one that is unique, I found that the message seemed to remain.  So, once it was there - it seems to be there forever.  Maybe it does check each time the user changes the file name, but I don't think so.
I like the way that when I change the file type, then it automatically changes the extension.  That's great.  But does it also check then whether that new name is to be overwritten?

Patowhiz:
@rdstern no it doesn't. But that could be implemented if it's important.

instat/ucrFilePath.vb
lloyddewit: Thanks for the explanation. I agree with the checks and their rationale. In my comment above, I was only suggesting to change line 101 to
```
            If String.IsNullOrEmpty(strFilePathDialogFilter) _
                    OrElse FolderBrowse _
                    OrElse String.IsNullOrEmpty(FilePath) Then
```
We could then delete lines 107 and line 116; and it's less indentation which makes the code more readable (in my view). 
It's an optional change. If you prefer to leave as is, then that's also fine

instat/dlgExportDataset.vb
lloyddewit: Thanks, understood

instat/dlgExportDataset.vb
lloyddewit: Thank you, now it's clear to me.
Maybe I'm not the only person who would find the check confusing (or maybe I am!).
Should we expand the comment? e.g.
```
        'If we are saving to multiple files, then path should be set to a folder.
        'If we are saving to a single file, then path should be set to a file.
        'Therefore `bSaveAsMultipleFiles` and `ucrFilePath.FolderBrowse` should always be equal.
        'If not, then we need to make the `ucrFilePath` control consistent with the state of `bSaveAsMultipleFiles`.
```

Patowhiz: Agreed. Couldn't be any better! Copied and pasted. Done.

rdstern (approved):
ok, approved then

lloyddewit (approved):

PR #7329, opened 2022-03-25
Added a feature to convert year from 2-digit year to 4-digit in the Climatic> Prepare> Conversions dialog

### LINKED ISSUE #7254: Convert 2 digit year to 4 digits
This would be useful to add in the (Climatic) Conversions dialog. It's fairly simple to do but not completely trivial for users to know how to do in the calculator.

It just needs a one line function. There should be an option of a base year to know what the year is i.e. if the base year is 1930 then anything above 30 has 19 added to the front of it and anything below has 20 added.

The reverse is less commonly needed and already easy in the Transform Text dialog so I don't think we need that also.

anastasia-mbithe:
This fixes #7254

@rdstern @N-thony, I have added a checkbox under the Units top button, such that when you have a two digit year column i.e 01, 23, 45; the function converts the two digit to four digit year column with a base year of 30. Years less than 30, 2000 is added to them and those greater than 30 1900 is added.
@rdstern you can confirm whether the checkbox in the design is in the right position.


anastasia-mbithe:
Made some minor changes to the function name

Vitalis95 (changes requested):
image
Remove line 192 if you don't need it

shadrackkibet:
@Vitalis95 have you completed your review? I also expect you to test this PR to confirm if it fixes the issue.

Vitalis95 (approved):
It looks fine

instat/dlgConversions.vb
```diff
@@ -237,18 +243,22 @@ Public Class dlgConversions
...
+        clsYearConversionFunction.SetRCommand("convert_yy_to_yyyy")
+
         clsPrecipitationFunction.SetAssignTo(ucrSaveConversions.GetText, strTempDataframe:=ucrSelectorConversions.ucrAvailableDataFrames.cboAvailableDataFrames.Text, strTempColumn:=ucrSaveConversions.GetText, bAssignToIsPrefix:=True)
         ucrBase.clsRsyntax.SetBaseRFunction(clsPrecipitationFunction)
     End Sub
 
     Private Sub SetRCodeforControls(bReset As Boolean)
         ucrReceiverElement.AddAdditionalCodeParameterPair(clsTemperatureFunction, (New RParameter("temperature", 0)), iAdditionalPairNo:=1)
         ucrReceiverElement.AddAdditionalCodeParameterPair(clsWindSpeedFunction, (New RParameter("wind_speed", 0)), iAdditionalPairNo:=2)
+        ucrReceiverElement.AddAdditionalCodeParameterPair(clsYearConversionFunction, (New RParameter("x", 0)), iAdditionalPairNo:=3)
```
lloyddewit, line 255: Currently, if someone runs this code after 2030, then all dates after 2030 will be converted to 1931, 1932 etc.
Even if R-Instat is replaced, then there is a big chance that this code will be reused in the replacement for R-Instat, so I think we should make it future proof.
Would it be possible to get the current date, and then pass the current year as the 'base'?

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -1235,6 +1235,10 @@ convert_to_dec_deg <- function (dd, mm = 0 , ss = 0, dir) {
   return(decdeg)
 }
 
+convert_yy_to_yyyy <- function (x, base=2030) {
```
lloyddewit, line 1238: I find it dangerous to hard-code the year (see comment above). Remove the default '2030' and make this a mandatory parameter?

lloyddewit (changes requested):

anastasia-mbithe:
@lloyddewit, I had a discussion with @rdstern and @N-thony on this and we resolved to change the design. So, there's a new top button added (Year) with a receiver and a up-down control that allows one to set the base year. It has a range of 2000-2100 with the default set at 2030.
You can test it.

rdstern (changes requested):
Asked for additional control to be added

lloyddewit:
> @lloyddewit, I had a discussion with @rdstern and @N-thony on this and we resolved to change the design. So, there's a new top button added (Year) with a receiver and a up-down control that allows one to set the base year. It has a range of 2000-2100 with the default set at 2030. You can test it.

Thank you for the explanation, I resolved my comments above. When @rdstern approves, then I'll review/approve again.

rdstern (approved):

lloyddewit (approved):

PR #7333, opened 2022-03-28
Updated Merge Additional Data dialog

### LINKED ISSUE #7029: Update Merge Additional Data dialog

N-thony:
Fixes #7029
This PR replaces PR #7309 due to the conflicts.
@rdstern I made the changes here from your review in PR #7309 https://github.com/africanmathsinitiative/R-Instat/pull/7309#pullrequestreview-912676623, Please take a look.

### DIFF
--- a/instat/dlgMergeAdditionalData.Designer.vb
+++ b/instat/dlgMergeAdditionalData.Designer.vb
@@ -161,7 +161,7 @@ Partial Class dlgMergeAdditionalData
         Me.Name = "dlgMergeAdditionalData"
         Me.ShowIcon = False
         Me.StartPosition = System.Windows.Forms.FormStartPosition.CenterScreen
-        Me.Text = "Merge Additional Data"
+        Me.Text = "Add (Merge) Data"
         Me.ResumeLayout(False)
         Me.PerformLayout()
 
--- a/instat/dlgMergeAdditionalData.Designer.vb
+++ b/instat/dlgMergeAdditionalData.Designer.vb
@@ -22,33 +22,54 @@ Partial Class dlgMergeAdditionalData
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()> _
     Private Sub InitializeComponent()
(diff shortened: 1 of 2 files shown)


rdstern (changes requested):
Looks nice, but:
I used the survey data (2 copies) and deleted 3 variables from one of them.  Then made a row variable in each data frame, called row.  Then tried the merge
a) There was a variable of the same name in each data frame.  I wonder why it didn't find it.  I was easily able to add it though.
b) I could not get Ok to be enabled.  So could not check if it works.

N-thony:
> Looks nice, but: I used the survey data (2 copies) and deleted 3 variables from one of them. Then made a row variable in each data frame, called row. Then tried the merge a) There was a variable of the same name in each data frame. I wonder why it didn't find it. I was easily able to add it though. b) I could not get Ok to be enabled. So could not check if it works.

@rdstern please could you re-test? For item a) I need a support from @dannyparsons on checking the link or updating the link when we add the same column in the two dataframes.

N-thony:
> 

@rdstern could you try this for item a)?
1. Load survey dataset two times you will have `survey` and `survey1`,
2. and then from `survey` delete some columns that you will like to add back from `survey1`,
3. go to `Prepare > Data Frame > Row Numbers/Names` then add row in both datasets,
4. use `Prepare > Keys and Links > Add Key` to define `row as key` for both datasets,
5. use `Prepare > Keys and Links > Add Link` to define `row as link` between the two datasets
6. then go to Options by `Context > Prepare > Merge Additional Data...` it will automatically detect `row` as link between the two datasets.
**Note**. it was not detecting row when you did before because row was now defined first `key` in both datasets and the as `link` between the two datasets

rdstern:
Discussed why the making of key variables above should not be necessary.  The code to detect linking should be identical to what is used in the general merge dialogue.  That compares on names.

N-thony:
> Discussed why the making of key variables above should not be necessary. The code to detect linking should be identical to what is used in the general merge dialogue. That compares on names.

@rdstern of course, I got the point now. Thanks.

PR #7338, opened 2022-03-29
Calculations to give NA if an incompatible summary statistic is chosen

lilyclements:
As discussed in #7123 - the summaries were giving an error if incompatible summary statistics were selected. For example, a sum for a factor.

This PR fixes this error.

The summary will be NA if the variable is a date or factor and the summary statistic is a sum or mean.

@rdstern what other data types or summary statistics do we want to give NA?

### DIFF
--- a/instat/static/InstatObject/R/Backend_Components/calculations.R
+++ b/instat/static/InstatObject/R/Backend_Components/calculations.R
@@ -1,41 +1,41 @@
 calculation <- R6::R6Class("calculation",
-                         public = list(
-                           initialize = function(function_name = "", parameters = list(), 
-                                                 calculated_from = c(), is_recalculable = TRUE,
-                                                 sub_calculations = list(), type = "", filter_conditions = list(),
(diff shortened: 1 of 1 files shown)


rdstern:
@lilyclements that is much better - great even!  It is a releif too.  I would be ok to approve with this.
If I am being fussy, then the following may be interesting:
a) `range` doesn't work for ordinal factors  - there is a range function, and I think that should be ok and wonder if we use it?
b) I was pleased - that dates do give more summaries.  So mean, max and min are all ok for a date variable.  But we have plenty of percentiles.  So I tried p25 and p75 and they don't work.  The ordinary default summary is ok - and it gives quartiles.  It would be a luxury for us, so don't spend too long, but I wonder why we don't?
c) On the second sheet I also tried the proportion > 0 (as a percentage) and this is poorly labelled and missing for all variables. 
d) I also made a variable into a logical column and that works fine.

Issue #7205, opened 2022-02-07
Three small improvements in File > Import from Library


rdstern:
@shadrackkibet these are pretty trivial - well 2 are, and the 3rd is optional, and only if you agree?
1. The menu item is now Import from Library.  So please change the title in the dialogue from `Open Dataset from Library` to` Import from Library`. 
2. There is a second Help button on the right of the dialogue, just below the list of files.  Change this from `Help` to `R Help`, so it is distinguished from the main Help button.  Alternatively it could be` Info`, but I think `R Help` will be fine.
3. I find I often do a double-click on a data set and expect it to work.  Now that is unusual in that we don't have any other dialogues that use double-click to implement Ok.  We do have double-click in the data selector to copy a selected variable to the receiver, so it is consistent with that.  I would certainly find it convenient.    Danny is Ok with the idea - at least not against it!

We will soon have a video on this dialogue.

shadrackkibet:
I agree with all these and can be done by an intern. @Gioche6 do you want to give it a try?

Gioche6:
@shadrackkibet , I'll take this up. Thanks

rdstern:
Thanks @Gioche6 .  I have a further query, in addition to these 3 tasks - of course.
My example is the library janeaustenr, which has 6 data sets.
In the dialogue we can currently only import one file at a time.  Could we implement the facility to import multiple files.  So implementing <ctrl> click and <shift> click to be able to select multiple files, and then import them into separate data frames.
If you need help, then @N-thony has just implemented something similar for File > Import.

N-thony:
> Thanks @Gioche6 . I have a further query, in addition to these 3 tasks - of course. My example is the library janeaustenr, which has 6 data sets. In the dialogue we can currently only import one file at a time. Could we implement the facility to import multiple files. So implementing click and click to be able to select multiple files, and then import them into separate data frames. If you need help, then @N-thony has just implemented something similar for File > Import.

@Gioche6 I think for the multiple selection, it may be simple to implement than I thought, I have tried to run
`data_book$import_data(data_tables=list(attenu=attenu, attitude = attitude, airquality = airquality))` and you can see the result in the screenshot below, and it loads correctly the three datasets with their corresponding names.
So, have a look at the VB code, and try it, let me know if you get stuck. You will need also to activate `multiselect` to `True` in the property of the `ListView` i.e `lstCollection` that contains dataset names on the dialogue.

[...]

PR #7244, opened 2022-02-18
Improvements on Import from Library Dialog

Gioche6:
Partially Fixes Issue #7205

@africanmathsinitiative/developers , This is ready for review. 
@rdstern , I have fixed item 1, 2 and 3 in [7205`](https://github.com/africanmathsinitiative/R-Instat/issues/7205). 
I am currently working on item 4 on the comments. Kindly test it. Thank you

@shadrackkibet  Kindly test and review. 
@lloyddewit , could you update the Translation dialog for the renaming's in this PR, so that they can appear on the run-time?

PR #7342, opened 2022-03-30
Added double-click (and other small improvements) to Import from Library dialog

derekagorhom:
Fixes Issue #7205 

Adding the double click to the import dialog and also fixing items 1, 2 and 3 in #7205 

@N-thony can you kindly test this PR. 

Also @Patowhiz had some suggestions for the double click in the import From library Dialog


derekagorhom:
This PR replaces PR #7244

N-thony (approved):

rdstern (approved):
@derekagorhom this seems to work fine.

Patowhiz:
> @rdstern could you also test? I think the next thing to do will be to load multiple datasets once.

@derekagorhom  yes, and please note that would require the dialog's redesign and implementation. For instance importing multiple datasets may mean no need to give new data frame name, just like the import dialog.

Patowhiz:
@lloyddewit when looking into this with @derekagorhom, I think it's justified to have a `performClickOk` subroutine in `ucrButtons`? We had to access it's inner controls which didn't seem right to me.

We could use the `ClickOk` event because technically it's just an `AfterClickOk` event.

lloyddewit:
@derekagorhom thanks for this PR. 
Does it fully fix issue #7205 or just partially?

lloyddewit (approved):

derekagorhom:
> @derekagorhom thanks for this PR. Does it fully fix issue #7205 or just partially?

this PR fully fixes issue #7205.

lloyddewit:
> @lloyddewit when looking into this with @derekagorhom, I think it's justified to have a `performClickOk` subroutine in `ucrButtons`? We had to access it's inner controls which didn't seem right to me.
> 
> We could use the `ClickOk` event because technically it's just an `AfterClickOk` event.

@Patowhiz @derekagorhom Understood. Currently, when the user double clicks the list collection, we pretend that the user has clicked the OK button instead. It would be cleaner if the event handlers for the OK click and list collection double click, both call the same shared function.
I looked in `ucrButtons.vb`. The OK click is handled by `cmdOk_Click()`. This function also raises click events, so even if we extract this code into a new shared function, then we are still tightly coupled to click events.
In summary, there may be arguments for refactoring this, but I don't think this needs to be done in this PR.
So I'll go ahead and merge this PR. I'm happy to discuss if you think further changes are needed.
Thanks

lloyddewit:
> @derekagorhom thanks for this PR. Does it fully fix issue #7205 or just partially?

@derekagorhom If the answer is 'fully', then please close issue #7205, thanks

N-thony:
> > @derekagorhom thanks for this PR. Does it fully fix issue #7205 or just partially?
> 
> @derekagorhom If the answer is 'fully', then please close issue #7205, thanks

@lloyddewit  It doesn't fully fixes #7205, we still need to add the possibility of loading `multiple datasets`. If you close this then we need to open a new issue for the remaining work.

PR #7344, opened 2022-03-30
Fixed the bug on the result produced by PICSA when the Facets is by row or column

### LINKED ISSUE #7332: Possibly corrected feature in PICSA graphs has resurfaced
Here it is:  I thought this was fixed ages ago?  It is an extremely annoying bug, because we are getting reasonable statistics wrong.  

image

Here is another example:

image

I checked with the development version (above was earlier, in case we have introduced this:

image

N-thony:
Fixes #7332
@rdstern the result produced when changing the facet option by row or column was still the same due to the missing of `dplyr::group_by(station)` in the R code. I have added that now, please test and confirm if everything is fine. Thanks.

### DIFF
--- a/instat/dlgPICSARainfall.vb
+++ b/instat/dlgPICSARainfall.vb
@@ -852,18 +852,30 @@ Public Class dlgPICSARainfall
         SetPipeAssignTo()
     End Sub
 
+    Private Sub GetParameterValue(clsOperator As ROperator)
+        Dim i As Integer = 0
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@N-thony great, and also a great relief.  That's the sort of bug I hate, because it gives a plausible, but wrong answer!
@lloyddewit or @shadrackkibet I don't think there will be too much to check here.  SWo hope it can be merged soon.

lloyddewit (approved):

PR #7345, opened 2022-03-31
Added an option to Add Key for the new column in Prepare>Data Frame>Row Numbers/Names dialog

### LINKED ISSUE #7323: Two small improvements to the Prepare > Data Frame > Row Numbers/Names dialogue
This is a simple task and should only take a few minutes I hope.
1) The first option says Copy Row Names into Column.  Change this into `Copy Row Numbers or Names into a Column`
2) Move the other options down, so there is room for a checkbox for this option.
3) Add a checkbox (default checked) with label Make the Column into a Key for the data Frame.

When checked it runs the same code as in the Prepare > Keys and Links and also prints the information in the output window, as is done when you run the Keys dialogue with that column.

(To see what's needed take any data frame (e.g. mtcars from the datasets.  Then run the current Row Numbers/Names dialogue and then the Prepare > Keys and Links > Make Key dialogue and make the row variable into a key.  (With our new "system" we will avoid that step.
@shadrackkibet who should be doing this?

anastasia-mbithe:
This fixes #7323 
@rdstern and @shadrackkibet, I have added a checkbox to Add Keys to the New column created by the first option (Copy row names into Column). The checkbox is checked by default. I have been using the mtcars dataset to test the dialogue.
I have also fixed some other bug issues on the dialog. You can have a look at it.

### DIFF
--- a/instat/dlgRownamesOrNumbers.vb
+++ b/instat/dlgRownamesOrNumbers.vb
@@ -20,6 +20,8 @@ Public Class dlgRowNamesOrNumbers
     Public bFirstLoad As Boolean = True
     Private bReset As Boolean = True
     Private clsRowNamesFunction As New RFunction
+    Private clsAddKeyFunction As New RFunction
+    Private clsAddKeyParam As New RParameter
 
     Private Sub dlgRowNamesOrNumbers_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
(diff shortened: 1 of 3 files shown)


rdstern:
@anastasia-mbithe great - thanks - will do

rdstern (changes requested):
@ana that's great - it works fine
a) In the label you can delete into and have a capital for Data in Data Frame
b) Something new in this dialogue recently is that the default (if you don't use the Position button) is for the row to be the last column in the data frame.  IOn earlier versions it used to be the first one.  I don't know when that changed, (certainly before your work) but could you please change it back.  

And your comment in the pull request was great - very clear.

rdstern (approved):
@anastasia-mbithe excellent - many thanks
Over to @lloyddewit or @shadrackkibet to check and merge

instat/dlgRownamesOrNumbers.vb
```diff
@@ -20,6 +20,9 @@ Public Class dlgRowNamesOrNumbers
     Public bFirstLoad As Boolean = True
     Private bReset As Boolean = True
     Private clsRowNamesFunction As New RFunction
+    Private clsAddKeyFunction As New RFunction
+    Private clsDummyFunction As New RFunction
+    Private ResetRCode As Boolean = True
```
lloyddewit, line 25:
```suggestion
```

lloyddewit (changes requested):

lloyddewit (approved):
@anastasia-mbithe I found your coding very clean and readable, thank you.
I just made one samll change, and I'm happy to approve

lloyddewit:
@rdstern I just made one small change that will not affect the functionality. If you approve again, then we can merge
thanks

rdstern (approved):
@lloyddewit this is still working fine, so I am approving.

It has uncovered a new problem, I think, namely when a variable is deleted, that is a key, (or part of a key), then the key remains.  That should not be the case.  I'll make a new issue.

PR #7346, opened 2022-03-31
Updating the correlations dialog to use the newer corrr package

### LINKED ISSUE #7257: Update the correlations dialog to use the newer corrr package

Vitalis95:
Fixes #7257
@rdstern @N-thony @shadrackkibet , this PR updates correlation dialog to use the newer corrr package

### DIFF
--- a/instat/dlgCorrelation.designer.vb
+++ b/instat/dlgCorrelation.designer.vb
@@ -38,76 +38,49 @@ Partial Class dlgCorrelation
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()>
     Private Sub InitializeComponent()
-        Me.ucrChkDisplayOptions = New instat.ucrCheck()
-        Me.ucrChkAbsolute = New instat.ucrCheck()
-        Me.ucrChkShave = New instat.ucrCheck()
         Me.rdoCompleteRowsOnly = New System.Windows.Forms.RadioButton()
         Me.rdoPairwise = New System.Windows.Forms.RadioButton()
-        Me.ucrPnlCompletePairwise = New instat.UcrPanel()
         Me.lblMethod = New System.Windows.Forms.Label()
-        Me.ucrInputRearrange = New instat.ucrInputComboBox()
-        Me.ucrChkRearrange = New instat.ucrCheck()
-        Me.ucrInputDisplayNas = New instat.ucrInputComboBox()
         Me.lblDisplayNas = New System.Windows.Forms.Label()
-        Me.ucrChkLeadingZeros = New instat.ucrCheck()
         Me.lblDecimalPlaces = New System.Windows.Forms.Label()
-        Me.ucrNudDecimalPlaces = New instat.ucrNud()
         Me.rdoTwoColumns = New System.Windows.Forms.RadioButton()
         Me.rdoMultipleColumns = New System.Windows.Forms.RadioButton()
-        Me.ucrPnlColumns = New instat.UcrPanel()
         Me.grpMethod = New System.Windows.Forms.GroupBox()
         Me.rdoKendall = New System.Windows.Forms.RadioButton()
         Me.rdoPearson = New System.Windows.Forms.RadioButton()
         Me.rdoSpearman = New System.Windows.Forms.RadioButton()
-        Me.ucrPnlMethod = New instat.UcrPanel()
         Me.grpMissing = New System.Windows.Forms.GroupBox()
         Me.cmdOptions = New System.Windows.Forms.Button()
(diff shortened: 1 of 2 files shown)


rdstern:
@Vitalis95 this is really coming on now.  Next time I need another opinion on the testing, so I hope @lilyclements can examine it too - it was her suggestion and the results are exciting.
a) If easy, I was pleased to see it is in the 2 places now.  Can you make it so that 2 variables is the default top radio button for the 2/3 variables and multiple (as now) for the multivariate.  (I think that is done for the regular sequence dialogue, which is different under numeric and dates.) Don't spend long on this.
b) You Have Display on Diagonal outside the display box, and Display NAs inside.  They seem roughly the same?  Could you therefore change the Display NAs to Display on Diagonal and delete the current Display on Diagonal control.
c) The rearrange works nicely except when I tick the Absolute checkbox.  Then it gave an error.  (I have not tried many options on the rearrange.)
d) The old version displays the results in the output window.  The new one displays them in a new data frame instead.  It should be as well.  So the correlations should always be displayed in the results window.
e) Similarly the current (old) correlations dialogue saves a correlation object.  It doesn't have an option to display in a new data frame.  (Which is a great feature of the new version.)  But the new version no longer saves a correlation, which it should do.
f) That control also doesn't seem always to keep the previous setting.  When I unset it, it sometimes comes back set.
g) So, there will always be a correlation object and the new feature is to disp[lay in a data frame.  Can that be a new control - checkbox, default unchecked with label something like.   `Display in Data Frame` . There is a space problem, so perhaps this could go into the display box, and on the same row as the Shave?

Vitalis95:
@rdstern I am closing this PR  and open a new one in order to include changes from @N-thony

PR #7347, opened 2022-03-31
Improved Options by Context > Prepare > Merge Additional Data dialog

### LINKED ISSUE #7029: Update Merge Additional Data dialog

N-thony:
Fixes #7029
This PR replaces PR #7333 due to the conflicts.
@rdstern I made the changes here from your review in PR #7333, please take a look. Thanks.#7123 

### DIFF
--- a/instat/dlgMergeAdditionalData.vb
+++ b/instat/dlgMergeAdditionalData.vb
@@ -50,6 +50,8 @@ Public Class dlgMergeAdditionalData
         ucrToDataFrame.SetLabelText("To Data Frame:")
 
         ucrFromDataFrame.SetLabelText("From Data Frame:")
+        ucrFromDataFrame.SetParameter(New RParameter("data_name", 0))
+        ucrFromDataFrame.SetParameterIsString()
 
         ucrReceiverSecond.Selector = ucrFromDataFrame
         ucrReceiverSecond.bForceAsDataFrame = True
@@ -97,6 +99,7 @@ Public Class dlgMergeAdditionalData
     End Sub
 
     Private Sub SetRCodeforControls(bResetControls As Boolean)
+        ucrFromDataFrame.SetRCode(clsGetVariablesFunction, bResetControls)
         ucrToDataFrame.SetRCode(clsLeftJoinFunction, bResetControls)
         ucrChkSaveDataFrame.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
     End Sub
@@ -122,7 +125,7 @@ Public Class dlgMergeAdditionalData
         clsByListFunction.ClearParameters()
         SetDataFrameAssign()
         SetMergingBy()
-        clsGetVariablesFunction.AddParameter("data_name", Chr(34) & ucrFromDataFrame.ucrAvailableDataFrames.strCurrDataFrame & Chr(34), iPosition:=0)
+        SetInputCheckVisibility(False)
     End Sub
 
     Private Sub ucrChkSaveDataFrame_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkSaveDataFrame.ControlValueChanged, ucrInputSaveDataFrame.ControlValueChanged
@@ -138,12 +141,15 @@ Public Class dlgMergeAdditionalData
                     ucrBase.clsRsyntax.SetBaseRFunction(clsLeftJoinFunction)
                     ucrInputSaveDataFrame.Visible = True
                     bJoinColsAreUnique = True
+                    cmdCheckUnique.Visible = False
+                    SetInputCheckVisibility(False)
                 Else
                     clsLeftJoinFunction.SetAssignTo(ucrToDataFrame.cboAvailableDataFrames.Text)
                     clsInsertColumnFunction.AddParameter("data_name", Chr(34) & ucrToDataFrame.cboAvailableDataFrames.Text & Chr(34), iPosition:=0)
                     clsInsertColumnFunction.AddParameter("col_data", clsRFunctionParameter:=clsLeftJoinFunction, iPosition:=1)
                     ucrBase.clsRsyntax.SetBaseRFunction(clsInsertColumnFunction)
                     ucrInputSaveDataFrame.Visible = False
+                    cmdCheckUnique.Visible = True
                 End If
             Else
                 clsLeftJoinFunction.RemoveAssignTo()
@@ -252,7 +258,7 @@ Public Class dlgMergeAdditionalData
                 ucrInputCheckInput.SetName("Entries not unique. Check New Data Frame to create new dataframe.")
                 ucrInputCheckInput.txtInput.BackColor = Color.LightCoral
             Else
-                ucrInputCheckInput.SetName("Column(s) key unique.")
+                ucrInputCheckInput.SetName("Column(s) key unique in From Data Frame.")
                 ucrInputCheckInput.txtInput.BackColor = Color.LightGreen
                 bJoinColsAreUnique = True
             End If
@@ -262,15 +268,23 @@ Public Class dlgMergeAdditionalData
 
     Private Sub cmdCheckUnique_Click(sender As Object, e As EventArgs) Handles cmdCheckUnique.Click
         CheckUnique()
+        SetInputCheckVisibility(True)
+    End Sub
+
+    Private Sub SetInputCheckVisibility(bCheck As Boolean)
+        ucrInputCheckInput.Visible = bCheck
     End Sub
 
     Private Sub ucrReceiverSecond_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverSecond.ControlValueChanged
         AddColumns()
         EnableDisableCmdCheckUnique()
+        SetInputCheckVisibility(False)
     End Sub
 
     Private Sub ucrInputMergingBy_TextChanged(sender As Object, e As EventArgs) Handles ucrInputMergingBy.TextChanged
         AddColumns()
+        CheckUnique()
+        SetInputCheckVisibility(False)
     End Sub
 
     Private Sub Controls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrToDataFrame.ControlContentsChanged, ucrFromDataFrame.ControlContentsChanged, ucrReceiverSecond.ControlContentsChanged
--- a/instat/dlgMergeAdditionalData.vb
+++ b/instat/dlgMergeAdditionalData.vb
@@ -159,7 +159,6 @@ Public Class dlgMergeAdditionalData
     End Sub
 
     Private Sub SetMergingBy()
-        Dim dctJoinColumns As New Dictionary(Of String, String)
         Dim lstJoiningColumns As New List(Of String)
         lstJoinColumns.Clear()
         If ucrFromDataFrame.ucrAvailableDataFrames.cboAvailableDataFrames.Text <> "" AndAlso ucrToDataFrame.cboAvailableDataFrames.Text <> "" Then
--- a/instat/dlgMergeAdditionalData.vb
+++ b/instat/dlgMergeAdditionalData.vb
@@ -25,8 +25,10 @@ Public Class dlgMergeAdditionalData
     Private lstJoinColumns As New List(Of String)
     Private clsLeftJoinFunction As New RFunction
     Private clsByListFunction As New RFunction
+    Private clsGetVariablesFunction As New RFunction
     Private bResetSubdialog As Boolean = True
     Private bBySpecified As Boolean
+    Private bUniqueChecked As Boolean = False
 
     Private Sub dlgMergeAdditionalData_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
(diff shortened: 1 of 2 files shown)


rdstern (approved):
@N-thony I am approving as it seems to work.  I tried the full merge which does work on a left join - as you said.  But the resulting data frame is much larger.  So I wonder if you could do a check that the resulting frame is ther same length as the original.  If not, then it either stops, or it then puts the result into a new data frame.  But that could come later.  I would be happy to merge first.  So hope that @shadrackkibet might check and merge.

rdstern:
@N-thony and @shadrackkibet I have been thinking a bit more about the problem with the joining in the Add (Merge) dialogue.  I think it only works if it is a 1:1 or Many:1 join?  If I am right, then a simple check, could be that the number of values in the right-hand side are unique.  So those variables could be a key.  So why not apply that check first.  We already have that check in the key dialogue.  Perhaps we could even add the check to the dialogue?

N-thony:
> @N-thony and @shadrackkibet I have been thinking a bit more about the problem with the joining in the Add (Merge) dialogue. I think it only works if it is a 1:1 or Many:1 join? If I am right, then a simple check, could be that the number of values in the right-hand side are unique. So those variables could be a key. So why not apply that check first. We already have that check in the key dialogue. Perhaps we could even add the check to the dialogue?

@rdstern I have added the check of variables uniqueness, could you test?

rdstern:
Started ok with our usual survey, with 3 variables missing in one of them.  OK not enabled.  Then added row and it was ok.
But then deleted row and added village in the joining columns.  This gave an error with message
image

Then the columns were incorrect here.

image

Thern it crashed!

So I tried again.  without the row I found OK wasn't enabled.  It was then enabled when I added the checkbox to put it into a new data frame.  Good.
Then I deleted all linking except Village - which still works into a new data frame.  But it doesn't let me.  I woulod like to be able to merge on village only into a new data frame.

I wonder about a checkbox (default unchecked) above Other and with with label `From Data Frame Checked for Unique Values for Join `  If checked it adds the control from the Key dialogue.  

Finally I notice the name at the top should change to `Add (Merge) Data`

N-thony:
> 

@rdstern I have tried several time to reproduce this bug but I couldn't. Perhaps as discussed on Skype if you find this again we can have a quick call again?

rdstern:
@N-thony I can't repeat it either. I assume the title will change automatically, and also expect @Vitalis95 to add it to the Prepare menu.  So approving as is.  Perhaps @shadrackkibet or @dannyparsons can also check the functioning as they examine the code?

rdstern (dismissed):

instat/dlgMergeAdditionalData.vb
```diff
@@ -20,10 +20,15 @@ Imports RDotNet
 Public Class dlgMergeAdditionalData
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
-    Private clsLeftJoin As RFunction
-    Private clsByList As RFunction
+    Private clsInsertColumnFunction, clsGetColumnsFromData As New RFunction
+    Private lstVariables As New List(Of String)
```
lloyddewit, line 24: make this a local variable in `AddColumns()`?

instat/dlgMergeAdditionalData.vb
```diff
@@ -93,120 +114,150 @@ Public Class dlgMergeAdditionalData
...
-                chrColumns = expTemp.AsList(0).AsCharacter
-                For i As Integer = 0 To chrColumns.Count - 1
-                    clsByList.AddParameter(Chr(34) & chrColumns.Names(i) & Chr(34), Chr(34) & chrColumns(i) & Chr(34))
-                    ucrReceiverSecond.Add(chrColumns(i), ucrSecondSelector.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
+    Private Sub SetMergingBy()
+        Dim dctJoinColumns As New Dictionary(Of String, String)
+        Dim lstJoinPairs As New List(Of String)
+        Dim lstFirstColumns As List(Of String)
+        Dim lstSecondColumns As List(Of String)
+        Dim i As Integer = 0
```
lloyddewit, line 159: For each of these local variables, if we move the declaration to the block where the variable is used, then it makes the code more readable. Example for `i` below.
However, if you prefer it the way it is, then please leave unchanged.
```suggestion
```

instat/dlgMergeAdditionalData.vb
```diff
@@ -93,120 +114,150 @@ Public Class dlgMergeAdditionalData
...
-                clsLeftJoin.AddParameter("by", clsRFunctionParameter:=clsByList, iPosition:=2)
-                PopulateMergeByText()
-                bBySpecified = True
             Else
-                ucrInputMergingBy.SetName("No link between these data frames. Click 'Join Options' to specify merging columns.")
-                clsLeftJoin.RemoveParameterByName("by")
-                bBySpecified = False
+                lstFirstColumns = frmMain.clsRLink.GetColumnNames(ucrFromDataFrame.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
+                lstSecondColumns = frmMain.clsRLink.GetColumnNames(ucrToDataFrame.cboAvailableDataFrames.Text)
+                i = 0
```
lloyddewit, line 170:
```suggestion
                Dim i As Integer = 0
```

instat/dlgMergeAdditionalData.vb
```diff
@@ -93,120 +114,150 @@ Public Class dlgMergeAdditionalData
...
-            clsItemOperator.AddParameter("left", clsRFunctionParameter:=clsGetLink, iPosition:=0)
-            clsItemOperator.AddParameter("right", "link_columns", iPosition:=1)
-            expTemp = frmMain.clsRLink.RunInternalScriptGetValue(clsItemOperator.ToScript(), bSilent:=True)
-            If expTemp IsNot Nothing AndAlso expTemp.Type = Internals.SymbolicExpressionType.List AndAlso expTemp.AsList.Count > 0 Then
-                chrColumns = expTemp.AsList(0).AsCharacter
-                For i As Integer = 0 To chrColumns.Count - 1
-                    clsByList.AddParameter(Chr(34) & chrColumns.Names(i) & Chr(34), Chr(34) & chrColumns(i) & Chr(34))
-                    ucrReceiverSecond.Add(chrColumns(i), ucrSecondSelector.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
+    Private Sub SetMergingBy()
+        Dim dctJoinColumns As New Dictionary(Of String, String)
```
lloyddewit, line 155: The key and value always seem to be equal. So could we use a list instead of a dictionary?

instat/dlgMergeAdditionalData.vb
```diff
@@ -20,10 +20,15 @@ Imports RDotNet
...
-    Private clsByList As RFunction
+    Private clsInsertColumnFunction, clsGetColumnsFromData As New RFunction
+    Private lstVariables As New List(Of String)
+    Private lstJoinColumns As New List(Of String)
+    Private clsLeftJoinFunction As New RFunction
+    Private clsByListFunction As New RFunction
+    Private clsGetVariablesFunction As New RFunction
     Private bResetSubdialog As Boolean = True
     Private bBySpecified As Boolean
+    Private bUniqueChecked As Boolean = False
```
lloyddewit, line 31: `bUniqueChecked` sounds like it should be true if uniqueness has been checked, else false.
However, this boolean is set to false if uniqueness checked but duplicates found.
Therefore this name is a bit confusing.
Rename (e.g. `bJoinColsAreUnique`) or add a comment?

lloyddewit (changes requested):
@N-thony Thank you, this looks very good. Just a few questions

instat/dlgMergeAdditionalData.vb
N-thony: @lloyddewit I think we may have a case that we will need to join by different column name but which has the same structure e.g we can link A.x = B.y where A, B are datasets and x,y are variables with same structure.

N-thony:
@lloyddewit I resolved your comments, please take a look

rdstern (dismissed):
I am approving, because it still works.
However, I tried with the linking variables, before filling the multiple receiver.  I had just village, which @N-thony is wrong.
Then I tried to fill the multiple receiver, and it gave me the message each time.  I'm not delighted by that, though most users will complete the multiple receiver first.  I would prefer not to check there and to add the check in the current key dialogue, as an option below.  (Default unchecked.) Then,  if the user wants to check uniqueness, there is an obvious way to do it.  (And doing that check would not trigger the message, because that would be a repeat of the check.
But it could come later if needed.  It would be good to get this dialogue merged and added to the Prepare menu.

N-thony:
> I am approving, because it still works. However, I tried with the linking variables, before filling the multiple receiver. I had just village, which @N-thony is wrong. Then I tried to fill the multiple receiver, and it gave me the message each time. I'm not delighted by that, though most users will complete the multiple receiver first. I would prefer not to check there and to add the check in the current key dialogue, as an option below. (Default unchecked.) Then, if the user wants to check uniqueness, there is an obvious way to do it. (And doing that check would not trigger the message, because that would be a repeat of the check. But it could come later if needed. It would be good to get this dialogue merged and added to the Prepare menu.

@rdstern please take a look. I made the change as agreed from our recent Skype call discussion.

N-thony:
> The functioning is to me, now perfect. Please could the messages change slightly. Then I will be really happy to approve. a) When ok, i.e. green, it says `Join column(s) unique in From Data Frame` I am not assuming people automatically understand about the concept of a key. So I am less keen on mentioning that. b) When not ok, i.e. red it says something like `From Data Frame entries not unique.` I am not mentioning the option of New Data frame in the message, because usually the action will be to make them unique instead.

@rdstern I have changed the message, please take a look. Thanks.

rdstern (dismissed):
Great and a good improvement.  There is work still for @Vitalis95 I think, so it is included also in the Prepare menu.  But that's a separate pull request.
@lloyddewit or @shadrackkibet I hope this can be merged soon.

instat/dlgMergeAdditionalData.vb
```diff
@@ -93,120 +119,176 @@ Public Class dlgMergeAdditionalData
...
             Next
+            clsGetColumnsFromData.AddParameter("data_name", Chr(34) & ucrFromDataFrame.ucrAvailableDataFrames.strCurrDataFrame & Chr(34), iPosition:=0)
+            clsGetColumnsFromData.AddParameter("col_names", frmMain.clsRLink.GetListAsRString(lstVariables, bWithQuotes:=True), iPosition:=1)
+        End If
+    End Sub
+
+    Private Sub EnableDisableCmdCheckUnique()
+        If ucrReceiverSecond.IsEmpty AndAlso ucrInputMergingBy.IsEmpty Then
+            cmdCheckUnique.Enabled = False
         Else
```
lloyddewit, line 236:
```suggestion
        cmdCheckUnique.Enabled = Not ucrReceiverSecond.IsEmpty OrElse Not ucrInputMergingBy.IsEmpty
```

instat/dlgMergeAdditionalData.vb
```diff
@@ -93,120 +119,176 @@ Public Class dlgMergeAdditionalData
...
-                        strMergeBy = strMergeBy & chrColumns(i) & "=" & chrColumns(i)
-                        ucrReceiverSecond.Add(chrColumns(i), ucrSecondSelector.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
-                    Next
-                    bBySpecified = True
-                Else
-                    strMergeBy = "No columns with the same name. Click 'Join Options' to specify merging columns."
-                    bBySpecified = False
-                End If
+            cmdCheckUnique.Enabled = True
+        End If
```
lloyddewit, line 238:
```suggestion
```

instat/dlgMergeAdditionalData.vb
```diff
@@ -93,120 +119,176 @@ Public Class dlgMergeAdditionalData
...
-            clsItemOperator.AddParameter("left", clsRFunctionParameter:=clsGetLink, iPosition:=0)
-            clsItemOperator.AddParameter("right", "link_columns", iPosition:=1)
-            expTemp = frmMain.clsRLink.RunInternalScriptGetValue(clsItemOperator.ToScript(), bSilent:=True)
-            If expTemp IsNot Nothing AndAlso expTemp.Type = Internals.SymbolicExpressionType.List AndAlso expTemp.AsList.Count > 0 Then
-                chrColumns = expTemp.AsList(0).AsCharacter
-                For i As Integer = 0 To chrColumns.Count - 1
-                    clsByList.AddParameter(Chr(34) & chrColumns.Names(i) & Chr(34), Chr(34) & chrColumns(i) & Chr(34))
-                    ucrReceiverSecond.Add(chrColumns(i), ucrSecondSelector.ucrAvailableDataFrames.cboAvailableDataFrames.Text)
+    Private Sub SetMergingBy()
+        Dim dctJoinColumns As New Dictionary(Of String, String)
```
lloyddewit, line 162:
```suggestion
```

lloyddewit (changes requested):

lloyddewit (approved):

lloyddewit:
@N-thony Thank  you I marked all my previous comments as resolved.
@rdstern I made some small adjustments, the functionality should be the same, please could you do a quick sanity check and approve again?
Thanks

rdstern (approved):
Great!

PR #7348, opened 2022-04-01
Added an option to select variables by value labels in Select Columns dialog

### LINKED ISSUE #7340: Extend select when data has value labels
Now I know why the sj packages includes searching in value labels.  I am afraid we need it in select.  @shadrack, it will become clear when you look at the SPSS data I sent in skype.
These data are excellent candidates for select, because there are quite a number of variables.  Some are labelled.  I was assuming the labelled variables would be the ones that become factors.  It is still important to be able to choose those variables with value labels.  But we need to go further, because some numeric variables, like age, are also labelled.  For example BH4Y Variable Label (Year of birth) also has value labels  `Inconsistent = 9997,  NO RESPONSE = 9999`

So I would like to be able to select variables with value labels that contain "NO RESPONSE" say!  

Interesting!

shadrackkibet:
fixes #7340

@rdstern this PR adds an option to select columns by value labels. Please test. 

More on select to come!


rdstern (approved):
image
Great.  I am approving.
1. I would still like an option to be able to (actually) hide the variables in the metadata.  In that case the initial row becomes red.  
2. And the extension, in the future, to be able to add a condition in the labelled data will be great.  Some of the labelled data are factors, while others are numeric, plus a few labels, e.g. starting with inconsistent is anumeric I think.

lloyddewit (approved):

shadrackkibet:
@lloyddewit really thanks for looking into this and merging. I really appreciate it.

PR #7352, opened 2022-04-04
Fixing limits issue in Temperature Quality Control dialog

### LINKED ISSUE #7331: Problems introduced on temperature quality control
This is in the new development version and has been introduced since version 0.7.4.  It seems ok there.
There may be more errors, but one of them is obvious.  I can supply data where I noticed it, if need be, but I think it will be obvious for any data, e.g. dodoma.

In the first check the default is the limits for tmax are 10 to 50 and tmin is 1 to 30.  In the results it was obvious (in my data) that it use the 10 degrees for tmax also for tmin.  I am not sure of the top end.

Please can careful checks of the code be made, for these and other options.  P)lease ccompare the code with 7.4, because it is ok there.

I am suggesting @shadrackkibet , because the R is quite complicated - at least to me it is.  But he may want to re-allocate.  Who changed this dialogue for the current version

Vitalis95:
Fixes #7331 
@rdstern @shadrackkibet , this PR makes tmax and tmin to use the limits specified at the bottom and at the top


rdstern (approved):
@Vitalis95 that's a relief - it now seems ok, from a quick check.  Over to @shadrackkibet or @lloyddewit

PR #7353, opened 2022-04-04
Fixed minor issues in the Prepare>Date>Make Date dialog

### LINKED ISSUE #7319: Issues in the Prepare>Date>Make Date dialog
@Patowhiz and I have been looking at the Prepare >Make Date Dialog, specifically the Year and day of the Year option.
We noticed 3 issues;
1. The dialog positions the new column at the end of the dataset instead of after the date of the year column. 
2. The cut-off up down control does not reset back to 0(current original value).
3. The selector should allow only numeric variables since the option only accepts numeric variables as Year or day of the year.

@rdstern are the 3 issues intentionally implemented that way or they are just bugs.

@rdstern, I do not understand the cut-off concept, suppose we have 2 digit year column(yy) of values 0-90 and we intend to make a 4 digit year column(yyyy) from 1900 to1990, what will be the cut off then? And what about if we want from year 2000 to 2090.

anastasia-mbithe:
This fixes #7319
@rdstern and @Patowhiz, In the Year and Day of the Year option, the selector allows Numeric variables for the two receivers and the up-down Cut Off now resets to 0 which is the original value. With the help of @N-thony, the dialog now positions the new column after the column used by the receiver.
In the Single Column option, the Origin option opens a pull-down with a default of "Excel (1899/12/30)"
You can have a look at this and test it.

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -2507,19 +2507,15 @@ DataSheet$set("public","make_date_yearmonthday", function(year, month, day, f_ye
 )
 
 # Not sure if doy_format should be a parameter? There seems to only be one format for it.
-DataSheet$set("public","make_date_yeardoy", function(year, doy, year_format = "%Y", doy_format = "%j", doy_typical_length = "366") {
-  year_col <- self$get_columns_from_data(year, use_current_filter = FALSE)
-  doy_col <- self$get_columns_from_data(doy, use_current_filter = FALSE)
+DataSheet$set("public","make_date_yeardoy", function(year, doy, base, doy_typical_length = "366") {
+  if(!missing(year)) year_col <- self$get_columns_from_data(year, use_current_filter = FALSE)
+  if(!missing(doy)) doy_col <- self$get_columns_from_data(doy, use_current_filter = FALSE)
   
-  if(missing(year_format)) {
-    year_counts <- str_count(year)
-    if(anyDuplicated(year_counts) != 0) stop("Year column has inconsistent year formats")
-    else {
-      year_length <- year_counts[1]
-      if(year_length == 2) year_format = "%y"
-      else if(year_length == 4) year_format = "%Y"
-      else stop("Cannot detect year format with ", year_length, " digits.")
-    }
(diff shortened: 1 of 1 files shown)


rdstern:
@anastasia-mbithe that is looking good.
a) On the default dialogue (one column) could you improve the tooltip 
image

I don't know why there are so many single quote symbols.  Do we need any?
b) On the Year /Day of Year, could you make the default 30 - so it is consistent with your other dialogue.  And could you check the code.  I tried it and it currently insists that dodoma starts in 2035, not 1935.

N-thony:
> @anastasia-mbithe that is looking good. a) On the default dialogue (one column) could you improve the tooltip image
> 
> I don't know why there are so many single quote symbols. Do we need any? b) On the Year /Day of Year, could you make the default 30 - so it is consistent with your other dialogue. And could you check the code. I tried it and it currently insists that dodoma starts in 2035, not 1935.

@anastasia-mbithe how is it going? We can discuss on Skype.

anastasia-mbithe:
@N-thony Okay

anastasia-mbithe:
@rdstern The dialog is now well with "base" added as a parameter, thanks to @N-thony. 
For the testing, I used Dodoma data, converted the Year column to a character, and used the Text>Transform dialog to substring the last two characters. I then converted the new column to numeric and used it in the dialog.
I set the base to 35 so that I could observe the difference. Setting it to 30, the new date falls into the 1900s.

codeyear

anastasia-mbithe:
@rdstern, you can have a look at this, and probably test it.

rdstern:
@anastasia-mbithe I can't now get the dialogue to work?
a) I used the library data from the climdex package - and one will do.
b) First with the year as given - the julian day is the 365/366 option
c) Then I used the calculator (logical keyboard) to get the 2-digit years.  

Either way I get this error?  Am I doing something wrong?  It used to work I think.  

image

anastasia-mbithe:
@rdstern, this now works fine as before. Kindly test it.

N-thony:
> @rdstern, this now works fine as before. Kindly test it.

@rdstern can you test?

rdstern (approved):
Looks good

lloyddewit (approved):

Issue #7299, opened 2022-03-13
Small changes in the Prepare > Check Data > Duplicates dialogue 


rdstern:
1. Could the name of the dialogue be Duplicate Rows.  That is at the top of the dialogue and hopefully also the Menu item (unless that should wait for other changes to the menus - see #7297 .
2. The most used option is the second, namely Selected Variables.  So could that become the first button, and hence the default when you first use the dialogue.  The Data Frame is then second.  The Successive values remains the last button.
3. In the 3rd radio button below the selector, the label i Index Number of Duplicates.  Then the pull down has the first entry as Count.  Please swap them, so the first entry is Index.
4. Currently this control doesn't "remember" the setting when you return to the dialogue.  Could it please remember.

@shadrackkibet who can do these.  It should only taker a few minutes?

derekagorhom:
@rdstern we looked into this issue with @Patowhiz  and we also realized that the dialog is producing duplicate R commands. I wonder if this was intentional.

What the dialog produces;
```
# Code generated by the dialog, Duplicate Rows

dup <- duplicated_count_index(x=data_book$get_columns_from_data(data_name="date_example", col_names="week", use_current_filter=FALSE), type="index")
data_book$add_columns_to_data(data_name="date_example", col_name="dup", col_data=dup, before=FALSE, adjacent_column="week")

dup <- data_book$get_columns_from_data(data_name="date_example", col_names="dup")
data_book$add_columns_to_data(data_name="date_example", col_name="dup", col_data=dup, before=FALSE, adjacent_column="week")
summary(object=data_book$get_columns_from_data(data_name="date_example", col_names="dup"))
rm(list=c("dup", "dup"))
```

What we think it should be producing;
```
# Code generated by the dialog, Duplicate Rows

dup1 <- duplicated_count_index(x=data_book$get_columns_from_data(data_name="date_example", col_names="week", use_current_filter=FALSE), type="index")
data_book$add_columns_to_data(data_name="date_example", col_name="dup1", col_data=dup1, before=FALSE, adjacent_column="week")
summary(object=data_book$get_columns_from_data(data_name="date_example", col_names="dup1"))
rm(list=c("dup1", "dup1"))
```

If the current duplicated R commands is not correct, should we correct it in PR #7305 or in a new PR?

PR #7305, opened 2022-03-15
Partially Fixes Duplicates Dialogue

derekagorhom:
Fixes (partially) issue #7299
Small changes in the duplicates Dialogue 

Selected Variables has become the first button, and hence the default when you first use the dialogue. The Data Frame is then second. The Successive values remains the last button. 
And the pull down has the first entry as Index.

@Ivanluv could you please test this to confirm

PR #7354, opened 2022-04-04
Small improvements to the Prepare > Check Data > Duplicates dialog

derekagorhom:
Fixes issue #7299
Small changes in the duplicates Dialogue

Selected Variables has become the first button, and hence the default when you first use the dialogue. The Data Frame is then second. The Successive values remains the last button.
And the pull down has the first entry as Index.

@Patowhiz  and I will make a new PR for the database changes recommended by @lloyddewit  in #7305


rdstern (changes requested):
I really did think this was only a small task:
image
I am keen that the option titled Index Number of Duplicates should have Index as the default.  It still has Count.
I changed it to Index and it works fine.  Then the screen shot show what happens when I return.  It has changed back to Count.  It is supposed to remember the previous setting.
I notice you have changed Index to be the first of the 2 options, where it was Count before.  That's fine, but it isn't sufficient.

derekagorhom:
Hello @rdstern i have done all these changes on the PR, but they do not show because of the database Translation, that is why we will make a PR for the database translation to correct this PR

Patowhiz:
@lloyddewit @rdstern,  @derekagorhom and I have created PR #7363 that contains the required translation changes. Once merged this will be ready for review. 
Thanks

rdstern (changes requested):
@derekagorhom and @Patowhiz I don't understand why this change is causing so many problems.
image

In this option Count is still given as the default (Though Index is the first of the two options.
And if I change to Index, then it does it fine.  But when I return to the dialogue it still has returned to count.

If this continues, perhaps instead we should have a 4th option for the radio button.  But I really don't know why Index couldn't be the default, and why it can't remember the setting.  That's what we usually do?

derekagorhom:
@rdstern sorry for the delay, PR #7363 will solve the translation issues with this PR

Patowhiz:
@rdstern you may have to wait for the translation PR #7363 to be merged. @lloyddewit is aware of that too.

derekagorhom:
@rdstern could you test this again, i merged the approved PR #7363 to this and it seems to work fine now

rdstern (approved):
Great!

lloyddewit (changes requested):

lloyddewit (approved):

lloyddewit:
@rdstern I made a small change that should not affect the functionality. If you test/approve again, then we can merge.
Thanks!

rdstern (approved):
Great

PR #7355, opened 2022-04-05
Fixed bug when colour is applied in the dataset from Data View window

### LINKED ISSUE #7232: Complete the Experiments > Define > On-Station dialogue
This is currently quite neat except it doesn't seem to do anything.
It currently shows the different metadata and you choose 1 property to colour.
But it doesn't colour yet!  Instead it adds a Colour property to the metadata and gives it a different number for each different value of the property you choose.

There is an option to `Remove colour`.  This keeps the Colour property, but changes the colour to -1.
You also get -1 if the property you use to colour has NA.

It would be nice to add colours!
Ideally we would colour columns in the data frame and rows in the column metadata.

1. I quite like the idea of separating the production of the colour meta-data from its application.  So should we have an `Apply Colour Property` checkbox?  Default is checked. We could go further and have Apply to data frame default checked, and also Apply to Column Metadata, default is unchecked.
I could picture including this dialogue in teaching about the meta-data and the different windows in R-Instat.
2. Then we need to choose the colours.  Could we do this simply by having a selection of palettes, and then choosing a palette?
The Climatic > Describe > Wind > Windrose dialogue has a neat option for palettes as a sub-dialogue.  Could we simply adopt those?
The code there must already associate colours with the numbers.
3. The class is the obvious property.  I suggest we have that as the default.  But allow any property to be used (as now)

@dannyparsons this looks like a "low-hanging fruit"?  I would like something, so the help can be written.  Any quick comments? 
 
I am liking this.  I started, feeling this was unimportant, but I think it can help in our teaching role - so making R-Instat easier to use!

N-thony:
Fixes #7232
@rdstern I have fixed the bug on not setting the colour to the dataset in the `Data View`, in addition while testing I fixed one small bug I found in the `Merge Additional Dialogue`. I also realised in version `0.7.2` that with dataset you shared in Skype, the merge dialogue seems to not apply the colour though the second or first dataframe has colour applied but with this version, it seems to keep the colour if applied from one of the dataset in the merged dataset. Please could test on your side and see if I'm right? Thanks.

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -2309,7 +2309,7 @@ DataSheet$set("public", "add_dependent_columns", function(columns, dependent_col
 
 DataSheet$set("public", "set_column_colours", function(columns, colours) {
   if(missing(columns)) columns <- self$get_column_names()
-  #if(length(columns) != length(colours)) stop("columns must be the same length as colours")
+  if(length(columns) != length(colours)) stop("columns must be the same length as colours")
   
   for(i in 1:length(columns)) {
     self$append_to_variables_metadata(columns[i], colour_label, colours[i])
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -1067,8 +1067,8 @@ DataSheet$set("public", "append_to_variables_metadata", function(col_names, prop
     for (curr_col in col_names) {
       #see comments in  PR #7247 to understand why ' property == labels_label && new_val == "" ' check was added
       #see comments in issue #7337 to understand why the !is.null(new_val) check was added. 
-      if (property == labels_label && !is.null(new_val) && new_val == "") {
-        #reset the column labels property 
+      if (((property == labels_label && new_val == "") ||(property == colour_label && new_val == -1)) && !is.null(new_val)) {
+        #reset the column labels or colour property 
         attr(private$data[[curr_col]], property) <- NULL
       } else {
         attr(private$data[[curr_col]], property) <- new_val
@@ -1079,8 +1079,8 @@ DataSheet$set("public", "append_to_variables_metadata", function(col_names, prop
     for (col_name in self$get_column_names()) {
       #see comments in  PR #7247 to understand why ' property == labels_label && new_val == "" ' check was added
       #see comments in issue #7337 to understand why the !is.null(new_val) check was added. 
-      if (property == labels_label && !is.null(new_val) && new_val == "") {
-        #reset the column labels property 
+      if (((property == labels_label && new_val == "") ||(property == colour_label && new_val == -1)) && !is.null(new_val)) {
+        #reset the column labels or colour property 
         attr(private$data[[col_name]], property) <- NULL
       } else {
         attr(private$data[[col_name]], property) <- new_val
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -2324,7 +2324,8 @@ DataSheet$set("public", "has_colours", function(columns) {
 }
 )
 
-DataSheet$set("public", "set_column_colours_by_metadata", function(columns, property) {
+DataSheet$set("public", "set_column_colours_by_metadata", function(data_name, columns, property) {
+if(!missing(data_name) && missing(columns)) columns <- names(self$get_data_frame(data_name = data_name))
   if(missing(columns)) property_values <- self$get_variables_metadata(property = property)
   else property_values <- self$get_variables_metadata(property = property, column = columns)
   
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -2309,7 +2309,7 @@ DataSheet$set("public", "add_dependent_columns", function(columns, dependent_col
 
 DataSheet$set("public", "set_column_colours", function(columns, colours) {
   if(missing(columns)) columns <- self$get_column_names()
-  if(length(columns) != length(colours)) stop("columns must be the same length as colours")
+  #if(length(columns) != length(colours)) stop("columns must be the same length as colours")
   
   for(i in 1:length(columns)) {
     self$append_to_variables_metadata(columns[i], colour_label, colours[i])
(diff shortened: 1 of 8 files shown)


rdstern:
@N-thony this now seems to work, so that's a relief.
You may like to pass this on, or continue, for some of the other aspects on this dialogue.
a) The remove colour seems to work fine.  But it could enable Ok, even when the receiver is empty.  (You don't need to know what property made the colours to clear them.)
b) I thought you might like to remove the property from the metadata, when it is always -1?
c) When there are no colours, perhaps the default could be the Class?
d) I would still like cho9ice on the palettes to be used.  This could be a useful very simple dialogue to illustrate the palettes that are usually available.  (If needed, this could become another issue, to be done later.)

N-thony:
> @N-thony this now seems to work, so that's a relief. You may like to pass this on, or continue, for some of the other aspects on this dialogue. a) The remove colour seems to work fine. But it could enable Ok, even when the receiver is empty. (You don't need to know what property made the colours to clear them.) b) I thought you might like to remove the property from the metadata, when it is always -1? c) When there are no colours, perhaps the default could be the Class? d) I would still like cho9ice on the palettes to be used. This could be a useful very simple dialogue to illustrate the palettes that are usually available. (If needed, this could become another issue, to be done later.)

@rdstern I resolved items a) to c), please could you check? Thanks.

ChrisMarsh82 (dismissed):
add _grid.UpdateWorksheetStyle(fillWorkSheet) back to ucrDataView

ChrisMarsh82:
@N-thony see the latest commit. it looks like there is an issue with reogrid when using entire range. This looks to solve it by changing the style of the top row separately. Please could you check to see if this solution fixes your issue while still allowing users to select a font through the options menu

N-thony:
> @N-thony see the latest commit. it looks like there is an issue with reogrid when using entire range. This looks to solve it by changing the style of the top row separately. Please could you check to see if this solution fixes your issue while still allowing users to select a font through the options menu

@ChrisMarsh82 thank you, I have checked by setting the colour with `Colour by Property dialogue` and also with the font through the options menu and it seems to work fine.
@rdstern could you re-test?

rdstern:
@N-thony this seems to work well now - at least initially and visually.
a) The Remove Colour checkbox is available all the time.  Should it be deleted if no colour is applied.

b) I removed colour and the result in the metadata seems odd.  Visually it is ok.  But it has NA for the first entry of metadata and still some numbers.  That's perhaps because of the Select I applied.  
image


c) So I deleted the selection and it gave an error.

image

N-thony:
@rdstern this seems to be fixed now.

rdstern:
@N-thony this time I tried it the other way round.  So I did a select, before doing the colour.  Got this:
image

Makes me wonder about doing a filter - I'll try.  That seems ok.

N-thony:
> @N-thony this time I tried it the other way round. So I did a select, before doing the colour. Got this: image
> 
> Makes me wonder about doing a filter - I'll try. That seems ok.

@rdstern could you test now?

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -2309,7 +2309,7 @@ DataSheet$set("public", "add_dependent_columns", function(columns, dependent_col
 
 DataSheet$set("public", "set_column_colours", function(columns, colours) {
   if(missing(columns)) columns <- self$get_column_names()
-  if(length(columns) != length(colours)) stop("columns must be the same length as colours")
+  #if(length(columns) != length(colours)) stop("columns must be the same length as colours")
```
N-thony, line 2312: @dannyparsons @shadrackkibet do we need this check of the length between columns and colours?

N-thony:
@rdstern could you try this now?

lloyddewit:
> @rdstern could you try this now?

@N-thony Before @rdstern tests again, please could you resolve my comment above (else @rdstern will need to test twice)? thanks

instat/static/InstatObject/R/data_object_R6.R
N-thony: @lloyddewit this if statement was needed, I have been looking at it with @lilyclements to understand the need of it, also fixed the bug somewhere else in the R code.

Patowhiz:
I'm now also getting this bug.
When I use 2 data frames with the similar column names, and I try to produce new columns from the dialogs that seem to also be the same across the 2 data frames. this error is thrown.

lloyddewit:
> > @rdstern could you try this now?
> 
> @N-thony Before @rdstern tests again, please could you resolve my comment above (else @rdstern will need to test twice)? thanks

@N-thony I saw that you resolved my comment, thanks

lloyddewit:
@N-thony Is this PR now ready for testing or do you still need to take action to address @Patowhiz 's comment above?
Thanks for clarifying

N-thony:
> @N-thony Is this PR now ready for testing or do you still need to take action to address @Patowhiz 's comment above? Thanks for clarifying

@lloyddewit I have discussed with @Patowhiz and found his comment has been fixed.

lloyddewit:
@rdstern Please can you test?
@ChrisMarsh82 You have an open change request above, which I think will block approval. Please can you approve (I will review/approve separately)?
Thanks

rdstern:
@N-thony I got build errors?

N-thony:
> 

@rdstern I'm not getting the build error, let's have a quick call?

lloyddewit:
> > 
> 
> @rdstern I'm not getting the build error, let's have a quick call?

@N-thony I also get build errors, `clsDataFramePage.vb` is incorrect:
image

N-thony:
> > > 
> > 
> > 
> > @rdstern I'm not getting the build error, let's have a quick call?
> 
> @N-thony I also get build errors, `clsDataFramePage.vb` is incorrect: image

@lloyddewit thanks, I think this is after resolving the conflicts, let me fix it.

N-thony:
> > > > 
> > > 
> > > 
> > > @rdstern I'm not getting the build error, let's have a quick call?
> > 
> > 
> > @N-thony I also get build errors, `clsDataFramePage.vb` is incorrect: image
> 
> @lloyddewit thanks, I think this is after resolving the conflicts, let me fix it.

@lloyddewit could you try it now?

lloyddewit:
@N-thony Thank you for the changes. The PR now builds without errors.
@rdstern You should be able to test now

rdstern (approved):
Good

lloyddewit (approved):

PR #7362, opened 2022-04-06
Improvements to the Prepare> Factor> View/Delete Labels dialog

### LINKED ISSUE #7341: Small addition needed in the Prepare > Column: Factor > View/Delete Labels dialogue
This should be a very simple task.  @shadrack to allocate.

It is on the View button:

It is a 4th option in the Display Options group.  That box can start higher, then add another checkbox at the bottom, default unchecked.  The label is `Max Labels:`.  If checked it opens an updown with default 15.  It can go up or down in integer jumps.  I assume down to 1, but try and check.  Can use MICS survey data.  If checked, then it adds max.len = value to the command.

anastasia-mbithe:
This fixes #7341 
@rdstern and @shadrackkibet , I have added a checkbox (Max Labels) that opens an up-down with a default of 15. When checked, it adds a parameter max.len in the function. I have used the Survey data for testing.

I noticed the selector in the dialog does not allow factors only, it allows all variables. Is it right that way, or should we allow only factors for the dialog?


rdstern (changes requested):
@anastasia-mbithe that's great and I could approve.
However, I have (of course!) found another problem!  This was there before, but I mention it in case it is easy to solve.  It may be messy, in which case we'll do it separately.
When you use the dialogue with a few variables then it works fine.
With lots of variables it repeats the (html) output as many times as it has lines of the command - which is also oddly, in the output window?  So, if you have (say) 30 variables then you get the results about 4 times!
Can the R co9de be given differently, so the result is just given once?

lilyclements:
@anastasia-mbithe why was this in "blocker"?

lilyclements:
@anastasia-mbithe why was this in "blocker"?

lilyclements:
@anastasia-mbithe why was this in "blocker"?

lilyclements:
@anastasia-mbithe why was this in "blocker"?

anastasia-mbithe:
@rdstern, this fixes the Html output repeating itself severally when using many variables. Kindly have a look at it.

rdstern (approved):
THis looks good now.

lloyddewit:
I will review

lloyddewit (approved):
@anastasia-mbithe clear and readable, thank you

lloyddewit:
@rdstern there's a couple of small changes since your last approval.
If you can test/approve again, then we can merge.
Thanks

anastasia-mbithe:
@rdstern and @lloyddewit, in the last commit I did away with the dplyr::select() and made the receiver give the columns selected. It still works the same way.

rdstern (approved):

PR #7375, opened 2022-04-13
In grid, ensured that 'Levels/labels' option is only enabled when cell is a factor or ordered factor

### LINKED ISSUE #7366: Disable Levels/labels in cell right
**Levels/labels** tool strip item should be disabled when the active cell is not a factor /ordered factor column.

image

Just check how this is done in the column right-click.

image

anastasia-mbithe:
This fixes #7366 
@shadrackkibet,I have disabled the Levels/Labels strip item when the active cell is not a factor. 
Screenshot (112)


lloyddewit (approved):

lloyddewit:

rdstern (approved):

PR #7383, opened 2022-04-17
Bug fixed on Add (Merge) Columns when merging with different datasets length

### LINKED ISSUE #7382: Add (Merge) Columns has problem with example for the Help
The data are 3 files from the climdex.pcic package.  They are 3 separate elements.  The longest is tmin, so add the others to that.
This works fine with the left-join in the ordinary merge, but gives an error in this dialogue. 
image

I would like to merge the flag field as well, but that's an option.  It is not needed, and should be deleted from the joining variables as shown above.  When merging I get:
image

As mentioned, the left join works fine in the usual merge dialogue and does just what I think this dialogue should do.

N-thony:
Fixes #7382
@rdstern I have fixed the bug as described in the corresponding issue, this problem is fixed by swapping the `x` and `y` parameters values in the `left_join function`. Please take a look. Thanks.

### DIFF
--- a/instat/dlgMergeAdditionalData.vb
+++ b/instat/dlgMergeAdditionalData.vb
@@ -20,7 +20,7 @@ Imports RDotNet
 Public Class dlgMergeAdditionalData
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@N-thony that's great now

lloyddewit (approved):

PR #7390, opened 2022-04-19
Added With Replacement check box to Sample/Permute Rows dialog

### LINKED ISSUE #7077: Small change in Prepare > Column: Numeric > Permute Columns dialogue
These are all minor linked changes and so are put together.  They may be done in successive pull requests, but all are small. 

First is in the Prepare > Column: Numeric > Permute Columns.
This is a minor change also in the menu item.  It currently says `Permute Columns...`  The change should be to `Permute/Sample Rows...`  (It permutes rows within a column).  

This is a task for @shadrackkibet and one of many in changes to frmmain to be collected and done.  The remaining tasks are small enough that he may wish to do them - or probably better to allocate to someone.

a) Then a very minor change in the dialogue is to add a checkbox `With Replacement ` and default unchecked.  (See the Prepare > Data Reshape > Random Subset for this control.)
b) And change the text at the top to become `Permute/Sample Rows ` here too.

Now in the Prepare > Data Reshape > Random Subset.
c) Move Set Seed up so it is in the same position as in the Permute Columns and in the Prepare > Column: Numeric > Random Samples dialogue.
d) At the same time, change the current `Seed` label to become `Set Seed`.

e) Finally in the Prepare > Column: Numeric > Random Samples dialogue add a checkbox with a pull-down control.  It has 7 options and each generate the RNKkind command, it would be before the set seed. 
The default is unchecked and then the command is not given.  If checked then there are 7 options, namely 
default, Mersenne, Wichmann, Marsaglia, Super, Knuth-TAOCP-2002, L'Ecuyer.
f) Add a tooltip "Chooses a different Random Number Generator.  Can usually be ignored."

Vitalis95:
Fixes #7077 
@rdstern @shadrackkibet this PR; 
I have added a with replacement control also sorted reset issue in set seed checkbox.


lloyddewit (approved):

rdstern (approved):
Excellent

PR #7391, opened 2022-04-19
Moved View/Delete labels dialogue from Factor menu to Check Data menu

### LINKED ISSUE #7385: Change the place of the Factor > View/Delete Labels dialogue
The View/Delete labels dialogue is being slightly improved, pull request #7362. 
In writing the Help it also seems to me to be in the wrong place.  It is hidden in the Prepare > Column: Factor menu.
I suggest we move it to the Prepare > Check Data menu, I suggest it goes just below One Variable Frequencies and above the horizontal line there.  I also suggest it is a move, so we delete it from its current place.

Vitalis95:
Fixes #7385 
@rdstern @shadrackkibet this PR moves View/Delete labels dialog from Factor menu to Check Data menu


lloyddewit (approved):

rdstern (approved):
Great

PR #7393, opened 2022-04-20
Fixed refreshing of Selector control when new column added

### LINKED ISSUE #7384: Something seems changed in the filter system?
In the merged version the filter system doesn't seem to work.
a) When I do filter from Factor, say using Village from the survey data, it doesn't work in that none of the villages are then defined as filters - I selected all the villages and it all works fine in Version 0,7.4.
b) When I define a filter, then it does work, and the filter is made correctly.  But it isn't saved as a filter.  So, if I delete the operation of the filter, then it isn't there to redo next time.

Patowhiz:
Fixes #7384

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -40,6 +40,11 @@ Public Class frmMain
     Private WithEvents timer As New System.Windows.Forms.Timer
     Private iAutoSaveDataMilliseconds As Integer
     Private clsDataBook As clsDataBook
+    Public ReadOnly Property DataBook As clsDataBook
+        Get
+            Return clsDataBook
+        End Get
+    End Property
 
     Public strAutoSaveDataFolderPath As String = Path.Combine(Path.GetTempPath, "R-Instat_data_auto_save")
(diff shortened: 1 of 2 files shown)


rdstern (approved):
Great - and a relief too.  I hope that was easy to do.
I did find a potential issue I have not considered before.  I report it here, but it does not affect approving this pull request.
a) I tried defining a factor as a set of filters and that now worked fine.  I made the fi8lter, and unmade it.  I then tried with another factor as well. All good.  With the usual survey data.
b) I also tried a select and that worked fine as well.

Now my concern to examine for the future.  I used a factor with just the levels/labels 0, 1, 2, 3.  So those are now the names of the filters.  Odd names, because you can't have variable names like that.  But ok.  But what happens if we define a different filter with the same names.  Have we allowed for that?

Patowhiz:
> Now my concern to examine for the future. I used a factor with just the levels/labels 0, 1, 2, 3. So those are now the names of the filters. Odd names, because you can't have variable names like that. But ok. But what happens if we define a different filter with the same names. Have we allowed for that?

@rdstern are you asking if we can have 'numeric' filter names?

lloyddewit (approved):

PR #7395, opened 2022-04-21
Improves the use of long datasets with the Regular sequence dialog. 

### LINKED ISSUE #7388: Improve the coding for the Regular sequence dialogue
@anastasia-mbithe perhaps you may like to try this task?

Could you look at the discussion topic #7378.  I hope you will find it interesting as a topic that could perhaps be included in a maths camp? 

Within that topic I found that the regular sequence dialogue is practically unusable for long datasets.  Could you check how it is coded?  I suspect that the problem may be related to the preview, but I don't know.  Perhaps you could start by omitting the preview as a start.  Then, if it is the preview, perhaps that could have a Preview button, so it is not done by default.  

The problems may be elsewhere, instead, or as well.

anastasia-mbithe:
This fixes #7388 

@rdstern, I have added a checkbox that when checked, we have the preview. I created a data frame of 200,000 rows and it worked well. Using 2,000,000 rows, it is well until you check the checkbox. So, I feel for a data frame with about 2,000,000 rows, it's safe having the regular sequence without checking for the preview at the moment. 
But since some people may find the preview useful, maybe we could show the preview for the first 1000 rows. What do you think?

### DIFF
--- a/instat/dlgRegularSequence.vb
+++ b/instat/dlgRegularSequence.vb
@@ -325,15 +325,9 @@ Public Class dlgRegularSequence
     End Sub
 
     Private Sub ucrChkPreview_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkPreview.ControlValueChanged
-        If ucrChkPreview.Checked Then
-            txtPreview.Visible = True
-            lblPreview.Visible = True
-            lblMessage.Visible = True
-        Else
-            txtPreview.Visible = False
-            lblPreview.Visible = False
-            lblMessage.Visible = False
-        End If
+        txtPreview.Visible = ucrChkPreview.Checked
+        lblPreview.Visible = ucrChkPreview.Checked
+        lblMessage.Visible = ucrChkPreview.Checked
         PreviewSequence()
     End Sub
 End Class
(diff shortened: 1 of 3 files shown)


instat/dlgRegularSequence.vb
```diff
@@ -315,4 +324,16 @@ Public Class dlgRegularSequence
...
+    Private Sub ucrChkPreview_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkPreview.ControlValueChanged
+        If ucrChkPreview.Checked Then
+            txtPreview.Visible = True
+            lblPreview.Visible = True
+            lblMessage.Visible = True
+        Else
+            txtPreview.Visible = False
+            lblPreview.Visible = False
+            lblMessage.Visible = False
+        End If
```
lloyddewit, line 336:
```suggestion
        txtPreview.Visible = ucrChkPreview.Checked
        lblPreview.Visible = ucrChkPreview.Checked
        lblMessage.Visible = ucrChkPreview.Checked
```

lloyddewit (changes requested):
@anastasia-mbithe code looks good, thank you
just one small suggestion

lloyddewit (approved):

lloyddewit:

rdstern (approved):
@anastasia-mbithe this seems good now.

PR #7412, opened 2022-04-25
Ensure that parameters can be edited in Random Samples dialog

### LINKED ISSUE #7406: Small correction needed in the Random Samples dialogue
@Vitalis95 this is a really curious one.  
a) With any data set use the random samples dialogue and go to Discrete Empirical.  You don't need to do anything there!
b) Then go to (say) Poisson, which just has a single parameter.  It can't now be edited, because discrete empirical had the first argument as non-editable.  (I think geometric has the same problem!)
c) It is easy to proceed - if you go to normal and then again to Poisson, you are ok.

Vitalis95:
Fixes #7406 
@rdstern this PR makes ucrInputParameter1 to be editable for some distributions.

### DIFF
--- a/instat/ucrDistributionsWithParameters.vb
+++ b/instat/ucrDistributionsWithParameters.vb
@@ -117,7 +117,7 @@ Public Class ucrDistributionsWithParameters
                 ucrInputParameter1.IsReadOnly = False
             End If
         End If
-            OnControlValueChanged()
+        OnControlValueChanged()
     End Sub
 
     Private Sub ucrInputParameter2_ContentsChanged() Handles ucrInputParameter2.ContentsChanged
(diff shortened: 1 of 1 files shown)


instat/ucrDistributionsWithParameters.vb
```diff
@@ -110,20 +110,23 @@ Public Class ucrDistributionsWithParameters
...
-            CheckParametersFilled()
+            If clsCurrDistribution.strNameTag = "Discrete_Empirical" Then
+                ucrInputParameter1.IsReadOnly = True
+                CheckParametersFilled()
+            Else
+                ucrInputParameter1.IsReadOnly = False
+            End If
         End If
-        OnControlValueChanged()
+            OnControlValueChanged()
```
lloyddewit, line 120:
```suggestion
        OnControlValueChanged()
```

lloyddewit (changes requested):

lloyddewit (approved):

rdstern (approved):
@Vitalis95 seems fine to me

PR #7416, opened 2022-04-25
Minor Changes to Column:Text > Transform Dialogue

### LINKED ISSUE #7398: Minor changes in the Prepare > Column: Text > Transform dialogue?
These are just cosmetic changes and I would like discussion, by one or two developers, even though they are trivial to do.  Perhaps @lilyclements could consider with whom?

a) I use this dialogue regularly, but the only option I have used is the last - namely Substring.  We seem to have an informal convention that the initial top radio button is the one on the left.  I am not sure we need that convention?  If you agree, we don't need it, then perhaps we could have Substring as the default.  If we keep to the convention, then perhaps move Substring to be the button on the left?
b) The dialogue is wide, purely because of the size of the top buttons.  Most of the other options are pretty trivial compared to our usual use of this top-radio-button feature in R-Instat.  Could (most of) the other buttons be made smaller, so the dialogue can be made considerably smaller - i.e. not so wide.

derekagorhom:
fixes #7398 

@rdstern @lilyclements 
This is ready for review

### DIFF
--- a/instat/dlgTransformText.vb
+++ b/instat/dlgTransformText.vb
@@ -369,7 +369,7 @@ Public Class dlgTransformText
             grpParameters.Size = New Size(grpParameters.Width, igrpParameterFullHeight / 3.04)
             ucrNewColName.Location = New Point(ucrNewColName.Location.X, iNewColMaxY / 1.39)
             ucrBase.Location = New Point(ucrBase.Location.X, iBaseMaxY / 1.36)
-            Me.Size = New Size(482, iFullHeight / 1.27)
+            Me.Size = New Size(Me.Width, iFullHeight / 1.27)
         ElseIf rdoLength.Checked Then
             grpParameters.Visible = False
             ucrNewColName.Location = New Point(ucrNewColName.Location.X, iNewColMaxY / 1.76)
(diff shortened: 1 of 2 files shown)


instat/dlgTransformText.vb
```diff
@@ -369,7 +369,7 @@ Public Class dlgTransformText
             grpParameters.Size = New Size(grpParameters.Width, igrpParameterFullHeight / 3.04)
             ucrNewColName.Location = New Point(ucrNewColName.Location.X, iNewColMaxY / 1.39)
             ucrBase.Location = New Point(ucrBase.Location.X, iBaseMaxY / 1.36)
-            Me.Size = New Size(Me.Width, iFullHeight / 1.27)
+            Me.Size = New Size(482, iFullHeight / 1.27)
```
lloyddewit, line 372: Do we need to hard-code this value? What's the problem with using `Me.Width`?

lloyddewit (changes requested):

derekagorhom:
Sorry @lloyddewit  i had trouble with reducing the width on the dialogue, i later changed the `Autosize` to `false`. i guess i forgot to change it back to `Me.width` in` line 372`

lloyddewit (approved):

lloyddewit:

rdstern (approved):
Looks fine to me and useful.  Can the code be checked by @lloyddewit or @ChrisMarsh82 so this can be merged?

PR #7417, opened 2022-04-25
In Prepare > Data Object > Rename Data Frame dialog, if original label exists, then include it

### LINKED ISSUE #7350: Small change in Prepare > Data Object > Rename Data Frame
Perhaps @Vitalis95 could make this change quickly as it includes a change to the menu.
I find the word Data Object to be very confusing on the menu.  Our data frames are made into a Data Book.  So I suggest we call this Data Book instead.  I'll check quickly with @volloholic .

But the small change can be seen by comparison with the Prepare > Data Frame > Rename Columns.  There you can also (for a single column) add or rename the Variable label.  If there is a variable label, then the original label is given.  Nice.

In this dialogue the original label isn't given.  That's because we haven't often used them in the past.  Now we are!  So the improvement is simply to include the original label there, when it exists.

This could be done by anyone, and I assume it will be a very quick change.

Vitalis95:
Fixes #7350 
@rdstern @N-thony , this PR adds the  original label to dataframe when it exists

### DIFF
--- a/instat/dlgRenameDataFrame.vb
+++ b/instat/dlgRenameDataFrame.vb
@@ -15,6 +15,7 @@
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 Imports instat.Translations
+Imports RDotNet
 Public Class dlgRenameDataFrame
     Public bFirstLoad As Boolean = True
     Dim strSelectedDataFrame As String = ""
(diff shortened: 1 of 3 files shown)


rdstern (approved):
Looks fine to me

lloyddewit (approved):

Issue #7351, opened 2022-04-03
Deleting key variables should also delete the key


rdstern:
We need key variables for our comments system.  And we have just added the default that making a column from the row variable usually automatically makes it a key variable.  
Slightly separate from this is what happens when we delete variables.  I think nothing happens to the key, so that remains.  I suggest that deleting a variable should check and if it is a key or part of a key, then that key should also be deleted?
@dannyparsons please confirm that this is needed, and isn't there yet?  Then I suggest this is simple for @shadrackkibet but needs his sort of understanding of the R-Instat functions, so he should do it.  Happy for it to be allocated if I am wrong.

rdstern:
@dannyparsons I have realised it is more complicated than this!
![image](https://user-images.githubusercontent.com/11226469/161515882-00597b0a-d180-4cf9-8523-a860f6426a96.png)

a) When I make a second key with different variables, then it appears as field key1 in the data frame metadata.
b) When I delete a key it still remains in the metadata.  So in the figure above I have deleted key from the hh data frame.  
c) In hh I made the second key, because I wanted to use it (rather than the first one) when I added comments.  But it didn't - it still used the first one.  When I deleted the first one, then it used the second one.
d) This has become more important now, because we now (by default) generate a key from the row variable.  

My solutions are as follows - at least for now.
a) When a data frame has multiple keys, then it uses the last one generated in future operations.  Rather than the first?
b) The metadata has a single key variable field.  This either has just the last set of key variables.  Or (like the value labels) it is a list with key=hh1,hh2 ; key1 = hh1, hh2, hh7, etc
c) When a key is deleted, then it is dropped from the data frame metadata.

rdstern:
@N-thony  I think you did this?  Can we close it?

lilyclements:
@N-thony @rdstern if I am not mistaken, this is a duplicate of #8296, in which case I fixed it in #8398

lilyclements:
@N-thony I have fixed the issue that when deleting a variable it should also delete a key if the variable was part of it - as outlined in the linked issue (#8296).
Given the title of this issue, I assumed that was a purpose of this PR. 
@N-thony What in _this_ issue do you think has not been fixed? :)

PR #7423, opened 2022-04-27
Made the change to the metadata to update the Data Frame Metadata grid

N-thony:
Fixes (partially) #7351 
@rdstern @lilyclements I have updated the metadata to be able to drop the key columns when there is no keys.
These changes concern items a) b) and c), please take a look.

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -2228,7 +2228,6 @@ DataSheet$set("public", "add_key", function(col_names, key_name) {
     self$append_to_variables_metadata(col_names, is_key_label, TRUE)
     if(length(private$keys) == 1) self$append_to_variables_metadata(setdiff(self$get_column_names(), col_names), is_key_label, FALSE)
     self$append_to_metadata(is_linkable, TRUE)
-    #self$append_to_metadata(next_default_item(key_label, names(self$get_metadata())), paste(col_names, collapse = ","))
     self$append_to_metadata(key_label, paste(private$keys[[key_name]], collapse = ","))
     cat(paste("Key name:", key_name),
         paste("Key columns:", paste(private$keys[[key_name]], collapse = ", ")),
@@ -2263,8 +2262,6 @@ DataSheet$set("public", "remove_key", function(key_name) {
   self$append_to_variables_metadata(private$keys[[key_name]], is_key_label, FALSE)
   private$keys[[key_name]] <- NULL
   self$append_to_metadata(key_label, NULL)
-  self$append_to_metadata(is_linkable, FALSE)
-  self$append_to_metadata(is_key_label, FALSE)
   cat("Key removed:", key_name)
 }
 )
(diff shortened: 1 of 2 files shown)


N-thony:
> Fixes (partially) #7351 @rdstern @lilyclements I have updated the metadata to be able to drop the key columns when there is no keys. These changes concern items a) b) and c), please take a look.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -2228,7 +2228,8 @@ DataSheet$set("public", "add_key", function(col_names, key_name) {
     self$append_to_variables_metadata(col_names, is_key_label, TRUE)
     if(length(private$keys) == 1) self$append_to_variables_metadata(setdiff(self$get_column_names(), col_names), is_key_label, FALSE)
     self$append_to_metadata(is_linkable, TRUE)
-    self$append_to_metadata(next_default_item(key_label, names(self$get_metadata())), paste(col_names, collapse = ","))
+    #self$append_to_metadata(next_default_item(key_label, names(self$get_metadata())), paste(col_names, collapse = ","))
```
lilyclements, line 2231: Remove this if we don't use it?

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1020,7 +1020,7 @@ DataBook$set("public", "delete_dataframes", function(data_names, delete_graph_bo
 
 DataBook$set("public", "remove_link", function(link_name) {
   if(!link_name %in% names(private$.links)) stop(link_name, " not found.")
-  private$.links[[link_name]] <- NULL
+  #private$.links[[link_name]] <- NULL
```
lilyclements, line 1023: and this?

lilyclements (commented):
@N-thony this looks good. However, there is still a column called "Is_Key" in the data frame metadata that is created when a key creates, but does not remove when the key is removed.

1. Upon opening the survey data I have the following metadata
image

2. I add a key and so the metadata shows the key column

image

3. I remove the key. But we now have a variable "Is_Key" in the metadata
image

instat/static/InstatObject/R/instat_object_R6.R
N-thony: Thanks @lilyclements , I resolved your comments.

lloyddewit (changes requested):
@N-thony There is still an open comment from @lilyclements . If you can resolve this, then I'm happy to approve

N-thony:
> @N-thony There is still an open comment from @lilyclements . If you can resolve this, then I'm happy to approve

@lloyddewit I think all is resolved now.

lilyclements:
Looks good! @lloyddewit @rdstern this is ready for you to review

rdstern (approved):
I'm not sure this is quite right yet, but it is much better than before, so I'm approving.  It looks perfect anyway.
a) I created key automatically in survey data through the Row Numbers/Names
b) Then key 1 from Village + Field.  This replaces the row in the metadata.  Good.  Could it also delete key?  I don't think it does - and that might be all that's needed.  It could even report in the Output Window that key1 defined, key delewted?

lloyddewit (approved):

PR #7419, opened 2022-04-27
Minor changes to the Translations

anastasia-mbithe:
This fixes (partly) #7405

@rdstern and @lloyddewit, this is a minor change to the Translations. It fixes the issue of some controls, on reopening, not remembering the state they were before closing the dialog. The controls fixed are the Method control in Text > Distance dialog and Distribution control in the Fit model > One Variable dialog.
dlgstringdistance

one var

PR #7425, opened 2022-04-27
Renamed Use Graph menu/dialog to Rename Graph

### LINKED ISSUE #7418: Should we rename the use graph menu/dialog to Rename graph?
@rdstern I looked into the **Use Graph** dialog, and the main functionality that makes it different from **View Graph** is it's ability to rename the graph. 

Shouldn't we just name it **Rename Graph**? 
The word use graph to me, meant like I want to use the graph, as it is, for other functions like combine it with others etc.

Vitalis95:
Fixes #7418 
@rdstern  @lloyddewit  this PR renames Use Graph menu and dialog to Rename Graph
 The  PR #7419  makes the database changes


lloyddewit:
@Vitalis95 I merged PR #7419 . Please could you pull the new database into this PR so that @rdstern can test?
Thanks

rdstern (changes requested):
Sorry @vitalis this name needs to revert to Use Graph in both places.
Then we do need to make some changes, because if the current dialogue was confusing to @Patowhiz it will also confuse others.
b) There is just an Options button.  We call it `Plot Options` elsewhere, so make it Plot Options here too.
c) Change the Save Graph to become Save New Graph.
d) Then the problem is that there are no other controls on the dialogue.  In the other dialogues we have controls on the right as well.
So let's add a control on the right.  One that is hidden generally is the legend.  So let's have a group box with the label Legend.  (It is currently hidden under Themes.)  Then there could be 2 controls, first is position and is the same control as exists there already.  Second is Title.  You will have to investigate how to define the legend title - I am not sure..

Vitalis95:
@rdstern  I have added the legend position on dialogue. For the legend title, it has many parameters as shown below and adding them, the dialogue might be large
image

lloyddewit:
@Vitalis95 Is this PR ready for testing again? thanks

rdstern:
@Vitalis95 I am happy with that small change to the dialogue for now.  But it doesn't yet work.  I tried a simple graph and changed the legend option to None - and I still got the legend, just as before.
I have edited the title of this pull request.

Vitalis95:
@rdstern , could you have a look it now.

Vitalis95:
@rdstern , could you have a look it now.

rdstern (approved):
Great.  @lloyddewit if the code is ok, then this would be a good addition to merge.

PR #7427, opened 2022-04-28
Fixed minor bug in Prepare>Data Frame>Rename Columns dialog

### LINKED ISSUE #7422: Minor bug with the Prepare>Data Frame>Rename Columns Dialogue
In the Rename Column Dialogue, under Multiple option, `OK button` seems to still be disabled though the grid is filled as shown in the screenshot below
image

@derekagorhom  can you have a look at it?

derekagorhom:
fixes #7422 , minor bug in Prepare>DataFrame>Rename Column Dialogue 

Enabling the `ok button` for the multiple option 

@africanmathsinitiative/developers  this PR is ready for review

### DIFF
--- a/instat/dlgName.vb
+++ b/instat/dlgName.vb
@@ -231,7 +231,7 @@ Public Class dlgName
                             OrElse Boolean.TryParse(strValue, parsedValue) _
                             OrElse strValue.ToLower.Equals("t") OrElse strValue.ToLower.Equals("f") _
                             OrElse IsNumeric(strValue) Then
-                        MsgBox("The column name must not be a numeric or contain space or french accent or be a boolean e.g TRUE, FALSE, T, F.")
+                        MsgBox("The column name must not be a numeric or French accent or be a boolean e.g TRUE, FALSE, T, F.")
                         bCurrentCell = False
                         Exit For
                     End If
--- a/instat/dlgName.vb
+++ b/instat/dlgName.vb
@@ -240,7 +240,6 @@ Public Class dlgName
             Case 2
                 bCurrentCell = True
         End Select
-        End Select
         TestOKEnabled()
     End Sub
 
--- a/instat/dlgName.vb
+++ b/instat/dlgName.vb
@@ -301,18 +301,6 @@ Public Class dlgName
         Return bFind
     End Function
 
-    Private Function CheckNames(strNewData As String, iColIndex As Integer) As Boolean
-        Dim bCheck As Boolean
-        Dim parsedValue As Boolean
-        If (containsFrench(strNewData) OrElse strNewData.Equals("") OrElse Boolean.TryParse(strNewData, parsedValue) _
-             OrElse strNewData.Equals("t") OrElse strNewData.Equals("T") OrElse strNewData.Equals("f") OrElse strNewData.Equals("F") OrElse IsNumeric(strNewData)) AndAlso iColIndex = 1 Then
-            bCheck = False
-        Else
-            bCheck = True
-        End If
-        Return bCheck
-    End Function
-
     Private Sub RenameColumns(strNewData As String, iRowIndex As Integer, iColIndex As Integer)
         GetVariables(strNewData, iRowIndex + 1, iColIndex)
         TestOKEnabled()
--- a/instat/dlgName.vb
+++ b/instat/dlgName.vb
@@ -137,7 +137,7 @@ Public Class dlgName
         ucrSelectVariables.Reset()
         dctRowsNewNameChanged.Clear()
         dctRowsNewLabelChanged.Clear()
-
+        bCurrentCell = False
         clsNewColNameDataframeFunction.SetRCommand("data.frame")
 
         clsNewLabelDataframeFunction.SetRCommand("data.frame")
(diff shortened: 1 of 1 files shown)


instat/dlgName.vb
```diff
@@ -220,18 +220,24 @@ Public Class dlgName
...
 
+    Private Sub CheckChange(dctName As Dictionary(Of Integer, String), iCol As Integer)
+        For Each value In dctName.Values
+            If Not CheckNames(value, iCol) Then
+                MsgBox("The column name must not be a numeric or contains space or french accent or be a boolean e.g TRUE, FALSE, T, F.")
+                bCurrentCell = False
+                Exit For
+            Else
+                bCurrentCell = True
+            End If
```
lloyddewit, line 231:
```suggestion
            End If
            bCurrentCell = True
```

instat/dlgName.vb
```diff
@@ -220,18 +220,24 @@ Public Class dlgName
         Return strValue
     End Function
 
+    Private Sub CheckChange(dctName As Dictionary(Of Integer, String), iCol As Integer)
+        For Each value In dctName.Values
+            If Not CheckNames(value, iCol) Then
```
lloyddewit, line 225: `CheckNames()` will only return False if `iCol` is 1. 
Therefore, we only need to call `CheckChange()` when `iCol` is 1.
If we only call this function once, then do we need a separate function at all?

instat/dlgName.vb
```diff
@@ -220,18 +220,24 @@ Public Class dlgName
...
-                    MsgBox("The column name must not be a numeric or contains space or french accent or be a boolean e.g TRUE, FALSE, T, F.")
-                    bCurrentCell = False
-                    Exit For
-                Else
-                    bCurrentCell = True
-                End If
-            Next
+            CheckChange(dctRowsNewNameChanged, iColIndex)
+        ElseIf iColIndex = 2 Then
+            CheckChange(dctRowsNewLabelChanged, iColIndex)
```
lloyddewit, line 240: Please see my comment above

lloyddewit (changes requested):

instat/dlgName.vb
derekagorhom:
```
ElseIf iColIndex = 2 Then
            CheckChange(dctRowsNewLabelChanged, iColIndex)
```
was used to enable the `ok buttom` for the multiple and single option in the dialogue but @N-thony  might be able to give a better explanation for why the separate function in  `checkchange()`

lloyddewit: @derekagorhom Could we delete `CheckNames()` and `CheckChange()`, and then rewrite `ValidateNamesFromDictionary() as below?
```
    Private Sub ValidateNamesFromDictionary(iColIndex As Integer)
        'TODO this needs to be implemented in the appropriate ucrControl
        Select Case iColIndex
            Case 1
                For Each strValue In dctRowsNewNameChanged.Values
                    Dim parsedValue As Boolean
                    If String.IsNullOrEmpty(strValue) _
                            OrElse containsFrench(strValue) _
                            OrElse Boolean.TryParse(strValue, parsedValue) _
                            OrElse strValue.ToLower.Equals("t") OrElse strValue.ToLower.Equals("f") _
                            OrElse IsNumeric(strValue) Then
                        MsgBox("The column name must not be a numeric or contain space or french accent or be a boolean e.g TRUE, FALSE, T, F.")
                        bCurrentCell = False
                        Exit For
                    End If
                    bCurrentCell = True
                Next
            Case 2
                bCurrentCell = True
        End Select
        TestOKEnabled()
    End Sub
```
Would that work?

I also noticed that the error message states that spaces are not allowed.
Please could you confirm this?
If spaces are not allowed, then please could you add a check for this?

Thanks

derekagorhom: It seems to work fine like now, i have not experienced the error message you mentioned above. maybe after i add this commit you could retest and confirm.

lloyddewit: @derekagorhom Thank you for the change.
- The `CheckNames()` function is no longer needed. Please could you delete it?
- Regarding the error message, I think there is a misunderstanding. I did not experience the error message during runtime, I just looked at the error message in the code. The error message states that spaces are not allowed. Is this correct (i.e. are spaces in a column name not allowed)? 
Please could you confirm this?
If spaces are not allowed, then please could you add a check for this?

Thanks

derekagorhom: @lloyddewit please i did the changes you requested, are there any more recommendations for this PR....

lloyddewit: @derekagorhom Please expand the `If` statement on line 229 so that it also checks that the column name does not contain any spaces.
Thanks

N-thony:
@derekagorhom is this ready for review?

N-thony:
> Not yet, I still needs the Contains function to check if the column name contains any spaces

@derekagorhom let's discuss on Slack.

derekagorhom:
Hello @lloyddewit please, I think the dialog does not need the contains function. if you check` line` 246 to 257, `ValidateRVariable` ensures labels do not contain spaces after you edit it. So could you test the dialog again, it should work.

instat/dlgName.vb
```diff
@@ -221,18 +221,25 @@ Public Class dlgName
...
+        Select Case iColIndex
+            Case 1
+                For Each strValue In dctRowsNewNameChanged.Values
+                    Dim parsedValue As Boolean
+                    If String.IsNullOrEmpty(strValue) _
+                            OrElse containsFrench(strValue) _
+                            OrElse Boolean.TryParse(strValue, parsedValue) _
+                            OrElse strValue.ToLower.Equals("t") OrElse strValue.ToLower.Equals("f") _
+                            OrElse IsNumeric(strValue) Then
+                        MsgBox("The column name must not be a numeric or contain space or french accent or be a boolean e.g TRUE, FALSE, T, F.")
```
lloyddewit, line 234:
```suggestion
                        MsgBox("The column name must not be a numeric or French accent or be a boolean e.g TRUE, FALSE, T, F.")
```

lloyddewit (changes requested):

lloyddewit (approved):

lloyddewit:
> Hello @lloyddewit please, I think the dialog does not need the contains function. if you check` line` 246 to 257, `ValidateRVariable` ensures labels do not contain spaces after you edit it. So could you test the dialog again, it should work.

Thank you for the explanation.
I wasn't questioning the functionality. The point I was trying to make was that the error check and the error message should be consistent.
If we do not check for spaces in the `If` condition, then the error message should not state that spaces are illegal (even if spaces are replaced with a '.' elsewhere in the code). 
I updated the error message and approved.
I hope this makes sense, thanks

lloyddewit:

rdstern (approved):
seems fine now

Issue #7433, opened 2022-04-30
Problem in Row Numbers /Names dialogue


rdstern:
I used the mrcars example.  In a default way it gave me the error shown

![image](https://user-images.githubusercontent.com/11226469/166100765-be7c3606-2821-423e-a29e-abe470489e4c.png)

N-thony:
@rdstern this is due to the fact that we are missing to pass the data name in the function `data_book$add_key(col_names="row")` which is supposed to be `data_book$add_key(data_name="mtcars", col_names="row")`

N-thony:
@rdstern this issue is related to issue `#7434` and it seems to be more general. @Patowhiz  I suspect that this bug was introduced in either your changes in `PR #7404` or `PR #7432`.
Another thing I found is when you click on `Reset`  in `Row Number /Names` it runs the code for adding the key twice as shown in the screenshot below.
![image](https://user-images.githubusercontent.com/68591383/166145331-d64096d7-fadd-45f0-9a21-541ba63b06a1.png)
@Patowhiz can you have a look of this?

Patowhiz:
@rdstern thanks for finding this bug. 
@N-thony in regards to the selector, it's because of how the dialog is set up. 
I have now factored such set ups in the data frame selector.

Patowhiz:
@N-thony in the dialog, code line 157, `ucrBase.clsRsyntax.AddToAfterCodes(clsAddKeyFunction, iPosition:=0)`, is what's causing the bug that you found in the your comment above.

Basically, the original author of the code anticipated code line 160, `ucrBase.clsRsyntax.RemoveFromAfterCodes(clsAddKeyFunction)`,  will work on reset. But that's not the case.

I have now fixed it in PR #7443

Issue #7438, opened 2022-05-03
Merge (and some other) dialogues don't work now


rdstern:
**Describe the bug**
Program crashes when I use the merge dialogue  This could be caused by the controls changes by @Patowhiz ?  As @N-thony has queried, there may be pull requests to merge that will fix this?

**To Reproduce**
Steps to reproduce the behaviour:
Import from library the 3 data sets from the climdex package (not sure this is needed even, but it is what I did.)
Go to merge.
You don't get the dialogue.  Instead

![image](https://user-images.githubusercontent.com/11226469/166402834-03263ff5-774a-4f73-b2c6-ef4d7e0f3aea.png)

I am checking further and other dialogues also have a problem.  Summary and unstack were ok, but here is what happens with stack - after opening the survey data.
![image](https://user-images.githubusercontent.com/11226469/166403167-62a3ca5f-9338-4955-9d4d-6811079408c3.png)

N-thony:
@rdstern this bug is fixed in PR #7415.

rdstern:
@N-thony and @ChrisMarsh82 and @lloyddewit and I also see I approved last week!  So at last I have found something that I'm ahead of the others!

lloyddewit:
> @rdstern this bug is fixed in PR #7415.

@N-thony I still have an open question in PR #7415. Please can you respond?
(or we can discuss verbally if that's easier)

N-thony:
@lloyddewit as discussed, it is true that PR #7415 fix the issue in the Merge dialogue and it can be merge since it doesn't harm anything else. However, the problem is still there and it will be good to fix from the source. We suspect that the `controlvaluechanged` event is called a bit early as it isn't supposed too.
@Patowhiz hopefully you will look at it and fix it as reported in some others issues.

N-thony:
Same problem with the `Duplicate Column`
![image](https://user-images.githubusercontent.com/68591383/166648301-220c2317-4dc2-496e-8946-5dc63a779abb.png)

PR #7441, opened 2022-05-04
Selector bug fix

### LINKED ISSUE #7437: Corrections needed in the new Tables dialogue
@Ivanluv in the (new combined) summary tables dialogue, I find the problem is likely to be caused by the changes in the selector control, and hence by @Patowhiz changes.

I think the selector control does not refresh when the receiver is changed, nor for other changes, like changing the class of data.  So this dialogue, and presumably others are not usable.

a) When I go to the summary button and the Variables receiver it only allows factors.  That's a change from version 7.4 and we need it to allow variables as I show in Version 0.7.4.
image

compare the above with your new version.
b) If I come back from Summary to Frequency after being in the Variables control no control seems to have the focus now.
c) Please check if there is more!

Patowhiz:
Fixes #7437 and partly issue #7433.

@africanmathsinitiative/developers @rdstern this is now ready for review.


rdstern (dismissed):
@Patowhiz phew.  Looks fine to me so am approving.  @Ivanluv could you also look quickly at the code, then we can perhaps get this merged soon.

Ivanluv (approved):

rdstern:
@dannyparsons could you approve and merge?

Patowhiz:
@rdstern @N-thony can you now test if this PR fixes issue #7433 selector issues. And also the duplicates [comment](https://github.com/africanmathsinitiative/R-Instat/issues/7438#issuecomment-1117061868)

N-thony:
> @rdstern @N-thony can you now test if this PR fixes issue #7433 selector issues. And also the duplicates [comment](https://github.com/africanmathsinitiative/R-Instat/issues/7438#issuecomment-1117061868)

@Patowhiz for the Row Numbers dialogue, it is still adding key code as long as you reset the dialogue
image
and the issue in the Duplicate dialogue is still there
image

Patowhiz:
@N-thony that's slightly a different issue relating to the dialog itself. Basically the dialog is not clearing the after codes on reset.
Let me quickly fix that for you in a separate PR.

rdstern (approved):
@Patowhiz I have checked the tables as before, and also the Row numbers/Names and the Transform dialogue.  All work fine for me.

N-thony:
> @N-thony that's slightly a different issue relating to the dialog itself. Basically the dialog is not clearing the after codes on reset. Let me quickly fix that for you in a separate PR.

@Patowhiz remember, the issue with the `Duplicate Column` is the same with the one we found and fixed with the `Merge Dialogue`, to fix that in the Merge Dialogue we have initiated 'New RFunction` globally in the dialogue but normally this is is done in the `SetDefaults` which is also called in the Loading of the form but currently when you open the dialogue, it is the `ControlValueChanged` that is triggered out first before the `Load` of the form, that is why we are getting that bug.

N-thony:
> > @N-thony that's slightly a different issue relating to the dialog itself. Basically the dialog is not clearing the after codes on reset. Let me quickly fix that for you in a separate PR.
> 
> @Patowhiz remember, the issue with the `Duplicate Column` is the same with the one we found and fixed with the `Merge Dialogue`, to fix that in the Merge Dialogue we have initiated `New RFunction` globally in the dialogue but normally this is is done in the `SetDefaults` which is also called in the Loading of the form but currently when you open the dialogue, it is the `ControlValueChanged` that is triggered out first before the `Load` of the form, that is why we are getting that bug.

Have a look of this, or I can demonstrate to make it more clear.

Patowhiz:
> @Patowhiz remember, the issue with the Duplicate Column is the same with the one we found and fixed with the Merge Dialogue, to fix that in the Merge Dialogue we have initiated New RFunction globally in the dialogue but normally this is is done in the SetDefaults which is also called in the Loading of the form but currently when you open the dialogue, it is the ControlValueChanged that is triggered out first before the Load of the form, that is why we are getting that bug.


@N-thony yes correct. The `ControlValueChanged` will always be called before the `form load` event because the `control load` event is called before the `form load` event. That's why in many dialogs we have the `RFunctions` initilised before the `form load` event, even though the same is still done in the `SetDefaults()`. 

So to fix the bug we just need to always make sure the dialogs don't call on null objects.

Issue #7448, opened 2022-05-04
Mapping problem


rdstern:
@lloyddewit This is (I think) another example of the same error you have fixed elsewhere?

It occurs when you produce a map and then try to add facets - bottom right.

![image](https://user-images.githubusercontent.com/11226469/166685150-7858dd25-d7cd-47df-8efe-6e506d0dd9f7.png)

It crashes out with the code below:

# Code generated by the dialog, Climatic Maps
gadm36_GHA_1 <- data_book$get_data_frame(data_name="gadm36_GHA_1")
GMet_Stn_corrected <- data_book$get_data_frame(data_name="GMet_Stn_corrected")
last_map <- ggplot2::ggplot() + ggplot2::geom_sf(data=gadm36_GHA_1, mapping=ggplot2::aes()) + theme_grey() + ggplot2::geom_point(data=GMet_Stn_corrected, mapping=ggplot2::aes(x=Lon, y=Lat)) + ggplot2::facet_wrap(dir="h", facets=,  ~ Station.Type) + ggrepel::geom_label_repel(data=GMet_Stn_corrected, mapping=ggplot2::aes(x=Lon, y=Lat, label=abbr))
data_book$add_graph(graph_name="last_map", graph=last_map, data_name="gadm36_GHA_1")
data_book$get_graphs(data_name="gadm36_GHA_1", graph_name="last_map")

@Patowhiz has shown me how to get round this and get the required results.  They are without the command, which is ok!

![image](https://user-images.githubusercontent.com/11226469/166685466-433fcf09-e0aa-462f-852d-7610a4ffe44f.png)

I hope this can be fixed, as were the others, but he also plans to (temporarily) catch this situation, if there are other instances, so the don't crash out during the courses.

N-thony:
> @lloyddewit This is (I think) another example of the same error you have fixed elsewhere?
> 
> It occurs when you produce a map and then try to add facets - bottom right.
> 
> ![image](https://user-images.githubusercontent.com/11226469/166685150-7858dd25-d7cd-47df-8efe-6e506d0dd9f7.png)
> 
> It crashes out with the code below:
> 
> # Code generated by the dialog, Climatic Maps
[...]

PR #7454, opened 2022-05-04
Bug fixed in the Mapping Dialogue

N-thony:
Fixes (partially) #7448 

### DIFF
--- a/instat/dlgClimaticStationMaps.vb
+++ b/instat/dlgClimaticStationMaps.vb
@@ -36,14 +36,14 @@ Public Class dlgClimaticStationMaps
     Private bResetSubdialog As Boolean = True
     Private bResetSFLayerSubdialog As Boolean = True
     Private clsLabelRepelAesFunction As New RFunction
-    Private clsCoordPolarFunction As RFunction
-    Private clsCoordPolarStartOperator As ROperator
+    Private clsCoordPolarFunction As New RFunction
+    Private clsCoordPolarStartOperator As New ROperator
     Private clsXScaleDateFunction As New RFunction
     Private clsYScaleDateFunction As New RFunction
     Private clsScaleFillViridisFunction As New RFunction
     Private clsScaleColourViridisFunction As New RFunction
     Private clsAnnotateFunction As New RFunction
-    Private clsGetDataFrame As RFunction
+    Private clsGetDataFrame As New RFunction
     Private clsRemoveFunc As New RFunction
     Private clsParamOperator As New ROperator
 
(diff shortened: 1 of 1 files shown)


lloyddewit (approved):

rdstern (approved):

Patowhiz:
@N-thony @lloyddewit please note, the same fix has been done in PR #7453, which I had to because @rdstern had to test the primary fix in that PR. 

I suggest this to just be closed instead to avoid conflicts. Sorry for any inconvenience. Thanks

Patowhiz:
@N-thony  after reading your  https://github.com/africanmathsinitiative/R-Instat/issues/7448#issuecomment-1117455915 , I think this can be merged then I will resolve the conflicts. Thanks.

Patowhiz:
@rdstern would be very useful for next week's workshop to have this PR merged. Thanks.

N-thony:
> @N-thony @lloyddewit please note, the same fix has been done in PR #7453, which I had to because @rdstern had to test the primary fix in that PR.
> 
> I suggest this to just be closed instead to avoid conflicts. Sorry for any inconvenience. Thanks

@Patowhiz we had 2 two bugs in issue #7448, the first was coming due to the new selector which you fixed by adding `New` the second is from the `R code` which was causing R-Instat to crash, you fixed the first and not the second, but you fixed the crash but the problem with the R code is still there, and this PR fixes the two bugs.
Your fix is more general to avoid R-Instat to crush in such problem, and this fix is concerns only this dialogue.
@lloyddewit since @Patowhiz fixes is more general, I would prefer to remove `dlgClimaticStationMaps.vb` in PR #7453 to avoid conflict.

lloyddewit:
> > @N-thony @lloyddewit please note, the same fix has been done in PR #7453, which I had to because @rdstern had to test the primary fix in that PR.
> > I suggest this to just be closed instead to avoid conflicts. Sorry for any inconvenience. Thanks
> 
> @Patowhiz we had 2 two bugs in issue #7448, the first was coming due to the new selector which you fixed by adding `New` the second is from the `R code` which was causing R-Instat to crash, you fixed the first and not the second, but you fixed the crash but the problem with the R code is still there, and this PR fixes the two bugs. Your fix is more general to avoid R-Instat to crush in such problem, and this fix is concerns only this dialogue. @lloyddewit since @Patowhiz fixes is more general, I would prefer to remove `dlgClimaticStationMaps.vb` in PR #7453 to avoid conflict.

@N-thony I looked at the changes to `dlgClimaticStationMaps.vb` in PR #7453, and I see that there are some overlapping changes. In this case, I think GitHub is smart enough to resolve the merge automatically, so there won't be merge conflicts. So I will go ahead and merge this PR.

PR #7456, opened 2022-05-04
Creating New RFunction instance when declaring private class member in the dialogue to prevent exception when event handled before `SetDefaults()` called

### LINKED ISSUE #7438: Merge (and some other) dialogues don't work now

N-thony:
Fixes #7438
@Patowhiz as agreed, I added keyword new in the declaration of `Rfunction` in each dialog to avoid bugs with the new `Selector`.
@lloyddewit take a look.

### DIFF
--- a/instat/dlgAddComment.vb
+++ b/instat/dlgAddComment.vb
@@ -22,7 +22,7 @@ Public Class dlgAddComment
     Private strSelectedColumn As String = ""
     Public bFirstLoad As Boolean = True
     Private bReset As Boolean = True
-    Private clsAddComment As RFunction
+    Private clsAddComment As New RFunction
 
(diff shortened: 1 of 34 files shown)


lloyddewit:
@rdstern I don't think there is anything you can test here and I am confident that these changes will not cause regression (famous last words!), so I will merge.

PR #7376, opened 2022-04-14
Random split dialog

mmulahvictor:
I have added the new Random Split dialog to the form main.
@shadrackkibet and @lloyddewit can you merge this for @rdstern to review PR #7272.

PR #7470, opened 2022-05-15
Adding the random split dialog to form main

mmulahvictor:
I have added the new Random Split dialog to the form main.

This PR is replacing #7376 

@shadrackkibet and @lloyddewit can you merge this for @rdstern to review PR https://github.com/africanmathsinitiative/R-Instat/pull/7272.

### DIFF
--- a/instat/frmMain.Designer.vb
+++ b/instat/frmMain.Designer.vb
@@ -458,6 +458,7 @@ Partial Class frmMain
         Me.mnuPrepareCheckDataOneVariableSummarise = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuPrepareCheckDataOneVariableGraph = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuPrepareCheckDataOneWayFrequencies = New System.Windows.Forms.ToolStripMenuItem()
+        Me.mnuPrepareCheckDataViewDeleteLabels = New System.Windows.Forms.ToolStripMenuItem()
         Me.ToolStripSeparator41 = New System.Windows.Forms.ToolStripSeparator()
         Me.mnuPrepareCheckDataExportOpenRefine = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuPrepareCheckDataImportOpenRefine = New System.Windows.Forms.ToolStripMenuItem()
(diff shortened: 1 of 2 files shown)


lloyddewit (changes requested):
Thanks, this looks good.
The only issue is the `.RHistory` file. I don't think that this should be in this PR, please could you remove it?

lloyddewit:
@rdstern This PR should open an empty dialog. Please could you check that the menus and tool strip are as you expect? We can then merge, and you can then test PR #7272 properly.
Thanks!

instat/frmMain.Designer.vb
```diff
@@ -5007,6 +5014,7 @@ Partial Class frmMain
         Me.ucrDataViewer.BackColor = System.Drawing.SystemColors.Control
         Me.ucrDataViewer.BorderStyle = System.Windows.Forms.BorderStyle.FixedSingle
         Me.ucrDataViewer.Dock = System.Windows.Forms.DockStyle.Fill
+        Me.ucrDataViewer.grdCurrSheet = Nothing
```
lloyddewit, line 5017: This line triggers a build error

image

instat/frmMain.Designer.vb
```diff
@@ -5067,11 +5075,11 @@ Partial Class frmMain
...
         '
-        'mnuPrepareCheckDataViewDeleteLabels
+        'mnuPrepareDataReshapeRandomSplit
         '
-        Me.mnuPrepareCheckDataViewDeleteLabels.Name = "mnuPrepareCheckDataViewDeleteLabels"
-        Me.mnuPrepareCheckDataViewDeleteLabels.Size = New System.Drawing.Size(245, 22)
-        Me.mnuPrepareCheckDataViewDeleteLabels.Text = "View/Delete Labels..."
+        Me.mnuPrepareDataReshapeRandomSplit.Name = "mnuPrepareDataReshapeRandomSplit"
+        Me.mnuPrepareDataReshapeRandomSplit.Size = New System.Drawing.Size(197, 22)
+        Me.mnuPrepareDataReshapeRandomSplit.Text = "Random Split..."
```
lloyddewit, line 5082: If the user selects clicks this menu option, then it will open an empty dialog box.
This is probably OK because we expect the real dialog to be added before the next release.
However, we should normally only merge 'releasable' chenges to the main branch.
In this PR, please could you set this menu option's visible attribute to false so that the user cannot see it?
When you merge these changes into PR #7272 , you can set 'visible' back to True.

lloyddewit (changes requested):

rdstern:
@lloyddewit sounds good

mmulahvictor:
@lloyddewit can you see if it is working correctly now?

lloyddewit (approved):
@mmulahvictor Thank you, this is good.
I noticed that you set 'enabled' to false. It's probably slightly cleaner to set 'visible' to false instead but this isn't so important.
I'm happy to approve and merge.

PR #7471, opened 2022-05-16
Ignore column selection when adding a new variable 

### LINKED ISSUE #7202: Improve the Position control to allow for selects
You notice I am using the noun here.  "a select" is to columns what "a filter" is to rows.  And a feature of R-Instat is that they are objects in the data book.
When a select is active you only see those variables in the grid and in the data selector.  Many dialogues have a Position control.  That dictates the position of further column(s).
Two options in the Position control are `Start` and `End`.  I propose that, when a select is in operation, they become the Start and End of the selection?  I assume that will need a change to the code.
More important is their role in the select?  I suggest the default is for them to be added to the current select?  Otherwise the results of a calculation won't be visible in the selected data frame, or in the data selector.  And it would then be a pain to do a further calculation on that variable.    
Whether we need an option in the Position control to not add to the selection I am not sure.  But my claim is that selects need to be dynamic.  
I am assigning @shadrackkibet because he needs to assess exactly what should be done.  Then I assume he will allocate the work - probably to Patrick, as he wrote the Position control.
(I also feel we are sort of on the right track with the current set of user controls.  If we hadn't a position button/control then we would have to produce it now, - with the changed role of start and end.)
And this sort of thing is also showing that select is one of our ways of "taming" the wide data sets.  I know there is a lot needed to improve the data selector as well, but that doesn't diminish the importance of the select feature.

shadrackkibet:
Fixes #7202

@rdstern here is the "temporary" fix to the bug adding a new variable when the column selection is active. 

@N-thony you may wish to test as well and proceed from here?

For example, It will be nice to alert the user in the output windows that a column selection has been ignored.


N-thony:
> Fixes #7202
> 
> @rdstern here is the "temporary" fix to the bug adding a new variable when the column selection is active.
> 
> @N-thony you may wish to test as well and proceed from here?
> 
> For example, It will be nice to alert the user in the output windows that a column selection has been ignored.

@shadrackkibet thanks, I will merge this in my branch and proceed.

rdstern (changes requested):
@shadrackkibet and @N-thony this already seems to work, which is great!
a) I tried with the usual survey data Prepare > Row Names/Numbers and it seems fine.  (I selected the first 3 variables plus yield)
b) I then tried with the Prepare > Calculations and found some oddities to examine.  Here is what I did.
1) Calculated Yield * 2, and saved it in calc
2) Repeated that, and put the position so it saves before the Yield.
Then I removed the selection.  I found
a) The resulting variable calc was only produced once.  I had expected calc and calc1
b) It was in a funny place - so the position option is working oddly.
c) It had many decimals and so is not quite the same as when there was no selection.  So compare calc and calc1 (done after removing the selection) below.
image

rdstern:
@shadrackkibet I have now found a surprise, but a possible excellent feature that works already!  
a) I did a select again with "Variables starting with y", ignoring case.  Plus then the first 2 variables.
b) I used calc again and saved the results in yield4, etc
Now they do appear in the selection - I assume you are taking off the selection, and then applying it again?  So the new variable starts with y and appears.  Nice!  Still in the wrong position, but also calculated perfectly this time!


I like that!!!

image

shadrackkibet:
@rdstern I made a small change that might be worth checking. Let me know if this works!

rdstern (changes requested):
@shadrackkibet I think that's just about "nailed" it.  I tried with the survey data, and selected 3 variables including yield.  I also add a starts-with and added x, even though there were no current variables starting with x!
a) I first added yield*2 with the default and then yield*4 putting it at the start.  They were in calc and calc1, and so did not appear - correctly.  So far so good.
b)  I took off the filter and they were in the correct positions.  Great.  The only problem is that they did not have a significant figures attached, and so looked like character variables with loads of decimals.  I attached them manually, but good if they could be given with the default automatically.
c) I then did the select again, and produced x1 at the start and x2 before the yield.  They both appeared at the end of the selected variables.  If that part of the positioning is tricky to fix, then I could live with that.  When I took off the selection, they were correctly positioned, which is great.  
d) Aha, I have found another little problem.  I want to define a new selection.  But as you see below, I only have the variables available from the previous selection.
image

shadrackkibet:
@rdstern I think the significant figures are now updated correctly.

@N-thony (d) is an issue with the selector not updating variables correctly. This should be easy to fix at the dialog level. Would you like to give it a try?

rdstern (approved):
@shadrackkibet this now seems to work well.  
I tried again with the survey data and made a selection including the yields - just used Columns.  Then new variables don't appear.
I then selected again and included Starts with c.  Now a calculated variable appears as the last column - independent of the settings of the position.  But it is in the overall correct position once the selection is removed.
I then tried adding the row variable.  This appears when you add starts with r.  Again at the end, but correctly positioned once the selection is removed.
Good improvement to our "system".

If @ChrisMarsh82 is ok, then this can be merged.

PR #7474, opened 2022-05-16
In Prepare > Add (Merge) Columns dialog, ensure that data frames with different key column names can still be merged

### LINKED ISSUE #7473: Minor bug with the Prepare > Add (Merge) Columns Dialogue
This is also found during the KMD workshop, the bug appears when the keys columns have different name with `Add (Merge) Columns Dialogue`, but it doesn't appear when using the general `Merge dialogue`.
image

N-thony:
Fixes #7473
@rdstern I have fixed the bug that occurs when the key columns have two different name when trying to merge. For the Joining columns option sub dialogue as you suggested that the selectors should reflect the selectors in the main dialogue, this can be done in a sepearate PR.

### DIFF
--- a/instat/dlgMergeAdditionalData.vb
+++ b/instat/dlgMergeAdditionalData.vb
@@ -236,7 +236,7 @@ Public Class dlgMergeAdditionalData
         Dim clsAnyDuplicatesFunction As New RFunction
         Dim iAnyDuplicated As Integer
 
-        If ucrInputMergingBy.GetText <> "" Then
+        If ucrInputMergingBy.GetText <> "" AndAlso lstJoinColumns.Count > 0 Then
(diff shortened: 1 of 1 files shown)


lloyddewit (approved):

lloyddewit:

rdstern (changes requested):
I can get the following error:
image

I used the Niger rds data with both the stations and the data.  I then changed the name in the Stations file to Station.  I then used the dialogue.  When I changed the second file to Data then it gave the error message shown.

I persisted and it made the change ok.

I notice the two files are still "the wrong way round" - perhaps  - in the sub-dialogue, but would be happy to merge without fixing that, once there are no error messages.

N-thony:
@rdstern I have resolved this minor bug.

rdstern (approved):
Great.  Now seems ok.
@lloyddewit there has been a small change just now.

lloyddewit (approved):

PR #7478, opened 2022-05-16
Added function and control to scale/distance dialog

### LINKED ISSUE #7436: Improving the Scale/Distance dialogue?
This is working fine.  But I am now interested in using [this little summary](https://www.statology.org/dist-function-in-r/) to calculate the help.  It gives different answers to our dialogue.

I have now realised that it is calculating the distances between the variables (columns) and we calculating it between the rows.  You can see this by trying the example and adding another row of data.

I tried to make the change, but the R code is (to me) not ideal, because the main calculations are in a single line.  
a) Could the getting of the data be separated from the use of the dist function.
b) Could we then use the transpose of the matrix, so we get the distances between the columns instead.
c) I hope the resulting data frame will then be labelled with the column names as in the example?
d) I suggest we don't have both, but hope that the R code (in the script window) would make it obvious what needs to change to give the current distances of the rows.

Vitalis95:
Fixes #7436 
@rdstern @N-thony @lloyddewit , this PR
-Separates getting of data  from the use of the dist function.
-transposes the data, so that we get the distances between the columns instead of rows
-Adds `Power option` control
-Sorts reset issue on `whole data frame` button and aligning controls


lloyddewit:
@N-thony Are you ok to peer review?

N-thony:
> @N-thony Are you ok to peer review?

@lloyddewit yes,

N-thony:
@Vitalis95 can you resolve my comment above? Then we can move to the next step?

N-thony (changes requested):

rdstern (changes requested):
@Vitalis95 this seems to work well now.  I think the scale was ok previously, so I was examining the distance option:
a) The word matrix on the left seems odd.  If I am right, then could you change the label to Save Result: (Just like the calculator dialogue)
b) The Power option seems to be visible all the time.  I think it only applies to the last option.
c) See below:
image
It seems here to be including the power function with the Euclidean distance.  I am not sure this makes any sense.  I don't think it is doing any harm, but it should presumably not be there?

Vitalis95:
@rdstern  have a look at it.

Vitalis95:
@rdstern  I have made the Save control invisible when the checkbox is unchecked.

rdstern (dismissed):
Great.  If @ChrisMarsh82 is ok, then this can be merged

N-thony (approved):

shadrackkibet:
@Vitalis95 I have a few points to report while testing this.

1) Why do you need the method Combobox editable? I think it should be non-editable because we only need specific methods listed in the Combobox. I understand we need to search through the drop-down but the list isn't much therefore I do not see the benefit of making this Combobox editable
image
2) The top radio buttons are not centred. Use vscode tools to centre horizontally.
3) I don't understand why you need an extra checkbox to hide/show the save control. Isn't it just ` ucrSaveDistance.SetCheckBoxText("New Data Frame Name:")`?
 4) You have two save controls in this dialog, which save a data frame. I think one should be enough, all you need is to set additional parameter pair to the other function i.e `ucrSave.AddAdditionalRCode(clsFunction, iAdditionalPairNo:=1)`

5) Tab order!

N-thony:
For item 5., in Visual Studio, there is an easy way to handle the Tab Order, see screenshots below
On the designer, click View > Tab Order
image
and then, by clicking on each number in the small square, you can make the order in the ascending order
image

Vitalis95:
> @Vitalis95 I have a few points to report while testing this.
> 
> 1. Why do you need the method Combobox editable? I think it should be non-editable because we only need specific methods listed in the Combobox. I understand we need to search through the drop-down but the list isn't much therefore I do not see the benefit of making this Combobox editable
>    image
> 2. The top radio buttons are not centred. Use vscode tools to centre horizontally.
> 3. I don't understand why you need an extra checkbox to hide/show the save control. Isn't it just ` ucrSaveDistance.SetCheckBoxText("New Data Frame Name:")`?
> 4. You have two save controls in this dialog, which save a data frame. I think one should be enough, all you need is to set additional parameter pair to the other function i.e `ucrSave.AddAdditionalRCode(clsFunction, iAdditionalPairNo:=1)`
> 5. Tab order!

@shadrackkibet , that checkbox is to convert the data to matrix (`as.matrix`), initially it was called `Matrix` but @rdstern  suggested I rename it has `Save Result` .so when its unchecked the save control should be hidden.

rdstern:
@Vitalis95 I am happy with @shadrackkibet suggestion here.  I don't see examples of New Data Frame Name, with a checkbox, but we often have `Save Result` with a checkbox, e.g. in Principal Components.  So, should we change the label to Save Result and just have the one control?

Vitalis95:
@rdstern @shadrackkibet . Have a look at it

rdstern:
@Vitalis95 and @shadrackkibet I am a bit concerned that we may be "going round in circles" here.  This now saves the objects well, but it used to (also) have the option of producing another data frame with the results.  That seems to have gone now?  Have I missed it, or how do I now get a data frame with the distances?

Vitalis95:
@rdstern , Now it  presents the distances in three ways the way I did in correlations i.e in the output, in a new data frame, and as an object
Have a look at it.

rdstern (approved):
@Vitalis95 I am approving this, though the saving still isn't quite right I think.  @shadrackkibet may suggest improvements to the naming of the new data frames and the distance objects, before merging.  But the naming conventions also need to be consistent with those for the new correlations, and with the (to be) improved PCAs.  So, I suggest we look at the naming again, once we have at least PCA, cluster, correlations and distances sorted.

shadrackkibet:
@Vitalis95 I think the confusion here is because of the checkbox "Display as data frame"

I claim that this checkbox should be removed then basically the save control saves the data frame when you check the checkbox and when unchecked this is printed in the output window.  

Check the calculator for a similar behaviour although that saves the column into the data frame, that should be a similar behaviour to the save control.

**Save Result** as the name of the Save control is sufficient for now.

image

Vitalis95:
> @Vitalis95 I think the confusion here is because of the checkbox "Display as data frame"
> 
> I claim that this checkbox should be removed then basically the save control saves the data frame when you check the checkbox and when unchecked this is printed in the output window.
> 
> Check the calculator for a similar behaviour although that saves the column into the data frame, that should be a similar behaviour to the save control.
> 
> **Save Result** as the name of the Save control is sufficient for now.
> 
> image

@rdstern , from our discussion last week  you suggested the distance to be displayed in 3ways; output, in a new data frame, and as an object like the correlations dialog also when `Save Result` is unchecked it saves as `last_distance` when checked saves as `distance 1,2...` for both the object and new dataframe.
@shadrackkibet if I go as per the above, will it still save the object? As it is now, `Display as data frame` when checked it converts the distance to matrix( `as.matrix`)  and `SetSaveTypeAsDataFrame()`  but when unchecked it `SetSaveTypeAsModel()` to save the object and displayed it in output window

rdstern:
@lloyddewit it would be quite good to include this copy, in this version, even with the discussion on the 3 ways of displaying the results.  It works and we could add issues later.

instat/dlgCluster.vb
```diff
@@ -155,12 +182,56 @@ Public Class dlgCluster
...
+                            ucrBase.OKEnabled(True)
+                        End If
+                    Else
+                        ucrBase.OKEnabled(False)
+                    End If
+                End If
+            End If
         Else
-            ucrBase.OKEnabled(True)
+            If rdoDistanceData.Checked Then
```
lloyddewit, line 210: make this block more concise and readable?
see example above

lloyddewit:
@rdstern The updates are related to whether the OK button is enabled or not. If you can test/approve, then this PR is ready to merge.

rdstern (approved):
Great

PR #7482, opened 2022-05-18
Bug fixed: File > Close Data Book

### LINKED ISSUE #7475: Minor Bug with File > Close Data Book
During the KMD workshop, we noticed that File > Close Data Book seems to not work in the following cases.

1. When the Filter is active and try to `Close Data Book` and it seems to not work
2. After saving a dataset with `File > Save As > Save Data As` then try to `Close Data Book` same as 1), it doesn't close the databook
image

N-thony:
Fixes #7475
@lloyddewit as discussed this bug was due to the fact the check when closing the databook was not done correctly, please take a look.

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -71,7 +71,7 @@ Public Class frmMain
     ''' it's set to false by ucrDataView control when state of data has been changed
     ''' it's set to true by dlgSaveAs dialog and save menu when data has been successfully saved 
     ''' </summary>
-    Public Property bDataSaved As Boolean = False
+    Public Property bDataSaved As Boolean
 
     Private strCurrLang As String
     Public Sub New()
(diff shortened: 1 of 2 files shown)


instat/frmMain.vb
```diff
@@ -71,7 +71,7 @@ Public Class frmMain
     ''' it's set to false by ucrDataView control when state of data has been changed
     ''' it's set to true by dlgSaveAs dialog and save menu when data has been successfully saved 
     ''' </summary>
-    Public Property bDataSaved As Boolean = False
+    Public Property bDataSaved As Boolean
```
lloyddewit, line 74: In VB dotnet, the default value for a boolean is False. 
Removing the assignment does not change the behaviour but it makes the code less readable.
Undo this change?

instat/frmMain.vb
```diff
@@ -1566,13 +1566,13 @@ Public Class frmMain
...
-                strSaveFilePath = ""
-            End If
+        If (Not bDataSaved AndAlso ucrDataViewer.GetWorkSheetCount() > 0 _
+                            AndAlso DialogResult.Yes = MsgBox("Are you sure you want to close your data?" &
+                                    Environment.NewLine & "Any unsaved changes will be lost.",
+                                    MessageBoxButtons.YesNo, "Close Data")) _
+                            OrElse bDataSaved Then
+            clsRLink.CloseData()
+            strSaveFilePath = ""
         End If
```
lloyddewit, line 1576: Is something like this more readable?

```suggestion
        If ucrDataViewer.GetWorkSheetCount() <= 0 Then
            Exit Sub
        End If

        If Not bDataSaved _
                AndAlso DialogResult.No = MsgBox("Are you sure you want to close your data?" &
                                    Environment.NewLine & "Any unsaved changes will be lost.",
                                    MessageBoxButtons.YesNo, "Close Data") Then
            Exit Sub
        End If

        clsRLink.CloseData()
```

lloyddewit (changes requested):

N-thony:
Thanks @lloyddewit, I resolved your comment.

lloyddewit:

rdstern (changes requested):
@N-thony it now works - but perhaps too well!
The previous Close - when it worked - also gave the message:
image

Now it seems to close fine, but without the message.  I rather liked the warning we had before.

N-thony:
> @N-thony it now works - but perhaps too well! The previous Close - when it worked - also gave the message: image
> 
> Now it seems to close fine, but without the message. I rather liked the warning we had before.

@rdstern the message is still there but only when the data are not saved, otherwise it will close without a message.
Do we still need to give a message even if the data are already saved and there is no changes?

rdstern (approved):
Good response @N-thony and you are correct!  Well done.  That's better!

Issue #7304, opened 2022-03-15
Help needed, plus serious bug on Find Replace of text variables


rdstern:
This messy dataset is supplied with the R rrefine package.  It may have to be installed.  It includes a messy data set, (together with a tidy version)
I would like to show how these data can be tidied in R-Instat.  I need some help on using regex (which is one way of tidying.  This could usefully involve one of the documentation team, maybe @Kelvin-23  as it is good stuff for a video.  I also found a serious bug here for @Patowhiz and @N-thony to investigate. 

![image](https://user-images.githubusercontent.com/11226469/158357530-973c8e26-5df2-42b5-b5eb-06939c792b7f.png)
 
The serious bug first.  If you type illegal regex, then the error isn't trapped in any way.  It just throws you out of the software.  What can be done about this?

rdstern:
@lilyclements after asking various people we think this is a puzzle you may be able to diagnose quite quickly.

With the data above I think there may be 2 slightly different things needed.  Both could usefully illustrate how easily we can tidy the data above (from rrefine package).  Perhaps initially how easily we could tidy the last variable:

a) In find/replace I would like to find something in a string, e.g. y and then replace the whole string by the replacement.  Currently I don't think I have that option, i.e. it replaces that part of the string.  The rest of the string is still included.  That is often useful, but how (in stringr) do I replace the whole string?
b) Then what is sensible regex for the task of locating the right find option?  And I wonder if our keyboard is sufficiently helpful?

lilyclements:
To (a):

```
string <- c("heya", "hiya", "hi", "hello", "hey")
string[str_detect(string, "\\y")] <- "Yes"
string
```

Although this isn't the "tidyverse" way to achieve this. We can achieve it with a `case_when` (or `ifelse`) option.
However, we then have an issue with factors. So have to convert the variable to a character (and then convert it back). We can keep the levels in the correct order by the `var_levels <- levels(...)` line

```
library(rrefine)
data(lateformeeting)

var_levels <- levels(lateformeeting$was.i.on.time.for.work)
lateformeeting %>%
    mutate(was.i.on.time.for.work = as.character(was.i.on.time.for.work)) %>%
    mutate(was.i.on.time.for.work = case_when(
      str_detect(was.i.on.time.for.work, "Y") ~ "Yes",
      str_detect(was.i.on.time.for.work, "N")  ~ "No",
      TRUE ~ was.i.on.time.for.work)) %>%
[...]

PR #7484, opened 2022-05-19
Added three Replace options in the Find/Replace dialog

anastasia-mbithe:
Fixes (partially) #7304 
This is still in progress. I am researching the function for the third option, replacing entire cell contents like it's done in Excel.


rdstern (changes requested):
I assume you have yet to implement the replace cell.  The radio buttons look fine.
Perhaps check the code to update the default name of the replace cell option.  If I go to the replace options it seems ok.  Then if I go to the cell option, then back to detect (radio button at the top) and then back to replace, then it still has the name detect.

anastasia-mbithe:
@rdstern, I found a function to replace the cell, you can test it now. I also fixed the bug you referred to in your earlier comment.
It is ready for review.

rdstern (dismissed):
Looks good.  I hope the code is also ok so it can be merged soon?

lloyddewit (approved):
@anastasia-mbithe Looks great, thank you.
I just renamed one of the variables.

PR #7490, opened 2022-05-20
Added Integer Keyboard to the Calculator

### LINKED ISSUE #7483: Add new arithmetic keyboard to the calculator
This keyboard will take some designing - 
a) I assume the keyboard could be called Arithmetic.
b) The top-left button could be called bigq and would produce gmp::as.bigq( ).
c) the R-Help button will call the gmp package.
d) Using this key will then be combined with the ordinary +, - , * / and power keys from the main keyboard.  I wonder about a small change to those keys (except ^ to have a space either side.  Perhaps @lilyclements can confirm.
e) Another general change I suggest is for the default to have the Save Result checkbox ticked by default.  It is what we often want.
Then we should investigate other keys on thew new keyboard (from the gmp package) and perhaps related?  I like prime which implements the gmp::is.prime( ) function and gmp::factorial

anastasia-mbithe:
Fixes #7483 
This is still in progress.


rdstern (changes requested):
@anastasia-mbithe that's a good start.
a) I think call it` Integer`, rather than arithmetic.
b) Change Prime as the label to `prime`.  And get rid of the dash in the function- it doesn't work currently.
c) Add the factorizeZ key and the chooseZ functions, as `factorial` and `choose` keys.
d) add a nextprime key. 
That should be quick - next some puzzles: 
e) Now some puzzles for you - not so simple, and each to be considered only if there is a reasonably easy solution.
1) I really like factorize, but it seems to work for just a single number, rather than a column of numbers.  And it gives the result in a short column, which then get repeated down the variable. Could you get it to produce the results in a single list each time, and to work for a variable, rather than a number?  
2) And a fib key would be nice.  Again that works only for a single number, while fib2 gives 2 Fibonacci numbers.  Could you add sapply as described under the Eulerian in the guide?
3) And look for more keys to add from the package

lloyddewit:
@anastasia-mbithe Thank you for this, is this PR ready again for test/review or is it still in progress?
Thanks

anastasia-mbithe:
Hello @lloyddewit, This is still in progress.

rdstern:
@anastasia-mbithe this is starting to look good.  And well done on using the Desctools package.
a) I know the keyboard is called integer, but is there any advantage using bigz, rather than bigq as the first function?  If so, then perhaps we include both?  But we may want to use the function on non-integer data, and bigz truncates, while bigq seems fine.  I tried on the fert column in the survey data.
b) Could you and @Wycklife add tooltips to the keys?
c) Then I am happy to approve.  But could you also, between you, find any other keys that could be useful to add?  I am particularly thinking of interesting things that would be fun in a maths camp.

anastasia-mbithe:
@rdstern , I have added some other functions as Fibonacci(), Divisors(), PercentRank() and DigitSum() from the DescTools package. I have also added the tooltips and bigq function. 
Have a look at it.

rdstern:
Sounds good.  Can you add an example of each that I can use.  Do they all work with columns of data, or are some simpler than that?

anastasia-mbithe:
> Fibonacci(c(1,2,5,7,8,10))
[1]  1  1  5 13 21 55
> Divisors(c(4,8,21,45))
$`4`
[1] 1 2

$`8`
[1] 1 2 4

$`21`
[1] 1 3 7

$`45`
[1]  1  3  5  9 15

> PercentRank(c(1,2,5,6,7))
[1] 0.2 0.4 0.6 0.8 1.0
> DigitSum(c(234,567,12345,987))
[1]  9 18 15 24

Yeah they all can work with a number, a list and even a column.

PR #7492, opened 2022-05-23
Fixed bug in the Start of Rains Dialog

### LINKED ISSUE #7488: Bug in the Start of Rains Dialogue
We encountered this bug during the GMet workshop, this bug is found in the `0.7.5` version of R-instat but not in the older version (`0.7.4` seems to work fine)
Start of Rains Bug

This bug occurred when we used the` Start of Rain` dialogue
1. Go to 'Climatic'
2. Click on 'Prepare'
3. Scroll down and select 'Start of Rains'
4. input the necessary Variables and conditions
5. The error is in the above the screenshot
Screenshot (427)

The R commands for this dialogue in the `0.7.5` version seem to work fine for the `0.7.4` version.

N-thony:
Fixes #7488 
@lilyclements thanks for your support, @rdstern I had a quick look of it then requested help from Lily, this seems to be fixed now, please take a look.
@derekagorhom @VimbayiChimonyo could you also test?
@dannyparsons could you review this? Thanks.

### DIFF
--- a/instat/static/InstatObject/R/Backend_Components/calculations.R
+++ b/instat/static/InstatObject/R/Backend_Components/calculations.R
@@ -476,38 +476,42 @@ DataBook$set("public", "apply_instat_calculation", function(calc, curr_data_list
     # note: important that there is *no* space between | for grepl function
     # and important there IS a psace in str_detect..!
     
-    # get the data type of the column
-    col_data_type <- self$get_variables_metadata(data_name = calc_from_data_name, column = col_name, property = "class")
-    # if it is a ordered factor...
-    if (any(stringr::str_detect("ordered", col_data_type))){
-      # put in here the ones that DO work for ordered factor
-      if (any(grepl("summary_count_non_missing|summary_count_missing|summary_count|summary_min|summary_max|summary_range", formula_fn_exp))){
-        curr_data_list[[c_data_label]] <- curr_data_list[[c_data_label]] %>%
-          dplyr::summarise_(.dots = setNames(list(as.formula(paste0("~", calc$function_exp))), calc$result_name))
+    if (exists(deparse(substitute(col_name)), parent.frame())){
+      # get the data type of the column
+      col_data_type <- self$get_variables_metadata(data_name = calc_from_data_name, column = col_name, property = "class")
+      # if it is a ordered factor...
+      if (any(stringr::str_detect("ordered", col_data_type))){
+        # put in here the ones that DO work for ordered factor
+        if (any(grepl("summary_count_non_missing|summary_count_missing|summary_count|summary_min|summary_max|summary_range", formula_fn_exp))){
+          curr_data_list[[c_data_label]] <- curr_data_list[[c_data_label]] %>%
+            dplyr::summarise_(.dots = setNames(list(as.formula(paste0("~", calc$function_exp))), calc$result_name))
+        } else {
+          curr_data_list[[c_data_label]] <- curr_data_list[[c_data_label]] %>%
+            dplyr::summarise_(.dots = setNames(list(NA), calc$result_name))
+        }
+        # if it is a factor or character, do not work for anything except...
+      } else if (any(stringr::str_detect("factor | character", col_data_type))){
+        # put in here the ones that DO work for factor or character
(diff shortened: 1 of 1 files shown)


rdstern (dismissed):
@N-thony this looks to be fixed.  So I am approving.  I hope that @ChrisMarsh82 can check and merge.
I did notice that the dialogue ignores a filter, if one is applied.  I checked back and this is also the case in Version 0.7.4, so perhaps to be examined later.

VimbayiChimonyo:
@N-thony forgive my ignorance - so do I run the script in R-INSTAT or do I download it again.

VimbayiChimonyo:
@N-thony I tested and it is now working. Thank you

ChrisMarsh82:
@shadrackkibet could you review this? Its R code which would be better for you to approve

instat/static/InstatObject/R/Backend_Components/calculations.R
```diff
@@ -476,38 +476,42 @@ DataBook$set("public", "apply_instat_calculation", function(calc, curr_data_list
...
-    } else if (any(stringr::str_detect("Date", col_data_type))){
-      # put in here the ones that DO NOT work for date
-      if (any(grepl("summary_sum", formula_fn_exp))){
-        curr_data_list[[c_data_label]] <- curr_data_list[[c_data_label]] %>%
-          dplyr::summarise_(.dots = setNames(list(NA), calc$result_name))
+    if (exists(deparse(substitute(col_name)), parent.frame())){
+      # get the data type of the column
+      col_data_type <- self$get_variables_metadata(data_name = calc_from_data_name, column = col_name, property = "class")
+      # if it is a ordered factor...
+      if (any(stringr::str_detect("ordered", col_data_type))){
```
shadrackkibet, line 483: There is a problem with these variable type checks. This is relying on obtaining the string from the metadata. I find this unsafe and might be difficult to maintain in future. For example, if we change how we display these variable types in the metadata then these checks will all be invalid.

A "correct" and safe way is to check the class of the column itself using `is.numeric(), is.factor()` etc

lilyclements: @shadrackkibet I do agree, but the issue with that is that this line is called in the `summary_tables` function. In that function, everything is converted to a character to create the summary tables. This is to allow for the calculations to take place without losing classes or gettings NA's (for example, for `date` variables). I'm happy to go through this in more detail if you'd like!

shadrackkibet: Great and thanks for giving more context. Please check my analysis below; 

`col_name` is always a single character therefore `col_data_type` will always be a single character which then means `any()` is redundant? but perhaps just for safety, it is better to have. I also find it unsafe to use `str_detect` here because this detects a pattern which means "partial" matching is then allowed for example `stringr::str_detect("ordered", "order")` will return `TRUE`. 

I suggest we use a strict check like `identical("ordered",col_data_type)` or just `"ordered"==col_data_type`? 
What do you think?

shadrackkibet: @lilyclements that's correct. But it is still beneficial to change to `identical()` or `%in%` and collapse with `any` instead of `str_detect`?

lilyclements: I'm getting that `identical` gives a `FALSE`:

```
col_data_type <- c("ordered", "factor")

any(identical("ordered", col_data_type))
any(str_detect("ordered", col_data_type))
```

But yes, happy to change it to `%in%`

`any("ordered" %in% col_data_type)`

shadrackkibet: What about `any(c("ordered", "character") %in% col_data_type)`?

lilyclements: That would cause problems if we had a variable that was a factor, but not ordered

```
col_data_type <- c("factor")
any(c("ordered", "factor") %in% col_data_type)
```

shadrackkibet (approved):
@rdstern if you test and re-approve then we can merge.

rdstern (approved):
That's a big relief.  I have been using it on data from Zambia and it seems to work fine now.

PR #7504, opened 2022-05-26
Added Temperature and General sub_menus to PICSA Menu

### LINKED ISSUE #7499: Implement the PICSA Temperature dialogue
@N-thony I suggest this can start initially in a very simple way.  So there is a chance it can be implemented in Version 0.7.6?  Later in the implementation it will need some skillful R, and I suggest @shadrackkibet may help with the R aspects.  I am not sure who is best placed to do this.
a) It can start as a copy of the PICSA rainfall dialogue, though the sub-dialogue will be slightly different. 
b) @dannyparsons may be able to advise, because the only real differences will be in the fitted lines.  
c) The current rainfall graphs work fine for the temperatures, except that we need to fit a line as a default, rather than having horizontal lines  as an option.
d) We also need to be able to use monthly data, and the current rainfall graph already works fine for that.
e) So the main addition initially is simply to add the controls from the main Line plot.  So Add Fitted Line (rather than line of best fit) ticked by default, and Add se, unticked by default.  
f) Perhaps instead of Lines in the PICSA options we have Labels.  There we have various options for labelling the fitted lines.  I like slope per 100 years, perhaps with a significance level as an option.
g) When facets is not None, then there needs to be an option to Fit parallel Lines.  We now have geom_parallel, but should also do the analysis to provide the significance levels.

Vitalis95:
Fixes (partially) #7499 
@rdstern @N-thony , this PR adds Temperature  and General sub-menus to PICSA Menu.
Should we rename the title of the dialogue to be more general my be `PICSA Graphs` ?
Once this PR is merged I will then continue with the other part of the issue.


lloyddewit (commented):
@Vitalis95 thanks for the changes.
I understand that:
- 3 of the climatic menu options will all open the same dialog
- this is the intention as described in issue #7499
- in future, for each menu option, the dialog will open in a different mode

Therefore we will need to change `frmMain` again later to specify the mode before `ShowDialog()` is called.

lloyddewit (approved):

lloyddewit:
@Vitalis95 @HawardKetoyoMsatsi This PR changes `frmMain`. We should merge quickly to reduce the risk of merge conflicts. So I took the liberty of testing/merging myself.
I hope that's OK.

PR #7506, opened 2022-05-27
Added missing Geoms to the plot options in the Climatic>Mapping>Map sub dialog

### LINKED ISSUE #7494: Improve the mapping dialogue in R-Instat
The Climatic > Mapping > Map dialogue was very popular in the recent climatic workshops.  There is currently a small improvement to the Plot Options button waiting to be merged (#7481).  This issue is to investigate a further related improvement in that dialogue.
The dialogue has the option to add points (usually station positions) and they can be labelled.  So, that adds geom_point and geom_label to the plot.  
But, when I use the plot options button I do not find those geoms in the list.  It just has the geom_sf from the left hand panel.  It would be good to "register" all the geoms, so they can be edited.  One obvious edit is to decrease the size of the labels, so more are visible.  Another could be to change geom_label into geom_text?
Once they are properly detected, then these geoms should also be added to the new drop-down that was implemented in #7481.
I am mentioning @N-thony here, though he may wish to allocate this.  I am sure he will need to be involved though.

EstherNjeriLiberatta:
Fixes #7494 

### DIFF
--- a/instat/dlgClimaticStationMaps.vb
+++ b/instat/dlgClimaticStationMaps.vb
@@ -45,7 +45,6 @@ Public Class dlgClimaticStationMaps
     Private clsAnnotateFunction As New RFunction
     Private clsGetDataFrame As New RFunction
     Private clsRemoveFunc As New RFunction
-    Private clsParamOperator As New ROperator
 
     Private Sub dlgClimaticMaps_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
@@ -329,24 +328,17 @@ Public Class dlgClimaticStationMaps
     Private Sub AddExtraGeoms()
         clsGGplotOperator.RemoveParameterByName("geom_label")
         clsGGplotOperator.RemoveParameterByName("geom_point")
-        clsGGplotOperator.RemoveParameterByName("geom_label_repel")
         clsGGplotOperator.RemoveParameterByName("facets")
         clsGGplotOperator.RemoveParameterByName("scale_shape_manual")
         If ucrChkAddPoints.Checked Then
             If Not ucrReceiverStation.IsEmpty Then
                 clsGGplotOperator.AddParameter("geom_label", clsRFunctionParameter:=clsLabelRepelFunction, iPosition:=2)
-            Else
-                clsGGplotOperator.RemoveParameterByName("geom_label")
             End If
             If Not ucrReceiverFacet.IsEmpty Then
                 clsGGplotOperator.AddParameter("facets", clsRFunctionParameter:=clsRFacetFunction, bIncludeArgumentName:=False, iPosition:=2)
-            Else
-                clsGGplotOperator.RemoveParameterByName("facets")
             End If
             If Not ucrReceiverShape.IsEmpty Then
                 clsGGplotOperator.AddParameter("scale_shape_manual", clsRFunctionParameter:=clsScaleShapeFunction, bIncludeArgumentName:=False, iPosition:=2)
-            Else
-                clsGGplotOperator.RemoveParameterByName("scale_shape_manual")
             End If
             clsGGplotOperator.AddParameter("geom_point", clsRFunctionParameter:=clsGeomPointFunction, bIncludeArgumentName:=False, iPosition:=2)
         End If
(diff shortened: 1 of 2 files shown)


rdstern:
@EstherNjeriLiberatta very well done, that looks great.
I was surprised, though, in the order of the layers.  It had the labels first, then the points and third was the sf.  I expected them to be the other way round.
Then I tried changing the label size to 0.3 and that gave me the following error, see below.  I wonder if that is because of the ordering of the layers?
image

N-thony:
> @EstherNjeriLiberatta very well done, that looks great. I was surprised, though, in the order of the layers. It had the labels first, then the points and third was the sf. I expected them to be the other way round. Then I tried changing the label size to 0.3 and that gave me the following error, see below. I wonder if that is because of the ordering of the layers? image

@EstherNjeriLiberatta thanks for this, is it ready for review?

EstherNjeriLiberatta:
@rdstern The error was coming about when you edit into the sub dialogue which was causing the data to be lost in the geom_point or geom_label_repel code. It is now fixed by adding the dataset to the ggplot() function. The order of the geoms is also fixed. This is ready for review.
@N-thony I have fixed the issue, it's now okay. Thanks

rdstern (changes requested):
@EstherNjeriLiberatta that's good and seems to work fine.
Ideally, the Point Options and Label Repel options would also be included in the drop-down.  
If that is tricky, then we could merge this and consider the extra geoms as an additional issue?

EstherNjeriLiberatta:
@rdstern I suggest we open another issue on this and merge this one. Thank you

rdstern (approved):
@EstherNjeriLiberatta that's good and seems to work fine.
Ideally, the Point Options and Label Repel options would also be included in the drop-down.  
If that is tricky, then we could merge this and consider the extra geoms as an additional issue?

After your comment I am approving and perhaps you could make a new issue of the point about adding the remaining geoms into the pull-down.

instat/dlgClimaticStationMaps.vb
```diff
@@ -124,14 +124,14 @@ Public Class dlgClimaticStationMaps
...
-        ucrChkAddPoints.SetParameter(New RParameter("data layers", strParamValue:=clsParamOperator), bNewChangeParameterValue:=False, bNewAddRemoveParameter:=True)
         ucrChkAddPoints.SetText("Add Points")
+        ucrChkAddPoints.AddParameterPresentCondition(True, "geom_point")
+        ucrChkAddPoints.AddParameterPresentCondition(False, "geom_point", False)
         ChangeSize()
 
     End Sub
 
     Private Sub SetDefaults()
-        clsParamOperator.Clear()
```
lloyddewit, line 134: Also delete declaration on line 48?

instat/dlgClimaticStationMaps.vb
```diff
@@ -346,17 +326,44 @@ Public Class dlgClimaticStationMaps
...
+            End If
+            If Not ucrReceiverFacet.IsEmpty Then
+                clsGGplotOperator.AddParameter("facets", clsRFunctionParameter:=clsRFacetFunction, bIncludeArgumentName:=False, iPosition:=2)
+            Else
+                clsGGplotOperator.RemoveParameterByName("facets")
+            End If
+            If Not ucrReceiverShape.IsEmpty Then
+                clsGGplotOperator.AddParameter("scale_shape_manual", clsRFunctionParameter:=clsScaleShapeFunction, bIncludeArgumentName:=False, iPosition:=2)
+            Else
+                clsGGplotOperator.RemoveParameterByName("scale_shape_manual")
```
lloyddewit, line 349: parameter already removed above. Do we need these lines?

lloyddewit (changes requested):

EstherNjeriLiberatta:
@lloyddewit I have resolved your comments above. Thank you.

lloyddewit (approved):
@EstherNjeriLiberatta Thank you, looks good

lloyddewit:
@rdstern Some code changes since your last approval. In theory, the functionality should not be affected. If you can test/reapprove, then we can merge. Thanks

rdstern (approved):

Issue #7477, opened 2022-05-16
Column Metadata doesn't open with the default selected dataset in the Data View


N-thony:
I have loaded the survey for `rwanda.RDS` file which comes with two datasets `rwanda_daily` and `rwanda_stations`, the default selected data in the `Data View window` is `rwanda_daily` but I select `rwanda_stations` then when I open `Column Metadata` it still open `rwanda_daily` by default rather than `rwanda_stations`. This was abit confusing for user during the workshop and always we were supposed to remind them to change the `dataset`.
![image](https://user-images.githubusercontent.com/68591383/168562225-24354577-3da8-49a9-bc3f-81ba778167d0.png)

rdstern:
I agree, but it is only (I assume) the first time the meta-data are loaded?  Once they have been loaded on one occasion I assume they will remember their position - that's what we do with dialogues.  So, I see this failure the first time as a bit of an inconvenience and would be happy for it to be fixed, but only if it is really quick and easy.

Patowhiz:
@rdstern from your above [comment](https://github.com/africanmathsinitiative/R-Instat/issues/7477#issuecomment-1127499183), did you mean that you only want this functionality when the user views the column metadata viewer the first time? If so, then please note that PR #7526 implementation seems to have added code that will always change the selected column metadata in the viewer, that is, "it won't remember it's position." @N-thony I'm I correct? And if so then @rdstern are you happy with that? 

Note also for the dialogs we simply decided to change the color of the selected data frame instead of selecting on behalf of the user.

PR #7508, opened 2022-05-30
Improved reordering of data frame tabs in Data View

N-thony:
Fixes (partially) #4557 
@shadrackkibet @lilyclements I checked the R code, then I found that on line 128, we were trying to check the obj class type as `data_object` i.e ` if(!is.list(new_data_objects) || (length(new_data_objects) > 0 && !all("data_object" %in% sapply(new_data_objects, class))))` to `if(!is.list(new_data_objects) || (length(new_data_objects) > 0 && !all("DataSheet" %in% sapply(new_data_objects, class))))`, please have a look, if everything is okay, I will then check the reodering in the front end.

### DIFF
--- a/instat/Model/DataFrame/clsDataBook.vb
+++ b/instat/Model/DataFrame/clsDataBook.vb
@@ -152,6 +152,9 @@ Public Class clsDataBook
                 _lstDataFrames.Remove(dataFrame)
             End If
         Next
+        If lstOfCurrentRDataFrameNames.Count = _lstDataFrames.Count Then
+            _lstDataFrames.OrderBy(Function(x) lstOfCurrentRDataFrameNames.IndexOf(x.strName))
+        End If
     End Sub
 
 
--- a/instat/Model/DataFrame/clsDataBook.vb
+++ b/instat/Model/DataFrame/clsDataBook.vb
@@ -153,7 +153,7 @@ Public Class clsDataBook
             End If
         Next
         If lstOfCurrentRDataFrameNames.Count = _lstDataFrames.Count Then
-            _lstDataFrames.OrderBy(Function(x) lstOfCurrentRDataFrameNames.IndexOf(x.strName))
+            _lstDataFrames = _lstDataFrames.OrderBy(Function(x) lstOfCurrentRDataFrameNames.IndexOf(x.strName)).ToList()
         End If
     End Sub
 
--- a/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
@@ -21,6 +21,7 @@ Public MustInherit Class ucrReoGrid
     Implements IGrid
 
     Protected _clsDataBook As clsDataBook
+    Private fillWorkSheet As Worksheet
 
     ''' <summary>
     ''' Gets current worksheet adapter
@@ -61,7 +62,7 @@ Public MustInherit Class ucrReoGrid
     End Property
 
     Public Function AddNewWorksheet(name As String) As clsWorksheetAdapter Implements IGrid.AddNewWorksheet
-        Dim fillWorkSheet = grdData.CreateWorksheet(name)
+        fillWorkSheet = grdData.CreateWorksheet(name)
         grdData.AddWorksheet(fillWorkSheet)
         fillWorkSheet.SelectionForwardDirection = unvell.ReoGrid.SelectionForwardDirection.Down
         fillWorkSheet.SetSettings(unvell.ReoGrid.WorksheetSettings.Edit_DragSelectionToMoveCells, False)
@@ -70,6 +71,20 @@ Public MustInherit Class ucrReoGrid
         AttachEventsToWorksheet(fillWorkSheet)
         Return New clsWorksheetAdapter(fillWorkSheet)
     End Function
+
+    Private Sub ReOrderWorksheets(strDataframe As String, iCount As Integer) Implements IGrid.ReOrderWorksheets
+        Dim iNewPosition As Integer
+        For Each tempWorkSheet In grdData.Worksheets
+            If tempWorkSheet.Name = strDataframe Then
+                iNewPosition = iCount
+                Exit For
+            End If
+        Next
+        If fillWorkSheet IsNot Nothing AndAlso iNewPosition < grdData.Worksheets.Count Then
+            grdData.MoveWorksheet(fillWorkSheet, iNewPosition)
+        End If
+        'grdData.Worksheets.OrderBy(Function(x) _clsDataBook.DataFrames.IndexOf())
+    End Sub
 
     Public Sub CopyRange() Implements IGrid.CopyRange
         grdData.CurrentWorksheet.Copy()
--- a/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
@@ -171,6 +171,7 @@ Public MustInherit Class ucrReoGrid
         End If
         If grdData.Worksheets.Count > 0 Then
             grdData.MoveWorksheet(fillWorkSheet, iNewPosition)
+            grdData.CurrentWorksheet = fillWorkSheet
         End If
     End Sub
 
--- a/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
@@ -157,25 +157,6 @@ Public MustInherit Class ucrReoGrid
         Next
     End Sub
 
-    Private Sub ReorderSheets(strSheet As String, iNewPosition As Integer) Implements IGrid.ReorderSheets
-        Dim bFoundWorksheet As Boolean = False
-        Dim grdWorksheetToMove As Worksheet = Nothing
-        If strSheet <> "" Then
-            For Each tempWorkSheet In grdData.Worksheets
-                If tempWorkSheet.Name = strSheet Then
-                    bFoundWorksheet = True
-                    Exit For
-                End If
-            Next
-
-            If bFoundWorksheet AndAlso grdData.Worksheets.Count > 0 Then
-                grdWorksheetToMove = grdData.GetWorksheetByName(strSheet)
-                grdData.MoveWorksheet(grdWorksheetToMove, iNewPosition)
-                grdData.CurrentWorksheet = grdWorksheetToMove
-            End If
-        End If
-    End Sub
-
     Public Sub UpdateWorksheetStyle(worksheet As clsWorksheetAdapter) Implements IGrid.UpdateWorksheetStyle
         UpdateWorksheetStyle(grdData.Worksheets.Where(Function(x) x.Name = worksheet.Name).FirstOrDefault)
     End Sub
--- a/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
@@ -157,6 +157,23 @@ Public MustInherit Class ucrReoGrid
         Next
     End Sub
 
+    Private Sub ReorderSheets(strSheet As String, iNewPosition As Integer) Implements IGrid.ReorderSheets
+        Dim bFoundWorksheet As Boolean = False
+        Dim fillWorkSheet As Worksheet = Nothing
+        For Each tempWorkSheet In grdData.Worksheets
+            If tempWorkSheet.Name = strSheet Then
+                bFoundWorksheet = True
+                Exit For
+            End If
+        Next
+        If bFoundWorksheet Then
+            fillWorkSheet = grdData.GetWorksheetByName(strSheet)
+        End If
+        If grdData.Worksheets.Count > 0 Then
+            grdData.MoveWorksheet(fillWorkSheet, iNewPosition)
+        End If
+    End Sub
+
     Public Sub UpdateWorksheetStyle(worksheet As clsWorksheetAdapter) Implements IGrid.UpdateWorksheetStyle
         UpdateWorksheetStyle(grdData.Worksheets.Where(Function(x) x.Name = worksheet.Name).FirstOrDefault)
     End Sub
--- a/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
@@ -92,6 +92,7 @@ Public MustInherit Class ucrReoGrid
                     grdData.MoveWorksheet(fillWorkSheet, iNewPosition)
                     grdData.CurrentWorksheet = fillWorkSheet
                     iCount += 1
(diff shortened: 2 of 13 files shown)


shadrackkibet:
@N-thony  we still need to check for  `data_object` for backward compatibility purposes. Therefore this check should change to something like `!any(c("DataSheet","data_object") %in% sapply(new_data_objects, class))`

N-thony:
@shadrackkibet @Patowhiz I have managed to update the reordering of the data frame from the front end as it is done in the back-end with the Data Frame MetaData, and fixed some bugs, please take a look.
@rdstern could you test?

N-thony:
> 

@Patowhiz thanks, I will discuss what I have done here with @ChrisMarsh82 and also about his approach.

rdstern (changes requested):
@N-thony this looks great and all seems to work.  So the points I make are all cosmetic.
a) The right-click at the bottom of the data frame remains disabled.  It is time now to enable it.  The dialogue works fine and is enabled.
b) When I go in to the dialogue for the first time, it is empty.  No data frames.  Reset doesn't help.  But if I go out, and then back in again, it is fine, and then ok from then onwards.
c) The control is on the left of the dialogue.  The same control for reordering columns is more on the right, because there is another control on the left.  I wonder if the control should be in the same position on the two dialogues.  (But you are better than me on design issues, so you decide on this.)
d) Then the important things seem to work fine - they reorder with no problem, and are correctly changed on the sheet metadata.
e) When finished it always moves to the last dataframe.  That seems to be independent of the last one that was used, or where you started, with this dialogue.  I don't have much of a problem with this, but wonder if it could return to the dat frame where I started?  That will often be the right-click data frame I guess, once that option works.
f) Finally I notice the "obvious" way - used in Excel, still works.  So I can click on a dataframe in reogrid and it changes position there.  But, of course, the metadata hasn't changed.  I am assuming this feature can't be trapped and made to work, and so wonder if it can be disabled?  Maybe that is a @Patowhiz question?

N-thony:
f) Finally I notice the "obvious" way - used in Excel, still works. So I can click on a dataframe in reogrid and it changes position there. But, of course, the metadata hasn't changed. I am assuming this feature can't be trapped and made to work, and so wonder if it can be disabled? Maybe that is a @Patowhiz question?
@rdstern I think there is an issue #7477 on this, it was reported during the KMD workshop.

Patowhiz:
@rdstern @N-thony  Like I said before, I think the approach done here needs to be consistent with how data frame metadata changes has been implemented.

Currently, we don't have the mechanism of tracking data book metadata changes at the .Net level. That's where I'd suggest we start with. It's very likely that in future we may want to track other data book metadata changes too. So I'd recommend adding that feature now.

If that's implemented, most of the code changes here may no longer be needed

Thanks.

N-thony:
> @rdstern @N-thony Like I said before, I think the approach done here needs to be consistent with how data frame metadata changes has been implemented.
> 
> Currently, we don't have the mechanism of tracking data book metadata changes at the .Net level. That's where I'd suggest we start with. It's very likely that in future we may want to track other data book metadata changes too. So I'd recommend adding that feature now.
> 
> If that's implemented, most of the code changes here may no longer be needed
> 
> Thanks.

@Patowhiz this is just what was before but now it is done with Chris structure, we still need to discuss approaches, as you said above, I will discuss that with @ChrisMarsh82 when he will be back.

N-thony:
> 

I have checked the code, and this is already happening but need some check with @ChrisMarsh82 in `RefreshDataFrames sub` in the `clsDataBook`.

N-thony:
@ChrisMarsh82 please have a look, thanks.
@rdstern could you re-test?

ChrisMarsh82:
@N-thony This looks good from the Reo grid side. Will need to be implemented for Linux grid. It is just tab pages used in the linux grid so you should be able to use something like this https://stackoverflow.com/questions/12001502/how-to-easily-reorder-tabcontrol

instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
```diff
@@ -71,6 +71,32 @@ Public MustInherit Class ucrReoGrid
...
+            If bFound Then
+                Dim fillWorkSheet = grdData.GetWorksheetByName(strName)
+                If fillWorkSheet IsNot Nothing AndAlso iNewPosition < grdData.Worksheets.Count Then
+                    grdData.MoveWorksheet(fillWorkSheet, iNewPosition)
+                    grdData.CurrentWorksheet = fillWorkSheet
+                    iCount += 1
+                End If
+            End If
+        Next
+    End Sub
```
Patowhiz, line 98: @N-thony I'm probably missing something, is it true that `bFound` will always be true when after it's set to true once?
Looking at this code I didn't see where it's set back to false.

Patowhiz: Or may be it doesn't really matter?

Patowhiz: Then also looking at `ReOrderWorksheets()` subroutine. Does it generally assume that all worksheets are always not in their previous position(order) and therefore attempts to always change all their indexes(positions)? That is,
`grdData.MoveWorksheet(fillWorkSheet, iNewPosition)`
will always be executed even for worksheets that retained their position?

N-thony: > Then also looking at `ReOrderWorksheets()` subroutine. Does it generally assume that all worksheets are always not in their previous position(order) and therefore attempts to always change all their indexes(positions)? That is, `grdData.MoveWorksheet(fillWorkSheet, iNewPosition)` will always be executed even for worksheets that retained their position?

If the position is still the same then there is no problem, it will retain the same position otherwise it will move it to the specific new position.

N-thony:
> @N-thony This looks good from the Reo grid side. Will need to be implemented for Linux grid. It is just tab pages used in the linux grid so you should be able to use something like this https://stackoverflow.com/questions/12001502/how-to-easily-reorder-tabcontrol

Thanks for the link, let me find out how to implement it in VB.

instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
Patowhiz: @N-thony understood.
I'm I correct to state that the ordering implementation here assumes the following;
1. The data book will always have all the data frames even the ones hidden from the user.
2. The grid may not have all the data frame worksheets equivalent, for instance hidden data frames may not be in the in the grid as work sheets.
3. All data frames in the data book are assumed to have changed their order positions

N-thony: Yes, you are correct.

ChrisMarsh82 (approved):

instat/Model/DataFrame/clsDataBook.vb
```diff
@@ -152,6 +152,9 @@ Public Class clsDataBook
                 _lstDataFrames.Remove(dataFrame)
             End If
         Next
+        If lstOfCurrentRDataFrameNames.Count = _lstDataFrames.Count Then
+            _lstDataFrames = _lstDataFrames.OrderBy(Function(x) lstOfCurrentRDataFrameNames.IndexOf(x.strName)).ToList()
+        End If
```
Patowhiz, line 157: @N-thony according to the way `RefreshDataFrames()` is implemented, at line 155 during runtime, the count of `lstOfCurrentRDataFrameNames` will always be the same as `_lstDataFrames`. I don't think we really need that check.

rdstern (approved):
Looks fine to me.

instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
Patowhiz: @N-thony okay.
Below is an alternative implementation that can be adapted to achieve the same. You may choose to ignore it, I have no strong objection.

```
Private Sub ReOrderWorksheets() Implements IGrid.ReOrderWorksheets
        'assuming the databook will always have all the data frames 
        'and the grid may not have all the data frame worksheets equivalent
        'and all data frames in the data book have changed their order positions 
        'get data frames sheets in the grid based on the databook data frames position order
        'and add it to the list.
        Dim lstWorkSheetsFound As New List(Of Worksheet)
        For Each clsDataframe In _clsDataBook.DataFrames
            Dim fillWorkSheet As Worksheet = grdData.GetWorksheetByName(clsDataframe.strName)
            If fillWorkSheet IsNot Nothing Then
                lstWorkSheetsFound.Add(fillWorkSheet)
            End If
        Next
        'then reorder the worksheets based on the filled list
        For i As Integer = 0 To lstWorkSheetsFound.Count - 1
            grdData.MoveWorksheet(lstWorkSheetsFound(i), i)
        Next
    End Sub
```

lloyddewit:
@N-thony Thanks for this. There are 2 approvals so in theory, we can merge.
However, there is still an open comment above from @Patowhiz related to `ReOrderWorksheets()`. The code he proposes looks more readable to me than the current implementation. If the code works, then would you consider pasting it in?

N-thony:
> @N-thony Thanks for this. There are 2 approvals so in theory, we can merge. However, there is still an open comment above from @Patowhiz related to `ReOrderWorksheets()`. The code he proposes looks more readable to me than the current implementation. If the code works, then would you consider pasting it in?

@lloyddewit okay, I made the change and fixed the small bug coming from the proposed code, have a look. Thanks.

instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
Patowhiz: @N-thony just 3 more questions before this is merged, 

1. Why do we need line 90,  `grdData.CurrentWorksheet = lstWorkSheetsFound(i)`? Does moving of worksheets change `grdData.CurrentWorksheet`  property? 
2. When I reorder the worksheet, should the active worksheet, that is,  `grdData.CurrentWorksheet` be changed to the last one?
3. Now that you decided to adapt my code suggestion, do you think it's a good idea to do the same for the data grid view?

Thanks.

N-thony:
> > @N-thony Thanks for this. There are 2 approvals so in theory, we can merge. However, there is still an open comment above from @Patowhiz related to `ReOrderWorksheets()`. The code he proposes looks more readable to me than the current implementation. If the code works, then would you consider pasting it in?
> 
> @lloyddewit okay, I made the change and fixed the small bug coming from the proposed code, have a look. Thanks.

It looks good the code but this is what you get after running, that why we need `grdData.CurrentWorksheet`
before reordering
image
after reordering
image

rdstern (dismissed):
I remain happy with the implementation of this feature and it would be good if we can include it in the forthcoming version.
It seems to work fine for me.  I did notice that it leaves you at the last data frame, rather than the one where you started the command.  I am sort of ok with that.

lloyddewit:
> I remain happy with the implementation of this feature and it would be good if we can include it in the forthcoming version. It seems to work fine for me. I did notice that it leaves you at the last data frame, rather than the one where you started the command. I am sort of ok with that.

@Patowhiz @N-thony It would also be good if @Patowhiz 's remaining comments above could be resolved. Would this be possible by end today (or latest tomorrow)? Or should we go ahead and merge and make a new PR for any further changes?

N-thony:
> > I remain happy with the implementation of this feature and it would be good if we can include it in the forthcoming version. It seems to work fine for me. I did notice that it leaves you at the last data frame, rather than the one where you started the command. I am sort of ok with that.
> 
> @Patowhiz @N-thony It would also be good if @Patowhiz 's remaining comments above could be resolved. Would this be possible by end today (or latest tomorrow)? Or should we go ahead and merge and make a new PR for any further changes?

@lloyddewit I will make the change if necessary in another PR for point 3), I don't see any problem with 1) and  2)

Patowhiz:
@rdstern what you noticed and sort of okay with, is what I'm asking @N-thony about.

If I have 5 worksheets and the 2nd is the one active, if I reorder the worksheets. 

1. Which worksheet should be the active one? 
2. Should it be the original one or the last one in the new order?

That one line of code is meant to always have the last worksheet in the new order be active one, that's what you have noticed.
Removing that one line of code retains the original active worksheet. 

I would suggest to retain the original worksheet as the active one. Reason being the user is just reordering data frames not change the active data frame.
@rdstern @N-thony  I'm happy to be convinced otherwise.

N-thony:
> > @rdstern what you noticed and sort of okay with, is what I'm asking @N-thony about.
> > If I have 5 worksheets and the 2nd is the one active, if I reorder the worksheets.
> > 
> > 1. Which worksheet should be the active one?
> > 2. Should it be the original one or the last one in the new order?
> > 
> > That one line of code is meant to always have the last worksheet in the new order be active one, that's what you have noticed. Removing that one line of code retains the original active worksheet.
> > I would suggest to retain the original worksheet as the active one. Reason being the user is just reordering data frames not change the active data frame. @rdstern @N-thony I'm happy to be convinced otherwise.
> 
> @Patowhiz it is fine, as it is okay at this point with Roger, I'm opening another issue where I will fix that.

@lloyddewit if you are happy with my suggestion, once this is merged, I will open another issue where I will work on that aspect of remaining to current worksheet.

rdstern (approved):
Agreed with a new issue of keeping the original data frame.

Patowhiz:
@rdstern @N-thony another test case scenario for you.

Load the climatic_guide_datasets.xlsx from the library the import all the sheets from the excel file.
Select the 3rd work sheet in the data viewer, then go to the calculator dialog, enter "surface", click Ok. 

The last worksheet is the one that becomes active. I think that's not what a user may expect.





 in addition try using 
I have 2 extra objections

N-thony:
> @rdstern @N-thony another test case scenario for you.
> 
> Load the climatic_guide_datasets.xlsx from the library the import all the sheets from the excel file. Select the 3rd work sheet in the data viewer, then go to the calculator dialog, enter "surface", click Ok.
> 
> The last worksheet is the one that becomes active. I think that's not what a user may expect.

See my comment above,

Patowhiz:
@rdstern @N-thony then also using the same file with the same number of worksheets(31).

Do the same calculator operation, notice how the grid blinks several times and it's just 31 sheets. This is similar to what we had before with the dialogs. This problem also compounds for very sheets added. Try using 62 sheets.

This is because this function now gets called every single time we have an output from an operation, and it's not just once. So we are always reordering data frames, which I'm objecting because of performance. 

@lloyddewit @N-thony like I said before, I understand getting this right requires several other aspects to be built in, and that may require time.

We can merge this if it's urgently needed by @rdstern for the next release but lets be aware of the problems introduced. If not urgent then I strongly suggest we pause and get it right first.

Patowhiz:
The other point that I can add. In this implementation, we are also always reordering data frames every time we create a new data frame. After now testing this, I'm convinced that's a future potential performance issue that will force us to refactor this feature.

lloyddewit (approved):
@N-thony @Patowhiz thanks for both the implementation and review comments.
As discussed in the comments above, I'll merge this PR and @N-thony can adress the potential performance and other concerns in a new issue.

PR #7516, opened 2022-05-31
Fixed grid row sizes for shape file data frames

### LINKED ISSUE #7165: New grid has row size problem for some datasets
The example I am using is a single row - but many other examples are possible.  The data are from [gadm](https://gadm.org/download_country.html).
This figure uses Version 0.7.2 so with the old coding for the grid.
image

Notice the metadata is a bit messy.  But the data grid is fine - well at least the rows are ordinary size.

Here is the same data in 7.3.1.  The multipolygon has become a list of lists.  The metadata is as before.

:
image
This cell in the grid now goes on forever - well almost!
I really like it that reogrid sometimes stores lists in a cell, and have other examples.  But can we either avoid wordwrap, or have an option to turn it on and off?

I assume this is initially a question to @ChrisMarsh82 or @dannyparsons and might then be pretty easy for someone to fix?

Patowhiz:
Fixes #7165 


lloyddewit (approved):

rdstern (approved):
Looks fine to me.  And helps our simple mapping that is in our climatic courses regularly now.

PR #7517, opened 2022-05-31
Adds a data book null check, in the data frame selector, to prevent design view error

### LINKED ISSUE #7489: VB.Net Designer Error
I have been getting this error whenever I try to open dlgCalculator designer from VB.Net solution Explorer.
How do I fix it?

image

Patowhiz:
Fixes #7489.
@africanmathsinitiative/developers @Wycklife this is ready for review.

@lloyddewit,

I'm not sure if we should generally allow committing changes like the one I've made `ucrDataFrame`. I did it here for @Wycklife to test, I can remove it if need be.

I think developers should be calling the `IsInDesignMode`  to make design changes then just remove or comment it out after that. 

I also understand there is no any significant performance difference in leaving such code. So I'm happy with any suggestion.

Thanks.

### DIFF
--- a/instat/mdlCoreControl.vb
+++ b/instat/mdlCoreControl.vb
@@ -135,23 +135,4 @@ Module mdlCoreControl
             Return "c(" & String.Join(",", enumerable.ToArray()) & ")"
         End If
     End Function
-
-    'temporary code for visual studio design time
-
-    ''' <summary>
-    ''' Checks if current system process is in design mode.
-    ''' <para>To be used by developers to prevent visual studio from executing code during design mode. 
-    ''' Especially code that can be source of design errors.</para>
-    ''' </summary>
-    ''' <returns>True, if current process is in design mode. False otherwise.</returns>
-    Public Function IsInDesignMode() As Boolean
-        'temporary for developers to prevent visual studio from executing code during design mode
-        Dim process As System.Diagnostics.Process = System.Diagnostics.Process.GetCurrentProcess()
-        If process.ProcessName = "devenv" Then
-            process.Dispose()
-            Return True
-        End If
-        Return False
-    End Function
-
 End Module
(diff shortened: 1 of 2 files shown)


lloyddewit:
@Patowhiz Thanks, this is a clever approach. I agree it has some disadvantages and I'm not sure we should commit it.

I understand that if a developer opens `dlgCalculator.Designer.vb` from the Solution Explorer, then `ucrDataFrame_Load()` is executed. This triggers an exception on the highlighted line below:

image

In a sense, this is fragile code because it assumes (without checking) that the objects are not nothing. Would it  be safer to add a check? for example:

```
        ElseIf Not IsNothing(frmMain) AndAlso Not IsNothing(frmMain.DataBook) AndAlso Not IsNothing(frmMain.DataBook.DataFrames) Then
            For Each dataFrame As clsDataFrame In frmMain.DataBook.DataFrames
                cboAvailableDataFrames.Items.Add(dataFrame.strName)
            Next
```
We could safely leave this check in the production code. 
Would this work? What do you think?

lloyddewit:
I just tested but could not recreate the error. Please see my comment in issue #7489 .

Patowhiz:
@lloyddewit I've added a data book check as suggested. I didn't see the need for `frmMain` and `DataFrames` null check because I can't see any scenario that they will ever be null.
Thanks.

lloyddewit (approved):

lloyddewit:
@Wycklife Please could you confirm that this solves your issue?
Thanks

Wycklife:
@Patowhiz @lloyddewit  I'm updating my VB.Net, the error seems to exist with the version am currently using.

Wycklife:
@lloyddewit I have tested it with the new version(16.11.15) and it looks fine after building the solution.

lloyddewit:
@Wycklife Thank you for confirming.

@rdstern @Patowhiz It's still not 100% clear to me if this fix is still needed for more recent versions of Visual Studio. Howeverer, the new check is still a valid quality improvement without side effects (as far as I can see).
So I willl merge this PR.

PR #7519, opened 2022-06-02
Added Random Split dialog

### LINKED ISSUE #7227: Add a new dialogue possibly called Random Split using rsample?

derekagorhom:
Fixes #7227
This PR replaces PR #7272 
This dialogue can be found in Prepare>Data Reshape > Random Split dialogue.
This is ready for review.
I have made the recommended changes requested by @lilyclements except `no.4`, since that needs @rdstern input/approval.


lloyddewit:
@N-thony - Please could you peer review?
Thanks

instat/dlgRandomSplit.vb
```diff
@@ -1,5 +1,219 @@
...
+                ucrBase.clsRsyntax.iCallType = 2
+            End If
+        End If
+    End Sub
+
+    Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
+        SetDefaults()
+        SetRCodeForControls(True)
+        TestOkEnabled()
+    End Sub
```
N-thony, line 197: One line space

N-thony (changes requested):

rdstern:
@derekagorhom please wait for @lilyclements confirmation before continuing with this pull request.  And make sure you fully understand the reasons for the suggested changes, before implementing them. 
a) The R code needs to change.  The rsample documentation shows why.  You need to generate the sample into x first and then produce the training and testing sets.  (You must ensure they complement each other, i.e. that the testing subset is what is left over, from the training set.  If you generate them separately - as you are doing now, then some observations may be in both.)
b) Produce and same both of them each time.
c) Don't include the arguments in the code that are not needed.  Currently you have breaks and pool always.  You rarely need them.)
So, the dialogue will change quite a lot.  
d) You don't need the data selector to be visible unless you tick the stratifying factor.  
e) Delete the label "Variables" and move the Stratifying factor checkbox to just above the control.  (It can act as the label as well.)
f) Delete the groups control.  (That's for stratifying by a numeric variable, which we are not allowing here.)
g) Omit the the 2 checkboxes for saving the training and testing sets - let's always save both of them.
h) Here is a more speculative query for @lilyclements .  Should we add a simple group box called Naming:  Inside are two radio buttons.  The first (default) is analysis/assessment and the second is training/testing.  (This issue is discussed in the rsample guide as a section in rsample called terminology.)  If default, then the resulting data frames are called analysis and assessment (and then analysis1, assessment1, etc.  Otherwise they are called training and testing, etc.
i) Should we also save the rsample object?  If so, possibly that should be optional, so have a checkbox, default unchecked.
j) Another @lilyclements question is that there is a rsample2caret function.  Should we allow this option.  If so, then this could be the object stored instead?

lilyclements:
@derekagorhom for (a) do you know the R code to achieve this? From #7227 

If "Sample" is selected then `initial_split` is the base function
* `data` is the data in the data frame
* `prop` is the proportion given in the fraction nud (default `0.75`)
* `strata` is default `NULL`. If ucrChkStratifyingFactor is checked, then `strata` is the variables in the receiver
* `breaks` could be included if we allow numeric variables in the strata receiver.
* `pool` is the number given in the pool nud (default `0.1`, but takes values `0`-`1`).

If "Time Series" is selected then `initial_time_split` is the base function
* `data` is the data in the data frame
* `prop` is the proportion given in the fraction nud (default `0.75`)
* `lag` is the number given in the lag nud (default `0`, takes any numerical integer).

This creates an object, which we can call `x` for now. That is, `x <- initial_split(....)` (or `initial_time_split(...)`).

For "save training data", run `training()` around the saved object (output from `initial_split`, or `initial_time_split`)
And save the output as a new dataframe.
That is, `training_data <- training(x)`

For "save testing data", run `testing()` around the saved object (output from `initial_split`, or `initial_time_split`)
And save the output as a new dataframe.
That is, `testing_data <- testing(x)`

@rdstern to:
h) Why not offer the user to name the data frame whatever they wish? With the default as the data frame name, followed by "_training" or "_testing" (e.g. "mtcars_training"). It is good to use the guide, but I'm not sure if that is overcomplicating it?
i) If we save the `rsample` object, what would we want to offer can be done with it? At the moment you can look at different objects, such as the row numbers that are used in one of the data sets. Is this something that could be viewed on the "View Objects" dialog?

j) The `rsample2caret` function does not work for `initial_split` unfortunately. It works for other functions (from what I can tell: `vfold_cv()`, `bootstraps()`, `mc_cv()`, `rolling_origin()`)

E.g. for `initial_split()` we run,

```
car_split <- initial_split(mtcars)
train_data <- training(car_split)
```

for `bootstraps()` we run:
```
car_splits <- bootstraps(mtcars, times = 2)
car_splits$splits  # gives a list of all the objects, like "car_split" object above.
car_splits$splits[[1]]  # returns just the first object, like "car_split" above
train_data_1 <- training(car_splits$splits[[1]])  # to get the training data for the first data set
rsample2caret(car_splits, data = c("analysis", "assessment"))
```

derekagorhom:
Thank you for the explanation @lilyclements, I will get to working on the Pull request.

N-thony:
> Thank you for the explanation @lilyclements, I will get to working on the Pull request.

@derekagorhom any progress?

N-thony:
> > Thank you for the explanation @lilyclements, I will get to working on the Pull request.
> 
> @derekagorhom any progress?

@derekagorhom any progress?

derekagorhom:
It's still in progress

lilyclements:
@derekagorhom how is this going? Is it waiting for particular feedback from anyone, or have you just got busy on other tasks? (both are absolutely fine! I just wanted to check in case there was anything I could do to help)

derekagorhom:
> @derekagorhom how is this going? Is it waiting for particular feedback from anyone, or have you just got busy on other tasks? (both are absolutely fine! I just wanted to check in case there was anything I could do to help)

@lilyclements please could you help me with implementing the rcode, i am having some difficultly with that

lilyclements:
> @lilyclements please could you help me with implementing the rcode, i am having some difficultly with that

@derekagorhom sure. The R-code is given above in this issue, what part of it are you having issue with? From a quick review, it looks like you have implemented the R-code quite well.

I had issues on the VB side of the code in:
* `lag` not giving the value if I selected the checkbox, but didn't change the nud.
* `pool` nud not increasing/decreasing
* `strata` running even if I unchecked the box

I think we would want two saves too - one for saving testing, one for saving training data. @rdstern can confirm this though!

rdstern:
@lilyclements and @derekagorhom I agree with the need for 2 saves.

derekagorhom:
@lilyclements can you test and review the PR, i have made the changes

derekagorhom:
@lilyclements  i did test, it does produce result but as @rdstern said there are results that appear in both. i was hoping you could help me solve it.

lilyclements:
@derekagorhom understood - so to avoid having values in both the training and testing data, you need to only run your `initial_split` (or `initial_split_time`) function once, and assign it to something.
Then you want to **call** your assigned object. 

For example,

```
rsample_data <- rsample::initial_split(data=survey, prop=0.75, strata="village", breaks=4, pool=0)

testing_data <- rsample::testing(rsample_data)
training_data <- rsample::training(rsample_data)
```

lilyclements (changes requested):
@derekagorhom this is looking great! One thing is that the "Save testing"/"Save training" checkboxes to be ucrSaves instead (and then the current ucrSave can be removed).
Coming along well though!

derekagorhom:
Hello @lilyclements and @rdstern while working on making the final changes to the dialogue, i come across this, if you check both screenshots, you will realize one value keeps appearing in both datasets (even after we made the corrections). 

At first i thought maybe it was because it had two `ucrSaves` now, but that was not the case. 
for clarity this is the script code

```
# Code generated by the dialog, Random Split
rsample <- rsample::initial_split(data=survey, prop=0.75, breaks=4, pool=0.10)
training <- rsample::training(x=rsample)
data_book$import_data(data_tables=list(training=training))


# Code generated by the dialog, Random Split
rsample <- rsample::initial_split(data=survey, prop=0.75, breaks=4, pool=0.10)
testing <- rsample::testing(x=rsample)
data_book$import_data(data_tables=list(testing=testing))
```


Screenshot (465)
Screenshot (464)

lilyclements:
@derekagorhom this is a great observation - good job for checking the output and finding this discrepancy.
The issue here is that you are running `rsample::initial_split` twice - once for the training data, and once for the testing data.
We just want to run it once. When it is run the once, it creates both the training and testing data sets, which we then call through the `training` and `testing` functions.

So, the code you want to be run in the dialog is:

```
# Code generated by the dialog, Random Split
rsample <- rsample::initial_split(data=survey, prop=0.75, breaks=4, pool=0.10)
training <- rsample::training(x=rsample)
testing <- rsample::testing(x=rsample)
data_book$import_data(data_tables=list(training=training))
data_book$import_data(data_tables=list(testing=testing))
```
`

N-thony:
> @derekagorhom this is looking great! One thing is that the "Save testing"/"Save training" checkboxes to be ucrSaves instead (and then the current ucrSave can be removed). Coming along well though!

@derekagorhom I got it now from this comment, I think we the current codes you have it is simpler to implement this.

derekagorhom:
@N-thony , @lilyclements  i have made the changes

N-thony:
@derekagorhom I have tested this, there is a small problem when you go under `Time Series` option, you can see in the script window when you check `Lag` there is no default value. 
And also reduce the space between the above controls and the save controls.
image

instat/dlgRandomSplit.vb
```diff
@@ -1,5 +1,210 @@
...
+
+    Private Sub ucrCore_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrNudFraction.ControlContentsChanged, ucrNudLag.ControlContentsChanged, ucrNudPool.ControlContentsChanged, ucrPnlRandomSplit.ControlContentsChanged, ucrChkLag.ControlContentsChanged, ucrChkStratifyingFactor.ControlContentsChanged, ucrNudBreaks.ControlContentsChanged, ucrSaveTrainingData.ControlContentsChanged, ucrSaveTestingData.ControlContentsChanged
+        TestOkEnabled()
+    End Sub
+
+    Private Sub ucrPnlRandomSplit_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlRandomSplit.ControlValueChanged '', ucrChkTestingData.ControlValueChanged, ucrChkTrainingData.ControlValueChanged
```
N-thony, line 193: you have comment some codes in this line, can you remove them if not needed?

instat/dlgRandomSplit.vb
```diff
@@ -1,5 +1,210 @@
...
+        Dim strDataframeName As String = ucrSelectorRandomSplit.ucrAvailableDataFrames.cboAvailableDataFrames.Text
+        If ucrSelectorRandomSplit.ucrAvailableDataFrames.cboAvailableDataFrames.Text <> "" AndAlso (Not ucrSaveTestingData.bUserTyped) AndAlso (Not ucrSaveTrainingData.bUserTyped) Then
+            If rdoSample.Checked Then
+                ucrSaveTestingData.SetPrefix(strDataframeName & "testing")
+                ucrSaveTrainingData.SetPrefix(strDataframeName & "training")
+            End If
+            If rdoTimeSeries.Checked Then
+                ucrSaveTestingData.SetPrefix(strDataframeName & "testing")
+                ucrSaveTrainingData.SetPrefix(strDataframeName & "training")
+            End If
```
N-thony, line 207: what about this?
            If rdoSample.Checked Then
                ucrSaveTestingData.SetPrefix(strDataframeName & "testing")
                ucrSaveTrainingData.SetPrefix(strDataframeName & "training")
            Else
                ucrSaveTestingData.SetPrefix(strDataframeName & "testing")
                ucrSaveTrainingData.SetPrefix(strDataframeName & "training")
            End If

instat/dlgRandomSplit.vb
```diff
@@ -1,5 +1,210 @@
...
+    Private Sub SetBaseFunction()
+        If rdoSample.Checked Then
+            If ucrSaveTestingData.IsComplete Then
+                clsTesting.AddParameter("x", clsRFunctionParameter:=clsInitialSplit)
+                ucrBase.clsRsyntax.SetBaseRFunction(clsTesting)
+            End If
+            If ucrSaveTrainingData.IsComplete Then
+                clsTraining.AddParameter("x", clsRFunctionParameter:=clsInitialSplit)
+                ucrBase.clsRsyntax.AddToAfterCodes(clsTraining)
+            End If
```
N-thony, line 169: What about this?
            If ucrSaveTestingData.IsComplete Then
                clsTesting.AddParameter("x", clsRFunctionParameter:=clsInitialSplit)
                ucrBase.clsRsyntax.SetBaseRFunction(clsTesting)
             ElseIf ucrSaveTrainingData.IsComplete Then
                clsTraining.AddParameter("x", clsRFunctionParameter:=clsInitialSplit)
                ucrBase.clsRsyntax.AddToAfterCodes(clsTraining)
            End If

instat/dlgRandomSplit.vb
```diff
@@ -1,5 +1,210 @@
...
+        Else
+            If ucrSaveTestingData.IsComplete Then
+                clsTesting.AddParameter("x", clsRFunctionParameter:=clsInitialTimeSplit)
+                ucrBase.clsRsyntax.SetBaseRFunction(clsTesting)
+            End If
+            If ucrSaveTrainingData.IsComplete Then
+                clsTraining.AddParameter("x", clsRFunctionParameter:=clsInitialTimeSplit)
+                ucrBase.clsRsyntax.AddToAfterCodes(clsTraining)
+
+            End If
```
N-thony, line 179: see my comment above

instat/dlgRandomSplit.vb
```diff
@@ -1,5 +1,210 @@
...
+        ucrBase.clsRsyntax.SetBaseRFunction(clsInitialSplit)
+    End Sub
+
+    Private Sub SetRCodeForControls(bReset As Boolean)
+        ucrSelectorRandomSplit.AddAdditionalCodeParameterPair(clsInitialTimeSplit, New RParameter("data", ucrSelectorRandomSplit.ucrAvailableDataFrames.clsCurrDataFrame, 0), iAdditionalPairNo:=1)
+        ucrNudFraction.AddAdditionalCodeParameterPair(clsInitialTimeSplit, New RParameter("prop", 1), iAdditionalPairNo:=1)
+        ucrNudLag.AddAdditionalCodeParameterPair(clsInitialTimeSplit, New RParameter("lag", 3), iAdditionalPairNo:=1)
+
+        ucrSelectorRandomSplit.SetRCode(clsInitialSplit, bReset)
+        ucrChkStratifyingFactor.SetRCode(clsDummyFunction, bReset)
```
N-thony, line 140: this can be done with the ucrNudFraction without using a DummyFunction

N-thony (changes requested):

N-thony:
@derekagorhom when `Satisfying Factor` is unchecked or checked I still get the same code, is this okay?
image

derekagorhom:
@N-thony in the new commits, i set up the `strata` to be linked to the `stratifying factor` so that it does not appear unless the `factor` is checked to prevent it from appearing on the script

instat/dlgRandomSplit.vb
derekagorhom: when i coded it like that i would only get one dataframe out of the two

N-thony:
@derekagorhom if you `check` or `unckeck` `stratifying factor` strata is still present in the code.
image

N-thony:
@derekagorhom could we have a quick call?

lilyclements:
> @N-thony , @lilyclements i have made the changes

@derekagorhom great! I'll review once you and @N-thony have finished his review!

instat/dlgRandomSplit.vb
```diff
@@ -1,5 +1,207 @@
...
+        If ucrSelectorRandomSplit.ucrAvailableDataFrames.cboAvailableDataFrames.Text <> "" AndAlso (Not ucrSaveTestingData.bUserTyped) AndAlso (Not ucrSaveTrainingData.bUserTyped) Then
+            If rdoSample.Checked Then
+                ucrSaveTestingData.SetPrefix(strDataframeName & "testing_data")
+                ucrSaveTrainingData.SetPrefix(strDataframeName & "training_data")
+            Else
+                ucrSaveTestingData.SetPrefix(strDataframeName & "testing_data")
+                ucrSaveTrainingData.SetPrefix(strDataframeName & "training_data")
+            End If
+        End If
     End Sub
```
N-thony, line 206: This sub is called no where, do we still need it?

N-thony (changes requested):

lilyclements (commented):
Looking great! Very minor points:

1. We should have a label above the receiver. If that is a stratifying variable then it should say "Stratifying variable (optional):"
2. What is the use of the checkbox? It only links to "pool", but the pool parameter always runs - 0.1 by default. I think we should remove this checkbox. What do you think?
3. Maybe needs to be a bit clearer on what "pool" is doing (see the help file)
4. The nud for breaks does not reset if I hit reset

Otherwise it's looking great - really nice job!

derekagorhom:
> Looking great! Very minor points:
> 
>     1. We should have a label above the receiver. If that is a stratifying variable then it should say "Stratifying variable (optional):"
> 
>     2. What is the use of the checkbox? It only links to "pool", but the pool parameter always runs - 0.1 by default. I think we should remove this checkbox. What do you think?
> 
>     3. Maybe needs to be a bit clearer on what "pool" is doing (see the help file)
> 
>     4. The nud for breaks does not reset if I hit reset
> 
> 
> Otherwise it's looking great - really nice job!

thanks i will work on the changes

derekagorhom:
@lilyclements i have made the changes, can you have a look and review
thanks

lilyclements (dismissed):
@derekagorhom looks good! I made some very minor design changes (removing an additional space where there shouldn't be one, aligning nuds, etc).

I am happy to approve. This is looking great!

derekagorhom:
@lloyddewit can you review this also

rdstern:
@derekagorhom I agree with @lilyclements that this is now looking very nice.  However, with apologies for coming late on the scene, I still have a few points and queries.
a) Unless I have misunderstood the Breaks and Pool are only relevant when there is a stratifying variable?  If so, could they only become visible when there is a variable there.
b) Derek could you check in the documentation that breaks is only when you are stratifying by a numeric variable.  If so, then either we can omit the control, because you currently only allow Factors in the data selector.  Or you could also allow the other variables too - not character. 
c) In the Pool the up-down goes in steps of 0.1.  I wanted to make it 0.01 and couldn't.  Could you go in steps of 0.01 instead?
d) I think the fraction is the Fraction of data in the training set?  So I find it odd that this is the second of the two.  Could you make it the first?  And call it Save Train to: which is more informative than Sample.
e) I'd like @lilyclements to check on this point, but I sometimes might just want a (say) 10% sample of the data - so the training set, without getting the other 90%.  I don't think we have another dialogue for that?  So could the Second have a checkbox with Label  `Save Test:`  Default is checked.  If unchecked then the control is made invisible.
f) Another @lilyclements question, from an example I tried and that worked fine.  The documentation says don't set Pool less than 0.1.  I wonder why they say that?
I tried with a 10% sample from the diamonds data.  I stratified by the cut variable.  The default didn't give me exactly 10% of any factor level.  I changed the pool to 0 and then it did.  That seemed nice.  Why do they imply that is so wrong?
I'd actually like to stratify by all 3 of the factors - which I can do through the Combine Factors dialogue.  But then very few (perhaps none, though I haven't tried) of the categories will be more than 10%?  So, I am quite keen on not having pooling!

derekagorhom:
@rdstern i have made the changes for a,b,c and d can you have a look
thanks

derekagorhom:
@rdstern and @lilyclements i made the changes roger suggested, could you review them

lilyclements:
@rdstern 

To e) I think we do not have a "Sample data frame" dialog, so this sounds good to me.

To f) The default is 0.1, and they recommend to not use under 0.1. However, I agree that while a recommendation, it can still be done so we should offer it. What's your thoughts on having a tool tip pop up if the user reduces it below 0.1 to check they are sure - or do you feel that is a bit too much?

rdstern:
@derekagorhom in my comments from above:
a) I think you have kept the control visible, and just made the up-down invisible.  I wonder why?  I was assuming the whole controls would not be visible, so it is a simpler looking dialogue when there is no stratifying variable.
b) You have done.
c) You have made it move in steps of 0.01, but I think it started at the default of 0.0, now, where it was (I think) 0.1 before.  Could it still start at 0.1.
d) You have renamed it to Save Train to as I hoped.  But I also asked for it to be first, where it is still the second.
e) You have ignored e).  Could you not have Save Test to, but just `Save Test: ` and with a checkbox, default checked, so you could uncheck and then just save the training set, and not have a test set.
f) And a new one.  I tried Reset after checking the Pool and breaks and it didn't seem to reset those settings.  It did other things though.
g) And I have just noticed you have breaks.  It should be Breaks.

rdstern:
@derekagorhom I am kind of losing the will to live here!  You really must work/read more carefully or ask me if there are things you are not sure you understand.

And you must make sure you really understand what is being done in the dialogue.  Programming and hoping it might be alright is not going to work.

a) Here is the dialogue when I first start:
image

I hope you can see that this is wrong.  There is no stratifying variable, but the two receivers are initially visible.  If I tick the Stratifying variable checkbox, then that's fine, and now if I untick it, they go away.

But that isn't what is needed.  **_You don't need the checkbox there._**  They should simply not be visible when there is no stratifying variable. 

b) You now have the Train and test swapped round, - at last.  Now it is saving the Test variable that needs the checkbox.  It should be optional.  In the dialogue the user needs to save the Training set, but not necessarily the Testing one.  However the default is to save both, so start with the `Save Test` checkbox checked.

c) And when you have made the changes please try them out before expecting me to look again.  You can use the diamonds data set as I show above.  And try just saving the training set.  Then try again saving both.

d) Then tell me when I can check again.

derekagorhom:
> @derekagorhom I am kind of losing the will to live here! You really must work/read more carefully or ask me if there are things you are not sure you understand.
> 
> And you must make sure you really understand what is being done in the dialogue. Programming and hoping it might be alright is not going to work.
> 
> a) Here is the dialogue when I first start: image
> 
> I hope you can see that this is wrong. There is no stratifying variable, but the two receivers are initially visible. If I tick the Stratifying variable checkbox, then that's fine, and now if I untick it, they go away.
> 
> But that isn't what is needed. **_You don't need the checkbox there._** They should simply not be visible when there is no stratifying variable.
> 
> b) You now have the Train and test swapped round, - at last. Now it is saving the Test variable that needs the checkbox. It should be optional. In the dialogue the user needs to save the Training set, but not necessarily the Testing one. However the default is to save both, so start with the `Save Test` checkbox checked.
> 
> c) And when you have made the changes please try them out before expecting me to look again. You can use the diamonds data set as I show above. And try just saving the training set. Then try again saving both.
> 
> d) Then tell me when I can check again.

sorry @rdstern i was not done yet with this PR.

derekagorhom:
i just pushed this changes so that i could work on PR #7991

rdstern (changes requested):
@derekagorhom I wanted to check this pull request, but the Random Split now seems to be disabled.  Could you please enable it?

derekagorhom:
@rdstern the dialogue has not been disabled

lloyddewit:
> @rdstern the dialogue has not been disabled

@rdstern I tested and I have the same behaviour as @derekagorhom  (dialog is not disabled). Could you try again and let us know if you still have problems? thanks

rdstern:
@derekagorhom this now seems to work fine.
a) Could you please change Save Test Variable into `Save Test Data`.  
b) And swap its position with the Lag, so it is lined up on the left-hand side of the dialogue.
c) And explain please, what the lag does in the time-series option on the dialogue?  I am curious at the step going in steps of 0.5?

derekagorhom:
@lilyclements could you review this and check if you are ok with the value for lag in the time series.
thanks

lilyclements (commented):
@derekagorhom at the moment we have
1. Just a checkbox for "Save test data"
2. The checkbox has a colon which leads to nothing
3. When checked it makes a ucr visible at a different point of the dialog.
4. Then currently there's not a checkbox for Save training data. 

I thought we would have two checkboxes: 

[] Save test data
[] Save training data

Then when checked, the ucrInput comes up _next_ to them.
They shouldn't have colons.


As for the lag:
1. Remove the colon from "Lag:" because it leads to nothing when unchecked
2. Shift the checkbox to be in line with the "F" from the "Fraction" label

Otherwise the lag parameter works well. Nice job

derekagorhom:
@lilyclements can you please test the dialogue again

N-thony:
> @lilyclements can you please test the dialogue again

@lilyclements can you test?

lilyclements (changes requested):
@derekagorhom This is looking really great - and thank you for being so patient over the past few weeks. This dialog is looking great, so it's clear you have been very thorough, it is good to also try clicking on bits and changing values and pressing "Reset" to see if everything resets correctly. There are only two things I noticed here:

1. The "Lag" checkbox does not reset to unchecked when you click reset
2. Presumably we want the new data frame names to reset to still have their prefixes? At the moment, when you press reset, the new data frame names just go back to "training_data" and "testing_data" and the prefixes go.
So, at the moment I have created a new data frame with the training data set, I reopen the dialog, and I have by default the name "training_data_1". This is good, and what we want as the new data frame name. This should be the new default name. If I change this to "shdsajdhk", and press reset, we want it to change back to "training_data_1". At the moment, it changes to "training_data" which would be a problem since it would overwrite the users current data set called "training_data" if they didn't notice.

Does this make sense? I think the "Random Subsets" dialog should have the prefixes set up in the way you want them, so it might be useful to have a look at that.

derekagorhom:
@lilyclements I made the changes you requested above but i have a little problem with the b. 
after i looked into the random subset i realized to save the `dataframe`, it picks name of the dataset first (here Survey) 
and the prefixes are added to it (in our case either training_data or testing_data) which creates further problems when the `dataframe` name becomes too wordy eg. `"survey_training_data"`, `"survey_training_data_testing_data"`, `"survey_training_data_training_data"`. 


I assumed maybe you wanted it to show only the prefix being `_1` but the random subset chooses from `available dataframes` and then adds the prefix so in our case it would have just picked `survey` and added `_1` to regardless of which `ucrsave` was selected.

instat/dlgRandomSplit.vb
```diff
@@ -1,5 +1,217 @@
...
+        ucrReceiverRanSplit.SetRCode(clsInitialSplit, bReset)
+        ucrChkStratifyingFactor.SetRCode(clsInitialSplit, bReset)
+
+        If bReset Then
+            ucrChkTest.SetRCode(clsInitialSplit, bReset)
+        End If
+    End Sub
+
+    Private Sub TestOkEnabled()
+        If ucrSaveTestingData.IsComplete Or ucrSaveTrainingData.IsComplete Then
```
lilyclements, line 165: We are missing some arguments here

lilyclements (changes requested):
@derekagorhom if I open the dialog, and press OK, I’m getting two data frames as the output (both the testing and training data), but only one of them should run.

I don’t know if I follow the issue with the ucrSave. Are you saying this is a general problem elsewhere?

derekagorhom:
@lilyclements I made the changes. this should work fine now

lilyclements:
@derekagorhom the TestOK still has issues - namely, I can press OK and have the nud's empty. They need a statement for if the radio button is selected and nud is empty, OK shouldn't be enabled (and for the other nud, if the radio button and checkbox are selected and nud is empty).

instat/dlgRandomSplit.vb
```diff
@@ -130,11 +130,9 @@ Public Class dlgRandomSplit
 
         clsTraining.SetPackageName("rsample")
         clsTraining.SetRCommand("training")
-        clsTraining.SetAssignTo("training_data", strTempDataframe:=ucrSelectorRandomSplit.strCurrentDataFrame, strDataFrameNames:="last_training")
 
         clsTesting.SetPackageName("rsample")
         clsTesting.SetRCommand("testing")
-        clsTesting.SetAssignTo("testing_data", strTempDataframe:=ucrSelectorRandomSplit.strCurrentDataFrame, strDataFrameNames:="last_testing")
```
lilyclements: Out of interest why did you remove these `SetAssignTo`'s?

derekagorhom: @lilyclements i removed them from here and added them to 
`Private Sub NewDefaultName()`
 you can check lines 200 - 202

lilyclements (commented):
Looks good! Can you put all these ucrs in the sub that calls `TestOKEnabled()`?

Edit - on second inspection, the TestOK won't work as it should here! Have a look at the dialog and work with what we want. We want it to be that if something is visible and empty, then OK shouldn't be enabled.

That is,
* If Sample is checked and Stratifying Variable is checked and the ucrReceiver is empty **or** the Breaks nud is empty **or** the pool nud is empty, then OK should not be enabled
* Regardless of what is checked: If the fraction nud is empty, or either save is not complete, then OK should not be enabled
* If Time Series is checked and Lag is checked and the Lag nud is empty, then OK should not be enabled. 

Then any control we are having in the check here for the TestOK, we need to call the TestOK sub

lilyclements:
@derekagorhom I see you've made some changes. Is this ready to review?

derekagorhom:
@lilyclements i made the changes and they work fine now, i wanted to ask if the max limit set on the `ucrNudlag` was ok with you (i set to 100)

lilyclements:
@derekagorhom good point - as on issue #7227, `lag` takes any numerical integer - this means that we want to set it to have 0 decimal points, and the min is the VB minimum and the max is the VB maximum.

derekagorhom:
@lilyclements i have made the changes, you review this now

instat/dlgRandomSplit.vb
```diff
@@ -59,7 +59,7 @@ Public Class dlgRandomSplit
         ucrChkStratifyingFactor.AddToLinkedControls({ucrNudPool, ucrReceiverRanSplit}, {True}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True, bNewLinkedUpdateFunction:=True, bNewLinkedChangeToDefaultState:=True, objNewDefaultState:=0.1)
 
         ucrNudLag.SetParameter(New RParameter("lag", 3))
-        ucrNudLag.SetMinMax(0, 100)
+        ucrNudLag.SetMinMax(Integer.MinValue, Integer.MaxValue)
         ucrNudLag.Increment = 0.5
         ucrNudLag.DecimalPlaces = 2
```
lilyclements, line 64: lag should not have any decimal places

lilyclements (changes requested):
image

When I open the dialog, OK is not enabled. Can you explain which ucr here needs to be filled/sorted for OK to be enabled?

derekagorhom:
> image
> 
> When I open the dialog, OK is not enabled. Can you explain which ucr here needs to be filled/sorted for OK to be enabled?

For Sample, `Stratifying Variable` should be checked and `ucrReceiverRanSplit` needs to be filled (also none of the `ucrNuds` needs to be empty), 
For Time series; `ucrChklag` needs to be checked (also none of the `ucrNuds` is supposed to be empty)

lilyclements:
> For Sample, `Stratifying Variable` should be checked and `ucrReceiverRanSplit` needs to be filled (also none of the `ucrNuds` needs to be empty), For Time series; `ucrChklag` needs to be checked (also none of the `ucrNuds` is supposed to be empty)

Good, I have two follow up points

1. From what you have said, then what needs to be sorted here for OK to be enabled?

image

2. Thank you for answering, it has helped clear up a bit of the confusion - I'll try to make sure I'm a bit clearer in the future :)
But, actually, for Sample, _if_ `Stratifying Variable` is checked, then `ucrReceiverRanSplit` needs to be filled (also none of the `ucrNuds` needs to be empty)
For Time series, _if_ `ucrChklag` is checked, none of the `ucrNuds` that are connected to that `ucrChklag` should be empty.

derekagorhom:
@lilyclements i have made the changes now, can you have a look 
thanks

lilyclements (commented):
@derekagorhom this is working a lot better. But OK is still not enabled all the times that it should be. 

For example, what is the reason for OK to not be enabled here:

image

We want to say **if** time series is checked and **if** lag is checked then ...
At the moment, I find that OK is only enabled **if time series and lag are checked**.
Do you see the difference there?

instat/dlgRandomSplit.vb
```diff
@@ -158,15 +158,36 @@ Public Class dlgRandomSplit
...
+                    If ucrSaveTrainingData.IsComplete AndAlso ucrSaveTestingData.IsComplete AndAlso Not ucrNudFraction.IsEmpty Then
+                        ucrBase.OKEnabled(True)
+                    Else
+                        ucrBase.OKEnabled(False)
+                    End If
+                Else
+                    ucrBase.OKEnabled(False)
+                End If
             Else
                 ucrBase.OKEnabled(False)
```
lilyclements, line 173: Why is this `FALSE`?
If `rdoSample` is checked, and we do not have any of these problems, then OK can be enabled 
Same for `rdoTimeSeries`

instat/dlgRandomSplit.vb
```diff
@@ -158,15 +158,36 @@ Public Class dlgRandomSplit
...
 
     Private Sub TestOkEnabled()
-        If (rdoSample.Checked AndAlso ucrChkStratifyingFactor.Checked AndAlso Not ucrReceiverRanSplit.IsEmpty AndAlso Not ucrNudBreaks.IsEmpty AndAlso Not ucrNudPool.IsEmpty) OrElse
-            (rdoTimeSeries.Checked AndAlso ucrChkLag.Checked AndAlso Not ucrNudLag.IsEmpty) Then
-            If ucrSaveTrainingData.IsComplete AndAlso ucrSaveTestingData.IsComplete AndAlso Not ucrNudFraction.IsEmpty Then
-                ucrBase.OKEnabled(True)
+        If rdoSample.Checked Then
+            If ucrChkStratifyingFactor.Checked Then
+                If Not ucrReceiverRanSplit.IsEmpty AndAlso Not ucrNudBreaks.IsEmpty AndAlso Not ucrNudPool.IsEmpty Then
+                    If ucrSaveTrainingData.IsComplete AndAlso ucrSaveTestingData.IsComplete AndAlso Not ucrNudFraction.IsEmpty Then
```
lilyclements, line 164: These ucrs are not controlled by the `ucrChkStratifyingFactor`. So therefore whether `ucrChkStratifyingFactor` is checked or not should not affect the state of these three controls

lilyclements: This is because these three controls affect the whole dialog, then to avoid repeating code, perhaps it makes sense to bring them to the "exterior" of the if-statement.

```
If ucrSaveTrainingData.IsComplete AndAlso ucrSaveTestingData.IsComplete AndAlso Not ucrNudFraction.IsEmpty Then
  < If statements for when rdos are checked > 
Else
   ucrBase.OKEnabled(False)
End If
```

Then we can build it up from there. So, next we can add in our if statement for rdoSample and rdoTimeSeries:

```
If ucrSaveTrainingData.IsComplete AndAlso ucrSaveTestingData.IsComplete AndAlso Not ucrNudFraction.IsEmpty Then
  If rdoSample.Checked Then 
    '<if statements for when rdo sample is checked >
  Else ' this here is if rdoTimeSeries.Checked
    '<if statements for when rdo time series is checked >
  End If
Else
   ucrBase.OKEnabled(False)
End If
```


Step 3: What if rdoSample is checked then? From looking at the dialog and the controls, there is a checkbox for "Stratifying variable". This brings up three controls. So we want to add in a statement that now says, `if ucrChkStratifyingVariable.Checked Then < new controls cannot be empty for OK to be enabled >`
Perfect though - since you have already done this on lines 162-163. So we can just add that in :) 
Then we just need to add in what if ucrChkStratifyingFactor isnt checked? Then OK should work fine.

```
If ucrSaveTrainingData.IsComplete AndAlso ucrSaveTestingData.IsComplete AndAlso Not ucrNudFraction.IsEmpty Then
  If rdoSample.Checked Then 
    If ucrChkStratifyingFactor.Checked Then
                If Not ucrReceiverRanSplit.IsEmpty AndAlso Not ucrNudBreaks.IsEmpty AndAlso Not ucrNudPool.IsEmpty Then
                   ucrBase.OKEnabled(True)
                Else
                   ucrBase.OKEnabled(False)
               End If
     Else ' And what if ucrChkStratifyingFactor isnt checked? Then OK should work fine.
             ucrBase.OKEnabled(True)
  Else ' this here is if rdoTimeSeries.Checked
    '<if statements for when rdo time series is checked >
  End If
Else
   ucrBase.OKEnabled(False)
End If
```

Then we just do the same for rdoTimeSeries.Checked.

Does this make sense?

derekagorhom:
@lilyclements i have made some changes to the testOK, can you have a look

instat/dlgRandomSplit.vb
```diff
@@ -158,37 +158,33 @@ Public Class dlgRandomSplit
     End Sub
 
     Private Sub TestOkEnabled()
-        If rdoSample.Checked Then
-            If ucrChkStratifyingFactor.Checked Then
-                If Not ucrReceiverRanSplit.IsEmpty AndAlso Not ucrNudBreaks.IsEmpty AndAlso Not ucrNudPool.IsEmpty Then
-                    If ucrSaveTrainingData.IsComplete AndAlso ucrSaveTestingData.IsComplete AndAlso Not ucrNudFraction.IsEmpty Then
+        If Not ucrSaveTrainingData.IsComplete Or Not ucrSaveTestingData.IsComplete Or ucrNudFraction.IsEmpty Then
```
lilyclements, line 161: out of interest, what's the difference between `or` and `orElse`?

lilyclements (commented):
Looks great and works great. Really nice job!

@rdstern @lloyddewit over to you to review!

instat/dlgRandomSplit.vb
derekagorhom: `Or` requires both expression to be evaluated while `OrElse` would still work even if only one expression is evaluated. so in this case if `ucrSaveTrainingData` (or `ucrsaveTestingData/UcrNudfraction`)  is empty then the `TestOK` will be disabled regardless of if the other controls are true.

lilyclements: Makes sense - thanks :)

lloyddewit: > Or requires both expression to be evaluated while OrElse would still work even if only one expression is evaluated. so in this case if ucrSaveTrainingData (or ucrsaveTestingData/UcrNudfraction) is empty then the TestOK will be disabled regardless of if the other controls are true.

@derekagorhom @lilyclements I'm not sure I completely agree with this. Both `Or` and `OrElse` provide a logical 'Or' operation and will always return the same logical result.
The only difference is that `OrElse` is a short-circuiting operator, `Or` is not. This means that with `OrElse` if the left-hand side is true then VB will not bother to evaluate the right-hand side (because the 'Or' condition has already been met). In contrast `Or` will always evaluate both sides. There may be rare conditions when you want to do this, e.g. if you want to force 2 functions to be called:
```
If myFunction1() Or myFunction2() Then
```
But this could be confusing for other developers so better to call functions explicitly above the 'If'.
`OrElse` is useful for avoiding exceptions, e.g.:
```
If myObject Is Nothing OrElse myObject.name = 'Derrick' Then
```
If we used `Or` above, and `myObject` was nothing, then it would raise an exception.
In summary, `OrElse` is more efficient and may be safer. Therefore best to always use `OrElse` (apart from very rare circumstances).
Does this make sense?

lloyddewit (changes requested):
@derekagorhom Thank you, this looks good.
Please could you read my comment and if you agree, then change the `Or`s to `OrElse`? It's not actually essential but it makes the code more consistent with the rest of R-Instat.
If you disagree, then that's also fine, I'm happy to discuss.
Thanks

derekagorhom:
@lloyddewit i have made the changes.... OrElse works fine also. thank you for the suggestion

N-thony:
> @rdstern Please could you test? thanks

@rdstern any plan about testing this?

rdstern (approved):
This seems to be working.  @derekagorhom there needs to be a possibly seperate pull request to add the rsplit package to R-Instat.
As a detail I found it odd that there was the usual data selector, when there was no receiver - unless you included the stratified option.  But I found it was working well - at least for the ordinary option.  I didn't check the time series.  I tried with the hh data from the MICS survey and it seems to be working fine.  So I am approving.

lloyddewit (approved):

Issue #7279, opened 2022-03-08
Corrections and an oddity in the climatic data entry dialogue


rdstern:
I have been using the Climatic > Tidy and Examine > Daily Data Entry dialogue as an editing facility to correct daily data for 2 stations in Tajikistan.  This is following the quality control, that showed a number of oddities.  For example there were 97 occasions with tmax <= tmin.  I now need to correct the data.  I am unable to get access to the paper records, and need to see the odd days in context.
I am finding the data entry tool to be excellent for this in a variety of ways.  

![image](https://user-images.githubusercontent.com/11226469/157190394-c2c69e69-b986-4006-b184-09ecd3c41624.png)

There are a couple of small improvements I suggest, and then an oddity that is shown here - a bug, perhaps to be found.
a) Please change the menu name to `Daily Data Editing/Entry` .  
b) And make the same change in the title of the dialogue.
c) When you return to the data entry after coming out of data entry it always asks if you really want to, because you will lose what was entered earlier.  Here is an example of the message:
![image](https://user-images.githubusercontent.com/11226469/157191414-12c1d2a7-f7e0-4ed7-9d68-a2112ad8d4e9.png)

Now, once you have entered, or corrected data, the number of corrections is shown in red at the bottom.  But the message appears whether, or not, you have made any corrections.  In the example above, I went into the sub-dialogue, then came straight out, and am now returning.  The message should not be given unless there are values to correct.

I hope these are quick and easy to do, because they include an edit to frmmain.

d) There is a bug that has just appeared.  It is that yellow box in the figure above.  This has appeared and is very annoying.  It is very persistent.  I have reset the main dialogue, and also the sub-dialogue.  But it has just appeared (I now realise it is in the master I should say - I was using Version 7.3 earlier??? and didn't get this!)    I'll start again - with the master, and see if it is a consistent problem.  Then - if so, I'll check with other versions.

I am suggesting @N-thony could fix at least a) to c) easily.  I really don't understand d) and will check further.

N-thony:
@rdstern I'm not able to reproduce item d), please check again otherwise we can close this?

rdstern:
[...]

PR #7520, opened 2022-06-02
Fixed minor bug in Climatic>Tidy and Examine>Data daily Dialog

EstherNjeriLiberatta:
Fixes (partially) #7279 
@rdstern Fixed item (a.)  It is now possible to add a comment without changing a day/more and test ok will be enabled.


lloyddewit (dismissed):

rdstern:
It seemed to allow ok, but then gave the error below:
image

N-thony:
> 

@rdstern I think this need to be solved on the R side, the `add_new_comment` function is executed inside `save_data_entry_data` function where they are some check that doesn't the functions to execute if there is nothing changed, I think @dannyparsons is the right person to advice on this.

EstherNjeriLiberatta:
@lilyclements I'll give you a call on skype.

N-thony:
@EstherNjeriLiberatta any progress?

EstherNjeriLiberatta:
@N-thony Thanks for checking, but I'm still waiting for @lilyclements response to the last comment you made.

lilyclements:
> @lilyclements we still need the `data changed` i.e `new_data` to pass this check on line 4287. What do you think? image

@N-thony I don't think I follow. Can you explain the problem here, and why we need the `new_data` to pass this check? Thanks.

N-thony:
> > @lilyclements we still need the `data changed` i.e `new_data` to pass this check on line 4287. What do you think? image
> 
> @N-thony I don't think I follow. Can you explain the problem here, and why we need the `new_data` to pass this check? Thanks.

@lilyclements could we have a quick call?

N-thony:
> @EstherNjeriLiberatta to sort this, you need to read in the `rows_changed` parameter (as well as the `row` option in the `comments_list`) and take the value of the row that has a comment associated with it.

@lilyclements looking at the code and trying to implement your idea, I find a difficulty that is when we only add a comment from where we read the `row` to pass it to `row_changed` it is fine but there is this check on line `4287` when the data change but in our case we didn't change the data, only added the comment.

N-thony:
> @N-thony @EstherNjeriLiberatta I suggest we have a call together and talk this through

@lilyclements let have a call.

EstherNjeriLiberatta:
@lilyclements @N-thony Thanks for your input the dialogue works okay now. 

lloyddewit (approved):

lloyddewit:
@N-thony Please could you test/approve? thanks

N-thony (approved):
Good, @rdstern could you test too? thanks

rdstern (approved):
@EstherNjeriLiberatta that's great.  I should have checked this earleir, because we could have used the feature on the Rwanda workshop this week.

PR #7526, opened 2022-06-06
Ensured that Column Metadata window opens with the default selected dataset in the Data View

### LINKED ISSUE #7477: Column Metadata doesn't open with the default selected dataset in the Data View

N-thony:
Fixes #7477 
@rdstern I have made the improvement to the column meta data to change to the current data frame when the current data frame has changed from data view. Please have a look. Thanks
@derekagorhom please have a look.

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -1505,6 +1505,10 @@ Public Class frmMain
         ucrColumnMeta.SetCurrentDataFrame(iIndex)
     End Sub
 
+    Public Sub UpCurrentColumnMetaDataFrame(strDataName As String)
+        ucrColumnMeta.SetCurrentDataFrame(strDataName)
+    End Sub
+
     Private Sub CummulativeDistributionToolStripMenuItem_Click(sender As Object, e As EventArgs) Handles mnuDescribeSpecificCummulativeDistribution.Click
         dlgCumulativeDistribution.ShowDialog()
     End Sub
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -1505,10 +1505,6 @@ Public Class frmMain
         ucrColumnMeta.SetCurrentDataFrame(iIndex)
     End Sub
 
-    Public Sub UpCurrentColumnMetaDataFrame(strDataName As String)
-        ucrColumnMeta.SetCurrentDataFrame(strDataName)
-    End Sub
-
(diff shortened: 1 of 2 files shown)


instat/frmMain.vb
```diff
@@ -1505,6 +1505,10 @@ Public Class frmMain
         ucrColumnMeta.SetCurrentDataFrame(iIndex)
     End Sub
 
+    Public Sub UpCurrentColumnMetaDataFrame(strDataName As String)
+        ucrColumnMeta.SetCurrentDataFrame(strDataName)
+    End Sub
+
```
lloyddewit, line 1511: Remove (see comment below)?
```suggestion
```

instat/ucrDataView.vb
```diff
@@ -230,6 +230,7 @@ Public Class ucrDataView
     End Sub
 
     Public Sub CurrentWorksheetChanged()
+        frmMain.UpCurrentColumnMetaDataFrame(_grid.CurrentWorksheet.Name)
```
lloyddewit, line 233: `frmMain.ucrColumnMeta` is public. Would it be simpler just to call like this?
```suggestion
        frmMain.ucrColumnMeta.SetCurrentDataFrame(_grid.CurrentWorksheet.Name)
```

lloyddewit (changes requested):

instat/ucrDataView.vb
N-thony: @lloyddewit this is resolved, thanks.

lloyddewit (approved):
@N-thony Thanks, looks good

lloyddewit:

rdstern (approved):
@N-thony great.  I am also very happy that these are changes proposed by other than me.  Also they are relatively small, but add to the convenience, and that is important for R-Instat.  Our catch-phrase is "making it easy" and this sort of thing helps!

N-thony:
@lloyddewit I have tried to test this with Linux version unfornately I was getting the following bug, I have fixed it in PR #7537
image

PR #7530, opened 2022-06-07
Minor bug fixed in the File > New Data Frame

### LINKED ISSUE #7529: Minor bug with File > New Data Frame
This bug was found by @berylwaswa when recording a video with this dialogue, the bug occurs when there is space or any other character that R doesn't support in column name.
image

N-thony:
Fixes #7529
This fix bug when creating a new data frame with column name containing character that R doesn't support like space...

### DIFF
--- a/instat/dlgNewDataFrame.vb
+++ b/instat/dlgNewDataFrame.vb
@@ -368,7 +368,7 @@ Public Class dlgNewDataFrame
                     clsEmptyRepFunction.AddParameter("x", Chr(34) & strDefault & Chr(34), bIncludeArgumentName:=False, iPosition:=0)
                 End If
 
-                Dim strColumnName As String = row.Cells("colNames").Value
+                Dim strColumnName As String = ValidateRVariable(row.Cells("colNames").Value)
                 If (strType = "Integer" OrElse strType = "Numeric") AndAlso Not strDefault = "NA" _
                                         AndAlso IsNumeric(strDefault) AndAlso strDefault.Contains(",") Then
 
@@ -412,6 +412,15 @@ Public Class dlgNewDataFrame
         strTemp &= ")"
 
         Return strTemp
+    End Function
+
+    Private Function ValidateRVariable(strNewDataText As String) As String
+        For Each chrCurr In strNewDataText
+            If Not Char.IsLetterOrDigit(chrCurr) AndAlso Not chrCurr = "." AndAlso Not chrCurr = "_" Then
+                strNewDataText = strNewDataText.Replace(chrCurr, ".")
+            End If
+        Next
+        Return strNewDataText
     End Function
 
     Private Function GetLabelAsRString(lstLabels As List(Of String)) As String


rdstern:
@N-thony this looks pretty good.  I could approve, but changing x1 to 1x, i.e. starting with a number, still gives an error.  Is there a standard function that copes with this situation as well?

lloyddewit:
> @N-thony this looks pretty good. I could approve, but changing x1 to 1x, i.e. starting with a number, still gives an error. Is there a standard function that copes with this situation as well?

@N-thony The R syntatic name rules are quite complicated (e.g. back ticks can be used and reserved key words are invalid). Rather than trying to implement these rules in VB, it may be better to use the R [`make.names()` function](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/make.names). This converts a string into a syntactically valid name (or if the string is already valid, then returns the string unchanged). This function will also be updated automatically if/when the R syntatactic rules change.

lloyddewit:
@rdstern @N-thony I also noticed that TidyVerse has its own guidelines for making names syntactic (see [section 3.6.2 here](https://principles.tidyverse.org/names-attribute.html#universal-names)). Should we follow the Tidyverse approach or is it unimportant?

rdstern:
We are trying to follow tidyverse where possible, but here I would not want us to force a name to be consistent with tidyverse.  If a user wanted a legal name that was not tidy, that's still up to them to be able to do.  So we just need the general function that will produce legal names, and so not produce an error.

instat/dlgNewDataFrame.vb
```diff
@@ -414,6 +415,20 @@ Public Class dlgNewDataFrame
...
     End Function
 
+    Private Function ValidateRVariable(strNewDataText As String) As String
+        For Each chrCurr In strNewDataText
+            If Not Char.IsLetterOrDigit(chrCurr) AndAlso Not chrCurr = "." AndAlso Not chrCurr = "_" Then
+                strNewDataText = strNewDataText.Replace(chrCurr, ".")
+            End If
+        Next
+        Dim chrfirst = strNewDataText.ToCharArray()
+        Dim match As Match = Regex.Match(chrfirst(0), "^[a-zA-Z ]*$")
```
lloyddewit, line 425: > A syntactically valid name consists of letters, numbers and the dot or underline characters and starts with a letter or the dot not followed by a number
[see [here](https://stat.ethz.ch/R-manual/R-patched/library/base/html/make.names.html)]

- This line allows the first character to be a space, is that correct?
- R reserved words (see [here](https://stat.ethz.ch/R-manual/R-patched/library/base/html/Reserved.html)) are illegal but this check does not catch them
- Starting with a dot is valid but this check does not allow it
- Having a number as the 2nd character is not allowed but this code does not catch this

As mentioned before, it may be simpler to use the R [`make.names()` function](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/make.names). An example of using an R function instead of VB code can be found in the `clsRLink.vb` `GetRFunctionDefinitionParameters()` function.

lloyddewit (changes requested):

instat/dlgNewDataFrame.vb
N-thony: @rdstern @lloyddewit have a look.

N-thony:
@lloyddewit if you are happy with this change, I can open another PR or make the change here for dlgName where I something similar i.e `ValidateRVariable` is needed, what do you think?

instat/dlgNewDataFrame.vb
```diff
@@ -16,6 +16,7 @@
 
 Imports instat.Translations
 Imports RDotNet
+Imports System.Text.RegularExpressions
```
lloyddewit, line 19:
```suggestion
```

instat/dlgNewDataFrame.vb
```diff
@@ -414,6 +415,14 @@ Public Class dlgNewDataFrame
         Return strTemp
     End Function
 
+    Private Function ValidateRVariable(strNewDataText As String) As String
+        Dim clsMakeNameFunction As New RFunction
+        clsMakeNameFunction.SetRCommand("make.names")
+        clsMakeNameFunction.AddParameter("names", Chr(34) & strNewDataText & Chr(34), iPosition:=0)
+        clsMakeNameFunction.AddParameter("unique", "TRUE", iPosition:=1)
```
lloyddewit, line 422: I think this is only used if we pass a list of strings
```suggestion
```

lloyddewit (changes requested):
@N-thony This looks nice and clean - thank you!
just a couple of small suggestions

N-thony:
> @N-thony This looks nice and clean - thank you! just a couple of small suggestions

Done

Patowhiz:
@N-thony following your conversations with @lloyddewit on this PR.

@N-thony I'm I correct to think `ValidateRVariable` function implementation is basically returning a valid R text? If so we have a function in `clsRLink` called `MakeValidText`. That's what the import dialog uses for making valid data frame names input by the user.

@N-thony @lloyddewit I'm not sure if @N-thony should ;

1. Just use that function just like the way the import dialog uses it(by calling it through `frmMain`), or 
2. He generalizes that function implementation in the `mdlCoreControl` module, then use it in this dialog. Then later we make dialogs and controls use the function based on the generalized parameters.

I suggest the 2nd option but happy with either instead of duplicating it here too. Thanks.

lloyddewit (approved):
@N-thony @Patowhiz Yes! This is the best solution of all. Thanks to @Patowhiz 's knowledge of the code.

lloyddewit:
@rdstern If you retest/approve, then we can merge, thanks

lloyddewit:
@N-thony Thanks for this fix.
I understand that this PR only partly fixes the issue, so I changed the PR description to 'Fixes (partially) #7529'.
Is that correct?

PR #7541, opened 2022-06-08
Ensured that data selector(columns, graphs, filters) refreshes when new data is created

### LINKED ISSUE #7539: Check on refreshing of data selector?
This may be a feature and hence not necessarily to be corrected?

With the usual survey data I go into a dialogue, say 1-variable summarise.  I then realise I need a calculated variable.
So I come out the dialogue and use the calculator to produce a new variable, called calc.

Now I return to the dialogue.  The selector is as I left it, i.e. calc isn't there.  If I press reset, then it is there.  

A second example.  I go into a dialogue (say summarise) and use the data options button.  There are no filters defined.  I come out of the dialogue and use the Filter system and Filter from Factors and choose village and make all 4 villages into possible filters.  Now I return to the (summarise) dialogue and use the Data Option.  The new filters are not there.  I choose define new filter, and then return and lo and behold, they are now there?

Patowhiz:
Fixes #7539.

### DIFF
--- a/instat/ucrSelector.vb
+++ b/instat/ucrSelector.vb
@@ -62,6 +62,9 @@ Public Class ucrSelector
         If bFirstLoad Then
             bFirstLoad = False
         End If
+        'always load selector contents on load event because contents may have been changed at R level 
+        'and the control needs to refresh the data frame names.
+        LoadList()
     End Sub
 
     Protected Sub OnResetAll()


lloyddewit (approved):

lloyddewit:
@rdstern if you can test/approve, then we can merge
Thanks

rdstern (approved):
Great

PR #7544, opened 2022-06-09
Fixed bug in selecting of all factors

### LINKED ISSUE #7538: Select all levels control for a factor doesn't seem to work now
I found this in 2 places - there may be more.  Try with the usual survey data.
a) Filter - then Filter from Factor - select all levels doesn't seem to do anything.
b) Filter , Define New filter and choose a factor.  The select all levels again seems to do nothing.

Patowhiz:
Fixes #7538.


instat/ucrFactor.vb
```diff
@@ -746,9 +746,8 @@ Public Class ucrFactor
         For iRowIndex As Integer = 0 To _grdSheet.Rows - 1
             If _grdSheet(iRowIndex, iCheckedColIndex) <> bChecked Then
                 bChangeDetected = True
-            Else
-                _grdSheet(iRowIndex, iCheckedColIndex) = bChecked
             End If
+            _grdSheet(iRowIndex, iCheckedColIndex) = bChecked
```
lloyddewit, line 750: move to inside `If` block?

lloyddewit (changes requested):

rdstern (approved):
Good

instat/ucrFactor.vb
Patowhiz: Agreed and done!

lloyddewit (approved):

PR #7552, opened 2022-06-10
Added menu items under the selector add button

### LINKED ISSUE #7551: Add options similar to the selector right click menu
When a user wants to add all contents of the selector to a multiple receiver, they have to right click then click on Add all option. 
Would be ideal if a user could get this option under the  Add options similar to what we have in the Ok and To Script buttons.

Patowhiz:
Fixes #7551.

### DIFF
--- a/instat/ucrSelectorByDataFrameAddRemove.vb
+++ b/instat/ucrSelectorByDataFrameAddRemove.vb
@@ -14,9 +14,18 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports System.ComponentModel
+
 Public Class ucrSelectorByDataFrameAddRemove
-    Private Sub btnAdd_Click(sender As Object, e As EventArgs) Handles btnAdd.Click
+    Private Sub btnAdd_Click(sender As Object, e As EventArgs) Handles btnAdd.Click, toolStripAddSelected.Click
         Add()
+    End Sub
+
+    Private Sub toolStripAddAll_Click(sender As Object, e As EventArgs) Handles toolStripAddAll.Click
+        AddAll()
+    End Sub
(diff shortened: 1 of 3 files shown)


instat/ucrSelectorByDataFrameAddRemove.vb
```diff
@@ -63,4 +72,16 @@ Public Class ucrSelectorByDataFrameAddRemove
...
+    Private Sub contextMenuStripAdd_Opening(sender As Object, e As CancelEventArgs) Handles contextMenuStripAdd.Opening
+        'todo. this code block should be set under CurrentReceiver change event
+        'once the event is added it can be removed from here.
+        toolStripAddSelected.Enabled = lstAvailableVariable.SelectedItems.Count > 0
+        toolStripAddAll.Enabled = False
+        If CurrentReceiver Is Nothing Then
+            toolStripAddSelected.Enabled = False
+        ElseIf TypeOf CurrentReceiver Is ucrReceiverMultiple Then
+            toolStripAddAll.Enabled = True
+        End If
```
lloyddewit, line 85: is something like this more readable?
```suggestion
        If CurrentReceiver Is Nothing Then
            toolStripAddSelected.Enabled = False
            toolStripAddAll.Enabled = False
        Else
            toolStripAddSelected.Enabled = lstAvailableVariable.SelectedItems.Count > 0
            toolStripAddAll.Enabled = TypeOf CurrentReceiver Is ucrReceiverMultiple
        End If
```

lloyddewit (changes requested):

instat/ucrSelectorByDataFrameAddRemove.vb
Patowhiz: @lloyddewit perfectly clear. Thanks.

lloyddewit (approved):

lloyddewit:
@rdstern if you can test, then we can merge, thanks

PR #7465, opened 2022-05-09
Adding the three Variable functionality to the Describe> Two\Three Summary dialogue

Ivanluv:
This PR has implementation for the three variable functionality to match the  `Three Variable` radio button  option on the Describe> Two\Three Summary dialogue

This is still in progress

PR #7560, opened 2022-06-15
Allowing consistent order of columns in the calculate_summary R function.

lilyclements:
Fixes bug in calculate_summary so that all the returned columns are in a consistent order regardless of if a percentage manipulation is performed or not.

@Ivanluv this should fix the issue you found.

Before, the order of columns would move if a percentage was available. This was inconsistent and creating a small issue in the `summary_calculations` R code.

```
# E.g., before, the following would give the column order as "variety" then "village"
# [1]
data_book$summary_table(data_name="survey",
                        summaries=c("summary_count"),
                        factors=c("variety", "village"))

# but this would give  "village" then "variety"
# [2]
data_book$summary_table(data_name="survey",
                        percentage_type="factors",
                        perc_total_factors="village",
                        summaries=c("summary_count"),
                        factors=c("variety", "village"))

# but now the code from [2] gives the order as "variety" then "village"
```

@Ivanluv this is ready. Should we merge straight into your branch, or should we merge it into the main first? (If your branch is not ready to merge, I suggest we do the latter in case of any other changes)

### DIFF
--- a/instat/static/InstatObject/R/Backend_Components/summary_functions.R
+++ b/instat/static/InstatObject/R/Backend_Components/summary_functions.R
@@ -228,7 +228,9 @@ DataBook$set("public", "calculate_summary", function(data_name, columns_to_summa
   if(return_output) {
     dat <- out$data
     # relocate so that the value_factors are first still for consistency
-    out$data <- (out$data %>% dplyr::select(c(tidyselect::all_of(value_factors), tidyselect::all_of(manip_factors)), tidyselect::everything()))
+      if (percentage_type != "none"){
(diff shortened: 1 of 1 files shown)


lloyddewit:
@lilyclements Thank you for this.
Is this PR related to an existing issue or PR? If so, please could you reference in the PR description (e.g. 'fixes issue #xxxx?').
thanks

lilyclements:
@lloyddewit thanks. It is related to Issue #7465 about a bug @Ivanluv and I discussed over slack. Is it more appropriate for this to be merged into his branch directly, or into the master file?

Ivanluv:
I would prefer to merge this in my branch then have it closed

lilyclements:
@Ivanluv good point! It should be fixed now

PR #7562, opened 2022-06-16
Added a Machine Learning Dialog 

### LINKED ISSUE #6671: Adding dialogues for machine learning - caret
caret appears to be a "one-stop-shop" for machine learning.  It is the only package that has a chapter devoted to it, in the [Introduction to Data Science](https://rafalab.github.io/dsbook/caret.html) book. 

I suggest a small set of dialogues - plus (possibly) keyboards in our Fit Model Keyboard and Use Model Keyboard.

I suggest the first dialogue be called `Machine Learning (caret) `and be under the `General `in the `Model > Fit Model `menu.

It will be for the `train `command in the caret package, plus various options that will imply a trainControl command is also given.
For example the default is to do 25 bootstrap samples.  Sometimes you may want many more - e.g. 250 and that would be done as an option in trainControl.  

The initial dialogue can be quite simple, but it will be useful to try the commands to check on the options needed.  I suggest as follows, to start with.  (Base it on the Fit Model > General
 a) Usual data selector
b) Single receiver for the Response Variable - as in the Fit > General dialogue
c) Also Have a Label Type as in the Fit Model General, but next to the single control.  Ideally it will also give the number of levels, but that is also the case for the Fit Model General.  So now just follow what is there.
d) Then have Explanatory O Variables O Model  - so with 2 radio buttons.  Default is Variables and that gives a Multiple receiver.  If Model is chosen then give the same receiver as in Fit Model > General, but without the keyboard below.
e) Where there is Distribution in the General dialogue, have Model instead and then include the Fit_Model_List.  Is that easy to do?  Perhaps the default is glm.
f) Then there is the Save Model control (as in General.
g) There will be more options added - so leave space for them.  There may be further suggestions before work on the dialogue starts?

anastasia-mbithe:
Fixes #6671 

I had to redo the design, due to outdated controls in the other PR.
This is still in progress.


N-thony:
> Fixes #6671
> 
> This is still in progress.

Great, I made a small change above, always start with Fixes then issue number so that if this PR is merged the issue will close automatically. Thanks.

lloyddewit:
> > Fixes #6671
> > This is still in progress.
> 
> Great, I made a small change above, always start with Fixes then issue number so that if this PR is merged the issue will close automatically. Thanks.

@N-thony Thank you for your eagle eye.
In fact I think what @anastasia-mbithe wrote ('This fixes #6671') also works. You can confirm this by looking at the list of linked issues on the right-hand side:

image


GitHub is not strict about linking issues. It auto links an issue if the PR description has 'fixes' directly in front of the issue number ('closes' and 'resolves' also work). See [here ](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) for details.

No action needed (updated description is fine), this comment is just background.

anastasia-mbithe:
@rdstern, We can have a model in this dialog by selecting more than one Explanatory variable. However, having an interactive model for this dialog is almost impossible, as I had explained to @lilyclements, because of the limitation of the caret function:
# For Regression;
sampled_data <- rsample::initial_split(data=survey, prop=0.7)
train <- rsample::training(sampled_data)
train_data <- dplyr::select(train,  c(field, size, variety))
caret::train(size ~ ., data=train_data, method="knn")

#For Classification
sampled_data <- rsample::initial_split(data=survey, prop=0.7)
train <- rsample::training(sampled_data)
train_data <- dplyr::select(train, c("Size","Yield","Field"))
train_classes <- train $ Village.
caret::train(train_data, train_classes, method="knn")

The train_data in both cases is gotten from the receiver by the select function which does not allow interaction. Do we explore other functions or maintain this, and if so, what's your take on the interactive model. ( where we can use +, *....)

So am actually stuck in this, and am open to more suggestions and comments.

lloyddewit:
@rdstern @anastasia-mbithe 
This has not been worked on for a year. If it's no longer active, then I suggest that we close it. Nothing is lost, and if anyone has time to look at this in the future, then we can reopen.
What are your views?
Thanks

lloyddewit:
Can be reopened if anyone wishes to work on again

PR #7570, opened 2022-06-27
Updated Boxplot dialogue's Plot Options to the new system.

### LINKED ISSUE #7556: Change the specific plots to the new system

EstherNjeriLiberatta:
Fixes #7556
In this PR, I have replaced the `Box Options` button in the Box plot dialogue with the new `Plot Options` button which includes a drop-down.
@rdstern @africanmathsinitiative/developers  This is ready for review.


rdstern (commented):
@N-thony this is great.  I have 2 small cosmetic points - but please only if you agree.  I am not so good on these aspects.
I wonder if the width of the Options button should be the same as the width of the selector?  And aligned with it.
Could there be a horizontal line below the first entry?#

What do you think.  This would apply to each of them.  

It does make the dialogue look simpler, which is a good thing.

N-thony:
> @N-thony this is great. I have 2 small cosmetic points - but please only if you agree. I am not so good on these aspects. I wonder if the width of the Options button should be the same as the width of the selector? And aligned with it. Could there be a horizontal line below the first entry?#
> 
> What do you think. This would apply to each of them.
> 
> It does make the dialogue look simpler, which is a good thing.

@rdstern I agrre with you, @EstherNjeriLiberatta please, could you make these changes?

rdstern:
@EstherNjeriLiberatta please check the horizontal line for yourself, as it may not be needed.  One example is Model > Fit Model > One Variable - the Test button and the button that says Binomial.  But looking at that, I am not sure whether a line is needed, given there are far fewer entries.

EstherNjeriLiberatta:
@rdstern  Thank you I have checked it, and I agree with you. The dialogue is ready for review.

rdstern (changes requested):
@EstherNjeriLiberatta it is doing more than I thought - but also therefore sometimes less!
When I am on boxplot, then I can get the ordinary plot options and boxplot options.  I can press violin, but I still get the boxplot options.  If I check tufte, then I get tufteboxplots.
I was thinking that I might get what I press, even though it isn't sensible.  I am happy that I can't, but is it possible to disable the ones that I can't get?
If I am on boxplot (or tufte), or violin, then I can check the points and this adds the jitter layer.  But I can't get it.  I should be able to.

EstherNjeriLiberatta:
@rdstern I have disabled the options that are not needed when you click on either button box/Tufte, jitter, or violin. It's possible to get the jitter layer when you check the Add points however only on the plot options.

rdstern (changes requested):
@EstherNjeriLiberatta that has been ambitious, to just have those enabled, that relate to the options chosen.  It is working well, except that the boxplot and violin plots allow points to be added.  When that checkbox is used, it is adding a jitter plot, so that option should then be enabled.

EstherNjeriLiberatta:
@rdstern This is ready for review. Thank you.

EstherNjeriLiberatta:
@N-thony Kindly review the changes. Thank you

instat/dlgBoxPlot.vb
```diff
@@ -368,69 +412,143 @@ Public Class dlgBoxplot
...
-            ucrSaveBoxplot.SetPrefix("jitter")
-            ucrSecondFactorReceiver.ChangeParameterName("colour")
-            clsCurrGeomFunc = clsJitterplotFunc
         Else
-            cmdBoxPlotOptions.Text = "Violin Options"
-            ucrSaveBoxplot.SetPrefix("violin")
-            ucrSecondFactorReceiver.ChangeParameterName("fill")
-            clsCurrGeomFunc = clsViolinplotFunc
+            'chkVarwidth.Checked = False
         End If
```
N-thony, line 428: Do you still need this?

instat/dlgBoxPlot.vb
```diff
@@ -368,69 +412,143 @@ Public Class dlgBoxplot
...
+        If clsCurrGeomFunction.GetParameter("varwidth") IsNot Nothing Then
+            If clsCurrGeomFunction.GetParameter("varwidth").strArgumentValue = "TRUE" Then
+                ' chkVarwidth.Checked = True
+                'Observe that changing the check of the chkVarwidth here doesn't trigger the checkchanged event.
+            End If
         Else
-            cmdBoxPlotOptions.Enabled = False
-            cmdOptions.Enabled = False
+            'chkVarwidth.Checked = False
         End If
```
N-thony, line 465: Same here too, do you need these if blocks since they don't have code inside

instat/dlgBoxPlot.vb
```diff
@@ -368,69 +412,143 @@ Public Class dlgBoxplot
...
+        bResetBoxLayerSubdialog = False
+        'Coming from the sdgLayerOptions, clsRgeom_boxplot and others has been modified. One then needs to display these modifications on the dlgBoxPlot.
+        If clsCurrGeomFunction.GetParameter("varwidth") IsNot Nothing Then
+            If clsCurrGeomFunction.GetParameter("varwidth").strArgumentValue = "TRUE" Then
+                ' chkVarwidth.Checked = True
+                'Observe that changing the check of the chkVarwidth here doesn't trigger the checkchanged event.
+            End If
+        Else
+            'chkVarwidth.Checked = False
+        End If
```
N-thony, line 502: here too

N-thony (changes requested):

EstherNjeriLiberatta:
@N-thony Thank you for reviewing. I have made the changes you requested.

N-thony (approved):
Great, @rdstern over to you

instat/dlgBoxPlot.vb
```diff
@@ -317,33 +318,74 @@ Public Class dlgBoxplot
...
+            clsCurrGeomFunction = clsJitterplotFunction
+        Else
+            ucrSaveBoxplot.SetPrefix("violin")
+            ucrSecondFactorReceiver.ChangeParameterName("fill")
+            clsCurrGeomFunction = clsViolinplotFunction
+        End If
+        'TODO Am not sure why the geomfunc parameter which carries clsCurrGeomFunction(current geom function) 
+        'does Not Update() properly when readio buttons are changed
+        'hence i have to force it to update properly after this if statement
+        clsBaseOperator.AddParameter(strFirstParameterName, clsRFunctionParameter:=clsCurrGeomFunction, iPosition:=2)
```
lloyddewit, line 347: The reason we need this line is because `clsCurrGeomFunction` is a reference. On lines 327, 333 etc. we change `clsCurrGeomFunction` to reference a different object. However the `clsBaseOperator` parameter still references the previous object.
Line 347 solves this by resetting the parameter to the new object referenced by `clsCurrGeomFunction`.
To make the code easier to understand, we could delete this block and set the `clsBaseOperator` parameter explicitly just after lines 327, 333, 338 and 342, for example
```
               clsCurrGeomFunction = clsTufteBoxplotFunction
               lsBaseOperator.AddParameter(strFirstParameterName, clsRFunctionParameter:=clsTufteBoxplotFunction, iPosition:=2)
``` 
What do you think?
(I hope this makes sense, if not, then let me know)

instat/dlgBoxPlot.vb
```diff
@@ -368,69 +410,139 @@ Public Class dlgBoxplot
...
+        'The aesthetics parameters on the main dialog are repopulated as required. 
+        For Each clsParam In clsRaesFunction.clsParameters
+            If clsParam.strArgumentName = "x" Then
+                If clsParam.strArgumentValue = Chr(34) & Chr(34) Then
+                    ucrByFactorsReceiver.Clear()
+                Else
+                    ucrByFactorsReceiver.Add(clsParam.strArgumentValue)
+                End If
+                'In the y case, the vlue stored in the clsReasFunction in the multiplevariables case is "value", however that one shouldn't be written in the multiple variables receiver (otherwise it would stack all variables and the stack ("value") itself!).
+                'Warning: what if someone used the name value for one of it's variables independently from the multiple variables method ? Here if the receiver is actually in single mode, the variable "value" will still be given back, which throws the problem back to the creation of "value" in the multiple receiver case.
```
lloyddewit, line 436: Also ensure comment does not exceed 100 columns.
```suggestion
                'In the y case, the value stored in the clsRaesFunction in the multiplevariables case is "value", however that one shouldn't be written in the multiple variables receiver (otherwise it would stack all variables and the stack ("value") itself!).
                'Warning: what if someone used the name value for one of it's variables independently from the multiple variables method ? Here if the receiver is actually in single mode, the variable "value" will still be given back, which throws the problem back to the creation of "value" in the multiple receiver case.

```

instat/dlgBoxPlot.vb
```diff
@@ -368,69 +410,139 @@ Public Class dlgBoxplot
...
+            ElseIf clsParam.strArgumentName = "y" AndAlso (clsParam.strArgumentValue <> "value" OrElse ucrVariablesAsFactorForBoxplot.bSingleVariable) Then
+                ucrVariablesAsFactorForBoxplot.Add(clsParam.strArgumentValue)
+            ElseIf clsParam.strArgumentName = "fill" Then
+                ucrSecondFactorReceiver.Add(clsParam.strArgumentValue)
+            End If
+        Next
+
+        'Question to be discussed: After running through the sdgLayerOptions, the clsCurrDataFrame parameters seem to have been cleared, such that in the multiple variable case, clsCurrDataFrame needs to be repopulated with "stack", "measure.vars" and "id.vars" parameters. Actually, even when repopulated, they are still not appearing in the script. ??
```
lloyddewit, line 444: Also ensure comment does not exceed 100 columns
```suggestion
        'TODO: After running through the sdgLayerOptions, the clsCurrDataFrame parameters seem to have been cleared, such that in the multiple variable case, clsCurrDataFrame needs to be repopulated with "stack", "measure.vars" and "id.vars" parameters. Actually, even when repopulated, they are still not appearing in the script. ??
```

instat/dlgBoxPlot.vb
```diff
@@ -368,69 +410,139 @@ Public Class dlgBoxplot
...
+        'This resets the factor receiver and causes it to be cleared of the correct variable. We don't want this.
+        'ucrVariablesAsFactorForBoxplot.SetReceiverStatus()
```
lloyddewit, line 446: delete?
```suggestion
```

instat/dlgBoxPlot.vb
```diff
@@ -317,33 +318,74 @@ Public Class dlgBoxplot
...
         sdgPlots.ShowDialog()
         bResetSubdialog = False
 
         'this syncs the coordflip in sdgplots and this main dlg
         ucrChkHorizontalBoxplot.SetRCode(clsBaseOperator, bReset)
-
     End Sub
 
-    Private Sub cmdBoxPlotOptions_Click(sender As Object, e As EventArgs) Handles cmdBoxPlotOptions.Click
+    Private Sub toolStripMenuItemBoxOptions_Click(sender As Object, e As EventArgs) Handles toolStripMenuItemBoxOptions.Click
```
lloyddewit, line 378: This function is almost the same as the 3 functions that follow.
Is it possible to combine into a single function?
Please see example in `dlgPICSARainfall.vb`.

image

instat/dlgBoxPlot.vb
```diff
@@ -368,69 +410,139 @@ Public Class dlgBoxplot
...
     End Sub
 
-    Private Sub ucrPnlPlots_ControlValueChanged() Handles ucrPnlPlots.ControlValueChanged, ucrChkTufte.ControlContentsChanged
-        SetGeomPrefixFillColourAes()
+    Private Sub DialogueSize()
+        If rdoBoxplotTufte.Checked Then
+            Me.Size = New Size(441, 505)
+            Me.ucrSaveBoxplot.Location = New Point(10, 384)
+            Me.ucrBase.Location = New Point(10, 413)
+        ElseIf rdoViolin.Checked Then
```
lloyddewit, line 525: would this work?
```suggestion
        If rdoBoxplotTufte.Checked OrElse rdoViolin.Checked Then
```

instat/dlgBoxPlot.vb
```diff
@@ -368,69 +410,139 @@ Public Class dlgBoxplot
...
-            clsBaseOperator.AddParameter(strStatSummaryParameterName, clsRFunctionParameter:=clsStatSummary, iPosition:=4)
-        Else
-            clsBaseOperator.RemoveParameterByName(strStatSummaryParameterName)
-        End If
+    Private Sub EnableDisableBoxplotOptions()
+        toolStripMenuItemJitterOptions.Enabled = rdoJitter.Checked OrElse ((rdoBoxplotTufte.Checked _
+            OrElse rdoViolin.Checked) AndAlso ucrChkAddPoints.Checked)
+        toolStripMenuItemViolinOptions.Enabled = rdoViolin.Checked
+        toolStripMenuItemBoxOptions.Enabled = rdoBoxplotTufte.Checked
+        toolStripMenuItemTufteOptions.Enabled = (rdoBoxplotTufte.Checked AndAlso ucrChkTufte.Checked)
```
lloyddewit, line 545:
```suggestion
        toolStripMenuItemTufteOptions.Enabled = rdoBoxplotTufte.Checked AndAlso ucrChkTufte.Checked

```

lloyddewit (changes requested):

rdstern (changes requested):
This is much closer than the other version.

@N-thony please check whether @EstherNjeriLiberatta has the time to do this quickly.  If not, then please re-allocate.

1) Here is the current version in 0.7.7
image
Notice that the new version has lost the Add Points row of controls.
That should be returned.  
And when it is ticked, then also enable Jitter as well as Box, because it is adding a layer
b) Add a Summary option into the pull-down.  When the Groups to Connect is ticked, then stat-summary is enabled.
c) And change Box to Boxplot.

PR #7579, opened 2022-06-28
Updated Mosaic dialogue's Plot Options to the new system.

### LINKED ISSUE #7556: Change the specific plots to the new system

EstherNjeriLiberatta:
Partially fixes #7556
In this PR, I have replaced the `Mosaic Options` button in the `Mosaic plot` dialogue with the new `Plot Options `button which includes a drop-down.


rdstern (commented):
There are the trivial changes mentioned in the boxplot pull request, that @N-thony has now seconded.  For consistency this is needed in each of the similar pull requests.
I looked at the mosaic plot itself and the Aesthetics are not there - all disabled.  I wonder if that is correct, or whether a small but can be sorted here.  This problem has been there a long time.  Otherwise, please there be an issue written on this, so we don't hold up the merging of this improvement.

PR #7580, opened 2022-06-28
Updated Parallel Coordinate dialog's Plot Options to new drop-down button system

### LINKED ISSUE #7556: Change the specific plots to the new system

EstherNjeriLiberatta:
Partially fixes #7556
In this PR, I have replaced the `Parcoord Options `button in the `Parallel Coordinate` plot dialogue with the new `Plot Options` button which includes a drop-down.


rdstern (commented):
Just the cosmetic changes.  Then can be accepted.

EstherNjeriLiberatta:
@rdstern This is ready for review.

N-thony (approved):

lloyddewit (approved):

lloyddewit:

rdstern (approved):

PR #7591, opened 2022-07-05
Prevented user-entered text from being translated (e.g. to French) in File Import and Rename Columns dialogs

### LINKED ISSUE #7590: Bug when trying to load csv file with the French translation
I'm able to load this data under the `English translation` but when I switch to `French translation`, I'm no longer able to load.
Here is the loading under English translation
image

this is what I get when I switch to French
image

this is the file [kisii_pyreth_data.csv](https://github.com/africanmathsinitiative/R-Instat/files/9045460/kisii_pyreth_data.csv)

### LINKED ISSUE #7615: A bug while Importing dataset using File>Import from File in French version
**Describe the bug**
When importing dataset using File>Import From File in French version, I get the following error;

image

But when I  use English version to import the same dataset , it works fine as shown below;

image

N-thony:
Fixes #7590 
Fixes #7615
@lloyddewit I made the following change in the database, I will add more if you agree with my suggestions in the corresponding issue.
image
image


lloyddewit:
@N-thony great! yes, please go ahead and make the other database changes (I added a comment to issue #7590 )

N-thony:
@lloyddewit this is an example with `input text control`, I'm trying to tidying the Gambia dataset but in the frecnh version, I noticed the `Nouveau nom input text control` is translating automatically.
image

lloyddewit:
> @lloyddewit this is an example with `input text control`, I'm trying to tidying the Gambia dataset but in the frecnh version, I noticed the `Nouveau nom input text control` is translating automatically. 

@N-thony Thank you for the discussion on 06 July and the example above. I agree that the contents of this text box should not be translated. As discussed, is it possible to fix this in a generalized way so that the contents of text boxes is never translated?
Thanks

N-thony:
> > @lloyddewit this is an example with `input text control`, I'm trying to tidying the Gambia dataset but in the french version, I noticed the `Nouveau nom input text control` is translating automatically.
> 
> @N-thony Thank you for the discussion on 06 July and the example above. I agree that the contents of this text box should not be translated. As discussed, is it possible to fix this in a generalized way so that the contents of text boxes is never translated? Thanks

We can simply set `DoNotTranslate` in the database to controls through their names  that are susceptible to create an enumeration parameter.

N-thony:
> > > @lloyddewit this is an example with `input text control`, I'm trying to tidying the Gambia dataset but in the french version, I noticed the `Nouveau nom input text control` is translating automatically.
> > 
> > 
> > @N-thony Thank you for the discussion on 06 July and the example above. I agree that the contents of this text box should not be translated. As discussed, is it possible to fix this in a generalized way so that the contents of text boxes is never translated? Thanks
> 
> We can simply set `DoNotTranslate` in the database to controls through their names that are susceptible to create an enumeration parameter.

@lloyddewit we can merge this for instance, I'm afraid now to make this change, on the second thought I think it will be good to decide what to do once we got the requested data for preparing the workshop from DRC Met. I think that will help on how to make the change because the translation is done also by switching the languages, if during the workshop it will require us to switch sometime @rdstern then we should be very carefull with some input text control.

lloyddewit:
I merged this without asking @rdstern for test/approval because it is a low-risk database change and I thought a quick merge was more important.

Issue #7565, opened 2022-06-17
Possibly add a primes key to the new integer keyboard?


rdstern:
This is only (at this stage) if it is really easy (less than 45 minutes?) for @lilyclements to code!  Then @anastasia-mbithe or @Wycklife could implement.   
It changes the current prime key to `is_prime` and adds a vectorised key called `prime` to the keyboard.  

a) We already have a `next prime` key, so (I presume) this could be done via purrr so the command prime(from = n) gives a set of prime numbers of length equal to the data frame.  Perhaps it needs a second argument such as values = m, where the default is the length of the data frame?  @lilyclements is this a really simple bit of code that could become the prime key?

The website on the [first 50 million primes](https://primes.utm.edu/lists/small/millions/) says "Usually it is faster to run a program on your own computer than to download them, but by popular demand, here they are!"

So that's one reason why this function would be useful.  It may also show how purrr avoids "old fashioned" loops?

[This site](https://stackoverflow.com/questions/19767408/prime-number-function-in-r) also gives various functions on primes.   I don't know if there is any help here?

lilyclements:
@rdstern I've had a look at this, mainly looking at stackoverflow.

So do we want two functions? One for `prime`, and one for `is_prime`, where `prime` gives a vector containing prime numbers, and `is_prime` checks if the current variable is a prime number.

Had a look on stack overflow, and to find `is_prime`, I think the following is a good solution:

```
is_prime <- function(x){
  div <- 2:ceiling(sqrt(x))
  !any(x %% div == 0)
}
```

The `prime` options found generally give a `to` parameter. Having this `m` option of to give the next `m` primes is a little tricky so the code uses the `nextprime` function. I give two options: give a `from` and `to`, or give a `from` and `n` for the number of primes you want after a given value.

```
list_prime <- function(from, to, n){
  is_prime <- Vectorize(is_prime)
  if (missing(n)){
    num <- from:to
    return(num[is_prime(num)])
  } else if (missing(to)){
    np <- NULL
    for (i in 1:n){
      np[i] <- as.integer(gmp::nextprime(from))
      from <- np[i]
    }
    return(np)
  }
}

list_prime(from = 1, to = 100)
list_prime(from = 8, n = 10)
```

anastasia-mbithe:
[...]

PR #7593, opened 2022-07-06
Added more keys to the Integer Keyboard

### LINKED ISSUE #7587: More possible keys on the integer keyboard - binary, octal and hexadecimal?
I notice that the lists provided, with the correction by @lilyclements now easily gives the first 1000 prime numbers, both as ordinary numbers and in binary.
That has made me wonder about including keys to transform from integer to binary - and also octal and hex, while we are there?  That should add another interesting teaching point for children at maths-camps?  

The R.utils package - already included, -  has the functions IntToBin, InttoOct and InttoHex functions.  Binary, Octal, Hex could perhaps be the key names?
These functions produce strings.  But base R also has as.octmode and as.hexmode.  So we could have the keys producing as.integer(R.utils::InttoBin(x)) 
as.octmode(R.utils::InttoOct(x))
as.hexmode(R.utils::InttoHex(x))

It is nice that this means we can do arithmetic on the hex and oct numbers.  (I don't know how to do this on binary in R?)    I suggest these could be fun???  @lilyclements and @anastasia-mbithe what do you think?

anastasia-mbithe:
Fixes #7587. 
I just added the three suggested keys to the Integer keyboard; octal, binary and hexadecimal.


lloyddewit (dismissed):

anastasia-mbithe:
@rdstern and @lloyddewit, in this PR, I have also added the split button for the R Help, and the Save checkbox also is checked by default when we have a dataset already loaded. These additional changes were mentioned in issue #7565.

lloyddewit (approved):

rdstern (approved):
Great

PR #7596, opened 2022-07-07
Ensured that date-type variables return `NA` if `sum` is performed in a calculation

lilyclements:
As @rdstern pointed out in #7465:

> I then tried with data-time variables. I used the mydata file from openair in the library. When you ask for sum there, (default with customised) it gives: image

This PR fixes this problem.

@rdstern @africanmathsinitiative/developers this is ready to review


lloyddewit (approved):

lloyddewit:

rdstern (approved):

PR #7604, opened 2022-07-08
Changes to Generate Date Sub Menus

### LINKED ISSUE #7588: Inconsistency between in Climatic > Dates > Generates Dates and the corresponding dialogue title
While discussing with @Wycklife , I found that the `Generates Dates...` sub menu title is different to its corresponding dialogue i.e `Regular Sequence`. @rdstern, could we make it consistent by changing the dialogue title from `Regular Sequence` to `Generates Dates`?
image
image

derekagorhom:
Fixes #7588 
This PR is ready for Review

Instead of Changing the `Regular Sequence` title dialogue to `Generate Dates` for the `Climatic` sub menu and `Prepare > Column: Date` sub menu
I changed the Sub menu names to Regular Sequence

@rdstern , @N-thony can you test and review this

### DIFF
--- a/instat/frmMain.Designer.vb
+++ b/instat/frmMain.Designer.vb
@@ -647,6 +647,7 @@ Partial Class frmMain
         Me.mnuOptionsByContextModel = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuOptionsByContextModelFitModel = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuOptionsByContextGeneralFitModel = New System.Windows.Forms.ToolStripMenuItem()
+        Me.mnuOptionsByContextCropModel = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuTools = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuToolsRunRCode = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuToolsRestartR = New System.Windows.Forms.ToolStripMenuItem()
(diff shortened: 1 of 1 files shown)


rdstern (commented):
@N-thony are you sure you want this change?
I was quite happy with the phrase `Generate Dates` as being clear before.  And I could see that perhaps (for consistency) you would wish for that phrase to also be the dialogue title.  It is now consistent, but "the other way round".  So it now says Regular Sequence in both places.  That's not so clear to me.  (I didn't really mind the inconsistency before, i.e. that the way to generate dates was via a regular sequence.)
I would prefer to keep the Generate Dates as the menu item, but don't have very strong views.  Perhaps someone else can decide.

N-thony:
> 

@rdstern yes, we can keep `Generate Dates` as it is more clear in both `sub menu` and `dialogues titles`
@derekagorhom what do you think?

derekagorhom:
@rdstern, @N-thony this 

> > 
> 
> @rdstern yes, we can keep `Generate Dates` as it is more clear in both `sub menu` and `dialogues titles` @derekagorhom what do you think?

Yes this can be done, but i have a small concern with if we change the dialogue title to `Generate Dates` does this not also affect the sub menu for `Regular Sequence`? 
and also the new column the Generate date produces should be changed to maybe `dates` instead of the usual `Regular`.

lloyddewit:
@derekagorhom The database is a binary file so I can't see the updates. Please could you update the PR description to explain exactly what you changed? For example, in his PRs, @N-thony normally provides a screenshot of the changed rows. I think this is the best approach.
If you can do this, and @rdstern approves, then we can merge.
Thanks

lloyddewit:
@derekagorhom I just merged PR #7591. Unfortunately this creates a conflict with the database file. This is unavoidable when two people change the same binary file in parallel. I'm afraid, that you will need to make the database changes again. I think you could partially resolve this with Git commands, but it's probably easier to create a new PR.
Sorry for the inconvenience.

derekagorhom:
hello @lloyddewit sorry for the late reply, i am ok with this PR being closed. i will create a new PR to fix #7588 later

Issue #6744, opened 2021-09-08
Improve parsing of multi-line R script


lloyddewit:
PR #6733 processes multiple lines of code in the 'Try' window. This PR works well for most cases. However, more sophisticated R parsing is available in the RScript library. I plan to create a NuGet package from this library in the coming months, and add it to R-Instat. I should then:

- [x] Review the changes in PR #6733 again, and see if the parsing can be improved with the RScript library
- [x] See if the Script window, or other dialogs, could be improved by using the RScript library

rdstern:
That sounds great.

lloyddewit:
I expanded this issue to include the proposal in PR #7610 :

RStudio has a good approach. They have an editor and a console. The user can type commands directly into the console (or paste/ctrl-enter commands from the editor). The console executes the lines sequentially. If the statement is incomplete then it changes the console prompt from '>' to '+'. If the user presses 'escape', then it clears the buffer and resets the prompt to '>'.

![image](https://user-images.githubusercontent.com/57253949/178672247-be48ce7d-a6a1-4d52-b606-3f266b7de808.png)

We could consider replacing the script window with a console window, and then using RScript to cover the whole R syntax. R-Instat's console window should then be nearly as good as the one in RStudio (we may omit some of its [advanced features](https://support.rstudio.com/hc/en-us/articles/200404846-Working-in-the-Console-in-the-RStudio-IDE)). I estimate that this would take around 40 hours including the changes to the RScript library (this is a very rough initial estimate).

rdstern:
@lloyddewit this is an exciting advance.  When you made the recent change (to include brackets in a multiline command) I also checked that this change also worked in the Edit > Script dialogue.  I am assuming the new console may make this dialogue obsolete, but I wonder if it may still have a role, because the scripts needed will often have a particular structure, that is made easy through this dialogue, namely:
a) Using particular packages
b) Getting data from an R-Instat data book
c) Doing something
d) Saving results back into the same data book, either adding to the data sheet, or maybe adding another data sheet into the data book
e) Tidying up - mainly deleting unneeded objects.

rdstern:
@lloyddewit Here is a script - well a function from the DescTools package documentation, function Fibonacci.  It gives an error.
```
# Fibonacci numbers without iteration
fibo <- function(n) {
phi <- (sqrt(5) + 1)/2
fib <- (phi^(n+1) - (1-phi)^(n+1)) / (2*phi - 1)
round(fib)
}
[...]

PR #7610, opened 2022-07-12
Ensured that R script with unclosed brackets treated as incomplete R statement

### LINKED ISSUE #7609: Improve the running of code from the script window
The following code does not run either from the script window, or using the new Edit > Script

I think that is probably simply due to the fact that an open bracket is not (yet?) counted as an indicator that the command line is continuing.  If all the as.factor commands are put on a single line, then the example does run. But if I put just one of those lines in the initial command, so the continuation is a comma, then it still doesn't run.   But it does if I include the final opening bracket in the line before.

```
library("sjPlot")
# -------------------------------
# random sample
# -------------------------------
# prepare data for 4-category likert scale, 5 items
likert_4 <- data.frame(
as.factor(sample(1:4, 500, replace = TRUE, prob = c(0.2, 0.3, 0.1, 0.4))),
as.factor(sample(1:4, 500, replace = TRUE, prob = c(0.5, 0.25, 0.15, 0.1))),
as.factor(sample(1:4, 500, replace = TRUE, prob = c(0.25, 0.1, 0.4, 0.25))),
as.factor(sample(1:4, 500, replace = TRUE, prob = c(0.1, 0.4, 0.4, 0.1))),
as.factor(sample(1:4, 500, replace = TRUE, prob = c(0.35, 0.25, 0.15, 0.25)))
)
# create labels
levels_4 <- c("Independent", "Slightly dependent",
"Dependent", "Severely dependent")
# create item labels
items <- c("Q1", "Q2", "Q3", "Q4", "Q5")
# plot stacked frequencies of 5 (ordered) item-scales

tab_stackfrq(likert_4, value.labels = levels_4, var.labels = items)
```
The example is from the sjPlot manual under the tab_stackfrq function.

lloyddewit:
Fixes #7609 

In the script window, R statements can be split across multiple lines for readability. When R-Instat parses the script, it parses line by line, and only executes the text when it judges that the R statement is complete.
It uses a set of heuristics to judge if the statement is complete (e.g. unclosed quotes, line ends in ',' etc.). This PR adds a check to ensure there that all open brackets have been closed. See issue #7609 for an example.

Note: The set of heuristics is quite simple but should cover most real-world examples. The RScript library parsing is much more sophisticated. However, it doesn't currently have a public function to check if the R script represents an incomplete R statement.
I suggest to stay with the simpler heuristics for now. If we find that we continue to expand the heuristics, then we should consider updating RScript and using this library instead (see issue #6744).

### DIFF
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -359,13 +359,18 @@ Public Class RLink
             '    then assume command is not complete
             Dim cLastChar As Char = strTrimmedLine.Last
             Dim strLast3Chars As String = ""
+            Dim iNumOpenRound As Integer = strScriptCmd.Where(Function(c) c = "("c).Count
+            Dim iNumClosedRound As Integer = strScriptCmd.Where(Function(c) c = ")"c).Count
             Dim iNumOpenCurlies As Integer = strScriptCmd.Where(Function(c) c = "{"c).Count
             Dim iNumClosedCurlies As Integer = strScriptCmd.Where(Function(c) c = "}"c).Count
             Dim iNumDoubleQuotes As Integer = strScriptCmd.Where(Function(c) c = """"c).Count
             If strTrimmedLine.Length >= 3 Then
                 strLast3Chars = strTrimmedLine.Substring(strTrimmedLine.Length - 3)
             End If
-            If cLastChar = "+" OrElse cLastChar = "," OrElse strLast3Chars = "%>%" OrElse iNumOpenCurlies <> iNumClosedCurlies OrElse iNumDoubleQuotes Mod 2 Then
+            If cLastChar = "+" OrElse cLastChar = "," OrElse strLast3Chars = "%>%" _
+                    OrElse iNumOpenRound <> iNumClosedRound _
+                    OrElse iNumOpenCurlies <> iNumClosedCurlies _
+                    OrElse iNumDoubleQuotes Mod 2 Then
                 Continue For
             End If
 


rdstern (approved):
Great.  This now works perfectly with that example.  I tried it both from the script window and also within the Edit > Script dialogue.  
Two questions:
a) Does it also work with square brackets [  ] and also curly brackets {   } which could also be used?
b) Now, if we open a bracket  and then don't close it (or a quote) then I assume we can type anything and it will just continue "eating" the code, until I close.  In interactive mode this could go on a long time.  If I suspect that something needs closing is there anything I can type that will close a command - and give an error - but enable me interactively to not go on forever?  I hope the question is clear?

lloyddewit:
@rdstern 

> Does it also work with square brackets [ ] and also curly brackets { } which could also be used?

It works with curly brackets but not square brackets. As I mentioned, currently there is a simple set of heuristics that aim to cover the most common cases. There are many aspects of R that is does not cover (e.g. single/double square brackets, single quotes, binary operators (apart from '+'), function declarations and other key words where the '{' is on the next line).

> Now, if we open a bracket and then ...

It will continue adding the code to a buffer ('eating' the code) until the heuristics consider the statement closed. The user cannot see what's in the buffer, so there is a risk that they don't know how to close the statement. Currently, the only way to clear the buffer is to exit R-Instat. 

Options:
- accept the current situation
- add a mechanism to allow the user to clear the buffer
- add extra heuristics for the most common situations
- implement a total solution.

I prefer the first or last option.

RStudio has a good approach. They have an editor and a console. The user can type commands directly into the console (or paste/ctrl-enter commands from the editor). The console executes the lines sequentially. If the statement is incomplete then it changes the console prompt from '>' to '+'. If the user presses 'escape', then it clears the buffer and resets the prompt to '>'.

image

We could consider replacing the script window with a console window, and then using RScript to cover the whole R syntax. R-Instat's console window should then be as good as the one in RStudio. I estimate that this would take around 40 hours including the changes to the RScript library (this is a very rough initial estimate).

What's your view?
thanks

rdstern:
@lloyddewit after discussion with @volloholic we are excited that you should go for your "total solution".  It fits so well with having a script button on each dialogue, with our "halfway" dialogues and so on.  Very exciting.  And just 40 hours too - I mean that's less than 2 days!

That suggestion made my day yesterday, and the teaching in India, I was doing, has only confirmed that there will be powerful people who would welcome this facility.  (The key person on that workshop, is a matlab user, who would quickly want to be using RStudio - i.e. commands, while having to start with R-Instat, because of its initial facilities in tidying and initial processing of the climatic data.)

This also fits my strategy, that we can never be the easiest R GUI, but we can be easy enough, and (by far) the most powerful!  And letting people start using R so easily, and then including the set of packages so they can proceed very quickly, is attractive in itself.

lloyddewit:
I merged this PR and will implement the R console against issue #6744

PR #7611, opened 2022-07-12
Improving the layout of the evapotranspiration dialogue

### LINKED ISSUE #7061: Improving the evapotranspiration dialogue
I have approved the small errors on wording in this dialogue.
@HawardKetoyoMsatsi I now suggest you start the work on the details of this dialogue, and this is quite a substantial task.  @N-thony should be the main tester.
1) I suggest you leave the layout issues until you have tested the dialogue as described below.  But the dialogue is too long and will not all be visible on some screens.  Also the options for missing values are scattered over the dialogue, etc.  So, a neater layout will be important to help in its use.
2) This is an important facility and the documentation will need to be extensive in its use.  It will be in the climatic guide and the help will also need to be extensive.  I hope we can use the examples of data from the evapotranspiration package as the main examples.  One example is 3-hour data, which is nice.  The second is used for PenmanMonteith in the guide, but I can't yet input it into R-Instat.

There is also a possible set of data - on a monthly basis - in the SPEI package.
And there is a set of data in the old Instat collection that could be useful.  It is called Bohicon and reads a single year of data (omit the first 13 lines)
[Old_Climatic.pdf](https://github.com/africanmathsinitiative/R-Instat/files/7677789/Old_Climatic.pdf)

I suggest to learn about evapotranspiration you could do worse than look in the relevant sections of the old Instat climatic guide.  It is Chapter 9 from section 9.5.

You may also wish to look in the [FAO guide] (https://www.fao.org/3/x0490e/x0490e04.htm#chapter%201%20%20%20introduction%20to%20evapotranspiration)

The idea is that you use the (complicated) Penman Monteith method even at sites where all the elements are not available - you estimate them then.  And this should all be clear from the dialogue.  As should the options for coping with missing values.

So in some of the examples you could miss a few days and then see what happens, etc.

Wycklife:
Fixes #7061

@africanmathsinitiative/developers Still in progress


rdstern:
@Wycklife what have you changed in the dialogue in the past 6 weeks, since my comments above.  It would be good to know.  That's partly for the MSc project, and also for R-Instat?

rdstern:
@MeSophie and @N-thony this is the pull request where I have given ideas for changes to the dialogue - see above.  Wycklife has left, so I wonder if you could take over this one.
@AmsaleEjigu you wanted to be kept up-to-date with this.

MeSophie:
> @Wycklife good you have started on this topic. I have been waiting for this. You should also be writing at the same time and having various examples of data that you are using in the chapter that is on evapotranspiration. I found I could download the data from ICRISAT (India) in 2 blocks and you may wish to add this station as an additional one. I think it probably has all the elements you need, and possibly with very few (or no) missing values. That will be useful. Look [here ](http://intranet.icrisat.org/nrmp/agroclimatology/weather.asp)for the data. So, do lots of writing concerning what you are doing in parallel with the changes in the dialogue and possibly attach that file - and the examples of the data if they are not currently in the library. a) I didn't see much change in the dialogue so far. I suggest you may want to have the existing dialogue as your starting point and then show your improved dialogue which is designed to make it easy for users to calculate the evapotranspiration. b) In your improvements one detail I notice is that the vertical distance to Tmax is not the same as to Tmin. That looks untidy. c) I notice it says Radiation and then below it says Solar. And one example of Solar is Sunshine Hours. Does that mean you should put the Sunshine hours under the Radiation entry if that is what you have. If that is right, then perhaps call the Receiver Solar rather than Radiation? Then in the pull-down, have Radiation as the 3rd entry, rather than Data? d) The 4th type of element is Wind speed and that is a bit separated from the others. If you improved the layout, then it would be clearer that what is needed for Penman/Monteith is temp, hum, solar, and wind. e) The whole dialogue looks confusing and cluttered. Consider having Missing Options as a button and putting all the controls concerning missing on a sub-dialogue. Would that help to make for a much clearer dialogue? Good if you could proceed quickly. It would be great if we could use the improved dialogue in our next climatic workshops.

@rdstern I make a PR#7958 according to your comment. And please I need some clarification

1) For a) is not really clear which are the improvements that need to be done  to make it easy for users to calculate the evapotranspiration? 

2) For c) If I understand you want that the dialog has two **Solar** one as the _receiver_ and another as a _selector_. I also notice that **Data** was already the 3rd position. It is okay like this or I need to modify it?
Thank you.

rdstern:
@MeSophie I thought I had sort of answered those questions, but here goes:
a) The original dialogue was very long, and our general idea is that the main dialogue is as simple as possible, with possible sub-dialogues when needed.  And sometimes the sub-dialogue may be complicated.  This was not the case with the original dialogue, that seemed very cluttered.  It is simpler now you have put the missing options on a sub-dialogue.  
b) I still think the way Radiation is specified remains a bit confusing, but if you are happy with that, then keep it in its current form.  I thought changing the label Radiation: into Solar, and then the word Data in the Solar pull-down into Radiation, might make the options clearer.  But happy if you think the current layout is clear.
Then I am convinced you may find you could make further improvements after you have tried using the dialogue for a range of the different options that it is supposed to handle.  So I would then like you to try using it, and see how it is standing up.

rdstern:
@MeSophie I also note you have a newer pull request on the same dialogue.  Can we close this one now?

MeSophie:
> @MeSophie I also note you have a newer pull request on the same dialogue. Can we close this one now?

Yes @rdstern we can close it.

PR #7612, opened 2022-07-12
Ensured that Y-axis settings are persistent when closing/re-opening PICSA Rainfall dialog

### LINKED ISSUE #7607: Fix remaining minor bug on PICSA graph dialogues
The  3 PICSA dialogues are working well now.  This is a great relief and they are also impressive.  
A bug that remains concerns the resetting of a control when the dialogue is used repeatedly.  I suspect there are more "features" like this, in these  - and in similar dialogues.
The simplest example I have found uses the dodoma data from the library - the rds file.
Get the annual summary of the total rainfall - and perhaps also the maximum rainfall (Climatic > Prepare > Summary).
Now use the Climatic > PICSA > Rainfall Graph dialogue.
Graph total rainfall against the year.  Use the PICSA options and set the low limit of the Y-axis to zero.  (Possible use lines and add the mean line - I am not sure if this is needed.)
Press Ok and it works fine.
Now return to the dialogue and press Ok again.  The graph is again Ok but the Y-axis setting at zero has now unset itself.  So it now starts at about 400mm for the Y axis!

Vitalis95:
Fixes #7607 
@rdstern @N-thony @lloyddewit  this PR fixes the issue when you set Y-axis  at any point  then it doesn't unset itself when dialogue is used repeatedly

### DIFF
--- a/instat/dlgPICSARainfall.vb
+++ b/instat/dlgPICSARainfall.vb
@@ -80,6 +80,8 @@ Public Class dlgPICSARainfall
     Private strLowerTercileName As String = ".lower_ter_y"
     Private strUpperTercileName As String = ".upper_ter_y"
 
+    Private clsDummyFunction As New RFunction
+
     Private clsAsDateYLimit As New RFunction
     Private clsGeomHlineMean As New RFunction
     Private clsGeomHlineMedian As New RFunction
@@ -220,6 +222,7 @@ Public Class dlgPICSARainfall
         clsBaseOperator = New ROperator
         clsGroupByFunction = New RFunction
         clsMutateFunction = New RFunction
+        clsDummyFunction = New RFunction
 
         clsRggplotFunction = New RFunction
         clsGeomLine = New RFunction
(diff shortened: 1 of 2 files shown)


rdstern (changes requested):
@Vitalis95 this is much better and the control now seems to behave itself when the y-axis is numeric.  But it still goes wrong when using dates.  I will send you the data for the plot below through skype.

It would be excellent if you can fix this too, at the same time?  

To be clear what I did.  I first did plots where y is the sum_rain.  I get the mean line and also set the y-axis to start at zero.  All seems ok now.

Now I change and plot the start_rain.  Now the y-axis is a date, and the default is for the y-axis to start at day 1.  I untick this.  
The graph works fine.
Then I return to the dialogue and the PICSA plot sub-dialogue.  Then I go to the Lines tab.  I don't change anything, but do the plot again.  Now the graphs start on Jan 1.  When I return the the PICSA sub-dialogue I find the y-axis control is ticked and at the default of 1.  I hope you can prevent that?

Vitalis95:
@rdstern , can you have a look at it now

rdstern (approved):
@Vitalis95 excellent.  Looks fine now.  Many thanks.  Hope @lloyddewit likes it so it can be merged.

lloyddewit (approved):

PR #7623, opened 2022-07-15
Improved the Describe > One Variable >Rating Data dialog

### LINKED ISSUE #7608: Further improve the Describe > One Variable >Rating data dialogue
This dialogue is now better than previously, following but I suggest there is scope for further improvements following #6786.  @N-thony to allocate.

Two examples to use are as follows.
a) the efc data for the 9 variables on caring.  This is the library data for the sjplot package.  Each rating has 4 options, so there is no neutral option.
b) Use File > New Data Frame > Random and then the 2nd to last example (survey2) which gives 10 examples of a 5-point rating.

a) Start with the R-code for the existing dialogue.  As requested earlier (but not implemented) the R-code should separate getting the data, in each of the 3 cases (functions from sjplot package) and applying the respective function.  (We could merge after this one if the following shows problems.
b) Be able to label the results with the variable names if there are no variable labels.  The example given in the documentation gives:

image

I suggest this could be a checkbox with label - Show Variable Names.  Default is unchecked.  The current situation is that it then shows the variable labels.  If there are no variable labels, then it shows nothing.  But there is an option var.labels at least for the table to implement.  I assume this may also apply for the other options, but if not, then ignore that.   

c) Consider whether there are other options for the 3 functions that can usefully be added to each dialogue.  Propose any you find in this issue before implementing.  There may not be any.

d) I note that it works for grouped data?  So try first and then we may wish to add the dialogue - with one or two grouping variables - to the 2/3 way dialogues.

anastasia-mbithe:
Fixes #7608 

This is still in progress, I have implemented parts (a) and (b).

### DIFF
--- a/instat/dlgRatingScales.Designer.vb
+++ b/instat/dlgRatingScales.Designer.vb
@@ -66,6 +66,7 @@ Partial Class dlgRatingScales
         Me.ucrReceiverOrderedFactors = New instat.ucrReceiverMultiple()
         Me.ucrBase = New instat.ucrButtons()
         Me.ucrSelectorRatingScale = New instat.ucrSelectorByDataFrameAddRemove()
+        Me.ucrChkVariableNames = New instat.ucrCheck()
         Me.grpSort.SuspendLayout()
         Me.grpSortTable.SuspendLayout()
         Me.SuspendLayout()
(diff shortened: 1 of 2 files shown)


N-thony:
> This fixes #7608
> 
> This is still in progress, I have implemented parts (a) and (b).

@anastasia-mbithe always start with `Fixes` not `This fixes` so that the corresponding issue will close automatically if this PR is merged.

rdstern:
@anastasia-mbithe it is so good that you have done item a).  I was surprised that you managed b) so quickly - and I see ytou haven't.
image

I show with the data from File >New Data Frame as mentioned above.  This has names - as you see, but no variable labels.  So you see there are still no names given to each variable in the output.

That is item b).  If it is too difficult, then an easy "work around" is always to copy the names into the labels - so don't spend too long on it.

Your work on implementing a) has had an immediate benefit for me.  In d) I had hoped that the dialogue could easily work with groups.  But I now see that this is not the case.  So ignore d).  And if b) is diffiult, then we'll merge with just the changes you have made.

anastasia-mbithe:
image

@rdstern, Is this what you meant for (b) when the checkbox is checked?

anastasia-mbithe:
This gives the variable labels and fixes some minor issues.
@rdstern kindly test it

image

anastasia-mbithe:
I have fixed that. Kindly test. 
Thanks

rdstern (approved):
@anastasia-mbithe that's great - @lloyddewit if you are happy, then it would be nice to have it merged.
image

@anastasia-mbithe  You can see the vertical alignment of the labels is not ideal.  So, not urgent, but I still wonder about another issue to see whether a few more options can be added?

lloyddewit (approved):
@anastasia-mbithe Looks great, thank you

PR #7629, opened 2022-07-16
Added filter for empty rows in Define New Filter dialog

### LINKED ISSUE #6406: Two small (?) changes to the filter dialogue
The really trivial change is to change the default name from Filter to filter - so no capital letters.

The (only do it is easy) is the change the Edit Filter to `Add to Filter`.  It would be possible that the sub-dialogue doesn't change at all.  It is just when you press Return, then it executes (original filter) & (new filter).

That's all I usually need.  For example if you filter to factor levels of station and one of them is Arkona.  Then I may want  Arkona & (year >= 1983)!

Of course the Add to Filter (replacing Edit Filter) is only enabled when a filter has been selected.

### LINKED ISSUE #7617: Very small change in New-Filter sub-dialogue
The problem is shown here

image

Here are the data:
[test_of_blank.csv](https://github.com/africanmathsinitiative/R-Instat/files/9113102/test_of_blank.csv)

It imports with variable b as a factor, so right-click and make it a character variable.  Then try and make the filter.

The correction is simple (I think) It is just that the Add Condition button is disabled if the field is empty.  All we need to do is enable it.

@N-thony to allocate please

shadrackkibet:
fixes #7617

@rdstern I added an option "isEmpty" into the dropdown. When you select this option it allows you to click Add Condition then you can select only empty rows. What do you think?
image


rdstern:
That's an interesting way to do it - and seems to work fine.  I thought we would just enable the `As Condition` button with an empty string, but I am also happy with this solution.  
a) I assume we can name it as we want, so would `is.empty`, rather than `isEmpty` be more consistent with the `is.na`?
b) We should then also add `!is.empty` for the opposite - as we have for `!is.na`.

shadrackkibet:
@rdstern I agree to the request above and I have actioned it accordingly.

rdstern (dismissed):
@shadrackkibet this is great.  Not the way I was thinking, but this way seems better.  And excellent that you are back to R-Instat, even though no longer with INNODEMS.  I hope @lloyddewit can approve, so you can proceed to the useful ADD facility after this is merged!

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -1788,6 +1788,13 @@ DataSheet$set("public", "get_filter_as_logical", function(filter_name) {
           } else {
             result[, i] <- !col_is_na
           }
+         }else if (condition[["operation"]] == "is.empty"||condition[["operation"]] == "! is.empty"){
+           col_is_empty <- self$get_columns_from_data(condition[["column"]], use_current_filter = FALSE) == ""
+          if (condition[["operation"]] == "is.empty") {
+            result[, i] <- col_is_empty
+          } else {
+            result[, i] <- !col_is_empty
+          }
```
lloyddewit, line 1797:
```suggestion
        } else if (condition[["operation"]] == "is.empty" || condition[["operation"]] == "! is.empty"){
          col_is_empty <- self$get_columns_from_data(condition[["column"]], use_current_filter = FALSE) == ""
          if (condition[["operation"]] == "is.empty") {
            result[, i] <- col_is_empty
          } else {
            result[, i] <- !col_is_empty
          }
```

lloyddewit (changes requested):

lloyddewit (approved):
@shadrackkibet Tidy fix, thanks, happy to approve.
I just fixed a small indentation issue.

PR #7636, opened 2022-07-20
Set the ucrInput control to DoNotTranslate to prevent bugs when switching in French

N-thony:
@lloyddewit  @lloyddewit I have made the following change as discussed earlier today in the database to avoid the bug when switching in French
 I used the following SQL code for this

image

this is the change

image


lloyddewit:
@N-thony thank you for this change, the SQL command to automate and your clear explanation. 
There are many dialogs changed and it won't be practical for @rdstern to test regression on all of them. Please could you suggest 3 or 4 dialogs for @rdstern to test in both French and English?
Thanks

N-thony:
> @N-thony thank you for this change, the SQL command to automate and your clear explanation. There are many dialogs changed and it won't be practical for @rdstern to test regression on all of them. Please could you suggest 3 or 4 dialogs for @rdstern (or his delegate) to test in both French and English? Thanks

@rdstern you can test Rename Column, Make Date, PICSA, and Regular Sequence dialogues

rdstern:
a) So I am checking Rename Columns and it seems to work fine.  There is a lot that is presumably not in the database, so is not translated.  In particular the rename options on the rename with radio button are all still in English.  (I assume that's different to this pull request.)
b) When I choose regular sequence I notice the whole Prepare > Column: Numeric > menu is still in English?
c) In regular sequence I was please to see the message about the length was in French when the sequence was the correct length.  But when the sequence was truncated, then the corresponding message was in English.  However - it all works!!
d) In Make date I found my first example of where the control was not wide enough in French - see below:
image
And another bit remains in English.
e) In PICSA the menu remains mainly in English.  
The rainfall dialogue looks pretty good.  The lines option needs the initial control to be wider, for the text to be shown completely.
image


However, after all this, I assume the main purpose of the changes here was to ensure it works.  So I am approving.

rdstern (approved):
I couldn't find anything that didn't work.  So I am approving.  

There is a lot that remains to be translated and a few instances where the controls need to be resized.  I assume this could happen over the next 2 weeks, so we will have an improved version for DRC, even if only for the better translations.

lloyddewit (approved):

PR #7638, opened 2022-07-20
Added prime keys to the Calculator dialog

### LINKED ISSUE #7565: Possibly add a primes key to the new integer keyboard?

anastasia-mbithe:
Fixes #7565 

@rdstern, I have added the nth_prime key and the prime key that generates primes with n= nrows(dataframe).
Kindly have a look at it. You can suggest other keys to add from the primes package.

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -188,7 +188,7 @@ Public Class dlgCalculator
                 Me.Width = iBasicWidth * 1.27
                 ucrBase.iHelpTopicID = 598
             Case "Integer"
-                Me.Width = iBasicWidth * 1.18
+                Me.Width = iBasicWidth * 1.37
             Case Else
                 Me.Width = iBasicWidth
         End Select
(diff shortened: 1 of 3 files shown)


lloyddewit:
@Ivanluv Please could you peer review?

Thanks!

Ivanluv (approved):

rdstern (approved):
@anastasia-mbithe looks fine.

lloyddewit (approved):

PR #7642, opened 2022-07-22
Allow a maximum of one factor as the base for the percentages in the Prepare>Data Reshape >Column Summaries Proportions and Percentages sub dialogue

Ivanluv:
This PR replaces the Multiple receiver, for Percentages, with a single receiver as suggested by @rdstern on PR #7416 

> Check with @lilyclements first, but I am suggesting (even in the general case) that the dialogue only allow a maximum of one factor as the base for the percentages. If she agrees, then the 3-variable could become a single receiver (as it is in the 2-variable case)

### DIFF
--- a/instat/sdgProportionsPercentages.Designer.vb
+++ b/instat/sdgProportionsPercentages.Designer.vb
@@ -33,10 +33,10 @@ Partial Class sdgProportionsPercentages
         Me.ucrReceiverFilter = New instat.ucrReceiverSingle()
         Me.ucrReceiverColumn = New instat.ucrReceiverSingle()
         Me.ucrButtonsProportionsPercentiles = New instat.ucrButtonsSubdialogue()
-        Me.ucrReceiverByFactor = New instat.ucrReceiverMultiple()
         Me.ucrSelectorProportionsPercentiles = New instat.ucrSelectorByDataFrameAddRemove()
         Me.ucrPnlBY = New instat.UcrPanel()
         Me.ucrChkProportionsPercentages = New instat.ucrCheck()
+        Me.ucrReceiverByFactor = New instat.ucrReceiverSingle()
         Me.grpOptions.SuspendLayout()
         Me.SuspendLayout()
         '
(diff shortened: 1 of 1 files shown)


lloyddewit:
@N-thony please could you review?
Thanks!

lloyddewit (approved):

lloyddewit:
@rdstern If you can test/approve, then we can merge, thanks

rdstern (commented):
@volloholic that's interesting and I have never used this option - perhaps wrongly.  So I need you to check what's to happen here.  For example, there is just one receiver - and it is optional.  So how is it used?  And, after quite some years, the other two buttons are still disabled.

Please confirm.  I am not aware that I asked for this change.

lloyddewit:
> @volloholic that's interesting and I have never used this option - perhaps wrongly. So I need you to check what's to happen here. For example, there is just one receiver - and it is optional. So how is it used? And, after quite some years, the other two buttons are still disabled.
> 
> Please confirm. I am not aware that I asked for this change.

@rdstern is your comment intended for @volloholic or @Ivanluv ?

Ivanluv:
> @volloholic that's interesting and I have never used this option - perhaps wrongly. So I need you to check what's to happen here. For example, there is just one receiver - and it is optional. So how is it used? And, after quite some years, the other two buttons are still disabled.
> 
> Please confirm. I am not aware that I asked for this change.

@rdstern  the receiver  adds percentages to the summary function used by the Column Summaries dialogue. Therefore 
 I changed the multiple receiver in the sub dialogue to a single receiver as per your suggestion of generally having percentages as single variables

rdstern:
@Ivanluv I suggest we leave it as a multiple receiver in this dialogue.  Sorry for the delay in thinking about this.  I suggest it is rare that we will need a multiple receiver, so we don't need that complication on the main dialogue for the main tabulation or on the 2-way tables.  But it could be needed, and hence allowing it on this sub-dialogue does no harm.

Ivanluv:
> @Ivanluv I suggest we leave it as a multiple receiver in this dialogue. Sorry for the delay in thinking about this. I suggest it is rare that we will need a multiple receiver, so we don't need that complication on the main dialogue for the main tabulation or on the 2-way tables. But it could be needed, and hence allowing it on this sub-dialogue does no harm.

@rdstern  thank you for the clarification. I will then close this PR

PR #7654, opened 2022-07-25
Improving PCA dialog

### LINKED ISSUE #6686: PCA Options subdialog, Save Tab Empty
Hi All, 

I have noticed that the PCA Options subdialog, Save Tab Empty. Is it waiting to be filled or should it be removed. 

Thanks, 

Rachel 

PCA Options

### LINKED ISSUE #7318: Extend the PCA dialogue
Looking in more detail this should be a reasonably easy improvement.
We currently already use the factominer package.  
Here is the current dialogue:
image

The factominer package says:
```
Description
Performs Principal Component Analysis (PCA) with supplementary individuals, supplementary
quantitative variables and supplementary categorical variables.
Missing values are replaced by the column mean.
Usage
PCA(X, scale.unit = TRUE, ncp = 5, ind.sup = NULL,quanti.sup = NULL, quali.sup = NULL, row.w = NULL,col.w = NULL, graph = TRUE,axes = c(1,2))
```
These are exciting additions, particularly:
a) Multiple receiver for additional numeric variables.  Suppl. Numeric
b) Multiple receiver for additional factor variables.  Suppl. Factors
c) Receiver for Additional individuals - a filter for us! 
In the examples it seems to code them numerically.  I assume we can use variable names?  I am not sure about c)
Then it is important that we get a pca object.  
There is a video, that then shows many alternative graphs from the pca object.
There is plot, predict, print and summary for the object.
There is a video, and it looks very nice.  How can we do that sort of investigation in R-Instat?

There is also fviz_pca and fviz_pca_var and get_pca from the factoextra package.

It looks as though (at least for multivariate, we want to get a pca object and then use it.  This is the parallel of our modelling ideas, i.e. fit and use, where we also have keyboards.  
We might even want to consider keyboards in multivariate, just as we have in the modelling???
We can possibly investigate by using the current modelling keyboard dialogues, before seeing if that could supply an "advanced" option for factoMiner to start with.  The fit keyboard could then not just be for pca, but for each of the methods?
 With dialogues we can start with Principal Components

Vitalis95:
Fixes #7318 
Fixes #6686 
@rdstern @N-thony @lloyddewit , this PR improves PCA dialog

The variables that are chosen for the `quanti.sup` or `quali.sup` receiver are limited to the explanatory variables in the multiple receiver for PCA,  you need to include here all variables which are  `quanti.sup` and `quali.sup`


lloyddewit:
@N-thony please could you peer review?
Thanks

rdstern (changes requested):
@Vitalis95 this is progressing well.  
a) But there is a problem with your rule that the variables in the supplementary need to be included in the initial variables.  It starts correctly with the main control having just the numeric variables.  So how are you supposed to be able to add supplementary factors?
b) There also seems to be a problem, with what is remembered from a previous use of the dialogue.  I made changes, and it still seemed to give me errors that were from the previous use of the dialogue.
c) So I pressed Reset.  Now it still kept the focus on a control that was no longer visible.  So it isn't quite resetting properly.
d) The output options need to change.  Currently you output the components themselves.  That was sort of ok with the survey data, but not at all with the diamonds.

Vitalis95:
@rdstern , the extra variables takes number of variables rather than variables names as it is always in our selector

Vitalis95:
@rdstern @lilyclements , could you have a look it
`Extra Variables` control adds the supplementary variables

lilyclements (commented):
Review for now - can you also sync the stand_alone_functions.R file that contains the `cbind_unique` function. Otherwise we cannot test it.
I suggest you update your branch before doing this to avoid conflicts.

Vitalis95:
closing the PR due to conflicts

PR #7658, opened 2022-07-26
Picsa rainfall graph

Fidel365:
Fixes (partially) #7556 
Modifying the plot option control in PICSA Rainfall graph
@N-thony , this PR is ready for review

### DIFF
--- a/instat/dlgBoxPlot.designer.vb
+++ b/instat/dlgBoxPlot.designer.vb
@@ -38,10 +38,9 @@ Partial Class dlgBoxplot
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()>
     Private Sub InitializeComponent()
-        Me.cmdOptions = New System.Windows.Forms.Button()
+        Me.components = New System.ComponentModel.Container()
         Me.lblByFactors = New System.Windows.Forms.Label()
         Me.lblBySecondFactor = New System.Windows.Forms.Label()
-        Me.cmdBoxPlotOptions = New System.Windows.Forms.Button()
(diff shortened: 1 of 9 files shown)


Fidel365:
@N-thony , the recent commit fixes minor overlapping issue in the PICSA rainfall graph

N-thony (commented):
@Fidel365 the title of this dialogue says Picsa rainfall graph but in the list of files changed I can see you made the change also in dlgBoxPlot, dlgMosaicPlot, is that correct?

Fidel365:
@N-thony , the commits of Mosaicplot and Boxplot should not be here, the changes are supposed to be on the PICSA rainfall graph

rdstern:
@N-thony and @Fidel365 this currently does not build.
I wonder what the main change is here.  I ask because I don't think the new drop-down system applies particularly for this dialogue.
In the other plotting dialogues, the Plot options is the main one being used, with the specific options used more rarely.  In contrast, in this dialogue, I use the PICSA options almost all the time and the general plot options only rarely.  So here I am happy to keep the 2 buttons separate.  the same applies to the new PICSA temperature graph and the general graph.

At this stage they don't need to change.

N-thony:
> @N-thony and @Fidel365 this currently does not build. I wonder what the main change is here. I ask because I don't think the new drop-down system applies particularly for this dialogue. In the other plotting dialogues, the Plot options is the main one being used, with the specific options used more rarely. In contrast, in this dialogue, I use the PICSA options almost all the time and the general plot options only rarely. So here I am happy to keep the 2 buttons separate. the same applies to the new PICSA temperature graph and the general graph.
> 
> At this stage they don't need to change.

Okay, so we don't need this change, we can close this PR.

PR #7682, opened 2022-07-29
Fixed the graphics axes issue in the General Graphics dialog.

### LINKED ISSUE #5349: Annoying small point about the graphics axes
I am pleased with the steady improvements in the "graphics system" in R-Instat.  Here is a small point that I have marked as a bug, but it is more an inconvenience.

Go to any Describe > Specific > Boxplot (say) but any example will do.  Then use Plot Options and either X-axis or Y-axis.  It always starts with Discrete as the default, which is inconvenient as we usually need Continuous.  If you change to Continuous it works fine and settings can be done.

But, when you return to the main dialogue and then return to that tab again on the sub-dialogue, it is back to Discrete again.  It just loves Discrete.

The settings under continuous are remembered, but I suggest that Continuous should be the default and it should also return to the setting that was used on a future occasion.

I assume (hope) that this is easy to fix, and I am not opening a "can of worms" here.  If so, then change the milestone.

anastasia-mbithe:
Fixes #5349 

@rdstern, the issue here was in the general graphics dialog and not the Plot options sub dialog. So, I traced the error and fixed it. Kindly test.
The condition is that when the receivers are empty, the X/Y axes are discrete, but when the receivers are filled, it changes to Continuous and maintains their state.


lloyddewit (approved):

lloyddewit:

rdstern (approved):
Looks good

PR #7718, opened 2022-08-03
Corrected misleading message in Tools > Install R package dialogue 

### LINKED ISSUE #7707: Correct misleading message in Tools > Install R package dialogue
Here is an example:

image

If you give an incorrect name, so there is no package in CRAN, then it says so.  

This seems to be an acceptable package name, but it is not (currently) in CRAN.  I get the same message with the mmtable2 package?

The splines package has been archived.  Perhaps a better message could be

`Not a current CRAN package.  Perhaps it has been archived.`

Ivanluv:
fixes #7707 


instat/dlgRPackages.vb
```diff
@@ -100,7 +100,7 @@ Public Class dlgInstallRPackage
                 ucrInputMessage.SetText("Cannot get package information.")
             End If
         Else
-            ucrInputMessage.SetText("Cannot get package information. Check your internet connection.")
+            ucrInputMessage.SetText("Not a current CRAN package. Perhaps it has been archived")
```
lloyddewit, line 103: How do we know that this is the reason that the `package_check` command failed? Could it also be an Internet connection error or another reason? Do we need to do some extra checking?

lloyddewit (changes requested):

instat/dlgRPackages.vb
Ivanluv: I have added an internet connection error check and slightly modified the checking function

instat/dlgRPackages.vb
```diff
@@ -95,12 +89,21 @@ Public Class dlgInstallRPackage
...
+                    ucrInputMessage.SetText("Not a current CRAN package. Perhaps it has been archived")
+                    ucrInputMessage.txtInput.BackColor = Color.LightSkyBlue
+                ElseIf chrOutput(0) = "5" Then
+                    ucrInputMessage.SetText("No internet connection.Try reconnecting")
+                    ucrInputMessage.txtInput.BackColor = Color.LightCoral
                 End If
-            Else
-                ucrInputMessage.SetText("Cannot get package information.")
             End If
         Else
```
lloyddewit, line 106: delete this `Else`?

instat/dlgRPackages.vb
```diff
@@ -78,13 +78,7 @@ Public Class dlgInstallRPackage
         If expOutput IsNot Nothing AndAlso Not expOutput.Type = Internals.SymbolicExpressionType.Null Then
```
lloyddewit, line 78: There are some conditions below where `ucrInputMessage` is not set to anything.
We should also follow the 'happy path stays left' principle.
Suggested code below, could you think of a better message?
```suggestion
        If expOutput Is Nothing OrElse expOutput.Type = Internals.SymbolicExpressionType.Null Then
            ucrInputMessage.SetText("Cannot get package information.")
            Exit Sub
        End If
```

instat/dlgRPackages.vb
```diff
@@ -78,13 +78,7 @@ Public Class dlgInstallRPackage
...
             chrOutput = expOutput.AsCharacter
             If chrOutput.Count >= 1 Then
```
lloyddewit, line 80:
```suggestion
        If chrOutput.Count < 1 Then
            ucrInputMessage.SetText("<add message here>")
            Exit Sub
        End If                
```

instat/dlgRPackages.vb
```diff
@@ -78,13 +78,7 @@ Public Class dlgInstallRPackage
...
-                If chrOutput(0) = "0" Then
-                    ucrInputMessage.SetText("No package with this name.")
-                    ucrInputMessage.txtInput.BackColor = Color.LightCoral
-                ElseIf chrOutput(0) = "2" Then
-                    ucrInputMessage.SetText("Package exists and not currently installed.")
-                    ucrInputMessage.txtInput.BackColor = Color.LightGreen
-                ElseIf chrOutput(0) = "1" Then
+                If chrOutput(0) = "1" Then
```
lloyddewit, line 81: Replace this `If` and `ElseIf` construction with a `Select Case` structure?

lloyddewit (changes requested):
@Ivanluv Thanks for adding the extra messages, I think this is a big improvement.
I added some extra comments

instat/dlgRPackages.vb
Ivanluv: Done

instat/dlgRPackages.vb
Ivanluv: Done

instat/dlgRPackages.vb
Ivanluv: Done

instat/dlgRPackages.vb
Ivanluv: Done

lloyddewit:
@Ivanluv I don't see the changes, did you push?

Ivanluv:
> @Ivanluv I don't see the changes, did you push?

@lloyddewit could you check again

lloyddewit (approved):

lloyddewit:

rdstern:
@lloyddewit It now seems to say the following whatever package I use.
image

So it is now unusable - what's going on here?

rdstern:
@N-thony I am not sure whether you could take this over, or re-allocate it.

N-thony:
@rdstern I have tested this and it works for me. What you need before testing this is installing `pingr` package Ivan used here to satisfy the connecting to the internet.
This package should be added in the list of default packages of R-Instat.

lloyddewit:
> What you need before testing this is installing pingr package Ivan used here to satisfy the connectivity check to the internet. This package should be added in the list of default packages of R-Instat.

@N-thony Please could you add the `pingr` package to R-Instat. An example of doing this is in PR #7557. If you add the `coin` package at the same time, then you can close PR #7719.
Thanks

rdstern (approved):
I installed the pingr package, and it now works fine.

PR #7764, opened 2022-08-12
Added Russian language option (beta)

lloyddewit:
Added Russian language option and merged Russian translations in CrowdIn PR #7756 into database.
Please note that:

- These are currently machine translations and have not been reviewed
- The menus and dialogs have not been tested in Russian, there may be positional and other problems

@rdstern Normally, I would not want to merge beta functionality into master. However, if we don't merge this PR very soon, then it will cause a merge conflict the next time we want to update the database. Also, this PR only affects users who specifically change the language option to Russian.
If you agree, then please could you do a sanity check and approve?
Thanks!


rdstern (approved):
Fantastic.
There is also an extra positive that the translations are so different, it makes it very clear, what still remains to be translated - not just into Russian, but into each language.
On the negative side I have an impression that sometimes Russian words may be even longer than French ones - so we will need to check also on the space for labels within the dialogues.

PR #7780, opened 2022-08-16
Modified Plot Control in  the Heat map dialog: Desribe>Graphs>Heatmap

### LINKED ISSUE #7556: Change the specific plots to the new system

EstherNjeriLiberatta:
Fixes #7556 
In this PR, I have updated the heatmap dialogue `plot options` control. 


lloyddewit:
@Ivanluv Please could you peer review?
Thanks!

Ivanluv (approved):

rdstern (approved):

lloyddewit (approved):

PR #7781, opened 2022-08-16
Improved behavior of position control in the Prepare > Data Frame > Row Numbers/Names dialog

### LINKED ISSUE #7772: Improve behavior of position control in (at least) the Prepare > Data Frame > Row Numbers/Names dialogue
When the row variable is added, then the default position is the Start, i.e. the first column.
So, try this example:
File > New Data Frame > Lists > Maths > Primes
Then
File > New Data Frame > Lists > Maths > Binary Primes 
Now use the Row Numbers/Names dialogue on each.
On the primes it is fine and the row variable is first.
But when you move to the Binary Primes it has changed to become last!

I don't know if this change is just a problem with this dialogue, or is also an issue with other dialogues?  So, check that possibility as well, please.
Perhaps @N-thony could suggest who might take this small point on?

Vitalis95:
Fixes #7772 
@rdstern @N-thony @lloyddewit , this PR makes the default position (Start),  when the row variable is added to different data frames.


lloyddewit (approved):

lloyddewit:

rdstern (approved):
Another good step.  Thanks @Vitalis95

Issue #7727, opened 2022-08-04
Graphs of high resolutions


Patowhiz:
Users asked if R-Instat could be produce graphs with high resolution.
When looking into this, I found that for the output viewer, we are technically taking a snapshot of the graphs produced using a fixed width and height. This interfere with the original resolution of the graph as produced by the functions.

I suggest getting the graphs images directly from the graph objects.

rdstern:
I like this, because I suggest it resolves a major problem in the current version of R-Instat.  This currently saves the graph itself, somewhere, and is therefore lost when the current R-Instat session finishes.
David has proposed we save the commands instead - and again, back as part of the R-Instat data book. 

I now propose that this also solves the resolution problem for the graph.  This has been considered in various questions - [here is one example](https://stackoverflow.com/questions/38907514/saving-a-high-resolution-image-in-r).

PR #7782, opened 2022-08-17
Enhances the output viewer to display html outputs (in windows machines), improves viewing of graphs and text outputs and improves processing of objects

### LINKED ISSUE #7121: Investigate and (hopefully) implement HTML in the new output window
A number of our existing dialogues produce results in html.  This currently displays in a browser.  It would be great if this could be in the output window instead?

The dialogues include:
Prepare > Column: Factor > View Levels/Labels 
Describe > One Variable > Rating Table
Describe > One Variable > Frequencies (optional)
Describe > Specific > Frequency Tables
Describe > Specific > Summary Tables

### LINKED ISSUE #7778: Output object types in R-Instat, how to process and display them at both R and .Net level
@dannyparsons @ChrisMarsh82 @lloyddewit @rdstern @lilyclements. 
Currently, we can only display 2 R outputs in the output viewer; **Graphs** and **Text**.

**Graph outputs**
These are displayed by internally running an extra R command, `grDevices::png`,  that takes a "snapshot" of the printed graph image with a specific resolution and saves it as a file that is then displayed in the output viewer.

**Text outputs**
These are displayed by internally running an extra R command, `utils::capture.output`, that takes the last R command and uses it as an input parameter to the `utils::capture.output` command. The results are then displayed in the output viewer.

In the .Net RLink, the detection of the expected output type is done by checking on; 

- Call type.
- Number of R script command lines. 
- Source of the R script. 

**My suggestions**

1. We detect the output type based on R-Instat "custom" R script commands, for instance we could have functions; `view_graph_object`, `view_text_object`, `view_html_object`. These R functions will detect if there is an R "viewer", if it's not there, like in the case of R-Instat, they save the object as a temporary file which can then be displayed in R-Instat. This will also simplify the dialogs implementations, developers won't need to change the `ICallType` for the different output types. All they need to do is use the custom R functions. 
2. At .Net level, we get all R outputs (the R scripts excluded) from the created files. This can be done by just getting the returned file path. This has the advantage of using less memory especially when the output viewer is implemented as a "_Recycle Viewer_". And also saving all the contents of the output viewer can be greatly simplified. This can also greatly improve the memory usage, we won't need to store big objects like html in the memory.

PR #7782 adds a html output type that can now be viewed in the output viewer. It also aims to demonstrate the above concepts.
Basically, dialogs that produce html objects will only need to call  `view_html_object` R function. That function checks of there is an R viewer, if not, creates a temporary file and returns it's file path. The at the .Net level, we file path is retrieved and the file is displayed in the output viewer.
This implementation can apply to graphs (by saving the graph objects with original resolution) and also text outputs.
I'm looking forward to your feedback.

### LINKED ISSUE #7808: How we process, store and retrieve objects in the R data book
@lloyddewit @ChrisMarsh82 @dannyparsons @rdstern @lilyclements 
Currently we rely on the class type of objects to identify and filter the objects contained in the data book in R.
Some packages produce objects that have multiple class types which overlap with other "category" of objects.
For instance;
1. Graphs could be in html "category". 
2. Tables could be in html or text "category".

That means if I want to identify graphs only in the set of objects. I will need to check more than the object class type.
Functions `add_objects` , `get_objects`  and `get_object_names` in R level demonstrates my above observations.

I suggest when adding objects in the data book we also require an extra  parameter for specifying the "type_label" of the object. 
This "type_label" can then be used to determine how we process, store and retrieve the object. This may require changing the data structure used for holding the objects.

Patowhiz:
Fixes #7121
Fixes #7808
Fixes #7778
Fixes partly #7727

**Dialogs affected**
1. Prepare > Check Data > One Variable Summarise. Demonstrates how text and table outputs are displayed. 
2. Prepare > Check Data > View/Delete Labels
3. Describe > Two/Three variables > Pivot Table
4. Describe > Specific Tables/Graphs > Frequency / Summary Tables

**Dialogs that need discussions before refactoring**
1. Describe > Two/Three variables > Summarise

Dialogs that can be refactored by the team to use the new implementation;
1. Describe > Two/Three variables > Two-Way Frequencies
2.  Describe > Two/Three variables > Three-Way frequencies
3. Any others?

**Issues known**
1. Every time a html output is produced, the output view adds the web control which initiates a new browser process similar to the way a new tab is created in Chrome. This means more memory usage. There may be need to prioritise how the output viewer can reclaim back that memory when the user is no longer viewing the html output..
2. Currently viewing of html objects that produce 2 html files is not supported. The R function added only supports viewing of a single html file.
4. Plotly html objects produce html files that don't return the correct CSS height. I'll have to investigate a work around.
5. Some dialogs may take some time to produce the html objects. This is not related to the new control that only reads from the produced html file dynamically.
6. At times the output window does not scroll smoothly. This is an existing issue.

New enhancements and concepts introduced by this PR
1. Viewing of html objects in R-Instat.
2. Maximising of outputs in the output window.


Patowhiz:
@rdstern could you try testing this. I'm not sure if the Nuget `CefSharp` libraries will be downloaded automatically.
If they don't I'll have to test this with @lloyddewit first to understand why.
If the packages are not copied directly R-Instat will fall back to using the browser.
kindly let me know

lloyddewit:
@Patowhiz 
I was able to build this PR.
I checked out the PR branch and it showed lots of red-underlined errors in my open editor windows.
When I started a build, I got a message in the VS status bar that it was restoring NuGet packages. This solved the compilation errors (see below).
image

When I ran the PR and generated a table, I go the following result. Is that what you would expect?

image

Patowhiz:
@lloyddewit yes! Thank you very much. 
Do you prefer we "git ignore" the files you suggested before @rdstern starts the testing?
I'm not sure how to do that, but I can try. Still your help will be highly appreciated. If you prefer doing that quickly and just updating this PR that's also okay.

lloyddewit:
@Patowhiz I could experiment a bit with the minimal number of cef files that we should include in this PR. I could do this today or tomorrow.
For me, it would be acceptable, if the first time a developer checks out this PR, then VS updates its development environment. I think this is a normal/acceptable approach.

lloyddewit:
@Patowhiz 
I investigated if we could remove the new NuGet packages from this PR.
I suggest to make the changes below to the `R-Instst/.gitignore' file. I didn't want to push this change directly to this PR.
I tried several times to create a temporary branch with this PR's changes without the NuGet packages for testing, but Visual Studio gave me an error when I tried to load the project.
If you make the change, and it works, then I don't know if the cef files will be automatically removed from this PR, or if we need to make a new PR without the cef files.

We could also consider a more general approach and remove all the NuGet packages from our git repo. See [here](https://docs.microsoft.com/en-us/nuget/consume-packages/packages-and-source-control).

image

```
# PR #7782 adds a feature to display html in the output window.
# This requires the cef packages. These packages are more than 0.4 GB.
# We want to prevent these packages from being stored in the R-Instat GitHub repository.
/packages/cef.redist.x64.104.4.18/
/packages/cef.redist.x86.104.4.18/
/packages/CefSharp.Common.104.4.180/
/packages/CefSharp.WinForms.104.4.180/
```

rdstern:
This is looking great - very exciting.

I also like the maximise feature.

However I am not supposed to praise, but be critical.  So, for the graphs, when I maximise the new version looks as though the top has been chopped a bit.  Compare a boxplot with the R-Viewer:
image

I realise there is just one format, currently, for saving graphs, and that will be extended later.

I really like the feature that I can just work in the output window and then maximise and look at any of the graphs in more detail later.

On the tables it seems to work.  Here is a table previously and now.

image

I note - again the top line is missing now, and the default font/formatting seems different.

I also had one instance (only 1) for the following black square for a simple table.  When I maximised it was there fine.
image

This was just the following table:
image

I suggest we will want to take advantage sometimes of the ability (at least with the tables) to produce latex - and I presume rtf, as these are the 3 for the gt package?

Patowhiz:
This PR now aims to fix issue #7121 and #7808.

lloyddewit:
@Patowhiz thank you for the recent changes. 
Is this PR now ready for final testing and review or do you still expect to make changes.
What's the status of the 700+ cef files?
thanks

Patowhiz:
@lloyddewit I amended the git ignore file as suggested. I think we may have to delete the already pushed `cef` files manually. 

I decided to make changes that affect processing of objects this week thus expounding the original issue. It strongly relates to the improvements we want in regards to outputs that are displayable by the output viewer. This includes major R level changes that @lilyclements will review.

The changes now affect all dialogs that produce an output that is displayed by the output viewer.
@rdstern changes made also allow us to save text summaries like `skimr` outputs.
  
@lloyddewit @rdstern I'm now going through every dialog to make sure that I'm aware of all the regression introduced in this PR. The bigger task is making the dialogs take advantage of the new simplified implementation. @lloyddewit I suggest you have a look at the code (especially the deeper) before I re-task the team here to undertake that task.

Patowhiz:
@lilyclements this is ready for your R review.
@lloyddewit this is ready for your review as well.

@rdstern I'm already aware of some of the cosmetic issues and the output viewer issues. I'll address those in a separate PR. This PR has big changes. You can test this as well. In particular notice how from dialog **One Variable Summarise** as indicated above, we can now save such summary objects and even maximise them just like graphs and tables.

Thanks.

rdstern:
@Patowhiz I tested with one-variable summarise, and with the Prepare > Check Data > View Labels that also directs data to the new output window.  Looks good.
I suggest it is time to go ahead and make the hopefully minor changes in the remaining dialogues so all the html output goes to the new window.
Two small points:
a) I got the black screen once again as mentioned above.
b) In a maximised graph it is not quite the same as the R-viewer.  There it starts as the same size as (I assume) the original and then maximise changes the graph to fill the screen.  And you can have intermediate sizes which is sometimes useful also.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -561,55 +592,123 @@ DataBook$set("public", "create_graph_data_book", function() {
...
-        else return(out)
       }
     }
-  }
+  
 }
 )
 
-DataBook$set("public", "get_object_names", function(data_name, include_overall = TRUE, include, exclude, type = "", include_empty = FALSE, as_list = FALSE, excluded_items = c(), internal = TRUE) {
+DataBook$set("public", "get_object_names", function(data_name, object_type_label, include_overall = TRUE, include, exclude, type = "", include_empty = FALSE, as_list = FALSE, excluded_items = c(), internal = TRUE) {
```
lilyclements, line 653: remove `type = ""`?

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -561,55 +592,123 @@ DataBook$set("public", "create_graph_data_book", function() {
...
+        } 
+      }else{
+        return(self$get_data_objects(data_name)$get_object_names(object_type_label, as_list = as_list, excluded_items = excluded_items))
+      } 
+    }
+    
+  }
+  
+  
+  #todo. deprecated. Delete after doing the necessary refactoring.
```
lilyclements, line 711: Is this still a "TODO" out of interest?
If so, is this written up in an issue?

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -2596,3 +2597,153 @@ read_corpora <- function(data){
...
+  #if there is a viewer, like in the case of RStudio then just print the object
+  #this check is primarily meant to make this function work in a similar manner when run outside R-Instat
+  r_viewer <- base::getOption("viewer")
+  if (!is.null(r_viewer)) {
+    #When print command is called in R-Studio, a temp file is
+    #automatically created. 
+    #TODO. Investigate how that can be done in R-Instat
+    print(graph_object)
+    #return empty file path. Important for RStudio to display the object
+    return("")
```
lilyclements, line 2631: Why can't it return the `graph_object`?

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -2596,3 +2597,153 @@ read_corpora <- function(data){
...
+  #   
+  # } else if ("openair" %in% object_class_names) {
+  #   
+  # } else if ("ggsurvplot" %in% object_class_names) {
+  #   
+  # } else if ("recordedplot" %in% object_class_names) {
+  #   
+  # }
+  
+  message("R viewer NOT detected. File saved in location ", file_name)
```
lilyclements, line 2660: Do we use capitals elsewhere in warning messages in R-Instat?

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -2596,3 +2597,153 @@ read_corpora <- function(data){
...
+  #if there is a viewer, like in the case of RStudio then just print the object
+  #this check is primarily meant to make this function work in a similar manner when run outside R-Instat
+  r_viewer <- base::getOption("viewer")
+  if (!is.null(r_viewer)) {
+    #When print command is called in R-Studio, a temp file is
+    #automatically created. 
+    #TODO. Investigate how that can be done in R-Instat
+    utils::capture.output(text_object)
+    #return empty file path. Important for RStudio to display the object
+    return("")
```
lilyclements, line 2679: Do you know why `return(utils::capture.output(...))` doesn't  work?

lilyclements (changes requested):
Looks good!

The View Objects and View Graph dialogs aren't displaying the object names for me if I create a graph (I haven't checked for other objects)
image

Otherwise, I just have some minor comments.

Patowhiz:
> @Patowhiz I tested with one-variable summarise, and with the Prepare > Check Data > View Labels that also directs data to the new output window. Looks good. I suggest it is time to go ahead and make the hopefully minor changes in the remaining dialogues so all the html output goes to the new window. Two small points: a) I got the black screen once again as mentioned above. b) In a maximised graph it is not quite the same as the R-viewer. There it starts as the same size as (I assume) the original and then maximise changes the graph to fill the screen. And you can have intermediate sizes which is sometimes useful also.

I will look into the black screen in a separate PR. It may be a HDPI issue.
The maximised graph is indeed not the same as the R-Viewer. When the R viewer is maximised, it literally reruns the R graphics code that draws the plot in the R viewer(sort of like redrawing or repainting). In principle that's how R graphics works, it paints on the given size of the screen.
Our custom maximised form just loads an existing image file, and displays it in the widened screen, the attempts to display it while maintaining the aspect ratio. We could explore doing what the R viewer does in a separate PR. We may have to be a bit clever here.

instat/static/InstatObject/R/stand_alone_functions.R
Patowhiz: It can.

instat/static/InstatObject/R/stand_alone_functions.R
Patowhiz: changed to small letters.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -561,55 +592,123 @@ DataBook$set("public", "create_graph_data_book", function() {
...
-      }
-      else {
-        out = sapply(self$get_data_objects(as_list = TRUE), function(x) x$get_objects(type = type))
-        if(include_overall) out[[overall_label]] <- private$.objects[self$get_object_names(data_name = overall_label, type = type)]
-        if(!include_empty) out = out[sapply(out, function(x) length(x) > 0)]
-      }
-      if(!missing(object_name) && length(object_name) == 1) {
-        if(print_graph && (ggplot2::is.ggplot(out) || any(c("gg", "ggmultiplot", "openair", "recordedplot") %in% class(out)))) return(print(out))
-        else return(out)
+DataBook$set("public", "get_objects", function(data_name, object_name, object_type_label, include_overall = TRUE, as_list = FALSE, include_empty = FALSE, force_as_list = FALSE, internal = TRUE, ...) {
```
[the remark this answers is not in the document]
Patowhiz: Because it's use has been removed. printing of objects has generally been moved to `view_object` function. That now applies to graphs as well.

instat/static/InstatObject/R/labels_and_defaults.R
```diff
@@ -58,10 +58,12 @@ structure_type_3_label="Measurement"
...
 #object labels
 overall_label="[Overall]"
-model_label="model"
 graph_label="graph"
 table_label="table"
-surv_label="surv"
+model_label="model"
+structure_label = "structure"
+summary_label = "summary"
+surv_label="surv" #todo. deprecated.
```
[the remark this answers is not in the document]
Patowhiz: Yes. Removed

instat/static/InstatObject/R/instat_object_R6.R
Patowhiz: Yes. removed

instat/static/InstatObject/R/instat_object_R6.R
Patowhiz: I've deleted the entire code block. No longer needed.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -2075,26 +2086,42 @@ DataSheet$set("public", "get_objects", function(object_name, type = "", force_as
...
-    lst = list()
-    lst[[self$get_metadata(data_name_label)]] <- out
-    return(lst)
-  }
-  else return(out)
+DataSheet$set("public", "get_object_names", function(object_type_label, as_list = FALSE, excluded_items = c()) {
+    if(missing(object_type_label)){
+      out = names(private$objects)
+    }else{
+      #todo. check if type is recognised?
```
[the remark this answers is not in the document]
Patowhiz: Removed. No longer necessary

lilyclements (approved):
Thanks for the changes - this looks good to me!

Patowhiz:
Thanks @lilyclements for reviewing this. Next task is to discuss how these objects can be saved and restored after the data book is saved by the user.
As discussed, @rdstern @lloyddewit I have written the issues expected and follow up fixes after this PR is merged in issue #7852.

rdstern (approved):
Now I have reproduced the problem systematically.  It is just the first time it uses the output window for html.
a) I first tried with the survey.  I used the Frequency/summary tables.  
When I use it again it is fine.
b) I then restarted R-Instat and used the Prepare > Check Data > Labels dialogue - same result.  Then I repeat the dialogue and it is fine.
c) So I start again and use a different dataset - the diamonds from the library.  Do a frequency table and again get the problem.  Then repeat and it is fine.
d) Fourth try.  Start again, use survey, and get a text output.  That's fine.  Then get a frequency table - same problem.
Repeat the dialogue and fine.

By the way it still gives the black screen (momentarily), but then it goes and the result is displayed.

e) Fifth try.  Start again. Survey dataset.  Use Pivot table.  Same problem.  Use it again and it is fine.

image

I can live with it, and so am approving.  If the code is ok I could accept it as now, so we merge quickly, and it can be fixed later.

rdstern (dismissed):
Now I have reproduced the problem systematically.  It is just the first time it uses the output window for html.
a) I first tried with the survey.  I used the Frequency/summary tables.  
When I use it again it is fine.
b) I then restarted R-Instat and used the Prepare > Check Data > Labels dialogue - same result.  Then I repeat the dialogue and it is fine.
c) So I start again and use a different dataset - the diamonds from the library.  Do a frequency table and again get the problem.  Then repeat and it is fine.
d) Fourth try.  Start again, use survey, and get a text output.  That's fine.  Then get a frequency table - same problem.
Repeat the dialogue and fine.

By the way it still gives the black screen (momentarily), but then it goes and the result is displayed.

e) Fifth try.  Start again. Survey dataset.  Use Pivot table.  Same problem.  Use it again and it is fine.

image

I can live with it, and so am approving.  If the code is ok I could accept it as now, so we merge quickly, and it can be fixed later.

Patowhiz:
@rdstern thanks for reviewing this.
I have added the issue you have identified in issue #7852 as one of the high priority follow ups.

Patowhiz:
As discussed with @lloyddewit this PR has now been replaced by PR #7853.

PR #7785, opened 2022-08-17
Fixed bug with Degree option in Climatic > Prepare > Transform dialog

### LINKED ISSUE #7766: Transform dialogue degree options don't seem to work now
This is the Climatic > Prepare > Transform > Degree option.

It was added by @anastasia-mbithe and seemed to work fine.  Recently @Vitalis95 added the option for tmax and tmin to be added by default if they were recognised as climatic.

Now we tried in the DRC workshop and it gives an error.  I tried the simple diurnal range and then the tmean option.  Both gave an error.  This was in a dataset with many stations.
I tried today with the dodoma (Tanzania) data from the library and also get an error.  It would be good to fix, and perhaps also check why it has appeared.  Did we just miss this on earleir checking or has something else changed - which might then affect more than just this option?

Vitalis95:
Fixes #7766 
@rdstern @lloyddewit @N-thony , this PR fixes a bug in Transform climatic dialog, Degree option.


rdstern (approved):
Great.  Thanks @Vitalis95

lloyddewit (approved):

Issue #7770, opened 2022-08-14
Add facilities to analyse lists


rdstern:
In  `File > New Data Frame > Lists > Words/Literature > Shakespeare/sonnets`, the variable list.lines comes into R-Instat as a list.

![image](https://user-images.githubusercontent.com/11226469/184530167-98e1eb32-4f2d-4bb3-a2e2-d65d83afed04.png)

There is a bit of initial work for @lilyclements and then (assuming she agrees) most could perhaps be done by @anastasia-mbithe ?

a) I have added the request for lists to have a proper data type into issue #7493 because that is on adding another type of data into R-Instat.  But having a type called (L) is more urgent, because currently it has nothing and hence looks numeric.
b) It looks like a complicated character variable, but the function `data_book$convert_column_to_type(data_name="data", col_names="list.lines1", to_type="character")`
doesn't work.  It converts all values to NA.  Please could this function work.  (This is probably a @lilyclements task?)
c) The menu `Prepare > Column: Text` mainly uses the `stringr` package.  These functions seem to work directly on lists, but most of the dialogue options don't allow the variable, because it isn't factor or character.  Please allow this option to be added.  Perhaps @anastasia-mbithe could do this?  The Split command does allow the list variables.
d) Improve the `Prepare > Data Reshape > Stack > Unnest ` option to be able to stack these data into line by line.  It is an excellent example of multiple response.  
0) The `unnest `function we use is from the tidytext package.  There is now one, with the same name in `tidyr`.  I think we are still ok using the one we have, but a check by @lilyclements would help.   
1) I assume it will work directly on lists, so allow that type of data - just as discussed in c) above.
2) The option `paragraph` allows a separator, called `paragraph.break` but there is no option in the dialogue.  So add this option.  We could call it `Pattern` perhaps?
3) Similarly when token is regex there is a `pattern =` possibility.
4) Check, but I think, in each case we could usefully add our regex keyboard as an option for the pattern.

Find what works for this example?  I am almost there below:
![image](https://user-images.githubusercontent.com/11226469/184533330-5340975e-b42b-413d-a8db-741e15cc50a6.png)

I used the lines as follows:

```
# Code generated by the dialog, Stack (Pivot Longer)

list.lines1 <- data_book$get_columns_from_data(data_name="data", col_names="list.lines1")
data <- data_book$get_data_frame(data_name="data")
[...]

PR #7790, opened 2022-08-19
Included list variables in the Prepare> Column: Text Options dialog

anastasia-mbithe:
This fixes part c. of #7770

@rdstern, I have added lists as variables to be included in the Text: Transform dialog. However, I noticed that the other Text options as Split, Combine ...and the others included lists before because they are not restricted to either factor, character, or even list, so am wondering if we should restrict the variables (for the other options) to only list, factor, and character?


lloyddewit (dismissed):

rdstern:
@anastasia-mbithe good spot.  I now have some numeric columns that I would like to trim the blanks.  So let's allow them too.  So really all types

anastasia-mbithe:
This is ready for review.

rdstern (approved):
Seems fine

lloyddewit (approved):

Issue #7792, opened 2022-08-19
Dialogues where the layout in French is incomplete


rdstern:
I simply list dialogues I have found where problems remain:
I have started with the Climatic menu:

a) Export to cpt.  The top radio buttons 2 and 3 seem incomplete

b) Climatic > Tidy and Examine > Stack.
![image](https://user-images.githubusercontent.com/11226469/185665578-e3e10830-3b06-4262-9221-4227b9c4d46b.png)
The plusiers line may be inclmplete.  It is certainly odd when the checkbox is checked.
c) In the same dialogue - and in many, the line below is complete, but without the colon that is there in English and it would be good if the receiver were moved just a bit to the right.  See the same in the unstack dialogue and append, and others.
d) One way frequencies.  Last option, when checked the receiver overlaps the label.  (The translation to English is odd if the Lan button is pressed.  It stays in French.)
e) Dates - nothing found
f) Define climatic Towards the bottom check duplicates is incomplete in French.
g) Climatic >Check data > Inventory Missing button at top.  Then when save details is checked the control needs to move right.
h) QC rainfall dialogue - missing months button is incomplete in French.  And next control overlaps.
![image](https://user-images.githubusercontent.com/11226469/185671540-87bb93dd-c13f-4a50-a372-10b94b799af1.png)
i) The position control, see below - the C is not available in the French.
![image](https://user-images.githubusercontent.com/11226469/185672079-44725732-aa07-460b-b39b-3fdd1370c28a.png)

j) The Climatic > Prepare menu now.  Starting with Transform.  Multiple Spells at the top.  Is that complete in French?
![image](https://user-images.githubusercontent.com/11226469/185672527-1e78cece-be3b-481a-a999-b889ff0bd5c9.png)
k) Transform there Day length  in incomplete see below:
![image](https://user-images.githubusercontent.com/11226469/185673895-8ced8315-17d3-4539-a3d1-76e742ea2254.png)

rachelkg:
@rdstern issue 
a) changed layout and updated crowdin translation
b) I have passed this on to Antoine as it was too complicated for me see issue: Layout issues with french version of Stack (Pivot) dialog #7794
c) is a problem with the Save.urc I would love to see it resolved but have not been able to get anyone to work on it, I have mentioned it again to Antoine in issue #7794. Please see issue #6700 for more details[https://github.com/africanmathsinitiative/R-Instat/issues/6700]
[...]

PR #7796, opened 2022-08-19
Adjusted layout of Climatic menu dialogs to allow longer French text

rachelkg:
@rdstern please could you check this. it is the issues you pointed out in issue #7792


rdstern:
a) looks ok now
b) Still a problem:
image
c) I think that was handed on.
d) Is now ok.  The comment about one phrase staying in French when Lan pressed, is (I think) something different.
e) Dates still ok I assume!
f) ok now
g) ok
h) ok
i) ok - I just note to remember for the future that I am not sure that spells into sorts (in French) is correct.  I think these are magic spells, but are they duration spells in French too.
j) ok

rdstern (approved):
I only had a single problem - item b), that I think has been passed to Antoine to fix elsewhere.  So I am approving.  @lloyddewit as it affects many dialogues it would be good to merge soon.

lloyddewit (approved):

rachelkg:
@rdstern I have resolved the i) spells issue in crowdin.

PR #7802, opened 2022-08-22
Cosmetic improvement to the stack dialog

### LINKED ISSUE #7798: Minor cosmetic improvement to the stack dialogue.
The stack dialogue is generally working well.  Here it is:
image

Usually I use it in default mode to carry all columns.  This time I chose to select columns to carry.  That's made it awkward and a bit of a pain to use.  This may be automatic and apply to other dialogues too.

a) When used in default mode, there is just the one multiple receiver. And all is fine.  
b)With the setting above there are 2 receivers, and it keeps moving to the second, while I usually want it to remain in the first receiver.
c) So with the above - try removing a single variable from the first receiver.  It now jumps automatically to the second.  In fact whenever you do anything to the the first receiver it seems to be so fond of the second receiver that it insists on jumping there!

Ivanluv:
fixes #7798 


lloyddewit (approved):
@Ivanluv Looks good, thank you for the clear/concise changes.

lloyddewit:

rdstern (approved):
Great

PR #7813, opened 2022-08-25
Added reverse function to the calculator string keyboard

### LINKED ISSUE #7807: Add reverse to the R-Instat string functions
I am suggesting @anastasia-mbithe for this, because she is already working on the calculator.  Also this is a function I am assuming for the maths camps!

Could a function called reverse be added to the text/string keyboard
image

I suggest it could go on the second-to-bottom row so the keys line up with the bottom row, each having 6 keys.
it gives stringi::stri_reverse( )  (notice it is stringi, and not stringr, which is why I missed it before)

anastasia-mbithe:
Fixes #7807 
@rdstern and @lloyddewit, This is ready for review.


lloyddewit (approved):

lloyddewit:
@rdstern If you can test/approve, then we can merge, thanks

rdstern (approved):
Good

PR #7823, opened 2022-08-29
Improved the Probability and Integer keyboards in the Calculator

### LINKED ISSUE #7800: Some simple improvements to the calculator
@anastasia-mbithe I hope you could do this.  I am writing the possible video script called **_Fun with numbers_** that uses this dialogue.

a) In our new integer keyboard, the binary key includes an as.integer function at the start.  This has the problem that all integers more than just over 1000 come as NA.  Try with a data frame of 10,000 (say) and the first variable being 1, 2, 3, ... (That's now easy to produce with the File > New Data Frame).
I've been looking to define it differently, because if you just omit the `as.integer`, then it comes as a character variable.  But I can't find the appropriate type, so I think we'll have to put up with it being a character.  Happy if you could explore too.
b) In the main keyboard could you add spaces round +, -, * and /.  So pressing them gives <space>+<space>, etc.
c) In the probability keyboard there is pbirth and qbirth.  pbirth currently gives `pbirthday( )`.  But that doesn't accept a column, just a number.  So please change it to `sapply( ,pbirthday)` with the cursor in the space before the comma.  That will allow a variable to be given.  Same for qbirth.  Check with a column of integer values.  (And with fractions for qbirth.)

### LINKED ISSUE #7819: Correct the pbirth and qbirth keys in the calculator
@anastasia-mbithe I hope it is ok, that I continue to suggest you work on the calculator?
pbirth currently gives an error if you give it a variable, rather than a number.
It is a good example to show the use of the sapply function.  
So it should change as a default to become a use of sapply, see below:

image

Also note that currently when you ask to see the parameters there is an extra close bracket.

The default, when you don't have the parameters is presumably 
sapply( , pbirthday), ready to put a variable before the comma.

The same with the qbirth function.

anastasia-mbithe:
Fixes #7819 and  Fixes #7800 also.
 @rdstern, This is ready for review.

### DIFF
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -2438,7 +2438,7 @@ Public Class ucrCalculator
 
     Private Sub cmd_Click(sender As Object, e As EventArgs) Handles cmdPbirth.Click
         If chkShowParameters.Checked Then
-            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("sapply(X = , FUN= pbirthday)", 18)
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("sapply(X = , FUN= pbirthday, classes= 365, coincident= 2)", 47)
         Else
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("sapply( , pbirthday)", 13)
         End If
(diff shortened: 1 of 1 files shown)


lloyddewit (approved):

lloyddewit:

rdstern (approved):
Excellent

PR #7831, opened 2022-09-05
 Correcting y axis on Describe > One Variable > Graph

### LINKED ISSUE #7821: Correct and hopefully also improve the Describe > One Variable > Graph dialogue
As with the stem plot I suggest this is for @N-thony to allocate.

In this dialogue the default bar chart can be pretty bad.  Here it is for the field variable in the survey, made into a factor, and flipped.  Note it is in character, rather than the factor (numerical) order.  In an ordinary factor like the variety it ignores the order of levels and just uses its own alphabetical.

image

Here is the code:
```
# Code généré par la boîte de dialogue, One Variable Graph
last_graph <- data_book$graph_one_variable(data_name="survey", columns="field", numeric="violin_jitter", categorical="geom_bar", coord_flip=TRUE)
data_book$get_graphs(data_name="survey", graph_name="last_graph")
```
I can't see what is wrong, because it is in our own graph_one_variable code.

Once this is corrected, it would be good to check other other options and also see if there can be useful additions to the dialogue.  Being able to add labels on the bar charts would be welcome.

lilyclements:
Fixes #7821

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -2407,8 +2407,13 @@ DataSheet$set("public", "graph_one_variable", function(columns, numeric = "geom_
     }
     else {
       stop("Cannot plot columns of type:", column_types[i])
-    }    
-    curr_data <- self$get_data_frame(stack_data = TRUE, measure.vars = columns)
+    }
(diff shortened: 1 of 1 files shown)


lloyddewit (approved):

lloyddewit:

Issue #7749, opened 2022-08-09
Translation issue on Data Frame, Check Data, Column:Numeric, and Data Reshape sub-menu of Prepare menu


MeSophie:
Using the French translation of R-Instat 0.7.7 there are some sub-menus text and dialog text in **Prepare** menu which are still in English (refer to figures )

![image](https://user-images.githubusercontent.com/106779233/183629687-7e800548-4b9b-4ef5-a774-e81706e014be.png)

1- Prepare>Data Frame
![image](https://user-images.githubusercontent.com/106779233/183630219-11be310b-5275-4460-a15c-59adfd7bc447.png)

2-Prepare>Check Data
![image](https://user-images.githubusercontent.com/106779233/183630708-14273cd7-1835-42fd-a2f4-0665988c9a63.png)

3-Prepare>Column:Numeric (all dialog text  are not translate)
![image](https://user-images.githubusercontent.com/106779233/183630950-0ce85ebc-7b29-41fb-816a-7adf93deabb0.png)

4-Prepare>Data Reshape
![image](https://user-images.githubusercontent.com/106779233/183631650-4815108d-8c2a-489c-a750-c2f18bdba782.png)

MeSophie:
I noticed that  **Random Split...** and **Scale/Distance...** are already translate in Crowdin but still in English in R-Instat. For **Merge (Join)...** and **Stack (Pivot Longer)...** I fixes it in Json file.
Please @lloyddewit and @N-thony  can you have a look ?

![image](https://user-images.githubusercontent.com/106779233/196410461-b3adeb07-a5b5-42ba-9eb1-a9d65d0672c7.png)

![image](https://user-images.githubusercontent.com/106779233/196412487-9399e88b-e0cb-40e2-857c-ddfbfbfab6ac.png)

![image](https://user-images.githubusercontent.com/106779233/196413094-e56c0b79-ff42-4652-8f91-86053e4efd40.png)

It is also the same  with **Number of Preview Rows:** (Tools>Options>Import).

![image](https://user-images.githubusercontent.com/106779233/196415278-88f2a952-8cbb-4710-9936-6170e76e5f73.png)

![image](https://user-images.githubusercontent.com/106779233/196415349-123cf5b1-623c-45b1-98de-d8b53be7a40d.png)

N-thony:
@MeSophie good, it is probably that those controls are not in the database or maybe they are but named differently.
@Vitalis95 would you like to look at it?

[...]

PR #7841, opened 2022-09-06
Update controls name in the database

N-thony:
Fixes (partially) #7749 
@lloyddewit I made the following changes in the database. I found that some controls name were named different in the database and it is one of the reason some controls are not translated
image
image


N-thony:
> Fixes (partially) #7749 @lloyddewit I made the following changes in the database. I found that some controls name were named different in the database and it is one of the reason some controls are not translated image image

I have added this control too
image

lloyddewit (approved):

PR #7851, opened 2022-09-07
Added Apsimx Examples dialog

### LINKED ISSUE #7532: Add a menu for Crop Models
This will be specified by me together with Sophie and Francis, once Sophie is on github.
I suggest the Options by Context Menu will (at the same time) be called Experiments
I also suggest we have a main menu item called Crop Model.  This will implement the apsimx package in R-Instat.  

a) One initial item could be Examples, which will open any of the built-in examples of data.  This will be a simple dialogue to implement the apsim_example function.    The command runs fine and all that needs to be added is to put the data into data frames in R-Instat.
b) Then I don't want Sophie to get stuck on the climatic data part.  And I hope she can work with Amsale on this aspect, so Sophie concentrates on the actual crop modelling.  On the other hand it looks as though using satellite-based data, e.g. chirps, is really easy.  So perhaps that can be included.
c) So then we move to using Apsim for crop modelling, assuming we have a met file.  The Apsim documentation says to use their gui, i.e. to simply use Apsim separately from R-Instat.  If that's best and there is nothing we can add into R-Instat to make that process easier, then so-be-it.  So here is where I need you to be able to use ApSim yourself, so you can provide guidance on what could usefully be added into R-Instat (if anything) the make Apsim easier, for R (Instat) users.

MeSophie:
Fixes #7532.

@rdstern and @lilyclements Please have a look.
Thank you.


N-thony:
@MeSophie this branch has conflicts, please resolve them.

lilyclements (dismissed):
Looks great! Nice job. @rdstern what do you think?

rdstern:
@MeSophie this is a good start.  These first comments are so you can complete this dialogue and it can be merged.
So please check each option and correct any aspects that are needed, so each example works.
I found Controlled Environment and Eucalyptus Rotation don't work and give the same error.  Then Rotation didn't work and gave a different error - I think it might be trying to write multiple data frames.

I stopped checking there.  Please check all the rest.  Once those are all working we should be able to merge the dialogue.  I didn't check what happens with Silent.

rdstern:
I wonder what your next thought was, when you saw the resulting data frames.  My immediate thought was - what's the corresponding climatic data for each example.  I think it is interesting that the assumption is that the analysis will just involve the yield data.  I would like to also be able to relate the variability in yields to those in the climatic data.  So I'd like both!

And, of course it would be good to know of the siol and management data too.  It is so useful that R-Instat makes it easy to look at multiple data frames.

So, I started by looking at the commands in the package - but couldn't see how to do this.  I did find that there is a link with Chirps and that there is a chirps package.  You could discuss this with @johnargentino as I wonder if that could become our easy way of importing chirps data in general?

Then I saw that the main vignette does seem to discuss the example data in much more detail.  I assume your current dialogue may work without installing Apsimx, but the other data need the package to be installed.  Good to investigate.

I imagine the dialogue will then be extended, so you could sort of import either just the yield data ar also the other data.  Let's see.

MeSophie:
@rdstern  The packages available are limited and all independent of each other. The ones I found interesting are **apsim** and **apsimx** which allow to read **.apsim and .apsimx files**, **apsimx_soil_profile** (which allows to create a soil profile), **as_apsimx_met** which transforms the climate data tables into **.met files**, **get_chirps_apsim_met/get_power_apsim_met**.  In order to link all these parameters, we need to understand the mathematical link between them and perhaps only then will one be able to simulate the model directly in R-Instat without using the APSIMX GUI. 

What we can do for the moment is to see to what extent we can use the **apsimx , inspect_apsimx and edit_apsimx** functions in a diaog. The user will be able to  import an .apsimx/.apsim file, inspect and edit it.

Also the majority of these packages require the installation of APSIMX software, this is also the case for the apsimx_examples and apsim_examples packages.

MeSophie:
@rdstern  I fixes  the bug generated by Controlled Environment, Eucalyptus Rotation and  Rotation models. Please, can you review the dialogue again? Thank you.

instat/dlgApsimx.vb
```diff
@@ -15,27 +15,135 @@
...
+        clsApsimExampleFunction.SetPackageName("apsimx")
+        clsApsimExampleFunction.SetRCommand("apsim_example")
+
+        ucrBase.clsRsyntax.SetBaseRFunction(clsBaseFunction)
+    End Sub
+
+    Private Sub SetRCodeForControls(bReset As Boolean)
+        ucrChkSilent.AddAdditionalCodeParameterPair(clsApsimExampleFunction, ucrChkSilent.GetParameter, iAdditionalPairNo:=1)
+        ucrInputComboList.AddAdditionalCodeParameterPair(clsApsimExampleFunction, New RParameter("example", 0), iAdditionalPairNo:=1)
+
```
N-thony, line 103: can you remove this line?

rdstern (commented):
For me it worked fine down to Wheat.  From then on it said as follows:
image

Now I had installed Apsim, but not Apsim classic.  I assume I needed to install the new Apsim or I would not have got any of the examples?  I am not sure I particularly need to install ApSim Classic, so I might have to do without those examples.  If that is the case, then perhaps the dialogue might make this all clear?

N-thony:
> Thanks @rdstern, I hadn't notice that because I had both apsim installed on my pc. To run all the examples below wheat you have to install Apsim classic because they are the examples of Apsim classic. The dialog uses two functions apsimx_example for the Apsimx Next Gen examples and apsim_example for the Apsim classic examples. If you don't need Apsim Classic I can just modify the code to stay with Apsim Next Gen examples only.

@rdstern what do you think?

rdstern:
@MeSophie we have spent so long on this dialogue, that is useful, but is a detail as it doesn't actually run ApSim, etc.  So we need to finish it quickly - I suggest for now either disabling or deleting those examples.  I am assuming a new user may not need to have classic as well as nextgen on their machine.

MeSophie:
> @MeSophie we have spent so long on this dialogue, that is useful, but is a detail as it doesn't actually run ApSim, etc. So we need to finish it quickly - I suggest for now either disabling or deleting those examples. I am assuming a new user may not need to have classic as well as nextgen on their machine.

@rdstern I already deleted apims_example on this dialog.

rdstern (dismissed):
@lloyddewit and @MeSophie it will be good to merge this, to show we have started work in this area.

instat/dlgApsimx.vb
```diff
@@ -15,27 +15,116 @@
...
+    Private Sub ucrInputComboList_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputComboList.ControlValueChanged
+        Dim strDataName As String = frmMain.clsRLink.MakeValidText(ucrInputComboList.GetText)
+        ucrInputSaveData.SetText(strDataName)
+        clsApsimxExampleFunction.SetAssignTo(strDataName)
+        clsDataListFunction.ClearParameters()
+        ucrBase.clsRsyntax.ClearCodes()
+        If ucrInputComboList.GetText = "Wheat" OrElse ucrInputComboList.GetText = "Barley" OrElse ucrInputComboList.GetText = "Controlled Environment" OrElse ucrInputComboList.GetText = "Sugarcane" OrElse ucrInputComboList.GetText = "Eucalyptus" OrElse ucrInputComboList.GetText = "Eucalyptus Rotation" OrElse ucrInputComboList.GetText = "Maize" OrElse ucrInputComboList.GetText = "Oats" OrElse ucrInputComboList.GetText = "Soybean" Then
+            clsDataListFunction.AddParameter(ucrInputSaveData.GetText, clsRFunctionParameter:=clsApsimxExampleFunction, iPosition:=0)
+            ucrBase.clsRsyntax.AddToBeforeCodes(clsApsimxExampleFunction, 0)
+            ucrBase.clsRsyntax.RemoveFromAfterCodes(clsReportOperator)
```
lloyddewit, line 122: I assume that we only need to remove this operator object if it is added elsewhere.
Where is `clsReportOperator` added to the `AfterCodes`?

lloyddewit (changes requested):
@MeSophie Thank you for this.
Just a few questions/suggestions

instat/dlgApsimx.vb
```diff
@@ -15,27 +15,116 @@
...
         TestOKEnabled()
     End Sub
+
+    Private Sub ucrInputComboList_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrInputComboList.ControlValueChanged
+        Dim strDataName As String = frmMain.clsRLink.MakeValidText(ucrInputComboList.GetText)
+        ucrInputSaveData.SetText(strDataName)
+        clsApsimxExampleFunction.SetAssignTo(strDataName)
+        clsDataListFunction.ClearParameters()
+        ucrBase.clsRsyntax.ClearCodes()
+        If ucrInputComboList.GetText = "Wheat" OrElse ucrInputComboList.GetText = "Barley" OrElse ucrInputComboList.GetText = "Controlled Environment" OrElse ucrInputComboList.GetText = "Sugarcane" OrElse ucrInputComboList.GetText = "Eucalyptus" OrElse ucrInputComboList.GetText = "Eucalyptus Rotation" OrElse ucrInputComboList.GetText = "Maize" OrElse ucrInputComboList.GetText = "Oats" OrElse ucrInputComboList.GetText = "Soybean" Then
```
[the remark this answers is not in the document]
MeSophie: @lloyddewit no we don't need to check all this. I just forgot to remove it. At the beginning there were two functions on this code since I have removed one we don't need this condition anymore.

instat/dlgApsimx.vb
MeSophie: @lloyddewit the operator is just useful for **Rotation mode** because  the result of this model is a list of two data frame and we just need one of them the **Report** so the operator help to select the report.

instat/dlgApsimx.vb
```diff
@@ -116,14 +112,12 @@ Public Class dlgApsimx
...
-            clsDataListFunction.AddParameter(ucrInputSaveData.GetText, clsRFunctionParameter:=clsApsimxExampleFunction, iPosition:=0)
-            ucrBase.clsRsyntax.AddToBeforeCodes(clsApsimxExampleFunction, 0)
-            ucrBase.clsRsyntax.RemoveFromAfterCodes(clsReportOperator)
-        ElseIf ucrInputComboList.GetText = "Rotation" Then
+        If ucrInputComboList.GetText = "Rotation" Then
             clsDataListFunction.AddParameter(strDataName, clsRFunctionParameter:=clsApsimxExampleFunction, iPosition:=0)
             ucrBase.clsRsyntax.AddToBeforeCodes(clsApsimxExampleFunction, 0)
             ucrBase.clsRsyntax.AddToBeforeCodes(clsReportOperator, 1)
+        Else
+            clsDataListFunction.AddParameter(ucrInputSaveData.GetText, clsRFunctionParameter:=clsApsimxExampleFunction, iPosition:=0)
```
lloyddewit, line 120: On lines 117 and 118, we add `clsApsimxExampleFunction` and `clsReportOperator` to the 'before codes'.
So should we remove `clsApsimxExampleFunction` and `clsReportOperator` from the 'before codes' in this `Else` block? or doesn't it matter?
(I don't completely understand how we should use RSyntax so my question may be naive)

lloyddewit (changes requested):
@MeSophie Thank you for the changes, just one open question

instat/dlgApsimx.vb
MeSophie: @lloyddewit Removing them in the else block doesn't matter too much because the results of the other models are already the **report**. So Removing or not will not change the results at all reason why I have deleted the two lines which removing them.

lloyddewit (approved):

lloyddewit:
@rdstern there have been some small changes since your last approval. If you can retest/approve, then we can merge, thanks!

rdstern (approved):
@MeSophie I have just noticed there is no help on the details of the examples.  I looked in the apsimx package and couldn't find them there.  It would be good to have those details - I assume they are somewhere?  But that doesn't affect approving the dialogue.

Issue #7630, opened 2022-07-17
Futher improvement to the Data View Grid Paging


N-thony:
We would like now to have a facility of going to a specific page by enter the page number.
@Patowhiz as discussed in our last trip together with @ChrisMarsh82 the code of going to a specific page is ready but what is required here is to have customized `input text control` that will fit within the labels for entering the page number. we can discussed this more during the Rwanda Met workshop.
![image](https://user-images.githubusercontent.com/68591383/179395520-d4be20f7-1b79-4d9b-93f4-23b6c7e2a73a.png)

PR #7858, opened 2022-09-08
Go to specific Row page

N-thony:
Fixes (partially) #7630
@rdstern I have added a tooltip showing where to click for getting the box where you can enter the row page number. Please have a look.
image


rdstern:
I could take that method.  I think the phrase could be Click to go to a specific window.  I am thinking that we have just a window onto the data.
Then once in the dialogue it could mention Window number:  Maybe it could even be an up-down with limits 1 to (say) 59 if there are over 59000 rows.  If you change the window size to (say) 2000, then it would adjust - or do we now not allow changes?
 
This could be the same for Columns.

It is only available when needed.  There have to be more than 2 clicks on the > or >> for this option to be worthwhile.  If there are 5 pages, or less, you can get to each page in (at most) 2 clicks.  So there need to be 6 or more windows (either down or across) before the corresponding option is visible.

N-thony:
> I could take that method. I think the phrase could be Click to go to a specific window. I am thinking that we have just a window onto the data. Then once in the dialogue it could mention Window number: Maybe it could even be an up-down with limits 1 to (say) 59 if there are over 59000 rows. If you change the window size to (say) 2000, then it would adjust - or do we now not allow changes?
> 
> This could be the same for Columns.
> 
> It is only available when needed. There have to be more than 2 clicks on the > or >> for this option to be worthwhile. If there are 5 pages, or less, you can get to each page in (at most) 2 clicks. So there need to be 6 or more windows (either down or across) before the corresponding option is visible.

@rdstern please take a look.  We keep this for now and further improvement as discussed will be done later.

rdstern:
I tried with the Pakistan hh data, which has 53400 rows.

a) The value isn't remembered.  It should be.
b) It did 25 well.
c) Then I typed 60.  It said the rows were 59000 to 60000 - which they were not.  And it moved back to page 1.  Either don't allow a number higher than 54 (for these data).  Or, if you type anything higher, then go to the last page, i.e. take the number mod 54.  You can keep the number typed - so it is possible to type a high number and go to the last page.
d) I tried -10 and it gave a funny message, saying I couldn't have 0.
e) I tried typing abc and it crashed.

You may feel that adding all these checks may make it easier to have your own simple dialogue, with an up-down control.  If so, then here it would be from 1 to 54 and the initial default would be halfway, i.e. 54/2 = 27.

N-thony:
> @N-thony I tried with the Pakistan hh data, which has 53400 rows.
> 
> a) The value isn't remembered. It should be. b) It did 25 well. c) Then I typed 60. It said the rows were 59000 to 60000 - which they were not. And it moved back to page 1. Either don't allow a number higher than 54 (for these data). Or, if you type anything higher, then go to the last page, i.e. take the number mod 54. You can keep the number typed - so it is possible to type a high number and go to the last page. d) I tried -10 and it gave a funny message, saying I couldn't have 0. e) I tried typing abc and it crashed.
> 
> You may feel that adding all these checks may make it easier to have your own simple dialogue, with an up-down control. If so, then here it would be from 1 to 54 and the initial default would be halfway, i.e. 54/2 = 27.

@rdstern thanks, I will just customized a sub dialogue for this.

N-thony:
@rdstern please test

rdstern (changes requested):
It is better with this control.
a) At the start I accidentally pressed right-click.  This gave me a developer error, but then was ok.  That was just the first time.  I didn't get it again.
b) It started at 1, which is where I was.  It knew that 10 pages was the maximum.  I would like the default to be (say) 5 if that is easy.  Not important though.
c) I added another data frame which had 50 pages.  It was still stuck with a maximum of 10.  It should have known that the data frame is different?  As in b) ideally it would then change to its new default for the new data frame - so 25.  But I have no problem if that is difficult, so it remains on the previous setting anyway.  (Though remember that I could start with a larger one. )
d) I loaded a small data frame, and it still gave the popup message - though it didn't go to the dialogue.
e) Will you be adding the same control to go wider, or do you want this merged and then do another pull request for moving sideways?
e) Finally - and trivially, the spacing of the 3 buttons isn't quite right.

N-thony:
@rdstern I have made some changes regarding your comments above, it will still start by 1 and also will remember the pages according the data frame you are, also the buttons in the sub dialogue is a bit general.
For item e) I would prefer this to be merge then I will do it together with item f)

N-thony:
@rdstern this is ready for your re-test.

rdstern (dismissed):
This is nice.  I am approving, so it can be merged as soon as possible.
Then there is a second pull request to be able also to choose the "page" in terms of the columns shown, as well as the rows.  When this new pull is done perhaps consideration can be given to the following minor points:
a) At the top it says choose window number.  Then the text says page number.  Should it be window?
b) It starts at 1.  I wonder about starting roughly halfway down the list.  I was trying up to 53400, so it could start perhaps at about 27.
c) I wonder if it could give me the choice in the phrase?  In another set I have 157899 rows.  It is not immediately obvious to me that the limits are currently 1 and 158.  Perhaps this could be included.  It could perhaps become `Enter page number (1-158): `  Perhaps that information could also be included in the tooltip?  `Click to go to a specific window (1-158)`;
d) Should this feature also be added to the right-click menu at the bottom.  And perhaps there could also be an entry in the Edit menu that leads to the ability to go to any window - down or across?
e) I have now checked what happens when the number of rows is changed in the Tools > Options menu.  Fun - see below.  Is 1000 hard-wired into your code currently?  But before changing let's proceed with the horizontal paging so we make progress.  I also suggest Chris may with to disable that option anyway?
image

N-thony:
@rdstern I have solved here items a) b) c), have a look. Thanks.

rdstern (commented):
This is looking good now - and I note you have also done the columns as well as the rows - great.
a) If there are less than 1000 rows it still asks you to right-click and says you can choose between 1 and 0.  Similarly, if there are less than 100 columns it asks you to right-click but nothing happens.  I assume the simplest is just not to have the popup unless there are more than 1000 rows (or 100 columns).
b) It still gets messed up when I change (in Tools > Options) to show 2000 rows (say) rather than 1000.  But I am wondering whether - now the default is (I think) 1000 by 100, that - instead of you changing that code, we (at least temporarily) don't allow those defaults to be changed.

N-thony:
> This is looking good now - and I note you have also done the columns as well as the rows - great. a) If there are less than 1000 rows it still asks you to right-click and says you can choose between 1 and 0. Similarly, if there are less than 100 columns it asks you to right-click but nothing happens. I assume the simplest is just not to have the popup unless there are more than 1000 rows (or 100 columns). b) It still gets messed up when I change (in Tools > Options) to show 2000 rows (say) rather than 1000. But I am wondering whether - now the default is (I think) 1000 by 100, that - instead of you changing that code, we (at least temporarily) don't allow those defaults to be changed. Perhaps @ChrisMarsh82 would like to advise on this?

@rdstern how about now?

N-thony:
> 

@rdstern wait a bit

rdstern (changes requested):
It still doesn't change the columns, see below for the example I am using:
image

N-thony:
@rdstern could you re-test? I also notice that when you change (in Tools > Options) to show 2000 rows rather than 1000 it doesn't apply directly to the `data view` unless you click one of the next button.
If you agree with me, we can consider it as a bug and fix it?

rdstern (dismissed):
@N-thony great - that all seems to work now, so I am approving and hope the code is such that @lloyddewit can approve soon and merge.
One tiny detail - if you want more work is the position of the cursor when you first use the popup.    It is at the start of the field, where I suggest it would be better at the end.  But that is not enough to stop the merge

N-thony:
> @N-thony great - that all seems to work now, so I am approving and hope the code is such that @lloyddewit can approve soon and merge. One tiny detail - if you want more work is the position of the cursor when you first use the popup. It is at the start of the field, where I suggest it would be better at the end. But that is not enough to stop the merge

@lloyddewit I can make this small change in the next PR concerning the paging.

instat/sdgWindowNumber.vb
```diff
@@ -1,3 +1,64 @@
...
+'
+' You should have received a copy of the GNU General Public License 
+' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports instat.Translations
+Public Class sdgWindowNumber
+    Public iPage As Integer
+    Public iNumPage, iStart, iEnd As Integer
+
+    Public enumWINNUMBERMode As String = WINNUMBERMode.Row
```
lloyddewit, line 22: Is there a special reason for using `String`?
I understood that enum variables were normally declared like
```suggestion
    Public enumWINNUMBERMode As WINNUMBERMode = WINNUMBERMode.Row
```

instat/sdgWindowNumber.vb
```diff
@@ -1,3 +1,64 @@
...
+
+        Select Case enumWINNUMBERMode
+            Case WINNUMBERMode.Row
+                iMax = iNumPage / iMaxRows
+                iDefault = Math.Round(iEnd / iMaxRows)
+            Case WINNUMBERMode.Col
+                iMax = iNumPage / iMaxCols
+                iDefault = Math.Round(iEnd / iMaxCols)
+        End Select
+        If Not Integer.TryParse(iMax, iOutput) Then
```
lloyddewit, line 47: `iMax` is not a string. What's happening on this line?

instat/sdgWindowNumber.vb
```diff
@@ -1,3 +1,64 @@
...
+        Dim iMaxRows As Integer = frmMain.clsInstatOptions.iMaxRows
+        Dim iMaxCols As Integer = frmMain.clsInstatOptions.iMaxCols
+
+        Select Case enumWINNUMBERMode
+            Case WINNUMBERMode.Row
+                iMax = iNumPage / iMaxRows
+                iDefault = Math.Round(iEnd / iMaxRows)
+            Case WINNUMBERMode.Col
+                iMax = iNumPage / iMaxCols
+                iDefault = Math.Round(iEnd / iMaxCols)
```
lloyddewit, line 45: The rest of this sub seems to assume that we are in `Row` mode. Is that correct?

instat/ucrDataView.vb
```diff
@@ -891,4 +891,58 @@ Public Class ucrDataView
...
+
+    Private Sub lblColDisplay_Click(sender As Object, e As EventArgs) Handles lblColDisplay.Click
+        If lblColNext.Enabled OrElse lblColBack.Enabled Then
+            sdgWindowNumber.enumWINNUMBERMode = sdgWindowNumber.WINNUMBERMode.Col
+            sdgWindowNumber.iNumPage = GetCurrentDataFrameFocus().iTotalColumnCount
+            sdgWindowNumber.iStart = GetCurrentDataFrameFocus().clsVisibleDataFramePage.intStartColumn
+            sdgWindowNumber.iEnd = GetCurrentDataFrameFocus().clsVisibleDataFramePage.intEndColumn
+            sdgWindowNumber.ShowDialog()
+            Dim iPageNum As Integer = 0
+            iPageNum = sdgWindowNumber.iPage
```
lloyddewit, line 919: combine into single line?

instat/ucrDataView.vb
```diff
@@ -891,4 +891,58 @@ Public Class ucrDataView
...
+
+    Private Sub lblRowDisplay_Click(sender As Object, e As EventArgs) Handles lblRowDisplay.Click
+        If lblRowNext.Enabled OrElse lblRowBack.Enabled Then
+            sdgWindowNumber.enumWINNUMBERMode = sdgWindowNumber.WINNUMBERMode.Row
+            sdgWindowNumber.iNumPage = GetCurrentDataFrameFocus().iTotalRowCount
+            sdgWindowNumber.iStart = GetCurrentDataFrameFocus().clsVisibleDataFramePage.intStartRow
+            sdgWindowNumber.iEnd = GetCurrentDataFrameFocus().clsVisibleDataFramePage.intEndRow
+            sdgWindowNumber.ShowDialog()
+            Dim iPageNum As Integer = 0
+            iPageNum = sdgWindowNumber.iPage
```
lloyddewit, line 903: combine into single line?

instat/ucrDataView.vb
```diff
@@ -891,4 +891,58 @@ Public Class ucrDataView
...
+                GetCurrentDataFrameFocus().clsVisibleDataFramePage.GoToSpecificColumnPage(iPageNum)
+                RefreshWorksheet(_grid.CurrentWorksheet, GetCurrentDataFrameFocus())
+            End If
+        End If
+    End Sub
+
+    Private Sub lblRowDisplay_MouseHover(sender As Object, e As EventArgs) Handles lblRowDisplay.MouseHover
+        If lblRowNext.Enabled OrElse lblRowBack.Enabled Then
+            Dim iOutput, iRowMax As Integer
+            Dim iTotalPage As Double = GetCurrentDataFrameFocus().iTotalRowCount / frmMain.clsInstatOptions.iMaxRows
```
lloyddewit, line 930:
```suggestion
            Dim dTotalPage As Double = GetCurrentDataFrameFocus().iTotalRowCount / frmMain.clsInstatOptions.iMaxRows
```

instat/ucrDataView.vb
```diff
@@ -891,4 +891,58 @@ Public Class ucrDataView
...
+            If Not Integer.TryParse(iTotalPage, iOutput) Then
```
lloyddewit, line 931: what's the purpose of this `If` statement?

instat/ucrDataView.vb
```diff
@@ -891,4 +891,58 @@ Public Class ucrDataView
...
+                iRowMax = Math.Floor(iTotalPage) + 1
+            End If
+            ttGoToRowOrColPage.SetToolTip(lblRowDisplay, "Click to go to a specific window 1-" & iRowMax)
+        End If
+    End Sub
+
+    Private Sub lblColDisplay_MouseHover(sender As Object, e As EventArgs) Handles lblColDisplay.MouseHover
+        If lblColNext.Enabled OrElse lblColBack.Enabled Then
+            Dim iOutput, iColMax As Integer
+            Dim iTotalPage As Double = GetCurrentDataFrameFocus().iTotalColumnCount / frmMain.clsInstatOptions.iMaxCols
```
lloyddewit, line 941:
```suggestion
            Dim dTotalPage As Double = GetCurrentDataFrameFocus().iTotalColumnCount / frmMain.clsInstatOptions.iMaxCols
```

instat/ucrDataView.vb
```diff
@@ -891,4 +891,58 @@ Public Class ucrDataView
...
+            If Not Integer.TryParse(iTotalPage, iOutput) Then
```
lloyddewit, line 942: please see comment above

lloyddewit (changes requested):
@N-thony Thank you for the clean and readable code.
The Scintilla pacakages should also be removed from this PR.

instat/ucrDataView.vb
N-thony: So, for this case, I'm using the if statement to check if the value returned from the division has decimal or not, if it has decimal then I have 1 to reach the limit of the page. Perhaps we can have a call and demonstrate that?

N-thony:
@lloyddewit I resolved your comments, please have a look. Thanks.

instat/sdgWindowNumber.vb
```diff
@@ -1,3 +1,65 @@
...
+    Private Sub InitialiseControls()
+        Dim dMax As Double
+        Dim iDefault As Integer
+        Dim iMaxRows As Integer = frmMain.clsInstatOptions.iMaxRows
+        Dim iMaxCols As Integer = frmMain.clsInstatOptions.iMaxCols
+
+        Select Case enumWINNUMBERMode
+            Case WINNUMBERMode.Row
+                dMax = iNumPage / iMaxRows
+                iDefault = Math.Round(iEnd / iMaxRows)
```
lloyddewit, line 42: I understand that 
- `iDefault` is the default number displayed in the Nud
- `iEnd` is the last row in the current window

In that case, I don't understand this calculation. Would it be better if the default was the current window + 10 for example? Or something similar?

instat/sdgWindowNumber.vb
```diff
@@ -1,3 +1,65 @@
...
+            Case WINNUMBERMode.Col
+                dMax = iNumPage / iMaxCols
+                iDefault = Math.Round(iEnd / iMaxCols)
+        End Select
+        Dim iRowOrColMaxPages = Math.Floor(dMax)
+        If Not iRowOrColMaxPages = dMax Then
+            iRowOrColMaxPages = iRowOrColMaxPages + 1
+        End If
+        If iNumPage = iEnd Then
```
lloyddewit, line 51: are we comparing a window number to a row/col number? is that correct?

instat/sdgWindowNumber.vb
```diff
@@ -1,3 +1,65 @@
...
+            iDefault = iRowOrColMaxPages
+        ElseIf iStart = 1 Then
+            iDefault = Math.Round(iRowOrColMaxPages / 2)
+        End If
```
lloyddewit, line 55: Can we calculate `iDefault` in a single line without these special checks?

lloyddewit (changes requested):
@N-thony Thanks for the updates.
I didn't finish my comments yet, I'll continue tomorrow.
Don't forget about removing the Scintilla packages.

instat/sdgWindowNumber.vb
N-thony: Could you make it more clear?

instat/sdgWindowNumber.vb
```diff
@@ -1,3 +1,66 @@
...
+' MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+' GNU General Public License for more details.
+'
+' You should have received a copy of the GNU General Public License 
+' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports instat.Translations
+Public Class sdgWindowNumber
+    Public iPage As Integer
+    Public iTotalRowOrColum, iStartRowOrColumn, iEndRowOrColumn As Integer
```
lloyddewit, line 20: It is hard to upderstand what these data members are.
Please add an XML header for each one.

instat/sdgWindowNumber.vb
```diff
@@ -1,3 +1,66 @@
...
+                dMax = iTotalRowOrColum / iMaxRows
+                iDefaultPage = iEndRowOrColumn / iMaxRows
+            Case WINNUMBERMode.Col
+                dMax = iTotalRowOrColum / iMaxCols
+                iDefaultPage = iEndRowOrColumn / iMaxCols
+        End Select
+        Dim iRowOrColMaxPages = Math.Floor(dMax)
+        If Not iRowOrColMaxPages = dMax Then
+            iRowOrColMaxPages = iRowOrColMaxPages + 1
+        End If
```
lloyddewit, line 50: would something like this work instead?
```suggestion
        ' the maximum number of rows/columns to display
        Dim iMaxRowsCols = If(enumWINNUMBERMode = WINNUMBERMode.Row,
                frmMain.clsInstatOptions.iMaxRows, frmMain.clsInstatOptions.iMaxCols)

        ' the current page number displayed
        Dim iDefaultPage As Integer = Math.Ceiling(CDbl(iEndRowOrColumn / iMaxRowsCols))

        ' the maximum number of row/col pages 
        Dim iRowOrColMaxPages As Integer = Math.Ceiling(CDbl(iTotalRowOrColum / iMaxRowsCols))
```

instat/sdgWindowNumber.vb
```diff
@@ -1,3 +1,66 @@
...
+
+        If iStartRowOrColumn = 1 Then
+            iDefaultPage = iRowOrColMaxPages / 2
+        ElseIf iTotalRowOrColum = iEndRowOrColumn Then
+            iDefaultPage = iRowOrColMaxPages
+        End If
```
lloyddewit, line 56: I still don't understand what's happening here, please could you explain?

instat/sdgWindowNumber.vb
N-thony: Thanks, resolved too

instat/ucrDataView.vb
```diff
@@ -891,4 +891,60 @@ Public Class ucrDataView
...
+    End Sub
+
+    Private Sub lblColDisplay_MouseHover(sender As Object, e As EventArgs) Handles lblColDisplay.MouseHover
+        If lblColNext.Enabled OrElse lblColBack.Enabled Then
+            Dim dTotalPage As Double = GetCurrentDataFrameFocus().iTotalColumnCount / frmMain.clsInstatOptions.iMaxCols
+            Dim iCols = Math.Floor(dTotalPage)
+            If Not iCols = dTotalPage Then
+                iCols += 1
+            End If
+            ttGoToRowOrColPage.SetToolTip(lblColDisplay, "Click to go to a specific window 1-" & iCols)
```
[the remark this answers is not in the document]
N-thony: Done

lloyddewit:
@rdstern On 01 Oct, you commented:

> It starts at 1. I wonder about starting roughly halfway down the list. I was trying up to 53400, so it could start perhaps at about 27.

The current implementation shows the current page number in the control. The only exception is when the user is on the first page it shows the number of the half way page (27 in your example above).

This is inconsistent. Is there a risk that this looks like a bug or confuses the user? 
Should we be more consistent (e.g. always display the current page; or always display the page number half way between the current page and the last page)?

rdstern:
Humph - @lloyddewit you are right, we should be consistent.  Sorry @N-thony please could you always show the current page number - so it will be 1 (as you had before) at the start.

N-thony:
> Humph - @lloyddewit you are right, we should be consistent. Sorry @N-thony please could you always show the current page number - so it will be 1 (as you had before) at the start.

@rdstern good, could you re-test?

N-thony:
> @N-thony Thank you for the clean and readable code. The Scintilla pacakages should also be removed from this PR.

@lloyddewit could you help by removing the Scintilla pacakages? Thanks

lloyddewit:
> > Humph - @lloyddewit you are right, we should be consistent. Sorry @N-thony please could you always show the current page number - so it will be 1 (as you had before) at the start.
> 
> @rdstern good, could you re-test?

@rdstern I suggest you wait before testing. We will probably need to make a new PR to get rid of some extra packages that crept in. It's better if you test the new PR.

instat/sdgWindowNumber.vb
lloyddewit: If we have a shared XML comment, then when the developer hovers over one of the variables, then it gives a confusing message.
Would it be better to have a separate XML header for each variable?

Do we need `iStartRowOrColumn`? Could we delete it?

image

lloyddewit (changes requested):
@N-thony This is looking good. There is just one open comment (see above).
If you can resolve this, then I will make a new PR, without the Scintilla packages, for @rdstern to test.
Thanks

instat/sdgWindowNumber.vb
N-thony: Done, thanks

lloyddewit:
Replaced by PR #7924

PR #7869, opened 2022-09-09
Extended the PCA dialog

### LINKED ISSUE #6686: PCA Options subdialog, Save Tab Empty

### LINKED ISSUE #7318: Extend the PCA dialogue

Vitalis95:
Fixes #7318 
Fixes #6686 
Replaces PR #7654 due to conflicts
@rdstern @lloyddewit @lilyclements , have a look at it


rdstern:
@Vitalis95 is this the new PCA dialogue?  It looks like the old one (so just a single receiver) and gave a problem with a simple example as well.

lilyclements:
@rdstern are you sure you synced your branch? I can see the changes on my version

@Vitalis95 I get the following bug if I fill the boxes in and press OK

image

rdstern:
@Vitalis95 @lilyclements was correct I wasn't using the right branch.
I got the same error as above. 
I also tried with the diamonds data and there was too much printout.  So I chose not to get the eigenvectors and that was much better.  I suggest not having the eigenvectors could be the default.

Then I tried the other plot options and only the scree plot seems to work.  The others all failed.  

From the presentation with this package there seemed to be many plotting options.  Are these updated yet to permit all that is possible from the new system, particularly those with the added variables. 

With the new output window perhaps @Patowhiz can also advise.  We are currently limited to a single graph at a time.  Will that limit still be there, or could we get multiple graphs now?

Patowhiz:
I still need to design for multiple outputs. Essentially, we need to detect when an R object has multiple outputs when printed. Once that's possible we can trivially have the output window display them.

We can do that once we are done with the new way of processing R objects.

Vitalis95:
@rdstern, to avoid the bug, I have excluded the variables from the  selector once added to Explanatory Variable receiver to avoid using them also in Suppl.Numeric Receiver . I did wrote an email to Francois Husson the author of `FactoMineR` package to inquire if it is possible to use the same  variable(s) in active variables  as well as in quantitative supplementary variables, he responded that he doesn't  have any example where a variable already used as active could be also used as illustrative (quantitative supplementary variables). 
From the documentation it says these supplementary variables  have no influence on the principal components of the analysis but they help to interpret the dimensions of variability-they are not taken into account in the PCA computations.

lilyclements (approved):
Looks great! @rdstern this is ready for you review

rdstern (approved):
@Vitalis95 this is looking really good now and I am approving.
Later we could revisit, or you could explain - as we write the help.  I am still not sure how we plot the options with the extra variables and factors.  That can be come later though!

lloyddewit (approved):

PR #7899, opened 2022-10-02
In the Script window, improved the usability of brackets and quotes

lloyddewit:
**Brackets**
If the user types an open bracket (e.g. `[`), then R-Instat automaticaly inserts a closing bracket.
Also, if the user moves the cursor, and the caret is then next to a bracket, then it highlights the paired open/close bracket. 
If it cannot find a paired bracket, then it displays the bracket next to the caret in red (to indicate an error).
For indented brackets, it shows a vertical line at different indentation levels:

image

**Quotes**
If the user types a quote character (e.g. `"`), then R-Instat automaticaly inserts a closing quote character.
It avoids inserting closing quotes in situations such as `don't`. 
It also ensures that the caret does not remain in the center upon multiple quote insertions. For example ("|" is cursor position; '=&gt;' is output):
- insert ' => '|' 
- insert ' again => ''| 

Visual Studio and RStudio have the same behaviour.

@N-thony Please could you peer review?
Thanks!

### DIFF
--- a/instat/ucrScript.Designer.vb
+++ b/instat/ucrScript.Designer.vb
@@ -40,15 +40,21 @@ Partial Class ucrScript
     Private Sub InitializeComponent()
         Me.components = New System.ComponentModel.Container()
         Me.mnuContextScript = New System.Windows.Forms.ContextMenuStrip(Me.components)
+        Me.mnuUndo = New System.Windows.Forms.ToolStripMenuItem()
+        Me.mnuRedo = New System.Windows.Forms.ToolStripMenuItem()
+        Me.ToolStripSeparator1 = New System.Windows.Forms.ToolStripSeparator()
         Me.mnuCut = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuCopy = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuPaste = New System.Windows.Forms.ToolStripMenuItem()
+        Me.mnuSelectAll = New System.Windows.Forms.ToolStripMenuItem()
+        Me.mnuClearContents = New System.Windows.Forms.ToolStripMenuItem()
+        Me.ToolStripSeparator2 = New System.Windows.Forms.ToolStripSeparator()
         Me.mnuRunCurrentLineSelection = New System.Windows.Forms.ToolStripMenuItem()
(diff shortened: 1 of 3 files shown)


N-thony (approved):
Great, the code is very clear and understandable

rdstern (approved):
This looks great.  I have tried with a range of bits of code and hadn't appreciated the role of this feature enough before.
I include an example:
```
compute_log_ratio <- function(mpg, wt) {
  log_base <- getOption("log_base", default = exp(1)) # gets external data
  results <- log(mpg/wt, base = log_base)
  print(mean(results))                                # prints to the console
  done <<- TRUE                                       # sets external data
  results
}
```
I mention this, because it is from this new version of the book called [Tidy Modelling with R](https://www.tmwr.org/).  I don't know how much it has changed, since I looked last but version 1 is dates 30 Sept 2022 and looks really good.
In terms of R-Instat it shows how easy and logical R programs can be when using the tidyverse principles.  So - of course - the writing of the R code is central and great that R-Instat is making that process attractive and easy.

Still can't see any reason not to simply approve!

rdstern:
@lloyddewit I am trying to think ahead to ways we might be using your new script facilities and I think the Edit > Script dialogue might still be useful?
If you agree, then I wonder if it is trivial to use your new editor also in that dialogue?

lloyddewit:
> @lloyddewit I am trying to think ahead to ways we might be using your new script facilities and I think the Edit > Script dialogue might still be useful? If you agree, then I wonder if it is trivial to use your new editor also in that dialogue?

@rdstern 
I need to better understand the rationale/aim of the Edit>Script dialog. Should we keep it separate from the normal script window? Or should we combine the features into a single editor dialog?
In any case, I think that all the script R-Instat displays should be displayed consistently (same font, syntax highlighting and structure):
- Script dialog
- Output window
- Log window
- Edit/Script dialog
- message boxes, help?
- is script displayed anywhere else in R-Instat?

To achieve the above, I could create a reusable/configurable editor window that could be used in some or all of the above (text would be set to read-only for most windows/dialogs).
You also had ideas for improving the layout/tiling of the different windows.
We can discuss all the above in our next catch up!

PR #7903, opened 2022-10-06
Added automatic indentation and code folding to the Script window

lloyddewit:
**Automatic indentation**
In the Script window, if the user presses the Enter key, then automatically sets the indent of the new line. Normally the indent is set to the same indent as the last previous non-empty line. However, if the caret was between '{}' when enter was pressed, then increases the indent by 2 spaces and moves the '}' to the next line. This results in code with nested indents. Please see example below. 

**Code folding**
In the margin, displays '+' and '-' buttons for each (nested) block. If the user clicks '-' then it collapses the block. If the user clicks '+', then it expands the block.
This will be useful if/when users display longer scripts in the Script window. I added it now because it was easy to do so while working on the indentation.

@N-thony Please could you review?
Thanks!

image

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -160,6 +160,60 @@ Public Class ucrScript
         End If
     End Sub
 
+    '''--------------------------------------------------------------------------------------------
+    ''' <summary>   Automatically sets the indent of a new line.<para>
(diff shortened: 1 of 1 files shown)


N-thony (approved):
Looks good :)

rdstern (approved):
Looks very nice.

PR #7909, opened 2022-10-12
Added create installer to workflows

ChrisMarsh82:
This updates the current installer to allow it to be done through github workflows.

There will still need to be some checks to be done once committed to the master branch, ensuring the workflow runs without any issues.

There is some tidying up to be done once workflow is refined:

- [ ] remove old installer files
- [ ] create installer for Linux
- [ ] update R install as currently brought in from ChrisMarsh82/R-RInstat
- [ ] improve how we install packages. currently, manual processes to create a list of packages
- [ ] Signing done manually
- [ ] Uploading to website done manually


instat/clsRLink.vb
```diff
@@ -218,7 +218,7 @@ Public Class RLink
         Dim iCurrentCallType As Integer
         Dim bClose As Boolean = False
         Dim strStaticPath = Path.GetFullPath("static")
-        Dim rHome = Path.Combine(strStaticPath, "R-" & strRBundledVersion)
+        Dim rHome = Path.Combine(strStaticPath, "R")
```
lloyddewit, line 221: I understood that there is a requirement that R-Instat shall use the R bundled with it, rather than the default R installed.
Did I misunderstand? Or is this requirement changed?

lloyddewit (commented):
@ChrisMarsh82 Thanks, just one open question

instat/frmMain.Designer.vb
```diff
@@ -5124,7 +5124,7 @@ Partial Class frmMain
         Me.IsMdiContainer = True
         Me.MainMenuStrip = Me.mnuBar
         Me.Name = "frmMain"
-        Me.Text = "R-Instat"
+        Me.Text = "R-Instat " + My.Application.Info.Version.ToString
```
lloyddewit, line 5127: Was this change made manually using the editor?
If so, then I think that the next time somebody makes a change to `frmMain` using the Designer, then this line will be automatically changed to `Me.Text = "R-Instat "` and the version information will be lost.
Could we set the string in `frmMain.vb` instead?

lloyddewit (commented):
one more comment

instat/clsRLink.vb
ChrisMarsh82: You haven't miss understood. The only thing I am changing here is the folder name. We are now calling the folder just 'R' instead of 'R-3.1' This simplifies the installation without loosing any functionality as we check the version through an R script and not the folder name

instat/frmMain.Designer.vb
ChrisMarsh82: When the build was created one of the steps was to add the version here. It will only reset it if they haven't got the latest before they make there change. I can make sure I keep an eye on this when doing the build.

lloyddewit (approved):
@ChrisMarsh82 Thanks for the feedback, happy to approve

lloyddewit:
@rdstern I don't think you can usefully test this PR, so I proceeded with the merge, I hope that's OK.

PR #7915, opened 2022-10-16
Fixed oddities caused by the translation when re-opening Enter and Calculator dialogues

### LINKED ISSUE #7914: Odd feature when reopening a control with an expression
Here is the curiosity with the Prepare > Numeric > Enter dialogue.  It is the same with the Prepare > Calculator.

image

I typed the expression `"Bachelor's Degree"`.  This works fine.  But press Close and then reopen the dialogue and the quote is repeated twice.  If you do this again (and possibly again you can get to `"Bachelor''''''''''''''''s Degree"`

I noticed this feature with the Prepare > Calculator when using the Wakefield keyboard and Education key. 

I wonder whether it is just this symbol or it could go wrong with others too?  

There may be a reason for this feature, so we have to live with it.  I don't see that as a major problem if it can be explained why, instead of fixing.

I wonder if @N-thony could look quickly.  It may be something needing someone else, so don't spend long before asking.

N-thony:
Fixes #7914
@rdstern @lloyddewit this fixed the oddities reported in the corresponding issue caused by the translation system trying to translate text in the `receivers` on `Enter` and `Calculator` dialogues
@lloyddewit this is the changes I made to fix that issue
image
image


rdstern (approved):
@N-thony that seems fine.  I am a little annoyed in myself that I didn't think it could be related to our translation work.  But delighted that I suggested you could fix it - though didn't expect you to find it to be so easy!
@lloyddewit over to you to check and merge.

N-thony:
@lloyddewit this PR contains database changes, could you peer review it to avoid conflicts? Thanks.

lloyddewit (approved):

Issue #7643, opened 2022-07-22
Correction and improvements to Select facility


rdstern:
This is an example of the current situation:
![image](https://user-images.githubusercontent.com/11226469/180415141-158b2266-1988-4b16-a8a5-90fd71d23ec0.png)

a) At the bottom it always seems to say Columns 1 to 100 of 100 (329).  As shown in the filter, the second 100 should be the number of selected variables - it always seems to be fixed at 100.  Look at the filter to see that it should be the number of variables selected out of the 329.

b) We will increasingly be using it to manage wide (many variable) data sets.  I suggest the default should therefore be to apply the select, also to the column metadata.  And hence the left-hand column in the metadata should also be coloured red.   Possibly we also allow an option where the select is only applied to the data, but that could come later.
c) When defining a select @shadrackkibet has included a Where option and one possibility there is to include all Labelled variables.  Now the labelled data are of two types:
1) A factor-type variable in which case every numeric value has a corresponding label.
2) A numeric variable where one (or a few) values are labelled.  (For example age is between 1 and 89, with 95 being unknown, and 99 being refused.)  
I would like extra options to select there 2 situations separately, in addition to being able to select all labelled variables - which is possible now.  This could perhaps be given as `Labelled completely` and `Labelled incompletely`.  Perhaps we don't need both, because we do have the Not button?  @shadrackkibet may be able to advise.  Then I wonder if this could be done by @N-thony or he may wish to allocate it?

N-thony:
@rdstern could you check again item a)?

lilyclements:
a) Is no longer an issue since we put this differently now (with a colon)
b) is solved in #7919

PR #7918, opened 2022-10-17
Improvement of the Column Metadata grid

N-thony:
Fixes (partially) #7643 
@rdstern please have a look. 
This set the color to red of the selected columns in the Column Metadata grid
image


N-thony:
@lloyddewit I'm closing this since it contains files not corresponding to this.

rdstern (commented):
@N-thony I like this.
In the future I would still like an option in the Select facility to be able to hide the not-selected variables.  I agree with @shadrackkibet that what we have now is a sensible default, i.e. that the selected data are hidden in the grid and also in the data selector, but not in the column metadata.  However sometimes it will be simpler if they are hidden there too.  And I think that sometimes we may want to be able to use a select, though not hide it in the grid.  That will need more discussion, but I see it as a facility to be able to repeat calculations for selected variables, rather than simply to restrict the visibility.  But that can come later - I hope not too much later.

PR #7921, opened 2022-10-18
Add Text in Json files

### LINKED ISSUE #7773: Translate the comments and improve the translation database
We used the French version last week - for the first time in anger!  
a) It went well in that there were no instances where the fact we used French seemed to be the cause of a specific problem
b) But it showed that there is a lot to add to the database, so the translation is more complete.
There were some on the DRC workshop who were relative beginners in computing, and in English.  A more complete translation would have helped them.

We have the 2 components, namely to make the database more complete and also check that the visible text is clear on the dialogues.  I would like to concentrate primarily - at least for the next 2 weeks - on the database.

In adding to the database please also be able to subtract items if they are not used?  Or is there a way this can be checked periodically?

And when there are changes in dialogues that will affect the translation database, can we add a suitable label to the issue, or pull request - I think perhaps the issue, and then change the pull request to "partially".  Then we can regularly check on the issues with that label.  @lloyddewit what do you think?

image

a) On the main menu it is just the new item `Experiments` that is not translated.
b) The Prepare menu, seems all to be translated.  Good.
c) The sub-menu shown has 4 items that need translation.  Two of the items may be new (Select Columns and Add (Merge) Columns, and the other two (Rename Columns and Duplicate Column are probably very minor edits of what was there before?)

Here is an example of a dialogue:

image

d) One recent addition in the dialogue needs to be translated.
But, I am also interested in 2 more general points:
e) The dialogue name at the top is not translated, even though the corresponding menu item is translated.
f) And the comment is all still in English.

I am particularly keen on the comment being translated into the new language.  I hope this is easy.  There are two parts to the comment.  The first is shown below from Tools > Options
image

The second is the name of the dialogue - which should pick up the translated name.

I hope this item (f) might be easy - perhaps done by @lloyddewit?

MeSophie:
Fixes (partially) #7773.
Fixes (partially) #7749.

@N-thony please have a look. I think this will close #7773 after translating Experiment in Crowdin .


N-thony:
@MeSophie good, you have just added one text, do you have other texts you would like to add before merging this?

MeSophie:
@N-thony  Please wait a little I'm still working through some translation issues. I will let you know when I will done.

MeSophie:
@N-thony I add some text I think you can merge now such that @Vitalis95 can procede.

lloyddewit (approved):

Issue #7777, opened 2022-08-15
Simplify the Prepare > Data Frame > Add (Merge) Data dialogue when possible


rdstern:
Here is an example to show what I would like.

`File > New Data Frame > Lists > Maths > Primes`
Then
`File > New Data Frame > Lists > Maths > Binary Primes `
Now add a row variable to each data frame - making it a key variable (uses Prepare >Data Frame > Row Numbers/Names dialogue.
Now I would like to add the binary primes variable to the primes data frame.

![image](https://user-images.githubusercontent.com/11226469/184616292-4cb0794f-810f-49f6-9d2d-bec20f9ab745.png)

Notice all variables are there, and I have to use the sub-dialogue to get rid of them all except the row variable.  (That's annoyingly complicated.)

I suggest that if the `From` data frame has a single key variable and the other data frame has a variable of the same name (here the name `row`), then that's the only linking variable that is suggested.  And that's what is needed here.  
I assume @N-thony for this, or he may propose someone else.

N-thony:
@derekagorhom can you try this? Let me know if you have any problem.

derekagorhom:
Hello Antoine, please i will need you help in fixing issue #7777, first is with the ucrFromDataFrame, when changing the setdatatype its now does not display any of the data types for the dataset.

N-thony:
@rdstern on item b) I think it is because we are using `data_book$add_columns_to_data(data_name="primes", col_data=primes)` and this is adding the column in the primes dataset, it overwrites if it find the same columns.

![image](https://user-images.githubusercontent.com/68591383/197731482-36e4755a-a316-47e2-8352-37a1391a7e02.png)

N-thony:
> Here is an example to show what I would like.
> 
> `File > New Data Frame > Lists > Maths > Primes` Then `File > New Data Frame > Lists > Maths > Binary Primes ` Now add a row variable to each data frame - making it a key variable (uses Prepare >Data Frame > Row Numbers/Names dialogue. Now I would like to add the binary primes variable to the primes data frame.
> 
> ![image](https://user-images.githubusercontent.com/11226469/184616292-4cb0794f-810f-49f6-9d2d-bec20f9ab745.png)
> 
> a) Notice all variables are there, and I have to use the sub-dialogue to get rid of them all except the row variable. (That's annoyingly complicated.)
> 
[...]

PR #7933, opened 2022-10-25
Fixed small bugs in Add Prepare > Data Frame > Add (Merge) Data dialog

N-thony:
Fixes (partially) #7777
@lloyddewit @rdstern this fixes part b) of the corresponding issue

### DIFF
--- a/instat/dlgMergeAdditionalData.vb
+++ b/instat/dlgMergeAdditionalData.vb
@@ -20,7 +20,7 @@ Imports RDotNet
 Public Class dlgMergeAdditionalData
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
-    Private clsInsertColumnFunction, clsGetColumnsFromData As New RFunction
+    Private clsInsertColumnFunction, clsGetColumnsFromData, clsListFunction, clsImportDataFunction As New RFunction
     Private lstJoinColumns As New List(Of String)
     Private clsLeftJoinFunction As New RFunction
     Private clsByListFunction As New RFunction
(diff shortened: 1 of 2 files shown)


rdstern:
@derekagorhom and @N-thony am I correct in waiting for item a) to be fixed, before checking?

N-thony:
> @derekagorhom and @N-thony am I correct in waiting for item a) to be fixed, before checking?

@rdstern I'm sharing this with @derekagorhom , for item b) this is ready for your test. Please test

rdstern (commented):
@N-thony thanks and the bug is fixed.  But it is still giving an extra copy of the list variable from the x.  Perhaps that is a feature?  here is what I got.
image

It has correctly passed the binary data, and then also added another copy of the primes, which it has called list.x!

These are minor inconveniences so you may decide to abandon the problem for now?

N-thony:
> @N-thony thanks and the bug is fixed. But it is still giving an extra copy of the list variable from the x. Perhaps that is a feature? here is what I got. image
> 
> It has correctly passed the binary data, and then also added another copy of the primes, which it has called list.x!
> 
> These are minor inconveniences so you may decide to abandon the problem for now?

@rdstern I suspect that you have done the operation twice i.e returning to the dialogue and clicking `Ok` again, it is only in that case you will have an additional `list` column otherwise you will only have `list.x` and `list.y`

rdstern:
@N-thony I don't think so.  list is there from the start, and it seems another copy has been made into list.x, while from the other data frame it has correctly copied list.y.  But I will check.

rdstern (approved):
@N-thony I suggest there are more important improvements than this one for now.  
So I am approving.  I am happy that the issue is just partially done.

lloyddewit (approved):

PR #7719, opened 2022-08-03
Added missing package

N-thony:
Fixes (partially) #5862
@rdstern @ChrisMarsh82 @lloyddewit I have added the missing packages as suggested in the corresponding.
@ChrisMarsh82 can you have a look? Thanks

PR #7934, opened 2022-10-26
Added missing packages

N-thony:
@lloyddewit this PR replaces PR #7719 and also adds the package needed in PR #7718.


lloyddewit:
@ChrisMarsh82 In PR #7557 , you added new packages and also updated the version numbers. When I look at your changes, it looks as if `versions` is a list with the same length as `packs` and corresponds one-to-one. If this is correct, then how to we guarantee that the lists stay aligned?

image

ChrisMarsh82:
> @ChrisMarsh82 In PR #7557 , you added new packages and also updated the version numbers. When I look at your changes, it looks as if `versions` is a list with the same length as `packs` and corresponds one-to-one. If this is correct, then how to we guarantee that the lists stay aligned?
> 
> image

There is no easy way to check. The R script that gets the packages also gets the versions so as long as someone copies and pastes them correctly they should be no issues.

lloyddewit:
> There is no easy way to check. The R script that gets the packages also gets the versions so as long as someone copies and pastes them correctly they should be no issues.

@ChrisMarsh82 Thank you for the clarification. The purpose of this PR is to add the `coin` and `pingr` packages to R. @N-thony followed the instructions in the wiki and added these packages manually to the `packs` list. I suspect that this has made the `packs` and `versions` lists inconsistent. Should we have added the packages using an R script rather than manually?
How did you add the packages in PR #7557? (we can go through this verbally next time we talk if it's easier and maybe involve @N-thony also)
Thanks for your help

PR #7935, opened 2022-10-26
Fixed error message in Script window

### LINKED ISSUE #7931: Minor bug in Script Window message
@Patowhiz can you assign this to who can fix it? It is a very minor one that I came across! But in ucrScript, we currently have

https://github.com/africanmathsinitiative/R-Instat/blob/ab00c9ea6ecfbb9c99bdd94812bc674307130212/instat/ucrScript.vb#L383

```
    Private Sub mnuPaste_Click(sender As Object, e As EventArgs) Handles mnuPaste.Click
        If Clipboard.ContainsData(DataFormats.Text) Then
            txtScript.Paste()
            EnableDisableButtons()
        Else
            MsgBox("You can only paste text data on the script window.", "Paste to Script Window")
        End If
    End Sub
```

If I try and paste something in the script window and I get the error message, it crashes. I think the issue is that we do not want to use `MsgBox` here, but instead have a different function. I'm not sure which function though. Happy to fix it if you can tell me the function, otherwise, do you know who can?

lilyclements:
Fixes #7931 

If I tried to paste an image (e.g. jpeg) into the script window, it would try to bring up the error message that this cannot be performed.
However, instead of bringing up the error message, it would crash.
This fixes that bug

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -380,7 +380,7 @@ Public Class ucrScript
             txtScript.Paste()
             EnableDisableButtons()
         Else
-            MsgBox("You can only paste text data on the script window.", "Paste to Script Window")
+            MsgBox(Prompt:="You can only paste text data on the script window.", Title:="Paste to Script Window")
         End If
     End Sub
 
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -380,7 +380,7 @@ Public Class ucrScript
             txtScript.Paste()
             EnableDisableButtons()
         Else
-            MsgBox(Prompt:="You can only paste text data on the script window.", Title:="Paste to Script Window")
+            MsgBox("You can only paste text data on the script window.", MsgBoxStyle.Exclamation, "Paste to Script Window")
         End If
     End Sub
 


lloyddewit:
@rdstern Please can you test? thanks

instat/ucrScript.vb
```diff
@@ -380,7 +380,7 @@ Public Class ucrScript
             txtScript.Paste()
             EnableDisableButtons()
         Else
-            MsgBox("You can only paste text data on the script window.", "Paste to Script Window")
+            MsgBox(Prompt:="You can only paste text data on the script window.", Title:="Paste to Script Window")
```
lloyddewit, line 383: The line above is perfectly correct 
I searched to see how `MsgBox` is used elsewhere in R-Instat and, when a title is specified, the type of message box is also specified. Should we be consistent and do it in the same way?
```suggestion
            MsgBox("You can only paste text data on the script window.", MsgBoxStyle.Exclamation, "Paste to Script Window")
```

lloyddewit (changes requested):

instat/ucrScript.vb
lilyclements: I was getting a crash before and by putting in the parameter names I'm no longer getting the crash - I assume because it knows that we want "Paste to Script Window" to be the title now? If this also fixes the crash then great!

lloyddewit:
@lilyclements Thank you for finding this bug.
I was able to recreate the error and your fix is correct. I took the liberty of changing it to be more consistent with the rest of the code. I don't think we need @rdstern 's test for this, so I'll go ahead and merge. I hope that's OK.

Patowhiz:
@lloyddewit how were you able to recreate the error?
It's interesting because the script window is designed to disable pasting if clipboard content is not text.

lloyddewit:
> @lloyddewit how were you able to recreate the error? It's interesting because the script window is designed to disable pasting if clipboard content is not text.

I tested by:
- I opened the script window, typed in some text, then cut/pasted it back again. This worked as expected.
- I made a screen shot (\<windows key\>\<shift\>S) and pasted this into the script window. Before @lilyclements 's change, this raised an exception. The exception was raised because line 383 attempted to pass a string to an integer parameter.
- After @lilyclements 's fix, it displayed a warning message box (the correct and expected behaviour)

Patowhiz:
@lloyddewit agreed that Message box should have been fixed either way.
Did you reproduce this in windows 11? My windows 10 behaves differently. The message box never pops up.

lloyddewit:
Interesting that you (and probably others) experience different behaviour. I have 64-bit Windows 10.

```
Edition	Windows 10 Home
Version	21H2
Installed on	‎01/‎10/‎2020
OS build	19044.2130
Serial number	PF11RZGQ
Experience	Windows Feature Experience Pack 120.2212.4180.0
```

Patowhiz:
@lloyddewit many thanks for your responses.

Maybe it's the windows version difference. 
Below is the windows version I tried to test this on. Still 64 bit.
```
Edition	Windows 10 Pro
Version	21H1
Installed on	‎07/‎12/‎2019
OS build	19043.1889
Experience	Windows Feature Experience Pack 120.2212.4180.0
```
Will test on my windows 11 machine and check on  its behavior.

The reason as to why these minor development stage details are important is because in production environment, they are at times the cause of bigger issues. During trainings we have encountered same windows machines behaving oddly different and at times we have to force users share 1 machine. @rdstern can testify how that at times is a bit frustrating (first session and already problems). Seems like in certain scenarios .Net behaves slightly different in different windows build versions.

lilyclements:
@Patowhiz @lloyddewit just if you were interested - my specs are the same version as @lloyddewit (`21H2`). But otherwise some differences

```
Edition	Windows 11 Home
Version	21H2
Installed on	‎03/‎04/‎2022
OS build	22000.978
Experience	Windows Feature Experience Pack 1000.22000.978.0
```

Patowhiz:
Thanks @lilyclements.

Issue #7812, opened 2022-08-25
Add alignmment options to the grid?


rdstern:
I assume this might add to an entry in the toolbar and also possibly in the Edit menu.

With some character fields an entry can become very wide.  I have also an extreme situation with the files supplied in #7811 

The Excel file has a wide column (very wide - over 2500 characters).  I would like the Excel options of being able to look at it left-justified, (possibly centred) or right-justified.
I like the feature (also in Excel) of double-clicking at the end to make it full width.  But then it is very awkward to try and get it back to a more reasonable width.  I don't know if there is a way to improve on this.

Word wrap is in Excel, and is also in R.  It is under Prepare > Column: Text > Transform > Wrap.  It works well in that it wraps in the grid, but it doesn't unwrap!  
An example is the data in the library in the package statquotes.  I used wrap 25 and later 70.  The grid still keeps the setting of the wrapping from the smallest that was used.  It doesn't unwrap the cell depth, though the text does.
![image](https://user-images.githubusercontent.com/11226469/186607385-49865dd9-3427-456b-a6f4-43fae78892f0.png)

rdstern:
I assume alignment and word-wrap (perhaps unlike Excel) only apply to whole variables (columns).  Alignment could be for multiple variables, while word-wrap will only be for a single variable in each data frame.  I assume there will be entries in the Edit menu and perhaps an additional item, with a pull-down, in the toolbar?  I am not so keen on adding to the right-click menu for these.

I describe 2 datasets where I would like an alignment option and also a word-wrap option.
a) File > Open from Library > statquotes.  This has a variable which is up to 700 characters.  So both changing alignment and word-wrap would be useful.  After wordwrap we need to be able to unwrap.
b) File > New data frame > 1000 rows and x1 = 1,1000.  Then use the Prepare > Calculator > Integer keyboard and Factorial to give the factorial of all numbers up to 1000!.  This is huge - over 2000 places, so very wide and (unlike words) with no spaces.  So when no spaces in blocks of the width specified, then it simply word-wraps, splitting the text at the specified width.

I assume wordwrap (unlike Excel) could perhaps just be for the current variable and (for simplicity) it would  only be for 1 variable in a dataframe.  If another is chosen , then first it is turned off  if it is in wordwrap mode.  If multiple variables are selected, then the word-wrap option may be disabled.

[...]

PR #7937, opened 2022-10-27
Ensured big numbers can be displayed in Data View grid

N-thony:
Fixes (partially) #7812 
@rdstern have a look of what we have now, when I use your example by creating a big number with `2^x1-1` in the corresponding issue I get this in the grid, with `Scientific = FALSE`
image

So, here when we set `Scientific = TRUE` we get the `e-notation` where the character are more than 9, you can see from row 30
image

### DIFF
--- a/instat/static/InstatObject/R/stand_alone_functions.R
+++ b/instat/static/InstatObject/R/stand_alone_functions.R
@@ -27,11 +27,12 @@ convert_to_character_matrix <- function(data, format_decimal_places = TRUE, deci
         #  format(data[[i]], digits = decimal_places[i], scientific = is_scientific[i])
         temp_data <- c()
         for(val in data[[i]]){
-            if(nchar(val) > 9){
+            #if(nchar(val) > 9 && is_scientific[i]){
+            #   temp_data <- append(temp_data, format(val, digits = decimal_places[i], scientific = is_scientific[i]))
                temp_data <- append(temp_data, format(val, digits = decimal_places[i], scientific = is_scientific[i]))
-            }else{
-               temp_data <- append(temp_data, val)
-            }
+            #}else{
+            #   temp_data <- append(temp_data, val)
+            #}
         }
         out[, i] <- temp_data
       }
--- a/instat/static/InstatObject/R/stand_alone_functions.R
+++ b/instat/static/InstatObject/R/stand_alone_functions.R
@@ -23,8 +23,19 @@ convert_to_character_matrix <- function(data, format_decimal_places = TRUE, deci
         #which are recognised oddly by the R.Net
         out[, i] <- as.character(data[[i]])
       } else {
-        out[, i] <-
-          format(data[[i]], digits = decimal_places[i], scientific = is_scientific[i])
+        #out[, i] <-
+        #  format(data[[i]], digits = decimal_places[i], scientific = is_scientific[i])
+                #out[, i] <-
+        #  format(data[[i]], digits = decimal_places[i], scientific = is_scientific[i])
+        temp_data <- c()
+        for(val in data[[i]]){
+            if(nchar(val) > 9){
+               temp_data <- append(temp_data, format(val, digits = decimal_places[i], scientific = is_scientific[i]))
+            }else{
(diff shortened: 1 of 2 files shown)


N-thony:
@lilyclements could we have a quick chat on this?

rdstern:
@N-thony that's good that it is so easy, but I also did like the option where the whole variable was in scientific.  For teaching I suggest it is useful to have the "all numbers scientific" option we had before.  I also like the FALSE option - where we can display very large numbers in non-scientific - and that's what Excel doesn't have!

So I suggest we need 3 options.  Is it possible for there to be FALSE and TRUE as before and NA as well!  If so, then NA would be your solution.  The message would also have to change when you make a mistake.  Perha[ps there could be a tooltip on the heading `Type T or TRUE for scientific, F or FALSE for numeric and N or NA for a mixture`.

N-thony:
> @N-thony that's good that it is so easy, but I also did like the option where the whole variable was in scientific. For teaching I suggest it is useful to have the "all numbers scientific" option we had before. I also like the FALSE option - where we can display very large numbers in non-scientific - and that's what Excel doesn't have!
> 
> So I suggest we need 3 options. Is it possible for there to be FALSE and TRUE as before and NA as well! If so, then NA would be your solution. The message would also have to change when you make a mistake. Perha[ps there could be a tooltip on the heading `Type T or TRUE for scientific, F or FALSE for numeric and N or NA for a mixture`.

@rdstern sounds good

lilyclements:
@N-thony did you want a chat about this or has it been cleared up?

N-thony:
> @N-thony did you want a chat about this or has it been cleared up?

@lilyclements let me try it again what @rdstern suggested now before your review.

N-thony:
> and N or NA for a mixture

@rdstern could you re-test?
@lilyclements can you review?

rdstern (dismissed):
@N-thony looks really good

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -23,8 +23,17 @@ convert_to_character_matrix <- function(data, format_decimal_places = TRUE, deci
         #which are recognised oddly by the R.Net
         out[, i] <- as.character(data[[i]])
       } else {
-        out[, i] <-
-          format(data[[i]], digits = decimal_places[i], scientific = is_scientific[i])
+        #out[, i] <-
+        #  format(data[[i]], digits = decimal_places[i], scientific = is_scientific[i])
```
lloyddewit, line 27:
```suggestion
```

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -23,8 +23,17 @@ convert_to_character_matrix <- function(data, format_decimal_places = TRUE, deci
...
+        temp_data <- c()
+        for(val in data[[i]]){
+            if(nchar(val) > 9 && is.na(is_scientific[i])){
+               temp_data <- append(temp_data, format(val, digits = decimal_places[i], scientific = is_scientific[i]))
+            } else{
+              temp_data <- append(temp_data, format(val, digits = decimal_places[i], scientific = is_scientific[i]))
```
lloyddewit, line 33: what's the difference between lines 31 and 33?

instat/ucrColumnMetadata.vb
```diff
@@ -174,12 +174,14 @@ Public Class ucrColumnMetadata
             If strBooleanValsAllowed.Contains(newValue) Then
                 If newValue(0) = "F" Then
                     newValue = "FALSE"
-                Else
+                ElseIf newValue(0) = "T" Then
                     newValue = "TRUE"
+                ElseIf newValue(0) = "N" Then
+                    newValue = "NA"
                 End If
```
lloyddewit, line 181: replace this if-else structure with a select-case ?

lloyddewit (changes requested):

rdstern (approved):
Still seems fine.  Must remember to include this property in the help.

lloyddewit (approved):

PR #7923, opened 2022-10-19
Added Options subdialog to Two variables, Frequency/summary and Column Summaries dialogs

anastasia-mbithe:
Fixes #7917 
Added Options subdialog to Describe > Two variables dialog, Describe > Specific Tables and graphs> Frequency/summary dialog and Prepare > Data Reshape > Column Summaries dialog.

@N-thony, have a look at this and test it as well.

PR #7942, opened 2022-10-31
Updated the label from Missing to Options in column summaries dialog.

Vitalis95:
@N-thony ,  - updated the label in the database as mentioned in  PR #7923

image


lloyddewit (approved):
@N-thony Thanks for change and documentation

PR #7946, opened 2022-11-01
 In Import dialog, read a JSON file with equal lengths of lists

### LINKED ISSUE #7714: Reading a json or html or txt file into R-Instat
Here is the file
[wikivital_mathematics.zip](https://github.com/africanmathsinitiative/R-Instat/files/9240241/wikivital_mathematics.zip)

It is a json file, which should import - but it seems not to be a data frame.  So File > Import Data Frame doesn't work.  Is there something that can be done, either with this dataset - which seems interesting - or with the import dialogue, for data of this type?

This is not a very important topic, so can wait until staff have time

anastasia-mbithe:
This Fixes #7714 
@rdstern, This is in progress.
So far the code can import a JSON file with equal lengths of lists. Am working towards being able to import JSON files with unequal lengths of lists.
I tried importing a simple txt file and the current options managed to import it to R-Instat, though I haven't tried with a bigger txt file!

### DIFF
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -7,7 +7,7 @@ Public Class dlgImportDataset
 
     Private clsImportTextFileFormats, clsImportCSVFileFormats, clsImportRDS, clsReadRDS, clsImportExcel, clsImport, clsImportfromJSON As New RFunction
     Private clsGetExcelSheetNames As New RFunction
-    Private clsJsonDataFunction As New RFunction
+    Private clsJsonDataFrameFunction As New RFunction
     Private clsRangeOperator As New ROperator
     ''' <summary>   
     ''' Ensures that any file paths containing special characters (e.g. accents) are 
@@ -297,7 +297,7 @@ Public Class dlgImportDataset
         clsImport = New RFunction
         clsReadRDS = New RFunction
         clsImportfromJSON = New RFunction
-        clsJsonDataFunction = New RFunction
+        clsJsonDataFrameFunction = New RFunction
         clsGetExcelSheetNames = New RFunction
         clsRangeOperator = New ROperator
         clsEnc2Native = New RFunction
@@ -335,6 +335,9 @@ Public Class dlgImportDataset
 
         clsImportfromJSON.SetPackageName("jsonlite")
         clsImportfromJSON.SetRCommand("fromJSON")
+
+        clsJsonDataFrameFunction.SetRCommand("as.data.frame")
+        clsJsonDataFrameFunction.AddParameter("x", clsRFunctionParameter:=clsImportfromJSON, iPosition:=0)
 
         'This R command ensures that any file paths containing special characters (e.g. accents) 
         'are correctly encoded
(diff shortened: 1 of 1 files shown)


rdstern:
@anastasia-mbithe Itried the wikivatal file again, but it still says I can't import it.

anastasia-mbithe:
> @anastasia-mbithe I tried the wikivatal file again, but it still says I can't import it.

Yeah, its because the wikivital file has unequal lengths of lists, and that's still to be implemented.

If you try importing a JSON file, then go to script and change the txt parameter from the file to maybe this URL (https://raw.githubusercontent.com/datacarpentry/r-socialsci/main/data/SAFI.json ) or a simple JSON file, it works because of the equal lists that can be converted to a data frame.

anastasia-mbithe:
[example of json.zip](https://github.com/africanmathsinitiative/R-Instat/files/9978734/example.of.json.zip)

@rdstern, @lilyclements and @Patowhiz 

Above is a simple example of a JSON file that can be imported using the current function in this PR.
I also have data from an URL, https://raw.githubusercontent.com/datacarpentry/r-socialsci/main/data/SAFI.json however it can only be imported via the script window since we don't have that option currently(Maybe it can be implemented in the future).

anastasia-mbithe:
@rdstern, just a small correction, it's not the complex datasets that we can import through the script window now, it's the files directly from a website( using a URL). So it's a point of discussion whether the import from file dialog (in the future) can have a feature/ input textbox to allow a user to input a website URL. Like the one I provided in my earlier comment.

anastasia-mbithe:
Oooh, I think I get it now. So this script imported the dataset in R, but failed to import it in R-Instat;

data1<- jsonlite::fromJSON(txt = "D:/Datasets/wikivital_mathematics.json", simplifyVector = FALSE, flatten=FALSE)
dat <-data.table::rbindlist(lapply(data1, function(x) {
  as.list(unlist(x))
}), fill=TRUE)
View(dat)

rdstern:
@anastasia-mbithe thank you for the clarification.  I am happy to separate the 2 aspects, and @Patowhiz has been discussing earlier, the possibility of direct import from a url.  I would like that - and through either the main import dialogue or another dialogue.  I hope we can do better than having to depend on a script.
Separately from that I was under the impression that we can now import some json files from the dialogue - which I just did - but not more complicated ones.  Will it be possible to import those through a script, even if it has to be adapted for each complicated json file?  If so, then I would like to see an example of a script for an example of a json file that can't be imported through the dialogue.
and by the way, I looked at the json file you supplied on the url.  Is that one that will import if I download, or is it an example of a complicated one?

anastasia-mbithe:
@rdstern , the JSON file I supplied on the URL can be downloaded/ imported with the current function in the dialog through the script window. So, you try importing the simple one, then send the codes to the script and change the path to the URL.

lloyddewit (approved):

lloyddewit:
@anastasia-mbithe Thank you for the changes. I see that there is still some discussion with @rdstern  above about this functionality. I'll go ahead and merge this PR with the understanding that the remaining points above will be implemented in a separate PR.
Thanks!

Patowhiz:
@anastasia-mbithe can your explore basic import from a url as outlined in issue #6196. If it takes you more than a day to make any progress, even a write up, leave it. Initially you could design it to only support json. I'm not sure if we may need another dialog for this, you finding will dictate our design.

PR #7949, opened 2022-11-03
Added more keys to the Prepare > Calculator >Transform keyboard.

### LINKED ISSUE #7944: Add 2 keys and change help button in Prepare > Calculator > Transform keyboard
@anastasia-mbithe there are a number of issues coming still on the calculator.  Here is the first - the transform keyboard.

Here is the current Transform keyboard:
image

a) In the second and third rows I hope there could be a change in the width, so there can be 5 keys in each.  After cumsum, could there also be cumprod.  I don't know why I omitted it.  It is there in the R base package!
b) So there should be the same for the mov, which uses the rollapply button and the function prod.  The existing moving minimum does the same.
c) It would then be good to swap the movmed and movmin keys so the cum row and the mov rows are equivalent.
d) Change the code for the movmin as follows: zoo::rollapply( c(2,3,5,7), width = 3 , FUN = min, align="center", fill = NA).  So I have put the 3 in the width, that is currently blank.  The k=3 isn't needed here and it would be good to have the align.  Have the same code for the prod function in movprod.
e) The tooltip could be : cumulative products.  For example cumprod(c(2,3,5,7)) =(2,6,30,210)  Also  moving products.  For example movprod(c(2,3,5,7,11))  = (NA, 30, 105, 385, NA)  
f) In the diff function could you add the default arguments always:  So c(NA, diff(x=x1 , lag = 1, differences =1)).  It needs lag and differences to be set to 1 for the single NA to work.
g) After diff add another new key called `reverse`. It runs the simple rev command, from R base.  The tooltip could be Reverse a variable.  For example rev(c(1,2,3,4,5)) = (5,4,3,2,1).
h) Have the new Help button.  And the default, and so the first package can be `dplyr` if you wish.  The other packages are `base` and `zoo`.

anastasia-mbithe:
Fixes #7944

@rdstern and @lloyddewit, This is ready for testing.


rdstern (commented):
@anastasia-mbithe That's great.  Really well and carefully done.  You are working now at a very high standard.
The only minor point, I would prefer the R Help button to not specify the base package - though that's what it would do - as now.  When you press the pull down it shows all 3 packages, with base first.  So you could get the base help either by just pressing the button, or by pressing the pull-down and then deciding to choose the first option.

Then I will be happy to approve.

anastasia-mbithe:
Am creating another PR with changes similar in this branch. solving the conflicts, resulted in more issues in the branch.

anastasia-mbithe:
I am closing this PR since I have replaced it with #7954.

PR #7951, opened 2022-11-03
Made small aesthetic fixes to Line Plot dialog

MeSophie:
I make some modifications according to @rdstern last comment in PR#7762.
@N-thony please have a look.


rdstern (commented):
@MeSophie thanks for this.  I can't tell really until it is merged.
I just take the opportunity to ask another question and at the same time suggest a very minor change that has been annoying me for a long time - but it has been too small to make a special issue.

So under the fitted line there is the option to "Add peaks and Valleys"  In French peaks is translated as pics.  I assume that is ok as the translation?
Now the very small change - the checkbox for valleys is too high!  Could you move it down so it is exactly lined up with the checkbox for the peaks (pics!).
Thanks

N-thony:
> @MeSophie thanks for this. I can't tell really until it is merged. I just take the opportunity to ask another question and at the same time suggest a very minor change that has been annoying me for a long time - but it has been too small to make a special issue.
> 
> So under the fitted line there is the option to "Add peaks and Valleys" In French peaks is translated as pics. I assume that is ok as the translation? Now the very small change - the checkbox for valleys is too high! Could you move it down so it is exactly lined up with the checkbox for the peaks (pics!). Thanks

@MeSophie can you make the changes and solve the conflicts?

rdstern (approved):
@lloyddewit these are small changes and I am approving.
I note that Derek is also working on the same dialogue.  Let's merge this quickly, so he can proceed with more major edits in #8017 .

lloyddewit (approved):

PR #7954, opened 2022-11-07
Added more keys to the Prepare > Calculator > Transform Keyboard

### LINKED ISSUE #7876: Even more for the Integer keyboard!
@anastasia-mbithe This is looking impressive already.  This is for after the existing pull request has been included.

First 4 minor points:
a) The new-style Help is excellent on the integer keyboard.  But it misses the `gmp` package, at least visibly.  It goes to gmp from the main button, but should also include it as the first in the drop-down list.   Could you please add that.  At the same time - see below, there will be a new section that uses the Zseq package.  Could you add that to the Help too.
b) I have had problems with the hex and oct keys.  This is in data that can't go into hexmode or oxmode.  So could you add 2 more keys, called octmode and hexmode.  They should have the code you currently have in octal and hex.  Then keep the octal and hex keys, but delete the as.octmode and as.hexmode from them.  Then it will give strings for those, just like binary.  Use the same tooltips for the new keys, with one showing the as.hexmode part of the key - as now, and the other omitting that part.

c) Change the tooltip for the bin key slightly, deleting the as.integer from the text.
d) There is also an as.roman function in the R utils package.  (I think R utils is not the same as R.utils?).  So add this key.  This will mean that another row will have 4 keys. The tooltip could be "Converts a small (up to 3899) positive integer to Roman numerals.  For example as.roman(14) = XIV.  

Now could you also arrange for the Zseq package to be added.  This has been improved recently, after I contacted the author.  The commands here are all added in the same way.  It is again up to the length of the data frame, so the nrow bit needs to be included.

Then I am adding proposed 12 keys - so a new section of 3 rows, below the primes - possibly as follows:
1. `palindrome`.  This key generates: `Zseq::Palindromic(n=nrow(x=(data)),gmp=FALSE)`, for a data frame called data. 
If there are 100 rows, this generates the first 100 palindromic numbers.
Tooltip "Generate palindromes to the length of the dataframe.  For example the 100th palindromic number is 909"
2. `fibonnaci2`. Tooltip: "Generates fibonacci numbers to the length of the data frame.  For example the 10th fibonacci number is just 34, but the 50th is much larger, 7778742049". 
The key may have to be: `Zseq::Fibonacci(n=nrow(x=(data)),gmp=TRUE)`.  I think the default is FALSE, though the documentation says otherwise.
3. `abundant`.  Key gives `Zseq::Abundant(n=nrow(x=(data)),gmp=FALSE)`. Tooltip: When divisors sum to more than the number.  So 12 is less than 1+2+3+4+6 = 16.  Opposite is Deficient.
4. `perfect`.  It takes ages if n>10 so key here gives rep(Zseq::Perfect(10),length=nrow(x=(data))) Tooltip: Divisors add to the number.  So 6 = 1+2+3.  Only 51 perfect numbers are currently known.
Now to the second row:
5. `frugal`.  Key gives `Zseq::Frugal(n=nrow(x=(data)),gmp=FALSE)`  125 is frugal because 125 = 5(cubed) and 5 and 3 are fewer digits (2) than 125 (3).
6. `powerful`. Key gives `Zseq::Powerful(n=nrow(x=(data)),gmp=FALSE)`.  36 = 2(squared)*3(squared) is powerful because for each divisor, here 2 and 3, its square is also a divisor.  
7.  `ugly`.  Tooltip "Also called Regular or Hamming or 5-smooth numbers.  Key gives `rep(Zseq::Regular(100),length=nrow(x=(data)),gmp=FALSE)`.  Tooltip is: Numbers for which the factors are only 2, or 3 or 5. 
8. `happy` ends the second row. Key gives `Zseq::Happy(n=nrow(x=(data)),gmp=FALSE)`.  Tooltip: 13 is happy because 1(squared) + 3(squared) = 10, then 1(squared)+ 0(squared) = 1. 
Now the third row:
9. `achilles` starts the 3rd row.  key gives  `Zseq::Achilles(n=nrow(x=(data)))`. Tooltip: Powerful numbers that are not perfect squares.  72 is the smallest Achilles number.
10. `padovan` key gives `Zseq::Padovan(n=nrow(x=(data)))` Tooltip: Sum of last but 1 and last but 2 values. So from ...7, 9, 12, next is 7+9 = 16.
11. `triangle`  key gives  `Zseq::Triangular(n=nrow(x=(data)))`  Tooltip: Number of objects in a triangle, so 0, 1, 3, 6, 10...
12. `square`key gives `Zseq::Square(n=nrow(x=(data)),gmp=FALSE)`  Squares if each integer, so 1, 4, 9, 16.

Finally, in the main keyboard, there is a factorize key which uses the Desctools package.  There is also a factorize key in the gmp package.  Please could that be added - close to the factorize key.  It could be called factorize2.  Tooltip is the same as for factorize.

### LINKED ISSUE #7944: Add 2 keys and change help button in Prepare > Calculator > Transform keyboard

anastasia-mbithe:
Fixes #7944 and also Fixes #7876
 
I have included the change that was requested under PR #7949.


rdstern (approved):
@anastasia-mbithe that's great - thanks

anastasia-mbithe:
@rdstern,  since this is still open, I've added the suggested changes to the integer keyboard from issue #7876, so that we can close the issue as well.
Kindly have a look.

rdstern (approved):
@anastasia-mbithe that's great

lloyddewit (approved):

PR #7930, opened 2022-10-25
Improving Model> Fit model>Two variable & One variable dialogs

Vitalis95:
Fixes (partially) #7634 
This work is still in progress
@rdstern @N-thony @lloyddewit I have added the Bayesian method to Two Variable Fit model dialog.
The function `bayes_inference`  currently supports two level factor, the response variable (**Second Variable** in our case) can be either categorical (2 level factor) or numeric , the explanatory variable(**First Variable** in our case) is always categorical.
I am using `nc`  data to test, 
When the **Second Variable** is numeric i.e `gained` and first variable `habit`, the `sample statistic` should be `mean`
When the **Second Variable** is categorical i.e `lowbirthweight` and first variable `habit`, the `sample statistic` should be `proportion`, then specify which of its levels is the `success` in the **y**  argument (**Second Variable**) i.e `low` - the level to do inference on,

@rdstern , the number of Monte Carlo drawn when the type of inference is `credible interval`, is more than 10,000,R-Instat output window cannot display all of them including the summaries at the end, I have set it to print a maximum of 1000, the reason for this is that by default the global options of R cut off outputs after the 1000th entry -it will cut off at 1000th entry and display summaries at the end. Alternatively, if you are not happy with this, we then not save the model so that it just display the summaries

PR #7958, opened 2022-11-10
Added new options to Evapotranspiration dialog

MeSophie:
I made some modifications on this Dialog according to @rdstern comment on PR#7611.
@Patowhiz, @N-thony , @AmsaleEjigu please can you have a look.


rdstern (commented):
@MeSophie that's a good start.  Could you also look at the other suggestions I made in the layout in #7611.
Then, in addition, what I haven't done is to make much use of the dialogue.  I suggest you also do this with the ICRISAT data and with other data sets.  And that could be in parallel with the cosmetic layout improvements.
That should make it easy to write the help, and possibly produce a video on the use of the dialogue.
To start I wonder what examples are supplied with the package itself.  Then there is the bohicon data in the Instat library climatic guide data sets (omit the top about 13 lines).  And you should download the ICRISAT data I mention in 7611.

Those examples will (I hope) be fairly trivial to run, but let's see.  

Then we need to see how to use it when there are missing values and when elements are missing.  I think the function will cope, but this will need investigation.  I suggest it will be efficient for you to start with the documentation from the package and we may eventually have to look at the function itself - though I hope now!

MeSophie:
@rdstern Please can you go and look the comment that I made on PR#7611. Thanks

MeSophie:
I noticed that after filling in the 4 main parameters, the **Ok** button is active but if we delete any of these parameters, the **Ok** button is still active, so I have corrected this problem.

image

rdstern (changes requested):
@MeSophie I am now c learer on the possible improvements in layout.
image

In the image above
a) Change the label Radiation to Solar.
b) In the pull-down, change Data to Radiation
c) Don't have the Wind in a block.  Instead just have the existing Checkbox with label Wind Speed, directly underneath the Solar label.  Default - as now, is checked and the receiver is there.  If unchecked, then - as now - the receiver is no longer visible.
d) Please check whether the options for missing are all the same for the second main radio-button.  If so, then just improve the spacing a bit.

I then tried with the Bohicon data - here it is.  It seems to work!
[bohicon.zip](https://github.com/africanmathsinitiative/R-Instat/files/10058258/bohicon.zip)

Finally I noticed that the humidity data is not yet in our defined climatic data, so that's a new issue.  But let's see if this one can be merged first - and then we are ready to test it!

MeSophie:
> Please check whether the options for missing are all the same for the second main radio-button. If so, then just improve the spaci

@rdstern  The missing options contained in the missing options sub-dialogue are common to both radio buttons. I change the Radiation label to Solar but is still Radiation. So @Vitalis95  will change also the database.

rdstern:
@MeSophie thank you for the explanation about Solar.
There is now one further (welcome) change I am afraid!
image

The change is in the Define Climatic data, where the humidity variables are now defined - and this means that they can all be included automatically when the dialogue is loaded.  And that will make the method simpler to use.

So can you check how the existing elements are added automatically into the dialogue, and extend that now to the humidity variables.  You may have to wait till the Define Climatic Data dialogue is merged.

MeSophie:
> @MeSophie thank you for the explanation about Solar. There is now one further (welcome) change I am afraid! image
> 
> The change is in the Define Climatic data, where the humidity variables are now defined - and this means that they can all be included automatically when the dialogue is loaded. And that will make the method simpler to use.
> 
> So can you check how the existing elements are added automatically into the dialogue, and extend that now to the humidity variables. You may have to wait till the Define Climatic Data dialogue is merged.

@rdstern I'm working on it. All climatic data are include automatically for  Bihicon data but for Patancheru there is small issue with radiation.

MeSophie:
@lloyddewit , @rdstern this PR is ready to review.
There is still the radiation Solar parameter which does't fill automatically but @N-thony suggest that after merge the PR we can create and issue for this.

lloyddewit (approved):

rdstern (approved):
GReat

lloyddewit:
@rdstern if you approve, then I think we can merge, thanks

@ChrisMarsh82 The compile checks are failing (see screenshot below). However, they are failing on the compilation error that was fixed in PR #7997 . Master now builds. I don't understand why this PR tries to merge/build against a previous master commit. I reran the GitHub build checks in 4 other PRs and they all passed so I don't believe there is a general problem. I propose to merge this PR anyway. What's your view?

image

lloyddewit:
>  I reran the GitHub build checks in 4 other PRs and they all passed so I don't believe there is a general problem.

@ChrisMarsh82 
I just found that the build checks also fail in PR #7930 for the same reason. Not sure why.

instat/sdgMissingOptionsEvapotranspiration.vb
```diff
@@ -0,0 +1,61 @@
+﻿' R- Instat
+' Copyright (C) 2015-2017
+'
+' This program is free software: you can redistribute it and/or modify
+' it under the terms of the GNU General Public License as published by
+' the Free Software Foundation, either version 3 of the License, or
+' (at your option) any later version.
```
lloyddewit, line 7: just a test, please ignore
```suggestion
' (at your option) any later version. 
```

lloyddewit (changes requested):

lloyddewit:
@ChrisMarsh82 Your theory was correct! thank you!
Checks seem OK now. I'll merge this PR tomorrow.

rdstern (approved):
@MeSophie seems to work fine - even adding the humidity!  But not the sunshine - so that can become an issue.  Perhaps together with the 3rd method to add!

lloyddewit (approved):

PR #7964, opened 2022-11-15
Enhanced New Dataframe dialog

### LINKED ISSUE #7957: Improve the New Data Frame dialogue
@N-thony on the new data frame dialogue, when I change the type of the first variable to a factor, the 1,1000 is not cleared. Please make it such that when the variable is a factor the 1,1000 is cleared since it's not relevant anymore. Below is an image of what the dialog looks like when the first variable is changed to factor.
New Data Frame 11_10_2022 2_21_16 PM

N-thony:
Fixes #7957 
@berylwaswa can you test this? Thanks.

### DIFF
--- a/instat/dlgNewDataFrame.vb
+++ b/instat/dlgNewDataFrame.vb
@@ -600,11 +600,11 @@ Public Class dlgNewDataFrame
         If dataTypeGridView.CurrentCell.ColumnIndex = 2 Then
             Dim iColumnLevelIndex As Integer = dataTypeGridView.CurrentRow.Cells("colLevels").ColumnIndex
             Dim iRowLevelIndex As Integer = dataTypeGridView.CurrentRow.Cells("colLevels").RowIndex
-            Dim iColDefaultIndex As Integer = dataTypeGridView.CurrentRow.Cells("colDefault").ColumnIndex
-            Dim iRowDefaultIndex As Integer = dataTypeGridView.CurrentRow.Cells("colDefault").RowIndex
             Dim selectedCombobox As ComboBox = DirectCast(sender, ComboBox)
             If selectedCombobox.SelectedItem = "Factor" Then
                 dataTypeGridView(iColumnLevelIndex, iRowLevelIndex).ReadOnly = False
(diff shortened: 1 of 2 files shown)


rdstern (dismissed):
@N-thony I am approving, because it does solve the problem that we found.  So it could be merged.
But there are also problems with the defaults of letters(1,10) and LETTERS(1,10) that you may want to resolve at the same time.
There are 2 problems with those options:
a) when they are used they currently only work if the length of the data frame is a multiple of 10.  It usually will be, but if you put the usual `rep` round it, then it would always work. 
b) They won't work, when the variable is numeric or integer.  So check if they are set.  I suggest if they are set and the type is numeric or integer, you make that variable into character instead.  
c) Also they won't work if the type is set to be a factor and you have set the levels.  So, again, check that and make it a character variable instead.

instat/dlgNewDataFrame.vb
```diff
@@ -600,9 +600,12 @@ Public Class dlgNewDataFrame
         If dataTypeGridView.CurrentCell.ColumnIndex = 2 Then
             Dim iColumnLevelIndex As Integer = dataTypeGridView.CurrentRow.Cells("colLevels").ColumnIndex
             Dim iRowLevelIndex As Integer = dataTypeGridView.CurrentRow.Cells("colLevels").RowIndex
+            Dim iColDefaultIndex As Integer = dataTypeGridView.CurrentRow.Cells("colDefault").ColumnIndex
+            Dim iRowDefaultIndex As Integer = dataTypeGridView.CurrentRow.Cells("colDefault").RowIndex
```
lloyddewit, line 604: these are only used on line 608.
Move these lines into the line 606 `If` block.

lloyddewit (changes requested):
@N-thony Thanks, just 1 samll comment. Will you also make the changes requested by @rdstern?

N-thony:
> @N-thony Thanks, just 1 small comment. Will you also make the changes requested by @rdstern?

@lloyddewit I will make them later.

rdstern (approved):
Fine

lloyddewit (approved):

PR #7965, opened 2022-11-16
Create empty Search dialog

### LINKED ISSUE #7950: Add a simple Find dialogue into the Edit menu?
I would like to see the Edit menu "completed".  I am including @Patowhiz in this issue, just to get his take on the suggestions below.  Also it may prompt him to advise on how the Copy > Paste features can be completed - I think there is still a bit to go here.  Then @N-thony could advise, but this might be a simple task for Derrick or Victor, once agreed.

I would like to be able to use reogrid a bit more as it is possible in Excel.

Here is an example - it is the statquotes data, from File > Open from Library > Statquotes package
image

I wanted to find a particular quote - the one about `liars`.

This is a common problem.  But it isn't as general as in Excel, because almost always there is a single variable that I need to examine.  So I would be quite happy with a `Find ` that brings up a simple dialogue that just searches in one variable.  
And, of course, we already (sort of) have the dialogue in the Prepare > Column Text > Find/Replace 

image

So I would be quite happy with just a simple copy of this as a new dialogue.  
a) The usual selector on the left 
b) Single receiver on right  (meant for strings, but can take anything)
c) Pattern field
d) Ignore case
e) Include Regular Expressions (is also available.)

It then uses the stringr command to locate the first occurrence of the pattern.
The difference is what it does then.  It should mark that row, or text in some way. and go there. 

It should work on what's visible - so if a filter is in place, then it just searches the visible text.  I think that will be automatic.

Initially I am quite happy if it always starts at the top.  Later we could permit it perhaps to start from a particular line - that's not so trivial when the data are filtered perhaps.  Perhaps it would be easy to have a control 
From (Visible) Row and then an up-down control?  If so, then it could perhaps start on the current row of the data frame, when the dialogue is opened, or re-opened.  

One use of this dialogue is that the help could direct users to the other text options.  Perhaps this Find dialogue  could even have label saying `Use Prepare > Column: Text > Find/Replace for more options`

We could also perhaps implement `Find Next`, fairly trivially.  It would be disabled if there is no Find text.  It is enabled if there is, and there is a short-cut - perhaps <ctrl>,shift>F?  It doesn't give a dialogue, just goes to the next occurrence.

Then I propose we delete the Replace and Cut items from the menu.

mmulahvictor:
partially fixes #7950 
@africanmathsinitiative/developers this is ready for review. This is an empty Search dialog which should be merged so that I may progress with it.
- create the search dialog
- added the search dialog to form Main

### DIFF
--- a/instat/dlgSearch.Designer.vb
+++ b/instat/dlgSearch.Designer.vb
@@ -1,0 +1,42 @@
+﻿<Global.Microsoft.VisualBasic.CompilerServices.DesignerGenerated()>
+Partial Class dlgSearch
+    Inherits System.Windows.Forms.Form
+
+    'Form overrides dispose to clean up the component list.
+    <System.Diagnostics.DebuggerNonUserCode()>
+    Protected Overrides Sub Dispose(ByVal disposing As Boolean)
+        Try
+            If disposing AndAlso components IsNot Nothing Then
+                components.Dispose()
+            End If
+        Finally
+            MyBase.Dispose(disposing)
+        End Try
+    End Sub
+
+    'Required by the Windows Form Designer
+    Private components As System.ComponentModel.IContainer
+
+    'NOTE: The following procedure is required by the Windows Form Designer
+    'It can be modified using the Windows Form Designer.  
+    'Do not modify it using the code editor.
+    <System.Diagnostics.DebuggerStepThrough()>
+    Private Sub InitializeComponent()
+        Me.SuspendLayout()
+        '
+        'dlgSearch
+        '
+        Me.AutoScaleDimensions = New System.Drawing.SizeF(96.0!, 96.0!)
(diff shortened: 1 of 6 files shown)


rdstern (commented):
@N-thony I was going to approve this - without looking at the code, as that's not my role, and so it could be merged.  But I was a bit puzzled that 6 files were changed.  So perhaps you could have a quick look and then @lloyddewit can merge

N-thony (changes requested):
@mmulahvictor can you set the search menu visible to False

mmulahvictor:
@N-thony I have made the change you have requested, please look it out

N-thony (approved):
@lloyddewit over to you

lloyddewit (approved):

PR #7967, opened 2022-11-18
Added Expand button to the Stack dialog

### LINKED ISSUE #7941: Adding Expand to the Stack dialogue?
I need a second opinion on my solution and the location of this function.  I have been very pleased to find it.

It builds on the problem I raised in #7929 where the data were frequencies.  I wanted the be able to stretch out the data to show how it would have looked in "ordinary" form.  

I have now found a solution in the package called `vcdExtra`.  We already have vcd (Visualising Categorical Data) and have recently added `vcdExtra`, which contains many interesting data sets and a function we need, see below.  In a vignette called [Creating and manipulating Frequency Tables](https://cran.r-project.org/web/packages/vcdExtra/vignettes/A_creating.html) we have the following:

image

So this generalises the task from Beryl where we entered a single variable plus frequencies.  

But, particularly for teaching  it would be good to show what it would look like as "ordinary data".  Now I think this is the function called expand.dft, from the vcdExtra package.  

But, where should it go?  I suggest it is not important enough to be a menu item on its own.  I think of it as a bit like stack (pivot longer) so suggest it should become a third button there.  It could be just called Expand.  

It would have the usual data selector.  Then a multiple receiver called Factors: and a single receiver, called Frequencies:
This is an option in the vcdExtra function to change the names of the resulting factors.  I suggest this be ignored.
Then the resulting data frame is specified.

The columns are made into a data frame and then handed to the function.

Please check what happens if the combination of factor levels is less than the number of frequencies.  So, in the example above, what would happen if you only gave a single factor?  I assume we should trap that.  On the other hand I assume there could be examples with zero frequencies and that row could then be omitted.  So, in the example above would the function still work if one frequency were zero, so there were just 5 rows of data.  Or should there always be exactly 6 rows.  So should the check be that the combinations of factor levels is equal to the number of rows of (frequency) data, or less than or equal to? 

Perhaps @lilyclements could do a quick check that this is all sensible.  But I am now even more convinced that this would be a useful addition.  If so, it should be a simple addition to make.  And,  @N-thony has already specified that the vcdExtra package needs to be added.

My reasoning is that most of the datasets in the tidyr package and also those in histdata are frequencies.  So they appear more often than I had thought before.  Good that they are well understood!

Vitalis95:
Fixes #7941 
@rdstern , @N-thony @lloyddewit , the button converts a frequency table to a data frame representing individual observations in the table. I have convert the dataframe generated from characters to factors .

### DIFF
--- a/instat/dlgStack.vb
+++ b/instat/dlgStack.vb
@@ -24,9 +24,12 @@ Public Class dlgStack
     Private clsExpandFunction As New RFunction
     Private clsTypeConvertFunction As New RFunction
     Private clsSplitColumnsFunction As New RFunction
+    Private clsGetColumnNamesFunction As New RFunction
     Private clsPipeOperator As New ROperator
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
+    Private lstEditedVariables, lstAllVariables As New List(Of String)
+    Private bResettingDialogue As Boolean = False
 
     Private Sub dlgStack_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
@@ -65,9 +68,8 @@ Public Class dlgStack
         ucrReceiverColumnsToBeStack.Selector = ucrSelectorStack
         ucrReceiverColumnsToBeStack.SetParameterIsString()
 
-        ucrReceiverExpand.SetParameter(New RParameter("cols", 1))
         ucrReceiverExpand.Selector = ucrSelectorStack
-        ucrReceiverExpand.SetParameterIsString()
+        ucrReceiverExpand.SetParameterIsRFunction()
 
         ucrReceiverFrequency.SetParameter(New RParameter("freq", 2))
         ucrReceiverFrequency.Selector = ucrSelectorStack
@@ -208,6 +210,7 @@ Public Class dlgStack
         clsExpandFunction = New RFunction
         clsTypeConvertFunction = New RFunction
         clsSplitColumnsFunction = New RFunction
+        clsGetColumnNamesFunction = New RFunction
         clsPipeOperator = New ROperator
 
         ucrSelectorStack.Reset()
@@ -236,6 +239,7 @@ Public Class dlgStack
 
         clsExpandFunction.SetPackageName("vcdExtra")
         clsExpandFunction.SetRCommand("expand.dft")
+        clsExpandFunction.AddParameter("x", "columns", iPosition:=0)
 
         clsTypeConvertFunction.SetRCommand("type.convert")
         clsTypeConvertFunction.AddParameter("x", clsRFunctionParameter:=clsExpandFunction, iPosition:=0)
@@ -251,7 +255,6 @@ Public Class dlgStack
     Private Sub SetRCodeForControls(bReset As Boolean)
         ucrReceiverColumnsToBeStack.AddAdditionalCodeParameterPair(clsSplitColumnsFunction, New RParameter("items", 0), iAdditionalPairNo:=1)
         ucrSelectorStack.AddAdditionalCodeParameterPair(clsReshapeFunction, New RParameter("data", ucrSelectorStack.ucrAvailableDataFrames.clsCurrDataFrame, 0), iAdditionalPairNo:=1)
-        ucrSelectorStack.AddAdditionalCodeParameterPair(clsExpandFunction, New RParameter("x", ucrSelectorStack.ucrAvailableDataFrames.clsCurrDataFrame, 0), iAdditionalPairNo:=2)
 
         ucrSaveNewDataName.AddAdditionalRCode(clsUnnestTokensFunction, iAdditionalPairNo:=1)
         ucrSaveNewDataName.AddAdditionalRCode(clsTypeConvertFunction, iAdditionalPairNo:=2)
@@ -279,7 +282,6 @@ Public Class dlgStack
         ucrChkDropPrefix.SetRCode(clsPivotLongerFunction, bReset)
         ucrFactorInto.SetRCode(clsReshapeFunction, bReset)
         ucrReceiverFrequency.SetRCode(clsExpandFunction, bReset)
-        ucrReceiverExpand.SetRCode(clsExpandFunction, bReset)
     End Sub
 
     Private Sub TestOKEnabled()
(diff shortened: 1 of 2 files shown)


rdstern:
@Vitalis95 this is looking promising.
a) Could you change Factor to Factor(s)

b) I don't see the frequencies, f. to put into the dialogue.  Notice they are numeric columns here.  I think they have to be a factor to see them.  I am happy for the factor to be a factor, but the frequencies is usually numeric.
image

c) So I made the frequencies into a factor.  Then I could use it and got this error:

image

I tried making the variable name freq and it still gave me the same error.

Finally I tried (stupidly) with x instead of f (or freq) twice and the dialogue worked, but with odd answers, of course!  So I wonder if you need to tell the vcdExtra the frequencies in the data frame you make?

Vitalis95:
@rdstern ,the frequencies is also included first in the multiple receiver `Factor(s):`, before adding it to single receiver, this is because the columns are made into a data frame and then handed to the function as you had indicated in the issue.

rdstern:
@Vitalis95 that now works fine.

Except I don't want users to have to put the frequencies into the factor(s) receiver.  That's very confusing.
Why can't I just put it into the frequencies receiver and you then add it into the data frame, before giving it to the vcdExtra function?

Vitalis95:
I am including @N-thony in the conversation, after we discussed he suggested we implement that way- like in Transpose dialog where the variable in single receiver is included first in multiple receiver. @N-thony, is there a way we can implement above suggestion? here is that function; 
 `expand.dft(x, frq="Freq")`  where `x`  is a table object, or a data frame in frequency form containing factors and one numeric variable i.e `frq` - the variable in frequencies receiver

Vitalis95:
@rdstern , have a look at now. Thanks

rdstern (approved):
@Vitalis95 and @N-thony it seems to work fine now.
I tried with @berylwaswa school data from 40 children - and fine.
Then I tried with abortion data, from vcdExtra which had 3 factors, plus the frequencies.  It works with all 3 factors, as well as when I put only a single factor.
Great!  I hope it can be merged soon.  
Does a new facility like this imply the translation should also be added.  If so, then that could come later?

instat/dlgStack.vb
```diff
@@ -269,7 +305,14 @@ Public Class dlgStack
             Else
                 ucrBase.OKEnabled(True)
             End If
+        Else
+            If ucrReceiverExpand.IsEmpty OrElse ucrReceiverFrequency.IsEmpty OrElse Not ucrSaveNewDataName.IsComplete Then
+                ucrBase.OKEnabled(False)
+            Else
+                ucrBase.OKEnabled(True)
+            End If
```
lloyddewit, line 313: Lines should not go past column 100.
Is this more readable?
```suggestion
            ucrBase.OKEnabled(Not ucrReceiverExpand.IsEmpty _
                              AndAlso Not ucrReceiverFrequency.IsEmpty _
                              AndAlso ucrSaveNewDataName.IsComplete)
```

instat/dlgStack.vb
```diff
@@ -369,4 +420,18 @@ Public Class dlgStack
         End If
         TestOKEnabled()
     End Sub
+
+    Private Sub ucrReceiverExpand_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverExpand.ControlValueChanged, ucrReceiverFrequency.ControlValueChanged, ucrSelectorStack.ControlValueChanged
+        Dim lstVariables As New List(Of String)
+        If ucrReceiverExpand.lstSelectedVariables.Items.Count > 0 Then
```
lloyddewit, line 426: Follow 'happy path stays left' principle?
```
        If ucrReceiverExpand.lstSelectedVariables.Items.Count <= 0 Then
            Exit Sub
        End If
```

instat/dlgStack.vb
```diff
@@ -369,4 +420,18 @@ Public Class dlgStack
         End If
         TestOKEnabled()
     End Sub
+
+    Private Sub ucrReceiverExpand_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverExpand.ControlValueChanged, ucrReceiverFrequency.ControlValueChanged, ucrSelectorStack.ControlValueChanged
+        Dim lstVariables As New List(Of String)
```
lloyddewit, line 425: declare on line 247?
```suggestion
```

instat/dlgStack.vb
```diff
@@ -369,4 +420,18 @@ Public Class dlgStack
         End If
         TestOKEnabled()
     End Sub
+
+    Private Sub ucrReceiverExpand_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverExpand.ControlValueChanged, ucrReceiverFrequency.ControlValueChanged, ucrSelectorStack.ControlValueChanged
+        Dim lstVariables As New List(Of String)
+        If ucrReceiverExpand.lstSelectedVariables.Items.Count > 0 Then
+            lstVariables = ucrReceiverExpand.GetVariableNamesAsList()
```
lloyddewit, line 427:
```suggestion
            Dim lstVariables As List(Of String) = ucrReceiverExpand.GetVariableNamesAsList()
```

lloyddewit (changes requested):

Vitalis95:
@lloyddewit , have a look at it. Thanks

rdstern (changes requested):
@Vitalis95 still looks good.  Except that when you change dataframe it correctly clears the receivers, but it should also go back to the first control - the one for the factors.  Currently it stays on the Frequencies.

Vitalis95:
@rdstern , have look at it. Thanks

rdstern:
@Vitalis95 it would be good to merge this one, if you have time for the small change above.

Vitalis95:
@rdstern , I did made the above change- when dataframe is changed, it correctly clears the receivers and  goes back to the first control and it works fine from my side. Thanks

rdstern:
@Vitalis95 thanks.  Let me check

rdstern (approved):
@Vitalis95 great.  Hope this passes the @lloyddewit tests, so it can be merged for the current release

lloyddewit (approved):

PR #7970, opened 2022-11-21
Added summaries for frequency data in the Calculator dialog

### LINKED ISSUE #7929: Adding summaries of data from frequency tables?
I asked @berylwaswa for a school example of data that could be typed in, after using the File > New Data Frame dialogue.
She has given me the following example:
image

I initially rejected it, because it is using data on frequencies and I keep hoping we will move past that.  So I would prefer examples that don't have grouped frequencies.  But perhaps I am wrong and should accept the (school) world as it is, rather than what I would like.  And it is a very nice example to show the use of the File > New Data Frame, which I show first.  Then I now suggest we add the calculation of the summaries for frequency data into the calculator.  But, before proceeding please could @lilyclements check what I am suggesting below.  Is this ok , and should there be more keys (i.e. summaries) added?

And I would like (simply) to reshape the data, into a data frame with 40 rows, from the frequency data entered into the 14 rows.  @lilyclements how can I do this with the current dialogues in R-Instat.  I am not sure it is worth a new dialogue, but don't see how to do this currently.  I am sure it is quite easy.

a) Data entry.  I suggest we do it twice:
1) First we can just change the 10 to 14 and press ok.
2) Then rename each variable to mass and number respectively.
3) Then enter the data into the first variable - so 47 to 60 into mass and the frequencies into the column called number.
4) Then make the 2 columns into type numeric.
5) Second time - we make the point it is good to check whether we could use more of the New data frame dialogue to make the data entry task easier.
6) So tick on the variable name option
7) Then name the 2 variables as mass and number
8) Now change the 1,1000 into 47,60
9) And make the second variable numeric
10) Then press ok and enter the second variable.

b) Now that's the end of the data entry.  But useful to complete the task and get the mean (and other summaries) of the frequency data.  So here is what I now suggest.  The comes from the answer to this [stack-overflow question](https://stackoverflow.com/questions/22644481/r-computing-mean-median-variance-from-file-with-frequency-distribution)). 

I suggest we add a section to the summary keyboard in the calculator.  Once we know what's needed, @anastasia-mbithe is very skilled at this.
Perhaps a box with label Frequencies.  Then keys with label mean, median, var, sd.  @lilyclements could you check it works fine and I wonder if we should add any more keys?
mean is the function mean(rep(mass, number)), so the key will be mean(rep(x , )) - no x, but the cursor there.
similarly for median, var and sd.  (Is it really so simple!)
And I now do think this works generally, so we should add a few more keys - for the mode, quantile, IQR, and possibly even the min and max - so people don't have to think about it.  And perhaps distinct as well!  I specify this below.

c)I have now also found how to "stretch the data?  I would like to also say why - if these are 40 students we often have more information about the students.  For example perhaps 20 are girls and the others are boys.  That's natural on the data as collected - in the 40 rows, but messy to add to the frequencies.

### LINKED ISSUE #7945: Improve the Prepare > Calculator > Summary keyboard
@anastasia-mbithe there is already an issue on adding another section to this keyboard.  This adds to the work on the existing keys.

a) The mode key can produce multiple modes if they exist.  This can currently only be saved into a new variable if it is a whole fraction of the length.  So add the rep function round it as you have done elsewhere:

So change the current : `statip::mfv()`

Into `rep(x=statip::mfv(), len=nrow(x=survey))`  of course changing to the current data frame.

b) Change the quantile function, so the default includes the argument `probs = 0.5` and include the same rep round it, because we can ask for multiple quantiles in one go if we wish.

c) Change the help into the new system.  You can keep the base package as the first and default.  The second is the stats package, then there is statip, and e1071, robustbase and raster.

d) Add tooltips to each of these entries.  I will supply them below:

anastasia-mbithe:
Fixes #7945 and also fixes #7929

@rdstern, this is ready for testing.


rdstern (changes requested):
@anastasia-mbithe this looks great already.
a) In the frequencies the only one that seems not to work is the mode1.  Could you use the same function as in the  in the ordinary summary, where it is statip::mfv1( )
b) The quantile could (I think) be made simpler in the frequencies part by quantile(x=rep( x, f ), probs=0.2, na.rm=TRUE) - so leaving the additional bit out.
c) Could you add the arguments in the frequencies if the corresponding checkbox is ticked.  Then they could become, for example  sum(x=rep(x= ,times=   ), na.rm=TRUE)
d) In the ordinary summaries they have the na.rm=TRUE when you ask for the arguments, but not otherwise.  Could you add it in - as you have for the frequencies andway.
e) I thought we might not need tooltips for the frequencies.  But have `complete the command by rep(d ,f) for data in variable called d and frequencies in f`.  That's for all except the propn and the quantile.  For propn `Complete as rep(d > 50, f)) to give the proportion of values in variable d more than 50`.  For quantile give `Complete as rep(d ,f), prop=0.2), for example, for the 20% point.`

anastasia-mbithe:
@rdstern, the issues/changes are fixed. Kindly have a look.

rdstern:
@anastasia-mbithe almost there:
a) the new mode1 in the frequencies needs the rep also  included.  `statip::mfv1(rep(x,f) , na.rm=TRUE)`
b) You have missed the na.rm=TRUE from the ordinary `mode1` and `kurtosis`.
c) Could you add the equivalent - which is `use="na"` for the cor and cov keys

anastasia-mbithe:
@rdstern, should we also change the value of the parameter "use", in cor and cov keys when we have the show parameters checkbox checked? currently, we have use= "everything"

rdstern:
@anastasia-mbithe good spot - we should be consistent.  Or should we change the Use to everything above?  I would just like it to cope with missing better than always giving NA.  I chose that option, because it said it copes with no data - rather than giving an error.

But now I tried "everything" and "na" with lots missing and I think I prefer being consistent with "na".

rdstern (approved):
Great

instat/ucrCalculator.vb
```diff
@@ -260,6 +261,51 @@ Public Class ucrCalculator
...
+        ttCalculator.SetToolTip(cmdFreqLength, "complete the command by rep(d ,f) for data in variable called d and frequencies in f")
+        ttCalculator.SetToolTip(cmdFreqMad, "complete the command by rep(d ,f) for data in variable called d and frequencies in f")
+        ttCalculator.SetToolTip(cmdFreqMax, "complete the command by rep(d ,f) for data in variable called d and frequencies in f")
+        ttCalculator.SetToolTip(cmdFreqMean, "complete the command by rep(d ,f) for data in variable called d and frequencies in f")
+        ttCalculator.SetToolTip(cmdFreqMedian, "complete the command by rep(d ,f) for data in variable called d and frequencies in f")
+        ttCalculator.SetToolTip(cmdFreqMin, "complete the command by rep(d ,f) for data in variable called d and frequencies in f")
+        ttCalculator.SetToolTip(cmdFreqMiss, "complete the command by rep(d ,f) for data in variable called d and frequencies in f")
+        ttCalculator.SetToolTip(cmdFreqMode1, "complete the command by rep(d ,f) for data in variable called d and frequencies in f")
```
lloyddewit, line 302: Can we remove the duplication by using a constant?
```
Const strCmdToolTip As String = ""complete the command by rep(d ,f) for data in variable called d and frequencies in f"
ttCalculator.SetToolTip(cmdFreqDistinct, strCmdToolTip)
ttCalculator.SetToolTip(cmdFreqIQR, strCmdToolTip)
'etc.
```

lloyddewit (approved):
@anastasia-mbithe Thank you, this looks good to me.
I just had one small comment but it's so small I suggest we ignore it, and I'll go ahead and merge.

Issue #7273, opened 2022-03-05
Navigating the grid using keyboard arrows


shadrackkibet:
In Excel, you can always navigate around the cells using keyboard arrows. I think this is a nice to have feature in the data view window.  
![image](https://user-images.githubusercontent.com/26301973/156886777-9ca28297-928e-467d-8ffe-92d4ea0ba5f2.png)

rdstern:
Agreed - didn't realise it didn't!

N-thony:
@mmulahvictor can you try this? Let me know if you need help.

mmulahvictor:
@N-thony I will try it out.

PR #7973, opened 2022-11-23
Implemented Tab and arrow keys in Data grid

### LINKED ISSUE #7956: Add tab and other keys to improve data entry into reogrid
Here is a dataset I have just entered into R-Instat, i.e. using reogrid.  I found it awkward in R-Instat compared to Excel, but also had a small problem in Excel.
image

Before adding to the facilities in reogrid, I would like the programmer to try the task I set myself.  

Now @berylwaswa who I hope will evaluate the improvements in the corresponding pull request, has done a video on the entry of these data.  But she entered a column at a time.  I now want to enter one row at a time.  I am simulating the system where we could ask successive children in a class to each enter a row.  There may be more than 2 columns to enter, but this example does fine to see what's wanted.

Perhaps you could also try in Excel or Open Office, because that seems to make the entry easier.

I suggest there are some aspects that can be improved immediately.  Then I have one problem even in Excel and one awkwardness in reogrid.

a) The Tab key works in Excel and moves across.  So you can type number <Tab> number <Tab> to go across.  And <Shift><Tab> in Excel goes back.
This doesn't seem to do anything in reogrid.  So could we implement the <Tab> in reogrid?
b) Similarly the arrow keys work in Excel, but don't seem to do anything in R-Instat.
c) Now an extra that I don't know how to do in Excel.  After finishing typing across I would like to go to the start of the next row.  In Excel there is no limit in the number of columns so <Tab> keeps going across.  In R-Instat could <Tab> do this from the last column, i.e. it then goes to the first variable on the next row?  That would be very nice!
d) I assume, if this works above, then it will work with select.  So we could then enter just a few variables and then take off the selection.  I suspect this may be automatic, so can be ignored first time.
e) And I sometimes find a problem typing data into a cell.  I think that's when I first start in that window.  Then it refuses.  If I click again then it moves in the cell and is ok - and it seems ok from then on.  Perhaps you could check this "feature" at the same time.

derekagorhom:
fixes partly #7956 and #7273

@N-thony and @rdstern  this PR is ready for review

### DIFF
--- a/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
@@ -174,8 +174,8 @@ Public Class ucrDataViewReoGrid
     End Sub
 
     Private Sub Worksheet_BeforeCellsKeyDown(sender As Object, e As BeforeCellKeyDownEventArgs)
-        e.IsCancelled = True
         If e.KeyCode = unvell.ReoGrid.Interaction.KeyCode.Delete OrElse e.KeyCode = unvell.ReoGrid.Interaction.KeyCode.Back Then
+            e.IsCancelled = True
             RaiseEvent DeleteValueToDataframe()
         End If
     End Sub


derekagorhom:
this PR fixes the a) and b) in #7956

rdstern:
@derekagorhom well done.  This already seems to work almost perfectly.  You have even included the <Tab> to go down to the next record.  
Just one small problem.  It doesn't work for me right at the start.
When I open a dataset (or restart by Closing all datasets and then opening anything again) You are in the place A1 (in Excel jargon).  Neither Tab or the arrow keys seem to be activated.  They don't do anything.

As soon as I move from there, it springs into life and seems to work perfectly.  Then, if I add another dataset, that's also ok.  It is just right at the start.
Once that's fixed, let's merge it.

derekagorhom:
@rdstern the issue you raised seems to also happen in both the column metadata and also the dataframe metadata. i will contact antoine and patrick to figure out the problem is coming from

rdstern:
@derekagorhom I hadn't thought of that - and good you noticed.  I assume it must be something to do with your (or some) code that is run when the reogrid position is changed.  It will be a good small puzzle to solve.

rdstern (approved):
@derekagorhom I am provisionally approving.  The problem we have found, of "waking up" the additional keys may be a feature?  Also I would like @lloyddewit to check the code and he may have a suggestion at the same time.  Meanwhile, if the code is ok, then I would like to proceed, even if sorting out the start-up becomes another issue.

lloyddewit (approved):

PR #7920, opened 2022-10-18
Adding the three Variable functionality to the Describe> Two\Three Summary dialogue

Vitalis95:
Replaces PR #7614 
@rdstern @N-thony , part a) and b) of the https://github.com/africanmathsinitiative/R-Instat/pull/7614#issuecomment-1239141364 is sorted
@lilyclements , could check on https://github.com/africanmathsinitiative/R-Instat/pull/7614#issuecomment-1277734379 for part c)

PR #7978, opened 2022-11-24
Adding the three Variable functionality to the Describe> Two\Three Summary dialogue

Vitalis95:
Replaces PR #7920 
@rdstern @N-thony ,these are  improvements to 2/3 summary dialog, the issue you raised of not getting numerical summaries  if you have ffn i.e  Village by Variety by Yield is now solved. Have a look at it. Thanks

### DIFF
--- a/instat/dlgDescribeTwoVariable.Designer.vb
+++ b/instat/dlgDescribeTwoVariable.Designer.vb
@@ -38,259 +38,73 @@ Partial Class dlgDescribeTwoVariable
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()>
     Private Sub InitializeComponent()
-        Me.cmdSummaries = New System.Windows.Forms.Button()
-        Me.lblFirstVariable = New System.Windows.Forms.Label()
-        Me.lbSecondVariable = New System.Windows.Forms.Label()
-        Me.grpOptions = New System.Windows.Forms.GroupBox()
-        Me.cmdMissingOptions = New System.Windows.Forms.Button()
-        Me.ucrChkOmitMissing = New instat.ucrCheck()
-        Me.lblSummary = New System.Windows.Forms.Label()
-        Me.lblFirstType = New System.Windows.Forms.Label()
-        Me.lblBy = New System.Windows.Forms.Label()
-        Me.lblSecondType = New System.Windows.Forms.Label()
-        Me.lblSummaryName = New System.Windows.Forms.Label()
-        Me.grpSummaries = New System.Windows.Forms.GroupBox()
-        Me.rdoCustomize = New System.Windows.Forms.RadioButton()
-        Me.rdoSkim = New System.Windows.Forms.RadioButton()
-        Me.lbSecondOpt = New System.Windows.Forms.Label()
-        Me.lblSecondFactor = New System.Windows.Forms.Label()
-        Me.lblNumericVariable = New System.Windows.Forms.Label()
-        Me.lblMarginName = New System.Windows.Forms.Label()
-        Me.grpDisplay = New System.Windows.Forms.GroupBox()
+        Me.ucrNudColumnFactors = New instat.ucrNud()
         Me.ucrReceiverPercentages = New instat.ucrReceiverSingle()
         Me.lblFactorsAsPercentage = New System.Windows.Forms.Label()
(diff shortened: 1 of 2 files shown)


rdstern:
@Vitalis95 this is now looking good.  I recognise that these will later have to be adapted for the new output window, but I would like to get these merged now.  Then do that as a separate issue.
The only problem I could find is that it  seems to make the tables twice.  Here is an example with our usual survey data:
image

I expect 2 3*way tables, but I get 4!

Here is the code:

```
# Code generated by the dialog, Describe Two Variables
multiple_receiver_cols <- c("yield", "fert")
list_of_tables  <- survey %>% purrr::map(.x=multiple_receiver_cols, .f= ~ data_book$summary_table("survey", summaries=c("summary_sum"), columns_to_summarise=.x, factors=c("variety", "village"), include_margins=TRUE, margin_name="All", j=1)) %>% purrr::map(.x=., .f= ~ mmtable2::mmtable(data=.x, cells=value) + mmtable2::header_top_left(variable="village") + mmtable2::header_left_top(variable="variety") + mmtable2::header_top_left(variable="summary-variable"))
list_of_tables 

```
I note for the change to the output window that it may not (yet) cope with multiple html outputs from the 1 command.  When we make the change we can either trap that or wait till it is allowed.

rdstern:
@Vitalis95 I tried again with the latest version.  Here is the output window, which shows the double output, and hence gives the tables twice:
image

@Patowhiz I get the output twice in the browser.  This will change anyway with your new system.  Should this be merged and then you either look at the code, or perhaps @Vitalis95 can do this - for your new output?
I am approving in case that is a route to the solution.

rdstern (approved):

lloyddewit:
> @lloyddewit is this ok, by you, to merge?

@rdstern There is still an open question for @Patowhiz (see comment from 04 Dec). I did not want to do the technical review before this was resolved. If you're happy to proceed without @Patowhiz 's response, then please let me know.

@Patowhiz Did you get a chance to look at this?

Thanks

image

Patowhiz:
@Vitalis95 does this PR contain the changes from the master? 
Any dialog that produces an output that goes to the output window should "add" and "get object". From the script produced by @rdstern above, it seems not to be adding an object.
@anastasia-mbithe was also working on multiple html outputs. From our last conversation, we agreed that she should research on how we could combine multiple html objects into one. Similar to how we do for graphs and grids ( we use `gridExtrapackage` for that). I'm not sure if there is search a package for html objects.
I would recommend you both (@Vitalis95 and @anastasia-mbithe) look into such possibilty, that will just be supported by the new system.
If it's not possible we can look at alternative ways.

Patowhiz:
This PR relates to [comment](https://github.com/africanmathsinitiative/R-Instat/pull/7999#issuecomment-1375215197) in PR #7999 . The both produce an object that has multiple html objects.

Vitalis95:
@Patowhiz , the object being produced by the function  is of class type `list`,  I have tried to add the  "add" and "get object" as implemented in **Summary Tables dialog** but still produces the output in the browser
Does the new output system also apply to objects of class type `list` or only `tables` ?
Thanks
image

rdstern (commented):
@vitalis this still went to the browser and then froze for me.  I tried a 3-way frequency table.
I note you are still waiting for input.

Vitalis95:
@rdstern , we looked at this issue during the sprint, we realized that the object being produced by  ` summary function` is a `list`.
`html` output does not support objects of class `list`, we couldn't immediately find a function that converts a `list` to `gt_tble` , an object supported by `html`  and @Patowhiz suggested he is going to look into it.

rdstern:
@Vitalis95 it would be very good to have this improvement included in the February update, so please keep following up with Patrick - he has a lot on his plate just now, so I hope he will find the time.  Are there any other options?

rdstern:
@Vitalis95 this seems to be another dialogue that still awaits either @ChrisMarsh82 and/or @Patowhiz to imporve the html windows.  Please can you keep asaking them.  I believe Patrick may soon be spending time in Maseno?

rdstern:
@anastasia-mbithe and @lilyclements am I right that you will be taking over this pull request?  I assume the changes will include getting rid of mmplot2 and possibly also making the first multiple control into a single one, so it produces just one table?

anastasia-mbithe:
@N-thony The changes were incorporated in #7999

PR #7988, opened 2022-11-27
Ensured that script window is not corrupted when auto-save triggered

lloyddewit:
Fixes 6720.
I also removed the `frmScript` class because this is no longer needed.

In the database, I deleted the following rows:

image

image

And I inserted the following new rows:

image


lloyddewit:
@rdstern @N-thony This PR updates the project file and translation database, so it risks creating merge conflicts in other PRs. There is not much in this PR to test or review. Therefore, I will go ahead and merge. I hope that's OK.

PR #7991, opened 2022-11-28
Added random sampling to Probability keyboard in Calculator dialog

### LINKED ISSUE #7953: Adding random sampling to the probability keyboard in the calculator
We don't yet have the commands to generate random samples.  I suggest they go in the probability keyboard they are a new section called `Random`. 

a) the existing probability keys are 5 across.  So I suggest we continue with that layout.
b) The first row might be `uniform, normal, uni_integer, bernoulli, binomial`
c) the second row could then be `poisson, nbinomial, gamma, beta, sample` 

d) Uniform is runif(n) - where the n is the usual n=nrow(x=survey), etc, because we must have a sample of size the same as the length of the particular data frame.  The tooltip `Random sample from the uniform distribution between 0 & 1. Use say runif(n,5,10) to change the range`.
(Use examples from the Integer keyboard to understand the code needed to produce the "n" above and in other examples.)
e) normal is rnorm(n) with tooltip `Random sample from the standard normal distribution.  Use, say rnorm(n,100,15) to change the mean and sd`.
f) uni-integer is sample.int(5, n,replace=TRUE) with tooltip `Random integers between 1 and 5. Use say sample.int(3,n,TRUE,prob=c(6,3,1)) to sample 1 to 3 with defined probabilities. `
g) bernoulli is rbinom(n,1,prob=0.5) with tooltip Random Bernoulli (0 or 1) sample.  Use say rbinom(n,1,1/6) for a random sample of a given dice value.
h) binomial is rbinom(n,3,prob=0.5) with tooltip `Random binomial sample with values between 0 and 3.  Use say rbinom(n,5,prob=0.1) for other distributions`.
i) poisson is rpois(n,lamda=1) with tooltip: `Random Poisson sample  with mean 1.  Change the mean as required.`
j) negative binomial is rnbinom(n,size=1,prob=0.5) tooltip Random geometric sample as given, i.e. number of failures before size=1 success.  Change value of size (must remain positive) for other negative binomials.  
k) gamma is rgamma(n,shape=1,scale=2) tooltip `Random sample with shape=1 is from the exponential distribution with mean 2.  Keep scale (>0) = 2 and change shape (>0) to 0.5 for chi-square distribution with 1d.f. or to 5, for 10 d.f.  Keep scale = 2 for chi-square.  
l) beta is rbeta(n,shape1=1,shape2=1) With tooltip "This special case is the uniform distribution between 0 and 1.  Change shape1 > 0 and shape2 > 0 for different beta distributions.
m) sample is sample(  ,n,replace=TRUE)  `Random sample with replacement, from a given variable.  Change replace to FALSE for a random permutation`.      `

derekagorhom:
fixes #7953

### DIFF
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -309,7 +309,7 @@ Public Class ucrCalculator
         ttCalculator.SetToolTip(cmduniform, "Random sample from the uniform distribution between 0 & 1. Use say runif(n,5,10) to change the range")
         ttCalculator.SetToolTip(cmdRan_normal, "Random sample from the standard normal distribution. Use, say rnorm(n,100,15) to change the mean and sd")
         ttCalculator.SetToolTip(cmduni_integer, "Random integers between 1 and 5. Use say sample.int(3,n,TRUE,prob=c(6,3,1)) to sample 1 to 3 with defined probabilities.")
-        ttCalculator.SetToolTip(cmdbernouli, "Random Bernoulli (0 or 1) sample. Use say rbinom(n,1,1/6) for a random sample of a given dice value.")
+        ttCalculator.SetToolTip(cmdbernoulli, "Random Bernoulli (0 or 1) sample. Use say rbinom(n,1,1/6) for a random sample of a given dice value.")
         ttCalculator.SetToolTip(cmdbinomial, "Random binomial sample with values between 0 and 3. Use say rbinom(n,5,prob=0.1) for other distributions")
         ttCalculator.SetToolTip(cmdpoisson, "Random Poisson sample with mean 1. Change the mean as required.")
         ttCalculator.SetToolTip(cmdnbinomial, "Random geometric sample as given, i.e. number of failures before size=1 success. Change value of size (must remain positive) for other negative binomials.")
@@ -4599,7 +4599,7 @@ Public Class ucrCalculator
         End If
     End Sub
 
-    Private Sub cmdbernouli_Click(sender As Object, e As EventArgs) Handles cmdbernouli.Click
+    Private Sub cmdbernouli_Click(sender As Object, e As EventArgs) Handles cmdbernoulli.Click
         If chkShowParameters.Checked Then
             ucrReceiverForCalculation.AddToReceiverAtCursorPosition("rbinom(n = ,size = 1,prob = 0.5)", 1)
         Else
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -4611,8 +4611,8 @@ Public Class ucrCalculator
         clsDataFunction.AddParameter("x", ucrSelectorForCalculations.ucrAvailableDataFrames.cboAvailableDataFrames.SelectedItem, iPosition:=0)
 
         clsUnitegerFunction.SetRCommand("sample.int")
-        clsUnitegerFunction.AddParameter("size", "5", iPosition:=0)
-        clsUnitegerFunction.AddParameter("n", clsRFunctionParameter:=clsDataFunction, iPosition:=1)
+        clsUnitegerFunction.AddParameter("n", "5", iPosition:=0)
+        clsUnitegerFunction.AddParameter("size", clsRFunctionParameter:=clsDataFunction, iPosition:=1)
         clsUnitegerFunction.AddParameter("replace", "TRUE", iPosition:=2)
 
         ucrReceiverForCalculation.AddToReceiverAtCursorPosition(clsUnitegerFunction.ToScript, 0)
@@ -4695,8 +4695,8 @@ Public Class ucrCalculator
         clsDataFunction.AddParameter("x", ucrSelectorForCalculations.ucrAvailableDataFrames.cboAvailableDataFrames.SelectedItem, iPosition:=0)
 
         clsSampleFunction.SetRCommand("sample")
-        clsSampleFunction.AddParameter("size", "", iPosition:=0)
-        clsSampleFunction.AddParameter("n", clsRFunctionParameter:=clsDataFunction, iPosition:=1)
+        clsSampleFunction.AddParameter("n", "", iPosition:=0)
+        clsSampleFunction.AddParameter("size", clsRFunctionParameter:=clsDataFunction, iPosition:=1)
         clsSampleFunction.AddParameter("replace", "TRUE", iPosition:=2)
 
         ucrReceiverForCalculation.AddToReceiverAtCursorPosition(clsSampleFunction.ToScript, 0)
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -316,8 +316,6 @@ Public Class ucrCalculator
         ttCalculator.SetToolTip(cmdRan_gamma, "Random sample with shape=1 is from the exponential distribution with mean 2. Keep scale (>0) = 2 and change shape (>0) to 0.5 for chi-square distribution with 1d.f. or to 5, for 10 d.f. Keep scale = 2 for chi-square")
         ttCalculator.SetToolTip(cmdRan_beta, "This special case is the uniform distribution between 0 and 1. Change shape1 > 0 and shape2 > 0 for different beta distributions.")
         ttCalculator.SetToolTip(cmdRan_sample, "Random sample with replacement, from a given variable. Change replace to FALSE for a random permutation.")
-
-
     End Sub
 
     Public Sub Reset()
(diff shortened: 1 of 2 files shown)


N-thony:
@derekagorhom good, @anastasia-mbithe can you peer review this?

rdstern:
@derekagorhom that's a very good start indeed.  Well done.  I did say you could use this example to also learn more that would help you generally.  When implementing I need you to reflect that your job includes thinking hard on how it will be used - so sometimes improving on what I have written as the issue.  Also, making sure you understand the issue well.
a) Keys well laid out with only one spelling I found, namely `bernoulli` has double l.
b) You have misunderstood the n in (possibly all) the keys I outlined in the issue.  They each need a sample of the same size as the runif key - which you coded fine.  So the code in runif should be used in the others.
c) You should notice there is a `Show Parameters` checkbox in the calculator.  With most keys we only show the minimum parameters for it to work when that is not checked.  Then we show all the parameters, with their default settings, when it is chec ked.  That may be relevant for some keys.
d) But for others these keys as examples where it is useful to have the parameters anyway.  Then it helps the user to know what to change if they want another instance of that key.  So the runif, please add the min = 0 and max = 1 into the ordinary use of the function.  So it won't change when the show parameters checkbox is checked, because there is nothing more to add.
e) Similarly for rnorm(n).  I now suggest that you add the full set of arguments anyway, i.e.` rnorm(n, mean = 0, sd = 1)`  Of course with your special adjustment for n.
f) Probably something similar for most of the other keys - some you have done already.
g) Then, your own testing should mean you produce or use a data frame and try each key yourself.

rdstern (changes requested):
@derekagorhom that is better - and quick too.  But I have deduced that you can't have tried your code.  You should not expect your tester to be the first person to try the different bits of code you write.  I would prefer you to take a bit longer, but be confident that you tested yourself, and critically, before expecting others.  Perhaps you were still going to do this and I am testing too soon?
a) My evidence is particularly the poisson key, which doesn't work.  I am sure you can discover and correct the error.
b) The uni_integer is a bit more complicated, but try with your default settings and (for example) the survey data, which has 36 observations.  That will give an error, because 5 does not go into 36.  Then you will discover that you need to swap n and size round.    
c) And the sample also doesn't work.  There you need to use size, where you have n, and you will have x = , ready for the user, usually to include a variable.  (You could alternatively have a number there, and then it is effectively the same as uni_integer.)

rdstern:
Progress - I knew you would be able to combine this task with learning a lot - both about what to do, and also how to do it.
 a) Poisson - that was the simple one and you found it!  Good.
b) int_uniform.  Your version was as follows: 
`sample.int(size=5, n=nrow(x=survey), replace=TRUE) `

I assume you tried it and found it didn't work.  Then look at the reference documentation on that command to make the correction - which I thought I had mentioned above.  If it did work for you, it is because your data frame was of length a multiple of 5 and you should have noticed that the values keep repeating.  It gave an error for me, because the worksheet was of length 36.  Changing to 
`sample.int(n=5, size=nrow(x=survey), replace=TRUE)`
c) Now can you change the sample key?

derekagorhom:
> Progress - I knew you would be able to combine this task with learning a lot - both about what to do, and also how to do it. a) Poisson - that was the simple one and you found it! Good. b) int_uniform. Your version was as follows: `sample.int(size=5, n=nrow(x=survey), replace=TRUE) `
> 
> I assume you tried it and found it didn't work. Then look at the reference documentation on that command to make the correction - which I thought I had mentioned above. If it did work for you, it is because your data frame was of length a multiple of 5 and you should have noticed that the values keep repeating. It gave an error for me, because the worksheet was of length 36. Changing to `sample.int(n=5, size=nrow(x=survey), replace=TRUE)` c) Now can you change the sample key?

Thank you @rdstern, i have made the changes to `uni_integer` and `sample`

rdstern:
@derekagorhom you are not learning to work as carefully as I hoped.  Did you really try the sample key?
a) Here is the documentation on the sample function.  I had hoped you would have looked for this?

image

Can you see what is wrong in your code now?
b) And I would then like to simply insert the column to sample from. But the cursor is at the end.  It should be after the x =  (so not the n = !  And before the comma.

derekagorhom:
hello @rdstern, i have made the changes, can you please check

rdstern (dismissed):
@lloyddewit I am at last ok with this.  So over to you.

instat/ucrCalculator.vb
```diff
@@ -306,6 +306,18 @@ Public Class ucrCalculator
...
+        ttCalculator.SetToolTip(cmdbinomial, "Random binomial sample with values between 0 and 3. Use say rbinom(n,5,prob=0.1) for other distributions")
+        ttCalculator.SetToolTip(cmdpoisson, "Random Poisson sample with mean 1. Change the mean as required.")
+        ttCalculator.SetToolTip(cmdnbinomial, "Random geometric sample as given, i.e. number of failures before size=1 success. Change value of size (must remain positive) for other negative binomials.")
+        ttCalculator.SetToolTip(cmdRan_gamma, "Random sample with shape=1 is from the exponential distribution with mean 2. Keep scale (>0) = 2 and change shape (>0) to 0.5 for chi-square distribution with 1d.f. or to 5, for 10 d.f. Keep scale = 2 for chi-square")
+        ttCalculator.SetToolTip(cmdRan_beta, "This special case is the uniform distribution between 0 and 1. Change shape1 > 0 and shape2 > 0 for different beta distributions.")
+        ttCalculator.SetToolTip(cmdRan_sample, "Random sample with replacement, from a given variable. Change replace to FALSE for a random permutation.")
+
+
```
lloyddewit, line 320:
```suggestion
```

instat/ucrCalculator.vb
```diff
@@ -306,6 +306,18 @@ Public Class ucrCalculator
...
+        ttCalculator.SetToolTip(cmdbinomial, "Random binomial sample with values between 0 and 3. Use say rbinom(n,5,prob=0.1) for other distributions")
+        ttCalculator.SetToolTip(cmdpoisson, "Random Poisson sample with mean 1. Change the mean as required.")
+        ttCalculator.SetToolTip(cmdnbinomial, "Random geometric sample as given, i.e. number of failures before size=1 success. Change value of size (must remain positive) for other negative binomials.")
+        ttCalculator.SetToolTip(cmdRan_gamma, "Random sample with shape=1 is from the exponential distribution with mean 2. Keep scale (>0) = 2 and change shape (>0) to 0.5 for chi-square distribution with 1d.f. or to 5, for 10 d.f. Keep scale = 2 for chi-square")
+        ttCalculator.SetToolTip(cmdRan_beta, "This special case is the uniform distribution between 0 and 1. Change shape1 > 0 and shape2 > 0 for different beta distributions.")
+        ttCalculator.SetToolTip(cmdRan_sample, "Random sample with replacement, from a given variable. Change replace to FALSE for a random permutation.")
```
lloyddewit, line 318:
```suggestion
        ttCalculator.SetToolTip(cmduniform, "Random sample from the uniform distribution between 0 & 1. Use say runif(n,5,10) to change the range")
        ttCalculator.SetToolTip(cmdRan_normal, "Random sample from the standard normal distribution. Use, say rnorm(n,100,15) to change the mean and sd.")
        ttCalculator.SetToolTip(cmduni_integer, "Random integers between 1 and 5. Use say sample.int(3,n,TRUE,prob=c(6,3,1)) to sample 1 to 3 with defined probabilities.")
        ttCalculator.SetToolTip(cmdbernoulli, "Random Bernoulli (0 or 1) sample. Use say rbinom(n,1,1/6) for a random sample of a given dice value.")
        ttCalculator.SetToolTip(cmdbinomial, "Random binomial sample with values between 0 and 3. Use say rbinom(n,5,prob=0.1) for other distributions".)
        ttCalculator.SetToolTip(cmdpoisson, "Random Poisson sample with mean 1. Change the mean as required.")
        ttCalculator.SetToolTip(cmdnbinomial, "Random geometric sample as given, i.e. number of failures before size=1 success. Change value of size (must remain positive) for other negative binomials.")
        ttCalculator.SetToolTip(cmdRan_gamma, "Random sample with shape=1 is from the exponential distribution with mean 2. Keep scale (>0) = 2 and change shape (>0) to 0.5 for chi-square distribution with 1d.f. or to 5, for 10 d.f. Keep scale = 2 for chi-square.")
        ttCalculator.SetToolTip(cmdRan_beta, "This special case is the uniform distribution between 0 and 1. Change shape1 > 0 and shape2 > 0 for different beta distributions.")
        ttCalculator.SetToolTip(cmdRan_sample, "Random sample with replacement, from a given variable. Change replace to FALSE for a random permutation.")
```

lloyddewit (changes requested):

Issue #7852, opened 2022-09-07
Follow up changes after PR #7782 is merged


Patowhiz:
@rdstern @lloyddewit @africanmathsinitiative/developers 
PR #7782 makes several underlying changes that require some dialogs implementations to be changed.
The following are the dialogs that will have to be modified after the PR is merged.

**Hight priority**
1. View object and View Graph dialog - @rdstern I propose the 2 dialogs to now be merged into one. We now have the underlying R and VB.Net implementation that allows this be easily done. Then also the concept of display options can now be changed. I assign myself this. We could have 1 dialog called view objects, capable of viewing; graphs, tables, models, structures and summaries.
2. Renaming, reordering and deleting of objects -  these dialogs and the underlying R code also needs to be refactored.
3. All dialogs that produce tables - they now have to supply the `object_format` depending on the table format. I suggest @EstherNjeriLiberatta  to do this.
4. All dialogs that produce models - they now have to supply the `object_format`. I suggest @Vitalis95 to do this.
5. All dialogs that produce text summaries - they can now specify the `object_format` this in their SetAssign. If @rdstern thinks it's a great idea to introduce the discussed aspects of saving these summaries as objects too, the dialogs can now have a save control. I suggest @anastasia-mbithe to do this. This promotes reproducibility aspects that we have with other objects
6. All dialogs that produce survival objects - From our discussions with @lilyclements , these are now structured objects which are in text formats. So the save control should adopt the 2 specifications.

**Low priority**
1. All dialogs that produce a graph. They just need to be amended to explicitly pass the `object_format`. This will make them consistent with the other dialogs. I suggest John to do this. 

Once the above is all done, we can simplify and optimise our R link and output implementations even more.
Note the above can be done once PR#7782 is merged.

Patowhiz:
In addition, as mention in PR#7782, **Describe > Two/Three variables > Summarise** dialog needs discussions with @rdstern on how we can produce 1 table at a time. Currently it produces 2 html tables in 1 command. I suggest @N-thony to look at this. He may wish to delegate too.

N-thony:
@Patowhiz as now PR #7853, we can have a quick discussion about everything above and then I will enable the team how can do it systematically.

N-thony:
> @Patowhiz as now PR #7853, we can have a quick discussion about everything above and then I will enable the team how can do it systematically.

@Patowhiz could we have this chat today? Let me know when you are available.

[...]

PR #7992, opened 2022-11-29
Refactors model dialogs to use the new output system

Patowhiz:
Fixes some items in issue #7852

### DIFF
--- a/instat/clsRCodeStructure.vb
+++ b/instat/clsRCodeStructure.vb
@@ -507,12 +507,20 @@ Public Class RCodeStructure
                     clsGetRObject.AddParameter("data_name", Chr(34) & _strDataFrameNameToAddAssignToObject & Chr(34))
                 End If
 
+                clsGetRObject.AddParameter("object_name", Chr(34) & _strAssignToName & Chr(34))
+
                 clsAddRObject.AddParameter("object_name", Chr(34) & _strAssignToName & Chr(34))
                 clsAddRObject.AddParameter("object_type_label", Chr(34) & _strAssignToObjectTypeLabel & Chr(34))
                 clsAddRObject.AddParameter("object_format", Chr(34) & _strAssignToObjectFormat & Chr(34))
-                clsAddRObject.AddParameter("object", _strAssignToObject)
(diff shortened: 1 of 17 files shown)


rdstern:
@Patowhiz so far so good.
I have quickly tried Fit Model 2 variables, and 3 variables and general with no problem.  Also Climatic > Extremes quickly.  All run ok so far.  Well done.

One variable the General case seems to run fine and I tried a Test and that was ok.
But the estimate t or normal gives me this error - I think not serious, as I explain below:

image

image

When I look at the results, they seem ok, but I don't have the R script in the output window.  Or the comment.  So that must be where the problem is?

rdstern:
@Patowhiz I have been trying more things and (at last) found a real problem to fix!  A different one.
I think it is caused by using plotly when it can't be used - and perhaps doesn't know?
a) Open dodoma.  b) Do an inventory plot of the 4 variables. c) All the graphics options work except plotly.  I don't mind not having plotly, but it then freezes up and I have to break out of R-Instat.

Incidentally I also tried plotly for a combined graph on the describe.  It didn't give it - which it did for facets.  But it didn't freeze then?

Patowhiz:
@rdstern it hangs because of the R process in the background taking too long to respond back. 
I tried this in R studio too, the R process hanged. 

I'm not sure why it also takes so long, probably because of the details of the plot. 
I tested the plotly option inventory plot with a smaller dataset of 100 rows. It worked fine. 

We can open a separate issue on ggplots that take long for plotly to render and the discuss with @dannyparsons @lilyclements on what we should do for such plots.

Patowhiz:
The bug you have encountered in the 1 variable fit is an existing one. Let me look into it.

rdstern:
@Patowhiz thanks for those replies.  Neither of them stop the new version being used on Monday.

lloyddewit (dismissed):
@Patowhiz Thank you for the changes.
I am not expert on the R-Instat assignment but the code looks good to me.
No action needed from your side. I may do some small formatting things. After that, I will go ahead and merge.

PR #7995, opened 2022-12-01
Improved consistency/readability of Describe menu

### LINKED ISSUE #7986: Possible menu changes in the Describe menu
@lilyclements I would like to check when we meet on Monday, before any implementation.  I would also like @N-thony view, because he is pretty good on design!  Once agreed, then perhaps he can suggest who could implement.
@lloyddewit you may have other ideas and the changes may not be quite so trivial, because I am suggesting small wording changes and we need to remember the language translation implications.  
Here is the One Variable:
image

I am suggesting that is ok.
Here is the 2/3 variable:
image

a) I suggest the Pivot Table should move to the top and have a line underneath, just like the Visualise Data  in the 1 Variable.  I suggest it can be a really important feature in R-Instat.  It is currently a hidden-jewel.  We need to document it and have a video, see #7985.  I don't think it quite warrants moving to the main menu, but having it at the top will give it more prominence - as we wish.
b) Should we also extend the name, because it has tables and graphs?  So we could still leave the name as it is.  Or we could change it to `Pivot Table/Graph`.  Or we could follow the pivot-table package word and so call it `Pivot Table/Chart`.  I am using chart essentially as a synonym of graph.  This might however be confusing.  What do you think between a) Leaving it at Pivot Table, or Table/Graph or Table/Chart?
c) While we are making that change I would like to simplify, and consider separating the next item

image

I suggest deleting the word Specific and having 2 separate entries, namely `Tables `and then `Graphs`.  The Tables will currently go to just the single item.  But later we may have Multiple Responses, so that may be a second.  
The Graph will then go to the set of graphs as now?

derekagorhom:
fixes #7986 
@N-thony this PR is ready for review


N-thony (approved):
@derekagorhom looks good, just a small comment if @rdstern agree you could also change 
`Tables > Frequency...`?
@rdstern over to you

derekagorhom:
@N-thony i have made the changes

rdstern (changes requested):
@derekagorhom looks good.  Just 2 small changes, then good to merge:
a)  The Describe > Tables > Frequency... should become Frequency/Summary...
And a new one:
b) Could you swap the Tables/Graphs menu items, so `Graphs` comes before `Tables`.

I checked with Danny and our reasoning is that with Tables having (now) just a single dialogue - it seems a bit feeble, so better to have graphs first.  (Later there may be a second one for multiple response.  If that is a third option in the existing Tables, then Tables will become Tables...

And, for reference only, we have discussed the Two/Three Variables menu item again, because there are some instances when there can be four variables.  But I would like to keep it at the Two/Three, title because it is mainly 2 variables, but you need 3 for a summary table.  We do sometimes occasionally have 4, but I would like to keep the idea that this menu is still to make these tables/graphs pretty simple, and when you want to be general, then you go to the many - namely the Graphs and Tables dialogues.

We need to mention three, mainly because you need 3 variables for a 2-way summary table.

derekagorhom:
Hello @rdstern I have made the changes.

rdstern:
@derekagorhom good progress.
You missed the new suggestion to make the Describe > Tables > Frequency?Summary Tables... into just `Describe > Tables...  `
Then it would be great to merge as soon as possible.  It makes the menu much better.

rdstern (changes requested):
see above

derekagorhom:
> @derekagorhom good progress. You missed the new suggestion to make the Describe > Tables > Frequency?Summary Tables... into just `Describe > Tables... ` Then it would be great to merge as soon as possible. It makes the menu much better.

Sorry about that, i just made the changes. 
can you have a loo at it.

rdstern:
@derekagorhom you continue to be quick and conscientious, but not careful enough!  We need the` ...` after the new Tables to signify you go straight into the dialogue.

rdstern (approved):
Great!  A good one to merge as soon as possible!  But not necessarily for today, though would do no harm!

lloyddewit (approved):

lloyddewit:
@rdstern To minimise regression risk, I will merge after the build planned for today

PR #7999, opened 2022-12-02
Improvements to the Describe > Two variable Summarise Dialog

### LINKED ISSUE #7262: Further improve the Describe > Two Variable Summarise dialogue
This work may need to wait for the skimr implementation on the same dialogue, though the proposed changes are separate.
a) There is (I think) already an issue for @Ivanluv to improve the option categorical by categorical, using the new tabulation facilities.  This will produce a series of separate 2-way tables, with one for each entry in the multiple receiver.
b) So we could consider sets of 3-way tables, if a further single receiver were a factor variable.  
c) If the third receiver were a numeric variable (more urgent).  Then the label will change to summary tables, and produce a set of 2-way summary tables.
d) The current option of numeric (multiple) by factor (single) produces a list of summaries.  This should be updated also to the new tabulation system.  Is there scope then for a third receiver, presumably another factor?
e) (Maybe for later?) The current option of factor(s) by numeric gives a one-way ANOVA table.  Less urgent, but there is a nice opportunity here to have a second factor, for a two-way ANOVA.  That would then be like the 3-way modelling, (special case) and we could then allow parallel lines or Interaction.  Maybe we could also have a graphics option here, with the new parallel geom and the ordinary plot, if interaction.  (We might even have a checkbox (default unchecked) for Add p-values.  
f) Might we consider other possibilities for glms, if the multiple variable is a factor (or a set of factors with all say 2 levels), or perhaps all having 5 levels (all ordered categorical?)  Perhaps they could ve a property of a select?  Currently this is the Frequency tables option.  That would become a pull-down with Analysis of Deviance as another option.

anastasia-mbithe:
Fixes #7262 

This is in progress.
I have added the new tabulation system when we have Numeric by Categorical variables.
@rdstern, the third receiver mentioned in points; b, c, d, and e, is it to be added under the two variable options?


rdstern (changes requested):
@anastasia-mbithe Don't worry about b), c), d) e) for now.  They are for the 3-way option.  The dialogue looks fine.  But it only did the count, when the default summaries button suggested many more summaries.  Perhaps make the default into count non-missing and mean and then execute these.

The output is to the browser while this has now moved on in many dialogues to use the output window.  Perhaps check with @Patowhiz whether anything can be done in this pull request, or we should wait till it is merged?

anastasia-mbithe:
@Patowhiz, I managed to have the output in the output window in Html format. however, the output is given as a single table as opposed to producing different tables depending on the variables in the multiple receiver. Initially, two variables in the multiple receiver gave an output of two tables.
Does the new format support production of different tables in the output window, from one chunk of code?
image

rdstern:
@anastasia-mbithe you were right to have the count missing, plus the same summaries that are given by the usual summary dialogue.  I had forgotten that!

rdstern:
@anastasia-mbithe while you are waiting for @Patowhiz and/or @ChrisMarsh82 I assume on extending the html for the summarise option.  This currently still goes to a browser for the multiple 2-way tables.  We now have our maximised window and I am hoping that can be improved soon to allow multiple maximised windows anyway?

This might then be an example that goes straight to the maximised window, rather than first into the output window.

Meantime could you look at the skim option:  Here is output for the usual survey data:

image

This has come from the following dialogue:

image

Now the code - with an extra line by me - that didn't work!

```
# Code generated by the dialog, Describe Two Variables

skimr::sfl(n_missing="n_miss", complete_rate=NULL) 
survey %>% dplyr::group_by(village, variety) %>% skimr::skim_without_charts(c("village","field","size"))
rm(survey)

```

a) So please impove on my feeble level attempt at R.  It appears that there is an sfl function in skimr where you can omit summaries or change the labels.  I wanted n_missing to become n_miss and then not to bother with the complete rate summary.
b) I was doing this partly so that all the summaries would come on one line.  There seems also to be a width possibility to organise this?
c) And could we now maybe reinstate the sparkline histograms - so change skim_without_charts into just skim?  Perhaps you will have to use the html output then?

This would make it much more attractive.  I hope your R works for all this?  If not, then perhaps Vitalis or Lily could help.

rdstern:
@anastasia-mbithe I assume you will be moving on to this dialog soon.  It is another where the skim may have fixed itself, or could do with a small change to all get onto a single line, per element.
And I assume you and @lilyclements will also now take over #7978 from @Vitalis95 at the same time?

rdstern:
Sorry @anastasia-mbithe but I get build errors

anastasia-mbithe:
This is not ready for testing.

rdstern:
@anastasia-mbithe I am really pleased you have now progressed to examining the improvements needed for the 2-variable (and perhaps 3-variable) summaries.  There is quite a lot to consider here:
a) Perhaps a quick win can be to nail the use of skim as mentioned above.  
b) Then let's start with the 2 variable summary stuff.  That is for 2 separate situations.  The first is numeric by categorical:

image

Here is the starting point, namely the result in the current R-Instat:

image

This used just the summary command.  The change was to use the new tabulation system.  Of course that would have put the result in a browser, and also uses mmtable2.  

I assume, unlike the option, below (where it is categorical by categorical) that the dialogue above will still produce a single table.  So a trick will be to add the factor to the one-variable dialogue situation that you have just completed - possibly in different ways?

c) The other option is factor by factor, currently here:

image

I suggest this is a really important sutuation, because it is so commonly used.  Currently it produces a list of tables and in the browser and uses mmtable2!  So lots needs to change.

1) I suggest we keep the multiple receiver (because it is used for the other situations.  But just process the first variable (at least for now.) So we only produce a single html table.  We may add a tooltip to explain this limitation.
2) But please discuss, and consider whether you could allow multiple "frequencies".  This needs discussion with @lilyclements , and I suggest it may also apply to the general frequency table situation?  We may decide not, because the freuencies dialogue also does that?  Here is an example:

image

And here are the results:

image

Aaah my suggestion for your 2-way frequencies is that we don't allow multiple percentages.  But, if easy, then we do allow Count as well as one percentage.  It is very rare (and a bit confusing) to have more - except for teaching.  (And, if that is difficult to code, then we might allow it only in the general situation.

d) Moving to 3-variables.  The main importance of that is as 2-way summary tables.

anastasia-mbithe:
@rdstern, I have made some changes to the two-variable and three-variable options, Please test.

rdstern:
This is really coming on now.  I start with this one:

image

And here is the new One Variable Summarise for comparison:

image

a) I like the layout of the `Choose Summaries` and then the `Order` control and then the `Format Table` control in the One Variable dialog.  The Order is very big, and could perhaps be reduced in both width and Depth generally.  That would be in each dialog where it is used.
b) I also prefer the layout of the Missing in the One Variable.
c) In the 2 variable above there is still the Percentages, which is not needed I think?
d) In the Margin control do we still need the Label `All` still to be on the main dialog.  I am assuming we will soon have a tab in the Format Table sub-dialog to be able to rename the Variables and I assume it could include a name for the Margin. there?
e) On the left it says numeric by categorical.  They are not lined up vertically.  Please could they be.  And I wonder why we are using categorical rather than factor - should we change it to numeric by factor?
f) Do we need the display significan figures on the main dialog?  Won't that come in the Format Table sub-dialog?
g) Finally the layout.  There are now 3 options, namely Summary, Variables and Factor, for the Columns.  Then, unless there is only a single summary variable chosen, this is a 3-way table.  So we can also choose which of the 2 row variables is first. So perhaps we shoiuld also have a checkbox underneath (perhaps still in the Columns: box and saying `Swap Row Variables`.

rdstern:
Now here are comments about the 2 variable Factors by Factor.:  Here is the new dialog:

image

And here are the current results:

image

That needs to be changed.

The "old results" - and from the current version, which also need to change urgently are:

image

The old results are the sort of thing we want, but they are in a browser and producing multiple tabs there.  

My first is a @lilyclements and @anastasia-mbithe question, namely can we provide anything with the dialog as shown above, or do we only process the first variable in the multiple receiver - then giving a nice set of options for the sort of 2-way tables we need?

If we can do anything with multiple elements then the columns would be the levels of the second variable, so New, Old, Trad., in the example above.   The rows would then be for each of the first variables in turn:  So the 2-way frequencies one-on top of the other.  Something like this:
                            New   Old    Trad
Village  Kesen      0          3        4
Village  Nando    2           7        5
Village  Niko       0           2         3
Village  Sabey     2           5         3
Fertgrp   0.5-2     0           6         4
Fertgrp    >2        4           10       3
Fertgrp      0         0           1        8
etc 

So a set of 2-way tables one on top of the other, and without margins - not a 3-way table!

Is that easy to do?  If so, then that would be someting with the full dialog as provided now..

In that case we could have an additional section of the dialog for 2-way table options, that is disabled when there are multiple factors in the first receiver.  If the above is difficult or not possible, then we just process the first variable in the multiple receiver for now.

There are then some similar points to make as above, but let's check this point first.

Then I am keen to merge these improvements for 2 variables as soon as possible - with the 3-way disabled.  Then we blast away on the 3-way.

rdstern (commented):
@anastasia-mbithe the 2-variable is looking good now, except for the factor(s) by factor, which is a frequency table and not a summary table.  So you want the frequencies and possibly percentages - which we had before.  And then margins as an additional option.

I was hoping before we could include multiple percentages and frequencies in the same table, but we already have that in the Describe > 2/3 way > Frequenies dialogue.  So if we return to having to choose just one of them, or perhaps allowing frequencies plus a choice of a single percentage as the maximum - then that's sufficient.  It is very rare that 2 different percentages are equally important.

rdstern:
@anastasia-mbithe I am relieved you are working on this, but I am not quite sure why this option is causing you all these problems? 
I am really keen that you "nail" this, so we can also have the 3-way summaries in the forthcoming release.

Remember the Factor by factor option should be a really simple way to get frequency tables.  So the dialog should be a simpler and easier way of your Describe > Tables > Frequency dialog, when there are just 2-way tables.  (And when all 3 are factors, then it is a simple way of doing that too.

Here is the general dialog for frequencies for 2-way tables:

image

And here is the result:

image

Now here is your 2-way dialog for the same factors:

image

Wow - that is looking more complicated than the general case above.  Remember this is supposed to be making 2-way frequency tables really easy for beginner users.  It is very popular.

And the result is this:
image

I hope you see that I am a bit confused that you have got to here with this dialog.  Up to now your work has always been so good?

Anyway I think most of the changes should be easy to do.  I hope you can do them today?

a) In your 2-variable case, the second variable is always the columns.   So that whole box is not needed.
b) You don't need the Choose Summaries button.  (You are not choosing summaries.  These are frequency tables!)
c) Move the Display Margins control so it is in the same place on the dialog as it is, on the general Frequency tables dialog.  We should be as consistent as we can.  This will help users.
d) The factor by factor option can then be a considerably smaller dialog.
e) Get (roughly) the same results - at least the same layout as you can with the general Frequencies dialog.  (I don't like the Summaries Count column there - but that's another story.)

And please ask for more help if you need it.
 
d) Get the right results

rdstern:
@anastasia-mbithe good to have you back - again.
But, I hope I am using your latest version - I don't see many changes:
image

You have not got rid of the Columns box.  That was my point a) above, i.e.
"a) In your 2-variable case, the second variable is always the columns. So that whole box is not needed."

You are still producing the wrong table.  This should be a 2-way table and it still isn't.

I am very puzzled.  You were so promising as an R-Instat programmer.  And you are doing great stuff on the graphs - I really like your improvements on the line graphs.  But you seem to be making very "heavy weather" on these tables.  I don't think you are enjoying it either?  And I was hoping that this part would be finished really quickly - even for 3 variables - so you could return the the Format Tables sub-dialog.  That's big and really important.  There's lots more that you could be doing.  Should we be allocating the tables stuff to someone else?

anastasia-mbithe:
> @anastasia-mbithe good to have you back - again. But, I hope I am using your latest version - I don't see many changes: image
> 
> You have not got rid of the Columns box. That was my point a) above, i.e. "a) In your 2-variable case, the second variable is always the columns. So that whole box is not needed."
> 
> You are still producing the wrong table. This should be a 2-way table and it still isn't.
> 
> I am very puzzled. You were so promising as an R-Instat programmer. And you are doing great stuff on the graphs - I really like your improvements on the line graphs. But you seem to be making very "heavy weather" on these tables. I don't think you are enjoying it either? And I was hoping that this part would be finished really quickly - even for 3 variables - so you could return the the Format Tables sub-dialog. That's big and really important. There's lots more that you could be doing. Should we be allocating the tables stuff to someone else?


Hi @rdstern , 

I tried to get the same layout as for the General Tables (Frequency tables). 
And as for the Column options, are you sure we should get rid of them since its part of gt flexibility that allows one to define what you want to set as your column factor? We can have no column factor, have the summary-variable or the Variable to be the column factor. And also for consistency with all gt tables.

rdstern:
Hi @anastasia-mbithe you write:

Hi @rdstern ,

I tried to get the same layout as for the General Tables (Frequency tables).
And as for the Column options, are you sure we should get rid of them since its part of gt flexibility that allows one to define what you want to set as your column factor? We can have no column factor, have the summary-variable or the Variable to be the column factor. And also for consistency with all gt tables.

That's an excellent question.
My reasoning is that you always have the 2 factors - here village and variety.  
And the simple analysis will then be to have them as a 2-way table, i.e. a rectangle.  
So the only question is which is the columns and which is the rows?
The user can dictate that by the order they include them, so which is first and which is second.  

And it is a frequency table.  It is an example of a summary, but a very specific one.  The only real decision is whether we present the frequencies or a percentage - and (if so, then) which percentage?  And you have that.

anastasia-mbithe:
@rdstern, 
image

Is this a two-way table?
Screenshot (327)
What of this?

anastasia-mbithe:
@rdstern, nice.
image
This is from the "no column option"

image
This is from the summary-variable option.

image
And this is from the Factor Variable option.
So, the first option is not a 2-way, but the rest are.

rdstern:
@anastasia-mbithe you have complicated your life by adding a third variable and none of those options are what is needed.  For village and variety we need the two-way information shoing the counts (or percentages) of data in each village for each variety.  So start with the data we need.  I use the Frequency tables dialog to show you:

image

Here are the results:

image

These are the correct results in a one-way layout:  **_We don't need those results from your 2 variable dialog._**

If I change to `summary variable `in that dialog I get the same thing, because there is only one summary variable!  _**We don't need that option in your 2-way dialog either.**_

If I change to `Factor variable `then I get:

image

At last I have the 2 variables in a 2-way layout.  **_We do want to be able to get that!_**

Or I could get:

image

which is the other way round.  **_We do want to get that as well._** 

I hope that is now clear.  Let's have a call if not.

And we can both of the 2-way layouts by choosing which is the first variable and which is the second.  So we don't need any of the set of options in the 2 variable dialog.  And that will make the dialog much simpler and smaller.

And check with your presentation above and you will also see that you are just not getting the 2-way information as all.  You have a one-way table, sometimes repeated multiple times.  

Look at the last example below where I have added the margins.  

image


I hope you see that, in all your examples you are just giving the 17, 15, and 4 values.  That's just oneway information.

rdstern:
@N-thony I have found another (two?) problems with the existing version of this dialog - that remains here, and which should be easy to fix.

image

a) The receiver has the scroll bar at the bottom.  That need not be there.
b) It wrongly says I can't have the 2 factors here.  This example is with the diamonds data and it seems to be because they are ordered factors.  It is ok if I make them ordinary factors.

I am keen - especially given the above points, that the new version should include an updated version of this dialog.  I wonder who could take it over?

Issue #7936, opened 2022-10-27
Add facilities for tabulating and summarising multiple response questions


rdstern:
We do need to add facilities for processing multiple response questions in survey-type data.  They are in SPSS and are very common.  There is little mentioned in R.  Here is one useful entry on [stack-overflow](https://stackoverflow.com/questions/9265003/analysis-of-multiple-response).

This mentions the UserFriendlyScience package as providing a route.  That has been removed from CRAN, because the Authors say it is being updated and split into different parts.  However, the ufs package is one of these updates.  It is currently available and includes a multiresponse function, which may be what we need.

I assume this could be a new dialogue, but it may be an additional button on the existing dialogue. There are currently frequency and summary buttons so this could be a multiple button?

At the same time we have tables and graphs on the same menu option.  I suggest we consider splitting them into two main menu items - one for tables and one for graphs?  So we have `Tables` and then `Specific Graphs`. 

Perhaps @lilyclements could first check whether ufs is suitable.  Then it should be reasonable simple to implement.

rdstern:
@anastasia-mbithe you are going to be improving this dialogue.  

Could you quickly change the top of the dialogue, which currently looks like this.

![image](https://user-images.githubusercontent.com/11226469/205509465-1d0f0151-618f-4722-ad83-9d40eda71c8d.png)

a) Change Frequency Table to just `Frequency` and Summary Table to just `Summary`.
b) Add a 3rd button saying `Multiple Response`. Disable this option and add a tooltip saying `Not Yet Available`.  (The reason for the tooltip is that we don't want users thinking there is something they may be able to do - perhaps in Tools > Options  that would make it available.)

In the pull request remember it only (partially) resolves this issue!

Thanks

rdstern:
@lilyclements I went to this topic hoping I could suggest some quick progress.  Instead I now suggest we make a big deal of it, but then we leave it for a bit and attack in stages.

The actual problem is quite simple.  You have respondents and also responses (for example Do you have sheep, chickens, goats, cows.  So you may have 4 columns with Yes No for each.  (That''s multip-le dichotomy.  Or farmers with multiple fields and you ask what is the main crop on Field 1, Field 2, etc, up to the number of fields owned by that farmer.  
Then you could ask for the number of respondents (farmers) that have maize.  Or what's the percentage of fields (responses) with maize.
So you can work at the resondent or the response level.

[...]

PR #8002, opened 2022-12-05
Added Multiple Response button in Frequency/Summary Tables

anastasia-mbithe:
Fixes Partially #7936 
@rdstern, this is ready. 
However, I have noticed when the button is not enabled the tooltip is not visible(like in this case).


rdstern (approved):
@anastasia-mbithe looks fine.  Thanks.
I am happy to see it merged without the tooltip working.

lloyddewit (approved):

PR #8005, opened 2022-12-05
Added Pair Plot Options sub dialog for Legend on  Graphs dialog

### LINKED ISSUE #7212: Add ggpairs to the Describe > Two Variables > Graphs dialogue
I suggest this is a clear "low-hanging" fruit.

This addition will be a new radio button at the top called Pairs.  I don't then have a good name for what is there now, which could be perhaps "By".

In the current dialogue there is a multi0ple receiver, plus a single receiver.  And the multiple rece3iver can only have variables of one type.
 The Pairs just looks at variables 2 at a time.  So, there is just the same data control plus a multiple receiver.  And it runs the GGally::ggpairs command.  The multiple receiver can take at least factors or numeric, so start with no restrictions.  Let's explore what happens with dates, character, logical, etc.

We already use this command with numeric variables in the Describe > Multivariate > Correlations dialogue.  Here we will include a variety of options, some on the dialogue and others on an Options sub-dialogue.  It is a very rich resource!

MeSophie:
Fixes #7212 
@rdstern  I added  legend on Graphs dialog for Pairs radio button.
@N-thony Please can you have a look.


rdstern:
@MeSophie the Options button On Describe > Two/Three Variables > Graph > Pairs is still disabled.  So I have not been able to check.

N-thony:
> @MeSophie the Options button On Describe > Two/Three Variables > Graph > Pairs is still disabled. So I have not been able to check.

@rdstern I checked this and it seems to work fine
image
@MeSophie I noticed now we lost the `Plot Options` button for the `By` option, but `Pair Plot Option` seems to work fine
image
image

N-thony:
@anastasia-mbithe can you peer-review this?

MeSophie:
@rdstern the Plot Options is visible for By now.

rdstern:
@MeSophie this is already looking nice.
a) I noticed that Ok is only enabled when you include the Colour (Optional).  It should be included as soon as the pairs can be used.  Possibly the Options button should only be enabled when the colour is completed?
b) Do we need 2 controls in the sub-dialogue?  I am happy with that, but perhaps Legend could be just a label, with None as the default?  I am happy either way though.

rdstern:
@MeSophie but that's why I would strongly prefer you (instead) activate Ok without needing to add the optional factor.  But disable the Options, or the legend instead!

Ok with the 2 controls.  But I assume now (with your options) you could tick the first one to have a legend, and then set it to none?  That's odd.

MeSophie:
> but that's why I would strongly prefer you (instead) activate Ok without needing to add the optional factor. But disable the Options, or the legend instead!

@rdstern according to @N-thony comment I think the dialog and sub-dialog are better now. I remove the first check box on sub dialog and made Pairs Plot Option sub dialog enable when we fill the Colour Factor.

rdstern (approved):
@MeSophie looks good.  Thanks.  I am approving

lloyddewit (changes requested):
@MeSophie thanks, looks good, just a couple of questions

MeSophie:
> @MeSophie thanks, looks good, just a couple of questions

@lloyddewit I revolved your comments thank you.

rdstern:
@lloyddewit I am not sure what should happen if I click on revert above?

I would like to reopen this pull request, or issue, by @MeSophie .  This is for the following reasons.
a) In Version 0.7.16 the Pairs Plot Options is disabled.  I thought this was the purpose of this pull request.
b) Our main plotting features have regressed again in a small way, and I wonder if this issue/pull request is related.  Namely a useful option in the plotting sub-dialogue is shown here:

image

The default option here used to be None - so to be able to not have a legend.  This has disappeared!

lloyddewit:
> I am not sure what should happen if I click on revert above?

@rdstern This would create a new PR with proposed changes that would undo the specific changes in this PR. 
We don't know for certain what caused the regression. So a full revert may not be the best option. I suggest that we start with the latest master version and investigate the bug. I assume that @MeSophie is the best person for this.

PR #8007, opened 2022-12-06
Ensured that the selected sheet in the Data View is maintained after reordering

### LINKED ISSUE #7550: Minor improvement on reordering of data frame tabs in Data View
Currently the reordering of data frame tab in the Data View works fine except that the current data frame tab is changed by the last data frame after reordering in Data View's tab. What we need to improve is to maintain the current data frame though reordering is applied.

N-thony:
Fixes #7550
@rdstern as discussed I have improved the code and also fixed the small bug in the `Data View` to maintain the selected sheet even though reordering is applied. Please, take a look. Thanks.

### DIFF
--- a/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
@@ -84,7 +84,11 @@ Public MustInherit Class ucrReoGrid
                 lstWorkSheetsFound.Add(fillWorkSheet)
             End If
         Next
-        If lstWorkSheetsFound.Count > 1 Then
+
+        Dim bReorder As Boolean = If(lstWorkSheetsFound.Count > 1, _clsDataBook.DataFrames.Select(Function(x) x.strName).ToList().
+                                                       SequenceEqual(grdData.Worksheets.Select(Function(x) x.Name).ToList()), True)
+
+        If Not bReorder Then
             'reorder the worksheets based on the filled list
             For i As Integer = 0 To lstWorkSheetsFound.Count - 1
                 grdData.MoveWorksheet(lstWorkSheetsFound(i), i)
--- a/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
@@ -85,13 +85,16 @@ Public MustInherit Class ucrReoGrid
             End If
         Next
 
+        'in the second condition we check if all data frames in the data book
+        'have the same order positions with all data frame sheets in the grid
+        'if not this check will return False which means the data frames in the data book are reordeder
         If lstWorkSheetsFound.Count > 1 AndAlso Not _clsDataBook.DataFrames.Select(Function(x) x.strName).ToList().
                                                        SequenceEqual(grdData.Worksheets.Select(Function(x) x.Name).ToList()) Then
             'reorder the worksheets based on the filled list
             For i As Integer = 0 To lstWorkSheetsFound.Count - 1
                 grdData.MoveWorksheet(lstWorkSheetsFound(i), i)
             Next
-            grdData.CurrentWorksheet = grdData.GetWorksheetByName(strCurrWorksheet)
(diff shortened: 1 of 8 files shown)


rdstern (approved):
@N-thony I am really pleased you have returned to this issue, because sorting out the reordering of the data frames was one of the last of these operations that seems now to be working.
It seems to be fine, so if @lloyddewit is ok, then I would be happy to merge it.  
I am not sure the title of the pull request is totally appropriate though?  When I tried, the sheets were reordered as I had wished, and the last sheet always became the active sheet.  That's fine by me, but it isn't what I think of as "maintaining the selected sheet".

N-thony:
> @N-thony I am really pleased you have returned to this issue, because sorting out the reordering of the data frames was one of the last of these operations that seems now to be working. It seems to be fine, so if @lloyddewit is ok, then I would be happy to merge it. I am not sure the title of the pull request is totally appropriate though? When I tried, the sheets were reordered as I had wished, and the last sheet always became the active sheet. That's fine by me, but it isn't what I think of as "maintaining the selected sheet".

@rdstern sounds good, it works fine on my side, I have sent some screenshots on Skype

instat/ucrColumnMetadata.vb
```diff
@@ -63,6 +63,7 @@ Public Class ucrColumnMetadata
 
     Private Sub AddAndUpdateWorksheets()
         Dim firstAddedWorksheet As clsWorksheetAdapter = Nothing
+        Dim strCurrWorksheet As String = If(_grid.CurrentWorksheet Is Nothing, Nothing, _grid.CurrentWorksheet.Name)
```
lloyddewit, line 66: move to just above line 77?

instat/ucrColumnMetadata.vb
```diff
@@ -73,6 +74,7 @@ Public Class ucrColumnMetadata
             End If
             RefreshWorksheet(worksheet, clsDataFrame)
         Next
+        _grid.ReOrderWorksheets(strCurrWorksheet)
```
lloyddewit, line 77: if `strCurrWorksheet` is `Nothing`, then should we still call this sub?
Unfortunately, `ReOrderWorksheets()` has no summary comment so it's hard to understand what will happen.
Please could you investigate and comment if needed?

instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
```diff
@@ -71,7 +71,7 @@ Public MustInherit Class ucrReoGrid
         Return New clsWorksheetAdapter(fillWorkSheet)
     End Function
 
-    Private Sub ReOrderWorksheets() Implements IGrid.ReOrderWorksheets
+    Private Sub ReOrderWorksheets(strCurrWorksheet As String) Implements IGrid.ReOrderWorksheets
```
lloyddewit, line 74: Please could you add a summary header (see comment below)?

lloyddewit (changes requested):
@N-thony Thanks, looks good, just a few open questions.

instat/ucrColumnMetadata.vb
N-thony: anytime there is a change in the data frame we load it back from R to the Data View, in `AddAndUpdateWorksheets()` there is a sub on line 75 we always load the data from R and set it as a selected/current dataframe in the Data View, so to the keep the selected dataframe active before reordering, I'm using this temp variable `strCurrWorksheet ` on line 66

N-thony:
@lloyddewit I have made the changes, please take a look. Thanks.

rdstern (dismissed):
@N-thony seems to work fine now!

lloyddewit:
@N-thony Thank you for the changes. There is just one open comment above. If you can add the summary header, then we can merge, thanks

instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
N-thony: @lloyddewit this is done

lloyddewit (approved):
@N-thony Thank you for the final changes.
The new header is the same as the sub name, so it doesn't add any value.
I was hoping that the header would explain more about the what the sub does. However, I looked through the subs again and with the new code comments you added, it's easier for a developer to understand what's happening. So I'm happy to approve/merge.

PR #8011, opened 2022-12-09
Added ability to show loading screen that keeps responsive from anywhere

ChrisMarsh82:
This is not used anywhere at the moment but the plan is to take code out of R-script that currently shows this screen. 
I have added a timeout on the screen as the code relies on hideLoadingScreen being called.

### DIFF
--- a/instat/frmSetupLoading.vb
+++ b/instat/frmSetupLoading.vb
@@ -19,16 +19,27 @@ Imports System.Threading
 Public Class frmSetupLoading
     Private dctMessagesLinks As Dictionary(Of String, String)
     Private iSelectedMessage As Integer
-
     Private Shared CancellationTokenSource As CancellationTokenSource
+
+    ''' <summary>
+    ''' Display a loading screen in seperate flag in order to keep the
+    ''' loading screens responsiveness
+    ''' </summary>
     Public Shared Sub ShowLoadingScreen()
         If CancellationTokenSource Is Nothing Then
             HideLoadingScreen()
         End If
         CancellationTokenSource = New CancellationTokenSource
-        'Ignore the warning - we only want to keep this dialog responsive
+        'Ignore the warning - we only want to keep this dialog responsive it doesn't mater
+        'that its not awaited 
+#Disable Warning BC42358 ' Because this call is not awaited, execution of the current method continues before the call is completed
         ShowLoadingScreenAsync(CancellationTokenSource.Token)
+#Enable Warning BC42358 ' Because this call is not awaited, execution of the current method continues before the call is completed
     End Sub
+
+    ''' <summary>
+    ''' Hide currently displayed loading screen
+    ''' </summary>
     Public Shared Sub HideLoadingScreen()
         If CancellationTokenSource IsNot Nothing Then
             CancellationTokenSource.Cancel()
@@ -40,8 +51,8 @@ Public Class frmSetupLoading
         Await Task.Run(Sub() ShowLoadingScreenAndWait(CancellationToken), CancellationToken)
     End Function
     Private Shared Sub ShowLoadingScreenAndWait(CancellationToken As CancellationToken)
-        Dim iWaitSecondsBeforeShow = 2
-        Dim iWaitSecondsBeforeAbort = 30
+        Dim iWaitSecondsBeforeShow As Integer = 2
+        Dim iWaitSecondsBeforeAbort As Integer = 30
         Dim stopwatch As New Stopwatch
         stopwatch.Start()
         'Wait before showing loading screen
--- a/instat/frmSetupLoading.vb
+++ b/instat/frmSetupLoading.vb
@@ -26,7 +26,7 @@ Public Class frmSetupLoading
     ''' loading screens responsiveness
     ''' </summary>
     Public Shared Sub ShowLoadingScreen()
-        If CancellationTokenSource Is Nothing Then
+        If CancellationTokenSource IsNot Nothing Then
             HideLoadingScreen()
         End If
         CancellationTokenSource = New CancellationTokenSource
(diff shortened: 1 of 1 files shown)


instat/frmSetupLoading.vb
```diff
@@ -14,10 +14,51 @@
...
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Imports System.Threading
+
 Public Class frmSetupLoading
     Private dctMessagesLinks As Dictionary(Of String, String)
     Private iSelectedMessage As Integer
 
+    Private Shared CancellationTokenSource As CancellationTokenSource
+    Public Shared Sub ShowLoadingScreen()
```
lloyddewit, line 24: - separate each function with single empty line
- add summarry header to each public function

instat/frmSetupLoading.vb
```diff
@@ -14,10 +14,51 @@
...
+        If CancellationTokenSource Is Nothing Then
```
lloyddewit, line 25: should this be `Not Nothing`?

instat/frmSetupLoading.vb
```diff
@@ -14,10 +14,51 @@
...
+            HideLoadingScreen()
+        End If
+        CancellationTokenSource = New CancellationTokenSource
+        'Ignore the warning - we only want to keep this dialog responsive
```
lloyddewit, line 29: Is the warning a user warning or just an ignored return value from `ShowLoadingScreenAsync()`?
clarify comment?

instat/frmSetupLoading.vb
```diff
@@ -14,10 +14,51 @@
...
+            CancellationTokenSource = Nothing
+        End If
+    End Sub
+    Private Shared Async Function ShowLoadingScreenAsync(CancellationToken As CancellationToken) As Task
+        Await Task.Run(Sub() ShowLoadingScreenAndWait(CancellationToken), CancellationToken)
+    End Function
+    Private Shared Sub ShowLoadingScreenAndWait(CancellationToken As CancellationToken)
+        Dim iWaitSecondsBeforeShow = 2
+        Dim iWaitSecondsBeforeAbort = 30
+        Dim stopwatch As New Stopwatch
```
lloyddewit, line 45: specify data types (e.g. `As Integer`)

instat/frmSetupLoading.vb
```diff
@@ -14,10 +14,51 @@
...
+    Private Shared CancellationTokenSource As CancellationTokenSource
+    Public Shared Sub ShowLoadingScreen()
+        If CancellationTokenSource Is Nothing Then
+            HideLoadingScreen()
+        End If
+        CancellationTokenSource = New CancellationTokenSource
+        'Ignore the warning - we only want to keep this dialog responsive
+        ShowLoadingScreenAsync(CancellationTokenSource.Token)
+    End Sub
+    Public Shared Sub HideLoadingScreen()
```
lloyddewit, line 32: are any explicit steps needed to hide the loading screen?
or will this happen automatically when `CancellationTokenSource` is set to `Nothing`?

instat/frmSetupLoading.vb
```diff
@@ -14,10 +14,51 @@
...
+        If CancellationTokenSource IsNot Nothing Then
+            CancellationTokenSource.Cancel()
+            CancellationTokenSource.Dispose()
+            CancellationTokenSource = Nothing
+        End If
+    End Sub
+    Private Shared Async Function ShowLoadingScreenAsync(CancellationToken As CancellationToken) As Task
```
lloyddewit, line 39: should this function return something?

lloyddewit (changes requested):

lloyddewit:
@ChrisMarsh82 
I wasn't sure if I should review this or not.
I'd prefer to merge this PR when all the code is called and there is some testable functionality. Otherwise I'm reviewing without understanding the context, and we risk having overlooked 'orphan' code in the code base.
I hope that's OK, happy to dicsuss if needed.

instat/frmSetupLoading.vb
ChrisMarsh82: It returns the task that can be run asynchronously. Its considered bad practice to use Sub/void as 'Exceptions thrown in an async void method can’t be caught outside of that method' ref: https://medium.com/@deep_blue_day/long-story-short-async-await-best-practices-in-net-1f39d7d84050

instat/frmSetupLoading.vb
ChrisMarsh82: The CancellationToken is a link back to the Async code. Both can see the value as soon as it changes. The ShowLoadingScreenAsync checks to see if this has changed.

instat/frmSetupLoading.vb
ChrisMarsh82: Done

instat/frmSetupLoading.vb
ChrisMarsh82: Added the code to suppress the warning.

ChrisMarsh82:
@lloyddewit I'm happy for this to wait till I have a use case.

instat/frmSetupLoading.vb
ChrisMarsh82: This is correct. It checks to see if the loading screen is already being shown before it shows it again. It is more of a fail-safe than required functionality

lloyddewit: If `CancellationTokenSource` is nothing, then `HideLoadingScreen()` won't do anything (see line 44. So what's the purpose of calling `HideLoadingScreen()`?

instat/frmSetupLoading.vb
lloyddewit: ok, thanks

lloyddewit (approved):

rdstern:
@ChrisMarsh82 are you happy for this pull request to keep sitting here?

lilyclements:
@ChrisMarsh82 just prodding you given @rdstern's question above

Issue #7493, opened 2022-05-23
Add more data types to the column header - and improve the display of lists


N-thony:
@rdstern @dannyparsons  could we now have the possibility to make datetime?
I have been looking this link https://stackoverflow.com/questions/44970127/remove-timezone-during-posixlt-conversion-in-r
@lilyclements what do you think?
Probably add an option in Year - Month -Day, a checkbox when it is checked it open a groupbox containing hour,min,sec input controls where we can fill and make a datetime and when it is unchecked we make date simply
![image](https://user-images.githubusercontent.com/68591383/169832891-c1ab2bf1-b793-4dfa-b824-246b06625c08.png)

N-thony:
I'm not sure if we need this since we can make datetime through the Calculator Dailogue

lilyclements:
We can run any R function through the calculator dialog so I don't understand why this is a problem?
In R-Instat we can give a front end to the code. Do you think this is something this is useful to have in a front end? I think it sounds sensible, and in the PLH work there are variables that are date-time. But what do you think? I'm not sure of the reasons why this hasn't been previously implemented.

rdstern:
For now I am happy to use the calculator, but later I suggest it will be good to also have a dialogue.  But also, for now, we currently have type (D) for all these types of data.  I feel the types would be clearer in R-Instat if we kept (D) for a date variable and had (DT) for date-time and (T) for time.   I have sometimes found it confusing when some current type (D) variables don't work, as I expect, because they are (DT). 
Perhaps @dannyparsons can confirm or deny?

dannyparsons:
Roger's suggestions sounds sensible to me.

rdstern:
@N-thony version 0.7.6 is holding up well in the important India (IMD) workshop that I am finishing today.  The only 2 bugs/inconveniencies have been a bug in the use of the PICSA rainfall graphs (Vitalis is looking at that) and the confusion here about date-time variables.
In the data I have sent you, the first sheet had a (current) date variable.  And it looks like an ordinary date too.
a) If you examine the metadata the variable called ...1(D)  appears as a class POSITct.POSITt rather than date.
b) If you use the Describe > One Variable > Summarise on it, then the summary includes the time element.
It is particularly confusing, because it presents with just the date part.

a) I suggest it be given the type (D.T) instead.  (That's like an ordered factor is (O.F).   
b) We may stop there, or also consider time variables as of type (T) at the same time.

Thanks

[...]

PR #8015, opened 2022-12-13
Improved the displays of the lists and column type header

### LINKED ISSUE #7947: Improve the Add Columns to Data function?
This is an interesting oddity, which I suggest needs just a small edit.  I wonder if @lloyddewit might be prepared to look at it.  Otherwise it could be @N-thony?  It could wait till the new integer keyboard, by @anastasia-mbithe is available.

Here is how to get this item.
a) File > New Data Frame, change rows to 100 and click to give the 2 variables including the first going 1 to 100.
b) Use Prepare > Calculator > Integer keyboard > factorize2 key, adding x1 as the variable, to factorize the numbers 1 to 100.
image

c) These are the commands generated - as a picture - which looks nice - and as a version that can be edited:
image

In this code it is from pressing the script button, but I added line 7 to look at the results - which is part of the point of this issue.

```
# Code generated by the dialog, Calculations

data <- data_book$get_data_frame(data_name="data", use_current_filter=FALSE)
attach(what=data)
factorize2 <- DescTools::Factorize(x1)
head(factorize2)
data_book$add_columns_to_data(data_name="data", col_name="factorize2", col_data=factorize2, before=FALSE)

detach(name=data, unload=TRUE)
data_book$append_to_variables_metadata(data_name="data", col_names="factorize2", property="labels", new_val="")
rm(list=c("factorize2", "data"))

```
Now here are the results:
image

On the right, you see the output from running the DescTools function, (for the first 6 observations) and on the left you see it in R-Instat.
So the function `data_book$add_columns_to_data` has introduced some inconsistencies in the variable in the data frame.  It usually adds the c(  ) and separates the data by commas, but sometimes it doesn't, and then has a colon.  I don't mind either, but would like it to be consistent!

N-thony:
Fixes (partially) #7493
Fixes #7947
Fixes partially #8029
@rdstern I have added the column type (S) for the list, (CX) for complex and also improved the display of lists in the grid. Take a look.


N-thony:
@rdstern I understand, I didn't check this example of sapply, let me look at it. Thanks

N-thony:
@rdstern this is seems to work fine. please have a look.
@ChrisMarsh82 please could you review this?

ChrisMarsh82:
@N-thony Looks good. I like the way you have done this. Very neat.

rdstern:
@N-thony here is the error I get when editing.
I produced the list in the same way as usual (calculate > Integer > Divisors)
Then duplicated the variable.  That worked fine.
Then edited row 6 from 1, 2, 3  - I changed the 3 to 4.

It then complains that there is a string variable:

image

As I mentioned somewhere, I am no longer worried if the display in reogrid is different to the R-viewer,  So you may wish to display it in reogrid as c(1, 2, 3) - if that makes the editing and subsequent translation for R easier?

N-thony:
> @N-thony here is the error I get when editing. I produced the list in the same way as usual (calculate > Integer > Divisors) Then duplicated the variable. That worked fine. Then edited row 6 from 1, 2, 3 - I changed the 3 to 4.
> 
> It then complains that there is a string variable:
> 
> image
> 
> As I mentioned somewhere, I am no longer worried if the display in reogrid is different to the R-viewer, So you may wish to display it in reogrid as c(1, 2, 3) - if that makes the editing and subsequent translation for R easier?

@rdstern this seems to work fine now? Please have a look. Thanks

rdstern (dismissed):
@N-thony that is great, so I am approving.  
1. I found I could replace a list by NA, but it complains if I make an element in the list into NA.  If quick, perhaps you could check this?  But otherwise let's stop for now.
2. We now need other dialogues to be able to produce lists.
3. I would also like lists be strings and possibly also matrices in the future.

But let's get this merged, and also the keyboard first.
It is really nice to be able to discuss lists, with this special case, and also to be able to use sapply in this way.

instat/Model/RCommand/clsPrepareFunctionsForGrids.vb
```diff
@@ -317,14 +317,18 @@ Public Class clsPrepareFunctionsForGrids
...
         'trim white space from ends of value
         strNewValue = strNewValue.Trim()
         clsReplaceValue.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$replace_value_in_data")
         clsReplaceValue.AddParameter("data_name", Chr(34) & _strDataFrame & Chr(34))
         clsReplaceValue.AddParameter("col_name", Chr(34) & strColumnName & Chr(34))
         clsReplaceValue.AddParameter("rows", Chr(34) & strRowText & Chr(34))
+        If bListOfVector Then
+            strNewValue = "list(c(" & strNewValue & "))"
+        End If
         If bWithQuotes Then
```
lloyddewit, line 332: will a list of vectors ever be surrounded by quotes?
If not, then should we change to `ElseIf`?
```suggestion
        ElseIf bWithQuotes Then
```

N-thony:
> @N-thony that is great, so I am approving.
> 
> 1. I found I could replace a list by NA, but it complains if I make an element in the list into NA.  If quick, perhaps you could check this?  But otherwise let's stop for now.
> 2. We now need other dialogues to be able to produce lists.
> 3. I would also like lists be strings and possibly also matrices in the future.
> 
> But let's get this merged, and also the keyboard first. It is really nice to be able to discuss lists, with this special case, and also to be able to use sapply in this way.

@rdstern I have fixed 1. and I agree for 2. and 3. which will help on improving this.

rdstern (approved):
Great

instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
```diff
@@ -71,7 +71,11 @@ Public Class ucrDataViewReoGrid
         strRowNames = dataFrame.DisplayedRowNames()
         For i = 0 To grdData.CurrentWorksheet.Rows - 1
             For j = 0 To grdData.CurrentWorksheet.Columns - 1
-                grdData.CurrentWorksheet(row:=i, col:=j) = dataFrame.DisplayedData(i, j)
+                Dim strData As String = dataFrame.DisplayedData(i, j)
+                If strData.Contains("(") AndAlso grdData.CurrentWorksheet.ColumnHeaders.Item(j).Text.Contains("(LT)") Then
+                    strData = strData.Split(New String() {"(", ")"}, StringSplitOptions.None)(1)
+                End If
```
lloyddewit, line 77: I understand that this code is setting `strData` to the text inside the brackets for strings with a format like `<txt>(<txt>)`.
Does it need to cope with strings in different formats, e.g. nested brackets? or no chars perceding the first '('?
Please could you add a comment with examples to help other developers understand the intention of this code?

lloyddewit (changes requested):
@N-thony Thank you, looks great. Just a couple of open questions

rdstern (dismissed):
@N-thony great.  I approve again.  I just note that the R viewer displays the first element as an empty list, while you display it as 0?  Should it be 0?
If it is empty I would also be happy if it was made into NA. 
Should it be 0?  I tried deleting it, and it complained.  Could it make NA when deleted?

instat/Model/RCommand/clsPrepareFunctionsForGrids.vb
N-thony: @lloyddewit if I commit this it means we will never pass the parameter value on line 333 or 335 if bListOfVector is True.
bListOfVector is helping only is just helping to change the parameter value i.e strNewValue to the format needed to replace a value if vector in the dataframe list column.

lloyddewit: yes, I agree, thanks for checking

N-thony:
@rdstern I tried to delete and it works fine.
image

When I display the data with the grid in RStudio, the first value is numeric(0) so for the nicer display as suggested in the issue, I found that numeric(0) returns a numeric vector of `length 0, `so when you add anything
to it you get the same result (it's basically a numeric NULL). I think it will be good to display as NA

rdstern:
Great - I agree

N-thony:
@rdstern I have made the changes as suggested replacing numeric(0) by NA.
@lloyddewit I resolved your comment above.
Please have look. Thanks

lloyddewit:
> @lloyddewit I resolved your comment above.

@N-thony I think there's still one unresolved comment above

instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
N-thony: @lloyddewit I'm just checking the first bracket because I think if that exist then there is a closing one.
I have also added some comments for the intention of these codes

lloyddewit (approved):

lloyddewit:

rdstern (approved):

rdstern (commented):
@N-thony I have now found a new minor oddity in the solution, so far.
If an operation on a list produces another list, then the NA is not carried forward.
Here is an example (that I have fixed):
image

It is as follows:
a) Usual new data frame, with 100 rows and x1 1 to 100.
b) x2 = Calculations > Integer > divisors put in x2.
c) calc is sapply (x2, with the key digitsum in the same keyboard).  

This gives the variable `calc`, but with the NA value as blank.
d) Then `sapply(calc,sum)` doesn't work - and says it can't, because calc is a list.
e) When I delete cell 1 in calc, then it replaces the blank in calc by NA, and then it works.

So a small tweak is needed in your code, when a LT variable produces another LT variable.

I would also like to be able to to do both steps together, i.e. 

`sapply(sapply(x2,DescTools::DigitSum),sum)`

and this currently doesn't work.  (by the way if sum is replaced by mean, then it does work!)

rdstern (changes requested):
@so I have withdrawn my approval for now!

rdstern:
@N-thony I have also found another interesting "feature" with your code.
Try the usual new data frame, length 100, with x1 as 1 to 100.  Then use calc > integer > factorize.  The result is shown in x2 below:

image

You see for 2 , 8 and 9 the separator is a colon (:) rather than a comma.  Don't spend too long on this, because it doesn't seem to cause any problems. It is still recognised as a list of 2 numbers.

N-thony:
> sapply(x2,DescTools::DigitSum)

@rdstern there is an oddity in the data produced by example with calc = `sapply(x2,DescTools::DigitSum)` i.e item c) the value in the first cell is `list()` and I think this needs also to be displayed as `NA` since there is no value but the difference wuth `numeric(0)` here we need to edit the data replace `list()` by `NA`, what do you think? 
`calc <- data_book$get_columns_from_data(data_name="data", col_names="calc")`
`calc[1]`, from this observation you cannot directly run `sapply(sapply(x2,DescTools::DigitSum),sum)` you will get the error in item d)
image

N-thony:
> > sapply(x2,DescTools::DigitSum)
> 
> @rdstern there is an oddity in the data produced by example with calc = `sapply(x2,DescTools::DigitSum)` i.e item c) the value in the first cell is `list()` and I think this needs also to be displayed as `NA` since there is no value but the difference wuth `numeric(0)` here we need to edit the data replace `list()` by `NA`, what do you think? `calc <- data_book$get_columns_from_data(data_name="data", col_names="calc")` `calc[1]`, from this observation you cannot directly run `sapply(sapply(x2,DescTools::DigitSum),sum)` you will get the error in item d) image

also, if try to change the first cell of `x2` to `NA` in the data by deleting the cell you get the following error. So I'm not sure if we should edit the data when we `numeric(0)` or `list()`?
image

rdstern (dismissed):
@lloyddewit I am approving as it seems sort of ok now & we need to move on. 
I still notice that the R viewer gives a blank, where R-Instat gives NA for a list of length zero.  If the code is ok, then let's accept.

rdstern:
@N-thony that's an improvement, but there is still a problem.  If I edit any entry in the list then that row doesn't work anymore.  Even if I just go into a row to edit, but don't change anything then it doesn't work.

I therefore asked Danny and David, if we are slightly opening a can of worms and should give up on all this.  They feel the opposite - so we continue.

The LT column is a list column and each element of the list is a numeric vector.  I think when sent to R these elements have a c(  ) round them, i.e. they look more like the way we used to display them in the grid.  
The R viewer distinguishes between that and the way the contents of the list are displayed to the user.

So, I am now more relaxed about having the display in reogrid different from the R-viewer than before - if that would also become simpler for you to code and clearer for the user.

But if we do display as you have now - that is still not quite the same as the R viewer - so the top element in reogrid is now 0, while it is blank in the R-viewer.   Anyway we should be able to change a value in reogrid, and it still works.  (Once we can do that I am particularly keen to see what happens if one of the elements is a missing value.)

I am happy that you are asking @ChrisMarsh82 to review the final code.  If you are not sure on the change to make now, and Chris is also unsure, then perhaps he could check with Danny or David.

In our statistical ideas, this is going to help us explain and discuss multiple response questions.

We will stop for now with these numeric vectors, but later will then include character vectors as the elements.  But that will be in another pull request.

And as a general idea, we are not taking a small step towards tibbles

N-thony:
@ChrisMarsh82 I have now made the changes I thought regarding Roger's comments above. Could you have a look?
@rdstern we can have a quick call about  what you have reported above?

rdstern (changes requested):
The display seemingly looks fine now - except it is giving the wrong answers!  I compare your display with the R-viewer here and you see that you are now just keeping the first digit of each number.

image

I guess this is a tricky task and keep thinking that @ChrisMarsh82 or @volloholic may need to add explanation.  It is important, but I am concerned at the time it is taking you.

I am delighted though that you must now have your new machine working.  That was quick!

N-thony:
@ChrisMarsh82 I used the Regex expressions with this pattern `.Regex.Matches(strData, "\d+|\bN\S*A\b")`. I think this is much more easier or better

rdstern (dismissed):
Great - that (almost?) seems to work.  I tried with a column of length 100 and with the pascal function - that will become a key.

`lapply( x1, function(x) {lapply(x, function(i) {choose(i, 0:i)})})`

I have approved - unless you want to try a last tweak to your code?  So, first, here is the good news:

image

It looks fine, doesn't it?  Very nice.  And it doesn't go into scientific notation either - which the R-viewer does.  

But later in the sequence there is a puzzle for me.  I am in the middle where the numbers are very large.  I assume there is an oddity here with the E notation?  Maybe if you recognised e and + in the regex it would become clear?  By the way I am having more trouble in the R-viewer now, than in your reogrid!

image

instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
```diff
@@ -90,6 +94,17 @@ Public Class ucrDataViewReoGrid
         grdData.CurrentWorksheet.RowHeaderWidth = TextRenderer.MeasureText(strLongestRowHeaderText, Me.Font).Width
     End Sub
 
+    Private Function GetVector(strData As String) As String
```
ChrisMarsh82, line 97: I would like to change how this works. Currently, you loop through all the values within the string. This could be inefficient if there are lots of values in the string. Also, GetVector does not really describe what the procedure is doing.
This is a proposed procedure that could be used instead:
`Private Function GetInnerBracketedString(strData As String) As String`
`    Dim strOutput As String = Strings.Left(strData, InStr(strData, ")") - 1) 'Get the first right bracket`
`    strOutput = Mid(strOutput, InStrRev(strOutput, "(") + 1) 'Get the last left braket`
`    Return strOutput`
` End Function`

ChrisMarsh82 (commented):
@N-thony

N-thony:
@ChrisMarsh82 I just notice a small issue that the grid get bigger as the number do
image

ChrisMarsh82:
> @ChrisMarsh82 I just notice a small issue that the grid get bigger as the number do 
The issue if it doesn't do this it could be more misleading. It looks like only a certain amount of data can fit on a row in the ReoGrid and it always shows the last row. therefore if it doesn't grow you would see something like this.
image
Row 44 now looks like it only contains '44, 1' when actually that's just the last 2 values.

If this is a problem for users we could add a text box at the top of the screen to show the value in the current cell?

ChrisMarsh82 (approved):

rdstern:
@N-thony and @ChrisMarsh82 sorry I was too quick in approving and now seem to have a bug in the merged version.  I only tested with the new pascal code and that works fine.

So I had a new data frame of length 100 with x1 1 to 100 and then the code:

`lapply( x1, function(x) {lapply(x, function(i) {choose(i, 0:i)})})`

This (in the merged version) still correctly, (and impressively) gives as follows:

image


But then the command shown above, i.e. DescTools::Divisors(x1) - which was working before - gives an error and stops R-Instat.

image

ChrisMarsh82:
@rdstern - I have a fix but I want to check. The code is designed to get anything for the inner brackets (if they exist). Therefore c(1,2) will become 1,2 and numeric(0) will become 0. Is this correct or is there an example where we need to display the whole return string including the brackets?

lloyddewit:
@rdstern Thank you for the comment above. This PR is merged so there's a risk that this comment will be overlooked.

Does this comment apply to one of the issues linked with this PR?
Or is the comment related to the new PR @ChrisMarsh82 made (#8050 )?

Should we move the comment?

PR #7814, opened 2022-08-25
Modifying the plot option control in Line plot dialog. >Describe>Specific Tables/Graphs>Line/Smooth/Dumbbell/Slope

Fidel365:
Fixes (partially) #7556 
@N-thony , @lloyddewit , this PR is ready for review

PR #8017, opened 2022-12-15
Modifying the plot option control in Line plot dialogue

derekagorhom:
fixes (partly) #7556 
replaces PR #7814 

@N-thony  this PR is ready for review


rdstern (changes requested):
@derekagorhom it is good you are working on this.  I would like to have the new system in place for all the graph dialogues in the next version if possible.
a) When I start quite a lot of the options are enabled.  It is all going to be a complicated set of enables and disables if you want to do it properly.  For example all these below are enabled when I have ticked nothing.

image

However smooth is not just a main option, but also on the line option.  And it is never enabled on this option as far as I can see.

image

A different small point now.  with line of best fit there is the se checkbox, which is now at the right height.  But could you move it so it is above the Add Valleys checkbox.

Then I suggest, rather than having all the complications of enabling and disabling, we enable all the pages at the same time.  @lloyddewit may wish to comment.  It will greatly simplify the code and I don't see that it does any harm.

lloyddewit:
@derekagorhom It's perfectly fine to close this PR. If we close a PR, then it's good to add a comment explaining why we're closing (e.g. we will implement the change in a new PR).
If you replace this PR with a new PR, then in the new PR, please reference this PR.
Thanks

PR #8021, opened 2022-12-16
In Make Date dialog, ensured that default position of the resulting date variable is before the year variable (not after)

### LINKED ISSUE #8018: Minor change in the Make Date dialogue
@Vitalis95 this should be a very quick fix (I hope?)
The option was used with the 3 variables.  This then (by default) positions the resulting date variable **_after_** the year variable.  It should be **_before_**.
Please check the situation and when there are 2 variables, so Year and day_of_year.  Again, have it before the year variable, and not afterwards.
When there is just one variable, then it is probably ok as it is.  That's probably after the single variable.  If the whole code is simpler if the date is always before the variables used, then that would be fine.

Vitalis95:
Fixes #8018 
@rdstern, @lloyddewit ,  @N-thony , this PR makes **`before`** position as default for Year and Day of Year & Year-Month -Day


rdstern (approved):
@Vitalis95 this looks fine.  If @lloyddewit is ok it can be merged.

lloyddewit (approved):

PR #8026, opened 2022-12-20
Added Complex Numbers keyboard to Calculator

### LINKED ISSUE #8013: Add a complex numbers keyboard to the calculator
@N-thony this should be easy to allocate - and perhaps fun, for someone who is interested in maths.  It is not urgent, but it fits well with using R-Instat to support maths courses for the sort of maths needed generally and also useful in statistics.  Perhaps @anastasia-mbithe will be interested to discuss with the education team?  Maybe @derekagorhom could do it, or @anastasia-mbithe herself, if it is quick and she is interested, or even someone new, and supervised by Ana?

They could play first to understand what this is all about.  R includes complex numbers as described [here](https://stat.ethz.ch/R-manual/R-devel/library/base/html/complex.html).

Whoever is doing it should first have an understanding.  They can do everything now, of course, by just typing the functions (and using the +, - *, / and ^ keys).  Here is an example:

a) `File > New Data Frame, values 1 to 100` as usual.
b) In the calculator type something like `complex(real = stats::rpois(100,1), imaginary = stats::rpois(100,4))`
c) Then add some simple calculations, e.g. x1 * x2, x2 ^ 3, etc to show those work.  
d) Then show the square root now works for negative numbers, e.g. sqrt(as.complex(-x1))
e) Then log and exp, etc.
f) And the circular functions (sin, cos, etc) from the maths keyboard also work.

image

The screen-shot also shows a (hopefully small task for @N-thony, namely that the significant figures don't work, so the results look messy.  (The scientific does.)  And the signif function from the maths keyboard works fine, so this should probably be a very small change.  Perhaps that could be done quickly on a separate pull request?

So to the keyboard.  
complex, real, imaginary, mod, arg
conjugate, sqrt, exp, log, round  Then Help of course!

The last 4 are just taken for the maths keyboard.  The others are for as.complex, Re, Im, Mod, Arg, Conj as described in the documentation.  Hopefully the special Help key could go straight to the Complex page. 

There is a bigger picture I am then considering here.  The mosaic project seems still to be going.  It includes a book about the use of R to support [the teaching of calculus](https://dtkaplan.github.io/RforCalculus/).  I wonder about the role of R-Instat and their mosaic project.  

I am not sure if that book is our best entry point - and it doesn't mention complex numbers much.  But it is consistent with the use of R to support maths teaching.  I will be happy discussing statistical aspects of mosaic and I wonder if there will be interest in innodems in the maths side?

derekagorhom:
fixes #8013 

This PR is ready for review


derekagorhom:
I wanted to find out if this keyboard would have tooltips and also i found out that some keyboards on the calculator do not have tooltips

`Maths` keyboard
`trunc`, `deg`, `sign`, `abs`, `exp`, `logit`, `logistic`, `atan2 `

`Text/String` keyboard
`reverse `

rdstern:
@derekagorhom this is excellent and I am also really pleased that you are questioning more broadly - with your question on tooltips.  (That is for later.)
As far as I have tested, it is working fine, so could be merged.  But I suggest some minor changes, plus then (sorry) some more keys.  It has become more important than I thought earlier.
a) Could you make `complex` as the second-to-last keyboard.  I really like keeping the Integer keyboard as the last one.
b) In the Help it goes to Base R correctly.  Even though there is really only that one package, could you use the drop-down control.  Then users can see the name of the package - call it just `Base`, without necessarily going into it.  
c) Now the extra keys!  Most are copies from the maths keyboard, but there are 3 new ones.  I am not sure why they are not also on the maths keyboard, but that is for another time!
I now wonder about this layout:
```
complex, real, imaginary, rad
mod, arg, conjugate, deg
exp, sin, cos, tan
log, sinh, cosh, tanh
sqrt, round, signif, pi
```

derekagorhom:
@rdstern i have made the changes, can you review them
thanks

rdstern (approved):
@derekagorhom I am impressed.  To me it is all ok.

lloyddewit (approved):

PR #8044, opened 2023-01-10
Added a Pascal key to the Calculator probability keyboard

### LINKED ISSUE #8036: Add a Pascal key to the probability or special keyboard
This is not to be done immediately.  And I need help, because I can't yet do it as I would like.

The murderous maths book on probability declares that Pascal's triangle is one of the most gorgeous things in maths.  

I would like to generate them as a list and it seems surprisingly easy.  It uses the choose function for which there is already a key.

I discuss this in #8034 .  I then found that the simple code:

`lapply(x1,choose,0:100) `is almost what I want:

image

But the list variable always has 101 elements.  Nice if it had, 2, 3, etc for each row.  Having 0:x1 gives just the first 2 values.  I assume I need to define and run a function?

Maybe this is nice for the propose Special keyboard I am planning to support the teaching of programming.

anastasia-mbithe:
Fixes #8036 

@rdstern this is ready for testing


N-thony:
@derekagorhom can you peer review this?

rdstern (changes requested):
@anastasia-mbithe this works fine.  Just for uniformity, could you have pascal with a small P.  Then this can be merged - at least as far as I am concerned.  

I note the code here, because I would like to use it to check @N-thony changes in the display of lists.

`lapply( x1, function(x) {lapply(x, function(i) {choose(i, 0:i)})})`

I note that your display of the results for 1 to 100 has the c(  ) round it, but otherwise displays fine and only goes to scientific much later than the R-viewer.  I am happy to "delay" the change to scientific.

rdstern (approved):
Great

derekagorhom (approved):

lloyddewit (approved):

PR #8047, opened 2023-01-12
When selecting data frames, allow user to add all/selected data frames

### LINKED ISSUE #7606: Extend the drop-down on the Add control?
This may need some discussion, so we do the right thing here.

image

@Patowhiz and @ChrisMarsh82 I am suggesting Chris may have a view here, because he never liked my right-click in the data selector!

a) The drop-down should be added to the selector also when it is selecting data frames.  For example here:
image

b) The reviewer also questioned whether we should have an` Add/Drop `button, rather than just the `Add,` as we have now.  If so, then I assume we could call it `Add` (as now), when the focus is on the data selector.  And this could change to `Drop`, when the focus is on a receiver?  There we have Clear on the current right-click and this would then be included on the drop down.  It might be there anyway, but disabled unless it is relevant.

N-thony:
Fixes #7606
@rdstern @ChrisMarsh82 this is ready for your review. Thanks


rdstern (approved):
Great

lloyddewit (approved):

PR #8049, opened 2023-01-13
Minor bug fix of display of list variables from PR #8015

N-thony:
Fixes bug remaining in PR #8015 .
@rdstern could you test this?

PR #8050, opened 2023-01-13
Updated procedure to allow for no brackets in string

ChrisMarsh82:
@rdstern @N-thony - this is the fix for #8015 it now checks to see if brackets exist.

The code is designed to get anything for the inner brackets (if they exist). Therefore c(1,2) will become 1,2 and numeric(0) will become 0. Is this correct or is there an example where we need to display the whole return string including the brackets?


rdstern:
@ChrisMarsh82 I am happy to review, but slightly confused because with the addition of #8049 it now seems to do that already.  So I am unsure what this pull request is adding?

ChrisMarsh82:
@rdstern it looks like @N-thony be me to it with #8049. Will close this off.

PR #8057, opened 2023-01-16
Fixed small bugs in the Regular Sequence dialog

### LINKED ISSUE #8051: Regular Sequence Dialog
When I open the regular sequence dialog, on the right side it says no preview available so I cannot see a preview of the sequence I am creating. 
I need this fixed so that it's okay in the next release.

**To Reproduce**
1. Go to 'Prepare>Column: Numeric>Regular Sequence'
2. Look at the right side of the dialog
3. See error

**Expected behaviour**
I expect to see a preview on the right side of the dialog when I open it but there's none.

**Screenshots**
Regular Sequence 1_13_2023 2_17_38 PM

**R script or commands**

regular <- rep(x=seq(from=1, to=100, by=1), each=1, length.out=100)
data_book$add_columns_to_data(data_name="data", col_name="regular", col_data=regular, before=FALSE)

rm(regular)

N-thony:
Fixes #8051
@rdstern @berylwaswa please take a look.

### DIFF
--- a/instat/dlgRegularSequence.vb
+++ b/instat/dlgRegularSequence.vb
@@ -21,6 +21,7 @@ Public Class dlgRegularSequence
     Public bFirstLoad As Boolean = True
     Private bReset As Boolean = True
     Private clsRepFunction, clsSeqFunction, clsSeqDateFunction As New RFunction
+    Private clsDummyFunction As New RFunction
     Private clsByOperator As New ROperator
     Private enumDefaultSequenceOption As DefaultSequenceOption = DefaultSequenceOption.NumericOption
     Private bDefaultOptionChanged As Boolean = False
@@ -129,6 +130,8 @@ Public Class dlgRegularSequence
         ucrPnlSequenceType.AddToLinkedControls({ucrInputComboDatesBy}, {rdoDates}, bNewLinkedHideIfParameterMissing:=True)
 
         ucrChkPreview.SetText("Preview")
+        ucrChkPreview.SetParameter(New RParameter("Checked", iNewPosition:=0))
+        ucrChkPreview.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
 
         ucrNewColumnName.SetPrefix("regular")
         ucrNewColumnName.SetDataFrameSelector(ucrSelectDataFrameRegularSequence)
@@ -147,6 +150,7 @@ Public Class dlgRegularSequence
         clsRepFunction = New RFunction
         clsSeqFunction = New RFunction
         clsSeqDateFunction = New RFunction
+        clsDummyFunction = New RFunction
         clsByOperator = New ROperator
 
         ucrSelectDataFrameRegularSequence.Reset()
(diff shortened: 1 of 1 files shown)


rdstern:
@N-thony this fixes the problem, but seems to have returned us to the earlier problem, namely that it appears to freeze when there is a long data frame.
I tried with the survey - length 36 and it seems to work fine.
Then with a new data frame of length 1 million and it still tries to give the preview - so it looks as though it is frozen for a time.  This is partly linked to the problem that changing the data frame and also reset don't seem to do much.
I suggest:
a) Changing the data frame should always untick the preview.  And it should review the default, so it is appropriate for that data frame.
b) Reset should also untick the preview, and should set the controls to the default for the data frame.
c) There should still be a preview not available if the data frame is long.  We could check different values, but I suggest perhaps 10,000 could be the limit - if longer, then no preview.

N-thony:
@rdstern I also notice the to input control by default in the code is the length of the dataframe but when you change the dataframe it still remain with the previous dataframe length.

rdstern:
This is much better, especially on the reset.  But there are still some oddities on the preview.
It still tries to give a preview when there are very long columns - and I think that is part of the problem.
When the column is long - say more than 10,000 then I wonder if the preview can quickly compare the length of the generated variable with the data frame - so give that information.  But not the data in the preview.

One big problem I found currently was when I moved from the Numeric > Regular Sequence to the Date > Generate.  Then it had remembered all the settings from the numeric use of the dialogue and took ages to come.  I was only asking for a data frame of length 100, but the preview was stuck because the other data frame is of length 1 million,

rdstern:
Maybe a separate problem is that I can't get a preview of dates?

image

It used to exist.  Here it is as the default in version 7.3

image

N-thony:
@rdstern following item c) I did set the limit to 10000, I think this solve the problem in your last comment?

> @N-thony this fixes the problem, but seems to have returned us to the earlier problem, namely that it appears to freeze when there is a long data frame. I tried with the survey - length 36 and it seems to work fine. Then with a new data frame of length 1 million and it still tries to give the preview - so it looks as though it is frozen for a time. This is partly linked to the problem that changing the data frame and also reset don't seem to do much. I suggest: a) Changing the data frame should always untick the preview. And it should review the default, so it is appropriate for that data frame. b) Reset should also untick the preview, and should set the controls to the default for the data frame. c) There should still be a preview not available if the data frame is long. We could check different values, but I suggest perhaps 10,000 could be the limit - if longer, then no preview.

rdstern (approved):
@N-thony great.  @lloyddewit I hope the code is ok to merge?

instat/dlgRegularSequence.vb
```diff
@@ -129,6 +130,8 @@ Public Class dlgRegularSequence
         ucrPnlSequenceType.AddToLinkedControls({ucrInputComboDatesBy}, {rdoDates}, bNewLinkedHideIfParameterMissing:=True)
 
         ucrChkPreview.SetText("Preview")
+        ucrChkPreview.SetParameter(New RParameter("Checked", iNewPosition:=0))
```
lloyddewit, line 133: R parameter names are case-sensitive. Does that matter in this context?
```suggestion
        ucrChkPreview.SetParameter(New RParameter("checked", iNewPosition:=0))
```

lloyddewit (changes requested):
@N-thony Thanks, just one small question

instat/dlgRegularSequence.vb
N-thony: yes, typo. Thanks

lloyddewit (approved):

PR #8059, opened 2023-01-17
Added factorize2 and primorial keys to the Calculator dialog integer keyboard

### LINKED ISSUE #8052: Add factorize2 and primorial keys to the integer keyboard
a) @derekagorhom could the `conf.design` package please be added to R-Instat.
b) Then add a factorize2 key after factorize on the integer keyboard.
c) Then move the digitsum key, so it fills the 2nd row of keys.
d) Then everything else just moves down - so Fibonacci and lucas now start the 3rd row.
e) The new tooltip for the factorize2 could say `Like factorize, but with simpler layout of results.  But much slower for large data sets.`

For the record, this package is written by a very prominent Australian statistician.  We could usefully write to him about the code to see whether he may wish to improve it.  Up to 10,000 it is quick.  It is considerably slower for 100,000 observations.  For 1 million the DescTools package took about 13 seconds, while this took about 30 minutes - so about 100 times slower.  

Useful that the main aim in this package is factorial designs, i.e. statistics!

Could you also add a primorial key to the end of the primes block of keys.

The code is `sapply(  ,primes::primorial_n)`  And the tooltip is Gives the primorial (like the factorial, but just the primes up to the number) for a variable.  For example primorial(c(7,8,9)) = 2*3*5*7 = (210, 210, 210)

derekagorhom:
fixes #8052 

This PR is review 
@rdstern can you test and approve


rdstern:
The primorial key is fine.
Label the factorize2 key with a small f.
Then the reason you are adding the conf.design package is to use the factorize function from that package.  So please use that function, rather than the one from primes.

Also check with @N-thony on how to specify that the package is to be added.

lloyddewit:
@derekagorhom Thank you for this PR.
The last commit deletes `install_packages.R`. This is not correct. I assume that you wanted to delete the changes you made rather than the file itself. This now needs to be fixed. The easiest way to resolve is to copy `install_packages.R` from the master branch back into your PR branch.
If you're unsure what to do, then @N-thony should be able to help.

rdstern:
@derekagorhom now I understand why you gave the command that way.  You were using the conf.design package, but should have known more about how to use a command from a package.  And you could have seen this from the code from many of the other keys.  I didn't realise I needed to specify the code as completely as that!

The command should be given as
`conf.design::factorize( )
`
I was confused by your addition of the primes argument.  And that doesn't work as you gave it either.  If you had wanted to give that, then you needed to add the con.design:: there too.

So you could have programmed it as:

conf.design::factorize(  ,conf.design::primes(  )).

But you don't need this extra argument, because (as shown in the conf.design manual) that is the default.

I deduce from this that you are still not trying the code you provide.  You should always test it yourself on examples of data.  Then usually you should also mention the examples you used for testing.

There is a second item to add - that I forgot.  Could you also add the conf.design package into the help menu.

Now also a big problem you need to discuss with @N-thony and either he or @ChrisMarsh82 need to fix.  Can you start by checking my problem below and then making an issue from this. 

When I try the correct code it gives an error.  It also (now) seems to give an error on any key that generates a list.  So now the divisors key doesn't work, and neither does factorize, etc.  I also get the same error in the master, so it isn't just your branch.  Try with a data set of 1 to 100 say and divisors.  Do you also get this?

image

N-thony:
> @derekagorhom now I understand why you gave the command that way. You were using the conf.design package, but should have known more about how to use a command from a package. And you could have seen this from the code from many of the other keys. I didn't realise I needed to specify the code for you as completely as that!
> 
> The command should be given as `conf.design::factorize( ) ` Most of the time you need to give the name of the package. I was confused by your addition of the primes argument. And that doesn't work as you gave it either. If you had wanted to give that, then you needed to add the con.design:: there too.
> 
> So you could have programmed it as:
> 
> conf.design::factorize( ,conf.design::primes( )).
> 
> But you don't need this extra argument, because (as shown in the conf.design manual) that is the default.
> 
> I deduce from this that you are still not trying the code you provide. You should always test it yourself on examples of data. Then usually you should also mention the examples you used for testing.
> 
> There is a second item to add - that I forgot. Could you also add the conf.design package into the help menu.
> 
> Now also a big problem you need to discuss with @N-thony and either he or @ChrisMarsh82 need to fix. Can you start by checking my problem below and then making an issue from this.
> 
> When I try the correct code it gives an error. It also (now) seems to give an error on any key that generates a list. So now the divisors key doesn't work, and neither does factorize, etc. I also get the same error in the master, so it isn't just your branch. Try with a data set of 1 to 100 say and divisors. Do you also get this?
> 
> image

@rdstern this bug is fixed in PR #8049, once merged this will be solved.

derekagorhom:
@rdstern thanks for the correction,  i have made the changes.
can you please have a look and approve. 
thanks

rdstern (dismissed):
@derekagorhom great - that looks fine now.  Thanks
It now awaits @lloyddewit for the second check.

instat/ucrCalculator.vb
```diff
@@ -4398,6 +4408,13 @@ Public Class ucrCalculator
 
         ucrReceiverForCalculation.AddToReceiverAtCursorPosition(clsZseqFunction.ToScript, 0)
     End Sub
+    Private Sub cmdPrimorial_Click(sender As Object, e As EventArgs) Handles cmdPrimorial.Click
+        ucrReceiverForCalculation.AddToReceiverAtCursorPosition("sapply(  ,primes::primorial_n)", 22)
```
lloyddewit, line 4412: Other subs in this class have the option to show the parameters.
Should we have this option in this sub also? You may want to ask @rdstern 's opinion.
For example:
```
    Private Sub cmdCeiling_Click(sender As Object, e As EventArgs) Handles cmdCeiling.Click
        If chkShowParameters.Checked Then
            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("ceiling(x= )", 1)
        Else
            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("ceiling( )", 1)
        End If
    End Sub
```

lloyddewit (changes requested):
@derekagorhom Thanks for the changes, just some small questions

instat/ucrCalculator.vb
derekagorhom: @lloyddewit i dont think so in this case since `sapply` takes list as input and gives output in vector or matrix.
but @rdstern can confirm

rdstern:
@derekagorhom I think that's right.  In this case the real function isn't sapply, which is just handling the looping.  So, for now I'm happy without it.  We may add later.

lloyddewit:
> @derekagorhom I think that's right. In this case the real function isn't sapply, which is just handling the looping. So, for now I'm happy without it. We may add later.

@rdstern @derekagorhom 
OK thanks, understood.
How about the new sub added just below (I assume that `factorize` is the real function)? Should this have the option to add parameters?
```
    Private Sub cmdFactorize2_Click(sender As Object, e As EventArgs) Handles cmdFactorize2.Click
        ucrReceiverForCalculation.AddToReceiverAtCursorPosition("conf.design::factorize( )", 3)
    End Sub
```

rdstern:
@lloyddewit usually yes, but this time the only argument is very confusing, so I am happy without it for now.

lloyddewit (dismissed):
Thanks @derekagorhom

lloyddewit:
@rdstern The code changes should not have affected the functionality, but still good to test. If you can reapprove, then we can merge, thanks

rdstern (changes requested):
@derekagorhom just a small correction, which I suggest is being caused by you still being optimistic and not trying your improvements yourself?

The factorize2 key needs to change the position of the cursor, so it is inside the bracket.

While you are doing that, I make a cheeky request.  You programmed the pascal key recently and it seems to work ok, but it works better if you could make a small change, namely to change the first letter to s from l.  So you are using sapply there, rather than lapply.  Thanks

rdstern (approved):
Looks fine now.  Back to you @lloyddewit

lloyddewit (approved):

PR #8060, opened 2023-01-17
Added new Export Data for Climsoft dialog

### LINKED ISSUE #6377: New dialogue to export data for Climsoft
1. This is to be added to the Climatic > File menu.  It is called `Export for Climsoft.`  It can be the first Export.  At the same time please change the others to `for`, rather than `to`.  (I suggest that the word for reflects that we are preparing the file correctly.  But we are not then going into Climsoft.)
2. The data must be defined as climatic.  In the definition of Climatic it must have a Station variable, which is the Station ID, even if it is for a single station.  Also all the elements must be defined in the list.
3. There is the usual data selector.  
4. There is a single receiver for the Station ID  - this is filled automatically
5.  There is the usual Date receiver - this is filled automatically.
6.  There is a multiple receiver for the elements.  Default is for all the elements to be included.
7.  Then the save file as, with the Browse.  (The default is csv and maybe that is all we allow.)
8. Interesting to allow Export Comments also.  If Excel is acceptable that would mean the 2 data frames can be exported together.

For this to work smoothly we will have to stack the data when there are multiple elements.
We will also have to include the Climsoft codes with the elements if this is not done already.  I doubt it.

And we need to know how to account for trace.  They are stored as 0.03 in R-Instat and we could arrange for a flag variable to be constructed?

Vitalis95:
Partially fixes #6377
It is still in progress.  
@rdstern ,@N-thony , @lloyddewit, it fixes the first part of the issue of writing a new dataframe. 
 The dialog is in **Climatic > File > Export to Climsoft**
@rdstern , have a look at the progress so far.


rdstern (changes requested):
@Vitalis95 that seems to work ok.  I tried with Dodoma, which doesn't have a station - so I had to add that variable.  I think that is ok, but you might consider, if easy, that you use the data frame name as the station information if it isn't provided.  I am not sure though.
1) Before merging I had the data defined as climatic.  So you should be able, then to include the variables automatically - as we do with many of the climatic dialogues.
2) And a question - perhaps also for @Patowhiz .  What happens on the importing if there are duplicate days?  Is that a problem?  Because you could easily add the control to check that - from the define climatic data.  And then possibly don't proceed if there are duplicates.

Then I would be happy to merge while you then add the proper exporting to the dialogue as another pull request.

And another 2 questions for @Patowhiz .  First what should we do about trace rainfalls?  In R-Instat we have the convention of making them 0.03mm, so we could detect and do something, e.g. even make it a text variable with tr, if that can be imported.  And second, what about flag variables?  We can export them, so they can be in R-Instat.  Can they be imported into Climsoft?

Patowhiz:
@rdstern 
Climsoft overwrites duplicate data. And by duplicate I mean duplicated dates.
Yes, Climsoft can import recognized flags. For trace values, 0 is what is expected not 0.03. The Climsoft team opinion is 0.03 could compound to observations that are not accurate.
Thanks.

rdstern:
Thanks for the clarification.
a) What do you think we should do as Climsoft overwrites duplicate values?  There seem to me to be 2 situations:
    1) The first value is in the database and the duplicate date is imported from R-Instat.  That will usually be a correction, so that's possibly ok.   
   2) There are duplicate rows in the data that is exported from R-Instat.  I suggest that is usually an error and should not happen.  Would it be good if that were checked in the Export dialogue, or are we getting too perfectionist.  It should be easy to add, because the code is there in the define climatic data dialogue.  
  3)  I suggest adding it to the dialogue would be sensible, because it would alert users to the general issue of duplicates.

Now the flags.

1) Pleased that Climsoft can import data with flags.  We need to make that an option in the dialogue.  So, there should be an Add Flag variables checkbox that would presumably open another multiple receiver for the flag variables (Text or factor variables).  Then it would check that there are as many flag variables as ordinary variables, and also possibly that the codes are acceptable?
 2) Rainfall would be a special case.  We could transform the rainfall data, on the fly as we export, i.e. change 0.03 into 0 and add T into a new flag variable.  For simplicity I suggest instead that we add a button at the top of the export dialogue, for that transformation.  It would produce the two new columns and (at least initially) be just for rainfall.  It could be labelled `Flag-precip`.

Patowhiz:
@rdstern I recommend adding the duplicate checks as suggested in item 2 above.
Also note, Climsoft expects the flag to be part of the value column not as a separate column. For examples; 0T or 23E. Then during importing it will detect the character and separate it out as the flag.

Vitalis95:
@rdstern ,it is now taking data frame name if the station information isn't provided, have a look at it

rdstern (approved):
@Patowhiz and @Vitalis95, I exported the data for climsoft into a csv and it looks like this:
image

I assume it can be imported like this into Climsoft, i.e. with the variable names as in R-Instat?  Can you confirm.  I am approving, because (if so) it would be good to include soon.

The next improvement, Vitalis is to complete the dialogue more automatically, when the data has been defined as climatic.  The date can be filled.  And I assume the Station is also filled automatically as a variable when the data are defined as climatic.  And I suggest all the climatic elements can be assumed for export too.  

But this is all cosmetic, so I am approving.  @Patowhiz can you give the second approval?

rdstern:
@Vitalis95 and @Patowhiz this will be useful in Zambia, and I hope we can have something in the February release.  Other features can be added later, and it doesn't even need trace if that is holding things up?

instat/dlgExportToClimsoft.vb
```diff
@@ -1,3 +1,189 @@
...
+    End Sub
+
+    Private Sub ucrReceiverElements_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverElements.ControlValueChanged
+        ucrBase.clsRsyntax.lstBeforeCodes.Clear()
+        clsCurrentNewColumnFunction = ucrReceiverElements.GetVariables(True).Clone
+        clsCurrentNewColumnFunction.SetAssignTo("columns")
+        ucrBase.clsRsyntax.AddToBeforeCodes(clsCurrentNewColumnFunction)
+    End Sub
+
+    Private Sub GetStationName()
```
lloyddewit, line 159: a `Get` function normally returns something.
Rename this sub?

instat/dlgExportToClimsoft.vb
```diff
@@ -1,3 +1,189 @@
...
+            Dim clsGetVariable As RFunction = ucrReceiverStation.GetVariables
+            clsGetVariable.SetAssignTo("station")
+            clsDataFrameFunction.AddParameter("station", clsRFunctionParameter:=clsGetVariable, iPosition:=1)
+            clsPipeOperator.RemoveParameterByName("right")
+        End If
+    End Sub
+
+    Private Sub ucrReceiverStation_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverStation.ControlValueChanged
+        GetStationName()
+        Dim strVar As String = ucrReceiverStation.GetVariableNames
```
lloyddewit, line 181: remove?

lloyddewit (changes requested):
@Vitalis95 thanks, this is looking good, just a couple of small suggestions

Vitalis95:
@Patowhiz , could you test it?

rdstern (changes requested):
@Vitalis95 the one item I notice is that the initial focus is on the date, even though all fields are completed except the multiple receiver.  
I am not sure what is needed next?  If it will take long to implement the direct csv output then we could merge this version once @lloyddewit is ok with the code.

Vitalis95:
@rdstern , I am doing export part on separate PR,

lloyddewit (approved):

lloyddewit:
> the one item I notice is that the initial focus is on the date, even though all fields are completed except the multiple receiver.

@Vitalis95 if you can resolve this, then we can merge, thanks

Vitalis95:
@rdstern , the focus now is on the multiple receiver. Have look at it

rdstern (approved):

lloyddewit (approved):

Issue #8030, opened 2022-12-22
Adding keys for fractions and decimals to the transform keyboard


rdstern:
@derekagorhom could you add 2 more keys to the bottom of the Transform keyboard:

a) The key called `fractions`.  This will run the `MASS::fractions( ) ` function.
If you produce a variable from a dataframe with say 6 observations, then with a regular sequence go from 1 to 0 in steps of 0.15.

![image](https://user-images.githubusercontent.com/11226469/209232207-da226a1d-c9da-405d-9614-a7431ee42628.png)

Then the fractions key produces a character variable

b) The inverse is called `decimals`.  That's a bit more complicated, but it works from the script window, and I hope it will as a calculation:  It is very nice - if it works to have a key that includes a function!

```
b <- c("5/2","3","7/2")  #this is the character column
sapply(X = b, FUN = function(v) {sapply(X = v,FUN = function(w) eval(parse(text=w)))})
```
So the user will supply the column that is given as b here.

c) Add MASS as another package for the help file.  They are in alphabetical order so it comes before zoo.

N-thony:
@derekagorhom how is it going with this? I did a quick check on this and I think the changes need to be made on the R level.

PR #8074, opened 2023-01-23
Added MASS Package

derekagorhom:
fixes partly #8030 
@lloyddewit can you approve this


rdstern:
@derekagorhom can you check.  The MASS package is already used - see Model > Fit Model > Fit Model Keyboards where it has its own keyboard.  
So, isn't it included already?

derekagorhom:
I am deleting this since MASS is already inclued

PR #8076, opened 2023-01-23
Improvement of the code quality of receiver logic that checks that when a dialog is reopened, only columns that still exist are shown in the list of selected columns

### LINKED ISSUE #8073: Improve code quality of receiver logic that checks that when a dialog is reopened, only columns that still exist are shown in the list of selected columns
PR #7581 contains some post-merge comments from @Patowhiz .
@N-thony Please could you review and implement those that you agree with?
Thanks

N-thony:
Fixes #8073
@lloyddewit @Patowhiz I have the changes suggested in PR #7581. Please take a look. Thanks

### DIFF
--- a/instat/ucrReceiver.vb
+++ b/instat/ucrReceiver.vb
@@ -130,7 +130,7 @@ Public Class ucrReceiver
 
     End Sub
 
-    'This refers to the selector list of column
+    'This refers to the selector list of columns
     Public Overridable Sub RemoveAnyVariablesNotInList()
 
     End Sub
--- a/instat/ucrReceiver.vb
+++ b/instat/ucrReceiver.vb
@@ -130,7 +130,7 @@ Public Class ucrReceiver
 
     End Sub
 
-    Public Overridable Sub RemoveAnyVariablesNotInList(lstViewItem As ListView)
+    Public Overridable Sub RemoveAnyVariablesNotInList()
 
     End Sub
 
--- a/instat/ucrReceiver.vb
+++ b/instat/ucrReceiver.vb
@@ -130,6 +130,7 @@ Public Class ucrReceiver
 
     End Sub
 
+    'This refers to the selector list of column
     Public Overridable Sub RemoveAnyVariablesNotInList()
 
     End Sub
--- a/instat/ucrReceiverMultiple.vb
+++ b/instat/ucrReceiverMultiple.vb
@@ -312,10 +312,9 @@ Public Class ucrReceiverMultiple
     ''' Removes any variable in the multiple receiver
     ''' that is not in the list of variables of the selector
     ''' </summary>
-    ''' <param name="lstViewItem"></param>
-    Public Overrides Sub RemoveAnyVariablesNotInList(lstViewItem As ListView)
+    Public Overrides Sub RemoveAnyVariablesNotInList()
         For Each strVar In GetVariableNamesAsList()
-            If lstViewItem.FindItemWithText(strVar) Is Nothing Then
+            If Selector.lstAvailableVariable.FindItemWithText(strVar) Is Nothing Then
                 Remove({strVar})
             End If
         Next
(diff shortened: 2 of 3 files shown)


lloyddewit:
@N-thony It all looks good to me. Is the comment below also addressed?

image

N-thony:
> @N-thony It all looks good to me. Is the comment below also addressed?
> 
> image

@lloyddewit yes, the single receiver already clear the variable

lloyddewit (approved):

lloyddewit:
@Patowhiz Thank you for these helpful comments. If you have time, would you also like to review?

thanks

Patowhiz:
@lloyddewit @N-thony thank you for addressing my comments
@N-thony I'm interested in knowing why the single receivers automatically removed the variables yet the multiple receivers don't, until this addition. The 2 receivers somehow relate thus this functionality should relate explicitly in regards to the 2.
I was expecting to see the single receivers calling the same function that you have added, or do they internally call a function that does the same thing?

instat/ucrReceiver.vb
```diff
@@ -130,7 +130,7 @@ Public Class ucrReceiver
 
     End Sub
 
-    Public Overridable Sub RemoveAnyVariablesNotInList(lstViewItem As ListView)
+    Public Overridable Sub RemoveAnyVariablesNotInList()
```
Patowhiz, line 133: @N-thony thanks for this.
Could we have a comment here as well. Something that makes it clear that the list referred here is actually the selector.

N-thony: > @lloyddewit @N-thony thank you for addressing my comments @N-thony I'm interested in knowing why the single receivers automatically removed the variables yet the multiple receivers don't, until this addition. The 2 receivers somehow relate thus this functionality should relate explicitly in regards to the 2. I was expecting to see the single receivers calling the same function that you have added, or do they internally call a function that does the same thing?

The implementation of the single receiver has itself already the possibility the clear it when the variable no longer exist in the selector list, we can do the same if we do refactor the code, we can discuss more about it in our next meeting in Kakamega next week.

rdstern (changes requested):
Good one to include.  However look below

image

I deleted 2 variables from the data frame.  They deleted from the selector - which is great.  But only 1 of them was deleted from the receiver.

N-thony:
@rdstern I have tried also by deleting two variables and it seems to work fine
image
 and again but with a different dialogue i.e `Convert columns`, this time I have deleted 4 columns and it seems fine too
image

instat/ucrReceiver.vb
Patowhiz: If the single receiver has a way of removing the deleted variable from itself. Then we should look into how it does that, at what event and whether we should use the same approach for the multiple receiver. What we have here seems to suggest explicitly that both the single and multiple receiver need to have this functionality implemented in this subroutine block, currently that is not the case.
It would be helpful if you could look into how the single receiver does it, check if the multiple receiver could have similar implementation then generalize it in a way that is consistent for both of them. I can make suggestions if you could outline how the single receiver does it.

PS what about `RemoveVariablesNotInSelector()`  as the name?
 
I'm happy we discuss more on the same.

N-thony: That is what I'm planning to do next after this PR which attends on resolving your comments and Stephen. But I will not consider it as higher priority for now. This was just an alternative instead of using the previous the R Function.

Patowhiz:
@N-thony @rdstern it would be good to also test this implementation with wide data sets (around 3000 variables). Especially for dialogs that have multiple receivers(multiple).

instat/ucrReceiver.vb
Patowhiz: So the single receiver uses an R Function to remove variables that are not in the selector?

rdstern (approved):
@N-thony it now seems to work well.  I tried with the survey data and summarise and it was fine now.  Then with rockart, from DAAG, which has 641 variables and deleted the first 100 - and that was ok.
Then back to another copy of the usual survey and single receivers in 3 variable regression, deleting two of them.  All ok.  
Nice

lloyddewit (approved):

N-thony:
> @Patowhiz @N-thony I'm personally happy with the changes so I approved. I understand that there are still some additional and related code-quality improvements that you're considering. I understand that you'll discuss these together when you're both in Kakmega. This PR is a code quality change rather than a functionality change, so it's not essential for the end January R-Instat release. When the changes are complete and you're both happy with this PR, then I'll merge. Thanks

@lloyddewit I'm also happy with this and think this can be merged. I will consider further suggestions in another PR.

PR #8077, opened 2023-01-23
Added Additional Keys to Transform Keyboard

### LINKED ISSUE #8030: Adding keys for fractions and decimals to the transform keyboard

### LINKED ISSUE #8075: Add scale function to the Calculator > Transform keyboard
@derekagorhom Here is the current transform keyboard:

image

1. Could you add the scale function as the first key on line 5.
 The default is simply `scale( )`  You should add the arguments if the corresponding box is ticked.  The tooltip is "Centre and scale the data - usually by producing (x - xbar)/s 

2. At the same time can you make the bottom row neater, so there are just the 4 calculations (on the left) with the keys the same size as the other rows.

3. Different, and very small I hope.  When the Clear button is pressed it currently doesn't return the focus to the Expression.  Could it please do that.

derekagorhom:
fixes #8075 
fixes #8030 

This PR is ready for review.

### DIFF
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -72,6 +72,7 @@ Public Class ucrCalculator
         ttCalculator.SetToolTip(cmdSiginf, "signif(x,3) to round to 3 significant figures")
 
         ttCalculator.SetToolTip(cmdSortF, "sorts a vector into ascending or descending order. For example sort(c(5,7,4,4,3)) = (3,4,4,5,7)")
+        ttCalculator.SetToolTip(cmdScale, "Centre and scale the data - usually by producing (x - xbar)/s")
         ttCalculator.SetToolTip(cmdLag, "shift a variable down. For example lag(1:5) = (NA,1,2,3,4); lag(1:5,3) = (NA,NA,NA, 1,2)")
         ttCalculator.SetToolTip(cmdLead, "shift a variable up. For example lead(1:5) = (2,3,4,5,NA); lead(1:5;3) = (4,5, NA,NA,NA)")
         ttCalculator.SetToolTip(cmdDiff, "difference between successive elements. For example diff(c(1,4,3,7)) = (NA 3,-1,4)")
@@ -4379,6 +4380,14 @@ Public Class ucrCalculator
         OpenHelpPage()
     End Sub
 
+    Private Sub MASSToolStripMenuItem_Click(sender As Object, e As EventArgs) Handles MASSToolStripMenuItem.Click
+        CalculationsOptions()
+        If ucrInputCalOptions.GetText = "Transform" Then
+            strPackageName = "MASS"
+        End If
+        OpenHelpPage()
+    End Sub
+
     Private Sub ZooToolStripMenuItem_Click(sender As Object, e As EventArgs) Handles ZooToolStripMenuItem.Click
         CalculationsOptions()
         If ucrInputCalOptions.GetText = "Transform" Then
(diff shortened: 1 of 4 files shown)


rdstern (commented):
@derekagorhom great - already works well.
Can you add tooltips - I forgot.
For fractions Changes decimal data into a character variable with fractions.  So 1.5, 0.25 becomes 3/2, 1/4.
For decimals.  The inverse of the fractions key.  So changes 3/2. 1/4 back into 1.5, 0.25.

Can you also discuss with @N-thony that the result in fractions appears fine in the try control.  But it is presented as a numeric decimal back in the data.  So it is as though the transformation has done nothing.  I hope it can be left as a character variable in fractional form.  
This may be a separate issue and should not affect the merging of this pull request.

rdstern (changes requested):
@derekagorhom that all looks good now.

I have been experimenting and for the fractions key can you add as.character(  ).
So it becomes as.character(MASS::fractions(  ))

Then the results display nicely, as fractions, in the grid.  I assumed @N-thony had to change something, but it was us instead!

rdstern (changes requested):
Great - thanks.  
I was just about to approve, when I checked that you had added the parameters when the box was ticked!  
The command is then `scale( , center = TRUE, scale = TRUE)`

And useful to do the same with the MASS fractions key,

derekagorhom:
@rdstern can you review this again. 
Thanks

rdstern (dismissed):
Great

instat/ucrCalculator.vb
```diff
@@ -72,6 +72,9 @@ Public Class ucrCalculator
         ttCalculator.SetToolTip(cmdSiginf, "signif(x,3) to round to 3 significant figures")
 
         ttCalculator.SetToolTip(cmdSortF, "sorts a vector into ascending or descending order. For example sort(c(5,7,4,4,3)) = (3,4,4,5,7)")
+        ttCalculator.SetToolTip(cmdScale, "centre and scale the data - usually by producing (x - xbar)/s")
+        ttCalculator.SetToolTip(cmdMASSFractions, "changes decimal data into a character variable with fractions. So 1.5, 0.25 becomes 3/2, 1/4.")
+        ttCalculator.SetToolTip(cmdDecimals, "the inverse of the fractions key. So changes 3/2. 1/4 back into 1.5, 0.25.")
```
lloyddewit, line 77:
```suggestion
        ttCalculator.SetToolTip(cmdMASSFractions, "changes decimal data into a character variable with fractions. So 1.5 becomes 3/2, 0.25 becomes 1/4 etc.")
        ttCalculator.SetToolTip(cmdDecimals, "the inverse of the fractions key. So 3/2 becomes 1.5, 1/4 becomes 0.25 etc.")
```

lloyddewit (changes requested):

lloyddewit (approved):

PR #8092, opened 2023-01-30
Minor bug fix on the window number feature

### LINKED ISSUE #8086: Small change on the window number feature
@N-thony here is an example:

image

I needed to go to row 3226.  So I know it is in the 3000 range, and I type 4 - rather than 3!  I did that quite a lot with these data and found it quite annoying!

My suggestion, for a minimal change is that in the message above you change 1-125 into 0-124.  So then I enter 3 to get the 4th window - which is usually the 3000's.

At the same time I have added a place in the Help file - so could you add the help context 294.  It will come in the February update to the Help!

N-thony:
Fixes #8086 
@rdstern I left some comments in the corresponded issue, also fix a small bug existing in the code if the user do a mistake to enter nothing in the window number nud control.


rdstern (commented):
I am not sure what has changed.  I accept your hesitation and am happy for you to also check with @Patowhiz or @ChrisMarsh82 if you wish, before making the change.  Here I show you my minor inconvenience with the current setting.

image

I wanted to visit row 3226, so (as it is in the 3000's I think the natural window would be number 3?) Not a big deal, but I was searching for many rows and found it distracting that I always had to add 1 to the window number.

So, if we consider the windows numbered from 0 to 124 then window 3 would be 3000-3999 usually and that would make my life slightly simpler.

I had forgotten the text at the bottom and that should also change to start from 0 - so 1 - 125 becomes 0 - 124 in this instance.

N-thony:
@lloyddewit please could you have a look? Thanks

rdstern (changes requested):
@N-thony that's perfect for the rows.

But I notice the change has affected the columns!  Can you correct that and then we can merge.

image

rdstern (changes requested):
@N-thony it is now (I hope) ok on the columns.
But the initial rows say 0 to 999, while it is actually 1 to 999.  I was happy before with 1 to 1000.
Then I am happier that the rows on the following pages remain as before, namely 3001 to 4000, etc.

N-thony:
> @N-thony it is now (I hope) ok on the columns. But the initial rows say 0 to 999, while it is actually 1 to 999. I was happy before with 1 to 1000. Then I am happier that the rows on the following pages remain as before, namely 3001 to 4000, etc.

@rdstern I'm now a bit confused, following your comment previously, I think this is what the current implementation is doing now or should I consider that you don't want it again and then we should go back to the initial state?
image

N-thony:
> > @N-thony it is now (I hope) ok on the columns. But the initial rows say 0 to 999, while it is actually 1 to 999. I was happy before with 1 to 1000. Then I am happier that the rows on the following pages remain as before, namely 3001 to 4000, etc.
> 
> @rdstern I'm now a bit confused, following your comment previously, I think this is what the current implementation is doing now or should I consider that you don't want it again and then we should go back to the initial state? We just need to update the text at the bottom of the DataView image

And the message at the bottom shows the starting row displaying in grid which is 1 but the text says 0, isn't it a bit confusing for the user?
image

rdstern:
@N-thony I didn't want any changes on the line numbers within a window.  They currently go from 1 to 1000, then 1001 to 2000, etc.  That was and is still fine.

The only change I asked for was on the window numbers. - not the line numbers within a window!

image

So, in the above, the 1 to 29 becomes 0 to 28.  Nothing to do with the 1 to 1000!

It was also only a minor convenience feature.  Perhaps we should omit it.  If it is so confusing, then perhaps other users will also have problems.  I never realised it would take so long.

lloyddewit:
@rdstern 
I assume that most users naturally start counting at one. So they consider the first row to be row 1 and the first page to be page 1.
I understand that issue #8086 requests that the first page is page 0. I find this a bit strange but agree that users might intuitively (but incorrectly in my view) expect row 3001 to be on page 3.
If we make page numbers start at zero and row numbers start at one, then is this inconsistent and potentially confusing?
I don't mean to stir up trouble. If you don't find it confusing and prefer it this way, then I'm happy to accept your decision.

rdstern:
@N-thony I owe you now an apology and would like to abandon this change and pull request!  Great shame, because otherwise I would have been able to approve it now.  But it is clear that starting with 0 caused you some confusion and @lloyddewit was also confused.  So I also discussed with Rachel, and she agreed with Stephen - and I think you.  (I think I owe you a fish meal at Fuel Junction for inflicting this work on you!)  Could you close it please!  And without the change - we will continue (for me a bit reluctantly) at window number 1!

N-thony:
@lloyddewit @rdstern I think we can merge this, I found a small bug when the `nud control` is empty. It is fixed now, and I did revert the code.

lloyddewit (approved):

lloyddewit:
@rdstern 
I understand that this PR:
- has page numbers and row numbers that both start at one (the same behaviour as previously)
- fixes a bug in the nud control

If you can test/approve, then we can merge, thanks

rdstern (approved):
@N-thony looks fine - thanks and sorry for all the confusion.

PR #8093, opened 2023-01-30
Fixed bug renaming columns in bug on Prepare > Data Frame > Rename

### LINKED ISSUE #7938: Minor bug on Prepare > Data Frame > Rename
I think this is simple for @N-thony to do, or allocate, as it has happened on other dialogues.  But it just may be tricky - and then we have to leave it as a "feature" rather than a bug.

When you rename multiple and then click ok when you are still in the last variable you have renamed then this last variable is ignored.  I don't know enough to suggest that the Ok key should close the previous control, or clicking somewhere else could close it?  I continually type new names as a, b, c and then Ok, to find the new names are a and b but not c!

derekagorhom:
Fixes #7938 

This PR is fixes the bug on Prepare > Data Frame > Rename Column  

This PR  is ready for Review.

### DIFF
--- a/instat/dlgName.vb
+++ b/instat/dlgName.vb
@@ -550,9 +550,12 @@ Public Class dlgName
     Private Sub grdCurrentWorkSheet_CellMouseLeave(sender As Object, e As CellEditTextChangingEventArgs) Handles grdCurrentWorkSheet.CellEditTextChanging
         If grdCurrentWorkSheet.ColumnCount > 0 Then
             Dim iCol As Integer = e.Cell.Column
+            Dim parsedValue As Boolean
             Dim strNewData As String = ValidateRVariable(e.Text, iCol)
-            RenameColumns(strNewData, e.Cell.Row, iCol)
-            ValidateNamesFromDictionary(iCol)
+            If Not strNewData.ToLower.Equals("t") AndAlso Not strNewData.ToLower.Equals("f") AndAlso Not IsNumeric(strNewData) AndAlso Not Boolean.TryParse(strNewData, parsedValue) Then
+                RenameColumns(strNewData, e.Cell.Row, iCol)
+                ValidateNamesFromDictionary(iCol)
+            End If
         End If
     End Sub
 End Class
--- a/instat/dlgName.vb
+++ b/instat/dlgName.vb
@@ -547,7 +547,7 @@ Public Class dlgName
         TestOKEnabled()
     End Sub
 
-    Private Sub grdCurrentWorkSheet_CellMouseLeave(sender As Object, e As CellEditTextChangingEventArgs) Handles grdCurrentWorkSheet.CellEditTextChanging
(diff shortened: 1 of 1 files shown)


N-thony:
@rdstern if you test and you are happy with this, I suggest we do the same changes everywhere we use the Grid?

rdstern (changes requested):
@derekagorhom I wonder how carefully you have tested this change.  It is what I keep asking you to do more of, before asking others.
Please try with the survey dataset and change the 7 names to a, b, c, d, e, f, g.  You will find - as before it doesn't like having f as the name.

Ok, now explain to me how to add any name now that starts with either f or t?  Now, with your change I think it is checking for the acceptable name letter by letter, and that's a disaster!

derekagorhom:
@rdstern sorry for the new issue found with the PR but i think instat recognizes t and f as boolean so it stops a user from having then as column names, see below
image
apart from that when you click ok and add more letters to the t or f it accepts them and the user can successfully rename the column.

N-thony:
> 

@derekagorhom we need the warning message to avoid the user to enter `Boolean` as column name, we can just check the string before passing it to the function if it is not a `Boolean`.

derekagorhom:
@rdstern this new commit ensures the user does not get hit with the `instat` warning when they are renaming the columns with starting words `t or f` but the warning message will pop up if they leave it as such (leaving it as `t or f `and moving to a new column).
this message will be repeated till the user makes that correction.

rdstern (approved):
@derekagorhom great - this now works fine.
As mentioned by @N-thony there are other places where this could be a good change.  Particularly entering data into the grid.  If agreed then maybe this pull request can be merged, but the issue could be extended and this pull request adjusted to only partially resolve the issue.

instat/dlgName.vb
```diff
@@ -546,4 +546,16 @@ Public Class dlgName
     Private Sub ucrCoreControls_ControlContentsChanged() Handles ucrInputNewName.ControlContentsChanged, ucrReceiverName.ControlContentsChanged, ucrPnlOptions.ControlContentsChanged
         TestOKEnabled()
     End Sub
+
+    Private Sub grdCurrentWorkSheet_CellMouseLeave(sender As Object, e As CellEditTextChangingEventArgs) Handles grdCurrentWorkSheet.CellEditTextChanging
```
lloyddewit, line 550: - The sub name should be consistent with the handled event. I would expect a name like `grdCurrentWorkSheet_CellEditTextChanging`.
- Normally we should change events in the designer so that `dlgName.designer.vb` stays consistent with `dlgName.vb`. Should we add this new event handler using the VS Designer?

lloyddewit (changes requested):

instat/dlgName.vb
derekagorhom: @lloyddewit i have made the changes

lloyddewit (approved):

PR #8106, opened 2023-02-07
Ensured that 'Signif_Figures' column in Data Grid sets list variables ro read only

### LINKED ISSUE #8067: Error with list variables and precision
I hope the figure shows the error:

image

@N-thony this may be an easy fix, but please try the problem for yourself, because it is followed by a possible further problem.  It may need discussion with @ChrisMarsh82 ?

Get it in the now usual way for lists.  So a New data frame with values say 1 to 100 in x1.  Then the calculator and any list variable in x2 - say divisors.

Now get the column metadata.  The list variables have NA for the precision.  You see from some of the other variables above that it may sometimes be useful to change that.  So, above I made it 3 and that caused the problem shown.

I am ok if list variables (at least for now) don't allow the signif element to be changed, but I don't like this error occurring.

I started again and tried other options in the metadata, but this seems the only one that gives this problem. 

But also, before fixing, could you also check what happens after this error, because R-Instat can still be used, but it is painful.  So try ignoring the advice and see what happens too.  I am not sure about the value of the advice, which may occur elsewhere.

So what happens currently is as follows:
a) It closed the data frame and gave the front screen.  When I try with multiple data frames it just closes the offending one..
b) I then started again - but without closing R-Instat - because I hadn't done much.
c) It seems to work fine - except I get the same error message many times (well perhaps 5 times) each time I press Ok.  That's the same problem if I keep working on another data frame, when more than one were used initially.

N-thony:
Fixes #8067 
@ChrisMarsh82 I have made the change as discussed few minutes ago that to disable the cell in the `Signif_Figures` where the `class` is list in the Column Metadata grid. Please have a look.


ChrisMarsh82 (approved):

N-thony:
@lloyddewit overt to you

instat/UserControls/DataGrid/ReoGrid/ucrColumnMetadataReoGrid.vb
```diff
@@ -43,6 +43,9 @@ Public Class ucrColumnMetadataReoGrid
         For i = 0 To grdData.CurrentWorksheet.Rows - 1
             For j = 0 To grdData.CurrentWorksheet.Columns - 1
                 grdData.CurrentWorksheet(row:=i, col:=j) = columnMetaData.Data(i, j)
+                If grdData.CurrentWorksheet.Item(row:=i, col:=j) = "list" Then
+                    grdData.CurrentWorksheet.GetCell(row:=i, col:=GetColumnIndex("Signif_Figures")).IsReadOnly = True
```
lloyddewit, line 47: The `GetColumnIndex()` function will potentially be called multiple times.
If we call the function outside the `For` loops (i.e. above line 43), then we only need to call the function once.
Should we move this function call?
Should we also check for the function returning `-1`?

lloyddewit (changes requested):
@N-thony thanks, just one question

instat/UserControls/DataGrid/ReoGrid/ucrColumnMetadataReoGrid.vb
```diff
@@ -40,9 +40,13 @@ Public Class ucrColumnMetadataReoGrid
         rngDataRange = New RangePosition(0, 0, grdData.CurrentWorksheet.Rows, grdData.CurrentWorksheet.Columns)
         grdData.CurrentWorksheet.SetRangeDataFormat(rngDataRange, DataFormat.CellDataFormatFlag.Text)
 
+        Dim iCoumnIndex As Integer = GetColumnIndex("Signif_Figures")
         For i = 0 To grdData.CurrentWorksheet.Rows - 1
             For j = 0 To grdData.CurrentWorksheet.Columns - 1
                 grdData.CurrentWorksheet(row:=i, col:=j) = columnMetaData.Data(i, j)
+                If grdData.CurrentWorksheet.Item(row:=i, col:=j) = "list" Then
```
lloyddewit, line 47: if `iColumnIndex` is -1, then the line below will trigger an exception.
Add a check like `AndAlso iColumnInex >= 0`?

instat/UserControls/DataGrid/ReoGrid/ucrColumnMetadataReoGrid.vb
```diff
@@ -40,9 +40,13 @@ Public Class ucrColumnMetadataReoGrid
         rngDataRange = New RangePosition(0, 0, grdData.CurrentWorksheet.Rows, grdData.CurrentWorksheet.Columns)
         grdData.CurrentWorksheet.SetRangeDataFormat(rngDataRange, DataFormat.CellDataFormatFlag.Text)
 
+        Dim iCoumnIndex As Integer = GetColumnIndex("Signif_Figures")
```
lloyddewit, line 43: fix typo?
```suggestion
        Dim iColumnIndex As Integer = GetColumnIndex("Signif_Figures")
```

lloyddewit (changes requested):

instat/UserControls/DataGrid/ReoGrid/ucrColumnMetadataReoGrid.vb
N-thony: this seems fixed now.

lloyddewit (approved):

lloyddewit (approved):

lloyddewit:
@berylwaswa if you can test, then we can merge, thanks

instat/UserControls/DataGrid/ReoGrid/ucrColumnMetadataReoGrid.vb
lloyddewit: @N-thony I may have started commenting before you were done with your changes ... sorry

berylwaswa:
@ChrisMarsh82 @lloyddewit @N-thony I have checked this and the cell where the class is list in the Column Metadata grid the Signif_Figures is disabled. It works fine.

Issue #8104, opened 2023-02-07
Implement the New output system and fix existing bugs


Vitalis95:
@rdstern, After reviewing the `Describe `menu dialogs with @Patowhiz and @anastasia-mbithe, we identified the following issues that need to be addressed.

**Describe> One Variable** dialogs
- [x]  **Summaries dialog** 
- [x] - Adding save controls to save objects for all the options 

- [x]  **Frequencies dialog**
- [x] - For `Table` option, we remove the **Output option** i.e **Text & HTML** and add save control.
- [x] - For `Graph` Option, fixing the bug
- [x] - Disabling `Both` option for now.
- [ ] - Find a way for the `stem and leaf` option to save an output object.

- [x] **Rating dialog**
- [x] - Adding save control to `Table` option

- [x] **Describe> Two/Three Variables**
- [x]  **Correlation** dilaog
- [x] In `Multivariable` option;
- [x] - Changing save label name
- [x] - Design changes

- [x] **Two Way Frequencies**  dilaog
- [x] In `Table` option
- [x] - Adding save control 

- [x]  **Three Way Frequencies**  dilaog
- [x] In `Table` option
- [x] - Adding save control

**Describe> General**
- [x]  **Column summaries**  dilaog
- [x] - Investigating the `calculate summary` function
- [x] - Adding save options i.e `save table` & `save dataframe`

- [ ] **Describe >Multivariate**
- [ ]  **Principal Component Analysis**  dilaog
- [ ] - Fix a bug where the receivers does not allow  to add variables
- [ ] - Adding save model for the new output system

- [x]  **Canonical correlation**  dilaog
- [x] - Fixing the bug

- [ ] **Cluster Analysis**  dilaog
- [ ] - Reviewing the design
- [ ] - Adding the save control for the new out put window

lilyclements:
Fixed in PRs attached, or, in those I couldn't see PRs for (Stem + Leaf, PCA dlg), checked manually and these problems are fixed

PR #8111, opened 2023-02-08
Implemented new output system in Describe >One variable >Frequencies dialog

anastasia-mbithe:
Fixes partially #8104 

@rdstern, @Patowhiz and @lloyddewit 
I have implemented the new output system in this dialog.
1. I have disabled the` Both option` since the new output system does not support producing two objects simultaneously. From the discussion with @Patowhiz, it's so easy for a user to go back and produce the other object when needed.
2. The dialog uses only one save control now.
3. The panel that initially controlled save as text(viewer) or HTML (browser) now controls save as a table and save as a data frame, for the `Table option`.


lloyddewit (approved):

lloyddewit:
@berylwaswa please could you test? thanks

berylwaswa:
@lloyddewit @anastasia-mbithe this looks good.
The Both option has been disabled and it's easy to go back and produce the other object.
The dialog uses only one save control.

PR #8113, opened 2023-02-08
Ensured Three-way Frequency dialog output is correct

derekagorhom:
fixes partly #8104 

@lloyddewit, @berylwaswa  This PR is ready for review.


berylwaswa:
@derekagorhom this works okay. With a large dataset, it takes some time but works fine.

lloyddewit (approved):

PR #8121, opened 2023-02-09
Reset the Scrollbar position on the DataView Grid  at the top when Filter is applied

### LINKED ISSUE #8116: Scrolling problem when data filtered
@N-thony I hope this will be simple to fix.  

When I apply a filter to a dataset, the immediate appearance depends on there I have scrolled to, in the main data.

[qcRain.zip](https://github.com/africanmathsinitiative/R-Instat/files/10694444/qcRain.zip)

To illustrate I think you can have any example - more than 1000 rows.  I think a solution may be simply to make sure the scrolling is at the top, before applying the filter.  

Otherwise the results seem to depend on where you have scrolled to.

In the example data I had a filter for December and also day number in month < 10.

a) If I am on row 1 when I apply the filter, then all is fine.
b) If I am halfway down then the screen is initially blank, but the 9 rows appear easily.
c) If I start at the bottom, then it is a pain - and rows appear one at a time as you press the top of the scroll bar.

As I said, I think if you could return to row 1, before doing the filter, then it should all be fine.

N-thony:
Fixes #8116
@rdstern @lloyddewit I made the change so that the scrollbar will always go at the top when the filter is applied. Have a look.


lloyddewit (approved):

lloyddewit:
@berylwaswa if you can test, then we can merge, thanks

rdstern:
@N-thony while you are here could you check the same thing "sideways".  I had 34 columns and only saw 30.  When I was at Variable 30 I went to the second page it was blank.  I think it is ok if I am at variable 1.

rdstern (commented):
@N-thony I confirm that the scrolling vertically seems fixed.  I then looked at scrolling horizontally with a new data frame of width 54 (my number of variables is set to 50) and hope the same can be done for this?
If you prefer it to be another pull request, then I'll approve this one.

N-thony:
The solution here fix the issue but it doesn't solve the problem in the Grid in general because the same problem found with the columns is also with the rows. I will try to fix it more general with the Grid.

rdstern (approved):
I am approving, because the sideways problem can be in a separate pull request.  And it would be good to have this feature fixed in time for the March workshops

PR #8131, opened 2023-02-10
Fixed memory leak

ChrisMarsh82:
Updated logging to log memory usage when opening the calculator
Updated translations to pass controls ByRef

@rdstern Hopefully this fixes the memory issues. It wasn't specific to the calculator but a problem with all forms. I have added extra logging around the calculator so if the problem still exists we should be able to see what's happening. 

About the logging.
It's still early of what we are putting in the logs and we will keep adding to them. 
The logs are stored in "../AppData/Roaming/RInstat/logs". 
What we log is defined in NLog.config. Therefore we can change the logging levels after the install. By default, we log 'Debug' and above.
The memory usage is under 'Trace' so the log file will need to be updated if we get the issue again. Change 'minlevel="Debug"' to 'minlevel="Trace"' in the NLog.config file.

### DIFF
--- a/instat/Translations.vb
+++ b/instat/Translations.vb
@@ -27,7 +27,7 @@ Public Class Translations
     '''
     ''' <param name="clsControl">    The WinForm control to translate. </param>
     '''--------------------------------------------------------------------------------------------
-    Public Shared Sub autoTranslate(clsControl As Control)
+    Public Shared Sub autoTranslate(ByRef clsControl As Control)
         If IsNothing(TryCast(clsControl, Form)) Then
             Exit Sub
         End If
@@ -44,7 +44,7 @@ Public Class Translations
     ''' <param name="tsCollection">     The WinForm menu items to translate. </param>
     ''' <param name="ctrParent">        The WinForm control that is the parent of the menu. </param>
     '''--------------------------------------------------------------------------------------------
-    Public Shared Sub translateMenu(tsCollection As ToolStripItemCollection, ctrParent As Control)
+    Public Shared Sub translateMenu(ByRef tsCollection As ToolStripItemCollection, ctrParent As Control)
         ' The 'WriteCsvFile' function call below should normally be commented out. 
         ' It only needs be uncommented and executed once, prior to each new release.
         'WriteCsvFile()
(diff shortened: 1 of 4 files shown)


instat/Translations.vb
```diff
@@ -27,7 +27,7 @@ Public Class Translations
     '''
     ''' <param name="clsControl">    The WinForm control to translate. </param>
     '''--------------------------------------------------------------------------------------------
-    Public Shared Sub autoTranslate(clsControl As Control)
+    Public Shared Sub autoTranslate(ByRef clsControl As Control)
```
lloyddewit, line 30: @ChrisMarsh82 
I understand that the default is `ByVal`. For an object, this will be a reference to the object. If we specify `ByRef`, then we are also passing the object's reference. The only difference is that the `Sub` is able to change the reference to point to a new object (which we don't need to do here).
Therefore, I don't understand how adding `ByRef` could solve a memory leak.
Am I missing something?

ChrisMarsh82: @lloyddewit leave this with me. You are right this doesn't solve the issue. I must have changed how I was testing it.

ChrisMarsh82:
I now think the memory leak is to do with refreshing the dataview grid. The memory goes up when adding rows.

rdstern:
@ChrisMarsh82 I may have some supporting evidence for your theory.  It usually takes a long time for the memory problem to appear.  But I have just been testing changes in the calculator in general and the list keyboard in particular.  And the memory problem appeared pretty soon.  So after perhaps 15-20 minutes.  Of course each calculation added a new (list) variable to the grid.  So that's consistent with your observation above.

ChrisMarsh82:
@rdstern I now think this is no memory leak and our problem comes from bringing back lists. In my tests the list I bring back gets steadily larger as it goes down the rows. If I bring back 1000 rows I get a memory issue. If I only bring back the first 50 I have no issues and the memory usage goes back down. 
We have a couple of options
Change the default rows that are displayed to 100 (also change the columns to 50 or 100 - I know this was already suggested by yourself)
Calculate memory usage when refreshing the gird and warn users if it goes over a certain size, then recommend displayed rows be changed. - This can have a few issues as the used memory may not be that accurate and I'm not sure of the limit we should add as it will change dramatically from PC to PC.

rdstern:
@ChrisMarsh82 I can imagine lists in the grid might make the problem worse, but we started having the problem before I started working on lists.  And, in the workshop, there were at least 2 instances on other machines, that had nothing to do with lists.  But did occur when R-Instat was left open for a long time - over a day in at least one of those instances.

In the workshop the row limit of 1000 works pretty well.  I would not want to reduce it to 100 as a default.  And when I got the memory problem quickly - and with lists - the limit was 1000, but I was only using data frames of 100?

Damn - I was enjoying the potential of the lists too!

By the way, does it still refresh the grid, when the grid is not visible?

Let's discuss tomorrow.  Looking forward.

Roger

ChrisMarsh82:
This is not required as there is no memory leak just inefficiencies with the grid.

Issue #8094, opened 2023-01-31
Improve the Describe > One Variable > Frequencies and Rating dialogue for the new html and objects changes


rdstern:
@Patowhiz, @derekagorhom (and @N-thony )

This started as a more general issue about all the One Variable and Two/Three variable summary dialogues.  But then I found enough to consider and do on just these two dialogues.  And a simple and useful starting point too

I suggest it is an excellent for Derrick to work on, after reflection by Patrick and then his guidance next week.  Though it can start this week if you both have the time.  

As data, you can try the One Variable > Frequencies with the usual survey data.  For the rating data use the efc data from sjmisc in the library - variables c82cop1 to cop9.

1. Compare the code for the current one variable frequencies, with Rating, and with the 2-variable frequencies that has already been updated.:

```
# Using the dialog: Frequencies

sjmisc::frq(x=data_book$get_columns_from_data(data_name="survey", col_names=c("village","field","size","fert","variety","yield","fertgrp"), force_as_data_frame=TRUE, remove_labels=TRUE, drop_unused_filter_levels=TRUE), out="txt")
```

Compare with the code for the rating dialogue:

```
# Using the dialog: Rating Data

columns <- data_book$get_columns_from_data(data_name="survey", col_names=c("fert","size"), use_current_filter=FALSE)
sjPlot::tab_stackfrq(items=columns, alternate.rows=TRUE, show.n=TRUE, show.total=TRUE, show.na=TRUE)
rm(columns)

```
and compare both of these with the 2-way frequencies that (I think?) Patrick has changed for the new system.

```
# Using the dialog: Two Way Frequencies

survey <- data_book$get_data_frame(data_name="survey", drop_unused_filter_levels=TRUE)
last_table <- sjPlot::sjtab(data=survey, village, variety, show.summary=FALSE, digits=0, fun="xtab", title="", string.total="Total")
data_book$add_object(data_name="survey", object_name="last_table", object_type_label="table", object_format="html", object=last_table)
data_book$get_object_data(data_name="survey", object_name="last_table", as_file=TRUE)
rm(list=c("last_table", "survey"))

```
And for reference the new code for skim from the summary dialogue.

```
# Using the dialog: One Variable Summarise

survey <- data_book$get_data_frame(data_name="survey")
last_summary <- skimr::skim_without_charts(data=survey, c("village","field","size","fert","variety","yield","fertgrp"))
[...]

PR #8146, opened 2023-02-16
Ensured Output system for One Way frequencies worked correctly

derekagorhom:
fixes partly #8094 
@rdstern , @N-thony 
I was able to fix the output system for the `Table` and `graph` options of the dialogue and also i fixed the issue of the `stem and leaf` not producing any results


rdstern (changes requested):
@derekagorhom that is much better now.  Could you change the code generated for the Table option, so it separates getting the variables from the sj function.  It now seems ok for the graph and I am happy we have the stem and leaf plot again.

derekagorhom:
@rdstern 
i did the code for the output to display

`columns <- data_book$get_columns_from_data(data_name="survey", col_names=c("village","field"), remove_labels=TRUE, drop_unused_filter_levels=TRUE)`

`one_way_freq <- as.data.frame(x=sjmisc::frq(x=columns, show.strings=TRUE, show.na=TRUE, grp.strings=NULL))
data_book$import_data(data_tables=list(one_way_freq=one_way_freq))`

`data_book$get_data_frame(data_name="one_way_freq")`

`rm(list=c("one_way_freq", "columns"))` 

on the script widow.

rdstern:
@derekagorhom thanks, this now seems to work fine.
However I am afraid I have stumbled on another small problem.  The earleir version permitted text or html output, with the html going to a browser.  Now we have html in the output window, but seem to have lost that option.  In the sjmisc::frq command it should come from adding out="viewer" to the function, and then the html option to the data_book_add_object function.  
This adds a third option to the dialogue.  @Patowhiz alternatively we now just have html as the output format in the output window?  What do you think?  I quite like the options of text or html, but don't feel strongly?

derekagorhom:
@rdstern i managed to add the maximise to the output for the dataframe button.

instat/dlgOneWayFrequencies.vb
```diff
@@ -22,6 +22,7 @@ Public Class dlgOneWayFrequencies
     Private clsSjMiscFrq As New RFunction
     Private clsSjPlot As New RFunction
     Private clsPlotGrid As New RFunction
+    Private clsGetColumnsFunction As New RFunction
```
N-thony, line 25: Can you move this on before line 330 i.e Dim clsGetColumnsFunction As New RFunction?

instat/dlgOneWayFrequencies.vb
```diff
@@ -164,6 +165,7 @@ Public Class dlgOneWayFrequencies
         clsAsDataFrame = New RFunction
         clsStemAndLeafFunction = New RFunction
         clsDummyFunction = New RFunction
+        clsGetColumnsFunction = New RFunction
```
N-thony, line 168: then you remove this

instat/dlgOneWayFrequencies.vb
```diff
@@ -216,7 +221,7 @@ Public Class dlgOneWayFrequencies
         ucrSaveGraph.AddAdditionalRCode(clsAsDataFrame)
 
         ucrReceiverWeights.SetRCode(clsSjMiscFrq, bReset)
-        ucrReceiverOneWayFreq.SetRCode(clsSjMiscFrq, bReset)
+        'ucrReceiverOneWayFreq.SetRCode(clsSjMiscFrq, bReset)
```
N-thony, line 224: remove?

instat/dlgOneWayFrequencies.vb
```diff
@@ -183,6 +185,9 @@ Public Class dlgOneWayFrequencies
 
         clsSjMiscFrq.SetPackageName("sjmisc")
         clsSjMiscFrq.SetRCommand("frq")
+        clsSjMiscFrq.AddParameter("show.strings", "TRUE")
+        clsSjMiscFrq.AddParameter("show.na", "TRUE")
+        clsSjMiscFrq.AddParameter("grp.strings", "NULL")
```
N-thony, line 190: Can you add the positions?

N-thony (changes requested):

derekagorhom:
@N-thony I have made the changes requested.

rdstern (changes requested):
@derekagorhom perhaps I didn't make myself clear last time.  Here is the dialogue in Version 0.7.8

image

You see that html output in the browser was then possible.  So now we no longer have html - and just when this is now an option in the output window.

You could add it, so the first option is txt, the second is the html - this time in the output window.  (That was supposed to be the main change with the new output window.)

I don't have a problem with the addition of the new option as dataframe, so that can stay if you wish.

Patowhiz:
@derekagorhom I think we discussed with @anastasia-mbithe and @Vitalis95 about similar options. Check with them.

derekagorhom:
@rdstern I was able to implement the HTML option to this dialogue, but the problem is that it is not able to produce any results in the output window so i made the result display in the browser. 
I have had conversations with both Anastasia and Vitalis and we still don't know why it does not produce the HTML result in the output.

Patowhiz:
@derekagorhom I'll now work on this.

rdstern:
@Patowhiz will you still be doing this?  I am keen to "finish" all these options for the Describe > One Variable and 2/3 Variable as soon as possible.

N-thony:
> @Patowhiz will you still be doing this? I am keen to "finish" all these options for the Describe > One Variable and 2/3 Variable as soon as possible.

@derekagorhom is this fixed/replaced by another PR already? If so, can you close this PR?

PR #8148, opened 2023-02-17
Ensured that Default Column and Variable Name checkbox function correctly in New DataFrame dialog

### LINKED ISSUE #8120: Minor bug in the File > New Dataframe Dialogue
@derekagorhom when I try to change the default it becomes black which makes it hard to change. Can you fix this?
image

derekagorhom:
fixes #8120 
This is not ready for review as it needs further testing. 
@N-thony , the fix works but it takes longer to load (probably from my laptop). 
Also i found out that when you set the column to say 14 and reduce it to say 8, you get a column number between those 2 numbers.


rdstern:
@derekagorhom while you are fixing that, I would like to explore with the `Variable Name` checkbox checked by default.  Could you please make that change too.

derekagorhom:
@rdstern, i and @N-thony managed to reduce the loading time for the columns till approximately 290 columns, after that dialogue takes sometime to load the columns (but not as long as previously). 
I believe this is ready for testing

rdstern (approved):
@derekagorhom this looks absolutely fine.  @lloyddewit over to you.

lloyddewit (approved):

PR #8152, opened 2023-02-20
Implementing the Find In Variable and Filter Dialogue

### LINKED ISSUE #7950: Add a simple Find dialogue into the Edit menu?

N-thony:
Fixes #7950
@rdstern following your comment in the corresponding issue, I have implemented the find into different pages. Could you have a look? The find in Filter is disable for now.


rdstern (commented):
So that's a bit better, but not great as a general find yet.
Try with the diamonds data and caret and pattern 3. - there are about 35 of those in the record.  But I have to step manually down.  

I don't know how you are doing it, but could ytou use R to find them all on all pages and then be able to step down.

It will also be good to have a From Row... but you need to be able to find on all pages first.  Maybe also try with the filter?

N-thony:
@ChrisMarsh82 I have made some changes to find rows with R. Can you have a look? They are still some minor things I need to fix but so far this seems to work.
@rdstern could you also test quickly and give the feedback? Probably will make the changes once.

rdstern:
@N-thony that's looking good.  It seems to work fine.
a) I tried some values that don't exist and it gave a message.  That could improve.  I wonder if it should be a popup or within the dialogue?
b) It goes round, and when you start a new text, it always starts at the top.  Could you add a Find Next button below, which starts from the current place - even for a new Finf pattern or variable and just goes to the end.  (So the Find option for a new variable, or pattern) always starts at row 1 and goes round, while Find next always starts from the current row and only goes down?
c) Can you add a regex option/keyboard as you have added in the ordinary dialogues, and possibly an ignore case option?

d) Oh here is a slight inconvenience:

image

It doesn't have the same pages anymore.  And I sometimes want to see the observation in context, so I wanted here to scroll up.  Can you keep the ordinary paging at 1000?

e) I like the green cell being marked.  But could you perhaps add a radio button, if easy, with Show (O) Cell (O) Row.  (You can see in the example shown I made a logical variable for rain > 100 and searched for when that is true.  

f) So, in addition maybe the pattern could have a drop down with suggestions.  The obvious ones are TRUE and FALSE.

Roger

rdstern (commented):
@N-thony I had a quick look though you are still working on it.  I used the dodoma data.  When I started ar the top and looked for day 31 it went to the next page and gave me row 1035.  Then went to the next page?
I notice the line option isn't working yet..

N-thony:
@rdstern you can test this.

rdstern:
@N-thony I got build errors

PR #8154, opened 2023-02-20
Added data export to Export for Climsoft dialog

Vitalis95:
Fixes Partly #6377 
@rdstern , @N-thony , @lloyddewit , I have implemented export data frame part. I am still working on the remaining part of  export comments


rdstern (approved):
@lloyddewit this looks ok to me.  If the code is ok, then it would be good to include this in the forthcoming release.  I don't know whether you want @Patowhiz to also look?  
I would like the export of the Comments sheet to be added, but it needn't be this pull request.  If added, then I suggest we don't need to choose another file name.  Instead the same name is used as for the data file, with perhaps _report added to the name.

So there could be an Add Report checkbox.  If checked it gives you the list of data frames and a single receiver.  If there is a .comments data frame then this is filled by default.

lloyddewit (approved):

lloyddewit:
@rdstern Thank you for the approval and comments.

> @lloyddewit this looks ok to me. If the code is ok, then it would be good to include this in the forthcoming release. I don't know whether you want @Patowhiz to also look?

I don't think there's anything controversial in the code so no essential need for @Patowhiz to review from the code perspective. However, he may wish to review the PR's functionality to ensure that it's compatible with Climsoft. Based on your approval, and to avoid delay, I'll merge anyway and we can make any future changes in a new PR.

> I would like the export of the Comments sheet to be added, but it needn't be this pull request. If added, then I suggest we don't need to choose another file name. Instead the same name is used as for the data file, with perhaps _report added to the name. So there could be an Add Report checkbox. If checked it gives you the list of data frames and a single receiver. If there is a .comments data frame then this is filled by default.

Will you raise a new issue for this?

Patowhiz:
@lloyddewit thanks for merging it. Now I will look at it and make further amendments.

PR #8157, opened 2023-02-21
Ensured that selecting variables works correctly in Climatic > Tidy and Examine > Tidy Daily Data dialog

### LINKED ISSUE #8155: Serious problem with the Climatic > Tidy and Examine > Tidy Daily Data dialogue
@Vitalis95 could you please look at this problem urgently.  

It has appeared in Version 0.7.8 and was not there (I gather) in Version 0.7.6.

You can try with any dataset that has the 31 days going across - so Marrupa from the Mozambique climatic is an example (make -- into missing) and separate the year and month.  Then the multiple receiver goes crazy when you tyy to fill it!

There is a similar problem with the 12 months going across - The Garoua data is an example here.  I have not checked the default third option of the years columns.

I assume it should be easy to find.  Then it would be good to understand why the change has happened in case there is a similar problem in some other dialogues?

Vitalis95:
Fixes #8155 
@rdstern @lloyddewit @N-thony , fixes a bug when variables added to the receiver are excluded from the selector


rdstern (changes requested):
@Vitalis95 great - that's much better.  

But it didn't "remember" the variables.  So, when I returned to the dialogue, the receiver was cleared.

Vitalis95:
@rdstern , have a look at it.

rdstern (approved):
Excellent.  @lloyddewit this was a nasty bug in the climatic, so good if we can get the corrections into the forthcoming release

lloyddewit (approved):

PR #8161, opened 2023-02-22
Ensured that options to carry all/selected columns works correctly in Stack dialog

### LINKED ISSUE #8156: Minor problem with the (new?) Stack dialogue
@N-thony can you sort this one, or allocate?

This seems to have escaped our quality control!  It would be good - and I assume simple to fix for this release, but not serious if it remains there for another month.

image

On the rhs you can choose to carry all columns - default.  Or you choose the columns to carry.

But notice that when you choose the columns to carry, then the Carry all columns checkbox is disabled, but still checked.  So it always carries all columns.  I assume these options should be radio buttons instead, as they are alternatives.

anastasia-mbithe:
Fixes #8156 
@rdstern and @N-thony This is ready for review.


rdstern (approved):
Looks fine now.  @lloyddewit over to you

lloyddewit (approved):

PR #8166, opened 2023-02-23
Changed grid to only update one cell if replace value is done

ChrisMarsh82:
Updated ReplaceValueInCell to only update the changed cell.

At the moment we still bring back the entire dataframe from R so there is still a delay. This would be a much bigger change and could run into errors. This change should at least half the time taken to refresh

### DIFF
--- a/instat/UserControls/DataGrid/Linux/ucrDataViewLinuxGrid.vb
+++ b/instat/UserControls/DataGrid/Linux/ucrDataViewLinuxGrid.vb
@@ -101,6 +101,7 @@ Public Class ucrDataViewLinuxGrid
         RaiseEvent IDataViewGrid_ReplaceValueInData(dataGrid.CurrentCell.Value.ToString(),
                         GetCurrentDataFrameFocus().clsVisibleDataFramePage.lstColumns(dataGrid.CurrentCell.ColumnIndex).strName,
                         GetCurrentDataFrameFocus().clsVisibleDataFramePage.RowNames()(dataGrid.CurrentCell.RowIndex))
+        RefreshSingleCell(dataGrid.CurrentCell.ColumnIndex, dataGrid.CurrentCell.RowIndex)
     End Sub
 
     'ToDo allow editing
@@ -145,5 +146,10 @@ Public Class ucrDataViewLinuxGrid
 
     Private Sub tcTabs_SelectedIndexChanged(sender As Object, e As EventArgs) Handles tcTabs.SelectedIndexChanged
         RaiseEvent WorksheetChanged()
+    End Sub
+
+    Public Sub RefreshSingleCell(iColumn As Integer, iRow As Integer)
+        Dim dataGrid = GetDataGridFromSelectedTab()
+        dataGrid.Rows(iRow).Cells(iColumn).Value = GetCurrentDataFrameFocus.DisplayedData(iRow, iColumn)
     End Sub
 End Class
--- a/instat/UserControls/DataGrid/Linux/ucrDataViewLinuxGrid.vb
+++ b/instat/UserControls/DataGrid/Linux/ucrDataViewLinuxGrid.vb
@@ -148,7 +148,7 @@ Public Class ucrDataViewLinuxGrid
         RaiseEvent WorksheetChanged()
     End Sub
 
-    Public Sub RefreshSingleCell(iColumn As Integer, iRow As Integer)
+    Private Sub RefreshSingleCell(iColumn As Integer, iRow As Integer)
         Dim dataGrid = GetDataGridFromSelectedTab()
         dataGrid.Rows(iRow).Cells(iColumn).Value = GetCurrentDataFrameFocus.DisplayedData(iRow, iColumn)
     End Sub
--- a/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
@@ -200,5 +200,4 @@ Public Class ucrDataViewReoGrid
         End If
     End Sub
 
-
 End Class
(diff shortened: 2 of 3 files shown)


rdstern (dismissed):
@ChrisMarsh82 I tried on a large dataset, where I had noticed the problem before.  It had 875,000 rows.
I needed to change temperaures measured such as 305 degrees to 30.5.  There were about 11 rows like this.  I first tried with a filter, so just the 11 rows were visible.  A cell change took about 4 seconds.  I bit long, but not impossible.
When I noted the row number and tried on the unfiltered data it took only 2 seconds.

I tried with your stated change and also (for comparison) with the released version.  There seemed no change in those times, between the 2 versions.

I had no problem with your new version, so am approving.

instat/UserControls/DataGrid/Linux/ucrDataViewLinuxGrid.vb
```diff
@@ -146,4 +147,9 @@ Public Class ucrDataViewLinuxGrid
     Private Sub tcTabs_SelectedIndexChanged(sender As Object, e As EventArgs) Handles tcTabs.SelectedIndexChanged
         RaiseEvent WorksheetChanged()
     End Sub
+
+    Public Sub RefreshSingleCell(iColumn As Integer, iRow As Integer)
```
lloyddewit, line 151: - should this sub be private?
- or could we copy these lines to line 104 and remove this sub?

instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
```diff
@@ -94,6 +94,10 @@ Public Class ucrDataViewReoGrid
         grdData.CurrentWorksheet.RowHeaderWidth = TextRenderer.MeasureText(strLongestRowHeaderText, Me.Font).Width
     End Sub
 
+    Public Sub RefreshSingleCell(iColumn As Integer, iRow As Integer)
```
lloyddewit, line 97: - should this sub be private?
- or could we copy this line to line 175 and remove this sub?

instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
```diff
@@ -195,4 +200,5 @@ Public Class ucrDataViewReoGrid
         End If
     End Sub
 
+
```
lloyddewit, line 203:
```suggestion
```

instat/UserControls/DataGrid/Linux/ucrDataViewLinuxGrid.vb
```diff
@@ -101,6 +101,7 @@ Public Class ucrDataViewLinuxGrid
         RaiseEvent IDataViewGrid_ReplaceValueInData(dataGrid.CurrentCell.Value.ToString(),
                         GetCurrentDataFrameFocus().clsVisibleDataFramePage.lstColumns(dataGrid.CurrentCell.ColumnIndex).strName,
                         GetCurrentDataFrameFocus().clsVisibleDataFramePage.RowNames()(dataGrid.CurrentCell.RowIndex))
+        RefreshSingleCell(dataGrid.CurrentCell.ColumnIndex, dataGrid.CurrentCell.RowIndex)
```
lloyddewit, line 104: This is an extra call. How does this increase performance?

instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
```diff
@@ -168,6 +172,7 @@ Public Class ucrDataViewReoGrid
                            GetCurrentDataFrameFocus().clsVisibleDataFramePage.lstColumns(e.Cell.Column).strName,
                            GetCurrentDataFrameFocus().clsVisibleDataFramePage.RowNames()(e.Cell.Row))
         e.EndReason = unvell.ReoGrid.EndEditReason.Cancel
+        RefreshSingleCell(e.Cell.Column, e.Cell.Row)
```
lloyddewit, line 175: This is an extra call. How does this increase performance?

instat/ucrDataView.vb
```diff
@@ -22,6 +22,7 @@ Imports instat.Translations
 Public Class ucrDataView
     Private _clsDataBook As clsDataBook
     Private _grid As IDataViewGrid
+    Private OnlyUpdateOneCell As Boolean = False
```
lloyddewit, line 25:
```suggestion
    Private bOnlyUpdateOneCell As Boolean = False
```

lloyddewit (changes requested):
@ChrisMarsh82 Thank you for the changes, just some small comments.
Also, I understand that the purpose of this PR is to increase performance. During his testing, @rdstern found the speed similar to the previous version. Before we increase the code complexity, I think we should demonstrate a speed improvement. Do you have a test case that could show this?

instat/UserControls/DataGrid/Linux/ucrDataViewLinuxGrid.vb
ChrisMarsh82: made private

instat/ucrDataView.vb
ChrisMarsh82: updated naming

instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
ChrisMarsh82: made private

ChrisMarsh82:
@rdstern @lloyddewit  The changes I have made only stop the grid being fully loaded from the dataframe. The dataframe is fully refreshed so in some cases it won't appear any quicker. 

Where the benefit will come in if you have a large dataset with a lot of rows and where those rows contain lists with lots of values.
For example: 
create a dataset with one column with numbers 1 to 1000 (note: create a new dataset is broken at the minute so it will need to contain 3 columns)
add a couple more columns using the calculator "lapply( x1, function(x) {lapply(x, function(i) {choose(i, 0:i)})})"
Now when a value is changed although it still goes back to R to refresh the dataframe it doesn't need to repopulate the list in the grid which was taking a very long time

instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
ChrisMarsh82: Although it is an extra call here ReplaceValueInData now doesn't repopulate the list

lloyddewit (approved):

PR #8169, opened 2023-02-24
Fixed regression bugs on the New Data Frame dialog

### LINKED ISSUE #8167: New bug with File > New Data Frame dialogue
@derekagorhom this was thankfully discovered by @ChrisMarsh82 - and was partly my fault that it was introduced! (and then not tested fully enough!)

a) You arranged that the Variable Name checkbox is checked by default.  I still like that, but it has helped to cause the new problem.  When that is unchecked the dialogue works fine for 2 variables. (Or even for one variable)  That's fine.  But when you reopen the dialogue the Variable Name control is checked again.  It should remember its previous setting and so remain unchecked if I uncheck it the last time.
b) And, when it is checked, it now gives multiple error.  At least it does with the 2 variables and 10 rows, which is the default.  Then it seems to crash, which is what Chris found.
c) It seems sort of ok after settling in to being used a few times without the Variable Names chec.  But, I also then tried with 4 variables, and made the second variable numeric.  It ignored that as well, see below.  So it needs correction for when other changers are made.  It ignores, for example, the changes in the variable names - and that worked ok before. 

image

Maybe you will need help from @N-thony in the changes to come.

derekagorhom:
Fixes #8167 

This fixes the recent bug found when the dialogue fails to create a dataframe with less than 3 columns. 

@ChrisMarsh82 , @rdstern, this PR is ready for review.

### DIFF
--- a/instat/dlgNewDataFrame.vb
+++ b/instat/dlgNewDataFrame.vb
@@ -122,8 +122,6 @@ Public Class dlgNewDataFrame
         clsDummyVarFunction.AddParameter("variablename", "True", iPosition:=0)
         clsDummyVarFunction.AddParameter("checked", "defaults", iPosition:=1)
 
-
-
         clsGetCategoriesFunction.SetPackageName("rcorpora")
         clsGetCategoriesFunction.SetRCommand("categories")
         Dim expCategoryNames As SymbolicExpression = frmMain.clsRLink.RunInternalScriptGetValue(clsGetCategoriesFunction.ToScript(), bSilent:=True)
--- a/instat/dlgNewDataFrame.vb
+++ b/instat/dlgNewDataFrame.vb
@@ -87,7 +87,7 @@ Public Class dlgNewDataFrame
 
         ucrChkVariable.SetText("Variable Name")
         ucrChkVariable.SetParameter(New RParameter("var", 0))
-        ucrChkVariable.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
+        'ucrChkVariable.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
 
         ucrChkIncludeLabel.SetText("Variable Label")
         ucrChkIncludeLabel.SetParameter(New RParameter("label", 0))
@@ -120,7 +120,10 @@ Public Class dlgNewDataFrame
         'reset the controls
         ucrNewDFName.Reset()
         ucrTryNewDataFrame.SetRSyntax(ucrBase.clsRsyntax)
-        'clsDummyVarFunction.AddParameter("variablename", "True", iPosition:=0)
+        clsDummyVarFunction.AddParameter("variablename", "True", iPosition:=0)
+        clsDummyVarFunction.AddParameter("checked", "defaults", iPosition:=1)
+
+
 
         clsGetCategoriesFunction.SetPackageName("rcorpora")
         clsGetCategoriesFunction.SetRCommand("categories")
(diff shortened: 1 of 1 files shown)


rdstern (changes requested):
@derekagorhom that's a releif that I think the main bug is fixed.  I couldn't get an error.  But there are still some more bits to correct.
a) As mentioned previously, when I untick the Variable Name checkbox it also works.  But when I return to the dialogue the Variable Name is ticked for me again.  The rule, in R-Instat is that the controls "remember" how they were left.  So please keep that control unticked if that was done the last time Ok was pressed.
b) I am also happy that changing a name in the variables now seems ok.  But changing the type to numeric, or factor doesn't work yet.  Please can those be fixed too.

rdstern:
@derekagorhom I hope this is still work in progress - but you need to comment more when you make changes, so I know when I should look again.  I am very keen that this is fixed in time for the February release this weekend.
So I now see the change has simply unticked the Variable Name (as it was originally).  That's not what I asked for.
I am happy that it starts by being ticked, as the initial default.
But when I then choose to untick it and press OK, it should then remember that setting for when I re-open the dialogue.

And the other problems - only character variables despite what I change - still remain.  Please also discuss with @N-thony because he did the original code and it worked fine then.

derekagorhom:
@rdstern i was not done yet... i am still making the changes

derekagorhom:
@rdstern i fixed the issue with the variable type and also made code changes to check variable name so that it is default and it can be ticked and unticked with no issues.

rdstern (approved):
@derekagorhom, well done it all seems to work fine now.
@lloyddewit it would be good to merge this, once you are happy with the code.  @ChrisMarsh82 found serious problems with the Version 0.7.8 version.

instat/dlgNewDataFrame.vb
```diff
@@ -87,12 +87,13 @@ Public Class dlgNewDataFrame
 
         ucrChkVariable.SetText("Variable Name")
         ucrChkVariable.SetParameter(New RParameter("var", 0))
-        ucrChkVariable.AddParameterValuesCondition(True, "variablename", "True")
-        ucrChkVariable.AddParameterValuesCondition(False, "variablename", "False")
+        'ucrChkVariable.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
```
lloyddewit, line 90:
```suggestion
```

instat/dlgNewDataFrame.vb
```diff
@@ -120,6 +121,9 @@ Public Class dlgNewDataFrame
         ucrNewDFName.Reset()
         ucrTryNewDataFrame.SetRSyntax(ucrBase.clsRsyntax)
         clsDummyVarFunction.AddParameter("variablename", "True", iPosition:=0)
+        clsDummyVarFunction.AddParameter("checked", "defaults", iPosition:=1)
+
+
```
lloyddewit, line 126:
```suggestion
        clsDummyVarFunction.AddParameter("checked", "defaults", iPosition:=1)
```

instat/dlgNewDataFrame.vb
```diff
@@ -182,10 +186,11 @@ Public Class dlgNewDataFrame
         ucrNewDFName.AddAdditionalRCode(clsListDfFunction, iAdditionalPairNo:=4)
         ucrNewDFName.SetRCode(clsConstructFunction, bReset)
         ucrChkIncludeLabel.SetRCode(clsDummyLabelFunction, bReset)
-        ucrChkVariable.SetRCode(clsDummyVarFunction, bReset)
+
```
lloyddewit, line 189:
```suggestion
```

instat/dlgNewDataFrame.vb
```diff
@@ -513,7 +519,7 @@ Public Class dlgNewDataFrame
                 Dim iRowHeader = i + 1
                 With dgrView.Rows
                     .Item(i).Cells(0).Value = iRowHeader
-                    .Item(i).Cells(1).Value = String.Concat("x", iRowHeader)
+                    .Item(i).Cells(1).Value = "x" & (iRowHeader)
```
lloyddewit, line 522:
```suggestion
                    .Item(i).Cells(1).Value = "x" & iRowHeader
```

lloyddewit (changes requested):

lloyddewit (approved):

lloyddewit:
@rdstern I noticed that this PR reversed some of the changes made in PR #8148 . Please could you test that the problem solved by PR #8148 is still fixed?
If you're happy, then we can merge, thanks

rdstern (approved):

PR #8184, opened 2023-03-01
Added Export Comments Feature to Export to Climsoft dialog

Vitalis95:
@rdstern , @lloyddewit , @N-thony , added export comments feature and also two buttons at the top. `Station` is disabled for now.


rdstern:
Good this is added.
a) Can you check what triggers the Ok.  I assume if any one of the 3 file or data frame options is ticked - and there is a file name - then Ok is enabled.

Currently if I unick everything, then Ok is still enabled:

image

b) Can you change the hour control.  Currently it (only) takes a value typed in.  Perhaps there could be a checkbox, just above, with label Single Value - default checked.  Then it gives the current control.  If unchecked, then nit expects a variable.

c) Now your comments option:

Currently it changes the whole shape of the dialogue - very odd

image

I wasn't able to check whether it always loads the comments by default.  It did sometimes.

rdstern:
@Vitalis95 this is coming on.

I suggest:
a) Have a group box for the alternative saving, with label Save:
b) Inside have the 3 options with the 2 main ones as radio buttons, so you do one or other.
c) The first one is called `New Data Frame`
d) The second is called `Export Data (csv)`
e) The third remains a checkbox called `Add Report`  It is disabled unless Export data is selected.

f) When I took the option to save as a new data frame is gave me an error, as shown:

image

g) Please check also trhe saving of the exported data.  I am not sure it works yet?
h) I suggest when you include Add Report you still just specify the one name for the data file.  Iy automatically exports that file and adds a second in the same directory with the name of the other data frame (e.g. comment) added.

Patowhiz:
I suggest we wait on this feature first until we discuss with the cClimsoft team on how comments will be imported in Climsoft. How Climsoft imports comments will strongly influence the dialog design and output.

Patowhiz:
@N-thony thanks for asking, Climsoft is yet to approve how comments will generally be stored. But in the new models that are being implemented for the next major release, any records in the database can have comments. So Climsoft will most likely expect a comment column as part of the data import. 

For Climsoft export, I'd lean towards having an implementation that has comments as a column of the exported data sheet instead of a separate sheet.

rdstern:
@Patowhiz do we still want to export comments via this dialog?  I am keen that they serve as a summary of possible changes needed in climsoft, even if they can not yet be part of the database.  But perhaps you would prefer that the comments sheet is exported via File > Export separately?  In that case should this be closed?

Patowhiz:
Considering version 4 supports csv only, this has to export comments separately. 

Then the user can decide where to file the comments.

Patowhiz:
@Vitalis95 kindly resolve the conflicts.

PR #8187, opened 2023-03-02
Correcting the issue on the Calculator when the result is not saved

### LINKED ISSUE #8176: Problem with the Prepare > Calculator when the result is not saved.
@N-thony I think this should be easy to correct.
a) Open any data set
b) Go into the calculator.
c) Do any simple calculation, e.g. 2+3+4+5
d) Untick the option to save the result.
e) Click ok
You now get the result in the output window, and also as a variable in the data frame - despite unticking that.
f) Return to the dialogue and run again - it is now just in the output window - as required.
g) The commands generated are very different.  I assume it is mistakenly still generating essentially the command to save in the data frame the first time.

In addition, I am now working where I don't show the R commands.  I suggest adding a checkbox, above the Save Result (which moves down a bit) saying Display Expression.  Default is unticked.  If ticked it displays the expression above the result.

If that is easy, could you try adding this too?  Otherwise just fix the bug for now.

N-thony:
Fixes #8176 
@rdstern I have fixed the issue reported in the corresponding. The addition feature quested there is a bit tricky to differentiate the type of script we would like to display or not, I did look at the code and will a have a quick discussion with @ChrisMarsh82 in few about. Meanwhile, you can check if the issue is fixed.

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -55,6 +55,7 @@ Public Class dlgCalculator
         ucrCalc.ucrSaveResultInto.ucrChkSave.Checked = True
         ucrCalc.chkShowParameters.Checked = False
         ucrCalc.ucrSaveResultInto.SetRCode(ucrBase.clsRsyntax.clsBaseCommandString)
+        ucrCalc.ucrChkShowExpression.chkCheck.Checked = True
         SaveResults()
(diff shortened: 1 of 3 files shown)


rdstern (approved):
Great

lloyddewit (approved):

PR #8204, opened 2023-03-08
Fixed a bug in Tidy Climatic Data dialog

### LINKED ISSUE #8201: New error in Tidy Climatic data in Version 0.7.9
Here is an example that works fine in Version 0.7.6 - shown:

image

It gives an error in Version 0.7.9.

I note also that this dialogue became unusable in Version 0.7.8 and @Vitalis95 corrected that error.  
Now it is possible to complete the dialogue, but not to get any results.  

As well as correcting the error for these data, I hope there can be an undestanding of what produced the problem, so we can also check that is has not causec problems for other data types or in other dialogues?

Vitalis95:
Fixes #8201 
@rdstern @lloyddewit , the bug is as result of missing  default value of  `element_name`. It was removed from the function and added as parameter,  I have restored it. Have a look it now.


N-thony (approved):

rdstern (approved):
@Vitalis95 This was a serious bug-find.  Many thanks.  It now works fine on the data we had for this workshop.  
@lloyddewit I hope this can be merged, and perhaps even an update released if it isn't too much work for Chris?  

It would also help on next week's workshop.

lloyddewit (approved):

lloyddewit:
@ChrisMarsh82 Please could you coordinate with @rdstern to see if it's worth creating a release for this bug fix? 
thanks

PR #8206, opened 2023-03-08
Fixed bug in Describe> Multivariate> Canonical Correlation Analysis dialog

anastasia-mbithe:
Fixes partially #8104 

This fixes the bug that was initially produced from the code.
Also, after a discussion with @Patowhiz during the Maseno sprint, we noticed that the options initially in the sub dialog were few and could be accommodated in the main dialog. And since the sub-dialog was not used in any other dialog, we settled on having the options in the main dialog. That has also been implemented in the branch.

@rdstern, @Patowhiz, and @N-thony this is ready for review.


rdstern (dismissed):
Now seems to work fine.

lloyddewit (changes requested):

anastasia-mbithe:
@lloyddewit, I made the changes

lloyddewit (approved):

Issue #8095, opened 2023-01-31
Improve the View Data dialogue for the new html system


rdstern:
@Patowhiz and @N-thony I hope this can be started within the mini-sprint?

Currently it has 2 options:
![image](https://user-images.githubusercontent.com/11226469/215706850-756cb69d-bd1d-4fc7-9071-d7a363183969.png)

This simply opens the R-viewer.  So it is the same as the Bottom-right-click > View Data Frame.  Here is is the default, and I am not sure that should continue to be the case?  But remember that a data frame can be very wide and very long.

The alternative is:

![image](https://user-images.githubusercontent.com/11226469/215708141-afb599f7-527b-4f02-912a-1776df96aa79.png)

Note the disabled html option here!  

In the redesign we could also question whether we keep those two top buttons as they are.  And do we want to include some nice options from the formattable package here.  

I would still like an option of the R viewer, but probably not as the default.  This will need some thought.

I suggest adding the `formattable` package for the next release will be a good start.

PR #8209, opened 2023-03-09
Added save option and HTML option to the View Data dialog

Vitalis95:
Fixes partially #8095 
@rdstern, @lloyddewit , @N-thony , I have added save control and enabled `html` option making it the default


rdstern (commented):
@Vitalis95 thank you for starting this one.  Great if you are happy to continue, but you could otherwise ask @derekagorhom if he would like to take it over?
a) The default is still the R viewer, for the whole data frame.  And it still comes behind the R-Instat windows.  Could you change so it comes in front.
b) The R-viewer is still the 3rd option on the View Columns.  So please change the label so it mentions the R-viewer.  It says extrenal and I thought that mightr be the new Maximised window.
c) I suggest perhaps the second option should still be the default.
d) Happy that you added the html.  But I am now trying with the diamonds data - 53,000 rows.  It has been going for over 30 minutes now, and shows no sign of any result, nor is there a message?  (It was all fine with the 36 observations from the survey data.)  But we can't allow that length of data into the output window I assume?  I suggest - like the pivot table stuff, it could go into a separate window?

Please also get Derek to look at the formattable package.  This might be useful at the same time.

N-thony:
> @Vitalis95 thank you for starting this one. Great if you are happy to continue, but you could otherwise ask @derekagorhom if he would like to take it over? a) The default is still the R viewer, for the whole data frame. And it still comes behind the R-Instat windows. Could you change so it comes in front. b) The R-viewer is still the 3rd option on the View Columns. So please change the label so it mentions the R-viewer. It says extrenal and I thought that mightr be the new Maximised window. c) I suggest perhaps the second option should still be the default. d) Happy that you added the html. But I am now trying with the diamonds data - 53,000 rows. It has been going for over 30 minutes now, and shows no sign of any result, nor is there a message? (It was all fine with the 36 observations from the survey data.) But we can't allow that length of data into the output window I assume? I suggest - like the pivot table stuff, it could go into a separate window? e) I have now restarted and tried the diamonds data again, but with the Describe > 2/3 Variables > Pivot Table. I added a row variable, so it would give me all the 53000 rows. It gives about 2000 in the output window, but then (quickly) gives them all in the maximise window. I wonder if your html should go into the maximise window in the same way, with perhaps a limit of 100, or perhaps a bit more in the output window?
> 
> Please also get Derek to look at the formattable package. This might be useful at the same time.

@Vitalis95 any progress?

N-thony:
@Vitalis95 I will have a look of item a), you can work on the remaining items.

Vitalis95:
@rdstern , have a look at this

rdstern:
@Vitalis95 for the first option, (the whole data frame) please add a label, below the data frame, that says 
Display in R viewer.
Then the viewer appears, but is then behind the R-Instat window.  
With the bottom right-click menu the viewer is in the front.  
Could you make it in the front for this option as well. 

Now the columns option:
a) Change the Display in Separate Viewer to `Display in R viewer`.
b) Add the formattable package to the next release.
c) Check with @Patowhiz what limits (if any) you may wish to add in terms of the number of columns and rows you display in the output window, when html?  

I tried with 3 datasets to get an idea of a possible answer.
a) the usual survey - all ok no limits needed.
b) The DAAG package called rockart.  This has 103 rows and 641 columns.  I thought it was quite presentable in the maximised window.  And somewhat useful there compared to the usual data frames, where you just see 50 columns at a time.  In the future we may improve further with neat formatting from the package.  For now it is ok.  So I propose no limit on the number of columns in the HTML output?  (We can check with loads of columns, when Patrick has finished his stuff on wide data.  )
c) The diamonds data with 53000 rows and 10 columns.  That did not mess up the output window, but did not appear.  It also did not appear in the maximised window.  So what should we do with long columns in html?
I tried filtering and it works fine with a filter.  And then I tried with 1500, then 5000 rows, then 20,000 rows.  It was a bit slow with the 20,000 but it works.  And looks neat in this table, so worth having.  So, I wonder about having 20,000 as a limit.  If you agree then perhaps add (Maximum 20,000 rows).  If the data is more than that, then either just do the first 20,000 - or say 10,000 if you think 20,000 is too much.  Or don't print it.

Later I would like to plan for interesting formatting options, making use of the formattable package properly.  But let's merge this option first.

Patowhiz:
@rdstern for now, I also propose no limits until we find issues with it. The control that displays html uses chromium browser engine, which is what the chrome browser uses. The chromium engine runs as a separate process (different thread) from that of R-Instat application. I expect the chromium engine to be well optimised. However, expect the control itself (not R-Instant) to 'hang' if it's a large dataset.

Thanks.

rdstern:
@Patowhiz just to check that I proposed no limit on the number of variables, but possibly a limit of 20,000 rows on the length?  It didn't seem to be over-happy with the diamonds data, which had 53000 rows.

In my suggestion of a limit I am not wanting anything clever.  If we have the limit, then if more than the limit is asked for it simply says "Sorry this option has a limit of 20,000 rows.  Perhaps you could apply a filter and retry."

But, if you don't want a limit, then it didn't give a bug.  It just didn't give the view in the output window.  So it could be ok like that.

@Vitalis95 can you add the package please.

rdstern:
@Vitalis95 thanks for the update and for the detailed way you have given that.  It is odd, that, on my machine the right-click remains in front while it is always behind from the dialog.

Yes please create another issue on further improvements in the View Data dialog.  The right-click is one point.  Could you include someting on Add a control to facilitate the use of options from the formattable package.  This to include control of decimals and heat maps, etc.  There should be a simple way to apply a format to all the columns, and also a control that permits different options for each variable.  The control to rename multiple columns can be a starting point.

rdstern (approved):
@Vitalis95 I am approving, as this is considerably better than the previous version.  It still does not display for the whole of the diamonds data - 53000 records, but it does no harm there.  In the new issue could you also add a possible check or restriction when more than 20,000 rows are being displayed.  

@lloyddewit over to you.

instat/dlgView.vb
```diff
@@ -266,4 +316,24 @@ Public Class dlgView
     Private Sub CoreControls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverView.ControlContentsChanged, ucrPnlDisplayWindow.ControlContentsChanged, ucrChkSpecifyRows.ControlContentsChanged, ucrNudNumberRows.ControlContentsChanged, ucrPnlDisplayFrom.ControlContentsChanged, ucrChkSortColumn.ControlContentsChanged, ucrReceiverSortCol.ControlContentsChanged, ucrPnlViewData.ControlContentsChanged
         TestOKEnabled()
     End Sub
+
+    Private Sub ucrSaveData_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSaveData.ControlValueChanged
+        'Dim strPrefix As String = clsOutputWindowFunction.GetRObjectToAssignTo 'If(ucrSaveData.GetText <> "", ucrSaveData.GetText, "last_table")
+        'clsGetObjectDataFunction.AddParameter("object_name", Chr(34) & strPrefix & Chr(34), iPosition:=1)
+        GetObjectName()
+    End Sub
+    Private Sub GetObjectName()
```
lloyddewit, line 325:
```suggestion

    Private Sub GetObjectName()
```

instat/dlgView.vb
```diff
@@ -266,4 +316,24 @@ Public Class dlgView
     Private Sub CoreControls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrReceiverView.ControlContentsChanged, ucrPnlDisplayWindow.ControlContentsChanged, ucrChkSpecifyRows.ControlContentsChanged, ucrNudNumberRows.ControlContentsChanged, ucrPnlDisplayFrom.ControlContentsChanged, ucrChkSortColumn.ControlContentsChanged, ucrReceiverSortCol.ControlContentsChanged, ucrPnlViewData.ControlContentsChanged
         TestOKEnabled()
     End Sub
+
+    Private Sub ucrSaveData_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSaveData.ControlValueChanged
+        'Dim strPrefix As String = clsOutputWindowFunction.GetRObjectToAssignTo 'If(ucrSaveData.GetText <> "", ucrSaveData.GetText, "last_table")
+        'clsGetObjectDataFunction.AddParameter("object_name", Chr(34) & strPrefix & Chr(34), iPosition:=1)
```
lloyddewit, line 322: delete?
```suggestion
```

lloyddewit (changes requested):
@Vitalis95 Thanks, looks good, I only had some trivial suggestions

rdstern (approved):
@Vitalis95 I assume the formattable package will be installed in the next update?
There will also be a new issue, that includes adding a Format button, in the Html row.  This will present a sub-dialog that gives options in the presentation.

lloyddewit (approved):

PR #8027, opened 2022-12-20
Added options in New Themes subdialog for frequency and summary tables.

anastasia-mbithe:
Fixes #7993

I have added controls, parameters, and functions (set R code) needed for the new Themes sub dialog. However, the subject requires more discussion with @lilyclements and @N-thony.

PR #8211, opened 2023-03-09
Fixed factor level order for `summary-variable` type in Table Summaries

lilyclements:
@anastasia-mbithe @rdstern this should fix the issue you highlighted in #8027 about reordering the summaries if `treat_column_as_factor = FALSE`.

This is ready to review, thank you again for being patient!

### DIFF
--- a/instat/static/InstatObject/R/Backend_Components/summary_functions.R
+++ b/instat/static/InstatObject/R/Backend_Components/summary_functions.R
@@ -1356,7 +1356,6 @@ SEDI <- function(x, y, frcst.type, obs.type, ...){
 
 ##TODO:Check if there are summaries that only apply to (Probabilistic-binary) types.
 
-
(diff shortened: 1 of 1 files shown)


rdstern (commented):
@lilyclements that's great - it seems to work very well.
But - I think this is new - when I return to the dialogue it has lost either the factors, or the variables in the dialogue.  I think it remembers the one that has the focus, but clears the other one?

lilyclements:
> @lilyclements that's great - it seems to work very well. But - I think this is new - when I return to the dialogue it has lost either the factors, or the variables in the dialogue. I think it remembers the one that has the focus, but clears the other one?

@rdstern interesting - I haven't changed anything in the dialog code, I have updated this from the master now, is it still happening here?
Is it happening in the master branch?

lloyddewit (approved):

PR #8222, opened 2023-03-20
Ensure that Filter dialog always reopens in default state

### LINKED ISSUE #8220: Small change in the Filter dialogue
@N-thony this should be easy for anyone to do.

The fact the dialogues remember the last setting is a major selling point in R-Instat.

There are just a few occasions where this should not happen.  One of these is the Filter Dialogue when you choose to copy to a new data frame, instead of applying a filter.  You almost never want to do that the next time round.  And it is easy to forget, and hence you do the wrong thing.  So please could that control always set back to the Filter default.

anastasia-mbithe:
Fixes #8220
 

rdstern (approved):
Looks fine.  Thanks.

derekagorhom (approved):

lloyddewit (approved):

Issue #8213, opened 2023-03-11
Simplify the climatic menu further


rdstern:
The problem, which is a good one, is that with beginners every click or change in a dialogue has to be explained and that always takes time.

In the climatic menu danny parsons arranged that, when a dataset is recognised as climatic, then much of a dialogue is often completed for you.  But there are a few instances this isn't happening.  With beginners this takes time.  The June e-siac will have many who are very inexperienced with computers so the simpler the better.  I give a few examples now that I remember and will add more as I review the dialogues again.  Others are welcome too.

I expect pull requests to each partially relate to this issue.  A pull request should usually relate to a single dialogue.

a) In Danny's impressive `PICSA rainfall and temperature` dialogues, the year variable is not made into the x.  I have never used it with anything else so suggest it be completed automatically.  
a again) The station is completed as the facet, but the default facet is set to None.  Set it to wrap.  

b) In the` Climatic > Tidy > Tidy Daily Data` dialogue I usually use the 3rd button (at the top), occasionally the second, and rarely the first.  I suggest we swap the first and third buttons.

c) Similarly -  In the `Make Date` dialogue I usually have Year, Month, Day.  So, the third option.  Occasionally the first - with a single variable - though that is - itself - more complicated.  Rarely the second.  I suggest, again we swap the 3 and 1 over.  Then, in the 3 variable case it goes nicely from year to month, but then does not often advance to the days.  Could you look at the code?

c again) And, in the same dialogue, we must have some neat code to recognise possible variable names - that we use in the define climatic data dialogue.  Could we recognise year and yr and month and mon and perhaps even mois, etc and day and jour, and fill them them automatically?  Also recognise year and doy and day in the 2 variable case.

c again, again) Could we add a Try button control to the dialogue.  Then we can check if it gives a date, sort of variable or NA, because we have completed incorrectly.  When adding this extra control, check on the spacing of the controls above, so the dialogue doesn't get too long.

d) In the `Fill Missing dates` dialogue the name column isn't filled by default.  If it is Station, or id, etc, could those be included by default - and perhaps check for more names.  There can be more than 1 here, so if there is both the names station and id, then include both.

e) In the `Define climatic data` dialogue our elements were x2 (tmax) x3 (tmin) and x5 (rain).  They were not recognised - could they be.  (There may be more element codes later, but these will do for now.)

e again) Please check the tab orders in this dialogue, particularly for the elements.  They currently go "asll over the place".

[...]

PR #8226, opened 2023-03-21
Ensured that year fills automatically in PICSA dialogs

Vitalis95:
Fixes partially #8213 
Fixes part a) of the issue
@rdstern , @lloyddewit , Made year variable be completed automatically and  wrap as default facet


rdstern (approved):
looks good.  I also checked temperatures and general, and the fix is there too.  Thanks

lloyddewit (approved):

PR #8228, opened 2023-03-22
Ensured that number of indices selected is in the middle of the message in Climdex Indices dialog

### LINKED ISSUE #8225: "ok" button in climndex is disabled
**Describe the bug**

**To Reproduce**
1. Go to 'climatic'
2. Click on 'prepare'
3. Scroll down to 'climndex'

**Expected behaviour**
"ok" button should be enabled.
IMG_20230320_222829
IMG_20230320_222829

**Details (optional):**
 - OS: [e.g. Windows 10]
 - R-Instat Version [e.g. 0.7.9]

derekagorhom:
fixes #8225 
I changed the text that was displaced - so that the number of indices selected would be in the middle of the message.
This PR is ready for review 
@rdstern , @N-thony  can you test/approve

### DIFF
--- a/instat/dlgClimdexIndices.Designer.vb
+++ b/instat/dlgClimdexIndices.Designer.vb
@@ -47,10 +47,10 @@ Partial Class dlgClimdexIndices
         Me.lblYear = New System.Windows.Forms.Label()
         Me.rdoAnnual = New System.Windows.Forms.RadioButton()
         Me.rdoMonthly = New System.Windows.Forms.RadioButton()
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@derekagorhom this seems to have solved the problem.  I would be happy to see it merged, if 
@lloyddewit is ok.
However, I note that this phrase (in the released version) seems to translate into Portuguese.  The new version does not.  It is now in English.  I suspect this may happen by default, but perhaps should not here, as there is  no reason why the words should be changed?

N-thony (approved):

PR #8233, opened 2023-03-23
Reset the Data View scroll bar to the top

N-thony:
@rdstern following our chat on Skype few minutes ago, I made the changes as you have noticed during the workshop when paging sometimes it is giving you blank page. This will always set the scroll bar at the top. You can test it also try with the filter for the rows.
I'm checking about the column if easy will just do it here.


rdstern (approved):
seems fine

lloyddewit (approved):

PR #8240, opened 2023-03-27
Ensured that 'Colour Columns by Structure' check box works correctly in Column Structure dialog

### LINKED ISSUE #7232: Complete the Experiments > Define > On-Station dialogue

derekagorhom:
fixes partially #7232 
I made the checkbox for the colour the default. So it starts with the checkbox ticked and also fixed the problem of the dialogue remembering the variables in each control when the dialogue is re-opened. 
@rdstern  this PR is ready for review. 
Thanks

### DIFF
--- a/instat/dlgColumnStructures.vb
+++ b/instat/dlgColumnStructures.vb
@@ -59,6 +59,9 @@ Public Class dlgColumnStructure
         ucrChkColourColumnsByStructure.SetText("Colour Columns by Structure")
         ucrChkColourColumnsByStructure.AddFunctionNamesCondition(True, frmMain.clsRLink.strInstatDataObject & "$set_column_colours_by_metadata")
         ucrChkColourColumnsByStructure.AddFunctionNamesCondition(False, frmMain.clsRLink.strInstatDataObject & "$remove_column_colours")
+        ucrChkColourColumnsByStructure.SetParameter(New RParameter("checked"))
+        ucrChkColourColumnsByStructure.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
+        ucrChkColourColumnsByStructure.SetRDefault("TRUE")
     End Sub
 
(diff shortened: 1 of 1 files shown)


rdstern (approved):
Fine - I have approved.  Sometime later I would like to add options to choose which colours, but that is very low now on my priority list.  So I have just changed the pull request into fixes partially.  And approved.  Over to @lloyddewit

lloyddewit (approved):

lloyddewit (approved):

PR #8243, opened 2023-03-28
Ensured that 'tmin' & 'rain' fill automatically in Climatic Boxplot dialog

Vitalis95:
Fixes partially #8213 
Fixes part h) of the issue
@rdstern , @lloyddewit , Made `tmin `variable be completed automatically and if `tmin` is missing, then `rain `be filled automatically

### DIFF
--- a/instat/dlgClimaticBoxPlot.vb
+++ b/instat/dlgClimaticBoxPlot.vb
@@ -420,13 +420,17 @@ Public Class dlgClimaticBoxPlot
         Dim strMonthCol As String
         Dim strDataFrame As String
         Dim strRainCol As String
+        Dim strTempCol As String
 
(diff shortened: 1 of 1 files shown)


lloyddewit (changes requested):
@Vitalis95 thanks, looks good, just one small suggestion

rdstern (approved):
Looks fine.  Thanks

lloyddewit (approved):

rdstern:
@Vitalis95 There is a serious problem here, that I hadn't noticed before:

image

It currently does just "what you say on the tin", so now includes tmin by default when loaded.

But the problem is that it remains most insistent that tmin be analysed.  So, if you change it to tmax, then it immediately changes the variable back to tmin!  That's not what is needed of course.  You should be able to analyse any element.

Issue #8151, opened 2023-02-18
Changing the File > New Data Frame and the Edit > Script dialogue to streamline running scripts in R-Instat


rdstern:
I build on comments introduced elsewhere.
We currently can prepare and execute scripts from 3 places:
a) `File > New Data Frame `  has a commands option.  It executes the commands to build a data frame.  The code provides the data frame and an extra line is then added to include the resulting data frame in the new, or existing data book.
b) The suggestion is to delete this option from the File > New Data Frame dialogue.  This has the plus point of deleting a complicated option, that is perhaps a bit daunting to our main audience for R-Instat as it the first dialogue that some may use - and they may be beginners.
c) @lloyddewit I note, however, one feature we will lose is the `Try control`.  I like that feature and it is also on the other "halfway" dialogues.  Could it, perhaps become a button directly available from a script window.  It opens effectively a single sub-dialogue, with just this one control.  And then a Return, etc, because you return to the script window.  The only difference from the current Try control is that it opens the sub-dialogue.  It is already there on our existing halfway dialogues.  Then it tries the marked code?
d) I also assume we will enhance and keep the `Edit > Script dialogue`.   It is designed to help users to construct a typical script if they start "from scratch".  It is an alternative to the To Script button on each dialogue.  
e) I don't have any sentimental attachment to this dialogue - just suggest it could remain pretty useful.  It is designed particularly to make it easy to include the "special" R-Instat components to get a data frame and also to write results back into a data frame or to the results window.

![image](https://user-images.githubusercontent.com/11226469/219849065-eddfdbad-218e-4044-b672-c6bb1d07de14.png)

f) Assuming we will keep it, then one simple change - even in the February release - could be to disable the Ok button.  So, when you are happy with the code you then do the ToScript and that sends it to the current script window. 

g) The New Data Frame could add a first option to the Data Frame shown above.  It would be O New Data Frame and the existing one becomes O Get Data Frame.  

h) I quite like the one-line examples that are in the current File > New dialogue and they might be included.  Bigger ones could perhaps be loaded from a file, or we might only allow the load into the ordinary script windows.  Aha this is becoming a bit more complicated!

i) If we keep the dialogue, in some form, then we should also check if the code side can include all your enhancements?  We could also check on alternatives.  I assume one of your new script windows will be read only and will include the log file?  Might another script window (or all of them) permit the sort of assistance that this dialogue is providing? 

[...]

PR #8255, opened 2023-04-03
Renamed first radio button in New Data Frame dialog

derekagorhom:
fixes partly #8151 
I changed the radio button for Empty to New, Then made Lists the second, Then Random, Construct, Commands - in that order and together. 

@rdstern , @lloyddewit this PR is ready for review.


rdstern (approved):
@derekagorhom that looks nice.  If @lloyddewit is ok then it can be merged

lloyddewit (approved):

PR #8257, opened 2023-04-04
Simplified the use of comments when a data frame does not have key variables in Add Comment dialog

### LINKED ISSUE #8216: Simplify the use of comments when a data frame does not have key variables
I will also confirm with @dannyparsons, but suggest this should be an easy addition to the code.

I need climatic beginners to be able to add commentson aspects of interst in their data.  This is easy if the data frame has a key field.  Otherwise it tells you to get one!  That's not something for beginners.

So, when a first comment is added and there is no key, then I would like a default key to be made.  That should be easy, because it is done by the Prepare > Data Frame > Row Numbers/Names dialogue.  When the key option is selected, see below:

image

So, when there is no key for a data frame - just make any data frame (with File > New Data Frame and there won't be.  Then (invisibly) run the code that comes from running this dialogue.  That will make a new row variable in their data frame, so include the check that the name doesn't exist and make it row1, etc (as is done in the dialogue) if it does.  In the output window please also include the comment from the row dialogue, so you are effectively running that dialogue.

@N-thony I assume this is one for you to allocate.  Thanks

Vitalis95:
Fixes partially #8216 
@rdstern @lloyddewit ,added a control  `Make a column a key for the data frame` to simplify the use of comments when a data frame does not have key variables.


rdstern (changes requested):
@Vitalis95 well done - it is fine, but working too well!

Change the label to `Make Row a Key for the Data Frame`
Don't have the saving option control.  It is just (always) called Row.  

Now the important bit.  It is only visible if there is no existing key.  I tried with a climatic dataset - which has Station and date as the key.  If there is a key, then we don't want this option.

And, of course, when you add a second comment to any data frame, then there must already be a key, so you don't have that option.


I hope that's clear?

Vitalis95:
@rdstern , have a look at it. Thanks

rdstern (dismissed):
Looks good now.  Can merge if @lloyddewit agrees

lloyddewit:
@rdstern In issue #8216 you wrote:

> I am changing 8257 to only Partially Fix this issue

Please note that if you write 'partially fixes #8216' then GitHub just sees the 'fixes #8216' and thinks that the whole of the issue is fixed. We need to write 'fixes partially' instead. I fixed it above.

image

instat/dlgAddComment.vb
```diff
@@ -135,12 +179,53 @@ Public Class dlgAddComment
         strSelectedColumn = strColumn
         bUseSelectedPosition = True
     End Sub
+    Private Sub AddRemoveKeyFromAfterCodes()
```
lloyddewit, line 182:
```suggestion

    Private Sub AddRemoveKeyFromAfterCodes()
```

instat/dlgAddComment.vb
```diff
@@ -135,12 +179,53 @@ Public Class dlgAddComment
...
+    Private Sub IdentifyKey()
+        If frmMain.clsRLink.IsVariablesMetadata(ucrSelectorAddComment.ucrAvailableDataFrames.cboAvailableDataFrames.Text, "Is_Key") Then
+            ucrChkMakeColumnIntoKey.Enabled = False
+            ucrChkMakeColumnIntoKey.Checked = False
+        Else
+            ucrChkMakeColumnIntoKey.Enabled = True
+            ucrChkMakeColumnIntoKey.Checked = True
+        End If
+    End Sub
+    Private Sub ucrChkMakeColumnIntoKey_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkMakeColumnIntoKey.ControlValueChanged
```
lloyddewit, line 208:
```suggestion

    Private Sub ucrChkMakeColumnIntoKey_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkMakeColumnIntoKey.ControlValueChanged
```

lloyddewit (changes requested):

lloyddewit (approved):

PR #8264, opened 2023-04-06
Implementing Edit Dialogue

### LINKED ISSUE #8239: Improve the column right-click menu and add an Edit Cell(s) option
@lloyddewit and @N-thony I'd like your confirmation that this is a sensible idea?

image

This is a data frame with all the months where the whole month was missing.  I am quite pleased with how easily it was able to get this.  I would now like staff to investigate the situation, by station and month by month to see if they can locate the missing data.  So I'd like a comment on what they find each station/year/month in the last column.

Now you can currently type into each cell in turn already, but I am wondering if it could be useful to also have an Edit Cell(s) option, below the Delete Cell(s) option.  This would sort of be similar to Excel, where you have the entry are the top of a sheetm where you can type and that goes automatically into the current cell.

This would usually be for a single cell, but ut could, like the delete cell(s) also put the same text into a block of cells.

It would open s imple dialogue with the cell selected and the current contents - for a single cell.  For a block of cells it would just give the column and rows.  Then (like the comment dialogue) you can type away.

By the way, when editing this right-click menu could we take the opportunity to delete the Paste option.  Paste seems never to be finisherd, and (even when it is) I don't want it there.  I don't want to particularly encourage pasting and the option is fine in the Edit menu.

derekagorhom:
fixes partly #8239 
This PR is not ready for review.
@lloyddewit as we discussed on zoom i have made the draft pull request 
the only problem with the dialogue not working is that it can not populate the columns and rows part of the rcode.


lloyddewit:
@derekagorhom Thanks, I'll try and look at this next week, and see if I can find the problem.

rdstern:
@derekagorhom I get build errors on this.
I note your message above where you say: "the only problem with the dialogue not working is that it can not populate the columns and rows part of the rcode."

I just note that this is a "partner" to the delete option - and that is doing well and populating the columns and rows with NA.  In your case, if there are multiple rows (and possibly columns" you want to populate them all with the contents.  Of course, one difference is that NA is allowed everywhere, while your result may not be permitted.  I am hoping that will already be catered for, automatically, when you type a problem text.

lloyddewit:
@derekagorhom In April, I offered to help with this PR but didn't do so yet - sorry for that.
I see that you are updating the PR again. Would you still like my help or are you OK for now?
thanks

derekagorhom:
@rdstern and @lloyddewit this PR is ready for review/Testing.
As at now the dialogue can only edit for characters, I still have to make some few minor changes.

derekagorhom:
@lloyddewit please are you on mattermost i would like to have a discussion on the dialogue

lloyddewit:
> @lloyddewit please are you on mattermost i would like to have a discussion on the dialogue

@derekagorhom Thank you for the quick chat. Great that you found a solution without needing my help. I understand that there is still one problem that needs to be fixed (the user need to press reset to update the selected cell). I suggest that @rdstern and I do the the test/review after this is resolved. Please let us know when this is done.
Thanks

derekagorhom:
> > @lloyddewit please are you on mattermost i would like to have a discussion on the dialogue
> 
> @derekagorhom Thank you for the quick chat. Great that you found a solution without needing my help. I understand that there is still one problem that needs to be fixed (the user need to press reset to update the selected cell). I suggest that @rdstern and I do the the test/review after this is resolved. Please let us know when this is done. Thanks

Sounds good to me, thanks

rdstern:
@derekagorhom this is very promising.  I thought I'd check early to see where you have got to.
a) The menu should say Edit Cell(s) to parallel the Delete Cell(s) just above.  However, there are more changes needed to the right-click menu so that can wait.
b) It seems to check the entry against a valid variable name.  That's not needed and is at least one reason I can't enter a number.
c) It seems one behind in what it remembers.  So when I moved from the Variety column ( a factor) to the Fert column, it was still checking against the allowed values in the variety column.  OK after reset - perhaps that's what you mentioned earlier.
d) When I marked multiple rows it only edited the first of them.
e) Like the Rename Column dialogue (also on right click) it could have a field with Current Value.  Then you edit the field you have, called New Value: which initially has the same contents as the Current value.

derekagorhom:
@rdstern when making the new changes we discussed yesterday. I realized that the reogrid recognizes rows according to their position i.e row ="1", so this makes it difficult to pick the value from the cell since it only pick its position rather (this is the same for the delete cells too)

rdstern:
@derekagorhom I assume that's because you are avoiding using R.  I assume if you use R it can easily give you the value and will also be able to inform you of the type of variable, 

My R isn't good enough, but when I use the survey data from the library and do:

as.numeric(survey[4,6])

I get the value 33.6, which is the yield in the 4th row of the yield column.


You could start by asking @fran2or because my R isn't good enough to take you further.  If that doesn't work, then we'll try Lily.

derekagorhom:
@N-thony I have made the changes and now the elements in the cell can be viewed in the dialogue and edited

instat/dlgEdit.vb
```diff
@@ -0,0 +1,118 @@
...
+        'set validation of ucrInputNewName as an RVariable.(input should not have any R reserved words like 'if','while')
+        ucrNewName.SetValidationTypeAsList()
+    End Sub
+
+    Private Sub SetDefaults()
+        clsReplaceValue = New RFunction
+        ucrSelectValues.Reset()
+
+        'trim white space from ends of value
+        strNewValue = strNewValue.Trim()
```
N-thony, line 71: remove? it seems to not be needed

instat/dlgEdit.vb
```diff
@@ -0,0 +1,118 @@
...
+        clsReplaceValue.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$replace_value_in_data")
+        clsReplaceValue.AddParameter("data_name", Chr(34) & ucrSelectValues.strCurrDataFrame & Chr(34))
+        clsReplaceValue.AddParameter("col_name", Chr(34) & strSelectedColumn & Chr(34))
+        clsReplaceValue.AddParameter("rows", Chr(34) & StrRowIndex & Chr(34))
```
N-thony, line 75: can you add the positions?

instat/dlgEdit.vb
```diff
@@ -0,0 +1,118 @@
...
+
+        ucrBase.clsRsyntax.SetBaseRFunction(clsReplaceValue)
+
```
N-thony, line 78: remove?

instat/dlgEdit.vb
```diff
@@ -0,0 +1,118 @@
...
+    End Sub
+
+    Private Sub SetRCodeForControls(bReset As Boolean)
+
```
N-thony, line 82: remove?

instat/dlgEdit.vb
```diff
@@ -0,0 +1,118 @@
...
+        TestOKEnabled()
+    End Sub
+
+    Public Sub SetCurrentColumn(strColumn As String, strRowNumber As String, strIRow As String)
+        strSelectedColumn = strColumn
+        strRowText = strRowNumber
+        StrRowIndex = strIRow
+        bUseSelectedColumn = True
+    End Sub
+
```
N-thony, line 117: remove?

N-thony (changes requested):

rdstern (changes requested):
@derekagorhom this is looking promising.
a) It worked for a Factor variable, but I got this odd error when changing a numeric value.

image

b) Please could the old Variable and the new Variable controls be the same width.

c) Could the new variable start with the same contents as the okd Variable ( as it would in Excel or Open Office)

d) Now a bigger challenge!  If the variable is a factor, then could there be a pull-down with the levels of the factor in the New Variable.  One in which you could perhaps type, and with the default being the current value.

I hope you could adapt the code from an existing dialogue.  Therre are probably more, but the Climatic > Tidy > Enter Data dialogue > Select station control.

image

e) Can you add the code from the main data selector which gives the column type - see below:

image

derekagorhom:
I have fixed items B -E . 
I am still working on fixing the issue of a)

rdstern (commented):
@derekagorhom that's good progress.

But, when a factor is selected we don't need an extra control for the levels.  It is the New Value control that becomes a pull-down..  Then it is easy to change the value there, without typing.

rdstern:
@derekagorhom I am not sure I meant to have the usual data selector and so have an "ordinary" dialogue here? 
Here is your current dialogue:

image

You will notice I changed the Vareiable, because ytour dialogue permits this!  But you did not change the cell contents.  If you wish to continue with the structure, then please change Column Name to Column:

Then also change your current `Select Factor :` to `Row:`

And I thought this would be very difficult, but I have just realised that this is now very similar the the idea and the code in the Add Comment dialogue!  This gas just been enhanced by @Vitalis95 so you could share with him if you wish.

In particular the Add Comment needs the data frame to have Key fields.  And Vitalis ahs just added the code for this in (I think) a pretty clever way.  So, you can add the same control here.  When there are key fields the row is shown in the Comments data frame, and will want it displayed here - so don't display anything if there are no key fields, but next time (when there are) you can display it.

This is looking very exciting now!

rdstern (commented):
@derekagorhom this now all seems to work.  Well done.

But it looks very odd with the extra control when there is a factor.

image
 So you are changing the value in the drop down, but the old value is still visible in the disabled control called New Value.

When you change a numeric variable that odd extra control is still visible, but disabled.

I'd really like just the one control for the New Value.  Which is as you have it for numeric and the drop down for a factor.

My ideal would be a control for the factor that you could alternatively type into.  But I am happy with the list if that is tricky.

One other small change.  Could you make the Column and Row controls back the the same "ordinary" size for them  So they remain left-aligned as now, but are not so wide.  I like the width of the old and new controls as they are.

(Finally, as you are getting so much better as a programmer now, we will merge this one once you have made these small changes.  But can you think of anything else I might be asking for later?  Still just in editing a single cell.)

rdstern:
@derekagorhom can you check please - I get build errors

lloyddewit:
Replaced by #8354

PR #8267, opened 2023-04-16
Merged CrowdIn changes from PR #8098 into translations table in database

lloyddewit:
This PR contains a newly generated translation database.
I only did a machine translation for Russian. This should significantly improve the amount of R-Instat text translated into Russian.
Updated text for all the other languages is uploaded to CrowdIn and ready for translation when required. This means that the translations for French and other languages are not improved by this PR (but can now be improved easily in CrowdIn).


lloyddewit:
@rdstern I will proceed and merge this PR without your testing. This is so that @ChrisMarsh82 can make a new release for us in the course tomorrow.
This is higher risk than I would like. I suggest we treat this version as a beta.

Issue #8242, opened 2023-03-27
Add labels to boxplot dialogues 


rdstern:
This is the stack-overflow discussion on [adding labels](https://stackoverflow.com/questions/33524669/labeling-outliers-of-boxplots-in-r).

There are 6 suggestions and I wonder if the solution involving stat_summary is one we could use?

It would apply to both the general and the climatic boxplot dialogue.  I assume we would (I assume) also add the link to geom_text into the Plot Options.

@Wezziebanda  will provide details including providing the R-code quickly and then maybe @MeSophie could attack?

rdstern:
On layout, I am keen that the dialogue does not need to become longer.  So I suggest 
a) Move the Horizontal Plot option to the right,instead of the (disabled) swap parameters option.  Put it there instead - as we have never done the work to enable that option. 
b) Add another option, called Label Outliers below the Add Points option.  Details will be given by @wezziebanda.
c) I assume it will only use geom_text?  If so, also add geom_text into the Plot Options pull down.
d) The R-code and an example will be given below.

WezzieBanda:
@rdstern I used the survey data from R-instat library

da11<-read.csv ("C:/Users/grace/Downloads/survey.csv")



library(ggplot2)

ggplot(da11, aes(x = factor(village), y = yield, fill = factor(village))) +
  geom_boxplot() +
  stat_summary(
    aes(label = round(stat(y), 1)),
    geom = "text",
    fun.y = function(y) { o <- boxplot.stats(y)$out; if(length(o) == 0) NA else o },
![Labels](https://user-images.githubusercontent.com/106272881/230734345-d1730397-70b2-4370-9c75-adb29bbdc332.PNG)

    hjust = -1
  )

rdstern:
@WezzieBanda that's great.
Now can you do the same boxplot (without the labels of course) in R-Instat.  Then press To Script to gwet the script, and add the stat_summary in the script file to get the same result.  

Could you also try without the script, but using the Describe > General dialogue which allows you to build up a graph layer-by-layer.  Could this be used to start with the boxplot and then add the summary layer?

(This sort of stuff would all have been soooo much easier to discuss when I had the time and we were together in Malawi!)

WezzieBanda:
@rdstern I tried using the first option and it I got the same results. 
![outlier1](https://user-images.githubusercontent.com/106272881/231007513-94237432-8d36-49fc-ac5f-95c541ca285e.PNG)

I am now trying out the second option.

PR #8271, opened 2023-04-18
Added Label Outlier Check Box and Minor Design changes on Boxplot Dialog.

MeSophie:
Fixes partially #8242.
@rdstern I fixes part a) and b).

@lloyddewit @WezzieBanda please can you have a look?


rdstern (changes requested):
@MeSophie good to get started.  
a) But sorry I misled you - it is not the `Plot Options` to move to the right, instead it is the `Horizontal Plot` option.

b) And I notice you already added the Geom_text option into that pull-down.  That's great, but could you just call it Text Options.

c) And move the Tufte Option up one control.  Later I am sure you will have more controls to add to the new row with the Label options.

And hopefully get help from Wezzie on the code and options to use for the labels

WezzieBanda:
@MeSophie The dialogue seems not to be working for me, I believe it has to do with how the sta_summary_fun was incorporated in the code.
Boxplot label

WezzieBanda:
@MeSophie looks good.

instat/dlgBoxPlot.vb
```diff
@@ -465,6 +514,18 @@ Public Class dlgBoxplot
         toolStripMenuItemTufteOptions.Enabled = (rdoBoxplotTufte.Checked AndAlso ucrChkTufte.Checked)
     End Sub
 
+    Private Sub ucrPnlPlots_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlPlots.ControlValueChanged, ucrChkTufte.ControlContentsChanged, ucrChkAddPoints.ControlValueChanged
```
lloyddewit, line 517: this sub doesn't seem to do anything, remove it?

lloyddewit (changes requested):
@MeSophie Looks good to me, just one small question

rdstern (changes requested):
@WezzieBanda and @MeSophie this is looking very promising.  It seems to work!  Well done.

But there are a couple of challenges still:

a) When I change the coefficient from 1.5 the boxplot changes, but the labelled outliers are still for the default value of 1.5.  Can the code for the outliers reflect the coefficient used?

image



b) I also tried an example where there are no outliers and it then gives an error.  An example is our usual survey data with yield against Variety.  (It works fine with yield against Village, where there is just one outlier.)
c) And a - i hope trivial design change.  Could you swap the Add Points and the Label outliers on the Left of the dialogue.  Then you could move the Tufte option up one place on the right.  (It looks lonely down there now!)  And could you change Tufte Boxplots to Tufte Boxplot (so singular), to be consistent with the option at the top. 
d) And another one.  The label outliers is only for the Boxplot option.  It does not apply to the jitter or the violin plots.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -2902,6 +2902,9 @@ get_vignette <- function (package = NULL, lib.loc = NULL, all = TRUE)
   else return(sprintf("file://%s", file))
 }
 
-
+# Function used in <<< DLG NAME >>> as `fun` parameter value
+stat_summary_fun <- function(y) {
+  o <- boxplot.stats(y)$out; if(length(o) == 0) NA else o
```
N-thony, line 2907: can you rename this variable correctly, something like
out <- boxplot.stats(y)$out; if(length(out) == 0) NA else out

MeSophie:
@rdstern I'm done with the parts a), c) and d). I added one control  for `Size` for part a) for now is the only way that I see to play with the size of outlier label. I'm still investigate of part b).

rdstern:
@MeSophie this is promising.  I think you can make some further progress, but then, for efficiency you might then wait for some help on the R-code from @lilyclements.  This may not be till the week of 12 June, when she will be visiting Kenya (though you are still likely to be working from home then.).
a) The outliers are still given for the coefficient of 1.5, even when you change the coeff.
b) As mentioned above, there is an error if there are no outliers (saurvey data with yield v variety is an example.
c) I am afraid a new one - the outlier is only the y variable.  The code gives examples where you may wish to label another variable, e.g. the date.  This will change the dialogue a bit to add space.
d) When it is working it would be good to add the same to the Climatic > Check Datat Boxplot dialogue - so consider this as the coding is done, i.e. that you will be using it in 2 dialogues.

On the boxplot design - point c, you may wish to move the Horizontal Plot control back to the Left.  Then you can move Tufte up and add a checkbox with label Variable: (default unchecked.  If checked, then it opens a single receiver that can take a variable to be plotted.

N-thony:
> @MeSophie this is promising. I think you can make some further progress, but then, for efficiency you might then wait for some help on the R-code from @lilyclements. This may not be till the week of 12 June, when she will be visiting Kenya (though you are still likely to be working from home then.). a) The outliers are still given for the coefficient of 1.5, even when you change the coeff. b) As mentioned above, there is an error if there are no outliers (saurvey data with yield v variety is an example. c) I am afraid a new one - the outlier is only the y variable. The code gives examples where you may wish to label another variable, e.g. the date. This will change the dialogue a bit to add space. d) When it is working it would be good to add the same to the Climatic > Check Datat Boxplot dialogue - so consider this as the coding is done, i.e. that you will be using it in 2 dialogues.
> 
> On the boxplot design - point c, you may wish to move the Horizontal Plot control back to the Left. Then you can move Tufte up and add a checkbox with label Variable: (default unchecked. If checked, then it opens a single receiver that can take a variable to be plotted.

@MeSophie any progress? Also, this PR has conflicts, can resolve them?

rdstern (commented):
@MeSophie this is not urgent compared to your other work.  But it would be nice if easy, given you have started.
a) When I change the coefficient for the boxplot, this is ignored and I still get the default labels from coeff 1.5.
b) I get an error if there are no outleirs.  For example survey yield against variety.  Against village it works fine, as there is one outlier.

MeSophie:
> @MeSophie this is not urgent compared to your other work. But it would be nice if easy, given you have started. a) When I change the coefficient for the boxplot, this is ignored and I still get the default labels from coeff 1.5. b) I get an error if there are no outleirs. For example survey yield against variety. Against village it works fine, as there is one outlier.

@rdstern I'm on it for part b) if we remove the `outlier.colour="red"` parameter as a defaut value of boxplot function ( `(ggplot2::geom_boxplot(outlier.colour="red"))`) so that the user can take this parameter from sub dialog it will solve the bug.
For part a) i'm still investigating on it if @lilyclements can help here it will be great.

MeSophie:
@rdstern I have fixed part b). For part a) I definitively need some help from @lilyclements because in R I have tried to edit the coef  in referents codes giving on [adding label](https://stackoverflow.com/questions/33524669/labeling-outliers-of-boxplots-in-r). But I still having the same problem.

MeSophie:
This is replace by PR #9804

PR #8273, opened 2023-04-19
Minor Changes on Find Non_Numeric Values and Replace Values Dialog Design.

### LINKED ISSUE #8272: More language improvements to the dialogue layouts
Early in our language work we used the French translations to ensure that labels in dialogues could accommodate all languages.  We assumed French would be the widest.

We have just been using the Russian version for the first time and this sometimes needs a considerably wider dialogue than we have allowed for.  Here is the Prepare > Check Data > Non-Numeric dialogue, (which is also in the climatic menu.)

image

And here is the Replace dialogue - also in check data.

image
 
Another "tour" of the dialogues may be useful?  I wonder if this is work that perhaps @rachelkg could do, or perhaps check, if someone else is able to do it?  If not Rachel to do, then perhaps @Vitalis95 could suggest who would do it - not him though.

MeSophie:
Fixes #8272 

@rdstern , @lloyddewit, @Vitalis95 Please can you have look. It is also good to know that these Dialogs are in  `Climatic>Tidy and Examine `. Thank you.


rdstern (approved):
@MeSophie many thanks for taking this on.  I am afraid there is mopre to come of the same type!

lloyddewit (approved):

PR #8205, opened 2023-03-08
Added Selected Items label to multiple receivers for dialogs in Prepare > Databook sub menu

derekagorhom:
fixes #8203 

I have added the Selected item label to `dlgdeletedataframe`, `dlgHideUnhidedataframe` and `dlgReorderdataframe`.
This PR is ready for review
@Patowhiz , @N-thony can you review this PR.

Thanks

PR #8286, opened 2023-04-24
Showing Seleted Items in Multiple receiver

### LINKED ISSUE #8203: Showing number of selected items in multiple receiver
I suggest the multiple receiver to always show the number of selected items. I think it will be very helpful in helping users be keen on the items selected especially for sensitive operations like deleting and reshaping of data frames.
For example have a look at the dialog below.

derekagorhom:
fixes #8203 
Replaces PR #8205 

I reverted the previous changes and enabled the multiple receiver to show selected items. 
This PR is ready for review. 

@Patowhiz , @rdstern  can you have a look.


Patowhiz:
image_2023_04_23T23_12_10_048Z

@derekagorhom thanks for this. Can the label be part of the receiver heading. For instance, Survey(2). This will avoid the extra label control.

rdstern (changes requested):
@derekagorhom - this is a very nice feature.  Note the request by Patrick.
Also that the code he requests is already implemented in the `Climatic > Tidy and Examine > Tidy daily data` dialogue.
So you could use that code?  (It was I think by Danny, so might be quite good?)  But also we don't need the feature twice in that dialogue.

image

derekagorhom:
@rdstern ,danny's code is similar to what was used in PR #8205

rdstern:
@derekagorhom great.  I look forward to seeing the next round then, for this change.  It is a very useful addition to the dialogues.

N-thony:
> @derekagorhom great. I look forward to seeing the next round then, for this change. It is a very useful addition to the dialogues.

@derekagorhom any progress?

N-thony:
> > @derekagorhom great. I look forward to seeing the next round then, for this change. It is a very useful addition to the dialogues.
> 
> @derekagorhom any progress?

@derekagorhom I'm ready to help.

N-thony:
@derekagorhom I'm ready for a call now about this.

derekagorhom:
@rdstern , @Patowhiz 
I was able to implement the item count for the multiple receiver, it works fine now
can you review this
Thanks

rdstern:
This is quite nice now.  But 2 problems, both of which I would like you to consider whether you could have found them yourself in the future:
a) You may remeber I described how I test new features and I always return multiple times to the dialog.  Here is the survey data after I came back four times.

image

The first time it correctly said (6) which was very nice.  The second time it said (6) (6) and then so on.

b) I immediately wondered what would happen if the data frame had a long name?  Here is an example:

image

What would you like to suggest here?  I wonder about just the first 8 characters, (if more than 8) with then the ... and then the number in brackets?  Maybe less than 8?

I also wanted to make sure there is enough space in case there are more than 1000 elements in the receiver - maybe even 10000, but I can't test that yet - R-Instat can't yet cope with very wide datasets.  I tried with rockart from the DAAG package and I think that might be ok?  

image

derekagorhom:
@N-thony  I looked into the first issue and it seems to be an individual dialogue problem rather than something is wrong with the code. i have a solution for that but for the second issue i might need you help on that

N-thony:
@derekagorhom be careful here, if you look at the base code, you will find sometimes we update the R code `data_name` parameter value from the multiple receiver header, so renaming it might make some dialogues to bug esp those have the selectors set parameter as `RFunction`

derekagorhom:
@N-thony are you available for a quick call?

derekagorhom:
> @derekagorhom thanks for changes. I suggest I quickly take over this so that I can solve it in PR #8465. Thereafter we can quickly discuss about my approach. What do you think?

@Patowhiz this is fine by me, Thanks

derekagorhom:
@Patowhiz had a look at the files being changed in PR #8465 and i realized it left out the corrections to PR #8203 so i think this PR can be allowed to make those corrections and your PR focuses on the multiple selector.

Patowhiz:
@derekagorhom @lloyddewit PR #8465 now has changes on `ucrMultipleReceiver.vb` and `dlgDeleteDataFrames.vb` files. I suggest this PR to wait until those changes are merged.

@derekagorhom kindly note some changes made to `ucrMultipleReceiver` relate to what you attempted to do in this PR. I have implemented the issue in a way that the dialogs that use multiple receivers don't need to have any changes. Have a look at subroutine `SetGroupHeaderVariablesCount()` in that PR.

Changes made to `dlgDeleteDataFrames` are meant to fix the a bug mentioned by @rdstern kind comment https://github.com/africanmathsinitiative/R-Instat/pull/8465#pullrequestreview-1589192452 . I have also removed the code you added for counting the data frames at this level. Kindly test and confirm in that PR. @rdstern thanks for noticing that bug. The bug related to how the dialog was being loaded, the sequence of commands were not correct leading to the regression.

PR #8289, opened 2023-04-26
Added a layer parameter for density ridges

### LINKED ISSUE #8283: Add a new option for the density ridges
This is a useful [article](https://cran.r-project.org/web/packages/ggridges/vignettes/introduction.html)

Here is an example of the code:

```
ggplot(iris, aes(x = Sepal.Length, y = Species)) + 
  geom_density_ridges(rel_min_height = 0.01)
```
Could you add this to the layer parameters for the geom_density_ridges.  It could be calle min_height.  The default is 0.01 and perhaps it could be an up-down from 0 to 0.1 in steps of 0.001, but perhaps start with that and I'll check.

At the same time, could you change the up-down for scale.  It could go from 0 to 5 in steps of 0.1.  Currewntly it goes in steps of 1.

That's a good start!  But the article also uses the other 2 geoms that we have and that give further options on the ridges - and I have been asked to produce a graph with colours related to temperature.  So later is would be good to be able to do this as well.

image

anastasia-mbithe:
Fixes #8283 

### DIFF
--- a/instat/ucrGeom.vb
+++ b/instat/ucrGeom.vb
@@ -874,7 +874,7 @@ Public Class ucrGeom
         clsgeom_density_ridges.AddLayerParameter("point_shape", "editablelist", "|", lstParameterStrings:={"|", "/", "$", "#", "*", "?", "!"})
         clsgeom_density_ridges.AddLayerParameter("point_size", "numeric", "3")
         clsgeom_density_ridges.AddLayerParameter("point_alpha", "numeric", "1", lstParameterStrings:={1, 0})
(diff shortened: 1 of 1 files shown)


N-thony (approved):

rdstern (changes requested):
@anastasia-mbithe very nice except the command should have a rel_ in front.  It is:
`geom_density_ridges(rel_min_height=0.01)`

rdstern (approved):
@anastasia-mbithe that now looks fine.  Great.

lloyddewit (approved):

Issue #8288, opened 2023-04-26
Improve and fix bugs on the bar chart dialogue


rdstern:
@COLIEWO is working on this topic, and so this issue may partly be dealt with elsewhere.  @derekagorhom I hope you can start on these points, and @COLIEWO will be the main tester.  Perhaps @WezzieBanda also

Here is the current bar chart dialogue:
![image](https://user-images.githubusercontent.com/11226469/234498087-b3be685e-d2f6-4f5d-8e28-e2ccf1bb4e8f.png)

a) Change the title to just `Bar Charts`.  (We don't give much importance to Pie Charts, so let's not keep it in the title.)
b) None of the 3 controls `Reorder Frequency, Reorder and Position` remember their settings when you re-open the dialogue.  They should all remember.  @lloyddewit Derek mentioned that you may be doing something to fix this?  I now note that this control was ok in Version 0.7.6.  I wonder what has happened? 
c) The position of the 3 addutional `Add Labels` controls is messy.  And they are incomplere in Russian - at least.  I suggest moving all 3 to be on the same horizontally, with `Position `moved to perhaps start just below the A of Add Labels.  Maybe the same horizontally as `Colour and Size`, so moving the Add Labels checkbox up slightly.  Check that each label is wide enough that is shows completely in Russian, and maybe also check the French also.

@COLIEWO is likely to be adding to this issue as we explore the dialogue more fully

derekagorhom:
@lloyddewit when trying to make title name change for this dialogue, it did not have the name change as expected.... does this need another change in the database before it shows the change?

lloyddewit:
@derekagorhom yes, you are correct, we'll need to change the database before you see the name change. I'm currently working on the database so I will make the change for you in a different PR. I assume that only part `a)` requires a database change. Is that correct?

derekagorhom:
@lloyddewit yes... thank you for the quick response

rdstern:
I have now found an item d) to add, and there may be more as I (and @COLIEWO ) test the dialogue more!

![image](https://user-images.githubusercontent.com/11226469/235339782-5847c441-2ede-4550-9ccf-eb381ef4b58b.png)

The bar chart can even be used when the x variable is numeric, rather than a factor, - though it is often not very sensible.
When this is a variate, as shown above than the ascending and descewnding options don't work, and give an error.  The correction is simple.

Here is the code generated - 

[...]

Issue #8280, opened 2023-04-22
Error message on PICSA graphs when R-Instat in language other than English


rdstern:
@lloyddewit there are 3 climatic PICSA dialogues, but they all use the same code.  Here is the message when in Russian:

![image](https://user-images.githubusercontent.com/11226469/233790367-d11fb2b1-ff5d-46d3-b94e-5c443532929e.png)

It is when I press the PICSA Options button.

Oh, forgive me, I then went back to English and have just found another problem.  The 10 previous commands are still in Russian!

![image](https://user-images.githubusercontent.com/11226469/233790534-70f71662-4bf4-4f74-a2b9-edc6d6e93adf.png)

Anyway, back to the message above with the PICSA graphs.  I get that Numeric message, and assumed it would be in non-English.  It comes twice and then you can use the dialogue and it seems to give no problem after that.  The graphs come fine.  I am afraid I can't send any code, because it is in the middle of dialogue use.

Now it is actually a lot more fun, and in a way that may help?  I don't get that message when using English - or French, or Swahili!.  But I do get it in Russian, Spanish, Portuguese and Italian!  That should make it a pretty interesting puzzle!

And the other problem of the last 10 dialogues seems to be just one language change behind!  But, no, they are much more fun than that.  Perhaps this is punishment for taking a week of leave?

I just went from Swahili to English and now - as you see below - to French.  And the last 10 dialogues seem to be half Swahili and half English. 

![image](https://user-images.githubusercontent.com/11226469/233791345-649ca1f1-1503-402a-9dd5-eeb4ae10fd4b.png)

Welcome back!

[dodoma_summary.zip](https://github.com/africanmathsinitiative/R-Instat/files/11301740/dodoma_summary.zip)

Here is a data set you can use

PR #8293, opened 2023-04-27
Ensured that Climatic > Check Data > Display Daily Data dialog works in non-English languages

### LINKED ISSUE #8275: Dialogue working in English, but not when translated
@lloyddewit this is the only dialogue, so far in this workshop, with this translation problem.  I suspect I have been lucky and we should envestigate in more detail before the next workshop.

I think it is general, but I will send the data I used separately.

It is in Russian, but I suspect will be the same problem in other languages.  
It is Climatic > Check Data > Display Daily Data

image

Notice, under the label Free Y the control in blank.  It is not blank when in English.  But I am not sure that is the main error.  When I run (after filling that with Station-Element, I get an error, which didn't help me.

I tried changing the Missing Colour to blue and (to my surprise) it ran.  Then I tried again and it said:

image

I thought that was perhaps more helpful, because it has tried to change the colour name into Russian!

Then I had a look in English and then Back into Russian and look at the Missing colour control now:

image

I hope this helps?

By the way I also get a developer error (twice) when I return to the dialogue.  That does no harm though.

### LINKED ISSUE #8298: Import from Library has regressed compared with Version 0.7.6?
@ChrisMarsh82 and @lloyddewit I am starting with you, because this is easy to fix, but I wonder if there may be a more general reason for the difference now, when it was working fine in Version 0.7.6?

Import from Library in Version 0.7.6.  I go to the dslabs package and import the research funding rates.  This works fine - and the same way in 0.7.6 and the current merged version:

image

Now I re-open the dialogue.  In Version 0.7.6 I get the same as I left it - as above.  That's our norm in R-Instat.  I sometimes do want this, to look possibly at the help on the dataset I just imported.

In the latest version it has reverted to the default of the datasets package, while still keeping the name of the previous file imported.  That's not so sensible.

image

Oh, I now find that one of my changes needed on bar charts, of controls not remembering their previous settings was also alright in Version 0.7.6.  This seems to be a more general problem.  I wonder what has happened?

lloyddewit:
Fixes #8275
Fixes #8298

@rdstern 
I was only able to recreate the #8275 error intermittently. I think it's fixed but please test thoroughly.

This PR fixes the above issues by setting all input fields (including combo boxes), in all dialogs, to 'do not translate'. My rationale is that these controls contain values to be used literally in the R code. Does this make sense? Please could you do a sanity check on a few other dialogs with input fields to see if the translations are still as expected.

Thanks for your help

### DIFF
--- a/instat/translations/translateIgnore.txt
+++ b/instat/translations/translateIgnore.txt
@@ -94,3 +94,8 @@ dlgClimaticBoxPlot_ucrInputWithinYear%
 # In the 'InventoryPlot' dialog, the selected 'Facet' should not be translated
 # Partially fixes issue #6989, related to PR #7070
 dlgInventoryPlot_ucrInputFacetBy%
+
+# In any dialog, input text should not be translated
+# Fixes issue #8275
+# TODO If by 2024, this line has not caused any problems, then delete ucrInput controls specified above (already covered by this pattern)
+%_ucrInput%


lloyddewit:
@rdstern I think this PR also fixes the first part of issue #8280. Please could you test this also?
Thanks

derekagorhom:
@lloyddewit i tested this PR and it solves the issue of item b) from Issue #8288. If @rdstern is ok with this i can make the Pull request for item a) and c) separately.

lloyddewit:
@rdstern I think this PR may fix some of the regression problems you're having. It would also be good to merge the database as soon as possible. Please could you test? thanks

lloyddewit:
> Notice a very cold winter in 2008 ...
Wow, amazing temperature ranges.
Happy it's working now. Apologies for the regression.

PR #8295, opened 2023-04-28
Ensure that list of last 10 opened dialogs is displayed in correct language

### LINKED ISSUE #8280: Error message on PICSA graphs when R-Instat in language other than English

lloyddewit:
Fixes #8280.
The main tool bar has a drop down list of the 10 most recent dialogs opened. This PR ensures that this list is always displayed in the current language (English, French etc.).


lloyddewit:
Ongoing, not yet ready for testing

lloyddewit:
@rdstern This PR is now ready for testing.

One small idiosyncracy is that if the language is non-English (e.g. Russian) and the user toggles to English, then the list of dialogs is translated from Russian to English (as expected). If the user then opens a new dialog, then the dialog is in Russian (as expected). When the user closes the Russian dialog, then the list of dialogs is shown in Russian (which may look odd to the user when the rest of the menu bar is still toggled to English).
I personally think this is OK, because the user has just been using the dialog with the Russian title. If needed, the user can still toggle the language again using the button on the menu bar.

I didn't resolve this idiosyncracy because it adds complexity to the code (I have to change the way the menu toggle works) and I wasn't sure it was worth it. 
But if you'd like the idiosyncracy resolved, then I'm happy to do it.

thanks

rdstern (approved):
@lloyddewit this all looks great.

If you are ever bored, then I notice that tooltips have yet to be translated - unless this mat be in the new database?

PR #8303, opened 2023-05-02
Database Changes to BarandPieChart Dialogue

derekagorhom:
fixes Partly #8288 
@lloyddewit i made a database change to the title of the dialogue

image

I made it `replacewithdynamictranslation` for any future change so that we don't have to go into the database to make the change.

I hope this is ok with you.


lloyddewit:
@derekagorhom Thank you for making this change. In parallel I had also changed the database so I implemented your change in PR #8307 and I will close this PR.

I agree that setting to `ReplaceWithDynamicTranslation` is more convenient in some ways. However, there are some other advantages to setting the text specifically where possible (i.e. for text that does not change at runtime). This is part of a bigger discussion and we should be consistent. For now, I set 'id_text' to 'Bar Chart'. I hope that's OK.

thanks

Issue #7695, opened 2022-07-30
Correct error on the mapping dialogue and possibly improve the R-code?


rdstern:
I think this may be an @anastasia-mbithe challenge?

I get the error in the following way:
a) Run the initial layer - a gadm file.
b) Add a second station file and plot points.  
![image](https://user-images.githubusercontent.com/11226469/181879658-b76b4695-7d18-431a-8552-d508d4f62069.png)
This runs fine.
c) Now I wanted to change the size of the labels - which is geom label repel.  I go to that geom and make the changes.  
Now it gives me an error.
Here is the R-code that gives the error:

```
# Code generated by the dialog, Climatic Maps

stations <- data_book$get_data_frame(data_name="stations")
gadm36_RWA_2_sp <- data_book$get_data_frame(data_name="gadm36_RWA_2_sp")
last_map <- ggplot2::ggplot(data=stations, mapping=ggplot2::aes(x=V1, y=V2, label=location1)) + ggplot2::geom_sf(data=gadm36_RWA_2_sp, mapping=ggplot2::aes(fill=NAME_1)) + ggplot2::geom_point(mapping=ggplot2::aes(x=V1, y=V2)) + ggrepel::geom_label_repel(label.size=0.1, label.r=0.08) + theme_grey() + ggplot2::theme(axis.text.x=ggplot2::element_text(), legend.position="none")
data_book$add_graph(graph_name="last_map", graph=last_map, data_name="gadm36_RWA_2_sp")
data_book$get_graphs(data_name="gadm36_RWA_2_sp", graph_name="last_map")
rm(list=c("last_map", "stations", "gadm36_RWA_2_sp"))

```
It says it can't find the variable V1, so perhaps it is looking at the wrong layer in the graph?

rdstern:
I suggest @NanyanziAlice might work on this issue and perhaps then @MeSophie could do the programming.

The mapping was a popular component of the Mozambique and Malawi workshops.  The error above was avoided, through not attempting to change the label size.  This should be investigated, and perhaps there should be an option on the main dialogue.  

In addition the item that was most popular was to filter the 2 data frames to just, then show a single district.  This could be made much easier if those filters could be on the main dialogue - both parts.

There should also be an investigation of other options, e.g. the colour of a map, etc.

We did not look at facets - usually having a facet for each station type.  But let's check that this is also working nicely.

MeSophie:
[...]

PR #8305, opened 2023-05-03
Add label options to the Mapping dialog

MeSophie:
Fixes partially #7695 

@rdstern I corrected the error on label maping Dialog.

For the Label and Legend option can we add a check box Add Legend and Add Label? And please more information about the filter part will be very helpful. Thank you.


rdstern:
@MeSophie I can now edit the labels, as you say.
But - oddly I can't change their size (or I think colour either).

MeSophie:
@rdstern I'm done with improvements. Please can you check.
`max.overlaps = Inf` is working well I try it with this code in r.

**- Without** `max.overlaps = Inf`

image



**- With** `max.overlaps = Inf`

image

rdstern (changes requested):
@MeSophie this is exciting. And looking great. 
a) Change the Label Rectangle to Text and have it unchecked by default.  It also doesn't remember the setting, so make it do that.
b) Have a checkbox, default unchecked for the Size option - as it is in the sub-dialogue.  When checked make the default 5, (as for the nlabel and text geoms, in both the dialogue and the sub-dialogue.)
c) The Label All doesn't remember the previous setting.  Make it do this.
d) The R-text calls the geom_label_repel twice.  I think that may be resulting in the text option not working (ot always gives a label).  And the size doesn't seem to work, from the main dialogue.  It does work from the sub-dialogue.
e) Could you move the box up so it starts just below the data selector.  That will give space for a 4th checkbox, with a copy of the Colour option from the sub-dialogue!

rdstern (changes requested):
@MeSophie This is looking excellent now.
There are just some option settings to tweal a bit and then you are through!
a) The default is correctly with the Text unchecked, and therefore it should be a label - so with the box round the names.  Currently the default (when it is unchecked) is now Text.  I would like the user to have to check to get the text.
b) Then sometimes the Colour and size reset to their default of Black and 5.  And sometimes they remember the previous setting.  I think they may remember for text and not for label, but I am not sure?

MeSophie:
@rdstern Please can you check the new changes? Thank you.

rdstern (approved):
@MeSophie very nice now.  Many thanks.  I hope @lloyddewit will like the code.

lloyddewit (approved):

PR #8307, opened 2023-05-03
Added specification of controls that should be translated dynamically

### LINKED ISSUE #6659: Themes subdialog of Plot options
Hello @shadrackkibet,

I have been working on the Themes subdialog and have found an number of areas that are unclear to me. For example, under XY-Axes multiple items are titled Element Text or Element Line, it is unclear what they refer to. There is also something called grpTickLength. There are similar things on each tab where multiple items have the same names.

Could we add tool tips or change the labels to make them clearer, or both? 

@rdstern suggests this could be addressed within the sprint.

AxesXY
Legend more options

Thanks, 

Rachel

lloyddewit:
Fixes #6659

This PR adds a specification file to specify WinForm controls that have dynamically changing text. This file is used to update the translation database so that the translation mechanism translates the control's current text, rather than the control's default text.
For example, this file can be used to ensure that a label that changes depending on the context is correctly translated into a language such as French or Portuguese.


lloyddewit:
@rdstern Please note that there are a few text items not yet translated into Russian. This is because the text was set in a non-standard way. I corrected this so that the text will be translated correctly in the release we make for the May training course. So please ignore this for now. Thanks

image

lloyddewit:

rdstern (changes requested):
@lloyddewit It is looking really good now.  I am using it is Russian and it has a much more "solid" air to it.
1) However there is still a problem with the Climatic > Check Data > Display Daily.  The By variable is still blank - as I remember it before on the course - but I think there was a correction?  And, like that it gives an error when trying to run.

When I switched back to English it ran ok as shown below:  There it gave me a default option of Station-Element and worked fine.  In English, from Russian it only offered Station which works, but was not the ideal graph

image

b) I tried a character variable to numeric - with right click (and with a non-numeric character and this is largely still not translated.  I assume that's for another time?

image

c) The last 10 dialogues is still working really fine.

d) I am having problems with the PICSA dialogues.  The default graph worked fine.  But when I used the sub-dialogue and returned I found the year variable had been translated into Russian in the dialogue.

image

Another oddity we had on the course, with this dialogue and I had once here (but it then went away, was that there are 2 stations and so should produce 2 separate graphs.  But sometimes the Facet-Wrap doesn't work, and there is just one of them.  Then, the next time, with no changes, there were (correctly) 2 graphs.  I need to examine the code next time I get it.
(Yes, it just ignored the facet-wrap., which appeared the second time?)

lloyddewit:
@rdstern Thanks for testing and finding the issues.
> a) There is still a problem with the Climatic > Check Data > Display Daily.  The By variable is still blank - as I remember it before on the course - but I think there was a correction?  And, like that it gives an error when trying to run.
> 
> When I switched back to English it ran ok as shown below: There it gave me a default option of Station-Element and worked fine.

I think this is fixed, please could you test?

> b) I tried a character variable to numeric - with right click (and with a non-numeric character and this is largely still not translated. I assume that's for another time?

Correct, pop up messages are not currently translated. See issue #3721.

> c) The last 10 dialogues is still working really fine.

Way hay :)

> d) I am having problems with the PICSA dialogues. The default graph worked fine. But when I used the sub-dialogue and returned I found the year variable had been translated into Russian in the dialogue.

I think this is also fixed now, please could you test?

> Another oddity we had on the course, with this dialogue and I had once here (but it then went away, was that there are 2 stations and so should produce 2 separate graphs. But sometimes the Facet-Wrap doesn't work, and there is just one of them. Then, the next time, with no changes, there were (correctly) 2 graphs. I need to examine the code next time I get it. (Yes, it just ignored the facet-wrap., which appeared the second time?)

Do you think this is a translation issue or something else?
Would it be worth creating a separate issue for this with steps to recreate the problem?

rdstern (approved):
@lloyddewit I have tested the PICSA graphs extensively and that seems fine now.

There is still a problem with the Climatic > Check Data > Display Daily Data dialogue.  Though this still seeems slightly different in non-English, compared to English as I describe below, I can also get the problem in English, so the issue is probably not mainly caused by the language setting.  Hence I am approving.  Then the issue below should be provided separately.

I think with any climatic dataset, when I press Reset in English I usually get Station in the By field.  Notice the Comment field is as follows `Code generated by the dialog, Display Daily Data`  .  I now claim that I will get different code at the end of the R command and it will not work.
image

Once I complete the dialogue, (with 2 variables) the By field changes to Station-Element - which is correct.  But I still get an error and the graph is not produced.

When I return to the dialogue I notice that the Comment has now changed.  It now is less informative and just says  `Code generated by the dialog`

It now runs fine.  

The old code - with then error is as follows:

```
# Code generated by the dialog, Display Daily Data
append <- data_book$get_data_frame(data_name="append", stack_data=TRUE, measure.vars=c("tmin","tmax"), id.vars=c("id", "date"))
append <- data_book$get_data_frame(data_name="append", stack_data=TRUE, measure.vars=c("tmin","tmax"), id.vars=c("id", "date"))
last_graph <- ggplot2::ggplot(data=append, mapping=ggplot2::aes(x=date, y=value, colour=variable)) + ggplot2::geom_line() + ggplot2::geom_rug(data=append %>% filter(is.na(value)), colour="black") + theme_grey() + ggplot2::theme(axis.text.x=ggplot2::element_text(), legend.position="none") + ggplot2::facet_wrap(scales="free_y", ncol=1, facet=id ~ variable, labeller=label_wrap_gen(multi_line=FALSE)) + ggplot2::xlab(NULL)
data_book$import_data(data_tables=list(append=last_graph))
data_book$get_data_frame(data_name="append")
```

The second - correctly running code is:

```
# Code generated by the dialog
append <- data_book$get_data_frame(data_name="append", stack_data=TRUE, measure.vars=c("tmin","tmax"), id.vars=c("id", "date"))
append <- data_book$get_data_frame(data_name="append", stack_data=TRUE, measure.vars=c("tmin","tmax"), id.vars=c("id", "date"))
last_graph <- ggplot2::ggplot(data=append, mapping=ggplot2::aes(x=date, y=value, colour=variable)) + ggplot2::geom_line() + ggplot2::geom_rug(data=append %>% filter(is.na(value)), colour="black") + theme_grey() + ggplot2::theme(axis.text.x=ggplot2::element_text(), legend.position="none") + ggplot2::facet_wrap(scales="free_y", ncol=1, facet=id ~ variable, labeller=label_wrap_gen(multi_line=FALSE)) + ggplot2::xlab(NULL)
data_book$add_object(data_name="append", object_name="last_graph", object_type_label="graph", object_format="image", object=check_graph(graph_object=last_graph))
data_book$get_object_data(data_name="append", object_name="last_graph", as_file=TRUE)
```
The last 2 lines are different.

In non-English the By field is initially blank and you may have to complete it manually.  The behavior is similar in that the first time (or after reset) it runs differently and gives an error.  The second time (assuming you complete the By field) it runs ok.

lloyddewit:
@rdstern I understand that the bug can be recreated in English but the behaviour is slightly different in non-English languages.
I think the best way to approach this is to first fix the English behaviour. If there are still problems after this in other langauges, then we should raise (and label) a new issue for the translation problem.

I raised issue #8309 for this. I will go ahead and merge this PR. I hope that's OK.
Thanks

PR #8315, opened 2023-05-09
Ensured that tool bar tool tips translate correctly between English and other languages

lloyddewit:


rdstern (approved):
Great - seems to work fine.  I checked, both on the toolbar and also in a dialogue.  Both seem to work - though my Russian has not improved.

PR #8319, opened 2023-05-09
Translation changes

### LINKED ISSUE #6556: Translation Layout Issue in Import Dataset Dialog
@Patowhiz I have been looking over the Import Dataset Dialog to check how the French translation looks. 

I have found that when I import an .RDS file the Overwrite existing if duplicates (checkbox text) is cut off in French. However, I cannot find it in Visual Studio to change it. Please could you tell me where/how to find the different layers? 

Also how many different configurations of this dialog are there? I need to check them all. 

It seems a very complex dialog! (I will stop working on it until I hear back) 

Thanks, 

Rachel 
Import dataset

### LINKED ISSUE #6563: Small point on our base control in multiple languages
Here is an example with our base control:

image

I am interested in the base control with the Comment line.  A trivial point is that the text could move over  very slightly, so the colon is clear.  But this is on every dialogue and I wonder if it could be translated?  There is an initial sentence, and then (I assume) the name of the dialogue as shown in the top left.  This then appears in the script and output windows, or can be edited by the user.

(That reminds me that we are inviting users to be using R-Instat with their own language settings on their machines.  I must mention this to our language testers, so we are prepared for possible problems!)

### LINKED ISSUE #6700: Save Urc issues in Load Dataset from Library, Row names and Numbers, Add Key and Recode Numeric Into Quantiles
Hi @Patowhiz,

 In Open Dataset from Library the Save area is not displaying properly. You should be able to see a colon at the end of Nouveau nom de tableau de donnees, when I make the item bigger instead of moving the box over it just makes the white box really big but it still does not show the colon and it looks cut off.  

Open data set from library

In Row Names and Numbers the position button is only just visible poking out from behind the white box. This should be shorter so that the Position button is fully visible. I know that it is the position button poking out as if you click on it it opens the positioning subdialog. 

Row Names or numbers no position button

In the Add Key dialog the Text box is very far away from Key (the translation is Cle so this is not necessary for the French). However, if I resize the control it doesn't move the box closer to its title it just cuts the box off.  

Add key

In the Recode Numeric into Quantile, this does not display correctly in English either 

Recode numeric into Quantiles

### LINKED ISSUE #7598: French lay out issues in the Import from file dialog
Hello @N-thony and @lloyddewit, 

Apologies for bombarding you. This is a dialog that we did not address on the first round as it was complicated and due to be updated. It has a number of issues. @N-thony please could you review it and correct them?

- Top Checkbox: Tous les fichier de ce type (type is not visible)
- Première ligne est les noms de colonnes (I updated the crowd-in translation to noms instead of en-têtes but cannot think of a shorter version than that. @N-thony can you think of a good way to shorten it or to expand the box?)
- Chiane de valeur manquante  the text overlaps the box. I have updated this to Valeurs manquantes in Crowd-in so this may resolve itself, but perhaps you could check.
- Convertir des chaînes en colonnes de facteur Checkbox: facteur cut off. 
- Nouveau nom de tableau de données: the colon at the end is cut off. 
- Drop Empty Row and Columns Checkbox: Not translated, perhaps this is simply a new item that is not yet in the system? I cannot see it in crowdin. 
- Is Cannot import this file with the current options.... message is not translated. Should it be? (some of the messages do not than that is ok.)

I also noticed that when I check the Import all files of this type box, I get a different view. This has a message that is translated but disappears off the page.

### LINKED ISSUE #7749: Translation issue on Data Frame, Check Data, Column:Numeric, and Data Reshape sub-menu of Prepare menu

### LINKED ISSUE #8247: Some simple language changes
Various points are shown here.  I use Portuguese, and am not sure which of the points shown are also in other languages.  I wonder who is best placed to fix them, but suggest these mostly don't yet need a Portuguese speaker.

image

a) In the dialogues the 5 controls at the bottom are hardly translated.  
b) The Close is translated, but into capitals.  That's just in portuguese - it seems ok in the French.
c) Help is not translated.  In the main menu it is translated, but I gather wrongly for Portuguese - It should be Ajuda and not Socorro (that's more like first aid).
d) Reset isn't translated.  
e) None of the buttons are translated in the new script window.

lloyddewit:
Fixes #8247
Fixes #7749
Fixes #6563
Fixes #7598
Fixes #6700 
Fixes #6556

### DIFF
--- a/instat/dlgImportDataset.Designer.vb
+++ b/instat/dlgImportDataset.Designer.vb
@@ -52,52 +52,21 @@ Partial Class dlgImportDataset
         Me.lblLinesToSkipCSV = New System.Windows.Forms.Label()
         Me.cmdBrowse = New System.Windows.Forms.Button()
         Me.grpCSV = New System.Windows.Forms.GroupBox()
-        Me.ucrNudMaxRowsCSV = New instat.ucrNud()
-        Me.ucrChkMaxRowsCSV = New instat.ucrCheck()
         Me.lblNAStringsCSV = New System.Windows.Forms.Label()
-        Me.ucrChkStringsAsFactorsCSV = New instat.ucrCheck()
-        Me.ucrNudRowsToSkipCSV = New instat.ucrNud()
-        Me.ucrInputHeadersCSV = New instat.ucrInputComboBox()
-        Me.ucrInputMissingValueStringCSV = New instat.ucrInputTextBox()
-        Me.ucrInputDecimalCSV = New instat.ucrInputComboBox()
-        Me.ucrInputSeparatorCSV = New instat.ucrInputComboBox()
-        Me.ucrInputEncodingCSV = New instat.ucrInputComboBox()
         Me.grpExcel = New System.Windows.Forms.GroupBox()
         Me.lblTo = New System.Windows.Forms.Label()
-        Me.ucrChkRange = New instat.ucrCheck()
         Me.lblFrom = New System.Windows.Forms.Label()
-        Me.ucrInputTextTo = New instat.ucrInputTextBox()
-        Me.ucrInputTextFrom = New instat.ucrInputTextBox()
-        Me.ucrNudMaxRowsExcel = New instat.ucrNud()
-        Me.ucrChkMaxRowsExcel = New instat.ucrCheck()
-        Me.ucrNudRowsToSkipExcel = New instat.ucrNud()
         Me.lblMissingValueStringExcel = New System.Windows.Forms.Label()
-        Me.ucrInputMissingValueStringExcel = New instat.ucrInputTextBox()
-        Me.ucrInputNamedRegions = New instat.ucrInputComboBox()
         Me.lblNamedRegion = New System.Windows.Forms.Label()
         Me.lblRowToSkipExcel = New System.Windows.Forms.Label()
-        Me.ucrChkColumnNamesExcel = New instat.ucrCheck()
-        Me.ucrChkTrimWSExcel = New instat.ucrCheck()
         Me.grpRDS = New System.Windows.Forms.GroupBox()
-        Me.ucrChkOverwriteRDS = New instat.ucrCheck()
-        Me.ucrChkImportObjectsRDS = New instat.ucrCheck()
-        Me.ucrChkImportCalculationsRDS = New instat.ucrCheck()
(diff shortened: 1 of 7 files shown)


lloyddewit:
Ongoing, not yet ready for testing

lloyddewit:
@rdstern 
I reviewed all the open issues related to translation. The majority were already fixed so I was able to close them. I fixed most of the remainder in this PR.
Please could you test?

When this PR is merged, the translations should be in a relatively good state, so I suggest we make a release. Is this OK from your side or are there other specific open PRs you would explicitly like in this release?

When this PR is merged, there will only be a couple of translation issues still open (e.g. issue #3721). These require changes to the translation mechanism and will take longer. So I suggest that we do these later and do not include in this release.

I hope that's OK, thanks for your help

rdstern (approved):
@lloyddewit this is now looking great.  You mention that only a couple of translation aspects remain, and mention #3721 .  I note that some points there, e.g. tooltips have now been translated (great).  I found a few things that remain, and note them here, but would really like to merge this now, so the team can start using this version in the different languages.

a) Some dialogues have items, like colours, that are used directly in commands.  An example is below.  If these are ever translated, then the English version still needs to  be used in the R-command.

Here is a Climatic > PICSA > Rainfall graph sub-dialogue that shows this feature:

image

2) The message at the bottom of a dialogue remains in English.  Also, the (sort of) tooltip shown here is still in English.  
a) I wonder if the tooltip is a different sort, because it doesn't  disappear after 30 seconds or so, like all the others.
b) Ther other messages at the bottom are quite important, but we don't have much freedom to change their width, because it is a bit dynamic.


3) The pop-up menus remain, or are just partially translated.  

 For example:

image

No, this is translated.  I'll find another one.  But this lets me ask something else - I am not sure how automatic the (machine translation is now.
The above is "Enter Window Number" in Englsish which becomes quite long in French and possibly also in other languages.  So here I am inclined to change the English into **_Enter the Number:_**  (It is obvious that it is the Window number from the title of the dialogue, just above.)  If I do that, then will it now automatically still be translated in all languages, or do we have to change individually?

Here is a popup that remains:  

image

 






:

PR #8322, opened 2023-05-11
Improved layout of New Data Frame dialog

### LINKED ISSUE #8321: The New Data Frame Dialog Buttons
On opening the new data frame dialog, for both Kiswahili and English, the buttons are separated. Initially, all five buttons were together, now the first and second buttons are on their own and the last three are together. I suggest that all 5 be stuck together for a better look.

Also when I am using Kiswahili as my language the translations for columns and rows overlap with the receivers. I suggest reducing the sizes of the box to allow the text to fit well.
Here is an image of the dialog.
new

derekagorhom:
fixes #8321 

I have put the radio buttons back together again from before and also moved the  drop down controls a bit to the right, so the Kiswahili text is complete. 

@rdstern and @berylwaswa this PR is ready for review 

Thanks


berylwaswa:
@derekagorhom here I see the buttons are now well together looking good but the words for column and rows are still not well placed with the boxes, please move them a bit to the right or reduce the size a bit for Kiswahil.
instat_BGibZvqxAX

derekagorhom:
@berylwaswa thanks for the correction. I have made the new changes, you have a look again.
thanks

rdstern (changes requested):
@derekagorhom thank you for making the changes so quickly.

Now the up=down controls are moved over, but they have alsao become very wide.

image

Here trhey are in your latest version:

image


I like their position now, because the Swahili is complete.  But could you return them to the previous width.

Thanks

derekagorhom:
@rdstern i was able to make the changes to the width of the up-down button

thanks

rdstern (changes requested):
@derekagorhom maybe I am doing something wrong here.  The up-down now seems still wide, but have moved back to where they were before.  That's the opposite of what I hoped for?
I hoped they would remain in their new position, but be the same width as in earleir versions.

rdstern (approved):
@lloyddewit this now seems to be fixed, so I am approving.  I hope the code is ok.

For the record, here it is on my screen now:

image

lloyddewit (approved):

PR #8324, opened 2023-05-11
Translation changes requested by PR #8319 comments

### LINKED ISSUE #8325: Odd right-click options not yet translated?
@lloyddewit I note the main data right-click menu seems now to be fully translated.

But the Row right-click omits the translations for the top 2 entries:

image

The bottom right-click seems to be fully translated, except for `R Viewer`

Now that one has become more interesting.

In the Column Information it is translated fine - and it is different - for no good reason.

image

There it is `View Data Frame` and IS translated fine.  In the Data View it is `R Viewer` and is not translated.  Could you please change it there into View Data Frame - for consistency, and then perhaps it will also be translated automatically? 

By the way I also checked the right-click in your script window and that looks fine.

lloyddewit:
Fixes #8325 

@rdstern Thank you for the discussion this morning about your comments in PR #8319. Here are the requested changes for the Window Number popup and the row/column status at the bottom of the data grid.
Please could you test?
If you're willing to approve today, then we can include in the new release.
Thanks

### DIFF
--- a/instat/Translations.vb
+++ b/instat/Translations.vb
@@ -74,6 +74,10 @@ Public Class Translations
             clsMenuItem.Text = GetTranslation(clsMenuItem.Text)
         Next
 
+        'The data grid status bar is dynamic and isn't translated as part of the main form.
+        '    So translate manually.
+        frmMain.ucrDataViewer.SetDisplayLabels()
+
         'The right mouse button menus for the 6 output windows are not accessible via 
         '    the control lists. Therefore, translate these menus explicitly
         HandleError(TranslateWinForm.clsTranslateWinForm.TranslateMenuItems(frmMain.ucrOutput.Name, frmMain.ucrOutput.UcrOutputPages.tsButtons.Items, strDbPath, strLanguageCode))
--- a/instat/sdgWindowNumber.Designer.vb
+++ b/instat/sdgWindowNumber.Designer.vb
@@ -32,29 +32,27 @@ Partial Class sdgWindowNumber
         '
         Me.lblTitleSize.AutoSize = True
         Me.lblTitleSize.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.lblTitleSize.Location = New System.Drawing.Point(6, 29)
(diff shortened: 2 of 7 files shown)


rdstern (commented):
@lloyddewit so far-so good, except that the control shown looks a bit odd now you have omitted all the text.  Could you just add `Number: `there?

image


@berylwaswa there seems an oddity in the current Swahili translation.  I notice both Row: and Column: are translated as Safu:  In a statistics package, having Rows and Columns as the same is desatrined to sow confusion - isn't it?

lloyddewit:
@rdstern I added the 'Number' label as requested, please could you test? thanks

berylwaswa:
@rdstern rows have to be Safu mlalo and columns Safu wima so as to avoid confusion.

rdstern (dismissed):
@berylwaswa thank you for the addition for the Kiswahili.  I assume that is a crowdin change that you can make yourself?  And we now have the video/practice on how to do this?

I am approving.  I am very excited about the language improvements for this release.

lloyddewit:
Technical note, database changes:

image

image

lloyddewit:
@rdstern I also fixed issue #8325 , please could you test?

Notes:
- In the Data Frame Metadata right click menu, I also changed 'R Viewer' to 'View Data Frame' just to be consistent, I hope that's OK.
- The 'View HTML' menu options are not translated. This is because this string is not yet in CrowdIn. This should be fixed automatically in the next release (but we should test!).
- In the Column Metadata right click menu, the Delete Column menu option does not toggle languages correctly. This is because the text is dynamic based on how many rows are selected. I think this is over complex. I suggest that we change it to 'Delete Column(s)' to be consistent with the other menu options. It should then translate/toggle correctly. If you agree, I'll do this in a separate PR.

rdstern (approved):
In answer to your comments:
1. Thanks for changing the R Viewer label.
2. The dynamic nature of the Delete Column/Columns is on both the Data View Column Right Click and the Metadata right click.  Happy for them both to change to Delete Column(s), as you suggested.

PR #8328, opened 2023-05-16
Add version number back in

ChrisMarsh82:
@lloyddewit this will add the version number back in. 

Looks like it was removed in this PR https://github.com/africanmathsinitiative/R-Instat/pull/8232

### DIFF
--- a/instat/frmMain.Designer.vb
+++ b/instat/frmMain.Designer.vb
@@ -5056,7 +5056,9 @@ Partial Class frmMain
         Me.IsMdiContainer = True
         Me.MainMenuStrip = Me.mnuBar
         Me.Name = "frmMain"
-        Me.Text = "R-Instat  .."
+        Me.Text = "R-Instat  " + My.Application.Info.Version.Major.ToString + "." +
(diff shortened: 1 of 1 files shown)


lloyddewit (approved):

lloyddewit:
@ChrisMarsh82 Thank you for fixing this so quickly.
I remember that we've had this problem before. I assume that if a developer changes `frmMain.Designer.vb` (which happens regularly), then the Visual Studio designer tool will automatically overwrites this change with a fixed string. It seems likely that this problem will keep recurring.
Should we update the release building documentation to include a check to see if the code is overwritten? Or should we move this code into `frmMain.vb` (I assume that it only needs to be called once at startup)?
What's your view?

ChrisMarsh82:
@lloyddewit This did keep happening at first as developers didn't pull the latest code before making changes. This is now not the case as it has been in for so long. The problem this time was it was deliberately changed to "R-Instat  .."

lloyddewit:
That makes sense. In that case, I agree that no changes to the process or code are required. Thanks for the explanation.

PR #8329, opened 2023-05-16
Ensured that Import Dataset dialog does not hang when opened in French

lloyddewit:
@rdstern When you tested R-Instat 0.7.12 you reported that if the user tries to open the Import Dataset dialog in French, then the dialog does not open and R-Instat hangs. This PR fixes this. Please could you test?
thanks

Notes:
- This PR fixes some regression inadvertently introduced by PR #8319
- This PR does not fix the issue with the apostrophe in the Import Dataset comment field. This will be fixed in a separate PR.

### DIFF
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -919,14 +919,12 @@ Public Class dlgImportDataset
         TryTextPreview()
         TryGridPreview()
         TestOkEnabled()
-        autoTranslate(Me)
     End Sub
 
     Private Sub cmdRefreshPreview_Click(sender As Object, e As EventArgs) Handles cmdRefreshPreview.Click
         TryTextPreview()
         TryGridPreview()
         TestOkEnabled()
-        autoTranslate(Me)
     End Sub
 
     Private Sub HideDropEmptyCheckBox()
@@ -947,7 +945,6 @@ Public Class dlgImportDataset
         TryGridPreview()
         TestOkEnabled()
         HideDropEmptyCheckBox()
-        autoTranslate(Me)
     End Sub
 
     Private Sub MissingValuesInputControls_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrInputMissingValueStringText.ControlContentsChanged, ucrInputMissingValueStringCSV.ControlContentsChanged, ucrInputMissingValueStringExcel.ControlContentsChanged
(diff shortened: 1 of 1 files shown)


rdstern (commented):
@lloyddewit it is much improved and doesn't hang.  However, I am not sure it is completely fixed.  

Here it is after opening 4 times.  You notice the string has become progressively longer.

image

And the fix you mentioned doesn't work, because the Comment window doesn't show the apostrophie.  

image

(I treid deleting and the string is still getting progressively longer!!!

So, it is better, but not cured.  Do you want me to approve, or you just merge?  It is certainly better!.

lloyddewit:
@rdstern Is it possible that the apostrophe is still in the comment but has been pushed down to a lower line and is no longer visible? 
If you click on `...` in the comment field, then do you still see an apostrophe?

Note that deleting the apostrophe is just a temporary workaround. To fix the apostrophe issue properly, I need to make a new version of the library and publish it on Nuget. This may take a few days.

rdstern (approved):
@lloyddewit you are right - that work-around works.
If I get rid of all the `'''''''''''` the first time, then it is ok ever after.  So I am happy now!  Can it be merged?

rdstern:
@lloyddewit I don't know if there is anything [here](https://stackoverflow.com/questions/1250079/how-to-escape-single-quotes-within-single-quoted-strings) that will help the long term solution to this problem.

It mentions the single quote problem and there are 25 proposed solutions there?  

I assume this is all linked to the fact that R uses regex by default?  I also want to be sure we don't have problermsd when we use a single quote in any text we might add ourselves - like the word don't!.  I tried putting my own `'`  and it gradually expanded into many of them.  I wonder if this is the only symbol like this.  I tried a " and that seems harmless!

lloyddewit:
@rdstern Thank you for the link. It describes similar issues with single quotes using bash scripting.
Our problem is related to SQL. SQL treats a single quote as a special character so we need to 'escape' any single quotes we have in the string. In SQL, the way you do this is that you specify `''` for each single quote that you want. The `TranslateWinForm` library we use does this correctly. However it has a bug. If it can't find a translation in the database, then it should return the original untranslated string. But it currently returns the original untranslated string **including** the 'escaped' single quotes. This means that every time R-Instat tries to translate the string it doubles the number of single quotes in the string. See code below.
The fix itself should be simple. The more time consuming part is making/publishing the new library. However this was on my to-do list before this bug because we anyway have a new version of this library. So this was work I needed to do anyway.
I hope this makes sense.

I will go ahead and merge this PR.

image

Issue #8248, opened 2023-03-30
Moving from mmtable2() to gt()


lilyclements:
@anastasia-mbithe after a discussion with @rdstern we have agreed that we should forgo the flexibility that `mmtable2` provides for `gt()` table directly. This is to allow for the flexibility that `gt()` offers.

Your work on the sub-dialogs is very impressive, and still holds really nicely so we don't need any changes on the sub dialogs from this move. In fact, doing this move will make things a lot easier on future sub-dialog code, and will add even more power to the tables!

The only changes are on the one variable, two/three variable, and (general) tables dialogs where we now remove all `mmtable2` code. 

E.g. we would replace
```
last_table <- mmtable2::mmtable(data=summary_table, cells=value) + mmtable2::header_left_top(variable='variable') + mmtable2::header_top_left(variable='summary')
```
with
```
last_table <- summary_table %>% gt()
```

I suggest we just focus on the "(general) tables" dialog for now, and can move to the other two later.

**Changes:**
There is one dialog change needed - and that is to remove the `Column Factors` nud and instead add in a set of rdos titled `Column Factor`.

If "Treat summaries as further factor" is **not** checked: Column Factor has three options - 
1. No column factor
2. Factor variable (if checked, a single receiver is displayed. This should be restricted to only variables in the factors receiver - this is implemented in another dialog, I'm just trying to remember an example. Do you know?)
3. Summary-Variable

If "Treat summaries as further factor" is checked: Column Factor has four options - the same two as above, but replace 3. with:
3. Summary
4. Variable

If 1. is checked, then the code is just:
`summary_table %>% gt()`

If 2. is checked, then the code is:
```
summary_table %>%
  pivot_wider(names_from = <variable in receiver>, values_from = value) %>%
  gt::gt()
```

If 3. is checked and "Treat summaries as further factor" is **not** checked, then the code is:
```
summary_table %>%
  pivot_wider(names_from = `summary-variable`, values_from = value) %>%
  gt::gt()
```

If 3 is checked and "Treat summaries as further factor" is checked, then the code is:
```
summary_table %>%
  pivot_wider(names_from = summary, values_from = value) %>%
  gt::gt()
```

If 4 is checked:
```
summary_table %>%
  pivot_wider(names_from = variable, values_from = value) %>%
  gt::gt()
```

Does this make sense?

lilyclements:
I'll put here the places we want to make these changes

[...]

PR #8330, opened 2023-05-16
Changed underlying R function in dlgOneVariable Summarise > customised tab (from mmtable function to gt function)

anastasia-mbithe:
This is a continuation of issue #8248 
@rdstern and @lilyclements, This is ready for review.


rdstern (changes requested):
@anastasia-mbithe this is great.
a) I am not sure if you changed the skim option, but it now does seem to provide all the numeric summaries on a single line, which is great.  
b) It includes the Summaries button which I assume should be hidden.
c) On the customised could there be a group box with a label Rows:  Then I suggest the first (default) option might be Variables, see below.  The other two options are just as you have them now, namely Variables and (lastly) None.
d) Could there be another option Display Missing as: , with a drop down for options, and the facility to type youe own.  NA is the (visible) default followed by (blank) which ideally is written like that, but gives a blank.  Otherwise it is just spaces.  Then perhaps . is another option and --- might also be provided?

lilyclements (commented):
@anastasia-mbithe this is looking great - I have nothing to add. Let me know once Roger's suggestions are implemented and I'll re-review. Nice one!

rdstern (commented):
@anastasia-mbithe that's looking very nice now.
You are quite right - only an idiot (me) would say Rows when it should be Columns!
And a tiny detail - I can't see exactly whether the Options button is lined up with the Omit Missing Values checkbox, or is slightly low?

rdstern (commented):
@anastasia-mbithe this is looking very nice now.  Here is an example, when I made fert into an ordered factor!  It shows that then there is a min and a max - which you don't have for an ordinary factor.  It even calculates the median - I didn't realise it could do that!

image


I hope it might now be easy to add 2 controls from the general summary tables dialogue

image

This would make it more useful, and it can also be a teaching tool.  What is a table?  How can we make it more attractive, and so on!

And could you also delete the As from the Display Missing As.  I think having the As there in English is sort of ok, but it could be tricky to translate.

rdstern:
@lilyclements and @anastasia-mbithe could you please also check the code you are using?  It is all working swimmingly with my little survey which has 36 rows.  Howver, when I tried with Dodoma and then a trivial summary of one variable (carat) and default summary for the diamonds I always had to break out.  It never finished.

anastasia-mbithe:
@rdstern, I have tested the dialog severally using Dodoma and Diamonds datasets, it gets to give an output, only that there is a lag for almost a minute, for the Diamonds data (maybe due to the dataset's size).
image

rdstern:
@anastasia-mbithe working now ok

anastasia-mbithe:
It's ready now, I just had to update the branch, for the sub-dialog functions to be consistent.

rdstern (changes requested):
@anastasia-mbithe This development is exciting.  

a) I occasionally get the developer error when pressing the Format button

image

b) The Format code seems to be missing!  I don't get a change of theme, or a title, when I ask for them.

c) Could you make the 2 labels wider for the French text.  That's the Omit Missing Values and Save Summary.  (Come to think of it, you are making it into a table, so I wonder about Save Table instead?

d) I would very much like to merge this in the version this week.  If you are still waiting for the Missing options, then perhaps disable for now, so it can be merged.

rdstern (approved):
@anastasia-mbithe this is looking great now.  What a teaching tool we are getting towards!  @lloyddewit over to you

Issue #7979, opened 2022-11-24
Continue to improve the output window


rdstern:
The output window now takes html.
This has been merged, so that other dialogues can use this major new feature.

There remain some oddities [reported here](https://github.com/africanmathsinitiative/R-Instat/pull/7853#issuecomment-1324094630). 

I hope they can soon be resolved too.

rachelkg:
Hi @Patowhiz, 
Exciting progress on the output window! 

So far, I have noticed a couple of oddities. 

- For one thing i cannot see my whole graph without maximizing which seems annoying, as you cannot look at the graph in reference to the data. 

- Second, sometimes the graph has a scroll bar and sometimes it does not, though the two graphs are the same size. The first time I made the graph there was no scroll bar, but then a scroll bar appeared on the first graph after I made a second one, though the second one never got one. The summary had a scroll bar right from the first time but didn't need one as was fully visible. 

- Finally, when I maximized the graph, the window kept disappearing when moved my mouse on to it. This happened three times and then I was able to click on it. 

Thanks, 

Rachel 

![2022-11-24_12-57-33](https://user-images.githubusercontent.com/85707390/203863942-b0afe50d-4542-4bf3-9f24-b002f1739e56.png)
![2022-11-24_12-58-10](https://user-images.githubusercontent.com/85707390/203863950-db296e8b-9cc9-4009-97a9-6d9eced773c5.png)

rdstern:
@Patowhiz and @ChrisMarsh82, we have now just finished the workshop for the Caribbean which went well.  We now will continue with a contract with them for the next year.  
It will be good if we can tell them that the next release includes some of the inconveniences they found.  One of these was in the output window - and has been there for some time.  I write without giving an example, but it is something that happens frequently - so is probably one you know.
Sometimes, when doing an analysis that output window is blank and you are scrolled right to the bottom.  You can either go up with the scroll bar, but it is usually quicker to return to the dialogue and repeat the Ok.  Then you are positioned with the output.

Patowhiz:
@rdstern thanks for feedback.
Does that happen with any dialog or specific dialog? An example will help. Does it occur even in the latest version?

rdstern:
@Patowhiz and @N-thony I am going through the list of tasks in your recent e-mail and linking up, where possible, to existing issues concerned with the output window.  I am also using this issue to make more suggestions - you may wish to move some elsewhere.

[...]

PR #8331, opened 2023-05-16
Changes default output startup commands, comments, heights of outputs and refactors R-Instat initialization implementation

### LINKED ISSUE #7854: Limiting heights of outputs in the output viewer
Some outputs are very long. When produced the user has to always scroll back to the top to view the output from the begining.
With the new maximise feature, I think it will be easier to always have the maximise link visible to the users. So that that they don't have to scroll. All they need to do is click on it and see the whole output.

I suggest limiting the heights of the outputs in the output viewer.
The maximum height value of the output in the output window can be part of settings in the options dialog.
This needs further discussions with @rdstern before implementation.

### LINKED ISSUE #8091: Improve the Output Window on startup when commands are not shown
@ChrisMarsh82 I propose that the February release of R-Instat would change to not show the R commands.  

@Patowhiz if you agree with the suggestions below, then perhaps @N-thony could implement, or allocate.

(That would be less off-putting to some of our users, and also give us an extra "ease-of-use" point in the next round of evaluations of R GUIs!)

That introduces (at least) 3 points to be considered first:
**_1. What should the output window show on start-up?_**
@Patowhiz improved the look of the data window.  Here are the current windows - when commands are switched off:

image

The output window doesn't look particularly inviting - or useful to me?

What should we put here.

I wonder about simply something like:
```

            Welcome to R-Instat
Data are shown in the Data View Window.
Results are shown in this Output Window.


```
But I am not imaginative on things like this?

**2. What should we show for the default comments.**

Here is the output window after opening a data frame and using the calculator a bit!

image

The default comments are not useful - particularly when we are not showing the commands.  I still like the Comments "system" whereby a comment can be added with each dialogue.  

One possibility is to turn Comments off - in the Tools menu.  But I quite like the default of having them on, to encourage users to include useful comments.  Either way we could usefully change the default - which the user can do as well.

I have changed it to `Using the dialog:` - so also having a colon instead of the comma - as the dialogue name starts with a capital letter.

3. Should there be an additional simpler way of changing the default settings.  Perhaps even on the toolbar?

It could mirror the Tools menu, or be just within the current session if done that way?  If we don't want an extra icon, then perhaps  we just add a down arrow to the icon for the output window.  The first is then `Output Window` then `R Code` and `Comments`.  I suggest that's a good and neat place to provide this facility?

I quite like the idea that the change on the toolbar is for that session only.  If you want to change it permanently, then use the Tools  menu.

Patowhiz:
Fixes #8091
Fixes #7854
Fixes partly #7979
@africanmathsinitiative/developers this PR has major changes on `frmMain`. I suggested any other changes to be done on `frmMain` to wait until this PR is merged. This will help in preventing conflicts.


rdstern (approved):
@Patowhiz this has done a lot.  I found only one small detail to change, and would like to see it merged as soon as checked by @lloyddewit 

The one very minor suggestion is that, at the start it says:
`# Setting display options (e.g Number of significant digits)`
 That's fine, but it also repeats that, whenever you use the dialogue.
I went in to change the language and it said that again!

The Options Dialog doesn't have the Comments control - perhaps it should?  But as a quick change could the Comment for this dialog be simply #Setting Options?  (Without changing it on startup, from what it is now.

I did try the output window and it seems much better.  I'd like to keep trying it, once it is merged.

That's so minor that I am approving.

Patowhiz:
@rdstern thanks for testing this even though I'm still making major changes on the different bits.
Should we then just have `Setting Options` even during startup?

rdstern:
@Patowhiz I quite liked the current, more detailed, message on Startup, with the brief message - possibly with access to the Comment control, thereafdter?  But, if that takes too much time, I could live with the brief message throughout.

And - assuming @lloyddewit is happy with the code, then I'd be happy to see these changes merged as soon as you are ready, and then another pull request for the further work.

lloyddewit:
@Patowhiz Is this PR ready for peer review or do you still wish to make some changes? thanks

Patowhiz:
@lloyddewit you can go ahead and review this while I keep making more additions

lloyddewit:
> @lloyddewit you can go ahead and review this while I keep making more additions
@Patowhiz I think it will be more efficient if I review when the PR is complete, thanks

rdstern:
@Patowhiz and @lloyddewit I would like to merge what has been done into the next release.  It changes the comments and startup, and therefore is captured in screeen shots for our documents and videos.  So good to have these changes soon.

Patowhiz:
@lloyddewit @rdstern 
Kindly test the following;
1. Switching on and off comments and commands
2. Startup comments and commands.
3. Switching on and off, heights of outputs. Graphs, summaries, tables, HTMLs etc
4. Auto recovered files. In particular notice how the recovered R script is executed and the differences in the options being reset. @lloyddewit it's still not clear to me why we assume some recovery files could be multiple. I noticed we may have to refactor different bits of recovery implementation.

@rdstern as discussed before, there is much more that I'm adding in regards to the outputs;
1. Right click menu that has; copy, save, view in plotly, view in R viewer, maximise, view command etc. Each option will depend on the output type.
2. Making sure that all outputs come in as file. The current exception is the calculator output, so I need to refactor it's code.
3. Each output should now have it's own custom control. These controls should be enhanced to work consistently with the output page they are contained in, this includes, resizing, saving etc.
4. Many more.

Considering the urgency of testing this. I can do the above in a separate PR.

lloyddewit:
@rdstern if you're willing to approve, then I can peer review and we can hopefully merge into the release this week. Thanks

Patowhiz:
@rdstern all previous settings should be retained. Only new users should have `Dialog:` as the default. Could try and test this by resetting the options to defaults? 
After this is merged `Dialog:` will also need it's own translation.

lloyddewit (approved):
@Patowhiz Thank you for this.
When I reviewed, I noticed lines that were commented out rather than deleted. Is this work in progress?
Also, some of the changes to RLink (also frmMain?) looked like refactoring changes rather than changes directly linked to this PR. Is that correct?

Happy to approve and merge.

Issue #8199, opened 2023-03-06
Prepare > Data Frame> Add (Merge) columns translations


Patowhiz:
@africanmathsinitiative/developers 
In the screenshot below, the sections highlighted are not being translated in Portuguese. I suspect it's the same for other languages too.
<img width="340" alt="1" src="https://user-images.githubusercontent.com/21179007/223132581-da1d7551-9636-47e6-8737-f15872c149cf.png">

lloyddewit:
The `Joining By` text field will not translate. This is because the string contains a new line and there is currently no mechanism for translating strings that contain new line characters. This is already documented in issue #8323.
Therefore I will set PR #8333 to close this issue.

![image](https://github.com/africanmathsinitiative/R-Instat/assets/57253949/c15db4b3-3d5b-4b4c-aac7-27910aef90b3)

PR #8333, opened 2023-05-17
Adjusted Layer on Add(Merge) Data Dialog

MeSophie:
Fixes Partially #8199.
The layer has been adjusted according to the Russian translation because it is the longest.
@lloyddewit, @N-thony  I wonder if adding the text to the json file will solve the translation problem. If so will both sentences be added on the same line or will each sentence have its own line? Thank you.


rdstern (commented):
Just checking the changes.  Before it used to say `From Data Frame` and `To Data Frame`.  Now - in English they both just say `Data Frame` which is not so clear.  I assume the changes to the database will revert these?

lloyddewit:
> Just checking the changes. Before it used to say `From Data Frame` and `To Data Frame`. Now - in English they both just say `Data Frame` which is not so clear. I assume the changes to the database will revert these?

@rdstern Yes, you're correct, this requires a database change. I fixed this in PR #8338 .
If you are willing to approve this PR, then we can merge, thanks

lloyddewit (approved):

rdstern (approved):

PR #8341, opened 2023-05-19
Right click modifications

derekagorhom:
Fixes Partly #8239

### DIFF
--- a/instat/ucrDataView.Designer.vb
+++ b/instat/ucrDataView.Designer.vb
@@ -102,13 +102,14 @@ Partial Class ucrDataView
         Me.ToolStripSeparator10 = New System.Windows.Forms.ToolStripSeparator()
         Me.mnuHelp2 = New System.Windows.Forms.ToolStripMenuItem()
         Me.statusColumnMenu = New System.Windows.Forms.ContextMenuStrip(Me.components)
-        Me.deleteDataFrame = New System.Windows.Forms.ToolStripMenuItem()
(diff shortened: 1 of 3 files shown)


rdstern (commented):
@derekagorhom that looks fine as far as it goes. Well done.  It now just remains to call the same right-click, from within the data frame with the Edit Cell and Delete Cell(s) now enabled..

derekagorhom:
@rdstern I have made the changes... but since the Edit cell is still a work in Progress, I disabled it till we were done with PR #8264 
after which it can be enabled.

rdstern (commented):
@derekagorhom looks pretty good.
a) It still says Edit Cell(s) when I am at the top of the column.
b) I think @lloyddewit has a problem with the Delete Column entry, that changes into Delete Columns when you select multiple coliumns.  That's pretty clever, but I think it makes the translation into different languages more difficult.  So could you please simplfy that to Delete Column(s).  So it doesn't change dynamically when you select multiple columns.

Then I would really like to merge this soon, because we are using it in many of our training documents.  So many thanks for doing this so quickly.

derekagorhom:
@rdstern I have fixed the issues with the Edit cell and also the name change for the delete columns. 
Thanks

rdstern (approved):
@derekagorhom great.  @lloyddewit over to you.
Very good if this could be in the forthcoming release.

lloyddewit (approved):

PR #8343, opened 2023-05-22
Fixes Bug When using One Variable in View/Delete Labels Dialog.

### LINKED ISSUE #8336: Prepare > Check Data > View/Delete Labels doesn't work when there is just one variable
@N-thony this should be very simple to get anyone to fix.  It happens with any datset I think?

image

The problem is only with a single variable to process - which is why I never noticed before.  I always used it with more than 1!

MeSophie:
Fixes #8336.

@rdstern I fixes the bug when using one variable on View/Delete Labels Dialog. 

@lloyddewit, @N-thony please have a look . Thank you.


rdstern (approved):
Looks good.  @lloyddewit over to you

N-thony (approved):

lloyddewit (approved):

PR #8344, opened 2023-05-22
Ensured that text containing apostrophes (single quotes) is correctly translated

lloyddewit:
Previously, the number of apostrophes in a text string doubled each time it was translated. See https://github.com/africanmathsinitiative/R-Instat/pull/8329 for more details.


rdstern (approved):
@lloyddewit that's great.  I tried a range of things, and it seems you have really nailed it!

PR #8356, opened 2023-05-25
Ensured that tab headings in Log/Script window are not translated

lloyddewit:
Includes database changes:

image

Deleted legacy rows containing ReplaceWithDynamicTranslation

image


lloyddewit:
@rdstern This PR fixes the bug we found during our meeting this morning. Please could you test?
Thanks

rdstern (dismissed):
Looks fine now.

rdstern:
@lloyddewit Here you see the problem with the released version - so not with the version after your fix in this pull request:  Notice the Replace with Dynamic Translation, right at the top and in the dialog - so including the OK and Script buttons.

image

Then (still on the released version - so not yet with the fix here)

image

You can see the width of the dialogs is affected, because of the wide text in the base controls.

OK, now with the fixed version:

1) I started in English and all seems ok.  I tried the 2 dialogs above and they are now fine.
2) I went to French and all ok still.  Then back to English and still all fine.
3) Then to Russian and not fine:  Here is a screen shot.  Notice the same problem at the top (where there should be the version number, and right at the bottom where there should be the name of the data frame.  Also the OK and Script keys again.  

image


Now I have switched back into English and some of the Russian problem has persisted:

image

I then moved to French, but the problem seems to persist.  This seems to be a "Special Operation" by Russian on R-Instat.  

No, it is not as bad as I thought.  All the other dialogs seem to be ok.  It is just the boxplot dialog that is still odd.  I assume if I could hit Reset this might also be fine, but (of course) the OK button is so wide that there is no reset button visible!

lloyddewit:
@rdstern @ChrisMarsh82 
First an apology. As the last step before the release, I rebuilt the translation database from scratch (this is the only way I know to ensure that all the controls added/updated since the previous release will be translated). Over the last weeks, I have been automating this process as much as possible and I thought I had reached the point where this was low risk. I was obviously wrong. I did a sanity check of the new version but I obviously did not test nearly deep enough.
Lessons learned:
- the process to rebuild the translation database still needs improving/fixing
- the PR updating the database should be tested/approved before merging, test should include all the main windows, multiple dialogs and multiple languages

In the short term, I'll fix the bugs reported above.
I assume that we'll also have to make a new release - sorry for that.

ChrisMarsh82:
That's fine. Let me know when you are ready and I will set everything going.

lloyddewit:
@rdstern I think I fixed the issues above. But I also spotted some issues with the OK, Reset and Help buttons on the dialogs. I'll fix these after my meetings today. So you don't need to test yet.

**Notes** 

Removed lines from translation table:

image

Ensured version number does not translate:

image

lloyddewit:
@rdstern I think I fixed the translation issues above. Please could you test?
thanks

rdstern (approved):
Excellent @lloyddewit .  I have been switching languages like crazy and it seems fine now.

Patowhiz:
> @rdstern @ChrisMarsh82 First an apology. As the last step before the release, I rebuilt the translation database from scratch (this is the only way I know to ensure that all the controls added/updated since the previous release will be translated). Over the last weeks, I have been automating this process as much as possible and I thought I had reached the point where this was low risk. I was obviously wrong. I did a sanity check of the new version but I obviously did not test nearly deep enough. Lessons learned:
> 
> * the process to rebuild the translation database still needs improving/fixing
> * the PR updating the database should be tested/approved before merging, test should include all the main windows, multiple dialogs and multiple languages
> 
> In the short term, I'll fix the bugs reported above. I assume that we'll also have to make a new release - sorry for that.

@lloyddewit I call a single function to rebuild the entire translations database in Climsoft. I do this through a function I added in the translations library. I understand R-Instat custom controls may not be fully supported with the function. Would be good to discuss this in our tomorrow catchup. I'm keen on your automation as well.

Issue #4108, opened 2017-10-27
Updated toolbar


rdstern:
I have been using it in the workshop.  It is working well.
Our use has been with climatic data, so quite long datasets, but not wide.  And with multiple data frames.
1. I propose we add the icon for the **Data frame metadata**.
2. I realise this just leaves the script window with no icon.  But that is different anyway, because it appears automatically when you use the script button on a dialogue.
3. Could there be something a bit wider to separate the sets of icons.  It is currently a very thin vertical line and is not so visible.  But also it uses much less space than in other software, e.g. acrobat.  I have  watched people clicking the recall last 10 dialogues icon and they are very close to the data-frame icon.

shadrackkibet:
I think 1 and 2 are done. For 3 I am not sure I clearly understand perhaps @rdstern can explain further, looks like it is a user experience issue. The current icons are sufficiently spaced so perhaps this is no longer an issue?

rdstern:
@Patowhiz I am progressively happier with the new toolbar.

I wonder if either of the following 2 items are reasonably easy to implement?  Ideally both!!
a) Could the Open From Library have a down-arrow, and remember the previous library items that were used?  This would parallel the facility for ordinary open files?  I realise this may be different for the R packages and the R-Instat options.  If it is only easy for one of those cases, I would still take that.
b) There is already a down arrow on the Log/Script Window.  Could there be an option there to Show R Commands.  This would be a quick way of the option that is in the Tools > Options > Output Window?

Patowhiz:
@rdstern you mean on the dialog itself? or the main form tool bar? or on the startup data view window?

rdstern:
@N-thony here is the current toolbar in Version 0.8.1.

![image](https://github.com/user-attachments/assets/5f3c75f4-65d6-4705-8c4d-807f4f9b9804)

I am liking the fact it looks neat.  The one missing down arrow is the title of this issue.  Could we make it easy to use library data sets again.  Maybe I would find that very convenient and it would help the appearance of the toolbar too.

I don't know whether it is easy though?

lilyclements:
@Ag-Derek this should be very simple - if you know the main form. I think you do. Do you know how to add that arrow back on for this? We want an arrow between the folder and the floppy disk so that the folder has an arrow again. (More or less directly under the first "r" in "Prepare")

Ag-Derek:
@anastasia-mbithe since you wanted a task to do....are you comfortable with working on this

[...]

PR #8359, opened 2023-05-30
Improved usability of Toolbar

derekagorhom:
Fixes partly #4108

This Fixes items A to D except adding the `Show R commands` to the `log/Script` window drop-down. 
I keep getting an error when adding that option to the log/script drop-down so i will fix that one separately. 

This PR is ready for review
@rdstern , @N-thony can you test/review. 

Thanks

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -1060,11 +1060,11 @@ Public Class frmMain
         dlgAppend.ShowDialog()
     End Sub
 
-    Private Sub mnuFileSaveAsOutputAs_Click(sender As Object, e As EventArgs) Handles mnuFileSaveAsOutputAs.Click, mnuSaveOutputTab.Click
+    Private Sub mnuFileSaveAsOutputAs_Click(sender As Object, e As EventArgs) Handles mnuFileSaveAsOutputAs.Click, mnuSaveOutput.Click
         ucrOutput.UcrOutputPages.SaveTab()
     End Sub
 
-    Private Sub mnuFileSaveAsLogAs_Click(sender As Object, e As EventArgs) Handles mnuFileSaveAsLogAs.Click
+    Private Sub mnuFileSaveAsLogAs_Click(sender As Object, e As EventArgs) Handles mnuFileSaveAsLogAs.Click, mnuSaveLog.Click
         ucrScriptWindow.SaveScript(True)
     End Sub
 
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -670,7 +670,7 @@ Public Class frmMain
         End If
     End Sub
 
-    Private Sub mnuTbSave_Click(sender As Object, e As EventArgs) Handles mnuTbSave.ButtonClick
+    Private Sub mnuTbSave_Click(sender As Object, e As EventArgs) Handles mnuTbSave.ButtonClick, mnuSaveData.Click
         mnuFileSave_Click(sender, e)
     End Sub
 
@@ -1173,7 +1173,7 @@ Public Class frmMain
         dlgImportOpenRefine.ShowDialog()
     End Sub
 
-    Private Sub mnuFileSaveAsDataAs_Click(sender As Object, e As EventArgs) Handles mnuFileSaveAsDataAs.Click, mnuSaveData.Click
+    Private Sub mnuFileSaveAsDataAs_Click(sender As Object, e As EventArgs) Handles mnuFileSaveAsDataAs.Click
         dlgSaveAs.ShowDialog()
     End Sub
 
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -610,13 +610,7 @@ Public Class frmMain
         UpdateLayout()
     End Sub
 
-    Private Sub mnuLogScript_Click(sender As Object, e As EventArgs) Handles mnuLogScript.Click, mnuViewLogScript.Click
-        mnuViewLogScript.Checked = Not mnuViewLogScript.Checked
-        mnuLogScript.Checked = mnuViewLogScript.Checked
-        UpdateLayout()
-    End Sub
-
-    Private Sub mnuTbLogScript_Click(sender As Object, e As EventArgs) Handles mnuTbLogScript.ButtonClick
+    Private Sub mnuLogScript_Click(sender As Object, e As EventArgs) Handles mnuLogScript.Click, mnuViewLogScript.Click, mnuTbLogScript.ButtonClick
         mnuViewLogScript.Checked = Not mnuViewLogScript.Checked
         mnuLogScript.Checked = mnuViewLogScript.Checked
         UpdateLayout()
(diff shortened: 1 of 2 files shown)


rdstern:
@derekagorhom that is looking good.
a) Could you also make the Down arrow on the first icon, `Import`, as wide as the othgers.
b) @Patowhiz (if @derekagorhom can not answer) would it be easy to have a down arrow on the open from Library with the recent files from there as well?
c) In the new Save down arrow delete the Tab on the output - so it just says `Save Output`.  And add `Save Log`.  
d) For later, because there is another issue on the saving of the Script and Output tabs.  I wonder, once there are more tabs produced, of either, whether they could be listed here.  (And possibly even disabled if they have not changed since the last save?)

rdstern:
@derekagorhom this is looking very promising now.
a) When I click on the Save button it is a Save, rather than a Save As.  (The Save option asks first time, but then overwrites the name on future occasions, without asking.) 
But when I click on down arrow and the Save Data option - which should be the same, it isn't.  It is a Save As.  I think all 4 options are Save As, rather than Save?  I am quite keen that they each become a `Save `button, rather than `Save As`.
b) I am discussing with @lloyddewit and @ChrisMarsh82 today, whether an * can be added to each option, in their respective windows, when there is new material to Save.  (That seems to be what RStudio does.)  If that is agreed, then it would be good to have the same in the 4 options in your Save.
c) Now a real detail - here is the current toolbar:

image

I notice a vertical line occasionally and even 2 of them between the new Save and the last 10 dialogs.  I quite like the one vertical line and wonder if it could be after each entry?

derekagorhom:
@rdstern I think for item c) the vertical line can be made very similar to the one on the sub menus 
image
That way the menu items are separated into groups, i think this makes it more presentable

rdstern:
@derekagorhom I like that idea a lot.

image

a) So, the first 3 icons are all loading and saving files.  So we don't have any vertical lines there.
b) We keep the line after your new Save - presumably just one line?
c) We keep the line after the last 10 dialogues.
d) Then the next 2 icons are the grid and output windows.  No line between them, but one after those 2.
e) Then the graph icon is on its own and there is a line after that.  
f) Then the R and the i open more windows and thew curly arrow closes them.  So let's think of them as a group and so just keep the vertical line there, after the curly arrow.
g) Then add a vertical line after the ? for help.  And possibly also one right at the end - after the language icon?

rdstern:
@lloyddewit you said adding the asterisk in the script (and log) tabs was easy to do.  I hope it might be equally easy to add in the new drop down in the toolbar?  If, so, then can you add this @derekagorhom?  That has (like the File > Save As) a special entry for saving the Log file, and is getting towards my minimum that we implement quickly, to save the important items.

lloyddewit:
> @lloyddewit you said adding the asterisk in the script (and log) tabs was easy to do. I hope it might be equally easy to add in the new drop down in the toolbar? If, so, then can you add this @derekagorhom? That has (like the File > Save As) a special entry for saving the Log file, and is getting towards my minimum that we implement quickly, to save the important items.

@rdstern This requires changes to the script window first (it has to catch the event when the script tab is changed, and then set a flag). So I suggest that I make the change above in a separate PR (when I add the asterix to the script tab). I don't think it make sense for @derekagorhom to try an implement an asterix in the toolbar in this PR.
I hope that's OK.

derekagorhom:
@rdstern I made the layout changes you requested and i was able to make the save data option save the same way the save icon operates. I could not change the other 3 option since they follow the existing `log/script/output` saves in R-instat

rdstern (dismissed):
@derekagorhom I am liking this a lot.  I am happy to see this merged now, and it will improve (or can be improved) once there are small changes in the log/script and the output windows.

So over to you @lloyddewit for this, so far.

instat/frmMain.vb
```diff
@@ -610,9 +610,15 @@ Public Class frmMain
...
 
-    Private Sub ScriptToolStripMenuItem_Click(sender As Object, e As EventArgs) Handles mnuTbLogScript.Click, mnuViewLogScript.Click
+    Private Sub mnuLogScript_Click(sender As Object, e As EventArgs) Handles mnuLogScript.Click, mnuViewLogScript.Click
         mnuViewLogScript.Checked = Not mnuViewLogScript.Checked
-        mnuTbLogScript.Checked = mnuViewLogScript.Checked
+        mnuLogScript.Checked = mnuViewLogScript.Checked
+        UpdateLayout()
+    End Sub
+
+    Private Sub mnuTbLogScript_Click(sender As Object, e As EventArgs) Handles mnuTbLogScript.ButtonClick
```
lloyddewit, line 619: Is this sub identical to `mnuLogScript_Click()`? If so, could we make `mnuLogScript_Click()` handle the `mnuTbLogScript.ButtonClick` event and delete this sub?

instat/frmMain.vb
```diff
@@ -587,7 +587,7 @@ Public Class frmMain
         End If
         mnuTbDataView.Checked = mnuViewDataView.Checked
         mnuTbOutput.Checked = mnuViewOutput.Checked
-        mnuTbLogScript.Checked = mnuViewLogScript.Checked
```
lloyddewit, line 590: where is `mnuTbLogScript.Checked` set?

instat/frmMain.vb
```diff
@@ -443,7 +443,7 @@ Public Class frmMain
 
         mnuTbDataView.Checked = True
         mnuTbOutput.Checked = True
-        mnuTbLogScript.Checked = False
```
lloyddewit, line 446: where is `mnuTbLogScript.Checked` set?

instat/frmMain.vb
```diff
@@ -2401,20 +2407,20 @@ Public Class frmMain
...
+    '        ucrScriptWindow.PasteText()
+    '    End If
+    'End Sub
 
-    Private Sub mnuPasteSpecial_Click(sender As Object, e As EventArgs) Handles mnuSubTbPasteSpecial.Click, mnuPasteSpecial.Click
-        dlgPasteNewColumns.ShowDialog()
-    End Sub
+    'Private Sub mnuPasteSpecial_Click(sender As Object, e As EventArgs) Handles mnuPasteSpecial.Click
+    '    dlgPasteNewColumns.ShowDialog()
+    'End Sub
```
lloyddewit, line 2423: delete rather than comment out?

lloyddewit (changes requested):

instat/frmMain.vb
derekagorhom: @lloyddewit  `mnuTbLogScript` was replaced with  `mnuLogScript` since `.checked` is not a member of `ToolStripSplitButton`

lloyddewit: @derekagorhom `mnuTbLogScript` still exists, it is referenced in `frmMain.Designer.vb` and `frmMain.vb`. If it's been replaced, then should we delete it completely?

derekagorhom: @lloyddewit 
I dont think so, `mnuTbLogScript`  is the design name for the toolbar icon of `log/script` while  `mnuLogScript` is the design name for the dropdown button of the `log/script`. 
something similar was done for the toolbar icon for the `column metadata/dataframe metadata`

lloyddewit: @derekagorhom Ok, thanks for investigating, happy to resolve this comment

lloyddewit (approved):

lloyddewit:
@rdstern there are just some small changes since your last approval. If you can retest/approve, then we can merge, thanks

rdstern (approved):
@lloyd and @derekagorhom  - still looks fine.  Thanks

Issue #8351, opened 2023-05-24
Incorrect naming of Frames in the panel tab of the Theme subdialogue


NanyanziAlice:
In the Theme dialogue -> panel tab, both frames are named "Major Grid Lines". On testing the options under both, the frame to the right should be named "Minor Grid Lines " to differentiate them. (Version 0.7.9)

Changing options in the left side frame as the figure below:
![image](https://github.com/africanmathsinitiative/R-Instat/assets/19998927/65340157-cf6b-4e1d-adb6-b3b85a276f19)

The resulting figure with changes made to the major grids is as shown below:
![image](https://github.com/africanmathsinitiative/R-Instat/assets/19998927/7de1d7cf-953c-4bc1-b066-9637b1ad262b)

On changing options in the right side frame, no changes occur to the major grids.

rdstern:
@NanyanziAlice well spotted.  I was worried that the "system" had regressed, because earlier man of the labels were the same!
But this seems the only one.
@N-thony who should make this correction?

PR #8364, opened 2023-06-01
Corrects the minor grid lines options label in the panel tab of the theme sub-dialog

MeSophie:
Fixes Partially #8351 
@rdstern I correct the name of frame but I'm still investigating on the reason why the `"Minor Grid Lines"` is not working because the code in to script looks nice.

image


rdstern:
@N-thony this seems odd.  Is this now back to you to check, or who might investigate further?  Does it await the sprint?

N-thony:
@MeSophie let me know when you are ready then we can have a look together.

rdstern:
@N-thony and @MeSophie how is this going?

N-thony:
> @N-thony and @MeSophie how is this going?

@MeSophie can you report what we found last when trying to include the `panel.minor.grid` in the map?

MeSophie:
@rdstern After investigation, we have discovered that despite the fact that the code is correct, the `panel.minor.grid` is still not visible and this only with the map dialogue and  at least since version 0.7.6 .

Patowhiz:
@MeSophie what is the status of this PR?

MeSophie:
> @MeSophie what is the status of this PR?

@Patowhiz It still partially solves issue #8351. Because  After investigation, we have discovered that there is no change when we modify the values for the `panel.minor.grid` in the map. So I don't know if @rdstern will  agree that despite that this pr can be merge.

Patowhiz:
@MeSophie I'm not sure if I understand you correctly, are you saying that the script produced by the minor grid lines options don't cause any changes to the final output? Have you tested the scripts in R Studio as well?

MeSophie:
> @MeSophie I'm not sure if I understand you correctly, are you saying that the script produced by the minor grid lines options don't cause any changes to the final output? Have you tested the scripts in R Studio as well?

Exactly @Patowhiz I Also asked help to @N-thony the result was the same.

Patowhiz:
@MeSophie from your research, is this an issue in the R package itself or it's an issue in R-Instat. It's important you do these tests in R Studio as well. And also check online if others have encountered the same issue. We could then document why we have disabled the feature with concrete references.

MeSophie:
> @MeSophie from your research, is this an issue in the R package itself or it's an issue in R-Instat. It's important you do these tests in R Studio as well. And also check online if others have encountered the same issue. We could then document why we have disabled the feature with concrete references.

@Patowhiz  Maybe the problem is in the package. We searched in R-Studio and online. Minor grid lines works everywhere else, except in the mapping dialog.

rdstern (approved):
I'm just approving!

Issue #8314, opened 2023-05-08
Fix bug in the climatic > Summary dialogue


rdstern:
@Vitalis95 this is in this option of the dialogue:

![image](https://user-images.githubusercontent.com/11226469/236847368-d9740029-f303-4362-9fec-c9ce40e47e98.png)

a) This code generates the following commands:

```
# Code generated by the dialog

doy_366 <- data_book$get_columns_from_data(data_name="dodoma", col_names="doy_366")
doy_366 <- data_book$get_columns_from_data(data_name="dodoma", col_names="doy_366")
day_filter <- instat_calculation$new(type="filter", function_exp="doy_366 >= 1 & doy_366 <= 366", calculated_from=calc_from_convert(x=list(dodoma="doy_366")))
data_book$calculate_summary(columns_to_summarise=c("tmax","tmin"), data_name="dodoma", factors=c("doy_366"), na.rm=TRUE, additional_filter=day_filter, na_type=c(), j=1, summaries=c("summary_count_non_missing", "summary_mean", "summary_min", "summary_max"), silent=TRUE)
rm(list=c("day_filter", "doy_366"))

```
Well, that's after my correction.  
1) I don't understand why the first line is repeated.
2) In the main command you see factors = c("doy_366").  That's after I corrected in the script.  Currently it puts "date" there and that is not what is wanted.

That's the bug.

b) In the main options there is a checkbox (see below) of Add Date Column.

![image](https://user-images.githubusercontent.com/11226469/236849186-5ffd525d-7a04-46ef-bb36-fa0e975c2cea.png)

Please also include that to this option.  The code is not quite trivial, because the year may be shifted and you may have less than the whole year.  In the simplest case (starting from January), just add the year 2000 to the doy.  If shifted by more than 2 months, then start in 1999.  (I am trying to ensure the doy includes a leap year!  I hope it will be easy.  If not, then perhaps check with @jkmusyoka for the R code?

N-thony:
> 

@Vitalis95 @rdstern in a) I don't see the need of the two first lines in the rest of the code.

Vitalis95:
> > 
> 
> @Vitalis95 @rdstern in a) I don't see the need of the two first lines in the rest of the code.

@N-thony , I removed one of it, Should I remove both?

N-thony:
> > > 
> > 
> > 
> > @Vitalis95 @rdstern in a) I don't see the need of the two first lines in the rest of the code.
> 
[...]

PR #8365, opened 2023-06-02
Fixed Minor bug in Climatic summary dlg

### LINKED ISSUE #8362: New bug has appeared in the Climatic > Prepare > Climatic Summaries dialog?
@lloyddewit it is in the Day Range control and is curious.  I am concerned that it is ok in the previous versions, and I just found it by chance.  And it is odd.  I hjope there is an obvious reason and not an indication that similar oddities may be elsewhere?

image

What is also odd (comforting or not?) is that we have the same control in various other later dialogs in the same menu and there it is fine.  For example:

image

When I run the dialog it works, because it is still a valid condition.  Here is the code:

```
# Diálogo: Resumo Climático

day_filter <- instat_calculation$new(type="filter", function_exp="1 & doy <= 366", calculated_from=calc_from_convert(x=list(Mozambique_3_stations="doy")))
data_book$calculate_summary(columns_to_summarise=c("TMPMAX","TMPMIN"), data_name="Mozambique_3_stations", factors=c("Station_Name", "year"), additional_filter=day_filter, j=1, summaries=c("summary_count_non_missing", "summary_mean", "summary_min", "summary_max"), silent=TRUE)
rm(day_filter)


```

Vitalis95:
Fixes #8362 
@rdstern @lloyddewit , seems to be fine now.


Vitalis95:
> @Vitalis95 did you test if item a) in issue #8314 is still fixed here too?

@N-thony , Yes it is fine

rdstern (approved):
@Vitalis95 and @lloyddewit many thanks again.  From my point of view it looks fine again.

PR #8373, opened 2023-06-12
Add controls to histogram dialog

### LINKED ISSUE #8335: Add controls to hisyogram plots just as the bar charts
@derekagorhom and @COLIEWO may wish to check this idea quickly.

In the bar chart dialogue you have now 2 useful controls under the Factor variable, namely Reoder and Position:

image

The idea of ggplot2 grammar is that logically the same options should work for the Histogram dialog - and (I think) for all 3 options of the top radio buttons at the top.

image

Could you add them please.  I hope both the VB part and the changes in code should be simple, because they should be the same as for the bar chart dialogue.

This may all make quite a nice teaching point - when we come to discuss the wonders of factors and ggplot!

COLIEWO:
Fixes #8335 
I have added the Reverse checkbox and the Position dropdown controls to the Histogram dialog.
@rdstern @N-thony This PR is ready for review.

### DIFF
--- a/instat/dlgHistogram.designer.vb
+++ b/instat/dlgHistogram.designer.vb
@@ -62,6 +62,9 @@ Partial Class dlgHistogram
         Me.ucrHistogramSelector = New instat.ucrSelectorByDataFrameAddRemove()
         Me.ucrBase = New instat.ucrButtons()
         Me.ucrPnlOptions = New instat.UcrPanel()
+        Me.lblPosition = New System.Windows.Forms.Label()
+        Me.ucrInputHistPositions = New instat.ucrInputComboBox()
+        Me.UcrChkReverse = New instat.ucrCheck()
         Me.contextMenuStripOptions.SuspendLayout()
         Me.SuspendLayout()
         '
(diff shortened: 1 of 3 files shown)


rdstern:
This seems to be coming on.  I notice the extra options are always there for the histogram.  The default should still be as before, i.e. without those options.
I assume the same will then apply to the density plots ans the frequency polygons, and possibly also to the ridge plots - at least reverse will apply.

N-thony:
Good, @EstherNjeriLiberatta you can start by taking over this PR and continue.

rdstern:
I am starting this reply without the internet, so am a bit handicapped as I can't see your latest branch.  But I think the changes we need should be simplewr than has been done so far.

You are adding 2 controls I think, namely Reorder and Position?  
a) We need the `Reorder` control and it is just the order of the levels of the factor.  So that is equally sensible to use when your data plotted are bar charts or histograms/density plots.  It is simply the order in which the factor levels are included.  For example if the levels of a factor are January to December, then the Reverse option goes from December to January for the ridge plots.  That's quite useful.
The way to include that is identical to the way in the bar chart.  It just includes a forecats function in the code. 
b) But the `Position` control, in the bar charts describes how the initial factor and the second factor are displayed.  There is only one factor variable in the histogram dialog, so that makes no sense.  So delete that control and the associated code.

N-thony:
> I am starting this reply without the internet, so am a bit handicapped as I can't see your latest branch. But I think the changes we need should be simplewr than has been done so far.
> 
> You are adding 2 controls I think, namely Reorder and Position? a) We need the `Reorder` control and it is just the order of the levels of the factor. So that is equally sensible to use when your data plotted are bar charts or histograms/density plots. It is simply the order in which the factor levels are included. For example if the levels of a factor are January to December, then the Reverse option goes from December to January for the ridge plots. That's quite useful. The way to include that is identical to the way in the bar chart. It just includes a forecats function in the code. b) But the `Position` control, in the bar charts describes how the initial factor and the second factor are displayed. There is only one factor variable in the histogram dialog, so that makes no sense. So delete that control and the associated code.

@rdstern  I think @EstherNjeriLiberatta took over this PR in PR #8395, and the changes should be done there.

rdstern:
@N-thony so should this be closed?

N-thony:
> @N-thony so should this be closed?

Yes

Issue #7980, opened 2022-11-24
Adapt the Describe > One Variable > Frequencies as no output now from the stem option


rdstern:
@N-thony I would like you to manage who does what on the set of issues for the new output window.
This one looks messy and needs a bit of detective work, to sort it out, that may be additional and unrelated.  I wonder if @Vitalis95 might be able to undertake it?  I expect others to be simpler and you may be able (in parallel) to do some yourself, in just a few minutes, or to allocate them.

This should be an easy one, but it may turn out to be a bit tricky.  It didn't work at all well for me just now.  I suspect what I have been finding may not be related to the main issue, but you may wish to correct it (them) first.

So first explore the current dialogue.
Here is what I did and the default task.  But read all below, because I then had (possibly unrelated) problems
a) Opened the usual survey data.
b) Tried the simplest option, namely all variables with the Table option, output to `Text (Output Window)`
c) Then tried the alternative HTML (Browser).
d) That's the simple task, namely the Browser option works fine and goes to a browser.  html can now go to the output window.  
e) So that's the main task.  And the change in code should be the same as in the Prepare > Check Data > View/Delete Labels that has already been done.  
f) So the html should not say (Browser) any more, and the Text should not say Output Window, because everything goes to the output window. 
g) I then tried a graph - with all the variables.  And it went straight to the R-viewer - which needs to change, because we don't use it often.  It was also an awful graph, but I suspect that should be left for now.
h) Then I tried stem and leaf for the yields - which is different and which didn't work.  I think there is some terrible programming here.
i) Then I restarted R-Instat and tried the first Table option.  Then I did stem and leaf again and it works fine.
j) But then when I return to the Table option for all variables it gives a totally different set of outputs!
k) I also tried the Both option and it now wanted to save (by default) both the graph and the table.  And it did save the table into another data frame - but not very well - though I asked it not to!

This may take too long - in which case leave it for now, but I hope not!  Or you may find you can quickly fix that Table and Stem parts - in which case disable the graph and Both options for now!

rdstern:
@N-thony I think the above may now already be fixed.  I'll perhaps check, but alternatively it is a good dialog for one of the team involved in maths camps to check.  

But the real problem is that I now don't get anything in the output window from the Stem and Leaf plot.  Please could a developer look at that.  Maybe @Fidel365?

[...]

PR #8375, opened 2023-06-13
Ensured that Stem and Leaf plots work correctly in Describe > One Variable > Frequencies dialog

Fidel365:
fixes partially issue #7980 
For part two of the issue, an error occurs when we replace the single receiver with a multiple receiver ; 'x must be a numeric value', this occurs when we use 2 or more variables in the multiple receiver.
@rdstern , this PR is ready for review


rdstern (dismissed):
@Fidel365 this now seems to work - at least with a single variable.  I am approving and would be happy for this to be merged as it is.  

Maybe you can also leave the issue open, in case it is not too difficult to include the multiple receiver.  It will need someone to change the R-code.  Could you ask @EstherNjeriLiberatta first to see if she can do it easily.  Emphasise that she should not spend more than 15 minutes on it - it will depend a) on her R skills and b) If it really is easy.  Then I hope @lilyclements can check her suggestions.  If it is difficult, then we leave it.  But I think it may be easy, and also instructive.

lilyclements:
@Fidel365 let me know when this is ready for me to check and we can go through together

Fidel365:
@rdstern , this PR is ready for  a re-review
But, we have to filter variables for the stem and leaf to be numeric and not for the Table and Graph; Hence when we set a datatype, we do not have a way of removing it; If we filter for numerics in stem and leaf the change applies also to Table and Graph, there should be an issue addressing this; i'll create one

rdstern (commented):
@fidel, this is looking pretty good now, and I really like the way you are explaining some of the limitations you have found.
a) In the old version - when it was a single receiver - it did just what you wanted, namely only the numeric variables are available below for stem and leaf, but not for the other options.  Why is that so difficult now?

image

And here is the output in Version 0.7.6 - so from a year ago.

image

Could we return and have that same, nice clean, output now - ideally still for multiple (numeric variables, but without the line numbers and the quotation marks?  Maybe check with @Patowhiz.

However, I note that the 0.7.6 output does not say which variable it is summarising.  It would be good to add that, particularly now that multiple variables are bing processed.
.

Fidel365:
@rdstern , the issue with the quotes in the stem and leaf has been fixed, but not for the line numbers, @lilyclements says it's not possible

Fidel365:
@rdstern , this PR is ready for a re-review

rdstern (dismissed):
@Fidel365 and @lilyclements I am really sorry I suggested this.

It is important we have a working stem and leaf, but the earleir one variable at a time was ok.  I just assumed it would be a minor change to be able to have a multiple receiver there!

How wrong I have been!  There are now 2 problems, 
a) We can't restrict to just numeric - which we had with one variable
b) The output is acceptable, but messier than with one variable.

image

I am approving, because it is working ok now.  But, I am very soory to say that changing back to the one variable is a lesser of evils.  

I hope that it doesn't take @lloyddewit too much time, so this can be merged for now.  Then my suggestion is to revert to single variable, neat output, and just numeric!

lloyddewit (changes requested):
Thanks, looks good. Only 2 small suggestions.

lilyclements:
@rdstern there's a bit of an issue with this function, and that's that it prints the item rather than stores it. It wasn't picked up on before, but this has always been an issue. @Patowhiz looked into it to find a way to "store" it, and unfortunately that's why the output is not as "clean" now. By storing it, we are enabling it to be viewed later on under bits like "View Objects".
So we have a choice: Would we rather have it as an object, and so viewable later on and stored within R-Instat
Or would we rather just have it print now, and un-saveable?

Maybe to help show, then run this:

```
# if we save it as an object (here "hello") it still runs the code! This is not like how R usually works!
hello <- stem(x = 1:100)

#  The decimal point is 1 digit(s) to the right of the |
#
#  0 | 123456789
#   1 | 0123456789
#   2 | 0123456789
#   3 | 0123456789
#   4 | 0123456789
#   5 | 0123456789
#   6 | 0123456789
#   7 | 0123456789
#   8 | 0123456789
#   9 | 0123456789
#  10 | 0
#
```
If I then call my object, I get no output, just `NULL`:
```
hello
NULL
```

rdstern:
@lilyclements and @Fidel365 my main interest now is to finish this task using the minimum of your time.  So we can exist with the current printout - so it is stored.

Should we return to the single receiver so we can restrict the variables to numeric?

lilyclements:
@rdstern great - @Fidel365 is working on it now. Thanks :)

Fidel365:
@rdstern , this is ready for review
The multi receiver for stem and leaf now only picks numerics

rdstern (approved):
@Fidel365 well done.  @lloyddewit over to you

lloyddewit (approved):

Issue #8244, opened 2023-03-28
More on the SPI dialogue


WezzieBanda:
I am getting a bug in the SPI dialogue. Initially, I was using Malawi data which has 4 stations, I then used Nigeria-Samaru 56 data from the library and I have noted that  I am getting the same bug. 

I went to: Cimatic>Prepare>SPI

![IMG_20230328_140957](https://user-images.githubusercontent.com/106272881/228243175-76ea1c2b-5db4-4a22-bf7b-18ee7354ebcc.jpg)
![IMG_20230328_141014](https://user-images.githubusercontent.com/106272881/228243266-f5705a6a-4df7-42db-be17-f216cc2a3d5b.jpg)
 

 - R-Instat Version [ 0.7.9]

rdstern:
@Vitalis95 or an alternative person @N-thony I second the problem with this dialog.  However this seems to be caused by the new code for the output window, so it should be easy to fix?

I tried with the single station, wichita, provided in the SPEI package.  

![image](https://github.com/africanmathsinitiative/R-Instat/assets/11226469/a785c911-e142-4d15-a8c7-8532875cf723)

In the old Version, 0.7.6 it works and here is the R-code:
```
# Dialog: SPI/SPEI

wichita <- data_book$get_data_frame(data_name="wichita", use_current_filter=FALSE)
data_ts <- spei_input(year="YEAR", month="MONTH", element="PRCP", data=wichita)
spi_mod1 <- SPEI::spi(data=data_ts, scale=1, kernel=list(type='gaussian', shift=0), na.rm=TRUE)
data_book$add_model(model_name="spi_mod1", model=spi_mod1, data_name="wichita")
spi <- spei_output(x=data_book$get_models(data_name="wichita", model_name="spi_mod1"), year="YEAR", month="MONTH", data=wichita)
data_book$add_columns_to_data(data_name="wichita", col_name="spi", col_data=spi, before=FALSE)

rm(list=c("spi", "spi_mod1", "data_ts", "wichita")) 
```
And this works fine, 

In the new version it gives the error above and the R code is now:

```
# Dialog: SPI/SPEI

wichita <- data_book$get_data_frame(data_name="wichita", use_current_filter=FALSE)
data_ts <- spei_input(year="YEAR", month="MONTH", element="PRCP", data=wichita)
spi_mod3 <- SPEI::spi(data=data_ts, scale=1, kernel=list(type='gaussian', shift=0), na.rm=TRUE)
[...]

PR #8376, opened 2023-06-14
Fixed a bug in the SPI dialog

Vitalis95:
Fixes (partially) #8244 
@rdstern , @N-thony @lloyddewit  , it works now well including plot for single station. I will check with @lilyclements  for  the multiple stations, I realized it was not working before.


rdstern (approved):
@Vitalis95 many thanks.  That seems to work and is a great relief.  I am approving so the changes can be tested and merged.

Please can it only partially fulfil the issue, because the plot isn't great and I have a new plan instead of trying to get the plot working.  I'll add to the issue later today.

PR #8377, opened 2023-06-14
 Split the Climatic > Tidy and Examine menu 

### LINKED ISSUE #8348: Split the Climatic > Tidy and Examine menu
Please proceed with these changes.  (I earlier wrote:  Please do not implement this issue, until others, including some of @N-thony and @Patowhiz and @rachelkg and @lloyddewit have given their views.)  I have some views now, and would like to proceed!

The climatic > Tidy and Examine menu is currently as follows:

image

I have proposed the addition of Pivot Tables to the bottom of this menu.  This will make it the longest menu un the whole of R-Instat! (The Prepare > Data Frame is the only other one that comes close.)

I have recently been pretty happy with some of the menu changes - particularly to the describe menu.

I have also been happy with the way that the climatic courses go steadily down the climatic menu.  I now tentatively propose we split this climatic Tidy and Examine menu.  

If so the first half becomes `Tidy Data` and the Second would be `Examine and Edit. `

Then the Tidy Data is the first 2 sections, so down to Append.

The Examine and Edit would swap the order of the remaining parts, so the 7 items would become:

One Variable Summarise...
One Variable Graph...
One Variable Frequencies...
----------------------
Pivot Table...
--------------------
Setup for Data Editing
Daily Data Editing
Compare Columns

(Note the change in name - no entry anymore that's for Climsoft!)

I am growing to like this as I write it.  

In addition I find I sometimes need the `Transform Text` and `Split Text` dialogs from the Prepare menu.  With the new setup I propose they could become a small 3rd section of the `Climatic > Tidy Data` menu.

What do you think?

MeSophie:
Fixes #8348 
@rdstern and @N-thony  I splitted  the Tidy and Examine menu. Please have a look. Thank you.

image


rdstern (changes requested):
@MeSophie this is great.  Well done.
a) In the Tidy Data add ... after Transform Text and Split Text.
b) Change Examine and Edit to `Examine/Edit Data` 
c) In the Examine menu add ... to the last 4 entries.
d) And I hope you can also auto-fill the pivot table if the data have been defined as climatic

Great.

PR #8380, opened 2023-06-14
Merged CrowdIn changes from PR #8379 into translations table in database

lloyddewit:
@rdstern I rebuilt the translation database from scratch. This is the step that caused problems in the previous release.
Please could you test?
If everything is OK, then I will merge and build release 0.7.15.
Thanks


rdstern (approved):
@lloyddewit looks fine to me.  I have switched languages like crazy and it seems pretty smooth now.

PR #8381, opened 2023-06-15
Ensured correct default for 'Maximum Factor Levels Shown', in Describe>One variable>Summaries dialog

### LINKED ISSUE #8371: Small correction needed for Describe > One Variable Summarise
In the default setting it looks like this:

image

When run, the R-code does not include `maxsum = 12`.  (It used to - at least in Version 0.7.10)
It includes the maxsum = 10 if the value is changed from 12 to 10.  But it needs to include the maxsum = 12 always, because 12 is our R-Instat default, but it isn't the R default.

@N-thony this is a very simple one for anyone to correct.

EstherNjeriLiberatta:
Fixes #8371 

This PR fixes the issue of `maxsum` not including the default value 12 in the code.

@rdstern @N-thony This is ready for review.


rdstern (approved):
@EstherNjeriLiberatta great to have you back!  Looks fine to me.

lloyddewit (approved):

PR #8383, opened 2023-06-15
Fixing error from Climatic maps> Themes background

### LINKED ISSUE #8345: Themes Background options throws an error
In the climatic module (and graphs), setting panel.background and panel.border options throw an error. Below is a snapshot of the sample dialogues.

image

The R code error thrown is below
image

anastasia-mbithe:
Fixes #8345 
This is ready for review. 
Though I didn't manage to test this since I didn't get any shapefile data with the geometry parameter, maybe @NanyanziAlice can do the first test.


rdstern:
@NanyanziAlice have you been able to check this?

NanyanziAlice:
@rdstern 
Yes, I did check it and it is now working fine. However, there was an issue that cropped up not necessarily an error but rather a concern about the user's experience when they leave the fill parameter of the panel. border unchecked which is set by default to white in the classic and grey themes so whatever is inside the panel such as a map would not be visible.  Anastasia and I would appreciate an opportunity to chat about it if necessary.

anastasia-mbithe:
Hi @N-thony, yes, for this issue it was done. The other was a different discussion/issue we wanted to get Roger's view, still on the same dialog.

derekagorhom:
@anastasia-mbithe 
the changes to the panels work fine but the map ends up not being visible to the user. i think Alice raised that concern when she did a review and its still the same.

derekagorhom:
@N-thony sorry for bring this here on this PR, but since this PR attempts to fix the theme issue i felt here so better. 
It seems Panel.border is the reason why the maps are not visible 
image

Here is when you fill the panel.border option 
with panel

Here is when you dont fill the panel.border option 
Without panel


Is it ok if we disable the panel.border option for the user?

Patowhiz:
@derekagorhom does this require @rdstern input?

derekagorhom:
@Patowhiz i think Roger can weight in, 
I do not know what the `Panel border` is used for (I have never used it). but if it is in regards to theme background then the `panel background` is the one responsible for the customization.

Patowhiz:
@derekagorhom thanks for your willingness to take on this task.

Having reviewed the issue and PR, it appears to be primarily a technical matter that doesn't need @rdstern's involvement at this stage.

It's my belief that completely disabling the feature without a comprehensive grasp of the underlying problem might not be the most prudent approach.

I would like to suggest that you delve into the R ggplot package, paying particular attention to the panel options, to gain a deeper understanding of why the script generated by the sub-dialog is triggering the issue described. Subsequently, I encourage you to scrutinize the alterations implemented in this PR to assess if they indeed address the issue in the most appropriate manner. If these changes prove inadequate, I suggest you create a new PR that attempts to fix the issue.

Your insights and feedback on this matter would be greatly valued as they will contribute to our progress.

Thanks.

PR #8391, opened 2023-06-20
Added count of the number of elements that have missing values by period (and station) to Inventory dialog

### LINKED ISSUE #8342: Add enhancements to Climatic > Inventory dialog
There are 3 small changes and then one tricky one!  The small ones first:

a) When you first use the Missing Option, could the Summary be enabled by default.
b) When you first select Datails, could the Months be the default, rather than the days.  Also make Date Order the default, rather than Element Order.
c) When it produces a data frame, could it add the key fields.  That needs the Station Name (if used), Element, From and Level variables.

The next one is trickier, so perhaps implement and merge these 3 changes first. @N-thony I suggest @Vitalis95 for this, but you may perfer someone else?  

d) The trickier one!  Could it add the cum (or even the cum1) variable - not both, which I have added by hand in the data frame below.  This needs a nifty bit of e-programming - probably @lilyclements and then implemented by @Vitalis95 ?

This adds a new variable to help the dialog be used more efficiently.

image

I think it is only added when the option is by month, and possibly only when the result is saved.  It facilitates the use of filters to look at the different sort of missing months (or Days, or Years!)

image
 
The column cum starts as zero.  Then look at the last 3 rows above.  All 3 elements were missing for the same set of months.  So the last value gives a 3 there.  (That's useful, because there only were 3 elements.  So if they were all missing, then perhaps the data sheets were never sent to the HQ?  So good to look to see if there is a copy at the station.)

The 2 rows above that were an example where rainfall was entered, but not tmax and tmin.  So there is a 2 there!  (Then the data must have been sent to HQ because one element was entered.)

Happy if that is all that can be done.  The complication is  when there is an overlap in the dates.  So I have added 0.5 to the number of rows above of the overlap.  I am then particularly interested in the 2.5, which says there are some months there that may have data at the station only!  So I could then filter on > 2!

Vitalis95:
Fixes #8342 
@rdstern @lloyddewit @lilyclements @N-thony , this PR fixes part d) of the issue; adding in a count of the number of elements that have missing values by period (and station)

### DIFF
--- a/instat/dlgInventoryPlot.vb
+++ b/instat/dlgInventoryPlot.vb
@@ -25,7 +25,7 @@ Public Class dlgInventoryPlot
     Private clsNewCAddKeyFunction As New RFunction
     Private clsCumulativeInventoryFunction As New RFunction
     Private clsDummyFunction As New RFunction
-    Private clsDataFramaFunction As New RFunction
+    Private clsDataFrameFunction As New RFunction
     Private bResetSubdialog As Boolean = True
 
     Private Sub dlgInventoryPlot_Load(sender As Object, e As EventArgs) Handles MyBase.Load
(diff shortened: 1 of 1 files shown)


rdstern (dismissed):
@Vitalis95 and @lilyclements that looks great.  I am not sure cum is needed, once we have cum1, but I'd like to keep both for now while we use it.  
The one change is that when we save the results into a data frame, I don't think we need them in the output window as well.  I hope that is easy to arrange.  I am also approving, because it will be great to add this feature.

Vitalis95:
@rdstern ,results are no more in the output window when we save it.

rdstern:
@Vitalis95 except you have now gone too far in the other direction!
a) When the results are in a data frame I still want to be able to have the comment reported in the output window

image

Here is the output window reporting the filts sub-dialog, but not that I used the inventory.

b) When the results are not saved into a dat frame, then they should go into the output window!
The results above are actually after I ran again, without saving into a data frame.  

Thanks

Vitalis95:
@rdstern , have a look at it. Thanks

rdstern (commented):
@Vitalis95 I could accept that.  Except it wasn't a tibble before.  Now it is and so only gives 10 rows and suggests the command to get more.  That is ok in the output window, but I would have liked more rows in the maximised window.  If it remains the same, then I suggest you need to make n larger than 10?

image

And I tried the summary dialog, which shows the comment fine - see figure above?  I wonder why I don't get it here?

Vitalis95:
@rdstern , I checked with Patrick on the  comment, probably there some controls which are not translated. That one can be done on a separate pull request

rdstern (approved):
@Vitalis95 this is great.  @lloyddewit I hope the code is acceptable so it can be merged soon?

instat/dlgInventoryPlot.vb
```diff
@@ -194,6 +192,8 @@ Public Class dlgInventoryPlot
         clsDummyFunction = New RFunction
         clsAddKeyFunction = New RFunction
         clsNewCAddKeyFunction = New RFunction
+        clsCumulativeInventoryFunction = New RFunction
+        clsDataFramaFunction = New RFunction
```
lloyddewit, line 196: Is this a typo or a deliberate name?
```suggestion
        clsDataFrameFunction = New RFunction
```

lloyddewit (commented):
@Vitalis95 Thanks, just one small question

rdstern (approved):
@lloyddewit this still looks ok to me.  This is also an improvement that will be good for the next workshops.

lloyddewit (approved):

PR #8398, opened 2023-06-21
Ensured that If a variable is deleted, and the variable is part of a key, then the key is also deleted

### LINKED ISSUE #8296: Deleting a variable, should also delete a key, if the variable was part of it.
A dataframe needs a set of key columns before it can be used to add Comments.

I recently had an error when trying to add another comment.  I found the cause of the error was that I had deleted the columns that made up the key.  The key still remained, but of course was no longer usable.  I then tried to delete the unusable key and it wouldn't let me do this.
I suggest that we should never allow this situation to occur.  Whenever variables are deleted, then, if any are also part of a key, the key is also deleted.  

@Vitalis95 perhaps you could allocate this.  I realise it may be a bit tricky to program.  But we already do have a dialogue to View and Delete keys.  We would just have to check if the column being deleted is part of a key. (Presumably the Comment code checks this?)  If so, then run that delete key code first.   The delete key dialogue also gives a message in the output window and that would also be good to include when the column is deleted.

lilyclements:
Fixes #8296 with @Vitalis95 

@rdstern @N-thony this is ready to review

### DIFF
--- a/instat/dlgDeleteRowsOrColums.Designer.vb
+++ b/instat/dlgDeleteRowsOrColums.Designer.vb
@@ -60,9 +60,10 @@ Partial Class dlgDeleteRowsOrColums
         '
         Me.lblNumberofRows.AutoSize = True
         Me.lblNumberofRows.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.lblNumberofRows.Location = New System.Drawing.Point(7, 108)
+        Me.lblNumberofRows.Location = New System.Drawing.Point(10, 162)
+        Me.lblNumberofRows.Margin = New System.Windows.Forms.Padding(4, 0, 4, 0)
         Me.lblNumberofRows.Name = "lblNumberofRows"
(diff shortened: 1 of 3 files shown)


rdstern (commented):
@lilyclements this seems to work fine.  Happy to approve, but I wonder about 2 related things.  
a) It now works when you delete a column, but not if you rename.  Could it also delete the key is a variable in a key is renamed?
b) When you make a key it is noted in the output window - as follows:

image

So it also notes when you remove it.  It doesn't say from which data frame.  I wonder if it could/should report even when it is removed "behind the scenes"

lilyclements:
@rdstern - 

> a) It now works when you delete a column, but not if you rename. Could it also delete the key is a variable in a key is renamed?
Sorted that now

> b) When you make a key it is noted in the output window - as follows:
This is already occurring if you remove it by the dialog. It doesn't if you remove it by the right click on the cell, because we don't have an `iCallType` for the grid. Is this ok for now?

lloyddewit (approved):

Issue #8301, opened 2023-05-02
Some short-cut keys don't work in the script window


rdstern:
@lloyddewit I have been using multiple script windows recently and all seems to be going well.

But <ctrl>C for copy and <ctrl>X for cut don't seem to work.  I think all the others are ok.

And, though I know it is a different topic I have found a very nice discussion on splitting a long line into multiple lines.  I think this will work nicely in R-Instat, but I wonder what to do in "the opposite situation".

So first what I think works well.  Namely you can split over multiple lines as long as R knows the command isn't finished.

So x <- 2 + 3 +
4
Is fine and R is happy, but don't do:
x <- 2 + 3
+ 4

A second example is

x <- ( 2 + 3
)/4

but not x <- (2 + 3)
/4

I like all this.  But sometimes the code I run has accidentally more open brackets than closed brackets.  Then, of course, R-Instat "appears to hang".  Then I need to put a line with just a close bracket by itself and - then relief - I usually get an error.

Am I right that absence of a close bracket is the usual reason for hanging oddly like this?  And also that the "solution" when you think it is hanging - because it might just be slow in executing - is to add a close bracket, because there is no other indication that it is waiting for me - rather than me waiting for it to execute?

lloyddewit:
@rdstern 

> [ctrl] C for copy and [ctrl X] for cut don't seem to work. I think all the others are ok.

I tested and confirmed the same behaviour. These short cuts used to be provided by the third-party control I used for the text part. I'll investigate and add the functionality myself if needed.

> you can split over multiple lines as long as R knows the command isn't finished.

For the examples you gave, I think the script window's behaviour is correct. 

However the behaviour is far from perfect. Currently ctrl-Enter just adds the current line to a buffer. If it thinks the buffer is a valid R statement, then it executes it and clears the buffer. It uses heuristics (such as hanging binary operators, missing brackets etc.) to judge if the buffer contains a valid statement. However this sometimes fails. 

I plan to improve the script window's back-end so that it uses the RScript library to find the complete statement at the cursor, execute the statement and then move the cursor to the start of the next statement. This is also what RStudio does and should solve the issues related to multi-line statements. This is documented in issue #7989.

rdstern:
@lloyddewit I am adding a different question here as perhaps it could be considered at the same time that you fix the keys above.  And I am not even sure it is much of an issue.  This is the Save and Save As keys/facilities.

[...]

PR #8405, opened 2023-06-22
Ensured that Script window short cut keys work correctly

Patowhiz:
Partly fixes issue #8301.

@lloyddewit as I understand it, the previous code replaced here, attempted to combine the top visible buttons with the context menu options.
My approach leans towards attempting to treat them differently in implementation. Unlike the top buttons, which are always visible, right click menu options are only visible when the user rights clicks. So I'm now disabling them during opening.
If a user attempts to use short cut keys, then the individual events call functions that give sensible feedback to the users.
This is what was previous there.

Thanks


rdstern (dismissed):
Another "first-time round".  Seems to work fine.  Maybe I am not critical enough.

instat/ucrScript.vb
```diff
@@ -236,40 +236,42 @@ Public Class ucrScript
...
     Private Sub EnableRunButtons(bEnable As Boolean)
         cmdRunLineSelection.Enabled = bEnable
         cmdRunAll.Enabled = bEnable
     End Sub
 
+    ''' <summary>
+    ''' Enables or disables all right click menu options
+    ''' </summary>
+    ''' <param name="bEnable">If true, enables all right click options,false otherwise</param>
+    Private Sub EnableRightClickMenuOptions(bEnable)
```
lloyddewit, line 260:
```suggestion
    Private Sub EnableRightClickMenuOptions(bEnable As Boolean)
```

instat/ucrScript.vb
```diff
@@ -236,40 +236,42 @@ Public Class ucrScript
     Private Sub EnableDisableButtons()
 
         Dim bIsLogTab As Boolean = TabControl.SelectedIndex = iTabIndexLog
-
-        mnuUndo.Enabled = clsScriptActive.CanUndo AndAlso Not bIslogTab
-        mnuRedo.Enabled = clsScriptActive.CanRedo AndAlso Not bIslogTab
-
         Dim bScriptselected = clsScriptActive.SelectedText.Length > 0
```
lloyddewit, line 239:
```suggestion
```

lloyddewit (changes requested):
@Patowhiz Thank you, this is a good approach.
Just 2 small suggestions

rdstern (approved):
Nice!

lloyddewit (approved):

Issue #8411, opened 2023-06-29
Minor bug in describe one variable frequencies


Fidel365:
When opening the dialogue for the first time, the multi receiver for Table and Graph is automatically selected unlike for the stem and leaf where you should click the receiver for it to be selected.

Fidel365:
@N-thony , this has already been merged

PR #8412, opened 2023-06-29
Ensured that stem and leaf fields set correctly when opening One Way Frequencies dialog for first time

Fidel365:
fixes issue #8411
@rdstern , this PR is ready for review


rdstern (approved):
@Fidel365 I am approving as it still seems to work.  In the future you could mention, in the pull request, what the minor change was.  Then I'll know better what I am supposed to be looking for.

lloyddewit (approved):

PR #8416, opened 2023-07-03
Ensured that columns in the Data Grid can be deleted even when there are no key columns

### LINKED ISSUE #8413: Delete columns sometimes doesn't work anymore
I am using the merged version of R-Instat

@Vitalis95 and @lilyclements I am pretty sure this bug has followed the merging of pull request #8398.

I read in the usual survey and immediately tried to delete the columns as shown below.  It doesn't work anymore:
It should be easy to fix as I explain below:

image
.
I suggest the problem is that deleting a column, now checks and deletes the corresponding key columns.  This data frame doesn't have a key.  If I add a key, then the delete works fine.  The code needs to cope with the situation where there is no key.

lilyclements:
Fixes #8413

@Vitalis95 can you review this, and then it can be ready for @rdstern to review

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -6,52 +6,52 @@ DataSheet <- R6::R6Class("DataSheet",
                                                  messages = TRUE, convert=TRUE, create = TRUE, 
                                                  start_point=1, filters = list(), column_selections = list(), objects = list(),
                                                  calculations = list(), keys = list(), comments = list(), keep_attributes = TRUE)
-{
-  # Set up the data object
-  self$set_data(data, messages)
-  self$set_changes(list())
-  #removed until this can be fixed.
-  #self$set_variables_metadata(variables_metadata)
-  
-  # Set first so that "no_filter" is added
-  self$set_filters(filters)
-  self$set_column_selections(column_selections)
-  if(keep_attributes) {
-    self$set_meta(c(attributes(private$data), metadata))
-  }
-  else {
-    self$set_meta(metadata)
-    self$clear_variables_metadata()
-  }
-  self$add_defaults_meta()
-  self$add_defaults_variables_metadata(self$get_column_names())
-  #self$update_variables_metadata()
-  self$set_objects(objects)
-  self$set_calculations(calculations)
-  self$set_keys(keys)
-  self$set_comments(comments)
-  
-  # If no name for the data.frame has been given in the list we create a default one.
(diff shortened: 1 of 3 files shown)


rdstern (approved):
@lilyclements many thanks - this seems to work fine now.  @lloyddewit over to you

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -771,12 +771,14 @@ DataSheet$set("public", "rename_column_in_data", function(curr_col_name = "", ne
...
-        if (self$get_variables_metadata() %>% dplyr::filter(Name == curr_col_name) %>% dplyr::pull(Is_Key)){
+      if (!is.null(self$get_variables_metadata()$Is_Key)){
+        if (self$get_variables_metadata() %>% dplyr::filter(Name == col_name) %>% dplyr::pull(Is_Key)){
           active_keys <- self$get_keys()
-          keys_to_delete <- which(grepl(curr_col_name, active_keys))
+          keys_to_delete <- which(grepl(col_name, active_keys))
           keys_to_delete <- purrr::map_chr(.x = keys_to_delete, .f = ~names(active_keys[.x]))
           purrr::map(.x = keys_to_delete, .f = ~self$remove_key(key_name = names(active_keys[.x])))
         }
+      }
```
lloyddewit, line 781: indents correct?

lloyddewit (changes requested):
@lilyclements Thanks looks good to me. If you can fix the indentation, then I can approve and merge

instat/static/InstatObject/R/data_object_R6.R
lilyclements: They should be! I just copied and pasted the file into R and then back into the VB document to autofix the indentations of everything in this file

lloyddewit: @lilyclements I assume that you pasted the whole file into the RStudio editor, and then pasted back into the Visual Studio editor. Unfortunately the GitHub diff interprets this as all 4000+ lines changing (probably because of different white space characters and/or line breaks). This makes it impossible to see the real changes in the PR and has also triggered a merge conflict.

I suggest that you go back to the previous version of the file and fix just the indents mentioned above manually.

Note: you can get the previous version of the file by checking it out from commit `102436e`. If you're unsure how to do this, then just let me know.

I'm sorry my comment caused such trouble, I only intended some spaces to be added to the few lines above.
Thanks for your patience.

image

PR #8418, opened 2023-07-04
Ensured that insert/delete row in Data grid does not trigger jump to the top of the page

### LINKED ISSUE #8415: Insert row and delete row jump to the top of the page
I have been inserting a row into a data frame.  It is working well.  I am using the right-click.  
However, once it is finished I am automatically moved to the top of the page.  I would prefer to stay on the same row in the data frame.

I use the Row > Right click > Insert row(s) option.   

There is also the Prepare > Data Frame > Insert Columns/Rows (and Delete Columns/Rows.
There the Insert Rows has the same problem.  I have not yet checked the other combinations.

(If there are multiple pages it takes me to the first row on the same page.  I am pleased it remains on the same page.)

N-thony:
Fixes #8415 
@rdstern please test.

### DIFF
--- a/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
@@ -86,7 +86,9 @@ Public Class ucrDataViewReoGrid
             End If
         Next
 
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@N-thony welcome back.  R-Instat has really missed you!  This looks fine.  Thanks

lloyddewit (approved):

PR #8423, opened 2023-07-06
Implementing the Find In Variable and Filter Dialogue

N-thony:
replaces #8152 
@rdstern please test.
I have also added `CTRL + F` to open the dialgoue.

### DIFF
--- a/instat/Interface/IDataViewGrid.vb
+++ b/instat/Interface/IDataViewGrid.vb
@@ -35,7 +35,7 @@ Public Interface IDataViewGrid
 
     Sub AddRowData(dataFrame As clsDataFrame)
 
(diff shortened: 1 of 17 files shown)


rdstern:
@N-thony have the conflicts been resolved?  It would be good to have this feature.

N-thony:
> @N-thony have the conflicts been resolved? It would be good to have this feature.

Resolved now

rdstern (commented):
@N-thony I am pleased you are back to this.  It seems to work fine, but just for the first "page".  If I have more than 1000 and go to another page, then it switches back to the starting 1000.  I don't mind if it is just for a page, but it needs to work on any one.

PR #8424, opened 2023-07-06
Ensured that Title values does not reset to default when changed in Plots sub dialog

### LINKED ISSUE #8417: Title, subtitle and caption size values reset to default values after a change
**Describe the bug**
When you change one or two of the three size values (i.e title, subtitle and caption ), the rest are reset to default implying the changed values are not captured.

**To Reproduce**
1. Go to 'Plot Options'
2. Click on 'Title Tab'. The default values (i.e title size=20, subtitle size=15 and caption size=8) are set as below:
image
3. On changing the title size value to 30. The change is captured in R code.  Corresponding snapshots are shown below
image

image

4. Then, changing the subtitle size value to 25.  The change is captured but the 'Title' size value is reset to the default value of 20 in the R code yet the dialog indicates otherwise. See snippets  below:
image

image

5. Further, by clicking on the 'Tag' checkbox, the three size values are set to default in the R code yet these values were changed on the dialog. See below
image

image

6. Finally, changing the 'Tag' size value to 20 and then checking the 'New Legend' tab, the tag size is not set in the R code yet the dialogue shows its value as 20.  
image

image

**Expected Behaviour**
Whatever is changed in the dialog should be correctly implemented in the R code.

Vitalis95:
Fixes #8417 
@rdstern @N-thony @NanyanziAlice , have a look at it


rdstern (changes requested):
I think this problem is solved.  But can you check on the default sizes.  I think that when you leave the title size at 20 then it isn't really 20 - so perhaps you include `plot.title=ggplot2::element_text(size=20)` whenever the rirle is included, even when it has the R-Instat default.

Vitalis95:
@rdstern , have a look at it

NanyanziAlice:
This works well now, Thanks @Vitalis95 ! 

@rdstern Is it necessary to add the Tag position (i.e top, topleft, topright, bottom, etc) as a feature?

rdstern (approved):
Seems fine to me.
@NanyanziAlice as the caption was easily available, it seemed useful to include as an option.
@lloyddewit over to you.  If ready in time, then it could be merged.  Otherwise it could wait for August.

lloyddewit (approved):

Issue #8408, opened 2023-06-28
Not able to save the output window


rdstern:
![image](https://github.com/africanmathsinitiative/R-Instat/assets/11226469/c2a36be2-b13a-402a-ad67-7ca8f256d8a8)

@Patowhiz what can we do for the participants on the workshop?

We are using 7.14, but I have trioed in 7.15 and this is the current merged version.

The participants want to save the output window.  Please can you help with the current version as well as sorting out a fix.  It seems to be a general problem and I am a bit at a loss that I did not notice it befoe, as I have been considering the saving aspects.

rdstern:
@Patowhiz yes, this remains the major problem we have found, with the software so far.  They were hoping to save a record of the results through the week, each on their own data.  REal blow that they can't   Hope it will be fixed for the next workshop, which is Tajikistan in a couple of weeks.

Particular problem here, because the team used R-Instat before, so it is an item that used to work, and has now stopped.

Patowhiz:
@ChrisMarsh82 has suggested, converting html to images as a start. I'll explore this as well.

rdstern:
@Patowhiz and @ChrisMarsh82 I find it hard to address your options, while something this important - trhat used to work - has stopped working.  And there was no isuue on that, so I found out only during a workshop.  How did I miss the problem as I have been worrying about saving options and why didn't it surface during your visit to UK, or the sprint?

In answer to your queries:

1. I am happy for you to explore options.  I note, for example, there is the [gembox software](https://www.gemboxsoftware.com/document) that seems to do the sort of thing we require, namely to transform output into different formats.   (That is commercial, but was can consider it, if it the best option, because it is royalty free.)  

There is a free version - limited to 20 paragraphs!  Even that would be ok for now.  We need something working by the next release, which is Wednesday!  I am happy with anything, as long as there is something.

As you know I have been asking for more options for your maximised window and, in the longer term it would be good to consider the two together.  We used the maximised window a lot in the workshop, see issues #8141 and #8189.

2. Agreed - if they are disabled, then don't need to print.

3. Not needed now.  Let's concentrate on saving something.

4.  In your more general assessment, please remember the gt software permits the option of tables in different formats.  Perhaps this also relates to options for you.  

[...]

PR #8429, opened 2023-07-10
Restores 'add to' and 'save' output window features

Patowhiz:
Fixes partly #8408

This is a temporary solution. It restore previously regressed features in a way that it was previously implemented.
Html outputs will still not be saved.

### DIFF
--- a/instat/Model/Output/clsOutputLogger.vb
+++ b/instat/Model/Output/clsOutputLogger.vb
@@ -77,36 +77,47 @@ Public Class clsOutputLogger
             Exit Sub
         End If
 
-        Dim outputType As OutputType
-        If String.IsNullOrEmpty(strOutput) Then
-            outputType = OutputType.Script
-        ElseIf Not bAsFile Then
-            outputType = OutputType.TextOutput
-        Else
-            Dim strFileExtension As String = Path.GetExtension(strOutput).ToLower
-            Select Case strFileExtension
-                Case ".png"
-                    outputType = OutputType.ImageOutput
-                Case ".html"
-                    outputType = OutputType.HtmlOutput
-                Case ".txt"
-                    outputType = OutputType.TextOutput
-                Case Else
-                    MessageBox.Show("The file type to be added is currently not suported",
-                                "Developer Error",
-                                MessageBoxButtons.OK,
-                                MessageBoxIcon.Error)
-                    Exit Sub
-            End Select
+        'add the R script as an output element
+        Dim rScriptElement As New clsOutputElement
+        rScriptElement.SetContent(strScript, OutputType.Script, "")
+        _outputElements.Add(rScriptElement)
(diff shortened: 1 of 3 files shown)


rdstern:
@Patowhiz looks very promising already.  I tried with one variable summarise and also one variable graph.

It saved the graph nicely, but not the summary - there it seems to save a link to the file instead.

On details it doesn't yet seem to remember the directory I save it to.  It always seemed to come back to the Documents directory.

In the future I am hoping the Save in the Output window might vbe a Save - rather than Save As?  The only Save As will be from the File menu? 

But I'll take anything working for the start - as soon as it saves the txt output, rather than a link we have the basics.

Patowhiz:
@rdstern I have fixed the text output and remembering of directory. 
Saving as links is definitely not desirable, considering that a user may want to share the file to another user with a different machine.

@rdstern I'm not sure if I should add links for htmls for now? Note this feature was there in the older versions anyway. 
To correctly implement html saving, I need some time.

rdstern (approved):
@Patowhiz that's a relief that it now works.  
As you mention, there's lots to improve still, including html, but I'd be very happy if this could now be merged.  So, over to @lloyddewit now.
I then hope that the work does not stop, so the output maximised windows have more possibilities in the mid-August release.

instat/UserControl/ucrOutputPage.vb
```diff
@@ -403,10 +418,17 @@ Public Class ucrOutputPage
...
             Case OutputType.Script
-                FillRichTextBoxWithFormatedRScript(richText, element.FormattedRScript)
+                FillRichTextWithRScriptBasedOnSettings(richText, element.FormattedRScript)
             Case OutputType.TextOutput
-                'todo. check if output is file or not. if file, read the contents of the file
-                AddFormatedTextToRichTextBox(richText, element.Output, OutputFont.ROutputFont, OutputFont.ROutputColour)
+                Dim strOutput As String = ""
+                If element.IsFile Then
+                    For Each strLine As String In IO.File.ReadLines(element.Output)
+                        strOutput = strOutput & strLine & Environment.NewLine
```
lloyddewit, line 426:
```suggestion
                        strOutput &= strLine & Environment.NewLine
```

instat/Model/Output/clsOutputLogger.vb
```diff
@@ -77,36 +77,47 @@ Public Class clsOutputLogger
...
+                                    "Developer Error",
+                                    MessageBoxButtons.OK,
+                                    MessageBoxIcon.Error)
+                        Exit Sub
+                End Select
+            Else
+                outputType = OutputType.TextOutput
+            End If
+
+            'add the output with it's R script as another output element
```
lloyddewit, line 111:
```suggestion
            'add the output with its R script as another output element
```

PR #8430, opened 2023-07-10
Added Help Button to Maximise window and Output window

### LINKED ISSUE #8278: Change the Help option from right-click menus and Windows
We have a variety of right-click menus.  Within them the Help does different things.  This issue proposes consistency.  

a) In the **_column metadata _**, if you right-click in the Window, then there is just the Help

image

If you choose it then you go to that window in the herlp file

image

I suggest we keep that, as it is.

b) Same for the Data Frame meta-data.  I suggest no change.

c) The new Script/Log window has both a help button and also has help at the bottom of the right-click menu:

image

They both go to the Script Window.  We need to change the help file, because there is now a Log/Script Window, but I suggest we keep both options.  So I propose no change.

d) The data view has 3 right-click menus.  I suggest we change the context number, so they each go to the Data View Window in the help file.  So change the context number for each of them to 134.  @derekagorhom could you do this?

e) @ChrisMarsh82 The output Window does not yet have a right-click menu, or a help button.  It has plenty of tab-type buttons.  If you agree, then please could **_Help_** be added, or could there be a right-click with Help, or both?  The help context number is 540.    

f) @Patowhiz the Maximise window currently has a File button, but not a Help button.  If you agree then could one be added, or a right-click with Help, or both?  The Context is 590.  The help file currently has a Browser Window - becasue that is what we needed before.  I propose to call it **_Maximise Window_**, if you can't suggest a better name. 

g)  The only other right-click (I think) are in the dialogues.  They go to a **_Controls in Each Dialogue_** entry in the help and I propose to leave that - at least for now.

derekagorhom:
Fixes #8278 
I have added help buttons on both Maximise and Output windows and also added the help context numbers

@rdstern , @N-thony can you review this PR

Thanks

### DIFF
--- a/instat/UserControl/ucrOutputPages.Designer.vb
+++ b/instat/UserControl/ucrOutputPages.Designer.vb
@@ -37,6 +37,7 @@ Partial Class ucrOutputPages
         Me.tabControl = New System.Windows.Forms.TabControl()
         Me.tpMain = New System.Windows.Forms.TabPage()
         Me.ucrMainOutputPage = New instat.ucrOutputPage()
+        Me.tbHelp = New System.Windows.Forms.ToolStripLabel()
         Me.TableLayoutPanel1.SuspendLayout()
         Me.tsButtons.SuspendLayout()
         Me.tabControl.SuspendLayout()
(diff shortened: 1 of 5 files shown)


rdstern (commented):
This is a useful and simple addition
a) In the output window the space between the Save (really still currently Save As) and the new Help seems smaller than between the other items.  Is that my imagination.
b) In the Maximise window the Help has a box round.  I don't think that's needed.

I note perhaps for @Patowhiz that when the output window is not maximised, some of the items are not visible.  I presume that's ok?  I mention it now in case any changes are foreseen for the August release.  
@lloyddewit I would like this to be merged as soon as possible, but not in the July release.  It puts us under good pressure to improve the help on the windows for the August release.

derekagorhom:
@rdstern I have made the changes, can you review them again 
thanks

rdstern (approved):
This looks fine.  @lloyddewit I would be happy to see it merged.  I am keen to update the help file for the August R-Instat release, and this feature is good motivation.

lloyddewit (approved):

lloyddewit:
> @lloyddewit @rdstern can we merge this?

yes! I thought I merged this on 19 July but I obviously did something wrong. Thanks for spotting

PR #8433, opened 2023-07-11
Change "rename_column_in_data" function to use curr_col_name not col_name

### LINKED ISSUE #8413: Delete columns sometimes doesn't work anymore

lilyclements:
fixes #8413

The issue before was that renaming columns would pull an error. This should resolve this. The issue occurred because the code was calling `col_name` not `curr_col_name`. This came from a copy-paste issue by me when putting similar code in the `"rename_column_in_data"` function that is in the `"remove_cols_from_data"` function.

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -771,7 +771,7 @@ DataSheet$set("public", "rename_column_in_data", function(curr_col_name = "", ne
           warning("Multiple columns have name: '", curr_col_name, "'. All such columns will be renamed.")
         }
         # remove key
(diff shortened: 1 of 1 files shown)


rdstern (approved):
Looks ok to me.  @lloyddewit over to you

PR #8431, opened 2023-07-11
Ensured that Pivot Table dialog in Climatic menu puts the factor variable in the right order

MeSophie:
Fixes #8420
@rdstern I corrected the order of factor in climatic menu. Please can you have a look?

PR #8437, opened 2023-07-13
Ensured that the Change of the Data Frame Works correctly in the Pivot Table dialog in Climatic Menu 

MeSophie:
@rdstern, @lloyddewit  base on this comment

>  @MeSophie and @lloyddewit of course I checked something slightly different this time, and found another problem with the new pivot table. I hope it might be easy to fix?
> 
> To see the problem:
> 
> a) open (say) Domoma from the library, or any of the climatic sets you have been using to check yourself.
> b) Then open a second data frame - I used the usual survey.
> c) Then go into your climatic pivot table with the wrong data frame.
> d) Then try to change to the correct data frame. I found it was impossible. It insists on the original data frame. Even going to the correct data frame and pressing Reset does no good. It insists on sticking with the wrong data frame.

 in PR #8431
 I corrected the bug when trying to change data frame in Pivot table. Please have a look.

### DIFF
--- a/instat/dlgThreeVariablePivotTable.vb
+++ b/instat/dlgThreeVariablePivotTable.vb
@@ -331,7 +331,8 @@ Public Class dlgThreeVariablePivotTable
                     ucrReceiverFactorLevels.Add(strMonthCol, strDataFrame)
                     ucrReceiverInitialColumnFactor.Add(strMonthCol, strDataFrame)
                 End If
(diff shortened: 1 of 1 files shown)


MeSophie:
> @MeSophie and maybe @N-thony too. Many thanks - it now seems to work fine. That's a big relief. @lloyddewit I hope we are not too late?

@rdstern most of the work was done by @N-thony

lloyddewit (approved):

PR #8446, opened 2023-07-17
Reset the Multiple Receiver when re-opening the dialog

### LINKED ISSUE #8409: Annoying message when re-opening a dialog
@N-thony this is a perhaps thorny one, so I also copy @lloyddewit on it.  The problem is annoying, But, on the positive side I am delighted that we are now getting to problems that occur rarely, and although annoying, they don't stop anything from working.

Here is a contrived example of the problem:

image

I opend diamonds from the library and went into Prepare > Data Frame > Replace Values with the last 6 (all numeric) variables.

image

Now close the dialog and make 3 of those variables into factors.

Now go back to the dialog and you get the error above.  It comes because you can't have variables of different types in the multiple receiver.  You could have all factors or all numeric, but now I have made a mixture.

I think we have trapped other situations where there have been changes in the variables, but this one just gives the funny message.  And it sometimes gives the message many times.  I am hoping this situation can be fixed, and perhaps when any dialog is re-opened, rather than just for this dialog?

(In the workshop I had 31 variables that were (C) and I needed to make them all numeric.  I replaced something and that worked for about half the variables.  Then I needed to replace something else - and hit this problems because some of the original 31 were now numeric, while others were still character.) 

I would be happy, in this situation, for the receiver to be empty.  I would not object if the solution was effectively to do a reset.

derekagorhom:
Fixes #8409 

This Fix clears previously selected variables in the multiple receiver when the dialog is re-opened. 

Thanks

### DIFF
--- a/instat/ucrReceiverMultiple.vb
+++ b/instat/ucrReceiverMultiple.vb
@@ -477,34 +477,34 @@ Public Class ucrReceiverMultiple
                         Clear()
                     End If
                     SetSelectorHeading("Variables")
-                    ElseIf strVariableTypes.Count > 0 Then
-                        If bCategoricalNumeric Then
-                            If strVariableTypes.Contains("categorical") Then
-                                SetIncludedDataTypes({"factor", "character", "logical"}, bStrict:=True)
-                                SetSelectorHeading("Categorical Variables")
-                            ElseIf strVariableTypes.Contains("numeric") Then
-                                SetExcludedDataTypes({"factor", "character"})
-                                SetSelectorHeading("Numerics")
-                            Else
-                                ' Else it is logical only
-                                RemoveIncludedMetadataProperty(strProperty:="class")
-                                RemoveExcludedMetadataProperty(strProperty:="class")
-                                SetSelectorHeading("Variables")
-                            End If
+                ElseIf strVariableTypes.Count > 0 Then
+                    If bCategoricalNumeric Then
+                        If strVariableTypes.Contains("categorical") Then
+                            SetIncludedDataTypes({"factor", "character", "logical"}, bStrict:=True)
+                            SetSelectorHeading("Categorical Variables")
+                        ElseIf strVariableTypes.Contains("numeric") Then
+                            SetExcludedDataTypes({"factor", "character"})
+                            SetSelectorHeading("Numerics")
                         Else
-                            If strVariableTypes(0) = "integer" OrElse strVariableTypes(0) = "numeric" Then
-                                SetDataType("numeric", bStrict:=True)
-                                SetSelectorHeading("Numerics")
-                            ElseIf strVariableTypes(0) = "ordered,factor" OrElse strVariableTypes(0) = "factor" Then
-                                SetDataType("factor", bStrict:=True)
-                                SetSelectorHeading("Factors")
-                            Else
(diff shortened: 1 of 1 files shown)


derekagorhom:
@N-thony , Thanks for the explanation. 

This is what i was able to do:
When 6 variables are previously selected and 3 are changed to `factor`, the `datatype` which was selected first into the multiple receiver will remain along with the variables with the same `datatypes`. 
and i also removed the error message from the code. 

can you review this again. 

thanks

derekagorhom:
@N-thony, @rdstern  the reason i made the change to this was because i thought as a user; 
the first `datatype` i selected should still be on the multiple receiver after i re opned the dialogue else the dialogue should clear the previously selected variables, so i have something to start from instead on having an empty receiver.

rdstern:
@derekagorhom I agree with the above.  When there is this complication, I would be happy for the receiver to become empty.

derekagorhom:
@N-thony , @rdstern I have made the change, can you review again
Thanks

rdstern (approved):
I could not make the Replace values dialog generate a bug now.

Patowhiz:
It's important to test other dialogs for regression. This fix affects all the dialogs that use the multiple receiver.

lloyddewit:
@derekagorhom I reviewed the code, but the only change I see is that the message box no longer appears. The rest of the functionality is unchanged.
I understood that this PR should clear the multiple receiver whenever it contains more than one data type, but I don't see any code that does this. Please could you clarify?
Thanks

derekagorhom:
@lloyddewit , this is not complete yet as there are still some issues with the multiple receiver not being able to clear all the items yet

derekagorhom:
> @derekagorhom I reviewed the code, but the only change I see is that the message box no longer appears. The rest of the functionality is unchanged. I understood that this PR should clear the multiple receiver whenever it contains more than one data type, but I don't see any code that does this. Please could you clarify? Thanks

@lloyddewit the code used to clear the items in the multiple receiver was already in function `CheckSingleType()` (see line 468, previously 469) so this was more of a code deletion that adding new code.

lloyddewit:
@derekagorhom thank you for the response. I looked at the `ucrReceiverMultiple.vb` and the rest of the issue seems to be fixed in PR #8076. However, it's a bit strange because the change was made several months before issue #8409 was raised.

@rdstern This PR just deletes a message box. Are you confident that the multiple receiver was already working perfectly before you raised issue #8409 and the only problem was that a message box was displayed, even when there was no error?

Thanks

lloyddewit (approved):

rdstern:
@derekagorhom just checking that this is not just the message box that is wrong in #8409.  The problem is that the multiple receiver in this, and a few other dialogs, can only have variables of one type.  They could be all factors, or all numeric, but not a mixture.

So, if a variable's type is changed and it is in the receiver then the contents of the receiver are no longer "legal".  A similar thing happens if variables are renamed, or deleted, but I think we already allow for that.  So, it becomes just another example of illegal contents of the receiver.  We therefore don't need the message, it is right to get rid of it.  And also the contents of the receiver need to be adjusted when this situation occurs.  I'd be happy if the receiver is simply cleared.

rdstern:
@lloyddewit I have checked and my comment above seems to be resolved.  So I think this can be merged.

Issue #8443, opened 2023-07-17
Handle list variables better in data frames


rdstern:
@N-thony I think and hope the obvious bug, shown below, might have an obvious and quick fix.  Then there is more for possible later discussion.  

I start by getting a data frame with 2 list variables.  This is an exciting feature of R-Instat data frames, but there are bugs, and oddities with one that requires an urgent fix.  This is the fact you cannot get a new data frame that is a subset (via the Filter Rows), if there are one, or more list variables.  I hope this might be simple to fix?

First, to understand the problem, we get a dataframe with 2 list variables.
a) File > New Data Frame, with 100 rows and 2 variables.
b) Prepare > Column: Calculator > Integer keyboard Key divisors of x1, put the result in x2.
c) Back to calculator, clear, factorise of x1 with the result into the default of calc.

Here is the resulting data frame, where I also show the data in R (from data frame right-click)

<img width="627" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/2c746609-05ad-46c8-ab39-f927c61b3722">

First the obvious bug, that I hope can be fixed quickly?

Right-click > Filter rows, and apply any filter, say x1 >50, then ask for a subset:

<img width="494" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/0a54dfef-a6ec-4fc9-a5db-05cf5b98dafb">

I hope this might have an easy fix?

Less urgent is whether we could display the list variables "better" and also be able to transform them, etc.  

a) Note the 0 at the top of x2 in the display.  In the R display it is shown as blank.  If we get the length of the items in x2 - from the calculator > list keyboard.  Then it correctly gives the length as zero.  
b) Notice the oddity in the calc column, that the second row in the calc column is given as `2:1`, rather than `2, 1` .  Same is for 8 and 9.  This doesn't seem to do any harm though.

Given the problem above I thought I would make them into character variables.  This makes them all into NA.  With our current procedures I can't make them into character variables.  I wonder if I should and also I wonder if I could be able to make a character variable into a list.  Perhaps this can be considered later?

rdstern:
[mersennes_with_divisors.zip](https://github.com/africanmathsinitiative/R-Instat/files/12206359/mersennes_with_divisors.zip)

@N-thony there is a more serious problem with the lists, now we have the (very nice) wrapping and unwrapping.  I suspect this may be quite a challenge to solve?  It is important for our handling of list variables.  

[...]

PR #7922, opened 2022-10-18
Created the Wrap dialogue

derekagorhom:
Fixes #7812 
@N-thony can you test and review, 
Thanks

PR #8447, opened 2023-07-17
Implemented wrapping and unwrapping on the DataView Grid

### LINKED ISSUE #7812: Add alignmment options to the grid?

N-thony:
Replaces #7922
Fixes #7812
Fixes partially #8443
@rdstern @derekagorhom please test the wrapping and unwrapping here.

### DIFF
--- a/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
@@ -104,7 +104,8 @@ Public Class ucrDataViewReoGrid
 
     Public Sub AdjustColumnWidthAfterWrapping(strColumn As String, Optional bApplyWrap As Boolean = False) Implements IDataViewGrid.AdjustColumnWidthAfterWrapping
         Dim iColumnIndex As Integer = GetColumnIndex(grdData.CurrentWorksheet, strColumn)
-        If iColumnIndex < 0 Then
+        If iColumnIndex < 0 OrElse grdData.CurrentWorksheet.ColumnHeaders(iColumnIndex).Text.Contains("(G)") Then
+            MsgBox("Cannot wrap or unwrap this type of varaible.")
             Exit Sub
         End If
 
--- a/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
@@ -110,12 +110,7 @@ Public Class ucrDataViewReoGrid
         End If
 
         If bApplyWrap Then
-            For i As Integer = 0 To grdData.CurrentWorksheet.ColumnCount - 1
-                If i = iColumnIndex Then
-                    grdData.CurrentWorksheet.AutoFitColumnWidth(i)
-                    Exit For
-                End If
-            Next
+            grdData.CurrentWorksheet.AutoFitColumnWidth(iColumnIndex)
             For i As Integer = 0 To grdData.CurrentWorksheet.RowCount - 1
                 grdData.CurrentWorksheet.AutoFitRowHeight(i)
             Next
--- a/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
@@ -72,6 +72,8 @@ Public Class ucrDataViewReoGrid
             textColour = Color.DarkBlue
         End If
 
+        Dim iMaxWidth As Integer = 0
+
         strRowNames = dataFrame.DisplayedRowNames()
         For i = 0 To grdData.CurrentWorksheet.Rows - 1
             For j = 0 To grdData.CurrentWorksheet.Columns - 1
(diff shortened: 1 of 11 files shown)


rdstern (changes requested):
@N-thony wow - well almost.  I will test a bit more, but it seems almost to be nailed!  Well done.  
The "almost" is a shape file.  When I originally import it is ok.

But when I try the workdwrap on the geometry variable in the shape file, then it gives a serious error and closes the file and tells me to restart.  Please could you check and see what is possible here.
If you can find where it is going wrong, then I see (at least) 3 options:
a) You may be able to fix it, so wordwrap is also possible on this type of variable.
b) It may be possible to fix under certain circumstances, e.g. the width must be more than a certain value.
c) You don't allow wordwrap on this type of variable.  (That would also be acceptable)

A second problem on the shape file is when I save it into a databook.  Then I open it again.  This is desirable, so I can make it easy for users to draw maps.  Then the data sheet comes wordwrapped - see below:
That's one of the reasons I was keen to have the wordwrap feature.  So that I could unwrap this variable!  This gives me an error as shown below:

image




However, after this error, it did unwrap, which is great.  So, just not doing that operation, when there are no \n, seems to work!
--------------------------------------------------------------------------------------------
That also made me try worwrap on a list variable, and that had a very surprising result - at least surprising (and attractive) for me.

I tried a new data frame of length 1000 and then the calculator divisors of x1 into x2.  Then wordwrap of x2.

This works, but it changes x2 into a character variable.  Now, if you look at my recent issue #8443 you will see that I am unable to make this transformation any other way.  (Making a list variable into a character variable makes it always NA!  So you may have a solution to part of that issue in your code!  I suggest you leave that point for now, till you see the solution for geometry-type variables.  Later we may choose to excluse list variables from those that can be wordwrapped.

rdstern (changes requested):
@N-thony this is just great.  And almost there I think.  I first tried with list variables and that was excellent.  As you say, you have also corrected the bug, so I can get a new data frame when there are list variables.  That's so good!

It now does detect a geometry variable and labels it with a (G) that I really like.  And I am happy that you can't wordwrap it.  

a) Please correct the spelling!  It says varaible and not variable!
b) I get an error when trying to wordwrap a variable in a dataframe that has missing values in the data.  It was fine with this variable first and then I made one row missing and got an error.

image

This is almost ready to merge!

rdstern (dismissed):
@N-thony this is exciting, both in its own right and in the implications for the use of the spreadsheets in R-Instat.  We need to keep testing it, but @lloyddewit it would be great to get this merged, as soon as you are happy with the code.

instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
```diff
@@ -102,6 +102,29 @@ Public Class ucrDataViewReoGrid
...
+            Exit Sub
+        End If
+
+        If bApplyWrap Then
+            For i As Integer = 0 To grdData.CurrentWorksheet.ColumnCount - 1
+                If i = iColumnIndex Then
+                    grdData.CurrentWorksheet.AutoFitColumnWidth(i)
+                    Exit For
+                End If
+            Next
```
lloyddewit, line 118: do we need this `For` loop?
```suggestion
            grdData.CurrentWorksheet.AutoFitColumnWidth(iColumnIndex)
```
If we did the above, then would we still need to check `iColumnIndex` for validity?

instat/dlgWordwrap.vb
```diff
@@ -18,5 +18,105 @@ Imports instat.Translations
...
+        ucrReceiverWrapText.SetRCode(clsWrapOrUnwrapFunction, bReset)
+        ucrPnlTextWrap.SetRCode(clsWrapOrUnwrapFunction, bReset)
+    End Sub
+
+    Private Sub TestOkEnabled()
+        If Not ucrReceiverWrapText.IsEmpty Then
+            ucrBase.OKEnabled(True)
+        Else
+            ucrBase.OKEnabled(False)
+        End If
```
lloyddewit, line 91: shorten? (if you prefer to leave as is, then that's also fine)
```suggestion
        ucrBase.OKEnabled(Not ucrReceiverWrapText.IsEmpty)
```

lloyddewit (changes requested):
@N-thony Loooks very good, just a couple of small questions

lloyddewit (approved):

lloyddewit:
@rdstern there were a couple of small changes, please could you retest/approve? thanks

rdstern (approved):

PR #8448, opened 2023-07-18
Added Use as Numeric and Display as Factor checkboxes to Graphics dialog

### LINKED ISSUE #8428: Improve the ribbon geom and the general graph dialog?
@lilyclements has prepared a nice climatic graph as follows:
image

This uses geom_ribbon which we have already in R-Instat.  She uses summary climatic data and then (of course) R code to produce the graph.  In the short-term I would like to be able to produce it using the Describe > General > Graph dialog.  Later I see this as a nice example of a special climatic climatic graph to go into the Climatic > Describe > Rainfall dialog - which is currently empty and disabled.

Here is the code and I suggest it indicates a need for an improvement in geom_ribbon, and also in the general dialog.

```
# Percentiles plot -----------------------------------
ggplot(data=CIMH_3_by_Year_Month_by_Month, aes(x=as.numeric(month_abbr))) +
  geom_ribbon(aes(ymin=value, ymax=max_sum_rain), fill = "grey") +
  geom_ribbon(aes(ymin=p10_sum_rain, ymax=p90_sum_rain), fill = "blue") +
  geom_ribbon(aes(ymin=p33_sum_rain, ymax=p67_sum_rain), fill = "lightblue") +
  geom_line(aes(y=median_sum_rain), size=2.0, group = "grey", colour="black") +
  viridis::scale_fill_viridis() + theme_grey() +
  theme(plot.caption=element_text(size=8),
        plot.title=element_text(size=20),
        plot.subtitle=element_text(size=15),
        axis.text.x=element_text()) +
  labs(title="", subtitle="", caption="",
       x = "Month", y = "Rainfall") +
  scale_x_continuous(breaks = 1:12, labels = levels(CIMH_3_by_Year_Month_by_Month$month_abbr)) 

```

a) @lilyclements could you look at the current geom_ribbon.  I think it needs the fill parameter added, Maybe both as an aeshetic and a parameter.  It seems the most obvious for ribbon and I wonder why it is missing.  I don't see anything else missing?

image

b) Then there is an "interesting" feature in this plot, namely that the x-axis is a factor, but the geom needs a numeric variable.  That seems to me to be a general point, so perhaps we could add this option into the dialog?  My first go - for comment by @lilyclements is as follows:

image

In this dialog, underneath the x-variable we move the Fill:Colour down and have 2 checkboxes.  Both are invisible initially.  If the x-variable is a factor, then the first checkbox becomes visible (default unchecked) and has label Use as Numeric.  

If checked, then it allows for the as.numeric() in the code above.
If checked a second checkbox becomes visible (default unchecked) with label Display as Factor.  If checked it adds roughly the last line of code above to the command.

MeSophie:
Fixes #8428 
@rdstern , @N-thony  I added `Use as Numeric` and `Display as Factor` checkbox on Graphics dialog. Please check it.

### DIFF
--- a/instat/dlgGeneralForGraphics.vb
+++ b/instat/dlgGeneralForGraphics.vb
@@ -24,7 +24,7 @@ Public Class dlgGeneralForGraphics
     'list of completed layers.
     Private iLayerIndex As Integer
     'current layer
-    Private clsGlobalAesFunction, clsAsNumericFunction, clsScaleContinuousFunction, clsLevelsFunction As New RFunction
+    Private clsGlobalAesFunction, clsScaleContinuousFunction, clsLevelsFunction As New RFunction
     Private clsBaseOperator As ROperator
     Private strGlobalDataFrame As String
     Public bDataFrameSet As Boolean
@@ -47,6 +47,7 @@ Public Class dlgGeneralForGraphics
     Private clsScaleFillViridisFunction As New RFunction
     Private clsScaleColourViridisFunction As New RFunction
     Private clsAnnotateFunction As New RFunction
+    Private clsDummyFunction As New RFunction
 
     Private Sub dlgGeneralForGraphics_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
@@ -88,16 +89,9 @@ Public Class dlgGeneralForGraphics
         ucrReceiverX.SetValuesToIgnore({Chr(34) & Chr(34)})
         ucrReceiverX.bAddParameterIfEmpty = True
 
-        ucrChkUseasNumeric.SetParameter(New RParameter("check", 4))
-        ucrChkUseasNumeric.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
-        ucrChkUseasNumeric.SetText("Use as Numeric")
-        ucrChkUseasNumeric.SetLinkedDisplayControl(ucrChkDisplayasFactor)
-        ucrChkUseasNumeric.SetRDefault("FALSE")
-
-        ucrChkDisplayasFactor.SetParameter(New RParameter("check", 5))
-        ucrChkDisplayasFactor.SetValuesCheckedAndUnchecked("TRUE", "FALSE")
-        ucrChkDisplayasFactor.SetText("Display as Factor")
-        ucrChkDisplayasFactor.SetRDefault("FALSE")
+        ucrChkUseasNumeric.SetText("Group X")
+        ucrChkUseasNumeric.AddParameterValuesCondition(True, "group", "True")
+        ucrChkUseasNumeric.AddParameterValuesCondition(False, "group", "False")
 
         ucrFillOrColourReceiver.Selector = ucrGraphicsSelector
         ucrFillOrColourReceiver.SetIncludedDataTypes({"factor"})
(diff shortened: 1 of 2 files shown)


rdstern:
@MeSophie I think this change is easier than the code we started with.  I need to check with @lilyclements, but suggest you keep the first checkbox and call it Group (or Group X - I am not sure.  The second checkbox and code is deleted.  
Then I show the code you had, and my change, which I prefer - found it on stack overflow of course!
Adding the group parameter does all we need!

```
# Dialog: Graphics

##last_graph <- ggplot2::ggplot(data=survey, mapping=ggplot2::aes(x=as.numeric(variety), y=yield)) + ggplot2::geom_line() + theme_grey()
last_graph <- ggplot2::ggplot(data=survey, mapping=ggplot2::aes(x=variety,group=1, y=yield)) + ggplot2::geom_line() + theme_grey()
data_book$add_object(data_name="survey", object_name="last_graph", object_type_label="graph", object_format="image", object=check_graph(graph_object=last_graph))
rm(list=c("last_graph", "survey"))
```
I have also made a new issue, on further possible improvements for this dialog, and suggest we now close this one, once this small change is merged.

MeSophie:
> I think this change is easier than the code we started with. I need to check with @lilyclements, but suggest you keep the first checkbox and call it Group (or Group X - I am not sure. The second checkbox and code is deleted.
> Then I show the code you had, and my change, which I prefer - found it on stack overflow of course!
> Adding the group parameter does all we need!

@rdstern I notice also that on your code you added `geom_line`. Will this geom_line function be added at the same time as the group parameter when the box is ticked?

MeSophie:
@rdstern Please can you check the new change.

rdstern (changes requested):
@MeSophie this works, but the group=1 is added at the wrong time.  If I do group before chooseing the geom, then it doesn't add group=1 into the code.  If I untick, after getting the geom, and then tick again, it is fine.  The action of ticking should add the code group=1 whatevwer the geom.

MeSophie:
@rdstern I think it is okay now.

rdstern (dismissed):
@MeSophie this looks fine now.  @lloyddewit there is more to improve on this dialog, but I would be happy to see this merged first.  And the corresponding issue can then be closed.

instat/dlgGeneralForGraphics.vb
```diff
@@ -237,7 +257,34 @@ Public Class dlgGeneralForGraphics
...
 
-    Private Sub AllControl_ControlContentsChanged() Handles ucrReceiverX.ControlContentsChanged, ucrVariablesAsFactorForGraphics.ControlContentsChanged, ucrSave.ControlContentsChanged
+    Private Sub AllControl_ControlContentsChanged() Handles ucrReceiverX.ControlContentsChanged, ucrVariablesAsFactorForGraphics.ControlContentsChanged, ucrSave.ControlContentsChanged, ucrChkUseasNumeric.ControlContentsChanged
         TestOKEnabled()
     End Sub
+
+    Private Sub AllControl_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrVariablesAsFactorForGraphics.ControlContentsChanged, ucrSave.ControlContentsChanged, ucrReceiverX.ControlContentsChanged, ucrChkUseasNumeric.ControlContentsChanged
+
+    End Sub
+
```
lloyddewit, line 267: This seems to be a duplicate. Delete?

instat/dlgGeneralForGraphics.vb
```diff
@@ -237,7 +257,34 @@ Public Class dlgGeneralForGraphics
...
+    Private Sub VariableXType()
+        ucrChkUseasNumeric.Visible = False
+        If Not ucrReceiverX.IsEmpty Then
+            clsGlobalAesFunction.AddParameter("x", ucrReceiverX.GetVariableNames(False), iPosition:=0)
+            If ucrReceiverX.strCurrDataType = "factor" Then
+                ucrChkUseasNumeric.Visible = True
+            Else
+                ucrChkUseasNumeric.Visible = False
+            End If
```
lloyddewit, line 276: simplify?
```suggestion
            ucrChkUseasNumeric.Visible = ucrReceiverX.strCurrDataType = "factor"
```

lloyddewit (commented):
@MeSophie looks good, just a couple of small suggestions

lloyddewit (approved):

lloyddewit:
@rdstern if you can retest/approve, then we can merge, thanks

rdstern (approved):
@lloyddewit this still looks fine

PR #8410, opened 2023-06-29
Added an option to rename columns in the Use Table dialog

anastasia-mbithe:
Fixes partially #8236 
This is not ready for testing. Working on fixing some bugs.

Issue #8236, opened 2023-03-23
More improvements in the Describe > Tables dialogue


rdstern:
@anastasia-mbithe and @lilyclements This is coming on well now that the reordering of the variables and the themes have been added.

I hope we can keep going on the improvements.  Here are a few simple suggestions, plus some that might take a bit longer.

Here is a 4-way table with Ana's new Excel theme:

![image](https://user-images.githubusercontent.com/11226469/227162911-5fdbdd36-1857-4f32-8f25-547df9aaf7bc.png)

I really like that we have the Excel theme even though it is an example of a poor table.

Here is the dialogue for the table above, with Example 1 from the `agriTutorial` package - that you will need to install if you would like to use the same example.

![image](https://user-images.githubusercontent.com/11226469/227163716-3f7d5622-40e2-4d40-86f6-fda63a1274f3.png)

Some simple suggestions first:
a) Please check the order of the factors in the first receiver.  I assumed that they would go in the order Replicate, then Management for the columns, followed by Nitrogen then Variety for the rows - that's to give the layout above, which corresponds to the textbook layout.  I don't have a serious problem with your current order as long as there is a simple logic to explain it.

b) There is just a single summary.  That's given as mean-yield in the table above.  When there is only a single summary could we have an option to not give it at all?  Or perhaps no option, but just don't display it?  Or it becomes a default totle or footnote?

c) I think the checkbox to Display Outer Margins now displays all margins - which is great.  So (if I am right), delete the word `Outer `from the label.

d) Similarly simplify Display Summary-Variables as Rows to `Display Summaries as Rows`.

Now how could I dictate that all data in the table above are shown to one decimal?  I think that may be easy through the gtsummary package.

e) If so, then add `gtsummary`.  Then I think it is `style_number` in that package.

f) Investigate adding the gtsummary theme - which seems to have some sub-themes and also permits the tables in different languages!

g) What else becomes easy once gtsummary is used?

lilyclements:
@rdstern to a - 
The order of the factors is be coming in from the `mmtable2` code.

We have two bits of code here - our `summary_table` code is first, and then the `mmtable2` code is second.

```
# Code generated by the dialog, Frequency/Summary Tables
summary_table <- data_book$summary_table(data_name="rice",
                                         columns_to_summarise="yield",
                                         factors=c("management","Replicate","variety","nitrogen"),
                                         treat_columns_as_factor=FALSE, summaries=c("summary_mean"))
[...]

PR #8450, opened 2023-07-18
Added an option to rename Table Columns

anastasia-mbithe:
Fixes partially #8236 

Replaces #8410
This is ready for testing.


rdstern (commented):
@anastasia-mbithe this is great, and I'm delighted you are back.  
 This already seems to work ok.  I don't see why it seems linked only to the Use Table dialog.  There should be just a single Format Table sub-dialog, so it should be automatically available everywhere, and particularly on the main Tables dialog.  Then we should soon be able to merge.

anastasia-mbithe:
@rdstern, we had a meeting with @Patowhiz during the sprint concerning that. For the grid to be able to read the columns, it could work well for an already existing table object. 
Having it in the other dialogs was something we could look into later. After determining, how best to have it since the columns are kinda detected by these options for the other dialogs. 
image

rdstern:
@anastasia-mbithe I am using the new dialog.  It seems to work - so that's great.  I am trying with a table of climatic summaries of the same type that @lilyclements used for a graph recently.  I'll give the data below.  

a) I just suggest it should be wider to make it easier to use.  Then it could be merged perhaps?
b) I found it under columns - which is correct for this table.  I thought it was elsewhere when I looked before?

image

c) I understand your reasons for just having it in the Use table dialog.  I still hope there is only a single (sub-)dialog for all the tables.  Then (for now) this option could be disabled.  Later, I assume as we make a table object anyway, it could be implemented in all dialogs, but just after the object is created.  

d) It seems to work, except for the last 2 entries.  I thought I did them all? 
image
 So I went back to the new table it had just made - and your sub-dialog, to find it had the last summary I had renamed, but it had not included.  And it has forgotten the earleir summaries thast it had included!

image

Here are the data I used:

[dodoma_tables.zip](https://github.com/africanmathsinitiative/R-Instat/files/12116635/dodoma_tables.zip)

I am also looking forward to also being able to control the number of decimals - here I would like none for the whole table.  And also perhaps having columns in different colours, as is easy in Excel.

anastasia-mbithe:
@lilyclements, I have been looking at the last issue mentioned above with @N-thony on reading the renamed table columns on the grid and we thought its a limitation of the `get_object_data` function. But it seems its a limitation of the ` gt::cols_label() ` 

I have tried with the below code and checked the names of the last table, still gives the original names. It can read Column names as column labels but not vice versa. Any suggestion?
`library(dplyr)
data("mtcars")

my_table <- mtcars %>% gt::gt()
my_data<- as.data.frame(my_table)
names(my_data)

label <- c("cylinder", "horsepower")   # value from the "new column name"
columns <- c("cyl", "hp")   # value from the "(current) column name"
cols_list <- as.list(label) %>% purrr::set_names(columns)
cars <- mtcars %>% gt::gt() %>% gt::cols_label(.list = cols_list)
cars
cars<- as.data.frame(cars)
names(cars)`

lilyclements:
Before this is taken over, I just want to check it's current state. We can then see what needs to be done next

lilyclements:
I think this can be ready to merge, and then we can make new amendments if we are happy with that?
@MeSophie, once merged I suggest we have a more in depth discussion about amendments to make to the tables.

One comment for changes in the future, is that if the "Themes" tab is opened on the sdg, we get a theme run even if we don't want to. We want to have by default no theme running. So can we add an rdo that is "No theme", which is selected by default. If this is selected, then no additional code is run .
I'll add this as a new issue. 

To answer the point made by @anastasia-mbithe, the change is that we want to do `cars$`_boxhead`$column_label` to view the variable names. But I think we should worry about this later.

```
library(dplyr)
data("mtcars")

my_table <- mtcars %>% gt::gt()
my_data<- as.data.frame(my_table)
names(my_data)

label <- c("cylinder", "horsepower")   # value from the "new column name"
columns <- c("cyl", "hp")   # value from the "(current) column name"
cols_list <- as.list(label) %>% purrr::set_names(columns)
cars <- mtcars %>% gt::gt() %>% gt::cols_label(.list = cols_list)
cars$`_boxhead`$column_label
```

rdstern:
@lilyclements this looks like a nice control to add.  I tried a simple example to see what is possible:

image

I then went into the Use Table dialog - because that is where it is.  Here is the current Columns dialog:

image

a) I quite like the control.
b) Nice also to have the alignment also.  However, in general I assume we may want the alignment to depend on the column.  So this might add to the rename control.  Perhaps both are needed, so if the alignment is for the whole table, then it is a single value.  However, perhaps there will also be a Table tab and overall alignment could then go there?
c) There are other things, particularly format, for the columns.  
d) An alternative structure is to have a Rename tab.  For example it would be good to be able to rename the summaries as 
well, at least in the table above.  
e) The first column above is one I can rename, but it doesn't appear in the table.

I then made some changes to the columns:

image

I then ran:

image

Oh it worked this time.  Last time I did it on a names table - don't know if that fact is relevant - and it didn't work!

I then went back into the dialog to make more changes.  But it has forgotten everything I did!  It is back to here:

image

So, all very useful to give us ideas, but not something to merge yet.  I suspect we will want to keep and build on the control, and include it in the main table dialog. Let's see.

Patowhiz:
@rdstern @N-thony who should we delegate this PR to? I understand some progress was made here but the issue is not yet fully solve.

rdstern:
@Patowhiz I am a bit lost on this one which is another important element of the Format Table set of sub-dialogs.  I think it may need input from @lilyclements and I suggest it is no longer urgent for this update.  It should fit somewhere into the sub-dialog and Lily did explain before why it was only in the Use table dialog and you couldn't do it earlier, but that is lost in the mists of time now.

I suggest the effective use of gt for tables output is the most important improvement we still need in R-Instat and we should continue the steps on that for the december update.  But I give in for November!

rdstern:
@Patowhiz is there anything here that is still useful, or should we close this pull request?

Patowhiz:
@rdstern let's close this. We already have it implemented in my PR

Issue #8451, opened 2023-07-19
Fix two recent regressions on the graphics sub-dialog


rdstern:
@N-thony I suggest these are both easy to fix and could be by anyone?  It may be worth a bit of detective work, at the same time, to see how they have arisen, in case there are more that we have yet to uncover.

a) This one seems to do no harm.  It has come after Version 0.7.15.  It has added 2 lines of exta text, namely `2 digit` after the first factor and a long text after the second!   Why has this happened and where have they come from?

<img width="512" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/7ac00c4f-aec0-415e-b908-147813b8fbf8">

b) The legend, see below:

<img width="350" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/365e3d98-c1fb-4145-a59e-66e937affcbb">

This is a problem.  The default used to be `None` and this has disappeared.  This was a problem I didn't notice in 0.7.14 and 15, but I think was ok before.

c) I don't know if this is related to b) In the pair-plot option, see below, I thought the pair-plot options were now enabled.  They may be linked to the legend problem above?  Maybe they get enabled later when using the dialog?  This was pull request #8005 and it mentioned the legend.

<img width="326" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/ea6da1d9-3d49-4385-9afb-eb2228897873">

lloyddewit:
@rdstern @derekagorhom 
For part `b)` above, the `None` option was deliberately removed in PR #8316. Was this correct?

![image](https://github.com/africanmathsinitiative/R-Instat/assets/57253949/c314f2bc-979e-4054-9006-598fa10dafde)

derekagorhom:
@lloyddewit 
Roger requested that change in issue #8288
see here 
![image](https://github.com/africanmathsinitiative/R-Instat/assets/76208189/7362c5bb-65cd-4466-b661-793c937a96e2)

PR #8456, opened 2023-07-21
Ensured that factor labels in Plot Options subdialog are correct

lloyddewit:
Fixes part a of #8451.
In the script window, I also changed the `Add` button to `New`.

@rdstern Please can you test?

Note:
The problem was caused by incorrect translations for these 2 labels for several languages in the translations database.
The JSON files and CrowdIn translations are all correct so the problem must have been introduced when I imported the JSON files into the database. I added a regression test so that if this happens again, then I should detect it before merging the new database into master.

### DIFF
--- a/instat/ucrScript.Designer.vb
+++ b/instat/ucrScript.Designer.vb
@@ -271,7 +271,7 @@ Partial Class ucrScript
         Me.cmdAddTab.Name = "cmdAddTab"
         Me.cmdAddTab.Size = New System.Drawing.Size(75, 23)
         Me.cmdAddTab.TabIndex = 4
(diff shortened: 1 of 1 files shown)


rdstern (approved):
I am approving, because the text seems fine now and I am also happy to have the New rather than Add.  

The only problem I could find when testing is - I think - not related to your changes.  I tried with the usual survey data and the bar chart dialog.  It worked fine.  But when I go back in, it reports a developer error.  

image

There are changes being made to the bar chart dialog that I think has introduced this.  It should be reported there (as well?) And I'll do that.

It is a message I report with different dialogs reasonably regularly.  It does no harm as the dialog then runs fine.  But it looks unprofessional.

Issue #8217, opened 2023-03-15
Add a Find Columns dialogue into the Edit menu


rdstern:
@N-thony you are working on the Find dialogue now.  There is becoming more work, so you may wish to allocate or share this work.

I suggest what you are working on becomes `Find Rows`, instead of just Find.  We don't need Find Next in the menu and I suggest we replace it by `Find Columns`.  This will again have 2 options and the select is the second - just like Filter is the second in the Find rows.  And it will indicate a column, just as the Find Rows indicates a row.

I strongly suggest the Select option be done first, because the ordinary option will take longer, and will essentially be a copy of the Select code.  It will then jump to the correct "page" sideways, just are your current find jumps to the right page going downwards.

rdstern:
@N-thony the Find is now working, though I think it may need a tweak.  I suspect that, by default it is using regula expressions, rather than the other way round.  I now suggest - for simplicity - we add to this dialog, rather than having a new one.
a) Change the name at the top of the dialog to `Find Rows or Columns`.
b) Add a third button, at the top, which is called `Select`. 

Then, as in the re-opened Find #7950 

The multiple receiver is just for Selects.  as in the Rename Objects dialog:

<img width="173" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/ecdaa2ae-02bb-44d8-ac95-e48aa216adb8">

Then the single receiver is just labelled Select:

then just the same as in #7950.

I think that should be pretty easy - at least for you?

PR #8459, opened 2023-07-24
Implemented find feature in Filter option in the Find dialogue

N-thony:
Fiexs partially #8217
@rdstern I have enabled and implemented the `Filter` option, have a look. If it is okay, we can merge it and will continue with the `Select` option.


rdstern (approved):
@N-thony great, and it is now listed in the last dialogs too.
@lloyddewit I am happy if this can be merged.  There is then another step, which is to find columns!  This could be a further pull request.

lloyddewit (approved):
@N-thony Looks good, thanks

PR #8462, opened 2023-07-26
Improved Describe>General menu and Describe<Graph options

### LINKED ISSUE #8421: Possible changes to the Describe menu
I am becoming increasingly comfortable with the describe menu.  It is reasonably short and pretty clear.
I need others, particularly @volloholic to check my suggestions below.  I'd also welcome comments of course from anyone else.

a) This has arisen, because I have, at last, found a simple graph that is very easy to do with the `Describe > General > Graphics` dialog. 
 And it doesn't seem possible with any of the options in our `Describe > Graphics` dialogs.  

It is described [here.](https://ggplot2.tidyverse.org/reference/scale_discrete.html) .  

b) This Describe > General > Graphics dialog is much more important than its current position in the menu.  Hence I propose the 2 following changes:

1) Here is the current Graphics menu:

image

I propose to get rid of the line below heatmap.  So there are just 2 sections so far.  They correspond to common graphs and particular graphs.  Later we will, I hope include maps in the heatmaps dialog.  That will change the name and move it above the line.

2) Then a line underneath and one entry, called General, which is the current Graph dialog that is currently hidden in the Describe > Graphics > General menu.

3) Then we keep the Describe > General menu item
We delete The Column Summaries dialog from here and also the Use Summary dialog.  It is currently a pretty feeble version of the bar chart dialog.

a) We add a copy of the Pivot-table dialog.  Just the same as in the Describe > 2/3 Variables > Pivot Table.
b) Then a horizontal line
c) Then keep the General graph dialog there.
d) Add a copy of the Tables dialog.

Fidel365:
Fixes partially #8421 
@rdstern this fixes part (b) of the issue and it's ready for review

### DIFF
--- a/instat/frmMain.Designer.vb
+++ b/instat/frmMain.Designer.vb
@@ -66,17 +66,13 @@ Partial Class frmMain
         Me.ToolStripSeparator26 = New System.Windows.Forms.ToolStripSeparator()
         Me.mnuDescribeSpecificMapPlot = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuDescribeSpecificDotPlot = New System.Windows.Forms.ToolStripMenuItem()
-        Me.ToolStripSeparator27 = New System.Windows.Forms.ToolStripSeparator()
         Me.mnuDescribeSpecificMosaic = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuDescribeSpecificCummulativeDistribution = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuDescribeSpecificParallelCoordinatePlot = New System.Windows.Forms.ToolStripMenuItem()
         Me.mnuDescribeSpecificTables = New System.Windows.Forms.ToolStripMenuItem()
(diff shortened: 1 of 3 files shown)


rdstern (changes requested):
@Fidel365 that's almost perfect.  Could you please just add a horizontal line before the Graphics.  Then we can merge.

Fidel365:
@rdstern , I have made the change for part (b), for part (a), I need some clarification from you, For parameter x, we will need a selector and a receiver that x gets the variables from. Secondly, we should create a new tab in plot options options, i.e. after the annotation option, where we should have a selector and receiver in that tab  plus a grid which contains factor, label and limits as shown below;
Screenshot (37)
Can we implement the changes this way?

rdstern:
@Fidel365 please I have misled you.  The part a) is now part of the discrete option in the axes sub-dialog that Vitalis is working on.  You don't need to include it.  I think this pull request is now complete.  I'll check again.

rdstern (approved):
That's fine.  @lloyddewit over to you.

lloyddewit (approved):

Issue #7161, opened 2022-01-24
Load and manipulate wide files in R-Instat more efficiently?


rdstern:
[ZZHR62FL.zip](https://github.com/africanmathsinitiative/R-Instat/files/7929163/ZZHR62FL.zip)

This is an spss file that reads fine into R-Instat.  It is about 3 mbytes, but is 50 mbytes as an spss sav file.

It has about 6000 records (rows) and over 7000 variables (columns).

When reading into R-Instat there is initially the usual screen that it is taking some time.  Then that stops as I assume it is getting to the last steps, which is to load the data into the R-Instat data book.  However this last step takes a long time and it looks as though R-Instat is frozen up.

I gather from @dannyparsons that this step can be done much more efficiently and that this should be quite easy to implement.  Possibly we should also include the waiting screen during this last step.  I am used to it taking (say) 10-20 seconds to complete and that is fine.  I have not timed, but suggest this is at least 10 minutes currently.

We might also wish to consider whether there is a maximum that we allow?  Perhaps this can - like other limits, and defaults, then be changed in the Tools > Options dialogue.  There isn't a problem in R, having (say) 1 million variables, perhaps by transposing a data frame with a million rows.  There is a problem loading these into a dialogue, and perhaps also into a grid?  The maximum in reogrid is 32,768.  This may not apply as we still only show a small subset, but that might also be a sensible limit for us in R-Instat.  It also has a limit of 1 million rows and that certainly doesn't apply to us.  (I have tried with our new grid system and 3 million rows.

dannyparsons:
Tagging @Patowhiz instead of assinging.

Patowhiz:
@rdstern @dannyparsons I imported the same data using R-Instat functions in R studio and below is a benchmark comparison of the 2 softwares in my machine.

Import command;
```
ZZHR62FL <- rio::import(file="C:/Users/LENOVO/Downloads/ZZHR62FL/ZZHR62FL.SAV")
data_book$import_data(data_tables=list(ZZHR62FL=ZZHR62FL))
```
Bench mark time;

| Trial | RStudio  | R-Instat |
| -------------| ------------- | ------------- |
| 1 | 7.480  | 5.876 |
| 2 | 7.416  | 5.181 |
| 3 | 7.449  | 5.125 |
| 4 | 7.456  | 5.540 |
| 5 | 7.393  | 5.820 |
| 6 | 7.449  | 5.546 |


[...]

PR #8465, opened 2023-07-27
Improved performance when importing wide datasets (and fixed bug when selecting multiple variables)

Patowhiz:
Fixes some items in issue #7161. Fixes #8286


rdstern (commented):
@Patowhiz that seems extrmely promising already.
a) I tried the ExtRemes > carcasson data from the library, which has 12,000 variables.  It now took just 23 seconds to import into R-Instat.  Great.
b) I then tried the Prepare > Data: Change > Transpose.  It took just 3 seconds to open and load the data selector.  That was (I think) fixed before.
c) I then added all into the data receiver.  That took 150 seconds.  I suspect that might be quite an easy fix?  Could that also be done in this pull request?
d) Then I pressed Ok.  It then took another 23 seconds to write the transposed data frame.  I was pleased initially, but I now wonder why that takes so long.  It is just writing a data frame of 4 variables and 12,000 rows.
e) Then I transposed back again.  How long does R-Instat take to write a data frame internally with 12,000 columns.  That took only the 23 seconds again - which is just great.
f) I then tried to see what - if anything you have done when I ask for the metadata.  This freezes up, and I have to restart R-Instat.
g) When I restart I was pleased to see that it works fine on running the log file.  

This is soooooooooooo promising.  Very exciting indeed.

To report further, there is a free-to-download dataset from DHS.  This has 8 data frames, one of which has 7654 variables (6290 rows) and it has others of 4200, 2200 and 1000 variables.  The rds file took about 31 seconds to open initially.  That is acceptable.

Patowhiz:
Like I explained before. The column metadata needs the "paging" structures similar to the data viewer. That's not quite trivial to do correctly, but very doable. I just need to countercheck with @ChrisMarsh82 about my ideas.
I've done a temporary fix with a todo comment. The fix loads the column metadata only when it's visible. 
The long term fix is to load only 1000 visible columns. This fix should also b used by the dialogs somehow.

I've been looking into the dialogs issue as well. I'll post my changes for testing once I'm done.
I'll also look at some of your items above.

Patowhiz:
@lloyddewit @ChrisMarsh82 when enhancing the performance of the multiple receivers I noticed the following;

1. Multiple Receivers are assumed to be capable of being attached to more than 1 data frame, that is, every variable can have a different data frame.
2. Selector uses receiver data frame to set the selector currently selected data frame in some events.
3. We are always transforming data from one format to another between the receiver and selector even though we are using a list view control in both  controls.
4. It is also assumed that the selector can have variables that belong to multiple data frames in the same dialog at the same time.

I have not yet seen any dialog which loads variables from more than 1 data frame in different options at the same time. In dialogs each selector is associated with 1 data frame. And each receiver is always associated with the data frame selected in the selector. @rdstern is there any dialog that deviates from this 'principle'.

I wasn't able to understand why @volloholic and @dannyparsons made such an implementation, I think this is unnecessary complication which has led to much more complexity in the design of selector and receiver.
@rdstern once I have answers to the above 4 items, I can make changes that have significant performance difference. I don't want to spend more time refactoring the selector and receiver on a feature I haven't seen used anywhere yet.
This PR focused on optimizing existing code implementations, to make them optimal for wide data sets.

On a separate note, I noticed the transpose dialog (dlgTranspose) uses the selector and receiver in an 'unconventional' way. It has to be refactored to make it use the performance gains made in this PR. I suggest @N-thony to have a look at it after this PR is merged.

Patowhiz:
@rdstern @lloyddewit below are my performance tests. I had 2 chrome windows, each with multiple tabs open, visual studio and calculator running.

Before this PR, in 1 trial it took (considering the time taken, I didn't have the time to run a second trial);

- 960 seconds to load the CarcasonneHeat data from extremes package
- 10 seconds to load the one variable summaries dialog.
- 370 seconds to add all the variables to the multiple selector of one variable summaries dialog.
- Over 6,000 seconds to reload the one variable summaries dialog. I had stopped the application.

I didn't bother proceeding with trial, it was clearly a waste of time

In this PR, in 2 trials it took;

- Average of 37.5 seconds to load the CarcasonneHeat data from extremes package
- Average of 6 .5 seconds to load the one variable summaries dialog.
- Average of 12.5 seconds to add all the variables to the multiple selector of one variable summaries dialog.
- Average of 25.0 seconds to reload the one variable summaries dialog.
- Average of 11.5 seconds to load the transpose dialog.
- Average of 15.5 seconds to add all the variables to the multiple selector of transpose dialog.
- Average of  6.5 seconds to reload the transpose dialog.

This shows exponential performance gains after optimizing the existing code despite the overengineering of the controls.

@rdstern you are welcome to do your own test trials using the 3 dialogs mentioned above.

Patowhiz:
@rdstern thanks for testing this. This is promising, considering it was very difficult to test such wide data sets before this PR.

 I spent quite a considerable amount of time (4 days) looking into this issue in a general way. Most solutions that I came up with risked regression in areas that aren't clearly understandable. I'm tempted to not loose the momentum (despite my many other tasks).

If a message is all you need, then that sounds as a good quick task for @N-thony.

I did write some prototype code that could have allowed me to support paging of the metadata. I now wonder whether I should spend more time in crafting something concrete for the August release.

Kindly have a look at the issues I have raised above as well. Would be good if you could get feedback from @volloholic as well, I still don't quite understand why we needed that complexity. I recommend removing it now when I understand it's pitfalls well enough.

The base code that supports paging of column metadata should be usable with columns selects. Basically, the selector, receivers and column metadata will probably use the same data source. So solving the paging feature is a step towards solving the column select feature.

I totally agree we definitely need to support and gain experience in developing for wide datasets. I can see them being increasingly adopted in this age of abundant computing power and AI tools.

I'm very optimistic that the output window can in future comfortably handle large outputs, it's design already supports addition of such implementation.

rdstern:
@Patowhiz If you could do the paging the metadata window quickly, that would be great.  Otherwise,  it would be good to write an issue while this is all fresh in your mind.

rdstern:
@Patowhiz and @ChrisMarsh82, you asked:

1. Multiple Receivers are assumed to be capable of being attached to more than 1 data frame, that is, every variable can have a different data frame.
2. Selector uses receiver data frame to set the selector currently selected data frame in some events.
3. We are always transforming data from one format to another between the receiver and selector even though we are using a list view control in both controls.
4. It is also assumed that the selector can have variables that belong to multiple data frames in the same dialog at the same time.

I have answers that relate at least to some of these questions:
1. A multiple receiver can (in principle) have data from different data frames.  

I do not know of any instance where we currently use this feature.  It is important for future plans.  It should enable dialogs (when needed, and in linked data frames) to analyse data at multiple levels, without them needing to be merged to all be at the same level.

An example (though not in a current multiple receiver) could be a water-balance calculation where you calculate 
 water balance = (rainfall - evaporation), between limits of say 0 and 100.  The evaporation is sometimes the average over the years, so just 366 values - in a special data frame, while the rainfall is for individual years.  Currently you have to merge the evaporation column first.  

So, you are welcome to have a switch for now, that simplifies (and hopefully speeds up) the work when thism option does not apply.  But don't get rid of the option completely.

Your point 4 is:
4. It is also assumed that the selector can have variables that belong to multiple data frames in the same dialog at the same time.
Most dialogs that use 2 data frames have 2 separate selectors.  The climatic > crops dialog is a rare exception that has a single selector, used at the daily level and also (for other single receivers) at the summary level.
The new define climatic data may be a second example.  Here most of the information is for a daily data frame, but (for the evapotranspiration dialog) we include lat and long now, and that comes from a station level.

If it would really help to avoid this, then we could have 2 selectors in these cases?

I don't quite understand your points 2 and 3.  Maybe @ChrisMarsh82 can help?

I hope this helps a bit?

rdstern (approved):
@Patowhiz this all seems to work now.  If @lloyddewit agrees, then I would be very happy to see this merged now, with another issue on what remains oustanding.
I tried with 1 medium-width example (1000), variables.  Then (initially accidentally) I opened a wide one 7400 (1 min 10 seconds) and 12,000 variables (2mins 30 seconds) when I had the metadata window open.  They looked fine!
Then I did them in 2 steps, so opening and then adding the metadata window.  
Here is the picture while the metadata window is being added:

image

It looks odd, but not terrible.

So, if merged, then I suggest an additional issue to be considered soon, perhaps for the September upgrade, as follows:
a) @Patowhiz said that the transpose dialog has different code, and maybe @N-thony could improve that.  This is an obvious dialog when dealing with wide data sets, so perhaps that can be done soon.
b) Discuss and probably install a limit in the Tools > Options menu.  What happens with (say) 50,000 variables, etc.  R doesn't have a particular limit.
c) Consider a message and perhaps the be patient dialog, when opening the information for the first time.

Patowhiz:
@rdstern I have now added the warning message. If you press no then you will only see an empty column metadata for the data frame that has wide data sets only. The warning is also shown if a wide data set is detected.

The oddity you are referring above is due to how visibility toggling of the "windows" is currently implemented. I think it was expected that the visibility toggling will happen very fast which in this case it's not due to the many things happening in the background in a single thread.  
I also noticed the oddity when you initially open R-Instat. I'll quickly have a look and see if I can correct the implementation.

rdstern (changes requested):
@Patowhiz I started checking the latest changes, and then found a more serious bit of regression.  The Right-click on data frame > Delete now seems to freeze up.  I think the ordinary menu item also doesn't work.

It is fine in the latest merged version.  But in your branch it seems to be stuck here?

image

I also have comments about your warning thing, but wanted to make tis response first

rdstern (approved):
@Patowhiz this seems fine and I am therefore approving.  
I wonder what your lower limit is, that ask?  I tried with 8 files simultaneously that is the free set from DHS and so a good one to show and use.  It asked about all 3 of the widest, which have CR: 2262, HR 7654, IR 4275 variables.  I tried saying Yes to everything and it was reasonably quick - but worth having the warning.
I then tried with yes to CR 2262, and No to the wider ones.  Your limit is clearly less than 2262 and I wonder about making it higher?  For example 3000 or even 4000 variables?

Apart from that - and the probably unrelated problem with the Delete data frame it is looking great.  I am approving though I worry slightly about the Delete problem.  Are there other vulnerable points if we merge?

rdstern (changes requested):
@Patowhiz I have now found a small problem with the warning and that is when it is given.   I am getting the warning too often.
a) I said no the first time for the 2 wide data sets.  That was fine.  Worked well.
b) I then closed the metadata and later I opened it again.  It asked me again about those 2 sheets.  That's questionable, but I suppose ok.
c) Then, when the metadata was open I changed a numeric to a factor variable on one of the small files (just 7 variables).  And it asked me yet again.  That's too often and when I was working happily on a small file.  
So, at the most it could ask repeatedly when I re-open the meta-data.  I would also not mind, if it only asked the first time.  But I should be able to work with the metadata window open without it ever asking again.

Patowhiz:
@rdstern thanks for testing this. I have now fixed the issues you raised above. 
My limit is actually 1000 variables. The decision behind that number is based on; 
1. The data viewer default rows limit is 1000. We have tested this to know there is no performance issues.
2. The copy feature limit is 1000 rows. We have tested this to know there is no performance issues.

I can't guarantee if there will be no performance issues beyond that number when filling the grid. If you strongly want more than 1000 I can quickly change the hard coded figure.

Regarding the prompt, I added an extra response button as a bonus. So you users can chose to not give a definitive answer by pressing 'cancel'. I'm happy to change that implementation.

Patowhiz:
@rdstern I have updated (and will regularly update) this branch with all changes from the master. I suggest you use it in your normal analysis work when not reviewing other PRs. This will help with knowing any breaking changes before it's merged to master.
Thanks.

rdstern (approved):
@Patowhiz I tried with the 8 DHS files, where 3 had over 1000 variables.
a) I checked on Delete data frames and that seems fine now.
b) When I ask for the metadata, the question has no gone to the other extreme.  I either get thyem all or none.  And I am just asked once.  I am happy with that.  
c) But I don't see the use/need for the Cancel button.  When I click on Cancel it asks me 3  times.  I suppose, if I know my data, I can assume it is asking in the order of the data frames.  So, it is effectively a No for each large dataframe in turn.    So I could cancel the first and then say Yes the second time I suppose. No - I can't see any reason for asking me 3 times.  Then, if it only asks once, perhaps the next time I open the metadata, then it will ask again.  But it doesn't.
So let's keep it simple and not have the Cancel button?
d) I am happy to keep testing, through using it, but I would also be very happy for it to be merged - if @lloyddewit is ok with it.

Patowhiz:
@rdstern I have removed the cancel option. I added it because I thought a user may want to view the column metadata later.
The correct implementation would be to track response for each data frame but I wasn't sure if spending time in crafting that solution would be a good use of time. After paging implementation, we won't even need the message prompt.

rdstern (changes requested):
I am happy with this change.  But I crashed R-Instat in the following way.  I wonder now if we are lucky that I tried this combination!
a) I first report that it asks me about the column metadata, when I only asked for the data frame metadata.  I don't really mind that.  I just said no and it gave me the result I wanted.
b) But the breaking problem is serious and perhaps obvious?  I said no and so I only have the column metadata for the not-wide data frames.  Then (as I had imported) I now decided to save the data book as an rds file.  And it crashed.  In hindsight that's perhaps obvious, because it didn't have the metadata.  So it couldn't write the data book out.  

I don't know what you want to do about this?  I don't want things to get too complicated.  And please read c) before deciding what to do. 

c) Next I started with the same data, already saved as an rds file.  I should have tried that a long time ago.  It only took 36 seconds to import compared to 1 minute, when I import.  And it is much smaller of course.  
Then I asked for the metadata.  It didn't give your warning at all,  So it seemed to freeze up again.  Then, of course, it all appears, with all the metadata, after the long wait.

So perhaps we don't give any choice when a data file is imported?  (Effectively we must say yes!>)  And we give the usual it is taking a long-time message.  Can you suggest better than that?

Patowhiz:
@rdstern saving of the data as an rds is done at the R (backend) level. Changes made in this PR are purely at the VB.NET (front end) level. Did you confirm that the saving bug is not in the master as well? I understand loading wide data sets in the master may not be a good idea, considering the time it will take.

I'm surprised why it loaded the data after you responded No. I'll check this and come back to you.

rdstern:
@Patowhiz I suggest it must be linked to your code, because I can't try this feature in the current master.
In the current version I always show all the metadata.  It takes a long time, but I have no other option.
In your pull request I have the option (for the first time) to not show the metadata for the wide data frames.  That's because it takes a long time.

Now I can imagine that when I save as an RDS file it is saving all the metadata, but I opted not to show it.  So perhaps it doesn't have it, in a convenient way to save?

Patowhiz:
@rdstern I've reproduced your bug (crush) when using a wide data set. It crushes even when not viewing the column metadata.
I think the crush is caused by the R command when attempting to save the data book with a "deep" list of changes. 
In R Studio, tried running the following R commands log produced by R-Instat. Notice the error shown below.

**R-Instat Log**
```
# Initialising R (e.g Loading R packages)
setwd(dir="C:/Users/patoe/source/repos/R-Instat/instat/bin/Debug/static/InstatObject/R")
source(file="Rsetup.R")
data_book <- DataBook$new()

# Setting display options (e.g Number of significant digits)
options(digits=4, show.signif.stars=FALSE, dplyr.summarise.inform=FALSE, R.commands.displayed.in.the.output.window=TRUE, Comments.from.dialogs.displayed.in.the.output.window=TRUE)

# Dialog: Import From Library
utils::data(package="extRemes", X=CarcasonneHeat)
data_book$import_data(data_tables=list(CarcasonneHeat=CarcasonneHeat))

# Dialog: Save Data As
saveRDS(file="C:/Users/patoe/Desktop/delete.RDS", object=data_book)
```

**RStudio Error Screenshot**


**Cause of the Bug**
According to [this](https://stackoverflow.com/questions/14719349/error-c-stack-usage-is-too-close-to-the-limit) stack overflow answer the error could be due to recursion or the default R stack size parameter used by the OS.
@volloholic @lilyclements when I commented out the amendments we did on the list that tracks the data book changes in `data_object.R`, the file was saved without an error. I did manage to also import back the file.
See the screenshot below that shows line 1134 commented. I think the list hits the R default stack size limit when tracked changes become many. I'm not sure if that will still happen if we restore line 1134. 

**Source of Bug Code Snippet**


**Thoughts**
@rdstern we replaced line 1133 with line 1134 to enable importing of such wide files. In my opinion, the size or "recursive" nature of the list hits the limitations of R.  I'll now wait for @volloholic and @lilyclements to give me their R suggestions.

The standard computing way of tracking changes is usually through a log file (to avoid such issues). But R being single threaded, I suspect the numerous number of Input/Output operations that will be done here, may affect performance significantly.

A quick fix would be to not track changes at all.

rdstern:
@Patowhiz I have been checking the saving of the wide data in this new version as well as the latest released version.  I have been using the Carcassonheat data from the ExtRemes package.  It is small, i.e. just 4 rows and 12054 variables.
a) I opened with the new (wide) version.  This takes about 29 seconds to open.  That's fine.  I asked for the metadata and said no.  Then I tried to save and it gave the overflow error trying to save the data book.  (That was file called Extremes_wide.)
b) I tried again and this time I said yes to the metadata.  That took about 2 minutes.  I then tried saving and got the same error.  File called extremes_wide2)
c) So I tried opening the same data in the "old" R-Instat (Version 0.7.16).  That took about 12.5 minutes to load from the library, and was painful, in looking as though the software had frozen.  I persisted and then saved.  It appeared to save instantly.  That's extremes_wide3.
d) Now I opened extremes_wide3 in the new test wide version.  It opens fine.  I added some small data frameas (from the survey).  I then tried to save.  It saves fine.  So, it is writing the first data book that is the problem.  Once it has one it seems fine.  That's test_wide4.  I can open it and save it fine with your new code.
e) I now copied the wide data frame.  That takes a little time as one would expect.  I now saved that data book. I called it test_wide_double, as there are now 2 wide data frames.  That also saves and then opens fine in your new test version.
f) I now appended the 2 wide data frames, so I have a new one with 8 rows instead of 4 as well as the original.  Now I get the error again, when saving.

image

I hope this helps.

Here are 2 data books that do open (and save) with your version:

[extRemes_wide_double.zip](https://github.com/africanmathsinitiative/R-Instat/files/12477839/extRemes_wide_double.zip)

rdstern (dismissed):
@lloyddewit this is a major improvement.

instat/ucrColumnMetadata.vb
```diff
@@ -54,10 +63,32 @@ Public Class ucrColumnMetadata
...
+        If dataFrame.clsColumnMetaData.iRowCount > 1000 Then
+            'if not asked or no response before then prompt for a response
+            If bWideDataSetPromptResponse = DialogResult.None Then
+                bWideDataSetPromptResponse = MessageBox.Show(Me, "Are you sure you need wide data set(s) column metadata?  If so, be patient.  It, will be slow to load the first time", "Wide Data Set(s) Detected",
+                                                             MessageBoxButtons.YesNo, MessageBoxIcon.Warning)
+            End If
+
+            'if response is no or no response given then don't fill the worksheet with data
+            If bWideDataSetPromptResponse = DialogResult.No Then
+                bFillData = False
```
lloyddewit, line 81: Should we `Exit Sub` after this line or is line 92 important?

instat/ucrColumnMetadata.vb
```diff
@@ -131,19 +166,19 @@ Public Class ucrColumnMetadata
         Dim clsDeleteLabelsFunction As New RFunction
 
         If strColumnName = strLabelsLabel Then
-            If MsgBox("This will delete the selected label(s) and replace with (NA)." &
+            If MsgBox("This will delete the selected label(s) And Replace Then With (NA)." &
```
lloyddewit, line 169: Do we need capitalized words (inconsistent with first half of message)? Did the VS editor do this automatically?
```suggestion
            If MsgBox("This will delete the selected label(s) and replace them with (NA)." &
```

instat/ucrColumnMetadata.vb
```diff
@@ -131,19 +166,19 @@ Public Class ucrColumnMetadata
...
-            If MsgBox("This will delete the selected label(s) and replace with (NA)." &
+            If MsgBox("This will delete the selected label(s) And Replace Then With (NA)." &
                                 Environment.NewLine & "Continue?",
                                 MessageBoxButtons.YesNo, "Delete Labels") = DialogResult.Yes Then
 
                 clsDeleteLabelsFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$append_to_variables_metadata")
                 clsDeleteLabelsFunction.AddParameter("data_name", Chr(34) & _grid.CurrentWorksheet.Name & Chr(34), iPosition:=0)
                 clsDeleteLabelsFunction.AddParameter("col_names", frmMain.clsRLink.GetListAsRString(_grid.GetSelectedColumns), iPosition:=1)
-                clsDeleteLabelsFunction.AddParameter("property", Chr(34) & "labels" & Chr(34), iPosition:=2)
+                clsDeleteLabelsFunction.AddParameter("Property", Chr(34) & "labels" & Chr(34), iPosition:=2)
```
lloyddewit, line 176: Auto change from VS editor?
```suggestion
                clsDeleteLabelsFunction.AddParameter("property", Chr(34) & "labels" & Chr(34), iPosition:=2)
```

instat/ucrColumnMetadata.vb
```diff
@@ -131,19 +166,19 @@ Public Class ucrColumnMetadata
...
                 clsDeleteLabelsFunction.AddParameter("new_val", Chr(34) & Chr(34), iPosition:=3)
                 frmMain.clsRLink.RunScript(clsDeleteLabelsFunction.ToScript())
             End If
         Else
-            MsgBox("Deleting cells is currently disabled. This feature will be included in future versions." & Environment.NewLine &
+            MsgBox("Deleting cells Is currently disabled. This feature will be included In future versions." & Environment.NewLine &
```
lloyddewit, line 181: Auto change from VS editor?
```suggestion
            MsgBox("Deleting cells is currently disabled. This feature will be included in future versions." & Environment.NewLine &
```

instat/ucrColumnMetadata.vb
```diff
@@ -370,7 +405,7 @@ Public Class ucrColumnMetadata
     End Sub
 
     Private Sub columnContextMenuStrip_Opening(sender As Object, e As CancelEventArgs) Handles columnContextMenuStrip.Opening
-        If IsOnlyOneDataFrameColumnSeleted() Then
+        If IsOnlyOneDataframeColumnSeleted() Then
```
lloyddewit, line 408:
```suggestion
        If IsOnlyOneDataframeColumnSelected() Then
```

instat/ucrReceiverExpression.vb
```diff
@@ -134,24 +126,24 @@ Public Class ucrReceiverExpression
...
-        Else
-            Return cboExpression.Text
-        End If
+    Private Sub cboExpression_KeyUp(sender As Object, e As KeyEventArgs) Handles cboExpression.KeyUp
+        iCurrentPosition = cboExpression.SelectionStart
+    End Sub
+
+    Private Sub cboExpression_MouseDown(sender As Object, e As MouseEventArgs) Handles cboExpression.MouseDown
+        iCurrentPosition = cboExpression.SelectionStart
+    End Sub
```
lloyddewit, line 135: combine these into a single handler?

instat/ucrReceiver.vb
```diff
@@ -150,18 +151,15 @@ Public Class ucrReceiver
...
     Public Overridable Function GetVariables(Optional bForceAsDataFrame As Boolean = False) As RFunction
-        Dim clsGetVariablesFunc As New RFunction
-        Return clsGetVariablesFunc
+        Return New RFunction
     End Function
 
-    Public Overridable Function GetVariableNames(Optional bWithQuotes As Boolean = True) As String
-        Dim strVarNames As String = ""
-        Return strVarNames
+    Public Overridable Function GetVariableNames(Optional bWithQuotes As Boolean = True, Optional strQuotes As String = Chr(34)) As String
```
lloyddewit, line 157: Is the `strQuotes` parameter used anywhere?

instat/ucrReceiverExpression.vb
```diff
@@ -134,24 +126,24 @@ Public Class ucrReceiverExpression
...
+    Private Sub cboExpression_TextChanged(sender As Object, e As EventArgs) Handles cboExpression.TextChanged
+        OnSelectionChanged()
+    End Sub
+
+    Public Overrides Function GetVariableNames(Optional bWithQuotes As Boolean = True, Optional strQuotes As String = Chr(34)) As String
+        If bWithQuotes Then
+            Return Chr(34) & cboExpression.Text & Chr(34)
+        Else
+            Return cboExpression.Text
+        End If
```
lloyddewit, line 146: Elsewhere you've used a line like below, do you have a preference?
```suggestion
        Return If(bWithQuotes, strQuotes & cboExpression.Text & strQuotes, cboExpression.Text)
```

lloyddewit (changes requested):
@Patowhiz thanks, looks great, I just had some small comments

instat/ucrColumnMetadata.vb
Patowhiz: Yes, line 92 is important. The property set is checked in line 62 to avoid duplicate checks.

Patowhiz:
@lloyddewit thanks for the code review.
I hope I have answered all of your above questions.

instat/ucrColumnMetadata.vb
lloyddewit: understood, thanks

instat/ucrReceiverMultiple.vb
```diff
@@ -224,71 +289,46 @@ Public Class ucrReceiverMultiple
         Return lstColumnFunctions
     End Function
 
-    Public Overrides Function GetVariableNames(Optional bWithQuotes As Boolean = True) As String
-        Dim strTemp As String = ""
-        Dim i As Integer
+    Public Overrides Function GetVariableNames(Optional bWithQuotes As Boolean = True, Optional strQuotes As String = Chr(34)) As String
+        Dim strTempBuilder As New Text.StringBuilder
```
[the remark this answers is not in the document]
lloyddewit: ok, thanks

instat/ucrReceiver.vb
lloyddewit: Understood. The `strQuotes` parameter was added in this PR. But I cannot see a call to this function (or one of its overrides) that uses this parameter. If the parameter is not used, then can we delete it?

lloyddewit:
> @lloyddewit thanks for the code review. I hope I have answered all of your above questions.

@Patowhiz I think there are still 8 unresolved comments above, please could you check? thanks

instat/ucrReceiverMultiple.vb
```diff
@@ -224,71 +289,46 @@ Public Class ucrReceiverMultiple
...
+        Dim strQuoteHolder As String = If(bWithQuotes, Chr(34), "")
```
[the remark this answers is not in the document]
Patowhiz: I have removed the parameter.

instat/ucrReceiverExpression.vb
Patowhiz: @lloyddewit the 2 events receive different event types. One has KeyEventArgs and the other MouseEventArgs. Note that these events are there in the master and not new.

Patowhiz:
@lloyddewit thanks. I think I've answered all your questions and committed some of your suggestions

instat/ucrReceiverExpression.vb
lloyddewit: yes, my mistake, thank you

lloyddewit (approved):

lloyddewit:
@rdstern if you can retest/approve, then we can merge, thanks

rdstern (approved):
I now don't get any issues saving the data.
I am havin g some interesting problems getting the metadata sometimes.
a) I first tried with the extRemes data, with 2 sheets of 12,000 and it never finished.
b) On another occasion it gave me the following:

image

I continued and it then worked.

c) I then tried with versions that were saved as rds files.  For the dhs data it was fine but with no message.  It looked as below.  For the extremes data it didn't give the message but then worked while I was writing this.

image


I am approving because the changes are so important.  I will keep checking.

Patowhiz:
@rdstern when you say no the first time, the message prompt never gets shown to you again, until you restart R-Instat. That means an other imports that follow won't prompt you for column metadata viewing. Same applies to saying yes. 
The yes or no prompt is not tied to a specific data frame. It's only tied to detection of the first data frame that has more than 1000 columns.
We could design the prompt to be tied to specific data frames if need be. 

As explained in my previous comments, if you accept to view the column metadata, expect some performance issues. This is due to filling the `Reogrid` with many rows. The solution needed to avoid this is similar to what we currently do with the data viewer.

rdstern:
@Patowhiz thanks.  Now I understand.  I tried again with the DHS wide data, which is for me now, an rds file.  And it asked me in the right way.
Incidentally that is the dataset with one over 7000 variables, one more over 4000 and so on.  It is actually the data set from DHS that they allow to be used freely for practice, and I hope sometime soon to tell them how nicely it could be analysed if only they considered R-Instat.

I also found it has a lot, really a lot, of empty variables.  When I get rid of those, which I did, then it just has one data frame of over 2000 and two with more than 1000 variables.  I said yes and it took only 7 seconds to get the metadata!

We have both approved now, so I hope this can be merged this week.

PR #8469, opened 2023-08-01
Added Secondary Axis to Plot Options

### LINKED ISSUE #8382: Add a secondary axis to our Plot Options
In May I raised the discussion topic on producing a climograph #8310.  This will be a very useful product for our forthcoming courses.  There was an excellent contribution by @COLIEWO that has emphasised the only extra we need is a secondary axis.  

It is too late now to have a special dialog, but I hope there may be time, even during the sprint, to add this feature, at least in an elementary way, to the main plotting sub-dialog?

It will (simply?) add the sec_axis function to the code.  Here it is in the example by Celestine:

`+ ggplot2::scale_y_continuous(sec.axis = sec_axis(~./5,name = 'Temp'))`

Here is the sub-dialog:
image

I suggest there is space for this block at the bottom of this Tab..  It is presumably on both the X-axis and the Y axis.  It is (at least for now) only for the `Continuous `option of these axes.  It could be called `Secondary Axis`.

Then the first item is called Transformation.  I think it may be optional, and (if not given it uses everything the same as the main axis.  So have a checkbox (default checked) with label Transformation.  Then there is a little pull-down control containing +,-,*,/ and perhaps with default being *.  Then a field into which any `number `can be typed.  (I think it is too general for a useful up-down.

Then there is the name of this axis.  I presume we call it Title to be consistent with the main axis.  Is it then just the same 3 options as for the main axis title.  

Do we need anything else?

MeSophie:
Fixes #8382 
@rdstern, @lloyddewit  I have Added Secondary axis to plot Options. Please can you check it.


rdstern:
@MeSophie that's already great and just about working.  Here is a list of suggestions:
a) The default seems to be with the Transformation, hence a second axis, ticked.  It should, by default be unticked.
b) The Title field is quite small.  Line it and make it the same size as the Axis Title field that is right above it.
c) Make the drop down with the *, etc just slightly smaller.
d) Make the next field into a drop down, with suggestions.  I remains the default.  Then various values that sound odd, but facilitate common uses for the second axis.  For example temperaure plots might have Centigrade on the left and Fahrenheit on the right.  So you can type, as now, with the drop down suggesting 1, 1.09361, 1.8, 2.20462, 2.54, 10, 25.4, 100 1000.
e) Make the whole box a bit wider, so you can add a checkbox (default unchecked), with label `Offset ` Then a pull-down with default 0 and possible suggested values -32 and 0 and 32.  

I tried changing your code to give an example with an offset `+ ggplot2::scale_y_continuous(sec.axis=sec_axis(~.*1.8+32))`
So if the Offset is checked, then you add (+or -) & <value of offset> as here.

MeSophie:
@rdstern  Please can you check the change I made?

rdstern (approved):
@MeSophie that's excellent.  Many thanks.  @lloyddewit over to you.

instat/ucrAxes.vb
```diff
@@ -537,6 +616,31 @@ Public Class ucrAxes
...
+                    clsXYSecondaryAxisFunction.AddParameter("trans", "~." & ucrInputSecondaryAxis.GetText & ucrInputTrans.GetText & ucrInputOffset.GetText, bIncludeArgumentName:=False, iPosition:=0)
+                    clsXYScaleContinuousFunction.AddParameter("sec.axis", clsRFunctionParameter:=clsXYSecondaryAxisFunction, bIncludeArgumentName:=True)
+                ElseIf ucrInputOffset.GetText = "32" Then
+                    clsXYSecondaryAxisFunction.AddParameter("trans", "~." & ucrInputSecondaryAxis.GetText & ucrInputTrans.GetText & "+" & ucrInputOffset.GetText, bIncludeArgumentName:=False, iPosition:=0)
+                    clsXYScaleContinuousFunction.AddParameter("sec.axis", clsRFunctionParameter:=clsXYSecondaryAxisFunction, bIncludeArgumentName:=True)
+                End If
+            Else
+                clsXYSecondaryAxisFunction.AddParameter("trans", "~." & ucrInputSecondaryAxis.GetText & ucrInputTrans.GetText, bIncludeArgumentName:=False, iPosition:=0)
+                clsXYScaleContinuousFunction.AddParameter("sec.axis", clsRFunctionParameter:=clsXYSecondaryAxisFunction, bIncludeArgumentName:=True)
+
```
lloyddewit, line 635:
```suggestion
```

instat/ucrAxes.vb
```diff
@@ -537,6 +616,31 @@ Public Class ucrAxes
...
+            End If
+        Else
+            clsXYScaleContinuousFunction.RemoveParameterByName("sec.axis")
+        End If
+        AddRemoveContinuousXYScales()
+    End Sub
+
+
```
lloyddewit, line 643:
```suggestion
```

instat/ucrAxes.vb
```diff
@@ -547,4 +651,17 @@ Public Class ucrAxes
             End If
         End If
     End Sub
+    Private Sub SetNameSecondaryAxis()
```
lloyddewit, line 654:
```suggestion

    Private Sub SetNameSecondaryAxis()
```

lloyddewit (changes requested):

PR #8470, opened 2023-08-02
Ensured position works correctly in Calculator dialog

### LINKED ISSUE #8434: Position sub-dialog may not always work correctly and overwriting an existing variable is too easy!
@derek, please see my suggestion below.  I think there is a really easy, and general, solution.

When I use the Position Sub-dialog from the Calculator it does not always work. It generally seems to work if i choose to position the new column at the beginning or the end of the data frame but it sometimes doesn't work when i try to position the new column before or after a column within the data frame. 

**To Reproduce**
1. Go to 'Calculator'
2. I used the Integer Keyboard to make a factorial column.
3. Use the Postion sub-dialog to place the new column in different places.
4. See where it comes up.

As you can see in my screen shot I have asked for fact2 to come after the fact column but it has been placed at the end of the data frame after the fives column. 

@rdstern could you have a look and maybe suggest someone else too? I am hoping someone could just play around with this and see if they experience the same thing or if perhaps I am doing something wrong. 

Thanks. 

Rachel

N-thony:
Fixes #8434
@rdstern have a look. Thanks.

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -119,7 +119,13 @@ Public Class dlgCalculator
     Private Sub SaveResults()
         If ucrCalc.ucrSaveResultInto.ucrChkSave.Checked AndAlso ucrCalc.ucrSaveResultInto.IsComplete Then
             clsRemoveLabelsFunction.AddParameter("col_names", Chr(34) & ucrCalc.ucrSaveResultInto.GetText() & Chr(34), iPosition:=1)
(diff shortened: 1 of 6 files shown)


rdstern (dismissed):
@N-thony I like the change, so I am approving.
a) I still think there could be a bug in the position control, but I would be happy to test that separately later unless you have found something obviously easy to fix?
b) It doesn't seem to work in quite the same way for the data frames.  In hindsight, I am actually happy that it is different - so you can choose to overwrite a column, but not a data frame.  (If you want to get rid of a data frame then you have to delete it!
c) The other change that affects all dialogs, that I am very keen on, is that when you return to a data frame the focus is always on the first receiver (so not on the receiver that you left it on.)  Patrick said that is also just a change in a single line.  Can you add that here - or in a second pull request - or should Patrick do it?

rdstern (approved):
Great - seems fixed

PR #8474, opened 2023-08-04
Restored "none"  legend option in plots sub-dialog

### LINKED ISSUE #8451: Fix two recent regressions on the graphics sub-dialog

Vitalis95:
Fixes #8451 
@rdstern , issue a) is no longer there for the updated master branch , b) the default  None  in the legends is  now restored and part c) of the issue is sorted in PR #8005

### DIFF
--- a/instat/sdgPlots.vb
+++ b/instat/sdgPlots.vb
@@ -312,9 +312,10 @@ Public Class sdgPlots
 
         'Theme Tab Checkboxes under grpCommonOptions
         ucrChkLegendPosition.SetText("Legend Position")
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@Vitalis95 great, and continuing my aim that the next release will be that much more"solid".

lloyddewit (approved):

PR #8481, opened 2023-08-07
Improved Titles options in Plots subdialog

### LINKED ISSUE #8477: Small improvement in the titles tab for the graphs menu?
@Vitalis95 this is only to be done soon if you agree and it is a simple change.  I think it is?

I am very pleased with the new facilities in the ggplot graphs for titles.
However it has made the default ggplot2 code very long, and complicated, because most of the title options are always included. 
 This is whether, or not, any effort to add titles has been added.

I wonder if you could parallel what we have for facets?  So there is an `Include Titles ` checkbox at the top.  If checked, then it gives the titles code - as you do now.  If unchecked then they are all omitted?

Vitalis95:
Fixes #8477 
@rdstern , @N-thony , @lloyddewit , this is ready for review.

### DIFF
--- a/instat/sdgPlots.Designer.vb
+++ b/instat/sdgPlots.Designer.vb
@@ -499,7 +499,7 @@ Partial Class sdgPlots
         '
         Me.ucrChkIncludeTitles.AutoSize = True
         Me.ucrChkIncludeTitles.Checked = False
-        Me.ucrChkIncludeTitles.Location = New System.Drawing.Point(11, 7)
+        Me.ucrChkIncludeTitles.Location = New System.Drawing.Point(6, 10)
         Me.ucrChkIncludeTitles.Name = "ucrChkIncludeTitles"
         Me.ucrChkIncludeTitles.Size = New System.Drawing.Size(121, 23)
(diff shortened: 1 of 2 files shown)


rdstern (approved):
@Vitalis95 nicely done.  This was perfect.  This is the second different edit to the code that I am approving first time round.
@lloyddewit just to report something very nice also about your script windows.  
a) I did the check by first sending the script without a title to a script window.
b) Then I made a new empty script window and added a title.  It then adds a whole set of titles to the script.
c) Then another script window and took away the title - to see if Vitalis also took it away from the code.  He did.

Now "your" script window properties.  This is important, because the graphs are a high-spot of R and are very logical, but also produce very long commands.  They are obvious bits of code to get the script and tweak.  I have been doing that and finding it a bit awkward with the single line being so long.  So I tried here and found:

a) It indents progressively.  Very nice.
b) Undo and Redo also both work really well.  I can easily get back to the long line if I wish - even 5 progressive undos.

My big idea for Antoine is for a sort of primitive undo - but very nice - for R-Instat as a whole.  How nice that I am now discovering it is working so sweetly in the script windows!

lloyddewit (approved):

PR #8482, opened 2023-08-08
Bug fix when wrapping or unwrapping a list column

### LINKED ISSUE #8475: Column metadata has new problems
@N-thony this is a problem that has been introduced recently.  It is in the merged version, but not in the latest release.

Here is a picture of the problem with the dodoma data:

image

The problem is general - here is the same problem with the usual survey data:

image

I get this problem when I do a filter and then save the subset.  @N-thony I think you worked on roughly this when we had list variables and tried to do the same thing?

N-thony:
Fixes partially #8443 
Fixes #8475
@rdstern this fixes the bug when wrapping/unwrapping a column to not lose the `meta data` and also correct the display of `2:1` to `2, 1`

### DIFF
--- a/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrDataViewReoGrid.vb
@@ -76,7 +76,8 @@ Public Class ucrDataViewReoGrid
         For i = 0 To grdData.CurrentWorksheet.Rows - 1
             For j = 0 To grdData.CurrentWorksheet.Columns - 1
                 Dim strData As String = dataFrame.DisplayedData(i, j)
-                If grdData.CurrentWorksheet.ColumnHeaders.Item(j).Text.Contains("(LT)") Then
+                If grdData.CurrentWorksheet.ColumnHeaders.Item(j).Text.Contains("(LT)") AndAlso
+                    strData IsNot Nothing Then
                     strData = GetInnerBracketedString(strData)
                     strData = If(strData.Contains(":"), strData.Replace(":", ", "), strData)
                 End If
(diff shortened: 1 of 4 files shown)


rdstern (changes requested):
@N-thony I am very releived you are making progress on this.  I am really keen on being able to handle lists properly in R-Instat and I think we are close.
a) The 2: 3 is improved, but it is now 2 ,3, while all the others would be 2, 3.
b) The wrapping, etc now all seems to work.  Great!
c) I just note that when calculating divisors the one difference now between your display and the R viewer is that you give 0 for the divisors of 1, while the R viewer gives NA.  (You may give 0, but the length is 0, so it isn't a real 0!)
d) But there is still a problem.  The filter to give a subset seems to work, but the metadata is still odd, see below:
image

rdstern:
@N-thony the fixes mean that now it seems exactly the same as the R viewer for the list variables.  That's great.
Is this a suitable pull request to check on the display in Roman - which remains different from R viewer?
And the subset still has to sort out the metadata.
I am very pleased with the progress though.

rdstern (approved):
@N-thony that's great.  I think the problem with displaying lists and also getting a subset, seems to have been solved.
I also tried everything with a dataframe where I am keen on wordwrapping, namely the statquotes and there then wrapping and everything seem to work perfectly.

There is an interesting "feature" about the wordwrapping for a list column and also for (say) factorial, where there are very long numbers - and no spaces.  In each case the wordwrap with 20, 0r 40, etc, does not wrap.  It just makes the field very wide.  But I don't care, because the unwrap works fine.  This is good enough.

@lloyddewit as soon as you are happy with the code, then I would like to have this merged.  I will then also inform @rachelkg  because she is working on items that can continue to test that this is ok.

lloyddewit (approved):

PR #8501, opened 2023-08-21
Database Name change for Output Window in View Menu

### LINKED ISSUE #8496: Make Output names consistent in the View menu and the Toolbar button
@derekagorhom we have `Output Window` in the toolbar and in the View menu we have only `Output`. Can you rename it for consistency?
image

image

derekagorhom:
Fixes #8496 
I made the name change in the database for the `Output` in the `View menu` to `Output window`

image

@N-thony, @lloyddewit  can you review this

Thanks


lloyddewit (changes requested):
@derekagorhom Thank you for the change. This will work in the short term but when we regenerate the database for the next translation release, the database will be overwritten with the text in the `.Designer.vb` file.
Therefore you should also update the text in the `.Designer.vb` file.

However, I added a note to the related issue, I suggest that you wait until this is resolved before further code changes.
Thanks

derekagorhom:
@lloyddewit Noted
I can revert the changes once Roger is ok with the name change you and Antoine suggested

rdstern (dismissed):
I am very happy with this change to the toolbar.

instat/frmMain.Designer.vb
```diff
@@ -3067,21 +3067,21 @@ Partial Class frmMain
         '
         Me.mnuOutputWindow.ImageScaling = System.Windows.Forms.ToolStripItemImageScaling.None
         Me.mnuOutputWindow.Name = "mnuOutputWindow"
-        Me.mnuOutputWindow.Size = New System.Drawing.Size(178, 22)
-        Me.mnuOutputWindow.Text = "Output Window"
+        Me.mnuOutputWindow.Size = New System.Drawing.Size(180, 22)
+        Me.mnuOutputWindow.Text = "Show Output "
```
lloyddewit, line 3071: Eventually, we will want to translate 'Show Output' into other languages. If we end the string with a space, then the translation will not work correctly.
```suggestion
        Me.mnuOutputWindow.Text = "Show Output"
```

lloyddewit (changes requested):

Issue #8281, opened 2023-04-22
Climatic data entry/editing has new serious problems


rdstern:
Update july 25th.  we used to get this message repeatedly on the recent workshop.  I am not sure whether it is because of the language, but I think that it is simply when cells are deleted repeatedly.  This has now been fixed it seems, but replaced by other problems reported in responses below.

![image](https://user-images.githubusercontent.com/11226469/233795804-852d923e-a502-41b5-8f94-1802f47e2a9c.png)

While correcting this perhaps adding the facility to use it as simply a viewer could also be included.  That may just need the editing button to be enabled, when there are variables in the view receiver, even if none are in the editing part.  Perhaps it is not as simple as that?

Could there be a right-click here to Delete Cell(s) - as there is on the right-click menu in the Data View?  Here, as you are editing, it might do it without the warning?

 Aha I changed back into English and got it just the same.  So it is general and not a language issue.

![image](https://user-images.githubusercontent.com/11226469/233796605-4eba2203-41b9-4383-a5bd-5c5696d7dfd4.png)

I deleted one entry and that was fine.  It went yellow, as it should.  Then I got this message when I deleted the second day.

@Patowhiz this would be good to fix.  Can you allocate it?

Fidel365:
@rdstern, I'm still working on this and it will be ready soon, I just need some aid on adding a command to my delete function, @Patowhiz will be helping me with that

rdstern:
@Fidel365 thanks for the update and pleased you are back.  Are you also looking for @N-thony to allocate more R-Instat work, or are you fully occupied.  And where are you based now?

Fidel365:
@rdstern,thanks for the welcome, So I'm currently at the Kakamega office as an employee of INNODEMS, I would love more work from R-instat, but I'm also involved in the G.P.I team for the documentation of PfR

N-thony:
@Fidel365 any progress? I can see you have a draft PR on this.

rdstern:
I have just been using the improved version of this dialog.  This is a powerful feature and it is working quite well, and I never had the nasty error reported at the top.  

But I also note problems and improvements that should each (I hope) be easy to implement.
Here is the dialog completed and ready to enter/edit data.  Note that Ok remains disabled and I like that.

<img width="313" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/d425ac5f-175f-4346-9a52-64af496793ad">

[...]

PR #8512, opened 2023-08-28
Ensured that rows can be changed and comments added in the Climatic > Examine/Edit > Data entry Dialog

N-thony:
Fixes partially #8281
@Fidel365 @rdstern this partially fixes the bug reported in the corresponding issue when the comment is added or a row values has changed. Have a look.

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -4343,7 +4343,7 @@ DataSheet$set("public", "save_data_entry_data", function(new_data, rows_changed,
     # This affects factor columns only  - we need to find out why and how to solve it best
     self$add_defaults_variables_metadata(self$get_column_names())
     self$data_changed <- TRUE
-   }
+  }
 })
 
 DataSheet$set("public", "add_flag_fields", function(col_names) {


rdstern (changes requested):
@N-thony that's excellent, I think one of the problems is fixed.  It now doesn't make that funny extra column.  
However, one critical problem still remains.  And I think I have found it.  I am not sure of the solution though.

After I make the change, I made it to Xai-Xai, and just added a comment.
Then I returned to the dialog.
In the returned dialog I now change the Station name.
Now I find that Ok is enabled - though I have not changed any data.  So it should not be.

I pressed To Script to see what it has produced:

```
# Dialog: Daily Data Editing/Entry

new_data <- data.frame(Station_Name="CHIMOIO")
data_book$save_data_entry_data(data_name="Mozambique_3_stations", new_data=new_data, rows_changed="22951", comments_list=list())
rm(new_data)

```
It has most of this code from last time, when I just changed a comment.  BUT it has a new station, i.e. the one I have now selected.  

I closed the dialog and checked the 3 stations - the 3 factors are fine.

Now I ran the script - so like pressing Ok.  Then I checked the stations again and found one row less in Xai-Xai has changed and become Chimoyo!

(When I hit the problem, as I was working before, I had changed a whole month, and it had changed a month from one station into the other.

I checked back in the dialog and when I change the variables to view it also enables Ok.  So that is currently another risky thing to do.  

I don't actually mind Ok being enabled, when there is nothing to change, as long as it then doesn't change anything!  But I suggest it would be simpler to only enable Ok, when there is at least a comment to change, or data to change, or both.

Here are the data I have been using.

[Mozambique_3_stations_shifted_summaries.zip](https://github.com/africanmathsinitiative/R-Instat/files/12458358/Mozambique_3_stations_shifted_summaries.zip)

rdstern (approved):
Yippee!  At last this now seems fixed.  Many thanks @N-thony.  @lloyddewit please check and merge as soon as you can.  This is useful for the forthcoming Zambia workshop.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -4315,6 +4315,7 @@ DataSheet$set("public", "get_data_entry_data", function(station, date, elements,
 })
 
 DataSheet$set("public", "save_data_entry_data", function(new_data, rows_changed, add_flags = FALSE, ...) {
+  if (ncol(new_data) > 1) {
```
lloyddewit, line 4318: indent lines 4319 to 4345?

lloyddewit (changes requested):
@N-thony thanks look good, just one small formatting suggestion

rdstern (dismissed):
Still looks fine to me

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -4315,33 +4315,35 @@ DataSheet$set("public", "get_data_entry_data", function(station, date, elements,
...
+        }
+      }
+    }
+    if(length(nrow(new_data)) > 0) cat("Row(s) updated: ", nrow(new_data), "\n")
+    self$set_data(changed_data)
+    # Added this line to fix the bug of having the variable names in the metadata changing to NA
+    # This affects factor columns only  - we need to find out why and how to solve it best
+    self$add_defaults_variables_metadata(self$get_column_names())
+    self$data_changed <- TRUE
+   }
```
lloyddewit, line 4346:
```suggestion
  }
```

lloyddewit (changes requested):

lloyddewit (approved):

Issue #8189, opened 2023-03-04
Add improvements to the new Maximise output window


rdstern:
@ChrisMarsh82 and @Patowhiz 

This is working well, but is minimal.
We go to it from the main Output window and also from the Help when we ask for R-Help.
Initial suggestions are:
1. Add a `Help` button.  Good if each window could call the help.
2. Give it a better name.  The R window - shown at the bottom - has a longer, but not particularly good name.  This window is currently called Output.  It is more than that, because it is also now used for the R-Help.  So we could call it `Output and R-Help`, and I hope that total lack of imagination will tempt someone to suggest something better.
3. Allow multiple "pages".  Currently with the R-help you usually come to the main page for a package, see below:

![image](https://user-images.githubusercontent.com/11226469/222879795-ebe308af-726f-4b76-b3d0-a8b23fccccc3.png)

Now you click on an entry and that's fine.  But you can't go back.
This also means we can build up many single output files.  Perhaps that's a different point though?
4. Be able to resize the graphs.  This is possible in the R-graphics window and was missed by participants in the recent workshop.
5. More???  For example we only save files in one way.  Would more options be useful? 

![image](https://user-images.githubusercontent.com/11226469/222879533-b8bb24de-9860-4b45-8c6a-41c16153b109.png)

rdstern:
@ChrisMarsh82 and/or @Patowhiz you have both been very quiet on the output window improvements recently?  Maybe you are both too busy, or assuming the other is doing it?  
Two of the pull requests #7999 and #7978  are also waiting for some improvements.  I think they may produce multiple html tables and that is currently not allowed for in the output window.  They currently go to the browser still.  Could they go to the maximise window if that permittied multiple pages?

Patowhiz:
@rdstern changing the title of the maximized output window should be trivial. It can be done dynamically when it's launched. @Vitalis95 can you look into doing this dynamically? Ideally if called by a help function, the title should be different.

Making it support going back and forward is possible. That functionality is currently not built in yet, I never anticipated to have the help as a use case. But it can be implemented.

rdstern:
@Patowhiz I have been finding the pivot tables very useful, particularly in the new Outpit window, whatever the name:
<img width="723" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/55e0376b-7f57-429d-ba2f-005ad2dccbae">

[...]

PR #8515, opened 2023-08-29
Enabled viewing of multiple R help maximised windows in Import from Library dialog

Patowhiz:
Fixes @rdstern comment https://github.com/africanmathsinitiative/R-Instat/issues/8189#issuecomment-1685168424

@rdstern quoted comment 

> @Patowhiz I have just managed to crash R-Instat - but I was being a bit ambitious. I have a maximised window open and I decided to keep it open for later. Then I used File > Open From Library and the R-Help - which also uses the maximised window. It seems this isn't (yet?) possible?


rdstern (changes requested):
@Patowhiz I think this may be 50%, which is already good.  A major dialog I have been using is Help > Package Documentation.  That's partiocularly where I hoped I could use a back button, and hopefully not crash.  When I use it twice I still get the following error.

image

I suspect you may not yet have changed the window from here?

Patowhiz:
@rdstern, I have now fixed the bug.

rdstern (approved):
@Patowhiz it is great now.  Perhaps I was missing this last time, but now anyway in the Package Documentation I can a) have multiple pages open and b) go back to the index, either from the bottom of the page, or right-clicking and using Back!

@lloyddewit I understand the wide data frames needs to be mergewd first.  These are good improvements.

Patowhiz:
@rdstern @lloyddewit this PR does not depend on any other PR. Thanks.

lloyddewit (approved):

PR #8516, opened 2023-08-29
Notify the user when there is a new version of R-Instat on the website

N-thony:
@ChrisMarsh82 there is no issue linked to this PR but as discussed this will notify the user when there is a new version online. Please have a look. Thanks.

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -228,13 +228,18 @@ Public Class frmMain
         Try
             ' Download the version information file from the website
             strLatestVersion = webClient.DownloadString("https://r-instat.org/version.txt")
+            Dim strCurrVersion = My.Application.Info.Version.ToString()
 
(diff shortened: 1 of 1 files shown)


Patowhiz:
@N-thony after quickly going through this, I suggest a silent message that just notifies the user when there are network or website related issues. This is a common trend in most software.

I have not done any tests regarding the loading performance that could be caused by the internet access. I would expect the function added here to slow the loading of the main form in intermittent connections. I recommend calling your function in a separate thread due to such scenarios. Would be a bad user experience when it takes a lot of time for R-Instat to load due to such issues. If you have tested this aspect and found it to not be an issue, ignore my suggestion. 

Thanks

rdstern:
On reviews I suggest @Patoand @lloyddewit for merging

ChrisMarsh82 (approved):

PR #8529, opened 2023-09-07
Ensured the Identify Key checkbox works correctly in Row Names or Numbers dialog

### LINKED ISSUE #8510: Apply Edit > Wordwrap for Factor and other variables and correct the Row Names Dialog?
@N-thony I have recently had some factor variables where the labels are very wide.  One example is the R-library, opening the mortality data from the FactoMineR package.  That has a messy (or attractive) label variable on the left.

image

I tried to make that an ordinary variable - that gave me another issue I am reporting.  Once it was the first variable I tried to make it a factor to check it was unique.  Then I thought it would be nice to wordwrap.  It doesn't apply to factor variables and I am happy if you decide it shouldn't?  But - only if it is easy - why not?

And, I assume it can't apply to numeric variables, because there are no spaces there.  Because we have some very large numbers it would be exciting if it could?

However, this is a luxury, while the other small issue on Prepare > Data Frame > Row Names dialog is a simple bug I hope.  My guess is that it could be easy for anyone to do?

image

a) The Make the key variable checkbox seems stuck.  I am happy this time that it is ticked, but the idea of a tick-box is that I can choose.  I can't untick it.
b) If you press Ok it says that the column isn't unique - which is odd, because that's an R rule.  And it gives a column of NA values.  Instead, press To script and you get the following:

```
# Dialog: Row Numbers or Names

row <- data_book$get_row_names(data_name="mortality")
data_book$add_columns_to_data(data_name="mortality", col_name="row", col_data=row, before=TRUE)
row <- as.numeric(x=data_book$get_columns_from_data(data_name="mortality", col_names="row"))
data_book$add_columns_to_data(data_name="mortality", col_name="row", col_data=row, before=TRUE)

data_book$add_key(data_name="mortality", col_names="row")
rm(row)
```
You can see that the first 2 lines are almiost repeated in the third and fourth.  That (I assume) gives a column of NA values - and it is true they are not unique!  

How did we get here I wonder?

derekagorhom:
fixes #8510
The changes are; 

a) Delete the second set of lines. 
b) To permit the key variable checkbox to be unchecked while still being checked by default.
c) added the function  `all(sapply(rownames(data), is.numeric))`

@N-thony , @rdstern this PR is ready for review

Thanks

### DIFF
--- a/instat/dlgRownamesOrNumbers.vb
+++ b/instat/dlgRownamesOrNumbers.vb
@@ -22,7 +22,9 @@ Public Class dlgRowNamesOrNumbers
     Private clsGetRowNamesFunction As New RFunction
     Private clsSetRowNamesFunction As New RFunction
     Private clsAddKeyFunction As New RFunction
-    Private clsAsNumericFunction As New RFunction
+    Private clsAsBaseFunction As New RFunction
+    Private clsAsSapplyFunction As New RFunction
+    Private clsAsRownamesFunction As New RFunction
     Private clsDummyFunction As New RFunction
 
     Private Sub dlgRowNamesOrNumbers_Load(sender As Object, e As EventArgs) Handles MyBase.Load
@@ -33,7 +35,6 @@ Public Class dlgRowNamesOrNumbers
 
         If bReset Then
             SetDefaults()
-            IdentifyKey()
         End If
         SetRCodeForControls(bReset)
         bReset = False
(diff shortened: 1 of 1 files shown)


rdstern (changes requested):
@derekagorhom that new first line generates an error.  The one:
`all(sapply(rownames(survey), is.numeric))`
I don't think it is needed, so you can delete it.
I am a bit confused how you managed to test it without finding that error yourself.

Originally I think the extra code was introduced because often the row names are just numeric, and the code finishes with them as character.  I am now prepared to live with that, at least for now!

derekagorhom:
@rdstern apologies, I think i read the issue wrongly. I have made the changes your requested. but you second paragraph is a bit confusing, Please did you want the dialog to have the original four lines again?

rdstern (approved):
@derekagorhom that now looks fine - thanks.  @lloyddewit for the last week I am afraid (at least for now) - over to you!

lloyddewit (approved):

PR #8540, opened 2023-09-13
Bug fixes in the Rename Dialogue when Select is applied

### LINKED ISSUE #8488: Serious bug found in rename dialog
Comment added - we have now found a more general issue with running commands when select is in operation.  So leave this issue, until the general point is solved.

@N-thony this may also need to wait until @MeSophie has completed her additions to the rename dialog.  But you may be able to investigate quickly, in which case perhaps she could incorporate the correction with her new changes.  I suspect it may be better for her improvements to be merged first.

Try with the usual survey data.  Do a select so (say) just the first 4 columns are selected.  
Now go to the Rename column dialog and rename with.  Try an option - I think any option will do.
a) The data are cleared.  It is just as though you have done a File > Close Data Book - but with no warning!
b) Now try to re-open the data frame again.  Nothing opens!  

This is serious - very serious.  I think we haven't seen it before, because we are not yet using select much.  We should start testing lots of dialogs with a select, to make sure they all work.  Let's correct this one first, but see (at the same time) whether there is anything that will give the same problem elsewhere.

N-thony:
Fixes #8488 
@rdstern @lloyddewit this fixes the bug reported in the Rename Dialogue when trying to rename a column from selected columns. Have a look. Thanks.
@MeSophie you can also have a look.

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -798,6 +798,7 @@ DataSheet$set("public", "rename_column_in_data", function(curr_col_name = "", ne
             purrr::map(.x = keys_to_delete, .f = ~self$remove_key(key_name = names(active_keys[.x])))
           }
         }
+        if(self$column_selection_applied()) self$remove_current_column_selection()
         # Need to use private$data here because changing names of data field
         names(private$data)[names(curr_data) == curr_col_name] <- new_col_name
         self$append_to_variables_metadata(new_col_name, name_label, new_col_name)
@@ -818,6 +819,7 @@ DataSheet$set("public", "rename_column_in_data", function(curr_col_name = "", ne
       curr_col_names <- names(private$data)
       curr_col_names[cols_changed_index] <- new_col_names
       if(any(duplicated(curr_col_names))) stop("Cannot rename columns. Column names must be unique.")
+      if(self$column_selection_applied()) self$remove_current_column_selection()
       names(private$data)[cols_changed_index] <- new_col_names
       for (i in seq_along(cols_changed_index)) {
         self$append_to_variables_metadata(new_col_names[i], name_label, new_col_names[i])
@@ -843,6 +845,7 @@ DataSheet$set("public", "rename_column_in_data", function(curr_col_name = "", ne
         .fn = .fn,
         .cols = {{ .cols }}, ...
       )
+    if(self$column_selection_applied()) self$remove_current_column_selection()
     new_col_names <- names(private$data)
     if (!all(new_col_names %in% curr_col_names)) {
       new_col_names <- new_col_names[!(new_col_names %in% curr_col_names)]
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -4400,7 +4400,7 @@ DataSheet$set("public", "remove_empty", function(which = c("rows", "cols")) {
   if(self$column_selection_applied()){
     df_with_Selection <- self$get_data_frame()
     df_without_Selection <- self$get_data_frame(use_column_selection = FALSE)
-    # Check for missing columns in B and remove them from A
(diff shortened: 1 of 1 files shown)


rdstern (changes requested):
@N-thony I had forgotten this one.  Delighted you have tried to fix it now.  
a) With the survey I first tried selecting the first 3 columns and then remove empty.  I was delighted it said it removed no empty variables.  But when I removed the selection it had still taken the other variables away.
b) I started again and added an empty variable.  Then I selected the first 3 variebles plus the empty one.  I then got rid of empty again.  It correctly said it had removed one variable.  But it then also closed all the data, and did not allow me to open any data again.  I had to close R-Instat and start again!

rdstern (changes requested):
@N-thony that's a big improvement.  When there are no empty columns to delete it now works ok.
I then used Prepare > Data Frame > Insert Columns/Rows so there were some empty columns to delete.
a) I had one in the selection and one outside the selection.  It deleted both.
b) It restored all the columns, (so effectively not having the selection.  But they remained red and stated that a swlwction was in process..

rdstern (changes requested):
@N-thony I am trying with the usual survey.
a) I first added 2 empty variables.  Then I did a select including the first 3 variables, plus the empty ones.  I then got rid of the empty variables.
It did this correctly, if correct is also to undo the selection.  I assume that is what you are doing.  I am ok with that.
(Let me just check that when I delete the empty variables you are intending to first remove the Selection?)
So far so good, but I then wantewd to continue, re-adding empty variables to try a different combination.  It gives me this error message, though I can continue.

image

Oh that is interesting.  I ignore that error - which keeps coming.
b) I now add 2 empty variables and then select a subset with just one of them.  Then I delete empty.  It only deletes that one and leaves the other one there.  Then it also undoes the selection!
c) But I also tried with no empty variables.  Then it didn't delete anything empty, but it left the select in place.  And when I undid the select, it had (as before) deleted all the other variables.
Bu

rdstern (changes requested):
@N-thony Just checked what happens when there are no empty columns:

image

Then it is still deleting all the non-selected variables!

rdstern (approved):
@N-thony great, now it seems to work fine.  Many thanks.
It unsets the select, and just deletes the empty variables in the select.
@Patowhiz over to you to check the code and merge.

MeSophie:
@N-thony Sorry but the rename dialog is still misbehaved when using selected dialog.
1-
a) import survey data>Column Selection dialog> select the three first variable
b) In Rename dialog>Single try to change the name of any variable (me I renamed village to Town)  
c) When click Ok we have the change but the data view part is showing the all dataset and the variables are red showing that you have a selection on this dataset.
image
 
2-
a) import survey data>Column Selection dialog> select the three first variable>on Define new column selection sub change colum to stard with prefix "f"
b) In Rename dialog >Rename With >Data Frame>change case to Upper Lower>ok
c)  The values of the variable have all disappeared.
image

MeSophie:
image
@N-thony this happens with all the options (Single, Multiple, Rename With ) in the Rename Columns dialogue not just in rename with>dataframe option .
If we select columns and try to rename one/many columns on the selected dataframe when click ok, it comes back to the all dataframe and remove the selection.

rdstern (approved):
@Patowhiz can you now check.  I am approving though we should make a new issue of a remaining problem.  But this is noiw solid, so it is an improvement.
The remaining problem is that when you rename variables, then some selects may no longer work, because they used the old names.  This does not apply if the select was defined by a starts with, etc - I think.  But it does apply if you just choose a set of names.  
There should be a check that when you apply an existing select it checks that the names it uses remain valid.  If not, then it states that and deletes the select.

rdstern:
@Patowhiz I see I already approved this one.  Can you test?  It is an important one to include, as renaming, with a selection, causes a nasty bug otherwise.  We may want to do better eventually, but this fix is ok for now.

ChrisMarsh82 (approved):

Issue #8487, opened 2023-08-09
Allow the user to easily reopen previously opened scripts


lloyddewit:
On 08/08/23 @rdstern wrote:

Having a list of past used scripts.

When using RStudio I have the past scripts as buttons at the top. Could we have a similar list in R-Instat.
And we have a very discrete way we could provide the list! Could it be added to the down arrow which is next to the Log/Script window. There is just one entry there currently, and it just says whether the window is open or not. We might keep that, but underneath is the recent scripts we have opened and saved.

I got to this recently when I was sent a set of data, in Excel and a script - so with an R extension. I used File > Open for both and it was equally simple. The data went to a data frame and the script went to the script window. I saved my own copy of the data as an RDS file and the code as an R file.

Now I returned today and the RDS file is available and in 3 places! It is on the front screen and is on the File menu and is on the toolbar drop-down. But my R file is nowhere?

So the initial opening is the same for data and code. But then it is different. We could, of course, save the R file in the list of past files, just like a data file, so that's another option. I am thinking that we are less obtrusive about the R files - but just as convenient. (I am also asking if the liobrary icon on the toolbar could have its drop-down, because we currently don't save a copy of those either. Though that's slightly different, because I anticipate that list just from opening a library file, not from saving a special copy.)

rdstern:
@N-thony after discussion with @lloyddewit please could you allocate this task - which is partly a further improvement in the toolbar.  I asume that this may be to @derekagorhom, as he has worked effectively on improvements to the toolbar before.  It may be that you could take a part of this yourself?

Note that File > Open can be used for data files - with an R extension.  The come automatically into the current script window.

Re-using past data files has 3 short-cuts currently as I show here:

<img width="469" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/4254b2c1-425d-4c31-8e7f-fe6e2568f219">

1 above is in the front menu.  Please could you also add the R files to this list, as well as data files.
2 above is that the same list is at the bottom of the file menu.  Could you please add recent R files to this list.
3 above is the drop-down from the toolbar.  Please leave this, as it is.  Do not add R files here.  Instead add them to the script/log window drop down.

<img width="384" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/47a5848a-14d4-45d7-a165-c64130d71f16">

This item on the drop down currently has the single item.  Please keep that an put a line below and then add the list of scripts recently used, below that item.

[...]

PR #8544, opened 2023-09-14
Insert Script and renaming tab in Script Window

### LINKED ISSUE #8570: Add a script tab to new Insert dialog
@N-thony Currently this Insert adds a line or so to the open script.

This tab would add a while script.  It will usually be to a new script tab.

Currently I go into a script from the r-Help on the Open from library.  Then I copy the script part - which is always the lines after the Example to the end of the file.  

I wonder if it would be easy to copy that script and insert it auromatically!

I assume choosing the place would use the same control as in the open from Library, namely:

image

The existing code will take us to the right data set.

I hope this might be easy?

If so, then it might be possible, at the same time, to have an option to use any package - not just those with data files.  

I really like the restriction to the data files, so this might be a different tab or a different drop-down?  And the hearing would say Function, rather than Data.

I can explain if this is unclear.  But also happy to start with just the data files!

N-thony:
Fixes partially #8487
Fixes partially #8151
Fixes #8570
@rdstern @lloyddewit I'm not sure about the specific issue related to this PR but as discussed this PR is about to insert script(s) to a specific cursor in the script window. Have a look. Thanks
I have added also a possibility of renaming a tab title in a script window.
So, if you double click in a specific tab title, it will open a small textbox where you type
image
then click Enter to rename it
image

### DIFF
--- a/instat/dlgScript.vb
+++ b/instat/dlgScript.vb
@@ -107,6 +107,8 @@ Public Class dlgScript
         ucrSaveModel.SetIsComboBox()
         ucrSaveModel.SetLabelText("Model Name:")
 
+        ucrInputPreviewLibrary.txtInput.ScrollBars = ScrollBars.Vertical
+
         'hide base button comment controls
         ucrBase.chkComment.Checked = False
         ucrBase.chkComment.Enabled = False
@@ -270,7 +272,7 @@ Public Class dlgScript
         For Each line As String In inputLines
             If Not String.IsNullOrEmpty(line) Then
                 Dim strTrimmedLine = ""
-                If strSelectedPackage <> "datasets" Then
+                If strSelectedPackage <> "datasets" AndAlso line.Contains("##D") Then
                     strTrimmedLine = line.Substring(line.IndexOf("##D") + 4).Trim
                 Else
                     strTrimmedLine = line.Substring(line.IndexOf(" ")).Trim
@@ -394,7 +396,7 @@ Public Class dlgScript
             If strSelectedPackage <> "datasets" Then
                 clsLibraryExpFunction.AddParameter("package", Chr(34) & strSelectedPackage & Chr(34), iPosition:=1)
             Else
-                clsLibraryExampleFunction.RemoveParameterByName("package")
+                clsLibraryExpFunction.RemoveParameterByName("package")
             End If
             LoadDatasets(strSelectedPackage)
             TestOkEnabled()


rdstern:
@N-thony this is great progress.  The issue is #8151, at least for the Edit > Script part.

I consider the rename and the Insert in turn.

For rename I suggest the closest is the rename data frame and here is that dialog. 

image

I picture a Rename button (and also a Rename Script on the script right-click.
Then an adaptation of the dialog pops up.  
a) The title is Rename Script
b) Underneath is an ordinary field with label `Current Name`.  In the control it might say   Untitled1.
c) Then, to the right is New Name as above.
d) Optionally there might be Script Label.  If that is given, then it should be wider than above, and would become a comment in the first line.  When the dialog is opened it mightr read that first line?  This is a luxury.
I need to check with David a bit, and @lloyddewit may have ideas.  I am not sure there is any need for R here?
e) There is a key difference between the script names and the data frame names though.  This is because we don't save individual data frames - just the data book.  So the data frame name is distinct and additional to the file name.  Here I suggest that, when a script is saved it suggests that name.  When a script is loaded, then the file name is used (as now), but without the .R.  So, the rename is a temprary feature of a script, unless it is kept as the filename.

Now the Script stuff.
a) Great that it already sort of works
b) Happy it is making use of the facilities and code in the current current Edit >Script dialog.
c) In #8151 I suggested `Add `for the button.  You have used `Insert`.  I don't have strong views, but perhaps slightly prefer your Insert.
d) Here is the current sub-dialog.  

image

1) I suggest it could have the title `Insert Line.`
2) I suggest the right-hand side is deleted and replaced, on the left, below, by our control for an R formula.  That's in all our halfway dialogs, as well as the calculator.
3) I suggest later we may add a keyboard.  For now all we need is the `Clear `button that is on all our halfway dialogs. 
 
Then, next time round, I would like to consider the different lines we will often need.  In the halfway dialogs we concentrate on the main R line and the extra code needed for R-Instat is handled for us, by the dialog.  This is the other way round.  The imported script has the R lines and we need to add the additions for R-Instat!

rdstern (changes requested):
@N-thony this is making excellent progress.  It is also a feature @Patowhiz was asking for, when I discussed today.
a) Rename:  The dialog looks ok.  I suggrest, for simplicity, you make the label for the label invisible for now.
b) Rename: When I click on the Save button, could it please suggest the name that is given there?
c) Rename: When a named file is saved, then could the name be given without the `.R ` added.

Now the Edit part:

I am not sure I am getting this right, yet.  I have realised that because it is a single line being generated we may not always need to see the command here.  We mainly need it in the script window.  
So I think the control we need is essentially the Preview control in the Prepare > Numeric > Transform dialog.

image

And sorry - this means the Clear isn't needed.  Perhaps we can discuss if you need?

Then could you have 5 tabs at the top.  
a) Library
b) Data Frame
c) the 3 that are already there.

Now if Library is chosen it gives roughly the following:

image

For now it is a single command - generating exactly what you currently have here as the Get Data > Library.  There are no buttons for now and it is exactly the option here, which seems to work perfectly.

The Data Frame command isn't there.  But it is there in your Save Results Tab.  I think it will be the same as there - which doesn't seem to work yet?

And it should produce the same command (I assume) as done currently in the File > New Data Frame dialog.  Essentially we will have the same twice, but I think that;s fine.  For the Data Frame Tab there are again no extra buttons for now.  It is a single command, namely:

`data_book$import_data(data_tables=list(data=data)) `

That's if you gave the name data in the entry field.

Those two are the important initial ones and we could then merge.  From then we need @Patowhiz in the discussion, because I think some of the others may need updating.

Wow!

rdstern (commented):
@N-thony I really, really like this.  I note the bug is also fixed where it used to default to the log window.  The names seem to work fine.  They are the suggested name when you save.  All good!

Usually I would suggest merging this, but I assume you may soon do some magic on the Insert part.  So I'll wait for that.

rdstern (commented):
@N-thony this really seems to be getting there.
a) The new line for the preview is only implemented for the first 2 tabs.  It should be there for all of them.
b) I suggest the Preview line that you added be moved down in the dialog, so it is in the same position for all 5 tabs.
c) It seems to return with the tick box ticked by default.
d) It could be a bit wider.  We will sometimes have quite complex commands.  I wonder if we will need to be able to scroll, or to sometimes have 2 lines there?

Very promising though!

rdstern (commented):
@N-thony this seems to be almost there.
a) Really pleased you made the preview 2 lines.  I was going to suggest that may sometimes be needed!
b) I tried an edit and thenpressed the Add.  It now included my edit, which is great.
c) So here is the only problem I found so far:  When I returned to the dialog the editr was empty, see below.  And it stayed empty whatever I clicked!

image

rdstern (commented):
@N-thony the preview is still oddly behaved.  I suggest whenever you enter the dialog that the `preview (and edit)` are `unchecked.`  Also they become unchecked whever you change either the tab or a different radio button.  Then, the action of checking the preview can easily become the current item? 

An alternative is, as now, that `preview and edit `keep their current setting.  Then, this setting must correspond to the tabs and buttons at the top - and not to anything else that has happened in R-Instat.  And the contents of preview must refresh whenever a tab or button is changed.

rdstern (approved):
@N-thony I am excited by this change and approving now.
@Patowhiz can you check and hopefully merge.  I have checkedthe starting code, namely the addition of the Library command, and then the writing of the data to a data frame.  I am keen for it to be merged even if some of the later code needs updating.

So, grateful if you can check the quality of the code and merge as soon as possible knowing that there will be some later options to alter, and perhaps new things to add.

We have to add the help context numbers as well.

@N-thony we should also delete the Edit > Script menu option.  It doesn't make much sense to use the dialog when there isn't a script open.  That's unless you would like it to open a new empty script window and then be ready to add a script, line-by-line.  I'm not convinced, but we could consider how that could work?

rdstern (changes requested):
@MeSophie that's looking great.
a) On the Aesthetics I thought everything inside the aes(  ) are the aesthetics, and that includes the Y and the X.  So I think the aesthetics box should include them?  @lilyclements am I Ok on this point?
b) On the bottom line can you move the control a bit further from the Save label and make sure that the space for the label is wide enough for translations.  Have a look in French.  Is it Enregistrer, or Enregistration?  Anyway it is longer than your current space.
c) Could you put a colon after Fill, so it is `Fill:`.  I am not sure we are consistent with colons in some other dialogs, but we are almost ok here?
Then we are ready to merge!

Patowhiz:
@N-thony thanks for this. I noticed a few oddities after a quick test. For instance,

1. I'm not sure what preview and edit are supposed to do. The text box is always blank. @rdstern  I'm not sure if we need the 'Edit' option in the insert script dialog, that's what the script window is supposed to do, I think this is just unnecessary duplication of functionality. I can see why we need a preview, that's a neat feature.
2. Whenever I navigated from tab to tab, the preview selection unchecked itself.
3. The preview seems to be unavailable in some of the other options like Library and Finally tabs.
4. In Get Data option, variables are not being shown in the selector. @rdstern notice we now have 2 'Add' buttons because of the selector. I would suggest renaming the 'Return' button to something like 'Insert Script' and 'Cancel' to 'Close'. Then we won't need all the multiple 'Add' buttons in different tabs. Of course this may need a wider discussion considering the base buttons will have different labels  compared to other dialogs. In my own opinion, the nature of this dialog is fundamentally different to other dialogs and therefore the variation can be justified.
5. Whenever I clicked on the 'Add' button, the scripts were not being added below the previously added scripts. So the last script added was on top instead of below the previous added scripts in the script window. Is that intentional?
6. The 'Remove' button in the 'Finally' section didn't add any script in the script window. And the label of the button makes it inconsistent with the other 'Add' buttons of the other options. This shouldn't be an issue if @rdstern agrees to my suggestion 4.

Patowhiz:
@rdstern @N-thony Why do we need the Rename button and dialog if as a user I can just double click on the script window tabs and rename them?

rdstern:
Ha, @Patowhiz you are doing my job - which is just to check "it" works.  Yours includes checking on the code!  And just to say I am really happy that you do this.  @N-thony occasionally does the same and @MeSophie is just starting.  I look forward to the others recognising that they could also do this sort of thing.

a) I agree with all your comments 2 to 6.  Very happy with all you observed.  If @N-thony is ok, then let's do them?
b) Your added comment on rename is (I think) what Antoine started with?  I was keen on a simple dialog for consistency, but it is a different rename to anything else.  @N-thony what do you think?  I am ok with the dialog, but I would also prefer to not have unnecessary buttons.  You decide, but I am veering towards going back to the double-click?

Your point 1 I would like to propose we keep the edit, at least for now (part of your point 1).  It isn't the default though and shouldn't be.

All our halfway dialogs allow the editing of a single command, before putting it in the script.  It also provides an easy correction place if some of the automatic provision of the command is wrong, or becomes wrong in the future.  It is sort of a visual part of the try button we have elsewhere.  We can't try it in situ, but if we did notice that it wasn't quite right for what we want, then we could do something about it here.

Patowhiz:
@N-thony I'm happy to retain the functionality of adding the script where the cursor is at initially. It will fit in nicely once we have the proposed single 'Insert' button that just closes the dialog when clicked.

@rdstern we could still retain the Rename button if you strongly feel so and have it do what the double click does. I can see a scenario where users with low computer literacy may be looking for such a button, but again, I'm not expecting such users to be looking at scripts.

@N-thony we have discussed and agreed with @rdstern on the following;

1. The Preview should always be visible and therefore you can remove the 'Preview' checkbox entirely. The Edit checkbox can be renamed to 'Edit Command'.
2. Changes made in the dialog should overwrite any edits made by the user.
3. The Add button can be removed in all the options. Then base buttons Return can be renamed to 'Insert' and 'Cancel' can be renamed to 'Close'. If you can (can be done later not urgent at all), you could have the Insert have similar functionality to Ok that we have for buttons. The options of Insert can be 'Insert and Close' (the default) and 'Insert and Keep.
4. The Finally tab can be renamed to Remove. Then the Remove button can be replaced with an input textbox that allows user just type in the objects he/she wants to remove.

rdstern (changes requested):
This is all very exciting.
a) The rename is now a double-click again.  However I have to press Enter, in the name field for the name to register.  If I type Roger and then click somewhere else, say in the script itself, the name reverts to Untitled1, etc?  
b) In the update dialog I really like the use of the 5 ordinary buttons at the bottom.  Why not!  
c) You need to make the Edit Line label wide enough for each language.  In French is may be Modifier la ligne, which is quite wide.  I don't want to move the editing field over too much, so could you allow the label to go over 2 lines if necessary.
d) I hope it will work that whenever you change anything, or refresh the top part of the dialog, then the script at the bottom is updated and changes.  It sometimes get empty even when I changed something.
Here is an example where I have been clicking and can't get any code in the line at the bottom.


e) Could you check the get data frame code.  I got this: - no I can;t now get anything here.  Byt when I got it earlier it used the data_book$get_data_frame function amd didn;t include the data frame name on the right-hand side?

f) I suggest, in this dialog, the default is not to have the Comment.  Then - if we do have the comment - it is transferred to then script window if we use the To Script button and not from the Ok button.  Anf it is currently done, but at the bottom of the script window.  Could it be before the line?

rdstern (changes requested):
@N-thony I note the renaming of the tab is now fine:
image
a) I don't know why the Edit Line is now in French, when I was still in English.  
b) It does show the label should change or the space for the label should be larger, perhaps 2 lines.  Alternatively call it Edit, or Line.  
c) The big point to change is that now the inserted line is always now at the top.  It should be at the cursor position.

Patowhiz:
@N-thony and @rdstern  is this ready for my review?

rdstern (approved):
@N-thony great I am approving
@Patowhiz I am very keen to get this merged as soon as possible.  I am conscious that some options may not work yet, but it is already useful in many ways.
We will inevitably have new issues on further improvements, but I would like them to start with this initial facility.  It builds well on the dialog you did earlier.

rdstern:
@N-thony that's almost a wow!
a) When you start, then Ok is not enabled.  It is, when you get back to datasets.
b) It then gives No example, even when there is an example - and copies it into exactly the right place

rdstern (approved):
@N-thony I take that back.  It imports nicely.
I am approving.  The slight problem with the Ok being enabled can be sorted later.  This is great.
@Patowhiz over to you.

Patowhiz:
@N-thony this now has conflicts. Could you resolve the conflicts then have it ready for me to review?
Thanks.

PR #8548, opened 2023-09-21
ChkRidges

Fidel365:
@rdstern , this is a small change in issue #8335 that only enables density ridges checkbox if there is a factor to edit


rdstern:
@Fidel365 I noticed that now seems to be the case in the histogram dialog anyway.  Is this still needed as well?  If not perhaps you can mention, and close this one.

rdstern:
@Fidel365 is this still needed?  If not, can you close it.

Fidel365:
I don't see this as a necessity

PR #8556, opened 2023-09-29
Allow the user to easily reopen previously opened scripts

### LINKED ISSUE #8487: Allow the user to easily reopen previously opened scripts

derekagorhom:
Partly fixes #8487 

This PR fixes part 1 and 2 of the Issues in #8487 

@rdstern , @N-thony can you review this PR 
Thanks

### DIFF
--- a/instat/dlgImportDataset.vb
+++ b/instat/dlgImportDataset.vb
@@ -716,6 +716,7 @@ Public Class dlgImportDataset
                               "Load Script From File", MessageBoxButtons.YesNo) = DialogResult.Yes Then
                     Try
                         frmMain.ucrScriptWindow.strActiveTabText = File.ReadAllText(strFilePathSystem)
+                        frmMain.clsRecentItems.addToMenu(Replace(strFilePathSystem, "\", "/"))
                     Catch
                         MessageBox.Show("Could not load the script from file." &
                               Environment.NewLine & "The file may be in use by another program or you may not have access to write to the specified location.",


rdstern (commented):
@derekagorhom that looks very good.
I think there is one component missing.  When I open a file it now puts it at the top of the list - and files of type R are now included.  Exactly what I was hoping.
However, when I save a data file that name is also registered in the same way.  So it also works from File > Save As > Save data As.
Could it also include the name if I save a Script file, so from File > Save As > Save Script As.
I assume using the Save button in the script window will automatically register in the same way.

I hope I am right @N-thony to ask for this?

derekagorhom:
@rdstern , @N-thony, I have made the changes you requested, i also ensured when a user loads directly from the load button on the script tab it automatically updates in the recent sub menu
can you review this 
thanks

derekagorhom:
@rdstern can you review this PR
Thanks

PR #8560, opened 2023-10-02
Added R packages

ChrisMarsh82:
Added:
    "agriTutorial", 
    "arm", 
    "rstanarm", 
    "statsr",
    "statquotes",
    "desplot",
    "formattable"

@derekagorhom I have removed rsplit as this seems to be a function not a package. Can you please check once the release is created if this is correct.


derekagorhom:
@ChrisMarsh82 sorry i looks like rsplit was a function but the package name is `collapse`, which has not been installed (Probably on my side) 
image

image

PR #8561, opened 2023-10-03
Made Column metadata consistent with Data View when selection is applied

N-thony:
@rdstern I'm not sure about the issue number this PR fixes, but I made the change as discussed on Skype so that the Metadata grid will reflect on the Data View when the selection is applied. Have a look.
image

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -463,8 +463,8 @@ DataSheet$set("public", "get_variables_metadata", function(data_type = "all", co
       if(self$column_selection_applied()) cols <- self$current_column_selection
       curr_data <- private$data[column]
     }
-    for(c in cols) {
-      col <- curr_data[[c]]
+    for (i in seq_along(cols)) {
+      col <- curr_data[[cols[i]]]
       ind <- which(names(attributes(col)) == "levels")
       if(length(ind) > 0) col_attributes <- attributes(col)[-ind]
       else col_attributes <- attributes(col)
@@ -487,7 +487,7 @@ DataSheet$set("public", "get_variables_metadata", function(data_type = "all", co
       #  col_attributes <- data.frame(class = NA)
       #}
       col_attributes <- data.frame(col_attributes, stringsAsFactors = FALSE)
-      out[[c]] <- col_attributes
+      out[[i]] <- col_attributes
     }
     #RLink crashes with bind_rows for some data frames with ~50+ columns
     #rbind.fill safer alternative currently


rdstern (changes requested):
@N-thony nice but needs a tweak,  I think it is just doing the first 3 columns, see below:

image

Good to check with wide data, when we may choose to not show the column metadata.

It is what I have wanted for a long time.  But before we merge I would like to check whether we could also have options on where to apply.

rdstern (commented):
@N-thony this is great - it now seems to work.  Here it is even for the extremes data:

image

I checked with @lilyclements though who is also excited by the option of being able to apply the select (or not) to these 3 alternatives (dataframe, metadata, dialog).  So I hope that isn't too  demanding a task.  

I had not expected this for the November upgrade even, so am really delighted!

rdstern (approved):
@N-thony agreed.  @Patowhiz over to you to check and merge

rdstern:
@N-thony I thought it is unlikely to do any harm and you suggested including it for now and then adding flexibility in where the select is applied later.  So, I'd be happy to have it in the version this week.

If it needs a check by @lilyclements first, then let's wait to see if she has the time this week.  Otherwise it could wait till the December version.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -456,13 +456,15 @@ DataSheet$set("public", "get_variables_metadata", function(data_type = "all", co
...
     }
     else {
       cols <- column
+      if(self$column_selection_applied()) cols <- self$current_column_selection
       curr_data <- private$data[column]
     }
-    for(i in seq_along(cols)) {
-      col <- curr_data[[i]]
+    for(c in cols) {
+      col <- curr_data[[c]]
```
lilyclements, line 467: @N-thony out of interest, why did you remove `seq_along`?
It means our `out` is now named not numbered. Does this affect code elsewhere or is this ok?

lilyclements, line 467: @N-thony out of interest, why did you remove `seq_along`?
It means our `out` is now named not numbered. Does this affect code elsewhere or is this ok?

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -456,13 +456,15 @@ DataSheet$set("public", "get_variables_metadata", function(data_type = "all", co
     if(missing(column)) {
       curr_data <- private$data
       cols <- names(curr_data)
+      if(self$column_selection_applied()) cols <- self$current_column_selection
```
lilyclements, line 459: From what I can tell, `column_selection_applied()` is either `FALSE` or `!curr_sel[["is_everything"]]`.
Have I misunderstood this? I'm sure you've checked this, but I just wanted to confirm is `!curr_sel[["is_everything"]]` logical (i.e., `TRUE`/`FALSE`)?

lilyclements (commented):
@N-thony just a couple of comments where I wanted to double check some code changes you've made. Apologies for double checking on them, I don't know this code as well as I'd like to so wanted to run by a couple of changes you made.

instat/static/InstatObject/R/data_object_R6.R
lilyclements: Ok great. So, `column_selection_applied()` gives a logical output then? If so, this looks good! :)

lilyclements (approved):

rdstern (commented):
@N-thony and @lilyclements it started ok, but then I got awkward and found this when I was trying to remove the selection:

image

I was a bit extreme, I know!  I made the selection and then did a summarise one variable and it worked fine.  Then I renamed variety to variety2 and noticed it is still there as variety in the selection.  

It is now a bit worse.  When I got rid of the error message, it took revenge and got rid of the data frame!

rdstern (approved):

Issue #8472, opened 2023-08-04
Cosmetic changes to the general Frequency tables main dialog


rdstern:
@lilyclements Here is the default dialog completed for a simple 2-way frequency table.

<img width="365" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/15e3c9c3-9e66-4ba5-9de7-7a6d3624b20b">

And here are the current results:

<img width="252" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/37e4f415-c5c2-46f9-af67-acf7485972ef">

Here are my suggestions, to make life easier for the user and to get clearer results.
a) @lilyclements do we need the Display box at all?  What's the point of the Summary Variable being the columns?  If we keep it, then I suggest we make the last factor in the list the columns always.  We are producing a table!
b) If we get rid of the Display box then we move the Significant figures control up.  Do we want to make that more general - perhaps then keep the display box, so we could alternatively contol the decimals in the display?
c) The results give NA for the missing counts.  Could we have options - perhaps we do need the display box - perhaps the default could be blank, but zero could be another option.  I think I'd like blank as the default?
d) Could the Margins be ticked by default.
e) When you click Percentages  - detail, small o for Of.  But could the last factor in the list be included as the default.  So, it would start with Variety in the example above.  It isn't often you want cell percentages, so let's make the default a commonly needed option.  

After @lilyclements has checked and possibly provided code then I wonder who could be given this. It should be simple.   @N-thony perhaps to discuss with Lily.  I am really keen we continue to progress on tables for the August release.

lilyclements:
> a) @lilyclements do we need the Display box at all? What's the point of the Summary Variable being the columns? If we keep it, then I suggest we make the last factor in the list the columns always. We are producing a table!

@rdstern you're right that we don't need this display box for the "Frequency" option since we don't have a column to summarise here. I think we want it for the Summaries option though.

> c) The results give NA for the missing counts. Could we have options - perhaps we do need the display box - perhaps the default could be blank, but zero could be another option. I think I'd like blank as the default?

@rdstern do you mean by this an option for what to display NAs as? I can code that in the R code

rdstern:
@Vitalis95 could you start with this issue, (which should be very quick) and then come back to #7999, which is very messy and needs some of the same changes as here.  
The only changes needed for now are as follows:
a) Move the Significant figures out of the display group.
[...]

PR #8562, opened 2023-10-03
Cosmetic changes to the summary/frequency tables

Fidel365:
Fixes partially #8472 , 
This fixes part (a, b and d), @Vitalis95 is working on part (c) of the issue because i have an issue with my DB browser that I'm trying to fix.
@rdstern this PR is ready for your review

### DIFF
--- a/instat/dlgSummaryTables.Designer.vb
+++ b/instat/dlgSummaryTables.Designer.vb
@@ -164,6 +164,7 @@ Partial Class dlgSummaryTables
         'rdoFactorVariable
         '
         Me.rdoFactorVariable.AutoSize = True
+        Me.rdoFactorVariable.Checked = True
         Me.rdoFactorVariable.Location = New System.Drawing.Point(14, 36)
         Me.rdoFactorVariable.Name = "rdoFactorVariable"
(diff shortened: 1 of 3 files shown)


rdstern (commented):
@Fidel365 great - this seems to work fine already.
Just one thing.  I tried with the usual survey data.  When I add a variable into the multiple receiver the dialog flickers around and black bits appear in a pretty crazy way.

Could you examine that part of the code, so it completes in a less dramatic way!  Then this is ready to merge from my point of view at least.

Fidel365:
@rdstern , this is ready for review

rdstern (approved):
@Fidel365 - nice one - well done.
@lilyclements this seems much better now, so I am approving.  Maybe you would like a quick check too?
It still needs a bit of perhaps Lily magic to display the NA differently, but I suggest we merge first and add this as a remaining item.  I am inclined to add it to the main dialoge as a Desplay Missing  option, and perhaps make blank the default.  NA is obviously another option.  We could specify I assume.  Should we suggest dot, or dash-dash, etc as others?
@N-thony could you please check the code.  Then, unless Lily needs changes we could merge?

PR #8580, opened 2023-10-17
Fixed Minor Translation Issue on Climatic>Unstack Dialog

### LINKED ISSUE #8576: Translation problem with Climatic > Tidy Data and others
@MeSophie here is the English version we have `Stack` and `Unstack`
image
but when you change it to French we get both `Depiler`
image

MeSophie:
Fixes #8576 
@N-thony  I made some small change on Unstack Dialog design, on data base and on Crowdin. Please have a look.
image


MeSophie:
@N-thony I removed the change in the data base.

rdstern:
@N-thony and @MeSophie great to see these improved.  
Here is that current meu:
image

a) I am not Francophone but I am surprised that both these options are `Depiler`.  Would `Empiler (Pivot plus long)` be better for the Stack option? And notice `long with l `in miniscule here as well for consistency?
b) The entry `Valeurs non numeric` is the only other entry with capital letters?

I assume if you make these changes they will also (automatically) also be made in the Prepare menu?

c) I am getting picky now, but could the Up-Down control in Hill missing values move right just a bit. 
d) In dates there is still the Use date translation to change.

MeSophie:
> @N-thony and @MeSophie great to see these improved. Here is that current meu: image
> 
> a) I am not Francophone but I am surprised that both these options are `Depiler`. Would `Empiler (Pivot plus long)` be better for the Stack option? And notice `long with l `in miniscule here as well for consistency? b) The entry `Valeurs non numeric` is the only other entry with capital letters?
> 
> I assume if you make these changes they will also (automatically) also be made in the Prepare menu?
> 
> c) I am getting picky now, but could the Up-Down control in Hill missing values move right just a bit. d) In dates there is still the Use date translation to change.

image
@rdstern It is better to for example **Transform Text** to `Transformer Le Texte` or `Transformer le Texte`?

image
Should I align this control with the lower radio buttons?
image

rdstern:
@MeSophie sorry - I was not clear on which dialog.  I have added a picture into your comment of earleir today.

Issue #8527, opened 2023-09-06
Further improvements to the Describe > Graphics > General dialog


rdstern:
This could possibly be done by @MeSophie who already has a pull request on this dialog.  I need to check with @lilyclements before the new work starts.

a) Consider making the buttons towards the bottom smaller and adding `Titles, X-Xais, Y-Axis` after Facets.
b) Add a simple `legend` checkbox, on the right, default unchecked and with the options as given in the Themes sub-dialog.
c) Add the usual `Flip Coordinates` Checkbox, that is on many dialogs.  Default unchecked
d) Simplify the Y variable (so get rid of the Single/Multiple option) and call it Y Variable.

More to come:
a) How to do a bar chart?
b) Second Y variable
c) Second data frame

Patowhiz:
I recommend using 2 selectors if the dialog needs 2 different data frames.  This makes it consistent with other dialogs.

lilyclements:
@rdstern these sound like useful and interesting changes to me. Just one small question - 

> b) Add a simple legend checkbox, on the right, default unchecked and with the options as given in the Themes sub-dialog.

Is this for changes like legend position, or for quite a few more changes? I like this dialog because it looks "clean" and simple and worry about it getting complex and cluttered!

MeSophie:
@rdstern please can you clarify a bit more b) and f)?
For b) Are the options given in the Themes sub-dialogue box you are talking about the position? If so, when the box is ticked should the control for choosing the position appear?
 For f) when the box is ticked, should a second receiver y appear or should the sec_axis function be added?

rdstern:
@N-thony  after talking to David, I wonder if you could quickly adapt or construct this new used control? 

<img width="209" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/7c2c6986-e15e-4362-9517-bd9e48d38bc2">

It may be that it can use the existing control?

I am keen for the Add button in the geom box to have the same sort of down arrow as the Add that is part of the data selector.  When clicked it would provide a list of geoms, (given above) and they would add into the control on the left, without the user going to a sub-dialog.

If just a few are added, then I suggest @MeSophie could complete the control.  But I suggest it could do with a start by you - or Patrick?

lilyclements:
This has been completed

PR #8585, opened 2023-10-20
Further Improvement On Graphic Dialog

MeSophie:
Fixes Partially #8527
@rdstern @N-thony  I Solved Part 1,2-a.
I have made some change in data base for part 2-a.
image

For part 2-b the group checcbox is still there.
image

I added the down arrow and the list of some Geom So @N-thony you can just continue with the coding part to make these geom working.


rdstern:
@MeSophie this is just great.  The list of geoms I saw is not as good as those you included above.

image


a) Do we need the bit on the left of the pull-down?
b) Could they be single-spaced - not sure, but that would enable us to include more geoms?
c) Could we add a scroll bar and I'll list a few more geoms

rdstern:
@MeSophie and @N-thony I am now understanding the 2 pull requests a bit better, so #8589 and this one.
a) On possible improvements to the pull-down, the main one needed is to add a scroll bar.  Then we can include more geoms without the control becomeing too long.
b) Second is to quest the shady bit on the left, whether we need that?
c) Then I quite like the extra space between the geoms, but wonder whether it is needed?  @N-thony is better than me on these cosmetic bits, so I leave that to you.
d) Then @MeSophie there is one correction needed to the list of geoms in your list above and I would like you to add some more.  The correction is your rug smooth which should be geom_rug and separately then geom_smooth.
e) So on adding the the list could you please add contour, dotplot, histogram, mosaic, parallel_slopes, tile and violin.

This is looking sooooo good!

rdstern (approved):
@N-thony that has all the geoms I like - and I do also like the new ones!  So, now I think they both need to merged and this will leap into action!

rdstern:
@N-thony and @MeSophie is this stuck or can it be resolved easily?

PR #8591, opened 2023-10-23
Fixed small bug in Column summaries sub-dialog

### LINKED ISSUE #8588: Small problem in Column summaries sub-dialog
@N-thony this should be able to be fixed by anyone.  It is a much unused sub-dialog.  I explain the example:

In R-Instat > Open from Library > Datasets > ACRAdmissions data.

Then Prepare > Data Reshape > Column Summaries.  Complete as shown below.  Also just one summary the sum.

image

After the summaries button, press the Percentages button - bottom right on the main dialog.  Tick the box, and you should get the Fig shown above.  So the receiver isn't shown!

If you click on the one button at the top that is enabled, then the receiver appears.  From then on, it appears all the time.  It is just the first time you go in - at least with this sequence.

derekagorhom:
Fixes #8588 
I have fixed the problem with the receiver not appearing when the sub-dialog is used for the first time 
@N-thony , @rdstern can you review this
Thanks

### DIFF
--- a/instat/sdgProportionsPercentages.vb
+++ b/instat/sdgProportionsPercentages.vb
@@ -84,8 +84,10 @@ Public Class sdgProportionsPercentages
     End Sub
 
     Private Sub ucrChkProportionsPercentages_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrChkProportionsPercentages.ControlContentsChanged
+        'temporarily fixes issue #8588
         If ucrChkProportionsPercentages.Checked Then
             rdoByFactors.Checked = True
         End If
     End Sub
+
 End Class
--- a/instat/sdgProportionsPercentages.vb
+++ b/instat/sdgProportionsPercentages.vb
@@ -82,4 +82,10 @@ Public Class sdgProportionsPercentages
             ucrReceiverColumn.SetMeAsReceiver()
         End If
     End Sub
+
+    Private Sub ucrChkProportionsPercentages_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrChkProportionsPercentages.ControlContentsChanged
+        If ucrChkProportionsPercentages.Checked Then
+            rdoByFactors.Checked = True
+        End If
+    End Sub
 End Class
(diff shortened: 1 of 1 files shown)


rdstern:
@derekagorhom I tried and it still seems the same:

image

Note I import the data, use the summaries as shown, then first the summaries button, and then the probability one.

instat/sdgProportionsPercentages.vb
```diff
@@ -83,4 +82,10 @@ Public Class sdgProportionsPercentages
             ucrReceiverColumn.SetMeAsReceiver()
         End If
     End Sub
+
+    Private Sub ucrChkProportionsPercentages_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrChkProportionsPercentages.ControlContentsChanged
+        If ucrChkProportionsPercentages.Checked Then
+            rdoByFactors.Checked = True
+        End If
+    End Sub
```
Patowhiz, line 90:
```suggestion
    Private Sub ucrChkProportionsPercentages_ControlContentsChanged(ucrChangedControl As ucrCore) Handles ucrChkProportionsPercentages.ControlContentsChanged
        'temporarily fixes issue #8588
        If ucrChkProportionsPercentages.Checked Then
            rdoByFactors.Checked = True
        End If
    End Sub

```

Patowhiz (changes requested):
@derekagorhom I have suggested a comment addition. Check if you agree then we can merge this.

rdstern (dismissed):
@Patowhiz I tried again and am happy to approve again.  I didn't do anything that provioked the comment though.
You also asked on the same dialog about the `j 1   Help` which does look a bit odd.  It is 2 separate controls.  Ideally the Help button would say R Help.  There is a comment to say why we have j as 1 in the other control.  It is all to do with the md measure.

Patowhiz (approved):

derekagorhom:
@Patowhiz should i look into the bug you found in another pull request?

Patowhiz:
@derekagorhom thanks the changes made here are enough. Once @rdstern clicks on approve, I will merge.

rdstern (approved):

PR #8592, opened 2023-10-23
Fixed Minor Developer Error Message in Rename Dialog 

### LINKED ISSUE #8586: Prepare > Data Frame > Rename gives an annoying sort of error message
@N-thony this is annoying, but not serious.  Perhaps ytou can allocate or even fix, while you complate #8540 - though it is different.  And I assume should be easy for anyone to fix.
a) Go into any dataset, I used the survey.
b) Go to Prepare > Data Frame > Rename
c) Click to include variable labels.
d) Add one or more labels
e) Press Ok
All is fine so far.
f) Go back into the dialog - it gives the common error we often get.  I would not mind an explanation and maybe a more general fix, or perhaps a better message?

image

g) When you go in to the dialog, it has not remembered that we asked for variable labels before.  So the option is unticked again. When you tick all is fine - it has remembered the previous labels entered.

MeSophie:
Fixes #8586
@rdstern , @N-thony  I have fixed the Developer error message. Please Have a look.


rdstern (approved):
@MeSophie great - looks fine now

@N-thony if you are happy this can be merged

PR #8597, opened 2023-10-25
Insert script and renaming tab in script window 

### LINKED ISSUE #8570: Add a script tab to new Insert dialog

N-thony:
Fixes partially #8487
Fixes partially #8151
Fixes #8570

@rdstern @Patowhiz this PR replaces PR #8544 due to conflicts, have a look. There is still a lot of work going on the script window, it will be good to get this merge quickly.


Patowhiz:
@rdstern could you test this and approve while I'm reviewing?

Patowhiz:
@rdstern we have looked into multiple issues with @N-thony in regards to this PR.

The final option is the one we couldn't agree on what you wanted. My understanding is, it's meant to produce the `rm()` function for removing objects from R global "memory".

Also I conspicuously noticed the **ToScript** button having the same functionality to **Ok** button. Any reason as to why we can't hide one of them to emphasize that the dialog is unique and simply inserts scripts to the script window?

Once @N-thony fixes the issues that we have discussed I can review.

rdstern (commented):
@N-thony this is looking really excellent.  @Patowhiz I note first what seems to me to be working well.
a) It fixes the issue that when you first click to open the script window it opened into the log window.
b) You can double-click on a tab and add a name.  It then suggests that name if you click save.
c) The big thing is that there is a new Insert button.  I have checked the first 2 elements of this, namely you can insert a library line.  And you can insert the line to add the data into an R-Instat data frame.  
d) We need to be able to do more, like getting data into a script and copying a graph, etc.  I have yet to check those - and they may become a new issue.  And they may need @Patowhiz 
e) The new big one is the copying of scripts from packages.  I am very happy to see this.  There is a bug, to be fixed that @N-thony is aware of.  Then it will be great to add.

rdstern:
@Patowhiz I know that the ToScript and Ok do the same in this dialog.  I am quite happy to still allow both for consistency with the other dialogs.  But I don't feel strongly about it.  If you disable one of them, will you keep the ToScript, or the Ok?  So I'm happy for you and Antoine to decide and implement whatever you agree on.

I am disappointed to lose e) because it is such a big deal.  And it seems to work for most scripts/packages.  But if you feel it will take time to sort out properly, then let's make it another PR.  I hope it will be in time for this new version though.

rdstern (changes requested):
@N-thony it looks great.
Except it is only showing data examples.  It should also allow examples from any function.  For example agricolae package has many functions, like AMMI and they also have an examples section.

And there are some packages that have useful examples, and don't have any data sets.  For example checkmate has many functions, some with examples, but no data files.

rdstern:
@N-thony I had not thought you would try that way?  I would like you to list the functions, just as you list the datasets.  Then we choose and you look for the Examples bit.
I would be very happy if there could be the two separate lists, but you could just list everything in a single list - as ythere is when you just choose the package in the Help dialog.

rdstern (approved):
@N-thony that's fine - we don't need the description for the functions.

When I upload an example it currently gives me some extra lines, see below.  So it is starting somewhere earleir than the Examples.  This isn't a problem (because they have hashes in front.) but a bit of a surprise.

I am approving anyway,  I am happy with what we have now, and would really like this to be merged.  @Patowhiz over to you.

image

ChrisMarsh82 (approved):

instat/dlgScript.vb
```diff
@@ -199,156 +215,165 @@ Public Class dlgScript
...
-        Dim strGetScript As String = clsSaveGraphFunction.Clone.ToScript(strScript:=strAssignedScript)
-        AppendTextScript(strAssignedScript.Trim & Environment.NewLine & strGetScript)
-        AddAssignToString(ucrSaveGraph.GetText)
+    Private Sub GetExample(strTopic As String)
+        Try
+            If Not String.IsNullOrEmpty(strTopic) Then
+                clsLibraryExpFunction.AddParameter("topic", Chr(34) & strTopic & Chr(34), iPosition:=0)
+                If clsLibraryExpFunction IsNot Nothing Then
+                    Dim strExampe = frmMain.clsRLink.RunInternalScriptGetValue(clsLibraryExpFunction.ToScript(), bSilent:=True).AsCharacter(0)
+                    Dim strResult = RemoveDoubleHashD(strExampe)
```
Patowhiz, line 332: Why do we need to remove double hashed D, that is, **##D**?

instat/dlgScript.vb
```diff
@@ -379,4 +399,170 @@ Public Class dlgScript
...
+        SetPreviewScript(clsImportNewDataFrame, strData)
+    End Sub
+
+    Private Sub ucrComboGetPackages_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrComboGetPackages.ControlValueChanged, ucrPnlExample.ControlValueChanged
+        If strSelectedPackage <> ucrComboGetPackages.GetText() Then
+            strSelectedPackage = ucrComboGetPackages.GetText()
+            If strSelectedPackage <> "datasets" Then
+                clsLibraryExpFunction.AddParameter("package", Chr(34) & strSelectedPackage & Chr(34), iPosition:=1)
+            Else
+                clsLibraryExpFunction.RemoveParameterByName("package")
```
Patowhiz, line 437: Why do we need to remove the package parameter when the selected package is not **datasets**?

Patowhiz (commented):
@N-thony, I recognize that this PR received expedited approval from @rdstern and was  merged by @ChrisMarsh82.

After discussions with @rdstern, I've undertaken an effort to modify and enhance the changes introduced here, in PR #8629.

In my review comments, I've posed a few questions to you. Your feedback will play a crucial role in enabling me to implement the "get examples" feature in PR #8629. 

Thanks.

instat/dlgScript.vb
```diff
@@ -199,156 +215,165 @@ Public Class dlgScript
...
+    End Sub
 
-        Dim strAssignedScript As String = ""
-        'clone the function first because the ToScript function modifies the contents of the function.
-        Dim strGetScript As String = clsSaveGraphFunction.Clone.ToScript(strScript:=strAssignedScript)
-        AppendTextScript(strAssignedScript.Trim & Environment.NewLine & strGetScript)
-        AddAssignToString(ucrSaveGraph.GetText)
+    Private Sub GetExample(strTopic As String)
+        Try
+            If Not String.IsNullOrEmpty(strTopic) Then
```
Patowhiz, line 328: Do we have a possibility of a null or empty topic? Any example?

instat/dlgScript.vb
```diff
@@ -199,156 +215,165 @@ Public Class dlgScript
...
+                clsLibraryExpFunction.AddParameter("topic", Chr(34) & strTopic & Chr(34), iPosition:=0)
+                If clsLibraryExpFunction IsNot Nothing Then
```
Patowhiz, line 330: When could `clsLibraryExpFunction ` be null?

rdstern:
@Patowhiz  I am contributing here, just to emphasise that I see this work by @N-thony and you to be a really important feaure in R-Instat, and so much the better if these questions result in you two working more closely together on the code generally. 
I'm not a programmer, but I view these questions as implying the importantance of including good explanatory comments in the code.  They explain why the code is there, and the code itself is just the "what".

That's the parallel of our underused comment feature in every dialog in R-Instat.  Anyway my purpose in this message is to suggest that Antoine could reply, where possible, by adding the relevant comments into the #8629?  @Patowhiz is this sensible?

Patowhiz:
Thanks for your input @rdstern. I agree that the collaboration between @N-thony, the broader team and myself is significant for enhancing R-Instat. Your perspective on the importance of explanatory comments in the code is spot on. Comments are indeed invaluable in helping developers understand the "why" behind the code, which is just as crucial as the "what." My hope is the broader team will take this into account as we continue our work on R-Instat.

Patowhiz:
@N-thony, I wasn't initially aware that this PR had been superseded by PR #8619. Nevertheless, I'd like to emphasize that my comments from this PR are still applicable and relevant to PR #8619.

instat/dlgScript.vb
Patowhiz: I've been answered by @rdstern [comment](https://github.com/africanmathsinitiative/R-Instat/pull/8629#issuecomment-1806332036) in PR #8629.

PR #8583, opened 2023-10-19
Bug fixes on the File > New DataFrame > Lists

N-thony:
Fixes #8574
@rdstern this fixes the bug with `US Airport codes` and `venues`. Have a look

PR #8608, opened 2023-11-01
New pr

### LINKED ISSUE #8574: Odd serious bug with file-new data frame on some lists
@N-thony I think this may be one for you, because it may be tricky for the others.  But up to you.

With File > New Data Frame > Lists, most of the lists are given fine.

But a couple consistently crash out of the program.  Running on version 7.16 it gives the crash error.  When I continued the screen for the example was like this:

image

The data were going across rather than down!

The only 2 examples I have found of this are in the category geography

image

They are `US Airport codes` - result as shown above - and `venues`.

Here is the error in reogrid:

image

My guess is that these lists don't have any data?

Patowhiz:
Fixes #8574.
Replaces PR #8583.

@rdstern, @N-thony, this PR serves as a response to my comments in PR #8583. Ensuring that these concerns are appropriately addressed here is pivotal for future enhancements in similar functionalities.

@N-thony, I invested a significant amount of time comprehending the underlying goals of the function. This enabled me to adopt an alternative approach to resolving the issue without compromising the expected functionality. I've taken the liberty to include comments in the code that elucidate the purpose of each code block, which I believe will greatly aid in your review.

@rdstern, I'd like to draw your attention to the consistency between the contents of the data viewer and R viewer. This was the specific aspect I was emphasizing in the tests. While PR #8583 effectively resolves the issue that was causing R-Instat to crash, it does introduce regressions in how the contents of list(LT) column types are displayed. Regressions can occur, but I strongly recommend documenting them explicitly, ideally after consulting with the original code author if feasible.


rdstern (approved):
@Patowhiz I can't tell the difference, of course!  It seemed to be fixed before and this version works fine too.  I'm really liking that it looks the same as in the R viewer.
@N-thony I assume it will be useful if you check, and also follow what's different.  Then let's merge!

PR #8549, opened 2023-09-21
Improve Colour Option- for Ggplot2 Graph

MeSophie:
Fixes partially #8282 
@rdstern , @N-thony  I added `Color Palette panel` in `colour tap` in `plot options`. Please can you have a look? I didn't work on part a) of this issue. I'm waiting for more explanation from @rdstern.

Issue #8282, opened 2023-04-23
Improve the colour options for ggplot2 graphs


rdstern:
@MeSophie could you also contact Wezzie on this.  I still can't contact her directly.

We need to improve the way colours are possible in our plots.  This looks like a great [article.](https://www.datanovia.com/en/blog/ggplot-colors-best-tricks-you-will-love/#:~:text=Change%20ggplot%20colors%20by%20assigning,or%20to%20the%20fill%20arguments.)  

We already have a Colour tab in the sub-dialogue.  This is where we need improvements.

I will write more later

lilyclements:
> My R knowledge isn't enough to know whether the palettes stay for a long time or just while the dialogue is set up? If we uncheck (Reset) do we get back to the default. I hope so. Good for someone the check these suggestions before they are implemented.

@rdstern in terms of R (not R-Instat dialogs), the colours is just adding on another layer. Therefore, it is just an option if you run it for that plot. It doesn't "remain" for all plots. 

E.g.

```
# Default colours before we specify anything
ggplot(iris, aes(x = Sepal.Length, y = Sepal.Width, colour = Species)) + geom_point()

# Adds a grey colour scale
ggplot(iris, aes(x = Sepal.Length, y = Sepal.Width, colour = Species)) + geom_point() + scale_colour_grey()

# Adds a different colour scale
ggplot(iris, aes(x = Sepal.Length, y = Sepal.Width, colour = Species)) + geom_point() + scale_colour_ordinal()

# Back to the default because nothing is added. 
ggplot(iris, aes(x = Sepal.Length, y = Sepal.Width, colour = Species)) + geom_point()
```

Have I understood your question correctly?

lilyclements:
Are we also only using `scale_colour_viridis` at the moment? There's loads of different options, some in the `ggthemes` package. I like to use `scale_colour_colourblind`. I prefer it to `viridis`' colour-blind scale - but I do not know what the colourblind community feel, which is a much more important point of view!

```
ggplot(iris, aes(x = Sepal.Length, y = Sepal.Width, colour = Species)) + geom_point() + ggthemes::scale_colour_colorblind()
```

rdstern:
@MeSophie I wonder if you could try this one?

PR #8609, opened 2023-11-01
Improve Colour Option for Ggplot2 Graph

MeSophie:
Fixes partially #8282
@rdstern , @lilyclements This PR replace #8549. I added some new options on color tap in plot options dialog. You can start checking when I'm still working on auto detection variable type and adjusting  the taps size.


MeSophie:
> @MeSophie you can pull my branch called `color_br` it fixes of the Axis type for color and added few geoms requested by @rdstern

@N-thony thank you. @lilyclements ,@rdstern  this PR is ready to be test.

rdstern (changes requested):
I am pleased we can test again.
a) The page seems sometimes to get very big - and empty for example:
image

I think on some screens the bottom controls will not be accessible.
b) I am not sure how to use the different controls but I got nice blue colours and then they always stayerd the same, see the code with scale_fill_brewer() almost whatever I did?
c) There are 4 checkboxes - are they needed, see below:

image


 - so you can have any combination ticked, even all 4.  Should they be radio buttons instead?  And I am not sure about pallete and scale?  Could you have a colour palette and a fill scale?

MeSophie:
image

@rdstern Certain controls appear depending on whether you are on ggtheme or viridis

image

PR #8612, opened 2023-11-02
Fixes Issue with Recent file List becoming too wide

### LINKED ISSUE #8595: Recent files list can become very wide.  Is this a problem?
I show an example and wonder what happens with other software?

image

You see that the file name is unreasonably long and hence the recent files is very wide.  I wonder what would happen if the file name were twice as long?  And we currently don't have any ability to edit this list?

What do other software do?  This is not a problem with the front screen, but it is the same width at the bottom of the file list.

derekagorhom:
Fixes #8595 
This PR is ready for review
@rdstern can you test this 
Thanks

### DIFF
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -746,7 +746,7 @@ Public Class ucrDataView
         Dim tooltip As New ToolTip()
 
         ' Set the tooltip texts for the labels.
-        'tooltip.SetToolTip(linkMenuItem, linkMenuItem.Text)
+        tooltip.SetToolTip(linkMenuItem, linkMenuItem.Text)
 
         'add the link control.
         panelRecentMenuItems.Controls.Add(linkMenuItem)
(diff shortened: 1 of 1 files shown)


rdstern (changes requested):
@derekagorhom you have done it the other way round from what I thought.  I didn't mind it being wide on the initial screen.  It didn't seem to do any harm there.

On the other hand, from the file menu it looks odd:
image


And also the toolbar option is still wide.

image

What do you think?

I also wonder if 25 is a bit low.  You could try (say) 35?

derekagorhom:
@rdstern i have made the change, also do you still want me to undo the changes in the main screen?

rdstern (changes requested):
@derekagorhom You still seem to be doing things one at a time:
a) Apply this also to the toolbar option, see below:
b) Change your limit to a higher value, say 35?
c) I am happy that the change is applied to the front screen, so keep that.

image

rdstern (approved):
@derekagorhom that looks fine.  Thanks.
@N-thony if ok by you, then happy, but not vital, to include it in the next version.

derekagorhom:
@N-thony I have made the changes, can you review again
Thanks

rdstern (approved):
Looks good

Issue #8541, opened 2023-09-14
Scripts and sub-dialog for gt tables


rdstern:
The gt package gives many useful examples of scripts [here](https://gt.rstudio.com/news/index.html).

And the gt package has a script mainly for sparkline tables [here](https://jthomasmock.github.io/gtExtras/articles/plotting-with-gtExtras.html). 

I have adapted one script from gt so it (almost) runs in our current script window.  I fould 4 problems with the script as provided and also wonder whether 2 of these may be fixed in pull request #8494?

a) There is a new symbol in base R from version 4.1 to handle piping, which is used a lot in the gt scripts.  This is `|>`,  Previously there was `%>%`, which the script window seems to handle.  It would be good to accommdate both.
b) I think currently commands with `%>%` need to be on a single line to be processed.
c) The results are automatically in a browser, rather than our output window.  Our current gt dialog produces the results in the output window.  Maybe we can add an R-Instat line, or edit the script in some other way, if we want the results in the output window.  This may be that tables parallel of the graphs in other scripts that only appear in the R-viewer.  At least in the short term I am fine with using a browser by default and having to tweak the script to use our own windows.
d) My adapted script still generates a few errors.  This may be errors I have introduced in the changes I made.  I am keen to know the time-scale on items a) and b) before investigating.  

```
library(gt)
library(tidyverse)

# Take the `islands` dataset and use some
# dplyr functionality to obtain the ten
# biggest islands in the world
islands_tbl <- tibble(name = names(islands),size = islands ) %>% arrange(desc(size)) %>%  slice(1:10)

# Display the table
islands_tbl
#> # A tibble: 10 × 2
#>    name           size
#>    <chr>         <dbl>
#>  1 Asia          16988
#>  2 Africa        11506

# Create a display table showing ten of
# the largest islands in the world
gt_tbl <- gt(islands_tbl)

# Show the gt Table
gt_tbl

# Make a display table with the `islands_tbl`
# table; put a heading just above the column labels
gt_tbl <-  gt_tbl %>% tab_header( title = "Large Landmasses of the World", subtitle = "The top ten largest are presented" )

# Show the gt Table
gt_tbl

# Use markdown for the heading's `title` and `subtitle` to
# add bold and italicized characters
gt(islands_tbl[1:2,]) %>% tab_header( title = md("**Large Landmasses of the World**"), subtitle = md("The *top two* largest are presented")  )

# Display the `islands_tbl` data with a heading and
# two source notes
[...]

PR #8614, opened 2023-11-02
Scripts and sub dialoguefor gt Table

Fidel365:
Fixes partially #8541 , @rdstern  you can try testing it from here


rdstern:
@Vitalis95 this is the latest attempt by @Fidel365 on his important tables work.  It has always had conflicts, so I have never been able to check.  He is now not around till Monday.  I wonder if you could check and resolve the conflicts?  Don't spend too long.  If necessary pass the task to Antoine.

Issue #8521, opened 2023-08-31
Adding a simple script to read multiple Excel files?


rdstern:
@Patowhiz we can import from multiple Excel sheets in the same worksbook.
If we have single sheets exported as csv files then we can read a directory of csv files (or sav files from SPSS).
But we can't (yet) read from a single (first) sheet of multiple Excel files.

I think you said that once loops were easy from the script window, then you could adapt the reading of a single sheet to enable this feature.

a) Well now we can handle loops in the script window.
b) And I am very keen to build up examples of scripts, for the R-Instat library, just as we have lots of datasets in the library.  It is becomeing a data and script library!
c) And this script would also be very useful for teaching purposes.
d) And I had examples where I could have used it on the recent workshop at CIMH.  Some participants had each station in a separate Excel file.

Do you have an example we could use.  Otherwise can you think of a better example than making one from the Cameroon data?  This is currently used for teaching data preparation, because each year is in a separate sheet.  We could make another version in its own directory, with each year in a single sheet.

The reason the script would be good for teaching is that it caters for a rare situation.  It is also a situation I don't want to encourage, i.e. lots of very simple Excel files.  It would be better to put them into a single book.  But I would like to explain the idea also that we want to avoid the "multiple copy-paste syndrome", whenever possible, i.e. when you repeat something routinely, then you would like the computer to do it, rather than you doing it!  

So if we have 3 Excel files, then we do it one-at-a-time.  But if we have 30, or 300, then we get the computer to do it!

By the way we could also add to this teaching point nicely to show how R permits us to do loops efficiently in many situations.  We could give a simple example, with lists, using the calculator.  That's already there.

Patowhiz:
@rdstern very excited to hear that @lloyddewit has added loops support. I'll test this. 
@rdstern we certainly need to test this feature in the script window (to know it's limits) then think of how it can be added in dialogs that may require it.
I understand that we have a bug when importing data frame through the script window, the data frame is imported twice. That should be a quick fix in the `RunScript()` of `clsRLink`. But I cannot touch it because @lloyddewit is looking at improving the same subroutine in another PR.

rdstern:
@lilyclements no, not yet.  And we sometimes could have used that, i.e. many Excel files, with a sheet on each.  Usually that can be quite old files, so sometimes xls.

[...]

Issue #4445, opened 2018-03-09
Giving R commands in R-Instat - options by context!


rdstern:
1. We have always felt that users should be encouraged to migrate from R-Instat to RStudio when they need to give commands to R, rather than just using the dialogues.  This route continues and we should also continue to check that it is as simple and encouraging as possible. 

2. An alternative is to give R commands within R-Instat.  David and Danny have always been reticent about this options, and for good reasons.  They include:
a) We don't have good error trapping - we just throw the commands at R.  So if they work, that's great, but if not, then the messages are usually difficult to use to make the necessary corrections.
b) There is nothing to stop us giving commands that mess the user up completely.  In particular commands like q() would throw us out of R-Instat - unceremoniously, and other commands may interact with our work and could result in it becoming a mess. (We should investigate that to advise on how to minimise the risk.
c) The sets of commands look messier than pure R (and pure R can be messy enough!)  In particular some sets of commands could be totally separate to our Instat object (they will be pure(ish) R, but why use R-Instat then - I explain why below!)  Or they make use of our Instat object and then they do look rather complicated.

Despite these issues I claim that this feature in R-Instat is so important that we keep working on it, to make this route as acceptable as possible.  Here is why, through 2 examples:

1. **Climatic audiences:**  They will love R-Instat.  But there are so many more analyses in R packages (some well and some poorly written) that will take ages to even consider adding to R-Instat.  In addition, there is a risk that adding many of them will clutter R-Instat and make it more complicated to use for the routine climatic work.   When these users are ready, they have access to everything from RStudio, but there will be plenty who just want one or two specific analyses.  And these may well be different for each Met service.  While they wait we could have a set of commands that could be provided (at least temprarily, and then executed from the script window.  If it is modest, then we could do it from R-Instat, and once it gets big, then we either recommend the RStudio route more strongly, or (in the other direction) check whether we could add a new dialogue.
2. **Agricultural Research audiences**  I still think we will be providing quite a lot that is useful within the R-Instat menu structure.  And this should be particularly following the forthcoming GCRF work.  However (for example) Ric stated that his reticence was because of the following:
[...]

Issue #4228, opened 2017-11-22
Adding a loop in R-Instat - using purrr?


rdstern:
Programmers used to love loops for doing repetitive operations.  And avoiding repetition is important.  Hadley Wickham states (in his book on data science and perhaps elsewhere) that if you do the same copy-paste more than twice then you should be looking for a better way to do this.  This is a nice reference to the repetitive way many people use Excel, when they should perhaps look to use a tool that helps them become more efficient and reliable in their work.

R-Instat is menu driven and we can't yet do loops.  But, in situations where loops are needed there may sometimes be another way to get the same result.  

In the longer term I suggest we will need some form of looping ability in R-Instat, e.g. to repeat an analysis for multiple (climate) stations.  But I want to share an example where I was convinced earlier we needed a loop to avoid repetition.  And looking on Stack Overflow confirmed my view.  But??? Well read on.

[Dodoma for multiple plot.xlsx](https://github.com/africanmathsinitiative/R-Instat/files/1494867/Dodoma.for.multiple.plot.xlsx)

Here is the data file.  I want to examine the 2 columns on sunshine hours.  First I want to do scatter-plots by month.  Then I want the correlations for the data for each month.  
It is odd to me that I can get the whole scatterplot, but not easily get the correlations.  I can do it step-by-step and this is what the participants are doing.  Many have data like this!

a) Go to the correlations dialogue.
b) Set up a filter so just using data for January.  Call it Jan.
c) Get the correlation
d) Return to the dialogue - easy!
e) Change the filter to Feb - well set up a new one for now.  We can't yet edit filters.
f) Get the correlation.
g) Repeat steps d) to f) for the remaining 10 months!

I am doing a loop - sort of copy-paste 12 times.  Is there an easier way?

If we are using R "properly", then we write a little function:

_xx <- data.frame(group = rep(1:4, 100), a = rnorm(400) , b = rnorm(400) )
Compute the correlations:

library(dplyr)
xx %>%
  group_by(group) %>%
  summarize(COR=cor(a,b))
The output:

Source: local data frame [4 x 2]

  group         COR
  (int)       (dbl)
1     1  0.05112400
2     2  0.14203033
3     3 -0.02334135
4     4  0.10626273_

I wonder what we can do in R-Instat on similar lines?  Is this an apply, lapply, etc.  Could we copy the command into the Script window and add something for now?

On the other hand can we do it already.  Here is a "solution" within the current R-Instat - which could be improved.

[...]

PR #8629, opened 2023-11-05
Insert script dialog

Patowhiz:


rdstern (commented):
@Patowhiz it is looking very promising.
a) Checked the previous feature that it should go to a script window.  It used to go to the log window and that caused problems.  It was fixed in @N-thony update.  That's ok here too
b) Checked on renaming of the tabs that was done by Antoine.  The renaming is still fine.  But it used to suggest that rename when Saving.  That's lost.  Good to re-instate it.
c) Then into Insert.  I tried to Common feature first.  They seem to work individually, but not both together.  I am happy to have one at a time, so could they have radio buttons?  That would keep it simple.
d) Could we have a better name.  I'd prefer `Commands` or `R Commands`? .Can you think of anything better than those?
e) I next tried the examples - It looks fine and seems to work ok, and as before.  The preview ignores the new lines.  It might be simpler to check, if it included them, but not if that is a big deal.
f) I next tried saving the data frame.  Why is there the extra first line here?

image

And when I press To Script I get a third line with a default comment.  

image

I can't change the comment, so please don't have that.  
And the extra line `plrv <- ()`  makes running the current script more difficult.  Does it do any good?

g) I understand why you got rid of Ok, but I now keep pressing Reset, because it is the one on the left and I'm used to pressing that.  I suggest you re-instat the Ok key.  (It does the same as To-Script, so I'm happy if you have just the 4 keys, but then move `To Script` to the left.  Or have the 5 keys and have `To Script` as the first and the last?)  I would like to be able to press the first (left-hand side) key!

Patowhiz:
@rdstern thanks for the review.
Can you countercheck that item (b) is really in the master branch, I'm not seeing that feature. This PR has not changed that addition.

I don't understand item (e), you mean when you type on the preview, any additions are ignore?

I've made changes that address the other points.

rdstern:
@Patowhiz I show the picture below:

image

I saved the dataframe and that's line 17 in the script added.
But it has also added that curious line 16, and also 15:  namely:

```
# Dialog: Script
ssp <- () 
```
Do we need them?  If we are going to add a comment, then I'd like the "usual" cntrol for comments in the dialog.  Then I can untick and not have it - or tailor it, as we can do in all the other dialogs.

And the ssp starting empty.  I am sure that will be ok, once Stephen finishes his improvements, but is it needed?

Second is that you ignored my point g above.  

image

I don't mind that there are only 4 buttons here.  I do mind that the left hand one is now Reset, because that's what I keep pressing - and that's annoying, because I then have to start again!

I had this problem immediately, in this testing.  Namely I just click on the left hand button at the bottom - which is always the Ok button - except here.  I would really like a left-hand button here to click on.  Could you move the To Script button to the left, or re-instate the 5th button - and maybe call it To Script.  (I don't even mind if we keep it as Ok here, because going to the script is what Ok does here.  I do mind that the left button is Reset, because that keeps messing me up!)

My point e) is shown below:

image

In some ways it is nice and consise that there is no new line after `###Name AMMI` ?  But perhaps it would be clearer if it were essentially the same look as when imported?

Patowhiz:
@rdstern thanks for the review.

I have now returned the **ok** button. You raised very valid points that I didn't see before (user consistency in being used to something).

I'm not getting the comment on my side. I'm surprised how you are able to get it since I've explicitly disabled it.

Line `ssp <- ()` is being added to show what or where the ssp parameter is coming from. Do you still want me to remove it?

I don't understand what you mean by same look as imported.
> In some ways it is nice and consise that there is no new line after ###Name AMMI ? But perhaps it would be clearer if it were essentially the same look as when imported?

Note if a user wanted the insert dialog to insert with a new line before the script, they can enter the new line in the script window before opening the dialog or the preview text in the dialog. I prefer letting this decision be made by the user using this dialog.

rdstern (approved):
This is looking nice now.
a) The extra comment line has gone now.
b) @lloyddewit That extra line below `ssp <- ( )` is maybe better code, but I currently comment it out, so your "system" runs easily.  I assume this will be ok before too long?  Can you confirm.
c) I am happy the Ok is now back again.
d) The renaming of a tab (one of the new features before) doesn't suggest that name when you save.  That's not enough to stop an approval.
e) Also, for the future, I would like just a few more commands in the Commands tab.  I have 2 in mind initially.  First is to put the data from a defined data frame into a pivot table, and the second is to define variables as factors. 

I suggest @N-thony could check and merge.  He is away for a few days, but returns next week.  



```
f <- system.file("external/ssp.csv", package="agricolae")
ssp<-read.csv(f)
#ssp <- ()
data_book$import_data(data_tables=list(ssp=ssp))
```

Patowhiz:
@rdstern the extra line added is not because of @lloyddewit . I've deliberately implemented it that way. I have explained why I added it above. And also asked you in light of my explanations if you still find it irrelevant.

Patowhiz:
@rdstern thanks for more suggestions in regards to this. I'll recommended less and less standard R commands in this dialog. This is a dialog that I think is for users ready to transition to using R through code. I'd lean towards pitching it as a dialog that helps R users learn the databook "api" as a package. This will enable them use R-Instat in a powerful way than R Studio, considering they won't have to write boilerplate code in regards to managing multiple data frames, data outputs, metadata etc. Which makes me wonder whether we are even naming this dialog correctly. 

I'd be interested in understanding the criteria you are leaning on of choosing the standard R functions/scripts to include or exclude. Once I understand that criteria then I'd lean towards another dialog that is suited for such purpose.

lloyddewit:
@rdstern
great to see the Script window being used and tested.
I'm still a bit confused about `ssp <- ()`. This is not valid R code. RStudio confirms this:

image

I assume this line is just a place holder and is not intended to be run without being updated first. Is that correct?
In that case, as far as I can see, R-Instat is behaving correctly by raising an error. What is is that you would like me to fix?
Thanks

Patowhiz:
@lloyddewit you are obsoletely correct. @rdstern line `ssp <- ()` is a dynamically added placeholder. I explained the intention above and asked you whether you still think we should remove it. Refer above to my explanations as to why I added it.

rdstern (changes requested):
Thanks @lloyddewit and @Patowhiz - let's remove it.

lloyddewit:
> Thanks @lloyddewit and @Patowhiz - let's remove it.

Or could we replace it with a comment that tells the user that they need to add soemting at this point in the script?

Patowhiz:
@rdstern as discussed online, I have made changes that add comment descriptions for the scripts produced (excluding Get examples). 

We have also agreed that item (b) is also not in the master branch. So can be done in a separate PR.

rdstern:
@Patowhiz you seem to have gone overboard on the comments now!  When I add a file it is now all comments!

image

You also seem to have changed the get_data_frame command.  It doesn't work now.

image

Patowhiz:
@rdstern not really.

Kindly note, as stated in my above [comment](https://github.com/africanmathsinitiative/R-Instat/pull/8629#issuecomment-1806267549) I didn't add any comments for the **Get Example** tab. I presume those examples have enough comments already. Any comments you get are being retrieved from the package. I have also maintained the initial implementation that is now in the master in regards to retrieving the examples from the packages.

Comments have only been added for **Save Data**, **Get Data** and **Commands** tabs.

For `data_book$get_data_frame()` function to work, the passed name should match a data frame that exists in the data book. Notice the error displayed in your above error message box.

rdstern:
@Patowhiz I tried the same example with the merged version and that is fine:

image

And when I load the example of alwan lamb I (correctly) get the following with the merged version:

image

Note your ##D from line 10 in your version above that is absent here.

Patowhiz:
@rdstern thanks for raising the double ##D. I have now resolved the issue.

rdstern:
@Patowhiz thanks for the ##D stuff.

I am still not able to save the data frame, i.e. transfer it to R-Instat.  I assume that's save data?  So what do I do to put ssp into a new R-Instat databook?  It used to work.

image

Patowhiz:
Damn! Yes, that's a wrong command. I'll fix this.

Patowhiz:
@rdstern I have now fixed the commands

rdstern (approved):
@Patowhiz great.  I am approving.

However, I found a problem that it would be nice if the Save > Data Frame could cope with.

image

The new feature we don't yet allow for, is when the data is not a single data frame.  We now cope with this in the Import from library.  Could we have an extended version of the command to add to the script, when we need it in the code?

I used a new package, called gosset - I'll explain the name when you visit!

This had the following line:

```
data("kenyachoice", package = "gosset")
# Save data frame "kenyachoice"

data_book$import_data(data_tables=list(kenyachoice=kenyachoice))
data_book$import_data(data_tables=lapply(X=kenyachoice, FUN=data.frame))
```
Your function is the first import_data and doesn't work when there are multiple dataframe.  When I copy the last line from the Import From Library it imports the 2 data frames fine!

I am approving, but wonder if this extra option could be added?  It could be a new pull request I suppose?

In the figure above I wonder if there could be 2 radio buttons.  The default is as shown, then there is a second which says Multiple Data Frames. (It is a nice simple example of lapply!)

rdstern:
@Patowhiz I now suggest the example above could be done before merging.  It is our nice first example of adding a loop.
Then, when you are in England I would like to discuss, also with @lilyclements how we extend this feature to add one way we facilitate adding loops in R-Instat.

I refer here to the pull requests for reference:

a) Multiple (Excel) files.  This is issue #8521.  I now suggest that the added line that is needed could become a third option on the existing Save Data tab with the Data Frame option!
b) The loops referred to in the reviews are:
1) Doing (manipulation) things for multiple columns when the dialog only works for a single column.  (So we need to select multiple columns.  We have a `select` dialog!)
2) Doing (manipulation) things for multiple levels of a factor when the simple manipulation is for a single level.  (We have a dialog to choose selected levels of a factor and make a set of `filters` from them.
3) Doing multiple analyses.  This could either imply 1 or 2 above.

I suggest we add a Loop tab to the dialog.  This has Factor Levels and Variables buttons.  It then calls on either the Filter control, or the Select control to provide the list of factor levels or variables.

Then I don't know.  My R isn't good enough.  Do we use lapply, or `purrr `or `foreach`.  All are installed in R-Instat already.  This stuff is referred to in issues #4386, #4228, #3928 and #4445.  

Very exciting.

Patowhiz:
@N-thony, would you be willing to include that particular line in the script window through a separate PR? Additionally, could you verify whether the line is currently absent from the master branch?

rdstern:
@Patowhiz I like the change.  I tested with the gosset package (which I had to install and which is becoming important for us) and the function ptpermute.  It uses the kenyachoice dataset, which is interesting in itself, and which has 2 data frames.  Wgen I untick single, it gives the correct command to import the 2 data frames into the R-Instat data book.

@N-thony can you check again.  It would be great to get this loaded.
Could you also add the code to suggest the renamed text, when a tab is saved?

rdstern (approved):

PR #8633, opened 2023-11-06
Add version number back in

ChrisMarsh82:
Added version number to main form


rdstern (approved):
Great.  I assume it will pick up the proper number when the version is prepared.  It is so tempting to think we are already at 1.0.0!

rdstern:
@Patowhiz or @ChrisMarsh82 could this be merged?

I keep using both the latest and the development version together - It would be nice that one of them now has a number, so I can tell which is which!

Issue #8546, opened 2023-09-20
Add emmeans dialog to R-Instat


rdstern:
This was sent by @fran2or and Issouf as an e-mail suggestion.  The teaxt was:

"Francis and I have put together a menu proposal for emmeans in R-instat. We need your opinion on this to improve it. "

<img width="560" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/69d33c0f-8b45-433a-a63a-63c86367c483">

I am copying it here so we can keep track of the suggestion.  Issouf what is your name on the github site?

IssoufKon:
**Here are our responses to the different comments**

a) we agree

b) Done

c) Done. We add a file with some sample outputs.

d) We've added a little box to click on before adding letter displays.

e) We add a file containing the emmeans

**Attached files**
[Model keyboards.odt](https://github.com/africanmathsinitiative/R-Instat/files/12859532/Model.keyboards.odt)
[New emmeans menu.odt](https://github.com/africanmathsinitiative/R-Instat/files/12859534/New.emmeans.menu.odt)
[Some emmeans output.odt](https://github.com/africanmathsinitiative/R-Instat/files/12859621/Some.emmeans.output.odt)

rdstern:
@IssoufKon that's great.  I'm sorry I missed it when you replied.  I did get your suggestions as an e-mail as well, but missed it while teaching.

I'm keen to start by building on this options, you gave:

<img width="494" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/910b75df-dcd1-4310-ba43-e7e6327401b7">

You have added emmeans to the general keyboard.  I suggest it should have its own keyboard.  This could usefully be the second keyboard after General.

That's because I suggest quite a lot more keys.

emmeans starts with a model and produces results - some that you have used.  It also produces an emmGrid object.

There are then some functions that start with this object, namely summary, confint, test, contrast, pairs, xTable 

I would separate and have perhhps the top row of the grid, with the functions in the emmeans package that can start with the model.  And the first is emmeans itself.

emmeans, trends, joint, refgrid  - need to investigate the extra keys perhaps?
Then there might be a small gap and perhaps 2 further rows of keys in alphabetical order:

confint, contrast, intplot, pairs  (intplot is emmip function)
plot, pwpp, summary, test, xTable

[...]

PR #8639, opened 2023-11-09
Added new Emmeans keyboard

derekagorhom:
fixes partly #8546 
This PR is ready for review. 

The emmeans keyboard can be found in Model > use Model > use Model Keyboard
the keys added are;
emmeans,  trends, joint, refgrid
confint, contrast, intplot, pairs
plot, pwpp, summary, test, xTable in that particular order. 
dataset to be used for this keyboard must be `.lm`, `.emm`, `.glm`
@rdstern would it be ok to add tooltips to this keyboard? 
@lilyclements can you review this
Thanks

### DIFF
--- a/instat/dlgUseModel.Designer.vb
+++ b/instat/dlgUseModel.Designer.vb
@@ -76,19 +76,22 @@ Partial Class dlgUseModel
         Me.ContextMenuStripSegmented = New System.Windows.Forms.ContextMenuStrip(Me.components)
         Me.ToolStripMenuSegmented = New System.Windows.Forms.ToolStripMenuItem()
         Me.grpEmmeans = New System.Windows.Forms.GroupBox()
-        Me.Button1 = New System.Windows.Forms.Button()
-        Me.Button15 = New System.Windows.Forms.Button()
-        Me.Button14 = New System.Windows.Forms.Button()
-        Me.Button13 = New System.Windows.Forms.Button()
-        Me.Button4 = New System.Windows.Forms.Button()
-        Me.Button5 = New System.Windows.Forms.Button()
-        Me.Button6 = New System.Windows.Forms.Button()
-        Me.Button7 = New System.Windows.Forms.Button()
-        Me.Button8 = New System.Windows.Forms.Button()
-        Me.Button9 = New System.Windows.Forms.Button()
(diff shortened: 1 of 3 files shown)


rdstern:
@derekagorhom that's an excellent start.  It works already.  It raises a new point that is a question for Patrick

image

@Patowhiz Here is the dialog for a model for the survey data.
This is the Use Model dialog, but this command needs to use both the models and the variables that make up the model.  How should we do this in the current data selector?  I assume we could simply list them both, but it would be nice if there could be a distinction after the models.

They will be from the same data frame, so could be in the single selector.

@derekagorhom one limitation here is there is the Try control, but it doesn't work.  Could it be implemented, as it is for the Fit Model keyboard I think.

Then another slightly annoying feature of this dialog is that the variable names someimes have to be in quotes  
, but not when they use the curly line.  So I suggest you add a few keyboard keys, including the curly line and the double quotes.

Then sabi should help with the testing, but I suggest it is mainly Issouf.  He says he is on github under issoufkon, but I can't find him here.  Is sabi on yet?

I like your idea of tooltips.  Perhaps Issouf can help here too?

Patowhiz:
@rdstern you want the selector to list the variables that make up the model only?

rdstern:
@Patowhiz that would be perfect.  I assumed that might not be easy, or even possible?  The selector would list all the models available.  Then you would choose one of them.  That would be the point at which you could presumably include those variables?

Patowhiz:
We currently don't have that support at the R level. We will have to investigate how we could do that. It's very possible to implement.

You now raise a very important question, whether we should store such kind of metadata for our output objects. I'm also not sure if by default all packages store this as part of the object metadata and we only need to read from the objects.

rdstern:
@Patowhiz many thanks for that.  Meantime I hope it might be easy for @derekagorhom to add the display of the variables, plus the models, into the data selector.  Then add a double-quote key to the keyboard.  And possibly get the Try working.  

Then - even now it is usable.  And Issouf had a 700-line code that used it!  So could he choose some examples of data - maybe real examples from his work, if the data can be shared - or perhaps (also) from agridat, and try it out.

rdstern (approved):
@derekagorhom well done.  There is more to do on the data selector, but I would like to see this merged now, and further improvements can then be added.

PR #8642, opened 2023-11-10
Fixed Small Issue On General Graphics Dialog

### LINKED ISSUE #8640: Small correction needed in the general graphics dialog
@MeSophie these should be simple to do

image

a) When I ask for geom_rug it still gives me rug_smooth!
b) And geom_smooth doesn't give me the geom!

MeSophie:
Fixed #8640 
@rdstern  I fixed the issue on Geom_rug and Geom_smooth. Please have a look.


rdstern:
@MeSophie great, this now all works.  But it has resulted in a new small oddity that could not be noticed before.  I hope it will be equally easy to fix?

image

There are 2 layers here.  But I also added 2 layers from the Plot sub-dialog:  Here:

image

It would be nice if they could also appear on the main General Plot dialog?

rdstern:
@MeSophie and @N-thony I have been using this dialog now in a new presentation.  It is brilliant, but I have found another few oddities in the R-code.  I suspect an @N-thony touch may be an efficient way to help.  I'll later add an example you can run through.  What I found:

a) Clicking on the Legend Position and setting it to None works fine for a time.  But if you go into another sub-dialog, e.g. Facets, and then change something, the Legend Position on the dialog (set to None, is still ticked, but it doesn't do anything.
b) On the Facets I ticked Free Scales Y, and it works fine.  Then I unticked and that's fine too.  Excel it is still ticked the next time.  It doesn't act as though ticked, so no harm, but once unticked it should stay unticked.
c) I then had a long look at the code that is generated and find it to be confusing.  Here it is on my example:

```
last_graph <- ggplot2::ggplot(data=datasaurus_dozen, mapping=ggplot2::aes(x=x, y=y, fill=dataset, colour=dataset)) 
+ ggplot2::geom_point() 
+ ggplot2::geom_smooth(method="lm") 
+ theme_grey() 
+ ggplot2::theme(plot.caption=ggplot2::element_text(), plot.title=ggplot2::element_text(size=20), plot.subtitle=ggplot2::element_text(), axis.text.x=ggplot2::element_text(), axis.title.x=ggplot2::element_text(), axis.text.y=ggplot2::element_text(), axis.title.y=ggplot2::element_text(), legend.position="none") 
+ ggplot2::facet_wrap(facets= ~ dataset, dir="h")
```
It all starts nicely.  Then it has an incredibly complicated section that I don't understand and I hope is not needed.  And I note that the legend.position is all in that complicated section.

I thi nk there is some unravelling to be done here.  I hope that long part of the command is not needed in that way.  I didn't touch the titles stuff, so hope it can be omitted from the command generated.

rdstern (dismissed):
@MeSophie that's good progress on the dialog.  The layers now works fine.  I hope you can also attack the new bits I found.
I looked at the code with, and without, the setting of the Legend to null.  The messy code comes because of that and has always been there I think!  At least it is the same mess when I set it from within themes as from your front menu.

So that, and also the anomaly with the facets is more general than your dialog.  Happy if you want to investigate, but we could also merge now, and make a new issue of that?  So I am approving for now

MeSophie:
> I looked at the code with, and without, the setting of the Legend to null. The messy code comes because of that and has always been there I think! At least it is the same mess when I set it from within themes as from your front menu.
> 
> So that, and also the anomaly with the facets is more general than your dialog. Happy if you want to investigate, but we could also merge now, and make a new issue of that? So I am approving for now

@rdstern  I'm still working on it when ready I will let you know.

rdstern (commented):
@MeSophie great - very well done I think.
One small change, you made the label Plot, rather than Options.  Could you change the first button, in English, to Options.

MeSophie:
> @MeSophie great - very well done I think. One small change, you made the label Plot, rather than Options. Could you change the first button, in English, to Options.

@rdstern I changed the text of the control from Plot to Options. Please have a look.

rdstern (approved):
@MeSophie many thanks - great.  @N-thony over to you!

PR #8643, opened 2023-11-10
Fixing a bug found in Climatic > Dates > Fill Date gaps dialog

### LINKED ISSUE #8603: Bug found in Climatic > Dates > Fill Date gaps...
During a recent R-instat presentation made by @fran2or along with John we were doing a data clean up when we found this bug. 

we tried to fill the date gaps for the data we were cleaning see below: 
Screenshot (600)

we filled the receivers and got this error
Screenshot (602)

Vitalis95:
Fixes #8603 
@rdstern ,  can you test from your side if the dialog stills works fine after commenting that line of code  suggested by @lilyclements  in the issue


rdstern:
When 2 or more factors were used there, @Vitalis95 that line was checking that they were essentially the same factor.  (So combining them didn't add any more levels.)  I still suggest that the error that started all this not an error, but was correctly detecting that what was done was not sensible.  I hope I answered that in the issue.

Maybe I have misunderstood what was done here.

lilyclements:
@rdstern the issue that was occurring here was triggered by having two factors in the receiver

1. Import rwanda data
2. Stack the elements and create a new column "element"
3. Remove some lines from the new data (e.g., lines 7-9)
4. Run infill dates for the stacked data and set "elements" and "station" in the receiver

That would trigger an error since the factor levels weren't the same in element and station. Is that what we expected to happen? If so, I'm not sure why.

> that line was checking that they were essentially the same factor.
That makes sense, I didn't know that's what the line was for. We should run this check in another way. But then what if someone puts in three variables where two are essentially the same factor? We can have checks, but it will get large. I assume we're happy with that?

lilyclements:
@Vitalis95 I think I have fixed this, given comments by @rdstern. Can you check? Thanks

Vitalis95:
@lilyclements ,thanks, it is fine, @rdstern  are you happy with that?

PR #8644, opened 2023-11-10
In Script window, implemented ?? operator and ensured tildas correctly parsed

### LINKED ISSUE #8533: More sample scripts and the third has piping which may work now?
@lloyddewit I am taking advantage of your last week, to raise a point I don't see yet in your list of script tasks!

The [site is here](https://waterdata.usgs.gov/blog/beyond-basic-plotting/): 

image

It is useful as being climatic.  And it is a sort of training exercise that uses ggplot in the ordinary way.  So useful to see if the code can run.  It startsd with code that needs a new package to be loaded.  I did that.

Here is the code:

```
library(dataRetrieval)
library(dplyr) # for `rename` & `select`
library(tidyr) # for `gather`
library(ggplot2)

# Get the data by giving site numbers and parameter codes
# 00060 = stream flow, 00530 = total suspended solids, 00631 = concentration of inorganic nitrogen
wi_daily_wq <- readNWISdv(siteNumbers = c("05430175", "05427880", "05427927"),
                          parameterCd = c("00060", "00530", "00631"),
                          startDate = "2017-08-01", endDate = "2017-08-31")

# Clean up data to have human-readable names + move data into long format
wi_daily_wq <- renameNWISColumns(wi_daily_wq,
                                 p00530 = "TSS",
                                 p00631 = "InorganicN") %>%
  select(-ends_with("_cd")) %>%
  gather(key = "parameter", value = "value", -site_no, -Date)

# Setup plot without facets
p <- ggplot(data = wi_daily_wq, aes(x = Date, y = value)) +
  geom_point(aes(color = site_no)) +
  theme_bw()

# Now, we can look at the plot and see how it looks before we facet
# Obviously, the scales are off because we are plotting flow with concentrations
p

# Add vertical facets, aka divide the plot up vertically since they share an x axis
p + facet_grid(parameter ~ .)

# Add vertical facets, but scale only the y axes freely
p + facet_grid(parameter ~ ., scales = "free_y")

p + facet_grid(parameter ~ ., scales = "free_y",
               switch = "y", # flip the facet labels along the y axis from the right side to the left
               labeller = as_labeller( # redefine the text that shows up for the facets
                 c(Flow = "Flow, cfs", InorganicN = "Inorganic N, mg/L", TSS = "TSS, mg/L"))) +
  ylab(NULL) + # remove the word "values"
  theme(strip.background = element_blank(), # remove the background
        strip.placement = "outside") # put labels to the left of the axis text
        
```
It runs.  It includes lots of comments.  They are like the comments we include in the dialogs.  But they do not appear in the output window.  Please could they?  Note that some comments are at the end of command lines.  If commands are on, then those commands are put into the output window.  

I assume it may be difficult to put them into the output window, without the commands?  That would be a bit disappointing, but I would be happy with that for now.

Then, on the extremely positive side, it all runs fine.   As mentioned recently it puts the (graphical) results into the R viewer.
It is also useful, because it mis retrieving hydrological data from the USGS (US Geological Service).  If people can use scripts then this might be all we provide initially, i.e. some examples.

The next section is essentially tidying the data - neat code, and perhaps we will sometimes continue to have that in code.
Then, just before the line `# Setup plot without facets` we can read the data into an R-Instat data book, and that works fine.

Then I hope we can do the same graphs from our dialogs.  If not, then we might, for now, have to "top" and "tail" each of the graphs with the code needed to get a data frame, and then to add the plot into our "system".

I see work already specified here.  The addition of comments to the output window is the only @lloyddewit query - perhaps to implement, or at least to discuss and add to the list of remaining tasks.

lloyddewit:
@rdstern 
For details of these changes, please see:

- lloyddewit/rscript#16
- lloyddewit/rscript#32

I think this PR also fixes #8533. I think that the only remaining parts of this issue are covered by lloyddewit/rscript#19. Please check.

Apart from keywords and curly brackets, do you have any other R code that is not correctly parsed by the RScript library?

Thanks

### DIFF
--- a/instat/instat.vbproj
+++ b/instat/instat.vbproj
@@ -155,8 +155,8 @@
     <Reference Include="RDotNet, Version=1.8.2.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\packages\R.NET.1.8.2\lib\netstandard2.0\RDotNet.dll</HintPath>
     </Reference>
-    <Reference Include="RScript, Version=0.1.0.9, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\RScript.1.0.10\lib\net461\RScript.dll</HintPath>
+    <Reference Include="RScript, Version=0.1.0.10, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\RScript.1.0.11\lib\net461\RScript.dll</HintPath>
     </Reference>
     <Reference Include="ScintillaNET, Version=3.6.3.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\packages\jacobslusser.ScintillaNET.3.6.3\lib\net40\ScintillaNET.dll</HintPath>
--- a/instat/packages.config
+++ b/instat/packages.config
@@ -11,7 +11,7 @@
   <package id="Microsoft.Win32.Registry" version="4.5.0" targetFramework="net461" />
   <package id="NLog" version="5.1.0" targetFramework="net461" />
   <package id="R.NET" version="1.8.2" targetFramework="net461" />
-  <package id="RScript" version="1.0.10" targetFramework="net461" />
+  <package id="RScript" version="1.0.11" targetFramework="net461" />
   <package id="Stub.System.Data.SQLite.Core.NetFramework" version="1.0.113.3" targetFramework="net461" />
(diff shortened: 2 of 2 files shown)


rdstern (approved):
@lloyddewit it is looking so good now.  Everything I tried now seems to run, except - as you know - poor Florence Nightinggale - and in Crimea too!
@Patowhiz if you are ok with the code, this is another step forward!

Patowhiz (approved):

PR #8648, opened 2023-11-13
Improved column selection to apply or not to dataframe, metadata or dialogue

### LINKED ISSUE #8600: Improving the column selection feature
@N-thony this is great - it now seems to work.  Here it is even for the extremes data:

image

I checked with @lilyclements though who is also excited by the option of being able to apply the select (or not) to these 3 alternatives (dataframe, metadata, dialog).  So I hope that isn't too  demanding a task.  

I had not expected this for the November upgrade even, so am really delighted!

_Originally posted by @rdstern in https://github.com/africanmathsinitiative/R-Instat/pull/8561#pullrequestreview-1655493037_

### LINKED ISSUE #8650: Be able to select partially labelled columns?
@N-thony your new select is looking very good now.  There is an option (in the `where` option) to select labelled variables. 

I would very much like a further option to select all just partially labelled columns.  Shadrack said that would be easy to add.  I checked and  can't find the answer for R in stackoverflow.  The nearest I got was that it is easy in Stata!

image

If it also isn't easy for you, then I wonder if @lilyclements or @Patowhiz can help.  Otherwise I could try Shadrack again.  Antoine I am sending the household data that provides an example.

N-thony:
Fixes #8600 
Fixes #8650
@rdstern have a look.
The dialogue checkbox is disabled and I was thinking that it will be good if the selection is applied to a specific dialogue?

### DIFF
--- a/instat/dlgSelect.vb
+++ b/instat/dlgSelect.vb
@@ -20,6 +20,7 @@ Public Class dlgSelect
     Private bReset As Boolean = True
     Private clsSetCurrentColumnSelection As New RFunction
     Private clsApplyAsSubset As New RFunction
+    Private clsDummyFunction As New RFunction
 
     Private Sub dlgSelect_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
@@ -52,6 +53,12 @@ Public Class dlgSelect
         ucrReceiverSelect.SetParameter(New RParameter("name", 1))
         ucrReceiverSelect.SetParameterIsString()
 
+        ucrPnlOptions.SetParameter(New RParameter("check", 0))
+        ucrPnlOptions.AddRadioButton(rdoApplyAll, "all")
+        ucrPnlOptions.AddRadioButton(rdoApplyToDataFrame, "dataframe")
+        ucrPnlOptions.AddRadioButton(rdoApplyToDialogue, "dialogue")
+        ucrPnlOptions.AddRadioButton(rdoApplyToMetaData, "metadata")
+
         ucrPnlApplyOptions.AddRadioButton(rdoApplyAsSelect)
         ucrPnlApplyOptions.AddRadioButton(rdoApplyAsSubset)
         ucrPnlApplyOptions.AddFunctionNamesCondition(rdoApplyAsSelect, frmMain.clsRLink.strInstatDataObject & "$set_current_column_selection")
@@ -68,7 +75,12 @@ Public Class dlgSelect
     Private Sub SetDefaults()
         clsSetCurrentColumnSelection = New RFunction
         clsApplyAsSubset = New RFunction
+        clsDummyFunction = New RFunction
         ucrSelectorForSelectColumns.Reset()
+
+        grpOptions.Visible = False
+
+        clsDummyFunction.AddParameter("check", "all", iPosition:=0)
 
         clsSetCurrentColumnSelection.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$set_current_column_selection")
 
@@ -84,6 +96,7 @@ Public Class dlgSelect
         ucrReceiverSelect.SetRCode(clsSetCurrentColumnSelection, bReset)
         ucrInputNewDataFrameName.SetRCode(clsApplyAsSubset, bReset)
         ucrPnlApplyOptions.SetRCode(ucrBase.clsRsyntax.clsBaseFunction, bReset)
+        ucrPnlOptions.SetRCode(clsDummyFunction, bReset)
     End Sub
 
     Private Sub TestOkEnabled()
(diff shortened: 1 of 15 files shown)


rdstern (commented):
@N-thony here is the current dialog:
image

I suggest 3 radio buttons (rather than 2) on the left.

O Select
O Subset
O Save Select

Select is the default and is the only one that shows the right-hand group.  Theat has 3 checkboxes.  Perhaps a label Apply To:
Then Data, Metadata, Dialogs.

All are checked by default.  

Roger.

rdstern (commented):
@N-thony this is starting to look very good.
a) I was able to save 2 selects and then use them later.  This seems to work fine.
b) But when I asked for it to apply to just the data (not to the metadata) it still applied to the metadata.
c) When I applied to the metadata and not to the data, I got a formatting curiosity, see below, for the many decimals in the size column.  When I remove the selection the display is back to normal.

image

d) And a new request.  I hope it is easy, so you can do it here.  Otherwise can you make a new issue.  When I make a filter, I can see the function on the main dialog.  That's very useful.  Could you add that feature for the selects?

e) I note that the select is currently always applied to the dialogs.  That option is greyed out.  I suggest that's fine for now, but could you tick it, so it is clear it will be applied to the dialogs.  And I can't apply it to just the dialogs?  Is that option difficult to add, see below:  If so, then stick with what we have now.  I am really happy that you can define selects ready for later use.

image

rdstern (changes requested):
@N-thony I'm not sure about this, but happy if you want to continue, then fine. I used the usual survey dataset:
However The only cases I would really like are:
a) To be able to save the select and apply later.  That's working.
b) To be able to not apply in the metadata - which we used to have.  So apply always in the data and the dialogs, and just have a checkbox whether you apply to the metadata.  Currently that seems not to be working, i.e. it always now applies in the metadata.  When not applied (as before) the red would indicate which were in the select.

I would be happy if you continue to offer the other options, but there seems an error with each.
c) When not applied to the data, the formatting of at least one column in the data is odd.
d) When not applied in the dialog it gives an error if you choose to use a variable that is not in the select.

rdstern:
@N-thony my first try was a selection now applying to the metadata.
a) Previously I thought it was able to mark them red, though you saw all of them.  Would that still be possible?
b) However then I did a one variable summary, and then the metadata columns were selected, i.e. only the selected variables were shown now.
c) I am still keen on being able to see the selection formula in the main dialog, just as you can in the filter.  Can you include that now, or would you prefer to add it to the issue?

rdstern (approved):
@N-thony my mistakes.  Thanks for the clarification.  It seems to work fine

rdstern (commented):
I am just testing the variables that are lebelled here and used hh from MICS.  The idea seems ok.
But when I ask for labelled variables I get 257 and these are fully or partially labelled.  When I ask for just the partially labelled I get 279 variables!  There's something wrong here.  I should get less.
I checked also by adding the not.  There are just 50 not labelled and that is working.

So you may be close!

rdstern:
@N-thony that's close!  I get 72 partially labelled, but they seem to be (instead) the variables that are not labelled.  When I ask for labelled variables and add the Not I get 72 - so the same.

rdstern (dismissed):
@N-thony very well done.  @Patowhiz can you check and merge?

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -469,11 +469,11 @@ DataSheet$set("public", "get_variables_metadata", function(data_type = "all", co
     if(missing(column)) {
       curr_data <- private$data
       cols <- names(curr_data)
-      if(self$column_selection_applied()) cols <- self$current_column_selection
+      if(self$column_selection_applied() && use_column_selection) cols <- self$current_column_selection
     }
```
Patowhiz, line 473: @N-thony why do we need an extra `use_column_selection` parameter instead of using the global `column_selection_applied()` only function. Is there a scenario where we will have a column selection yet `column_selection_applied()` will return false?

instat/dlgSelect.vb
```diff
@@ -52,10 +54,24 @@ Public Class dlgSelect
         ucrReceiverSelect.SetParameter(New RParameter("name", 1))
         ucrReceiverSelect.SetParameterIsString()
 
+        ucrChkDataframe.SetText("DataFrame")
```
Patowhiz, line 57: @N-thony did you intend to have this as "Data Frame" instead of "DataFrame" ? Same applies to "MetaData" instead of "Metadata".

Patowhiz (commented):
@N-thony thanks for this. Just a few comments and question.

I noticed that you have introduced checking of column selections in different areas in the front end. Not unless I have misunderstood, this seems to suggest that the feature will be managed at the front end. Is the column selection not supposed to be managed at the R level? I  expected individual controls to internally check with the R data book as to whether there is column selection or not or send commands with appropriate selection parameters.
In summary I expected any new column selection global functionalities to be managed at the R level not VB.Net. Then for components that wish to not use them, they simply call R functions with correct flag parameters or functions that ignore the selection internally.

Patowhiz (commented):
@N-thony thanks for your response. Very useful. Just a few comments and suggestions then we can merged.

Could you add TODO comments in specific instances that you think could be improved in future in regards to managing the entire column selection at the R level. A reference to this PR in the specific instances will be helpful as well. I find this helpful when in future we gradually refactor code and delete bits that are no longer needed.

Thanks.

rdstern (approved):
@Patowhiz still seems fine to me!

PR #8649, opened 2023-11-13
Returning the focus to the original receiver

### LINKED ISSUE #8256: Returning the focus to the original location?
@Patowhiz I hope this mightbe a simple change that is automatically applied to all dialogues.  That is, if we agree.  Here is the boxplot dialogue completed:
image

When you go in the first time, the focus - so coloured - is on the Y variable.  Filling it in usually puts the focus as shown above - so in a place rarely used.

When I return it is usually either to change what is plotted - so the original variable with the focus.  Or it is to change some of the options, in which case I don't care where the focus is.

With beginners - one of my pre-occupations now - I suggest it would be simpler if the focus was always on the original first receiver, every time the dialogue is re-opened.

If agreed, then could this be a "rule" and programmed just in one place - or do we have to tweak each dialogue?

N-thony:
Fixes #8256
@rdstern I have tested this with some dialogues in the Describe menu, have a look

### DIFF
--- a/instat/ucrSelector.vb
+++ b/instat/ucrSelector.vb
@@ -68,7 +68,9 @@ Public Class ucrSelector
             Dim lstVisibleReceivers As List(Of ucrReceiver)
             lstVisibleReceivers = lstOrderedReceivers.Where(Function(ctrl) ctrl.Visible).ToList()
             lstVisibleReceivers = lstVisibleReceivers.OrderBy(Function(ucr) ucr.TabIndex).ToList()
-            SetCurrentReceiver(lstVisibleReceivers(0)) 'set the focus to the first Receiver in the dialogue.
+            If lstVisibleReceivers.Count > 0 Then
+                SetCurrentReceiver(lstVisibleReceivers(0)) 'set the focus to the first Receiver in the dialogue.
+            End If
         End If
     End Sub
 
--- a/instat/ucrSelector.vb
+++ b/instat/ucrSelector.vb
@@ -65,7 +65,9 @@ Public Class ucrSelector
         LoadList()
         'always return the focus to the first Receiver when re-opening the dialogue.
         If lstOrderedReceivers.Count > 0 Then
-            SetCurrentReceiver(lstOrderedReceivers(0))
+            Dim lstVisibleReceivers = lstOrderedReceivers.Where(Function(ctrl) ctrl.Visible).ToList()
+            lstVisibleReceivers = lstVisibleReceivers.OrderBy(Function(ucr) ucr.TabIndex).ToList()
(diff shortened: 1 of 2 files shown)


rdstern (commented):
@N-thony it seems to work fine on some dialogs.  Others resturn and nothing is in turquoise.  Here is an example:

image

The stem and leaf is another like that:

image

rdstern (approved):
Looks fine.  Perhaps @Patowhiz can confirm

instat/ucrSelector.vb
```diff
@@ -63,6 +63,12 @@ Public Class ucrSelector
         'always load selector contents on load event because contents may have been changed at R level 
         'and the control needs to refresh the data frame names.
         LoadList()
+        'always return the focus to the first Receiver when re-opening the dialogue.
+        If lstOrderedReceivers.Count > 0 Then
+            Dim lstVisibleReceivers = lstOrderedReceivers.Where(Function(ctrl) ctrl.Visible).ToList()
```
Patowhiz, line 68: @N-thony could you specify the type of `lstVisibleReceivers` explicitly. This has many merits relating to long term code maintenance.

Patowhiz (changes requested):

Patowhiz:
@rdstern I hope you have tested this with dialogs that have multiple features. I'm particularly interested with dialogs that have **multiple features** yet use **different receivers**. The sequence of reload may cause regression in the changes introduced here.

instat/ucrSelector.vb
Patowhiz: @N-thony I understand what the predicate is doing in the where clause.
Could you then specify the type explicitly, for example `Dim lstVisibleReceivers As List(Of Control)`. I understand that VB.Net allows dynamic types, but to be consistent with the code base and also for self documenting code, I strongly recommend static typing in this case.

rdstern (dismissed):
@Patowhiz I have checked again.  I really like this improvement and did check with many dialogs.  I can't see any problems.  My nervousness is that you know of some problems and are really just trying to teach me the lesson that I'm not checking carefully enough?
If that's not the case, then please continue with your checks and then merge!

rdstern (approved):

Patowhiz:
@rdstern thanks for your response. I'm glad you were able to check again on the dialogs that I was concerned about. 
I'm still not sure whether this change may affect performance of dialog reloads or not when it comes to wide data sets. 
@N-thony thanks for the change, would be useful to test this with dialogs that have different receivers that are visible in different options/features when using wide data sets. Nevertheless, I'm merging this.
@rdstern keep testing this feature in other PR's as well, I'll do the same.

Patowhiz (approved):

PR #8658, opened 2023-11-23
Added a basic Autocompletion to the script window

N-thony:
@lloyddewit as discussed yesterday, I have implemented here a basic autompletion following the C# example from Jacob. Some keywords are probably missing but have a look.

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -952,6 +952,17 @@ Public Class ucrScript
         End If
 
         RunScript(clsScriptActive.Text, "Code run from Script Window (all text)")
+
+        SetFocusAndScrollCaret()
+    End Sub
+
+    ' Ensure the cursor is visible by scrolling it into view
+    Private Sub SetFocusAndScrollCaret()
+        ' Set focus back to the ScintillaNET editor control
+        clsScriptActive.Focus()
+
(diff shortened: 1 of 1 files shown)


lloyddewit:
@N-thony Thank you, this looks like a good approach.
Is it ready for @rdstern to test?
When @rdstern has tested/approved, then I can do a full peer review.

rdstern:
@N-thony so I had a look, though I am not sure what to expect.

I couldn't get it to do anything when I put a script in and hovered, etc.

So I started typing new lines and it gave me some words I could use.  I chose 1 and it included it.  I chose cut

What's next?

Can it include the names easily of many R commands, from different packages?
Can it include the names of our special R-Instat functions?
Can it suggest allowed parameters for the functions names it chooses?  Or is it just their names?

rdstern:
@N-thony I am still not sure whether more time on autocomplete can be a priority now?  I am looking for "quick-wins" and this feature is more for R code developers than users.  

The same issue, number #  has a hopefully small task, that I would really like and that is to make the cursor visible in a script.  I wonder if this could be included here?

rdstern (commented):
@N-thony I am very happy to see the cursor and would very much like to merge this.  But the autocompletion is now also automatically there and it never gives me any useful suggestions at this stage.  So, could that be disabled for now?

PR #8615, opened 2023-11-03
Scripts and sub dialogue for gt Table

Vitalis95:
Fixes partially #8541 
@rdstern , this is @Fidel365 pull request, I managed to solve the merge conflicts

Issue #8554, opened 2023-09-26
Controls for editing text in Tables in R-Instat


lilyclements:
@rdstern and I have been discussing the tables and come up with a new set of controls for the Titles tab when editing the subtitle.

@N-thony we wanted to know if the following is possible, and, if you had any additional requests or ideas for this?

We want to have a control that is similar to this:

<img width="234" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/21180424/d32574ff-5de7-4ff4-9b90-66dc975c3c54">

But with the options given in the `gt` package in R, so instead we would have something like this:

![image](https://github.com/africanmathsinitiative/R-Instat/assets/21180424/d1513706-5589-4fdd-b477-aab1718b192c)

I've numbered the different bits and go through the questions I have on this for @N-thony 

1. **Font** - In here, we have an editable drop down giving the font to choose. Is there already a control like this in Visual Basic that we can use?
2. **Size** - This is the size of the font, and this can be an editable drop down that allows numerical values only. We could have in it the options that Word has: 8, 9, 10, 11, 12, 14, 16, 18, 20, 22, 24, 26, 28, 36, 48, 72
3. **Bold** - This could be like a button, with an arrow in the corner to select further options (like we have on our OK) to add further bold options in R (e.g., _bolder, lighter_).
If you click the button, it makes it **bold**, if you unclick it, it removes that parameter and parameter.
Is this possible?
6. This is a colour changer to change the colour of your font. Is it possible to have this as an option in R-Instat as a button with an arrow of the different colours?

4-5, 7-9. Are the same idea as 3., but with **Italic**, **Underline**, **Sentence Case**, and **Alignment** There are additional options if you click on the little arrow in the corner that are offered in R. No additional questions here that isn't asked in 3.!

Let me know what you think @N-thony :)

lilyclements:
_Sidenote on the R-Code Side for myself_
These are all in the `gt::tab_style` function. 
1. `font`
2. `size`
3. `weight`, options are: `"normal"`, `"bold"`, `"lighter"`, `"bolder"`. We run `"bold"` if the button is pressed, `"normal"` if not unpressed, and `"bold"`, `"lighter"`, and `"bolder"` are options on the drop down.
4. `style`, options are: `"normal"`, `"italic"`, `"oblique"`. We run `"italic"` if the button is pressed, `"normal"` if not unpressed, and `"italic"` and `"italic"` are options on the drop down.
[...]

PR #8662, opened 2023-11-23
Adds some GT table formatting functionalities

Patowhiz:
Fixes partly issue #8554 #8541
Replaces PR #8615


rdstern:
@Patowhiz delighted that even a new draft is there.  I don't know whether I should be testing yet?  When I try to go in it says:
image

It does then go in.  I sort of expected the Format button to work, because that was where you were were working when in the UK.  It doesn't seem to?  

I also hope that Ana's work on the themes will soon be put back.

Tell me and @lilyclements when you are ready for us to look properly.

Patowhiz:
@rdstern you can now format title, subtitle, foot and source notes. 
Foot notes can have expressions for the each columns. Only valid R expressions will yield results. You can have a look at the [gt documentation](https://gt.rstudio.com/articles/gt.html) of how to add expressions for foot notes. In future we could find a way of making it easy to add the expressions using a GUI instead of users having to type them.

As outlined in the package documentation, 'advanced' text formats are only supported for html output format. That means some text formats will be ignored if you export the table in other formats other than html. This is probably because the package uses `CSS` for these 'advanced' formats 

@rdstern in regards to themes, I simply returned what was already implemented, I didn't delve into refactoring the code nor the design. Test for it's regression only not any additions. 

The new additions are only available through **Describe > Tables...** dialog. Once merged they can be adapted to other tables that need the new features. I have deleted lots of now redundant code in this dialog. The aim was not to refactor it but rather show how gt table dialogs can now use the features added. Test for any regression in this dialog.

Kindly note that the PR doesn't fix all the items in the issues mentioned. We can keep building on top of what is currently here.

rdstern:
@lilyclements this has arrived just as I am boarding for the return to England.  Very exciting, and I note that @Patowhiz seems to have been "burning the midnight oil" to update this pull request.  Please can you check too?

rdstern (commented):
@Patowhiz very nice indeed so far. 
a) I like the idea of the control for the footers.  But how do I use the Column and Row expressions?
b) The Themes > Select Theme doesn't work yet, but it is a trivial fix to do.  All the themes listed come from the gtExtras package.  So you just have to add gtExtras:: in front of that element and they all work fine.  

I'll check on the Manual themes next time.

There's more tabs to come, but @N-thony do you want to check too?  As soon as these parts are working I would like to include this sub-dialog in the other commands and merge.

Patowhiz:
I have fixed the bugs that you found. Kindly retest.

In regards to columns and row expressions, like I said in the above [comments](https://github.com/IDEMSInternational/R-Instat/pull/8662#issuecomment-1955371798), I'm still not sure how best to design a UI for making it easy for users to add the columns and row expressions. As outlined in the same comment, the documentation shows that footers can be used to make references to table cells. You do this by passing R expressions as parameters as shown in the example code below 

```
# Add footnotes (the same text) to two different
# cell; data cells are targeted with `data_cells()`
gt_tbl <- 
  gt_tbl |>
  tab_footnote(
    footnote = "The Americas.",
    locations = cells_body(columns = name, rows = 3:4)
  )
```

The `locations` parameter shown in the above code is done for you by the dialog, So you simply type in the column name or expression (can be a vector etc, check on  this [documentation](https://gt.rstudio.com/reference/cells_body.html)) and row expression as shown above.

The cells_body function also allows us to style those cells. We could add

rdstern:
@Patowhiz I like the new controls and think you have gone a long way - broken the back? - of this task.
With the sirvey data I got this table:
image

Odd that the 1 at the bottom isn't on the same row as the footer?

Here is the current Footer for the table above:

image

The Column and Row fields need a pull-down to help with the options.  My first suggestion (to keep the options and the pull-downs simple is as follows:
a) Make it "shorter" so perhaps just the space for 3 rows, before it scrolls.
b) `Repeat it 3 times! ` For the first, your current Footers Notes becomes a checkbox default checked with label `Single.`
c) The Column pull-down has the names of each column
d) The row pull-down just had the integers, with 1 to the number of rows.
e) The Headers become `Column `and `Row`.  They could be a bit narrower for this option.
f) You choose ust one of each.
g) There is also Title and Sub-title in the row pull down, so you could have a footer for eitrher.
h) There is a Stub option in the Columns so you can link the footer to any cell.  (I think we will need a Stub tab.

i) For the second instance of the control the checkbox has the label `Multiple`.  Default unchecked, and the control is only visible if checked.
j) The header labels are Column(s), Row(s)
k) The column pull-down is again the different column names.  But you can choose multiple columns to include.  Then the code will add the c()
l) The row pull-down is either the numbers again, or some examples.  
m) That's all.  For Multiple it is only for the internal Cells of the table

n) For the third it is called General.  The headers as for Multiple.  The pull-down gives the different functions.  You just choose one for each cell, and then complete that cell.

You may think of something much better.  I am keen that you can have the single option without needing to know coding?

The general themes now works fine.  I haven't checked the more complicated theme elements yet.

lilyclements:
@Patowhiz this is looking great - really exciting stuff!

Patowhiz:
@rdstern I was able to have 2 footer grids; header and cells. 

I managed to implement the header grid, which now has title and subtitle drop down. Note, you should only set the header footer when you supply the title or the subtitle values.

To implement a feature that allowed users to select the table columns, I would have to produce the `gt` table object then read the object to get the column names produced and populate them in a grid drop down. The `gt` package doesn't provide functions for reading the `gt` table object variables. I think their methodology assumes that you are meant to produce the table iteratively through code (meaning you are aware of the output already). Because of the package limitation, I'm not able to implement the feature you have suggested.

Patowhiz:
@rdstern I have also disabled alignment for now because I'm yet to find away of styling the footer superscript label to be consistent with alignment style. The  footer superscript label is added by the gt package which by default styles the parent tag.

rdstern:
This is looking nice, so far.  Here is a table with some footnotes

image

a) Am I limited to numbers in these controls, so which row, or column it is.  There could then be pull-downs?
b) How do I get a footnote attached to the header row.  I assumed row 1, but it seems not?  When I tried 0 it gave an error.
c) I also added atitle and sub-title - and with footnotes too.  Nice.


image


d) The format option, above seems pretty comprehensive.  Where is Bold?
e) The dialog always starts with None everywhere - could it start with the current settings for whatever is being formatted.

f) I look forward to the stub tab - I wonder whether that includes the summary variable, so we could choose to omit that column?

Then there is a lot to look forward to, on the remaining tabs.  For example I need to be able to tailor the decimal places, choose whether NA could be replaced by blanks and so on.  And I am excited about the nano-plots.  You mknetioned in UK that you were aiming for a set of tabs that would be better than those for the graphs.  We have a lot to look forward to.

Patowhiz:
@rdstern

**Bold** is under the weight option.

Would you be okay with me replacing "None" with "Default"? We currently lack a way to directly determine the current settings of the `gt` object intended to be produced.

I define the Header as the title and subtitle only, not the columns. You can see parts of a gt table [here](https://gt.rstudio.com/articles/gt.html).

The header footnote grid is designed to work with the title and subtitle. They are the only options in the grid. As I mentioned earlier, the title or subtitle must exist in the gt object, or you'll encounter an error. To avoid this, could we place this functionality under the header tab?

Regarding pull-downs for columns and rows, during my exploration of the package, most functionalities assume that the user will format the output iteratively. The inputs for the output are expected to come from the `gt` output itself. This differs from our dialogs where most inputs come from the data frame itself. For example, to implement the pull-downs you suggest, I would need to implement the ability to read the HTML output to determine the columns produced. It would have been easier if the `gt` object had such internal functions (which I haven't been able to find).

Would you be open to expanding items **g**, **h**, and **i** in separate issues? You could explain the sample outputs you have in mind, and we can discuss the relevant implementation needed to achieve them.

PR #8665, opened 2023-11-24
Added aov and tidy 4-variable modelling dialogue

### LINKED ISSUE #8645: Adding aov and 4-variable modelling?
@lilyclements there is a lot to do on the modelling and @volloholic is hoping we might spend some time on it, early next year.  But I now suggest 2 maybe very quick wins.  I may check with David tomorrow, but I think they are pretty uncontroversial.  

a) We don't (yet) have an easy way or running aov.  

image

But the 3 variable, 4 variable and general dialog all have this distribution control, shown above.
I suggest we add `normal_aov ` under the `normal` entry here.  We could change that to `normal_lm`, but I suggest we leave it just as it is.
If aov is chosen then it simply changes lm to aov in the function fitted.  I haven't checked carefully if there are new options, but think that might be all we need to do?

b) The 4 variable dialog looks pretty complete.  It says it is for demonstration only.  I have a feeling it might be pretty easy to bring it into operation.  The code behind should be identical to the 3 variable?

These should be things that Sabi could check too.  They are mainly for our agriculture work.

derekagorhom:
Fixes #8645 
@rdstern @lilyclements @N-thony 
This PR is ready for review.
Thanks


rdstern:
@derekagorhom that's nearly there for aov.  You made it more complicated than it should be:

`last_model <- glm(data=survey, formula=yield ~ fert + variety, na.action=na.exclude, family=aov())`

Instead, just put the aov where the glm is now, and not as the family.  So simply:

`last_model <- aov(data=survey, formula=yield ~ fert + variety, na.action=na.exclude)`

On the 4 variable you could enable the Function button on all of them.
And the second default could also be a plus and not *.

derekagorhom:
@rdstern I have made the changes, can you review this again
Thanks

rdstern:
@derekagorhom please also see the comment above, but I am repeating here.
a) In the general it does not need the `family = aov()`.  That gives an error. Also change the `glm` in the command to `aov`.  
b) In the general, when I set to `normal_aov` it does not remember the setting - at least in the general option.  When I return to the dialog it is` normal` again.  (That seems ok in the other options.)

derekagorhom:
@rdstern i have made the changes, can you review
thanks

derekagorhom:
> @derekagorhom are you done making changes?

@N-thony yes please

Issue #8513, opened 2023-08-28
"New" options in `visdat` package


lilyclements:
I say "new" because I'm not sure how new they are. But some nice neat new ways to easily visualise your data frame in the [visdat package](https://github.com/ropensci/visdat) We currently have under 

Prepare > Check Data > Visualise Data

the `vis_dat`, `vis_miss`, and `vis_guess` functions run. There are now more we could add, or at least consider for elsewhere.

**Visualising Correlations**
We can visualise correlations between all variables, or facet by another variable. For example,

`vis_cor(airquality, facet = Month)`
![image](https://github.com/africanmathsinitiative/R-Instat/assets/21180424/11c4550b-a67d-44c1-9b53-ff4b248c2380)

This looks somewhat similar to something to the right hand panel in the `desplot` package that @rdstern has written a little guide on
<img width="381" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/21180424/9d1b3de7-1d1b-449c-a87b-21a35ad07735">

**Compare Data Frames**
Perhaps relevant around duplicating data frames bit, but this is useful to compare two data frames
```
aq_diff <- airquality
aq_diff[1:10, 1:2] <- NA
vis_compare(airquality, aq_diff)
```
![image](https://github.com/africanmathsinitiative/R-Instat/assets/21180424/8950675c-5e4c-422a-a370-ab8098a0adda)

**Expected values**
From their guide - 
"vis_expect visualises certain conditions or values in your data. For example, If you are not sure whether to expect values greater than 25 in your data (airquality), you could write: vis_expect(airquality, ~.x>=25), and you can see if there are times where the values in your data are greater than or equal to 25:
`vis_expect(airquality, ~.x >= 25)`
![image](https://github.com/africanmathsinitiative/R-Instat/assets/21180424/9648a2a9-8357-4178-b2a5-0513a8de8ad9)
"
I'm not sure where this would be most appropriate in R-Instat. But, perhaps @rdstern can see a use. Either way, useful to mention it as another new option in `visdat`. 

**Visualising Numerical Values**
`vis_value()` is a quick way to visualise just the numerical values in a data frame, on a 0-1 scale. This throws an error if there is a non-numerical variable in the data frame, so if we used it we would want to be careful of that. Not sure again how useful this is, but it shows some structure in the data. Running for just numerical values in the Niger data:

`vis_value(Niger %>% dplyr::select(where(is.numeric)))`

[...]

PR #8667, opened 2023-11-24
Added new options in Visualize data dialogue

Raymond-Andam:
Fixes partly #8513 
@rdstern , @lilyclements this PR is ready for review
I have added numeric option and also added the facets for both data and missing options


rdstern:
@Raymond-Andam I had a look and this is a good start.  

a) Numeric option - it seems to work if all the variables are numeric.  But it gives an error otherwise, e.g. for the survey dataset, which is a good simple one for testing.  The R code could select the numeric variables itself, when you use the whole dataframe.  Or don't allow the whole dataframe options, and just include numeric variables in the data selector.
b) The first option now doesn't give a graph - it is blank.
c) The facet option should show the column selector even for the whole dataframe option.  Currently you can't choose a factor.

derekagorhom:
@rdstern this fixes points a and b of the changes you requested,
c could not be done because the checkbox for the `wholedataframe` prevents the user from viewing the available variables in the dataset.

rdstern (commented):
@Raymond-Andam and @derekagorhom on the numeric option please set the selector to only allow numeric (or I presume integer or also perhaps logical variables.  This should be an easy setting.  It is done, for example, in the correlations dialog.

The numeric option doesn't include the possibility of facets.  Is there any reason why not?

On these options I still don't understand why (when you have the whole data frame) you can't show the usual selector, with just factors, when you click on the facets checkbox?  Maybe check with @N-thony?  If you can't, then don't have the facets option with the whole data frame.  If I want facets with all variables I can then always have that option and choose them all.

derekagorhom:
@rdstern we have;
1. Enabled the `numeric option` to allow for only numeric values 
2. allow for the selector to show the variables for the wholedataframe which use the facets. 
3. could not enable facets for the numeric option since the vis_value option does not have a facet parameter. see below
image

rdstern (changes requested):
@derekagorhom and @Raymond-Andam can you allow ordered factors as well as factors here, please?
image

And this looks very odd.  Ideally we should only show the variables when we can select them:  Please can we just show the data frame here.  This is all taking you too long, so perhaps you need some help from @N-thony?

image

On Numeric this also looks very odd don't you think?  You should allow Data Frame.  There don't show the variables, but it picks all the numeric variables.  Then what you have here - with just the numeric variables - is the second option.

image

Raymond-Andam:
@rdstern this fixes the issue of the display so the facet can display the selector when in the whole dataframe options.
We allowed the receiver to take ordered factors.

Antoine is still checking on the whole dataframe options  in the numeric option of the dialog

rdstern (commented):
@Raymond-Andam the dialog now seems to work fine.  However, I now don't get any facets.  I thought I did before.
This makes me wonder how well you tested your changes yourself.  I ask because sometimes developers seem to just change the code, and then hope the results will be ok.  We expect you to do the initial tests yourself, and it is good practice to include a mention of which datasets you used.

rdstern (changes requested):
@Raymond-Andam thank you for reporting.  You said you tried with the diamonds data, so I started with that.  I used visualise and just pressed ok.  Here is the result:
image

So, no graph.  That makes me wonder what you tested?  Remermber you don't just try the facets, which you have changed.  It has to work generally as well.  

Back to you.

derekagorhom:
@rdstern , @Raymond-Andam has pushed the changes that have the fixes for the dialogue not producing graphs... can you please review this.
Thanks

rdstern (commented):
@Raymond-Andam well done - it looks very nice now.  I just found one oddity, that I guess has been there all the time?  When I used the dodoma data from the R-Instat library, I wanted to select the variables.  I found that the selection of the variables receiver seems not to permit date variables?  Could you check this and make sure that any type of variable is acceptable.  Then I'd like to see this merged as soon as possible.

rdstern (dismissed):
@Raymond-Andam great, thanks.  @N-thony could you check and merge, when you are happy.  This is a really nice new feature.

rdstern:
@N-thony you approved earlier.  Is it still ok - so could it be merged.  It is a nice addition.

derekagorhom:
@N-thony i have send you a message on skype, can you respond to it please

Raymond-Andam:
@N-thony The requested changes have been made.

rdstern (commented):
@Raymond-Andam and @N-thony I thought it was still working well.  But then I found a curious problem, just at the end.  
image

Here is the code that has given the error:

```
# Dialog: Visualise Data
colnames <- data_book$get_columns_from_data(data_name="survey", col_names=c("field","size","fert","yield","village"), force_as_data_frame=TRUE)
last_graph <- visdat::vis_value(colnames, viridis_option="A", na_colour="grey")
data_book$add_object(data_name="survey", object_name="last_graph", object_type_label="graph", object_format="image", object=check_graph(graph_object=last_graph))
```
It shows the problem, namely that this open does not allow facets, but the village variable, that was the facet variable (village) in the earlier options is still included in the command?

I also notice, oddly, that the multiple receiver still says there are 6 variables.  It has not refreshed as we changed the options at the top.

I suspect a simple solution is to clear the multiple receiver when moving to, and from this option.  @Raymond-Andam unless you can make the corrections quickly, I suggest that perhaps @N-thony should make the smell change and you check what he has done.  If so, then @N-thony could still approve.

rdstern:
@N-thony thanks, that's great.  And if @derekagorhom can check what you have done, then you and I can still approve and merge.

derekagorhom:
@N-thony , @rdstern , i have had a look and this works.
Thanks

rdstern (approved):
@N-thony this seems to work fine now.  I thought it had been merged long ago, but not.  Over to you.

PR #8673, opened 2023-11-30
Improved the Add and R Help buttons with scrollbar on the General Graphic Dialog

N-thony:
@rdstern I have the change for the Add and R Help in the General Graph since both have long list and not display nicely for users.
image
image

now they are using a scrollbar, have a look
image
image

### DIFF
--- a/instat/ucrAdditionalLayers.vb
+++ b/instat/ucrAdditionalLayers.vb
@@ -88,9 +88,9 @@ Public Class ucrAdditionalLayers
 
     Private Sub ucrAdditionalLayers_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         SetEditDeleteEnabled()
-        cmdAdd.AddItemsToListbox({"geom_boxplot", "geom_bar", "geom_count", "geom_col", "geom_density", "geom_tile",
-                                "geom_jitter", "geom_histogram", "geom_contour", "ggridges", "geom_density_ridges",
-                                "geom_label", "geom_label_repel", "geom_line", "geom_point", "geom_rug", "geom_smooth"})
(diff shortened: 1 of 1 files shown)


rdstern (commented):
@N-thony very nice.  I am also happy with the size, but just wondered if a bit longer may be even better?
But then they stopped working, which makes me wonder about the code?  I used it twice for the geoms and was pleased it also remembered its place in the list.   Then I went to the R-help and that worked fine.  Then back to the geom and then the second time, the help drop-down didn't work.  
From then on neither down arrow works.  Nothing happens.  I tried reset, different data frame, closing the dialog.  I have to start R-Instat again for them to now work again.

rdstern (commented):
@N-thony in the geoms the order is odd and some are repeated.  They are ok and alphabetical in the long version:
image

I can't though get it to stop working so that seems well fixed.

rdstern (approved):
@N-thony that's very nice now.  @Patowhiz good for you to check and merge.

Patowhiz:
@N-thony why did you choose to add a list box to the split button instead of just making the slit button context menu scrollable?

rdstern:
@Patowhiz yes you can have the same geom multiple times.  This could be for the same layer, but the more common examples are when you plot, say tmax and tmin against date.  You start with tmax against date, say with geom_line and then again for tmin.
I like your query for @N-thony.  If I understand correctly, you feel we could be re-using an existing custom control, (changing the properties), rather than adding a new one?  This is becoming an important, and one of our nicest dialogs - in my view anyway - so I look forward to merging the improved version.

Patowhiz:
@rdstern thanks for your feedback.
@N-thony try looking into whether the context menu can be made scrollable and it's size fixed. If that's possible, then we won't need a list box, which means no additional controls and public functions will be needed.

rdstern:
So @Patowhiz can this now be accepted and merged?  I am keen to proceed well, but also quickly - so not striving for perfection if what is here is ok.

PR #8674, opened 2023-11-30
Some Change Made on Climatic>Describe Menu

### LINKED ISSUE #8660: Improving the Climatic > Describe Menu
This is currently a pretty pathetic menu.  It currently only has one working dialog, and that is the only one that is a sub-sub-sub menu:

image

I suggest we make it into an impressive menu ready for the January release, and even describe it in the new wmo guide.

Most of the items are copeid or moved from elsewhere.  Some need to be tested carefully, because they use packages updated in 2023 and we need to do that anyway!

a) `Climatic > Describe > Summarise 2/3 Variables` - this is simply a copy of the `Describe > Two/Three Variable > Summarise` dialog.  It is being improved but that's ok.  It is simply a copy at its current state.
b) `Climatic > Describe > Graph 2/3 Variables` - this is simply a copy of the `Describe > Two/Three Variables > Graph `dialog.
c) A `separator line`.  (Note we have most of the Describe > One Variable dialogs earlier as part of the tidying and initial viewing of the data.  The pivot tables are also earlier.  The Correlation dialog is later in the Climatic > Compare menu.  I propose we leave the General graph and table dialogs just in the describe menu, but mention them in any documentation.
d) `Climatic > Describe > SPI/SPEI`.  This is moved, rather than copied from the Climatic > Prepare menu.  It uses monthly data so is more appropriate here.  This package has been updated and we need to check we are taking advantage of the latest version.  We need this to be working well, because I think an old version is in ClimPact2 and we should be including that now.
e) `Climatic > Describe > Climograph`.  This is new and described in #8657 .
f) `Climatic > Describe > IDF`.  We are working on this now.  I suggest the climatol version is appropriate here.  If we use the IDF package, that would go under the Climatic > Models menu.
g) A `separator line`. 
h) The 3 last dialogs from the Structured Menu, namely Wind Rose, Wind/Pollution Rose and Other Rose Plots.  This is a copy here. I hope we will have time to check these dialogs toom from the recently updated openair package, but that work has lower priority.  They are copied over anyway.

MeSophie:
Fixes #8660 
@rdstern I have made change on Climatic >Describe Menu please have a look.


rdstern (changes requested):
@MeSophie that's great.
a) Please can you return the 3 rose plots to the structured menu.  They should be there as well.
b) Then get rid of the original Climatic > Describe menu items.  So it starts with the Summarise 2/3 Variables.
c) Also ad the `...` after each entry.  Thery all lead to a dialog.
d) Then the hard work will begin, on the Climograph - but itwould be good to merge this first.

MeSophie:
@rdstern Please can you check the change?

rdstern (commented):
@MeSophie sorry, I am clearly not explaining myself well.
a) The "old" menu - from Rainfall, down to General, is not needed.  It does nothing.
b) Then, under IDF the 3 circular ones from the structured menu are needed.  Copy them from the Structured menu, but also don't delete them from there.  They will appear twice.

MeSophie:
@rdstern Please can you check again?

rdstern (dismissed):
@MeSophie that's great.  Thanks.  @N-thony could you test and merge please

MeSophie:
@N-thony , @Patowhiz Please it will be good if we can review this PR and merge. Thank you.

rdstern (approved):
@N-thony still looks fine to me.  So back to you.  There's lots still to do on this menu, but good to merge this first, when it is ok?

PR #8689, opened 2023-12-14
 Added key and link in Restore Hierarchy option on Prepare > Data Reshape > Unstack dialog

### LINKED ISSUE #8686: Add key and link in Restore Hierarchy option on Prepare > Data Reshape > Unstack dialog
@N-thony I provide the example to motivate this new feature.  It should be very easy for anyone who understands a bit about our Prepare > Data Reshape > Column: Summarise dialog.  Perhaps this is you?  The code from there needs to be copied.  Perhaps it would be simpler if I share my screen and explain, but see below.

Use the library data, called evals, from the moderndive package.  This is data on 463 University courses, given by 94 staff.
Make the second variable into a factor.

image

Now use the Prepare > Data Reshape > Unstack, and the restore hiererchy option.  The prof_ID is the factor and age to rank are the variables at the prof-ID level to carry.

This works perfectly.  But it should also make prof_id into a key variable, and link the evals data frame - with this variable to the new data frame.

To understand what is needed you could use Prepare > Data Reshape > Summarise Columns.  Then summarise the score, getting, for example, mean, median, max, min, range and count-non-missing.  That includes the key and link that is needed.  So, all that needs to be done is to implement that code in the Restore Hierarchy function.

N-thony:
Fixes #8686 
@rdstern have a look,

### DIFF
--- a/instat/dlgUnstack.vb
+++ b/instat/dlgUnstack.vb
@@ -365,7 +365,9 @@ Public Class dlgUnstack
                 clsAddKeyFunction.AddParameter("data_name", Chr(34) & strDataFrame & Chr(34), iPosition:=0)
                 clsAddKeyFunction.AddParameter("col_names", ucrReceiverFactorToUnstackby.GetVariableNames(), iPosition:=1)
                 clsAddLinkFunction.AddParameter("to_data_frame", Chr(34) & strDataFrame & Chr(34), iPosition:=1)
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@N-thony that's excellent.  @Patowhiz please can you check the code and then merge.  It is a great example of R-Instat making life simple for users.  I really like it!

Patowhiz (approved):

PR #8699, opened 2023-12-20
Fixed Bug in the Edit Cell dialog

### LINKED ISSUE #8664: Bug in the Edit Cell dialog
@derekagorhom this is quite an odd one.  

Use any dataset - see below with the survey.  Then use a filter that does not select consecutive cells.  I selected New varieties below.  Then it gives the correct tow, but alwaways the top value to edit.

image

Fun one to find!

derekagorhom:
Fixes #8664
@rdstern  this PR is ready for review
Thanks

### DIFF
--- a/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
+++ b/instat/UserControls/DataGrid/ReoGrid/ucrReoGrid.vb
@@ -219,8 +219,8 @@ Public MustInherit Class ucrReoGrid
     Private Function GetRowIndex(currWorkSheet As Worksheet, strRowName As String) As Integer
         If currWorkSheet IsNot Nothing Then
             For i As Integer = 0 To currWorkSheet.Rows - 1
-                Dim strCol As String = currWorkSheet.RowHeaders(i).Text
-                If strCol = strRowName Then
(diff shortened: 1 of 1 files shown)


rdstern (commented):
@derekagorhom thanks.  It is improved, but still the behavior is not quite the same was without the filter.  You are working well, so I am feeling this sort of thing is something you should also be checking yourself, before expecting me to test?

So it is now in the right place and I can edit.  But when there is no filter, then it remains in the same cell afterwards.  When there is a filter it resets, to the left.  I haven't checked if it resets to the top, because I didn't have enough date in the filter.

It is usable if the change is a consequence of the filter, and so difficult to do anything about.  Could you either fix, so the behavior is identical, with and without a filter.  Or report on the difference, with a reason that this is difficult to do anything about.

rdstern:
@derekagorhom I wonder what you will do on this one?

derekagorhom:
@rdstern I had a conversation with antoine on this and we suggest this be merged first since it already fixes the main issue, while we look at it in a different PR, i have also realized this behavior problem is also occurring in issue #8711 (I hope i am wrong). 
What do you think?

rdstern (approved):
@derekagorhom I agree, that's fine.  @N-thony over to you

PR #8700, opened 2023-12-20
Corrected the message when running the Tools > Install R Package

### LINKED ISSUE #8632: Correct the message when running the Tools > Install R Package
@N-thony we have agreed that we will keep the current limitation on this dialog, that you have to run R-Instat as an adminstrator to be able to add an R package.

However it used to allow you to check whether the package is available before.  Now it quickly tells you you can't install.  

a) I would like it to check - as before - whether the package is in CRAN and whether it is already installed and up-to-date etc.  That's without the current message that you can't write there.  The message used to come later, i.e. when you pressed Ok
b) Can you change the mesage (again!) when there isn't a package.
image

The second sentence could become `Perhaps spelled wrongly, or archived?`

@rachelkg the Help goes to the right place.  Adding help would be useful here.  
This dialog allows you to check whether a particular R package is supplied with R-Instat.  If not then you can add or update an R package, but you need to run R-Instat as an administrator to do this.  That's when you start R-Instat.  Instead of clicking on the Icon, right-click and choose to Run as Administrator.

image

derekagorhom:
Fixes #8632
I have corrected the message for the R package from `Not a current CRAN package. Perhaps it is been  archived?` to `Not a current CRAN package. Perhaps spelled wrongly, or archived?`

@rdstern  this PR is ready for review
Thanks

### DIFF
--- a/instat/dlgRPackages.Designer.vb
+++ b/instat/dlgRPackages.Designer.vb
@@ -55,18 +55,18 @@ Partial Class dlgInstallRPackage
         Me.ucrInputMessage.AutoSize = True
         Me.ucrInputMessage.IsMultiline = False
         Me.ucrInputMessage.IsReadOnly = True
-        Me.ucrInputMessage.Location = New System.Drawing.Point(105, 81)
+        Me.ucrInputMessage.Location = New System.Drawing.Point(95, 81)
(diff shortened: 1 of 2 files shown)


rdstern (approved):
@N-thony very simple change to check.

PR #8701, opened 2023-12-20
Adjusting the auto save dialogue to fix overlappping in french layout

Fidel365:
fixes partially # 8693.
@rdstern this fixes the overlapping issue. On adding the title to the database, I will clarify with @N-thony  first. Lastly, how would you like us to implement this dialogue on the edit menu?

### DIFF
--- a/instat/dlgAutoSaveRecovery.Designer.vb
+++ b/instat/dlgAutoSaveRecovery.Designer.vb
@@ -49,15 +49,15 @@ Partial Class dlgAutoSaveRecovery
         Me.cmdSaveData = New System.Windows.Forms.Button()
         Me.lblBackupDataDetected = New System.Windows.Forms.Label()
         Me.lblDataSavedTo = New System.Windows.Forms.Label()
-        Me.ucrInputSavedPathData = New instat.ucrInputTextBox()
-        Me.ucrInputSavedPathLog = New instat.ucrInputTextBox()
         Me.lblLogSavedTo = New System.Windows.Forms.Label()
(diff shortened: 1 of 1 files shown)


rdstern:
I got there and it is much improved.  The buttons at the bottom need to be wider.  
And, of course the main text needs to be added.  @Fidel365 I think you mentioned that?

rdstern:
@N-thony I also add some questions on this dialog for you and also @Patowhiz 

Here is the current recovery dialog:

image

 
a) Should we be able to access a version of this dialog, or adaptation when we start any session, so not just when we collapsed last time.  So is the past log file and the last backup of the data saved anyway, or only when we crash?
b)  If so, then where should it go.  We have a disabled Tools > Load Options currently.  Should we change that to Load Backup and run a version of this dialog from there as an additional option?
c) If so, then I propose it is available at the start, and is disabled as soon as you start a new session, i.e. once you load a dataset, etc.
d) At the same time I don't understand the internal debugging log.  That's the third option to save.  What is it?
e) It also defaults to sending the debugging log to the R-Instat team.  That sounds really professional, but I must have sent lots over the years and have no idea where they are or what to do.  I suggest we untick that option and disable it for now.
f) There is also an inactive Help button.  The Load options has a help ID of 411.  Could that help number be used anyway?
g) If we do proceed, then this should probably become an issue.  Could we please also look at some of the other inactive options on the tools menu?

rdstern (commented):
@N-thony please can you check.  If you are ok, then I'll approve so you can merge.

rdstern (approved):

PR #8717, opened 2024-01-08
Swapping Data View and Log/Script Windows

### LINKED ISSUE #8712: Add a View for Data with (and not in) R-Instat?
@N-thony don't do this till I check with @ChrisMarsh82 and @lloyddewit that it is a good idea?!  But you might like to confirm if it is quick and easy to do?

With the new version of R-Instat we have easy access to loads of R scripts.  They generate or use data and then do analyses, but they don't (automatically) put them into an R-Instat data book.   Usually we want to do that and the new Insert dialog makes it easy.  But otherwise they just run and we get the results in the Output Window.

I am considering this as "Analysing data **_with_** R-Instat" as opposed to **_in_** R-Instat.

When we analyse **_with_**, then the data view is irrelevant (as are the metadata windows). 

The important windows are the Script window and the Output Window.  

Now you already have a useful option in the View Menu, namely `Swap data and Metadata`  Could we also have `Swap data and Script`?  Then the script is on the left and the results are on the right?  The other obvious possibility is already there, namely Script above and Output below.  (Just open the script and close the data view!)

N-thony:
Fixes #8712 
@rdstern @lloyddewit have a look. Thanks

### DIFF
--- a/instat/dlgScript.vb
+++ b/instat/dlgScript.vb
@@ -117,6 +117,7 @@ Public Class dlgScript
         ucrBase.chkComment.Visible = False
         ucrBase.txtComment.Visible = False
         ucrBase.bAddScriptToScriptWindowOnClickOk = False
+        ucrBase.bMakeVisibleScriptWindow = False
 
     End Sub
 
--- a/instat/frmMain.Designer.vb
+++ b/instat/frmMain.Designer.vb
@@ -5104,7 +5104,9 @@ Partial Class frmMain
         Me.MainMenuStrip = Me.mnuBar
         Me.Margin = New System.Windows.Forms.Padding(4)
         Me.Name = "frmMain"
-        Me.Text = "R-Instat .."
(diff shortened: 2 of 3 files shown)


rdstern (commented):
@N-thony this looks great.  I really like it.
a) When I open a dataset, using Insert, the 2 screens changes to 3, with the original data view appearing, as so:

image

This isn't needed.  I like the fact I can have the data like this, but it should not be automatic, every time I open a new dataset!

b) And a separate small point, and perhaps not this issue.  As the window is now half the screen, could the gap between the sets of buttons be slightly smaller, and could the remaining buttons become a bit smaller - perhaps the same size as the Run and Run All buttons?

Very exciting!

rdstern (approved):
@N-thony this now seems to work.  It is great and I am approving.  @ChrisMarsh82 I wonder if you could check quickly and merge?  If not, I'll ask Patrick.  Thanks

Roger

ChrisMarsh82 (approved):

PR #8718, opened 2024-01-08
Fixed Bug And Added Facet Controls On General Graphics Dialog. 

### LINKED ISSUE #8695: Correct small bug and add another feature to the General Graphics dialog.
@MeSophie this is only after your pull request #8656 is merged.

a) There is a small bug that when a factor is added as the x-variable you have the option to add the group=1 option.  That's all correct.  But if you now change the x-variable to become a numeric variable, then the control correctly becomes invisible, but the code remains.  So it still adds to group=1 option.

b) Now a suggestion for a neat addition - I hope you will agree.  There is now spaceand it would be great to be able to include facets (one facet only) without going to the sub-dialog!
, 
That's of course quite complicated, except we already have the control and code from Danny in the PICSA dialogs.  Here is a picture:

image

I hope you can simply copy the control and the corresponding code.  In the label, omit the Optional, because that's obvious!

MeSophie:
Fixes #8695
@rdstern, @N-thony   I have fixed the bug and added facet controls on general graphics dialog. Please have a look.

### DIFF
--- a/instat/dlgGeneralForGraphics.Designer.vb
+++ b/instat/dlgGeneralForGraphics.Designer.vb
@@ -40,17 +40,11 @@ Partial Class dlgGeneralForGraphics
     Private Sub InitializeComponent()
         Me.components = New System.ComponentModel.Container()
         Me.grpAethetics = New System.Windows.Forms.GroupBox()
-        Me.ucrChkUseasNumeric = New instat.ucrCheck()
-        Me.ucrReceiverY = New instat.ucrReceiverSingle()
         Me.lblYVariable = New System.Windows.Forms.Label()
         Me.lblXVariable = New System.Windows.Forms.Label()
-        Me.ucrReceiverX = New instat.ucrReceiverSingle()
         Me.lblLabel = New System.Windows.Forms.Label()
-        Me.ucrLabelReceiver = New instat.ucrReceiverSingle()
         Me.lblColour = New System.Windows.Forms.Label()
-        Me.ucrColourReceiver = New instat.ucrReceiverSingle()
         Me.lblFill = New System.Windows.Forms.Label()
-        Me.ucrFillReceiver = New instat.ucrReceiverSingle()
         Me.ContextMenuPackagesList = New System.Windows.Forms.ContextMenuStrip(Me.components)
         Me.ListBaseToolStripMenuItem = New System.Windows.Forms.ToolStripMenuItem()
         Me.GeomtextpathToolStripMenuItem = New System.Windows.Forms.ToolStripMenuItem()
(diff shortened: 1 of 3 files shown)


rdstern (commented):
@MeSophie that's really excellent - very well done.  This is becoming a wonderful dialog.
a) In the Add Code control I got the arguments wrong for the hline and vline.  I assumed we could shorten yintercept to y and xintercept to x.  But it seems not.  So, in the Add Code, could you give those in full?
b) In the Add Code, could you add a blank, and make that the default?  Then the pull-down shows the other possibilities - as you have. 
c) The facets seems to work well.  When I then look at the facets sub-dialog, it has implemented facets, but has not included the variable I used on the main dialog.  I don't know if this could be added.  Then the result from facets on the main dialog will be the same as using the sub-dialog?
d) Now a more questionable change.  I am not good on design, so you may wish to discuss this suggestion with Antoine first.  I would like the Add Code drop-down to be wider, so it is easier to type and correct the code.  So more code is visible.  
My suggestion is to move the Legend control to where you currently have the Add Code control.  And make it a single row, as you have for the Add Code.
Then raise the Geom control a bit and put the Add Code, (on 2 lines) below it.  And make the pull down go to the right as far as you can.  
Finally I wonder a bit about the "Add Code" label?  It adds a plus to the ggplot command - does that have a technical name?  If we keep the label Add Code, then what would you give in French?  Is it clear.  And, if we keep the Add Code, then it needs a colon after it.

MeSophie:
@rdstern  I have fixed part a) b) and d) I'm still on part c).
image
Here can we change the label to something like **Add code to customise the graph:**? It is true that just **Add Code** is not clear for user.

rdstern (commented):
@MeSophie one small change is that you have yintercept = 5, etc correctly, but (for the first 3 examples of the code that should be instead of the y = 5 and not as well, so could you delete those.
Then I now find the space for the command line is wider than I had thought.  As mentioned previously I wonder about moving the Facet and the Legend row down a bit, so the Legend label is where the Add Code label is now.
Then the Add Code checkbox goes underneath the Geoms control and the field that then opens is underneath and goes from the middle right to the right.  @lilyclements I wonder if you could look at the layout and see if you agree?
 Then I would like to merge, even while you see if item c can be resolved.

MeSophie:
@rdstern please can you have a look at the new changes? Part c) is also fixed.

rdstern (dismissed):
@MeSophie I am approving so @N-thony can check and merge.  I am writing the help this weekend, on this dialog.
I still would have preferred the Add Code checkbox up one line and the pull down starting below the Add Code, rather than next to it.  If that is clear and you are happy to make that small change, then please do.  Otherwise it is ok as it is/
.

MeSophie:
@rdstern  I made the change on the design. I moved Add Code a little to the right because of the french translation of Plot Options.

rdstern (approved):
@MeSophie great.  @N-thony over to you to check and hopefully merge.

derekagorhom (dismissed):

MeSophie:
Sorry @rdstern  and @derekagorhom there was a small bug with facet. I fixed it please can you check the change again?

derekagorhom:
@rdstern,  problem with facet @MeSophie found on my screen is that the `~` operator was after the parameter and it was producing an error, I think that is fixed now.

rdstern (approved):
Still seems ok, so approving again.  @N-thony back to you.

PR #8719, opened 2024-01-08
Fixed bug in Column Selection > Partially labelled option

### LINKED ISSUE #8705: Select with partial labels should only work in one direction
@N-thony The partial label featyre you added is designed for situations where there are more (numerical) values than value labels.

The other way round, namely more labels than levels is quite common and is different and should not be included.  That's when there are extra factor levels with zero frequency and we handle that differently.
I found this problem with IPUMS data, where they are from one country, but the labels gives all the possible countries.  You must be doing a check somewhere for not equal, when ut should be that that there are more values in the data than in the labels.

N-thony:
Fixes #8705 
@rdstern check this if it is okay now.

### DIFF
--- a/instat/dlgSelectColumns.vb
+++ b/instat/dlgSelectColumns.vb
@@ -53,7 +53,7 @@ Public Class dlgSelectColumns
         ucrInputSelectOperation.SetItems({"Columns", "Starts with", "Ends with", "Contains", "Matches", "Numeric range", "Last column", "Where"})
         ucrInputSelectOperation.SetDropDownStyleAsNonEditable()
 
-        ucrInputColumnType.SetItems({"Numeric", "Factor", "Character", "Logical", "Variable label", "Value label", "Empty columns", "NA columns", "Partially labelled"})
+        ucrInputColumnType.SetItems({"Numeric", "Factor", "Character", "Logical", "Variable label", "Value label", "Empty columns", "NA columns", "Partially labelled", "Unused label"})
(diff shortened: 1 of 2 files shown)


rdstern (commented):
@N-thony it is still giving me columns where there are more labels than values, rather than just those where there are more values than labels.  I see the function also on the right and wonder about the last part, where it says  `(fn=is.containPartialValueLabel))), and_or="|")`  Is the second part `and_or="|"` needed?  Does it cause any problems?


image

rdstern (commented):
@N-thony it now seems to work, so I am approving.  Well done.
I would like @Patowhiz to check, but first I wonder if you could see if there is anything you could do with this, and perhaps other options to make it work faster?
In the data I sent, I took the hh data and looked for the partially labelled variables.  There are 37.  I found it executes the command when you press Return on the sub-dialog, rather than when you press Ok.  I am not sure if that is correct.
a) It takes a long time.  
b) So I then undid it - cleared and it took ages before getting the dialog - and closed the screen twice during that time.
c) I have a feeling that everything may take a long time once the select is in place?  So it only works slowly?
d) I then decided to get another select of all the variables once I had this select.  That took a long time to do.  But then re-getting that select was very quick

rdstern (approved):

rdstern:
@Patowhiz could you check and approve, so this can be merged?

Patowhiz (approved):

Issue #8694, opened 2023-12-15
Make Filter and Select dialogs more consistent


rdstern:
**EDIT: All points here are complete except (g)**

@N-thony I think all except one of these options is trivial.  I hope option g) is easy, but otherwise we can merge and this becomes a separate issue.

Here are the 2 current dialogs:

![image](https://github.com/africanmathsinitiative/R-Instat/assets/11226469/278d96b2-aa53-4687-a3f1-b6b7cc122ce0)

And for filters:

![image](https://github.com/africanmathsinitiative/R-Instat/assets/11226469/36bac872-e77f-4f55-8df5-1c6b1a4abc57)

a) Change the `Filter` title to `Filter Rows`
b) In Filter change `Apply Options` to `Apply`
c) In Filter delete the Apply from the 2 options, so it is just `As Filter` and `As Subset`
d) In Filter add a third button with label `Save Filter`.   (This is like the option to Save the select.)
e) In Filter get rid of the word Selected in Selected Filter Preview.  So it is just Filter Preview.
f) In Select Delete the Apply from the first button, so it is just `As Selected Columns`.
g) In Select add the `Select Preview ` control if possible to parallel the Filter Preview.
h) Check the database that these phrases are translated.

Fidel365:
I will be working on this.

lilyclements:
@Fidel365 I note in #8830 you mention you're working on (g) in a separate PR. Would you mind linking to that? Or if that work has ceased then let me know so we can put (g) in a separate issue.

lilyclements:
@Ag-Derek can you assign?

Ag-Derek:
@anastasia-mbithe can you take this up
Thanks

anastasia-mbithe:
@Ag-Derek yes I can.

PR #8723, opened 2024-01-10
Changes to the Filter and Select dialogue to make them more consistent

Fidel365:
Fixes partially issue #8694 
This fixes parts b,c, g and partially d. I will first consult with @N-thony for the remaining part of the issue.


rdstern (commented):
@Fidel365 I like your comment where you detail exactly what you have done so far.
1) You say partially d).  But make sure you understand exactly what Save Filter is doing.  Then you will realise that you copied the Subset option and you should have copied the simpler Use option instead.  We don't need a new data frame.  Then you should realise that Use Filter, will save the filter and then apply it.  The Save Filter will save it and won't apply it.  So it should be much simpler and you should be able to complete it.
2) You say you have done g) and I see the control is there.  But the contents are blank.  So you do need to finish g.
If, for efficiency you ask @N-thony for help and he completes it, then please study what he has done, and test it and report on that, before I need to test again.

PR #8749, opened 2024-01-22
Fixes Export of Workspace 

Patowhiz:


rdstern:
@Patowhiz I can't test with my quaint method.  But it says all checks are passed.  Doesn't that mean you can merge and then I'll look at the merged version.  I'd be ok with that.  I see you made the changes 2 months ago.

Patowhiz:
If you can approve it then @N-thony can merge. Note this only fixes the export bug to make sure the dialog works. The objects won't be imported back you import the file in R-Instat but they should be part of the saved rds file and can be viewed in R studio.

rdstern (approved):
@N-thony if you are ok, this is important and I would really like to see it merged.
@Patowhiz you mention this is a one-way process, and that's fine.

Issue #8739, opened 2024-01-18
Fault in boxplot dialog when using sub-dialog


rdstern:
@MeSophie etc

rdstern:
@N-thony could you allocate please.  I thought @MeSophie, but she already has quite a lot for James.


a) When doing the violin plot it works with a single variable and no x given explicitely.  For example carat for the ggplot2 diamods data.
It does this by adding `x=""` into the code.  But if you return to the dialog and use the down arrow to go into the Violin plit sub-dialog, then out again and try, then it gives an error.  And the `x="" is no linger in the code.

(Curiously I think there should be the same problem with the boxplot, but it seems ok.)

b) On the main dialog in the left button at the top it should just say `Boxplot`.  Please delete the `\Tufte Boxplot` from the label.

c) If easy could you swap the second and third buttons, so the second is Violin and the third is Jitter.  (Don't do this change if it could cause complications.

lilyclements:
Completed

PR #8751, opened 2024-01-22
Changing the label  Boxplot \Tufte Boxplot to Boxplot in Boxplot dialog

Vitalis95:
Fixes partially #8739 
@rdstern @N-thony , It changes the button label from` Boxplot \Tufte Boxplot `to `Boxplot`

Here is the database change;

image


rdstern:
@Vitalis95 I still get the problem with the violin plot.  If I do a single violin plot, say yield from the survey data it is fine.  If I come back to the dialog and repeat it is fine.
But if I use the Plot Options down arrow and go to the subdialog and then press Return - I don't do anything - then it gives an error and says it needs the x.

Vitalis95:
@rdstern , this PR is changing the label and also updating the database. The bug is solved in PR #8750

rdstern (approved):

PR #8753, opened 2024-01-23
Made the Context Menu Control in the Split Button Scrollable

N-thony:
@rdstern this replaces PR #8673 in case we decide to use the same control i.e `contexmenustrip` in the split button. I made it scrollable, have a look.

### DIFF
--- a/instat/ucrSplitButton.vb
+++ b/instat/ucrSplitButton.vb
@@ -585,9 +585,60 @@ Public Class ucrSplitButton
 
         State = PushButtonState.Pressed
 
-        If _contextSplitMenuStrip IsNot Nothing Then
-            _contextSplitMenuStrip.Show(Me, New Point(0, Height), ToolStripDropDownDirection.BelowRight)
+        Dim tmpForm As New Form With {
+            .AutoScaleMode = AutoScaleMode.None,
+            .FormBorderStyle = FormBorderStyle.None,
+            .StartPosition = FormStartPosition.Manual,
+            .ShowInTaskbar = False
+        }
+
+        '' Event handler for item click
+        'Dim itemClickHandler As EventHandler = Sub(sender As Object, e As ToolStripItemClickedEventArgs)
+        '                                           ' Handle the item click here
+        '                                           ' You can perform actions based on the selected item
+        '                                           ' ...
+
+        '                                           ' Detach the event handler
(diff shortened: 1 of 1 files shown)


rdstern:
@N-thony great that you are back to finish this.
I like the fact it is scrollable.  I like this one - is that now yours?  I am pleased this has not changed.

image

Now yours looks roughly the same as the scroll bar:

image

and:

image

Except it is too short and the spacing between items remains large.  So, for example, I would not expect the last one to need a scroll bar.  

By the way I notice that the spacing is changed in @MeSophie 's latest pull request on this dialog.

rdstern:
@N-thony that's better.  One problem resolved, namely the contol is a good length now.
But when the blank lines are removed we don't need that scrolling element for the pull down on the plot options.
It is also there on ok and keep, where there are just 2 elements.  It really doesn't need to be everywhere, just when the number of entries is long.
And we do need to get rid of the spaces between the items as well.

image

That's more spread out than this one!

image

rdstern:
@N-thony I am pleased there is no scroll bar now on the Plot Options down arrow.
I still don't understand the spacing between the geoms, and maybe the control here, and on the main general graph dialog:
image

Does it have to be different to the one on the geoms here?:

image

I am a bit at a loss of what the problem is here?  I assumed it would be a very small task?  Do we need a skype to discuss?

Just to mention that the list of common geoms on the main dialog and the drop down in the sub-dialog, will not get much longer, but we will be adding to it.  It could get up to double its current length.

rdstern:
@N-thony thanks for the explanation.  I assume the increased spacing, between items, is due to this control?  Anyway this is not a big deal, so I am approving.  There are many more urgent improvements, and this works and is just cosmetic.

rdstern (approved):
@Patowhiz can you check and merge please.

Patowhiz:
@rdstern as explained above. I know it works but the code implementation introduces new ways of using the control instead of building on existing functionality.  This in the long term leads to unnecessary complexity. Like I said before,  I can merge this if you need it urgently(taking into account my comments) and if @N-thony finds my above suggestion time consuming to implement.

rdstern:
@Patowhiz @N-thony claims he has fitted in with your suggestion, and explained that above.  So it is just to check the code and merge.  Thanks

Patowhiz (approved):

Issue #8758, opened 2024-01-24
More improvements in Describe > Graphs > Boxplot and other dialogs


rdstern:
@N-thony please could you allocate.  I think @Vitalis95 might be happy to work in this area, but you choose.  There will be more improvements in the five key graph dialogs, so there will be plenty of useful tasks soon.  This suggests an improvement on each of the five dialogs

This is the existing dialog:

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/45f76345-8796-41db-9014-d3935c13a0eb)

There are various improvements that can make the dialog even more exciting.  I am listing them one at a time:

There are 2 new controls in a nice row at the bottom of the Describe > Graphs > General dialog:

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/f4a98e75-49ae-4b1b-bc27-8c0d177fcbd1)

Please could they be added to the same position of the Boxplot and the other dialogs.

They should be in a line, and the code should be such that when used it is the same as when either is used from the sub-dialog.  I assume the code will be identical, so it is all part of the control and hence only done once.

Please could this be added to the boxplot dialog.  Ideally, it may be efficient to add it, at the same time, to the other 4 graph dialogs  (bar, scatter, line, histogram).  I leave up to @N-thony to propose whether this will be one pull request or five separate pull requests.

PR #8766, opened 2024-01-25
Improvements in Describe > Graphs > Boxplot dialog

Vitalis95:
Fixes Partially #8758 
@rdstern , @N-thony , The PR adds` 'legend' `and `'facets'` options to the boxplot option in the Boxplot dialog

### DIFF
--- a/instat/dlgBoxPlot.vb
+++ b/instat/dlgBoxPlot.vb
@@ -112,7 +112,7 @@ Public Class dlgBoxplot
         ucrPnlPlots.AddFunctionNamesCondition(rdoJitter, "geom_jitter")
         ucrPnlPlots.AddFunctionNamesCondition(rdoViolin, "geom_violin")
         ucrPnlPlots.AddToLinkedControls(ucrChkAddPoints, {rdoBoxplotTufte, rdoViolin}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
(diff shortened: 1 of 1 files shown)


rdstern (commented):
@Vitalis95 this is brilliant.  I tried with both boxplots and with the general.  Both work and I note that both also set the control in the sub-dialog - which is great.
Could you add the two controls, identically in the jitter and violin plot options.  Then we can merge.  And later they need to go into the other graph dialogs.
@Vitalis95 and @N-thony I assume it is difficult to code "the other way round"?  If I make the change in the sub-dialog, then I didn't expect it to reflect automatically in the main dialog.  But I wondered if it might set the main dialog when the boxplot dialog is re-opened?

rdstern:
@N-thony you asked " don't understand that, isn't it a bit of complication?. When you make the change on the sub dialogue, it should reflect directly in the main dialogue. Why it should reflect after re-opening the dialogue?"

That's because I was wrong.  I assumed it would be difficult to sybchronise the two controls completely.  I had only checked the consistency from main dialog to sub-dialog, after re-opening the dialog.  I now see that going from main dialog to sub-dialog it immediate - they are really synchronised.  So my new request is whether this synchronisation could be "both ways".  That would be perfect.

Vitalis95:
@rdstern , the change I made is adding the two controls in the jitter and violin plot options

PR #8778, opened 2024-01-31
Fixes issue where the Levels/Labels dialog does not label the last level well

### LINKED ISSUE #8747: Levels/Labels dialog does not label the last level well
This dialog is on the Column right-click menu.  It is also on Prepare > Column: Factor > Levels/Labels

Here it is with the amis data from the boot package:

image

After importing from the library make all variables except the first into Factors.  Then Use the Levels.Labels dialog to change the Labels

image

For period you can make them as shown into before, when, after.
For Warning they can be yes, no.

The problem is the last label.  You are usually still in the cell and press Ok.  It then doesn't register the last changed value.  So, in the example above you get before, when and 3 as the labels.  

Could this either be fixed by registering the value, - that's what I would like.  If that is difficult could Ok be disabled while you are in the cell.  Then you will lmow to click somewhere else to activate Ok.

@N-thony I suspect you could do this very quickly.  But you may also wish to allocate?

derekagorhom:
Fixes #8747 
This fixes the problem where the levels/labels dialog does not label the last level being changed
Thanks

### DIFF
--- a/instat/ucrFactor.vb
+++ b/instat/ucrFactor.vb
@@ -936,4 +936,25 @@ Public Class ucrFactor
         lblSelected.Visible = iSelectCol > 0
         SetToggleButtonSettings()
     End Sub
+
+    Private Sub _grdSheet_CellEditTextChanging(sender As Object, e As CellEditTextChangingEventArgs) Handles _grdSheet.CellEditTextChanging
+        Dim bValid As Boolean = True
+        'do levels entry validation
+        If _grdSheet.ColumnHeaders(e.Cell.Column).Text = DefaultColumnNames.Level Then
+            If Not IsNumeric(e.Text) Then
+                MsgBox("Levels must be numeric values", MsgBoxStyle.Information, "Invalid Value")
+                bValid = False
+            ElseIf e.Text.Contains(".") Then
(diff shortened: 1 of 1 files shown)


rdstern (commented):
@derekagorhom it works perfectly - except it is working too well.  It labels the last level in all the factors!  I just want it to apply to the factor I was labelling!

I did Yes/No on a different factor and look here on a new factor!

image

derekagorhom:
> @derekagorhom it works perfectly - except it is working too well. It labels the last level in all the factors! I just want it to apply to the factor I was labelling!
> 
> I did Yes/No on a different factor and look here on a new factor!
> image

@Patowhiz or @lloyddewit can any of you help me to understand why when i try to use Refresh() i still end up with the same problem?

Patowhiz:
@derekagorhom which problem.

Patowhiz:
@derekagorhom the problem is probably in the subroutine that fills the grid with the contents of the data frame that contains the factors. You would have to debug it and check why it fails to do so.

rdstern:
@derekagorhom you have quite a lot of tasks now.  Is this one where you need help from @N-thony, or perhaps he could take this over?

rdstern (approved):
Perfecr.  @N-thony ready to merge!

PR #8782, opened 2024-02-01
Changes to the filter and select dialogue's

Fidel365:
Fixes partially #8694 , these are the changes i had implemented before, I'm working on the remaining part of the issue.


rdstern (commented):
@Fidel365 great to finish this.  I note that on the Save Filter option that was visible earlier you have still the option for the subset, which isn't needed.  So it isn't just the remaining items.  You also need to fix this option. 

Please note that the first option (Apply as Filter), saves the filter and then applies it.  The option you are adding just saves the filter and doesn't apply it!  So you have made life difficult for yourself, by using the secon (subset) option.  The idea is to make life simple for yourself - and also for a reviewer.

Fidel365:
Changes for this are in PR #8830, new issue for part(g) which is the only yet to be solved is #8861

Issue #8786, opened 2024-02-01
Adding Help IDs for New Dialogs - 4 (Climatic)


rdstern:
@Fidel365 this is (I hope) the last one for this task just now:

I assume the pull request #8785 has been merged.  This restores the Climatic > PICSA > General dialog.

- [x] a) `Climatic > PICSA > Temperature` currently has ID 118.  Please change to 479.
- [x] b) `Climatic > PICSA > General` currently also has ID 118.  Please change to 521
- [x] c) `Climatic > Describe > Trend Graph` is the same as Climatic > Describe > General.  So I assume that is automatically changed to 521.  If not, then please make the change here too.
- [x] d) `Climatic > Describe > IDF` add Help ID of 523
- [ ] 
- [x] e) `Climatic > Compare > Taylor Diagram`  Add ID of 643
- [x] f) `Climatic > Compare > Conditional Quantiles` Add ID of 642
- [x] g) `Climatic > Compare > Seasonal Plot`  Add ID of 640
- [x] h) `Climatic > Compare > Time Series Plot`  Add ID of 639
- [x] i)  `Climatic > Compare > Time Series Plot`  `Summary Annotations `Sub-dialog Add ID of 268
- [x] j) `Climatic > Compare > Summary`.  Add ID of 635
- [x] k) `Climatic > Compare > Calculate`  Make sure Help key is active.  Currently the button doesn't do anything.   Add ID 634
- [x] l) `Climatic > Check Data > Check Station locations` Add ID of 622
- [x] m) Climatic > Mapping > Map Make sure Help key is active.  Add ID of 645 
- [x] n) `Climatic > Models > Extremes` Add ID of 488
- [x] o) `Climatic > Models > Markov Modelling`  Add ID of 490
- [ ]

lilyclements:
Fixed by #8812

PR #8791, opened 2024-02-05
Adding help id's

Fidel365:
Fixes partially issue #8786.
@rdstern, some of the help id's you have provided are not working; that is for part (b), (c), (d) , the id's provided are not working but if itry to use different valid id's , they work just fine. Otherwise, the remaining part of the issue is ready for review


rdstern:
@Fidel365 it is understandable that some of the new IDs don't yet work.  They need the new help file first.

rdstern:
@Fidel365 I just checked #8786 so the help 4 issue.  I assume you did not change anything from 1, 2, and 3, yet.  Tell me when you do.
I checked the results and just found 2 that don't work.
In point i, the sub-dialog help doesn't work yet.  It should have help ID of 268.
And o went to the wrong place.  That may be my mistake, but could you check you put ID of 490 there.

Thanks so far.

rdstern:
@Fidel365 you have gone quiet again?  I assumed the updates for the 4 issues would be about one day of work together.  I wonder what has happened?  Could you please keep @N-thony updated when there are days when you are not abler to work?

Fidel365:
Pr #8799 fixes the issue

Issue #8730, opened 2024-01-12
Implement the Drop Unused Levels checkbox in the Column Summaries dialog 


rdstern:
This is the second of three improvements in the R-Instat Summary system.  It also awaits a check with @volloholic.
The Drop unused levels checkbox is in the existing dialog, but disabled.  I suggest it be implemented - as long as it doesn't "open a can of worms".   
If it does, then the implementation remains on our to-do list, but the disabled option should have the checkbox ticked, because it currently does remove the unused levels.
My possible concern is that the current system works very nicely with a filter in operation and I want that to remain.  I hope it isn't dropping levels as its way of accomodating the filter.
There are perhaps 2 separate situations on the levels.  
a) A single factor could haved unused levels.
b) Each factor could be ok, but there are some zero frequencies in the combinations.  For example, in the survey data, Variety has 3 levels and Village has 4 levels, but only 10 of the 12 combinations are non-zero.  Currently we only give the 10 combinations.

Note that one way of getting the unused levels in the summaries would be to proceed initially with the summaries as now.  Then get the cross factor (forcats) with all levels and then merge with the data.  The only complication I forsee is that the cross factor would have to allow for a filter being in place.

rdstern:
After discussion with @lilyclements she claims it should now be easy.  Let's see!

rdstern:
@N-thony I think we finally have a solution to this problem.  I hope you can make the small checges to the dialog (or allocate, but probably you, for efficiency?)  Then @lilyclements will separately make the partner change in the calculation system.  At least that's my understanding!

For you:
a) The Drop unused levels checkbox becomes checked by default.  (Currently it is unchecked.)
b) If the Summarise By... multiple receiver is empty or includes any non-factor terms, then the drop levels is checked, and also disabled. 
c) If the Summarise By... has just factor variables  then the drop levels control becomes enabled.  
d) @lilyclements could the summarise by work ok if there were logical variables as well as factor variables?  If so, then they could be allowed in the control?

PR #8794, opened 2024-02-07
Updating R code in calculations system

lilyclements:
This updates the R code in the calculation system. In particular, here we are moving from the now depreciated functions like `filter_` to `filter` in `dplyr`, which means that updates in the verbs in `dplyr` can be implemented in R-Instat.

@rdstern I set this as a draft for now. This is because I have **hard coded** in to have `.drop = FALSE` and `.preserve = TRUE` here. This means that we do **not drop unused levels** here as requested in #8730.
This is not ready to be reviewed or merged, but, if you wanted to see in action how the results would look if we did not drop unused levels in the column summaries dialog then you can try this out (the option on the dialog is still disabled, but hard coded in the R code -- not visual basic code). I've put a GIF below to illustrate -

Untitled video - Made with Clipchamp

I need to have a conversation with either @volloholic or @dannyparsons on how to add `drop` and `preserve` as parameter options in the summary system, to then set `.drop = drop` and `.preserve = preserve`.


rdstern:
@lilyclements that's great.  It works!  I know I shouldn't be surprised.
The only problem I found was that the summary variables are now Lists (given as LT in R-Instat), rather than numeric.
But I tried first to a new summary data frame, which was fine.
Then I got brave and added to an existing summary (yield by variety and village) which already had the 10 summaries without the zeros before - and it just added them!  Very nice.

lilyclements:
> The only problem I found was that the summary variables are now Lists (given as LT in R-Instat)

This is good to know. This PR is quite a big change in the calculation system, so I am not surprised that there are little bugs like this! I will do a much more comprehensive review of these changes to eliminate these sorts of bugs, but this is really useful to know. Thanks. I have now fixed this problem (I hope!)

lilyclements:
> Then I got brave and added to an existing summary (yield by variety and village) which already had the 10 summaries without the zeros before - and it just added them! Very nice.

@rdstern that is brave - I am _very_ relieved that that worked!

rdstern:
@lilyclements to me it works perfectly now!  Well done.

I am alsoi very happy that you are taking the opportunity to update the calculationj system.  Will this (does this) affect other dialogs?  For example will the general summaries be changed a bit, and start of the rains, etc?

lilyclements:
@rdstern yes. These are similar changes to what we want to do for when we manage to add the censoring into the Start of Rains dialogs. I believe that censoring requires another step after this, where we are facing issues. But this is a step in that direction.

lilyclements:
I have a meeting with @volloholic tomorrow, where we hopefully will have time to cover this. If not, I'll set up another meeting for early next week (of course he's now booked out for this week!)

lilyclements:
@rdstern I accidentally worked out why this wasn't working for us to get your censoring. We need to have year as a factor on the PR (it needs to be a factor because the `.drop` is with missing factor levels) to get your censoring (finally!):

E.g., with `ghana_two_stations` I created two new variables from the "Use Date" dialog: Year (factor) and doy
Then I filled out the start of rains dialog like so:

image

We get the censoring - finally!

1944 is NA due to a lack of data, 1945 is NA due to no start of rains fitting the given definition (100mm rainfall in 2 days)

image

rdstern:
@lilyclements that's great, really great - and also helpful in the end of the season where there were confusing years and we were not sure whether missing data, or no end to the season.
a) But I would prefer to have year as a numeric for the plotting later, if possible.  Could we leave it numeric in the define, and add as.factor in the R code for the start and end of the rains/season.  I can see it is behaving as a factor there.
b) And could we (I mean you of course) tweak the R code to give 3 values in the logical column?  So 1944 becomes NA, rather  than TRUE?  So missing in the data gives NA, leaving TRUE as a date, and FALSE being no ending?

lilyclements:
@rdstern 

a) This is less pretty than I want it to be!

Some ideas:

1. We could add it another argument in the calculation system when using `type = "by"`, and add it in as a parameter to `treat_as_factor` or something?
Then for year, we would do `drop = FALSE, treat_as_factor = TRUE` as arguments?
If `treat_as_factor` is true, then we run `group_by(factor(<variable>))` instead of `group_by(<variable>)`
But then we need to have a bit of a think about converting it back to numeric. This could be straightforward to read into the summary calculation - I don't know the calculation system well enough yet to know.

Alternatively, could we have a checkbox next to "Year" to "Convert to factor"? Not quite what you want, I know, but might be a way around this for now. 

b) Yes, we can do this by adding at the end of the function (after the `data_book$run_instat_calculation` line):

```
start_of_rains_status <- instat_calculation$new(type="calculation", function_exp="ifelse(start_rain_status == TRUE & is.na(start_rain), NA, start_rain_status)", result_name="start_rain_status", save=2, calculated_from=list("ghana_by_station_year"="start_rain", "ghana_by_station_year" = "start_rain_status"))
data_book$run_instat_calculation(calc=start_of_rains_status, display=TRUE)
```

rdstern:
@lilyclements I am happy that b) is true.
I am also delighted that we now have a solution - as you said, we have been waiting 5 years for that.  But I am not able to judge about this issue if year having to be a factor for this particular calculation, and this needing to be done explicitely.  

We used to have to define the year as a factor before, when making the climatic summaries, and this was in the general and climatic summaries dialog.  It made tutorial 2 much more complicated than it is now, so it was a great relief to me, when @dannyparsons declared it wasn't needed.  

It is only you, Danny and @volloholic who understand this key (and very powerful) component of R-Instat.  We need a short-term solution, and for this I suggest discussion with @volloholic is needed.  I would really like to make quick progress, on this issue.  Longer-term (only slightly longer I hope) is to document the system, and I suggest we include R-Instat documentation but also aim for a paper somewhere.  This is way overdue anyway.  Can we add @jkmusyoka into this.  His R skills are much better than mine, and it would be good it he understood the calculation system as well.  And it is particularly used for the climatic component, so maybe the paper (or papers?) should be for that audience?

rdstern:
@lilyclements things are moving on, in relation to the R code in R-Instat.  I also need to understand the implications of the 3 R-Instat R packages.  For the smaller point of the improved calculations I am hoping your hard-coding of .`drop = FALSE `and `.preserve = TRUE `here. This means that we do not drop unused levels here as requested in https://github.com/IDEMSInternational/R-Instat/issues/8730. and means that you could complete pull request #8986, or instruct @N-thony on what need be done?

lilyclements:
@rdstern if we're happy for this to be hard coded in for now, then this is ready for review!

rdstern (commented):
@lilyclements this looks fine:
image

Now how can this be inserted into @N-thony #8986 and also not be hard-wired.  Is this for Antoine to do, under your instruction, or will you do it?

lilyclements:
@rdstern

lilyclements:
@rdstern Did we want it hard-wired in for now, or, do we want to wait until I can make the changes?

rdstern:
@lilyclements wait till the changes

rdstern:
@lilyclements is this something you can now add into #8986 that @N-thony has prepared.  Or is it a job for someone else?

lilyclements:
@rdstern it's currently hard-wired in in this version, so I need to set it to be a parameter and amend the calculations system to do that

rdstern:
@lilyclements is that something for you alone, or do you need to work with @volloholic on it?  If so, I note you are both in Kenya together?

lilyclements:
@rdstern I have now done this! (I hope, I have been testing my end).
**Are we happy for `drop = FALSE` to be the default? Can make it that `drop = TRUE` is the default instead.**

@N-thony this means this can be merged into your branch 

**In Start of Rains: Changes to Make to Implement this:**
To now get the `drop = TRUE` option in the Start of Rains dialog, we need to add a new parameter to the `data_book$run_instat_calculation` function:

```
data_book$run_instat_calculation(calc=start_of_rains_combined, display=TRUE, param_list = list(drop = TRUE))
```

To get the `drop = FALSE` option, we can either make no change (i.e., this is the default) or run:

```
data_book$run_instat_calculation(calc=start_of_rains_combined, display=TRUE, param_list = list(drop = FALSE))
```


**In Column Summaries: Changes to Make to Implement this:**
We just use the `drop` parameter here. Set `drop` to be `TRUE` to drop, and `FALSE` otherwise. This is for the `"calculate_summary"` function. So this should be straightforward to make the amendments in the VB code.

Let me know elsewhere we want this, so I can amend the functions as necessary.

rdstern:
@lilyclements I just want to check with drop = FALSE what happens when there is a filter in operation?  In the survey data if I filter, on village, to Kesen and Niko, then with drop=TRUE I get 4 rows.  
What do I get with drop=FALSE?  Do I get 6 rows, because there are only 2 villages (after the filter) or do I always get 12 rows, because Village has 4 levels and Variety has 3 levels.

I don't want anything to change in your code.  But if you get 12 rows, then I suggest we have drop = TRUE as the default.

lilyclements:
> @lilyclements I just want to check with drop = FALSE what happens when there is a filter in operation? In the survey data if I filter, on village, to Kesen and Niko, then with drop=TRUE I get 4 rows. What do I get with drop=FALSE? Do I get 6 rows, because there are only 2 villages (after the filter) or do I always get 12 rows, because Village has 4 levels and Variety has 3 levels.
> 
> I don't want anything to change in your code. But if you get 12 rows, then I suggest we have drop = TRUE as the default.

@rdstern so filter has it's own option `.preserve` for this situation.

The answer is different in R depending if you do `group_by`, then `filter`, or if you do `filter` then `group_by`.
In R-Instat, we would do `filter` then `group_by` in something like the column summaries dialog (different for dialogs like SoR where we write out a whole function).

So we have these two ideas: `.preserve` and `.drop`:

1. **`.drop = TRUE/FALSE` in `group_by()`**:
   - `.drop = TRUE`: This is the default setting. It drops groups that do not have any data after filtering. For example, if you group by `village` and then filter the data, only the groups that still have data will be kept.
   - `.drop = FALSE`: This keeps all groups, even if some groups have no data after filtering.

2. **`.preserve = TRUE/FALSE` in `filter()`**:
   - `.preserve = TRUE`: This retains the original grouping structure even if some groups have no rows after filtering.
   - `.preserve = FALSE`: This is the default setting and allows the filtering operation to potentially change the grouping structure by dropping groups that no longer have data.

### When **Filter is Applied First**:
- If you filter first and then apply `group_by()`, the `.preserve` option in `filter()` does not affect the outcome of the grouping. 
- The `.drop` option in `group_by()` will determine whether groups with no data (after filtering) are kept or dropped.
- This is usually the case that we have: we would usually filter and then group_by (like in the column summaries dialog). So, for us, **we would get all 12 columns even if we filtered previously provided that `drop = FALSE`.**

```r
# These give the same result because filtering is done first
survey %>%
  dplyr::filter(village %in% c("NANDA", "KESEN"), .preserve = TRUE) %>%
  dplyr::group_by(village, .drop = FALSE) %>%
  dplyr::summarise(n())

survey %>%
  dplyr::filter(village %in% c("NANDA", "KESEN"), .preserve = FALSE) %>%
  dplyr::group_by(village, .drop = FALSE) %>%
  dplyr::summarise(n())
```

In both cases, `.preserve` does not affect the final result because grouping is applied after filtering.

### When **Group_by is Applied First**:
- If you group the data first and then apply `filter()`, the `.preserve` option in `filter()` *does* influence whether groups with no data (after filtering) are kept or dropped. This is where `.drop` and `.preserve` interact.
- This is relevant in our R-Instat code for bits like the SOR code.

```r
# Here, .preserve matters because the data is grouped first
survey %>%
  dplyr::group_by(village, variety, .drop = FALSE) %>%
  dplyr::filter(village %in% c("NANDA", "KESEN"), .preserve = TRUE) %>%
  dplyr::summarise(n())

survey %>%
  dplyr::group_by(village, variety, .drop = FALSE) %>%
  dplyr::filter(village %in% c("NANDA", "KESEN"), .preserve = FALSE) %>%
  dplyr::summarise(n())
```
**In this scenario, only if `.preserve = TRUE`, then groups that have no data after filtering are still kept.**

### Overall:
- **Filter then Group_by**: Only `.drop` in `group_by()` determines if empty groups are retained.
- **Group_by then Filter**: `.preserve` in `filter()` and `.drop` in `group_by()` together determine whether groups without data are kept.

Does this make sense or have I gone off too far?

lilyclements:
@N-thony if you’re happy with it, then yes. The code changes are all done on my end (provided no one spots a bug!)

PR #8796, opened 2024-02-08
Moved the Percentage sub-dialog into the Summaries sub-dialog in Prepare > Data Reshape > Column Summaries

### LINKED ISSUE #8729: Move the Percentage sub-dialog into the Summaries sub-dialog in Prepare > Data Reshape > Column Summaries
The calculation system in R-Instat is one of our "selling points".  It needs a presentation (including video) and to be well reported in the help.  As preparation this is the first of 3 urgent(?) issues we need to undertake.

I need an initial discussion with @volloholic before this work.  I'll report once that's done.

This one is to get rid of the Propn and Percentages button and sub-dialog from this dialog.  Instead it adds a tab to the summaries sub-dialog. 

image

Instead - and I hope this will be easy to do - it is just called `Percentages`, and becomes a further tab on the Summaries sub-dialog.

I think this (with small changes) just needs to be transferred (and corrected) to become the last tab in the existing Summaries sub-dialog. 

Here it is: 

image

a) We still need the checkbox at the top, simplify the label to become `Calculate Percentages`.
b) Then, either we implement the other radio buttons at the top, or let's get rid of them.
c) Check, notice this is a different data frame here to the main dialog.  I think this should not be allowed, so the dataframe part of the data selector is disabled.
d) Correct the R code.  I think it works ok when factors are included, but it should also work when they are omitted?  I assume, with this example it should add the percentage of values for each variety.  Currently it doesn't.  (This may need a quick input from @lilyclements, but I suggest that could come after the transfer has been completed.

N-thony:
Fixes #8729 
@rdstern have a look.

### DIFF
--- a/instat/dlgColumnStats.Designer.vb
+++ b/instat/dlgColumnStats.Designer.vb
@@ -48,7 +48,6 @@ Partial Class dlgColumnStats
         Me.ucrChkPrintOutput = New instat.ucrCheck()
         Me.ucrChkDropUnusedLevels = New instat.ucrCheck()
         Me.cmdSummaries = New System.Windows.Forms.Button()
-        Me.cmdProportionsPercentages = New System.Windows.Forms.Button()
         Me.ucrChkWeights = New instat.ucrCheck()
         Me.ucrReceiverWeights = New instat.ucrReceiverSingle()
         Me.ucrReceiverByFactor = New instat.ucrReceiverMultiple()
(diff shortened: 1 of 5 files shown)


rdstern (commented):
@N-thony excellent.  And it all seems to work fine and is a great addition. 
The only thing I found was that it doesn't clear the percentage checkboxes when I press reset.
@lilyclements I think you will enjoy this.  Could you test and approve.  I assume your approval can be used to merge.  If not, we should ask Chris to change the settings.  

Lily, in your checking I am not yet checking whether we could improve the way the percents are being given.  I'm just checking that it all works as it did when they were in separate buttons.
I first tested with the survey, yield by variety with counts and mean, and adding percents - without the button for then sub-sub-dialog.  So that's a one-way table.
Then yield by variety and village.  Again without the sub-sub-dialog to give cell percents.  Then with the sub and putting vfillage there.

rdstern (commented):
@N-thony one more small thing.
If I use the survey and get a set of summaries, anhd then add the percentage checkbox, but ignore the button, then it doesn't produce the percentages.  I need to press the button, and then return - without doing anything perhaps.  Then it works.  For overall percentages you don't need the button, so it would be perfect if it gave them once the checkbox is ticked.

rdstern (approved):
@N-thony that's really great.  @Patowhiz this is a small but important change.  Could you please check and merge?

Patowhiz (approved):

PR #8799, opened 2024-02-09
Adding help ID's

Fidel365:
Fixes issue #8786 .
@rdstern, this is ready for review except for (o).New Markov, the id is invalid and takes us to the wrong place.

### DIFF
--- a/instat/dlgClimaticStationMaps.vb
+++ b/instat/dlgClimaticStationMaps.vb
@@ -64,7 +64,7 @@ Public Class dlgClimaticStationMaps
     End Sub
 
     Private Sub InitialiseDialog()
-        ucrBase.iHelpTopicID = 558
+        ucrBase.iHelpTopicID = 645
         ucrBase.clsRsyntax.iCallType = 3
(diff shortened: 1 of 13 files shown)


rdstern:
@Fidel365 many thanks for the update.  I wonder if the Marlkov modelling ID is currently 491?  If so, it should be changed to 490.  
I'll check later today.

rdstern:
@Fidel365 I am not aboe to run your branch.  It says you have a syntax error.  When fixed, could you also correct the Markov Chain value, and tell me which issues it resolve?  Is it just the 4th, or for more of the help IDs?

Fidel365:
@rdstern I wonder where this errors are coming from on your side, I will get someone else to test and if it still doesn't work, I will just create another PR, this one fixes the issue except for new Markov that I think(having tested it) both 490 and 491 are invalid IDs for it.

rdstern:
@Fidel365 could you get someone else to check:  I still get:

image

which I haven't seen before!

Fidel365:
Let's work from PR #8812

Issue #8728, opened 2024-01-11
Correct the labels on the Prepare > Data Reshape > General Summaries dialog.


rdstern:
@N-thony this should be quick for anyone to do.

Something has gone wrong with these controls:

![image](https://github.com/africanmathsinitiative/R-Instat/assets/11226469/8f592c00-acc6-4e5a-8bfd-c7f6ff86b1b9)

I assume they just need to be moved to the right for each of the buttons to become visible?  And, as on the main dialog they are Add, Edit, Delete, Duplicate?

In the sub-calculations that's what they are:

![image](https://github.com/africanmathsinitiative/R-Instat/assets/11226469/18d9e1f1-e375-428a-a753-33006c86a522)

So in the Manipulations the up-down arrows seem to have messed that up?

lilyclements:
Fixed by #8801

Issue #8744, opened 2024-01-20
Prepare > Data Reshape > General Summaries needs bugs fixing


rdstern:
This is an incredibly powerful dialog.  Once fixed it will still be a pain to use, but we could use it to calculate the complicated summaries that we use, for example, for the start of the rains.  So it really shows the power and potential of R-Instat.  I found the bugs when trying to write the help for the dialog.  I mention 2 (of the?) bugs below.  I think if those are fixed it should be easy to proceed and for the team to do anything more - if more is needed.
@ChrisMarsh82 David has (I am afraid) proposed you to fix the obvious bugs, given below.  
@volloholic is hoping you could have a quick look and either fix them - he thinks that should be easy for you.  Or you then allocate.

The example shown here is trivial, but demonstrates that the system does work.  It uses the ggplot2 diamonds data and the caclulator and gives the mean price for each level of the color factor. 

Here is the output it produced:
![image](https://github.com/africanmathsinitiative/R-Instat/assets/11226469/572b6407-02a0-498f-8c7c-7004f17587d0)

And here is the (corrected) script, which produced it.

```
# Dialog: General Summaries

calc1 <- instat_calculation$new(name="calc1", type="summary", save=2, result_name="bob", manipulations=list(manip1=instat_calculation$new(name="manip1", type="by", save=0, result_name="", calculated_from=list(diamonds = "color"))), function_exp="mean(price , na.rm = TRUE)")
data_book$run_instat_calculation(calc=calc1, display=TRUE)

rm(calc1)
```
Now this was corrected by David.  The script directly from the dialog was as follows:

```
# Dialog: General Summaries

calc1 <- instat_calculation$new(name="calc1", type="summary", save=1, result_name="", manipulations=list(manip1=instat_calculation$new(name="manip1", type="by", save=0, result_name="", calculated_from=list(diamonds = "color",diamonds = ""))), function_exp="mean(price , na.rm = TRUE)")
data_book$run_instat_calculation(calc=calc1, display=TRUE)

rm(calc1)
```
So, he made 2 changes.  The first was to add a result name on the first line.  The second was to delete the `,diamonds-""` on line 3 

Now the 2 or 3 bugs so far.  Go into the dialog and click on Add.  You then get this:
![image](https://github.com/africanmathsinitiative/R-Instat/assets/11226469/9f8f6805-09e5-467f-90bc-1a403eca4a59)

It doesn't find the diamonds data.  That's when the type is calculation or summary.
[...]

PR #8801, opened 2024-02-12
small changes to the general summaries dialogue

Fidel365:
Fixes issue #8728 
@rdstern , this is  ready for review


rdstern (approved):
@Fidel365 I am really happy you have done this already.  @N-thony this is the same dialog where there is also a major problem #8744, which you also mentioned today, that you will investigate.  It may be that Fidel could have the first look?  Up to you both.
Meantime I am approving in case you wish to proceed step-by-step?

Issue #8790, opened 2024-02-04
Improvements in the Describe > Graphics > Heatmap dialog


rdstern:
Here is the current dialog:
The greenrice data is in the agritutorial package.

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/d5da30c1-c97a-4a1f-ad6f-bda2929e60a8)

(One change, specified separately, is to add the legend, and facet control to the dialog.  Probably wait until that has been merged.)
a) At the top the word Chorolopleth is wrong.  It should be spelled `Choropleth`  
b) The top control should chenge to become just a Single receiver.  That will enable the controls on the right to move up a bit, which is good.
c) The Add points control is wrong.  It should go to the left, and use jittered points and not include a new variable.  The control could be a checkbox, default Unchecked.  Label Add Jitter.  If checked it shows the other two controls (copied from the Scatterplot dialog, see below.)  And it would add the same code.

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/7f96456f-24c9-4d11-baa4-1637e0810cc6)

d) The options for colours should become much richer and perhaps give a button to go to the sub-dialog?  But I wonder if we give the gradient options here and that would be enough?  This needs some thought first.

I am working towards this sort of result to become easy from the main dialog:

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/4a992cde-dc6a-4266-aaa0-4eff9eb3c096)

Issue #8768, opened 2024-01-26
Translation problems on layout and words Prepare > Data Reshape > Column Summary


rdstern:
@jkmusyoka and @MeSophie 

In this dialog here is the More tab on the sub-dialog in French and Russian:

MeSophie:
> @jkmusyoka and @MeSophie
> 
> In this dialog here is the More tab on the sub-dialog in French and Russian:

@rdstern Please could you specify the place where the issue is? After investigate on the translation issues I notice some issues but they are already repport on #8767. 
Also I notice that the buttons on `proportion and percentage` button are not translate. I don't know if you're talking about this problem here?
![image](https://github.com/IDEMSInternational/R-Instat/assets/106779233/a59f7e89-2b02-46b8-9d81-c6bc6a111dd8)

There is also this `Options` button where the translation of the last sentence has been cut.
![image](https://github.com/IDEMSInternational/R-Instat/assets/106779233/3377dab4-f058-424a-ad19-e259b8aaedef)

Issue #8767, opened 2024-01-26
Translation problem on some dialogs on returning to English


rdstern:
@lloyddewit NS @N-thony 

Here is the Prepare > Data Reshape > Column Summaries dialog:

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/46fce40e-aca0-4261-afab-5e239118be18)

I was looking for possible problems with spacing in other languages, so went to French, and then Russian and then back to English

Here is the main dialog:

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/af98c029-30ff-47d6-94fc-14339ff8860a)

There are just 2 items that are not translated back into English.  Don't worry about the Pourcentage button, because that will be changed soon.  This is also a feature I keep seeing and I can look for other instances.  But is there something about those senetences that causes the problem, so there could be a general solution?  I don't know whether it is with the translation mechanism, or the sentence - e.g. does it start with a space?

MeSophie:
@rdstern I notice also a layers translation problem on the summaries sub dialog

![image](https://github.com/IDEMSInternational/R-Instat/assets/106779233/d3896eb8-cdbc-4197-9e0e-eb1115e261f8)

PR #8803, opened 2024-02-12
Small Text Change in Data Base.

MeSophie:
Fixes Partially #8767 
Fixes Partially #8768
Fixes partially #8790
@rdstern , @N-thony I made small change on data base to correct the text `Choropleth Map`. When PR #8802 will be merge the text will be okay.

image

@rdstern  I also solve some translation issue on Prepare>Reshape Date>Column Summaries dialog and Summaries sub dialog (`Print Resuts to Output Window` box in the main dialog and `proprotions/percentage/count` on summaries sub dialog).

image
image

@lloyddewit On the summaries sub dialog there is still some translation issue in More tab.  
When translating the dialog in french and click on `Lan` `Percentile and Kurtosis` check boxes are still in french. 

I checked on json file crowdin data base controls and data base translation as I did previously  but I can see where the problem is. I also check try to solve the issue with Antoine but we but didn't find where the issue is.
image
For `percentile` I also try to add text in the data base translation but it is not working. So please could you  have a look?


lloyddewit:
> On the summaries sub dialog there is still some translation issue in More tab. When translating the dialog in french and click on Lan Percentile and Kurtosis check boxes are still in french.

@MeSophie 
For `Kurtosis` the `form_controls` table looks correct:

image

However the `translations` table looks strange:

image

Translating from French to English, the `translations` table has 2 possible translations for `Kurtose` (`Kurtose` and `Kurtosis`). 
`Kurtose` does not appear in the `form_controls` table so therefore it should also not appear in the first column of the `translations` table. If you remove all `translation` table rows that have `Kurtose` in the first column, then it will fix the problem.
The real question is why does the `translation` table contain these incorrect rows? I suggest that you investigate the JSON files and CrowdIn to see if there is a problem. The next time that @N-thony or yourself rebuilds the translation database, then you could also check to see if this problem has returned.

You also mentioned there was a problem with `Pourcentage`. I don't have time to investigate now but it is likely to be a similar problem. Please could you investigate?
Thanks

MeSophie:
@lloyddewit Thank you I did further investigations and was able to solve the problem with percentages and Kurtosis.
For `Kurtosis` as you said I delete Kurtose in the data base.
image

For `Percentage` group box I added Controls in the data base and also change P of percentages to Upper-case.
image
image

But for `Percentile` check box in Prepare>Reshape Data>Column Summaries>Summaries>More the problem is still there. We already have the text in the data base translation, the control in the data base controls
image
image

lloyddewit:
@MeSophie Thank you for the feedback.
I looked at the Prepare>Reshape Data>Column Summaries>Summaries>More dialog in the main branch and I saw a number of translation issues but not the one you mentioned. I'm not officially working on R-Instat so I have limited time to spend. If you can fix what you can and merge the PR, then I can look again at the main branch and see if I can identify the problem.
Thanks for your understanding

Issue #8760, opened 2024-01-24
Corrections and improvements to the plot geom sub-dialog


rdstern:
@N-thony I suggest this should be @Vitalis95 .  After he has made the corrections I will be interested if he can diagnose a later task - not urgent - for him?

As I use the sub-dialog I keep finding minor omissions, errors and inconveniences.  Here I start with boxplots and maybe add more.  They should all be easy corrections.  Sometimes making the same correction for other geoms may also be possible.

This shows the boxplot parameters:
<img width="368" alt="image" src="https://github.com/IDEMSInternational/R-Instat/assets/11226469/07106663-37cc-4635-be1d-97cc9307943b">

a) As shown NA is given with quotes.  Please could they be deleted, just from the NA.

b) Changes needed on both the stat_cor and stat_regline_equation.  Same changes on each:
![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/fad1cbe6-6add-424a-b9e7-a08bef90be4f)

The corrections are simple for Label.X and Label.Y.  The existing option, "NULL" should be replaced by the simple control into which you can type a number.  It can't be an up-down and it can be any number - positive or negative, because it depends on the values in the graph.

It is more complicated for the npc (sort of percent controls), that are just above.  That can either be a number between 0 and 1, or the categories you have there.  I suggest we add 2 radio buttons on each line.  The first has no label, but has the up-down control from 0 to 1 in steps of 0.01.  Then there is the second radio button, again I don't think a label is needed.  But it is followed by the existing control.

c) Perhaps more to come!

lilyclements:
Fixed by #8804

PR #8804, opened 2024-02-12
Iimprovements to the plot geom sub-dialog

Vitalis95:
Fixes Partially #8760 
@rdstern this fixes **a)** and **b)** on corrections  for `Label.X `and` Label.Y` 
@N-thony  on the part of ` Label.X.npc`  and `Label.Y.npc`  the  `radio button`  for linking them is not defined in `ucrGeoms`. I have sent you a message about this on skype


rdstern (commented):
@Vitalis95 that seems to work.  For the remaining bit, would it make the change simple to do if you simply changed the control for the label.x.npc options.  So, just a single receiver that can take a number - any number including negative and entrered, rather than an up-down. m Maybe you could check in the documentation, because I suggest that's what is usually needed?

rdstern:
@Vitalis95 if you are right, that they are 0 to 1 then an up-down would be fine.  But I thought they were tied to the actual x and y scales, so could be anything?  I hope I am wrong, because that might get tricky if one axis is a date!

Vitalis95:
@rdstern  I think it is  preferable for `label.x.npc` and` label.y.npc `to stay as characters because it specify the  positions  top, middle, and center..., rather than being numeric

Vitalis95:
@rdstern , have a look at now

rdstern (approved):
@Vitalis95 great, that's working now.  @N-thony I am approving.  (I note that if I type `roger `there instead of a number I get an error of course.  But I am becoming quite happy with that error, because if I go to a script and run `roger <- 35`, then it is quite happy!)

rdstern:
@N-thony is this ok to merge?

PR #8805, opened 2024-02-13
Disabled horizontal scrollbar in multiple receiver

### LINKED ISSUE #8797: Get rid of the sideways scroll in the multiple receiver
@N-thony I hope this is a 5-minute job.  Good if you could do it, because that receiver is in so many dialogs.

Here (as an example) is the summarise dialog:

image

It was not there before and I now understand why it has appeared.  It is the extra at the top that notes the number of items in the receiver.  Could you make an edit, so we keep that feature but don't have the horizontal scroll at the bottom.

There is just one complication, that sometimes we have a long sheet name.  Here is an example:

image

In this case, without the horizontal scroll bar we don't see the number of variables in the reciever.  I woulod be ok with that.  It is a price the user pays for having a long sheet name.  But if that is difficult (or you would prefer that feature to be available) then I could accept the horizontal when it isw needed.

N-thony:
Fixes #8797
@rdstern have a look

### DIFF
--- a/instat/ucrReceiverMultiple.vb
+++ b/instat/ucrReceiverMultiple.vb
@@ -167,18 +167,19 @@ Public Class ucrReceiverMultiple
         'it's not clear when the receiver will ever have more than one data frame
 
         'reset the header text with the name
-        lstSelectedVariables.Groups(0).Header = ShortenString(lstSelectedVariables.Groups(0).Name, 10)
+        lstSelectedVariables.Groups(0).Header = ShortenString(lstSelectedVariables.Groups(0).Name)
         If lstSelectedVariables.Groups.Count = 1 AndAlso lstSelectedVariables.Items.Count > 0 Then
             lstSelectedVariables.Groups(0).Header = lstSelectedVariables.Groups(0).Header & " (" & lstSelectedVariables.Items.Count & ")"
         End If
     End Sub
 
-    Private Function ShortenString(strText As String, maxLength As Integer) As String
+    Private Function ShortenString(strText As String) As String
+        Dim maxLength As Integer = 10
         If strText.Length <= maxLength Then
             Return strText
         Else
             ' Trim the string to the specified length and add ellipsis
-            Return strText.Substring(0, maxLength - 3) & "..."
+            Return strText.Substring(0, maxLength) & "..."
         End If
     End Function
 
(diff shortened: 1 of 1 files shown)


rdstern (dismissed):
@N-thony great.  @Patowhiz could you please check, and merge if ok?

instat/ucrReceiverMultiple.vb
```diff
@@ -168,12 +167,21 @@ Public Class ucrReceiverMultiple
         'it's not clear when the receiver will ever have more than one data frame
 
         'reset the header text with the name
-        lstSelectedVariables.Groups(0).Header = lstSelectedVariables.Groups(0).Name
+        lstSelectedVariables.Groups(0).Header = ShortenString(lstSelectedVariables.Groups(0).Name, 10)
```
Patowhiz, line 170: If the `maxLength` parameter of the `ShortenString()` is hard coded as 10, why not remove it as a parameter and use the 10 inside the function?
 
 When I look at code line 181, it's not clear what will happen when the value `maxLength` parameter is less than 3

Patowhiz (changes requested):
@N-thony looks good. Just a few comments and code suggestion.

instat/ucrReceiverMultiple.vb
```diff
@@ -168,12 +167,22 @@ Public Class ucrReceiverMultiple
...
 
+    Private Function ShortenString(strText As String) As String
+        Dim maxLength As Integer = 10
+        If strText.Length <= maxLength Then
+            Return strText
+        Else
+            ' Trim the string to the specified length and add ellipsis
+            Return strText.Substring(0, maxLength) & "..."
+        End If
+    End Function
```
Patowhiz, line 184:
```suggestion
    Private Function ShortenString(strText As String) As String
    Const maxLength As Integer = 10
    If strText.Length > maxLength Then
        Return strText.Substring(0, maxLength) & "..."
    End If
    Return strText
End Function
```

Patowhiz (commented):
@N-thony I have made some suggestion on the `ShortenString()` function. Have a look and let me know if the suggested changes are more clear and optmised.

Patowhiz (approved):

rdstern (approved):
@N-thony and @Patowhiz I am not sure exactly what has changed, but it looks better to me now. Very happy for it to be merged, because the horizontal scroll bar was both annoying and looked un-professional.

PR #8806, opened 2024-02-13
Added more commands in the Script > Insert dialog Commands Tab 

### LINKED ISSUE #8775: A simple script leading to a new command for the Insert > Command dialog
Here is the script from the Circular package for the wind data.  It produces ordinary graphs, i.e. not ggplot graphs:

```
### Name: wind
### Title: Col De La Roa wind direction
### Aliases: wind
### Keywords: datasets

### ** Examples

  data(wind)
  wind <- circular(wind, template='geographics')
  par(mfcol=c(1,2))
  plot(wind)
  plot(density(wind, bw=40), main='')
```

I have added commands to make it more useful in R-Instat.  In particular it now produces ggplot graphs.

```
### Name: wind
### Title: Col De La Roa wind direction
### Aliases: wind
### Keywords: datasets
# Load library "ggplotify"
library(package="ggplotify")


### ** Examples
# Load library "circular"
library(package="circular")

  data(wind)
  wind <- circular(wind, template='geographics')
  par(mfcol=c(1,2))
  plot1 <- as.ggplot(~plot(wind))
  plot1
  
  # Load library "ggplotify"
library(package="gg  plotify")

plot2 <- as.ggplot(~plot(density(wind, bw=40), main='')) 
plot2
```

This runs fine.  It is one of many example scripts that produce ordinary plots.  In R-Instat we usually manage ggplots.  So I have added the command to change an ordinary plot into a ggplot.  This would be a good line to add into the `Insert button, Commands tab`, shown here:

image

I suggest the second radio button (rm) is moved down, so it remains the bottom one.  Add two extra buttons, of which this is the second.  It can be labelled as `ggplotify` - which is quite expressive!

If chosen, then it gives a label `Graph object`, into which the user types a name, say `plot1`

That's the same control as here:

image

When the name is typed the results window has 2 lines:

#Make graph a ggplot
<name, so here plot1> <- as.ggplot(~   )

Ideally the cursor is after the squiggle.  Then the user copies, or types, the graph code here.  

It would be good to do the other addition to the Command tab at the same time.  This is described in issue #8696. 

@N-thony I don't know if this is one you could do quickly - or allocate?

Thanks

Roger

Vitalis95:
Fixes #8775
@rdstern , @N-thony , Added` file.choose()`,` readRDS` and  `as.ggplot( )` commands
Adding cursor  after the squiggle in  `as.ggplot( )` is a bit tricky  because the `PreviewScript` control runs more other commands. We can consider this as a separate issue


rdstern:
@Vitalis95 this is coming on.  Let me start with the ggplotify example:  Here is the current dialog:  I got it wrong.  We need 2 controls and don't need to fiddle in the generated command code.

image

Add a control with label Graph Command.  (e.g. plot(wind)
Then the Graph Object is just below it.  
So you then copy the graph command into the brackets on the command and add the squiggle.
Also add ggplotify:: into the command.

So this would generate

  ```
# Make Graph a ggplot 
graph1 <- ggplotify::as.ggplot(~plot(wind))
graph1

```

Second is the View command.  That's above the plotify

**Review by rdstern — COMMENTED** [review]
@Vitalis95 iyt is almost there:
image

a) A trivial point, but the new lines are indented 2 spaces.
b) You have added a further plot, where I was copying the whole command - including the plot.  So you just need to add the squiggle.
c) And could you add the further line to show the plot.

Thanks

**Comment by rdstern** [comment]
@Vitalis95 I would like another option added.  Underneath Get Package add List Data.  
a) This has first the package, just the same as the Get Package control.  
b) Then after, and hopefully on the same line, is the control where you can type a name.  (That's the same as the Save Data tab, Data Frame control.) The label is `Into (Optional)`

The comment can just be `#List of data sets`
The command generated is `vcdExtra::datasets("package_name")`
If you add data as the optional name, then it also adds the same line as you would get with the Save Data > Dataframe, so:
```
# Save data frame "data"
data_book$import_data(data_tables=list(data=data))
```
So it generates the 4 lines.

Vitalis95:
@rdstern , have a look at it

rdstern (commented):
@Vitalis95 that is excellent - and almost works already!
a) Let's change the Into (Optional) into a `checkbox` with label `Into `.  Default is unchecked.  If checked, then it shows the field as now, to type the data-frame name.
b) When that is checked and used, then the first generated command needs to assign the datasets - so add this at the front.  It becomes:
data <- vcdExtra("dslabs") 

And great you are doing this so soon.  @rachelkg is ready to use it in a new script/video!

Vitalis95:
@rdstern , have a look at it

rdstern (approved):
@Vitalis95 Thanks and I am approving. @N-thony over to you.

Issue #8779, opened 2024-02-01
Adding help IDs for new dialogs - 1


rdstern:
@Fidel365 here they are.  @Vitalis95 can help if you need it.

a) Climatic > File > Export to Climsoft - 476
b) Climatic > File > Export to Climpact - 506
c) Prepare > Column Define > Circular - 520
d) The 3 extra keyboards for the calculator.  The current calculator goes to each of the 9 keyboards, when that is active.
There is now also a keyboard for List - ID is 439
and Complex - ID is 455
and Integer - ID is 463

I am stopping here, so this makes a reasonable pull request.  There is more to come!

lilyclements:
Fixed by #8809

PR #8809, opened 2024-02-14
adding help id's-1

Fidel365:
Fixes issue #8779    , @rdstern , the help file you sent may be the reason we are having conflicts, so I was able to test all except one because I do not have the recent help file log(i will add it once I'm done with these issues). So all work except for part (b), the id takes us to the wrong place and and part (a) I added it but i was not able to test it from my side.

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -198,10 +198,13 @@ Public Class dlgCalculator
                 ucrBase.iHelpTopicID = 598
             Case "Integer"
                 Me.Width = iBasicWidth * 1.5
+                ucrBase.iHelpTopicID = 463
             Case "Complex"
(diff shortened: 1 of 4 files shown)


rdstern (approved):
@N-thony I am approving as you found nothing wrong with the code.
I was still unable to test, because I (and I think @Fidel365 ) get an error when trying.  There are 2 possibilies I see.  The first is that this is fixing the extra keyboards in the calculator, which didn't exist earlier (in the help file).  The ordinary help key is clever here and goes to the right keyboard help.  Maybe there is a problem here.
The other point is that this help also goes to new dialogs that were not present earlier.

I am approving and will test as soon as it is merged.  That means (I hope) that you could easily revert, if there is an unforseen problem.

PR #8812, opened 2024-02-14
Adding help ids-4

Fidel365:
@rdstern, this fixes issue #8786 , let's work from here now. I still don't have the id for the New Markov, but you can test these ones first.


rdstern (approved):
@Fidel365 this now works.  I assume you will close the old one.  I assume the help number is 490 for the only one that doesn't work (Markov Chains)?  I hope it may "spring into life" automratically, when we include the new help file.  (I found an oddity with that entry in the help file, which I have chenged.  So I am optimistic.
@N-thony could you please check the code and merge if ok?

Fidel365:
PR #8812 fixes the issue.

PR #8842, opened 2024-02-26
Improvement in scatterplot, when x is a factor

### LINKED ISSUE #8836: Small improvement in scatterplot, when x is a factor
@N-thony this should be quite easy for anyone.  I show with the survey data.  

image

This works fine if fert is numeric. It gives the scatterplot with the fitted line.  Now make fert variable into a factor and the line doesn't come.

The correction needed is that if x is a factor then add a `group aeshetic to the geom_smooth` as so:

`last_graph <- ggplot2::ggplot(data=survey, mapping=ggplot2::aes(y=yield, x=fert)) + ggplot2::geom_point() + ggplot2::geom_rug() + ggplot2::geom_smooth(aes(group=1),method="lm", se=TRUE) + theme_grey()
`
That's if there is no factor in the `By (Optional)` control.  If x is a factor and (say) Variety is in the by, then the command becomes:

`last_graph <- ggplot2::ggplot(data=survey, mapping=ggplot2::aes(colour=variety, y=yield, x=fert)) + ggplot2::geom_point() + ggplot2::geom_rug() + ggplot2::geom_smooth(aes(group=variety),method="lm", se=TRUE) + theme_grey()`

Vitalis95:
Fixes #8836 
@rdstern @N-thony , ready for review

### DIFF
--- a/instat/dlgScatterPlot.vb
+++ b/instat/dlgScatterPlot.vb
@@ -20,6 +20,8 @@ Public Class dlgScatterPlot
     Private clsRScatterGeomFunction, clsLabelFunction As New RFunction
     Private clsRaesFunction As New RFunction
     Private clsLocalRaesFunction As New RFunction
+    Private clsGroupAesFuction As New RFunction
+    Private clsGroupAesVarFuction As New RFunction
     Private clsBaseOperator As New ROperator
     Private bFirstLoad As Boolean = True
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@Vitalis95 very well done.  You are getting good at this.  @N-thony can you please check and merge.

PR #8846, opened 2024-02-28
Added New Climagraph Dialogue On Climatic> Describe Menu

### LINKED ISSUE #8657: Add climographs to the climatic menu
@jkmusyoka  Here is the Climatic > Describe menu.  It is very poor.

image

If you ask for climatic graphs it suggests bar charts and climographs are popular.  The recent dodoma summaries provide data with just monthly summaries of rainfall and temperatures.  I was not able to get a climaograph from the Describe > Graphs > General dialog - perhaps you can?

In addition could we have various attractive climographs perhaps using R code you suggest, but I would like to see what chatGPT could come up with?  I assume that will be built on code and hope it could suggest several different alternatives?

Then could we quickly construct an impressive dialog that implements these alternatives.

The basic structure is clear.

a) Usual data selector
b) Single receiver with label Rainfall.
c) Multiple receiver with label Temperature(s)  (It could just have mean monthly temperatures or Mean min and mean max, and perhaps other possibilities including extremes as well.
d) I assume there might be a second axis label to allow Multiplier and Offset for the secondary axis?
e) Be able to save the graph of course.
f) Other options dependin g on what there are.  For example maybe the max and min could have bands? Could there be an Alphs control to help the visibility of the different components?

I am keen of checking the sort of AI process as well as the dialog.

And I wonder about a Walter-Leith graph, from climatol, on another button on the same tab.  I think it can use the same data?

I would also like to buuld that Climatic > Describe menu more.

MeSophie:
Fixes #8657
@rdstern this Pr is ready to be tested. Please have a look.


rdstern (commented):
@MeSophie and @lilyclements brilliant - it now seems to be working - though I don't quite understand some of the inputs.  Here is the result for Dodoma:

image

Here is the completed dialog:
image

Nice-eh.  Now the questions:

What is Absolute Temperature?  I kind of assumed it is the mean temperature rather than needing the max and min?  But I note it is essential as Ok isn't enabled without it.

The station name seems also to be needed?  That's fine as it will usually be a variable and it is easy to add if not.

It would be good to get this merged as soon as we can.  More generally we might make mean_tmax and mean_tmin into climatic, so parts of the dialog fill automatically.  But that can come later.

lilyclements:
@rdstern I answer your questions below:

> What is Absolute Temperature? I kind of assumed it is the mean temperature rather than needing the max and min? But I note it is essential as Ok isn't enabled without it?

A `ta_min` parameter came from the original example. I am not sure what ta_min is (as asked on the issue), so I don't know how important it is for the graph itself. Is it absolute minimum temperature? We can calculate that internally in the function if so :) 

To the bug you found, I've now fixed it in the `prepare_walter_lieth` function. @MeSophie I hope you don't mind, but I fixed this in your R code and have merged it into this PR. @rdstern this problem should now be fixed.

rdstern:
@MeSophie and @lilyclements thanks for the fixes and also the explanation.  I checked, and now found this:

The "Absolute monthly minimum temperature" is needed, so you are perfectly correct in not enabling Ok, until it is given.  I found this definition, which we can include in the help:  It is the: 

"Lowest daily minimum temperature observed during a given calendar month over a specified period of years." Copyright 2022 American Meteorological Society (AMS).

So it is simply the min of tmin in the 2 monthly data frames.

lilyclements:
@rdstern I thought there was just one monthly data frame. Which is the second one that you are referring to?

Once I know, I'll edit the function so that we just calculate the minimum of the min_tmin column in the R function. Then @MeSophie can remove it completely from the dialog side once I've done that.

lilyclements:
@rdstern thanks - I was hoping that was the case.

@MeSophie can you remove the `ta_min` option completely from the dialog side? I've amended the code and merged the changes into this branch so we no longer need the `ta_min` option, it's automatically calculated now.

rdstern:
@lilyclements great, though I am slightly surprised you are able to omit this element, if it is really the min of the daily min.  On the one hand I didn't see what changed in the graph when I changed what I put in that control.  But, if it really uses the min of the daily min, then I don't see how that can be calculated from the mean of the min?

I'll wait for the changes by @MeSophie and then check again.

lilyclements:
@rdstern this is a good point. I misunderstood the mean_tmin as the min_tmin. I’ll add the ta_min variable back in!

MeSophie:
@lilyclements Sorry the revert that you made brings some troubles on stand alone function. I was working then I pull your last changes and stating get error when trying to import data after investigation I saw that the problem if coming from stand alone function. Please can you have a look?

lilyclements:
@MeSophie sorry about that - completely my mistake. Can you try now? It should be fixed

MeSophie:
I will close this because of merge conflicts.

Issue #8726, opened 2024-01-11
Corrections on the Describe > Graphics > Bar Chart dialog


rdstern:
@N-thony I wonder if this might be a good task for @Fidel365 ?  It is currently a small change, but there may be further small edits needed as I write the Help for this graph.  I already have a second edit needed, see below.

![image](https://github.com/africanmathsinitiative/R-Instat/assets/11226469/9a7d6842-ab4b-494e-880f-5e68bae7a582)

a) This first change is a minor layout change where the arguments for the Lollipop charts need to move up so they are on the same line as the checkbox.  Also delete the Colour option, so it fits in the line above.

b) The second change is that the Reorder option seems to work fine with the ordinary Bar, but is ignored with the lollipop option.  Also, when you include the By factor (Optional), it adds the fill argument into the ggplot aes.  That's right, for the bar chart, but is ignored, by the lollipop.  So when lollipop is selected, then change fill to colour.

c) And the back-to-back option doesn't work anymore.  I think the code needs 2 tweaks.  It will be good for @Fidel365 to understand the graph as well as the coding, so I describe in some detail.  These are also called population pyramids   I suggest you can use the efc data from the library, with the sjmisc package.

Make e16sex variable into a factor and then use the dialog as shown below:

![image](https://github.com/africanmathsinitiative/R-Instat/assets/11226469/7e6c5a64-8627-4aaa-bda8-424fdebbfe50)

Note the back-to-back and flip checkboxes.  

The result is close, but the bar charts ore not back-to-back.  Here is a picture of the R-code generated:

![image](https://github.com/africanmathsinitiative/R-Instat/assets/11226469/a4f46e61-e05e-4804-8e94-65f87d0d08f3)

These are the corrections I suggest: (my R isn't good so I hope @jkmusyoka can also test the corrections and maybe have more suggestions.)  Note also the [stackoverflow here](https://stackoverflow.com/questions/14680075/simpler-population-pyramid-in-ggplot2). 

a) Delete lines 4 and 5.  - we don't need 3 repeats of the same line
b) Omit line 9, i.e. `ggplot2::geom_bar(stat="count") +  `.  Note this line changes if the dialog sets to dodge, etc.  Ignore that too.  The setting doesn't apply to the back-to-back graph.
c) In line 10 change `aes=aes(y=..count..*(-1)), stat="count"` to `aes(y=after_stat(count)*(-1))`

So here it is (I think) which works:

```
[...]

PR #8851, opened 2024-02-29
Corrections to the Describe>Graphs> Bar chart dialog

Fidel365:
Fixes issue #8726 , @rdstern , can this be reviewed.


rdstern (changes requested):
@Fidel365 if I understand correctly, then @Vitalis95 has made (many of) the last set of changes?  Thanks Vitalis.  But Fidel, did you then check that it is working?  Otherwise how are you learning, from these experiences when you get help?

I ask because I think the lollipop option is working well now.  But the example, in the issue, for the back-to-back, no longer works.  It should not take you long to find the problem and make the correction.  Please could you do this.

And in future Fidel you are still writing very short notes.  Please can they, in future, say what has changed since the last review abd how you checked, e.g. which data you used.  That will help me in the future.

Finally I note this is a new pull request.  Does this mean that the previous one is now redundant?  If so, then please could you close it.

rdstern:
@Fidel365 has repleid very constructively on skype and could not repeat the error I found.  It is more complicated to reproduce than I thought.  It arises in a quite common way (I am afraid) for our code, namely that an option remains in the R-code from a previous analysis with the same dialog, when particular settings are changed.

a) I used the efc data, made sex a factor and first got a back-to-back.  No problem:

image

b) Then click to the `Value` option at the top and prtess Ok.  That gives an odd bar chart - not very sensible.
c) Then back to `Frequency` and press Ok.

This gives an error.  Here is the R code generated.

```
# Dialog: Bar Chart
efc <- data_book$get_data_frame(data_name="efc")
last_graph <- ggplot2::ggplot(efc, mapping=ggplot2::aes(x=e17age, fill=e16sex)) + ggplot2::geom_bar(data=subset(efc, e16sex == levels(x=e16sex)[1])) + ggplot2::geom_bar(data=subset(efc, e16sex == levels(x=e16sex)[2]), aes(y=after_stat(count) * (-1)), stat="identity") + lemon::scale_y_symmetric(labels=abs) + theme_grey() + ggplot2::coord_flip()
data_book$add_object(data_name="efc", object_name="last_graph", object_type_label="graph", object_format="image", object=check_graph(graph_object=last_graph))
data_book$get_object_data(data_name="efc", object_name="last_graph", as_file=TRUE)

```
You can see the problem!  The move to the Value option adds `stat="identity"` to the code, and this still remains in the code, when we return to the frequency option.  I suspect more than back-to-back will go wrong?

So we should not keep that option hanging, when moving back to the Frequency option.  Maybe we could consider changing `geom_bar` to `geom_col`  when we change from Frequency to Value?  That's instead of keeping geom_bar and adding `stat="identity"` .  (There are more options on this dialog, and they each change the geom.)

rdstern (approved):
@Fidel365 that's great.  @N-thony can you please check and merge when you are ok?

Fidel365:
Issue #8864 is linked to this issue

PR #8853, opened 2024-03-01
Ensured that R script only analysed if it has changed since the last analysis

### LINKED ISSUE #8852: In Script window, make script analysis more efficient
Each time the user asks to execute a single statement, R-Instat currently analyzes the entire script (with `New clsRScript(clsScriptActive.Text)`). This is not very efficient. Change R-Instat so that if the script has not changed since the last time the script was analyzed, then do not analyze it again.

lloyddewit:
Fixes #8852 

@rdstern 
This PR should not change the script window functionality. It just makes the statement-by-statement execution faster by storing the R code analysis and only reanalysing if/when the script changes. The analysis is anyway very fast (the time taken to execute the statement is mainly in the R environment) so you will not notice any difference in statement-by-statement execution speed for short scripts. But on an older laptop with very long scripts this PR should make execution faster.

'Run All' and 'Run Selected' functionality is not affected (it's not worth optimizing this because these options create a temporary script with a comment at the top).

Please could you test?
Thanks

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -25,6 +25,7 @@ Public Class ucrScript
     Private bIsTextChanged = False
     Private iMaxLineNumberCharLength As Integer = 0
     Private Const iTabIndexLog As Integer = 0
+    Private clsRScript As RScript = Nothing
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@lloyddewit and @N-thony I'm not sure how to test this, except just by checking that it still seems to run scripts as before - which I did.  So, I'd be happy for it to be merged if you are ok Antoine.

Issue #8817, opened 2024-02-15
Box cut off in many dialogs


rachelkg:
Hello @N-thony and @rdstern, 

I think I have mentioned this before but it still has not been fixed and I just think it looks unprofessional. This box is cut off in many dialogs including Open from library and Paste to new Data Frame. 

Can it be fixed? 

Thanks, 

Rachel 

![2024-02-15_11-44-00](https://github.com/IDEMSInternational/R-Instat/assets/85707390/cfe0dd51-72c8-48c8-b846-95c178adda81)
![2024-02-15_11-32-52](https://github.com/IDEMSInternational/R-Instat/assets/85707390/f6c5097c-f55a-4ef0-b66b-0101a0876e80)

Raymond-Andam:
@rdstern I have been working on Issue #8800. There were issues with the code which I discussed with @N-thony  and Derrick. The issue has been fixed and I am working on creating the PR for it.

rdstern:
@Raymond-Andam thanks for letting us know, and it is good to hear from you in github.  In general, you should be working on more than 1 task at a time.  @N-thony also has short daily reporting meetings with the team, and you should be part of them.  Maybe you are? And, if allocated a task, then you should also report in github - even if it is to say you are busy on other work/tasks - so just as you have now done above.

PR #8857, opened 2024-03-05
Design change to  NewDataFrameName

Fidel365:
Fixes issue #8817 , @rdstern , this is ready for review and I will be keeping an eye on other dialog(s). This works for the mentioned dialog(s) and other dialogs as well; note that I will continue to check on this while I work


rdstern (approved):
@Fidel365 great, that does seem to have worked.  I am approving.  @rachelkg Fidel seems to clain that the particular "unprofessional looking" issue has been fixed more generally.  Really happy if you want to check, and also point out other untidy bits in our dialogs.
@N-thony over to you to check and merge.

Issue #8800, opened 2024-02-10
Improvements to Describe > Graph > Scatterplot


rdstern:
@derekagorhom this is another task mainly for @Raymond-Andam .  I hope it can be done in parallel with his other work.  I am keen for developers to have multiple tasks, because they may have to wait for help, or testing on each one.

This should initially be easier than some of the tasks on #8789.

Here is the current scatterplot dialog:

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/72c35488-6abd-40d9-a6af-3e90d9768694)

a) Could you make a group box with label `Add.`  That is round the current `Add Line of Best Fit`, and `Add Rug Plot`  
b) In each of those please delete the word Add.  It doesn't have to be repeated.
c) Add another checkbox with label `Side Plot`.  For now it can be disabled.
d) Decrease the vertical space slightly between the 3 options in the group box labelled Geoms.  Also delete the s, so it is Geom.
e) Make that group box wider.  Then, in the Point row, add a checkbox before Size and another before Shape.  Default is unchecked, and then the up-down and drop down controls are hidden.
f) You don't need checkboxes for the other two, but could move theor controls over, so they line up neatly with those for the point option.
g) Change the R code generated slightly.  I think it now gives you the size, etc inside the geom_point(), and they should now only be when the checkbox is ticked.  (The reason is that it is possible also to have a size and/or a shape aesthetic (i.e. a variable).  If size and/or shape are given as a single value inside geom_point, then it will ignore the aesthetic.

That's all for you now, but I have a question for @lilyclements .

I am very keen to be able to add marginal plots to the scatterplots. 

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/4693fae9-b14f-4f91-83c8-bc2d4e31bf6e)

I think they are useful anyway and partly for teaching.  The popular way most people seem to use is with ggExtra, asa shown above.  It is simple and nice and powerful, but has 2 defects to me anyway.
a) It is a slightly different syntax.
b) It seems not to cope with facetted plots.
A more recent package is ggside.  It will be a bit more complicated to add to the dialog, but I think it looks great.  Do you agree?

Raymond-Andam:
@rdstern This partly fixes issue #8800

lilyclements:
@rdstern only took me 8.5 months, not 5 years.
I do agree. I don't like the "wrapper" syntax in `ggExtras`.
[...]

PR #8865, opened 2024-03-07
Improvements to Describe > Graph > Scatterplot 

Raymond-Andam:
Fixes partly #8800
This PR fixed parts a, b, c, d . 
This PR is ready for review @rdstern @N-thony

### DIFF
--- a/instat/dlgScatterPlot.Designer.vb
+++ b/instat/dlgScatterPlot.Designer.vb
@@ -42,8 +42,6 @@ Partial Class dlgScatterPlot
         Me.lblFactorOptional = New System.Windows.Forms.Label()
         Me.lblXVariable = New System.Windows.Forms.Label()
         Me.lblVariable = New System.Windows.Forms.Label()
-        Me.lblSize = New System.Windows.Forms.Label()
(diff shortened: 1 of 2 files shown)


rdstern (commented):
@Raymond-Andam please can you also inform @fran2or when Items are ready for testing.  @fran2or I hope you are now able to be a tester, because that would help me and also allow you to monitor Raymond's progress.
@Raymond-Andam did you check the work, before asking me to test?  You said you have done a to d.  Reporting what you have done in this way is great, so I know what to test.

I like the new layout.  But neither of those options now work, either line of best fit, or the rug plot.  They used to.  And even the settings are no longer remembered when we return to the dialog.

We call this "regression", namely when an improvement somewhere causes bits that used to work fine, to stop working.  Please fix and quickly.  Assuming you are still keen to continue.

In preparation for the next steps could you also arrange for  ggside to be added to the packages in R-Instat.  

And @derekagorhom maybe you need to become involved here too.  It would be good for these improvements to be in the next release - scheduled for 15 March.  I had assumed these changes would take a few days - rather than months.

PR #8868, opened 2024-03-08
Upgraded  R.NET

ChrisMarsh82:


ChrisMarsh82:
@N-thony I think this will fix the issue with 7.16 not running on Windows 11. Let me know when a new build is done and I will double-check it solves the issue before closing it.

rdstern (approved):
@N-thony not sure what to check, but approving, so it can be merged!

PR #8870, opened 2024-03-11
Correcting the browsing in the Export Dataset Dialogue

### LINKED ISSUE #8866: Undo the regression in the File > Export > Export DataSet and maybe other dialogs
@N-thony I show the dialog from the merged version:

image

This is nopt the same as when I use Version 0.7.18.

This looks like the same (new) dialog from the file.choose( ), that we have added to the Insert dialog in the Script window improvements.  It doesn't work now, and I wonder where else it is now used in error?

N-thony:
Fixes #8866
@rdstern I have used this as a good example to show the team in our today meeting on how to debug the code and fix when there is an issue. The problem was more general and coming from `GetVariableNamesList` function from `ucrReceiverMultiple` control. Have a look.

### DIFF
--- a/instat/ucrReceiverMultiple.vb
+++ b/instat/ucrReceiverMultiple.vb
@@ -322,7 +322,7 @@ Public Class ucrReceiverMultiple
     End Function
 
     Public Overrides Function GetVariableNamesList(Optional bWithQuotes As Boolean = True, Optional strQuotes As String = Chr(34)) As String()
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@N-thony many thanks - that's a relief.  @Patowhiz if you could check the code, and merge if ok?

Patowhiz (approved):
Thanks for showing the team @N-thony .

PR #8869, opened 2024-03-10
Exploratory prototype for RSyntax and RCodeStructure classes

lloyddewit:
@rdstern @N-thony 
Please ignore this PR. I am only using this PR to manage some temporary changes while I explore the RSyntax and RCodeStructure family of classes. I will close this PR when done.
Thanks

PR #8881, opened 2024-03-17
Refactor RSyntax

lloyddewit:
@rdstern @N-thony Please ignore for now, thanks

### DIFF
--- a/instat/clsRSyntax.vb
+++ b/instat/clsRSyntax.vb
@@ -53,25 +53,15 @@ Public Class RSyntax
     ' 'TODO SJL It's not valid for an object of this class to be more than one of the 3 types above. 
     '     However the booleans potentially allow this. Replace with an enumeration?
 
-    ''' <summary>   An R function of the form 'RCommand(param1=param1Val, param2=param2Val, ...)'. </summary>
-    Public clsBaseFunction As New RFunction
+    ''' <summary>   If true then don't include the output part in the script (i.e. the part of the 
+    '''             script to the left of the assignment operator '&lt;-'). </summary>
+    Public bExcludeAssignedFunctionOutput As Boolean = True
 
-    ''' <summary>   An R operation of the form 'leftSide Operator rightSide' (e.g. "x+y"). </summary>
-    Public clsBaseOperator As New ROperator
+    ''' <summary>   If true then run the R script in a separate thread. </summary>
+    Public bSeparateThread As Boolean = True
 
-    ''' <summary>   An R command (of any type). </summary>
-    Public clsBaseCommandString As New RCodeStructure 'TODO SJL 17/04/20 What's the connection between this and 'bUeseCommandString' and 'strCommandString'? 
-
-
-    ''' <summary>   The R command in the form of a string. </summary>
-    Public strCommandString As String = ""
-
-    ''' <summary>   The R functions/operators/commands that should be run before the base R code. </summary>
-    Public lstBeforeCodes As New List(Of RCodeStructure)
-
-    ''' <summary>   The R functions/operators/commands that should be run after the base R code. </summary>
-    Public lstAfterCodes As New List(Of RCodeStructure)
-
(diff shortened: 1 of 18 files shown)


lloyddewit:
@rdstern 
In PR #8869 I am working on a proof-of-concept. The most efficient first step is to simplify some of the current classes. This could also benefit the current version of R-Instat. This PR contains a refactored version of `RSyntax`. More refactoring is possible but this would increase the risk of regression and require significant testing. For this PR, all the dialogs should work exactly as before. Some suggested dialogs to test are below. If these still work correctly, then all the other dialogs should also be fine:
- dlgTwoVariableUseModel
- dlgModelling
- one of the graphics dialogs (e.g. Box plot)

@N-thony I removed about 300 lines of unused code and made some other small improvements. I also put the data members and functions in the correct order. I realise now that the reordering has made the `RSyntax` diff bigger and your peer review more complex. Sorry for that. The diffs are not as big as they appear, many of the functions are unchanged and have just been moved.
I tried hard to keep the refactoring as low risk as possible, but obviously `RSyntax` is one of the core classes. If you think it's necessary, then you could also ask another of the senior developers to review.

Thank you!

rdstern:
@lloyddewit this is exciting.  

a) I followed your suggestions and started with the boxplot dialog.  I used the usual survey data and it worked fine.  It seemed to work better than before when I tried the plot options button.  Was that my imagination?  It used to be very slow in going to the sub-dialog the first time, and it seems to be quicker now.  I was going to ask @N-thony to look at the code, but maybe don't need to?

b) Then I fitted a simple 2-variable model:
image

That's so I could test the dlgTwoVariableUseModel dialog:

That gives an error, as follows:

image

c) I started again, but am not sure whether I am testing the dlgModelling function. Is it the Fit > Model > General?  

image

If so, my initial tests it seems to work fine.

lloyddewit:
@rdstern Thank you for testing.

> a)

I would love to take credit for the perceived speed increase but I doubt it was my change :)

> b)

I was able to recreate the error in this PR but I was also able to recreate the error in master branch. Are we sure that the error is due to this PR's changes?

> c) I started again, but am not sure whether I am testing the dlgModelling function. Is it the Fit > Model > General

You can access the `dlgModelling` via:

image

lloyddewit:
@rdstern 
For part `b)` we could check the R code in the log file that is generated by the dialog. If the R code generated by master branch is the same as the R code generated by this PR, then this PR has not caused any regression. Would this be an adequate test to approve this PR?
Obviously the error still needs fixing but this should be done in a separate issue/PR.
What do you think?

Patowhiz:
@N-thony, please note that the `get_models` R function has been deprecated and replaced with `get_object_data`. This change is documented in detail in issue #7808, where I've explained the new concept for processing output objects.

If you encounter any dialogs still generating `get_models`, they need to be updated. I believed all model dialogs had been refactored to use the new `get_object_data` function, so any remaining instances of `get_models` are unexpected and should be corrected.

rdstern:
@N-thony I hope the discovery of the `get_models` R-Instat function is the cause of the apparent regression in the modelling dialog.  Should there be a new issue on this topic or do we continue here?  I assume there could be lurking `get_models` elsewhere in the modelling dialog, so this is a good time to check more broadly?

lloyddewit:
@rdstern If the `get_models` is the only known problem (to be fixed in a separate PR), then is this PR ready to merge or does it need more testing?
Thanks

rdstern (approved):

lloyddewit:
@N-thony I think I just found an error. It looks like it is caused by this PR. Please could you confirm the error?
If you can recreate the error and agree that it is caused by this PR, then please could you revert the merge?
I am very sorry for the inconvenience.

@N-thony @rdstern 
I found an issue in the Enter dialog. If I try and add a new column containing then the correct R code is not generated. Details below. I will investigate the error tomorrow.

image

image

Expected script (from 0.7.16):

image

Script from this PR:

image

PR #8887, opened 2024-03-20
Displaying multiple tables (html) in the output window

### LINKED ISSUE #8855: Having multiple tables (presumably html) in the output window?
@Patowhiz I am trying to redo the Describe > Two/Three Variable dialog, and I keep coming up agianst the limit that we can currently have only a single (html) table, at a time in the output window.  The dialog produces a different type of summary for each of the combinations (2 variables) or cat/cat, (frequency tables) cat/num, (summary tables) num/cat (ANOVA) and num/num (ANOVA).  

The ANOVA options are very happy with multiple tables.  The frequency and summary tables are not. 

I wonder how much work it is to change that?  I assume it is the same limitation if we produce an rtf output?

Ideally we could havce multiple html, but otherwise could the (multiple) tables more easily be non-html ones?  

If this is realoly difficult I wonder - maybe @lilyclements could advise, - whether we could quickly only use gt for the output, when there is a single variable in the first control, and not bother with it, if there's more than one?  

I just want to make sure we have looked at all the options, before spending too much time on changes to this dialog, simply to fit within current output restrictions?

N-thony:
Fixes #8855
@rdstern I couldn't reproduce your example in the corresponding issue on having multiple tables to the default browser. 
This is a quick check but will still investigate on having the multiple tables in output window rather than default browser. Can you test this?

### DIFF
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -1006,7 +1006,7 @@ Public Class RLink
                 For Each _path In arrFilesPaths
                     ' Add each HTML file as an output to clsOutputLogger
                     ' strScriptWithComment: the script with associated comments
-                    ' line: the path to the HTML file
+                    ' _path: the path to the HTML file
                     ' bAsFile: a boolean indicating whether the output should be treated as a file
                     ' bDisplayOutputInExternalViewer: a boolean indicating whether to display the output in an external viewer
                     clsOutputLogger.AddOutput(strScriptWithComment, _path, bAsFile, bDisplayOutputInExternalViewer)
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -996,10 +996,9 @@ Public Class RLink
                 End If
             End If
 
-            ' Split the strOutput into an array of lines, removing empty entries
-            Dim arrFilesPaths() As String = strOutput.Split({Environment.NewLine}, StringSplitOptions.RemoveEmptyEntries)
-
             If bAsFile Then
+                ' Split the strOutput into an array of lines, removing empty entries
+                Dim arrFilesPaths() As String = strOutput.Split({Environment.NewLine}, StringSplitOptions.RemoveEmptyEntries)
                 ' Iterate through each HTML files
                 For Each _path In arrFilesPaths
                     ' Add each HTML file as an output to clsOutputLogger
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -997,11 +997,14 @@ Public Class RLink
             End If
 
             Dim lines() As String = strOutput.Split({vbCrLf, Environment.NewLine}, StringSplitOptions.RemoveEmptyEntries)
-            For Each line In lines
-                'log script and output
-                clsOutputLogger.AddOutput(strScriptWithComment, line, bAsFile, bDisplayOutputInExternalViewer)
-            Next
-
+            If Not String.IsNullOrEmpty(strOutput) AndAlso lines.All(Function(line) File.Exists(line) AndAlso New FileInfo(line).Length > 0 AndAlso
+                    Path.GetExtension(line).Equals(".html", StringComparison.OrdinalIgnoreCase)) Then
+                For Each line In lines
+                    clsOutputLogger.AddOutput(strScriptWithComment, line, bAsFile, bDisplayOutputInExternalViewer)
+                Next
+            Else
+                clsOutputLogger.AddOutput(strScriptWithComment, strOutput, bAsFile, bDisplayOutputInExternalViewer)
+            End If
 
         Catch e As Exception
             MsgBox(e.Message & Environment.NewLine & "The error occurred in attempting to run the following R command(s):" & Environment.NewLine & strScript, MsgBoxStyle.Critical, "Error running R command(s)")
@@ -1036,7 +1039,7 @@ Public Class RLink
             'if not, just return empty file path
             strFilePath = String.Join(Environment.NewLine, expTemp.AsCharacter())
             Dim lines() As String = strFilePath.Split({vbCrLf, Environment.NewLine}, StringSplitOptions.RemoveEmptyEntries)
-            lines = lines.Where(Function(line) File.Exists(line) OrElse New FileInfo(line).Length = 0).ToArray()
+            lines = lines.Where(Function(line) File.Exists(line) OrElse New FileInfo(line).Length > 0).ToArray()
             strFilePath = If(lines.Length = 0, "", String.Join(Environment.NewLine, lines))
         End If
         Return strFilePath
--- a/instat/clsRLink.vb
+++ b/instat/clsRLink.vb
@@ -999,17 +999,17 @@ Public Class RLink
             ' Split the strOutput into an array of lines, removing empty entries
             Dim arrFilesPaths() As String = strOutput.Split({vbCrLf, Environment.NewLine}, StringSplitOptions.RemoveEmptyEntries)
 
-            ' Check if strOutput is not empty and all lines correspond to existing HTML files with content
-            If Not String.IsNullOrEmpty(strOutput) AndAlso arrFilesPaths.All(Function(line) File.Exists(line) AndAlso New FileInfo(line).Length > 0 AndAlso
-                    Path.GetExtension(line).Equals(".html", StringComparison.OrdinalIgnoreCase)) Then
+            ' Check if strOutput is not empty and all files correspond to existing HTML files with content
+            If Not String.IsNullOrEmpty(strOutput) AndAlso arrFilesPaths.All(Function(_path) File.Exists(_path) AndAlso New FileInfo(_path).Length > 0 AndAlso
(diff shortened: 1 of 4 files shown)


rdstern:
@N-thony that would be really great, but I don't have any current dialog to test this with?  The obvious one is the Describe > 2/3 Variable summarise.  Maybe you could tweak the code for this dialog in this pull request?

image

Currently it just gives the table for village.  It ignores the 2nd factor in the multiple receiver.

rdstern (dismissed):
@N-thony that's great.  I like it a lot.  @Patowhiz can you check and merge if ok?

Patowhiz (commented):
@N-thony kindly add comments to code changes in` clsRLink.vb` and `instat_object.R`. This will help me understand the changes better and will be useful in future maintenance.

Patowhiz (commented):
Also try using RStudio formatting options when editing the R files. It help with having consistency in the line formats in those files.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -641,18 +641,43 @@ DataBook$set("public", "get_object", function(data_name = NULL, object_name) {
...
+        out <-
+          view_object_data(object = out$object,
+                           object_format = out$object_format)
+      }
+    } else {
+      out <-
+        view_object_data(object = out$object,
+                         object_format = out$object_format)
+    }
+  } else {
```
Patowhiz, line 675: I recommend you implement these changes inside `view_object_data` function  instead of `get_object_data` function. They relate more to viewing the object data as file than getting the object data. I also think it will be more efficient and clear when done inside `view_object_data`.

instat/clsRLink.vb
```diff
@@ -996,9 +996,26 @@ Public Class RLink
                 End If
             End If
 
+            ' Split the strOutput into an array of lines, removing empty entries
+            Dim arrFilesPaths() As String = strOutput.Split({vbCrLf, Environment.NewLine}, StringSplitOptions.RemoveEmptyEntries)
```
Patowhiz, line 1000: Considering that strOutput always uses `Environment.NewLine`, you don't have to use `vbCrLf` when splitting it to array. Using `Environment.NewLine` is enough.

instat/clsRLink.vb
```diff
@@ -996,9 +996,26 @@ Public Class RLink
...
+
+            ' Check if strOutput is not empty and all files correspond to existing HTML files with content
+            If Not String.IsNullOrEmpty(strOutput) AndAlso arrFilesPaths.All(Function(_path) File.Exists(_path) AndAlso New FileInfo(_path).Length > 0 AndAlso
+                    Path.GetExtension(_path).Equals(".html", StringComparison.OrdinalIgnoreCase)) Then
```
Patowhiz, line 1004: The file exist check is done by `GetFileOutput` function. You can simply use `bAsFile` to check whether you should iterate the strOutput or not.

instat/clsRLink.vb
```diff
@@ -1029,14 +1046,20 @@ Public Class RLink
...
+            ' Extract the file path names from the expTemp and join into a single string
+            ' Split the string into an array of file path names, removing empty entries
+            ' Check if each file path name corresponds to an existing file with content
+            ' If so, filter out invalid file paths
+            ' Combine the valid file paths into a single string separated by newline characters
+            ' If no valid file paths remain, set strFilesPath to an empty string
+            strFilesPath = String.Join(Environment.NewLine, expTemp.AsCharacter())
+            Dim arrFilesPath() As String = strFilesPath.Split({vbCrLf, Environment.NewLine}, StringSplitOptions.RemoveEmptyEntries)
+            arrFilesPath = arrFilesPath.Where(Function(path) File.Exists(path) AndAlso New FileInfo(path).Length > 0).ToArray()
+            strFilesPath = If(arrFilesPath.Length = 0, "", String.Join(Environment.NewLine, arrFilesPath))
```
Patowhiz, line 1059: You don't need to recreate the `strFilesPath` from the array, The array was derived by `arrFilesPath`.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -2734,46 +2734,64 @@ view_text_object <- function(text_object){
...
-    #"mmtable2" package produces "gt_tbl" objects 
-    gt::gtsave(html_object,filename = file_name)
+  for (i in seq_along(html_objects)) {
+    # Get the current HTML object
+    html_object <- html_objects[[i]]
+    
+    object_class_names <- class(html_object)
+    
+    # Generate a unique temporary file name
+    file_name <- tempfile(pattern = paste0("viewhtml_", i, "_"), fileext = ".html")
```
Patowhiz, line 2758: Calling `file_name <- tempfile(pattern = "viewhtml", fileext = ".html")` will guarantee unique name, so the concatenation of the index `i` is not really necessary. I recommend removing it for simplicity.

instat/static/InstatObject/R/stand_alone_functions.R
```diff
@@ -2735,6 +2735,18 @@ view_text_object <- function(text_object){
 #it saves the object as a file in the temporary folder
 #and returns the file path.
 view_html_object <- function(html_object){
+  # Check if html_object is a list and has more than one element
+  if (is.list(html_object) && all(sapply(html_object, class) == class(html_object[[1]]))) {
+    # If html_object is a list with multiple elements of the same class, 
+    # recursively call view_html_object on each element
+    file_names <- vector("list", length(html_object))
+    for (i in seq_along(html_object)) {
+      file_names[[i]] <- view_html_object(html_object[[i]])
```
Patowhiz, line 2744: I noticed that in scenarios where `html_object[[i]]` is a list ,  `file_names[[i]]` will also end up having a list, which might cause unexpected behavior for functions dependent on this one.

I understand that you are trying to use recursions to achieve what could be done with a simple `for` loop. While recursion can be elegant, in this case, a `for` loop would be more optimal and clearer. For instance, in scenarios where `html_object` is a list, the check `if (is.list(html_object) && all(sapply(html_object, class) == class(html_object[[1]])))` is executed for each element in `html_object`. This redundancy makes the recursion inefficient.

Additionally, it took me a while to reproduce the possible bug that I just outlined above, which further supports the need for a more straightforward approach.

I recommend refactoring the recursive logic into a `for` loop to enhance both performance and code readability.

Patowhiz (changes requested):
@N-thony I have made a few comments and request for changes. Thanks.

Patowhiz (changes requested):
@N-thony, this is quite insightful. We will definitely want to extend this functionality to handle other formats like text and graphs in the future, but that can be addressed later.

I noticed that some valuable comments have been removed from the `view_html_object` function in R. Please re-insert these comments into their appropriate code blocks. These comments were added to explain the rationale behind certain decisions related to various outputs. In the future, I plan to introduce generic solutions based on these decisions. Removing these comments could hinder future developers or myself from quickly grasping the reasons behind these decisions. It's worth noting that the comments include the dates when these decisions were made, highlighting the importance of retaining these comments within the function.

Patowhiz (approved):

Patowhiz:
@rdstern, kindly retest this before giving your approval. Additionally, please test other dialogs that produce various types of output objects to ensure that we haven't inadvertently introduced any regressions. Apologies for the inconvenience, but you are our "unit tests" for these scenarios.

rdstern:
I tried the same as shown in the dialog above.  It works fine in the dialog.

I then did To Script and tried the script, which was as follows:

```
# Dialog: Describe Two Variables

last_table <- survey %>% purrr::map(.x=c("village","variety"), .f=~data_book$summary_table(data_name="survey", treat_columns_as_factor=FALSE, summaries=count_label, factors=c("fertgrp",.x), columns_to_summarise=.x) %>% pivot_wider(names_from={{ .x }}, values_from=value) %>% gt::gt())
rm(list=c("last_table", "survey"))

```
This gives an error, see below.  When running line-by-line the error is in the `data_book$get_object_data function`.

image

rdstern (approved):
@N-thony this is great.  I am approving.  I tried just a few more dialogs.
My only query is the code generated.  I made variety into a factor, so I had 3 tables generated.  
The code, when I ran from the script looks fine and now runs well.  It seems a lot neater even than the code when run from the dialog, because it just repeats the last main line,before each table.  When running from the dialog it repeats the whole code before each of the 3 tables?

Patowhiz:
@rdstern did you test other dialogs that produce various types of output objects to ensure that we haven't inadvertently introduced any regressions?

rdstern:
@Patowhiz I have tried some - including the calculator, and various from the describe menu that produce graphs and tables.  So far, so good.  I also wonder about these dialogs where we disabled the Both option, when you redid the output window?

Perhaps they could now be re-instated?  It is not high on my list of priorities, but might be a good additional test to see if they now work easily:

This is 2-way - there is also 3-way:

image

Patowhiz:
@rdstern sounds like a good task for @N-thony and @Vitalis95 to look into those dialogs once they become a priority.

Patowhiz (approved):

Issue #8886, opened 2024-03-20
Improve the circular keyboard and a link to numeric convert


rdstern:
@Fidel365 I think this is nearly the last one - oh, except for factor keyboard!  Again a mixture of small improvements, but mainly tooltips. 

a) There is a popup dialog, called Convert to Numeric.  It only appears when you attempt to convert a variable to numeric and some rows aew non-numeric.  Could you please add the help ID of 614.
b) On all keys (except 1), in circular there is `na.rm=FALSE`.  Change our default to `TRUE`.
c) The A1 key gives circular::bessel(  ).  Change bessel to A1.

Tooltips for circular:
circular: define a variable as circular.  Specify whether the data are in radians (default), degrees, or hours.
mean the circular mean.  For example with 1, 2, 3, 6, mean(circular(c(1,2,3,6)) gives 1.51.
median circular median.  For example with 23 and 2 representing hours, median(circular(c(23,2), units="hours")) gives 0,5.
medianHL median using Hodges-Lehmann estimate. For example with 1,2,3,6, medianHL (and median) give 1.5.  
range: circular range is the shortest arc containing the data. For example with 1,2,3,6 gives 3.28 (6 is also -0.28 on circle from 0 to 2*pi)
sd: The circular standard deviation.  This is not the sqrt(circular.var).  It is reasonably close to the ordinary sd for data in radians, so from 0 and 6.28. 
ang.dev: the angular deviation is square root of the angular variance, so between 0 and sqrt(2).  
var: the circular variance is (1 - rho), so between 0 and 1, with small values implying low (circular) variation.  For 1,2,3,6 var is 0.5964.  With 1,2,3,4,5,6 var is 0.951.  
A1: Ratio of Bessel functions for values of kappa parameter.  Used in the von Mises (circular normal) distribution kappa = 0.9 gives A1 same as rho value for 1,2,3,6 data.
ang.var the angular variance is twice the circular variance, so between 0 and 2.
quantile: Defined quantile round the circle.  With 0.5 it is the (circular) median, so is 1.5 for 1,2,3,6.
rho: the mean resultant length is between 0 and 1. Small values imply large (circular) variation. For 1,2,3,6 rho is 0.4036. With 1,2,3,4,5,6 it is 0.049.
min:  Smallest value round the circle. For 1,2,3,6 min is 6. (values are from 6 (almost `2 * pi ` to 3)
q1: Lower quartile round the circle. For  1,2,3,6, q1 is 0.68
[...]

PR #8888, opened 2024-03-20
Improments on the circular keyboard in the calculator.

Fidel365:
Fixes issue #8886 , @rdstern this Pr is ready for review, also, according to point(b), you say all except one button should have the na.rm=FALSE(Changed to TRUE now), I can see two don't have; circular and A1, should be in one on either of these?

### DIFF
--- a/instat/frmConvertToNumeric.vb
+++ b/instat/frmConvertToNumeric.vb
@@ -68,9 +68,7 @@ Public Class frmConvertToNumeric
     End Sub
 
     Private Sub cmdHelp_Click(sender As Object, e As EventArgs) Handles cmdHelp.Click
-        Help.ShowHelp(Me.Parent, frmMain.strStaticPath & "\" & frmMain.strHelpFilePath, HelpNavigator.TableOfContents)
-        'Replace with specific page once written
-        'Help.ShowHelp(Me.Parent, frmMain.strStaticPath & "\" & frmMain.strHelpFilePath, HelpNavigator.TopicId, iHelpTopicID.ToString())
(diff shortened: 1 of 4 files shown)


rdstern:
@Fidel365 many thanks and you are becoming pretty effective in the improvements you are making to R-Instat.  
a) When changing the setting of missing to true, the default position of the cursor seems to be changed in each keyboard - at least when using the checkbox to "show parameters".  It is no before the equal sign.  (This may have been a problem before, but more likely introduced by your change.  

`circular:: mean.circular(x new_data= , na.rm = TRUE, control.circular = list())`

I hope that is clear?  Here is an example, after I have added the variable `new_data` into the formula.

b) In the medianHL key, please add quotes as shown below.  (I think this was always an error.)

`circular::medianHL.circular(x =new_data , na.rm = TRUE, method = c("HL1","HL2","HL3"), prop = NULL)`

c)  In the circular sd tooltip delete the confusing last phrase "so from 0 and 6.28

d) Add an s to the the quantile key - so it becomes `quantiles`.  Also add s in the tooltip - so:
`Defined quantiles`

e) I have been inconsistent in starting the tooltips with lower case or upper case.  Could you make them consistent and start each with upper case please.

f) You are right, above, when you said there are 2 keys, and not just one, where there is no change in the missing option. 

g) Finally, the issue, #8886 in point a) adds help link to a popup dialog.   Did you make this change?  (I can't test this, with the current help. )

These are all small changes, so I hope this can be merged soon.

Fidel365:
@rdstern , I have made the changes requested, among the changes made in this PR was adding a help id of 614 to convert to numeric; as seen below
Screenshot (55)

rdstern (approved):
@fidel.  Great, thanks.  @N-thony over to you

Issue #8882, opened 2024-03-17
Improving the Wakefield keyboard in the calculator


rdstern:
This is a continuation from #8878 which was becoming too long.

The Wakefield keyboard has many keys, for which it is easy to add useful tooltips. Also:
a) Change date_stamp to dates (D)
b) For iq, can you change the sd from 10 to 15?
c) For sat change the mean to 1000, sd to 150, maximum to 1600 and minimum to 400.
d) For upper, change the command to upper_factor. Keep just as upper on keyboard, but change the c to F.
e) Overall change (f) to (F) and (c) to (C), and (o.f) to (O.F).
f) Add buttonlower (F)on bottom row, and just after upper (F). It produces the lower_factor command.
g) Correct mis-spelling of linkert7 to likert7.
h) Change level key to lower, with correct coding for lower command. (It gives a single lower case letter.) (Note the very similar code for upper is already included)
i) The key called died is not needed. Instead include grade_letter, with code for that command
j) The key called sex is not needed. Instead have times (T)as key name, giving time-stamp as the command.
k) The key called ela is not needed. Instead include minute (T) with code for that command
l) There will then need to be a new issue and pull request for @N-thony to recognise H:M:S data and give the (T) for time in the data sheet.

Here goes on the tooltips!

age: Sample of ages, with default from 20 to 35
animal: Sample of animals, with default of 10 from a list of 591 animals!
pet: Sample of pets. Default is dog, cat, none, bird horse, with given probabilities
answer: Sample of No or Yes, with defaults of equal probability.
car: Sample of cars, with datasets mtcars makes as default.
children: Sample of number of children, with default 0 to 10 and defined probabilities
coin: Sample giving heads or tails with default of equal probability
colors: Sample by default from the 657 colours in the grDevices package.
primary: Sample by default from the colours, red, green, blue, yellow, black, and white. Are those your primary colours?
dates: Sample of dates with default being dates in order, by month, for the past year.
death: Sample giving FALSE/TRUE with default being equally likely.
died: (Key not needed. See tooltip for grade_letter below.)
dice: Sample with default being equally likely from a 6 sided dice.
dna: Sample with default being equally likely from Guanine, Adenine, Thymine, Cytosine
dob: Sample of dates, with default being for 2 years, starting 15 years ago
dummy: Sample of 0 and 1, with default equally likely.
education: Sample of 12 education levels from No education to doctorate with defined probabilities
[...]

PR #8892, opened 2024-03-22
Improvements to Wakefield keyboard in the calculator

Fidel365:
Fixes partially issue #8882.
Partially because of the changes needed for Time(T) and also, we have two lowers; the lower to replace levels and the lower(F) below Upper(F), are both of them needed? If so, I will need the tooltip for lower that replaces level key.
@rdstern , can this Pr be reviewed as we continue to work on this? thanks.


rdstern (changes requested):
@Fidel365 well done.
a) You have changed likert into likert7.  Could you please revert that one to be called just likert and to produce likert with 5 categories as before.  The change I requested was simply to the key name for likert7, which is at the bottom right and spelled wrongly.  It is currently called linkert7.  It should be just likert7.
b) You have added lower twice.  There is one in the middle and a second at the end.  Delete the second one.  And Add the (F) to the other one.  Also change LETTERS to `letters`.  It is supposed to give lower case - currently it says lower, but gives letters in capitals.
c) pet tooltip - add comma before the horse.
d) Delete the full-stop at the end of some tooltips.  (Sorry, that was probably me being inconsistent!)
e) Add (F) to the grade letter key.
f) Add arguments into the new times and minutes keys to parallel those for the dates.

rdstern:
@Fidel365 I'll check again as soon as you resolve the builde errors

Fidel365:
@rdstern because of the conflicts, I am closing this PR and we will be working from PR #8896 from now on.

PR #8895, opened 2024-03-25
Fixed a Bug on Climatic > Inventory Dialogue

### LINKED ISSUE #8893: Bug on Climatic > Inventory Dialogue
When you to the dialogue as shown in the screenshot below and click `OK`, nothing happen even tho the expection is to have the summary in the output. You get the output in this case only if you uncheck and then check again again on Summary checkbox and click `OK`
image

Vitalis95:
Fixes #8893 
@N-thony , @rdstern  have a look at it


rdstern (approved):
Looks finer now.  @N-thony good spot - I am really keen that we work hard to make R-Instat more solid, so there are as few surprises as possible.

Back to you now Antoine.

PR #8909, opened 2024-04-02
Improvements to Describe > Graph > Scatterplot

Raymond-Andam:
Fixes Partly Issue #8800.
This PR fixes parts e, f, and g. This PR is ready for review.
@fran2or, @rdstern.


rdstern (commented):
@Raymond-Andam thank you for the work.  Here is a screen-shot:
image

a) In the bits you have changed there is now have size and shape twice.  Why?
b) This is for both Point (shown) and for Count.
c) When you change that you won't need to have the checkboxes quire so close to the radio buttons.  That's important, because we have translation of the labels into many other languages and that looks messy if the controls are too close together.
d) I wonder whether you are being careful enough about the positioing of the controls.  It may be my imagination, but I think the 3 radio buttons are no longer quite aligned.  Neither (vertically) are the 3 checkboxes in the Add group box.

@fran2or are you now also able to check these different branches, or do you still have problems?

PR #8914, opened 2024-04-03
Sorting the display of numbers in the output window and the grid

### LINKED ISSUE #8906: Sort out the display of numbers in the grid and the output window
@N-thony I think this is p-robably one for you - at least it will need your guidance.

In the grid complex numbers don't get formatted  here is an example:

image

To get this, get the default data frame from File > New Data Frame.
Then, in the calculator type
complex(real=sqrt(x1)) and put the result in x2

Then complex(real= sqrt(x1), imag=2*sqrt(x1)) into a third variable

It is worse in the output window.  There no numbers are formatted.  For example calculating sqrt(2) and putting it into the output window gives:

image

I think this was ok (much) earlier, because there is a setting for the output window precision in the Tools > Options > Output Window

N-thony:
Fixes #8906
Fixes partially #7979 
@rdstern have a look

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -132,7 +132,7 @@ Public Class dlgCalculator
         Else
             ucrBase.clsRsyntax.RemoveFromAfterCodes(clsRemoveLabelsFunction)
             ucrBase.clsRsyntax.RemoveAssignTo()
-            ucrBase.clsRsyntax.iCallType = 1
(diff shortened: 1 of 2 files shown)


rdstern (approved):
@N-thony that's great.  It seems to fix both the grid (for complex numbers, where it went crazy before - and the output window.  I'm very happy and hope I haven't missed anything.  @Patowhiz could you check and merge if happy too!

Patowhiz (approved):

PR #8913, opened 2024-04-03
Fixed regression bug in PR #8881

lloyddewit:
@N-thony Please could you test that this PR fixes the regression error introduced by PR #8881?

@rdstern If @N-thony approves, please could you test the Calculator and Enter dialogs?

Thanks!

PR #8919, opened 2024-04-05
Ensured that Enter dialog assignments are correctly deleted

lloyddewit:
I found another regression in PR #8881:
The last line of the script generated by the `Enter` dialog should be `rm(list=c("enter", "survey"))
`. However it was only `rm(list=c("survey"))`.
I fixed this in this PR.
I also reversed an unnecessary (but harmless) change I made in PR #8913 (making `strScript` public).

@rdstern Please could you check that the Enter and Calculator dialogs still work as expected?

Thank you

### DIFF
--- a/instat/clsRSyntax.vb
+++ b/instat/clsRSyntax.vb
@@ -94,8 +94,6 @@ Public Class RSyntax
     ''' <summary>   The R command in the form of a string. </summary>
     Public strCommandString As String = ""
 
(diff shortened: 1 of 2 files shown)


rdstern (approved):
@lloyddewit I am currently working on the calculator - so that fits well.  I tried Enter too.  So far, so good.  @N-thony over to you.

PR #8920, opened 2024-04-08
Convert to Numeric for logical variables

### LINKED ISSUE #8917: Maybe Improve our right-click Convert to Numeric for logical variables
@N-thony I hope you can investigate and fix this.  It should be simple, but may become a bit of R that needs @lilyclements ?

Here is an example.  a) New data frame - 20 rows, press Ok.
b) Calculator  with x1 > 10 to make x2 a logical variable.
c) Right-click and Duplicate Column (keep the type the same) to make x21.
d) Right click on x21and Convert to numeric and it complains that values are not numeric.  If I then do `Ordinal Convert` then I think all is ok.  (And is the same as Duplicate Column and convert there.)

I would like it to do this automatically to numeric and not complain.

e) Now duplicate x2 again, and this time make the new variable a factor.
f) Right-click on this new variable and Convert to Numeric.  Here is the very odd message:

image

Again, I think Ordinal Convert does the right thing.  I would like that conversion to be automatic - without the popup menu at all in either case.

N-thony:
Fixes #8917 
@rdstern have a look.

### DIFF
--- a/instat/ucrDataView.vb
+++ b/instat/ucrDataView.vb
@@ -659,7 +659,6 @@ Public Class ucrDataView
             Dim iNonNumericValues As Integer = GetCurrentDataFrameFocus().clsPrepareFunctions.GetAmountOfNonNumericValuesInColumn(strColumn)
             If iNonNumericValues = 0 Then
                 GetCurrentDataFrameFocus().clsPrepareFunctions.ConvertToNumeric(strColumn, True)
-                'Continue For
             Else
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@N-thony that looks great now - and a relief that it is another step in making R-Instat simpler to use.  And alother change where I am approving first-time round!  Excellent.  @Patowhiz I hope the code is ok, so it can be merged?

Patowhiz (approved):

Issue #8889, opened 2024-03-21
Improvements in Help file with the addition of snippets


rdstern:
@Fidel365 the help numbers towards the bottom of this dialog are now ready for you to check and add.  Most, but not all, will work in the current help.  I have made a few edits in the current help file, so a few will not.

I will add more to this issue, so make this a "partially" pull request.  But it will be good to merge these changes and then have a separate pull request for the rest of them.  I estimate that I'm about halfway through.

I continue to work, with @rachelkg on the help file for R-Instat.  This uses software called helpNdoc.  The current help file with R-Instat has a problem that the software developers have just rectified - at least for our file.

This particularly affects our use of "snippets".  They are entries that are repeated more than once in the file.  We currently use them in 2 ways:
a) The bottom keys (Ok and to To Script) are described in a table towards the bottom of the help on each dialog.  That isn't repeated, but calls a single snippet.
b) Some dialogs are duplicated in different places in the menu system.  For example the pivot-table dialog is in the Prepare menu, in Describe and in Climatic.  The help is a single snippet, rather than being repeated 3 times.

Until version 9.1 of helpNdoc, the snippets in our help were not being shown.  Now they are.  We will therefore be including the new help in the next release.  

We would also like now to jump to the right location in the help whenever we use a dialog that is repeated.  Currently we always jump to one of the instances.  I am listing the multiple help numbers below.  @Fidel365 has already made this adaptation for one example (Visualise).  Here are the help numbers for the remaining dialogs - mainly in the prepare menu.

a0) `Replace Values` in Prepare > Data Frame, Confirm 47; in Climatic > Tidy data 604
a) `Pivot Table` - in Prepare > Check data 319; in Describe 2/3 variable - confirm 417; in Climatic > Examine/Edit data 558
b) `Duplicate rows` in Prepare > Check data confirm 547; in Climatic tidy data 605.
c)  ` Compare columns `in Prepare > Check data confirm 546; in Climatic examine/edit data 562
d) `Non-numeric values` in Prepare > Check data confirm 545; in Climatic tidy data 603
e) `Boxplot `in Prepare > Check data 548 In Describe > Graphs > Boxplots Confirm 436
f)  `One variable summarise` in Prepare > Check data 550; In Describe > One Variable Confirm 410 ; In Climatic Edit Data 615
g) `One variable graph` in Prepare > Check data 549; In Describe > One Variable Confirm 412 ; In Climatic Edit Data 616
[...]

PR #8923, opened 2024-04-08
Adding help id's for snippets in R-instat

Fidel365:
Fixes partially issue #8889.
Partially because part (a) is not yet fixed.
@rdstern this PR is ready for review


rdstern (approved):
@Fidel365 great.
I have kept your partially, though the point a) you mention - good spot - is now ok.  But there is a second set of help IDs that are to come later, but are only partially there now. So no more work on this issue yet.
@N-thony over to you.

PR #8924, opened 2024-04-08
Corrected error after position control is used

### LINKED ISSUE #8908: Correct error after position control is used.
@N-thony I think this could be very quick for you - and probably take longer to explain to anyone else?  But up to you.

image

It occurs with the use of multiple data frames.

Here in our survey data I have a summary data frame as well.  It does not have village in it.

a) I do a calculation in the ordinary survey data and then use the Position sub-dialog to save it after the Village variable.  I like the fact that this "rule" is remembered if I do another calculation with the same data frame.

b) But I then move to another data frame.  I don't like the fact it still remembers the rule.  So I now get the error above, because there is no village variable in this data frame.

c) I tried with a third data frame - that does have Village and confirm it is still in operation.  I pressed Reset and that still does not currently reset that sub-dialog.  It should.

Thanks

Vitalis95:
Fixes #8908 
@rdstern @N-thony , have a look at it

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -58,6 +58,13 @@ Public Class dlgCalculator
         SaveResults()
         ucrCalc.ucrSelectorForCalculations.bUseCurrentFilter = False
         ucrCalc.ucrTryCalculator.SetRSyntax(ucrBase.clsRsyntax)
+        ucrBase.clsRsyntax.SetAssignTo(ucrCalc.ucrSaveResultInto.GetText(), strTempColumn:=ucrCalc.ucrSaveResultInto.GetText(),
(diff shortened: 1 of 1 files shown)


rdstern:
@Vitalis95 really pleased you are attacking this one.  However the problem is still there so far:

image

To get this essor I started with the survey data, that includes summary dialogs.
a) First did calculation with field * 2 in position after field.variable.
b) Then moved to summary data frame, then reset and count * 2, without touching the position button.  This gives the confusing error above.
c) If I try the calculation again - still not touching the Position button it is ok.  And the calculated variable is correctly the last one.

Vitalis95:
@rdstern , have a look at it now

rdstern (approved):
@Vitalis95 excellent - I think you have solved this one.  It is also a good indication that R-Instat is becoming more solid!
@N-thony over to you.

Issue #8885, opened 2024-03-19
Adding dialog to export climatic definitions (and data)


lilyclements:
As part of an EPICSA project, we are currently developing a series of functions in [epicsawrap](https://github.com/IDEMSInternational/epicsawrap) which aim to read the climatic definitions from R-Instat and export them to google buckets.

The R code on this is in progress, but we would like to have a front end for it in R-Instat. This should just require one function to be called in the dialog with some simple parameters. I don't think this would be a complex dialog at all, and if someone could start on the design that would be great.

Here is the design:

![image](https://github.com/IDEMSInternational/R-Instat/assets/21180424/a67a36bf-3e25-4251-ac50-49b1ed67c38c)

* **Name:** Export Climatic Definitions
* Then we have a selector
* Six single receivers with the labels: "Annual Rainfall", "Annual Temperature", "Monthly Temperature", "Crop Success Probabilities", "Season Start Probabilities", "Extremes"
* Then a checkbox with label "Include summary data"

@N-thony could you suggest someone who might be suitable for this task?

Vitalis95:
@lilyclements ,  should be added the last item on  `Climatic > File`, that is below `Export to Climpact`?

lilyclements:
@Vitalis95 that sounds suitable - thanks for the suggestion

Vitalis95:
@lilyclements , I am done with the design

lilyclements:
@Vitalis95 this is great, well done! There's two new receivers to add if that's ok:
* Year
* Month

The R Function is called `export_r_instat_to_bucket`. There are a few parameters:

* `data`: value in the data receiver (read as a string)
* `data_by_year`: value in the data_by_year receiver (read as a string)
* `data_by_year_month`: value in the data_by_year_month receiver (read as a string)
* `crop_data`: value in the data receiver (read as a string)
* `rain`: value in the rainfall receiver
* `tmin` : value in the tmin receiver
* `tmax`: value in the tmax receiver
* `year`: value in the year receiver
* `month`: value in the month receiver
* `summaries`: We have `summaries = c(...)` where in `c` we list the items that were checked - `"annual_rainfall", "annual_temperature", "monthly_temperature", "extremes", "crop_success", "start_season"`. Is it ok to be done this way? I can simplify it if you would like to `annual_rainfall = TRUE` if checked, `annual_rainfall = FALSE` if not. Let me know.
* file_path`: The file path in the file saver
[...]

PR #8930, opened 2024-04-12
Added  Climatic > File > Export climatic definitions dialog

Vitalis95:
Still in progress.


lilyclements:
@Vitalis95 if you update `epicsadata`, the issue you previously reported should be sorted. Thank you for finding and reporting that one!

We now have a new problem with exporting to the buckets. We need to authenticiate first. This is achieved using the following code:

```
epicsawrap::gcs_auth_file(" <file_authenticator_token.json> ")
```

where `file_authenticator_token.json` is the `json` file containing the token to authenticate the buckets.

@Vitalis95 @Patowhiz @ChrisMarsh82 I'm not sure the best way to go about this. Do we want to have the ability to authenticate on the main dialog, as a sub-dialog from the main dialog (where OK isn't enabled until it's filled and clicked), or as a separate dialog?

Let me know your thoughts. Thanks!

Patowhiz:
@Vitalis95 a screenshot of the current dialog design would be helpful in answering @lilyclements question.

Vitalis95:
@Patowhiz here is the dialog, the define button leads to sub-dialog. 

This PR  is linked to issue #8885 


image

Patowhiz:
@lilyclements considering the amount of details already on the dialog, I think a sub-dialog would be sensible. 
For Climsoft, we  currently do authentication to the Climsoft database through a sub-dialog.

Vitalis95:
@lilyclements , there is this new error after updating ` epicsadata`

image

lilyclements:
@Vitalis95 thank you for finding this and being patient, once again! I've fixed the problem. Can you update your `epicsawrap` package to get the fix?

`devtools::install_github("IDEMSInternational/epicsawrap")`

rdstern:
@Vitalis95 we (including @jkmusyoka have the script from @lilyclements .  It looks as though the dialog is almost ready.  Is that true.

lilyclements:
@rdstern quick update - @Vitalis95 and I have a meeting tomorrow (it is a public holiday today in Kenya). The only hold-back at the moment is with opening the browser from the dialog, I think. I hope this can be fixed by the `iCallType`, and then it is simply just fixes required on the `TestOK` and it's ready!

If it can't be fixed by the `iCallType`, or even if it is, we might want to have an option to connect to google buckets through the dialog - similar to how we do it on the "Import from Climsoft" dialog, but the user just has to put in the file containing their "key/token" to connect to the buckets.

Vitalis95:
@lilyclements , have a look at the progress so far

PR #8932, opened 2024-04-15
Added Select/Deselect All checkbox to the output window

N-thony:
Fixes partially #7979 
@rdstern I have added a checkbox to Select/Deselect All elements to the output window, have a look
image
image
image

### DIFF
--- a/instat/Model/Output/clsOutputLogger.vb
+++ b/instat/Model/Output/clsOutputLogger.vb
@@ -110,7 +110,7 @@ Public Class clsOutputLogger
 
             'add the output with it's R script as another output element
             outputElement.SetContent("", outputType, strOutput)
-            '_outputElements.Add(outputElement)
+            _outputElements.Add(outputElement)
             'raise event for output pages
             RaiseEvent NewOutputAdded(outputElement, bDisplayOutputInExternalViewer)
 
--- a/instat/UserControl/ucrOutputPage.vb
+++ b/instat/UserControl/ucrOutputPage.vb
@@ -500,9 +500,29 @@ Public Class ucrOutputPage
             Dim startIndex As Integer = _checkBoxes.IndexOf(lastCheckedBox)
             Dim endIndex As Integer = _checkBoxes.IndexOf(currentBox)
 
+            ' Toggle check state for checkboxes between startIndex and endIndex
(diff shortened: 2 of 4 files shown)


rdstern:
@N-thony I had not recognised this from the title.
a) Am I correct that you have also fixed the initial black blob.  At least I didn't get it, when I did my first visualise.  No perhaps not.  It was there for the first pivot table.
b) But I also really like the new feature, with the count of the number selected and also that clicking is also the option to deselect all.  

Would it be easy to add the shift-click to selecting multiple items?  You clicl on one box, then move down and shift-click would select all between the last one selected, and this one?

Then I would be happy to see this merged as it is.  I really like it.

rdstern:
@N-thony I may need instruction, because it seems to be working, but I'm not sure how it is deciding what to include!
I was hoping I could click on a checkbox,  then I move down, or up, and if I shift-click, then it selects all check-boxes between the first and the shift-click one.  
So its action is identical to what happens currently in selecting multiple cells in a column of the grid?

rdstern:
I think the shift-click works as I would expect now - not sure without more testing.  So could accept it.  I wonder whether the converse could also work, namely if I click on an item that is already selected it is noIf I then shift-click anywhere else (selected or not) then everything between is deselected.  Is that easy to implement?

rdstern (approved):
@N-thony it is great.  Another important item ticked off my list of things we should include in the bug release!  It is better than I thought, because I hadn't considered assing a counter!
@Patowhiz over to you to check and, I hope soon merge.

rdstern (commented):
@N-thony one more change I hope, before @Patowhiz checks and merges this improvement.  I selected all and then used Delete.  It works fine, but your new checkbox remains checked and I can't now uncheck it.  It should uncheck when I use delete.  Thanks

rdstern (approved):
@Patowhiz over to you

rdstern:
@Patowhiz could you check and hopefully merge?

Patowhiz (changes requested):
@N-thony, could you please add comments to the newly added functions and global variables in this pull request? Although we haven't consistently done this in the past, I've found that adding comments improves readability and greatly assists with future maintenance. Thank you.

rdstern (approved):
@Patowhiz I like this additional functionality.  Back to you to check anf hopefully merge.

Patowhiz (approved):

Issue #8931, opened 2024-04-12
Tidy the Integer keyboard and add catalan numbers


rdstern:
@Fidel365 Here is the existing integer keyboard.  I am now writing the help for it.

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/f1d283a6-bf11-434b-a32d-8596239cb222)

a) There are 3 sections and I would like to keep that.  But the vertical gaps between the rows within a section are all over the place.  Could you please make it a neater set of keys please.

b) And I have forgotten catalan numbers.  They are also in the Zseq package.  The key is catalan.  The Command is `Zseq::Catalan(n=nrow(x=data))`.  The tooltip is `generates Catalan numbers to the length of the data frame. The 5th is 10!/(5!*6!) =7 * 6 = 42 `.  Put it instead of the fibonacci2 key.  But also take the command for the Fibonacci2 key and use it instead of the code in the first block for fibonacci.  That's `Zseq::Fibonacci(n=nrow(x=data2))`.  The simple tooltip on the current fibonacci key can remain.

c) In the 6 keys, twin, cousin, sexy, third, triplets, k-tuple please change the 100 under max to 1000000 (add 4 zeros).

d) the k-tuple key currently is `rep(x=primes::k_tuple(min=0, max=100, tuple=c(0, 2)), len=nrow(x=data1))`
Change the tuple to become `c(0,2,8)`.  So the command will become
`rep(x=primes::k_tuple(min=0, max=1000000, tuple=c(0, 2, 8)), len=nrow(x=data1))`

e) The code given in the perfect key is wrong:  It currently gives:
`rep(x=Zseq::Perfect(n=10), length=nrow(x=data2), len=nrow(x=data2))`
Delete the len argument.

f)This is more challenging and may need help from @N-thony .  Try the roman key on a data-frame with x1 (say) 1, 2, 10.  If you use Try, then it displays as the Roman numerals and (I am sure, in a past version it displayed as `I, II, ... V... X` in a previous version.  And the Roman numerals are there in the R-viewer.  But in the grid they just display as `1, 2, 3, ...`  Could this be fixed?

g) Finally, for me, the `phi `key gives an error.  Please check that iot also does for you.  The totient function is in the documentation for the primes package, but the function doesn't seem to be there.  Is that the same on your machine?

lilyclements:
Remaining items: (f) and (g)

@rdstern what should we do about the phi problem here?
[...]

PR #8947, opened 2024-04-24
Improvements to the integer keyboard

Fidel365:
Fixes partially issue #8931 
@rdstern , this PR is ready for review, partially because of the issue with the roman button in the grid window which I am still looking at and will include it in the next PR for calculator.
Lastly, the phi function is not valid anymore on my side too.


rdstern (approved):
@Fidel365 looks fine.  @N-thony over to you when you are back.

PR #8961, opened 2024-05-01
Fixed bug when using Hide/Unhide Data Frame Dialogue

### LINKED ISSUE #8956: Hide/Unhide Data Frame Crashing R-Instat
Hello @rdstern and @N-thony, 

I have moved on to the help for Hide/Unhide Dataframe and I have found that it keeps crashing my r-instat. When I hide a data frame it seems fine but then when I go to unhide them it crashes which I add data frames in the Unhide tab.

It has happened three times when I have been using this dialog so I thought I would report it. FYI I am using what I think is the most up to date version of R-Instat thought Visual Studio

N-thony:
Fixes #8956 
@rachelkg @rdstern have a look.

### DIFF
--- a/instat/ucrReceiver.vb
+++ b/instat/ucrReceiver.vb
@@ -69,6 +69,7 @@ Public Class ucrReceiver
     ' If not attached to primary data frame, should only data frames linked to the primary data frame be shown.
     Public bOnlyLinkedToPrimaryDataFrames As Boolean = True
 
+    ' If True then this receiver can only contain hidden data frames from the selector's list
     Public bOnlyHiddenDataFrames As Boolean = False
 
     ' When calling GetVariableNames() this is the R package & function name the variables will be inside


rdstern (approved):
@N-thony that now seems to work fine and a good bug to have resolved!  @rachelkg may not have time to check just now.  @Patowhiz if it is ok then can you merge.

rdstern (commented):
@N-thony as @Patowhiz has yet to check and merge this one, could you just check the help context number.  It should be 538?
I am still approving so we get it merged.  Maybe @Patowhiz could check/change this also?

instat/ucrReceiverMultiple.vb
```diff
@@ -180,12 +180,15 @@ Public Class ucrReceiverMultiple
     End Sub
 
     Private Function ShortenString(strText As String) As String
-        Dim maxLength As Integer = 10
-        If strText.Length > maxLength Then
-            ' Trim the string to the specified length and add ellipsis
-            Return strText.Substring(0, maxLength) & "..."
+        Dim strShortText As String = ""
+        If Not String.IsNullOrEmpty(strText) Then
```
Patowhiz, line 184: I understand that this function is used in on place(line 176). What makes/causes name of the group to ever be Null? Is there a situation where the group is created without a name?

Patowhiz: What do you mean by "just adding the values and not the keys on line 51"? The data structure is `KeyValuePair` which means it must have a key. Line 51 adds a `KeyValuePair` to a list of `KeyValuePair`. 

What scenario will make key of the `KeyValuePair` be `Nothing`? That should never happen because the list view control of the multiple selector expects each list view item to belong to a `Group`.

Patowhiz: @N-thony,

I did review the `AddMultiple` function that's why I'm concerned why the `KeyValuePair` keys can be `null`. This scenario should ideally never occur, as each item in the list view control of the selector is supposed to be tagged—this tagging is crucial for grouping items displayed by the multiple receiver.

The presence of `null` tags suggests a developer oversight that needs thorough investigation. Implementing a `null` check as a bug fix implies an expectation of encountering untagged items, which contradicts the design intent of always having tagged list view control items. Have a look at `FillListView` subroutine in the `clsRLink` to understand the design intent. `FillListView` is used by the `ucrSelector` in line 159 to populate itself with different "kinds" of contents.

Before making any changes, it would have been beneficial to investigate why the tag is `null`, particularly when using `dlgHideDataFrame`. I've looked into this for you, and the problem originates from line 148 of `dlgHideDataFrame`. However, I recommend against simply correcting this line in isolation. The entire dialog likely needs refactoring. For instance, the `SetHiddenColumns()` subroutine is redundant since its functionality is already covered by the selector. Furthermore, the subroutine’s name does not accurately reflect its purpose given that the dialogue primarily deals with data frames.

Just like I have mentioned to @rdstern before, I strongly advise against patchwork bug fixes that do not consider the broader context and how components interact. Our goal should be to enhance the overall quality of the codebase, ensuring robustness and coherence in our product's functionality.

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -1193,11 +1193,12 @@ DataBook$set("public","set_hidden_data_frames", function(data_names = c()) {
 } 
 )
 
-DataBook$set("public","get_hidden_data_frames", function() {
+DataBook$set("public","get_hidden_data_frames", function(as_list = FALSE, ...) {
```
Patowhiz, line 1196: @N-thony 

I recommend modifying the `get_data_frames` function to allow filtering by hidden or non-hidden data frames. This function currently includes some unused parameters that align with those found in `get_column_names`.

By updating `get_data_frames` appropriately, we can minimize the changes needed on the front end. We could then phase out `get_hidden_data_frames` by either updating the blocks that use it to utilize the revised `get_data_frames`, or by having it directly invoke the updated `get_data_frames`.

Please review the implementation of `get_data_frames` and compare it to how `get_column_names` is utilised, particularly the `include` and `exclude` parameters which are modified by the receiver. I believe adopting a similar approach for `get_data_frames` was likely the original author's intent. This would also eliminate the need for adjustments to the `clsRLink.FillListView` subroutine consequently eliminating the need of introducing more global variables to manage internal states.

Patowhiz (changes requested):
@N-thony have a look at how the selector determines

rdstern (approved):
@N-thony seems to work ok now.  @Patowhiz over to you.

instat/static/InstatObject/R/instat_object_R6.R
Patowhiz: @rdstern and @N-thony, if the fix is urgent and my earlier suggestions for a coherent resolution present technical challenges that would require significant effort from @N-thony to address, then I recommend making adjustments to line 148 of `dlgHideDataFrame`. This would be a patchwork fix, but it would resolve the bug and keep the changes confined within the dialog itself.

Patowhiz: @N-thony, I'm not sure what you mean by "I will work on this in a separate PR once this is merged."

I suggest leaving this PR as it is and creating a new PR to make adjustments to line 148 of `dlgHideDataFrame`. I will then merge that new PR as a patchwork fix to resolve the bug, keeping the changes confined within the dialog itself.

Currently, I can't merge this PR because it includes changes to multiple files that, in my opinion, do not collectively resolve the broader issue. These changes introduce potential coherence and robustness issues in how components interact with the backend.

Patowhiz: @N-thony apologies for the confusion. I meant` get_data_names` instead of `get_data_frames` (which doesn't exist). Thank you for pointing out the confusing suggested function name.

Please work on implementing `get_data_names` as discussed, and then we can proceed with merging the PR. Alternatively, you can edit this PR or make a new one that makes adjustments to line 148 of `dlgHideDataFrame`. That should be a quick task that will patch the outlined bug.

PR #8965, opened 2024-05-02
Displaying results as a new data frame when in scale option

### LINKED ISSUE #8952: Scale and Distance dialog
Hello @rdstern and @N-thony, 

I have been working on the help for the Prepare > Data Reshape > Scale/Distance dialog and I have some question about how it is working. 

In the Help it looks as though when you press ok it should create a New Data frame, and in fact the dialog has a New Dara Frame Name box. However, when I use the dialog it presents the result in the Output window. @rdstern please clarify how you would like this to function. Where should the results display, a new data frame, the output window or both (in this case an extra item will need to be added to the dialog)?

Also the first time I tried to use this dialog I kept pressing ok and nothing happened until i closed and reopened R-Instat. I have not had this happen again so I don't know if it is something to be concerned about but wanted to mention it in case it was symptomatic of a problem.

Vitalis95:
Fixes #8952 
@rdstern @N-thony , this a minor change which display results as a new data frame when in scale option


rdstern (approved):
@Vitalis95 looks fine now, thanks.  @N-thony over to you

Issue #8905, opened 2024-03-31
Add a functions keyboard to the calculator


rdstern:
@Fidel365 I am going to take advantage that you have become very effective in improving the R-Instat calculator.  

This is not ready for work yet, and I hope that @lloyddewit will also be interested.  It builds on a discussion topic, #8034 plus work "completing" the complex keyboard.  

It will consider functions now in 2 ways.  From an R-computing perspective and also (now) from a maths perspective.  

I would like to move the pascal key from probability keyboard into this keyboard.  And I would like to add the polyroot function.  This is in base or utils and gives the roots (maybe complex) of any polynomial.  It provides a variable that is 1 value less than the number of coefficients so we give it as:
c(polyroot(x1),NA).  Maybe we could alternatively add 0 instead of NA?
Then there is the fundamental theorem of algebra that relates the roots to the coefficients.  This is provided by:
sum(combn(roots, m=2,FUN=prod))
This is usually real so we just give the summary number:
Re(sum(combn(roots, m=2,FUN=prod)))

we repeat sum( ), and we have omitted prod(  ,na.rm=TRUE) from the summary keyboard.  So we include it here.  There is a summary section on this keyboard.

I would like to include minimisation of functions here too, partly because Nelder-Mead is still popular - and I knew them both well!  Maybe not - there is even a neldermead package, but this could take too long!

Fidel365:
@rdstern , can I start working on this or it's not yet ready?

rdstern:
Ok, I am starting it now.  I suggest 4 keys across.  The first row is:

`ave, ave_fun, ave_fac, ave_both`

The full ave function in the standard stats package is [described here](https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/ave).

As the keys, 
a) ave, without parameters gives `ave( )`, with parameters it gives `ave(x=  )`

b) ave_fun without parameters gives `ave(  , FUN=mean)`, 
with parameters it gives   `ave(x=   ,FUN=function(x) mean(x, na.rm=TRUE))`

c) ave_fun without parameters gives` ave(  ,   )`, with the cursor before the first comma. With parameters it is `ave(x=  ,  )`

d) ave_both gives `ave(  ,   , FUN=mean)` without parameters.
with parameters it gives   `ave(x=   ,   , FUN=function(x) mean(x, trim=0.1))`

Nice and easy I hope?

Second row of keys:

`roots, algebra, sum, prod`, see above


Third row:
[...]

PR #8968, opened 2024-05-03
Adding functions keyboard to calculator

Fidel365:
Fixes partially issue #8905 
@rdstern this is ready for review execept for the following comments;1.We don't have coeffs2 key working yet, and 2.I am missing tooltips for the first set row of keys

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -27,7 +27,7 @@ Public Class dlgCalculator
     Private iBasicWidth As Integer
     Private strDefaultKeyboard As String
     ' Note: This list needs to be updated when a new keyboard is added.
-    Private strKeyboards() As String = {"Functions","Basic", "Maths", "Logical and Symbols", "Transform", "Summary", "Probability", "Factor", "Text/Strings (Character Columns)", "Dates/Times", "Circular", "Wakefield", "Goodness of Fit", "List", "Complex", "Integer"}
+    Private strKeyboards() As String = {"Basic", "Functions", "Maths", "Logical and Symbols", "Transform", "Summary", "Probability", "Factor", "Text/Strings (Character Columns)", "Dates/Times", "Circular", "Wakefield", "Goodness of Fit", "List", "Complex", "Integer"}
 
     Private Sub dlgCalculator_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -27,7 +27,7 @@ Public Class dlgCalculator
     Private iBasicWidth As Integer
     Private strDefaultKeyboard As String
     ' Note: This list needs to be updated when a new keyboard is added.
-    Private strKeyboards() As String = {"Basic", "Maths", "Logical and Symbols", "Transform", "Summary", "Probability", "Factor", "Text/Strings (Character Columns)", "Dates/Times", "Circular", "Wakefield", "Goodness of Fit", "List", "Complex", "Integer"}
+    Private strKeyboards() As String = {"Functions","Basic", "Maths", "Logical and Symbols", "Transform", "Summary", "Probability", "Factor", "Text/Strings (Character Columns)", "Dates/Times", "Circular", "Wakefield", "Goodness of Fit", "List", "Complex", "Integer"}
 
     Private Sub dlgCalculator_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
@@ -158,6 +158,9 @@ Public Class dlgCalculator
 
     Private Sub ucrInputCalOptions_NameChanged() Handles ucrCalc.NameChanged
         Select Case ucrCalc.ucrInputCalOptions.GetText
+            Case "Functions"
+                Me.Width = iBasicWidth * 1.5
+                ucrBase.iHelpTopicID = 439
             Case "Maths"
                 Me.Width = iBasicWidth * 1.38
                 ucrBase.iHelpTopicID = 126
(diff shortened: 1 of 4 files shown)


rdstern:
@Fidel365 what an excellent start.  
a) The keyboard, in the drop-down, should not be the second one.  It should be last in the list.  
b) Tooltip for ave: Gives the mean, as a column.  For example ave(c(1,2,3,4,10)) = 4, 4, 4, 4, 4.
c) Tooltip for ave_fun: As ave key, but change mean (in the function) to any summary, e.g. sd, for std dev or raster::cv for coefficient of variation.
d) ave_fac: In the function add a space after the comma to make it clearer.  Tooltip: Gives the means for one or more factors.
e) ave_both: Add a space before the first comma and another space before the second.  Tooltip: Gives any summary measure for one or more factors.
f) coeffs: You have repeated the ave function.  Instead use `coeffs: polynom::poly.calc( )` , with arguments (x= ).
g) prod, combn, coef keys - also still ave function!  See the 4th row of keys in the issue.
h) coeffs2 : That was a long way further down in the issue.  I wondered if you would look?  It is:
`as.numeric(purrr::map(.x=(nrow(data_RDS)-1):0, .f = ~sum(combn(rootsx3[1:nrow(data_RDS)-1], .x , FUN=prod))))`
Note here the data_RDS, which needs to give the name of the current data frame.  (That's in a lot of the other keys, e.g. prob keyboard, random sample keys.)
i) dsum is again currently your favourite ave code.  See the issue row 5.
j) Swap the order of the fractions and decimals keys.
k) Tooltip for Pascal Gives Pascal traingles, e.g. for c(1,2,3,4) gives 1, (1,1), (1, 2, 1), (1, 3, 3, 1).

rdstern (commented):
@Fidel365 this is looking very good now.  I can see it isn't yet ready, so just a single comment for now that digitsum is the same as the old dsum.  So you don't need the lonely key on the bottom row any more.

rdstern (commented):
@Fidel365 impressive and almost there!
a) The polynomial key is mis-spelt.  add the o!
b) My change as I write the help! Could you move rows 3 and 4 (starting polynomial and prod) to the bottom.  They are currently rows 3 and 4 and should become rows 6 and 7 in the keyboard.
c) The digitsum key needs correction.  Get rid of the dsum:, so it starts with `sapply`.  Also `Function becomes function - lower case f`
d) the digitsqu has a number of errors.  get rif of the capital letters in function and split, get rid of the space between < and -, and get rid of the brackets after split.

Incidentally it would be good if you could also test that these run, before handing them to me.  Make a File New Data Framew, with 1000 rows and then try your keys on x1, once you have corrected them.  A good developer must also be a good user.

e) digitssq - almost works.  Add a `", so split = ", becomes split = ""`

I hope that's it.  Exciting new keyboard.  Well done.

rdstern (commented):
@Fidel365 it is getting there.  I looked first at the pull request you did earleir in the evening, and felt it had retreated some steps!
This is good, except for (I hope) 3 buttons.
a) coef:  Here is the code, completed, that works:

`sum(combn( x4[1:nrow(data_RDS)-1],4 ,FUN=prod, na.rm=TRUE))`

Here is what you have currently:

`coef: sum(combn( , ,FUN=prod, na.rm=TRUE))`

Get rid of the `coef:`.  That was just me saying which key it relates to.  It has gone in the other keys and needs to go here too.
Now I have realised we can't use the whole variable, because the last value is NA.  It works but gives the wrong answer.  So, before the first comma please add [1:nrow(dataframename)-1]  Here you have to get the actual data frame name, to insert, which is done in various other keys - e.g. random numbers in probability keyboard.
The cursor is just before the square bracket.  Then I suggest you could include 0 (zero) after the comma.  That will always be ok.  The user can change it to 1, 2, 3, 4 if needed. 

b) The coeffs key.  Here is a working version
`polynom::poly.calc(x4[1:nrow(data_RDS)-1] )`

Your current version gives:
`coeffs: polynom::poly.calc( )`
So, as above get rid of `coeffs:`

And add the same [1:nrows  stuff as in the coef above.

c) Coeff2 looks very odd.  Please see if you can figure out what to do from the instructions in the issue.  You seem to have hard-coded the name of a data frame, there is an odd mean=0, sd=1, which is curious.

Roger

Fidel365:
Still more work tomorrow morning.

Fidel365:
@rdstern I have made the changes requested here and this is ready for review

rdstern (changes requested):
@Fidel365 great, I have run the functions code that should go inyo the library, and am checking each key.  
a) Key 4 row 2, I have changed the name of the function to match your key.  No change from you, but I will send the new set of functions to you.
b) Key 4 row 4, change your code to match the key.  The key says digitssq, while the current code is dssq(  ).  Make it digitssq(  ).
c) Row 6 we produce polynomial objects.  We have a "rule" in the R-Instat data sheet, that objects with an odd class have an (S) after the name.  This may need some input from @N-thony?
d) In combn change and simplfy the tooltip.  So change 2 to 3 in the example to match the calculation.  Then omit the "so part" which is confusing in the tooltip, because the multiplication is missing - and good to make it shorter anyway!  
e) In coef, please add Re(   ) round the whole thing, so the result given is not complex.
f) The coeff2 seems to work fine - well done you!

Fidel365:
@rdstern this PR is ready for rereview.

Fidel365:
@rdstern this is ready for rereview.

rdstern (commented):
@Fidel365 well done, and almost there on some impressive stuff.
a) Thre polynomials are now (s), great!
b) ave(fun puts the cursor in slightly the wrong place in the expression.  It is before the bracket.
c) digitssq* is misssing - nothing there when I press the key.
d) the tooltip for combyn should be product of the values 2 at a time (not 3)

Then looking forward to merging

Fidel365:
@rdstern and now ready again.

rdstern (approved):
@Fidel365 that's great.  Really great.  @N-thony over to you.

Issue #8872, opened 2024-03-12
Improve running of scripts generated via the calculator.


rdstern:
@Patowhiz I think you may be able to explain the following quickly, as I assume the code is yours.  Then could either you, or Antoine make the corrections?

@N-thony there are some problems running scripts, when the script is generated from the Prepare > Column Calculator.  They work, but can be very slow.  They also generate (extensive results in the output window that are not generated when using the calculator from the dialog.

Here is the script file:

[liouville.zip](https://github.com/IDEMSInternational/R-Instat/files/14565851/liouville.zip)

It is to be used in a practice and video.  I give the first draft here, in case it is of interest to see the context  But I give a shorter code below that can be used directly.

[Are you a natural mathematician.docx](https://github.com/IDEMSInternational/R-Instat/files/14565791/Are.you.a.natural.mathematician.docx)

Here is the start of the script:

```
# Dialog: New Data Frame

liouville <- data.frame(x1=as.numeric(rep(seq(1,100000), each=1, length.out=100000)))
data_book$import_data(data_tables=list(liouville=liouville))

rm(liouville)
# Dialog: Calculations

liouville <- data_book$get_data_frame(data_name="liouville", use_current_filter=FALSE)
attach(what=liouville)
factors <- conf.design::factorize(x1 )
data_book$add_columns_to_data(data_name="liouville", col_name="factors", col_data=factors, before=FALSE)
data_book$get_columns_from_data(data_name="liouville", col_names="factors")
liouville <- data_book$get_data_frame(data_name="liouville", use_current_filter=FALSE)
detach(name=liouville, unload=TRUE)
data_book$append_to_variables_metadata(data_name="liouville", col_names="factors", property="labels", new_val="")
rm(list=c("factors", "liouville"))
# Dialog: Calculations

liouville <- data_book$get_data_frame(data_name="liouville", use_current_filter=FALSE)
attach(what=liouville)
length <- sapply(  factors ,length)
data_book$add_columns_to_data(data_name="liouville", col_name="length", col_data=length, before=FALSE)
data_book$get_columns_from_data(data_name="liouville", col_names="length")
liouville <- data_book$get_data_frame(data_name="liouville", use_current_filter=FALSE)
detach(name=liouville, unload=TRUE)
data_book$append_to_variables_metadata(data_name="liouville", col_names="length", property="labels", new_val="")
rm(list=c("length", "liouville"))

```
[...]

PR #8972, opened 2024-05-08
Fixed the error caused by attach and detach functions when there is no data in Calculator

N-thony:
Fixes partially #8872 
@rdstern have a look.

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -86,8 +86,6 @@ Public Class dlgCalculator
         clsAttachFunction.AddParameter("what", clsRFunctionParameter:=ucrCalc.ucrSelectorForCalculations.ucrAvailableDataFrames.clsCurrDataFrame)
         clsDetachFunction.AddParameter("name", clsRFunctionParameter:=ucrCalc.ucrSelectorForCalculations.ucrAvailableDataFrames.clsCurrDataFrame)
         clsDetachFunction.AddParameter("unload", "TRUE")
-        ucrBase.clsRsyntax.AddToBeforeCodes(clsAttachFunction, 0)
-        ucrBase.clsRsyntax.AddToAfterCodes(clsDetachFunction, 0)
         ucrBase.clsRsyntax.SetCommandString("")
 
         ucrCalc.ucrSaveResultInto.SetPrefix("calc")
@@ -112,7 +110,6 @@ Public Class dlgCalculator
 
     Private Sub ucrCalc_SaveNameChanged() Handles ucrCalc.SaveNameChanged
         SaveResults()
-        TestOKEnabled()
     End Sub
 
     Private Sub ucrCalc_ControlValueChanged() Handles ucrCalc.ControlValueChanged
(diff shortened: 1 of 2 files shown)


rdstern:
@N-thony it is looking good and I especially like the way it now works cleanly also when there is no data frame.
With a data frame the one bit I still don't like is here:

image

With the survey I do a simple calculation calc <- fert * 2, which is fine with Ok.  But when run from the script the data_book$get_columns_from_data puts the result correctly into the grid, but also gives a copy in the output window.  Could there be a small change in that function, so it only outputs to the grid?

rdstern:
@N-thony great.  What do you suggest?  
a) Should `To Script `therefore not generate those lines, when using the calculator?
b)  or just generate them, but with the comment `(#)` added?  
c) Or is that a job for the user (to add the comment) when running from the script?

I wonder if this is just a `prepare > calculator `script problem or also in other dialogs?

rdstern:
@N-thony now it seems to work fine from the script window.  

But, when I do fert * 2 from the survey data in the calculator and don't save, so it goes into the output window I now get:

image

I thought it was ok in the output window before?

rdstern:
@N-thony this is all still a big improvement, so should I approve, so we can merge so far, and then discuss further?

rdstern (approved):
@Patowhiz I am really happy with the changes so far.  There is still a small question for a follow-up issue, but I would like to see this merged as soon as you are happy with the change.  Thanks

instat/dlgCalculator.vb
```diff
@@ -131,24 +127,45 @@ Public Class dlgCalculator
         Else
             ucrBase.clsRsyntax.RemoveFromAfterCodes(clsRemoveLabelsFunction)
             ucrBase.clsRsyntax.RemoveAssignTo()
-            ucrBase.clsRsyntax.iCallType = 5
+            ucrBase.clsRsyntax.iCallType = 1
```
Patowhiz, line 130: I recommend removing the dependency on `iCallType`. Currently, within the `clsRLink.RunScript` subroutine, only the calculator dialog relies on `iCallType` to determine script execution. Initially, all dialogs were required to set `iCallType`, but we later shifted towards a model where scripts are executed in the same manner as they would be through the script window. This change is detailed in issue #7778.

If you can be able to use an approach that removes the dependency on `iCallType` we can then remove the code block in line 980 of  `clsRLink.RunScript` subroutine. This will improve the robustness and coherence of our base code. 

If doing the above needs more time, I'm happy to merge this as it is and later look into it.

Patowhiz (changes requested):

instat/dlgCalculator.vb
Patowhiz: @N-thony I'm not sure what you mean by " I am afraid this can be the same with other dialogues?"
Did you have a look at my suggestions above? If you can address them then I'll merge this. In particular evaluating the time that it could take to remove the `iCallType` dependency for reasons outline in the issue I mentioned.

Patowhiz: No it was not correct, I agree with the changes made on line 312 of `clsRSyntax` in this PR.

Patowhiz: @N-thony I've seen that you have reverted the `iCallType` change.

Have you evaluated removing the dependency on `iCallType`? I've been wishing that for a long time, but if you feel that it will take a lot of your time, I'm happy to merge this as it is.

rdstern (commented):
@N-thony I am really happy with the results when there is no data in R-Instat.

But once there is a data frame - survey for example,  In Version 7.6 I get this:

image
  Not nice and not well formatted.

In the current pull request I get:

image

So numbers well formatted, but do I have to also get the output below the data.  That'll be off-putting to R-Instat users - and me!

rdstern:
Thanks @N-thony I look forward to the results from your discussions.
There is a small - but would be very useful task hidden in #8931 which may need similar discussions?.  This is to display roman numerals in a data frame.  There is a calculator key (in the integer keyboard), and they display nicely as roman numerals in the Try and also in the R viewer.  But, in the data frame, they are transformed back into ordinary numbers, which destroys the point of showing roman numerals!

rdstern (approved):
Forgot this one.  @Patowhiz could you merge please!

PR #8978, opened 2024-05-16
Gt Table Dialogs

Patowhiz:
Still draft.


rdstern:
@Patowhiz was I supposed to start testing?

image

I first tried with these data and don't get any table?

Maybe you could take me through an example?

Could you include the sub-dialog again to the Summaries dialog?

Patowhiz:
@rdstern yes you can start testing features implemented. I'm now making changes to the stub and and adding a row expression sub-dialog(so anywhere you see a row expression, the user will have the ability to write the expression through a sub-dialog - just like the calculator). I'm also adding styling for the individual features.

I'm honestly surprised why you are unable to get the dialog, the code in the output window looks good, do you get results when you run the code through the script window?

I'll update the examples that demonstrate each feature in a follow up comment after I finish the pending features.

Like I mentioned before, it's not quite possible to add it in the summaries dialog because it uses our custom controls that require a data sheet that is in the data book. The summaries dialog produces a new data frame internally, this data frame has to be added in the data book for the presentation dialog controls to use.

rdstern:
@Patowhiz here is your list of remaining items:

1. Nano plot.
2. Column summaries
3. Hiding of columns.
4. Styling columns.
5. Styling stub.
6. Styling footers.
7. Styling summaries.
8. Cleaning and documenting the code.

I hope you are happy to complete the above, and then we look, with @volloholic and perhaps @lilyclements on how to connect this all with the summary dialogs.  Ideally perhaps this could start on the next Thursday meeting?

I have started to check what's there so far and it looks clean and impressive.  I wonder about having Nano Plots as an additional tab, rather than within Columns, and also moving Hide there?  I'm not sure we'll need to hide much (because we can select the variables for the plot) except when we have the plots?

Patowhiz:
@rdstern I've had to disable several GT functionalities of the row summaries because they are not supported by the R package used by R-Instat (R 4.1.2/3)

Patowhiz:
@rdstern 

I'm disabling nanoplots and certain row summaries functionalities for now because it is only supported in R 4.2.0 and above. Current R-Instat ships with R 4.1.3.
I believe we would still have to evaluate other dialogs using different packages when upgrading? For instance I noticed several breaking changes in the GT R package.

I also didn't see the need for hiding columns.

This is now ready for your review.

rdstern:
Thanks for the update.  I also don't see the need for hiding variables until we have nanoplots.  Then the need will arise, because we may wish to include a set of variables to produce the plot, but then we sometimes don't need them as well.

I see you have nop choice but to disable nanoplots if they really need R Version 4.20.  I assume you are sure about this?

Assuming so, then we return to the issue of updating the version of R and hence only catering for 64 bit machines.  I would like to re-open this discussion once @ChrisMarsh82 is back from leave.

We decided to delay updating R, because there was nothing important in the recent versions that is needed for R-Instat.  Now there is!  R us currently Version 4.4.

We have - I hope recently sorted out most of mthe recent regressions.  We still should continue, but then I wonder about a last Version 0.7.x that is for 32 bit machines.  From then on - perhaps in one month?  we update R and just support 64 bit?

Patowhiz:
@rdstern I agree, we can decide to freeze any new features now, put effort in testing and cleaning (including cosmetics) all that is in R-Instat. To make sure we have a stable 32 bit version. Key word stable.

rdstern (changes requested):
@Patowhiz this looks really, really good.  So we should merge as soon as we can - and before we finalise the 32 bit 
c) Add the Table Options button to the `Prepare > Data Frame >View Data `dialog.  Specify the issue - whether on both buttons or just the Selected Columns.  Decide also whether you will quickly also add this yourself or allocate.
d) On the titles (header tab )see below) the fields should each be considerably wider:

image

e) On the Columns tab, see below, only include the Columns in the data selector that have been selected for the table; so those on the main dialog.

image

f) @N-thony please can you check the code.  I'd like to merge pretty quickly and continue checking steadily.  I don't see the need to check each option before an initial merge.

g) Going back repeatedly I get the following when pressing on options:

image

h) In the font I don't thisn the size, sty;e, etc are working yet.  The size starts with NULL.  Could that be a pull-down?

image
:

i I need to check the save button on the maximise dialogs and also saving (including multiple tables from the output Window(s).  However, when I do a Copy/Paste from the Maximised output it already goes strainght into Word as a table.  Nice!

rdstern:
@Patowhiz I have just gone into the styles for the month column and it has "remembered the serring from the title.  Shouldn't it be separate.?

image

A more major point is that gt says it needs R >= Version 3.60.  And =that includes nanoplots of course.  Are you sure you can't add nanoplots, etc in the current version?

rdstern:
On your comments:
a) Showing just the columns used in the table.  For now we can do a select before running the table dialog.  Then I assume we'll just have those columns.  I'll try.
b) Agreed to merge this PR before adding the code to other dialogs.
c) Still surprised that we will need to update R, when it says it only needs R Version 3.6 or higher.  But happy with our plan to freeze a 32 bit version soon and then update R.
NowI'll try your improvements - exciting!

rdstern:
Really happy so far.
a) Made a nice title.  Then have returned to the dialog and get this developer error when I pressed the Format button again.  (It doesn't stop anything, so I am continuing.)

image

(I have now done a lot more and not got this again!  We got it a lot more often before.)

b) Size now works on the title - great.

c) The title of the dialog is just Table.  Have it as Presentation Table instead.

Patowhiz:
@rdstern this is now ready for your review.
I could not reproduce the bug you found in item (f) .

rdstern (approved):
@N-thony over to you to check and hopefully merge soon

PR #8986, opened 2024-05-23
Implemented the drop unused levels checkbox in the Column Summaries dialog

### LINKED ISSUE #8730: Implement the Drop Unused Levels checkbox in the Column Summaries dialog

N-thony:
Fixes #8730
@rdstern I made the change as requested for the drop unused levels checkbox. @lilyclements can now make the change in the calculation system.

### DIFF
--- a/instat/dlgColumnStats.vb
+++ b/instat/dlgColumnStats.vb
@@ -123,7 +123,6 @@ Public Class dlgColumnStats
         ucrBase.clsRsyntax.SetBaseRFunction(clsDefaultFunction)
         bResetSubdialog = True
 
-        ucrChkDropUnusedLevels.Enabled = False
     End Sub
 
     Public Sub SetRCodeForControls(bReset As Boolean)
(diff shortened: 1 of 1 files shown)


rdstern:
@lilyclements I am under the impression you have the change in R, but had "hard-wired it, in your version to test.  I hope it could now be included in this pull request, as the condition that the by variables have to be factors is now there.  The checkbox should be disabled if they are not.

rdstern:
@lilyclements can you incorporate your new code in this pull request, so the option can be completed?

rdstern:
@lilyclements are you ok to add your magiv here.  You said you had the solution, but "hardwired".  Note it will be needed again later in the Tables dialog.

lilyclements:
> @lilyclements are you ok to add your magiv here. You said you had the solution, but "hardwired". Note it will be needed again later in the Tables dialog.

@rdstern as you pointed out the changes are in #8794 - thanks!

rdstern:
@lilyclements I am unable to test changes just now (misplaced laptop!).  It would be great if this now works.  Could you test?

rdstern:
@lilyclements has agreed to test this.  Then, or in parallel, could @Patowhiz check the code and merge.  This is an important new feature in the calculation system.  (It is also helping an improvement in the climatic dialogs.)

rdstern:
@N-thony if I understand correctly, then the check unused levels should not be disabled.  Let me give an example to check the different options now are possible,
In the survey dataset, if you summarise yield by both the variety (3 levels) and village (4 levels), then you get a summary data frame with either 10 or 12 (3 times 4) rows.  

image

In the past there were always 10 rows, because 2 combinations have zero frequencies.  (NEW variety is only in 2 of the 4 villages.)  

image

Now, with the @lilyclements and your changes, the user can choose - depending on the state of the ``Drop Unused Levels ``checkbox.

So if you don't drop the unused levels then I think you should now get the following table.  

image

I suggest the 10 rows remain the default, so the Drop Unused Levels checkbox is checked by default.

lilyclements (approved):
@N-thony looks good! Approving now, @Patowhiz over to you

rdstern:
@Patowhiz this is a small, but important improvement.  I'll explain when we meet.  It links to "your" tables too!  Hope the code is ok, so it can be merged.

rdstern:
@Patowhiz I am back testing again and very happy with this new feature.  Over to you to check and merge.

Patowhiz (approved):
Thanks for reviewing this @lilyclements

Issue #8990, opened 2024-05-25
Add Help context number to the Help > Packages Documentation dialog


rdstern:
@Fidel365 it is the bottom dialog in the Help menu and the context number is 695.
You can't test it, because the help entry is missing in the current help file.

Fidel365:
@rdstern an @N-thony this too

PR #8991, opened 2024-05-27
Adding Help context number to the Help > Packages Documentation dialog

Fidel365:
Fixes issue #8990 
@rdstern I have tested this with a different context that is valid in the current help file and it is working well, hence, this will work once the file is updated


rdstern (approved):
Thanks @Fidel365 I am approving and assuming there isn't much for @N-thony to check.  We can only test when I provide the new Help file in the next version, but it isn't a major problem if it still needs to change later.

PR #8999, opened 2024-06-03
Added scalar object to dataframe metadata

N-thony:
@rdstern can you do the first test here before moving to the next step? The feature is available in the calculator dialogue.

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -301,6 +301,9 @@ Public Class dlgCalculator
         ManageScalarStorageAndAttachDetach()
     End Sub
 
+    Private Sub ucrCalc_TryCommadClick() Handles ucrCalc.TryCommadChanged
+        ManageScalarStorageAndAttachDetach()
+    End Sub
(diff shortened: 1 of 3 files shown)


rdstern (commented):
@N-thony that's a wow to me that quite a lot is working already.  Well done.  
a) I used survey with 2 other data frames as summaries.
b) I was then able to define another scalar and save it.  Perfect.  And see it in the sheet metadata.
c) It still also rean the code to make it a vector, so I presume that's the next step?
d) I also could also define only 1 scalar.  The next time it didn't add another one.

rdstern:
@N-thony that's good progress as I can now define multiple scalars.

image

a) Position of the keyboards is messed up.  I assume that's a single item to be change and they will revert.
b) The Position button is only disabled the first time it is used.  Later it remains enabled.

I am thinking for now you might comment out the lines that produce nrows and ncols, as they might ncomplicate the code later and so make it take longer.  Unless you feel it is simpler, because the sclars are then always there?

PR #9006, opened 2024-06-05
Added scalar object to the metadata

N-thony:
@rdstern this replaces #8999. Have a look

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -110,8 +110,8 @@ Public Class dlgCalculator
         ucrBase.clsRsyntax.AddToBeforeCodes(clsAttachFunction, 0)
         ucrBase.clsRsyntax.AddToBeforeCodes(clsAttachScalarsFunction, 1)
 
-        ucrBase.clsRsyntax.AddToAfterCodes(clsDetachFunction, 0)
-        ucrBase.clsRsyntax.AddToAfterCodes(clsDetachScalarsFunction, 1)
+        ucrBase.clsRsyntax.AddToAfterCodes(clsDetachFunction, 1)
+        ucrBase.clsRsyntax.AddToAfterCodes(clsDetachScalarsFunction, 2)
         ucrBase.clsRsyntax.SetCommandString("")
 
         ucrCalc.ucrSaveResultInto.SetPrefix("calc")
@@ -137,10 +137,10 @@ Public Class dlgCalculator
     End Sub
 
     Private Sub checkBoxScalar_CheckedChanged()
-        GetScalarValue()
+        SetItemType()
     End Sub
 
-    Private Sub GetScalarValue()
+    Private Sub SetItemType()
         If Not String.IsNullOrEmpty(ucrCalc.ucrSelectorForCalculations.strCurrentDataFrame) Then
             If ucrCalc.ucrSelectorForCalculations.checkBoxScalar.Checked Then
                 ucrCalc.ucrReceiverForCalculation.strSelectorHeading = "Scalars"
@@ -167,7 +167,7 @@ Public Class dlgCalculator
     Private Sub SaveResults()
         If ucrCalc.ucrSaveResultInto.ucrChkSave.Checked AndAlso ucrCalc.ucrSaveResultInto.IsComplete Then
             clsRemoveLabelsFunction.AddParameter("col_names", Chr(34) & ucrCalc.ucrSaveResultInto.GetText() & Chr(34), iPosition:=1)
-            ucrBase.clsRsyntax.AddToAfterCodes(clsRemoveLabelsFunction, 1)
+            ucrBase.clsRsyntax.AddToAfterCodes(clsRemoveLabelsFunction, 3)
             ucrBase.clsRsyntax.bExcludeAssignedFunctionOutput = True
             ucrBase.clsRsyntax.iCallType = 0
         Else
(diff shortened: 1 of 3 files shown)


rdstern:
@N-thony great that it is working again.
a) The other keyboards are ok again now.
b) When I save a scalar, the Position is still enabled and it also presents the saved column.  Aha, maybe the scalar part has to be added back in again, because no scalars are produced.  
c) That would be consistent with the fact I don't see thenm in the data frame window.
d) But I see that you have already added Scalars to the Rename objects dialog.  Great.  Not yet to the delete objects.

So excellent so far.

rdstern:
@N-thony I found

a) The first time I had the scalar checkbox as before and could save the scalar.  But the Position wasn't disabled, and I still got the column generated.
b) Then, when I went back to the dialog the scalar checkbox was absent from the dialog:

image

c) I also looked in the sheet metadata and there were no scalars.

d) Great that the scalar option is also added to the Rename and Delete and I look forward to trying once I can find the scalars.

rdstern:
@N-thony the scalar checkbox is now there consistently.
a) Position is not disabled when I check to save as a scalar.  
b) When I save as a scalar I still get a variate of that name.
image


c) Having saved as a scalar, when I check on scalars, there are none.
d) there is also no hint of scalars in the data frame metadata.

rdstern (commented):
@N-thony it IS almost working.  Well done you!  Not just the calculator, but rename and delete also.  And this is just like memories on an ordinary calculator!
a) I typed 10 into the calculation and left the name as calc.  I saved as scalar and it disabled the the Position button correctly.  It displayed the result in the Output window, rather than a column and that seems fine.
b) I checked the sheet information and it has added Scalars and calc is there.  Perfect so far!
c) I used the rename and made it` ten`.  That works and changes the name properly. 
d) Now the first part that I thought wasn't working yet.  I couldn't make a second scalar.  Now I have realised I need to press Try first and then it works.  Is that to help the coding or do you feel it is useful for users?  I would prefer it to be optional, but can live with it.
e) I made 3 scalars and deleted them.  I notice the sheet metadata then says NA=null, so once scalars are there for a session, the column remains in the sheet metadata.  If that is trivial to change, then fine, but it does no harm, as it is overwritten by the first new scalar.

This is great.  Can you finsh off from your point of view, particularly if the dependence on Try is to be removed.  Then I'll approve and we can hand it to Patrick.

rdstern (commented):
a) First a problem when I tried the scalar as a summary.

@N-thony I just started.  I first checked that mean of yield works fine still as a variable.  When I ask for it to be a scalar it gives this error and crashes out:

image

b) I then tried the "system" saving 10 as `ten`, and 30 as `thirty.` I found them as scalars, renamed them. examined them in the metadata and then deleted them.  Brilliant.  It all works.

c) I then (of course!) logically tried without a dataframe.  It can't work, and duly gave an error.  I'm inclined to just disable them, but they could be made invisible.  Your choice.

d) I wonder what you do in the metadata now, when there are many scalars?

image

I think, for the value labels, there is a limit of 4 shown.  I would be happy with that here too.  I don't think users will often have many, but here isn't the place then to view them all!  If easy, when there are more than 4, perhaps add` ...` after the fourth.

It will be great to merge this one.

rdstern (commented):
@N-thony great, that's working, but I have now hit - I hope a small problem - using the scalars.  Look at this screen and I'll explain:
image


I now made the mean(yield) into a scalar called mean.  Then sd(yield) into sd.  Then I tried (yield - mean)/sd  and it doesn't work.

In fact nothing works when I try to combine a sclalr with a variable.  Currently I can do calculations on scalars and on variables, but I can't mix them.  So, for example `yield - ten` or `yield - mean` don't work.

I can't find the `class(ten)`, it gives an error.  The `class(mean) `is `function`.  I hope it needs a small change in the `data_book$add_scalar` function to make them work together?

rdstern:
@N-thony I thought it was working ok before?  Now I don't have ok enabled?  Am I doing something wrong?
image

rdstern:
@N-thony great.  The `Save Result `checkbox still seems stuck on being ticked?  That's whether or not a data frame is open.  (I like the fact it is disabled when `Store Scalar` is used.)

Then I still had a temporary problem using the scalars - which then went away and now seems fine.  I think it might be the first time I use something like the formula below and scalars are still ticked.  I may investigate further, but otherwise it is great now!

image

Aha gotit again in a way you can reproduce.
a) Start and open survey
b) Summary keyboard and get mean and save in scalar Mn.
c) Clear and get sd in scalar Std
d) Clear and get (yield - Mn)/Std as shown.  Then press Try with Scalar keyboard still there.  
Error as shown.  The Errort is the the yield variable is not recognised!

image

In the last run it was then ok after I unticked the Scalars.  This time it does not recognise Mn.  So similar error.

This time it was more persistent.  But I then cleared and just tried Mn and it was fine.  Then the expression above was also ok from then on.  

Another new item now.  I saved the data book (with the scalars).  Restarted R-Instat and opened it again.  Yes they are still there in the metadata.  But not in the Calculator?  And they are not in Rename or Delete either.

image

rdstern:
@N-thony excellent.  Really like it.  I assume there is just tidying to be done?

image

a) I opened a file saved with scalars previously.  I'll try from the start in a few minutes.  This gave an error message, but then opened ok.  I checked and - as you said - the scalars are now in "all the right places".  So they are in the calculator, the rename and delete.  I even renamed them as you see below.

b) I then went back to the calculator.  This had the new names, but the Sheet metadata did not, see below.
c) I tried the calculation as below and it still has the error - and doesn't recognise yield to start with.  There remains a small tweak for you to make there.
image

d) I started again and now always get an error message when importing any RDS file:  Here it is:

image

It then imports correctly, so presumably doesn't do any harm?

e) I now produced scalars and they update correctly.  So that all looks fine.  The error above is still there.  If you fiddle about, opening again, etc and changing the script slightly, then it becomes ok.  From then on it seems ok for that expression and it saves fine as another variable.

f) Then I tried another formula that mixed data and scalars and that works first time!

Almost there!

rdstern:
@N-thony I'm ready to test, but it says there are conflicts

Issue #8989, opened 2024-05-25
Improve the Data Window start-up screen


rdstern:
The data side used to be really boring and @Patowhiz produced this greatly improved screen. 

 I have been very happy with the change and simply now suggest some further items we could add.  This includes first impressions, for R-Instatand so is important.  Ideally it should include something for all our users, many of whom are R-shy, so I include @rachelkg for comments.  Some are not so shy and I report below on comments from @volloholic.  I'd like to proceed this week, so we continue to check on them with the version to be produced at the end of June.

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/2a8ec5f3-53ea-46f1-9dad-159f4610015c)

`Section 1, Start. ` This has 3 options.  I like it and don't plan any changes.  Perhaps a slightly smaller space between sections, so I can add to Sections 3 and 4.

`Section 2, Recent: ` I like this too.  I have a tiny suggestion, but otherwise suggest we leave as it is.  My small suggestion is for the RDS files.  Note there is also an R file.  When I click on that, then it goes straight into a script window.  When I click on an RDSW file it goes to the Import File Dialog.  I suggest a double-click would open it automatically?
This is only for RDS files - I am happy if a recent file like Excel, then it always opens the dialog.  So I suggest it is only for RDS files that double-click opens them straight away.  

`Section 3, Help. ` I like the existence of this section and would like to add further options:  I suggest 6 options here:
**_Introduction_** (as now, it goes to the top of the help)
**_Getting Started_** (It goes to the Getting started section in the Help)
**_Data_** (We have promoted the Data section in the help - this would go there.)
**_Prepare Menu_** (Goes to the top of this menu.  It will also indicate that each menu has its help and this is an important one towards the start of any anlysis.)
**_Climatic_**  (It is our key niche area!  We have special menus and this is one of them!)

`Section 4:  If you must!:`  (This is a proposed new section:)
**_Paste data from clipboard
Restore from Backup
Swap Data and Script Window
Add R package to R-Instat_**

The first 2 now work pretty well and are both recent additions.  Together with the 3 Start options they are the obvious ways to start with data.  
The other two options here, are an indication that R-Instat is a pretty surprising front end!  (The swap data and script window is another recent addition and I hope the adding R packages will soon be an improved facility.)

I suggest this is also now an interesting opening screen to draw our reviewer's attention to!

[...]

PR #9007, opened 2024-06-07
New help file

N-thony:
@rdstern I have added the new help file you shared here, have a look


rdstern (approved):
Looks ok.  @Patowhiz I am not sure if there is much to check, so (if ok) then please could you merge.  Not there has to be an improved version before we can release, but this version is much better for the forthcoming video that Beryl will make.

@N-thony that remainds me that it would be good, for that video, if the new front screen could be done.  Could you allocate that please - perhaps it could even be done today?  It is #8989 Is it a good @derekagorhom task?

Patowhiz (approved):

Issue #9004, opened 2024-06-05
Time series data should be defined and should also (if possible) convert


rdstern:
@N-thony this is a problem that @MeSophie and I have sort of fallen into when trying to improve the line graphs.  Some data from the datasets packagage are imported as time series (their class is ts) and we don't yet process those data well.  

I don't want to get side-tracked into adding time series methods into R-Instat yet.  So I have a limited target with this issue.  I have 2 example datasets from the standard datasets package and I'.m sure there are more.  They are LakeHuron and WWWusage.

a) When I look at the column metadata they are of class ts, so they are recognised on import automatically, but (like the polynomials) they are not indicated in the data frame.  Could we indicate them as ts, or TS, or S.  I think this will become important enough that we might want a special label, but I'll take what is easy for now.  

Would it be possible to maike this general, rather than every new case being coded separately?  So any variable that is a new type would be coded as S automatically in the grid? 

b) I am unable to convert the ts object back to numeric?  I checked [here ](https://stackoverflow.com/questions/2576876/converting-a-ts-time-series-object-to-a-vector-in-r)and it should be easy.

In R-Instat it is in multiple places.  I assume most use `data_book$convert_column_to_type`.

I tried in the `Duplicate Column`, which has an option to change the type and also in  `Prepare > Data Frame > Convert Column`.  Neither work yet, at least with options I tried?  I assume they both use the `data_book$convert_column_to_type`

I tried `Right-click > Convert to Numeric `and it converts ok but keeps the extra tsp information in the metadata.

I succeeded with the calculator, as shown below:

`calc` is just` as.numeric(value)` and gets rid of the extra metadata.
`calc1` is `unclass(value)` and makes it numeric, but leaves the tsp metadata.  Maybe we want both options in the duplicate and convert dialogs.

![image](https://github.com/IDEMSInternational/R-Instat/assets/11226469/4aec1431-0edb-4f0b-bcb6-0c0fba2e2521)

a)

Patowhiz:
@rdstern excuse my ignorance. In regards to the context of a single column header, why would we not want it labeled as numeric, is there a fundamental difference between a TS and numeric column? 

On a separate note, I suggest whenever possible, we lean towards limiting R jargons in the front end. I have no issue with class being shown as ts, a class is already an R jargon.

rdstern:
[...]

PR #9014, opened 2024-06-11
Defining the Timeseries Data 

derekagorhom:
fixing partly #9004 
This Fixes part a) and the second part of part b)
for the part b 
I took a look at the function in the `Right-click > Convert to Numeric` and realised it had ignore labels as part of the convert to function.
this option is already available in the `Prepare > Data Frame > Convert Column` but under a different name (simple convert) but not available in the duplicate dialog so i added it there.

Thanks


rdstern (approved):
@derekagorhom very nice, I am approving.
@N-thony over to you.  This is great.  The only problem, which is not here is that the right-click Convert to Numeric has "regressed".  @Patowhiz has also noticed that.  This needs to return to its former glory!  I'm not sure when that happened?
I suggest we merge this change, however and address the Convert to numeric separately.  It now always adds labels to any variable with problems and so can convert, even a character variable to numeric without complaining.  We need it to give the options it raised previously.

Issue #9025, opened 2024-06-19
Alterations to the Start Screen


rachelkg:
Hello @Fidel365, @N-thony and @rdstern 

I would like to suggest some changes to the opening screen of R-Instat. 

- Move the Paste Data From Clipboard to the end of the Start section
- Remove the Horizontal scroll bar from below the recent files, it causes an odd gap and is unnecessary. 

![2024-06-19_13-55-46](https://github.com/IDEMSInternational/R-Instat/assets/85707390/0b19a348-3753-430e-8504-4b2f63ac160f)

-  In the Help Section just have 3 items, namely: `Getting Started`, `Data` and `R-Instat Resources Site` (which links to the moodle site [https://ecampus.r-instat.org/course/view.php?id=14]
- Change the name of If you Must to `Advanced`. 
- Advanced to include 3 items 1) Swap Data and Script Window, 3) Add R Package to R-Instat, 4) Restore Data from Backup  _(RDS: see below for a minor change to the Swap Data item in this section)_

@rdstern please correct or add to this as needed.

Thanks, 

Rachel

rdstern:
@Fidel365 I support most of the changes suggested by @rachelkg and hope they will be easy to do.  The only change I suggest is to one item in Advanced.  That to have 3 items namely:
```
Paste Script from Clipboard
Add R Package to R-Instat
Restore Data or Log from Backup
```
The first item is the only one that has a different action, on your part.  I suggest, instead of swapping the script and data windows, that you just open  a script window.  (With the swap it is not obvious how you would undo that action.  Opening the script window - ready for pasting - it is clearer, with the arrow on the toolbar, how you get back.

lilyclements:
Fixed by #0-44

PR #9032, opened 2024-06-22
Alterations to the start screen

Fidel365:
Fixes issue #9025 
@rdstern and @rachelkg this is ready for review

### DIFF
--- a/instat/ucrDataView.Designer.vb
+++ b/instat/ucrDataView.Designer.vb
@@ -709,9 +709,9 @@ Partial Class ucrDataView
         Me.lblAdvanced.ImeMode = System.Windows.Forms.ImeMode.NoControl
         Me.lblAdvanced.Location = New System.Drawing.Point(5, 4)
         Me.lblAdvanced.Name = "lblAdvanced"
-        Me.lblAdvanced.Size = New System.Drawing.Size(114, 25)
+        Me.lblAdvanced.Size = New System.Drawing.Size(108, 25)
         Me.lblAdvanced.TabIndex = 0
-        Me.lblAdvanced.Text = "Advanced:"
+        Me.lblAdvanced.Text = "Advanced"
         '
         'linkStartPasteScriptData
         '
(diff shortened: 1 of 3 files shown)


rdstern (commented):
@Fidel365 good changes:
a) In the top group the vertical space to the Paste Data from Clipboard is smaller than the spaces between the others.  I don't mind making those vertical spaces smaller.
b) You have also got rid of the vertical scroll bar in the Recent section.  That needs to stay.  I assume we can have the vertical without the horizontal scroll bars.
c) Can you change the action on the Paste Script from Clipboard so it takes you here, rather than swapping the windows:

image
 
Ideally could you go further - maybe ask @N-thony, so that if you have a script in the clipboard, then it is pasted automatically into the script window?  Here is [an example](https://ggplot2.tidyverse.org/reference/geom_path.html) halfway down on this page:

image

rdstern (commented):
@Fidel365 I am pleased you are working on this, because it would be great to include the new start-up in the forthcoming version.  I'm not sure what's changed, 

image

a) But my initial view is that the spacing between the items is different in the 3 different sections.  This is what you tidied so nicely in the calculator.  Please could you do the same here?  So the vertical spacing between the advanced items is less than those between the Help items and the new Paste at the top is closer to the others.  
I think I slightly prefer the smaller spacing throughout, but it needs to look professional and hence be consistent.
b) I still can't scroll to the bottom of the More files.

rdstern:
@Fidel365 it would be good to finish this for the next release. The tasks:
a) The vertical spacing - you neatened the keyboards well in the calculator.  I assume this is simpler.
b) For the recent files, restore the horizontal line under that section (i.e. back to the previous situation) if it is difficult to get the `More` scrolling working without it.  There is no scrolling possible now.
c) Keep the Paste Script from Keyboard as now, i.e. with the full window for the script, and adding in the pasting if there is an R file in the clipboard.

Fidel365:
Work on this continues on Pr #9044

PR #9034, opened 2024-06-25
Added Scalars to the Enter Dialogue

### LINKED ISSUE #9023: Add scalars to the Prepare > Column: Numeric > Enter dialog?
@N-thony I hope this will be a quick task - and also a good test of the extent to which the scalars have been well coded.  Here is the current Calculator dialog - with scalars:
image

And here is the current Enter Dialog:

image

Could we add the scalar checkbox here as well?  Maybe the top one in the calculator goes below the keyboard in Enter.  If checked, then the Scalars are shown, just as in the calculator.  There is also the same code for the Save Scalar checkbox as in the calculator.

N-thony:
Fixes #9023
@rdstern  have a look

### DIFF
--- a/instat/dlgEnter.Designer.vb
+++ b/instat/dlgEnter.Designer.vb
@@ -78,12 +78,12 @@ Partial Class dlgEnter
         Me.ttEnter = New System.Windows.Forms.ToolTip(Me.components)
         Me.ucrSaveEnterResultInto = New instat.ucrSave()
         Me.ucrTryModelling = New instat.ucrTry()
-        Me.ucrDataFrameEnter = New instat.ucrDataFrame()
         Me.ucrReceiverForEnterCalculation = New instat.ucrReceiverExpression()
         Me.ucrBase = New instat.ucrButtons()
         Me.cmdRHelp = New instat.ucrSplitButton()
         Me.ContextMenuStripBase = New System.Windows.Forms.ContextMenuStrip(Me.components)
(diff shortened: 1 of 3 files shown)


rdstern (commented):
@N-thony that's great.  One small edit, before merging and then a query, either for now, or for another issue:  
image

a) You notice the save control is displaced in the dialog.  It needs to return to its previous position.
b) A more general scalars point.  The scalar is given in the list in the top, in "all its glory", so with many decimal places.  I don't mind that the exact number is stored, but could it be given to the defined decimals, as in Try and in the output window?
c) This is more a question.  I am happy that "scalars" could be quite wide, see below:

image

Should there be a limit, e.g. 4 of the number of scalars shown in the metadata?  Or won't it matter.  I wonder if it would be easy to show a maximum of (say) 4 scalars in the metadata, and add a `...` if there are more?

rdstern (commented):
@N-thony great.  Just 2 trivial points:
a) In the calculator @rachelkg asked whether the Save Scalar could be in the position you now have it in Enter.  Could you put it into the same place in the Calculator too.
b) The checkbox Save Scalar, in Enter, is not quite vertically aligned with the other 2 checkboxes.  Could it be moved left just a little?

rdstern (approved):
@N-thony great.  @Patowhiz over to you to check and maybe merge.

Patowhiz (approved):

PR #9049, opened 2024-07-03
Fixed problem with R-instat start up

### LINKED ISSUE #9042: Problem at start-up with the new release
@N-thony thanks for the 28 June pre-release.  It installed well.

Two small items so far to correct is there is an update within this week.  There also seems to be a problem in the critical export to google buckets for the Malawi data, so an updated release within the week may be useful if that error is fixed:  Two small things I have found are not critical.  We may find more when the workshop starts.

a) When you start it now always says it didn't shut down properly previously.  That used to work well before.  I suspect it is a by-product of the of the new Tools > Restore Backup dialog.  Ideally we return to the previous situation, where R-Instat gives that message if (and only if) the previous session crashed out.  If that is difficult now, then (at least temporarily) switch that off, so it never shows.  
b) The Version is 0.7.17.  I already have 0.7.17 and 0.7.18, so it should be 0.7.19 at least.

derekagorhom:
Fixes #9042 

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -1120,59 +1120,29 @@ Public Class frmMain
     End Sub
 
     Public Sub DeleteAutoSaveFiles()
-        Try
-            If Directory.Exists(strAutoSaveDataFolderPath) OrElse Directory.Exists(strAutoSaveLogFolderPath) OrElse Directory.Exists(strAutoSaveInternalLogFolderPath) Then
-                ' Define the retention policy (keep last N autosaves)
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@N-thony I don't see what I can check here, as it now doesn't show on my machine anyway.  So I am approving and leaving the check to you

derekagorhom:
@N-thony Thanks for this, I have pulled the changes

rdstern (approved):
I am simply approving

PR #9063, opened 2024-07-09
Fixed The Regression in Climatic>Prepare>Transform Dialog

### LINKED ISSUE #9050: Regression in Climatic > Summary > Transform dialog
This was still working fine in Version 0.7.18  (note our current numbering should be 0.7.19)
@N-thony good to check who was working on this.  I fell into this, this morning calculating the simplest case, namely rainday (called count in the dialog)

Here is the dialog now when you enter:

image

It is odd it says `cumsum`.  It should say `count`.
 If I click ok it gave me an error.  Others got a crazy result.  It should give 0 or 1 in the resulting column.

If you go at the top to Cumulative and then Back, it now says Count.  Now it works, but gives the wrong answers.  It is just the data again and not the 0 1 values.  See below:

image

The example above is with Malawi data but I get the same problem with the dodoma data from the library.

MeSophie:
Fixes #9050 
@rdstern  @N-thony I have fixed the bug between Count and Cumulative buttons on Transform Climatic dialog. Please have a look.


rdstern (approved):
@MeSophie brilliant, that now seems fine again.  Could you say a bit more about what the problem was, and (if you discover) why? @N-thony I hope this will be easy to check and merge.  It was a very unwelcome surprise in the workshop.

MeSophie:
image

@rdstern   I really don't know, the code for this part was identical to the previous version. Maybe they weren't just more precise, because I had to modify some lines of code.

PR #9065, opened 2024-07-09
Minor Improvement Made on Climatic>Define As Climatic Dialog.

### LINKED ISSUE #9059: Minor improvement in the Prepare Climatic Data dialog
When data are imported from Climsoft they look like this:

image

We then use the Climatic > Define Climatic Data and the initial dialog looks like this:

image

Notice the dialog recognises PRECIP and automatically fills the receivers.  But it does not recognise TMPMIN or TMPMAX, so we have to do this ourselves.  Could it recognise these names too please.  (I think these will always be the names from Climsoft, so good to make life as easy as possible for Climsoft users.  

@Patowhiz can you confirm?  Would it be easy to become optimistic and also give the names for the remaining elements.  That would be useful, but less urgent.

@N-thony could you suggest who might do this.  @MeSophie has already worked on this dialog?

MeSophie:
Fixes #9059
@rdstern @N-thony I added some text on Define Climatic Data dialogue code. The dialogue can now recognize TMPMAX and TMPMIN. Please have a look.


rdstern (approved):
Nice.  Thanks @MeSophie and over to @N-thony

Issue #9058, opened 2024-07-08
Further minor changes in the Front menu


rdstern:
@Fidel365 it is looking much better now.

a) Could you please add `(Online)` to the 2 items, so: `R-Instat Resource Site (Online)` and `Add R-Package to R-Instat... (Online)`

b) @rachelkg writes: "Also there are still slightly different gaps between the Section names and the items on the home screen. This bothers me, but am I just being picky? If we are making further changes I think we should standardise this to make it look more professional. "

It is now much better than before, but I am very keen we look professsional - at least here right as people start each session of R-Instat.  Could you make the gaps the same, and also mention Rachel in the pull request, so she checks it as well.

c) Then @N-thony could we check also that all these entries are in the database, so they are translated into the different languages.  If not there is an additional task.

PR #9069, opened 2024-07-11
Further improvements to the front menu

Fidel365:
Fixes issue #9058 
@rdstern this is ready for review, but I will also invite @rachelkg to check if this in now ok.


rdstern (commented):
@Fidel365 thanks for adding the Online bit.  
Now I'm not good at this spacing stuff and @N-thony .  But I would like this initial screen to look as professional as possible.
There seems to me to be a bigger space before the Advanced section, compared to the others.
And is the spacing between the 3 items in Advanced, the same as those between the Help items?  And the Paste data - is that exactly the same space below the items as the other?
@Fidel365 you  were really good at making the calculator keyboards much neater.  Can you have a similar go at the spacing here?
This is important, because it is going to be visible at the start every time anyone uses R-Instat.  And I'm keen that we give a really good initial impression.

Fidel365:
@rdstern have a look at the recent channges.

rdstern (commented):
@Fidel365 I think that looks much better now.  Many thanks.
A trivial point.  The `Restore Data from Backup...` needs the 3 dots.  And I take the opportunity to delete the `or Log`, so it is a simpler message.
@rachelkg could you also cast your critical eye over this, before it is merged - though changes could also happen later.

rdstern (approved):
@Fidel365 perfect - I like it.  @N-thony over to you to check and merge.  Maybe @rachelkg will comment before?, if not, then don't wait.  I am keen to have this new screen

PR #9072, opened 2024-07-15
Additional feature in the Find dialog

### LINKED ISSUE #9056: Additional feature in the Find dialog in the edit menu
@N-thony I hope this will be an easy feature to add.  Here is the current dialog:

image

During the workshop I looked for 2 particular values in TMPMIN

a) TMPMIN = 23.8 and this worked fine
b) TMPMIN = 2.1 and this did not work so well.  I wanted just 2.1 but there were also many days with 22.1 that contained 2.1 in the value.

Could you add a checkbox, maybe next to the Ignore Case.  It has label `Whole Value`, like the `Match Entire Cell Contents`   in Excel.  It just checkes for the value being that value, rather than that value being in the cell.  The default is unchecked.

Thanks

derekagorhom:
Fixes #9056 
I have added the Whole Value checkbox which does the Match Entire Cell Contents in the Find Dialog

@rdstern , @N-thony This is ready for review
Thanks

### DIFF
--- a/instat/dlgFindInVariableOrFilter.Designer.vb
+++ b/instat/dlgFindInVariableOrFilter.Designer.vb
@@ -28,7 +28,6 @@ Partial Class dlgFindInVariableOrFilter
         Me.grpSelect = New System.Windows.Forms.GroupBox()
         Me.rdoRow = New System.Windows.Forms.RadioButton()
         Me.rdoCell = New System.Windows.Forms.RadioButton()
-        Me.ucrPnlSelect = New instat.UcrPanel()
         Me.lblLabel = New System.Windows.Forms.Label()
         Me.rdoVariable = New System.Windows.Forms.RadioButton()
         Me.rdoInFilter = New System.Windows.Forms.RadioButton()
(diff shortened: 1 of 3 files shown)


rdstern (commented):
@derekagorhom great - it works!
a) the default should be unchecked.  I think you made the default checked.
b) while on the dialog, please could you improve the "other way" of using find, and that is via a filter.  I hope @N-thony will tell you this is an easy addition to make.

To illustrate, and show you the current situation, could you make a filter.  I assume you know how.  Anyway here are the steps.  
Right-click and choose option Filter Rows
Define New Filter
Use tmin as the variable and make the condition `< 3`
Add the condition
Back to main menu and the 3rd option, so just `Store Filter`.  (You don't apply the filter!)
Now back to your Find dialog and the start of the data.  Option at the top to use Filter.

image

It now says `Variable s`till.  Could that change to `Filter`?
Click on Find and it says it has found row 14552 - see above.
But it hasn't gone there.  It is at the top of the 14000's.  You can't scroll down without closing the dialog.  Could it go down, so 14552 is visible and could the row be marked as it is for the ordianry Find option?  

@N-thony I hope this is not too difficult?

Why are we doing this.  Well it came from the climatic analysis and the quality control showed 2.1 degrees was an odd value.  We would like to see it in context.

derekagorhom:
@rdstern for the part b of the issue you wanted me to have look, i think they already work.
See the screenshot below
image
the `variable` changes to `filter` and the find shows where the row is

rdstern:
@derekagorhom this is just to report what we discovered when looking at the Find dialog together.
a) When you define a filter first, then it works fine.  It finds the rows and they are marked, just as I would expect.
b) But when you start by looking for the cell or row , i.e. an ordinary Find first, - which now works including the all cell addition, which seems to be working fine - then it doesn't work in the same way.
So the ordinary Find is setting some things that are presumably not being cleared when moving to the Filter option.
c) In other dialogs you could press Reset, but this is disabled in this dialog.  That's possibly because the ordinary Find is just in the front end - so nothing to do with R?

derekagorhom:
@rdstern I have added a reset button to the dialog
for the part B i tried to recreate that error with different branches but i am still not able to get that bug. I will ask vitalis tomorrow to check his branch and see if we can recreate that error, if that is ok with you

rdstern (approved):
@derekagorhom looks fine and now seems to work.  @N-thony over to you

PR #9073, opened 2024-07-16
Fixed Some Bugs in Line Plot Dialog

### LINKED ISSUE #9055: Line plot dialog
I was using the line plot dialog and after producing the graph I needed, on going back to the dialog I found some of the features I had used before had gone back to their default. So the dialog is not the same as I left it.
The data used here is ggplot2's economics.

The first feature is the Add Line check box, on producing the graph I ticked the check box but on returning the check box is unticked. I would love to find it checked for the next graph I want to produce. Below is an image showing the box checked, when I close the dialog and open it is no longer checked.

instat_hBO5J9FMal

The second feature is the color option in the Line options subdialog. I changed the color to red, and when I returned to produce another graph that still needed the red color, it changed back to the default, blue. I need it to stay red so that I don't have to keep going back to change it. Below is the image of the subdialog.

instat_AwqmgyfJfs

MeSophie:
Fixes #9055
@N-thony  @rdstern  @berylwaswa I fixed some bugs in line plot Dialog. Please have a look.


berylwaswa (approved):
Thank you @MeSophie this is working fine as expected. @N-thony you can merge it.

rdstern:
@MeSophie I'm still confused by the options here.
a) With the economics data I first just used the Line option with unemploy by date.  Then go into the Line options and change the colour to red.  When I return to the main dialog the date has been cleared from the receiver?  I wonder why?
b) Then I went to the smooth options as shown below:

image

I added the line - as @berylwaswa stated, but the line isn't red - though the red remains set.  

Thanks

Roger

MeSophie:
@roger i fixed the problem with the  x receiver. If you want to change the default color or any other parameter of the line function outside the `line button`, you need to click on the line options again and make the change. This is because the line function used for the` line button` is different from the one used in the other buttons, to ensure independence between the different buttons.

rdstern (approved):
Looks ok now.  @N-thony to you

Issue #9079, opened 2024-07-23
Another very small change in the front screen


rdstern:
@Fidel365 I have just noticed the inconsistency here!

![image](https://github.com/user-attachments/assets/788a738e-4de4-4190-8528-dc42ef950863)

Could you please swap the top 2 items in the introductory screen, so they match the order of the 3 options in the File menu.  Sorry - I should have noticed before!

Thanks

rachelkg:
Hello @Fidel365

On my review of the front screen I notice that the 'Paste data from clipboard...' is not properly capitalised.  Please could you make it 'Paste Data from Clipboard...' so that it matches the other items.
![2024-07-25_15-07-45](https://github.com/user-attachments/assets/93e85750-8199-4edc-a10a-60c19f93b2a9)

Thanks, 

Rachel

rdstern:
@Fidel365 @N-thony I am happy!  And closing.

PR #9080, opened 2024-07-24
Small change in the front menu

Fidel365:
Fixes issue #9079 
@rdstern should I add the shortcuts also i.e Ctrl +N and Ctrl + O to the corresponding 2 items as it appears on the file menu? Otherwise this is ready for review.

### DIFF
--- a/instat/ucrDataView.Designer.vb
+++ b/instat/ucrDataView.Designer.vb
@@ -842,7 +842,7 @@ Partial Class ucrDataView
         Me.linkStartNewDataFrame.AutoSize = True
         Me.linkStartNewDataFrame.ImeMode = System.Windows.Forms.ImeMode.NoControl
         Me.linkStartNewDataFrame.LinkBehavior = System.Windows.Forms.LinkBehavior.NeverUnderline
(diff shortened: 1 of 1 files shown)


rdstern (dismissed):
@Fidel365 thanks for making the changes so soon.  Also for your ctrl N and O suggestions.  You know I hadn't noticed them on the file menu.  They work nicely - as they do for Excel and Word!
So I was very happy with the fact you are thinking in this way and that's convinced me to forward you a message related to a possible different bit of work!  We don't need to draw extra attention to them, so I'm just hoping this can be merged now.  @N-thony could you check and merge please.

rdstern:
@Fidel365 just noticed another very minor inconsistency.  Sou could you add `...` after the Help > Getting Stered and Data options.  If you don't have the front screen available, then you can use the Help menu, and those entries have the `...` there.

Thanks

Fidel365:
@rdstern thank you for the recommendation, I am following up on that. Otherwise this is ready again.

rdstern (approved):
Looks fine.  Thanks.
@N-thony over to you

PR #8954, opened 2024-04-29
Bug fix on pasting multiple columns

N-thony:
Fixes #8815 
@rdstern @rachelkg @Patowhiz this fixes the pasting of multiple columns. I used the scripts suggested by Patrick in the issue. Have a look.

Issue #8815, opened 2024-02-14
Edit Menu - Paste New


rachelkg:
Hello @rdstern and @N-thony, 

I the edit menu there is an item called Paste New. This takes the data from the clipboard and makes it into its own data frame. Should this menu item be called Paste New Data Frame? I think it would be clearer, Paste new feels a bit ambiguous. 

![2024-02-14_11-10-28](https://github.com/IDEMSInternational/R-Instat/assets/85707390/7232d679-0103-4cef-a3f7-7f25fe381329)

rdstern:
@Fidel365 Here are Rachel's questions, to answer, so she can write the help for this dialog.  I hope you can help her.  You may have to check with Patrick if there are bits you can't answer.  There may not be any code changes needed, in which case you can continue the discussion here.  Then @rachelkg can close this issue when she has the information needed:

She writes:
"In the paste to new dataframe help you immediately give an example that doesnt work. I suggest we start by outlining when this does work. i see that it works well with data from other data frames and from excel. Would it work well coming in from other stats programs? 

I assume the disadvantage is that it doesn't bring any of the column metadata so we recommend importing? 

How about the columns feature? What are the parameters for this? I have tried some but keep getting errors. I guess because I am trying to add my data to data frames that do not match it in some way? Maybe good to have a quick chat about it tomorrow if you have time. "

My answer so far.  When you open this dialog it simply copies whatever text is in the clipboard.  So, in the current help we could assume the Edit > Copy New Data Frame dialog is open and the Help has been clicked.  Now they could go down to the data shown (the 4 rows).  They could mark the 4 rows, then either right-click and choose Copy, or press <ctrl>C, which is the same thing.
Now close the help and lo and behold, nothing has happened!  But click on the Reset button and the text copies into the dialog.  Then untick the option to use the first row as the header.  Then you are ready to go!

It can be very quick and convenient to copy data into R-Instat in this way.  But it is only for small datasets, because it is limited to the visible window onto the data.  Usually it is just for a few rows and columns.  So, from Excel, etc, you usually us the File > Import from mFime dialog.

You can also add these data to an existing data frame, that's the Columns option, but here there are some restrictions.  Perhaps, in the Help that might be an Anchor point? Usually what you copy will be the same length as the data frame.  So you could add another opy of the same data.

[...]

PR #9091, opened 2024-07-31
Bug fixes in adding columns

### LINKED ISSUE #9086: Addition of columns
**Describe the bug**
The add columns R function of the databook has gone through changes several changes recently.  This has caused some regressions. 
For instance;
1. If I use the split dialog, to split contents of a column into individual columns, only 1 column gets produced.
2. The new Climsoft import dialog implemented in PR #9070 also throws an error when importing metadata due to internal conversion of numeric columns. This error was reported by @rdstern in this [comment](https://github.com/IDEMSInternational/R-Instat/pull/9070#pullrequestreview-2198758499)

**To Reproduce**
1. Use the split text dialog to split contents of a column into fixed number of components 
2. Use the import climsoft dialog in PR #9070 to import the metadata.

**Expected behaviour**
@N-thony you recently worked on the add columns R function,  kindly investigate what may have caused the regressions outlined above. 

@rdstern we may have to wait for this fix before we release a new version. We may have to also test other areas that add or change columns, to make sure they still work as expected. 

**Screenshots**
3 - Split WND Column

N-thony:
Fixes #9086
@rdstern @Patowhiz I don't remember the reason that code has been changed but this seems now to be fixed. It will be good to test this PR seriously and see if there is no regression. I did some test and all seems to work fine. Over to you.

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -712,7 +712,7 @@ DataSheet$set("public", "add_columns_to_data", function(col_name = "", col_data,
       if(require_correct_length) stop("Length of new column must be divisible by the length of the data frame")
       else curr_col <- rep(curr_col, length.out = self$get_data_frame_length())
     }
-    if(use_col_name_as_prefix) curr_col_name = self$get_next_default_column_name(col_name[i])
+    if(use_col_name_as_prefix) curr_col_name = self$get_next_default_column_name(col_name)
     else curr_col_name = col_name[[i]]
     curr_col_name <- make.names(iconv(curr_col_name, to = "ASCII//TRANSLIT", sub = "."))
     new_col_names <- c(new_col_names, curr_col_name)
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -712,8 +712,9 @@ DataSheet$set("public", "add_columns_to_data", function(col_name = "", col_data,
       if(require_correct_length) stop("Length of new column must be divisible by the length of the data frame")
       else curr_col <- rep(curr_col, length.out = self$get_data_frame_length())
     }
+    
     if(use_col_name_as_prefix) curr_col_name = self$get_next_default_column_name(col_name)
-    else curr_col_name = col_name[[i]]
+    else curr_col_name = col_name
     curr_col_name <- make.names(iconv(curr_col_name, to = "ASCII//TRANSLIT", sub = "."))
     new_col_names <- c(new_col_names, curr_col_name)
     if(curr_col_name %in% self$get_column_names()) {
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -693,9 +693,9 @@ DataSheet$set("public", "add_columns_to_data", function(col_name = "", col_data,
     }
   }
   
-  if(length(col_name) != num_cols) {
+  if(length(col_name) != num_cols && (num_cols == 1 || length(col_name) == 1)) {
     use_col_name_as_prefix = TRUE
-  }
+  } else use_col_name_as_prefix = FALSE
   
   replaced <- FALSE
   previous_length = self$get_column_count()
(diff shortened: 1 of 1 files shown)


rdstern:
Thanks, @N-thony I'll check soon, but am assuming @Patowhiz will check first, as it resolves a problem he found.  So easy for him to check that it now, at least, solved the original problem he found.

Patowhiz (changes requested):
@N-thony 

Here are a few comments and suggestions:

Regarding item 1 of the issue:

- It is solved, but I suspect your current solution may cause regressions elsewhere. I think the issue relates to:
  1. Line **685-694** of the `data_object_R6.R` file. I'm not sure when line 685 was added/modified and why the check (`length(col_name) != 1)` is relevant. Ideally, the check in line 685 should stop any further execution of the function.
  2. The script produced by the dialog:
  ```
  X <- data_book$get_columns_from_data(data_name="flags_metadata", col_names="X", use_current_filter=FALSE)
  wind_cols <- stringr::str_split_fixed(string=X, pattern=stringr::coll(pattern=","), n=5)
  data_book$add_columns_to_data(data_name="flags_metadata", col_name="wind_cols", col_data=wind_cols, before=FALSE, adjacent_column="X")
  rm(list=c("wind_cols", "X")) 
  ```
  Notice that `wind_cols` will contain multiple columns, yet the dialog sets the `col_name` parameter. I think the correct solution would be to make the dialog set the `use_col_name_as_prefix` parameter to true and use the passed `col_name` as the prefix of all the newly generated column names. If `use_col_name_as_prefix` is set to true, you can make changes at line 685 that will make the length of `col_name` equal to the value of `num_cols`, which means you won't need to make inconsistent changes as currently done in line 715.

Regarding item 2 of the issue:

- I checked with the new Climsoft import dialog in PR #9070 but still got the error. Note that I explained where I think the bug is in [this comment](https://github.com/IDEMSInternational/R-Instat/pull/9070#issuecomment-2244135966). In the PR, I have fixed it by not using the data book R functions for factor conversion, so if you want to test my original issue, use my branch named `climsoft_updated_dialog_metadata_bug`.

Patowhiz:
@N-thony 

Note that closing this PR means the first item is not yet fixed.

Regarding the second item in the issue, it is not specific to the Climsoft import but applies generally to any similar operations. I hope you intend to fix that in another PR as well?

I'm  accepting to close this PR and will reopen the issue issue it attempts to solve. 

On a separate note, I made a few comments above. They are quite critical and generally apply to other dialogs with similar operations. It's essential that you consider the items I have outlined when attempting to fix the issue.

Thanks.

rdstern:
@N-thony and @Patowhiz can anything be merged yet?

Patowhiz:
@rdstern Considering my review comments above, which explain in detail where the problem might be and also point out the potential regression that may arise due to the solution provided here, from my side, this is not ready to merge.

I'm happy for @lilyclements to take over, review, and merge this one. The R function has gone through several modifications, and I may be missing something that @N-thony is already familiar with, which might render my review irrelevant.

lloyddewit:
> @lloyddewit can you review this when possible?

@N-thony Thanks, but I don't think I have anything to add. I suggest that one of the normal approvers does the review/approval.

lilyclements:
Looking at the code history, this [[i]] was added in by "Bug fix by pasting multiple columns", which relates to PR #8954 on issue #8815

If we want to merge this, I suggest we reopen #8954 with an explanation as to why it's reopen and a link to this PR.

rdstern (approved):
@N-thony this seems fine now on the few checks I made.  I checked an example on splu=it and also ranks on the row statistics.  Both produce multiple variables.

I am also really happy that:
a) There seem fewer regression bugs
b) It is other than me finding them.  This was an important @Patowhiz discovery and @rachelkg has also been making suggestions recently.  It will be good if (at least the) rest of the development team become more proficient users/testers too.

rdstern:
@N-thony I have approved.  @Patowhiz back to you.  I agree it would be great to merge this one in the forthcoming release

Patowhiz:
@rdstern my comments and questions above are yet to be responded to by @N-thony . I suspect we may regress other parts if I merge this without knowing the answers to my questions above. In my above comments, I've also given an alternative solution that involves solving the problem at the dialog level (this has far less risk). 

I'm happy for you and @lilyclements or someone else to review and merge this.

rdstern:
@N-thony are you reasonably confident that your solution is ok and will not cause the regressions @Patowhiz is concerned about?  If so, then I suggest we ask @ChrisMarsh82 to check and merge?  It worked nicely for the examples I looked at - which is seems were a problem before.  So I am keen to progress.

rdstern:
@N-thony I am unhappy delaying the merging of this bug-fix.  I have tested the Prepare > Text > Split with data from tidyr (world populations) and it works fine with your fix.  
I have now looked at the same example in the latest version and it doesn't work.  That's what started this off with Patrick's issue on July 30.  So, unless we merge, we are waiting 2 months for the fix of an obvious problem.

I also checked the only other examples I could think of, where multiple variables are produced and added to an existing data frame.  They are Prepare > Numeric > Random samples.  That doesn't work on the merged version and seems fine with your fix.

And the Prepare > Numeric > Row Statistics (with ranks) is the only other example I can think of for now.  That also works fine with your fix.

So I find it curious that the 2 best programmers won't take responsibility in this case.  I would like this to be merged in the forthcoming version, because it is the only known (obvious) bugs.  Assuming @Patowhiz is still not prepared to merge, then please @ChrisMarsh82 could you do so.  Whether you check the code or just merge I don't care any more - it is time this fix was included.

Patowhiz (approved):
Merging this to facilitate further testing and new release that needs the bug fixed.

Issue #8847, opened 2024-02-28
In the Script window, add an option to reformat R code to tidy spaces, indents and line breaks


lloyddewit:
@rdstern , @N-thony
RStudio allows reformatting of R code. We could provide something similar in the R-Instat Script window.
I considered adding this feature to the `RInsightF461` library. However R code formatters already exist and are available as R packages. [styler](https://styler.r-lib.org/index.html) seems like a good candidate.
Therefore, we could pass `clsScriptActive.Text` to an R formatting package and then redisplay the returned formatted code. We would also need to ensure that the cursor (aka 'caret') is set to the equivalent location in the formatted code.
Could this issue be implemented by the development team?

lloyddewit:
> @lloyddewit sounds good, should we have a right click menu too or only a button alongside with others existing buttons for formatting option?

Personally, I think we should copy RStudio as much as is practical (to make it easier for users to transition between RStudio and R-Instat). See screenshot below. But I will leave the final decision to you and @rdstern.

![image](https://github.com/IDEMSInternational/R-Instat/assets/57253949/0850724e-a107-4ad4-9e18-da0ce9cfca23)

rdstern:
@lloyddewit and @N-thony I am liking this discussion.  Agreed to be as similar as possible to RStudio, both in the terms we use, and in the short-cuts we implement.  But I am not sure we need a new main menu with scripts in mind.  The Edit menu is currently very short.  Why not a horizontal line at the bottom of the current menu.  Below the line is for scripts?  

@N-thony can we discuss how to implement some of these features?  Could this be a job (a few tasks here) for you?  You could allocate and support.  But I am also keen to discuss the team writing articles, at least for local conference presentations, and you should certainly be doing at least one.  They would be general (R-Instat), but then focussing on an aspect that is partly your work.  This could link to your past contributions here, and then these enhancements?

lloyddewit:
@derekagorhom If you add an option to the 'Edit' menu to reformat the code, then I assume you will want the cursor to be placed in the same statement after the reformatting is done. The easiest way to do this may be to use the `RInsightF461` library. I may need to add some features to this library to support your work. When you reach this point in your work, then let me know and we can discuss.
Thanks

derekagorhom:
@N-thony I think this issue is completed

PR #9107, opened 2024-08-19
Added the reformat option into the script window via right click

derekagorhom:
fixes partly #8847
To fully implement the Reformat code in the Edit menu, i need to test how well it work on the script window itself so i implemented
a right click reformat option and this is the result
 I think i was able to do the white spaces and indents (the screen recording did not show me right clicking and selecting the reformat option so i used a screenshot to show)

https://github.com/user-attachments/assets/903db2d0-4923-4531-8d4c-db0ba4ac106a

image

@lloyddewit or @N-thony can this be reviewed to show whether that is what you guys wanted the reformat to work like

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -237,8 +237,6 @@ Public Class ucrScript
         FormatCode = Join(formattedLines, vbCrLf)
     End Function
 
-
-
     Private Function GetTextBefore(ByVal text As String, ByVal position As Long) As String
         ' Returns the text before the specified position
         If position > 0 Then
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -144,6 +144,120 @@ Public Class ucrScript
         End If
     End Sub
 
+    Public Sub ReformatText()
+        Dim fullText As String
+        Dim selectedText As String
+        Dim formattedText As String
+        Dim startPos As Long
+        Dim endPos As Long
+
+
+        If clsScriptActive.SelectedText.Length > 0 Then
+            selectedText = clsScriptActive.SelectedText
+
+            formattedText = FormatCode(selectedText)
+
+            startPos = clsScriptActive.SelectionStart
+            endPos = startPos + Len(selectedText)
+
+            fullText = clsScriptActive.Text
+
+            ' Replace the selected text with the formatted text
+            clsScriptActive.Text = GetTextBefore(fullText, startPos) & formattedText & GetTextAfter(fullText, endPos)
+
+            ' Optionally, you can call the Copy function if needed
+            clsScriptActive.Copy()
+
+            EnableDisableButtons()
+        End If
+    End Sub
+
+    Private Function FormatCode(ByVal code As String) As String
+        ' This function will format the input code string
+        Dim lines() As String
+        Dim formattedLines() As String
+        Dim i As Integer
+        Dim indentLevel As Integer
+        Dim formattedLine As String
+        Dim inAssignment As Boolean
+        Dim assignmentPos As Long
+
+        ' Split the code into lines
+        lines = Split(code, vbCrLf)
+        ReDim formattedLines(0 To UBound(lines))
+
+        ' Initialize the indentation level
+        indentLevel = 0
+        inAssignment = False
+
+        ' Loop through each line and apply formatting rules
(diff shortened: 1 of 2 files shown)


instat/ucrScript.vb
```diff
@@ -144,6 +144,112 @@ Public Class ucrScript
...
+        ' This function will format the input code string
+        Dim lines() As String
+        Dim formattedLines() As String
+        Dim i As Integer
+        Dim indentLevel As Integer
+        Dim formattedLine As String
+        Dim inAssignment As Boolean
+        Dim assignmentPos As Long
+
+        lines = Split(code, vbCrLf)
```
lloyddewit, line 185: We are splitting on `CrLf`. I think that script read in from files may have line breaks that are just `Cr` or `Lf`.

instat/ucrScript.vb
```diff
@@ -144,6 +144,112 @@ Public Class ucrScript
         End If
     End Sub
 
+    Public Sub ReformatText()
```
lloyddewit, line 147: Does this need to be public?

lloyddewit:
@derekagorhom Thank you, this looks like a great feature.
I think the official review should be done by one of the normal senior developers. However, I can add some extra comments:
- Currently, only the selected text is reformatted. Is this the expected usage or will the user normally prefer to reformat the whole script? In many editors the script is reformatted automatically, either on save or when the user hits return or ends a statement. Does @rdstern have a view?
- Currently, the cursor is not repositioned after the reformatting, so it will effectively jump to a different place in the script (because of the added/deleted spaces). This could confuse the user. It may be tricky to reposition the cursor exactly where it was but it may be enough to move the cursor to the end of the current statement. I added some technical notes below.
- Currently, the reformatting is limited. As far as I can see, it only reformats some assignments and commas. It does not handle other types of reserved words or add/remove line breaks. This may frustrate the user. Adding the level of formatting provided by other editors will be a lot of work. It may be better to reuse an existing formatter. Issue #8847 states 'R code formatters already exist and are available as R packages. [styler](https://styler.r-lib.org/index.html) seems like a good candidate.'. What's your view regarding R formatters available in the R environment?

Hope this helps

**Technical notes**

You can use the `RInsightF461` library to reposition the cursor after the reformatting. This library provides a dictionary that contains the start position of each statement in the script. You may need to upgrade to the latest version. 
One approach could be:
- Before reformatting, find the current statement (e.g. the ninth statement in the dictionary)
- After reformatting, regenerate the RScript object from the reformatted script. Then find the end position of the ninth statement and reposition the cursor there.

https://github.com/lloyddewit/RInsightF461/blob/master/RInsightF461/RScript.cs

image

lilyclements:
> Currently, only the selected text is reformatted. Is this the expected usage or will the user normally prefer to reformat the whole script? In many editors the script is reformatted automatically, either on save or when the user hits return or ends a statement. Does @rdstern have a view?

@lloyddewit I would generally want to reformat the whole script, but there might be times when I don't. For example, if I do a long "ifelse", and want bits on separate lines, it can make it really wide and I can't read it all. This is particularly true in R-Instat where we have to put the package names too which take up extra space. Here's an example of a function I've written in R in a package where I'd like it on separate lines for readability, but if I were to "reformat all" here, I would have a lot of blank space, however, there are other parts in this where I would want to reformat a section of the code:

https://github.com/user-attachments/assets/e2f190f4-6a1d-4164-a330-416611f4a560

lloyddewit:
@lilyclements Thanks for the feedback and example. The code in the video could be a good test case.
@rdstern @derekagorhom What's your view on formatting all the text, or just the highlighted part?

rdstern:
@derekagorhom I am coming late to this discussion.  With @lilyclements did you resolve the point about using an R package, which is our usual method in R-Instat?  I suggest it will also be a safer method than writing our own code?  If so, then either `styler` or `formatR` seem to be the obvious choices? 

Then I am quite happy to (usually?) format the selected part of the code.  If nothing is selected, does it reformat the current line?  If you want to select all, then you do Select All, or have a different right-click option?

derekagorhom:
> pass the script text through one of these packages and re-display

Hello @N-thony can we have a meeting on this

derekagorhom:
This PR is being replaced by #9153

PR #9116, opened 2024-08-29
Minor Change to Factor level for One Variable Summarise

### LINKED ISSUE #9114: Minor edit on one-variable summarise
The default `Describe > One Variable Summarise `now has the maximum factor levels as 12.  That's fine.  What is not fine is that the control cannot be set to more than 12.  Please take this limit away.  There doesn't need to be a limit.

image

@N-thony this is simple, and so could be allocated to anyone.

derekagorhom:
Fixes #9114 
i have changed the `masum` maximum limit from 12 to the `integer.maxvalue` which is 2147483647
when you open the dialog it will display 12 as the initially but it can be increased now
@jkmusyoka can you review this
Thanks

### DIFF
--- a/instat/dlgOneVariableSummarise.Designer.vb
+++ b/instat/dlgOneVariableSummarise.Designer.vb
@@ -245,11 +245,11 @@ Partial Class dlgOneVariableSummarise
         Me.ucrNudMaxSum.AutoSize = True
         Me.ucrNudMaxSum.DecimalPlaces = New Decimal(New Integer() {0, 0, 0, 0})
         Me.ucrNudMaxSum.Increment = New Decimal(New Integer() {1, 0, 0, 0})
-        Me.ucrNudMaxSum.Location = New System.Drawing.Point(246, 249)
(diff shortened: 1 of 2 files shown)


rdstern:
@derekagorhom I am having second thoughts about this very high limit!  The max of 12 is certainly too low.  Let's make the new limit 100.

In the documentation we can then run from a script and change that value if more is needed.

Thanks
Roger

derekagorhom:
@rdstern I have changed the upper limit to 100 now...also are you fine with the maxsum showing 12 when the dialog opens
Thanks

derekagorhom:
@jkmusyoka please can you review this PR.
Thanks

lilyclements (approved):
@derekagorhom looks good! Happy to approve

@N-thony over to you

rdstern:
@N-thony just to confirm another pull request that should be easy to merge.
Although very simple, this is another small inconvenience solved.  I am optimistic we can soon have a "proper" release that is also on the web site and is the final one for 32 bit machines.

Issue #9005, opened 2024-06-05
REturning presentation of roman to the calculator


rdstern:
@Fidel365 I think - except for presenting Roman numerals in the reogrid (that may need discussion with @N-thony) that the improvements to the calculator (again except for the great addition to scalars by Antoine) are now done.

There are 9 open issues that relate to the calculator, namely
#8940 
#8931 
#8925 
#8918 
#8905 
#8904 
#8902 
#8884 
#8820 

I think they may all be done and hence could be closed.  Could you please check, report and close - or get @N-thony to close if you are not sure.

Thanks

Roger

Fidel365:
@rdstern I have closed all execept issue #8931 because of the issue with Roman numbers which is still in progress. So this can be closed when you are happy.

rdstern:
@Fidel365 that's great.  Please also remember the item of adding the functions file (for the functions keyboard) into the R-Instat library.

PR #9119, opened 2024-08-29
Fixed the issue of roman numbers not displaying in the grid

Fidel365:
Fixes issue #9005 
@rdstern not very long ago, I realised that the issue we are experiencing likely arises from the fact that the as.roman() function in R converts numbers to Roman numerals as a character vector. When you add this character vector back to your data frame using data_book$add_columns_to_data(), the column might be interpreted as numeric rather than character, resulting in "normal" numbers instead of Roman numerals.
Problem:
Character Conversion: as.roman(x3a) returns Roman numerals as characters.
Data Frame Column Type: When you add the Roman numeral characters to your data frame, they might be converted to numeric types if the data frame expects numeric values, leading to the display of normal numbers.

So, I went ahead and made the changes in the calculator so that instead of *utils::as.roman()), we have *as.character(utils::as.roman())*
Well, this is working well now for both the reogrid and the output window, have a look!

### DIFF
--- a/instat/ucrCalculator.vb
+++ b/instat/ucrCalculator.vb
@@ -4691,9 +4691,9 @@ Public Class ucrCalculator
 
     Private Sub cmdRoman_Click(sender As Object, e As EventArgs) Handles cmdRoman.Click
         If chkShowParameters.Checked Then
-            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("utils::as.roman(x= )", 2)
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("as.character(utils::as.roman(x= ))", 3)
         Else
-            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("utils::as.roman( )", 2)
+            ucrReceiverForCalculation.AddToReceiverAtCursorPosition("as.character(utils::as.roman( ))", 3)
         End If
     End Sub
 


rdstern:
@Fidel365 that's great.  I am temporarily not able to test properly as I don't have my laptop.  
But I tried your solution in the existing calculator.  It works, but has the problem that I can no longer do arithmetic on the resulting variables.  
With the previous code (so without the as.character) it displays nicely as roman numerals in the R viewer (right-click on the tab at the bottom and choose View Dataframe.  So I had a column with 1 to 31 as roman in calc and it works fine.  Then I can also do calc + 2, or calc * 3, etc and they work nicely.  That's nice to show you can do arithmetic on Roman numerals.  And if calc and calc1 are both Roman I can do calc + calc1 etc.

I can't do any of that if I make them into character variables - which is a shame.  I assume it should be possible to display the original code as Roman in the grid, if it works so nicely in the R viewer?

rdstern:
@Fidel365 is this still something you continue with, or should it be handed over to @N-thony or maybe @ChrisMarsh82 ?

rdstern:
@Fidel365 you have been very quiet on this.  An additional point is that the roman variable is of type roman.  Could this please also be noted, either as R for Roman if we are not using that already, or S, for a special structure.

rdstern:
@Fidel365 and @N-thony can this now be closed?

Issue #9109, opened 2024-08-22
Serious issue on where R-Instat looks for R-packages?


rdstern:
@N-thony and @ChrisMarsh82 I think we have uncovered the serious problem that is plaguing our export for Google buckets.

I write because if I am right, then I want all the praise, but @jkmusyoka and @lilyclements have been doing all the work!

The problem is that the export dialog fails on most machines and seemingly in different ways.  It works perfectly on James and Lily's machines!  It worked on my machine in Malawi, and now it did not.

Lily has (magically) diagnosed that it could be purrr that is the problem and I find that I have purrr 1.01, while it was updated to purrr 1.02 on 10 August.  James has updated purrr and also rlang on my machine and now it works.

Now our main discovery.  On Micah's machine she is using purrr 0.34 and it gives an error.  James has deleted R-Instat purrr and checked and she is still using purrr 0.34!!!!

So we deduce that when R is installed already, then R-Instat appears to be looking first in the R packages and only afterwards does it look in the R-Instat ones?

That is at least an explanation of the different errors we get, that seem to be machine dependent.

Could you please check urgently - maybe even include Danny if you feel it is necessary?  

In making the changes could you also please remember we are hoping in the future that R packages in R-Instat could be addedm witht needing to be an adminstrator.  Maybe that could be added at the same time?

James may add!

rdstern:
@ChrisMarsh82 I assume this is your sort of problem and wonder about your silence?

rdstern:
Delighted to hear that @N-thony has been contacting @ChrisMarsh82 and a solution may be in sight.  I assume earlier, that the packages were separate, and understand from Danny that this is a line or so that needs to be included in the installation procedure.  So I hope we can revert to the previous situation.

I saw more evidence of the new problem and think I understand why it happened.  I note two instances:
a) On Micah's machine in Zambia, she was the only one who could not do (I think) an inventory graph.  The error message said she was using ggplot2 version 3.46 and needed at least 3.5.  This satisfies my "theory" that the issue may have been around for quite a long time.  It is not evident for anyone involved in development, because they use the R packages than - so they are kept up-to-date.  It was not evident in West Africa, because I assume there was no R installed separately.
[...]

PR #9120, opened 2024-09-03
Fixing the issue on where R-Instat looks for packages

N-thony:
@rdstern @ChrisMarsh82 this PR is attempting to fix the issue #9109 on where R-Instat looks for package. I have added
`if (length(.libPaths()) == 2) .libPaths(.libPaths()[2])` in the `installPackages.R` to specify the library after installation of R-Instat to looks for it packages by default. This can be tested only after creating a test release and I'm planning to do that once this PR is merged.

### DIFF
--- a/instat/static/InstatObject/R/InstallPackages.R
+++ b/instat/static/InstatObject/R/InstallPackages.R
@@ -156,6 +156,9 @@ packs <- c("abind", "agricolae", "agridat",
 
 install.packages(packs, dependencies = FALSE, repos='https://cloud.r-project.org', type="win.binary")
 
+# Only use internal library
+if (length(.libPaths()) == 2) .libPaths(.libPaths()[2])
+
 #install development packages not on CRAN
 devtools::install_github("ianmoran11/mmtable2")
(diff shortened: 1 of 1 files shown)


rdstern:
@N-thony and @jkmusyoka that's an interesting addition.  I'm a bit concerned that we are looking for a general solution and might there not be a situation (maybe in the future) where the length of libPaths is more than 2?  

Currently my understanding is that the path we want, when installing Version 0.7.20 (say) is just:
"C:/Program Files/R-Instat/0.7.20/static/R/library"?
Is that somewhere else in the installation?  I think you are saying it is?  In that case perhaps you are assuming this will be the second path if there is more than 1?
Anyway, let's try!
I am more comfortable now, that we can, if necessary, check and alter these paths once we start R-Instat!

jkmusyoka (approved):

Patowhiz:
I'm not sure about this. @ChrisMarsh82 would be the right person to approve.

rdstern (approved):

Patowhiz:
@N-thony I'm glad that this has been broadly approved, sounds good to merge.

PR #9125, opened 2024-09-04
Small Change in the Start and End of Rain Dialogs.

### LINKED ISSUE #9117: Small (or large) change in the start and end of the rains dialogs
Here is the current start of the rains dialog:

image

a) @lilyclements has solved the problem of the Occurrence control, at the bottom right.  It would be wonderful if that can now be implemented?  (It is on both the start and the end dialogs.)
b) Meantime a trivial edit is to make the default to have the Date checkbox checked.  So the default is to produce 2 variables.  (When the Occurrence is done, then the default should be to have that checked as well.  But start with the Date checkbox.

The reason is that the new Export to Google buckets exports the dates when they exist.  So let's make it easy to produce them.

MeSophie:
Fixes #9117
@rdstern , @N-thony I Added drop parameter on Start and End of rain and set the default of Date box to be checked. Please have a look.


lilyclements:
@MeSophie, try instead with "200mm rain over 3 days". This drop is to distinguish if an NA in the start of rains is due to no data, or if there's no start date under that definition. 
Asking for 200mm over 3 days should give years which do not fit this definition at all, and so you should then see a difference if you do drop = TRUE vs drop = FALSE

(Note that in R, it automatically adds new columns to the bottom of the data frame, so you may want to rearrange by year if you do drop = TRUE then drop = FALSE)

MeSophie:
image
@rdstern Please do you mean the occurrence box to be checked by default? It is mean that the three boxes are checked by default?

rdstern:
@MeSophie that's right.  All 3 are now checked

MeSophie:
@rdstern I tested the dialogue with Lily's help and the changes work well.

rdstern:
@MeSophie itis certainly different now!
a) The date and status were not yet ticked by default.
b) I tried dodoma (data filr from Jan to Dec), 20mm in 3 days, just in November.
Here were the results:

image

I hope you can see the 2 problems here:

a) You are only showing the successful years - those are correctly TRUE.  Please show all the years.  The dat and date should be missing and the logical is then FALSE.
b) The last 2 years have a missing value.  Then the logical should be missing.

MeSophie:
@rdstern the boxes are checked by default on my side.
image

Concerning the missing year and the TRUE when NA's, 
image 
I thing @lilyclements  is the best to have a look on that. Thebug is may be coming from the R code because we have the same issue with vs 0.7.20.

image

lilyclements:
I am not getting an issue with (a)?

to b)  - 

We run a filter to: `rain >= 0.85 & roll_sum_rain > 20) or is.na(x=rain) or is.na(x=roll_sum_rain)`
Then we define occurrence to be when there are rows of data after applying that filter. 

In, say, 1947, we get `NA` for a start date, and get `FALSE` for the occurrence because our filter gives no rows of data.
This is because we had rainfall, but never had a rolling sum over 20mm.

In 2012, we get NA for a start date, and get `TRUE` for the occurrence.
The rainfall is **always NA**, so when we run our filter, while there's no rolling sum over 20mm, we do have rain and the rolling sum as NA. This means there are rows of data, so the occurrence is not NA.

Perhaps we want to define occurrence to be when there are rows of data after applying that filter, **unless if is.na(start_rain) or is.na(start_rain_date), then set occurrence to be NA**

Is that what we want @rdstern?

If so, 

```
start_of_rains_status <- instat_calculation$new(type="summary", function_exp="n() > 0", result_name="start_rain_status", save=2)
```

becomes

```
start_of_rains_status <- instat_calculation$new(type="summary", function_exp="ifelse(is.na(start_rain), NA, n() > 0)", result_name="start_rain_status", save=2)
```

Except, we can't call `start_rain` here, since it hasn't been defined at that stage, so it may have to be we repeat the `ifelse` statement which creates `start_rain` - 

```
start_of_rains_status <- instat_calculation$new(type="summary", function_exp="ifelse(is.na(ifelse(test=is.na(x=dplyr::first(x=rain)) | is.na(x=dplyr::first(x=roll_sum_rain)), yes=NA, no=dplyr::first(x=doy_366, default=NA))), NA, n() > 0)", result_name="start_rain_status", save=2)
```

Maybe there's a more logical way for this that my brain cannot get at the moment!

MeSophie:
@lilyclements Concerning part a) the years of dodoma begin in 1935 and the poster results begin in 1940, so we don't know what happened between 1935 and 1939. I think that is what @rdstern wants.

rdstern:
@lilyclements and @MeSophie if you look above in the examples the big problem remains.  
a) The years with FALSE are omitted from the results.  So quite a lot of years are omittied from the summaries.  I would like them there, with the date and day number given as NA and the logical as FALSE.  Looking above, I don't see any FALSE there at all.  (This is roughly the column summaries option when no combinations are dropped.)
b) When there are missing values in the data (in Dodoma the last 2 years), then all 3 columns should be NA.  Currently the logical seems to be TRUE.

lilyclements:
>  Concerning part a) the years of dodoma begin in 1935 and the poster results begin in 1940, so we don't know what happened between 1935 and 1939.

To a) @MeSophie we're missing the years where there is no occurrence due to nothing that fits the criteria (which is 1935-1939, but also, 1942, 1943, etc). We expect to have `NA` as the start date and `FALSE` as the occurrence here.
In other words, the `drop = FALSE` is not working.
@rdstern @MeSophie, this is working for me. Have you set Year to be a factor? 
Can you share your script? It will not be working if Year is not a factor.

> b) When there are missing values in the data (in Dodoma the last 2 years), then all 3 columns should be NA. Currently the logical seems to be TRUE.
To b) - I've commented on why this is above in the R code with a potential solution (but I'm hoping there's a sleeker solution. Sometimes my brain forgets how to work.

rdstern:
@lilyclements I'm liking the reason that we need year to be a factor.  Let me try.

rdstern:
@lilyclements and @MeSophie I made year into a factor and that's really nice now.  It needs the missing option and we are done with this important improvement.

Except I prefer year to be numeric for later work as it is important in the PICSA graphs and they become annoying if year is a factor.  @lilyclements why can't the code include an `as.factor `adjustment in the R code?  Then it stays numeric in general, but is treated aas a factor here.

Then I am ready to merge with just the start of the rains finished.  I have realised the end of the rains/season is rather different and will discuss that in the issue.

lilyclements:
@rdstern because of how the calculation system works, to set as a factor, we would have to make changes in the back end. There will be a nice way to do this, and I've spent a few hours now trying to understand the system and work it out.

I have a "yucky" solution:

Amend the summary_functions.R file and add the following in after line 12. 
```
  # if the class is different, set to be the same or throw a useful warning
  if (class(curr_data[[by]]) != class(new_data[[by]])){
    warning(paste0("Type is different for ", by, " in the two data frames. Setting as numeric in both data frames."))
    if (class(curr_data[[by]]) == "factor"){
      curr_data[[by]] <- as.numeric(as.character(curr_data[[by]]))
    } else if (class(curr_data[[by]]) == "numeric"){
      new_data[[by]] <- as.numeric(as.character(new_data[[by]]))
    } else {
      stop(paste0("Type is different for ", by, " in the two data frames."))
    }
  }
```

Then, what we would need to do in the dialog code is:

1. Save the class that year currently is
2. Convert `year` to be a factor
3. Run the usual `start of rains` calculation
4. Convert `year` back to the class we saved it as in `1.`

That is,

```
year_type <- data_book$get_column_data_types(data_name="dodoma", columns="year")
data_book$convert_column_to_type(data_name="dodoma", col_names="year", to_type="factor")

# Then the usual SOR code

data_book$convert_column_to_type(data_name="dodoma", col_names="year", to_type=year_type)
```

@rdstern what are your thoughts?


### Other Ideas (more for my benefit I think so I can find this when I pick it up next!)
I'm meanwhile trying to work on a much nicer solution. The issue is that calling it `factor(year)` vs `year` throws an error if I do, say, 

```
grouping_by_year <- instat_calculation$new(type="by", calculated_from=list("dodoma"="factor(year)"))
```

Then I've changed it to say in the R code that if it detects this `factor(year)` to read it as `year` for the naming purposes, but it's doing that for the usage too so just undoing the fact we made it a factor! I'm not sure the best approach. I guess I would want to have something more internal. Probably a conversation with David when he's back from Kenya, in two weeks, so I have a temporary solution for the time being given above?

rdstern:
@lilyclements I'm sorry this is giving you so much trouble.  I havwe been pondering an alternative solution, but I fear we may need your change - hopefully eventually neat!

I realised that usually we don't do start of the rains as the first (or only) summary.  So usually we will have all the years in the annual data frame.  So I tried that and first keeping the year as numeric.

That works as before with the logical being TRUE correctly and NA for the FALSE and NA conditions.  Then I made year into a factor in the daily data and that gives an error - shown below.  (We don't need to fix it, because the reason is obvious!

image

I just wonder if there could be a different way of "tweaking" the calculation, when the logical variable is included.  Would it be possible to check on the start date, which may be a)  really missing or b) not yet happened - remember we can have multiple missing value codes in R as well as SPSS or SAS or Stata!  If it is the second type of missing, then set the logical to FALSE!?

lilyclements:
> Maybe there's a more logical way for this that my brain cannot get at the moment!

@MeSophie and I discussed this earlier and managed to get to that much more logical, neater, sleeker way (for the issue of occurrence putting `TRUE` when we had no data). I knew my brain wasn't functioning properly!

We simply want to change
```
start_of_rains_status <- instat_calculation$new(type="summary", function_exp="n() > 0", result_name="start_rain_status", save=2)
```

to be 

```
start_of_rains_status <- instat_calculation$new(type="summary", function_exp="ifelse(n() > 0, ifelse(all(is.na(roll_sum_rain)), NA, TRUE), FALSE)", result_name="start_rain_status", save=2)
```

That is just changing the `function_exp` parameter value, and it can be hard coded in because `roll_sum_rain` is the name that we internally give it!

@MeSophie are you happy to do this?

rdstern:
@MeSophie and @lilyclements I am not sure if this was yet ready for me to check again?  That's after Sophies change late Friday.

It is still as before.  
image

a) The NAs in the status need to become FALSE.
b) The TRUEin the status, when the data are NA, need to become NA!

lilyclements:
@rdstern I think you were a little premature with your review! I think @MeSophie needs to implement the changes outlined in 1. and 2. from my previous comment still.
@MeSophie thank you for merging in my changes to the `summary_functions.R` file

Issue #9118, opened 2024-08-29
Possible change in the recovery options after a crash?


rdstern:
If there is a crash then R-Instat detects this on the restart and you can try to save the log file or the data from the last backup, before really starting R-Instat.

@N-thony I suggest we continue with this detection, but stop then using the dialog that follows?  This needs perhaps more discussion, but the new Tools > Restore Data is now working well.  So I propose we no longer need both options.  Instead we could simply provide the message with the suggestion and people then click to continue.

The message could now be something like:
We have detected ....
The `Tools > Restore Backup `dialog allows a return to the backed-up data, or to save the corresponding log file.  Click Yes for this option..
Yes   No 
If Yes, then go into R-Instat and run  the Tools > Backup Data dialog.

rdstern:
@N-thony when I first checked, it still gave the same options as before, i.e. you could still use the original recovery (before going into R-Instat)  and it still warned you that this step was sort of final?  

Could you show me a scewwn shot of the message it gives now?

I find it tricky to crash on purpose!

rdstern:
@N-thony exactly and I suggest that was ok before we had the new improved stuff in the tools menu!  

So now (as I proposed above) I suggest we stop the Yes option and replace that message by something like the one I gave above.  Wgat do you think?

rdstern:
So I crashed and got this and then it said this:

![image](https://github.com/user-attachments/assets/bc6a14bc-c957-43bc-9ca4-c259dbae0133)


That is no longer true is it?

rdstern:
@N-thony this can be closed I think?  Please check, but I am doing that.

PR #9138, opened 2024-09-17
Recovery improvement

N-thony:
Additional fixes to #9118 
@rdstern as discussed yesterday, I made the changes to recovery message when R-Instat has crashed, and opening the Tools > Restore Dialogue if you click Yes. Have a look.

### DIFF
--- a/instat/dlgRestoreBackup.vb
+++ b/instat/dlgRestoreBackup.vb
@@ -198,6 +198,14 @@ Public Class dlgRestoreBackup
         End If
     End Sub
 
+    Public Function GetScript() As String
+        Return strScript
+    End Function
+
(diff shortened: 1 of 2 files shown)


rdstern (approved):
@N-thony that looks fine to me now.
@Patowhiz please can you test as well as checking the code.  We used to have recovery before going into R-Instat.  Now you can recover anyway, with Tools > Recover, even if you didn't crash.  (Maybe you just forgot to save the data.) 

But when you do crash (try ctrl alt delete for example) then usinf this dialog instead of the previous one seems much cleaner.  I hope you agree.

ChrisMarsh82 (approved):
Code for this looks fine. I have left the functionality testing to @rdstern

PR #9139, opened 2024-09-18
Additional library for R-Instat

### LINKED ISSUE #7064: Possible change in location for the packages in R-Instat?
One powerful feature in R-Instat is in the Tools Menu where you can import a new package into R-Instat.  The dialogue is useful and I often also use it to check whether a particular package is already installed.  If so, is it up-to-date.
This works fine on the development (merged) version of R-Instat.  It is also a very nice feature (even nicer!) in RStudio - as we might expect.
However in the "proper" version of R-Instat it isn't quite so nice.  You have to be an administrator to install it.  That's not difficult to arrange on my machine, but it isn't so easy for everyone, and it isn't needed in RStudio.  Could we store the packages in a different location - on installation - so packages could be added that bit more easily?
 I think improved installation is to be considered soon, and perhaps this element could be included.  I find the current installation is pretty easy, but the pain is that we have no upgrading facility.  You just have to download and install the whole of the next version and then delete the old one.  Almost all other software does upgrading more smoothly.  (This is important for us, because we claim to be appropriate for countries with limited internet access.  We are mostly, but expecting people to upload 700 mbytes each time is an exception.)

N-thony:
Fixes #7064
@ChrisMarsh82 As discussed yesterday,, I added an additional package directory to R-Instat where no administrator privilege will be required. The directory will be located in something like `C:\Users\Antoine\AppData\Roaming\R-Instat\library`.
@rdstern @ChrisMarsh82 this will need a new release to be able to test. I also made sure that R-Instat looks for package in the two libraries i.e the one in something like `C:\Program Files\R-Instat\0.7.9\static\R\library` and `C:\Users\Antoine\AppData\Roaming\R-Instat\library`.

### DIFF
--- a/instat/static/InstatObject/R/InstallPackages.R
+++ b/instat/static/InstatObject/R/InstallPackages.R
@@ -1,6 +1,13 @@
 Sys.setenv(TZ='GMT')
 Sys.setlocale("LC_TIME", "C")
 
+# Define the custom package library path (in AppData, no admin rights needed)
+app_data_lib <- file.path(Sys.getenv("APPDATA"), "R-Instat", "library")
+dir.create(app_data_lib, recursive = TRUE, showWarnings = FALSE)
+
+# Add the custom library path to .libPaths() for installing user-level packages
+.libPaths(c(app_data_lib, .libPaths()))
+
 #Install packages from CRAN archive
 install.packages("http://cran.r-project.org/src/contrib/Archive/signmedian.test/signmedian.test_1.5.1.tar.gz", repos=NULL, type="source")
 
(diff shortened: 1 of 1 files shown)


rdstern:
@N-thony this would be great.  
a) I assume it looks in the Antoine/roaming one first and then the main R-Instat one?  
b) And when there is a new installation, the Antoine/roaming directory is empty?
c) So when there is an update there is a new (empty) Antoine update? 
d) And when R-Instat is removed, then it removes the Antoine directory as well?

rdstern:
@N-thony please check a) with Chris if you like, but it should look in the new (Antoine) directory first.  Remember that's the one the user can write to.  So updates will be there and they should be used rather than the older version in the main directory.

rdstern:
@ChrisMarsh82 and @Patowhiz could you approve and merge?

ChrisMarsh82 (approved):
This looks fine, although checking should be done once merged. 
To check:
On initial install all r-instat installed r packages are in the program library files
If a new package is installed the appdata library is used (regardless of admin access)

rdstern (approved):
@N-thony I am approving, following Chris Marsh, on the grounds that this is one we can only really check once the version is made!

PR #9150, opened 2024-09-23
Creating additional library on vb.net level rather than R

N-thony:
@ChrisMarsh82 this follows our discussion on Mattermost, the create of the additional folder for package will be done on vb level now. Have a look.

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -411,7 +411,7 @@ Public Class frmMain
 #If Not DEBUG Then
             ' Add the custom library path to R's .libPaths for user-level package installation
             Dim strScript As String = $".libPaths(c('{strLibraryPath.Replace("\", "/")}', .libPaths()))" & Environment.NewLine &
-                   "if (length(.libPaths()) >= 2) {
+                   "if (length(.libPaths()) > 2) {
                          current_paths <- .libPaths()
                          valid_indices <- c(1, 3)[c(1, 3) <= length(current_paths)]
                          .libPaths(current_paths[valid_indices])


ChrisMarsh82:
@N-thony This still would not work. InstallPackages runs in the Github actions to create the R environment that gets copied over. It will not run on the installed machine. The code needs to run once R-Instat is installed. This could be the first time it is run

ChrisMarsh82:
@N-thony Are you sure this will work? If you add the folder to the last install (the build you made to originally to test this change) does R start using that path? 
Does it need also to run the r-script at this point?

rdstern:
@N-thony when installing R-Instat the R packages are all installed in Program Files.  App Data is likely to be empty at the start.  Then R-Instat looks first in App Data.  And App Data is where the Tools > Install R Package dialog will send updates and additional packages.

@N-thony and @ChrisMarsh82 the one item I am unclear about is what is planned when R-Instat is updated next time?  Does each new installation have its own App Data or is there just one?  And when we uninstall a version, does it also delete the App Data directory.  (For R-Instat users I suggest we may want to link the App data to a particular version only?)

ChrisMarsh82 (changes requested):
Add the r script into the if statement where we check the library is created. Then the R code will not run every time R-Instat is started

ChrisMarsh82:
@rdstern it is only set up to have one library for any of the r-instat versions. This means if they upgrade they keep any of the packages they have added. It also makes it easy to delete the library if they are experiencing issues with R-Instat. 
I personally think it would be overkill to have different libraries for different versions as I would hope there is no need for users to have multiple versions of R-Instat. 
This being said I would be happy if you overrule me on this as you understand user needs more than me.

rdstern:
@lilyclements I wonder if you are up-to-date on this innovation to have the main R-Instat library, in program files, as now on installation, and to have a sec,ond location, (in App Data)  where we can add extra packages, or update packages.  We don't need to have admin privedges any more to add packages, or to update packages, which is great.

When searching for a package R-Instat will look first in App Data and then in Program files.

This is all great.  This is also to be a single location that remains when R-Instat is upgraded.  I am very happy with all that, but with one caveat.  Say, in the original installation we have package aaa, version 0.7, (in program files) and we update to version 0.8 in app data.  All is fine.

Now we upgrade R-Instat and package aaa is now version 0.9.  So now App Data is behind the program files.  Is there a way we can manage that, and delete the version in App Data, if it is an older version than a copy in program files?
 
@ChrisMarsh82 am I worrying unnecessarily?  Is this a non-problem?

lilyclements:
> Is there a way we can manage that, and delete the version in App Data, if it is an older version than a copy in program files?

@rdstern while I'm not sure how we can do this on the technical side outside of R (I think that is much more @ChrisMarsh82's domain and why he's involved!), it is absolutely possible to uninstall packages to then reinstall particular package versions using `install_version`:

E.g. for package `aaa`
```
remotes::install_version("aaa", "0.8")
```

Is this the sort of thing you're looking for, or is it a bit more technical than this?

rdstern:
@lilyclements in the "old" system you can install packages as an administrator.  The next update will ignore those packages, and you will have to install them again.
One option. with the new system, would therefore be to empty app-data when installing a new version.  That would be safe and no worse than the current system.  I think I'd be inclined to do that?

rdstern (approved):
@ChrisMarsh82 I am approving - so you have all the responsibility!

ChrisMarsh82:
> > Add the r script into the if statement where we check the library is created. Then the R code will not run every time R-Instat is started
> 
> @ChrisMarsh82 Please correct me if I'm wrong. As we are doing this, it seems that we will have to add the additional in every new R or R-Instat session. Moving this in the if statement where we check the library means we will add this only in the current R or R-Instat, in the new session the AppData library won't be added in the `.libPaths()`. Can you test this too? Thanks

@N-thony I don't see why we need to do this in every session. In R if you add a library you don't loose the library if you shut down and re-open. They shouldn't be a reason why that would be the case in R

ChrisMarsh82 (approved):

ChrisMarsh82:
Happy with this. You can look at doing it with the bash script if you think that would be better but not sure how this would work and you might need to look into the innosetup scripts to do so.

PR #9153, opened 2024-09-25
Formatting Script

N-thony:
fixes partly https://github.com/IDEMSInternational/R-Instat/issues/8847
@rdstern @derekagorhom have a look of the formatting of the script.
image
image

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -19,6 +19,7 @@ Imports System.IO
 Imports System.Windows.Controls
 Imports RInsightF461
 Imports ScintillaNET
+Imports RDotNet
 
 Public Class ucrScript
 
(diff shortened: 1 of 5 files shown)


derekagorhom:
@N-thony I tested this PR and it works, it also fixes the other issues with regards to my PR. 
@rdstern maybe able to give a more detailed review than me

rdstern:
@derekagorhom I must be missing something?  I loaded a script and pressed right-click and here is what I got:

image

I don't see any difference, or formatting options?  What should I have done?  Should I have installed a package?

derekagorhom:
@rdstern The reformat option in on the Edit Menu now

rdstern (approved):
@N-thony that's great. 
I like the right-click, and I like it being called Reformat Code.
Do we still want the Format Code in the edit menu?  If so, should there be a line above and should it be Reformat Code for consistency with the right-click option?

I am approving while you consider this.  @Patowhiz could you check, react and merge?

rdstern (approved):
@Patowhiz this should be an easy check and merge?

instat/ucrScript.vb
```diff
@@ -1025,4 +1068,7 @@ Public Class ucrScript
         sender.Dispose()
     End Sub
 
+    Private Sub mnuReformatCode_Click(sender As Object, e As EventArgs) Handles mnuReformatCode.Click
+        FormatRCode()
```
Patowhiz, line 1072: @N-thony if `FormatRCode()` is used once, then just put it's functionality inside the `mnuReformatCode` click event.
Same with `EscapeDoubleQuotes()` function. It's just one line and if used in one place only, just call the single line where it is used.
This will simplify the functionality added consequently making it easier to maintain.

Patowhiz (changes requested):

rdstern (dismissed):

instat/ucrScript.vb
```diff
@@ -1025,4 +1026,37 @@ Public Class ucrScript
...
+            expTemp = frmMain.clsRLink.RunInternalScriptGetValue(clsStylerFunction.ToScript(), bSilent:=True)
+            If expTemp IsNot Nothing AndAlso expTemp.Type <> Internals.SymbolicExpressionType.Null Then
+                formattedCode = expTemp.AsCharacter().ToArray
+            End If
+
+            ' Join the formattedCode array into a single string
+            Dim formattedText As String = String.Join(Environment.NewLine, formattedCode)
+
+            ' Check if there is any selected text
+            If clsScriptActive.SelectionStart <> clsScriptActive.SelectionEnd Then
```
Patowhiz, line 1050: @N-thony I don't quite understand the need for this check. Line 1023 `Dim scriptText As String = clsScriptActive.SelectedText.Replace("""", "\""")` alreadu assumes there is selected text. That's why it is formatting the it. Why do you need to check for selection again? I though if there is no selection, formatting shouldn't happen.
If there is a case when you are allowed to format nothing, shouldn't the test for selection be done before line 1023 is executed?

instat/ucrScript.vb
```diff
@@ -1024,5 +1025,37 @@ Public Class ucrScript
...
+
+            Dim expTemp As SymbolicExpression
+            Dim formattedCode As String() = Nothing
+            expTemp = frmMain.clsRLink.RunInternalScriptGetValue(clsStylerFunction.ToScript(), bSilent:=True)
+            If expTemp IsNot Nothing AndAlso expTemp.Type <> Internals.SymbolicExpressionType.Null Then
+                formattedCode = expTemp.AsCharacter().ToArray
+            End If
+
+            ' Join the formattedCode array into a single string
+            Dim formattedText As String = String.Join(Environment.NewLine, formattedCode)
```
Patowhiz, line 1051: What will happen when `formattedCode` is nothing? I think you can just place lines 1051 and 1054 inside the if block that checks for `expTemp`. For instance,

```
If expTemp IsNot Nothing AndAlso expTemp.Type <> Internals.SymbolicExpressionType.Null Then
                 Dim formattedCode As String() = expTemp.AsCharacter().ToArray
 ' Join the formattedCode array into a single string
            Dim formattedText As String = String.Join(Environment.NewLine, formattedCode)

' Replace the selected text with the formatted text
            clsScriptActive.ReplaceSelection(formattedText)
 End If
```

Then I think you won't need the try catch. You could have a message box in the else block that notifies users when formatting could not be done.

Patowhiz: But that's what your try catch will do when `formattedCode` is nothing. In you code, you are changing `formattedCode` to a concatenated string even when it's nothing. That means an exception will be raised and the catch block will be executed, thus not 'providing your smoother experience'. That's why I asked about it. If the 'disruptive message' is not needed then you don't need the else block, that's why it's omitted in my code example above.

Patowhiz (dismissed):
@rdstern the code now looks good. Kindly retest then I can merge.

Patowhiz (approved):
@rdstern over to you to review.

rdstern (approved):
@N-thony and @Patowhiz that's very nice!

Issue #9143, opened 2024-09-19
Add the Table Options button to the Prepare > Data Frame > View dialog


rdstern:
@Patowhiz I assume this is a simple task, once your Presentation (sub-)dialog is merged?  I assume anyone can do this, so @N-thony can allocate.  It would be good to include this in the next release - for 26th September?  I suggest it can be allocated now.  

Here is half the current dialog:

![image](https://github.com/user-attachments/assets/5d095be2-9871-43d1-8fa4-b0c35ed270bc)

a) I suggest we get rid of the radio buttons at the top.  It is very easy from here to choose all the columns.  Agreed?
b) So we just have this half of the dialog, without those top radio buttons.  
c) In the last option, delete Window, so it says `Display in R Viewer`

Here is the dialog with the 2nd option chosen:

![image](https://github.com/user-attachments/assets/45650df7-2f3e-4015-bd60-35b4688ae4b9)

d) For both the first 2 options (html and output window)  have the Number of Rows control and move it up, so it is aligned with the multiple receiver and just below it.  Make the default 6 (assuming the number of rows in the data frame is at least 6.  The upper limit is already set to the length of the data frame - which is good.

e) For the html option make the new Table Options button visible.  It is the same Table Options button and call that is in @Patowhiz Describe > Tables > Presentation Table dialog.  It calls the same sub-dialog.

f) For the Output window option, delete the radio buttons.  Replace them with a checkbox - default checked, with label `Display from Top`.  So the default is to display the top lines.  If you uncheck it, then it displays from the bottom, e.g. the last 6 lines.

rdstern:
@Fidel365 the Presentation Tables dialog has now been merged.  So I hope you are ready to work on this dialog?  Maybe you already started?

Patowhiz:
@rdstern, I'm uncertain why we have HTML display options here when they already exist in the presentation dialog. I see this dialog as a 'prepare' dialog, intended for exploring the data frame rather than presenting it (which I view as a 'describe' activity).

If the original purpose of this dialog was indeed for 'preparatory' activities only, I’d suggest removing the HTML display and store options entirely (as I believe the store option is primarily for what you've 'described'). This would allow us to streamline the dialog's functionality specifically for 'preparatory' tasks, including the ability to view the full data set.

I also don’t support removing the top radio buttons, as I think the selector currently only shows unhidden columns (please correct me if this functionality has changed recently).

[...]

PR #9154, opened 2024-09-25
Adding the table options button to Prepare>Dataframe>View dialog

Fidel365:
Fixes partially issue #9143 
@rdstern this PR is ready for review, partially bacause there are more changes to come on the table options button.


rdstern:
@Fidel365 it is almost there already.  And I got it a bit wrong.  Here is the Describe > Tables > Presentation dialog:
image

a) Please change Format Table in the Html option to `Table Options`.
b) Change your Label Number of Rows to a checkbox that says just `Rows`.  Default is checked with the 6 as now.  If unchecked then the up-down is no longer visible.
c) The Rows checkbox also applies to the html option.  (Just as in the preview option - which we will change to Rows) in the Presentation dialog.
d) So you can delete  the Specify Rows checkbox in the Output window option.
e) Hence the `Display from Top` checkbox can move up to be level with the Output viewer option.

Thanks

Roger

Fidel365:
We move to PR #9174

PR #9163, opened 2024-09-30
Bug fix for new machine running R-Instat

N-thony:
@rdstern @jkmusyoka this is now fixed. I noticed this was due to poor code review. We have to be more carefull in reviewing

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -511,11 +511,11 @@ Public Class frmMain
                 'get the data file path if selected by the user
                 strDataFilePath = dlgRestoreBackup.GetDataFilePath()
             End If
-        End If
 
-        Using writer As StreamWriter = New StreamWriter(strMarkerFilePath, False)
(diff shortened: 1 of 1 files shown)


instat/frmMain.vb
```diff
@@ -511,11 +511,11 @@ Public Class frmMain
...
             End If
-        End If
 
-        Using writer As StreamWriter = New StreamWriter(strMarkerFilePath, False)
-            writer.WriteLine("Running")
-        End Using
+            Using writer As StreamWriter = New StreamWriter(strMarkerFilePath, False)
+                writer.WriteLine("Running")
+            End Using
+        End If
```
Patowhiz, line 518: @N-thony other changes in spacing/tabs what is the difference between what was previously there and the new changes? I can't spot the difference.

Patowhiz:
I've just noticed you have closed this. If you plan to create a new, would be to indicate the issue it fixes. Thanks.

PR #9165, opened 2024-10-01
Bug fix for new machine running R-Instat 

### LINKED ISSUE #9164: The changes in directories in the R-Instat merged version and an explanation of the recent problems?
@N-thony I have had a question from Beryl, who has the same problem I have, in running the merged version from Visual Studio.
Earlier, the merged version correctly picked the R directory with all my R files.  Now you have added roaming, I assume it should have the correct R directory as the secone one.  

But it chooses an incorrect directory.  My guess is that it does the same for Beryl, and maybe for others.

So, on my machine, the current merged version gives the following result from running `.libPaths() `command  once I run:
```
[1] "C:/Users/rdste/AppData/Roaming/R-Instat/library"
[2] "C:/Program Files/R/R-4.1.3/library"             

```
Here the first is now your new roaming directory.  The second is the incorrect R library directory and that is incorrect.  Maybe that's because I have 2 versions of R on my machine?  It should choose the directory it used earlier, that has all the R packages.  This, on my machine is:

`.libPaths(new= "C:/Users/rdste/Documents/R/win-library/4.1" )  `

That's the directory on my machine that the merged version used earleir.  Ideally the merged code would now have that as the second package instead of the 4.1.3 directory - that has just has a few R packages - not many.

I now have to correct the package location each time I run the merged version, or any individual branch.  This is boring.

At the same time please - once the latest version has been fixed - please add a detailed explanation of the latest fix, to give @jkmusyoka and me, re-assurance that you have a full solution to our recent installation problems.  This has now been a problem in the installation in 5 countries.

@N-thony and @ChrisMarsh82 this all started with the Malawi and Zambia installations. R-Instat worked on our machines, and some participants and then gave the dreaded:

image

on some desktop machines.  Usually those with no previous R-Instat and possibly no R on the machines.

We thought this was fixed, but the same problem occurred in Zimbabwe.  You fixed overnight - many thanks.  I assumed the problem was over.

But now - in both Bangladesh and Uganda, we have the same problem again.  And it still seems that we can't test, because it is still not a problem on our machines.  

This needs to stop, and we also need to understand it more so we can explain what is happening.  Please provide the explanation of the problem and the fix.  And please can you re-assure us that this fix will work on any machine, independent of their past R and R-Instat history.  I have become so nervous with each installation - it is becomeing a nightmare!

I have the impression it is mainly related to having a writeable directory for extra packages.  If necessary let's abandon that idea, and stick to administrator rights being needed.  But we do still need to ensure the R-Instat directory (or directories)  is separate from those used by R.  

@N-thony I trust your fix tomorrow(now today) morning will work.  I look forward to downloading and installing as soon as possible.  Of course with a gigabyte to download that is another challenge!  We must return to a more professional level of product.

Vitalis95:
Replaces PR #9163 
Partially fixes #9164
@N-thony @rdstern  have a look it

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -511,11 +511,11 @@ Public Class frmMain
                 'get the data file path if selected by the user
                 strDataFilePath = dlgRestoreBackup.GetDataFilePath()
             End If
-        End If
 
-        Using writer As StreamWriter = New StreamWriter(strMarkerFilePath, False)
-            writer.WriteLine("Running")
-        End Using
+            Using writer As StreamWriter = New StreamWriter(strMarkerFilePath, False)
+                writer.WriteLine("Running")
+            End Using
+        End If
 
         '---------------------------------------
 
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -511,11 +511,15 @@ Public Class frmMain
                 'get the data file path if selected by the user
                 strDataFilePath = dlgRestoreBackup.GetDataFilePath()
             End If
-
-            Using writer As StreamWriter = New StreamWriter(strMarkerFilePath, False)
-                writer.WriteLine("Running")
-            End Using
         End If
+
+        If Not Directory.Exists(strAutoSaveLogFolderPath) Then
+            Directory.CreateDirectory(strAutoSaveLogFolderPath)
+        End If
+
+        Using writer As StreamWriter = New StreamWriter(strMarkerFilePath, False)
+            writer.WriteLine("Running")
+        End Using
 
         '---------------------------------------
 


instat/frmMain.vb
```diff
@@ -511,11 +511,11 @@ Public Class frmMain
...
             End If
-        End If
 
-        Using writer As StreamWriter = New StreamWriter(strMarkerFilePath, False)
-            writer.WriteLine("Running")
-        End Using
+            Using writer As StreamWriter = New StreamWriter(strMarkerFilePath, False)
+                writer.WriteLine("Running")
+            End Using
+        End If
```
Patowhiz, line 518: @Vitalis95 @N-thony I still see no difference in the code other than the tab/spaces introduced. Kindly explain to me what has been introduced here. 
Give my ignorance on this issue of new machine running R-Instat, why do we need write "Running" on a file?

Patowhiz (changes requested):

rdstern (approved):
@Patowhiz I don't know how to review this, but we do need a fix.  I'm therefore approving blind on the assumption this is the right thing to do, and we can test the result here.

rdstern:
@N-thony and @Patowhiz please produce a new fixed version.  We are in the middle of training here.

Patowhiz (approved):
I'm approving this. Good to tag @ChrisMarsh82 for review in all future tasks that relate to debug, log and library directories. He is more familiar with recent changes than me.

Issue #9136, opened 2024-09-14
Addition of a new Distances dialog


rdstern:
I suggest this (at least for now) to go at the bottom of the Climatic > Check data menu.
It can have the label distances:

It is designed for the station data file with climatic metadata.  Often the Long and Lat variables will have been defined earlier.

a) Usual data selector.
b) Two single receivers, one labelled `Long: ` the other `Lat:` If defined in the climatic, these are filled automatically. 
c) A Group box labelled `From:`.  Witrhin it are 2 radio buttons, The first is labelled `Values `and underneath is the second labelled `Variable`.  
d) If Values are selected then 2 controls are opened, labelled `Long` and `Lat`.  A number can be typed into each field.  For Long there is a check that it is between -180 and 180.  for Lat it is between -90 and 90.  
e) If Variable is chosen then a single receiver is opened for the variable to be chosen.    Underneath is a receiver where you choose a value from that variable.  This is a drop-down.  You choose one value from that variable.  (It can be any variable, not necessarily a factor.  But the value chosen must be unique.  Perhaps that check can wait, or even that option can wait?
f) Then the standard store result in control, with ther Position possibility.  It produces a new column.
g) Then the standard base buttons.

It uses the geosphere packager that is already installed, see[ here](https://stackoverflow.com/questions/72719074/how-to-calculate-the-distance-between-points-and-a-fixed-location-in-r-studio).

I'll give more details

rdstern:
@Vitalis95 that's very good, and already seems to work.

a) Could you please change the up-down controls, so you just type a number.  They are such a pain to use.  I would then be quite happy if the checks on the acceptable values are only when you type ok.

b) I also assume the answers are in metres.  Is that correct?  If so, then could you divide them by 1000 to give the answers in kms.

PR #9166, opened 2024-10-01
Added Distances dialog to Climatic > Check data

Vitalis95:
Fixes partly #9136 
@rdstern @N-thony , implemented the first part - `Values`


Vitalis95:
@rdstern , have a look at it

rdstern (approved):
@Vitalis95 great, I am approving.
One small question below.  but @N-thony can you check and merge if ok?

Here are some results for Malawi:

image

They are given with 4 decimal places, so that is the distance to the nearest 10cms!  We don't really need that accuracy.  

image
Is it easy to round to the nearest 10 metres, so 2 decimals?

Vitalis95:
@N-thony , let me do it here then you can merge

PR #9175, opened 2024-10-07
Added Fraction of Evaporation to the Start of the Rains Dialog

### LINKED ISSUE #9144: Add fraction of evaporation to the start of the rains dialog
@lilyclements and @MeSophie this is the latest request from CIMH of something we had in the original Instat that is not yet in R-Instat.  Here is the dialog in the old Instat:

image

@volloholic suggests it would be (very?) easy to add to R-Instat.  

If we do, then I assume we will have an evaporation variable in the data with no missing values.  We don't give the option of a constant.  Instead, if constant, we just make a variable with that value.  If there are missing, then (before using the dialog), we suggest using replace to insert a sensible value, e.g. 5mm per day.  If we only have the values for a single year, then we use Merge to stretch it out for every year.

Then the calculation gets the moving total for the rainfall exactly as it does now.  It does the same for the fractional moving total of the evaporation.  Then the start is that first day (each year) that the moving rainfall total exceeds the moving evaporation total.

Now here is the current start of the rains dialog:

image

I'm not clear just how to add this extra component.  This is also the only dialog that is so complicated.  So I wonder if this is our occasion to change the structure of the by adding another sub-dialog button.  Could we perhaps keep the first line - with all possibilities - on this dialog and then have a Further Conditions button for the other 3 conditions?  Maybe after the Further Conditions button - if one or more is specified, then the word Specified appears on the main dialog after the button?

Then for Evaporation option (on the main dialog, we have a `Fraction` up-down control, with a default of 0.5 and steps of ,01 with lower limit of 0.01 and upper limit of 5.  Then a single `Evaporation: `receiver for the evaporation variable.

### LINKED ISSUE #9185: Error messages from start of rains analysis using shifted climatic data
@N-thony for you to check quickly. Seems something small to fix I hope

There are 2 error messages that have popped up during the start of rains analysis with shifted data here in Eswatini.

The first message below appeared when opening the start of rains dialog. Clicking OK still opened the dialog

image

Submitting a filled-in start of rains dialog then gave the following second error message. Again closing the message box still gave the output in a new dataframe as expected.

image

The dialog seems to work despite the errors. But we do need to find out why the errors and fix the issues

MeSophie:
Fixes #9144
Fixes #9185
@rdstern  @lilyclements  @N-thony  I added fraction evaporation and Add sub dialog on Start  of Rain dialog. Please have a look

### DIFF
--- a/instat/dlgStartofRains.Designer.vb
+++ b/instat/dlgStartofRains.Designer.vb
@@ -39,24 +39,13 @@ Partial Class dlgStartofRains
     <System.Diagnostics.DebuggerStepThrough()>
     Private Sub InitializeComponent()
         Me.grpConditionsForSatrtofRains = New System.Windows.Forms.GroupBox()
-        Me.ucrNudDPOverallInterval = New instat.ucrNud()
-        Me.ucrNudDSMaximumDays = New instat.ucrNud()
-        Me.ucrNudDSLengthOfTime = New instat.ucrNud()
-        Me.lblDPOverallInterval = New System.Windows.Forms.Label()
-        Me.ucrNudRDMinimumDays = New instat.ucrNud()
-        Me.ucrNudDPRainPeriod = New instat.ucrNud()
-        Me.lblDPLength = New System.Windows.Forms.Label()
-        Me.ucrChkDryPeriod = New instat.ucrCheck()
-        Me.lblDSLengthofTime = New System.Windows.Forms.Label()
-        Me.lblDPMaxRain = New System.Windows.Forms.Label()
-        Me.ucrNudRDOutOfDays = New instat.ucrNud()
-        Me.ucrNudDPMaxRain = New instat.ucrNud()
-        Me.ucrChkDrySpell = New instat.ucrCheck()
-        Me.lblDSMaximumDays = New System.Windows.Forms.Label()
-        Me.lblRDMinimum = New System.Windows.Forms.Label()
+        Me.cmdAdditionnal = New System.Windows.Forms.Button()
+        Me.ucrChkAdditional = New instat.ucrCheck()
+        Me.ucrReceiverEvap = New instat.ucrReceiverSingle()
+        Me.lblFraction = New System.Windows.Forms.Label()
+        Me.ucrNudEvapo = New instat.ucrNud()
+        Me.rdoEvapo = New System.Windows.Forms.RadioButton()
         Me.lblTROverDays = New System.Windows.Forms.Label()
-        Me.lblRDWidth = New System.Windows.Forms.Label()
-        Me.ucrChkNumberOfRainyDays = New instat.ucrCheck()
         Me.ucrNudTROverDays = New instat.ucrNud()
         Me.ucrNudTRAmount = New instat.ucrNud()
         Me.lblTRVal = New System.Windows.Forms.Label()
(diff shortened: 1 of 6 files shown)


rdstern:
@MeSophie this is impressive.  Perhaps @lilyclements may have suggestions to improve on the new layout of the controls.  But it is better than I could have done.
a) I first checked that the original definition is still working.  Pretty good, but I think I have found one problem.  It may have been there before.

1) I tried with dodoma and 20mm in 3 days from 1 Nov to 31 december.  Saved all 3 variables and that worked fine!  Great!
2) Now I added your additional option, of the default dry spell condition.  Named them differently, so we have both - now 7 variables in the summary.  Seems fine - great.
3) Now unticked the Additional, so I'm back to the original, i.e. situation 1.  Renamed again to get a total of 10 columns.  Now I found I got the same results as in 2), i.e. still with the dry-spell condition.  It should have taken that condition away, so I should have got the same as 1).
4) Now I did a reset and tried again.  It now gave me an error, that mentioned dry spells.  So I wonder if it is not clearing properly?  I then went into the sub-dialog, and saw that it had cleared the dialog.  When I now tried again it all worked!

Now I moved to check your new evaporation option.  Here is the dialog I used:

image

And, to my amazement it worked first time.  Congratulations to @lilyclements and @MeSophie.  Just one small thing - of course!  Here are the results:

image

Now I had made the evaporation column using the calculator, to be simply `20/3 `.  So, with the fraction as 1 I should get the same result as before.  And I usually do, but not always!  I wonder why not?  So I had a look at the data, for 1951 when there was a big difference.  I was happy with day 350 for the ordinary start, but the evap had started a lot earlier, when there was one day with 14.7mm.  So I think there needs to be a small tweak in the calculation system for the evap situation.

How exciting to be so close so quickly though!

MeSophie:
image
@rdstern I was unable to reproduce this error. Everything works fine at home. I don't know if Lily will be able to reproduce it.

image
@lilyclements Could you please have a look at this part when you have time? Thank you.

rdstern:
@MeSophie @jkmusyoka has just found an odd error in the start of the rains.  It may (or may not) be due to using the shifted year - I started with Zambia data from August.
I haven't seen it before, and perhaps it is because I went straight to the start of the rains dialog.  However, I wondered if it would be just the first time I used the dialog, but I then got it whenever I repeated, even after the data frame was created earlier.

It is annoying, but it doesn't seem to stop the results being created.  Perhaps you could investigate as you continue to improve this dialog.  By the way I got it with a simple definition of the start, and nothing to do with evaporation.

Thanks

MeSophie:
@rdstern @jkmusyoka We have found a solution for the second error reported in #9185. Please have a look.
We were unable to reproduce the first error.

rdstern:
@MeSophie great that you can include the solution to @jkmusyoka error.  Can you also check with @lilyclements on the problem above, that the current R code seems to be incorrect in the date of the start occasionally.  And @lilyclements may also have some ideas on further improvements in the new layout?

rdstern:
@MeSophie I see there are now conflicts.  Can you resolve?

And have you and @lilyclements been able to resolve the occasional wrong results with the current code?

MeSophie:
> @MeSophie I see there are now conflicts. Can you resolve?
> 
> And have you and @lilyclements been able to resolve the occasional wrong results with the current code?

@rdstern I can't resolve the conflicts. I'm just going to change branch and make the changes again, separating the change made in rInstatTranslations.db from the others to avoid any potential conflicts.

image
Concerning this question @rdstern  please are  you referring to part 4 or to the last comment about the calculator? Because I couldn't reproduce the error in part 4. I also asked Vitalis to try it it from his side, but he couldn't. And if it's about the calculator comment, I think Lily's still working on it. I don't know if she's had time to look at it this week.

rdstern:
@MeSophie and @lilyclements it is about the fact that the result, with the evaporation, is sometimes not what I feel I should expect.  I thought I had contrived it, that the result with the evaporation should be identical to 20 mm in 3 days.  Sometimes it wasn't!

MeSophie:
This PR has been replaced by PR #9194 because of conflicts.

PR #9184, opened 2024-10-09
Fixing sort_dataframe

### LINKED ISSUE #9177: Sort no longer working in Version 0.7.21!
This is the latest bug that @jkmusyoka has fallen into in the Eswatini workshop.  The right-click > sort no longer works!
 
In Version 0.7.21 load any dataset.  I tried with the survey and also with the diamonds.  The problem is the same.
Sort on a numeric variable, eg size for the survey or carat in diamonds.  It is just right-click and the sort option from within a variable.  Or the Prepare > Data Frame > Sort dialog.

It gives the following error message:

image

Checking back it is the same error in Version 0.7.20, but works fine in Version 0.7.19.

I don't think the cause is likely to be an update in dplyr, because that was last updated in November 2023.  Notice at the bottom of the error message it says it is running our `data_book$sort_dataframe` function, so that is presumably the place to look.

When it is found, I hope the solution can also pin-point any other functions that might have become similarly affected?  Why does this keep happening to us!

lilyclements:
Fixes #9177

@rdstern @Vitalis95 thank you for your comments on the issue for this. This has now been fixed, and is ready for review.


rdstern (approved):
@lilyclements many thanks.  Seems ok now.
@N-thony can you check and hopefully merge in time for the next version

Issue #6264, opened 2021-03-13
Addition of a timer to the tools menu in R-Instat


rdstern:
I wonder if it would be really easy to add a timer?  
[Here](https://www.r-bloggers.com/2017/05/5-ways-to-measure-running-time-of-r-code/) is information on five ways this could be done.
I hope this might be very easy to do, as follows:
a) Have Add timing to the tools menu.  Default is unticked .  If you select it, then it is ticked.
b) If ticked, then when you press OK it gives the time in the output window.  Then it also gives the time again as the last command in whatever code it runs.  Ideally it also gives the difference.

rdstern:
You have mentioned that the Wine version seems to run slower than the ordinary version.  I think you are using a Windows machine, so might you have the ordinary R-Instat and now the Wine version on the same machine.
Could you please quantify what you mean by slower?
This could be by installing a timer.  But it could initially be used through the script window.  I suggest we anyway install the tictoc program, which seems very easy, and does exactly what I wanted.  
Once this is installed (which I did), then we can run it through the script window as follows:

I did the one variable summarise command, then copied it to the script window.  Then added tic() before and toc() after!

```
library("tictoc")
tic()
summary(object=data_book$get_columns_from_data(data_name="Data", col_names=c("Name","date","year","month_abbr","doy","day_in_month","tmax","tmin","rain","hmax","hmin","sunh","ws","wd","cumsum","rainday","day5","day20")))
toc()
rm()
```
It then said:
![image](https://user-images.githubusercontent.com/11226469/111548546-c8cf9180-8772-11eb-9d33-3803001aae13.png)

Neat eh?

@maxwellfundi doesn't this tempt you to add the command.  It is just a check in the tools menu!
@TobbyKe aren't you tempted to do some timing trials?

rdstern:
@maxwellfundi in testing for issue #4918 I am afraid I don't see how to do this easily without adding the timer properly to R-Instat.  The main testing will be on the time to refresh the grid change, for various grid sizes.

I wonder if this can be done easily.  If so, then I think TicToc is such a nice name, that we could use that in the menu, rather than the more boring Timer.  Or even TicToc Timer!

rdstern:
@lloyddewit as we may be doing more stress tests, I wonder if it could be convenient to add this into R-Instat.  What do you think?

lloyddewit:
[...]

PR #9189, opened 2024-10-12
Implementing undo in Reogrid

N-thony:
@rdstern, this is still a draft, but I would appreciate it if you could test it and share your feedback. This PR implements an undo operation for any changes made to the dataframe, which will then be reflected as usual in ReoGrid.
Just like in Excel, try anything and then click `Ctrl+Z` or `Ctrl+Shift+Z` to undo the operation on the grid.

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -172,7 +172,7 @@ DataSheet$set("public", "set_enable_disable_undo", function(disable_undo) {
 })
 
 DataSheet$set("public", "save_state_to_undo_history", function() {
-  self$set_undo_history(list(private$data))
+  self$set_undo_history(private$data, attributes(private$data))
 })
 
 DataSheet$set("public", "is_undo", function() {
@@ -209,19 +209,31 @@ DataSheet$set("public", "has_undo_history", function() {
 
 DataSheet$set("public", "undo_last_action", function() {
   
-  # Perform the undo action
+  # Check if there's any action to undo
   if (length(private$undo_history) > 0) {
+    # Get the last state from the undo history
     previous_state <- private$undo_history[[length(private$undo_history)]]
-    self$set_data(as.data.frame(previous_state))  # Restore the previous state
     
-    private$undo_history <- private$undo_history[-length(private$undo_history)]  # Remove the latest state
+    # Restore the data and its attributes
+    restored_data <- previous_state$data  # Extract the dataframe
+    restored_attributes <- previous_state$attributes  # Extract the attributes
+    
+    # Set the dataframe in the DataSheet
+    self$set_data(as.data.frame(restored_data))
+    
+    # Restore attributes
+    restored_attributes <- previous_state$attributes  # Extract the attributes
+    for (property in names(restored_attributes)) {
+      self$append_to_metadata(property, restored_attributes[[property]])
+    }  
+    # Remove the latest state from the undo history
+    private$undo_history <- private$undo_history[-length(private$undo_history)]
     
     # Trigger garbage collection to free memory
     gc()
   } else {
     message("No more actions to undo.")
   }
-  
 })
 
 
@@ -302,32 +314,38 @@ DataSheet$set("public", "set_scalars", function(new_scalars) {
 )
 
 # Set undo_history with memory management
-DataSheet$set("public", "set_undo_history", function(new_undo_history) {
-  if (!is.list(new_undo_history)) stop("undo_history must be of type: list")
-  if(!private$disable_undo){
-  # Define memory and undo_history limits
-  MAX_undo_history_SIZE <- 10  # Limit to last 10 undo_history states
-  MAX_MEMORY_LIMIT_MB <- 1024  # Limit the memory usage for undo undo_history
+DataSheet$set("public", "set_undo_history", function(new_data, attributes = list()) {
+  if (!is.data.frame(new_data)) stop("new_data must be of type: data.frame")
   
-  # Check current memory usage
-  current_memory <- monitor_memory()
-  
-  # If memory exceeds limit, remove the oldest entry
-  if (current_memory > MAX_MEMORY_LIMIT_MB) {
-    message(paste("Memory limit exceeded:", round(current_memory, 2), "MB. Removing oldest entry."))
-    private$undo_history <- private$undo_history[-1]  # Remove the oldest entry
-    gc()  # Trigger garbage collection to free memory
-  }
-  
-  # Limit undo_history size
-  if (length(private$undo_history) >= MAX_undo_history_SIZE) {
-    private$undo_history <- private$undo_history[-1]  # Remove the oldest entry
-    gc()  # Trigger garbage collection to free memory
-  }
-  
-  private$undo_history <- append(private$undo_history, list(new_undo_history))
+  if (!private$disable_undo) {
+    # Define memory and undo_history limits
+    MAX_undo_history_SIZE <- 10  # Limit to last 10 undo_history states
+    MAX_MEMORY_LIMIT_MB <- 1024   # Limit the memory usage for undo_history
+    
+    # Check current memory usage
+    current_memory <- monitor_memory()
+    
+    # If memory exceeds limit, remove the oldest entry
+    if (current_memory > MAX_MEMORY_LIMIT_MB) {
+      message(paste("Memory limit exceeded:", round(current_memory, 2), "MB. Removing oldest entry."))
+      private$undo_history <- private$undo_history[-1]  # Remove the oldest entry
+      gc()  # Trigger garbage collection to free memory
+    }
+    
+    # Limit undo_history size
+    if (length(private$undo_history) >= MAX_undo_history_SIZE) {
+      private$undo_history <- private$undo_history[-1]  # Remove the oldest entry
+      gc()  # Trigger garbage collection to free memory
+    }
+    
+    # Package the new data and attributes into a list
+    new_undo_entry <- list(data = new_data, attributes = attributes)
+    
+    # Append the new entry to the undo history
+    private$undo_history <- append(private$undo_history, list(new_undo_entry))
   }
 })
+
 
 DataSheet$set("public", "set_keys", function(new_keys) {
   if(!is.list(new_keys)) stop("new_keys must be of type: list")


rdstern:
@N-thony I tried on the survey data, but nothing happened when I pressed ctrl Z (or ctrl Shift Z?  Would be great if it did - probably better than my suggestions, mentioned earlier, if this could work.

rdstern:
@N-thony that's pretty amazing  to me!
Undo seems to work, when I am in the grid and press `ctrl + shift + Z`.  I have tried:
a) Edit cell, and delete cell and then undo.
b) Delete a column
c) Rename a column
d) Make a column into a factor

I tried first with the small survey data - 36 rows and 7 variables.
Then with dodoma that has 29000 rows, so lots of pages.  I tried changes on the last page - edit cell, and also, from there, did delete of a column.

I then tried with the MICS survey data, where I made changes to other than the first 50 columns.

I'm not sure what you have done, but it seems excellent.  It would be nice if ctrl Z would work as well, but not a problem if only ctrl + shift + Z is what has to be used.

I tried ctrl  Y for a redo, and couldn't get that to work.  Might that be possible.

I tried 2 changes on the same sheet and I think it only did undo for the first of them.  Is that correct?  I am happy with the one change limit per sheet.  I deleted 2 columns, one at a time, and it got the second one back.  Not the first.  Then I deleted 2 variables in the same delete operation and it got both back.

I tried one change on each of 2 sheets.  Then I corrected the first change I had made.  That worked, and moved me to the second sheet.  I did undo there and it worked too. 

This is much better than my suggestions in an earlier issue.  But I still don't understand how it works - can you explain easily?

Will there be a dialog, or will it just be those keys?  If only the keys, because the focus needs to be in the grid, then that would be fine.  We might then still have a simple dialog that mainly gives access to the help, and tells you to press the keys!

rdstern:
@N-thony exciting and good to have you back in this way.  before I test, I wonder on your views on redo?  Might that also be possible?
If you do a dialog, then I assume it would be in the Edit menu?  Where?  Should it be under Find?  Anyway, if you do, the nthe Help context number is 702.

rdstern (commented):
@N-thony it is still looking excellent.  I enthused to Danny, who said he hopes it is being done through R, or he is not interested.  I think so, as I see you have implemented a data_book$undo_last_action function?

I just note, while I reemember:

rdstern (approved):
@N-thony this is very nice indeed.  I am approving even though it is only a draft.  
@Patowhiz could you check and perhaps merge if ok
a)  There's just one text to delete.  When deleting rows it still sats it can't be undone.  That's no longer true.

I checked with filtered data, there even delting all.  I also tried reordering columns, and changing multiple names.  

I would still like to know what is behind the function, especially as you can do multiple undeletes.  Does it therefore make multiple copies of each data frame.  If so, then is there a size limit we could set, so it doesn't slow down visibly.  Might it also slow down - or get confused with multiple sheets.  It doesn't seem to?

 You are wondering about adding a menu item - to match Excel.  No problem with that, partly so the Help can tell users about it, and it can indicate the short-cut keys.  But may not be necessary.   I am not keen on adding it to the column right-click, because that's already long enough, and I am sure we should consider a redo as well?  But I would be happy to  add it to the row-right click menu.  That could include the short-cut keys .  I know it works on more than rows, but it does work there.  Abd that has quite a small number of items, and it has a Help button already.

rdstern:
@N-thony am I right that all the data frames, i.e. the whole data book is copied up whenever any sheet is changed?  Or is it just that sheet.  So we soon have 10 progressive data books?
Because I remain extremely keen on this feature, but we then can't permit this for very large data books.  In particular we don't automatically prodiuce the column metadata when there are 1000 variables or more.  I suggest 1000 variables might be a strict limit, but we might want a lower limit as the default, maybe with an offer for Undo to be enabled on larger dataframes (in the tools > Options dialog, for any particular session.  I suggest, if we add that feature, then it needs to be set every session where we want quite large datasets to offer undo.

I call this spreadsheet-style undo, and we already have R-stype undo via the Tools > Restore Backup dialog.

So I see this as great for teaching, and really useful as a practical tool.  But we don't want R-Instat to be slowed down too much by this feature.  Also I anticipate that users of the spreadsheet-style undo will rarely be using very large data books.

rdstern:
@N-thony that's even better, namely that only the chosen sheet is copied.  That means we can have a simple rule for allowing your undo to operate.  It works by default unless the length of the longest data frame is more than xxx rows, or wider than yyy variables.  This then gives us a chance to discuss the sizes we can handle in R-Instat easily.  
Remember "your" undo is for beginners, and we have the R undo (if needed) for those who are happy to learn a bit more.  Explaining about undo fits well into our training, with yours mentioned in the basic workshops and both methods discussed in the Further workshop!
@jkmusyoka you may have views, and @Patowhiz could be well placed to help specify the limits.  I'm trhinking that the user can decide to ignore the limits for a given session - but that maybe silly if a data frame bepomes very wide (Patrick?)  I don't mind it being quite silly - because then the user learns not to try the simple undo with such crazy data frames?  
I'm wondering if copying a long data frame is always quite quick?  It is only with wide daya frames that we have real time problems?  
If we always have a limit then my first go is that the default limit (for now) is <600,000 rows or perhaps 1 million rows (is this too high).  And < 500 variables.  And we can optionally take off the limit for rows (unlimited), and at the same time increase the variable limit to 1000 variables.  @Patowhiz you know all about speeds, when data frames are wide.  You also made a limit of 1000 variables, so any dataframe wider than that R-Instat asks if the metadata is to be calculated?  I suggest our undo limit is always the same as your metadata limit.  I did wonder whether that was a bit conservative?  Would you consider raising that to 2000 variables?

rdstern:
@N-thony now I am learning more!  Does that mean that earlier we didn't include the data frames in the history object, and now we do, so we can have our undo?  
Or it was always included, and now we are simply taking advantage and using it?

rdstern:
@Patowhiz I am still happy - when you are - to have this merged as it stands.  You may have other ideas.

There should then still be a further issue/pull request, to 

a) Consider redo, but not if it takes long.
b) Put it somewhere as well as having the short-cut key.  I would be happy if you decided that in the column right-click is sufficient.
c) Adding limits, so it is not for large data frames - whatever we decide is "large".

Here is a picture of R-Instat:

image

I am keen to show that undo doesn't just change the data, it can also change the meta-data at both column.  That's really nice for our teaching about the meta-data.

However, I assume changing the data and metadata will be a strain, particularly on very wide datasets, (just as it is for reading them in) .  I am keen for R-Instat to remain as efficient as possible for large datasets and don't want to always wonder whether some problems are caused by the undo facility.  I also see no problem, because people handling very large datasets shouldn't need the simple undo so much.  (We also have the R undo for them!)

ChrisMarsh82:
@N-thony This will need to be tested with multiple large datasets where multiple changes are made on each of the datasets and R-Instat is kept running for at least half a day. We need to make sure there are no memory issues with the changes.

rdstern:
@N-thony this is looking very promising:
image

a) Please can you make the up-down for the rows wider.  So it shows at lease 8 values.  
b) And I suggest it goes below the columns one.  With translation I am sure you might need more space.
c) And make the lower limit 100,000 for the rows and maybe even zero for the columns.  The would be a neat way to switch it off..  Certainly allow 100.  (Currently the lower limits are the defaults.)
d) I used a recent data frame of 550,000 rows and 11 columns.  So it didn't work with the defaults.  Previously you gave a message, but now it just didn't work.  Can you re-instate the message when ctrl Z is pressed?  It can direct them to the Tools dialog.
e) Just to check that the big data frame was actually only 15,000 rows visible because I had a filter in operation.  I assume it copies the ndata frame, and not just the filtered part? 
f) I changed the 200,000 rows to 600,000.  It took a long time to make the change.  Maybe worth checking why, or adding the standard message we give when operations take long?
g) We will have to put the facility elsewhere as well.  I hope it will be really easy to add it to the row right-click menu.  That can show the short-cut and it has a help attached, where we can explain what it does.


f) It does need to be somewhere

rdstern (approved):
@N-thony that's excellent.  I am approving.  I hope the addition in the row right click can come soon in a separate pull request.
@Patowhiz this is a really excellent addition.  Please check and hopefully then you can merge.

Patowhiz:
@rdstern @N-thony thanks for this. @ChrisMarsh82 will check this and merge.

ChrisMarsh82:
@N-thony  I really like what you have done but I want to see some testing from it. I need you to do some load testing. i.e. checking speed difference. 
Check if number dataframs makes a difference. i.e. 1 df, 10 df and 100 df - checking undo on all dataframes
Check multiple large dataframes - checking undo over 10 times
It should also show memory usage just checking it doesn't get too large
It should also have dialog actions that run scripts on the datasets, does this change the speed.
I know this is a bit of a pain but I think we need this testing report so that we can be confident this wont have adverse effects for users


It also needs an option to turn off completely, not just on large datasets. 

@rdstern Are we happy that if I delete a row, then run a bunch of R commands on the dataframe that I can still press undo and go back to a dataframe that could be very out of data?

rdstern:
@N-thony I think you have still missed what I interpet to be the main challenges that @ChrisMarsh82 has tried to set you.   And reflecting on that, will be useful for you much more generally than just this particular task.

As I see it, you have added a really great facility to R-Instat.  I am also really happy that you chose to add it - far too many of the additions to R-Instat seem to come from me. One reason is it so nice is that it targets our "spreadsheet" style users, and these are the extra people who can now use R even if they hate the idea of scripts.

But you stopped too early.  You should have recognised when an addition is risky.  This is both in your own work, and in your work supervising the team.  For example you made the cavalier statement that you could now do undo back 10 steps and you could make that unlimited.  And you didn't seem to show a concern that this might cause R-Instat to crash?  So, remember that this is a convenience for some R-Instat users.  We have got away without it for 10 years.  We shouldn't have anything crashing R-Instat, and particularly not a convenience.

So, in the future, as our senior programmer, I would expect you to have included those limits that are now in the Tools > Options.  It shouldn't have needed me to ask you to do it.  And the checks from Chris are also items you should have wanted to do.

So let me suggest some things you could now be doing - and note if I am making any sense here, then it shouldn't be me making these suggestions to you - it is all computing and so it should be you maybe checking your ideas with me - or with Chris.

a) Chris suggests it could go wrong when you have 100 reasonably large data frames.  I wonder how you will get them?  You could create a new data frame, then fill it say with random numbers, then duplicate it.  Then repeat another 98 times.    You would obviously do this in a little script - nice perhaps in our new script directory.  I am not yet good on loops in R.  You probably are, otherwise check with Vitalis and (failing that) maybe with James.  (Then remember, this was just one quick suffestion from Chris.  You may think of other things to try.  And there is no magic with just 100.)
,b) Then I see 2 main risks, one is that it may take too long, (under certain circumstances) and the second is that it uses too much memory and possibly never clears it.  Do you clear stuff out?  Anyway I have already pointed you towards timing checks - remember issue #6264 that we discussed a couple of days ago?
c) So how about checking on memory?  You are adding backup files and I have no idea whether you are doing good housekeeping on those files.  Do they ever get deleted?  So I checked - as usual - with stack-overflow of course and found this, called [Monitor Memory Usage in R](https://stackoverflow.com/questions/7856306/monitor-memory-usage-in-r).  That seems quite relevant to your task don't you think?

Now, in practical terms, how are you going the do this work - and I don't want you to spend too long, because you have a lot of other tasks, if we are going to get a good October release.

I suggest adding the timing and memory checks should be pretty quick, if you only do them in a script.  And those will be quite interesting scripts to add to our growing library.  You are doing the undo properly as an R command.  So pretty neat that you can do this checking in a script!  So don't bother adding more dialogs. at this stage.  But do comment the scripts nicely so we can use them in teaching later.

rdstern:
@N-thony very nice that you also added Undo to the Edit menu (with the short-cut) and it seems to work well there.

image

a) The up-downs could be still a bit wider still.
b) You say they reset each time - but they don't.  I set it to 1.6 million and it remained at that the next time.
c) The switch off undo is checked by default - but doesn't switch off.  Change the default to unchecked please.
d) In the whole dialog we have been inconsistent with most words starting capitalised.  Could you chege that to be consistent.  Maybe leave the resetting in brackets mainly in lower case?
e) I suggest 200,000 for the default rows may be ok?
f) You say you have added the times.  Great - but where.?  Can you guide on how it is used?

rdstern:
@N-thony you've done it the other way round now.  I wanted R-Instat to start with undo enabled.  So the message in Tools Options should be unticked, and say Switch Off.  When you tick, then it will switch off.
You are right, at this stage I am happy that you have added the timer simply to be used in the script window.
Have you looked at whether you can also add something to check memory, in the same way - also in the script window?

Then I am ready to ask @ChrisMarsh82 to check the code and merge.  His requested tests will still need to be done, but I see that as a useful task as we test the release more generally.  (If it shows a particular problem we can quickly hard wire the off switch to be the default, and disable it, for the released version.  I don't think it will!)

rdstern:
You chenged the message correctly to switch off.  But it is still off to start with.  

image


I want to be positive, so to start with Undo switched on.  That's now consistent with the dialog - so that doesn't need to change.  Just start with undo switched on.

image


By gthe way I noticed I ahd to swich off  (i.e. tck) to switch it on.  Then it enabled me to undo the changes I made before!!!  Would that be the case for every data set or just the small undoabole data frames?  Are you sure that's not dangerous?

rdstern (dismissed):
@N-thony great.  I am approving.  @ChrisMarsh82 over to you, and I'd really like this in the next version if possible.  My one concern is that I turned off the undo.  That was fine.  Then I turned it on again and I could now undo what I couldn't when it was turned off.
So I wonder how that happened and what actually is turned off?  Maybe we should chat.

rdstern:
@N-thony I'll check, but that's getting more complicated than I think is needed.  I see this undo as an immediate one.  So I would even be ok if the undos are cleared when you move data frames.  Maybe you could discuss with Chris, when you meet him tomorrow morning.

rdstern (approved):
@N-thony still looks fine to me.  @ChrisMarsh82 maybe back to you?

ChrisMarsh82 (approved):

PR #9218, opened 2024-10-29
updated packages

N-thony:
@rdstern @ChrisMarsh82 I have updated the packages to align with the current R version we are using.
I noticed after this, I have installed the packages again, and I did not need to ask a specific package as it has happened this morning on your side.


rdstern:
@N-thony I m not sure what I need to do here?  I tried running the branch in the usual way.  Then what?
I wondered if this could also relate to the Tools > Install R packages minor issue where it now always says Cannot get package information?  I don't think so?

rdstern (approved):
@N-thony I'll approve.  Maybe @lilyclements could be the second approver, so you could merge and we can all update?

lilyclements (approved):

PR #9228, opened 2024-11-04
Bug Fixes in Parallel Coordinate Plot

### LINKED ISSUE #9201: Parallel Coordinate Plot
Hello @N-thony, @derekagorhom and @rdstern,

When I try  to open Parcood Options

I get this message. 

Then when i close the message this has opened. 

This seems odd so I thought I would mention it. 

Thanks, 

Rachel

derekagorhom:
Fixes #9201 
I have fixed the issue with the `developer error` and also error with `column_numbers`
@rdstern @rachelkg , @N-thony 
Can you have a look 
Thanks

### DIFF
--- a/instat/dlgParallelCoordinatePlot.vb
+++ b/instat/dlgParallelCoordinatePlot.vb
@@ -173,11 +173,14 @@ Public Class dlgParallelCoordinatePlot
         ucrSaveGraph.Reset()
         bResetSubdialog = True
 
+        clsMatchFunction.SetRCommand("match")
+        clsMatchFunction.SetAssignTo("column_numbers")
+
         clsGGParCoordFunc.SetPackageName("GGally")
         clsGGParCoordFunc.SetRCommand("ggparcoord")
-        clsGGParCoordFunc.AddParameter("columns", "column_numbers", iPosition:=1)
+        clsGGParCoordFunc.AddParameter("columns", clsRFunctionParameter:=clsMatchFunction, iPosition:=1)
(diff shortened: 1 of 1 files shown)


rdstern (changes requested):
@derekagorhom well done.  That has fixed the problems Rachel found.  Unfortunately I have found a new one!  But is should be easy to fix.

Here it is:
image

Now, in the dialog, it says that Factor is optional, so I have left it out.  And that has triggered the error.  I show you the R code, because I think that points to the fix:

```
# Dialog: Parallel Coordinate Plot

iris <- data_book$get_data_frame(data_name="iris")
column_numbers <- match(c("Sepal.Length","Sepal.Width","Petal.Length","Petal.Width"), names(iris))
last_graph <- GGally::ggparcoord(data=iris, columns=column_numbers, missing="exclude", order="anyClass", centerObsID=1) + theme_grey()
data_book$add_object(data_name="iris", object_name="last_graph", object_type_label="graph", object_format="image", object=check_graph(graph_object=last_graph))
data_book$get_object_data(data_name="iris", object_name="last_graph", as_file=TRUE)
rm(list=c("last_graph", "iris", "column_numbers"))
```
@derekagorhom by the way, in the code I really like the second line, which enables us to use numbers, rather than names in the command.  I assume that's from @lilyclements ?

Anyway, notice the order = "anyclass", in the command, which is causing the problem.  If there is no factor, then just leave the order argument out and it works!

rdstern (approved):
@derekagorhom perfect, thanks
@N-thony over to you

PR #9230, opened 2024-11-04
Load the data in the grid directly after double-clicking .RDS file

N-thony:
@rdstern @ChrisMarsh82 this will load directly the data in the grid after double click on .RDS file. Have a look.
I noticed that each time we double-click an .RDS file, it opens in a new R-Instat instance. This might have been because we allow multiple R-Instat instances. But I think it would be better if, when we double-click, the data always loads into the first opened R-Instat instance if there are multiple open. What do you think?

### DIFF
--- a/instat/ApplicationEvents.vb
+++ b/instat/ApplicationEvents.vb
@@ -9,7 +9,8 @@
     Partial Friend Class MyApplication
         ''' <summary>   Handles the event raised when launching a single-instance application. 
         '''             If the event was triggered by double-clicking on a data file, then the sub 
-        '''             opens R-Instat (if not already running), and load the data in the Grid.</summary>
+        '''             opens R-Instat (if not already running), and load the data in the Grid
+        '''             with the double-clicked file's path.</summary>
         '''
         ''' <param name="sender">   Source of the event. </param>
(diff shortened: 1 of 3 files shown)


rdstern (commented):
@N-thony am I doing something wrong?  My double-click still gives the open dialog?

rdstern:
@N-thony I think we are at cross-purposes here!
I am just asuming I am already in an instance of R-Instat.  There I have:

image

It is here that, when I click on the .R file it opens immediately into the script window.
I would like the rds extension files from here (or from the bottom of the file menu) to open directly - maybe to add to an already open data book, when I double-click.  When I single click it opens the dialog as now.

rdstern:
@N-thony I am not happy associating rds files outside R-Instat with R-Instat.  I strongly suggest this is confusing for everyone who also uses R with RStudio.

I was just hoping, once in R-Instat, and returning to an rds file (usually from the front screen, in the recent files, as shown above,  that a double-click would open the file in R-Instat without going to the dialog.  But it is a minor convenience, so happy to drop the request if you wish.

rdstern:
I am closing this now and am explaining in detail in the issue

PR #9232, opened 2024-11-05
adding if statement to param list

### LINKED ISSUE #5482: Column summaries output display for no by factors
Originally from here #5465 

> With the more general Column Summaries with multiple variables and multiple summaries it gives an awful display in the output window - as it assumes one row and as many columns as there are variables Times summaries. If you save it has a special bug - @dannyparsons knows so only a single number is displayed.
> When there are no factors then could the (default) display be with a variable (column) for the summary and each variable is a row. This could perhaps be an option more generally? That's a stack, but difficult to specify currently manually until we can have groups of variables.

lilyclements:
Fixes  #5482

@rdstern this should fix the factor issue that Lisa reported. This is ready to test/review

### DIFF
--- a/instat/static/InstatObject/R/Backend_Components/summary_functions.R
+++ b/instat/static/InstatObject/R/Backend_Components/summary_functions.R
@@ -239,7 +239,7 @@ DataBook$set("public", "calculate_summary", function(data_name, columns_to_summa
     curr_filter_name <- curr_filter[["name"]]
     curr_filter_calc <- self$get_filter_as_instat_calculation(data_name, curr_filter_name)
     manipulations <- c(curr_filter_calc, manipulations)
-  }
+  } 
   if(!missing(additional_filter)) {
     manipulations <- c(additional_filter, manipulations)
   }
@@ -247,10 +247,12 @@ DataBook$set("public", "calculate_summary", function(data_name, columns_to_summa
 
   # setting up param_list. Here we read in .drop and .preserve
   param_list <- list()
-  for (i in 1:length(combined_calc_sum$manipulations)){
-    if (combined_calc_sum$manipulations[[i]]$type %in% c("by", "filter")){
-        param_list <- c(param_list, combined_calc_sum$manipulations[[i]]$param_list)
-    }
+  if (length(combined_calc_sum$manipulations) > 0){
+      for (i in 1:length(combined_calc_sum$manipulations)){
+          if (combined_calc_sum$manipulations[[i]]$type %in% c("by", "filter")){
+              param_list <- c(param_list, combined_calc_sum$manipulations[[i]]$param_list)
+          }
+      }
   }
   out <- self$apply_instat_calculation(combined_calc_sum, param_list = param_list)
   # relocate so that the factors are first still for consistency	


rdstern (commented):
@lilyclements many thanks for trying to fix this.  Both Lisa and @dannyparsons said it used to work.  
With your fix I don't get the same error, so it clearly helps.  However:
a) I suggest it only gives a single value in the new summary sheet.
b) If you ask for (say) mean, min, max, count, it still just gives the first value.
c) It works when you put the results into the output window.

With your fix (I had missing values in my data), I just got a single NA in the results window.  When I asked for the missing in the data to be ignored it made no difference.
So I then filtered out the missing values and it now gives an error.

So, I think there is either a bit more to do, or we (at least temporarily) just try to return to an adaptation of what worked previously.  I need to check more and I'll then continue.

rdstern:
@N-thony if @lilyclements agrees with my suggestion below, then I suggest you could quickly do this.  It is "front-end" so I think much easier for you to do than Lily.
When there are no factors, I suggest that automatically the results only go to the Output window and not to a new data frame, or to add anything to the calling data frame.  

Here is an example of the dialog:  Note nothing in the Factors control on the right:

image

That is the situation where Lily has fixed a bug.   I am suggesting that what you do is the same as you could achieve by unticking the Store Results and Ticking the Store results in Output window controlm i.e. if the dialog looked like this:

image

I am not suggesting you change these controls on the dialog, but that's what happens when Ok is pressed, if there are no factors.

lilyclements:
@rdstern is your (a), (b), (c) an issue you are currently getting, or something you are proposing (I assume the latter). In which case, this isn't R code bits and is front end stuff as you say. 

I am not sure about removing the option to have it stored as a data frame if there are no factors. If I have all the R code up on the right hand side, or output, it can be difficult to flick through it and find that value I calculated before. Instead, I quite like being able to find it in a data frame. However, I am not our expected user, and if it is more confusing to have it show up this way then I understand.

rdstern:
@lilyclements my a) , b) c) is what currently happens and also was the case in Version 0.7.6, where it was claimed to work.
I am also quite keen to propose that the solution of outputting, from this dialog only to the Output window, when there are no factors, is now (for us) a pretty good practice to recommend.  That's because of the following:
a) If you want a one-line summary in its own new data frame, then it is quite easy to do.  Just make a factor column with a single level and use that in the Summary dialog.  The factor (with just one level) is not just for the heck of it.  It is also the linking variable.  So further summaries add to the same data frame.
b) If we can fix the one line situation without a factor, then it won't be linked.  So not quite so good.
c) With just one or two summaries you probably want to look at them anyway - nice in the output window.  
d) Now the final (new) important point - the sort of good practice stuff.  In our "calculation system" we can produce 3 "levels" of summaries, or transformations.  We can produce scalars, or columns or whole data frames.

We have recently added scalars to our datafrmae metadata.  So I suggest that, if Lisa wants the overall mean to be able to use in further calculations then it is a scalar result from the calculator.  So, instead of using the Summarise dialog and producing a whole dataframe for a single number, she just adds a scalar to the yearly dataframe as below.  Then it is easily available for further calculations.

image


Nice, now?

lilyclements:
@rdstern great. Are you happy with the code changes in here for someone else to take on the front-end changes?

rdstern:
@N-thony I wonder if you could do this quickly?  It would then be a fix for Lisa's bug.
So from the code, as now, 
If ok is pressed and there are no factors:
a) The output is is to the output window - so as though the `Print Results to output window` was ticked.  (Whether or not it was ticked!)
b) The results are not stored in another data frame - so as though the Store Results was unticked.  (Whether, or not it was unticked!)

I hope that will be easy?

rdstern (changes requested):
@N-thony it is almost there - working too well really
a) When no factor it seems fine.  Great.
b) But when I add a factor, then it still goes to the output window.  Ehen there is a factor, then it should obay the dialog.

image

rdstern (changes requested):
@N-thony that's better, but now it is in both the new data frame (good) and still the output window (bad).  

image

rdstern:
@N-thony and that's just in your branch I think.  It should not be checked by default.  It isn't in Version 0.8,0, but it is in your branch.  So please change that.

That's a relief that you found it!

rdstern (approved):
@N-thony perfect.  I'd really like to get this merged in the forthcoming release.
@Patowhiz could you please check?

rdstern:
@Patowhiz please could you check now, and merge if ok.  This is an important change and I thought it was in 0.8.0.  I'd really like it in the version this week.

Thanks

Patowhiz:
Hi @rdstern, I believe this was merged about two weeks ago. However, it seems the changes might have been reverted or undone. Confirm with @N-thony?

rdstern:
@N-thony and @Patowhiz my mistake, sorry.  It was merged 2 weeks ago as Patrick says.

PR #9234, opened 2024-11-06
Issue when checking all packages installed

N-thony:
@ChrisMarsh82 @rdstern I realized after a discussion on Skype with @Patowhiz that `AllPackagesInstalled()` is called immediately after starting the R engine but before executing setup scripts via `ExecuteSetupRScriptsAndSetupRLinkAndDatabook()`. This ordering causes issues because `AllPackagesInstalled()` depends on R scripts or functions loaded later in the process.

To resolve this, I rearrange the order so that `ExecuteSetupRScriptsAndSetupRLinkAndDatabook()` is executed immediately after starting the R engine, allowing it to configure the necessary R environment before checking for installed packages. Have a look.
Note that this is only for developers to be in track also of installed packages in the development. It is not a big issue because it doesn't affect users. We can leave it open and use it if needed.

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -179,11 +179,8 @@ Public Class frmMain
         '---------------------------------------
 
         '---------------------------------------
-        'start the R engine and install any missing R packages
-        'only proceed if;
-        '1. R engine has been started
-        '2. All packages have been installed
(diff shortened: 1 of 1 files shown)


lilyclements:
Closing as we have an updated approach to handle this

PR #9237, opened 2024-11-06
Fixes paste from clipboard bug

### LINKED ISSUE #9207: Paste no longer works in the test version
@N-thony so-far-so-good on the test version.
I have been writing the help on your undo and it is working well.

Now I have found the first bug and that is that the paste operation no longer seems to work.

I tried in diamonds, and then survey, just copying 3 cells from a single column and pasting them lower down into the same column.  By the way the warning may have to change, because (when it works) I assume and hope, when it works that it can now be undone> 

image
Anyway now I get this error:

image

So there is a problem now in the data_book$paste_from_clipboard function to be fixed!

I know this gives you new work, but in a way I'm quite pleased - this is what our testing is all about!!!

Patowhiz:
Fixes #9207
@N-thony @rdstern @jkmusyoka this is ready for your review.

I've also provided detailed information on how this bug originated in issue #9207. Would be good to read the comment because it may relate to other future potential problems.


rdstern (approved):
@Patowhiz and @N-thony I think it is now working fine, so I am approving.  But I note as follows:
a) pasting within a dataframe gave an error, saying clipr was not installed.  I assume that's just my machine, and it will be installed in the proper version?  Anyway I installed clipr and it now seems to work fine.
b) The paste new now works - at least starting from Excel.  I have sent a message and data in word, as it didn't work for me directly from word (It did paste into Excel and then from Excel to R-Instat.

Patowhiz:
@rdstern can you confirm if it pastes from word? I sent a response that you may need to download the file and copy from MS word not from the browser preview. It worked on my end.

rdstern:
@Patowhiz I did download the file and then opened word on the desktop.  It still didn't paste - though it seems fine from Excel.  I looked for a reason from stackoverflow, but couldn't resolve that.

Patowhiz:
@rdstern I tested using the ZMD participants file that you sent. The limitation is in `clipr` package due to the way the contents are copied from word to the clipboard. When you open the word document, notice the spacing in the first participant. that's what is making `clipr` to fail. Pasting to excel or even notepad first then to R-Instat works because that first line after header is corrected.

**Copying from Word to R-Instat**
word

**Copying from Excel or Notepad to R-Instat**
excel

**Copying directly from clipboard using clipr** - This confirms it's a `clipr` issue.
pasting

PR #9244, opened 2024-11-12
Minor correction in the Prepare > Data Reshape > Column Summary Dialogue

N-thony:
@rdstern take a look if this helps with the output

### DIFF
--- a/instat/dlgColumnStats.vb
+++ b/instat/dlgColumnStats.vb
@@ -230,9 +230,13 @@ Public Class dlgColumnStats
                 clsDefaultFunction.RemoveParameterByName("return_output")
                 If ucrChkStoreResults.Checked Then
                     clsDefaultFunction.AddParameter("store_results", "TRUE", iPosition:=3)
+                Else
+                    clsDefaultFunction.AddParameter("store_results", "FALSE", iPosition:=3)
                 End If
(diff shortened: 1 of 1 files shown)


rdstern (commented):
@N-thony excellent.  I think that's fixed the error.  That's a relief.  

But this worked fine in Version 0.7.6.  Can you re-instate easily?

image

If you write back to the original level it doesn't matter whether there are factors in the By or not?

rdstern (approved):
@N-thony perfect.  Great, thanks.
@Patowhiz please can you check and merge when ok.  This is the fix of one of Lisa's problems

Patowhiz:
@rdstern @N-thony I'm not sure what needs to be checked here. Is there an existing issue that explains what this PR aims to address?

rdstern:
@Patowhiz I agree you might find it hard to see what to check.  It was an important, but minor correction to @N-thony big correction that was merged last week.  The whole change is important, and the correction here is just an important bit of tidying.

Patowhiz (approved):
@rdstern @N-thony I still don't understand the context of the change and I'm not sure on the significance of the change. 

I'm approving this because the changes are scoped to the dialog only and therefore low risk.

Will merge once @rdstern approves.

Issue #9088, opened 2024-07-30
Three small points from the latest exercise


rdstern:
@N-thony I have been preparing a new exercise to interoduce people to R-Instat, who are too strong to need the tutorials.  I have (maybe for the first time) fould nothing major that needs correction.  Two of the minor issues are shown here:

![image](https://github.com/user-attachments/assets/83162d55-d8b9-470f-9aae-3989a906a529)

a) The scalars are working fine.  But they don't obay the formatting rules in the metadata.  Could this be added?  I assume that may be for you?
b) I have not checked the backing up.  But whenever the data are backed up, then at the bottom it says "No data loaded", which is clearly untrue!  I assume that is probably a @derekagorhom task to change the message, perhaps to "Backup Completed".  Currently that is there until you change the data frame, and I think that's fine.
c) I have, for a long time, been concerned that in the graphics the first time you go to the sub-dialog it takes a long time - at least 5 seconds on my machine.  That's only the first time you do this.  After that it is almost instananeous.  Could that please be examined?  I assume that will have to be by you or perhaps @Patowhiz?

rdstern:
@N-thony and @derekagorhom could you please see if you can fix point b) above?

Issue #9236, opened 2024-11-06
Why do we repeat our backing up in R-Instat if nothing has changed in the log file?


rdstern:
@ChrisMarsh82 and @N-thony and @Patowhiz I am writing the help and have got as far as the log file.  Here is a bit of my latest:

![image](https://github.com/user-attachments/assets/a6954c1f-394f-403a-89ea-1c7daaa13d7f)

I often have R-Instat open, while I am doing other things.  I notice, from the figure above that it backs up every 10 minutes, even if I have not used it, since the last one?  Couldn't it check that, and only backup if something has changed?  I can imagine situations (with large datasets) where this could be very annoying?

rdstern:
@Patowhiz happy with your support of my first point.  The times of backing up have been included recently - and I'm afraid at my request!  That's by @N-thony and after discussions with Stephen.  They form part of our recent "ultimate undo system" for those who can cope - and I hope you'll like it.
That's also why I was so pleased to have Antoine's new "spreadsheet style" undo system in the latest versions.  That's in contrast to the R-user type undo, which uses the information in the log file!  
Remember we back-up the data every 10 minutes (by default).  Now suppose we have a problem and need to go back a bit.
a) Use Tools > Backup to restore the data up to the last backup.
b) Use the log file, from the date/time of the last backup, as much as you need to complete your data rescue.

Neat eh, for those who can cope?

rdstern:
@N-thony after my experience in the Bangladesh workshop I would like to take the backing-up a stage further.
a) As above, it should only backup the data book, if something has changed.
b) Could there be a record kept, perhaps in the data frame metadata, of the last time each data frame is changed.  Then, after the initial backup, could a data sheet only be backed up differently, if that sheet has changed?

This is particularly important for our climatic data.  The main data file is usually there.  It is often used to get summaries, etc, without being changed.  And soon we will be dealing with larger files from sub-dialy data.  Then often the daily data will be generated.  It will still be useful to keep the primary data available, but it will rarely be changed and hence will need backing up anew.

Patowhiz:
@rdstern,

What has been the current backup experience with all the data frames (hourly, daily, summaries, etc.) open? I recall hearing that you were working with sub-daily data containing 7 million records, which I believe will become increasingly common in the next 5–10 years.

I'm also curious about how this operation performed on users' machines. Did they notice any performance degradation during the process?

[...]

PR #9246, opened 2024-11-12
Set No loaded data text correctly after auto saving

N-thony:
fixes partially #9088

@rdstern have a look

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -1169,12 +1169,7 @@ Public Class frmMain
             clsSaveRDS.AddParameter("file", Chr(34) & strCurrentAutoSaveDataFilePath.Replace("\", "/") & Chr(34))
             clsRLink.RunInternalScript(clsSaveRDS.ToScript(), bSilent:=True, bShowWaitDialogOverride:=False)
 
-            If strCurrentStatus.Equals("No data loaded") Then
-                tstatus.Text = GetTranslation(strCurrentStatus)
-            Else
-                tstatus.Text = strCurrentStatus
-            End If
-
+            tstatus.Text = strCurrentStatus
             Cursor = Cursors.Default
             bFirstBackupDone = True
         End If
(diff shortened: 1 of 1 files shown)


rdstern (dismissed):
@N-thony I am not sure, and you don't say what should happen.  For me it now seems to just remain with the name of the current sheet.  There is no extra message.  I am quite happy with that, so am approving.  
@Patowhiz can you check and merge?

rdstern:
@N-thony thanks, you have explained it perfectly.  @Patowhiz previously, after backup it said no data loaded.  That is item b in #9088.  Now it is fine, and I am happy we are at last getting these details sorted out.

Patowhiz:
@rdstern @N-thony thanks for the explanations. 

Have you both tested this using a different language to just make sure everything shown is as expected? You both speak french, would be good to us it.

I mention this because the code changes seem to suggest it was a translation issue.

Patowhiz (changes requested):
@N-thony thanks for the changes. I think all you need here is to do the following:

1. Retain changes made in line 1153.
2. Retain the deletion you have made in line 1175.
3. Revert other areas to the way they were before the new additions.

Reason being, line 1175 is the one that sets the `tstatus.Text` to "No data loaded" because it uses translations from the database. Note the "No data loaded" is set in `ucrDataView` in line 298 which already translates it, so when translations like French are on, your new addition in line 1172 will never be called.

rdstern:
Doesn't that point also relate to my #9236 ?  Remember the autosaving is also saving the updated log file.  I wonder whether you could have time (maybe with @ChrisMarsh82 ) to only auto-save when there has been a change either the log, or the data, or both.
But having no data (if previously there was data) is to be "saved".  And it seems to do no harm.

Patowhiz:
@N-thony, my three points above were meant to outline the translation process and your response. Point 1 refers to translating "Auto saving data," and Point 2 involves removing the form translation handled within the subroutine. I hope this clarifies things a bit more.

rdstern (approved):

Issue #9241, opened 2024-11-09
Small changes in our summary dialogs


rdstern:
@N-thony this should be easy for anyone:

a) In the Prepare > Reshape Data > Summarise > Subdialog we now only tick N-non Missing and Sum as the default.  That seems sensible.  Nothing to change.  In the Describe > Tables > Summarise the default is the mean.  I like that too.  Nothing to change.
b) But in the Climatic > Prepare > Summarise we still (by default) tick the N-Total as well.  Please not.  Let's make it consistent with the Prepare > Reshape Data > Summarise.
c) A more important one now.  The name (in each case) of the summary is `count_non-missing_tmax`, etc.  This is too long and a pain both in a data frame, and now in the tables.  And we use it often.  The N-total, that we hardly use, is called count, which is much better.
I suggest we call the count_non_misssing `count` instead.  Than perhaps call the N-total `count_all`.  Or we could be even shorter and have n and n_all.  @lilyclements and @N-thony what do you think?
Let's also change count_missing to count_miss, or n_miss?

I like the percentiles.  They are called p75_tmax, etc.

lilyclements:
@rdstern happy for us to rename these variables. I agree they are quite long as it is!

rdstern:
@lilyclements would you prefer `count` and `count_all` and count_miss, or n and n_all and n_miss?  Or something else?

lilyclements:
@rdstern I think `count` is good - especially since in R they're now using `count()` as a function instead of `summarise(n())` (although the column in R is called `n` by default - so I'm not against `n` at all).
What do you think?

rdstern:
@derekagorhom can you change the current count_non_missing_name into `count_name`.
Then the current count (which comed from N-total into count_all_name
And count_missing_name, into count_miss_name.

Then @lilyclements I just found a bug, from dplyr, in N-distinct - it gives an error.  Could you help, so @derekagorhom corrects that at the same time.

PR #9248, opened 2024-11-12
Getting n_distinct working for summaries

lilyclements:
@rdstern the smallest change I think I've ever done in a PR (which is very pleasant!). This is ready to review.

As mentioned in issue #9241, there was an error in `n_distinct`. This should fix that and means it now runs fine.


rdstern (approved):
@lilyclements nice, and thanks and I'm getting encourages we are now fixing these rare problems.
@N-thony please can you check and merge.  This will not be your most difficult task this week!

PR #9170, opened 2024-10-02
Changes to the Use table dialog

derekagorhom:
Fixes partly #9142 
I have done parts a, b, and c of this issue
@rdstern can you review this 
Thanks

PR #9249, opened 2024-11-12
Adds viewing and exporting of saved tables in use table dialog

### LINKED ISSUE #9142: Sort out the Use Table dialog
@Patowhiz here is the current dialog

image

I suggest:

a) Move the Format Table button to under the data selector. Call it Table Options and disable it for now.
b) Hopefully enable the options to produce some of these options and decide what to do with the results.  Could there be an option to copy to the clipboard.  Apart from that I assume they could be saved to a file.  Does that include html nor is that simply put in the output window?

Patowhiz:
Fixes #9142 
Replaces #9170

@rdstern @N-thony this is ready for review.

@rdstern @lilyclements saving of **pdf**, **png** and **word** fails in R-Instat but works in RStudio. You can test this by running the log file produced by R-Instat in RStudio.

Exporting `gt` to word in R-Instat produces error `Pandoc document conversion failed with error 127`, I tried looking into but I'm not sure how this could be solved from R side. Same saving operation in RStudio worked seamlessly.

Exporting `gt` to pdf and png (after applying the table options) produces error `execLater(f, delay, loop$id): Failed to get callback timer` in R-Instat. RStudio only required me to install certain packages as shown in the screenshot below.
webshot2

@N-thony when I used RStudio to download the webshot R package, the other dependencies were also automatically added.   I hope the same would have applied to R-Instat or will users be required to install the other R packages in the screenshot below.
extras

@lilyclements @rdstern I couldn't find solutions that are readily applicable to R-Instat. Some seem to suggest that we have to install dependencies that are out of R because these dependencies come with RStudio by default. Should we disable these export 'types' for now while we explore viable solutions or we leave them and wait for @lilyclements suggestions?


Patowhiz:
I change the dialog design slightly and the issues explained in the above comment are related to R.Net, R and the relevant R packages.

PR #9275, opened 2024-11-25
Enhance Define Climatic Dialogue Autofill

### LINKED ISSUE #9272: Minor improvements needed on the Define Climatic Dialog
Heris a snapshot of the data, together with the define dialog:

image

There are 3 issues that arose with the data from the recent Bangladesh workshop.

a) Note the names for the Max and Min temperatures and wind speed were not recognised.  **_Do not fix that._**  The problem is different.  Please check the tab orders generally.  After entering Min into Minimum Temperatures the next field should be Maximum, but it isn't.
b) When I return to the dialog (after defining the data as climatic) it seems to "remember.  But other times - perhaps after closing and the re-opening the data book, it doesn't - as shown above.  Could you check the code, so when the data are defined as climatic, they are always filled in, on the dialog.
I hope these can be done quickly, so we can merge.
c) I am less sure here, but these data included dry bulb, which is one of many possible additional elements.  I am happy with the ones on the list, but would also like a facility to define more - and there are so many!  I suggest we omit Cloud Cover (to make space) and add a More button.  If checked it opens a simple sub-dialog with the data selector and a multiple receiver.  It has the label More Elements:  They are just entered here.  Then press Return.  They are then called more1, more2, etc in the climatic type.

N-thony:
Fixes #9272
@rdstern have a look? This fixes items a) b)

### DIFF
--- a/instat/static/InstatObject/R/instat_object_R6.R
+++ b/instat/static/InstatObject/R/instat_object_R6.R
@@ -2990,6 +2990,10 @@ DataBook$set("public", "import_from_cds", function(user, dataset, elements, star
   if (is_win) close(pb)
 })
 
+DataBook$set("public", "get_column_climatic_type", function(data_name, col_name, attr_name){
+  self$get_data_objects(data_name)$get_column_climatic_type(col_name = col_name, attr_name =attr_name)
+})
+
 DataBook$set("public", "add_flag_fields", function(data_name, col_names, key_column_names) {
   if (!self$has_key(data_name)) {
     self$add_key(data_name, key_column_names)
--- a/instat/DlgDefineClimaticData.vb
+++ b/instat/DlgDefineClimaticData.vb
@@ -55,18 +55,18 @@ Public Class DlgDefineClimaticData
         Dim kvpLongitude As KeyValuePair(Of String, List(Of String)) = New KeyValuePair(Of String, List(Of String))("lon", {"lon", "lont", "longitude"}.ToList())
         Dim kvpLatitude As KeyValuePair(Of String, List(Of String)) = New KeyValuePair(Of String, List(Of String))("lat", {"lat", "latitude"}.ToList())
         Dim kvpCloudCover As KeyValuePair(Of String, List(Of String)) = New KeyValuePair(Of String, List(Of String))("cloud_cover", {"cloud", "cldtot"}.ToList())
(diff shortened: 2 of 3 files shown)


rdstern:
@N-thony you have missed the main point in item a) in #9272 .  I have now put a bit there in bold!  Now you have fixed to recognise max and min, you can leave that.  But the point is that occasionally the dialog will NOT recognise certain elements, and you then have to fill yourself what they are.  That's fine.

What is going wrong is that when you return to that fialog in another session, having opened that same file, then it should remember the entries you typed manually, because they are still the climatic elements - and recognised in the metadata.  But it doesn't.

You could try, with dodoma is you rename tmin to say nnn and tmax as xxx.  Then use the dialog.  Then save the file and close R-Instat.  Then return and open these data again.

In the data from Bangladesh it now recognises max and min, but not wind - which is the wind speed.  I am happy to fill in manually, but it then needs to fill in for me, when I return to the dialog in a later session.

I hope that's clear now?

rdstern (approved):
@N-thony nice.  You said you had now understood the problem fully, and that's clear.
@Patowhiz over to you to check, and hopefully merge

instat/static/InstatObject/R/instat_object_R6.R
```diff
@@ -2990,6 +2990,10 @@ DataBook$set("public", "import_from_cds", function(user, dataset, elements, star
   if (is_win) close(pb)
 })
 
+DataBook$set("public", "get_column_climatic_type", function(data_name, col_name, attr_name){
+  self$get_data_objects(data_name)$get_column_climatic_type(col_name = col_name, attr_name =attr_name)
+})
+
```
Patowhiz, line 2996: @N-thony, this looks good on the VB side. I’ll proceed with merging it after @lilyclements provides feedback.

@lilyclements, @N-thony has added a new R function, `get_column_climatic_type`, in the data book. Do you think this was necessary, or could an existing function have been reused? Since you recently worked on the data book and climate-related functions, your input would be valuable here.

When I was making significant changes to the R data book previously, I noticed a number of R functions performing similar tasks where a single function with a parameter could suffice (what I refer to as "code bloats"). I worked to remove redundant functions during my updates to the output objects functions. Your thoughts on this would be appreciated.

Patowhiz (approved):
@N-thony as agreed by @lilyclements . I've now approved this.

PR #9276, opened 2024-11-25
Fixed a  minor bug in Line plot dialog 

### LINKED ISSUE #9273: Line Plot dialogue is not producing multiple plots for multiple variables
**Describe the bug**
Line Plot dialogue is not producing multiple plots for multiple variables, but a single plot

**To Reproduce**
 Steps to reproduce the behaviour (could use  any data set). For example
1. Open the ggplot2 data frame in "economics"
2. Click on Describe > Graphs > Line Plot..
3. Select "Multiple variables" and add 2/3 variables to the receiver for y and "year" for x
image
the output obtained is the following 
image

**Expected behaviour**
We should get  a line per variable (graph obtained with the same dialogue in R-Instat 0.7.6
image

Vitalis95:
Fixes #9273 
@fagiothree @rdstern @N-thony , it now produces multiple plots after removing  `colour = "blue" in geom_line`

### DIFF
--- a/instat/dlgLinePlot.vb
+++ b/instat/dlgLinePlot.vb
@@ -536,7 +536,6 @@ Public Class dlgLinePlot
 
         clsGeomLineFunction.SetPackageName("ggplot2")
         clsGeomLineFunction.SetRCommand("geom_line")
-        clsGeomLineFunction.AddParameter("colour", Chr(34) & "blue" & Chr(34))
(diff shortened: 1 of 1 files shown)


rdstern (approved):
Thanks @Vitalis95 looks fine now.  @N-thony can you check and hopefully merge.

PR #9279, opened 2024-11-26
Corrected the Autofilling in the Fill Date Gaps Dialog

### LINKED ISSUE #9271: Correct the autofilling in the Fill Date Gaps dialog
@N-thony this should be very quick, so could be you.  If not then I suggest @MeSophie .

[t1_100.zip](https://github.com/user-attachments/files/17892412/t1_100.zip)

The problem is minor and I don't understand why it occurs.  Here is the dialog for the data attached:

image

The `Station.` variable does not fill automatically.  It cBut, with other datasets the Station variable is filled automatically, so why not here.  I don't think it is the dot at the end of the station name, because it still doesn't fill when that is removed.

When the code is corrected I would welcome an explanation of why it doesn't work now?

MeSophie:
Fixes #9271
@rdstern I Fixed the auto-filling issue with the station receiver but unfortunately i can say what was the cause of this issue. I tried to create a date column to check if the error was linked to the metadata. Unfortunately I couldn't go through with it because when I went back to the Fill Gaps dialog it crashed my machine. Maybe because of this notification that appears and disappears as it pleases or because the data isn't daily? In short, I haven't been able to determine the cause of the error.  

image

### DIFF
--- a/instat/dlgInfill.vb
+++ b/instat/dlgInfill.vb
@@ -42,7 +42,6 @@ Public Class dlgInfill
         SetHelpOptions()
         bReset = False
         autoTranslate(Me)
-        AutoFillStation()
     End Sub
 
     Private Sub InitialiseDialog()


rdstern:
@MeSophie I think you may have fixed the problem for the daily data.  I tried with t2 data and I now get the station.

But please can you look again at the code, to start the dialog when you use the t1 data.  As you say, the fill date gaps won't work, with that dialog, because the data are not daily.  And also (unless you remove them) there are missing values in the date variable.  But the dialog should still fill easily.  People do make mistakes so they may try this dialog with the t1 data.  (And we can make it work later I hope.)  
On my machine it didn't break out, but clearly got into a long (perhaps endless) loop filling the data selector repeatedly.  Eventually it gave up.  Ask @N-thony if you are not able to correct the loading code for the dialog.

MeSophie:
@rdstern Could you please explain again the changes to be made in this dialogue? From what I understand, the station receiver fills up infinitely when t_1 is used?

rdstern (approved):
@MeSophie great, that change seems to have worked.
@N-thony over to you

PR #9283, opened 2024-11-28
Auto-Saving Data: Triggering Saves Only on Data Changes

### LINKED ISSUE #9236: Why do we repeat our backing up in R-Instat if nothing has changed in the log file?

N-thony:
Fixes #9236
@rdstern I hope this fixes what you have suggested in issue #9236. Have a look.

### DIFF
--- a/instat/dlgBarAndPieChart.vb
+++ b/instat/dlgBarAndPieChart.vb
@@ -135,7 +135,6 @@ Public Class dlgBarAndPieChart
         Dim dctLegendPosition As New Dictionary(Of String, String)
 
         ucrBase.clsRsyntax.bExcludeAssignedFunctionOutput = False
-        ucrBase.bUpdateGrids = False
         ucrBase.clsRsyntax.iCallType = 3
         ucrBase.iHelpTopicID = 438
 
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -184,12 +184,10 @@ Public Class dlgCalculator
             ucrBase.clsRsyntax.AddToAfterCodes(clsRemoveLabelsFunction, 3)
             ucrBase.clsRsyntax.bExcludeAssignedFunctionOutput = True
             ucrBase.clsRsyntax.iCallType = 0
-            ucrBase.bUpdateGrids = True
         Else
             ucrBase.clsRsyntax.RemoveFromAfterCodes(clsRemoveLabelsFunction)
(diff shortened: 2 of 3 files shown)


rdstern (changes requested):
@N-thony delighted to see the change.  But this backing up information - if true shows a problem at the start, with many macking up in the first 2 minutes!
[Log for backing up.zip](https://github.com/user-attachments/files/17947129/Log.for.backing.up.zip)

rdstern:
@N-thony it is odd now in a different way.  It says backup at the start, (at the bottom) when I think it is checking about undo instead.  And now you have taken the notification in the log file away it seems?  At least it never eregistered any more, and I was deliberately making changes?

rdstern:
@N-thony I am getting nervous now, on what we are trying to achieve here.  And I'm worried we might be regressing as we make a change.  Here is the log file I just loaded, which looks really odd still:

image

I suggest was abandon this for now and return in January where we try for the full change, namely to only backup a data frame, when something has changed there.

I don't want to get into details of which commands change the "data", because almost all the commands/dialogs have an option to change the metadata.

And I start to worry about how the backing up is supposed to work when we are using scripts as well as dialogs.

rdstern:
In Version 0.8.1 I loaded R-Instat in the afternoon.  I had a break in the evening, and have come back the next day.  I set the backing up to be every 15 minutes.  Here is a bit of the log file.
image

This seems to show that the backup is being done whenever the computer is being used, whether using R-Instat or not.  This could be both preplexing and annoying with large data sets.

The message also seems to be wrong, namely that it is backing up the log file as well as the data file?  I thought from @N-thony that the log file is backed up whevenever it changes?

I assume it should be easy to only backup when something has changed in R-Instat?  Or perhaps only when the log file has changed - and not simply the backing up message?

This would seem to be urgent.  The luxury of just backing up those data sheets that have changed can come later.  When that is done, please remember that changing the metadata objects (that are attached to each sheet) is a change.  So the trigger to change a sheet, could be anything that uses, or opens that sheet.  Not just those commands that make an obvious change.

PR #9284, opened 2024-11-28
Improving barchart and histogram dislogs

### LINKED ISSUE #9210: Small changes in barchart and histogram dialogs
@N-thony these were spotted by @rachelkg when writing the help.

a) Here is the current histogram dialog:

image

Please change the up-down into a control where the user types a number.  I am happy with the default of 1.5.  (The similar control with a number is already in the boxplot dialog - the Cut Width control there.)

b) Here is the bar chart dialog:

image

Usually the x-variable is a factor.  But it can be a numeric variable.  (With the multiple option consider the type of the first variable.)

a) Please add the same Bin Width control as for the histogram.  It could go just below the Plot Options.  As with the boxplot dialog it is hidden, unless the x-variable is numeric (and it is only for the frequency option on the dialog.)  When numeric it becomes visible and the default is unchecked.  
b) The R code currently is as follows:

`last_graph <- ggplot2::ggplot(diamonds, mapping=ggplot2::aes(x=carat, fill="")) + ggplot2::geom_bar(stat="count") + theme_grey()`

That's fine when the x is a factor.  But when it is a variate, as with carat in the example above, then please change `stat="count"` to `stat="bin"`

If the bin width checkbox is checked, then use the value given there, by adding the binwidth = value to the R code.  Here is an example:

`last_graph <- ggplot2::ggplot(diamonds, mapping=ggplot2::aes(x=carat, fill="")) + ggplot2::geom_bar(stat="bin",binwidth=0.1) + theme_grey()`

Vitalis95:
Fixes #9210 
@rdstern @N-thony , added `Bin Width` control

### DIFF
--- a/instat/dlgBarAndPieChart.Designer.vb
+++ b/instat/dlgBarAndPieChart.Designer.vb
@@ -115,6 +115,8 @@ Partial Class dlgBarAndPieChart
         Me.lblFacetBy = New System.Windows.Forms.Label()
         Me.ucrInputLegendPosition = New instat.ucrInputComboBox()
         Me.ucrChkLegend = New instat.ucrCheck()
+        Me.ucrInputWidth = New instat.ucrInputTextBox()
+        Me.ucrChkBinWidth = New instat.ucrCheck()
         Me.contextMenuStripOptions.SuspendLayout()
         Me.SuspendLayout()
         '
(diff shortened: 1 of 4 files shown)


rdstern (changes requested):
@Vitalis95 that seems fine except the Single Variable control now seems misplaced?

rdstern:
@Vitalis95 The Single Variable (top right) is still misplaced here.  This is the only one I have checked.

image

PR #9290, opened 2024-12-02
Improved the skim option in the 2/3 Variable summarise dialog

### LINKED ISSUE #9252: Improve the skim option in the 2/3 Variable summarise dialog
@Vitalis95 this is the same dialog as you are trying to finalise.  I hope the suggested corrections will be simple to add, and you might include them as you finish the 3-variable case.  But they should not distrcat you too much, so you may want to consider them as a separate pull request later.

a) The dialog works, but does not "remember" the numeric items in the multiple receiver.  They seem to be cleared when you return to the dialog.
b) And the second receiver has the focus when you return, not the first.

Lastly a question for @Patowhiz ,partly because it may be more general than this dialog.
The default output for numeric columns is a bit of a mess, see below:

image

There is one really simple solution - in Rstudio.  This is mentioned in the skimR documentation.  It is to give the options command, e.g 

`options (width=120)`

This also works fine in R-Instat - as a command (one-line script)  Then I get:

image

which is a lot nicer!

@Patowhiz How should we do this in R-Instat.  I assume the default is 80.  Is there any reason to have that as the R-Instat default, or what should we do?

Vitalis95:
Fixes #9252 
@rdstern @N-thony , have a look at it, also added `Width to Display` in Options dialog  with default =80

### DIFF
--- a/instat/clsInstatOptions.vb
+++ b/instat/clsInstatOptions.vb
@@ -30,6 +30,7 @@ Imports RDotNet
     Public bIncludeRDefaultParameters As Nullable(Of Boolean)
     Public iPreviewRows As Nullable(Of Integer)
     Public iMaxRows As Nullable(Of Integer)
+    Public iMaxWidth As Nullable(Of Integer)
     Public iMaxCols As Nullable(Of Integer)
     Public iUndoColLimit As Nullable(Of Integer)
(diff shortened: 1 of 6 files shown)


rdstern (approved):
@Vitalis95 excellent, both the dialog memory and the option in the controls.
@N-thony over to you.

PR #9291, opened 2024-12-02
Improved the display of shapefile data

### LINKED ISSUE #9160: R-Instat, probably reogrid has a problem with shape files
The `Climatic > File > Import and tidy Shape files` is used routinely for our climatic mapping work.

image

Here is an example of data from Zimbabwe.  I found the problem with similar data from Bangladesh.  I think it is with all the shape files, at least those from gadm.

image

Note that this is a challenging type of file for reogrid to be able to handle.  The one shown only has 91 rows, but the geometry variable is sufficiently large that the data frame is over 5 mbytes.  (The one where I found the problem was over 100 mbytes.)

a) The first problem - perhaps minor - is that R-Instat now runs slowly, and I assume this may be to do with refreshing the reogrid.  For example making (say) NAME_1 into a factor takes over 1 minute on the Bangladesh data.

b) So I then tried to do something about the geometry variable.  I tried right-click and delete, then right-click and rename.  Each time I get this error message and it then closes the data frame.  Other data frames stay open.

image

 @N-thony I wonder what can be done.  I note it adds special types the the metadata.  I fon't need to be able to dit, or do anything with the contents of the geometry variable.  It works, in that the maps are drawn.  I would just like reogrid to handle the data frame better - and also not throw me out of R-Instat.

I wonder if this problem may need some @ChrisMarsh82 advice?

N-thony:
Fixes #9160 
@rdstern, this tempts to fix the corresponding issue, so far the display of the geometry data has been replaced by "MULTIPOLYGON" and set the column to read only in the reogrid. Can you also check the speed issue and probable share with me the data you have used?

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -957,7 +957,6 @@ DataSheet$set("public", "rename_column_in_data", function(curr_col_name = "", ne
         if(self$column_selection_applied()) self$remove_current_column_selection()
         # Need to use private$data here because changing names of data field
         names(private$data)[names(private$data) == curr_col_name] <- new_col_name
-        
         if(any(c("sfc", "sfc_MULTIPOLYGON") %in% class(private$data[[curr_col_name]]))){
           # Update the geometry column reference
           sf::st_geometry(private$data) <- new_col_name
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -1005,9 +1005,7 @@ DataSheet$set("public", "rename_column_in_data", function(curr_col_name = "", ne
   } else if (type == "rename_with") {
     if (missing(.fn)) stop(.fn, "is missing with no default.")
     curr_col_names <- names(curr_data)
-    current_geom_col <- attr(private$data, "sf_column")
     private$data <- curr_data |>
-      
       dplyr::rename_with(
         .fn = .fn,
         .cols = {{ .cols }}, ...
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -956,7 +956,12 @@ DataSheet$set("public", "rename_column_in_data", function(curr_col_name = "", ne
         }
         if(self$column_selection_applied()) self$remove_current_column_selection()
         # Need to use private$data here because changing names of data field
-        names(private$data)[names(curr_data) == curr_col_name] <- new_col_name
+        names(private$data)[names(private$data) == curr_col_name] <- new_col_name
+        
+        if(any(c("sfc", "sfc_MULTIPOLYGON") %in% class(private$data[[curr_col_name]]))){
+          # Update the geometry column reference
+          sf::st_geometry(private$data) <- new_col_name
+        } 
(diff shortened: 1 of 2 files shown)


rdstern (commented):
@N-thony very promising so far.  I have tried with the Climsoft Stations file first.  This has the station data from Zambia, plus gadm files all saved as an rds file.  They now open and seem to work fine.
Now the Bangladesh files.  To be continued.

rdstern:
@N-thony I had another look at the data before you made the changes - so Bangladesh in 0.8.0 to try to see why it behaves so differently to the Zambia data.  It is because the grometry variable is much, much "longer".  
image

I notice that the R viewer has formatted the values in the geometry, and you have not.  I assume you can, and this may even help, because there is probably something ridiculous about the length of that field in the viewer.

rdstern:
@N-thony I wonder what next?  I have the impression that:
a) It doesn't crash now when I try to change a name.
b) It always displays nicely
c) The timing problems may largely be simply because it is a very large file for Bangladesh?

I'm still a bit concerned at the time it takes for other operations on that sheet, like making a variable a factor, or renaming another variable.  Is that also a problem on RStudio?

But overall I am wondering about approving and merging now, and maybe still wondering in a later issue about what remains?  Should we?

rdstern (approved):
@Patowhiz can you check please

Patowhiz (commented):
@N-thony thanks for this. The VB.Net implementation looks good.

@lilyclements, could you please double check the R changes made in this update? The modifications affect a general R function that is widely used by other functions. Would be good to ensure that I haven't missed any potential regressions.

Thank you.

instat/static/InstatObject/R/data_object_R6.R
```diff
@@ -956,7 +956,12 @@ DataSheet$set("public", "rename_column_in_data", function(curr_col_name = "", ne
         }
         if(self$column_selection_applied()) self$remove_current_column_selection()
         # Need to use private$data here because changing names of data field
-        names(private$data)[names(curr_data) == curr_col_name] <- new_col_name
+        if(any(c("sfc", "sfc_MULTIPOLYGON") %in% class(private$data[[curr_col_name]]))){
+          # Update the geometry column reference
+          sf::st_geometry(private$data) <- new_col_name
+        }
```
lilyclements, line 962: @N-thony can you explain this a bit to me? What if `"sfc""`, `"sfc_MULTIPOLYGON"` are not in there too?

lilyclements (commented):
@N-thony looks good to me, just a quick question on the R code

Patowhiz (approved):
Thanks @lilyclements for approving this. I'm now approving it as well.
@lilyclements I'm merging hoping @N-thony comments have answered your question.

Patowhiz:
@rdstern once you approve this, I'll merge it.

rdstern (approved):
@Patowhiz I've approved so you can merge.
@N-thony as well, just wanted to add why I am really happy with this.  I suggest we have made some really important progress recently.  At the start of the year there were quite a few mainly regression problems, etc and it all seemed rather unprofessional.  
This was disappointing partly because R-Instat is so clearly very useful to the NMSs, and some will be soon really depending on it.  It is a good partner to Climsoft and between them, they are highly appropriate for many countries.
Now, recently, it is looking so much better.  In the recent workshops there were no - and I mean no - big problems, just small cosmetic issues - like this one to sort out.  

So all becoming very satispactory.

Issue #9280, opened 2024-11-27
Add a "labels with" button to the rename dialog?


rdstern:
This is maybe "the icing on the cake" but might be an easy puzzle for @lilyclements first, and then for someone else to implement?

We have started using the rename with option on the variable names.  Very powerful!  

Could we also apply this the get tidier variable labels?  Here is a discussion in [stack-overflow](https://stackoverflow.com/questions/72969805/rename-variables-with-variable-labels-in-r).

I suggest, to avoid clutter, we might make it a separate (4th) button at the top.  And maybe just the edit option?

Might it be easy?  It shows the power of R, currently, but also how messy it is as a language perhaps, and hence our challenge to bring these tasks within reach of non-R-whizzes!

And to encourage good statistical practice by making these tasks easy!

lilyclements:
1. Could it be an option to have not just Data Frames in the receiver, but the option to choose the Column Metadata on this dialog as a way to change the labels?
2. Agreed we could have the `starts with`, `ends with`, `matches`, and `contains` in the Find/Replace dialog.

rdstern:
@lilyclements and @N-thony I am really liking the improvements coming here.  I may be exaggerating the importance, but I think we can really be comprehensive here, and that is making a difficult job easier.

@lilyclements suggested a very simple addition to be able to rename with for the variable labels:

![image](https://github.com/user-attachments/assets/391054b8-23ed-4d67-95fc-b9b18af144c4)

Here is the dialog now.

a) Add a third radio button to this option, just saying `Labels`. 
b) @lilyclements can you confirm that with the Labels option we omit the first options.  So the whole box can have the label "Edit"?
c) So, in the box, the first label could be Command and then the rest is the same?
d) Then it needs you to give the code, so it edits the Variable label?

e) Now an addition for @N-thony .  Lily, I think you said that `Matches `is the only command that uses regex?  Could the group box be deeper, for both the `data frame` and the `labels` options, with the regex keyboard labelled here `Regex Keyboard`.  This is disabled except for the Matches option, when it is enabled!

lilyclements:
@rdstern
a) I can confirm that with the Labels option we can omit the first options. However, we don't have to if we want to have the other options.
b) I gave some code in my last message which is on amending the label to be the variable name. Would we want this as an additional radio button?
[...]

PR #9292, opened 2024-12-02
Adding option to rename labels

lilyclements:
Adding an option to rename labels, which has the same flexibility as our option to rename columns. This is as discussed in #9280

### DIFF
--- a/instat/static/InstatObject/R/data_object_R6.R
+++ b/instat/static/InstatObject/R/data_object_R6.R
@@ -1011,6 +1011,25 @@ DataSheet$set("public", "rename_column_in_data", function(curr_col_name = "", ne
       self$data_changed <- TRUE
       self$variables_metadata_changed <- TRUE
     }
+  } else if (type == "rename_labels"){
+    # to rename column labels. Here, instead of renaming a column name, we're giving new values in a column.
+    curr_metadata <- self$get_variables_metadata()
+    curr_col_names <- names(curr_data)
+    
+    new_metadata <- curr_metadata |>
+    dplyr::mutate(
+        dplyr::across(
+            {{ .cols }},
+            ~ .fn(., ...)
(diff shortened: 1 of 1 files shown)


rdstern:
@lilyclements sounds perfect.  I assume @Vitalis95 just needs to add a `Labels With`, as a 4th button at the top, (so just after Rename With) and then make sure it uses your code?

lilyclements:
@rdstern yes. Then everything else is the same code as for the previous tab, but with `type = "rename_labels"` not `type = "rename_with"`.

Vitalis95:
@lilyclements @rdstern , then also with this new tab  we include all the  the options in `Rename with` tab,   `Rename Options` box?

lilyclements:
@Vitalis95 yes we can, if @rdstern is happy with that too!

rdstern:
@Vitalis95 I'd be very happy

Vitalis95:
@lilyclements , there is an error when you select a column

`data_book$rename_column_in_data(data_name="Responses", type="rename_labels", .cols="Which.province.do.you.work.in.", .fn=tolower)`


image

Even when you get the column first,
```
colnames <- data_book$get_columns_from_data(data_name="Responses", col_names="Which.province.do.you.work.in.")
data_book$rename_column_in_data(data_name="Responses", type="rename_labels", .cols="colnames", .fn=tolower)
```

lilyclements:
@Vitalis95 thanks! Can you try again now? I hadn't set up the selecting function correctly before but it should be working now

lilyclements:
@Vitalis95 so it should change the labels associated to that variable. 

Open the variable metadata (View > Column Metadata), and add a label to some variables, and then try it

image

Vitalis95:
Thanks @lilyclements , I get it now

PR #9297, opened 2024-12-09
Update One Variable Summarise and changing the Format options to Table Options

### LINKED ISSUE #9288: Update One Variable > Summarise and connect to new tables sub-dialog
@N-thony over to you to allocate

The third option is Customised:

a) Change the Format Table to Table Options and tidy the code.  It currently goes to the old Format table.  I suggest that dialog and probably some code behind is now redundant.
b) Change the defaults on the Summaries to Count_non_missing and Mean.  Also update so it gives the count name, and no longer the count-non-misssing.  (I thought that would have been changed everywhere when it was changed in the main sub-dialog.  Please check, why the name is not updated?)

derekagorhom:
Fixes #9288 
This Fixes the part a for the issue where we change the Format Table to Table Options and connect it to the table sub-dialog.
for part b the changes done previously to change the summary  Count_non_missing are still there, please check below
image

### DIFF
--- a/instat/dlgOneVariableSummarise.Designer.vb
+++ b/instat/dlgOneVariableSummarise.Designer.vb
@@ -48,7 +48,7 @@ Partial Class dlgOneVariableSummarise
         Me.rdoVariable = New System.Windows.Forms.RadioButton()
         Me.rdoSummary = New System.Windows.Forms.RadioButton()
         Me.grpColumns = New System.Windows.Forms.GroupBox()
-        Me.cmdFormatTable = New System.Windows.Forms.Button()
+        Me.ucrPnlColumnFactor = New instat.UcrPanel()
         Me.ucrReorderSummary = New instat.ucrReorder()
         Me.ucrSaveSummary = New instat.ucrSave()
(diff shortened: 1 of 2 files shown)


derekagorhom:
@rdstern can you review this
Thanks

rdstern (commented):
@derekagorhom thats, good as far as it goes.
I am not sure if it is making a gt table, because you can go into @Patowhiz sub-dialog fine, but it doesn't seem to use any changes you make.  For example I added a title and also made it an Excel theme, but all is ignored.  Maybe @N-thony can help?

derekagorhom:
@rdstern i have fixed the issue with the gt table, it should make titles and the themes work well now. 
Can you review this again
Thanks

rdstern (changes requested):
@derekagorhom that's great, well done.
Except that the layout used to work.  Now it only does the option with no column factor, whatever the radio button on the layout.  Once that's fixed I'm excited to see this merged.    See below:

image

derekagorhom:
@rdstern , I have fixed the issue with the columns radio buttom
Can you review this again
Thanks

Issue #8397, opened 2023-06-21
Removing a datatype


Fidel365:
If we declare a datatype, we do not have a way of removing it. For example in describe<one variable frequencies, we need to set the variables for the stem and leaf to be numeric and not for the Table and Graph, If we declare the datatype and we do not remove it, the   changes will also apply to Table and Graph. So we need to find a way of removing the datatype once it's  set.

rdstern:
@Fidel365 could you look at the code for the Describe > 2-variables > Summarise?
There are currently 2 options, namely skim and 2-variables.
a) In the multiple receiver for skim you can currently have any data type.
b) In the multiple receiver for 2-variables, they can only be of a single type.
When you fill the receiver in skim and then move to the 2-variable the multiple receiver is empty.  If you then return to skim, the receiver has been cleared.
This should allow the correct procedure where you empty the receiver on entry to the stem and leaf - and on returning from it.  Ideally you keep the contents when moving from table to graph.  It is just the stem and leaf that clears it.

rdstern:
@Fidel365 I haven't yet been checking your changes in the pull request for the 2 variable summarise.  That's partly because it has had many recent changes - including its name - it is now also for 3 variables.
But I wonder if you have mis-understood the message above from @N-thony in all this.
As I read the thread it started with an idea from you on the one-variable frequencies dialog.  Then @N-thony suggested you look at the 2-variable summarise.  BUT I read hgis post as assuming the code there may help you to solve your one-variable frequency problem, with stem-and-leaf, rather than set you working on the 2-variable summarise, in parallel with Vitalis.

Could you discuss and check this point maybe with both of them?

Fidel365:
@rdstern aah now I get it, I will just make the changes in the same PR.

Fidel365:
@N-thony this was fixed, closing..

PR #9302, opened 2024-12-11
Small changes to describe>two variable>summarize dialog

Fidel365:
Fixes issue #8397 partly.
@rdstern this fixes the problem with the summarize dialog, still looking on how to fix for frequencies

### DIFF
--- a/instat/dlgDescribeTwoVariable.vb
+++ b/instat/dlgDescribeTwoVariable.vb
@@ -705,6 +705,7 @@ Public Class dlgDescribeTwoVariable
         cmdMissingOptions.Visible = False
         ucrChkOmitMissing.Visible = False
         If rdoSkim.Checked Then
+            ucrReceiverFirstVars.strSelectorHeading = "Variables"
             clsDummyFunction.AddParameter("checked", "skim", iPosition:=0)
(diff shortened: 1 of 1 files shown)


Fidel365:
@rdstern I see this dialog still has the same problem so this could be useful?

Fidel365:
@rdstern the changes are here clears the receiver when moving to and from the stem and leaf radio button.I also removed the uneeded changes for the describe two variable dialog

rdstern (approved):
@Fidel365 I'm pleased this was your issue and it's now fine, so I'm approving.
@N-thony over to you to check and merge.  Fidel says he has taken out the changes to the 2-variable dialog, so this is just his issue on stem and leaf.
Good to get it merged.
@fidel, if you would like to make another issue on step-and-leaf, here are the current results:
image

a) I don't see why we need those numbers in square brackets, though they do no harm.  But I don't like the Nulls at the bottom.
b) I really miss the fact there is no information on the variable being analysed.  Could it say Stem and leaf for xxxx, where xxxx is the variable analysed?

This would be a new issue and then pull request.  Not urgent, so I'dd include the timeline Version 0.8.1 if you do.

Issue #8536, opened 2023-09-09
Insert rows gives errors when rows are not the positive integers


rdstern:
<img width="509" alt="image" src="https://github.com/africanmathsinitiative/R-Instat/assets/11226469/84915490-0f67-4524-b8ce-aeaddfd5ff5d">

This is - I hope just poor programming and so should be easy to fix?  It is the row right click or Prepare > Data Frame > Insert Rows/Columns dialog.  The bug is in our Insert_row_in_data function.

Note this is because I am using an example where the row variable is not the positive integers - well it is, but they go in 2's rather than from 1 upwards.  I assume the function can find the r'th row of the data without using the row variable.

Note that in the datasets package the mtcars data also doesn't work, because the makes of car are the row variable there.  So getting insert rows to work there will be the same problem.

lilyclements:
@rdstern I don't really understand your data and how you have 2, 4, 6, 8, ... as row numbers. Do you mind sharing how that happened so I can investigate?

A current solution here could be to change the start of the code to:

```
DataSheet$set("public", "insert_row_in_data", function(start_row, row_data = c(), number_rows = 1, before = FALSE) {
  curr_data <- self$get_data_frame(use_current_filter = FALSE)
  curr_row_names <- rownames(curr_data)
  if (!start_row %in% curr_row_names) {
        curr_row_names <- as.numeric(make_factor(rownames(curr_data)))
          if (!start_row %in% curr_row_names) {
              stop(paste(start_row, " not found in rows"))
          }
  }
  row_position = which(curr_row_names == start_row)
  row_data <- curr_data[0, ]
  for(i in 1:number_rows) {
    row_data[i, ] <- NA
  }
... etc etc
```

This will change the rownames to be numbers for mtcars when applied there, however, so we would most definitely want a warning at the very least for the user to know that this occurs!

An alternative is that we set the rownames to be a variable, and then switch it back out. However, that's creating new issues in the system. 

It would be good to know your data @rdstern so that I can work out if this is a problem that will be resolved by switching to tibbles. I checked in filter and it is not a problem there.

rdstern:
@lilyclements great that you are going through this.
[...]

PR #9303, opened 2024-12-11
Improvements to insert row dialog for dataframes with rownames

Fidel365:
Fixes issue partly #8536 
@rdstern and @lilyclements I have added the tibble function to check for rownames, when we get rownames;  the error mesage is displayed.


Fidel365:
@rdstern still looking for a way to initiate this better in the dialog, I'm still having a look but I thought a better error message could be useful, looking forward to this while exploring different ways to solve this.

rdstern (changes requested):
@Fidel365 here is your solution for the dialog:

image

a) I don't usually use the dialog, I use the right-click in the row names, and then Insert Rows.  Can you trap the error message there and give the same message as shown above.
b) I was trying to think of the correct message and yours is better than my attempts.  So great.  Why did I not think of using the Help.  @rachelkg we must make sure that it is explained there.
c) But can you edit your message a bit.  Make a space in `aresome` and leave the s off `details`  Also could you have the `We explain in detail in the help.` on a new line.

lilyclements:
This should instead be performed in `databook` now. I have moved this PR into an issue there ([233](https://github.com/IDEMSInternational/databook/issues/233))

PR #9311, opened 2024-12-16
Fixes bug in adding variable in multiple receiver

### LINKED ISSUE #9306: Adding and display problem in multiple receiver
@N-thony here is an example:

image

There are 3 data frames and I right-click on the second and choose Delete.  Now:

a) The name of the data frame is cut-off.  I don't know why we can't see more of the name?
b) More serious is that I can't add the 3rd data frame into the receiver.  I can double-click or click and press Add and it doesn't go there.

In the workshop this same problem was found in various dialogs including the Prepare > Data Reshape > Column Summaries.  So I think there is a glitch in the general code for moving single variables from the selector to a multiple receiver, that is already partially full.

Note I can clear from the receiver and then add the two data frames.  So participants have always found a work-around, but noce to fix, so it works more smoothly.

N-thony:
Fixes #9306 I hope!

@derekagorhom have a look of this PR, it attempts to fix b) from the issue by adding of variable in the multipe reveiver, I think the issue was in the code with `FindItemWithText` which might only check for the beginning of the text (prefix matching) instead of the full string, or it could be performing partial matches therefore we were not able to add another variable.
You can work on item a)

### DIFF
--- a/instat/ucrReceiverMultiple.vb
+++ b/instat/ucrReceiverMultiple.vb
@@ -47,7 +47,8 @@ Public Class ucrReceiverMultiple
         'first eliminate all items that already exist
         'this improves perfomance significantly for wide data sets
         For Each kvpTempItem As KeyValuePair(Of String, String) In lstItems
-            If lstSelectedVariables.FindItemWithText(kvpTempItem.Value) Is Nothing Then
+            Dim item As ListViewItem = lstSelectedVariables.FindItemWithText(kvpTempItem.Value)
+            If item Is Nothing OrElse Not item.Text.Equals(kvpTempItem.Value, StringComparison.Ordinal) Then
                 lstActualItemsToAdd.Add(kvpTempItem)
             End If
         Next
--- a/instat/ucrReceiverMultiple.vb
+++ b/instat/ucrReceiverMultiple.vb
@@ -47,11 +47,18 @@ Public Class ucrReceiverMultiple
         'first eliminate all items that already exist
         'this improves perfomance significantly for wide data sets
         For Each kvpTempItem As KeyValuePair(Of String, String) In lstItems
-            Dim item As ListViewItem = lstSelectedVariables.FindItemWithText(kvpTempItem.Value)
-            If item Is Nothing OrElse Not item.Text.Equals(kvpTempItem.Value, StringComparison.Ordinal) Then
+            Dim isMatchFound As Boolean = False
+            For Each item As ListViewItem In lstSelectedVariables.Items
+                If item.Text.Equals(kvpTempItem.Value, StringComparison.Ordinal) Then
+                    isMatchFound = True
+                    Exit For
+                End If
+            Next
+            If Not isMatchFound Then
                 lstActualItemsToAdd.Add(kvpTempItem)
             End If
         Next
+
 
         If lstActualItemsToAdd.Count = 0 Then
             Exit Sub


derekagorhom:
@N-thony, It works now

rdstern (commented):
@N-thony I believe you may have found that one.  Note it became apparent in more dialogs, including the summarise one.  Really noire to make R-Instat run more smoothly.
@derekagorhom as Antoine says, I hope you can fix the other feature in this issue, maybe in the same branch?

rdstern:
@derekagorhom are you ok to complete this pull request, or do you need help, or will it take time - in which case we could merge this change and have a new pull request?

derekagorhom:
@rdstern this pull request can be merged, it will be done on another pull request

rdstern (dismissed):
@N-thony,  @derekagorhom has suggested this be merged now and the other part becomes a separate pull request.
@Patowhiz please can you "do the honours"

instat/ucrReceiverMultiple.vb
```diff
@@ -47,7 +47,8 @@ Public Class ucrReceiverMultiple
         'first eliminate all items that already exist
         'this improves perfomance significantly for wide data sets
         For Each kvpTempItem As KeyValuePair(Of String, String) In lstItems
-            If lstSelectedVariables.FindItemWithText(kvpTempItem.Value) Is Nothing Then
+            Dim item As ListViewItem = lstSelectedVariables.FindItemWithText(kvpTempItem.Value)
+            If item Is Nothing OrElse Not item.Text.Equals(kvpTempItem.Value, StringComparison.Ordinal) Then
```
Patowhiz, line 51: @N-thony based on the official documentation in this [link](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.listview.finditemwithtext?view=windowsdesktop-9.0) , I think it's safer to use a custom loop that searches for a full string match in the list of selected variables instead of relying on what is returned by `FindItemWithText`.

Patowhiz (changes requested):

Patowhiz:
@rdstern Could you please test this with wide datasets to check for any significant performance degradation? Specifically, try using dialogs that have multiple receivers, as well as the Pivot Table dialog.

The code looks good, but I’d like to evaluate whether optimisation is necessary for wide datasets.

rdstern:
@N-thony I was looking for @Patowhiz wide data set when I stumbled over a different problem.
Try with our usual survey and `Describe > One Variable Summarise`.  Then add all variables and press Ok.

I get an error as follows:

image

I give the code above and the error is correct, there is no object.

If you press Run again it is fine and the code is different - and back to what it should be.

Patowhiz:
@rdstern is that bug found in this PR only or it is also in the master branch. I'm also still waiting on you feedback on wide data sets.
Thanks.

rdstern:
@N-thony there isn't an issue on this.  I just stumbled over it, when testing this pull request for wide data sets.  So it was just reported here.  I'm now releived that the bug appeas=rs to be resolved, though I'd be even happier to discover why it happened as well.  Now I'm back to testing this one with wide data.

rdstern (approved):
@Patowhiz I tried with the first dataset from the extRemes package - with 12,000 variables.
It took 123 seconds to load into R-Instat, but then (in summarise) was reasonably instantaneous in transferring variables into the multiple receiver.
It did expose a small bug with wide data, but not to do with this code change.  So I'm approving.  Good to have fixed the bug.
@N-thony or @Patowhiz to merge.  Thanks to both.

Patowhiz (approved):

PR #9317, opened 2024-12-18
Added the table options  into the Summarise 2/3 variables dialog

### LINKED ISSUE #9286: Add the table options back into the Summarise 2/3 variables dialog.
@Vitalis95 this used to be there.  I list below, where the button should be re-instated and hope this might be a simple task - with some exceptions for now!

As we progress I suggest we merge progressively.

**_ 2 variables frequency table(s)_**

image

a) There should be the Margins option here, could you add it?
b) Could you explain what is stored here?  Is it a set of gt objects, or data frames?
c) Could you add the `Table Options` button, I assume below the Percentages group box?
Then also for @Patowhiz I assume this is tricky for multiple tables?  If so, then for now have the button, but it is only enabled when there is a single table.
d) Can the `Summary Variable` column (in the table) be omitted?  Ideally `perc_count_all_village`, etc is the title for the table instead?

**_2 variable summary table(s)_**

image

a) There is a half-hidden button saying Format Table.  Re-instate that and call it Table Options.  Again I assume it is disabled for multiple tables
b) The table options will permit us to present the decimals as we wish.  I suggest that is more natural if the summaries are the columns.  So swap the rows and columns.  I hope the names can be just the summary, without the yield there.  Can Summaries for yield become a title?
c) Add the Margins group box.  I hope that is coded?  They are the overall values, with the corresponding summary, i.e. count, min, mean, max.

That's enough for now.  Once you are ready we'll do the 3 variables!

Vitalis95:
Fixes #9286 
@rdstern @N-thony , added Table Options .

### DIFF
--- a/instat/dlgDescribeTwoVariable.Designer.vb
+++ b/instat/dlgDescribeTwoVariable.Designer.vb
@@ -99,10 +99,9 @@ Partial Class dlgDescribeTwoVariable
         Me.rdoThreeVariable.FlatAppearance.CheckedBackColor = System.Drawing.SystemColors.ActiveCaption
         Me.rdoThreeVariable.FlatStyle = System.Windows.Forms.FlatStyle.Flat
         Me.rdoThreeVariable.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.rdoThreeVariable.Location = New System.Drawing.Point(432, 12)
-        Me.rdoThreeVariable.Margin = New System.Windows.Forms.Padding(4)
+        Me.rdoThreeVariable.Location = New System.Drawing.Point(288, 8)
         Me.rdoThreeVariable.Name = "rdoThreeVariable"
-        Me.rdoThreeVariable.Size = New System.Drawing.Size(150, 42)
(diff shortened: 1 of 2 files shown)


rdstern:
@Vitalis95 started by checking 2-way frequency tables.  And that is a wow, from you and @lilyclements I think!
Here is the first result:

image

You have managed to get this to work with multiple tables.  I wasn't expecting that!  This is really important, for one of the important points I am making generally now, and particularly to climatic staff.  They are all addicted to spreadsheets.  I would like to tempt them with the idea you can now easily do great tables and graphs, and easily,  without going back to Excel for the presentation work.  So thyese are deliberately Excel theme tables.
@Patowhiz note in these tables that we need the option to hide variables, which I assume is coming in your updates.
I'll check a bit more!

rdstern (approved):
@Vitalis95 excellent
@N-thony this is a great advance, and one I have been waiting for, for a long time.  Very well done Vitalis on all this.  @N-thony Please check and merge when you can.
It will also be further improved with the new table features from @Patowhiz.  They are imminent.

PR #9320, opened 2024-12-19
Adds hiding of columns, replacing missing values, grand summaries, nano plots and group summaries features

Patowhiz:
@rdstern this is ready for your review.
This PR adds the following in relation to  `gt` R package:
1. Hiding of columns.
2. Nano plots.
3. Replacing missing values.
4. Group summaries.
5. Grand summaries.

It doesn't exhaust the functionalities of the above `gt` features but attempts to simplify how to use them.

### DIFF
--- a/instat/UserTables/Columns/ucrColumnHide.Designer.vb
+++ b/instat/UserTables/Columns/ucrColumnHide.Designer.vb
@@ -1,0 +1,83 @@
+﻿<Global.Microsoft.VisualBasic.CompilerServices.DesignerGenerated()> _
+Partial Class ucrColumnHide
+    Inherits System.Windows.Forms.UserControl
+
+    'UserControl overrides dispose to clean up the component list.
+    <System.Diagnostics.DebuggerNonUserCode()> _
+    Protected Overrides Sub Dispose(ByVal disposing As Boolean)
+        Try
+            If disposing AndAlso components IsNot Nothing Then
+                components.Dispose()
+            End If
+        Finally
+            MyBase.Dispose(disposing)
+        End Try
+    End Sub
+
+    'Required by the Windows Form Designer
+    Private components As System.ComponentModel.IContainer
+
+    'NOTE: The following procedure is required by the Windows Form Designer
+    'It can be modified using the Windows Form Designer.  
+    'Do not modify it using the code editor.
+    <System.Diagnostics.DebuggerStepThrough()> _
+    Private Sub InitializeComponent()
+        Me.ucrReceiverMultipleCols = New instat.ucrReceiverMultiple()
+        Me.Label1 = New System.Windows.Forms.Label()
+        Me.ucrSelectorCols = New instat.ucrSelectorByDataFrameAddRemove()
+        Me.SuspendLayout()
+        '
+        'ucrReceiverMultipleCols
+        '
+        Me.ucrReceiverMultipleCols.AutoSize = True
+        Me.ucrReceiverMultipleCols.frmParent = Nothing
+        Me.ucrReceiverMultipleCols.Location = New System.Drawing.Point(335, 37)
+        Me.ucrReceiverMultipleCols.Margin = New System.Windows.Forms.Padding(0)
+        Me.ucrReceiverMultipleCols.Name = "ucrReceiverMultipleCols"
+        Me.ucrReceiverMultipleCols.Selector = Nothing
+        Me.ucrReceiverMultipleCols.Size = New System.Drawing.Size(180, 123)
+        Me.ucrReceiverMultipleCols.strNcFilePath = ""
+        Me.ucrReceiverMultipleCols.TabIndex = 356
+        Me.ucrReceiverMultipleCols.ucrSelector = Nothing
+        '
+        'Label1
+        '
+        Me.Label1.AutoSize = True
+        Me.Label1.ImeMode = System.Windows.Forms.ImeMode.NoControl
+        Me.Label1.Location = New System.Drawing.Point(340, 12)
+        Me.Label1.Margin = New System.Windows.Forms.Padding(4, 0, 4, 0)
+        Me.Label1.Name = "Label1"
+        Me.Label1.Size = New System.Drawing.Size(85, 20)
+        Me.Label1.TabIndex = 357
+        Me.Label1.Text = "Column(s):"
+        '
+        'ucrSelectorCols
+        '
+        Me.ucrSelectorCols.AutoSize = True
+        Me.ucrSelectorCols.bDropUnusedFilterLevels = False
+        Me.ucrSelectorCols.bShowHiddenColumns = False
+        Me.ucrSelectorCols.bUseCurrentFilter = True
+        Me.ucrSelectorCols.Location = New System.Drawing.Point(9, 6)
+        Me.ucrSelectorCols.Margin = New System.Windows.Forms.Padding(0)
+        Me.ucrSelectorCols.Name = "ucrSelectorCols"
+        Me.ucrSelectorCols.Size = New System.Drawing.Size(289, 292)
+        Me.ucrSelectorCols.TabIndex = 355
+        '
+        'ucrColumnHide
+        '
+        Me.AutoScaleDimensions = New System.Drawing.SizeF(9.0!, 20.0!)
+        Me.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Font
+        Me.Controls.Add(Me.ucrReceiverMultipleCols)
+        Me.Controls.Add(Me.Label1)
+        Me.Controls.Add(Me.ucrSelectorCols)
+        Me.Name = "ucrColumnHide"
+        Me.Size = New System.Drawing.Size(523, 312)
+        Me.ResumeLayout(False)
+        Me.PerformLayout()
+
+    End Sub
+
+    Friend WithEvents ucrReceiverMultipleCols As ucrReceiverMultiple
+    Friend WithEvents Label1 As Label
+    Friend WithEvents ucrSelectorCols As ucrSelectorByDataFrameAddRemove
+End Class
--- a/instat/UserTables/Columns/ucrColumnHide.resx
+++ b/instat/UserTables/Columns/ucrColumnHide.resx
@@ -117,10 +117,4 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <metadata name="colExpression.UserAddedColumn" type="System.Boolean, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089">
-    <value>True</value>
-  </metadata>
-  <metadata name="colExpression.UserAddedColumn" type="System.Boolean, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089">
-    <value>True</value>
-  </metadata>
 </root>
--- a/instat/UserTables/Columns/ucrColumnHide.vb
+++ b/instat/UserTables/Columns/ucrColumnHide.vb
@@ -1,0 +1,56 @@
+﻿Imports Antlr.Runtime
+
+Public Class ucrColumnHide
+    Private clsOperator As New ROperator
+    Private bFirstload As Boolean = True
+    Private clsColHideRFunction As RFunction
+
+    Private Sub ucrColumnHide_Load(sender As Object, e As EventArgs) Handles MyBase.Load
+        If bFirstload Then
+            InitialiseControl()
+            bFirstload = False
+        End If
+    End Sub
+
+    Private Sub InitialiseControl()
+        ucrReceiverMultipleCols.Selector = ucrSelectorCols
+        ucrReceiverMultipleCols.SetMeAsReceiver()
+    End Sub
+
+    Public Sub Setup(strDataFrameName As String, clsOperator As ROperator)
+        Me.clsOperator = clsOperator
+
+        ' Set up the selector and receiver
+        ucrSelectorCols.SetDataframe(strDataFrameName, bEnableDataframe:=False)
+        ucrReceiverMultipleCols.Clear()
+
+        Dim lstRParams As List(Of RParameter) = clsTablesUtils.FindRFunctionsParamsWithRCommand({"cols_hide"}, clsOperator)
+        If lstRParams.Count > 0 Then
+            clsColHideRFunction = lstRParams(0).clsArgumentCodeStructure
+            Dim arrColumnNames As String() = clsTablesUtils.SplitRText(clsColHideRFunction.GetParameter("columns").strArgumentValue)
+            For Each columnName As String In arrColumnNames
+                ucrReceiverMultipleCols.Add(columnName)
+            Next
+        Else
+            clsColHideRFunction = New RFunction
+            clsColHideRFunction.SetPackageName("gt")
+            clsColHideRFunction.SetRCommand("cols_hide")
+        End If
+    End Sub
+
+    Public Sub SetValuesToOperator()
+        ' Remove any previous cell footers 
+        Dim lstRParams As List(Of RParameter) = clsTablesUtils.FindRFunctionsParamsWithRCommand({"cols_hide"}, clsOperator)
+        For Each clsRParam As RParameter In lstRParams
+            clsOperator.RemoveParameter(clsRParam)
+        Next
+
+        ' Add new changes
+        If Not ucrReceiverMultipleCols.IsEmpty Then
+            Dim strParamValue = ucrReceiverMultipleCols.GetVariableNames(bWithQuotes:=False)
+            clsColHideRFunction.AddParameter("columns", strParameterValue:=strParamValue, iPosition:=0, bIncludeArgumentName:=True)
+            clsOperator.AddParameter(New RParameter(strParameterName:="cols_hide_param", strParamValue:=clsColHideRFunction, bNewIncludeArgumentName:=False))
+        End If
+    End Sub
+
+End Class
--- a/instat/UserTables/Columns/ucrColumnMissingTexts.Designer.vb
+++ b/instat/UserTables/Columns/ucrColumnMissingTexts.Designer.vb
@@ -1,0 +1,183 @@
+﻿<Global.Microsoft.VisualBasic.CompilerServices.DesignerGenerated()> _
+Partial Class ucrColumnMissingTexts
+    Inherits System.Windows.Forms.UserControl
+
+    'UserControl overrides dispose to clean up the component list.
+    <System.Diagnostics.DebuggerNonUserCode()> _
+    Protected Overrides Sub Dispose(ByVal disposing As Boolean)
+        Try
+            If disposing AndAlso components IsNot Nothing Then
+                components.Dispose()
+            End If
+        Finally
+            MyBase.Dispose(disposing)
+        End Try
+    End Sub
+
+    'Required by the Windows Form Designer
+    Private components As System.ComponentModel.IContainer
+
+    'NOTE: The following procedure is required by the Windows Form Designer
+    'It can be modified using the Windows Form Designer.  
+    'Do not modify it using the code editor.
+    <System.Diagnostics.DebuggerStepThrough()> _
+    Private Sub InitializeComponent()
+        Me.lblMissingTexts = New System.Windows.Forms.Label()
+        Me.btnClear = New System.Windows.Forms.Button()
+        Me.ucrSelectorCols = New instat.ucrSelectorByDataFrameAddRemove()
+        Me.dataGrid = New System.Windows.Forms.DataGridView()
+        Me.colMissingTexts = New System.Windows.Forms.DataGridViewTextBoxColumn()
+        Me.ucrReceiverMultipleCols = New instat.ucrReceiverMultiple()
+        Me.Label1 = New System.Windows.Forms.Label()
+        Me.lblMissingText = New System.Windows.Forms.Label()
+        Me.ucrTxtMissingText = New instat.ucrInputTextBox()
+        Me.btnAdd = New System.Windows.Forms.Button()
+        CType(Me.dataGrid, System.ComponentModel.ISupportInitialize).BeginInit()
+        Me.SuspendLayout()
+        '
+        'lblMissingTexts
+        '
+        Me.lblMissingTexts.AutoSize = True
+        Me.lblMissingTexts.ImeMode = System.Windows.Forms.ImeMode.NoControl
+        Me.lblMissingTexts.Location = New System.Drawing.Point(347, 303)
+        Me.lblMissingTexts.Margin = New System.Windows.Forms.Padding(4, 0, 4, 0)
+        Me.lblMissingTexts.Name = "lblMissingTexts"
+        Me.lblMissingTexts.Size = New System.Drawing.Size(108, 20)
+        Me.lblMissingTexts.TabIndex = 359
+        Me.lblMissingTexts.Text = "Missing Texts:"
+        '
+        'btnClear
+        '
+        Me.btnClear.ImeMode = System.Windows.Forms.ImeMode.NoControl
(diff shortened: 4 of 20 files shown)


rdstern:
@Patowhiz the first thing I wanted to check was the hiding of the columns.  It is excellent that the sub-dialog is now available from many dialogs.  But, in the hiding it is showing the original data file, (which is irrelevant), and not the summary file, which has the column I need to hide. 

image


Looking at more now.

rdstern:
> Looking at more now.

I next tried to replace missing and (of course) got an error - which is the same problem I presume.

image

rdstern:
@Patowhiz it is now 3 days since the problem I found, described above.  I would have a much better Christmas if you could fix this soon, or at least reply and convince me it is easy to fix.  Currently I have the fear that you might return again to only having gt for a presentation data frame, omitting the main use, which is a presentation (summary) table!   That complicated the work for many months and I assumed was resolved by the discussions when you visited England in October.

I still live in hope.

rdstern:
@Patowhiz that's very disappointing.  We had the team working on the one-variable summarise, two variables summarise and the main Tables > Summaries dialogs since your visit in October.  And you, and @ChrisMarsh82 knew all along that the main parts of your sub-dialog would not work for any of those dialogs.  I was really hoping that by mid-December our tables nightmare would be over.  Is the main nightmare just beginning?  Was @volloholic aware of this, following the October discussions?

You seem to be saying that there is quite complicated and important steps that still remain to be done, and that they are outside what you consider to have been your terms of reference?  Should I have known this?  How do you plan for your sub-dialog to work with the summaries?  Do you plan to be involved, at least to supervise, or to delegate?  If delegating, then to whom, and when?

Patowhiz:
@rdstern, I believe I have explained thoroughly (and, I hope, with minimal technical jargon) what I think is needed. Initially, I thought the requirement was simply to translate the `gt` package into a sub-dialog, similar to what we have for `ggplot`. However, it seems you might be looking for a custom R package that leverages `gt` to produce the specific tables you have in mind. 

This would also require new custom controls capable of interpreting the scripts generated by the main dialogs, enabling users to style or format the desired tables. As I’ve mentioned before, in my opinion, assuming the expected table outputs are clearly defined, this would involve a two-step process, each requiring a new set of components to be implemented. That said, I could be misunderstanding what is truly needed, so I’m eager to see what the team’s suggestions are.

In the meantime, I recommend focusing first on having the dialogs produce the desired outputs without styling. I think this can be done by any R-Instat developer as long as the inputs and expected outputs are clearly defined. This would allow us to fully understand the requirements for the sub-dialog and inform the subsequent implementation of the two-step process.

I also want to apologise for not probing more deeply into your initial vision for this. That was an oversight on my part.

Finally, to clarify, the current format sub-dialog assumes you are formatting or styling any data frame in the data book—that’s what `gt` essentially does: it presents a data frame. This PR builds on that assumption and aims to complete the `gt` features we couldn’t finish earlier due to the limitations of the supported R version.

lilyclements:
We have two stages to the tables:
1. Calculation
2. Design

I'm going to break this into three stages (breaking Design into two)
1. Calculation (our summary function)
2. Pivoting (`pivot_wider` function) - this only sometimes is performed depending on the users design choice
3. Styling (the `gt()` function).

image

**The issue**

In our `gt` table, the variable `village` does not exist, so we cannot call it. Instead we have columns called `KESEN`, `NANDA`, etc.

From @rdstern's code, it looks like you're trying to replace the `NA`s in the summary output in the village variable.
The code in VB is trying to do this to our `gt` object. Instead, this is something we want to do to our "long table" object, and then we can re-pivot and re-create the gt object from that.
This would mean that in terms of the R code, when the user edits a `gt` table, the R code is actually amending that `gt` table's corresponding "long table", and re-pivoting and re-building the `gt` object afterwards.


**To summarise**

My understanding is that when we save the `gt` table, we want to save the corresponding "long table" with it. Then, we can call the variables from our original data frame to amend things in our `gt` table.

(I would say we also want to save the calculation that creates the wider table too, so that those column names (KESEN, NANDA, etc, in our example) can be accessed in the receiver in the dialogs if the user wished to change something on them - e.g., case, colour, etc).

rdstern:
@lilyclements delighted you are attacking this topic.  There are currently 3 dialogs that summarise data and then are able to call gt - and it is all working fine for aspects that are not column or presumably row, or cell dependent.  
In your strategy I wonder if looking at what the gtsummary package does, could be useful.  That does summaries - as we do, and then (now) calls gt for the final presentation.  It has an intermediate step where it produces a gtsummary object, and then uses that to call gt (or other presentartion engines - but now seemingly always suggests gt).  

In gtsummary, the summaries can (I think) be the sort we produce, or regression/modelling summaries, so what we would term model objects can also be presented nicely in production tables via gt.  Perhaps gt could present our anova tables neatly.  No let's stick for now, with the summaries!!!

lilyclements:
@rdstern I want to update and add some corrections to what my previous message said, following our meeting earlier today. 
Should I put this as a separate issue?

## What do we display on the sub-dialogs when creating a table
We should not display any options on amending the table contents - such as the column or row names - since the table does not yet exist. This can all come later in an edit-table dialog.
We can add other options -- but not any that impact the table grid itself (at least for now). So, we can have options on, say, adding a title, a subtitle, etc, to the table. Just nothing that calls the actual table contents itself.
For now we could either:

a) disable either the sub-dialog completely, and go through the options a bit more systematically to decide what we can and cannot display
b) disable the options relating to the grid, and do a more in-depth explore later on what makes sense and what does not.

@rdstern @Patowhiz what do you think?

## How can we call the columns to edit in the `gt` table?

When we have a `gt` object, and we want to specify columns to amend, we can quite easily access the data. There are two ways to access the column names.

### 1. Approach One: Calling the data frame using `_data`

Let's use the survey data and say we create a table of our three factors. Our table is called `last_table`. Then to get a "R-data-frame" table of it, then we run `last_table$_data`:

```
frequency_table <- data_book$summary_table(data_name="survey", factors=c("village","variety","fertgrp"), store_results=FALSE, treat_columns_as_factor=FALSE, summaries=count_label)
last_table <- (frequency_table %>% pivot_wider(names_from=c(village), values_from=value)) %>% gt::gt()

# get a tibble of the gt object. 
last_table$`_data`
```

On the left we have a tibble of our `gt` object, on the right we have the `gt` object itself:

image

So, we can get the variables in our `gt` table object by internally running `last_table$_data` to get the column names, not `last_table` (the `gt` object).


**But what if we have two headers?**

```
frequency_table <- data_book$summary_table(data_name="survey", factors=c("village","variety","fertgrp"), store_results=FALSE, treat_columns_as_factor=FALSE, summaries=count_label)
last_table <- (frequency_table %>% pivot_wider(names_from=c(village, variety), values_from=value))%>%
  gt::gt() %>%
  gt::tab_spanner_delim(delim = "_")

# get a tibble of the gt object. 
last_table$`_data`
```

On the left we have a tibble of our `gt` object, on the right we have the `gt` object itself:

image

As before, we can get the variables in our `gt` table object by internally running `last_table$_data` to get the column names, not `last_table` (the `gt` object). But then, how do you distinguish `KESEN_OLD` from `KESEN_TRAD` if you just want to amend all of `KESEN`. 

I would argue that using this approach isn't the right way because of having multiple headers. But I'd like to hear your thoughts @Patowhiz @rdstern 

### 2. Approach Two: Calling the column names metadata using `_boxhead`

As before, let's use the survey data and say we create a table of our three factors. Our table is called `last_table`. This time, we will get the column metadata using `last_table$_boxhead`:

```
frequency_table <- data_book$summary_table(data_name="survey", factors=c("village","variety","fertgrp"), store_results=FALSE, treat_columns_as_factor=FALSE, summaries=count_label)
last_table <- (frequency_table %>% pivot_wider(names_from=c(village), values_from=value)) %>% gt::gt()

# get a tibble of the gt object. 
last_table$`_boxhead`
```

On the left we have the metadata of our columns in the `gt` object, on the right we have the `gt` object itself:

image

There are two columns here to draw attention to: `var`, which is the variable used for that column, and `column_label`, which is the name given in the R code. If we amend this metadata, then it will reflect in our table. 

E.g.,

```
last_table$`_boxhead`$column_label[[7]] <- "Sabey"
```

results in a table with "Sabey" not in capitals. (We would do this in a more elegant way / a way which works with R-Instat. Currently in R-Instat we change column values with a grid option, for example).

I appreciate this is a bit more work. However, this works a lot better when we come to tables with two rows of column headers.

**Two column headers**

```
last_table <- (frequency_table %>% pivot_wider(names_from=c(village, variety), values_from=value))%>%
  gt::gt() %>%
  gt::tab_spanner_delim(delim = "_")

# In here we can see the column_label info under column_label
last_table$`_boxhead`
```

gives the following. Our `gt` table is on the right and our `column_label` metadata is on the left:

image


### Overall

I'm not sure how we want to do it. But I can see several options:

1. We can use the first approach, which uses the `_data` option, but does not work as nicely when we have two column headers
2. We can use the second approach, which calls the metadata. If we do this, do we want to have that the user amends the unique values in the `column_labels` row? This would mean that we set it up that if you call "OLD" once, it will change all of the "OLD"'s
3. We can use the second approach, which calls the metadata. What if we had a grid where the user could change the name of the column labels? Would that work for other changes to the column, such as making it bold?
4. A mix of both! Use a grid and `_column_labels` when amending the name in the columns. Use `_data` for getting column names for amending other items in there like making it bold or changing colours etc.
5. Another option?

Any feedback or questions would be useful for us to then discuss on this further. Happy to move this into an issue for discussion if we don't want to here.

rdstern:
@lilyclements I hope @Patowhiz will respond more generally.  I'll also have a careful look tomorrow.  

I note for now that there are currently 3 summary dialogs that include the gt tables sub-dialog button.  They are 
a) Describe > One Variable > Summarise (Customised Option), 
b) Describe > Two/Three Variables Summarise (Two Variables and three Variables) 
c) Describe > Tables > Summaries dialog.

Both a) and b) have a pivot wider of a maximum of one factor.  It is only the most important c) that could currently have more than 1.

lilyclements:
> Both a) and b) have a pivot wider of a maximum of one factor. It is only the most important c) that could currently have more than 1.

@rdstern this is a very useful distinction. Regardless how our table looks, I suppose in all three cases we do not want any 
options on amending the table "grid" itself - instead just on, say, adding a title, a subtitle, etc, to the table. @Patowhiz what do you think?

Then anything on editing the table "grid" itself can come in a later dialog - using either approach 1 or 2.

Patowhiz:
@lilyclements yes, we could initially just focus on amending the table "grid" on the main dialogs. The main dialogs have all the information about the final table and therefore they can have controls that amend the table "grid". Then the sub-dialog can be used for adding general `gt` stuff that are applicable to any table, for instance, title, subtitle, themes etc. The approaches mentioned can also be implemented at the individual dialog level.

I think that approach can accelerate development of what's desired in each dialog.

rdstern:
@lilyclements asked (I think) what is needed from the gt sub-dialog for the "simple" tables, i.e. those where there is at most pivot wider by one factor:

Here are two tables from the 2/3 way summary dialog:  they are without going into the gt sub-dialog.

image
  Then also 

image

They appear one at a time in the Maximise window.  I have not examined the store option as to whether they are one or two objects?

a) I note there are multiple tables produced at the same time.  Currently we can use the gt-sub dialog and I assume it will act in exactly the same way on each table.  Nice if it does, but not a tragedy is the Table Options button is only available when we produce one table?
b) I would like to be able to hide columns - as you can see here.  Or maybe the summarise shouldn't give the column with the variable repeated?
c) I would like to replace NA.  Possibly the default should be to make it blank?
d) Good to be able to format the rows, and/or columns, e.g 1 decimal, add dollar sign?

lilyclements:
> a) I note there are multiple tables produced at the same time. Currently we can use the gt-sub dialog and I assume it will act in exactly the same way on each table. Nice if it does, but not a tragedy is the Table Options button is only available when we produce one table?

Yes, it will act the same way on each table since we will run it in the mapping statement

> b) I would like to be able to hide columns - as you can see here. Or maybe the summarise shouldn't give the column with the variable repeated?

We can make this the case when there is one variable repeated by default. I think this can be an option by default, but, we wouldn't want to get into hide/show columns in the sub-dialog options when generating the table since the table object doesn't exist until OK is pressed.

> c) I would like to replace NA. Possibly the default should be to make it blank?
> d) Good to be able to format the rows, and/or columns, e.g 1 decimal, add dollar sign?

(c) and (d) are similar in their approach and potential issues they can introduce. By setting `""` to be an option, and/or by adding units like dollar signs, we have to set these numerical variables to be characters. This is absolutely fine, but then of course means we can't change the decimal places afterwards.
If we set to be a currency, then it is easy as we know to just do 2dp. But, if we set to be a different type of unit, it would be difficult for the user to convert it back afterwards (the code isn't impossible though!).
Other than setting decimal places, what other options would we want that apply to only numerical variables?

We also mentioned on our call with David that we want to make any changes to the table itself after the table has been created, and so potentially these are options to add into the presentation/edit table dialog.
Options like titles, footnotes, etc (bits that don't affect our actual cells) can be made when the `gt` object is created.

lilyclements:
> @lilyclements yes, we could initially just focus on amending the table "grid" on the main dialogs. The main dialogs have all the information about the final table and therefore they can have controls that amend the table "grid". Then the sub-dialog can be used for adding general gt stuff that are applicable to any table, for instance, title, subtitle, themes etc. The approaches mentioned can also be implemented at the individual dialog level.

@Patowhiz glad we are agreed on this. We should define next steps. 

1. I do not know what is on the current sub-dialogs, but we want these to have bits that don't amend the table "grid" itself (table values - just headers, footers, etc). Who should do that?
2. I do not know what is on the "Use Table" dialog at the moment. I've outlined how we can access the data behind the gt table object above. Does this mean you are now happy to go ahead and implement the changes on the Use Table dialog? What is needed from me?

rdstern:
@berylwaswa I hope you can help me to test this branch?  It will be quite a challenge!  I am keen to start with the nanoplots!
The examples of data are all in the library, in the gt package.  I would like to see how many of the examples [from here](https://gt.rstudio.com/reference/cols_nanoplot.html) we can reproduce using the current dialog?  I suggest we start, in each case, by checking that we can run an example from the script window.  I copied the simplest, namely illness below and explain what we need to do:  
Here it is, with the top 2 lines added, so it runs in R-Instat:

image

Now I would like to get this table, and the subsequent ones via R-Instat including Patrick's sub-dialog.

The first line in the script can be done by:
a) Prepare > Data Frame > Row Names/Numbers, just press Ok.  
b) Filter for row <= 10.

Then go to Describe > Tables > Presentation with test, units and all the day variables:

image

Neat eh.

There are 6 examples.  Each can easily be copied and pasted into successive script windows.  The 2 extra lines above, are only needed once per session.  So the other scripts all run as they are. 

I suggest the code for each, shows how wonderful gt is for R whizzes, and also how complicated it looks for mere mortals!  So a solution that uses dialogs, or a mixture of dialogs, plus tweaking the corresponding scripts could make them accessible to many more - especially our typical R-Instat users.

berylwaswa:
@rdstern I have copied the codes and added the first lines you said and each was produced via the R-Instat script.

msedge_1RRXxDWlvx
msedge_Cu8TX6cTIG
msedge_Jrb3Arskys
msedge_nRPajpkSbG
msedge_qdg4Q81mNL
msedge_Xf3m6n2BeJ

I then followed the steps you gave to produce the tables for the 6 and below are the images of the tables produced. I labeled the 1 to 6 each representing the respective example.







 
And yes the tables are pretty neat!

berylwaswa:
@rdstern @Patowhiz 


I managed to create the nano plot for the first example with the illness data. It is pretty neat! 

This is the one produced by the code in the example;

msedge_pZXCP0D7Gz

Below is what I managed to produce;

instat_DFeR3sI5Ry

I then worked with @derekagorhom PR 9413 to add the title and the footnote. Is there a way to make the word test bold? And can I add the words in the brackets?

instat_WoU8JJPIpn

I think this came out beautifully I am now working on the second example.

berylwaswa:
@Patowhiz this is working well so far but when I go back to the sub-dialog, the receiver is cleared. I'd prefer it stays filled so I can edit it as I please if I'm still working with the same variables.

rdstern:
@Patowhiz I second the request from @berylwaswa, at least for the nano-plots, tab to include the "memory" for the other controls.  Also we wonder whether colours can be added to the lines or bars.
And two more general queries now.  
a) When we summarise data, and go straight into your sub-dialog, can you confirm that `Header , Themes and Other Styles` are available, but the remaining options are not?
b) I see you have Missing in the Columns tab.  Could there be a Missing main tab which gives options for the display of Missing for the whole table/dataframe, whatever the columns?

berylwaswa:
@Patowhiz the second example came out beautiful too. However I can't change the colors, so as @rdstern has suggested, we can have that feature added.

This how the one from the example looks like;

msedge_1RRXxDWlvx

And below is what I managed to make with R-Instat. For some reason, the nano plot button is disabled in the PR I was to use to add the title @derekagorhom so I am sharing this without the title.

instat_49Ce0PKScC

berylwaswa:
@rdstern I wanted to make our dates look like the ones in the example but noticed for our dates we only allow the day to be displayed as numeric. It would be nice to have them as abbreviations too.

berylwaswa:
@Patowhiz @rdstern 

I have been working on the third example, which is a combination of two nano-plots. 

msedge_Cu8TX6cTIG

I tried combining the two but kept getting the error below, maybe I missed something?

instat_4vTQduC2ZA

Then I produced them separately and they both came out okay see below.

instat_pPuG8xH1UR

instat_v9BnHxg6pn

@Patowhiz in the spanners tab the clear button for the expression is disabled so one has to reset the dialog in order to use it again, please enable it. 

instat_lfuNqFe0s9

@rdstern help me figure out how to combine the two, thanks.

berylwaswa:
@rdstern @Patowhiz 

I went through the six examples [here](https://gt.rstudio.com/reference/cols_nanoplot.html) and I managed to produce the first two. 

The third one produced the individual plots well but the example needed the two plots to be combined, I tried but didn't see how to do that with our current subdialog maybe tell me how?

I did not manage to produce the fourth, fifth, and sixth plots. These need more options in the sub dialog; and needed some data frame mutations that I got some trouble getting around but @rdstern helped me out with some.

Is it possible to add scatter plots to the plots we produce? I also noticed the fifth example produces a line plot of the median as the reference line. How do I specify this in R-Instat?

I think the tables sub dialog is impressive since it was my first time using it and I still managed to produce some good plots. I'll continue checking other examples from gt and let you know what I find.

rdstern:
No, @Patowhiz the problem is still there.  When a table is provided your sub-dialog needs to use it!

image

berylwaswa:
@Patowhiz I was making a table with the sp500 data in gt. The table came out okay. 

Although, the last column in the table is in billions R-Instat when producing the table makes it with the exponential value how do I make it come as the one in the example? 

Is there a way to identify a column as a currency in R-Instat the way it shows in the example?

See below;

|instat_CxuS6mO1Pd|
msedge_I5KZA8sbHy|

berylwaswa:
@Patowhiz I was creating another table with the exibble data. The stub button works really well. I managed to add the stub column and also group it and it worked fine. So I wanted to put the row group labels within a secondary column within the table stub but I don't know how to do that can you tell me? I tweaked the code below and managed to get what I wanted so I want to know if I can do it in the dialog.

# Dialog: Presentation Table

last_table <- data_book$get_columns_from_data(data_name="exibble", col_names=c("num","char","fctr","date","time","datetime","currency","row","group")) %>% utils::head(100) %>% gt::gt(rowname_col="row", groupname_col="group", row_group_as_column = TRUE)
data_book$add_object(data_name="exibble", object_name="last_table", object_type_label="table", object_format="html", object=last_table)
data_book$get_object_data(data_name="exibble", object_name="last_table", as_file=TRUE)
rm(last_table)

I managed to get this first;

instat_jTzSv74qna

Then I tweaked the code to get this;

instat_iOdtg8hqME

berylwaswa:
@Patowhiz Today I tested the cells and the source notes tabs and these are working well. I was working with the islands data in the datasets package following the examples [here](https://gt.rstudio.com/articles/gt.html).
 
I managed to add source notes and they look good, though there's no option to change the text style and the clear button is not enabled. 

instat_onxJgW85ql

With the cells tab, I was adding footnotes for specific cells and this also worked fine. In this tab also we have no option for the text style.

instat_zapNan49bh

Below is the image of what I produced. I could change the text style for the titles. What I wanted was to change styles for just one or two words in the title but that can't be done here because the option styles seem to take the whole text.

I'd say the sub-dialog is doing well.

instat_0alqFwRedH

berylwaswa:
@Patowhiz I continued testing the sub-dialog. Today I checked the stub and the row tabs.

The stub works well in the contents section I found no problem but then when I went to the styles button I couldn't see any add button. You should add it there or how does the user add the expression?

instat_wZdKXcdEQV

The rows tab also worked so well when grouping the contents. This looks good, see below.

instat_DehAFgDNfo

berylwaswa:
@Patowhiz I was checking on column labels. I added spanners and customized the column labels so that they are more descriptive. I was working with the air-quality data in the datasets package.

I added the spanners and it looked good, see below;

instat_W5VYLiK6Ld

When adding the column labels I noticed the dialog does not produce a table when I add all four labels at once, the output window stays empty. So I added one it worked fine. Then I added the second one so that the column labels table showed two of them and it worked fine. And I had to keep adding one at a time for the table to be produced. The other tabs I've worked with like the footnotes the receiver can take more than one addition at once.

instat_noNQGhiqU2

This is what I wanted to produce,

chrome_nZI2Sk9aJi
 
And below is what I got in R-Instat. Is there a way to make the labels look like the example? Reading in two lines?

instat_PO12AI0MD7

berylwaswa:
@Patowhiz @Vitalis95

I was checking the group summaries in the row tab on the tables sub-dialog and so I changed the format to show three decimal places and it keeps giving an error below when the decimal places are at the default 2 it's working fine.

instat_zytxWLQwqt

I managed to produce tables by first getting the stub and then grouping the rows, and doing the group summaries

msedge_LEzwa3GXyZ


But then I noticed whenever I'd add any option from the sub dialog below it gave an error.

instat_SbLVcra9Qg

berylwaswa:
@Patowhiz @Vitalis95 

I had unstacked some columns and noticed an error so I deleted the data frame and did it again. But when I went to the table dialog, the variables shown were the same ones on the data frame I deleted. I reset the dialog and still, the variables didn't change.
I then clicked to choose a different data frame in the drop-down, then selected the data frame I wanted again,n and this time the variables changed. 

instat_vQRSMK7XFC

 But when I went into the sub-dialog the variables still showed the ones from the deleted data frame.

instat_vdpLzg3INb

I have now had to restart R-Instat for me to do what I needed to be done.

rdstern (dismissed):
@N-thony I am agreeing with @Vitalis95 suggestion that we merge this first, to avoid conflicts.
Could you "do the honours".  Thanks

Patowhiz:
@rdstern @N-thony I have resolved the conflicts. I noticed that there are changes in the presentation dialog and the themes section of the sub-dialog. Please note that these changes originate from the master branch.  

@Vitalis95, to implement what @rdstern suggested, you may need to make significant changes to the design and development approach across all sub-dialog components. I recommend gradually rebuilding all components from scratch and, once complete, removing the table-related components I previously added.

rdstern (approved):
@N-thony I am approving.

berylwaswa:
@Patowhiz @Vitalis95 
I was working on the sza data to create a table that meets the following requirements;

- filter the data to just use the 20°N data and remove the latitude column

- NA values from sza column are to be removed

- reshape the table so that columns of tst (true solar time) contain angles in degrees (from the sza column)

- the gt output table will have the months as row labels in the stub

- missing values will be replaced with an empty string (so that those cells are blank)

- a stubhead label will state what’s inside the stubs (months, at 20°N)

- the table will have a heading decorated with the HTML Black Sun with Rays (&#x2600;) symbol

- to fit the large amount of data in a small area, use some table options to reduce text size and row padding

This is what the original looks like;

chrome_mCy37aoWNQ

And this is what I got in R-Instat;

instat_yGgc27d4o6

This is really good!

PR #9322, opened 2024-12-21
Update one variable summarise

### LINKED ISSUE #9321: Bug in Describe > One Variable > Summarise Dialogue
Using the survey dataset, I tried to summarise all the variables using the Describe > One Variable > Summarise dialogue, I get the following error. This works fine in version 0.8.0, and it seems like this was recently introduced in the merged version?
image

derekagorhom:
Fixes #9321 
when i looked at the script code the default produced and pasted them in the standalone versions, they run just fine. 
so what i did was remove the `clsGtTableROperator` which was used to pass the function from Patrick's Gt tables and used the original `clsSummaryOperator` and added the third parameter to the `clsSummaryOperator` which would call Patrick's Gt tables.
We tried that when doing the Gt tables for two/three variables with Vitalis but instead of adding the parameter to the pivot wider operator, in this dialog we passed it to the clsSummaryOperator.
Thanks

### DIFF
--- a/instat/dlgOneVariableSummarise.vb
+++ b/instat/dlgOneVariableSummarise.vb
@@ -33,7 +33,7 @@ Public Class dlgOneVariableSummarise
 
     Private clsPipeOperator, clsJoiningPipeOperator As New ROperator
     Private clsGetGtTableFunction As New RFunction
-    Private clsGtTableROperator As New ROperator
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@derekagorhom thanks, that's a relief!
@N-thony over to you.

PR #9331, opened 2025-01-06
Tidy code for Prepare > Dataframe > View Dialog

### LINKED ISSUE #9324: Tidy the code for the Prepare > Data Frame > View dialog
@derekagorhom here is the new dialog - I think as done by @Fidel365 recently?

image

If I use the To script, then it is clear it is not following our "good practice", namely a) get data, b) do something, c) present results.

This code currently does a) and b) together.  Here is the current script:

```
# Dialog: View Data

last_table <- utils::tail(n=10, x=data_book$get_columns_from_data(data_name="dodoma", col_names=c("year","month","day","date","month_abbr","doy_366","rain","tmax","tmin","sunh"), force_as_data_frame=TRUE))
data_book$add_object(data_name="dodoma", object_name="last_table", object_type_label="table", object_format="text", object=last_table)
data_book$get_object_data(data_name="dodoma", object_name="last_table", as_file=TRUE)
rm(last_table)
```
Could you please edit, so it confirms to our good practice?
It is currently the same for the html option, so both need to be done.

For the third option, there is just a and b and they are currently together.  This makes for an odd top line in the R viewer, see below.  I hope that may be improved by the change made.

image

I can alternatively get the R viewer from the bottom right-click.  Then it comes with the nice title instead:

image

I am suggesting you, rather than @Fidel365 to make the changes, because you have already done that, for other dialogs.  I hope Fidel will be observing, so that we adhere to the good practice as we improve further dialogs.

derekagorhom:
Fixes #9324 
This PR implements the naming system for the View dialogue
The Naming system works for the Display HTML in Output window and Display in Output window
For the Display in R Viewer Window i get an error which states that `data=survey` is an `unused argument`, so that option only does; 
b) do something, c) present results.
@rdstern , @N-thony this is ready for review.
Thanks

### DIFF
--- a/instat/dlgView.vb
+++ b/instat/dlgView.vb
@@ -158,6 +158,11 @@ Public Class dlgView
 
         clsViewColumnsFunction.SetPackageName("utils")
         clsViewColumnsFunction.SetRCommand("View")
+        clsViewColumnsFunction.SetAssignToOutputObject(strRObjectToAssignTo:="last_table",
+                                      strRObjectTypeLabelToAssignTo:=RObjectTypeLabel.Table,
+                                      strRObjectFormatToAssignTo:=RObjectFormat.Html,
+                                      strRDataFrameNameToAddObjectTo:=ucrSelectorForView.strCurrentDataFrame,
+                                      strObjectName:="last_table")
     End Sub
 
     Private Sub SetRCodeForControls(bReset As Boolean)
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@derekagorhom that looks good.  @Fidel365 have you seen the changes?  It will be good to ensure that, this change in the commands is checked, at least for all dialogs that are corrected, or improved in the future.

I am curious that the change could not be made in the code for the R-viewer:  Here is the picture.  It shows that the top line is simply the contents of the View command.  This remains messy here, in contrast to what must be the code for viewing the whole file on the right-click option.

So I am approving as this is purely cosmetic, but perhaps @N-thony can suggest a further improvement, before merging.

image

rdstern:
@derekagorhom I approved earleir, and now have approved #9334 that sorts out the naming.  You have made changes now that nullify my approval, but without saying what they are?  

So what's new, and how does your last change relate to #9334?

derekagorhom:
@rdstern I was removing the code changes made for the R-viewer made on this PR
Can you review this again

rdstern (approved):
Still seems fine.  @N-thony over to you

PR #9333, opened 2025-01-07
Added missing themes for ggplot2 graphs

### LINKED ISSUE #9332: Add misssing themes for ggplot2 graphs
@N-thony I'm not quite sure why we missed any - maybe they were added after we made the sub-dialog. 
It should be easy for the remaining ones to be found and added, as they seem to come either only from ggplot2 or from ggthemes, and we already use both.

We already have a lot so whoever add them will have to look (and report) carefully.  I noticed, in particular, that we are missing 2 themes from Excel, and that's important!  We want to be able to show we can at least do as well as a spreadsheet.  We have calc already, so should add Excel.

I assume this could be anyone?

Vitalis95:
Fixes #9332 

@rdstern @N-thony , added `theme_excel` and `theme_excel_new` for excel themes, `theme_solarized_2`for the Solarized  palette, `theme_economist_white` which approximates the style of the 'Economist' and theme Settings Functions; `theme_update, theme_replace, theme_get, theme_test` and `reset_theme_settings`

### DIFF
--- a/instat/clsGgplotDefaults.vb
+++ b/instat/clsGgplotDefaults.vb
@@ -67,7 +67,7 @@ Public Class GgplotDefaults
         Get
             Dim strTemp As String()
             'TODO need to make proper functions and set package names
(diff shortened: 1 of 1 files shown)


rdstern:
@Vitalis95 very nice, and seems to work.

I tried the 4 new themes, and also like the reser_theme_settings.  I wasn't sure however, what replace, get, update and test are supposed to do.  They don't seem to do any harm, but don't they need something more added?  Maybe using them is ready for a tweak to the script?

Then I found a bug - not from here, because I trhink it has been there a time.  If tricky to fix, I'll approve, and this can be a new issue.  If using the boxplot in the Climatic > Check Data > Boxplot, then it doesn't work with facets.  I tried with the zambia Eastern data, where there are 5 stations.  In the dialog the facets are not set properly.  Facets is disabled, which is fine, but Station should be set there, to avoid the bug.

rdstern (approved):
@Vitalis95 thanks.
@N-thony over to you.  A simple but really nice improvement.

PR #9338, opened 2025-01-08
Implementing Comment/Uncomment lines in Script Window

### LINKED ISSUE #8568: Add easy way to comment set of lines in Script window
In PR #8551, @rdstern wrote:

I have been commenting out different blocks of code - so putting a # at the start of multiple lines. Is there a way of doing that easily on a block - perhaps via a right-click?

It would be an R-Instat of this:

"The easiest way to create a multi-line comment in RStudio is to highlight the text and press Ctrl + Shift + C. You can just as easily remove the comment by highlighting the text again and pressing Ctrl + Shift + C"?

N-thony:
Fixes #8568
@rdstern, as the title said, have a look. 
You can find this under the insert button

### DIFF
--- a/instat/ucrScript.vb
+++ b/instat/ucrScript.vb
@@ -247,62 +247,6 @@ Public Class ucrScript
     End Sub
 
     ''' <summary>
-    ''' Toggle comments on selected lines.
-    ''' </summary>
-    Private Sub ToggleComments()
-        Dim originalCaretPosition As Integer = clsScriptActive.CurrentPosition
-
-        ' Get the start and end positions of the selected text
-        Dim selectionStart As Integer = clsScriptActive.SelectionStart
-        Dim selectionEnd As Integer = clsScriptActive.SelectionEnd
-
-        ' Get the start and end lines of the selection
-        Dim startLine As Integer = clsScriptActive.LineFromPosition(selectionStart)
-        Dim endLine As Integer = clsScriptActive.LineFromPosition(selectionEnd)
-
-        ' Begin updating text
-        clsScriptActive.BeginUndoAction()
-
-        Try
-            ' Check if all lines are commented or not
-            Dim allCommented As Boolean = True
-            For lineIndex As Integer = startLine To endLine
-                Dim lineText As String = clsScriptActive.Lines(lineIndex).Text.TrimStart()
-                If Not lineText.StartsWith("#") Then
-                    allCommented = False
-                    Exit For
-                End If
-            Next
-
-            ' Toggle comment status for each line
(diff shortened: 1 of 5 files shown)


rdstern:
@N-thony I don't see what's different?

image

On the other hand I have rediscovered the R Help button in the top right-hand corner, which is very nice.

rdstern:
@N-thony I wonder if you got it wrong in an excellent way, and have implemente a much more important feature!  That is to be able to comment, or uncomment a selection easily?  Let me see.  That's a different issue and far more useful!

rdstern:
@N-thony it is what I hoped for, and there is an issue requesting this.  Great.

It needs some small tweaks and then will be a great little feature.

image

a) I think it currently takes a single `# ` and deletes it, and the next symbol - usually space.  It should just delete the "#" and leave the space, or whatever is after it.
b) It should delete multiple "###" - however many there are.  Currently it only recognises a single `#`.  
c) I am happy that when a section has some comment lines and at least one without a `#`.  Then it puts another hash in front of every line.  That's fine to keep.

rdstern (dismissed):
@N-thony that's nice.  I'm approving.  As a trivial detail, could you call then Comment/Uncomment, i.e. without the s.
@Patowhiz another one for you to check and merge.  Thanks

rdstern (approved):
@N-thony thanks
@Patowhiz over to you

instat/ucrScript.vb
```diff
@@ -246,6 +246,62 @@ Public Class ucrScript
         End Using
     End Sub
 
+    ''' <summary>
+    ''' Toggle comments on selected lines.
+    ''' </summary>
+    Private Sub ToggleComments()
```
Patowhiz, line 252: @N-thony is this function used elsewhere? If not, I suggest you place it under the event block that calls it. This will reduce the number of functions and help with maintainability. Should we ever want this functionality elsewhere, we can always put it in a subroutine.

Patowhiz (commented):
@N-thony thanks for this, just a comment below.
@rdstern could you test what happens when you have say 8 out of 10 commented lines and 2 out of 10 uncommented lines. Let the 2 uncommented lines be in between the 8 lines. Then let me know if the output is what you expect. I'm expecting if I click on uncomment and I only have 2 lines uncommented it should still uncomment the 8 or do you think otherwise?

rdstern (approved):
@Patowhiz thanks for the suggestion.  I have tried various combinations now - after the last changes by @N-thony and it seems reasonable now.
So I'm approving again.

Patowhiz (approved):

Issue #9325, opened 2024-12-27
Convert to numeric occasionally doesn't know to activate the popup dialog


rdstern:
@N-thony I am using the efc data from the library - sjmisc, or sjplot package.  

It is labelled data so I make a variable into a factor - e42dep.  That's fine.  Then if I reverse that and make them numeric again, it just makes them all missing.  I was expecting it to activate the popup menu to let me reverse the operation.  (Fortunately your undo works, but I don't know how to return the variable to numeric? 

I note a second problem here with these data with the Prepare > Check Data > View/Delete labels dialog

![image](https://github.com/user-attachments/assets/a8a25201-9983-4dbc-8fd7-9421a8b09cb4)

It works nicely when there is no filter.  Now make a filter, choosing maybe e16sex==1 (so male).  It works on the filtered data, but the results are now different - there are now no labels except for the variables defined as factors.

It would be good if this dialog could work with filters.

derekagorhom:
@rdstern for the second part of this issue, i think that is how the view dialog works. see below the survey data.
![image](https://github.com/user-attachments/assets/0eb9fe27-3417-4822-b969-ae2cbe754f49)

PR #9341, opened 2025-01-09
Fixed Issue with Convert Numeric dialog not popping up

derekagorhom:
Fixes partly #9325 
While going through the issue i realised the frmConvertoNumeric's cmdLabelledConvert alternates between labelled convert and ordinal convert. it seems the ordinal convert option works because the frmConvertoNumeric is able to pop up when the column being convert to numeric is not labelled but does not for the case where column is labelled. This PR should fix that issue (I hope) without causing any regression. 
@rdstern , @N-thony can your review this PR 
Thanks


rdstern (approved):
@derekagorhom excellent detective work and very useful.
You say only partially fixes the issue.  Is that just being cautious - it seems fixed to me?
@N-thony another bug sorted.  Over to you to check and merge.

PR #9343, opened 2025-01-09
Formatting script from preview window in Insert Dialogue

N-thony:
@rdstern, is this what you wanted?
image

### DIFF
--- a/instat/dlgScript.vb
+++ b/instat/dlgScript.vb
@@ -675,8 +675,7 @@ Public Class dlgScript
     End Sub
 
     Private Sub PreviewScript(strNewScript As String)
-        Dim strScript As String = strNewScript & Environment.NewLine
-        txtScript.Text = GetFormattedScript(strScript)
+        txtScript.Text = GetFormattedScript(strNewScript & Environment.NewLine)
         txtScript.SelectionStart = txtScript.Text.Length
         txtScript.ScrollToCaret()
         txtScript.Refresh()
--- a/instat/dlgScript.vb
+++ b/instat/dlgScript.vb
@@ -675,11 +675,28 @@ Public Class dlgScript
     End Sub
 
     Private Sub PreviewScript(strNewScript As String)
-        txtScript.Text = strNewScript & Environment.NewLine
+        Dim strScript As String = strNewScript & Environment.NewLine
+        txtScript.Text = GetFormattedScript(strScript)
         txtScript.SelectionStart = txtScript.Text.Length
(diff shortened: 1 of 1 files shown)


rdstern (dismissed):
@N-thony nice - yes it is
@Patowhiz another small one for you

instat/dlgScript.vb
```diff
@@ -675,12 +675,29 @@ Public Class dlgScript
     End Sub
 
     Private Sub PreviewScript(strNewScript As String)
-        txtScript.Text = strNewScript & Environment.NewLine
+        Dim strScript As String = strNewScript & Environment.NewLine
+        txtScript.Text = GetFormattedScript(strScript)
         txtScript.SelectionStart = txtScript.Text.Length
```
Patowhiz, line 680: I think the 2 lines can be merged into one; `txtScript.Text = GetFormattedScript(strNewScript & Environment.NewLine)`

instat/dlgScript.vb
```diff
@@ -675,12 +675,29 @@ Public Class dlgScript
...
         txtScript.ScrollToCaret()
         txtScript.Refresh()
     End Sub
 
+    Private Function GetFormattedScript(strScript As String) As String
```
Patowhiz, line 685: @N-thony is there a reason as to why you couldn't use the functions used in `frmMain` for getting formatted scripts? It would be good to reuse those functions if possible (to promote DRY principle within the code base) .
I understand we don't use them to display on the script window but check if it's possible.

Patowhiz (commented):
@N-thony thanks for this. A few comments below.

instat/dlgScript.vb
Patowhiz: @N-thony sorry I think I misled you by mentioning `frmMain` which contains an instance of `clsRLink`. Have a look at `GetRunnableCommandLines` function in `clsRLink`. Is it possible to use that?

Patowhiz: @N-thony you’re absolutely correct. I just realized that we are formatting scripts in multiple places, either for checking script contents or displaying them. Ideally, we should consolidate all the code related to script formatting into a single location. However, achieving this will require careful examination and thoughtful implementation.

Thank you for all scripts related tasks.

Patowhiz (approved):
@rdstern the code looks good now. I will proceed with merging once you approve. 

Regarding the scripts produced by the insert dialog, I’m not sure if they’re in the correct order or if it’s just an issue with my stale branch. However, that's a separate matter.

PR #9347, opened 2025-01-10
Corrected the Functionality of Items in the View Menu

### LINKED ISSUE #9268: Hiding Structured and Procurement Menus upon opening
Hello @rdstern, @N-thony and @Patowhiz, 

While I think that it is great that R-Instat offers a  wide range of options. I sometimes think this contributes to the program feeling complicated and inaccessible. As such I want to question whether we should hide more of the infrequently used or currently unfinished menus on opening. 

I question whether we could hide:

- Model (as i understand it this menu is not really ready for use? Perhaps I am wrong? If it is usable then not) 
- Structured (as i understand it this menu is not really ready for use? Perhaps I am wrong? If it is usable then not) 
- Procurement (as i understand it this is a specialized menu  that is not fully developed.)
- Experiments  (as i understand it this is a specialized menu  that is not fully developed.)

This is just an idea. If people feel strongly against, I am interested to hear your thoughts. 
Thanks, 

Rachel

MeSophie:
Fixes #9268
@rdstern @N-thony I corrected the functionality of Structured, Procurement and Experiments items in the View menu. Please have a look.

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -159,9 +159,7 @@ Public Class frmMain
 
         '---------------------------------------
         'toggle the optional form menu items based on set opyions
-        'mnuViewStructuredMenu.Checked = clsInstatOptions.bShowStructuredMenu
         mnuViewClimaticMenu.Checked = clsInstatOptions.bShowClimaticMenu
-        'mnuViewProcurementMenu.Checked = clsInstatOptions.bShowProcurementMenu
         mnuIncludeComments.Checked = clsInstatOptions.bIncludeCommentDefault
         mnuShowRCommand.Checked = clsInstatOptions.bCommandsinOutput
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@MeSophie looks perfect now.  Thanks
@N-thony over to you.
@rachelkg just to report that the default is that we now start with the climatic menu only.  So the initial screen for now will be different.  The user can always choose to view another menu too - or instead.  If so, then what is then viewed will be the starting point next time.

rdstern:
@N-thony I wonder if this is either something you might take-over quickly, or we leave till after the important release of 0.8.3.  

There's quite a lot I would like to see included there, and this is lower on my priority list.  I am concerned that it could be time consuming for them.  If it is also time consuming for you, then I suggest we "park it" for now and return after the next release.

rdstern:
@MeSophie can this be closed?

PR #9342, opened 2025-01-09
Enabling R viewer in the Main menu

Fidel365:
This issue corresponds to PR #9334 
@rdstern you can approve this separately.

PR #9349, opened 2025-01-10
Enabled R viewer in the main menu

Fidel365:
Replaces PR #9342 
@rdstern and @N-thony sorry for being too ambigous on this issue, but I had to make a fresh pr because of conflicts, and now it's ready.


rdstern (commented):
@Fidel365 I like it, except it now comes "behind" the main window.  It is currently the same as the Bottom right-click, but there is stays in front of the main R-Instat windows, which I prefer.  Earleir I think your option did too?

Fidel365:
@rdstern I have made the changes requested, and again this is ready, If you are happy then see you on monday!

rdstern (approved):
@Fidel365 thanks, this seems to work nicely now.
I note that (like the bottom right-click, it then disappears and you need to get it from the taskbar.  There if you right-click it stays on top.  We should find a place to put that into the help somewhere!

@N-thony back to you to check and merge.

PR #9355, opened 2025-01-13
Fixed the problems in the Describe > One Variable > Summarise

### LINKED ISSUE #9354: Fix the problems in the Describe > One Variable > Summarise
@N-thony a relatively minor problem to start with.   But there is also a serious problem, which worked fine in Version 0.7.6 and doesn't work here.  I'll leave you to assess whether this is serious enough to delay the release on Monday?

a) I have usually tried the dialog with the small survey data.  This problem was clearly introduced with the recent improvements: Here is the dialog:

image

It is with all the variables and works fine the first time:  The commands generated are as follows:

```
# Dialog: One Variable Summarise

summary_table <- data_book$summary_table(data_name="survey", data=survey, treat_columns_as_factor=TRUE, margins="summary", columns_to_summarise=c("village","field","size","fert","variety","yield","fertgrp"), summaries=c("summary_count", "summary_count_all", "summary_sum"))
last_table <- (summary_table %>% pivot_wider(names_from=summary, values_from=value) %>% gt::gt())
rm(list=c("last_table", "summary_table", "survey"))
```

Now return to the dialog, perhaps press Reset and do the same as above.  Then the layout has changed, and there are no summary variables across.  You can see why, because the commands generated are now as follows:

image

You can see the difference on line 6 above.  It no longer does the pivot wider element.

b) That's the easy one to fix.  The other is more serious and may be easier to investigate once a) is fixed.

When I do the same with the diamonds data, it is so slow it has never finished for me.  The problem is I think in line 5 of the above code.  It has probably been there for quite a long time.  It seems similar in Version 0.8, but it all works fine in version 0.7.6 from 2 years ago.  So, something has regressed, and I'm both relieved and disturbed by this!

Here is the dialog in Version 0.7.6 with all variables and count, countall and mean as the summaries:
image

Here are the result, from version 0.7.6, fairly instantaneously:

image

And here are the commands generated in Version 0.7.6:

```
# Dialog: One Variable Summarise

summary_table <- data_book$summary_table(data_name="diamonds", treat_columns_as_factor=TRUE, margins="summary", columns_to_summarise=c("carat","cut","color","clarity","depth","table","price","x","y","z"), summaries=c("summary_mean", "summary_count", "summary_count_non_missing"), j=1)
mmtable2::mmtable(data=summary_table, cells=value) + mmtable2::header_left_top(variable='variable') + mmtable2::header_top_left(variable='summary')
rm(summary_table)

```

Vitalis95:
Fixes #9354 
@rdstern @N-thony ,please check  both a) and b) of the issue

### DIFF
--- a/instat/dlgOneVariableSummarise.vb
+++ b/instat/dlgOneVariableSummarise.vb
@@ -192,7 +192,6 @@ Public Class dlgOneVariableSummarise
                                                    strObjectName:="last_summary")
 
         clsSummaryTableFunction.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$summary_table")
-        clsSummaryTableFunction.AddParameter("data", clsRFunctionParameter:=ucrSelectorOneVarSummarise.ucrAvailableDataFrames.clsCurrDataFrame, iPosition:=0)
         clsSummaryTableFunction.AddParameter("treat_columns_as_factor", "TRUE", iPosition:=1)
         clsSummaryTableFunction.AddParameter("margins", Chr(34) & "summary" & Chr(34), iPosition:=2)
         clsSummaryTableFunction.AddParameter("summaries", clsRFunctionParameter:=clsSummariesList, iPosition:=5)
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@Vitalis95 excellent and thanks.  Please can you add an explanation of what you did and why if possible?  How did these happen?  Is there anything we can learn from trhem?
@N-thony please check and merge when ready.  We may be back on track for the version today!

PR #9360, opened 2025-01-15
Fixed bug in Append dialog

### LINKED ISSUE #9352: Small oddity in adding data frames in Append dialog
@N-thony Here it is:

In the Prepare > Reshape > Append I might have this:

image

I can keep adding and that's fine.  But if I Close (or press Ok) and then Reopen and add I get this:

image

Now I have checked briefy and think this may be the only dialog with this problem.  If I have a multiple receiver and add more columns it seems fine.  There are not many dialogs with multiple data frames and the other I found clear the receive on re-opening - and I am happy with that there.  But I'd prefer not to clear on re-opening in Append.

I hope it will be trivial to fix.

derekagorhom:
Fixes #9352 
The fix ensures that the dialog on reopening clears the receiver of the previously selected dataframes.
@rdstern , @N-thony this is ready for review

### DIFF
--- a/instat/dlgAppend.vb
+++ b/instat/dlgAppend.vb
@@ -36,6 +36,7 @@ Public Class dlgAppend
         End If
         SetRCodeForControls(bReset)
         SetHelpOptions()
+        ReopenDialog()
         bReset = False
         autoTranslate(Me)
     End Sub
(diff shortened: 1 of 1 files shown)


rdstern (changes requested):
@derekagorhom these don't seem to be your finest corrections.
The problem has gone, but because you are clearing the multiple receiver.  When I return to the dialog I would still like the receiver to include the sheets I appended before.

derekagorhom:
Closing this PR because it is being fixed in #9410

PR #9369, opened 2025-01-17
Bug fix in Levels/Labels dialogue when filter is applied

### LINKED ISSUE #9263: The levels/labels dialog should respect filters
@N-thony I hope this will be easy, either for you, or to allocate?  I show here the script from the One Variable Describe, which does take account of the filter, and levels/labels that does not.

image

a) I wonder if adding the first line from the Describe One Variable script is all that is needed for a start.
b) However, I note that the script generated doesn't produce any results when run as a script.  Could it be altered so it can be run as a script?  Or could we give instructions on what to do to get this display from a script?

N-thony:
Fixes #9263
@rdstern have a look.


rdstern (approved):
@N-thony great.  I use this dialog a lot in workshops, and it has been an anomaly that it didn't respect filters.
@Patowhiz over to you.

Patowhiz (approved):

PR #9371, opened 2025-01-17
Additional button in the Insert > Save Results Tab

### LINKED ISSUE #9329: Add another button in the Insert > Save Results Tab
@N-thony this is a small task to allocate.

Here is the current tab in the Insert dialog:

image

a) Could you change **_From Data Frame_**  to **_Into Data Frame._**
b) Move that option down and add an option at the top that says **_From Variables_**
In that option there are 2 fields to complete, namely 
Name(s) [           ] (separated by commas)
Data Frame [         ]
If a,b,c,d is the column names and dat is the data frame name, then it generates 3 lines
# Save variables "a,b,c,d" into data frame "dat"
dat <- data.frame( a,b,c,d )
data_book$import_data(data_tables=list(dat=dat))

Note Ok is still enabled if the name field is blank.  Then the 2nd line is just dat <- data.frame(  ).  In that case the names can be added within the script, which will sometimes be easier.

derekagorhom:
Fixes #9329
I have added the `From Variables` options with the two `fields` to complete
@rdstern , @N-thony 
This is ready for review

### DIFF
--- a/instat/dlgScript.Designer.vb
+++ b/instat/dlgScript.Designer.vb
@@ -24,21 +24,38 @@ Partial Class dlgScript
     Private Sub InitializeComponent()
         Me.tbFeatures = New System.Windows.Forms.TabControl()
         Me.tbPageSaveData = New System.Windows.Forms.TabPage()
+        Me.ucrInputSaveRFile = New instat.ucrInputTextBox()
         Me.rdoFromRFile = New System.Windows.Forms.RadioButton()
         Me.rdoDataFrame = New System.Windows.Forms.RadioButton()
+        Me.ucrChkDisplayGraph = New instat.ucrCheck()
+        Me.ucrChkSaveDataFrameSingle = New instat.ucrCheck()
+        Me.ucrInputSaveDataFrame = New instat.ucrInputTextBox()
         Me.lblSaveObjectFormat = New System.Windows.Forms.Label()
+        Me.ucrCboSaveOutputObjectFormat = New instat.ucrInputComboBox()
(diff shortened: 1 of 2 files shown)


rdstern (changes requested):
@derekagorhom great - it already works!  So the suggestions are just cosmetic.
Here is the current dialog:
image

a) It should be From Variables: (so add the s)
b) Put the Name(s) and Data Frames on their own lines.  I don't mind if the dialog is a bit longer - and can be longer for all options if needed.
c) Similarly put the From R file, below the Single checkbox, so the 3 buttons are each below and lined up.
d) In the comment that is generated, add a space before the into.
Thanks

derekagorhom:
@rdstern I have made the changes, can you review this again
Thanks

rdstern (approved):
@derekagorhom great - that looks fine now
@N-thony over to you

rdstern (changes requested):
@derekagorhom I approved earleir, but please first add the s to Variable:, so it becomes `Variables:`.

derekagorhom:
@rdstern I have made the change

rdstern (approved):

rdstern:
@N-thony I am re-opening this pull request. I find there are 2 small problems.  They should be easy to fix.

a) In the screen below the Variables section was added at the top:

image

The 3rd option, From R file needs to move down a bit.  The top is cut off.

b) The controls from the Variables section also appear on the two other top buttons.  One is shown below:

image

derekagorhom:
@rdstern PR #9433 fixes the new problems you found.

PR #9381, opened 2025-01-22
Added Outfilling dialog

### LINKED ISSUE #9363: Outfilling Dialog
@N-thony  this is mainly for @Vitalis95. 

As part of our ePICSA work, we would like to construct a dialog to simplify the outfilling of station data using satellite estimates mainly for our current work in Zambia.

Below is the R script which we recently used in Zambia perform some outfilling. The outfilling uses @lilyclements  `outfillingR` package. 
`library(outfillingR)
zambia_data <- data_book$get_data_frame(data_name="zambia_data")
#perform outfilling
infilled_data <- do_infilling(
  data = zambia_data,
  station = "station",
  date = "date",
  rainfall = "rainfall",
  rfe = "tamsat",
  lon = "longitude",
  lat = "latitude",
  station_to_exclude = "STATION A",
  rainfall_estimate_column = "tamsat",
  custom_bins = c(1, 5, 10, 15, 20),
  count_filter = 5,
  min_rainy_days_threshold = 30,
  target_months = c(5, 6, 7, 8, 9),
  distribution_flag = "gamma",
  markovflag = TRUE
)
#display resulting dataframe
data_book$import_data(data_tables=list(infilled_data=infilled_data))`

See an initial design of the dialog below - largely based on the arguments of the `do_infilling` function in the script. I think this should be enough to get you @Vitalis95 started.  Let me know any questions you may have. 

Outfilling Dialog

@lilyclements the script currently results in a new dataframe with the outfilled column. Wouldn't it simpler and neat if the outfilled column is just added to the dataframe we are trying to infill? Is this possible?

Vitalis95:
Fixes #9363 
@rdstern @lilyclements @jkmusyoka , please check the progress.

`Omit Months` is not working yet, and I have raised that in the issue.

### DIFF
--- a/instat/frmMain.Designer.vb
+++ b/instat/frmMain.Designer.vb
@@ -2570,7 +2570,6 @@ Partial Class frmMain
         '
         'mnuClimaticModelOutfilling
         '
-        Me.mnuClimaticModelOutfilling.Enabled = False
         Me.mnuClimaticModelOutfilling.Name = "mnuClimaticModelOutfilling"
         Me.mnuClimaticModelOutfilling.Size = New System.Drawing.Size(271, 34)
         Me.mnuClimaticModelOutfilling.Text = "Outfilling..."
--- a/instat/static/InstatObject/R/dlgOutfillingStationData.Designer.vb
+++ b/instat/static/InstatObject/R/dlgOutfillingStationData.Designer.vb
@@ -23,22 +23,346 @@ Partial Class dlgOutfillingStationData
     <System.Diagnostics.DebuggerStepThrough()> _
     Private Sub InitializeComponent()
         Me.ucrBase = New instat.ucrButtons()
+        Me.ucrSelectorOutfilling = New instat.ucrSelectorByDataFrameAddRemove()
+        Me.ucrReceiverStation = New instat.ucrReceiverSingle()
+        Me.ucrReceiverRain = New instat.ucrReceiverSingle()
+        Me.ucrReceiverLongitude = New instat.ucrReceiverSingle()
+        Me.ucrReceiverLatitude = New instat.ucrReceiverSingle()
+        Me.ucrReceiverDate = New instat.ucrReceiverSingle()
+        Me.ucrReceiverEstimates = New instat.ucrReceiverSingle()
+        Me.lblStation = New System.Windows.Forms.Label()
+        Me.lblDate = New System.Windows.Forms.Label()
+        Me.lblLatitude = New System.Windows.Forms.Label()
+        Me.lblLong = New System.Windows.Forms.Label()
+        Me.lblRain = New System.Windows.Forms.Label()
+        Me.lblEstimates = New System.Windows.Forms.Label()
+        Me.ucrInputBins = New instat.ucrInputComboBox()
+        Me.lblBins = New System.Windows.Forms.Label()
+        Me.ucrNudCount = New instat.ucrNud()
+        Me.ucrNudDays = New instat.ucrNud()
+        Me.lblDays = New System.Windows.Forms.Label()
+        Me.lblCount = New System.Windows.Forms.Label()
+        Me.cmdOmitMonths = New System.Windows.Forms.Button()
+        Me.ucrInputDist = New instat.ucrInputComboBox()
+        Me.ucrInputMarkov = New instat.ucrInputComboBox()
+        Me.lblDist = New System.Windows.Forms.Label()
+        Me.lblMarkov = New System.Windows.Forms.Label()
+        Me.ucrChkOmitMonths = New instat.ucrCheck()
         Me.SuspendLayout()
         '
         'ucrBase
         '
         Me.ucrBase.AutoSizeMode = System.Windows.Forms.AutoSizeMode.GrowAndShrink
-        Me.ucrBase.Location = New System.Drawing.Point(16, 439)
+        Me.ucrBase.Location = New System.Drawing.Point(16, 530)
         Me.ucrBase.Margin = New System.Windows.Forms.Padding(4, 4, 4, 4)
         Me.ucrBase.Name = "ucrBase"
         Me.ucrBase.Size = New System.Drawing.Size(615, 78)
         Me.ucrBase.TabIndex = 0
         '
+        'ucrSelectorOutfilling
+        '
+        Me.ucrSelectorOutfilling.AutoSize = True
+        Me.ucrSelectorOutfilling.bDropUnusedFilterLevels = False
+        Me.ucrSelectorOutfilling.bShowHiddenColumns = False
+        Me.ucrSelectorOutfilling.bUseCurrentFilter = True
+        Me.ucrSelectorOutfilling.Location = New System.Drawing.Point(9, 17)
+        Me.ucrSelectorOutfilling.Margin = New System.Windows.Forms.Padding(0)
+        Me.ucrSelectorOutfilling.Name = "ucrSelectorOutfilling"
+        Me.ucrSelectorOutfilling.Size = New System.Drawing.Size(318, 274)
+        Me.ucrSelectorOutfilling.TabIndex = 1
+        '
+        'ucrReceiverStation
+        '
+        Me.ucrReceiverStation.AutoSize = True
+        Me.ucrReceiverStation.frmParent = Me
+        Me.ucrReceiverStation.Location = New System.Drawing.Point(492, 65)
+        Me.ucrReceiverStation.Margin = New System.Windows.Forms.Padding(0)
+        Me.ucrReceiverStation.Name = "ucrReceiverStation"
+        Me.ucrReceiverStation.Selector = Nothing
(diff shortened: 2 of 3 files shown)


rdstern:
@Vitalis95 many thanks.  I show the current dialog here to help @jkmusyoka and @lilyclements to contribute to the review:

image

And very well done.  This looks really nice and neat.  I suggest it is important also for our audience.  I note you already have the automatic filling of the controls, when the data are defined as climatic.  I suggest having the dialog - rather than just the script - is important in our strategy of involving the ZMD staff centrally in this stage of the research on outfilling.  I suggest many staff could contribute well via the dialog route and it would be much harder for them otherwise.

Also:

a)  @jkmusyoka I assume you have a version of the data from Eastern Province that we used for the outfilling.  Can you add a copy here, so we can try it with the dialog?
b)  @lilyclements I assume we do want to use "our" version of the Omit Months, (see the Rainfall QC dialog)? 
 This feature is important to get the testing working ok for Zambia.  I am happy if we decide instead to have a simpler control - used in the current function, which uses the month numbers.  But I am nervous that we may want to use this facility with the shifted data - James what did we do in Zambia in December?  We could alternatively tell users this currently is only for non-shifted data, so easy to give month numbers as Lily does currently? 
c) Lily, what did Emily say about the `Station to Exclude `control.  Before asking Vitalis to improve it, I wonder on the reply as to whether it is needed at all - given we can use filters in R-Instat.  Can we omit that?
d) Vitalis, one omission you could sort out now is that we need a `Store Result` control at the bottom- the same as in the Calculator and other dialogs.  Lily, you are going to make the corresponding change in the function?  (It currently makes a new data frame, despite just producing a single column of the same length.  It will be so much easier to use when it just adds to the existing dataframe.  What should the default name be?  Maybe `outfilled`!
e) While you are doing that could you add a blank at the top of the Omit station control, - currently we have to omit a station.  
Could you also add a numeric version of the Omit Months checkbox, on the left instead of "ours".  If checked this opens a drop-down (same control as for your current bins) with 5,6,7,8,9 as the default and one can type numbers into it.  Have 1, 2, 3, 11, 12 as a second and 1,2,12 as a third. 
And in the drop down for bins, change 5 to 2 in the second option there.
f) Lily do we need a More button towards a sub-dialog that facilitates changing the other arguments, or can that wait?

Vitalis95:
@rdstern , sorted out` e)`. I will add `Store Result` control once the function is changed.

rdstern:
[zambia_data_for outfilling2.zip](https://github.com/user-attachments/files/18519940/zambia_data_for.outfilling2.zip)

@Vitalis95 @lilyclements and @jkmusyoka this is wonderful.  The dialog is working already!

a) I attach the data file for Eastern Province (first data frame) together with the results from 3 runs of the dialog  (Tamsat, chirps and era5) :

image

Amazing it is working immediately.  So full marks to Lily, for the outfillingR package.  I installed it directly from the menu.  (I'm still not clear whether the importing dialog needs the James adjustment - I just use the dialog.)

Then it ran first time!  

Lily I still would far prefer it to produce a new column in the same data frame, rather than a whole new data frame.
And I still think it is important to have an option of a specified random number seed.  The whole thing is comparisons, and this includes comparisons of the different methods - which should use the same sequences.

But should we merge this now, while we wait?

rdstern (commented):
@Vitalis95 and @lilyclements that's wonderful - what I have tested.  It seems to work.

Vitalis, the only trivial point is that the label Stations to Exclude is too long, and is therefore only incomplete.  Could you please change it to `Omit Stations`.  

@N-thony this is a major new facility.  Can you please check so it can be merged?

@lilyclements I tried with the 3 methods and all seem to work on these data.  (tamsat, chirps and era5.)  I also tried era5 twice with the same initial random number seed and got the same answers.  Brilliantly done therefore.

The only result I don't understand is that tamsat outfilling gave missing values when tamsta was missing.  But the rainfall was not missing then.  I thought that it would give the station rainfall except when that was missing.   So now I'm not sure what the results are?

lilyclements:
@Vitalis95 nice. I ran this for my data (attached) with `custom_bins=c(1, 3, 5, 10, 15, 20)` and got an error in the R code. I did not get this error with `custom_bins=c(1, 5, 10, 15, 20)`. Since this is a problem with the R code, I will document it in the outfillingR repo [here](https://github.com/IDEMSInternational/outfillingR/issues/27).

I do not understand the system well enough to know a suitable fix. But essentially what is happening is an `NA` is generated for the SD of rainfall in one month due to small bin sizes, and so it throws an error. Something to discuss with Emily, but @rdstern if you could take a look at the issue in case it is something you can understand then that would be very useful!

rdstern:
@lilyclements did you see Emily's answer to my query above?  Essentially that it might not yet be running the second part of the function?  If that's the case, we should also check whether the current results are useful as part of the checking process. of how to adapt the different methods.  We then leave the data for the main stations intact, which is what we want, when applying them to more stations.

lilyclements:
@rdstern yes, it sounds like something we can go through next week when we meet? I'm not sure if I will be able to come in person due to my ankle but I can certainly join on teams for the afternoon.

Vitalis95:
@rdstern , have a look at it
@N-thony , is it possible to set the prefix for the two columns produced at the same time, eg `tamsat_est`, and `tamsat_out`. ?

lilyclements:
@Vitalis95 how about somehting ike

```
infilled_data <- outfillingR::do_infilling(...)  # run this as you currently do (I'm not sure what you assign it to, but just put `infilled_data` as a hypothetical name for now

new_colnames <- colnames(infilled_data) 
colnames(infilled_data) <- paste0("tamsat", "_", new_colnames)
```

I believe the output data frame is just two columns - our `outfilled_rain` and `generated_rain`.

Vitalis95:
@lilyclements ,  Roger suggested that the two new columns should have the suffixes _est and _out. @rdstern, is that correct? Do we still need the save control?

lilyclements:
@Vitalis95 can you try this:

```
   infilled_data <- outfillingR::do_infilling(...)  # run this as you currently do (I'm not sure what you assign it to, but just put `infilled_data` as a hypothetical name for now
    colnames(infilled_data) <- c(paste0(rainfall_estimate_column, "_gen"), paste0(rainfall_estimate_column, "_out"))
```

Where `rainfall_estimate_column` is the variable name for the variable in the receiver for the rainfall estimate
And `generated_weather` is the name of the `do_outfilling` script.

Can the save control still do a prefix check to it - so we add a prefix to it if that column already exists?

lilyclements:
@rdstern I am not getting a corrupt error when downloading it. Can you try removing it completely by running this:

```
remove.packages("outfillingR")
```

Then try downloading it again?

rdstern:
@Vitalis95 and @lilyclements I followed Lily's instruction above and it is now working!
I don't quite understand why - but that's not particularly important.  
On this topic, if that is something that we may want to do in the future, then I wonder if:
a) We add a third button to the Tools > Install R-Package.  Maybe called Remove?
b) The command `remove.packages("outfillingR")`. Gave zero output.  It does give an error if I mis-type the name, but nicer if it said if it works?

Ok , but to the results.  Here is a snapshot:

image

It looks sensible.  I did 2 runs,  tamsat is the tamsat original data.  tamsat_ and tamsat_2 are the 2 estimates we had last time, while tamsat_1 and tamsat_21 are the outfilled variables.  

Lily and Vitalis I thinking that early on we probably just want tamsat_ and tamsat_2 when we are checking which method to use.  Then we don't need to outfill.  Later, once it is working well we need the outfilled data, but maybe not the checking variables?
If you agree we should have two individual storing variables, each with a checkbox, and default checked.  The checkbox labels could be Estimated and Outfilled and the default names maybe tam_est and tam_out.  (what's the better word than estimated?) So with the first 3 letters from the Estimates receiver.

Vitalis95:
@rdstern , have a look at it

rdstern:
@Vitalis95 it is looking very good.
Could you swap the filled names round please.  So chirps.out goes with outfilled, and gen with generated!

That's all for you, but one for @lilyclements as I hit the error she reported.  

image

It is a nice message, but reminded me that you and Emily wqere going to tweak the code so the bin limits and somethoing else, were unneceassary?

When you do, then I wonder about bin sizes?  I guess the more data you have the more bins you could allow and I assume (up to a limit) the more the merrier?  And if the contents will be roughly equal then could we choose what the value will be, say between roughly 10% so 10 bins and maybe 20%?  And maybe the top bin has half the others, because we need data there, but it is the most important?
Trying to tempt your discussion with Emily!

lilyclements:
@rdstern thanks for the reminder. I have just messaged Emily about it and we will hopefully meet next week

rdstern:
@lilyclements I have also added to the comment above suggesting another task for you!  That's for the function to add a report of the settings for each run into the output window!  I hope that's easy.

Vitalis95:
@rdstern , if you want the first factor by default, then  we make the first label by default be selected, not all selected as it is now.

PR #9382, opened 2025-01-22
Removing Scalars from non-Scalar Calculations

### LINKED ISSUE #9376: Check if we always need scalars in each calculation?
@N-thony I am now using R-Instat with R commands visible.  Here is a simple calculation:

Image

I notice it always has 2 special lines/commands for the scalars.  Do we always need them.  Here I am using a data frame with no scalars, so could that be checked first?  Or could they become part of the data in the first line?

derekagorhom:
Fixes #9376 
This PR should prevent the issue of the scalars function appearing when the user is not using the scalar option.
@rdstern , @N-thony can you review this code
Thanks

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -122,10 +122,8 @@ Public Class dlgCalculator
         clsDetachScalarsFunction.AddParameter("unload", "TRUE")
 
         ucrBase.clsRsyntax.AddToBeforeCodes(clsAttachFunction, 0)
-        'ucrBase.clsRsyntax.AddToBeforeCodes(clsAttachScalarsFunction, 1)
 
         ucrBase.clsRsyntax.AddToAfterCodes(clsDetachFunction, 1)
-        'ucrBase.clsRsyntax.AddToAfterCodes(clsDetachScalarsFunction, 2)
 
         ucrBase.clsRsyntax.SetCommandString("")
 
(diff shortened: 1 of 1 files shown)


rdstern (changes requested):
@derekagorhom it is only partly working well.
Please can you get a scalar and the try to use it.  I used the survey and saved mean(yield) into mean_yield as a scalar.  Then tried a calculation of yield * mean_yield.
It says it doesn't recognise mean_yield.  It used to do that fine.

derekagorhom:
@rdstern I have made the fix which prevented the check scalar options from producing results
Can you have a look at it again
Thanks

rdstern:
@derekagorhom you have solved that problem, but seem to have created some more.  You may need help, because it looks to be rather a mess now:
And I also wonder what testing you did, before my tests?

image

Here is what I did:
a) With the survey data I tried yield * 2 .  That was to check there was no longer a use of the scalar lines in the output window,  That seems fine.
b) Next I used, **_Clear_**, then the **_summary keyboard_**, got **_mean(yield)_** and saved it as **_a scalar_** called mean_yield.  That seems almost fine, except I got a bonus of the result also in the output window.  I am not sure whether that was happening before?
c) Then I used Clear, and got yield * mean_yield (pressed the scalar button for that), and saved it into a column, namely calc1.  I worked but also gave the mess.
d) I now see it has also saved a copy of calc1 (a vector) as though it is another scalar.  So now I have the columns I want, plus this bonus that I really, really don't want.

PR #9386, opened 2025-01-23
Skipping enable disable undo script in the output window

### LINKED ISSUE #9385: Minor issue with undo script in the output window
@N-thony when you are checking, there is one small item that maybe for you to change.  Here is the output window after running the dialog:
image

Now I like the fact there is a brief report in the output window, when the main results are the 2 new data frames.  But there are now also 2 lines at the bottom about undo, and I wonder if those are needed?  They seem to me to be a distraction.  If they are needed, then I have no problem with them coming in the log window, but do they need also to be here?

Thanks

_Originally posted by @rdstern in https://github.com/IDEMSInternational/R-Instat/pull/9372#pullrequestreview-2567932972_

N-thony:
Fixes #9385
@rdstern have a look.
This issue was reported in PR #9372
image

### DIFF
--- a/instat/Model/DataFrame/clsDataFramePage.vb
+++ b/instat/Model/DataFrame/clsDataFramePage.vb
@@ -236,7 +236,7 @@ Public Class clsDataFramePage
 
         Dim strDisable As String = If(bDisable, "TRUE", "FALSE")
         clsEnableDisableUndoRFunction.AddParameter("disable_undo", strDisable)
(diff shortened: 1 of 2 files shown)


rdstern (approved):
@N-thony great - that looked like a simple change.  Thanks
@Patowhiz I think this is a very small task to check and merge.  Thanks

Patowhiz (approved):

PR #9389, opened 2025-01-24
Fix Bug in Define Climatic Loading and Address Usability Inconvenience

### LINKED ISSUE #9388: A bug loading the Define Climatic and also a possible inconvenience.
@N-thony Here is the likely bug:

Image

Let me explain the route.
a) I loaded Moorings from the library.  Added variables with the Climatic > Dates > Used Date and then defined it as climatic.
b) I then finished with Moorings.  So I used the File > Close Data Book and loaded the Zambia Eastern tidy data.  It is already defined as climatic, as you see here.  But I wanted to re-define it, so I used the Define Climatic again.

This gave the error, shown above, on loading the Define Climatic.  It still expected Moorings.  Worse than this, it gives me the same error many times.  I still don't know how many. Now I do, about 130 times!!

This ties with the issue I was going to raise anyway.  When I use this dialog for the first time it takes about 10 seconds on my fast machine to show the dialog.  The loading code must be poor.  So could this be examined at the same time the bug.

I record here also that I don't know why the Zambia data - which is still defined as climatic, has lost the key variables.  They are produced by this dialog.  I wonder if they are not well saved.  I will explore, and probably make a second issue.

N-thony:
Fixes #9388
@rdstern, I was not able to reproduce the error you have reported in the issue but I have tried to improve here the time when loading the dialogue.

### DIFF
--- a/instat/DlgDefineClimaticData.vb
+++ b/instat/DlgDefineClimaticData.vb
@@ -32,8 +32,6 @@ Public Class DlgDefineClimaticData
     Private bIsUnique As Boolean = True
 
     Private Sub DlgDefineClimaticData_Load(sender As Object, e As EventArgs) Handles MyBase.Load
-        Dim stopwatch As New Stopwatch()
-        stopwatch.Start()
         autoTranslate(Me)
         If bFirstLoad Then
             InitialiseDialog()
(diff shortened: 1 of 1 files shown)


rdstern:
@N-thony I think it is quicker, but the error message remains.  It is the second time I go into the dialog, and it is still looking for the previous data frame.

image

I found it by 
a) starting with the Zambia Eastern data 
b) and going into the Define climatic.  Accepting and pressing Ok.
c) Then I used file Close data book
d) Then Opened Moorings
e) Then used Dates > Use Date and added the day of year, and day of month.
f) Then Define Climatic gives the error above.  It is still expecting the Zambia data.

rdstern (approved):
@N-thony good to have found that as well as improving the code.  Thanks
@Patowhiz over to you to check and hopefully merge

rdstern:
@Patowhiz could you please check and merge

Patowhiz (approved):

PR #9394, opened 2025-01-28
Some Change Made on Length Of Season Dialog 

### LINKED ISSUE #9391: Small change in labels in Climatic > Prepare > Length dialog
Here is the current dialog:

Image

We now seem to have fixed on the designation of Status as the logical variables for the start and end.

So please add Status to the relevant labels.  So Start (L) becomes `Start Status (L) ` and similarly End (L) becomes `End Status (L)`

MeSophie:
Fixes #9391
@rdstern @N-thony I have added the length_more variable (In accordance with the discussion in PR #9157 ) and corrected some labels on the Length Of Season Dialog. Please have a look.
image


rdstern (approved):
@MeSophie that's good and I am approving.
@N-thony could you check and merge if ok?
Meantime @MeSophie can you make a new issue, for later to replace the up_down control with our usual Day Range control, but call the button Day Number here.  Then there is the usual sub-dialog, (maybe) where you disable everything except the To with Fixed Day.  The receiver on the main dialog can perhaps be less narrow, because it is only giving the number.

Issue #9351, opened 2025-01-11
Small inconvenience to fix for wide data and the viewing of the column metadata


rdstern:
@N-thony or @Patowhiz I go into a wide dataset as shown below.  (I used CarcassoneHeat, from the extRemes package.) I then pressed the option to open a script/log window.  Lo and behold I am asked about whether I really want to get the column metadata.

The trigger for the question seems to be asking for another window.  Please could it be limited to the metadata windows and not the script window.  That's confusing!

![image](https://github.com/user-attachments/assets/656104f0-ca58-4304-86bd-d7513c337a38)

It is correct in opening the log/script window, rather than the blank metadata window.  I say No, because I certianly don't want it now, then I've lost the chance to get it later, because it doesn't seem to ask again.

Patowhiz:
@rdstern thank you for spotting the bug, I agree that it should be limited to the column metadata window.
@N-thony, this issue can be delegated and fixed.

rdstern:
@N-thony when I click on Log/Script I certainly was not expecting to get the blank Column Metadata window.  Getting that is all to do with the error I hope can be rectified.  I hope I have now explained more clearly in your post just above.

rdstern:
@N-thony I now have another bug, reported separately, but it all started when @Patowhiz was sorting out the wide datasets, so I think it may be related - not sure.
Here is how to get it.  
a) Go into R-Instat and get the tutorial_after data from the R-Instat side of Import from Library.
b) Go to View and choose Swap Data and Metadata.

Here is the result:

![Uploading image.png…]()

You now see the non-existent sheet1 in all its glory.  You seem to have lost the real data frames, though the bottom left shows that R-Instat still thinks it is showing diamonds!

berylwaswa:
@ARMSTRONGOPONDO please check if this is still an issue, thanks.

ARMSTRONGOPONDO:
@berylwaswa I have looked through the issue and it seems to be all addressed ill close it.

PR #9396, opened 2025-01-29
Fix bug: Additional Empty Sheet Issue in Column Metadata

N-thony:
@rdstern, I don't remember the issue number. But this fixes the issue when opening the column metadata window it comes with an additional empty sheet named `Sheet1`
image


rdstern:
It is certainly mentioned in issue #9351.  Let me check.

rdstern:
@N-thony I think some bits are fixed, but I still get this!

image

a) It has still added the Sheet1.
b) It shouldn't be here anyway, because I opened the file - first one from extRemes package, and then pressed to go to the Log/Script - why is it asking about metadata then?

rdstern:
@N-thony that's much better.  It now works fine when I go into the log/script window.
Then I tried getting the column metadata, and the extra Sheet1 is still there.  However, I then clicked Yes, and when the metadata finally appeared, I was happy to see that Sheet1 is no longer there.  So that's no longer a particular problem.  (So far I could call that a feature!

Let me test some more options.
a) I tried adding the data frame meta-data window, instead of the log-script window.  That worked fine and also did not ask anything about the column metadata - which is good.  It shouldn't!
b) Then I tried switching the data and column metadata.  Now it should have asked me, but instead here is the result:

image

I also noticed an oddity in the view menu, that I never noticed before:

The Column Metadata has now moved to the top of the View menu.  I didn't even know items could move, in the menus, and would prefer they didn't.

Then I tried volunteering for the column metadata.  Now I correctly got the warning message and accepted.  This gave another issue, that may be easy to fix, or may be done separately.  Namely it does take a long time, which I understand - but it doesn't show the warning screen that it is still working away.  So it looks as though R-Instat has frozen.  Could the message be added?

When it loaded the metadata I could make the switch in windows, and all was fine, and quick.  The order of items in the metdata was still changed.  But no hint now of Sheet1
image

It is certinly better than before.  Will you attack some of the above, or should I approve and the other bits come later?

rdstern:
@N-thony should I approve so the current improvements get into this version?

rdstern (dismissed):
Approving in the hope it can be included in the forthcoming release.
@Patowhiz please do a final check and merge

Patowhiz (approved):

Patowhiz:
@N-thony will merge this after  the conflicts are resolved.

rdstern (approved):
@N-thony I think it is alright now:  I still get this:

image

But after saying yes, the metadata appears without that extra sheet.  I think it is ok now.
@Patowhiz back to you to merge.

Patowhiz:
@rdstern do you want this to be merged as it is or you want @N-thony to fix the message issue? Is it an issue on this PR only or also in the master?

rdstern:
@Patowhiz I am ok for it to me merged as it is.

Patowhiz (approved):
Approving as requested by @rdstern

PR #9403, opened 2025-01-30
Added Loops to R-Instat

Vitalis95:
Still in progress

### DIFF
--- a/instat/dlgScript.Designer.vb
+++ b/instat/dlgScript.Designer.vb
@@ -95,11 +95,16 @@ Partial Class dlgScript
         Me.lblPreview = New System.Windows.Forms.Label()
         Me.txtScript = New System.Windows.Forms.TextBox()
         Me.ucrBase = New instat.ucrButtons()
+        Me.tbPageRepeat = New System.Windows.Forms.TabPage()
+        Me.lblSelectColumns = New System.Windows.Forms.Label()
+        Me.ucrReceiverRank = New instat.ucrReceiverSingle()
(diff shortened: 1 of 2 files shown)


rdstern:
@Vitalis95 great, it really seems to work, and I like the double-use of the data selector.  (First for the select, and second for the column) I didn't realise that was possible!
There is a dangerous flaky bit if I try to change the contents of the column.  It goes out of R-Instat and puts the code back into Visual Studio!  See below:

image

I did that because the code updates when I complete the column component, but not when I add the statement command.

Wow - it is all there and really works!

Vitalis95:
@rdstern , have a look at it

rdstern (commented):
@Vitalis95 that's great.  You and probably also with the support from @N-thony are working really well now.
Only 2 minor points.
a) Reset does not clear the controls.  I am realising that this is an "odd" dialog.  It is sort of a main dialog, with the usual controls at the bottom.  But I think it is then the only main dialog that is Tabbed.  We have Tabbed sub-dialogs of course (e.g. for graphs), but they don't have a reset button!
I think ideally, pressing Reset could usefully Clear the current Tab and remain on the tab.  Is this easy to do?
b) Great you have added a start of a comment.  Could it maybe say: 
```
# Get the variables in the chosen select.
# Then do the calculation for each variable in turn, 
# So produce a new column for each variable in the select.  
```

rdstern (approved):
@Vitalis95 good work and an excellent example to the others of an efficient way to complete a difficult task.
@N-thony could you please check and merge?  Thanks

PR #9336, opened 2025-01-07
Adding Titles and Themes to the Presentation Tables dialog

Fidel365:
Fixes partially #9323 
I still got to add the functonalities for title and footnote, will be ready in the morning.

PR #9413, opened 2025-02-03
Added title and themes to the presentation table dialog

### LINKED ISSUE #9323: Add title and themes to the presentation table dialog
@N-thony this should be easy to allocate, possibly either @derekagorhom or @MeSophie?

It is the `Describe > Tables > Presentation `dialog.

1) Earlier @Patowhiz had the title (and footnote) on the main dialog.  Could it be returned?  (Just the main title, with footnote, and exactly as it is on the sub-dialog.  Not the sub-title.
2) Also the `Select Theme` from the Themes sub-dialog.  It should be a checkbox, default unchecked, rather than a radio button.  @Patowhiz would a checkbox also be ok - maybe even preferred also on the sub-dialog?  Though there we would have to code that you don't tick both Select and Manual.

derekagorhom:
Fixes #9323 
Replaces #9336
I have added title, footnote and themes to the presentation table dialog
I also disabled the style button from @Fidel365 original PR. 
@Fidel365 can you check and confirm this is what Roger asked you to do.
Thanks


Fidel365:
@derekagorhom 
1. The changes should sync with the subdialog and vice versa, have a look at my pr on ow the selet theme checkbox works
2. The select theme checkbox and manual checkbox should work separately, i.e when i uncheck the select theme it's not supposed to directly select manual and vice versa, have a look at my latest changes in pr #9336 
that's all, then I can close mine, also @rdstern  can add to this.

rdstern (commented):
@derekagorhom and @Fidel365 it is coming on.  However I didn't think it was as hard as you are making it.  There is more coming on the tables, so I hope you can finish this soon and then progress.
I am including @berylwaswa on this message as she will also be testing this pull request.  She has been asking about how titles and footnotes can be added to tables, so this will be a nice answer for her.

Here is the dialog using the illness data that comes with the gt package.

image

I have now changed to the Excel theme and tried again.  Here is the result

image

Now there are - at this stage just 2 problems left:
a) Why is the Styles button not enabled on the main dialog.
b) Why is the footnote not "connected" to the title?

When I look at the sub-dialog, the answer to b) is becoming clearer:

image

You can see here that the title line is also "magically" also in the sub-dialog, but the footer remains blank.

@derekagorhom I hope this helps in finalising this improvement, and @berylwaswa I hope this also helps you to see the sort of thing to look out for, when you are testing.

rdstern:
@derekagorhom that's great.  The footnote works and is also in the sub-dialog.
As well as activating the Styles - item a) above,  There is also:

c) I used Theme Excel which works fine.  But could it also please say the same thing in the Themes sub-dialog tab?

berylwaswa:
@rdstern , @derekagorhom 

I managed to add the title and the footnote to the first example with the illness data. I don't know if there is a way to make the word test bold as it is in the one produced by the code.

@rdstern it works fine when I just select to show ten lines so we don't have to filter anymore. I guess that means we also don't have to create the row numbers column.

@derekagorhom see what I got below.

instat_WoU8JJPIpn

rdstern:
@derekagorhom I saw more examples from gt and they all use 10 rows to demonstrate the package.  So could you please also change the 100 to 10, as the new default in the Presentation dialog?  Thanks

derekagorhom:
@berylwaswa I have enabled item a and changes the parameter value from 100 to 10
@rdstern , for item c, I think drop downs do not usually work the way textbox or checkbox do in relation to main dialogs and sub-dialogs (the implemented legend options in the main dialogs does not show in its sub-dialog).
So although on the main dialog the user does something with a drop down, in its sub-dialog the inputs do not show there.

Fidel365:
@derekagorhom for part c, look at my pr #9420

derekagorhom:
> @derekagorhom for part c, look at my pr #9420

I have sent you a skype message

PR #9414, opened 2025-02-04
Improvements to Make Date dialog 

### LINKED ISSUE #9399: Oddity in the Make Date dialog - numeric variables not shown?
@N-thony I have an excellent dataset to illustrate tidying in R-Instat.

I wanted to tidy the columns of dates and also those with lat and long - and all in R-Instat.  I have succeeded, though it will be slightly easier when this bug is resolved and also when the Prepare > Column: Text > Find/Replace is further improved!

The dates come from Excel and illustrate a tricky problem.  Just over half the dates import as numeric and the others (407 rows) look like dates.  On stack-overflow it says to sort this out in Excel first - so before importing.  I tried and failed, so decided to try in R-Instat instead.  This was the only real bug:

Image

With this option in the dialog, it only shows numeric variables.  Well some are numeric, but it isn't showing any variables!  Why not?

Once allocated I will send the data directly.  I hope we can add it to the library, but don't want to assume that yet, so don't want to put it on github.

Thanks

Vitalis95:
Fixes #9399 
@rdstern @N-thony , allowed all variables and added convert numeric function


rdstern (approved):
@Vitalis95 looks fine.  Thanks
@N-thony over to you

PR #9420, opened 2025-02-05
Syncing titles and themes in presentation table 

Fidel365:
Fixes partially #9323,
@rdstern , @derekagorhom , @berylwaswa I made this mainly for syncing the data between the dialog and sub dialog all should work except for the style button.


Fidel365:
@rdstern currently throws an exception , but it works, I'm working around it.

PR #9425, opened 2025-02-06
Fixed Minor bug in Edit > Find Dialog

### LINKED ISSUE #9422: Minor change needed in Edit > Find dialog.
@N-thony the default is to not use regex.  But I think it does currently:

Image

I had a variable where I thought there might be just one row where there was a dot.  When I search for dot, I get every non-missing row.  When I search for `\\.` it works fine.  So the default seems to be assuming regex.

I wondered whether it was coded "the wrong way round"? So I tried with regex ticked.  It seems it always uses regex?

derekagorhom:
Fixes #9422
This should fix the problem with the find needing to use regex to find items in a cell/Row
Thanks

### DIFF
--- a/instat/dlgFindInVariableOrFilter.vb
+++ b/instat/dlgFindInVariableOrFilter.vb
@@ -217,10 +217,16 @@ Public Class dlgFindInVariableOrFilter
                 Dim bApplyToRows As Boolean = (rdoVariable.Checked AndAlso rdoRow.Checked) OrElse rdoInFilter.Checked
                 frmMain.ucrDataViewer.SearchRowInGrid(rowNumbers:=lstRowNumbers, strColumn:=strColumn,
                                                        iRow:=iRow, bApplyToRows:=bApplyToRows)
+                lblFoundRow.Visible = True
                 lblFoundRow.Text = "Found Row: " & iRow
                 lblFoundRow.Visible = True
                 iCountRowClick += 1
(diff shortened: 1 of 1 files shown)


rdstern (commented):
@derekagorhom I am liking my example with these data:
[Zambia_stations_2015.xlsx](https://github.com/user-attachments/files/18693221/Zambia_stations_2015.xlsx)

Here I can look for dot in Longitude.
I found you have fixed it - well except for one small item below:  When I use regex and search for `\\.` it works fine until there are no more.  Then it give me this:

image

I can live with if it is tricky to find (the large negative number).  But I don't get the nsame problem looking for just dot, without regex.  So if easy to find and fix?

And I thought I had written a second small problem, but clearly not!  I don't know how to clear if I would like to, i.e. to get rid of the green line(s) or cells?  I wonder if it would be easy for Reset to also clear any green?  Maybe a @N-thony question?

derekagorhom:
@rdstern , @N-thony has added the option to remove the colour background
Also i made sure the sentinel values no do appear on the display
can you review this 
thanks

rdstern (approved):
@derekagorhom and @N-thony thanks, that's great.
@rachelkg I tested with our new station data that I hope can go into the library.  I looked for a dot in the Longitude first , both with and without regex.  I suggest these data and also this dialog is perhaps ok for introducing regex in the help.
@N-thony can you check and merge please.

PR #9431, opened 2025-02-10
Added title and themes to the presentation table dialog

### LINKED ISSUE #9323: Add title and themes to the presentation table dialog

derekagorhom:
Fixes #9323
This contains the work done on the previous PRs as well as the corrections requested by @N-thony 
@rdstern , @N-thony 
This PR is ready for review
Thanks


Fidel365:
@rdstern I wonder if this is ok with you.

rdstern (approved):
@derekagorhom thanks for the reminder.  It looks very good.

berylwaswa:
@Patowhiz @Vitalis95 

I was checking the group summaries in the row tab on the tables sub-dialog and so I changed the format to show three decimal places and it keeps giving an error below when the decimal places are at the default 2 it's working fine.

instat_zytxWLQwqt

I'm working with the exibble data in gt package.

PR #9432, opened 2025-02-10
Downgrade getPass to 0.2-2 to ensure compatibility

N-thony:
@ChrisMarsh82, I have made the changes here as discussed, have a look. If, this is merged I will create a test version and check if this issue is resolved.

I have also reported the issue on their GitHub https://github.com/wrathematics/getPass/issues/20


rdstern:
@N-thony thanks for continuing with this.  It is really important that we can use Patrick's importing from Climsoft into R-Instat.  
You may want to include @Patowhiz in the tests, so that he can check also that his feature of being able to read from the `Initial `as well as the `Final` table, is working.  

When @jkmusyoka was in Zimbabwe recently, the reading from Initial wasn't working.  But he was using an older version of R-Instat (because reading from Final was n't working in Version 0.8.0, or 0.8.1).  So it is possible that reading from Initial works fine in the new version.

@Patowhiz MSD may have solved this issue before you arrive.  If not, now that I hope everyone is agreed that the Final table is not final, I assume that one solution is to quickly transfer everything there.  We now know the quality control is pretty feeble, so maybe do it after putting all the data together?

ChrisMarsh82 (approved):

Patowhiz:
Great that this has been solved.

PR #9433, opened 2025-02-10
Fixed a bug in the Insert dialog

derekagorhom:
While looking to start working on #8800
I realized the if the from variable checkbox on the savedataframe was still check and you move to a option (say the column option). the textboxes from the variable appear on top on the original textboxes on the new option.
This PR fixes that bug
@rdstern , @N-thony can you review and approve
Thanks

### DIFF
--- a/instat/dlgScript.Designer.vb
+++ b/instat/dlgScript.Designer.vb
@@ -159,7 +159,7 @@ Partial Class dlgScript
         'rdoFromRFile
         '
         Me.rdoFromRFile.AutoSize = True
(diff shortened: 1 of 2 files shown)


derekagorhom:
@rdstern this PR fixes the cosmetic change you requested

rdstern (approved):
@derekagorhom many thanks for noticing and fixing these two points.  If only I had read your work earleir I would have saved my own time, re-discovering what you had already fixed!
@N-thony when you are well again this is ready for you.

PR #9434, opened 2025-02-10
getPass version correction

N-thony:
@ChrisMarsh82 I have made the correction for the downgrade getPass version to 0.2-2. Have a look.


instat/static/InstatObject/R/Rsetup.R
ChrisMarsh82, line 1: Rsetup checks all packages are installed on startup. getPass wants to be included here even though it installed manually. similar to mmtable

ChrisMarsh82 (changes requested):

rdstern (approved):
@N-thony I'm not sure quite how I should be testing this pull request, because af the different way packages are loaded in the VS and released versions.
I tried without loading anything, and it works - at least as far as giving the right input menu.
So I'm approving.  Over to you @ChrisMarsh82

rdstern:
After @ChrisMarsh82 message I don't think we need mmtable2 anymore!  That's a different issue maybe, but can we soon do a check of what packages we are still using?  I would guess we can get rid of quite a few.  And there are quite a large number that we load, every time R-Instat starts.  I wonder if we need them all,  from the beginning?

rdstern (approved):
My only check is that it now gives the popup dialog.  I assume @Patowhiz will check, then the test version is made,because MSD will be the first users.

jkmusyoka:
I get an error message that I do not have the getPass package installed! I assume that the new R-Instat installer will now come with this package?

Patowhiz:
Would be good to get this correction made. In the meantime @jkmusyoka @rdstern we could evaluate how to connect to the database through a script just incase the package fails again in future. The other alternative that we have is getting raw data from either initial or final through the Climsoft front end. We could then document the steps for QC and products when a user chooses that path.

rdstern:
@Patowhiz I like the idea of a script as a backup - but only that.
The new test version is being produced today by @N-thony specifically to test that this feature is working.  So your urgent job is to confirem that.  Hopefully both from Initial and Final.
Very happy with the export from Climsoft also being possible.  Is there currently an easy way that the current it will export all the data and not simply average or total the values from the duplicate records.  Is there an export for the synoptic data that could be used?

Patowhiz:
@rdstern I'm happy to test. I'm hoping to get a stable version that I can use for ZINWA activities as well. 
In regards to exporting raw data. Yes, the administrator can export raw data from either tables. Note, by raw, I mean data in the state that it is in the database. I haven't had the opportunity to test the export performance for big data sets but we are working on improving data export as part of the UNDP project. Exploration of this option could be good for other Met services as well and it will support the web operations.

instat/static/InstatObject/R/Rsetup.R
ChrisMarsh82: install_packages is the list of packages that we need (without dependencies) 
Rsetup is created from install_packages and should include all dependencies and versions. 
installPackages are all packages installed including dependencies. This is used by the installer to install all the packages. This should also be created from install_packages however due to issues with installing some packages there are manual edits. 

It would be worth reviewing all packages in install_packages making sure we only have listed the required packages. Then run this locally and create the packs and version lists for RSetup, update this and installpackages. The awkward bit comes in getting a blank R version to test installpackages before it runs in actions. It may be trial and error to get the actions to install all packages.

ChrisMarsh82 (approved):

rdstern:
@Patowhiz and @jkmusyoka please can you download and check 
a) It now works from Final and 
b) It also works from Initial.  

@Patowhiz I assume you have access to a climsoft database with some data in Initial, or you can transfer some data there for the checks?

Patowhiz:
@N-thony @rdstern I'm not sure on how I can download from the link above. Could you send us a link that directly leads to a download page?

rdstern:
@Patowhiz unusual for me to do something you can't .  The option to downlaod is at the bottom of the  [link above](https://github.com/IDEMSInternational/R-Instat/actions/runs/13259765065) :

Patowhiz:
Thanks @rdstern for precisely pointing out where the download link is located.  I've been able download it. 
As shown in the screenshot below, I still get the `getPass` error. 
@N-thony did you test the set up and does it work on your end? 
@jkmusyoka do you get the same error as well?

jkmusyoka:
@Patowhiz yes we get the same error. I believe @N-thony is at the moment working to resolve this issue

PR #9437, opened 2025-02-12
Improving the Find-replace dialog

### LINKED ISSUE #9326: Improve the text find-replace dialog, and the setting items to NA?
@lilyclements I suggest there are 2 items missing.  
a) We don't have a good way here to set values as missing, i.e. NA.  We actually do, but in our Replace dialog?  The one item not yet in that code is to be able to ignore case in the string set into missing.  
We could use the facility in the naniar package, but I wonder if our existing code is good enough that we could just add that element.  I have an example, of station data from Zambia, that would be good to test.
b) We have nicely added the starts, end, contains commands into the rename dialog.  Should they also be added here?

Vitalis95:
Fixes #9326
@rdstern @N-thony , the PR sets items to NA


rdstern:
@Vitalis95 this already seems to work.  Just 2 points:
a) Could you change the `Replace NA` into `From NA`
b) The other buttone have lost the Pattern control, e.g.

image

rdstern (approved):
@Vitalis95 very nice.
@N-thony over to you to check and merge.
This is only small, but important for R-Instat.  I want to minimise the things that should be simple - and they are for R-whizzes.  But they can be frustrating for R-Instat users!

Issue #9026, opened 2024-06-20
Add Find/Replace to script window, or a Specific change to piping in scripts 


rdstern:
As context I give some background to this issue.
I am preparing for "practices" and videos on the graphics and tables.  I am finalising the first, which is on line plots.  I am following the tidyverse example script and showing how it can be done in 2 ways.  First using dialogs as much as possible and second, by adapting the script for the example.
I plan to do the same for the tables.  The first example is the script provided for the pizzaplace data.
The gt documentation has a whole collection of useful scripts.  They all follow the same pattern and use piping obsessively.  I would like to show how these can easily be adapted for use in R-Instat.  The problem is that they use the modern symbols for piping, namely `|>`  and this causes problems, when adapted in our scripts (it crashes R-Instat).  There is a simple "work-around", namely to replace `|>`  by the original `%>%`.  All works fine then.

If we had a simple Find/Replace in the script editor, then this would be an example.  @Patowhiz and @N-thony, Stephen says this could be quite a bit of work.  So, I wonder if it would be easy to add an option in Insert to make this specific change.  That would be good to have anyway, because the addition also provides a reason to explain (in the help) why it is needed.  Hence, if we did have a general Find/Replace, the I would be requesting this special case as well.  Would it be easy to start with this special case?

Patowhiz:
@rdstern to insert what specifically, the `%>%` symbol?

rdstern:
@Patowhiz Exactly.  That stops R-Instat from crashing!

Patowhiz:
@rdstern I can see how find and replace dialog is useful but I'm not sure how using a dialog to just insert `%>%` is better/efficient compared to just typing it. 

As a  user If I'm using a script that I have pasted elsewhere, most likely it will have lots of `|>` symbols. Replacing each one every time I paste a script could be very hectic.

rdstern:
@Patowhiz you write "As a user If I'm using a script that I have pasted elsewhere, most likely it will have lots of |> symbols. Replacing each one every time I paste a script could be very hectic."
You are right, the special replace `|>` by `%>%` would be a replace All.  So it replaces throughout the script.  Currently I am having to use <ctrl>V repeatedly and it is "hectic", and also easy to miss some.

rdstern:
@lilyclements and @fran2or might it be easy to use R's powerful find replace on our scripts?
We might, at least initially add a Find tab into the Script windows Insert dialog.
[...]

PR #9441, opened 2025-02-13
Added the Side Plot option for the graph option in the insert dialog

derekagorhom:
Fixes partly #8800 
I have added the side plot option for the graph option of the Insert dialog. 
Currently still working on the Text path option. 
@fran2or can you review this
Thanks 
This is still not ready for review

### DIFF
--- a/instat/dlgScript.Designer.vb
+++ b/instat/dlgScript.Designer.vb
@@ -159,7 +159,7 @@ Partial Class dlgScript
         'rdoFromRFile
         '
         Me.rdoFromRFile.AutoSize = True
-        Me.rdoFromRFile.Location = New System.Drawing.Point(49, 202)
+        Me.rdoFromRFile.Location = New System.Drawing.Point(49, 206)
         Me.rdoFromRFile.Margin = New System.Windows.Forms.Padding(2)
         Me.rdoFromRFile.Name = "rdoFromRFile"
         Me.rdoFromRFile.Size = New System.Drawing.Size(81, 17)
@@ -174,7 +174,7 @@ Partial Class dlgScript
         Me.ucrInputSaveRFile.AutoSize = True
         Me.ucrInputSaveRFile.IsMultiline = False
         Me.ucrInputSaveRFile.IsReadOnly = False
-        Me.ucrInputSaveRFile.Location = New System.Drawing.Point(67, 228)
+        Me.ucrInputSaveRFile.Location = New System.Drawing.Point(67, 231)
         Me.ucrInputSaveRFile.Margin = New System.Windows.Forms.Padding(9)
         Me.ucrInputSaveRFile.Name = "ucrInputSaveRFile"
         Me.ucrInputSaveRFile.Size = New System.Drawing.Size(145, 21)
@@ -259,7 +259,7 @@ Partial Class dlgScript
         '
         Me.ucrChkDisplayGraph.AutoSize = True
         Me.ucrChkDisplayGraph.Checked = False
-        Me.ucrChkDisplayGraph.Location = New System.Drawing.Point(88, 217)
+        Me.ucrChkDisplayGraph.Location = New System.Drawing.Point(88, 225)
         Me.ucrChkDisplayGraph.Margin = New System.Windows.Forms.Padding(6)
         Me.ucrChkDisplayGraph.Name = "ucrChkDisplayGraph"
         Me.ucrChkDisplayGraph.Size = New System.Drawing.Size(113, 34)
--- a/instat/dlgScript.Designer.vb
+++ b/instat/dlgScript.Designer.vb
@@ -100,6 +100,18 @@ Partial Class dlgScript
         Me.lblSelectColumns = New System.Windows.Forms.Label()
         Me.ucrReceiverRank = New instat.ucrReceiverSingle()
         Me.ucrSelectorForRank = New instat.ucrSelectorByDataFrameAddRemove()
+        Me.tbGraphs = New System.Windows.Forms.TabPage()
+        Me.rdoSidePlot = New System.Windows.Forms.RadioButton()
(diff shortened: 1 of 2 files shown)


fran2or (commented):
@derekagorhom I've reviewed your progress, and it aligns with our discussion. However, is there a way to ensure that the insert is placed directly where expected within the script? @rdstern, could you also check this? 

Currently, the user must determine the exact location in the code to insert the side plots' code; otherwise, it defaults to the end of the R-Instat graph code, causing it to malfunction. 

@rdstern, is this acceptable, considering that users utilizing this feature are expected to have some coding knowledge? If they can read the code to that level, they might not need an automated insert function for side plots—they could simply write the code themselves. Thoughts?

rdstern:
@fran2or I am about to check the dialog.  But in answer to your query I was assuming that the user has to understand the script sufficiently to put the cursor manually in the right place.  Let's ponder later if we can do anything better.

rdstern:
@derekagorhom and @fran2or this is already working.  Great:
a) The current labels for the side plots are `X Side` and `Y Side`.  For R-Instat users I wonder if `Top` and `Right` would be simpler?  Up to you.
b) A question for @fran2or as an R whizz.  The side plots make use of the aeshetics from the main plot.  I suggest we could usefully modify the code generated for the main plot, e.g. the point plot, so when it adds a colour aesthetic it also adds fill for the same variable and v.v.?  I suggest this could be general in the plot dialogs.  It will not change anything in the main plot and sometimes the side plots could use the other?
c) A more general question for @N-thony.  We don't yet have a find-replace facility for scripts.   This is discussed in a few places, for example #9026.  I wonder if it might be quite a simple job - at least could we check.  If it existed, then we could copy the last item from the graph into the dialog, 
e.g `theme_grey()` in my example.  This would then be replaced by:
`theme_grey() + ggside::geom_xsideboxplot() + ggside::geom_ysideboxplot()`

rdstern:
@derekagorhom that's fine.  Is there a problem adding the other part?

rdstern:
@derekagorhom this is a start.  However there remains a lot to correct here in the code, and possibly in the controls.  I suggest:
a) You disable this option for now.  That is what you had earleir.  Then we can approve with the simple, and useful, controls for the side plots.
b) I am keen to get to a stable release very soon, and there are more important tasks than this one.  So I will be suggesting other issues.
c) You return to this pull request after the next release and @fran2or will be the main tester of your progress.

derekagorhom:
@rdstern I have disabled the Textpath option

rdstern:
@derekagorhom sorry, I keep messing you up on this.  That's where I wish you were doing this with @fran2or as well as with me.
So let's leave it for now - I am beginning to doubt whether it is in the right place, given it works very well just adding a single control.
So, we still have the disabled Side Plot control in the scatterplot dialog.  Maybe have 2, so 1 for Sode Plot (Top) and the other for Side Plot (Right).

But that's not quite right, because the side plots are then just for scatterplots.  They can be added to any plot.  So maybe there is a Sideplot tab in the sub-dialog.  And maybe both, so the feature is on the main dialog for scatterplots, and in the sub-dialog for everything else.

 There are many other features I would like to add quickly into the forthcoming version.  Given the time on this one feature they have higher priority than this one.

fran2or:
> A question for @fran2or as an R whizz. The side plots make use of the aeshetics from the main plot. I suggest we could usefully modify the code generated for the main plot, e.g. the point plot, so when it adds a colour aesthetic it also adds fill for the same variable and v.v.? I suggest this could be general in the plot dialogs. It will not change anything in the main plot and sometimes the side plots could use the other?

That's an interesting suggestion! Modifying the main plot code to automatically add `fill` when `colour` is specified can be done 
 and could indeed make it easier for side plots to inherit the aesthetics seamlessly. However, this change would also mean that side plots, such as density plots, would always be filled, removing the option for unfilled density plots.  

This could be a limitation in cases where **overlapping density plots** need to be interpreted, as unfilled plots often provide a clearer visual comparison. Perhaps a more flexible approach could be to give users the option in the dialog—either allowing `fill` to be added automatically when `colour` is set or letting them decide manually when needed.  

What do you think? Would this balance usability and interpretability?

fran2or:
> So, we still have the disabled Side Plot control in the scatterplot dialog. Maybe have 2, so 1 for Sode Plot (Top) and the other for Side Plot (Right).
> 
> But that's not quite right, because the side plots are then just for scatterplots. They can be added to any plot. So maybe there is a Sideplot tab in the sub-dialog. And maybe both, so the feature is on the main dialog for scatterplots, and in the sub-dialog for everything else.

@rdstern Is there a specific reason why we need the Side Plot controls on the scatterplot main dialog? If not, I think having it present in the sub-dialog would be the best placement. This way, the Side Plot feature will be available not just for scatterplots but for other plots as well, ensuring broader functionality and then avoids duplication.

rdstern:
Thanks @fran2or for the quick reply.  I went through the same steps as you have done.
a) I assumed adding the side plots would be quite complicated, and so thought the script- insert route was the way to go.
b) Then you and @derekagorhom showed we could add them with just a single control.  That made me reflect that perhaps the single control could go onto the main scatterplot dialog instead.
c) But, as you say, the side plots can be added to any main plot, so maybe a sub-dialog is better.  That could be an easy way also to add the same options for that extra as for the other geoms.  But maybe we could have a single "page" for all the geoms, with just the control at the top of the page dictating the exact geom?
d) I noiw wonder about both.  Nice to be able to show them, really easily for one type of graph.  Then we add them, just from the sub-dialog, for the other geoms.

lilyclements:
@Ag-Derek can you comment on the state of this, and the plan going forward

rdstern:
Closed as side plots now in 2/3 variable plot dialog.

PR #9445, opened 2025-02-14
Improvement in View Menu > Swap options

### LINKED ISSUE #9368: View menu buggy?
Hello @rdstern, @N-thony and @Patowhiz 

Please could someone take a look at the view menu and check it is all working as expected. I am having trouble with it. I used the swap data and log window option at the bottom and now it seems to be stuck. 
 
Image

When i click on Reset Default Layout nothing happens. When i click on the black arrow in the tool bar nothing happens either. 

When i uncheck the log script window from the view list I get this. 

Image

But then when I click on the data window in with view menu it shows up like this. Top and bottom lay out instead of side by side. 

Image

Then when I use the reset default layout arrow or menu item it goes back to this. 

Image

Can anyone tell me if this is a bug or if i am using it wrong? 

Thanks, 

Rachel

N-thony:
Fixes #9368
@rdstern, have a look.


rdstern:
@N-thony it is looking good on the toolbar data drop-down options.
The script one does not work however, see below:

image

It works from the View menu but not from the toolbar.  That has correctly been ticked, but needs to be re-activated.

rdstern (approved):
@N-thony very nice.  Thanks.
Can you merge or perhaps @Patowhiz may still have time to check and hopefully merge?

Patowhiz (approved):

Issue #9440, opened 2025-02-13
Correct and improve the calculator and our summary system for ordered factors


rdstern:
The diamonds data has 3 ordered factors.  We don't yet handle them very well, and I suggest this is important to correct and improve in R-Instat.

Why important?  Because we should discuss how to tidy and analyse data of different types.  (The main types are numeric, factor and character, but also date, logical etc.  

Factors are categories and really important and not well understood - maybe because SPSS doesn't have factors?  

Anyway note that - an answer to "can you get percentiles for ordered factors" gave:

`Yes, you can calculate percentiles for ordered factors, essentially treating them like a continuous variable where each level represents a certain position within the ordered distribution, allowing you to determine the percentile rank for each category based on its position within the ordered factor levels; however, it's important to understand that this interpretation is not as precise as with numerical data due to the inherent discrete nature of categorical variables.   `

In R the quantile function provides the general way to calculate percentiles.  I suggest a simple improvement and also a correction to the calculator.
Then I need more information to see how to improve the summary system.  

@lilyclements here is the dialog:

![Image](https://github.com/user-attachments/assets/ea54427d-8d48-48af-a752-e64497e7af5b)

 I am not getting any summaries from the main tab - min and max work fine, and I've ticked all the candidates from the More tab:

![Image](https://github.com/user-attachments/assets/302dd869-cc55-4a11-aba2-7bf18b3bd7d4)

This gives an error currently - not surprised - for an ordered factor.  Here is the one-line script generated:

```
# Dialog: Column Summaries

data_book$calculate_summary(data_name="diamonds", probs=0.5, store_results=TRUE, return_output=FALSE, factors="color", columns_to_summarise="cut", 
count_test="'=='", count_value=1, prop_test="'=='", prop_value=1, j=1, summaries=c("summary_quantile", "proportion_calc", "count_calc"), silent=TRUE) 
```

Could you include the code for the summary_quantile, proportion_calc and count_calc functions.  I am hoping that some, or all, use the R quantile function, which is easy to adapt for ordered categorical data.

In our summary system we have the median and lots of individual quantiles on the main page.  I propose we leave them for only numeric data, and just adapt one or more of the 4 above, to cope with ordered categorical.

Now to the calculator.  Assuming @Fidel365 is still working on R-Instat, he is the ideal person for this!

a) Please correct and simplify the quantile key.  It is currently as follows:

[...]

PR #9456, opened 2025-02-24
Changes to summary keyboard in calculator

Fidel365:
Fixes partially #9440
@rdstern I saw to do this, on the calculator part, it's ready for review.


rdstern:
@Fidel365 

a) The keyboard layout is fine.  Well done.
b) But look below:

image

you have included the name of the data frame after x = .
Instead it should be blank and with the cursor there, so the correct variable can be included.

This is both for quantile and quantile3.  
And can you make the tooltip a bit shorter. So:  **_for the variable, from the ggplot2 diamonds, gives_**

Fidel365:
@rdstern ready once more.

PR #9458, opened 2025-02-24
Improvements to Prepare > Data Frame > Insert Columns/Rows dialog

### LINKED ISSUE #9455: Small improvements for the Prepare > Data Frame > Insert Columns/Rows dialog
Here is the current dialog, with the usual survey data:

Image

a) A cosmetic change.  Delete the label `Column: `in the dialog.  Then move those 2 controls left in the box, and the dialog can become narrower.  
b) You can see the result from the default use of the dialog.  The new column is filled with NA and is also a logical column.  Could you make the default into a numeric column please.
c) Get rid of the tooltip, which is on the Value control.  It is incorrect.

Image

d) In the drop-down change True to TRUE and False to FALSE.  If either of those are used as the default, then make the resulting column logical.  (Currently you can't include TRUE - the dialog changes it to True.  That's all wrong!

Vitalis95:
Fixes #9455 
@rdstern @N-thony , have a look at it

### DIFF
--- a/instat/dlgInsertColumn.vb
+++ b/instat/dlgInsertColumn.vb
@@ -81,8 +81,8 @@ Public Class dlgInsertColumn
         dctDefaultValue.Add("0", "0")
         dctDefaultValue.Add("Kisumu", Chr(34) & "Kisumu" & Chr(34))
         dctDefaultValue.Add("New Zealand", Chr(34) & "New Zealand" & Chr(34))
-        dctDefaultValue.Add("True", "TRUE")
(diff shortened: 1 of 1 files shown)


rdstern (changes requested):
@Vitalis95 that's all very nice.  Many thanks.
Could you just change True to `TRUE` and False to` FALSE` in the drop-down list.

rdstern (approved):
@Vitalis95 thanks, that looks fine now.
@N-thony please can you check and merge?

PR #9466, opened 2025-02-27
Correcting packages versions

N-thony:
@rdstern @ChrisMarsh82 I have here corrected the getPass version to fix the issue we had last time not being able to get the small password window.
@rdstern, I'm wondering if we can consider in this PR the attachment of only ```"gt", "instatExtras", "instatClimatic", "databook", "instatCalculations" ``` packages?

### DIFF
--- a/instat/static/InstatObject/R/Rsetup.R
+++ b/instat/static/InstatObject/R/Rsetup.R
@@ -219,10 +219,7 @@ load_R_Instat_packages <- function() {
   # tidyr loaded because unite() function is required by mmtable() function from mmtable2 package
   # purrr loaded because map_int() is required  by function(s) such as header_top_left() from mmtable2 package
   # mc2d loaded because of triangular and continuous empirical distributions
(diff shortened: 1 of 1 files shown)


Patowhiz:
I thought the `getPass` issue was fixed. @jkmusyoka @rdstern keep tab on this before the Zambia work in. @jkmusyoka also evaluate using R scripts for setting the database connection credentials. We may have to change the internal data book code to allow a connection credentials to be set from a script. The other alternative is using the **Administrator** menu on Climsoft desktop to get the raw data from the **Update Observations** dialog.

rdstern:
@N-thony that packages drop would be interesting.  Perhaps also keep ggplot2 to emphasise graphs and tables!  But good to try that.

rdstern (approved):
I have checked and it still looks fine!
Maybe @jkmusyoka can be the second approver here!

jkmusyoka (approved):

PR #9474, opened 2025-03-03
Improvement in Reorder Levels Dialogue

### LINKED ISSUE #9451: Make our factor dialogs simpler to use
Our factor menu is really impressive - that's because one reason why R is so good for data analysis is because of factors.

And forcats is an impressive package for handling factors.  

Factors are categorical columns.  We need to make using them as simple as possible.  Here is an example:

Image

This looks a bit daunting.  Notice there are not many variables in the selector, because we only have the factor variables available there.  In the example above the corsor is in the Variety variable - which is a factor.  So, please could the dialog start with the selected variable in the receiver, if it is a factor.  So here the dialog would start as below:

Image

That's clearer.  Why not.

I further suggest that if the current variable in the data view (or the metadata view) is not a factor, then the dialog starts with the first factor variable in the selector, copied into the receiver, as shown below.

Image

What is also nice about this scheme, is that occasionally we have data still without factors - and that's often because we have forgotten to make those variable into factors.  So now the only reason that the receiver is empty is when there are no factors in the data.

In the factor menu I propose this change (selected variable if a factor pr ordered factor, and first factor otherwise) for the following dialogs that have a single receiver - except 1, namely Unused Levels.

1. Recode Factor
2. Dummy Variables
3. Levels/Labels   - this does it nicely from right click already.
4. Reorder Levels - as shown above
5. Reference Level
6. Contrasts
7. Factor Data Frame

Unused Levels is different.  There we check the factors from the selector and fill the receiver with the first factor that has unused levels.  If no factors have unused levels then we don't fill the receiver.

N-thony:
Fixes #9451
@rdstern, have a look at it if it works as expected. Then @derekagorhom will continue with other dialogues.

### DIFF
--- a/instat/dlgReorderLevels.vb
+++ b/instat/dlgReorderLevels.vb
@@ -287,19 +287,32 @@ Public Class dlgReorderLevels
     Private Sub SetSelectedColumn()
         Dim strTempSelectedVariable As String = ""
         Dim strDataName As String = ucrSelectorFactorLevelsToReorder.strCurrentDataFrame
-        If Not String.IsNullOrEmpty(_strSelectedColumn) AndAlso frmMain.clsRLink.GetDataType(strDataName, _strSelectedColumn).Contains("factor") Then
+        Dim strTemp As String = ""
+
+        ' Retrieve parameter value safely
+        Dim clsParam = clsDummyFunction.GetParameter("strVal")
+        If clsParam IsNot Nothing Then
+            strTemp = clsParam.strArgumentValue
+        End If
+
+        ' If _strSelectedColumn is valid and a factor, use it
+        If Not String.IsNullOrEmpty(_strSelectedColumn) AndAlso
+       frmMain.clsRLink.GetDataType(strDataName, _strSelectedColumn).Contains("factor") Then
             strTempSelectedVariable = _strSelectedColumn
-            Dim strTemp As String = clsDummyFunction.GetParameter("strVal").strArgumentValue
-            If Not String.IsNullOrEmpty(strTemp) AndAlso strTempSelectedVariable <> strTemp Then
-                strTempSelectedVariable = strTemp
-            End If
         ElseIf ucrSelectorFactorLevelsToReorder.lstAvailableVariable.Items.Count > 0 Then
+            ' If no selected column, use first available variable
             strTempSelectedVariable = ucrSelectorFactorLevelsToReorder.lstAvailableVariable.Items(0).Text
         Else
-            ' Handle the case where there are no available variables
+            ' No available variables, exit
             Exit Sub
         End If
 
+        ' Ensure strTemp takes precedence if it's valid
+        If Not String.IsNullOrEmpty(strTemp) AndAlso strTempSelectedVariable <> strTemp Then
+            strTempSelectedVariable = strTemp
+        End If
+
+        ' Add the selected variable to the receiver
         ucrReceiverFactor.Add(strTempSelectedVariable, strDataName)
     End Sub
 


rdstern:
@N-thony very nice, I think it is almost there.  I tried with survey and then also diamonds.  It seems to be working ok for both factor and ordered factor variables.
a) Going in for the first time you can conform, but I think it is exactly right.  Namely if the focus is on the variety (factor) variable, then that it added.  If the initial focus is on a non-factor variable, then village is added (as the first factor.
b) Going back to the dialog, the rule that it is as it was left has precedence.  So it still re-opens as it was left.
c) Pressing reset empties the control, so that then nothing is selected.  That is right too.
d) The only oddity is shown here:  I have pressed Reset to empty the control.  The 3rd factor has the focus.  Earlier the first factor (Village) was selected.  

image

If I press Close and then Re-open the Village is filled, and I would prefer it to be the fert one?

rdstern (dismissed):
@N-thony this is great.  I am approving.
@lilyclements and/or @berylwaswa you may also wish to check and I hope you might agree this is promising.  This is only working in the Prepare > Column: Factor > Reorder levels dialog.  But I hope it will soon be added to most of the other factor dialogs, and also some others, including some text dialogs.  But it is for the factors that I am particularly keen.
I am convinced we should have a training video and practice titled something like `"I'm completely comfortable with categorical columns"`  
So I want to make it as simple as possible to use the factor dialogs.  Hence - in the Reorder dialog @N-thony has made it that:
a) If, in the data, the focus is in a categorical (factor) column, then, when you go into the dialog, that factor is automatically in the receiver.
b) If you are not in a factor column, then the first factor in the data-frame is loaded.
c) So, if nothing is loaded when you go into the dialog it is only because you don't yet have any factors in your data.  (I must check that - I have not tried yet with any datasets with no factors.)
d) So this is a bit like the climatic dialogs.  There, when you define the data as climatic, then some receivers are filled in for you.  Here when you define factors, the same happens.

@ChrisMarsh82 you may also wish to check the code, before merging, as we plan to re-use it in other dialogs?

lilyclements (dismissed):
Reapproving on @rdstern's behalf - I tested and it's working great for me. Nice one, @N-thony. I'll wait for @ChrisMarsh82 to review though too.

rdstern (dismissed):
@N-thony I am approving so we could merge.  However, I don't understand why @derekagorhom can't apply the same ideas in #9492 to the other factor dialogs.  It will be odd if that feature only works for the one dialog.
So I prefer to wait, with merging until either Derrick examines your code and hence finishes #9492, or (someone else takes over - maybe it has to be you) or we don't have this feature in R-Instat?

rdstern:
@N-thony I may owe an apology to @derekagorhom as I had assumed your pull request was working fully.

image

Please tell me if I am wrong with what I expected here, see above:

a) I am in a non-factor and go into the Reorder Levels dialog.  It correctly put cut (the first factor column) into the receiver.
b) But I wanted the second column, so pressed color- as shown above.
c) Now I press Close and then re-open.  I suggest, in this case the "remembering of the last state" should have happened, so color should have been there when I re-opened.  But it went back to its original default of cut?


1.  I report 2 other possibilities that are currently working well.  So I'm keen we keep them with any improvements:

a) I press Reset in the dialog as shown below and then Close:

image

b) I remain in a none-factor column and re-open.  It now goes beck to the original default, with cut.  I am very happy with this.

2. I am now in a factor column.  
 
a) It correctly opens that factor, namely claity. 
b) But I move to another factor (color) as shown below.  I now close.

image

c) Now I re-open.  And it is just as I left it, i.e. it still has color - and has not reverted to clarity.  Whoopee.  So it is perfect when I am starting in a factor variable.  It should do the same when I start in a non-factor variable.

I hope that is clear?

Issue #9464, opened 2025-02-26
Fix error and slightly improve the Prepare > Column: Text > Split dialog


rdstern:
Here is an [example of data](https://stackoverflow.com/questions/19339296/plotting-contours-on-an-irregular-grid) where the Split dialog fails:

The data can be copied and pasted into R-Instat using Edit > Paste New.  Here it is:

![Image](https://github.com/user-attachments/assets/f033050b-46c9-4871-a7df-2f96168830b1)

I find it annoying that the data does not paste into 3 variables - unless we go via Excel, but that is a separate issue.  And it meant I could try the Split dialog that led to this issue:

![Image](https://github.com/user-attachments/assets/9638e6b2-1c5f-4ca8-9527-e61085e41c06)

So here are the data, together with the dialog and the script:

a) I think the R line in the code producing x is working Ok.  But the 3rd line, putting the results back into the data frame, has problems.  It used to produce multiple columns, while now it makes just one and a bit.  So that's the bug.
b) The data adds a problem, because there are multiple spaces and so it makes more than 3 new columns.  Please add a new option called spaces(  ) - with 2 spaces inside the bracket.  This to allow for one or more spaces as a separator.  There is a squish function in stringr that changes multiple spaces into just one.  So that just needs to be added to the code in line 2.  That's the enhancement!

rdstern:
@lilyclements I asked @Vitalis95 to fix the split bug and also to enhance the dialog a bit.  
He has identified - I think - that the bug is caused by the move to the packages.  It is working fine in Version 0.8.2 anyway.  I hope it is as easy to fix as the others seem to be, so far.

Then Vitalis work, enhancing the dialog/function will be a good start to your new "system" for the developers.

lilyclements:
Thanks for finding this. I found another one earlier today for sorting variables (you cannot sort by descending). This is fixed [here](https://github.com/IDEMSInternational/databook/pull/45). If you update `databook`, it should hopefully now work!

lilyclements:
Fixed by #9464. Updates to this given in #5992

PR #9477, opened 2025-03-04
Improving the Prepare > Column: Text > Split dialog 

Vitalis95:
Fixes partly #9464 
@rdstern @N-thony , the pr removes extra spaces and ensure proper splitting


rdstern:
@Vitalis95 I note the bug is still there, namely that it produces just the 2 variables (not the three as is needed in the example I used.  I think that is maybe a @lilyclements fix, as it was working ok in Version 0.8.2.
But also I was keen that you add an addition to the pull-down called Spaces(   ), and that is where you include the squish.

I would still like the space( ) option to work as before - where (if there are say 3 spaces) it gives one extra variable per space.  There are occasions where that might be useful.

Incidentally I am also hoping you will add the spaces(  ) option into the Paste New dialog.  There is will not give options on the separator, and it will use the squash code.  So it is just a checkbox to split on the multiple spaces option.

Vitalis95:
@rdstern , update` databook`. It should work

Vitalis95:
@rdstern , added Space Separated control to Paste New dialog

rdstern (changes requested):
@Vitalis95 this is now very nice indeed.  I was about to approve but  I note you have also added the checkbox in the Paste New dialog.  

It is working better than I thought possible, so I am very happy about this feature too.  The way I suggested - with Split, would not have split the variable names.
image

I note that (like the tables dialogs) Ok needs to be pressed to make the multiple columns.  

I tried with Ok and Keep, but then you have to refresh the display, and there is a button for this.  That seems to be done when the dialog is re-opened, so perhaps that could be included after the Ok and keep too?

The one item to correct - I think - is that the Space Separated is usually only applicable when there is a single variable.  So, when more than 1 column is pasted, thenit is unticked and disabled?

image

rdstern (approved):
@Vitalis95 this looks great now.
@N-thony there are 3 main changes:
a) The bug in the split is resolved
b) There is an extra option in split when there can be multiple spaces between items.
c) There is a great additional checkbox in the Edit > Paste New to split items when there are spaces.

Really nice.  Over to you to check.

PR #9482, opened 2025-03-06
Adding package name to prevent error in Describe Two Variables

lilyclements:
Fixes #9821

@rdstern this is ready to review

### DIFF
--- a/instat/dlgDescribeTwoVariable.vb
+++ b/instat/dlgDescribeTwoVariable.vb
@@ -317,6 +317,7 @@ Public Class dlgDescribeTwoVariable
         clsDummyFunction.AddParameter("var", "False", iPosition:=5)
         clsDummyFunction.AddParameter("percent", "col", iPosition:=6)
 
+        clsPivotWiderFunction.SetPackageName("tidyr")
         clsPivotWiderFunction.SetRCommand("pivot_wider")
         clsPivotWiderFunction.AddParameter("values_from", "value", iPosition:=2)
 


rdstern:
@lilyclements and @N-thony I report on how I am now testing, using the new branches - my runs with version 1.0.0.
a) I used sessionInfo() to find which versions of the packages were attached.  
b) I checked with the Tools > Add Packages and it said that all 4 packages need be iupdated.
c) I detached them in turn and found that Extras and databook still needed updating, but it said the other 2 were up-to-date.
d) I updated Extras and databook, from the existing dialog, and it worked fine.  So now said all packages are up to date.
By the way it keeps saying databook is version 0.1.5?
Calculations is 0.1.3 and the others are 0.2.2.

Now to the check.  I report, in passing, an odd problem with the survey data that the One Variable Summarise and 2/3 variables summarise don't list any of the variables in the survey data.  It does in the other dialogs, and does in those for other datasets.  So I tried the 2 way frequency table again and got a different error:

image

I tried other summaries and they worked fine

lilyclements:
Ok so we need to call `count_label` in R-Instat still. The labels we call in our R package (`databook`) are just for use in the R package, but when we call in R-Instat we need them to be in R-Instat. 
So, I've updated the labels file to include only labels that are in R-Instat.
This means you need to update the `databook` (again, sorry) to get the other labels that are in the `databook` and aren't in R-Instat.

rdstern (approved):
@lilyclements sorry I overlooked that!
@N-thony I am approving.  Can you check and merge.
I note that we currently still need to include dplyr, tidyr and purrr to now seemingly get everything to work.

PR #9483, opened 2025-03-06
Improvement in the setting of the user directory

### LINKED ISSUE #9467: The issue with getPass seems to be back
@lilyclements @rdstern 
After moving from the instat files to a separate R packages, I realize that the issue we had with `getPass` is back in the test version 0.8.3 of R-Instat. We have fixed this by downgrading `getPass` from version `0.2.4 to 0.2.2`, but this seems like another package in the installation process might have getPass as a dependency and required version 0.2.4 and I suspect the `databook` package.

N-thony:
Fixes #9467

Thanks @lilyclements for updating `instatExtra::set_library_paths` in https://github.com/IDEMSInternational/instatExtras/pull/65
@rdstern, this will make sure that the user directory is created if not yet and set it as a primary library for packages before running the code to attach the packages in `Rsetup.R`. Now, when you update a package which goes to the user directory, it will be attached after restarting R-Instat.


rdstern:
@N-thony not sure how and what to test here?
I started with sessionInfo() and found still the old versions were attached.

But I can't anymore load our packages.  it says:

image

Should I be doing things in a different way?

lilyclements:
Hi @N-thony I am going to investigate this. It's been working for @rdstern, but for me I've been getting this issue too. I think perhaps the `devtools` function has changed so our codes for `1`, `2`, ..., etc have changed
(We would say "If you get this warning, give this output", and for me it's been wrong for a few weeks now, but since @rdstern hasn't had an issue I've left it to prioritise other more urgent things before investigating). In the meantime can you install by `devtools::install_github("IDEMSInternational/databook", update = "never")`

lilyclements:
@N-thony great - glad it worked in the end. I'm not sure why the databook directory isn't there though

rdstern:
@N-thony I am not sure if the following observation helps.
On the merged version (i.e. not this branch) when I go into R-Instat I run the first line of the following:
```
sessionInfo()
detach("package:instatCalculations", unload=TRUE)
library(instatCalculations)
detach("package:instatClimatic", unload=TRUE)
library(instatClimatic)
detach("package:instatExtras", unload=TRUE)
library(instatExtras)
detach("package:databook", unload=TRUE)
library(databook)
```
The sessionInfo() tells me that I have the original versions of the 4 Instat packages, i.e. not @lilyclements updated ones.
So I run the remaining lines, to detach and then attach the packages again.  Then run the sessionInfo() again and I get the new versions, e.g. version 0.2.2 of instatExtras rather than the original 0.2.1. 

Note I am not installing anything new, just changing the attached version.

When I try the same with your branch then:
a) The first sessionInfo() is the same as the merged version.
b) And the second sessionInfo() is unchanged from the first.  
So, have you got the default (first directory) the wrong way round?


a) The

ChrisMarsh82:
@N-thony you have the option to build from a branch. This means you can create a build for this branch and test it before merge.

ChrisMarsh82:
@N-thony try using your fork. The branch you have specified lives on your fork https://github.com/N-thony/R-Instat/tree/improvement_libPaths

rdstern:
@N-thony I tgried downloading and installing.  And it seems to run.

But the packages are the "old" versions.  

image

I thougbht I could update them, but it doesn't seem to recognise that I am online.  I think I am an the CRAN option seems to run.  So I can't update.

I am also not sure why I need to update?

rdstern (dismissed):
@N-thony I am not sure what to test here.  I would like a new version, and maybe even the merged version, to start with the most recent version of the packages. 
@ChrisMarsh82 I am approving so you can check and merge

rdstern (approved):
@ChrisMarsh82 I am approving again as I am not sure what concernes me.
Could you now check and merge, or maybe involve @Patowhiz if that would be better?  Can this be merged before @N-thony leaves?

rdstern:
@Patowhiz I am trying to get a version that we can test well during the sprint.  This one relates to your importing from Climsoft.  Can you approve so we can merge and then see if the problems with getPass are now behind us?

PR #9487, opened 2025-03-10
Improving the pivot-table dialog

### LINKED ISSUE #9473: Improve the pivot-table dialog?
This needs discussion with @lilyclements before being done.

Here is the current pivot table dialog:

Image

I have been using the usual diamonds data and this has led me to two  suggestions for improvements.  

a) I would like to start with just one row factor (cut) and two column factors, (clarity and color).  The current dialog only allows for a single column receiver.  I suggest changing it to a (small) multiple receiver.
b) To keep the dialog from getting deeper I suggest moving the bottom Factor control to the left-hand side?  That also links with c).
c) The "system" doesn't cope well with respecting the order of factor levels.  There is a "tweak", that they suggest, added by Lily for a single factor.  I would like it, for the diamonds, to be possible for 2 factors.  I don't think it need be more general than that.  Lily, would that be easy to include?

Here is the current code, where the change to a multiple column receiver will be trivial.  But the addition of another factor to be relevelled is not so clear?

```
# Dialog: Pivot Table

diamonds <- data_book$get_data_frame(data_name="diamonds")
survey_levels <- stringr::str_flatten(string=paste0("\"", levels(x=diamonds $ clarity), "\","))

diamonds <- data_book$get_data_frame(data_name="diamonds")
relevel_variables <- paste0("function(attr) {  var sortAs = $.pivotUtilities.sortAs;  return sortAs([", survey_levels,"]); }")
last_table <- rpivotTable::rpivotTable(data=diamonds, rows=c("cut","color"), cols="clarity", sorters=relevel_variables, subtotals=FALSE, rendererName="Table", aggregatorName="Average")
data_book$add_object(data_name="diamonds", object_name="last_table", object_type_label="table", object_format="html", object=last_table)
data_book$get_object_data(data_name="diamonds", object_name="last_table", as_file=TRUE)
rm(list=c("last_table", "diamonds", "relevel_variables", "survey_levels"))  
```

Vitalis95:
Fixes #9473 
@rdstern @N-thony , added `Factor Labels in Order` control and changed `Initial Column` receiver to multiple receiver .

### DIFF
--- a/instat/dlgThreeVariablePivotTable.Designer.vb
+++ b/instat/dlgThreeVariablePivotTable.Designer.vb
@@ -37,32 +37,33 @@ Partial Class dlgThreeVariablePivotTable
         Me.ucrChkIncludeSubTotals = New instat.ucrCheck()
         Me.ucrReceiverSelectedVariable = New instat.ucrReceiverMultiple()
         Me.ucrChkSelectedVariable = New instat.ucrCheck()
-        Me.ucrReceiverInitialColumnFactor = New instat.ucrReceiverSingle()
         Me.ucrBase = New instat.ucrButtons()
         Me.ucrSelectorPivot = New instat.ucrSelectorByDataFrameAddRemove()
-        Me.lblFactorLevels = New System.Windows.Forms.Label()
-        Me.ucrReceiverFactorLevels = New instat.ucrReceiverSingle()
         Me.ttFactorLevels = New System.Windows.Forms.ToolTip(Me.components)
+        Me.ucrReceiverInitialColumnFactor = New instat.ucrReceiverMultiple()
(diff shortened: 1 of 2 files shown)


rdstern (changes requested):
@Vitalis95 and @lilyclements this already works and that's a wow!  I really like it.
a) The one tiny edit is to check the intial order factor levels checkbox by default.  So it should always start checked, and reset should make it checked again.  
b) Initially the checking of that item didn't work in my test version.  That's a temprorary limitation that the original version of the 4 packages is originally attached.  So it doesn't find Lily's new function.  Just detach and then attach (library) again and all works.

Vitalis95:
@rdstern , it's now checked by default

PR #9488, opened 2025-03-10
Added find replace/options to script window right mouse button menu

lloyddewit:
Fixes (partly) #9026.

### DIFF
--- a/instat/ucrScript.Designer.vb
+++ b/instat/ucrScript.Designer.vb
@@ -77,6 +77,11 @@ Partial Class ucrScript
         Me.cmdRunStatementSelection = New System.Windows.Forms.Button()
         Me.TabControl = New System.Windows.Forms.TabControl()
         Me.toolTipScriptWindow = New System.Windows.Forms.ToolTip(Me.components)
+        Me.ToolStripSeparator6 = New System.Windows.Forms.ToolStripSeparator()
+        Me.mnuFindNext = New System.Windows.Forms.ToolStripMenuItem()
+        Me.mnuFindPrev = New System.Windows.Forms.ToolStripMenuItem()
+        Me.mnuReplace = New System.Windows.Forms.ToolStripMenuItem()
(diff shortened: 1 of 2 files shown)


rdstern:
Looks great so far.  I note:
a) Already in the right-click menu.
b) When I copy a phrase, e.g. data_book, then the options are enabled.
So far it seems to stay where I have copied, i.e. Find Next and Find Previous don't yet move up or down.  I wonder how Replace will work?

lloyddewit:
@rdstern Thanks for looking at this. This PR is still draft. I'll let you know when it's ready for testing.

lloyddewit:
@rdstern I added find next and find previous functionality. You may test if you wish but it may be more efficient for you to wait for the replace functionality and then test everything together.

lloyddewit:
@rdstern I added the `Replace` and `Replace All` functionality.
All the find and replace functionality should now behave exactly as explained in the Script window right mouse button tool tips. Basically, the search text is the current selection and the replace text is the clipboard.
This should complete step 1 of issue #9026. It is not a very polished interface but it should solve your immediate problem.
I did not yet look to see how trivial it will be to call this functionality from the `Insert` pull-down menu.
Please could you test?
Many thanks

lloyddewit:
@rdstern Thank you for testing. You understood everything correctly.
This PR includes all the back-end functions needed to perform the finds and replaces. I thought it was most efficient for me to write these function because I already have experience with the `Scintilla` library used in the Script window.
I wrote these functions so that they could be re-used by a GUI. But creating the GUI will still be an interesting and challenging task. I think it would be a good task for one of the developers.

> Antoine - could we (easily?) change the name Insert to Edit and add those Find and Find/Replace options also to the pull-down that you added to that button.

@N-thony I thought a bit more about this. If I add it in this PR then it will need to be changed again in the 'part 2' PR. So I agree with @rdstern and I think this should be done in the 'part 2' PR.
Please could you review this PR?
Thanks

PR #9489, opened 2025-03-10
Improvement to Prepare > Column: Factor Dialogues 

### LINKED ISSUE #9451: Make our factor dialogs simpler to use

derekagorhom:
Fixes #9451 
This is the continuation of the improvements done in #9474
I have done the following dialogues;
Recode Factor 
Dummy Variables
Levels/Labels
@rdstern , @N-thony 
can you review this PR 
Thanks


berylwaswa (changes requested):
@derekagorhom this works fine, when I open the dialogs the first factor is already in the receiver and the levels visible. 
However, when I enter a different variable in the receiver then close the dialog, when I come back, the focus is on the variable I had selected before but the first variable is the one in the receiver. Can the dialog remain as I had left it?

derekagorhom:
> @derekagorhom can you pull and test PR #9474 and see the issue is also there?

@N-thony I did two updated branches and i still have the same problem

derekagorhom:
> > @derekagorhom just to second the testing by @berylwaswa. a) When you return to the dialog, which had a factor, then it remembers the way it was left. b) The rule for filling with a factor, if that has the focus, or the first factor otherwise applies only after Reset was pressed, (which empties the receiver), and this is presumably the same as the first time the dialogue is used.
> 
> @derekagorhom this is how the dialogue is expected to work.

@N-thony I have been trying since yesterday, i do not know how it works on yours and not on mine, perhaps someone can try and different PR and see if it works

derekagorhom:
#9492 replaces this PR

PR #9499, opened 2025-03-25
Table options improvement

N-thony:
@Vitalis95 after our meeting with @lilyclements, I have made some changes here. @lilyclements we realized that can't use cols_label instead we had to use the rename function from dplyr, I think your approach was good but missing something. This is now fixed and implemented here.
Can you pull the changes here and continue as discussed few minutes ago?


lilyclements:
@N-thony interesting! What was wrong with using `cols_label` in the end? And how did you fix using `dplyr::rename`? Nice one

rdstern:
@Patowhiz are you able to check the code in this one, or do we wait till Monday?  I'd like to look at all @N-thony pull requests before next Wednesday.

rdstern:
@N-thony and @Vitalis95 I am suggesting closing this pull request and freezing #9522 for now.  Neither seem to be doing very well and the Themes in the Use Table dialog no longer works.
I assumre @vitalis is a later pull request so we don't need both?  Am I right?

PR #9502, opened 2025-03-26
Improved the Count in Factor Dialog

### LINKED ISSUE #9447: Improve the count in factor dialog
@N-thony this should be a very small amount of work,  you may want to check yourself, because one of the points has come up before in other dialogs, so I am hoping for a general fix.

Here is the current dialog:

Image

a) Note the selector presumably should say Factors or something, but is very truncated.  We previously had a similar issue with a receiver.
b) This is one of a number of examples where it would be simpler if a variable, by default, is copied into the receiver.  Use the first factor.  If another column is needed -  it is no further further work - because it overwrites.  

Note that it is particularly when the contents are just factors, or dates, because there are often relatively few of them in the data frame.  We often copy the first date variable into the receiver, and let's do the same for the first factor, when convenient.

MeSophie:
Fixes #9447
@rdstern @N-thony  I made some changes on Count in Factor Dialog. Please have a look.


rdstern (approved):
@MeSophie thanks that works in that I can see the title fully now
@N-thony I am approving, but I am concerned that this may overlap with the work you and @derekagorhom have been doing on the factor dialogs.
I note that Sophie has included a simple addition of the first factor into the receiver.  Her "solution" is less comprehensive than you have, sn getting the order right on merging could be important.

rdstern (commented):
@N-thony and @MeSophie can this be closed now that we have a more recent one to merge?

PR #9495, opened 2025-03-18
Hiding The Structured Menu

derekagorhom:
Fixes Partly #9268 
For this PR, i tried implementing the hide option for the Structured menu only. When you start R-instat, the Structured menu should be hided as usual when you click on the structured menu click in the dataview buttom it should appear.
When you close the software with the structured menu still clicked and restart the software, the structured menu will appear with the other existing menus.
image
@rdstern , @N-thony can you review this PR thanks

PR #9504, opened 2025-03-27
Hiding Structured and Procurement Menus upon opening

N-thony:
Fixes (partly) #9268
Replaces #9495

@rdstern @derekagorhom have a look of this if it does what is expected.

### DIFF
--- a/instat/clsInstatOptionsDefaults.vb
+++ b/instat/clsInstatOptionsDefaults.vb
@@ -18,10 +18,10 @@ Public Class clsInstatOptionsDefaults
     Public Shared ReadOnly DEFAULTbIncludeRDefaultParameters As Boolean = False
     Public Shared ReadOnly DEFAULTbCommandsinOutput As Boolean = False
     Public Shared ReadOnly DEFAULTbIncludeCommentDefault As Boolean = True
-    Public Shared ReadOnly DEFAULTbShowProcurementMenu As Boolean = True
+    Public Shared ReadOnly DEFAULTbShowProcurementMenu As Boolean = False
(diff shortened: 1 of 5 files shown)


derekagorhom:
@N-thony i tried this PR and the structure menu was hidden when i first opened the frmmain but the procurement and option by context menu was still on
image
maybe @rdstern can confirm this

rdstern:
@N-thony this is timely for a topic I mention below.  I have 2 trivial cosmetic suggestions:
a) I first tried with the View Menu.  That seems to work fine, but please could you change Experiments to `Experiments Menu` to make it consistent with the other special menus.
There it seems to work fine.  I made some changes and then re-started R-Instat and it correctly started as I left it.  
b) Then I wnet to look at the Tools ? Options menu.  Do we need both these options?  They are just an alternative way of doing the same thing.  Assuming we keep both, then please change Options by Context to `Experiments` to make it consistent with the View menu.

Then, I assume you may want to merge quickly, but I take the opportunity to mention the imminent addition of another menu, called `Tricot`, which is to be added and (I assume) just after the climatic menu.  So, I hope the code will permit more menus to be added easily.

I report also for @derekagorhom as this new Tricot menu is very likrly to be linked to a forthcoming R-Instat sprint at the end of April, in Kenya.  Expect more details by Monday!

rdstern (approved):
@N-thony many thanks I am approving.
@Patowhiz could you please check and hopefully merge.  I would like to proceed as soon as possible with the new (tricot) menu - that's the main impetus for the sprint.  This needs to be merged first.

Issue #9505, opened 2025-03-31
New tricot menu in R-Instat


rdstern:
I outline the structure of the proposed menu.  Then @lilyclements will enlarge and correct me on it.  @lloyddewit is also interested to know the structure of the menu and dialogs.

a) The menu is called `Tricot` and comes after the Climatic menu.
b) I suggest, once the minor changes in #9504 are completed and merged, that the `Experiments` menu should swap with the Procurement menu, so it comes straight after the Tricot menu.
c) The Tricot menu should be added to the View Menu, and the Tools > Options menu, in the new order and should behave identically to them.  The "Factory Setting" remains that Climatic is visible and all the other menus are hidden, including Tricot.

d) Now the provisional structure of the Tricot Menu.  
1) There are about 4 initial dialogs, including `Define Tricot Data...` .  They are probably direct dialogs, i.e there is no need for a sub-menu.  Lily will add the names and correct if this is wrong.
2) There is a Line - just like the end of the Prepare section in Climatic.
3) Then there are Describe and Model sub-menus.
4) In the Describe sub-menu there are 2 dialogs - Lily to add the names.
5) In Model there are 3 dialogs - Lily to add the names.

Once we have confirmation on the structure and the names I further suggest all the dialogs are started as follows:
1) They have our `usual selector`.
2) They have our usual 5 `bottom buttons` and the Comment control.
3) They have a Dialog Under Construction note in the middle and Ok is disabled. 
4) They each have a help context number - given by Roger - once the names or each dialog is known, see below.

The help entries are constructed:
I will make the same structure in the R-Instat help file, once the structure of the menu is confirmed and the names of each dialog is given.  Then I can report the help context numbers, so the help button will connect properly.

The R-Instat scripts are added the the library.
@lilyclements has R scripts adapted for R-Instat.  Once she is happy with them, a copy will be zipped and included here.  They will then either be added to the Scripts directory in our library.  Or they could go into a Tricot sub-directory in the script sirectory - Lily to determine.

lilyclements:
@rdstern I've started writing up [here](https://docs.google.com/presentation/d/1ppsQ-jfVYWwCp-fxZN8OnuGCg3Lrt-RRwejpGVXwYwY/edit#slide=id.g346728dc158_0_10) for the menu design and dialogs. I will do a separate issue for each of the dialogs and give the R code, but just as an overall view of it I'm jotting bits down there for now. 

@rdstern what do you think of the following names for the new dialogs:

[...]

PR #9506, opened 2025-03-31
Setting up new objects for Tricots menu

lilyclements:
Adding a new object (ranking and grouped_ranking) for tricots.

@N-thony do you mind taking a look at this to check I've added bits in the correct place? You're much more of a wizz at this stuff!

### DIFF
--- a/instat/Enum/RObjectTypeLabel.vb
+++ b/instat/Enum/RObjectTypeLabel.vb
@@ -12,6 +12,6 @@ Public Structure RObjectTypeLabel
     Public Const Graph As String = "graph"
     Public Const Table As String = "table"
     Public Const Model As String = "model"
-    Public Const StructureLabel As String = "structure" ' e.g Survival objects
+    Public Const StructureLabel As String = "structure" ' e.g Survival, ranking, grouped_ranking objects
     Public Const Summary As String = "summary" ' e.g objects of text formats
 End Structure
--- a/instat/clsRCodeStructure.vb
+++ b/instat/clsRCodeStructure.vb
@@ -168,6 +168,8 @@ Public Class RCodeStructure
     ''' <param name="strTempModel">                 (Optional) The new value for the model. </param>
     ''' <param name="strTempGraph">                 (Optional) The new value for the graph. </param>
     ''' <param name="strTempSurv">                  (Optional) The new value for the surv. </param>
+    ''' <param name="strTempRanking">               (Optional) The new value for the ranking. </param>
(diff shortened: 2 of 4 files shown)


lilyclements:
@Patowhiz @N-thony - in R, I would run the following line to save the object:

```
data_book$add_object(data_name="breadwheat_by_ID_variety", object_name="rankings_list", object_type_label="structure", object_format="text", object=rankings_list)
```

Are these changes the correct changes to facilitate that? The R code works, but, after @Patowhiz's comment in #9505 this is definietly somewhere where I would like you to let me know if this is the right direction

rdstern:
@lilyclements I assume @Patowhiz is the right person to do the second approval and merging.  

I query whether - while you are on a roll with new objects - it might be a quick step for you to add the calculation or definition objects that you produce for e-picsa?  That would be a separate pull request, but might be useful to have in place before the sprint?

Patowhiz:
@lilyclements That’s why I asked what makes the ranking objects special or different from summaries.

The original idea behind object types was to distinguish what’s possible with each type. For example:
- You can edit and reuse graph and table objects.
- You can apply a model object to datasets, etc.

Based on that, we left most non-summary objects under the "structure" object type. At the time, I didn’t fully understand the purpose of survival objects—none of the survival dialogs were working—so I left them as "structure" as well (I think I consulted you about that too).

If you've added a new "rankings" object type, then your `object_type_label` should be `ranking`. If not, it depends on how you’ve classified the ranking objects.

On a separate note, @rdstern I also made a distinction between data objects and output objects. I considered things like filters, definitions, and column selections as data objects. One object I had hoped to also classify as a data object was shapefiles. Currently, they’re imported as data frames, but that never felt quite right. I always thought that shapefiles should be treated as data objects of their own type, but that would have required rethinking how R-Instat mapping dialogs work, something I didn’t pursue at the time. Still, it would have greatly improved the experience, especially around the `reogrid` issues you’ve faced. Most importantly, it would have enhanced user experience considering shapefiles and their close cousin GeoJSON, which I now work with frequently, were never meant to be used or viewed as plain tables.

lilyclements:
@Patowhiz thanks for clarifying. So this is like an object which is used in modelling. I think it makes sense under structure now I have seen the definition.
This looks like it might mean I don’t need to set up anything new then - instead I’ll save it as a structure type.

For us to save a structure object in the ucrSave, I guess I don’t need to add any new code? In which case I am happy to close this PR. Thanks.

Issue #9513, opened 2025-04-03
Tricots Dialog: Modelling (no covariates)


lilyclements:
This is our first modelling dialog in our tricots menu! It is called "Placket-Luce Model (No Covariates)"

A nice and simple main dialog:

![Image](https://github.com/user-attachments/assets/5d2a8bc4-54af-420d-b489-300309de58e9)

We have:
1. A data selector, which here is "breadwheat_by_ID_variety" in our example
2. A multiple receiver, which in the R code below is assigned to be called `vars`:
`vars <- c("lastassessment_overallperf", "firstassessment_earlygrowth", "lastassessment_grainquality")`
3. There's then two buttons leading to sdgs, which I'll outline in my next message
4. And a save option, to save the model, like in our other modelling dlgs

```r
get_index_names <- data_book$get_variables_metadata("breadwheat_by_ID_variety")
rankings_object <- data_book$get_object(data_name="breadwheat_by_ID_variety", object_name="rankings_list")$object
vars <- c("lastassessment_overallperf", "firstassessment_earlygrowth", "lastassessment_grainquality")
rankings_object <- get_ranking_items(data = get_index_names, vars_to_get = vars,
                                       index = "rankings_index", rankings_object)
mod <- purrr::map(.x = rankings_object, .f = ~PlackettLuce::PlackettLuce(.x))
names(mod) <- vars
```

lilyclements:
Display Options:

This is how the current display options looks (and I've added in this some checkboxes we want to add)

![Image](https://github.com/user-attachments/assets/a5dad44c-1ed3-45dd-bdeb-6238e91c1926)

Under Display, we want all these options. We additionally want a new checkbox next to "Estimates" stating `log`, default unchecked.

We then have eight more checkboxes below it
- AIC
- Deviance
- Coefficients
- Pairwise Probabilities of Coefficients
- Reliability
- itempar
- Variance-Covariance Matrix
- qvcalc of itempar

For Reliability and Plot Log-worth, the additional also must be run:

```
breadwheat_by_ID_variety <- data_book$get_data_frame("breadwheat_by_ID_variety")
variety_baseline <- levels(factor(breadwheat_by_ID_variety$variety))
```

| Checkbox | R Code | Notes |
| -----------| ------------------------------------------------------------------------------------------------------| -------------------|
| Summary | `purrr::map(.x = mod_list, .f = ~summary(.x))` | |
[...]

PR #9514, opened 2025-04-03
Added Tricot Menu

N-thony:
Fixes (partly) #9513
@rdstern, I have added the Tricot menu and sub menus as requested in the corresponding issue. Have a look.


rdstern:
@N-thony this is great.
a) The tricot menu itself looks exactly what we need.  @lilyclements do you agree?
b) Here is a picture of the view menu:

image

1) The symbol for Tricot and even the tick looks slightly different to the others - and the shaded bit on the left is slightly wider.  
2) Could you please change the order, so Experiments comes straight after Tricot, both here and in the Tools > Options dialog.

This is an excellent start though!

rdstern (commented):
@Patowhiz I have also checked and am happy to approve also.  But I would still like you to check the code and then do the second approval and merge.  This is the menu that's giving us the sprint.  I think it is a very small task for you?

PR #9531, opened 2025-04-14
Fixed feature of shortening names in multiple receivers not being consistent

### LINKED ISSUE #9497: Fix feature of shortening names in multiple receivers
Here is an example:

Image

I think this is general and is am consequence of the code that added the count - here it says there are 2 variables in the control.  And the names don't go past that point.  It occasionally fixes itself!

So, here in a copy of the data frame, with an even shorter name, and it is fine:

Image

And now, if I return to the old data frame it is now fine too (with this dialog)

Image

But, in a new dialog it is not fine:

Image

And, in some dialogs it seems not to fix itself so easily?

Image

derekagorhom:
Fixes #9497 
This should fix the problem of the shortening names feature not being inconsistent in the multiple receivers
Thanks

### DIFF
--- a/instat/ucrReceiverMultiple.vb
+++ b/instat/ucrReceiverMultiple.vb
@@ -524,12 +524,6 @@ Public Class ucrReceiverMultiple
         CheckSingleType()
     End Sub
 
-    Private Sub lstSelectedVariables_ClientSizeChanged(ByVal sender As Object, ByVal e As EventArgs) Handles lstSelectedVariables.ClientSizeChanged
-        If lstSelectedVariables.Columns.Count > 0 Then
-            lstSelectedVariables.Columns(0).Width = lstSelectedVariables.ClientSize.Width
-        End If
-    End Sub
-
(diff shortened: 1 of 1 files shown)


rdstern:
@derekagorhom can you point me towards datasets and dialogs that have been changed by this.  Here is the first example I looked at.  You can see the names are quite long and I am not clear they need to be shortened so much in this particular multiple receiver?

image

derekagorhom:
@rdstern 
Please what length is ok for variables, because the shorten function set a limit of 10 for each variable

rdstern:
@derekagorhom that's interesting and useful.  I note that each full name is shown when you hover over any entry - and that's useful.  I am keen to avoid a bottom scroll, which is presumably what is achieved by the shorten function.  Can the number of characters be related to the width of the receiver - say 4 less, so there is room for the dots.  Otherwise I assume it could be a bit larger, e.g. 12 and still ok in each multiple receiver?

rdstern (approved):
@derekagorhom thanks, I am accepting!
@N-thony over to you

PR #9536, opened 2025-04-22
Improvement to Count in Factor Dialog

### LINKED ISSUE #9447: Improve the count in factor dialog

derekagorhom:
Fixes #9447 
Replaces #9502 
This includes the new utility class for columns selection for the factor dialogs which is also implemented here 
Thanks


rdstern (approved):
@derekagorhom and @MeSophie this now looks fine.
@N-thony can you please check and merge if possible.

lilyclements (approved):

Issue #9400, opened 2025-01-29
Adding "Import from ClimMob" for Tricot Data Analysis


lilyclements:
This is related to a dialog to "Import from ClimMob" for our new set of Tricot Analyses menu:

![Image](https://github.com/user-attachments/assets/4c0717ea-955b-4beb-8972-19b4d6dfdb7c)

1, "Server Name" runs the parameter `server`. This has a drop down list of the four server names that this variable takes: `climmob3`, `1000FARMS`, `AVISA`, `RTB`. `server = "climmob3"` is the default
2. API Key runs the parameter `key`. This can have a button next to it where you browse .txt files for the stored key in a .txt file. This is similar to how we get the key in the `Import from OpenAppBuilder` and `Import from RapidPro` dialogs

```
key <- read.table("path/to/key.txt", quote="\"", comment.char="")
```

3. The "Find forms" button runs the R code `getProjectsCM(server = <server name>, key = <key>)`. This R code gives a table, and we just want to return in the drop down list the `project_id` column in this list. I assume this is similar to how the "Import from ODK" dialog works.
4. There is a checkbox, default unchecked:

a) When you click OK, the following runs:

```r
user_owner <- project_list %>% dplyr::filter(project_id == project) %>% dplyr::pull(user_owner)

ClimMobTools::getDataCM(key = <key>
                        project = < project ID selected in the drop down >,
                        userowner = user_owner,
                        server = <server>,
                        pivot.wider = TRUE)
```

b) We get one data frame, which is called the name given in the ucrSave after "Data Frame:"

**R Code Example:**

```r
# get a list of projects
key <- "d39a3c66-5822-4930-a9d4-50e7da041e77"
# or in our case we will store the key in a .txt file (without the quotes) and call by:
# key <- read.table("C:/Users/lclem/Downloads/tricot_api_key.txt", quote="\"", comment.char="")
project_list <- ClimMobTools::getProjectsCM(key = key, server = "climmob3")

# display the project_list$project_id in the list
```

Second bit that runs gets the data from ClimMob
```r
# Let's say they choose the project titled "breadwheat":
# and that the name in the ucrSave is "breadwheat"
project <- "breadwheat"
user_owner <- project_list %>% dplyr::filter(project_id == project) %>% dplyr::pull(user_owner)
breadwheat <- ClimMobTools::getDataCM(key = key,
                             project = project,
                             userowner = user_owner, # we run this ourselves
                             server = "climmob3",
[...]

PR #9538, opened 2025-04-23
Implementing Import from ClimMob dialog

derekagorhom:
@lilyclements here is the dialog, it corresponds to issue #9400
Please can you have a look
Thanks


instat/dlgImportClimMob.vb
```diff
@@ -0,0 +1,200 @@
...
+
+        clsSecondOperator.SetOperation("%>%")
+        clsSecondOperator.AddParameter("left", clsRFunctionParameter:=clsProjectsFunction, iPosition:=0)
+        clsSecondOperator.AddParameter("right", clsROperatorParameter:=clsFourthOperator, iPosition:=2)
+        clsSecondOperator.bSpaceAroundOperation = False
+        clsSecondOperator.SetAssignTo("user_owner")
+
+        clsThirdOperator.SetOperation("==")
+        clsThirdOperator.AddParameter("left", "project_id", iPosition:=0)
+        clsThirdOperator.AddParameter("right", "project_name", iPosition:=1) ' right = value in the ucrInputBox
```
lilyclements, line 96: @derekagorhom I've made some changes here to the code, and it is nearly there. The only bit remaining is for this value here where it says "project_name" to just be the value that is selected in the ucrInput box.
It's got pretty late here so I'll leave this for the day - are you OK to have a look at this?

We do this in the "Import from ODK" dialog, which might be useful to check how it is done there for an idea.

Can you also make sure you pull the changes from here into your branch? There's been a few other changes made which have helped some of the other bits!

lilyclements (commented):
@derekagorhom this should now work! However, you should check it, I haven't properly stress tested it, but it's at least reading in the correct bits now

instat/dlgImportClimMob.vb
lilyclements: @derekagorhom you can ignore that message - it's all sorted now

rdstern (commented):
@lilyclements I am not sure how to test this properly, but happy to approve as soon as you want to get something in the menu.
I assume it corresponds to issue #9400, so have added that to the pull request.
I note there isn't yet a button for the sub-dialog?

lilyclements:
@rdstern yes I am talking to @derekagorhom about it at the moment. He's working on the second sdg now, and we've also discussed some other bits on the main dialog.

To test it, if you were interested, at the moment, then I have shared a token and video with you via email

lilyclements (dismissed):

PR #9542, opened 2025-04-28
Small improvemens to the Two/Three Summarise dialog

### LINKED ISSUE #9530: Two small improvements with the Describe > Two/Three variables > Summarise dialog
This is an important dialog.  There is already one improvement to fix a bug in the ANOVA, which is #9503 .

These suggestions are cosmetic and could be combined, or be separate.  There will be further improvements once the tables stuff is further along.

a) The first is shown here.  In the 2-variable case there is an option to swap the first and second.  It works fine, but when you do, the Display Margins Checkbox on the left remains visible.  When there is the swap it should be hidden.  

Image

b) When you move between the top buttons the receivers clear.  Here is an example:

Image

If I now move to two variables, then it clears to this:

Image

It would be nice if the first 2 receivers remained.  And if I were to return the the 3-variable case, then it would be ideal if it still remembered that third variable.

Note there is one potential problem, namely that the first option, of Skim, allows a mix of data types in the multiple receiver, and the other buttons allow only one data type.  In that situation, I don't mind, it could clear the multiple receiver, if that is the simplest option for the coding.

Emmanuel-Afrifa:
Fixed #9530 
@rdstern, this is the fi for the reported issue

### DIFF
--- a/instat/dlgDescribeTwoVariable.vb
+++ b/instat/dlgDescribeTwoVariable.vb
@@ -173,6 +173,7 @@ Public Class dlgDescribeTwoVariable
         ucrChkSwapXYVar.SetText("Swap First/Second Variables")
         ucrChkSwapXYVar.AddParameterValuesCondition(True, "var", "True")
         ucrChkSwapXYVar.AddParameterValuesCondition(False, "var", "False")
+        ' This controls the visibility of the display margins check box
         ucrChkSwapXYVar.AddToLinkedControls({ucrChkDisplayMargins}, {False}, bNewLinkedHideIfParameterMissing:=True)
 
         ucrPnlDescribe.AddRadioButton(rdoTwoVariable)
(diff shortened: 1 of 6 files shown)


rdstern:
Afre @berylwaswa has approved this is a good task for @Patowhiz to show the developers how he checks the code.

berylwaswa:
@Emmanuel-Afrifa 
This still doesn't solve the first problem of display margins disappearing when I check to swap the first and second variables.

instat_5miycgJKcP

The second improvement needed was that the first receiver does not clear when you move between the top buttons, but that doesn't happen either. 

I am in the two variables.

instat_agyHsn4PFq

Then I move to the three variables

instat_7rhyL4H6FB

And now I'm in the three-variable

instat_6AOOqG2C8l

And moved to two variables

instat_C0Ymi16wfZ

PR #9547, opened 2025-04-28
Improvements to hide the display margins checkbox and retain the selected columns

### LINKED ISSUE #9530: Two small improvements with the Describe > Two/Three variables > Summarise dialog

Emmanuel-Afrifa:
Fixes #9530

### DIFF
--- a/instat/dlgDescribeTwoVariable.vb
+++ b/instat/dlgDescribeTwoVariable.vb
@@ -173,6 +173,8 @@ Public Class dlgDescribeTwoVariable
         ucrChkSwapXYVar.SetText("Swap First/Second Variables")
         ucrChkSwapXYVar.AddParameterValuesCondition(True, "var", "True")
         ucrChkSwapXYVar.AddParameterValuesCondition(False, "var", "False")
+        ' This controls the visibility of the display margins check box
+        ucrChkSwapXYVar.AddToLinkedControls({ucrChkDisplayMargins}, {False}, bNewLinkedHideIfParameterMissing:=True)
 
         ucrPnlDescribe.AddRadioButton(rdoTwoVariable)
         ucrPnlDescribe.AddRadioButton(rdoSkim)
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@Emmanuel-Afrifa great, that seems to work fine now, both deleting the Margins and also keeping the variables.  I am approving.
@Patowhiz this is an example to demonstrate checking the code.  Then merging!

lilyclements:
@Patowhiz are you happy to approve after looking this over with the others?

instat/dlgDescribeTwoVariable.vb
```diff
@@ -1005,10 +1007,10 @@ Public Class dlgDescribeTwoVariable
...
 
     Private Sub ucrPnlDescribe_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlDescribe.ControlValueChanged
-        ucrReceiverFirstVars.Clear()
-        ucrReceiverThreeVariableSecondFactor.Clear()
-        ucrReceiverThreeVariableThirdVariable.Clear()
-        ucrReceiverSecondTwoVariableFactor.Clear()
+        'ucrReceiverFirstVars.Clear()
+        'ucrReceiverThreeVariableSecondFactor.Clear()
+        'ucrReceiverThreeVariableThirdVariable.Clear()
+        'ucrReceiverSecondTwoVariableFactor.Clear()
```
Patowhiz, line 1013: @Emmanuel-Afrifa I note that you have commented out the previous code. Why not just remove it entirely?

Patowhiz (approved):

Issue #9518, opened 2025-04-07
Tricots Dialog: Modelling (tree)


lilyclements:
### Design - Main dlg

This is taken from the dlgFitModel, where I've made a few minor changes in this design

![Image](https://github.com/user-attachments/assets/12d03927-0eef-4a98-9ecd-b54d9fe8366c)

We have:

* A multiple receiver, which in the R code below is assigned to be called vars:
`vars <- c("lastassessment_overallperf", "firstassessment_earlygrowth", "lastassessment_grainquality")`
* An explanatory variable box to give the explanatory variables in. The user will most definitely use a different data frame to what is used to specify the vars, so note that we want to call both data frames.
* There's then two buttons leading to sdgs, outlined below
And a save option, to save the model, like in our other modelling dlgs

```r
# vars takes the value of the variables in the "Traits" receiver
# breadwheat_by_ID_variety is the value of the df in the selector for the traits
get_index_names <- data_book$get_variables_metadata("breadwheat_by_ID_variety")
rankings_object <- data_book$get_object(data_name="breadwheat_by_ID_variety", object_name="grouped_rankings_list")$object
vars <- c("lastassessment_overallperf", "firstassessment_earlygrowth", "lastassessment_grainquality")
rankings_object <- get_ranking_items(data = get_index_names, vars_to_get = vars, index = "rankings_index", rankings_object)

# We need to get the variable "variety" for the reference level. This here uses the Data Frame used to get the trait variables above.
breadwheat_by_ID_variety <- data_book$get_data_frame("breadwheat_by_ID_variety")
var_name <- data_book$get_variables_from_metadata(data_name = "breadwheat_by_ID_variety", property = "Tricot_Type", property_value = "variety")
variety_baseline <- levels(factor(breadwheat_by_ID_variety[[var_name]]))[1]

# get the X variables.
# call their data frame. This data frame will be different to the one called above.
breadwheat <- data_book$get_data_frame("breadwheat")
plfit <- PlackettLuce::plfit # due to a bug in their code, we need to run this for now.

mod_list <- purrr::map(.x = rankings_object,
                       .f = ~PlackettLuce::pltree(.x ~ gender,  # If gender is the only variable in the explanatory receiver
                                    data = breadwheat,   # the data frame associated with teh variables in the explanatory receiver
                                    ref = variety_baseline))
names(mod_list) <- vars
```

### Design - Model Options

[...]

PR #9560, opened 2025-04-30
Added the tricot dialog: Modelling (tree)

Emmanuel-Afrifa:
Fixes Partly #9518 
@rdstern, @lilyclements Kindly review the modelling (tree) tricot dialog.

### DIFF
--- a/instat/dlgModellingTree.Designer.vb
+++ b/instat/dlgModellingTree.Designer.vb
@@ -53,6 +53,8 @@ Partial Class dlgModellingTree
         Me.ucrSelectorByDataFrameAddRemoveForModellingTree = New instat.ucrSelectorByDataFrameAddRemove()
         Me.ucrBase = New instat.ucrButtons()
         Me.UcrSelectorByDataFrameForModellingTreeSecond = New instat.ucrSelectorByDataFrameAddRemove()
+        Me.btnCheck = New System.Windows.Forms.Button()
+        Me.ucrInputCheck = New instat.ucrInputTextBox()
         Me.grpSecondCalc.SuspendLayout()
         Me.grpFirstCalc.SuspendLayout()
         Me.SuspendLayout()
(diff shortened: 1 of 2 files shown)


lilyclements:
Can you add one new feature. 

It will work similarly like the "Check()" feature in [here](https://github.com/derekagorhom/R-Instat/blob/339fc5b140e1b19823896eecfe2afa4e14b693c9/instat/dlgTransformTricotData.vb)

1. Update your `databook` R package (you should do this in R, not R-Instat)
2. Add in a new button and input into the dialog, above the ucrSave, where the button says "Check"
3. Add in the code! Have a look at the `  Private Sub Check()` in this R code [here](https://github.com/derekagorhom/R-Instat/blob/339fc5b140e1b19823896eecfe2afa4e14b693c9/instat/dlgTransformTricotData.vb) for how to do it. But you need to change the error messages. You won't need Case Else, so you can ignore that. 
4. Also add in for the `ucrInput....` receiver to clear when you press Reset.  (So add this line into the Defaults sub: `ucrInput.....SetName("")`)
5. And add in for your `Check()` sub to run when you press it, e.g., if it's called `cmdCheck`, then you'd run:

```r
 Private Sub cmdCheck_click(sender As Object, e As EventArgs) Handles cmdCheck.Click
        Check()
    End Sub
```

lilyclements (dismissed):
Approving - very good @Emmanuel-Afrifa! @rdstern over to you!

The sub-dialogs are still in progress so one of the sub-dialogs is not yet created. This is being done by Sophie as it is used for another modelling dialog

rdstern (approved):
@Emmanuel-Afrifa looks great.  I am approving so it can be merged.
a) Then a small change is that there are 2 selectors, so please could both receivers be turquoise.  If needed I can show another dialog that has this feature.
b) I then tried display options and a plot and it gave errors.  Could you please check this?

lilyclements:
@rdstern for the plots, that is currently being worked on in the sdg by Sophie in another PR. The plan is to link the plots in here later (tomorrow or later tonight)

PR #9574, opened 2025-04-30
Improvements to R Packages Dialog

Ag-Derek:
I do not know what issue this fixes;
This PR solves the problem with the R Packages dialog not properly installing packages from IDEMSInternational and also added the 4 new packages which were created by @lilyclements in the Packages dropdown.
This PR also has removed some commented code which was found in PR #9547.

@lilyclements , @rdstern this is ready for review.
Thanks

### DIFF
--- a/instat/dlgDescribeTwoVariable.vb
+++ b/instat/dlgDescribeTwoVariable.vb
@@ -1007,10 +1007,7 @@ Public Class dlgDescribeTwoVariable
     End Sub
 
     Private Sub ucrPnlDescribe_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrPnlDescribe.ControlValueChanged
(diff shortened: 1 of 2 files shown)


lilyclements:
If the receiver contains: 

```
"ggplot2", "gt", "dplyr", "tidyr", "purrr", "instatExtras", "instatClimatic", "databook", "instatCalculations"
```

then can we run `library( <that package>)` afterwards?

lilyclements:
@derekagorhom I fixed the bug - can you update your `instatExtras` package, it should be in there. Can you change the function to `detach_package` from `package` too 
and then there's one more thing I noted above

Ag-Derek:
@lilyclements i have made the changes

lilyclements:
@rdstern over to you

PR #9577, opened 2025-04-30
Transform Tricot Data

lilyclements:
Fixes https://github.com/IDEMSInternational/R-Instat/issues/9515
Replaces https://github.com/IDEMSInternational/R-Instat/pull/9544
This PR includes the Main dialog and sub-dialog for the Transform Tricot dialog.
Work by @derekagorhom


lilyclements (commented):
@rdstern this is ready to review

rdstern:
@derekagorhom I first tried with the breadwheat.  Here is what it said:

image

I tried making the id into a factor and this made no difference.

I then tried with cassava and got the same message.

The sub-dialog looks good.

rdstern (approved):
@lilyclements It didn't allow me to add the branch, so perhaps it is already merged.
@derekagorhom I tried with the cassava data from gosset.  It works and looks great.  

I was slightly disappointed that I didn't get any results in the output window?  I think it added the keys and links as well as making the 3 data frames.  If I understand correctly you don't now need the define dialog. Is that true?

lilyclements:
@rdstern can you test again - if you update your `databook` and `instatExtras` package, you should now get a comment in the output window that the rankings object is created, the keys are there, and what is defined

PR #9580, opened 2025-04-30
Adding checks in dlgDefineTricotData

### LINKED ISSUE #9550: Add Checks to Tricot Define Dialog
We want to add in checks for duplicates within each group box on the Tricots Define dialog 

1. In the "ID Level", then we want to check that the ID variable has no duplicates
2. In the "ID-Variety" level then we want to check that the ID-Variety combination has no duplicates
3. In the "Variety" level then we want to check that the Variety variable has no duplicates.

An example of where this has been done is in the Climatic Define

lilyclements:
Replaces #9573
Fixes #9550

@Rockjunior due to some issues in your PR with amendments to the main form, I've moved things to a new branch for now. In the future make sure you keep your branch updated from the master, build your branch from the master, and keep one PR per branch. It's all new stuff, and you've picked up a lot in the last three days so it's very impressive stuff!


lilyclements:
@rdstern this is ready for review

Issue #9561, opened 2025-04-30
Check and improve the climatic >prepare >climdex dialog


rdstern:
a)  One small item is that for Tricot we will be getting the rainfall and temperature indices separately.  Currently the dialog needs all 3 before Ok is enabled.  Please enable Ok, once either (or any) are included. 
b) Try all indices you can with the Tricot data. Start by locating simply those that give errors - I think one of the rainfall indices does.  Then correct the errors
c) Check you get sensible output, the indices should also add to others at the same station level that you may have from other Prepare dialogs.

For data you should use the rainfall data from the R-Instat library.  They have to be tidied and help information, with the data in R-Instat could be used.  Discuss with @berylwaswa 

d) Confierm from the earlier #5367 that all is ok and we were right to close it.

Rockjunior:
This issue has previously been addressed by @lilyclements and Francis but they couldn't manage to crack it. I’ve been advised that you, @jkmusyoka, may be able to assist in resolving it

lilyclements:
@jkmusyoka Error 2 is related to the R function which I will fix.

Error 1 and 3 are occurring because for some reason it is reading in `prec = "r"` and `tmax = "t"` even though our precipitation column we put in is called "rain", and our tmax column is called "tmax"!

rdstern:
I am now hitting pause on the changes we should do to this dialog.  We need to resolve the issues here for Tricot.  But I have now found a problem with the spells and extremes dialogs as well - one that should be relatively easy to resolve.
There seem to be multiple problems with this dialog.  So we should return to them soon.  We must get it working again for general use as well as Tricot.  But, for Tricot, we will also need the same fix as for the spells and extremes dialogs, i.e. to get the spells at a station level - rather than each year or month.  It will only link properly if we add that option.

rdstern:
I have commented in the InstatClimatic that the first steps here are to get the dialog working, as before (Version 0.7.16) with "ordinary data".  Test this with the Dodoma data from the R-Instat library and maybe others.
a) It was ok in Version 0.7.16, both in calculating the indices and in writing them to the yearly, or monthly dataframes.
b) Since moving to our packages (e.g. Version 0.8.4) it seems to run, as before, but does not write to the output data frame.
c) In trying to improve the dialog, we seem to have gone backwards.  Please revert to the previous dialog, and then see how to correct the error, so it writes to the summary data frames - as before.

[...]

PR #9591, opened 2025-05-01
Issues 9561 related about the R code in IntatClimatic

Rockjunior:
@rdstern and @lilyclements
this relates to the issues #9561, despite I haven't been able to finish, but am puting it here so that someone can provide an idea on how to solve the issue, althought am still working on it behind. I'll appreciate if someone shares his idea

### DIFF
--- a/instat/dlgClimdexIndices.vb
+++ b/instat/dlgClimdexIndices.vb
@@ -172,12 +172,65 @@ Public Class dlgClimdexIndices
     End Sub
 
     Private Sub TestOkEnabled()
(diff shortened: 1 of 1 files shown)


lilyclements:
@Rockjunior looks good. Is this just the changes on that all three receivers don't need to be filled? I think this is good, if it is. I think the other issue that you found can be reported in an issue?

lilyclements (approved):
@Rockjunior I'm approving 

@rdstern can you check this please
@Rockjunior can you write up an issue on the other problem you found when on this?

lilyclements (approved):

Issue #9601, opened 2025-05-01
Create Plot for Trees


Emmanuel-Afrifa:
@lilyclements  @Emmanuel-Afrifa Kindly make sure that the plot options under the Graphics tab in the display options sub-dialog of the general tricot modelling dialog works (produces a graph/plot)

- Same issue for the modelling (trees) dialog

lilyclements:
@Emmanuel-Afrifa thanks - has this been resolved? I noticed it's now working for them all.

Emmanuel-Afrifa:
@lilyclements Yes, it has been resolved now.
But the plot function itself, you said you would work on it, hence the plot radio button has been disabled for both the trees and general. Kindly alert me when that's ready so that I can add that as well.

lilyclements:
@Emmanuel-Afrifa perfect. I’ll work on that either today or tomorrow - I have quite a few meetings today since I didn’t have any while we were at the sprint. But if I can get to it today then I will. Thanks!

Emmanuel-Afrifa:
@lilyclements Well noted

lilyclements:
I've finally made this plot - it is in the `instatExtras` package, so you should be able to get it if you update that pakage.

Can you add this in @Emmanuel-Afrifa? Just to have only in the "Trees" sub-dialog, and to be the plot that comes if you click "Trees" for your plot. 

The R code is:

```r
list_of_plots <- purrr::map2(.x=mod_list, .y=names(mod_list), .f=~plot_pltree(.x) + ggplot2::labs(caption = .y))
list_of_plots
last_graph <- patchwork::wrap_plots(list_of_plots)
```

This follows the same structure as `gosset::worth_bar` did, just the function itself is different

lilyclements:
@Emmanuel-Afrifa great! There might be a problem or two in that function - if there is, can you update it to the newest `instatExtras`?

Emmanuel-Afrifa:
@lilyclements I have added the function but when I run that line, I get the error in the image. I have updated my `instatExtras` package too but the error is still persistent.

<img width="960" alt="Image" src="https://github.com/user-attachments/assets/002481b5-4b0d-4598-9061-5d1d98d45d63" />

lilyclements:
@Emmanuel-Afrifa when did you last update it? If there's another update available can you try that?

If there's not, can you try running `library(patchwork)` before trying this code to see if that fixes the issue?

PR #9604, opened 2025-05-02
Fixed Issue with plots not apppearing

Emmanuel-Afrifa:
Fixed Partly #9601 
@lilyclements @rdstern 
Fixed the code to make the graphs show in the output window when a plot type is selected in the Graphics tab in the display options sub-dialog for both modelling dialogs (Trees and General) 

@lilyclements I have made the following observation and I will like your opinion.

So in the modelling (general) dialog, the function that runs when the 'check variety' button is clicked throws an error when you use a different dataset (other than the tricot data). This error occurs in the 'check_variety_data_level' function with the error message;
'Object 'Tricot_Type' not found'. 

This is actually great because the specified dataset does not actually have the 'Tricot_Type' object, but from a UI perspective, is it great?

I think it'll work better the way the 'Check' button works in the Modelling (tree) dialog. (There, it doesn't throw an error instead,  it shows the message 'Cannot get data information.' in the input box next to the button). 

Is the a way to fix the  'check_variety_data_level' function to have the same behaviour?

In the function 'check_variety_data_level', the error occurs in the following command:
'dplyr@@filter(., Tricot_Type == tricot_variety_label)'.

I have also Implemented the auto-check of the model in both the tree and general dialogs (similar to what was done for the one covariate dialog).

Lastly, in configuring the sub-dialogs, I realised that if a certain control is disabled under one dialog and you open another dialog that uses the same sub-dialog but requires the same control to be enabled, it isn't. I have fixed this issue for the tree and general dialogs.
- The approach I used was to explicitly enable/disable all the controls for the corresponding main dialogs (depending on what options were needed in that dialog). I have not checked the no covariate and one co variate dialogs by @MeSophie and @anastasia-mbithe,  but I recommend a similar procedure (or a better way around it. That's if they face the same issue in their dialogs).

### DIFF
--- a/instat/dlgModellingTree.Designer.vb
+++ b/instat/dlgModellingTree.Designer.vb
@@ -53,8 +53,8 @@ Partial Class dlgModellingTree
         Me.ucrSelectorByDataFrameAddRemoveForModellingTree = New instat.ucrSelectorByDataFrameAddRemove()
         Me.ucrBase = New instat.ucrButtons()
         Me.UcrSelectorByDataFrameForModellingTreeSecond = New instat.ucrSelectorByDataFrameAddRemove()
-        Me.btnCheck = New System.Windows.Forms.Button()
         Me.ucrInputCheck = New instat.ucrInputTextBox()
+        Me.lblCheckVareity = New System.Windows.Forms.Label()
         Me.grpSecondCalc.SuspendLayout()
         Me.grpFirstCalc.SuspendLayout()
         Me.SuspendLayout()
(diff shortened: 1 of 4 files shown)


lilyclements:
@Emmanuel-Afrifa thanks for outlining this!

1. I'll check: Plots display in trees
2. Good point - I like it. I think instead I should throw an error that says "Data is not Tricot Defined. Use Define Dialog", I can make this change in both functions on my end. I'll do this shortly, 
3. I'll check: autocheck in trees and general
4. I'll check: Thanks for outlining this, I'll check if it's the case for them, and I'll write an issue if it is

lilyclements:
@Emmanuel-Afrifa great find on the bug too - I've added a PR for "No Covariates". I need to add for the "One Covariate" - but first I need to review how it currently is (might make sense for me to make the change). That'll be in a couple of hours I think as my plane is waiting for me.

Fixes #9601

Emmanuel-Afrifa:
@lilyclements well noted. Safe travels on your way back

PR #9490, opened 2025-03-13
Improving summary system for ordered factors

Vitalis95:
Fixes partly #9440 

@rdstern @lilyclements @N-thony , allowed proportion and count input to accept factor labels. If a non-numeric value is entered, quotes are  correctly added. Also fixed a bug where selecting quantile automatically activated the percentile checkbox.

PR #9608, opened 2025-05-02
Allowed Proportion and count input to accept factor labels.

### LINKED ISSUE #9440: Correct and improve the calculator and our summary system for ordered factors

bayakarisa:
Fixes #9440 
First in the proportion and count, at the prepare data dialogue we were limited to numeric only and not neither factors or ordered factors and that check needed to be removed.

Its now working for  both ordered factors and factor labels for proportion and count and not limited like before
you just need to update the databook.
collumn_summary
erro

@rdstern  I have implemented what we discussed
@lilyclements kindly review

### DIFF
--- a/instat/sdgSummaries.vb
+++ b/instat/sdgSummaries.vb
@@ -217,8 +217,7 @@ Public Class sdgSummaries
         ucrInputComboPropTest.SetDropDownStyleAsNonEditable()
 
         ucrInputPropValue.SetParameter(New RParameter("prop_value", 6))
-        ucrInputPropValue.SetValidationTypeAsNumeric()
-        ucrInputPropValue.AddQuotesIfUnrecognised = False
+
 
         ucrInputComboCountTest.SetParameter(New RParameter("count_test", 7))
         Dim dctCountTest As New Dictionary(Of String, String)
@@ -231,8 +230,7 @@ Public Class sdgSummaries
         ucrInputComboCountTest.SetDropDownStyleAsNonEditable()
 
         ucrInputCountValue.SetParameter(New RParameter("count_value", 8))
-        ucrInputCountValue.SetValidationTypeAsNumeric()
-        ucrInputCountValue.AddQuotesIfUnrecognised = False
+
 
         ucrChkProportion.SetParameter(New RParameter("proportion_calc", 42), bNewChangeParameterValue:=True, bNewAddRemoveParameter:=True, strNewValueIfChecked:=Chr(34) & "proportion_calc" & Chr(34), strNewValueIfUnchecked:=Chr(34) & Chr(34))
         ucrChkProportion.SetText("Proportion")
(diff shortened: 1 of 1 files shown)


lilyclements:
@rdstern are you OK to review this? I have started, but am getting a few issues. I'm not sure if I am just handling this dialog incorrectly. Thanks

lilyclements:
@bayakarisa yes, but I can't remember what value we should put in there - is it a factor level (e.g., "SABEY") or the variable itself (e.g., "Village"). Does that make sense?

rdstern:
@lilyclements and @bayakarisa it is a factor level, for an ordered factor

lilyclements:
@rdstern great. @bayakarisa can you try with it having a factor level in there when you test? I see in your screenshot example you've put "Variety", but instead can you try with a factor level from variety - e.g., "OLD"

Vitalis95:
@bayakarisa , is this similar to what was done in PR #9490? If so, please you may close the other PR.

lilyclements (approved):
Really nice by @bayakarisa and @Vitalis95. Approving now. @rdstern over to you to review

PR #9629, opened 2025-05-06
Made design changes to the define tricot dialog 

### LINKED ISSUE #9624: Design changes to Tricot Define dialog
From @rdstern in #9611: David wondered whether the Tricot Define dialog was too large. It needs to work on low resolution screens and we do have some dialogs where the key OK line is not visible, I suggest reducing the space between the data frame and the variables in the selectors, making the selectors one variable shorter and generally saving a bit more vertical space where possible.

In the other direction could the Traits multiple receiver be made a bit bigger?

Emmanuel-Afrifa:
Fixes #9624 
@rdstern @lilyclements  I have made some modifications to the design of the define tricot dialog. Kind review if it's okay now.
Thank you.


rdstern (approved):
Thanks @Emmanuel-Afrifa I am approving too.
@lilyclements can you merge

Issue #9636, opened 2025-05-07
Repeated Factors in Column(s) within the Describe > Tables > Summarise Dialog


fran2or:
**Is your feature request related to a problem? Please describe.**
When using the **Describe > Tables > Summarise** dialog, certain tabs generate redundant factor columns due to repeated entries. This results in unnecessary duplication, affecting table readability and efficiency. The issue needs investigation to ensure factors are correctly structured without redundancy as a similar issue was addressed in PR 9592. I  show the two **tab** cases  below and their accompanying results:

**Frequency Tab**

![Image](https://github.com/user-attachments/assets/76462021-f1f2-47ef-8250-486244bc1d76)

_Output Results Highlighting Problematic Column_

![Image](https://github.com/user-attachments/assets/fbaec89e-bae8-4a16-a2bd-529a20fd07d1)

**Summary Tab**
 There are 3 possible output types  that could be generated here with the _summary-variable_  column here being split into two separate columns. I outline them as follows:

_i) One Variable - One Summary_

![Image](https://github.com/user-attachments/assets/dd167371-74fc-4348-8240-abad9b5e0568)

Results show repeated factor  in each of the split two columns.

![Image](https://github.com/user-attachments/assets/1995db4e-dba2-496c-9d70-1c8c5300cc3b)

_ii) Multiple Variables - One Summary_

![Image](https://github.com/user-attachments/assets/ad889c8a-35e7-41d7-9ecf-c3bfeb25b92b)

Results show repeated factor  in summary column 

![Image](https://github.com/user-attachments/assets/0450402c-adea-4e71-b9c8-38bbda1054ed)

_iii) One Variable - Multiple Summaries_

![Image](https://github.com/user-attachments/assets/ee0dc93c-6253-4e43-a968-2a42c41e48ed)

Results show repeated factor  in variable column 

![Image](https://github.com/user-attachments/assets/5e774f32-e6e6-4a34-8fe7-f5ad2915144e)

_iv) Multiple Variables - Multiple Summaries_

![Image](https://github.com/user-attachments/assets/6bfe7db9-9f62-4c11-97f4-be3956e0d781)

Results here seem desirable

![Image](https://github.com/user-attachments/assets/cde1c67a-db1e-484d-ab57-bafb3628660d)

**Describe the solution you'd like**
 Use the similar ideas in PR 9592 used by @bayakarisa  

**Describe alternatives you've considered**
[...]

PR #9639, opened 2025-05-07
Removed repeated Factors in Column(s) within the Describe > Tables > …

bayakarisa:
fixes partly issue #9636 
Removed repeated factors in column(s) of Describe>Tables>Summarise>Frequency  dialogue
Tablesumm
I have fixed it to this 
tablesum
@rdstern @lilyclements  Kindly check to confirm
@fran2or I have implemented what we discussed

### DIFF
--- a/instat/dlgSummaryTables.vb
+++ b/instat/dlgSummaryTables.vb
@@ -221,8 +221,8 @@ Public Class dlgSummaryTables
 
         ' Gt function
         Dim clsGtFunction As New RFunction
-        clsGtFunction.SetPackageName("gt")
-        clsGtFunction.SetRCommand("gt")
+        clsGtFunction.SetPackageName("instatExtras")
+        clsGtFunction.SetRCommand("generate_summary_tables")
 
         ClsTabSpannerDelimFunction.SetPackageName("gt")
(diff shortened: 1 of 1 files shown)


lilyclements:
@bayakarisa this is nice, but @fran2or suggested we just apply this to the "Frequency Tab" for now. This is because in the "Summary Tab" there are some additional bits to account for. Can you apply this just to the frequency tab?

bayakarisa:
@lilyclements This has been applied  to the "Frequency Tab" only as i wait further directives on what to do with the "Summary Tab"  from @rdstern and @fran2or
frequency
frequency1

lilyclements:
@bayakarisa ignore that comment for now, I have made a suggestion in #9636 in changing the R function which means that this function would then work regardless of if Frequency or Summary were checked. Just waiting for a response from @fran2or and if he's happy then I'll implement the function changes.

lilyclements:
@bayakarisa I made the changes to the R function anyway so this should now work for the Summary tab. Can you update your `instatExtras` package, and check?

lilyclements (approved):
@rdstern this is ready to test. This should mean that we don't have a repeated column for summary, variables, or summary-variables (if it's always the same value). You need to update `instatExtras` to get the most recent changes there.

This is working for me, so I'll approve!

PR #9640, opened 2025-05-07
Improved No Covariate Tricot Dialog

### LINKED ISSUE #9610: Tricots Dialog: Amendment in the Display Options for dlgPlackettLuce
Two very minor things - I would do them my end but do not want to create a conflict in case you're working on them

1. For `No Covariate` (dlgPlackettLuce) the following message from @Emmanuel-Afrifa applies:
> In configuring the sub-dialogs, I realised that if a certain control is disabled under one dialog and you open another dialog that uses the same sub-dialog but requires the same control to be enabled, it isn't. I have fixed this issue for the tree and general dialogs.

> The approach I used was to explicitly enable/disable all the controls for the corresponding main dialogs (depending on what options were needed in that dialog).

If you're unsure can you look in the enable/disable in the main dialog for the trees (dlgModellingTree)

2.  Can you add to your `ucrReceiver`: `ucrReceiverNAME.SetTricotType({"traits"})`

MeSophie:
Fixes #9610
@lilyclements @rdstern I have Improved the no covariate tricot dialog. Please have a look.


instat/dlgPlacketLuceModel.vb
```diff
@@ -380,4 +384,10 @@ Public Class dlgPlacketLuceModel
         sdgPLModelOptions.ShowDialog()
         bResetSubdialog = False
     End Sub
+
+    Private Sub ucrSelectorTraitsPL_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSelectorTraitsPL.ControlValueChanged
+        If clsGetObjectRFunction.ContainsParameter("data_name") Then
+            clsGetObjectRFunction.AddParameter("data_name", Chr(34) & ucrSelectorTraitsPL.strCurrentDataFrame & Chr(34))
+        End If
+    End Sub
```
lilyclements, line 392: This crashes it for me. I think I know the fix - I'll sort it now from GitHub

lilyclements (dismissed):
Great fixes @MeSophie! Nice one.

@rdstern this is ready for you to review.

lilyclements (approved):

rdstern:
@MeSophie I tried with my small cochran_bib data and note the following.  I ran using the ranks in each location, and the overall reversed ranks, and even the yields!
a) The default works - I assume that's summary.  Great.
b) I ran with more of your display options and they worked too.  Then @lilyclements I noticed a result I hadn't quite expected - I guess it may be obvious if I understood PlackettLuce?  The results are identical for the overall ranks and the ranks within each block!  Nice!  It also runs with the yields and I have no idea whether that is a ludicrous thing to do!
Of course this doesn't apply to tricot data, but I still think that's interesting!
Now some comments about the sub-dialogs:
c) The Model Options sub-dialog looks messy:

image

1. I am not sure whose this is, but could the vertical distance between the checkboxes be made the same.  The sub-dialog may become smaller.
2. Even the 3 buttons at the bottom are not equally spaced.
3. The Help does not link yet.  @MeSophie Do you have the numbers?  I don't think any of them are there yet, i.e. main dialog and the 2 sub-dialogs.
4. I wonder what the default model options are?  Ideally they would be ticked and (if Normal and Gamma are the only 2 alternatives for a prior, perhaps they should be radio buttons instead).  Can you run with both Normal and Gamma or is that all Bayesian, with the default being neither?

I tried checking everything - is that allowed.  I don't understand a couple of them

image

I ran with that and oh-my-gosh it works!  Wow.
5) The one item I noticed is that I don't get the ANOVA results, etc, though they remain ticked.  That's perhaps because when they came earlier they needed almost 3000 iterations.  I suggest in the maximum iterations you allow up to 10,000 rather than 1000 and maybe go up in steps of 100 with a minimum of 100?

d) The display Options sub-dialog looks excellent.  I tried the options below and they seemed to work.

image

1) When I first came in to the sub-dialog, nothing was ticked.  I assume @lilyclements the default is summary?  If so, then could that be ticked on the sub-dialog when you first enter.  And I wonder (if you then untick all the options do you get no display?  I assume you should.  Or maybe the Summary could be ticked, and also disabled, because it will always be given?!
2) I tried a plot and it seemed to give one and then it disappeared?

lilyclements:
1. @rdstern to the design parts, I will sort that once this is merged
2. @fran2or can you remind us what the help numbers are for the "No covariates" dialog, and its sub dialogs?
3. @rdstern you can run with both gamma and normal. I also don't understand a couple of the options, but, after our workshop I hope that we can modify this as we should understand a bit more.
4. @MeSophie can you set "Summary" to be ticked as the default?
5. @rdstern you should get a display even if everything is unticked, because you can get just the default model object

MeSophie:
@lilyclements I resolved your comment 4) and 5) Please have a look

lilyclements (approved):
@MeSophie great! Looks good. Just waiting to see if we get the help numbers from @fran2or or @Ag-Derek. Otherwise, we can merge and add them in later

Ag-Derek:
@lilyclements , @MeSophie this help code for the model and its sub-dialogs are below
Model no Covariance - 704
sub 1 - 721
sub 2 - 722
it was not specified which code belongs to which sub

lilyclements:
@Ag-Derek thanks! I checked by putting 721 and 722 as help IDs in random places in R-Instat to see where it took me. 

721 - Model Options
722 - Display Options
(704 - main dlg)

@MeSophie can you add these in? Thanks!

rdstern (approved):
@lilyclements and @MeSophie I suggest the help numbers be added later, along with those for the other dialogs in this menu.  I'll write an issue, initially for @rachelkg on this.

PR #9647, opened 2025-05-08
Changes to the Tricots Menu Layout

### LINKED ISSUE #9637: Add Stack and Unstack to the Tricot menu
This is a simple task, but please wait until @lilyclements agrees.

@volloholic is happy with the new extended menu meaning that some tricot analysts can just about use that menu from the start.  So I am now considering day 1, where I would like to process the survey-type questions.

a) Which dataset should we start with?  You supplied breadwheat from India with 45 variables and 493 cases.  I quite like that version, because there are extra variables to check and tidy, compared to the copy in gosset with just 19 variables.  Though it doesn't seem to add anything - except the ID variable, which I would produce.  So I am thinking of useing the gosset breadwheat as my first example.  It will have 20 variables, once I add the ID (on day 1).  Are you ok with this?
b) I would like to include more than one example.  I have looked at the Cameroon/Nigeria dataset, (cassava), with 1000 respondents (600 from Nigeria and 400 from Cameroon.)  That has som nice, small differences, including differences in the 2 countries and 2 extra columns that give reasons for the ranking and are really messy data.  They don't have date of planting, so I assume climatic is not needed for that set.   I suggest I don't include nicabean as well, (which has 2 data frames) , but I quite like the beans example from PlackettLuce, because that has just one trait, but then compares with a local and is nice to  consider more quality control!

Now to the point of this message.  The varieties_a, b and c are an example of a multiple choice variable.  So it is nice to stack the data.  Hence I suggest adding Stack and therefore also Unstack into the Tidy part of the menu.  They go just above the Add (Merge) Columns dialog.  It works very nicely.  I also want to use the Add Key dialog and that is already there.  

If you are happy with that, then this could be the same pull request as the one that changes the position of the Transform dialog.  I copy that [aragraph here:

Rename and change the position of the Transform Tricot dialog.  This largely came from Rachel.  There are now 2 ways to get the data as Tricot, namely using Transform, or Define.  So can we change the name of the Transform dialog to `Transform and Define as Tricot`.  And move it back to the main menu.  And put it just before the Define as Tricot .  (I am also deleting Data from the Define as Tricot, so the names are not too long.)

Rockjunior:
as stated in issue  
Kindly @rdstern and @lilyclements and @derekagorhom  please review

Fixes #9637
@lilyclements and @rdstern
This PR adds  Stack and Unstack to the Tricot menu, renaming changes to R-Instat menu, Can you Review this,
Thanks


lilyclements (dismissed):
@Rockjunior great! Approving now. @rdstern over to you

rdstern (changes requested):
@Rockjunior looks good.
a) Here is your current main menu:
image

The `...` after the name indicates that when you click it goes into a dialog.  So please can you delete the `...` from each of the 3 names when there is a sub-menu, i.e. Examine, Tidy and PlackettLuce.

I am afraid (unless @lilyclements objects) I have 2 more items to go ito the `Tidy sub-menu`, namely `Reorder Factor Levels...` and `Transform Text...`.

They go, in that order, after `Duplicate Rows...`, and above the horizontal line.

The first is Prepare > Column: Factor > Reorder Levels and the second is Prepare > Column: Text > Transform.

We will then merge and ask @rachelkg to amend the help file.  Then there will be another small issue to add the help numbers.

lilyclements (commented):
@rdstern I'll leave this for you to review since you have a much clearer idea on how you want this to look!

rdstern (approved):
@Rockjunior great, that looks perfect to me, and very clean.  I am approving and hope this can be the menu used for the workshop.
Great to have it merged and we can then prepare the new Help menu and then add the help numbers.

rdstern:
@lilyclements can you approve and merge?

lilyclements (approved):

PR #9649, opened 2025-05-08
Removed ellipses from menu names with sub-menus and add new items to Tidy sub-menu

Rockjunior:
@rdstern and @lilyclements and please review

Fixes #9647 

This PR Removes ellipses and updates Tidy sub-menu items
Thanks


rdstern:
@Rockjunior here is the new menu:

image

It has gone forward and also backwards.  Maybe I'm using the wrong version, because I had warnings when moving to your branch.

But, if this is your latest, then it is based on the old menu.  So before you added the stack and unstack, and also before you moved the Transform Tricot back to the main menu and renamed it.  And herethe Reorder Factor Levels is given twice.

Rockjunior:
This changes have been done in this PR #9647

PR #9652, opened 2025-05-08
Added the Table Tab option to the 2 Table Options

### LINKED ISSUE #9632: Add Table tab to all three Table Options sub-dialogs

Image

Image

Image

Ag-Derek:
Fixes #9632
This adds the New Tables tab option to the Tables sub-dialog (Before and Same, After was supposed to be implemented by Patrick)
While working this, i realized the options were added as a whole (each option having it own selector, receiver etc) so to have them as group boxes would not be ideal. So i followed the structure and made them separately with radio button assigned to the two options.
I believe this will make it simpler for the user to be able to use them and also for any developer when there are changes to be made would find them easy to do
@rdstern , @lilyclements can you review this
Thanks

### DIFF
--- a/instat/UserTables/Cells/Formats/ucrNewCellFormats.Designer.vb
+++ b/instat/UserTables/Cells/Formats/ucrNewCellFormats.Designer.vb
@@ -1,0 +1,220 @@
+﻿<Global.Microsoft.VisualBasic.CompilerServices.DesignerGenerated()> _
+Partial Class ucrNewCellFormats
+    Inherits System.Windows.Forms.UserControl
+
+    'UserControl overrides dispose to clean up the component list.
+    <System.Diagnostics.DebuggerNonUserCode()> _
+    Protected Overrides Sub Dispose(ByVal disposing As Boolean)
+        Try
+            If disposing AndAlso components IsNot Nothing Then
+                components.Dispose()
+            End If
+        Finally
+            MyBase.Dispose(disposing)
+        End Try
+    End Sub
+
+    'Required by the Windows Form Designer
+    Private components As System.ComponentModel.IContainer
+
+    'NOTE: The following procedure is required by the Windows Form Designer
+    'It can be modified using the Windows Form Designer.  
+    'Do not modify it using the code editor.
+    <System.Diagnostics.DebuggerStepThrough()> _
+    Private Sub InitializeComponent()
+        Me.ucrRowExpression = New instat.ucrRowExpression()
+        Me.Label1 = New System.Windows.Forms.Label()
+        Me.Label2 = New System.Windows.Forms.Label()
+        Me.cboSelectFormat = New System.Windows.Forms.ComboBox()
+        Me.ucrReceiverMultipleCols = New instat.ucrReceiverMultiple()
+        Me.lblFormats = New System.Windows.Forms.Label()
+        Me.btnClearFormats = New System.Windows.Forms.Button()
+        Me.lblColumns = New System.Windows.Forms.Label()
+        Me.btnEnterFormat = New System.Windows.Forms.Button()
+        Me.ucrSelectorCols = New instat.ucrSelectorByDataFrameAddRemove()
+        Me.dataGridFormats = New System.Windows.Forms.DataGridView()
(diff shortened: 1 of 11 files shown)


rdstern:
@derekagorhom that's a good start - maybe?
image

a) It is Table, rather than Tables.
b) Delete the controls you don't need.  In the picture above you don't need 

1) The data selector
2) The multiple receiver
3) The row expressions control
4) The Format's control

Put another way you only need 2 controls above, namely the Select Format and the Enter Format.

And that's also why I suggested, in the issue that Select Format might become a checkbox, default unchecked and when checked you choose one of the radio buttons rather than a drop down.  And each radio button has a format Button to open the options @Patowhiz has constructed.

Similarly for the Missing values, where I think there is just a single control.  Have another read of the issue I wrote.

Ag-Derek:
@rdstern i want to understand what you want the table tab to produce so i have the script code below
So when we remove 
    The data selector
    The multiple receiver
    The row expressions control
    The Format's control
    we should this in our script window?

```

# Dialog: Summarise Two/Three Variables

summaries <- c("summary_mean")

factors <- "village"

combinations <- purrr::cross_df(list(factor=factors, summary=summaries))

last_table <- survey %>% purrr::map2(.x=combinations$factor, .y=combinations$summary, .f=~data_book$summary_table(summaries=.y, data_name="survey", treat_columns_as_factor=FALSE, columns_to_summarise="yield", factors=c(.x,"variety")) %>% tidyr::pivot_wider(names_from=variety, values_from=value) %>% instatExtras::generate_summary_tables() %>% gt::fmt_number( decimals=1, n_sigfig=1))
rm(list=c("last_table", "survey", "summaries", "factors", "combinations"))
```
The second part is the Replacing NA options
```
# Dialog: Summarise Two/Three Variables

summaries <- c("summary_mean")

factors <- "village"

combinations <- purrr::cross_df(list(factor=factors, summary=summaries))

last_table <- survey %>% purrr::map2(.x=combinations$factor, .y=combinations$summary, .f=~data_book$summary_table(summaries=.y, data_name="survey", treat_columns_as_factor=FALSE, columns_to_summarise="yield", factors=c(.x,"variety")) %>% tidyr::pivot_wider(names_from=variety, values_from=value) %>% instatExtras::generate_summary_tables() %>% gt::fmt_number( decimals=2, n_sigfig=1) %>% gt::sub_missing(columns=, missing_text="...."))
rm(list=c("last_table", "survey", "summaries", "factors", "combinations"))
```

rdstern:
@derekagorhom amy R isn't good enough, but those both look sensible to me.  @lilyclements can you confirm?

lilyclements:
@rdstern @Ag-Derek this looks sensible to me too. I assume the idea is to apply these things "globally" to the whole table

Ag-Derek:
@rdstern , @lilyclements I have redone the table Tab
can you have a look
Thanks

rdstern (changes requested):
@derekagorhom you are still making it much more complicated than it is.
a) I am not sure which dialogs to try it with?  It seems to crash on the 2/3 variables summarise.
b) On the Tables > Summarise it seems to work (in not crashing), but the Dark Theme was implemented, when I didn't ask.  Can you keep them independent - by=ut that might have to wait until the main options are working.
c) On the Format, you now have the radio buttons, etc.  Great.  Can you line them up with the buttons.
d) I'm still not convinced you fully follow what you are doing.  Please make sure you understand fully.  When I take the format option it still opens this control, from Patrick.

image

Can you see that this control has Columns and Rows and they can't mean anything here, because you are applying the format to the whole table.  So delete that control and apply just one format to the whole table.

e) Similarly for the missing values, which is even simpler:

image

Here delete the Add button and also the whole Missing texts control.  If I accept NA becomes 0 then that's just changed in the table.  It should be really simple.

Can you try initially with the usual survey data.  Here is an example:

image

It shouldn't be that difficult to make the NA in this table into zero, or blank?

Ag-Derek:
@Patowhiz can we have a meeting on the GT tables, i want to do some modifications to them and i need your advise on the way forward

PR #9655, opened 2025-05-08
Fixed Regression found in the Column Factor dialogs

### LINKED ISSUE #9651: Error on Opening 'Prepare>Column: Factor >Reorder Levels' Dialog
@rdstern @lilyclements @fran2or @jkmusyoka @derekagorhom
An error occurs in the newly released version (0.8.4.5) when navigating to Prepare > Column: Factor > Reorder Levels....

This happens whether a dataset is loaded or you're simply trying to open the dialog with no data present.
The application throws an error message: "Unhandled exception has occurred in your application."

I've attached this screenshot showing the error prompt.

Image

This issue is worth noting.

Ag-Derek:
Fixes #9651 
I did a couple of test runs and they seem to run fine, the Column factor dialog don't seem to cause the problems found anymore. 
@lilyclements please can you do a stress test and see if everything works fine for you too. 
Thanks

### DIFF
--- a/instat/clsRecentFiles.vb
+++ b/instat/clsRecentFiles.vb
@@ -211,27 +211,6 @@ Public Class clsRecentFiles
             'Note: Translate both sides of the comparison in case the 2 sides are in different languages.
             '      This is possible if the language was recently changed.
             If GetTranslation(dfTemp.Text) = GetTranslation(DirectCast(sender, ToolStripMenuItem).Text) Then
-                If dfTemp.Name = "dlgReorderLevels" Then
-                    frmMain.SetDefaultValueInReorderLevels()
-                End If
-                If dfTemp.Name = "dlgRecodeFactor" Then
-                    frmMain.SetDefaultValueInReorderLevels()
-                End If
-                If dfTemp.Name = "dlgDummyVariables" Then
-                    frmMain.SetDefaultValueInReorderLevels()
-                End If
-                If dfTemp.Name = "dlgLabelsLevels" Then
-                    frmMain.SetDefaultValueInReorderLevels()
-                End If
-                If dfTemp.Name = "dlgReferenceLevel" Then
-                    frmMain.SetDefaultValueInReorderLevels()
-                End If
-                If dfTemp.Name = "dlgFactorDataFrame" Then
-                    frmMain.SetDefaultValueInReorderLevels()
-                End If
-                If dfTemp.Name = "dlgCountinFactor" Then
-                    frmMain.SetDefaultValueInReorderLevels()
-                End If
                 dfTemp.ShowDialog()
                 Exit Sub
             End If
--- a/instat/clsRecentFiles.vb
+++ b/instat/clsRecentFiles.vb
@@ -211,6 +211,14 @@ Public Class clsRecentFiles
             'Note: Translate both sides of the comparison in case the 2 sides are in different languages.
             '      This is possible if the language was recently changed.
             If GetTranslation(dfTemp.Text) = GetTranslation(DirectCast(sender, ToolStripMenuItem).Text) Then
+                If frmEditor.IsDataSetLoaded() Then
+                    Select Case dfTemp.Name
+                        Case "dlgReorderLevels", "dlgRecodeFactor", "dlgDummyVariables",
+                            "dlgLabelsLevels", "dlgReferenceLevel", "dlgFactorDataFrame",
+                             "dlgCountinFactor"
+                            frmMain.SetDefaultValueInReorderLevels()
+                    End Select
+                End If
                 dfTemp.ShowDialog()
                 Exit Sub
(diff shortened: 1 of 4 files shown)


instat/clsRecentFiles.vb
```diff
@@ -211,27 +211,6 @@ Public Class clsRecentFiles
...
-                If dfTemp.Name = "dlgReferenceLevel" Then
-                    frmMain.SetDefaultValueInReorderLevels()
-                End If
-                If dfTemp.Name = "dlgFactorDataFrame" Then
-                    frmMain.SetDefaultValueInReorderLevels()
-                End If
-                If dfTemp.Name = "dlgCountinFactor" Then
-                    frmMain.SetDefaultValueInReorderLevels()
-                End If
                 dfTemp.ShowDialog()
```
lilyclements, line 214: What did this code do? I want to check that before reviewing. 
I know that we have the functionality to "remember" the variable you accessed the dlg through if you have gone via the right-click menu. Is this impacting that?

Ag-Derek: This was implemented in such a way that when the user uses the last dialog button it remembers the `setDefaultValueInValueInReorderLevels()`
`setDefaultValueInValueInReorderLevels()` was set such that when the factor dialog was clicked this would automically pick the the first factor in the dataset.

lilyclements: Ok, so out of interest why is this being removed?

Ag-Derek: my understanding is that, that particular section gives the error Rodgers found. because there is not dataset in the reogrid, when you select some of the factor dialogs because it cannot pick the factors in the dataset it gives the error.

rdstern:
@derekagorhom I assume you have resolved the regression by getting rid of the new feature of automatically including a factor variable?
If so, then am afraid I have changed my mind, and hence agree with @lilyclements reticence above.  I prefer we live with this problem for now, until you are able to fix it properly, i.e. not try to load a variable when there is no data.

instat/frmMain.vb
ChrisMarsh82, line 1: Revert changes back and add a check at line 951 to check to see if the dataset exits. If it doesn't the exit out of the procedure

instat/clsRecentFiles.vb
ChrisMarsh82: Revert these changes and just update SetDefaultValueInReorderLevels() to break out if no dataset exists

Ag-Derek:
@ChrisMarsh82 @lilyclements I have added the check
can you review this 
Thanks

rdstern (dismissed):
@Ag-Derek this looks fine to me - and a great relief, and well done.  It is also a great example of a critical issue found and reported by a new member @Rockjunior of the development team.
I really like the example, because who would think of using a factor dialog when there is no data at all.  Answer - well somebody might, and we are trying to produce a package to be used by anyone.  And that sets a much higher standard than just producing a program!

I am approving.  @Emmanuel-Afrifa do you feel able to check the VB code?  If so, and you are happy with it, then please tell @lilyclements so she can merge.  It is fine if not and we'll  ask Vitalis or Patrick or Chris?

instat/frmMain.vb
ChrisMarsh82, line 1: Simplyfy these changes IsDataSetLoaded should be in SetDefaultValueReorderLevels(). Then the check can be done just once. 

Can you also get IsDataSetLoaded to be in frmMain it would be better if the code within frmMain didn't have to reference frmEditor. Can you check the clsDataBook to see if the dataframe is loaded?

ChrisMarsh82 (changes requested):
Try to write less code. If you need to check a dataset is loaded do it in one place not multiple

instat/frmMain.vb
Ag-Derek: Sure i can try that too

Emmanuel-Afrifa:
> I am approving. @Emmanuel-Afrifa do you feel able to check the VB code? If so, and you are happy with it, then please tell @lilyclements so she can merge. It is fine if not and we'll ask Vitalis or Patrick or Chris?

@rdstern I think @ChrisMarsh82 may have already started reviewing the code. Also, as this change touches some other very important and global files like frmMain, I think it will be best for a more experienced developer like @ChrisMarsh82 to review these code changes.

instat/frmMain.vb
Ag-Derek: @ChrisMarsh82 Please is it ok if we leave the IsDataSetLoaded function in the frrmEditor, since moving it would require us to have unvell.ReoGrid.Worksheet in the frmmain which requires `Imports unvell.ReoGrid.Events` into Frmmain and setting this up adds more lines of code which we wanted to avoid

ChrisMarsh82 (approved):
Have not completed functionality test but the code looks fine

PR #9676, opened 2025-05-11
Ensuring the data name runs in the Tricot > Describe > Traits dialog

### LINKED ISSUE #9663: Error message when running the Tricot > Describe > Traits dialog
@lilyclements I have defined the data as Tricot and get this message consistently:

Image

I tried earlier with Breadwheat and think the message is the same.  I seem to remember it from last week?

I also tried the correlations dialog and that seems ok.  I'll now try some of the modelling dialogs.

lilyclements:
Fixes #9663


lilyclements:
`data_name` was not being read into the "data_book$get_object" function, causing an error. 

image


this fixes that issue

Issue #9622, opened 2025-05-06
Use date instead of doy in climatic summary dialog, Station option.


rdstern:
Here is the dialog concerned, Climatic > Prepare > Climatic Summary.

![Image](https://github.com/user-attachments/assets/b644f666-3b2b-4e2b-afac-88592028d39e)

I am in the Station button, so will get one summary for each station.  There are 557 stations.  

For the Day Range (button at the bottom of this dialog) we want to choose the 45 days from the planting date.  The data are over multiple years, so we cannot use the doy for the range.  It will still be useful sometimes for this option, but here we would like to specify the day range based on the date, rather than the doy.

Here is the sub-dialog:

![Image](https://github.com/user-attachments/assets/046fbc1f-d582-44bf-934b-e9dff542b8fb)

Please could there be a checkbox, currently disabled unless there is the Station setting in the main dialog.
It says `Use Date not Day of Year (doy) 

The selector then (or anyway) includes date columns, and they can n ow be chosen.  
Then the Condition gives a date rather than a doy.  All should then work?

Note this will initially not work with choosing a Fixed date in the formula above.  Don't worry at this stage and we'll have another pull request for that.

Note the same sub-dialog is also used in the Spells and Extremes dialogs.  It can (I think) be enabled there and may work easily.  I'm not sure about other dialogs, but it will do no harm.  `

rdstern:
@lilyclements, @jkmusyoka  and @Vitalis95 I have hit a problem with this task here that exceeds my R skills.  My task is to make it really simple, in Tricot, to filter on dates, rather than the day of year at least within the Climatic Summary dialog, when we use the Station option.  

This is particularly because Tricot get a lot of climatic indices in the 45 days from planting.  I can do this (a long way) by filtering the climatic data before using the climatic summary, etc dialogs.

But we already have a neat (and powerful) system of filtering within the climatic dialogs, based on the day of year (doy).  I thought that all that is needed is to replace the doy by the date, and it should all work!?

I then realised that I could actually use the current dialog and simply change the resulting script to use the new variables!  How simple eh?

I also found the existing dialog here, uses doy, simply because that' in the corresponding receiver.  So, with no change in code I replaced it by `date`, and lo and behold the condition uses the date!

![Image](https://github.com/user-attachments/assets/3f6c5867-ad5b-491c-a7b7-5288ea4acddc)

Now I go to the sub-dialog - it doesn't yet show date columns, so I use something else (age) for now:

[...]

PR #9678, opened 2025-05-12
Improved climatic summary sub-dialog, Station option

Vitalis95:
Fixes partially #9622 

@rdstern @derekagorhom @Rockjunior , fixed part of the issue - Use Date instead of Day of Year (doy)


rdstern:
@Vitalis95 many thanks for taking this up.
Here is the sub-dialog:
image

a) The first thing needed is to permit date-type variables in the data selector.  I assume that should be easy?
b) Then there could be a checkbox, towards the right, default unchecked that says something like Use date rather than doy for condition.
c) If so, then date is substuted for doy in the condition and it should work fine?  Both there and then back in the main dialog.

rdstern:
@Vitalis95 I now suggest we should investigate the R-code first to facilitate this change.  It would provide an easy way for users to provide the filter used in Tricot.  This applies to the summary, spells and extremes dialogs.  But I am having trouble getting the R-code to work,  I suggest working with @jkmusyoka and maybe @lilyclements on this first.

lilyclements:
@rdstern what on the R code needs to be discussed here? I thought the R code to run was;

```r
# Then our code here
day_filter <- instatCalculations::instat_calculation$new(type="filter", function_exp="date >= planting_date & date <= planting_date + 44", calculated_from=databook::calc_from_convert(x=list(data_RDS_stacked=c("date", "planting_date"))))
data_book$calculate_summary(columns_to_summarise="rain", data_name="data_RDS_stacked", factors=c("id"), additional_filter=day_filter, summaries=c("summary_count", "summary_sum"), silent=TRUE)
rm(day_filter)
```

lilyclements:
Assuming that `planting_date` is one variable, and `date` is another. Here, `planting_date`, is the date variable given in the sdg under "variable day"

rdstern:
@lilyclements that's what I have been trying, but I kept getting an error with that condition, (probably because my R is so feeble).  It said my version of your condition didn't work on dates.  I now managed it with the dplyr::between function, but yours works fine and is the obvious solution!

I attach the code where I started with the current sub-dialog, using doy instead of date and age instead of planting_date and then edited the line.

```
# Dialog: Climatic Summary

#day_filter <- instatCalculations::instat_calculation$new(type="filter", function_exp="doy >= age & doy <= age + 44", calculated_from=databook::calc_from_convert(x=list(data_RDS_stacked="doy", covar=c("age"))))

#day_filter <- instatCalculations::instat_calculation$new(type="filter", function_exp="dplyr::between(date,planting_date ,planting_date + 44 )", calculated_from=databook::calc_from_convert(x=list(data_RDS_stacked="date", covar=c("planting_date"))))


day_filter <- instatCalculations::instat_calculation$new(type="filter", function_exp="date >= planting_date & date <= planting_date + 44", calculated_from=databook::calc_from_convert(x=list(data_RDS_stacked="date", covar=c("planting_date"))))
data_book$calculate_summary(columns_to_summarise="rain", data_name="data_RDS_stacked", factors=c("id"), additional_filter=day_filter, j=1, summaries=c("summary_count", "summary_max", "summary_sum"), silent=TRUE)
rm(day_filter)
```
So I hope the pull request can now be completed, initially just for the sub-dialog when the Climatic > Prepare > Summary is used. 
a) Allow date variables to be shown in the second data frame.  For simplicity they can always be shown. 
b) Add the checkbox to use the date.  That's only when the main dialog is on the Station button.
c) Generate your code with those options.  
Lovely!

rdstern:
@lilyclements I don't know why my version of your code didn't seem to work earlier for me.  Your code above  works fine and I prefer it to my between attempt above. 
So the steps are as follows:
a) In the sub-dialog data frame - accessed via variable date, then currently only numerical variables are allowed.  Please allow date variables as well.  For simplicity, don't have any conditions on that.  Whenever that data frame is shown it will do no harm to always allow dates there.
b) Add a checkbox sort of on the right in the sub-dialog, and reasonably low down with Use date rather than day of year (doy).  Default is unchecked.  For now have that anyway, i.e. not limited to the Station option.
c) If checked, then it uses a Lily-type condition (see above) instead of the one with doy!

PR #9679, opened 2025-05-12
Improved Describe>Table >Summaries Dialog

### LINKED ISSUE #9529: Minor problems with the Describe > Tables > Summarise dialog
@Vitalis95 I am very happy with how solidly this dialog has held up.  There are minor issues, that I hope can easily be resolved.  It never gave an error message, so is holding up very well.

I used the usual diamonds data from ggplot2.  

Here is the first dialog:

Image

And here are the results:

Image

a) Could the table be given without the column labelled summary variable?  That's a general simplification for the summary table too (and for the 2/3/ variable summarise dialog!)
b) Could the names be given for the two column factors?

I then added a title and an Excel theme to give this table - nice.

Image

c) But when I returned to the table options I found it had not remembered those settings in the sub-dialog.  That's from the Header and the Themes.  I tried putting something in Stub, which it ignored, but did remember that?

d) I then gave a summary table.  It works fine, with just one small hiccup, namely the order of the (summary) factor in the columns.  Here is the dialog:

Image

And here is the result:

Image

With the summaries having the value 4 - so after the 3 "real" factors, it should be below the color factor, not above it.  If I change the position to 3, then I get the layout that I should have with 4!

MeSophie:
Partly Fixes #9529
@rdstern @N-thony  I solved a) and b). Please have a look.
For c) I realized that the Table Options input are  not also remember theirs values in Describe>Table >Presentation Dialog I'm still investigate on it.


berylwaswa:
@MeSophie @rdstern 
I have tested this and it looks good. I used the diamonds data from ggplot2.

Here is the first dialog;

instat_Fbt9sNvJq7

And here is the output;

instat_bNGkCoRMSK

- The table is given without the column labelled summary variable. 

- The names are given for the two-column factors.

lilyclements (approved):
@rdstern can you look at this?

rdstern:
@lilyclements and @MeSophie I am confused, because I thought this was what @fran2or was working on with one of the Innodems developers during the sprint?  I thought the column was transformed into the tatle for this dialog and also for the more general Describe > Tables > Summarise (Frequency) option dialog?  Is that somewhere else.  I though it was merged already?

lilyclements:
@rdstern good point. We did it in Two Variable Describe, and then in this dialog in #9639.

@MeSophie were you still getting it as an issue in this PR before you made the changes?

MeSophie:
@lilyclements I will then close this PR

Issue #9686, opened 2025-05-13
Update how packages are installed and checked


ChrisMarsh82:
Need to improve how packages are installed through the installer and then checked when users first go into R-Instat

Currently install_pacakages.R is where all of our packages are defined. This then can be run locally to create Rsetup.R and InstallPackages.R

Rsetup.R is a list of all required packages and their versions. This is used by R-Instat to determine if all packages are installed.

InstallPackages.R is used by GitHub actions to create the installer. 

The issue is that these files are quite hard to maintain and is currently happening manually. The other issue is that we now have quite a few packages not available on CRAN therefore the InstallPackages.R is getting more complicated. 

Can we simplify these files? We need a list of packages we want to install. Which should be install_packages (updated by developers) This list should determine where the package should be installed from.

Does Rsetup.R need version numbers and dependencies? This would be simpler if it was just the packages in install_packages.R, and when we install, we always install dependencies? 

@lilyclements @dannyparsons Your views on this would be good. I want to simplify how builds are done without losing useful functionality

lilyclements:
I don't think I am the person that can help with this. Perhaps could this be something to discuss with @volloholic

PR #9687, opened 2025-05-13
updated installer

ChrisMarsh82:
Updated installer to generate Rsetup.R when run. 
Removed InstallPackages.R and setup installer to use install_packages.R

@lilyclements Does this look a sensible approach to you? It means when a new package needs to be added to 'install_packages.R' Developers will need to decide at this point how the package should be installed? either via CRAN or other method. 
There is an order to how some of the packages are installed I assume this is due to dependencies?

The installer will then add packs and versions to the top of Rsetup.R with all the packages and versions installed at the time the installer is created.


ChrisMarsh82:
#9686

ChrisMarsh82:
> @ChrisMarsh82 this looks far neater than what we're currently doing, and of course I'd trust your code process over mine. This sounds sensible to me. How does a developer add a new package into `install_packages.R` and how they define if it is from CRAN or elsewhere?

There are headers in the file:
#Install packages from CRAN archive
#Install packages from win.binary
#Install packages from CRAN 
#install development packages not on CRAN

depending how it should be installed depends where you put the code. It does rely on the developer knowing what would be the best option but it would be better than the current process where they all get installed from CRAN and then I do a trail and error process until I get them installed

PR #9694, opened 2025-05-14
Fixes an error when calling the right-click Levels-Labels dialog repeatedly 

### LINKED ISSUE #9670: Prevent the common Developer Error when calling the right-click Levels-Labels dialog repeatedly
Here is an example with the cassava data from the gosset package, but I think it may occur with any dataset, containing multiple factors.

Recently @N-thony improved this dialog to take account of an active filter, and this may have been a by-product?  It seems to occur when I use this dialog on multiple variables, where the number of levels of the factor is different.  It is fine the first time, but occurs later.

Image

Note that it is annoying, but not serious.  After clicking OK, the Levels/Labels dialog opens fine.

@MeSophie I think you have met this problem elsewhere?  Could you maybe fix it, and also hopefully report why it happens and what you have done - so the new developers can avoid it in the future?

Vitalis95:
Fixes #9670 
@rdstern @MeSophie , the ucrFactorLabels attempted to set control values before the grid was properly populated, resulting in a row mismatch.


MeSophie:
@Vitalis95 you are right. Thank you. There's one question I haven't answered yet: why does this only happen with the cassava data `reason_like`? At first I thought it was because of the length of the data but by converting the date of `Ghana two stations bad` data (53023 rows) into factors and using the Labels/Levels dialog I don't get this notification.
@rdstern I've also noticed that when we click on the `Add Factor Level` button it disable the `Ok` button and we only have to reset the dialog. What exactly does this button? If we are still implement it can we temporarily disable it?

rdstern (approved):
@Vitalis95 many thanks.  I am approving.
@MeSophie that's a good spot.  But when you add a new level it initially does not have a label - so Ok is disabled.  If you add a label then it seems alright and Ok is then enabled.
@lilyclements can you merge.  And we need to get someone else on this task too?

lilyclements (approved):
great stuff - approve and merging now

PR #9695, opened 2025-05-14
Updated the R-help labels for the transform, summary, factor and text/string keyboards.

### LINKED ISSUE #7619: Improve the R-Help control on the Prepare > Column: Calculator
The last calculator keyboard (integer) has the improved control - but it is not quite right there.  So:

- [x] a) Improve that control.  It currently adds the main package - gmp - when you press the key - and that is fine.  But it should also be the first key - called gmp - when you press the arrow for the list.  So there will be the 3 packages there.
- [x] b) Implement the same R-Help with down arrow, on each of the keyboards - even if there is only a single package used.  When you use the arrow it gives the name of the package.  (This name should be in lower, or upper case to match the name of the package.
- [x] c) So, for hydroGOF it just gives the one package hydroGOF
- [x] d) For circular it just gives circular
- [x] e) for wakefield it just gives wakefield 
- [x] f) for Transform there is now just the one key.  The main and first package is dplyr, then zoo, then base.
- [x] g) Dates/Times lubridate and also hms
- [x] h) Probability just the stats package
- [x] i) Factor.  Not base as now, but forcats, and also labelled.
- [x] j) Text/Strings just stringr.  (But is there also something clear for help on regex?)
- [x] k) Summary stats, then base, then dplyr, then raster (cv) to go to the function rather than the package as a whole.  Similarly robustbase (mc), e1071 (skewness), e1071 (kurtosis), statip (mode), statip (mode1)
- [x] l) Logical base and dplyr.  (Also correct the tooltip for when, by adding 2 commas, so it makes sense!)
- [x] m) Maths base, car (logit), stats (logistic) - in each case go to the function itself for those.

Emmanuel-Afrifa:
Fixed #7619 
@lilyclements @rdstern Kindly check if changes meet your requirements. Thank you

### DIFF
--- a/instat/ucrCalculator.Designer.vb
+++ b/instat/ucrCalculator.Designer.vb
@@ -150,6 +150,7 @@ Partial Class ucrCalculator
         Me.cmdTransformHelp = New instat.ucrSplitButton()
         Me.ContextMenuStripFactor = New System.Windows.Forms.ContextMenuStrip(Me.components)
         Me.FactorForcatsToolStripMenuItem = New System.Windows.Forms.ToolStripMenuItem()
+        Me.LabelledToolStripMenuItem = New System.Windows.Forms.ToolStripMenuItem()
         Me.cmdShuffle = New System.Windows.Forms.Button()
         Me.cmdShift = New System.Windows.Forms.Button()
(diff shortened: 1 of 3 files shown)


lilyclements (approved):
Great! Approve and merging now

PR #9706, opened 2025-05-16
Removing Filter(if any) before adding the Row names/numbers

### LINKED ISSUE #5608: Minor edits on Prepare > Data Frame > Row Names dialogue
This should not take much time, partly because we hardly make any use of row names.  

1) the first option says Copy Row names into First Column.  Delete the First.
2) Move the options box over, and make the dialogue smaller
3) Now the more serious problem.  It doesn't work when a filter is in operation.  There is no reason why it should, but it shouldn't give an error.  I think we will find this problem with other commands too.  There are 3 obvious options and I don't mind what we do - but we should do something.
a) Trap this situation and just give a message ourselves to remove the filter to do this operation.  Or:
b) Simply get rid of the filter and then proceed with the dialogue.  It is very easy to re-apply a filter.  We could give a message to say what we have done.  I would be happy without, particularly if we detect a filter, then we just issue:
```
 # Remove Current Filter
data_book$remove_current_filter(data_name="dodoma") 
```
c) Get rid of the filter, do the command and then re-apply the filter

This point is worth discussing with @danny and @volloholic as it is more general than just this dialogue.

Vitalis95:
Fixes  #5608 
@rdstern @lilyclements @Ag-Derek , fixes part c) of the issue. It gets rid of the filter ( if any) and then proceeds with the dialogue


lilyclements (approved):
This looks good to me. @rdstern over to you

rdstern (approved):
@Vitalis95 thanks, looks fine.
@lilyclements I have also approved

PR #9522, opened 2025-04-08
Gt table improv

Vitalis95:
Still in progress

@N-thony , I have implemented what we discussed, please check

PR #9708, opened 2025-05-16
Added Table Tab in Table Options Sub Dialog

### LINKED ISSUE #9498: Adding to the Tables sub-dialog?
Here is an example of a table from the new `Describe > Tables > Presentation` dialog.

Image

This builds on what @Patowhiz produced, and his presentation of a data frame is now the third button at the top.  It is also a special case of the second button, when no pivotting is needed.

Many presentation tables will be the first button, as in the example above, and this is what pivot tables make easy.  @lilyclements and @MeSophie a feature of these tables is that they are for a single variable.  Hence most of the presentation options will be the same for all the numbers in the table, and so can presumably, like the theme and the title, be specified while making the gt object.  

In the example above I would like to present NA differently (including by blank) and also be able to specify the format, e.g. 2 decimals, or whole numbers, or scientific notation, etc.  I assume this could maybe be a table tab on the sub-dialog?  

If I am right, could you consider the R code, so this can be done?  Then we can have the tab, plus one or two more common items (like the recoding of NA) on the main dialog?

MeSophie:
Partly fixes #9498
@rdstern @lilyclements I have added Table options on Describe>Table>Presentation Dialog Please have a look. 
@lilyclements  Please I need your help with Summaries Table Dialog. The function I used for Presentation Dialog are not working On Summaries Table dialog. I use `fmt_number` for display decimal place and `sub_missing` for Display missing values.


rdstern:
@lilyclements and @MeSophie thanks for this.  But Derek is also working on this topic, in pull request #9652.
I am wondering if you could, instead, look at #9722, which maybe is what Vitalis started in #9522.  I am not sure, so please wait for a reaction from Lily.  (She has a new label for issues of Beginner Item.  This is NOT a beginner item!)

lilyclements:
@rdstern @MeSophie great. I'll have a look at #9722 and advise there later today.

MeSophie:
I will close this since it is already in PR #9652

PR #9724, opened 2025-05-20
Improvements to Extremes and Spells dialogs

### LINKED ISSUE #9703: Improve the Climatic > Prepare > Spells and Extremes dialogs for Tricot
@jkmusyoka and @lilyclements Here is the data set we could use for testing:

[tricot test of date filter2.zip](https://github.com/user-attachments/files/20242366/tricot.test.of.date.filter2.zip)

I was getting this to attack the climdex problems.  I then realised we have a small problem in the Extremes and Spells too.  Once we fix the new problems in climdex we will still have this one.  So I suggest we sort these 2 out first - which should be simple - then return to attach climdex seriously.

Here is the current extremes dialog:

Image

My suggestion is to add a Station button at the top.  It is a 4th button, so in the same position as the station button in the summary dialog.    Then it will produce the same dialog as shown here, but without the year receiver.  

Currently (with the option shown) it produces extremes for each of the 3 years in the data file.  I can get it to produce the correct answers and also link correctly if I trick the dialog and put the id instead of the year into the year receiver!  (The calculation system is working well!)

Here is the Spells dialog, together with another small problem:

Image

The initial task, as for the extremes, is to add a Station Button at the top.  
If that is used the dialog is the same as shown, but omitting the Year receiver.

I will also sort out a bit more and then add the changed R code.  It is a very small change!

The second problem is indicated by the NA at the top of the spells column.  That's because it doesn't know what length the dry spell might have been before the data started.  I am concerned this might be a general problem, but let's solve it for (at least) the new Station option.  Please can the R code change so that effectively dat=y zero for each station had rain.  So it starts counting with 1 (or zero) for the first day at each station.

R-code still to come.  I am now wondering whether @Vitalis95 could do this.  I suggest it could be very easy for him to do this well, and difficult for beginners?

I describe why and give the full task in the next comment.

Vitalis95:
Partially fixes #9703 
@rdstern @lilyclements , added station option to the dialogs


lilyclements (approved):
@Vitalis95 great stuff!

PR #9725, opened 2025-05-20
Added Identical button into the calculator keyboard

### LINKED ISSUE #5906: Adding identical button into the calculator keyboard
I give a case example where I had to use `identical` function. When infilling missing values using the new Climatic>Check Data>Fill Missing values dialog I had to quickly check if two columns are the same to see if infilling by factors, using different methods etc are really different. @rdstern might have other examples where this is useful.

I, therefore, suggest adding this as a button into the **Logical and symbols** keyboard. The button is named **Identical** and the function is `identical` from base R. It returns a logical value so Ideally, it will use the try button but nobody prevents the user from pressing OK. I suggest a tooltip **Tests if two objects are identical: Returns a single logical value, TRUE or FALSE, never NA and never anything other than a single value.**

I also note that expression label at the top of the calculator should have colons! (Expression:)

@rdstern Is there any reason why we should allow the user to write into the try textbox? I think this should be read-only.

anastasia-mbithe:
Fixes #5906 

@rdstern and @lilyclements I have added the Identical button. It returns the number of rows not identical. Also added a tooltip that inform the users that. @rdstern Please check whether the output is what was intended.


lilyclements (approved):
@anastasia-mbithe in terms of your code this is great. 
@rdstern it would be useful to know if this is the intended output from the issue. (If it's not, that's a me-problem which I can fix). Thanks!

rdstern (approved):
@anastasia-mbithe and @lilyclements it is working very well.
I tried with data of different types and was pleased it works for numeric columns, factors, character variables and dates.  When the 2 columns are of different types it gives NA which is fine.  So I am approving - but still have a suggestion for an improvement - of course!

@lilyclements the ave command is designed for one factor, but you can include more.  Which is great.  
Now correlations, etc get more complicated with more than 2 columns, but could we allow this function to permit identical in any number of columns?  Could the function be extended Lily?  For example in breadwheat there are 4 traits.  I wonder how often they are different?

lilyclements:
@rdstern absolutely, let me extend this now. In the meantime, I'll merge these changes as the R code changes are independent to this now.

lilyclements:
@rdstern this is now done in the updated version of `instatExtras`

PR #9734, opened 2025-05-21
Use date instead of doy in climatic summary dialog

### LINKED ISSUE #9622: Use date instead of doy in climatic summary dialog, Station option.

Vitalis95:
Fixes #9622 
@rdstern @lilyclements , it uses date instead of doy in climatic summary dialog when Use date checkbox is checked

### DIFF
--- a/instat/dlgClimaticSummary.vb
+++ b/instat/dlgClimaticSummary.vb
@@ -22,6 +22,7 @@ Public Class dlgClimaticSummary
     Private iReceiverMaxY As Integer
     Private iReceiverLabelMaxY As Integer
     Private bResetSubdialog As Boolean = False
+
     Private clsDefaultFunction, clsIfElseFirstDoyFilledFunction, clsConcFunction, clsSummariesList,
         clsDefaultFactors, clsDayFilterCalc, clsDayFilterCalcFromConvert,
         clsDayFilterCalcFromList, clsAddDateFunction,
@@ -150,7 +151,7 @@ Public Class dlgClimaticSummary
         ucrPnlAnnualWithin.AddToLinkedControls({ucrChkAddDateColumn}, {rdoAnnual, rdoAnnualWithinYear}, bNewLinkedAddRemoveParameter:=True, bNewLinkedHideIfParameterMissing:=True)
         ucrReceiverYear.SetLinkedDisplayControl(lblYear)
         ucrReceiverWithinYear.SetLinkedDisplayControl(lblWithinYear)
-
+        AddDateDoy()
         ucrInputFilterPreview.IsReadOnly = True
     End Sub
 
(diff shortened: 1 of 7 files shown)


lilyclements:
@Vitalis95  is this ready for review?

Vitalis95:
@lilyclements , please have look at it

lilyclements:
@Vitalis95 this is great! Nice one and really well done! The only one thing is very small, I hope, and that is that is it possible for the selector on the sub-dialog to be "frozen", such that it is always the same data frame in the selector as that on the main dialog?
If not possible then it's not a big deal!

rdstern (approved):
@Vitalis95 and @lilyclements this looks fine.  I also (if I understand Lily's request correctly), don't want the sub-dialog data selector frozen, because it could be a linked data frame. 

I attach the data, because the pull request was giving me a real problem:

[tricot test of date filter3.zip](https://github.com/user-attachments/files/20382460/tricot.test.of.date.filter3.zip)

Now I have realised - it is probably working too well - must check.

rdstern (commented):
@Vitalis95 and @lilyclements this looks fine.  I also (if I understand Lily's request correctly), don't want the sub-dialog data selector frozen, because it could be a linked data frame. 

I attach the data, because the pull request was giving me a real problem:

[tricot test of date filter3.zip](https://github.com/user-attachments/files/20382460/tricot.test.of.date.filter3.zip)

Now I have realised - it is probably working too well - must check.

lilyclements:
> I also (if I understand Lily's request correctly), don't want the sub-dialog data selector frozen, because it could be a linked data frame.
@rdstern good point! @Vitalis95 ignore my suggestion

a) I assume to (a) that this is related to changes in the define climatic dialog. Very nice though!
b) Investigating this now

lilyclements:
@rdstern this is working for me - and very nicely too with the link in there! Is the issue just that the "Omit Missing Values" isn't checked by default?

rdstern:
@lilyclements let's merge then.  I need to make more progress on the climatic side.  This is a good step.

PR #9742, opened 2025-05-23
Extended the One Covariate and Modelling General dialogs

### LINKED ISSUE #9627: Maybe extend the PlackettLuce with one covariate to run with just plot-level data?
@lilyclements I don't want you to extend the functions they have used, but currently the Tricot PlacketLuce (with one covariate) will only run with Variety data that is just at the variety level.

In my cochran-bib example I contrived a variety covariate, but put it at the plot level, i.e. with the y variable.  In general covariates could indeed be at the plot level (e.g. presence or absence of dead plants in each plot.  So I'm surprised it is not allowed by the function.  If it is allowed, then so should we.

I was quite easily able to get the data in the right form.  In the Unstack Dialog there is a Restore Hierarchy option to do just that.  Then I had to use the Define again, this time with the variety data as well as the plot data.  That worked fine, and now the one covariate dialog also runs.

That makes me also suggest we could maybe have a second example of the cochran data and maybe have ID-level data as well as genotype data - in addition to the plot-level data.  What about making the 13 blocks as Rainy, Middling, or Dry?  That would link with the climatic variables presumably coming into the model at the ID level.  If so, then I'd still be inclined to also add the new climate variable at the plot level, because that's what is usually done.

Emmanuel-Afrifa:
Fixes #9627
@lilyclements @rdstern 
I have added a new check and modified the code generated depending on which check passed for both the modelling general and one covariate dialogs. 

@lilyclements the changes you proposed worked and it doesn't result in an error anymore. Kindly review the changes and see if everything is as you wanted it.

### DIFF
--- a/instat/dlgTricotModelOneVarCov.vb
+++ b/instat/dlgTricotModelOneVarCov.vb
@@ -21,22 +21,25 @@ Public Class dlgTricotModelOneVarCov
     Private bReset As Boolean = True
     Public bResetSubDialog As Boolean = False
     Private bIsUnique As Boolean = True
+    Private bCheck7Passed As Boolean = False
     Private clsGetVariablesMetadataFunction As New RFunction
     Private clsWrapBarFunction, clsWrapPlotFunction,
         clsHeatFunction, clsPlotFunction, clsBarfunction, clsAnnovaFunction, clsSummaryFunction, clsAICFunction,
         clsCoefFunction, clsConfidenLimFunction, clsDevianceFunction, clsEstimatesFunction, clsItemsFunction,
         clsPariPropFunction, clsQuasivarianceFunction, clsReliabilityFunction, clsSecondEstimatesFunction,
         clsStatsFunction, clsRegretFunction, clsTopItemFunction, clsNodeRuleFunction, clsNodeLabFuction,
-        clsTreeFunction, clsVarianCovaMatrixFunction, clsImportDataFunction, clsWrapTrees As New RFunction
+        clsTreeFunction, clsVarianCovaMatrixFunction, clsImportDataFunction, clsWrapTrees,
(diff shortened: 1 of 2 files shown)


lilyclements (approved):
@Emmanuel-Afrifa great stuff nice one! Approving from my end.

@rdstern over to you -- this is for the "One Covariate" and "General" dialogs to work with plot level data as well as variety level data.

rdstern:
@lilyclements maybe I am asking too much?
It can now handle a genotype covariate that is at the plot level.  But it complains with a location covariate?

image

With this example I can "trick" it, of course, because there is symmetry in these data between the location and the genotype:  Then it works - but it is a locatrion model and not a genotype model:

=
image

I suggest this is useful so far, so am approving.  But it shows I have misunderstood the baseline - which is the no covariate model.

I had assumed it was a rank equivalent of the analysis of variance - so a model with both location and genotype.  So, there is an order to the fitting, but both are in the model?  But it clearly isn't working that way.   I'm not so clear now how to incrporate the climatic covariates in the model - partly because I don't fully undertand what the trees are doing.

rdstern (approved):
So I am approving!
[cochran_bib.zip](https://github.com/user-attachments/files/20481482/cochran_bib.zip)

I attach the data I am using.  With both a genotype and a climatic covariate.

lilyclements:
@rdstern ok, it is interesting to see how you are wanting to use it. This is the `pladmm` model. In [this vignette](https://cran.r-project.org/web/packages/PlackettLuce/vignettes/PLADMM.html), `pladmm` is known also as “Plackett Luce with Item (Variety) Covariates”, so it is built for the item level.
So it will work in the plot-level data, provided that those variables are unique across the varieties.
Perhaps me, you, and @volloholic should have a meeting on this to discuss. 
I assume there are other models at different levels, but it is unclear which models they use in tricots.

PR #9743, opened 2025-05-23
Changed the functions that run when the first three keys in the extRemes keyboard is used

### LINKED ISSUE #5539: Adding new keys to the extRemes keyboard in Use Model dialogue

Emmanuel-Afrifa:
Fixes #5539 
@rdstern @lilyclements 
I have made the changes you requested. 

The `print.fevd`, `summary.fevd` and `plot.fevd` keys now run the functions `print`, `summary` and `plot` respectively.

I tried to change their names on the designer as well to just  `print`, `summary`, and `plot` but for some reason, it didn't reflect when I run the solution. I even tried setting the `.Text` properties explicitly in the code but that didn't work either.

I don't know if that's what you even wanted. I just assumed that since the functions have changed, the names on the buttons might change as well, although the issue seemed more concerned about the code they produce rather than what they were called. If that's actually something you want, I can look into it more carefully and resolve it later.

### DIFF
--- a/instat/dlgUseModel.vb
+++ b/instat/dlgUseModel.vb
@@ -147,27 +147,27 @@ Public Class dlgUseModel
     Private Sub cmdSummaryFevd_click(sender As Object, e As EventArgs) Handles cmdSummaryFevd.Click
         Clear()
         If ucrChkIncludeArguments.Checked Then
-            ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("extRemes::summary(object=, ...)", 6)
+            ucrReceiverForTestColumn.AddToReceiverAtCursorPosition("summary(object=, ...)", 6)
         Else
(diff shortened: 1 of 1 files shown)


lilyclements (approved):
@Emmanuel-Afrifa looks good! I removed the package name as that's not needed. Approving it. 
@rdstern over to you

PR #9746, opened 2025-05-26
Tools menu Improvements

### LINKED ISSUE #8988: Improve the Tools Menu
@N-thony and @ChrisMarsh82 and @Patowhiz here is the current tools menu, together with some questions.  I think it was done by Danny right at the start of R-Instat and has hardly been changed.

image

a) `Run R Code: `This is disabled.  I am not sure what it is supposed to do and suggest we simply delete it?  (It could open the Log/Script window, but that is in the View menu.)
b)  `Restart R: ` This is disabled:  Is it possible to do this?  If so, then that would be wonderful, because we could now rescue the current session from the backup?  A problem is that while a command is running, I don't seem to be able to get to the menus.  [Here is a method perhaps in general?](https://github.com/randy3k/radian/issues/296#issuecomment-974583380)
c) `Check for Updates:`  This is enabled, but doesn't do anything!  It would be nice if it were available.  I suggest we keep for now and disable.
d) `Clear Output Window: ` This seems to work, and gives a warning too.  I suggest we keep it as it is.
e) `Save Current Options: ` This is disabled.  Currently the options do save.  I suggest we delete it and also the lines.  That will make the Tools menu simpler!
f) `Restore Backup:`  This is the new dialog that used to be just when we crashed.  I love it and need to add the help.
g) `Install R package:`  Currently this assumes Admin rights and installs in the single same directory that is used for the other packages.  This is totally separate from R packages with RStudio - if that is also installed.  I like the separation and would like to keep it.  But I would very much like to make it easier to add (or update) the R packages in R-Instat.  So, could we have a new writeable directory for these packages.  However, I would like to clarify what happens when a user updates R-Instat.  I am ok if there is then a new directory, which is empty, so they need to add their own packages again.  But, in that case, we would need to be able to know what was in the user directory.  Perhaps we could also make an editable script available to help with the restore of the packages needed?
h) `Options:   `  Not much to say at this time. There's lots here for someone else to make suggestions!  For a start I notice:
image

Do we disable this option?

In the Data view: 
image

Could we display the backup period better?

What does the option above mean?  It isn't checked, but I thought we did that anyway.

More generally what does "Factory Reset" do.  I have been scared to press it - so I haven't - in 8 years - but now I need to write the help about it!!!

Ag-Derek:
Fixes #8988 
From checking R-instat during the weekend i found out that 
items 1 to 5 were already done so i did items 6-7
@lilyclements can review this 
Thanks

### DIFF
--- a/instat/dlgOptions.Designer.vb
+++ b/instat/dlgOptions.Designer.vb
@@ -38,12 +38,12 @@ Partial Class dlgOptions
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()>
     Private Sub InitializeComponent()
-        Dim TreeNode13 As System.Windows.Forms.TreeNode = New System.Windows.Forms.TreeNode("Languages")
-        Dim TreeNode14 As System.Windows.Forms.TreeNode = New System.Windows.Forms.TreeNode("Comments")
-        Dim TreeNode15 As System.Windows.Forms.TreeNode = New System.Windows.Forms.TreeNode("Import")
-        Dim TreeNode16 As System.Windows.Forms.TreeNode = New System.Windows.Forms.TreeNode("Output Window")
-        Dim TreeNode17 As System.Windows.Forms.TreeNode = New System.Windows.Forms.TreeNode("Commands")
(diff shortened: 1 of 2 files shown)


lilyclements (commented):
@Ag-Derek perfect! Approving. @rdstern can you check that this is how you want the options dialog to look?

PR #9758, opened 2025-05-30
Adding new ucr for the use table dialog

lilyclements:


lilyclements:
@Vitalis95 this is as far as I could get on adding in our columns from the `gt` object.

I've created new ucrs:

**1. ucrTable** 

This contains the ucrDataFrame to get the data frame name, and a new label saying "Table:", and finally a new combo box which gives the table names of the corresponding data frame.
 This successfully fills with your table object names but only if you change data frame to a new one, and back again. 

**2. ucrSelectorByTable**
This aims to work just like `ucrSelectorByDataFrame`. 
I cannot get the list to show the column names. The R code is correct. I suspect the issue is in that the `ucrSelectorTable` inherits from `ucrSelector`, which is the same that `ucrSelectorDataFrame` inherits from. 

**3. ucrSelectorByTableAddRemove**
This aims to work just like `ucrSelectorByDataFrameAddRemove`, and I think this is fine code-wise. 

I have put in prints. And I can see that we are getting the correct column names read in. I can see the r script in `clsRLink` is running 

```r
colnames(data_book$get_object_data(data_name = "survey", object_name = "frequency_table1", as_file = FALSE)[['_data']])
```

and it is given the correct output (`village`, `value`). 

But I can also see from the output window that we get:

```r
Error in self$get_data_objects(data_name) : village not found
```

I suspect this is coming from `ucrSelectorByTable` inheriting from `ucrSelector`. But I don't know. Do you have any ideas? If not, then I will put this to @Patowhiz as I have spent far too long on this now.

lilyclements:
@rdstern I am not good at VB code, and with the Robots help I've started to get there but only partially there. I think @Patowhiz is going to have to take this over since I have spent far too long on this. 

The approach I'm seeing is:

**1. Create a new set of ucrs. This is to ultimately have one like this:**

image

I have outlined the ucrs in my message above to @Vitalis95 

**2. I have updated `GetColumnNamesFromTable` (in `frmMain.clsRLink`)**

Used the _data slot from data_book$get_object_data() to retrieve column names of a table object.
Confirmed via debugging that the R script returned valid names (e.g., "village" and "value").
This can be seen in changes in the clsRLink` file

**3. Modified `LoadList()` in ucrSelectorByTablesAddRemove**

It successfully fetched and displayed column names from the table's _data slot -- verified by debug print logs like:

```r
Column names found: village, value
```

**4. The Problem (I think)**
The issue came from the `Add()` method, from `ucrSelector`. I think that we need to create our own `ucrSelector` that `ucrSelectorByTables` inherits. 

**What am I on about?**
I have no idea.
What I do know is that the R code is there.
The changes in the `clsRLink` hopefully at least help show the R code we want to be doing here. 
I am leaving this now to @Vitalis95 to see if he can offer any guidance, and, failing that, for @Patowhiz to step in and take this over. He might have a much better way to do it -- which would be far more efficient than what I'm doing too. My understanding was that he did not know the R code side. I have here hopefully illustrated the R code side works.

rdstern:
@Patowhiz I note another week has gone by with no further work on the Use Table dialog.  I am expecting to use this on the Tricot workshop from 30 June.  So, I hope you haven't forgotten us?  I also hope you can take it on from here, because @lilyclements tried valiently, but is not a VB person.  I hope she did enough at the R end, because she is now on leave for 2 weeks!

lilyclements:
@Patowhiz how do we get the data from the table object if we do not use `gt$_data`?

PR #9779, opened 2025-06-03
Implementing Omit Missing Value Options In Describe>Two Variable> Summarise

### LINKED ISSUE #6012: Separate missing value options on summary dialogs

Ag-Derek:
Fixes #6012 
The checkbox and the sdg were already implement but it looks like they were hidden, also they were only implemented for the two variable option on the Summarise dialog when you have a categorical and numberic which gives a `summary table`. 
@lilyclements when the user has a summary table summary they should be able to see it now.
This is ready for review
Thanks

### DIFF
--- a/instat/dlgDescribeTwoVariable.Designer.vb
+++ b/instat/dlgDescribeTwoVariable.Designer.vb
@@ -130,7 +130,7 @@ Partial Class dlgDescribeTwoVariable
         'cmdMissingOptions
         '
         Me.cmdMissingOptions.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.cmdMissingOptions.Location = New System.Drawing.Point(359, 293)
+        Me.cmdMissingOptions.Location = New System.Drawing.Point(146, 365)
         Me.cmdMissingOptions.Name = "cmdMissingOptions"
         Me.cmdMissingOptions.Size = New System.Drawing.Size(105, 23)
         Me.cmdMissingOptions.TabIndex = 21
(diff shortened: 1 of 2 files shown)


lilyclements (commented):
@Ag-Derek Nice! Can you also put in for the Three variable case? Have you been able to test thist oo?

lilyclements (commented):
@Ag-Derek Is the missing options checkbox under both Two Variable and Three Variable tabs? I can't see it under the Two Var tab but perhaps something happened in merging. But then I can see it under the three var tab

berylwaswa (dismissed):
@Ag-Derek @lilyclements 
I have checked both the two variables and three variables tabs and the missing values checkbox is visible in both. See below.
Two variables
instat_QljRwHlmDx

Three variables
instat_Xp4kf4Quoe

Thanks.

lilyclements (commented):
@Ag-Derek this is great! A few things to report

1. There is something occurring on the R side if more than one check box is selected on the missing values. I will set up a separate issue on this.
2. If I click on "Omit Missing Values", then I can see my button for more options. However, after pressing OK and reopening, then my missing values options box is no longer visible. Can you look into this? I think this is only on the Two Variables case.

Very nice once again!

lilyclements (approved):
@Ag-Derek great! Well done
@berylwaswa if you're still happy with this can you re-approve it?

PR #9799, opened 2025-06-06
  Improving dlgSplit dialog to use multiple variables efficiently

### LINKED ISSUE #3751: dlgSplitText as an example of using dialogs efficiently on multiple variables
This is with the split dialog, but probably applies with other dialogs too.

The split dialog has a single receiver because R will only allow the user to edit one column at a time.
However, often when I'm preparing data, I want to use the same command on multiple columns.

For example, I was dealing with
Period_50
Period_60
Period_70
As three different columns and I want to split by the "_".

Instead of reopening the dialog to keep having to change my column, would it be possible to say run the function x times (for the x columns in the receiver).

Not top of priorities, but just an idea.

Vitalis95:
Fixes #3751 
@lilyclements @rdstern @Ag-Derek , this ready for review

### DIFF
--- a/instat/dlgSplitText.Designer.vb
+++ b/instat/dlgSplitText.Designer.vb
@@ -59,13 +59,14 @@ Partial Class dlgSplitText
         Me.rdoMultiple = New System.Windows.Forms.RadioButton()
         Me.rdoSingle = New System.Windows.Forms.RadioButton()
         Me.ucrPnlColumnSelectOptions = New instat.UcrPanel()
+        Me.grpVar = New System.Windows.Forms.GroupBox()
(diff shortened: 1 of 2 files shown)


Vitalis95:
@lilyclements , have a look at it

lilyclements (dismissed):
@Vitalis95 amazing! Nice one. @rdstern over to you to review
Or @Ag-Derek or @berylwaswa

Vitalis95:
@rdstern , I have added the group box

rdstern (approved):
@Vitalis95 that's looking great, many thanks.
@Patowhiz please could you do a final code check and merge?

PR #9806, opened 2025-06-09
Added new format to the Dates> Make Date dialog

### LINKED ISSUE #9774: Add a new option to the Dates > Make Date dialog
@derekagorhom I think this should be very easy, by anyone?

Here is the current dialog:

Image

Notice, in the data that the date-like column (called date) is a character column.  This dialog makes it into a column of type date.

The format of the column is:
`Year(4 digit).Month.Day `

The first option in the drop down is very similar, namely `Year(4 digit)-Month-Day `.  

Please could you add the option above at the top of the list, i.e. `Year(4 digit).Month.Day ` to the top of the drop-down list.

Here are the data as shoen above to test:

[chirps_rain_date.zip](https://github.com/user-attachments/files/20563261/chirps_rain_date.zip)

anastasia-mbithe:
Fixes #9774

@rdstern and @berylwaswa I have added the new format as specified.
@rdstern I have 2 questions.
1. what is the expected output since it returns an output in this format: YYYY-MM-D
2. Do you want it to be the default such that when someone clicks te dropdown, its the one highlighted or we retain YYYY-MM-D?


berylwaswa (commented):
@rdstern The format has been added to the list.

instat_9zpuSymW5u

 
However, @anastasia-mbithe when I apply it, it doesn't appear on the date column produced. I assume it should?
 
instat_qu1tKynoRN

anastasia-mbithe:
@berylwaswa Yes I noticed that.  That's why I had asked @rdstern about the expected output.

I also tried researching a bit about it and got this; 

> R converts the character "2014.06.23" into a Date object, and the internal display format for Date objects is always "YYYY-MM-DD".

rdstern (approved):
@anastasia-mbithe I like it.  Many thanks.  It seems to work fine.  And we will be using it in the Tricot workshop on day 3.

image

I am not sure about the comment from @berylwaswa ?  Beryl if I understand correctly, then to overwrite the existing variable you use the down arrow on the Store Date control.  Then you can choose to save the result back into the existing date variable.  Was that your query?

I am approving.  @berylwaswa  if you are now Ok, could you approve too.

@Patowhiz this should be a simple final check for you to then merge.

berylwaswa (approved):
@rdstern I am approving this.

PR #9808, opened 2025-06-10
Implemented Data Format option for the table Option in the GT Tables

### LINKED ISSUE #9632: Add Table tab to all three Table Options sub-dialogs

Ag-Derek:
Fixes #9632 
This adds the Data format option on the Table option for the two GT tables (Before and Current) 
@rdstern, @lilyclements  this is ready for review. 
Thanks

### DIFF
--- a/instat/UserTables/sdgBeforeTablesOption.vb
+++ b/instat/UserTables/sdgBeforeTablesOption.vb
@@ -44,8 +44,8 @@ Public Class sdgBeforeTablesOption
         ucrChkSelectTheme.SetText("Select Theme")
         ucrChkManualTheme.SetText("Manual Theme")
 
-        ucrChkDataFormat.SetText("Specify the Data Format")
-        ucrChkMissingValues.SetText("Replace NA")
+        ucrChkDataFormat.SetText("Format the Data:")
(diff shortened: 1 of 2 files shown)


Ag-Derek:
@rdstern the reason why nothing is remembered is because we are calling a sub-dialog from a sub-dialog, the usual way to make dialogs remember their default state cannot be done on this sub-dialog,

lilyclements:
@Ag-Derek is this something that can be fixed in the system? Perhaps @Patowhiz has a solution we can add so that we can remember default states?

lilyclements:
@Ag-Derek I can see the sub-sub-dialog works for the Themes options:

Under the boxplot dialog, click on the sub for plot options, then go to the themes tab, and select "Simple Options". If you click around on there, close the sub, and reopen it, it has remembered it.
Is there a reason why it works there and wouldn't work here?

Patowhiz:
@Ag-Derek thanks for tagging me. I'm now working on the sub-dialog. I'll take over this task. Thanks.

Ag-Derek:
closing this since Patrick is taking over

PR #9811, opened 2025-06-11
Re-opening of R Package datasets from the Toolbar

### LINKED ISSUE #4108: Updated toolbar

anastasia-mbithe:
Fixes #4108 

@Patowhiz here is a PR for the issue. 

I've tried saving the package name and dataset name as a file to be accessed from the recent files and also to facilitate its display and recalling but seems something is still wrong.

Unfortuntely, there are some frmMain changes, so it will have conflicts. Once we figure out how to implement, I can open a different pr.

### DIFF
--- a/instat/clsRecentFiles.vb
+++ b/instat/clsRecentFiles.vb
@@ -16,6 +16,7 @@
 
 Imports System.IO
 Imports instat.Translations
+Imports RDotNet
 
 Public Class clsRecentFiles
     Public lstRecentDialogs As New List(Of Form)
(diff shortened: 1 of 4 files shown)


instat/clsRecentFiles.vb
```diff
@@ -120,6 +125,19 @@ Public Class clsRecentFiles
         UpdateRecentFilesMenuItems()
     End Sub
 
+    Public Sub addToMenu(dfdataframe As DataFrame)
```
Patowhiz, line 128: @anastasia-mbithe I think this should be `"{packageName}::{datasetName}"` not the data frame object.

PR #9813, opened 2025-06-11
R viewer for metadata

### LINKED ISSUE #9357: Tidy the metadata bottom right-click
There are three items here, either to be taken together, or as separate pull requests.

1. @Fidel365 here is the bottom metadata right-click menu currently:

image

You have just tidied the data frame right click from the data window.  I suggest it should be the same from here.

So get rid of the View Data option, and reorder the others so they are identical to the one you tidied earlier.

2. At the same time the option for the R-viewer, in the toolbar, ignores the meta-data window.   Could it either:
a) Show the data for the window - data or metadata - that has the current focus,  Or
b) If that is difficult, then (as now) show the window for the data view, unless that is closed.  If closed, then it shows the data corresponding to the current metadata?
 
3. When implementing the R-viewer from the toolbar it still says:

`# Right click menu: View R Data Frame` in the output window.

Of course it is no longer a Right click menu.  So now it could either say `#Toolbar Option: View R Data Frame`, or just `#View R Data Frame`

Thanks

Emmanuel-Afrifa:
Fixes #9357
@lilyclements  @Ag-Derek @rdstern 

Please, I have added the functionality for the R-Viewer pop-up window to display the column metadata information when the  `Swap Data with Column Metadata` is checked from the dropdown on the `Toggle data view open and closed` toolbar before choosing the `R-Data Viewer` option.

Although it works, I realized that the last column in the main data view `Signif_Figure` isn't showing in the R-Viewer pop-up. In the code, I used `data_book$get_variables_metadata`. Perhaps @lilyclements might have a different option to this.

Also, I realized that when I load the `experimental_survey` data for the first time and I check `Swap Data with Column Metadata`, the main Data View becomes blank. I have to uncheck and check it again before it displays the column metadata. I can't figure out why this is happening. Perhaps, it will be best to check to see if the problem existed before I made the changes or not, so that we can narrow the source.

 

Lastly, although it seems to work, perhaps one of the senior developers @ChrisMarsh82 @Patowhiz might review it to make sure it doesn't break this in other places, since I made changes to `frmMain` and `clsPrepareFunctionsForGrids`, which affect almost all dialogs.


Patowhiz:
@Emmanuel-Afrifa The code looks good, user reviewers should now proceed with testing.

Regarding your comment about the data viewer appearing blank, I'm a bit unclear on what you mean. Please note that the **column metadata viewer** and the **data viewer** are separate components. Your changes shouldn't affect the data viewer at all.

You also mentioned you're unsure whether this issue existed before your changes. I recommend testing the behavior on the `master` branch. If the problem persists there, it's likely unrelated to your changes. In that case, please create a separate issue for it and include clear steps to reproduce the bug.

Thanks.

lilyclements (commented):
@Emmanuel-Afrifa awesome! Nice one, it works really nicely. The issue you reported with the column metadata being blank at the first try is also an issue in the master branch. This means that it's not a bug you introduced. As a result, I will merge this, and report that as an issue in our issues.
It would also be great if this worked for the "Data metadata" as well - you've set it up nicely for the "Column Metadata"

PR #9815, opened 2025-06-13
Adding Select options to Prepare > Column:Text > Find/ Replace dialog

### LINKED ISSUE #4892: Small (I hope) change in Prepare > Column: Text > Replace All dialogue
Could we make the single receiver into a multiple receiver?
When it is a multiple receiver we hide or disable to saving option.
Then it saves back into the same columns.

I think saving multiple columns into the same worksheet is not too difficult.  If not it is needed elsewhere as well, so I hope it can be added?  (Perhaps it is similar - or the same as when we save both the residuals and fitted values after fitting a model?)

I address to Danny first, largely to check that this is a simple change.

anastasia-mbithe:
Fixes #4892 

@rdstern I have added the Select Options under the Replace button.
Everything works well when the Replace Options: First Occurence and Replace All are selected.

However for Cell, I will have to wait for @lilyclements to help with the right code.


anastasia-mbithe:
@rdstern and @Vitalis95 This is ready for testing.

lilyclements (commented):
@anastasia-mbithe this is great! Just two small changes in the R code and then it should be good to go! These are both changes made in the `Column: Text > Split` dialog when similar changes were implemented there last month.

1. Can `select` instead equal the value in the ucrSave (e.g., `"detect"`) instead of the data frame name? (so `select <- "detect"`, say).
2. Can we add to `col` at the end for 

```r
%>%
  dplyr::select(-any_of(names(survey)))
```

So our R code becomes, say:

```r
col <- survey %>%
  dplyr::mutate(dplyr::across(everything(.), .fns=~replace(., grepl("SABEY", .), values="KESEN"),
                              .names=paste0("{.col}_", select))) %>%
  tidyr::unpack(cols=ends_with(paste0("_", select)), names_sep=".") %>%
  dplyr::select(-any_of(names(survey)))
```

Looks great!

(I've just updated on the issue to have those code changes too if that helps)

anastasia-mbithe:
@lilyclements I'm not sure I understood the first change in the ucrSave, Set it to "select" please check if it's what you meant.

lilyclements:
@anastasia-mbithe to the extra `select <- "detect"` thing I just meant:

```r
new_name <- "NEWNAME" # our new column name (suffix)
col <- survey %>%
  dplyr::mutate(dplyr::across(everything(.), ~stringr::str_replace_all(string=.x, pattern = stringr::coll("B", FALSE), replacement="_hello_") %>%
                                as.data.frame(), 
                              .names = paste0("{.col}_", new_name))) %>%
                              tidyr::unpack(cols = ends_with(paste0("_", new_name)), names_sep = ".") %>%
dplyr::select(-any_of(names(survey)))
data_book$add_columns_to_data(data_name="survey", col_data=col, before=FALSE)
rm(list=c("col", "survey"))
```

Here I've called it `new_name <- << new column name suffix in the ucrSave >>`
Does that make sense? We just want that `new_name` to be in there

anastasia-mbithe:
@rdstern one thing I noticed with the new select function(in this dialog), In replace first and replace all it works when the columns are Factors, Replace cell with select will work when the columns are in Character format.

anastasia-mbithe:
Hi @lilyclements , some issues with the Select option in this dialog:
When multiple is checked:
1. Replace First occurrence and Replace All work when the columns selected are Factors only and bugs when using Characters
2. Replace Cell works when columns are characters, when they are factors It returns NA 
3. The auto update of the newname from the ucrSave as requested, is not functioning as required, tried it with Vitalis but we haven't figured out yet how to make it work appropriately.   
Just thinking if the dialog needs to be used next week, we can have it as it is, then the ucrSave issue can be checked into after the workshop

lilyclements:
@anastasia-mbithe I've replied below: 

> 1. Replace First occurrence and Replace All work when the columns selected are Factors only and bugs when using Characters

This is working for me. What issue are you getting?

> 2. Replace Cell works when columns are characters, when they are factors It returns NA

This is because you are replacing a whole cell value with something that isn't defined as a new factor level. The user would have to first define it as a new factor level. This is OK for now, we could make it clearer by adding something to the help or a tooltip?

> 3. The auto update of the newname from the ucrSave as requested, is not functioning as required, tried it with Vitalis but we haven't figured out yet how to make it work appropriately.

Two things:

1. Move the `%>% as.data.frame()` to be _after_ the `dplyr::select()`
2. You are setting `select <- selection1`, or select to be your selector name. You want it to be your ucrSave name:

```r
select <- "NEWNAME"
col <- ((survey %>% dplyr::mutate(dplyr::across(everything(.), ~stringr::str_replace_all(string=.x, pattern=stringr::coll("9", FALSE), replacement="L"),
                                                .names=paste0("{.col}_", select))))) %>%
          tidyr::unpack(cols=ends_with(paste0("_", select)), names_sep=".") %>%
  dplyr::select(-any_of(names(survey))) %>%
  as.data.frame()
```

anastasia-mbithe:
Hi @lilyclements , maybe we should have a call today about this issue.

anastasia-mbithe:
@lilyclements this was completed and its ready for testing.

lilyclements (commented):
Looks good. One very small comment just on syntax, but otherwise all good. @rdstern over to you

anastasia-mbithe:
Thank you for noticing that, just one line was different so edited it to this 

image

rdstern:
@anastasia-mbithe that's very nice.  I like that the Tricot data sets are an obvious example, so I tried with the breadwheat

image

This is great and really useful, both in what it does and the message it sends, namely that dialogs can be reasonably efficient.

However, sorry after my long breat, but four questions for you:

a) The saving label doesn't change.  I assumed the method and changes would build on the Prepare > Numeric > Transform dialog, here:

image

Note the Position button is absent.

b) Is there a way of saving back into the same variables?  That's what I would usually want to do?  Could there be a checkbox for that?  (It is easy with a single variable.)  

c)  I should have checked earlier, but wonder why it is only for Replace?  Why not the other options?  Saving back into the same variable may not apply, (see b), but I assumed this would be for the dialog, rather than just one tab.  (And I am hoping the code will also be easily transferrable, e.g to the Taxt > Transform dialog.  

(Could the generalision to more options and more dialogs, become a simple task, for one of the others in Kakamega, and you check?)

anastasia-mbithe:
@rdstern for the Position button being absent in Numeric::Transform dialog is it by design or we should have it there? 
If so, that can be done in a separate issue.

lilyclements (changes requested):
1. If "overwrite" is selected, can the ucrSave be disabled?
2. Did we want overwrite as an option even if "single" is selected? (Or is this what you meant by doing bits in a new PR)
3. How come it always runs this line:

```r
data_book$add_column_selection(data_name="survey", name="selection", column_selection=list(C0=list(operation="tidyselect::where", parameters=list(fn = is.factor))), and_or="|")
```

This line was meant to be an example of setting a column selection, but I didn't mean for it to be in this dialog -- just as part of sample code for testing. Apologies for the confusion

anastasia-mbithe:
@lilyclements to respond on the above
1. I can make that happen
2. Yes, Roger had mentioned having it for the whole dialog, such that there is always that option. Also having the multiple in the other tabs not just replace. These can be done separately from this PR.
3. I can remove that since its a before code
4. Lemme look into that

anastasia-mbithe:
@lilyclements This should be good now.

For the last comment the bug appears in Survey because of the Factor to Character conversion. When I use Breadwheat, it works since the variables are characters already so no effect to the select.

lilyclements:
@anastasia-mbithe perfect, great! Let me review again now, but your codes are looking great. Thanks for sharing!

lilyclements (dismissed):
@anastasia-mbithe this is excellent. Well done!
@Vitalis95 or @Patowhiz can you have a look at the VB code and approve if you're happy
@rdstern over to you too if you wanted another review

lilyclements (approved):

lilyclements:
@anastasia-mbithe can you write up the new changes that need to be made into an issue? (or are they already reported?)

PR #9818, opened 2025-06-16
Amending ucrAdditionalLayer to have up/down controls

### LINKED ISSUE #9713: Amending ucrAdditionalLayer to have up/down controls
It would be great if our ucrAdditionalLayer had a way that you could move the items up and down. This would be like we see in our `ucrReorder`. 

E.g., here, it would be nice to be able to move our layers up and down:

Image

This would help with issues such as seen in #4352 as you could add the Density on the main dialog so you can speicify you want the "counts" option, then add the Histogram layer on the sub, and then switch the layers.

bayakarisa:
fixes #9713

@rdstern @Ag-Derek @lilyclements I did what was discussed
@berylwaswa kindly check

### Amending ucrAdditionalLayer to have up/down controls
**now it's possible to move our layers up and down**

reodeerlayers

i used the survey data


Ag-Derek:
@berylwaswa can you review this 
Thanks

lilyclements (commented):
@bayakarisa this is great and a really impressive start! There are a couple of bits:

1. There should be a difference between the two up options. One should take it to the very top (^^) and the other should just move it up one place (^). Similarly for the down option.

2. The order doesn't actually change in the R code of the layers. E.g., let's have a look at the boxplot dialog for the survey data. I ask for a boxplot with just "yield" as the x variable

image

Then I ask for the jitter option to come in too. 

image

By default, we get the boxplot then the jitter plot:

```r
# Dialog: Boxplot
last_graph <- ggplot2::ggplot(data=survey, mapping=ggplot2::aes(y=yield, x="")) +
ggplot2::geom_boxplot(outlier.colour="red") +
ggplot2::geom_jitter() +
theme_grey() + ggplot2::theme(legend.position="none")
```

If I rearrange the arrows, I would expect to get the jitter then the boxplot. So I would want this to now be my R code:

```r
# Dialog: Boxplot
last_graph <- ggplot2::ggplot(data=survey, mapping=ggplot2::aes(y=yield, x="")) +
ggplot2::geom_jitter() +
ggplot2::geom_boxplot(outlier.colour="red") +
theme_grey() + ggplot2::theme(legend.position="none")
```

I'm not sure how advanced that bit is to implement - it might require a chat with @Ag-Derek to help outline it, but hopefully it should be reasonably straight forward. 
I can understand if we have to have boxplot as the first layer, but then would it at least be possible for any layers added in the sub-dialog to be moveable? E.g., if we add in geom_point, then to be able to change the order of geom_point and geom_jitter?

Here's an image of me adding geom_point, and I set size = 10 so we can tell easily if it switches order with the jitter

image
image

Ag-Derek:
@bayakarisa please do you have any updates so far on what lily asked you do to

bayakarisa:
> @bayakarisa please do you have any updates so far on what lily asked you do to

I am still working on this

bayakarisa:
@rdstern @Ag-Derek @berylwaswa Kindly  check 
1. There's now a  difference between the two up options. One takes it to the very top (^^) and the other  moves it up one place (^). Similarly for the down option.
Reorderlayers

Here is the result 
geom_point

PR #9819, opened 2025-06-16
Added Linked Station Data in Define Climatic Data Dialog

### LINKED ISSUE #9736: Fix the Define Climatic Data for situations where there is linked station data
@lilyclements and @MeSophie this is an exciting extension to the define climatic data that was done by Sophie, buit never really tested - and it doesn't work yet - bi=ut I think it is close.  And it is perfect for our Tricot climatic work.

Image

Here is the data:

[tricot test of date filter2.zip](https://github.com/user-attachments/files/20383306/tricot.test.of.date.filter2.zip)

Sophie I don't see how this can work, because we need 2 station ids to be able to link them.
Could you have a button and call it `Linked Station Data` - which is probably clearer than metadata?  

Then it opens a simple sub-dialog with the data selector and those receivers.  Then the Station on the main dialog is always from the climatic data dateframe, while the sub-dialog gives the station name in the station-level dataframe.

You might include a Check Duplicates control on the sub-dialog too.  There it is just the station that must be unique.

MeSophie:
Fixes #9736
@rdstern @lilyclements @Ag-Derek I added Linked Station Data in Define Climatic Data Dialog. Please Have a look.

### DIFF
--- a/instat/DlgDefineClimaticData.vb
+++ b/instat/DlgDefineClimaticData.vb
@@ -25,7 +25,7 @@ Public Class DlgDefineClimaticData
     Private lstReceivers As New List(Of ucrReceiverSingle)
     Private lstNewReceivers As New List(Of ucrReceiverSingle)
     Private lstRecognisedTypes As New List(Of KeyValuePair(Of String, List(Of String)))
-    Private clslLinkedAnyDuplicatesFunction, clsLinkedConcFunction, clsDefaultFunction, clsLinkedDefautFunction, clsLinkedGetColFunction As New RFunction
+    Private clslLinkedAnyDuplicatesFunction, clsDefaultFunction, clsLinkedDefautFunction, clsLinkedGetColFunction As New RFunction
     Private clsAnyDuplicatesFunction, clsConcFunction, clsGetColFunction, clsDummyFunction As New RFunction
     Private strCurrentDataframeName As String
     Private bIsUnique As Boolean = True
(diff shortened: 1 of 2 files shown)


rdstern:
@MeSophie from first impressions this looks great.  Well done.
One small thing is that the checking of duplicates in the sub-dialog control has the 2 terms overlapping.  Could you make them smaller.
It also takes a long time to open this sub-dialog?  Is there a reason for that?  

The main improvement is that it should also make the nstation in the linked file into a key variable.  And then link that to the station id in the other data frame.  I assume that may become a new data_book function? Can you do that, or will it have to wait for @lilyclements next week?

rdstern (approved):
@MeSophie I am using these on day 3 next week.  So I am approving now, because it would be good to get the dialog merged.  
@lilyclements there is a function I think needs editing and perhaps with a new name, but that will be invisible to users.  Can you see if you can approve as it is, and merge for now?

lilyclements:
@MeSophie works great! The bit of code you want to add is:

```r
data_book$add_link(from_data_frame="data_RDS_stacked", to_data_frame="covar", link_pairs=c(id="id"), type="keyed_link", link_name="link")
```

* `from_data_frame` - this gives the name of our data frame on the main dialog
* `to_data_frame` - this gives the name of our data frame on the subdialog
* `link_pairs=c(id="id")` - here `id` is the name of our station variable in the main dialog, and `"id"` is the name of our station variable in the sdg
* `type="keyed_link"` will always be the case
* `link_name="link"` is the name of the new link. I'm not sure if this overwrites if there's currently a link between them and it's called "link". 

I generated this from using dlgAddLink, so if you're uncertain on how to do something in here (e.g., the `id = "id"` bit), then have a look at that.

MeSophie:
> @MeSophie are you in a position to work on adding the link now? If not, I will merge this as per @rdstern's request and we can add this as a new issue for a separate PR.

Yes @lilyclements  I can do it.

lilyclements (approved):
@MeSophie great stuff!

PR #9822, opened 2025-06-19
Added Click Reset Warning

### LINKED ISSUE #3709: Clicking Reset giving a warning
This was discussed for 0.4.1, would you mind adding a milestone for that and removing this comment @dannyparsons 

When clicking reset, it should give a warning to let the user know it will clear everything. This is particularly important for dialogs, such as General Graphics where you may build lots of layers and then not realise you will lose it all by clicking reset. There are other dialogs like this which would be good to have this feature. For consistency, we will have it in all dialogs. However, because it may be irritating to the user, it should have a checkbox saying "Don't show me this again" and if checked, the warning does not come up again that session.

Ag-Derek:
Fixes #3709
First implemented a warning message when the user clicks reset, this class message in only implemented in dlgTraits for now. 
This is not ready for review


Ag-Derek:
@lilyclements i think you can have a look at the warning message now (Only implemented on dlgTraits) and let me get the feedback

Ag-Derek:
Sorry i forgot to add it in the description,  there is check in the Edit menu called disable warning message it acts as a global level for disabling the warning message. it also doubles as disabling warning message for when a user delete a column

Issue #9825, opened 2025-06-23
Change default for column factors in the Describe > Tables > Summary dialog


Ag-Derek:
In the Describe > Tables > Summary Dialog we want to change the default for Column: Factors from 0 to 1

PR #9834, opened 2025-06-23
solved issue #9825 partially - changed the default for Column: Factors from 0 to 1

Rockjunior:
Set ucrNudColFactors.Minimum = 1 to make 1 the new default instead of 0 for Column: Factors.

PR #9835, opened 2025-06-23
Changing minimum value for Column Factor

Rockjunior:
@lilyclements this now solves issue #9825 and PR #9834

### DIFF
--- a/instat/dlgSummaryTables.vb
+++ b/instat/dlgSummaryTables.vb
@@ -616,17 +616,17 @@ Public Class dlgSummaryTables
         Dim selectedVariables As List(Of String) = ucrReceiverFactors.GetVariableNamesAsList ' Example, adjust based on your control
         Dim selectedCount As Integer = selectedVariables.Count
         ' Ensure ucrNudColFactors.Maximum does not exceed the number of selected variables
-        If selectedCount > 0 Then
+        If selectedCount > 1 Then
             ucrNudColFactors.Maximum = selectedCount
 
             If ucrNudColFactors.Value > selectedCount Then
                 ucrNudColFactors.Value = selectedCount
             End If
 
-            ucrNudColFactors.Minimum = 0
+            ucrNudColFactors.Minimum = 1
 
         Else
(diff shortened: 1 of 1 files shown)


lilyclements:
partially fixes #9825

instat/dlgSummaryTables.vb
```diff
@@ -616,17 +616,17 @@ Public Class dlgSummaryTables
         Dim selectedVariables As List(Of String) = ucrReceiverFactors.GetVariableNamesAsList ' Example, adjust based on your control
         Dim selectedCount As Integer = selectedVariables.Count
         ' Ensure ucrNudColFactors.Maximum does not exceed the number of selected variables
-        If selectedCount > 0 Then
+        If selectedCount > 1 Then
             ucrNudColFactors.Maximum = selectedCount
 
             If ucrNudColFactors.Value > selectedCount Then
                 ucrNudColFactors.Value = selectedCount
             End If
```
lilyclements, line 624: @Rockjunior not quite - you've changed the **minimum** value for the NUD, but we want to change the **default** value.

You want to set a default value for `ucrNudColFactors` when there is at least one variable in the "Factors" multiple receiver.

I think what you want to run is here:

```r
        If selectedCount > 0 Then

' if we have one item in the count then set as 1
            If selectedCount = 1 Then
                ucrNudColFactors.Value = 1
            End If

' then all your other bits
            ucrNudColFactors.Maximum = selectedCount

            If ucrNudColFactors.Value > selectedCount Then
                ucrNudColFactors.Value = selectedCount
            End If
```

(Note it might not be `ucrNudColFactors.Value`. @Ag-Derek might be able to know the exact code, or have a play around with the options you can see)

lilyclements (commented):
@Rockjunior just one small comment

Ag-Derek:
@Rockjunior 
You should input 
`ucrNudColFactors.SetRDefault(1)`
in the 
`InitialiseDialog()` section of the dialog and undo all the changes you have done

PR #9842, opened 2025-06-25
adjusted the default value for column factor

Rockjunior:


Rockjunior (commented):
@Ag-Derek hope this now set

lilyclements (commented):
@Rockjunior @derekagorhom the issue here is that it sets the default as 1. But then it sets it to be "0" because we have no column factors, I think. We want instead to run:

```r
            If selectedCount = 1 Then
                ucrNudColFactors.Value = 1
            End If
```

I've added that in the appropriate place into your branch in [this commit](https://github.com/IDEMSInternational/R-Instat/pull/9842/commits/ad1daf7251b645eb3978245b572306d63fcbfd81).

So, to fully close this issue, we just want to now do similar for `UcrNudColumnSumFactors`.
But, this one is more fun. Because in here there are two controls changing if it is "1". You don't want it to reset to 1 when the user adds in more variables, just to go to "1" the first time that we have any column factors.

Ag-Derek:
@lilyclements makes @Rockjunior kindly follow the steps above and do the same for `ucrNudColumnSumFactors`

lilyclements:
@Rockjunior how is this coming on? It would be great if we could get it in the version today - otherwise we can just merge as it is, and work on the other case afterwards

rdstern (changes requested):
@Rockjunior I think you have now half-done it correctly?  It now has a default of 1 - I think for the Frequency Tables - but not yet for the summary tables.
If I am correct, then could you implement the same change in default for the summary tables please.

lilyclements (dismissed):
@Rockjunior I've made some changes my end. This should work now. Can you review and see that you are happy with it?

lilyclements (approved):

Issue #9824, opened 2025-06-21
Issues for Day 2 Practical 1


rdstern:
a) The first is the Prepare > Column: Factor > Recode Factor.  Here is the figure, where I was recoding:

![Image](https://github.com/user-attachments/assets/04e8dd99-26f5-43ca-a491-562919d9c351)

Note the scroll bar in this dialog, in the fig. above.  In my version (Merged code) I now get:

![Image](https://github.com/user-attachments/assets/2b3a81d8-8c62-491a-89f7-c8e0f5ec06f2)

I seem to have lost the scroll bar?

b) The Define Tricot dialog?  This is not a bug, and maybe can be ignored, because this example is an exception.
Here is the current dialog:

![Image](https://github.com/user-attachments/assets/e8bb5f7b-4a20-4579-993a-7163b52b62f1)

We don't have any others in R-Instat that are so empty.  We do have some that changes size when options are checked.  I wonder if these could do the same.  I think perhaps, there is so much to do overall next week, that anything should be left until after the workshop.

c) Results from the Model > No Covariates.  

This works fine, but I don't understand the results.  I would also like the ANOVA table to be displayed more clearly and wonder if it is correct.  The results seem very different to the simpler ANOVA done earlier.

I wonder if I am incorrect that these models can be used with these data?

Could you check also which results correspond to which settings in display.  When I asked for ANOVA and nothing else extra I got this as the first results:

![Image](https://github.com/user-attachments/assets/370f06de-6921-4da3-a1fc-b92e288b55e9)

I don't understand the result, paticularly the tie2 line and the degrees of freedom?

Then the ANOVA, which is here:

![Image](https://github.com/user-attachments/assets/85a6c0a6-e90b-4b2a-8260-24ab2af4a928)

d) I rthen used the ANOVA with one covariate.

![Image](https://github.com/user-attachments/assets/69c3520a-45ea-4d26-a130-b463cf760109)

I wonder why there is no ANOVA now?

And I now get an error message:

![Image](https://github.com/user-attachments/assets/67d2ffc8-6a75-4fd5-a663-7180fa23d79f)

I think this may simply be just thatbecause it still expected the covariate to be at the genotype level?  So it may just need the error message not to be triggered.

And the result looks very sensible:

![Image](https://github.com/user-attachments/assets/0fd8ec86-2b03-49d2-9c7d-b5afd7159b92)

lilyclements:
[...]

PR #9843, opened 2025-06-25
Setting the correct data frame to read into Tricot Modelling

lilyclements:
Renaming variables; debugging setRCode, and adding data statement for `pladmm` to be the correct data.

Fixes partially #9824

@rdstern this should fix the issue you've had with the `cochran.bib` data for the One Covariate (and General) case. Can you test?


rdstern (approved):
@lilyclements this is now resolved - no error message.

I hope there will be a new version with thnis included - that would be neat!

PR #9849, opened 2025-06-27
Fixes 9848

lilyclements:


lilyclements:
@rdstern I'm going to merge so that we get this into the new issue. But do review the master copy and let me know if it is not working your end. It's working for me now

PR #9854, opened 2025-06-29
Fixes to Tricot Items

lilyclements:


lilyclements:
1. Fixes in Modelling (General and One Covariate)
2. Design for Import from ClimMob
3. Fix in typo in Display Options
4. Adding TestOKEnabled into Import from ClimMob
5. Updating ucrSave name in ClimMob - dynamically updates as a data frame is read in

PR #9884, opened 2025-07-10
Additions to the tricot transform dialog and sub-dialog

### LINKED ISSUE #9870: Additions to the Transform Dialog for Tricot Data
This is in our Tricot > Transform and Define as Tricot Data dialog. 

After updating `instatExtras` and `databook`, then we can do the following:

On the sub-dialog under the "Data" tab:

1. Add in Multiple receiver to "Specify Variety Variables".
This is added/removed by a checkbox
If checked, this reads into `databook::create_tricot_datasets` and in `databook::define_tricot_data` as a new parameter `variety_cols`
If unchecked, then variety_cols = NULL in `databook::create_tricot_datasets` and in `databook::define_tricot_data`. This is the default.

2. Add in a single input for to "Specify Rank Values".
This is added/removed by a checkbox
If checked, this reads into `databook::create_tricot_datasets` as a new parameter `rank_values`.
By default have in there `c("A", "B", "C")` when checked.
If unchecked, then `rank_values = NULL` in `databook::create_tricot_datasets`. This is the default.

3. Add in a multiple receiver to "Specify Trait Variables" in sdg.
This is added/removed by a checkbox
If checked, this reads into `databook::create_tricot_datasets` as a new parameter `data_trait_cols`
If unchecked, then `data_trait_col = NULL` in `create_tricot_datasets`

4. Add in a multiple receier to "Carry Columns" in sdg.
This is added/removed by a checkbox
If checked, this reads into `databook::create_tricot_datasets` as a new parameter `carry_cols`
If unchecked, then `carry_cols = NULL` in `create_tricot_datasets`

5. < Removed now >

6. We noticed that it is not giving the output every time that you run it to say that it is defined/etc. Is the call type correct for this when we run the `define_tricot_data` and `create_ranking_object` functions?

7. Can the check clear on (a) re-opening the dialog and (b) going from the sdg to the dialog?
It currently only resets when you press OK.

Emmanuel-Afrifa:
Fixes #9870 
@lilyclements @rdstern 

Kindly review the changes and check if everything is as expected. Thank you.

### DIFF
--- a/instat/dlgTransformTricotData.vb
+++ b/instat/dlgTransformTricotData.vb
@@ -249,7 +249,15 @@ Public Class dlgTransformTricotData
 
         sdgTransformations.SetRFunction(clsNewRFunction:=clsOutputDataLevel, clsNewDefaultFunction:=clsCreateTricotData, clsNewIDColsFunction:=clsIDColsFunction, clsNewVarietyColsFunction:=clsVarietyColsFunction, clsNewTraitColsFunction:=clsTraitColsFunction, clsNewDefineTricotDataFunction:=clsDefineTricotDataFunction, ucrNewBaseSelector:=ucrSelectorTricotData, bReset:=bResetSubdialog)
         sdgTransformations.ShowDialog()
+        ClearCheckOnSdgClose()
+        TestOKEnabled()
         bResetSubdialog = False
+    End Sub
+
+    Private Sub ClearCheckOnSdgClose()
+        bUniqueChecked = False
+        ucrInputTricotData.SetText("")
+        ucrInputTricotData.txtInput.BackColor = Color.White
     End Sub
 
     Private Sub TestOKEnabled()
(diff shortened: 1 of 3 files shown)


lilyclements:
@Ag-Derek can you advise on how to fix this problem with the failing test? I believe this branch is up to date

Ag-Derek:
@lilyclements i am having Anna to a new PR to see if she would get the check fail, if she gets it, we would have to assumed a recently merge PR had the old duplicated packages and did not update their branches

lilyclements:
> @lilyclements i am having Anna do a new PR to see if she would get the check fail, if she gets it, we would have to assume a recently merged PR had the old duplicated packages and did not update their branches

@Ag-Derek I can't see a PR that was merged since @ChrisMarsh82's, where this "issue" began. I notice your PR isn't failing - is it updated? Or do you have a version of something that we need?
I've been told that James doesn't need the new version until the week after next so we have time to fix it.

Emmanuel-Afrifa:
@lilyclements Sorry for the late reply. I will work on  your recommendations.

lilyclements:
@Emmanuel-Afrifa nice! This is looking great - good job! I'll merge it once we have fixed the issue with the test R check -- which shouldn't require anything from you. Chris is looking at this now.

Issue #7661, opened 2022-07-27
Minor bug in the Climatic Maps Dialogue


N-thony:
We need to add the facility of reducing the Station Labels size
![image](https://user-images.githubusercontent.com/68591383/181198598-f8d819df-9101-4845-997f-94325e83841a.png)

Ag-Derek:
@ARMSTRONGOPONDO can you have a look at this
Thanks

ARMSTRONGOPONDO:
Am trying to work on this issue but am trying to reproduce the map first before I do the changes could you please help with the dataset to use and how to implement the map couse I've tried doing it and it gives me errors maybe it's the way am implementing it @Ag-Derek  or @lilyclements

lilyclements:
@ARMSTRONGOPONDO a great point, and to be honest I don't know what data is used here either.

I can see a package called `spData`, and so in the "Import from Library" dialog, I've imported the first data set (afcon) as an example:

<img width="645" height="583" alt="Image" src="https://github.com/user-attachments/assets/203179d9-4c9b-4fe3-bf58-bc094a809004" />

And I played around with that on the dialog. I am not sure what is meant to go into "geometry", but I've just put the ID variable in there. I suspect we're meant to have a particular variable type go in there, but this is fine for the sake of us testing.

<img width="1139" height="664" alt="Image" src="https://github.com/user-attachments/assets/3d208752-5f65-4c8d-9695-9e4467071f69" />

1. Upon investigation, the errors you're seeing are an R code issue where the `geometry` parameter isn't being read in correctly. We should be reading in `geometry` as a parameter name (with the corresponding receiver as it's parameter value) into the R code.
2. The labels issue is fixed. There is an option to change labels size on the main dialog, and that does exactly as we want. 

The only issue here is with the `geometry` parameter which has been fixed in #9887

PR #9887, opened 2025-07-11
Calling geometry correctly in Climatic Maps Dialog

lilyclements:
As outlined in #7661

We want to call in the `geometry` parameter into the correct place when the receiver is filled, and to otherwise call in the `data_book$get_geometry(...)` option.


Ag-Derek:
@berylwaswa can you review this, You would probably need a shapefile to confirm

PR #9893, opened 2025-07-16
Improvements to Tricots dialog 

### LINKED ISSUE #9872: Small Enhancement to the "Tricot Traits" Dialog
@Ag-Derek this should be a very small one, I hope perhaps for one of the INNODEMS team?

This currently runs 

```r
....
last_graph <- plot_network(rankings_object[[1]])
...
```

We want to add some small additions:

1. Can we change the name of the dialog to "Check Traits Connectivity"
2. Can we add in two checkboxes below the single receiver.
3. Can the first checkbox be called "Print connectivity output", and this runs `instatExtras::check_connectivity(rankings_object[[1]])`. Can this be default checked.
4. Can the second checkbox be called "Print adjacency matrix", and this runs `PlackettLuce::adjacency(rankings_object[[1]])`. Can this be default unchecked.
5. Can we have some tool tips on these two new checkboxes
a) "Checks the connectivity of the network underlying a set of rankings."
b) "For "N" varieties in the trait, this gives an N by N matrix. Here, element (i, j) is the number of times item i wins over item j. For example, in the ranking 1 > 3 > 2, item 1 wins over items 2, 3, and item 3 wins over item 2."

To add for @rachelkg, this "Traits" dialog checks the randomisation of the varieties was done properly. This is more of a check (NTS: see DOI 10.1002/CSC2.20817).

Vitalis95:
Fixes #9872 
@lilyclements @Ag-Derek , added two checkboxes  for checking connectivity and printing adjacency matrix


lilyclements (commented):
@Vitalis95 I've made some minor changes. I need to pop out, so I will continue a review later, but this looks good. I think ready to merge (and if not, then I imagine only needs minor edits which I can make when I continue my review later)

PR #9880, opened 2025-07-10
Fix DataFrame reference bug in "Three Variables" tab when selector changes without reset

Rockjunior:
@lilyclements @rdstern @anastasia-mbithe kindly review
Partially solves issue #9863

PR #9895, opened 2025-07-16
Moved the cmdMissingOptions button to the correct position

Rockjunior:
Updated the position of the `cmdMissingOptions` button. Partially fixes the issue raised in [PR #9880](https://github.com/IDEMSInternational/R-Instat/pull/9880#pullrequestreview-3015987158).
@berylwaswa @Ag-Derek @lilyclements @rdstern. Kindly review and confirm if this resolves the issue.

### DIFF
--- a/instat/dlgDescribeTwoVariable.Designer.vb
+++ b/instat/dlgDescribeTwoVariable.Designer.vb
@@ -130,7 +130,7 @@ Partial Class dlgDescribeTwoVariable
         'cmdMissingOptions
         '
         Me.cmdMissingOptions.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.cmdMissingOptions.Location = New System.Drawing.Point(146, 365)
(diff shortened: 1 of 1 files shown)


lilyclements (changes requested):
Where was the "Omit Missing Values" button before, and where do we want to move it to? I noticed that I had this issue of the button being placed behind the labels in the boxes:

image

lilyclements:
I would have expected it to be next to the "Omit Missing Values" checkbox, on the right hand side of it. @Rockjunior what do you think?

Ag-Derek:
@Rockjunior I think we never call the `Omit missing value` checkbox in the skim option, so what you should have done was to make it only visible in the two variable and three variable options.

berylwaswa (commented):
@Rockjunior the omit missing values checkbox is well positioned. However it's only visible in the Two variable button, should it also be visible in the three variables @Ag-Derek ?

PR #9898, opened 2025-07-16
Addtions to the Model Options sub-dialog in the Modelling (Trees) dialog

### LINKED ISSUE #9869: Improvements in the Trico dialogs?
1. When defining a data set as Tricot it seems useful to specify one of the traits as the overall one.  To avoid another receiver could we make either the first or the last trait into the "Overall" one?  Then we could just have a checkbox on the dialog, or maybe both the transform and the define dialogs.  If so, then default could be checked, and the label could be `Last trait defined as Overall`.
2. Once defined, then on the modelling dialogs we don't give them all anymore.  We still have a multiple receiver, but only infill the Overall.  What do you think?
3. In the define, and maybe transform,  we probably want to add a Carry, maybe 2 of them.  I asked about some trials measuring yields and they would then be provided as 3 variables - so like the 3 treatments.  (There could be others, including the comparisons with the control, or anything else at the plot level.  This could be a `Carry Multiple`  and in that receiver you just have Selects.  There could also be a Carry Single and this would take single variables (like planting date, or soil type, or age or max temperature) measured at the ID level and stack them.
4.  I wonder about automatically making the current character variables into factors?  This might be automatic when the data are read using import.  And the rds files in the library could have them as factors?
5. I found the current transform doesn't seem to work with the traits defined as factors.  Can this be fixed?
6. Here is the current modelling multiple dialog:

Image

Kaue seemed to be clear that we always have plus.  So why not replace the single line receiver with a multiple receiver and omit the mini keyboard completely.  Then the user just includes all the x's and the plus is added in the code?

Emmanuel-Afrifa:
Fixes #9869 
@lilyclements @rdstern 
Please, kindly review the requested additions.
Thank you.

@lilyclements I wanted t use the tricot_example dataset you sent to me last time to test it, but no matter the dataset I put in the second selector, the check didn't pass. I presume it has something to do with the accepted data types for the expression receiver (which is currently numeric and factor). I'd like to believe this is intentional and has the expected behaviour. If so, can you recommend a dataset and how to fill in the receivers to get the check to pass? Thank you.

### DIFF
--- a/instat/dlgModellingTree.vb
+++ b/instat/dlgModellingTree.vb
@@ -614,6 +614,9 @@ Public Class dlgModellingTree
         sdgPLModelOptions.SetRCode(clsNewRSyntax:=ucrBase.clsRsyntax, bReset:=bResetSubDialog, clsNewPlacketFunction:=clsPlackettLuceFunction)
         sdgPLModelOptions.ucrChkMultivariateNormal.Enabled = False
         sdgPLModelOptions.ucrChkGamma.Enabled = False
+        sdgPLModelOptions.ucrChkGamma.Enabled = True
+        sdgPLModelOptions.ucrChkMinSize.Visible = True
+        sdgPLModelOptions.ucrChkPValue.Visible = True
         sdgPLModelOptions.ShowDialog()
         bResetSubDialog = False
     End Sub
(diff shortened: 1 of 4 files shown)


lilyclements:
@Emmanuel-Afrifa can you try converting "Gender" to be a factor in your ID level dataset? Or try converting any variables to be a factor in there. That should then allow you to test using those factor variables

lilyclements (approved):

lilyclements:
Nice! Looks great @Emmanuel-Afrifa well done

PR #9899, opened 2025-07-17
Fixed Bug Where Missing Option Button was Appearing on the Skim Option

Ag-Derek:
Replaces #9895
Partly Fixes #9880

I ensured the missing option button will not be able to appear on the Skim option but on both the Two and Three Variable options
@lilyclements , @berylwaswa , @Rockjunior can you review this 
Thanks

### DIFF
--- a/instat/dlgDescribeTwoVariable.vb
+++ b/instat/dlgDescribeTwoVariable.vb
@@ -625,7 +625,7 @@ Public Class dlgDescribeTwoVariable
                 cmdSummaries.Visible = False
             End If
             ucrChkOmitMissing.Visible = IsFactorByNumericByFactor() OrElse IsFactorByFactorByNumeric()
-            cmdMissingOptions.Visible = ucrChkOmitMissing.Checked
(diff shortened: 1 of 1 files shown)


Vitalis95:
@Ag-Derek , `cmdMissingOptions` is still misplaced. When you check Omit Missing under the "Two Variable" option and then switch to "Three Variable", the button appears at the position of the OK button.

berylwaswa (commented):
@Ag-Derek the missing options button is well positioned in the two variables button, but I can't see it in the three variables button. 

instat_V8P3nayzLk
instat_X5cm83uFDa

lilyclements:
@berylwaswa we don't want to have the missing options for the correlations option, at least not yet. This isn't set up in the same way, to my knowledge, at the moment. 

It should be there if you have it that it's creating a summary/frequency table, rather than a correlations table (e.g., factor, factor, factor)

berylwaswa:
@lilyclements thanks for letting me know. @Ag-Derek has also shown me how this works and I think this looks good the missing options button appears as it should in both places.

berylwaswa (approved):
@Ag-Derek this looks good on my end.

PR #9881, opened 2025-07-10
Fixed Bug found in General dialog title input Text

Ag-Derek:
Fixes #9879

I have added `ResetText()` to both ucrInputTitle and ucrInputTitleFooter and placed them on the clickReset command. 
@berylwaswa , @lilyclements this is Ready for review
Thanks

Issue #9879, opened 2025-07-10
A bug in the title receiver of the tables dialog


berylwaswa:
@Ag-Derek @rdstern @Patowhiz, I hope this a quick fix?

I am testing the tables subdialog, and when I moved to work on another data frame, when I reset the dialog, the title that I gave the previous table is not erased. 

I was working on the illness data frame, and now I added the pizzaplace data, but when I reset everything cleared except the title.

<img width="773" height="457" alt="Image" src="https://github.com/user-attachments/assets/14419efe-ea91-4bca-b382-640c3e83ddbd" />
<img width="476" height="606" alt="Image" src="https://github.com/user-attachments/assets/fb8ef5dc-99c0-40d8-a40d-a1d02600493b" />

Issue #9878, opened 2025-07-10
A review of the tables sub-dialog


berylwaswa:
@Patowhiz  @rdstern 

As I test the tables sub-dialog again, I have created this issue with my findings. I will make a list here and you can check it and improve what needs improving or fixing.

I have been following different examples to test the dialog and sub-dialog.

The first examples are [here](https://gt.rstudio.com/reference/cols_nanoplot.html).

Making nanoplots with the [illness](https://gt.rstudio.com/reference/illness.html) dataset.

Below is the code used;

illness |>
  dplyr::slice_head(n = 10) |>
  gt(rowname_col = "test") |>
  tab_header("Partial summary of daily tests performed on YF patient") |>
  tab_stubhead(label = md("**Test**")) |>
  cols_hide(columns = starts_with("norm")) |>
  fmt_units(columns = units) |>
  cols_nanoplot(
    columns = starts_with("day"),
    new_col_name = "nanoplots",
    new_col_label = md("*Progression*")
  ) |>
  cols_align(align = "center", columns = nanoplots) |>
  cols_merge(columns = c(test, units), pattern = "{1} ({2})") |>
  tab_footnote(
    footnote = "Measurements from Day 3 through to Day 8.",
    locations = cells_column_labels(columns = nanoplots)
  )

It produced the nanoplot below.

![Image](https://github.com/user-attachments/assets/2c4182d8-0047-4480-856c-9be1d5602776)

In R-Instat, I started with adding all the variables.

Then in the table options subdialog;

- In the header tab I added the Title label

![Image](https://github.com/user-attachments/assets/66714f03-9cfa-4bb6-94a3-7f04f2f41118)

- I then went to the stub tab and added the test column and gave it the label Test. There is no option to make the label bold as it is in the original image.
- I moved to the Hide button and hid the norm columns.
- In the nanoplots button, I selected the day columns, left it as a line plot, gave it the label, and left the hide columns check box checked.
- When I go to the footnote button to add the footnote, I get the error below.

![Image](https://github.com/user-attachments/assets/404acb5d-51c1-45e5-8e53-5d00d4549022)

- Without the footnote, I get the graph below

<img width="491" height="454" alt="Image" src="https://github.com/user-attachments/assets/c0898b04-ec74-44fa-b776-182bb4c7d441" />

The units are added in the original example but in brackets next to the tests, but here they have a column of their own. I may have to do a merge there, @rdstern. I don't know how to you can tell me.

[...]

PR #9900, opened 2025-07-17
Implements Before, Current and After Table Options

### LINKED ISSUE #9632: Add Table tab to all three Table Options sub-dialogs

Patowhiz:
Replaces PR #9758, #9808, #9878
Fixes #9632, #9879, #9881

Not yet ready for a full review.


Patowhiz:
@rdstern I’m still struggling to understand the need to include the options highlighted below directly in the main dialog. For consistency with other dialogs that use table sub-dialogs, I recommend keeping these options within the sub-dialog instead.

Having to click once more to set the title and subtitle isn’t a major issue, especially since it's the first tab, and it helps maintain a clean and consistent interface.

Additionally, I find the **Themes** option unnecessary at this stage. I suggest we wait and gather user feedback to identify which options are frequently used before deciding what to include in the main dialog.

Overall, I lean towards keeping the main dialog focused on essential options only, and letting the sub-dialog handle reusable ones. This approach promotes simplicity, consistency, and reduces code duplication and maintenance overhead.

Let me know if I should leave the highlighted options or remove them before I make the final commits.

berylwaswa:
@rdstern I think we can just group the rows I didn't see a feature that does what is on the pivot table, I'll keep looking.

rdstern:
@berylwaswa thanks for looking.  I hope we will soon also hear something from @Patowhiz .  I thought David had convinced him of the urgency of continuing his work on the tables, even before his leave.  He has been back over a week now, but so far has been a paragon of silence on R-Instat.

Patowhiz:
@rdstern as mentioned by @lilyclements above:

> @Patowhiz I am confused. My understanding was that you were going to work on this PR, or from scratch, to get the variable names that are needed from the \_data`in thegt` object. Have I understood this correctly? From our meeting with @volloholic we agreed that "other problems" would be addressed later, but for now you were to get the basics from https://github.com/IDEMSInternational/R-Instat/pull/9758 to work.

I have also responded above that this PR aims to implement same and after table sub-dialog options. It aims to make the after table sub-dialog be able to access columns of the underlying data frame that created the `gt` table.
Please let's not include discussions on other table features needed so that we make progress on what David and Lily explained was urgent. There are plenty of table features and most can be solved at the main dialog levels.

@lilyclements yes, this is what this PR intends to solve. You can have a look at the code approach as well compared to the other PR. I'm waiting for you to confirm that the after sub-dialog which is currently used by the use table dialog is now able to do what you attempted to do. Also take note of my above https://github.com/IDEMSInternational/R-Instat/pull/9900#issuecomment-3101362556 about `rowname_col` and `groupname_col`. I'm happy to jump on a quick call if that helps speed up testing and confirmation.

lilyclements:
@Patowhiz you say this is not yet ready for a full review. What is the next stage? This is working nicely for the Use Table (rows and columns tabs). Really nice stuff.

rdstern:
@Patowhiz I am afraid I am stuck at the start.

image

I made 2 simple tables, then went to the UseTable dialog, then the sub-dialog, and don't see any of the columns?

In terms of tyour strategy to finish this task I wonder if the recent pull request, by Stephen provides a model.  He is adding the capabilities to cope with quarto files.  He has listed the 25 steps that are needed and then has shown he has (so far) implemented 8 of them.  So I know exactly what I should be testing.

rdstern:
@Patowhiz very good discussion last Friday (12 September).  You were going to update your branch and make small changes.  I can't tell from above whether it is now ready for @berylwaswa and me to continue testing?  Please tell us when it is.

lilyclements:
@Patowhiz yes, I've set up `format_gt_table` so that it reads in any argument from `gt`. You can see this [here](https://github.com/IDEMSInternational/instatExtras/blob/main/R/format_gt_table.R) by the `...`, which simply means it reads in those extra parameters from `gt`

berylwaswa:
@Patowhiz, is there anything for me to test?  This seems to have taken quite a long time.  @rdstern seems keen to complete that by the end of the sprint.

Patowhiz:
@rdstern Could you clarify what causes the crash? If a PR makes the whole application crash, then it shouldn’t be merged. I’m happy to fix the issue you found if you can share the steps to reproduce it.

This PR affects other dialogs that others might be working on, so I suggest doing a quick round of testing before any conflicts arise.

rdstern:
@Patowhiz many thanks for the latest work.  The crash is when I try to use the table sub-dialog from Prepare > Data Frame > View Data.  Maybe you are right and you should then fix it yourself.  I tried the other instances Describe 2/3/ variable summarise, the main dialog and Use Table and they went in fine.  
Then you may want to check with Chris, whether he has time to check and merge quickly?  If not, should we merge anyway and keep testing in the merged version?

rdstern:
@lilyclements I have approved, so we can get this merged quickly.
Could you check quickly on this "interesting" point, that the table objects are all character variables?  I had never realised this before.  Is there someone else who could help on this?

lilyclements:
@rdstern yes, they will be character variables. I need to think about this. So, when we run our summary table, it is initially in "long" format.

e.g.,

```r
# Dialog: Import From Library
utils::data(package="ggplot2", X=diamonds)

data_book$import_data(data_tables=list(diamonds=diamonds))

rm(diamonds)

# Dialog: Frequency/Summary Tables
summary_table <- data_book$summary_table(data_name="diamonds", columns_to_summarise=c("x","y"),
                                         factors="cut", treat_columns_as_factor=TRUE,
                                         drop=TRUE, summaries=c("summary_mean"))
```

image

This is in long format, and our "value" column is all character variables. This means that they are character variables after we pivot-wider later in the calculation

However, I cannot remember the implications if we set these to be numeric. Maybe that could be a task for someone to investigate? Or @rdstern can you think of an example where we might have the values in the table as non-numeric? The only thing that comes to mind is the "Replace NA with" option -- so if you want to replace NA with "MISSING", say, then that needs to be a character variable to be able to have words that are not just "NA", and to have the numbers.

Issue #9897, opened 2025-07-16
Improve the reliability of different window settings in R-Instat


rdstern:
@derekagorhom and @lilyclement I am trying to improve the simplicity of using R-Instat for wide datasets.  By wide I meandatasets that have at least 100s of variables.

In that situationlooking at just 50 variables at a time doesn't do much.  Instead, like SPSS, I like the option of being able to swap the data and (column) metadata.  That is working Ok, and it is very nice that the options are on the toolbar as well as the view menu:

<img width="951" height="1098" alt="Image" src="https://github.com/user-attachments/assets/9edeb178-6828-47bc-a8bd-5567d251b3ca" />

However, it is a bit "flaky".  For example when I now press the Data frame metadata it stays on the Column Metadata but says it is on the data frame matadata.

<img width="952" height="1072" alt="Image" src="https://github.com/user-attachments/assets/66c15a6e-7f8e-4ea9-9608-ed47e0ea4455" />

And, similarly, when I press - from the Column metadata - on the Script window swap, it says it has - but it hasn't!

<img width="968" height="1088" alt="Image" src="https://github.com/user-attachments/assets/21f476cc-b80e-4eea-bb15-f0165fe52b0e" />

So, @derekagorhom with your computing background, I suggest you might like to take this up.

At the same time let's simplify the windows that are possible.  We currently have 2 alternatives:

a) This swapping of the data for the column-metadata, data frame metadata, or log-script windows.
b) The addition of any of those windows to the top of the display.

I suggest we make those 2 options mutually exclusive.  So:

If we have (say) column metadata visible **_in addition to the data and output windows_**, then the swapping of the windows is disabled.

And the View menu becomes correct.  Here it says the data and output are visible, when that is not true.  It is the column metadata and the output:

<img width="973" height="1079" alt="Image" src="https://github.com/user-attachments/assets/b8e9e1f9-f0f8-4e21-a952-978056393496" />

There is one exception to this mutual exclusivity.  When you are showing the column or dataframe metadata (instead of the data) you can still have the script window at the top.  (That's working already well, from the To Script button so we leave that intact.  But when the script window is open at the top, then swapping it, is disabled.

So here look below:  I have the Column Metadata open, but it says I don't!  If I click on the Column Metadata now (when it is already open) then the data opens at the top!

<img width="986" height="1093" alt="Image" src="https://github.com/user-attachments/assets/41cc49f0-11ff-474a-b2f8-6b1aa265c298" />

[...]

PR #9905, opened 2025-07-23
Return to Start Screen After Closing Data Book

Ag-Derek:
Making sure when using File>close data book, it automatically sends user back to starting screen view

fixes partly #9897

This PR fixes the issue below
@derekagorhom one addition to the changes is when I use File > Close Data Book and I have switched the Window to the Column (or maybe other) types of metadata, then I don't get the usual starting screen. I can use the curly arrow to get it back. Could that please be done automatically.

@lilyclements , @rdstern this is ready for review 
Thanks


lilyclements:
I'll approve and merge since I believe that @rdstern is half-away for a few days

PR #9911, opened 2025-07-23
Corrected Regression in facets Sub-dialog and Added Further Options

### LINKED ISSUE #9902: Correct the R code for graphs where facets included and add further options
@MeSophie I think this is one where you could usefully attack.

a) The facet row and facet column seem both to be working.  But could they also populate the Facet tab in the sub-dialog.
b) It would also be good to add the margin options to the rows and columns.  This could be in the pull-down, just adding those options, maybe as Row + O (for overall) and Col + O.  When they are chosen, then it just adds ,margin=TRUE into the facet grid command.
c) The terms in the sub-dialog are Horizontal and Vertical.  Could (Col) be added to Horizontal and (Row) to vertical?  Or should we just change them to Column and Row, for consistency with the main dialog?  I'm thinking let's change them.  
d) The R-code generated with the addition of the facet on the main dialog does not seem to clear properly when data frames are changed.  I show one error message below.

Here is an error message that has come when I have moved from the diamonds data to the survey data and the code isn't clearing.  The error still refers to the diamonds - which it can't find now of course!

Image

MeSophie:
Fixes #9902
@rdstern @Ag-Derek @lilyclements I fixed regression on Facet in Plot Options sub-Dialog and added col+O and row+O on Facet by pull-down control in R-Instat dialogs. Please have a look.


berylwaswa (dismissed):
@MeSophie This worked looks good on my end. I checked the boxplot, bar and histogram. 

@rdstern the margin options have been added to the rows and columns. It now adds margin=TRUE into the facet grid command.

And I got no error when I changed the data frames. Please take a look too. Thanks.

Ag-Derek:
@MeSophie  i am looking into the code and it looks fine but the only small problem would be the bRow0 and strFacetColO, can you rename them to say bRowZero etc
Thanks

Ag-Derek (approved):
The code looks good for me @berylwaswa the file changes are in multiple dialogs
e.g climatic boxplot, general graphics dialog, heatmap etc.
I think 15 dialog in total so if it is ok for the boxplot and histogram then you can approve and @lilyclements can merge this

lilyclements:
@rdstern I am merging since there are so many changed files. However, can you still review in case there are still bugs from your initial request

PR #9914, opened 2025-07-28
Removing unused R packages from install_Packages.R*

### LINKED ISSUE #9882: Checking and removing R packages in R-Instat
R-Instat has become a big file for download, and part of this is due to all the R packages in it.
Now we have moved things out of R-Instat and into R packages, there are some R packages that we no longer need to load in R-Instat.

So, this leads to a pretty important task which I'm hoping one of the INNODEMS team can take on: Checking which R packages are and aren't read into R-Instat. 

The task:

1. Update your branch
2. Open up `install_packages.R`
3. You'll see on line 15-193 there are a bunch of R packages we install into R-Instat. We want to check these
4. To check, we just do a CTRL+F and insert the package name, and look across the **entire solution**.
5. If that package is called in a file ending with `.vb`, then we want to keep it!
If it is only seen in a `.R` file then we can list to remove that package.
If it is not seen at all in R-Instat, then we can list to remove it.

Image

I have a video (<2 mins!) I've attached where I look through the first five packages: 

https://github.com/user-attachments/assets/dc7bcc82-50d9-4397-bd4d-75d8a3790737

```r
    "reshape2",
    "lubridate",
    "plyr",
    "dplyr",
    "rtf",
```

I've found that **the first four are all used in ".vb" files** throughout, but `rtf` is never used. So, we can comment to remove `rtf`.

@bayakarisa do you have the capacity to take on this task? Otherwise @Ag-Derek can you suggest someone to take this on? I suggest one of the INNODEMS team.

Ag-Derek:
Fixes #9882
@lilyclements below are some of the R Packages i commented out from the List 
```
openxlsx
DAAG
plotrix
candisc
Circstats
lazyeval
htmlTable
rje
faraway
lahman
RColorBrewer
colorRamps
rworldmap
rworldxtra
chillR
HistData
```
 we should double check and confirm since  i found out some packages like httr and DBI while are not called in any file are very important for relational ddatabase and HTTP request
 @lilyclements this is ready for review


lilyclements (approved):
@ChrisMarsh82 here @Ag-Derek has removed about 12 packages. Can you check you're happy with this?

Ag-Derek:
> @Ag-Derek amazing - well done for finding these. Do you suggest we keep `httr` and `DBI` then?

Yes please

lilyclements:
@Ag-Derek great! Thanks for confirming

lilyclements:
@Ag-Derek great! Thanks for confirming. I just want @ChrisMarsh82 to have a quick look too in case there is something else that is needed when removing packages (I think the system changed recently). Thanks for sorting this.

rdstern:
@Ag-Derek and @lilyclements that's great to have this list.  Out of it I have one surprise and that is the chillR package being in the list.

I suggest we do want to keep chillR anyway -- if only for the future.  And I think it should be used in the Climatic > Check Data > Fill Missing Values dialog.

That dialog gives a developer error when you go in - annoying, but not the point.  Then the Multiple button is the one.  

image

This calls `data_book$patch_climate_element`  so could you check that one?

In addition I saw that someone said R Version 4.5 is now released.  Should we update?  If so there is a famous Penguins dataset called `palmerpenguins` and this is in R-Instat, but is now in the datasets package.  So we won't need it now.

Ag-Derek:
@lilyclements , @rdstern i have added back the chillR.

PR #9922, opened 2025-08-11
Fixed Regression found in  Boxplot

Ag-Derek:
This PR fixes the issue below
image
@lilyclements @berylwaswa, while working on fixing the issue with Beryl Repo we found out that the main had a failed check, i hope this fixes it and then beryl make another pull request and it should work for her.
Thanks


lilyclements (approved):
Great find @Ag-Derek!

PR #9937, opened 2025-08-21
Add option in Multiple Receiver to return variables separated by +

### LINKED ISSUE #9901: Multiple Receiver that separates by "+"
We want this in the Plackett Luce trees modelling dialog in R-Instat.

Currently we have an input with a calculator where the user can write their model. However, we want to offer the option to have a multiple receiver instead (or as well as? Perhaps you flick between them through a radio button @rdstern?).

We want to have a multiple receiver. Usually the items in the multiple receiver are given like: `c("item A", "item B", ...)`. But, we want to have an option to give these items as `item A + item B + ...`

@Ag-Derek I'm assigning this to you. Do you know how this can be done, or if we have something that works like this already in R-Instat? 
Can you look into this, or, does this need more @Patowhiz's input?

Thanks.

Vitalis95:
Fixes #9901 
@lilyclements  @Emmanuel-Afrifa @Ag-Derek , This PR adds a new property `bSeparateAsAddition` to the multiple receiver control.
When `bSeparateAsAddition = TRUE`, the multiple receiver returns variables separated by `+`,

### DIFF
--- a/instat/ucrReceiverMultiple.vb
+++ b/instat/ucrReceiverMultiple.vb
@@ -21,7 +21,6 @@ Public Class ucrReceiverMultiple
     Public bSingleType As Boolean = False
     ' If bSingleType and bCategoricalNumeric then categorical and numeric are the only considered types
     Public bCategoricalNumeric As Boolean = False
-    Public Property bSeparateAsAddition As Boolean = False
     Public iMaxItems As Integer
 
     Private Sub ucrReceiverMultiple_Load(sender As Object, e As EventArgs) Handles Me.Load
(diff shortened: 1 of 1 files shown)


Vitalis95:
@Emmanuel-Afrifa 
Currently, the multiple receiver `defaults `to returning variables in the existing `c(...)` format.

If you want it to return variables separated by `+`, you can set this in the dialog when configuring the receiver. For example, in `InitialiseDialog()` you would add:

`ucrReceiverColumnsToBeStack.bSeparateAsAddition = True`

PR #9939, opened 2025-08-27
Improve the copy-paste in the Factor > Recode Factor dialog

### LINKED ISSUE #9669: Improve the copy-paste in the Factor > Recode Factor dialog
This is purely a front-end correction.  It has nothing to do with R.  I explain with the example I used, but any data set could be used.

I use the `cassava data` from the `gosset package`.  

Image

In the example above copy the label in row 8 and paste it into row 6.  Others (if you want more)  are the label in row 2 copied into row 12.  Row 5 is copied into row 13 and also into row 15.

It looks fine, as though it has pasted nicely.  But it doesn't work.  Note, currently you can double-click in row 6, and then copy.  That does work, and then the resulting label goes yellow.  But that is messy and should not be necessary.  Please could the ordinary Paste be made to work?

As a strictly computing task, is @derekagorhom best placed to do this?

bayakarisa:
Fixes #9669 
@rdstern @lilyclements @berylwaswa Kindly check as i have implemented what was discussed

I used the cassava data from the gosset package.

### 1. Now you can **copy and paste and it works**
Here I  changed the label from Woman to Man

gosset1

### Here is the result to show its working 
gosset2

### DIFF
--- a/instat/dlgRecodeFactor.vb
+++ b/instat/dlgRecodeFactor.vb
@@ -296,9 +296,16 @@ Public Class dlgRecodeFactor
 
             'add the parameters
             For i = 0 To lstCurrentLabels.Count - 1
-                'todo. why not use the double quotes here??
-                'Backtick(') needed for names of the vector incase the levels are not valid R names
-                clsReplaceFunction.AddParameter(Chr(96) & lstCurrentLabels(i) & Chr(96), lstNewLabels(i), iPosition:=i + 1)
+                ' Clean and normalize labels before adding to the replace list
+                Dim oldLabel As String = lstCurrentLabels(i).Trim()
+                Dim newLabelRaw As String = lstNewLabels(i).Replace(Chr(34), "").Trim()
+
+                ' Optional: reorder compound labels alphabetically (if separated by ;)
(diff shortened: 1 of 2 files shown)


lilyclements:
I want to check with @rdstern and @bayakarisa that I understand what we want here

Do we want it such that:

1. If you copy and paste, it should send the pasted cell as yellow to indicate it has been altered
2. and, if you double click a cell, it should not set it as yellow, since it has not actually been altered

berylwaswa:
@Vitalis95 please take this up and help baya

bayakarisa:
@berylwaswa, @lilyclements  , @rdstern Kindly Check
Fixes #9669
Improve the copy-paste in the Factor > Recode Factor dialog 
I used the cassava data from the gosset package.
one

 level 8 is repeated 11 times and level 6 is just once. And they are the same, but with the four times in a different order.

Now it works and i have shown it here 

two

berylwaswa (commented):
@bayakarisa @Vitalis95 I was working on the dialog, and the levels changed from 19 to 18, I expected them to become 13. So when I went back to the dialog to check, I saw what I had pasted on level six, and i just pressed enter, and the dialog broke, bringing me the error below.

recode dialog

PR #9840, opened 2025-06-25
Changes in the Plot Options subdialog

anastasia-mbithe:
Fixes #9826 

Part c is fixed, 
However still working with @Vitalis95 and @derekagorhom to fix part d.

PR #9953, opened 2025-09-11
Linking Facets in Graphs

anastasia-mbithe:
@berylwaswa @Ag-Derek 

This is in line with the changes made in #9840 
Linking the facet in the main dialog and the subdialog, solves issues the issues we had before with the facets
In this branch, I have linked facets in the Barchart, Scatter plot, Line Plot, General Graphics, and the Histogram (I've restored the older code as well (had made some changes in #9840, that were not complete))


berylwaswa (approved):
@Ag-Derek @anastasia-mbithe this looks okay on all the mentioned dialogs.

PR #9955, opened 2025-09-12
Temporarily renaming Dataframe in the Calculator

### LINKED ISSUE #9428: Problem with a numeric variable in the calculator
This originally came from @berylwaswa and I'm surprised it hasn't appeared before.  I think it may be serious enough to fix quickly.

@N-thony This initially used the `sza dataset` from the` gt package` in the library.  Start with that to see the problem.

In the dataset try the calculation `90 - sza `and you get an error.  I think I have been over-complicating the problem, which is probably simply due to the fact that sza is both the name of the variable and also the name of the data frame.  When you use it in the calculator it is trying to do the calculation using the data frame - and the error that the data sheet is not numeric is correct.

If I right-click and `rename column sza to sz `then it works fine.

To check this, I tried with our usual survey data.  I renamed yield as survey.  And it is then a mess here too.  Using the try button shows it is getting the sheet called survey and not the variable!
 
I wonder if this affects any other dialogs?

anastasia-mbithe:
Fixes #9428

@lilyclements @rdstern 
This fixes the issue with a dataset and a column in the dataset having the same name i.e `sza`

This is ready for testing.

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -112,8 +112,8 @@ Public Class dlgCalculator
         clsRemoveLabelsFunction.AddParameter("new_val", Chr(34) & Chr(34), iPosition:=3)
 
         clsGetDataframe.SetRCommand(frmMain.clsRLink.strInstatDataObject & "$get_data_frame")
+        clsGetDataframe.AddParameter("use_current_filter", "FALSE")
         clsGetDataframe.SetAssignTo("`_df`")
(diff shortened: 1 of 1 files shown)


rdstern:
@Patowhiz could you please review the code and hopefully merge.  I am becoming increasingly concerned that we have not been having the code reviews done properly recently.  I'm still hoping that Vitalis wil also be able to help here, but that has not happened yet, though he continues to remain involved very construcively as a volunteer.

Patowhiz (approved):
Code looks good.

PR #9967, opened 2025-09-18
Made Data Options control can open in the correct data frame

### LINKED ISSUE #9957: Data Options control can open in the wrong data frame
The problem is shown below.  It applies to the Data Options control button, which is on most dialogs.  It mainly gives access to the filter dialog from within any other dialog.  Here is the route to an example:

a) Open the usual survey dataset
b) Use the Prepare > Data Reshape > Column Summaries dialog
c) Summarise yield, then use 2 factors - Village and Variety.
d) Use the summaries button and change sum to mean.  Maybe add max and min.  (It doesn't matter what summaries you use.)
e) Press Ok and you should be a second data frame with 10 rows.
f) From the toolbar, return to the last dialog.  See below:
g) Notice the data frame (survey) is in red.  That's a warning that maybe we are in the wrong data frame, but this time it is fine.
h) Press the Data Options button to get the figure below:

Image

This is the problem, namely it should also use the survey data frame, and not the currently visible one.
I assume this should be a trivial change.  It needs to be done well, because it is in a control that is in most dialogs.

Vitalis95:
Fixes #9957 
@rdstern @lilyclements @Ag-Derek ,This PR ensures that the Data Options sub-dialog always opens with the correct data frame from the parent dialog


rdstern (approved):
@Vitalis95 many thanks, seems fine.
@Patowhiz please could you check and merge.  I suspect it will be trivial for you, but it is a change in a control that's in most of the dialogs.  So important.

lilyclements:
@Patowhiz please review here too so we can get this into the new version

Issue #9960, opened 2025-09-17
New Dialog: Import from e-PICSA


lilyclements:
This comes under the EPICSA contract work. @rdstern am I correct that this means for now we want @Emmanuel-Afrifa on this (or can we have an INNODEMS member on it yet?). Can the hours towards this be noted.

There is a new dialog under the "Climatic > File > Import" section. This has the name "Import from e-PICSA"

* This has 3 buttons at the top: Definitions, Stations, Data
* All three have the option "Token" with an input and button, just like in the "Export to Google Buckets". This works the same as in there, same R code too, as the input there for the token.
* We then have a label "Country" with a drop down input. This has the same options as the "Country drop down" input box in the "Export to Google Buckets" dlg.

**If Definitions is checked:**
* There is an editable input, with label "Definitions ID:". This takes any value. 
* There is a button titled "Find Files", and a non-editable drop down next to this. This drop down will be filled with values once "Find Files" is pressed. This is just like in the dlgImportFromClimMob dialog.

The R Code for this is in two. First, if Find Files is pressed:

```r
```

Then, if OK is pressed:

```r
```

Then Definition ID: [ ] followed by a Button with the label Import
Then there are Definitions : [  ]. They become available once the Import button is pressed. Usually choose the one with the latest date.
This would import all the definitions. They are objects and a short report on them is provided in the output window.

The second tab is Station and is only used if corrections are needed to the Station data. This would import the station details into a separate data frame.

The third tab is the summary data. This again imports into a separate data frame. It is only used if the staff want to examine and check the data.

Emmanuel-Afrifa:
@lilyclements Please, can I work on this or should I wait for @rdstern response first? Thnk you

lilyclements:
@rdstern and I discussed it in person yesterday. If you're ready to work on it then that's great! Note you need to update the `epicsawrap` R package

Emmanuel-Afrifa:
@lilyclements Well noted.

Emmanuel-Afrifa:
@lilyclements and @Ag-Derek Please, is someone working on `frmMain` now? If not, I'd like to add the empty new dialog to `frmMain` so that it can be merged now to avoid any conflicts later, and I can now start populating the dialog.

lilyclements:
@Emmanuel-Afrifa apologies, I just saw this. Thanks for checking. To my knowledge, no one is on `frmMain` at the moment

Emmanuel-Afrifa:
[...]

PR #9973, opened 2025-09-23
Added the Import from E-PICSA dialog.

Emmanuel-Afrifa:
Fixes partly #9960

@lilyclements I have worked on all the things you mentioned with the exception of the `Data` tab and also the issue with the `Find Files` button on the `Definitions` tab. 

I thought I should push it so that you can check and give any feedback. Then I'll be able to work on all the needed bits once you send me your response.

For easy reference, I'll mention that problem here again.

> Please, can you help me resolve the error below?
> 
> It happens in the `list_definition_versions` function that runs internally when I click on the `Find Files` button.
> 
> Image

### DIFF
--- a/instat/dlgImportFromEPicsa.Designer.vb
+++ b/instat/dlgImportFromEPicsa.Designer.vb
@@ -22,21 +22,242 @@ Partial Class dlgImportFromEPicsa
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()> _
     Private Sub InitializeComponent()
+        Me.ucrPnlImportFromEPicsa = New instat.UcrPanel()
+        Me.rdoDefinitions = New System.Windows.Forms.RadioButton()
+        Me.rdoData = New System.Windows.Forms.RadioButton()
+        Me.rdoStation = New System.Windows.Forms.RadioButton()
+        Me.ucrInputTokenPath = New instat.ucrInputTextBox()
+        Me.lblToken = New System.Windows.Forms.Label()
+        Me.cmdChooseFile = New System.Windows.Forms.Button()
+        Me.ucrInputComboCountry = New instat.ucrInputComboBox()
+        Me.lblCountry = New System.Windows.Forms.Label()
+        Me.ucrInputDefinitionsID = New instat.ucrInputTextBox()
+        Me.lblDefinitionsID = New System.Windows.Forms.Label()
+        Me.cmdFindFiles = New System.Windows.Forms.Button()
+        Me.ucrInputComboFindFiles = New instat.ucrInputComboBox()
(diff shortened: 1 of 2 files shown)


lilyclements:
@Emmanuel-Afrifa can you try wiht

`country="internal_tests"`

and

`definition_id="250909"`

I'll set a better error meanwhile for an unfound definition_id!

Emmanuel-Afrifa:
@lilyclements Still clicking on the `Find Files` button results in the `Invalid token error`. 

In the code snippet below, the `file` parameter is supposed to be read from `ucrInputComboFindFiles`. And the values for this combo box are to be derived from the `epicsa::list_definition_versions` function, which runs internally when the `Find Files` button is clicked. Since that function is currently giving us the `Invalid token` error, we're not getting anything from it, and hence the reason why the `file` parameter is set to an empty string. Other than that, everything should work fine.

So should I take out the `file` parameter for now?

```
last_definition <- epicsawrap::get_definitions_data(country="internal_tests", definitions_id="250909", file="")
```

Emmanuel-Afrifa:
@lilyclements Alright. So just to be clear, I'll run the function below both internally and in the script window, right?

```
epicsawrap::gcs_auth_file(filename="C:/Users/HP/Downloads/epicsa_token.json")
```

Please, will that account for the http error, or we'll take a look at that later.

Lemme do this now.

lilyclements:
@Emmanuel-Afrifa yes run in both. I’ll dig into the http error shortly

lilyclements:
To the http error, it should be that file = just the numbers, not .json, or definitions/. So. 

file = “250909.20250909171926”

Emmanuel-Afrifa:
@lilyclements Please, I have fixed both issues, and now we have no errors. Thanks for the tip. Please, I just want a quick clarification.

Although there are now no errors and things seem to work fine, there are no visible outputs (not even in the output window), unlike for the station tab (where we get a dataframe). Is that the intended purpose? If not, then perhaps I have to change the saveType.

Vitalis95:
@Emmanuel-Afrifa  after @lilyclements has checked this , we can go through the code together

lilyclements (dismissed):
@Emmanuel-Afrifa this is incredible. You've done this far quicker than planned, and picked it up very easily. It's working well.

This is a fun one, for discussion in the issue. But that's that the definitions file isn't associated to a data frame. Which means that ... how do we view it? All of our other objects are always associated to a data frame. Like a table, or a graph. Definitely a discussion to have with @rdstern, but not now.

Great stuff, and impressively fast too!

Vitalis95 (approved):
@Emmanuel-Afrifa , great one!
@lilyclements ,there were a few minor code changes  -  you can go ahead and approve again.

lilyclements (approved):

PR #9981, opened 2025-09-29
Implemented Find Next in Edit > Find dialog

### LINKED ISSUE #8941: Edit Menu  - Find Dialog
Hello @N-thony and @rdstern 

I notice on the find dialog that after you click find it highlights the row and then "Found Row" shows up on the dialog saying which row you are on.

It would be much more helpful if this could tell you how many instances of the item there are, that way I would know how many times to click find and it would feel like an overview of the data. 

Could we change this? 

Also I notice that the Find Next button always seems to be greyed out. Does it work? If so how do you activate it? If not can you? 

Thanks, 

Rachel

anastasia-mbithe:
Fixes #8941 

@rdstern  and @rachelkg 
This is ready for testing.


Ag-Derek:
@berylwaswa can you have a look at this PR and review 
Thanks

Issue #9984, opened 2025-09-30
More Issues on the New Multiple Select Option.


anastasia-mbithe:
These are issues that emerged from the discussion in #9815 

- [x] Add the Multiple Select Option to the other tabs in Prepare > Column:Text > Find/ Replace dialog. Add overwriting checkbox too.
- [x] Add the Mutiple Select Option in the Column:Text> Transform dialog
- [x] Do we still want the checkbox in the Column:Numeric > Transform dialog? Since the overwriting still happens when the suffix is empty. When a user types in the suffix, it creates new columns. If so, add the Overwriting checkbox  here too.
- [ ] Factor Variables converting to Characters after Overwriting in Replace and Replace All options. @lilyclements you said "We can add in some before and after codes which check its type and converts it, like we did in the SOR dialog but I’m not so keen on that." Which is the SOR dialog?
- [ ] ucrSave name in Multiple Select; New Column Name i.e Select in Replace, does not update automatically after returning to the dialog. @lilyclements This is an issue we found sometime back, remember? normal ucrSave implemetation doesn't work in Multiple to always update

@lilyclements and @rdstern you can confirm if I left any out.

@lilyclements for the other tabs and Column:Text> Transform, I hope the select function or "frame" doesnt change alot. We apply it the same way as we did in Replace.

PR #9990, opened 2025-10-06
Added Multiple Select Options to Text> Replace dialog

anastasia-mbithe:
Fixes first part in #9984

@lilyclements I need your help in Locate All under the Find tab, and the right function for To Na Tab, these two bug, see below screenshots.
@rdstern and @lilyclements  I had left some questions in the issue #9984 about the Overwrite.

image
To Na

image
Locate All.

### DIFF
--- a/instat/dlgStringHandling.Designer.vb
+++ b/instat/dlgStringHandling.Designer.vb
@@ -47,38 +47,33 @@ Partial Class dlgStringHandling
         Me.rdoReplace = New System.Windows.Forms.RadioButton()
         Me.rdoFind = New System.Windows.Forms.RadioButton()
         Me.grpModifiers = New System.Windows.Forms.GroupBox()
-        Me.ucrChkComments = New instat.ucrCheck()
-        Me.ucrChkMultiline = New instat.ucrCheck()
         Me.cmdAddkeyboard = New System.Windows.Forms.Button()
         Me.grpDetectOptions = New System.Windows.Forms.GroupBox()
         Me.rdoEnds = New System.Windows.Forms.RadioButton()
         Me.rdoStarts = New System.Windows.Forms.RadioButton()
         Me.rdoDetects = New System.Windows.Forms.RadioButton()
-        Me.ucrChkNegate = New instat.ucrCheck()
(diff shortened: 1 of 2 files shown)


anastasia-mbithe:
@rdstern thank you for the response. 
What of Count Option under Find? do we want to overwrite there as well?

anastasia-mbithe:
This changes the ucrSave label to Suffix Name when in the Multiple option is selected and makes the Overwrite checkbox not visible in the Detect Option.

anastasia-mbithe:
@rdstern the second is implemented in #10004 
I did them separately just in case some conflicts occur before merging, so that I have to worry about few files at once.
I didn't find issues with any transformation in Numeric > Transform so I implemented for all

rdstern:
@anastasia-mbithe I jave been looking again and also trying the Overwrite.  I didn't notice it last time, because it wasn't next to the New Column Names.  So:
a) Could you move the Overwrtite control, so it is at the bottom and just next to the New Column Name - which becomes `New Column Suffix`. when Multiple is chosen.
b) A new feature (maybe for another pull request) would be to also include it also for the Single Option!  (We can already do that, by choosing the same column name, but this would be simpler!  If so it could be called Overwrite Column(s), so the label remains the same for the Single and Multiple Case.  (If that takes a bit long, then you may want to put the control there now, and disable it.
c) The Overwrite control is disabled for each option - including Detect and Count -  when it isn't desirable.  I haven't yet checked the NA stuff.  For now disable for any option where you are not sure.
d) And just to note a nice feature that is working ok, is Edit >Undo!  I am using the little survey data and it is quite sweet.
e) Finally a slightly different small change is that Reset should set back to Single.  It currently sticks on Multiple.

Overall this is very nice indeed.  @rachelkg and @berylwaswa this is also going to be good for our teaching and maybe a video!  

@anastasia-mbithe I hope the changes are small and can maybe be before the sprint?  Sorry for the delay and it would be very nice to get something merged soon.  Now let me look at your Transform!

lilyclements:
> @rdstern do we need the Replace by in the FromNA tab to be optional?

I have discussed this with @anastasia-mbithe and we have realised it does not need to be optional

lilyclements (approved):

Issue #9963, opened 2025-09-18
Addition of the legend to the mapping dialog


berylwaswa:
I am working on the mapping dialog and I just noticed that the legend checkbox is not on the main dialog as with the graph dialogs. At the moment to access the legend I have to go to the plot options sub-dialog then themes tab, I would prefer to have it on the main dialog too. Thanks. 

<img width="756" height="434" alt="Image" src="https://github.com/user-attachments/assets/e503d434-adaa-4850-9f54-69f3ca6b4995" />

Rockjunior:
@berylwaswa This PR #10003 addresses this issue

PR #10003, opened 2025-10-15
Add controls for map legend visibility and position

Rockjunior:
Add configuration options for controlling map legend appearance and location.
Addresses part of issue #9963


Rockjunior:
@berylwaswa Check this out

PR #10004, opened 2025-10-16
Added Overwrite Columns Checkbox in Prepare> Column:Numeric > Transform dialog

anastasia-mbithe:
Fixes third point in #9984

@rdstern @lilyclements 
I assumed that we would need the checkbox in this dialog as suggested by Roger for consistency as the other dialogs with Multiple select.

@berylwaswa @Ag-Derek This is ready for testing.

### DIFF
--- a/instat/dlgTransform.vb
+++ b/instat/dlgTransform.vb
@@ -746,7 +746,6 @@ Public Class dlgTransform
         ucrInputDivide.SetRCode(clsScaleDivideOperator, bReset)
         ucrInputMultiply.SetRCode(clsScaleMultiplyOperator, bReset)
         ucrChkAddConstant.SetRCode(clsConstantDummyFunction, bReset)
-        ucrPnlNumericOptions.SetRCode(clsNumericDummyFunction, bReset)
         ucrPnlNonNegative.SetRCode(clsNonNegativeDummyFunction, bReset)
         ucrChkOmitNA.SetRCode(clsMeanFunction, bReset)
         ucrChkPreview.SetRCode(clsConstantDummyFunction, bReset)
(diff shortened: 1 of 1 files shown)


instat/dlgTransform.vb
```diff
@@ -1186,4 +1200,13 @@ Public Class dlgTransform
             clsScaleDivideColsOperator.RemoveParameterByName("z")
         End If
     End Sub
+
+    Private Sub ReopenDialog()
+        'This is hardcoded here so that the checkbox is always unchecked when the dialog is reopened
+        ucrChkOverWriteColumns.Checked = False
+    End Sub
+
```
Ag-Derek: @anastasia-mbithe does this dialog still have that problem of the checkbox remembering it last setting even when you set it to not be checked in the initialisedialog block?

anastasia-mbithe: Do you mean in set defaults? I hadn't checked before but I just did and yes, it could remain checked. So we still retain this sub.

anastasia-mbithe:
I've made a few changes after testing the changes that I had made, having being suggested by the AI assist, only to find some other issue was introduced. 
I just fixed that. In the Multiple, the save sets "select" as default but also allows user typed and the checked overwrite works well too.

rdstern:
@anastasia-mbithe this is great.
a) It seems to work well, I haven't tested fully though yet.  
b) Only small point I found is that when I choose Multiple the label correctly changes from Column to Select.  But when I return to the dialog it is still correctly in Multiple, but the label is back to Column: again.
c) As with the Find/ReplaceI wonder about keeping the option of overwriting even for Single.  Therefore changing the checkbox label to `Overwrite Column(s)`.
This is great to merge once the small changes are done.
@rachelkg and @berylwaswa please note this is another dialog with an important change for the help and documentation.

Vitalis95:
@anastasia-mbithe , for the `b)` above , you can  add;

```
Private Sub dlgRank_VisibleChanged(sender As Object, e As EventArgs) Handles Me.VisibleChanged
        If Me.Visible Then
            If rdoMultiple.Checked Then
                lblSelectColumns.Text = "Select:"
            Else
                lblSelectColumns.Text = "Column:"
            End If
        End If
    End Sub
```

berylwaswa (approved):
@anastasia-mbithe this is looking good. @Ag-Derek @Vitalis95 over to you

PR #10017, opened 2025-10-27
Fixed Translation Issue On Selector Control

### LINKED ISSUE #10011: More translation points to consider?
@lloyddewit I hope you can continue to advise us and @MeSophie I hope you are ok to continue  leading the improvements in the translation system.  @berylwaswa I am pleased you remain involved and also guide others.

Image

This is a typical dialog.  

1a)  The right-click popup is not translated at all. 
1b)  There is another (simpler) in the receiver.  Also not translated.  
1c) In the Add button the down arrow gives another small menu, also not translated.  

Are these easy?  They are in most dialogs, so commonly seen.

Note at the bottom there is Store Summary - not translated.  That's another very common problem.  It is that we add controls and slightly change the text in many dialogs.  Sometimes, as here, it used to be translated - that's when it was `Save Summary`.  But the word Save was confusing, so we have changed it to Store? This is also on many dialogs.  I assume this is easy to fix and if fixed once, then it may detect (or not) the same label on each dialog, so will then be changed everywhere.

I wonder if we can handle this, in the future, by including a `Needs Translation `label to each issue, where it is relevant.  Then the actual fixes could be done in batches, by addressing all the new dialogs where there is a `Needs Translation `.  Maybe, once fixed the lebel is removed?

MeSophie:
Fixes #10011
@rdstern @Ag-Derek @berylwaswa I Fixed translation issue on selector control. Please have a look.

@lloyddewit Please I need your expertise here. The translation here is cut off for texts that are too long (I assume this is also the case for Russian, as the translation is often long for that language as well). I have tried to solve this problem with AI, but to no avail. I hope that, as you are more familiar with the project than I am, you will be able to help me solve this problem. Thank you.
image

### DIFF
--- a/instat/Translations.vb
+++ b/instat/Translations.vb
@@ -16,6 +16,7 @@
 Imports System.Data.SQLite
 Imports System.IO
 Imports System.Reflection
+Imports System.Windows.Forms
 
 Public Class Translations
 
@@ -28,6 +29,7 @@ Public Class Translations
     ''' <param name="clsControl">    The WinForm control to translate. </param>
     '''--------------------------------------------------------------------------------------------
     Public Shared Sub autoTranslate(clsControl As Control)
+        ' This helper intentionally only translates Forms (TranslateForm expects a Form)
         If IsNothing(TryCast(clsControl, Form)) Then
             Exit Sub
         End If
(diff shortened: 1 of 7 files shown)


berylwaswa (commented):
@Ag-Derek please check this out because on my side the selector control is not translated

lilyclements:
@MeSophie can you summarise the issues here as to why this stopped development? Is it worth closing this, and linking to it in an issue instead?

PR #10044, opened 2025-11-01
Refactored faceting logic in dlgPICSARainfall class

Rockjunior:
Streamlined the faceting logic by removing unused operators and consolidating facet parameters. This enhances code clarity and maintainability.

### DIFF
--- a/instat/dlgPICSARainfall.vb
+++ b/instat/dlgPICSARainfall.vb
@@ -1069,6 +1069,3 @@ Public Class dlgPICSARainfall
         openSdgLayerOptions(clsPointsFunc)
     End Sub
 End Class
-
-
-


lilyclements (commented):
Looks good! Only issue I'm finding is when "none" is selected. At the moment it runs `ggplot2::facet_wrap()`, but actually we want to run nothing. We want to remove that function entirely, rather than the parameter value.

Rockjunior:
> Looks good! Only issue I'm finding is when "none" is selected. At the moment it runs `ggplot2::facet_wrap()`, but actually we want to run nothing. We want to remove that function entirely, rather than the parameter value.

Fixed Now

lilyclements (dismissed):
@Rockjunior nice! Works well and looks good to me. Good stuff. 
@anastasia-mbithe I believe you were reviewing earlier. Are you happy or do you have any comments?

PR #10046, opened 2025-11-01
Fixed QC Rainfall - Issue if Month is not a factor

### LINKED ISSUE #10036: QC Rainfall - Issue if Month is not a factor
In the "Climatic > Check Data > QC Rainfall...", when Dry Month is checked, an Omit Months selector becomes visible. This works correctly when Month is a factor.

However, when Month is numeric, the grid in the sdg appears empty, which confused users during a workshop. They didn’t realise that Month must be a factor to populate the levels.

Image

**Expected Behaviour:**
If the Month column is numeric, the dialog should provide guidance explaining why no month levels appear.

**Suggested Fix**
When the Month variable is numeric, instead of showing an empty selector, display helper text such as:

"Month variable is numeric and has no month labels.
Convert Month to a factor (e.g., via Climatic > Dates > Use Date)."

Ag-Derek:
Fixes #10036 
I have added the display helper text which will display when the Month column is not a factor in the omit button of the QC rainfall dialog
@berylwaswa @lilyclements this is ready for review

### DIFF
--- a/instat/sdgSelectMonth.vb
+++ b/instat/sdgSelectMonth.vb
@@ -55,7 +55,8 @@ Public Class sdgSelectMonth
 
                 ' Show popup message instead
                 MsgBox(GetTranslation("Month variable is numeric and has no month labels." & vbCrLf &
-                                  "Convert Month to a factor (e.g., via Climatic > Dates > Use Date)."),
+                                  "Convert Month to a factor (e.g., via Climatic > Dates > Use Date)." & vbCrLf & vbCrLf &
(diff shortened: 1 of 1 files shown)


lilyclements (changes requested):
@Ag-Derek this is good, except it doesn't update. I put a numeric in, and view the sdg and get the warning. If I then put a factor in, it doesn't update the message.

image

Patowhiz:
@lilyclements I've approved the code, so you can merge once the functionality needed is implemented as expected.

PR #10055, opened 2025-11-03
Added a dataframe selector to the Import from EPICSA dialog

### LINKED ISSUE #10052: Add data frame selector into "Import from EPICSA"
In "Import from EPICSA", when the "Definitions" tab is imported, we want to have a data frame selector.

We just need `ucrDataFrame` -- the one that looks like this

Image

Then we need to amend the Save code slightly to attach it to the data frame selector (I think this is just the line

```r
        ucrSaveNAME.SetDataFrameSelector(ucrSelectorNAME.ucrAvailableDataFrames)
```

Emmanuel-Afrifa:
Fixes #10052
This PR is ready for review


Patowhiz (approved):
Code looks good. Very minimal business logic changes.

PR #10058, opened 2025-11-03
Improving PICSA Rainfall Graphs

### LINKED ISSUE #10045: Amendments to the PICSA Rainfall Dialog
We want to add in an optional new layer into our PICSA Rainfall dialog.

Can we have a checkbox that says "Include Status", and when checked, a receiver appears.

Then, the R code, when this is checked, is:

```r
+ ggplot2::geom_rug(
    data = dplyr::filter( <data frame name> , <variable in receiver > == TRUE),
    sides = "b",
    alpha = 0.8,
    linewidth = 1
  )
```

anastasia-mbithe:
Fixes #10045 
@lilyclements @rdstern @Patowhiz this is ready for testing

I have added the Include status option, as well as the second Y Varible receiver that adds geom_segment and another geom_points functions. 
The include status adds in geom_rug to the main function. Absence of the variables, removes the functions from the main function.

### DIFF
--- a/instat/dlgPICSARainfall.vb
+++ b/instat/dlgPICSARainfall.vb
@@ -1134,6 +1134,7 @@ Public Class dlgPICSARainfall
             clsEqualToOperator.AddParameter("left", ucrReceiverIncludeStatus.GetVariableNames, iPosition:=0, bIncludeArgumentName:=False)
             clsBaseOperator.AddParameter("geom_rug", clsRFunctionParameter:=clsGeomRugFunction, iPosition:=6)
         Else
+            clsEqualToOperator.RemoveParameterByName("left")
             clsBaseOperator.RemoveParameterByName("geom_rug")
         End If
     End Sub


lilyclements (approved):
@anastasia-mbithe really nicely done! Works very nicely. Only one small point in the TestOK, but looking great.

lilyclements:
@rdstern can you review?

lilyclements (approved):

anastasia-mbithe:
@lilyclements and @rdstern this is ready

rdstern:
@anastasia-mbithe and @lilyclements this is really great.  We must get it merged soon - though I still have a small only cosmetic suggestion!  Sorry! 

Here is the current dialog:

image

a) The checkbox currently says Include Status.  Could you change it to `Include Status Variable`.  
b) Then the label says Include Status (Optional).  We don't need Include twice, but for consistency I would like to repeat Variable twice instead!  So could you change it to just `Status Variable: `.  I have deleted the Optional, because once you have checked the Status checkbox it isn't optional!  (You can see it isn't optional, because Ok is disabled until I include it.  Good coding!

@N-thony please can you confirm you are still happy with the code.  You checked earlier.  Then we can merge as soon as these small changes are made.

Now some more general observations.  I note it works for the start of the rains, both ways round!  So you can start with the dry spells and the lines go down to the second.  Or start with the simple definition and add the dry spells.

I strongly suggest we also need to write this up.  There could be a general paper on the start, etc, but I suggest the implementation in R-Instat is also worth writing, with Anastasia a key author on that one.  That could become a really punchy conference paper at least.  Let me also bring David and Danny into the discussion.

instat/dlgPICSARainfall.vb
```diff
@@ -1068,4 +1127,26 @@ Public Class dlgPICSARainfall
     Private Sub toolStripMenuItemPointOption_Click(sender As Object, e As EventArgs) Handles toolStripMenuItemPointOption.Click
         openSdgLayerOptions(clsPointsFunc)
     End Sub
-End Class
\ No newline at end of file
+
+    Private Sub ucrChkIncludeStatus_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrChkIncludeStatus.ControlValueChanged
+        If ucrChkIncludeStatus.Checked Then
+            ucrReceiverIncludeStatus.SetMeAsReceiver()
+            clsEqualToOperator.AddParameter("left", ucrReceiverIncludeStatus.GetVariableNames, iPosition:=0, bIncludeArgumentName:=False)
```
N-thony: "left" is added but never removed when `ucrChkIncludeStatus` is unchecked, please remove it in the Else

N-thony (changes requested):
@anastasia-mbithe just a small comment

lilyclements:
> Though it was in the geom_rug so when geom_rug is removed, everything in it could be removed as well.

@anastasia-mbithe this is a good point, thanks for explaining. I think in this case it didn't need to be removed, but, no worries. I will merge since we have two approvals now

PR #10067, opened 2025-11-05
Adds sorting in alphabetical order on the selector control and support for using a data frame filter on the Climsoft import dialog

### LINKED ISSUE #9219: Add an option to order the variables in the selector alphabetically
In the App and Chatbot data, we quite often have a lot of variables (hundreds, sometimes thousands). These can often be placed in a seemingly random order because of the nature of the data. This then means that when we try to add variables from the selector into the receiver, it can take some time to find them. This is true even when creating our "selects".

@fagiothree has suggested having the option to order the variables in the selector alphabetically - perhaps on a right click, or to have a little arrow in the selector box, or both.

E.g., here, have "Order ascending" as an option, and "Order descending", or something?

image

@rdstern how does SPSS handle this? I have seen SPSS data sets with a large number of variables, so wasn't sure if they had an approach.

Is this something we could possibly add? I think it would make it a lot easier to use when we have so many variables to just select them.

Patowhiz:
Fixes #9219
@rdstern @jkmusyoka this is ready for review.


MeSophie:
image
@rdstern I made a PR #10017 but I'm still working on it because when I click on Lan button the right click menu items remain in french. I hope this change will made my task easier or not.

Patowhiz:
I've unintentionally implemented support for using a **data frame filter** on the **Climsoft import dialog** in this PR, so I've updated its title accordingly.

This PR now does the following:
* Adds sorting functionality on the selector.
* Enables the use of a data frame filter on the Climsoft import dialog.

@rdstern @jkmusyoka this is ready for review. Please note that this PR should be approved soon to avoid potential conflicts, as it includes changes to the selector.

To use the data frame filter:
1. Import the station metadata.
2. Apply a filter on the station metadata data frame.
3. Open the Climsoft import dialog and select the filtered data frame in the sub-dialog, as shown in the screenshot below.

jkmusyoka:
@Patowhiz I have re-tested and found that the data frame filter method works ok. I was able to first retrieve the station metadata, apply a filter and use the filter in the dialog to extract data for the the filtered stations.

I also tested the "sort alphabetically" and found that it works fine except that the list is unsorted with every selection of a station. It is therefore inefficient with selection of multiple stations one at a time. Is this something that can be improved?

Patowhiz:
@jkmusyoka thanks for testing this. The issue you have pointed above was due to the selector not caching the condition for reloading data when it comes to database variables.

I have now fixed the issue. Kindly retest and let me know if it's fixed.

PR #10081, opened 2025-11-13
Fixed Issue with To Script button

Ag-Derek:
Fixes Partially #9897 
This PR fixes the issue of part A of the new comment below

When that happens there is an "interesting" problem when a To Script button is pressed. It then automatically opens the top window with (now) the data. That's because it is programmed to open the ntop window with the normal defaults to show the script. When the script window has been swapped the top window should not open. I hope that will be a simple change.

@N-thony , and @rdstern this PR is ready for review


rdstern:
@Ag-Derek many thanks foir the confirmation.  @Patowhiz please could you check and merge if ok?

PR #10100, opened 2025-11-19
Added Label outliers in Climatic Boxplot

### LINKED ISSUE #10078: Label Outliers in Climatic Boxplot
In Climatic > Check > Boxplot, we have a checkbox that says "Label Outliers". 

We want this to have a receiver associated to it, and then we can put a variable in the receiver which we want to get the outlier values for.

`stat_summary()` only sees the y values, not the whole data frame, so it can't "know" the year that produced each outlier. What we need to do is:

1. Pre-compute the outliers per month (with their years).
2. Plot the usual boxplot.
3. Add a `geom_text()` layer using that outlier data, labelling with the year.

So, this means that we need to do two changes. Firstly, if this checkbox is checked, and the receiver is filled, we need to run a new data frame that calculates the outliers

```r
outliers <- daily_niger %>% # the data in the selector
  dplyr::group_by(month) %>%   # whichever variable is in the "Within Year" receiver
  dplyr::mutate(is_out = tmin %in% grDevices::boxplot.stats(tmin, coef = 1.5)$out) %>%   # coef = the value in the outlier coefficient nud, and tmin is the variable in the Element receiver.
  dplyr::filter(is_out) %>%
  dplyr::ungroup() %>%
  dplyr::mutate(month = instatExtras::make_factor(month)) # we have to make this a factor since we do that in ggplot2 too to our daily data.
```

```r
# Then we run our code as usual
last_graph <- ggplot2::ggplot(data=daily_niger, mapping=ggplot2::aes(y=tmin, x=instatExtras::make_factor(month))) +
  ggplot2::geom_boxplot() +
  ggplot2::theme_grey() +
  ggplot2::theme(axis.text.x=ggplot2::element_text(angle=90, hjust=1, vjust=0.5)) +
  ggplot2::xlab(NULL) +

  # but here, we remove our stat_summary, and instead run this:
  ggplot2::geom_text(data = outliers,
                     ggplot2::aes(label = year),    # the variable in the new receiver
                     hjust = -0.2,
                     position = ggplot2::position_nudge(x = 0.05))
```

Image

Emmanuel-Afrifa:
Fixes #10078 
This PR is ready for review.

### DIFF
--- a/instat/dlgClimaticBoxPlot.Designer.vb
+++ b/instat/dlgClimaticBoxPlot.Designer.vb
@@ -58,9 +58,10 @@ Partial Class dlgClimaticBoxPlot
         'cmdBoxPlotOptions
         '
         Me.cmdBoxPlotOptions.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.cmdBoxPlotOptions.Location = New System.Drawing.Point(10, 241)
+        Me.cmdBoxPlotOptions.Location = New System.Drawing.Point(15, 362)
+        Me.cmdBoxPlotOptions.Margin = New System.Windows.Forms.Padding(4)
         Me.cmdBoxPlotOptions.Name = "cmdBoxPlotOptions"
-        Me.cmdBoxPlotOptions.Size = New System.Drawing.Size(166, 25)
+        Me.cmdBoxPlotOptions.Size = New System.Drawing.Size(249, 38)
         Me.cmdBoxPlotOptions.TabIndex = 20
         Me.cmdBoxPlotOptions.Tag = "Boxplot_Options"
         Me.cmdBoxPlotOptions.Text = "Boxplot Options"
@@ -69,9 +70,10 @@ Partial Class dlgClimaticBoxPlot
         'cmdOptions
         '
         Me.cmdOptions.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.cmdOptions.Location = New System.Drawing.Point(10, 272)
+        Me.cmdOptions.Location = New System.Drawing.Point(15, 408)
+        Me.cmdOptions.Margin = New System.Windows.Forms.Padding(4)
         Me.cmdOptions.Name = "cmdOptions"
-        Me.cmdOptions.Size = New System.Drawing.Size(166, 25)
+        Me.cmdOptions.Size = New System.Drawing.Size(249, 38)
         Me.cmdOptions.TabIndex = 21
         Me.cmdOptions.Tag = "Options"
         Me.cmdOptions.Text = "Plot Options"
@@ -85,9 +87,10 @@ Partial Class dlgClimaticBoxPlot
         Me.rdoViolin.FlatAppearance.CheckedBackColor = System.Drawing.SystemColors.ActiveCaption
         Me.rdoViolin.FlatStyle = System.Windows.Forms.FlatStyle.Flat
         Me.rdoViolin.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.rdoViolin.Location = New System.Drawing.Point(291, 12)
+        Me.rdoViolin.Location = New System.Drawing.Point(436, 18)
+        Me.rdoViolin.Margin = New System.Windows.Forms.Padding(4)
         Me.rdoViolin.Name = "rdoViolin"
-        Me.rdoViolin.Size = New System.Drawing.Size(120, 27)
+        Me.rdoViolin.Size = New System.Drawing.Size(180, 40)
         Me.rdoViolin.TabIndex = 3
         Me.rdoViolin.TabStop = True
         Me.rdoViolin.Text = "Violin Plot"
@@ -102,9 +105,10 @@ Partial Class dlgClimaticBoxPlot
         Me.rdoJitter.FlatAppearance.CheckedBackColor = System.Drawing.SystemColors.ActiveCaption
         Me.rdoJitter.FlatStyle = System.Windows.Forms.FlatStyle.Flat
         Me.rdoJitter.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.rdoJitter.Location = New System.Drawing.Point(173, 12)
+        Me.rdoJitter.Location = New System.Drawing.Point(260, 18)
+        Me.rdoJitter.Margin = New System.Windows.Forms.Padding(4)
         Me.rdoJitter.Name = "rdoJitter"
-        Me.rdoJitter.Size = New System.Drawing.Size(120, 27)
+        Me.rdoJitter.Size = New System.Drawing.Size(180, 40)
         Me.rdoJitter.TabIndex = 2
         Me.rdoJitter.TabStop = True
         Me.rdoJitter.Text = "Jitter Plot"
@@ -119,9 +123,10 @@ Partial Class dlgClimaticBoxPlot
         Me.rdoBoxplot.FlatAppearance.CheckedBackColor = System.Drawing.SystemColors.ActiveCaption
         Me.rdoBoxplot.FlatStyle = System.Windows.Forms.FlatStyle.Flat
         Me.rdoBoxplot.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.rdoBoxplot.Location = New System.Drawing.Point(55, 12)
+        Me.rdoBoxplot.Location = New System.Drawing.Point(82, 18)
+        Me.rdoBoxplot.Margin = New System.Windows.Forms.Padding(4)
         Me.rdoBoxplot.Name = "rdoBoxplot"
-        Me.rdoBoxplot.Size = New System.Drawing.Size(120, 27)
+        Me.rdoBoxplot.Size = New System.Drawing.Size(180, 40)
         Me.rdoBoxplot.TabIndex = 1
         Me.rdoBoxplot.TabStop = True
         Me.rdoBoxplot.Text = "Boxplot"
@@ -132,9 +137,10 @@ Partial Class dlgClimaticBoxPlot
         '
         Me.lblWithinYear.AutoSize = True
         Me.lblWithinYear.ImeMode = System.Windows.Forms.ImeMode.NoControl
-        Me.lblWithinYear.Location = New System.Drawing.Point(242, 264)
+        Me.lblWithinYear.Location = New System.Drawing.Point(363, 396)
+        Me.lblWithinYear.Margin = New System.Windows.Forms.Padding(4, 0, 4, 0)
         Me.lblWithinYear.Name = "lblWithinYear"
-        Me.lblWithinYear.Size = New System.Drawing.Size(65, 13)
+        Me.lblWithinYear.Size = New System.Drawing.Size(95, 20)
         Me.lblWithinYear.TabIndex = 17
         Me.lblWithinYear.Text = "Within Year:"
         '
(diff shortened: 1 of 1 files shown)


instat/dlgClimaticBoxPlot.vb
```diff
@@ -360,6 +383,59 @@ Public Class dlgClimaticBoxPlot
...
+        clsDollarSignOperator.AddParameter("x", clsRFunctionParameter:=clsBoxplotStat2Function, iPosition:=0, bIncludeArgumentName:=False)
+        clsDollarSignOperator.AddParameter("y", "out", iPosition:=1, bIncludeArgumentName:=False)
+        clsDollarSignOperator.bSpaceAroundOperation = False
+
+        clsInOperator.SetOperation("%in%")
+        clsInOperator.AddParameter("y", clsROperatorParameter:=clsDollarSignOperator, iPosition:=1, bIncludeArgumentName:=False)
+
+        clsMutateFunction.SetPackageName("dplyr")
+        clsMutateFunction.SetRCommand("mutate")
+        clsMutateFunction.AddParameter("x", clsROperatorParameter:=clsInOperator, iPosition:=0, bIncludeArgumentName:=False)
```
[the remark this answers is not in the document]
Emmanuel-Afrifa: @lilyclements I've pushed the changes.

lilyclements (changes requested):
Looking good, a couple of things:

1. Outlier Coefficient - two n's in there. Can you correct that label?  (I think that was there before, but just while you're on this dialog)
2. Have you tested this? I ran it without filling in the receiver first, and it worked. Then I filled in the receiver and the R code is missing the variables in the `group_by` and the `make_factors`:


```r
# Dialog: Climatic Boxplot
observations_unstacked_data <- data_book$get_data_frame(data_name="observations_unstacked_data")
outliers <- observations_unstacked_data %>% dplyr::group_by() %>% dplyr::mutate(is_out=TMPMAX %in% grDevices::boxplot.stats(TMPMAX, coef=1.5)$out) %>% dplyr::filter(is_out) %>% dplyr::ungroup() %>% dplyr::mutate(month_val=instatExtras::make_factor())
```

lilyclements (commented):
Oh yes, I think the outlier coefficient is an issue with the translation - I believe that's been pointed out to Sophie on another issue so ignore that one!

instat/dlgClimaticBoxPlot.vb
```diff
@@ -725,7 +822,39 @@ Public Class dlgClimaticBoxPlot
         AddRemoveFacets()
     End Sub
 
+    Private Sub ucrReceiverWithinYear_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverWithinYear.ControlValueChanged
+        If Not ucrReceiverWithinYear.IsEmpty Then
+            clsGroupbyFunction.AddParameter("x", ucrReceiverWithinYear.GetVariableNames(bWithQuotes:=False), iPosition:=0, bIncludeArgumentName:=False)
+            clsAsFactor2Function.AddParameter("x", ucrReceiverWithinYear.GetVariableNames(bWithQuotes:=False), iPosition:=0, bIncludeArgumentName:=False)
```
lilyclements: Do you know the difference between the `GetVariableNames`, `GetVariableNames(bWithQuotes := False)`, and `GetVariables`?

I think here you want to have `GetVariables`

I suggest you try running with 

`GetVariables` and `GetVariableNames` to have a look at the difference (or have a look in another dialog, one where we have GetVariables, and one with GetVariableNames, to compare).

Essentially, if you run `.SetParameterIsString` for the ucr, then that is equivalent to running `GetVariables`
Running `.SetParameterIsRFunction` is equivalent to your `GetVariablesNames`.

instat/dlgClimaticBoxPlot.vb
```diff
@@ -390,9 +466,18 @@ Public Class dlgClimaticBoxPlot
...
+        'ucrReceiverElement.AddAdditionalCodeParameterPair(clsBoxplotStat2Function, New RParameter("x", 0, bNewIncludeArgumentName:=False), iAdditionalPairNo:=1)
+        'ucrReceiverElement.AddAdditionalCodeParameterPair(clsInOperator, New RParameter("is_out", 0), iAdditionalPairNo:=2)
+        'ucrNudOutlierCoefficient.AddAdditionalCodeParameterPair(clsBoxplotStat2Function, New RParameter("coef", 1), iAdditionalPairNo:=2)
+
+        'ucrReceiverWithinYear.SetRCode(clsGroupbyFunction)
 
+        ucrReceiverLabelOutliers.SetRCode(clsRaes2Function, bReset)
         ucrSavePlot.SetRCode(clsBaseOperator, bReset)
         ucrSelectorClimaticBoxPlot.SetRCode(clsFilteredDataOperator, bReset)
+        ucrSelectorClimaticBoxPlot.SetRCode(clsPipeOperator, bReset)
```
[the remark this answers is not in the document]
Emmanuel-Afrifa: @lilyclements I'll give this another look, but I remember I tried it yesterday and it didn't work so well. That's if you use `.SetRCode`, it gives the parameter name and value (`parameter_Name = Value`) from the control setup, but in almost all of these functions, they were not keyword arguments, hence using the `.SetRCode` would have posed some problems. The alternative was to use the `.AddAdditionalCodeParameterPair`, as this allows you to change the parameter name and even specify whether it should be included or not, but as I mentioned earlier, this also wasn't working (The values weren't being read in), hence the reason I set up the parameters manually using the `.ControlValueChanged` event handlers.

instat/dlgClimaticBoxPlot.vb
Emmanuel-Afrifa: @lilyclements 
So the return type of `GetVariables` is a `RFunction`, that's the values would be read as objects, and `GetVariableNames` returns the column names (in the receiver) as strings. The `bWithQuotes` controls whether the names should be quoted (`"column_name"`) or without quotes (`column_name`)

lilyclements:
@N-thony good point and great spot! @Emmanuel-Afrifa how are you getting on?

instat/dlgClimaticBoxPlot.vb
```diff
@@ -725,6 +827,12 @@ Public Class dlgClimaticBoxPlot
         AddRemoveFacets()
     End Sub
 
+    Private Sub ucrReceiverWithinYear_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrReceiverWithinYear.ControlValueChanged
+        If Not ucrReceiverWithinYear.IsEmpty Then
+            clsMutate2Function.AddParameter(ucrReceiverWithinYear.GetVariableNames(bWithQuotes:=False), clsRFunctionParameter:=clsAsFactor2Function, iPosition:=0)
+        End If
```
N-thony: When a parameter is added, we need to also remove the parameter when the receiver is empty. For example:
```
If Not ucrReceiverWithinYear.IsEmpty Then
    clsMutate2Function.AddParameter(...)
Else
    clsMutate2Function.RemoveParameterByName(...)
End If
```

N-thony (changes requested):
@Emmanuel-Afrifa looks good, I left some comments.

instat/dlgClimaticBoxPlot.vb
Emmanuel-Afrifa: @N-thony Please, I've added the `else` clause, but I have a question. In this particular situation, I'm setting the parameter name to be the name of the variable in the receiver (`ucrReceiverWithinYear.GetVariableNames(bWithQuotes:=False)`), so in the else clause, I need to have something like,
```
clsMutate2Function.RemoveParameterByName(ucrReceiverWithinYear.GetVariableNames(bWithQuotes:=False))
```
But since we're checking that `ucrReceiverWithinYear` is empty, I presume when the `else` clause runs, `ucrReceiverWithinYear.GetVariableNames(bWithQuotes:=False)` will probably be an empty string. Won't that make it a bit redundant?

N-thony (approved):
@Emmanuel-Afrifa good
@rdstern @berylwaswa over to you.

rdstern (changes requested):
@Emmanuel-Afrifa I loaded a climatic file and got this error when I tried to use the climatic boxplot dialog:

image

Emmanuel-Afrifa:
@rdstern Please, could you retry again?

rdstern:
@lilyclements currently if you leave the control blank, then you get the actual outlier, e.g 36.8.  If you tick the checkbox labelled `With: `then it labels the outlier using that Variable.  Maybe label it `Variable:` instead?  Would that be clearer?

rdstern:
@Emmanuel-Afrifa there is still work to be done, but this graph shows we are on the right-track:
a) I first looked at all the stations together and there may be some label problems - but look at the point below first:


b) Now this is serious stuff, so useful to examine a station at a time. (This uses our filter by Station option). Then here is the first station, namely Chipata:

image

Excellent.  I have learned an important new feature - all these low outliers are in 3 consecutive months and all in 2016.  This is a great way to show that!

On the presentation side, I suggest these labels are a good size.  So, whatever that size is, don't change it!

This is all very promising.  I hope the other cosmetic changes can come soon so this can be merged.

rdstern:
@Emmanuel-Afrifa will do.  Exciting.  But I first wanted to show you this graph - which is now your work!

image

It is easy to get sidetracked now.  For example look at how small the variability of tmin is, in the 3 rainy Summer months (Dec to Feb) compared to the dry season!  And that makes you wonder we hope - is it the same for tmax?  Is it the same for the whole of Zambia, or just that station?  You are really on a pretty neat voyage of discovery.  Is that no fun at all?  And then when you think how long it has taken for the data to be colected and then put into a neat shape, why is it that hardly any time has ev er been spent in looking at the resulting data?

Anyway, back to the testing - but after dinner!

rdstern (dismissed):
@Emmanuel-Afrifa lovely, I am very happy now.
I am approving.
@N-thony has already approved, so I'll also merge!

Emmanuel-Afrifa:
@lilyclements Please, I've made the changes. Please, can you check if everything is in order now? Thank you.

lilyclements (commented):
Almost! But we want to have some labels there of the values themselves when the labels checkbox is checked, and no variable is given

Emmanuel-Afrifa:
@rdstern and @lilyclements, Thanks for the explanation. I misunderstood the needs initially.  Now, I've changed the code such that when the `Label Outlier` checkbox is checked and the associated receiver has a variable, the newly added labels are in effect, if the checkbox is checked and there's no variable in the receiver, then the old labels (before this PR) are in effect and if the checkbox isn't checked, then there are no labels. I hope this was the interaction that was needed. Could you please check if they work as expected? Apologies once again for not fully understanding what was needed.

lilyclements (commented):
Tiny tiny tiny one!

rdstern (approved):
@Emmanuel-Afrifa many thanks for continuing this week.  This now seems fine to me and I'd really like to get it merged.  I note that @N-thony has approved.  Maybe good if @lilyclements has the last word and merges, once she returns.
I'm approving.  As a serparate task, I hope that #10126 might now be quick, as a separate pull request, now this is fresh in your mind.

lilyclements (changes requested):
The first time I run this, it does not read in the `coef = 3` parameter for `geom_boxplot`.

If I then check the checkbox (Label Outliers) and press OK then it does give `coef = 3` (we can tell from the R code, but also from the fact that there are fewer outliers in the plot)
And if I then reopen, uncheck the checkbox, and press OK, it works great. So just something in the first run not reading it in.

https://github.com/user-attachments/assets/257c7f70-2315-4873-ae8c-6c6ad550f334

Emmanuel-Afrifa:
@lilyclements Oh no. It isn't fraustrating at all. I'm also eager to learn why this isn't working as I had hoped and what I may have missed, hence the reason I asked @N-thony for assistance.

N-thony (changes requested):
@Emmanuel-Afrifa @lilyclements, I had a quick look and left a few comments. The dialog works well overall, but the code could be improved to make it more stable and easier to read.

Emmanuel-Afrifa:
@lilyclements @MeSophie @Ag-Derek Fixed overlapping controls mention in #10171

PR #10103, opened 2025-11-24
Fixed SwapMetatdata bug in the main window

### LINKED ISSUE #8724: An odd sheet1 appears the first time that the column metadata is opened.
@N-thony Here is an example:

image

This extra and empty Sheet1 only appears the first time the metadata is loaded.  It is an extra array that must be a result of some of the recent improvements made by @Patowhiz .

I assume it will be easy to spot and stop.  If it takes you (or whoever you allocate) at all long, then I suggest you ask Patrick.  I assume he is not needed initially, but could check.

Ag-Derek:
Fixes #8724
This fixes the bug with the sheet appearing empty when the either of the the swap option is used.
@rdstern , @N-thony this is ready for review

### DIFF
--- a/instat/frmMain.vb
+++ b/instat/frmMain.vb
@@ -642,6 +642,7 @@ Public Class frmMain
             mnuViewSwapDataAndScript.Checked = False
             mnuSwapDataDataframeMetadata.Checked = False
             mnuSwapDataLogScript.Checked = False
+            ucrColumnMeta.RefreshGridData()
         Else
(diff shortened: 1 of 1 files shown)


berylwaswa (approved):
@Ag-Derek this is okay, there is no extra empty sheet when the metadata window is opened. @rdstern looks good to me.

rdstern (commented):
I am happy to approve.  @N-thony please could you check the code and merge if ok?

Issue #10048, opened 2025-11-01
Adding Sliders as Interactive elements in the ggplot2 Graphs in R-instat


anastasia-mbithe:
As the suggestion from @rdstern 
"Consider adding sliders and other interactive elements to R-Instat ggplot2 graphs. This could be examined generally, as a low-hanging fruit. I note that one method involves using plotly. I suggest we start with that, because we already include plotly in R-Instat. Ideally the work would include a slider for the time series graphs used in e-picsa, so R-Instat can produce something similar."

I looked into this and also found this example that works so well in R Studio, 
```
library(ggplot2)
library(plotly)

# Create a ggplot2 plot with 'cyl' mapped to the frame aesthetic
p <- mtcars |> 
  ggplot(aes(x = wt, y = mpg, frame = cyl)) + 
  geom_point() + 
  labs(title = "Example with mtcars")

# Convert the ggplot2 plot to an interactive plotly object with a slider
ggplotly(p)
```

So it includes the frame parameter (which should be a variable that can be used as a frame) and passes the object into the ggplotly function
I have tested this in R-instat,

```
# Dialog: Scatter Plot
survey <- data_book$get_data_frame(data_name="survey")
last_graph <- ggplot2::ggplot(data=survey, mapping=ggplot2::aes(x=field, y=yield, frame= fert)) + ggplot2::geom_point() + theme_grey()
last_graph <- plotly::ggplotly(last_graph)
data_book$add_object(data_name="survey", object_name="last_graph", object_type_label="graph", object_format="image", object=instatExtras::check_graph(graph_object=last_graph))
data_book$get_object_data(data_name="survey", object_name="last_graph", as_file=TRUE)
rm(list=c("last_graph", "survey"))
```
I've randomly tested in Scatter plot, boxplot, histogram and line plot.
it works for all except histogram which I get this error 

<img width="577" height="291" alt="Image" src="https://github.com/user-attachments/assets/7de18b1c-ac35-41d2-a886-36d70f3d8ea7" />

The only downside is that the output gets printed/viewed in the browser currently.

<img width="1877" height="811" alt="Image" src="https://github.com/user-attachments/assets/08a6cb06-cd1a-4909-a222-de5d23c010c7" />

anastasia-mbithe:
@lilyclements @rdstern 
This is going to be a checkbox in the main dialogs named; Add Slider

lilyclements:
[...]

PR #10115, opened 2025-11-25
Implementing Facets in Plot Options Subdialog

anastasia-mbithe:
@rdstern and @lilyclements 

This adds and links correctly the facets in the main dilogs and subdialog.
This is still in progress.

We have used vars() function in the facet_wrap(facets= vars(factor)) to work round the linking. 

The two issues @N-thony and I are working on is to have facet_grid (rows= vars(factor1), cols=vars(factor2)) when the two receivers in the subdialog are filled. The facet wrap takes facet_wrap(facets= vars(factor1, factor2)) according to the function documentation and its working well.
The other issue we working on, is when margins are checked in the sub-dialog, coming back to the main dialog, the station input should select Facet Row+O or Col+O

### DIFF
--- a/instat/sdgPlots.vb
+++ b/instat/sdgPlots.vb
@@ -2319,7 +2319,7 @@ Public Class sdgPlots
                         bReset As Boolean, Optional clsNewGlobalAesFunction As RFunction = Nothing, Optional clsNewXScaleDateFunction As RFunction = Nothing, Optional clsNewYScaleDateFunction As RFunction = Nothing, Optional clsNewFacetVariablesOperator As ROperator = Nothing, Optional clsNewColVarsFunction As RFunction = Nothing,
                         Optional clsNewScaleFillViridisFunction As RFunction = Nothing, Optional clsNewScaleColourViridisFunction As RFunction = Nothing, Optional strMainDialogGeomParameterNames() As String = Nothing, Optional clsNewAnnotateFunction As RFunction = Nothing, Optional clsNewRowVarsFunction As RFunction = Nothing,
                         Optional bNewEnableFill As Boolean = True, Optional bChangeAesParameter As Boolean = False, Optional bNewChangeScales As Boolean = False, Optional bNewEnableColour As Boolean = True, Optional bNewEnableDiscrete As Boolean = True, Optional strNewAxisType As String = "discrete")
-        Dim clsTempParam As RParameter
+
         bRCodeSet = False
 
         If Not bControlsInitialised Then
@@ -3151,6 +3151,8 @@ Public Class sdgPlots
                     Else
                         clsFacetFunction.AddParameter("dir", Chr(34) & "v" & Chr(34))
                     End If
+                    clsFacetFunction.RemoveParameterByName("rows")
+                    clsFacetFunction.RemoveParameterByName("cols")
                 Else
                     clsFacetFunction.SetRCommand("facet_grid")
                     If rdoHorizontal.Checked Then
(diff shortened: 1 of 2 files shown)


lilyclements:
@N-thony if I am following correctly, on the sdg do we have the option to do facet wrap and facet grid? I would have expected so. Shouldn't the main dlg reflect the choices on the sdg?

anastasia-mbithe:
In the subdialog, if `Margins` or` Free space` are checked, It automatically switches to facet_grid. If these two are not checked, then it uses facet wrap.
@N-thony , yes in the old code, `facet_wrap` didn't use two variables, but from the `Facet_wrap function` documentation, It can use two variables.

anastasia-mbithe:
@lilyclements @rdstern This is ready for testing. 
Kindly test using the Line Plot dialog.

Since in the main dialog, selecting of the function to run is done manually and In the subdialog, the function depends on the conditions/parameters chosen, we have linked them in a way that if the conditions set in sdgPlots satisfy Facet Row +O, on returning from the sdg, the Facet Station updates automatically to that, for consistency.

That led to adding Facet Row & Col, and Facet Row & Col + O in the dropdown, when we have both receivers in the subdialog filled. That runs Facet_grid(rows= vars(x), cols=vars(y)) and overall when margins is checked. 

However, this brings along a small issue, that we haven't solved yet, but investigating on how to solve it: These two options can't be used in the main dialog, it has to be made possible in the Plot Options by filling second receiver. So some restrictions are needed.

Please test on the existing functionality at the moment, so that it can be implemented to the other dialogs, as we figure out the issue above.

anastasia-mbithe:
@rdstern Yeah that will be nice 
@Patowhiz could you please review the code

instat/sdgPlots.vb
```diff
@@ -3175,7 +3201,7 @@ Public Class sdgPlots
                 clsFacetFunction.RemoveParameter(ucrNudNumberofRows.GetParameter())
             Else
                 clsFacetFunction.RemoveParameterByName("space")
-                clsFacetFunction.RemoveParameterByName("margins")
+                clsFacetFunction.RemoveParameterByName("margin")
```
Patowhiz: @anastasia-mbithe any reason as to why `margin` is preferred over `margins` ?
Could you also counter check that no any other dialog that uses this sub-dialog is using the parameter name `margins` to reference this parameter. Some dialogs usually use the parameter names.

anastasia-mbithe: @Patowhiz my bad it should have remained as `margins`

N-thony (changes requested):
@anastasia-mbithe looks good, just a few comments, then I would like @rdstern @anastasia-mbithe to test as well.

anastasia-mbithe:
I have removed the sliders from this dialog and branch as the discussion in #10048,

rdstern:
@anastasia-mbithe I report some oddities first.
I am using the survey data and the line plot graph.  I first got errors in relation to the scales package.
It currently is rsion 1.3 and the error said 1.4 is needed.  I installed in R-Instat and it seemed to say 1.4 was installed, but I still had the error.  I started again and it seems it didn't install properly - still version 1.3.  

I couldn't solve that in R-Instat - odd, so had to install using RStudio.  Then the line plots worked fine and seemed to synch with the sub-dialog.

Great - well done.

Except I then tried a boxplot - without any facets:

image

And it crashed with this error:

image

I'll try again!

So I tried again and roughly that boxplot worked.  And it  worked with a facet too.
Then I pressed the plot options and got the same error.  

I'll keep trying!

This time I just went straight to boxplots - I have been assuming that (as this control is in all graphs) I can do this?  

This time it was consistent with the last.

I did yield against variety - no facets - worked fine.
Added village as facet - still works fine
Pressed the plot options sub-dialog and got the error above.

rdstern:
@lilyclements thanks for all that.  In reply:
a) I am really keen on being able to use these facets in their own right and also to be able to explain what tables (in general) are all about.
b) So, I found it very exciting that we can have more than a single facet on either rows or columns.  It really is a good grammar of the graphics!
c) So I would really like to be able to have more than 2 facets.
d) Now you seem to insist that we have the same controls on the main and sub-dialog. e
e) So I suggest a single multiple control for both.  
f) It can be quite small, at least on the main dialog, but it allows any number of factors on each.
g) It won't be used often with more than 2 factors, so the first is always the columns and all the others are the rows.
h) If you want a different layout, then you tweak the code in the script.
i) I am also intrigued to know what it does with margins when there are 3 (or more) factors.  Whatever it is I am happy that we have (or at least can start with) eithert you have all margins or you have none!  
j) But if there is a chance that ggplot2 is better than a pivot table in giving margins, then I would like to continue that discussion with you!
Must get the train!

lilyclements:
@anastasia-mbithe great. Let me review and I'll merge if it is working well. We can then move the other comments into a new issue

PR #10119, opened 2025-11-27
Small addition to the Rainfall quality control sub-dialog

### LINKED ISSUE #6228: Rain day threshold for the consecutive option on Rainfall quality control
- [ ] Shouldn't there also be a rain day threshold for the consecutive option as well? If so, I would suggest there should be a dialog option for rain day threshold that will apply to all the options, instead of multiple options for this. Slightly less flexible but clearer I think.

- [x] Also, for "Dry Month" there is a label called "Missing Values Threshold" but what it's used for is the monthly rainfall total. So shouldn't this be something like "Monthly rainfall less than"?

_Originally posted by @dannyparsons in https://github.com/africanmathsinitiative/R-Instat/issues/6115#issuecomment-766786676_

### LINKED ISSUE #10083: Small addition to the Rainfall quality control sub-dialog?
I would like @jkmusyoka or @lilyclements support on this small change, before it is implemented.

Here is the dialog and sub-dialog:

Image

R-Instat is about making it simple for the users.  In the figure above I have added the extra Why variable and made it a character column.  In ZMD the provincial staff are asked to report on these months as part of their data checking.

So I would like to include an option here to add the Why variable for the user.  Then they don't need to add it as a comment, in a separate data frame.  They just add the explanation here.

Now, I don't see the need for this extra variable for all the options, in this dialog, but it is very useful here.  And the dialog is very crowded.  So I am suggesting we add a checkbox to the bottom of this sub-dialog, default unchecked.  It says Add why Variable.  If checked, then it adds this character variable, either with NA or blank as the default.

An alternative is to add the checkbox to the main dialog - or to let the user add the variable themselves, if they need it?

bayakarisa:
Fixes #6228  and
Partly fixes #10083

@berylwaswa @rdstern 
1. Changed the outlier option which currently says "Omit zero" to  "Omit below threshold"
2. changed rain days to says days > 0.05 - with the 0.05 changeable.
3. changed Consecutive default of 10  changed to 20
4. Added a checkbox at the bottom of this dialog that says "Add Why Variable"

Scre
 
###  @MeSophie I need help as this is a database problem the **rain days is not changing to days** 
 
Screen1

### Here is how it works when you just omit Months without selecting the months 

image
image

### DIFF
--- a/instat/dlgClimaticCheckDataRain.vb
+++ b/instat/dlgClimaticCheckDataRain.vb
@@ -60,7 +60,7 @@ Public Class dlgClimaticCheckDataRain
     Private clsCumulativeList As New RFunction
     Private clsIfelseFunction As New RFunction
     Private clsIsNaFunction As New RFunction
-    Private strCumulativeCalc As String = "Wet_days"
+    Private strCumulativeCalc As String = "wet_days"
     Private strCumulativeTestCalc As String = "wet_days"
     Private strDryMonthCalc As String = "dry_month"
     Private strDryMonthTestCalc As String = "dry_month_calc"
@@ -852,5 +852,6 @@ Public Class dlgClimaticCheckDataRain
             clsListSubCalc.RemoveParameterByName("why")
         End If
     End Sub
+
 
 End Class


rdstern:
@bayakarisa this is now a week old, and you have been silent.  Can I check that you are still working fully on R-Instat, so that the corrections for this dialog can be included when we make a new version this Friday (11th December) 
To help I am simplifying the suggested tasks, see below.

a)  I suggest you start again from the original code.  Maybe on a new pull request.
b) You only make one change in the dialog, namely you add the Why Variable, and it is a character variable as you did above.  Now it is right at the end.  It is the last variable you produce.  And it is only for the Dry Month option.  
c) To do this task  you include your Add Comment Variable as you did before, but 1) It is now called `Add Why Variable` and 2) it is disabled unless the Dry Month option is ticked.
d) In your changes earlier, you also allowed the limit or what is dry to be chosen and not fixed.  Do not do that.  It is less important and was not working.  If needed, it can come in a later pull request.
e) @lilyclements did respond last week, see above, and that was to enhance the Consecutive option.  Leave that also for now.  Please just concentrate on this one task.  Also no lower limit change on either option.  Please leave as before.  Other improvements can come later.

I hope this is clear.  Happy to chat, if not.

Here is the data I used in the discussion above:
[eastern_with summaries.zip](https://github.com/user-attachments/files/23982035/eastern_with.summaries.zip)

lilyclements:
@bayakarisa what is the state of this? It looks from your last message that you were waiting for @berylwaswa to review. Can you prompt her to review this? It should be the responsibility of the PR author to follow up on this

berylwaswa (dismissed):
@bayakarisa this is looking good!
@rdstern below is how the dialog looks like,

instat_OvydsTfxrX

@bayakarisa has implemented all that he listed above. I tested without errors using the same data he used. @rdstern you can do the final check.

instat/dlgClimaticCheckDataRain.vb
```diff
@@ -753,25 +796,115 @@ Public Class dlgClimaticCheckDataRain
...
+
+        ' Toggle all assignments using ternary operator
+        clsFilterMonthFunction.SetAssignTo(If(bClear, "", "filter_months"))
+        clsGroupByMonthYearFunction.SetAssignTo(If(bClear, "", "grouping_month_year"))
+        clsDryMonthCalculationFunc.SetAssignTo(If(bClear, "", "dry_month_calculation"))
+        clsWhyVariable.SetAssignTo(If(bClear, "", "why_variable"))
+        clsGroupByFunc.SetAssignTo(If(bClear, "", "grouping"))
+        clsRainFilterFunc.SetAssignTo(If(bClear, "", "rainfall_filter"))
+        clsDayFilterCalcFunction.SetAssignTo(If(bClear, "", "day_calculation"))
+        clsIdCalculationFunc.SetAssignTo(If(bClear, "", "id_calculation"))
```
lilyclements: Can you explain this bit? Why are we removing all the assignments? This gives a very horrible long long long long long function which is very difficult to read!

lilyclements (approved):

lilyclements:
This looks good. Only thing is the "Why" variable, but while that looks trivial I'm going to have to dig into the code a bit more on that! I've reported it in #10184

PR #10124, opened 2025-12-04
Adding Interactive Slider in Use Graph dialog

### LINKED ISSUE #10048: Adding Sliders as Interactive elements in the ggplot2 Graphs in R-instat

anastasia-mbithe:
This Fixes #10048 

@rdstern @lilyclements 
This works well now. Note I have named the produced plotly object "last_plot" to avoid overwriting the last_graph since the produced plot when used again it bugs, but the selector still categorizes it as a graph. (After Lily adds a plotly object, it will be easy in the code to differentiate)

@Vitalis95 @N-thony I just noticed one small issue that I didnt manage to solve. If I select my last_graph, and go to the plot options add a title, then return and  add my slider, everything works well. 
But if I don't use that order and I select the last_graph, then add slider, then go to plot options and add a title, the code produced for the before codes is repeated and on unchecking the checkbox, the before codes aren't removed. Could you look at it and advise?


lilyclements:
We should discuss the status of this, and outline what is left to do.

PR #10161, opened 2026-01-07
Changing Header to Titles in sdgTableOptions

### LINKED ISSUE #9227: Fix bug and add controls to the use table dialog
@Patowhiz the current `Describe > Tables > Presentation` dialog (and sub-dialog) is really a Presentation of a data-frame and so has essentially the same role as `Prepare > Data Frame > View`.  But there is one key distinction, and that is that `Describe > Tables > Presentation` has a whole dialog available.  So please can some presentation controls be added to the main dialog.  This then makes it more like the graphics dialogs, where the main dialog is often sufficient, then with loads of additional flexibility from the sub-dialog.

The sub-dialog still needs to remain complete, so it has everything for the View (and later the 2/3 way Summarise dialogs.  For those dialogs an option will be to save the default table and then go to the Describe > Use Table dialog.  So the Use Table will have a similar set of controls as the Presentation dialog.

Ideally the main dialog would also allow a discussion of some of the presentation concepts of a table too.

a) Title:  Here is the Header tab in the sub-dialog.  Please could we call it Titles instead to be consistent with the Graph sub-dialg.

image

Towards the top I suggest the Titles half of this tab, could usefully be copied there.  Including Format and Footnote, but not the sub-title section.

I'm liking this as I write it.  I think @Patowhiz included it originally.  The title is important, and @Patowhiz has a great `Format` sub-dialog that can be discussed here.  And footnotes are important so nice to have one to discuss from the main dialog.

b) Similarly, towards the bottom of the main dialog, could we have Themes:

We just have the standard themes on the main dialog.  If you want a manual theme, then go to the sub-dialog.

c) One item for the middle - that I don't immediately know how to do from the current sub-dialog is `Replace NA`.  How is this done?  Might it even benefit from another tab?

d) What else, in the middle?  Could there be a simple version of part of Stub?  I assume spanners are not relevant for a dataframe?  Maybe an item or two for nanoplots.  And might Column titles be possible?  (Though maybe we should rename the columns beforehand?  That would then be more for the Use Table dialog?

The three items in a), b) and c) would also be the same in the `Use Table` dialog.  I am wondering whether Drop Variables might also be useful there, especially after the 2/3 variable summaries tables is implemented?

lilyclements:
This PR changes "Header" to "Titles" in the sdgTableOptions

Fixes #9227

- [x] Runs without errors  

### DIFF
--- a/instat/UserTables/sdgTableOptions.Designer.vb
+++ b/instat/UserTables/sdgTableOptions.Designer.vb
@@ -60,29 +60,30 @@ Partial Class sdgTableOptions
         'ucrSdgBaseButtons
         '
         Me.ucrSdgBaseButtons.AutoSize = True
-        Me.ucrSdgBaseButtons.Location = New System.Drawing.Point(268, 465)
-        Me.ucrSdgBaseButtons.Margin = New System.Windows.Forms.Padding(4)
+        Me.ucrSdgBaseButtons.Location = New System.Drawing.Point(402, 715)
+        Me.ucrSdgBaseButtons.Margin = New System.Windows.Forms.Padding(6, 6, 6, 6)
         Me.ucrSdgBaseButtons.Name = "ucrSdgBaseButtons"
-        Me.ucrSdgBaseButtons.Size = New System.Drawing.Size(224, 30)
+        Me.ucrSdgBaseButtons.Size = New System.Drawing.Size(336, 46)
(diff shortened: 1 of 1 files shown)


lilyclements:
image

lilyclements:
@berylwaswa this is ready to review

PR #10164, opened 2026-01-12
Changes to Save As Dialog

### LINKED ISSUE #5702: Save As Dialog
Everytime the user wants to save data, they are first prompted by the Save As Dialog, which is one of our normal dialogs. This can be frustrating especially because users are used to the normal Save Dialog that they find in Excel, Word etc which takes them directly to the file system. In the ICPAC workshop users requested if the Save As could take them directly to the file system instead of having to click on browse again.

bayakarisa:
@lilyclements @rdstern 
@berylwaswa  Kindly check 

Fixed #5702 

I did Two small changes:

Added the TriggerBrowse() method to ucrFilePath.vb
Called it from dlgSaveAs_Load

**This will automatically open the file browser when the Save As dialog loads.**

https://github.com/user-attachments/assets/68ff65cf-5a08-4ef0-978b-5bdbb62f9b86

### DIFF
--- a/instat/ucrFilePath.vb
+++ b/instat/ucrFilePath.vb
@@ -235,9 +235,9 @@ Public Class ucrFilePath
     End Sub
 
     'Triggers the browse button click programmatically
-    Public Sub TriggerBrowse()
-        btnBrowse.PerformClick()
-    End Sub
+    'Public Sub TriggerBrowse()
+    '    btnBrowse.PerformClick()
+    'End Sub
 
     Private Sub ucrFilePath_Load(sender As Object, e As EventArgs) Handles Me.Load
         If bFirstLoad Then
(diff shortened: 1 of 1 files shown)


bayakarisa:
@Patowhiz I tested without BeginInvoke and it works perfectly. I've simplified the code

instat/ucrFilePath.vb
```diff
@@ -234,6 +234,11 @@ Public Class ucrFilePath
         FilePath = ""
     End Sub
 
+    'Triggers the browse button click programmatically
+    Public Sub TriggerBrowse()
+        btnBrowse.PerformClick()
```
Patowhiz: @bayakarisa Instead of calling `btnBrowse.PerformClick()`, you could move the logic inside the `btnBrowse` click event handler into the `TriggerBrowse()` subroutine.

You can then call `TriggerBrowse()` both from the click event and anywhere else it’s needed. This makes the intent clearer and keeps the behavior semantically explicit.

PR #10166, opened 2026-01-12
Improvements to 2 and 3 way Summarise dialog

### LINKED ISSUE #10127: Improvements and correction in the 2 and 3 way Summarise dialog
I hope that @Vitalis95 may be prepared, and could do these changes quickly.

Here is the data, with then dialog and the result:

Image

a) In the output table the header says count_all Joblevel.  We don't need the name count_all
b) In ordinary R functions the NA in these frequency tables are zeros.  Could they be given here as 0 instead of NA?
c) The margin checkbox, both here, and with 2 variables, doesn't do anything.  It doesn't give the margins!

@rachelkg and @berylwaswa there is a video on tables called [Creating Tables in R](https://www.google.com/search?sa=X&sca_esv=d7139353ec3d0296&udm=7&sxsrf=AE3TifP4hOyOHlOdr1nvbKTEoeISp6ENXA:1765128553605&q=Create+data+table+in+R&ved=2ahUKEwi4sK6jgKyRAxXOYEEAHY_UCCQQ8ccDKAJ6BAg_EAQ&biw=1280&bih=631&dpr=1.5#fpstate=ive&ip=1&vld=cid:8e22fd47,vid:Ja_CM253oDQ,st:0).  

We can do the whole of that video nicely with this dialog.  The margins is a luxury, because he doesn;t mention margins.  I suggest this will be great in our help and also as a simple video.  The data are called EmployeeDemographics.csv and are available from [his github site.](https://github.com/davidcaughlin/R-Tutorial-Data-Files)

Vitalis95:
Fixes #10127 
@rdstern @lilyclements , this is ready for review

### DIFF
--- a/instat/dlgDescribeTwoVariable.vb
+++ b/instat/dlgDescribeTwoVariable.vb
@@ -600,9 +600,11 @@ Public Class dlgDescribeTwoVariable
     (rdoThreeVariable.Checked AndAlso (IsFactorByFactorByNumeric() OrElse IsFactorByNumericByFactor() OrElse IsFactorByFactorByFactor()))
 
         ucrInputMarginName.Visible =
-    ucrChkDisplayMargins.Visible AndAlso
-    ucrChkDisplayMargins.Checked
-
+    ucrChkDisplayMargins.Checked AndAlso
+    ((rdoTwoVariable.Checked AndAlso (IsFactorByFactor() OrElse IsFactorByNumeric())) OrElse
+     (rdoThreeVariable.Checked AndAlso (IsFactorByFactorByNumeric() OrElse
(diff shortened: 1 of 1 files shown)


rdstern:
@Vitalis95 that seems great, many thanks.  I am not sure if you will change anything, following the gemini comments above.
@rachelkg this is the dialog you will be using for the practice document and video on tables in R-Instat.  Could you try the pull request with the data set proposed for that.
@ARMSTRONGOPONDO I read, from your recent work, that you aim to concentrate more on the VB side.  Could you please check and report on the code used here.  Then @lilyclements could decide whether more checks are needed, or it could then be merged.

Issue #10093, opened 2025-11-17
New Check Summary dialog to construct


rdstern:
This is a new dialog called `Climatic > Describe > Check Summary`, and I suggest just before the SPI/SPEI, dialog.
@jkmusyoka and maybe also @lilyclements to check, before work starts.

At the top there are four buttons with maybe `Recent, Trend, Outliers, Table` as the headings.

The `Recent` button is new, and maybe I explain it and the others, each in a sub-issue.

It is to produce a 3-level factor.  I need to check with @lilyclements and @jkmusyoka first.

The `Trend` button can start with the main controls from the climatic > Describe > Trend Graph dialog, but with the single/multiple y-variable control used on the main graph dialogs.

The `Outliers` button provides boxplots - without the buttons from that dialog for jitterplots, etc.
The `Table` button provides a table of counts of the different variables, probably by the Recent factor.

lilyclements:
The Recent button is awaiting responses on the PR
The Trend and Outlier button look ready to go!

@jkmusyoka do you have any additional thoughts on @rdstern's suggestions for this new dialog?

PR #10222, opened 2026-02-09
Added Check summary dialog

ARMSTRONGOPONDO:
Partial work Issue no #10093 (Recent and Trend button)

### Before asking for review
Please confirm that you have:

- [ ] Runs without errors  


ARMSTRONGOPONDO:
@rdstern @lilyclements and @jkmusyoka 
I have finished work on the check summary new dialog the only bits remaining are the plot options in the outlier tab and the bug on the circular plot instead of the box-plot and disabling the table options

jkmusyoka:
@rdstern and @jkmusyoka have tested the completed parts of this dialog using this rainfall summary data from Zambia
[eastern_summaries.zip](https://github.com/user-attachments/files/26937830/eastern_summaries.zip)

This is a good start to the dialog but we suggest the following improvements:

- [x] the recent tab works fine but can we simplify the default name of the resulting column to "recent"
- [x] In the outlier tab, changing the single variable to multiple variable the box overlays the x-variable text so this needs to be corrected
- [x] Still in the outlier tab, if the dataframe includes a "recent" column can this be the default x-variable
- [x] Still in the outlier tab, if the dataframe includes a station name column, then can you make this the default in the Facet_by box and the default type should be facet_wrap
- [x]  Still in the outlier tab, the dialog crashes largely because the facet_wrap code does not include a ~. We changed this in the code and the dialog works.

rdstern:
I am adding to your list.

image

a) If there is a `recent` variable (and maybe only for single variable?) you get the boxplots in colour, see above, if you also put the recent in the fill-by.  So, if there is a recent variable, could you put it in both.

b) It is worth having the variable width.  So could the default, please be to have that ticked.

c) Could you add a checkbox, default unchecked, with label `Year for Outliers`.  This is tricky, except that @Emmanuel-Afrifa has already provided it for the `climatic > check data > boxplot`.

In the climatic boxplot you can choose which variable.  But you only allow the year, so you don't need the receiver.  If the data are defined as climatic, then a year column will have been defined.  If no year has been defined, then disable that checkbox. 

Here is the dialog for that: (so you can check)

image


And here is the resulting graph - nice eh?

image

Next will be improvements in the trend graph.  And @jkmusyoka will provide the R code for the last option, namely the table.

rdstern:
I think I now have a possible design of the tables option for this dialog.  If you agree, then it is sort of a special option from our 2/3 variable summarise, which is also a dialog in Climatic Describe.  I suggest this is done next, because the trend graph (though @jkmusyoka doesn't like that name, can wait until the Climatic > Describe > Trend graph is sorted.  

I suggest the table is for multiple summaries for an individual station.  And, because we use it for checking a station and the proposing that it be accredited we deliberately do not try to get efficient, i.e. do multiple stations at once.

So, the dialog is simple.

a) Usual data selector
b) Multiple receiver to take many summary columns.  
c) Two controls to choose a station.  

c is potentially complicated except we already have these controls in the Climatic > Examine Data > Daily Data Editing dialog.  It is the single receiver with label Station, and the one below, which is called Select Station.  And they can be exactly as there, with the same labels and also hopefully with the same code behind.

image


Then it should produce the table of counts non missing, including the margins!  

Note the table below is not right, because it is for multiple stations and one summary.  The table we want is for multiple summaries and a single station.  I hope @jkmusyoka can help with the R code if needed.

Here is the summary dialog:

image
 And here is the resulting table:

image

That's nice, and we may discuss this option in the help.  But we only want the Count-non-missing and we always want the margins.  So we don't need those controls.

And we want multiple summaries, with the heading being the chosen Station.

ARMSTRONGOPONDO:
@jkmusyoka and @rdstern  thanks for the detailed feedback ill implement the corrections and the suggestions and next look into the table options part

ARMSTRONGOPONDO:
> @rdstern and @jkmusyoka have tested the completed parts of this dialog using this rainfall summary data from Zambia [eastern_summaries.zip](https://github.com/user-attachments/files/26937830/eastern_summaries.zip)
> 
> This is a good start to the dialog but we suggest the following improvements:
> 
>     * [ ]  the recent tab works fine but can we simplify the default name of the resulting column to "recent"
> 
>     * [ ]  In the outlier tab, changing the single variable to multiple variable the box overlays the x-variable text so this needs to be corrected
> 
>     * [ ]  Still in the outlier tab, if the dataframe includes a "recent" column can this be the default x-variable
> 
>     * [ ]  Still in the outlier tab, if the dataframe includes a station name column, then can you make this the default in the Facet_by box and the default type should be facet_wrap
> 
>     * [ ]   Still in the outlier tab, the dialog crashes largely because the facet_wrap code does not include a ~. We changed this in the code and the dialog works.

@jkmusyoka I have succefully implemented this suggestions and corrections

ARMSTRONGOPONDO:
> I am adding to your list.
> image
> 
> a) If there is a `recent` variable (and maybe only for single variable?) you get the boxplots in colour, see above, if you also put the recent in the fill-by. So, if there is a recent variable, could you put it in both.
> 
> b) It is worth having the variable width. So could the default, please be to have that ticked.
> 
> c) Could you add a checkbox, default unchecked, with label `Year for Outliers`. This is tricky, except that @Emmanuel-Afrifa has already provided it for the `climatic > check data > boxplot`.
> 
> In the climatic boxplot you can choose which variable. But you only allow the year, so you don't need the receiver. If the data are defined as climatic, then a year column will have been defined. If no year has been defined, then disable that checkbox.
> 
> Here is the dialog for that: (so you can check)
> image
> 
> And here is the resulting graph - nice eh?
> image
> 
> Next will be improvements in the trend graph. And @jkmusyoka will provide the R code for the last option, namely the table.

@rdstern I have implemented the fill by and the variable width checked by default
image

The only remaining change is requirement c which Ill be doing so by the end of this week

rdstern:
@jkmusyoka is checking this dialog and we would like it merged for the next release.  So I'll let him react to your suggestions and addd one more:  

a) You suggest adding the label with year, as Emmanuel has done for the climatic boxplot.  It is a very good idea to add the label with a variable.  He has now done this for the main boxplot at good idea to include it here.  But don't restrict to the year.  That's useful for the ordinary climatic boxplot, which is for the daily data.  Here the year is usually the x-axis so we wouldn't need it twice.  But allowing any column would be very useful.
b) I hope the summary tab is now available.  That's important and I haven't checked recently.
c) I wonder about having something clear for the extremes?  For example could there be 2 check-boxes.  One says First and Last and that other says Smallest and Largest.  If you tick first and last it gives a box from below the x-axis to the top around the first and another one around the last year? The other gives 2 horizontal boxes.  (Maybe you can suggest something better than a box?

ARMSTRONGOPONDO:
@rdstern 
I have added the label outliers but due to your last message I have not restricted it to the year variable only and I have named it as check outliers since their was an existing label as "label outliers"  can you please confirm if that is what was required or do I need to make further adjustments

ARMSTRONGOPONDO:
@rdstern 
Could you clarify your last request regarding b and c so I can make sure I implement it correctly?
Also, just to confirm—by the summary tab, did you mean the table tab?
It would also be helpful to complete all updates and refinements for the current three tabs and merge them first. I’d prefer not to start work on the last tab (the table section) until the existing three tabs are fully merged and all requested refinements are addressed.

jkmusyoka:
@ARMSTRONGOPONDO these are the changes we have agreed for the Check Summary dialog: 
**The Outlier tab**
The label outlier box does not seem to work as expected. When ticked it does not give the appropriate box where the user can specify what variable to use for labelling e.g. year. The check outliers seems to do that but it doesn't work either. Between the two, I prefer label outliers to check outliers and this is consistent with the other dialogs. 
The default jitter width is set to 0.2. which is different to the default R setting of 0.4. Is there any reason for this difference? The other problem is varying the width does not seem to reflect expected changes in the graph. It could be that the default value is hard-coded and so changing it in the dialog doesn't make a difference. Please investigate
Re-position the Store Graph box so it is consistent with other dialogs
Re-position the Facet By box so it is consistent with other dialogs

**The trend tab**
When mean lines are added, the mean value displayed is the same across the stations. @rdstern suggested copying the corrections made in the PICSA graphs to resolve this.
The add points box - I don't think we need to jitter the points here like in the outlier tab because the points are part of the graph automatically and they are joined by a line. So we can remove the jitter-related fields
Re-position the Store Graph box so it is consistent with other dialogs
Label first and last - this does not seem to work at the moment. I believe it is work in progress but not very urgent for merging next week.

@rdstern you can add what I may have missed

I have send the sample dataset for testing by email

ARMSTRONGOPONDO:
> @ARMSTRONGOPONDO these are the changes we have agreed for the Check Summary dialog: **The Outlier tab** The label outlier box does not seem to work as expected. When ticked it does not give the appropriate box where the user can specify what variable to use for labelling e.g. year. The check outliers seems to do that but it doesn't work either. Between the two, I prefer label outliers to check outliers and this is consistent with the other dialogs. The default jitter width is set to 0.2. which is different to the default R setting of 0.4. Is there any reason for this difference? The other problem is varying the width does not seem to reflect expected changes in the graph. It could be that the default value is hard-coded and so changing it in the dialog doesn't make a difference. Please investigate Re-position the Store Graph box so it is consistent with other dialogs Re-position the Facet By box so it is consistent with other dialogs
> 
> **The trend tab** When mean lines are added, the mean value displayed is the same across the stations. @rdstern suggested copying the corrections made in the PICSA graphs to resolve this. The add points box - I don't think we need to jitter the points here like in the outlier tab because the points are part of the graph automatically and they are joined by a line. So we can remove the jitter-related fields Re-position the Store Graph box so it is consistent with other dialogs Label first and last - this does not seem to work at the moment. I believe it is work in progress but not very urgent for merging next week.
> 
> @rdstern you can add what I may have missed
> 
> I have send the sample dataset for testing by email

@jkmusyoka I have implemented most of this corrections feel free to test it out again

jkmusyoka:
@ARMSTRONGOPONDO this is good and I think ready for merging into a new version for next week. 

So far the outlier and recent tabs work fine and that is all we need for next week.

There are changes I would like to see in the trend graph but those can wait and I will list them at a later time. 

So for now, can we merge this?

jkmusyoka (dismissed):
Approved!

lilyclements:
@Vitalis95 can you review the VB code?

instat/dlgCheckSummary.vb
```diff
@@ -0,0 +1,2354 @@
...
+    Private clsPasteUpperTercileY As New RFunction
+    Private clsFormatUpperTercileY As New RFunction
+
+    Private clsGeomHlineMedian As New RFunction
+    Private clsGeomHlineLowerTercile As New RFunction
+    Private clsGeomHlineUpperTercile As New RFunction
+
+    Private clsGeomHlineAesMedian As New RFunction
+    Private clsGeomHlineAesLowerTercile As New RFunction
+    Private clsGeomHlineAesUpperTercile As New RFunction
```
Vitalis95: remove these

instat/dlgCheckSummary.vb
```diff
@@ -0,0 +1,2354 @@
...
+        ucrNudJitter.SetLinkedDisplayControl(lblJitter)
+
+        ucrNudTransparency.SetParameter(New RParameter("alpha", 2))
+        ucrNudTransparency.SetMinMax(0, 1)
+        ucrNudTransparency.DecimalPlaces = 2
+        ucrNudTransparency.Increment = 0.01
+        ucrNudTransparency.SetLinkedDisplayControl(lblTransparency)
+        ucrNudTransparency.SetRDefault(1)
+
+        rdoBrokenStick.Enabled = False
```
Vitalis95, line 377: not implemented yet? Then put a comment

instat/dlgCheckSummary.vb
```diff
@@ -0,0 +1,2354 @@
...
+
+        ucrOutlierInputStation.SetItems({strFacetWrap, strFacetRow, strFacetCol, strFacetRowAll, strFacetColAll, strNone})
+        ucrOutlierInputStation.SetDropDownStyleAsNonEditable()
+
+        strOutlierGeomParameterNames = {strOutlierFirstParameterName}
+
+        UpdateVisiblePanels()
+    End Sub
+
+    Private clsWithFunction As New RFunction
```
Vitalis95: move this up

instat/dlgCheckSummary.vb
```diff
@@ -0,0 +1,2354 @@
...
+
+        rdoMeanLine.Checked = False
+        rdoFittedLine.Checked = False
+        rdoBrokenStick.Checked = False
+        ucrChkPoints.Checked = True
+        ucrChkFirstAndLast.Checked = False
+        ucrChkWithSE.Checked = False
+        ucrChkLabel.Checked = False
+        ucrChkLabeloutliers.Checked = False
+        ucrChkAddPoints.Checked = False
```
Vitalis95: Could we avoid hardcoding the default checked states here? Please follow the pattern used in dialogs such as `dlgCluster`, where defaults are managed through a dummy function and `SetRDefault/AddParameterValuesCondition`, then applied through `SetRCode`, rather than explicitly setting `.Checked = True/False` for each control.

instat/dlgCheckSummary.vb
```diff
@@ -0,0 +1,2354 @@
...
+        ucrChkLegend.SetRCode(clsOutlierThemeFunction, bReset, bCloneIfNeeded:=True)
+        ucrInputLegendPosition.SetRCode(clsOutlierThemeFunction, bReset, bCloneIfNeeded:=True)
+        ucrOutlier1stFactorReceiver.SetRCode(clsOutlierFacetVariablesOperator, bReset)
+
+        If bReset Then
+            AddRemoveFacets()
+        End If
+    End Sub
+
+
```
Vitalis95: please remove unnecessary blank lines and extra spaces elsewhere in the dialog for consistency.

Vitalis95 (commented):
@ARMSTRONGOPONDO  good work
I've  left a few comments. Gemini has also raised some concerns that I think are worth addressing

ARMSTRONGOPONDO:
@Vitalis95 I implemented the recommended changes

instat/dlgCheckSummary.vb
Vitalis95: @ARMSTRONGOPONDO 
I noticed this comment was marked as resolved, but the hardcoded .`Checked = True/False` lines are still present in the code.

Did you get a chance to check how this is handled in `dlgCluster`, or other dialogs? If you have a few minutes, could we have a quick chat?

ARMSTRONGOPONDO:
@jkmusyoka can you please test again and approve .

jkmusyoka (approved):
Approved!

Vitalis95 (approved):

PR #10230, opened 2026-02-11
Added Column details to the Column Metadata display in reogrid

### LINKED ISSUE #10220: Add Column details to the Column Metadata display in reogrid
The bottom of the data view gives a useful summary of the number of rows and columns in the data frame, see below:

Image

It is now easy to swap and view the metadata:

Image

Please could the Column information also be shown here?  It could just be a copy of the same column information.

Note, though That it is possible to have a select.  This is also shown nicely in the data view and would be equally nice in the metadata view.

Image

And then add here in the Column metadata view:

Image

@N-thony if you are still volunteering, this is an area you worked on before?  Otherwise it seems an obvious task for @derekagorhom 

@lilyclements I wonder if there is a function already, or one could be included to help, when the append dialog gives an error, because columns have the same name, (in 2 data frames) but are of different types.  The error message is not helpful.  In this example, ELEVATION imported as numeric in one data frame, and character in another.  I wonder if this might be a simple function, and just provided as code - and in our script library.  All it does is to take 2 data frames and record, in the output window, if they have the same name but the type is different.  So here it might just record:

a) No different variable types - or -
b) Variable          Name1            Name2
     ELEVATION     Numeric          Character
     etc?

Ag-Derek:
Fixes #10220 
I have added to the Column Metadata display the column details @rdstern requested in #10220.
@rdstern , @lilyclements this PR is ready for review

- [x] Runs without errors  

### DIFF
--- a/instat/Interface/IColumnMetaDataGrid.vb
+++ b/instat/Interface/IColumnMetaDataGrid.vb
@@ -29,4 +29,6 @@ Public Interface IColumnMetaDataGrid
 
     Sub SetNonEditableColumns(lstNonEditableColumns As List(Of String))
 
+    Event WorksheetChanged()
+
 End Interface
--- a/instat/UserControls/DataGrid/Linux/ucrColumnMetadataLinuxGrid.vb
+++ b/instat/UserControls/DataGrid/Linux/ucrColumnMetadataLinuxGrid.vb
@@ -23,6 +23,15 @@ Public Class ucrColumnMetadataLinuxGrid
 
     Public Event EditValue(iRow As Integer, strColumnName As String, strPreviousValue As String, newValue As Object) Implements IColumnMetaDataGrid.EditValue
 
+    Public Shadows Event WorksheetChanged() Implements IColumnMetaDataGrid.WorksheetChanged
+
+    Private Sub tcTabs_SelectedIndexChanged(sender As Object, e As EventArgs) Handles tcTabs.SelectedIndexChanged
+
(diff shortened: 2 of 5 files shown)


rdstern:
@Patowhiz can you answer this, or should we ask Chris instead?

ARMSTRONGOPONDO:
Hi @Ag-Derek,
I've reviewed the changes and tested them locally. 
However, I did confirm the issue @rdstern reported: the metadata view currently doesn't update when you switch between different worksheets (dataframes).

The ucrColumnMetadata control isn't currently listening for when the active worksheet changes. In the Data View (ucrDataView), there is a specific event listener (WorksheetChanged) that triggers a refresh when a tab is clicked, but the Metadata view is missing this mechanism.
The Fix:
I managed to get this working locally by mirroring the approach used in the Data View. You'll need to:
1.  Update the Interface: Add a WorksheetChanged event to IColumnMetaDataGrid so the main control knows it exists.
2.  Implement in Grids: In both the ReoGrid and LinuxGrid user controls, implement this new interface event. For the ReoGrid specifically, you need to catch its internal CurrentWorksheetChanged event and bubble it up using your new event.
3.  Connect in the Main Control: Finally, in ucrColumnMetadata.vb, add a handler in SetupInitialLayoutAndGrid that listens for this WorksheetChanged event. When it fires, it should simply call AfterStateChange() to update the labels and buttons for the new active sheet.
With those structural updates, the row/column counts and navigation buttons update immediately when switching tabs.

image


image

PR #10244, opened 2026-02-19
Added `Insert Column/Rows` dialog into the `Climatic > Tidy` menu and dynamic help for dialog

### LINKED ISSUE #10219: Small edit in databook to insert a character, rather than a numeric variable
@lilyclements the dialog Prepare > Dataframe > Insert Columns/Rows on the Column code, runs the following function, at least when it inserts NA:

```
# Dialog: Insert Columns/Rows
data_book$add_columns_to_data(data_name="survey", before=FALSE, col_data=NA_real_, col_name="X", use_col_name_as_prefix=TRUE)

```
Please could you change, so it inserts col_data=NA_character_ instead?

It sounds trivial, but our recent use has been to insert a Comment column.  When this happens it would be better if the new column were immediately defined as a character column.

In addition, most users have experience in changing character (C) variables to Numeric.  They have never previously had to go from Numeric to Character.

I think the change is only needed if NA is chosen.  The existing code correctly produces a numeric column if 0 is the default, and character if a text is given.

@berylwaswa could you also allocate that this dialog also be added to the Climatic > Tidy Data menu.  It should be just before Duplicate Rows in the menu.

@rachelkg this change should also be reflected in the Help file.  It can be the same help, but ideally the figure  with the 2 columns inserted, the X is now X(C) rather than X(L).  

Rachel: In the default Value section in the help you could mention "The resulting variables are of Character type, unless the default value is numeric."

Rachel: Leave still the bit in red at the bottom for now.  And add the help context number please.

Rockjunior:
- Added InsertColumnRowsToolStripMenuItem to the UI and connected it to open dlgInsertColumn.
- Introduced InsertColumnMode enum and a public variable in dlgInsertColumn to track context (Prepare or Climatic).
- Implemented SetHelpOptions to set help topic ID based on mode, allowing dynamic help content.

### Before asking for review
Please confirm that you have:

- [x] Runs without errors  

Partially fixes #10219

### DIFF
--- a/instat/dlgDeleteRowsOrColums.vb
+++ b/instat/dlgDeleteRowsOrColums.vb
@@ -14,16 +14,8 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-Imports instat.dlgInsertColumn
 Imports instat.Translations
 Public Class dlgDeleteRowsOrColums
-    Public enumDeleteRowsOrColumnsMode As String = DeleteRowsOrColumnsMode.Prepare
-
-    Public Enum DeleteRowsOrColumnsMode
-        Prepare
-        Climatic
-    End Enum
-
     Private bFirstLoad As Boolean = True
     Private bReset As Boolean = True
     Private clsOperatorRowNames As New ROperator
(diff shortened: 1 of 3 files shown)


Rockjunior:
@lilyclements @berylwaswa @rachelkg 
I have made changes. Kindly review and let me know if any further revisions are required.

PR #10273, opened 2026-03-19
Improvement to the Right Click Rename Column Dialog 

### LINKED ISSUE #10251: Improve the right-click to Rename Column dialog
@derekagorhom or @Ag-Derek this looks like one for you please.

I supply the data used recently for this dialog:

[cameroon rainfall stations.zip](https://github.com/user-attachments/files/25699598/cameroon.rainfall.stations.zip)

Note that this dialog us usually accessed via the column right-click as shown here:

Image

When Rename is chosen, here is the resulting dialog:

Image

I think it is the only right-click dialog where there are multiple radio buttons at the top.

Usually we use the default, as shown here, and that's fine.  But recently I have been switching the top buttons and then I am not happy with the next time I right-click, which may be in a different data frame.

I suggest the answer is that when using the right-click > Rename, please always go to this first top-radio button (the simplest use of the dialog).  And with the current data frame that corresponds to the right-click.

Note that you could alternatively return to this dialog, as the last one - using the toolbar.  Or it is on `Prepare > Data Frame > Rename Columns`.  I think they are fine, so don't change anything there.  They correctly remember the past, while the right-click to rename is now always starting with the current data frame, current column and simplest setting of the toolbar.

@berylwaswa please can you check when the changes are made for this issue.  There is a follow-up, so Derek, please do a "partially" on the pull request.  Once that is done I suggest the second option `Multiple` is very nice, but rather flaky.  It has a funny check of a name being empty that comes at the wrong time, and occasionally can crash.  I suggest it may be time to include use of gemini to examine the code.  But only after this first change is made.

Ag-Derek:
Fixes #10251

This implements the improvement to the column rename dialog which is located on the right click menus where we make sure the first options is always selected when the user uses the dialog from the Right click menus and this does not affect it when the users uses it from it other options. 
I used the survey dataset for the testing 
@rdstern @berylwaswa this is ready for review. 
Thanks 

- [x] Runs without errors  


rdstern (dismissed):
@Ag-Derek many thanks, that looks fine now.  I am approving.
Sorry @Patowhiz but another one that arose from our climatic work for you to check and hopefully merge.
It would be good for this to get into the forthcoming release.

instat/dlgName.vb
```diff
@@ -827,4 +829,16 @@ Public Class dlgName
             cmdAddkeyboard.Visible = ucrChkIncludeRegularExpressions.Checked
         End If
     End Sub
+
+    Private Sub ReopenDialog()
+        SetSingleOptionStatus(bDefaultToSingle)
+    End Sub
```
Patowhiz: Given that `ReopenDialog` is private, has a single line and is called once , you could simply call `SetSingleOptionStatus` inside `dlgName_Load` event and include a comment that explains why it needs to be called every time the dialog is reopened.

instat/ucrDataView.vb
```diff
@@ -272,6 +272,7 @@ Public Class ucrDataView
 
     Private Sub mnuColumnRename_Click(sender As Object, e As EventArgs) Handles mnuColumnRename.Click
         dlgName.SetCurrentColumn(GetFirstSelectedColumnName(), _grid.CurrentWorksheet.Name)
+        dlgName.bDefaultToSingle = True
```
Patowhiz, line 275: Instead of calling `dlgName.bDefaultToSingle = True` why not just make `SetSingleOptionStatus`  to be public, then call it here and in line 695 of `ucrDataView` and re-implement it as below?

```
Private Sub SetSingleOptionStatus()
         rdoSingle.Checked = True
End Sub
```

This means you won't need the `ReopenDialog `  method and `bDefaultToSingle ` property.
If you agree with this suggestion, you may want to the new function a more suggestive name.

Patowhiz (changes requested):
@Ag-Derek thanks for this. Just a few questions and comments below, then we can retest and merge.

instat/ucrDataView.vb
Ag-Derek: I think keeping this within the dlg Name file is more appropriate compared with making it public, if it were to be used by multiple dialogs then that would be the way to do it but since it is only one dialog i suggest we let it be private.

instat/dlgName.vb
Ag-Derek: I agree and i have made the change

instat/ucrDataView.vb
Patowhiz: @Ag-Derek what my comment above suggests is that you entirely remove the public property `bDefaultToSingle` and have the private function `SetSingleOptionStatus` to be public after re-implementing it as explained in the same comment. 

As I understand it, the only reason you added property `bDefaultToSingle` is to have it set to true by the right click menu toolstrips in the data viewer, so that they can set the single radio button to be in a checked status. Instead of having a method and property for this, you can just have a public method that sets the radio button directly before opening, given the default R code is to have the single radio button checked. 

That said, I understand the addition you have made of using properties and methods for this kind of set up follows the `dlgName.SetCurrentColumn` implementation pattern, which in my own opinion is not ideal, but is safer given how the whole dialog is implemented. So I'm approving.

Patowhiz (approved):

Patowhiz:
@rdstern I've approved this. Kindly test then we can merge.

rdstern (approved):
Thanks @Ag-Derek and @Patowhiz I am approving

Issue #10183, opened 2026-01-20
@rdstern and @lilyclements


rachelkg:
@rdstern and @lilyclements 

I have been using this dialog and notice that it treats the percentages differently from the Two-Way Frequencies dialog. When I use the Two-Way Frequencies dialog, I check the Row Percentages box I get each row totally 100%: 
<img width="433" height="471" alt="2026-01-19_12-06-51" src="https://github.com/user-attachments/assets/2440a681-720e-4fcd-87b3-109e0804fa45" />
<img width="253" height="317" alt="2026-01-19_12-06-10" src="https://github.com/user-attachments/assets/2e4726cb-c609-4d21-b5bc-0bda8ca84950" />

On the other hand, when I am in the Summarise two/three Variables dialog and set up what I think is the same table with the Row% radio button checked, I get all of the columns adding up to 100. 
<img width="484" height="478" alt="2026-01-19_12-07-44" src="https://github.com/user-attachments/assets/e76b3c7a-9d71-4c5b-8a38-a19fd9b1cc2e" />
<img width="297" height="340" alt="2026-01-19_12-10-06" src="https://github.com/user-attachments/assets/682f1a0a-6c97-411f-8647-f6c12e13bfa5" />

Equally, if I select the Col% radio button, I get rows of 100
<img width="254" height="339" alt="2026-01-19_12-11-10" src="https://github.com/user-attachments/assets/36cf5fad-92f1-4a93-86a4-a048b4cebb03" />

The receiver below the Percentages box fills automatically, and while I can select it I dont see to be able to change the contents of it. I was trying to as I wondered if I put the other variable in here I would get the outcome I expected. I found it strange that this was selectable but not changeable. 

Is this a mistake or am I misunderstanding? If I am filling the dialog wrong or misunderstanding this, feel free to ignore this. I just wanted to highlight it, in case it is an error?

_Originally posted by @rachelkg in https://github.com/IDEMSInternational/R-Instat/issues/10166#issuecomment-3769975159_

rdstern:
There is a lot to do here, but I hope it is all straightforward.  I suggest we might merge progressively (maybe 3 times) as we make the different improvements.  However, the dialog is worth it.  It is really important, both for work and for our teaching too.  So here is my list, with all but item 1 taken from the topics above.

[...]

PR #10288, opened 2026-04-13
More improvements to Two and Three Summaries dialog

Vitalis95:
Fixes partially #10183 
@rdstern , @lilyclements , @berylwaswa  The PR fixes the order of variable in 3-way tables


rdstern:
@Vitalis95 that's promising.  It worksg going down.  But it is odd going across?

image

Interestingly I changed the order of the factor levels in that fertgrp factor, to go from 0, and the table didn't change?

lilyclements:
@rdstern this is coming from `pivot_wider`, not in our stuff! How nice, I think. 

The solution is to use `names_expand = TRUE` as an additional argument in `pivot_wider`:

E.g., 

```
survey %>% pivot_wider(names_from = fertgrp, values_from = yield, names_expand = TRUE)
```

Vitalis95:
@rdstern , have a look at it and also update your databook

rdstern:
@Vitalis95 and @lilyclements That solves the order nicely.  But it now includes the NA level, even though there are no NAs in the data?

Of course that makes me realise that when there are NAs in any factor, then it would be nice to have an option to have that level, or exclude it.  Maybe that could be a general option?  Is that a @lilyclements question?

Interesting that the NA column is currently only for that factor?

Oh and my databook didn't need updating.

lilyclements:
@rdstern so `names_expand = TRUE` fixes the order, but also adds in all other levels. The issue here is we had a "bonus level" that is "NA".
I've amended the databook code to drop unused factor levels to bypass this. So, if you update your databook, it should hopefully now work

rdstern (approved):
@Vitalis95 and @lilyclements that looks great now.  I am approving, so we can merge with this change first.  Lily can you approve and merge?
@Vitalis95 I hope you are happy to continue with the Outer Margins option and also the addition of the percentages option for the 3-way?  If not, then to supervise one of the others?

lilyclements (approved):

PR #10317, opened 2026-05-08
Added the pivottable package as an archive into r-instat

### LINKED ISSUE #10314: The pivot table dialog is not in the latest version of R-Instat
@ChrisMarsh82 and @lilyclements and @berylwaswa I report that Version 0.8.14 does not include the rpivotTable package.

Its omission raises a number of points.  It is not there presumably because the package was removed from CRAN in March 2026.

First I checked with James.  He informs me he is also using Version 0.8.14 and it is there in his version.  Are there 2 Versions both called 0.8.14?

So, Beryl, it is interesting that this omission was not picked up in the testing?  (But maybe you were testing the other 0.8.14.  I am very confused.
And Chris, with so many packages in R-Instat, it is likely that occasionally one will be removed.  Can this be checked in the installation procedure?
And Lily, I wonder what we should do about it?  I really like the facilities it offers.  It hasn't been updated for about 8 years, so it is clear that the developers are no longer interested.  I wonder if there is an alternative?  In the short term it is not the only non-cran package we use, so can we add it?  And can we instruct users into how they can do it?

Ag-Derek:
Fixes #10314 
I have added the PivotTable package as an archive into R-instat

- [x] Runs without errors  


Ag-Derek:
@lilyclements i have the pivot table as an archive here unfortunately we have had lights since yesterday but you can review this and when you are ok you can merge else you can leave a comment and i will address it when i get light back

lilyclements (approved):
@Ag-Derek looks good to me! @berylwaswa can you review and approve?

berylwaswa (approved):
@Ag-Derek looks good.

PR #10341, opened 2026-05-20
Refactored dlgCalculator parameter handling and adjacent column support

### LINKED ISSUE #10331: Small problems with the Calculator
@berylwaswa here is the calculator with the maths keyboard:

Image

Two problems:

a) Tiny typing error in the keyboard is the typing error in the siginf key!  It should be `signif`.
b) The second problem is more general for the calculator.  The Position facility doesn't work.  In the figure I have asked for the new variable to be after the away score column.  But it doesn't do that.  The new column is always at the end.
I checked with the Prepare: Column: Numeric > Transform and the Position control seems to work fine there.  So that could become a model, though perhaps the developer could use another dialog to check it is ok there, at the same time, i.e. it is only in the calculator that it has gone wrong.

Beryl, I hope this will be easy for someone to fix.  At the same time I report that this bug came to light as I was preparing a chapter for @lilyclements in our variant of the lsr textbook.  I think some developers are working on other chapters.  I have been pleasantly impressed that R-Instat seems pretty solid, but have been reporting various problems, like this, as I go along.  I wonder what R-Instat problems the others have been finding?  Maybe good to check?

Rockjunior:
### Before asking for review
Please confirm that you have:

- [x] Runs without errors  

Refactored dlgCalculator parameter handling logic by updating SetAssignTo to use ucrAvailableDataFrames.Text for strTempDataframe, modifying clsAddColumnsToData to dynamically retrieve the before parameter, and adding support for the adjacent_column parameter with quote stripping logic. Also added inline comments to improve clarity around the new parameter handling flow.

@berylwaswa @rdstern @lilyclements @Vitalis95 @ARMSTRONGOPONDO @bayakarisa @kisinga Could you review please.

Fixes #10331

### DIFF
--- a/instat/dlgCalculator.vb
+++ b/instat/dlgCalculator.vb
@@ -90,7 +90,7 @@ Public Class dlgCalculator
         ucrCalc.ucrSelectorForCalculations.bUseCurrentFilter = False
         ucrCalc.ucrTryCalculator.SetRSyntax(ucrBase.clsRsyntax)
         ucrBase.clsRsyntax.SetAssignTo(ucrCalc.ucrSaveResultInto.GetText(), strTempColumn:=ucrCalc.ucrSaveResultInto.GetText(),
-                                   strTempDataframe:=ucrCalc.ucrSelectorForCalculations.ucrAvailableDataFrames.cboAvailableDataFrames.Text,
+                                   strTempDataframe:=ucrCalc.ucrSelectorForCalculations.ucrAvailableDataFrames.Text,
                                            bAssignToIsPrefix:=ucrBase.clsRsyntax.clsBaseCommandString.bAssignToIsPrefix,
                                            bAssignToColumnWithoutNames:=ucrBase.clsRsyntax.clsBaseCommandString.bAssignToColumnWithoutNames,
                                            bInsertColumnBefore:=ucrBase.clsRsyntax.clsBaseCommandString.bInsertColumnBefore,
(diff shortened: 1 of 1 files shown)


lilyclements:
@Rockjunior is this ready for review?

lilyclements:
@Rockjunior I'm still having the following issue in this dialog: "b) The second problem is more general for the calculator. The Position facility doesn't work. In the figure I have asked for the new variable to be after the away score column. But it doesn't do that. The new column is always at the end."

rdstern:
@Rockjunior I have just started checking and notice:
a) In Integer > Fibonnaci you changed the Zseq package into DescTools.  Was this intentional?  I ask because DescTools also has a Fibonnaci function, but it works a bit differently!  Maybe change back?
b) The next key, Lucas, the data frame (survey for me)  is not in quotes.  

I have not checked many others yet, but wonder how you are making the changes?  Did you change individually?

This could be a long check if this start is any indication!

lilyclements:
@Rockjunior nice, but this also needs to be changed in `ucrCalculator` as well. Just a quick "CTRL+F" (or and replace) for "nrow" should be enough to find and fix this

rdstern:
@Rockjunior and @berylwaswa I hope the corrections in the calculator can be finished and merged this week, so before your leave, Beryl.  It is one of the few problems in R-Instat that showed no visible progress during the sprint.  Beryl, you spotted the main problem in the first place.  It is the one item that is holding up to production of a new release.  @Rockjunior if it is too difficult for you, then perhaps hand it over to someone who can complete it?

Rockjunior:
I cannot continue working on this branch as we had initially made changes to the calculator. If I proceed with the current branch, I would end up undoing those changes.

I’m planning to first restart the implementations done here using a new branch after pulling the merged content from the previous `nrow` changes, and then test whether the reported issue in **b** still occurs. If the issue persists, I will make a new PR with the fix.

In the meantime, as you’re all aware, we no longer have `nrow` problems in the calculator. We fixed this using:

`frmMain.clsRLink.strInstatDataObject & "$nrows"` and other implementations

I now need to investigate this further and find a way to address the issue without introducing another problem in the calculator.

For now, I’ll consider this PR no longer useful and will close it.

berylwaswa:
@Rockjunior would you like to hand this task over to someone else or you are okay getting it done before the end of the week please?

Rockjunior:
That’s the one I’m trying to fix at the moment. Additionally, with every change I make, I need to test at least 75% of the other keyboards in the calculator to make sure I’m not introducing any regressions from the previous changes I made.

However, if you’re willing to hand it over to someone else, I’m also okay with that.

rdstern:
@Rockjunior please continue - if you are making progress.  But note our concern that it is good you are now posting on this pull request.  Since May 20th and even during the sprint, there is silence from you in in this github thread.  We have to resolve issues, not just keep silent.

You earlier made excellent progress solving @berylwaswa bug, on the whole Wakefield keyboard.  She had also found the same problem on the Integer keyboard and I assumed in less than a day you should have been able to apply the same fix to the other keys.  But that was many weeks ago.  So please either continue or hand it on.  And if you do hand it on then please report on the way you made progress so far, to help the new person.  Or work together with them.  I think @lilyclements mentioned something that is needed also?

Given the problems you mentioned above I am also a bit at a loss to know why you didn't use the sprint to finish the corrections - where you had so much access to help from others if needed.  Please keep reporting on progress and send bits for testing regularly.  Hopefully this week, before @berylwaswa goes on leave.

PR #10200, opened 2026-02-02
Added Save Definitions  Option to Spell dialog

Rockjunior:
Added "SaveDefinitions" checkbox and save control to dlgSpells, allowing users to optionally save spell calculation definitions. Updated dialog layout and integrated new controls with linked control logic. Implemented supporting RFunction objects and logic to extract and save definitions based on selected data and grouping variables. Controls dynamically update visibility and parameters in response to user input.
Fixes partially #10041

PR #10344, opened 2026-05-20
Saved output to a different data name

### LINKED ISSUE #10342: Save output to a different data frame
Now that #10326 is in, we want to amend where we save the definitions to. @Emmanuel-Afrifa can you take this up? We want to fix this in:

- [x] Start of Rains
- [ ] End of Rains
- [ ] End of Season
- [ ] Climatic Summary
- [ ] Crops
- [ ] Spells

This is just adding the line

```r
        ucrSaveDefinition.SetDataNameAsRVariable("linked_data_name")
```

(I believe it is called "linked_data_name" in all of those places, but if it is not then let me know and I'll double check)

Emmanuel-Afrifa:
Fixes #10342 

@lilyclements I have changed the save name as specified in the issue. The dialogs affected are as follows
- End of Rains / End of Seasons
- Climatic Summaries
- PICSA Crops

In PICSA Crops, the new data name I used was not `linked_data_name` (as was done in the other dialogs). Instead, when we're saving proportions, we have `new_crop_prop`, and when we're saving crop definitions, we have `new_crop_def`

Lastly, I have not added the option for the spells dialog here because there's an open PR (#10200) on that, and I believe this change was the last bit needed to get it merged. I'll add it in there now.

- [X] Runs without errors  
- [X] OK disabled when dialog is incomplete or invalid  
- [X] OK enabled only when required inputs are valid
- [X] Reset returns dialog to its default/sensible state
- [X] Invalid inputs are handled cleanly (e.g. negative, too-large, empty, impossible combos)
- [X] Running twice with different settings behaves consistently (e.g., open → run → close → reopen → change options checked → run again)

### DIFF
--- a/instat/dlgClimaticSummary.vb
+++ b/instat/dlgClimaticSummary.vb
@@ -165,6 +165,7 @@ Public Class dlgClimaticSummary
 
         ucrSaveObject.SetSaveType(strRObjectType:=RObjectTypeLabel.StructureLabel, strRObjectFormat:=RObjectFormat.Text)
         ucrSaveObject.SetDataFrameSelector(ucrSelectorVariable.ucrAvailableDataFrames)
+        ucrSaveObject.SetDataNameAsRVariable("linked_data_name")
         ucrSaveObject.SetLabelText("Name:")
         ucrSaveObject.SetIsComboBox()
         ucrSaveObject.SetAssignToBooleans(bTempAssignToIsPrefix:=True)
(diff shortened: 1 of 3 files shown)


lilyclements (approved):
@Emmanuel-Afrifa works well! Nice one @Ag-Derek can you review

Ag-Derek (approved):

lilyclements (approved):

PR #10348, opened 2026-05-22
Added r package check for PRs

ChrisMarsh82:
Added new script to check all R packages are installed.
Updated action to add results to the PR


ChrisMarsh82:
> ## ❌ R Package Check
> **6 of 149 package(s) are not installed.**
> 
> Metric	Count
> Total packages checked	149
> Installed	143
> Missing	6
> ### Missing Packages
> * `clifro`
> * `maptools`
> * `ggmosaic`
> * `ggalt`
> * `waffle`
> * `CGPfunctions`
> 
> **Missing packages detected.** Ensure all packages in `instat/static/InstatObject/R/install_packages.R` install correctly on the target R version.
> 
> _This comment is automatically updated by the R package check workflow._

@lilyclements I have updated the check for install packages. It will now update the PR with the results (see above). Before I look into updating the 6 packages that failed to install can you just confirm if these are need. (Don't look into it too much i just wanted to make sure these are obvious packages that are no longer needed)

lilyclements:
@ChrisMarsh82 great. These six seem to be removed from CRAN. Of these six:

`ggmosaic` - needed
`ggalt` - needed
`maptools` - let me know - used in some dialogs at the moment (these are very specific climate dialogs which I don't think are used at the moment anyway, so if this one is being difficult then let me know)
`clifro` - needed (although not a huge deal)
`waffle` - from a CTRL+F check on VB, I couldn't see this one. 
`CGPfunctions` - from a CTRL+F check on VB, I couldn't see this one.

It could be `waffle` and `CGPfunctions` were needed back in t' day when we had all of the R code (databook, etc) in R-Instat. Or, could they be dependencies? 



`ggalt` looks like it'll come back onto CRAN from their repo
`ggmosaic` has had movement in the last month, so I hope will be back to CRAN soon
`maptools` I don't know. I'll keep an eye.
`clifro` hasn't had action lately, and says "This repository was archived by the owner on Feb 7, 2025. It is now read-only."
`waffle` has asked if anyone wants to take over maintenance of the package, which we can do if it is an important one, but I cannot see where we use it.
`CGPfunctions` looks like it's off CRAN for good

lilyclements:
@rdstern do we use `waffle` in R-Instat? I thought we did waffle charts, but I cannot see the use of this package at all.

ChrisMarsh82:
@lilyclements I have now updated the list of packages and put them into r_packages.json This defines where they get installed from and the order (priority) This makes it easier to keep the check file and install file in sync. It should be able to be pulled out into a table format.

lilyclements (approved):

Issue #10332, opened 2026-05-18
Improve the Stack dialog when stacking multiple sets.


rdstern:
@lilyclements these are cosmetic and (I assume) may just be in the instatExtras package?  I use the first dataset from the lsr book, which is filtered to just the finals.

Here is the dialog:

<img width="452" height="595" alt="Image" src="https://github.com/user-attachments/assets/ebcf99af-ed90-4c28-8a0d-48e09ea837d3" />

I have added some variables, but that is not relevant.  You will have 8 variables, not 11, to drop.

Here is the result.  It works fine!

<img width="332" height="488" alt="Image" src="https://github.com/user-attachments/assets/9222e115-55ef-4246-b0e5-3327a715e0ae" />

The minor changes I would like:

a) The rows are complicated.  We don't use rows.  Could they be just the default of 1 to 400?
b) The dialog says location will be a factor, but it isn't.  Could it be made into a factor?
c) The id is an extra variable, a sort of bonus.  Could it be the first, rather than the last variable?

Here is the current code:
```
# Dialog: Stack (Pivot Longer)

afl <- data_book$get_data_frame(data_name="afl")
afl_stacked <- stats::reshape(data=afl, varying=instatExtras::split_items_in_groups(items=c("home.team","away.team","home.score","away.score"), num=2), timevar="variable", drop=c("calc","winner","margin","year","round","weekday","day","month","is.final","venue","attendance"), direction="long", idvar="id")
data_book$import_data(data_tables=list(afl_stacked=afl_stacked))

rm(list=c("afl_stacked", "afl"))
```

lilyclements:
@rdstern I had a look into `pivot_longer` to see if we could use this instead of `stack` to achieve what we want to here. 

I want to share with you the two outputs

1. This is what we currently run:

```r
# Dialog: Stack (Pivot Longer)
afl <- data_book$get_data_frame(data_name="afl")

afl_stacked <- stats::reshape(data=afl,
                              varying=instatExtras::split_items_in_groups(items=c("home.team","away.team","home.score","away.score"), num=2),
                              timevar="location",
                              drop=c("year","round","weekday","day","month",
                                     "is.final","venue","attendance"), 
                              direction="long", 
                              idvar="id")

data_book$import_data(data_tables=list(afl_stacked=afl_stacked))
```


2. This is what we could instead run if we changed the function to `pivot_longer` (and then we remove our function of `split_items_in_groups`.

```r
afl_long <- afl %>%
[...]

PR #10353, opened 2026-05-26
Improvements to the Stack dialog when stacking multiple sets

Vitalis95:
Fixes partially #10332 
@rdstern @lilyclements @berylwaswa ,  implemented `pivot_longer` function instead of  `split_items_in_groups` function

### Before asking for review
Please confirm that you have:

- [x] Runs without errors  


lilyclements:
@rdstern can you test to check this now does as you wanted it to in your original issue?

rdstern (approved):
@lilyclements and @Vitalis95 that's interesting - and different!  

Image

It assumes double-barreled names - which there were.  So the dialog worked fine!

a) The row variable is now fine.  Good.
b) The id is different, and this one is much better!

I like this a lot.  Many thanks.
I am approving.  @lilyclements do you want to confirm and merge, or do we ask Steve to look too?

@rachelkg once merged the help will need to change.  These data are quite a nuce example!

rdstern:
Except, in the old version, I was easily able to do a multiple stack of the famous Anscombe data.  This is in lsrdata, and also elsewhere, including the datasets package.  And I am using it in my variants chapter 2

Here it is:

image

Now I want to stack into 2 columns, each of length 44.  The old version did it as follows:

image

I don't know how to do this with the new version?  By the way I thought the current new dialog should be able to do it?  I tried by adding a period to each name, as below:

image

Then I hoped it would know the first 4 columns are one set and then second 4 are the other?  So it would stack to 44?  But instead it did this?

image

Here is what I would have liked instead:

image

I attach the data file where I have added the period to each name.  So it should be a correction to the code, I hope?

[anscombedot.zip](https://github.com/user-attachments/files/28784964/anscombedot.zip)

But, once it works, could you also add None to the pull down, or (maybe simpler for the code) add 1,2,3,4,5 to the options.  Then here we would have 1, so just after 1 character is the end of the first half.  If we have None instead, then you have to work out how many characters for the first part of the name.

If I have understood correctly, then this will be really nice.  But then maybe we should also have a Try, or a Check Names control.  Like on the Define Climatic, where we check for duplicates.  Here we check for an acceptable list of names.  

I hope I have understood correctly, because (if so) then I really like this.  It is a great example where R (used tidyly - spelling?) encourages you to be careful and consistent with the name columns.  And that's very different to a spreadsheet.

rdstern (changes requested):
@lilyclements and @Vitalis95 if I have understood correctly then I really like the new changes.  I tried first on the example given in the issue and it worked perfectly.  Then I tried with another, which works with the old system, but not now.  I explain in detail.  I hope the second example can also be made to work, with the only change in the dialog being that maybe a None should be added.  

But, in addition, there is now a new limitation, (assuming I have understood correctly), on the set of names in the columns.  So I am suggesting we have an additional Check Names control, like the Check Duplicates in the Define Climatic.  If red is gives a message and also  directs you to the help.

rdstern:
@lilyclements and @Vitalis95 any chance of progress on this?  I use it in the lsr Variant chapter 2.

rdstern:
@Vitalis95 the conflicts now need to be resolved.  Then I am not sure if the needed small change is possible by you, or needs @lilyclements ?

Vitalis95:
I had a look at Roger's example, and I think the difference comes from how `pivot_longer()` interprets the column names.

With `pivot_longer()`, the grouping is inferred from the names via `names_to` and `names_sep`. Our current implementation assumes `names_to = c("variable", ".value")`, which works for examples like `home.team`, `away.team`, `home.score`, `away.score`, but Roger's Anscombe example requires `names_to = c(".value", "set")` instead.
I'm wondering whether the dialog should let the user choose the order of `names_to` instead of always using the same order. That would allow both use cases while keeping the `pivot_longer()` approach.

@lilyclements, what do you think?

rdstern:
@Vitalis95 and @lilyclements I don't quite follow Vitalis question, but I am interested if there is are simple rules for the repeat stacks, that mean that I could do it if I changed the column names.    Are you saying that if the names were x1.x, x2.x, x3.x, x4.x and y1.y, y2.y, y3.y, y4.y then they would stack ok?

If so, then I would be happy with that solution.

rdstern:
@Vitalis95 there are now conflicts.  When resolved I would like to try changing the names in anscombe to see if that works.

rdstern:
@jkmusyoka and @Vitalis95 here is my problem and ther almost solution  with the possible dialog:

image

James, Vitalis is changing this - for multiple sets - to use pivot_longer and I really like this.  

The existing dialog works fine when I changed the names, as you can see in the script above.  I would like to use the existging names.  Could we check first - via the script - for the existing names?

I think we just need to change `names_sep =` in the script to `names_pattern =`.  I would like (in `names_pattern` to just use the first character of the names, so x for the first 4 variables and y for the next 4.  But what do I put in the script - I am no good at regex?

My suggestion in the dialog is that there are 2 radio buttons.  The first gives the separator, as in the existing dialog.  The second has label Pattern and then an up-down control, with default 1 and going from (say) 1 to 10!  This then takes always the first n characters of the names.

But can we try try in the script first?

lilyclements:
OK, so the user should either be able to use the names separator, or stack into N sets (but not both)

If they use the names separator, then the code we currently run is fine. 

```r
afl_long <- afl %>%
  dplyr::select(-c("year","round","weekday","day","month",
                   "is.final","venue","attendance")) %>%
  dplyr::mutate(id = dplyr::row_number()) %>%
  tidyr::pivot_longer(
    cols = c(home.team, away.team,
             home.score, away.score),
    names_to = c("location", ".value"),
    names_sep = "\\."
  )
```

If instead they say to stack into N (here, 2) sets, then we run:

```r
anscombesquartet_stacked <- anscombesquartet %>%
  dplyr::mutate(id = dplyr::row_number()) %>%
  tidyr::pivot_longer(
    cols = c(X1, X2, X3, X4, Y1, Y2, Y3, Y4),
    names_to = c(".value", "variable"),
    names_pattern = make_names_pattern(items = c("X1", "X2", "X3", "X4", "Y1", "Y2", "Y3", "Y4"), 2)
  )
```

where, `2` is the number in the nud, and the `items` parameter value is your variables in the nud


I will add the following functino into instatExtras tomorrow so that this works

```r
make_names_pattern <- function(items, num) {
  groups <- instatExtras::split_items_in_groups(items, num)
  
  # Find the longest common prefix of a vector of names
  common_prefix <- function(x) {
    if (length(x) == 1) {
      return(x)
    }
    
    chars <- strsplit(x, "", fixed = TRUE)
    n <- min(lengths(chars))
    
    i <- which(vapply(
      seq_len(n),
      function(i) length(unique(vapply(chars, `[`, character(1), i))) > 1,
      logical(1)
    ))[1]
    
    if (is.na(i)) {
      paste0(vapply(chars[[1]], identity, character(1)), collapse = "")
    } else {
      paste0(chars[[1]][seq_len(i - 1)], collapse = "")
    }
  }
  
  prefixes <- vapply(groups, common_prefix, character(1))
  
  variable_names <- substring(
    groups[[1]],
    nchar(prefixes[[1]]) + 1
  )
  
  paste0(
    "(",
    paste(prefixes, collapse = "|"),
    ")(",
    paste(variable_names, collapse = "|"),
    ")"
  )
}
```

Vitalis95:
@rdstern , please check

rdstern (changes requested):
@Vitalis95 and @lilyclements that's great - it now works!

I'm happy, though I thought it would work differently, i.e. by noring that the first letter was x and then y in the 2 sets.  So we would use the 1 or x-character prefix in the name, rather than the suffix.  Here is would be 1 character.

Assuming we keep the current code I have 2 small suggestions:

Image

a) In the dialog it says the resulting Factor, by default is called variable.  Could it be made into a factor?  

b) Could you put the Radio button Stack into N sets on the line below.  Then omit the Number of Sets label, and just have the up-down

lilyclements:
@Vitalis95 for (a), you just awnt to add in "  dplyr::mutate(variable = factor(variable))" at the end, i.e., 

```r
anscombesquartet_stacked <- anscombesquartet %>%
  dplyr::mutate(id=dplyr::row_number()) %>%
  tidyr::pivot_longer(cols=c("X1","X2","X3","X4","Y1","Y2","Y3","Y4"),
                      names_pattern=instatExtras::make_names_pattern(items=c("X1", "X2", "X3", "X4", "Y1", "Y2", "Y3", "Y4"), 2),
                      names_to=c(".value", "variable")) %>%
  dplyr::mutate(variable = factor(variable))
```

Vitalis95:
moved to #10456

PR #10417, opened 2026-07-16
Improvements to Record Quarto System

### LINKED ISSUE #10403: Improvements to Record Quarto System
1. When we press OK, it should essentially run a "To Script and OK" on Quarto Mode
2. When we are in "Quarto Mode" we need to work out which buttons should change here and how. And any other features we should disallow. (E.g., check how the GetPass stuff works with it!)

Image

3. An error occurs with PDF if `tinytex` is not installed. Quarto fails to set up its PDF template because LaTeX isn't available. The user needs to install TinyTeX to fix PDF rendering, or switch to html format to avoid the dependency entirely.

The code that will fix this is:

```r
install.packages("tinytex")
tinytex::install_tinytex(force = TRUE)
```

a) Is `tinytex` already installed into R-Instat?
b) We need to investigate a way to do this in R-Instat.

Options: 
* Default to HTML output while this is being resolved, since HTML works without any additional dependencies
* If PDF is selected and LaTeX is not detected, surface a clear message explaining the issue and the fix
* Add a tooltip in R-Instat near the PDF option explaining that TinyTeX must be installed for PDF to work
* Add a check when the user selects "PDF" on the "Record Quarto" dialog that detects whether TinyTeX is installed, and warns the user proactively if it is not
* Add a note to the docs that the above R code must be run before PDF output will work

Vitalis95:
Fixes #10403 
@lilyclements @AmsaleEjigu , I have added the Knitr options

- [x] Runs without errors  


AmsaleEjigu:
@Vitalis95, it seems something is disabling the selection of columns. Once I start Quarto Record, I cannot select a column into the receivers (no column is shown). I restarted the session and checked dialogs before Record Quarto, and they are fine, but not when Record Quarto is started. Note that Stop Quarto did not restore the functionality.

image


Screenshot 2026-07-17 180849

AmsaleEjigu:
@Vitalis95, almost there. Now I can select columns, and the OK button is active - great! However, the OK button produces the quarto report as opposed to producing a graph for a dialog and the To Script button is inactive.

image

lilyclements:
@Vitalis95 great! So, pressing "OK" on a dialog currently

1. Adds the R code to the qmd script
2. Renders the qmd script

But, we want pressing "OK" on a dialog to:

1. Add the R code to the qmd script
2. Run the R code in the output window

Essentially, for it to work just like if we pressed "To Script + OK" on the "Non-quarto" mode

Vitalis95:
@lilyclements , please check it again

lilyclements (commented):
@Vitalis95 nice! Works well and looks good! Good job!

@rdstern over to you to review.

We just need to think of where to run the line:

```
tinytex::install_tinytex(force = TRUE)
```

jkmusyoka (approved):
This is good, ready to be merged

lilyclements (approved):

PR #10423, opened 2026-07-23
ci(translation-check): allow fork PR checkout under pull_request_target

kisinga:
Recenntly, GitHub released actions/checkout v7 with a new security default: in privileged workflows (pull_request_target, workflow_run), and as a result, actions/checkout now refuses fork PR code in pull_request_target workflows unless opted in, failing the check before it runs. 

Opt-in is reviewed-safe: PR checkout is data-only (persist-credentials false, no PR code executed; only base-repo scripts run against it).

Also bump checkout v5, setup-python v6, github-script v8 to clear Node 20 deprecation warnings (all run on node24).

### Before asking for review
Please confirm that you have:

- Completed the developer testing checklist below (ticking items as you go)
- Responded to all AI/bot comments (either fixed or explained)
- Added a short note in the PR describing how you tested these changes
- Noted which issue(s) this PR relates to

**Developer Testing Checklist**
- [x] Runs without errors  
- [ ] OK disabled when dialog is incomplete or invalid  
- [ ] OK enabled only when required inputs are valid
- [ ] Reset returns dialog to its default/sensible state
- [ ] Invalid inputs are handled cleanly (e.g. negative, too-large, empty, impossible combos)
- [ ] Running twice with different settings behaves consistently (e.g., open → run → close → reopen → change options checked → run again)
- [x] All AI/bot comments addressed (fixed, intentionally ignored with explanation, or queried)

Issue #10065, opened 2025-11-05
Changing current pull requests workflow to be directly on the main repository through branches instead of developer's repositories


Patowhiz:
@ChrisMarsh82 as discussed with @kisinga we could have developers working on the main repositories and making PR's through branches that are created in the main repository then configuring the Git to automatically delete the branches after PR's are merged.

I think this will make new developers onboarding easier and accelerate development workflows.

@kisinga may add more to this.

kisinga:
@Patowhiz  Thanks for looping me in.

To help me better understand the motivation behind the current workflow, I’d like to check if I’m missing any obvious advantages. As I understand it, the current flow is:

Fork R-Instat → create feature branch → open PR to upstream

This has the benefit of separation and safety, but beyond that, I don’t see many other clear gains, especially for developers who are part of the core team or for non-dev collaborators who just want to review or test PRs directly.

@ChrisMarsh82 The proposal we discussed with @Patowhiz simplifies this to:

A single shared repository → feature branches created directly on the main repo → branches auto-deleted after merge (with “squash commits” enforced)

The main goal is to reduce friction for active contributors and make onboarding smoother—particularly for those who may not be familiar with managing forks or setting up remotes. It also makes it easier for reviewers to check out PR branches locally and run tests without needing to fetch from multiple origins. We debugged an issue with @berylwaswa relating to this, and I thought a better approach might be to simplify the entire workflow altogether.

Are there any specific drawbacks you foresee with this approach, maybe, around permissions, repo protection rules, or contributor access levels? If not, I think this could streamline the workflow quite a bit.

lilyclements:
@Patowhiz can you please follow up with @ChrisMarsh82 on this? Or give his response here so we know how to proceed. Thanks.

PR #10426, opened 2026-08-03
simplifying translation check

ChrisMarsh82:
Testing changes to translation check

### Before asking for review
Please confirm that you have:

- [ ] Runs without errors  

### DIFF
--- a/.github/workflows/translation-check.yml
+++ b/.github/workflows/translation-check.yml
@@ -6,6 +6,9 @@ on:
     branches: [ "master" ]
   pull_request:
     branches: [ "master" ]
+    paths:
+      - 'instat/**/*.vb'
+      - 'instat/translations/**'
 
 permissions:
   contents: read
@@ -38,7 +41,7 @@ jobs:
         run: |
           git remote add base "https://github.com/${{ github.repository }}.git"
           git fetch --no-tags base "refs/heads/${{ github.event.pull_request.base.ref }}:refs/remotes/base/${{ github.event.pull_request.base.ref }}"
-    
+          fetch-depth: 0   
 
       - name: Setup Python
         uses: actions/setup-python@v6
@@ -47,12 +50,9 @@ jobs:
 
       - name: Run translation check
         id: translation-check
-        working-directory: pr
-        # Run trusted scripts from the base checkout against the PR checkout.
-        # The PR checkout is treated only as data; no PR code is executed.
         run: |
           set +e
-          python ../base/scripts/check_translations.py --ci --base="base/${{ github.event.pull_request.base.ref }}" > translation-report.txt 2>&1
+          python scripts/check_translations.py --ci > translation-report.txt 2>&1
           status=$?
           cat translation-report.txt
           exit $status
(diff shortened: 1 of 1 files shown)


kisinga:
@ChrisMarsh82 thanks for pushing on this. You're right that pull_request + single checkout is simple and better.

Some history: the double checkout on master exists because I wanted the check to post its report as a PR comment. Commenting on fork PRs needs a write token, which forces pull_request_target, which forces PR code to be treated as data only. 

Your draft made me reconsider whether the comment is worth keeping and I dont think it is. The report, the missing strings, and the fix instructions can all live on the check page itself whereby clicking a failed check is good enough, and it works for forks with no token at all.

I have a commit ready that takes your draft to that end state:

- Drops the comment step entirely as it would 403 on fork PRs anyway (read-only token under pull_request), and everything it showed is now on the job summary
- `permissions: contents: read` nothing in the workflow writes anymore
- Fixes the push trigger to diff against the previous commit (--base), so it actually checks something. As drafted it compares master against itself and would always report "0 files checked"
- Drops workflow_dispatch for the same reason above it could only ever produce a decorative green run
- Restores the paths filter, extended to the checker scripts and this workflow, so tooling changes test themselves
- checkout v3 → v5

With this, #10423 is superseded and I'll close it. I'm happy to push here if you are.

ChrisMarsh82:
@kisinga I am happy for you to push your changes here. Can I just check. Why do you want to get rid of the comment? It would still be useful on the PR. It may not work on a forked repository but I don't think that would matter; we still want the comment added on this repository

kisinga:
> Why do you want to get rid of the comment? 

@ChrisMarsh82 I was under the impression that most PR's are from forks? I'm probably missing something, but I dont know what value this would have for most devs? Plus it would be consistent if the behaviour is same for both forks and "internal PRs"?

kisinga:
Just to clarify in case theres a confusion, This is my understanding: PR comments always live on this repo and a fork PR's conversation is right here. However, the workflow's token is read-only for fork PRs, so the comment can't be created into our parent repo here. Does this align?
Meanwhile, I've pushed changes having removed the commenting. I'm happy to add it if we decide to

ChrisMarsh82:
@kisinga I am happy to merge this, although I think we should put the comments back in. But that may be a longer conversation about contributors not working on forks. I don't see any benefit to developers working from a forked repo. I think we originally set it up like this before really understanding how github works. The only reason for someone forking is when they plan to make changes that aren't being pushed to the main repo

@lilyclements @rdstern Should we look into moving away from forks?

kisinga:
> @kisinga I am happy to merge this, although I think we should put the comments back in. But that may be a longer conversation about contributors not working on forks. I don't see any benefit to developers working from a forked repo. I think we originally set it up like this before really understanding how github works. The only reason for someone forking is when they plan to make changes that aren't being pushed to the main repo
> 
> @lilyclements @rdstern Should we look into moving away from forks?

Thanks @ChrisMarsh82 
I had actually opened a separate issue in alignment with this #10065. I'm happy to discuss further.

PR #10427, opened 2026-08-03
Fixes a bug when saving data with the select feature in R-Instat

### LINKED ISSUE #10424: Problem saving with the select feature in R-Instat
@Vitalis95 I hope this will be an easy bug (or two?) to fix?

[survey with selects.zip](https://github.com/user-attachments/files/30369943/survey.with.selects.zip)

a) `**_Problem when saving.`_** - I now find this is the only real problem.

The data is the ordinary survey and I have added 2 selects.  One of them selects the first 2 columns, plus the other numeric columns.  
The second also has the first 2 columns plus the factors.

Now selects deal with columns like filters deal with rows.  The filters are working fine, so I first explain something you can do with a filter.
a) Define a filter - say the first 2 villages.
b) Apply the filter.  The row column goes red.
c) Save the data with the filter applied.  File > Save As > Save Data As.
d) Use File > Close Data Book
e) Re-open the data you just saved.

The data is as you saved it, i.e. the filter is still applied.  (That's nice, but not essential.)
Right-click and Remove Current Filter.  
Now I have all the 36 rows - as expected.

Ideally the same should happen with selects.
a) So I apply the numeric select.  The names go red and I (correctly) have the 5 names.
b) Save the data with the Select applied, i.e. File > Save As > Save Data As.
c) Close Data Book
d) Reopen the data you just saved.

Just as for the filter, it re-opens with the select applied.  
But when I right-click and remove the Select it still has only the 5 columns.  The others have gone!

I don't mind if the File Save turns off the Select first.  So, the data comes back without the select applied.  (It is easy to apply the select again.)  I do mind that it has lost the remaining data!

Vitalis95:
Fixes #10424 
@rdstern @berylwaswa , This turns off the Select first before saving

- [x] Runs without errors  


berylwaswa (commented):
@Vitalis95 this removes the select when the data is saved, @rdstern I hope this does close to what you described

rdstern (approved):
@Vitalis95 many thanks, it works fine.  That's a relief.  I am approving.
@lilyclements could you merge?

lilyclements (approved):

PR #10439, opened 2026-08-05
Added Ggside Plots to 2/3 Variable Graphs

### LINKED ISSUE #8800: Improvements to Describe > Graph > Scatterplot

Emmanuel-Afrifa:
Fixes #8800
@rdstern @lilyclements @berylwaswa @derekagorhom @fran2or @Vitalis95 @Ag-Derek 

This PR adds the side plots.

@rdstern, please can you review and let me know if I may have left anything out? Thank you

- [x] Runs without errors  


lilyclements:
@Emmanuel-Afrifa just a small issue here with the layout on the "Save" overlapping with the controls (I don't think this has come from you, but while you're here!)

image

A couple of points

1. Can we rename how it is said in the input drop down (instead of `ysidedensity` say "Density", etc)

2. There are issues with the new checkboxes displaying/removing when you change around. E.g., if you have numerics in the first two receivers, and then switch to factors, then you get an error.

image

3. If we have Factors in the multiple receiver, and Numerics in the second receiver, then we want to have both "xsideplot" and "ysideplot" as visible.
Currently, only one of the checkboxes are (although it still ran both)

image

4. If we have Numerics in the multiple, and Factors in the second, do we want to have both "xsideplot" and "ysideplot" visible?

Emmanuel-Afrifa:
@lilyclements Please, I have worked on the issues you raised. 

To answer your fourth question, when we have a numeric variable in the multiple receiver and a categorical variable in the second receiver, we only have the `Y Side Plot` option being available (visible).

lilyclements:
@Emmanuel-Afrifa do we want only the y side plot to be visible in that instance, or would it make sense to have the x side plot too? I suppose it makes sense to only have y-side plot? 

1. The "fill" variable cannot be the same as the variable in the multiple receiver (only an issue for factors anyway)
2. The "Second variable" cannot be the same as the variable in the multiple receiver
3. If I have numerics in the first and second receivers, and change the drop down options, and press OK, and then reopen the dialog, the drop down options have defaulted to "Density" again.
4. (3. might also be true for factors/numerics, etc)
5. Similarly, if I have two numerics in the multiple receiver, and change to, say, "Histogram". Then, I add a third numeric to the multiple receiver, it defaults back to "Density"

lilyclements:
@Emmanuel-Afrifa looks good. Just some points:

1. If I run with nothing in the fill, it tries to do `ggside::geom_xsidefreqpoly(ggplot2::aes(colour=))`

(or `colour=, fill=` too at points)

image

2. In `geom_xsidefreqpoly` and `geom_xsidehistogram` if we have factors in the multiple, and a numeric in the single, we want to have the extra parameter `stat = "identity"` (otherwise it throws an error about discrete/continuous issues)

3. If we have numerics in the multiple, and a factor in the single, then it currently runs with "x side" (if you have previously run with "x side"). But, this is good. It shows that we actually do want to have both x side and y side plots when we have numerics in the multiple and a factor in the single. So, can you make it so that both "Y Side" and "X Side" options are visible when we have numerics in the multiple, and a factor in the single.

E..g, -

image


4. With 3., in `geom_xsidefreqpoly` and `geom_xsidehistogram` , if we have numerics in the multiple, and a factor in the single, we want to have the extra parameter `stat = "identity"` (otherwise it throws an error about discrete/continuous issues)

Emmanuel-Afrifa:
@lilyclements I have implemented your most recent suggestions. Thank you.

lilyclements (commented):
@Emmanuel-Afrifa really good, well done. Final one is just for the two new checkboxes to reset when you press OK (the checkboxes reset, but the inputs associated to them does not). Otherwise, looks great and works really nicely!

Emmanuel-Afrifa:
@lilyclements I have implemented the change such that the input boxes resets to their default state (Show all available options, with the default text being `Density`. That's undo all the filtering done based on the datatypes of the variables in the first and second receiver) when the `Reset` button is pressed.

@lilyclements, Please, is this what you meant?

lilyclements (approved):
Very good work @Emmanuel-Afrifa! Works brilliantly! Well done!

@Vitalis95 can you review the VB code?
@rdstern can you review?

Vitalis95 (approved):

Issue #10270, opened 2026-03-17
Avoid cef runtime message?


rdstern:
@ChrisMarsh82 this now seems more general and for everyone.  I don't get it elsewhere than R-Instat.

Whenever there is an error in R-Instat everyone now seems to get this error message first:

<img width="861" height="1067" alt="Image" src="https://github.com/user-attachments/assets/0050a3db-dc05-4ba4-9c31-20845ecf5c49" />

Is there something we can do about it?  And is this a you, or a @Patowhiz question?

rdstern:
@ChrisMarsh82 and @Patowhiz are ignoring this problem for now.  @Ag-Derek is this something you could investigate - maybe this week?  If you can't solve it, then I am with @ChrisMarsh82 next week, so I can ask him again!

Patowhiz:
@rdstern @ChrisMarsh82

The CEF runtime error should only be displayed when **R-Instat** is first launched, not when a dialog executes code.

When implementing this, I added a check for CEF runtime initialization specifically during application startup. This was intentional, since some machines may not have up-to-date Windows installations and could fail to initialize the runtime successfully.

The intended behavior was that if the CEF runtime fails to initialize, **R-Instat** would silently fall back to using the machine’s default browser for displaying HTML outputs, without interrupting the user with errors.

It’s also worth noting that the CEF runtime is only required for rendering HTML outputs. The *Tidy Daily* dialog produces a data frame, not an HTML output, so it should not trigger this error.

This behavior is clearly a regression and should be investigated and fixed to align with the original design.

rdstern:
Thanks @Patowhiz for the clarification.  I just add that it now appears on seemingly all implementations and for every error on any dialog.  So there are now always 2 error messages.  The cef comes first and then the specific error message.  I don't know if it helps locate the problem?

@Patowhiz can you suggest who might best examine, and hopefully fix this?  Can it be anyone from our main development team?

Patowhiz:
@rdstern I don’t have a specific person to suggest, but anyone familiar with the base code, particularly how scripts inputs and outputs link to R, and who is comfortable using Visual Studio debugging tools (especially breakpoints) should be able to trace and fix this regression.

kisinga:
Closed by #10443

PR #10443, opened 2026-08-05
Fix spurious/repeated Cef runtime error message (#10270)

kisinga:
- InitialiseCefRuntime now returns True when Cef is already initialised instead of False, which callers treated as an initialisation failure
- Log Cef initialisation failures (full exception) via NLog so the root cause can be diagnosed from user logs
- Move Cef initialisation and its error message from frmMain's constructor into frmMain_Load so the message box is properly owned by the visible main form and can only appear once at startup
- Add temporary PowerShell diagnostic script (scripts/check_cef_runtime.ps1) to verify Cef payload files, architecture, VC++ redist, and cache locks on machines showing the error

### Before asking for review
Please confirm that you have:

- [ ] Runs without errors  


ARMSTRONGOPONDO:
@kisinga I tested the CEF error that was being experienced in both the current branch and the master branch.

To do this, I reused a previously reported issue as a test case, since earlier there were multiple CEF errors occurring before the actual error would appear. Just as a heads-up, the issue I used had already been reported I simply reused it to verify the CEF error behavior.

To Reproduce

1. Go to Climatic > Tidy Data / Prepare > Data Reshape
2. Click on Unstack (Pivot Wider) ...
3. Went to Restore Hierarchy

the CEF error fires first
image

then the actual error
image

the cef error fires again
image

then the another error
image

the cef error fires again
image

then the last error is shown 

image

this was the scenario before testing @kisinga changes.

The CEF error that used to occur at runtime and when opening dialogs in VB is now resolved. For example, when navigating to the solution, searching for `frmMain`, and opening it in the designer, the CEF error would previously trigger before the design view loaded. This no longer happens.

so this PR has successfully resolved the firing of the CEF error in both the runtime and the vb side.
In the above example I now do not get the CEF errors firing at first and in between the subsequent errors.

kisinga:
I highly appreciate your help @ARMSTRONGOPONDO. 
Could you validate that actual errors are still surfaced correctly? Thanks

kisinga:
Cc @lilyclements @rdstern @berylwaswa 
Fixes #10270

lilyclements (approved):

lilyclements:
Been told that this has been reviewed by @ARMSTRONGOPONDO who is approving

Issue #10442, opened 2026-08-05
Change Frmain to include 4 new dialogs


berylwaswa:
Three of these dialogs are in the experiments menu. 
a) Experiments Greater than Design is the first.
b) Experiments greater than Model greater than Multiple Comparisons is the second, and is below the two entries that are there now.
c) Experiments greater than Model greater than One button is the third and is below b.
d) Prepare Data Reshape Greater than Unnest is the forth and is in Prepare > Data Reshape and below the current Merge dialog
Create a dummy empty dialog for each so that whoever is working on those can fill in.

@bayakarisa please work on this first before the other tasks.

PR #10447, opened 2026-08-05
Created  dummy empty dialogs for each so that whoever is working on t…

bayakarisa:
@Vitalis95 @berylwaswa 
This is just putting names of the diaogs in the frmMain and making empty Dialogs at frmMain
Created dummy empty dialogs for each so that whoever is working on those can fill in.

a) Experiments > Design is the first.
b) Experiments > Model > Multiple Comparisons is the second, and is below the two entries that are there now.
c) Experiments > Model > One button is the third and is below b.
d) Prepare Data Reshape > Unnest is the forth and is in Prepare > Data Reshape and below the current Merge dialog


berylwaswa (commented):
@bayakarisa this looks good, @rdstern can you confirm that the names are okay?

lilyclements (approved):

lilyclements:
@bayakarisa I think the idea is for ">" to signify going to the next level, not be interpreted as "greater than".

(I don't know how to "undo an approve")

lilyclements:
Fixes #10442

PR #10456, opened 2026-08-06
Stack dialog improvements

### LINKED ISSUE #10332: Improve the Stack dialog when stacking multiple sets.

Vitalis95:
Replaces PR #10353 
Fixes #10332 
@rdstern @lilyclements , I have made into a factor the resulting variable

- [x] Runs without errors  


lilyclements:
@Vitalis95 great, I am reviewing now. @rdstern can you also review?

rdstern (approved):
@vitalis, many thanks, I am approving.
@kisinga please could you merge

kisinga (approved):

PR #10471, opened 2026-08-11
Added the New Unnest Dialog

### LINKED ISSUE #10339: Request for Table Data Extraction and Transformation Feature in R-Instat
While conducting the data collection research, I used a data collection form that contained single-choice, multiple-choice, and tabular questions. In the exported spreadsheet, all the questions are stored within the same dataset structure.

One issue that captured my attention is how the information from the tabular questions is stored. The entire table data is saved in a single cell, making it very difficult to analyse or perform any meaningful data manipulation because all the values are grouped together.

We tried using another tool, specifically Excel, which worked very well in separating and restructuring the table data into usable columns and rows.

Below is the dataset used:

[Champ.xlsx](https://github.com/user-attachments/files/28013178/Champ.xlsx)

In Sheet1, column C contains the table data. Our objective was to transform this information into a clean and structured dataset. Currently, this cannot easily be achieved in R-Instat, whereas Excel provides a convenient way to do this.

The following are the steps used in Excel to convert the column data into a clean table suitable for data manipulation:

1. Open the file in Excel.
2. Go to the Data menu and select From Table/Range on the left side. This opens the Power Query Editor.
3. Select the table column (in our case, Inventaires_Champs_exploitation), right-click it, choose Transform, then select JSON.
4. On the column header, click the two-arrow expansion icon on the right side and select Expand to New Rows. Repeat the process and select the rows or fields you would like to extract. In our case, all rows were selected.
5. Click OK, then close the Power Query Editor. Excel will prompt you to keep the changes — select Keep.
6. A new table containing the extracted and structured data will then be created.

This is exactly the type of functionality we would like to achieve in R-Instat. However, this feature or workflow does not yet seem to be available or identified within the software.

My concern and request is: would it be possible to implement such functionality in R-Instat to allow easier extraction and further data manipulation for datasets containing tabular JSON-like data stored in single cells?

I will also attach a copy of the questionnaire used for data collection, in addition to the dataset already shared above. 

[Questionnaire ExploitationHM.docx](https://github.com/user-attachments/files/28014125/Questionnaire.ExploitationHM.docx)

@rdstern @lilyclements @volloholic @Rockjunior @berylwaswa @bayakarisa @kisinga @ARMSTRONGOPONDO @Ag-Derek @jkmusyoka  @Emmanuel-Afrifa

Emmanuel-Afrifa:
Fixes #10339 
@rdstern @lilyclements @berylwaswa @Aboubacari @jkmusyoka @Vitalis95 @kisinga 

I have added the new unnest dialog and this PR is ready for review.

@lilyclements, I noticed that when I supply the new ID name, I get the following error message (also shown in the image below)
```R
Error in self$get_columns_from_data(col_names, use_current_filter=FALSE):
Not all column names were found in data

.temp_val <- instatExtras::view_object_data(object = data_book$get_dataframes_from_JSON_columns(data_name = "Feuil1", json_cols=json_cols,key_columns="Matricule", carry_columns=c("Statut", "Gradient"), id_column="ffdys"), object_format="text")
```
json error

@lilyclements, @rdstern  Also, I was thinking about this. Because if you set the ID name, that's the name that be used for the unique ID column created across all the new dataframes. I'm thinking perhaps, it might be a bit misleading in some instances (say for instance, we call it `field ID`, as we have as the default now, then all the new ID columns created for all the dataframes will have this name). Perhaps, could we hardcode the ID name in the R code instead (something like `.ID`). And then it removes the possibly confusion (and it also resolves the error above, since we'd no longer be using).

Please, what are your thoughts on this? Thank you.

- [x] Runs without errors  


rdstern:
@Aboubacari  please can you check and also supply me with the data for testing?  I have copied the data from the issue, but I think that is already partly expanded.  Could you send the data with 290 cases and 5 columns you showed during the sprint?

rdstern:
Thanks @Emmanuel-Afrifa for the data and it is great that this worked for me already.  Very well done.

a) A detail - could you make the label Key Columns into `Key Column(s)`.
b) Could you shorten Check Unique Keys to just `Check Unique`.
c) Could you line up the New ID Name label better with the field, and just call it New ID.
d) Then also suggest a name for the New ID.  I suggest the name of the column, except there is more than 1!  So should it be given automatically and named as the asame as the column each time?  Maybe that control isn't needed.
(The only problem is if that name is repeated in the resulting data frame, but then it could be edited first.)
e) I would prefer that the new ID column is just given as the number.  Then also it, plus the original key, are defined as the key fields.  
f) You also say Column(s) define a key in the check so could you also define the key in the original data frame if it has not been defined as a key earlier?
g) I suggest you don't need to add the structure demographique, etc to the names in the resulting data frames.  
h) And, like the define climatic, could you report what you have done in the output window.  

Very well done - all this.

Aboubacari:
@Emmanuel-Afrifa  I’m very pleased to hear about the progress of this dialogue. @rdstern You can download the data here. 
[Champ.xlsx](https://github.com/user-attachments/files/30984366/Champ.xlsx)

Emmanuel-Afrifa:
@rdstern, I'm sorry for the long week of no response. Please, I need a few clarifications/suggestions on some of the points you raised.
d. I agree that the `new ID` column name would be given automatically and hence that control could be take out. My question is, you suggested reusing the column name as the name of the ID column (that's if there's only one column). Could we instead use something static like `.ID` for the new ID column that signals that it's the ID column always? 

Roger: That would be fine

e. You suggested that we use only numbers as the new IDs. I think that's okay. My worry (and one that I think came up during the discussions on this at the sprint is that will it be descriptive enough if we use only numbers?) But if that won't be an issue, I'm happy to get this done. It'll probably be that @lilyclements makes the changes and we update our databook versions.

Roger: In the example you added to the contents, and I suggest this isn't needed.

f. I believe this is already being done in the new bit @lilyclements  added in.

g. I think this will need @lilyclements' input quite a lot. One of the challenges I had was that at the end of the transformation, I was just selecting the widened (newly extracted) columns, the key columns and the carry columns into the new data frame. In selecting the  extracted columns, the easiest way I could identify them was to use the prefix (which was the original JSON column name), and hence that's why all the new columns (except the ID column) have that prefix. In removing the prefix, it might be a bit challenging for me to select these columns. I believe it'll be good for @lilyclements to have a look and give me some ideas of how I might go about it or alternatively, she fixes that on her end.

Roger: I am really keen to include long variable names unless it is essential.

As all the other changes you pointed out are mostly cosmetic changes, I'm thinking I'll hold on with that temporarily and the next commit I make on this PR will (which will hopefully have all what's been discussed implemented) will that the cosmetic changes as well. Or I can do only that now and commit the changes.

rdstern:
@Emmanuel-Afrifa I am also handing on further checking and discussion to @jkmusyoka who is becoming more involved in this part of R-Instat now.  I note there is now a conflict, and reply to your queries in your comment above.

Emmanuel-Afrifa:
@rdstern @lilyclements @jkmusyoka 
Sorry for being away for a while. We just finished the Maths Camp last week.

This PR seems to have accumulated a lot of conflicts (possibly from its direct link to master now).

It has over 700 conflicts now.

I think it'll be best to close this PR and open a new one.

@lilyclements, @jkmusyoka  Please, have you been able to find a workaround for point g (that @rdstern raised), so that we can hopefully shorten the names of the newly extracted columns in the new dataframe? Also, can I make the changes in the databook package to set the name of the new ID column, or should I wait until the other R bits are ready and I put them all in (or perhaps you put it in)?

Thank you.

PR #10474, opened 2026-08-12
Fixed tolerance control in Compare Columns dialog

### LINKED ISSUE #10440: Prepare > Check Data > Compare Columns has 2 problems to fix
We needed to use this dialog in our recent Zimbabwe work.

It didn't work, so we resorted to using the calculator.  Two problems that should be simple to fix.

Image

a) the tolerance control doesn't work.  It resets to 0 when you return.  When put as a value other than 0 it ignores the value and still uses 0.
b) It only seems to work once.  The next time you use it, there is no error, but also no activity!

I have tried it again today on a small data set, and can't get it to fail on point b).  Could someone please correct a) and look to see if there is something in the code that might cause b).  Don;t spend too long, and @jkmusyoka and I will then test.

I note issue #6418 that suggests an extension.  That's still for later.

Rockjunior:
Fixes #10440, specifically the tolerance control issue.

The tolerance value was resetting to 0 when reopening the dialog, and any non-zero value entered was being ignored. The dialog now correctly retains and uses the specified tolerance value.

@rdstern @berylwaswa Kindly review

@rdstern I haven’t worked on (b) yet. I’d probably prefer to have a short session or make a quick video demonstrating how the dialog works once and then stops, as it seems to be working correctly on my side.

### DIFF
--- a/instat/dlgCompareColumns.vb
+++ b/instat/dlgCompareColumns.vb
@@ -254,7 +254,7 @@ Public Class dlgCompareColumns
                 If rdoByRow.Checked Then
                     ucrInputTolerance.Visible = True
                     clsLessorEqualToOperator.AddParameter("first", clsRFunctionParameter:=clsAbsoluteFunction, iPosition:=0)
-                    clsLessorEqualToOperator.AddParameter("tol", "0", iPosition:=1)
+                    'clsLessorEqualToOperator.AddParameter("tol", "0", iPosition:=1)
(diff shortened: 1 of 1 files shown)


rdstern (approved):
@Rockjunior looks good.  I am approving.
@Vitalis95 another easy one for you I hope?

Vitalis95:
Moved to PR #10504

Issue #10433, opened 2026-08-04
Feature Request: Add a Multiple Mean Comparison Dialog with Significance Letters


Aboubacari:
## Problem

Currently, R-Instat does not provide a dedicated dialog for performing **multiple mean comparisons** with **Compact Letter Display (CLD)**. Users need to rely on R code or several manual steps to obtain pairwise comparisons and significance letters, making the workflow less accessible and less efficient.

## Proposed Solution

Add a new **Multiple Mean Comparison** dialog under **Experiments → Model**.

The dialog could be implemented using the **Biometryassist**  package and should support multiple statistical models. It should provide an intuitive interface that allows users to:

* perform multiple mean comparisons;
* automatically generate significance letters (Compact Letter Display);
* customize tables and graphical outputs;
* export publication-ready tables and figures.

## Benefits

This feature would:

* simplify post-hoc analyses for users with little or no R programming experience;
* provide a consistent workflow for multiple comparison procedures;
* improve the presentation and interpretation of experimental results through customizable outputs.
 

@rdstern @berylwaswa @lilyclements @jkmusyoka

lilyclements:
@Aboubacari great. Have you, or someone, looked into the `biometryassist` package, and can you/they outline what we want from it?

Aboubacari:
@lilyclements  The **BiometryAssist** package provides a simple and effective workflow for performing **multiple mean comparisons** following an ANOVA fitted with `lm` or `aov`. We have found this package to be particularly useful in our context.

Currently, the package examples mainly focus on **single-factor ANOVA**, but the same approach could be extended to support **models with multiple factors and interactions**, which are commonly used in agricultural and experimental research.

The core workflow is straightforward:

### 1. Fit an ANOVA model

```r
model <- aov(Petal.Width ~ Species, data = iris)
```

### 2. Perform multiple mean comparisons

```r
multiple_comparisons(model, classify = "Species")
```

### 3. Produce a graph with significance letters (Compact Letter Display)

```r
autoplot(
  multiple_comparisons(model, classify = "Species"),
  label_height = 0.95
)
```

berylwaswa:
@bayakarisa  look into this please

rdstern:
@bayakarisa here is my attempt at the structure of the dialog.  

Note the dialog comes after the user has run aov, or lm, or glm, etc.  And it uses the model objects created.  This is the first dialog to use the model objects

[...]

Issue #9751, opened 2025-05-28
multiple comparisons


IssoufKon:
**Is your feature request related to a problem? Please describe.**
In our research community (that of West Africa), after performing ANOVA analyses, researchers seek to make comparisons between different options. They want to know whether or not there is a significant difference between the options. This question has come up several times in our research method support activities. 

**Describe the solution you'd like**
We would like to add an option, perhaps in the “experimental design” menu, a comparison sub-menu. In this sub-menu, we would like to see:
1) Design the model,
2) Use the model for comparisons: Here we only want to see options for comparisons such as emmeans. 
3) Make graphical representations of these comparisons using packages such as ggplot2, postHoc.

**Describe alternatives you've considered**
Any other multiple comparison option will be welcome. The most important thing is to add this option to R-Instat and make it easy to use for researchers. 

**Additional context**
Some examples of outings we would like to see:

![Image](https://github.com/user-attachments/assets/fe1430cc-5af6-4117-acf9-beb5db24f2af)

![Image](https://github.com/user-attachments/assets/d620301a-e25f-4615-91d1-df7720d2d572)

![Image](https://github.com/user-attachments/assets/43085a71-5c05-4eb6-8c35-027883be55dc)

![Image](https://github.com/user-attachments/assets/46314f28-890c-4f51-9535-5d3f5d1af61c)

![Image](https://github.com/user-attachments/assets/156d97cc-5de5-4627-8e6b-f46f01f9e531)

![Image](https://github.com/user-attachments/assets/4f6165f8-5749-4f1a-8b40-0ba7c7ad210e)

lilyclements:
@IssoufKon this is great timing, and a good reminder for myself too on the tricots work we are on at the moment! We want to be able to compare models there too. 

Particularly to point 2 and 3 -- can you outline what different comparisons you would like?
These plots are good too - have you made them with ggplot2 and can you share codes on them?

Emmanuel-Afrifa:
@lilyclements @IssoufKon @Ag-Derek @rdstern @fran2or

<img width="295" alt="Image" src="https://github.com/user-attachments/assets/079a9cbd-5c34-4044-9bbb-b8b613c52053" />

This is a sample design of how the dialog may look based on our discussions last week.

The 3 factor receivers will receive values from the second (data frame) selector and these are the values that will appear after the `~` operator in the emmeans function...
`emmeans(mod, ~ Species)`

[...]

PR #10475, opened 2026-08-12
Add Multiple Mean Comparison dialog using biometryassist

bayakarisa:
Fixes partly #9751 and #10433

@lilyclements This touches on the frmMain as i had disabled it before. Kindly Check

@rdstern I made the dialog for Multiple Mean Comparison Under Experiments then Model
This is just what I Have done. I am still working on this is the first push

image

@berylwaswa For testing I used the Iris data. First You have to have a model for this dialog
I implemented that the two receivers for label Variable and By take only Factors as this was what was requested by @rdstern 

The Transformation Button is disabled for now

image

---
This is the dialog and the output

image

This is the code just Three lines

```
# Dialog: Multiple Comparisons

model_tmp <- data_book$get_object_data(data_name="iris", object_name="biometry_model")
mct_obj <- biometryassist::multiple_comparisons(model.obj=model_tmp, classify="Species")
mct_obj
rm(list=c("model_tmp"))
rm(mct_obj)
```

@Vitalis95 Kindly check the code

### DIFF
--- a/instat/dlgModelMultipleComparisons.vb
+++ b/instat/dlgModelMultipleComparisons.vb
@@ -142,11 +142,9 @@ Public Class dlgModelMultipleComparisons
             ucrChkAdjustment.SetRCode(clsMultipleComparisonsFunction, bReset)
 
             ucrBase.clsRsyntax.AddToBeforeCodes(clsAssignModelOperator)
-
             ucrBase.clsRsyntax.SetBaseRFunction(clsMultipleComparisonsFunction)
-            ucrBase.clsRsyntax.SetAssignTo(ucrSaveModelMultipleComparisons.GetText())
-
             ucrBase.clsRsyntax.AddToAfterCodes(clsRmFunction)
+
         End If
     End Sub
 
@@ -187,7 +185,12 @@ Public Class dlgModelMultipleComparisons
 
     Private Sub ucrSaveModelMultipleComparisons_ControlValueChanged(ucrChangedControl As ucrCore) Handles ucrSaveModelMultipleComparisons.ControlValueChanged
         If ucrSaveModelMultipleComparisons.GetText() <> "" AndAlso ucrSaveModelMultipleComparisons.IsComplete() Then
-            ucrBase.clsRsyntax.SetAssignTo(ucrSaveModelMultipleComparisons.GetText())
+            clsMultipleComparisonsFunction.SetAssignToOutputObject(
+            strRObjectToAssignTo:=ucrSaveModelMultipleComparisons.GetText(),
+            strRObjectTypeLabelToAssignTo:=RObjectTypeLabel.Summary,
+            strRObjectFormatToAssignTo:=RObjectFormat.Text,
+            strRDataFrameNameToAddObjectTo:=ucrSelectorModelMultipleComparisons.strCurrentDataFrame,
+            strObjectName:=ucrSaveModelMultipleComparisons.GetText())
         End If
     End Sub
 
--- a/instat/dlgModelMultipleComparisons.vb
+++ b/instat/dlgModelMultipleComparisons.vb
@@ -14,66 +14,196 @@
 ' You should have received a copy of the GNU General Public License 
 ' along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-Imports instat
 Imports instat.Translations
+
 Public Class dlgModelMultipleComparisons
-
-
     Private bFirstLoad As Boolean = True
+    Private bReset As Boolean = True
+
+    Private clsMultipleComparisonsFunction As New RFunction
+    Private clsGetModelFunction As New RFunction
+    Private clsAssignModelOperator As New ROperator
+    Private clsRmFunction As New RFunction
 
     Private Sub dlgModelMultipleComparisons_Load(sender As Object, e As EventArgs) Handles MyBase.Load
         If bFirstLoad Then
             InitialiseDialog()
             bFirstLoad = False
         End If
-
+        If bReset Then
+            SetDefaults()
+        End If
+        SetRCodeForControls(bReset)
+        bReset = False
+        autoTranslate(Me)
+        TestOkEnabled()
     End Sub
 
     Private Sub InitialiseDialog()
 
+        ucrBase.iHelpTopicID = 0
+        ucrBase.clsRsyntax.bExcludeAssignedFunctionOutput = False
+        ucrBase.clsRsyntax.iCallType = 2
 
+        ucrReceiverMultipleMeanComparisonUseModel.SetItemType(RObjectTypeLabel.Model)
+        ucrReceiverMultipleMeanComparisonUseModel.Selector = ucrSelectorModelMultipleComparisons
+        ucrReceiverMultipleMeanComparisonUseModel.SetMeAsReceiver()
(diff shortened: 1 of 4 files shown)


bayakarisa:
@rdstern
I have changed how changed how mct_obj's output is assigned

image


This is the code now 

```
# Dialog: Multiple Comparisons

model_tmp <- data_book$get_object_data(data_name="iris", object_name="biometry_model")
mct_obj <- biometryassist::multiple_comparisons(model.obj=model_tmp, classify="Species")
data_book$add_object(data_name="iris", object_name="mct_obj", object_type_label="summary", object_format="text", object=mct_obj)
data_book$get_object_data(data_name="iris", object_name="mct_obj", as_file=TRUE)
rm(list=c("model_tmp"))
rm(mct_obj)
```
I am going to look for @Vitalis95 and discuss with him the vb code

rdstern:
@bayakarisa that's great, that you have it working so soon.  So we can already get down to the details:

image

a) The vertical spacing, of the receivers could be slightly improved.
b) The Descending option doesn't work yet.  You don't yet include the changed R code, when you change that option.
c) I missed one additional option.  Add another checkbox please with label `Display Letters`.  It is ticked initially and has a drop-down with TRUE, FALSE (like Descending) and TRUE is the default.  It is above Descending. It is always given in the command you generate.  It is the groups parameter, so add groups = TRUE to the code you generate.
d) The Store mct needs a checkbox, with default unchecked (as we usually do)
e) I like the name of the incoming objects also statiing the y-variable, e.g. Anova_Sepal.Width.  So can the names here also include that feature.  So, in the example above, it would start as `mct_Sepal.Width`.
f) In the R code you generate, see above, I note you always start with the first object and then also the one analysed.  Please omit the 2 extra lines in the code.
g) The results, in the output window don't say which variable is being analysed.  Could you add a line that includes this?  For example Multiple Comparisons for Sepal.Width?
h) Finally, (I hope) there will later need to be a change in the way the Label Variable and the By receivers are done.  I like what you have done now, because it works and is simple.  But they will later need to become drop-downs (I assume) with the factors from the anova object - and maybe filling automatically if there is just one.  This will need help from @jkmusyoka or others on the R side.

bayakarisa:
@rdstern I have done necessary changes to this that is all except g and h. I am working on completing g since h is for later on.

image

image

This is the script 

```
# Dialog: Multiple Comparisons

model_tmp <- data_book$get_object_data(data_name="iris3", object_name="anova_Sepal.Width")
biometryassist::multiple_comparisons(model.obj=model_tmp, classify="Species", sig=0.05, int.type="ci", descending=TRUE, adjust="tukey", groups=TRUE)
rm(list=c("model_tmp"))
```

Display Letters is shown as **groups** just as in the documentation and Confidence Interval as **int.type**  and Alpha as **sig**

rdstern (dismissed):
@bayakarisa that's very nicely done.  Great.  Many thanks.
The only thing to change for now is that the saving of the mct object, should increment, so it doesn't overwrite.  
That could come later, if it takes you a long time, as it would be good to get this merged as soon as possible.  So I am approving for now, and asking @Vitalis95 to check the code please, and merge once you are ok with it.
@lilyclements there remains one major improvement needed to the dialog, that needs R skills.  I assume the functions may exist.  Currently the second and third receivers in the dialog list the factors in the data-frame.  That's ok for now, but they should just list the factors (in a pull down in the receiver), that are in the model.  That's the terms on the rhs of the aov, or lm, etc model that are factors.

instat/dlgModelMultipleComparisons.vb
```diff
@@ -14,66 +14,226 @@
...
     Private Sub ucrBase_ClickReset(sender As Object, e As EventArgs) Handles ucrBase.ClickReset
         SetDefaults()
         SetRCodeForControls(True)
-        TestOkEnabled()
-    End Sub
 
-    Private Sub ucrReceiverFirstVariables_Load(sender As Object, e As EventArgs) Handles ucrReceiverFirstVariables.Load
+        If Not String.IsNullOrEmpty(ucrSelectorModelMultipleComparisons.strCurrentDataFrame) Then
+            clsGetModelFunction.AddParameter("data_name", Chr(34) & ucrSelectorModelMultipleComparisons.strCurrentDataFrame & Chr(34))
+        End If
```
Vitalis95: `data_name` is added both in ucrBase_ClickReset and DataFrameChanged. Could we keep this in DataFrameChanged only, to avoid potentially adding the same parameter more than once?

instat/dlgModelMultipleComparisons.vb
```diff
@@ -14,66 +14,226 @@
...
+
+            ucrChkAlpha.SetRCode(clsMultipleComparisonsFunction, bReset)
+            ucrChkConfidenceInterval.SetRCode(clsMultipleComparisonsFunction, bReset)
+            ucrChkDisplayLetters.SetRCode(clsMultipleComparisonsFunction, bReset)
+            ucrChkDescending.SetRCode(clsMultipleComparisonsFunction, bReset)
+            ucrChkAdjustment.SetRCode(clsMultipleComparisonsFunction, bReset)
+
+            ucrBase.clsRsyntax.AddToBeforeCodes(clsAssignModelOperator)
+            ucrBase.clsRsyntax.SetBaseRFunction(clsMultipleComparisonsFunction)
+            ucrBase.clsRsyntax.AddToAfterCodes(clsRmFunction)
```
Vitalis95: I think the `base syntax` setup should be moved into `SetDefaults(),` as we normally initialise the R functions and before/after codes there. This would also make the reset behaviour clearer and avoid relying on `SetRCodeForControls(True)` to initialise the syntax.

Vitalis95 (commented):
@bayakarisa great work, some few comments

When the `Store mct `checkbox is checked, clicking Reset does not reset the save control back to its default unchecked state. Could you check the reset handling for `ucrSaveModelMultipleComparisons`? The reset should also clear the save/assignment state so that after Reset the generated code no longer contains the stored `mct `assignment.

instat/dlgModelMultipleComparisons.vb
bayakarisa: Removed this from set defaults

instat/dlgModelMultipleComparisons.vb
bayakarisa: Moved this to set defaults

bayakarisa:
@Vitalis95 thanks for the review. I have completed the code changes should be good to merge.

1. I have moved  the Base syntax setup to SetDefaults()
2. I removed the duplicate data_name from SetDefaults()
3. I have fixed the Store mct checkbox resetting and now it is good

Vitalis95 (approved):
@bayakarisa , thanks
@rdstern , please check and approve again

rdstern (approved):
Many thanks @bayakarisa and @Vitalis95 I am now approving and merging.

I note there is still work to be done on this dialog, so there will very soon be another pull request.
a) In the receivers we need to get the factors from the object and not from the data frame.
b) There is still the Transform sub-dialog to add.  (That can wait.) 
b) There is an autoplot function in the package.  I suggest we will probably have a separate dialog, as well as a simple plotting option on this dialog.  I still need to think this out more carefully for the different design options.

But meantime this corresponds to the main request from the McKnight team, so the testing. plus planning of how to use this new feature, can start now.

I am delighted this has been done so quickly.  Great.

PR #10476, opened 2026-08-18
Refactor dlgEnter UI and add new button functionalities

ARMSTRONGOPONDO:
@rdstern @lilyclements @Vitalis95 @berylwaswa 
I have implemented the keyboard changes on the enter dialog 
 Reorganized extra keyboard (rep_times, rep_each, rnormal, rfactor, etc.)
 Length label disabled until Try succeeds
 Not in this PR
 Filter support  waiting for Calculator filter implementation to be confirmed before implementing it to Enter


ARMSTRONGOPONDO:
@rdstern  can you please confirm the filter in calculator so that I could work on it in the enter dialog.

rdstern:
@ARMSTRONGOPONDO I would like to check the current branch first, but you seem to be on the new "system" and I can't locate it, under the remotes?  Can you help me to find it?

ARMSTRONGOPONDO:
@rdstern am not sure but here is the exact name of the branch `origin/enetr_calc_keyboard ` maybe @kisinga can confirm how to do that.
cause on my end its their maybe try to sync your local files.
image

rdstern:
@ARMSTRONGOPONDO and @kisinga I have just re-loaded the master and now only have a small list of remotes!  And yours is one of them, so my problem is solved.  I'm testing now.

rdstern (commented):
@ARMSTRONGOPONDO this is really nice.  I would like to see this merged before you "attack" the filtering part.
You have included a new feature I really like in 2 of the keys here:

Image

The new feature is that you include `n =  36` here - given just like that!  That's the length of the data frame.  How did you do that?

I ask because the calculator has lots of keys like that.  There the expression is given as 
`n=data_book$nrows(x="survey")`, instead.  That's by @Rockjunior of course.  Now I prefer what you have done, because it is much simpler for the user.  Except, I note if I filter the data - e.g. just include the first 3 villages for example, that the function used in the caclulator gives the filtered count (e.g. 26) while you always give 36.   

This is also making me reflect that possibly we keep Enter as not allowing filters (i.e. if data are filtered we remove it first.  While calculate does permit filters!  I need to ask others about this.  In that case what you have done there is great.  Can you please explain how?

rdstern (approved):
@ARMSTRONGOPONDO very well done.  I am now approving.  Having discussed with Danny I now suggest that we leave Enter to ignore filters.  Danny is also keen that we indicate in the calculator whenever it does take account of filters.  
@Vitalis95 could you please check and hopefully merge?

ARMSTRONGOPONDO:
Thanks @rdstern 
  For **rnormal** and **rfactor**, when the key is clicked the dialog reads the row count in VB from the selected data frame and inserts it as a plain integer in the R expression. 
In `dlgEnter.vb`:
  ```vb
  Dim iNrows As Integer = ucrSelectorEnter.ucrAvailableDataFrames.iDataFrameLength
  clsRfactorFunction.AddParameter("n", iNrows.ToString(), iPosition:=0)
  ```
  `iDataFrameLength` (in `ucrDataFrame.vb`) returns:
  ```vb
  frmMain.DataBook.GetDataFrame(dataframeName).iTotalRowCount
  ```
  So for `survey` with 36 rows, the user sees e.g.:
  ```r
  wakefield::r_sample_factor(n = 36, x = c("a", "b", "c"), prob = NULL)
  ```
  rather than the Calculator style:
  ```r
  wakefield::r_sample_factor(n = data_book$nrows(x = "survey"), ...)
  ```
 Why I did it this way (including the Try errors I hit while testing) My first implementation copied the Calculator pattern (`data_book$nrows(x = "...")`). When I pressed **Try** in Enter, that failed with errors like:
  ```text
  Error in rnorm(n = data_book$nrows(x = "survey"), mean = 0, sd = 1) :
    attempt to apply non-function
  ```
  **rfactor** failed similarly, because `n` did not evaluate to a valid length.Enter evaluates expressions in its own way: it **attaches the data frame** before running the expression through **Try**. In that path, the dynamic `data_book$nrows(...)` call did not work reliably for me. Using the row count from the VB data book (`iTotalRowCount`) and inserting a literal integer (e.g. `n = 36`) was a practical fix that makes **Try** succeed and is easier for users to read and edit.

Vitalis95 (approved):

PR #10488, opened 2026-08-23
Fixed Bug when Plot Options is used in the Use Graph Dialogue

### LINKED ISSUE #10486: Use Graph dialog has error when Plot Options used
I think this is general, i.e. for any data.  I first found it in the development version and here it is again in version 0.8.15.  Earleir I used a bar chart, here a scatterplot.  I think it is for anything.

@derekagorhom please could you have a look?

@Ag-Derek which name do you now usually use?

Image

In the deverlopment version it is here - for a different dataset, but same error message:

Image

Ag-Derek:
Fixes #10486 
This fixes the issue when the user selects the plot option in the Use graph dialogue which creates and error in both the developer version and the standalone. 

- [x] Runs without errors  


rdstern:
@Ag-Derek that's great - that bug fixed! 

But sorry it was just fixed until the next one!  Which is here:

image

I fear there is more to do and delighted you have started so soon.

I used the usual survey data and tried a simple bar chart of the Varieties with the Villages as facets.  As is has crashed I'll start again and describe in detail below.

Here is the dialog and the results:

image

Then I go into the Use Graph dialog:

Now I can get to the Plot Options.  I went to the layers and they were empty.  See below:

image

Clicking on Add gives the error.  

The solution might be complicated, because ideally the bar geom would be here, as read from the ggplot2 object.  Is this something you could attack?  It may need Francis, or Lily R skills I assume?

Ag-Derek:
@Vitalis95 and @Emmanuel-Afrifa  do you guys have any idea why the layers option does not show the geom bar? since you have worked on the plot option sub dialog before. if not then we can ask lily or Francis

rdstern:
@Ag-Derek @Vitalis95 and @lilyclements note it isn't just the layers that are "empty", but also the facets and (I assume) others.  That's because the current Use Graph doesn't try to "read" the ggplot2 object at all.  
So, I assume we can add to the command (literally add additional `+ `bits, to the command) but maybe not change details of existing layers, etc.  If so, we can use the titles and themes tabs, (maybe the axes?) but should not mess with the others - so it would be a limited sub-dialog that is available?

Ag-Derek:
@rdstern i suggest this is done as a separate issue since this involves objects along with more explanation given.
Thanks


> @Ag-Derek @Vitalis95 and @lilyclements note it isn't just the layers that are "empty", but also the facets and (I assume) others. That's because the current Use Graph doesn't try to "read" the ggplot2 object at all. So, I assume we can add to the command (literally add additional `+ `bits, to the command) but maybe not change details of existing layers, etc. If so, we can use the titles and themes tabs, (maybe the axes?) but should not mess with the others - so it would be a limited sub-dialog that is available?

rdstern (approved):
@Ag-Derek I agree this is already a useful step forwards, so good to merge.
@Vitalis95 this should be an easy one?

Vitalis95 (approved):

PR #10500, opened 2026-08-25
 Suppress verbose output from Install R Package dialog (#10484)

ARMSTRONGOPONDO:
@lilyclements @rdstern 
Set iCallType to 0 so pak/install.packages progress is not shown in the output window,
Fixes excessive `pak`/`install.packages` output reported in #10484

### DIFF
--- a/instat/dlgRPackages.vb
+++ b/instat/dlgRPackages.vb
@@ -29,7 +29,7 @@ Public Class dlgInstallRPackage
         Dim dctPackages As New Dictionary(Of String, String)
 
         ucrBase.iHelpTopicID = 592
-        ucrBase.clsRsyntax.iCallType = 2
+        ucrBase.clsRsyntax.iCallType = 0
         ucrInputTextBoxRPackage.SetParameter(New RParameter("pkgs", 1))
         ucrPnlRPackages.AddRadioButton(rdoCRAN)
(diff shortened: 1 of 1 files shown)


lilyclements:
@ARMSTRONGOPONDO can you just check for me - did we always have the `options(...)` bit come up before these changes were made (under Installing from CRAN, not from GitHub):

image

Also, we just want this for the "Install from GitHub" option, not both.

ARMSTRONGOPONDO:
@lilyclements thanks  updated.

**On `options(...)`:** yes, that was already there before this PR. For CRAN installs the dialog has long used `options(warn = 2)` before `install.packages(...)` and `options(warn = 0)` after (since the GitHub option was added). This PR did not introduce that.

**On `iCallType`:** agreed  we only want suppressed output for **Install from GitHub**. The latest commit keeps `iCallType = 2` for CRAN, and sets `iCallType = 0` only when the GitHub option is selected.

lilyclements (approved):
@ARMSTRONGOPONDO great! Thanks for this. @rdstern over to you

rdstern:
@Vitalis95 I see Lily has approved.  Could you therefore check and merge.

Vitalis95 (approved):

PR #10503, opened 2026-08-25
Implements in the Multiple Comparisons dialog plus the autoplot function

bayakarisa:
@rdstern Kindly Check this .
In this Pr here is what I was doing for this dialog. Kindly Check

First is this - In the receivers we need to get the factors from the object and not from the data frame.

> a) The By (Optional): seems a "bigger deal" than I thought earlier. So, could you please add a checkbox at the front, default unchecked. When unchecked the receiver is not shown. (When checked it is as now.)
> b) The code, when the By variable is included, is a bit changed from the obvious. The By is given as you have included already, but the Classify is given as "nitrogen:variety" rather than just variety (so variety within nitrogen, which is what it does.
> c) In the code generated, could you also always include descending = FALSE (or TRUE). (Usually, as you have now, we don't include the default, but here it would help if a user wants to tweak the script.)
> d) Could you also always include the adjust argument, in the command, with the default if it isn't changed.
> e) Could you please reorder the checkboxes on the left, so, under alpha next could be Descending and the third is Adjustment. (This is because they are also used in the other options - pairwise and reference - while the others are not.)
> f) Could you also always include By in the command too. It has By = NULL if not used in the dialog.
> g) Could you add a Graph checkbox, default unchecked. If checked it produces a graph - just as you have done for the One Button dialog. Also, on the same line, if checked, then a Type label becomes visible. This has a drop-down with the 3 options Point, Line, Bar. This generates the additional line biometryassist::autoplot(mct_reg, type="point", label_height = 0.1) as shown in the figure above.
> In the R code please can you always give the type argument and also include the default label_height = 0.1 argument. That will then be easy to edit, if it isn't quite right.
> h) And can you add the Store Graph option - as you have already done in the One Button dialog. This is disabled unless the Graph checkbox is ticked.

------

image

Store Graph is only activated after Generate Plot is checked and the By Receiver is activated only after the checkbox is checked

----------
image

Full dialog with everything checked

-----

image

image

image

image

----------

Also the plot options on Use Graph don't really work . It throws a warning error so it is not just the development version. I tried with version 0.8.16.133
image

### DIFF
--- a/instat/dlgModelMultipleComparisons.Designer.vb
+++ b/instat/dlgModelMultipleComparisons.Designer.vb
@@ -24,8 +24,11 @@ Partial Class dlgModelMultipleComparisons
     Private Sub InitializeComponent()
         Me.lblMultipleMeanComparisonModeltoUse = New System.Windows.Forms.Label()
         Me.lblVariabletoUse = New System.Windows.Forms.Label()
-        Me.lblBy = New System.Windows.Forms.Label()
         Me.btnTransformation = New System.Windows.Forms.Button()
+        Me.ucrSaveGraph = New instat.ucrSave()
+        Me.ucrInputGenerateMultipleComparisonGraphs = New instat.ucrInputComboBox()
+        Me.ucrChkGenerateMultipleComparisonPlot = New instat.ucrCheck()
+        Me.ucrChkByOptional = New instat.ucrCheck()
         Me.ucrInputComboBoxAdjustment = New instat.ucrInputComboBox()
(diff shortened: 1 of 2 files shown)


rdstern (approved):
@bayakarisa that's already very nice.  I know there is more to do  on this dialog, but I am approving this for now.
@Vitalis95 can you check the code so far please.  This would be excellent to merge soon,

Vitalis95 (approved):

PR #10504, opened 2026-08-25
Fixed 2 problems in Prepare > Check Data > Compare Columns dialog

### LINKED ISSUE #10440: Prepare > Check Data > Compare Columns has 2 problems to fix

Vitalis95:
Fixes #10440 

@rdstern @jkmusyoka , fixes  both issue (`a` and `b`)

### DIFF
--- a/instat/dlgCompareColumns.vb
+++ b/instat/dlgCompareColumns.vb
@@ -132,7 +132,6 @@ Public Class dlgCompareColumns
         clsLessorEqualToOperator = New ROperator
         clsDummyOperator = New ROperator
 
-        ucrBase.clsRsyntax.ClearCodes()
         rdoByRow.Checked = True
 
         ucrSelectorCompareColumns.Reset()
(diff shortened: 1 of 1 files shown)


rdstern:
@jkmusyoka you said you would do a bit of R-Instat this week.  Excellent way to spend your leave.  Here should be the fixes to the dialog we had problems with in Zimbabwe.  Can you check?

rdstern:
@jkmusyoka are you going to check this one.  I was hoping it might have been included?

Issue #10436, opened 2026-08-04
Add a One Button dialog to the Experiments menu


Aboubacari:
# Feature Request: Add BKBread functions to the experiments menu

## Is your feature request related to a problem? Please describe.

We frequently receive requests from researchers in the **West Africa Community of Practice (COP)** asking how to perform **multiple mean comparisons** and obtain **significance letters (Compact Letter Display, CLD)** in a simple and intuitive way.

Currently, this workflow requires writing R code or using several manual steps, which can be challenging for many users. We believe that R-Instat could address this need by providing a dedicated dialog under the **Model** menu.

## Describe the solution you'd like

Add a **Multiple Mean Comparison** dialog that supports post-hoc comparisons and automatically generates **Compact Letter Displays (CLD)**.

The dialog could be implemented using the **BKBreed** package (and other compatible packages where appropriate). It should allow users to:

* perform multiple mean comparisons after fitting a model;
* automatically generate significance letters;
* choose among available multiple comparison methods;
* customize tables and graphical outputs;
* export publication-ready results.

## Describe alternatives you've considered

At present, users can perform these analyses by writing R code or by calling functions from packages such as **BKBreed** directly. However, these approaches require familiarity with R and are not accessible to many R-Instat users. A dedicated dialog would provide a simpler, more consistent, and user-friendly workflow.

@rdstern @lilyclements @berylwaswa @jkmusyoka

rdstern:
This suggestion is first to add the BKBreed package to R-Instat.
It is then to become a new dialog called `One Button` in the Experiments > Model menu.
It is after the Multiple Comparisons dialog described in Issue #10433.

One complication is that the package uses its own data sets, but does not supply them as is done by other packages.  They are beiing collected by @Aboubacari into a single Excel file that then needs to be added to the Instat side of the library.  Maybe the file could be called BKBreed and so would not even need to be in a separate directory.

Now to the dialog.  
a) I suggest 2 top radio buttons maybe called `General` and `Special`.
b) Then another set of ordinary radio buttons to choose the design.  The label could be design.  In General the first and default is labelled `rbd`, then `factorial`, then  (to be completed)
Add tootips, so rbd has the tooltip "randomised complete block design", factorial has "two factors and possibly in randomised blocks".  (to be completed)
c) Then our usual data selector.
d) On the right the first, like other dialogs is a **_single receiver_** called `Response Variable`.
e) Underneath the **_second single receiver_** is called `Treatment`
[...]

PR #10515, opened 2026-08-31
Added radio button for Pairwise and Reference

bayakarisa:
Partly solve #10433 @rdstern Kindly Review
Taking from last conversation this was what was wanted and what I have been working on

> a) Can there now be 3 radio buttons at the top, namely Multiple, Pairwise, Reference. The default - is the current dialog, which is Multiple. (Note I describe Pairwise before Reference, but the changes are almost the same for both.)
> b) If Pairwise, then swap multiple in biometryassist::multiple_comparisons( function for pairwise, i.e. biometryassist::pairwise_comparisons( described on page 8 of the manual.
> c) For pairwise the selector, and controls on the top right are the same as for Multiple. The Transform is deleted.
> d) On the left is the new top 3 checkboxes, namely Alpha, Descending and Adjustment. The next 2 (Display Letter and Confidence Intervals are omitted.
> e) Descending now has a third option, namely NULL and that is the default.
> f) descend has a different default, namely holm and tukey is omitted.
> g) The Plot checkbox is included, but the additional pull down (with point, line, bar) is not.
> h) The remaining store, etc are as for multiple.
> i) In the command please also include pairs = NULL, contrasts = NULL. (I need longer to think how they could be added to the dialog. It works ok with the default and the user could tweak them to change.
> 
> Now for Reference:
> 
> Ref_a) As Pairwise but running biometryassist::reference_comparisons( instead, see page 13.
> Ref_b) Selector and controls on right as for others, with no Transform. But new control, below By, added to choose a Reference level from the Label Variable factor. (We have this control in the Climatic > Examine/Edit Data > Daily Data Editing - Select Station control.)
> Ref_c) Same checkboxes on the left as for pairwise. Descding and Plot are exactly the same. The Adjustment now has Dunnet as the default and (like Pairwise) no Tukey.

------------------

image
-----
image

------

image

------
For Pairwise 

image

-----
For Reference

image

### DIFF
--- a/instat/dlgModelMultipleComparisons.Designer.vb
+++ b/instat/dlgModelMultipleComparisons.Designer.vb
@@ -443,7 +443,7 @@ Partial Class dlgModelMultipleComparisons
         Me.MinimizeBox = False
         Me.Name = "dlgModelMultipleComparisons"
         Me.StartPosition = System.Windows.Forms.FormStartPosition.CenterScreen
-        Me.Text = "Comparisons"
+        Me.Text = "Multiple Comparisons"
         Me.ResumeLayout(False)
         Me.PerformLayout()
 
--- a/instat/dlgModelMultipleComparisons.Designer.vb
+++ b/instat/dlgModelMultipleComparisons.Designer.vb
@@ -22,6 +22,10 @@ Partial Class dlgModelMultipleComparisons
     'Do not modify it using the code editor.
     <System.Diagnostics.DebuggerStepThrough()>
     Private Sub InitializeComponent()
+        Me.ucrPnlComparisonType = New instat.UcrPanel()
+        Me.rdoMultiple = New System.Windows.Forms.RadioButton()
+        Me.rdoPairwise = New System.Windows.Forms.RadioButton()
+        Me.rdoReference = New System.Windows.Forms.RadioButton()
         Me.lblMultipleMeanComparisonModeltoUse = New System.Windows.Forms.Label()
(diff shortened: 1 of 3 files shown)


rdstern (changes requested):
@bayakarisa this is excellent.  Very well done, and I hope we can merge very soon.  I only found 2 minor points:
a) Top left the dialog is called Comparisons.  I would prefer keeping the name `Multiple Comparisons`.
b) The two controls below, `Descending and Adjustment` don't "remember" their settings, when returning to the dialog.

Image

bayakarisa:
@rdstern Kindly check as I have addressed the issues

image

rdstern (commented):
@bayakarisa thanks
@lilyclements good to merge please.
More changes are needed, and I suggest the code can be checked then.  They include:
a) Proper choice of treatment factor.  Currently it is from the data frame.  It should be from the object.
b) The transform sub-dialog included.
c) Hopefully more checks that the object is suitable - maybe a Try will be useful.  I first fitted a wrong model with no residual degrees of freedom.  This gave an error - quite understandable, but not the sort of error that helped me find the problem quickly!

lilyclements (approved):

lilyclements:
@bayakarisa does this link to an existing issue? If so, can you link them. Thanks

bayakarisa:
> @bayakarisa does this link to an existing issue? If so, can you link them. Thanks

I have just linked them it was issue #10436

